PR #2, opened 2015-01-13
move course id page to config, add logging for bad course ids

ragesoss:
Nate: might we want to move the wiki page that holds the course ids into the configuration file? I think this is right, but I'm going to avoid pushing substantive changes to master.


ragesoss:
I already added the new application.example.yml bit to the application.yml files for production and master, so it shouldn't break anything to make this change, I think.

PR #15, opened 2015-02-07
Wrap some string literals for i18n

adamwight:
Note that some of the language and capitalization is slightly changed
to reuse strings, but may need to be checked for consistency throughout each
page.

The formatting was also slightly kludged, on labels inside of <small> tags.


ragesoss:
Why do the tag labels need to be kludged like that?

ragesoss:
We can probably use a method like titleize to get those capitalization changes working as intended. The inconsistent capitalization is annoying enough that I wouldn't want to push to WikiEdu production without fixing that.

ragesoss:
This also break the popup for number of students who "have completed training" on the global/cohort view. (It works properly on a single course view, though.)

ragesoss:
I pushed it live to staging, FYI: http://dashboard-testing.wikiedu.org/

ragesoss:
Hmm... the tag kludge also muddles the design some; the font color and size gets changed. Compare testing with production for the header: http://dashboard.wikiedu.org/

PR #17, opened 2015-02-07
wrap more literals

adamwight:
TODO:
* Don't decorate numbers with header tags because that's the incorrect meaning.
  Use a class or something else.


ragesoss:
I'm going to go ahead and merge these, since we're taking a pause in mainline development for the time being, so there's plenty of time in the next two weeks or so to sort out the style issues introduced by these changes.

PR #40, opened 2015-03-08
Basic non-en-wiki support

ragesoss:
This needs a corresponding change in the WikiEduDashboardTools endpoint on labs to handle the lang parameter.

### DIFF
--- a/app/helpers/course_helper.rb
+++ b/app/helpers/course_helper.rb
@@ -12,7 +12,4 @@ module CourseHelper
     value.html_safe
   end
   
-  def base_url
-    "https://#{Figaro.env.wiki_language}.wikipedia.org/wiki/Education_Program:"
-  end
 end
--- a/app/helpers/course_helper.rb
+++ b/app/helpers/course_helper.rb
@@ -11,4 +11,8 @@ module CourseHelper
     end
     value.html_safe
   end
+  
+  def base_url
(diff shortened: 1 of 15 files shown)


ragesoss:
@adamwight can you take a look and point out any insanity you spot?

adamwight:
This looks like a big improvement!  One small thing that could be TODO is to write a function like Article#get_url, rather than building the URL in various places via string manipulation.

ragesoss:
@adamwight I implemented Course#url, which simplifies things a bit. I'm imagining that if/when we get to the point of wanting to handle multiple language wikis in the same database, we'll want to incoporate the language into the Article table. So in the meantime, pulling it from config should be good enough.

ragesoss:
@ntdb Can you take a look at this and merge it if it looks good to you?

ntdb:
@ragesoss will do this afternoon. I'll take the time to build the change to the tools as well.

ragesoss:
@ntdb thanks much! Be sure to add wiki_language: en to your dev application.yml. (I did that pre-emptively on staging and production.)

ntdb:
Everything is wired up! I don't know how the dashboard should behave using a language value other than 'en' so I'll leave the testing/setup of that to you.

ragesoss:
Thanks @ntdb! I'll give it a spin this weekend to see if it actually works.

PR #41, opened 2015-03-11
Rubocop cleanup

ragesoss:
@ntdb Merge if you like it.


ntdb:
@ragesoss if rubocop likes it then so do I. :)

PR #49, opened 2015-03-18
Resolves #48

### LINKED ISSUE #48: Logs show invalid course_id warnings for already-delisted courses
If a course isn't part of the on-wiki cohorts, we shouldn't warn about them.

ragesoss:
@ntdb Does this look right to you?

Only update courses that are listed on-wiki or are listed: true in
the database.


ntdb:
The only problem with this change is that once a course is de-listed it will never be picked back up by the dashboard (if it is later re-listed). Is that an acceptable trade-off until we move courses to our system?

ragesoss:
(Per our chat, relisting should still work.)

PR #69, opened 2015-04-20
Initialize cohorts

ragesoss:
How does this look, @ntdb?


ntdb:
Looks perfect, thanks!

PR #72, opened 2015-04-25
Add check to not remove ArticlesCourses for users with both student and non-student roles.

ragesoss:
Does this make sense to you, @ntdb? Or is this unnecessary in a way I didn't realize?


ntdb:
Looks solid. Good thought about users who have both student and non-student roles: I hadn't considered that case. Thank you!

PR #217, opened 2015-07-18
Doc tweaks

adamwight:


ragesoss:
I'm quite confused about why this build is breaking, since the only differences are to the readme. Let's see if it still breaks after merge.

PR #231, opened 2015-07-28
Prepend language to article title

thenickcox:
@ntdb Do you mind reviewing this? Merging this would depend on your implementation of the DashboardTools endpoint.


public/stylesheets/rtl/main.css
```diff
@@ -1,140 +1,100 @@
-/* /////////////////////////////////////////////////////
```
thenickcox: Not sure about all these changes. I think I picked these up rebasing…

ragesoss: I wouldn't worry about the /rtl/main.css at this point.

This gets built automatically via some of the i18n framework, but we don't actually have users in RTL languages at this point. When we do, I'm sure we'll find plenty of RTL issues, and we can start figuring out the build quirks then.

app/assets/javascripts/components/assignments/assignment.cjsx
```diff
@@ -15,7 +16,8 @@ Assignment = React.createClass(
     className = 'assignment'
     ratingClass = 'rating ' + article.rating
     ratingMobileClass = ratingClass + ' tablet-only'
-
+    languagePrefix = if article.language then "#{article.language}:" else ''
+    formattedTitle = "#{languagePrefix}#{article.title}"
```
thenickcox: @ragesoss Fixed

README.md
```diff
@@ -80,7 +80,16 @@ For analytics (ie, tracking traffic), we use [Piwik](https://github.com/piwik/pi
...
+In order to use and develop the authenticated features of the application (course creation, the assignment design wizard, user management, etc) you will need to create a MediaWiki OAuth consumer).
+
+[Log in to mediawiki](https://www.mediawiki.org/w/index.php?title=Special:UserLogin&returnto=Special%3AUserLogout&returntoquery=noreturnto%3D) with your username and password for mediawiki.org. Once you're logged in, click on "Preferences" in the upper right-hand corner. In the "User profile" tab under "Preferences" (selected by default), change the email address to your email. You'll need this to confirm your account and get your token and secret key in the next step.
```
thenickcox: @ragesoss Removed reference to `application.yml`.

public/stylesheets/rtl/main.css
ntdb: These are probably here because `gulp build` was not run before commit (there's no documentation of this and I didn't explain it, no blame). It's not strictly necessary at any rate.

PR #276, opened 2015-08-07
Only pull revisions for a user on click

thenickcox:
@ntdb and I were looking through the app with a profiler gem he added, and noticed a large amount of time spent in the database on /students. For a large course like [cyberlaw](http://localhost:3500/courses/University_of_California,_Berkeley/Cyberlaw_(Spring_2015)/students) with a large number of active students, the `users` API was returning revisions for all users, which seemed unnecessary if they were all just hidden on load and had to be loaded through clicking.

Our work resulted in a decrease of the request time for the users API by 57%.

Before:
screen shot 2015-08-07 at 9 19 51 am

After:
screen shot 2015-08-07 at 9 22 53 am

_(Note that the difference in overall request time for the page isn't possible to determine until this commit is staged, but there should be a significant difference.)_


ragesoss:
We merged this work from within the main repo.

PR #281, opened 2015-08-10
Wiki edits refactor

ragesoss:
@thenickcox: I'd love to get a code review for this. (I'll probably be showing up at the WINTR office today, FYI.)


lib/wiki_assignment_output.rb
```diff
@@ -0,0 +1,97 @@
...
+    # article page actually exists. This also servces to make sure that we
+    # only post to talk pages of mainspace articles, as we assume that pages
+    # like Talk:Template:Citation or Talk:Wikipedia:Notability do not exist.
+    if initial_page_content.nil?
+      return nil if Wiki.get_page_content(title).nil?
+      initial_page_content = ''
+    end
+
+    # Limit it to live assignments for this article-course
+    title_assignments = title_assignments.select { |a| !a['deleted'] }
```
thenickcox: This would be better expressed as

`title_assignments = title_assignments.reject { |a| a['deleted'] }`

to avoid the negation. It's easier to think of you rejecting the ones that are deleted than selecting the ones that aren't deleted.

lib/wiki_assignment_output.rb
```diff
@@ -0,0 +1,97 @@
...
+  # the set of assigned users for the article for a single course changes.
+  # The strategy here is to only update the tag for one course at a time, so
+  # that the user who updates the assignments for a course only introduces data
+  # for that course. We also want to make as minimal a change as possible, and
+  # to make sure that we're not disrupting the format of existing content.
+  def self.build_assignment_page_content(new_tag,
+                                         course_page,
+                                         page_content)
+
+    dashboard_url = Figaro.env.dashboard_url
```
thenickcox: Using `Figaro.env` couples us to that gem. I've been meaning to bring this up. Many gems implement an interface for accessing `ENV` variables. Because `ENV['dashboard_url']` would also work, I think we should switch to that so that we could, if need be, switch to [dotenv](https://github.com/bkeepers/dotenv). Ideally we wouldn't be unnecessarily coupled to an environment gem.

lib/wiki_assignment_output.rb
```diff
@@ -0,0 +1,97 @@
...
+    # Return if tag already exists on page
+    unless new_tag.blank?
+      return nil if page_content.include? new_tag
+    end
```
thenickcox: How about `return if new_tag.blank? || page_content.include?(new_tag)`

lib/wiki_assignment_output.rb
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  ###################
+  # Helper methods #
+  ###################
+  def self.assignments_tag(course_page, title_assignments)
+    return '' if title_assignments.empty?
+
+    # Make a list of the assignees, role 0
+    tag_assigned = build_wikitext_user_list(title_assignments, 0)
```
thenickcox: Can we store `role 0` and `role 1` as named constants so it's clear what they represent and we don't have any [magic numbers](https://sourcemaking.com/refactoring/replace-magic-number-with-symbolic-constant)?

lib/wiki_edits.rb
```diff
@@ -284,15 +235,14 @@ def api_post(data, tokens)
       #    {"code"=>"protectedpage",
       #     "info"=>"The \"templateeditor\" right is required to edit this page",
       #     "*"=>"See https://en.wikipedia.org/w/api.php for API usage"}}
-
       if response_data['error']
-        raise StandardError.new response_data['error']['code']
+        raise StandardError.new response_data['error']['info']
```
thenickcox: I'm not convinced `StandardError` is the most semantic class for this error.

lib/wiki_assignment_output.rb
```diff
@@ -0,0 +1,97 @@
...
+    end
+
+    # Check for existing tags and replace
+    old_tag_ex = "{{course assignment | course = #{course_page}"
+    new_tag_ex = "{{#{dashboard_url} assignment | course = #{course_page}"
+    page_content.gsub!(/#{Regexp.quote(old_tag_ex)}[^\}]*\}\}/, new_tag)
+    page_content.gsub!(/#{Regexp.quote(new_tag_ex)}[^\}]*\}\}/, new_tag)
+
+    # Add new tag at top (if there wasn't an existing tag already)
+    if !page_content.include?(new_tag) && !new_tag.blank?
```
thenickcox: We already `return`ed if `page_content.include?(new_tag)`, right?

lib/wiki_edits.rb
```diff
@@ -157,32 +117,23 @@ def self.notify_users(current_user, recipient_users, message)
...
-               page: course_page_slug,
-               section: '0',
-               prop: 'wikitext',
-               format: 'json' }
-
-    response = api_post params, tokens
-    puts response.body
-    response.body
-  end
+  def self.assignments_by_article(course, assignments, delete)
```
thenickcox: Can this take a hash instead of three arguments that are opaque when called? See below in the test.

spec/lib/wiki_assignment_output_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+                        id: 10001,
+                        title: 'Course Title',
+                        school: 'School',
+                        term: 'Term',
+                        slug: 'School/Course_Title_(Term)')
+        talk_title = 'Talk:Selfie'
+        selfie_talk = Wiki.get_page_content(talk_title)
+
+        course_page = course.wiki_title
+        assignment_titles = WikiEdits.assignments_by_article(course, nil, nil)
```
thenickcox: Having two `nil` arguments here confuses me. Why not just default them to nil and make them optional?

spec/lib/wiki_assignment_output_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+        title_assignments = assignment_titles['Selfie']
+        assignment_tag = WikiAssignmentOutput.assignments_tag(course_page,
+                                                              title_assignments)
+        page_content = WikiAssignmentOutput
+                       .build_assignment_page_content(assignment_tag,
+                                                      course_page,
+                                                      selfie_talk)
+        pp page_content
```
thenickcox: Did you mean to leave this debugging output in here?

lib/wiki_assignment_output.rb
ragesoss: @thenickcox sounds great. I'll make that change here, and we can start doing the same elsewhere as we go.

lib/wiki_assignment_output.rb
ragesoss: @thenickcox Because if new_tag is the empty string, we actually don't want to return here. new_tag is empty string when we need to remove the existing tag (this is the delete=true situation from elsewhere), so if it's the empty string, then we want to proceed with the replacement of the old tag with the empty string.

During this refactor, I waffled at several points over completely separating the 'delete' flow from the normal flow, but I decided to work around it for now.

lib/wiki_assignment_output.rb
ragesoss: That was before the gsub, in which we might have replaced an old tag with the new tag. This block is only for the case where we didn't already find somewhere else to put the new tag, so we want to add it to the top section of the page.

lib/wiki_edits.rb
ragesoss: My reasoning was that they are set from .update_assignments (where they do default to nil and false) — and this method is tightly coupled to .update_assignments — so it seemed clearer to rely on those being passed along. I trust your judgment, but I don't understand why setting defaults in both methods would be better. Can you explain your reasoning?

ragesoss: Ah, I guess the point is mainly relevant for the corresponding specs, where these methods get handled in isolation?

lib/wiki_edits.rb
ragesoss: Agreed. We haven't done any error classes yet, so I was waiting for someone else to set the standard for how they should be done.

lib/wiki_assignment_output.rb
thenickcox: Will `new_tag` become blank after the `gsub`? It doesn't look like it. If not, we can remove that second check.

lib/wiki_assignment_output.rb
ragesoss: This makes total sense. (We have magic numbers in many places.)

What's the best place to store such magic numbers? Should these be global constants, or defined within the class?

lib/wiki_edits.rb
thenickcox: In that case, how about

```ruby
raise ExecptionError.new(response_data['error']['info'])
...
class ResponseError < Exception
end
```

lib/wiki_assignment_output.rb
thenickcox: I typically define them at the top of the class.

```ruby
STUDENT_ROLE_TYPE = 1
...
tag_assigned = build_wikitext_user_list(title_assignments, STUDENT_ROLE_TYPE)
```

A great use case for these is to make explicit what those role types are for new devs. Like, I don't actually know if student is 1.

PR #375, opened 2015-09-13
Course update manager

ragesoss:
@thenickcox I experimented here with removing some of the responsibilities of Course into a separate class. Is this a good way to do it?

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -75,7 +75,9 @@ class Course < ActiveRecord::Base
   scope :legacy, -> { where('courses.id <= ?', LEGACY_COURSE_MAX_ID) }
   scope :not_legacy, -> { where('courses.id > ?', LEGACY_COURSE_MAX_ID) }
 
-  scope :unsubmitted_listed, -> { where(submitted: false).where(listed: true).merge(Course.not_legacy) }
+  scope :unsubmitted_listed, -> {
+    where(submitted: false).where(listed: true).merge(Course.not_legacy)
+  }
 
   scope :listed, -> { where(listed: true) }
 
@@ -92,12 +94,12 @@ class Course < ActiveRecord::Base
   # A course stays "current" for a while after the end date, during which time
   # we still check for new data and update page views.
   scope :current_and_future, lambda {
-    update_length = Figaro.env.update_length.to_i.days.seconds.to_i
+    update_length = ENV['update_length'].to_i.days.seconds.to_i
     where('end > ?', Time.now - update_length)
   }
 
   before_save :order_weeks
-  validates :passcode, presence: true, unless: :is_legacy_course?
+  validates :passcode, presence: true, unless: :legacy?
 
   def self.submitted_listed
     Course.includes(:cohorts).where('cohorts.id IS NULL')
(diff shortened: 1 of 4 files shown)


app/models/course.rb
```diff
@@ -137,62 +134,60 @@ def delist
...
   end
 
   def students_without_instructor_students
     students.where.not(id: instructors.pluck(:id))
   end
 
+  def new_articles
+    articles_courses.live.new_article
+      .joins(:article).where('articles.namespace = 0')
+  end
```
thenickcox: Any reason this is an instance method? It reads more like a class method.

app/models/course.rb
```diff
@@ -137,62 +134,60 @@ def delist
...
   def new_article_count
-    self[:new_article_count] || articles_courses.live.new_article
-      .joins(:article).where('articles.namespace = 0')
-      .size
+    return_or_calculate :new_article_count
+  end
+
+  def return_or_calculate(attribute)
+    update_cache unless self[attribute]
+    self[attribute]
```
thenickcox: Nice metaprogramming. :tada:

lib/course_cleanup_manager.rb
```diff
@@ -0,0 +1,25 @@
+#= Class for callback cleanup methods for Courses
+class CourseCleanupManager
+  def self.cleanup_articles(course, user)
+    # find which course articles this user contributed to
+    user_articles = user.revisions
```
thenickcox: [extract temp to query](http://www.refactoring.com/catalog/replaceTempWithQuery.html)

lib/course_cleanup_manager.rb
```diff
@@ -0,0 +1,25 @@
...
+                    .where('date >= ? AND date <= ?', course.start, course.end)
+                    .pluck(:article_id)
```
thenickcox: [extract temp to query](http://www.refactoring.com/catalog/replaceTempWithQuery.html)

lib/course_cleanup_manager.rb
```diff
@@ -0,0 +1,25 @@
...
+    # have these articles been edited by other students in this course?
+    to_delete = []
+    possible_deletions.each do |pd|
+      other_editors = Article.find(pd).editors - [user.id]
+      course_editors = course.students & other_editors
+      to_delete.push pd if other_editors.empty? || course_editors.empty?
+    end
+
+    # remove orphaned articles from the course
+    course.articles.delete(Article.find(to_delete))
```
thenickcox: `delete` won't use callbacks. I think you want `destroy` here. [source](http://stackoverflow.com/a/22757533/931934)

lib/course_update_manager.rb
```diff
@@ -0,0 +1,66 @@
...
+    if data.blank?
+      data = CourseImporter.get_course_info id
+      return if data.blank? || data[0].nil?
+      data = data[0]
+    end
+    # Symbol if coming from controller, string if from course importer
+    course.attributes = data[:course] || data['course']
+
+    return unless save
+    if data['participants']
```
thenickcox: Can data['participants'] be `nil`? If it's an empty array, you don't need a check. (If it's not an empty array, it should be.)

```
[8] pry(main)> [].each_with_index do |x, i|
[8] pry(main)*   puts x
[8] pry(main)*   puts i
[8] pry(main)* end
```

thenickcox:
I'm gonna merge this.

PR #404, opened 2015-09-28
Recent edits feed

ragesoss:
@thenickcox I got this working, and I merged and deployed it onto staging. But I'm still nervous, so I'd like you to take a look before moving this to master/production

### DIFF
--- a/app/assets/javascripts/components/activity/activity_table.cjsx
+++ b/app/assets/javascripts/components/activity/activity_table.cjsx
@@ -1,0 +1,85 @@
+React     = require 'react/addons'
+UIStore   = require '../../stores/ui_store'
+UIActions = require '../../actions/ui_actions'
+ActivityTableRow = require './activity_table_row'
+TransitionGroup = require '../../utils/TransitionGroup'
+
+ActivityTable = React.createClass(
+  displayName: 'ActivityTable'
+  getInitialState: ->
+    activity: @props.activity
+  clearAllSortableClassNames: ->
+    Array.prototype.map.call document.getElementsByClassName('sortable'), (el) ->
+      el.classList.remove('asc')
+      el.classList.remove('desc')
+  sortItems: (e) ->
+    sortOrder = if e.target.classList.contains('asc') then 'desc' else 'asc'
+    @clearAllSortableClassNames()
+    e.target.classList.add(sortOrder)
+    key = e.target.dataset.sortKey
+    activities = _.sortByOrder(@state.activity, [key])
+    activities = activities.reverse() if sortOrder is 'desc'
+    @setState @state.activity = activities
+
+  render: ->
+    activity = @state.activity.map (revision) =>
+      roundedRevisionScore = Math.round(revision.revision_score) or 'unknown'
+      revisionDateTime = moment(revision.datetime).format('YYYY/MM/DD h:mm a')
+      talkPageLink = "https://en.wikipedia.org/wiki/User_talk:#{revision.user_wiki_id}"
+
+      <ActivityTableRow
+        key={revision.key}
+        rowId={revision.key}
+        articleUrl={revision.article_url}
+        diffUrl={revision.diff_url}
+        talkPageLink={talkPageLink}
+        title={revision.title}
+        revisionScore={roundedRevisionScore}
+        author={revision.user_wiki_id}
+        revisionDateTime={revisionDateTime}
+      />
+
+    drawers = @state.activity.map (revision) ->
+      courses = revision.courses.map (course) ->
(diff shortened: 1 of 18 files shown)


thenickcox:
@ragesoss So, a lot of this looks great. In particular, it looks like you're really getting your head around React, Flux, and the relationship between stores, actions, and components.

A few things. Firstly, you'd certainly know better than I would, but getting the last 200 revisions seems like a lot. Is that helpful? It takes a while to load, and there's a lot of data there. If that's what's needed, great, but it seems like a _lot_ of data. It's probably fine for admins, but if we did something like this for other users, I'd want to paginate or chunk the results, or do something so that it loaded faster.

Secondly, and this is pretty minor, but the Revision Date/Time sorting seems weird; it only seems to sort every 2 clicks. I mostly blame myself, because I wrote those components so that the sorting happens through setting the state in the components directly because I couldn't figure out Nate's UI store, but his way is definitely better. For now, though, this probably works, since it's an admin-only feature.

Lastly, when you wrote some of the suspected plagiarism component, I refactored that and the DYK-eligible components to use an `ActivityTableRow` component, which is great, but now all three still have a good bit of shared behavior, in particular `clearAllSortableClassNames`, and the sorting method itself, which tends to only differ based on the model name, but we could make that dynamic. These should probably be broken out into a separate object. (You can see an example of that on commit 8ecf46843e706c757f7d5ac5f1fd92eed298915b.) 

And stepping back a level, their `render` methods are almost identical: get the models and the drawers, and `zip` them so that they alternate. Then get the headers, and put it all into a table. This could – and should – be refactored into a more abstract component that can handle all.

Other than that, though, if this does what you want, :+1: .

thenickcox:
Also, it's probably a good idea to write a smoke test that hits this page and makes sure it renders in case we break it. You can add it to `spec/features/activity_page_spec.rb`.

ragesoss:
Thanks @thenickcox!

I tried more ambitious things, like basically adapting the RevisionsLlist from courses, but I got lost and had to back up to basically copying the format from the other to recent-activity pages. It's making a little more sense, but it's still way more moving parts than I can think about at once.

I'll add that test, and then merge.

thenickcox:
@ragesoss Do you want help with the refactor?

ragesoss:
I'll take a shot at it. Thanks!

app/assets/javascripts/components/activity/activity_table.cjsx
```diff
@@ -0,0 +1,83 @@
...
+        component='tbody'
+        enterTimeout={500}
+        leaveTimeout={500}
+      >
+        {elements}
+      </TransitionGroup>
+    </table>
+)
+
+module.exports = ActivityTable
```
thenickcox: This is _exactly_ what I had in mind! :tada:

app/assets/javascripts/components/activity/did_you_know_handler.cjsx
```diff
@@ -21,85 +19,20 @@ DidYouKnowHandler = React.createClass(
...
-        leaveTimeout={500}
-      >
-        {elements}
-      </TransitionGroup>
-    </table>
+    <ActivityTable
+      activity={@state.articles}
+      headers={headers}
+      noActivityMessage={noActivityMessage}
+    />
```
thenickcox: Look how much duplication that took out. Awesome!

app/assets/javascripts/components/activity/recent_edits_handler.cjsx
```diff
@@ -0,0 +1,39 @@
...
+      { title: 'Revision Author',    key: 'user_wiki_id' },
+      { title: 'Revision Date/Time', key: 'revision_datetime' },
+    ]
+    noActivityMessage = 'Loading recent edits...'
+
+    <ActivityTable
+      activity={@state.revisions}
+      headers={headers}
+      noActivityMessage={noActivityMessage}
+    />
```
thenickcox: This `render` method is so slim!

spec/features/activity_page_spec.rb
```diff
@@ -77,17 +76,27 @@
 
         it 'displays a list of revisions suspected of plagiarism' do
           view_plagiarism_page
-          assert_page_content article.title.gsub('_', ' ')
+          assert_page_content article.title.tr('_', ' ')
```
thenickcox: Curious about this change.

ragesoss: When I upgraded rubocop, it started throwing a warning about this, saying to replace gsub with tr. Apparently, it's a faster, less flexible string method: it only replaces a single character with another single character.

thenickcox: @ragesoss Huh.
giphy

PR #427, opened 2015-10-18
Hard-coded strings to i18n strings

Kenua:
Replaced in the main application layout, RevisionList and CourseCreator.


ragesoss:
@Kenua Hi!

You've gotten started in the right direction with this, but the corresponding strings need to be added to config/locale/en.yml (and qqq.yml) for it to work as intended.

The idea here is that the `I18n.t` function uses the strings encoded in the locale files to insert the appropriate language string. See, for example, the use of it here: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/components/students/student_list.cjsx#L52

The `I18n.t` function takes `users.name` and looks up the corresponding string for a given language: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/config/locales/en.yml#L183

So here we need to take each of those English strings, and add them to the en.yml file, and then call them by their message keys instead of the literal English strings we use now.

ragesoss:
@Kenua Were you able to get a development environment up and running? If so, you can try out your changes by running `rake i18n:js:export` on the command line. That command will regenerate the I18n javascripts from the locale .yml files, so that you can verify that the strings are showing up as expected when you're using the app.

One other little details: be sure to fetch the latest version of `master` and rebase your changes onto master. Let me know if I can help with that.

Kenua:
Thank so much for instruction, hope now I really understand what to do with this function. I still have a problem with installing capybara-webkit (whatever it is) and so can't run rake i18n:js:export, but I'll cope with it as soon as possible.

Kenua:
Hi! Tell, please, does an another way to verify correctness of my changes
in code exist? I've made my tasks (and even a little more) according to
your instruction long ago, but still can't deal with a development
environment. Virtual machines work too slowly on my old netbook, so I tried
to install all these things on Windows and was very close to success, but a
little bug in Qt installation embitters me for about a week. A lot of time
was spent trying to fix it, and I'm afraid I will not have sufficient time
for Dashboard tasks if keep doing. So can I do anything?

ragesoss:
@Kenua Getting a development environment running would be essential to working on this as an Outreachy internship. Unfortunately, I don't have any experience with trying to run a Rails development environment in Windows, but from what I've read, it's not easy. Your best option might be to install Ubuntu on your netbook.

Kenua:
@ragesoss Please, tell  something about the last version of RevisionList, Course Creator, and main application layout.

ragesoss:
@Kenua Hi! I didn't see the new commits until your comment yesterday... sorry about that delay. I'll take a look soon... Monday, if not today.

Were you able to get a development environment up and running? If so, did you try it out and verify that the substituted strings are working as expected?

Kenua:
@ragesoss Yes, I can't believe, but I've installed all these programs, rechecked the written and debugged it. Now it must work.

And one question: should I commit .js-files?

ragesoss:
Merged as #449.

@Kenua No, the .js files don't get committed. Those files get generated from the `gulp build` process, which runs as part of deployment, but we don't check them in.

PR #446, opened 2015-11-01
Adding support for I18N to did_you_know handler.cjsx

WinnySilva:
Adding support for I18N to did_you_know_handler.cjsx. The files en.yml
and qqq.yml have been modified to support the addition.

### DIFF
--- a/app/assets/javascripts/components/activity/did_you_know_handler.cjsx
+++ b/app/assets/javascripts/components/activity/did_you_know_handler.cjsx
@@ -25,13 +25,13 @@ DidYouKnowHandler = React.createClass(
 
   render: ->
     headers = [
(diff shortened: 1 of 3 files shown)


ragesoss:
@WinnySilva nice work! Were you able to get a development environment up and running and test it out?

The one improvement I'd like to see before merging would be adjust the string titles to match the snake-case convention of the rest of the i18n stringe. I'd suggest `did_you_know` (leaving off the 'handler' part, since that's an implementation detail that doesn't matter to the user), and then I'd suggest `no_activity` in place of `noActivityMessage`.

ragesoss:
Winny, I'm sorry that the Outreachy project didn't work out, but thanks for getting involved! I hope you'll apply again in the future!

PR #449, opened 2015-11-03
Hard-coded strings are replaced with i18n strings in the main applica…

ragesoss:
…tion layout, RevisionList and CourseCreator.


ragesoss:
@Kenua nice work! I've rebased your pull request onto master, squashed it into a single commit, and cleaned up a few little things. I'm going to merge it now.

The one thing that should be improved is the descriptions in the qqq file. In some cases, the descriptions you've added basically just repeated the message text. Instead, they should describe how those strings are used within the interface.

ragesoss:
Merged.

PR #461, opened 2015-11-10
Add the github deploy key as an attempt to fix the ssh forwarding issue.

bmathews:
Unfortunately we won't know if this works until we push the change to staging or production.

The issue I'm trying to solve here is that the server does not have access to our github repo via ssh from travis because ssh forwarding is on in capistrano. The travis build environments aren't forwarding that access along. This change adds a github deploy key to travis's ssh agent that _should_ forward onto the server.


ragesoss:
@bmathews feel free to push this to staging. If you do so and it works, then we should revert staging to its current situation, which is an up-to-date version of `training_tool_rebase`.

bmathews:
Will do! Thanks!

bmathews:
https://travis-ci.org/WikiEducationFoundation/WikiEduDashboard/builds/90225562#L2766

bmathews:
@ragesoss @thenickcox Let me know if you see any issues with the capistrano deployments finally running automatically on the staging/production branches!

PR #475, opened 2015-11-14
Parallel travis specs

ragesoss:
@thenickcox @bmathews 

This puts the slow feature specs into a separate build. We could take it further and run (say) the course creation specs in on build, other feature specs in another build, and all the rest of the (fast) tests in another build.

The one significant wrinkle is automated deployment. I solved this by attaching the deployment script only to the unit-test build. Since we'll generally only be pushing to production if a build already passed on master, this shouldn't really be a problem (and it also makes deployment reliable, even if some of the feature tests fail intermittently).

What do you two think about this approach?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,12 @@
+# Run feature specs in a parallel build from the rest, to speed up the process.
+# Only run the deployment script on the faster non-feature branch.
+# We should only be pushing healthy builds to staging or production anyway.
+env:
+  - TEST_SUITE='~type:feature'
+    AFTER_SUCCESS='./travis-deploy.sh'
+  - TEST_SUITE=type:feature
+    AFTER_SUCCESS=''
+
 language: ruby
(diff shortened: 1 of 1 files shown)


spec/features/activity_page_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 
-describe 'activity page', type: :feature, js: true do
+describe 'activity page', type: :feature, js: true, feature: true do
```
thenickcox: Was there not a way to do this besides the `feature: true` hash? That seems redundant after `type: :feature`. Is there not a way to hook into the `type`?

ragesoss: Yep. I just didn't look far enough to find it. New version does that: `--tag type:feature`.  If we want to break up those feature specs into a bunch of different categories for even more builds, we may need to add some tags though.

Issue #371, opened 2015-09-10
Show a placeholder image for uploads that have been removed


bmathews:
We should handle thumbnails for removed images/uploads better. 

http://dashboard-testing.wikiedu.org/courses/University_of_Detroit,_Mercy/Social_Psychology_(Summer_2015)/uploads

![image](https://cloud.githubusercontent.com/assets/848347/9793143/d8b5dfa6-5798-11e5-8805-3c99e76af71c.png)

PR #476, opened 2015-11-15
Track deleted files

### LINKED ISSUE #403: dashboard should recognize when uploads have been deleted
We currently don't handle deleted images well. From a UI perspective, see #371. But We should also update the database to reflect deleted status (and not show such images at all in the feed), so that we can keep track of images that got uploaded and then deleted.

ragesoss:
The first step is to get this up and running on staging, and make sure it works and doesn't hurt performance.

Then, we can adjust the Uploads tab of courses to only show non-deleted files, and adjust the Analytics routines to account for deleted files.

A goal for later on will be to surface info about deleted files — as well as deleted articles — to users (see #371).

Resolves #403

### DIFF
--- a/lib/commons.rb
+++ b/lib/commons.rb
@@ -30,6 +30,14 @@ class Commons
     usages
   end
 
+  def self.find_missing_files(commons_uploads)
+    missing_query = build_info_query(commons_uploads)
+    pages = get_image_data(missing_query, 'pageid', '')
+    missing_pages = pages.select { |page| page['missing'] }
+    missing_page_ids = missing_pages.map { |page| page['pageid'] }
+    commons_uploads.select { |file| missing_page_ids.include? file.id }
+  end
+
   def self.get_urls(commons_uploads)
     url_query = build_url_query commons_uploads
     file_urls = get_image_data(url_query, 'imageinfo', 'iicontinue')
@@ -82,6 +90,14 @@ class Commons
                     continue: ''
                   }
     usage_query
+  end
+
+  def self.build_info_query(commons_uploads)
+    file_ids = commons_uploads.map(&:id)
+    info_query = { pageids: file_ids,
+                   continue: ''
+                 }
+    info_query
   end
 
   def self.build_url_query(commons_uploads)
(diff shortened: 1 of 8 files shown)


ragesoss:
@thenickcox a once-over check of this would be appreciated.

lib/commons.rb
```diff
@@ -84,6 +92,14 @@ def self.build_usage_query(commons_uploads)
     usage_query
   end
 
+  def self.build_info_query(commons_uploads)
+    file_ids = commons_uploads.map(&:id)
+    info_query = { pageids: file_ids,
```
thenickcox: Why not

```
file_ids = commons_uploads.map(&:id)
{ page_ids: file_ids, continue: '' }
```
?

lib/importers/upload_importer.rb
```diff
@@ -19,6 +19,18 @@ def self.update_usage_count(commons_uploads)
     end
   end
 
+  def self.find_deleted_files(commons_uploads)
+    Utils.chunk_requests(commons_uploads) do |file_batch|
+      deleted_files = Commons.find_missing_files file_batch
```
thenickcox: What is the data type of `deleted_files` here?

ragesoss: deleted_files is an array of CommonsUploads. Commons.find_missing_files takes an array of CommonsUploads, and selects the ones that the Wikimedia Commons API reports as 'missing'.

lib/commons.rb
ragesoss: I find this pattern a little easier to make sense of. It connects the output of the method with the context in which it gets used. For example, in `self.get_usages`, the first line is `usage_query = build_usage_query commons_uploads`

It makes it easier to quickly see 'oh, yeah, something called "usage_query" is exactly what comes out of `build_usage_query`.

It's not a big deal either way, but to me, being a little more explicit by naming the return value is worth the extra line of code.

(More immediately, the answer for this commit is: "because that's the pattern established in the rest of this class.")

lib/importers/upload_importer.rb
thenickcox: That's what I thought. Then can we do `deleted_files.each { |file| file.update_attribute(:deleted, true) }`?

ragesoss: Done.

Issue #317, opened 2015-08-22
Dashboard does not inform user when they need to reauthorize OAuth


ragesoss:
In some cases, a user's OAuth credentials may become invalid, causing the WikiEdits.get_tokens method (and the subsequent edit attempt) to fail. If get_tokens fails with the error `mwoauth-invalid-authorization` then we should prompt the user to reauthorize.

ragesoss:
A first pass at this has been implemented; the user gets logged out and there is a flash message informing the user of that they need to log in again.

However, the flash message doesn't play well with React. As long as the user is still on the same course page (and therefore within the same react-root, I guess?) the page will not render the flash message or show the user that they have been logged out.

How can we push the error message and logged_out status to the user through React?

thenickcox:
@ragesoss bump. @bmathews seemed to address this at least. Any further action required?

ragesoss:
@thenickcox There's still a ways go with this:
* The frontend does not always put up a notification when an auth issue has happened. It works for enrolling a student, but it does not seem to be working for Timeline or Description edits.
* Even if the user gets the notification, the page does not give a clear indication of what is wrong and what they should do, and it does not show that they are logged out. A user can dismiss the notification, and continue navigating through the tabs and taking other actions — all of which will fail because they are no longer logged in.
![error](https://cloud.githubusercontent.com/assets/848483/12098912/083fbd7c-b2da-11e5-8e99-3027abd9b9f2.png)

Here's a flow for testing it:
* Load one of the courses you can edit on dashboard-testing.wikiedu.org. Submit it if it's not submitted already.
* In another tab, revoke access for dashboard-testing.wikiedu.org: https://test.wikipedia.org/wiki/Special:OAuthManageMyGrants
* Make an edit to the course that will trigger a wiki edit (such as editing the description, or timeline, or enroll a student).
* There *should* be an error notification for any of those actions, and the UI should no longer show the user as logged in.

Currently, there is no notification for some cases — I can only get a notification for the 'enroll a student' case — and the UI still shows the user as logged in.

thenickcox:
Let's talk about an approach for this. GMail has a UI for this, where you sign out in another tab, then switch back and it pops up an alert and signs you out. Implementation-wise, it seems to poll some sort of auth endpoint to make sure you're still logged in. Not totally sold on that approach, but that's certainly one way to do it. Thoughts?

ragesoss:
[...]

PR #478, opened 2015-11-16
Improve UI error handling

### LINKED ISSUE #378: Course page should inform user when a save action fails
When a course page or timeline or other post of course data fails, the UI still behaves as if it succeeded; the updated data is reflected in the rendered view (until the page is refreshed and the old data returns), and the UI does not inform the user that their action failed.

Even just a popup and refresh... or a popup and displaying the unsaved data... would be a nice improvement, and help us make sure that users are aware of, and report, when the system is failing to save.

### LINKED ISSUE #380: 'e.responseJSON is undefined' javascript error in Firefox
This one appears to be exclusive to Firefox: http://sentry.ragesoss.com:8080/wiki-ed/production/group/577/

bmathews:
Resolves #378. All API related errors show notifications now.
Resolves #380. Non-JSON responses are handled better now.
Partially addresses #317. The UI will show an 'Unauthorized' notification.

### DIFF
--- a/app/assets/javascripts/components/common/notifications.cjsx
+++ b/app/assets/javascripts/components/common/notifications.cjsx
@@ -23,7 +23,7 @@ Notifications = React.createClass(
   _handleClose: (notification) ->
     NotificationActions.removeNotification(notification)
 
-  renderNotification: (notification, i) ->
+  _renderNotification: (notification, i) ->
     if notification.type is "error"
       message = (
         <p>
@@ -46,7 +46,7 @@ Notifications = React.createClass(
     </div>
 
   render: ->
-    notifications = @state.notifications.map (n, i) => @renderNotification(n, i)
+    notifications = @state.notifications.map (n, i) => @_renderNotification(n, i)
 
     <div className="notifications">
       {notifications}
--- a/app/assets/javascripts/components/common/notifications.cjsx
+++ b/app/assets/javascripts/components/common/notifications.cjsx
@@ -1,30 +1,52 @@
 # Requirements
 #----------------------------------------
 
-React = require 'react'
+React             = require 'react'
+NotificationStore = require '../../stores/notification_store'
+NotificationActions = require '../../actions/notification_actions'
 
 
 # Component
 #----------------------------------------
 
 Notifications = React.createClass(
-  render: ->
-    notifications = []
-    
-    if @props.notifications
-      notifications = @props.notifications.map (notification, i) =>
-        if notification.type is "error"
-          message = (
-            <p>
-              <strong>There was an error:</strong> {notification.message}
-            </p>
-          )
 
-        <div key={i} className='notice'>
-          <div className='container'>
-            {message}
-          </div>
-        </div>
+  mixins: [NotificationStore.mixin]
+
+  getInitialState: ->
+    notifications: NotificationStore.getNotifications()
+
+  storeDidChange: ->
+    @setState
+      notifications: NotificationStore.getNotifications()
+
+  _handleClose: (notification) ->
+    NotificationActions.removeNotification(notification)
+
+  renderNotification: (notification, i) ->
+    if notification.type is "error"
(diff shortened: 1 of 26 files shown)


bmathews:
@thenickcox review please!

app/assets/javascripts/components/common/notifications.cjsx
```diff
@@ -1,30 +1,52 @@
...
+    notifications: NotificationStore.getNotifications()
+
+  storeDidChange: ->
+    @setState
+      notifications: NotificationStore.getNotifications()
+
+  _handleClose: (notification) ->
+    NotificationActions.removeNotification(notification)
+
+  renderNotification: (notification, i) ->
```
thenickcox: I like your convention of underscoring method names that aren't part of the lifecycle, like `_handleClose` above. Why not for `renderNotification`?

bmathews: Oversight. Fixed, thanks!

app/assets/javascripts/utils/router.cjsx
```diff
@@ -47,15 +49,9 @@ routes = (
...
-        <Redirect from="overview" to="overview" />
       </Route>
-      <DefaultRoute handler={CourseCreatorButton} />
     </Route>
-    <Route handler={CourseCreatorButton}>
-      <Route name='course_creator' path='course_creator' handler={CourseCreator} />
-    </Route>
-    <DefaultRoute handler={CourseCreatorButton} />
-
+    <Route name='course_creator' path='course_creator' handler={CourseCreator} />
```
thenickcox: Thanks for cleaning this up! :+1:

spec/features/activity_page_spec.rb
```diff
@@ -62,7 +62,7 @@
       it 'filters the courses to my courses' do
         # Admin is admin of course 1, should only see user1's revision
         # when checked
-        click_link 'Recent Activity'
+        visit '/recent-activity'
```
thenickcox: I could totally see this going either way, and I'm sure your reasoning is good for doing this. One thing I like about the `click_link` approach is that, if that link accidentally gets deleted, this covers that. Of course, one annoying thing is if you change it from 'Recent Activity' to 'Recent activity', this fails. What's your reasoning?

spec/rails_helper.rb
```diff
@@ -21,6 +22,9 @@
   config.allow_url 'cdn.ravenjs.com'
 end
 
+Capybara::Screenshot.webkit_options = { width: 1400, height: 768 }
+Capybara.save_and_open_page_path = "tmp/screenshots/"
```
thenickcox: We should consider having a cron task that clears this directory out every so often so it doesn't get bloated.

bmathews: Just need to add `Capybara::Screenshot.prune_strategy = :keep_last_run`.

I only considered it running on travis, where it'll automatically destroy everything. But it could fill up locally after a while.

test/components/common/notifications.coffee
```diff
@@ -0,0 +1,51 @@
...
+
+    rendered = ReactTestUtils.renderIntoDocument(
+      <Notifications />
+    )
+
+    notification =
+      responseJSON:
+        error: 'Test error'
+
+    rows = ReactTestUtils.scryRenderedDOMComponentsWithClass rendered, 'notice'
```
thenickcox: Is this a jest test?

spec/features/activity_page_spec.rb
bmathews: The headless webkit browser wasn't able to click the link due to the `position: fixed` header nav covering it up after some style changes I made. This was the easier fix, but I can switch to using selenium for this test instead if you'd rather.

test/components/common/notifications.coffee
bmathews: Nope, mocha

PR #482, opened 2015-11-17
Check for nil values of slug_index.

### LINKED ISSUE #479: error in TrainingProgressManager upon creating course
I hit this on staging: http://sentry.ragesoss.com:8080/wiki-ed/staging/group/1342/

I got an error upon submitting the wizard.

ragesoss:
If last_slide_completed is a slide that is no longer part of the module, then slug_index will return nil. That's not good.

Resolves #479 

@thenickcox make sense?


thenickcox:
Wait, why would this happen?

thenickcox:
Also, I like that you made the `module_started?` method, but it seems like that should be private.

ragesoss:
> Wait, why would this happen? What do you mean "no longer part of the module"?

Say I have a module that I got all the way to slide.id 504 on.  Now that module gets changed, so that slide 504 is not part of the module. The TMU record knows that last_slide_completed for that module is 504, but since that's no longer part of the module, slug_index returns nil... because `@training_module.slides.collect(&:slug).index(slug)` does not find the corresponding slug. Or something. I didn't check how .last_slide_completed works, but I think the problem comes from last_slide_completed being a slide that got removed from the module.

ragesoss:
> Also, I like that you made the module_started? method, but it seems like that should be private.

Done.

thenickcox:
Oh, okay. Yeah, I guess we should handle that case. Thanks! :tada:

ragesoss:
That unbroke the timeline, but there were similar errors upon starting the module with the bad last_slide_completed data. That second commit will hopefully take care of those.

PR #485, opened 2015-11-20
Enable separate signup vs. login paths via omniauth-mediawiki fork

ragesoss:
@bmathews I think these might be the droids you're looking for.


bmathews:
Merged via 050d7a4.

PR #492, opened 2015-11-25
Upgrade react and router

thenickcox:
[Tests are passing](https://travis-ci.org/WikiEducationFoundation/WikiEduDashboard/builds/93222377) :tada: :tada: :tada: 

Lots of commits here. And lots of changes. Perhaps this is just a good way to acquaint you with what doesn't work in React & Router anymore. Couple big ones:

* `<RouteHandler />` (meaning rendering a component into another) [has become](https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/master...upgrade_react_and_router?expand=1#diff-db06c3242d889159716403a51173145aL26) `{@props.children}`
* No more named routes, unfortunately, e.g., `<Link to="timeline" props={{ foo: bar }}`. It has to be an actual route `<Link to={"/courses/#{course.id}/timeline" />`
* `transitionTo` [has become](https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/master...upgrade_react_and_router?expand=1#diff-967f9bc48a90e925d8d696031abb8d23L61) `history`.
* In Mocha tests, rendered DOM components [have to be composite components](https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/master...upgrade_react_and_router?expand=1#diff-755c790f76288e5de404b8a70f6e1534R8) for some stupid reason (i.e., a component nested inside another component)

Read over this and definitely ask about anything that's not clear.

## Edit

Apparently those links to line numbers don't work.


ragesoss:
@thenickcox what about the activity page specs that are still commented out?

thenickcox:
@ragesoss Those components literally just say "Hello from plagiarism!" So I removed them, but kept the tests in commented form to give us a good starting place for when we bring those pages back.

ragesoss:
@thenickcox I think you're misremembering. Those components are both working in production:

https://dashboard.wikiedu.org/recent-activity/recent-edits
https://dashboard.wikiedu.org/recent-activity/plagiarism

(And, we're seeing actual hits from the plagiarism feed for the first time!)

thenickcox:
@ragesoss *facepalm* I found [what I think I thought was the plagiarism handler](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/9ce452073568f3dc38da434fc88ca9fce30c9483/app/assets/javascripts/components/activity/mainspace_handler.cjsx) and deleted it. I think I was too deep in it.

Fixing now.

app/assets/javascripts/components/activity/activity_table.cjsx
```diff
@@ -39,9 +43,9 @@ ActivityTable = React.createClass(
 
     drawers = @state.activity.map (revision) ->
       courses = revision.courses.map (course) ->
-        <li><a href="/courses/#{course.slug}">{course.title}</a></li>
+        <li key={Math.random() * course.id}><a href="/courses/#{course.slug}">{course.title}</a></li>
 
-      <tr className='activity-table-drawer'>
+      <tr key={Math.random() * revision.key} className='activity-table-drawer'>
```
ragesoss: Won't this generate new keys every time this component rerenders? If so, will changing the keys have performance implications (such as forcing React to recreate each of those elements instead of reusing the existing one)?

Maybe there's some other way to ensure unique keys without introducing random numbers?

app/assets/javascripts/components/activity/did_you_know_handler.cjsx
```diff
@@ -39,6 +37,7 @@ DidYouKnowHandler = React.createClass(
         Show My Courses Only
       </label>
       <ActivityTable
+        loading={@state.loading}
```
ragesoss: I assume this adds that loading animation while it's waiting for data from the server. We should do the same thing for the plagiarism and recent edits handlers.

app/assets/javascripts/components/activity/activity_table.cjsx
thenickcox: I hadn't considered that, but I think it's safe to say `Math.random()` isn't an expensive operation.

app/assets/javascripts/components/activity/did_you_know_handler.cjsx
thenickcox: Good call.

app/assets/javascripts/components/activity/activity_table.cjsx
ragesoss: Yeah, Math.random() is going to be cheap. But changing the key might be expensive. What about... say... concatenating the revision and course ids as a string,  and adding '-drawer' onto the end for the drawer? Multiplying by a random number just seems... not right.

thenickcox: I agree about the "not right"-ness of it. But what do you mean when you say "But changing the key might be expensive"?

ragesoss: Keyed elements in React can be reused when the page rerenders; if the keys change every time the page rerenders, then it will keep creating new elements instead of reusing existing ones.

It's probably not really a performance issue, but if any of the components with randomized keys have their own state (for example, an open/closed state, or some other state related to behavior we haven't built yet) then that state would be lost on every render, because the state is tied to the key.

https://coderwall.com/p/jdybeq/the-importance-of-component-keys-in-react-js

thenickcox: Fixed via 841da57. With a higher article limit, I saw revisions coming through where the id and course id were duplicates, though that doesn't seem possible. We'll have to keep an eye on that.

thenickcox: Oops. `"#{revision.key}-#{course.id}"` was not unique.

thenickcox: Fixed via 7dad1a2

thenickcox:
@ragesoss With those last few commits, this should be good to go.

ragesoss:
@thenickcox looks like the latest changes broke the npm tests.

thenickcox:
Welp. Those tests pass separately, but not together. I'm out for the holiday. I'll look at them on Monday.

thenickcox:
Re-fixed Mocha tests. This is greeeeeeen.

PR #494, opened 2015-11-27
[WIP] Adds home page and begins aesthetic consistency/transition

bmathews:
I'm opening this PR a little early to have/track some discussion.

I've finished the majority of the home page portion of onboarding. I've made all of these changes in a way that would let us ship/test these changes before the rest of onboarding is finished, if we'd like. 

#### Questions:
1. @ragesoss / @thenickcox: How should we manage the content for the home page's "Impact since 2010:" stats and the university logos?
1. @ragesoss: I'm assuming the content of the home page needs further external review/refinement/localization. How/when should that get done?

#### Changes: 

- Added home page
  - Authenticated users are routed to /courses
  - Unauthenticated users are routed to the home page
- Improved code re-use for application html/nav
  - Split up reusable parts of the erb/haml files (header/footer)
  - Moved nav into a reusable partial
  - Use that nav partial in the training page (with `fluid` param)
- Transitioned the rest of the app to the new purple color for consistency
- Changed notification colors to new green/red colors
- Changed tooltip colors to primary purple
- Switched font from Roboto to Open Sans
  - Added variables with proper fallbacks to improve consistency/reuse
- Improved alignment inside of notifications

This starts _some_ improvements to the overall maintainability/modularity of the CSS, but once these changes get merged in I'd like to take a broader pass at that. 

Here're screenshots of some of the changes:

## Home page

nav

image
image
image
image
image

## Dashboard (color and font updates)

image
image

## Training nav
image
image
image


ragesoss:
@bmathews This is looking great. One thing we definitely want to add is some copy — not just accessible via a '?' icon — that explains what the dashboard does with the OAuth permissions it requires. Because signup / login don't correspond neatly to 'already has signed in before vs new to the dashboard', I think we probably need to do this on the homepage right before the signup / login segment.

Looks like there's still a bit of the spec that needs to be updated with new copy for the login / logout links.

I'll seek feedback on the design early next week.

ragesoss:
@bmathews The Wikipedia 'W' icon isn't rendering for me locally.

ragesoss:
@bmathews one little bug with the header: descender — the `g` in ask.wikiedu.org and the `p` in Help! — are being cut off. In your screenshot, only the `p` is affected, but for me locally, the `g` is also truncated.

Also, at mobile width, the layout of the header will break for a logged in user, because of 'Explore + My Dashboard + username + Log out'.

thenickcox:
@ragesoss I thought we were "supporting mobile," not, you know, *supporting mobile*?

ragesoss:
@thenickcox I was told that mobile would *just work*.

bmathews:
@ragesoss Polished up that text rendering and hid the logged in username on mobile. @rileywho has some newer designs for the right side of the navigation, so this should tide us over until that's implemented.

PR #513, opened 2015-12-04
Remove arrow key helper from first slide

ragesoss:
That's all there is to it, right?


ragesoss:
cc @bmathews quick sanity check?

bmathews:
:shipit:

PR #531, opened 2015-12-14
Only show strictly current courses to admins.

ragesoss:
Admins have so many courses, and the distance between terms is so short, that we want to immediately remove courses from the list when they end.

@bmathews Does this look okay to you? It's the case that we now only use the _user_courses partial for admins, right?


bmathews:
Yeah, looks good! 

:shipit:

PR #532, opened 2015-12-15
Adds a routine to post automatic greetings to students during batch update

ragesoss:
@thenickcox @bmathews 

If one of you have a chance, I'd love to get a review of this. It adds a feature for posting automatic welcome messages, and does some refactoring of the request stub methods along the way.

### DIFF
--- a/lib/student_greeter.rb
+++ b/lib/student_greeter.rb
@@ -1,0 +1,46 @@
+class StudentGreeter
+  def new
+    @greeters = User.find_by(wiki_id: 'Ragesock') # User.where(greeter: true)
+  end
+
+  private
+
+  def greet_all_ungreeted_students
+    courses = Course.strictly_current
+    courses.each do |course|
+      nonstudents_in_course = course.nonstudents
+      possible_greeters = @greeters & nonstudents_in_course
+      next if possible_greeters.empty?
+      greeter = possible_greeters[0]
+      course.students_without_nonstudents.each do |student|
+        greet_if_ungreeted(student, greeter)
+      end
+    end
+  end
+
+  def self.greet_if_ungreeted(student, greeter)
+    return if student.greeted
+    if talk_page_blank? student
+      greet student
+      return
+    end
+    # TODO: check whether any of the greeters have edited the talk page. return if so.
+    greet student
+  end
+
+  def talk_page_blank?(student)
+    Wiki.get_page_content(student.talk_page).nil?
+  end
+
+  def welcome_message(greeter)
+    # TODO get first name from real name, add to subst template
+  end
+
+  def self.greet(student, greeter)
+    response_data = WikiEdits.notify_users(greeter, [student], welcome_message)
+
+    return if response_data['edit'].nil
+    return unless response_data['edit']['result'] == 'Success'
+    student.update_attributes(greeted: true) if result
+  end
+end
--- a/lib/student_greeter.rb
+++ b/lib/student_greeter.rb
@@ -8,9 +8,10 @@ class StudentGreeter
   end
 
   def greet_all_ungreeted_students
-    courses = Course.strictly_current
-    courses.each do |course|
+    Course.strictly_current.each do |course|
       nonstudents_in_course = course.nonstudents
+      # Only users who both have the greeter flag and are enrolled as nonstudents
+      # are possible greeters for a class.
       possible_greeters = @greeters & nonstudents_in_course
       next if possible_greeters.empty?
       greeter = possible_greeters[0]
@@ -46,9 +47,10 @@ class StudentGreeter
 
   def ids_of_contributors_to_page(page_title)
     contributors_response = Wiki.query contributors_query(page_title)
-    # TODO: exception handling for unexpected response data
-    # currently, that will just cause a NoMethodError, which is okay but not
-    # optimal, because it will likely break a rake task.
+    # TODO: Add exception handling for unexpected response data.
+    # Currently, that will just cause a NoMethodError, which is okay but not
+    # optimal, because it will likely break a rake task. But it's at the end
+    # of the rake batch anyway, so it's not a huge deal.
     contributors = contributors_response.data['pages'].values[0]['contributors']
     contributor_ids = contributors.map { |user| user['userid'] }
     contributor_ids
--- a/lib/student_greeter.rb
+++ b/lib/student_greeter.rb
@@ -1,5 +1,9 @@
 class StudentGreeter
-  def new
+  def self.greet_all_ungreeted_students
+    new.greet_all_ungreeted_students
+  end
+
+  def initialize
     @greeters = User.find_by(wiki_id: 'Ragesock') # User.where(greeter: true)
   end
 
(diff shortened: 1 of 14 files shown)


lib/student_greeter.rb
```diff
@@ -0,0 +1,84 @@
+class StudentGreeter
+  def self.greet_all_ungreeted_students
+    new.greet_all_ungreeted_students
+  end
+
+  def initialize
+    @greeters = User.where(greeter: true)
```
thenickcox: I've read through this and come back to here. What it is a `greeter` user? An admin?

thenickcox:
I just don't have enough context overall. What's the purpose of this?

ragesoss:
The purpose is automate a simple, time-consuming task of Wiki Ed's two Content Experts, Ian and Adam. One of them is assigned to each course we support, and they usually post welcome messages on the Wikipedia talk pages of the students in each of their classes (manually).

What this does is use the dashboard to post those automatically. A 'greeter' is a manually assigned flag on a user that says that, if they are in the Wiki Ed Staff role for a course, then their account can be used to to greet students for that course (if they haven't already been greeted).

lib/student_greeter.rb
ragesoss: A greeter, at this point, is either of two users (Ian and Adam). They both are admins, but they are greeting in their capacity as the Wiki Ed Content expert assigned to a course. (There is usually also another staff member in the Wiki Ed Staff role for any given course, the Classroom Program Manager who is the instructor's point of contact, but that user is not a greeter.)

So, the greeter flag is intended to say 'this is one of the few users who will automatically greet students'. (In general, the system should not be making edits that don't directly result from an intentional action that the user just took. But staff accounts are the one exception.)

thenickcox:
Great. Then what you're talking about sounds like it should be on `courses_users`, no?

lib/student_greeter.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  def greet_all_ungreeted_students
+    courses = Course.strictly_current
```
thenickcox: This feels like it could go on one line, since `Course.strictly_current` is more descriptive.

`Course.strictly_current.each do |course|`

lib/student_greeter.rb
```diff
@@ -0,0 +1,84 @@
...
+    courses.each do |course|
+      nonstudents_in_course = course.nonstudents
+      possible_greeters = @greeters & nonstudents_in_course
```
thenickcox: If `non_students_in_course` are possible greeters, that sort of waters down the meaning of the `greeter` flag.

lib/student_greeter.rb
```diff
@@ -0,0 +1,84 @@
...
+      next if possible_greeters.empty?
+      greeter = possible_greeters[0]
+      course.students_without_nonstudents.each do |student|
```
thenickcox: Add the `ungreeted` flag here, so you don't need to do `greet_if_ungreeted` on the next line:

`course.students_without_nonstudents.ungreeted.collect { |s| greet(s) }`

lib/student_greeter.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  def ids_of_contributors_to_page(page_title)
+    contributors_response = Wiki.query contributors_query(page_title)
+    # TODO: exception handling for unexpected response data
+    # currently, that will just cause a NoMethodError, which is okay but not
+    # optimal, because it will likely break a rake task.
+    contributors = contributors_response.data['pages'].values[0]['contributors']
+    contributor_ids = contributors.map { |user| user['userid'] }
+    contributor_ids
```
thenickcox: Are you trying to stay under 100 chars for RuboCop? Couldn't you make this whole method two lines?

```
contrib = Wiki.query(contributors_query(page_title))
contrib.data['pages'].values.first['contributors'].map { |u| user['userid'] }
```

If your argument is that isn't clear, the method name is pretty descriptive, and it's a private method.

lib/student_greeter.rb
```diff
@@ -0,0 +1,84 @@
...
+    # Split on either whitespace or underscore, so that it works for usernames
+    # like Ian_(Wiki Ed) too.
+    name.split(/[\s_]/)[0]
+  end
+
+  def greet(student, greeter)
+    message = welcome_message(greeter)
+    response_data = WikiEdits.notify_user(greeter, student, message)
+    return if response_data['edit'].nil?
+    return unless response_data['edit']['result'] == 'Success'
```
thenickcox:
```
response = WikiEdits.notify_user(greeter, student, message)
return if response['edit'].nil? || !response_data['edit']['result'] == 'Success'
```

ragesoss:
If the greeter flag is only set manually, then it needs to be on User.

lib/student_greeter.rb
ragesoss: *shrug*

I woudn't have rewritten it from your form, but I find it easier to grok when there's not so much going on per line.

lib/student_greeter.rb
ragesoss: greet_if_ungreeted is still necessary, because a student may have been greeted (manually) even though the flag is not set. So greet_if_ungreeted does a couple of things:
* check for the greeted flag
* check for a blank talk page (which clearly indicates that the user has not been greeted)
* checks if any greeter has posted to the talk page before (and therefore, the user is actually greeted)

The check for the greeted flag can go here, but that will just remove the first line of greet_if_ungreeted.

lib/student_greeter.rb
ragesoss: Agreed.

lib/student_greeter.rb
ragesoss: possible greeters for a course are users who are both a) enrolled as nonstudents, and b) have the greeter flag.

lib/student_greeter.rb
ragesoss: You mean, change the 'greeted' flag to an 'ungreeted' flag? Or add an ungreeted scope that just looks for 'greeted: false'?

thenickcox:
Why is that?

lib/student_greeter.rb
thenickcox:
```ruby
class User < ActiveRecord::Base
  ...
  def self.greeted
    where(greeted: false)
  end
...
end
```

ragesoss:
Me: "If the greeter flag is only set manually, then it needs to be on User."
@thenickcox: "Why is that?"

Because otherwise, I'd need to make a manual command-line update to the database for every single course, as opposed to just once each for the two users who are greeters.

thenickcox:
But you mention this:

>They both are admins, but they are greeting in their capacity as the Wiki Ed Content expert assigned to a course.

If they're assigned as greeters to a course, doesn't it need to be course specific?

ragesoss:
They are not assigned as greeters. They are assigned as Wiki Ed Staff, but there are other users who are also assigned as Wiki Ed Staff who are not greeters.

thenickcox:
@ragesoss Fair enough!

PR #568, opened 2015-12-22
Switch from character count to word count in course and cohort stats

ragesoss:
@bmathews @thenickcox Can either of you take a peek at this?

### DIFF
--- a/app/helpers/course_helper.rb
+++ b/app/helpers/course_helper.rb
@@ -15,17 +15,4 @@ module CourseHelper
   def pretty_course_title(course)
     "#{course.school} - #{course.title} (#{course.term})"
   end
-
-  def word_count_from_character_sum(character_sum)
-    # Aaron Halfaker has done some analysis on the relationship between
-    # byte length and visible characters in English Wikipedi articles.
-    # According to his regression, the ratio is 1.15 bytes per visible character.
-    # Combining that with standard English ratio of 4.5 letters per word, and
-    # we get 5.175 characters / word.
-    # See discussion here: https://lists.wikimedia.org/pipermail/wiki-research-l/2013-August/002999.html
-    # See graph with regression line here: https://commons.wikimedia.org/wiki/File:Bytes.content_length.scatter.correlation.enwiki.png
-    # This is just a rough estimate, pending research on how bytes *added*
-    # relates to changes in readable word count for work by student editors.
-    (character_sum / 5.175).to_i
-  end
 end
--- a/app/helpers/course_helper.rb
+++ b/app/helpers/course_helper.rb
@@ -15,4 +15,17 @@ module CourseHelper
   def pretty_course_title(course)
     "#{course.school} - #{course.title} (#{course.term})"
   end
+
+  def word_count_from_character_sum(character_sum)
+    # Aaron Halfaker has done some analysis on the relationship between
+    # byte length and visible characters in English Wikipedi articles.
+    # According to his regression, the ratio is 1.15 bytes per visible character.
+    # Combining that with standard English ratio of 4.5 letters per word, and
+    # we get 5.175 characters / word.
+    # See discussion here: https://lists.wikimedia.org/pipermail/wiki-research-l/2013-August/002999.html
+    # See graph with regression line here: https://commons.wikimedia.org/wiki/File:Bytes.content_length.scatter.correlation.enwiki.png
+    # This is just a rough estimate, pending research on how bytes *added*
+    # relates to changes in readable word count for work by student editors.
+    character_sum / 5.175
+  end
 end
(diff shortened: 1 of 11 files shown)


thenickcox:
@ragesoss @bmathews Looking.

app/helpers/course_helper.rb
```diff
@@ -15,4 +15,17 @@ def current?(course)
...
+    # Aaron Halfaker has done some analysis on the relationship between
+    # byte length and visible characters in English Wikipedi articles.
+    # According to his regression, the ratio is 1.15 bytes per visible character.
+    # Combining that with standard English ratio of 4.5 letters per word, and
+    # we get 5.175 characters / word.
+    # See discussion here: https://lists.wikimedia.org/pipermail/wiki-research-l/2013-August/002999.html
+    # See graph with regression line here: https://commons.wikimedia.org/wiki/File:Bytes.content_length.scatter.correlation.enwiki.png
+    # This is just a rough estimate, pending research on how bytes *added*
+    # relates to changes in readable word count for work by student editors.
+    (character_sum / 5.175).to_i
```
thenickcox: Magic Number! 

ysbkhg1

app/views/courses/_header_stats.html.erb
```diff
@@ -12,8 +12,8 @@
     </div>
   </div>
   <div class="stat-display__stat">
-    <div class="stat-display__value"><%= number_to_human presenter.courses.sum(:character_sum) %></div>
-    <small><%= t("metrics.char_added") %></small>
+    <div class="stat-display__value"><%= number_to_human word_count_from_character_sum presenter.courses.sum(:character_sum) %></div>
```
thenickcox: Can we use parens here. It's hard to tell what the argument is.

app/views/courses/_row.html.erb
```diff
@@ -12,7 +12,7 @@
       </div>
 
       <div class="course-list__row__characters">
-        <p><span class="tablet-only-flex"><%= t("metrics.char") %>: </span><%= number_to_human course.character_sum %></p>
+        <p><span class="tablet-only-flex"><%= t("metrics.word_count") %>: </span><%= number_to_human word_count_from_character_sum course.character_sum %></p>
```
thenickcox: Same here.

app/views/courses/course.json.jbuilder
```diff
@@ -17,7 +17,7 @@ json.course do
   json.edit_count number_to_human @course.revisions.count
   json.student_count @course.user_count
   json.trained_count @course.students_without_nonstudents.trained.count
-  json.character_count number_to_human @course.character_sum
+  json.word_count number_to_human word_count_from_character_sum @course.character_sum
```
thenickcox: Here also. If you want to avoid a double set of parens, it could be `number_to_human word_count_from_character_sum(@course.character_sum)`

spec/features/course_page_spec.rb
```diff
@@ -134,7 +135,7 @@ def js_visit(path)
       find('#student-editors').click
       expect(page.find('#trained-count')).to have_content user_count / 2
       characters = revision_count * 2
-      expect(page.find('#characters-added')).to have_content characters
+      expect(page.find('#word-count')).to have_content word_count_from_character_sum characters
```
thenickcox: Here also

app/helpers/course_helper.rb
ragesoss: I thought that, since it only is used in this one place, it wouldn't count as a magic number. Nice gif, though.

spec/features/courses_spec.rb
```diff
@@ -1,4 +1,5 @@
 require 'rails_helper'
+include CourseHelper
```
thenickcox: The fact that this functionality is included in a helper is something I don't love. It's then globally available. Your argument for that might be that we need it in the jbuilder as well as in the views, but that could be handled by a presenter or service object. Exposing that information to the global namespace is, to me, not preferable to wrapping this functionality in a separate object.

app/helpers/course_helper.rb
thenickcox: For me, a magic number is any number that isn't 1 or 0.

ragesoss:
My reasoning for a helper was that its primarily a view-oriented routine, and it spans across courses and cohorts. But as a separate library it also makes sense... it just requires a little more webbing to make sure it's available where needed — both in a courses presenter and for an individual course. I was trying to avoid adding a method to Course. But, I think it's fine either way. The last commit moves it to lib/WordCount and adds the calls to it from Course instances and from the CoursesPresenter.

ragesoss:
@thenickcox cool with this version?

ragesoss:
yolo

PR #587, opened 2016-01-04
TrainingModulesUsersController should handle nonexistent slides

ragesoss:
This should fix the 404 slide spec introduced in 7ac48cd1c577ca3185995aa1862ebd7b54cfe9d3

@thenickcox you good with this?

### DIFF
--- a/spec/controllers/training_modules_users_controller_spec.rb
+++ b/spec/controllers/training_modules_users_controller_spec.rb
@@ -1,27 +1,30 @@
 require 'rails_helper'
 
 describe TrainingModulesUsersController do
-
   describe '#create_or_update' do
     let(:user)  { create(:user) }
     let(:t_mod) { TrainingModule.all.first }
     let(:slide) { t_mod.slides.first }
-    let(:post_params) {{
-      module_id: t_mod.slug,
-      slide_id: slide.slug,
-      user_id: user.id
-    }}
+    let(:post_params) do
+      { module_id: t_mod.slug,
+        slide_id: slide.slug,
+        user_id: user.id }
+    end
+
     before { allow(controller).to receive(:current_user).and_return(user) }
+
     it 'sets a slide complete' do
       post :create_or_update, post_params, format: :json
       expect(TrainingModulesUsers.last.last_slide_completed).to eq(slide.slug)
     end
+
     context 'first slide' do
       it 'does not set the module complete' do
         post :create_or_update, post_params, format: :json
         expect(TrainingModulesUsers.last.completed_at).to be_nil
       end
     end
+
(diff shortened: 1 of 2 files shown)


spec/controllers/training_modules_users_controller_spec.rb
```diff
@@ -30,6 +33,17 @@
...
+    context 'nonexistent slide' do
+      let(:nonexistent_slide_params) do
+        { module_id: t_mod.slug,
+          slide_id: 'not-a-real-slide',
+          user_id: user.id }
+      end
 
+      it 'renders a response' do
+        post :create_or_update, nonexistent_slide_params, format: :json
+        expect(response).to be
```
thenickcox: `.to be` ?

ragesoss: Just, to be. Shakespeare-like.

I wasn't sure what the convention is. Basically, I just want to test that a request for a nonexistent slide (when the user is logged in) returns a response and doesn't throw any errors.

thenickcox: (I assume you've seen [this](http://qz.com/283268/if-shakespeare-wrote-in-javascript-heres-what-it-would-look-like/).)

Anyway. Maybe check the response status?

thenickcox: `expect(response.status).to eq(302)` (I think, because it redirects)

ragesoss: In this case, it's a json request fired by the client to say 'I made it through this module all the way to this slide)', rather than the html request that redirects to a 404. So, at least as I patched it here, it's a 200 rather than a 302. Would it be better to redirect that json request to the 404 page too?

thenickcox: Not sure. It's complicated by the fact that the React Router doesn't handle our 404s. That's why there's a `window.location` call in the React component. So here, it's hard to say what the test should do.

ragesoss: I think the test makes sense for the current controller behavior. The question is whether the controller behavior is right.

I think it is, because this controller is basically the endpoint for an ajax request that says 'I made it to this slide', and the response it expects back is 'Okay, you made it to that slide'. Current implementation is to reply 'Okay, you made it to null-slide.' It's up to the client to figure out what to do from there. There's already another route by which the client is similultaneously being redirected to the 404 page, but this controller doesn't know or care.

It also makes for a pretty clean method: the render happens the same way for existing and nonexisting slides, and then it just returns before the business logic if the guard clause gets triggered.

PR #594, opened 2016-01-05
Set environment variables in config/environments/test.rb

### LINKED ISSUE #592: Flaky tests because of `application.yml` dependencies
```yaml
wiki_language: en
disable_wiki_output: 'false'
```

ragesoss:
This should ensure that the local application.yml settings don't interfere with specs. For example, the student_greeter_spec will fail if wiki output is disabled. This allows you to disable it for your development environment, but still have it set in the test environment to satisfy the specs.

Resolves #592


thenickcox:
LGTM :+1:

PR #600, opened 2016-01-08
On `capistrano deploy:rollback`, do not overwrite assets

### LINKED ISSUE #599: cap deploy:rollback syncs local assets
I ran `cap production deploy:rollback`, but it didn't have the intended effect... because the deploy process still ran the asset sync process, meaning that the previous javascript that I wanted to roll back to got overwritten by the buggy version on my local machine.

We need to disabled that step on deploy:rollback.

ragesoss:
The build and upload asset steps of capistrano's deployment script have also been applied to rollback, meaning that the compiled assets were not being rolled back but instead compiled and uploaded anew... and deployed against the older version of the ruby code. I couldn't find a better way to switch off asset steps for rollback, and this method seems to work.

Resolves #599


ragesoss:
@bmathews what do you think of this?

bmathews:
That makes sense to me!

PR #611, opened 2016-01-12
React router upgrade

thenickcox:
@ragesoss This is basically a placeholder. Take a look at some of the API changes for your reference. I fixed everything I could see on all the pages that I could think of (and find in the code) that used the router.

We won't merge this until the 2.0.0 final version comes out, but this will put us _ahead_ of the curve, rather than scrambling to play catch up in about 90 days when the next major version comes out.

### DIFF
--- a/app/assets/javascripts/components/common/calendar.cjsx
+++ b/app/assets/javascripts/components/common/calendar.cjsx
@@ -29,7 +29,6 @@ Calendar = React.createClass(
     else
       exceptions.push formatted
       utils = CourseDateUtils
-      console.log utils.wouldCreateBlackoutWeek(@props.course, day, exceptions)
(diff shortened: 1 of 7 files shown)


ragesoss:
Great! I'm going to merge this to staging, so I can poke at it across a few browsers.

ragesoss:
58 commits since RC5

o_O

ragesoss:
RC6 is out now, and advertises a new breaking change: https://github.com/rackt/react-router/releases

ragesoss:
@thenickcox 2.0.0 is out now. From what I can tell, we don't import from react-router anyway, so the breaking change in RC6 of `Removed default top-level <Router> export` shouldn't affect us. I'll update the react-router version and rebase this.

PR #615, opened 2016-01-14
Switch legacy courses to a Course subtype

ragesoss:
@thenickcox @adamwight 

I'd love to get a few eyes on this before I push it, especially to verify the migration.

@adamwight, this should resolve https://phabricator.wikimedia.org/T123618

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -78,9 +78,8 @@ class Course < ActiveRecord::Base
 
   # Legacy courses are ones that are imported from the EducationProgram
   # MediaWiki extension, not created within the dashboard via the wizard.
-  LEGACY_COURSE_MAX_ID = 9999
-  scope :legacy, -> { where('courses.id <= ?', LEGACY_COURSE_MAX_ID) }
-  scope :not_legacy, -> { where('courses.id > ?', LEGACY_COURSE_MAX_ID) }
+  scope :legacy, -> { where(type: 'LegacyCourse') }
+  scope :not_legacy, -> { where.not(type: 'LegacyCourse') }
 
   scope(:unsubmitted_listed, lambda do
     where(submitted: false).where(listed: true).merge(Course.not_legacy)
@@ -119,6 +118,7 @@ class Course < ActiveRecord::Base
   validates :passcode, presence: true, unless: :legacy?
 
   COURSE_TYPES = %w(
+    LegacyCourse
     ClassroomProgramCourse
     VisitingScholarship
     Editathon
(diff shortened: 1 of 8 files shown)


app/models/course.rb
```diff
@@ -142,10 +142,7 @@ def to_param
   end
 
   def legacy?
-    # If a course doesn't have an id yet, it's a new, unsaved course, and
-    # therefore not a legacy course. Legacy courses get their ids from the wiki.
-    return false if id.nil?
-    id <= LEGACY_COURSE_MAX_ID
+    type == 'LegacyCourse' ? true : false
```
thenickcox: Well, that cleans _that_ up nicely.

app/models/legacy_course.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  def wiki_title
+    prefix = 'Education_Program:'
+    escaped_slug = slug.tr(' ', '_')
+    "#{prefix}#{escaped_slug}"
+  end
+
+  def update(data={}, should_save=true)
+    # For legacy courses, the update may involve pulling data from MediaWiki
+    require "#{Rails.root}/lib/course_update_manager"
```
thenickcox: Can we put `require`s at the top of the file uniformly?

db/migrate/20160114203143_change_type_of_legacy_courses.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeTypeOfLegacyCourses < ActiveRecord::Migration
+  def up
+    Course.where('id < ?', 10000).update_all(type: 'LegacyCourse')
+  end
+
+  def down
+    Course.where('id < ?', 10000).update_all(type: 'ClassroomProgramCourse')
```
thenickcox: Magic numbers!

thenickcox:
Few minor notes, but I like how this simplifies things. With those fixed, it LGTM :+1:

db/migrate/20160114203143_change_type_of_legacy_courses.rb
ragesoss: But the constant that defines this is going away. Should I define that as a constant within the migration? Seems pointless, since it's a migration.

app/models/legacy_course.rb
ragesoss: This convention of inlining the requires was introduced by @ntdb, and I've found — especially lately — that it's quite nice when it comes to refactoring. In this specific case, it doesn't make a big difference... but have that require be inline for the older version of this method (on course.rb) made it easy to see that this was the only place that needed that... and so there was no need to double and triple check the top level requires.

PR #619, opened 2016-01-22
Change course interface strings depending on course type

ragesoss:
@bmathews @thenickcox @adamwight 

This is my first pass at having different types of courses use different interface strings appropriate to their type. For example, an edit-a-thon event (Editathon course type) should call the participants 'Editors' rather than 'Students'.

The strategy I've taken is to add a `#string_prefix` to each course type, and then use that to determine which string from the i18n file to use, via the function `CourseUtils.i18n` (which sets a default for before `@state.course` gets set on initial page load). This makes it pretty easy to further customize strings for different course types, but it doesn't seem very elegant.

If anyone has a better idea for implementing this, I'm eager to hear it.

### DIFF
--- a/app/assets/javascripts/components/common/list.cjsx
+++ b/app/assets/javascripts/components/common/list.cjsx
@@ -37,7 +37,8 @@ List = React.createClass(
     elements = @props.elements
     if elements.length == 0
       if @props.store.isLoaded()
-        text = I18n.t(@props.table_key + '.none')
+        text = @props.none_message
+        text ||= I18n.t(@props.table_key + '.none')
         text ||= 'This course has no ' + @props.table_key
       else
         text = I18n.t(@props.table_key + '.loading')
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -118,15 +118,14 @@ en:
     current: Current Courses
     duration: Course Duration
     expected_students: Expected Students
-    expected_editors: Expected Editors
     gradeables_none: This course has no gradeable assignments.
     instructor:
       one: Instructor
       other: Instructors
     students: Students
+    student_editors: Student Editors
     students_none: This course has no students.
     students_short: Students
-    editors: Editors
     timeline: Assignment Details
     weeks_none: This course does not have a timeline yet.
     new_week: Add New Week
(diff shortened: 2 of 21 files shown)


ragesoss:
@thenickcox Please leave a little bit of time this week to review this.

app/assets/javascripts/components/common/list.cjsx
```diff
@@ -37,7 +37,8 @@ List = React.createClass(
     elements = @props.elements
     if elements.length == 0
       if @props.store.isLoaded()
-        text = I18n.t(@props.table_key + '.none')
+        text = @props.none_message
+        text ||= I18n.t(@props.table_key + '.none')
```
thenickcox: Using `+` here creates a new string needlessly. How about `text ||= I18n.t("#{@props.table_key}.none")`?

config/locales/en.yml
```diff
@@ -65,17 +65,17 @@ en:
     loading: Loading articles...
     assigned: Assigned Articles
     assigned_to: Assigned To
-    character_doc: The sum of characters added to articles by the course's students during the course term
+    character_doc: The sum of characters added to articles by enrolled editors betweent the start and end dates
```
thenickcox: Typo here `betweent` -> `between`

thenickcox:
Nice use of duck typing. I had a few minor comments, but nothing major. Was there anything in particular about this that made you uncertain enough to want another set of eyes on it?

ragesoss:
@thenickcox Mainly just that it seems like it will be easy for this to get quite complex, and if there are better patterns that you (or the Wikimedia folks) know about for this type of thing, I don't want to go down this path and regret it later.

thenickcox:
It seems pretty simple to me for now. Are you anticipating adding anything to it that would complexify it?

ragesoss:
I'm not anticipating anything *specific*...

app/assets/javascripts/components/common/list.cjsx
ragesoss: I left that part as-is; the `+` format is used throughout the rest of the file.

PR #627, opened 2016-01-27
Remove user cache columns

ragesoss:
@thenickcox That came out pretty cleanly, with all the tests still green.


thenickcox:
:tada:

PR #637, opened 2016-02-03
Test should be insensitive to configuration

adamwight:


ragesoss:
@adamwight The strategy we've used to accomplish this is for other tests is to set all the test-sensitive values upon loading the test environment: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/config/environments/test.rb

I overlooked that one.

PR #639, opened 2016-02-04
Replace some hardcoded strings with t() calls

adamwight:
TODO: Unsure about "#{" vs "{" substitution style.

Bug: T125630

### DIFF
--- a/app/assets/javascripts/components/course.cjsx
+++ b/app/assets/javascripts/components/course.cjsx
@@ -36,7 +36,7 @@ Course = React.createClass(
     @state.current_user
   submit: (e) ->
     e.preventDefault()
-    return unless confirm "Upon submission, this course will be automatically posted to Wikipedia with your user account. After that, new edits to the timeline will be mirrored to Wikipedia.\n\nAre you sure you want to do this?"
+    return unless confirm I18n.t("courses.warn_mirrored")
     to_pass = $.extend(true, {}, @state.course)
     to_pass['submitted'] = true
     CourseActions.updateCourse to_pass, true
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/components/course.cjsx
```diff
@@ -36,7 +36,7 @@ Course = React.createClass(
     @state.current_user
   submit: (e) ->
     e.preventDefault()
-    return unless confirm "Upon submission, this course will be automatically posted to Wikipedia with your user account. After that, new edits to the timeline will be mirrored to Wikipedia.\n\nAre you sure you want to do this?"
+    return unless confirm #{I18n.t("courses.warn_mirrored")}
```
ragesoss: Here you only need the I18n.t() part. "#{}" is for string interpolation in coffeescript. The curly braces outside of a string, elsewhere in this file, are for adding coffeescript code within react cjsx fragments. This is just a function that returns a  string, so we need neither, I think.

adamwight:
@ragesoss I couldn't find a reference for that--could you drop a link to the docs?  Isn't this Reactified coffeescript which gets compiled to Javascript?  I'm not understanding when and how the Ruby substitution happens.

ragesoss:
@adamwight Ruby isn't involved with this substitution. During the `gulp build` process, we generate I18n javascript files. Here's the tooling: https://github.com/fnando/i18n-js

So the client gets the locale data in the form of a javascript file like en.js (`/public/assets/javascripts/i18n/en.js`) and the I18n-js module (required globally for the main javascript) composes the strings.

This is Reactified coffeescript, but the line in question doesn't have any special React stuff going on. It's just plain coffeescript that is supposed to look like `confirm STRING`. `I18n.t('message_key')` will return that string.

ragesoss:
@adamwight here are the relevant docs: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/i18n.md

app/assets/javascripts/components/course.cjsx
ragesoss: This is keeping the specs from passing, because it doesn't generate the confirmation dialog that is part of the course creation flow. Changing that line to `return unless confirm I18n.t("courses.warn_mirrored")` should fix it.

ragesoss:
This got done as PR #653

PR #655, opened 2016-02-06
Use course type to vary more UI; provide default course type setting

adamwight:
TODO:
* Provide the default course type to the course_creator component.

### DIFF
--- a/app/assets/javascripts/components/overview/details.cjsx
+++ b/app/assets/javascripts/components/overview/details.cjsx
@@ -48,7 +48,7 @@ Details = React.createClass(
         <fieldset>
           <TextInput
             onChange={@updateDetails}
-            value={@props.course.passcode}
+            value={display_code}
             value_key='passcode'
             editable={@props.editable}
             type='text'
--- a/app/assets/javascripts/components/overview/details.cjsx
+++ b/app/assets/javascripts/components/overview/details.cjsx
@@ -37,6 +37,11 @@ Details = React.createClass(
     online = <InlineUsers {...@props} users={@props.online} role={2} title='Online Volunteers' />
     campus = <InlineUsers {...@props} users={@props.campus} role={3} title='Campus Volunteers' />
     staff = <InlineUsers {...@props} users={@props.staff} role={4} title='Wiki Ed Staff' />
+    if @props.course.school
+      # FIXME: Convert lego to parameterized messages.
+      school = <p>{CourseUtils.i18n('school', @props.course.string_prefix)}: {@props.course.school}</p>
+    if @props.course.term
+      term = <p>{CourseUtils.i18n('term', @props.course.string_prefix)}: {@props.course.term}</p>
 
     if @props.course.passcode
       passcode = (
@@ -54,6 +59,58 @@ Details = React.createClass(
         </fieldset>
       )
 
+    if @props.course.expected_students
+      expected_students = (
+        <fieldset>
+          <TextInput
+            onChange={@updateDetails}
+            value={@props.course.expected_students}
+            value_key='expected_students'
+            editable={@props.editable}
+            type='number'
+            label={CourseUtils.i18n('expected_students', @props.course.string_prefix)}
+          />
+        </fieldset>
+      )
+
+    # TODO: course.timeline?
+    if @props.course.timeline_start || @props.course.timeline_end
+      timeline_start_props =
+        minDate: moment(@props.course.start)
+        maxDate: moment(@props.course.timeline_end).subtract(1, 'week')
+      timeline_end_props =
+        minDate: moment(@props.course.timeline_start).add(1, 'week')
+        maxDate: moment(@props.course.end)
+
+      timeline_start = (
+        <fieldset>
+          <TextInput
+            onChange={@updateDetails}
+            value={@props.course.timeline_start}
+            value_key='timeline_start'
+            editable={@props.editable}
+            type='date'
+            label={CourseUtils.i18n('assignment_start', @props.course.string_prefix)}
+            date_props={timeline_start_props}
+            required=true
+          />
+        </fieldset>
+      )
+      timeline_end = (
+        <fieldset>
+          <TextInput
+            onChange={@updateDetails}
+            value={@props.course.timeline_end}
+            value_key='timeline_end'
+            editable={@props.editable}
+            type='date'
+            label={CourseUtils.i18n('assignment_end', @props.course.string_prefix)}
+            date_props={timeline_end_props}
+            required=true
+          />
+        </fieldset>
+      )
+
     cohorts = if @props.cohorts.length > 0
       _.pluck(@props.cohorts, 'title').join(', ')
     else 'None'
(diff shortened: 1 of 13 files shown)


adamwight:
> BasicCourse?

Sure!

app/controllers/courses_controller.rb
```diff
@@ -15,7 +15,9 @@ class CoursesController < ApplicationController
   def create
     slug_from_params if should_set_slug?
     @course =
-      Course.create(course_params.merge('passcode' => Course.generate_passcode))
+      Course.create(course_params.merge(
+        passcode: Course.generate_passcode,
+        type: ENV['default_course_type']))
```
ragesoss: Currently, the default is handled by the database, but if `ENV['default_course_type']` is nil, this would create a course record that won't pass validation and won't get saved.

Let's do `type: ENV['default_course_type'] || 'ClassroomProgramCourse'`

That way, nothing will break for instances (including development instances) that leave their current application.yml as-is.

config/application.example.yml
```diff
@@ -44,6 +44,10 @@
 # EducationProgram extension (DEPRECATED).
   enable_legacy_courses: 'false'
 
+# Set the default course type for newly created courses.  Use either
+# "ClassroomProgramCourse" or "BasicCourse".
```
ragesoss: Why specify just those as valid options? Any default should work fine.

ragesoss:
This needs a test to exercise BasicCourse. Easiest way is to add a simple one here: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/spec/models/course_spec.rb#L531

And configure a factory here: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/spec/factories/courses.rb

If we add any unique behavior, we can break that out into a new spec, like https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/spec/models/visiting_scholarship_spec.rb

app/controllers/courses_controller.rb
adamwight: Okay, I see now that setting `type: nil` actually sets to null rather than the db default.  I don't like repeating the hardcoded default in multiple places, though.

What if we create a file for default configuration instead, and leave the setting commented in application.example.yml to make it clear that it's an optional value?

config/application.example.yml
adamwight: Thanks, I wasn't clear enough.  I was trying to suggest that there were two reasonable defaults, but you're right, no reason to be so inflexible!

ragesoss:
The admin role spec for adding a tag seems to be broken. (Many of the feature specs have flickering failures on travis, so one failure on travis doesn't necessarily mean that something broke. But this spec seems to be consistently failing for the PR, and the PR affects the part the part where it fails, which is the Details component.)

ragesoss:
@adamwight the 'add tag to course' test within the admin_role_spec is still failing on travis. Does it pass locally for you?

app/assets/javascripts/components/overview/details.cjsx
```diff
@@ -37,13 +37,23 @@ Details = React.createClass(
...
+      # FIXME: Convert lego to parameterized messages.
+      school = <p>{CourseUtils.i18n('school', @props.course.string_prefix)}: {@props.course.school}</p>
+    if @props.course.term
+      term = <p>{CourseUtils.i18n('term', @props.course.string_prefix)}: {@props.course.term}</p>
+
+    if @props.course.passcode
+      if @props.current_user.role > 0 || @props.current_user.admin
+        display_code = @props.course.passcode
+      else
+        display_code = '*****'
```
ragesoss: @adamwight The passcode is not sent to the client at all unless the user is an admin or instructor in the course. See https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/views/courses/course.json.jbuilder#L28 (magic number, whoops)

We could instead just check change the check to `if @props.course.passcode`.

app/assets/javascripts/components/overview/details.cjsx
```diff
@@ -54,6 +64,58 @@ Details = React.createClass(
...
+            value_key='expected_students'
+            editable={@props.editable}
+            type='number'
+            label={CourseUtils.i18n('expected_students', @props.course.string_prefix)}
+          />
+        </fieldset>
+      )
+
+    # TODO: course.timeline?
+    if @props.course.timeline_start || @props.course.timeline_end
```
ragesoss: This checks for either one, but moment code below expects both timeline_start and timeline_end to be defined. It's probably not functionally relevant, as timeline start and end come from the same json data and should always be present. But it would be clearer to either just check for one, or check for both with &&.

app/controllers/courses_controller.rb
```diff
@@ -14,8 +14,12 @@ class CoursesController < ApplicationController
 
   def create
     slug_from_params if should_set_slug?
-    @course =
-      Course.create(course_params.merge('passcode' => Course.generate_passcode))
+
```
ragesoss: Take this or leave it, but it reads nicer to me to build up `overrides` in parallel.
```ruby
overrides = {}
overrides[:passcode] = Course.generate_passcode
overrides[:type] = ENV['default_course_type'] if ENV['default_course_type']
```

app/assets/javascripts/components/overview/details.cjsx
adamwight: Thanks, that's nicer.  I did want to display something to alert that there is a passcode, so I'll shuffle the display logic a bit.

app/assets/javascripts/components/overview/details.cjsx
adamwight: Thanks!

adamwight:
> I'm cool with this. I note that this does affect rendering for all courses, not just generic ones.

Ah, good point!  I'll just put it back like I found it :)

ragesoss:
No, it's fine. I think it's a good idea; just wanted to document the fact, since this PR is otherwise not going to affect Wiki Ed's dashboard.

spec/factories/courses.rb
```diff
@@ -47,6 +47,15 @@
     passcode 'pizza'
   end
 
+  factory :basic_course, class: 'BasicCourse' do
```
ragesoss: This doesn't create a save-able course. Passcode, school and term are part of the validation.

https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/models/course.rb#L118
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/models/course.rb#L306

This is relevant for the follow-on work with #662

ragesoss: I can just add this on after merging, if @AndrewGreen gives the +1 to this PR. https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/4bff6f770b789f83f7ee31cf94d36e776b9d36f5

ragesoss:
@adamwight Let's save that passcode validation bit in the last commit for a different PR. In addition to the points above, it also breaks lots of tests in its current form.

adamwight:
Oops, thanks for pointing this out!

AndrewGreen:
+1!!! LGTM ;)

PR #662, opened 2016-02-08
Generic course, part 2

adamwight:


app/views/dashboard/index.html.haml
```diff
@@ -1,17 +1,18 @@
 - content_for :after_title, " - " + t("application.my_dashboard")
 
-#react_root
+#react_root{data: {default_course_type: ENV["default_course_type"],
```
ragesoss: Instead of calling the env variables from the template, let's set those values in DashboardPresenter and call them like other stuff through @pres.

app/views/dashboard/index.html.haml
```diff
@@ -1,17 +1,18 @@
...
+                   course_string_prefix: ENV["default_course_type"].constantize.new.string_prefix}}
```
ragesoss: This breaks if default_course_type is not set. Let's do this in the presenter.

I'm also thinking that it makes sense to make string_prefix a class method instead of an instance method, so that we can do things like this without creating an instance just to get a hard-coded string out of the course type.

app/assets/javascripts/components/course.cjsx
```diff
@@ -61,7 +61,7 @@ Course = React.createClass(
     user_role = if user_obj? then user_obj.role else -1
 
     if (user_role > 0 || @getCurrentUser().admin) && !@state.course.legacy && !@state.course.published
-      if CourseStore.isLoaded() && !(@state.course.submitted || @state.published)
+      if CourseStore.isLoaded() && !(@state.course.submitted || @state.published) && @state.course.type == 'ClassroomProgramCourse'
```
ragesoss: `:type` is not being sent to the client. To use this, we need to add type to /app/views/courses/course.json.jbuilder

ragesoss: The changes to course.cjsx — I'm not sure why — are causing problems only with Capybara webkit.

After adding course.type to the jbuilder, I was able to get the course_page_spec to pass (at least, didn't try others) by adding `Capybara.current_driver = :selenium`.

I'd love to get to the bottom of *why*, because that might help us figure out why some other tests only work with selenium and not headless webkit. (Webkit tests are significantly faster, so any of the slow feature specs we can move away from selenium is a win.)

ragesoss:
Here's one approach to making the string_prefix method accessible: 3c597e0988965c33cd1fbbd00f47097a87455ba1

app/assets/javascripts/utils/course_utils.coffee
```diff
@@ -17,8 +20,12 @@ class CourseUtils
 
   # This builds i18n interface strings that vary based on state/props.
   i18n: (message_key, prefix, default_prefix = 'courses') ->
-    key_prefix = prefix
-    key_prefix ||= default_prefix
-    I18n.t("#{key_prefix}.#{message_key}")
+    if prefix
+      message = I18n.t("#{prefix}.#{message_key}")
+
+    if !message || message.startsWith("[missing")
```
AndrewGreen: Is this right?
http://apidock.com/rails/ActionView/Helpers/TranslationHelper/translate

ragesoss: This uses a js library, i18n.js, but it's got pretty much the same
interface as the rails i18n framework.

AndrewGreen: @ragesoss thanks! Mmm so, yes looks like it is correct, my bad:
https://github.com/fnando/i18n-js
(A comment somewhere that this depends on leaving on the default behaviour for missing messages might be useful, in any case? I don't see any other error flag for missing messages provided by the library... Seems ironic that an i18n library makes you check for a literal English string, though... )

AndrewGreen: Aaarg, I take that back! You should use the library's fallback for missing scope, instead:
`I18n.t("some.missing.scope", {defaults: [{scope: "some.existing.scope"}]});`

AndrewGreen:
Cool!!! This is a huge step forward! My only concern is the lack of encapsulation: I'm not sure if we should be hardcoding strings of course types all about and switching functionality on that. Maybe instead some course-type-agnostic switches for the types of functionality that we wish to control, and have the course types somehow able to move those switches as necessary?

app/views/dashboard/index.html.haml
adamwight: I agree, but I'd prefer to make that change in a separate code quality PR?

app/assets/javascripts/utils/course_utils.coffee
adamwight: Awesome, thanks for finding that!

app/views/dashboard/index.html.haml
ragesoss: I'd rather not put off the presenter bit and adding a default fallback, because otherwise it will force every development environment and every live server to update application.yml.

I'm fine with putting off the string_prefix bit, if that's what you meant.

adamwight:
The most general solution would be something like, the course types would include a map of which fields and features they support, which fields are required vs optional, and so on.  That map could be modulated by the course itself, even.

We should probably postpone that design until we have more use cases for multiple course types, though--currently we're straddling a gray zone between simply forking the UI, and actually supporting course types.

adamwight:
Still need to fix the tests...

PR #664, opened 2016-02-09
Add footer with links to privacy policy and terms of service

ragesoss:
@rileywho FYI

@thenickcox @bmathews I can't figure out how to make the footer go to the bottom when there is not enough content to fill the screen. This is especially relevant for training slides, where we want to make sure the footer isn't in the middle of the screen. Also, just hacking my way through with CSS and haml, so let me know if this is the wrong way altogether to add a footer.


ragesoss:
I'm gonna go ahead and merge this.

PR #665, opened 2016-02-10
Flag to disable mandatory training and onboarding

adamwight:


app/views/shared/_nav.html.haml
```diff
@@ -8,7 +8,8 @@
         %li= link_to t('application.my_dashboard'), root_path, class: class_for_path(request, root_path)
       - if current_user && current_user.admin?
         %li= link_to t('application.recent_activity'), recent_activity_path, class: class_for_path(request, recent_activity_path)
-      %li= link_to t('application.training'), training_path, class: class_for_path(request, training_path)
+      - if !ENV['disable_training']
```
ragesoss: Because of the way ENV gets populated from application.yml, the ENV variables are all strings. For other such feature switches, the strategy we use for checking them to check for the exact string that signifies the relevant state. So in this case:
```ruby
unless ENV['disable_training'] == 'true'
```
would be more consistent with the rest of the codebase.

In this case, the setting specified in application.example.yml will mean that `ENV['disable_training']` is the string `'false'` rather than the boolean, which means it won't behave as intended.

app/views/courses/_users.json.jbuilder
```diff
@@ -4,7 +4,9 @@ json.users course.courses_users.eager_load(:user) do |cu|
   json.call(cu.user, :id, :wiki_id, :contribution_url, :sandbox_url)
   json.admin cu.user.permissions == User::Permissions::ADMIN
   json.recent_revisions RevisionStat.recent_revisions_for_user_and_course(cu.user, cu.course).count
-  json.course_training_progress ctp_manager.course_training_progress
+  if ENV['disable_training'] != 'true'
```
ragesoss: The `unless` form would be more consistent with the rest of the codebase. We try to follow the ruby style guide (https://github.com/bbatsov/ruby-style-guide#unless-for-negatives) as much as possible.

app/views/shared/_nav.html.haml
```diff
@@ -8,7 +8,8 @@
         %li= link_to t('application.my_dashboard'), root_path, class: class_for_path(request, root_path)
       - if current_user && current_user.admin?
         %li= link_to t('application.recent_activity'), recent_activity_path, class: class_for_path(request, recent_activity_path)
-      %li= link_to t('application.training'), training_path, class: class_for_path(request, training_path)
+      - if ENV['disable_training'] != 'true'
```
ragesoss: `unless` here too.

app/views/courses/_users.json.jbuilder
```diff
@@ -4,7 +4,9 @@ json.users course.courses_users.eager_load(:user) do |cu|
   json.call(cu.user, :id, :wiki_id, :contribution_url, :sandbox_url)
   json.admin cu.user.permissions == User::Permissions::ADMIN
   json.recent_revisions RevisionStat.recent_revisions_for_user_and_course(cu.user, cu.course).count
-  json.course_training_progress ctp_manager.course_training_progress
+  unless ENV['disable_training'] == 'true'
+    json.course_training_progress ctp_manager.course_training_progress
```
ragesoss: We should move all the training-related bits of this endpoint into the conditional.

```ruby
unless ENV['disable_training'] == 'true'
  ctp_manager = CourseTrainingProgressManager.new(cu.user, cu.course)
  json.course_training_progress ctp_manager.course_training_progress
  json.modules_overdue ctp_manager.first_overdue_module.present?
end

ragesoss:
This one will work find independently of the changes, and probably not much point in more CR. @adamwight make that above change and I'll merge this.

ragesoss:
I went ahead with 4b02b7550e1843ac095dc0308f78bce979006e77

PR #667, opened 2016-02-11
Multiwiki schema and migration

adamwight:
This adds a wiki property to Article, Assignment, and Revision, and makes space for a User.home_wiki, which we may or may not use.  Onwiki IDs are stored in native_id, plus Revision.page_id.  We'll slowly break the assumptions that primary id == onwiki ID, and refactor code to use the .wiki property rather than assume enwiki


adamwight:
This branch shouldn't be merged until native_id is updated correctly.

spec/factories/assignments.rb
```diff
@@ -22,6 +22,7 @@
     course_id 481
     article_id 124_884_99
     article_title 'Siderocalin'
+    wiki
```
ragesoss: should be `wiki_id`. If `wiki` works, that's Rails magic.

adamwight: Strange--if I try to use wiki_id instead, I get "Trait not registered: wiki_id"

ragesoss: Weird indeed. Maybe factory_girl does something different if you don't include a value? Maybe we should set an id for the wiki fixture, and set wiki_id to that in this factory.

db/migrate/20160210013622_create_wikis.rb
```diff
@@ -0,0 +1,33 @@
...
+    default_wiki = Wiki.find_or_create_by(
+      language: ENV['wiki_language'],
+      project: 'wikipedia'
+    )
+
+    add_column :articles, :wiki_id, :integer, index: true
+    add_column :assignments, :wiki_id, :integer, index: true
+    add_column :revisions, :wiki_id, :integer, index: true
+
+    reversible do |dir|
```
thenickcox: `Revision.update_all(:wiki_id, ENV['wiki_language'])`

app/models/article.rb
```diff
@@ -96,4 +101,11 @@ def update_cache
   def self.update_all_caches(articles=nil)
     Utils.run_on_all(Article, :update_cache, articles)
   end
+
+  def set_defaults_and_normalize
+    self.title &&= title.tr(' ', '_')
```
thenickcox: `return unless title.present?`

app/models/article.rb
```diff
@@ -96,4 +101,11 @@ def update_cache
...
+    # FIXME: transitional only
+    self.wiki_id ||= Wiki.default_wiki.id
```
thenickcox: `wiki_id ||= Wiki.default_wiki.id`

ragesoss: @thenickcox `self` is necessary for a setter.

adamwight:
It turns out, database column default value isn't kosher here.  The default_wiki.id is calculated, but gets saved in schema.rb as a constant.

app/models/article.rb
```diff
@@ -30,17 +32,18 @@ class Article < ActiveRecord::Base
...
   scope :namespace, -> ns { where(namespace: ns) }
 
-  # Always save titles with underscores instead of spaces, since that's the way
-  # they are in the MediaWiki database.
   validates :title, presence: true
-  before_validation do
-    self.title = title.tr(' ', '_')
-  end
+
+  before_validation :set_defaults_and_normalize
```
AndrewGreen: Why is before_validate the place to put transitional data into the new fields? Shouldn't we override getters instead?

AndrewGreen:
Is there a written doc about how the transition will work somewhere? Any thoughts or experiences on smoke testing the data transition?

app/models/article.rb
```diff
@@ -30,17 +32,18 @@ class Article < ActiveRecord::Base
   has_many :articles_courses, class_name: ArticlesCourses
   has_many :courses, -> { uniq }, through: :articles_courses
   has_many :assignments
+  belongs_to :wiki
+
+  # Allows us to directly #update from the Tools response
+  alias_attribute :page_id, :native_id
```
AndrewGreen: Are you talking about being able to use the `page_id` key on both sides of the assignment operator [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/773d72ce4169dc78d93fa9cb5c63799406cd9a07/lib/replica.rb#L71)? It seems that a detail of how the importer works shouldn't be in the model... I'd remove the alias, and maybe just call the real attribute/column something other than `native_id` (and do the same for other columns being added) to make the code that uses the model clearer.

AndrewGreen:
Cool!!!! Some general stuff:
- Maybe document somewhere the desired relationships between the new columns and the "native" mediawiki IDs?
- Instead of the name `native_id`, I'd recommend something like `id_in_mediawiki`. Sorry to make this suggestion so late in the game...
- How exactly is the data migration planned to work? Is it expected that each time an existing object is instantiated from a legacy row, then saved, the new columns will be filled? Wouldn't it be cleaner to have a single-use migration script? Though it seems improbable, in theory this migration strategy could be borked by changing the default wiki before all the legacy data has been fixed up. (Or am I completely misunderstanding?)
- The migration strategy should be documented somewheres. :)
- If the above migration strategy is to be pursued, I think the filling in of values for the new attributes/columns maybe shouldn't go in the `before_validation` callback? Even if it would be more verbose, I'd sooner use `after_initialize`, or [override getters](http://stackoverflow.com/questions/21835116/how-can-i-overwrite-a-getter-method-in-an-activerecord-model) and use `before_save`, or something like that.
- Another option might just be to never touch the legacy rows in the DB, and only override getters to provide the correct value when objects representing legacy rows are accessed.

app/models/article.rb
adamwight: The data needs to be in the database proper, cos we join on it, select values from the column, and so on.  As for `before_validate` vs `after_initialize` (in your comment below), I agree that after_initialize would make more sense, but I believe there were two issues.  One is that, in theory I'd like the normalized + default values to be available to any other hooks, so setting early in the lifecycle is better.  The second issue is probably just my own confusion, I seem to remember tests failing like crazy if I used the after_initialize hook.  Which in hindsight doesn't sound right, maybe I hadn't figured out the build (no hooks) vs create (runs hooks) distinction at that point.

I'll review whether I can move defaults to the after_initialize hook.

adamwight:
> Is there a written doc about how the transition will work somewhere? Any thoughts or experiences on smoke testing the data transition?

Thanks, I should write a doc!  The transformations are both reversible at first, therefore low-stress.  However, things will get tricky if we have to rollback after restarting the application, and I should document why that is.  Long story short, we'll be in danger of collapsing rows with identical native_ids from different wikis.

app/models/article.rb
adamwight: Great advice, I'll change the field name in the importer.  More descriptive field names will be an improvement, and will clarify code that consumes these records.

adamwight:
> - Instead of the name `native_id`, I'd recommend something like `id_in_mediawiki`. Sorry to make this suggestion so late in the game...

Yes!  How about `mw_page_id`, namespace first, and the rest of the column name actually matches the name in MediaWIki?

adamwight:
I'm convinced that documenting the migration is a good idea ;-)

My thought is to do a one-time migration.  The default value stuff is just a safety mechanism so that any code which was missed in the refactor doesn't produce records with null columns.  The next phase is to take the default value code out and add non-null constraints so that any incorrect code fails fast.  Maybe I should go ahead and do that now?  My thought was that the one-time migration can be merged separately from the refactor, if it includes default values added at runtime.

docs/multiwiki.md
```diff
@@ -0,0 +1,60 @@
+Background
+==========
+This application was originally written for English Wikipedia, and for
+convenience the primary IDs of many entities were kept in sync with MediaWiki
+database IDs.
+
+Multiwiki feature
+==================
+In February 2016, we've added support for articles from any of the [Wikimedia
```
AndrewGreen: Maybe use commit SHA instead of date?

docs/multiwiki.md
```diff
@@ -0,0 +1,60 @@
...
+project wikis](https://meta.wikimedia.org/wiki/Wikimedia_wikis#Core_projects),
+which involves database schema and code changes.  The `Wiki` table will hold
+one record for each wiki referenced from the application, and all onwiki
+entities will link to that row.  The entity primary IDs will be decoupled from
```
AndrewGreen: When you say "all onwiki entities will link to that row", maybe you mean something like, "all Dashboard entities that refer to an entity on a given wiki, or that must specify a wiki for any reason, will link to the corresponding row in the wiki table"? Rrrrg sorry for curmudgeonliness.

app/models/revision.rb
```diff
@@ -17,16 +17,25 @@
 #  wp10_previous  :float(24)
 #  ithenticate_id :integer
 #  report_url     :string(255)
+#  wiki_id        :integer
+#  native_id      :integer
+#  page_id        :integer
```
AndrewGreen: If I understand correctly, `page_id` (soon to be `mw_page_id`, I think?) will the the "new" multiwiki-aware version of `article_id`? If so, maybe a comment would be helpful to explain this? (I stared down the MW DB schema for a few, attempting to grok... ;p )

docs/multiwiki.md
```diff
@@ -0,0 +1,60 @@
...
+This is a pretty bad guess, and will be changed in the future.  Having a home
+wiki is helpful to minimize code changes during this refactor, but we don't
+know whether we even want such a property.
+
+Migration
+=========
+There will be a one-time schema migration, which is mostly reversible in case
+of rollback.  Since everything in the application database is on the default
+wiki, we can simply copy the primary IDs over to the MediaWiki ID columns, and
+set the wiki to our default record.
```
AndrewGreen: Thanks so much for writing this explanation!!! :D This sounds like the right approach. Am I correct in understanding that there will be a one-off script that we'll run to do this, though said script isn't yet in any GitHub PR?

docs/multiwiki.md
```diff
@@ -0,0 +1,60 @@
...
+
+The refactor is split into two branches, the first, `wiki_schema`, creates
+a schema without making logic changes.  The second, `use_native_ids`, rewrites
+the application to distinguish between local article ID and MediaWiki page ID.
+That branch is organized with safe patches towards the beginning, and riskier
+ones towards the end, so that a partial rollback is possible if required.
+Partially rolled back code will be protected by the default value logic, and
+will create valid records, but will only handle English Wikipedia articles, so
+should be corrected and redeployed as quickly as possible.
+
```
AndrewGreen: K I think I understand... Mainly I'd suggest that we limit possible rollback plans on the basis of what's actually risky and what's controllable given the current install base. It seems that a script copying existing primary keys and setting `wiki_id` is sure to succeed and can be easily verified by queries. This would happen before any of the code that actually uses the new columns is rolled out, right? So the default values set in `before_validation` in the models wouldn't be needed... Then when it's ready, multiwiki code can be rolled out on WikiEdu's production instance, but without providing any UI features that let you do anything anywhere other than a default wiki. WikiEdu I'm sure could guarantee they won't switch the default wiki during whatever testing period they need. That would guarantee no duplicate ids during that testing period. When they're satisfied with the stability of the new code, they could roll out UI for multiwiki features, if they wish (or maybe they'll want some switches to hide it, not sure...).

A similar plan could be enacted for the semi-production instance on labs.

For the real WMF P&E production instance, there's no worry about rollback, since it doesn't yet exist. :)

Does that make sense? Have I misunderstood something? BTW I'm not sure what you mean by, "Partially rolled back code will be protected by the default value logic, and will create valid records." Maybe the idea isn't to fill in the new columns in one go with a script? Apologies if I'm being thick-skulled!!

AndrewGreen:
Note that it's not guaranteed that all titles in the WikiEdu production database are normalized to use underscores instead of spaces. Would we want to check and if necessary normalize via a script? (From IRC: @ragesoss mentioned this seems OK.)

docs/multiwiki.md
adamwight: This'll just be the built-in "rake db:migrate" running db/migration/* files.

docs/multiwiki.md
adamwight: That's a great point--let's not advertise the potential for "partial rollback".  It's something I considered when ordering the patches, but if we actually need to do that, it's a world of pain.

adamwight:
> Note that it's not guaranteed that all titles in the WikiEdu production database are normalized to use underscores instead of spaces. Would we want to check and if necessary normalize via a script?

Ah!  Thanks, I'll add that migration.

db/migrate/20160210013622_create_wikis.rb
adamwight: So, I went ahead and used the models for all of my migrations.  However, I found this recently:
http://guides.rubyonrails.org/v3.2.8/migrations.html#using-models-in-your-migrations

The argument is that models will eventually diverge enough that they cannot be used to manipulate older schema.  Seems pretty compelling...

app/models/assignment.rb
```diff
@@ -46,4 +50,12 @@ def sibling_assignments
...
       .where.not(user: user_id)
   end
+
+  private
+
+  def set_defaults_and_normalize
+    self.article_title = article_title.tr(' ', '_') unless article_title.nil?
+    # FIXME: transitional only
+    self.wiki_id ||= Wiki.default_wiki.id
+  end
```
ragesoss: A more robust approach would be to use `Utils.format_article_title`, which also takes care of capitalizing the first letter.

adamwight: Great, thanks!

ragesoss: Also, note that we already do that via AssignmentManager before attempting to save: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/lib/assignments_manager.rb#L28

AndrewGreen:
@adamwight wrote:
> This'll just be the built-in "rake db:migrate" running db/migration/* files.

Aaah K I didn't get that. Cool! :)

ragesoss:
I moved this stuff over the the Wiki Ed repo. We can continue the discussion at #703. I'm going to merge it to staging and test it out soon.

PR #683, opened 2016-02-17
Anyone can create a course, when configured

adamwight:
Adds a configuration variable to give permissions by default.  This allows any
new user to create courses.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -18,6 +18,7 @@ class CoursesController < ApplicationController
     overrides = {}
     overrides[:passcode] = Course.generate_passcode
     overrides[:type] = ENV['default_course_type'] if ENV['default_course_type']
+    overrides[:cohorts] = [Cohort.default_cohort] if ENV['open_course_creation']
     @course = Course.create(course_params.merge(overrides))
     handle_timeline_dates
     CoursesUsers.create(user: current_user,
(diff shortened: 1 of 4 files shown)


app/models/user.rb
```diff
@@ -136,4 +138,8 @@ def to_json(options={})
     options[:except] ||= [:wiki_token, :wiki_secret, :remember_token]
     super(options)
   end
+
+  def set_defaults
+    self.permissions = ENV['default_permissions']
```
ragesoss: `||=`

ragesoss:
Setting the default permissions in the database on save doesn't seem like the right approach, especially since the goal is not 'make everyone an instructor' but rather 'let anyone create a course'.

DashboardPresenter showing or not showing the 'create course' button is the only thing that determines whether a user can create a course. (A non-instructor could create a course by navigating directly to `/course_creator`.)

I think rather than default permissions, this should be a configuration for whether anyone can create a course. Probably it would only need to be checked at `DashboardPresenter#can_create_course?`

ragesoss:
So, that part of it looks good, but given the name of the new config option and its description, this only takes it halfway. If you're about to work on having that option also affect the way 'submit' and 'approve' works, then we'll wait until that's included? That will keep this PR conceptually consistent.

adamwight:
Okay, that works for me.  It looks like we'll indeed make the workflow change, but need to design it a bit more.

ragesoss:
I didn't realize this was complete, or I would have merged this a while ago. I copied it over as fdc8b4ffe1642ce5ac4c39ba522cf76d54897bf0, since I had to amend it to make sure it handled the ENV variable as a string rather than a boolean.

ragesoss:
Okay, this functionality is on master now.

PR #695, opened 2016-02-23
Catch up with schema and annotations

adamwight:
Note that weeks.title was missing from db/schema.rb

### DIFF
--- a/app/models/basic_course.rb
+++ b/app/models/basic_course.rb
@@ -1,3 +1,40 @@
+# == Schema Information
+#
+# Table name: courses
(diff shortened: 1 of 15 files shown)


ragesoss:
I think we removed week.title when we redesigned the Timeline in late 2015.

ragesoss:
Actually, that doesn't appear to be right. We must have accidentally removed it from schema.rb at some point. I think we no longer use week.title, though.

adamwight:
Seems to be a flapping test.

PR #696, opened 2016-02-24
Multiwiki: Use native ids everywhere

adamwight:
This includes all the logic changes needed for multiwiki support.  Wiki associations are managed, loops are broken into groups of operations on one wiki at a time, and assumptions about primary ID == onwiki ID ("native id") are corrected.


db/migrate/20160210013622_create_wikis.rb
```diff
@@ -0,0 +1,25 @@
+class CreateWikis < ActiveRecord::Migration
+  def change
+    create_table :wikis do |t|
+      t.string :language, limit: 16
+      t.string :project, limit: 16
+    end
```
AndrewGreen: Any use in having one or more indices?

adamwight: yes, thanks!  I'll make a multi-column index.

db/migrate/20160210013622_create_wikis.rb
```diff
@@ -0,0 +1,26 @@
...
+    add_index :wikis, [:language, :project], unique: true
```
AndrewGreen: Doesn't have the same options as the equivalent line (231) in schema.rb.

AndrewGreen:
Maybe the URL logic in [5bd20](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/5bd20ab377eca60f6df11f42ca55a60e3b82ef0a) could be refactored as class methods of `Wiki`?

app/models/user.rb
```diff
@@ -3,6 +3,7 @@
 # Table name: users
 #
 #  id                  :integer          not null, primary key
+# FIXME: rename to username
```
AndrewGreen: Or maybe `mw_username`?

app/models/user.rb
```diff
@@ -12,6 +13,7 @@
 #  article_count       :integer          default(0)
 #  revision_count      :integer          default(0)
 #  trained             :boolean          default(FALSE)
+# TODO: Use global_name instead
```
AndrewGreen: `mw_global_username`? (Not sure how it works, if that's exactly right...)

AndrewGreen:
Commit message for [054f0ee](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/054f0ee72ebc7f02e741f7651f6a907e074a4abf) is potentially a bit confusing.....

db/migrate/20160210013622_create_wikis.rb
adamwight: The extra options are transparently added by the schema migration, it seems...

app/models/user.rb
adamwight: yes!

adamwight:
> Maybe the URL logic in 5bd20 could be refactored as class methods of Wiki?

I had the same thought, but decided to leave it alone for now, to avoid philosophical debate about model vs view helpers...

app/models/user.rb
ragesoss: I overlooked this before writing a patch for User.username. But I think `username` is preferable to to `mw_username`. Although we do make some queries against username where we assume that it's the same one on Wikipedia, most of the uses throughout the app are just invoking the username that should be shown for a user.

Unlike the revision and page ids, this *is* a global, cross-wiki username. If you can log in with OAuth, then you have a unified account and the same name across Wikimedia wikis (right?).

app/models/user.rb
ragesoss: `global_id` (which comes from the OAuth auth hash as `uid`, and is in fact the central_auth user id) is used to fetch the local user_id via replica (and local user_id is used other data from replica). Aside from that, we don't use it at all. Global username should always match regular username, as far as I know.

The main reason to use ids instead of usernames is to avoid encoding issues. Between ruby, php, mysql, and the wide range of allowed characters in mediawiki usernames, I've run into loads of bugs... which all went away when I switched to ids here: 40c2bbae28c4760d2435103d3ec93c2232e1d8b8

lib/analytics/course_statistics.rb
```diff
@@ -35,7 +35,7 @@ def report_statistics
   end
 
   def articles_edited
-    Article.where(namespace: 0, id: @page_ids)
+    Article.where(namespace: 0, id: @article_ids)
```
ragesoss: page_ids is correct here. In this file, `page_ids` is mw_page_ids, and `article_ids` is the ones that are namespace 0 (ie, not just pages but mainspace articles).

lib/analytics/course_statistics.rb
```diff
@@ -46,20 +46,19 @@ def articles_edited
...
       revision_ids << revisions.pluck(:id)
-      page_ids << revisions.pluck(:article_id)
+      article_ids << revisions.pluck(:article_id)
       upload_ids << course.uploads.pluck(:id)
     end
     @revision_ids = revision_ids.flatten.uniq
-    @page_ids = page_ids.flatten.uniq
+    @article_ids = article_ids.flatten.uniq
     @upload_ids = upload_ids.flatten.uniq
-    @article_ids = Article.where(namespace: 0, id: @page_ids).pluck(:id)
```
ragesoss: Remove this changes the (under-tested) behavior of this method; it will include ids from outside mainspace.

lib/replica.rb
```diff
@@ -79,179 +85,181 @@ def self.parse_revision(revision)
...
 
   # Given a list of articles, see which ones have not been deleted.
-  def self.get_existing_articles_by_id(articles, language=nil)
-    article_list = compile_article_id_string(articles)
-    existing_articles = api_get('articles.php', article_list, language)
+  # FIXME: inconsistent signature
+  def get_existing_articles_by_id(page_ids)
+    article_list = compile_article_id_string(page_ids)
+    existing_articles = api_get('articles.php', article_list)
+    # FIXME: What does this do?
```
ragesoss: The `unless` statement was left over from a refactor: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/0865d9a301913da0e0eb6c41c64aab1f31ad6cd2

lib/replica.rb
```diff
@@ -79,179 +85,181 @@ def self.parse_revision(revision)
...
-        user_list += '&' if i > 0
-        user_list += "user_ids[#{i}]='#{user.id}'"
-      end
-      user_list
-    end
+  # Compile a user list to send to the replica endpoint, which might look
+  # something like this:
+  # "usernames[]='Ragesoss'&usernames[]='Sage+%28Wiki+Ed%29'"
+  def compile_usernames_string(users)
+    # FIXME: don't quote
```
ragesoss: Why?

lib/replica.rb
```diff
@@ -79,179 +85,181 @@ def self.parse_revision(revision)
...
-        tag_list += "oauth_tags[#{i}]='OAuth CID: #{id}'"
-      end
-      tag_list
-    end
+  # Compile a user list to send to the replica endpoint, which might look
+  # something like this:
+  # "user_ids[0]='Ragesoss'&user_ids[1]='Sage (Wiki Ed)'"
+  # FIXME: deprecated
+  def compile_user_ids_string(users)
+    { user_ids: users.map(&:id) }.to_query
```
ragesoss: This is awesome. I didn't know about `.to_query`, but it makes this so much less painful!

lib/importers/assigned_article_importer.rb
```diff
@@ -6,14 +6,18 @@ class AssignedArticleImporter
...
-    Assignment.where(article_title: new_article_titles).each do |assignment|
-      article = Article.find_by(title: assignment.article_title)
-      next unless article.title == assignment.article_title # guard against case variants
-      assignment.article_id = article.id
-      assignment.save
+    assignments_missing_articles = Assignment.where(article_id: nil)
+    assignments_missing_articles.group_by(&:wiki).each do |wiki, assignments|
+      missing_titles = assignments.map(&:article_title).uniq
+      ArticleImporter.new(wiki).import_articles_by_title(missing_titles)
+      # FIXME: How are these "new"?
```
ragesoss: They are 'new' as in 'newly imported to the dashboard'. These are all the new Article records for Assignments that previously on had a title but weren't linked to an Article.

ragesoss:
I merged #712, which covers the same ground.

PR #703, opened 2016-03-01
Multiwiki schema

ragesoss:


ragesoss:
@adamwight @AndrewGreen I tried to deploy this to staging, and ran into errors during the migration.

`Mysql2::Error: Duplicate entry '692-25108540-Behavioural_confirmation-0' for key 'by_course_user_article_and_role': update assignments set article_title = replace(article_title, ' ', '_')`

ragesoss:
The way around it is probably to add a rescue to the migration and destroy the record if it's a duplicate.

ragesoss:
@adamwight @AndrewGreen Since the big `use_native_ids` is supposed to be a pure refactor, I think it will be easiest to deal with it one patch at a time. Once this schema stuff gets merged, we should be able to incrementally make all the changes in `use_native_ids` without breaking anything.

The one thing that might be a good idea *before* we merge this schema branch, though, would be to do the rename of User.wiki_id to User.username. That will be much easier to do with find-and-replace before we have other `wiki_id` columns referring to `Wiki.id`.

adamwight:
@ragesoss
> Since the big use_native_ids is supposed to be a pure refactor, I think it will be easiest to deal with it one patch at a time. Once this schema stuff gets merged, we should be able to incrementally make all the changes in use_native_ids without breaking anything.

This might be a nice way to deploy, but `use_native_ids` isn't currently severable.  There are plenty of things to fix in response to your code review (thank you!), so when I do that, I can spend more time making sure that dependencies come in the right order.

ragesoss:
@adamwight if you don't mind taking this `use_native_ids` branch as a first try and working in increments to reimplement it while keeping the tests passing, then I can still start merging parts of it.

I've now finished the User.username switch, I could start with making WikiApi, WikiEdits, and Commons use Wiki; those should be pretty independent, and won't change behavior as long as the default Wiki is in place.

PR #724, opened 2016-03-15
Fix recent activity nav

marythought:
This is now navigating to the correct route and persisting the correct URL (as far as I can tell!). I did have to change it from a Link to: to an href which I hope will not cause problems elsewhere.

### DIFF
--- a/app/assets/javascripts/components/activity/recent_activity_handler.cjsx
+++ b/app/assets/javascripts/components/activity/recent_activity_handler.cjsx
@@ -13,13 +13,13 @@ RecentActivityHandler = React.createClass(
     <div className='container recent-activity__container'>
       <nav className='container'>
         <div className="nav__item" id="dyk-link">
-          <p><Link to="recent-activity">Did You Know Eligible</Link></p>
+          <p><Link to="/recent-activity">Did You Know Eligible</Link></p>
         </div>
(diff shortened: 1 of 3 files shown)


ragesoss:
@marythought I think the `Link` elements will work if you just add a slash, like this: `<Link to="/recent-activity">`

marythought:
yup, that works! Sorry I missed that, I thought I tried every combination of dots and slashes!

ragesoss:
Great! Can you squash these into one commit?

marythought:
Did I do this correctly? In the course of rebasing I fetched from upstream which incorporated your commit into my branch :/

ragesoss:
Not quite. You've incorporated duplicate changes from my commit.

What command did you use to rebase? To just combine several commits (ie, "squash" them), I would normally do this: `git rebase --interactive HEAD~3` and then change "pick" to "squash".

And to rebase on origin/master would be an optional separate step after that.

marythought:
yup, I did exactly that, but I couldn't find the directions to force push to update the PR before I tried some other things (which brought in the latest commits). going to close out this PR and start fresh since I don't think there's a way to easily "unsquash"!

ragesoss:
That's fine. One way to use the same branch and PR would be to:
1. Reset hard to origin/master
2. Make the fix again in a new commit
3. git push -f

PR #726, opened 2016-03-15
add plagiarism and recent-revisions load bar

### LINKED ISSUE #355: recent-activity feed should have a 'loading revisions' placeholder
Currently, it says there are no DYK-eligible pages before it loads them. Instead, this should be a 'loading revisions' placeholder like we use on the /Students, /Articles, and /Activity pages of a course, and only show the 'none' message once the loading finishes (if there are none).

marythought:
This now mirrors the recent edits messaging and behavior
resolves #355

### DIFF
--- a/app/assets/javascripts/components/activity/plagiarism_handler.cjsx
+++ b/app/assets/javascripts/components/activity/plagiarism_handler.cjsx
@@ -5,6 +5,7 @@ ServerActions = require '../../actions/server_actions'
 
 getState = ->
   revisions: SuspectedPlagiarismStore.getRevisions()
+  loading: true
 
 PlagiarismHandler = React.createClass(
   displayName: 'PlagiarismHandler'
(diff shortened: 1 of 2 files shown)


ragesoss:
This only works on the assumption that there are revisions that will load. If there *are* no plagiarism revisions, then the old message will be correct. The idea here is to only show a loading message until the query for revision data resolves. I guess we don't actually handle that on recent-revisions, we just assume that there are *some* revisions. But we can't assume that there are plagiarism revisions.

See instead the behavior of the did-you-know component, where there is a loading animation until the data has loaded.

ragesoss:
That'll do nicely! One possible follow-up would be to move those messages for 'no revisions' (for all three of the activity components) into the i18n framework.

marythought:
Sounds like good practice for my Outreachy proposal -- I'll see what I can do! Thanks!

PR #729, opened 2016-03-16
plain to i18n text in recent activity section

marythought:
includes recent activity labels, plagiarism, recent edits, and DIY
added labels to en.yml and descriptions to qqq.yml
follow-up to PR #726

### DIFF
--- a/app/assets/javascripts/components/activity/recent_activity_handler.cjsx
+++ b/app/assets/javascripts/components/activity/recent_activity_handler.cjsx
@@ -10,16 +10,21 @@ RecentEditsHandler = require './recent_edits_handler'
 RecentActivityHandler = React.createClass(
   displayName: 'RecentActivityHandler'
   render: ->
+
+    dIYEligible = I18n.t('recent_activity.did_you_know_eligible')
+    plagiarism = I18n.t('recent_activity.possible_plagiarism')
+    recentEdits = I18n.t('recent_activity.recent_edits')
+
     <div className='container recent-activity__container'>
       <nav className='container'>
         <div className="nav__item" id="dyk-link">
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -305,6 +305,16 @@ en:
     user_char_added: Userspace chars added
     user_ns: User
 
+  recent_activity:
+    show_courses: Show My Courses Only
+    no_eligible: There are not currently any DYK-eligible articles.
```
ragesoss: This key should be more specific. Without reading the message, it's not clear that it's specifically about DYK.

app/assets/javascripts/components/activity/recent_activity_handler.cjsx
```diff
@@ -10,16 +10,21 @@ RecentEditsHandler = require './recent_edits_handler'
 RecentActivityHandler = React.createClass(
   displayName: 'RecentActivityHandler'
   render: ->
+
+    dIYEligible = I18n.t('recent_activity.did_you_know_eligible')
```
ragesoss: This variable name is misleading. You meant `dYKEligible`, I assume.

In my opinion, a better option more generally here would be to just put the I18n calls directly into the markup, instead of storing them in these variables first. With a long string specified in the file itself, having a variable for the message makes it easier to see what's going on with the html without having the content there to confuse things, but the i18n message keys do just as good or better a job than variable names at concisely indicating what's going on.

marythought: works for me! I didn't realize I could render the i18n calls by just sticking in a handlebar, but of course that makes sense if the handlebars can render a variable assigned to them.

ragesoss:
Thanks @marythought!

PR #741, opened 2016-03-23
Create a Features class to handle all features toggles

ragesoss:
This prevents us from dragging around the knowledge of how the ENV-based feature flags work as strings, and lays the groundwork for replacing the way we do feature toggles in the future.


ragesoss:
@adamwight thoughts on this?

adamwight:
Great idea!

Is that a transient test failure?

ragesoss:
Yep, transient test failure.

PR #843, opened 2016-05-04
Survey notification reminders

### LINKED ISSUE #790: Add reminder option to survey email notifications
In addition to an initial survey email, we'd like to be able to send reminder emails (eg, two weeks after the first email) to users who still have not completed a survey.

### LINKED ISSUE #791: Whenever config should only send emails Monday - Thursday
When we set up the cron job for automatic survey notification emails, we should set it up to send emails only on Mondays through Thursdays.

thenickcox:
This should be everything we need for this feature. If you're wondering where the rake spec is, it would be redundant; it just calls a `SurveyMailer` method that is itself tested. That is, in my opinion, how rake tasks should be. I did neglect, however, to delete the current spec for `send_notifications`.

If merged:

closes #790 
closes #791 

Update: stand by, running the whole test suite.

### DIFF
--- a/app/models/survey_notification.rb
+++ b/app/models/survey_notification.rb
@@ -10,10 +10,18 @@ class SurveyNotification < ActiveRecord::Base
   def send_email
     # In these environments only send emails to the users specified in ENV['survey_test_email']
     return if ['development', 'staging'].include?(Rails.env) && !ENV['survey_test_email'].split(',').include?(user.email)
-    return if email_sent
+    return if email_sent_at.present?
     return if user.email.nil?
-    SurveyMailer.notification(self).deliver_now unless email_sent
-    update_attribute('email_sent', true)
+    SurveyMailer.notification(self).deliver_now
+    update_attribute(:email_sent_at, Time.now)
+  end
+
+  def send_follow_up
+    return unless survey_assignment.follow_up_days_after_first_notification.present?
+    return if follow_up_sent_at.present? || user.email.nil?
+    return if Time.now < email_sent_at + survey_assignment.follow_up_days_after_first_notification.days
+    SurveyMailer.follow_up(self).deliver_now
+    update_attribute(:follow_up_sent_at, Time.now)
   end
 
   def survey_assignment
--- a/app/models/survey_notification.rb
+++ b/app/models/survey_notification.rb
@@ -18,7 +18,7 @@ class SurveyNotification < ActiveRecord::Base
 
   def send_follow_up
     return unless survey_assignment.follow_up_days_after_first_notification.present?
-    return if follow_up_sent_at.present? || user.email.nil?
+    return if follow_up_sent_at.present? || user.email.nil? || email_sent_at.nil?
     return if Time.now < email_sent_at + survey_assignment.follow_up_days_after_first_notification.days
     SurveyMailer.follow_up(self).deliver_now
     update_attribute(:follow_up_sent_at, Time.now)
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -28,3 +28,12 @@ end
 every 1.day, at: '4:30 am' do
   rake 'batch:update_daily'
 end
+
+every 1.day, at: '12:30 am' do
+  rake 'cache:warm:homepage'
+end
+
+every [:monday, :tuesday, :wednesday, :thursday], at: '10:15 am' do
+  rake 'surveys:send_notifications'
+  rake 'surveys:send_notification_follow_ups'
+end
--- a/lib/tasks/surveys.rake
+++ b/lib/tasks/surveys.rake
@@ -12,7 +12,7 @@ namespace :surveys do
   end
 
   desc 'Send follow-up survey notifications if configured on the parent notification'
-  task send_survey_notification_follow_ups: :environment do
+  task send_notification_follow_ups: :environment do
     SurveyNotification.active.each(&:send_follow_up)
   end
 end
--- a/lib/tasks/surveys.rake
+++ b/lib/tasks/surveys.rake
@@ -10,4 +10,9 @@ namespace :surveys do
   task send_notifications: :environment do
     SurveyNotification.active.each(&:send_email)
   end
+
+  desc 'Send follow-up survey notifications if configured on the parent notification'
+  task send_survey_notification_follow_ups: :environment do
+    SurveyNotification.active.each(&:send_follow_up)
+  end
 end
--- a/app/assets/stylesheets/_utils.styl
+++ b/app/assets/stylesheets/_utils.styl
@@ -140,6 +140,9 @@ show-over(breakpoint, display)
 .text-center
   text-align center
 
+.text-left
+  text-align left
+
 .no-bullets
   list-style none
   &.no-margin
--- a/app/assets/stylesheets/survey_modules/form.styl
+++ b/app/assets/stylesheets/survey_modules/form.styl
@@ -114,3 +114,6 @@ form:focus
 
 .input.string input[type=text]
   width 100%
+
+small.form__small-dark
+  color #787878
--- a/app/controllers/survey_assignments_controller.rb
+++ b/app/controllers/survey_assignments_controller.rb
@@ -109,6 +109,7 @@ class SurveyAssignmentsController < ApplicationController
     params.require(:survey_assignment)
           .permit(:survey_id, :send_before, :send_date_relative_to,
                   :send_date_days, :courses_user_role, :published,
+                  :follow_up_days_after_first_notification,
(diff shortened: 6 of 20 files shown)


ragesoss:
Since this morning, the (unchanged) survey admin spec is timing out on travis.

Also, I maybe have just introduced an edit conflict when I was clearing out lots of unecessary stuff in the survey rake file. On the plus side, the rake tasks are much smaller and fewer, as of yesterday. :-)

thenickcox:
I'll look at the survey admin spec. I'm also mulling over whether or not I buy that [rake tasks should be tested](https://robots.thoughtbot.com/test-rake-tasks-like-a-boss). Right now I'm leaning toward a sanity check test that would cover a case of the rake task being renamed or something because, again, the functionality itself is tested elsewhere. If you're on board with that, then I'll also collapse the current rake spec on the sending of the emails, since that functionality is also tested elsewhere.

ragesoss:
Re testing rake tasks, I'm with you on your current leanings. Basically, as of now, the most complex logic in any rake tasks is the stuff for `batch:pause` and `batch:resume`. I moved the hairy glob of crazy in the other batch update tasks into new libraries yesterday. With the rake test helper that Jonathan set up, it's easy enough to do sanity checks that I think it's worth it: `expect(SomeClass).to receive(:some_message)` and invoke the task, and that should be it in most cases.

thenickcox:
Yeah, that's exactly what I was thinking. Heading off to do that now.

ragesoss:
BTW, a couple annoyances I noticed with the rake task helper and its usage: you must use a specific task as the top-level `describe` subject, and you can only use the `subject.invoke` command for that one. So if you are testing multiple tasks from the same rake file, there's some disconnect between the `describe` listing and the actual things being tested.

Also, if you test multiple tasks in one file, the coverage report is based strictly on the last you test, so it looks like tasks from the earlier tests are not covered even if they are.

app/models/survey_notification.rb
```diff
@@ -10,10 +10,18 @@ class SurveyNotification < ActiveRecord::Base
...
-    SurveyMailer.notification(self).deliver_now unless email_sent
-    update_attribute('email_sent', true)
+    SurveyMailer.notification(self).deliver_now
+    update_attribute(:email_sent_at, Time.now)
+  end
+
+  def send_follow_up
+    return unless survey_assignment.follow_up_days_after_first_notification.present?
+    return if follow_up_sent_at.present? || user.email.nil?
+    return if Time.now < email_sent_at + survey_assignment.follow_up_days_after_first_notification.days
```
ragesoss: Will this raise an error if email_sent_at is not present? I don't see anything that would guard against getting this far even if the initial email was not sent.

lib/tasks/surveys.rake
```diff
@@ -94,8 +94,11 @@ namespace :surveys do
...
   task send_notifications: :environment do
-    SurveyNotification.active.each do |notification|
-      notification.send_email
-    end
+    SurveyNotification.active.collect(&:send_email)
+  end
+
+  desc 'Send follow-up survey notifications if configured on the parent notification'
+  task send_survey_notification_follow_ups: :environment do
+    SurveyNotification.active.collect(&:send_follow_up)
```
ragesoss: Why `collect` instead of `each`?

thenickcox: Yeah, it should be `each`.

On Wed, May 4, 2016 at 11:56 AM, Sage Ross <notifications@github.com> wrote:

> In lib/tasks/surveys.rake
> <https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/843#discussion_r62094891>
> :
>
> > @@ -94,8 +94,11 @@ namespace :surveys do
> >
> >    desc 'Find SurveyNotifications that haven\'t been sent and send them'
> >    task send_notifications: :environment do
> > -    SurveyNotification.active.each do |notification|
> > -      notification.send_email
> > -    end
> > +    SurveyNotification.active.collect(&:send_email)
> > +  end
> > +
> > +  desc 'Send follow-up survey notifications if configured on the parent notification'
> > +  task send_survey_notification_follow_ups: :environment do
> > +    SurveyNotification.active.collect(&:send_follow_up)
>
> Why collect instead of each?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/843/files/1aea670cfe086046833f0728c2e69d2ba85be33e..49d5a3b3970830afb541a3bd7613ff4ef929388c#r62094891>
>



-- 
*W* *I* *N* *T* *R*  |  Nick Cox |  Lead Engineer |  o. 206.783.4570  |
wintr.us

thenickcox: Because we don't care about the return value of the loop.

On Wed, May 4, 2016 at 12:02 PM, Nick - WINTR <nick@wintr.us> wrote:

> Yeah, it should be `each`.
>
> On Wed, May 4, 2016 at 11:56 AM, Sage Ross <notifications@github.com>
> wrote:
>
>> In lib/tasks/surveys.rake
>> <https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/843#discussion_r62094891>
>> :
>>
>> > @@ -94,8 +94,11 @@ namespace :surveys do
>> >
>> >    desc 'Find SurveyNotifications that haven\'t been sent and send them'
>> >    task send_notifications: :environment do
>> > -    SurveyNotification.active.each do |notification|
>> > -      notification.send_email
>> > -    end
>> > +    SurveyNotification.active.collect(&:send_email)
>> > +  end
>> > +
>> > +  desc 'Send follow-up survey notifications if configured on the parent notification'
>> > +  task send_survey_notification_follow_ups: :environment do
>> > +    SurveyNotification.active.collect(&:send_follow_up)
>>
>> Why collect instead of each?
>>
>> —
>> You are receiving this because you authored the thread.
>> Reply to this email directly or view it on GitHub
>> <https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/843/files/1aea670cfe086046833f0728c2e69d2ba85be33e..49d5a3b3970830afb541a3bd7613ff4ef929388c#r62094891>
>>
>
>
>
> --
> *W* *I* *N* *T* *R*  |  Nick Cox |  Lead Engineer |  o. 206.783.4570  |
> wintr.us
>



-- 
*W* *I* *N* *T* *R*  |  Nick Cox |  Lead Engineer |  o. 206.783.4570  |
wintr.us

thenickcox:
That must be what I'm running into _right now_!

app/models/survey_notification.rb
thenickcox: That needs a test!

thenickcox: Fixed via a122991d806df78eb998aaa01e5a782b590e89fa

thenickcox:
Just rebased and force pushed. I'll take a look at the Travis weirdness.

ragesoss:
@thenickcox I wrote in a workaround for the travis weirdness, which your rebase will have picked up. This should be good to go.

thenickcox:
:tada:

ragesoss:
@thenickcox I'm deploying this to staging now, and assuming not migration problems on the production-sized database, I'll merge this in a bit.

thenickcox:
The migrations should be super fast.

ragesoss:
I no longer trust my intuition about how long a migration will take.

config/schedule.rb
```diff
@@ -28,3 +28,12 @@
 every 1.day, at: '4:30 am' do
   rake 'batch:update_daily'
 end
+
+every 1.day, at: '12:30 am' do
+  rake 'cache:warm:homepage'
```
ragesoss: This snuck back in. This no longer does anything useful, and I deleted that rake task.

ragesoss:
There's still the case of separating creation of notifications from sending of email. We need an option to send no email for a survey assignment.

config/schedule.rb
thenickcox: Doesn't the homepage have a lot of stats still?

thenickcox:
Ah, right. I'm in meetings for the rest of the day, but I'll add that tomorrow.

config/schedule.rb
ragesoss: Not that much, now. Just top level stats for the default cohort, and only for logged-out people, and the load time is not bad.

thenickcox:
Oh, do we not need the no email option before merge?

ragesoss:
It's pretty independent of the rest, and we're not going to use that option on the one survey we currently have up and running, so I didn't see any reason to hold up the merge. (Actually, it's already deployed on production.)

PR #848, opened 2016-05-05
Remove mouseover behavior of block Edit buttons

### LINKED ISSUE #844: Timeline block edit button should always be visible to users who can edit
The hover-to-reveal edit buttons are not discoverable enough, and this is a frequent source of confusion for new instructors.

ragesoss:
Resolves #844
 
@thenickcox Could you review this? It does what I want, but the css parts are just trial-and-error for me.

### DIFF
--- a/app/assets/javascripts/components/timeline/block.cjsx
+++ b/app/assets/javascripts/components/timeline/block.cjsx
@@ -38,6 +38,10 @@ Block = React.createClass(
     className = 'block'
     className += " block-kind-#{@props.block.kind}"
 
+    blockTypeClassName = 'block__block-type'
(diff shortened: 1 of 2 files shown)


thenickcox:
If you don't want it to animate, I wouldn't handle this with CSS; I would make it just not display in React if there weren't permissions.

ragesoss:
But it also requires some CSS so that the block type is still visible. The behavior without this change is that on mouseover, the button slides in *over* the "In Class" or whatever, but I want that to still be visible. That's why I have the &.editable bit to shift the position if the Edit button is present.

thenickcox:
Then LGTM! 👍

PR #850, opened 2016-05-06
Consolidate webpack configs and improve gulp dev watch flow.

### LINKED ISSUE #839: Hot loading/No hot loading build process lacks documentation
Coming back to this project after a while, it was frustrating for JavaScript not to work as expected when running `gulp`, and to have to dig around to find out why. I still lack the necessary understanding of the history behind why there are two options (and what exactly they do and why), so I wouldn't be a good candidate to write this, but I'd love for the next person to come along not to have to hunt down how to make JavaScript Work as Expected :tm:.

bmathews:
This resolves #839.

Now, `gulp` will Just Work™ better.  The file watching for the non hot-loading flow wasn't working. Production external sourcemaps are back now, too.


ragesoss:
@thenickcox can you try this out and see if it works for you, and if so, merge it?

thenickcox:
@ragesoss Looking into this now.

PR #854, opened 2016-05-10
Survey follow up count

### LINKED ISSUE #847: Surveys should be configurable for multiple followup emails
We want to sent a series of follow-up emails, rather than just one.

I think we could do this by adding a followup count to the SurveyNotification, as well as to the SurveyAssignment, and then use follow_up_sent_at to indicate the most recent followup's send date. Then we just increment the count each time, and add a little more date logic.

ragesoss:
Resolves #847

@thenickcox Can you review this when you get a chance?

### DIFF
--- a/app/models/survey_notification.rb
+++ b/app/models/survey_notification.rb
@@ -18,10 +18,11 @@ class SurveyNotification < ActiveRecord::Base
 
   def send_follow_up
     return unless survey_assignment.follow_up_days_after_first_notification.present?
-    return if follow_up_sent_at.present? || user.email.nil? || email_sent_at.nil?
-    return if Time.now < email_sent_at + survey_assignment.follow_up_days_after_first_notification.days
+    return if user.email.nil?
+    return unless ready_for_follow_up?
     SurveyMailer.follow_up(self).deliver_now
-    update_attribute(:follow_up_sent_at, Time.now)
+    update_attributes(follow_up_sent_at: Time.now,
+                      follow_up_count: follow_up_count + 1)
   end
 
   def survey_assignment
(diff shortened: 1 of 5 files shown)


app/models/survey_notification.rb
```diff
@@ -39,4 +40,24 @@ def user
...
+
+  MAX_FOLLOW_UPS = 3
+  def ready_for_follow_up?
+    return false if email_sent_at.nil?
+    return false if Time.now < last_email_sent_at + time_between_emails
+    return false if follow_up_count >= MAX_FOLLOW_UPS
+    true
+  end
+
+  def time_between_emails
```
thenickcox: This method name originally made me think it was for the time between follow up emails. Maybe something like `time_between_initial_and_follow_up`?

app/models/survey_notification.rb
```diff
@@ -39,4 +40,24 @@ def user
...
+    survey_assignment.follow_up_days_after_first_notification.days
+  end
+
+  def last_email_sent_at
+    sent_at = follow_up_sent_at
+    sent_at ||= email_sent_at
+    sent_at
```
thenickcox: Couldn't this just be `follow_up_sent_at || email_sent_at`?

db/schema.rb
```diff
@@ -296,6 +296,7 @@
     t.datetime "updated_at",                                     null: false
     t.boolean  "completed",                      default: false
     t.datetime "follow_up_sent_at"
+    t.integer  "follow_up_count",      limit: 4, default: 0
   end
```
thenickcox: Seems like we should also change the `follow_up_sent_at` to `last_follow_up_sent_at` since it can change.

app/models/survey_notification.rb
ragesoss: It's both.

app/models/survey_notification.rb
ragesoss: Could be, but I found clearer to do it this way, to emphasize the order more explicitly.

app/models/survey_notification.rb
thenickcox: I find that really confusing. Maybe `time_between_intial_and_most_recent_follow_up`?

app/models/survey_notification.rb
thenickcox: It's pretty verbose. Why not `follow_up_sent_at.present? ? follow_up_sent_at : email_sent_at`?

ragesoss:
Thanks @thenickcox! I made some tweaks and changed the column name.

PR #867, opened 2016-05-17
Multiwiki assignments

ragesoss:
This is a significant reshuffling of how assignments get created and deleted. Previously, the AssignmentsController has `create` and `destroy` methods that were not used at all, while the actual updating of assignments happened for a whole course at once via `UsersController#save_assignments`. (This was partly to allow for the instructor/admin feature of saving a batch of assignments at once. However, that UI was confusing anyway.)

With this, all assignment creation and deletion goes through AssignmentsController. It also parses the language and project of a URL pasted in as assignment input, and creates an assignment for the correct wiki.


app/assets/javascripts/components/high_order/editable.cjsx
```diff
@@ -28,13 +28,14 @@ Editable = (Component, Stores, Save, GetState, Label) ->
       if permissions && @state.editable
         unless save_only
           className = 'controls'
-          cancel = (
-            <button onClick={@cancelChanges} className='button'>{I18n.t('editable.cancel')}</button>
-          )
+          unless SaveOnly
```
mattfordham: @ragesoss I'm not 100% in touch with what's happening here, but the nested `unless` blocks and the two different "Save Onlys" makes this area a little tough to decipher. Perhaps adding some comments at least might help us refactor in the future?

mattfordham:
@ragesoss This is great work, and a big effort; nice! I walked through everything and it seems cool. Tests pass as expected. I made one comment related to a particularly confusing spot, but not a reason to avoid merging.

app/assets/javascripts/components/high_order/editable.cjsx
ragesoss: Yeah, good point. This was the hackiest bit that I did. The 'Editable' component takes a provided set of controls, and mixes in another one to toggle between Edit mode and adds a Save and Cancel button when edit mode is activated.

The students_list relies on this to properly render the *other* controls for Enrollment and Notify Students, but I wanted to change the assignments part to be just a toggle that adds or removes the Assign buttons from each cell. The original save_only option I couldn't figure out how to utilize cleanly, so I added that other one.

Okay, yeah. I knew this was a bad thing when I did it, but I figured... the Editable component is a complex mess anyway, so why not pile on?

Sigh.

ragesoss: To clarify... the original save_only option seems to be incompatible with the layout of the other controls. Using that instead of adding the new one resulted in shifting the position of the other controls, and so I opted for adding new logic that I knew would not affect any other usages of Editable in other contexts.

ragesoss:
@mattfordham Thanks for taking a look. There's a bug I discovered on staging that I'm stuck on. If there are multiple assignments in the same assign_cell, now only one of them is shown. I think this may have to do with `AssignmentStore` and the things it listens to or triggers.

Assignments reach an assign cell via a `Student` component, and that component is supposed to get a filtered set of assignments from the AssignmentStore. But as far as I can tell, my changes have made it so that only one assignment ends up in the store for each user-role, when the page first loads.

If you add a new assignment, it gets added to the store and behaves as expected, but upon refreshing the page, it goes back to showing only one assignment — even though the database has all the assignments. If you delete one and refresh, then one of the others appears.

ragesoss:
Fixed with that last commit.

mattfordham:
@ragesoss 👍

PR #932, opened 2016-08-10
Allow programs to be only 1 day long

MusikAnimal:
Bug: T140793


ragesoss:
I cleaned up some extra loose ends and merged it.

MusikAnimal:
Thanks! :)

PR #951, opened 2016-09-07
[WIP] Adding start/end times to courses

MusikAnimal:
**NOT READY FOR MERGE**

Bug: T125546


app/assets/javascripts/components/common/date_picker.jsx
```diff
@@ -9,10 +9,11 @@ const DatePicker = React.createClass({
 
   propTypes: {
     id: React.PropTypes.string,
-    value: React.PropTypes.string,
+    value: React.PropTypes.object, // should be a moment object
```
ragesoss: I think we can do `React.PropTypes.instanceOf(moment)`

MusikAnimal: I didn't have luck with that, e.g. `moment().constructor.name === "n"` whatever the heck "n" is, not sure!

ragesoss: That's weird. It works in the console on http://momentjs.com/docs/ — that is, `moment() instanceof moment` returns true — but it doesn't work in the dashboard.

ragesoss: I think an alternative could be:

```
React.PropTypes.shape({
  _isAMomentObject: React.PropTypes.bool
})
```
Per http://andrewhfarmer.com/validate-nested-proptypes/

app/models/course.rb
```diff
@@ -304,4 +305,11 @@ def ensure_required_params
     self.timeline_start ||= start
     self.timeline_end ||= self.end
   end
+
+  def set_end_times
+    unless self.use_start_and_end_times
```
ragesoss: No need for self here.

app/models/course.rb
```diff
@@ -304,4 +305,11 @@ def ensure_required_params
...
+      self.end = self.end.end_of_day
+      self.timeline_end = self.timeline_end.end_of_day
```
ragesoss: No need for self on the right side of the assignment here. (Still needed for `self.end` because end is a special ruby word.)

app/assets/javascripts/components/common/date_picker.jsx
```diff
@@ -25,7 +26,8 @@ const DatePicker = React.createClass({
     onFocus: React.PropTypes.func,
     onClick: React.PropTypes.func,
     append: React.PropTypes.string,
-    date_props: React.PropTypes.object
+    date_props: React.PropTypes.object,
```
ragesoss: Along with datepicker unit tests, it would be great to use `React.PropTypes.shape` to validate the shape of date_props more precisely. Not strictly necessary as part of this PR, but it would be nice.

app/models/course.rb
ragesoss: Use a guard class here: `return if use_start_and_end_times`.

app/models/course.rb
```diff
@@ -304,4 +305,11 @@ def ensure_required_params
     self.timeline_start ||= start
     self.timeline_end ||= self.end
   end
+
+  def set_end_times
```
ragesoss: Add documentation of the relationship between this the `use_start_and_end_times` method on course types, and how we interpret it to mean overriding the default of starting at 00:00 and ending at 23:59.

db/migrate/20160911210500_change_courses_end_to_datetime.rb
```diff
@@ -0,0 +1,12 @@
+class ChangeCoursesEndToDatetime < ActiveRecord::Migration
+  def up
+    change_column :courses, :end, :datetime
+    Course.all.each do |course|
```
ragesoss: Let's do this data migration in a separate migration. That will reduce the risk of breaking the migrations if travis times out during the migration.

db/migrate/20160911210531_change_courses_timeline_end_to_datetime.rb
```diff
@@ -0,0 +1,13 @@
+class ChangeCoursesTimelineEndToDatetime < ActiveRecord::Migration
+  def up
+    change_column :courses, :timeline_end, :datetime
+    # FIXME: doesn't appear to work on the first run?
+    Course.all.each do |course|
```
ragesoss: Likewise, let's do this in a separate migration, which hopefully will also fix the bug.

app/models/course.rb
ragesoss: This method should also set the start times to 00:00. That will ensure consistency if a course changes type from one that uses start/end times to one that doesn't.

spec/controllers/revisions_controller_spec.rb
```diff
@@ -47,6 +55,13 @@
       end
     end
 
+    it 'does not return revisions that happened after the basic course ended' do
```
ragesoss: `does not return revisions from the final day of the basic course but after it ended`

spec/controllers/revisions_controller_spec.rb
```diff
@@ -3,9 +3,10 @@
 
 describe RevisionsController do
   describe '#index' do
```
ragesoss: Add a comment that we're using this spec with different course types to check the behavior of course start/end times and how these interact with the Course.revisions scope,

PR #973, opened 2016-09-23
basic version of Issue#930-Add user profile page that shows associate…

sejalkhatri:
issue#930 resolved

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -43,6 +43,20 @@ class UsersController < ApplicationController
     end
   end
 
+  ####################################################
+  # Page for showing course info for particular user #
+  ####################################################
+  def show
+    @user = User.find_by_username(params[:username])
+    if @user
+      @course = @user.courses
+    else
+      flash[:notice] = 'User not found'
+      redirect_to controller: 'dashboard', action: 'index'
+    end
+  end
+
+
   private
 
   #################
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -199,6 +199,7 @@
 
   get '/styleguide' => 'styleguide#index'
 
+  get 'users/:username' => 'users#show'
```
ragesoss: This should be near the other /users/ path at line 24.

app/controllers/users_controller.rb
```diff
@@ -43,6 +43,20 @@ def enroll
     end
   end
 
+  ####################################################
+  # Page for showing course info for particular user #
+  ####################################################
+  def show
+      @user = User.find_by_username(params[:username])
```
ragesoss: This should conform to the project's Rubocop style rules. Use two spaces for indentation.

app/controllers/users_controller.rb
```diff
@@ -43,6 +43,20 @@ def enroll
...
+      if @user
+        @course = @user.courses
+      else
+        flash[:notice] = "User not found"
```
ragesoss: Use single quote strings except where necessary for interpolation.

app/controllers/users_controller.rb
```diff
@@ -43,6 +43,20 @@ def enroll
...
+        redirect_to :controller => 'dashboard', :action => 'index'
```
ragesoss: Use the newer hash syntax: `controller: 'dashboard'`

ragesoss:
Thanks @sejalkhatri! This is a solid beginning!

PR #977, opened 2016-09-28
update user profile

sejalkhatri:


config/locales/en.yml
```diff
@@ -584,7 +584,13 @@ en:
...
     username_placeholder: Username
-
+    role:
+      visitor: Visitor 
+      student: Student
+      instructor: Instructor
+      campus_volunteer: Campus Volunteer
+      online_volunteer: Online Volunteer
+      wiki_ed_staff: Wiki Ed Staff
   wiki_edits:
```
ragesoss: Great

app/views/users/show.html.erb
```diff
@@ -36,6 +36,18 @@
...
+        <div class='col'>
+          <div class='course-details_title'>
+            Role
+          </div>
+          <div class='course-details_value'>
+            <% r = c.courses_users %>
+            <% r.each do |p| %>
+              <% @pass = p.role %>
+            <%end%>
+            <%= role(@pass) %>
```
ragesoss: This loops through each courses_user in the course `c`, reassigning `@pass` for each one, and then rendering the last one. That means courses_user may not be the same user that we want.

There are a few alternatives. Maybe the most straightforward one would be to set @courses_users instead of @course in the controller, and use that as the basis for the entire view. (You can get the course for a CoursesUsers record with courses_user.course.)

ragesoss: This also might be a good time to refactor this entire loop for the course into a partial and convert it to haml. (Optional for this pull request.)

app/helpers/users_helper.rb
```diff
@@ -5,4 +5,20 @@ def contribution_link(courses_user, text=nil, css_class=nil)
     options = { target: '_blank', class: css_class }
     link_to((text || courses_user.user.username), courses_user.contribution_url, options)
   end
+  def role(value)
+		case value
+  	when  -1
```
sejalkhatri: ok

app/helpers/users_helper.rb
```diff
@@ -5,4 +5,20 @@ def contribution_link(courses_user, text=nil, css_class=nil)
...
+  		return t('users.role.visitor')
+  	when 0
```
sejalkhatri: Ya that will be better ,changed and works .Thanks

app/helpers/users_helper.rb
```diff
@@ -5,4 +5,20 @@ def contribution_link(courses_user, text=nil, css_class=nil)
     options = { target: '_blank', class: css_class }
     link_to((text || courses_user.user.username), courses_user.contribution_url, options)
   end
+  def role(value)
```
ragesoss: I suggest renaming this something more specific, like course_role_name. `value` should also have a more specific name, such as `courses_users_role`, to make it very clear how the method is supposed to be used.

sejalkhatri: roger that!

app/views/users/show.html.erb
sejalkhatri: Thanks it's sorted now ,loop is removed . I ll convert the file to haml in next pull request and also add controller tests.

sejalkhatri:
Thanks Sir , learning something new everyday !( with respect to Closed #978
<https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/978>
via 13fc071
<https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/13fc071782b8bedde9c9432202ff5a008d8d03fe>
.)

PR #979, opened 2016-09-30
[WIP] Change vocabulary for dashboard

Niharika29:
For https://phabricator.wikimedia.org/T141327 this task. 

I've created new keys where the key was too generic, in case it might be used in other places. I've used existing keys otherwise. Let me know if I should change this!


app/assets/javascripts/components/overview/details.jsx
```diff
@@ -222,10 +222,6 @@ const Details = React.createClass({
         </div>
         <div className="module__data extra-line-height">
           {instructors}
-          {online}
-          {campus}
-          {staff}
-          {school}
```
Niharika29: Here, I could probably remove all the computation done earlier to display this information as well. Should I do that?

app/helpers/course_helper.rb
```diff
@@ -22,4 +22,8 @@ def date_highlight_class(course)
     return 'table-row--info' if course.start > 1.week.ago
     return ''
   end
+
+  def course_i18n(prefix, messageKey)
```
Niharika29: I did not actually end up using this.

config/locales/en.yml
```diff
@@ -152,7 +152,7 @@ en:
     assign_self: Assign myself an article
     confirm_add_available: Are you sure you want to add %{title}?
     confirm_addition: Are you sure you want to assign %{title} to %{username}?
-    confirm_deletion: Are you sure you want to delete this assignment?
+    confirm_deletion: Are you sure you want to delete this article?
```
Niharika29: I let this be because in most places "article" is used instead of "assignment".

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -222,10 +222,10 @@ const Details = React.createClass({
...
+          {Features.wikiEd ? online : null}
+          {Features.wikiEd ? campus : null}
+          {Features.wikiEd ? staff : null}
```
ragesoss: Let's keep the pattern already in place with `staff` where is defined (line 75), and put the conditionals there. That way, we only set up the components in the first place if they are needed.

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -222,10 +222,10 @@ const Details = React.createClass({
...
+          {Features.wikiEd ? school : null}
```
ragesoss: We should keep `school` (which will renders as Institution for P & E programs.) Otherwise, users will not be able to change that if needed.

config/locales/en.yml
```diff
@@ -130,7 +130,7 @@ en:
     title_example: Article title
     view: View Article
     view_all: View All
-    view_doc: The sum of all views to each article during the course term
+    view_doc: The sum of all views to each article during the program period
```
ragesoss: Let's call this "project" instead of "program period". That will be general enough to work for both courses and programs.

config/locales/en.yml
ragesoss: I think this is a little unclear: it risks confusion about deleting the Wikipedia article itself. I suggest 'article assignment'.

config/locales/en.yml
```diff
@@ -323,23 +325,23 @@ en:
...
       other: Volunteers
+    weeks_none: This course does not have a timeline yet.
     warn_mirrored: >
       Upon submission, this course will be automatically posted to Wikipedia with your user account.
       After that, new edits to the timeline will be mirrored to Wikipedia.
 
       Are you sure you want to do this?
     wiki_staff: Wiki Ed Staff
-    word_count_doc: An estimate of the number of words added to mainspace articles by the course's students during the course term
+    word_count_doc: An estimate of the number of words added to mainspace articles by the program's editors during the program term
```
ragesoss: The programs version should be in courses_generic rather than replacing the course version.

app/presenters/features.rb
```diff
@@ -8,6 +8,10 @@ def self.wiki_ed?
     false
   end
 
+  def self.default_course_type
+    ENV['default_course_type'] || 'ClassroomProgramCourse'
+  end
+
```
ragesoss: This is not used. (It could be used to replace the same method in DashboardPresenter; this is a better place for it.)

app/views/courses/_cohort_courses.html.erb
```diff
@@ -17,7 +17,8 @@
     </div>
     <div class="cohort-select">
       <select class="cohorts">
-        <option selected disabled><%= t("cohort.view_other") %></option>
+        <option selected disabled>
+        <%= Features.wiki_ed? ? t("courses.view_other") : t("courses_generic.view_other") %></option>
```
ragesoss: This pattern shows up over and over. This is what we should use the helper method for. We could change the signature to `course_i18n(message_key)` and handle all the conditionals there. That would make each of these cases be simply `course_i18n('view_other')` and so on.

Niharika29: Actually, this is the only erb file where this pattern shows up. The other instances are html.haml and jsx files. If the method is added in /app/helpers/course_helper.rb (where I put it), will it be accessible to JSX and HTML files too?

ragesoss: It will be available in all the rails view templates, whether erb or haml. It won't be available in the jsx.

Niharika29: Got it.

app/helpers/course_helper.rb
```diff
@@ -23,7 +23,11 @@ def date_highlight_class(course)
     return ''
   end
 
-  def course_i18n(prefix, messageKey)
-    return t(`${prefix}.${messageKey}`);
+  def course_i18n(messageKey)
+    if Features.wiki_ed?
+      return t(`courses.#{messageKey}`);
```
Niharika29: Is the t() function available here?

ragesoss: Ah, that's it! Outside of views, call it as `I18n.t()`

app/helpers/course_helper.rb
```diff
@@ -25,9 +25,9 @@ def date_highlight_class(course)
 
   def course_i18n(messageKey)
     if Features.wiki_ed?
-      return t(`courses.#{messageKey}`);
+      return I18n.t(`courses.#{messageKey}`)
     else
-      return t(`courses_generic.#{messageKey}`);
+      return I18n.t(`courses_generic.#{messageKey}`)
```
ragesoss: Use double quotes instead of backticks. That should fix the no-key error.

app/helpers/course_helper.rb
```diff
@@ -25,9 +25,9 @@ def date_highlight_class(course)
 
   def course_i18n(messageKey)
```
ragesoss: `message_key` rather than `messageKey` is the standard style for Ruby.

app/helpers/course_helper.rb
```diff
@@ -22,4 +22,12 @@ def date_highlight_class(course)
     return 'table-row--info' if course.start > 1.week.ago
     return ''
   end
+
+  def course_i18n(messageKey)
```
ragesoss: Standard Ruby style is "message_key" rather than "messageKey".

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -70,25 +70,27 @@ const Details = React.createClass({
...
-          type="text"
-          label={CourseUtils.i18n('school', this.props.course.string_prefix)}
-          required={true}
-        />
-      );
+      online = <InlineUsers {...this.props} users={this.props.online} role={2} title="Online Volunteers" />;
+      campus = <InlineUsers {...this.props} users={this.props.campus} role={3} title="Campus Volunteers" />;
+      if (this.props.course.school || this.props.current_user.admin) {
+        school = (
+            <TextInput
```
ragesoss: This is failing on travis-ci because of the indentation causing eslint failures. Indent with two spaces to fix it. We have the tests set to be strict on javascript linting.

ragesoss:
squashed and merged separately.

PR #987, opened 2016-10-06
[WIP] Time zone functionality

MusikAnimal:
**Work in progress, do not merge**

Bug: [T140306](https://phabricator.wikimedia.org/T140306)

### DIFF
--- a/app/assets/javascripts/components/common/time_zone.jsx
+++ b/app/assets/javascripts/components/common/time_zone.jsx
@@ -1,0 +1,92 @@
+import React from 'react';
+
+const DatePicker = React.createClass({
+  displayName: 'TimeZone',
+
+  propTypes: {
+    id: React.PropTypes.string,
+    value: React.PropTypes.string,
+    value_key: React.PropTypes.string,
+    spacer: React.PropTypes.string,
+    valueClass: React.PropTypes.string,
+    label: React.PropTypes.string,
+    editable: React.PropTypes.bool,
+    enabled: React.PropTypes.bool,
+    p_tag_classname: React.PropTypes.string,
+    onChange: React.PropTypes.func,
+    append: React.PropTypes.string
+  },
+
+  getDefaultProps() {
+    return {
+      label: I18n.t('courses.time_zone')
+    };
+  },
+
+  /**
+   * Update parent component with new timezone.
+   * @return {null}
+   */
+  onChangeHandler(e) {
+    this.props.onChange(this.props.value_key, e.target.value);
+  },
+
+  getDropdownOptions() {
+    return Object.keys(window.TimeZones).map((tz, index) => {
+      const value = window.TimeZones[tz];
+      return (
+        <option value={value} key={`timezone-dropdown-${index}`}>
+          {tz}
+        </option>
+      );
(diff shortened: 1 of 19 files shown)


app/assets/javascripts/components/common/time_zone.jsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+
+const DatePicker = React.createClass({
```
ragesoss: TimeZone, not DatePicker

db/schema.rb
```diff
@@ -11,304 +10,294 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160930190451) do
+ActiveRecord::Schema.define(version: 20161005185412) do
```
ragesoss: Be sure to sanitize the schema.rb changes to just make the minimum change, before we merge.

app/models/course.rb
```diff
@@ -39,6 +39,7 @@
 #  syllabus_file_size    :integer
 #  syllabus_updated_at   :datetime
 #  home_wiki_id          :integer
+#  timezone              :string           default("UTC")
```
ragesoss: let's be sure to validate this against inclusion in the set ActiveSupport knows about.

db/schema.rb
MusikAnimal: Hmm that I don't know how to do. I did find it a bit odd that after my migration this file changed so much!

ragesoss: schema.rb gets recreated every time you run `rake db:migrate`... based on the state of your own database. But in tests and production, the schema is what gets loaded to control behavior.

In this case, the only things that should change are this line defining the prefix of the latest required migration...

db/schema.rb
```diff
@@ -11,304 +10,294 @@
...
+    t.integer  "upload_count",                        default: 0
+    t.integer  "uploads_in_use_count",                default: 0
+    t.integer  "upload_usages_count",                 default: 0
+    t.string   "syllabus_file_name"
+    t.string   "syllabus_content_type"
+    t.integer  "syllabus_file_size"
     t.datetime "syllabus_updated_at"
-    t.integer  "home_wiki_id",          limit: 4
+    t.integer  "home_wiki_id"
+    t.string   "time_zone",                           default: "UTC"
```
ragesoss: And this line, the only column that has changed.

ragesoss: So you can sort it out by reverting to the original version, and then just manually updating those two lines.

app/views/shared/_head.html.haml
```diff
@@ -15,6 +15,8 @@
   I18n.defaultLocale = "#{I18n.default_locale}";
   I18n.locale = "#{I18n.locale}";
 
+  TimeZones = #{ActiveSupport::TimeZone.all.each_with_object({}) { |v,h| h[v.name] = v.formatted_offset }.to_json};
```
ragesoss: Maybe instead of doing this translation on each page load, we should set up the array in an initializer.

MusikAnimal:
Closing in favour of #993

PR #993, opened 2016-10-14
Allow user to enter and be shown times in their own time zone

MusikAnimal:
This shows the start and end times of courses in the user's local time zone, but still saves in UTC. The time zone abbreviation is also shown in the Details view, after the course is saved.

Bug: [T140306](https://phabricator.wikimedia.org/T140306)

### DIFF
--- a/spec/features/open_course_creation_spec.rb
+++ b/spec/features/open_course_creation_spec.rb
@@ -14,6 +14,8 @@ end
 describe 'open course creation', type: :feature, js: true do
   let(:user) { create(:user) }
   before do
+    @system_time_zone = Time.zone
+    Time.zone = 'Eastern Time (US & Canada)'
     ENV['default_course_type'] = 'BasicCourse'
     Capybara.current_driver = :poltergeist
     page.current_window.resize_to(1920, 1080)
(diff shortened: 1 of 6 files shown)


ragesoss:
The latest version is deployed on dashboard-testing.wikiedu.org.

It looks like this is working fine, with the exception of how it affects courses that don't `use_start_and_end_times`: the start dates are off from what gets saved in the database. For example: https://dashboard-testing.wikiedu.org/courses/test/Timezone_test

Here the start date renders for me as 10-30, but it is saved in the database as 10-31 00:00

MusikAnimal:
Ah, so just convert to UTC for courses without `use_start_and_end_times`. Will fix!

spec/features/open_course_creation_spec.rb
```diff
@@ -14,6 +14,8 @@ def fill_out_open_course_creator_form
 describe 'open course creation', type: :feature, js: true do
   let(:user) { create(:user) }
   before do
+    system_time_zone = Time.zone
```
ragesoss: I think this needs to be either an instance variable or it needs to be outside the 'before' block so it is in scope for the 'after' block.

MusikAnimal: whoops, duh! :)

PR #1007, opened 2016-10-25
Change all instances of cohort to campaign

MusikAnimal:


app/presenters/courses_presenter.rb
```diff
@@ -1,29 +1,29 @@
...
 
-  def cohort
-    return NullCohort.new if cohort_param == 'none'
-    @cohort ||= Cohort.find_by(slug: cohort_param)
-    raise NoCohortError if @cohort.nil? && cohort_param == ENV['default_cohort']
-    @cohort
+  def campaign
+    return NullCampaign.new if campaign_param == 'none'
+    @campaign ||= Campaign.find_by(slug: campaign_param)
+    raise NoCampaignError if @campaign.nil? && campaign_param == ENV['default_campaign']
```
ragesoss: We'll need to duplicate this variable with the new name in application.yml before deployment, and then clean up the old version afterwards.

db/schema.rb
```diff
@@ -188,17 +188,17 @@
 
   create_table "question_group_conditionals", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
     t.integer  "rapidfire_question_group_id"
-    t.integer  "cohort_id"
+    t.integer  "campaign_id"
     t.datetime "created_at",                  null: false
     t.datetime "updated_at",                  null: false
-    t.index ["cohort_id"], name: "index_question_group_conditionals_on_cohort_id", using: :btree
+    t.index ["campaign_id"], name: "index_question_group_conditionals_on_campaign_id", using: :btree
```
ragesoss: Is this regenerated from migrations, or did you find-and-replace the schema.rb file?

I have no idea how indexes interact with renaming and mgirations.

db/migrate/20161025193948_rename_cohorts_tables_to_campaigns.rb
```diff
@@ -0,0 +1,7 @@
+class RenameCohortsTablesToCampaigns < ActiveRecord::Migration[5.0]
+  def change
```
ragesoss: Do each of these renames as a separate migration, please.

db/schema.rb
ragesoss: And the two others on campaigns_survey_assignments, I guess.

I'm pretty sure if running the migrations didn't change this, then the old indexes will still be there and the new ones won't be.

MusikAnimal: I was wrong, Rails does this for you! I confirmed this with `SHOW INDEX` in the db console

ragesoss:
Seems to be working without a hitch, after deployment yesterday to staging! Nice work!

PR #1011, opened 2016-10-27
test added for home page redirecation in case of invalid user

sejalkhatri:


spec/controllers/users_controller_spec.rb
```diff
@@ -160,4 +160,15 @@
       expect(user.locale).to eq('es')
     end
   end
+
+  describe '#show' do
+    context 'when user not found' do
+      before {allow(controller).to receive(:current_user).and_return(nil) }
```
ragesoss: This is unnecessary. In this scope, since we didn't do anything like `allow(controller).to receive(:current_user).and_return(user)`, `current_user` will be `nil` anyway (which means visiting the page while logged out).

spec/controllers/users_controller_spec.rb
```diff
@@ -160,4 +160,15 @@
...
+      it 'redirects to the home page' do
+       @controller = DashboardController.new
```
ragesoss: This means you are using the DashboardController rather than the UsersController. This shouldn't be necessary here.

spec/controllers/users_controller_spec.rb
```diff
@@ -160,4 +160,15 @@
...
+       get :index
```
ragesoss: This does not test the behavior of `UsersController#show`. You need to do `get :show` with the required :username parameter, like this:

`get :show, params: { username: 'not a real user' }`

You can verify that it's testing the right thing by commenting out the redirect in UsersController#show, which should cause the test to fail.

db/schema.rb
```diff
@@ -201,8 +201,8 @@
 
   create_table "rapidfire_answer_groups", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
     t.integer  "question_group_id"
-    t.string   "user_type"
     t.integer  "user_id"
+    t.string   "user_type"
```
ragesoss: Remove this unrelated change and this will be ready to merge.

sejalkhatri: ok

PR #1022, opened 2016-11-09
Campaign overview page

MusikAnimal:
Also adding some missing translations

Bug: [T148804](https://phabricator.wikimedia.org/T148804)

### DIFF
--- a/app/assets/javascripts/utils/editable.js
+++ b/app/assets/javascripts/utils/editable.js
@@ -1,3 +1,22 @@
+/**
+ * Editable - a tiny jQuery plugin to turn blocks of text into inputs
+ *
+ * To use:
+ * - Build a normal Rails form, adding the class '.rails_editable' to the <form> tag.
+ * - Add the blocks of text you want to be editable, each having an associated input field as a sibling.
+ * - Each pair of a text block and input should be wrapped in a element with the class '.rails_editable-field'.
+ * - Blocks of text should have the class .rails_editable-content
+ * - Inputs should have the class .rails_editable-input (they are hidden via CSS, otherwise use the inline style 'display:none').
+ * - Add an edit button somewhere within the form with the class '.rails_editable-edit'.
+ *
+ * - When the Edit button is clicked:
+ *   - The script will prevent the form from being submitted, so you can use a <button> element.
+ *   - The inputs are supplied with the content of the associated text blocks, and the text blocks are hidden.
+ *   - Textareas will have the same size as the corresponding text block
+ *   - Cancel and Save buttons are appended to the parent. Construct your markup accordingly.
+ *   - Cancel will revert back to showing the text blocks and hide the inputs.
+ *   - Save will submit the form as a normal submit button would.
+ */
 $(() => {
   const readMode = $parent => {
     $parent.find('.rails_editable-cancel, .rails_editable-save').remove();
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/utils/editable.js
```diff
@@ -0,0 +1,30 @@
+$(() => {
+  const readMode = $parent => {
+    $parent.find('.editable-cancel, .editable-save').remove();
```
ragesoss: This naming seems a little risky; if we accidentally one of these `editable-save` or `editable-cancel` classes onto something in a React view, then we'd have jquery meddling with the view, right?

Maybe maybe it explicit that this is for rails forms? `rails_editable-cancel` and so on? (I'm not too experienced with jquery, so maybe this isn't something to worry about?)

app/assets/javascripts/utils/editable.js
```diff
@@ -0,0 +1,30 @@
+$(() => {
```
ragesoss: Should have documentation of the usage of this.

spec/features/campaign_overview_spec.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+describe 'campaign overview page', type: :feature, js: true do
+  let(:slug)  { 'spring_2016' }
```
ragesoss: two spaces

spec/features/campaign_overview_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  let(:user) { create(:user) }
+  let(:admin) { create(:admin) }
+  let(:campaign) {
```
ragesoss: use do/end for multiline block

spec/features/campaign_overview_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    create(:campaign,
+           id: 10001,
+           title: 'My awesome Spring 2016 campaign',
+           slug: slug,
+           description: 'This is the best campaign')
+  }
+
+  before do
+    Capybara.current_driver = :poltergeist
+    stub_token_request
```
ragesoss: `stub_token_request` is for when the system tries to get a csrf token from Wikipedia. We don't need that here.

spec/features/campaign_overview_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  context 'as an user' do
+    it 'should not show the edit buttons' do
+      login_as(user, scope: user)
+      visit "/campaigns/#{campaign.slug}"
+      sleep 1
```
ragesoss: `have_no_css` will wait until it becomes true, so no need to sleep here.

spec/features/campaign_overview_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(page).to have_no_css('.campaign-description .editable-edit')
+    end
+  end
+
+  context 'as an admin' do
+    before do
+      login_as(admin, scope: :user)
+      visit "/campaigns/#{campaign.slug}"
+      sleep 1
```
ragesoss: These pass locally for me without the sleep. Did you find this necessary, or was it just following the pattern in other specs?

app/views/campaigns/overview.html.haml
```diff
@@ -1 +1,31 @@
...
+    .primary
+      = form_for(@campaign, url: campaign_path(@campaign.slug), html: { class: 'module campaign-description editable' }) do
+        = hidden_field(:campaign, :slug)
+        .section-header
+          %h3
+            = succeed ':' do
+              = t('campaign.campaign')
+            = @campaign.title
+          .controls
+            - if current_user&.admin?
```
ragesoss: instead of checking current_user at each point in the view, maybe this makes more sense to set once in the controller as `@editable`. That way it will be easy to patch in the `campaigns_users` permissions when we get there.

PR #1033, opened 2016-11-17
Basic functionality for CampaignsUsers

MusikAnimal:
Bug: [T148804](https://phabricator.wikimedia.org/T148804)

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -90,8 +90,7 @@ class CampaignsController < ApplicationController
   end
 
   def is_organizer?
-    @campaigns_user = @campaign.campaigns_users.where(user_id: current_user.id).first
-    @campaigns_user && @campaigns_user.role == CampaignsUsers::Roles::ORGANIZER_ROLE
+    @campaign.campaigns_users.where(user_id: current_user.id, role: CampaignsUsers::Roles::ORGANIZER_ROLE).any?
   end
 
   def csv_for_role(role)
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -4,7 +4,7 @@ require "#{Rails.root}/lib/analytics/campaign_csv_builder"
 #= Controller for campaign data
 class CampaignsController < ApplicationController
   layout 'admin', only: [:index, :create, :edit]
-  before_action :require_admin_permissions,
+  before_action :require_write_permissions,
                 only: [:create, :update]
   before_action :set_campaign, only: [:overview, :programs, :edit, :update]
 
@@ -65,6 +65,13 @@ class CampaignsController < ApplicationController
   end
 
   private
+
+  def require_write_permissions
+    require_permissions
+    unless Features.open_course_creation?
+      require_admin_permissions
+    end
+  end
 
   def set_campaign
     @campaign = Campaign.find_by(slug: params[:slug])
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -4,9 +4,9 @@ require "#{Rails.root}/lib/analytics/campaign_csv_builder"
 #= Controller for campaign data
 class CampaignsController < ApplicationController
   layout 'admin', only: [:index, :create, :edit]
-  before_action :require_write_permissions,
-                only: [:create, :update]
   before_action :set_campaign, only: [:overview, :programs, :edit, :update]
+  before_action :require_create_permissions, only: [:create]
+  before_action :require_update_permissions, only: [:update]
 
   def index
     @campaigns = Campaign.all
@@ -66,11 +66,17 @@ class CampaignsController < ApplicationController
 
   private
 
-  def require_write_permissions
-    require_permissions
+  def require_create_permissions
     unless Features.open_course_creation?
       require_admin_permissions
     end
+  end
+
+  def require_update_permissions
+    return if current_user&.admin? || is_organizer?
+
+    exception = ActionController::InvalidAuthenticityToken.new('Unauthorized')
+    raise exception
   end
 
   def set_campaign
@@ -81,6 +87,10 @@ class CampaignsController < ApplicationController
     CampaignsUsers.create(user: current_user,
                           campaign: @campaign,
                           role: CampaignsUsers::Roles::ORGANIZER_ROLE)
+  end
+
+  def is_organizer?
+    current_user.is_a?(CampaignsUsers) && current_user.role == CampaignsUsers::Roles::ORGANIZER_ROLE
   end
 
   def csv_for_role(role)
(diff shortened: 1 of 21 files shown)


app/models/campaigns_users.rb
```diff
@@ -0,0 +1,32 @@
...
+#
+#  id          :integer          not null, primary key
+#  created_at  :datetime
+#  updated_at  :datetime
+#  campaign_id :integer
+#  user_id     :integer
+#  role        :integer          default(0)
+#
+
+require "#{Rails.root}/lib/utils"
```
ragesoss: unused

app/models/campaigns_users.rb
```diff
@@ -0,0 +1,32 @@
...
+
+#= Campaign + User join model
+class CampaignsUsers < ActiveRecord::Base
+  belongs_to :campaign
+  belongs_to :user
+
+  validates :campaign_id, uniqueness: { scope: [:user_id, :role] }
+
+  scope :current, -> { joins(:campaign).merge(Campiagn.current).distinct }
```
ragesoss: Campaign has no `current` scope. We also don't have a need as this point for a current scope on CampaignsUsers, I think.

app/models/campaigns_users.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  ####################
+  # CONSTANTS        #
```
ragesoss: white space!

app/models/campaign.rb
```diff
@@ -18,6 +18,7 @@
 #= Campaign model
 class Campaign < ActiveRecord::Base
   has_many :campaigns_courses, class_name: CampaignsCourses
+  has_many :campaigns_users, class_name: CampaignsUsers, dependent: :destroy
```
ragesoss: User needs the same `has_many` (without the dependent bit, as we don't destroy users)

app/models/campaign.rb
```diff
@@ -18,7 +18,9 @@
 #= Campaign model
 class Campaign < ActiveRecord::Base
   has_many :campaigns_courses, class_name: CampaignsCourses
```
ragesoss: I guess while we're at it, we should destroy CampaignsCourses.

spec/controllers/campaigns_controller_spec.rb
```diff
@@ -64,16 +64,27 @@
     let(:user) { create(:user) }
     let(:admin) { create(:admin) }
     let(:campaign) { create(:campaign) }
+    let(:campaign_user) do
+      create(:campaigns_user, user_id: user.id, campaign_id: campaign.id,
```
ragesoss: You'll need to create a new factory in `spec/factories`. Just follow the example from `spec/factories/campaigns_courses.rb` and do `annotate`.

app/controllers/campaigns_controller.rb
```diff
@@ -83,6 +89,10 @@ def add_organizer_to_campaign
                           role: CampaignsUsers::Roles::ORGANIZER_ROLE)
   end
 
+  def is_organizer?
+    current_user.is_a?(CampaignsUsers) && current_user.role == CampaignsUsers::Roles::ORGANIZER_ROLE
```
ragesoss: `current_user` is a `User` object. Maybe instead set @campaigns_user first and then check the role on that.

spec/controllers/campaigns_controller_spec.rb
MusikAnimal: Wow! Here I was adding the schema information by hand to each model!

app/controllers/campaigns_controller.rb
```diff
@@ -90,7 +90,8 @@ def add_organizer_to_campaign
   end
 
   def is_organizer?
-    current_user.is_a?(CampaignsUsers) && current_user.role == CampaignsUsers::Roles::ORGANIZER_ROLE
+    @campaigns_user = @campaign.campaigns_users.where(user_id: current_user.id).first
```
ragesoss: This implicitly assumes there's only one CampaignsUsers per user-campaign pair. Otherwise, `first` might be another role for the same user and campaign.

I suggest something like:
```ruby
@campaign.campaigns_users.where(user_id: current_user.id, role: CampaignsUsers::Roles::ORGANIZER_ROLE).any?
```

spec/controllers/campaigns_controller_spec.rb
```diff
@@ -64,16 +64,27 @@
...
-    it 'returns a 401 if the user is not an admin and feature flag is off' do
+    it 'returns a 401 if the user is not an admin and not an organizer of the campaign' do
       allow(controller).to receive(:current_user).and_return(user)
       allow(Features).to receive(:open_course_creation?).and_return(false)
       post :update, params: { campaign: campaign_params, slug: campaign.slug }
       expect(response.status).to eq(401)
     end
 
+    it 'updates the campaign if the user is an organizer of the campaign' do
+      allow(controller).to receive(:current_user).and_return(campaign_user)
```
ragesoss: You need to create the CampaignsUsers record first, but you still want :current_user to return :user.

app/models/campaign.rb
```diff
@@ -17,8 +17,10 @@
 
 #= Campaign model
 class Campaign < ActiveRecord::Base
-  has_many :campaigns_courses, class_name: CampaignsCourses
+  has_many :campaigns_courses, class_name: CampaignsCourses, dependent: :destroy
+  has_many :campaigns_users, class_name: CampaignsUsers, dependent: :destroy
   has_many :courses, through: :campaigns_courses
+  has_many :users, -> { distinct }, through: :campaigns_users
```
ragesoss: I think 'users' is a dangerous name for this. If I weren't looking at the definition, I would assume it scopes all the users in all the courses, rather than just the organizers of the campaign itself.

If it's not going to be used at this point, we should leave it out entirely. If it is going to be used, let's name it 'campaign_managers' or something else that makes it clear that it's not going through courses.

spec/controllers/campaigns_controller_spec.rb
```diff
@@ -153,9 +153,9 @@
 
     it 'removes the given organizer from the campaign if the current user is a campaign organizer' do
       create(:campaigns_user, user_id: user.id, campaign_id: campaign.id,
-                                                role: CampaignsUsers::Roles::ORGANIZER_ROLE)
+                              role: CampaignsUsers::Roles::ORGANIZER_ROLE)
       allow(controller).to receive(:current_user).and_return(user)
-      put :remove_organizer, params: { slug: campaign.slug, id: user2.id }
+      put :remove_organizer, params: { slug: campaign.slug, id: organizer.id }
```
ragesoss: Should be `organizer.user_id`

MusikAnimal: I'm so stupid >:( It originally went off of the CampaignsUsers id, but then I realized in the view we have the User and not the CampaignsUsers, so to make it easier I adjusted the `remove_organizer` action to go off of user ID.

MusikAnimal:
The failing test in instructor_role_spec.rb passes locally. Maybe try to rebuild (it doesn't appear I'm able to do it)?

ragesoss:
That test is one of the ones that I just converted to Poltergeist and it's still flapping.

On the plus side, now the feature tests are exclusively Poltergeist, which makes it considerably less complicated to set up a working dev environment.

ragesoss:
This seems to be in order.

One UX issue is the flash notice: there's no discernible change if you save a second change, since the flash notice remains the same. It's a bit confusing.

PR #1036, opened 2016-11-20
Don't let left:50% flip in RTL pages

Ladsgroup:
This would solve some parts of https://phabricator.wikimedia.org/T127358


ragesoss:
Nice!

It looks like the same `/*@noflip*/` annotations don't work with the styling of the popover buttons, which are positioned via `transform`: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/stylesheets/modules/_popover.styl#L17

PR #1037, opened 2016-11-22
Proof of concept for confirmation dialog

ragesoss:
@mattfordham I've taken a pass at a confirmation dialog.

Is this a bad way to do it? The idea here is to use the usual React data flow, so the state gets reset by sending an action, receiving it in a store, emitting a change signal, and then reseting the state from storeDidChange. This is overkill for the use case, since no data is being passed. But it doesn't introduce any new concepts or dependencies.


mattfordham:
This is great... I don't think it's overkill and it does indeed correctly follow the pattern; data *could* be passed in the future 👍👍

Issue #985, opened 2016-10-06
It's possible to create slugs with URL-encoded spaces


ragesoss:
See for example: https://dashboard.wikiedu.org/courses/Hunter_College_CUNY/AFPRL_29026%C2%A0Intro_to_Hip_Hop_Culture_and_History_%28Fall_2016%29

That breaks the return links from Wikipedia, and disrupts the url.

The slug function on the front-end should handle this situation.

ragesoss:
Here's the method that generates the slugs from user input: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/utils/course_utils.js#L4

PR #1042, opened 2016-11-30
Prevent whitespace in slugs

ethanlee16:
[Bug T150068](https://phabricator.wikimedia.org/T150068), #985 

Prevents non-breaking spaces from slipping past the `slugify` function.


ragesoss:
Nice! My fix would have been more complex.

ragesoss:
Now covered in the mocha test as well: aa7a8312acab8b8e49683200d85d1aae3b264d60

PR #1055, opened 2016-12-08
Add optional start and end dates (needs feature tests)

MusikAnimal:
Uses HTML5 date input field, with validations

Minor CSS fix to top nav

### DIFF
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -1,6 +1,6 @@
 nav.top-nav
   top 0
-  z-index 10000
+  z-index 100
   background-color white
   border-bottom 1px solid rgba(0,0,0,.1)
   overflow hidden
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -1,6 +1,6 @@
 nav.top-nav
   top 0
-  z-index 100
+  z-index 10000
   background-color white
   border-bottom 1px solid rgba(0,0,0,.1)
   overflow hidden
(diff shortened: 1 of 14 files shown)


ragesoss:
A few things to toss in along with the dates:

* https://phabricator.wikimedia.org/T152675 (error from app/views/campaigns/overview.html.haml when logged out)
    * Can be regression tested by having a controller test for logged out user viewing campaign with `render_views` added to the test
* Hide 'Create Program' button until it is functional, and hide it also if not `Features.open_course_creation?` (because it shouldn't allow you to create a campaign-associated course on Wiki Ed Dashboard)
* The case of `add_organizer` with a non-existent user is not tested.

app/assets/stylesheets/modules/_nav.styl
```diff
@@ -1,6 +1,6 @@
 nav.top-nav
   top 0
-  z-index 100
+  z-index 10000
```
ragesoss: What's this for? If this is necessary to make the styling work as intended, let's fix the underlying cause rather than punch up the z-index higher.

MusikAnimal: A comment from me would have helped :) As you scroll down the page, the alert banners will appear on top of the top nav. I assume this was not intentional. I could have also decreased the z-index of `.notice` but I wasn't sure if that would break anything. It seemed more likely that the top nav would always be on top of everything else.

ragesoss: Okay. Let's leave that alone for now then. The banner behavior has some more general issues that will require some CSS reworking, and this behavior will be part of that.

Also, can you make a screenshot and add an issue?

ragesoss: thanks much!

MusikAnimal: https://phabricator.wikimedia.org/T152803

PR #1059, opened 2016-12-09
Email ID of the user displayed in the profile pages  and visible to admin and user itself.

sejalkhatri:
…o admin

### DIFF
--- a/app/views/users/show.html.haml
+++ b/app/views/users/show.html.haml
@@ -1,9 +1,14 @@
-%br/
+#profile
+  %h2
+    User: #{@user.username}
+  - if current_user == @user || current_user&.admin?
+    %h4
+      &#x2709; #{@user.email}
+
 %br/
 .container
-  %h2
+  %h3
     = t("courses.course_details")
-    #{@user.username}
   - @courses_user.each do |c|
     %a.course{:href => "/courses/#{c.course.slug}"}
       %span.button.border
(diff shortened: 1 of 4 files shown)


app/views/users/show.html.haml
```diff
@@ -1,9 +1,14 @@
-%br/
+#profile
+  %h2
+    User: #{@user.username}
+  - if user_signed_in? && current_user.returning_instructor? && current_user.admin?
```
ragesoss: The email should show if either:
* The current user is the same one whose profile is being shown (as in, current_user == @user )
* The current user is an admin

app/views/users/show.html.haml
```diff
@@ -1,9 +1,14 @@
-%br/
+#profile
+  %h2
+    User: #{@user.username}
+  - if user_signed_in? && (current_user == @user || current_user.admin?)
```
ragesoss: This is fine, but one slightly more concise way to do it would be:

`if current_user == @user || current_user&.admin?`

The safe navigation operator `&.` is really useful in conjuction with 'current_user' because you can still call methods on it without first checking whether there is a current_user.

sejalkhatri: Great! Thanks Sage!

ragesoss:
This looks good. It'll be ready to merge once it has a test demonstrating that the email shows up for an admin, shows up for the same user, but does not show up for another user.

PR #1061, opened 2016-12-11
Create campaign modal

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T144864

### DIFF
--- a/app/assets/stylesheets/modules/_campaign.styl
+++ b/app/assets/stylesheets/modules/_campaign.styl
@@ -49,3 +49,6 @@
 
 .create-campaign-button
   margin-bottom: 50px
+
+.wizard .right
+  width: 100% !important
--- a/app/assets/stylesheets/modules/_campaign.styl
+++ b/app/assets/stylesheets/modules/_campaign.styl
@@ -46,3 +46,6 @@
 
     .campaign-use-dates
       display: block
+
+  .campaign-list
+    margin-top: 50px
--- a/app/assets/stylesheets/modules/_campaign.styl
+++ b/app/assets/stylesheets/modules/_campaign.styl
@@ -47,8 +47,8 @@
     .campaign-use-dates
       display: block
 
-.create-campaign-button
-  margin-bottom: 50px
+  .create-campaign-button
+    margin-bottom: 50px
 
-.wizard .right
-  width: 100% !important
+  .wizard .right
+    width: 100% !important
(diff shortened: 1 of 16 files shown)


app/assets/stylesheets/modules/_campaign.styl
```diff
@@ -46,3 +46,9 @@
 
     .campaign-use-dates
       display: block
+
+.create-campaign-button
+  margin-bottom: 50px
+
+.wizard .right
+  width: 100% !important
```
ragesoss: This seems very dangerous, as `wizard` is used in a lot of places. Can we find a better way to ensure the correct width in this context without `!important`?

app/views/campaigns/index.html.haml
```diff
@@ -1,21 +1,23 @@
...
   .container
     %hr
-    %h1
-      Create New Campaign
-    %div
-      =form_tag do
-        = text_field :campaign, 'title', placeholder: 'Title'
-        = submit_tag 'Create Campaign', class: 'button dark'
-%section
+    - if current_user && Features.open_course_creation?
```
ragesoss: Admins on Wiki Ed dashboard also need to be able to create a campaign, even with open_course_creation disabled.

app/views/campaigns/_create_modal.html.haml
```diff
@@ -0,0 +1,44 @@
...
+                = label(:campaign, :start, t('courses.creator.start_date') + ':')
+                = date_field(:campaign, :start, placeholder: '2016-12-31')
+              .form-group
+                = label(:campaign, :end, t('courses.creator.end_date') + ':')
+                = date_field(:campaign, :end, placeholder: '2016-12-31')
+
+          .column
+            .form-group
+              = label(:campaign, :description, t('campaign.description') + ':')
+              = text_area(:campaign, :description)
```
ragesoss: This needs a maximum size to guarantee that it won't be more than the database will accept.

app/models/campaign.rb
```diff
@@ -118,6 +124,11 @@ def validate_dates
     end
   end
 
+  def set_slug
+    # Strip everything but letters and digits, and convert spaces to underscores
+    self.slug = title.downcase.gsub(/[^\w0-9 ]/, '').tr(' ', '_') unless self.slug.present?
```
MusikAnimal: #1062

PR #1066, opened 2016-12-13
Start on creating program from campaign, new campaigns.json endpoint

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T152792


app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -51,6 +51,12 @@ const CourseCreator = React.createClass({
 
   componentWillMount() {
     CourseActions.addCourse();
+
+    const campaignParam = location.search.match(/\?.*?campaign_id=(\d+)/);
```
ragesoss: You'll need to do window.location.

app/views/campaigns/index.json.jbuilder
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+if @campaign
```
ragesoss: This doesn't make sense as an index to me. Its purpose is to return a single campaign, so this would be more like `show`, where we'd have the campaign id in the route. So that would mean calling `/campaigns/1.json` to get campaign 1.

I think that makes more sense than overloading the index method to do very different things in html vs json.

spec/features/course_creation_spec.rb
```diff
@@ -346,6 +348,36 @@ def go_through_researchwrite_wizard
...
+  describe 'course created from a campaign', js: true do
+    let(:campaign) do
+      create(:campaign,
+             id: 10001,
+             title: 'My Awesome Campaign',
+             description: 'This is the best campaign')
+    end
+
+    before do
+      ENV['default_course_type'] = 'BasicCourse'
```
ragesoss: instead of setting ENV and unsetting it, you can do `allow(Features).to receive(:default_course_type)`

PR #1074, opened 2016-12-15
Sidekiq alert email#1013

patmbolger:
Convert from sync to async operation via sidekiq: Emailing target user from AlertsController#create

### DIFF
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -67,7 +67,8 @@ class Alert < ActiveRecord::Base
   def email_target_user
     return if emails_disabled?
     return if target_user.nil?
-    AlertMailerWorker.schedule_email(alert: self, user: target_user)
+    AlertMailer.alert(self, target_user).deliver_now
+    update_attribute(:email_sent_at, Time.now)
   end
 
   # Disable emails for specific alert types in application.yml, like so:
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -67,8 +67,7 @@ class Alert < ActiveRecord::Base
   def email_target_user
     return if emails_disabled?
     return if target_user.nil?
-    AlertMailer.alert(self, target_user).deliver_now
-    update_attribute(:email_sent_at, Time.now)
+    AlertMailerWorker.schedule_email(alert: self, user: target_user)
   end
 
   # Disable emails for specific alert types in application.yml, like so:
--- a/app/workers/alert_mailer_worker.rb
+++ b/app/workers/alert_mailer_worker.rb
@@ -3,15 +3,12 @@
 class AlertMailerWorker
   include Sidekiq::Worker
 
-  def self.schedule_email(alert)
-    perform_async(alert.id, alert.target_user.id)
+  def self.schedule_email(alert_id)
+    perform_async(alert_id)
   end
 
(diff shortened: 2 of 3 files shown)


app/models/alert.rb
```diff
@@ -67,8 +67,7 @@ def email_course_admins
   def email_target_user
     return if emails_disabled?
     return if target_user.nil?
-    AlertMailer.alert(self, target_user).deliver_now
-    update_attribute(:email_sent_at, Time.now)
+    AlertMailerWorker.schedule_email(alert: self, user: target_user)
```
ragesoss: The place we should intervene is the controller, rather than the model.

In some cases, there are batches of alerts generated and sent, and we throttle those email sending events to make gmail happy. The only reason we want to do them async is to avoid the specific case of sending an email within an http request.

The place to call the worker would be `AlertsController#email_target_user` (passing just the `alert.id` to `perform_async`)

patmbolger: I see now ... I was assuming we wanted to send *all* alert emails asynchronously.

See adjusted PR.

app/workers/alert_mailer_worker.rb
```diff
@@ -3,14 +3,14 @@
 class AlertMailerWorker
   include Sidekiq::Worker
 
-  def self.schedule_email(alert:, user:)
-    perform_async(alert.id, user.id)
+  def self.schedule_email(alert)
```
ragesoss: For consistency in the way workers are set up, we can leave alert as a named parameter.

app/workers/alert_mailer_worker.rb
```diff
@@ -3,14 +3,14 @@
...
+    perform_async(alert.id, alert.target_user.id)
```
ragesoss: We can just pass the `alert.id` to `perform_async`

app/workers/alert_mailer_worker.rb
```diff
@@ -3,14 +3,14 @@
...
   end
 
   def perform(alert_id, user_id)
     alert = Alert.find(alert_id)
     user  = User.find(user_id)
-    
+
     AlertMailer.alert(alert, user).deliver_now
```
ragesoss: We should just run alert.email_target_user inside the `perform`. The less it knows about it's job the better.

app/workers/alert_mailer_worker.rb
patmbolger: I'm following then pattern of the other workers and the `schedule_...` method receives AR objects, which negates the need to an object retrieval in that method.  (?)

ragesoss: The way sidekiq works is that it receives very simple things that lack state, because it stores them in Redis. So typically it should only get record IDs and not full objects. That's all fine here.

In this case though, we only need one object id — for the alert — to do all the work we need to do. We removed a call to `alert.email_target_user`, so we should just fetch that alert and call `email_target_user` on it. No need to give the worker any knowledge about the target user or how the method works.

ragesoss: The `self.schedule_email method` here, following the pattern I set up on other workers, is just to make the interface more explicit, so that you don't have to know the sidekiq limitations (or do the conversion of object to object id) when just using the worker.

PR #1077, opened 2016-12-16
Find programs page

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T148803

### DIFF
--- a/app/assets/stylesheets/modules/_selects.styl
+++ b/app/assets/stylesheets/modules/_selects.styl
@@ -34,5 +34,6 @@
     right 191px
   select.sorts
     background url('../images/sort.svg') no-repeat 48% white
+    color: transparent
     -webkit-appearance none
     text-indent -99px
--- a/app/assets/stylesheets/modules/_selects.styl
+++ b/app/assets/stylesheets/modules/_selects.styl
@@ -15,7 +15,6 @@
   cursor pointer
   float left
   margin 10px 10px 10px 0
-  overflow hidden
   +above(desktop)
     margin 0
     position absolute
--- a/app/assets/stylesheets/modules/_selects.styl
+++ b/app/assets/stylesheets/modules/_selects.styl
@@ -14,7 +14,6 @@
 .campaign-select, .sort-select
   cursor pointer
   float left
-  height 42px
   margin 10px 10px 10px 0
   overflow hidden
   +above(desktop)
@@ -34,6 +33,5 @@
     right 191px
   select.sorts
     background url('../images/sort.svg') no-repeat 48% white
-    color: transparent
     -webkit-appearance none
-    text-indent -99px
+    text-indent -200px
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -138,7 +138,8 @@ class CampaignsController < ApplicationController
   end
 
   def set_campaign
-    return if @campaign = Campaign.find_by(slug: params[:slug])
+    @campaign = Campaign.find_by(slug: params[:slug])
+    return if @campaign
     raise ActionController::RoutingError.new('Not Found'), 'Campaign does not exist'
   end
 
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -138,7 +138,8 @@ class CampaignsController < ApplicationController
   end
 
   def set_campaign
-    @campaign = Campaign.find_by(slug: params[:slug])
+    return if @campaign = Campaign.find_by(slug: params[:slug])
+    raise ActionController::RoutingError.new('Not Found'), 'Campaign does not exist'
   end
 
   def set_presenter
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -25,6 +25,10 @@ class CampaignsController < ApplicationController
     end
   end
 
+  def new
+    redirect_to campaigns_path(create: true)
+  end
+
   def create
     @campaign = Campaign.create(campaign_params)
 
--- a/app/controllers/explore_controller.rb
+++ b/app/controllers/explore_controller.rb
@@ -6,9 +6,8 @@ class ExploreController < ApplicationController
   def index
     # 'cohort' is the old name for campaign. We accept 'cohort' as an alternative
     # Redirect to new campaign overview page if a parameter is given, for backwards compatibility
-    if campaign = params[:campaign] || params[:cohort]
-      redirect_to campaign_path(campaign)
-    end
+    campaign = params[:campaign] || params[:cohort]
+    redirect_to campaign_path(campaign) if campaign
 
     @presenter = CoursesPresenter.new(current_user, ENV['default_campaign'])
   end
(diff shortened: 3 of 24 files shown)


lib/course_creation_manager.rb
```diff
@@ -74,9 +74,7 @@ def set_initial_campaign
     return unless Features.open_course_creation?
 
     if @initial_campaign_params.present?
-      parent_campaign = Campaign.find_by_id(@initial_campaign_params[:initial_campaign_id])
-      @overrides[:campaigns] = [parent_campaign]
-      @overrides[:description] = parent_campaign.template_description
+      @overrides[:campaigns] = [Campaign.find_by_id(@initial_campaign_params[:initial_campaign_id])]
```
ragesoss: whoops! thanks for catching that.

ragesoss: Oh, I see this, is on an older branch...

app/models/campaign.rb
```diff
@@ -42,7 +43,7 @@ class Campaign < ActiveRecord::Base
   # Scopes #
   ##########
 
-  scope :active, -> { where('end > ?', Time.zone.now) }
+  scope :active, -> { where('end > ? OR (start IS NULL AND end IS NULL)', Time.zone.now) }
```
MusikAnimal: Having just one NULL would cause an error, currently https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/models/campaign.rb#L120

app/views/courses/_campaign_courses.html.haml
```diff
@@ -1,7 +1,10 @@
 %section#courses
   .section-header
-    %h3= [t("#{presenter.course_string_prefix}.courses"), presenter.campaign.title].join(" | ")
-    .sort-select.inset
+    %h3
+      = t("#{presenter.course_string_prefix}.courses")
+      - if current_user && (current_user.admin? || Features.open_course_creation?)
```
ragesoss: Since we're using a presenter already, this logic should be done as a clearly named method in the presenter.

app/views/campaigns/_campaigns.html.haml
```diff
@@ -0,0 +1,77 @@
+%section#campaigns
+  .section-header
+    %h3
+      = t('campaign.active_campaigns')
+      - if current_user && (current_user.admin? || Features.open_course_creation?)
```
ragesoss: Do this with a presenter function.

app/models/campaign.rb
ragesoss: So then why not just check for end IS NULL? that matches the use case here, and isn't coupled to whether or not we validate co-presence of start and end.

ragesoss:
@sejalkhatri this one is ready for review. I'll be reviewing it; if you would like to as well, please do.

app/controllers/campaigns_controller.rb
```diff
@@ -134,7 +138,8 @@ def require_write_permissions
   end
 
   def set_campaign
-    @campaign = Campaign.find_by(slug: params[:slug])
+    return if @campaign = Campaign.find_by(slug: params[:slug])
```
ragesoss: Set the campaign separately from the conditional return. Doing assignment in a condition is a rubocop style violation, because of the potential confusions between `=` and `==`.

app/controllers/explore_controller.rb
```diff
@@ -1,14 +1,15 @@
...
 
   def index
     # 'cohort' is the old name for campaign. We accept 'cohort' as an alternative
-    # parameter to keep old incoming links from breaking.
-    campaign = params[:campaign] || params[:cohort] || ENV['default_campaign']
-    @presenter = CoursesPresenter.new(current_user, campaign)
-    return unless @presenter.campaign.nil?
-    raise ActionController::RoutingError.new('Not Found'), 'Campaign does not exist'
+    # Redirect to new campaign overview page if a parameter is given, for backwards compatibility
+    if campaign = params[:campaign] || params[:cohort]
```
ragesoss: Avoid assignment in condition here too.

app/models/campaign.rb
```diff
@@ -43,7 +43,7 @@ class Campaign < ActiveRecord::Base
   # Scopes #
   ##########
 
-  scope :active, -> { where('end > ?', Time.zone.now) }
+  scope :active, -> { where('end > ? OR end IS NULL', Time.zone.now) }
```
ragesoss: Did you get a chance to try the ActiveRecord `.or` approach? If it works, I'd rather use that than the raw SQL.

app/presenters/courses_presenter.rb
```diff
@@ -22,15 +22,36 @@ def campaign
...
+  def active_courses_by_recent_edits
+    active_courses.sort_by { |course| [-course.recent_revision_count, course.title] }
+  end
+
+  def campaigns_by_num_courses
+    # Sort first by number of courses, then by campaign title
+    campaigns.sort_by { |campaign| [-campaign.courses.count, campaign.title] }
+  end
+
+  def can_create?
```
ragesoss: Since we're mixing courses and campaigns here, this is probably better named `can_create_campaign` to avoid confusion.

app/assets/stylesheets/modules/_selects.styl
```diff
@@ -34,5 +34,6 @@
     right 191px
   select.sorts
     background url('../images/sort.svg') no-repeat 48% white
+    color: transparent
```
ragesoss: Is this the cause of the white-on-white select menu?

white on white

app/controllers/explore_controller.rb
```diff
@@ -1,14 +1,15 @@
...
+      redirect_to campaign_path(campaign)
```
ragesoss: Does this remove the ability to access the list of unsubmitted courses (the `NullCampaign` in `CoursesPresenter`)? We need some way to see that set of courses. It could even be a separate view and route that just renders the courses list with `campaign_param == 'none'`. Alternatively, we could continue the same approach and special-case the unsubmitted set as it's own fake campaign, and handle that in campaign controller to show it like any other campaign.

app/presenters/courses_presenter.rb
MusikAnimal: Well that's just it, I'm using it to see if they can create a course as well as a campaign. Maybe `can_create_campaign_or_course`?

app/assets/stylesheets/modules/_selects.styl
MusikAnimal: Dammit Firefox. The issue was before, the selected option was showing up on top of the icon of the button. I observed that in Chrome. I'll try some other tricks

app/presenters/courses_presenter.rb
ragesoss: Ah, I see. For Wiki Ed, this will add the create buttons from explore, but only for admins. Which is fine.

app/assets/stylesheets/modules/_selects.styl
```diff
@@ -35,4 +33,4 @@
   select.sorts
     background url('../images/sort.svg') no-repeat 48% white
     -webkit-appearance none
-    text-indent -99px
+    text-indent -200px
```
MusikAnimal: @ragesoss Tested this CSS on Mac in Chrome, Firefox and Safari and it looks good. Not sure about other platforms

MusikAnimal:
That test passes locally for me

app/assets/stylesheets/modules/_selects.styl
ragesoss: lgtm locally

PR #1078, opened 2016-12-16
Implement Confirm component for JoinCourse Button

sejalkhatri:


app/assets/javascripts/components/common/confirm.jsx
```diff
@@ -1,29 +1,63 @@
...
 const Confirm = React.createClass({
   displayName: 'Confirm',
 
   propTypes: {
     onConfirm: React.PropTypes.func.isRequired,
     onCancel: React.PropTypes.func.isRequired,
-    message: React.PropTypes.string.isRequired
+    message: React.PropTypes.string.isRequired,
+    showInput: React.PropTypes.bool,
+    passcode: React.PropTypes.string,
```
ragesoss: We should name all the props in a general way, so that they don't refer specifically to this use case. In this case, I don't think we need to have a prop for this at all. We can just pass the state to the onConfirm function.

app/assets/javascripts/components/common/confirm.jsx
```diff
@@ -1,29 +1,63 @@
...
-    this.props.onConfirm();
+    this.props.onConfirm(this.props.passcode);
   },
 
   onCancel() {
     this.props.onCancel();
   },
 
+  onChange(_valueKey, value) {
+    this.setState({ passcode: value });
```
ragesoss: We can call this userInput instead of passcode, so that it works as a general prompt.

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -13,21 +16,40 @@ const AvailableActions = React.createClass({
...
   },
 
-  mixins: [CourseStore.mixin],
+  mixins: [CourseStore.mixin, ConfirmationStore.mixin],
 
   getInitialState() {
-    return getState();
+    return ({
+      course: CourseStore.getCourse(),
+      showConfirm: false,
```
ragesoss: You can probably leave out all the null values in initial state. If not defined, they will all be falsey by default, so it should behave the same way without them.

app/assets/javascripts/components/common/confirm.jsx
sejalkhatri: ok, didn't know we can do that. Thanks!

Issue #995, opened 2016-10-17
WikiEdits should handle Wikimedia API downtime gracefully


ragesoss:
Currently, it causes errors: https://sentry.wikiedu.org/wikiedu/production/issues/967/

```
JSON::ParserError: 784: unexpected token at '<!DOCTYPE html>
<html lang=en>
<meta charset=utf-8>
<title>Wikimedia Error</title>
<style>
* { margin: 0; padding: 0; }
body { background: #fff; font: 15px/1.6 sans-serif; color: #333; }
.content { margin: 7% auto 0; padding: 2em 1em 1em; max-width: 640px; }
.footer { clear: both; margin-top: 14%; border-top: 1px solid #e5e5e5; background: #f9f9f9; padding: 2em 0; font-size: 0.8em; text-align: center; }
img { float: left; margin: 0 2em 2em 0; }
a img { border: 0; }
h1 { margin-top: 1em; font-size: 1.2em; }
p { margin: 0.7em 0 1em 0; }
a { color: #0645AD; text-decoration: none; }
a:hover { text-decoration: underline; }
code { font-family: sans-serif; }
.text-muted { color: #777; }
</style>
<div class="content" role="main">
<a href="//www.wikimedia.org"><img src="//www.wikimedia.org/static/images/wmf.png" srcset="//www.wikimedia.org/static/images/wmf-2x.png 2x" alt=Wikimedia width=135 height=135></a>
<h1>Error</h1>
<p>Our servers are currently under maintenance or experiencing a technical problem. Please <a href="" title="Reload this page" onclick="window.location.reload(false); return false">try again</a> in a few&nbsp;minutes.</p><p>See the error message at the bottom of this page for more&nbsp;information.</p>
</div>
<div class="footer">
<p>If you report this error to the Wikimedia System Administrators, please include the details below.</p>
<p class="text-muted"><code>
Request from 2600:3c03::f03c:91ff:fe08:7973 via cp1053 cp1053, Varnish XID 741374242<br>Error: 503, Service Unavailable at Tue, 08 Nov 2016 13:49:01 GMT</code></p></div></html>'
  from /usr/local/rvm/gems/ruby-2.3.1/gems/passenger-5.0.29/src/ruby_supportlib/phusion_passenger/utils.rb:113:in `block in create_thread_and_abort_on_exception'
[...]

PR #1083, opened 2016-12-17
[WIP] Added downtime exception handling

feelfreelinux:
Issue #995 
I don't know, should I use nice message, or exception report? Probably that's not all, but i need code review

### DIFF
--- a/lib/wiki_edits.rb
+++ b/lib/wiki_edits.rb
@@ -113,8 +113,11 @@ class WikiEdits
     return { status: 'no current user' } unless current_user
     @access_token = oauth_access_token(current_user)
     get_token = @access_token.get("#{@api_url}?action=query&meta=tokens&format=json")
-
-    token_response = JSON.parse(get_token.body)
+    begin
+      token_response = JSON.parse(get_token.body)
+    rescue JSON::ParserError => e
+      Raven.capture_exception(e)
+    end
     WikiResponse.capture(token_response, current_user: current_user,
                                          type: 'tokens')
     return { status: 'failed' } unless token_response.key?('query')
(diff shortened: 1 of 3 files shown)


feelfreelinux:
Looks unrelated for me

lib/wiki_edits.rb
```diff
@@ -113,8 +113,11 @@ def get_tokens(current_user)
     return { status: 'no current user' } unless current_user
     @access_token = oauth_access_token(current_user)
     get_token = @access_token.get("#{@api_url}?action=query&meta=tokens&format=json")
-
```
ragesoss: I think the better approach here, to make this as specific as possible, will be something like returning or skipping the JSON parse if the response code is 503.

This method should always return a hash that includes the status (see the return line further down).

This is a good case for writing a failing test case first, and then proceeding with the solution.

spec/support/request_helpers.rb
```diff
@@ -20,13 +20,7 @@ def stub_token_request_failure
...
 
   def stub_wikimedia_error
-    wikimedia_error = '<meta charset=utf-8><title>Wikimedia Error</title><div class="content" role="main">
-      <a href="//www.wikimedia.org"><img src="//www.wikimedia.org/static/images/wmf.png" srcset="//www.wikimedia.org/static/images/wmf-2x.png 2x" alt=Wikimedia width=135 height=135></a>
-      <h1>Error</h1>
-      <p>Our servers are currently under maintenance or experiencing a technical problem. Please <a href="" title="Reload this page" onclick="window.location.reload(false); return false">try again</a> in a few&nbsp;minutes.</p><p>See the error message at the bottom of this page for more&nbsp;information.</p></div>
-      <div class="footer">
-      <p>If you report this error to the Wikimedia System Administrators, please include the details below.</p>
-      <p class="text-muted"><code>'
+    wikimedia_error = '<!DOCTYPE html><html lang=en><meta charset=utf-8><title>Wikimedia Error</title>'
```
ragesoss: close the html tag to keep it valid-ish html.

PR #1084, opened 2016-12-17
Sidekiq course delete#1013

patmbolger:
Convert from sync to async operation via sidekiq: Deleting course from CoursesController#destroy

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -46,7 +46,7 @@ class CoursesController < ApplicationController
 
   def destroy
     validate
-    DeleteCourseWorker.schedule_edits(course: @course, current_user: current_user)
+    DeleteCourseWorker.schedule_deletion(course: @course, current_user: current_user)
     render json: { success: true }
   end
 
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -46,11 +46,8 @@ class CoursesController < ApplicationController
 
   def destroy
     validate
+    DeleteCourseWorker.schedule_edits(course: @course, current_user: current_user)
     @course.destroy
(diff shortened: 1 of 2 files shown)


ragesoss:
Yikes. I didn't think this one through. We can't do this async as-is, because the course will already be destroyed by the time sidekiq gets to it.

We should either keep this one syncronous, or destroy the course inside the worker. I *think* the latter would be fine.

ragesoss:
Yeah, destroying the course within the worker should be fine.

patmbolger:
Yes, I swapped the order of actions so that the local course is destroyed *after* the async job is scheduled.  Unless we think there is a chance that the course `destroy` will fail, I did not see much difference between doing it like this or performing the `destroy` within the worker.

Do you still want to do it in the worker?

ragesoss:
It would fail when the worker tried to perform, because the course would already be gone from the database. You could schedule it, but looking up the course by ID would cause an error. (Doesn't surface in the test, because sidekiq runs in-thread syncronously in the test suite.)

patmbolger:
Ah yes, of course.  Good point.

app/controllers/courses_controller.rb
```diff
@@ -46,11 +46,7 @@ def update
 
   def destroy
     validate
-    @course.destroy
-    WikiCourseEdits.new(action: :update_course,
-                        course: @course,
-                        current_user: current_user,
-                        delete: true)
+    DeleteCourseWorker.schedule_edits(course: @course, current_user: current_user)
```
ragesoss: Since this does more than just delete it, let's name the method differently. `schedule_deletion` perhaps?

ragesoss: Change name here too.

PR #1086, opened 2016-12-18
[Untested] Add wikidata

Ladsgroup:


ragesoss:
This is the right idea, but requires some improvements to the validations. I took a pass at it here: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/f32db0d7889af20a26fc0c6db02f446340c8b714

After that, it will also require some tweaks to the frontend to allow selection of wikidata.

ragesoss:
Updated PR  #1087

PR #1087, opened 2016-12-19
Add support for Wikidata

ragesoss:
This adds Wikidata as a project that can be selected for assignments; revisions to Wikidata will be imported just like from any other project.

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -102,7 +102,7 @@ class AssignmentsController < ApplicationController
     home_wiki = @course.home_wiki
     language = params[:language].present? ? params[:language] : home_wiki.language
     project = params[:project].present? ? params[:project] : home_wiki.project
-    @wiki = Wiki.find_or_create_by(language: language, project: project) || home_wiki
+    @wiki = Wiki.get(language: language, project: project) || home_wiki
   end
 
   def set_new_assignment
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -102,7 +102,7 @@ class AssignmentsController < ApplicationController
     home_wiki = @course.home_wiki
     language = params[:language].present? ? params[:language] : home_wiki.language
     project = params[:project].present? ? params[:project] : home_wiki.project
-    @wiki = Wiki.get(language: language, project: project) || home_wiki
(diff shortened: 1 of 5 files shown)


app/controllers/assignments_controller.rb
```diff
@@ -102,7 +102,7 @@ def find_or_create_wiki
     home_wiki = @course.home_wiki
     language = params[:language].present? ? params[:language] : home_wiki.language
     project = params[:project].present? ? params[:project] : home_wiki.project
-    @wiki = Wiki.find_or_create_by(language: language, project: project) || home_wiki
+    @wiki = Wiki.get(language: language, project: project) || home_wiki
```
sejalkhatri: this looks good.
although i got confused with the rails routing get at the first.

sejalkhatri: Ok then I think there's no need to change the method name `find_or_create_by` works

ragesoss: `find_or_create_by` is built-in method that comes from ActiveModel, while `get` here is an original method in `Wiki`. Although we could rename it `find_or_create_by`, in which case it would override the ActiveModel version.

The key change here is that I updated `get` so that it will handle the special case of wikidata, where it should create a wiki record that has langauge `nil` (even if the user inluded a language).

sejalkhatri: I get it! In that case we can rename it as `get_or_create` or somethng relevant...As `get` is confusing.

ragesoss: Done.

app/models/wiki.rb
```diff
@@ -44,23 +50,53 @@ class Wiki < ActiveRecord::Base
...
-    "https://#{language}.#{project}.org"
+    if language
+      "https://#{language}.#{project}.org"
+    else
+      "https://#{MULTILINGUAL_PROJECTS[project]}"
+    end
   end
 
   def api_url
     "#{base_url}/w/api.php"
```
sejalkhatri: should we mention which api url are we talking about in the comment?

ragesoss: If there's a comment you would find helpful here, you could add it in a separate pull request. Since this PR doesn't change `api_url` or how it is used, that would usually be out of scope for the code review.

sejalkhatri: Fine! everything else makes sense to me..

PR #1090, opened 2016-12-19
resolve #654 Course creator slug updates when data is valid.

### LINKED ISSUE #654: Course creator slug should only update when data is valid
Generally, there should be indication in this dialog of which fields are required.  For this bug, there is a string "/()" in the lower left corner, which shows the course slug once course title, school, and term are filled out.  I would suggest not displaying the slug until there is data in all of these fields.  Update interval could be much longer as well, like a second.

sejalkhatri:


app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -65,7 +65,12 @@ const CourseCreator = React.createClass({
 
   storeDidChange() {
     this.setState(getState());
-    this.state.tempCourseId = CourseUtils.generateTempId(this.state.course);
+    if (this.state.course.term !== '' && this.state.course.school !== '' && this.state.title !== '') {
```
ragesoss: The term field is not shown for the global WMF version of the course creator. This switches on the default course type (`this.state.default_course_type === 'ClassroomProgramCourse'`).

A slug valid without a term, it should just be shown without the trailing "_()".

sejalkhatri: So this condition should execute only when this.state.default_course_type === 'ClassroomProgramCourse'  and the trailing _() removed in the else section....
OR do I need to make changes in the course_utils.js

ragesoss: I think this function should just be in charge of returning empty string if either title or school is missing, and course_utils.js should be updated to not include the _() if term is empty.

sejalkhatri: ya thats efficient

ragesoss:
This works nicely. It should also include a test case for the new CourseUitls behavior, in `test/utils/course_utils.spec.js`.

PR #1091, opened 2016-12-20
[WIP] Popup is not eager to popdown

### LINKED ISSUE #648: Popup is not eager to popdown
no_popdown

Clicking outside the box should cause it to disappear.
The <escape> key should also do so.

Neither does, I have to click outside the popup but within the course row to make the popup go away.

arreche:
fix #648

### DIFF
--- a/app/assets/javascripts/components/high_order/expandable.jsx
+++ b/app/assets/javascripts/components/high_order/expandable.jsx
@@ -24,11 +24,8 @@ const Expandable = function (Component) {
       return UIActions.open(this.refs.component.getKey());
     },
 
-    handleClickOutside() {
-      // TODO: Fire UIActions.close?
-      this.setState({
-        is_open: false
-      });
+    handleClickOutside(e) {
+      if (this.state.is_open) this.open(e);
     },
 
     render() {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/high_order/expandable.jsx
```diff
@@ -22,6 +24,13 @@ const Expandable = function (Component) {
       return UIActions.open(this.refs.component.getKey());
     },
 
+    handleClickOutside() {
+      // TODO: Fire UIActions.close?
```
ragesoss: It's poorly named, but `UIActions.open` actually toggles the state of it. (See how UIStore handles the `OPEN_KEY` signal for the details.)

With this version, it closes the popover, but the UIStore still registers it as open, so it will take two clicks to open it again. This function also tries to set state on every outside click, even when the popover is not open.

If you change the setState to instead call `this.open(e)`, and only if it's already open, then this seems to work perfectly.

arreche:
Currently, I am struggling writing a test.

ragesoss:
One thing I noticed before was that the onClickOutside function was firing on every click for every expandable on the page. We should probably use the enable and disable functions to regulate that as well: https://github.com/Pomax/react-onclickoutside#regulate-which-events-to-listen-for

ragesoss:
@arreche thanks for showing the way on this!

I took the solution you found, and forked the 'Expandable' component so that some components could retain the old behavior and not close on outside click: da22ff184d4d60bc67302dac8bdb07cea778da40

PR #1098, opened 2016-12-23
Added icon to unexisting article

feelfreelinux:
Added icon to unexisting article in courses page.

### DIFF
--- a/app/views/courses/_assignments.json.jbuilder
+++ b/app/views/courses/_assignments.json.jbuilder
@@ -8,6 +8,10 @@ json.assignments course.assignments do |assignment|
     json.article_rating assignment.article.rating
     json.article_rating_num rating_priority(assignment.article.rating)
     json.article_pretty_rating rating_display(assignment.article.rating)
+  else
+    json.article_rating nil
+    json.article_rating_num nil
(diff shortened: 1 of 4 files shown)


ragesoss:
This looks like a good start. It needs an explanation of the symbol on mouseover, as well. For that, it will require passing both a pretty_rating and a rating (`does_not_exist`), and a new interface message (on `config/locales/en.yml`, with a description of the message in `qqq.yml`) for the rating.

app/views/courses/_assignments.json.jbuilder
```diff
@@ -8,6 +8,10 @@ json.assignments course.assignments do |assignment|
     json.article_rating assignment.article.rating
     json.article_rating_num rating_priority(assignment.article.rating)
     json.article_pretty_rating rating_display(assignment.article.rating)
+  else
+    json.article_rating nil
```
ragesoss: 'does not exist', not `nil`

ragesoss:
This looks good. One minor tweak for consistency: let's handle `rating_num` the same way as the other values, so that if we change the json value from `nil`to something else, it will make it all the way through to the Assignment component.

PR #1099, opened 2016-12-23
Multi conditional questions

ragesoss:
@mattfordham This mostly works, but if the last question is one of the multiple conditionals, it can display some incorrect behavior around the 'Submit' vs. 'Next' button.

Setup:
* Create question with three options: Yes, No, Maybe
* Create three questions conditional on different answers (in a different order): Yes, Maybe, No

Incorrect behavior:
* Select ['Yes'] — Conditional question button says 'Next' but actually advances to the end.
* Select ['Yes', 'No', 'Maybe] — Second conditional (for 'No') says 'Submit' but actually advances to third conditional, which says 'Next' but actually advances to the end.

### DIFF
--- a/app/assets/javascripts/surveys/modules/Survey.js
+++ b/app/assets/javascripts/surveys/modules/Survey.js
@@ -531,7 +531,6 @@ const Survey = {
       }
 
       this.addConditionalQuestionToStore(question_id, $question);
-      
       this.addListenersToConditional($question, conditionalOptions);
       this.surveyConditionals[question_id].currentAnswers = [];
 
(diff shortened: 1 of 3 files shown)


mattfordham:
Will check it out. Thanks for the detail

ragesoss:
Cool. I think the only way we're going to be happy with the survey code is if we move rapidfire directly into the codebase and then rewrite all the view logic to happen in the javascript. Meanwhile, one more jquery hack is no big deal.

mattfordham:
Yup, I totally agree

PR #1102, opened 2016-12-24
Added links to campaigns list.

feelfreelinux:
In course overview details, added link to campaigns list.

Bug: T153926


ragesoss:
It's necessary to include a unique `key` for each span, as done in `InlineUsers`. Otherwise it causes console errors when you go to edit details.

PR #1109, opened 2016-12-27
Fix sandbox training for students

Protonk:
Correct errors in sandbox training, inc:
* Correct article namespace
* Warn students about copying from edit mode

### DIFF
--- a/training_content/wiki_ed/slides/9-sandboxes/902-what-is-a-sandbox.yml
+++ b/training_content/wiki_ed/slides/9-sandboxes/902-what-is-a-sandbox.yml
@@ -3,7 +3,7 @@ title: "What's a sandbox?"
 summary:
 content: |
   A "sandbox" is your personal wiki page where you can practice editing, plan
-  out articles, draft articles, or just experiment, without impacting the
+  out work, draft articles, or just experiment, without impacting the
   article "mainspace” on Wikipedia—which is what we call live articles.
 
   To go to your default sandbox page, simply click the sandbox link, which can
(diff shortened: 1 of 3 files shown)


Protonk:
I can squash if need be, but I'll have to fire up my gh client (just did this in the browser)

ragesoss:
How about: 
 
If you're editing an existing article, a sandbox can be a good place to prepare your first updates by copying a small portion of the article that you want to change or extend. Do not try to overhaul an entire article from the sandbox.

1. Open the original article in **Edit** mode. (References and other templates will break if you copy from **Read** mode.)
2. Select the portion you want to work on — a few paragraphs at most — and copy it.
3. Open your sandbox in Edit mode and paste the copied article content, then save it.
4. Re-enter edit mode in your sandbox, make your changes, and save them.
5. When you're ready to move your small updates back into the article, use the same procedure in reverse: copy from your sandbox in **Edit** mode, and paste the updated content into the article.
  
Once you've gotten the hang of it you can make further edits directly in the article, or repeat the sandbox procedure one portion at a time.

If you're working with a new article that you've created, and doesn't exist on
Wikipedia already, you can "move" your sandbox page into mainspace. In the
next slides, we'll show you how.

Protonk:
I've updated w/ those changes. @ragesoss, do I need to manually wrap the text?

ragesoss:
It parses this yaml file in `|` mode, which means a single linebreak is ignored. So for source code consistency, you can add line breaks. Regardless, wrapping is ultimately done in the browser and doesn't match the source.

ragesoss:
Tweaked format and merged.

PR #1110, opened 2016-12-27
Show organized campaigns on My Dashboard

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T152742

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -111,17 +111,6 @@ class User < ActiveRecord::Base
     course.users.role('student').include? self
   end
 
-  def campaign_organizer?
-    organized_campaigns.any?
-  end
-
-  def organized_campaigns
-    CampaignsUsers.where(
-      user_id: id,
-      role: CampaignsUsers::Roles::ORGANIZER_ROLE
-    ).collect(&:campaign).compact
-  end
-
   def role(course)
     # If this is a new course, grant permissions.
     return CoursesUsers::Roles::INSTRUCTOR_ROLE if course.nil?
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -49,6 +49,7 @@ class User < ActiveRecord::Base
   has_many :campaigns_users, class_name: CampaignsUsers, dependent: :destroy
   has_many :survey_notifications, through: :courses_users
   has_many :courses, -> { distinct }, through: :courses_users
+  has_many :campaigns, -> { distinct }, through: :campaigns_users
   has_many :revisions, -> { where(system: false) }
   has_many :all_revisions, class_name: Revision
   has_many :articles, -> { distinct }, through: :revisions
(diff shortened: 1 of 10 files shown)


app/models/user.rb
```diff
@@ -110,6 +111,17 @@ def student?(course)
     course.users.role('student').include? self
   end
 
+  def campaign_organizer?
+    organized_campaigns.any?
+  end
+
+  def organized_campaigns
```
ragesoss: Why not just use `.campaigns` in place of this? We don't have any other roles at this point, and if/when we do, we may still want to show all a user's campaigns.

ragesoss:
@sejalkhatri would you like to review this?

app/models/user.rb
MusikAnimal: So should this be moved to DashboardPresenter? There we want to show only campaigns the current user organizes. I just thought it was a nifty method that could exist on the model

ragesoss: My point is that, until/if we have another role, user.organized_campaigns will be the same as user.campaigns. So why not just use user.campaigns in the view or presenter or wherever, and dispense with this method entirely?

sejalkhatri:
Ya sure, m on it!

app/views/campaigns/_campaigns.html.haml
```diff
@@ -47,34 +47,7 @@
...
-              = campaign_presenter.word_count
-          %td
-            %span.views-human
-              = number_to_human(campaign_presenter.courses.sum(:view_sum))
-            %span.views.hidden
-              = campaign_presenter.courses.sum(:view_sum)
-          %td
-            %span.students
-              = campaign_presenter.campaign.students_without_nonstudents.count
+        = render 'campaigns/row', presenter: campaign_presenter, campaign: campaign
```
sejalkhatri: If this is the only place we are using this view then shouldn't we put the entire loop(lines 48/49) in that?

MusikAnimal: It is also used in [dashboard/index.html.haml](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1110/files#diff-6ba95d8ef933598e22ac06d1e009b57eR135). The convention of having `_row.html.haml` just be the `<tr>` markup itself is modeled after the [courses/row](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/views/courses/_row.html.haml) partial

sejalkhatri: ohk! that makes sense.

spec/features/dashboard_spec.rb
```diff
@@ -146,4 +146,27 @@
       expect(page).to have_content 'Recent Title'
     end
   end
+
+  context 'campaigns', focus: true do
```
sejalkhatri: focus: true -- should be removed I guess

MusikAnimal: whoops! Thank you :)

sejalkhatri:
everything else looks good to me!

MusikAnimal:
Thanks!

PR #1112, opened 2016-12-28
Added new 'StudentList'' spec test

feelfreelinux:
Added new 'StudentList'' spec test


ragesoss:
I got it working by adding a few tweaks. One key is to make sure you set up all the global patches before rendering the component. 75a4b781ce547a174ef1ac13e253ad049963657b

feelfreelinux:
@ragesoss Thanks! This `global.confirm` fix is the key.

PR #1114, opened 2016-12-28
Remove programs from the Campaign:programs tab

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T152748

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -106,7 +106,7 @@ class CampaignsController < ApplicationController
     campaigns_course.destroy if campaigns_course
     flash[:notice] = t('campaign.course_removed', title: params[:course_title],
                                                   campaign_title: @campaign.title)
-    redirect_to overview_campaign_path(@campaign.slug)
+    redirect_to programs_campaign_path(@campaign.slug)
   end
 
   #######################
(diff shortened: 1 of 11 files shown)


app/controllers/campaigns_controller.rb
```diff
@@ -101,6 +101,14 @@ def remove_organizer
     redirect_to overview_campaign_path(@campaign.slug)
   end
 
+  def remove_course
+    campaigns_course = CampaignsCourses.find_by(course_id: params[:id], campaign_id: @campaign.id)
```
ragesoss: `:id` is an ambiguous name for this param. Let's use course_id instead.

app/controllers/campaigns_controller.rb
```diff
@@ -101,6 +101,14 @@ def remove_organizer
...
+    campaigns_course.destroy if campaigns_course
```
ragesoss: If we aren't going to handle the case of failure — which will mainly happen when you try to remove a course that already got removed since the page was loaded — then it's better to let this raise a 500 instead of incorrectly flashing the 'removed' notice.

Alternatively, you could do something like this:
```ruby
message = campaign_course&.destroy ? 'campaign.course_removed' : 'campaign.course_already_removed'
flash[:notice] = t(message, …)
```

ragesoss:
This looks good and works nicely. Ready to merge after addressing the minor issues above.

PR #1116, opened 2016-12-29
Add Contribution Statistics to the Profile Pages

sejalkhatri:


ragesoss:
UnsubmittedCoursesController needs to be updated to use courses_list instead of the 'none' campaign.

I think now is also a good time to change the method signature of `CoursesPresenter#initialize` to use keyword arguments (which means changing each place that calls `CoursesPresenter.new` to use the keyword arguments).

app/views/users/_profile_cumulativestats.html.haml
```diff
@@ -0,0 +1,46 @@
+%h3
```
ragesoss: Use underscores to separate words in the file name.

app/views/users/_profile_cumulativestats.html.haml
```diff
@@ -0,0 +1,46 @@
...
+  = t("users.contribution_statistics")
+%br/
+%h5
+  = t("users.user_impact")
+  #{@user.username}
```
ragesoss: Instead of handling this separately, you should pass the username to the message function. Doing it this way is called a 'lego' message (because it builds the whole from parts), and it's bad for internationalization because not all languages will put the parts in the same order.

app/views/users/_profile_cumulativestats.html.haml
```diff
@@ -0,0 +1,46 @@
...
+        %h4
+          = number_to_human presenter.trained_percent
+          \%
+        %p= t("users.up_to_date_with_training")
+%br/
+%br/
+%h5
+  = t("users.user_impact")
+  #{@user.username}'s
+  = t("courses.students")
```
ragesoss: Same here. Use a single message that takes the username as a parameter.

app/views/users/_profile_cumulativestats.html.haml
```diff
@@ -0,0 +1,46 @@
...
+    .stat-display__stat.tooltip-trigger
+      .stat-display__value= presenter.user_count
+      %small= t("#{presenter.course_string_prefix}.students")
+      .tooltip.dark
+        %h4
+          = number_to_human presenter.trained_percent
+          \%
+        %p= t("users.up_to_date_with_training")
+%br/
+%br/
```
ragesoss: better to use css within the users module than handle spacing with empty elements and breaks.

Issue #1115, opened 2016-12-28
Unsubmitted Courses view should show more relevant information


ragesoss:
These current columns are not relevant for unsubmitted courses: 
* Recent Edits
* Words Added
* Views
* Editors

These other data should be added instead:
* Username of first instructor
* list of tags for the course (`course.tags`)
* Expected students

PR #1118, opened 2016-12-29
Unsubmitted data clean-up

feelfreelinux:
Fixes: issue #1115

### DIFF
--- a/app/views/unsubmitted_courses/index.html.haml
+++ b/app/views/unsubmitted_courses/index.html.haml
@@ -21,9 +21,9 @@
             = t("courses.creation_date")
             %span.sortable-indicator
           %th.sort.sortable{"data-default-order" => "desc", "data-sort" => "instructor", :style => "width: 165px;"}
-             = t("courses.instructor.one")
+            = t("courses.instructor.one")
           %th.sort.sortable{"data-default-order" => "desc", "data-sort" => "expected_students", :style => "width: 200px;"}
-             = t("courses.expected_students")
+            = t("courses.expected_students")
           %th.sort.sortable{"data-default-order" => "desc", "data-sort" => "tags", :style => "width: 125px;"}
             = t("metrics.tags")
             %span.sortable-indicator
--- a/app/views/unsubmitted_courses/index.html.haml
+++ b/app/views/unsubmitted_courses/index.html.haml
@@ -10,9 +10,6 @@
       .sort-select
         %select.sorts{:rel => "courses"}
           %option{:rel => "asc", :value => "title"}= t("courses.title")
-          %option{:rel => "desc", :value => "instructor"}= t("courses.instructor.one")
-          %option{:rel => "desc", :value => "expected_students"}= t("courses.expected_students")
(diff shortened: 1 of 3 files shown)


app/views/unsubmitted_courses/index.html.haml
```diff
@@ -10,14 +10,10 @@
       .sort-select
         %select.sorts{:rel => "courses"}
           %option{:rel => "asc", :value => "title"}= t("courses.title")
-          %option{:rel => "desc", :value => "revisions"}= t("metrics.revisions")
-          %option{:rel => "desc", :value => "characters"}= t("metrics.word_count")
-          %option{:rel => "desc", :value => "average-words"}= t("metrics.word_count_average")
-          %option{:rel => "desc", :value => "views"}= t("metrics.view")
-          %option{:rel => "desc", :value => "students"}= t("users.editors")
+          %option{:rel => "desc", :value => "instructor"}= t("courses.instructor.one")
+          %option{:rel => "desc", :value => "tags"}= t("metrics.tags")
```
mdew192837: Why does tags go before expected students here and vice versa above? Should it be consistent?

config/locales/en.yml
```diff
@@ -297,6 +297,7 @@ en:
       user_exists: "Sorry, %{username} is not an existing user."
       invalid_language_or_project: Invalid language/project
     expected_students: Expected Students
+    expected_students_doc: Number of expected students count
```
mdew192837: This is redundant. Remove the count at the end or change to something like 'Expected students count'

config/locales/en.yml
```diff
@@ -334,7 +335,7 @@ en:
       your timeline, submit it for approval by Wiki Ed staff. Once approved, you
       will be given an enrollment URL that students can use to join the course.
       (You'll still be able to make edits later.)
-    revisions_doc: The number of edits in the past %{timeframe} days
+    instructor_doc: The name of first instructor
```
mdew192837: probably should be a 'the' in front of 'first'

config/locales/en.yml
```diff
@@ -458,8 +459,10 @@ en:
     char: Chars Added
     date_time: Date/Time
     edit_count_description: Total Edits
+    expected_students: Expected students
```
mdew192837: This message is already on line 299. Is there a reason to add it again?

feelfreelinux:
image

config/locales/en.yml
```diff
@@ -334,7 +335,7 @@ en:
       your timeline, submit it for approval by Wiki Ed staff. Once approved, you
       will be given an enrollment URL that students can use to join the course.
       (You'll still be able to make edits later.)
-    revisions_doc: The number of edits in the past %{timeframe} days
```
ragesoss: This is used elsewhere. Do not delete it.

feelfreelinux:
Looks like tests need fix. Unrelated.

ragesoss:
Good start!

A few problems:

* Sorting by expected students, instructor, and tag do not work. The sorting code for these should just be removed; it's not important to be able to sort by them.
* The mouseover docs for instructors and expected students are not helpful. They should be removed.

PR #1121, opened 2017-01-06
Update setup instructions for Windows

markspolakovs:
See #1097.

PR #1122, opened 2017-01-06
Make work on Windows

### LINKED ISSUE #1097: Setting up a development environment in Windows is not documented

markspolakovs:
This makes the setup work on Windows. Changes:

 - Windows does not include time zone information, so added gem `tzinfo-data`
 - `zeus` is not supported on Windows
 - In the Raven initializer, `/dev/null` is replaced with `nul` on Windows

Together with #1121, fixes #1097.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -53,6 +53,9 @@ gem 'paper_trail'
 
 gem "paperclip"
 
+# TZ information is not available on Windows, needs to be installed separately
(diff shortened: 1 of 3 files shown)


ragesoss:
I can't confirm that it's working on Windows, but I tested this out in my environment and it doesn't break anything.

PR #1129, opened 2017-01-09
Improve styling of user profile pages & add student contribution Statistics.

sejalkhatri:


app/models/user.rb
```diff
@@ -115,6 +115,10 @@ def student?(course)
     course.users.role('student').include? self
   end
 
+  def isstudent?
+    courses.any? { |course| student?(course) }
+  end
+
```
ragesoss: This method and the `isinstructor?` method are going to be easy to confuse with the prior `student?` and `instructor?` methods.

Perhaps instead, `student_role?` and `instructor_role?`. Or maybe `course_student?` and `course_instructor?`.

app/presenters/courses_presenter.rb
```diff
@@ -60,6 +60,10 @@ def word_count
     WordCount.from_characters courses.sum(:character_sum)
   end
 
+  def student_word_count
+    WordCount.from_characters CoursesUsers.where('user_id = ?', current_user).sum(:character_sum_ms)
```
ragesoss: For all these new individual methods, they are based on `current_user`. How will that work for viewing another user's profile? Maybe we should instead pass an additional optional argument, `user:`, to represent the user whose courses are being presented. And then these methods might be called `individual_word_count` and so on, perhaps.

app/controllers/users_controller.rb
```diff
@@ -57,7 +57,11 @@ def show
     @user = User.find_by_username(username)
     if @user
       @courses_user = @user.courses_users
```
ragesoss: `@courses_users`

app/presenters/courses_presenter.rb
```diff
@@ -87,5 +107,29 @@ def user_count
...
+  def individual_article_views
+    article_views = 0
+    courses.each do |c|
+      article_views += c.all_revisions.where(user_id: @user&.id).sum(:views)
+    end
+    article_views
+  end
+
+  def individual_article_created
+    article_created = 0
```
ragesoss: articles_created (plural)

app/presenters/courses_presenter.rb
```diff
@@ -87,5 +107,29 @@ def user_count
     courses.sum(:user_count)
   end
 
+  def individual_article_count
+    article_count = 0
+    courses.each do |c|
+      article_count += c.all_revisions.where(user_id: @user&.id).distinct.count(:article_id)
```
ragesoss: `article_ids = c.all_revisions.where(user_id: @user.id).pluck(:article_id).uniq`

app/presenters/courses_presenter.rb
```diff
@@ -87,5 +107,29 @@ def user_count
     courses.sum(:user_count)
   end
 
+  def individual_article_count
+    article_count = 0
```
ragesoss: `article_ids`

app/presenters/courses_presenter.rb
```diff
@@ -87,5 +107,29 @@ def user_count
...
+    courses.each do |c|
+      article_count += c.all_revisions.where(user_id: @user&.id).distinct.count(:article_id)
+    end
+    article_count
```
ragesoss: `article_ids.uniq.count`

app/presenters/users_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+require "#{Rails.root}/lib/word_count"
+
+#= Presenter for users view
+class UsersPresenter
+  def initialize(user:)
+    @user = user
+  end
+
+  def individual_courses
+    @user.courses.where('courses_users.role = 0')
```
ragesoss: Instead of the "magic number" 0 here, you should use the constant, `CoursesUsers::Roles::STUDENT_ROLE`.

app/presenters/users_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+    individual_courses.each do |c|
+      upload_usage_count += c.uploads.where(user_id: @user.id).sum(:usage_count)
+    end
+    upload_usage_count
+  end
+
+  def individual_article_count
+    article_ids = 0
+    individual_courses.each do |c|
+      article_ids += c.all_revisions.where(user_id: @user.id).pluck(:article_id).uniq.count
```
ragesoss: if this is going to collect article_ids, and not just count them, then you don't need `count`. If you want this method to count them (as the current name indicates) then you should count them at the end, after collecting and deduplicating them.

app/presenters/users_presenter.rb
```diff
@@ -48,17 +48,18 @@ def individual_article_views
...
       individual_articles.each do |a|
-        article_views += a.revisions.where(user_id: @user.id).order('date ASC').first.views
+        earliest_revision = a.revisions.where(user_id: @user.id).order('date ASC').first
+        article_views += earliest_revision.views if earliest_revision.present?
       end
     end
     article_views
   end
 
   def individual_articles_created
```
ragesoss: You can use a similar strategy here as in the article count to avoid overcounting new articles that occur during multiple overlapping courses: in the .each loop, collect the new article revisions, then deduplicate and count them at the end.

ragesoss:
It occurs to me that some scope like like `user.course_revisions`, which would provide all the revisions a user made during courses where they were a student, would make all this a lot easier. Then we wouldn't have to iterate over courses and worry about duplication. I'm not sure how feasible it is to make a scope like that, though. I'll look into it.

sejalkhatri:
Yeh! That would reduce the complexity. I'll also give it a try.

app/controllers/users_controller.rb
```diff
@@ -56,9 +56,10 @@ def show
     username = params[:username].tr('_', ' ')
     @user = User.find_by_username(username)
     if @user
-      @courses_user = @user.courses_users
-      @courses_list = @user.courses.where('courses_users.role = 1')
-      @presenter = CoursesPresenter.new(current_user: current_user, courses_list: @courses_list)
+      @courses_users = @user.courses_users
+      @courses_list = @user.courses.where('courses_users.role = ?', CoursesUsers::Roles::INSTRUCTOR_ROLE)
+      @courses_presenter = CoursesPresenter.new(current_user: current_user, courses_list: @courses_list)
+      @users_presenter = UsersPresenter.new(user: @user)
```
ragesoss: The name `UsersPresenter` doesn't do much clarify the specific role of this presenter. Maybe `IndividualContributionsPresenter`?

sejalkhatri: IndividualStatisticsPresenter?

ragesoss: +1

spec/presenters/individual_statistics_presenter_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+describe IndividualStatisticsPresenter do
+  describe 'articles view count ' do
+    let(:course1) { create(:course) }
+    let(:course2) { create(:course) }
+    let(:user) { create(:user) }
+    let(:article) { create(:article) }
+    subject { described_class.new(user: user) }
+
+    context 'Select unique articles' do
```
ragesoss: The conventional way to name the context block is so that it combine with the rest of the spec to form something like a complete sentence, like "IndividualStatisticsPresenter articles_view_count, when a user is in two courses that overlap, doesn't double count the same article in multiple courses".

(If you create `.rspec` in the project root directory and add formatting options, you'll see each of these messages as the tests run. Mine looks like this:
```
--color
--format documentation
```
)

app/presenters/individual_statistics_presenter.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+require "#{Rails.root}/lib/word_count"
+
+#= Presenter for users view
```
ragesoss: A more precise description would be helpful here. The key point is that this class is intended to provide statistics about a user's contributions *during courses in which the user was a student*.

spec/presenters/individual_statistics_presenter_spec.rb
sejalkhatri: Just tried it out, looks great! Thanks:)

Issue #1023, opened 2016-11-09
Overview tab of course should be renamed Home


ragesoss:
The term 'overview' can be a little confusing in the context of a course... some users think it means something closer to timeline, a detailed overview of the course plan/content.

PR #1132, opened 2017-01-14
Change overview text to home

hin101:
Fixes: #1023 

Will need to update translations in other languages


ragesoss:
This commit removes example `.yml` files, and should not.

For a full fix of #1023, this should also update the routing, so that the url becomes `:course_slug/home` rather than `:course_slug/overview`. The caveat there is that `/overview` should still work, so that urls pointing to overview pages still work.

ragesoss:
@hin101 That works nicely! The last bits will be to change the link target and associated variable names, in course.jsx starting around line 230, and to add a comment in the router about why the 'overview' variant remains and should not be removed (to preserve URLs from when that was the default).

app/assets/javascripts/components/course.jsx
```diff
@@ -229,7 +229,7 @@ const Course = React.createClass({
 
     let overviewLinkClassName;
```
ragesoss: This name should be changed to 'home' too, since it matches the `homeLink`.

dump.rdb
```diff
@@ -0,0 +1 @@
+REDIS0006�ܳC�Z��V
```
ragesoss: Stray file. You could add dump.rdb to the git ignore file, if you like, or just remove it from the commit.

ragesoss:
@hin101 the translations will gradually get updated; they come from (volunteers, usually active Wikipedians) at translatewiki.net.

PR #1146, opened 2017-01-19
Display instructor as well as student stats in User Profile Page when user plays both the roles.

sejalkhatri:
Also improving the profile page styling.

### DIFF
--- a/app/assets/stylesheets/modules/_stats.styl
+++ b/app/assets/stylesheets/modules/_stats.styl
@@ -38,11 +38,16 @@
     &:before
       left 50%
 
-.instructor-profile-stats
-  display block
-  margin-bottom 70px
+.user_stats
   .stat-display
-    justify-content initial
+    margin 0 0 30px 0
 
-    .stat-display__stat
-      margin 0 20px 0 20px
+  #instructor-profile-stats
+    display block
+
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/modules/_stats.styl
```diff
@@ -38,11 +38,16 @@
...
-    justify-content initial
+    margin 0 0 30px 0
 
-    .stat-display__stat
-      margin 0 20px 0 20px
+  #instructor-profile-stats
+    display block
+
+    .stat-display
+      justify-content initial
```
ragesoss: What does this do?

sejalkhatri: Handles positioning of the items, here it is used to avoid the elements to space out in the horizontal space. So here I am overwriting the css properties of .stat-display for #user-prfile-stats block as there are only 2 elements in that row

app/assets/stylesheets/modules/_users.styl
```diff
@@ -7,7 +7,7 @@
 	h3,h5
 		color #808080
 
-#coursedetails #display
```
ragesoss: This should be called course-details to be easier to read quickly.

PR #1154, opened 2017-01-26
Add revision(s) info to DiffViewer modal (num chars, dates)

patmbolger:
Waffle story: https://waffle.io/WikiEducationFoundation/WikiEduDashboard/cards/585826f063d5f70e00c0e78b

For single revision:
- Num characters changed
- Date of edit

For range of revisions:
- Span of time over which revisions were made

### DIFF
--- a/app/assets/javascripts/components/activity/activity_table_row.jsx
+++ b/app/assets/javascripts/components/activity/activity_table_row.jsx
@@ -65,7 +65,9 @@ const ActivityTableRow = React.createClass({
 
     let diffViewer;
     if (this.props.revision && this.props.revision.api_url) {
-      diffViewer = <DiffViewer revision={this.props.revision} />;
+      diffViewer = <DiffViewer revision={this.props.revision}
+                               charCount={this.props.revision.characters}
+                   />;
     }
 
     return (
--- a/app/assets/javascripts/components/activity/activity_table_row.jsx
+++ b/app/assets/javascripts/components/activity/activity_table_row.jsx
@@ -65,9 +65,7 @@ const ActivityTableRow = React.createClass({
 
     let diffViewer;
     if (this.props.revision && this.props.revision.api_url) {
-      diffViewer = <DiffViewer revision={this.props.revision}
-                               charCount={this.props.revision.characters}
-                   />;
+      diffViewer = <DiffViewer revision={this.props.revision} />;
     }
 
     return (
--- a/app/assets/javascripts/components/revisions/diff_viewer.jsx
+++ b/app/assets/javascripts/components/revisions/diff_viewer.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import OnClickOutside from 'react-onclickoutside';
 
 const DiffViewer = React.createClass({
-  // displayName: 'DiffViewer',
+  displayName: 'DiffViewer',
 
   // Diff viewer takes a main (final) revision, and optionally a first revision.
   // If a first revision is supplied, it fetches a diff from the parent of the
@@ -20,7 +20,6 @@ const DiffViewer = React.createClass({
   getInitialState() {
     return {
       showDiff: false,
-      chars: this.props.charCount
     };
   },
 
@@ -116,13 +115,19 @@ const DiffViewer = React.createClass({
         dataType: 'jsonp',
         url: diffUrl,
         success: (data) => {
-          const revisionData = data.query.pages[this.props.revision.mw_page_id].revisions;
+          const firstRevisionData = data.query.pages[this.props.revision.mw_page_id]
+                                      .revisions[0];
+          let lastRevisionData = null;
+          if (data.query.pages[this.props.revision.mw_page_id].revisions.length === 2) {
+            lastRevisionData = data.query.pages[this.props.revision.mw_page_id]
+                                      .revisions[1];
+          }
           this.setState({
-            diff: revisionData[0].diff['*'],
-            comment: revisionData[0].comment,
+            diff: firstRevisionData.diff['*'],
+            comment: firstRevisionData.comment,
             fetched: true,
-            firstRevDateTime: revisionData[0].timestamp,
-            lastRevDateTime: revisionData.length === 2 ? revisionData[1].timestamp : null
+            firstRevDateTime: firstRevisionData.timestamp,
+            lastRevDateTime: lastRevisionData ? lastRevisionData.timestamp : null
           });
         }
       });
@@ -160,7 +165,11 @@ const DiffViewer = React.createClass({
 
     let diffComment;
     let revisionDateTime;
-    let firstRevTime, lastRevTime, days, hours, minutes;
+    let firstRevTime;
+    let lastRevTime;
+    let days;
+    let hours;
+    let minutes;
     let timeSpan = '';
 
     // Edit summary for a single revision:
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/components/activity/activity_table_row.jsx
```diff
@@ -65,7 +65,9 @@ const ActivityTableRow = React.createClass({
 
     let diffViewer;
     if (this.props.revision && this.props.revision.api_url) {
-      diffViewer = <DiffViewer revision={this.props.revision} />;
+      diffViewer = <DiffViewer revision={this.props.revision}
```
ragesoss: the `revision` object already includes the character count, so there's no need to pass it as a separate prop. Within the DiffViewer, you can do `this.props.revision.characters`.

patmbolger: For a single revision yes, but I did not see that available for rev / first rev usage - so I added the prop and used it uniformly throughout (even though for single rev I could get it from the revision itself).

(please check me on that - I may have missed something re two-revision usage)

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -115,11 +116,13 @@ const DiffViewer = React.createClass({
         dataType: 'jsonp',
         url: diffUrl,
         success: (data) => {
-          const revisionData = data.query.pages[this.props.revision.mw_page_id].revisions[0];
+          const revisionData = data.query.pages[this.props.revision.mw_page_id].revisions;
```
ragesoss: This and the following setState values would be clearer if you separately set up `revisionData` (which corresponds to `this.prop.revision`) and `firstRevisionData`.

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -19,7 +19,8 @@ const DiffViewer = React.createClass({
 
   getInitialState() {
     return {
-      showDiff: false
+      showDiff: false,
+      chars: this.props.charCount
```
ragesoss: This character count is not being used yet? In any case, if you're pulling it from props, then there's no need to copy it to state. You can just do `this.props.revision.characters` whenever you want to access that value.

app/assets/javascripts/components/activity/activity_table_row.jsx
ragesoss: Ah, yes. In the cumulative diff usage (when it is provided both a revision and a first revision), the revision does not include the character count. In this context, it is showing all the changes to an article over a large span of time, so we don't want to show the character count of that last revision. If we were to show anything, it would be the cumulative change in characters across the diff range displayed, which we'd need to get from the Wikipedia API rather than passing it in as props.

In any case, we should be able to do this without adding a new required prop to the component.

patmbolger: I am showing the cum count for the range of revs.  I could calculate the cum char change by looking at the `size` attribute of each rev - but since that is already calculated by an API call, and is available in the parent component, I thought it would be better to just add the props var.  

Or, if the story does not require it, we could just show the char count only for the single-rev case.

Thoughts?

ragesoss: I think only showing it for the single-rev case is good for the time-being.

arreche:
Travis CI has [some complains](https://travis-ci.org/WikiEducationFoundation/WikiEduDashboard/jobs/195659739) about the code style ;)

I think you can lint the code locally by running `gulp lintjs` but It may be convenient using `.editorconfig` and `.eslintrc` in your editor.

patmbolger:
@arreche - yes, I've addressed those warnings in the file that I changed in this PR.

The other files it is complaining about have been removed from the PR (that is, those are unchanged from what is in production).

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -116,13 +115,19 @@ const DiffViewer = React.createClass({
         dataType: 'jsonp',
         url: diffUrl,
         success: (data) => {
-          const revisionData = data.query.pages[this.props.revision.mw_page_id].revisions;
+          const firstRevisionData = data.query.pages[this.props.revision.mw_page_id]
+                                      .revisions[0];
+          let lastRevisionData = null;
+          if (data.query.pages[this.props.revision.mw_page_id].revisions.length === 2) {
```
arreche: I would assign this condition to a variable with a meaningful name for readability. ;)

i.e.
```javascript
hasRevisions = 'Something hard to read or to understand by humans'
if(hasRevisions)
  doSomething 
```

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -156,11 +164,43 @@ const DiffViewer = React.createClass({
...
+        timeSpan = `${days} day${days > 1 ? 's' : ''}, `;
+      }
+      if (hours !== 0) {
+        timeSpan += `${hours} hour${hours > 1 ? 's' : ''}, `;
+      }
+      if (minutes !== 0) {
+        timeSpan += `${minutes} minute${minutes > 1 ? 's' : ''}`;
+      }
+      if (!timeSpan) {
+        timeSpan = 'less than a minute';
```
arreche: I think we should use `I18n.t` here.

app/assets/javascripts/components/revisions/diff_viewer.jsx
patmbolger: I simplified the assignment of the variable.

app/assets/javascripts/components/revisions/diff_viewer.jsx
patmbolger: I'm not sure of the extent to which text should be localized or not. 

@ragesoss - could you explain?

Also, if so, perhaps this should be a separate PR?

ragesoss: I haven't closely reviewed this part of the code or what you are trying to do — it looks like you're recreating something along the lines of the Rails 'distance between time in words' or whatever it's called — but this, I think we should start by just displaying the times of the first and last revisions.

In general, we do want to use the i18n framework for any new user-facing text that is introduced.

patmbolger: Yes, that's exactly what I was doing.  There's probably a package for this(?).  In any case, this is how I interpreted this from the story:```Time range for cumulative diff```

I will change to just show the start and end date-times.

Re I18n - do I need to just add something to the .en file?

ragesoss: When you add an interface message, you should add it to both the en.yml and qqq.yml files. The latter is the documentation file, where you describe how the message is used within the interface (so that translators can understand the context, which is often necessary for correctly translating it, especially from a language as flexible as English).

For something like what you were trying to do, we would definitely want to use an internationalized library. (Perhaps moment.js, which we already have in the app, does this?). Rolling it by hand makes it very problematic for i18n, because it creates so-called 'lego' messages, where the whole message is built up from smaller parts. Variations in syntax from one language to another means that lego messages run into trouble when the word / phrase order needs to be switched to be correct in another language, or when gendered languages need to rely on one part of the message to get the right gender for another part.

arreche: I think you are suggesting to use the [Time from now](https://momentjs.com/docs/#/displaying/fromnow/) function in moment.js

patmbolger: Kind of like that .... but I don't want to be limited to ```now``` as the reference time, since the span I was looking for was the duration between datetime1 and datetime2.  

What I needed was the JS equivalent to: http://api.rubyonrails.org/classes/ActionView/Helpers/DateHelper.html#method-i-distance_of_time_in_words

But, it is moot since we'll just use timestamps for the editing period end points.

Thanks for the explanation explanation of hot the i18n should be done.

ragesoss: Moment can handle that with 'Time from X', which does it using any two moment objects.

patmbolger: Ah yes - if I used "time from X" without the suffix that would do it.  Moment is amazingly powerful.

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -156,11 +161,27 @@ const DiffViewer = React.createClass({
...
+    let timeSpan;
+
+    // Edit summary for a single revision:
+    //  > Edit date and number of characters added
+    // Edit summary for range of revisions:
+    //  > First and last times for edits to article (from first applicable rev to last)
     if (!this.props.first_revision) {
-      diffComment = <p className="diff-comment">{this.state.comment}</p>;
+      revisionDateTime = moment(this.props.revision.date).format('YYYY/MM/DD h:mm a');
+      diffComment = <p className="diff-comment">{this.state.comment}&nbsp;
```
ragesoss: This is looking pretty clean now! I'll give it a spin as soon as I can. After this change, `diffComment` is doing a more than the name implies. It was named that because it just showed the comment (aka, the edit summary) left along with the diff.

ragesoss:
@patmbolger thanks! This is working nicely, and nicely implemented.

I think putting all that information into a single line — the actual diff comment, plus the time info, plus the edit size, using parentheses an semicolons — makes it hard to make sense of. Both in the UI and the code, all that info is being added to the comment, rather than displayed along with it. Maybe just a different paragraph for each bit of info would clear things up, until I can get some design attention for it.

patmbolger:
I made a change but I'm not sure I am getting at what you are describing.  If this does not do it, lets discuss after scrum on Monday.

ragesoss:
@patmbolger you got it. Looks a bit better now. Fix the eslint error and I'll merge it. Still needs some design improvement to put these three pieces of data into a user-friendly format, but we can do that later.

Issue #234, opened 2015-07-29
Notices (like the submit message) are easy to overlook


ragesoss:
Several instructors have not readily noticed the instructions in the notices that appear below the nav section of course pages.

In the case where a user makes changes to the timeline and then saves using the bottom save button, they do not get returned to the top of the page. This can lead to a sense of being lost, in terms of next steps.

In the longer term, we probably want to have a more sophisticated notification system. One option in the short term (which I'm not convinced is the right approach, but it might be) would be to make the notifications stick at the top of the screen (once they get that high), like the login/home bar.

This isn't a high priority right now, I just want to capture it as an issue.

rileywho:
/* Note: This solution is still being thoroughly vetted but I wanted to share where I'm at.*/
Changes:
+ Remove course data summary if the course hasn't yet been approved
+ Green bar details:
- Total height: 80px
- Color: #BBDB94
- Text: Roboto, Normal, #000, font size: 17px
- Button: background #fff, color: #656e92

![screen shot 2015-08-04 at 5 49 57 pm](https://cloud.githubusercontent.com/assets/2120466/9075992/5884a878-3ad1-11e5-874f-6fee00cfd875.png)

ragesoss:
I like this. It will work for the submit notice and the enrollment URL
notice, since in both cases there won't be any stats yet.

It won't work for other (not yet existing) notices that appear when a
course is already underway.

ragesoss:
This looks great! I think it doesn't do enough to prevent the message being overlooked, though; I really like the idea of sticky header on scroll up as the next iteration. Not a priority for the time being, though.

ragesoss:
Currently, we have the `Notifications` component as the first thing inside `app.jsx`, while the sticky nav header of course page is implement via `Affix` in `course.jsx`

On possible approach would be to shift the Notifications to live at the end of the nav, inside the `Affix` in `course.jsx`, and then add `Notifications` in the other higher level components are not part of the `Course` hierarchy in in the router (`router.jsx`).

Alternatively, the `Notifications` component in `app.jsx` could also be affixed, and then the one in `course.jsx` could be adjusted to play nicely with it.

ragesoss:
* Move Notifications to course.jsx, in the nav section
* Add Notifications component to other routes besides Course
* Ensure that messages appear as expected on screen when screen is scrolled down (and they would be above the fold in the old design)

FedericoEsparza:
[...]

PR #1156, opened 2017-01-28
Display course update scheduled message in a notification component

### LINKED ISSUE #1045: 'needs update' action should give info about what to expect
UX feedback: 
> In the interface, it is not clear how to retrieve data from past programs. I
> created a new program for the Edit-a-thon on Nov 2, and clicked on "Schedule
> data update" and the message I got is "update scheduled" but it doesn't say
> anything about when the data will be available, or that you already
> scheduled an update when you click the button the second time. In the demo,
> you mentioned that within 24 hours organizers could retrieve the data. Can
> we add this in the message that pops up?

We can solve this by creating a notification with a relevant explanation.

The trigger for this is in `app/assets/javascripts/components/overview/available_actions.jsx`.

An example of creating a notification via `NotificationActions.addNotification` can be found in `app/assets/javascripts/components/students/enroll_button.jsx`.

arun1595:
Fixes #1045

### DIFF
--- a/app/assets/javascripts/stores/notification_store.js
+++ b/app/assets/javascripts/stores/notification_store.js
@@ -87,6 +87,15 @@ const NotificationStore = Flux.createStore(storeMethods, (payload) => {
       handleErrorNotification(data, payload.actionType);
       NotificationStore.emitChange();
       break;
+    case 'NEEDS_UPDATE':
+      addNotification({
+        message: payload.data.result,
+        closable: true,
+        type: 'success'
+      });
+
+      NotificationStore.emitChange();
+      break;
     default:
       // no default
   }
--- a/app/assets/javascripts/utils/api.js
+++ b/app/assets/javascripts/utils/api.js
@@ -505,13 +505,12 @@ slide_id=${opts.slide_id}`,
         type: 'GET',
         url: `/courses/${courseId}/needs_update.json`,
         success(data) {
-          alert(data.result);
           return res(data);
         }
       })
       .fail((obj) => {
         console.error('Couldn\'t request update');
-        return rej(obj);
+        rej(obj);
       })
     );
   },
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -112,7 +112,8 @@ class CoursesController < ApplicationController
   def needs_update
     @course = find_course_by_slug(params[:id])
     @course.update_attribute(:needs_update, true)
-    render json: { result: 'Update scheduled.' }, status: :ok
+    render json: { result: I18n.t('courses.creator.update_scheduled') },
+           status: :ok
   end
 
   def notify_untrained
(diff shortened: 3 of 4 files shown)


app/assets/javascripts/utils/api.js
```diff
@@ -1,4 +1,5 @@
 import { capitalize } from './strings';
+import NotificationActions from '../actions/notification_actions.js';
```
ragesoss: If we're going to couple this to the response from the server — which is a good idea, although not what I was imagining — then we should not doing this in API.js. As much as possible, API.js should be strictly about querying th endpoint and returning the data.

Adding the Notification could be done by adding handling for the NEEDS_UPDATE with the notification store. ServerActions.needsUpdate() already sends out the 'NEEDS_UPDATE' signal when this action succeeds. Following the patterns of this app would mean adding the 'NEEDS_UPDATE' case to the NotificationStore.

arun1595: > As much as possible, API.js should be strictly about querying th endpoint and returning the data

Agree. That is why I was initially working on adding this to `available_actions.jsx` where it make it sense to include a notification component.

> Following the patterns of this app would mean adding the 'NEEDS_UPDATE' case to the NotificationStore.

Ah! Makes sense. Let me make the changes and update the PR.

arun1595:
@ragesoss The PR is ready for review. The animation is needed for UX, without it, the user will have no idea about the notification message until he scrolls all the way to the top.

app/assets/javascripts/utils/api.js
```diff
@@ -503,15 +503,12 @@ slide_id=${opts.slide_id}`,
...
       $.ajax({
         type: 'GET',
-        url: `/courses/${courseId}/needs_update.json`,
-        success(data) {
-          alert(data.result);
-          return res(data);
-        }
+        url: `/courses/${courseId}/needs_update.json`
       })
+      .done(obj => res(obj))
```
ragesoss: You can do this in the 'success' call to match the pattern through most of api.js.

app/assets/javascripts/stores/notification_store.js
```diff
@@ -87,6 +87,17 @@ const NotificationStore = Flux.createStore(storeMethods, (payload) => {
...
+      addNotification({
+        message: payload.data.result,
+        closable: true,
+        type: 'success'
+      });
+
+      $('html, body').animate({scrollTop: 0}, 200)
+        .promise()
+        .then(() => NotificationStore.emitChange());
+      break;
```
ragesoss: Although this makes sense on it's own in the current UI, this should be addressed by fixing the general problem of notifications appearing when the visible page doesn't include the Notifications section: #234. I think we'll work on a fix for that one soon, so let's cut out the animation here.

arun1595: Okay.

app/assets/javascripts/utils/api.js
arun1595: Done.

PR #1158, opened 2017-01-30
Add article size graph

ragesoss:


sejalkhatri:
Thanks Sage! I'll update the patch accordingly.

ragesoss:
This works nicely. I think we can get rid of the remaining flickering on graph load by waiting to render the initial graph until state.graphData is available. Could use the `<Loading />` component instead.

Issue #1068, opened 2016-12-14
Admins should be able to mark an Alert as 'resolved'


ragesoss:
We have assorted types of Alert records that get created in different situations. For some, like deletion alerts and 'continued activity' alerts, it would be helpful if admins could indicate that the issue has been dealt with, and then the system would create a new alert if the alert conditions come up again. For example, if an article is nominated for deletion, then that gets resolved and it is not deleted, but later it gets renominated, the system would again create a new alert about that. Or if course dates are extended to include student activity after the initial end date, but then there is even more activity after the new end date, the system would issue another alert.

MrWilhelm:
If I am not wrong, reason for alerts not triggering for the second time, this line on ContinuedCourseActivityAlertManager;

```next if Alert.exists?(course_id: course.id, type: 'ContinuedCourseActivityAlert')```

I think we can add another alert manager to lib/alerts which checks for existing alerts and deletes if there is any. And then we can add a callback to Course model after updating course date which calls existing alert. We can also use this manager to expire Alerts when needed like when article resolved.

 @ragesoss

ragesoss:
@MrWilhelm you're correct about where the system looks to avoid triggering an alert twice. We don't want to delete old alerts, though. I think we'd instead want to do some thing like add a `resolved` column, defaulting to `false`, to the Alert records. Then we'd do something like add `resolved: false` to that `Alert.exists?` call. (It also needs a UI to mark an alert resolved.)

If you're interested in working on this, I suggest starting with the case of `ArticlesForDeletionAlert`.

MrWilhelm:
So is this how I should do it;

- Create migration to add alerts resolved boolean column default false
- Add resolve action for alerts controller to update resolved column
- Add button to resolve alerts.

Can we add something like '?resolved=true' parameter to the end of the link sent with email, then we can resolve alert on page load automatically?

PR #1159, opened 2017-01-31
Admins can resolve alerts

MrWilhelm:
I've initiated this PR for #1068 

@ragesoss


spec/controllers/alerts_controller_spec.rb
```diff
@@ -61,4 +61,22 @@
...
+    let(:alert) { create(:alert) }
+    let(:user) { create(:user) }
+
+    before do
+      allow(controller).to receive(:current_user).and_return(user)
+      allow(controller).to receive(:user_signed_in?).and_return(true)
+      # controller.instance_variable_set(:@alert)
+    end
+
+    it 'should update Alert resolved column to true' do
```
ragesoss: Resolving an alert should only be allowed if the user is an admin.

spec/controllers/alerts_list_controller_spec.rb
```diff
@@ -7,14 +7,27 @@
...
       before { allow(controller).to receive(:current_user).and_return(admin) }
+
       it 'renders the alerts list' do
         get :index
+
         expect(response.status).to eq(200)
         expect(response.body).to have_content(alert.type)
       end
+
+      it 'should not render resolved alerts' do
```
ragesoss: Rather than hiding resolved alerts altogether, it would be better to still show them, but indicate that they are resolved.

app/controllers/alerts_controller.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 class AlertsController < ApplicationController
   before_action :require_signed_in
+  before_action :set_alert, only: [:resolve]
 
   def create
```
ragesoss: Now that this AlertsController is doing more than just creating NeedHelpAlerts, it needs a bit of documentation and/or method renaming to clarify that `create` is specifically for creating a `NeedHelpAlert`, while other alert types are created via the update cycle instead of direct action by a user.

app/controllers/alerts_controller.rb
```diff
@@ -19,8 +22,12 @@ def create
     end
   end
 
+  # Resolves alert if it is resolvable? Requires admin permission.
+  # Normally, same alerts won't be created for the second time.
+  # Resolving alert, allows it to be created for the second time if conditions are met.
   def resolve
-    ensure_alerts_are_enabled { return}
+    ensure_alerts_are_enabled { return }
```
ragesoss: This isn't necessary, as this 'alerts are enabled' check is specifically about whether the 'get help button' that allows any user to create a 'NeedHelpAlert' is enabled. Even if that is disabled, other alerts are being created and may be resolved.

app/models/alert.rb
```diff
@@ -77,6 +77,12 @@ def email_target_user
   def emails_disabled?
     ENV["#{self.class}_emails_disabled"] == 'true'
   end
+
+  def resolvable?
```
ragesoss: Instead of type checking in `Alert`, this should return false, and 'true' version (or `!resolved`) should be in the `ArticlesForDeletionAlert` subclass to override it.

spec/models/alert_spec.rb
```diff
@@ -24,6 +25,8 @@
   let(:course) { create(:course) }
   let(:revision) { create(:revision) }
   let(:user) { create(:user) }
+  let(:alert) { create(:alert) }
```
ragesoss: one confusing thing here is that the test of `.resolvable?` below is implicitly relying on the default 'alert' factory to return an ArticlesForDeletionAlert. You should specify the type and the `resolved` value explicitly here, so that it's more clear exactly what you're testing.

MrWilhelm: In alert factory, it was defined so(ArticlesForDeletionAlert) with type, but resolved? function doesn't override resolvable?, so I changed class name with this. 

I'm adding type here explicitly to make it more clear.

MrWilhelm:
What do you think about adding a callback to course model which works only on update and if end date of the course changed to sometime in the future?

We can expire last alert (Or alert meets some condition), if there is any.

ragesoss:
This looks great. I think we should avoid a callback like that; I'm not eager to add extra ActiveRecord queries to something as central to the system as course updates.

But I think we can do the exact same thing with ContinuedCourseActivityAlert. The workflow is very much the same; the person who gets the alert manually takes a look at the course and chooses whether to extend the course dates to include the continued activity, and with resolvability, they could choose whether to reset the alert system to alert them again if there is further activity.

I'm going to go ahead and merge this now. If you want to take the ContinuedCourseActivityAlert change, open a new PR for it. Should be really easy to add that on.

Issue #1013, opened 2016-10-27
Process slow requests involving external services asyncronously via sidekiq


ragesoss:
We have sidekiq workers implemented for a few WikiEdits actions to process wiki edits asyncronously. We can expand this to more cases to further improve performance.

The initial implementation introduced two sidekiq worker classes for two triggering to kinds of WikiEdits actions: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/29552181d90bada9055afe94966dba6825a2715d

Here are other actions where we can apply the same pattern: instead of invoking WikiEdits or another slow process directly, create a new worker class to do that action via sidekiq, and replace the call to it in a controller with a call to the worker.

To get started on this, you'll need to set up a development environment. Follow the setup instructions here, and let us know if you run into any problems: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/setup.md

Each of these should be a separate task with its own pull request:

- [x] [IN PROGRESS] Emailing target user from AlertsController#create
- [x] WikiCourseEdits#remove_assignment, from AssignmentsController#destroy
- [x] Course deletion WikiCourseEdits, from CoursesController#delete
- [x] Notify untrained users, from CourseController#notify_untrained
- [x] course submission emails, from CoursesController#handle_course_announcement
- [x] WikiCourseEdits#announce_course, from CoursesController#handle_course_announcement
- [x] enrollment edits and preference setting, from SelfEnrollmentController
- [x] enrollment edits, from UsersController
- [x] assignment removal, from UsersController#remove_assignment_templates

ragesoss:
For reference, see:

* https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/workers/update_course_worker.rb
* https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/29552181d90bada9055afe94966dba6825a2715d

feelfreelinux:
I'll make assignment removal

feelfreelinux:
I made assignment removal

FedericoEsparza:
https://youtu.be/d-8MafzIxIU?t=26m50s Some discussion in the 2017-02-21 AV Americas Scrum recording.

PR #1161, opened 2017-02-01
Refactor notify untrained users into asynch sidekiq job per #1013

mtc2013:
Refactor notify untrained users into asynch sidekiq job per #1013


app/controllers/courses_controller.rb
```diff
@@ -117,7 +117,7 @@ def needs_update
 
   def notify_untrained
     @course = find_course_by_slug(params[:id])
-    WikiEdits.new(@course.home_wiki).notify_untrained(@course, current_user)
+    NotifyUntrainedUsersWorker.schedule_notifications(course: @course, notifying_user: current_user)
```
ragesoss: require the worker code at the top.

Issue #1164, opened 2017-02-02
Admins should be able to search for users by email


ragesoss:

MrWilhelm:
Should we create another admin page like feedbacks, alerts etc., which lists and search users? Or should we include it anywhere else?

Also, do we need something like typeahead(https://twitter.github.io/typeahead.js/) or just basic search with a button?

ragesoss:
@MrWilhelm I think a `/users` index page would make a lot of sense for this (added the the admin nav like the others).

Typeahead would be nice if it's easy to add, but just a basic search with a button would be fine.

Another nice-to-have feature would be a way to get a list of all the instructors in the system.

MrWilhelm:
I have started tests and some implementation and send it soon.

However I'm not sure about scope of the search. I did something like this, but depends on it, we can create separate query builder on lib, but for now it is simple class method on model. Query is something like:

```ruby
User.('lower(email) like ?', 'search_term%')
```

ragesoss:
Sounds good to me.

ragesoss:
Implemented in #1170

Issue #1141, opened 2017-01-19
Campaign course data CSVs should include number of users who are newbies


ragesoss:
Per Vojtěch Dostál, a useful bit of data to add to the dashboard would be the number of participants who are newly registered — that is, their wiki accounts were created after the start date. To do this, we'd need to:

* Get account creation date during user creation process (`UserImporter`), store it in the User model.
* Calculate the newly registered count during course cache update, store it in the Course model.
* Add this data to the campaign csv output.

ragesoss:
To get the account creation date, we can use the MediaWiki 'list users' API and request the 'registration' property: https://en.wikipedia.org/w/api.php?action=help&modules=query%2Busers

There's an example query of that 'list' API for user data in `WikiApi`.

ragesoss:
Done with #1290

PR #1165, opened 2017-02-02
Continued Course Activity Resolvable

MrWilhelm:
@ragesoss Here is the one for ContinuedCourseActivityAlert

### DIFF
--- a/spec/models/alert_spec.rb
+++ b/spec/models/alert_spec.rb
@@ -59,9 +59,24 @@ describe Alert do
       end
     end
 
-    it 'should be resolvable only for ArticlesForDeletionAlert' do
-      expect(alert.resolvable?).to be(true)
-      expect(active_course_alert.resolvable?).to be(false)
+    it 'should be resolvable for resolvable alert types' do
+      Alert::RESOLVABLE_ALERT_TYPES.each do |type|
+        # Equals to ArticlesForDeletionAlert.new
+        alert = type.constantize.new
+
+        expect(alert.resolvable?).to be(true)
+      end
+    end
+
+    it 'should not be resolvable for certain types' do
+      unresolvable_alert_types =
+        Alert::ALERT_TYPES - Alert::RESOLVABLE_ALERT_TYPES
+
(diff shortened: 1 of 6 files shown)


ragesoss:
Lovely! Code looks great. Assuming tests pass, I'll merge it shortly.

spec/models/alert_spec.rb
```diff
@@ -59,9 +59,24 @@
       end
     end
 
-    it 'should be resolvable only for ArticlesForDeletionAlert' do
-      expect(alert.resolvable?).to be(true)
-      expect(active_course_alert.resolvable?).to be(false)
+    it 'should be resolvable for resolvable alert types' do
+      Alert::RESOLVABLE_ALERT_TYPES.each do |type|
```
ragesoss: This is nice.

MrWilhelm:
Do you have any suggestion for the next issue I will work on? @ragesoss

ragesoss:
@MrWilhelm This one might be a good fit for you, on the strictly Ruby side: #1141

I can provide additional details on how to get necessary data. I'll start with a few pointers on that issue, and you can give me a ping if I don't say enough.

Another option, if you're up for some javascript/css, is #234 

And since you seem to know your way around Rails quite well, this one might be pretty easy for you: #1164 

I've never implemented search before, so I'm not really sure where to start.

ragesoss:
@MrWilhelm one more option: you could do a code review of either of the larger work-in-progress PRs I have open: #1167 and #1152

Since you seem very comfortable with Rails conventions, you'd probably be able to spot ways to improve the code.

MrWilhelm:
I think I want to get #1164 . 

I'll also check out #1167 If there is anything I can help.

I'll usually go with backend stuff since I feel more comfortable with it and easier for me to get into.

@ragesoss

ragesoss:
@MrWilhelm great! Ping in those places if you have any questions.

spec/models/alert_spec.rb
MrWilhelm: Thanks!

PR #1170, opened 2017-02-07
Users search for admins

MrWilhelm:
Here is minimal implementation,

I will add role-based search soon, but first I will look if I can convert user_permissions to an enum. Many things would be easier with enum.

For example:
``` user.admin? ```

``` User.instructor.where(....) ```

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -7,9 +7,12 @@ require "#{Rails.root}/app/workers/update_course_worker"
 #= Controller for user functionality
 class UsersController < ApplicationController
   respond_to :html, :json
-  before_action :require_participating_user, only: [:enroll]
 
+  before_action :require_participating_user, only: [:enroll]
   before_action :require_signed_in, only: [:update_locale]
+  before_action :check_user_auth, only: [:index]
+
+  layout 'admin', only: [:index]
 
   def signout
     if current_user.nil?
@@ -46,6 +49,17 @@ class UsersController < ApplicationController
     elsif request.delete?
       remove
     end
+  end
+
+  ####################################################
+  # User listing page for Admins                     #
+  ####################################################
+  def index
+    @users = if params[:term].present?
+               User.search(params[:term])
+             else
+               User.all
+             end
   end
 
   ####################################################
@@ -147,6 +161,12 @@ class UsersController < ApplicationController
     username = enroll_params[:username]
     @user = User.find_by(username: username)
     @user = UserImporter.new_from_username(username) if @user.nil?
+  end
+
+  def check_user_auth
+    return if current_user&.admin?
+    flash[:notice] = "You don't have access to that page."
+    redirect_to root_path
   end
 
   def enroll_params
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: app/controllers/users_controller.rb [hunk 1]
```diff
@@ -7,9 +7,12 @@
 #= Controller for user functionality
 class UsersController < ApplicationController
   respond_to :html, :json
-  before_action :require_participating_user, only: [:enroll]
 
+  before_action :require_participating_user, only: [:enroll]
   before_action :require_signed_in, only: [:update_locale]
+  before_action :check_user_auth, only: [:index]
```
ragesoss: You can use `require_admin_permissions`, defined in ApplicationController, instead.

app/controllers/users_controller.rb
```diff
@@ -49,6 +52,17 @@ def enroll
   end
 
   ####################################################
+  # User listing page for Admins                     #
+  ####################################################
+  def index
+    @users = if params[:term].present?
+               User.search(params[:term])
+             else
+               User.all
```
ragesoss: This is slow and not super useful. for the default, I suggest either `User.none` or a query for the last 20 users where permissions equal `User::Permissions::INSTRUCTOR`.

app/controllers/users_controller.rb
```diff
@@ -49,6 +52,17 @@ def enroll
   end
 
   ####################################################
+  # User listing page for Admins                     #
+  ####################################################
+  def index
+    @users = if params[:term].present?
```
ragesoss: Calling this parameter `term` is confusing, in part because it's vague and in part because the Course model has a `term` attribute that means something different. In the current implementation, it's specifically an email search, so maybe call it `email`?

app/models/user.rb
```diff
@@ -73,6 +73,13 @@ module Permissions
   scope :ungreeted, -> { where(greeted: false) }
 
   ####################
+  # Class method(s)  #
+  ####################
+  def self.search(term)
```
ragesoss: This method name doesn't communicate the specific kind of search (email). I suggest either naming it `search_by_email`, or adding a second parameter — `field` or similar` that has a default value of `:email`, and then write the query so that it works with any field.

Being able to also search by `real_name` would be a very useful thing to add (not necessarily in this PR).

app/views/users/index.html.haml
```diff
@@ -0,0 +1,22 @@
...
+  %hr
+
+  %h1 Users
+  %span= params[:term].present? ? "Results for: \"#{params[:term]}\"" : nil
+
+  %table.table.table--hoverable
+    %thead
+      %tr
+        %th Username
+        %th E-mail
```
ragesoss: Showing the `real_name` field as well would be very useful.

app/views/users/index.html.haml
```diff
@@ -0,0 +1,22 @@
...
+    %tbody
+    - @users.each do |user|
+      %tr
+        %td= user.username
```
ragesoss: The username should link to the profile for that user: `/users/:username`.

MrWilhelm:
@ragesoss All should be resolved. I'm thinking about to create separate PR for search when I have more time.

I'm thinking about the write a query builder, if params(whatever we name it) have '@' sign, builds a query only for email colon and otherwise searches for real_name.

ragesoss:
@MrWilhelm That sounds like a great next step.

PR #1174, opened 2017-02-14
course campaign select list ordered by create time, waffle #1142

patmbolger:
On the course page a course can be associated with a campaign.  This story consists of:

1) Present the list of campaigns in time-sorted order, most recent at top of list
2) Add test expectations for this behavior

### DIFF
--- a/spec/features/admin_role_spec.rb
+++ b/spec/features/admin_role_spec.rb
@@ -42,9 +42,11 @@ describe 'Admin users', type: :feature, js: true do
            course_id: 10002,
            role: 1)
 
-    create(:campaign,
-           id: 1,
-           title: 'Fall 2015')
+    create(:campaign, id: 1, title: 'Fall 2015',
+                      created_at: Time.now + 2.minutes)
+    create(:campaign, id: 2, title: 'Spring 2016',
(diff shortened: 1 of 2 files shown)


ragesoss:
LGTM! I'll merge it as soon as the tests pass.

spec/features/admin_role_spec.rb
```diff
@@ -70,7 +72,14 @@
 
       # Edit details and add campaign
       click_button('Edit Details')
+
       page.all('.button.border.plus')[4].click
+
+      # Ensure campaigns appear in select list ordered by time (descending)
+      campaign_options = all('select[name=campaign]>option')[1,2]
```
sejalkhatri: can u explain this line, just for info!

ragesoss: `all('select[name=campaign]>option')` returns an array of the options in that select element.

`array[1,2]`returns a portion of the array, starting from index 1, with length 2. https://docs.ruby-lang.org/en/2.3.0/Array.html#method-i-5B-5D

sejalkhatri: Thanks @ragesoss  :+1:

PR #1192, opened 2017-02-22
Polish UI for ended courses

### LINKED ISSUE #1149: It should be hard to miss if a course is ended when a user tries to join it
If a course has already ended, users will not be able to join. The fact that it has ended should be clear to users *before* they try to create accounts and/or click 'Join'. One scenario this should help fix is if the dates of an edit-a-thon were entered incorrectly; participants should be able to easily notice the problem and tell the organizers.

We can probably do this via changes to the `EnrollCard` component.

arreche:
Fix #1149

### DIFF
--- a/app/assets/javascripts/components/enroll/enroll_card.jsx
+++ b/app/assets/javascripts/components/enroll/enroll_card.jsx
@@ -15,7 +15,13 @@ const EnrollCard = React.createClass({
 
   render() {
     let messageBody;
-    if (this.props.enrolledParam !== undefined) {
+    if (this.props.course.ended) {
+      messageBody = (
+        <div>
+          <h1>{I18n.t('courses.ended')}</h1>
+        </div>
+      );
+    } else if (this.props.enrolledParam !== undefined) {
       // Enrollment is complete
       if (this.props.enrolledParam === 'true') {
         messageBody = (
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/enroll/enroll_card.jsx
```diff
@@ -35,13 +35,21 @@ const EnrollCard = React.createClass({
...
     // User is logged in and ready to enroll
     } else if (this.props.user.id && this.props.userRole === -1) {
-      messageBody = (
-        <div>
-          <h1>{I18n.t('courses.join_prompt', { title: this.props.course.title || '' })}</h1>
-          <a className="button dark" href={this.props.course.enroll_url + this.props.passcode}>{I18n.t('courses.join')}</a>
-          <a className="button border" href={this.props.courseLink}>{I18n.t('application.cancel')}</a>
-        </div>
-      );
+      if (this.props.course.ended) {
```
ragesoss: This should be higher up in this 'if' chain. We want to show the 'course ended' message whether or not the user is logged in.

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -120,11 +120,11 @@ const AvailableActions = React.createClass({
       }
     // If user has no role or is logged out
     } else {
```
ragesoss: This could change from an `else` to an `else if`, and then it won't add an extra level of nesting.

PR #1196, opened 2017-02-24
Add User Profile Form, Improve Profile UI and resolve #1181.

### LINKED ISSUE #1181: JavaScript error on many pages caused by user profile jquery
When I load course pages and some other non-profile pages, I get the following error:

`TypeError: $(...).offset(...) is undefined`

I think this is related to the user profile jquery plugin.

sejalkhatri:


config/routes.rb
```diff
@@ -19,6 +19,12 @@
                           as: :true_destroy_user_session
   end
 
+  #UserProfilesController
+  controller :user_profiles do
+    get 'user_profiles/:username' => 'user_profiles#show' , constraints: { username: /.*/ }
```
ragesoss: We want to keep the same route `/users/:username`. Otherwise, it will break links and cause confusion.

config/routes.rb
```diff
@@ -19,6 +19,12 @@
...
+    post 'user_profiles/update/:username' => 'user_profiles#update'
```
ragesoss: I suggest keeping the routes all under `/users/`

db/schema.rb
```diff
@@ -353,6 +353,11 @@
     t.datetime "completed_at"
   end
 
+  create_table "user_profiles", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
ragesoss: This change should also update the migration version timestamp at the top of the file.

spec/factories/user_profiles.rb
```diff
@@ -0,0 +1,5 @@
+FactoryGirl.define do
+  factory :user_profile do
+    bio "MyString"
```
ragesoss: Since `bio` is not a required field, it may be better to leave it out of the factory definition (so that it would default to `nil`) and including it via `create(:user_profile, bio: 'foo')` when an initial value is needed.

ragesoss:
edit mode

I find the design here confusing: the white-background disclaimer looks like an editable text field, while the actual edit field has a purple background and does not at first glance seem like it changed into an editable field.

I suggest coloring the disclaimer background, possibly with the 'Warning' color used in tables: https://dashboard.wikiedu.org/styleguide#tables

And I suggest switching to a white background for the profile in edit mode.

I think the username also needs more space from the top of the purple box.

db/schema.rb
sejalkhatri: It didn't get updated automatically, is there a command we have to run for this?

app/controllers/user_profiles_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  def destroy
+    @user_profile.destroy
+  end
+
+  private
+
+    def require_write_permissions
```
ragesoss: These don't need to be indented further after the `private` keyword.

sejalkhatri: oops! I'll correct that right away!

app/controllers/user_profiles_controller.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+class UserProfilesController < ApplicationController
+  before_action :set_user, only: [:show, :update, :destroy]
```
arun1595: This line could be written as `before_action :set_user` as every action in the controller needs a `user` object to work with. Even in the long run, if you add any new actions to this controller, those will always require a `user` object. So, IMO, it is totally fine having

```ruby
before_action :set_user
```

Less keystrokes :wink:

app/models/user.rb
```diff
@@ -57,6 +57,7 @@ module Permissions
   has_many :assignments
   has_many :uploads, class_name: CommonsUpload
   has_many :training_modules_users, class_name: 'TrainingModulesUsers'
+  has_one :user_profile, class_name: UserProfile, dependent: :destroy
```
arun1595: I think the `class_name` option is superfluous. Wouldn't rails automatically infer the model name from the association?

From rails docs,

> If the name of the other model cannot be derived from the association name, you can use the :class_name option to supply the model name.

app/controllers/user_profiles_controller.rb
sejalkhatri: yes! that will be efficient! Thanks @arun1595

spec/factories/user_profiles.rb
```diff
@@ -0,0 +1,4 @@
+FactoryGirl.define do
+  factory :user_profile do
```
arun1595: Shouldn't the factory definition require a `user`? Since we already have a `user` factory, you could add something like the following:

```ruby
FactoryGirl.define do
  factory :user_profile do
    user
  end
end
```

What do you think @sejalkhatri ?

sejalkhatri: I think Its better if the user creates that in the test itself(controller spec file) as the test can vary for different kinds of user.

ragesoss: @sejalkhatri I agree that, at this point, keeping the factory independent is best. If it becomes an issue with further development of the related tests, we can reconsider it.

ragesoss:
This looks good. I'm deploying it to staging just to double-check in a live environment and make sure the migration goes smoothly. Will merge and deploy to production after that. Great work!

sejalkhatri:
Thank you :)

ragesoss:
@sejalkhatri I can't type in the bio field.

It looks like this, but clicking on the text area does not add the cursor or let me fill it in.

edit view

sejalkhatri:
ohk! It works fine for me! but I think I know what the problem is, its css related. I'll update it

PR #1199, opened 2017-02-27
Make Profile Section in the profile pages responsive and add image field to the user form

sejalkhatri:
Using Paperclip file uploader gem for uploading user images.

user_profile_update1

user_profile_update2


ragesoss:
Where did the icons come from?

db/schema.rb
```diff
@@ -356,8 +356,14 @@
...
-    t.string  "bio"
-    t.integer "user_id"
+    t.string   "bio"
+    t.integer  "user_id"
+    t.string   "image_file_name"
+    t.string   "image_content_type"
+    t.integer  "image_file_size"
+    t.datetime "image_updated_at"
+    t.string   "location"
+    t.string   "work_status"
```
ragesoss: We should call this field 'institution' rather than 'work status'.

ragesoss:
I'll add a tweak to the image styling.

PR #1218, opened 2017-03-04
Change the order of Courses displayed in Course Details, latest to oldest and some other tweaks to contrbution stats rendering in user profile pages

sejalkhatri:


app/views/user_profiles/show.html.haml
```diff
@@ -25,7 +25,7 @@
           %h3
             = t("courses.course_details")
           %br/
-          - @courses_users.each do |c|
+          - @courses_users.reverse_each do |c|
```
ragesoss: I've never seen that method before. o_O

PR #1221, opened 2017-03-04
Implement Contribution Statistics section in React using Flux Architecture.

sejalkhatri:


app/controllers/user_profiles_controller.rb
```diff
@@ -25,6 +27,15 @@ def update
     end
   end
 
+  def student_stats_data
```
ragesoss: I suggest using a single endpoint for all the user profile stats. You can structure the jbuilder view to separate student vs instructor stats, if you need to. But just having one API call and one path through the React data flow will make this easier to work with and reason about.

ragesoss:
I think you've got the right idea with it so far.

You'll need to create something like `ProfileActions.fetch_stats(userId)` that gets called when the stats component mounts, which in turn calls a new ajax method in api.js to get the json data, and then handle receiving that data in the profile store.

You'll need a `getState()` function in the stats component to fetch data from the profile store, and it'd also be good to have the store implement `.getLoadingStatus()` so that you can render a loading spinner before the data arrives. `OverviewHandler` is one example you can follow here.

app/controllers/user_profiles_controller.rb
sejalkhatri: okay!

app/assets/javascripts/components/user_profiles/instructor_stats.jsx
```diff
@@ -0,0 +1,27 @@
...
+// import ByStudents from './by_students'
+// import StudentStats from './student_stats.jsx';
+
+const InstructorStats = React.createClass({
+  render() {
+    return (
+      <div>
+        <h5> InstructorStats by </h5>
+        <div className= "user_stats">
+          <div className= "stat-display">
```
arun1595: At first sight, it looks like the `InstructorStats` and `StudentStats` component have similar UI. Why don't we create a functional component encompassing the `.user_stats` div?

arun1595: And if the `InstructorStats` and `StudentStats` components look dumb to me. You could make them functional components as well.

@sejalkhatri @ragesoss What do you guys think?

sejalkhatri: elements are yet to be added! It's not complete yet

app/assets/javascripts/actions/profile_actions.js
```diff
@@ -0,0 +1,13 @@
+import McFly from 'mcfly';
+const Flux = new McFly();
+import API from '../utils/api.js';
+
+const ProfileActions = Flux.createActions({
+  fetch_stats(username) {
+    return API.fetchStatsData(username)
```
ragesoss: This is not a specific enough function name in `API`. It needs to clearly indicate that it's fetching user profile stats.

app/assets/javascripts/components/user_profiles/by_students.jsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+
+const ByStudents = React.createClass({
```
ragesoss: I think this component's name should include 'stats', like `ByStudentsStats`, to more clearly indicate what it's for.

app/assets/javascripts/components/user_profiles/contribution_stats.jsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import InstructorStats from './instructor_stats.jsx';
+import StudentStats from './student_stats.jsx';
+import ProfileStore from '../../stores/profile_store.js';
+import ProfileActions from '../../actions/profile_actions.js';
+import Loading from '../common/loading.jsx';
+
+const getState = function () {
+  const isstudent = $('#react_root').data('isstudent');
```
ragesoss: Use camelCase for all these javascript identifiers.

app/assets/javascripts/components/user_profiles/by_students.jsx
```diff
@@ -0,0 +1,74 @@
...
+  render() {
+    return (
+      <div>
+        <h5>
+          {I18n.t('user_profiles.instructors_student_impact', { username: this.props.username })}
+        </h5>
+        <div className= "stat-display">
+          <div className= "stat-display__stat">
+            <div className="stat-display__value">
+              {this.props.stats.word_count}
```
ragesoss: 'stats.word_count' is not specific enough. It took me a while to figure out from the code whether this is 'by students' word count or 'as student' word count. Maybe in the jbuilder and here, it could be called 'word_count_by_students' or something else that clearly identifies which word count it is.

app/views/user_profiles/stats_data.json.jbuilder
```diff
@@ -0,0 +1,25 @@
...
+if @user.course_student?
+  json.course_string_prefix @individual_stats_presenter.course_string_prefix
+  json.individual_courses_count @individual_stats_presenter.individual_courses.count
+  json.individual_word_count number_to_human @individual_stats_presenter.individual_word_count
+  json.individual_article_views number_to_human @individual_stats_presenter.individual_article_views
+  json.individual_article_count number_to_human @individual_stats_presenter.individual_article_count
+  json.individual_articles_created number_to_human @individual_stats_presenter.individual_articles_created
+  json.individual_upload_count number_to_human @individual_stats_presenter.individual_upload_count
+  json.individual_upload_usage_count number_to_human @individual_stats_presenter.individual_upload_usage_count
+end
```
ragesoss: I think it might make sense to add some structure to this entire json output, to address the ambiguous names of the instructor stats. Instead of a flat json object, something more like:

```
{
  by_students: {},
  as_instructor: {},
  as_student: {}
}
```

That would make it much clearer which stats are part of which displayed component and which section of the UI.

config/routes.rb
```diff
@@ -22,6 +22,7 @@
   #UserProfilesController
   controller :user_profiles do
     get 'users/:username' => 'user_profiles#show' , constraints: { username: /.*/ }
+    get 'user_stats' => 'user_profiles#stats_data'
```
ragesoss: Maybe just call this controller method `#stats` instead of `stats_data`.

app/controllers/user_profiles_controller.rb
```diff
@@ -25,8 +27,10 @@ def update
     end
   end
 
-  def destroy
-    @user_profile.destroy
+  def stats_data
+    @individual_stats_presenter = IndividualStatisticsPresenter.new(user: @user)
+    @courses_list = @user.courses.where('courses_users.role = ?', CoursesUsers::Roles::INSTRUCTOR_ROLE)
+    @courses_presenter = CoursesPresenter.new(current_user: current_user, courses_list: @courses_list)
```
ragesoss: Hmm... with this change, it means that we have to do the pretty slow queries of `CoursesPresenter` twice; once for the html view, and another for the json stats data.

I think that's okay for now, but eventually we'll want to move the courses list to React as well.

app/assets/javascripts/components/user_profiles/by_students_stats.jsx
```diff
@@ -0,0 +1,74 @@
...
+  render() {
+    return (
+      <div>
+        <h5>
+          {I18n.t('user_profiles.instructors_student_impact', { username: this.props.username })}
+        </h5>
+        <div className= "stat-display">
+          <div className= "stat-display__stat">
+            <div className="stat-display__value">
+              {this.props.stats.word_count}
```
sejalkhatri: Should I still change the word_count keyword to word_count_by_students?( i.e after making the changes in the stats json view)

ragesoss: No, changing the view should cover it nicely.

PR #1223, opened 2017-03-05
Sidekiq worker for WikiCourseEdits.announce_course edits

keer25:
WikiEducationFoundation/WikiEduDashboard#1013


ragesoss:
Nice work @keer25!

PR #1224, opened 2017-03-06
added campaign institution/rows column

### LINKED ISSUE #1140: Campaign course list should include institution/term for each course

mz99:
Added to programs table the additional field "Institution/Term". Added in new partial _courses_row to campaign table without affecting other contexts

Resolves #1140

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -357,6 +357,7 @@ en:
     revisions_doc: The number of edits in the past %{timeframe} days
     revisions_none: This course has no recent Wikipedia editing activity.
     school: School
+    school_and_term: Institution/Term
     start: Start
     start_typing: Start typing
     students: Students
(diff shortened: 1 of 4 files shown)


sejalkhatri:
Everything looks good to me.

config/locales/en.yml
```diff
@@ -450,6 +451,7 @@ en:
     published: "This project has been published! Editors may enroll by visiting the following URL:"
     revisions_none: This project has no recent editing activity.
     school: Institution
+    school_and_term: Institution/When
```
sejalkhatri: /When doesn't convey the message for me

ragesoss: It's not great, but it matches the term we use for the 'term' field for courses that aren't actually classroom/academic projects.

sejalkhatri: Cool!

PR #1228, opened 2017-03-11
Sidekiq workers for slow processes in SelfEnrollmentController

keer25:
Implemented sidekiq workers for enrollment edits and preference setting, from SelfEnrollmentController #1013


ragesoss:
@keer25 thanks! I'll take a look at these soon. You should try a different issue next. Each of the items here are very similar, and each one makes good first-time contributions.

Issue #1197, opened 2017-02-26
Too long assignment titles cause MySQL errors


ragesoss:
Putting in a very long assignment title results in an error in MySQL. Too-long titles should be caught and handled earlier in the data flow, with a sensible error message presented to the user.

ragesoss:
@keer25 You asked:

> It will be helpful if someone can suggest where the the create and update actions of the AssignmentsController are triggered

These get called via the javascript frontend. The starting point is the AssignButton component: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/components/students/assign_button.jsx#L284

Submitting the input for a new assignment triggers the `assign` function of that component, which does some validation, then triggers a confirmation dialog, which triggers the `onConfirm` function, which triggers `ServerActions.addAssignment`, which triggers `API.createAssignment`, which does a POST to `/assignments.json`, which gets routed (via `routes.rb`) to the create method of AssignmentsController.

sejalkhatri:
Control Flow: 
/assets/javascripts/components/students/assign_button.jsx handles the Assign button submit action and calls assign() method onSubmit which updates the store and calls for a server action.

From /javascripts/utils/api.js createAssignment method is called which redirects control to assignments_controller, where lib/assignment_manager.rb is imported and used to update the assignment model.

PR #1229, opened 2017-03-12
Alert message if assignment title too large

keer25:
#1197 If title exceeds 255 characters alert is shown and is not added to databe.

### DIFF
--- a/app/assets/javascripts/components/students/assign_button.jsx
+++ b/app/assets/javascripts/components/students/assign_button.jsx
@@ -109,6 +109,10 @@ const AssignButton = React.createClass({
 
     if (assignment.title === '' || assignment.title === 'undefined') {
       return;
+    } else if (assignment.title.length > 255) {
+      // Title shouldn't exceed 255 chars to prevent mysql errors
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/students/assign_button.jsx
```diff
@@ -109,6 +109,9 @@ const AssignButton = React.createClass({
 
     if (assignment.title === '' || assignment.title === 'undefined') {
       return;
+    } else if (assignment.title.length > 255) {
```
ragesoss: Here it would be helpful to add a comment explaining why we need to limit the title length.

app/assets/javascripts/components/students/assign_button.jsx
```diff
@@ -109,6 +109,9 @@ const AssignButton = React.createClass({
...
+      alert(I18n.t('assignments.title_too_large'));
```
ragesoss: We'll also need a corresponding message string in the locale file (`config/locates/en.yml`, with an description of what the message is for in `qqq.yml`).

Issue #1206, opened 2017-02-28
Close the popup once an available article has been created


arun1595:
## Steps to reproduce:

1. Login as an instructor
2. Create a course and try to add a new available article.

Even after creating the article, the popup remains opened. I guess this could be fixed by removing the `open` class from the popup on clicking the 'Assign' button.

ragesoss:
The best way to do this is to use the `open` function (`this.props.open(e)`) that comes from Expandable and is a prop on `AssignButton`.

PR #1230, opened 2017-03-12
Close the popup once an available article has been created

keer25:
Fix for issue #1206


app/assets/javascripts/components/students/assign_button.jsx
```diff
@@ -123,7 +123,12 @@ const AssignButton = React.createClass({
       return;
     }
 
+    const closePopup = this.props.open;
```
ragesoss: A comment here would be useful, explaining that it should only close the popup if it's the 'add available' version of this button.

app/assets/javascripts/components/students/assign_button.jsx
```diff
@@ -123,7 +123,12 @@ const AssignButton = React.createClass({
...
+    const addAvailable = this.props.add_available;
```
ragesoss: `addAvailable` is being used to determine whether to close the popup on confirmation. Naming it something like `closeOnConfirm` will make this a bit clearer, and will separate the role of the `onConfirm`, and will keep all the references to add_available together in the two setup lines.

PR #1231, opened 2017-03-13
Move enrollment edits to an async process

arun1595:
Fixes part of https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1013

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -96,8 +96,9 @@ class UsersController < ApplicationController
   def make_enrollment_edits
     return unless enroll_params[:role].to_i == CoursesUsers::Roles::STUDENT_ROLE
     # for students only, posts templates to userpage and sandbox
-    WikiCourseEdits.new(action: :enroll_in_course, course: @course,
-                        current_user: current_user, enrolling_user: @user)
+    EnrollInCourseWorker.schedule_edits(course: @course,
+                                        editing_user: current_user,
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -96,8 +96,7 @@ def ensure_enrollment_success
   def make_enrollment_edits
     return unless enroll_params[:role].to_i == CoursesUsers::Roles::STUDENT_ROLE
     # for students only, posts templates to userpage and sandbox
-    WikiCourseEdits.new(action: :enroll_in_course, course: @course,
-                        current_user: current_user, enrolling_user: @user)
+    EnrollmentEditsWorker.edit_enrollment(@course.id, current_user.id, @user.id)
```
ragesoss: This is the same worker that just got created as `EnrollInCourseWorker`. Use that new worker instead of adding a new one.

arun1595: Ah! Let me update the PR :)

app/workers/enroll_in_course_worker.rb
```diff
@@ -2,19 +2,19 @@
 require "#{Rails.root}/lib/wiki_course_edits"
 
 class EnrollInCourseWorker
-	include Sidekiq::Worker
+  include Sidekiq::Worker
```
arun1595: There are no significant changes made to this file, only cosmetic ones. I replaced tabs with two spaces

PR #1233, opened 2017-03-15
Adding validations for blank/whitespace-only values during course creation

medhabansal:


lib/course_creation_manager.rb
```diff
@@ -48,8 +51,9 @@ def set_wiki
   end
 
   def set_slug
-    slug = String.new("#{@course_params[:school]}/#{@course_params[:title]}")
-    slug << "_(#{@course_params[:term]})" unless @course_params[:term].blank?
+    slug = @course_params[:school].blank? ? '' : "#{@course_params[:school]}"
+    slug += "/#{@course_params[:title]}" unless @course_params[:school].blank?
```
sejalkhatri: u want `unless @course_params[:title].blank?` here I think

app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -171,7 +171,7 @@ const CourseCreator = React.createClass({
     // This regex is intended to match ascii word characters, dash,
     // whitespace, comma, apostrophe, and any unicode "letter".
     // Adapted from http://stackoverflow.com/questions/150033/regular-expression-to-match-non-english-characters#comment19644791_150078
-    const courseSlugRegex = /^[\w\-\s,'\u00BF-\u1FFF\u2C00-\uD7FF]+$/;
+    const courseSlugRegex = /^[\w\-\s,'\u00BF-\u1FFF\u2C00-\uD7FF]+[\w\u00BF-\u1FFF\u2C00-\uD7FF][\w\-\s,'\u00BF-\u1FFF\u2C00-\uD7FF]*$/;
```
ragesoss: This needs some explanation in the comments. It's a complex regex, so we need to say clearly what we intend for it to be doing.

ragesoss:
This will need some accompanying tests. The easiest place to add them for the server-side validation will probably be to the existing CoursesController spec — `spec/controllers/courses_controller_spec.rb`. There are already test cases for the other failure modes (invalid wiki and duplicate slug), so you can just add new ones for empty school and empty title.

Another useful improvement that you could include in this PR would be to replace the English error messages in `CourseCreationManager` with i18n message strings, so that the error messages can be translated.

lib/course_creation_manager.rb
medhabansal: @sejalkhatri yes, thanks, I will make the change.

app/assets/javascripts/components/course_creator/course_creator.jsx
medhabansal: Thanks! Adding the comment in my latest commit.

medhabansal:
@ragesoss Will make the appropriate changes and add tests as suggested in your last comment. Also, during functionally testing the course creation flow, I found that it isn't possible to clone a course with a bad slug(missing title/school). Should I address the same here?

ragesoss:
@medhabansal nicely done!

I think it should also include tests for the regex, with some examples of valid and invalid input. The best way to do this is probably to move the regex out of `CourseCreator` and into `CourseUtils` (`app/assets/javascripts/utils/course_utils.js`, with corresponding tests in `test/utils/course_utils.spec.js`). We already do basically the same thing with `CourseDateUtils`, which has a validationRegex for dates and is tested in `course_date_utils.spec.js`.

spec/controllers/courses_controller_spec.rb
```diff
@@ -261,6 +261,38 @@
...
 
+      context 'when blank values are given for course school, title or both' do
+        let(:course_params) do
+          { school: ' ',
+            title: '  ',
+            term: 'Fall 2015',
+            start: '2015-01-05',
+            end: '2015-12-20',
+            language: 'arrr',
+            project: 'wikipirates' }
```
ragesoss: You've copied the language and project from another invalid example, so this is not testing your change. These tests all pass even without your change to CourseCreationManager. Use 'en' and 'wikipedia' instead.

One way to improve these new tests as well as the old ones would be to test for the expected error message. For example, `expect(response.body).to have_content('Blank school/title for course')`.

ragesoss:
@medhabansal you asked:
> Also, during functionally testing the course creation flow, I found that it isn't possible to clone a course with a bad slug(missing title/school). Should I address the same here?

If I understand you correctly, no, you don't need to address that.

I manually fixed the existing course in production that had a bad slug, and after this, it won't be possible to create new ones with bad slugs. We should make sure it isn't possible to create a cloned course with a bad slug, but we don't need to do anything to account for existing courses with bad slugs.

medhabansal:
@ragesoss Ok, so that case need not be handled here. Also, updated the pull request with suggested changes, please check. Thanks!

lib/course_creation_manager.rb
```diff
@@ -63,7 +63,7 @@ def invalid_wiki?
   end
 
   def invalid_slug?
-    @slug.split('/').length <= 1
+    @slug.split('/').reject { |slug_part| slug_part.blank? }.size <= 1
```
ragesoss: What this method is doing and why is not very clear. A comment and/or rewriting the code to make the concept more clear — that to be valid, a course slug must consist of some non-blank text followed by a slash followed by some more non-blank text.

test/utils/course_utils.spec.js
```diff
@@ -32,6 +32,18 @@ describe('courseUtils.generateTempId', () => {
   });
 });
 
+describe('CourseUtils.courseSlugRegex', () => {
+  it('returns a regex that validates course slug', () => {
+    const courseSlugRegex = courseUtils.courseSlugRegex();
+    expect(courseSlugRegex.test(' valid ')).to.eq(true);
```
ragesoss: Using more realistic valid values here would be useful. We're using this to test the 'term', 'school', and 'title' fields from user input, so we should use some of the kinds of real values we typically see, including some non-latin scripts.

You can find plenty of examples here: https://outreachdashboard.wmflabs.org/campaigns/miscellanea/programs

medhabansal:
@ragesoss Made the changes and squashed the commits. Please let me know in case of more changes/improvements. Thanks.

Issue #1195, opened 2017-02-24
Add support for Multilingual Wikisource


ragesoss:
Wikisource has a catch-all multi-lingual version for cases where texts are in a language (or multiple languages) that don't have its own wiki: https://wikisource.org

We'll need to special-case that particular wiki in the `Wiki` model.

PR #1235, opened 2017-03-15
Extend existing validations to allow wikisource

bgreg:
Addresses: https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1195

These changes are 2 fold, adjustments to the validation & creating
methods for wikisource and some refactoring of the specs for speed and
readability.

Wikisource changes

Typically a wiki page has a language support, but in some edge cases a
given page will support many languages. This has led to specific logic
around wikidata and wikisource. They are both multilingual in nature but
wikisource can SOMETIMES have a language assigned depending on what
resource within wikisource is being worked on.

I consider these changes an intermittent step. Long term we want a way
to ask the services what languages they support for a given project and
do validation that way. Until that work is completed, this code will
  solve our current use case.

Refactoring of the spec

In some case, a create statement was not needed, so we can save time by
using FactoryGirl#build. Other places I was able to reduce the lines of
code by using object equality checks instead of going through property
comparisons. I also did my best to add some context descriptions to try
and help the documentation quality a little.

### DIFF
--- a/app/assets/javascripts/utils/course_utils.js
+++ b/app/assets/javascripts/utils/course_utils.js
@@ -58,9 +58,9 @@ const CourseUtils = class {
       };
     }
 
-    const wikiSourceUrlParts = /wikisource\.org\/wiki\/([^#]*)/.exec(articleTitle);
-    if (wikiSourceUrlParts) {
-      const title = decodeURIComponent(wikiSourceUrlParts[1]).replace(/_/g, ' ');
+    const wikisourceUrlParts = /wikisource\.org\/wiki\/([^#]*)/.exec(articleTitle);
+    if (wikisourceUrlParts) {
+      const title = decodeURIComponent(wikisourceUrlParts[1]).replace(/_/g, ' ');
       const project = 'wikisource';
       const language = 'www';
       return {
--- a/app/assets/javascripts/utils/course_utils.js
+++ b/app/assets/javascripts/utils/course_utils.js
@@ -58,6 +58,19 @@ const CourseUtils = class {
       };
     }
 
+    const wikiSourceUrlParts = /wikisource\.org\/wiki\/([^#]*)/.exec(articleTitle);
+    if (wikiSourceUrlParts) {
+      const title = decodeURIComponent(wikiSourceUrlParts[1]).replace(/_/g, ' ');
+      const project = 'wikisource'
+      const language = 'www'
+      return {
+        title,
+        project,
+        language,
+        article_url: articleTitle
+      };
+    }
+
     return {
       title: articleTitleInput,
       project: null,
--- a/app/assets/javascripts/utils/course_utils.js
+++ b/app/assets/javascripts/utils/course_utils.js
@@ -61,8 +61,8 @@ const CourseUtils = class {
     const wikiSourceUrlParts = /wikisource\.org\/wiki\/([^#]*)/.exec(articleTitle);
     if (wikiSourceUrlParts) {
       const title = decodeURIComponent(wikiSourceUrlParts[1]).replace(/_/g, ' ');
-      const project = 'wikisource'
-      const language = 'www'
+      const project = 'wikisource';
+      const language = 'www';
       return {
         title,
         project,
--- a/app/models/wiki.rb
+++ b/app/models/wiki.rb
@@ -48,13 +48,13 @@ class Wiki < ActiveRecord::Base
     sg sgs sh si simple sk sl sm sn so sq sr srn ss st stq su sv sw szl ta te
     tet tg th ti tk tl tn to tpi tr ts tt tum tw ty tyv udm ug uk ur uz ve
     vec vep vi vls vo vro w wa war wikipedia wo wuu xal xh xmf yi yo yue za
-    zea zh zh-cfr zh-classical zh-cn zh-min-nan zh-tw zh-yue zu www
+    zea zh zh-cfr zh-classical zh-cn zh-min-nan zh-tw zh-yue zu
   ).freeze
   validates_inclusion_of :language, in: LANGUAGES + [nil]
 
   MULTILINGUAL_PROJECTS = {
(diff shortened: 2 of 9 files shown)


bgreg:
Start a work in progress PR so I can checkout travis and start to get feedback.

app/models/wiki.rb
```diff
@@ -48,12 +48,13 @@ class Wiki < ActiveRecord::Base
...
     vec vep vi vls vo vro w wa war wikipedia wo wuu xal xh xmf yi yo yue za
-    zea zh zh-cfr zh-classical zh-cn zh-min-nan zh-tw zh-yue zu
+    zea zh zh-cfr zh-classical zh-cn zh-min-nan zh-tw zh-yue zu www
   ).freeze
   validates_inclusion_of :language, in: LANGUAGES + [nil]
 
   MULTILINGUAL_PROJECTS = {
-    'wikidata' => 'www.wikidata.org'
+    'wikidata' => 'www.wikidata.org',
+    'wikisource' => 'www.wikisource.org'
```
ragesoss: The base url of multilingual wikisource is wikisource.org rather than www.wikisource.org. (www will redirect, but for interacting with the API and to avoid unnecessary redirects, we should use simply wikisource.org)

ragesoss: This is in contrast to wikidata.org, which uses www.wikidata.org as its canonical base url.

app/models/wiki.rb
```diff
@@ -73,14 +74,18 @@ def api_url
...
-    # Multilingual projects must have language == nil.
+    # Most multilingual projects must have language == nil,
+    # but wikisource is the exception, it can have nil, or a language.
     # Standard projects must have a language.
-    if MULTILINGUAL_PROJECTS.include?(project)
+    #
+    # TODO: Validate the language/project combination by pinging it's API.
+    if project == "wikidata"
       self.language = nil
+    elsif project == "wikisource"
```
ragesoss: To make the `www` convention consistent, we should probably switch on *that* to set language to nil, and save all multilingual project wikis in the database with `language: nil`.

ragesoss: Note that translating `www` to `nil` here keeps the behavior consistent with the `Wiki` spec we added earlier: 'for projects with a multilingual version in addition to language support'.

What we don't want is two Wiki records that represent the same wiki, for example, one with `www` and one with `nil` for the language. The spec should probably be tweaked to make sure that when you create one, and then `.get_or_create` the other, it returns the first one.

app/models/wiki.rb
```diff
@@ -96,7 +101,7 @@ def self.default_wiki
   end
 
   def self.get_or_create(language:, project:)
-    language = nil if MULTILINGUAL_PROJECTS.include?(project)
+    language = nil if project == 'wikidata'
```
ragesoss: Per above comment, I guess this is the place where we'd want to convert 'www' to `nil`. (Doing it in `#ensure_valid_project` as well doesn't hurt, especially since we use FactoryGirl rather than this `.get_or_create` method in some of the specs.)

app/models/wiki.rb
bgreg: ok, fixed

app/models/wiki.rb
bgreg: ok, so should we have the "www" language code only used on the front-end?

ragesoss: The setting of defaults is handled on the backend, so we currently pass in many nils from the frontend that get converted to English Wikipedia. See for example the `set_wiki` methods of CourseCreationManager and of AssignmentsController.

I think handling that "www" here will be the cleanest solution (although I could be convinced otherwise).

bgreg:
I need to do some cleanup, added too many commits with a rebase.

app/models/wiki.rb
```diff
@@ -101,7 +96,7 @@ def self.default_wiki
   end
 
   def self.get_or_create(language:, project:)
-    language = nil if project == 'wikidata'
+    language = nil if MULTILINGUAL_PROJECTS.include?(project)
```
ragesoss: Rolling back to this means the single-language version of Wikisource won't work; they'll all get saved as `nil` even if they are supposed to be 'en.wikisource.org' or the like.

bgreg: ok, I probably misunderstood your advice earlier. I'll take a look tomorrow when I can get back on this.

app/assets/javascripts/utils/course_utils.js
```diff
@@ -58,6 +58,19 @@ const CourseUtils = class {
       };
     }
 
+    const wikiSourceUrlParts = /wikisource\.org\/wiki\/([^#]*)/.exec(articleTitle);
```
ragesoss: 'wikisource' is one term, so changing the caps will avoid ambiguity in this name: `wikisourceUrlParts`

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -64,16 +59,27 @@
...
+      end
+    end
+  end
 
-      expect(Assignment.last.wiki.language).to eq('es')
-      expect(Assignment.last.wiki.project).to eq('wikisource')
-      expect(Assignment.last.article.title).to eq('No_le_des_prisa,_dolor')
-      expect(Assignment.last.article.wiki.language).to eq('es')
-      expect(Assignment.last.article.wiki.project).to eq('wikisource')
+  it 'will create a valid assignment for multiliungual wikisource projects' do
```
ragesoss: spelling: multilingual

lib/assignment_manager.rb
```diff
@@ -42,6 +41,8 @@ def set_clean_title
   def set_article_from_database
     # We double check that the titles are equal to avoid false matches of case variants.
     # We can revise this once the database is set to use case-sensitive collation.
+    #
+    #
```
ragesoss: stray empty comment lines

PR #1238, opened 2017-03-20
Fixing Datepicker time zone problem during course creation

medhabansal:
@ragesoss The bug could be reproduced only for positive timezones, it behaved in the same manner for all positive timezones, and could not be reproduced for any negative timezone. I have thoroughly tested this across all positive and negative timezones and the change here appears to be a valid fix. Could you please review.
Also, I could not find an appropriate file to add related tests. Please help.

### DIFF
--- a/app/assets/javascripts/components/common/date_picker.jsx
+++ b/app/assets/javascripts/components/common/date_picker.jsx
@@ -222,7 +222,7 @@ const DatePicker = React.createClass({
   },
 
   moment(...args) {
-    return this.props.showTime ? moment(...args) : moment(...args).utc();
+    return moment(...args);
   },
 
   showCurrentDate() {


ragesoss:
@medhabansal I think this reintroduces an earlier bug, which unfortunately we didn't find a way to test before. For my time zone, although it shows the correct date in the UI before I save, it changes it to the prior day upon saving.

This is because of they way times are handled in the `Course` model: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/models/course.rb#L317

For a `ClassroomProgramCourse` where the start and end times are not specified, we want the start to be the beginning of the UTC day, and the end to be the end of the UTC day. The frontend defaults to 0:0 for the hours and minutes, so if the datepicker value is not converted to to UTC, then the date will be different after you save and reload the page.

Maybe the solution is to do some UTC conversion within that `#set_default_times` method.

ragesoss:
@medhabansal I think I've found a way we can do testing on this: https://github.com/ariya/phantomjs/issues/10379#issuecomment-36058589

Our feature specs use poltergeist as the Capybara driver, which uses phantomjs. It spawns a separate process from Ruby, so we should be able to set (for example) `ENV['TZ'] = 'Asia/Kolkata'` in a `before` block, then set it (back) to `nil` in an `after` block. I tried it out with logging the timezone from the javascript, and it looks like this works to change the browser timezone in the test.

ragesoss:
As we discussed last week, it's looking like there's no easy fix for this. The right solution is probably to convert `ClassroomProgramCourse` to use the user's local times, and showing the timezone and start/end times in the UI.

PR #1240, opened 2017-03-20
Add Visualizations for User Profile Stats to Profile Pages.

sejalkhatri:
Adding 2 Graphs to the Instructor Stats Component in the User Profile Pages.
1)Courses Taught Graph - Plotting all the courses taught by the instructor on the time scale(highlighting start and end period of the course)
2)Students Taught Graph - Plotting all the students taught by an instructor on the time scale and highlighting the time when the student had enrolled. 

Link to the user profile pages -> /users/[username]


ragesoss:
@keer25 @medhabansal please review!

ragesoss:
@sejalkhatri there's a merge conflict you'll need to fix before we can test it out.

app/assets/stylesheets/modules/_user_profiles.styl
```diff
@@ -143,6 +143,12 @@
 		#inner_statistics
 			padding-top 10px
 
+			.user_stats
+				.stat-display
+					.button
+						border-radius 10px
+						padding 3px 0 8px
+
```
ragesoss: This file is using tabs instead of spaces, it looks like.

app/views/user_profiles/stats_graphs.json.jbuilder
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+if @user.course_instructor?
+  json.instructor_stats do
+    i = 0
```
ragesoss: You could use `each.with_index do |course, i|` instead  of indexing manually.

ragesoss:
There's a stray escape character in the % trained popover:

percent

For the courses taught graph, the heading needs a better title, like 'Courses taught' / 'Programs taught'. The axis label also should switch based on which dashboard it is, like the stats itself does ("Programs Taught" below).

courses taught

Same thing with Students taught: 'StudentsTaughtGraph' should be just 'Students taught', and that header along with the axis label should change depending on whether 'editors' or 'students' is the default.

Also, the visualization here can use some improvement. I don't think the number labels are helpful, and some kind of line plot (or area plot like the structural completeness one) may more clearly convey the data.

students taught

app/assets/stylesheets/modules/_user_profiles.styl
sejalkhatri: yes.. It should work tab = 2spaces, the file looks good

config/locales/en.yml
```diff
@@ -484,6 +485,7 @@ en:
     school_and_term: Institution/When
     student_editors: Editors
     students: Editors
+    students_taught: Editors Inspected
```
ragesoss: Maybe "Total Editors" instead. "Inspected" in this context is maybe particular to Indian and British English.

app/views/user_profiles/stats_graphs.json.jbuilder
```diff
@@ -0,0 +1,23 @@
...
+      json.course_end course.end
+      json.articles_edited course.article_count
+      json.articles_created course.new_article_count
+      json.by_students_common_uploads course.upload_count
+    end
+  end
+
+  json.student_count do
+    i = 0
+    json.array! CoursesUsers.where(course: @courses_presenter.courses, role: 0).each do |course_user|
```
sejalkhatri: each.with_index  do |course_user,  i|   is not working in this case, because of creating json array in the same loop I guess!

PR #1241, opened 2017-03-20
Wrote preliminary draft of #1187

### LINKED ISSUE #1187: Improve styling and relevant info shown in AlertsList
* The resolve link should be a button.
* Instead of showing either 'N/A' or the resolve link, that column should show a check mark or similar indicator if the alert is resolved, and should show nothing at all if it is unresolvable.

mz99:
Added AlertHelper method to handle alert.resolvable and alert.resolved conditionals

Closes #1187

### DIFF
--- a/app/controllers/alerts_list_controller.rb
+++ b/app/controllers/alerts_list_controller.rb
@@ -2,7 +2,6 @@
 
 # Controller for showing Alert records
 class AlertsListController < ApplicationController
-  include AlertsHelper
   layout 'admin'
   before_action :check_user_auth
 
--- a/app/controllers/alerts_list_controller.rb
+++ b/app/controllers/alerts_list_controller.rb
@@ -2,6 +2,7 @@
 
 # Controller for showing Alert records
 class AlertsListController < ApplicationController
+  include AlertsHelper
   layout 'admin'
   before_action :check_user_auth
 
--- a/app/helpers/alerts_helper.rb
+++ b/app/helpers/alerts_helper.rb
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 module AlertsHelper
-  def alert_display
-    if alert.resolvable? == true && alert.resolved == true then
-      '✓'
-    elsif alert.resolvable? == true && alert.resolved == false then
-      button_to('Resolve', resolve_alert_path(alert.id), method: :put,
-      class: 'button small danger dark')
-    end
+  def alert_display(alert)
+    return '✓' if alert.resolved?
+    resolve_button(alert) if alert.resolvable? # implicit return
+  end
+
+  def resolve_button(alert)
+    button_to('Resolve', resolve_alert_path(alert), method: :put, class: 'button small danger dark')
   end
 end
--- a/app/helpers/alerts_helper.rb
+++ b/app/helpers/alerts_helper.rb
@@ -1,3 +1,11 @@
 # frozen_string_literal: true
 module AlertsHelper
+  def alert_display
+    if alert.resolvable? == true && alert.resolved == true then
+      '✓'
+    elsif alert.resolvable? == true && alert.resolved == false then
+      button_to('Resolve', resolve_alert_path(alert.id), method: :put,
+      class: 'button small danger dark')
+    end
+  end
 end
--- a/app/views/alerts_list/index.html.haml
+++ b/app/views/alerts_list/index.html.haml
@@ -14,4 +14,4 @@
         %td= link_to(alert.type, "/alerts_list/#{alert.id}")
         %td= link_to(alert.course.slug, "/courses/#{alert.course.slug}") if alert.course
         %td= link_to(alert.user.username, "/users/#{alert.user.url_encoded_username}") if alert.user
-        %td= alert.resolvable? ? link_to('Resolve', resolve_alert_path(alert.id), method: :put) : 'N/A'
+        %td= alert.alert_display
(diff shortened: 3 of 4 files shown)


app/views/alerts_list/index.html.haml
```diff
@@ -14,4 +14,4 @@
         %td= link_to(alert.type, "/alerts_list/#{alert.id}")
         %td= link_to(alert.course.slug, "/courses/#{alert.course.slug}") if alert.course
         %td= link_to(alert.user.username, "/users/#{alert.user.url_encoded_username}") if alert.user
-        %td= alert.resolvable? ? link_to('Resolve', resolve_alert_path(alert.id), method: :put) : 'N/A'
+        %td= alert.alert_display
```
sejalkhatri: I think you'll have to pass alert object as a parameter to the alert_display function  ```alert_display(alert)```

app/controllers/alerts_list_controller.rb
```diff
@@ -2,6 +2,7 @@
 
 # Controller for showing Alert records
 class AlertsListController < ApplicationController
+  include AlertsHelper
```
sejalkhatri: its not necessary to include ```AlertsHelper``` module in the controller

app/helpers/alerts_helper.rb
```diff
@@ -1,3 +1,11 @@
 # frozen_string_literal: true
 module AlertsHelper
+  def alert_display
+    if alert.resolvable? == true && alert.resolved == true then
```
arun1595: Where is the `alert` object passed?

```ruby
if alert.resolvable? == true && alert.resolved == true then
```
The above line of code sounds imperative. You can simplify it as

```ruby
if alert.resolvable? && alert.resolved?
```

arun1595: Well, thinking more about it, I see both the branches having a common condition. How about rewriting the method as

```ruby
def alert_display(alert)
  return unless alert.resolvable?
  return '✓' if alert.resolved?
  return resolve_button(alert)
end

def resolve_button(alert)
  button_to('Resolve', resolve_alert_path(alert), method: :put, class: 'button small danger dark')
end
```

app/controllers/alerts_list_controller.rb
arun1595: @sejalkhatri  I am not sure. **Is this a new behaviour in Rails 5?** 

AFAIK, the helper file which matches the name of the controller is only included in the controller.

For eg: If the controller name is `AlertsListController`, `AlertsListHelper` is automatically included for you.

app/helpers/alerts_helper.rb
patmbolger: @arun1595, @mz99  - the story says this: ``` Instead of showing either 'N/A' or the resolve link, that column should show a check mark or similar indicator if the alert is resolved, and should show nothing at all if it is unresolvable.```

Also, I think the logic is constrained by this: `resolvable?` will return `true` _if_ the item is, in fact, resolvable _and_ has not been resolved - otherwise `resolvable?` returns `false`.

So, Arun's proposed logic needs to reflect these items.

arun1595: @patmbolger I see. Updated the code. That should do.

app/controllers/alerts_list_controller.rb
sejalkhatri: yes! u are right it says that!! But I don't know why it still works for me when I am not including the helper module in the controller. Would be great if someone has additional input on this.

arun1595: @sejalkhatri I think I was wrong. Please retract my previous comment. All I wanted to convey was that the helper methods matching the resource are automatically available in the views, not in the controllers.

You don't really want to pollute the `controller` by including the all the methods in the helper. In fact. So, I don't think rails will automatically include it for you unless you _ask_ it to by `including` it in the controller.

Someone correct if I am wrong.

> But I don't know why it still works for me when I am not including the helper module in the controller.

Because you're only accessing the helper methods in the views?

sejalkhatri: >For instance: If the controller name is AlertsListController, AlertsListHelper is automatically included for you. 
In this particular case when the helper name is not ```AlertsListHelper```, it still works if you don't include it in the controller (i.e you are able to access the helper method in the alerts_list view) .

mz99: So do we have a consensus whether or not the module should be explicitly included in the controller?

app/helpers/alerts_helper.rb
```diff
@@ -1,3 +1,11 @@
 # frozen_string_literal: true
 module AlertsHelper
+  def alert_display(alert)
+    return '✓' if alert.resolved?
+    resolve_button(alert) if alert.resolvable? # implicit return
```
ragesoss: This has three possible results: a check mark, the button, or nothing.

That will be clearer if you make this three lines, with a guard clause before this line: `return unless alert.resolvable?`

app/helpers/alerts_helper.rb
```diff
@@ -1,3 +1,11 @@
 # frozen_string_literal: true
 module AlertsHelper
```
ragesoss: Since this is a helper for the AlertsListController's index view, we should probably rename this class AlertsListHelper (and change the file name). With that change, there will also be no possibility of confusion over whether the include the module in the controller.

PR #1243, opened 2017-03-23
 Add Information Icon to the Metrics element which have tooltip associated with them

sejalkhatri:
final_info_icon


ragesoss:
This is nice! Seeing it in action locally, I think the icons should be smaller, though. It's rather prominent, and will be a little distracting for people who already know about the mouseover data.

sejalkhatri:
ok! that makes sense, I have made the changes - keeping image width 10%, So that the icon will be easy to understand. Let me know if this works!

ragesoss:
These icons should be desktop only, because the tooltips are not available on mobile layout. Other than that, it's good to go.

Issue #1208, opened 2017-03-02
Update how Article #rating data is imported: use the new PageAssessments API


ragesoss:
Currently, `RatingImporter` gets article ratings by calling `WikiApi.get_article_ratings`, fetching the text of the talk page and using regular expressions (in `ArticleClassExtractor`) to figure out what the rating of an article is.

This only works for English Wikipedia, and is potentially subject to error if the format of common WikiProject templates changes.

The new PageAssessements API provides structured data about article ratings, is available for all languages: https://en.wikipedia.org/w/api.php?action=query&prop=pageassessments&titles=Selfie

`RatingImporter` and related code should be rewritten to use this new API, so that rating data can be imported from every language version of Wikipedia (and Wikivoyage, but it looks like not other projects).

keer25:
PageAssessments API also works only only for English Wikipedia (I referred to [here](https://meta.wikimedia.org/wiki/Community_Tech/PageAssessments). But it has structured data so it will be good to update to PageAssessments API

ragesoss:
Done via #1245

PR #1245, opened 2017-03-26
Get article rating through pageassessments API

keer25:
As the existing method uses reqular expressions to extract classes from talk page getting article class through the structured
page assessments API is more efficient and less prone to error. Issue #1208


ragesoss:
@keer25 nice work!

A few things:

* Using the `formatversion=2` option would probably make the json processing significantly simpler.
* This is an opportunity to move even more of the ratings-specific code out of `WikiApi`. Although it's not there yet, `WikiApi` ought to be concerned only fetching data from MediaWiki, not with processing it. The new code for processing that data should live in a separate class — either in `RatingImporter` or better yet, a new class.
* Can you turn the continue-handling code into something more general? If so, that could stay in `WikiApi`. If not, it should be moved out along with the other methods that deal with processing pageassessments data.

Another thing this may do is result in assessment values that we don't account for. With `ArticleClassExtractor`, we only get ratings that we know how to deal with on the frontend (in `ArticleHelper#rating_priority`). It looks like there are no limits to what values might come out of the pageassessment API. In that case, we need to add a default to that helper method. It would also be useful to add some additional tests that explicitly handle more of the possible ratings that the API may return.

keer25:
image
These are the exhaustive list of classes. bplus should default and nil for other unidentified classes as per article extractor

keer25:
After going through this [https://en.wikipedia.org/wiki/MediaWiki:Gadget-metadata.js](https://en.wikipedia.org/wiki/MediaWiki:Gadget-metadata.js). I thing we should also add a/ga to a. 
And al, bl, cl, sl to list.

ragesoss:
I agree. I don't think al, bc, cl, and sl existed when we first implemented the ruby adaptation.

keer25:
image
What can be the reason this failure also occurs in the master branch of the project.

ragesoss:
Hmm... Is it just that test, or all the feature specs?

I think I've seen that a couple of times locally, but I'm not sure what the cause is. It's not directly related to the app or spec code, I don't think.

keer25:
I attempted a generalised continue handling code that can work on fetching a bunch of pages with any property. As specified in article helper{ fa fl a ga b c start stub list} is the exhaustive list of rating that can be returned. Guess documentation of this should be in relevant files

lib/article_class_fetcher.rb
```diff
@@ -0,0 +1,54 @@
...
+  	if %w(fa fl a ga b c start stub list).include? rating
+  		return rating
+  	elsif rating.eql? 'bplus'
+  		return 'b'
+  	elsif rating.eql? 'a/ga'
+  		return 'a'
+  	elsif %w{al bl cl sl}.include? rating
+  		return 'list'
+  	else
+  		return nil
```
ragesoss: I think we should save whatever rating the API provides, and this winnowing of known / simplified classes should happen at the view layer (by changing ArticleHelper). That way, we'll be saving a fuller set of data that we can do more things with later on if necessary.

ragesoss:
I will review it more fully soon, but this looks excellent.

lib/article_class_fetcher.rb
keer25: Agreed

lib/wiki_api.rb
```diff
@@ -51,51 +51,56 @@ def get_article_rating(titles)
...
-    ArticleClassExtractor.new(wikitext).extract
+    query_parameters = { titles: titles,
+                         prop: 'pageassessments', 
+                         formatversion: '2',
+                         redirects: 'true' }
+    response = get_responses_continue(query_parameters)
+    return [] unless response
+    return [] unless response['query'] 
+
+    info = response['query']['pages']
```
ragesoss: Instead of these guard clauses, we can use `.dig` to safely retrieve the expected `pages` value.

lib/wiki_api.rb
```diff
@@ -51,51 +51,56 @@ def get_article_rating(titles)
...
+    return [] unless info
```
ragesoss: Let's have ArticleClassFetcher handle `nil` input by returning `[]`.

lib/article_class_fetcher.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+#= Fetches article classes from PageAssessments response
+# We simplify this parser by folding the nonstandard ratings
+# into the corresponding standard ones. We don't want to deal with edge cases
+# like bplus and a/ga.
+
+class ArticleClassFetcher
+	def initialize(pageassessments)
+		@assessments = pageassessments
```
ragesoss: By changing this to `pageassessements || []` we can smoothly handle `nil` input.

lib/wiki_api.rb
```diff
@@ -51,51 +51,56 @@ def get_article_rating(titles)
...
+    ArticleClassFetcher.new(info).get_article_rating
```
ragesoss: This class isn't fetching any data, it's just parsing and extracting the ratings from the API data. 'Class' is also a potentially confusing term here; it's Wikipedia jargon, but it collides with 'class' in the programming sense. Maybe we can call this `ArticleRatingExtractor' or `ArticleRatingParser` or similar?

ragesoss: The method name here should be plural; maybe simply `ratings`?

lib/wiki_api.rb
```diff
@@ -51,51 +51,56 @@ def get_article_rating(titles)
...
+          response = new_response
+      end
+      if new_response['continue'] != nil
+        new_response['continue'].map do |mod, parameter|
+          query_parameters[mod] = parameter 
+        end
+      else
+        break
+      end
+    end
```
ragesoss: This is a very complex method, especially with the nesting and the if/else blocks. I suspect this can be simplified considerably. The `continue` implementation in `lib/commons.rb` may be a useful point of reference.

keer25: I generalised it that's is why it is a little more complex and using .dig will simplify the code

ragesoss: I took a closer look and got sucked into the task... formatversion 2 is nicer for single queries, but it makes continuation very painful!

I think a general continuation implementation is only worth it if we can make it return the same data that a single query would return if there was no limit. I took a pass at it (without formatversion 2) here: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/9f7213db4a5276aa2bf056dddd13eff490f5525d

ragesoss:
There are lots of rubocop style violations that need to be cleaned up as well: two spaces (not tabs) for indentation, and no trailing whitespace.

lib/article_class_fetcher.rb
keer25: As of now the ratings are stored in Article.rating attribute which is used as a parameter to get rating_docs and the methods display_rating and rating_helper which can only handle finite ratings ( I hope this is the exhaustive list)

We can change these methods to be inclusive of more ratings like bplus, etc... which is a pretty huge task and can have a commit of its own

And save any rating we get in an rating_all : string(255) attribute in Article model and save the default ratings in the raing atribute

lib/importers/rating_importer.rb
```diff
@@ -55,7 +55,8 @@ def self.update_ratings(article_groups)
   ####################
   def self.update_article_ratings(articles, ratings)
     articles.each do |article|
-      article.rating = ratings[article.title]
+      article.rating_all = ratings[article.title]
```
ragesoss: Instead of saving the original full-spectrum rating separately, I think we should only store that rating (as `article.rating`) and then basically do what you are doing with default_rating() below within the ArticleHelper methods 'rating_display' and 'rating_priority'

I don't want to add a new column to the database for this.

ragesoss: (Continuing from IRC)... I see your point about how the use of article.rating is not completely abstracted for use with the ArticleHelper method... we pass three different versions of the rating data to the frontend, via the course articles.json endpoint: 'rating' (the one from the model directly), 'pretty_rating' (from `ArticleHelper#display_rating`) and 'rating_num' (from `ArticleHelper#rating_priority`). However, I think it makes more sense to save only the full ratinig, and deal with whatever display issues come up later on. For the rating docs, the best solution will probably be to add more docs for ratings that aren't currently supported.

app/helpers/article_helper.rb
```diff
@@ -62,11 +63,28 @@ def rating_priority(rating)
   # rubocop:enable Metrics/CyclomaticComplexity
 
   def rating_display(rating)
+    ratiing = default_class(rating)
```
ragesoss: typo: `ratiing`

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170407203227) do
+ActiveRecord::Schema.define(version: 20170409172645) do
```
ragesoss: The schema change got left behind.

keer25:
I will create another issue to add rating_docs in i18n and appropriate CSS classes in _course.styl ?

ragesoss:
Sounds good.

-Sage

db/schema.rb
```diff
@@ -49,6 +49,7 @@
     t.date     "average_views_updated_at"
     t.integer  "wiki_id"
     t.integer  "mw_page_id"
+    t.string   "rating_all"
     t.index ["mw_page_id"], name: "index_articles_on_mw_page_id", using: :btree
```
ragesoss: One more bit left.

lib/article_rating_extractor.rb
```diff
@@ -0,0 +1,40 @@
...
+      ratings[page['title'].tr(' ', '_')] ||= get_article_class(page['pageassessments'])
+    end
+    return ratings
+  end
+
+  private
+
+  def get_article_class(pageassessments)
+    # Returns nil if no page assessment data found
+    return nil if pageassessments.nil?
```
ragesoss: This is unnecessary since the following check will also work for `nil`.

lib/article_rating_extractor.rb
```diff
@@ -0,0 +1,40 @@
...
+    return nil unless pageassessments.is_a? Enumerable
```
ragesoss: What is the purpose of this check, beyond the nil check? Did you find cases where this would receive a non-enumerable non-nil argument, or is it just double-checking to making sure the following operation is safe?

lib/article_rating_extractor.rb
```diff
@@ -0,0 +1,40 @@
...
+    pageassessments.map do |project, info|
```
ragesoss: Doing a 'map' here is a little confusing, since we only care about the first value, and we definitely don't care about the return value. We could instead do something like `pagesassessments.values.first` to get the wikiproject we want to extract the class from.

PR #1246, opened 2017-03-28
Disallow courses to be created with end time before start time

### LINKED ISSUE #1220: Course can be created with end time before start time
Although the dates work pretty well in terms of front-end validation, the times are not checked.

AM/PM confusion in particular can lead to invalid dates such that no data will be collected.

There should be validation in both the javascript and in the Course model.

arun1595:
Fixes #1220

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -33,6 +33,8 @@ class CoursesController < ApplicationController
              status: 404
       return
     end
+    # TODO: Add strict datetime validations to course creation and
+    # add a clause to handle validation errors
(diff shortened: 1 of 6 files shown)


app/controllers/courses_controller.rb
```diff
@@ -33,6 +33,7 @@ def create
              status: 404
       return
     end
+    # TODO: Update the below line
```
arun1595: We need to add a conditional to make sure that if a rollback occured, we should return json with the error messages. We might have to update the create method of `CourseCreationManager`.

Issue #1180, opened 2017-02-16
Add support for incubator.wikimedia.org


ragesoss:
Some projects involve new users editing on the incubator wiki. We'll need to special-case this in `Wiki` as we do with wikidata.

bachounda:
hi 
Is it possible to integrate incubator projects into the dashboard because students in our education program.https://outreachdashboard.wmflabs.org/courses/Chlef_university__Algeria/Wikipedia_education_program__2016-2017

They also contribute to the incubator

Otherwise I really use the dashboard / and I think it's a great program

We discovered another way to use the dashboard in contest - we "jury members" can follow contributors contests with the dashboard and it's very easy

PR #1251, opened 2017-03-29
[WIP] Adding support for incubator.wikimedia.org

medhabansal:
@ragesoss Please review.
Done with the assumption that the incubator project does not offer language support.
Also, name of the incubator project in the Wiki model has been kept as 'wikimedia'. Other options in mind were 'incubator_wikimedia' and 'wikimedia_incubator', but they exceeded number of allowed column characters in the DB, please suggest if it needs to be changed. 

Refers #1180

### DIFF
--- a/lib/replica.rb
+++ b/lib/replica.rb
@@ -155,12 +155,15 @@ class Replica
     "#{base_url}#{endpoint}?#{language_and_project_params}&#{query}"
   end
 
-  # We must special-case Multilingual Wikisource because its labs database name
-  # does not follow the normal pattern; it is called 'sourceswiki', which is the
-  # same one we'd get if a Wikipedia's language was "sources".
+  # We must special-case Multilingual Wikisource/ Wikimedia Incubator because
+  # their labs database names do not follow the normal pattern; they are called
+  # 'sourceswiki' and 'incubatorwiki' respectively, which are the same as we'd
+  # get if a Wikipedia's language was "sources" or "incubator".
   def language_and_project_params
     if @wiki.project == 'wikisource' && @wiki.language.nil?
       'lang=sources&project=wikipedia'
+    elsif @wiki.project == 'wikimedia' && @wiki.language == 'incubator'
+      "lang=incubator&project=wikipedia"
     else
       "lang=#{@wiki.language}&project=#{@wiki.project}"
     end
--- a/spec/features/multiwiki_assignment_spec.rb
+++ b/spec/features/multiwiki_assignment_spec.rb
@@ -82,4 +82,22 @@ describe 'multiwiki assignments', type: :feature, js: true do
       end
     end
   end
+
+  it 'will create a valid assignment for multilingual wikimedia incubator projects' do
+    VCR.use_cassette 'multiwiki_assignment' do
+      visit "/courses/#{course.slug}/students"
+      click_button 'Assign Articles'
+      click_button 'Assign an article'
+      within('#users') do
+        first('input').set('https://incubator.wikimedia.org/wiki/Wp/kiu/Heyder_Cansa')
+      end
+      click_button 'Assign'
+      click_button 'OK'
+      visit "/courses/#{course.slug}/students"
+
+      within('#users') do
+        expect(page).to have_content 'incubator:wikimedia:Wp/kiu/Heyder Cansa'
+      end
+    end
+  end
 end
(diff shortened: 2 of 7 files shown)


ragesoss:
As we discussed on IRC, we should implement this with 'language' as 'incubator' and 'project' as 'wikimedia'. That will leave room for other 'wikimedia.org' subdomains, and it means we may not need to special-case this within the Wiki model... only in Replica.

medhabansal:
@ragesoss I have made the changes but have a few questions:
1) How do we handle an invalid language(apart from incubator currently) for the Wikimedia project, should we store nil or raise an error in this case?(I have commented the test for the same as of now)
2) Should we include WikiEduDashboardTools [cleanup](https://github.com/WikiEducationFoundation/WikiEduDashboardTools/blob/9a1b62f37ae9a2f82ce8edaddcd9d965b084aec6/public_html/common.php#L64) in this issue?
3) I am having trouble understanding VCR wrt the replica spec tests, could you please help me there?
Thanks!

lib/replica.rb
```diff
@@ -155,12 +155,16 @@ def compile_query_url(endpoint, query)
...
-  # same one we'd get if a Wikipedia's language was "sources".
+  # We must special-case Multilingual Wikisource/ Wikimedia Incubator because
+  # their labs database names do not follow the normal pattern; they are called
+  # 'sourceswiki' and 'incubatorwiki' respectively, which are the same as we'd
+  # get if a Wikipedia's language was "sources" or "incubator". Special case for
+  # Wikimedia however also handles addition of future sub-domains.
   def language_and_project_params
     if @wiki.project == 'wikisource' && @wiki.language.nil?
       'lang=sources&project=wikipedia'
+    elsif @wiki.project == 'wikimedia' && @wiki.language
```
ragesoss: I'm not sure how general this is. For now, we should just handle the explicit case of project 'wikimedia', language 'incubator'. here.

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -89,7 +89,7 @@
       click_button 'Assign Articles'
       click_button 'Assign an article'
       within('#users') do
-        first('input').set('https://wikisource.org/wiki/Wp/kiu/Heyder_Cansa')
```
ragesoss: This test with the wikisource url should stay in place, but we can add an additional test for incubator.

ragesoss: Whoops, nevermind. I was just looking at the single commit, not the current whole PR. Ignore that.

spec/lib/replica_spec.rb
```diff
@@ -178,6 +178,21 @@
         expect(response.count).to eq(28)
       end
     end
+
+    it 'functions identically on wikimedia incubator' do
+      VCR.use_cassette 'replica/wikimedia_incubator_revisions' do
+        all_users = [
+          build(:user, username: 'Ragesoss')
```
ragesoss: This should confirm that we can actually get results back. Choose a user and date change that return some revisions. For example: http://tools.wmflabs.org/wikiedudashboard/revisions.php?lang=incubator&project=wikipedia&usernames[]=Daad%20Ikram&start=20170311000000&end=20170317000000

Username: `Daad Ikram`

dates: 20170311000000 to 20170317000000

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -82,4 +82,22 @@
...
+      click_button 'Assign an article'
+      within('#users') do
+        first('input').set('https://incubator.wikimedia.org/wiki/Wp/kiu/Heyder_Cansa')
+      end
+      click_button 'Assign'
+      click_button 'OK'
+      visit "/courses/#{course.slug}/students"
+
+      within('#users') do
+        expect(page).to have_content 'wikimedia:Wp/kiu/Heyder Cansa'
```
ragesoss: This should check for the language as well: `incubator:wikimedia:...`

spec/lib/replica_spec.rb
medhabansal: Yes, I was having trouble looking for actual data, thanks!

spec/features/multiwiki_assignment_spec.rb
medhabansal: Made the change.

PR #1253, opened 2017-04-04
Training caching

ragesoss:
@mattfordham This scheme of loading from yaml to the Rails cache came from you, I think.

I'd love a quick sanity check on what I'm doing... basically pushing the main caching to class instance variables, since that's much faster than reading from the Rails cache ever time the `all` method is called. Does this seem like a good idea?

### DIFF
--- a/lib/training/training_base.rb
+++ b/lib/training/training_base.rb
@@ -27,19 +27,38 @@ class TrainingBase
     check_for_duplicate_ids
   end
 
+  # Called during initialization, and also via manual :training_reload action.
+  # This should regenerate all training content from yml files and/or wiki.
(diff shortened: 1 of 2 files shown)


mattfordham:
@ragesoss This looks good to me. The only thing I'd watch out for is server memory consumption. I may make no significant difference, but the class variable will keep it in memory, as opposed to just loading it when needed from cache 👍

ragesoss:
Great. I think we're okay for memory usage, and the trainings aren't super big. They might get big if on-wiki creation and translation really take off on Programs & Events Dashboard, but we can worry about that if/when it becomes a problem.

Issue #1184, opened 2017-02-16
Wikis should be validated to ensure they exist


ragesoss:
`Wiki` currently has a basic validation for new records based on a list of language codes and projects. However, not all project/language codes actually exist. For example, `dk.wikipedia.org` does not exist; it redirects to `da.wikipedia.org`. Invalid wikis result in data not being fetched and lots of noise in the Sentry logs for all the API failures when when `Replica` cannot connect to the (nonexistent) database.

At the time of creation, each wiki should be validated via a ping to its API.

ragesoss:
* During validation before a new `Wiki` record is created, ping the corresponding API to confirm that it exists. If not, catch the network error and raise an `InvalidWiki` error.

FedericoEsparza:
https://youtu.be/d-8MafzIxIU?t=52m2s Some discussion and point voting in the 2017-02-21 AV Americas Scrum recording. _(Link bookmarks and directs to the starting time)_

PR #1256, opened 2017-04-06
Validate wiki

hin101:
Closes Waffle #1184 

Also cleaned up spec/models/wiki_spec.rb to comply with rubocop.


ragesoss:
Please squash this to a single commit, to avoid adding all the merges to the history.

ragesoss:
Unfortunately, this wreaks havoc with the test suite because it adds an external web request to many tests.

I'm not sure what the best way to handle this is. I guess one way would be to stub all of the needed requests from the spec_helper.

To do that, basically we'd edit request_helpers to add a method like `stub_wiki_validations`, and then put multiple stub requests for all the wikis that are causing problems in the test suite.

I'll take a pass at that.

ragesoss:
The test side of this is way trickier than I had anticipated. I have one possible approach here: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/101a47b9020ef2548fd2c0411f8c354df483a635

I'd love to get some other opinions on what the best way to solve this is. @bgreg, this might be something you have some insight on.

bgreg:
@ragesoss  I like what you got going on in https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/101a47b9020ef2548fd2c0411f8c354df483a635

ragesoss:
Continued in https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1257

PR #1262, opened 2017-04-14
Passing project and language as-is for querying in Replica

medhabansal:
@ragesoss Related to [PR](https://github.com/WikiEducationFoundation/WikiEduDashboardTools/pull/9). Please review.

### DIFF
--- a/lib/replica.rb
+++ b/lib/replica.rb
@@ -150,20 +150,21 @@ class Replica
     Net::HTTP::get(URI.parse(url))
   end
 
+  # Query URL for the WikiEduDashboardTools repository
   def compile_query_url(endpoint, query)
     base_url = 'http://tools.wmflabs.org/wikiedudashboard/'
-    "#{base_url}#{endpoint}?#{language_and_project_params}&#{query}"
+    "#{base_url}#{endpoint}?#{project_database_params}&#{query}"
   end
 
-  # We must special-case Multilingual Wikisource/ Wikimedia Incubator because
-  # their labs database names do not follow the normal pattern; they are called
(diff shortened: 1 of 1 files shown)


lib/replica.rb
```diff
@@ -150,29 +150,30 @@ def do_query(endpoint, query)
     Net::HTTP::get(URI.parse(url))
   end
 
+  # Query URL for the WikiEduDashboardTools repository
   def compile_query_url(endpoint, query)
     base_url = 'http://tools.wmflabs.org/wikiedudashboard/'
-    "#{base_url}#{endpoint}?#{language_and_project_params}&#{query}"
+    "#{base_url}#{endpoint}?lang=#{@wiki.language}&project=#{@wiki.project}\
```
ragesoss: We need either the lang/project params or the db param, but not both.

lib/replica.rb
```diff
@@ -150,29 +150,30 @@ def do_query(endpoint, query)
...
 
   def compile_usernames_query(users)
     { usernames: users.map(&:username) }.to_query
   end
 
+  # Returns special Labs database names for databases not meeting
+  # project/language naming conventions, else nil
+  def db_params
+    if @wiki.project == 'wikidata'
+      'wikidatawiki'
```
ragesoss: Instead of just doing the db value, you could return the whole query param, `db=wikidatawiki` and such, and then fall back to `"lang=#{@wiki.language}&project=#{@wiki.project}"` at the end.

medhabansal:
@ragesoss Made changes and rebased the branch.

PR #1272, opened 2017-04-29
Search by real name

### LINKED ISSUE #1255: Admins should be able to search for users by real name
Currently, we can search by email via `/users`. Adding a separate option to search by real name (`User.real_name`) would be handy. This should be straightforward to implement by following the pattern of the initial search-by-email implementation.

hin101:
Closes #1255


ragesoss:
@hin101 please take a look at the failing specs, which are the ones for the existing email search functionality: `spec/models/user_spec.rb`.

app/models/user.rb
```diff
@@ -91,8 +91,12 @@ module Permissions
   ####################
   # Class method(s)  #
   ####################
-  def self.search_by_email(email)
-    User.where('lower(email) like ?', "#{email}%")
+  def self.search_by_email_or_real_name(email, real_name)
+    if email.nil?
+      User.where('lower(real_name) like ?', "#{real_name}%")
```
ragesoss: If we change it to "%#{real_name}%" then it will match the input if it occurs anywhere in the real name field (allowing search for just last name, for example).

app/models/user.rb
```diff
@@ -91,8 +91,12 @@ module Permissions
   ####################
   # Class method(s)  #
   ####################
-  def self.search_by_email(email)
-    User.where('lower(email) like ?', "#{email}%")
+  def self.search_by_email_or_real_name(email, real_name)
```
ragesoss: This would be clearer if instead of a single method that works differently depending on the parameters, we used two different methods. Since this only works for one or the other type of input, there's nothing we need a combined method for.

app/controllers/users_controller.rb
```diff
@@ -55,8 +55,8 @@ def enroll
   # User listing page for Admins                     #
   ####################################################
   def index
-    @users = if params[:email].present?
-               User.search_by_email(params[:email])
+    @users = if params[:email].present? || params[:real_name].present?
+               User.search_by_email_or_real_name(params[:email], params[:real_name])
```
ragesoss: This would become an if / elsif / else if we switch to two different query methods on `User`.

ragesoss:
@hin101 changes look good, but tests need an update.

ragesoss:
Nice job @hin101! I rebased and merged.

PR #1273, opened 2017-05-01
Added support for videos in Wiki slides

OmarShehata:
Right now, in order to embed a video in a slide generated from a Wiki page, you would have to enter the raw HTML, which would look good on the dashboard slide but wouldn't render correctly on Wiki.

This change adds support for a video template in the same way the image template works. The Wiki page would have a [template](https://meta.wikimedia.org/wiki/Template:Training_module_video) like this:

```
{{Training module video
| video = File:How to Use the Dashboard (2 of 5).webm
| source = https://upload.wikimedia.org/wikipedia/commons/7/79/How_to_Use_the_Dashboard_%282_of_5%29.webm
| caption = How to Use the Dashboard (2 of 5)
}}
```

On Wiki, this would embed the file normally (like `[[{{{video}}}|thumb|{{{caption}}}]]`). But on the dashboard it would insert an `<iframe>` tag. 

I've also added a test in the relevant spec file for it. 

This is my first contribution to a Ruby project so I welcome any advice/tips!

### DIFF
--- a/lib/training/wiki_slide_parser.rb
+++ b/lib/training/wiki_slide_parser.rb
@@ -12,6 +12,7 @@ class WikiSlideParser
     remove_translate_tags
     extract_quiz_template
     convert_image_template
+    convert_video_template
   end
 
   # The first translated line is the slide title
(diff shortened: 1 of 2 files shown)


ragesoss:
Nice work! In trying it out, I turned up a bug that would prevent multiple different templates on the same slide, which I fixed with 5970665885150f599dd689233b3d44bcc7a4a889.

If there is a way to embed the webm files without having them autoplay, that would be a nice thing.

PR #1275, opened 2017-05-05
Start of a switchover to Redux instead of Flux stores

ragesoss:
Before going any further down this path, I would love to get feedback on whether I'm going in the right direction here.

Here's what I've done:
* Added the start of a Redux store with a single reducer, which does the same thing as UIStore (the store for which single UI component is 'open' currently). https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/ReduxYay?expand=1#diff-6cb8b824811965837dbd2e078c12de7d
* Wrapped the Router with Provider to inject the store https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/ReduxYay?expand=1#diff-1d40ff7b193d8b42fbff64c9f957da6f
* Used `connect()` to make the store available to the Course component, and passed down the store state (`reduxState`) and the store dispatch method as props https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/ReduxYay?expand=1#diff-a5f6be881fdf9e1920a4ee8b20122d9d
* Replaced the use of UIStore / UIActions in StudentDrawer and ArticleDrawer with the Redux version (which allowed bypassing 'Expandable' and just rendering based on the props). in part: https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/ReduxYay?expand=1#diff-910f38803dc117ebb0cc3f8e6b72109a

What I mainly want to know:
* Is this way of connecting up the Redux store and dispatching actions to it a good way? (It seems a little convoluted to be passing the store state and the dispatch function, but that seems preferable to using `connect()` in more than one place.)
* Anything important I'm missing?

cc @mattfordham


mattfordham:
Nice work @ragesoss!

So, I believe the recommended practice, and the way I've been using Redux, is to NOT pass down the store or the dispatch method. 

In other words, each component that needs access to the store uses `connect` to subscribe to *only the data it cares about*. This makes things more performant, as Redux will only re-render components if their specific store concerns are updated. By passing the store as a prop, everything that receives it will re-render with every change. 

Not all components should need `connect` however. Some should be "Presentational Components" components and only receive the data they are concerned about as props. This discusses the difference: http://redux.js.org/docs/basics/UsageWithReact.html#presentational-and-container-components.

So, consider using `connect` on the child components AND/OR just passing down only the props that the child components need. 

Make any sense? ;)

ragesoss:
@mattfordham thanks, that's helpful. I'm most of the way through the [getting started with Redux tutorials](https://egghead.io/courses/getting-started-with-redux) and this is making a little more sense... the key being to use container components as the place where all the action functions get defined, and they just get passed as props rather than passing dispatch itself.

mattfordham:
Yeah, that's a great series. 

Here is roughly the pattern I am using: https://gist.github.com/mattfordham/9bb8587866b955f670e6b6062ff77a52

Notice how I am exporting the MyComponent class and then exporting the "connected" version as default. This essentially makes MyComponent presentational and very easy to test, as there are no dependencies on Redux.

mattfordham:
... and then I have other components that have no "connected" counterpart... they simply present passed down data. In my example, perhaps there would be a totally presentational "Person" component for each item in the "people" array.

ragesoss:
Thanks! After seeing the refactoring of components into presentational ones that are actually just functions, I think working on *that* will be a better way to start, instead of starting with Redux itself. For example, 51571bc74014ec1aaf31cfa8feaa17bb0985d4d3

mattfordham:
Agreed! 👍

PR #1276, opened 2017-05-10
Wiki Slide Parsing Multiple Images & Remove Categories tag

OmarShehata:
This PR mainly:

- Removes the category tag (ex: `[[Category:Programs & Events Dashboard]]`) which is useful to have on Wiki pages but should not show up on the slides
- Refactors the image parsing to allow for parsing multiple images correctly. Right now if there are multiple image templates, the parser will replace them all with the last match. 
- Does the same thing for videos as well. 

This is needed so that the [Using the dashboard](https://outreachdashboard.wmflabs.org/training/learning-and-evaluation/using-the-dashboard) module works correctly. There is at least one slide with multiple images, and all the slides have the category tag.

### DIFF
--- a/lib/training/wiki_slide_parser.rb
+++ b/lib/training/wiki_slide_parser.rb
@@ -10,6 +10,7 @@ class WikiSlideParser
     remove_noinclude
     remove_translation_markers
     remove_translate_tags
+    remove_category
     extract_quiz_template
     convert_image_template
     convert_video_template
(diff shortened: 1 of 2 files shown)


lib/training/wiki_slide_parser.rb
```diff
@@ -50,6 +51,12 @@ def remove_noinclude
     @wikitext.gsub!(%r{<noinclude>.*?</noinclude>\n*}m, '')
   end
 
+  # Category tags are useful for categorizing pages, but we don't want them to show up in the slides
+  # Example: [[Category:Programs & Events Dashboard]]
+  def remove_category
+    @wikitext.gsub!(%r{\[\[Category:.*?\]\]\n*}m, '')
+  end
```
ragesoss: This should have a test.

Alternatively, we could just wrap the categories in noinclude tags and remove this method, which maybe makes more sense anyway in this context. (I can imagine trainings where we want to show how to add categories, and so unparsed category markup might appear on the slide, which would then get incorrectly removed by this code.)

ragesoss:
I removed the category bit and merged it.

PR #1279, opened 2017-05-16
Start of switch from Flux to Redux

ragesoss:
This sets up most of the core Redux plumbing and replaces UIStore for handling the state of the ArticleList subcomponents.

I've followed the naming conventions from the Redux todo example: https://github.com/reactjs/redux/tree/master/examples/todomvc/src


ragesoss:
@mattfordham here's take 2. This one follows the standard pattern pretty closely, I think.

mattfordham:
Nice! Yes, that looks much more "standard" for sure :)

PR #1280, opened 2017-05-18
[WIP] Adding editing suggestions

keer25:
Adding editing suggestions can help users to choose articles to edit based on
the editing requirements on the article

### DIFF
--- a/app/assets/javascripts/components/articles/available_article.jsx
+++ b/app/assets/javascripts/components/articles/available_article.jsx
@@ -3,6 +3,7 @@ import CourseUtils from '../../utils/course_utils.js';
 import ServerActions from '../../actions/server_actions.js';
 import AssignmentActions from '../../actions/assignment_actions.js';
 import NotificationActions from '../../actions/notification_actions.js';
+import EditingSuggestions from '../common/editing_suggestions.jsx';
 
 const AvailableArticle = React.createClass({
   displayName: 'AvailableArticle',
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/articles/available_article.jsx
```diff
@@ -70,6 +81,16 @@ const AvailableArticle = React.createClass({
       );
     }
 
+    // Shows editing suggestions only when suggestions are available for that class
+
+    let feedbackButton;
+    if (this.isEnglishWikipedia &&
+        !I18n.t(`suggestions.suggestion_docs.${assignment.article_rating}`).includes("missing")) {
```
ragesoss: The `.includes()` is causing an error in the feature spec, probably because that String method is not available in the poltergeist environment. It would also break on Internet Explorer: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes#Browser_compatibility

More generally this is a fragile way of handling the choice of whether to display it. I think it would be better to leave off the check against article_rating altogether, and provide a fallback message within `EditingSuggestions` that describes the case of an unrated article (or where relevant, an not-yet-created article), noting explicitly that the article is unrated. That will make more sense for users; it would be confusing to have some available articles have the suggestions button and other not, without an clear indication of why.

ragesoss: Also: `this.isEnglishWikipedia` is the function itself, rather than the return value of the function. Should be `this.isEnglishWikipedia()`.

PR #1293, opened 2017-06-01
Campaign articles

ragesoss:
Adds a very simple list of articles edited to the Campaign view, as a new tab.


ragesoss:
@keer25 @sejalkhatri @medhabansal If any of you would like to review this, please do!

PR #1296, opened 2017-06-02
Added button for feedback

keer25:
Added buttons in the MyArticles component to present some feedback on the
articles being edited


ragesoss:
@medhabansal @sejalkhatri please review!

app/controllers/revision_feedback_controller.rb
```diff
@@ -3,19 +3,54 @@
...
+    @article = Article.find(params[:article_id])
+
+    query = { prop: 'revisions', pageids: @article.mw_page_id, rvprop: 'ids' }
+    response = WikiApi.new(@wiki).query(query)
+    results = response&.data
+    revisions = results.dig('pages', @article.mw_page_id.to_s, 'revisions')
+
+    # The API sends a response with the id of the last revision
+    unless revisions.nil? || revisions.length == 0
+      @revId = revisions[0]['revid']
```
ragesoss: Standard Ruby style is to use snake_case for variable names: `@rev_id` in this case.

app/controllers/revision_feedback_controller.rb
```diff
@@ -3,19 +3,54 @@
...
-    @revision = Revision.new(wiki_id: @wiki.id, mw_rev_id: params[:rev_id], mw_page_id: 0)
-    @revision.save!
-    RevisionScoreImporter.new.update_revision_scores([@revision])
-    @revision.reload
+  def initialize_ores
+    @wiki = Wiki.find_by(language: 'en', project: 'wikipedia')
+    @ores_api = OresApi.new(@wiki)
+  end
+
+    def ores_features(revId)
```
ragesoss: This and the related methods duplicate a lot of logic from `RevisionScoreImporter`. Instead, we should probably just add a method to `RevisionScoreImporter` that fetches the ores data for a single rev_id without saving anything to the database. It looks like that just means moving this method to `RevisionsScoreImporter` (maybe as `#fetch_ores_features_for_revision_id`) and then call that instead of having the controller need to know about using and parsing responses from the ORES api.

app/controllers/revision_feedback_controller.rb
```diff
@@ -3,19 +3,54 @@
...
+    ores_data = @ores_api.get_revision_data(revId)
+    score = extract_score ores_data
+    return extract_features(ores_data)[revId.to_s]
+  end
+
+  def set_revision_id
```
ragesoss: This method name doesn't convey enough about what the method does; it specifically sets the revision id variable for the *latest* revision to the article. It also relies on the article being present in the database already, and the client knowing its id, then it does an api call to get the revisions and extract the latest one from the response. Maybe `set_latest_revision_id`

We could simplify things by having the client send the article title, and querying by title instead of pageids. Then we wouldn't rely on our database for any of the functionality.

lib/revision_feedback_service.rb
```diff
@@ -1,12 +1,18 @@
 # frozen_string_literal: true
 class RevisionFeedbackService
-  def initialize(revision)
+  # Imports the features from the revisions if revisions is nil otherwise uses the features passed
+  def initialize(features = nil, revision = nil)
```
ragesoss: If we're not going to use this with a revision as the argument at this point, then we may as well remove that branch and just take a features hash as the required argument.

For interfaces that need to be flexible in terms of taking one kind of argument or another, it's nicer to use keyword arguments instead of positional arguments. Otherwise, using the revision option would require passing nil as the first argument.

app/assets/javascripts/components/common/feedback.jsx
```diff
@@ -31,16 +46,25 @@ const Feedback = React.createClass({
...
+
+    for (let i = 0; i < messages.length; i++) {
+      feedbackList.push(<li>{messages[i].message}</li>);
+    }
     if (!this.state.show) {
       modal = <div className="empty" />;
     } else {
       modal = (
         <div className="article-viewer feedback">
           <h2>Feedback</h2>
```
sejalkhatri: **Styling suggestion:** You could add padding-bottom to this tag, for proper spacing between header and feedback details.

app/assets/javascripts/components/common/feedback.jsx
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    let modal;
+    if (!this.state.show) {
+      modal = <div className="empty" />;
+    } else {
+      modal = (
+        <div className="article-viewer feedback">
+          <h2>Feedback</h2>
+          <a className="button small diff-viewer-feedback" href="" target="_blank">{I18n.t('courses.suggestions_feedback')}</a>
```
sejalkhatri: Styling: The margin-right property of diff-viewer-feedback class must be set to 0, So as to align it with the Okay button.

app/controllers/revision_feedback_controller.rb
keer25: Importers seem to be mostly dealing with fetching and saving it in the database. As of now I will refactor RevisionScoreImporter to have nothing to do with ORES. Maybe when ORES is saved in the database in the future RevisionScoreImporter can import from OresApi

keer25: Is RevisionScores imported for anything other than the Feedback Service

app/controllers/revision_feedback_controller.rb
keer25: The data flow is such that article always exists when the controller is called

app/controllers/revision_feedback_controller.rb
ragesoss: RevisionScoreImporter is used in the update cycle to pull in data on all the English Wikipedia mainspace and userspace revisions that don't have score data already. That's used for the recent activity page: https://dashboard.wikiedu.org/recent-activity/dyk

You could maybe extract the code that deals with parsing the ORES output and getting the scores. ArticleRatingExtractor does that for the page assessments; you could similarly make an OresDataExtractor.

app/controllers/revision_feedback_controller.rb
ragesoss: The current flow is such that the article always exists, but the client doesn't have the id if the article was just added. The part of the app that creates the Article records is pretty independent from this, so it's better if we can avoid coupling it. It's also just not necessary to rely on the article existing; the data flow is simpler if we use article title as the main.

spec/controllers/revision_feedback_controller_spec.rb
```diff
@@ -3,28 +3,34 @@
...
+            .with(have_key('feature.enwiki.revision.cite_templates'))
+          expect_any_instance_of(RevisionFeedbackService).to receive(:feedback)
+            .and_return(have_at_least(1))
+          get :index, params: params
+        end
+      end
+
+      it 'assigns feedback properly' do
+        feedback = assigns(:feedback)
+        expect(feedback.length).to be >= 1
```
ragesoss: These specs could use some comments.

It took me a while to realize that these tests — I think — don't depend on the the state of the latest revision of article 27697087. (They just depend on it existing and ORES returning the normal features hash for it.)

Adding what the article is, and explaining in a comment how this is just acting as an integration test with `RevisionFeedbackController` and `RevisionFeedbackService` — but without testing the particular revision-dependent results of either one — would be helpful.

spec/controllers/revision_feedback_controller_spec.rb
```diff
@@ -3,28 +3,34 @@
 
 describe RevisionFeedbackController do
   describe '#index' do
-    let(:params) { { rev_id: 123456 } }
-    let(:revision) { build(:revision, mw_rev_id: 123456) }
+    let(:params) { { article_id: 1 } }
```
ragesoss: Better to do this as `article_id: article.id`.

ragesoss:
Per discussion in the checkin today, this can be merged once the feedback link is fixed. You can point it to something like `/feedback?subject=Article Viewer — ${articleTitle}`.

For follow-up work, we'll want to reword the messaging to avoid jargon like "ORES" and to explain things like "FA" (Featured Article), change the language on the feedback button (to avoid overlap of the meanings of "feedback" and "feature"), move the button to immediately after the feedback.

It also would be nicer to use a React-based form to send feedback instead of the current UI of opening a new tab for the feedback form, although this is not urgent.

app/views/revision_feedback/index.html.haml
```diff
@@ -1,6 +1,12 @@
 %section
   %h1
     Suggestions
+  %h4
+    Rating
+  %p
+    = @rating
+  %h4
+    More Suggestions
```
ragesoss: Any reason to keep the html version of this, since we're only using the json one?

keer25: We can make it admin only for referring to the suggestions on receiving feedback

ragesoss: Ah, I see... so that's why you included that URL as the subject!

PR #1305, opened 2017-06-18
 Add info mouseovers to explain Assign and Review buttons within the MyArticles component Resolve #1044

### LINKED ISSUE #1044: My Articles buttons should have info mouseovers to explain Assign and Review
These buttons show up for students when they join a course, but they just say 'Assign' and 'Review', and what that means isn't apparent.

Within the `MyArticles` component, these two `AssignButton` instances should have mouseovers that explain the intended use of each.

sejalkhatri:
Also adding OnClickOutside(higher order component) to AssignButton instance within the MyArticles component.

### DIFF
--- a/app/assets/javascripts/components/overview/my_articles.jsx
+++ b/app/assets/javascripts/components/overview/my_articles.jsx
@@ -51,6 +51,7 @@ const MyArticles = React.createClass({
               student={this.props.current_user}
               assignments={assigned}
               prefix={I18n.t('users.my_assigned')}
+              tooltip_message="Assign yourself an article to edit"
             />
             <AssignCell
               id="user_reviewing"
@@ -62,6 +63,7 @@ const MyArticles = React.createClass({
               student={this.props.current_user}
               assignments={reviewing}
               prefix={I18n.t('users.my_reviewing')}
+              tooltip_message="Assign yourself an article to review"
             />
           </div>
         </div>
--- a/app/assets/javascripts/components/overview/my_articles.jsx
+++ b/app/assets/javascripts/components/overview/my_articles.jsx
@@ -51,7 +51,7 @@ const MyArticles = React.createClass({
               student={this.props.current_user}
               assignments={assigned}
               prefix={I18n.t('users.my_assigned')}
-              tooltip_message="Assign yourself an article to edit"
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/overview/my_articles.jsx
```diff
@@ -51,6 +51,7 @@ const MyArticles = React.createClass({
               student={this.props.current_user}
               assignments={assigned}
               prefix={I18n.t('users.my_assigned')}
+              tooltip_message="Assign yourself an article to edit"
```
ragesoss: These tooltip messages should go through I18n.

`An assigned article is one that you plan to create or improve.`

app/assets/javascripts/components/overview/my_articles.jsx
```diff
@@ -62,6 +63,7 @@ const MyArticles = React.createClass({
               student={this.props.current_user}
               assignments={reviewing}
               prefix={I18n.t('users.my_reviewing')}
+              tooltip_message="Assign yourself an article to review"
```
ragesoss: `This if for keeping track of which article(s) you plan to peer review or provide feedback on.`

ragesoss:
After playing around with this for a little bit, I think the tooltip ought to be separated from the button. Like, we could have the tooltip indicator to the right of the button, and only mousing over the indicator itself would trigger the tooltip. That would avoid the awkward interaction between the clicking UI and the mouseover UI. What do you think?

sejalkhatri:
Once the user has assigned himself/herself an article then it doesn't appear, So the current pattern should work.

ragesoss:
What I'm talking about is when the tooltip is rendering at the same time as the dialog (underneath it, so it can't be read, but the top edge is visible.
both

ragesoss:
Could be fixed with an extra conditional...

```
      if (this.props.tooltip_message && !this.props.is_open) {
        tooltipIndicator = (
          <span className="tooltip-indicator"></span>
          );
        tooltip = (
          <div className="tooltip">
            <p>
              {this.props.tooltip_message}
            </p>
          </div>
      );
```

sejalkhatri:
Ok! I understand that, will add the extra condition

PR #1310, opened 2017-06-22
Enabling Programs & Events Dashboard to make automatic edits on connected wikis

medhabansal:


medhabansal:
Things done in initial draft :
- Created associated templates for P&E Dashboard.
- Created a mapping class which returns template names according to dashboard.
- Generalized template usage in edit and output classes.

lib/wiki_edit_mappings.rb
```diff
@@ -0,0 +1,30 @@
...
+      "instructor" => "course instructor",
+      "course_assignment" => "course assignment",
+      "table" => "students table",
+      "table_row" => "students table row",
+      "course" => "course details",
+      "timeline" => "start of course timeline"
+    },
+    "outreachdashboard.wmflabs.org" => {
+      "editor" => "participant editor",
+      "instructor" => "program instructor"
```
ragesoss: Missing comma

lib/wiki_edit_mappings.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+# Class returns template mappings for WikiEdu and P&E Dashboards, respectively
+class WikiEditMappings
```
ragesoss: This will let us get started on English Wikipedia for P&E Dashboard, but we'll want to be able to use different templates for each language / project. (For dashboard.wikiedu.org, it will remain enabled only on English Wikipedia.)

ragesoss: We can see how far we get with hard-coding the template names; I think it's an okay way to do things for now.

Alternatives include putting the mapping into a config file or locale file, or putting it into the database (which would let us make a simple frontend for setting the templates for a particular wiki).

lib/wiki_course_edits.rb
```diff
@@ -22,7 +23,7 @@ def initialize(action:, course:, current_user:, **opts)
   # set of participants, articles, timeline, and other details.
   # It simply overwrites the previous version.
   def update_course(delete: false)
-    return unless @course.submitted && @course.slug
```
ragesoss: We still need a check for unsubmitted courses for Wiki Ed Dashboard.

This is something we could delegate to the specific Course subclasses via something like `course.wiki_course_page_enabled?`. That would check for submitted  status for ClassroomProgramCourse, and return true for others. I'm not sure why the check for a slug is here, as it shouldn't be possible to have a course without one.

ragesoss:
Per discussion today, in addition to implementing the .yml configurations for template names, next steps include:

* A clear way to configure which wikis we will and won't make edits on
* Tests for that behavior, showing that an edit is attempted with the right template for one wiki, and no edit is attempted for the same event on another wiki that doesn't have edits configured/enabled

config/templates/dashboard.wikiedu.org_en.yml
```diff
@@ -0,0 +1,10 @@
+templates:
```
ragesoss: let's change this to `default` instead of `templates`

lib/wiki_assignment_output.rb
```diff
@@ -1,12 +1,15 @@
 # frozen_string_literal: true
 require './lib/wiki_api'
+require "#{Rails.root}/lib/wiki_edit_mappings"
```
ragesoss: remove this and the file

lib/wiki_course_edits.rb
```diff
@@ -12,17 +13,20 @@ def initialize(action:, course:, current_user:, **opts)
     @course = course
     # Edits can only be made to the course's home wiki through WikiCourseEdits
     @home_wiki = course.home_wiki
+    return unless @home_wiki.edits_enabled?
     @wiki_editor = WikiEdits.new(@home_wiki)
     @dashboard_url = ENV['dashboard_url']
     @current_user = current_user
+    template_file_path = "config/templates/#{@dashboard_url}_#{@home_wiki.language}.yml"
+    @templates = YAML.load_file(Rails.root + template_file_path)
```
ragesoss: Pass @templates to WikiAssignmentOutput and WikiCourseOutput, instead of redoing this in each file.

lib/wiki_course_edits.rb
```diff
@@ -117,12 +122,14 @@ def repost_with_sanitized_links(wiki_title, wiki_text, summary, blacklist)
 
   def add_course_template_to_instructor_userpage(instructor)
     user_page = "User:#{instructor.username}"
-    template = "{{course instructor|course = [[#{@course.wiki_title}]] }}\n"
+    instructor_template_key = @templates['templates']['instructor']
```
ragesoss: Let's extract all these _key variables into a method, `template_name(key)`, which will raise an error if the key is missing.

lib/wiki_course_output.rb
```diff
@@ -1,14 +1,18 @@
 # frozen_string_literal: true
 require "#{Rails.root}/lib/wikitext"
 require "#{Rails.root}/lib/course_meetings_manager"
+require "#{Rails.root}/lib/wiki_edit_mappings"
```
ragesoss: remove

lib/wiki_course_output.rb
```diff
@@ -90,14 +93,16 @@ def course_week(week, week_number)
       week_output += content_block(block)
     end
 
-    week_output += "{{end of course week}}\r"
+    week_end_template = @templates['templates']['end_of_week']
```
ragesoss: Instead of a method in the other class, we should make a module, like `WikiOutputTemplates`, so we can reuse it here.

spec/lib/wiki_course_edits_spec.rb
```diff
@@ -1,23 +1,25 @@
-# frozen_string_literal: true
+;# frozen_string_literal: true
```
ragesoss: remove this character

ragesoss: @medhabansal one last thing ^

PR #1311, opened 2017-06-22
Display - "Data not updated" when Article pageViews stats is zero

sejalkhatri:


app/assets/javascripts/components/overview/course_stats.jsx
```diff
@@ -2,9 +2,24 @@ import React from 'react';
...
 
 const CourseStats = ({ course }) => {
+  let viewData;
   if (course.upload_usages_count === undefined) {
     return <div className="stat-display"></div>;
   }
+  if (course.view_count === '0') {
+    viewData = (
+      <div className="stat-display__data">
+        Data not updated
```
ragesoss: Put this in i18n. "No view data available"

sejalkhatri: even some styling tweaks are yet to be made

app/assets/javascripts/components/overview/course_stats.jsx
```diff
@@ -2,9 +2,23 @@ import React from 'react';
...
+  if (course.view_count === '0') {
+    viewData = (
+      <div className="stat-display__data">
+        {I18n.t('metrics.view_data_unavailable')}
+      </div>
+    );
+  } else {
+    viewData = (
+      <div className="stat-display__value">
+        course.view_count
```
ragesoss: Needs braces, or it gets treated as a string literal.

One of the feature specs caught this, but this would be a good opportunity to add a javascript component test.

sejalkhatri:
The enzyme test are not functioning currently, will fix them in next commit cycle.

ragesoss:
On taking another look at this in action, I think we should only put in that message if there are some articles edited.

If articles edited is zero, then it makes more sense to keep the row of all zeros, and only add the 'no data' message if there are other stats and so the user would be expecting views as well.

ragesoss:
The feature spec is broken with that last commit. We are now testing that behavior in the component spec, so I'd be fine with deleting the feature spec instead of fixing it. Do you think it's a valuable test for anything that isn't covered by the javascript one?

sejalkhatri:
yes it is because currently we don't test all the stats in the component spec but we can add that and then delete the feature spec

PR #1312, opened 2017-06-23
Change the styling for DayPicker div in the Create New Course Page - Resolve #1094

### LINKED ISSUE #1094: The pop-up for date selection in course_creator page disappears while scrolling down using the scroll bar.
The course_creator Page date selection pop-up vanishes if instructor scrolls down(using the scroll bar) for determining the dates.

It works if he/she clicks the pop-up and then uses arrow keys for scrolling down but then the year field in the pop-up responds to the arrow key behavior(decrements with the use of up arrow key and increments with the use of down arrow key)

**Behavior1: The pop-up appears on clicking the Start date field.**
before_scrolling
 
**Behavior2: It vanishes on using the Scroll bar.**
using_the_scroll_bar

**Behavior3: The Year field in the pop-up responds to the arrow key behavior.**
after_scrolling

**Issue observed in**
Operating System: Ubuntu 16.04 (64-bit)
Browsers: Chromium[Version 51.0.2704.79 ] and Firefox[Version 48.0]

sejalkhatri:
Avoid Scrolling down while using pop-up for date selection in course_creator page by making some styling tweaks.

### DIFF
--- a/app/assets/javascripts/components/common/date_picker.jsx
+++ b/app/assets/javascripts/components/common/date_picker.jsx
@@ -5,7 +5,7 @@ import InputMixin from '../../mixins/input_mixin.js';
 import Conditional from '../high_order/conditional.jsx';
 import CourseDateUtils from '../../utils/course_date_utils.js';
 
-const DatePicker = OnClickOutside(React.createClass({
+const DatePicker = React.createClass({
   displayName: 'DatePicker',
 
   propTypes: {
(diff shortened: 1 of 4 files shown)


ragesoss:
Looks good to me. Does this fix the issue for you?

ragesoss:
On taking another look at this in action, I think we should only put in that message if there are some articles edited.

If articles edited is zero, then it makes more sense to keep the row of all zeros, and only add the 'no data' message if there are other stats and so the user would be expecting views as well.

PR #1319, opened 2017-07-03
Improving the Feedback Feature

keer25:


app/assets/javascripts/components/common/feedback.jsx
```diff
@@ -37,16 +44,32 @@ const Feedback = React.createClass({
...
+  handleSubmit(event) {
+    const subject = `/revision_feedback?title=${this.titleParam()}`;
+    const body = event.target.value;
+    this.setState({ feedbackSent: 'sending' });
+    API.postFeedbackFormResponse(subject, body).then(() => {
+      this.setState({ feedbackSent: true });
+    });
+    event.preventDefault();
+  },
+
```
keer25: The submit button posts to Rails API to create Feedback Form Response. This approach doesn't check if the request was successful just that it receives a response. I couldn't figure out a way to figure out if the response was successful. The data received on success is a HTML page.

ragesoss: In the feedback form controller, `create` redirects to the 'confirmation' page. What you can probably do is add a json version of the `confirmation` view and change the url in api.js to add '.json', and then you'll get that back instead of html.

The controller flow is such that there's no explicit error handling, but it will return an error message and code if something on the rails side throws an error. That should be enough to fail the promise (which you could handle with a `.catch()` after the `.then()`).

keer25: The catch() function was what I was looking for.

keer25:
The failure on revision_score_importer_spec passes for me locally. Not sure what is the problem

app/assets/javascripts/components/assignments/assignment.jsx
```diff
@@ -47,6 +48,11 @@ const Assignment = React.createClass({
     if (assignees.length) { assignees.pop(); }
     if (reviewers.length) { reviewers.pop(); }
 
+    let feedback;
+    if (this.props.assignmentGroup.length === 1 || article.article_id) {
```
ragesoss: This could use a comment to explain the intention:

If the article exists (and therefore has an article id) then show Feedback, which will be based on the current version of the article. If the article does not exist, then Feedback is based on a user's sandbox, so we only want to show it if there is a single user assigned; otherwise, we don't know which sandbox to base it on.

app/controllers/revision_feedback_controller.rb
```diff
@@ -16,13 +16,13 @@ def index
...
-
-    query = { prop: 'revisions', pageids: @article.mw_page_id, rvprop: 'ids' }
+    query = { prop: 'revisions', titles: params['title'], rvprop: 'ids' }
     @wiki = Wiki.find_by(language: 'en', project: 'wikipedia')
     response = WikiApi.new(@wiki).query(query)
-    results = response&.data
-    revisions = results.dig('pages', @article.mw_page_id.to_s, 'revisions')
+    page = response&.data['pages']
+    # The Page ID is the only key in the response
+    page_id = page.keys()[0];
```
ragesoss: Parens are not necessary here.

app/controllers/revision_feedback_controller.rb
```diff
@@ -16,13 +16,13 @@ def index
...
   def set_latest_revision_id
-    @article = Article.find(params[:article_id])
-
-    query = { prop: 'revisions', pageids: @article.mw_page_id, rvprop: 'ids' }
+    query = { prop: 'revisions', titles: params['title'], rvprop: 'ids' }
     @wiki = Wiki.find_by(language: 'en', project: 'wikipedia')
     response = WikiApi.new(@wiki).query(query)
-    results = response&.data
-    revisions = results.dig('pages', @article.mw_page_id.to_s, 'revisions')
+    page = response&.data['pages']
```
ragesoss: The safe navigation operator here doesn't help. If response is nil, doing `page.keys` below will still cause an error.

PR #1320, opened 2017-07-05
Test cases for automatic edits

medhabansal:


app/models/wiki.rb
```diff
@@ -104,6 +104,10 @@ def ensure_wiki_exists
     raise InvalidWikiError unless base_url == "https://#{servername}"
   end
 
+  def edits_enabled?
+    ENV["edit_#{domain}"]
```
ragesoss: My tests were failing because of this. We need to add the default of en.wikipedia.org to the test environment.

ragesoss: This will return a truthy value if it is set to "false" in application.yml. We should follow the convention used elsewhere in the app for config switches: `ENV["edit_#{domain}"] == 'true'. That way, it always returns true or false (as a `?` method should) and only returns true if it's unambiguously set to 'true'.

lib/wiki_course_edits.rb
```diff
@@ -120,13 +124,14 @@ def repost_with_sanitized_links(wiki_title, wiki_text, summary, blacklist)
   def add_course_template_to_instructor_userpage(instructor)
     user_page = "User:#{instructor.username}"
 
-    instructor_template_key = WikiEditMappings.get_template('instuctor')
+    instructor_template_key = @templates['templates']['instuctor']
```
ragesoss: 'instructor' mispelled

spec/lib/wiki_course_edits_spec.rb
```diff
@@ -1,23 +1,25 @@
-# frozen_string_literal: true
+;# frozen_string_literal: true
```
ragesoss: stray character

PR #1322, opened 2017-07-07
Add explanation for the submitted field in the course details section

sejalkhatri:


app/assets/javascripts/components/overview/submitted_selector.jsx
```diff
@@ -35,7 +35,15 @@ const SubmittedSelector = React.createClass({
     if (this.props.editable) {
       selector = (
         <div className="form-group">
-          <label htmlFor={this.state.id}>Submitted:</label>
+          <span htmlFor={this.state.id}>Submitted:</span>
```
ragesoss: Let's move this to i18n.

config/locales/en.yml
```diff
@@ -432,6 +432,7 @@ en:
     suggestions_feedback: Are these suggestions helpful?
     user_role: Role
     course_details: Course Details
+    course_submitted: Edits of course don't happen until you submit your course
```
ragesoss: How about...

"Submitted status is used for education program courses where a course plan must be approved before the project begins. It is only relevant for the ClassroomProgram course type."

PR #1324, opened 2017-07-07
Improve the page layout for course cloned page and add validation for the title input in the details section of course page

sejalkhatri:
Things covered:
1) Improve the page layout for course cloned page
2) Added validation for the title,school and term input in the details section of course page
3) Add info icon for the when field in program cloned page

### DIFF
--- a/app/assets/javascripts/components/course_creator/course_creator.jsx
+++ b/app/assets/javascripts/components/course_creator/course_creator.jsx
@@ -35,7 +35,7 @@ const getState = () => {
 const CourseCreator = React.createClass({
   displayName: 'CourseCreator',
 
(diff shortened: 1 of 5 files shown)


ragesoss:
@sejalkhatri the 'term' and 'school'; fields also need to be validated with the same regex.

ragesoss:
One thing to note is that this creates a misleading error message if you blank the Term field: it says invalid characters.

If a course has already been created with no term, then the other fields can be changed without adding a term. But if the user puts input into that field and then removes it, or if there is already a term and the user tries to remove it, this will be invalid.

I think this is okay, as it will save problems with slug collisions, but we may want to revisit the validation on this some time in the future if it causes confusion.

PR #1333, opened 2017-07-18
Instructor delete unsubmitted timeline

### LINKED ISSUE #955: Instructors should have an option to delete the entire timeline of an unsubmitted course
Many times, instructors need to delete each week in a timeline so that they can go through the wizard again. For unsubmitted courses, we should provide an easy way to reset the timeline and launch the wizard fresh. (For submitted courses, the assumption is that they will only change incrementally; we don't want to miss total rewrites after a course has been approved.)

hin101:
Closes #955


app/controllers/weeks_controller.rb
```diff
@@ -6,4 +6,12 @@ def destroy
     Week.find(params[:id]).destroy
     render plain: '', status: :ok
   end
+
+  def delete_multiple
```
ragesoss: Instead of `delete_multiple` that operates on a set of week ids, I think this should operate at the level of the course, taking a course_id param and doing something like course.weeks.destroy_all. That way, there's less danger of leaving the data in an inconsistent state, if the browser doesn't have the up-to-date set of week ids (which may happen if another user was editing the timeline in the meantime). The addWeek action to create a week on the client-side sets a placeholder ID using Date.now(), so it's also possible that there would be a week where the id differs from what is in the database.

Before the action, it should also verify that the current user has permission. (I notice that we don't do this with the existing `destroy` method, which we should do.)

ragesoss:
@hin101 WeeksController test is failing now.

spec/controllers/courses_controller_spec.rb
```diff
@@ -545,4 +545,20 @@
...
+    let!(:user) { create(:admin) }
+
+    before do
+      allow(controller).to receive(:current_user).and_return(user)
+      allow(controller).to receive(:user_signed_in?).and_return(true)
+    end
+
+    it 'deletes all the weeks' do
+      delete :delete_all_weeks, params: { id: course.slug }
+      expect(course.weeks.count).to eq(0)
```
ragesoss: I don't think this course has any weeks to begin with.

hin101: Should I remove this test or assign a week to the course then test?

ragesoss: I think adding two weeks to the course would be the best option.

ragesoss:
Thanks @hin101. I'm happy with the architecture, and I've given it a spin locally. It needs a little bit more on the user flow side:

* The action needs a confirmation step. We don't want to immediately delete the timeline if the user misclicks that button.
* Upon success, the page should reload. This will clear out the gradeables store.
* The button should not appear if the course has been submitted (`this.props.course.submitted`).

ragesoss:
Rebased and merged after a couple final tweaks. Thanks @hin101!

Issue #1330, opened 2017-07-13
Training Libraries index should have links directly to individual trainings


ragesoss:
It currently names them, but does not link to them.

vdineva:
Created a PR: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1336
Attaching a screenshot of before and after. One thing I wasn't quite sure about is if the default underline for the link should be there and about the spacing between the 'Instructors' and 'Students' and the block below them. I'd be more than happy to address any visual changes
BEFORE
![1330-before](https://user-images.githubusercontent.com/791418/28571253-ffa5bafe-710f-11e7-95d6-4f46d6647d12.png)
AFTER:
![1330-after](https://user-images.githubusercontent.com/791418/28571258-035d8172-7110-11e7-8201-5e39f9011aec.png)

ragesoss:
@vdineva Thanks! I think the newly introduced spacing between the header and the list of modules should be removed. The link styling looks good.

PR #1336, opened 2017-07-25
Issue #1330 Training Libraries index should have links directly to in…

vdineva:
…dividual trainings

### DIFF
--- a/app/views/training/index.html.haml
+++ b/app/views/training/index.html.haml
@@ -9,8 +9,10 @@
               = library.slug.humanize.titleize
             %span.icon-container
               %i.action-card-icon.icon.icon-rt_arrow
-          .action-card-text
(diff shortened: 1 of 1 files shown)


ragesoss:
I played around with it a little bit, and it looks like it's going to be a little tricky to fix the styling issue. The current design relies on the whole library card being a link, and the border gets added around that whole card when it's clicked.

Adding links within the card creates illegal link nesting, which is why the space gets added (and which makes the border only show up around the header when the link is clicked).

I'm okay with changing the clicked-link styling on this page to make it work, but we should make sure it does not disrupt the styling on other pages (like the pages for the individual training libraries).

ragesoss:
Closing this; I can reopen if you're still working on it.

PR #1339, opened 2017-07-31
Add Custom Messages Support

keer25:
Users can add a feedback to an article in addition to the existing feedback


app/assets/javascripts/components/common/feedback.jsx
```diff
@@ -44,6 +45,19 @@ const Feedback = React.createClass({
...
+  handleFeedbackInputChange(event) {
+    this.setState({ feedbackInput: event.target.value });
+  },
+
+  handleFeedbackSubmit(event) {
+    const feedback = this.state.feedbackInput;
+    this.setState({ feedbackInput: '' });
+    API.createCustomFeedback(this.props.assignment.id, feedback).then(() => {
+      this.props.feedback[this.props.assignment.id].custom.push(<li>{feedback}</li>);
+    });
```
keer25: This doesn't exactly give the required behavior. If the feedback is stored as a state it will be easier to update but there is no callback to update state after mapStatetoProps (after fetching from API).

ragesoss: You should move the API call into FeedbackActions, and then dispatch an action that includes the feedback input as data. The action would be something like in alert_actions:

```
export function createCustomFeedback(input) {
  return function (dispatch) {
    dispatch({ type: types.CUSTOM_FEEDBACK_SUBMITTED, data: input }); // handle this in the reducer to add the new feedback into the set of already-received feedback. connect() will then pass in the updated feedback prop and cause the component to re-render.
    return API.createCustomFeedback(input)
      .then(() => (dispatch({ type: types.CUSTOM_FEEDBACK_CREATED }))) // optional, to confirm successful creation of feedback
      .catch(response => (ApiFailAction.fail(response)));
  };
}
```

app/assets/stylesheets/feedbacks.css
```diff
@@ -0,0 +1,4 @@
+/*
+  Place all the styles related to the matching controller here.
+  They will automatically be included in application.css.
+*/
```
ragesoss: This shouldn't be included unless it's being used.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,9 @@
+class FeedbacksController < ApplicationController
```
ragesoss: Why a separate controller instead of putting this create method into RevisionFeedbackController?

app/models/application_record.rb
```diff
@@ -0,0 +1,3 @@
+class ApplicationRecord < ActiveRecord::Base
```
ragesoss: Did this get automatically generated via a console command? I'm not sure what this does, or whether it makes it any different from the rest of the models which inherit from ActiveRecord::Base directly.

app/models/feedback.rb
```diff
@@ -0,0 +1,3 @@
+class Feedback < ApplicationRecord
```
ragesoss: I think this class name is not specific enough. Maybe AssignmentFeedback or AssignmentSuggestion ?

That would help reduce confusion with the unrelated 'FeedbackFormResponse', which is a more general set of "feedback" records from many different parts of the application.

app/views/feedbacks/create.html.haml
```diff
@@ -0,0 +1,2 @@
+%h1 Feedbacks#create
+%p Find me in app/views/feedbacks/create.html.haml
```
ragesoss: Remove this unused boilerplate.

app/controllers/feedbacks_controller.rb
```diff
@@ -0,0 +1,9 @@
...
+  def create
+  	Feedback.new(form_params).save
+  end
```
ragesoss: This should return an 'okay' json response of some sort, if it doesn't already.

app/models/assignment.rb
```diff
@@ -22,6 +22,7 @@ class Assignment < ActiveRecord::Base
   belongs_to :course
   belongs_to :article
   belongs_to :wiki
+  has_many :feedbacks
```
ragesoss: If these are *only* linked to an assignment, they should be destroyed if the assignment is destroyed.

app/models/feedback.rb
```diff
@@ -0,0 +1,3 @@
...
+	belongs_to :assignment
```
ragesoss: We should also record the user who creates this.

app/controllers/feedbacks_controller.rb
keer25: If edit and delete features are added I think it is better to have a separate controller.

app/models/application_record.rb
keer25: Yeah, I used a generator to create the Feedback model. I don't know what this does too.

ragesoss: Looks like the idea is to add functionality to that instead of monkey-patching ActiveRecord::Base: http://blog.bigbinary.com/2015/12/28/application-record-in-rails-5.html

But if we switch to that at some point, we should do it separately, for all classes at once.

keer25:
Much more work to be done for the delete option. I will add a user column and show delete only for their own comments.

ragesoss:
@keer25 the migrations are not working with this latest one? What errors are you getting?

ragesoss:
If you're getting migration errors, one option is to reset your own database, and then rewrite the migration as a single one. (That model also still needs to be renamed.)

keer25:
I don't have any migration errors now atleast in my system

keer25:
@ragesoss I have incorporated the changes requested previously. Do review it.

keer25:
I have replaced the feedback form as suggested aboive User Suggestions. But still it is confusing the two input boxes look similar. 
Do you think it is better to go back to the button for giving feedback like previouly

PR #1340, opened 2017-08-01
Updating YAML files, template names for edits

medhabansal:


config/application.production.yml
```diff
@@ -31,6 +31,8 @@ campaign_spring_2015: "Wikipedia:Education_program/Dashboard/course_ids"
 campaign_fall_2014: "Wikipedia:Education_program/Dashboard/Fall_2014_course_ids"
 
 default_campaign: "spring_2016"
+edit_en.wikipedia.org: "true"
+edit_pt.wikipedia.org: "true"
```
ragesoss: This file represents the dashboard.wikiedu.org production config. I'll fork this file so we have the outreachdashboard.wmflabs.org config in the repo as well.

ragesoss:
Looks like this breaks the WikiCourseOutput spec.

PR #1343, opened 2017-08-08
Blocked edits alert

### LINKED ISSUE #1176: Dashboard should detect, and create alerts for, blocked edits
Occasionally, automated edits from the dashboard are blocked. This can happen either because of an autoblock, where-in subsequent users of the same IP as an initial blocked user get automatically blocked, or because of a range block that affects the IP of the dashboard's server (which is currently on Linode).

Blocked edits should trigger an alert with an email to the dashboard's `technical_help_staff` user, but only up to once per day. (We don't want an email flood during such a blockage incident.)

hin101:
Closes #1176


hin101:
How do I squash the commits above?

ragesoss:
Maybe you can just do "git fetch upstream/master" and "git rebase".

app/models/blocked_edits_alert.rb
```diff
@@ -0,0 +1,9 @@
+class BlockedEditsAlert < Alert
+  def main_subject
+    "#{article.title}"
```
ragesoss: There is no article for this alert; it gets created with only a user_id and target_user_id.

app/models/blocked_edits_alert.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+
+  def url
+    article_url(article)
```
ragesoss: Likewise, there's no article so this doesn't make sense.

app/models/blocked_edits_alert.rb
ragesoss: Maybe instead, "Edit by #{user.username} was blocked"

app/models/blocked_edits_alert.rb
ragesoss: For url, you could use `user_profile_url`.

lib/blocked_edits_monitor.rb
```diff
@@ -0,0 +1,27 @@
...
+    alert = Alert.create!(type: 'BlockedEditsAlert',
+                          user_id: user.id,
+                          target_user_id: technical_help_staff&.id)
+    alert.email_target_user
+  end
+
+  def alert_already_exists?
+    Alert.exists?(['created_at >= ? AND type = ?',
+                   (Time.now - (8 * 60 * 60)).to_s,
+                   'BlockedEditsAlert'])
```
ragesoss: This could be simpler:
```ruby
BlockedEditsAlert.where('created_at >= ?', 1.day.ago).exists?
```

lib/blocked_edits_monitor.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def create_alert(user)
+    return if alert_already_exists?
+    alert = Alert.create!(type: 'BlockedEditsAlert',
+                          user_id: user.id,
+                          target_user_id: technical_help_staff&.id)
+    alert.email_target_user
+  end
+
+  def alert_already_exists?
```
ragesoss: This method should be renamed and/or documented, to make clear that what we're checking for is whether any recent BlockedEditAlerts already exist.

lib/blocked_edits_monitor.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class BlockedEditsMonitor
```
ragesoss: Maybe instead, 'BlockedEditsReporter', since unless like the other similar classes, this one does not do any monitoring; it's job is only to create and email new alerts.

spec/lib/wiki_edits_spec.rb
```diff
@@ -58,6 +58,16 @@
     expect(User.first.wiki_token).to eq('invalid')
   end
 
+  it 'handles edit requests that are blocked' do
```
ragesoss: These should also expect the BlockedEditsMonitor (or BlockedEditsReporter) to receive `:create_alerts_for_blocked_edits`, which tests the interaction between WikiResponse and the new alert code.

ragesoss:
Merged after a rebase.

PR #1349, opened 2017-08-17
Add email alerts for blocked edits

### LINKED ISSUE #1176: Dashboard should detect, and create alerts for, blocked edits

ragesoss:
Closes #1176

### DIFF
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -30,6 +30,7 @@ class Alert < ActiveRecord::Base
   ALERT_TYPES = %w[
     ActiveCourseAlert
     ArticlesForDeletionAlert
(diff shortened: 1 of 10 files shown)


ragesoss:
@hin101 I checkout out your branch and then did `rebase master` to get rid of all the merge commit noise.

PR #1359, opened 2017-08-24
add real_name attribute to CoursesUsers

mattwarren10:


mattwarren10:
Submitted this before reading your comments. I'll close and resubmit later.

ragesoss:
You can feel free to leave a work-in-progress PR open (just mark it in the title as WIP) if you like. The benefit is that the tests will run on travis-ci each time you add new commits.

PR #1361, opened 2017-08-24
CMU Fall 2017 experiment

ragesoss:
This adds code for inviting instructors by email to participate in an experiment, and for recording whether they have opted in or out.


ragesoss:
Here's what the invitation looks like:

email

PR #1362, opened 2017-08-25
[WIP] Add real_name attribute to CoursesUsers

mattwarren10:
The most recent commit on this branch is updated according to your comments. In the future, I'll submit PRs with WIP in the title and keep them open.


app/controllers/mass_enrollment_controller.rb
```diff
@@ -27,7 +27,7 @@ def add_users
   def add_user(username)
     user = find_or_import_user(username)
     return { failure: 'Not an existing user.' } unless user
-    JoinCourse.new(course: @course, user: user, role: CoursesUsers::Roles::STUDENT_ROLE).result
+    JoinCourse.new(course: @course, user: user, role: CoursesUsers::Roles::STUDENT_ROLE, real_name: user.real_name).result
```
ragesoss: As with UsersController, this is used by one user adding another user to a course, so if they do not provide a real name, we don't want to associate one with the new CoursesUsers record. This can stay the same as before (with a change to the mass enrollment interface required before it makes sense to make any changes to allow real names here).

PR #1366, opened 2017-08-28
[WIP] Add real_name attribute to CoursesUsers

mattwarren10:
Hopefully this PR is the way it needs to be. No changes to gemfile.lock or package.json. If I am missing something, let me know. Thanks.

### DIFF
--- a/app/services/join_course.rb
+++ b/app/services/join_course.rb
@@ -6,10 +6,11 @@ require "#{Rails.root}/lib/chat/rocket_chat"
 class JoinCourse
   attr_reader :result
 
-  def initialize(course:, user:, role:)
+  def initialize(course:, user:, role:, real_name: nil)
     @course = course
     @user = user
     @role = role
+    @real_name = real_name
     process_join_request
   end
 
@@ -59,7 +60,8 @@ class JoinCourse
     CoursesUsers.create(
       user_id: @user.id,
       course_id: @course.id,
-      role: @role
+      role: @role,
+      real_name: @real_name
     )
   end
 
--- a/app/views/courses/_users.json.jbuilder
+++ b/app/views/courses/_users.json.jbuilder
@@ -17,7 +17,7 @@ json.users course.courses_users.eager_load(:user, :course) do |cu|
   # an instructor of the course.
   # Emails and names of greeters are shown to all users
   if show_email_and_real_name || cu.user.greeter
-    json.real_name cu.user.real_name
+    json.real_name cu.real_name
     # Student emails are not shown to anyone.
     json.email cu.user.email unless cu.role == CoursesUsers::Roles::STUDENT_ROLE
   end
--- a/spec/services/join_course_spec.rb
+++ b/spec/services/join_course_spec.rb
@@ -23,6 +23,34 @@ describe JoinCourse do
     enroll_as_instructor
   end
 
+  context 'with real name' do    
+    let(:course) { create(:basic_course) }
+    let(:subject) do
+      described_class.new(course: course, user: user,
+                          role: CoursesUsers::Roles::STUDENT_ROLE,
+                          real_name: 'student name')
+    end
+    it 'allows a course to be joined' do
+      result = subject.result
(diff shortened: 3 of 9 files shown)


app/views/courses/_users.json.jbuilder
```diff
@@ -4,7 +4,7 @@ show_email_and_real_name = user_signed_in? && current_user.role(course).positive
 
 json.users course.courses_users.eager_load(:user, :course) do |cu|
   json.call(cu, :character_sum_ms, :character_sum_us, :character_sum_draft, :role,
-            :recent_revisions, :content_expert, :program_manager, :contribution_url, :sandbox_url)
+            :recent_revisions, :content_expert, :program_manager, :contribution_url, :sandbox_url, :real_name)
```
ragesoss: Instead of adding it here, the call to cu.real_name should replace `cu.user.real_name` on line 20.

app/services/join_course.rb
```diff
@@ -5,11 +5,13 @@
 #= Adds a user to a course
 class JoinCourse
   attr_reader :result
+  attr_accessor :real_name
```
ragesoss: We don't expect to read or set :real_name except via JoinCourse.new, I suggest leaving this out. (You may have added it to facilitate testing, but it's not necessary.)

spec/services/join_course_spec.rb
```diff
@@ -23,6 +23,23 @@
     enroll_as_instructor
   end
 
+  context 'optionally with or without a real_name' do
```
ragesoss: Instead of modifying the subject within the individual tests, I think it would be clearer if we create a context for 'with real name' and 'without real name', and set the subject within that context so that it's more explicit that you're basically testing `JoinCourse.new` with and without the optional param. That will also let us avoid adding the attr_accessor that isn't otherwise used.

ragesoss:
I realized one more place that requires a change: `lib/course_creation_manager.rb` Currently it creates a CoursesUsers record directly, rather than use `JoinCourse`, but it can be changed to use `JoinCourse` with the real name (set from the user record). When a user creates a course, they are automatically added as the initially instructor, and this is equivalent to self-enrollment for students.

ragesoss:
I can resolve the merge conflict when I merge it. It's just another schema change that went to master in the meantime.

spec/services/join_course_spec.rb
```diff
@@ -23,6 +23,34 @@
     enroll_as_instructor
   end
 
+  context 'with real name' do    
+    let(:course) { create(:basic_course) }
```
mattwarren10: I wasn't quite sure which course you wanted to test this on, but I assumed a `basic_course` would work find. Let me know.

ragesoss: Testing it with `create(:course)` is the default pattern, but this is fine as well.

spec/controllers/users_controller_spec.rb
```diff
@@ -152,7 +152,8 @@
       before do
         create(:courses_user, user_id: user.id,
                               course_id: course.id,
-                              role: CoursesUsers::Roles::STUDENT_ROLE)
+                              role: CoursesUsers::Roles::STUDENT_ROLE,
+                              real_name: user.real_name)
```
ragesoss: Is this necessary?

spec/services/join_course_spec.rb
mattwarren10: I can change it if needed.

spec/controllers/users_controller_spec.rb
mattwarren10: In my local repo I don't have that. Maybe I need to force push.

spec/services/join_course_spec.rb
ragesoss: no need. :-)

spec/controllers/users_controller_spec.rb
mattwarren10: This change should be gone.

PR #1367, opened 2017-08-28
Adding Hamburger menu and enhancing usability of dashboard on Mobile devices- resolve #756

### LINKED ISSUE #756: Implement hamburger menu for mobile
@ragesoss Since the survey supports mobile, I was thinking that it would be a good idea to update the mobile design make it easier to log in and navigate to the survey. I'll be attempting to find specific actions and will log them into github. Thoughts?

Current mobile experience:
img_0515 png

sejalkhatri:
Making the Navigation menu accessible on small screen devices.


app/assets/javascripts/components/CustomLink.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+
+const CustomLink = React.createClass({
```
ragesoss: A top-level comment explaining the purpose of this component would be helpful, since it has a very generic name. It should point to which circumstances you want to use this instead of the non-custom `Link`.

app/assets/javascripts/components/hamburger_menu.jsx
```diff
@@ -0,0 +1,160 @@
+import React from 'react';
+import { slide as Menu } from 'react-burger-menu';
+import CustomLink from './CustomLink.jsx';
+
+const HamburgerMenu = React.createClass({
+  propTypes: {
+    rootUrl: React.PropTypes.string,
+    logoPath: React.PropTypes.string,
+    fluid: React.PropTypes.bool,
+    exploreclass: React.PropTypes.string,
```
ragesoss: All these props should ideally be the standard javaScriptCase, with first letters of words after the first capitalized.

app/assets/javascripts/components/hamburger_menu.jsx
```diff
@@ -0,0 +1,160 @@
+import React from 'react';
+import { slide as Menu } from 'react-burger-menu';
+import CustomLink from './CustomLink.jsx';
+
+const HamburgerMenu = React.createClass({
```
ragesoss: Looks like this could be a stateless functional component.

app/assets/javascripts/utils/hamburger_menu.js
```diff
@@ -0,0 +1,11 @@
+// jQuery for user profile pages
```
ragesoss: This comment looks incorrect.

app/helpers/application_helper.rb
```diff
@@ -2,11 +2,33 @@
...
 
+  def permissions
+    if Features.wiki_ed? && current_user&.permissions == User::Permissions::NONE
+      'true'
+    else
+      'false'
+    end
+  end
+
+  def languageswitcherenabled
```
ragesoss: Nonstandard method name.

Why does this need to return strings? And if strings are really necessary, `Features.enable_language_switcher?.to_s`

db/schema.rb
```diff
@@ -12,416 +12,416 @@
 
 ActiveRecord::Schema.define(version: 20170602231912) do
 
-  create_table "alerts", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
-    t.integer  "course_id"
-    t.integer  "user_id"
-    t.integer  "article_id"
-    t.integer  "revision_id"
-    t.string   "type"
+  create_table "alerts", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
```
ragesoss: I think this is unrelated to the rest, and just got included by accident. Shouldn't be any schema changes.

app/helpers/application_helper.rb
sejalkhatri: If its not a string, then i m not able to send it as a parameter.

app/assets/javascripts/components/CustomLink.jsx
sejalkhatri: have added a comment `Replacement for non-custom Link component. The react Link component doesn't highlight the active link if it is not handeled by react router, CustomLink is used to provide that functioning.`

app/assets/javascripts/main.js
```diff
@@ -17,6 +17,7 @@ $(() => {
   require('./utils/language_switcher.js');
   require('./utils/editable.js');
   require('./utils/users_profile.js');
+  require('./utils/hamburger_menu.js');
```
ragesoss: It looks like this is not being used; the file itself is all commented out.

app/views/layouts/admin.html.haml
```diff
@@ -3,15 +3,15 @@
...
+  = hot_javascript_tag("main")
 
 !!!
 %html{ dir: rtl? ? "rtl" : nil, class: body_class(request)}
   = render "shared/head"
   %body{class: body_class(request)}
     = render "shared/flash"
     %div.wrapper
-      = render "shared/nav"
+      #nav_root{"data-rooturl" => main_app.root_url, "data-logopath" => logo_path, "data-fluid" => "false", "data-exploreurl" =>  main_app.explore_path, "data-explorename" => t(Features.wiki_ed? ? "application.explore" : "courses_generic.explore"), "data-usersignedin" => user_signed_in? ? 'true' : 'false', "data-ifadmin" => current_user&.admin? ? 'true' :'false', "data-trainingurl": main_app.training_url, "data-disable_training": Features.disable_training? ? 'true' : 'false', "data-help_disabled" => Features.disable_help? ? 'true' : 'false', "data-ask_url" => main_app.ask_url, "data-user_permissions" => permissions, "data-wiki_ed" => Features.wiki_ed? ? 'true' : 'false', "data-language_switcher_enabled" => language_switcher_enabled, "data-username" => current_user&.username, "data-destroyurl" => main_app.destroy_user_session_url, "data-omniauth_url" => main_app.user_mediawiki_omniauth_authorize_url}
```
ragesoss: In all these menus where you do something like `Features.wiki_ed? ? 'true' : 'false'`, you could just do `Features.wiki_ed?.to_s` instead.

Issue #1347, opened 2017-08-10
DiffViewer does not work for Wikidata


ragesoss:
The button should not be shown for Wikidata items.

ragesoss:
The button no longer shows up for Wikidata items. Nicer would be for the DiffViewer to work for them.

heysusgarcia:
If no one has started on this I can take it! I think I know what I need to do to get the DiffViewer to work for wikidata.

PR #1368, opened 2017-08-29
Fix diffviewer for wikidata #1347

heysusgarcia:
The code I added hides the button for wikidata items. However, I did notice when you're in the Articles section and open a drawer for an edited article for the first time you will briefly see the “Show Cumulative Changes” button.

### DIFF
--- a/app/assets/javascripts/components/revisions/diff_viewer.jsx
+++ b/app/assets/javascripts/components/revisions/diff_viewer.jsx
@@ -56,7 +56,6 @@ const DiffViewer = React.createClass({
     if (this.props.first_revision) {
       return this.findParentOfFirstRevision();
     }
-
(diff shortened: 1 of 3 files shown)


ragesoss:
This looks good. Can you squash it into a single commit?

heysusgarcia:
hey @ragesoss not sure if I did this correctly. It's my first time squashing commits. I noticed that my commit now includes changes to 2 other files.

ragesoss:
Looks like it ended up adding in some other changes that are in recent
commits on the main branch.

One option, since it's a small change, is to cut a new branch from
origin/master and recommit the same change.

Fetching upstream/master and rebasing onto to it might also work, but I'm
not sure.

ragesoss:
Merged via #1373

Issue #1205, opened 2017-02-28
Horizontal overflow when an instructor tries to create an available article


arun1595:
## Steps to reproduce:

1. Login as an instructor
2. Create a course and try to add a new available article.
3. You can find the popup causing a horizontal overflow.

### Screenshot

![overflow](https://cloud.githubusercontent.com/assets/7039523/23436802/d95a2764-fe31-11e6-85a4-7caacf2b8531.png)

coleworsley:
Hi there! I’m a first-time contributor and was hoping to help out with this issue. I noticed nobody was assigned to it, but if there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

ragesoss:
@coleworsley go for it! Let me know if I can help with anything.

PR #1371, opened 2017-08-29
Fix horizontal overflow on issue 1205

coleworsley:
'Fixes issue #1205'

@ragesoss Updated PR

Heres the message from before: We were able to prevent the overflow by right aligning the popovers with the right side of the container. This affected the button in the issue at hand, but we also noticed this affected the enrollment button on the students page. We adjusted the placement of .pop.open ::before pseudoelement to a fixed offset of 50px from the right. Please let us know if this is an acceptable solution.

Thanks!

Authors: @buji405 @dstock48 @coleworsley
### Screenshot 1
screen shot 2017-08-29 at 2 08 10 pm

### Screenshot 2
screen shot 2017-08-29 at 2 08 21 pm

### DIFF
--- a/app/assets/javascripts/components/common/popover.jsx
+++ b/app/assets/javascripts/components/common/popover.jsx
@@ -14,7 +14,9 @@ const Popover = React.createClass({
     if (this.props.is_open) {
       divClass += ' open';
     }
-
+    if (this.props.right_aligned) {
+      divClass += ' right-aligned';
+    }
     return (
       <div className={divClass}>
         <table>
--- a/app/assets/javascripts/components/common/popover.jsx
+++ b/app/assets/javascripts/components/common/popover.jsx
@@ -4,6 +4,7 @@ const Popover = React.createClass({
   displayName: 'Popover',
 
   propTypes: {
+    right_aligned: React.PropTypes.bool,
     is_open: React.PropTypes.bool,
     edit_row: React.PropTypes.node,
     rows: React.PropTypes.node
--- a/app/assets/javascripts/components/students/assign_button.jsx
+++ b/app/assets/javascripts/components/students/assign_button.jsx
@@ -26,6 +26,7 @@ const AssignButton = React.createClass({
     add_available: React.PropTypes.bool,
     assignments: React.PropTypes.array,
     open: React.PropTypes.func.isRequired,
+    right_aligned: React.PropTypes.bool,
     tooltip_message: React.PropTypes.string
   },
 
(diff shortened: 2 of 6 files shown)


ragesoss:
This works well for these two buttons, but there's some style regression in the PopoverButton components that use a plus sign as the entire button text (in edit mode of the Details section on the Home tab, for example).

popover plus button alignment

ragesoss:
One possible fix would be to add another prop to Popover to add an additional class that adjusts the alignment further, and then pass that prop from PopoverButton (but not from other components that use Popover).

coleworsley:
Sure we'll give that a try!

Thanks!

ragesoss:
There are some linting issues that are breaking the build; check the `gulp` output for the linter errors.

Does this fix all the cases of the "+" button, or just for `AssignButton` and `EnrollButton`?

coleworsley:
Hey @ragesoss! This fixes just the `Assign` and `Enroll` buttons

We took your advice and added a new "right_aligned" prop to the **popover.jsx** components that are rendered in the **assign_button.jsx** and **enroll_button.jsx** components. We then added logic to the render method of the **popover.jsx** component to give it a "right-aligned" class when the "right_aligned" prop was `true`. We noticed the **popover_button.jsx** component wasn't actually being used in the files affecting the issue -- not sure if this is intentional or an oversight, as other files seem to be using it. Happy to make changes or refactor as needed. Thanks!

ragesoss:
This doesn't seem to be having the intended effect. I see the `right-aligned` class for the assign button in the Students table, but it does not change the styling; it looks like it's being overridden by other css.

Also, there are several components in the Home tab Details section that (indirectly) use PopoverButton, including TagButton and some others (see the screenshot I included from the Home tab).

One alternative to the props approach I suggested above might be to utilize the `plus` class, which is used wherever the small form of the button is present. I'm not sure, but it might be possible to do it all in css.

coleworsley:
Hey @ragesoss, we looked at your comments regarding our most recent PR, but we're not really sure  by what you mean by not having the intended effect. Please see below for a screenshot of what were seeing. Does yours show differently?

image

Apologies for the delay, we are currently students in a web development school and have been assigned a new project which has taken up the majority of our time. We would like to continue working on this issue, however, if anyone else can provide a better solution we are happy to hand it off.

ragesoss:
@coleworsley see my screenshot above; the case I'm talking about is when the button consists of just a plus sign or similar very small amount of text. In that case, the popover's arrow does not end up point to the button, but instead points to the left of it.

(This happens on the Assign button in your screenshot *after* you've already added at least one thing to the list.)

ragesoss:
This work is being finished up in  #1399

ragesoss:
Thanks!

PR #1374, opened 2017-08-30
DiffViewer does not work for Wikidata #1347

heysusgarcia:


heysusgarcia:
@ragesoss i recommitted the diffviewer change. Let me know if this is good/if i should do anything else

ragesoss:
I merged this already from a rebased branch, #1373.

Issue #1356, opened 2017-08-22
Programs should be searchable (by name and/or institution)


ragesoss:
It would be useful also be have searches that could be scoped by Campaign (for example, so search by name or institution from a Campaign)

PR #1379, opened 2017-09-06
Have ability to search courses by title, institution, or term #1356

mattwarren10:


mattwarren10:
Please let me know if this works out ok. I tried looking to see if the courses nav would be responsive. Turns out my Chrome browser--when in responsive view for iPhone--shows a different view than on my actual iPhone.

app/assets/stylesheets/modules/_nav.styl
```diff
@@ -239,6 +239,12 @@ nav
         line-height 55px
         overflow hidden
         -webkit-font-smoothing auto
+    ^[2] nav
```
ragesoss: What does this do? I've never seen that kind of thing is a .styl file before.

app/views/campaigns/programs.html.haml
```diff
@@ -67,5 +67,9 @@
             %th{style: 'width:100px'}
               = t('courses.actions')
       %tbody.list
+      - if @keywords.nil?
```
ragesoss: Rather than if-check a negative, the standard Ruby way to handle this would be to switch the order and do
```
if @keywords.present?
  <do search version>
else
  <do normal version>
```

ragesoss:
Here's how the layout looks for me.

search

app/assets/stylesheets/modules/_nav.styl
mattwarren10: It's a partial reference. [http://stylus-lang.com/docs/selectors.html#partial-reference](url). I took it out though. We don't need it because it's easier to adjust the selector's indention.

mattwarren10:
So the `^ [2] nav` selector is a [partial reference](http://stylus-lang.com/docs/selectors.html#partial-reference). I took it out though. We don't need it because I can simply adjust the selector's (`nav .campaign-nav__ul`) indention.

Also, the course search form is neatly inline with the nav bar for me. 
screen shot 2017-09-08 at 4 28 10 pm

spec/presenters/courses_presenter_spec.rb
```diff
@@ -13,6 +13,12 @@
       expect(subject.trained_count).to eq(1)
       expect(subject.trained_percent).to eq(50.0)
     end
+    context 'search campaign' do
+      it 'returns courses when searching' do
+        q = 'basket-weaving'
```
ragesoss: This implicitly relies on the course title set by the factory ­— the "mystery guest" code smell, I think — so it's not clear just from looking at the test whether we expect it to pass or not.

Here we care about how the test interacts with the title, term and school, so you should set those explicitly for the course records used in this test.

It's also written against `nil` as the failure condition, but a search that returns no results will be an empty ActiveRecord relation, not `nil`, so this test will pass no matter what query you set. You can instead use `be_empty`. (It's very useful to make sure your tests fail in the expected ways when you change them, in addition to passing as written.)

mattwarren10:
Travis-ci says the red tests failed to pass:
image

In my test suite, they pass:

```
$ rspec spec/lib/article_status_manager_spec.rb 
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}
...........

Finished in 8.39 seconds (files took 6.87 seconds to load)
11 examples, 0 failures
```

What do I need to look at to fix this problem?

ragesoss:
This was a problem on the wmflabs server. I'll restart the test.

ragesoss:
The second job is a superset and it passed, so this build is fine.

mattwarren10:
If this PR is successful, I'll take on another issue.

PR #1381, opened 2017-09-08
Remove 'disable_training' feature toggle

### LINKED ISSUE #1375: Remove the 'disable_training' feature toggle
This is no longer used or needed, since both production instances have training enabled. The config should be removed from the example application.yml file, the relevant method in `Features` presenter should be removed, and all the places that call it — most in view templates — will also need to be updated.

majakomel:
Closes #1375


ragesoss:
This is going in the right direction.

It looks like you did this based on a slightly older version of the main branch, from before we merged a rewrite of the top nav. Now there's a call to the toggle in each of the layouts, and there's a little bit of React component code that will need to be adjusted to remove the client-side toggling as well.

majakomel:
Oh, forgot to check for `disableTraining`, thanks

PR #1384, opened 2017-09-09
Prevent the overshadowing of the course main section by the course nav bar section

sejalkhatri:


ragesoss:
Nice. The Notifications component still gets covered up by the course nav in mobile layout, but this helps a lot for most common situations where there is no Notification message.

PR #1389, opened 2017-09-12
improved diffviewer to work with wikidata #1347

heysusgarcia:


app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -60,7 +60,16 @@ const DiffViewer = React.createClass({
   },
 
   wikiUrl() {
-    return `https://${this.props.revision.wiki.language}.${this.props.revision.wiki.project}.org`;
+    const wikiProject = this.props.revision.wiki.project;
+    let url;
+
+    if (wikiProject === 'wikipedia') {
```
ragesoss: Most projects work the same way as Wikipedia in terms of the url: es.wikisource.org, ta.wiktionary.org, etc. There are just a handful that have a different url pattern; I think the only ones we currently support are wikidata (www.wikidata.org) and multilingual wikisource (wikisource.org).

heysusgarcia: That is good to know. As it stands, my logic will only work for wikipedia and wikidata projects. If I reverse my logic as follows: check if wikiProject === 'wikidata' , else return the url pattern we had previously been returning that supports wikipedia, wikitionary, etc url patterns, will that suffice?

ragesoss: I think so. You can add some activity from users active on a couple different wikis and try it out.

heysusgarcia:
@ragesoss my branch has been updated with the revised logic.

PR #1390, opened 2017-09-12
changed 'ruby 2.3.1' to 'ruby >= 2.3.1' in Gemfile

alpha721:
changed 'ruby 2.3.1' to 'ruby >= 2.3.1' in Gemfile to make it compatible with higher versions of ruby.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.3.1'
+ruby '>=2.3.1'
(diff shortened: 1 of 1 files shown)


ragesoss:
@alpha721 can you amend the commit message? It has the wrong version, which might be confusing later.

ragesoss:
@alpha721 you changed the PR title, but not the commit message. You can do `git commit --amend` and then `git push -f` to change the commit itself.

PR #1396, opened 2017-09-15
improved diffviewer to work with wikidata #1347 #1389

heysusgarcia:
@ragesoss  i went ahead and created a new pull request for this enhancement. I noticed that even though I had updated my branch with new logic, the original PR was still showing the old commit and none of the improved logic.

### DIFF
--- a/app/assets/javascripts/components/revisions/diff_viewer.jsx
+++ b/app/assets/javascripts/components/revisions/diff_viewer.jsx
@@ -60,7 +60,11 @@ const DiffViewer = React.createClass({
   },
 
   wikiUrl() {
-    return `https://${this.props.revision.wiki.language}.${this.props.revision.wiki.project}.org`;
+    if (this.props.revision.wiki.language) {
+      return `https://${this.props.revision.wiki.language}.${this.props.revision.wiki.project}.org`;
+    }
+
+    return `https://${this.props.revision.wiki.project}.org`;
   },
 
   diffUrl() {
@@ -89,7 +93,7 @@ const DiffViewer = React.createClass({
   },
 
   findParentOfFirstRevision() {
-    const wikiUrl = `https://${this.props.first_revision.wiki.language}.${this.props.first_revision.wiki.project}.org`;
+    const wikiUrl = this.wikiUrl();
     const queryBase = `${wikiUrl}/w/api.php?action=query&prop=revisions`;
     const diffUrl = `${queryBase}&revids=${this.props.first_revision.mw_rev_id}&format=json`;
     $.ajax(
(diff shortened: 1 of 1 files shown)


ragesoss:
Cool. I'll probably review it early next week. Thanks!

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -60,7 +60,16 @@ const DiffViewer = React.createClass({
...
+    const wikiProject = this.props.revision.wiki.project;
+    let url;
+
+    if (wikiProject === 'wikidata') {
+      url = `https://${this.props.revision.wiki.project}.org`;
+    } else {
+      url = `https://${this.props.revision.wiki.language}.${this.props.revision.wiki.project}.org`;
+    }
+
+    return url;
```
ragesoss: This is more complicated than it needs to be. Maybe instead:

```javascript
if (this.props.revision.wiki.project === 'wikidata') { return 'https://www.wikidata.org' }
return `https://${this.props.revision.wiki.language}.${this.props.revision.wiki.project}.org`;
```

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -89,7 +98,15 @@ const DiffViewer = React.createClass({
...
-    const wikiUrl = `https://${this.props.first_revision.wiki.language}.${this.props.first_revision.wiki.project}.org`;
+    const wikiProject = this.props.revision.wiki.project;
+    let wikiUrl;
+
+    if (wikiProject === 'wikidata') {
+      wikiUrl = `https://${wikiProject}.org`;
+    } else {
+      wikiUrl = `https://${this.props.first_revision.wiki.language}.${wikiProject}.org`;
+    }
+
```
ragesoss: This could all be just:

```javascript
const wikiUrl = this.wikiUrl();
```

The wiki for `first_revision` will always be the same as the wiki for `revision`, so no need to duplicate any of that logic.

app/assets/javascripts/components/revisions/diff_viewer.jsx
ragesoss: Instead of that, you could make it a little more general by switching on language:

```javascript
if (this.props.revision.wiki.language) {
  return  `https://${this.props.revision.wiki.language}.${this.props.revision.wiki.project}.org`;
} else {
  return  `https://${this.props.revision.wiki.project}.org`;
}

That should correctly handle both wikidata and multilingual wikisource, the two wikis that have a project but no language.

heysusgarcia:
I updated my branch to reflect requested changes. Hopefully this will do the trick.

ragesoss:
thanks @heysusgarcia!

PR #1398, opened 2017-09-15
partial fixes for issue 1395

jstoebel:
move application controller rescues to their own module (errors/rescue_errors)
respond with useful error messages for various errors in application controller

I still have two questions:

 1. I'm not sure how to handle point 1: "redirect /404 and /500 to the prettier error pages." AFAIK, rails first tries to fulfill a request by looking for static assets in `public/`. `routes.rb` is only consulted if they are not found. Since we don't want to remove `public/404.html` I'm not sure how to have the router handle requests for `/404` or `/500`
2. How specifically do you want to handle errors in the application controller when responding to html? Do you want to redirect somewhere for example?

### DIFF
--- a/lib/errors/rescue_errors.rb
+++ b/lib/errors/rescue_errors.rb
@@ -30,8 +30,9 @@ module Errors
     def self.rescue_not_signed_in(base)
       base.rescue_from AuthenticationErrors::NotSignedInError do |e|
         respond_to do |format|
-          format.json { render json: { message: e.message }, status: :unprocessable_entity }
-          format.html {} # TODO
+          format.json { render json: { message: e.message }, status: :unauthorized }
+          # TODO: need more user friendly error handling for html
+          format.html { render plain: e.message, status: :unauthorized } 
         end
       end
     end # rescue_not_signed_in
@@ -39,8 +40,9 @@ module Errors
     def self.rescue_not_permitted(base)
       base.rescue_from AuthenticationErrors::NotPermittedError do |e|
         respond_to do |format|
-          format.json { render json: { message: e.message }, status: :unprocessable_entity }
-          format.html {} # TODO
+          format.json { render json: { message: e.message }, status: :unauthorized }
+          # TODO: need more user friendly error handling for html
+          format.html { render plain: e.message , status: :unauthorized }
         end
       end
     end # rescue_not_permitted
@@ -48,8 +50,9 @@ module Errors
     def self.rescue_not_admin(base)
       base.rescue_from AuthenticationErrors::NotAdminError do |e|
         respond_to do |format|
-          format.json { render json: { message: e.message }, status: :unprocessable_entity }
-          format.html {} # TODO
+          format.json { render json: { message: e.message }, status: :unauthorized }
+          # TODO: need more user friendly error handling for html
+          format.html { render plain: e.message, status: :unauthorized }
         end
       end
     end # rescue_not_admin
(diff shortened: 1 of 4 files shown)


ragesoss:
1. Did you test out that snippet for get '/404' in the routes file? Maybe it works for /404 but not /404.html because rails uses routes to handle the implicit .html format?

2. I think to start with, we should keep the same current behavior, but add new error types (and rescue them in the new errors module). I think the handling of different formats can happen in the errors module after that. To answer your question more fully, I'm not yet sure what should happen in the html case, but whatever does happen should be controlled by that errors module rather than at the ApplicationController level.

lib/errors/rescue_errors.rb
```diff
@@ -21,6 +29,26 @@ def self.rescue_unknown_format(base)
       end
     end # self.rescue_unknown_format
 
+    ##
+    # dynamically define each rescue in terms of the exception it will catch
+    AUTHENTICATION_ERRORS.each do |err|
+      send(:define_singleton_method, "rescue_#{err}") do |base|
+        base.rescue_from "Errors::AuthenticationErrors::#{err.to_s.camelcase}Error".constantize do |e|
```
ragesoss: Let's not metaprogram these definitions, as we will want to have different messages for each one, and in some cases, the way we handle them may differ. Plus, we want to be able to search the codebase for the error name and find out where the handling is defined.

jstoebel:
**item 1** If I'm understanding this correct, we want `get /404` to resolve to the pretty 404 page, but `get /404.html` to resolve to the rails default page. If so, I am stumped on how to solve this other than by removing the static page. The problem seems to be that Rails first tries to fulfill the request by looking in the static folder and if it doesn't get a match then falls back to  `config/routes.rb`. I wasn't able to find a way to ask the app to look in `routes.rb` first (and that would have a performance hit on serving of static assets). If there is a better idea that I am missing let me know.

**item 2** I've made the requested changes to not use meta programming. I also had to slightly tweak a test to match a new error message. The way it works now, the error message is passed into the custom exception and when caught is sent out the door as the http response.  Please let me know if this isn't what you had in mind.

ragesoss:
Removing the 404.html and 500.html will probably be fine, I guess. Not sure whether there will be unintended consequences, but I guess I'll find out in the long run.

ragesoss:
Cool. This looks like it's coming along. Now that we have the named errors with good default messages, I think we can streamline ApplicationController more and also improve the behavior for differentiating between errors that are caused by the user being signed out and the user otherwise not being allowed to do something. Something like this...

```ruby
  def require_signed_in
    raise NotSignedInError unless user_signed_in?
  end

  def require_permissions
    require_signed_in
    course = Course.find_by_slug(params[:id])
    raise NotPermittedError unless current_user.can_edit?(course)
  end

  def require_admin_permissions
    require_signed_in
    raise NotAdminError unless current_user.admin?
  end

  def require_participating_user
    require_signed_in
    course = Course.find_by_slug(params[:id])
    # Course roles for non-students are greater than STUDENT_ROLE.
    # Non-participating users have the VISITOR_ROLE, which is below STUDENT_ROLE.
    return if current_user.role(course) >= CoursesUsers::Roles::STUDENT_ROLE
    raise ParticipatingUserError
  end
```

The rescue for `ActionController::InvalidAuthenticityToken` should definitely get a json version, as that's probably the most common error that will come from json requests, if a user gets signed out but their tab doesn't know it. Currently, the explanation doesn't get through, and it just displays 'Unauthorized' to the user.

jstoebel:
Let me know what you think about this.

ragesoss:
@jstoebel I'm testing this out, and it looks like `respond_to` doesn't work in these rescue blocks. It always servers the plain version, even for json requests.

I found a workaround like this, although I feel like there must be a better way...

```ruby
    def json?(request)
      request.media_type.include? 'json'
    end

    def self.rescue_invalid_token(base)
      base.rescue_from ActionController::InvalidAuthenticityToken do
        if json?(request)
          render json: { message: 'Please sign in' }, status: :unauthorized
        else
          render render plain: t('error_401.explanation'), status: :unauthorized
        end
      end
    end # self.rescue_invalid_token
```

ragesoss:
This looks good! Despite having to do the workaround, I think it came out pretty understandable, and I think it'll accomplish the main goals of making it easier to a) handling controller errors consistently, and b) show useful messages to users.

I'm going to make a few further tweaks to the messages and add separate handling for 'not signed in' vs. 'signed in *but* something else is now allowed'.

PR #1399, opened 2017-09-16
Fix horizontal overflow on Popover components for issue #1205

mattwarren10:
screen shot 2017-09-15 at 7 57 43 pm

I simply added the additional `plus-right-aligned` class to overwrite `right-aligned` on `<Popover />` components that have the `<button>` with class `plus`.

### DIFF
--- a/app/assets/stylesheets/modules/_popover.styl
+++ b/app/assets/stylesheets/modules/_popover.styl
@@ -19,17 +19,19 @@
     // so we need to turn it off and add the -webkit- prefix explicitly to make
     // it work together with css-flipper.
     /* autoprefixer: off */
-    /*@replace: translateX(50%) translateY(-30px) scale(.7, .7)*/ transform translateX(-50%) translateY(-30px) scale(.7, .7)
-    /*@replace: translateX(50%) translateY(-30px) scale(.7, .7)*/ -webkit-transform translateX(-50%) translateY(-30px) scale(.7, .7)
+    /*@replace: translateX(0px) translateY(0) scale(1, 1)*/ transform translateX(0px) translateY(0) scale(.7, .7)
+    /*@replace: translateX(0px) translateY(0) scale(1, 1)*/ -webkit-transform translateX(0px) translateY(0) scale(.7, .7)
     /* autoprefixer: on */
     transition all .2s
     z-index 10
+    left auto
+    right 0
     &.open
       opacity 1
       pointer-events all
       /* autoprefixer: off */
-      /*@replace: translateX(50%) translateY(0) scale(1, 1)*/ transform translateX(-50%) translateY(0) scale(1, 1)
-      /*@replace: translateX(50%) translateY(0) scale(1, 1)*/ -webkit-transform translateX(-50%) translateY(0) scale(1, 1)
+      /*@replace: translateX(0px) translateY(0) scale(1, 1)*/ transform translateX(0px) translateY(0) scale(1, 1)
+      /*@replace: translateX(0px) translateY(0) scale(1, 1)*/ -webkit-transform translateX(0px) translateY(0) scale(1, 1)
       /* autoprefixer: on */
     &:before
       border-bottom 6px solid white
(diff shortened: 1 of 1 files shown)


ragesoss:
Cool. I think we can step back some of the complexity... as far as I can tell, the 'right-aligned' style is or could be used everywhere, so that prop could maybe be removed and those styles put into `.open` as the other PR did initially. Then just the `plus-right-aligned` prop would be needed.

I'm not sure about that, but I `think` we only need one prop to control this styling.

mattwarren10:
With these changes:
``` // _popover.styl
    &.open
      opacity 1
      pointer-events all      
      /*@replace: translateX(0px) translateY(0) scale(1, 1)*/ transform translateX(0px) translateY(0) scale(1, 1)
      /*@replace: translateX(0px) translateY(0) scale(1, 1)*/ -webkit-transform translateX(0px) translateY(0) scale(1, 1)
      /* autoprefixer: on */
      left auto
      right 0    
    &:before
      border-bottom 6px solid white
      border-right 6px solid transparent
      border-left 6px solid transparent
      content ''
      display block
      left 50%
      margin-left -3px
      position absolute
      top -6px
    &.open:before
      left auto
      right 20px
    &.plus-right-aligned:before
      left auto
      right 5px
```
I have this issue when  when the `.open` class is added.
sep-15-2017 21-11-12

Not sure how to fix this.

ragesoss:
That's weird. That looks like the transition we use in the assignment wizard. I don't know either.

mattwarren10:
Well this PR works with `right-aligned` in tact. The changes above that include removing `right-aligned` and putting all the styles in the `.open` class were not committed. So it's good to go, unless you want to fix this transition.

ragesoss:
Maybe rather than in 'open', those just go in 'pop'. The idea being that 'right-aligned' can be the default styling and we don't need a prop for it.

mattwarren10:
Putting them in `.pop` doesn't look too good either.

ragesoss:
Okay. Can we just remove the prop and conditional and always use right-aligned? If we can avoid using two new props, we should.

mattwarren10:
So keep `right-aligned` as a `divClass` but remove conditional and prop. Correct?

ragesoss:
Yeah, if that doesn't cause any problems. Should reduce the surface area of this change considerably.

mattwarren10:
So I removed the prop and conditional and it looks like this now:
``` 
//popover.jsx
render() {
    let divClass = 'pop';
    if (this.props.is_open) {
      divClass += ' open';
      divClass += ' right-aligned';
    }
...
}
```
I removed the `.right-aligned` prop from all the other components. No changes in CSS. No errors in gulp, but I am still getting the same css transition bug exactly like the gif above.

mattwarren10:
Wait.. I figured it out.

mattwarren10:
Let me know of any other changes needed before merging.

app/assets/javascripts/components/students/enroll_button.jsx
```diff
@@ -23,6 +23,7 @@ const EnrollButton = React.createClass({
     inline: React.PropTypes.bool,
     open: React.PropTypes.func,
     is_open: React.PropTypes.bool,
+    right_aligned: React.PropTypes.bool,
```
ragesoss: This can be removed?

app/assets/javascripts/components/common/popover.jsx
```diff
@@ -4,17 +4,23 @@ const Popover = React.createClass({
...
   propTypes: {
+    plus_right_aligned: React.PropTypes.bool,
     is_open: React.PropTypes.bool,
     edit_row: React.PropTypes.node,
     rows: React.PropTypes.node
   },
 
   render() {
     let divClass = 'pop';
+    divClass += ' right-aligned';
```
ragesoss: One last bit of cleanup is possible now... may as well move this into the `let` in the line above.

ragesoss:
I tested this out, and the alignment looks good... but the transition animation no longer happens. I don't want to lose those, as it makes these popovers feel much nicer.

mattwarren10:
I have class here in a few minutes until 5:15pm est. Around that time I can take a look at it and make the necessary adjustments.

mattwarren10:
So I was able to completely remove both the `right-aligned` and `plus-right-aligned` classes and have the transition run smoothly.

ragesoss:
Looks good to me. I'm glad it's a css-only fix!

Issue #1360, opened 2017-08-24
Campaign 'Articles' tab should show more detail about each article


ragesoss:
In particular, characters added — and sorting by that value — would be very nice.

testa19:
Hi
There is same feature implemented on Course's Articles page. See example: https://dashboard.wikiedu.org/courses/Ohio_University/Writing_for_Biologists_(Spring_2017)/articles

Would it be better to reuse the same logic(? or React code)?
Or is there anything against that?

ragesoss:
@testa19 it would be rather more complicated to do it in React, because unlike the course pages, the Campaign pages don't currently use any React. It would be nice to port them to React at some point, but it's a fair bit of work.

testa19:
Sage, is there some kind of _articles_cources_ relation is needed for Compaign model (_like has_many :articles, -> { distinct }, through: :courses_) to get the view and characters count.
Or we should just use `article.articles_courses.where(course_id: @campaign.courses)`

`       - @campaign.articles.includes(:wiki).order(:title).each do |article|`
`          %tr
              %td.title
                = link_to full_title(article), article_url(article)
              %td
                = article.articles_courses.where(course_id: @campaign.courses).sum(&:character_sum)
              %td
                = article.articles_courses.where(course_id: @campaign.courses).sum(&:view_count)
              %td
                %small
                  = "#{article.wiki.language}.#{article.wiki.project}" `
   ?

ragesoss:
@testa19 very good question!

I think adding an articles_courses relation for Campaign would make the most sense. This page may have very large numbers of articles, so it's going to be important to avoid introducing any N+1 queries.

So you probably want to add a Campaign.articles_courses scope, and do something like @campaign.articles_courses.includes(:articles). Then you can rewrite that view to iterate over the articles_courses, but each AC will already have the Article record loaded without any additional queries.

testa19:
Thanks for advice!
Now there is `@campaign.articles_courses.includes(article: :wiki).order('articles.title')` for querying.

I have another question:
I've madde some changes to campaign JS ('/assets/javascripts/') for sorting on the page to work but it seems like it is not loading (js from public is (?)). 
There could be an issue of me using ubuntu VM on Windows host with shared folder and some guard/smth-else is not working properly.
Or is there some asset related task i should run to update the script?

testa19:
[...]

PR #1400, opened 2017-09-18
Articles tab show's more detail about each article

testa19:
Changes for enhancement https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1360

Campaign.articles_courses relation is added
campaigns/articles view changed to show table header row and two additional info columns
Sorting articles option is added to javascript

### DIFF
--- a/app/assets/javascripts/campaigns.js
+++ b/app/assets/javascripts/campaigns.js
@@ -84,10 +84,22 @@ $(() => {
     });
   }
 
+  // Article sorting
+  // only sort if there are tables to sort
+  let articlesList;
+  if ($('#articles table').length) {
+    articlesList = new List('articles', {
+      valueNames: [
+        'title', 'views', 'char_added'
+      ]
+    });
+  }
+
   return $('select.sorts').on('change', function () {
     const list = (() => {
       switch ($(this).attr('rel')) {
         case 'campaigns': return campaignList;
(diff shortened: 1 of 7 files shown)


ragesoss:
Nicely done! One last tweak that would improve it a bit more would be to add sorting by wiki (the last column). @testa19 let me know if you want to add that now; otherwise, I'll merge it as-is.

testa19:
@ragesoss ok, let me take a quick look (i don't know how to modify a PR with another commit yet)

ragesoss:
@testa19 you just checkout your branch again, add a new commit, and push it. The PR will update automatically.

testa19:
@ragesoss There is an issue with column name: there is no translation for it

ragesoss:
@testa19 you can add one by editing `config/locales/en.yml`.

testa19:
@ragesoss So i can just add `wiki_language` field to `articles` "model"?

ragesoss:
@testa19 in this case, the column is both the wiki language and project (like `en.wikipedia`), so just 'wiki' is probably the best name for it.

testa19:
@ragesoss Ok, seems like the second commit has been checked:
Wiki column name and sorting were added
.table--expandable CSS class was removed from the table (nothing to expand on)

PR #1405, opened 2017-09-19
For course slugs, multiple internal spaces should be collapsed into one space

alpha721:
made a few changes to `course.utils.js` , to collapse multiple whitespaces into one space for course slugs.


ragesoss:
That looks good. Please also update cleanupCourseSlugComponents() so that it does the same thing to each of the attributes that make the slug: school, title, and term.

ragesoss:
The spec for cleanupCourseSlugComponents() should also be updated to test the new behavior.

Issue #1391, opened 2017-09-13
Real name field can have leading/trailing whitespace


ragesoss:
This should be removed for consistency.

Real names are set via the /onboarding flow.

testa19:
Does adding simple before save callback to User model is ment?
before_save do
  self.real_name = self.real_name.squish
end

ragesoss:
I would rather sanitize it at the controller level instead of adding a callback. Empty is not the problem; it's things like ` My name `.

asmigos:
Hello, 
I am new-comer,
This could be resolved by using `.trim` method

testa19:
@ragesoss so is it just simple private method for Onboarding controller like below that is called by  onboard action:
```
validate_params
**sanitize_real_name**
@user = current_user
set_new_permissions
```
```
def sanitize_real_name
  params[:real_name] = params[:real_name].squish
end
```
?

ragesoss:
Yes, something like that should work.

PR #1406, opened 2017-09-19
Add real_name 'sanitization' to Onboarding controller

testa19:
Minor fix to a minor bug https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1391

Private method in `OnboardingController` for `Real_name` sanitization.

### DIFF
--- a/app/controllers/onboarding_controller.rb
+++ b/app/controllers/onboarding_controller.rb
@@ -15,7 +15,7 @@ class OnboardingController < ApplicationController
     validate_params
     @user = current_user
     set_new_permissions
-    @user.update_attributes(real_name: params[:real_name],
+    @user.update_attributes(real_name: sanitized_real_name,
                             email: params[:email],
                             permissions: @permissions,
                             onboarded: true)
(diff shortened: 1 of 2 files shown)


app/controllers/onboarding_controller.rb
```diff
@@ -40,4 +41,8 @@ def validate_params
       obj.require(key)
     end
   end
+
+  def sanitize_real_name
+    params[:real_name].squish!
```
ragesoss: Rather than modifying the params with `.squish!, maybe make this method `sanitized_real_name`, use `.squish`, and do `@user.update_attributes(real_name: sanitized_real_name)`

ragesoss:
The new behavior should also be tested via onboarding_controller_spec.

testa19:
@ragesoss Should there be a separate `it "should strip name field of excessive whitespace"` test case like below?

it "should strip name field of excessive whitespace" do
  params = { real_name: " Name  \n Jerry ", email: 'email@email.org', instructor: false }
  put 'onboard', params: params
  expect(user.real_name).to eq('Name Jerry')
end

ragesoss:
@testa19 yes, a separate test case would be fine.

Issue #1385, opened 2017-09-10
Create `SpecialUsers` presenter to fetch users in special system-wide roles


ragesoss:
Currently, we use ENV variables to identify some specific user roles like classroom_program_manager who receive certain types of emails.

Instead of using ENV, those usernames should be set in a `Setting` record, and accessed via a new `SpecialUsers` class. Then all the places that currently use ENV to get a specific user can use that instead.

(Moving that config data to the `Setting` record will pave the way to create a UI for changing those users.)

alpha721:
Hi, I am a first time contributer.
I would like to take up this issue if no one else is working on it.

BellPepperLab:
Hi ragesoss,

Can it be achieved by adding a "role" column to the user table and then creating an ENUM with all user roles in the user model.

ragesoss:
There are many possible ways, but in this case the setting more represents a system-wide setting rather than a trait for a particular user. The idea here is to refactor the way configuration settings are accessed to move them from the appllcation.yml into a `Setting` record in the database. The benefits includes: 1. no changes to the database schema are necessary, and 2: because these are system-wide settings, they will probably make more sense and be easier to deal with in the long run if they get treated as `Setting` records.

testa19:
@ragesoss hi
So is it should be like this:

./presenters/special_user
```
class SpecialUser
  def self.classroom_program_manager
    User.find(Setting.first.value[:classroom_program_manager])
  end
end
```

Somewhere in code:
`staffer = SpecialUser.classroom_program_manager`
?

Second question:
How are settings stored in `Settings` table? What are the contents of the `key` property?

ragesoss:
The `key` property is basically the name of the setting, so in this case, it might be 'special_users', which you would use to find it. And then the `value` would be a hash of all the roles and usernames, so something like 
```ruby
{ classroom_program_manager: 'Username 1', another_special_user_role: 'Username 2' }
```

testa19:
@ragesoss So would there be a Settings record and searching would be:
`Setting.find_by(key: :special_users).value`

And should there be a presenter SpecialUser:
```
class SpecialUser
  def self.classroom_program_manager
    username = Setting.find_by(key: :special_users).value[classroom_program_manager]
    User.find_by(username: username)
  end
end
```

Or is there some kind of god practise of dealing with these tipes of tasks?

ragesoss:
[...]

Issue #1386, opened 2017-09-10
Create a super-admin UI for modifying application settings


ragesoss:
We currently use lots of ENV variables for assorted system-wide settings, which occasionally must be changed. Some of these would make more sense if moved to a `Setting` record.

Build a UI, accessible only to users who are among the system's super-admins (a designation that does not exist yet) for modifying the system-wide user roles. (This depends on #1385 being completed first.)

In this UI, a super-admin will be able to set and change other system-wide user roles like classroom_program_manager, and will also be able to modify the permissions level of any user (to add or remove admin rights, or change a student to instructor or vice-versa).

testa19:
Hi @ragesoss 
Have some questions:
`App Settings` link could be added to Admin Navigation bar.
So can we use `Admin` controller? `Index action` in paticular: rendering empty `index.haml` for admins and some `super_index` for super_admin (with app settings)
So `App Settings` nav tab would lead to `Admin:index`
And `Admin:update` would be a place for changing `Settings` table (Is it ok?)

Changing user roles (remove admin rights, change a student to instructor or vice-versa):
is it for `User` Nav tab and for `User Controller`?

testa19:
@ragesoss Or it better be a `Settings_controller`
`resources :settings` in routes
With `Admin `layout
`App Settings` link could be added to Admin Navigation bar.

Changing user roles (remove admin rights, change a student to instructor or vice-versa):
is it for User Nav tab and for User Controller?

ragesoss:
I think a Settings controller is the best option. The changing of admin rights could be done from there. I think one good way to start might be to render a list of all the admins, and have a 'New Admin' form where the user puts in a username, and if they already exist, they get turned into an admin. Similarly, there could be a 'remove admin rights' button for each current admin.

Users can switch themselves between student and instructor, so there is no need to provide tools to do that here. When admin rights are removed, the user should be left with 'instructor' permissions.

jstoebel:
after speaking with @ragesoss he suggested I attempt this ticket. @testa19 if you are still working on this ticket, let me know. I don't want to step on toes!

testa19:
@jstoebel no problem. 
Can't find the time lately (

PR #1411, opened 2017-09-22
Create SpecialUser Presenter

testa19:
Simple presenter for https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1385

### DIFF
--- a/app/presenters/special_users_presenter.rb
+++ b/app/presenters/special_users_presenter.rb
@@ -1,0 +1,26 @@
+#= Presenter for Setting / special users
+class SpecialUser
+  def self.special_users
+    Setting.find_or_create_by(key: 'special_users').value
+  end
+
+  def self.communications_manager
+    User.find_by(username: special_users[:communications_manager])
+  end
+
+  def self.classroom_program_manager
+    User.find_by(username: special_users[:classroom_program_manager])
+  end
+
+  def self.technical_help_staff
+    User.find_by(username: special_users[:technical_help_staff])
+  end
+
+  def self.survey_alerts_recipient
+    User.find_by(username: special_users[:survey_alerts_recipient])
+  end
+
+  def self.super_admin
+    User.find_by(username: special_users[:super_admin])
+  end
+end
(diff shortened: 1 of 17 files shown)


ragesoss:
This looks sensible to me; now all the places that relied on the old method of finding those users need to be replaced.

testa19:
@ragesoss Hi!
Have replaced old methods.
But there should be the special_users record in Setting with all appropriate stuff for changes to work.
Is it done manually?

ragesoss:
> Have replaced old methods.
But there should be the special_users record in Setting with all appropriate stuff for changes to work.
Is it done manually?

Yes, I'll manually move those setting from application.yml to a Setting record before I deploy it.

testa19:
@ragesoss Yeah, there is a small issue: i dont know how to move those settings to DB without using console :)

app/presenters/special_users_presenter.rb
```diff
@@ -0,0 +1,26 @@
+#= Presenter for Setting / special users
+class SpecialUser
```
ragesoss: The class name and file name should correspond to each other. In this case, since the class is a presenter for all special users, like the existing `Features` presenter which does something similar, I think `SpecialUsers` is the best name for the class (and so, `special_users.rb` for the file name).

spec/controllers/alerts_controller_spec.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 require 'rails_helper'
+require "#{Rails.root}/app/presenters/special_users_presenter"
```
ragesoss: These 'require' statements won't be necessary if the file name and class name follow the Rails convention.

app/presenters/special_users_presenter.rb
testa19: Thanks!
There was (is) a mess of concepts in my head: model name vs table name, presenters and "helper" wrapper class in presenters' folder. And not so good knowledge of tests environment and no complete understanding of how and what is loaded for tests.
Good to have a real life example to get some ground

ragesoss:
@testa19 one more thing: the 'super admin' role is currently both unused and untested. It should be removed from this PR; it can be added in for #1386.

Issue #1407, opened 2017-09-19
Article titlle should not say 'null' for Wikidata items, etc


ragesoss:
When an article is shown that has a wiki with no language attribute, such as wikidata or multilingual wikisource, the dashboard shows 'null' as the language code. Instead, should omit the language code in this case.

![null](https://user-images.githubusercontent.com/848483/30607158-e873999c-9d28-11e7-9e37-57646f6cff61.png)

mauditecandela:
I would like to pick this as my first issue :)

mauditecandela:
Hi! I would like to pick this as my first contribution to the program but unfortunately cannot reproduce the bug. Could someone help me? Thanks.

ragesoss:
You can reproduce it by creating a course, adding a user who has been active in Wikidata (e.g., Ragesoss if you set the course to start about a month ago), adding an available article from Wikidata to connect the course to that wiki, and then visiting [course page]/manual_update.

mauditecandela:
I created a PR for a possible fix.

testa19:
@mauditecandela Hi
Could it be simple check for `null `or `undefined ` in `shouldShowLanguagePrefix()`?
seems like it is the place responsible for "should omit the language code in this case"

`if (!this.props.article.language) { return false; }`

or 
`if (this.props.article.language == undefined) { return false; }`
`== null` returns `true` if `this.props.article.language` is also `undefined`
`== undefined` returns `true` if `this.props.article.language` is also `null`

mauditecandela:
@testa19 indeed that was a better solution as it's more consistent with the logic of the whole article.jsx and prevents for false positives. I already did a commit for it and it's already added to my PR :) https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1413

testa19:
@mauditecandela there could be a bug introduced with your previous commits - so Travis CI build failed.
I'm not a specialist in git PRs: i would get rid of previous commits and left only the last meaningfull one.

And seems like `!this.props.article.language` does that check for `null` and `undefined`. Try to play in dev cosole in a browser.

ragesoss:
You can look at the CI build log to see the error. In this case, it's a javascript eslint error: `Unexpected whitespace before semicolon  semi-spacing`.

mauditecandela:
[...]

PR #1413, opened 2017-09-25
Fixes issue #1407: language in title

mauditecandela:
https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1407


app/assets/javascripts/components/articles/article.jsx
```diff
@@ -21,7 +21,7 @@ const Article = React.createClass({
 
   shouldShowLanguagePrefix() {
     if (!this.props.course.home_wiki) { return false; }
-    return this.props.article.language !== this.props.course.home_wiki.language;
+    return this.props.article.language !== this.props.course.home_wiki.language && null;
```
ragesoss: I'm not clear about what the `&& null` is doing here. It looks like this would always return either `false` or `null`.

mauditecandela: yes, you are right. reworking on it right now, sorry.

ragesoss:
@mauditecandela looking this over, I'm thinking that the best approach might be to have every article go through `CourseUtils.formattedArticleTitle`. That way we only have one place that is responsible for generating formatted article titles. With that approach, it may make sense to add another helper function to `CourseUtils` that takes a course object and an article object and handles the 'home_wiki' part of the current implementation in `article.jsx`.

A big benefit to moving it over to `CourseUtils` will be that it'll be easy to write unit tests for it (in `test/utils/course_utils.spec.js`).

mauditecandela:
I will try that out.

mauditecandela:
@ragesoss and @testa19 here it is a test. I am not 100% sure if the logic is the desirable, when testing it in my local environment it seemed fine but could be that I am missing something. Could you please have a look and give me some feedback? Thanks a lot.

ragesoss:
@mauditecandela there's another linting error that is breaking the build. If you run `gulp` locally, it will re-run the linter ever time you save a change so you can see any linting errors quickly. I also recommend installing an eslint plugin to your code editor, so you can see any linting errors as you're codeing.

mauditecandela:
aaaand it's green! Waiting for your feedback :) In my local machine seems to work and worked on the approach of transferring everything to course_utils.js in order to make easier to do some tests.

ragesoss:
@mauditecandela cool!

It looks like the only place that users CourseUtils.formattedArticleTitle is in assignment.jsx, and it already has an article object that will include the language and project.

So I think we should combine those into a single function that takes article and defaultWiki as arguments (which your new function does) and call it from assignment.jsx

I think `formattedArticleTitle` is a better name for that function. Changing the signature of formattedArticleTitle will cause the tests to break, so you'll need to adjust the tests as well. But it'll also be a good chance to test it more thoroughly and intuitively, around the idea of 'given an article on some wiki, and a default wiki, how do we want the title to be displayed'.

You can run the javascript tests locally by running `jest --watch`. That will rerun the tests whenever you change a js file or corresponding test.

mauditecandela:
Hi @ragesoss here my first proposal as a possible solution. As you suggested, I have merged both functions but in order to work finely, I also separated the title formatting action also from articleFromAssignment() function. I think this is more consistent, but also lead me to change different files where the article.formatted_title was expected to be available after creating the article from the assignment. 

I have also noticed that the problem that appears in this test: https://travis-ci.org/WikiEducationFoundation/WikiEduDashboard/jobs/280854551 was not appearing in my local jest, do you know why?

Btw, thanks for the hint, jest is a great tool and now I love it! 

I look forward your feedback :)

ragesoss:
@mauditecandela we have two different test frameworks: Jest for unit testing of Javascript, and RSpec for doing Ruby unit tests as well as browser tests. Those 'multiwiki assignments' tests are browser tests, so they don't get run by Jest. (If you are running `guard`, it will try to automatically run relevant tests similarly to what Jest does, but because the browser tests aren't linked to a single file, they generally don't get run unless you run them explicitly or you're running the entire test suite.)

test/utils/course_utils.spec.js
```diff
@@ -141,7 +141,7 @@ describe('courseUtils.articleFromTitleInput', () => {
   });
 });
 
-describe('courseUtils.articleFromAssignment', () => {
+describe('courseUtils.formattedArticleTitle', () => {
```
ragesoss: This is still a test of `articleFromAssignment`.

app/assets/javascripts/utils/course_utils.js
```diff
@@ -92,18 +92,12 @@ const CourseUtils = class {
     const language = assignment.language || 'en';
     const project = assignment.project || 'wikipedia';
     const articleUrl = assignment.article_url || this.urlFromTitleAndWiki(assignment.article_title, language, project);
-    const formattedTitle = this.formattedArticleTitle(
```
ragesoss: Instead of removing this and setting formatted_title in other places every time this is called, you could do article.formatted_title  = this.formattedArticleTitle(article) before the return.

If necessary, you could also make articleFromAssignment take an optional homeWiki argument to include.

In general, the more shared logic we can move into CourseUtils, the better. Isolating that logic into the shared vanilla Javascript function makes it much easier to test and maintain.

app/assets/javascripts/utils/course_utils.js
```diff
@@ -117,22 +111,22 @@ const CourseUtils = class {
     return `https://${language}.${project}.org/wiki/${underscoredTitle}`;
   }
 
-  formattedArticleTitle(language, project, articleTitle) {
+  formattedArticleTitle(language, project, title, home) {
     let languagePrefix = '';
-    if (language === undefined || language === 'en') {
+    if (!home || language === home.language || language === null || language === undefined || language === 'en') {
```
ragesoss: We can probably do away with the special-casing of 'en' here. Also, see if you can simplify this conditional any further.

test/utils/course_utils.spec.js
```diff
@@ -153,8 +153,6 @@ describe('courseUtils.articleFromAssignment', () => {
     expect(article.url).to.eq('https://es.wikipedia.org/wiki/Autofoto');
     expect(article.title).to.eq('Autofoto');
     expect(article.language).to.eq('es');
-    expect(article.project).to.eq('wikipedia');
```
ragesoss: This should still be true.

app/assets/javascripts/components/assignments/assignment.jsx
```diff
@@ -17,12 +17,10 @@ const Assignment = React.createClass({
   },
   render() {
     const article = this.props.article || CourseUtils.articleFromAssignment(this.props.assignmentGroup[0]);
-
-    article.formatted_title = CourseUtils.formattedArticleTitle(
-        this.props.assignmentGroup[0].language,
-        this.props.assignmentGroup[0].project,
-        article.title,
-        this.props.course.home_wiki);
+    article.formatted_title = CourseUtils.formattedArticleTitle(article);
```
ragesoss: This line seems unnecessary if articleFromAssignment does this already.

app/assets/javascripts/components/assignments/assignment.jsx
```diff
@@ -17,12 +17,10 @@ const Assignment = React.createClass({
...
+    if (!article.formatted_title) {
+      article.formatted_title = CourseUtils.formattedArticleTitle(this.props.assignmentGroup[0]);
```
ragesoss: Here you can do `article.formatted_title = CourseUtils.formattedArticleTitle(article, this.course.home_wiki)`

The fewer times we need to call `assignmentGroup[0]`, the more understandable this code will be.

app/assets/javascripts/utils/course_utils.js
```diff
@@ -99,10 +99,11 @@ const CourseUtils = class {
       url: articleUrl,
       title: assignment.article_title,
       article_id: assignment.article_id,
-      language,
-      project,
+      language: language,
+      project: project,
```
ragesoss: This is fine, but means the same thing as what was there before, which is es2015 shorthand notation: `{ foo }` is the same as `{ foo: foo }`.

mauditecandela:
I have already reduced the code as you suggested the only problems that I see are: 
- Any formatted title created from articleFromAssignment is not showing any language or project. I think it's because they are not defined yet. I think this was the behaviour that I saw at the beginning, can you confirm? Any other suggestions?
- There's a test that keeps failing, but I will try to fix it tomorrow :)
Any other feedback is very welcome, of course.

ragesoss:
@mauditecandela I think the way to solve both of these is to pass course.home_wiki articleFromAssignment, and set the language and project from that if it is not provided by the `assignment` argument.

mauditecandela:
This is passes all tests and adds the defaultWiki to articleFromAssignment. I also removed te unneeded part of the object.

app/assets/javascripts/components/students/assign_button.jsx
```diff
@@ -241,7 +241,8 @@ const AssignButton = React.createClass({
       let removeButton;
       let articleLink;
       ass.course_id = this.props.course_id;
-      const article = CourseUtils.articleFromAssignment(ass);
+      const article = CourseUtils.articleFromAssignment(ass, this.props.course.home_wiki);
+      if (!article.formatted_title) { CourseUtils.formattedArticleTitle(article, this.props.course.home_wiki); }
```
ragesoss: Is this necessary? Unlike in `assignment.jsx`, this article object is always generated by `CourseUtils.articleFromAssignment`, so it should always have a formatted_title.

mauditecandela: true!

app/assets/javascripts/utils/course_utils.js
```diff
@@ -117,22 +112,22 @@ const CourseUtils = class {
     return `https://${language}.${project}.org/wiki/${underscoredTitle}`;
   }
 
-  formattedArticleTitle(language, project, articleTitle) {
+  formattedArticleTitle(article, defaultWiki) {
     let languagePrefix = '';
-    if (language === undefined || language === 'en') {
+    if (defaultWiki === undefined || article.language === defaultWiki.language || article.language === null || article.language === undefined) {
```
ragesoss: These can be simplified somewhat, I think. Instead of checking article.language and defaultWiki against multiple precise conditions, we can just check for falsiness: `!defaultWiki` and `!article.language`.

ragesoss:
@mauditecandela this is almost there. I really like how much this simplifies both article.jsx and assignment.jsx.

What remains:
* Simplify the logic of the formattedArticleTitle function as much as possible.
* Update the course_utils spec to test the various possibilities more thoroughly.

mauditecandela:
Thanks a lot for all the guidance and the comments @ragesoss Here it is the new version. I added some tests and simplified the logic are formattedArticleTitle. Let me know what you think :)

ragesoss:
@mauditecandela the tests still have some room for improvement, to reflect the different situations where article titles get displayed. It looks like everywhere except for `components/students/assign_cell.jsx` that calls `articleFromAssignment()` now passes course.home_wiki, but the tests don't reflect that; no defaultWiki is included in the tests for that function. The home_wiki can be added to the call in AssignCell, and the tests should include it. In particular, that test was trying to make sure that when you assign an article from Spanish Wikipedia in the context of an English Wikipedia course, that it shows the language prefix, and that project correctly to Wikipedia.

For the tests of formattedArticleTitle, similarly, it looks like we're going to be able to use it always with the defaultWiki included, so the tests should reflect that, but they should test all the main possible situations: when the article is on the same wiki as the course; when the article is the same project but different language; when the article is same language but different project; when the article is different langauge *and* different project; and when the article is a different project that doesn't have a language (ie, Wikidata).

mauditecandela:
@ragesoss you were right, I was too tired yesterday and today I saw that the tests were far from accurate. I corrected them, please let me know what you think. I am excited! I enjoyed very much working on this. :)

test/utils/course_utils.spec.js
```diff
@@ -191,3 +207,81 @@ describe('courseUtils.hasTrainings', () => {
...
+    const article = {
+      title: 'Judith Butler',
+      language: null,
+      project: 'wikidata'
+    };
+    article.formatted_title = courseUtils.formattedArticleTitle(article, defaultWiki);
+    expect(article.formatted_title).to.eq('wikidata:Judith Butler');
+  });
+
+  it('returns true for a weeks array with trainings', () => {
```
ragesoss: Duplicate

ragesoss:
> I am excited! I enjoyed very much working on this. :)

:-)

Please add the course home_wiki to the `AssignCell` component, and remove the duplicated hasTrainings() test. I'll test it out on Monday to make sure that everything is behaving as expected, and if so, it should be good to merge.

mauditecandela:
Done!!!

Issue #1417, opened 2017-09-26
Refactor CoursesController into multiple controllers


ragesoss:
`CoursesController` is rather large, because it is responsible for a bunch of different independent things.

It could be reorganized into multiple smaller controllers. `#update` in particular has a lot of helper methods that are not reused by other actions, so most of the other actions could probably be moved to new controllers (or vice versa). One possible approach to doing that is described here: http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

testa19:
@ragesoss  Is it just another file `course_update_controller.rb` (`class CourseUpdateController < ApplicationController`)? 
(There would be nearly 170 and 100 LOC in two course-controller files if other update actions will be moved)

or is there will be some kind of Courses namespase (with folders)? CoursesController and Courses::UpdateController

Do i get the article right: other suggests controllers have to have only CRUD actions and any other actions could be moved to another controller and "translated" to CRUD there?

ragesoss:
@testa19 I don't have set answers to those questions. If you're going to take the CRUD-only strategy from the blog post, I think that would involve leaving update (and create, and destroy, and show), adding a namespace, and creating new namespaced controllers for each of the other remaining actions.

On the other hand, separating update into its own controller makes the most sense in terms of single responsibility and isolating the most complex action into it's own file. I that case, Courses::UpdateController would probably be the best way to go. Doing it that way might help set up a pattern we can use for other controllers as well to start to simplify routes.rb, which is getting pretty messy without how far we've gone so far with most routes being explicitly defined.

If you want to take this on, you can play around with it and report what you discover. I'm not set on a particular solution, but the problem to solve here is that CoursesController is confusing because it does too many things.

Aeropio:
hello @ragesoss, is this issue still open?

ragesoss:
Yes.

chepyego:
Is this issue still open? i would like to help out

chepyego:
@ragesoss is this issue stil available?

ragesoss:
yes, it is. if you're working on it, please just make a note of it here, and post an update if you stop working on it.

chepyego:
Yes am working on it,, I'll update you

chepyego:
hey, am sorry for not giving out feedback about the issue was to work on
due to COVID-19 schools were closed so getting internet has been a problem
to me and i see deadline is near ,I don't know how to start will you please
give some tips on how to go with this issue

[...]

PR #1419, opened 2017-09-29
[WIP] Splitting CoursesController (issue #1417)

testa19:
Steps on orginizing Courses Controller structure https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1417

1. Splitting CoursesController: CoursesUpdateController created (no modules or namespaces)

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -6,9 +6,7 @@ require "#{Rails.root}/lib/wiki_course_edits"
 require "#{Rails.root}/lib/list_course_manager"
 require "#{Rails.root}/lib/tag_manager"
 require "#{Rails.root}/lib/course_creation_manager"
-# require "#{Rails.root}/app/workers/update_course_worker"
 require "#{Rails.root}/app/workers/notify_untrained_users_worker"
-# require "#{Rails.root}/app/workers/announce_course_worker"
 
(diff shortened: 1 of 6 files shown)


ragesoss:
It worked! It looks like that was a relatively straightforward refactor:
* move update and its helper methods
* move the section of the specs that test `#update`
* add new explicit routing for update, and add routing exception for basic course resource routing

What do you think about this structure versus going down the namespace path?

Cleanup remaining:
* Don't change `consider_all_requests_local`
* Fully remove commented-out code that was copied to new files

testa19:
@ragesoss Oh, it was fust a first step: to see how two separate controllers wil look and work.
Now i'm trying to understand things (and make them work) with namespacing or modules:
and seems i am missing something

testa19:
@ragesoss can't make them work with namespacing: had a mess with partials - there is too much interlacement (with other controllers? in routes?) for my qualification.

ragesoss:
I'm fine with doing it the first way, if you can't figure out the namespacing solution. If it's not a straightforward refactor, then maybe it's best to not introduce namespacing, which I haven't used well enough to feel totally comfortable with myself.

ragesoss:
Please reopen if you take this up again, although it now has conflicts with the main branch.

Issue #1394, opened 2017-09-13
Update ArticleViewer to use whocolor for all available languages


ragesoss:
The whocolor API now supports Basque (`eu`) as well as English Wikipedia. `ArticleViewer` should work for Basque as well, in a way that will be easy to extend to more languages as the number supported by whocolor grows.

----
https://api.wikiwho.net/

jyoti0208:
I am working on implementing this feature.

jyoti0208:
I am not able to find ArtcleViewer on any page.
I have created a dummy campaign on my local as `http://localhost:3000/courses/Wiki/React_Training_(first)/articles`
Also, there is article_viewer.jsx in common, So if I modify anything to this file which page will get change?

ragesoss:
@jyoti0208 Have you done the steps of populating the course with data? https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/user_roles.md

jyoti0208:
Yes, I have 2 new users as participants and run the update routine.
Also, I am getting this page `http://localhost:3000/courses/Wiki/React_Training_(first)/articles`
 Is this the Article viewer page?

ragesoss:
@jyoti0208 I just noticed that now German and Turkish are available too. :)

ragesoss:
All available languages are set now, and it should be trivial to add more as they are available.

PR #1421, opened 2017-09-30
Article viewer lang

jyoti0208:
Adding eu as new language for #1394


app/assets/javascripts/components/common/article_viewer.jsx
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import OnClickOutside from 'react-onclickoutside';
 import { trunc } from '../../utils/strings';
+import SUPPORTED_LANG from '../../constants/mediawiki_lang';
```
ragesoss: Since this is only used in ArticleViewer, we may as well just do this in the same file. Also, whocolor is different from mediawiki (which is the software that runs Wikipedia and other wikis).

app/assets/javascripts/components/common/article_viewer.jsx
```diff
@@ -69,7 +70,7 @@ const ArticleViewer = React.createClass({
   },
 
   isEnWiki() {
-    return this.props.article.language === 'en' && this.props.article.project === 'wikipedia';
+    return SUPPORTED_LANG.includes(this.props.article.language) && this.props.article.project === 'wikipedia';
```
ragesoss: The name of this method should be changed; it no longer answers the question of whether it's English Wikipedia, but rather whether it's a wiki with whocolor support.

app/assets/javascripts/components/common/article_viewer.jsx
jyoti0208: I have removed this extra const file.

app/assets/javascripts/components/common/article_viewer.jsx
jyoti0208: Changed the name of method as `isWhocolorLang`

ragesoss:
Looks good!

PR #1422, opened 2017-10-02
Iss1408

jstoebel:
adds rules from `react` and `jsx-a11y` which have no current violations.


ragesoss:
`Failed with 593 errors`. Quit `gulp` and then start it again and you should see the errors.

jstoebel:
I was failing to search for errors in `.jsx` files and thus missed most of the errors. I'm going to close this and start over.

Issue #1402, opened 2017-09-18
Convert `article_url`, `full_title`, and `escaped_full_title` in ArticleHelper into Article instance methods


ragesoss:
I often find myself wanting to get the url or full title of an article by calling `article.url` or `article.full_title`, etc.

Let's replace those `ArticleHelper` methods with instance methods on `Article`. This will require changes to every place in the app that currently calls those methods.

unnatii:
I would like to take this up as my first issue

alpha721:
@unnatii are u still working on this issue , because otherwise I would be happy to take it up.

mauditecandela:
@unnatii and @alpha721 are you both still working on this issue? I find it very interesting and therefore I will be happy to take it up if you are not still working on it :)

mauditecandela:
@ragesoss I created a first proposal for this issue, you can see it in 18e9bd4. All my local tests pass. I am not sure if I should transfer the namespaces constant to the article.rb model or if I should keep them at article_helper.rb. What do you think?

PR #1424, opened 2017-10-03
#1402 Creates new instance methods for url, full_title, escaped full_title

mauditecandela:
Also replaces them in every place it was used.


ragesoss:
@mauditecandela this looks good! Failing test was unrelated to the change. Just remove the commented-out code and I think this will be good to merge.

mauditecandela:
done! :D

Issue #1387, opened 2017-09-10
BlockedEditAlert should include block data


ragesoss:
Currently we send the block data that we get from MediaWiki through Raven to Sentry, but we don't include it in the BlockedEditAlert that we email to Wiki Ed staff.

It would be very useful to save that data on the Alert record (requires a new db field) and send it in the email. That way the email would include enough info for someone to get the block sorted out. This is particularly important for when the dashboard's IP address gets blocked because of a student with a bad username; the user affected by the block will not necessarily be the one who was actually blocked, and it's hard to track down who was blocked without that block data.

mattwarren10:
What file(s) is the block data that we get from MediaWiki referenced? What would you want the new attribute in the `alerts` table to be called? `block_data`?

ragesoss:
Nothing so specific as `block_data`, since the Alert table holds all the different alert types. Maybe `details`, as a serialized Hash.

PR #1427, opened 2017-10-03
Add details attribute to alerts table and update AlertMailer #1387

mattwarren10:


db/migrate/20171003172534_add_details_to_alerts.rb
```diff
@@ -0,0 +1,5 @@
+class AddDetailsToAlerts < ActiveRecord::Migration[5.1]
+  def change
+  	add_column :alerts, :details, :text
```
ragesoss: yikes, tabs!

mattwarren10:
This is a simple way I thought of to serialize a text attribute to be a Hash. Let me know what you think.

ragesoss:
Yes, that's what I had in mind.

mattwarren10:
Is this something you were looking for? If it is, I'll take on another issue.

ragesoss:
This is good so far. It still needs to:

1. Include the details in the rendered alert email (by changing the relevant view template)
2. pass in the mediawiki response data as details to new BlockedEditAlert records

The second item starts from `lib/wiki_response.rb` and goes to `BlockedEditsWorker` and `BlockedEditsReporter`.

ragesoss:
This will be the first time we would use Redis to store something more complex than an integer record ID — passing in the details hash as a sidekiq argument in `BlockedEditsWorker` — but I think it's okay and jives with https://github.com/mperham/sidekiq/wiki/Best-Practices

That details hash is not application state, and it starts its life as plain JSON anyway, so it should survive being serialized and deserialized.

mattwarren10:
So the check isn't passing because of the amount of arguments. Since `WikiResponse` is not a model, hence it doesn't have any factories, would it be best to just include empty hashes like this? Or is there a more optimal approach?

```
describe BlockedEditsReporter do
  describe '.create_alert' do
    let(:student) { create(:user, username: 'student') }
    let(:student2) { create(:user, username: 'student2') }
    it 'creates an Alert record for a blocked edit' do
      BlockedEditsReporter.create_alerts_for_blocked_edits(student, {})
      expect(Alert.count).to eq(1)
    end

    it 'does not create multiple alerts' do
      BlockedEditsReporter.create_alerts_for_blocked_edits(student, {})
      BlockedEditsReporter.create_alerts_for_blocked_edits(student2, {})
      expect(Alert.count).to eq(1)
    end
  end
end
```

ragesoss:
I would pass it something more like the expected data. Here's roughly what an actual response that would get passed to WikiResponse looks like:

```ruby
{:servedby=>"mw1229",
 :error=>
  {:info=>"You have been blocked from editing.",
   :blockinfo=>
    {:blockedtimestamp=>"2017-09-16T14:12:01Z",
     :blockid=>7824827,
     :blockedby=>"Orangemike",
     :blockreason=>
      "{{uw-softerblock}} <!-- Promotional username, soft block -->",
     :blockedbyid=>42464,
     :blockexpiry=>"infinite"},
   :*=>
    "See https://en.wikipedia.org/w/api.php for API usage. Subscribe to the mediawiki-api-announce mailing list at &lt;https://lists.wikimedia.org/mailman/listinfo/mediawiki-api-announce&gt; for notice of API deprecations and breaking changes.",
   :code=>"blocked"}}

```

(I copied this from Sentry; the actual response would I think have the same shape, except with strings instead of symbols for all the keys.)

mattwarren10:
Should be good to go.

PR #1430, opened 2017-10-03
[WIP] remove clone links

yeli-buonya:
removed 'clone' links from:

/views/survey_assignments/_survey.html.haml

and

/views/rapidfire/question_groups/_question_group.html.haml

Todos:

update routes.rb
survey_controller.rb
any other code associated with cloning

### DIFF
--- a/app/views/rapidfire/question_groups/_question_group.html.haml
+++ b/app/views/rapidfire/question_groups/_question_group.html.haml
@@ -19,6 +19,8 @@
   .survey__admin-row__col.survey__admin-row__col--actions
     .course-list__col
       = link_to "Edit", edit_question_group_path(question_group)
+    .course-list__col
+      = link_to "Clone", "/surveys/question_group/clone/#{question_group.id}", method: :post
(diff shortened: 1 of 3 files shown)


app/views/rapidfire/question_groups/_question_group.html.haml
```diff
@@ -19,8 +19,6 @@
   .survey__admin-row__col.survey__admin-row__col--actions
     .course-list__col
       = link_to "Edit", edit_question_group_path(question_group)
-    .course-list__col
```
ragesoss: This is the cloning link for a QuestionGroup, not a Survey. This should be kept.

yeli-buonya: Okay, thanks for the update, let me correct that right away

ragesoss:
@yeli-buonya you can check the CI build to see what tests failed. In this case, I think just `survey_admin_spec` is all that was affected.

ragesoss:
Done with #1458

Issue #1408, opened 2017-09-20
Enable additional javascript style linting rules


ragesoss:
We're using eslint rules from several plugins — the airbnb config, plus some React plugins — to maintain fairly standard style and Javascript usage. However, we have many of the rules disabled via `.eslintrc`, because there were violations that appeared after an update of the eslint config.

Most of these should be re-enabled along with fixes for any violations in the codebase.

This is best done one rule per commit (and at most a few rules per PR).

jstoebel:
I'll give this one a shot!

jstoebel:
currently working on `no-find-dom-node` and `forbid-prop-types`

bryszard:
Working on `no-danger`. See #1492

ragesoss:
@nicktu12 go for it! The a11y rules are especially appreciated, as fixing the javascript so that those rules are not violated may concretely improve the accessibility of the app.

alpha721:
@ragesoss can I give some of these a try?

ragesoss:
Yep!

helenatxu:
Hello, I am an applicant for the Outreachy internship, can I take this issue for my first contribution? Is it still available?

Ajitesh13:
Hello, Can I proceed with this open issue ?

Nikuunj:
this issue still available? 
plz give guides for repo

Nikuunj:
Hi @ragesoss,

I hope this message finds you well. I'm interested in joining the WikiEducationFoundation Slack channel to stay informed and contribute to the discussions. Could you please add me to the channel?

Looking forward to being a part of the community.

utkarsh-gaur:
Hello @ragesoss , I hope you are doing well. I am willing to the WikiEducationFoundation's slack channel to keep up with latest news. It would be grateful of you if you add me to the channel.

Eagerly waiting to be a part of the discussions taking place at WikiEducationFoundation .
Regards

ragesoss:
Let me know your email address if you would like a Slack invite.

utkarsh-gaur:
Thank you for your response.
Here's my email address associated with slack.
Email: ***@***.***

utkarsh-gaur:
i want to try my hands on this issue.
can you please assign this issue to me?

ragesoss:
@utkarsh-gaur your email address did not come through on github.

utkarsh-gaur:
E-mail: utkarshgaur6014@gmail.com

[...]

PR #1432, opened 2017-10-06
refactor to not use findDOMNode

jstoebel:
partially addresses #1408 by removing `findDOMNode`


jstoebel:
after discussion with @ragesoss its preferable to refactor to use enzyme instead. Closing so I can resubmit.

PR #1434, opened 2017-10-08
Refactor dyk reducer

KarmaHater:
Task: 
`We currently have a mix of older McFly-based Flux stores (in app/assets/javascripts/store) and a Redux store built from reducers (in app/assets/javascripts/reducers). We want to transition to using Redux for everything.`

What I did:

+ Fix performance issue in feedback reducer. (You should try not use `import * as types` but just import `const` that are needed)

+ Refactor feedback reducer to use the same code style as the rest.

+ Delete McFly `DidYouKnowStore` and created Did You Know reducer.

+ Broke up action type file. (It is not a good idea to use one file because they get big real fast).

+  Wrote tests for DidYouKnow reducer.

+ Refactor `did you know handler`. I moved the `HEADERS` and `NO_ACTIVITY_MESSAGE` vars out of the component. The only code that should be in a component is what is necessary to make it run. Components get big fast removing this kind of fluff is always a good idea :) I also deleted `storeDidChange` method that was not being used.

I would like to make a suggestion. I would highly recommend we use `reselect`. It keeps, code clean, helps with performance, and prevents silly errors. The Redux part of the project is small making it the perfect time to add this helpful npm package. I would be more then willing to do it if you think it is useful.

https://github.com/reactjs/reselect

There is no logger at the moment making the project harder to work with. I was going to add it to this pull request but I thought maybe I should ask first. This is one I find extremely helpful.

https://github.com/evgenyrodionov/redux-logger

Cheers!


ragesoss:
Cool! Feel free to open a separate PR adding reselect and the logger.

PR #1437, opened 2017-10-09
add selectors

KarmaHater:
REFACTOR:

+ I add the reselect libaray (https://github.com/reactjs/reselect).
+ Refactored components that are using Redux reducers to use selectors.

Cheers!

### DIFF
--- a/app/assets/javascripts/components/activity/activity_table.jsx
+++ b/app/assets/javascripts/components/activity/activity_table.jsx
@@ -5,6 +5,7 @@ import _ from 'lodash';
 import moment from 'moment';
 
 import * as UIActions from '../../actions';
+import { selectUIOpenKey } from "../../selectors";
 import ActivityTableRow from './activity_table_row.jsx';
 import Loading from '../common/loading.jsx';
 
@@ -152,7 +153,7 @@ const ActivityTable = React.createClass({
 });
 
 const mapStateToProps = state => ({
-  openKey: state.ui.openKey
+  openKey: selectUIOpenKey(state)
 });
 
 const mapDispatchToProps = dispatch => ({
--- a/app/assets/javascripts/components/articles/article_list.jsx
+++ b/app/assets/javascripts/components/articles/article_list.jsx
@@ -4,6 +4,7 @@ import { connect } from 'react-redux';
 
 import * as UIActions from '../../actions';
 import * as ArticleActions from '../../actions/article_actions';
+import { selectUIOpenKey, selectArticleDetails } from "../../selectors";
 
 import Editable from '../high_order/editable.jsx';
 import List from '../common/list.jsx';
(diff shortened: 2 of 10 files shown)


ragesoss:
@KarmaHater it looks like this doesn't use reselect at all... I guess because none of the selectors do any data transformation?

KarmaHater:
@ragesoss Silly me I forgot to add the package.json. I will push that up soon.

KarmaHater:
I was thinking maybe it is a good time to consider using a library like immutable? I can also do this refactor over the weekend of the old reducers. 

https://facebook.github.io/immutable-js/

KarmaHater:
@ragesoss I thought about this a little more. I don't think we should add `reselect` to the package.json. However, I do think we should keep to this selector pattern because it keeps your store and util functions slimmer.  Consider the following scenario.... 

`You need to calculate the price per serving from properties on a product object and display it to the dom on the PDP page.`

You don't need to do this calculation and save it to your store and putting it in your component will bulk it up. You could make a utils function. However, I think writing functions that directly deal with your current state should be separate from utils.

This is when you can create the following selector...

```js
const selectPricePerServing = (state, id) => {
  const product =  state.products.find(id === product.id);
  return  product.price /product.meals / product.size
}
```

We could also make this better and combine selectors in the following way. 

```js

const selectProduct = (state, id) => 
   state.products.find(id === product.id);

const selectPricePerServing = (state, id) => {
  const product =  selectProduct(state, id);
  return  product.price /product.meals / product.size;
}
```

You would then use it the following way..

```js

const mapStateToProps = (state, param) => ({
  propuct: selectProduct(state, id),
  pricePerServing: selectPricePerServing(state, id)
})
```

Much cleaner right? 

This is another reason why I think this pattern is extremely useful. If you are still not convinced let me know and I will think of other scenarios.

ragesoss:
Thanks for the explanation! This makes sense for the kinds of selects you note here, but all the ones you've refactored here are trivial selectors where it seems like the redirection doesn't seem to gain us anything.

Is there any benefit to using this pattern in cases where all you're doing is something like `state.ui.openKey`? (I could see the being relevant for the User Courses refactor you're doing in another PR, where we do that `_.reject`operation for undocumented reasons; giving that selection a name and its own file makes sense to me for that.)

KarmaHater:
@ragesoss I see your point. I suppose adding it here makes it consistent everywhere. Also, coming in as a new developer I do not need to go look at the Reducer for the data structure. I know that the property I need will always be `selectPropertyName.` In my option, this would speed up development. I do not need to remember any structure like `state.ui.key.blah.blah.blah` or go back to the reducer. In this case, it is really pretty simple so hard to proves its true value but consistent is always good. I find it super helpful. Should I close this PR?

ragesoss:
I think this is worth revisiting later, once we have some more complex cases that would make use of the selector pattern; at that point, for consistency we'd also want to do it for the trivial selectors. Until then, though, I'd rather not add the extra indirection, which just adds another thing you have to wrap your head around to work with the app's data flow.

PR #1439, opened 2017-10-09
[WIP] remove findDOMNode

jstoebel:
I removed `findDOMNode` from the project. Also, refactored many tests to use enzyme.

Some tests are still failing (see `course_creator.spec.jsx:86`). For reasons I don't understand, when I `simulate('change') on a `TextInput` component, I expect the method `updateCourse` to be called, but it isn't. Any ideas on how to fix this?


ragesoss:
Done in a different PR.

PR #1443, opened 2017-10-10
Redux logger take two

KarmaHater:
Add Redux Logger for easier development

npm package added
https://github.com/evgenyrodionov/redux-logger


ragesoss:
I never figured out how to use this without breaking things.

Issue #1295, opened 2017-06-02
Expanding the length of the list of activities (Activity tab)


JoalpeBrasil:
On the P&E Dashboard, the list of activities only shows last ~50 edits; the possibility to show more than that should be available, as cohorts might be much larger than 50 students

ragesoss:
Having an option to load a larger number of activities would be pretty easy to implement; just add a button the Activity view, and pass a parameter for the max activities to fetch... more or less.

mauditecandela:
Hi! I have been having a look to this issue. Where's the activity view rendered? Do you mean the activity_table.jsx? I am having problems finding the limit of rows shown. Any hint is super welcome! Thanks a lot :)

ragesoss:
@mauditecandela in broad strokes, the data flow is:
* user navigates to the activity tab, which renders RevisionHandler component
* as it mounts, that component initiates fetching the revisions
* api.js makes a call to the course_controller show revisions endpoint, which renders `app/views/courses/revisions.json.jbuilder`. That's the file that sets the limit of 50.
* the frontend receives the set of revisions and puts it in RevisionStore
* the RevisionsList component renders all the revisions in the store.

mauditecandela:
@ragesoss thanks a lot for the explanation. 

I was having a look and I have some questions regarding this:
- Would you render all the list of revisions from the server and then show only 50 in the UI? Or add different calls based on user interaction?  I don't know how many activity could be displayed if I don't limit it from the model directly, and if that would slow down the application. What do you think?
- I also had a look to both /ui_actions.js and /server_actions.js. TBH, I don't have a very clear overview yet on how to create new actions. I will continue reading about Flux, but if you have any hint that could help me it would be very appreciated. 

Of course if you think this is not the right approach let me know :) And thanks again!

ragesoss:
@mauditecandela I think the way to handle it will be something like:
* add a 'limit' param to the controller, and use that (with a default of 50) to generate the json list.
* add some kind of UI for using the limit. Have a look at `recent_edits_handler.jsx` for an example of something similar (a check box that affects the 'scoped' param on a query).

The component itself should be able to continue with the current behavior of rendering every revision that is in the store; for both performance and simplicity, we I think it'll make sense for the limit to continue to be handled on the server, in the query that pulls the records from the database.

mauditecandela:
PR: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1446

[...]

Issue #1005, opened 2016-10-24
For edit-a-thons (at least), dashboard should give user info about when to expect updated stats


ragesoss:
A common question / point of confusion is why edits aren't showing up yet during edit-a-thons. The dashboard should make it obvious to users that data gets pulled in periodically and it may take a while to show up.

Ideally, it could show actual info about how long the update cycle takes (based on the latest one) and/or when to expect the next update.

Ladsgroup:
:+1:

keer25:
Design for displaying the message : How about a info icon that opens a popup message on hovering over it.

ragesoss:
@keer25 yes, something like that would make sense.

keer25:
The constant update that happens every 15 minutes pulls revisions and hence except for article views and commons uploads all other metrics get updated every 15mins.? What about articles creates? How often is it updated?

ragesoss:
`ConstantUpdate` runs every 15 minutes, but it exits immediately if another instance of that process is already running. Currently, updates are taking about 4 hours each on Wiki Ed production, and 8 hours each on Programs & Events Dashboard. The update process reports the time the update took to the logging service Sentry — see the `Raven.capture_message` call in `BatchUpdateLogging`.

One way to handle this could be to also record that update time somewhere else — such as using Rails.cache — and then use that to show the user a) how long each update usually takes, and b) when the last update finished.

ragesoss:
An update on this... Currently, `ConstantUpdate` updates take about 15 minutes on Wiki Ed production, and about 2 hours on P&E Dashboard. `DailyUpdate` takes about 1 hours on Wiki Ed, and 4 hours on P&E.

We now have a `Setting` model which could be used to keep track of update data, such how long each of the last 10 constant and daily updates took and which update started most recently when.

mauditecandela:
I'm checking this issue for my next contribution, but tbh I think it is to backend-ish for me to properly understand. I have several questions regarding the issue, and I would like to understand them better as it's a relevant topic at the art+feminism program. 

- I was checking the Setting model but I am unable to find the Service table. Why? I mean, is not part of the database yet? 
- What do you mean with using the Setting model in order to keep track of the updated data? Something like creating some methods that are searching for the last updates (e.g. Rails.cache.fetch('Update finished')) and counts the avge time an update needs to finish?
- About the UI, I thought it could be added to the hover bubble at articles.cross_wiki_tracking. What do you think?

[...]

PR #1446, opened 2017-10-11
adds see more button #1295

mauditecandela:
- adds button to revisions_handler and onClick method
- adds translation to english for the message in the button
- includes a limit param to the courses controller
- creates new server_action and api functions to generate the endpoints

to do:
- show the new limit in the UI

### DIFF
--- a/app/assets/javascripts/actions/server_actions.js
+++ b/app/assets/javascripts/actions/server_actions.js
@@ -76,7 +76,7 @@ const ServerActions = Flux.createActions({
 
   fetchCourseRevisions(courseId, limit) {
     const actionType = 'RECEIVE_REVISIONS';
-    return API.fetchMoreRevisions(courseId, limit)
+    return API.fetchCourseRevisions(courseId, limit)
       .then(resp => ({ actionType, data: resp }))
       .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
   },
--- a/app/assets/javascripts/actions/server_actions.js
+++ b/app/assets/javascripts/actions/server_actions.js
@@ -13,6 +13,14 @@ const ServerActions = Flux.createActions({
       .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
   },
 
+  fetchMoreRevisions(courseId, limit) {
+    const actionType = `RECEIVE_MORE_REVISIONS`;
+    return API.fetchMoreRevisions(courseId, limit)
+      .then(resp => ({ actionType, data: resp }))
+      .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
+  },
+
+
   add(model, courseId, data) {
     const actionType = `${model.toUpperCase()}_MODIFIED`;
     return API.modify(model, courseId, data, true)
--- a/app/assets/javascripts/actions/server_actions.js
+++ b/app/assets/javascripts/actions/server_actions.js
@@ -13,14 +13,6 @@ const ServerActions = Flux.createActions({
       .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
   },
 
-  fetchMoreRevisions(courseId, limit) {
-    const actionType = `RECEIVE_MORE_REVISIONS`;
-    return API.fetchMoreRevisions(courseId, limit)
-      .then(resp => ({ actionType, data: resp }))
-      .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
-  },
-
-
   add(model, courseId, data) {
     const actionType = `${model.toUpperCase()}_MODIFIED`;
     return API.modify(model, courseId, data, true)
@@ -79,6 +71,13 @@ const ServerActions = Flux.createActions({
   fetchRevisions(studentId, courseId) {
     return API.fetchRevisions(studentId, courseId)
       .then(resp => ({ actionType: 'RECEIVE_REVISIONS', data: resp }))
+      .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
+  },
+
+  fetchMoreRevisions(courseId, limit) {
+    const actionType = 'RECEIVE_REVISIONS';
+    return API.fetchMoreRevisions(courseId, limit)
+      .then(resp => ({ actionType, data: resp }))
       .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
   },
 
--- a/app/assets/javascripts/actions/server_actions.js
+++ b/app/assets/javascripts/actions/server_actions.js
@@ -74,7 +74,7 @@ const ServerActions = Flux.createActions({
       .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
   },
 
-  fetchMoreRevisions(courseId, limit) {
+  fetchCourseRevisions(courseId, limit) {
     const actionType = 'RECEIVE_REVISIONS';
     return API.fetchMoreRevisions(courseId, limit)
       .then(resp => ({ actionType, data: resp }))
--- a/app/assets/javascripts/components/revisions/revisions_handler.jsx
+++ b/app/assets/javascripts/components/revisions/revisions_handler.jsx
@@ -26,12 +26,9 @@ const RevisionHandler = React.createClass({
   },
 
   showMore() {
-    if (this.state.limit === 50) {
-      this.state.limit = this.state.limit + 50;
-    } else {
-      this.state.limit = this.state.limit + 100;
-    }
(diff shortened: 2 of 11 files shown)


mauditecandela:
@ragesoss this is my first commit to the topic. Afaik, it works from server side perspective but not in the frontend. I will check now how to serve the new limit of rows. meanwhile, I thought it could be nice if you could give me some feedback, specially on creating new methods for the server_actions and api. Thanks a lot!

app/assets/javascripts/actions/server_actions.js
```diff
@@ -13,6 +13,14 @@ const ServerActions = Flux.createActions({
       .catch(resp => ({ actionType: 'API_FAIL', data: resp }));
   },
 
+  fetchMoreRevisions(courseId, limit) {
+    const actionType = `RECEIVE_MORE_REVISIONS`;
```
ragesoss: I think the disconnection with rendering the revisions is here. The action type determines which stores will receive and process the fetched data, and there is no store that handles `RECEIVE_MORE_REVISIONS`. If you change it to `RECEIVE_REVISIONS`, that may fix it.

I would suggest, rather than 'fetchMoreRevsions`, you just call this `fetchRevisions`, and use the same action (just with the default limit) in componentWillMount.

app/assets/javascripts/utils/api.js
```diff
@@ -66,6 +66,23 @@ const API = {
     });
   },
 
+  fetchMoreRevisions(courseId, limit) {
```
ragesoss: Similarly, here you can probably just adjust fetchRevisions to take a limit, and then stick with a single fetch method here.

app/controllers/courses_controller.rb
```diff
@@ -215,6 +216,10 @@ def set_endpoint
     @endpoint = params[:endpoint] if SHOW_ENDPOINTS.include?(params[:endpoint])
   end
 
+  def set_limit
+    @limit = (params[:limit] || 50)
```
ragesoss: This points to a problem with how we've set up the multipurpose `show` endpoint for courses. The limit is relevant only for a single endpoint, but this will set it for every endpoint. In the short term, another option might be to handle the limit param inside the jbuilder view.

In the longer term, I think we need to rewrite the routing and controller to have an explicit endpoint for each type of json data for a course.

app/assets/javascripts/actions/server_actions.js
mauditecandela: What do you exactly mean? 'fetchRevisions' already exist and I think it does retrieve different endpoint (revisions for each student). I have tried and didn't work, but I will check it further. 
I added though 'RECEIVE_REVISIONS' and indeed works. Thanks.

app/controllers/courses_controller.rb
mauditecandela: I limited the set_limit to the revisions endpoint with an IF statement, as I think it's clearer if it's in the controller. Why do you think it's better to add it to the json jbuilder?

app/assets/javascripts/actions/server_actions.js
ragesoss: You're right, I misread the code. Currently, the default fetch revisions for the course is going through API.fetch(courseId, model). So the way to avoid using two different methods for basically the same thing, you could call the new method `fetchCourseRevisions`, and use it instead of general-purpose fetch method for both the default and 'more' data flows.

app/controllers/courses_controller.rb
```diff
@@ -217,7 +217,7 @@ def set_endpoint
   end
 
   def set_limit
-    @limit = (params[:limit] || 50)
+    @limit = (params[:limit] || 50) if params[:endpoint] = 'revisions'
```
ragesoss: Should be `==`

app/controllers/courses_controller.rb
ragesoss: I'm okay with the if statement. That should make it clear that this is only relevant for the one endpoint.

mauditecandela:
@ragesoss the tests should pass. 
I added states in the revision_handler in order to define the number of activities to show. I also changed the name of the fetching method + added to componentWillMount in order to have the same structure. Let me know what do you think! :)

app/assets/javascripts/components/revisions/revisions_handler.jsx
```diff
@@ -8,17 +8,29 @@ const RevisionHandler = React.createClass({
...
+
+  getInitialState() {
+    return {
+      limit: 50
+    };
   },
 
   componentWillMount() {
-    return ServerActions.fetch('revisions', this.props.course_id);
+    this.setState({ limit: 100 });
```
ragesoss: This shouldn't be in componentWillMount, I think.

app/assets/javascripts/utils/api.js
```diff
@@ -65,6 +65,23 @@ const API = {
       });
     });
   },
+  
+  fetchMoreRevisions(courseId, limit) {
```
ragesoss: This should be named fetchCourseRevisions as well.

app/assets/javascripts/components/revisions/revisions_handler.jsx
mauditecandela: Do you mean that it's a mistake or that shouldn't be placed there?

ragesoss: This hook runs immediately when the component starts to load, so it would overwrite the default of 50 from getInitialState. I think you also can't set state from this React lifecycle hook anyway.

mauditecandela: I thought so but it works in my machine. I tried it with lower limit numbers and worked perfectly. componentWillMount was both taking the limit from the getInitialState and setting it on 100 for the next interaction. The same with the setState in showMore(), it worked as expected.

app/assets/javascripts/utils/api.js
mauditecandela: Done! 👍

mauditecandela:
I don't understand why the last commit could deliver to that error 😓 Do you think it's related?

ragesoss:
Nope, that's unrelated. It's a flaky feature spec that fails randomly on occasion. I restarted the build; it should pass this time, unless it gets unlucky twice in a row.

app/assets/javascripts/components/revisions/revisions_handler.jsx
ragesoss: I see we have setState a few places elsewhere within componentWillMount, so that seems to be okay technically. But it should be necessary as far as I can tell.

app/assets/javascripts/components/revisions/revisions_handler.jsx
```diff
@@ -8,17 +8,28 @@ const RevisionHandler = React.createClass({
...
+    return ServerActions.fetchCourseRevisions(this.props.course_id, this.state.limit);
   },
 
   sortSelect(e) {
     return UIActions.sort('revisions', e.target.value);
   },
 
+  showMore() {
+    this.setState({ limit: this.state.limit + 100 });
+    return ServerActions.fetchCourseRevisions(this.props.course_id, (this.state.limit + 50));
```
ragesoss: I don't understand what is going on here. Why is setState adding 100, but +50 is used in the following line?

mauditecandela: I agree this solution was rather confusing - tbh I was not that happy with it. I just commited a new one, let me know if you think it's better. :)

ragesoss: Huh... I didn't think this would work, but it does... I guess it's possible to assign to this.state directly, even though it's considered "wrong": https://reactjs.org/docs/state-and-lifecycle.html#using-state-correctly

But I don't think we need to treat the first 'more' any differently from the rest. Here's what I suggest:

```
const newLimit = this.state.limit + 100;
this.setState({ limit: newLimit });
return ServerActions.fetchCourseRevisions(this.props.course_id, newLimit);
```

ragesoss:
@mauditecandela nice work! I'll go ahead a merge this. A further improvement would be to only show the button if the number of revisions at or close to the limit. That way, it won't appear a) while fetching of new revisions is happening, and b) when all the available revisions are already being shown.

mauditecandela:
@ragesoss Thanks for the hint! :) I was planning on having a look to issue #1005 but I will check what you are mentioning as well.

ragesoss:
@mauditecandela I opened a new issue, #1451, for it, but see my notes there... it might be trickier than it seems.

Issue #1342, opened 2017-08-07
Convert remaining McFly Flux actions/stores to Redux


ragesoss:
We currently have a mix of older McFly-based Flux stores (in `app/assets/javascripts/store`) and a Redux store built from reducers (in `app/assets/javascripts/reducers`). We want to transition to using Redux for everything.

The easier ones are the Flux stores that don't use the generic StockStore as the basis. `ConfirmationStore`, `DidYouKnowStore`, `UserCoursesStore`, and some others are among the easier ones. `CourseStore` is a non-StockStore one that is fairly complex, especially because it receives so many different actions. Ultimately, with StockStore, I think we should implement each of the specific stores individually, with only the explicit actions required, rather than trying to re-implement the generic model store like StockStore does.

jyoti0208:
@ragesoss I would like to work on this.

ragesoss:
@jyoti0208 cool. Once you've poked around a bit, you should note which one(s) you're working on. One at a time is probably best.

jyoti0208:
@ragesoss I have to convert all McFly-based Flux stores to the redux store?

ragesoss:
@jyoti0208 that's the ultimate goal, but it can be done one store at a time.

KarmaHater:
@jyoti0208 @ragesoss Could I also work on this? Maybe, I start with `DidYouKnowStore` ?

KarmaHater:
@jyoti0208 Here is the pull request for the DidYouKnow reducer. 

https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1434

I will move onto `UserCoursesStore` I expect to be done by October 11, 2017 :)

KarmaHater:
I just finish UserCourseStore.

@ragesoss 

https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1448

ragesoss:
Done!

PR #1448, opened 2017-10-12
Refactor user courses store reducer

KarmaHater:
Refactor of `User Course` mc fly store to reducer.

Issue: 
https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1342

Some side notes...

I had to export the component two different ways.

```
export const ConnectedCourseCreator = connect(mapStateToProps, mapDispatchToProps)(CourseCreator);
 export default CourseCreator;
```
I did this because of testing. ` __Rewire__ ` will only let me test a component if I export it using default.  This component is really big and using a lot of stuff, mc store, internal state, and now recompose making it almost impossible to test. I would suggest refactoring this component. Maybe use a Container component to handle the state and smaller components that renders UI. This would make it super easy to do the UI tests that are already included. 

I would love to come back and do this refactor when I have more time.


app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -27,15 +29,19 @@ import { getDefaultCourseType, getCourseStringPrefix, getUseStartAndEndTimes } f
 const getState = () => {
   return {
     course: CourseStore.getCourse(),
-    error_message: ValidationStore.firstMessage(),
-    user_courses: _.reject(UserCoursesStore.getUserCourses(), { type: 'LegacyCourse' })
+    error_message: ValidationStore.firstMessageprops
```
ragesoss: This is what broke the course creation feature spec.

PR #1450, opened 2017-10-13
fix for following bug: Course pages can end up in a broken state after a series of clicks 

heysusgarcia:


ragesoss:
@heysusgarcia thanks much! nice work find the cause of that bug!

PR #1457, opened 2017-10-14
Index articles on title

KarmaHater:
I decided to fix all missing and unused indexes in the DB. This is the gem I used to help with the task. 

https://github.com/gregnavis/active_record_doctor

### DIFF
--- a/db/migrate/20171014204201_add_missing_indexes.rb
+++ b/db/migrate/20171014204201_add_missing_indexes.rb
@@ -1,0 +1,27 @@
+class AddMissingIndexes < ActiveRecord::Migration[5.1]
+  def change
+    add_index :alerts, :subject_id
+    add_index :articles, :wiki_id
+    add_index :articles, :title
+    add_index :assignment_suggestions, :user_id
+    add_index :assignments, :article_id
+    add_index :assignments, :course_id
+    add_index :assignments, :user_id
+    add_index :assignments, :wiki_id
+    add_index :blocks, :gradeable_id
+    add_index :campaigns_courses, :campaign_id
+    add_index :campaigns_courses, :course_id
+    add_index :courses, :chatroom_id
+    add_index :courses, :home_wiki_id
+    add_index :feedback_form_responses, :user_id
+    add_index :gradeables, :gradeable_item_id
+    add_index :rapidfire_answer_groups, :course_id
+    add_index :revisions, :ithenticate_id
+    add_index :revisions, :mw_page_id
+    add_index :revisions, :mw_rev_id
+    add_index :training_modules_users, :training_module_id
+    add_index :user_profiles, :user_id
+    add_index :users, :chat_id
+    add_index :users, :global_id
+  end
+end
--- a/db/migrate/20171014212112_remove_unneeded_index_revisions_on_article_id.rb
+++ b/db/migrate/20171014212112_remove_unneeded_index_revisions_on_article_id.rb
@@ -1,0 +1,5 @@
+class RemoveUnneededIndexRevisionsOnArticleId < ActiveRecord::Migration[5.1]
+  def change
+    remove_index :revisions, name: :index_revisions_on_article_id
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171014204201) do
+ActiveRecord::Schema.define(version: 20171014212112) do
 
   create_table "alerts", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
     t.integer "course_id"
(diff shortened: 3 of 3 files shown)


ragesoss:
Interesting! I wonder what the performance cost of adding all the indexes is. For most of these, we probably don't do any queries that would hit the index, but for a few we might. I'll want to poke around a little bit, and maybe run this on staging to see if there are any noticeable performance effects.

KarmaHater:
@ragesoss no problem I'm going to clean up the commit history still getting used to using a forked repo.

ragesoss:
@KarmaHater hmmm... on thinking it through a little more and reading up a bit, I think I'm good with adding indexes for all the foreign key relationships. Some of these are not foreign key relationships, though — at least, not foreign keys within the dashboard database.

Were all these indexes recommended by that gem, or is it a mix of that plus others?

KarmaHater:
@ragesoss This was the result from the gem. Should I go back and remove everything that is not a foreign key relationships?

ragesoss:
Yeah, I think that would be good. My understanding is that adding an index adds a bit of time to transactions whenever the indexed column is changed, so unnecessary indexes can hamper performance on the write side (without helping anything on the read side). We do a lot more reading than writing, but I think a safe approach would be to index the foreign keys (as this is a default if you generate a model in certain ways, anyway) but only add other indexes based on specific performance needs.

KarmaHater:
@ragesoss The only index that was added that is not a foreign key it the title for the article. However, this is what the issue asked to be added. Can you just clarify that you want me to remove article title index and keep all the other foreign key index?

ragesoss:
I closed the issue because we already have an index which addresses the need that prompted it: we now do queries on wiki+title+namespace, and we have an index for that. I'll add comments explaining the things that are not foreign keys.

db/migrate/20171014204201_add_missing_indexes.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_index :articles, :title
+    add_index :assignment_suggestions, :user_id
+    add_index :assignments, :article_id
+    add_index :assignments, :course_id
+    add_index :assignments, :user_id
+    add_index :assignments, :wiki_id
+    add_index :blocks, :gradeable_id
+    add_index :campaigns_courses, :campaign_id
+    add_index :campaigns_courses, :course_id
+    add_index :courses, :chatroom_id
```
ragesoss: This is not a foreign key; there is no Chatroom model. The chatroom id is an id for a separate service.

db/migrate/20171014204201_add_missing_indexes.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_index :courses, :home_wiki_id
+    add_index :feedback_form_responses, :user_id
+    add_index :gradeables, :gradeable_item_id
+    add_index :rapidfire_answer_groups, :course_id
+    add_index :revisions, :ithenticate_id
```
ragesoss: This is not a foreign key within our database.. It's the id we use to pass to an external service to get a plagiarism report.

db/migrate/20171014204201_add_missing_indexes.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_index :revisions, :mw_page_id
```
ragesoss: This is not quite a foreign key; it's the page ID in mediawiki for a particular page, but because we store revisions from multiple wikis, it would be possible for two revisions from different wikis to have the same mw_page_id.

db/migrate/20171014204201_add_missing_indexes.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_index :revisions, :mw_rev_id
```
ragesoss: Likewise, this is a key within the mediawiki database, but it's not guaranteed to be unique in our database.

db/migrate/20171014204201_add_missing_indexes.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_index :training_modules_users, :training_module_id
+    add_index :user_profiles, :user_id
+    add_index :users, :chat_id
```
ragesoss: Same as with chatroom_id, this is not a foreign key within the context of our database.

db/migrate/20171014204201_add_missing_indexes.rb
```diff
@@ -0,0 +1,27 @@
...
+    add_index :users, :global_id
```
ragesoss: This is not actually a foreign key within our database — is the global user id for a multi-wiki user account in Wikimedia — but there might be occasions where we would search via this. I think this one is okay to keep.

db/migrate/20171014204201_add_missing_indexes.rb
```diff
@@ -0,0 +1,27 @@
+class AddMissingIndexes < ActiveRecord::Migration[5.1]
+  def change
+    add_index :alerts, :subject_id
+    add_index :articles, :wiki_id
+    add_index :articles, :title
```
ragesoss: Not necessary, as we have one for a combination of title + wiki + namespace.

Issue #1452, opened 2017-10-13
Upgrade enzyme to version 3


ragesoss:
The enzyme javascript testing library has a breaking change for version 3; it looks like we need to add additional configuration in the form of an 'adapter' to make enzyme work.

Edit package.json to change enzyme to the latest release, then run `yarn`, then run `jest` to see the error messages, then make the `jest` tests pass again.

ragesoss:
https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md

sljohnson32:
@ragesoss Hi there! I’m a first-time contributor and was hoping to help out with this issue. I noticed nobody was assigned to it, but if there’s already a solution in progress I’m happy to try helping out elsewhere. Thanks!

ragesoss:
@sljohnson32 cool, go for it. How did you find your way to this project?

sljohnson32:
@ragesoss Awesome!  I am a student at Turing School of Software & Design in Denver and we have a project to contribute to an open source project.  Our instructor @brittanystoroz gave us this repo as an option.

ragesoss:
Thanks much. @brittanystoroz, how did *you* find your way to this project?

brittanystoroz:
@ragesoss Haha I honestly don't remember, I made a lot of pleas on twitter for people to help me find good open-source projects for my students to work on, which is really difficult. Someone must have linked me to it on one of those threads.

PR #1460, opened 2017-10-16
Update enzyme to v3.1.0

sljohnson32:
This PR addresses issue #1452 

After updating the version of Enzyme to 3.1.0, we added enzyme-adapter-react-15 to the dev dependencies and used it to configure an adapter in the testHelper.js. We also installed react-test-renderer which we are using to render React components as pure JS objects. 

Based on the documentation for upgrading [here](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md), we updated tests to pass by replacing any references to `.nodes()` with `.getElements()` and any references to `.node()` with `.instance()`.


ragesoss:
@sljohnson32 the javascript test suite is failing hard with this change. You can `jest` locally to run the js tests and see the errors.

sljohnson32:
@ragesoss We just noticed those errors after installing Yarn and running tests again locally.  We will work on getting those fixes in tomorrow.

sljohnson32:
@ragesoss We pushed up a fix for the testing issue.  All tests are now passing.

ragesoss:
@sljohnson32 thanks. I fixed the package.json conflicts on a new branch, and will merge that if it passes.

PR #1465, opened 2017-10-17
add packages needed to update to react 16

KarmaHater:
Updated React to 15.6.2.

+ add `react-addons-shallow-compare`
+ add `prop-types`
+ add `create-react-class`
+ add `react-addons-shallow-compare`
+ add `react-test-renderer`
+ update `react-select`
+ update `react-typeahead`


ragesoss:
Cool! I take it that the remaining PropTypes warnings are from npm packages rather than our code?

KarmaHater:
@ragesoss Yes, I updated a few which solved many of the warnings but there were a few I could not fine. I also could not solve the `isMounted()` issue. I believe this also has something to do with an npm package.

PR #1466, opened 2017-10-18
#1005 Temporary Fix for not updated stats

mauditecandela:
- Adds an info bubble to the home view
- Warns the user of the possible lack of updated statistics
Note: This is only a temporary fix, as the real fix would add a avge time of delay + avge time needed for the update, as stated in #1005


ragesoss:
I don't think the mouseover info on the Articles Edited count will be an intuitive place to look for info about how the update process works, as it's about all the stats rather than just the article count.

I would rather find a different design solution for how to present info about the update cycle.

mauditecandela:
I will work in another proposal during the weekend! same as with issue #1005 ;)

mauditecandela:
Here another option with the UI. In order to make it clearer what the stats are showing and which information to find, I added a header and an Info bubble that will display the information when hovering.

Issue #1445, opened 2017-10-10
Javascript errors don't include source file line numbers in stack trace


ragesoss:
It's harder than it should be to track down the source of errors, because even in a development environment, the console stack traces use the compiled javascript rather than the source files.

There should be some sort of configuration that will make source mapping work better for our build pipeline.

heysusgarcia:
@ragesoss i will look into this one. Was it working at some point?

ragesoss:
Go for it! It was never working before, but it's something available in the
dev environments of many projects. Not really sure how to make it happen
though.

heysusgarcia:
i created PR #1470

PR #1470, opened 2017-10-20
issue #1445: Edited webpack.js so JS errors include source file line numbers in stack trace

heysusgarcia:
After doing some research ([devtool config](https://webpack.js.org/configuration/devtool/)) it seems like cheap-module-eval-source-map is the best option, but for some reason it's not working correctly. None of the devtool options that contain `eval` in them appear to work correctly. I'm not sure if it's due to the webpack version the project is set up on or some other reason. Using the `cheap-module-source-map` will give us what we want. 

screen shot 2017-10-19 at 7 12 52 pm

The documentation says the `cheap-source-map` option results in a slightly faster build & rebuild performance but I thought it could be a little confusing looking at transformed code if you're viewing the sources within the dev console. Let me know if you think we should go about this some other way.

### DIFF
--- a/gulp/tasks/webpack.js
+++ b/gulp/tasks/webpack.js
@@ -84,7 +84,7 @@ gulp.task('webpack', ['bower'], (cb) => {
       'i18n-js': 'I18n'
     },
     plugins: wpPlugins,
-    devtool: config.development ? 'cheap-module-eval-source-map' : 'source-map'
+    devtool: config.development ? 'cheap-module-source-map' : 'source-map'
   };
 
   const wp = webpack(wpConf);


ragesoss:
Sweet! This only worked for me in Chrome by default, but I was able to make it work in Firefox too by changing a config setting: `devtools.sourcemap.locations.enabled` set to `true` via about:config.

FedericoEsparza:
@ragesoss I know you just merged this, but it caught my eye, because I had reviewed the hot-dev/webpack-related code a week or two ago, AND we just encountered the issue in the AV React Mob, today.
 
I would recommend just going with `'eval"` for the devtool value:
 

 
If you want to keep the conditional code intact, in case there's future revisions, you could go with `devtool: config.development ? 'eval' : 'eval',`
 
It's the fastest build option, and works in more cases, because it regenerates the files and lines, With some of the other options you'll still get `main.js` references in the browser console errors.
 
See https://webpack.js.org/configuration/devtool/ for reference.

ragesoss:
@FedericoEsparza Nice! `eval` works even better than `cheap-module-source-map`. I think I want to keep `source-map` for production builds, though, as `eval` results in files about 3 times as large.

FedericoEsparza:
@ragesoss Good point. Glad that helps. Thanks for the feedback.

PR #1475, opened 2017-10-25
adds temporary fix for statistics #1005

mauditecandela:
- Adds UI for #1005 
- Adds a message that warns the user that the course metrics might not be updated


mauditecandela:
@ragesoss this is the temporary fix for the statistics of the issue #1005 A review is appreciated :)
I have decided to split the UI and the computation in different PRs as they are independent by the current state. 
In this PR I only show a possibility on how to show the time left for the statistics to update. Travis is giving me an error I don't understand. Do you know what could happen?

ragesoss:
Thanks for splitting these PRs! Very useful.

Looks like travis is behaving now?

update ui

I think adding a header like this just to add a place to attach the mouseover is not worth the tradeoff. The header isn't necessary otherwise, and it adds a significant amount of vertical space before the real content, pushing more of it below the fold.

I still haven't thought of a perfect solution — @sejalkhatri, any ideas? — but maybe a good-enough approach would be a small line of text underneath the course stats.

ragesoss:
merged in #1485

PR #1482, opened 2017-10-31
remove no find dom node

jstoebel:
`findDOMNode` is no longer the recommended. I've removed it.

The tests effected by this change have also been switched to Enzyme which is the general direction we'd like the test suite to move in.


ragesoss:
Sweet! Looks good to me.

Issue #1473, opened 2017-10-21
In the Program Clone Page, programs should be displayed from newest to oldest in the dropdown list.


sejalkhatri:
Similar to issue #1472 

![clone](https://user-images.githubusercontent.com/14351471/31852236-dff3c13e-b691-11e7-9fd7-a44ae6d7e306.JPG)

PR #1483, opened 2017-11-01
order dropdown choices for cloning course

appleJax:
clonecourseordered

Order "Clone Course" dropdown choices from newest to oldest.

References: #1473

### DIFF
--- a/app/assets/javascripts/components/course_creator/course_creator.jsx
+++ b/app/assets/javascripts/components/course_creator/course_creator.jsx
@@ -220,7 +220,10 @@ const CourseCreator = createReactClass({
     const cloneOptions = showNewOrClone ? '' : ' hidden';
     const controlClass = `wizard__panel__controls ${courseFormClass}`;
     const selectClass = showCloneChooser ? '' : ' hidden';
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -220,7 +220,10 @@ const CourseCreator = createReactClass({
     const cloneOptions = showNewOrClone ? '' : ' hidden';
     const controlClass = `wizard__panel__controls ${courseFormClass}`;
     const selectClass = showCloneChooser ? '' : ' hidden';
-    const options = this.props.user_courses.map((course, i) => <option key={i} data-id-key={course.id}>{course.title}</option>);
+    const options = this.props.user_courses
+                              .sort((a, b) => b.id - a.id) // Newest to Oldest
```
ragesoss: Instead of writing our own sort function, this will be clearer if we use lodash's sortByOrder function: `_.sortByOrder(courses, 'id', 'desc').map(...)`

ragesoss: Yes, I agree... that's a better place to do the sorting.

appleJax:
@ragesoss I heard about it through AgileVentures.

PR #1484, opened 2017-11-02
Adds Last Update information for the user #1005

mauditecandela:
This PR adds an information bubble at the courses overview to give the user more information about the statistics available. This info is create through a new model (metrics_updates) that writes the information in the setting record. 
I also created a new method in the course method that retrieves this information and added the info to the json builder so it's available by the frontend. 
I added the hree locales the information needed (EN, ES, CA).
Any review/comment is welcome! Thanks :)


app/models/course.rb
```diff
@@ -265,6 +265,10 @@ def average_word_count
     word_count / user_count
   end
 
+  def last_update
```
ragesoss: This is not specific to a course, so it shouldn't be a method on course. MetricsUpdate can provide this directly.

app/models/metrics_updates.rb
```diff
@@ -0,0 +1,9 @@
+class MetricsUpdates
```
ragesoss: Can you think of a better name that `MetricsUpdate`? Maybe `UpdateLog`, since the job of this is to log data about the update cycles? Or something else, but I think `MetricsUpdate` doesn't clearly communicate what this is for.

app/models/metrics_updates.rb
```diff
@@ -0,0 +1,9 @@
...
+
+  def self.last_update(time)
```
ragesoss: This method is a setter, so maybe call it `last_update=` or `log_update` or something else that communicates what the method does.

Relatedly, `last_update` should probably be the getter that you put in `course.rb`.

ragesoss:
Merged in #1485

Issue #1289, opened 2017-05-26
Formatting of datetimes in Revision list is not localized


ragesoss:
The datetimes we show come from moment: `moment(this.props.revision.date).format('YYYY-MM-DD   h:mm A')`.

That's not a standard format in some locales. Brazilian Portuguese, for example, is normally DD/MM/YYYY.

PR #1485, opened 2017-11-03
New Ux for update log #1005

mauditecandela:
With this PR I create a new proposal for the statistics update UX. In this case I just added an extra line under the stats, so it doesn't take that much vertical space. 
I created a separate PR so is easily merged. This PR also includes the changes of https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1484


mauditecandela:
captura de pantalla 2017-11-03 a las 17 16 11
Here's an overview of what the new proposal would look like. 

Note: Tests are now green but they were previously red as replica.rb randomly failed.

ragesoss:
Good. I suggest aligning that text to the right, though, and keeping it as short as possible.

mauditecandela:
Done! Let me know what you think :)

ragesoss:
@mauditecandela cool. Will have a look soon.

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -121,6 +121,9 @@ const Overview = createReactClass({
         { syllabusUpload }
         <CourseStats course={this.state.course} />
         {userArticles}
+        <div className="pull-right">
+          <small>{I18n.t('metrics.are_updated')}. {I18n.t('metrics.last_update')}: {this.state.course.last_update !== 0 ? this.state.course.last_update : '-'}</small>
+        </div>
```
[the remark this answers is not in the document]
ragesoss: In particular, what I think we'll want to do is to maybe use moment-with-locales to show the user how long ago the last update was., in words and localized... like "2 hours ago". That should definitely wait until later; moment-with-locales may also be the solution for #1289.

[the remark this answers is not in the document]
mauditecandela: @ragesoss I don't understand what you mean with not to show it for "Wiki Education Dashboard". Could you explain a bit more please? Thanks.

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -116,11 +116,21 @@ const Overview = createReactClass({
       <div className="sidebar" />
     );
 
+    let courseStatistics;
+    if (!this.state.course.ended || !Features.wikiEd) {
```
ragesoss: I think you need && here.

mauditecandela: True.

lib/data_cycle/batch_update_logging.rb
```diff
@@ -59,6 +59,7 @@ def log_end_of_update(message)
     log_message 'Update finished'
     total_time = distance_of_time_in_words(@start_time, @end_time)
     Rails.logger.info "#{message} Time: #{total_time}."
+    UpdateLog.log_update(@end_time.to_datetime) unless message == 'Daily update finished.'
```
ragesoss: Relying on that message string is fragile... especially with matching on an `unless` condition. If a DailyUpdate fails partway through, it still hits this code but with the message 'Daily update failed'.

Maybe instead do `if self.class == ConstantUpdate`.

mauditecandela: Definitely. I was certainly not sure about using the message. I just pushed your proposal asi I think it's a better solution, indeed.

ragesoss: Cool. Assuming the tests pass — I *think* that suggestion will work, but not entirely sure — I'll merge this. I think later we'll want to log more info, including info about all types of updates (we also have SurveyUpdate in addition to Constant and Daily), but for an initial version, this looks solid.

PR #1486, opened 2017-11-06
hardcode bower dependencies to allow gulp to run main-bower-files without bower.json 

### LINKED ISSUE #1453: Migrate away from bower

tansaku:
fixes #1453


tansaku:
Notes
-----

history
* bower - package manager for front end js
  - bower.json --> downloads files in correct place 
  - .bowerrc --> everything goes in `vendor`
  - bower not deterministic in terms of version
  --> use npm or yarn that have lock files etc.

* replacing bower with yarn
  - yarn also a package manager
  - works of package.json
  - everything installed in node_modules

* gulp
  - orchestrates all of things that needs to be done for asset 
  - replacement for asset pipeline
  - creates vendor.js (which includes all project js requirement) - and minified css
  - listjs and other things that don't have to be here
  - gulp main_bower_files is reading bower.json and has expectations about vendor

* bower-away 
  - interprets bower.json and works out what would yarn do
  - then I'm going to add that stuff into package.json
  - and then put it inside node_modules in a special dir
  - so then that special dir can be used in the build



yarn replacing bower
bower-away - moves bower stuff into another location

gulp-tasks/bower.js <-- takes all the bower files that we need and it concats them into vendor.js 

tricky thing is the main bower files - what are the main js files from all these vendor distribution packages and tries to work out the best minified js file for each

bower-away doesn't handle this above gulp/bower config

Sage approach
---------------

in nodejs console - require bower files and output an array of files, and that will define all the files that we need - and that could be hardcoded and rewrite it to be the new path to copy the bower-away move to node modules trick


Another approach?
------------------

make everything available in vendor.js globally (by moving all from bower.json to package.json), but perhaps not such a good approach 

Another approach?
------------------

alternative gulp package that works with yarn??

tansaku:
I just switched to work on another branch, but fell foul of having the symlink setup:

```
lrwxr-xr-x     1 tansaku  staff     103  6 Nov 18:25 vendor -> /Users/tansaku/Documents/GitHub/WikiEducationFoundation/WikiEduDashboard/node_modules/@bower_components
```

we might need some instructions for how to handle that for working with this branch - or switching from one approach to the other

ragesoss:
@tansaku this never passed the feature specs on travis, so I think something with generating vendor.js worked differently than it did for us locally.

tansaku:
I wonder if travis can handle the symlinks ...

tansaku:
ah, looks like `bower install` is still being run - maybe need to remove that from the travis instructions ...

ragesoss:
Yeah, I made a branch earlier that cleaned up travis.yml, but it was failing the same way this one is now:

```
ActionController::RoutingError:

              No route matches [GET] "/assets/javascripts/vendor.js"
```

I take that to mean that vendor.js didn't get created, or it got created in the wrong place.

ragesoss:
Please reopen if you take this back up. This work will be useful when someone takes another crack at the problem, but we can open a new PR then.

Issue #1487, opened 2017-11-06
ArticleViewer cuts off author list if it's too long


ragesoss:
The ArticleViewer's author list at the bottom doesn't handle large numbers of usernames very well, and if it ends up at three rows worth, it can extend beyond the visible area (depending on screen size).

![cutoff](https://user-images.githubusercontent.com/848483/32463071-866a7804-c2f0-11e7-87f5-5c8d4a50b1c8.png)

The handling of that section should be adjusted so that it can be viewed even with very long author lists (even if scrolling is necessary). The styling of it when it stretches to two rows also has lots of room for improvement.

The relevant styles are in `app/assets/stylesheets/modules/_article_viewer.styl`.

It may or may not require some adjustments in the `ArticleViewer` React component.

ragesoss:
See this bug in your own browser here: https://dashboard.wikiedu.org/courses/Marshall_University/Principles_of_Cell_Biology_(Fall_2017)/articles

PR #1489, opened 2017-11-10
Add overflow: scroll to ArticleViewer author list

appleJax:
Prevents the author list from flowing off the page. Bug viewable on the following link:
https://dashboard.wikiedu.org/courses/Marshall_University/Principles_of_Cell_Biology_(Fall_2017)/articles

The rows after the first row are accessible by scrolling, but they are kinda hidden. I can increase the `height` of `.user-legend-wrap` to `60px` to show the top of the second row, but it didn't look good to me.

screen shot 2017-11-09 at 9 17 31 pm

References #1487

### DIFF
--- a/app/assets/stylesheets/modules/_article_viewer.styl
+++ b/app/assets/stylesheets/modules/_article_viewer.styl
@@ -5,7 +5,7 @@
   /*@noflip*/ left 50%
   width 100%
   max-width 1200px
-  max-height calc(95vh \- 70px)
+  max-height calc(95vh \- 55px)
   z-index 10
   border 1px solid mischka
   border-radius 3px
@@ -38,6 +38,8 @@
 .user-legend-wrap
   width 90%
   background white
+  height 110px
+  overflow scroll
 
 .user-legend
   margin 10px
@@ -99,7 +101,7 @@
   color bayoux
 
 .article-scrollbox
-  max-height calc(95vh \- 200px)
+  max-height calc(95vh \- 225px)
   overflow-y scroll
   overflow-x hidden
 


ragesoss:
I don't think this solution is discoverable enough. Two-row lists currently work okay... not ideal, but usable, and this will be a step back in usability for that case.

appleJax:
Ok, I increased the size of the footer, and added `overflow: scroll` to the author list. When the author list is shorter, the height of the footer stays the same. That's why I chose to cut off the third row. It reduces the amount of empty whitespace when the author list is short, and it lets the reader know there are more authors when the list is long.

The way the css is currently structured, the height of the footer is determined by a combination of the heights of `.article-viewer` and `.article-scrollbox`. The content of the footer has no bearing on its own height. In order to make the footer's height fluctuate according to its content, the css would need more significant changes. I wanted to offer this solution, which only changes 6 lines, before messing too much with how the css is structured.

screen shot 2017-11-10 at 7 55 54 pm

ragesoss:
This one significantly worsens the styling of the most common case of just one line. I wonder if we can fix the bottom of that div to the bottom of the article viewer, and let it take up more space upward into the article area when it needs more lines.

appleJax:
What's the easiest way to seed all the necessary information into the development environment to replicate this issue?

ragesoss:
Create a course with the same start and end dates as that one from production, and then go to the 'Students' tab and add in each of the usernames that edited that 'Cell membrane' article. Then, run `rake batch:update_constantly`.

ragesoss:
This issue is fixed with #1515

PR #1491, opened 2017-11-12
enable no danger react linting

tansaku:
related to #1408

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -25,7 +25,7 @@
     # Rules below here should be enabled once violations are fixed.
     "react/no-find-dom-node": 2,
     "react/forbid-prop-types": 0,
-    "react/no-danger": 0,
+    "react/no-danger": 2,
     "react/no-children-prop": 0,
     "react/no-unused-prop-types": 0,
     "react/no-unescaped-entities": 0,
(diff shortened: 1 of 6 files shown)


bryszard:
I've looked into the issue a bit more and I think our solution is too naive - we need to have a way to parse the input into html (something like `html_safe` method), but without using `dangerouslySetInnerHTML`.

One possible solution would be to use some different parser of html (see also - https://stackoverflow.com/questions/23616226/insert-html-with-react-variable-statements-jsx).
I think we could try with this library to do this - https://www.npmjs.com/package/html-react-parser

ragesoss:
I'm not sure there's a way around it for some of our use cases, in particular the training slides and the article viewer. We take in html from an outside source (Wikipedia) and render it, so it's going to be 'dangerous' in the sense of injecting whatever html comes from Wikipedia, but that's what we need.

bryszard:
@tansaku I thought it'll be easier if I create a new PR, directly from the branch in my forked repo - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1492.

If you think that it's ok to deal with it this way, you can close this PR :).

PR #1492, opened 2017-11-12
Issue 1408 - enable no-danger rule for react linting

bryszard:
Refers to issue #1408.

I'm using the html parser library - https://www.npmjs.com/package/html-react-parser, as suggested here - https://stackoverflow.com/questions/23616226/insert-html-with-react-variable-statements-jsx.


tansaku:
@bryszard thanks for the extra work on this - so is this done now?  Is there some explicit block we're up against?

tansaku:
I've restarted the travis jobs - they seemed to be failing on weird stuff ...

ragesoss:
There's one feature spec failing with a Javascript promise error:
`Unhandled promise rejection TypeError: First argument must be a string`

That suggests that this change introduced a new JS error, probably in one of the two training components that changed.

Try using the dev build and go through the 'editing basics' training module while monitoring the browser console. The failing spec basically just goes through that module from start to finish.

bryszard:
Ok, I've adjusted the PR. Let's see if Travis passes now.

ragesoss:
Interesting! Can you explain in the PR message what Parser is doing and why it's an improvement over using dangerouslySetInnerHTML?

bryszard:
@ragesoss The main advantage is that the lint is not moaning :).

But what is the difference besides of it? It does more or less the same job, but it's cleaner in my opinion cause you don't need to set it as an element attribute, but directly in the content.

If you think it makes sense to do something else (eg. sanitize the html) we could add DOMPurify library).

ragesoss:
@bryszard Thanks! The other option would be to just keep the rule disabled, and move it higher up in the .eslintrc file to the section of rules that we plan to keep disabled. :-)

I'm just not sure whether there would be any circumstances where we are using the "dangerous" method that would be safer with this. I'm guessing the performance hit from adding a parser isn't big enough to worry about, but after looking over the violations and the fix for them, I think I prefer just using the built-in dangerous method over adding a pre-1.0 dependency that doesn't change the functionality at all.

bryszard:
Sounds reasonable @ragesoss. I'll just move this rule up and add a comment about this then.

PR #1501, opened 2017-11-20
Remove 'up-to-date training' mouseover unless there are trainings

mattwarren10:
Hopefully this helps. Let me know of any changes you prefer.


ragesoss:
@mattwarren10 this looks like it implements the switching behavior to hid the mouseover, but it does it just based on student count. The case we're trying to solve is for is like this: https://outreachdashboard.wmflabs.org/courses/Jacob_Lawrence_Gallery/Seattle_Jacob_Lawrence_Gallery_AF_2017

There are participants, but they are not assigned to complete any training modules.

I think the easiest way to fix this 95% of the way would be to switch on course.timeline_enabled. If a course does not have the timeline enabled, then there won't be any assigned trainings, and courses with timelines but no trainings will be very rare edge cases.

mattwarren10:
I changed the condition to check if `course.timeline_enabled` is true.

ragesoss:
Excellent.

PR #1504, opened 2017-11-20
Issue 1408 - eslint rule: no-children-prop 

bryszard:
Refers to #1408 

Seems that this rule was not violated anywhere.


.eslintrc
```diff
@@ -26,7 +26,7 @@
     "react/no-find-dom-node": 2,
     "react/forbid-prop-types": 0,
     "react/no-danger": 0,
-    "react/no-children-prop": 0,
+    "react/no-children-prop": 2,
```
ragesoss: Instead of explicitly enabling this, we should just remove the line that disables it.

The baseline rules come from the eslint modules in package.json, and new rules get added or removed when we update those packages.

ragesoss: You could also remove the no-find-dom-node line; I should have noted the same thing when that change got merged, but I didn't think of it.

bryszard: Ok, I've done it here and on other branches :).

PR #1505, opened 2017-11-20
Issue 1408 - eslint rule: no-unused-prop-types

bryszard:
Refers to #1408 

Rule was violated in 3 places. See more explanation here - https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md.


ragesoss:
Looks good to me. Please remove the .eslintrc line altogether rather than explicitly enabling the rule.

PR #1506, opened 2017-11-20
Issue 1408 - eslint rule: no-unescaped-entities

bryszard:
Refers to #1408 

More info - https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unescaped-entities.md


ragesoss:
Cool! Same thing as the others: please remove the eslintrc line instead of enabling explicitly.

Issue #1499, opened 2017-11-18
On P&E Dashboard, layout of training index is bad


ragesoss:
Almost all the libraries are on the left.

![training index](https://user-images.githubusercontent.com/848483/32985484-04313c1c-cc71-11e7-869b-1b3b18978a2e.png)

mattwarren10:
I went to the training page on the dashboard, but couldn't find this issue. What is the route where the issue exists?

ragesoss:
The route is the training index: `/training`: https://outreachdashboard.wmflabs.org/training

This only happens on P&E Dashboard, which suggests that it's related to how on-wiki trainings get stored.

You can configure your dev environment to pull in the same trainings using these settings in application.yml

```
  training_path: 'training_content/none'
  disable_training: 'false'
  enable_wiki_trainings: 'true'
  training_slides_wiki_page: 'Training modules/dashboard/slides'
  training_modules_wiki_page: 'Training modules/dashboard/modules'
  training_libraries_wiki_page: 'Training modules/dashboard/libraries'
```
After that, you may need to visit `/reload_trainings?module=all` to load the training content from meta.wikimedia.org.

mattwarren10:
I changed the `application.yml` file to fit these settings, and after I reloaded the trainings, I couldn't reproduce the issue. Maybe there's another setting to change. I'll look for one.

ragesoss:
Interesting. Screenshot of what it looks like for you?

-Sage

mattwarren10:
![training-library](https://user-images.githubusercontent.com/13757305/33090706-ee32e01a-cec2-11e7-83d3-36827b41d0d5.png)

ragesoss:
It looks like you haven't reloaded the trainings, so you're still seeing the yml-based wiki ed versions.

Visit `/reload_trainings?module=all` and check back. It may take a while.

mattwarren10:
I visited that route and did it several times. I'll just wait if it takes time.

ragesoss:
If you saved those changes in application.yml (and there aren't the same config lines with different values elsewhere), and you've restarted rails since that change, then the reload should take a while. It's almost immediate if enable_wiki_trainings is false.

mattwarren10:
Got the issue working. Forgot to reload the server. I'll see what I can do.

PR #1508, opened 2017-11-21
Modify training list items to be organized #1499

mattwarren10:
Looks like this now:

clear-left


app/assets/stylesheets/training_modules/_training.styl
```diff
@@ -128,6 +128,13 @@
   col(12/12)
 .training-libraries__individual-library
   col(6/12)
+  width 48%
+  margin 0% 1%
+  display inline-block
+
+.training-libraries__individual-library:nth-child(odd)
```
ragesoss: nice.

PR #1515, opened 2017-11-29
bugfix: 1487 AritcleView no longer cuts off author list

AlexYanai:
In `ArticleView`, when a list of authors grows beyond 2 rows, the footer would formerly extend out of the `div`. 

This change adds an `overflow: scroll` to `.article-footer` and changes the height so that overflow only occurs when a 3rd row is added. Cases with 1 or 2 rows are aligned properly.

#1487

### DIFF
--- a/app/assets/stylesheets/modules/_article_viewer.styl
+++ b/app/assets/stylesheets/modules/_article_viewer.styl
@@ -20,9 +20,10 @@
 .article-footer
   border-top 1px solid mischka !important
   font-size 90%
-  overflow: scroll
-  max-height: 92px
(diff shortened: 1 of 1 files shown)


ragesoss:
This is just about the right fix. To avoid scrollbars when they aren't necessary, instead of `overflow scroll` it should use:

```
  overflow-y auto
  overflow-x hidden
```

Also, no need for the colons with .styl syntax.

AlexYanai:
Sounds good, I just made the updates. Thanks!

ragesoss:
Nice work, thanks!

Issue #1518, opened 2017-11-30
Convert react classes to stateless functional components


ragesoss:
These components don't use any React class features, and so could be converted into pure functions:

- [x] /app/assets/javascripts/components/app.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1519
- [x] /app/assets/javascripts/components/activity/recent_activity_handler.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1520
- [x] /app/assets/javascripts/components/articles/article_list.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1521
- [x] /app/assets/javascripts/components/articles/available_article_list.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1521
- [x] /app/assets/javascripts/components/articles/salesforce_media_buttons.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1522
- [x] /app/assets/javascripts/components/common/course_link.jsx -https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1523
- [x] /app/assets/javascripts/components/common/list.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1523
- [x] /app/assets/javascripts/components/common/loading.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1523
- [x] /app/assets/javascripts/components/common/popover.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1523
- [ ] /app/assets/javascripts/components/high_order/conditional.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1524
- [x] /app/assets/javascripts/components/onboarding/index.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1525
- [x] /app/assets/javascripts/components/overview/inline_users.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1525
- [x] /app/assets/javascripts/components/revisions/revision_list.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1525
- [x] /app/assets/javascripts/components/timeline/orderable_block.jsx - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1525
[...]

PR #1524, opened 2017-11-30
Convert Conditional component to a stateless functional component

arun1595:
Converts `Conditional` high order component to a stateless functional component.


ragesoss:
Looks like this one uses refs, so it can't become a stateless functional component.

ragesoss:
@arun1595 how about, instead of this:

1. Remove the line in .eslintrc that disables the rule for preferring stateless functional components
2. Add markup to this file to disable the rule for only this component.

With that, we could close #1518

ragesoss:
@arun1595 there is a linting rule that tries to detect when a component could be an SFC, and it if it is enabled it would break the build.

From the error in this build for this PR, it looks like this Conditional component can't actually become an SFC. So one option would be to enable the rule (by removing a line from .eslintrc) and then disabling it inline for just conditional.jsx.

Alternatively, maybe you can figure out why Conditional is throwing errors as an SFC and fix that.

ragesoss:
Not sure this one is possible, but if it is, it should probably be redone from a fresh branch.

PR #1532, opened 2017-12-05
Forbid prop types

jstoebel:
this partially addresses the issue of the `forbid-prop-types` rule. Sadly I don't have the time at the moment to continue with this. I _hope_ to have some time later this month, but I would hate for someone else to work on this one only to find out their efforts were duplicated. Feel free to either accept the partial fix and merge it in, or leave it alone and I will pick it up later.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -20,7 +20,7 @@
     "no-alert": 0,
     "new-cap": [2, { "capIsNew": false }],
     "no-use-before-define": 0,
-    "react/forbid-prop-types": 2,
+    "react/forbid-prop-types": 0,
     # Rules below here should be enabled once violations are fixed.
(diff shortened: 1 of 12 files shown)


ragesoss:
Thanks @jstoebel. This is valuable work; it's great to see what it would look like to add `shape` specifications to some of the key props.

I think we're not ready to use that level of detail, though. Eventually, I'd like to see that shape spec somewhere near the top of the component hierarchy. Or maybe we could specify the shape of `course` in its own file and just import that any time we want it in the props for another file.

The volume of boilerplate would be pretty significant if we used the full shape spec everywhere.

ragesoss:
Closing this, per my previous comment.

PR #1535, opened 2017-12-06
#1005 Adds Average Delay to Dashboard statistics

mauditecandela:
- Modifies old last_update method and instead of requesting the 'last_update' key, looks for the last update that took place
- Deletes old log_updates method and instead creates a new method that updates the last 10 updates
- Adds the expected delay time to the Overview Handler
- Adds translations to ES, CA and EN
- Includes a test for the new methods for the update_log.rb


ragesoss:
@mauditecandela You should rebase this onto the latest master branch, or redo the update logging changes by themselves. The linting problems are fixed separately already, and it might be complicated to merge this branch.

app/models/update_log.rb
```diff
@@ -4,14 +4,31 @@ def self.setting_record
...
+
+  end
+
+  def self.log_updates(time)
     setting = self.setting_record
-    setting.value["last_update"] = time
+    last_update = setting.value.keys.last
+    if last_update === nil
+      new_update = 0
+    elsif last_update >= 10
```
ragesoss: This number is a "magic number", which should be replaced by a constant, so that the name can help show what it's for — in this case, the number of updates to keep.

app/models/update_log.rb
```diff
@@ -4,14 +4,31 @@ def self.setting_record
...
+      setting.value.delete(last_update - 10)
+      new_update = last_update + 1
+    else 
+      new_update = last_update + 1
+    end
+    setting.value[new_update] = time
```
ragesoss: I've I'm reading this right, this data structure is something like
```
{
  1 => time of update 1,
  2 => time of update 2,
 ...
}

And the numbers will just keep increasing, with the older values being removed?

I think that will work fine. It'll be more flexible and resilient to future changes if you store data about the update as another hash, instead of just the time value. For example, the value for each update number key could be something like `{ start_time: time, end_time: time }`. That way, as we add more things to the update logs (eg, keeping track of how many programs were active at the time of the update), we'll still be able to access the data the same way.

File: app/models/update_log.rb [hunk 3]
```diff
@@ -4,14 +4,31 @@ def self.setting_record
...
     setting.save
   end
 
-  def self.last_update
-    self.setting_record
-    Setting.find_by(key: 'metrics_update').value['last_update']
+  def self.average_delay
```
ragesoss: This gets called for every page load, so we should calculate it each time the record is updated and save that value on the record itself rather than crunching the numbers again and again.

That will probably require adding another level of nesting to the data structure. So setting.value could be something like:

```
{
  constant_updates: {
    1 => { ... },
    2 => { ... }
   },
  average_constant_update_time: time
}
```

app/models/update_log.rb
```diff
@@ -4,14 +4,42 @@ def self.setting_record
...
+      setting.value['average_delay'] = seconds.inject(:+)/(seconds.length-1) if seconds.length > 1
+    end
     setting.save
   end
 
+  def self.log_updates(starttime, endtime)
+    setting = self.setting_record
+    update_record = setting.value['constant_update'] ? setting.value['constant_update'] : setting.value['constant_update'] = Hash.new
+    last_update = update_record.keys.last
+    max_updates = 10
```
ragesoss: This is a constant, so instead of setting a variable within the method, you can define it within the class, outside the method, like `MAX_UPDATES_TO_KEEP = 10`.

app/models/update_log.rb
```diff
@@ -4,14 +4,42 @@ def self.setting_record
...
+    if constant_update 
+      seconds = constant_update.keys.length > 1 ? constant_update.map {|key, value| (Time.parse(constant_update[key]["end_time"].to_s) - Time.parse(constant_update[key-1]["end_time"].to_s) if key != constant_update.keys[0]).to_i} : [0]
+      setting.value['average_delay'] = seconds.inject(:+)/(seconds.length-1) if seconds.length > 1
+    end
     setting.save
   end
 
+  def self.log_updates(starttime, endtime)
+    setting = self.setting_record
+    update_record = setting.value['constant_update'] ? setting.value['constant_update'] : setting.value['constant_update'] = Hash.new
```
ragesoss: Looks like this ternary conditional isn't doing anything.

app/models/update_log.rb
```diff
@@ -4,14 +4,42 @@ def self.setting_record
...
+    max_updates = 10
+    if last_update === nil
+      new_update = 0
+      update_record[new_update] = Hash.new
+    elsif last_update >= max_updates
+      update_record.delete(last_update - max_updates)
+      new_update = last_update + 1
+    else 
+      new_update = last_update + 1
+    end
```
ragesoss: See if you can find a way to write this in a way that makes the meaning and intention easier to decipher.

One thing to consider is the variable names: `new_update` and `last_update`. These are variables are not "update" objects that have all the information about an update, they are integers that represent the index into the hash of update data.

app/models/update_log.rb
```diff
@@ -4,14 +4,42 @@ def self.setting_record
...
-    setting.value["last_update"] = time
+    constant_update =  setting.value['constant_update']
+    if constant_update 
+      seconds = constant_update.keys.length > 1 ? constant_update.map {|key, value| (Time.parse(constant_update[key]["end_time"].to_s) - Time.parse(constant_update[key-1]["end_time"].to_s) if key != constant_update.keys[0]).to_i} : [0]
+      setting.value['average_delay'] = seconds.inject(:+)/(seconds.length-1) if seconds.length > 1
+    end
     setting.save
   end
 
+  def self.log_updates(starttime, endtime)
```
ragesoss: This class will benefit from separating out the high-level concepts of what it is doing, and the low-level details of how it is doing them... along with separating out the methods intended to be public from the methods that are helpers and should be considered private because we only want to call them from other methods within this class.

At the high level, there are basically two entry points: the one for fetching data to display on the frontend, and the one for logging a new update. Those should be at the top of the file.

One strategy I like is to write those public methods completely in terms of private helper methods that tell a clear story of what steps are involved in the process. So for `log_update`, it might be something like:
```ruby
def self.log_update(update_data)
  add_new_log update_data
  delete_oldest_log
  update_average_delay
  save_setting_record
end
```

mauditecandela: Nice feedback. Would you create that many methods to do so? Or is just the example?

ragesoss: Yes, I would create separate methods for each of those steps.

app/models/update_log.rb
```diff
@@ -1,17 +1,62 @@
...
   end
 
   def self.last_update
-    self.setting_record
-    Setting.find_by(key: 'metrics_update').value['last_update']
+    setting = self.setting_record
+    setting.value['constant_update'] ? setting.value['constant_update'].values.last['end_time'] : nil
   end
+
+  private
```
ragesoss: This won't actually make the class methods private. It's important that the *actually* be private methods, though... the important part is documenting which methods are intended to be called from other classes, and which are not.

To get private class methods, you must either set each one private using `private_class_method`, or use `class << self` to open up the singleton class.

app/models/update_log.rb
```diff
@@ -1,17 +1,62 @@
...
   def self.last_update
-    self.setting_record
-    Setting.find_by(key: 'metrics_update').value['last_update']
+    setting = self.setting_record
+    setting.value['constant_update'] ? setting.value['constant_update'].values.last['end_time'] : nil
   end
+
+  private 
+
+    def self.add_new_log
```
ragesoss: This looks like it just initializes the hash if it's not already there. That doesn't match up with the method name. My suggestion is to have `add_new_log` insert the start/end time from the new update (and first initialize the hash, if necessary), then have a separate method to remove oldest one if there are more than 10 updates already stored.

app/models/update_log.rb
```diff
@@ -1,17 +1,63 @@
...
-    self.setting_record
-    Setting.find_by(key: 'metrics_update').value['last_update']
+    setting_record
+    @setting.value['constant_update'] ? @setting.value['constant_update'].values.last['end_time'] : nil
   end
+
+  class << self
+
+    def setting_record
+      @setting = Setting.find_or_create_by(key: 'metrics_update')
```
ragesoss: Instead, you can do `@setting ||= Setting.find_or_create_by(key: 'metrics_update')`.

Then, you don't having to call `setting_record` first and use `@setting` elsewhere... you can just do `setting[key] = value` or `setting.save` or whatever.

app/models/update_log.rb
```diff
@@ -1,17 +1,63 @@
...
+      if @last_update === nil
+        index = 0 
+      else
+        index = @last_update + 1
+      end
+      @setting.value['constant_update'][index] = {'start_time' => starttime, 'end_time' => endtime}
+    end
+
+    def delete_old_log
+      if @last_update != nil && @last_update >= MAX_UPDATES_TO_KEEP
```
ragesoss: The most idiomatic Ruby way to do this kind of thing is with a guard clause instead of an if statement:

```ruby
return if some_conditions
do_the_thing
```

app/models/update_log.rb
```diff
@@ -1,17 +1,63 @@
...
+        @setting.value['constant_update'].delete(@last_update - MAX_UPDATES_TO_KEEP)
+      end
+    end
+
+    def log_delay
+      constant_update = @setting.value['constant_update']
+      if constant_update
```
ragesoss: same here: 'return unless constant_update`

mauditecandela: Thanks a lot for this guidance :)

app/models/update_log.rb
```diff
@@ -1,17 +1,57 @@
...
+      if @last_update === nil
+        index = 0
+      else
+        index = @last_update + 1
+      end
+      setting_record.value['constant_update'][index] = {'start_time' => starttime, 'end_time' => endtime}
+    end
+
+    def delete_old_log
+      return setting_record.value['constant_update'].delete(@last_update - MAX_UPDATES_TO_KEEP) if @last_update != nil && @last_update >= MAX_UPDATES_TO_KEEP
```
ragesoss:
```ruby
return unless @last_update?. >= MAX_UPDATES_TO_KEEP
```

ragesoss:
Let's start from other branch rather than using this one.

PR #1541, opened 2017-12-07
Linter update dangling commas

AdamMescher:
This PR references Issue #[1408](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1408).

Removing the `“comma-dangle”: 0` from `.eslintrc` resulted in changes to 248 files to resolve over 1000 linter errors. We saw the number of errors increase in only a couple days and realized that the problem would only continue to grow if it wasn’t addressed immediately. 

We felt this contribution would be beneficial to the project as a whole in mitigating the problem before it became insurmountable. 

All tests are passing after committing changes. 

Contributions by: @bbp5280, @nicktu12, @adammescher

### DIFF
--- a/app/assets/javascripts/actions/api_fail_action.js
+++ b/app/assets/javascripts/actions/api_fail_action.js
@@ -4,7 +4,7 @@ const Flux = new McFly();
 const ApiFailAction = Flux.createActions({
   fail(response) {
     return ({ actionType: 'API_FAIL', data: response });
-  }
+  },
 });
 
 export default ApiFailAction;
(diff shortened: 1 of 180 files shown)


ragesoss:
@bbp5280, @nicktu12, @AdamMescher Thanks. It looks like you made these changes manually, with one commit per file change? Dangling commas appear to be fixable automatically by eslint, with `eslint --fix`.

This rule is also one of the ones above the "rules below should be enabled" comment in `.eslintrc`; this is a rule that we are planning to leave in place. It's only the ones below that comment that we'd like to enable.

In this case, it's not a huge deal either way, but leaving off the dangling comments makes it more consistent with Ruby style, where dangling commas are never used.

Issue #1517, opened 2017-11-29
For live edit-a-thons, add an 'Update stats' button to AvailableActions


ragesoss:
During edit-a-thons, people want ongoing info about the contributions. This can be done via the hidden `/manual_update` path for a program, but for edit-a-thons that are currently happening, we should make this discoverable.

For editathons (only), between the start and end times, add a button that links to the manual_update for that program.

You should create a stateless functional component, UpdateStatsButton or the like, which takes the course slug as its prop, and add it from AvailableActions under the appropriate conditions.

mauditecandela:
One nice reference to consider is the join button from available_actions.jsx

PR #1544, opened 2017-12-12
#1517 - Adds Manual Update Button for Edit-a-thons

mauditecandela:
- Adds a button to generate a manual update of the statistics only available to Admins and Instructors in case the Course is published and inside the P&E Program 
- After the button is clicked a modal is generated asking for confirmation 
- After confirmation a GET request is sent to /courses/ID/manual_update 
- The user is also sent to this url. Note: Is this the desired behaviour? 
- Adds locales and translations. Key: update_stats

### DIFF
--- a/app/assets/javascripts/components/overview/available_actions.jsx
+++ b/app/assets/javascripts/components/overview/available_actions.jsx
@@ -44,6 +44,13 @@ const AvailableActions = createReactClass({
     this.props.initiateConfirm(confirmMessage, onConfirm, true, joinDescription);
   },
 
+  updateStats() {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -114,6 +122,13 @@ const AvailableActions = createReactClass({
         <p key="join"><button onClick={this.join} className="button">{CourseUtils.i18n('join_course', this.state.course.string_prefix)}</button></p>
       ));
     }
+    // If the user is an instructor or admin, and the course is published, show a manual stats update button
```
ragesoss: update comment

PR #1548, opened 2017-12-13
Average delay

ragesoss:
@mauditecandela I did some basic refactoring (in separate commits so you could take a look at each refactor).

Remaining suggestions will be added as comments.

### DIFF
--- a/app/models/update_log.rb
+++ b/app/models/update_log.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class UpdateLog
   ####################
   # CONSTANTS        #
@@ -13,7 +15,7 @@ class UpdateLog
   end
 
   def self.average_delay
-    setting_record.value['average_delay'] ? setting_record.value['average_delay']: nil
+    setting_record.value['average_delay'] ? setting_record.value['average_delay'] : nil
   end
 
   def self.last_update
@@ -21,7 +23,6 @@ class UpdateLog
   end
 
   class << self
-
     def setting_record
       @setting ||= Setting.find_or_create_by(key: 'metrics_update')
     end
@@ -47,8 +48,8 @@ class UpdateLog
     def log_delay
       constant_update = @setting.value['constant_update']
       return unless constant_update && constant_update.keys.length > 1
-      times = constant_update.map {|key, value| Time.parse(value["end_time"].to_s)}
-      seconds = (times[times.length-1] - times[0])/(times.length-1)
+      times = constant_update.map { |key, value| Time.parse(value['end_time'].to_s) }
+      seconds = (times[times.length - 1] - times[0]) / (times.length - 1)
       @setting.value['average_delay'] = seconds
     end
 
@@ -56,5 +57,4 @@ class UpdateLog
       @setting.save
     end
   end
-
 end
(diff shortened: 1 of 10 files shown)


app/models/update_log.rb
```diff
@@ -1,17 +1,60 @@
...
 class UpdateLog
+  ####################
+  # CONSTANTS        #
+  ####################
+  # We want to keep a maximum number of 10 updates, so the last key we want to save is the ninth
+  MAX_UPDATES_KEY = 9
 
-  def self.setting_record
-    Setting.find_or_create_by(key: 'metrics_update')
+  def self.log_updates(starttime, endtime)
```
ragesoss: I think this interface is too brittle. If there is more we want to save, we'll continually need to change the expected arguments along with the logic for saving it.

I suggest instead that we take in all the data we want to save as a hash, and then just save it like:

```ruby
setting_record.value['constant_update'][this_update] = input_hash
```

That hash will have keys for the current things we require (start_time, end_time) but it will then be easy to save more data without breaking anything.

ragesoss:
I've been testing this out locally, and there are a few problems I noticed.

I didn't consider thread safety when I suggested using a class instance variable for the record. That means different threads — like the webserver vs. the update thread — can have different data, and the webserver won't know that new data got saved by the update thread... so even after an update, the latest data won't show up on program pages. The whole update part of it should probably be converted to instance methods, with a call to `new` when we log a new update.

Also, average delay isn't doing the calculations as expected. I get a value of class `Rational`, like `(139368885407/194400000000000)`, and the displayed value is something very high, like `19:16:47`, even though the updates I ran to populate it were very short.

mauditecandela:
@ragesoss I created a new pull request to fix the first problem. 
Although when testing locally, I don't get the same error when it comes to average_delay, I get in fact a float that it's correctly translated in the frontend. Where could this difference come from?

ragesoss:
@mauditecandela the layout is bad when the MyArticles component is present (ie, for users participating as students).

Also, there's a stray period at the beginning of the message.

enrolled user update data

app/models/update_log.rb
```diff
@@ -1,17 +1,51 @@
...
+      return unless @last_update&. >= MAX_UPDATES_KEY
+      setting_record.value['constant_update'].delete(@last_update - MAX_UPDATES_KEY)
+    end
+
+    def log_delay
+      constant_update = @setting.value['constant_update']
+      return unless constant_update&.length&. > 1
+      times = constant_update.values.map { |log| log['end_time'] }
+      seconds = (times[times.length - 1] - times[0]) / (times.length - 1)
+      @setting.value['average_delay'] = seconds.to_i
```
ragesoss: The subtraction of times is where I get the weird results locally; using `to_f` on datetimes will convert it into seconds consistently.

We can also use some of Array's methods to make this read more clearly:

```ruby
      average_time = (times.last.to_f - times.first.to_f) / (times.count - 1)
      @setting.value['average_delay'] = average_time
```

ragesoss:
@mauditecandela once the layout issue is fixed, this is good to go.

ragesoss:
@mauditecandela A few more things to polish this up...

I think if the 'next update' estimate is in the past, then we should not show it. It will be very confusing when the estimate is off or updates have stalled.

last update next update

Per the discussion with Jonathan, the 'next update' message should make it clear that it's an estimate. Maybe just change the message to 'Next update estimated"

Also, we don't need the mouseover info on the title now, I think.

PR #1550, opened 2017-12-14
Deletes one argument to the methods of UpdateLog

mauditecandela:
- In order to be more stable, deletes one of the arguments that we were passing to log_updates and add_new_log
- These arguments is now a hash that should have the key values "start_update" and "end_update" included
- Changes also batch update logging to include the hash with the values needed.

### DIFF
--- a/app/models/update_log.rb
+++ b/app/models/update_log.rb
@@ -7,8 +7,8 @@ class UpdateLog
   # We want to keep a maximum number of 10 updates, so the last key we want to save is the ninth
   MAX_UPDATES_KEY = 9
 
-  def self.log_updates(starttime, endtime)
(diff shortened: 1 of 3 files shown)


app/models/update_log.rb
```diff
@@ -7,8 +7,8 @@ class UpdateLog
   # We want to keep a maximum number of 10 updates, so the last key we want to save is the ninth
   MAX_UPDATES_KEY = 9
 
-  def self.log_updates(starttime, endtime)
-    add_new_log(starttime, endtime)
+  def self.log_updates(input_hash)
```
ragesoss: `input_hash` is not a very useful name.

PR #1557, opened 2017-12-16
added fixes to common problems in troubleshooting.md  

aSquare14:
- When I followed setup.md, I noticed that for running mysql the command "sudo mysql" did not work for me as the default password was blank in database.yml,so I added a workaround.

- In a debian system to confirm that the redis-server runs as a daemon I added the command 
$ `ps aux | grep redis-server`  . (Just so that one can crossscheck).

- I faced a few issues while setting up gulp. One of them was that version of gulp in CLI and Local mismatched.This is a common issue and I added a warning so that this can be avoided. $ `gulp -v`


ragesoss:
These are all useful tips. I think we should put them in troubleshooting.md instead of here. Having too much detail — especially detail that doesn't apply to most people — is itself a source of confusion and problems during setup.

aSquare14:
Yes sure. I'll update troubleshooting.md .

ragesoss:
Looks like there are some accidental whitespace changes to README. Clean those up and this will be ready to go.

aSquare14:
There were no changes in the README made by me. I also did not find extra whitespaces in setup.md or troubleshooting.md

docs/setup.md
```diff
@@ -107,10 +107,10 @@ If you know your way around Rails, here's the very short version. Some additiona
       $ `zeus start`
 
 2. **Start Redis** (if not already running as daemon)
-    - Redis is used by Sidekiq. Some features — especially related to making
-    edits on Wikipedia — will not work when Redis is down. On a Linux-based system,
```
ragesoss: Here is the whitespace change.

Issue #1380, opened 2017-09-07
For courses with no passcode, 'Join' button in Available Actions should not request a passcode


ragesoss:
This means we'll need to let the frontend know that the password is blank, in the cases where the password has been removed.

ihsavru:
Hi! I am a newcomer and would like to work on this issue.

ragesoss:
Hi @ihsavru! Great. This is a relatively straightfoward change to describe, but it will probably touch both the server and client code in a few places, so don't hesitate to ask if you have any questions or want any clarification about how the system works.

ihsavru:
@ragesoss I am having trouble setting it up locally on my machine. Is this the right place to ask :sweat_smile: ?

ragesoss:
@ihsavru You can post specific problems you're having with the setup process here: https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1351

I also sent you a slack invite where you can post any questions you have; a few folks there can probably help out.

alpha721:
@ragesoss I figured that the some changes have to be made in `self_enrollment_controller.rb` but I am not able to find the code which requests the passcode on the front end. Can you give a slight hint as to where I should be looking.

ragesoss:
The frontend does not currently know anything about the password (unless the user is an admin or instructor). I think the way to handle it will be to add an attribute to send through course.json in the case where there is no password: `app/views/courses/course.json.jbuilder`. This is the main course object that gets passed through most of the frontend.

Set passcode to empty string if the course password is not present, and then use that on the frontend in the join course action flow.

alpha721:
` # If course has no passcode set, treat any submission as valid.
    return true if @course.passcode.blank?`
should this still hold?

ragesoss:
I think so. Where is that from?

alpha721:
` def passcode_valid?
    # If course has no passcode set, treat any submission as valid.
    return true if @course.passcode.blank?
    params[:passcode] == @course.passcode
  end`
its from `self_enrollment_controller.rb`.

ragesoss:
Yeah, that should be fine still. Enrolling without a passcode works, it's just the 'Join course' action modal that doesn't handle it elegantly because it doesn't know that the course requires no password.

alpha721:
![wiki](https://user-images.githubusercontent.com/18638840/30615999-5ef5fdaa-9dae-11e7-8a4e-2b01ca5adf93.png)

[...]

PR #1559, opened 2017-12-18
Allow to join a course without a passcode #1380

mauditecandela:
For those cases where we want to allow the users to join a program without having a passcode, we will display a different kind of button that doesn't require to fill out a passcode input.
This PR: 
- Adds a new method to the Course model that checks if the course requires a passcode and if this passcode is blank. Will return true only if the passcode IS blank and IS NOT required
- Adds a new functionality to available_actions displaying a different kind of button if the course allows registration without a passcode
- Includes a new method in the SelfEnrollment Controller that will allow the user indeed to register in this course they don't provide the passcode. 
- Adds an English translation for the Modal Prompt.

### DIFF
--- a/spec/features/student_role_spec.rb
+++ b/spec/features/student_role_spec.rb
@@ -135,7 +135,7 @@ describe 'Student users', type: :feature, js: true do
 
       expect(page).to have_content 'An Example Editathon'
 
-      click_button 'Join course'
+      click_button 'Join program'
       within('.confirm-modal') do
         click_button 'OK'
       end
--- a/spec/features/student_role_spec.rb
+++ b/spec/features/student_role_spec.rb
@@ -36,18 +36,6 @@ describe 'Student users', type: :feature, js: true do
            end: '2020-01-01'.to_date)
   end
 
-  let!(:second_editathon) do
-    create(:editathon,
-           title: 'Second Example Editathon',
-           school: 'University',
-           term: 'Term',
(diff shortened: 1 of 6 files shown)


ragesoss:
Did you get the i18n working?

This should have tests too. There are related tests *with* a passcode in `spec/features/student_role_spec.rb`.

mauditecandela:
Yes, the i18n thing is working. 
I will work on the tests now.

mauditecandela:
@ragesoss do you think writing a test in the same place (student_role_spec) is a good idea? I think it makes sense.

ragesoss:
Yep!

ragesoss:
@mauditecandela if you have the dev build of the assets, then that's normal; the error-level warnings for deprecated react features cause the capybara tests to fail. Do `gulp build` and those errors go away.

If you are using the production build (`gulp build`) and you have errors still, we should look for the cause.

spec/features/student_role_spec.rb
```diff
@@ -24,6 +24,17 @@
...
+    create(:editathon,
+           title: 'An Example Editathon',
+           school: 'University',
+           term: 'Term',
+           slug: 'University/An_Example_Editathon_(Term)',
+           submitted: true,
+           passcode: '',
+           start: '2015-01-01'.to_date,
+           end: '2020-01-01'.to_date)
+  end
```
ragesoss: Need to add this to a campaign as well.

PR #1565, opened 2017-12-21
Default_course_type based on the campaign

mauditecandela:


ragesoss:
Looks like there are two failing campaign specs... possibly again related to validation of the new field? Probably upgrading the course type field to a `select` with only the valid options will fix it.

mauditecandela:
@ragesoss that's right but not only that, I think. The "nil" as option in the default course type is only working when creating the course, but not when updating. So we are forcing the user to add a Default Course Type to every campaign if the user tries to edit it.

We have a weird user experience regarding the default course type. We are only showing it to the user once they want to edit the course and only if the course has start/end dates. I will check more throughly but I would: 
- Fix the validation so a nil default course type also passes
- Add a Default Course Type Option to the Campaign creator
- Show the Default Course Type independently of the start-end dates
- Add a dropdown to the edit function

Issue #1459, opened 2017-10-16
On-wiki course page shows the instructor with the first User record, rather than the first CoursesUsers record


ragesoss:
On-wiki course pages (from `WikiCourseOutput`) list only the first instructor. (See also #1175) This 'first' instructor is not necessarily the first one enrolled in the course, but it should be.

Also, the real name is fetched from the User record. It should instead by fetched from the CoursesUsers record.

aSquare14:
I'd like to work on this issue.

ragesoss:
@aSquare14 great! would like me to set up a GCI task for it?

aSquare14:
By GCI do you mean Google Code In?  Because if you do, I'm not eligible for
GCI, I'm a sophomore at college(not in the age gap 13-17) . I'm just
getting started with Open Source and have a keen interest in learning and
exploring. :D

mariahajmal:
@ragesoss Hi, Is this fixed? If not, can you please assign it to me

ragesoss:
Yeah, it looks like this did get fixed incidentally via https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/1d0268c84e0d579831e951e1d5a31a3e6a011552

PR #1566, opened 2017-12-21
#fixes 1459 ,wiki-course page shows first instructor and real name fetched from CourseUsers

aSquare14:
On Wiki course pages, only first instructor is listed.He is the first enrolled.
The real name is fetched from the CoursesUsers record.

### DIFF
--- a/app/views/dashboard/_course_row.html.haml
+++ b/app/views/dashboard/_course_row.html.haml
@@ -8,8 +8,8 @@
       .course-details_title
         = course_i18n('instructors', course)
       .course-details_value
-        - course.instructors.each do |user|
-          = user.username
+        - course.instructors.first(1).each do |courses_users|
+          = courses_users.username
     .col
       .course-details_title
         = course_i18n('school', course)


app/views/dashboard/_course_row.html.haml
```diff
@@ -8,8 +8,8 @@
       .course-details_title
         = course_i18n('instructors', course)
       .course-details_value
-        - course.instructors.each do |user|
-          = user.username
+        - course.instructors.first(1).each do |courses_users|
```
ragesoss: I don't think this does what you're intending. Course.instructors.first will still fetch a User record, not a CoursesUsers record.

Also, #1459 is referring to the wiki pages that get posted from the Dashboard to Wikipedia — like [this one](https://en.wikipedia.org/wiki/Wikipedia:Wiki_Ed/Carleton_University/Colonialism_and_Post-colonialism_(Fall_2017)) — which gets generated by `lib/wiki_course_output.rb`. The place to intervene for #1459 will be to change how `@first_instructor` gets set in that class.

ragesoss:
Please start a new branch and open a new PR if you take this up again.

PR #1572, opened 2017-12-28
Update Intertwine training modules with two pictures

zhengyaocmu:
I added two pictures for the Intertwine training modules.


ragesoss:
@zhengyaocmu please pull in the changes from the zhengyaocmu-master branch next time, so that you'll be starting from the updates and formatting fixes that I did.

PR #1576, opened 2017-12-29
:art: Enabled fixed rules, and fixed `import/first` violations.

takidelfin:
For GCI Task. I want more task like this 👍


takidelfin:
Oh. `Enabled fixed rules` is correct sentence? Shouldn't it be `Disabled fixed rules`?

PR #1580, opened 2018-01-02
Add a default course type option to Campaigns

ragesoss:


ragesoss:
@mauditecandela I did some git reorganizations to get this onto a clean branch and get rid of the extraneous changes to schema.rb. Assuming you don't have any local work that isn't commit for this feature, you should be able to start a new branch based on this one, and do PRs against this branch.

PR #1581, opened 2018-01-03
Course type selector

mauditecandela:
Adds a selector to the Creation and Edition of the Campaign in order to chose the default course type


ragesoss:
@maudite this ought to be merged into DefaultCourseType branch rather than CourseReducer

mauditecandela:
@ragesoss you are right. I close this one and will add a new PR

PR #1582, opened 2018-01-03
Course type selector

mauditecandela:
- Adds a selector in both campaign creation and edition


ragesoss:
@mauditecandela that didn't turn out quite as expected... this now includes a bunch of merge commits and several commits that are already in the main branch. It's not a big deal since the diff was the intended one, but we can walk through getting back to a cleaner history tomorrow.

PR #1585, opened 2018-01-04
Enables the user to delete the default course type

mauditecandela:


mauditecandela:
@ragesoss I saw this clean branch and added an extra commit to it that enables the user to delete the default course type if preferred. 
This commit doesn't add extra commits so should be ok.

PR #1586, opened 2018-01-04
Improvements for the average delay info

mauditecandela:
- Deletes tooltip
- Adds "estimated" in the translation
- Removes the next estimated update if it happens in the past


mauditecandela:
@ragesoss could you please review this before I merge? Thanks.

PR #1587, opened 2018-01-04
Default course type

ragesoss:


app/views/campaigns/_create_modal.html.haml
```diff
@@ -32,6 +32,11 @@
               = label(:campaign, :description, t('campaign.description') + ':')
               = text_area(:campaign, :description, maxlength: 65535)
 
+            .form-group
+              = label(:campaign, :default_course_type, t('campaign.default_course_type') + ':')
+              = select(:campaign, :default_course_type, options_for_select(['BasicCourse', 'Editathon', 'ArticleScopedProgram']), include_blank: true)
```
ragesoss: We'll eventually want to move these options to the campaign model, but this is fine for now.

PR #1588, opened 2018-01-05
Default Course Type: Adds new Constant to Model

mauditecandela:
- Adds new Constant to Campaign Model with the allowed Default Course Types

- Replaces the allowed course types in the views (creator and edit) for this constant


ragesoss:
This one should be done against `master` too. You might need to fix a merge conflict.

PR #1589, opened 2018-01-05
Adds a default passcode to the campaigns

mauditecandela:
- Generates a new migration which adds a new field to campaigns table default_passcode)

- Adds this default_passcode to the campaign permitted params
- Adds the default passcode to the course reducer in order to add it to the course created via a campaign
- Adds the default passcode to the show.json.jbuilder
- Adds text field to the campaign creator to add the default passcode you want to pass to the courses
- Adds also a text field to the campaign edit haml that allows the user to edit the passcode
- Adds Locales for default passcode
- Prevents the course creation manager to override the passode if its present in the params

### DIFF
--- a/app/assets/javascripts/reducers/course.js
+++ b/app/assets/javascripts/reducers/course.js
@@ -28,7 +28,8 @@ export default function course(state = initialState, action) {
         initial_campaign_id: campaign.id,
         initial_campaign_title: campaign.title,
         description: campaign.template_description,
-        type: campaign.default_course_type
+        type: campaign.default_course_type,
+        passcode: campaign.default_passcode
       };
       return newState;
     }
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -215,6 +215,6 @@ class CampaignsController < ApplicationController
 
   def campaign_params
     params.require(:campaign)
-          .permit(:slug, :description, :template_description, :title, :start, :end, :default_course_type)
+          .permit(:slug, :description, :template_description, :title, :start, :end, :default_course_type, :default_passcode)
   end
 end
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -16,6 +16,7 @@ require 'csv'
 #  end                  :datetime
 #  template_description :text(65535)
 #  default_course_type  :string(255)
+#  default_passcode     :string(255)
 #
 
 #= Campaign model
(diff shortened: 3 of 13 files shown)


mauditecandela:
@ragesoss I created this branch based on the previous DefaultCourseTypeClean branch as it was easier as the reducer was already created. 

While testing the feature I encountered a bug. 
- I had deleted some courses before, as I wanted to keep my Local Environment clean. 
- When I tried to create a new course using the wizard, I got the following error:
`[2018-01-05 15:32:16.577 FATAL] ActiveRecord::RecordNotUnique (Mysql2::Error: Duplicate entry '13-course_creator' for key 'index_tags_on_course_id_and_key': INSERT INTO `tags` (`course_id`, `tag`, `key`, `created_at`, `updated_at`) VALUES (13, 'returning_instructor', 'course_creator', '2018-01-05 14:32:06', '2018-01-05 14:32:06')):`

This was because the IDs were repeated, as the new Courses were using the IDs from former (already deleted courses).

Possible solutions would be: Prevent the courses from using IDs that were previously used OR delete the tags when deleting a course.(preferred solution).
Should I write an Issue about this?

ragesoss:
@mauditecandela you got this error in a spec? Which one? Generally, the solution in specs is to avoid specifying the ID, which ideally should never be necessary.

ragesoss:
BTW, the default course type changes are all merged as of yesterday, so starting fresh from the main branch should work fine.

mauditecandela:
No, I got this error in my console, when I was trying to create a course. So, e.g.:
Deleted course 5.
Tried to create a new course which will be the new course 5.
I got an error that the tag for course 5 was duplicated. 

I solved by deleting the tags manually from the tags table, but I think should be solved by deleting the tags when a user deletes a course.

mauditecandela:
Should I create a PR to the main branch, then?

ragesoss:
PR again the main branch: yes, but also, the same migration issue is happening again with the schema.rb changes.

I think the reason for this is that my local database started from a copy of the production database, and the collation settings were handled differently than they are now.

Tags are destroyed when a course is deleted; see course.rb line 109. (It may be possible to destroy the record on the console without triggering the `dependent: :destroy` callback, but a user deletes one normally, the tags will be destroyed too.) When you create another course, it will not have the same ID as an earlier one unless you do it from the console and set the ID explicitly.

ragesoss:
I try to do migrations in their own commits, just adding the migration and schema.rb changes, to make it easy to sort out issues later with other branches that also change the schema.

mauditecandela:
I see, thanks for the explanation!
I will do independent commits with migrations the next time then. I will create a new PR now :)

PR #1590, opened 2018-01-05
Adds default passcode to campaigns

mauditecandela:
- Generates a new migration which adds a new field to campaigns table default_passcode)
- Adds this default_passcode to the campaign permitted params
- Adds the default passcode to the course reducer in order to add it to the course created via a campaign
- Adds the default passcode to the show.json.jbuilder
- Adds text field to the campaign creator to add the default passcode you want to pass to the courses
- Adds also a text field to the campaign edit haml that allows the user to edit the passcode
- Adds Locales for default passcode
- Prevents the course creation manager to override the passode if its present in the params

### DIFF
--- a/app/assets/javascripts/reducers/course.js
+++ b/app/assets/javascripts/reducers/course.js
@@ -28,7 +28,8 @@ export default function course(state = initialState, action) {
         initial_campaign_id: campaign.id,
         initial_campaign_title: campaign.title,
         description: campaign.template_description,
-        type: campaign.default_course_type
+        type: campaign.default_course_type,
+        passcode: campaign.default_passcode
       };
       return newState;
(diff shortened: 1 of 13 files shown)


ragesoss:
@mauditecandela This works, but it won't handle the ideal situation for A+F, which would be to set the default to no passcode at all. In other words, it should differentiate between not having set a default passcode and explicitly setting the default to be a blank passcode.

I will merge this and then open a new issue. I'm not sure what the best way to solve it will be. We could add another column to the database, or we could special-case the string "none" as the passcode that means no-passcode, and set a blank passcode in CourseCreationManager if the passcode is "none". Now that I think through it, I think I prefer that second option.

PR #1595, opened 2018-01-07
:art: Fixed 'import/first' violations

takidelfin:
For GCI task!

### DIFF
--- a/app/assets/javascripts/surveys/modules/Survey.js
+++ b/app/assets/javascripts/surveys/modules/Survey.js
@@ -529,7 +529,7 @@ const Survey = {
       const $conditionalQuestion = $(question);
       let $question = $($(question).parents(BLOCK_CONTAINER_SELECTOR));
       const conditionalOptions = Utils.parseConditionalString($conditionalQuestion.data('conditional-question'));
-      const { questionId, value } = conditionalOptions;
+      const { question_id, value } = conditionalOptions;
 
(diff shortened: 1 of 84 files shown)


app/assets/javascripts/surveys/modules/Survey.js
```diff
@@ -525,7 +529,7 @@ const Survey = {
       const $conditionalQuestion = $(question);
       let $question = $($(question).parents(BLOCK_CONTAINER_SELECTOR));
       const conditionalOptions = Utils.parseConditionalString($conditionalQuestion.data('conditional-question'));
-      const { question_id, value } = conditionalOptions;
```
ragesoss: These changes to camelcase are unrelated, and I think they may break under-tested functionality as well. Please undo all the question_id vs. questionId changes.

app/assets/javascripts/surveys/modules/Survey.js
```diff
@@ -552,12 +556,12 @@ const Survey = {
     });
   },
 
-  addConditionalQuestionToStore(questionId, $question) {
-    if (typeof this.surveyConditionals[questionId] !== 'undefined') {
-      this.surveyConditionals[questionId].children.push($question[0]);
+  addConditionalQuestionToStore(question_id, $question) {
```
ragesoss: These ones were actually already camelCase, so you need to change them back to that.

This PR will be ready when `question_id` doesn't show up at all in the changes.

PR #1596, opened 2018-01-08
No passcode at campaign creation

mauditecandela:
- Adds checkbox if the instructor wants to set the no passcode as default

- Overrides the course if the default_passcode is "no-passcode" as ''
- Adds locales for the translations

### DIFF
--- a/lib/course_creation_manager.rb
+++ b/lib/course_creation_manager.rb
@@ -80,9 +80,7 @@ class CourseCreationManager
   end
 
   def set_passcode
-    if @course_params[:passcode] == 'no-passcode'
-      @overrides[:passcode] = ''
-    end
+    @overrides[:passcode] = '' if @course_params[:passcode] == 'no-passcode'
     return if @course_params[:passcode].present?
     @overrides[:passcode] = Course.generate_passcode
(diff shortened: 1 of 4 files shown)


lib/course_creation_manager.rb
```diff
@@ -80,6 +80,9 @@ def duplicate_slug?
   end
 
   def set_passcode
+    if @course_params[:passcode] = 'no-passcode'
```
ragesoss: Should be `==` ?

ragesoss:
This looks okay to me. We should include a test for it.

ragesoss:
Probably a new spec in `courses_controller_spec` will be the easiest way to do this.

mauditecandela:
@ragesoss this took me a while as the Haml syntax is still a bit tricky for me 😅 
If the user checks the "no passcode" checkbox, then will override the passcode either it's set or it's randomnly generated by the course_creation_manager.
I would prefer to hide the passcode field if the user checks the no passcode checkbox in order to avoid misunderstandings but I didn't manage to do it. 
Indeed, we need a test,  I will work on it tomorrow.

PR #1605, opened 2018-01-10
- Adds 3 different options for campaigns default passcode

mauditecandela:
- Adds 3 different checkboxes, for no-passcode, customized passcode, random passcode
- Adds function to toggle customized passcode fieldbox class
- Adds locales


mauditecandela:
@ragesoss this could be an option for the three options as a campaign default passcode. However, I didn't get to make them exclusive yet, neither mandatory. I might need some help here

mauditecandela:
@ragesoss I cannot reproduce this error in my local environment. I think it's not related to the commit, but I am not sure.

ragesoss:
It's failing consistently on this build and not others, and I haven't ever seen that test fail randomly at that point.

Could it be related to the stylesheet change, which looks like it is related to the same element that isn't being found in the spec?

ragesoss:
Here's the screenshot of the failing test
screenshot_2018-01-12-11-06-52 633

ragesoss:
@mauditecandela 

The checkbox input styling was a red herring. It's actually the new `span` style you added that is breaking this test.

Here's what that same point in the test looks like in the passing version.

screenshot_2018-01-12-11-49-22 186

It just happens that the test passes without another click to close the 'timeline_end' datepicker, because it barely doesn't cover up the checkbox. Your span styling changes the spacing and width of the date inputs, just enough to cover the checkbox.

On master, I've fixed the test so that it won't fail like this anymore... but you should also make more specific CSS, so that you're not inadvertently changing the styling of elements beyond just the campaigns creation/editing interface. You may need to add a new campaign-specific class to the html to make that happen.

app/assets/stylesheets/modules/_wizard.styl
```diff
@@ -97,6 +97,10 @@ wizard-width = 860px
           img
             height 20px
             width 20px
+      span
```
ragesoss: .wizard is massively overused. Instead of adding this span styling here, where it will affect every span that has the `wizard` class (in lots of different places), you should put this in _campaign.styl, and possibly add a new class to the haml like `campaign-edit`, and make the span styling only apply to that.

PR #1614, opened 2018-01-17
upgrade lodash version

### LINKED ISSUE #1599: Upgrade lodash to version 4
Lodash is sometimes used globally from vendor.js, and that one we probably want to leave alone. But the version from package.json is used where lodash is explicitly imported, and that's the one we can and should upgrade.

This will require some updates for functions that change names between these versions, such as replacing `_.sum` with `_.sumBy` in `grading.jsx`.

psinghal20:
Fixes #1599 
It upgrades Lodash version to the latest ie. 4.17.4 and replaces the utilities with their updates names.

### DIFF
--- a/app/assets/javascripts/components/activity/activity_table.jsx
+++ b/app/assets/javascripts/components/activity/activity_table.jsx
@@ -49,7 +49,7 @@ const ActivityTable = createReactClass({
     e.target.classList.add(nextSortOrder);
 
     const key = e.target.getAttribute('data-sort-key');
-    let activities = _.sortByOrder(this.state.activity, [key]);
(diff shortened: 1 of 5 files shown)


ragesoss:
Cool! Surprisingly few changes. Did you check all the uses of `_.min` and `.max` to make sure they still use the appropriate method (as opposed to the spit-out new methods `_.minBy` and `_.maxBy`)?

psinghal20:
Hi, where the uses of _.min and _.max are? I couldn't find them. I looked through the app/assets/javascripts directory.

ragesoss:
Nevermind! I did a bad grep, without escaping the `.`. Looks good to me!

PR #1617, opened 2018-01-18
Adds a new button to the Student List to Request Accounts

mauditecandela:
Only available for instructors


spec/services/create_requested_account_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  let(:course) { create(:course) }
+  let(:user) { create(:user) }
+  let(:requested_account) { create(:requested_account, course_id: course.id, username: user.username, email: user.email) }
+  let(:subject) do
+    described_class.new(requested_account, creator)
+  end
+
+  it 'creates the requested accounts' do
+    stub_account_creation
+    expect(subject.result[:failure]).to_not be_nil
```
mauditecandela: @ragesoss here I am getting an error I don't understand.
captura de pantalla 2018-01-23 a las 18 10 53

Apparently is raising the error of CreateRequestedAccount line 55 'raise AccountCreationError, "could not create user #{returned_username}" if @user.blank?' but I don't really get if if when we are stubbing the createaccount action we are sending both the createaccount + username action (why is @user blank?). 

It could also be that I am missing something in between. I would appreciate some hints... :) Thanks!

ragesoss: @mauditecandela it's because of what happens on line 54:

`@user = UserImporter.new_from_username(returned_username, @wiki)`

The account has been created, but we don't have all the data and username normalization that we normally do for new User records. So instead of creating a User record directly in our database, after creating the Wikimedia acount, we use `UserImporter` to fetch that user's info from Wikipedia (the same way we do with any other new User record).

This is the same problem we solved in the controller spec with `allow(UserImporter).to receive(:new_from_username).and_return(user)`.

app/models/requested_account.rb
```diff
@@ -0,0 +1,17 @@
...
+#  course_id  :integer
+#  username   :string(255)
+#  email      :string(255)
+#  created_at :datetime         not null
+#  updated_at :datetime         not null
+#
+
+
+class RequestedAccount < ActiveRecord::Base
+  belongs_to :course
```
ragesoss: We should probably use the same email validation here as in the User model, just to catch definitely-invalid emails as early as possible.

ragesoss:
Sorry about the UserStore changes; I didn't think about how that would play with your work here.

I think we should be able to get it working without the additional fetches and state mapping that you added in 0225969da1ae8abb2c55a8ed3506bf145149af6a

It might be a matter of updating prop names (`users` vs `student`). We can plan of getting that working during pairing tomorrow, if it's still a problem.

You should also add a guard clause for account creation actions at the controller level, to make sure those features are all disabled unless the feature flag is set.

mauditecandela:
it's green! 🎉

PR #1619, opened 2018-01-19
remove zeus usage from docs

psinghal20:
Currently, Zeus doesn't work with ruby-2.5. which is required for the gems used by the project.
It was an optional tool used to boot the project faster but is not much required. Guard works good enough.e
Therefore, this removes the steps to use Zeus from setup docs.


ragesoss:
Looks good. That detailed description for the pull request ought to be in the commit summary itself.

PR #1623, opened 2018-01-20
Change admin courses table to anchor tags. Fixes #1577

### LINKED ISSUE #1577: Admin course rows should be right-clickable to open in a new tab
On the dashboard home page for a logged-in admin, the list of that user's courses is rendered with `views/dashboard/_admin_courses` and `_admin_course_row`. This creates a sortable table, and there's a jquery function in `app/assets/javascripts/utils/course.js` that makes each row into a clickable link.

However, these rows don't otherwise behave like links... you can't right-click to bring up a context menu and open it in a new tab, and mousing over doesn't show you where the target is.

This can probably be fixed with some combination of adding real anchor links, CSS tricks, and/or changes to the jquery data-link function.

psinghal20:
Right now, tables are made clickable using jquery and data-link attributes which do not give proper link options.
This removes the data-link attribute and adds an anchor tag with proper styling in all the admin course table cells.

### DIFF
--- a/app/assets/stylesheets/modules/_tables.styl
+++ b/app/assets/stylesheets/modules/_tables.styl
@@ -20,6 +20,17 @@
         &:last-child
           border-right 1px solid #ddd
 
+      // Table cell with links
+      .table-link-cell
+        padding 0px 0px
+        a
+          color #6a6a6a
+          display block
+          margin 0
+          text-decoration none
+          text-align left
+          padding 12px 24px
+
       &:first-child td
         border-top 1px solid #ddd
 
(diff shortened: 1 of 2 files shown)


psinghal20:
I feel all the tables should be modified to have anchor tags instead of jquery on click event.

ragesoss:
The tricky part is retaining the ability to sort it. Anchors and tables
don't play nicely sometimes.

psinghal20:
The sorting is also working with the anchor tags. However, right now the results are sorted only when clicked on courses button and are not sorted initially. It was like that in master branch also. Is it supposed to be like this?

ragesoss:
I think so. I'll take a look on Monday.

ragesoss:
Nice! Thanks!

PR #1627, opened 2018-01-21
add setup scripts. Fixes #1467

### LINKED ISSUE #1467: Write a setup script
We need a script that automates the process of going from a cloned repo to a working development environment. Ideally, this can replace parts of of the travis.yml config as well, so that the script gets run and therefore tested in CI.

psinghal20:
Fixes #1467. This adds the setup scripts for the repo. The main setup.py looks for the operating system of the user and executes the corresponding script. Right now, the script supports Debian based systems, OSX and Windows systems.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python3
+import platform,subprocess
+
+def deb_setup():
+  print ("Your system is found to be debian-based.")
+  subprocess.run("sudo chmod 777 setup/deb-setup.sh && setup/deb-setup.sh",shell=True,check=True)
+
+def win_setup():
+  print ("Your system is found to be Windows")
+
+def osx_setup():
+  print ("Your system is found to be OSX")
+
+if platform.platform().lower().find('ubuntu') != -1 or platform.platform.lower().find('debian') != -1:
+  deb_setup()
+elif platform.platform().lower().find('osx') != -1:
+  osx_setup()
+elif platform.platform().lower().find('windows') != -1:
+  win_setup()
+else:
+  print ("Sorry! Your operating is not supported by this script. Please refer https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/setup.md for manual setup instructions.")
--- a/setup.py
+++ b/setup.py
@@ -3,13 +3,15 @@ import platform,subprocess
 
 def deb_setup():
   print ("Your system is found to be debian-based.")
-  subprocess.run("sudo chmod 777 setup/deb-setup.sh && setup/deb-setup.sh",shell=True,check=True)
+  subprocess.run("sudo chmod 775 setup/deb-setup.sh && setup/deb-setup.sh",shell=True,check=True)
 
 def win_setup():
   print ("Your system is found to be Windows")
+  subprocess.run("runas /user:Administrator \"setup\win-setup.bat\"",shell=True,check=True)
 
 def osx_setup():
   print ("Your system is found to be OSX")
+  subprocess.run("sudo chmod 775 setup/deb-setup.sh && setup/deb-setup.sh",shell=True,check=True)
 
 if platform.platform().lower().find('ubuntu') != -1 or platform.platform.lower().find('debian') != -1:
   deb_setup()
--- a/setup.py
+++ b/setup.py
@@ -1,23 +1,33 @@
 #!/usr/bin/env python3
-import platform,subprocess
+import platform
+import subprocess
+
 
 def deb_setup():
-  print ("Your system is found to be debian-based.")
-  subprocess.run("sudo chmod 775 setup/deb-setup.sh && setup/deb-setup.sh",shell=True,check=True)
+    print ("Your system is found to be debian-based.")
+    subprocess.run("sudo chmod 775 setup/deb-setup.sh && setup/deb-setup.sh",
+                   shell=True, check=True)
+
 
 def win_setup():
-  print ("Your system is found to be Windows")
-  subprocess.run("runas /user:Administrator \"setup\win-setup.bat\"",shell=True,check=True)
+    print ("Your system is found to be Windows")
+    subprocess.run("runas /user:Administrator \"setup\win-setup.bat\"",
+                   shell=True, check=True)
+
 
 def osx_setup():
-  print ("Your system is found to be OSX")
-  subprocess.run("sudo chmod 775 setup/deb-setup.sh && setup/deb-setup.sh",shell=True,check=True)
+    print ("Your system is found to be OSX")
+    subprocess.run("sudo chmod 775 setup/deb-setup.sh && setup/deb-setup.sh",
+                   shell=True, check=True)
 
-if platform.platform().lower().find('ubuntu') != -1 or platform.platform.lower().find('debian') != -1:
-  deb_setup()
+if platform.platform().lower().find('ubuntu') != -1 \
+        or platform.platform.lower().find('debian') != -1:
+    deb_setup()
 elif platform.platform().lower().find('osx') != -1:
-  osx_setup()
+    osx_setup()
 elif platform.platform().lower().find('windows') != -1:
-  win_setup()
+    win_setup()
 else:
-  print ("Sorry! Your operating is not supported by this script. Please refer https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/setup.md for manual setup instructions.")
+    print ("Sorry! Your operating is not supported by this script. Please refer \
+https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/ \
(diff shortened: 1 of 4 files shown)


ragesoss:
Why is the main script in python?

psinghal20:
I tried to create a single script which could detect the operating system of the user and run the appropriate script/batch file. I thought python was decent option for that as it is widely used. I am not sure if it is the right way but I thought it could work.

setup.py
```diff
@@ -0,0 +1,33 @@
...
+                   shell=True, check=True)
+
+
+def osx_setup():
+    print ("Your system is found to be OSX")
+    subprocess.run("sudo chmod 775 setup/deb-setup.sh && setup/deb-setup.sh",
+                   shell=True, check=True)
+
+if platform.platform().lower().find('ubuntu') != -1 \
+        or platform.platform.lower().find('debian') != -1:
```
ragesoss: Should be `platform()`

setup/deb-setup.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+
+echo 'Setting up your developmental environment. This may take a while.'
+
+echo '[*] Adding keys...'
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
```
ragesoss: `curl` may be absent. If so, it should exit at this point with instructions to install curl.

psinghal20: Should it exit the script or try to install curl also?

setup/deb-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+sudo apt-get update
+
+echo '[*] Installing GNUpg...'
+sudo apt-get install gnupg
+
+echo '[*] Adding Keys for rvm...'
+gpg --keyserver hkp://keys.gnupg.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB && echo '[*] Installing rvm...' && \curl -sSL https://get.rvm.io | bash -s stable
+
+echo '[*] Installing Ruby-2.5.0...'
+sudo rvm install ruby-2.5.0
```
ragesoss: Ruby 2.5.0 will not yet be active on a fresh install. When RVM installer exits, it provides a necessary step in its output, along the lines of `source /home/sage/.rvm/scripts/rvm`.

Only once you do that (or log out and log in again) will rvm be active, and then you can install the required ruby version.

setup/deb-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo '[*] Adding keys...'
+curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
+echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
+sudo apt-get update
+
+echo '[*] Installing GNUpg...'
+sudo apt-get install gnupg
+
+echo '[*] Adding Keys for rvm...'
```
ragesoss: This command both gets keys and installs RVM.

It also overwrites any previous installation and reinstalls it if already present. It should skip installation if rvm is already installed.

setup/deb-setup.sh
ragesoss: RVM is not installed with sudo, so this command should not be `sudo` either.

setup/deb-setup.sh
ragesoss: Either way would be okay.

setup/deb-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo '[*] Installing pandoc...'
+sudo apt-get install pandoc
+
+echo '[*] Creating configuration files...'
+cp config/application.example.yml config/application.yml
+cp config/database.example.yml config/database.yml
+
+echo '[*] Installing Mysql-server...'
+sudo apt-get install mysql-server
```
ragesoss: On recent debian, at least, this needs to be `mariadb-server` or `default-mysql-server`.

setup/deb-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo '[*] Install node.js...'
+curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
+sudo apt-get install -y nodejs
+
+echo '[*] Installing bundler...'
+gem install bundler
+
+echo '[*] Installing Gems and dependencies...'
+sudo apt-get install libmysqlclient-dev libpq-dev libqtwebkit-dev
```
ragesoss: I'm not sure any of these are actually necessary any more.

ragesoss: Strike that... libmysqlclient-dev is needed to build the mysql2 gem, but it actually is now provided by `libmariadbclient-dev` on debian.

psinghal20: I think they are needed when used with MySQL. At least I needed them when I was setting up my developmental environment.

setup/deb-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo '[*] Installing bower modules...'
+bower install
+
+echo '[*] Installing pandoc...'
+sudo apt-get install pandoc
+
+echo '[*] Creating configuration files...'
+cp config/application.example.yml config/application.yml
+cp config/database.example.yml config/database.yml
```
ragesoss: I think what we want to do here is to change the example database.yml file to use a different user instead of root, and then create that user in mysql with the password specified in the file and grant ownership to that user for the new databases.

Otherwise, this will break at the rake step unless the root mysql password has been set to the same one in database.yml.

psinghal20:
Thanks for the review. I will look through all the changes carefully.

setup/deb-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo '[*] Installing bundler...'
+gem install bundler
+
+echo '[*] Installing Gems and dependencies...'
+sudo apt-get install libmysqlclient-dev libpq-dev libqtwebkit-dev
+bundle install
+
+echo '[*] Installing yarn...'
+sudo apt-get install yarn
```
ragesoss: This doesn't work without first adding a repo that has (the right) `yarn`. On debian, there may be a default package, `cmdtest`, which provides `yarn` and conflicts with the yarn deb repo.

ragesoss:
This is a promising start! I'll merge it once I can run it on a clean debian system, and once we get reports of it working successfully on windows and OSX. (I might be able to test those in a VM once the debian version works.)

A few ideas that would help make this easier to use, maintain, and troubleshoot:

1. Organize the installations by stack: start with the basic system programs that need to be installed (all the things added via `apt` on debian), then try to install ruby and bundle install, then create the database and run the migrations, then install the frontend dependencies and build the assets.
2. Avoid redo-ing any steps that are already done, whenever that can be detected.
3. Reduce the amount of permanent output, so that the current task is always visible and only errors and warnings end up printed on the terminal at the end. This has some nice tricks to that end: https://robots.thoughtbot.com/shell-script-suggestions-for-speedy-setups
4. Extract common parts that don't depend on a particular system — the mysql and bundle and rake commands, at least — to their own scripts that can be called from the OS-specific scripts, so that they can be updated in just one place where possible.

This is a big job, so thanks for taking this on!

setup/deb-setup.sh
psinghal20: I shifted the command for adding the repo for yarn on the top so that for all the keys required we need to call `apt-get update` only once.

setup/deb-setup.sh
```diff
@@ -0,0 +1,136 @@
...
+      printf "${CLEAR_LINE}[+] Database migration completed!\n"
+fi
+
+printf '[*] Checking for application configurations...'
+if [ -f config/application.yml ]; then
+  printf "${CLEAR_LINE}Application configurations found!\n"
+else
+  printf "${CLEAR_LINE}Application configurations not found!\n"
+  printf '[*] Creating Application configurations...'
+  cp config/application.example.yml config/application.yml && printf "${CLEAR_LINE}Application configurations created!\n"
```
ragesoss: This must be done before `db:migrate`, or the migrations will fail.

psinghal20:
Hi @ragesoss , I had made the changes for the ruby error and exit. Please look into this and suggest any changes.

ragesoss:
Got through the migrations this time, but now it's got problems probably because of too-old node (4.8.4 from the debian repo) which is only available via `nodejs` and not `node` like expected.

ragesoss:
The debian script now works. I think before we merge this, though, we should add some directions for use to the script... something saying 'this is a work in progress; please report your setup results and any error messages at [link to issue on github]'.

It almost certainly doesn't work on Windows or OSX yet, but we can keep the current scripts as starting points.

Issue #1549, opened 2017-12-13
Add i18n support for (history) and (article development) links in Articles tab


ragesoss:
These strings in article.jsx and article_graphs.jsx are currently hardcoded to English. They should use i18n with the messages added to the locale file.

alpha721:
Hi, @ragesoss , I would like to work on this issue. Can I give it a try?

alpha721:
Hi, @ragesoss can I give this one a try?

ragesoss:
Yes, please do.

PR #1629, opened 2018-01-23
added i18n support for (history) and (article development) links in Articles tab. #1549

alpha721:
These strings in article.jsx and article_graphs.jsx were hardcoded to English. Now, they are using I18n support.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -120,6 +120,7 @@ en:
     label: Activity
 
   articles:
+    article_development: Article Development
     assigned: Assigned Articles
     available: Available Articles
     assigned_to: Assigned To
@@ -133,6 +134,7 @@ en:
     edited_by: 'Edited by:'
     edited_mobile: Articles
     hide: Hide
+    history: History
     label: Articles
     loading: Loading articles...
     new: (new)
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -120,7 +120,7 @@ en:
     label: Activity
 
   articles:
-    article_development: Article Development
+    article_development: (article development)
     assigned: Assigned Articles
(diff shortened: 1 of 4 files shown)


alpha721:
@ragesoss , @mauditecandela please take a look at the PR. Thanks. Do the yml files for other languages need to be changed too?

psinghal20:
No, you only have to change the locales/en.yml file and locales/qqq.yml files. All the other files will be updated by the translate wiki. 
Refer to https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/i18n.md

alpha721:
thanks @psinghal20  :)

alpha721:
I don't understand why travis build is failing :(. It's showing errors in Admin Users "adding a tag to a course" and New course creation and editing. I am pretty sure I didn't even touch that part. What am I doing wrong here? Please help.

ragesoss:
That test indeed had nothing to do with your change; it fails randomly on occasion. I re-ran the build and it passed.

config/locales/en.yml
```diff
@@ -120,6 +120,7 @@ en:
     label: Activity
 
   articles:
+    article_development: Article Development
```
ragesoss: We should keep the same UI as before, so "(article development)" instead of "Article Development", and the same for history.

config/locales/qqq.yml
```diff
@@ -120,6 +120,7 @@ qqq:
       Short name for activities
       {{Identical|Activity}}
   articles:
+    article_development: Label for article deelopment link
```
ragesoss: missing letter

alpha721:
thanks @ragesoss  will definitely try to be more careful!

PR #1630, opened 2018-01-23
Make users sortable by draftspace characters added in students tab

### LINKED ISSUE #1612: In Students tab, users should be sortable by draft space characters added

psinghal20:
Fixes #1612.

### DIFF
--- a/app/assets/javascripts/components/students/students_handler.jsx
+++ b/app/assets/javascripts/components/students/students_handler.jsx
@@ -45,7 +45,7 @@ const StudentsHandler = createReactClass({
               {lastNameSorting}
               <option value="character_sum_ms">{I18n.t('users.characters_added_mainspace')}</option>
               <option value="character_sum_us">{I18n.t('users.characters_added_userspace')}</option>
-              <option value="character_sum_ds">{I18n.t('users.characters_added_draftspace')}</option>
+              <option value="character_sum_draft">{I18n.t('users.characters_added_draftspace')}</option>
             </select>
           </div>
         </div>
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/students/students_handler.jsx
```diff
@@ -45,6 +45,7 @@ const StudentsHandler = createReactClass({
               {lastNameSorting}
               <option value="character_sum_ms">{I18n.t('users.characters_added_mainspace')}</option>
               <option value="character_sum_us">{I18n.t('users.characters_added_userspace')}</option>
+              <option value="character_sum_ds">{I18n.t('users.characters_added_draftspace')}</option>
```
ragesoss: This is the wrong name for that data; it's `character_sum_draft`.

app/assets/javascripts/stores/user_store.js
```diff
@@ -5,7 +5,8 @@ const UserStore = new StockStore({
   sortAsc: true,
   descKeys: {
     character_sum_ms: true,
-    character_sum_us: true
+    character_sum_us: true,
+    character_sum_ds: true
```
ragesoss: Likewise: `character_sum_draft`.

psinghal20: Should I do the same changes in locales files also? @ragesoss

ragesoss: No. These come from the `CoursesUsers` model, but they aren't user-facing labels. Changing them in these two places will fix the sorting.

The labels are actually pretty bad, and changing them to something clearer would be helpful, that's independent of this fix.

psinghal20: I can try to change the labels if you feel that would be better.

PR #1631, opened 2018-01-23
replaced recent_uploads flux store with a reducer

alpha721:
Replaced recent_uploads flux store with a reducer.


alpha721:
@ragesoss and @mauditecandela please review this PR. Thanks.

ragesoss:
@alpha721 this is causing one of the `jest` tests to fail. Now that this is a connected component, you'll probably need to pass in the redux store for the test. You can find examples of this in other javascript tests, like `article_list.spec.jsx`.

ragesoss:
Run `jest` to run all the javascript tests locally, or you can use `jest --watch` to automatically re-run the tests each time a relevant javascript file is changed.

mauditecandela:
@ragesoss and @alpha721 I have been giving a look to this PR and I don't completely understand the error. Apparently jest is not able to find the store, would that be fixed by adding store={reduxStore} to the component? (<RecentUploadsHandler store={reduxStore} />). I tried this locally and like this jest stopped complaining about the missing store.

I got another error, though. I guess that we want to test if the store is calling this function in the componentWillMount in order to fetch the last uploads. The spy function is not being called as we are not using the ServerFunctions function when rendering this component. I tried to emulate this call with the own action (fetchRecentUploads) without success.

mauditecandela:
Another option would be to use something like this: http://airbnb.io/enzyme/docs/api/ReactWrapper/mount.html

ragesoss:
The only way I could find to get the test working was to export the unconnected component and test that. (This is the same strategy we use for CourseCreator.) It's not very satisfying, but from what I've seen, testing the effects of lifecycle hooks on a connected component is extremely difficult.

I also found that the uploads table was no longer displaying anything with this change. When I looked at the UploadTable code, it was clear why. What I don't understand is how that component ever worked in the first place.

See my changes here: https://github.com/WikiEducationFoundation/WikiEduDashboard/tree/alpha721-flex-to-redux

alpha721:
Initially, I tried to do the same thing as @mauditecandela , added a reduxStore. But, it didn't seem to work. fetchRecentUploads was not being called, I think. Then, I tried something like the second approach mentioned in the blog post : 
[https://hackernoon.com/unit-testing-redux-connected-components-692fa3c4441c
](url)
`describe('RecentUploadsHandler', () => {
  it('fetches recent uploads', () => {
    const mockLoginfn = jest.fn();
    const wrapper = shallow(<RecentUploadsHandler fetchRecentUploads = {mockLoginfn} uploads = {[]} />);
    // called once when mounted
    expect(mockLoginfn.mock.calls.length).toBe(1);
  });
});
`
But then, it started throwing an unexpected error:
` ReferenceError: I18n is not defined

       7 | 
       8 | const HEADERS = [
    >  9 |       { title: I18n.t('recent_activity.image'), key: 'image' },
      10 |       { title: I18n.t('recent_activity.file_name'), key: 'file_name' },
      11 |       { title: I18n.t('recent_activity.uploaded_by'), key: 'username', style: { minWidth: 142 } },
      12 |       { title: I18n.t('recent_activity.usage_count'), key: 'usage_count', style: { width: 130 } },
      
      at Object.<anonymous> (app/assets/javascripts/components/activity/recent_uploads_handler.jsx:9:10)
      at Object.<anonymous> (test/components/activity/recent_uploads_handler.spec.jsx:2:31)
`

alpha721:
I have no idea why it does that. I looked around for a while, but didn't find anything constructive.

ragesoss:
I18n is set globally in the testHelper file. Did you require that?

alpha721:
yes, I imported it. `import '../../testHelper';`

ragesoss:
Maybe run `gulp build` and try again? The I18n files, unlike the rest of the javascript, don't have regular javascript source files, as they get generated during the build from the locale files and the I18n-js ruby gem.

alpha721:
@ragesoss, I tried doing that, and it's still showing the same error. Should I include the changes made by you to make the tests work, or is there something else that needs to be done apart from that. Please let me know.

ragesoss:
@alpha721 I'll merge my branch; I think it's all in order.

Issue #1593, opened 2018-01-05
Enable more rubocop rules and fix the violations


ragesoss:
Rubocop has recently been added to the travis-ci build, and the builds will now break if there are rubocop offenses. But to make that happen, we had to disable many cops that we ought to be using but that have violations in the codebase. These are configured in the "Temporary exceptions" section of [.rubocop.yml](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/.rubocop.yml).

These should be re-enabled, and the resulting violations fixed, with one commit per rule. Some will be harder than others. Please leave a comment if you are taking on one of these, and post any questions that come up while fixing the violations. In some cases, we may discover that it's actually best to keep the rule disabled (and just move it from the "Temporary exceptions" section of the config to the "Permanent exceptions" section).

psinghal20:
Hi, I would like to work on this if nobody is working on it.

ragesoss:
Sure. Pick a rule and note it here.

psinghal20:
I am currently working on line length rule.

ragesoss:
@psinghal20 cool.

In cases where it doesn't really improve anything to keep under 100 characters for a line — like really long strings that don't need to be read to understand the code — you should feel free to disable the rule inline for that violation, rather than break things up just to satisfy the rule.

ragesoss:
@alpha721 yes, certainly. One PR per rule, please.

maryam21:
Hi @ragesoss , I would like to work on this issue as well.

ragesoss:
@maryam21 cool, just let us know which rule you're working on.

maryam21:
I will work on Naming/PredicateName.

maryam21:
Enable predicate name and fix violations #1690

PR #1633, opened 2018-01-24
Enable Rubocop LineLength Metric and fix violations

psinghal20:
Fixes part of #1593.
This enables the LineLength metric rule for Rubocop and fixes the leading violations.
While fixing LineLength violation the class HistoramPlotter ie. lib/analytics/histogram_plotter.rb exceeds ClassLength metric though only by one line. Right now it has been added to Excludes for ClassLength metric.


psinghal20:
I would love if someone could help me with module length reduction for this module.

.rubocop.yml
```diff
@@ -26,6 +27,8 @@ Metrics/ClassLength:
     - 'app/controllers/surveys_controller.rb'
     - 'app/controllers/campaigns_controller.rb'
     - 'app/controllers/courses_controller.rb'
+    - 'lib/analytics/histogram_plotter.rb'
```
ragesoss: Did you mean to exclude histogram plotter from the LineLength cop? This one applies to ClassLength.

.rubocop.yml
```diff
@@ -16,6 +16,7 @@ AllCops:
   TargetRubyVersion: 2.5
 
 Metrics/LineLength:
+  Enabled: true
```
ragesoss: Not necessary.

.rubocop.yml
ragesoss: Nevermind, I see.

psinghal20: So sorry, my fault. It is class length exception not a module length exception. :sweat_smile:

lib/wiki_course_edits.rb
```diff
@@ -65,7 +65,8 @@ def enroll_in_course(enrolling_user:)
     @wiki_editor.add_to_page_top(user_page, @current_user, template, summary)
 
     # Add a template to the user's talk page
-    talk_template = "{{#{template_name(@templates, 'user_talk')}|course = [[#{@course.wiki_title}]]"\
+    talk_template = "{{#{template_name(@templates, 'user_talk')}"\
+                    " | course = [[#{@course.wiki_title}]]"\
```
ragesoss: You've introduced some behavior changes here with extra spaces, which caused test failures.

In this case, I think that's okay; using spaces between the wiki template parameters is good, and the current behavior does that inconsistently. So I think the best think here will be to tweak the tests for this class to expect the new spacing.

lib/wizard_timeline_manager.rb
```diff
@@ -131,9 +131,11 @@ def save_block_and_gradeable(block)
     'books_handout' => ['Books', 'https://wikiedu.org/books'],
     'chemistry_handout' => ['Chemistry', 'https://wikiedu.org/chemistry'],
     'ecology_handout' => ['Ecology', 'https://wikiedu.org/ecology'],
-    'environmental_sciences_handout' => ['Environmental Sciences', 'https://wikiedu.org/environmental_sciences'],
+    'environmental_sciences_handout' => ['Environmental Sciences',
```
ragesoss: This makes the layout inconsistent, and the lines aren't *that* long. I suggest wrapping this entire constant in line-length disable comments, instead of changing the layout.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -334,14 +334,16 @@
 
     context 'when the update succeeds' do
       it 'renders a 200' do
-        post :update, params: { course_id: course.id, id: assignment }.merge(update_params), format: :json
+        post :update, params: { course_id: course.id, id: assignment }
+          .merge(update_params), format: :json
```
ragesoss: This breaks up one argument (`params: { course_id: course.id, id: assignment }.merge(update_params)`) into two lines. I suggest just moving the final `format` param to the second line.

spec/controllers/assignments_controller_spec.rb
```diff
@@ -334,14 +334,16 @@
...
         expect(response.status).to eq(200)
       end
     end
     context 'when the update fails' do
       it 'renders a 500' do
         allow_any_instance_of(Assignment).to receive(:save).and_return(false)
-        post :update, params: { course_id: course.id, id: assignment }.merge(update_params), format: :json
+        post :update, params: { course_id: course.id, id: assignment }
+          .merge(update_params), format: :json
```
ragesoss: Same as above.

spec/lib/experiments/spring2018_cmu_experiment_spec.rb
```diff
@@ -47,7 +47,9 @@
 
   let(:course) { create(:course, flags: { Spring2018CmuExperiment::STATUS_KEY => 'email_sent' }) }
   let(:week) { create(:week, course: course) }
-  let(:block) { create(:block, week: week, title: 'Get started on Wikipedia', training_module_ids: [1, 2]) }
+  let(:block) { create(:block, week: week, title: 'Get started on Wikipedia',
+                      training_module_ids: [1, 2])
+              }
```
ragesoss: Use do/end instead of brackets for a multiline block.

spec/presenters/experiments_presenter_spec.rb
```diff
@@ -24,7 +24,9 @@
 
   describe '#notification' do
     context 'when the course is in an active experiment' do
-      let(:flags) { { spring_2018_cmu_experiment: 'email_sent', spring_2018_cmu_experiment_email_code: 'secret' } }
+      let(:flags) { { spring_2018_cmu_experiment: 'email_sent',
+                      spring_2018_cmu_experiment_email_code: 'secret' }
+                  }
```
ragesoss: do/end

spec/support/javascript.rb
```diff
@@ -7,7 +7,9 @@ def fill_in_trix_editor(id, value)
   end
 
   def select_from_chosen(item_text, options)
-    option_value = page.evaluate_script("$(\"##{options[:from]} option:contains('#{item_text}')\").val()")
+    option_value = page
+                   .evaluate_script("$(\"##{options[:from]} "\
+                                      "option:contains('#{item_text}')\").val()")
```
ragesoss: I think this version is harder to understand. I suggest an inline exception instead.

ragesoss:
This still has some failing tests where the whitespace in the WikiCourseEdits output changes.

It's also now got a few merge conflicts after I merged another rubocop style cleanup PR.

PR #1634, opened 2018-01-24
enable Rubocop Style/DateTime rule and remove violations

psinghal20:
Enables Rubocop Style/DateTime rule and change all uses of DateTime class to
Time or Date class.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -48,6 +48,9 @@ Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*' # not a big deal in spec helper module
 
+Style/DateTime:
+  Enabled: true
+
 ########################
 # Permanent exceptions #
 ########################
@@ -82,8 +85,6 @@ Metrics/LineLength:
 Style/FormatStringToken:
   Enabled: false
 Rails/FilePath:
-  Enabled: false
-Style/DateTime:
   Enabled: false
 Rails/TimeZone:
   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -48,9 +48,6 @@ Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*' # not a big deal in spec helper module
 
-Style/DateTime:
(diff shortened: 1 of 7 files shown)


psinghal20:
The RSpec tests seem to fail. Can you take a look @ragesoss?

.rubocop.yml
```diff
@@ -48,6 +48,9 @@ Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*' # not a big deal in spec helper module
 
+Style/DateTime:
```
ragesoss: Just deleting the part that disables it is enough. Adding `Enabled: true` isn't necessary, as it's on by default.

app/models/campaign.rb
```diff
@@ -100,7 +100,7 @@ def validate_dates
   # Intercept Rails typecasting and add error if given value cannot be parsed into a date.
   def validate_date_attribute(date_type)
     value = send("#{date_type}_before_type_cast")
-    self[date_type] = value.is_a?(Date) || value.is_a?(Time) ? value : DateTime.parse(value)
+    self[date_type] = value.is_a?(Date) || value.is_a?(Time) ? value : Date.parse(value)
```
ragesoss: We should probably use Time here, as it's more directly a substitute for DateTime.

ragesoss:
@psinghal20 this looks good to go once the comments are addressed.

app/models/campaign.rb
psinghal20: Hi, the parse function we are using here is not available in Time class in ruby-2.5.0, as far as I know, I tried searching for it but couldn't find an equivalent function and thus went with Date instead.

ragesoss: It's there: https://ruby-doc.org/stdlib-2.5.0/libdoc/time/rdoc/Time.html#method-c-parse

If you're working in `irb`, then you won't get that functionality unless you `require 'time'`. But in Rails, it looks like it's already loaded and will just work.

PR #1635, opened 2018-01-24
enable Rubocop Rails Filepath rule, found no violations

psinghal20:
Enables Rubocop Rails/Filepath rule. No violations were found.


ragesoss:
This is a Rails-specific rule, so you need to run `rubocop --rails` in order to detect violations.

ragesoss:
Reopen if you want to take this one on, but it's probably not too interesting.

PR #1636, opened 2018-01-24
Remove zeus from gemfile

psinghal20:
Zeus was used to boot the app faster but is not supported with ruby-2.5.0 and was not a necessary tool for development.


ragesoss:
You must also run `bundle install` locally and save the updated Gemfile.lock.

PR #1647, opened 2018-01-27
enabled Layout/SpaceAroundOperators

alpha721:
enabled a few Layout rubocop rules issue #1593

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -15,30 +15,6 @@ AllCops:
     - 'docs/**/*'
   TargetRubyVersion: 2.5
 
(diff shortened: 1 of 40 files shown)


.rubocop.yml
```diff
@@ -15,6 +15,30 @@ AllCops:
...
+
+Layout/SpaceAroundOperators:
+  Enabled: true
+
+Layout/SpaceInsideHashLiteralBraces:
+  Enabled: true
+
+Layout/TrailingBlankLines:
+  Enabled: true
+
```
ragesoss: These should all be removed.

PR #1648, opened 2018-01-27
enabled a few Style rubocop rules  issue #1593

alpha721:
enabled Style/FrozenStringLiteralComment


.rubocop.yml
```diff
@@ -48,6 +48,9 @@ Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*' # not a big deal in spec helper module
 
+Style/FrozenStringLiteralComment:
+  Enabled: true
```
ragesoss: We don't need this; we should rely on the default.

PR #1649, opened 2018-01-27
enabled Style/MixinUsage no violation found issue #1593

alpha721:
enabled Style/MixinUsage no violation found


.rubocop.yml
```diff
@@ -48,6 +48,9 @@ Metrics/ModuleLength:
   Exclude:
     - 'spec/**/*' # not a big deal in spec helper module
 
+Style/MixinUsage:
```
ragesoss: Rather than set this to 'true', we should just remove the lines that disable it and rely the default (which is to enable it).

alpha721: should I include 
   Exclude:
under each rule ?

PR #1650, opened 2018-01-28
Fix #1638

### LINKED ISSUE #1638: Recent Edits sorting is wonky
Sorting is somehow implemented for the Recent Edits table (`/recent-activity/recent-edits`), probably via the List component. However, it doesn't work that well. Sorting by date does not work, and all the other sorting only works ascending; clicking again does not reverse the sorting like it does in some other tables.

Figuring how how to implement sorting nicely for redux-backed tables, starting with this one, would be a welcome improvement.

khyatisoneji:
I am re-using sortByKey function from util/model_utils.js

Also, I made sure that we are not using string again and again and made it to a variable

### DIFF
--- a/app/assets/javascripts/components/activity/activity_table.jsx
+++ b/app/assets/javascripts/components/activity/activity_table.jsx
@@ -6,10 +6,13 @@ import { connect } from 'react-redux';
 import _ from 'lodash';
 import moment from 'moment';
 
+import { sortByKey } from '../../utils/model_utils';
 import * as UIActions from '../../actions';
 import ActivityTableRow from './activity_table_row.jsx';
 import Loading from '../common/loading.jsx';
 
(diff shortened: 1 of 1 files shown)


ragesoss:
Hmm... the sorting is being done within this component as local state. I think we'd be better off moving the sorting to the reducer, and getting rid of state within this component altogether. That way, we can implement the sorting in a general way that can easily be re-used in other components.

Although this table is used in two places, so it might require a bit of refactoring and maybe depend on the completion of converting the plagiarism store to redux as well.

khyatisoneji:
@ragesoss I don't have experience working with Redux. If possible, assign it to someone else

ragesoss:
@khyatisoneji thanks! I think Pratush is going to give it a try.

PR #1651, opened 2018-01-29
replace JSON with Oj for parsing

### LINKED ISSUE #1632: Use 'oj' instead of 'json' for Ruby json parsing
`oj` is a faster json parser than the Rails default `json` gem: https://github.com/ohler55/oj

It's also now used by the i18n gem if available, the app uses a lot of i18n, so switching to it might result it a measurable performance improvement.

psinghal20:
Fixes #1632.
Oj is a faster JSON parser than default JSON gem of rails and thus is replacing the default JSON gem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ gem 'jbuilder', '~> 2.0' # DSL for building JSON view template
 gem 'haml-rails' # HTML template language, used instead of ERB
 gem 'bootsnap', require: false # Makes rails boot faster via caching
 gem 'figaro' # easy access to ENV variables. Deprecated.
+gem 'oj'
 
 ### Database and caching
 gem 'mysql2' # MySQL integration for ActiveRecord
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ gem 'jbuilder', '~> 2.0' # DSL for building JSON view template
 gem 'haml-rails' # HTML template language, used instead of ERB
 gem 'bootsnap', require: false # Makes rails boot faster via caching
 gem 'figaro' # easy access to ENV variables. Deprecated.
-gem 'oj'
 
 ### Database and caching
 gem 'mysql2' # MySQL integration for ActiveRecord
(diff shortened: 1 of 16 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'oj'
```
ragesoss: These requires should not be necessary, since it's included in the Gemfile. Unless it has the `require: false` option, each of those gems will be required when rails boots.

ragesoss:
@psinghal20 please have a look at the failing test.

Gemfile
```diff
@@ -7,6 +7,7 @@ gem 'jbuilder', '~> 2.0' # DSL for building JSON view template
 gem 'haml-rails' # HTML template language, used instead of ERB
 gem 'bootsnap', require: false # Makes rails boot faster via caching
 gem 'figaro' # easy access to ENV variables. Deprecated.
+gem 'oj'
```
ragesoss: Please move this the 'http and api tools' section, and add a comment that describes the gem. With a name like `oj`, its role in the app will be opaque otherwise.

Gemfile
```diff
@@ -48,6 +47,7 @@ end
 gem 'faraday' # Standard HTTP library
 gem 'mediawiki_api', '0.7.1' # Library for querying mediawiki API
 gem 'restforce', git: 'https://github.com/ejholmes/restforce.git' # Salesforce API access
+gem 'oj'
```
ragesoss: Needs a comment too, like `# JSON parsing`

vzamanillo:
I am late, but I think this PR could be more abstract setting the `ActiveSupport` JSON encoder gem via an initializer.
```
# config/initializers/json_encoding.rb
module ActiveSupport::JSON::Encoding
  class Oj < JSONGemEncoder
    def encode value
      ::Oj.dump(value.as_json)
    end
  end
end

ActiveSupport.json_encoder = ActiveSupport::JSON::Encoding::Oj
```

PR #1652, opened 2018-01-29
enabled Style/CommentedKeyword

alpha721:
enabled Style/CommentedKeyword


lib/errors/authentication_errors.rb
```diff
@@ -6,24 +6,30 @@ class NotSignedInError < StandardError
       def initialize(msg='Please sign in.')
         super
       end
-    end # NotSignedInError
+    end
+    # NotSignedInError
```
ragesoss: I think instead of moving them, we can do without these comments altogether.

PR #1659, opened 2018-02-01
complete upload.spec.jsx

HartiganHM:
- Complete `upload.spec.jsx` tests for `uploads` component


ragesoss:
@HartiganHM Nice set of tests! We're trying to move away from react-test-utils, and using enzyme instead for new specs. Could you convert it to use enzyme instead?

HartiganHM:
@ragesoss Thanks! Yes, this is absolutely doable. We generally write tests using Enzyme but wanted to stick to the format we found in other tests. I think this should be a fairly quick fix! I'll get to it ASAP.

PR #1662, opened 2018-02-05
New Campaigns Reducer and new campaign button

mauditecandela:
- Converts Campaigns Store to Redux
- Includes a new Select option alphabetically ordered to display all the campaigns when editing a course
- Deletes all code related to campaigns from the Details component and leaves everything at the Campaign Button component
- The select button only offers the campaigns that have not been selected
- Includes specs


app/assets/javascripts/actions/campaign_actions.js
```diff
@@ -84,9 +84,38 @@ export const addCampaign = (courseId, campaignId) => dispatch => {
...
         });
       })
   );
 };
+
+const fetchAllCampaignsPromise = () => {
+  return new Promise((res, rej) => {
+    return $.ajax({
+      type: 'GET',
+      url: `/all_campaigns`,
```
mauditecandela: I would like to know your opinion on this, @ragesoss 

I don't know if I am overcomplicating this issue, but I think we need to also retrieve all the campaigns available so we can add new more. What do you think to add an action that fetches them? Or would you do it differently?

app/assets/javascripts/components/overview/campaign_button.jsx
```diff
@@ -40,26 +44,25 @@ const CampaignButton = createReactClass({
...
     });
 
+    const allCampaigns = ['A', 'B', 'C', 'D'];
     let buttonClass = 'button';
     buttonClass += this.props.inline ? ' border plus' : ' dark';
     const buttonText = this.props.inline ? '+' : CourseUtils.i18n('campaigns', this.props.course.string_prefix);
     const button = <button className={buttonClass} onClick={this.props.open}>{buttonText}</button>;
     return (
       <div className="pop__container" onClick={this.stop}>
+        {campaignList}
```
mauditecandela: I think maybe this makes more sense. I would add the list of current Campaigns + a Remove Button and in the Button underneath, the access to add new campaigns. What do you think of this?
Also, do we need this button or should I create a Select component?
Thanks in advance.

ragesoss: I don't completely understand what you're asking. I think keeping the UI the same during this conversation is a good idea.

mauditecandela:
@ragesoss what I meant was if this design is fine: 
captura de pantalla 2018-02-06 a las 18 02 23
captura de pantalla 2018-02-06 a las 18 02 31
Or if it would be better to include a select option.

app/assets/javascripts/components/overview/campaign_button.jsx
```diff
@@ -29,13 +30,24 @@ const CampaignButton = createReactClass({
   },
 
   removeCampaign(campaignId) {
+    this.addDeletedCampaign(campaignId);
```
mauditecandela: @ragesoss I've been struggling with this as I was trying to perform these changes in the reducer -- without sucess. Right now is partially working: I would need to send the allCampaigns that remain available to the state the app doesn't show all the campaigns in the next interaction. I will research on that tomorrow (suggestions?). Also, would you create these new methods to add and delete campaigns from the state or would you just add this actions to addCampaign/removeCampaign without creating specific methods? What do you think is a best practice?

ragesoss: I would say that you want the redux state to hold two things: the campaigns for this course, and the set of all campaigns. Then, for displaying a list of possible campaigns to add, use a selector to remove the course's campaigns from the full set. If the server returns the list of a course's campaigns when you add or remove, then it can be the same redux action fetching course campaigns initially, adding one, or removing one... the full list of course campaigns comes fresh from the server in each case..

mauditecandela: For me to see if I understood you correctly: Would you add e.g. a second argument to the action (fetchAllCampaigns) with all the course campaigns (e.g. fetchAllCampaigns(campaignId, courseCampaigns)) and then in the success case of the action, compare the two arrays and send as data only the campaigns difference?

mauditecandela: Sorry, it would be "fetchAllCampaigns(courseCampaigns)" as we are not sending the campaignId to fetch campaigns.

app/assets/javascripts/actions/campaign_actions.js
```diff
@@ -91,12 +93,13 @@ export const addCampaign = (courseId, campaignId) => dispatch => {
...
 };
 
-const fetchAllCampaignsPromise = () => {
+const fetchAllCampaignsPromise = (courseCampaigns) => {
   return new Promise((res, rej) => {
     return $.ajax({
       type: 'GET',
       url: `/lookups/campaign.json`,
       success(data) {
+        data.values = _.difference(data.values, courseCampaigns);
```
mauditecandela: something like this? @ragesoss

ragesoss: I was thinking that the action would just pass the values to the reducer, without worrying about the interaction between course campaigns. Then a selector (ie, a reselect function like with the student list) would be used in mapStateToProps to send the unused campaigns to the component that displays the options for adding a new one.

mauditecandela:
@ragesoss I would say this is more or less ready. I need to update the spec though. Could you please have a look and give me some feedback? Thanks.

ragesoss:
@mauditecandela I'm not sure whether I will be able to review it this week, but I will if I can.

mauditecandela:
Great, thanks.

app/assets/javascripts/components/overview/campaign_button.jsx
```diff
@@ -1,25 +1,109 @@
...
+  },
+
+  render() {
+    let campaignList;
+    let campaignSelect;
+    let campaignClass;
+    const campaignOptions = this.props.allCampaigns.map(campaign => {
+      return { label: campaign, value: campaign };
+    });
+    if (this.props.editable) {
```
ragesoss: One possible way to handle this would be to split this into two display components, one for 'edit' mode and one for 'display' mode. Then the 'fetchAllCampaigns', if it hasn't been done yet, could be done when the 'edit' mode version mounts. This nested if/else structure is also likely to be confusing for future readers of this file, and breaking it into simpler display components would make the possible render paths more clear.

app/assets/javascripts/components/course.jsx
```diff
@@ -46,7 +47,8 @@ const Course = createReactClass({
     ServerActions.fetch('course', courseID);
     this.props.fetchUsers(courseID);
     ServerActions.fetch('users', courseID);
-    return ServerActions.fetch('campaigns', courseID);
+    this.props.fetchAllCampaigns();
```
ragesoss: This will fetch all campaigns every time the page is loaded (whether or not the user is even able to add/remove campaigns). Instead of doing that here, I think this should be done closer to when/where it's needed... for example, when the user first focuses the campaign button input.

app/assets/javascripts/reducers/campaigns.js
```diff
@@ -0,0 +1,36 @@
+import {
+  RECEIVE_CAMPAIGNS,
+  RECEIVE_ALL_CAMPAIGNS,
+  ADD_CAMPAIGN,
+  DELETE_CAMPAIGN
+} from '../constants/campaigns.js';
+
+const initialState = {
+  campaigns: [],
+  all_campaigns: [],
```
ragesoss: This should be `allCampaigns` just for style consistency.

app/assets/javascripts/components/overview/campaign_button.jsx
```diff
@@ -1,25 +1,109 @@
...
+      </div>
     );
-  })
-;
+  }
+
+});
+
+const mapStateToProps = state => ({
+  allCampaigns: getAvailableCampaigns(state),
```
ragesoss: Better to call this `availableCampaigns`?

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -357,4 +348,4 @@ const saveCourseDetails = (data, courseId = null) => {
   }
 };
 
-export default Editable(Details, [CourseStore, CampaignStore, TagStore], saveCourseDetails, getState, I18n.t('editable.edit_details'));
```
ragesoss: All that red in this file — lovely!

ragesoss:
I left some detailed comments about a few things I think we should fix before merging this, but overall this is solid.

I think the UI is a little bit of a regression compared to the current design, where clicking the plus button makes the popover appear with an input box and the current campaigns that may be removed show up underneath. I'm okay with merging it anyway like this and filing an issue for fixing the UI later, but if that's something you want to tackle, that would be a fine next task.

PR #1667, opened 2018-02-08
Limiting articles for performance enhancement.

### LINKED ISSUE #1656: Limit course articles.json results for programs with huge numbers
There are some P&E Dashboard programs with huge numbers of articles edited, such as this one: https://outreachdashboard.wmflabs.org/courses/CIS-A2K/Wikidata_India_Workshop

The `articles.json` query takes a very long time, and the app can't render so many articles without problems either.

The easiest solution would be to limit the query to some reasonable value (~500) and to show a message for courses with more than the lmit of articles edited to note that only that many are being shown.

AmitJoki:
Fixes #1656.


ragesoss:
Looks like there are some linting errors.

AmitJoki:
@ragesoss yes! I have not been able to run the tests at cloud9 because the memory keeps running out :( Windows sucks! Trying to run Lubuntu on Virtual Box now.

AmitJoki:
@ragesoss also, currently, I have limited the articles in the backend itself with limit(500) so there's no way to tell if articles were in excess of that limit, but it results in better performance server-side.

Or should the limiting be done in the client-side so we could conditionally display a message when its > 500. As of now, with server-side limiting, the message is always displayed and is generic like - "The first 500 are shown".

What way should I go?

ragesoss:
Server-side is good.

AmitJoki:
What is the eslint configuration that WikiEducationFoundation uses? I am looking for a file `eslintrc.json` so I could use it in my IDE.

AmitJoki:
@ragesoss the tests are passing. Is the code alright?

app/assets/javascripts/components/articles/articles_handler.jsx
```diff
@@ -51,6 +55,12 @@ const ArticlesHandler = createReactClass({
      categories = <CategoryHandler course={this.props.course} current_user={this.props.current_user} />;
    }
 
+    this.props.addNotification({
```
ragesoss: We only want to show a message if the limit has been reached.

Now that I think on it more, I also think it's important to retain a way for users to request the full list when they really want it. I think that could be done by:
* showing a notification only when the limit has been reached
* included a link or button in that notification that would request the full set of articles, with a warning that it might take a while.
* adding an optional `fetch_all` parameter one the server side of things, so that we use the same action for both limited and unlimited requests.

ragesoss:
See my in-line comment. On further thought, I think we need it to be able to fetch the full set of articles if desired.

AmitJoki:
@ragesoss cool. Can I modify the articles.json to also return the count of the articles so as to make a decision of whether to show them a link to fetch all?

AmitJoki:
@ragesoss can you please tell me how could I dispatch newly GET articles so that they'd render in the article list?

Suppose, I get,

```
$.getJSON("articles.json?next=5", function(data){
    console.log(data) // is the newly fetched articles that aren't yet rendered
    // how can I access the Article Store and add the fetched articles so it would render by itself?
});
```

Also please let me know if I am asking too many questions :(

ragesoss:
Have a look at the 'Activity' tab of courses, which rendered with `revision_handler.jsx`. It fetches revisions, with a limit, and then allows the user to click 'show more' to fetch larger amounts.

I think you could do the same thing here.

AmitJoki:
@ragesoss I've been getting Net:Timeout error when logging through Wiki API locally, and the rake tasks aren't doing anything. I ran it in the console and it didn't output anything for a long time.

So, I have not been able to test it visually, so please see if there's anything more I should be doing.

AmitJoki:
@ragesoss up until now, I had issues with my code, and I've cleared them up now, but the build is still failing with regards to Category to  which I've made no changes and I've made sure that the changes I've made don't affect normal operations.

Can you please look into it?

AmitJoki:
@ragesoss this one is finally building alright.

app/assets/javascripts/actions/server_actions.js
```diff
@@ -6,9 +6,9 @@ const Flux = new McFly();
 const ServerActions = Flux.createActions({
 
   // General-purpose
-  fetch(model, courseId) {
+  fetch(model, courseId, args) {
```
ragesoss: Rather than modify this widely-reused general 'fetch' function, we should follow the example of what we did for course revisions, with `fetchRevisions` in `revisions_actions.js`. This might be the right time to convert ArticleStore to redux, which would let you follow the revisions example pretty closely. But even if we're sticking with Flux and ArticleStore for the moment, we can still make a new fetchArticles action in articleActions to replace the use of API.fetch.

AmitJoki: @ragesoss since, the argument is directly passed as data and is the third argument, I thought it wouldn't do any harm. But yes, I'll change that.

app/assets/javascripts/components/articles/articles_handler.jsx
```diff
@@ -19,10 +20,21 @@ const ArticlesHandler = createReactClass({
   },
 
   componentWillMount() {
-    ServerActions.fetch('articles', this.props.course_id);
+    state.limit = 500;
+    state.article_count = 501; // Make sure Show More button is shown at least once
+    // In case the following fails to resolve before rendering
+    $.getJSON(`/courses/${this.props.course_id}/article_count.json`, (data) => {
```
ragesoss: I don't like the idea of adding an extra query for every page load, just to fetch the limit which will only be passed a small portion of the time. I think a better strategy is the one we use for revisions (implemented in the revisions reducer) of comparing the actual number of articles received with the limit, and assuming there are more articles only if the actual number reaches the limit.

AmitJoki: @ragesoss but articles are present within article_list.js I don't suppose I have the this.props.articles on hand in here?

AmitJoki:
@ragesoss wouldn't that conflict with Editable which requires ArticleStore?

ragesoss:
I don't think ArticleList actually needs to be Editable. I'm not sure why we use Editable for it, as there are no edit controls and I don't think we ever call the save function for it.

AmitJoki:
that's what I thought too, so can it be removed then?

ragesoss:
Yes, I think so.

AmitJoki:
there's still one question though. The articles need to be present in Article Handler but it doesn't have access to it as the articles are rendered within ArticleList. What do I do with this?

ragesoss:
Have a look at RevisionsHandler and RevisionList. As part of the switch to redux, we'll get the articles from the redux store and pass them to the component that needs them via mapStateToProps. In this case, we probably can do that in ArticleList. I don't think ArticlesHandler needs the articles.

AmitJoki:
@ragesoss I don't know, the first build is succeeding and the second one is not. Why so?

ragesoss:
@AmitJoki have a look at the failure stack traces at the end of failing build. There some feature specs that are failing because of javascript errors. Looks like you need to import the `API_FAIL` constant.

ragesoss:
Looks like the failing spec is related to the PR.

AmitJoki:
It says "edited view", is there any chance this is because I took Editable away?

ragesoss:
Not directly. The `Articles` tab of a course has three sections, one of which is called "Articles Edited". The failing test is expecting to find multiple entries in that table, but it's finding none. Are the edited articles showing up locally for you?

AmitJoki:
@ragesoss List uses Flux Store for sorting. But I have removed the usage of Editable, which takes away the store component. What should I do?

ragesoss:
@AmitJoki have a look at `RevisionsHandler` and `RevisionList`, which was recently reimplemented with Redux. You can pass a sortBy function to List, in which case it will use that instead of a Flux store.

AmitJoki:
Thanks! Will have a look at it.

psinghal20:
@AmitJoki On a quick look, I would like to suggest some changes:
- You have added an `I18n` string in the en locale. You should all add its documentation in `qqq.yml`. Refer to https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/i18n.md.
- Also, you should remove your changes from the widely-used `fetch` function.
- The `article_actions.js` was actions file for `article_details.js` reducer. In my opinion, you should create a separate actions file for `articles` reducer.
- You have to remove the schema changes.
- One more, though not required, can you look at the `constants/index.js` and `reducers/index.js`. Both are in alphabetical order. Just improves the presentation of code.
About the sorting, I think, you should implement store based sorting using the `sortByKey` method from `model_utils.js`.

AmitJoki:
Closing this in favor of #1755

Issue #1655, opened 2018-01-31
account creation cleanup and polish


ragesoss:
* Use initiateConfirm instead of `confirm` for the confirmation dialog in AvailableActions
* After successfully enabling account requests, the course should be updated (using CourseActions.updateCourse) to set the flag, so that the button disappears
* In StudentList, NewAccountButton should be used instead of NewAccountModal; that component should be in charge of launching the modal and closing it. StudentList should just determine whether to include the button at all and to pass in the necessary props (including the currentUser that gets used by NewAccountModal).
* In NewAccountModal, there should also be a direct link to Wikipedia account creation (the same link that appears if account requests are not enabled), so that it's also possible for users to try to create an account directly instead of doing it through the new flow.

PR #1671, opened 2018-02-09
#1655 Account creation cleanup

mauditecandela:
- Substitutes confirm with initiateConfirm
- Includes new account button instead of new account modal in the student list
- Adds a button that enables the user to create the account directly on wikipedia in the modal
- Adds new label naming for the new account button -- that i think makes more sense for the usability

### DIFF
--- a/app/assets/javascripts/components/enroll/new_account_modal.jsx
+++ b/app/assets/javascripts/components/enroll/new_account_modal.jsx
@@ -36,17 +36,17 @@ const NewAccountModal = ({ course, passcode, currentUser, closeModal, newAccount
   }
 
   let newAccountInfo;
+  let wikipediaAccountCreation;
   if (currentUser.role === INSTRUCTOR_ROLE) {
     newAccountInfo = <div><p>{I18n.t('courses.new_account_info_admin')}</p></div>;
   } else {
     newAccountInfo = <div><p>{I18n.t('courses.new_account_info')}</p></div>;
+    wikipediaAccountCreation = (
+      <a href={`/users/auth/mediawiki_signup?origin=${window.location}`} className="button auth signup border margin">
+        <i className="icon icon-wiki-logo" />{I18n.t('application.sign_up_extended')}
+      </a>
+    );
   }
-
-  const wikipediaAccountCreation = (
-    <a href={`/users/auth/mediawiki_signup?origin=${window.location}`} className="button auth signup border margin">
-      <i className="icon icon-wiki-logo" />{I18n.t('application.sign_up_extended')}
-    </a>
-  );
 
   return (
     <div className="basic-modal left">
--- a/app/assets/javascripts/components/enroll/new_account_modal.jsx
+++ b/app/assets/javascripts/components/enroll/new_account_modal.jsx
@@ -42,6 +42,12 @@ const NewAccountModal = ({ course, passcode, currentUser, closeModal, newAccount
     newAccountInfo = <div><p>{I18n.t('courses.new_account_info')}</p></div>;
   }
 
+  const wikipediaAccountCreation = (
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -100,7 +101,8 @@ const AvailableActions = createReactClass({
     const course = this.state.course;
     const onConfirm = function () {
       enableRequests(course);
-      return alert(I18n.t('courses.accounts_generation_enabled'));
+      alert(I18n.t('courses.accounts_generation_enabled'));
+      return CourseActions.updateCourse(course);
```
mauditecandela: @ragesoss this is not working as I expected. It's not rerendering the component so the button is still appearing until the user refresh the page. Do you have any idea why?

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -84,16 +88,24 @@ const AvailableActions = createReactClass({
...
   enableRequests() {
-    if (confirm('Are you sure you want to enable the account requests?')) {
-      this.props.enableAccountRequests(this.state.course);
-      return alert(I18n.t('courses.accounts_generation_enabled'));
-    }
+    const enableRequests = this.props.enableAccountRequests;
+    const course = this.state.course;
+    const onConfirm = function () {
+      enableRequests(course);
+      alert(I18n.t('courses.accounts_generation_enabled'));
```
ragesoss: Instead of an alert, this would be nicer as a Notification. Part of benefit of switching from 'confirm' to 'initiateConfirm' is that it avoids the slightly jarring UI of the browser-based popup dialogs.

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -84,16 +88,24 @@ const AvailableActions = createReactClass({
...
+      return CourseActions.updateCourse(course);
+    };
+    const confirmMessage = 'Are you sure you want to enable the account requests?';
```
ragesoss: Needs i18n.

spec/features/student_role_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'rails_helper'
 
-describe 'Student users', type: :feature, js: true do
+describe 'Student users', type: :feature, js: true, focus: true do
```
ragesoss: focus: true... whoops!

app/assets/javascripts/components/enroll/new_account_modal.jsx
```diff
@@ -42,6 +42,12 @@ const NewAccountModal = ({ course, passcode, currentUser, closeModal, newAccount
     newAccountInfo = <div><p>{I18n.t('courses.new_account_info')}</p></div>;
   }
 
+  const wikipediaAccountCreation = (
+    <a href={`/users/auth/mediawiki_signup?origin=${window.location}`} className="button auth signup border margin">
+      <i className="icon icon-wiki-logo" />{I18n.t('application.sign_up_extended')}
+    </a>
+  );
+
```
ragesoss: This should only be there if the user if not in the instructor role. I think we also need to change to interface messages to make it clear that clicking this is an alternative to requesting an account: the user can either create their own account directly and then sign in and join the program, or they request an account.

app/assets/javascripts/components/overview/available_actions.jsx
ragesoss: This is also where you could send another action to update the flag, or re-fetch the course from the API, so that the 'enable' button goes away.

ragesoss:
@mauditecandela I'll go ahead and merge this now; please do the UI change for the 'create account yourself' link that we discussed today in a separate commit.

PR #1673, opened 2018-02-11
replaced training_status_store with a reducer

alpha721:
replaced training_status_store with a reducer


ragesoss:
@alpha721 it looks like the failing spec might be related to this change. Can you run it locally and see if it passes, and check whether the 'student drawer' that opens when you click on a user in the Students tab works the same way before and after this change?

alpha721:
@ragesoss, I got held up by my college exams for some days, I will look into it right away. Thanks.

ragesoss:
Looks like this is a bit stale now. Feel free to open a new PR if you take this up again.

PR #1677, opened 2018-02-13
Enable useless assignment cop

arun1595:
Associated with https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1593

### DIFF
--- a/lib/revision_feedback_service.rb
+++ b/lib/revision_feedback_service.rb
@@ -18,7 +18,7 @@ class RevisionFeedbackService
   MINIMUM_REFERENCES = 1
   def citation_feedback
     ref_tags = @features['feature.wikitext.revision.ref_tags']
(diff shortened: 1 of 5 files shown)


lib/revision_feedback_service.rb
```diff
@@ -18,7 +18,7 @@ def feedback
   MINIMUM_REFERENCES = 1
   def citation_feedback
     ref_tags = @features['feature.wikitext.revision.ref_tags']
-    cite_templates = @features['feature.enwiki.revision.cite_templates']
+    # cite_templates = @features['feature.enwiki.revision.cite_templates']
```
arun1595: @ragesoss I have commented this out because I am not sure if the logic to determine citation_feedback needs to do check something against `cite_templates`. If that's not the case, we can remove this line.

PR #1681, opened 2018-02-13
Enable rails output cop and fix violations

arun1595:
Associated with #1593


ragesoss:
conflicts

lib/training/training_base.rb
```diff
@@ -129,7 +129,7 @@ def initialize(content, slug)
       instance_variable_set("@#{key}", value)
     end
   rescue StandardError => e
-    puts "There's a problem with file '#{slug}'"
+    Rails.logger.info "There's a problem with file '#{slug}'"
```
ragesoss: This breaks the spec, which you can probably fix within the failing spec. If it's going through Rails logger, it should be done with level `error`.

ragesoss:
No activity for a while, so I'll close this.

PR #1692, opened 2018-02-14
Adds a link instead of a button to register at Wikipedia

mauditecandela:
We want to give some hierarchy to the new account modal, as we would prefer that users register through our new request account implementation (if available). Therefore we decided to offer the registration at wikipedia just as a button.
This PR includes:
- Implementation of a link that drives the user to the wikipedia account formular
- new locale for translations


ragesoss:
@mauditecandela can you post a screenshot of this?

mauditecandela:
Here you are
captura de pantalla 2018-02-14 a las 16 49 39

Issue #1628, opened 2018-01-22
In Details section on P&E Dashboard, facilitators or admins should be able to change the 'private' setting for a program


ragesoss:

khyatisoneji:
@ragesoss , can you please provide information about which page are you talking about? Also talk briefly about what do you mean by changing private setting.

ragesoss:
There's a `private` attribute on the Course model, which can be set during course creation. If set to true, it makes it so that only the instructors for that course, or admins, can view it.

Users who can edit the course details should be able to change this setting, in the same way that admins can change the `submitted` value while in Edit mode in the Details section of a course Home tab. See `SubmittedSelector` for details on how that was implemented.

ragesoss:
@mauditecandela this would be a good one for you handle. It's already been an issue with several A+F events in the last few weeks.

PR #1694, opened 2018-02-14
#1628 Course Privacy Selector

mauditecandela:
- Adds a new PrivacySelector component
- This PrivacySelector component Includes a new selector to set a course as private or public only viewable by admin users
- Adds locale for translations


ragesoss:
Nice. I think we should only show this in Edit mode. Other than that, looks good to me.

PR #1696, opened 2018-02-14
Converts Campaign Editable into a Popover

mauditecandela:
- Deletes the conditional from Details and includes both campaign list and editable components
- Converts campaign editable into a popover
- Shows campaign editable only in editable mode

### DIFF
--- a/app/assets/javascripts/components/overview/campaign_editable.jsx
+++ b/app/assets/javascripts/components/overview/campaign_editable.jsx
@@ -64,28 +64,22 @@ const CampaignEditable = createReactClass({
       });
       campaignSelect = (
         <Select
-          className="edit"
+          className="fixed-width"
           ref="campaignSelect"
           name="campaign"
-          placeholder="Campaign"
+          placeholder="Select the campaigns you want to add"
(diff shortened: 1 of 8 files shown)


mauditecandela:
@ragesoss could you please check this approach? I would like still to improve how does it look: the popover size may differ a lot depending on the campaign name length and I would prefer to have a fix size. What do you think overall?

app/assets/javascripts/components/overview/campaign_editable.jsx
```diff
@@ -64,30 +64,24 @@ const CampaignEditable = createReactClass({
       });
       campaignSelect = (
         <Select
+          className="fixed-width"
           ref="campaignSelect"
           name="campaign"
-          placeholder="Campaign"
+          placeholder="Select the campaigns you want to add"
```
ragesoss: Needs i18n.

ragesoss:
Almost there!

Looks like there is a styling bug; the campaigns should have class `form-group` rather than `course campaign`. 
campaigns in box

In edit mode, the button should also be at the end of the campaign list rather than the next line, if there is room.
campaign edit mode

PR #1698, opened 2018-02-14
Enable rails environment comparison cop and fix violations

arun1595:
Associated with #1593

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -98,8 +98,6 @@ Lint/UnneededDisable:
   Enabled: false
 Rails/SkipsModelValidations:
   Enabled: false
-Rails/EnvironmentComparison:
-  Enabled: false
 Rails/OutputSafety:
   Enabled: false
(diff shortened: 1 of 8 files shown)


psinghal20:
@arun1595 You cannot compare strings directly like that. The correct syntax will be 
```ruby
Rails.env.eql?('development')
```

arun1595:
@psinghal20 

>  You cannot compare strings directly like that.

`Rails.env.development` does not return a string. It returns a `ActiveSupport::StringInquirer` object. It responds to `test?` and `development?` messages. That's why `Rails.env.development?` works.

You can check it out by firing up the rails console and trying the following:

```
Rails.env
=> "development"
Rails.env.class
=> ActiveSupport::StringInquirer
Rails.env.development?
=> true

If you're interested, have a look at the source code [here](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/string_inquirer.rb)

PR #1702, opened 2018-02-15
Pluralize method names

arun1595:
Associated with #1593


arun1595:
@ragesoss Done :smile:

PR #1706, opened 2018-02-15
Enable predicate name cop

maryam21:
#1593

### DIFF
--- a/app/helpers/surveys_helper.rb
+++ b/app/helpers/surveys_helper.rb
@@ -96,7 +96,7 @@ module SurveysHelper
     return question.conditionals
   end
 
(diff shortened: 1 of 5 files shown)


app/helpers/surveys_helper.rb
```diff
@@ -96,7 +96,7 @@ def question_conditional_string(question)
     return question.conditionals
   end
 
-  def has_course_data(question_form)
+  def course_data(question_form)
```
ragesoss: This should be `course_data?`.

maryam21: That makes more sense I will change it.

PR #1708, opened 2018-02-15
Migrate away from bower

### LINKED ISSUE #1453: Migrate away from bower

psinghal20:
Fixes #1453.
Bower has been deprecated and even its developers advise to use yarn or npm. This shifts away from bower by adding bower dependencies in package.json and using these for public assets. It uses copy-webpack-plugin and gulp concat plugin to copy the previous bower modules from their npm module to public directory.


gulp/tasks/webpack.js
```diff
@@ -3,11 +3,14 @@ import loadPlugins from 'gulp-load-plugins';
 import ManifestPlugin from 'webpack-manifest-plugin';
 import webpack from 'webpack';
 import WebpackDevServer from 'webpack-dev-server';
+import CopyWebpackPlugin from 'copy-webpack-plugin';
 import path from 'path';
 import config from '../config.js';
+import Assets from './assets';
```
ragesoss: This should have a more precise name... these are global assets that are expected to be loaded before main.js, so maybe 'GlobalAssets', or maybe 'UniversalLanguageSelectorDependencies' since that's all that's left that we need these globals for?

ragesoss:
This looks good! The PR should also remove references to bower in the docs.

psinghal20:
@ragesoss made the required changes in docs, .travis.yml, package.json, yarn.lock, watch.js and changed the name of assets.

PR #1718, opened 2018-02-18
Replaced jquery-ujs with rails-ujs

dianameca:


ragesoss:
Looks like this is not working. The key thing that this does is allow non-get requests from html links; that seems to be why the specs are failing, because that functionality isn't working.

PR #1719, opened 2018-02-18
Good article nomination

### LINKED ISSUE #1669: Dashboard should create alerts for Good Article Nominations
Similar to the DYK alerts, the Dashboard should generate alerts and send emails when relevant articles get nominated for Good Article on English Wikipedia.

It'll probably look a lot like `DYKNominationMonitor`

AmitJoki:
Fixes #1669


ragesoss:
This task shouldn't require any frontend changes; similarly to ArticlesForDeletionMonitor, it's just a backend alert.

AmitJoki:
@ragesoss ah! FIne. But there's another issue. category.spec is failing everytime for me though I am pretty sure I've not touched anything relating to it!

AmitJoki:
category.spec is failing for my another PR too with limiting articles.json 

What's happening?

ragesoss:
It looks like the Wikipedia category that is being tested against has changed, so that spec will need to be updated. I'll do that probably on Tuesday.

AmitJoki:
Okay. I'll probably still continue to submit PRs and wait for Tuesday for the build to run successfully.

ragesoss:
Just fixed it, actually.

AmitJoki:
ah that was fast! Thanks.

lib/alerts/ga_nomination_monitor.rb
```diff
@@ -0,0 +1,69 @@
...
+    course_articles = ArticlesCourses.joins(:article)
+                                     .where(articles: { title: @page_titles, wiki_id: @wiki.id })
+    course_articles.each do |articles_course|
+      create_alert(articles_course)
+    end
+  end
+
+  private
+
+  GA_CATEGORY = 'Wikipedia:Good Article nominations'
```
ragesoss: This should be "Category:Good article nominees"

lib/alerts/ga_nomination_monitor.rb
```diff
@@ -0,0 +1,69 @@
...
+  GA_CATEGORY_DEPTH = 0
+  def find_pending_ga_nominations
+    @ga_titles = CategoryImporter.new(@wiki)
+                                 .page_titles_for_category(GA_CATEGORY, GA_CATEGORY_DEPTH)
+  end
+
+  def extract_page_titles_from_nominations
+    @ga_article_titles = @ga_titles.map do |ga_title|
+      ga_title[%r{Template:Good Article nominations/(.*)}, 1]
```
ragesoss: This will need to adjusted as well.

AmitJoki:
@ragesoss as I have said previously, running rake batch:update_constantly and update_daily are doing nothing. They get stuck and I've to use Ctrl+C to terminate it.

ragesoss:
@AmitJoki let's try to figure out the problem with running rake tasks locally, so that you can test this out. What exactly happens when you run `rake batch:update_constantly`? You could also run this routine from the rails console.

AmitJoki:
@ragesoss the command just goes blank and outputs nothing. update_daily worked though. update constantly is just getting stuck and have to keyboard interrupt it.

Is it supposed to show some feed back or is that how it works?

ragesoss:
Try requiring and then running `ConstantUpdate.new` in a rails console. That way, you should see the debug output... lots of queries and updates. It could be that you have enough high-volume editors in your courses that it just takes a long time. In that case, you could use `/manual_update` on a single course to pull in data for just that one.

PR #1720, opened 2018-02-20
Request account campaign

mauditecandela:
- Creates a new controller that handles the requests from the campaigns.
- The user is able to enable and disable the account requests from the campaign
- Creates a new field in the campaigns table that registers if the account creation for the campaign is enabled
- Includes specs
- Includes locales
- Modifies the javascript tests to pass again


mauditecandela:
@ragesoss I create this PR so it's easier for me to talk and ask my questions. I am not sure about the approach on how to do this. I went for a rails form linked from the requested accounts campaign index. I would like to know if this would be the right approach.

db/migrate/20180219125251_add_campaign_id_to_requested_accounts.rb
```diff
@@ -0,0 +1,5 @@
+class AddCampaignIdToRequestedAccounts < ActiveRecord::Migration[5.1]
```
ragesoss: Instead of this, you can just use a scope: Campaign has_many RequestedAccounts through Courses.

ragesoss:
We can talk it through further, but on a quick skim this seems like a good way to do it. I think we should not add campaign_id to the model; we can go through associations to find all the requests for a campaign without changing the database, and that will be easier to reason about, since we won't need to worry about which campaign to put in the record for requests from a course with multiple campaigns.

config/routes.rb
```diff
@@ -46,6 +46,18 @@
   post 'mass_enrollment/:course_id'  => 'mass_enrollment#add_users',
       constraints: { course_id: /.*/ }
 
+  put 'requested_accounts_campaigns' => 'requested_accounts_campaigns#request_account'
+  delete '/requested_accounts_campaigns/*campaign_slug/*id/delete' => 'requested_accounts_campaigns#destroy',
```
ragesoss: These two routes are no longer used, right?

PR #1722, opened 2018-02-21
Migrate Validation store to redux [WIP]

psinghal20:
It migrates validation store to the redux store and converts the input mixin to a High-order component. It extracts out the utility functions associated with ValidationStore to validation utils.

### DIFF
--- a/app/assets/javascripts/actions/course_actions.js
+++ b/app/assets/javascripts/actions/course_actions.js
@@ -46,7 +46,7 @@ const CourseActions = Flux.createActions({
     };
   },
 
-  updateClonedCourse(data, courseId, tempId) {
+  updateClonedCourse(data, courseId, tempId, checkServer, updatePersistingStatus) {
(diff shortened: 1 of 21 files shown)


psinghal20:
@ragesoss Can you take a look at this PR when you have time? I have been working on this PR for long. I have separated the changes to InputMixin from this PR. A lot of the Jest tests are failing due to conversion to redux. I also tried to run the some of the feature specs which used to access the validation store earlier, but I am facing a MySQL Duplicate entry and Deadlock errors. As many of the jest tests are failing, even the Travis build log is not helping with the feature specs which could help in verifying the behavior.

psinghal20:
Hi @ragesoss, Can you take a look at the failing Travis Build log? I was able to fix the earlier errors but right now the Travis build Failure shows `Ambiguous match` error which should not be the case there is only one next button and there is no change in the UI.

PR #1724, opened 2018-02-22
add sorting on redux state for recent_edits via sortByKeys

### LINKED ISSUE #1638: Recent Edits sorting is wonky

psinghal20:
Fixes #1638. Adds sorting on the redux state for recent_edits via using sortByKeys from model_utils.


psinghal20:
@ragesoss The feature specs for language switcher seems to fail in Travis Build and I don't think it is related to this and must be random failing spec.

ragesoss:
Yes, that one is flaky. I'll restart the build.

ragesoss:
I'm trying this out, and sorting by date still is weird; I have to click twice before it will sort by date. Also, there are no icons in the header to indicate which columns are sortable and which are not.

app/assets/javascripts/components/activity/activity_table.jsx
```diff
@@ -121,7 +108,7 @@ const ActivityTable = createReactClass({
   _renderHeaders() {
     return this.props.headers.map((header) => {
       return (
-        <th style={header.style || {}} key={header.key} onClick={this.sortItems} className="sortable" data-sort-key={header.key}>
+        <th style={header.style || {}} key={header.key} onClick={this.sortItems} className="sortable asc" data-sort-key={header.key}>
```
ragesoss: Ideally, this should be done the same way as the `/explore` table: the class changes according to the direction it's being sorted, so the sorting indicator shows which way that column is being sorted. I'll open a separate issue for this.

psinghal20: @ragesoss Okay, I understand. I can also work on that issue if you want

PR #1726, opened 2018-02-22
Migrate from bower

### LINKED ISSUE #1453: Migrate away from bower

psinghal20:
Fixes #1453.
Bower has been deprecated and even its developers advise to use yarn or npm. This shifts away from bower by adding bower dependencies in package.json and using these for public assets. It uses copy-webpack-plugin and gulp concat plugin to copy the previous bower modules from their npm module to public directory.


package.json
```diff
@@ -140,7 +142,8 @@
     "test": "jest",
     "start": "gulp dev",
     "hot": "gulp hot-dev",
-    "build": "gulp build"
+    "build": "gulp build",
+    "postinstall": "node -e \"try { require('fs').symlinkSync(require('path').resolve('node_modules/@bower_components'), 'vendor', 'junction') } catch (e) { }\""
```
ragesoss: Is this necessary? Looks like this is doing symlinks to vendor, but where are those symlinks used?

package.json
```diff
@@ -153,5 +156,8 @@
       "app/assets/javascripts",
       "node_modules"
     ]
+  },
+  "engines": {
+    "yarn": ">= 1.0.0"
```
ragesoss: What does this do?

psinghal20: This line was added by the migrate away package. It actually just specifies what our system runs that is basically what package manager we use with its version info. It will generate a warning when using with npm install as we have specified yarn as our engine. 
https://docs.npmjs.com/files/package.json#engines

package.json
psinghal20: This line was also added by the bower-away package to provide symlinks for bower_components as the vendor directory. So that files which are using vendor directory can now use this symlink to follow to the bower components. In our case we don't need so I will remove this command from the package.json

gulp/tasks/jquery.uls.js
```diff
@@ -0,0 +1,24 @@
...
+import loadPlugins from 'gulp-load-plugins';
+import config from '../config.js';
+const plugins = loadPlugins();
+
+
+//--------------------------------------------------------
+// Concatenate jquery-uls libraries
+//--------------------------------------------------------
+const jqueryUlsPath = [
+  'node_modules/@bower_components/jquery/dist/jquery.js',
```
ragesoss: Could we just as easily remove the `@bower_components/` prefix both here and in package.json, and this would still work?

psinghal20: No, earlier I did the same thing and tried to use the same jquery packages here and that's what to lead the survey errors. I tried it with bower-away and seemed to work. If we use the same package here it leads to jquery declaration again and leads to those errors though I am not sure.

ragesoss: I see. So maybe we can do that separately later on via getting rid of the webpack `externals` for jquery and explicitly requiring jquery everywhere it's needed.

psinghal20: Yes, we can try that later though I haven't looked much into it. Will try to take a look and see what alternatives we have so that we don't need to manage separate modules for these bower modules.

ragesoss: I think a good strategy will be to modify `views/shared/_head.html.haml` to only include the jquery-uls.js if `Features.enable_language_switcher?`, then fix all the ensuing chaos. ;)

psinghal20: @ragesoss I will try getting rid of the jquery from externals and bundling it with webpack as well and then add the check for language switcher and try to look into the issues from surveys. I will update you about the status.

PR #1733, opened 2018-02-24
add post methods for replica service to get articles by titles

### LINKED ISSUE #1575: Replica article queries should be broken into multiple queries when the URI is too long
`Replica` is the class for fetching data from copies of the Wikimedia databases. It's a PHP service that connects to the wmflabs replica databases: https://github.com/WikiEducationFoundation/WikiEduDashboardTools

One failure mode it has is querying for lots of article titles, which can result in a URI that is too long and causes an error. This mainly happens with queries for languages with non-latin scripts, where the URI encoding of article titles makes for much longer URIs. We currently work around this by special-casing the languages that have caused errors, and fetching fewer articles per query for those languages. See `ArticleStatusManager#articles_per_replica_query`.

A better approach would be to update `Replica` to handle breaking up large queries into smaller ones when necessary, based on URI length.

Another option, which might be even better, would be to use POST instead of GET requests and send the article titles as data rather than part of the URI. That would probably require making changes to the wmflabs replica endpoint tool.

psinghal20:
Fixes #1575. This adds post methods in replica.rb to access the replica service with post methods and uses the said function in article_status_manager.rb to retrieve articles by titles without breaking the data into chunks.

### DIFF
--- a/lib/article_status_manager.rb
+++ b/lib/article_status_manager.rb
@@ -124,18 +124,6 @@ class ArticleStatusManager
     true
   end
 
-  # This is limited by the URI length of the combined titles. For most languages,
-  # 100 titles per query is no problem, but languages with unicode titles hit the
-  # URI length limit.
-  # TODO: move the chunking to Replica and set the size dynamically depending on the
-  # length of the URI.
-  LONG_URI_LANGUAGES = %w[he ar ml mk].freeze
-  HIGH_REPLICA_LIMIT = 80
-  LOW_REPLICA_LIMIT = 20
-  def articles_per_replica_query
-    LONG_URI_LANGUAGES.include?(@wiki.language) ? LOW_REPLICA_LIMIT : HIGH_REPLICA_LIMIT
-  end
-
   # Check whether any deleted pages still exist with a different article_id.
   # If so, update the Article to use the new id.
   def update_article_ids(deleted_page_ids)
--- a/lib/article_status_manager.rb
+++ b/lib/article_status_manager.rb
@@ -140,16 +140,13 @@ class ArticleStatusManager
   # If so, update the Article to use the new id.
   def update_article_ids(deleted_page_ids)
     maybe_deleted = Article.where(mw_page_id: deleted_page_ids, wiki_id: @wiki.id)
-
     # These pages have titles that match Articles in our DB with deleted ids
-    same_title_pages = Utils.chunk_requests(maybe_deleted, articles_per_replica_query) do |block|
-      request_results = Replica.new(@wiki).get_existing_articles_by_title block
-      @failed_request_count += 1 if request_results.nil?
-      request_results
-    end
+
+    request_results = Replica.new(@wiki).post_existing_articles_by_title maybe_deleted
+    @failed_request_count += 1 if request_results.nil?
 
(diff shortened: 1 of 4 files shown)


ragesoss:
The wmflabs.org endpoint is updated with your patch to handle POST now. Looks like this is still failing with the new endpoint.

psinghal20:
Sorry, Forgot to change the variable used in functions. Made the correction and pushed the code.

ragesoss:
The ArticleStatusManager specs fail locally for me.

psinghal20:
@ragesoss Can you help me with the failing specs? I tried to look through them and found that for all tests that show `each method error` ie the 1,2,3 and 5 failing specs are caused because the article_list received is empty and thus not getting sent with the post request. For the 4th failing spec, I am not sure what is causing it as I have checked the response of the PHP endpoint with both get and post methods with same parameters is same.

ragesoss:
I think I found the issues here...

First, Utils.chunk_requests will return `[]` if all the requests fail. Now that you're just doing one request and not wrapping it in chunk_requests, you just need to handle the case of a failed query in `ArticleStatusManager#update_article_ids`, for exampe by returning before the request_results.each step, or using `request_results&.each`.

Second, the `does not delete articles by mistake` spec relies on stubbing `:get_existing_articles_by_title`. The name of that method isn't referring to the GET http method, it's more like just 'fetch'. I suggest that, instead of adding the new `post_existing_articles_by_title` method, you just change the implementation of the `get` method to use post. That will mean that the existing replica specs are now exercising the `post` path, as well, so we won't need to worry about writing new specs for that.

psinghal20:
@ragesoss Thank you! I have made the required changes and removed the get method and replace it with the post method. The specs now pass locally for me.

psinghal20:
Hi @ragesoss, The first build passes and I think the second build was a random Travis failure. Can you restart the build?

lib/replica.rb
```diff
@@ -128,11 +128,35 @@ def api_get(endpoint, query='')
...
   end
 
+  def api_post(endpoint, key, data)
+    tries ||= 3
+    response = do_post(endpoint, key, data)
+    return unless response.is_a? Net::HTTPSuccess
+    parsed = Oj.load(response.body)
+    return unless parsed['success']
+    parsed['data']
+  rescue StandardError => e
```
ragesoss: Needs tests for this error handling.

lib/article_status_manager.rb
```diff
@@ -140,16 +140,13 @@ def articles_per_replica_query
   # If so, update the Article to use the new id.
   def update_article_ids(deleted_page_ids)
     maybe_deleted = Article.where(mw_page_id: deleted_page_ids, wiki_id: @wiki.id)
-
     # These pages have titles that match Articles in our DB with deleted ids
-    same_title_pages = Utils.chunk_requests(maybe_deleted, articles_per_replica_query) do |block|
```
ragesoss: The `articles_per_replica_query` method above is no longer needed.

psinghal20:
@ragesoss I have added the test for the error handling in replica_specs. Please review the changes. Also, The Travis build seems to be a random failure.

PR #1739, opened 2018-02-26
UNFINISHED Faster Update for Editathons

mauditecandela:
- Creates a new scope for courses that need update + are editathons
- Includes a new update called short to run more frequently on editathons
- This update cycle is based on  constant update but omits some of their steps or reduces its scope
- Adds a new method to the batch to run this update
- Adds a PID file to batch update logging

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -157,7 +157,7 @@ class Course < ApplicationRecord
   scope :ready_for_update, -> { current.or(where(needs_update: true)) }
 
   scope :ready_for_short_update, lambda {
-    strictly_current.where(type: 'Editathon')
+    strictly_current.where('end <= ?', 1.day.from_now)
   }
 
   def self.will_be_ready_for_survey(opts)
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -156,6 +156,10 @@ class Course < ApplicationRecord
 
   scope :ready_for_update, -> { current.or(where(needs_update: true)) }
 
+  scope :ready_for_short_update, lambda {
+    where(type: 'Editathon').where(needs_update: true)
+  }
+
   def self.will_be_ready_for_survey(opts)
     days_offset, before, relative_to = opts.values_at(:days, :before, :relative_to)
     today = Time.zone.now
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -157,7 +157,7 @@ class Course < ApplicationRecord
   scope :ready_for_update, -> { current.or(where(needs_update: true)) }
 
   scope :ready_for_short_update, lambda {
-    where(type: 'Editathon').where(needs_update: true)
+    strictly_current.where(type: 'Editathon')
   }
 
   def self.will_be_ready_for_survey(opts)
--- a/lib/data_cycle/batch_update_logging.rb
+++ b/lib/data_cycle/batch_update_logging.rb
@@ -6,6 +6,7 @@ module BatchUpdateLogging
   include ActionView::Helpers::DateHelper
 
   UPDATE_PID_FILES = {
+    short: 'tmp/batch_update_shortly.pid',
     daily: 'tmp/batch_update_daily.pid',
     constant: 'tmp/batch_update_constantly.pid',
     views: 'tmp/batch_update_views.pid',
--- a/lib/data_cycle/constant_update.rb
+++ b/lib/data_cycle/constant_update.rb
@@ -113,7 +113,6 @@ class ConstantUpdate
   end
 
   def conflicting_updates_running?
-    return true if update_running?(:short)
     return true if update_running?(:daily)
     return true if update_running?(:constant)
     return true if update_waiting_to_run?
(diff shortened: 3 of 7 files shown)


mauditecandela:
@ragesoss I will add some comments to this PR to explain my thoughts regarding the new short update

app/models/course.rb
```diff
@@ -156,6 +156,10 @@ def self.unsubmitted
 
   scope :ready_for_update, -> { current.or(where(needs_update: true)) }
 
+  scope :ready_for_short_update, lambda {
```
mauditecandela: I created a new scope, and in this case I just included editathons, set as needs_update true. As the editathons normally last less than one month, current was (as in ready_for_update) is not needed.

lib/data_cycle/constant_update.rb
```diff
@@ -113,6 +113,7 @@ def remove_needs_update_flags
   end
 
   def conflicting_updates_running?
+    return true if update_running?(:short)
```
mauditecandela: Added this here in order to prevent conflicts. I set the short one as the first one as is expected to be the shortest one.

lib/data_cycle/short_update.rb
```diff
@@ -0,0 +1,112 @@
...
+    return if updates_paused?
+    return if conflicting_updates_running?
+
+    run_update_with_pid_files(:constant)
+  end
+
+  private
+
+  def set_courses_to_update
+    @courses = Course.ready_for_short_update.to_a
```
mauditecandela: here I used the new scope

lib/data_cycle/short_update.rb
```diff
@@ -0,0 +1,112 @@
...
+    log_message "Ready to update #{@courses.count} courses"
+  end
+
+  def run_update
+    log_start_of_update 'Short update tasks are beginning.'
+    update_revisions_and_articles
+    update_new_article_views unless ENV['no_views'] == 'true'
+    update_new_article_ratings
+    import_uploads_for_needs_update_editathons
```
mauditecandela: Do we need the uploads and to format the categories for this update? Normally this will also happen in the daily update, so it's not that relevant. Anyway, I reduced also the scope of courses that should be updated in the methods definitions.

lib/data_cycle/short_update.rb
```diff
@@ -0,0 +1,112 @@
...
+    update_categories_for_needs_update_editathons
```
mauditecandela: Same question as in `import_uploads_for_needs_update_editathons`

lib/data_cycle/short_update.rb
```diff
@@ -0,0 +1,112 @@
...
+    update_all_caches # from CacheUpdater
+    generate_alerts # from UpdateCycleAlertGenerator
```
mauditecandela: In this case we don't remove the `needs_update`flag so the bigger update can take place later.

mauditecandela: Also, I don't think is needed to update the ungreeted students, as in editathons this is not relevant.

app/models/course.rb
```diff
@@ -156,6 +156,10 @@ def self.unsubmitted
...
+    where(type: 'Editathon').where(needs_update: true)
```
ragesoss: The `needs_update` attribute is only ever set manually for an individual course (by clicking the 'update stats' button that appears after the event has ended), and it gets reset to false after the next time it gets updated.

In this case, I think we could something like `strictly_current.where(type: 'Editathon')`. That would only include editathon events while they are going on. Alternatively, instead of basing it on type, you could write a scope based on the program being short or nearly over: strictly_current ending in less than a day from now, for example.

lib/data_cycle/constant_update.rb
ragesoss: If we can't run the short update concurrently with other updates, then splitting off a separate process probably won't help much; we'll still need to stop short updates for ~5 hours every time constant update runs, and longer for daily update.

I think instead we should fix any conflicts such that we can be running short updates on a continual basis, without worrying about what other updates are running.

lib/data_cycle/batch_update_logging.rb
```diff
@@ -6,6 +6,7 @@ module BatchUpdateLogging
   include ActionView::Helpers::DateHelper
 
   UPDATE_PID_FILES = {
+    short: 'tmp/batch_update_shortly.pid',
```
ragesoss: Calling it `update_shortly` is kinda confusing, but kinda amusing.

lib/data_cycle/short_update.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  def run_update
+    log_start_of_update 'Short update tasks are beginning.'
+    update_revisions_and_articles
+    update_new_article_views unless ENV['no_views'] == 'true'
+    update_new_article_ratings
+    import_uploads_for_needs_update_editathons
+    update_categories_for_needs_update_editathons
+    update_all_caches # from CacheUpdater
```
ragesoss: This is actually the most time-consuming part of the update cycle, and will update all `current` courses, I think.

lib/data_cycle/short_update.rb
```diff
@@ -0,0 +1,49 @@
+# frozen_string_literal: true
+
+require "#{Rails.root}/lib/data_cycle/batch_update_logging"
```
ragesoss: Probably need to require UpdateCourseRevisions.

PR #1740, opened 2018-02-26
[WIP] Heard from

### LINKED ISSUE #1529: For Wiki Ed ask new instructors how they heard about Wiki Ed
This could probably be done in the /onboarding flow. Maybe create an Alert and email someone each time it happens.

AmitJoki:
Fixes #1529 

Needs another commit for sending an alert.


ragesoss:
I tried this out, and the functionality works smoothly. I haven't reviewed the code in depth yet, but a few things:

* Remove the schema.rb changes
* "Heard" and "HeardFrom", etc, are names that are likely to cause confusion. I think the alert could be `OnboardingAlert`, and that will still make sense if we later add additional questions or change them. The more general concept here is collecting supplemental onboarding info, so something like 'onboarding supplement' might work.
* This needs a feature spec for going through onboarding as an instructor, and confirming that the alert gets created after the form is filled out.
* The form starts out with a red 'invalid' border; it should not. Leaving it blank should be fine, in which case it should not create an alert.
* The messages should use I18n.

AmitJoki:
will submit a fresh PR tomorrow incorporating these changes.

AmitJoki:
Also, could you clarify the last point a bit more? I haven't seen I18n being used in `form.jsx` What do you mean by messages?

ragesoss:
I mean interface messages. Although most of the other onboarding components have the text hard-coded into the jsx files, there's no reason to skip i18n for new components. You can find plenty of other components that use `I18n.t()` instead of directly including message text. The message text lives in config/locales/en.yml, which makes it available for people to translate on translatewiki.net and eventually populate the other yml locale files.

AmitJoki:
closing this in favour of  #1747

PR #1745, opened 2018-03-01
Allow home_wiki of a course to be updated

### LINKED ISSUE #1741: Home wiki for a course should be changeable
Currently, the home wiki for a course is set at the time of course creation, and cannot be changed. It would be useful to allow instructors/facilitators to change this.

psinghal20:
Fixes #1741.
It allows the home_wiki of a course to changed by the user. Provides `HomeWikiLanguageSelector` and `HomeWikiProjectSelector` components in editable course details component. Further, modifies `course_controller#update` to allow update of home_wiki_id for a course.

### DIFF
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -17,6 +17,8 @@ import PrivacySelector from './privacy_selector.jsx';
 import WithdrawnSelector from './withdrawn_selector.jsx';
 import TimelineToggle from './timeline_toggle.jsx';
 import CourseLevelSelector from '../course_creator/course_level_selector.jsx';
+import HomeWikiProjectSelector from './home_wiki_project_selector.jsx';
+import HomeWikiLanguageSelector from './home_wiki_language_selector.jsx';
 
 import Editable from '../high_order/editable.jsx';
 import TextInput from '../common/text_input.jsx';
@@ -222,6 +224,8 @@ const Details = createReactClass({
     let courseLevelSelector;
     let timelineToggle;
     let withdrawnSelector;
+    let projectSelector;
+    let languageSelector;
     if (this.props.current_user.admin) {
       const tagsList = this.props.tags.length > 0 ?
         _.map(this.props.tags, 'tag').join(', ')
@@ -295,6 +299,22 @@ const Details = createReactClass({
       );
     }
 
+    if (this.props.editable) {
+      projectSelector = (
+        <HomeWikiProjectSelector
+          course={this.props.course}
+        />
+      );
+    }
+
+    if (this.props.editable) {
+      languageSelector = (
+        <HomeWikiLanguageSelector
+          course={this.props.course}
+        />
+      );
+    }
+
     return (
       <div className="module course-details">
         <div className="section-header">
@@ -347,6 +367,8 @@ const Details = createReactClass({
           {privacySelector}
           {timelineToggle}
           {withdrawnSelector}
+          {projectSelector}
+          {languageSelector}
         </div>
       </div>
     );
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -299,15 +299,12 @@ const Details = createReactClass({
       );
     }
 
-    if (this.props.editable) {
+    if (this.props.editable && !Features.wikiEd) {
       projectSelector = (
         <HomeWikiProjectSelector
           course={this.props.course}
         />
       );
-    }
-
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/components/overview/details.jsx
```diff
@@ -295,6 +299,22 @@ const Details = createReactClass({
...
 
+    if (this.props.editable) {
+      projectSelector = (
+        <HomeWikiProjectSelector
+          course={this.props.course}
+        />
+      );
+    }
+
+    if (this.props.editable) {
```
ragesoss: These go together, so use the same `if` statement for both.

Also, these should not appear on the Wiki Ed dashboard, so add `!Features.wikiEd` to the conditional.

app/assets/javascripts/components/overview/home_wiki_project_selector.jsx
```diff
@@ -0,0 +1,52 @@
...
+  },
+
+  render() {
+    const options = JSON.parse(WikiProjects).map((project, index) => {
+      return (<option value={project} key={index}>{project}</option>);
+    });
+
+    const selector = (
+      <div className="form-group">
+        <label htmlFor={this.state.id}>Home Wiki Project:</label>
```
ragesoss: Needs i18n.

app/controllers/courses_controller.rb
```diff
@@ -41,7 +41,16 @@ def update
     validate
     handle_course_announcement(@course.instructors.first)
     slug_from_params if should_set_slug?
-    @course.update course_params
+
+    if params[:course].key?(:home_wiki)
```
ragesoss: This new data handling should be done in a helper method, like with the timeline and slug routines.

psinghal20:
@ragesoss Can you help me with the feature spec for home_wiki? I tried adding the allow statement in the existing before statement for the test, it broke all the specs. Tried it with the describe statement in the editing examples but still wasn't working. Finally added the allow statement in the particular example without before statement. I haven't been able to test because of some local MySQL errors in the spec, which are related to some duplicate entries in my local Database. I pushed the code to see if the Travis build passes but it seems to fail because of some error in the I18n gulp task. Can you please take a look at the feature spec?

ragesoss:
Whoops. You hit travis while master was broken, before I did the actual fix for those gulp warnings.

spec/features/course_page_spec.rb
```diff
@@ -224,6 +226,20 @@ def js_visit(path, count=3)
       end
       expect(Course.last.passcode).to eq(previous_passcode)
     end
+
+    it 'allows edit in home_wiki_id' do
+      admin = create(:admin, id: User.last.id + 1)
+      login_as(admin)
+      allow(Features).to receive(:wiki_ed?).and_return(false)
```
ragesoss: This might work. The typical way to do it would be to put this example inside a `context`. See feature_toggles_spec for a simple example of this.

psinghal20:
Yes, I didn't know about the I18n fix. Can you restart the build?

spec/features/course_page_spec.rb
```diff
@@ -224,6 +192,24 @@ def js_visit(path, count=3)
       end
       expect(Course.last.passcode).to eq(previous_passcode)
     end
+
+    context 'allows edit in home_wiki_id' do
```
ragesoss: The `context` text should refer to the setup being done in this block, which in this case is the 'before'.

Then the `it` block text should refer to what you're actually testing.

And together, they should form a complete-ish, natural language description of the test.

spec/features/course_page_spec.rb
```diff
@@ -224,6 +192,24 @@ def js_visit(path, count=3)
...
+      before { allow(Features).to receive(:wiki_ed?).and_return(false) }
+      it do
+        es_wikibooks = Wiki.get_or_create language: 'es', project: 'wikibooks'
```
ragesoss: This is okay, but could be done with a `let` instead.

spec/features/course_page_spec.rb
```diff
@@ -43,6 +45,7 @@ def js_visit(path, count=3)
                     timeline_end: course_end.to_date,
                     school: 'This university.foo',
                     term: 'term 2015',
+                    home_wiki_id: es_wiktionary.id,
```
ragesoss: This is what is changing the behavior of the later test, and making it no longer test what it was designed for. That test is supposed to be for how article titles show up when they are a different home wiki than the course. Previously, the home wiki relied on the default of en.wikipedia.

I suggest making it explicitly en.wikipedia here, so that no changes to other tests will be needed.

PR #1746, opened 2018-03-02
Campaigns Alerts 

mauditecandela:
- Creates a new alerts handler to include the alerts by campaign (not finished)
- Creates a new method in the campaigns controller that returns an array of alerts ids
- Adds the new alerts handler to the react router
- Includes locales

### DIFF
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -67,10 +67,10 @@ class CampaignsController < ApplicationController
   def alerts
     alerts = []
     @campaign.course_ids.each do |course_id|
-      alert = Alert.find_by(course_id: course_id)
-      alerts << alert.id unless alert.nil?
+      alert = Alert.where(course_id: course_id)
+      alert.each { |alert| alerts << alert.id unless alert.nil? }
     end
-    alerts
+    render json: { campaign_alerts: alerts }
   end
 
   def edit
--- a/app/controllers/campaigns_controller.rb
+++ b/app/controllers/campaigns_controller.rb
@@ -8,7 +8,7 @@ class CampaignsController < ApplicationController
   layout 'admin', only: %i[index create]
   before_action :set_campaign, only: %i[overview programs articles users edit
                                         update destroy add_organizer remove_organizer
-                                        remove_course courses ores_plot articles_csv]
+                                        remove_course courses ores_plot articles_csv alerts]
   before_action :require_create_permissions, only: [:create]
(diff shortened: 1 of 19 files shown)


app/controllers/campaigns_controller.rb
```diff
@@ -64,6 +64,15 @@ def users
     ).includes(:user, :course).order(revision_count: :desc)
   end
 
+  def alerts
```
mauditecandela: @ragesoss I would like to have your opinion here. I created a new method in the campaigns controller that returns an array with alerts method. This way we'll be able to call sth like "this.props.campaign.alerts" in the alerts_handler (react). What do you think of this approach?

Thanks in advance!

ragesoss: This makes sense, but you want to be able to do something like @campaign.alerts, which should come strictly from scopes.

app/controllers/campaigns_controller.rb
```diff
@@ -67,10 +67,10 @@ def users
   def alerts
     alerts = []
     @campaign.course_ids.each do |course_id|
-      alert = Alert.find_by(course_id: course_id)
-      alerts << alert.id unless alert.nil?
+      alert = Alert.where(course_id: course_id)
+      alert.each { |alert| alerts << alert.id unless alert.nil? }
     end
-    alerts
+    render json: { campaign_alerts: alerts }
```
mauditecandela: @ragesoss do you mean by scope adding it to the campaign model? That makes sense, yes. I was also thinking about adding a reducer and action and therefore creating here a json so we are able to call something like /campaigns/:campaign_slug/alerts.json and receive a json with the information. 
I am not sure what would be better specially in terms of performance or good practices. Any suggestion?

mauditecandela:
captura de pantalla 2018-03-05 a las 18 13 39
@ragesoss this is the current state of the implementation. I would like to have your opinion on both technical and the solution to continue :) thanks!

ragesoss:
The code looks clean and makes sense so far!

mauditecandela:
@ragesoss I modified a bit the outlook. So far it looks it passes al tests. 
I don't know but the "type" attribute from alerts is not being rendered to the json. Do you know why is that happening?

ragesoss:
Cool!

alerts

To be useful, this will need to include article (when there is one) as well. The course, username, and article should all be links. I think 'resolved' should not be shown here. What it means is not very intuitive — it's a feature designed mainly for Wiki Ed staff.

PR #1747, opened 2018-03-02
Onboarding-Alert

AmitJoki:


app/models/alert.rb
```diff
@@ -46,6 +46,7 @@ class Alert < ApplicationRecord
     SurveyResponseAlert
     UnsubmittedCourseAlert
     UntrainedStudentsAlert
+    OnboardingAlert
```
ragesoss: This should be placed alphabetically.

db/schema.rb
```diff
@@ -33,10 +33,10 @@
     t.index ["user_id"], name: "index_alerts_on_user_id"
   end
 
-  create_table "articles", id: :integer, force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8 COLLATE=utf8_unicode_ci" do |t|
```
ragesoss: schema.rb changes need to be removed.

app/models/onboarding_alert.rb
```diff
@@ -0,0 +1,32 @@
...
+#  target_user_id :integer
+#  subject_id     :integer
+#  resolved       :boolean          default(FALSE)
+#  details        :text(65535)
+#
+
+# Alert for from where the instructor learnt about Wiki Edu.
+class OnboardingAlert < Alert
+  def main_subject
+    'WikiEdu: Onboarding Alert'
```
ragesoss: I suggest something more like "Onboarding Alert: [username] / [real name]"

This is the subject line of the email, so seeing the username and real name of the user will make these more useful.

app/controllers/onboarding_controller.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "#{Rails.root}/lib/alerts/onboarding_alert_manager"
```
ragesoss: Use `require_dependency` instead of `require`. The latter can cause a lot of warnings in some circumstances.

app/assets/javascripts/utils/api.js
```diff
@@ -675,6 +675,24 @@ slide_id=${opts.slide_id}`,
     );
   },
 
+  supplement(data) {
```
ragesoss: Instead of tossing another method into api.js, this can be put into an onboarding_actions file.

lib/alerts/onboarding_alert_manager.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class OnboardingAlertManager
+  def create_alert(user_name, msg)
+    user = User.find_by(username: user_name)
+    unless Alert.exists?(user_id: user.id, type: 'OnboardingAlert')
```
ragesoss: I think this check is not helpful. It'll be better to create a second alert if the same user goes through onboarding again and submits a new response.

lib/alerts/onboarding_alert_manager.rb
```diff
@@ -0,0 +1,14 @@
...
+      alert = Alert.create(type: 'OnboardingAlert',
+                           user: user,
+                           message: "Heard about Wiki Edu from: #{msg}",
```
ragesoss: Just pass the message itself, I think.

AmitJoki:
It's past midnight here in India. Will work on all these and submit a fresh PR tomorrow :+1:

AmitJoki:
closing in favour of #1758

PR #1748, opened 2018-03-05
Course edits now can be disabled for single course, closes issue #1440

### LINKED ISSUE #1440: Allow course edits to be disabled for a single course
Probably this could be done the same way we handle enabling/disabling timelines, with another entry in the course flags.

See https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/1b66177d705d1836340ca7612b478f570c1c59f2

vzhz:
This PR closes #1440 :
- add front-end toggle for `wiki_edit_enabled`
- adds flag for single-course edits
- refactors `wiki_edit_enabled` into the `Course` class, now removed from subclasses


psinghal20:
@vzhz Please take a look at the failing Travis build. There are several linting errors, which are causing the build to fail.

app/assets/javascripts/components/overview/course_edit_toggle.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import CourseActions from '../../actions/course_actions.js';
+import uuid from 'uuid';
+
+const CourseEditToggle = React.createClass({
```
ragesoss: This looks like it duplicates a lot of the logic from YesNoSelector, which was recently extracted. Could you use that instead? See PrivacySelector, for example.

app/controllers/courses_controller.rb
```diff
@@ -208,6 +208,17 @@ def set_timeline_enabled
     end
   end
 
+  def set_course_edits_enabled
```
ragesoss: Doesn't look like this method gets called.

ragesoss:
This should include a test that shows a course that responds 'wiki_edits_enabled?' true initially, then after interacting with the selector and saving it, responds false.

vzhz:
Hey team, I've got the linter errors taken care of (locally, will push it all together) and am about to work on tracking down the uncalled "set_course_edits_enabled" that @ragesoss pointed out doesn't get called. Thanks for the comments, will be working through them today.

app/controllers/courses_controller.rb
vzhz: Fixed, thank you!

ragesoss:
@vzhz please have a look at the travis build and check out the failing specs.

What I expect from this change is that the current behavior will be unchanged, except for a course that has a flag set to *disable* edits. The flag would just override the case where edits are currently enabled and disable them for that course.

vzhz:
@ragesoss, do I understand that all courses should be editable by default and our flags should override this starting condition? Thanks!

ragesoss:
Thanks for this! I just merged a fix for #1440, and reviewing this PR was useful as I was sorting out the flow.

PR #1753, opened 2018-03-05
Use /srv instead of /var for storing DB dumps

ragesoss:
The majority of space on the outreachdashboard.wmflabs.org VPS is mounted at /srv, and storage is at a premium on the root partition currently.

This should give us some extra breathing room.

### DIFF
--- a/server_config/outreach_mysql_backup.sh
+++ b/server_config/outreach_mysql_backup.sh
@@ -1,11 +1,12 @@
 #!/bin/bash
 
 # This script backs up the DB_NAME database into the specified BACKUP_PATH
-#   and syncs it to REMOTE_USER_HOST under the same path
+#   and syncs it to REMOTE_USER_HOST under REMOTE_BACKUP_PATH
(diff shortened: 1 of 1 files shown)


ragesoss:
@MusikAnimal This is a tweak the backup script you wrote for P&E Dashboard. I was wondering if you might take a look at this and make sure I'm not missing anything that will cause problems.

MusikAnimal:
Oh man, it's been so long! =p I assume you'll want to manually move whatever is left in `/var/outreach_backup/db/weekly/dashboard` (the old location) to `/srv` before deploying this, so those files don't get deleted on the remote server. The code otherwise looks right to me, but I have not tested it.

ragesoss:
Yep, I already copied the stuff over to /srv (and can delete the copies in /var after this is deployed).

Thanks! I'll try it out tomorrow.

ragesoss:
@MusikAnimal it worked! thanks!

PR #1754, opened 2018-03-06
Default campaign creation

### LINKED ISSUE #1749: Error message for no default campaign is unhelpful
After you finishing setting up a dev environment initially, loading the app will result in an error message about no campaigns. It actually means that the default campaign does not exist, so even if you create a campaign you will not necessarily fix this problem.

The handling of this error in `RescueDevelopmentErrors` should just create the default campaign, instead of displaying an error message.

AmitJoki:
Fixes #1749

### DIFF
--- a/lib/errors/rescue_development_errors.rb
+++ b/lib/errors/rescue_development_errors.rb
@@ -25,14 +25,9 @@ module Errors
       end
     end
 
-    NO_CAMPAIGNS_EXPLANATION =
-      'Error: There are no campaigns.' \
-      "\n\n" \
-      'Go to "/campaigns" to create one.'
     def self.rescue_from_no_campaigns(base)
       base.rescue_from CoursesPresenter::NoCampaignError do
-        render plain: NO_CAMPAIGNS_EXPLANATION,
-               status: 500
+        Campaign.create(title: "Default Campaign", slug: ENV['default_campaign'])
       end
     end
   end
--- a/lib/errors/rescue_development_errors.rb
+++ b/lib/errors/rescue_development_errors.rb
@@ -28,7 +28,7 @@ module Errors
     def self.rescue_from_no_campaigns(base)
       base.rescue_from CoursesPresenter::NoCampaignError do
         Campaign.create(title: 'Default Campaign', slug: ENV['default_campaign'])
-        redirect_to '/campaigns'
+        redirect_to '/'
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/errors/rescue_development_errors.rb
```diff
@@ -25,14 +25,10 @@ def self.rescue_from_rev_manifest(base)
...
-    NO_CAMPAIGNS_EXPLANATION =
-      'Error: There are no campaigns.' \
-      "\n\n" \
-      'Go to "/campaigns" to create one.'
     def self.rescue_from_no_campaigns(base)
       base.rescue_from CoursesPresenter::NoCampaignError do
-        render plain: NO_CAMPAIGNS_EXPLANATION,
-               status: 500
+        Campaign.create(title: 'Default Campaign', slug: ENV['default_campaign'])
+        redirect_to '/campaigns'
```
ragesoss: Probably redirect to `/` would make the most sense, as that's generally the route where this error is encountered.

AmitJoki: Alright. Just a sec.

ragesoss: You might also need to update `spec/lib/errors/rescue_development_errors_spec`.

spec/lib/errors/rescue_development_errors_spec.rb
```diff
@@ -39,7 +39,7 @@ def index
 
     it 'renders an explanation with helpful advice' do
```
ragesoss: This is no longer accurate.

ragesoss: Instead of the mock index above that raises the error, you could do a before { Campaign.destroy_all }, then visit the normal index page, and expect the (newly created) default campaign to show up.

AmitJoki: Sure. Will work on it.

spec/lib/errors/rescue_development_errors_spec.rb
```diff
@@ -29,17 +29,13 @@ def index
...
+    before do
+      Campaign.destroy_all
+      get '/'
     end
 
-    it 'renders an explanation with helpful advice' do
-      get :index
-      expect(response.body).to match(/There are no campaigns/)
+    it 'creates a default campaign' do
+      expect(Campaign.count).to eq(1) # The default campaign has been created
```
ragesoss: Try this:

```ruby
  describe 'when CoursesPresenter::NoCampaignError is raised' do
    controller(ApplicationController) do
      include Errors::RescueDevelopmentErrors

      def index
        raise CoursesPresenter::NoCampaignError
      end
    end

    it 'creates a default campaign' do
      Campaign.destroy_all
      expect(Campaign.count).to eq(0)
      get :index
      expect(Campaign.count).to eq(1) # The default campaign has been created
    end
  end
```

PR #1755, opened 2018-03-06
Limit-Articles

### LINKED ISSUE #1656: Limit course articles.json results for programs with huge numbers

AmitJoki:
Fixes #1656


app/assets/javascripts/components/articles/articles_handler.jsx
```diff
@@ -1,37 +1,52 @@
...
   },
 
   componentWillMount() {
-    ServerActions.fetch('articles', this.props.course_id);
     ServerActions.fetch('assignments', this.props.course_id);
+    return this.props.fetchArticles(this.props.course_id, this.props.limit);
+  },
+
+  showMore() {
+    return this.props.fetchArticles(this.props.course_id, this.props.limit + 100);
```
ragesoss: This should go up by 500 each time. Fetching 500 initially and then just 100 more each extra fetch will be pretty tedious for very active programs.

app/assets/javascripts/components/articles/article_list.jsx
```diff
@@ -70,7 +63,7 @@ const ArticleList = ({
       table_key="articles"
       className="table--expandable table--hoverable"
       none_message={CourseUtils.i18n('articles_none', course.string_prefix)}
-      store={ArticleStore}
```
ragesoss: Is ArticleStore used anywhere in the app still? Since the fetch has moved to redux, we need to make sure ArticleStore is being replaced with redux state everywhere it is used, and the store should be deleted.

ragesoss: ArticleStore is also being used substantively by AssignmentList (and required but not needed for anything as far as I can tell in AvailableArticlesList). AssignmentList should be updated to use the `articles` props that it receives, instead of calling ArticleStore to find the article (if any) corresponding to the assignment title, and the usage of ArticleStore can probably just be removed from AvailableArticlesList with no further changes.

AmitJoki: Can I remove Editable from AssignmentList too?

ragesoss: No, I don't think so. It needs it for refreshing the assignment list when AssignmentStore changes.

ragesoss: Once AssignmentStore gets converted to redux, then we can remove Editable.

AmitJoki: yeah, I thought so too.

app/assets/javascripts/components/assignments/assignment_list.jsx
```diff
@@ -28,14 +27,30 @@ const AssignmentList = createReactClass({
     });
   },
 
+  getFilteredArticles(options) {
```
ragesoss: see getFiltered in `utils/model_utils.js`.

ragesoss:
Regarding the `descKeys` feature from ArticleStore, I forgot that we already have a way to handle that. See the `users` reducer, and how it uses `SORT_DESCENDING`.

app/assets/javascripts/reducers/articles.js
```diff
@@ -8,6 +8,14 @@ const initialState = {
   sortKey: 'character_sum'
 };
 
+const SORT_DESCENDING = {
+  rating_num: true,
+  title: true,
```
ragesoss: To match the current behavior, just character_sum and view_count should be here.

app/assets/javascripts/reducers/articles.js
```diff
@@ -22,13 +30,9 @@ export default function articles(state = initialState, action) {
...
       const newState = { ...state };
-      if (action.key === state.sortKey) {
-        newState.articles = _.sortBy(state.articles, action.key).reverse();
-        newState.sortKey = null;
-      } else {
-        newState.articles = _.sortBy(state.articles, action.key);
-        newState.sortKey = action.key;
-      }
+      const sorted = sortByKey(newState.articles, action.key, state.sortKey, SORT_DESCENDING[action.key]);
+      newState.users = sorted.newModels;
```
ragesoss: `newState.articles`

app/assets/javascripts/reducers/articles.js
```diff
@@ -8,6 +8,14 @@ const initialState = {
   sortKey: 'character_sum'
```
ragesoss: This does not work to actually sort them initially.

I suggest instead to do a second dispatch in the fetchArticles action, right after RECEIVE_ARTICLES, to sort them by `character_sum`.

PR #1758, opened 2018-03-07
Onboarding alert

AmitJoki:


ragesoss:
Looks like this is all passing except for the spec that is looking for the wrong bit of text from the earlier version on the supplementary info screen of onboarding.

AmitJoki:
Yeah. Changing and pushing the change in a minute.

app/models/onboarding_alert.rb
```diff
@@ -0,0 +1,32 @@
...
+#  target_user_id :integer
+#  subject_id     :integer
+#  resolved       :boolean          default(FALSE)
+#  details        :text(65535)
+#
+
+# Alert for from where the instructor learnt about Wiki Edu.
+class OnboardingAlert < Alert
+  def main_subject
+    "Onboarding Alert: [#{user.username}] / [#{user.real_name}]"
```
ragesoss: Remove the brackets.

AmitJoki: Okay.

spec/features/onboarding_spec.rb
```diff
@@ -63,4 +63,21 @@
...
+      login_as(user, scope: :user)
+    end
+
+    it 'should go to supplementary' do
+      visit onboarding_path
+      find('.intro .button').click
+      find('input[type=radio][value=true]').click
+      find('form button[type=submit]').click
+      sleep 1
+      expect(page).to have_content 'us?'
```
ragesoss: The `sleep 1` is almost certainly not necessary. `have_content` will wait for a while until it matches.

This is not a very good way to test this. The language may change again the in the future, and this is a small enough snippet that it might match accidentally even if this feature was not working as intended.

You should instead do something like interact with the form, which should have a specific selector (such as an ID on the input) that is unique to this component.

AmitJoki: So, I can be certain that after `find('form button[type=submit]').click` is done, the next page will be loaded alright right? Also, I can just test if a specific component is present in the page right?

AmitJoki:
@ragesoss the build is failing due to some Training Translation VCR spec. Not my changes.

PR #1770, opened 2018-03-10
Persist wizard

### LINKED ISSUE #131: Users don't expect 'back' button in browser to immediately exit the wizard
If I'm in the middle of the wizard, the back button will immediately take me to the previous page, and my progress through the wizard will be lost. A better behavior would be to take me one step back in the wizard.

AmitJoki:
Fixes #131

### DIFF
--- a/app/assets/javascripts/components/wizard/wizard.jsx
+++ b/app/assets/javascripts/components/wizard/wizard.jsx
@@ -32,6 +32,7 @@ const Wizard = createReactClass({
     return getState();
   },
   componentWillMount() {
+    history.pushState({ index: 0 }, 'wizard', 'step-1'); // Initial States
     return ServerActions.fetchWizardIndex();
   },
   componentWillUnmount() {
@@ -111,5 +112,10 @@ const Wizard = createReactClass({
   }
 }
 );
+
+window.onpopstate = function (event) {
+  WizardActions.rewindWizard(event.state.index);
+};
+
 
 export default Wizard;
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/wizard/wizard.jsx
```diff
@@ -19,6 +19,24 @@ const getState = () =>
...
+  window.onpopstate = function (event) { // Listen to changes
+    if (event.state) {
+      WizardActions.goToWizard(event.state.index);
+    }
+  };
+};
+
+const unloadEvents = function () {
+    window.onpopstate = null;
+    window.onbeforeunload = null;
```
ragesoss: These should probably be camelCase?

AmitJoki: @ragesoss They are vanilla JS events. They're not something which a user can define. We can only listen to these events. See [onpopstate](https://developer.mozilla.org/en-US/docs/Web/API/Window/onpopstate) & [onbeforeunload](https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload)

ragesoss:
This works really nicely! The only glitch I notice is that the step in the page title does not change when you use forward/back. If you can consolidate that into the main functions for changing steps, that would be good. (If not, the title steps could be removed.)

AmitJoki:
@ragesoss can you please check it now? Hopefully, the titles must now be synced. If you still find some errors, we could probably take the title steps off.

ragesoss:
@AmitJoki it works for most parts of the wizard, but it gets out of sync when you navigate to an earlier step from the final Summary step. The step number rendered in the panel itself is always correct, though. Is there a good way to tie it to that?

AmitJoki:
@ragesoss everything is fixed now. I ran it locally and every step is getting synced.

PR #1771, opened 2018-03-10
Commons wiki type

### LINKED ISSUE #1764: Create new Course type for Wikimedia Commons upload campaigns
Some events are specifically focused on commons. Users should be able to select Commons as the home wiki, so that edits from other wikis do not get pulled in.

This will require converting `CommonsWiki` to a database-backed normal `Wiki`.

AmitJoki:
Fixes #1764


lib/commons.rb
```diff
@@ -134,6 +134,6 @@ def parse_image_data_and_update_continue(response)
   private
 
   def api_get
-    WikiApi.new(CommonsWiki.new).query(@query)
+    WikiApi.new(Wiki.new(language: nil, project: 'commons')).query(@query)
```
ragesoss: This should be `Wiki.get_or_create`.

Also, you'll probably want to add Commons as a fixture for the specs.

AmitJoki: Can I put `en` as the language in the fixture? If not, should I leave it nil, that is don't define language property at all in the fixture?

ragesoss: it should probably be `commons` as the language and `wikimedia` as the project.

AmitJoki:
@ragesoss can you please check why it is failing? Is it because of the fixture that I added?

psinghal20:
@AmitJoki Yes, the specs are failing due to the fixture added. The error displayed is `Mysql2::Error: Table 'test.commons' doesn't exist: DELETE FROM 'commons'`. In my opinion, as now you are using the Wiki model for the Common Wikis, you should add your fixture in `Wiki.yml` instead of creating a separate file.

app/models/wiki.rb
```diff
@@ -23,6 +23,7 @@ class Wiki < ApplicationRecord
   validates_uniqueness_of :project, scope: :language
 
   PROJECTS = %w[
+    commons
```
ragesoss: this should go in 'language'.

app/models/wiki.rb
```diff
@@ -58,7 +59,8 @@ class Wiki < ApplicationRecord
 
   MULTILINGUAL_PROJECTS = {
     'wikidata' => 'www.wikidata.org',
-    'wikisource' => 'wikisource.org'
+    'wikisource' => 'wikisource.org',
+    'commons' => 'commons.wikimedia.org'
```
ragesoss: This isn't necessary since we are using `commons` as the language.

lib/commons.rb
```diff
@@ -134,6 +134,6 @@ def parse_image_data_and_update_continue(response)
   private
 
   def api_get
-    WikiApi.new(CommonsWiki.new).query(@query)
+    WikiApi.new(Wiki.get_or_create(language: nil, project: 'commons')).query(@query)
```
ragesoss: This needs to match the fixture.

spec/fixtures/wikis.yml
```diff
@@ -11,3 +11,8 @@ default:
   id: 1
   language: en
   project: wikipedia
+
+commons:
+  id: 2
```
ragesoss: You'll need to fix the failing spec that has a hard-coded wiki id. See travis.

AmitJoki:
@ragesoss can you please look into what is causing the spec to fail? I have used another id instead of the one in fixture to avoid collision. But an entirely unrelated part of the spec is failing. Could you please give me some pointers?

ragesoss:
Those do indeed look unrelated. Weird. I'll look when I poke it when I get a chance. Maybe not today.

ragesoss:
Oh, you did change that article status manager spec. I assume it's failing locally too. During the spec, try verifying that each article corresponds to the expected wiki (en.wikipedia vs es.wikipedia).

AmitJoki:
@ragesoss yes it is failing the spec even after I reset the contents of the file. I can see that the language 'es' isn't getting set up right. I tried setting `id: 666` to eliminate any conflicts but the spec is still failing on that one. By the way, I am running it like `bundle exec rspec 'path/to/the/file/'` Is it the problem with the fixture setting?

ragesoss:
Running the spec like that should be fine. Try doing something like `Wiki.all.each { |wiki| pp wiki }` right before the line where it fails, to see all the wikis in the database and whether they look like what you expect.

AmitJoki:
Actually, it is setting it right:

```
[#<Wiki id: 666, language: "commons", project: "wikimedia">, #<Wiki id: 1, language: "en", project: "wikipedia">, #<Wiki id: 2, language: "es", project: "wikipedia">]
```
I guess something wrong's with `update_article_for_course` ?

ragesoss:
@AmitJoki try rebasing on master, and deleting the `fixtures` directory.

ragesoss:
Hmm... thinking this over more, I think we should just do the changes in `Wiki` at this point, and leave `CommonsWiki` as it is. That will do the minimum of allowing a user to pick commons.wikimedia.org as the home wiki for a program, without having to work around existing tests or adding extra database calls for every Commons upload query.

AmitJoki:
Should I just restore the `CommonsWiki` then?

ragesoss:
I think if you just include the changes to `wiki.rb`, that will be the minimal fix.

Issue #1769, opened 2018-03-10
For Editathons, dashboard stats should refresh automatically


ragesoss:
Now that we have a quick update cycle for short-running events, it will be practical to auto-refresh the course stats while a user has the overview page open. Re-fetching the course every 5 minutes or so might be all it takes to do that.

vzhz:
I'd be happy to work on this next, if it isn't addressed by @AmitJoki 's fix on #1772 .

Issue #1768, opened 2018-03-10
'Join' confirmation message should set expectations for activity tracking


srish:
In the edit-a-thon (that I attended at SFMOMA recently) guidelines, it was mentioned that an editor should keep their dashboard window open in a browser tab as this helps track their edit activity on Wikipedia. 

If this is true, then a popup could be displayed at the top of the app window when a user logs in for the first time in the dashboard at an event. I noticed that many participants didn't read the guidelines at first, and didn't keep their tab open while they were editing. As the edit count in the dashboard does not update in real time, they were not sure if their edits were tracked or not.  

A message explaining how the participants' edit activity will be tracked might help add clarity.

ragesoss:
@srish there's no need to keep the dashboard window open. Once a user joins an event on the dashboard, their edits will be tracked. I think keeping the window open was just suggested so that users can easily see the stats during the event (but not in real time, and only if they refresh).

srish:
@ragesoss WOW! This is helpful to know. Thanks!

ragesoss:
This can probably be done by expanding the `courses.join_successful` interface message to include more details about what the dashboard does.

PR #1772, opened 2018-03-10
Make the dashboard realtime through polling

AmitJoki:
Hopefully fixes both #1768 and #1769

### DIFF
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -25,6 +25,7 @@ import Editable from '../high_order/editable.jsx';
 import TextInput from '../common/text_input.jsx';
 import DatePicker from '../common/date_picker.jsx';
 import CourseActions from '../../actions/course_actions.js';
+import ServerActions from '../../actions/server_actions.js';
 
 import CourseStore from '../../stores/course_store.js';
 import TagStore from '../../stores/tag_store.js';
@@ -39,6 +40,8 @@ const getState = () =>
     error_message: ValidationStore.firstMessage()
   })
 ;
+
+const POLL_INTERVAL = 300000; // 5 minutes
 
 const Details = createReactClass({
   displayName: 'Details',
@@ -58,6 +61,15 @@ const Details = createReactClass({
     return getState();
   },
 
+  componentDidMount() {
+    this.timeout = this.poll(); // Start polling
+  },
+
+  componentWillUnmount() {
+    if (this.timeout) {
+      clearInterval(this.timeout); // End it
+    }
+  },
   updateDetails(valueKey, value) {
     const updatedCourse = this.props.course;
     updatedCourse[valueKey] = value;
@@ -88,6 +100,15 @@ const Details = createReactClass({
     // On P&E Dashboard, anyone with edit rights for the course may rename it.
     return true;
   },
+
+  poll() {
+    if (this.state.course.type === 'Editathon') {
+      return setInterval(() => ServerActions.fetch('course', this.state.course.slug), POLL_INTERVAL);
+    }
+    return null;
+  },
+
+  timeout: null,
 
   render() {
     const canRename = this.canRename();
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -103,7 +103,11 @@ const Details = createReactClass({
 
   poll() {
     if (this.state.course.type === 'Editathon') {
-      return setInterval(() => ServerActions.fetch('course', this.state.course.slug), POLL_INTERVAL);
+      return setInterval(() => {
+        if (!this.props.editable) {
+          ServerActions.fetch('course', this.state.course.slug);
+        }
+      }, POLL_INTERVAL);
     }
     return null;
   },
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -102,10 +102,10 @@ const Details = createReactClass({
   },
 
   poll() {
-    if (this.state.course.type === 'Editathon') {
+    if (this.props.course.type === 'Editathon') {
       return setInterval(() => {
         if (!this.props.editable) {
-          ServerActions.fetch('course', this.state.course.slug);
+          ServerActions.fetch('course', this.props.course.slug);
         }
       }, POLL_INTERVAL);
     }
--- a/app/assets/javascripts/components/overview/overview_handler.jsx
+++ b/app/assets/javascripts/components/overview/overview_handler.jsx
@@ -11,7 +11,6 @@ import ThisWeek from './this_week.jsx';
 import CourseStore from '../../stores/course_store.js';
 import AssignmentStore from '../../stores/assignment_store.js';
 import WeekStore from '../../stores/week_store.js';
-import CourseActions from '../../actions/course_actions.js';
 import Loading from '../common/loading.jsx';
 import CourseClonedModal from './course_cloned_modal.jsx';
 import SyllabusUpload from './syllabus-upload.jsx';
--- a/app/assets/javascripts/components/overview/overview_handler.jsx
+++ b/app/assets/javascripts/components/overview/overview_handler.jsx
@@ -50,6 +50,7 @@ const Overview = createReactClass({
 
   componentDidMount() {
     ServerActions.fetch('timeline', this.props.course_id);
+    this.timeout = poll();
     return ServerActions.fetch('tags', this.props.course_id);
   },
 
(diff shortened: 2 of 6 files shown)


AmitJoki:
#1771  is passing, while the build errors on this one has `translation_spec` and `common_spec` failing which is weird.

psinghal20:
@AmitJoki If you see the Travis log you can find that there are javascript linting errors in your code leading to failure. Also, you forgot to add the documentation for added I18n strings in `qqq.yml` which are required by the WikiTranslators to translate the strings in supported languages.

AmitJoki:
@psinghal20, I was not sure if my description would be accepted and thought that part could be done in the reviewing processes. Anyhow, I have added some corresponding documentation right now. Thanks! :+1: for the heads up on linting errors. I had forgot to push the changes I'd made locally.

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -50,10 +53,23 @@ const Overview = createReactClass({
...
+      clearInterval(this.timeout);
+    }
+  },
+
   storeDidChange() {
     return this.setState(getState());
   },
 
+  timeout: (function () {
+    if (this.state.course.level === 'Editathon') {
```
psinghal20: @AmitJoki I don't this function is bound therefore the components state is not accessible by this function. Please look into the Travis Build log for more info.

AmitJoki: ah yes! Forgot only fat arrow functions make use of the contextual `this`. Have to bind now.

AmitJoki:
@ragesoss many irrelevant builds are failing. An issue with the Travis CI?

ragesoss:
@AmitJoki No, those look like JS bugs.

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -47,13 +50,29 @@ const Overview = createReactClass({
 
   componentDidMount() {
     ServerActions.fetch('timeline', this.props.course_id);
+    this.timeout = poll();
```
psinghal20: @AmitJoki shouldn't this be `this.poll()`. If you see the Travis log, all the failing specs are caused by this statement.

AmitJoki: aaarrrgggggh!

AmitJoki: Yeah, it should be smh :|

config/locales/en.yml
```diff
@@ -447,7 +447,10 @@ en:
     join: Join
     join_course: Join course
     join_prompt: Join '%{title}'?
-    join_successful: You’ve successfully joined %{title}.
+    join_successful: >
+     You’ve successfully joined %{title}.
+     In case of an Editathon, the dashboard updates itself about every 5 minutes or so.
+     For other course levels, you could always refresh this page to get the latest data.
```
ragesoss: You’ve successfully joined %{title}.

The dashboard will automatically keep track of your contributions with periodic updates. Check back later to see what impact you've made!

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -47,13 +50,29 @@ const Overview = createReactClass({
...
   },
 
+  poll() {
+    if (this.state.course.level === 'Editathon') {
+      return setInterval(() => CourseActions.updateCourse(), POLL_INTERVAL);
+    }
+    return null;
+  },
+
+  timeout: null,
```
ragesoss: Doing it like this seems weird. Why keep the timeout on the class directly, instead of in state?

AmitJoki: IMO, it need not be in state because, it is not going to change once its assigned. Also, I found `this.timeout` more semantic because `Overview` has a timeout, *not the state* of `Overview`.

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -47,13 +50,29 @@ const Overview = createReactClass({
...
+      clearInterval(this.timeout);
+    }
+  },
+
   storeDidChange() {
     return this.setState(getState());
   },
 
+  poll() {
+    if (this.state.course.level === 'Editathon') {
```
ragesoss: `course.level` doesn't look right.

AmitJoki: @ragesoss could you elaborate a little bit more please? Is the literal 'Editathon' I am comparing with, wrong?

ragesoss: `course.type` is what I would expect, unless I'm missing something. Did you get this working locally?

AmitJoki: `course.type` contains values like `"ClassroomProgramCourse"` which I don't think is the case here. I checked out that the "BasicCourse", "Editathon" were under `course.level`. It worked for me locally, in the sense I could see the states being refreshed at intervals.

ragesoss: course.type will be one of the values from `Course::COURSE_TYPES`.

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -47,13 +50,29 @@ const Overview = createReactClass({
...
+    }
+  },
+
   storeDidChange() {
     return this.setState(getState());
   },
 
+  poll() {
+    if (this.state.course.type === 'Editathon') {
+      return setInterval(() => CourseActions.updateCourse(), POLL_INTERVAL);
```
ragesoss: I don't see how this could have worked for you. CourseActions.updateCourse() does not fetch the course, it just updates the store when the user makes changes to to the course. Calling it with no data is basically a noop.

I had to change it to `ServerActions.fetch('course', this.state.course.slug)`.

ragesoss:
One wrinkle just occurred to me: if the user is editing the course via Edit Details, they may have some changes that are not yet persisted to the server which would get undone whenever the polling occurs. The fetch should only happen if the user is *not* in Edit Details mode.

AmitJoki:
@ragesoss how do I check for it? Is there any prop I could use?

psinghal20:
@AmitJoki The only thing I can think of signifying the Edit mode will be the `editable` which is part of Editable component's state. One way I think of handling this is saving this state in some store so that the overview handler has access to it and in turn can take care Polling in such cases.

ragesoss:
Or we could move the polling to Details. Probably want to wait until the edit mode redesign is merged though.

ragesoss:
I just merged the Details change.

AmitJoki:
So, I'd have to call the fetch in Details right?

ragesoss:
Yeah, that's the cleanest way I can think of to do it.

ragesoss:
Looks like this introduces a JS error: `Can't find variable: ServerActions`

ragesoss:
@AmitJoki now you should be able to use `this.props.editable` in `Details` to turn off polling while in edit mode.

AmitJoki:
@ragesoss completed :+1:

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -89,6 +101,19 @@ const Details = createReactClass({
     return true;
   },
 
+  poll() {
+    if (this.state.course.type === 'Editathon') {
+      return setInterval(() => {
+        if (!this.props.editable) {
+          ServerActions.fetch('course', this.state.course.slug);
```
ragesoss: This needs to use `this.props.course` instead of state. (No reason not to use props for the type check too.)

With state, if you edit details, make a change to the course slug (not persisted to the server), then cancel, the state will still have a modified slug and it will start throwing 404 errors.

config/locales/qqq.yml
```diff
@@ -406,7 +406,9 @@ qqq:
     join_prompt: |-
       Do you want to join this course?
       {{Identical|Join}}
-    join_successful: Message that join was successful
+    join_successful: Message that join was successful and explains the functioning of Dashboard
+      with emphasis on how Editathons are updated every 5 minutes while other course level
```
ragesoss: This is not accurate after the changes above.

app/assets/javascripts/components/overview/details.jsx
AmitJoki: Yeah. Makes sense. We want to encapsulate within `details.jsx` without bothering about the state residing outside of it. Will do.

config/locales/qqq.yml
AmitJoki: What would you have it changed to instead? Something more precise?

ragesoss: Something more vague, I think. "Message that joining a program was successful, with an explanation of how the Dashboard will track contributions."

PR #1775, opened 2018-03-10
Convert InputMixin to a high-order Component

### LINKED ISSUE #1456: Get rid of `InputMixin`
We use this mixin for most of the input fields in the app, mainly for handling validations in an extensible way. However, this one in particular seems to be the source of a lot of warnings that will need to be fixed for the React 16 upgrade, and this is the only mixin that isn't just a store subscription.

psinghal20:
Fixes #1456.
It replaces the `InputMixin` with `InputHOC` component which is then used with several input fields to validate entries by using validation entries.


psinghal20:
@ragesoss I separated the `InputMixin` conversion to another PR as it was getting harder to find and fix the errors with changes done to the validation store. I will continue with `ValidationStore` changes in another PR.

psinghal20:
@ragesoss Is the failing Travis build a random failure? I don't think I made any changes to CourseCloneModal. If it is a random failure, can you please restart the Travis build.

ragesoss:
Will do. Can you confirm that that spec passes locally? It does some unique things in terms of state, inputs, and validations, and this one is not one that I've seen fail randomly in a long while.

AmitJoki:
Have you tried checking which input components use `InputHOC` in the cloning process? That might help you in singling out the problem.

psinghal20:
@ragesoss The specs seems to fail when the same course term is used as the original course, but in my local branch, the failure message does appear when the course term is same as the original course. When I try to run this spec on my local branch, the test fails and raises the error `PhantomJS client died while processing`. I am not sure about the error on my local branch and in case of Travis build, as the message does appear on my local, I think updating the spec with a sleep statement like `sleep 2` maybe can help with the failing spec.

ragesoss:
I take it that course cloning works as expected when you do it manually? Putting in the same term as the original will show the error message and let you change them term and proceed?

ragesoss:
`expect(page).to have_content('This course already exists')`

That uses Capybara's `wait` behavior, so it will wait up to 15 seconds for the expectation to be met before timing out. Sleep is unlikely to help.

psinghal20:
Yes, it does behave as expected when I try it manually but I don't know why the local spec makes my PhantomJS client die.

ragesoss:
Okay. I'll poke at it next week.

psinghal20:
@ragesoss Found what was causing my local failure, it was because of console statement I added to debug. Removed and ran all the feature specs. All specs pass on my local build.

PR #1777, opened 2018-03-10
Better 404 for courses

### LINKED ISSUE #365: 404 for a /course url should indicate the slug of the nonexistent course
The message should be something like 'no course foo/bar_(baz)' exists.

AmitJoki:
Fixes #365

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -34,10 +34,8 @@ class ErrorsController < ApplicationController
   private
 
   def not_found_message
-    if params[:endpoint] == 'overview'
-      return I18n.t 'error_no_course.header', slug: "#{params[:school]}/#{params[:titleterm]}"
-    else
-      return I18n.t 'error_404.header'
-    end
+    if params[:endpoint] == 'overview' # only if it is a course
+      return I18n.t 'error_no_course.explanation', slug: "#{params[:school]}/#{params[:titleterm]}"
+    return I18n.t 'error_404.explanation'
   end
 end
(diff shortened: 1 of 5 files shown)


ragesoss:
I'd rather find a way to do this without JS.

AmitJoki:
I'll do it in `errors_controller.rb` then :+1:

AmitJoki:
@ragesoss removed javascript and have added it to the controller instead.

app/controllers/errors_controller.rb
```diff
@@ -20,6 +21,16 @@ def incorrect_passcode
     render status: 401
   end
 
+  private
```
psinghal20: I don't think the login_error method should be private. That is causing the failing specs in Travis Build. Also, there are some Rubocop linting errors in the code. Please have a look at the Travis build for more info.

AmitJoki: yeah. I just pushed the fix.

psinghal20:
@AmitJoki Please update the `error_pages_spec.rb` for the new message that you have added.

app/controllers/errors_controller.rb
```diff
@@ -29,4 +30,14 @@ def login_error
       render status: 200
     end
   end
+
+  private
+
+  def message
```
ragesoss: A more specific name would be helpful here, as this is specific to 404 errors.

AmitJoki: is `course_not_found` okay?

ragesoss: It's also the message for other not found errors, so I'd go with `not_found_message`.

ragesoss:
I don't think it works well in the header. Maybe instead, it could switch the message beneath. It should not say anything about courses if it's *not* a course page, and the smaller-text message beneath should include the slug if it is a course.

Also, it's no longer the case that "If it was recently registered, it may take a little while for the system to add it."

no such course

AmitJoki:
@ragesoss the failing spec is unrelated. Can you please rebuild this?

PR #1778, opened 2018-03-10
Add project based filtering for Edited Articles list

### LINKED ISSUE #1686: Add filtering by project to Articles Edited table
The Articles Edited table may contain edits from multiple wikis. For example: https://dashboard.wikiedu.org/courses/University_of_North_Carolina_at_Chapel_Hill/Helping_Giving_Away_Psychological_Science_(Spring_2018)/articles

It would be useful to be able to filter it (with controls that only show up if there are multiple wikis present) to show one wiki at a time.

This will depend on ArticleStore being converted to redux first; then we can handle it pretty much just like the redux todo example plus a `reselect` selector.

psinghal20:
Fixes #1686.
This adds `reselect` based filtering for the Articles Edited list corresponding to projects present. The filter selector is displayed only in the case when there are more than two projects are present which have edited articles. `Lodash` is used to compute the `projects` list which is used to generate the options of the filter selector.
**ScreenShot**
peek 2018-03-11 22-50


ragesoss:
@psinghal20 this is my fault for the issue description, but it should actually filter by wiki rather than by project. As in, en.wikipedia.org vs es.wikipedia.org vs wikidata.org.

psinghal20:
Oh, okay. I will make the changes. I think changing this to Wiki won't take much effort.

psinghal20:
@ragesoss Made the changes.
New Screenshot:
peek 2018-03-13 01-38

PR #1779, opened 2018-03-10
Adds tool tip for timeline toggle input field

### LINKED ISSUE #1765: Timeline toggle should have an tooltip explaining what the timeline is good for

psinghal20:
Fixes #1765.
Adds a tooltip displaying a message about what the timeline is good for, uses I18n locales.
**Screenshot:**
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -538,6 +538,7 @@ en:
     timeline: Assignment Details
     timeline_enabled: Timeline enabled
     timeline_link: Timeline
+    timeline_tooltip: Timeline help you to keep track of your course by breaking the course duration in weeks which have proper agendas and tasks associated with them.
     time_zone_message: Start and end times are displayed in your local time zone.
     title: Title
     training_due:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -538,7 +538,7 @@ en:
     timeline: Assignment Details
     timeline_enabled: Timeline enabled
     timeline_link: Timeline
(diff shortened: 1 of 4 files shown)


psinghal20:
Suggestions for the tooltip message will be appreciated. :smile:

config/locales/en.yml
```diff
@@ -538,6 +538,7 @@ en:
     timeline: Assignment Details
     timeline_enabled: Timeline enabled
     timeline_link: Timeline
+    timeline_tooltip: Timeline help you to keep track of your course by breaking the course duration in weeks which have proper agendas and tasks associated with them.
```
vzhz: The only change I would make is a grammatical one: I would say "...into weeks with agendas and tasks associated with them." It's tighter as a whole and "into" is more clear because, for example, you can break something into pieces but don't break something in pieces. Just a suggestion! @psinghal20

config/locales/en.yml
```diff
@@ -538,6 +538,7 @@ en:
     timeline: Assignment Details
     timeline_enabled: Timeline enabled
     timeline_link: Timeline
+    timeline_tooltip: Timeline help you to keep track of your course by breaking the course duration into weeks which have proper agendas and tasks associated with them.
```
ragesoss: "Enabling the Timeline will let you specify week-by-week activities and assign training modules. It is useful for Education Program courses and other long, highly structured programs."

psinghal20:
@ragesoss Updated the tooltip message. Thanks :smile:

PR #1781, opened 2018-03-11
Add option to retype the passcode

### LINKED ISSUE #1496: If password in enroll link is incorrect, user should have a chance to put in the correct one
For edit-a-thons and other events, an enroll link may get distributed before the passcode is changed (or it may be miscopied). Handling incorrect passcodes silently, and prompting the user for the correct one, would be a nice improvement.

AmitJoki:
Fixes #1496 

image

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -18,7 +18,7 @@ class ErrorsController < ApplicationController
   end
 
   def incorrect_passcode
-    @path = request.referer.split('?')[0]
+    @path = params[:retry] || ''
     render status: 401
   end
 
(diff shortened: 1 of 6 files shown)


psinghal20:
@AmitJoki Maybe you can add a separate feature spec example for this feature than changing the previous one.
- Checks that user is allowed to enter the passcode again, after one failed attempt which you can do in the existing example.
- Checks that if the passcode is invalid even the second time, then display the corresponding error message.

AmitJoki:
@ragesoss I tried working on the spec but couldn't quite set up the environment. Can you please give me some pointers as to creating a course, logging the user in as a student? I have worked out the other parts.

ragesoss:
Check out `spec/features/student_role_spec.rb`, which includes a spec for putting in an incorrect passcode. You should be able to just extend that spec to include retyping.

AmitJoki:
@ragesoss can you please checkout my code for spec testing. I am not able to ascertain what's wrong as PhantomJS is crashing every time for me in here.

PR #1782, opened 2018-03-11
Salesforce IDs are now editable by admins

### LINKED ISSUE #1539: Admins should be able to change salesforce_id for a course
Mistakes happen, and they must currently be corrected manually from the console.

AmitJoki:
image

Fixes #1539


ragesoss:
Currently, we set the salesforce ID via a button in `AvailableActions`. To avoid unnecessary UI clutter, I think it would be better to keep in there. Maybe can just modify the existing action so that if the Salesforce ID is already present, it just changes the button text to 'update Salesforce ID'.

AmitJoki:
Closing in favour of #1791

PR #1785, opened 2018-03-11
Claim remove

### LINKED ISSUE #1783: For Available Articles, dual-role particants should have both 'claim' and 'remove' options
Right now, the remove option is replaced if the user is both instructor and student.

AmitJoki:
Fixes #1783

### DIFF
--- a/app/assets/javascripts/components/articles/available_article.jsx
+++ b/app/assets/javascripts/components/articles/available_article.jsx
@@ -62,13 +62,13 @@ export const AvailableArticle = createReactClass({
 
     let actionSelect;
     let actionRemove;
-    if (this.props.current_user.role === 0) {
+    if (this.props.current_user.isStudent) {
       actionSelect = (
         <button className="button dark" onClick={this.onSelectHandler}>{I18n.t('assignments.select')}</button>
       );
     }
 
-    if (this.props.current_user.role === 1 || this.props.current_user.admin) {
+    if (this.props.current_user.isNonstudent) {
       actionRemove = (
         <button className="button dark" onClick={this.onRemoveHandler}>{I18n.t('assignments.remove')}</button>
       );
--- a/app/assets/javascripts/components/articles/available_article.jsx
+++ b/app/assets/javascripts/components/articles/available_article.jsx
@@ -62,13 +62,13 @@ export const AvailableArticle = createReactClass({
 
     let actionSelect;
     let actionRemove;
-    if (this.props.current_user.role == 0) {
+    if (this.props.current_user.role === 0) {
       actionSelect = (
         <button className="button dark" onClick={this.onSelectHandler}>{I18n.t('assignments.select')}</button>
       );
     }
 
-    if (this.props.current_user.role == 1 || this.props.current_user.admin) {
+    if (this.props.current_user.role === 1 || this.props.current_user.admin) {
       actionRemove = (
         <button className="button dark" onClick={this.onRemoveHandler}>{I18n.t('assignments.remove')}</button>
       );
--- a/app/assets/javascripts/components/articles/available_article.jsx
+++ b/app/assets/javascripts/components/articles/available_article.jsx
@@ -60,7 +60,7 @@ export const AvailableArticle = createReactClass({
     const ratingMobileClass = `${ratingClass} tablet-only`;
     const articleLink = <a onClick={this.stop} href={article.url} target="_blank" className="inline">{article.formatted_title}</a>;
 
-    let actionButton;
+    let actionCell;
     // Show 'Select' button to students
     if (this.props.current_user.role === 0) {
       actionCell = (
@@ -68,7 +68,7 @@ export const AvailableArticle = createReactClass({
           <button className="button dark" onClick={this.onSelectHandler}>{I18n.t('assignments.select')}</button>
         </td>
       );
-    // Show 'Remove' button to admins and facilitators
+    // Show both 'Select'and Remove' button to admins and facilitators
     } else if (this.props.current_user.admin || this.props.current_user.role > 0) {
       actionCell = (
         <td className="table-action-cell">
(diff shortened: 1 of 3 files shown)


AmitJoki:
@ragesoss the build is failing because of `language_switcher.spec` which fails randomly sometimes. Could you please rebuild this?

ragesoss:
This looks like it will show the Select and Remove buttons to any admin, regardless of whether they are also in the student role.

Instead, it should work like this:
* User sees 'Select' if they are a student
* User sees 'Remove' if they are instructor or admin

Those should be independent, so the user would only see both if both conditions apply.

For this, we should be able to use `current_user.isStudent` and `current_user.isNonstudent`, since (unlike user.role, which is just a single value) these not mutually exclusive.

app/assets/javascripts/components/articles/available_article.jsx
```diff
@@ -60,16 +60,21 @@ export const AvailableArticle = createReactClass({
     const ratingMobileClass = `${ratingClass} tablet-only`;
     const articleLink = <a onClick={this.stop} href={article.url} target="_blank" className="inline">{article.formatted_title}</a>;
 
-    let actionButton;
+    let actionCell;
     // Show 'Select' button to students
     if (this.props.current_user.role === 0) {
```
AmitJoki: @ragesoss this checks for "Students"

app/assets/javascripts/components/overview/salesforce_link.jsx
```diff
@@ -16,6 +16,12 @@ const SalesforceLink = createReactClass({
     ServerActions.linkToSalesforce(this.props.course.id, salesforceId);
   },
 
+  updateSalesforceId() {
```
ragesoss: Unrelated change.

app/assets/javascripts/components/articles/available_article.jsx
```diff
@@ -60,16 +60,27 @@ export const AvailableArticle = createReactClass({
...
+        <td className="table-action-cell">
+          <button className="button dark" onClick={this.onRemoveHandler}>{I18n.t('assignments.remove')}</button>
+        </td>
+      );
+    } else if (this.props.current_user.isNonStudent && this.props.current_user.isStudent) {
+      actionCell = (
+        <td className="table-action-cell">
+          <button className="button dark" onClick={this.onSelectHandler}>{I18n.t('assignments.select')}</button>
+          <button className="button dark" onClick={this.onRemoveHandler}>{I18n.t('assignments.remove')}</button>
+        </td>
```
ragesoss: The logic looks right, but this could be done more clearly and with less duplication if the two buttons were set independently of each other. Maybe just leave the `td` there but empty if neither button is present?

app/assets/javascripts/components/overview/salesforce_link.jsx
AmitJoki: ugh! Mixed it up with salesforce id branch. Will remove.

AmitJoki:
Have made the requested changes!

ragesoss:
It's still `if ... else if ... else if`.

I meant that it should just have two independent `if` statements:

`if (isStudent) { set the select button }`
`if (isNonstudent) { set the delete button }`

test/components/articles/available_article.spec.jsx
```diff
@@ -7,7 +7,7 @@ describe('AvailableArticle', () => {
   const props = {
     course: { home_wiki: { language: 'en', project: 'wikipedia' } },
     assignment: { article_title: 'two' },
-    current_user: { role: 0 }, // student role
+    current_user: { isStudent: true }, // student role
```
AmitJoki: How do I do such a thing for `course_page_spec.rb:279`?

ragesoss: You create two CoursesUsers records, one for each role.

ragesoss: There shouldn't be any changes to the ruby specs needed, that I can think of.

AmitJoki: `course_page_spec.rb` was failing somehow.

AmitJoki: Is my code alright in the first place? With the replacing the whole thing with just two mutually exclusive if conditions?

ragesoss: Not sure. I'll check it out today.

AmitJoki: I think, I should just use role itself. 

if(role == 0)
  showSelect
if(role == 1 || course.user.admin)
  showRemove

AmitJoki:
@ragesoss it was some problem with `isStudent`,`isNotStudent`. I used the vanilla `course.current_user.role` and it worked out fine.

AmitJoki:
@ragesoss  The build is failing randomly, can you please rebuild this. Yesterday it was passing.

AmitJoki:
Build is successful :+1:

app/assets/javascripts/components/articles/available_article.jsx
```diff
@@ -60,15 +60,16 @@ export const AvailableArticle = createReactClass({
     const ratingMobileClass = `${ratingClass} tablet-only`;
     const articleLink = <a onClick={this.stop} href={article.url} target="_blank" className="inline">{article.formatted_title}</a>;
 
-    let actionButton;
-    // Show 'Select' button to students
+    let actionSelect;
+    let actionRemove;
     if (this.props.current_user.role === 0) {
```
ragesoss: Using current_user.role is not going to be compatible with showing both at once, except for admins.

If a user is enrolled with both the instructor and student roles in the same course, but is not an admin, then they should see both buttons. In that case, the role will probably be 1, but isStudent and isNonstudent should both be true.

AmitJoki: @ragesoss I got errors using `isStudent` and `isNonStudent` while in specs. That's why I backtracked on this.

PR #1790, opened 2018-03-12
details.jsx: made course details edit mode into a modal

ihsavru:
Converted the edit Details mode to a modal so that all the editable fields can be seen on one screen and doesn't stretch below the fold.
out-5
screenshot from 2018-03-12 23-09-05

### DIFF
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -19,6 +19,7 @@ import TimelineToggle from './timeline_toggle.jsx';
 import CourseLevelSelector from '../course_creator/course_level_selector.jsx';
 import HomeWikiProjectSelector from './home_wiki_project_selector.jsx';
 import HomeWikiLanguageSelector from './home_wiki_language_selector.jsx';
+import Modal from '../common/modal.jsx';
 
 import Editable from '../high_order/editable.jsx';
 import TextInput from '../common/text_input.jsx';
@@ -314,7 +315,7 @@ const Details = createReactClass({
       );
     }
 
-    return (
+    const shared = (
       <div className="module course-details">
         <div className="section-header">
           <h3>{I18n.t('application.details')}</h3>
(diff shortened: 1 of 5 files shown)


ragesoss:
@ihsavru cool! I think the centered titles above each input are awkward. Most of those could be inline, as the input fields don't need to be very wide.

psinghal20:
@ihsavru Please look into the Travis log for some linting issues in your code, some minor tweaks in the UI and we will be good to go. Also, Please ensure that all the feature specs related to the editable are passing now. I think there is a chance that some of them fail due to this new layout.

ragesoss:
The only feature spec that failed was the flaky 'admin adds tag' spec, which was probably a random failure (so I restarted the build). :-)

ihsavru:
@ragesoss I reduced the width of the modal and left aligned the labels.  I can make them inline if this still looks awkward

screenshot from 2018-03-13 00-21-24

ragesoss:
@ihsavru looks much better!

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -314,62 +315,97 @@ const Details = createReactClass({
       );
     }
 
-    return (
-      <div className="module course-details">
-        <div className="section-header">
-          <h3>{I18n.t('application.details')}</h3>
-          {this.props.controls()}
+    if (!this.props.editable) {
+      return (
```
ragesoss: Could this be done without duplicating all the content, by setting the shared content to a constant first, and then just switching on whether to wrap that content in `Modal`?

ragesoss:
@ihsavru the same spec failed again. Could you run it locally to see if it's passing for you?

`rspec ./spec/features/admin_role_spec.rb:111`

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -314,62 +315,97 @@ const Details = createReactClass({
...
+      <Modal>
+        <div className="module course-details modal-course-details">
+          <div className="section-header">
+            <h3>{I18n.t('application.details')}</h3>
+            {this.props.controls()}
+          </div>
+          <div className="module__data extra-line-height">
+            <Instructors {...this.props} />
+            <div className="details-form">
+              <div className="group-left">
```
ragesoss: Hmm... it's still a lot of duplication, I guess because of these extra divs. Would it work to keep the same arrangement in both modes, and just switch the classnames in edit mode?

psinghal20:
@ihsavru I can't find the option to add tags in the screenshots you have attached above. I think that is causing the failure of the Travis Build and the `admin_role_spec` at the example where the admin is trying to add the Tag for the course.

ihsavru:
@ragesoss reduced duplication by adding an additional class to the edit mode. Also, build passes now since I missed `{tags}` previously as @psinghal20 mentioned.

screenshot from 2018-03-13 13-10-00

ragesoss:
@ihsavru This modal shows up on top of the confirm modals that are used by some of the features with Details. For example, if you try to add another facilitator, the confirmation dialog is hidden behind this, and you won't see it until you exit Edit Details mode.

ihsavru:
Confirm modal as well as error notice comes on top now
screenshot from 2018-03-14 20-52-36
screenshot from 2018-03-14 20-52-39

Issue #1284, opened 2017-05-22
improve Uploads tab layout


ragesoss:
The Uploads tab currently acts as simply a recent uploads feed, and doesn't help much with either getting an overview of all contributed Uploads (where a justified gallery would fit many more images for a good overview) or evaluating uploads on a user-by-user basis. (Currently, you can sort by username, but can't easily get to the Commons upload page for a user, and also there's no visible grouping by user even when you've sorted by username.)

aSquare14:
I'd like to work on this !

ragesoss:
Great! Starting with wireframes might be helpful.

yunjuc:
I'll tackle on this as well.

yunjuc:
Here is what I have in mind - a grid view photo gallery with the following information:
- Description (instead of filename, which is less informative)
- Thumbnail (link to source page)
- Uploaded by
- Usage count (omit when usage is 0)

![upload](https://user-images.githubusercontent.com/18255878/36301146-ae5bd976-12b8-11e8-84bc-be2a811103f7.png)

When users click on author name, they will be taken to another page with the same layout, only the page title is slightly different. Users can go back to the previous page using the back button or the upload button.

![author](https://user-images.githubusercontent.com/18255878/36301167-b8525180-12b8-11e8-8f35-c943c5627e3b.png)

Thought?

ragesoss:
This is going in the right direction. I'm not sure about the separate view for a single user; that should definitely not be done in the initial implementation, but maybe something like that would make sense later. I think we'd want to do it as a filter on the same set of images, rather than a separate page.

For the layout, I think it's important to use as much space as possible; padding should be limited or removed. Justified gallery layouts are ideal for this kind of thing, in my opinion.

yunjuc:
Agree on the filter usage.

Also agree justified gallery looks better. So maybe mouseover to surface text and click linking to source page?

ragesoss:
Yeah, something like that might work nicely.

yunjuc:
Ok, I'll start working on this.

ragesoss:
We've made a lot of progress on this, and I think the design is in pretty good shape now.

PR #1797, opened 2018-03-14
Change upload table to photo gallery layout 

yunjuc:
This is an improvement of issue [#1284](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1284)

**Updates**
* "Uploads" table is replaced with photo gallery layout - mouse over will show image usage (no display if is 0), username, filename. No table header, need to add sort/filter functions to the top handler as next step **-- upload.jsx, upload_list.jsx, _uploads.styl --**
* Replace "Uploads" sorting option 'Filename' with "Date/time" - filename is usually random and doesn't provide much added value. Date/time is more meaningful. **-- upload_handler, jsx --**
* "Recent Uploads" table is replaced with the same layout - no sort/filter handler, need to add it as next stop. **-- upload_table.jsx --**

*Demo 1*
ezgif-2-cd2099c97e

*Demo 2*
ezgif-2-86e571e023

**Need Review/Discussion**
* **upload_list.jsx, upload_table.jsx** - some codes are commented out, need to check if they are not needed anymore.
* **upload_handler.jsx, recent_upload_handler.jsx** - some code may not needed anymore, need check and clean up.
* Do we want to show photo info (usage, username, filename) by default? - pro: information is available at first glance, no further step to reveal information, con - less interactive  (perhaps I can use another color layer for mouse over interaction)

*Show image info by default*
screen shot 2018-03-14 at 5 09 51 am


ragesoss:
This is coming along nicely!

I think the most useful way to handle the info will be to have it hidden by default, but have a toggle that would make it show for all images.

The layout crops some images, depending on the other items in that row and the dimensions of the image. Many end up with cropping on the top and bottom. I think the way Flickr does it, by varying the height of the rows when necessary to preserve the aspect ratios — see https://www.flickr.com/search/?text=panorama — is a better option. (I have no idea how easy/hard to implement it would be.)

photos

ragesoss:
BTW, I added a script to populate the database with some example data. If you pull down the latest version, you can run `rake dev:populate`, then `rake batch:update_constantly` and `rake batch:update_daily`. It takes a few minutes, and you should end up with most of these same photos in some example courses.

You may want to do `rake db:reset` first to clear out your database, if it's got enough courses that the updates take too long.

yunjuc:
I ran the rest -> populate -> update and get an empty dashboard. Do I miss anything? 

Also, I have one question about the batch update - In my local, are those two scripts run automatically in the background , or I have to manually run them to pull in new data?

yunjuc:
Instead of a having rectangle gallery container, I was playing with uneven width rows. I think it's rather interesting, or you think the rectangle shape is better? I slightly in favor this design but can do the other one if you think it's better.

For the hide/show info layer, I think having a toggle function would add burdens to the users, because It's not conventional, so it takes some efforts for users to discover and learn to use it. It also makes the handler section busy once we add sorting/filter there.

I tweaked the css from yesterday and now I have a new design to display a text overlay by default and remove it upon hover. This design can achieve several goals:
* It contains both visual effect and important sense
* Users can get a quick sense of what the photos are about by scanning through them
* The big "usage" message has the effect of inviting clicking and also encouraging users to upload images themselves.

_Demo3_
ezgif-2-46a493e971

ragesoss:
@yunjuc I like this new layout.

The primary function of this page is to let people quickly see all the images, either to just enjoy them or to look for particular ones of interest (either great photos, or to see which ones look like copyright violations that need to be looked into more). So I think defaulting to the info hidden (as in the current version I tried out locally) is preferable to the info UI you show here.

A secondary function is to look at metadata-related details, for example to see which of all of them are actually in use. That's not easy if you can only see one image's data at a time, but it's also an activity more for more advanced users with that particular purpose in mind.

I suggest starting with the show-on-hover UI, and we can worry about the secondary function later on.

ragesoss:
One thing you might consider to get most of that secondary function covered would be to add a much smaller indicator of usage... maybe even just something like a different border color for images that have at least one usage... so that you can quickly see the ones in use and then mouse over for the full details.

yunjuc:
Understood. Just push another code to change the row width elastic and hide the image info card.

ragesoss:
@yunjuc be sure to have a look at the failing `jest` test. (Run `yarn test` to run these locally.)

You can probably just delete the first failing test, as it's no longer relevant. The second might be easily fixable with a tweak to the test.

yunjuc:
Fixed.

test/components/activity/upload_table.spec.jsx
```diff
@@ -72,6 +69,6 @@ describe('UploadTable', () => {
     );
 
     const rowElements = TestTable.querySelectorAll('tr.upload');
-    expect(rowElements.length).to.eq(2);
+    expect(rowElements.length).to.eq(0);
```
ragesoss: This change means it's no longer testing what it is intended to test, which is that UploadTable will render elements for the uploads that are passed in as props. It's doing that by looking for `tr` nodes with the class `upload`, but since it's not longer rendered as a table with rows, you should change the selector to look for the new structure. Maybe that's just `div.upload` instead of `tr.upload`.

yunjuc:
I see, fixed again. 

I have one side question. In that test case above this, how is the <div><UploadTable {headers}> being used in the new layout?

ragesoss:
The `headers` props is not being used, as far as I can tell. You commented out the code that turned the headers prop into table header markup, which was previously being added to the top of the table in UploadTable.

yunjuc:
Should it fail when <gallery> is empty(no recent uploads)?

ragesoss:
Yes, but this spec is just testing the individual component, so it has whatever uploads are passed to it as props, which are defined at the top of the test file.

yunjuc:
I thought once the gallery is built, it's a success case. If upload has no data, there could be other messages on the page?

yunjuc:
I clean up upload_table header code and add checking gallery elements.

ragesoss:
@yunjuc Cool. Now that the javascript tests are passing, the full test suite runs on Travis. In particular, job 2 is the complete suite that includes the feature specs, which has two failing tests related to this PR.

The first one is searching for the text of an upload filename on the rendered page. It's failing because that text isn't visible, even though it's on the page. That should be solvable by changing the test to have it mouse over the image so that the text becomes visible. You can do this with `find('some.selectors').hover`.

The second spec fails because it's looking for a placeholder message when the Uploads tab renders with no uploads. To fix that, you should implement a 'no images' placeholder message.

PR #1811, opened 2018-03-17
Move mass enrollment functions to a service

### LINKED ISSUE #1796: Extract user registration and enrollment functions from MassEnrollmentController into a service
MassEnrollmentController has code to take a list of usernames and a course, and add all the users to the course, creating the users if they do not already exist in the database.

This should be extracted into a service object.

srish:
Fixes #1796. 

LMK, if this the correct approach!

### DIFF
--- a/app/controllers/mass_enrollment_controller.rb
+++ b/app/controllers/mass_enrollment_controller.rb
@@ -13,29 +13,9 @@ class MassEnrollmentController < ApplicationController
 
   def add_users
     @course = Course.find_by(slug: params[:course_id])
-
     usernames_list = params[:usernames].lines.map(&:strip)
-    @results = {}
-    usernames_list.each do |username|
-      @results[username] = add_user(username)
-    end
-
+    @results = AddUsers.new(course: @course, usernames_list: usernames_list).add_all_at_once
     UpdateCourseWorker.schedule_edits(course: @course, editing_user: current_user)
-
     render :index
-  end
-
-  private
-
-  def add_user(username)
-    user = find_or_import_user(username)
-    return { failure: 'Not an existing user.' } unless user
-    JoinCourse.new(course: @course, user: user, role: CoursesUsers::Roles::STUDENT_ROLE).result
-  end
-
-  def find_or_import_user(username)
-    user = User.find_by(username: username)
-    user = UserImporter.new_from_username(username, @course.home_wiki) if user.nil?
-    user
   end
 end
--- a/app/services/add_users.rb
+++ b/app/services/add_users.rb
@@ -1,0 +1,30 @@
+# frozen_string_literal: true
+
+class AddUsers
+  def initialize(course:, usernames_list:)
+    @course = course
+    @usernames_list = usernames_list
+  end
+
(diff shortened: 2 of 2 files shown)


app/services/add_users.rb
```diff
@@ -0,0 +1,32 @@
...
+    @current_user = current_user
+  end
+
+  def add_all_at_once
+    usernames_list = @usernames.lines.map(&:strip)
+    results = {}
+    usernames_list.each do |username|
+      results[username] = add_one_by_one(username)
+    end
+    UpdateCourseWorker.schedule_edits(course: @course, editing_user: @current_user)
```
ragesoss: This should be left in the controller. The job of the service will just be to add the users to the course; updating the on-wiki course is a separate idea, one that we carry out right afterwards when a user submits the username list through the UI, but which we might not want to do if we are calling this service in another context (like that setup script).

That will also remove the need to include current_user in the initializer.

app/services/add_users.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+class AddUsers
```
ragesoss: Here and a few other spots there is whitespace that the `rubocop` linter does not like. This is causing the build to fail. (I should mention rubocop in the PR template I just added.)

Depending on your code editor, you can probably install a rubocop plugin to see such things while you're editing. You can also run `rubocop` from the command line to see a list of linter violations.

app/services/add_users.rb
```diff
@@ -0,0 +1,32 @@
...
+
+class AddUsers 
+  def initialize(course:, usernames:, current_user:)
+    @course = course
+    @usernames = usernames
+    @current_user = current_user
+  end
+
+  def add_all_at_once
+    usernames_list = @usernames.lines.map(&:strip)
```
ragesoss: I think doing this in the controller makes a little more sense. That way, the interface to this service expects an array of usernames (rather than a multiline string), which will be more intuitive and easier to reuse.

The Rails convention is for controllers to handle user input, and service objects to carry out business logic.

app/controllers/mass_enrollment_controller.rb
```diff
@@ -13,29 +13,7 @@ def index
...
-
-    usernames_list = params[:usernames].lines.map(&:strip)
-    @results = {}
-    usernames_list.each do |username|
-      @results[username] = add_user(username)
-    end
-
-    UpdateCourseWorker.schedule_edits(course: @course, editing_user: current_user)
-
+    AddUsers.new(course: @course, usernames: params[:usernames], current_user: current_user).add_all_at_once
```
ragesoss: To keep the results behavior we currently have, you need to make `add_all_at_once` return the `results`, and assign that to `@results` here. (I guess that's what you were talking about in issue #1812?)

`@results` gets used by the corresponding view for this controller, to render the success/failure of each username submitted.

srish:
Thanks a lot for your detailed comments! I've made the changes. With these changes, specifically by returning `@results` in the controller, I can now view them on the UI and also realized there is no such bug that I mentioned in #1812.

PR #1824, opened 2018-03-17
Postioned diffViewer below Navbar to avoid overlapping

### LINKED ISSUE #1807: DiffViewer is partially hidden by course navbar
The addition of `z-index` 0 to the `course-main` class is making the DiffViewer render beneath the course nav.

For example, click one of the +/- buttons here: https://dashboard.wikiedu.org/courses/University_of_Southern_California/Social_Media_for_Scientists_and_Engineers_(Spring_2018)/articles

overlap

It should be like this:
correct

ihsavru:
Positioned the top of the diff viewer below the nav bar so it doesn't hide beneath it

screenshot from 2018-03-17 21-24-43

fix #1807

### DIFF
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -23,7 +23,6 @@ import Modal from '../common/modal.jsx';
 
 import Editable from '../high_order/editable.jsx';
 import TextInput from '../common/text_input.jsx';
-import Notifications from '../common/notifications.jsx';
 
 import DatePicker from '../common/date_picker.jsx';
 import CourseActions from '../../actions/course_actions.js';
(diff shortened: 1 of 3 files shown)


ragesoss:
Hmm... I don't like the solution of shifting the ArticleViewer (and DiffViewer, which is the one that comes up with the "+/-" icon) down beneath the course nav. Especially on short screens, that will make the article viewer pretty cramped.

If there's not a good way to keep the previous layout for these components, I think sticking with the extra `Notifications` component is fine.

ragesoss:
Per above, let's stick with the duplicate Notifications for now. Maybe eventually we can find a solution for the different stacking contexts of the nav vs. the components like DiffViewer that take over the screen.

PR #1825, opened 2018-03-18
Added filter for Alert Types

### LINKED ISSUE #1792: Improve campaign Alerts list
* Implement filtering by type, with a default filter that only includes ArticlesForDeletion and DiscretionarySanctions. User should be able to select as many types as they want to include in the filter.

ihsavru:
fix #1792
out-7

### DIFF
--- a/app/assets/javascripts/components/alerts/alerts_handler.jsx
+++ b/app/assets/javascripts/components/alerts/alerts_handler.jsx
@@ -4,15 +4,30 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 
 import AlertsList from './alerts_list.jsx';
-import { fetchAlerts, sortAlerts } from '../../actions/alert_actions';
+import { fetchAlerts, sortAlerts, filterAlerts } from '../../actions/alert_actions';
+import MultiSelectField from '../common/multi_select_field.jsx';
+
+const ALERTS = [
+  { label: 'ActiveCourseAlert', value: 'ActiveCourseAlert' },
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/common/filter.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import createClass from 'create-react-class';
+import PropTypes from 'prop-types';
+import Select from 'react-select';
+
+const MultiSelectField = createClass({
```
ragesoss: The component name should correspond to the file name.

app/assets/javascripts/components/alerts/alerts_handler.jsx
```diff
@@ -5,21 +5,49 @@ import { connect } from 'react-redux';
...
 
-
   propTypes: {
     fetchAlerts: PropTypes.func,
     alerts: PropTypes.array
   },
 
+  getInitialState() {
+    return {
+      selected: [
```
ragesoss: This is okay, but I think moving the filter state to redux would be better. It doesn't make a significant difference for this feature, since it's an isolate React component where state will be lost with the user navigates away. But once the whole of Campaigns gets migrated to React, having the filters in redux will mean that filter state gets retained when the user navigates away. It also keeps the architecture more consistent with the redux standard, where every significant change in application state is done via a redux action.

ihsavru:
@ragesoss renamed filter.jsx to multi_select_field.jsx and moved selected filters from component state to redux

ragesoss:
Looks good!

PR #1826, opened 2018-03-19
Migrate Upload store to redux

psinghal20:
Migrates Upload Store to redux.
Uses `sortByKey` method for sorting the uploads.
Screenshot: 
uploadstore


app/assets/javascripts/reducers/uploads.js
```diff
@@ -0,0 +1,36 @@
...
+  usage_count: true,
+  date: true,
+  uploader: true
+};
+
+export default function uploads(state = initialState, action) {
+  switch (action.type) {
+    case RECEIVE_UPLOADS: {
+      const dataUploads = action.data.course.uploads;
+      //Intial sorting by file name
```
ragesoss: Initial sorting should be by date.

app/assets/javascripts/reducers/uploads.js
```diff
@@ -0,0 +1,36 @@
...
+
+const initialState = {
+  uploads: [],
+  sortKey: null,
+};
+
+const SORT_DESCENDING = {
+  usage_count: true,
+  date: true,
+  uploader: true
```
ragesoss: Username should not be descending.

ragesoss:
This looks mostly good. Sorting by clicking the column header is broken.

The sorting for usage count is buggy: it should put all the entries with `null` count at the end, but they end up at the beginning. Also, Date sorting should be added to sorting pulldown. (Both of those bugs are already present in the current version, but now's a good time to fix them.)

app/assets/javascripts/utils/model_utils.js
```diff
@@ -28,10 +28,17 @@ export const sortByKey = (models, sortKey, previousKey = null, desc = false) =>
     newKey = sortKey;
   }
```
ragesoss: Documentation of this would be helpful.

psinghal20:
@ragesoss I have added the documentation and the failing Travis build seems to be a random failure as the superset build is passing.

PR #1832, opened 2018-03-21
Alter the student List - Making Header stick to the top

### LINKED ISSUE #1469: In mobile, making sure key information is accessible as the user scrolls
In the Course Editors Page, the header of the page should stick to the top while user is scrolling down -- to help them to understand the information being displayed.
**Current Version**
Before scrolling
scroll_stick2
After scrolling
52stick_row_tothetop_editorssection
**Improvement**
This feature would make it easy to understand the information displayed in the table.
**Development**
Work involves fixing the table header at the top when user scrolls it out of view.

sejalkhatri:
The header of the page sticks to the top when the user scrolls down -- to help them understand the information being displayed. Fixes #1469

- Added a StickyHeader bool PropType to List Component which can be set to true if this functionality is required.

[
stickyheader
](url)

### DIFF
--- a/app/assets/javascripts/components/common/list.jsx
+++ b/app/assets/javascripts/components/common/list.jsx
@@ -1,118 +1,162 @@
 import React from 'react';
+import createReactClass from 'create-react-class';
 import PropTypes from 'prop-types';
 import Loading from './loading.jsx';
 import UIActions from '../../actions/ui_actions.js';
 
-const List = ({
-  store,
-  keys,
-  sortable,
-  table_key,
-  className,
-  elements,
-  none_message,
-  loading,
-  sortBy
-}) => {
-  const sorting = store && store.getSorting();
-  const sortClass = (sorting && sorting.asc) ? 'asc' : 'desc';
-  const headers = [];
-  const iterable = Object.keys(keys);
+const List = createReactClass({
 
-  const sortByFunction = (tableKey, key) => {
-    if (sortBy) {
-      return () => {
-        sortBy(key);
-      };
+  propTypes: {
+    store: PropTypes.object,
+    keys: PropTypes.object,
+    sortable: PropTypes.bool,
+    table_key: PropTypes.string,
+    className: PropTypes.string,
+    elements: PropTypes.node,
+    none_message: PropTypes.string,
+    sortBy: PropTypes.func,
+    stickyHeader: PropTypes.bool
+  },
+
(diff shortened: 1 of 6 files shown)


ragesoss:
The positioning can shift when it switches into sticky mode.

Before sticking:
normal

After sticking:
offset

ragesoss:
Is it possible to do this with the `Affix` component, as we do with CourseNavbar in the `return` call in `course.jsx`?

sejalkhatri:
I don't think it's possible because Affix component is wrapped around a complete section which has to be fixed. 
In this particular case we don't want the entire List to be fixed but just the Table header.

app/assets/javascripts/components/common/list.jsx
```diff
@@ -19,18 +19,19 @@ const List = createReactClass({
...
+    if (persistAreaElements.length !== 0) {
+      for (let i = 0; i < persistAreaElements.length; i++) {
+        const persistHeaderRow = persistAreaElements[i].getElementsByClassName('persist-header');
+        const clonedHeaderRow = persistHeaderRow[0].cloneNode(true);
+        const style = window.getComputedStyle(persistHeaderRow[0]);
+        const width = style.getPropertyValue('width');
+        clonedHeaderRow.style.width = width;
+        clonedHeaderRow.classList.add("floatingHeader");
+        persistAreaElements[i].childNodes[0].insertBefore(clonedHeaderRow, persistHeaderRow[0]);
+      }
```
sejalkhatri: @ragesoss Is this what you were suggesting? I have refactored some of the code but I don't see a point why not to use JQuery!!

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -54,6 +54,7 @@ const DiffViewer = createReactClass({
 
   showDiff() {
     this.setState({ showDiff: true });
+    $('.floatingHeader').css({ opacity: 0 });
```
ragesoss: Eek! Any way to avoid putting workarounds in unrelated components?

sejalkhatri:
Changing z-index does not work because both the divs are in different stacking context. So, I have tried to fix the problem by changing the styling. Let me know if you can think of another workaround.

ragesoss:
That stacking context thing is no end of trouble! This is okay with me.

PR #1837, opened 2018-03-23
Uploads PR clean up

yunjuc:
This is the updated version of [ RR #1797](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1797)

Some changes since the last update:
* Add no data message on course uploads page
* Check image exists and hover text on course upload and activity upload page


app/assets/javascripts/components/uploads/upload.jsx
```diff
@@ -39,22 +23,22 @@ const Upload = ({ upload, linkUsername }) => {
     uploader = upload.uploader;
   }
 
+  let usage = '';
+  if (upload.usage_count) {
+      usage = `${upload.usage_count} usage${upload.usage_count !== 1 ? 's' : ''}`;
```
ragesoss: This should be done with an i18n message to which you pass the count. There's already a message, `uploads.usages`, which I'm not sure where it's used. You might be able to repurpose that and add pluralization support, or you might need to add a new message key.

app/assets/javascripts/components/uploads/upload.jsx
```diff
@@ -39,22 +23,22 @@ const Upload = ({ upload, linkUsername }) => {
...
-      <td className="desktop-only-tc">{moment(upload.uploaded_at).format('YYYY-MM-DD   h:mm A')}</td>
-      <td />
-    </tr>
+    <div className="upload">
+      <a href={upload.url} target="_blank">
+        <img src={imageFile} alt="" />
+        <div className="info">
+          <p className="usage"><b>{usage}</b></p>
+          <p><b>{fileName}</b></p>
+          <p className="uploader"><b>By <a>{uploader}</a></b></p>
```
ragesoss: `By` should be moved to an i18n message as well.

app/assets/javascripts/components/activity/upload_table.jsx
```diff
@@ -25,46 +20,20 @@ const UploadTable = createReactClass({
...
-            <th />
-          </tr>
-        </thead>
-        <tbody>
-          {uploads}
-        </tbody>
-      </table>
+      <div className="gallery">
+        {uploads}
+      </div>
```
ragesoss: I think you could replace the usage of `UploadTable` altogether, using `UploadList` in `RecentUploadsHandler instead of this component.

yunjuc:
Updates per change requests.

app/assets/javascripts/components/activity/recent_uploads_handler.jsx
```diff
@@ -2,16 +2,41 @@ import React from 'react';
...
 
-const HEADERS = [
-      { title: I18n.t('recent_activity.image'), key: 'image' },
-      { title: I18n.t('recent_activity.file_name'), key: 'file_name' },
-      { title: I18n.t('recent_activity.uploaded_by'), key: 'username', style: { minWidth: 160 } },
-      { title: I18n.t('recent_activity.usage_count'), key: 'usage_count', style: { width: 160 } },
-      { title: I18n.t('recent_activity.datetime'), key: 'uploaded_at', style: { width: 200 } },
-    ];
+
+export const UploadTable = createReactClass({
```
ragesoss: This component should have its own file.

yunjuc: So bring back the upload_table? 

I saw there is another `export const RecentUploadsHandlerBase` in the same file. It's legal because it's part of the `export default connect` while `UploadTable` is not?

Also, there is a prop `onSort={this.props.sortRecentUploads}` in the returned Upload table. This no sorting function right now. Is it for later? Should I keep it?

ragesoss: `RecentUploadsHandlerBase` was the only component defined in the file; while `export default connect()` returns a component, it doesn't call createReactClass or similar, so it doesn't trigger the eslint rule.

I missed this before, but the last change didn't fix the duplication issue. We should keep either UploadTable or UploadList — probably UploadList, for consistency with the similar ArticleList and RevisionList components — and use that component in both UploadHandler and RecentUploadsHandler.

The sorting function is working currently for the recent uploads tab: https://dashboard.wikiedu.org/recent-activity/recent-uploads

upload_actions.js has a similar sorting function that gets connected in UploadsHandler. With gallery design, there is no sorting to be done via table headers, so we'll need to add a sorting `select` that uses the sorting function in RecentUploadsHandler, like in UploadsHandler.

yunjuc:
Yarn test gives me an error `Invariant Violation: Element type is invalid:` when it tries to do `renders activities`, not sure how to resolve it.

yunjuc:
I think I can further merge upload and upload_list. Is there any pro and con to keep them separate vs. in one file?

ragesoss:
@yunjuc generally, we prefer separate components for discrete portions of the interface, such as the display of an individual upload. The main advantage of smaller components with limited responsibility is that it makes the app easier to reason about; if you want to change how a single image displays, you won't have to worry about its context in the list and how it interacts with the list functionality.

yunjuc:
I just realized I missed one rspec checker...

Thanks for the answer. I thought they can be combined since uploads doesn't share its layout with other pages. Now I see the component can be potentially used to check if a page has data as well. 

I removed sort by filename from both uploads and recent activity sorting list, because I don't see much use case for it. Hope it's okay.

ragesoss:
@yunjuc thanks! I just squashed and merged this as https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/d90b94788bdfab40bebae3db63b899481d0e58d3

Also made a few final tweaks:
* https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/939278eda750bf62802828e9873e58e8450a7bc3
* https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/8989e393054ec32b56b0954965c5f26a213b3afa

PR #1838, opened 2018-03-25
Add exec permission to setup/macOS-setup.sh

bekicot:
This will potentially remove the needs to change file mode at

https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/setup.py#L24
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/setup.py#L36
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/setup.py#L18


psinghal20:
Hi, I can't seem to see any changes in any of the files. Only the size of the macOS script is somehow different. Can you please elaborate what you mean by giving the exec permissions to the script?

bekicot:
@psinghal20 it is just running
`chmod x setup/macOS-setup.sh`

Issue #1802, opened 2018-03-15
Add ability to view and set SpecialUsers via /settings


ragesoss:
Some users have specific unique roles on the Dashboard, such as being the recipients of certain emails. These roles are stored in the `special_users` Setting record, and are accessible via the `SpecialUsers` presenter.

The super-admin-accessible `/settings` page should display these special users and allow them to be set or changed.

bekicot:
I'm currently working on this, do you mind to assign this to me?

ragesoss:
@bekicot we don't usually use the 'assign' feature, but thanks for taking this on!

ragesoss:
done!

PR #1839, opened 2018-03-25
Add special_users endpoint on settings namespace

bekicot:
Add special users endpoint on special name space.

Related to #1802 

This is the endpoint to list all of the special users. It will be required when creating the view on the `/settings` page.


ragesoss:
This looks reasonable. I'll wait to merge it until the frontend part of things is ready as well.

app/assets/javascripts/components/settings/containers/special_user_container.jsx
```diff
@@ -0,0 +1,4 @@
+import { connect } from 'react-redux';
+import SpecialUser from '../views/special_user';
+
+export default connect()(SpecialUser);
```
ragesoss: This is unnecessary. `connect` isn't doing anything here; you may as well just use SpecialUser directly in SettingsHandler.

bekicot: The file was intended as a placeholder, we will need it to add a new action. I think I should make a complete implementation as you suggested.
this implementation is not obvious.

PR #1845, opened 2018-04-02
Link Training Libraries index directly to individual trainings

### LINKED ISSUE #1330: Training Libraries index should have links directly to individual trainings

ihsavru:
Fix #1330
screenshot from 2018-04-03 00-22-13


ragesoss:
This looks good except for the active link styling, which includes a white line from the module list.

active link border

ihsavru:
@ragesoss i removed the outline around the headers

PR #1850, opened 2018-04-07
Add pagination for campaign articles

### LINKED ISSUE #1847: Articles tab for campaigns may time out if there are too many articles
We need to add pagination... maybe 1000 per page.

ihsavru:
added pagination for 1000 articles per page using "will_paginate" gem
fix #1847

out-8


ragesoss:
This is great! So simple!

Issue #1842, opened 2018-03-28
Replace jquery-uls with a simpler dropdown language selector


ragesoss:
We're using a forked version of the Wikimedia jquery universal language selector to let users choose their locale. However, the global jquery dependency is not ideal, we can't easily update it because it relies on bootstrap class names that we changed in our fork, and we don't really need the full complexity of it. A simpler language dropdown selector would fine, and would let us cut out the separate jquery-uls.js script.

Something like the pageviews tool uses, including the link to translatewiki, would do nicely: https://tools.wmflabs.org/pageviews/?project=en.wikipedia.org&platform=all-access&agent=user&range=latest-20&pages=Cat|Dog

srish:
I would like to work on this issue, and I've started to understand the code flow a bit.

srish:
I am a bit stuck on this issue! So, following the same approach as Pageviews does, I considered writing a function in `/language_switcher.js`, make it extract all available languages (from `/i18n/ `) and then populate them in the drop-down selector. As a first step, I tried using the npm package glob. But, I am running into the following error: `Module not found: Error: Can't resolve 'fs' in '../Projects/WikiEduDashboard/node_modules/glob'
`

I'm wondering:
* First, is this the right approach I am following?
* If so, have we encountered such errors in the past while using a package that has fs dependency? What am I missing here?

ragesoss:
@srish if you push your WIP to github, I can take a look.

There will be many ways of doing this. 

To recap the overall set of things going on:
* language_switcher.js uses jQuery to load and configure the language switcher and define the behavior when a language is selected. This is included in main.js, so it happens once when the page is loaded.
* uls_box.jsx is a React component that reloads/reconfigures it every time time it gets re-rendered. I believe this is a workaround so that it is able to work from within the React-based navbar, even if the page gets resized and the navbar switches from desktop to mobile layout.
* nav.jsx is the navbar, and it controls whether to include the uls-trigger button to which the language switcher attaches. (Language switcher is not present on the Wiki Education configuration.)

The approach I suggest would be:
* remove language_switcher.js and uls_box.jsx altogether
* create a new component, like `language_picker.jsx`, and insert that in place of the uls-trigger button in nav.jsx.
[...]

PR #1855, opened 2018-04-17
[WIP] Language selector

srish:
Through the changes in this pull request, we are replacing the use of `jquery-uls` with a simple dropdown language selector as used by the [Pageviews tool here](https://tools.wmflabs.org/pageviews/?project=en.wikipedia.org&platform=all-access&agent=all-agents&range=latest-20&pages=Cat|Dog). Attempts to fixing #1842 

screen shot 2018-04-16 at 8 35 26 pm

I would like to get some feedback on my code, perhaps how to improve it. More specifically, I need help in modifying the test cases in `spec/features/language_switcher_spec.rb.` I tried replacing the CSS `.uls-trigger.` which is in there with the new class name `.language-picker.` I tried a bunch of different combinations in Capybara to find the element, but I couldn't. I've been getting "Element not found" error every single time.


app/assets/javascripts/components/language_picker.jsx
```diff
@@ -0,0 +1,63 @@
...
+      const nativeName = iso.getNativeName(code);
+      if (nativeName && (nativeName !== enN && nativeName !== esN && nativeName !== frN)) {
+        return { label: nativeName, value: code };
+      }
+      return undefined;
+    });
+
+    const defLocales = _.without(allLocales, undefined);
+    const newLocales = translateLink.concat(popularLocales).concat(defLocales);
+    const curLocale = <span><img src="../../assets/images/icon-language.png" alt="Translate this page" /> {I18n.locale} </span>;
```
ragesoss: This image is not found for me. Removing the ../.. fixed it.

package.json
```diff
@@ -110,6 +111,7 @@
     "react-redux": "^5.0.6",
     "react-router": "3.2.0",
     "react-select": "^1.0.0-rc.10",
+    "react-select-plus": "^1.2.0",
```
ragesoss: This should also include a change to `yarn.lock`.

If you run `yarn` locally, it should add the exact versions of these dependencies to the lock file.

ragesoss:
I haven't looked over the component code closely yet, but on a quick overview it's looking good.

A few styling aspects have room for improvement. In particular, the select is separate from the current locale and image:
* There is no clear visual connection between the unbordered image/locale and the select box that appears beneath it.
* You can scroll the page to make the image/locale disappear, but the select does not scroll away.

I guess this is a combination of the behavior of react-select-plus and the styling you added to it.

There's also a visual artifact... maybe a border for zero-width selected element? ... that appears for me on Firefox whenever I activate the Select. It's the two dots that appear to the left of the picker image.
language picker

The tests are failing because the nav component is not rendering at all in the feature specs. (Screenshots, like the one below, should appear in `tmp/screenshots` whenever feature specs fail.) I don't know why, as there are no obvious problems or javascript errors, and it works as expected outside of the tests. I'll look into this more when I have a chance, but it may not be until I get back from Berlin.

screenshot_2018-04-18-19-24-51 856

srish:
@ragesoss The select was separate from the current locale and image previously because I override the default styles to make the element on the nav look similar to the previous one (uls-trigger). But, I guess that wasn't the right approach. 

I've made some changes that fixed the UI concerns you have addressed in your comment. 

Though with not overriding the styles, the Select seems to take up a bit of space and comes with a border and arrow like in: 
screen shot 2018-04-24 at 9 14 51 pm

Would this work?

ragesoss:
I think it looks fine without overriding the styles, and that fixes the scroll-related problem as well. Lemme see if I can figure out where the tests are going wrong...

ragesoss:
Okay, I'm closer now! I noticed that near the end of `gulp build` output, there was an error message:

```
ERROR in main.3f9ebedf52a5663c116a.js from UglifyJs
Unexpected token: name (ISO6391) [main.3f9ebedf52a5663c116a.js:148860,6]
```

I gather that this is because Uglify is incompatible with ES6 syntax, and the `iso-639-1` module uses some of that.

I adjusted LanguagePicker to hard-code the language names instead of using the iso-639-1 module, and this allows the feature specs to run without error.

I found that upgrading the webpack Uglify plugin to the latest version (instead of the one that is bundled with webpack 3.*) silences the error, but the tests still fail in the same way (not rendering any of the Javascript-based components). Switching to a different minification plugin (babel-minify) also silences the error, but it takes 5 minutes to minify the code (instead of a few seconds) and also does not fix the specs.

My next thought would be to try just importing the LANGUAGE_LIST (https://github.com/meikidd/iso-639-1/blob/master/src/data.js) from the module, and reimplementing `getNativeName()`.

ragesoss:
Another option might be to generate the language-code-to-name mapping server-side, using https://github.com/alphabetum/iso-639

ragesoss:
Copying over the language list from the npm module worked well: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/a43cde15613adf5f66be12bf3b3953fead55303c

The build is still failing, but it's just the language switcher feature spec now. It'll probably just require some tweaks to the spec file to interact with the new language picker.

One other tweak would be to point the 'translate' link to the dashboard page on translatewiki: https://translatewiki.net/wiki/Translating:Wiki_Ed_Dashboard

ragesoss:
@srish is there any reason why we need `react-select-plus` instead of `react-select`? It looks like this doesn't use the 'groups' feature that the `-plus` for adds, and we already have `react-select` elsewhere in the project. If we don't need the `-plus` fork, that's one less dependency to worry about.

srish:
@ragesoss thank you for your help! I couldn't have figured out that issue :P I've made changes to the test suite `language_switcher_spec.rb` (all cases pass) and a couple others that you've addressed above! Some tests were failing locally for me still, but with a msg that they are the expected ones.

ragesoss:
It's getting close to ready! Making a new branch was not necessary, but perfectly fine. To be continued on #1860

PR #1859, opened 2018-04-24
Fix #1856 Campaign Alert Table Not Showing under Mobile Layout

### LINKED ISSUE #1856: Campaign Alerts list is not shown on mobile
When it switches to mobile layout, the list just disappears: https://outreachdashboard.wmflabs.org/campaigns/artfeminism_2018/alerts

wendyyuchensun:
Fix [#1856](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1856)

**The bug**

The article column in alert table has a `desktop-only-tc` class, which hides the whole column under mobile layout.

Since article column should be the only thing alert tables shows under mobile layout, that class makes the table looks like 'totally disappeared' when switches to mobile layout.

**What I did**

* Remove the problematic `desktop-only-tc` class of article column.
* Also add right border to that column under mobile layout.

**Result of fix (viewport width < 920px)**

screen shot 2018-04-24 at 9 53 28 pm

Let me know if there is any problem with this PR. Thanks for reviewing!

### DIFF
--- a/app/assets/javascripts/components/alerts/alert.jsx
+++ b/app/assets/javascripts/components/alerts/alert.jsx
@@ -6,7 +6,7 @@ const Alert = ({ alert }) => {
   return (
     <tr className="alert">
       <td className="desktop-only-tc date">{moment(alert.created_at).format('YYYY-MM-DD   h:mm A')}</td>
-      <td className="desktop-only-tc">{alert.type}</td>
+      <td className="alert-type">{alert.type}</td>
       <td className="desktop-only-tc"><a target="_blank" href={`/courses/${alert.course_slug}`}>{alert.course}</a></td>
       <td className="desktop-only-tc"><a target="_blank" href={`/users/${alert.user}`}>{alert.user}</a></td>
       <td><a target="_blank" href={alert.article_url}>{alert.article}</a></td>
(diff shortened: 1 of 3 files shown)


ragesoss:
@wendyyuchensun cool! I can merge this now if you like, but if you want to take it further, it would be ideal to include more info in the mobile layout. The most useful thing would be to at least show the alert type.

wendyyuchensun:
Hi @ragesoss, I've added a new commit to include the type column in alert table in mobile layout. 

screen shot 2018-04-25 at 9 31 33 am

However, CI fails. I think it's related to the current version of Node (10.0.0, which just released), not this PR. I reproduced same error showed in CI logs in my local environment using Node 10.0.0. Using Node 8.11.1 (LTS version) works fine. 

Seems like someone bumped into same problem: https://github.com/gulpjs/gulp/issues/2162

ragesoss:
@wendyyuchensun thanks! I'll look into the node/gulp issue soon, probably tomorrow. Did you try upgrading gulp to 4.0.0?

wendyyuchensun:
Not yet. I'll see if I have time to try it out recently. I'll share the result here when I'm done.

ragesoss:
The nodejs 10 compatibility is sorted out now... lemme give this a try.

ragesoss:
Nice work @wendyyuchensun. Thanks!

PR #1860, opened 2018-04-28
[WIP] Language Selector v2

srish:
Finishing bits and pieces for issue #1842, previously addressed in the request https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1855. 

Test suite `language_switcher_spec.rb` now works. Few minor changes include:
* Uses module `react-select` in place of `react-select-plus`
* New translate wiki link for text 'Help translate' in dropdown 
* Additional style changes to `.language-picker` so that the element appears on the right on small screens

Though with this initial push, some test are failing locally with a message "Failures listed here are expected and do not affect your suite's status"


ragesoss:
On CI, one of the specs is still failing. I had trouble with that one too; I think the main problem is the inaccuracy of PhantomJS's rendering. I worked around it [like this](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/c15ba349952029ffb63896c7cd717915363a18a3#diff-60efbe0eff76a8bb4ccebad992e8a3e9R38), setting the width to 3000.

Not sure whether that will be a stable solution or whether we'll need to add that to the "Failures listed here are expected" set.

One other thing I forgot to point out that I think we should change is the handling of available locales that don't have a name; this includes all codes that are not just two letters. Falling back to the code in those cases will be an easy first step, eg, [like this](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/c15ba349952029ffb63896c7cd717915363a18a3#diff-e5f5561b5321df856335f782fb815b11L44).

srish:
@ragesoss Oops! My bad. I didn't realize you were working on the spec in the other branch. Resizing the window to what I had previously, saved me from errors locally, but looks like it needed more increment for CI.  I've pushed a couple of changes from your commit as you suggested above!

ragesoss:
@srish this looks good to me. I can merge it if there's nothing else you're planning to change.

One area for improvement, which we could just file a new issue for, is adding names for all the available locales that are missing from the data copied over from the iso-639-1 module.

srish:
@ragesoss merging it and opening another issue for adding names for missing locales sounds good to me!

ragesoss:
@srish one more thing I missed! The styling is broken for the 'training' layout; it uses a separate main stylesheet, which doesn't include all the style modules.

A start to fixing it will be to add these to `training.styl`:

```
@import "modules/_language_picker"
@import "modules/_react_select"
```
There's a bit more to be done after that, to fix the alignment and the text color on mouseover.

training picker styling

ragesoss:
I finished up the styling bit and merged this. Thanks @srish! I'll likely deploy this today.

srish:
thanks @ragesoss! :)

PR #1871, opened 2018-05-17
[WIP] Basic prototype for Article Finder Tool

psinghal20:
It adds a basic prototype of the new Article Finder Tool implementing client-side data fetch operations.


psinghal20:
Hi @ragesoss, I have added the incorporated redux hooks with each query. Please take a look. I have currently tested it with the help of `Redux DevTools`. I will add the list component ASAP.

app/assets/javascripts/reducers/article_finder.js
```diff
@@ -0,0 +1,69 @@
...
+export default function articleFinder(state = initialState, action) {
+  switch (action.type) {
+    case CLEAR_FINDER_STATE: {
+      return {
+        articles: [],
+        loading: false
+      };
+    }
+    case RECEIVE_CATEGORY_RESULTS: {
+      let newStateArticles = state.articles.map(article => ({ ...article }));
```
ragesoss: You should be able to do `newStateArticles = { ...state.articles }`. With that pattern, unchanged `article` objects get reused, which should prevent unnecessary rerendering of ArticleFinderRow.

app/assets/javascripts/utils/article_finder_utils.js
```diff
@@ -0,0 +1,82 @@
+import logErrorMessage from './log_error_message';
+
+const pageviewBaseUrl = 'https://wikimedia.org/api/rest_v1/metrics/pageviews/per-article/en.wikipedia/all-access/user/';
```
ragesoss: This should not be en.wikipedia specific. We might roll this out for just en.wikipedia to start with, but we should write to only keep the specific wiki in one place (like the top ArticleFinder component) so that it''ll be easy to adapt for other langauges.

app/assets/javascripts/reducers/article_finder.js
```diff
@@ -20,6 +20,7 @@ export default function articleFinder(state = initialState, action) {
         newStateArticles[data.title] = {
           pageid: data.pageid,
           ns: data.ns,
+          fetchState: 1
```
ragesoss: Don't do this with magic numbers. You can add all the states to a constants file and use the named constants instead. Otherwise, every person who reads any of the fetchState-related code will need to refer to each of these places in the reducer to understand what the fetchState values are supposed to mean.

app/assets/javascripts/constants/article_finder.js
```diff
@@ -0,0 +1,14 @@
...
+export const RECEIVE_ARTICLE_REVISION = 'RECEIVE_ARTICLE_REVISION';
+export const RECEIVE_ARTICLE_REVISIONSCORE = 'RECEIVE_ARTICLE_REVISIONSCORE';
+
+export const fetchStates = {
+ TITLE_RECEIVED: 1,
+ PAGEASSESSMENT_RECEIVED: 2,
+ REVISION_RECEIVED: 3,
+ REVISIONSCORE_RECEIVED: 4,
+ PAGEVIEWS_RECEICED: 5
+};
```
psinghal20: @ragesoss I added this object to map fetchState to an integer for comparing. Redux state will save the keys of this object so that it is easy for anyone to understand the code.

app/assets/javascripts/constants/article_finder.js
```diff
@@ -4,3 +4,9 @@ export const RECEIVE_ARTICLE_PAGEVIEWS = 'RECEIVE_ARTICLE_PAGEVIEWS';
 export const RECEIVE_ARTICLE_PAGEASSESSMENT = 'RECEIVE_ARTICLE_PAGEASSESSMENT';
 export const RECEIVE_ARTICLE_REVISION = 'RECEIVE_ARTICLE_REVISION';
 export const RECEIVE_ARTICLE_REVISIONSCORE = 'RECEIVE_ARTICLE_REVISIONSCORE';
+
+export const TITLE_RECEIVED = 1;
+export const PAGEASSESSMENT_RECEIVED = 2;
+export const REVISION_RECEIVED = 3;
+export const REVISIONSCORE_RECEIVED = 4;
+export const PAGEVIEWS_RECEICED = 5;
```
ragesoss: typo in this name (also copied everywhere it's used)

Issue #1861, opened 2018-05-01
Add language names for available locales that lack them


ragesoss:
We just switched to a simple dropdown language picker in place of the jquery universal language selector. To provide language names that correspond to the locale codes, we started from an iso 639-1 data set that has names for all of the two-letter language codes: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/utils/language_names.js

For the other available locales (with three letter codes or hypenated codes), the picker currently just shows the code. We should extend that language_names file to cover all the locales that we have available.

PR #1872, opened 2018-05-19
[WIP]Solved #1861 Added language names for available locales that lack them

jerry-74:
Added missing language codes (with three letter codes or hypenated codes) to language_names.js


db/schema.rb
```diff
@@ -10,9 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
```
sejalkhatri: @jerry-74 These changes are not related to this Pull request, So its suggested to revert back to the original file.

app/assets/javascripts/utils/language_names.js
```diff
@@ -573,10 +697,18 @@ const LanguageNames = {
       name: 'Sinhala',
       nativeName: 'සිංහල',
     },
+//    'simple_form.en': { // does it have to be there?
```
ragesoss: We have a `simple_form.en` locale file, but it's not for Simple English. I think it's used by a dependency called 'simple_form', and was generated automatically when that was installed. But it doesn't show up as an available locale and we don't need it in this file.

app/assets/javascripts/utils/language_names.js
```diff
@@ -729,14 +893,35 @@ const LanguageNames = {
...
+      name: 'Simplified Chinese',
+      nativeName: '中文（简体）‎',
+    },
+    'zh-hant': {
+      name: 'Traditional Chinese',
+      nativeName: '中文（繁體）‎',
+    },
+    'zh-TW': {
+      name: 'Traditional Chinese',
+      nativeName: '中文（繁體）‎',
```
ragesoss: It looks like zh-TW and zh-hant use the same name and native name, and the same is true for zh-CN and zh-hans. However, they have separate translation files (and separate sets of translators).

But even in the comments from translatewiki.net, the language code is the only thing that shows they are different. Both zh-hant and zh-TW say "Messages for Traditional Chinese (中文（繁體）‎)", for example.

After this change, the language picker component will show two entries in a row for `中文（繁體）` and for `中文（简体）`.  I would love some advice from @amire80 or @Nikerabbit or another i18n expert on this. I guess for most people it won't be a problem, because the language code gets selected via browser preferences, so users from Taiwan should get zh-TW by default.

There are a few other cases with the same native name, such "ug-arab" vs "ug", but in those cases, we only actually have one of the versions available (only ug-arab, for example), so it won't cause any problems.

jerry-74: I used the names from the comments you mentioned but I crosschecked them with data on the Wikipedia. Regarding the Chinese names, I have found right now this page https://en.wikipedia.org/wiki/Chinese_Wikipedia where the names are described better, I hope. Shall I replace them?

app/assets/javascripts/utils/language_names.js
jerry-74: removed in the last checkout

app/assets/javascripts/utils/language_names.js
ragesoss: @jerry-74 I guess you mean the names from the table in the Solutions section? That seems like a good solution to me. I'm sure someone will report it if it doesn't make sense. Please do replace those, and then I'll merge.

ragesoss: I guess those names from the table correspond to the part in parentheses? So cn-TW would be `中文（台灣正體）`, and so on, I think. That seems consistent with how Google Translate interprets it: Chinese (Taiwanese).

jerry-74: Done. See my last commit. Hoping it will be ok.

ragesoss:
Thank you! I'll likely deploy this today.

PR #1874, opened 2018-05-20
[WIP]Implement a justified image gallery

ihsavru:
...and added timestamps to the image info
screenshot from 2018-05-20 20-44-26

### DIFF
--- a/app/assets/javascripts/components/uploads/upload.jsx
+++ b/app/assets/javascripts/components/uploads/upload.jsx
@@ -3,12 +3,19 @@ import PropTypes from 'prop-types';
 
 const Upload = ({ upload, linkUsername }) => {
   const fileName = upload.file_name;
-
   let imageFile;
   if (upload.deleted) {
     imageFile = '/assets/images/deleted_image.svg';
   } else {
     imageFile = upload.thumburl;
+  }
+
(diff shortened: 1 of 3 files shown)


ragesoss:
This looks great. The build failed because of a rubocop violation: the json.builder file has a line that's now too long.

I'll try this out later today to make sure it works with a wide variety of images, and it can probably be merged and deployed later today.

ihsavru:
@ragesoss i just figured there's a flaw with this since i'm using `thumbwidth` and `thumbheight` but a lot of images have that set as null, i'll set this as a WIP

ragesoss:
@ihsavru I already merged it, so if you have a fix we can do it in the next PR.

I'm not noticing any problems of images that aren't loading in production. However, the justified layout doesn't work out for some series of images. See https://dashboard.wikiedu.org/recent-activity/recent-uploads

staggered
staggered layout

ihsavru:
@ragesoss they aren't justified because of what I mentioned above, it fails in case when `upload.thumbwidth` or `upload.thumbheight` are null. 
Regarding the images that weren't loading in production was due to the `upload.thumburl` being null for some images so there were just empty `<div></img>` blocks

PR #1876, opened 2018-05-24
Add Pagination and change timestamp format on Uploads

ihsavru:
screenshot from 2018-05-24 21-37-35

### DIFF
--- a/app/assets/javascripts/components/uploads/uploads_handler.jsx
+++ b/app/assets/javascripts/components/uploads/uploads_handler.jsx
@@ -31,17 +31,15 @@ const UploadsHandler = createReactClass({
     this.setState({ data: data, pageCount: nextProps.uploads.length / this.state.perPage });
   },
 
-  setUploadData() {
-    const data = this.props.uploads.slice(this.state.offset, this.state.offset + this.state.perPage);
-    this.setState({ data: data });
+  setUploadData(offset) {
+    const data = this.props.uploads.slice(offset, offset + this.state.perPage);
+    this.setState({ offset: offset, data: data });
   },
 
   handlePageClick(data) {
     const selectedPage = data.selected;
     const offset = Math.ceil(selectedPage * this.state.perPage);
-    this.setState({ offset: offset }, () => {
-      this.setUploadData();
-    });
+    this.setUploadData(offset);
   },
 
   sortSelect(e) {


ihsavru:
Pagination for 100 images per page
screenshot from 2018-05-25 14-58-29

app/assets/javascripts/components/uploads/uploads_handler.jsx
```diff
@@ -13,10 +14,37 @@ const UploadsHandler = createReactClass({
...
+  setUploadData() {
+    const data = this.props.uploads.slice(this.state.offset, this.state.offset + this.state.perPage);
+    this.setState({ data: data });
+  },
+
+  handlePageClick(data) {
+    const selectedPage = data.selected;
+    const offset = Math.ceil(selectedPage * this.state.perPage);
+    this.setState({ offset: offset }, () => {
+      this.setUploadData();
```
psinghal20: Is there a specific reason you are using a callback to update the state for new `data`?

package.json
```diff
@@ -109,6 +109,7 @@
     "react-dropzone": "^4.2.1",
     "react-motion": "^0.5.2",
     "react-onclickoutside": "^6.6.2",
+    "react-paginate": "^5.2.3",
```
psinghal20: You didn't commit the `yarn.lock` file for this change.

app/assets/javascripts/components/uploads/upload.jsx
```diff
@@ -1,29 +1,40 @@
...
+      this.setImageDimensions();
+    });
+  },
+
+  setImageDimensions() {
     const img = new Image();
-    img.src = imageUrl;
+    img.src = this.state.imageFile;
     const component = this;
     img.onload = function () {
```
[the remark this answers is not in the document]
ihsavru: Nope, the requests are only made once as far as I can tell by looking at the Networks tab in Dev tools, the images are cached most probably.

app/assets/javascripts/components/uploads/uploads_handler.jsx
ihsavru: Yes, or else there's an unwanted effect, the `setUpload` executes before the state is actually set

package.json
ihsavru: Oh, my bad. I was under the impression that I don't have to commit `yarn.lock` file  :sweat_smile:

app/assets/javascripts/components/uploads/uploads_handler.jsx
psinghal20: Well actually, I thought you could set the state with only one call to `setState` with `offset` and `data`. Just compute the `data` array after you calculate the `offset`  and then pass them together to `setState`. I just thought right now, this will rerender the component by setting the state again in the callback.

ihsavru: Surely, I can pass offset to `setUploadData` and then set the state together after computing the data.

PR #1882, opened 2018-06-01
UI changes for Article Finder

psinghal20:
**Screenshots**
image
image

@ragesoss I was also considering adding a range input field for `Max Completeness` field. I was thinking of using a React component I found online [react-input-range](https://github.com/davidchin/react-input-range). I am not sure how much it will affect our bundle size but it will indeed be fit for our use if we go with a range type input.


psinghal20:
**Screenshot**
simpler UI with combined article quality filter
image

PR #1884, opened 2018-06-04
[WIP] Add keyword based search for article finder

psinghal20:
It adds the keyword based article search with the help of Mediawiki search API and removes the default sorting based on page views.


psinghal20:
**Screenshots**:
image
image
image

Issue #1853, opened 2018-04-13
When stats for a course are updated, there should be a visual indication of the update


ragesoss:
Some courses — editathons while they are going on, in particular — automatically refresh the stats while the home tab is open.

It would be nice to add some subtle visual indication when the stats change. I'm not sure what that would look like exactly. Maybe some CSS animation of the stats that fades out within a second or two. (It should only happen if the stats actually change, not when the stats get refreshed but did not change.)

yunjuc:
How can I simulate the data update? I tried to add new editors to a course and ran `rake batch:update_constantly` but nothing changed?

ragesoss:
@yunjuc there are two parts to this.

On the server, the stats need to actually change; that normally only happens if there are additional Wikipedia edits that happened since the previous update, but you can simulate it by directly modifying the `Course` record in the rails console. For example, `Course.last.update(article_count: 101)`.

On the browser side, if the 'home' tab of a course is open, it will request new course stats from the server every 5 minutes. You can change the interval to something short for testing in the `details.jsx` component. Then, shortly after you do an update to the `Course` record in the rails console, you should see the change in the browser.

ragesoss:
@yunjuc `update_constantly` is a little different than it was a few months ago; it's no longer what fetches new data for courses. The easiest way to do that for a single course is to add `/manual_update` to the end of the base URL. The way we do it now on the server is via `rake batch:schedule_course_updates`; that puts all the courses that need to be updated into the Sidekiq job queue, but those jobs won't run unless you also start the right sidekiq workers to process them. (This reminds me that the sidekiq rake task needed an updated to account for this new course update architecture, which I just pushed a fix for.)

yunjuc:
Ok, I'll try that. I'm also trying to find a way to re-format 1000 to 1k on the frontend. Right now it's done with `number_to_human` but the 'k' can't be animated because it's not a number.

yunjuc:
Also, the `byte_added` count doesn't format the number so it runs long and overlaps with next number sometimes, like this one:
https://outreachdashboard.wmflabs.org/courses/Wikidata/Wikidata_WikiProject_India

ragesoss:
@yunjuc the campaigns are currently all done server-side, but change the numbers for one of the courses should be all it takes to change the campaign numbers.

yunjuc:
[...]

PR #1885, opened 2018-06-04
[WIP] Move number formatting to client side

yunjuc:
Fixes for #1853 

I tried a couple of approaches to do this:
1. There is a jquery method can create number animation, I add some code so the animation and number formatting can happen at the same time. This works on campaign stats now.
2. I found a [react component](https://www.npmjs.com/package/react-countup) can create number animation, but it can't do number formatting and these two actions need to happen together. If we want to have both animation and the format, I'll need to dig into the source code and change it.

I read about using jquery in react is not a good idea. Now I have a numbers.js holds the animation code (a jquery method) and number formatting function. What I want to do is to add a script link on the head so it can select the numbers and run the code, similar to i18n, but only on the course and user profile pages.

For maintenance point, which one will be the better way to do it?

### DIFF
--- a/app/assets/javascripts/components/overview/course_stats.jsx
+++ b/app/assets/javascripts/components/overview/course_stats.jsx
@@ -1,6 +1,8 @@
+import CountUp from 'react-countup';
 import React from 'react';
 import PropTypes from 'prop-types';
 import CourseUtils from '../../utils/course_utils.js';
+//import numFormat from '../../numbers.js';
 
 const CourseStats = ({ course }) => {
   let viewData;
@@ -11,7 +13,9 @@ const CourseStats = ({ course }) => {
   if (course.home_wiki.language === 'en') {
     contentCount = (
       <div className="stat-display__stat" id="word-count">
-        <div className="stat-display__value">{course.word_count}</div>
+        <div className="stat-display__value">
+          <CountUp stat={0} end={course.word_count} duration={2} />
+        </div>
         <small>{I18n.t('metrics.word_count')}</small>
       </div>
     );
--- a/app/assets/javascripts/numbers.js
+++ b/app/assets/javascripts/numbers.js
@@ -1,0 +1,34 @@
+//select number elemets from DOM and add animation
+$('.stat-display__value').each(function () {
+  $(this).prop('Counter', 0).animate({
+    Counter: $(this).text()
+  }, {
+    duration: 1000,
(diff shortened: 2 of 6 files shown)


ragesoss:
@yunjuc thanks! I'll take a look at this tomorrow.

One quick thing: the number formatting and the update animation should be done i separate PRs.

For the animation, I had imagined something that could be implemented just with CSS, so that adding a certain class (via React) would trigger the animation. I'm not an expert with CSS animations and maybe that's not practical, but I think it's the ideal in terms of maintainability, separating the logic part of it — in this circumstance, trigger some sort of visual effect — and the actual visual effect styling. I hadn't thought of animating the actual increase, like the `react-countup` package you found. I like that concept, although it sounds like it's not a good fit for any numbers that we want to treat with some kind of 'for humans' reformatting. The core of the feature will just be to draw the eye whenever stats do update, so that (for example) if a course page is being projected on a screen at an event, there will be some kind of change that lasts for a few seconds and increases the chances that people will notice that the numbers changed.

yunjuc:
My bad. I tried another way approached to indicate the data update - to add and remove elements from the dom, something like this. 
https://jsbin.com/tocokop/edit?html,js,output

I use Mutation Observer to watch the dom, so I need to add a js script link to the course page. How can I do that?

I'll open another PR from the data update visual cue, and keep this one as number formatting change.

yunjuc:
Since now I move to another direction on the animation so it's not number depended. Is there still a need to move the number formatting to frontend? What's the benefit of it?

ragesoss:
@yunjuc I think the benefit of moving the number formatting to the frontend is pretty minimal. For React views where the HTML is completely rendered client-side, it's a little nicer to have everything related to localization and formatting also done client-side, just to consolidate display-related logic in one place (and to make the API data a bit more usable for potential 3rd-party consumers: if we're providing numbers directly from the API, they'll be more useful as real numbers than strings like `'100k'`).

The easiest fix for the too-large bytes-added figures would be to format those server-side, but it means adding more display logic to the server (so it's basically adding a small amount of technical debt for when we eventually straighten that out and move the display logic to the client). But until we have a solution that brings all the for-humans formatting together in one place, it's probably better to just do the easiest thing that will work.

I did find this project — https://github.com/emcien/number-helpers-coffeescript — which is aimed at the problem we have, but it doesn't support locales, unfortunately.

PR #1887, opened 2018-06-05
[WIP] Add spinner animation to course overview dashboard

yunjuc:
Fixes for #1853

The animation is added to the course dashboard. To watch the data change on the DOM, `statsUpdate.js` file needs to add to the template. Not sure how to do it yet, looks like a gulp + webpack task, any pointer is appreciated.

The same feature can be added to the campaign dashboard too once it's done on the course page.

### DIFF
--- a/app/assets/javascripts/components/overview/course_stats.jsx
+++ b/app/assets/javascripts/components/overview/course_stats.jsx
@@ -6,20 +6,28 @@ const CourseStats = ({ course }) => {
   let viewData;
   let infoImg;
   let trainedTooltip;
-
   let contentCount;
+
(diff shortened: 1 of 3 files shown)


ragesoss:
I went a different direction for #1853. Now that the main `course` object is going through redux, I added code to track which stats have updated whenever course data gets refreshed, and the specific stats that changed will show a brief animation.

yunjuc:
Thanks for closing this and the other notification PR. I've been busying with some works this summer and also studying up React and JS. I'll get back here when I get more time. Also feel free to ping me if there is something you think I can help with. Cheers!

PR #1889, opened 2018-06-06
Fix bytes-added number format issue

yunjuc:
Fix bytes-added number format issue.

### DIFF
--- a/app/views/courses/course.json.jbuilder
+++ b/app/views/courses/course.json.jbuilder
@@ -25,7 +25,7 @@ json.course do
   json.trained_count @course.trained_count
   json.word_count number_to_human @course.word_count
   json.view_count number_to_human @course.view_sum
-  json.character_sum number_to_human @course.character_sum_human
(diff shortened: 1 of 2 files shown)


app/views/courses/course.json.jbuilder
```diff
@@ -25,6 +25,7 @@ json.course do
   json.trained_count @course.trained_count
   json.word_count number_to_human @course.word_count
   json.view_count number_to_human @course.view_sum
+  json.character_sum number_to_human @course.character_sum
```
ragesoss: This overwrites the raw `character_sum` value from the `json.call` above.

I think would be better to keep the raw `character_sum`, and call this `character_sum_human` or similar, instead. To make that work as intended, we would also need to change the `CourseStats` component to show the new value.

yunjuc: Updated. If that overwrites the original value, what it doesn't matter to other data?

PR #1894, opened 2018-06-11
[WIP] Integrate Article finder with home wiki

psinghal20:
This PR integrates the home wiki language and course of a course with the article finder and adds a loader to show the status of fetching.


ragesoss:
The ArticleFinderRow still treats everything as en.wiki, even if the search results come from a different language.

psinghal20:
@ragesoss missed those changes earlier. Updated the PR, please review. Thanks!

ragesoss:
Looks good. If you start noticing nice ways to refactor the actions file, that would be good to do soon. It's getting complicated enough that it will be much more understandable and maintainable if the logic of *what* to do gets separated from the code for *how* to do it. One option might be to extract the queries themselves into their file.

PR #1897, opened 2018-06-13
[WIP] add uploads count for course users

ihsavru:
The tooltip description could be better and need to add translations
screenshot from 2018-06-13 19-43-57

### DIFF
--- a/app/models/courses_users.rb
+++ b/app/models/courses_users.rb
@@ -88,6 +88,7 @@ class CoursesUsers < ApplicationRecord
 
   def update_cache
     revisions = live_revisions
+    self.total_uploads = course.uploads.where(user_id: user_id).count
     self.character_sum_ms = character_sum(revisions, Article::Namespaces::MAINSPACE)
     self.character_sum_us = character_sum(revisions, Article::Namespaces::USER)
     self.character_sum_draft = character_sum(revisions, Article::Namespaces::DRAFT)
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -934,6 +934,7 @@ en:
     revisions_doc: The number of edits made by this editor in the past 7 days
     reviewers: Reviewer(s)
     reviewing: Reviewing
+    total_uploads: "Total Uploads"
     training_complete:
       one: "is up-to-date with training"
       other: "are up-to-date with training"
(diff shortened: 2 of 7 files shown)


app/models/courses_users.rb
```diff
@@ -86,6 +86,10 @@ def live_revisions
     course.revisions.joins(:article).where(user_id: user_id).live
   end
 
+  def total_uploads
+    course.uploads.where(user_id: user.id).count
```
ragesoss: Doing it this way will result in one extra database query for each user, whenever the users.json for a course is requested.

To avoid the performance impact, the next steps should be:

1. Generate a migration to add `total_uploads` to the `courses_users` database table.
2. Move this query into the update_cache method, so that instead of doing the query on each read, it gets done once and stored as an attribute along with the other stats.
3. Check whether all of the database indexes to make the query as performant as possible are already present, and add them if necessary.

config/locales/en.yml
```diff
@@ -950,6 +951,8 @@ en:
     training_module_name: Training module
     training_module_status: Status
     up_to_date_with_training: 'up-to-date with training'
+    uploads_doc: "The total number of uploads a user has added to Wikimedia
```
ragesoss: Maybe `Number of file uploads to Wikimedia Commons`.

We should either avoid the term 'course', or move this message into the `courses` section with an alternative `courses_generic` version that uses the term 'program' instead. For any message that uses terminology specific to students and courses, we always need alternative messages that don't use school-focused terms.

ragesoss:
This looks good. I haven't tested it out locally yet; I will on Monday.

The one thing missing is the schema.rb change. (It should just change the version number at the top, and add a line for the new column; if you run `rake db:migrate` and more than that gets changed, you can manually fix it to just change the two lines. I can also add this to the commit when I test it out locally, if you don't get to it before then.)

db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_06_04_211453) do
+ActiveRecord::Schema.define(version: 2018_06_15_155436) do
```
ragesoss: Did you run another migration after the one included in this PR? The version timestamp will normally match the filename of the latest migration that's been run on the DB, so locally I get `2018_06_13_173211` after running the migration.

ragesoss:
Merged as https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/1bd3dab3abfaeac7c8cd78b731b00a44e76ed15d

PR #1898, opened 2018-06-13
[WIP]Clean up UI of Article Finder

psinghal20:
- Clean up the search bar
- Add show filter button
- Add sorting icons for table headers

**Screenshots:**
image
image
image

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -28,6 +28,7 @@ const ArticleFinder = createReactClass({
   getInitialState() {
     return {
       isSubmitted: false,
+      showFilters: false,
     };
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -73,22 +80,22 @@ const ArticleFinder = createReactClass({
         value_key="search_term"
         required
         editable
-        label={I18n.t('article_finder.category')}
-        placeholder={I18n.t('article_finder.category')}
+        label="Search"
```
ragesoss: These changed interface messages should be done in the locale files.

psinghal20: Done!

Issue #1877, opened 2018-05-25
Data download options should include aggregate CSV of edits for a campaign


ragesoss:
We have the option to download all the edits for a course, via the 'Download Stats' modal. A similar option for an entire campaign (like the others currently accessible from /campaigns) would be useful.

robotscissors:
@ragesoss I am a new contributor, I would like to work on this.

I may have a few follow-up questions :)

ragesoss:
@robotscissors cool! Ping here, ask on Slack, for any questions you have.

robotscissors:
Hi @ragesoss,
Having trouble with the installation process. I am almost there, but need a nudge.

OS: Mac Sierra
I have installed on two separate machines.


Used two methods (manual and script)
- Succeeded in getting through mysql installation.
- Rails s - server starts nicely 
- localhost:3000 produces the following error:
No such file or directory @ rb_sysopen - /Users/christopher/work/bloc/wiki-dashboard/public/assets/stylesheets/rev-manifest.json
This error occurs when the asset build process has not generated the required rev-manifest.json files, which specify the filenames of the compiled stylesheet and javascript files.
Run `gulp` or `gulp build` and make sure there are no build errors.

**Running `gulp` or `gulp build` produces these errors:**
Let's go (~/work/bloc/wiki-dashboard$) > gulp build
[09:20:18] Failed to load external module @babel/register
[09:20:18] Failed to load external module babel-register
[09:20:18] Failed to load external module babel-core/register
[09:20:18] Failed to load external module babel/register
[09:20:18] Local gulp not found in ~/work/bloc/wiki-dashboard
[09:20:18] Try running: npm install gulp

Then, I burst in tears repeatedly.

ragesoss:
@robotscissors try `npm install -g gulp` (or `sudo npm install -g gulp`), then re-run `gulp build`.

robotscissors:
@ragesoss, I appreciate the attention.

`sudo npm install -g gulp` then `gulp build` 

Yielded this:

Let's go (~/work/bloc/wiki-dashboard$) > gulp build
[10:18:48] Failed to load external module @babel/register
[10:18:48] Failed to load external module babel-register
[10:18:48] Failed to load external module babel-core/register
[10:18:48] Failed to load external module babel/register
/Users/christopher/work/bloc/wiki-dashboard/gulpfile.babel.js:1
[...]

PR #1900, opened 2018-06-17
Fixing issue 1877 download csv for campaigns

robotscissors:
## Before the Pull Request ##
* Tested manually
* Ran the Javascript tests locally (`yarn test`) and confirmed that they are passing.
* Ran the specs for any Ruby files you changes, along with the feature specs that are related to your changes. 

* Added a link in the campaign view to download all **edits app/views/campaigns/index.html.haml**
* Updated routing to accommodate new view and action
* Created a new lib-analytic-campaign-csv-builder: lib/analytics/campaign_edits_csv_builder.rb 
* This address issue #1877

### DIFF
--- a/spec/lib/analytics/campaign_edits_csv_builder_spec.rb
+++ b/spec/lib/analytics/campaign_edits_csv_builder_spec.rb
@@ -6,7 +6,14 @@ require "#{Rails.root}/lib/analytics/campaign_edits_csv_builder"
 describe CampaignEditsCsvBuilder do
   let(:campaign) { create(:campaign) }
   let(:course) { create(:course) }
-  let(:edits) { CampaignEditsCsvBuilder.new(campaign).generate_csv }
+  let(:user) { create(:user) }
+  let(:article) { create(:article) }
+  let!(:revision) { create(:revision, article: article, user: user, date: course.start + 1.hour) }
+  before do
+    campaign.courses << course
+    create(:courses_user, course: course, user: user)
+  end
+  let(:edits) { described_class.new(campaign).generate_csv }
 
   it 'creates a CSV with a header' do
     expect(edits).to include("revision_id")
@@ -21,4 +28,12 @@ describe CampaignEditsCsvBuilder do
     expect(edits).to include("dashboard_edit")
   end
 
+  it 'creates a CSV with a header and a row of data for a course revision' do
+    expect(edits.split("\n").count).to eq(2)
+  end
+
+  it 'creates a CSV with a header and a row of data for a course revision' do
+    expect(edits).to include(course.title)
+    expect(edits).to include(article.title)
+  end
 end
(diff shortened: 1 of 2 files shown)


spec/lib/analytics/campaign_edits_csv_builder_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require "#{Rails.root}/lib/analytics/campaign_edits_csv_builder"
+
+describe CampaignEditsCsvBuilder do
+  let(:campaign) { create(:campaign) }
+  let(:course) { create(:course) }
```
ragesoss: This spec isn't being run with any edits. You can add a user to the course along with an edit or two, following the example of the CourseEditsCsvBuilder spec, so that it exercises the `row` method.

app/controllers/campaigns_controller.rb
```diff
@@ -176,6 +177,15 @@ def articles_csv
     end
   end
 
+  def all_edits
```
ragesoss: This should have a test.

spec/lib/analytics/campaign_edits_csv_builder_spec.rb
robotscissors: Understood. I'll get to it.

app/controllers/campaigns_controller.rb
robotscissors: Great feedback. I will also work on this.

spec/lib/analytics/campaign_edits_csv_builder_spec.rb
robotscissors: HI @ragesoss, I added a couple of tests as requested. I am having trouble adding more than one row but I did add one row and checked for content and total number of rows in CSV. Let me know if this is acceptable.

ragesoss:
@robotscissors tests look good, and cover all the new code. The rubocop violations are still there, and causing the build to fail. If you don't get to it first, I can clean those up when I test it out manually before merging.

lib/analytics/campaign_edits_csv_builder.rb
```diff
@@ -0,0 +1,51 @@
...
+    diff
+    username
+    bytes_added
+    new_article
+    dashboard_edit
+  ].freeze
+  def row(revision)
+    row = [revision.mw_rev_id]
+    row << @campaign.title
+    row << @course.title
```
ragesoss: Instead of the title, this should be `@course.slug`. Titles may not be unique, but slugs will be.

lib/analytics/campaign_edits_csv_builder.rb
```diff
@@ -0,0 +1,51 @@
...
+    row << revision.date
+    row << revision.wiki.base_url
```
ragesoss: `revision.wiki.domain` would be better, I think, so that the useful info (the language and project) is easier to scan.

PR #1906, opened 2018-06-21
Add switchable views for uploads

ihsavru:


ragesoss:
Looks like this breaks some of the `jest` tests.

Issue #1878, opened 2018-05-25
Notification banner doesn't display properly on the Training page


yunjuc:
On program & event dashboard, the notification banner is off place and doesn't display properly on the Training page.
https://outreachdashboard.wmflabs.org/training

ragesoss:
This is because the training layout uses a completely separate stylesheet. Copying over the relevant styles from the main css to one of the training css files will probably fix it.

yunjuc:
How can I create a notification?

RDenPen:
I am fairly new to open source, and am currently working on this issue. I've been trying to modify the _training.styl files in favor of the notification problem but I am having trouble seeing any changes take effect when I run tests

yunjuc:
You mean you can't see the notification or you can't see the style changes you made to training page? 

If it's former, you need to uncomment the sitenotice in ```config/application.yml``` file to enable the notification. 

If it's later, _training.styl should be right file.

ragesoss:
Changes to a `.styl` file will only show up after the CSS has been recompiled, which happens via `gulp`. If you aren't seeing the expected changes in your development environment, make sure `gulp` is running in a terminal. On each save, you should see messages related to recompiled files.

PR #1911, opened 2018-06-24
Add css style to training notification. Add spacing between components.

yunjuc:
Fix for #1878

Add style to notification on training page and increase the spacing between modules.

screen shot 2018-06-24 at 12 11 46 am
    
    
Global change to increase the spacing between sections.

screen shot 2018-06-23 at 11 05 12 pm

### DIFF
--- a/app/assets/stylesheets/modules/_tables.styl
+++ b/app/assets/stylesheets/modules/_tables.styl
@@ -20,6 +20,7 @@
 .table
   width 100%
   border-collapse separate
+  margin 2% 0
 
   thead,
   tbody
--- a/app/assets/stylesheets/_layout.styl
+++ b/app/assets/stylesheets/_layout.styl
@@ -72,7 +72,6 @@ footer
 
 section
   position relative
-  margin 2% 0 2% 0
 
 .details
   padding-top 20px
--- a/app/assets/stylesheets/_layout.styl
+++ b/app/assets/stylesheets/_layout.styl
@@ -44,7 +44,7 @@ footer
   .container
     padding-bottom 10px
 
-.footer, .push 
+.footer, .push
   height 30px
 
 .section-header
@@ -54,15 +54,15 @@ footer
   h3
     color body-color
 
-  .section-header__actions  
-    // 
-      
+  .section-header__actions
+    //
+
   +above(tablet)
     padding 15px 0 0
-    
-    .section-header__actions  
+
+    .section-header__actions
       position absolute
-      right 0 
+      right 0
       top 20px
 
   +above(desktop)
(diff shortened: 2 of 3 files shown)


ragesoss:
This adds some margins in places that shouldn't have them. For example, below the nav bar:

unwanted margin

app/assets/stylesheets/modules/_tables.styl
```diff
@@ -20,6 +20,7 @@
 .table
   width 100%
   border-collapse separate
+  margin 2% 0
```
ragesoss: @yunjuc can you explain more about what this is for, and perhaps illustrate it?

yunjuc: It adds 2% margin on the top and bottom of a table, and 0 margin on the right and left. Therefore, the spaces between a table and the elements above and after it become wider.

ragesoss: This is not necessary for fixing the training notification issue, right?

I browsed a little bit, and in many cases I think this extra space spreads things out too much. It's worthwhile to consider what global layout adjustments might improve usability (and maybe my first impression of this one is off base), but I'd like to have a clear understanding of the rationale, and something like that should be done in a separate PR.

yunjuc: Yes, this is one separate from the notification issue and I should do it on sperate PR.

The original thought of the fix is I saw the campaign text is right after the previous table and I want to give it some breathing space. It's enclosed in section tag so I first added the margin on the section (see dom samples below). It caused the unexpected space because the purple welcome section use a background color and it's only seen when you're not logged in, so I missed that. 
<body>
   ....
   <section>
     <table></table>
  </section>
   ...  
</body>

Later I changed it to add margin to the table tag. I can do it either only on this specific table vs. a global change on all table. The reason I did it globally is to make the style consistent, so when a table is created, it has the same spacing as other tables. And the trade-off is the added space you see, and I think some extra space to make each element more distinct from each other is not a bad thing, as long as it's not an awkward gap. Thoughts?

On a side note, I'm having a `ERR_TOO_MANY_REDIRECTS` issue and can't open localhost these two days. There are a lot of rollback and 302 redirect on guard, and I tried to clear the cookies but that doesn't help.

ragesoss: No, I've never seen that before on localhost. Usually when I've run into that in other places, it's because of the browser caching bad 302 routes. The way to clear redirect caches varies by browser, but try it's worth a try. As a first step, see if the works properly in a different browser; if so, it's probably cached redirects.

yunjuc: I'll try to resolve this. Thanks.

ragesoss:
Thanks again @yunjuc! I copied the notificaton and the library margins into another commit that I just merged.

PR #1914, opened 2018-06-26
add relevance sorting and make it default

psinghal20:
- add relevance sorting
- add sort select dropdown


ragesoss:
Location of the sorting pulldown is unexpected:

sorting location

I think this would work better if the relevance index was explicitly listed in the table (eg, in the first column before title), so that it's a little more clear what's going on.

psinghal20:
@ragesoss Removed the sort select component and added the column with relevance index.
image

Issue #1852, opened 2018-04-13
Replace multiendpoint CoursesController#show action with separate actions for each


ragesoss:
A high proportion of traffic goes through the the `show` action, which is responsibile for variably rendering many different course-specific JSON views (see `SHOW_ENDPOINTS`).

Performance monitoring tools treat these all as the same endpoint, so it's not easy to single out performance problems with a single view.

Pulling the whole set of `show` endpoints to a separate namespaced sub-controller, and then making explicit actions for each one would probably be the best way to fix this. See also #1417

jstoebel:
I'm going to take a stab at this one

PR #1930, opened 2018-07-01
[WIP] Iss1852

jstoebel:
I am still working on this PR for #1852 and am running into a number of errors where I can't see how they're related to my code. For example, I am getting a number of errors like the following

```
  1) AssignmentsController DELETE #destroy when the user owns the assignment when the assignment_id is provided destroys the assignment
     Failure/Error: @templates ||= YAML.load_file(Rails.root + template_file_path)
     
     Errno::ENOENT:
       No such file or directory - bs_fetch:open_current_file:open
     # ./app/models/wiki.rb:116:in `edit_templates'
     # ./lib/wiki_course_edits.rb:23:in `initialize'
     # ./app/controllers/assignments_controller.rb:61:in `new'
     # ./app/controllers/assignments_controller.rb:61:in `remove_assignment_template'
     # ./app/controllers/assignments_controller.rb:25:in `destroy'
     # ./spec/controllers/assignments_controller_spec.rb:46:in `block (5 levels) in <top (required)>'
```

Any ideas why that would be?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -88,7 +88,7 @@ class ApplicationController < ActionController::Base
 
   def course_slug_path(slug, args={})
     slug_parts = slug.split('/')
-    show_path(args.merge(school: slug_parts[0], titleterm: slug_parts[1]))
+    courses_path(args.merge(school: slug_parts[0], titleterm: slug_parts[1]))
   end
   helper_method :course_slug_path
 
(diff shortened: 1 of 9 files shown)


ragesoss:
One possibility is that an unexpected ENV value is causing this. What do you have for `dashboard_url` in you application.yml, and when was the last time you rebased on upstream/master? I can poke at this next week, but see if rebasing fixes it. If so, it's probably because the test expected `dashboard.wikiedu.org` as the url but your environment had something different. We merged a fix for that (by setting the url in the test.rb environment file) a few days ago.

It looks like that spec passes on CI.

PR #1932, opened 2018-07-03
Create Syllabus controller

NickSchimek:
Moved syllabus controller and associated method to its own controller
file.

Moved specs for controller syllabus to its own file.

Updated syllabus controller route. Action name is now 'update'.
Routing endpoint remains the same.

This helps to keep controller's small and manageable.

This starts to address issues referenced in #1417

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -61,6 +61,8 @@ Metrics/ModuleLength:
 
 Layout/SpaceAroundEqualsInParameterDefault:
   Enabled: false
+Style/ClassAndModuleChildren:
+  Enabled: false
 Style/ColonMethodCall:
   Enabled: false
 Style/NumericLiterals:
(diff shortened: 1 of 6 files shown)


NickSchimek:
1 rubocop failure
```
app/controllers/courses/syllabuses_controller.rb:2:7: C: Style/ClassAndModuleChildren: Use nested module/class definitions instead of compact style.
class Courses::SyllabusesController < ApplicationController
```

nested style is default, but compact style is supposedly good too.
https://rubocop.readthedocs.io/en/latest/cops_style/#styleclassandmodulechildren

what do you think?

ragesoss:
Yes, compact style is fine. You can disable that in the robocop config.

ragesoss:
Nice, thanks!

PR #1938, opened 2018-07-05
July5 timeline edits

cassiville:
Rearranged some blocks on the Timeline so the Assignment module always appears before an Exercise or Discussion module. Made one copy-edit Week 1.


ragesoss:
already merged

PR #1945, opened 2018-07-07
Add credit for each upload

ihsavru:
peek 2018-07-07 22-12

### DIFF
--- a/app/assets/javascripts/reducers/uploads.js
+++ b/app/assets/javascripts/reducers/uploads.js
@@ -7,7 +7,6 @@ const initialState = {
   sortKey: null,
   view: GALLERY_VIEW,
   selectedFilters: [],
-  fetchState: false,
 
 };
 
@@ -34,7 +33,6 @@ export default function uploads(state = initialState, action) {
         ...state,
         uploads: sortedModel.newModels,
         sortKey: sortedModel.newKey,
-        fetchState: false,
       };
     }
     case SET_UPLOAD_METADATA: {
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/reducers/uploads.js
```diff
@@ -1,11 +1,14 @@
...
+import _ from 'lodash';
+import { RECEIVE_UPLOADS, SORT_UPLOADS, SET_VIEW, GALLERY_VIEW, FILTER_UPLOADS, SET_UPLOAD_METADATA } from '../constants';
 import { sortByKey } from '../utils/model_utils';
 
 const initialState = {
   uploads: [],
   sortKey: null,
   view: GALLERY_VIEW,
   selectedFilters: [],
+  fetchState: false,
```
ragesoss: `fetchState` is not very descriptive, especially if it's a boolean. Since this gets used elsewhere, naming it in a way that clearly communicates *which* fetch it's tracking state for would be good. Maybe `metadataFetched`?

app/assets/javascripts/reducers/uploads.js
```diff
@@ -31,6 +34,35 @@ export default function uploads(state = initialState, action) {
...
+        if (data.query) {
+          fetchedData = { ...fetchedData, ...data.query.pages };
+        }
+      });
+      const updatedUploads = state.uploads.map(upload => {
+        if (fetchedData && fetchedData[upload.id]) {
+          if (fetchedData[upload.id].imageinfo[0].extmetadata.Credit) {
+            upload.credit = fetchedData[upload.id].imageinfo[0].extmetadata.Credit.value;
+          }
+          if (!fetchedData[upload.id].imageinfo[0].extmetadata.Credit) {
```
ragesoss: This would be clearer and less repetitive as an `else`, since we'll always either find a credit or set the fallback.

ihsavru: I replaced this with lodash's  _.get to avoid checking for every key if it exists

PR #1950, opened 2018-07-11
add articles navbar in articles handler

psinghal20:
Screenshot:
image


ragesoss:
The underlines after each item don't look great. Also, looks like we have room to shrink the navbar so that the tables have more horizontal space.

psinghal20:
@ragesoss made the suggested changes!
Screenshot:
image

ragesoss:
In that first screenshot, and replicated locally for me, the Articles Edited table is narrowing than it should be. (This only happens if the table is not empty.) It looks like this is caused by an empty `th` in the table header.

ragesoss:
I'm not sure why that empty `<th />` is there in the render for `List` anyway, but removing it seems not to cause any problems. I guess it's there because some (formerly all?) tables have a final column without a header, for the 'open drawer' icon.

Also, on scroll, I get lots of `TypeError: this.refs.articlesEdited is undefined`.

psinghal20:
@ragesoss 
- Used hash locations similar to the timeline component
- Added a `onUpdate` function in `router.jsx`, for scrolling to the hash locations.
- Used redux UI state for the scroll workaround, found some predefined actions in `index.js`, should I move those to the new `ui_actions_redux.js` file?
- Should I create redux selectors for *Assigned Articles* and *Available articles*? Currently, both are part of a single object and differentiated by `used_id`.

ragesoss:
I get `ReferenceError: types is not defined` when I click 'Add available articles'.

Also, the switching from article finder to one of the articles lists and back is very sluggish, at least on the course I was testing it with. I wonder if this is because of mounting and unmounting a table with hundreds of items. One option that might help would be to replace the separate ArticleFinder route with a CSS `hidden` class to hide and unhide the articles tables and article finder, so that they would all stay mounted in the same component. That would be more consistent with the UI, and avoid the full-page rerendering when the route changes.

Issue #1926, opened 2018-06-29
'Expected editors' count in Details should only be shown for ClassroomProgramCourse


ragesoss:
The course creator does not collect the expected editor count for other course types, and we forgot to limit it to only the relevant course type within `Details`. Other fields are already conditonal on course type, we just need to add expected_editors to that set of fields.

khyatisoneji:
@ragesoss , I would like to work on this issue, can you please give information about the relevant files and the URL that displays course information.

ihsavru:
@khyatisoneji The file you're looking for is `details.jsx` in the overview folder in components, it is responsible for displaying the course details which you can view on the 'Home' tab of a course. There are other examples in that component where certain fields are displayed conditionally to a program type. Goodluck!

khyatisoneji:
@ihsavru @ragesoss I got the details that were mentioned here, I was looking at the props that are passed to the component and I saw that the prop expected_editors is not being passed on to the component. Also, there is no mention of expected_editors anywhere in the code. Please guide me as to from where shall this field be arriving.

sejalkhatri:
@khyatisoneji You are looking for expected_students which is analogous to expected_editors. The value depends on the course type, expected_students for the `ClassroomProgramType` and  expected_editors for `BasicCourse` type. We use [Rails internationalization](http://guides.rubyonrails.org/i18n.html) in the project, you can view that up in `config/locales/en.yml file`.   Also, expected_students can be accessed from the `course` prop  in the `details.jsx` file

khyatisoneji:
@sejalkhatri , so I can use expected_students as expected_editors, right? Do I need to consider any other fields in expected_editors?

sejalkhatri:
@khyatisoneji Those conditions are already handled through internationalization. 
You have to add a condition for 'Expected students' count in Details. `expectedStudents` element should be shown only for the ClassroomProgramCourse. For eg: (Line 315) Condition for `courseLevelSelector`, it is shown only for ClassroomProgramCourse.

Ankitanvesh:
@sejalkhatri I have read the comments above and added the condition for expectedStudents element to be shown only for the ClassroomProgramCourse. I wanted to know if there is anything else that needs to be done in regards to this issue.

sejalkhatri:
@Ankitanvesh No! Nothing else is to be done.

PR #1959, opened 2018-07-26
'Expected editors' count in Details should only be shown for ClassroomProgramCourse #1926

Ankitanvesh:


Ankitanvesh:
@ihsavru thanks for the suggestion. I will do it

Ankitanvesh:
Fixes #1926

ragesoss:
Thanks!

PR #1961, opened 2018-07-30
add recent uploads on user profiles

ihsavru:


app/assets/javascripts/components/user_profiles/course_details.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+const CourseDetails = createReactClass({
+  propTypes: {
+    courses: PropTypes.array
+  },
+
+  render() {
+    const elements = this.props.courses.map(course => {
+      return (
+        <a className="course" key={course.id} href={`/courses/${course.course_slug}`}>
```
ragesoss: The `course` object here is not shaped the same way as the ones from course pages; it's actually derived from a CoursesUsers record, like this:

```
course_id: 201
course_school: "test"
course_slug: "test/Sci_Comm_brochure_yos_(now)"
course_term: "now"
course_title: "Sci Comm brochure yos"
user_count: 1
user_role: "Student"
```

To ensure unique keys, it should be something like:
```
`${course.course_slug}-${course.user_role}`
```

app/assets/javascripts/components/user_profiles/user_profile.jsx
```diff
@@ -0,0 +1,64 @@
...
+
+
+const UserProfile = createReactClass({
+  propTypes: {
+    params: PropTypes.object,
+    fetchStats: PropTypes.func.isRequired,
+    stats: PropTypes.object.isRequired,
+    isLoading: PropTypes.bool.isRequired
+  },
+
```
ragesoss: I had to add a getInitialState() to set an empty object, to avoid an error from `this.state.statsGraphsData` before the graph data loads.

app/assets/javascripts/components/user_profiles/user_uploads.jsx
```diff
@@ -0,0 +1,27 @@
...
+  },
+
+  render() {
+    let uploadList = (<UploadList uploads={this.props.uploads} view={GALLERY_VIEW} />);
+    if (this.props.uploads.length === 0) {
+      uploadList = (<span>{I18n.t('courses.user_uploads_none')}</span>);
+    }
+    return (
+      <div id="recent-uploads">
+        <h3>Recent Uploads</h3>
```
ragesoss: Needs i18n.

PR #1962, opened 2018-07-30
made datepicker onscreen by setting a max-width

RDenPen:


ragesoss:
@RDenPen could you post an illustration of this issue before and after?

Positioning these datepickers has been tricky, and there are still some layouts that it fails for, but I want to make sure we don't accidentally introduce new problems when we tweak the layout to solve existing problems.

Also, please remove the deletion of the two example config files from the commit. (It looks like you renamed those files rather than copying them when you set up your dev environment.)

ragesoss:
Looks like this has been abandoned.

PR #1969, opened 2018-08-17
[WIP] add UploadViewer component

ihsavru:
screen shot 2018-08-17 at 9 37 55 pm


ragesoss:
This is coming along nicely! You probably already have some of this in mind, but some UI bits I noticed:
* The extenal links in the Upload component are still present, so clicking one will both open up a new tab and open the UploadViewer in the current tab. Any click should only do one of these things. (Maybe `event.stopPropagation()` is needed?)
* The 'Original file' link goes to the thumbnail rather than the full-size original file. There's also no space between 'Original file' and the size annotation.
* There is no link to the main file page from UploadViewer. Clicking the image should open it, and an explicit button (like in ArticleViewer) would also be good.
* The Upload component should trigger the hand pointer, so that users know they can click anywhere on that component to see more detail.
* Unlike ArticleViewer, clicking outside of UploadViewer does not close it.

PR #1844, opened 2018-04-02
Add initial Dockerfile support

tuxology:
Fixes #1787

This sets ground for using a containerized image for build/test an deploy. Currently this is a single image containing the DB as well as the complete application, but there can be options to separate the logic for flexible deployments using `docker-composer`. Apart from Docker volumens could also be used eventually and if the maintainers wish, we could publish a prebuilt image to Docker Hub or Docker Store from the official WikiEduDashboard org! 

As a reference for now, a prebuilt working WikiEdu dashboard image is at my own repo : https://hub.docker.com/r/tuxology/wiki-edu-dashboard/

To test run the pre-built image, 
```
$ docker pull tuxology/wiki-edu-dashboard
$ docker run --net=host -it tuxology/wiki-edu-dashboard
```

To build and run your own image,

```
$ docker build -t wiki-edu-dashboard .
$ docker run --net=host -it wiki-edu-dashboard
```

PR #1970, opened 2018-08-19
Upgrade Ruby to 2.5.1

greysteil:
I'm pretty sure the only reason 2.5.0 is used is [this comment](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1844#issuecomment-378012713). Since Ruby 2.5.1 has now been out for some time, and since all Ruby requirements now specify 3 digits, I'm pretty sure upgrading to 2.5.1 will be fine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ env:
 # Run all tests in job 2
 - TEST_SUITE='~type:foo' AFTER_SUCCESS='' CC_TEST_REPORTER_ID=f0d5d763ddc6b3f980ba0fb0c38e7c27c0dffc4a5787cd7d5b8c2b0c3b2e27e2
 language: ruby
-rvm: 2.5.0
+rvm: 2.5.1
 sudo: required
 dist: trusty
 cache:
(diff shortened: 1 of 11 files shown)


ragesoss:
Thanks! The reason we're still on 2.5.0 is that deploying a ruby version upgrade to production usually requires some downtime. But this patch is great to have on hand with all tests passing. I'll keep this around until we're ready to do the upgrades on the production servers.

greysteil:
Makes sense!

Was also wondering if you'd be up for running Dependabot on this repo to keep the `Gemfile` / `Gemfile.lock` up-to-date. I ran it on my fork and got [these PRs](https://github.com/greysteil/WikiEduDashboard/pulls).

I built it and am keen to get as many open source repos using it as possible so I can use the CI data from each update run to detect and report any buggy library releases (pointing to the failing test suite as a reproduction case for the maintainer). Obviously lots of clever filtering to be done, but it's already got a few thousand users so should be possible.

Link if you want to give it a go is https://github.com/apps/dependabot. Free for open source forever, obviously.

PR #1981, opened 2018-09-09
Fixes #1978

### LINKED ISSUE #1978: Upgrade webpack to version 4

AmitJoki:
This is the initial migration to Webpack 4.

### DIFF
--- a/gulp/tasks/webpack.config.js
+++ b/gulp/tasks/webpack.config.js
@@ -1,0 +1,80 @@
+import path from 'path';
+import webpack from 'webpack';
+import ManifestPlugin from 'webpack-manifest-plugin';
+import config from '../config.js';
+
+const jsSource = `./${config.sourcePath}/${config.jsDirectory}`;
+const doHot = config.development && !config.watch_js;
+const appRoot = path.resolve('../../');
+const entries = {
+  main: [`${jsSource}/main.js`],
+  raven: [`${jsSource}/raven.js`],
+  styleguide: [`${jsSource}/styleguide/styleguide.jsx`],
+  survey: [`${jsSource}/surveys/survey.js`],
+  survey_admin: [`${jsSource}/surveys/survey-admin.js`],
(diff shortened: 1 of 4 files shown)


AmitJoki:
Some details of what went into this migration:

1. Instead of "loaders", we now have "rules" attribute.
2. Loading JSON is now taken care by default and hence, there's no need for a separate JSON rule now.
3. The other notable changes are upgrading packages webpack-dev-server, webpack itself and the additional package webpack-cli. Also webpack-manifest-plugin has to be upgraded to 2.0.0-rc.2
4. Another noticable change is that we don't need UglifyJS as it is inbuilt now.

This is still a work in progress but should not be long before it is production ready.

ragesoss:
Hmmm... the output looks like it's working fine for the production build, but the slow builds are definitely a problem for development.

In watch mode, it takes about 10 seconds to emit the changed files, and it was nearly instant with webpack 3.

It looks like `hot-dev` mode, which automatically reloads the page upon changes, doesn't work after the upgrade. I get an error with trying to load the `raven.js` file. (To test this, you need to set `hot_loading: 'true'` in application.yml and then do `gulp hot-dev`.)

ragesoss:
I think I found the problem. `config.development` is not coming through properly; it only gets evaluated at the beginning of the `gulp` process, before the `set-development` step changes it to `true`. So it always ends up as the default `false` value, resulting in a full rebuild on every change instead of an incremental rebuild.

I think this is because the webpack config gets imported to the webpack.js file, rather defined within the closure during the gulp "webpack" task.

One possible solution might be to define three separate, static webpack configs (for dev, hot-dev, and production), and then use the config inside the webpack gulp task to determine which one to use.

AmitJoki:
@ragesoss that was it! Javascript can come and bite you in unexpected ways! :smile: Anyway, I wanted to use CLI firstly and hence the separate config file. Now that I've removed it and put it inside the `webpack.js` that hoisting issue is resolved :)

PR #1992, opened 2018-09-17
Initial Babel Upgrade

### LINKED ISSUE #1991: Upgrade Babel to 7.0

AmitJoki:
Fixes #1991 

Running `npx babel-upgrade --write --install` took care of most things. `@babel/register` now has to be installed as a dev dependency.

Had to remove `babel-polyfill` as it no longer had any use.


app/assets/javascripts/components/common/multi_select_field.jsx
```diff
@@ -47,4 +47,4 @@ const MultiSelectField = createClass({
   }
 });
 
-module.exports = MultiSelectField;
+export default MultiSelectField;
```
AmitJoki: The previous line was causing a warning. Hence made this change to remove the warning and to keep things more in line with how it is elsewhere.

ragesoss:
@AmitJoki can you explain what's going on with the last commit? I don't know what the 'transformations' property in package.json does.

ragesoss:
Hmm... now feature specs are failing with variations on this javascript error:
`jQuery.Deferred exception: Can't find variable: Promise fetchDykArticles@http://127.0.0.1:38169/assets/javascripts//main.318a9f77ac802569ce9d.js:47:308673`

I guess this has something to do with removing the polyfill.

ragesoss:
Maybe something else needs to be done to enable the built-in polyfill capabilities of Babel?

ragesoss:
Yeah, looks like this will provide `Promise`: https://babeljs.io/docs/en/babel-polyfill

AmitJoki:
There's an error where sub-modules of Jest have Babel 6.26.3 as their dependency. The resolutions field is  yarn specific, and what it does is it resolves the dependency correctly to the one I've mentioned. So, they point to `7.0.0.bridge.0`.

I've tried putting the polyfill back. Though it was supposed to be left out as per the migration guide.

PR #1994, opened 2018-09-17
add deleteAllWeeks function as part of redux

### LINKED ISSUE #1988: finish converting CourseActions to Redux
We're nearing the end of converting to Redux from a Flux framework.

The last remaining function in CourseActions is `deleteAllWeeks`, which is called from the Timeline component. It isn't event handled by any Flux store, except in the case of an API failure. It can be converted to a Redux action in `course_actions_redux.js`. After that, the entire file can be renamed to just `course_actions.js`, since the old course_actions.js file will be gone.

Also, instead of doing a complete page reload (as currently happens after the action resolves), the `timeline.js` reducer should reset the timeline.

khyatisoneji:
Fix #1988


ragesoss:
@khyatisoneji the tests all passed on travis-ci, and on a quick skim, the implementation looks good to me. Please do go ahead with the rename.

To test this locally, you can:
* run `npm test` to run the javascript tests
* run `gulp build` to build the production version of the javascript
* create a course with a timeline locally, then on the timeline tab click `Delete Timeline` to test it out manually

It looks like we don't have a feature test for this. It would be great to add another test in `spec/features/timeline_editing_spec.rb` to test that the whole related feature is working properly, although that's not a requirement for this PR.

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -13,7 +13,7 @@ import CourseLink from '../common/course_link.jsx';
 import Affix from '../common/affix.jsx';
 import EditableRedux from '../high_order/editable_redux';
 
-import CourseActions from '../../actions/course_actions.js';
+import { deleteAllWeeks } from '../../actions/course_actions_redux.js';
```
ragesoss: Using `deleteAllWeeks` directly from an import does not work, as it's a redux "action creator" rather than an action. It needs to be connected to `dispatch` in order to work, so instead of importing it here, it should be imported in the connected parent component `TimelineOverview`, added to `mapDispatchToProps` and then passed to `Timeline` as a prop. See `deleteWeek` and other action creators in `TimelineHandler` for the pattern to follow.

I missed this the first time I skimmed the PR, and that lack of an integration test gave me false confidence in the passing tests. Sorry about that.

ragesoss: I just added a feature spec for this.

khyatisoneji: @ragesoss In all the other files where we have used course_actions_redux, we have directly used the function. For example: `import { fetchCourse } from './course_actions';`. Why in this case is it required to import the file and not the function?

ragesoss:
All the redux action creator's pass through `connect` before they are called. Have a look at `fetchCourse` in `course.jsx`, for example. The imported `fetchCourse` function is only referenced directly in `mapDispatchToProps`, which gets passed as the second argument to `connect`. The way `connect` works is that all the functions passed in mapDispatchToProps are availabe as props in the exported, connected component. So in `componentDidMount`, it gets called as `this.props.fetchCourse(courseID)`.

The reason for this is that a redux action creator is a function that returns another function, because it needs access to the redux store's `dispatch` function. So the action creator looks something like this:

```javascript
const doSomething = argument => dispatch => {
  somethingHappens(argument);
  dispatch({type: REDUX_ACTION_TYPE, data: argument });
}
```

`connect` takes that action creator, and basically injects dispatch into it, so the final function that is available in the component props is something like this:

```javascript
const connectedAction = argument => {
  doSomething(argument)(reduxStore.dispatch)
}
```

khyatisoneji:
@ragesoss I didn't understand why the test is failing. Can you please provide some information about this.

ragesoss:
@khyatisoneji it's failing because clicking the the 'Delete Timeline' button throws an error: `this.props.deleteAllWeek is not a function`. That's because in TimelineHandler, `deleteAllWeeks` is being imported from `timeline_actions`, but the function is currently in `course_actions`. (`timeline_actions` is probably the better place for it.)

You should try that feature out locally. (Let us know if you need any guidance on how to do that.)

PR #1998, opened 2018-09-18
add deleteAllWeeks function as part of redux and rename course_action…

### LINKED ISSUE #1988: finish converting CourseActions to Redux

khyatisoneji:
Fix #1988


khyatisoneji:
@ragesoss Wanted to start from scratch. Please ignore the previous PR and check this one. Thanks!

ragesoss:
great! looks good, works fine, tests are passing. Thanks!

PR #2001, opened 2018-09-19
[WIP] Migrate server_actions to use redux

### LINKED ISSUE #1995: convert many ServerActions functions to redux
As part of the migration from Flux to Redux, we need to rewrite all the remaining Flux actions as redux action creators. Several miscellaneous actions in ServerActions are not actually handled by Flux (except in the case of request failures) and can be easily converted to redux.

In particular, replace all the usages of:
* `ServerActions.deleteCourse`
* `ServerActions.updateSalesforceRecord`
* `ServerActions.notifyOverdue`
* `ServerActions.greetStudents`
* `ServerActions.remove(` (which can be replaced by a more specific redux action creator for the one place that uses it)

khyatisoneji:
Fix #1995


ragesoss:
I'm trying to do the final push in the redux migration, so I moved converted the remaining server actions except the one that is connected to ValidationStore.

PR #2002, opened 2018-09-19
fix #1995

### LINKED ISSUE #1995: convert many ServerActions functions to redux

khyatisoneji:
fix #1995


app/assets/javascripts/components/students/student_list.jsx
```diff
@@ -18,7 +18,7 @@ import { notifyOverdue } from '../../actions/server_actions_redux.js';
 import CourseUtils from '../../utils/course_utils.js';
 import { getFiltered } from '../../utils/model_utils.js';
 
-const StudentList = createReactClass({
+export const StudentList = createReactClass({
```
ragesoss: Since you're now exporting StudentList as a named export, I think there will be a linting error wherever the default export is being used, since it's probably doing `import StudentList`, using the same name as the named export. You can probably get around that by changing the import to `import ConnectedStudentList` or the like.

khyatisoneji: It is not showing linting error in my vscode.

khyatisoneji: You were right, I will make that change

app/assets/javascripts/components/students/student_list.jsx
```diff
@@ -18,7 +18,7 @@ import { notifyOverdue } from '../../actions/server_actions_redux.js';
 import CourseUtils from '../../utils/course_utils.js';
 import { getFiltered } from '../../utils/model_utils.js';
 
-export const StudentList = createReactClass({
+export const ConnectedStudentList = createReactClass({
```
ragesoss: The naming is backwars here. This is the unconnected component. The connected one is the default export at the end of the file.

If you leave this as StudentList, then you can do `import ConnectedStudentList` in StudentsHandler (and change the usage of it in line 60 to match).

khyatisoneji: Yeah, so sorry. I didn't realize this. Can you make this change as a maintainer?

ragesoss:
It looks like this was most of the way there, except for the 'leave course' action which looks from the broken spec like something went wrong there.

PR #2009, opened 2018-09-22
Include other training progress as well

### LINKED ISSUE #1530: Data on a user's training progress through all modules should be available
We currently only show such info for assigned modules on the Students tab of a course. For other use cases like the WMDE Thank You campaign (https://phabricator.wikimedia.org/T181305), it would be useful to provide data on any user about which modules they've started or completed.

There should probably be both a JSON API and a new component on user profile pages to display such data.

AmitJoki:
Fixes #1530


ragesoss:
Looks like this is passing the tests except for rubocop violations in some of the new code.

Per our chat discussion, the reducer needs to be either refactored to use the same data structure as it already has for the context of the StudentDrawer, or split out into a separate reducer for user profile training data.

AmitJoki:
@ragesoss I have made the requested changes. Please have a look at it. I figured separate reducers spell the intention much clearly.

config/locales/en.yml
```diff
@@ -972,6 +972,8 @@ en:
       other: "%{completed_count}/%{count} training modules completed"
     training_module_name: Training module
     training_module_status: Status
+    user_training_status: Training Status
+    user_no_training_status: You have not started any training modules yet.
```
ragesoss: "This user has not completed any training modules."

AmitJoki:
@ragesoss I've added another PR for that - https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2016 

Is this what you had in mind?

Issue #2004, opened 2018-09-20
'Credit' column data overflows in the Uploads table


ihsavru:
Under list-view of Uploads, sometimes the table overflows horizontally due to the Credit being too long for certain uploads in the list.
<img width="1680" alt="screen shot 2018-09-20 at 11 45 55 am" src="https://user-images.githubusercontent.com/22816171/45799384-f2ef8800-bcca-11e8-9982-0400fd7a6963.png">

Similar issue happens for Articles Table in campaigns, read more here: #1883

ShivangiKakkar:
@ihsavru Hey, May I work on this issue?

ihsavru:
@ShivangiKakkar Sure!

siwaniagrawal:
![articles](https://user-images.githubusercontent.com/35132729/45948671-910a8780-c016-11e8-8085-c476fa572eba.png)
Is this okay? or should I make some more changes?

siwaniagrawal:
Hey! @ihsavru  I was going through this issue, I inspected it in the browser and intend on adding required wraps. Can you please guide me through the process of adding necessary resources so, that I can access the respective page in my local.

ShivangiKakkar:
@siwaniagrawal I am working on it :)

siwaniagrawal:
oh okay :)

ragesoss:
It's not just the 'Credit' column that can cause overflow. Another situation is that before the thumbnail has been fetched, the filename is shown in the first column, and that may be longer than the image width. So the 'Image' column is extra wide until the file is loaded, and this may cause the table width to jump several times, as the longest remaining filename changes.

Long filenames with no dashes can also cause the `File Name` column to be too wide without wrapping.

Here's a good page to see all that behavior: https://outreachdashboard.wmflabs.org/courses/Uffizi/WDG_-_AF_2018_Florence/uploads

Especially, eg, pages 4 and 5.

Sanjana1308:
@ragesoss before:
![screenshot from 2018-09-25 23-41-03](https://user-images.githubusercontent.com/31213584/46034027-41fb4a00-c11d-11e8-913d-c4e8b30c2d9a.png)

Thank you for reviewing my pull request. As my prior observation, the upload column had either taken just long texts or links which were creating issues which I fixed. But , then I noticed the credit column  could take even images.
[...]

PR #2010, opened 2018-09-24
'Credit' column data overflows in the Uploads table #2004

Sanjana1308:
If provided with a link in 'Credit' column, it causes overflow. So I have reproduced this issue and fixed it by breaking this url(Table was already responsive) using css.

Signed-off-by: Sanjana suman <sumansanjna13@gmail.com>

## Before the Pull Request ##
* Test out your changes manually to confirm that it works and search for bugs.
* Run the specs for any Ruby files you changes, along with the feature specs that are related to your changes. Optionally, you can run the entire test suite locally (`rake spec`). For feature specs, you will need to use the production build of the assets by running `gulp build` first.

* If the PR is not complete but you want feedback on it, include `[WIP]` in the title to indicate work-in-progress. Edit the title when it's ready.

## After the PR ##
    * Fix any failures.

### DIFF
--- a/app/assets/stylesheets/modules/_uploads.styl
+++ b/app/assets/stylesheets/modules/_uploads.styl
@@ -144,3 +144,8 @@
 
     .upload-viewer-button
       margin 10px
+
+a.external.text {
+    word-break: break-word;
+    overflow-wrap: break-word;
+}


ragesoss:
Can you provide before/after screenshots? Testing this locally, it doesn't seem to help with the overflow for most uploads.

Adding wrap for `a.external` might make sense in general, since I'm pretty sure only HTML content pulled from a wiki is going to have the `external` class on a link, but I don't think that can be a complete fix for this table. If we do add CSS for `a.external`, that should either be specific to children of `.gallery` if it's meant to affect only the Uploads tab, or it should be in one of the general style modules if its meant to be global. Only CSS that is scoped specifically to uploads-related classes should be in `_uploads.styl`, to avoid unintended effects elsewhere.

PR #2016, opened 2018-09-26
Remove N+1 Queries

AmitJoki:
The User Training Status had N+1 Queries. This PR optimises the query.


ragesoss:
This doesn't work for me locally. The API endpoint returns an empty array.

PR #2017, opened 2018-09-26
Fixed overflow of article table #1883

### LINKED ISSUE #1833: Switch to headless Chrome as the driver for feature specs
Our current driver for browser tests is Poltergeist, which uses the now-unmaintained PhantomJS to run headless browser tests.

We can likely replace this with headless Chrome, which has been available for some time now. There will likely be some trial-and-error getting the configuration worked out properly; in preliminary exploration, I was able to get some tests to pass but ran into trouble related to URL blacklists.

### LINKED ISSUE #1883: Campaign > Articles table runs over container range
The article table under Campaign > Articles is out of the border on the live site, but looks normal on the local dev environment.
https://dashboard.wikiedu.org/campaigns/spring_2018/articles

siwaniagrawal:
Fixes #1883  
Fixed Program and Title column in the Article table.(https://dashboard.wikiedu.org/campaigns/spring_2018/articles)
Screenshot after the CSS changes

### DIFF
--- a/app/assets/stylesheets/modules/_tables.styl
+++ b/app/assets/stylesheets/modules/_tables.styl
@@ -57,6 +57,12 @@
       &:first-child td
         border-top 1px solid #ddd
 
+.course_title
+  word-break break-word
+  max-width 450px
+
+
+
   // Transparent header
   thead
     tr
--- a/app/assets/stylesheets/modules/_tables.styl
+++ b/app/assets/stylesheets/modules/_tables.styl
@@ -61,7 +61,9 @@
   word-break break-word
   max-width 450px
 
-
+.title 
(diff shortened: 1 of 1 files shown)


ragesoss:
Hmm... upon playing around with this, it seems like the article titles aren't much of a problem because they have spaces, and they break on words already. The problem is the program titles, which don't break on the underscores. At least for this particular table, I think the best solution might be to replace the underscores with spaces in the template: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/views/campaigns/_article_row.html.haml

ragesoss:
Please feel free to reopen if you want to continue working on this.

Issue #2026, opened 2018-10-01
At the end of a training module, the user should get a clear indicator of when the server has registered module completion


ragesoss:
Currently, the client sends an update to the server each time a new training slide is visited, and if it's the last slide then the server marks the module as completed.

It would be better if, on the last slide, there's an indicator that shows when the module has been registered as complete (and guide the user to remain on the page rather than navigate away, if it's not yet confirmed).

Issue #2025, opened 2018-10-01
When there are network errors during training modules, the UI should surface them


ragesoss:
Currently, if you lose network connectivity in the middle of a training module, you can continue to the end but the server will not be updated on your progress.

This is probably best handled after #1989

PR #2033, opened 2018-10-04
Migrate TrainingStore to  Redux

### LINKED ISSUE #1989: convert TrainingStore to redux
The TrainingStore is a relatively straightforward Flux store, which we need to convert into a Redux reducer.

The corresponding actions are spread between ServerActions and TrainingActions, and these should all be moved to a training_actions.js file and reimplemented in redux.

AmitJoki:
Fixes #1989, #2025, #2026.


ragesoss:
Looks like the quizzes are broken, eg, http://localhost:3000/training/students/wikipedia-essentials/five-pillars-quiz-1

ragesoss:
It's not working because the redux actions are mutating the state rather than returning a new object. If the same object gets returned (with mutations or not) then React won't re-render because it only does shallow comparison to determine if the props have changed.

The reducer is supposed be a pure function. Since this is a relatively complicated reducer, it might be worthwhile to write unit tests for it. The examples in the Redux tutorial videos are probably a good pattern to follow, where the tests use an immutability library to verify that the reducer isn't doing any mutation.

ragesoss:
Here's the example I'm thinking of: https://egghead.io/lessons/react-redux-avoiding-array-mutations-with-concat-slice-and-spread

app/assets/javascripts/reducers/training.js
```diff
@@ -0,0 +1,114 @@
...
+      break;
+    default:
+      newState = state;
+  }
+  return {
+    ...newState,
+    previousSlide: getPreviousSlide(newState),
+    nextSlide: getNextSlide(newState),
+    isFirstSlide: isFirstSlide(newState),
+    slides: newState.module.slides
```
ragesoss: This will return a new object on every redux action and call several functions, even for actions that are not handled by this reducer.

The default case for any reducer should always be `return state`.

AmitJoki: Makes sense. Will update it.

ragesoss: I'm still a little uncomfortable with this, as there's a lot of state getting updated repeatedly no matter what action it is.

In all the other reducers, each case explicitly returns the new state. I think that's a good pattern, because it lets us isolate the minimum changes necessary to process any given action. If something within an object changes, then we want to create a new object, but for anything that doesn't change, the ideal situation is to return the original rather than a new object. It's easier to understand that flow when each action is isolated.

app/assets/javascripts/reducers/training.js
```diff
@@ -0,0 +1,114 @@
...
+      newState = setCurrentSlide(state, data.slide);
+      break;
+    case RECEIVE_ALL_TRAINING_MODULES:
+      newState = { ...state, modules: data.training_modules };
+      break;
+    case SLIDE_COMPLETED:
+      newState = setEnabledSlides(state, data.slide);
+      break;
+    case MODULE_COMPLETED:
+      redirectTo(data);
```
ragesoss: This should not be handled in the reducer, as redirecting is a side-effect.

The redirection can happen in the action creator.

AmitJoki:
@ragesoss I've incorporated your requested changes and also fixed the related issues as well.

app/assets/javascripts/training/components/training_slide_handler.jsx
```diff
@@ -3,97 +3,94 @@ import PropTypes from 'prop-types';
...
       module_id: this.moduleId(),
       user_id: userId
     });
   },
 
-  storeDidChange() {
-    return this.setState(getState());
+  next() {
+    if (this.props.training.valid) {
+      this.setSlideCompleted(this.props.training.currentSlide.slug);
```
ragesoss: This will work like before if you do nextSlide here instead of currentSlide.

app/assets/javascripts/training/components/training_slide_handler.jsx
```diff
@@ -3,97 +3,94 @@ import PropTypes from 'prop-types';
...
+      this.props.setCurrentSlide(this.props.training.nextSlide.slug);
+    }
+  },
+
+  prev() {
+    if (this.props.training.valid) {
+      this.setSlideCompleted(this.props.training.currentSlide.slug);
```
ragesoss: No need to do this for prev, as it will never be the latest slide visited so it will be a noop on the server.

Issue #1286, opened 2017-05-22
Users should be able to cycle through diffs in the Activity view


ragesoss:
A 'Next' button and/or keyboard shortcut would make this list of diffs much more useful.

Kachulio1:
I would love to take this.

ragesoss:
For more context, here's a typical activity page for a course: https://dashboard.wikiedu.org/courses/University_of_Oklahoma/Cold_War_Science_(Spring_2018)/activity

Clicking the +/- button in the last column of one of the table rows will open up the DiffViewer component. To view another diff, you currently have to close the DiffViewer and then click a different row to reopen it with the new diff.

The idea here is to make to more efficient to go through each of the diffs in the Activity tab quickly.

ragesoss:
To test this locally, you need a course with some imported data. See https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/user_roles.md#populating-a-course-with-data

PR #2057, opened 2018-10-08
[WIP]Users should be able to cycle through diffs in the Activity view #1286

Kachulio1:
#### To manually  test this  : http://localhost:3000/courses/The_Lazy/Learn_How_To_Contribute_To_Any_Open_Source_Project/activity

screen shot 2018-10-10 at 22 24 31

[https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1286](#1286)

### DIFF
--- a/app/assets/javascripts/components/revisions/diff_viewer.jsx
+++ b/app/assets/javascripts/components/revisions/diff_viewer.jsx
@@ -29,7 +29,7 @@ const DiffViewer = createReactClass({
 
   getInitialState() {
     return {
-      showDiff: false,
+      showDiff: false
     };
   },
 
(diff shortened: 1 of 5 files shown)


Kachulio1:
@ragesoss The PR is ready for review 🤕

ragesoss:
@Kachulio1 cool! I just tested it out, and it looks like it's working smoothly.

I think the placement of the buttons needs to change. Currently, they'll float underneath the DiffViewer if the window is narrow enough.

occluded

In general, the whole app ought to show up in the center 1100 px `.container` div. One approach in this case could be to add the buttons on the edges of the DiffViewer, within the same `.diff-viewer` div.

I'm also having trouble getting a clear picture of which changes are functional and which are just reformatting the existing code. If you can isolate the functional changes (and optionally do a separate PR for strictly formatting changes) that would be helpful.

ragesoss:
Feel free to reopen if you want to take this up again.

PR #2070, opened 2018-10-09
[WIP] Extract Users::Enrollment_Controller

### LINKED ISSUE #1949: Refactor UsersController into multiple controllers
`UsersController` is somewhat too large and does too many things. In particular, `enroll` action and all the private methods that support it ought to be moved to a new controller, such as `Users::EnrollmentController`.

Relatedly, MassEnrollmentController and SelfEnrollmentController could be moved under the `Users::` namespace.

See also #1417 for a similar issue, and #1932 for an example of this type of refactoring.

LuanvP:
Extract EnrollmentController from UsersController
`Fixes #1949`


ragesoss:
This was going in the right direction I think, but still had some remaining spec failures.

PR #2108, opened 2018-10-16
Filter articles by new, existing or both

### LINKED ISSUE #97: Articles view for a course should be filterable by 'new'
The Articles view for a course should be able to be filtered to only new articles.

AmitJoki:
Fixes #97


AmitJoki:
The Travis CI failure is unrelated to this change.

ragesoss:
@AmitJoki could you rebase this so we can see if the tests are green?

AmitJoki:
@ragesoss I've rebased it.

app/assets/stylesheets/modules/_articles.styl
```diff
@@ -0,0 +1,20 @@
...
+    font-weight: 600;
+
+.article-sort
+  width: 60px
+  .sorts
+    background: url(../images/sort.svg) no-repeat 48% #fff;
+    -webkit-appearance: none;
+    text-indent: -200px;
+
+.tooltip-trigger
```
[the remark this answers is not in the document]
AmitJoki: What do you want it to be? I think it is better to be throughout because the tooltips don't need to be that wide. If you just want it to be for articles tab, that's fine too. Your call.

AmitJoki:
Have made the CSS tweaks. It should be fine now. Do let me know what you think of it after you've checked it out.

ragesoss:
I'm trying this out, and there's some buggy behavior when switching among filter states. In particular, the 'New and Existing' option doesn't work as expected.

I have a course that has both new and existing articles from English Wikipedia, plus some existing but not new ones from Wikidata.

If I change to either New or Existing, after that, changing back to 'New and Existing' does not have an effect.

Guardfile
```diff
@@ -34,7 +34,7 @@ end
 #  * zeus: 'zeus rspec' (requires the server to be started separately)
 #  * 'just' rspec: 'rspec'
 
-guard :rspec, cmd: 'bundle exec rspec' do
+guard :rspec, cmd: 'echo "done"' do
```
ragesoss: This looks like it was included by accident.

app/assets/javascripts/reducers/articles.js
```diff
@@ -61,6 +67,15 @@ export default function articles(state = initialState, action) {
       }
       return { ...state, wikiFilter: action.wiki };
     }
+    case SET_NEWNESS_FILTER: {
+      switch (action.newness) {
+        case 'new':
+        case 'existing':
+          return { ...state, newnessFilter: action.newness };
+        default:
+          return state;
```
ragesoss: Here's the bug. Changing from one of the other filters to 'both' falls through to the default case and does not change state.

I suggest matching the implementation for SET_PROJECT_FILTER:

```javascript
    case SET_NEWNESS_FILTER: {
      if (action.newness === 'both') {
        return { ...state, newnessFilter: null };
      }
      return { ...state, newnessFilter: action.newness };
    }
```

In this case, because of how you implemented the selector, it was also work to just do it unconditionally:

```javascript
    case SET_NEWNESS_FILTER: {
      return { ...state, newnessFilter: action.newness };
    }
```

That could even be extended to the `SET_PROJECT_FILTER` case, if we tweak the selector implementation to use a case statement with the default fallback like in new selector.

ragesoss:
I made the last changes, and will merge once the build passes.

app/assets/javascripts/reducers/articles.js
AmitJoki: Yes, I've edited it so that we could use a simple reducer.

ragesoss:
I merged it already.

PR #2127, opened 2018-10-19
Add absolute sorting for character added sorting

### LINKED ISSUE #101: Sorting Activity by chars added/removed is not ordered intuitively
If I sort by characters added on the Activity tab of a course, it will start from the highest negative numbers, then go down to -1, then start from teh highest positive numbers and go down to zero.

I think that the two most useful sorting patterns would be:
- Sorting by magnitude (highest to lowest on first click)
- Sorting in number order (high to zero to negative)

Either would be preferable to the current sort order. Ideally, that sorter would cycle from magnitude to number order to reverse magnitude to reverse number order.

AmitJoki:
Fixes #101

### DIFF
--- a/app/assets/javascripts/utils/model_utils.js
+++ b/app/assets/javascripts/utils/model_utils.js
@@ -19,7 +19,7 @@ export const getFiltered = (models, options) => {
 // already sorted by, and the default sorting direction.
 // If you sort a second time by the same key, then it will reverse the sorting.
 // Otherwise, it will just sort by that key.
-export const sortByKey = (models, sortKey, previousKey = null, desc = false) => {
+export const sortByKey = (models, sortKey, previousKey = null, desc = false, absolute = false) => {
   const sameKey = sortKey === previousKey;
   let newKey;
   if (sameKey) {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/utils/model_utils.js
```diff
@@ -40,7 +40,14 @@ export const sortByKey = (models, sortKey, previousKey = null, desc = false) =>
 
   const reverse = !sameKey !== !desc; // sameKey OR desc is truthy, but not both
   let newModels;
-  if (reverse) {
+  if (absolute) {
+    const sorted = _.sortBy(models, [o => Math.abs(o[sortKey])]);
+    if (reverse) {
```
ragesoss: This logic could probably be refactored to handled reverse separately from the rest. Like, instead of `if/else if/else`, you first do the basic sorting with whatever options you need, and then you reverse the results at the end if that option is enabled.

AmitJoki: There's a problem with that. We can't simply reverse the sorted because the `sort` function is specifically written to take care of `null` values to end at last instead of beginning when descending order is to be obtained.

PR #2138, opened 2018-10-24
Corrected Text-Alignment for pageviews & revScore

### LINKED ISSUE #2064: In article finder, the numbers in the last two columns should be right-aligned
For the pageviews, in particular, that will make it easier to compare numbers of different magnitudes.

Currently, they're left-aligned:

alignment

Jigar3:
Fixes #2064 

This is how the alignment looks now.
actual

### DIFF
--- a/app/assets/stylesheets/modules/_article_finder.styl
+++ b/app/assets/stylesheets/modules/_article_finder.styl
@@ -11,6 +11,13 @@
 .fetch-more
   height 3rem
 
+.pageviews 
+  text-align justify
+  direction rtl
+
+.revScore 
+  text-align center
+
 .search-bar
   display flex
   flex-direction row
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/modules/_article_finder.styl
```diff
@@ -11,6 +11,14 @@
 .fetch-more
   height 3rem
 
+.pageviews 
+  text-align justify
+  direction rtl
+  padding-right: 3rem !important
```
ragesoss: Can you find a way to do this without `!important`? (Also, the colon is not necessary here.)

ragesoss:
Thanks! If you can rebase this, or just squash the relevant changes to a single commit, or write a fresh commit on a clean branch, this is ready to go.

I think you could achieve the extra spacing you had before you if add more specificity to the selector, for example, by putting a CSS id in the top div of ArticleFinder and making all the classes in the article_finder.styl file specific to `#article_finder`. But it's not a big deal either way.

Jigar3:
@ragesoss Please review

PR #2140, opened 2018-10-25
Enabled and fixed 3 esLint rules

takidelfin:
#1408

Enabled and fixed violations for rules:
- `arrow-parens`
- `import/newline-after-import`
- `react/jsx-closing-tag-location`

Made for GCI task

### DIFF
--- a/test/components/settings/settings_handler.spec.jsx
+++ b/test/components/settings/settings_handler.spec.jsx
@@ -23,9 +23,11 @@ describe('SettingsHandler', () => {
       },
       notifications: [],
     });
-    const wrapper = shallow(<Provider store={store} >
-      <SettingsHandler />
-    </Provider>);
+    const wrapper = shallow(
+      <Provider store={store}>
+        <SettingsHandler />
+      </Provider>
+    );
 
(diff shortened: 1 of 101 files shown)


test/components/settings/settings_handler.spec.jsx
```diff
@@ -23,9 +23,7 @@ describe('SettingsHandler', () => {
       },
       notifications: [],
     });
-    const wrapper = shallow(<Provider store={store} >
-      <SettingsHandler />
-    </Provider>);
+    const wrapper = shallow(<Provider store={store}><SettingsHandler /></Provider>);
```
psinghal20: Can we add an exception here? In my opinion, it was much more readable earlier.

takidelfin:
```jsx
    const wrapper = shallow(
      <Provider store={store}>
        <SettingsHandler />
      </Provider>
    );
```
?

takidelfin: I will rebase that commit, so the git log will look clean :smile:

Issue #1709, opened 2018-02-15
Improve Setup Script


psinghal20:
For those who faced issues with using the setup scripts, Please comment below with the errors you faced and the logs for your installation. You can find the logs in setup directory.

vzhz:
I run macOS High Sierra and when I ran `python setup.py` from the root directory, I was referred to the manual setup process. 
I am happy to attempt this setup again as a test as work on this continues.
Here is my complete log:
`Admins-MacBook-Air:WikiEduDashboard Admin$ python setup.py
WARNING! This is a work in progress script. It has been tested to work for debian-based systems.   There are individual operating system dependent scripts being called from this one.   You can find them in setup directory. The scripts for windows and OSX are still not complete.   You can use them as a starting point or as a reference. If you run into any errors while running this script,   please comment with your issue on https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1709   Please upload your logs for installation with your issue reports.   The logs can be found in the setup directory. If you can help improve this script,   We would love your contributions.
Please install ruby-2.5.0 before running this script.
Sorry! Your operating is not supported by this script. Please refer https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/ docs/setup.md for manual setup instructions.`

ragesoss:
The message for when ruby 2.5 is not available should just say to install it, without insisting on a specific method. It can link to https://rvm.io/ as one option for doing so, but without any more specific instructions than that.

The exact instructions in the current script didn't work for me on a Mac, but other instructions from the RVM website did.

jeffblack360:
The `setup/deb-setup.sh` complains if it doesn't find ruby 2.5.0 installed. I have ruby 2.5.1: should the script at merely check for a match on 2.5 vs. 2.5.0?  

I changed the script to complete setup.

arcayn:
Hi, just leaving my notes on the setup files here. It's a ~1000 word markdown file with notes on both ubuntu and windows installations, and which documents the changes in the PR I made.
[_installing_wikiedudashboard_.txt](https://github.com/WikiEducationFoundation/WikiEduDashboard/files/2517049/_installing_wikiedudashboard_.txt)

hardy1334:
Hey,I'm setting up https://github.com/WikiEducationFoundation/WikiEduDashboard
repo on my system, its front end is working but its not fetching anything from database
[...]

PR #2147, opened 2018-10-25
Improves win-setup.bat to work smoothly #1709

arcayn:
This fixes several issues with the win-setup.bat file (issue #1709)

1) Adds a fix to ruby devkit not installing from choco - still unclear as to why this occurs, but trying the fix as discussed [here](https://github.com/chocolatey/choco/issues/846) seemed to work, so the version is explicitly stated
2) Added the path to yarn directories to PATH in two places to make sure the installer can use yarn properly as it previously would crash as it could not find yarn
3) Created a user "wiki" for the database as otherwise migration would not work (this copied from the debian setup file)
4) Switched all instances of `rake` and `yarn` to `call rake` and `call yarn` to stop the batch process hanging
5) Added a message to let the user know the batch has in fact completed!

These fixes are all down to my own personal experience of a broken batch file whilst installing. On a fresh install, the new batch works with no tweaking to the script!

(This is part of a Google Code-In task)


arcayn:
@psinghal20 no worries! Echoes added, will look into that issue

psinghal20:
Looks good to me! @ragesoss you can go ahead and merge this. We can raise a separate issue to bring this one closer to the Linux scripts.

PR #2148, opened 2018-10-26
First pass at positioning the banner correctly

### LINKED ISSUE #927: notification banners get covered in mobile and narrow desktop layouts
When the nav cannot fit all the links and the get help button:
notif hidden 2

When it switches to mobile layout:
notif hidden

AmitJoki:
Fixes #927

I think this should be able to do it, but I don't have lot many devices to test on and it requires a fair trail and error method of finding the right margin-value.

### DIFF
--- a/app/assets/stylesheets/modules/_modules.styl
+++ b/app/assets/stylesheets/modules/_modules.styl
@@ -96,3 +96,6 @@
   .course-alerts
     height 0px
     margin-top -30px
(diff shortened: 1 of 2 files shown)


AmitJoki:
Before:

before

After:

after

ragesoss:
This is a clear improvement. Mobile layout still isn't 100% compatible with the notifications and banners, though:

better

PR #2149, opened 2018-10-26
Special users endpoint

AmitJoki:
I've made the requested changes, except removing the placeholder connected component which I am guessing will be needed for creating further UI actions.

Apart from that, this is an initial pass at the requested changes on this PR. What actions should be there in addition to what's present?


ragesoss:
Looks like it will need a bit of work to rebase it, and there are also JS linting errors.

ragesoss:
Testing this out, and in addition to the failing tests in settings_controller_spec, it looks like it also breaks the current functionality of the Settings page. For example, I tried to add a new admin, and on this branch I get  an error saying the user isn't found, while it works properly on master. (To test this, your User record need to be a super admin, `permissions: 3`.)

PR #2156, opened 2018-10-26
Fixed a typo

Bjornskjald:
Basically, fixed "zajdź" to "znajdź" (find in Polish)


ragesoss:
@Bjornskjald thanks! The normal place to make translation updates is translatewiki.net: https://translatewiki.net/wiki/Translating:Wiki_Ed_Dashboard .

I'll merge this now although I'm not 100% sure it won't get overwritten later from translatewiki.

Issue #2128, opened 2018-10-22
Bad interwiki prefixes added while editing dashboard related pages by Dashboard


urbanecm:
Hi, 

I was informed Dashboard is using bad interwiki prefixes (always directing to English language project regardless on project language. 

This is example: https://cs.wikipedia.org/w/index.php?title=Wikipedie:N%C3%A1st%C4%9Bnka/Wikimedia_%C4%8CR/Editaton_v_%C4%8Cesk%C3%BDch_Bud%C4%9Bjovic%C3%ADch_(podzim_2018)&diff=16547465&oldid=16546868 . 

Fix: https://cs.wikipedia.org/w/index.php?title=Wikipedie:N%C3%A1st%C4%9Bnka/Wikimedia_%C4%8CR/Editaton_v_%C4%8Cesk%C3%BDch_Bud%C4%9Bjovic%C3%ADch_(podzim_2018)&diff=16547473&oldid=16547465 (I know fixing things after Dashboard is useless, this is just for demonstration). 

In short, instead of `[[:language:family:page]]` Dashboard should use `[[:family:language:page]]`. This is fully qualified link and will work in all WMF projects. 

Thank you for fixing this, 

Martin Urbanec
cs.wikipedia admin

JAnDud:
and family should be shortened to standard interwiki prefix:
wikisource:->s:
wiktionary:->wikt:
wikiquote:->q:
etc.

urbanecm:
That's another possible solution, but wikisource: and s: is the same (on WMF wikis), and as nobody should edit dashboard-maintained page manually and insert the prefix manually, it can do no harm I think.

ragesoss:
Just deployed a fix, and it appears to be working: https://cs.wikipedia.org/w/index.php?title=Wikipedie%3AN%C3%A1st%C4%9Bnka%2FWikimedia_%C4%8CR%2FEditaton_v_%C4%8Cesk%C3%BDch_Bud%C4%9Bjovic%C3%ADch_%28podzim_2018%29&type=revision&diff=16569665&oldid=16547476

PR #2157, opened 2018-10-27
Fixed interwiki prefixes, added shorthands and tests

takidelfin:
#2128

Fixed interwiki prefixes by reordering them (as stated in issue), added shorthands from [en.wikipedia.org/wiki/Help:Interwiki_linking](https://en.wikipedia.org/wiki/Help:Interwiki_linking#Project_titles_and_shortcuts). I also wrote tests for `lib/wikitext.rb`

GCI Task

### DIFF
--- a/lib/wikitext.rb
+++ b/lib/wikitext.rb
@@ -65,11 +65,51 @@ class Wikitext
     language = assignment.wiki.language
 
     # For other wikis a language prefix is required, except for wikidata where the language is nil
-    language_prefix = language ? ":#{language}:" : ''
+    language_prefix = language ? ":#{language}" : ''
     # If the project is different, a project prefix is also necessary.
-    project_prefix = project == home_wiki.project ? '' : "#{project}:"
+    project_prefix = project == home_wiki.project ? '' : ":#{project}"
 
-    language_prefix + project_prefix + title
+    # Interwiki shorthands, source: https://en.wikipedia.org/wiki/Help:Interwiki_linking#Project_titles_and_shortcuts
+
+    case project_prefix
+    when ':wikipedia'
+      project_prefix = ':w'
+    when ':wiktionary'
+      project_prefix = ':wikt'
+    when ':wikinews'
+      project_prefix = ':n'
+    when ':wikibooks'
+      project_prefix = ':b'
+    when ':wikiquote'
+      project_prefix = ':q'
+    when ':wikisource'
+      project_prefix = ':s'
+    when ':wikiversity'
+      project_prefix = ':v'
+    when ':wikivoyage'
+      project_prefix = ':voy'
+      # # Disabled due to Invalid Project:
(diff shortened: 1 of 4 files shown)


spec/lib/wikitext_spec.rb
```diff
@@ -84,28 +84,73 @@
...
-      output = subject.assignments_to_wikilinks(assignments, en_wiki)
-      expect(output).to include('[[Selfie]], ')
-      expect(output).to include('[[:Category:Photography]]')
-      expect(output).to include('[[Bishnu Priya]]')
-      expect(output).to include('[[:es:Blanca de Beaulieu]]')
-      expect(output).to include('[[:es:wiktionary:agrazarías]]')
-      expect(output).to include('[[wikidata:Q60]]')
+      VCR.use_cassette('multiwiki_assignment') do
+        output = subject.assignments_to_wikilinks(assignments, en_wiki)
+        puts output
```
psinghal20: I don't think we should log any output while running ruby specs.

takidelfin:
@ragesoss can you review my pull request?

lib/wikitext.rb
```diff
@@ -64,12 +64,28 @@ def self.format_assignment_title(assignment, home_wiki)
     project = assignment.wiki.project
     language = assignment.wiki.language
 
+    # Interwiki shorthands, source: https://en.wikipedia.org/wiki/Help:Interwiki_linking#Project_titles_and_shortcuts
+    shorthands = {
```
ragesoss: Instead of defining this in the method, this hash should be a constant.

spec/lib/wikitext_spec.rb
```diff
@@ -84,15 +84,46 @@
...
+    let(:pl_wikiquote) { create(:wiki, language: 'pl', project: 'wikiquote') }
+    let(:es_wikisource) { create(:wiki, language: 'es', project: 'wikisource') }
     let(:wikidata) { create(:wiki, project: 'wikidata') }
+    let(:de_wikiversity) { create(:wiki, language: 'de', project: 'wikiversity') }
+    let(:en_wikivoyage) { create(:wiki, language: 'en', project: 'wikivoyage') }
+    # Disabled due to Invalid Project:
+    # (`raise InvalidWikiError unless PROJECTS.include?(project)`) in app/models/wiki.rb
+    # let(:metawikipedia) { create(:wiki, language: 'en', project: 'metawikipedia') }
+    # let(:meta) { create(:wiki, language: 'en', project: 'meta') }
+    # let(:commons) { create(:wiki, language: 'en', project: 'commons') }
```
ragesoss: Commons is a valid project, it's just done as `language: 'commons', project: 'wikimedia'`

takidelfin: what do to with `Commons => c` shorthand then? should I do something like ` shorthands[language_prefix] || language_prefix`?

Issue #2015, opened 2018-09-25
Replace `$.ajax` requests with the Fetch API


ragesoss:
We currently use jQuery through much of the frontend for web requests of all sorts. To lay the ground work for getting rid of jQuery as a dependency, we can start by switching to use Fetch for all the requests.

I suggest starting by replacing just one core request with a Fetch implementation (such `fetchCourse` from `course_actions.js`) to get it working smoothly, and then proceed with converting the many other usages of jQuery `.ajax`.

For each request that you change, be sure to test it manually to verify that it's working the same way before and after. One important difference with `fetch` is that authentication credentials will not be included unless they are [explicitly enabled](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/fd9264617323aba70e6fc6de23eeb89eac5b8838); without the credentials, a request will be treated like it comes from a logged out user, which may introduce subtle (or not-subtle) bugs.

Nshul:
Can i work on this??

Jigar3:
Has PR #2029 solved this issue or some work still remaining?
@ragesoss

ragesoss:
This is close to done. There are just a few uses of `ajax` left in the codebase.

Aminehassou:
How do I test the AJAX calls in the `Survey.js` and`SurveyAdmin.js` files? I'm accessing the surveys page locally through `http://localhost:3000/surveys` but I can't seem to find where this AJAX call is used:

```
  submitQuestionGroup(index) {
    if (this.submitted.indexOf(index) !== -1) { return; }
    this.submitted.push(index);
    const $form = $(`form[data-question-group='${index}']`);
    const url = $form.attr('action');
    const method = $form.attr('method');
    const _context = this;
    $form.on('submit', function (e) {
      e.preventDefault();
      const data = _context.processQuestionGroupData($(this).serializeArray());
      return $.ajax({
        url,
        method,
        data: JSON.stringify(data),
        dataType: 'json',
        contentType: 'application/json',
        // success(d) { return console.log('success', d); },
        // error(er) { return console.log('error', er); }
      });
    });
    $form.submit();
  }
```

ragesoss:
I believe this function is called when a user completes the last question of a survey. You'll need to set up a survey and then go through it in order to test it. (Check out this script for a quick way to set up some surveys in your dev environment: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5269 )

PR #2158, opened 2018-10-27
Replaced `$.ajax` with fetch API

Jigar3:
Issue #2015 
@ragesoss  Please review

### DIFF
--- a/app/assets/javascripts/utils/api.js
+++ b/app/assets/javascripts/utils/api.js
@@ -89,200 +89,226 @@ const API = {
   },
 
   createCustomFeedback(assignmentId, text, userId) {
-    return new Promise((res, rej) =>
-      $.ajax({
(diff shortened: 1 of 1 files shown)


ragesoss:
Looks like several feature specs are failing with this change. You can run them with a visible browser if you first remove 'headless' from the Capybara config in `spec/rails_helper.rb`.

Jigar3:
> You can run them with a visible browser if you first remove 'headless' from the Capybara config in spec/rails_helper.rb

How exactly can I see those changes ?

ragesoss:
You can run the specs in a single file with `rspec spec/features/some_name_spec.rb`. If the headless option is not there in rails_helper.rb, then running a feature spec will launch Chrome and you can watch the test as it runs.

PR #2159, opened 2018-10-27
Enabled and fixed 4 eslint rules (Issue #1408)

arcayn:
Enalbled and fixed 4 eslint rules (as requested [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1408)):

1. `comma-status`
2. `space-unary-ops`
3. `no-restricted-syntax`
4. I also enabled `no-else-return` but with `allowElseIf` - this is the default, but has been overriden somewhere outside of WikiEduDashbooard's .eslintrc, and allowing else if blocks does make the code more readable than multiple disjointed if statements

This is for a GCI task

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -65,7 +65,6 @@
     "prefer-destructuring": 0,
     "react/prop-types": 0,
     "semi-style": 0,
-    "operator-linebreak": 0,
     "react/jsx-tag-spacing": 0,
     "implicit-arrow-linebreak": 0,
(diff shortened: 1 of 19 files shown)


psinghal20:
The changes look good to me. @ragesoss you can go ahead and merge this.

arcayn:
Enabled an extra rule whilst I wait for the GCI task to be accepted - `operator-linebreak`

PR #2172, opened 2018-10-31
[WP] Test that all actions don't modify passed state for course reducer

ledestin:
I expect `Proxy` might be better extracted somewhere.

### DIFF
--- a/test/reducers/course.spec.js
+++ b/test/reducers/course.spec.js
@@ -1,12 +1,31 @@
 import deepFreeze from 'deep-freeze';
 import course from '../../app/assets/javascripts/reducers/course';
-import { UPDATE_COURSE } from '../../app/assets/javascripts/constants';
+import * as ACTION_TYPES from '../../app/assets/javascripts/constants/course';
 import '../testHelper';
 
 describe('course reducer', () => {
+  it("doesn't modify passed state", () => {
+    const initialState = {};
+    deepFreeze(initialState);
+
+    Object.values(ACTION_TYPES).forEach((actionType) => {
+      const action = { type: actionType };
+      const actionBlackHole = new Proxy(action, {
+        get: function (target, key) {
+          if (key === 'type') {
+            return target.type;
+          }
+
+          return actionBlackHole;
+        }
+      });
+      course(initialState, actionBlackHole);
+    });
+  });
+
   it('updates an attribute via UPDATE_COURSE', () => {
     const initialState = { title: 'old title', term: 'old term' };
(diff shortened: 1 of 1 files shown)


ragesoss:
Can you explain what's going on here, both in terms of the strategy and the implementation?

ragesoss:
Thanks! It seems like a useful approach, especially if you're trying to be extremely thorough with reducer testing. My current overall goal with the javascript test suite is to improve the understandability, as with the JS tests in particular, the biggest frustration is when they break without a clear picture of why (when the intended functionality is not broken), and they aren't clear enough to easily fix or refactor to be less brittle.

I don't think we have that problem with reducer tests as this point, as reducers are one of the areas where test setup can be really simple and match the mental model for writing the code in the first place. It seems like this black hole pattern might break that, especially for when a reducer gets extended and some actions could throw errors if the action doesn't have the expected shape beyond the type. I'm also having trouble imagining what errors it might catch that would otherwise sneak through. I'd rather get in the habit of explicitly testing all the action types handled by a given reducer, and trust the simple and always-the-same case-statement pattern of the reducers to ensure that any other action will just return the state unchanged.

ledestin:
> I don't think we have that problem with reducer tests as this point, as reducers are one of the areas where test setup can be really simple and match the mental model for writing the code in the first place. It seems like this black hole pattern might break that, especially for when a reducer gets extended and some actions could throw errors if the action doesn't have the expected shape beyond the type.

When a reducer gets extended, there's a high chance you'd need to update the passed action, because a new property was introduced. Without the update, the test will break. That would make the tests more brittle than the black hole approach, though easier to understand. Magic vs Java explicitness is what it is, in my opinion.

> I'm also having trouble imagining what errors it might catch that would otherwise sneak through. 

Oh, that was just designed to catch the passed state modifications. Writing down all action types seems like a big effort, and will break on changes like property renames or moving things around in the action object.

> I'd rather get in the habit of explicitly testing all the action types handled by a given reducer, and trust the simple and always-the-same case-statement pattern of the reducers to ensure that any other action will just return the state unchanged.

There could be a hybrid approach, where you use black hole in many of the action tests, just so that you don't have to prepare the action that will satisfy the given action type.

Overall, it looks like a lot of work writing down action parameter for each of the action types, and I wonder whether there's a better way, be it what I suggested or something else.

ledestin:
I had a sleep on it, and if you're unwilling to accept it as a check that state isn't mutated by the reducer, there's no point in using it otherwise. Specific action tests wouldn't need black hole, they'd check that new state is valid, and that requires a valid input action.

PR #2173, opened 2018-10-31
[WIP] Enable timezone-related rubocop rules

kaine119:
1. move FormatStringToken rule to permanent exception; not applicable if only string interpolation (`#{foo}`) is used
2. enabled Rails/TimeZone and Rails/Date

There's one Rubocop violation in `app/models/campaign.rb`; the `validate_date_attribute` function there used `Time.parse`, which violates Rails/Timezone. Changing it to `Time.zone.parse()` conforms to Rubocop, but [broke the spec](https://travis-ci.org/kaine119/WikiEduDashboard/jobs/448672011#L5332) for date validation. I left it there as `Time.parse()` for now, but I'm not sure how to fix the Rubocop violation.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -77,6 +77,8 @@ Style/RedundantReturn:
   Enabled: false # Sometimes a redundant return enhances clarity.
 Style/SymbolArray:
   Enabled: false # Rare syntax that is potentially confusing to newcomers.
+Style/FormatStringToken:
+  Enabled: false # Not very helpful, string interpolation used instead of Kernel::Format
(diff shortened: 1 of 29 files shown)


ragesoss:
As a start, I would suggest just adding inline comments to disable the rule for that one place. Since these rules are related, there may be some interaction between them, so to be confident in these changes, I'd like to see builds for each of the rules independently. Could you open a new PR for each rule separately? That might help us pin down exactly how to work with the Campaign one that breaks the spec.

kaine119:
Alright; closing this PR first then.

Issue #2150, opened 2018-10-26
Provide a clearer notice when a program has no campaign


Samwalton9:
A user had an issue (https://meta.wikimedia.org/wiki/Talk:Programs_%26_Events_Dashboard#Password_not_accessing_dashboard) with a program not having a campaign set. This meant the program couldn't be joined by other users.

When this happens the message users get is, I believe, "This course has not yet been approved for enrollment." It seems like we could display a more informative notice here, specifying the problem (no campaign) and noting more explicitly that users won't be able to join until it's set.

ragesoss:
This is specifically for the case of the P&E Dashboard, rather than the Wiki Education Dashboard (which already has a different, detailed notification about the next steps for an unlisted course).

This is currently handled in the "Admin / Instructor notifications" of `course.jsx`. Tackling this issue would be a good opportunity to extract the logic for all the course-alerts to a separate component (and to further extract this particular set of messages related to the 'life cycle' of getting a new course ready for use into its own component).

Samwalton9:
Done in https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2174.

PR #2174, opened 2018-11-01
'No campaign' warning for 'Course' view

takidelfin:
#2150 

Moved some messages to i18n (to make them translatable), added `no-campaign` alert and moved alerts to new `CourseAlert` component.

A new message was written by @ragesoss

GCI Task

screenshot_20181101_122459


ragesoss:
This looks like it's working as intended! The 'warning' class should not be used; it's basically the same kind of alert that most of the others are: not an error, just informational.

The extraction of the individual alert component is good. As a next step, I think the entire logic for which notification alerts to display should be extracted. That will make it a lot easier to understand both the Course component itself, and the component responsible for all the notifications when you want to add a new one (like this new one).

ragesoss:
This message isn't being rendered:
create timeline message

ragesoss:
I think in `CourseAlert`, the message should be passed as a prop, with explicit I18n done outside of the component. So the interface would look like this:

```jsx
<CourseAlert
  key="submit"
  message={I18n.t('courses.submitted_note')}
/>
```

The way you implemented it works, but it's a little harder to understand the correct usage, because of its reliance on `this.props.children`.

ragesoss:
(I haven't checked the rest of the messages, but if they have similar issues, those will need to be fixed as well.)

takidelfin:
Oh, sorry! initially I wanted pass message using prop, but I thought that
`this.props.children` will look cleaner.

takidelfin:
@ragesoss Could you please review my pull request?

PR #2178, opened 2018-11-01
[WIP] Permanently exempt rubocop rule: FormatStringToken

kaine119:
Moves the `Style/FormatStringToken` in rubocop.yml to permanent exemptions; it's not applicable if only string interpolation (#{foo}) is used.

(from #2173)

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -77,8 +77,6 @@ Style/RedundantReturn:
   Enabled: false # Sometimes a redundant return enhances clarity.
 Style/SymbolArray:
   Enabled: false # Rare syntax that is potentially confusing to newcomers.
-Style/FormatStringToken:
-  Enabled: false # Not very helpful, string interpolation used instead of Kernel::Format
 Naming/VariableNumber:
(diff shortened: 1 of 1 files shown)


kaine119:
The same spec failed in [the first test](https://travis-ci.org/WikiEducationFoundation/WikiEduDashboard/jobs/449328001#L4440) but passed in [the second test](https://travis-ci.org/WikiEducationFoundation/WikiEduDashboard/jobs/449328002#L4885), is that normal?

ragesoss:
Yeah, that happens every once in a while with that spec in particular. If the second, more complete test suite passes, then generally everything is fine. I restarted the first one, which will probably pass.

ragesoss:
Are there even any violations of this rule? If not, we may as well just enable it.

kaine119:
There aren't any violations of this, don't think [`Kernel::format`](http://ruby-doc.org/core-2.5.3/Kernel.html#method-i-format) is used anywhere.

kaine119:
Should I just delete the config line for this rule?

PR #2184, opened 2018-11-01
[WIP] Enforce rubocop Rails/Timezone rule

kaine119:
Enable the Rubocop `Rails/Timezone` rule.

This rule breaks date validation in `app/models/campaign.rb`, so for now I've ignored the problem line

(from #2173)


ragesoss:
Could you rebase this one now? The other timezone one definitely looks fine so I merged it. For this one, I'd like to poke around with it first, to see whether updating the spec is a better option.

Issue #2082, opened 2018-10-10
Write tests for Redux reducers that cover all action types and ensure that state is not mutated


ragesoss:
The reducers that make up the main Redux reducer should all be pure functions that are free of side effects and do not mutate their inputs. This makes unit tests particularly easy to implement, but important for making sure we avoid state-mutation bugs.

The handful of existing specs in `test/reducers` offer clear examples for how individual actions can be tested for these reducers, with `deepFreeze` being called on the initial state to guarantee that nothing in the reducer is mutating it.

Choose one or more reducers, and add tests for the functionality of each action type supported by that reducer. This should be done one reducer per commit. If you're working on this, please post a comment to say which reducer you're working on, so that others can work on different reducers.

Ideally, the set of tests for each reducer will provide 100% coverage of that reducer. You can check the coverage by running `jest` and then opening `coverage/lcov-report/index.html`, which includes coverages maps of each javascript file.

sunny-lan:
(GCI 2018)

I'll be making tests for `uploads.js`

sunny-lan:
Is it possible to get the tests working without setting up the whole env?
I tried just doing `yarn` and `gulp` to build, but that results in a large amount of errors from gulp concerning bundler (I have bundler installed)

ragesoss:
@sunny-lan you must do `bundle install` before `gulp` will run, because one of the steps in the gulp pipeline involves calling the ruby script that generates the JS versions of the interface message translations.

ragesoss:
That said, I don't think the javascript tests even depend on generating the assets. If you did `yarn` and it ran without error, you should be able to do `yarn test` to just run the jest tests.

kagehina919:
@ragesoss can I work on adding some of the tests? I have set up the development environment (and opened a PR for #2197).

AZReed:
I'll take user related reducers:
user_profile
user_revisions
user_training_status
user

If someone is doing these let me know.

ragesoss:
Here's what the current coverage on the reducers looks like:
![reducer coverage](https://user-images.githubusercontent.com/848483/161113266-582ef796-217f-455d-a190-73ffb5af34ac.png)

shruti862:
@ragesoss 
I want to work for adding test for some of the reducers.

fillingtothemomo:
@ragesoss  I would like to pick up this issue, can you assign it to me.

ragesoss:
[...]

PR #2208, opened 2018-11-07
Test reducer course.spec.js

AZReed:
100% coverage for course reducer.

First pull request of my life actually, so I'm assuming I miss something or I'm not following some rule. Feel free to point me in the right direction, it will be appreciated.

### DIFF
--- a/test/reducers/course.spec.js
+++ b/test/reducers/course.spec.js
@@ -1,9 +1,40 @@
 import deepFreeze from 'deep-freeze';
 import course from '../../app/assets/javascripts/reducers/course';
-import { UPDATE_COURSE } from '../../app/assets/javascripts/constants';
+import { UPDATE_COURSE, RECEIVE_COURSE, RECEIVE_COURSE_UPDATE, PERSISTED_COURSE } from '../../app/assets/javascripts/constants';
 import '../testHelper';
 
 describe('course reducer', () => {
+
+  it('receives a course via RECEIVE_COURSE', () => {
+    const initialState = {};
+    const action = { type: RECEIVE_COURSE, data: { course: { title: 'title' } }  };
+    deepFreeze(initialState);
+
+    const newState = course(initialState, action);
+    expect(newState).to.deep.eq(action.data.course);
+  });
+
+  it('receives a course update via RECEIVE_COURSE_UPDATE with newStats', () => {
+    const initialState = {};
+    const action = { type: RECEIVE_COURSE_UPDATE, data: { course: { title: 'title' } }  };
+    deepFreeze(initialState);
+
+    const newStats = [ "createdCount", "editedCount", "editCount", "studentCount", "wordCount", "viewCount", "uploadCount"];
+
+    const newState = course(initialState, action);
+    expect(newState.title).to.deep.eq(action.data.course.title);
+    expect(newState.newStats).to.have.all.keys(newStats);
+  });
+
+  it('persisted course via PERSISTED_COURSE', () => {
+    const initialState = {description: 'initial description'};
+    const action = { type: PERSISTED_COURSE, data: { course: { title: 'new title' } }  };
+    deepFreeze(initialState);
+
+    const newState = course(initialState, action);
+    expect(newState).to.deep.eq({description: 'initial description', title: 'new title'})
+  });
+
   it('updates an attribute via UPDATE_COURSE', () => {
     const initialState = { title: 'old title', term: 'old term' };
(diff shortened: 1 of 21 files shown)


test/reducers/course.spec.js
```diff
@@ -1,16 +1,225 @@
...
+      end: null,
+      timeline_start: null,
+      timeline_end: null,
+      day_exceptions: '',
+      weekdays: '0000000',
+      editingSyllabus: false
+    };
+
+    const newState = course(undefined, { type: null });
+    expect(newState).to.deep.eq(expectedState);
```
ragesoss: This tests is tied to the exact form of the current initialState from the reducer. A better approach would just be to test for the presence of a few properties that you expect. In particular, something like expecting it to have keys for title and description, and maybe expect that weekdays is '0000000'. That will leave it open for extension if new properties get added to the initial state, which should not break the test.

test/reducers/course.spec.js
```diff
@@ -1,16 +1,225 @@
...
+      type: RECEIVE_COURSE,
+      data: { course: { title: 'title' } }
+    };
+    deepFreeze(initialState);
+
+    const newState = course(initialState, mockedAction);
+    expect(newState).to.deep.eq(mockedAction.data.course);
+  });
+
+  it('should handle RECEIVE_COURSE_UPDATE with newStats', () => {
```
ragesoss: The point with this one should be to test the two possible cases: one where stats don't change and the current state is the same as the newly received data (which means newStats should equal false); and one where the stats do change (so newStats should equal true).

You probably want to test this with something like `{ title: 'title', createdCount: 0, editedCount: 0 }` as this initial state, and then for the action, update one of those and expect its newStats entry to be true, and don't update the other and expect it's newStats entry to be false.

ragesoss: The description here could be more precise: it 'keeps track of which stats have updated with RECEIVE_COURSE_UPDATE' or something like that.

test/reducers/course.spec.js
```diff
@@ -1,16 +1,225 @@
...
+
+    const newState = course(initialState, mockedAction);
+    expect(newState.title).to.deep.eq(mockedAction.data.course.title);
+
+    Object.keys(newState.newStats).forEach((key) => {
+      expect(newState.newStats[key]).to.eq(false);
+    });
+  });
+
+  it('should handle PERSISTED_COURSE', () => {
```
ragesoss: This could be more descriptive. The idea behind it is something like 'updates the course state with newly received data'.

test/reducers/course.spec.js
```diff
@@ -1,16 +1,225 @@
...
+    deepFreeze(initialState);
+
+    const newState = course(initialState, mockedAction);
+    expect(newState).to.deep.eq({
+      description: 'initial description',
+      title: 'new title'
+    });
+  });
+
+  it('should handle UPDATE_COURSE', () => {
```
ragesoss: it 'handles UPDATE_COURSE by adding or updating properties from the action, and leaves other properties as they are', or something along those lines.

test/reducers/course.spec.js
```diff
@@ -1,16 +1,225 @@
...
+      default_course_type: 'foobar',
+      default_passcode: 'pass'
+    };
+    const mockedAction = {
+      type: RECEIVE_INITIAL_CAMPAIGN,
+      data: { campaign: campaign }
+    };
+    deepFreeze(initialState);
+
+    const expectedState = {
```
ragesoss: The idea here is that it uses the initial campaign to provide a default course description, type, and passcode, and also adds the properties for the initial campaign id and title.

Instead of expecting deep equality using the same code as from the reducer — which ties it closely to the implementation — this should pass in an initial state that is more like `{ title: 'default title', description: 'default description', type: 'default type', passcode: 'default passcode' }, and then check that the description, type and passcode got changed, the title was't changed, and the new properties are there and have the expected values.

Avoiding `deep.eq` in this case is a way to keep the test decoupled from the precise implementation, and will make the code more amenable to change without breaking the test.

ragesoss:
@AZReed it looks like you've merge from master and introduced some conflicts.

It's usually easier to handle this kind of thing if you rebase rather than merge: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/git.md

On a quick look, your changes look good. I'll take a final look once it's cleaned up of conflicts.

AZReed:
yep, don't know what happened exactly but looking into it, thanks

PR #2209, opened 2018-11-08
Add tests for reducer alerts

ledestin:
I personally think that `deepFreeze` should go right after the `initialState` declaration as it's easy to forget. Otherwise, the PR is ready.

#2082


ledestin:
@ragesoss thanks! Updated.

PR #2217, opened 2018-11-12
User related test for reducers and refactorings

AZReed:
Covers tests for reducers:

- user_profile.js
- user_revisions.js
- user_training_status.js
- users.js

### DIFF
--- a/app/assets/javascripts/reducers/user_training_status.js
+++ b/app/assets/javascripts/reducers/user_training_status.js
@@ -1,6 +1,6 @@
 import { RECEIVE_USER_TRAINING_STATUS } from '../constants';
 
-const initialState = [];
+const initialState = {};
 
 export default function userTrainingStatus(state = initialState, action) {
   switch (action.type) {
--- a/app/assets/javascripts/reducers/user_training_status.js
+++ b/app/assets/javascripts/reducers/user_training_status.js
@@ -1,6 +1,6 @@
 import { RECEIVE_USER_TRAINING_STATUS } from '../constants';
 
-const initialState = {};
+const initialState = [];
 
 export default function userTrainingStatus(state = initialState, action) {
   switch (action.type) {
--- a/test/reducers/user_profile.spec.js
+++ b/test/reducers/user_profile.spec.js
@@ -1,25 +1,24 @@
 import '../testHelper';
 import deepFreeze from 'deep-freeze';
-import user_profile from '../../app/assets/javascripts/reducers/user_profile';
+import userProfile from '../../app/assets/javascripts/reducers/user_profile';
 import { RECEIVE_USER_PROFILE_STATS } from '../../app/assets/javascripts/constants';
 
 describe('user profile reducer', () => {
   it('should return initial state when no action nor state is provided', () => {
-    const newState = user_profile(undefined, { type: null });
-    expect(newState.stats).to.an('object');
+    const newState = userProfile(undefined, { type: null });
+    expect(newState.stats).to.be.an('object');
     expect(newState.isLoading).to.eq(true);
   });
 
   it('receives stats from action and updates state attribute with RECEIVE_USER_PROFILE_STATS', () => {
-    const initialState = { title: 'title' };
+    const initialState = { isLoading: true };
     deepFreeze(initialState);
     const mockedAction = {
       type: RECEIVE_USER_PROFILE_STATS,
       data: {},
-      isLoading: true
     };
 
-    const newState = user_profile(initialState, mockedAction);
+    const newState = userProfile(initialState, mockedAction);
     expect(newState.isLoading).to.eq(false);
     expect(newState.stats).to.be.an('object');
   });
--- a/test/reducers/user_profile.spec.js
+++ b/test/reducers/user_profile.spec.js
@@ -1,0 +1,26 @@
+import '../testHelper';
+import deepFreeze from 'deep-freeze';
+import user_profile from '../../app/assets/javascripts/reducers/user_profile';
+import { RECEIVE_USER_PROFILE_STATS } from '../../app/assets/javascripts/constants';
+
+describe('user profile reducer', () => {
+  it('should return initial state when no action nor state is provided', () => {
+    const newState = user_profile(undefined, { type: null });
+    expect(newState.stats).to.an('object');
+    expect(newState.isLoading).to.eq(true);
+  });
+
+  it('receives stats from action and updates state attribute with RECEIVE_USER_PROFILE_STATS', () => {
+    const initialState = { title: 'title' };
+    deepFreeze(initialState);
+    const mockedAction = {
+      type: RECEIVE_USER_PROFILE_STATS,
+      data: {},
+      isLoading: true
+    };
+
+    const newState = user_profile(initialState, mockedAction);
+    expect(newState.isLoading).to.eq(false);
+    expect(newState.stats).to.be.an('object');
+  });
+});
--- a/test/reducers/user_revisions.spec.js
+++ b/test/reducers/user_revisions.spec.js
@@ -1,27 +1,30 @@
 import deepFreeze from 'deep-freeze';
 import '../testHelper';
-import user_revisions from '../../app/assets/javascripts/reducers/user_revisions';
+import userRevisions from '../../app/assets/javascripts/reducers/user_revisions';
 import { RECEIVE_USER_REVISIONS } from '../../app/assets/javascripts/constants';
 
 describe('user revisions reducer', () => {
   it('should return initial state when no action nor state is provided', () => {
-    const newState = user_revisions(undefined, { type: null });
+    const newState = userRevisions(undefined, { type: null });
     expect(newState).to.be.an('object');
   });
 
   it('receives user revisions data with user identification as key via RECEIVE_USER_REVISIONS', () => {
-    const initialState = {
-      title: 'title'
-    };
+    const initialState = {};
     deepFreeze(initialState);
     const mockedAction = {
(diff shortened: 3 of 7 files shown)


app/assets/javascripts/reducers/user_training_status.js
```diff
@@ -1,6 +1,6 @@
 import { RECEIVE_USER_TRAINING_STATUS } from '../constants';
 
-const initialState = [];
+const initialState = {};
```
ragesoss: The data received for this reducer is an array, so we should keep the initial state in the same form.

test/reducers/user_profile.spec.js
```diff
@@ -0,0 +1,26 @@
+import '../testHelper';
+import deepFreeze from 'deep-freeze';
+import user_profile from '../../app/assets/javascripts/reducers/user_profile';
+import { RECEIVE_USER_PROFILE_STATS } from '../../app/assets/javascripts/constants';
+
+describe('user profile reducer', () => {
+  it('should return initial state when no action nor state is provided', () => {
+    const newState = user_profile(undefined, { type: null });
+    expect(newState.stats).to.an('object');
```
ragesoss: missing a `be`?

test/reducers/user_profile.spec.js
```diff
@@ -0,0 +1,26 @@
...
+    expect(newState.isLoading).to.eq(true);
+  });
+
+  it('receives stats from action and updates state attribute with RECEIVE_USER_PROFILE_STATS', () => {
+    const initialState = { title: 'title' };
+    deepFreeze(initialState);
+    const mockedAction = {
+      type: RECEIVE_USER_PROFILE_STATS,
+      data: {},
+      isLoading: true
```
ragesoss: This should be in the initialState, rather than the action.

test/reducers/user_revisions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+  it('receives user revisions data with user identification as key via RECEIVE_USER_REVISIONS', () => {
+    const initialState = {
+      title: 'title'
+    };
+    deepFreeze(initialState);
+    const mockedAction = {
+      type: RECEIVE_USER_REVISIONS,
+      data: { course: { revisions: true } },
+      userId: 'username'
+    };
```
ragesoss: The actual data looks a little different. `userId` is an integer, and the value of `revisions` is an array (of objects).

test/reducers/user_revisions.spec.js
```diff
@@ -0,0 +1,27 @@
+import deepFreeze from 'deep-freeze';
+import '../testHelper';
+import user_revisions from '../../app/assets/javascripts/reducers/user_revisions';
```
ragesoss: As a convention and to match how it's used in the app, this import should match the name of the imported function in the other file: `userRevisions`.

test/reducers/user_revisions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    };
+    deepFreeze(initialState);
+    const mockedAction = {
+      type: RECEIVE_USER_REVISIONS,
+      data: { course: { revisions: true } },
+      userId: 'username'
+    };
+
+    const newState = user_revisions(initialState, mockedAction);
+    expect(newState.username).to.eq(true);
```
ragesoss: It took me a while to understand why this is passing right now...

The idea is that userRevisions holds a set of revisions for any number of users, accessible via their user ID. So it should be something like newState[123] is equal to `[]`.

test/reducers/user_revisions.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    expect(newState.title).to.eq('title');
```
ragesoss: Title is not part of this bit of state. The actual initial state is an empty object. If you want to test how it behaves when a second action is received, it would be better to first receive data for one user, than then for a different user, and then expect that the data for the first user is also still present.

test/reducers/user_training_status.spec.js
```diff
@@ -0,0 +1,23 @@
...
+import { RECEIVE_USER_TRAINING_STATUS } from '../../app/assets/javascripts/constants';
+
+describe('user training status reducer', () => {
+  it('should return initial state when no action nor state is provided', () => {
+    const newState = user_training_status(undefined, { type: null });
+    expect(newState).to.be.an('object');
+  });
+
+  it('returns user training status with RECEIVE_USER_TRAINING_STATUS', () => {
+    const initialState = { title: 'title' };
```
ragesoss: Here initial state is just an empty array. This particular reducer only is designed to hold a single array of data (for the user whose page we're on).

test/reducers/user_training_status.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    deepFreeze(initialState);
+    const mockedAction = {
+      type: RECEIVE_USER_TRAINING_STATUS,
+      data: { user: { training_modules: { key: 'value' } } }
```
ragesoss: The expected data shape here is that training_modules is a an array of objects.

test/reducers/users.spec.js
```diff
@@ -0,0 +1,79 @@
...
+describe('users reducer', () => {
+  it('should return initial state when no action nor state is provided', () => {
+    const newState = users(undefined, { type: null });
+    expect(newState.users).to.be.an('array');
+    expect(newState.isLoaded).to.eq(false);
+    expect(newState.sort).to.be.an('object');
+  });
+
+  it('returns the previous state and updates users array from action via RECEIVE_USERS, ADD_USER and REMOVE_USER', () => {
+    const initialState = { title: 'title', isLoaded: false };
```
ragesoss: This isn't the shape of the initial state. If you want a shortcut to it, you can just do `initialState = users(undefined, { type: null });`

test/reducers/users.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      type: RECEIVE_USERS,
+      data: { course: { users: [1, 2] } }
+    };
+
+    let newState = users(initialState, mockedAction);
+    expect(newState.title).to.eq('title');
+    expect(newState.users).to.deep.eq([1, 2]);
+    expect(newState.isLoaded).to.eq(true);
+
+    mockedAction.type = ADD_USER;
```
ragesoss: Reusing the same action and mutating it is bit confusing here, since that's now how actions get used in the app itself. I think making a new, more specifically named action would make this clearer.

test/reducers/users.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    expect(newState.isLoaded).to.eq(false);
+    expect(newState.sort).to.be.an('object');
+  });
+
+  it('returns the previous state and updates users array from action via RECEIVE_USERS, ADD_USER and REMOVE_USER', () => {
+    const initialState = { title: 'title', isLoaded: false };
+    deepFreeze(initialState);
+    const mockedAction = {
+      type: RECEIVE_USERS,
+      data: { course: { users: [1, 2] } }
```
ragesoss: This and the following actions should have a shape that matches how the app behaves more closely. In particular, the users in this array are objects (with `id`, `username`, `role`, and other properties, if you want to get that detailed).

test/reducers/users.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    };
+
+    let newState = users(initialState, mockedAction);
+    expect(newState.title).to.eq('title');
+    expect(newState.users).to.deep.eq([1, 2]);
+    expect(newState.isLoaded).to.eq(true);
+
+    mockedAction.type = ADD_USER;
+    mockedAction.data.course.users = [1, 2, 3];
+    newState = users(initialState, mockedAction);
```
ragesoss: If we're going to test a series of related actions, we may as well test it as a series, with the state output from the previous action used as the initial state for the next one.

test/reducers/users.spec.js
AZReed: Should each type be in its own block? It's kinda big chunk for one block on a second look.

test/reducers/users.spec.js
AZReed: In these three cases the reducer returns the exact same users array passed from the action. Does it make sense to test it as a series when the output is going to be the same as the input in every test?

ragesoss:
Either way. Doing them in series would match how they actually get used,
but the are handled in a way that it doesn't actually matter.

On Nov 13, 2018 8:06 PM, "AZReed" <notifications@github.com> wrote:

*@AZReed* commented on this pull request.
------------------------------

In test/reducers/users.spec.js
<https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2217#discussion_r233308542>
:

> +  it('returns the previous state and updates users array from action via RECEIVE_USERS, ADD_USER and REMOVE_USER', () => {
+    const initialState = { title: 'title', isLoaded: false };
+    const mockedAction = {
+      type: RECEIVE_USERS,
+      data: { course: { users: [1, 2] } }
+    };
+
+    let newState = users(initialState, mockedAction);
+    expect(newState.title).to.eq('title');
+    expect(newState.users).to.deep.eq([1, 2]);
+    expect(newState.isLoaded).to.eq(true);
+
+    mockedAction.type = ADD_USER;
+    mockedAction.data.course.users = [1, 2, 3];
+    newState = users(initialState, mockedAction);

In these three cases the reducer returns the exact same users array passed
from the action. Does it make sense to test it as a series when the output
is going to be the same as the input in every test?

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2217#discussion_r233308542>,
or mute the thread

PR #2221, opened 2018-11-13
Training active record

### LINKED ISSUE #1759: Convert TrainingModule to ActiveRecord
Following the recent conversion of TrainingSlide (see https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1732 ), we should convert TrainingModule into a database-backed model as well.

AmitJoki:
Fixes #1759


db/schema.rb
```diff
@@ -10,9 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2018_08_30_173705) do
+ActiveRecord::Schema.define(version: 2018_11_12_100217) do
 
-  create_table "alerts", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+  create_table "alerts", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4", force: :cascade do |t|
```
ragesoss: lots of extraneous changes to the DB here. Only the `create_table "training_modules"` block and the version bump should be included. (Also, the default charset is still just utf8 in production, and we should probably keep that consistent with the new table as well.)

PR #2226, opened 2018-11-17
Converted TrainingModule to ActiveRecord

### LINKED ISSUE #1759: Convert TrainingModule to ActiveRecord

AmitJoki:
Fixes #1759

### DIFF
--- a/app/models/training_module.rb
+++ b/app/models/training_module.rb
@@ -61,7 +61,7 @@ class TrainingModule < ApplicationRecord
     # Reload the requested module's slides
     training_module = TrainingModule.find_by(slug: slug)
     raise ModuleNotFound, "No module #{slug} found!" unless training_module
-    # TrainingSlide.load(slug_list: training_module.slide_slugs)
+    TrainingSlide.load(slug_list: training_module.slide_slugs)
   end
 
   def self.inflate(content, slug, wiki_page = nil) # rubocop:disable Metrics/MethodLength


spec/lib/training/training_base_spec.rb
```diff
@@ -54,30 +53,6 @@
       end
     end
 
-    context 'when there are duplicate slugs' do
```
ragesoss: These specs are testing important things. Maybe they can be moved to specs for the new class, instead of being tests of TrainingBase?

AmitJoki:
I have made the requested changes. Have a look at it when you can :blush:

lib/training/training_base.rb
```diff
@@ -33,17 +33,10 @@ def self.load(slug_list: nil, content_class: self)
...
-      TrainingModule.all.each { |tm| TrainingSlide.load(slug_list: tm.slide_slugs) }
-      TrainingModule.flush
-      TrainingModule.load
-      TrainingLibrary.load
-    else
-      TrainingLibrary.load
-      TrainingModule.load
-      TrainingSlide.load
-    end
+    TrainingSlide.delete_all
```
ragesoss: Did you find this to be necessary? I'd really like to avoid any situation where slides need to be deleted; the idea is that we overwrite the existing records instead, whenever we need to update them. The same should be true for the modules.

lib/training/wiki_training_loader.rb
```diff
@@ -35,10 +35,9 @@ def load_from_wiki
                           level: 'info', extra: { wiki_pages: source_pages }
 
     thread_count = [CONCURRENCY, source_pages.count].min
-    threads = source_pages.in_groups(thread_count, false).map.with_index do |wiki_page_group, i|
-      Thread.new(i) { add_trainings_to_collection(wiki_page_group) }
+    source_pages.in_groups(thread_count, false).each do |wiki_page_group|
```
ragesoss: Since this gets reid of the conccurency altogether, there's some additional cleanup above here that needs to be done. The constant and thread_count is now useless.

lib/training/training_base.rb
ragesoss: I think we should only need to flush the TrainingLibrary after this. For module and slide, we can now just do `.load` and it should overwrite the old records.

But for wiki trainings, it might still be relevant to do `TrainingModule.all.each { |tm| TrainingSlide.load(slug_list: tm.slide_slugs) }` instead of Slide.load.

app/models/training_module.rb
```diff
@@ -0,0 +1,115 @@
...
+    TrainingBase.load_all
+  end
+
+  def self.flush
+    delete_all
+  end
+
+  # This reloads all the library and module content, but only updates the slides
+  # for the module with the given slug.
+  def self.reload_module(slug:)
```
ragesoss: I think this can be simplified, now that TrainingModule is ActiveRecord, instead of long-lived mutable objects in shared memory. We shouldn't ever need to use `.flush`, or `load` without the slug list, except for with TrainingLibrary.

A big part of the benefit of doing this transition to ActiveRecord is to simplify the reloading process and remove unnecessary slow network calls to fetch and refetch Wikipedia content.

app/models/training_module.rb
```diff
@@ -55,15 +55,15 @@ def self.reload_module(slug:)
     # and can load slides for brand-new modules.
     TrainingLibrary.flush
     TrainingLibrary.load
-    self.load(slug: slug)
+    load(slug: slug)
```
ragesoss: Whoops. I thought `load` was a Ruby keyword (in which case `self.` would be necessary), but it's actually just a method on `Kernel`, so it's just being overridden by the `load` method in this file. Sorry about that!

app/models/training_module.rb
```diff
@@ -37,7 +37,7 @@ def self.trim_id_from_filename
   end
 
   def self.load(slug: nil)
-    TrainingBase.load(content_class: self, slug_list: slug)
+    TrainingBase.load(content_class: self, slug_list: slug.nil? ? nil : [slug])
```
ragesoss: Is there anywhere that calls `load` without the argument, or can we remove the default and just always pass `[slug]`?

app/models/training_module.rb
```diff
@@ -55,15 +55,15 @@ def self.reload_module(slug:)
     # and can load slides for brand-new modules.
     TrainingLibrary.flush
     TrainingLibrary.load
-    TrainingModule.load(slug: [slug])
+    self.load(slug: slug)
```
ragesoss: I'm testing this out, and I was wrong about being able to do this based on the slug only. It doesn't actually know the slugs from the wiki page that defines the available modules, because they are all json pages that don't embed the slug (unlike the index of slides, which do). So I think we'll need to load all the modules each time (but still, only the slides for the selected module).

ragesoss:
For testing this out:

```
# application.yml
  wiki_education: 'false'
  training_slides_wiki_page: 'Training modules/dashboard/slides'
  training_modules_wiki_page: 'Training modules/dashboard/modules'
  training_libraries_wiki_page: 'Training modules/dashboard/libraries'
```

You may need to do `TrainingModule.destroy_all` and `TrainingSlide.destroy_all` before testing.

Then visit (for example) http://localhost:3000/reload_trainings?module=keeping-events-safe

That should load all the slides for that module, and put it in working condition. I tweaked the `load` method to load all the modules (per my comment, I was confused about how the slug_list could work here), but after that I still ran into problems with the `description` of each module being `nil` (which also breaks rendering when you visit a module page). I'm not sure what is happening differently, but it's not following through to the wiki page to get the descriptioon.

For example, this module:
https://meta.wikimedia.org/wiki/User:JSutherland_(WMF)/dashboard_modules/keeping_events_safe.json

It should be getting the description (and, as a next step, translations) from https://meta.wikimedia.org/wiki/Training_modules/Keeping_events_safe/description.

I'm not sure what's happening differently, as the non-ActiveRecord version does load the description as expected (although it doesn't do the description translations).

app/models/training_module.rb
AmitJoki: The `load` with arguments is used only in `reload_module`. Otherwise it is just `.load`

AmitJoki:
@ragesoss I've always wondered why you had a dislike towards meta-programming. I now know why. So the problem was basically this.

I was migrating TrainingModule to ActiveRecord blindly. I referred to [TrainingSlide.rb#L59](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/models/training_slide.rb#L59) and had no idea what it was doing, which also led me into believing there was no translations field as I only focused on what was available in the `lib/training_module`.

I then proceeded without wanting to use the meta-programming and was doing assignments in the old fashioned way like
```
training_module.description = content['description']
```

But there was an anomaly as to how the keys were stored due to this: [WikiTrainingLoader.rb](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/lib/training/wiki_training_loader.rb#L142)

While all other attributes were using String keys, the aforementioned code had Symbols as keys. So there was no `content['description']` but `content[:description]`. It worked fine during TrainingSlide because at [Line 60](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/models/training_slide.rb#L60) the String Interpolation implicitly converted Symbol to String.

The old way worked because of how Hashugar works.
```
{ "string" => 1, symbol: 2 }.to_hashugar => { "string" => 1, "symbol" => 2 }
```

The fix was simple. I maintained the uniformity of keys as String and it works alright now! :smile:

ragesoss:
I *like* metaprogramming... it's really cool and fun and flexible. Just don't want any more of it that is completely necessary in my production code. :-)

ragesoss:
I tried this again, and I first loaded the `keeping-events-safe` module and it works smoothly for the modules. (And it saved the translations, yay! Next step will be to make the training module *show* page use those translations.)

But when I tried to load a second module 'defining-your-event' that I didn't already have imported slides for, none of the slides loaded. I guess that's because you have the slide loading step commented out.

Looks like it's working as expected once that is re-enabled.

I'll plan to merge and deploy this next week, as I want to have plenty of time to troubleshoot if any problems happen in production.

PR #2231, opened 2018-11-19
reducer articles test

AZReed:

100% coverage for reducer articles test

### DIFF
--- a/test/reducers/articles.spec.js
+++ b/test/reducers/articles.spec.js
@@ -1,0 +1,121 @@
+import deepFreeze from 'deep-freeze';
+import articles from '../../app/assets/javascripts/reducers/articles';
+import {
+  RECEIVE_ARTICLES,
+  SORT_ARTICLES,
+  SET_PROJECT_FILTER,
+  SET_NEWNESS_FILTER
+} from '../../app/assets/javascripts/constants/articles';
+import '../testHelper';
+
+const articles_array = [
+  {
+    id: 1,
+    title: 'articles_1',
+    new_article: false,
+    language: 'us',
+    project: 4
+  },
+  {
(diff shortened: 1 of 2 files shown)


test/reducers/articles.spec.js
```diff
@@ -0,0 +1,121 @@
...
+} from '../../app/assets/javascripts/constants/articles';
+import '../testHelper';
+
+const articles_array = [
+  {
+    id: 1,
+    title: 'articles_1',
+    new_article: false,
+    language: 'us',
+    project: 4
```
ragesoss: `project` here will be a string like `wikipedia` or `wiktionary` or `wikidata`.

test/reducers/articles.spec.js
```diff
@@ -0,0 +1,121 @@
...
+    expect(newState.articles[2].id).to.eq(3);
+  });
+
+  it('sets filter conditionally with SET_PROJECT_FILTER', () => {
+    const initialState = {};
+    deepFreeze(initialState);
+    const mockedAction = {
+      type: SET_PROJECT_FILTER,
+      wiki: {
+        project: 'all'
```
ragesoss: The `wiki` object here will have both `project` and `language` properties.

test/reducers/articles.spec.js
```diff
@@ -0,0 +1,121 @@
...
+      }
+    };
+
+    const state = articles(initialState, mockedAction);
+    expect(state.wikiFilter).to.eq(null);
+
+    const newMockedAction = {
+      type: SET_PROJECT_FILTER,
+      wiki: {
+        project: ''
```
ragesoss: as above, project and language.

PR #2232, opened 2018-11-20
Convert ValidationStore to redux, finishing the Flux->redux migration

ragesoss:
This set of changes converts the remaining uses of the McFly Flux library to redux.

The final, tricky bit was ValidationStore, which was responsible for storing validation state for inputs, with that state passed to the inputs themselves via `InputHOC`. ValidationStore could be mutated both through actions and via several functions, most importantly `.isValid()` — which would take unactivated validations (such as for the initial state of empty but required inputs) and mark them as 'changed' so that they would then show the 'invalid' class. This approach appears in handling the 'Save' button for forms, so that required inputs only get highlighted as invalid if the user tries to save without filling them in. It also relied on storeDidChange to update the validation state on input components.

This now all goes through redux actions, with change propagation done via the standard redux way of passing redux state as props with `connect`. The tricky part here is limiting the amount of rerendering necessary, to avoid making the connected inputs too slow. The approach I took, which I hope covers all the use cases, is to check for changes in the `editable` prop and the three core bits of state tracked by each instance of InputHOC: `value`, `invalid`, and `id`. Thus, inputs that don't have their value or validaton status change, and don't switch to/from editable mode, will skip re-renders when other parts of the redux validation state changes.

This incorporates some of Pratyush Singhal's work from https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1722 , which was very helpful in sorting out how to handle `connect` for high order components, and which provided a good starting point for the validations reducer.


ragesoss:
FYI @psinghal20 :-)

PR #2234, opened 2018-11-23
Special users

AmitJoki:
This PR displays the special users and also provides the UI to add them.


ragesoss:
This is working smoothly now. On a quick look through the build results and the PR, there are a few things we'll need to fix before merging:

* Most of the new functionality is uncovered by tests. (See the uncovered lines listed at the end of the build for job #2.) I suggest adding a feature spec for adding and removing special users, as well as adding a non-existent user, to start with (and see whether there are remaining coverage gaps after that).
* The `set` and `remove` methods on `SpecialUsers` make it more into a model instead of presenter. One way around this could be to move those methods (like `set_special_user` and `remove_special_user`) into `Setting` itself.
* I'd rather avoid coupling `User` to SpecialUsers through the new `.is?` method. That seems like it ought to be the job of the SpecialUsers presenter itself, like with a new method that takes the user and the role and provides the same answer.
* The git history now doesn't include attribution for bekicot. I suggest squashing the entire PR into one commit, and then including attribution and a link to #1839 in the commit message.

PR #2236, opened 2018-11-23
Add the ability to embed vega graphs

### LINKED ISSUE #2233: Provide a way to embed a widget of stats from a course into another website
The main set of course stats that is shown at the top of a course page is useful info for program organizers who are showing off the impact of their programs. Users sometimes take screenshots so that they can highlight these stats other places. It would be useful if, instead, they could use a snippet of code to embed a widget on an arbitrary site that renders the latest stats for a given course.

I'm not at all sure what the best way to do that is.

AmitJoki:
This PR could be an extra added feature for those looking to embed the vega graphs.

### DIFF
--- a/app/assets/javascripts/components/articles/course_quality_progress_graph.jsx
+++ b/app/assets/javascripts/components/articles/course_quality_progress_graph.jsx
@@ -2,6 +2,7 @@
 import React from 'react';
 import createReactClass from 'create-react-class';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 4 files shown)


ragesoss:
The stats people want to embed are the ones rendered by CourseStats.

ragesoss:
I'm going to close this, as I don't think we have any demand for embedding the graphs at the moment, and if/when we do, it should be pretty easy to implement following the pattern of the other already-merged embed feature.

PR #2237, opened 2018-11-25
Adds the ability to embed the course stats

### LINKED ISSUE #2233: Provide a way to embed a widget of stats from a course into another website

AmitJoki:
Fixes #2233


ragesoss:
I tried this out, but I get blocked by CORS whenever I try to embed it on another website. It did work when I embedded it on a different page within the development environment.

I have some questions and ideas related to the implementation...

Doing it as an iframe seems fine. Are there any other options that are worth considering, like just using a `script` tag for the embed? I think that's possible, I'm just not sure how complex it is.

I don't think sending the entire main bundle makes sense just to render a single, static component. If we're going to do it as an iframe, we probably don't even need javascript at all. The campaign stats, which look the same style-wise, get rendered by the `courses/_header_stats.html.haml` template. That approach could be used here as well, with a new haml template mostly copied from that one. Going with that approach, we'd probably also want to have the stats embed have its own route and controller that doesn't interact with the existing CoursesController and routing.

Basically, my biggest priority for how this gets implemented is to isolate it from the rest of the application as much as possible.

ragesoss:
I get this when I try to embed it on (for example) travis-ci.org or wikiedu.org.

```
Blocked loading mixed active content “http://localhost:3000/stats/courses/University_of_Chicago/History_of_Skepticism_(Spring_2018)”
```

ragesoss:
This looks good code-wise. The one thing I'd like to change to further isolate it is to put it in its own controller instead of in CoursesController (and put the view in its own directory instead of `views/courses`).

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -189,6 +190,9 @@ const AvailableActions = createReactClass({
       controls.push((
         <p key="download_course_stats"><CourseStatsDownloadModal course={course} /></p>
       ));
+      controls.push((
+        <EmbedStatsButton title={course.title} key="embed_course_stats"/>
```
ragesoss: Needs to be wrapped in a `<p>` to the margin is right:

button margin

app/assets/javascripts/components/overview/embed_stats_button.jsx
```diff
@@ -0,0 +1,63 @@
...
+          />
+          <small>{this.state.status}</small>
+        </td>
+      </tr>
+    );
+
+    return (
+      <div className="pop__container embed_stats">
+        <a onClick={this.props.open} className="button">
+          Embed Course Stats&nbsp;&lt;/&gt;
```
ragesoss: i18n

config/locales/en.yml
```diff
@@ -433,6 +433,9 @@ en:
     duration: Course Duration
     edit_count: Edit Count
     edit_course_dates: Edit Course Dates
+    embed_course_stats: Embed Course Stats
+    embed_course_stats_heading: To Embed Course Stats
```
ragesoss: Maybe instead of a header and then more copy, just the full explanation without a header would be clearer:

`To embed the stats for this program in another website, copy and paste this code snippet:`

AmitJoki:
Have pushed the requested changes.

ragesoss:
Okay! I'll merge and deploy, and then we'll see if we can get some feedback.

PR #2243, opened 2018-11-27
Special Users End Point

AmitJoki:
This PR adds the ability to list and add/remove Special Users.


app/presenters/special_users.rb
```diff
@@ -6,23 +6,38 @@ def self.special_users
...
+
+  def self.remove(role)
+    users = Setting.find_or_create_by(key: 'special_users')
+    users.value.delete(role)
+    users.save
+  end
+
   def self.communications_manager
-    User.find_by(username: special_users[:communications_manager])
+    User.find_by(username: special_users['communications_manager'])
```
ragesoss: These changes from symbol to string will change the existing behavior, because all the current (manually-set) values in the production database were done with symbols.

PR #2248, opened 2018-11-28
Fix pagination in Uploads

### LINKED ISSUE #2028: Pagination element disappears if we go to another tab and then come back to Uploads

aSquare14:
fixes #2028

### DIFF
--- a/app/assets/javascripts/components/uploads/uploads_handler.jsx
+++ b/app/assets/javascripts/components/uploads/uploads_handler.jsx
@@ -13,18 +13,19 @@ const UPLOADS_PER_PAGE = 100;
 
 const UploadsHandler = createReactClass({
   displayName: 'UploadsHandler',
-
   propTypes: {
     course_id: PropTypes.string,
     course: PropTypes.object
   },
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/uploads/uploads_handler.jsx
```diff
@@ -13,18 +13,19 @@ const UPLOADS_PER_PAGE = 100;
...
 const UploadsHandler = createReactClass({
   displayName: 'UploadsHandler',
-
   propTypes: {
     course_id: PropTypes.string,
     course: PropTypes.object
   },
 
   getInitialState() {
+    const length_selecteduploads = this.props.selectedUploads.length;
```
ragesoss: For style consistency, this should be done with camelCase, like `selectedUploadsLength`.

Alternatively you could just do the whole pageCount calculation inside the return object, without the intermediate const.

aSquare14: @ragesoss  Done !

PR #2258, opened 2018-11-29
Upgrade jeet to 7.2.0

### LINKED ISSUE #2059: Upgrade Jeet to version 7
We currently use Jeet 6.x to provide grid layouts in our CSS. Upgrading to the latest version will require some changes: the location of the files we need to import is different, and some of the functions that jeet provides are used in various stylesheets and need to be renamed (`cf()`, at least).

Migration guide: https://github.com/mojotech/jeet/wiki/Migrating-from-6-to-7

aSquare14:
fixes #2059

### DIFF
--- a/app/assets/javascripts/components/articles/add_available_articles.jsx
+++ b/app/assets/javascripts/components/articles/add_available_articles.jsx
@@ -33,7 +33,7 @@ const AddAvailableArticles = createReactClass({
   },
 
   chainSubmissions(assignments, promise) {
-    const assignment = assignments.shift();
+    const assignment = assignments.move();
     if (assignment === undefined) { return promise; }
     let extendedPromise;
     if (promise) {
--- a/app/assets/stylesheets/_layout.styl
+++ b/app/assets/stylesheets/_layout.styl
@@ -4,7 +4,7 @@ body > .container
 
 .container
   center(1100px)
-  cf()
+  clearfix()
   padding 0 10px
   position relative
   .container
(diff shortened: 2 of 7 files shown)


aSquare14:
@ragesoss This says that for styl files. Path is same. https://github.com/mojotech/jeet/wiki/Migrating-from-6-to-7#700-removes-superfluous-directories-and-renames-jeetscss-and-jeetstyl-to-indexscss-and-indexstyl

Travis CI is showing that it cannot import jeet

ragesoss:
I don't think that bit of the documentation is correct, as the file looks like it's at `index.styl` instead of `jeet.styl`: https://github.com/mojotech/jeet/tree/master/styl

aSquare14:
How can I reproduce this failing test on my local machine ? Running gulp build or yarn run build passes for me.

ragesoss:
Hmm... I don't know. `gulp build` is the same thing that runs on CI, so I would expect the same error to show up for you locally.

aSquare14:
It doesn't !

ragesoss:
I'll take a look at it soon to see if I can figure it out.

aSquare14:
Thanks !

ragesoss:
I do get those same errors locally.

I had to change the import paths to `node_modules/jeet/styl` to fix them.

ragesoss:
However, while the CSS compiles with that change, the grid layout isn't working:

no grid

What output do you get when you run `yarn`? I can't figure out why this would work locally for you without changing the `@import` paths in `training.styl` and `_shared.styl`, unless you still have the old version of jeet in your `node_modules` directory.

aSquare14:
I'll work on it . Thanks for testing it out !

PR #2271, opened 2018-12-03
Avoid loading scripts and assets from other domains on P&E Dashboard

ragesoss:
Since P&E Dashboard is subject to the Wikimedia Cloud Services terms of use, we can't serve anything from third-party domains.

This switches to serving TinyMCE locally, removes piwik.js from P&E Dashboard, and skips loading the 'Open Sans' and 'Source Sans Pro' fonts from fonts.googleapis.com for P&E Dashboard (falling back instead to arial, which is pretty close).

See https://phabricator.wikimedia.org/T210946

### DIFF
--- a/app/assets/javascripts/tinymce.js
+++ b/app/assets/javascripts/tinymce.js
@@ -1,0 +1,10 @@
+// Import TinyMCE
+import 'tinymce/tinymce';
+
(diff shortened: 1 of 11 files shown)


framawiki:
Thanks for your work on this!

PR #2272, opened 2018-12-04
Migrate to react-transition-group v2

### LINKED ISSUE #2265: Upgrade to react-transition-group v2
See https://github.com/reactjs/react-transition-group/blob/master/Migration.md

We use this is a handful of places. I *think* the usage in `TimelineHandler` doesn't do anything and can just be removed, although I'm not 100% sure.

AmitJoki:
Fixes #2265


ragesoss:
An unrelated spec broke in the meantime. Could you rebase this?

ragesoss:
Nice! One thing I noticed is that I don't think the transition in CourseCreator actually does anything... is that correct?

ragesoss:
Yeah, I'm pretty sure it doesn't do anything.

Issue #2175, opened 2018-11-01
Remove trailing spaces and add a linter that cheks for trailing spaces in this repo


refeed:
From my investigation on this repo, there are about 369 trailing spaces on this repository. It's not a big problem though, but I think it'd be better if this repository had no trailing spaces.
![screenshot from 2018-11-01 19-14-39](https://user-images.githubusercontent.com/20320125/47851701-0afb1100-de0c-11e8-8ea9-327612dcfa7c.png)

PR #2277, opened 2018-12-04
Remove trailing white spaces

ankitkataria:
Addresses #2175 

- Removed trailing white spaces from repo and added appropriate lint configs.


ragesoss:
Cool, thanks! I wonder why that isn't a default rule for either eslit or rubocop.

One little change and this can be merged: the eslint rule should go in the upper section as an intentional rule, rather than in the bottom section with the other rules that we will enable at some point once the violations are fixed.

ankitkataria:
Moved the eslint rule above. :heavy_check_mark:

PR #2278, opened 2018-12-05
I18n training module

AmitJoki:
This PR internationalizes name and description of the training modules.

### DIFF
--- a/app/models/training_module.rb
+++ b/app/models/training_module.rb
@@ -97,17 +97,5 @@ class TrainingModule < ApplicationRecord
     end
   end
 
-  def translated(key)
-    translations.dig(I18n.locale.to_s, key)
-  end
-
-  def t_name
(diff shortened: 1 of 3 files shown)


app/models/training_module.rb
```diff
@@ -97,5 +97,17 @@ def slides
     end
   end
 
+  def translated(key)
```
ragesoss: These are all mainly presentation-level concerns, so I think putting these in the presenter rather than the model would be a little nicer.

app/models/training_module.rb
```diff
@@ -97,5 +97,17 @@ def slides
...
+    translations.dig(I18n.locale.to_s, key)
+  end
+
+  def t_name
```
ragesoss: Spell this out instead of abbreviating it.

PR #2297, opened 2018-12-08
Enable rubocop cops

pjht:
One yarn test failed, but I think this was a problem before I started work. This is the failure: https://pastebin.com/7F1pDQVj

Also, the rspec tests continue to fail after the rebasing. The failures look like this:
F  HTML screenshot: /Users/peterterpstra/Desktop/Wikimedia/WikiEduDashboard/tmp/screen
shots/screenshot_2018-12-07-18-46-06.325.html
  Image screenshot: /Users/peterterpstra/Desktop/Wikimedia/WikiEduDashboard/tmp/screen
shots/screenshot_2018-12-07-18-46-06.325.png

Fixes: #1593


ragesoss:
@pjht that failure can be fixed locally if you run `yarn` again, I think. It's related to a version bump for a node module. I expect it won't fail like that in CI.

pjht:
Yarn tests fixed.

ragesoss:
This includes unrelated changes, with both yarn.lock and the re-addition of lib/training_module.rb

PR #2298, opened 2018-12-08
Isolate displayed date and stored date

### LINKED ISSUE #1234: Datepicker takes wrong date on course creation page
Date selection on the course creation page for course start date and end date through datepicker enters the selected date's previous date in the input field.

Observed behaviour:

courses_datepicker_bug

AmitJoki:
Fixes #1234

### DIFF
--- a/app/assets/javascripts/components/common/date_picker.jsx
+++ b/app/assets/javascripts/components/common/date_picker.jsx
@@ -68,6 +68,7 @@ const DatePicker = createReactClass({
     if (dateObj.isValid()) {
       this.setState({
         value: dateObj.format('YYYY-MM-DD'),
+        displayValue: dateObj.local().format('YYYY-MM-DD'),
         hour: dateObj.hour(),
         minute: dateObj.minute()
       });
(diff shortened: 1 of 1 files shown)


ragesoss:
This looks likes it fixes the UI bugginess in all the situations I saw, and I can't see that it was doing anything useful...

However, it doesn't fix the ultimate bug of server time vs. local time. In particular, any timezone east of UTC ends up with dates shifted by 1 day after the user saves.

I tested this by:
* Setting timezone to a European one
* Setting ClassroomProgramCourse as the default course type
* Creating a course and setting start date December 1, end date December 31
* Now on the wizard page (which shows up after the new course is created on the server and sent back to the client) the start/end dates are November 30 and December 30.

ragesoss:
I'm not hinting at anything in particular, just thinking through what the right behavior ought to be...

I think the only way to make this work the way we really want is to put timezone all the way through the process. So, we want to have the course start/end dates for the ones that don't specify exact time to be start of the day *in local time* for the start date, and to end at end of the day *in local time* on the last day. I think that means both showing local timezones on the frontend, and setting start/end dates in local time on the backend... and then making sure to pass the exact start/end times back to the frontend so that if someone looks at the same course but from a different timezone, they see that it starts at 1am and ends at 12:59am, or whatever.

PR #2304, opened 2018-12-11
Enable rubocop cops

### LINKED ISSUE #1593: Enable more rubocop rules and fix the violations

pjht:
Submission for the GCI task https://codein.withgoogle.com/dashboard/task-instances/4834317768327168/

Fixes #1593

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -116,8 +116,6 @@ Rails/InverseOf:
   Enabled: false
 Rails/UnknownEnv:
   Enabled: false
-Performance/RegexpMatch:
-  Enabled: false
 Rails/Delegate:
   Enabled: false
 Rails/HasAndBelongsToMany:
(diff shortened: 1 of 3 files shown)


app/models/campaign.rb
```diff
@@ -115,9 +115,7 @@ def validate_dates
   # Intercept Rails typecasting and add error if given value cannot be parsed into a date.
   def validate_date_attribute(date_type)
     value = send("#{date_type}_before_type_cast")
-    # rubocop:disable Rails/TimeZone
```
ragesoss: The removal of this is causing the build to break. Locally, you need to run `rspec --rails` in order to include the Rails-specific cops (in which case, this comment won't itself be a violation).

PR #2315, opened 2018-12-13
Make Jquery a webpack dependency

pjht:
Removes the jquery-uls task
Adds a webpack dependency for jquery
Changes the head HAML template to use the webpack jquery
diff_viewer.jsx now uses main jquery

Fixes #2132


ragesoss:
This looks like progress, but it looks like jquery isn't being made available for the survey and survey_admin JS files, and it also breaks some of the JS tests.

Maybe the answer is to get rid of the separate jquery file, and instead include campaigns.js into main.js (as we already do with course.js), and then explicitly import jquery in each file that uses it and put it into the window global for surveys explicitly.

ragesoss:
Still some build errors, possibly just eslint violations, that are causing the gulp build to fail on travis-ci.

ragesoss:
Feel free to reopen if you take this up again.

PR #2329, opened 2018-12-18
adds wrapping in arrange timeline

shivanshbindal9:
pr2

### DIFF
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -91,14 +91,16 @@
   cursor move
 
   > h4
-    margin-bottom 0
-    margin-right auto
-    white-space nowrap
-    text-overflow ellipsis
+    line-height inherit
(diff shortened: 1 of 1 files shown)


ragesoss:
Thanks! This looks like a bit of style regression for single-line titles: the whitespace is no longer symmetric above vs. below the title, and the text is smaller. Ideally the fix should allow wrapping but otherwise leave the style as-is.

ragesoss:
Wow, that's multiple kinds of bad layout!

Fixing the red border so it don't go under the block as well would be nice, although that can be a separate issue if necessary.

ragesoss:
@shivanshbindal9 are you planning to continue working this issue? As I said above, I think with a few further tweaks, we can fix all the main problems.

shivanshbindal9:
Yes, I will try to fix it!

ragesoss:
I think this is largely fixed now.

PR #2332, opened 2018-12-21
[WIP] Fixes #2312 (Help Required)

### LINKED ISSUE #2312: Move jquery-uls to the webpack config

This will involve:

devanshbatra04:
@ragesoss 

## Changes made 
completed the steps as mentioned in #2312 #
* creating a jquery.js entry file in app/assets/javascripts
* updating package.json to treat jquery as normal dependency instead of the current config that puts it in * @bower_components/jquery

## Problem:
jQuery tests for React keep failing, I can't come up with any explanation.


devanshbatra04:
Here are some failing specs. Probably something to do in the Webpack file.
screenshot from 2018-12-21 15-20-59

devanshbatra04:
Solved, the issue was I had named jquery.js as the entry file name, it collided with jquery module in node_modules. The test script instead imported this file. Shifting the order of modules in node_modules solved the issue.

devanshbatra04:
screenshot from 2018-12-21 19-39-36
I have narrowed down, the ruby spec errors. This seems to be causing all of them. Probably a spec used something from the old jquery.js which appeared at the same place due to a gulp task? Any help will be appreciated

devanshbatra04:
Its working well in webpack, and javascript specs. Not sure why ruby specs are failing. There was a similar error with Javascript specs. 
So gulp copied the bower jquery as jquery.js into the same location as this entry file, that required adding the entry file directory as where to look for javascript modules, this conflicted the require statements (in JS specs) with the npm package in node_modules. Changing the order in package.json did the trick for JS specs. 
Probably something similar happening with Ruby?

ragesoss:
The js specs run individual javascript files in relative isolation, and they operate on the source files rather than the webpack output.

The ruby `feature` specs that are failing are browser tests, which load the site (including the javascripts present on any given page being tested) and which will fail if there are any errors in the javascript console. I expect that you'll see some of the same JS errors if you load the site locally.

PR #2351, opened 2019-01-04
Update setup.md and troubleshooting.md

Ace-Bansal:
screenshot from 2019-01-05 00-42-13
Users have been facing this problem while setting up the dev environment. I've updated the setup docs to include the command `yarn --ignore-engines` which solves this problem.
I've also mentioned using node v10 or lower to avoid any problems as a pre-requisite to running the project.


Ace-Bansal:
@ragesoss could you please review and point out any mistakes that I may have made. Thanks!

docs/setup.md
```diff
@@ -38,6 +39,8 @@ You can also contact us on slack for any further queries.
 - If you face the error that `Sorry! Your operating is not supported by this script`
   - You can try running the system dependent scripts from setup directory, according to your system
   - You can try manual installation
+- After running 'gulp', if you face the error `Failed to load external module @babel/register`
```
ragesoss: Instead of putting `--ignore-engines` in the documentation here, the setup script itself should be changed to use that flag.

Also, the manual setup instructions below should be updated to include that flag.

package.json
```diff
@@ -34,7 +34,7 @@
     "eslint-plugin-import": "^2.14.0",
     "eslint-plugin-jsx-a11y": "^6.1.2",
     "eslint-plugin-react": "^7.11.1",
-    "gulp": "^3.8.7",
+    "gulp": "^3.9.1",
```
ragesoss: The changes to package.json and yarn.lock should not be included in this PR.

package-lock.json
```diff
@@ -0,0 +1,2011 @@
+{
```
ragesoss: This file should not be added. We use yarn.lock instead.

psinghal20:
@Ace-Bansal yes, that is correct. You can iterate that over all the shell scripts.

Ace-Bansal:
Changes made in latest commit:
1. Include the --ignore-engines flag in all setup scripts.
2. Include the tag in the manual setup instructions in setup.md
3. Remove package-lock.json
4. Undo the changes done in package.json & yarn.lock by previous commit.

psinghal20:
Looks good to me! Sage can have a look tomorrow and merge it.

PR #2352, opened 2019-01-04
Fixes #2350 added strip course params in set_slug

### LINKED ISSUE #2350: Courses can be created with a space at the end
[This](https://outreachdashboard.wmflabs.org/courses/Wikimedia_ZA/Afrikaans_Writing_Competition) course was initially created with an additional space at the end of the title, which meant the URL ended with an additional _. I've since renamed it, but presumably additional whitespace should be trimmed before course creation.

garry329:
I have the striped the various the params which make up the slug on course creation.This is my first contribution to this project.Please point any mistakes or points I didn't notice.

### DIFF
--- a/lib/course_creation_manager.rb
+++ b/lib/course_creation_manager.rb
@@ -56,9 +56,9 @@ class CourseCreationManager
   end
 
   def set_slug
-    slug =  @course_params[:school].presence || ''
-    slug += "/#{@course_params[:title]}" if @course_params[:title].present?
-    slug += "_(#{@course_params[:term]})" if @course_params[:term].present?
+    slug =  @course_params[:school].strip.presence || ''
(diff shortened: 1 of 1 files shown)


ragesoss:
@garry329 were you able to replicate the bug? This is something that already has some client-side sanitization going on, and I'm not quite sure where the trailing underscore slips through that.

garry329:
Yes @ragesoss, I was able to replicate the bug by adding a single space after course title or institute name in the course creation form.

ragesoss:
Thanks! This would benefit from a test, ideally a feature spec that fills out the course creation form with trailing spaces and verifying that the created course has a proper slug without the extra spaces. That way we can make sure it doesn't get re-broken later on.

ragesoss:
I'll go ahead and merge this now.

PR #2355, opened 2019-01-07
Alignment done on Editing Details Component

hardy1334:
## Before the Pull Request ##
* Test out your changes manually to confirm that it all works and search for bugs.
* Run the specs for any Ruby files you changed, along with the feature specs that are related to your changes. Optionally, you can run the entire test suite locally (`rake spec`).

* If the PR is not complete but you want feedback on it or you just want to trigger a build, include `[WIP]` in the title to indicate work-in-progress, and add comments about anything you're stuck on. Edit the title when it's ready.

## After the PR ##
    * Fix any failures.


ragesoss:
There a bunch of unrelated changes (formatting, renaming of the example application.yml). This PR should only include the changes for the issue at hand.

It looks like the fix is to shorten the interface message that was wrapping? I think we should keep that interface message as-is, if possible, and find an HTML/CSS solution to prevent that unecessary wrapping.

hardy1334:
@ragesoss  I'll look into that

hardy1334:
@ragesoss it has been updated,please can you have a look at it 
screenshot from 2019-01-08 21-10-37

ragesoss:
Hmm... having a look at the behavior a little more closely, this alignment issue is tied to window size. I don't like the idea of changing the margin throughout the whole Details interface, and it also still doesn't solve the alignment problem for narrower windows.
10px left margin

Instead, I think we need that interface message to be able to extend out beyond the date segment to the time segment as well. The label is inside the same element as the first input, and could probably be moved outside of that instead. The place to do that change is the DatePicker component.

hardy1334:
@ragesoss  probably now I'll try to do it by doing some changes in Datepicker component

hardy1334:
screenshot from 2019-01-09 23-00-21

hardy1334:
screenshot from 2019-01-09 23-03-05

app/assets/stylesheets/modules/_course.styl
```diff
@@ -199,6 +199,7 @@ rating(color)
 
     .course-details
       width 90%
+      max-width 920px
```
ragesoss: This section is part of the media query, and only applies below 920px screen width. The max-width should be in the same class as the changed width above.

app/assets/stylesheets/modules/_course.styl
```diff
@@ -159,7 +159,7 @@ rating(color)
     z-index 1
 
   .course-details
-    width 50%
+    width 80%
```
ragesoss: Making this 90%, along with the max-width, is a little better I think.

hardy1334: Sorry for the discrepency,I'll change that

app/assets/stylesheets/modules/_course.styl
hardy1334: I'll change it

PR #2367, opened 2019-01-09
Redirect when users visit /update_locale/:locale to change languages

ragesoss:
This will save a bit of confusion when we direct users to update their locale via https://dashboard.wikiedu.org/update_locale/en

### DIFF
--- a/app/controllers/users/locale_controller.rb
+++ b/app/controllers/users/locale_controller.rb
@@ -11,6 +11,7 @@ class Users::LocaleController < ApplicationController
     if request.method == 'POST'
       render json: { success: true }
     else
+      flash[:notice] = 'Locale preference updated!'
       redirect_to '/'
     end
   end
--- a/app/controllers/users/locale_controller.rb
+++ b/app/controllers/users/locale_controller.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+class Users::LocaleController < ApplicationController
+  before_action :require_signed_in
+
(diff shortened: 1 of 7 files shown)


ragesoss:
@bwreid is there anything you'd do differently here? I want to learn what I can from you while your eyes are still fresh on this codebase!

bwreid:
Looks pretty solid to me! Should we do anything for users that end up [at the wrong place](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/7c62faab08998bd193bf8e1eb590806d7d98be5e/app/controllers/users_controller.rb#L30-L33) as well?

As a side note, I'm assuming users get here just based off of something a staff member tells them?

ragesoss:
@bwreid I don't think the failure condition really needs any special handling; it should be extremely rare, is likely to be confusing no matter what.

Yeah, someone asks about it maybe once a term or less, and just point them to https://dashboard.wikiedu.org/update_locale/en

ragesoss:
I guess one possible approach to improve that would be to insert the language picker even on dashboard.wikiedu.org _if the user's locale is not English_.

ragesoss:
I had second thoughts about it. I think it might be weird to have that language picker show up by default, then you change your language to English (because it's there, so you are invited to play around with it), and after that it's gone. I think if we went that route, it would be more like:

* show language picker if user locale is not English *or* if user has an explicit locale preference set on their user record.

PR #2368, opened 2019-01-09
Update referral question on /onboarding/supplementary page

bwreid:
- Update question from open-ended to radio buttons
- Conditionally include text area for users to add more details
- Include components level tests, controllers tests, and integration tests
referral


ragesoss:
Looks good!

PR #2371, opened 2019-01-10
Fix course description bug where field resets to initial state

bwreid:
Ended up being a real simple change.  :)

This test also correctly fails with the previous code but now passes.

### DIFF
--- a/app/assets/javascripts/utils/course_utils.js
+++ b/app/assets/javascripts/utils/course_utils.js
@@ -178,14 +178,23 @@ const CourseUtils = class {
 
   newCourseStats(oldCourse, newCourse) {
     return {
-      createdCount: oldCourse.created_count !== newCourse.created_count,
-      editedCount: oldCourse.edited_count !== newCourse.edited_count,
-      editCount: oldCourse.edit_count !== newCourse.edit_count,
-      studentCount: oldCourse.student_count !== newCourse.student_count,
-      wordCount: oldCourse.character_sum_human !== newCourse.character_sum_human,
-      viewCount: oldCourse.view_count !== newCourse.view_count,
-      uploadCount: oldCourse.upload_count !== newCourse.upload_count
+      created_count: oldCourse.created_count !== newCourse.created_count,
+      edited_count: oldCourse.edited_count !== newCourse.edited_count,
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/utils/course_utils.js
```diff
@@ -187,6 +188,16 @@ const CourseUtils = class {
       uploadCount: oldCourse.upload_count !== newCourse.upload_count
     };
   }
+
+  // Given a course and camelized stats from the above `newCourseStats`
+  // function, return only the key-value pairs of what needs to be updated
+  // in the course.
+  courseStatsToUpdate(course, newStats) {
+    const decamelizedStats = decamelizeKeys(newStats);
```
ragesoss: hmm... I'm okay with this, but I don't really like the indirection here.

The camelCase values happen to correspond to the snake_case ones that newStats uses, but that's because it mirrors the coupling that is present in CourseStats, where the camelCase values are used for class names. The snake_case values are still what the course object is keeping as properties, and what is being rendered in CourseStats.

Ultimately, I think the right way to fix all these snake_case properties is to change it at the interface between Ruby and JavaScript... that is, at the point where we generate the JSON.

bwreid: I'd agree that just switching the casing here is not great. I did it this way to try and minimize the impact elsewhere... but I'm totally down to make more changes.

A small first step could just be to change the `course_stats.jsx` to use snake_case versions. I think that would get rid of any need to change any casing in this function and would modify the `newCourseStats` function to just use snake_case. If that seems acceptable, I can make those changes to this PR.

Agreed, the right way would be to pick a casing for the frontend and make sure everything fits that convention. I think there's an opportunity to solve this and #2015 at the same time. We could build a request module that uses `fetch` and camelizes any response from the API.

yarn.lock
```diff
@@ -6853,6 +6853,11 @@ https-proxy-agent@^2.2.1:
     agent-base "^4.1.0"
     debug "^3.1.0"
 
+humps@^2.0.1:
```
ragesoss: This got left behind.

bwreid: Yikes, sorry bout that. Updated now.

PR #2374, opened 2019-01-11
Separate course_creator component into child components

Cressence:
## Before the Pull Request ##
* Test out your changes manually to confirm that it all works and search for bugs.
* Run the specs for any Ruby files you changed, along with the feature specs that are related to your changes. Optionally, you can run the entire test suite locally (`rake spec`).

* If the PR is not complete but you want feedback on it or you just want to trigger a build, include `[WIP]` in the title to indicate work-in-progress, and add comments about anything you're stuck on. Edit the title when it's ready.

## After the PR ##
    * Fix any failures.

### DIFF
--- a/app/assets/javascripts/components/course_creator/clone_course.jsx
+++ b/app/assets/javascripts/components/course_creator/clone_course.jsx
@@ -1,0 +1,12 @@
+import React from 'react';
+
+const CloneCourse = ({ cloneClasss, chooseNewCourseAction, showCloneChooserAction, createLabel, cloneLabel }) => {
+  return (
+    <div className={cloneClasss}>
+      <button className="button dark" onClick={chooseNewCourseAction}>{createLabel}</button>
+      <button className="button dark" onClick={showCloneChooserAction}>{cloneLabel}</button>
+    </div>
+  );
+};
+
+export default CloneCourse;
--- a/app/assets/javascripts/components/course_creator/course_creator.jsx
+++ b/app/assets/javascripts/components/course_creator/course_creator.jsx
@@ -16,7 +16,7 @@ import Modal from '../common/modal.jsx';
 import CourseUtils from '../../utils/course_utils.js';
 import CourseDateUtils from '../../utils/course_date_utils.js';
 import CourseType from './course_type.jsx';
-import CloneCourse from './clone_course.jsx';
+import NewOrClone from './new_or_clone.jsx';
 import ReuseExistingCourse from './reuse_existing_course.jsx';
 import CourseForm from './course_form.jsx';
 
@@ -273,7 +273,7 @@ const CourseCreator = createReactClass({
           <div className="wizard__panel active" style={formStyle}>
             <h3>{CourseUtils.i18n('creator.create_new', this.state.course_string_prefix)}</h3>
             <p>{instructions}</p>
-            <CloneCourse
+            <NewOrClone
               cloneClasss={cloneOptions}
               chooseNewCourseAction={this.chooseNewCourse}
               showCloneChooserAction={this.showCloneChooser}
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/components/course_creator/clone_course.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const CloneCourse = ({ cloneClasss, chooseNewCourseAction, showCloneChooserAction, createLabel, cloneLabel }) => {
```
ragesoss: Maybe call this NewOrClone, or something else that will make it easy for us to remember what it's for?

Cressence: It is for the screen which is displayed asking the user to either create new course or clone existing course when the create course button is clicked.

ragesoss: Right. So I think 'CloneCourse' may be confusing, as I might assume it's a component focused on choosing a course to clone, or filling out the details to finalize cloning a course, rather than choosing whether to clone a course or start from scratch.

ragesoss:
Looks like the build is failing because of an unused variable, `chooseNewCourseAction`.

Cressence:
> Looks like the build is failing because of an unused variable, `chooseNewCourseAction`.

The variable is actually used.
selection_009

ragesoss:
@psinghal20 @ihsavru if either of you have a chance to review this, that would be fabulous.

ihsavru:
@Cressence this.courseSelect is undefined after you moved out the logic for `ReuseExistingCourse`
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/84053f37f4e33614d063642a70ecf21faf424532/app/assets/javascripts/components/course_creator/course_creator.jsx#L204

ihsavru:
@Cressence this.courseSelect is undefined, therefore, the tests are failing. Did you try cloning in your local server? I'm trying to do that but doesn't work for me. You might want to look into the ref of your `ReuseExistingCourse` component.

app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -448,101 +272,41 @@ const CourseCreator = createReactClass({
...
+              courseSelect={reuseCourseSelect}
+              options={options}
+              useThisClassAction={this.useThisClass}
+              cloneThisLabel={CourseUtils.i18n('creator.clone_this', this.state.course_string_prefix)}
+              cancelCloneAction={this.cancelClone}
+              cancelLabel={CourseUtils.i18n('cancel', this.state.course_string_prefix)}
+            />
+            <CourseForm
+              courseFormClass={courseFormClass}
+              course_utils={CourseUtils}
```
ragesoss: `CourseUtils` is just an import, so there's no reason to pass it down as a prop rather than import it in the child component.

app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -448,101 +272,41 @@ const CourseCreator = createReactClass({
...
+              string_prefix={this.state.course_string_prefix}
```
ragesoss: For consistent style, you can make this `stringPrefix` instead of `string_prefix`.

ragesoss:
@Cressence This a solid start at refactoring CourseCreator.

I think there's a little more that could be done to push functions down to the components that use them, in cases where only one component uses that function. Since many of the functions are built out of the basic functions like `updateCourse`, you could just pass down the basic function, and move the derivative function into the child component. `updateCourseDates` and `updateCourseType` are two that jump out as appropriate for that type of refactoring. The smaller we can make `CourseCreator` itself, the better.

PR #2377, opened 2019-01-11
Bug fix for keyboard navigation on training module pages

bwreid:
I am _still_ trying to get some of these JS tests to be useful, but in the meantime we should just put this up.  :)

### DIFF
--- a/spec/features/training_tool_spec.rb
+++ b/spec/features/training_tool_spec.rb
@@ -101,6 +101,15 @@ describe 'Training', type: :feature, js: true do
       expect(tmu.reload.last_slide_completed).to eq(module_2.slides.second.slug)
     end
 
+    it 'allows for navigation with the left and right buttons' do
+      click_link 'Start'
+      expect(page).to have_content('Page 1 of')
+      find('html').native.send_keys :right
+      expect(page).to have_content('Page 2 of')
+      find('html').native.send_keys :left
+      expect(page).to have_content('Page 1 of')
+    end
+
     it 'sets the module completed on viewing the last slide' do
       login_as(user, scope: :user)
       sleep 1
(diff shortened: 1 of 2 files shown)


spec/features/training_tool_spec.rb
```diff
@@ -101,6 +101,17 @@
       expect(tmu.reload.last_slide_completed).to eq(module_2.slides.second.slug)
     end
 
+    it 'allows for navigation with the left and right buttons' do
+      click_link 'Start'
+      sleep 1.5
+      tmu = TrainingModulesUsers.find_by(user_id: user.id, training_module_id: module_2.id)
```
ragesoss: What's this for?

spec/features/training_tool_spec.rb
```diff
@@ -101,6 +101,17 @@
...
+      expect(find('body').text).to have_content('Page 1 of')
```
ragesoss: Is there any benefit to `find('body').text` over `page` here? I usually use `expect(page).to have_content`, just because it is closer to English.

spec/features/training_tool_spec.rb
```diff
@@ -101,6 +101,17 @@
       expect(tmu.reload.last_slide_completed).to eq(module_2.slides.second.slug)
     end
 
+    it 'allows for navigation with the left and right buttons' do
+      click_link 'Start'
+      sleep 1.5
```
ragesoss: I think with `page`, at least, the sleep won't be needed.

spec/features/training_tool_spec.rb
ragesoss: Oh, I see... copied from the other examples.

I think the sleep is only necessary because we want to find a record that got created (and later, updated) when the user clicked start. But for just testing navigation, we can skip anything to do with the database altogether, and also avoid the sleep delay.

PR #2380, opened 2019-01-13
add react-select in yes_no_selector

shivanshbindal9:
private
timeline_enabled
This adds react-select in both private and time_enabled.


app/assets/javascripts/components/overview/yes_no_selector.jsx
```diff
@@ -11,10 +13,15 @@ const YesNoSelector = createReactClass({
     editable: PropTypes.bool,
     updateCourse: PropTypes.func.isRequired
   },
-
-  _handleChange(e) {
+  componentWillMount() {
```
ragesoss: This can probably be done in `componentDidMount` rather than `componentWillMount`, which is not the recommend way to do it any longer: https://reactjs.org/docs/react-component.html#mounting

app/assets/javascripts/components/overview/yes_no_selector.jsx
```diff
@@ -11,10 +13,15 @@ const YesNoSelector = createReactClass({
...
+    this.setState({
+      selectedOption: { value: this.props.course[this.props.courseProperty] ? 'yes' : 'no', label: this.props.course[this.props.courseProperty] ? 'yes' : 'no' },
```
ragesoss: This would be a good time to add i18n support to the yes/no labels.

ragesoss: You can probably also refactor this to only do `this.props.course[this.props.courseProperty] ? 'yes' : 'no'` once.

app/assets/javascripts/components/overview/yes_no_selector.jsx
shivanshbindal9: Can we do it by getInitialState function ? https://reactjs.org/docs/react-without-es6.html

app/assets/javascripts/components/overview/yes_no_selector.jsx
shivanshbindal9: When I am using I18n to translate 'yes' to various languages it is showing not found. So, how should I add translation of 'yes' in various languages. @ragesoss

ragesoss:
@shivanshbindal9 can you rebase this? Hopefully the build will pass after this is rebased; it was failing because of an unrelated issue.

app/assets/javascripts/components/overview/yes_no_selector.jsx
ragesoss: Yes, getInitialState should work fine.

app/assets/javascripts/components/overview/yes_no_selector.jsx
ragesoss: You'd need to add them to the default locale file, `config/locales/en.yml`. They should probably go in the `application` section, with keys of `yes` and `no`.

shivanshbindal9:
Can you please review the changes @ragesoss .

ragesoss:
This looks mostly solid. One thing I noticed while testing it out, which I now realize also applies to other instances of using React-Select, is that it behaves differently from a native select in terms of which option is selected by default when you open the dropdown.

For example, clicking this input with the native `select` automatically shows the the current value as the selected one, and using `tab` to exit that input results in no change to the selection.

no

With the version from this PR, the first option is always the active one, even if a different value is the current selection, and using tab to exit the input then results in selecting a different option than what it had previously.

no yes

I think this has a high chance of causing problems with accidental changes. Is there a way to make it behave like the native version here?

shivanshbindal9:
recomended_changes
I have made both the changes. Now, the first option will not be active and also on using tab to exit the value does not changes.

ragesoss:
Thanks! I'll likely review this sometime next week, unless someone else can do so earlier.

Issue #2089, opened 2018-10-11
Course with all Saturdays as the meetings dates does not generate a timeline


ragesoss:
We ran into a bug today for a course where all the meeting dates are on Saturdays: going through the wizard and clicking 'Generate Timeline' does not generate a timeline, it just returns an empty one.

This seems like probably a bug on the Rails end of the wizard.

ragesoss:
Fixed!

PR #2381, opened 2019-01-15
Bug fix to allow classes to be scheduled on Saturdays

bwreid:
Fixes issue #2089.

`.end_of_week(:<day>)` is exclusive. For example:

```ruby
> '2019-01-15'.to_date.end_of_week(:saturday)
=> Fri, 18 Jan 2019
```

### DIFF
--- a/lib/course_meetings_manager.rb
+++ b/lib/course_meetings_manager.rb
@@ -97,7 +97,8 @@ class CourseMeetingsManager
     meetings = []
     @timeline_week_count.times do |wk|
       week_start = @beginning_of_first_week + wk.weeks
-      week_end = week_start.end_of_week(:saturday)
+      # This excludes Sunday, putting the end of the week at Saturday.
+      week_end = week_start.end_of_week(:sunday)
       week_mtgs = []
       @meeting_dates.each do |meeting|
         next if (meeting < @course.timeline_start) || (@course.timeline_end < meeting)
--- a/spec/lib/course_meetings_manager_spec.rb
+++ b/spec/lib/course_meetings_manager_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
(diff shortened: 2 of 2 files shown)


ragesoss:
Nice. Do you think it's worth it to keep the new feature spec, and add that 20 seconds or so to the full build? (Either way is fine with me, I just want to consider it.)

lib/course_meetings_manager.rb
```diff
@@ -97,7 +97,7 @@ def calculate_week_meeting_dates
     meetings = []
     @timeline_week_count.times do |wk|
       week_start = @beginning_of_first_week + wk.weeks
-      week_end = week_start.end_of_week(:saturday)
+      week_end = week_start.end_of_week(:sunday)
```
ragesoss: Probably ought to have a comment explaining that this code is supposed to return the last day, Saturday, of the week.

bwreid: Can do. I actually think just `.end_of_week` will work as well. Perhaps that's better syntax anyhow?

bwreid:
I'd be happy to remove the feature spec if you can help me figure out what other tests to write in the `course_meetings_manager_spec.rb`. The one test I added is only to the `#meeting_dates_of` method. Are there other methods it would be useful to evaluate?

bwreid:
It looks like a different issue is hanging this up, so I'm going to look after lunch.

lib/course_meetings_manager.rb
ragesoss: Yeah, `end_of_week` default is probably good, but maybe merits a comment anyway, since the intention remains to return a saturday but it's not obvious that's what will happen.

bwreid: `end_of_week` _does_ work but if you call it on a Sunday it just picks the same day. Added a comment and left in the `:sunday` keyword.

ragesoss:
This was merged separately.

PR #2395, opened 2019-01-16
Onboarding Alert displayed on course page for admins

bwreid:
Pushing this up at the moment to ask a few questions. This is not ready to merge.

As is, this pull request:
- Shows the specified alert
- Allows for filtering through the URL on the `/alerts_list` page
- Allows for a `json` response on the `/alerts_list` page
- Still needs more testing and organization of code

image

#### Question / comments:
* I had some trouble with the JS tests (surprise) in part because of some library issues. The [isomorphic-fetch](https://github.com/bitinn/node-fetch/issues/481) library seems to only though issues while testing.
* [This is gross.](https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/onboarding-alert-course-page-feature?expand=1#diff-8e5b00cd1fb9d3693bdddddcd92ab7c0R69) Either we'll need some tricky string manipulation and/or will want to change the input going to the messages. I would suggest:
  1. Update the format of information being sent so that it is perhaps easier to parse.
  2. Move this code into `courseUtils` and expand it out so it's easier to read.

  I think the ideal solution would be to start to store these messages as JSON and then present them in whatever way we need.


ragesoss:
@bwreid For that gross bit, it occurs to me that we don't need to be storing that data as strings in the first place. I think I added the original implementation of the onboarding alerts before we added the `details` attribute to the Alert model; it's basically a hash that we can put whatever data we want into, so we could easily switch to a structured format for handling the input when it originally comes in.

That leaves the problem of dealing with already-created OnboardingAlerts, but an uglier fallback based on alert#message (just render the whole message, without worrying about the structure) when alert#details is not there would be fine. (Keeping that gross implementation as a fallback, and moving it to courseUtils, and maybe expanding it to be understandable, would all be okay as well; as long as it's tested and clearly labeled, it probably won't give us any future trouble.)

For the isomorphic-fetch issue... yuck. Nothing comes immediately to mind.

app/assets/javascripts/components/overview/overview_handler.jsx
```diff
@@ -44,6 +46,7 @@ const Overview = createReactClass({
 
   componentDidMount() {
     if (this.props.current_user.admin) {
+      this.props.fetchOnboardingAlert(this.props.current_user);
```
ragesoss: Current user is the admin, rather than the instructor.

PR #2400, opened 2019-01-18
Onboarding alert refactor

bwreid:
This relates to PR #2395 and is additional refactor on top of it. We can either merge this one first, and then the base PR, or just worry about this one separately.

This refactor moves the course alerts to their own reducer and action file. It also tests both the actions and the reducer, in part mocking out API requests.


ragesoss:
Looks reasonable to me. I suggest doing it in whatever order will be easier.

PR #2401, opened 2019-01-19
Store onboarding alert information in serialized hash

bwreid:
This will store OnboardingAlert information as a hash in the `details` column. This will make it easier to present this information back on the frontend.


bwreid:
Still compatible @ragesoss. At the moment this just adds additional information to that one field. The final step would then be to make changes to the frontend so that we use the text as a fallback and use the `details` json as the information by default.

PR #2409, opened 2019-01-22
[WIP] Allow for backup account when creating requested accounts

bwreid:
Adds a new environment variable that specifies an ID for an account that will be used as a backup in the event a requested account creation fails. If the environment variable is not there, it will just default to the previous implementation.

After you've had a chance to take a look at the code, let's just chat on Slack about this. I'd like to just run by you what I'll do to be manually testing this.

### DIFF
--- a/app/services/create_requested_account.rb
+++ b/app/services/create_requested_account.rb
@@ -6,7 +6,7 @@ require_dependency "#{Rails.root}/lib/importers/user_importer"
 # Processes a RequestedAccount by creating a new mediawiki account, and
 # creating the User record upon success.
 class CreateRequestedAccount
-  attr_reader :result, :user
+  attr_reader :result, :user, :creator
 
   def initialize(requested_account, creator)
     @creator = creator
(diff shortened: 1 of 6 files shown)


app/services/create_requested_account.rb
```diff
@@ -36,38 +37,55 @@ def process_request
...
   end
 
+  def retry_request_with_backup_account
+    backup_account_id = ENV['account_creation_backup_creator_id']
+    backup_account = User.find_by(id: backup_account_id)
+    return log_unexpected_response if !backup_account.is_a?(User) || backup_account == @creator
+
+    @creator = backup_account
+    creation_reason = "Created #{@username} at the request of #{@creator}."
+    process_request(backup_account, creation_reason)
```
ragesoss: This will loop if the retry is also failing.

bwreid: I think the line above setting `@creator = backup_account` should solve this? On a second run through, the backup_account and the @creator should be the same.

ragesoss: What happens if trying with the backup account fails (for example, because the account's credentials expired or it got blocked)? I don't see how to avoids an infinite loop in that circumstance.

ragesoss:
merged separately.

PR #2410, opened 2019-01-23
solve on_tab_press issue in course_type_selector and home_wiki_projec…

shivanshbindal9:
This fixes the issue of auto selection of first option in course_type_selector and home_wiki_project_selector.


ragesoss:
Works smoothly, thanks!

PR #2424, opened 2019-01-26
add react-select in block_type_select without css

shivanshbindal9:
block_type
This adds react-select in block_type_select but the css in not perfect. @ragesoss can you please give some suggestions how should it look as increasing the size of select component disrupts in the css of entire block.

### DIFF
--- a/app/assets/javascripts/components/high_order/input_hoc.jsx
+++ b/app/assets/javascripts/components/high_order/input_hoc.jsx
@@ -55,9 +55,9 @@ const InputHOC = (Component) => {
 
     shouldComponentUpdate(nextProps, nextState) {
       if (this.state.value === nextState.value
-            && this.state.id === nextState.id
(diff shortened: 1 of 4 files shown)


ragesoss:
Hmm... It looks like this select is considerably narrower than the current one:

block editable

Making it wider to match the current layout would fix most of the problem (if not all of it).

shivanshbindal9:
block_type
I have improved the styling. Can you please review it !

ragesoss:
It looks good now from the screenshot. Could you remove the whitespace changes (those could be done in a separate PR)? That'll make it clearer which changes are related to this issue, and keep the git history more useful.

shivanshbindal9:
I have made a new PR for adding react-select to block_type_select #2429 @ragesoss . The whitespace changes were added unintentionally. Sorry, for any inconvenience.

PR #2429, opened 2019-01-29
Block type select

shivanshbindal9:
block_type
This adds react-select in block_type_select.

### DIFF
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -140,20 +140,8 @@
     margin-right 80px
   select
     margin-right 0
-
-
-.react-select_dropdown
-  color $text_dk
-  font-size 14px
-  font-weight 300
-  position relative
-  -webkit-font-smoothing auto
-  top 2px
-  min-width 125px
-  &.editable
-    margin-right 80px
-  select
-    margin-right 0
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/modules/_timeline.styl
```diff
@@ -131,11 +131,12 @@
 
 .block__block-type
   color $text_dk
-  font-size 14px
+  font-size 14  px
```
ragesoss: Is that a typo?

shivanshbindal9: Yeah, sorry I didn't notice it. 
I have fixed it.

ragesoss:
This looks pretty  good — although Assignment still gets truncated.

There's one small regression: if you change the type, but then click 'Cancel', the input does not reset to the original value (while it does reset correctly in the current implementation).

shivanshbindal9:
@ragesoss It is working fine on my system i.e. you click 'Cancel' the input does not reset.

ragesoss:
If you click cancel, it 'should' reset.

shivanshbindal9:
Sorry, my bad!! Can you please review it. @ragesoss

shivanshbindal9:
Can you please review it @ragesoss !

PR #2433, opened 2019-01-30
Upgrade rails version from 5.2.2 to 6.0.0.beta1

### LINKED ISSUE #2432: Upgrade Rails to 6.0.0.beta1

faithngetich:
What does this PR do.
------------------------------------------
This PR updgrades rails version from 5.2.2 to 6.0.0.beta1 version

- Fixes #2432


ragesoss:
Looks like this is coming along! You might need to rebase it on `master` to avoid the gemfile.lock confict, so that the tests will run.

ragesoss:
@faithngetich it looks like the things failing now are mostly, if not all, because of the activerecord-import version bump. See https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2440

I recommend taking a step back and doing a separate PR to fix just the activerecord-import version change, and then rebasing the rails upgrade on that.

ragesoss:
@faithngetich another helpful way to approach this will be to create separate PRs (or one PR with multiple commits) for each of the changes that is required for Rails 6.0 but is compatible with our current Rails version, so that we can merge those changes immediately, even before the final release of Rails 6.

faithngetich:
@ragesoss I think, creating a separate PR to fix the existing issue is better.  To avoid working on a really big PR.

faithngetich:
@ragesoss I believe we are having the `activerecord-import`  error because before I changed the rails version, I did run `bundle update` which upgraded the version of all our dependencies thus `activerecord-import` was upgraded.  

I will go ahead and create a new branch and open another a PR that's cleaner i.e I'll only run `bundle update rails`  to only update the rails version. I think this is a more cleaner way to go about fixing the mistake I made. Then the `activerecord-import` upgrade error can be fixed separately. 

 Does this sound reasonable to you?

ragesoss:
Yeah, sounds good. Looks like it's passing, too!

Issue #2435, opened 2019-01-30
Add email to instructors for unsubmitted courses alerts


ragesoss:
We currently have an UnsubmittedCoursesAlert that is supposed to let Wiki Education staff know about courses that never got submitted, after they pass the `start` date.

This will be more effective as an automated email that goes to the instructors themselves, which will require reworking these alerts a bit.

The alert should be generated (instead of the current conditions) for an submitted course, if it was created at least 3 days ago and if it starts within the next three days (and not more than a week ago). That will generally mean that instructors with unsubmitted courses will get an email three days before the start date, as long as they created the course 3+ days ago.

Here's draft copy:

Reminder:Submit your Wiki Education course page
----

Dear [Instructor full name],

Just a quick reminder that you still need to hit "submit" on your Wiki
Education course page. [Would it be possible to link the submit option
right in the email?]

After you hit submit, we'll review your Wikipedia assignment and make it
live on our system. Your students cannot begin their Wikipedia projects
until you've submitted your page for review.

Thank you, and we look forward to working with you!

Best,
[classroom program manager full name]

PR #2436, opened 2019-01-30
Send UnsubmittedCourseAlerts to instructors before course starts

bwreid:
This should switch emails to go out to the instructor for a course 3 days before the start of a course, assuming they created the course at least 3 days before that. Created to solve Issue #2435.

### DIFF
--- a/app/models/alert_types/unsubmitted_course_alert.rb
+++ b/app/models/alert_types/unsubmitted_course_alert.rb
@@ -34,4 +34,9 @@ class UnsubmittedCourseAlert < Alert
   def reply_to
     user&.email
   end
+
+  def send_email
+    return if emails_disabled?
+    UnsubmittedCourseAlertMailer.send_email(self)
+  end
 end
--- a/app/models/alert_types/unsubmitted_course_alert.rb
+++ b/app/models/alert_types/unsubmitted_course_alert.rb
@@ -31,10 +31,6 @@ class UnsubmittedCourseAlert < Alert
     course_url
   end
 
-  def reply_to
-    user&.email
-  end
-
   def send_email
     return if emails_disabled?
     UnsubmittedCourseAlertMailer.send_email(self)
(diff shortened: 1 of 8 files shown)


ragesoss:
Moving to an email that goes to instructors instead of staff means we'll need to add a new mailer and view, so we can make it a nicely formatted message along the lines of the draft in #2435.

(This ticket isn't urgent, though; we should return to it after we do the server work we've got planned.)

app/models/alert_types/unsubmitted_course_alert.rb
```diff
@@ -34,4 +34,9 @@ def url
   def reply_to
     user&.email
```
ragesoss: This method is now dead code, I think.

lib/alerts/unsubmitted_course_alert_manager.rb
```diff
@@ -8,26 +8,17 @@ def create_alerts
...
-  # Returning instructors work with the Classroom Program Manager.
-  def alert_recipient(course)
-    if course.tag? 'first_time_instructor'
-      SpecialUsers.outreach_manager
-    else # 'returning_instructor', and fallback just in case a course has neither tag.
-      SpecialUsers.classroom_program_manager
-    end
+    ClassroomProgramCourse.unsubmitted
+                          .where('courses.created_at <= ?', TIME_WINDOW.days.ago)
+                          .where('courses.start <= ?', TIME_WINDOW.days.from_now)
```
ragesoss: It looks like this won't exclude old unsubmitted courses. We only want to do this basically at the point when a course is just nearing its start date. Maybe, following the pattern from the old version, `.where(start: TIME_WINDOW.days.ago..TIME_WINDOW.days.from_now)`.

Also, maybe either rename the constant or include the days in it (`TIME_WINDOW = 3.days`), just so that it's clear what exactly the value means.

PR #2451, opened 2019-02-09
Fix #1286 - With help from Kachulio1

### LINKED ISSUE #1286: Users should be able to cycle through diffs in the Activity view

khyatisoneji:
screen shot 2019-02-09 at 5 04 19 pm

### DIFF
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -22,12 +22,6 @@ const Article = createReactClass({
     selectedIndex: PropTypes.number
   },
 
-  getInitialState() {
-    return {
-      isArticleOpened: false
-    };
-  },
-
   fetchArticleDetails() {
     if (!this.props.articleDetails) {
       this.props.fetchArticleDetails(this.props.article.id, this.props.course.id);
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -11,6 +11,7 @@ const Article = createReactClass({
 
   propTypes: {
     article: PropTypes.object.isRequired,
+    index: PropTypes.number,
     course: PropTypes.object.isRequired,
     fetchArticleDetails: PropTypes.func.isRequired,
     articleDetails: PropTypes.object,
@@ -76,6 +77,7 @@ const Article = createReactClass({
           />
           <DiffViewer
             fetchArticleDetails={this.fetchArticleDetails}
+            index={this.props.index}
             revision={this.props.articleDetails && this.props.articleDetails.last_revision}
             first_revision={this.props.articleDetails && this.props.articleDetails.first_revision}
             showButtonLabel={I18n.t('articles.show_cumulative_changes')}
@@ -84,6 +86,7 @@ const Article = createReactClass({
             showSalesforceButton={Boolean(Features.wikiEd && this.props.current_user.admin)}
             course={this.props.course}
             article={this.props.article}
+            shouldShowDiff={this.props.shouldShowDiff}
             showDiff={this.props.showDiff}
             hideDiff={this.props.hideDiff}
             isFirstArticle={this.props.isFirstArticle}
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -18,10 +18,34 @@ const Article = createReactClass({
     showOnMount: PropTypes.bool
   },
 
+  getInitialState() {
+    return {
+      isArticleOpened: false
+    };
+  },
+
   fetchArticleDetails() {
     if (!this.props.articleDetails) {
       this.props.fetchArticleDetails(this.props.article.id, this.props.course.id);
     }
+  },
+
+  showDiff() {
+    this.setState({
+      isArticleOpened: true
+    });
+  },
+
+  shouldShowDiff() {
+    return this.state.isArticleOpened;
+  },
+
+  isFirstArticle() {
+    return true;
+  },
+
+  isLastArticle() {
+    return true;
   },
 
   render() {
@@ -71,6 +95,10 @@ const Article = createReactClass({
             showSalesforceButton={Boolean(Features.wikiEd && this.props.current_user.admin)}
             course={this.props.course}
             article={this.props.article}
+            showDiff={this.showDiff}
+            shouldShowDiff={this.shouldShowDiff}
+            isFirstArticle={this.isFirstArticle}
+            isLastArticle={this.isLastArticle}
           />
         </td>
       </tr>
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -93,8 +93,6 @@ const Article = createReactClass({
             isLastArticle={this.props.isLastArticle}
             showPreviousArticle={this.props.showPreviousArticle}
             showNextArticle={this.props.showNextArticle}
-            revisionAuthor={this.props.articleDetails && this.props.articleDetails.editors[0]}
-            revisionName={this.props.article.title}
           />
         </td>
       </tr>
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -93,6 +93,7 @@ const Article = createReactClass({
             isLastArticle={this.props.isLastArticle}
             showPreviousArticle={this.props.showPreviousArticle}
             showNextArticle={this.props.showNextArticle}
+            articleTitle={this.props.article.title}
           />
         </td>
       </tr>
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -93,6 +93,8 @@ const Article = createReactClass({
             isLastArticle={this.props.isLastArticle}
             showPreviousArticle={this.props.showPreviousArticle}
             showNextArticle={this.props.showNextArticle}
+            revisionAuthor={this.props.articleDetails && this.props.articleDetails.editors[0]}
+            revisionName={this.props.article.title}
           />
         </td>
       </tr>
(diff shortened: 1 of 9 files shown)


ragesoss:
This is coming along for the Activity tab. Some improvements that I think should be made before it's ready to deploy:
* Better visual indication of which diff is the active one. Currently, the +/- button disappears for the open diff. Since you can now cycle through them, it would make sense to highlight the entire row of the open diff in some way.
* Show the article name and username within the DiffViewer. Since this information is only present in the table at present, there's no way to tell whether the current diff is by the same user and to the same article as the previous one.

There are also some warnings, including one regarding the update of state during render.

There are some bugs in the Articles and Students tabs, at least when there are multiple DiffViewer instances being rendered. In both cases, the 'close' button doesn't work for me, and on the Students tab, I get this — maybe from lots of instances all displaying at once?

lots of diffs

ragesoss:
Nice!

The layout of information outside the diff could be improved a lot:

* For attributing the edits, use the same language as the ArticleViewer legend: "Edits by". (The author of the edits isn't necessarily the author of the whole article.)
* List multiple editors for a cumulative diff that includes multiple editors (the same way that ArticleViewer does)
* The article title and the editor usernames are highly relevant information, while the labels are not. The title probably doesn't need a label if it's formatted appropriately. The way ArticleViewer handles it might be a good pattern to follow.
* The information that is already there is center-aligned, while the new info is left-aligned. I recommend moving it all to left-aligned.
* With the new info making this even more of a complex UI, the edit summary, timestamp and byte change details should probably get their own labels.

There's also still the warning about updating state from within a render method, which I only see happening on this branch. That should be fixed before this is merged.

ragesoss:
In particular, it looks like the warning is from the call to `fetchRevisionDetailsIfOpened` (which modifies state) at the beginning of the render method. That should be moved outside the render, perhaps by using `getDerivedStateFromProps`.

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -171,11 +203,51 @@ const DiffViewer = createReactClass({
...
+        <p>{I18n.t('articles.article_title')}: {this.props.articleTitle}</p>
+      </div>
+    );
+  },
+
+  authorsHTML() {
+    return (
+      <div className="user-legend-wrap">
+        <div className="user-legend">{I18n.t('users.edits_by')}</div>
+        <div className="user-legend">{this.props.editors.join()}</div>
```
ragesoss: This needs a space between each username as well.

ragesoss:
This works pretty smoothly now, and functionality-wise I'm happy to merge it with the tweak to fix the spacing between usernames.

Architecture-wise, I don't really like the control structure here, with a whole lot of duplication across each of the parent components (StudentDrawer, ArticleList, RevisionList) and 'index' being the bridge between the parent and child components, with each DiffViewer instance using index to determine for itself whether to be in the 'show' state or not. I don't have a better idea at the moment, though.

@bwreid can you take a look and let us know what you think?

bwreid:
@khyatisoneji Wow, this is a really great feature! Nice work. :)

Addressing Sage's comments:

#### Using `index` as the bridge between components
I can see why you used index here and are passing down much of the state management. Because each article and index will be re-rendered whenever the article list changes, using `index` should be fine but it does feel a bit vague by the time you get down to the actual `diff_viewer.jsx` component.

A different way I can think of to solve this would be to add a key like `article.visible = false` into each article as [you're mapping over the articles.](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2451/files#diff-33ebcdfde49c13a1cf72016b40a7db39R106) Then, instead of always checking the index you're instead checking a specific named property. The code might look something like:

```js
const articleElements = this.props.articles.map(article => {
  article.visible = false;
  <Article
    .../>
  // cont...
})
```

This would lead to the functions that determine the first/last/next/previous article to be a bit more complex. @ragesoss Would this address some of the concern you have?

#### Duplication of show/hide code

This is getting to be very un-DRY code. All of these have to do with state though, so building utility functions might get a bit gross if you're having to pass around `this.setState` as a function or something like that. This might be a good time to try using some ES6 classes and create a parent class for all of these components.

```js
class Rotator extends Component {
  constructor () {
    this.state = {
      selectedIndex: -1
    }

  // methods...
}

// revision_list.jsx
class RevisionList extends Rotator {
  // ...
}
```

Either way, this might be a problem to tackle apart from this feature. Ideally right afterwards.  :)

@khyatisoneji Let me know if this feedback makes sense -- or if you've already tried one of these and it totally didn't work, you're more of the expert on these components than I am! 😄

Side note, should [this method](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2451/files#diff-053fa52348b1a517b814b5e72b3c80fcR100) just be deleted at this point?

khyatisoneji:
@bwreid Thanks for your inputs.

I agree that the deduplication of hiding/show code can be done by a higher order component.

I am unsure of how instead of passing an index and passing a boolean value `visible` would be helpful in our case. I am curious about how the showDiff() function would work then, because if it doesn't get index, it won't know which particular DiffViewer to keep open.

For example:
```
// diff_viewer.jsx
showDiff() {
    this.props.showDiff(); // Here we don't pass the information about which row to open, as we are not passing index
},
render() {
       <button onClick={this.showDiff}>Open</button>
}
```

```
// revision_list.jsx
showDiff() {
    // confused about what do we set here to know that a particular row is opened.
}

const articleElements = this.props.articles.map(article => {
  article.visible = false;
  <Article
        showDiff={this.showDiff}
    .../>
})
```

Please let me know what are your thoughts on this!

ragesoss:
@khyatisoneji yeah, we'd need to pass some kind of identifier to the showDiff function that would set the new state for which article or revision is the open one.

@bwreid changing it from index to an article property (or a revision property, for RevisionList and StudentDrawer) doesn't really help with the part that makes me a little unhappy, which I think is coupling of state between the list components and DiffViewer. DiffViewer changes the state of the list, and that triggers a change in state of the DiffViewer.

Can we get rid of the `isArticleOpened` state within the DiffViewer altogether, and make whether or not to be opened a function of just the props? I'm not sure, but I think this could be simplified so that there are only four control-related props that need to be passed: `index`, `selectedIndex`, `lastIndex`, and `setSelectedIndex`. All the rest — show, hide, next, previous, isFirst, isLast — could be implemented inside DiffViewer so that there's much less duplicated code.

khyatisoneji:
Sorry if there were some unnecessary commits.
Please review the latest changes, I have incorporated the changes that @ragesoss suggested.

Hope it works seamlessly now

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -11,11 +11,21 @@ const Article = createReactClass({
...
     articleDetails: PropTypes.object,
     wikidataLabel: PropTypes.string,
-    showOnMount: PropTypes.bool
+    showOnMount: PropTypes.bool,
+    setSelectedIndex: PropTypes.func,
+    lastIndex: PropTypes.number,
+    selectedIndex: PropTypes.number
+  },
+
+  getInitialState() {
```
ragesoss: Does this do anything in Article? I don't see state used at all in this component.

app/assets/javascripts/components/articles/article_list.jsx
```diff
@@ -8,84 +9,101 @@ import List from '../common/list.jsx';
...
+  },
+
+  showDiff(index) {
+    this.setState({
+      selectedIndex: index
+    });
+  },
+
+  render() {
+    const keys = {
```
ragesoss: No need for this constant to be defined inside `render`, where it will get redefined on every render. (I realize that's what's already happening in the stateless functional version of this component, but it's worth fixing now.) It can be constant outside of the class.

ragesoss: Whoops, nevermind, looks like the `keys` object gets modified.

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -53,20 +56,44 @@ const DiffViewer = createReactClass({
...
+    return this.props.index === 0;
+  },
+
+  isLastArticle() {
+    return this.props.index === this.props.lastIndex - 1;
   },
 
   handleClickOutside() {
-    this.hideDiff();
+    // this.props.hideDiff();
```
ragesoss: For handleClickOutside, it should either be reimplemented to only call `hideDiff` if it's currently open (because I assume interference between the multiple instances of this component is why it needed to be commented out), or this component shouldn't use the `OnClickOutside` high order component wrapper at all.

I think keeping the OnClickOutside behavior and just fixing it so that it works with this new functionality is slightly preferable, but I don't particularly mind removing it.

app/assets/javascripts/components/revisions/revision.jsx
```diff
@@ -3,37 +3,64 @@ import PropTypes from 'prop-types';
 import moment from 'moment';
 import DiffViewer from './diff_viewer.jsx';
 import CourseUtils from '../../utils/course_utils.js';
+import createReactClass from 'create-react-class';
 
-const Revision = ({ revision, wikidataLabel, course }) => {
```
ragesoss: It's not a big deal either way, but is it necessary to convert this from a function component to a class component?

bwreid:
Is this working for you locally still, @khyatisoneji? I'm encountering two problems that seem to start at [this commit](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2451/commits/6576aa84f6175df95bf796f263c919fed4414300):

1. The error `Cannot read property 'fetched' of null` is appearing when clicking on the Diff Viewer from the Articles tab on a course.
1. Clicking *Next* and *Prev* don't seem to work as expected. Clicking these buttons once shows the Loading GIF and nothing else. Only after clicking back and forth on Next and Prev does the article load.

khyatisoneji:
@bwreid There was one issue where we were not using nextProps and still using this.props.
I have made the fix, please pull the latest code and verify.

Thanks.

bwreid:
@khyatisoneji This looks great. Thanks for sticking with this! :)

PR #2453, opened 2019-02-10
fix #2452: show due date only when block type is assignment

### LINKED ISSUE #2452: Due date still visible when Block type is changed
Once we change the block type of any block to "assignment" we can set its due date and it is shown alongside the Edit button. But as we change the type to any other type like "in class" or "milestone" still the due date is shown.
issue

shivanshbindal9:
Now, the due will be shown only when the block type is "assignment".


shivanshbindal9:
Can you please review this. @ragesoss

PR #2472, opened 2019-02-13
Upgrade from rails 5.2 to rails 6 beta

### LINKED ISSUE #2432: Upgrade Rails to 6.0.0.beta1

faithngetich:
What does this PR do.
------------------------------------------
This PR updgrades rails version from 5.2.2 to 6.0.0.beta1 version

- Fixes #2432


Gemfile
```diff
@@ -2,9 +2,9 @@ source 'https://rubygems.org'
 ruby '2.5.0'
 
 ### Basic Framework
-gem 'rails', '5.2.2'
+gem 'rails', '~> 6.0.0.beta1'
 gem 'jbuilder' # DSL for building JSON view template
-gem 'haml-rails' # HTML template language, used instead of ERB
+gem 'haml-rails', git: "https://github.com/indirect/haml-rails.git" # HTML template language, used instead of ERB
```
ragesoss: I assume this will be updated with a new release before or soon after Rails 6 final is released.

app/controllers/salesforce_controller.rb
```diff
@@ -24,7 +24,7 @@ def create_media
     url = CreateSalesforceMediaRecord.new(article: @article, course: @course, user: @user,
                                           before_rev_id: params[:before_rev_id],
                                           after_rev_id: params[:after_rev_id]).url
-    redirect_to url
+    redirect_to url, allow_other_host: true
```
ragesoss: Hmm... did you find these yourself, did tests fail to lead you to these cases of external redirects? I wonder if there are others that aren't covered.

faithngetich: The tests were failing locally before I made the change, but running them now passes without this change. I must have run the test in the old branch and assumed it applies to this new branch.

Gemfile
faithngetich: Hopefully, this will be the case.

app/controllers/salesforce_controller.rb
ragesoss: Cool. I don't see any mention of the default being changed for allow_other_hosts — https://github.com/rails/rails/releases/tag/v6.0.0.beta1 . If this is necessary, it shouldn't be changed.

faithngetich: It actually exists, in the Action Pack section.

> Ensure external redirects are explicitly allowed
      Add fallback_location and allow_other_host options to redirect_to.
      Gannon McGibbon

What I am not sure is if we need to `add fallback_location`

ragesoss: Ah, I guess "Ensure external redirects are explicitly allowed" means that the default changed from true to false? In Rails 5, I think this only applied to 'redirect_back' *and* it defaulted to true anyway.

faithngetich:
Hey, @ragesoss I am assuming we are waiting for the rails 6 **Beta 2**, **Release Candidate**, **Release Candidate 2**, and the  **Final release** version then we can build up on this and finally merge it to our project. Or what did you have in mind?

ragesoss:
@faithngetich we don't merge a bump in the rails version until then, although if the changes in files other than Gemfile/Gemfile.lock are compatible with with Rails 5, then those could be done separately and merged. Ideally, everything or almost everything would be changed ahead of time, so that the final release just requires the gem switches but no code changes.

bwreid:
@ragesoss Is there any reason to keep this at this point?

ragesoss:
Nope.

Issue #2474, opened 2019-02-13
Set Classroom Program Manager as default instructor for Classroom Program courses


bwreid:
For the Classroom Program, typically the Classroom Program Manager is added as a Wiki Ed Staff member for every single course. To make this user experience easier, we should by default add that user.

See `courses_controller.rb` and `list_course_manager.rb`.

PR #2475, opened 2019-02-13
Set Classroom Program Manager as default instructor

bwreid:
Issue #2474.

When a campaign is added to a course (i.e. approved), the user designated as the Classroom Program Manager should be added by default as a Wiki Ed Staff member.

### DIFF
--- a/app/assets/javascripts/components/overview/campaign_editable.jsx
+++ b/app/assets/javascripts/components/overview/campaign_editable.jsx
@@ -12,6 +12,7 @@ import Popover from '../common/popover.jsx';
 import Conditional from '../high_order/conditional.jsx';
 
 import { removeCampaign, fetchAllCampaigns, addCampaign } from '../../actions/campaign_actions';
+import { fetchUsers } from '../../actions/user_actions';
 
 
 const CampaignEditable = createReactClass({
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/overview/campaign_editable.jsx
```diff
@@ -55,7 +56,11 @@ const CampaignEditable = createReactClass({
   },
 
   addCampaign() {
-    this.props.addCampaign(this.props.course_id, this.state.selectedCampaign.value);
+    // After adding the campaign, request users so that any defaults are
+    // immediately propogated.
```
ragesoss:
```suggestion
    // immediately propagated.
```

app/assets/javascripts/components/overview/campaign_editable.jsx
```diff
@@ -55,7 +56,11 @@ const CampaignEditable = createReactClass({
...
+    this.props.addCampaign(this.props.course_id, this.state.selectedCampaign.value)
+    .then(() => this.props.fetchUsers(this.props.course_id));
```
ragesoss: Nice!

lib/list_course_manager.rb
```diff
@@ -39,6 +40,17 @@ def add_instructor_real_names
...
   end
 
+  def add_classroom_program_manager_if_exists
+    cpm = SpecialUsers.classroom_program_manager
+    if cpm
+      CoursesUsers.create(user: cpm,
+                          course: @course,
+                          role: CoursesUsers::Roles::WIKI_ED_STAFF_ROLE,
+                          real_name: cpm.real_name,
+                          role_description: 'Wiki Ed Staff')
```
ragesoss: I think role_description is better left out. We don't normally enter it for staff, and if present it will show up in the list of users in the Details section, and it duplicates the label "Wiki Ed Staff" that is already at the front of the list for this role.

PR #2480, opened 2019-02-14
Update activerecord-import

### LINKED ISSUE #2479: Upgrade activerecord-import to version 1.0
See https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2440

Upgrading this gem will require a few code changes to retain the old behavior, which we rely on.

khyatisoneji:
Fix #2479


ragesoss:
nice!

PR #2488, opened 2019-02-15
Properly name engagements in salesforce

bwreid:
These changes should add more information to the Engagement record on Salesforce.

### DIFF
--- a/app/services/create_salesforce_media_record.rb
+++ b/app/services/create_salesforce_media_record.rb
@@ -29,6 +29,7 @@ class CreateSalesforceMediaRecord
     {
       Title__c: @article.full_title,
       Name: @article.full_title,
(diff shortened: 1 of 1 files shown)


app/services/create_salesforce_media_record.rb
```diff
@@ -28,6 +28,8 @@ def create_salesforce_record
   def salesforce_media_fields
     {
       Title__c: @article.full_title,
+      Name: @article.full_title,
+      Is_New_Article__c: ArticlesCourses.find_by(article: @article).new_article,
```
ragesoss: This will not necessarily be for the correct course, if multiple courses have edited the same article. Add `course: @course` to the find params and it should be right.

Issue #2473, opened 2019-02-13
Show count of ungreeted users in admin `dashboard` view


ragesoss:
Wikipedia Experts would really like to be able to see which of their courses have (large numbers of) ungreeted students.

Doing this immediately may or may not introduce a performance bottleneck, since we don't cache the ungreeted count on the `Course` record. On a quick test, I think it might be fine, as long the courses query gets an `includes(:students)` tacked on so that it's not an N+1.

PR #2489, opened 2019-02-15
Display number of students left to be greeted on home page

bwreid:
Issue #2473 
- Adds a sortable list to the home page for how many students are
  left to be greeted
- Should display for Wiki Edu admins only

If you'd rather not add another feature test, I will understand.  :)

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -35,11 +35,11 @@ class DashboardController < ApplicationController
 
     return unless current_user.admin?
     @submitted = Course.submitted_but_unapproved
-    @strictly_current = current_user.courses.strictly_current
+    @strictly_current = current_user.courses.strictly_current.includes(:students)
   end
 
   def current_courses
-    current_user.courses.current_and_future.includes(:students)
+    current_user.courses.current_and_future
   end
 
   def past_courses
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -39,7 +39,7 @@ class DashboardController < ApplicationController
   end
 
   def current_courses
-    current_user.courses.current_and_future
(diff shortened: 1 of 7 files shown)


ragesoss:
Feature test looks good. This has N+1 query behavior now, though, with the ungreeted query running for each course.

app/controllers/dashboard_controller.rb
```diff
@@ -39,7 +39,7 @@ def set_admin_courses_if_admin
   end
 
   def current_courses
-    current_user.courses.current_and_future
+    current_user.courses.current_and_future.includes(:students)
```
ragesoss: The admin courses come from `@strictly_current` on line 38, rather than here.

app/presenters/dashboard_presenter.rb
```diff
@@ -32,6 +32,10 @@ def campaigns
     current_user.campaigns
   end
 
+  def ungreeted_students(course)
+    course.students.ungreeted.length
```
ragesoss: Even with the `includes(:students)` moved to the right place above, this still triggering a query for each course, because `ungreeted` is an ActiveRecord scope.

For the case where there are no students, doing it as an array operation makes it 2x faster and removes the N+!, but I haven't tested what it's like in more like production conditions where most courses have many students. `course.students.reject(&:greeted).count`

But I think finding a way to do the calculation of the count in that main query (or a separate query for the counts per course, all in one) will be the ideal way to do it, if we can figure out how.

ragesoss:
My worry is that that the `to_a` approach might be noticeably slow when there are lots of students (~3000 or more, across all of a Wikipedia Expert's classes), just because of all the extra ruby objects that will need to get created.

If there's no nice query-based solution to that, we could add a `greeted_count` db column to `Course` and update it along with the other course stats during the update cycle.

But we can deploy this for now and see if there actually is a performance problem or not.

Issue #2032, opened 2018-10-04
Some test are failing when run in different timezone


Kachulio1:
These are some of the tests that are failing when run in **East Africa Time (EAT)**

![screen shot 2018-10-04 at 10 52 58](https://user-images.githubusercontent.com/22243090/46460137-bdb36180-c7c3-11e8-9a0c-7b80f56422b4.png)

./spec/features/campaign_overview_spec.rb:212
./spec/features/campaigns_spec.rb:77
 ./spec/features/course_page_spec.rb:158
./spec/models/campaign_spec.rb:90
./spec/models/campaign_spec.rb:99
./spec/models/campaign_spec.rb:107
./spec/models/campaign_spec.rb:135

chalchuck:
Hello @Kachulio1 @ragesoss please assign me to this issue.

Kachulio1:
@chalchuck Great, you can work on the issue.

ragesoss:
I think we've taken care of all the cases where this happened.

PR #2496, opened 2019-02-19
Fix Timezone Failing Tests

bmwenda:
#### What this PR does
Fixes issue  #2032 

#### Context
Some tests were failing on East African Time (EAT). I tracked this down to [this line](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/bb81026d3a1a0797c89c4a0e806ab56a98b83b8b/app/models/campaign.rb#L119).

The bug was that `Time.parse` uses system time to parse time. Since time is stored in `UTC` on the DB, this causes some inconsistencies. My approach was to prefer `Time.zone.parse` which uses the configured application timezone.

The side effect of this was that `Time.zone.parse(invalid string)` returns `nil`, rather than raise an exception like `Time.parse`. Also, this fix means that we will always default to `UTC`

Not sure if this is okay for now? I imagine that a long term viable support for timezone would be allowing users to configure their timezone on setup, which would require some backend changes to allow for this.

### DIFF
--- a/app/assets/javascripts/components/common/date_picker.jsx
+++ b/app/assets/javascripts/components/common/date_picker.jsx
@@ -214,7 +214,7 @@ const DatePicker = createReactClass({
   },
 
   moment(...args) {
-    return this.props.showTime ? moment(...args) : moment(...args).utc();
+    return this.props.showTime ? moment.parseZone(...args) : moment.utc(...args);
   },
 
   showCurrentDate() {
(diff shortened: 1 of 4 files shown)


bwreid:
Thanks for this PR, @bmwenda. I'll take a closer look at this within the next couple days.

Could you fix the rubocop error that is surfacing in the meantime?

bmwenda:
> Thanks for this PR, @bmwenda. I'll take a closer look at this within the next couple days.
> 
> Could you fix the rubocop error that is surfacing in the meantime?

@bwreid Sure, looking into it

Issue #2497, opened 2019-02-19
Initiate cloning from a course, if it's cloneable


bwreid:
As an admin, we want the ability to start cloning a course directly from an old course page as opposed to having to click the "Create Course" button. This will decrease the number of steps needed to recreate a course.

Currently much of the user-facing code for cloning exists in `course_creator.jsx`.

PR #2506, opened 2019-02-20
Add Course Clone button to Course Overview for admins

bwreid:
Solves issue #2497.

- Adds a course clone button for admins on WikiEd
- Adds AvailableActions component testing for both dashboards

### DIFF
--- a/app/assets/javascripts/components/overview/clone_course_button.jsx
+++ b/app/assets/javascripts/components/overview/clone_course_button.jsx
@@ -1,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { cloneCourse } from '../../actions/course_creation_actions.js';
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/overview/clone_course_button.jsx
```diff
@@ -0,0 +1,43 @@
...
+  );
+}
+
+function onClickConfirmation(props) {
+  return () => {
+    const clone = () => {
+      props.cloneCourse(props.courseId).then(({ data }) => {
+        const course = data.course;
+        props.updateCourse(course);
+        window.location = `/courses/${course.slug}`;
```
bwreid: I'm not super happy about this. Ideally we'd use `withRouter` to give this component the `history` object so that we can use `history.push()`. However, doing so does not trigger the form to pop up for the cloned course. It looks like it happens on `UPDATE_COURSE`...

Issue #2500, opened 2019-02-19
Change Role field to a dropdown with the ability to enter your own text


bwreid:
As a user creating a course, I want to be able to set my role via a dropdown as opposed to having to enter free form text. At the same time, I want the option to be able to define my own role if applicable. In general, the goal is to make course creation easier on the user and standardize our own data on the backend.

We should use the [react-select](https://react-select.com/) library to create the form field. Roles should include "Instructor" and "Teaching assistant" and should be included in a translation file. Right now the placeholder text can be [found here](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/config/locales/en.yml#L406).

PR #2513, opened 2019-02-21
Allow for selection or typing in role field

bwreid:
Issue #2500 

This changes the `role` field to be a dropdown field where you can also type in a custom response.


ragesoss:
This is  a little wonky. It works fine if you select one of the known roles, and it works if you type and then hit enter, or tab.

However, if you type and then click out of the input (eg, clicking another input, or clicking 'Create my Course', the input reverts to the previous value (which will be nothing, if this is the first time you've interacted with it).

Maybe there's a way to treat a blur event the same way as selecting a value or tab/enter?

bwreid:
Interesting! I'll take a look.

ragesoss:
Looks good to me now.

Issue #2503, opened 2019-02-19
Make some blocks un-deletable or un-editable


bwreid:
As admin, we want the ability to not allow instructors to manually edit or delete certain blocks from their created courses. A good place to start for this functionality may be in one of the [content.yml](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/config/wizard/researchwrite/content.yml) files that sets up a block. Likely we want to add some attribute that specifies a block as unchangeable. Then, we will need to reflect that value in the UI.

PR #2514, opened 2019-02-21
[WIP] Allow for blocks to be uneditable or undeletable

bwreid:
Pull Request to solve Issue #2503.

image

Any other fields we shouldn't be allowing edits too? The only other one I can see is Block Type but that actually seems like it might be useful for instructors to always be able to modify.

### DIFF
--- a/app/assets/javascripts/components/timeline/block.jsx
+++ b/app/assets/javascripts/components/timeline/block.jsx
@@ -111,7 +111,13 @@ const Block = createReactClass({
     // let graded;
     if (isEditable) {
       if (!this.props.block.is_new) {
-        deleteBlock = (<div className="delete-block-container"><button className="danger" onClick={this.deleteBlock}>Delete Block</button></div>);
+        const button = this.props.block.is_deletable ? (
+          <button className="danger" onClick={this.deleteBlock}>Delete Block</button>
+        ) : (
+          <button disabled={true}>You may not delete this block</button>
+        );
+
+        deleteBlock = (<div className="delete-block-container">{ button }</div>);
       }
       className += ' editable';
       if (this.props.isDragging) { className += ' dragging'; }
@@ -163,7 +169,7 @@ const Block = createReactClass({
       <span className="block__due-date-spacer"> - </span>
     ) : undefined;
 
-    const editButton = this.props.editPermissions ? (
+    const editButton = this.props.editPermissions && this.props.block.is_editable ? (
       <div className="block__edit-button-container">
         <button className="pull-right button ghost-button block__edit-block" onClick={this._setEditable}>Edit</button>
       </div>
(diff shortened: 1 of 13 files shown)


app/models/course_data/block.rb
```diff
@@ -50,4 +54,18 @@ def calculated_date
...
   end
+
+  private
+
+  def editable?
+    throw :abort unless is_editable
+  end
+
+  def content_fields_changed?
+    content_changed? || title_changed? || training_module_ids_changed?
```
ragesoss: Are these magic Rails metaprogramming methods?

app/models/course_data/block.rb
```diff
@@ -23,6 +25,8 @@ class Block < ApplicationRecord
   belongs_to :week
   has_one :course, through: :week
   serialize :training_module_ids, Array
+  before_update :editable?, if: :content_fields_changed?
+  before_destroy :deletable?
```
ragesoss: These blocks should still be deleted if an instructor is deleting the entire timeline (so that they can go throught the wizard again), but they shouldn't be able to delete the week containing an indestructible block, nor the block directly.

This will probably shake out in the feature specs, once we configure the wizard with the relevant blocks.

app/assets/javascripts/components/timeline/block.jsx
```diff
@@ -111,7 +111,13 @@ const Block = createReactClass({
     // let graded;
     if (isEditable) {
       if (!this.props.block.is_new) {
-        deleteBlock = (<div className="delete-block-container"><button className="danger" onClick={this.deleteBlock}>Delete Block</button></div>);
+        const button = this.props.block.is_deletable ? (
+          <button className="danger" onClick={this.deleteBlock}>Delete Block</button>
+        ) : (
+          <button disabled={true}>You may not delete this block</button>
```
ragesoss: These button labels should have i18n.

bwreid:
So sorry, I don't think I saw these comments @ragesoss ! I'll get to them now.

app/models/course_data/block.rb
bwreid: Yes! Rails magic indeed ✨

app/models/course_data/block.rb
bwreid: Sorry, I'm not sure I quite understand this first part. The request is:

1. If an instructor is deleting all of the blocks, they should be able to delete the block even if `deletable?` returns false
1. If an instructor is attempting to delete a week that contains an indestructible block, the entire week should not be able to be deleted. (?)
1. The instructor should not be able to delete an indestructible block.

Did I get all the use cases right?

bwreid: One option for this is just to ignore validation on the server side. It seems odd someone with access to edit the backend would try and go around the UI.

ragesoss: Yeah, you got all those use cases right. There's a feature to delete the entire timeline, which we still need to keep as an option. But deleting a week with an indestructible block should be impossible for an instructor on the frontend.

Issue #2471, opened 2019-02-13
Admins should be able to notice and act on requested accounts more easily


ragesoss:
On both production instances of the Dashboard now, we've got the requested accounts feature in use, but the biggest pain point seems to be account requests that never get processed; there are currently about 80 of them on Programs & Events Dashboard, likely because of cases where the program organizer didn't realize they needed to take action to create the requested accounts.

I'm not sure what the ideal fix for this is, but we need something to make it easy to act on the requested accounts that are pending.

bwreid:
So we now have notifications for how many requested accounts are pending appearing on the main course page. If this is not sufficient and we need an additional way to notify instructors that these exist, the only other avenues we have would be email and/or notifying them via Wikipedia. It doesn't look like you need an email to sign up for a Wikipedia account but would that still be a reliable way to get in touch?

If so, we could likely do something similar to the Unsubmitted Course Alerts except for requested accounts.

ragesoss:
We show them on an individual course page; I think the idea would be to highlight them in a system-wide way, so that an admin can notice when there's a requested account for a program they don't know about.

The recent change will probably take care of 80% of the problem for P&E Dashboard, but doesn't really help much for Wiki Education Dashboard, where the current best option is for staff to periodically check the campaign page for the current term (which links to the aggregate requests for all courses in that campaign).

bwreid:
Thanks @ragesoss. So I think my current idea is to move another banner to the home page that will say something like "12 requested accounts pending." Then, it'll link to a new page that will aggregate all the requested accounts on one page.

ragesoss:
@bwreid the home page (only), or the nav component?

This could be a good starting point for a notifications counter, if the nav does a query (for admins) of how many requested accounts there are.

bwreid:
I was just thinking the homepage, but adding it to the nav would be great! I'll first start on just creating the new page and then will start to add to the nav.

PR #2523, opened 2019-02-26
[WIP] Create page to show all requested accounts for a user

bwreid:
Related to issue #2471 

image

notifications

### DIFF
--- a/app/assets/javascripts/components/nav/nav.jsx
+++ b/app/assets/javascripts/components/nav/nav.jsx
@@ -4,6 +4,7 @@ import Confetti from 'react-confetti';
 import CustomLink from './CustomLink.jsx';
 import HamburgerMenu from './hamburger_menu.jsx';
 import LanguagePicker from './language_picker.jsx';
+import fetch from 'cross-fetch';
 
 const Nav = createReactClass({
   displayName: 'Nav',
@@ -53,6 +54,12 @@ const Nav = createReactClass({
   },
 
   componentDidMount() {
+    if (Features.wikiEd && this.state.ifAdmin) {
+      fetch('/requested_accounts.json')
+        .then(res => res.json())
+        .then(({ requested_accounts }) => this.setState({ requested_accounts }))
+        .catch(err => err); // If this errors, we're going to ignore it
+    }
     window.addEventListener('resize', this.updateDimensions);
   },
   componentWillUnmount() {
@@ -72,6 +79,7 @@ const Nav = createReactClass({
     let navClass;
     let myDashboard;
     let forAdmin;
+    let notifications;
     let help;
     let Sandbox;
     let wikiEd;
@@ -86,11 +94,30 @@ const Nav = createReactClass({
       );
     }
     if (this.state.userSignedIn) {
+      if (this.state.ifAdmin && this.state.wikiEd) {
+        notifications = (
+          <li aria-describedby="notification-message" className="notifications">
+            <a href="/requested_accounts" className="icon icon-notifications_bell"/>
+            {
+              this.state.requested_accounts
+                ? (
+                  <span className="bubble red">
+                    <span id="notification-message" className="screen-reader">You have new notifications.</span>
+                  </span>
+                )
+                : (
+                  <span id="notification-message" className="screen-reader">You have no new notifications.</span>
+                )
+            }
+          </li>
+        );
+      }
       loginLinks = (
         <span>
           <li>
             <b><a href={`/users/${this.state.currentUser}`} className="current-user">{this.state.currentUser}</a></b>
           </li>
+          { notifications }
           <li>
             <a href={this.state.destroyUrl} className="current-user">{I18n.t('application.log_out')}</a>
           </li>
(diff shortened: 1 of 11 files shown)


app/controllers/requested_accounts_controller.rb
```diff
@@ -42,8 +42,19 @@ def enable_account_requests
     @course.save
   end
 
+  # List of requested accounts for a user's courses.
+  def index
+    if user_signed_in?
+      @courses = current_user.courses.select do |course|
```
ragesoss: I think this would be more useful as an unconditional list of requested accounts, rather than limited to the viewing admin's courses. In most cases, the desired course of action is just to create all of the requested accounts... the only time we wouldn't is if there was someone trying to create a bunch of accounts with bad names, or someone was mass-requesting so many that it's obviously not legit. Minimizing the time until the request is processed  and avoiding cases where requests sit around without being handled are the main goals.

ragesoss:
I like this this layout, breaking it up by course. It provides more/better context than the current campaign requested accounts list.

app/controllers/requested_accounts_controller.rb
bwreid: In addition to the above buttons then, should we also have a "Create all accounts" button?

ragesoss: I think that would be good.

bwreid: Will there be an issue here with using the backup account at all? I imagine this button would often get us over the 6 accounts limit. Last I remember we had a limited amount of additional accounts we could create with our backup account.

ragesoss: The backup account isn't set on Wiki Education Dashboard, but it should be used if necessary on Programs & Events Dashboard.

I believe that the Wiki Education Dashboard is not subject to those IP limits anyway, because its IP is whitelisted.

bwreid:
At the moment I've just added Notifications as a text link. How can we update the font to include a little bell?  :)

ragesoss:
Maybe something like https://www.npmjs.com/package/react-notification-badge ?

app/assets/javascripts/components/nav/nav.jsx
```diff
@@ -56,7 +56,8 @@ const Nav = createReactClass({
   componentDidMount() {
     fetch('/requested_accounts.json')
```
ragesoss: This should only be for admins?

app/assets/javascripts/components/nav/nav.jsx
```diff
@@ -136,6 +142,24 @@ const Nav = createReactClass({
           <CustomLink to="/admin" name="Admin" />
         </li>
       );
+      notifications = (
```
ragesoss: Maybe this should only show for admins as well, unless/until there is the possibility of notifications for other users as well.

app/controllers/requested_accounts_controller.rb
```diff
@@ -42,8 +43,18 @@ def enable_account_requests
     @course.save
   end
 
+  # List of requested accounts for a user's courses. @courses set in before action
+  def index
+    raise_unauthorized_exception unless user_signed_in? && current_user.admin?
+
+    respond_to do |format|
+      format.html { render }
+      format.json { render json: { requested_accounts: @courses.present? } }
```
ragesoss: This seems like it would be doing a whole lot of unnecessary work for the json case, where all we need to a boolean. The `set_course` is going through all the current courses to find requested accounts (maybe an N+1, where it queries each course for its requests?). `RequestedAccount.any?` is probably what we want.

app/controllers/requested_accounts_controller.rb
```diff
@@ -62,15 +73,32 @@ def create_accounts
...
+    end
+
+    render :index
+  end
+
   private
 
+  def set_courses_with_requested_accounts
+    has_requested_accounts = ->(course) { course.requested_accounts.present? }
+    @courses = Course.current_and_future.select(&has_requested_accounts)
```
ragesoss: An alternative approach could be to go through `RequestedAccount` and group by course. Then you only need to load the course records for the ones that have requested accounts, which are going to be a very small fraction of all courses.

app/controllers/requested_accounts_controller.rb
```diff
@@ -62,15 +74,31 @@ def create_accounts
...
+      results << create_account(account)
+    end
+
+    render :index
+  end
+
   private
 
+  def all_requested_accounts
+    RequestedAccount.all.group_by { |account| account.course.title }
```
ragesoss: Not likely to ever be a performance problem, this gets rid of the N+1 query.

```suggestion
    RequestedAccount.all.includes(:course).group_by { |account| account.course.title }
```

app/controllers/requested_accounts_controller.rb
```diff
@@ -62,15 +74,31 @@ def create_accounts
     end
   end
 
+  def create_all_accounts
+    raise_unauthorized_exception unless user_signed_in? && current_user.admin?
+    @courses = all_requested_accounts
+    @results = RequestedAccount.all.inject([]) do |results, account|
```
ragesoss: This is slightly simpler if we `map` instead of `inject`.

bwreid:
bell-icon

Look at this little cutie. 😍

ragesoss:
love love love. I want to merge and deploy now, and review and test later, just to get the reactions from the team as soon as possible. I guess I'll have another look first instead.

PR #2540, opened 2019-02-28
Refactoring of UserContoller done

hardy1334:
For the issue #1949  ,I have divided UserController is further divided into **EnrollCourseController** and in that controller action regarding enrolling a program is there.


ragesoss:
Please have a look at the failing controller specs.

app/views/enroll_course/index.html.haml
```diff
@@ -0,0 +1,26 @@
+.container
```
ragesoss: Is this copied from another location? If so, it should probably be moved instead of duplicated, or if it's used in more than one place, the original file can be explicitly rendered instead of the default rails rendering by path.

app/views/enroll_course/users.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
```
ragesoss: As above: is the a duplicated view?

app/views/enroll_course/index.html.haml
hardy1334: @ragesoss  Yeah,they were copied from **users** view,I'll definetly modify them and remove the duplicate one

app/views/enroll_course/users.json.jbuilder
hardy1334: Yeah, it is ,and I'll definetly resolve the problems in it

PR #2550, opened 2019-03-04
Query params url#2547

### LINKED ISSUE #2547: ArticleFinder query parameters should be included in URL
When an ArticleFinder query is executed, the URL should be updated with those parameters (search term, search type, and any thresholds for views or quality) so that the URL can be shared to replicate the query.

hjhimanshu01:

* Note the issue that the PR addresses. Include `Fixes #2547` 
* Added query parameters to the URL address ( search_term , search_type , article_quality , min_views) for sharing the URL
*On refreshing the page , URL is restored to the original one without any search parameters
*Added screenshot 
screenshot from 2019-03-04 00-23-55


ragesoss:
Cool!
"On refreshing the page , URL is restored to the original one without any search parameters" - this is not the desired behavior. The main use case is the copy the URL with the parameters so that it can be shared with someone else and the same search can be replicated.

hjhimanshu01:
I do understand that though if the user refreshes the page (even by mistake) , old query params would persist in the URL while the article searched would be none , leading to inaccurate results if URL is shared at that point , just an edge case , shall I still remove it ?

ragesoss:
I'm not sure I understand the problem you're describing.

Here's what I think the ideal behavior would be:
* URL does not change until I click 'Submit', at which point it updates the URL to include the params
* Refreshing the page with those params in the URL results in the search term and other params automatically updating the initial state of the refreshed page, so that clicking 'Submit' will re-run the same search.
* Any changes to the search that I make will get updated in the URL when I click 'Submit' again.

hjhimanshu01:
@ragesoss ,done ! added all the functionalities ideally required !

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -52,20 +55,38 @@ const ArticleFinder = createReactClass({
...
   },
-
+  buildURL() {
+    let queryStringUrl = window.location.href.split('?')[0];
+    const search_term = encodeURIComponent(this.props.search_term);
+    const search_type = encodeURIComponent(this.props.search_type);
+    const article_quality = encodeURIComponent(this.props.article_quality);
+    const min_views = encodeURIComponent(this.props.min_views);
+    queryStringUrl += `?search_term=${search_term}&search_type=${search_type}&article_quality=${article_quality}&min_views=${min_views}`;
+    history.pushState(window.location.href, 'query_string', queryStringUrl);
```
ragesoss: I think `pushState` is not what we want here; this fills up the browser history with every change. `replaceState` is probably right here.

ragesoss:
Aside from the `pushState` issue, this works very nicely!

I'm finding the control flow pretty confusing, though. The trip from `getQueryVariable`, to (async) `this.updateFields`, `this.props.updateFields`, and `buildUrl`, and then back to `this.updateFields` for user interactions takes a while to think through, and I'm still unclear about what the `async, await` is doing. If you can think of any ways to refactor this for clarity, please do. (Maybe hooks?)

@bwreid can we get your feedback on this?

db/schema.rb
```diff
@@ -522,4 +522,4 @@
     t.index ["language", "project"], name: "index_wikis_on_language_and_project", unique: true
   end
 
-end
+end
```
ragesoss: unrelated change

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -52,20 +55,38 @@ const ArticleFinder = createReactClass({
...
-  updateFields(key, value) {
-    return this.props.updateFields(key, value);
+  getQueryVariable() {
+    const query = window.location.search.substring(1);
+    const vars = query.split('&');
+    for (let i = 0; i < vars.length; i += 1) {
+        const pair = vars[i].split('=');
+        this.updateFields(pair[0], pair[1]);
+    }
+    },
```
bwreid: I think there might be a simpler way to solve this problem. It looks like this function is doing two jobs:
1. Parsing the query string
2. Updating each field

We already have the [query-string](https://www.npmjs.com/package/query-string) package installed, so I would suggest making use of it like so to solve this first problem:
```js
import qs from 'query-string';
//...
qs.parse(window.location.search) // { article_quality: "100", min_views: "0" ... }
```

Then you'll end up with an object to represent the query params. If you want to turn this into an array of arrays, you can do the following:
```js
const query = qs.parse(window.location.search)
Object.entries(query) // [ [ "article_quality", "100" ], [ "min_views", "0" ], ... ]
```

The reason you have `async/await` (which, is awesome!) is because we want all the updates to finish before we can build the URL, right? Typically when you have multiple asynchronous actions you would use `Promise.all`, which [evaluates an array of promises](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) and then returns an array with all of the final results. An async function will return a promise, so to create an array of promises you could do something like:
```js
const query = qs.parse(window.location.search);
const entries = Object.entries(query);
const updates = entries.map(([key, val]) => this.updateFields(key, val));
// At this point, the updates haven't been made yet!

Promise.all(updates).then(() => this.buildURL());
```

This reads a bit more cleanly to me. You can either remove or rename the `getQueryVariable` function

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -52,20 +55,38 @@ const ArticleFinder = createReactClass({
...
       showFilters: !this.state.showFilters
     });
   },
-
+  buildURL() {
+    let queryStringUrl = window.location.href.split('?')[0];
+    const search_term = encodeURIComponent(this.props.search_term);
+    const search_type = encodeURIComponent(this.props.search_type);
+    const article_quality = encodeURIComponent(this.props.article_quality);
+    const min_views = encodeURIComponent(this.props.min_views);
```
bwreid: Using `.map()` or `.reduce()` might make this a bit easier to read. Let me know if you need help thinking through how to do this!

app/assets/javascripts/components/article_finder/article_finder.jsx
hjhimanshu01: Got it ! Although had intentionally used .pushState() to save the searches to history if a user wants to retrieve old queries , but you are right , it's better not to clutter history with each and every change

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -35,6 +35,9 @@ const ArticleFinder = createReactClass({
   },
 
   componentWillMount() {
+    if (window.location.href.split('?').length !== 1) {
```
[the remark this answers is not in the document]
hjhimanshu01: Hey @bwreid , that's awesome! , would change it , mine's to difficult to comprehend.

app/assets/javascripts/components/article_finder/article_finder.jsx
hjhimanshu01: The async await is there exactly for this reason , without async await it was not updating queries while calling buildURL() instantly , your use of query-string module is again great , sorry hadn't noticed the libraries installed already otherwise I would have used that !

app/assets/javascripts/components/article_finder/article_finder.jsx
hjhimanshu01: I think .forEach() would be more favourable as .map() would create another array and iterate , what do you think ?

app/assets/javascripts/components/article_finder/article_finder.jsx
hjhimanshu01: window.location.search and Object.entries(query) , both of them are producing a console warning. I'm a bit confused why did it happen?

Warning: Failed prop type: "value" must be a number or a range object

hjhimanshu01: I just found out that  the .updateFields(key,val) function is the reason for the warning . Any idea about how could I resolve this , 
`const updates = entries.map(([key, val]) => this.updateFields(key, val));`

Warning: Failed prop type: "value" must be a number or a range object

app/assets/javascripts/components/article_finder/article_finder.jsx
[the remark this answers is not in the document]
bwreid: @hjhimanshu01 I am actually getting this issue without any of the above changes. Steps to reproduce:
* Go to `/:course_slug/article_finder`
* Perform a search
* Refresh the page

Are you not encountering the same issue?

app/assets/javascripts/components/article_finder/article_finder.jsx
bwreid: Ah, I see now. So, I don't think the issue is any of the suggested changes but is rather the fact that when pulling values from the query string, each value will be a string when it might in fact need to be a number (e.g. in the case of `article_quality`).

Two options I can imagine here:
1. For the keys that we know will be a number, convert them to a number before calling `this.props.updateFields()`
1. Dynamically check if each value _could_ be a number. If it is, convert it, otherwise leave it as is

The second option seems more "clever" but I could see it introducing problems and might require a bit of commenting to make it understandable. I think either would work though.  :)

app/assets/javascripts/components/article_finder/article_finder.jsx
bwreid: Ultimately, any of the options are fine. You're right to be suspicious of taking up new memory, but I don't think creating this small array will be too bad if it adds clarity.  :)

I think the ideal scenario is one where, if we were to add a new search parameter, we would only need to add a key to one place and the function wouldn't grow that much. Creating an array of strings that represents the keys would likely work with this. Does that make sense?

app/assets/javascripts/components/article_finder/article_finder.jsx
bwreid: I would add, I actually think that `.then(() => this.buildURL())` is probably unnecessary. I forgot you were building the URL on each change, so it should likely just work.

hjhimanshu01: I'm currently stuck on the dynamic part , everything else is done , would create a PR soon

app/assets/javascripts/components/article_finder/article_finder.jsx
hjhimanshu01: Made the changes! :)

app/assets/javascripts/components/article_finder/article_finder.jsx
bwreid: @hjhimanshu01 No worries! I think the first option will just fine. Good luck with exams!

hjhimanshu01: @bwreid , I just found out that to implement dynamic checking there is a small roadblock in the already implemented code , in the props that are passed to article_finder , "min_views" is stored as a string while "article_quality" as a number , I was stuck here , as I had been passing both as numbers which ideally should be the case as in dynamic checking as both are numbers.

The logic implemented is this :

`getParamsURL() {`
    `const query = qs.parse(window.location.search);`
    `const entries = Object.entries(query);`
   ` entries.map(([key, val]) => {`
     ` const copy_val = val ;`
      `val = (isNaN(parseInt(copy_val))) ? val : parseInt(val) ; `
     ` this.updateFields(key, val)`
   ` });`
   ` },`

hjhimanshu01: Update :

@bwreid @ragesoss  , encountered another bug that might have been there in the code for a while , details below :

On the Article Finder Page , if you click on the submit button without any search_term , it would fire up messages in console as well as a loading message that doesn't show anything.
Screenshot from 2019-03-08 14-48-53

ragesoss: Yes, it looks like that is the current behavior if you submit an empty query.

Feel free to add in a fix for that if it's very simple; alternatively, we could open a new issue for it.

hjhimanshu01: Let's have a new issue , I would take it up as this thread is going long , also I'm currently waiting for @bwreid for dynamic update checking issue where "article_quality" and "min_views" have a different data type , that is another issue if looked on or I could just keep it like the one which I had already made a commit with, making  a little changes for each one of the parameters.

bwreid: @hjhimanshu01 I would just go with the first option of only changing the value to a number for specific fields.

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -52,20 +55,40 @@ const ArticleFinder = createReactClass({
...
+    entries.map(([key, val]) => {
+      val = (key === 'article_quality') ? parseInt(val) : val;
+      return this.updateFields(key, val);
+    });
+  },
   updateFields(key, value) {
-    return this.props.updateFields(key, value);
+    const update_field = this.props.updateFields(key, value);
+    Promise.resolve(update_field).then(() => {
+    if (this.props.search_term.length !== 0) { this.buildURL(); }
```
bwreid: Let's fix this indentation. 👍

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -52,20 +55,40 @@ const ArticleFinder = createReactClass({
...
     });
   },
-
+  buildURL() {
+    let queryStringUrl = window.location.href.split('?')[0];
+    const params_array = ['search_type', 'article_quality', 'min_views'];
+    queryStringUrl += `?search_term=${this.props.search_term}`;
+    params_array.map((param) => {
+      return queryStringUrl += `&${param}=${this.props[param]}`;
+    });
```
bwreid: In this case, you're really just using `.map()` as a `.forEach()`. You can change it to a `.forEach()` or we could use `.reduce()`. I think the following would work:

```js
queryStringUrl = params_array.reduce((url, param) => {
  return url + `&${param}=${this.props[param]}`;
}, queryStringUrl);
```

I don't really have a preference! In fact, I think @ragesoss might like the `.forEach()` better.  :)

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -52,20 +55,40 @@ const ArticleFinder = createReactClass({
...
   },
 
   toggleFilter() {
     return this.setState({
       showFilters: !this.state.showFilters
     });
   },
-
+  buildURL() {
+    let queryStringUrl = window.location.href.split('?')[0];
```
bwreid: Should we use `window.location.search` here as well?

app/assets/javascripts/components/article_finder/article_finder.jsx
ragesoss: Heh. `.reduce()` is fine here. But I probably would have used `.forEach()` instead if I'd written it myself, yeah.

hjhimanshu01: Let's go with .forEach() ! :)

app/assets/javascripts/components/article_finder/article_finder.jsx
hjhimanshu01: window.location.search gives results after the "?" , it's better to use window.location.href as it gives the full address , that's why didn't changed it.

Reference : https://webplatform.github.io/docs/apis/location/search/

app/assets/javascripts/components/article_finder/article_finder.jsx
hjhimanshu01: Okay!

app/assets/javascripts/components/article_finder/article_finder.jsx
bwreid: Ah, great!  :)

hjhimanshu01: Done! :D

PR #2556, opened 2019-03-04
Issue#2520

### LINKED ISSUE #2520: Change passcode detail to be a link
image

Most instructors don't need the passcode as much as they need the actual link to send out to students to enroll. When the details section is _not_ being edited, we should change the passcode from plain text to actually being the enrollment link. When editing, the regular fields should still show.

twisted-sres:
Fixes #2520 
screenshot from 2019-03-05 00-53-20
Hey, it looks like this, but I guess there is some problem with courseLinkParams, it is not able to fetch the home page address and hence it is displaying undefined instead of it.

### DIFF
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -173,7 +173,7 @@ const Details = createReactClass({
 
    if (!this.props.editable) {
    const enrollEquals = '?enroll=';
-   const url = window.location.origin + this.props.courseLinkParams + enrollEquals + this.props.course.passcode;
+   const url = window.location.origin + this.props.course.slug + enrollEquals + this.props.course.passcode;
    passcode = (
      <TextInput
        onChange={this.updateDetails}
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -171,6 +171,23 @@ const Details = createReactClass({
       );
     }
 
+   if (!this.props.editable) {
+   const enrollEquals = '?enroll=';
+   const url = window.location.origin + this.props.courseLinkParams + enrollEquals + this.props.course.passcode;
+   passcode = (
+     <TextInput
+       onChange={this.updateDetails}
+       value= <a href={url}>{url}</a>
+       value_key="passcode"
+       editable={this.props.editable}
+       type="text"
+       label={I18n.t('courses.passcode')}
+       placeholder={I18n.t('courses.passcode_none')}
+       required={!!this.props.course.passcode_required}
+     />
+      );
+    }
+
     let expectedStudents;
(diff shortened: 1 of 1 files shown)


ragesoss:
The base URL comes from the course slug. `courseLinkParams` is used in some places as the base URL, but for the `Details` component that is not included as either a prop (as in `CourseAlerts`) or as a function (as in `EnrollButton`. However, the slug itself is included as the `course_id` prop in `Details`.

twisted-sres:
Hey, now it looks like this, the link is breaking between the lines, should I increase the width of the details box, but that would cover most of the page then.
screenshot from 2019-03-05 02-09-30

ragesoss:
I think rather than showing the whole link (and labelling it Passcode), it could be with `Enrollment link` with the target being the URL, and ideally with a functional `copy to clipboard` icon button next to it. Then the user could either click on the URL and copy it from the browser, or copy it straight to the clipboard to paste it into a an email.

twisted-sres:
Do we have this copy to clipboard button anywhere else in this project?

ragesoss:
No, I don't think we do. You can find an example here of a change that involved adding a new icon: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/4455c50328bc03ad02a7b35fe1ea73e4857b8c1e

For sourcing the icon, you could either find a freely licensed one from commons.wikimedia.org or The Noun Project, or create a new one.

twisted-sres:
Hey, can I implement copy to clipboard functionality using react-copy-to-clipboard package?
But that would include changing gitignore files as well

twisted-sres:
I have installed the copy-to-clipboard npm package in the node_modules folder, how do I import it? Is there anything similar which I could look on?

twisted-sres:
screenshot from 2019-03-08 01-19-29
@ragesoss how about this? The link gets copied on clicking on the text area and a Copied message is displayed. This way we are showing the link also.

ragesoss:
It's moving in the right direction, but I don't like using an un-editable input field for this purpose.

I think, in this case, a cleaner UI is a higher priority than showing the full URL.

twisted-sres:
@ragesoss so basically for copying, we need to select first and then copy, and select function is only supported for textarea, so one approach can be having a textarea but hiding it and having a button which would trigger the textarea select and copy. Haven't seen copy without textarea. Can't think of any other approach.

ragesoss:
@twisted-sres I see there are a few examples of doing it without a text area, eg, https://css-tricks.com/native-browser-copy-clipboard/

twisted-sres:
Okay, thanks, will try implementing it, I had few queries and wanted to chat on direct message on slack, is it possible?

ragesoss:
Please re-open if you take this up again. Thanks! Your progress will likely be useful for whoever finishes this up.

Issue #2499, opened 2019-02-19
Change Subject field to a multi-select dropdown


bwreid:
As a user creating a course, I want to be able to set my subject via a multi-select dropdown as opposed to having to enter free form text. At the same time, I want the option to be able to define my own subject if applicable. In general, the goal is to make course creation easier on the user and standardize our own data on the backend.

We should use the [react-select](https://react-select.com/) library to create the multi-select and can pull the list of subjects from the [wizard.yml](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/config/wizard/researchwrite/wizard.yml#L191) file.

neozenweb:
Outreachy applicant here, waiting for an invite to contribute. I would like to take this up. Just forked and cloned the repo. Thanks.

bwreid:
@neozenweb Now that you have things set up, I think this would be great feature to work on.

neozenweb:
@bwreid  I have set up the dev on my localhost but cannot access course creator. The console shows a select statement for user=29 and the screen returns to the homepage. Here is the screenshot

![image](https://user-images.githubusercontent.com/37545009/53354866-c254ef80-38f5-11e9-8e2f-9ddd8cbe5873.png)

neozenweb:
I have a few questions for this one:
1. I see that there is a separate course_level_selector.jsx component for the select element for the level. So for the subject, should I create a similar separate component or just use multi_select_field.jsx from the common components. 
2. How should I use the yaml file to populate this select field? Any examples in the code that have used this feature?
Thanks for your time.

puthalalitha:
Hello I am an Outreachy student, and I would like to work on this issue. is it still open? I have my environment set up.

bwreid:
@puthalalitha I believe this is still open, yes.

puthalalitha:
I don't see a subject field in the current course/program creation flow. Below is the screen shot of that page. Am I doing something wrong? 
![Screen Shot 2019-10-16 at 9 30 23 PM](https://user-images.githubusercontent.com/43020183/66978557-41eb2680-f05f-11e9-8c70-bb36f963ceaf.png)

ragesoss:
To see that version of the CourseCreator interface, you need to switch your development environment to `wiki_education: 'true'` mode in `application.yml`.

puthalalitha:
Hi Sage Ross,

Thank you. Now I am able to see the CourseCreator interface. 

Thanks,
[...]

PR #2561, opened 2019-03-04
[WIP]Change-Subject-field-to-a-multi-select-dropdown-#2499

neozenweb:
Change-Subject-field-to-a-multi-select-dropdown-#2499

Instead of entering subject field as a text, the value can be chosen from a multi-select drop-down. Additionally "Other" values for subjects can be added in the "Other" text field.
image


bwreid:
@neozenweb Please let us know if you need some help solving this conflict!

neozenweb:
> @neozenweb Please let us know if you need some help solving this conflict!

Thanks @bwreid , I did not realize that there were conflicts because there was no red cross mark at the PR description. I have tried to put in the config.example.yml, and added yarn.lock to my gitignore. Do you think I should add/delete any files from the gitignore?

bwreid:
@neozenweb I think a lot of these issues are stemming from all the changes to the config files. I believe the only changes that should have needed to happen would be in the `app/` folder. It looks like maybe you were having some issues getting the project set up at all. What were the other changes necessary for?

neozenweb:
> @neozenweb I think a lot of these issues are stemming from all the changes to the config files. I believe the only changes that should have needed to happen would be in the `app/` folder. It looks like maybe you were having some issues getting the project set up at all. What were the other changes necessary for?

@bwreid Most or all of the changes except for the app/,  were for the setup. One of them for example in .eslintrc was not recognizing the <end of line> which is different for unix vs windows. Another build failure may be due to the yarn codeclimate-test-reporter dependency which was not included in my package.json as it might be a new addition.  Is there anyway to block out these in .gitignore or what can I do to solve these build errors?

bwreid:
The more sustainable solution is to fix these errors. At the moment, you could just have your PR be the relevant files for the change. That should help in getting the tests to pass and will make it easier to evaluate our code.

Essentially, any changes that are specific to your machine should not be included in this Pull Request.

bwreid:
@neozenweb If I were you, I would try starting a new branch and/or pull request. If you want to keep some of the changes you made to the build, you can do so. Just add the files that are relevant to this PR (e.g. `git add ./app`) and only commit those files. That way we can review your PR and you can continue working on getting the build setup correctly.

neozenweb:
@bwreid I think I will do that. Thanks.

neozenweb:
I have reinstalled the dev on Ubuntu instead of windows, so please delete this pull request. I will do a new PR. Thanks.

PR #2581, opened 2019-03-07
Update jeet require path through node_modules/

bwreid:
Not sure when this got changed, but it doesn't look like we're able to update our CSS without changing this first.

### DIFF
--- a/app/assets/stylesheets/_shared.styl
+++ b/app/assets/stylesheets/_shared.styl
@@ -1,4 +1,4 @@
 @import "_variables"
 @import "_easing"
 @import "_utils"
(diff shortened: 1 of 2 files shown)


ragesoss:
It got changed here, recently: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/03d273da64132fc3c980c009d3512ed126c76be8#diff-f454a89f933fec6b5f4eb30318bffc13

ragesoss:
This doesn't work locally for me, or on Travis. Maybe different path lookup behavior on your system?

bwreid:
Interesting. Alright, I'll check it out!

Issue #2583, opened 2019-03-07
In Articles tab, navigation to Assigned Articles and Available Articles sections is hard


ragesoss:
This is what https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/1950 tries to address, although there could be other possible solutions as well.

bwreid:
What about just the following? I changed the style of the button to feel somewhat distinct from the "Get Help" button right above it. Clicking on each button would show the selected one. This also should narrow the screen any further.

![image](https://user-images.githubusercontent.com/1316902/54061296-1644bb80-41b5-11e9-816c-0ecf3e677aed.png)

For fanciness, I can try and add some transition effects.

ragesoss:
works for me!

PR #2594, opened 2019-03-08
Add navigation for article tab

bwreid:
Issue #2583 
- Swap between Edited, Assigned, and Available articles
- Some light reorganization of the associated files

article-switch

### DIFF
--- a/app/assets/javascripts/components/articles/article_list.jsx
+++ b/app/assets/javascripts/components/articles/article_list.jsx
@@ -7,6 +7,7 @@ import { connect } from 'react-redux';
 import * as ArticleActions from '../../actions/article_actions';
 import List from '../common/list.jsx';
 import Article from './article.jsx';
+import CourseOresPlot from './course_ores_plot.jsx';
 import CourseUtils from '../../utils/course_utils.js';
 
 const ArticleList = createReactClass({
@@ -18,7 +19,7 @@ const ArticleList = createReactClass({
     current_user: PropTypes.object,
     actions: PropTypes.object,
     articleDetails: PropTypes.object,
-    sortBy: PropTypes.func,
+    sortArticles: PropTypes.func,
     wikidataLabels: PropTypes.object,
     sort: PropTypes.object
   },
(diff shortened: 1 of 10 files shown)


ragesoss:
Cool. Considerations:
* Article finder still works?
* What happens when user is a student and there are no available articles? (We currently hide that table completely for students if it's empty, but show it for instructors so they can add to it.)

ragesoss:
I guess we should hide the nav button in the cases where we're also hiding the table.

no available

Article finder works fine. (I was thinking, perhaps, about some UI ideas we were kicking around that would have the rendering of ArticleFinder interact more with the Articles tab, like as a fourth nav option in addition to these three.)

ragesoss:
Hmm... and this should probably use the URL to reflect state, so that a user could link directly to the Available or Assigned views.

bwreid:
Yep, Article Finder works. I didn't know that about Available Articles -- I'll hide that button as well if the table is hidden.

And happy to add url params to this! Was actually thinking about that as well after pushing it up. :)

bwreid:
I thought the following might be more clear:

image

There is a weird side effect of updating the search string on this page. When *Articles Assigned* is clicked, it correctly changes the URL. But, when you click elsewhere on the page, it's returned to just `/articles`. I believe this is due to [these lines in the article viewer.](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/components/common/article_viewer.jsx#L113-L115) I can inspect a bit more tomorrow to see if there's a reasonable way to get around this. It doesn't really affect the feature though.

bwreid:
One thing I forgot is permissions to see the table if you're an instructor or admin. Will continue on this tomorrow.

ragesoss:
If we're going with this sub-nav approach where we only render one list at a time, the best solution to that URL issue is probably to use routes instead of params, and extract `assignments` and `available_articles` into their own components that are independent of `articles`.

Just thought of one other concern: the UI for selecting an available article needs a bit of enhancement. Currently, selecting an available artifcle essentially just moves it from `available` to `assigned`, which is good enough when both tables are visible at once, but may be confusing now since the newly assigned article you select won't be visible, and the one you click will simply disappear. A confirmation message should probably be added.

bwreid:
@ragesoss There is already a notification that appears when someone enrolled in the course selects an article from the list of Available articles.

image

Is there another place you were thinking of?

ragesoss:
Nope, I forgot that was there.

ragesoss:
I think this is good to go once the tests are passing, but it looks like a few need to be updated.

PR #2596, opened 2019-03-09
refactoring-controller

hardy1334:
For the issue #1949 ,I have divided UserController is further divided into **EnrollCoursesController** and in that controller action regarding enrolling a program is there.


hardy1334:

ragesoss:
It looks like this is still duplicating, rather than renaming, the views that the new controller needs.

hardy1334:
@ragesoss In the view of  **enroll_courses** there seems that code of  **users** view is used again for the **enroll_courses** but without code will break and test cases wont pass

ragesoss:
@hardy1334 in that case, you can override the template used to render enroll_courses, so that you don't need to duplicate it: https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-an-action-s-template-from-another-controller

hardy1334:
@ragesoss  I have tried implementing, renderign **users** view in the **enroll_courses_controller.rb** but that fails  many test cases in **EnrollCoursesController**, I'm not quite sure why this happened, if we go by the current methodology,it works completly fine.

hardy1334:
I guess what we can do is merge this one and open new **issue** to remove the duplicate view in **enroll_courses** , as current version of it works completly fine.

ragesoss:
Since this is a refactor, I don't want to merge something that introduces new issues that require subsequent cleanup.

It's important that those two views are the same, because on the frontend, we use the same data handling in redux for both receiving the list of users and updating that list after a user is added or removed from a course... so if the duplicated views were to diverge, it could introduce subtle bugs.

hardy1334:
@ragesoss  okay,no problem let me see through it

ragesoss:
Feel free to reopen if you take this up again.

PR #2597, opened 2019-03-10
refactor: jQuery-uls removed from bower

### LINKED ISSUE #2312: Move jquery-uls to the webpack config

bhawna-singh:

## What this PR does
closes #2312


ragesoss:
It looks like the rspec tests are passing, but there's a problem with how jquery gets provided in the jest test helper.

ragesoss:
Feel free to reopen if you return to this.

PR #2603, opened 2019-03-11
[WIP] Add option for image link and remove.

### LINKED ISSUE #1924: Allow profile pics from Commons
Currently, we store the user profile pic locally. It would be nice to allow a user to choose an existing photo from Wikimedia Commons as an alternative.

dustmop44:
## What this PR does
Added an option for an image link for profile picture and also added a remove button to remove the profile picture.

fixes #1924


bwreid:
@dustmop44 Thanks for the PR! Please check the failing build. It looks as though you have some errors stemming from `rubocop` to start. I think it would also be worthwhile to update and or add some tests for this feature.

dustmop44:
Ok. Will get on it. Just wanted to see if I was doing everything ok. Thanks!

dustmop44:
I updated the formatting errors. As for the tests, I seem out of my element for the test specs. What tools are we using there? Is there some documentation I can read?

ragesoss:
@dustmop44 here's the docs for our test setup: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/testing.md

You can find many examples of feature specs in the `spec/features` directory; that sort of browser-based test is probably what makes sense for this feature.

dustmop44:
Oh, that makes sense! I forgot it was in the spec folder. Was looking at the test folder was completely confused. Will get on it now. Thanks!

dustmop44:
Done with the tests. Let me know if they were adequate or not. And thank you for letting me work on this issue!

bwreid:
This looks really great, @dustmop44!

My only comment is that I think it'd make more sense for the route to be `DELETE /profile_image` as opposed to a `GET`. While we can do either, I would say `DELETE` makes more sense here. What do you think?

dustmop44:
Sounds about right! Fixed the route and the tests. I thought the RESTful routes should be used only on models. But after further research it can be used on anything, and it is just a naming convention to organize our routes! Thanks!

bwreid:
@dustmop44 I think you'll still need to change the front end so that when you click on the link to remove the profile image, it sends a DELETE (as opposed to a GET, which is the default behavior for a link).

dustmop44:
So sorry! Would this be about right?

bwreid:
Looking good @dustmop44! Thanks for working on this!

dustmop44:
Thank you for helping me through it!

Issue #2613, opened 2019-03-12
Upgrade tinymce and @tinymce-react to latest versions


ragesoss:
It looks like tinymce version 5 has breaking changes — maybe just a change in location for the theme, or maybe more. See the gulp build errors here: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2607

Find and fix the cause of these errors with the new tinymce release, and manually test the timeline block editing feature where tinymce is used to check for differences in the user experience.

hardy1334:
I'll like to work on it

PR #2620, opened 2019-03-13
tinymce-react and tinymce updated to Latest versions

hardy1334:
For the issue #2613   In this **Tinymce-react** and **tinymce** have been updated to their latest versions of **3.0.1** and  **5.0.2** respectively


hardy1334:
@ragesoss  can you please have a look at it

ragesoss:
It looks like this changes the theme. Can you add before/after screenshots, along with some explanation of the change (as in, was it just a renaming, or did they replace one theme with another, or what)?

hardy1334:
**Before:** (Version _4.9.2)_

Screenshot from 2019-03-14 14-46-31

**After:** (Version _5.0.2)_

Screenshot from 2019-03-14 11-48-53

There are some changes in UI in version 5.0.2.

ragesoss:
Hmm... I notice that the link icon and the list icons are no longer there in the version. Those icons are the result of the the `link` and `lists` plugins, which may not be working. We also use the `code` plugin to provide source HTML, so you should figure out if that is no longer working as well.

hardy1334:
Current Version:

Screenshot from 2019-03-17 05-05-33

bwreid:
@hardy1334 This looks pretty great to me! Thanks for working on this.

ragesoss:
@hardy1334 did you look into the `code` plugin? It looks like that got removed in the PR.

ragesoss:
Looks like it works with just adding it back in: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/160867af338995544b8b68318a0790723dd77251

hardy1334:
@ragesoss  shall I add code plugin

ragesoss:
I did already.

ragesoss:
It adds it to the menu, which is how it was before. I think it doesn't need to be in the toolbar also.

PR #2631, opened 2019-03-15
Improve revision import behavior when there are unimported revisions before the first imported ones

ragesoss:
RevisionImporter starts from the last already-imported revisions for each user, and only imports later revisions.

In cases where the start date of a course has changed, this could mean that early revision that are now within the course dates never get imported.

This modifies the behavior of `/manual_update` and the `needs_update` flag so that updates from either of those user-initiated update modes will (re-)fetch all revisions, not just ones later than the already-known revisions. It also automatically sets the `needs_update` flag for a course whenever the start date gets changed, so that updating the stats in such cases will happen automatically.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -150,7 +150,7 @@ class CoursesController < ApplicationController
 
   def manual_update
     @course = find_course_by_slug(params[:id])
-    UpdateCourseStats.new(@course) if user_signed_in?
+    UpdateCourseStats.new(@course, full: true) if user_signed_in?
     redirect_to "/courses/#{@course.slug}"
   end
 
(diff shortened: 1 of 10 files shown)


ragesoss:
Don't know how common it is, but I do it fairly often to simplify the interface, for classes where there's one entry point for other code to call it and there's no need for the other code to know or care about whether or not there are instances of the class involved. All else being equal, I'd usually rather invoke `CourseRevisionUpdater.import_revisions(foo)` than `CourseRevisionUpdater.new(foo).import_revisions`. But most cases where it's used, it's more for state management: the entry point may handle a bunch of records, while each instance of the class is concerned with a single record.

In this case, it's a remnant from the previous version of this code that was designed to processes a whole set of courses (either one at a time or concurrently). In those cases, the class instance was designed around having a single course as the relevant state, so each different course would need its own instance. Now that I removed the code for processing multiple courses at once, it doesn't make much of a difference.

bwreid:
Alright, makes sense. Let's merge it and perhaps can revisit at some point. 👍

PR #2641, opened 2019-03-17
conditionally rendered link issue#2546

hjhimanshu01:
## What this PR does
Conditionally renders link in ArticleViewer , issue#2546

### DIFF
--- a/app/assets/javascripts/components/common/article_viewer.jsx
+++ b/app/assets/javascripts/components/common/article_viewer.jsx
@@ -328,7 +328,7 @@ const ArticleViewer = createReactClass({
           <div className="article-header">
             <p>
               <span className="article-viewer-title">{trunc(this.props.article.title, 56)}</span>
-              {!this.props.article.id && (
+              {this.props.article.id && (
                 <span><a className="icon-link" href={`?showArticle=${this.props.article.id}`} /></span>
               ) }
               {closeButton}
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/common/article_viewer.jsx
```diff
@@ -328,7 +328,9 @@ const ArticleViewer = createReactClass({
           <div className="article-header">
             <p>
               <span className="article-viewer-title">{trunc(this.props.article.title, 56)}</span>
-              <span><a className="icon-link" href={`?showArticle=${this.props.article.id}`} /></span>
+              {this.props.article.id === 'undefined' && (
```
bwreid: I don't think this would/should be able to `=== 'undefined'` as triple equals is used for strict equality. If there is no id, the value should be `undefined`, not a _string_ with the contents `'undefined'`.

A perhaps simpler way to do this is to just use `!this.props.article.id`.

hjhimanshu01: Ah! My bad ! , that's a grave mistake on my part .

bwreid: Did you try this out? I think we were both wrong here. Instead, it should be:
```js
this.props.article.id && ( ... )
```

hjhimanshu01: Absolutely right !, had tested it with 'undefined' keyword , made changes

PR #2652, opened 2019-03-19
Issue 1611

davidnoahball:
## What this PR does
This addressed the feature request from issue 1611. It adds a button to the Timeline page to edit week titles, and a button to reset to default titles. I also moved dates to be below the titles to accommodate the change.

## Screenshots
Before:
Screenshot from 2019-03-18 19-50-23

After:
Screenshot from 2019-03-18 19-49-49

### DIFF
--- a/app/assets/javascripts/components/timeline/timeline.jsx
+++ b/app/assets/javascripts/components/timeline/timeline.jsx
@@ -65,7 +65,7 @@ const Timeline = createReactClass({
   },
 
   usingCustomTitles() {
-    return this.props.weeks.find(week => week.title) !== undefined;
+    return this.props.weeks.some(week => week.title);
   },
 
   hasTimeline() {
@@ -316,14 +316,14 @@ const Timeline = createReactClass({
         {saveChangesButton}
         {cancelChangesButton}
       </div>
-    ) : undefined;
+    ) : null;
     const titlesActionButtons = this.props.editableTitles ? (
       <div>
         {saveChangesButton}
         {cancelChangesButton}
         {resetTitlesButton}
       </div>
-    ) : undefined;
+    ) : null;
 
     let reorderableControls;
     let editWeekTitles;
--- a/app/assets/javascripts/components/timeline/timeline.jsx
+++ b/app/assets/javascripts/components/timeline/timeline.jsx
@@ -35,12 +35,16 @@ const Timeline = createReactClass({
     deleteAllWeeks: PropTypes.func.isRequired,
     saveGlobalChanges: PropTypes.func,
     cancelGlobalChanges: PropTypes.func,
+    updateTitle: PropTypes.func,
+    resetTitles: PropTypes.func,
     saveBlockChanges: PropTypes.func,
     cancelBlockEditable: PropTypes.func,
     all_training_modules: PropTypes.array,
     edit_permissions: PropTypes.bool,
     reorderable: PropTypes.bool,
-    enableReorderable: PropTypes.func
+    editableTitles: PropTypes.bool,
+    enableReorderable: PropTypes.func,
+    enableEditTitles: PropTypes.func
   },
 
   getInitialState() {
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -60,6 +64,10 @@ const Timeline = createReactClass({
     return week.blocks;
   },
 
+  usingCustomTitles() {
+    return this.props.weeks.find(week => week.title) !== undefined;
```
bwreid: So, this line is trying to determine whether or not _anything_ has a title, correct? If that's the case, the following might provide a more clear intention:
```js
return this.props.weeks.some(week => week.title)
```
Alternatively, you could just return a boolean value with your `.find()` result instead of having to compare to `undefined`:
```js
return !!this.props.weeks.find(week => week.title)
```

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -280,18 +296,37 @@ const Timeline = createReactClass({
...
-        <button className="button dark button--block" onClick={this.props.saveGlobalChanges}>
-          {I18n.t('timeline.save_all_changes')}
-        </button>
-        <button className="button button--clear button--block" onClick={this.props.cancelGlobalChanges}>
-          {I18n.t('timeline.discard_all_changes')}
-        </button>
+        {saveChangesButton}
+        {cancelChangesButton}
+      </div>
+    ) : undefined;
```
bwreid: This is minor, but for an intentional "nothing" value, I would typically use `null`.

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -280,18 +296,37 @@ const Timeline = createReactClass({
...
+    const titlesActionButtons = this.props.editableTitles ? (
+      <div>
+        {saveChangesButton}
+        {cancelChangesButton}
+        {resetTitlesButton}
       </div>
     ) : undefined;
```
bwreid: Same goes for here.

app/assets/javascripts/reducers/timeline.js
```diff
@@ -39,6 +41,13 @@ const blocksInWeek = (blocks, weekId) => {
   return count;
 };
 
+const validateTitle = (title) => {
+  if ((typeof title === 'string' || title instanceof String) && title.length <= 20) {
+    return true;
+  }
+  return false;
+};
```
bwreid: You can just shorten this to:
```js
return (typeof title === 'string' || title instanceof String) && title.length <= 20)
```

What's the case where this would not be a string?

app/assets/stylesheets/modules/_timeline.styl
```diff
@@ -191,7 +191,7 @@ button.week__add-block
 
 .week__week-dates
   position relative
-  top 16px
+  top -12px
```
bwreid: If this is going to appear right below the title, I think it would be better to rely on more standard techniques like `margin`. We probably don't need to reposition this anymore.

davidnoahball:
@bwreid Thanks for the review. I've addressed your comments and fixed the bug. I think it should be good to go now.

bwreid:
@davidnoahball Thanks so much! I think this looks great.

I would like to wait until @ragesoss gets back so he can also weigh in on the UI changes. As soon as he does so, we'll merge it. Great work!

app/assets/javascripts/components/timeline/week.jsx
```diff
@@ -68,21 +71,35 @@ const Week = createReactClass({
...
+    } else {
+      weekTitleContent = I18n.t('timeline.week_number', { number: this.weekNumber() });
+    }
+    const weekId = this.props.week.id;
+    const weekTitle = this.props.editableTitles ? (
+      <input
+        className="week-index week-title-input"
+        defaultValue={weekTitleContent}
+        maxLength={20}
+        onChange={text => this.props.updateTitle(weekId, event.target.value, text)}
```
ragesoss: This is giving me an error. Should be `event => this.props.updateTitle(weekId, event.target.value)`?

ragesoss: This error just happens in Firefox, it looks like.

ragesoss:
The Milestones thing doesn't need to be done in this PR; there's already a separate issue for it. This should be good to merge once the `onChange` is fixed and the button label is updated.

davidnoahball:
@ragesoss `onChange` fixed, button label updated

davidnoahball:
Forgot about updating the button in the test; fixed now.

Issue #2635, opened 2019-03-15
Send DYK alerts to instructors


bwreid:
### Current Behavior:
At the moment, DYK alerts only go to the content experts associated with a course.

### Desired Behavior:
In addition to content experts, send an email to the creator of the course/program.

### Related Files & Code:

* [dyk_nomination_monitor.rb](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/lib/alerts/dyk_nomination_monitor.rb)
* [dyk_nomination_monitor_spec.rb](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/spec/lib/alerts/dyk_nomination_monitor_spec.rb)

Aeropio:
hello @bwreid : I would like to work on this issue, could you please assign this to me?

Aeropio:
> ### Current Behavior:
> At the moment, DYK alerts only go to the content experts associated with a course.
> 
> ### Desired Behavior:
> In addition to content experts, send an email to the creator of the course/program.
> @bwreid Is creator of the course same as instructor?
> ### Related Files & Code:
> * [dyk_nomination_monitor.rb](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/lib/alerts/dyk_nomination_monitor.rb)
> * [dyk_nomination_monitor_spec.rb](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/spec/lib/alerts/dyk_nomination_monitor_spec.rb)

bwreid:
Hi @Aeropio -- this would be a fine issue to work on, but I would first complete the other issue you started on. Unless you would like to skip that one and work on this instead?

arpitadash:
Hello @bwreid, I'd like to work on this issue if it is still open.

bwreid:
@arpitadash This commit was responded to relatively recently. If we don't hear from @Aeropio in the next day or so, please feel free to take it.

Aeropio:
> Hi @Aeropio -- this would be a fine issue to work on, but I would first complete the other issue you started on. Unless you would like to skip that one and work on this instead?

Hi @bwreid I would like to skip the other issue and take this issue instead

Aeropio:
Hello @bwreid : I am almost done with the implementation of this issue, but I am not sure how to test it, so can you please provide steps to test this?

Aeropio:
Please review this PR
[...]

PR #2659, opened 2019-03-20
sending DYK alerts to instructors

Aeropio:

## What this PR does
It addresses isssue number 2635, DYK alerts now being send to creator of the course/program inaddition to content experts associated with a course.

## Screenshots
Before:

After:
How can i test if DYK alerts are being send to both instructors and content experts in my local machine

### DIFF
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -118,6 +118,17 @@ class Alert < ApplicationRecord
     update_attribute(:email_sent_at, Time.zone.now)
   end
 
+  def email_course_instructor
+    return if emails_disabled?
+    return if course.nil?
+    course_creators = course.instructors
+    return if course_creators.empty?
+    course_creators.each do |course_creator|
+      AlertMailer.alert(self, course_creator).deliver_now
+    end
+    update_attribute(:email_sent_at, Time.zone.now)
+  end
+
   # Disable emails for specific alert types in application.yml, like so:
   #   ProductiveCourseAlert_email_disabled: 'true'
   def emails_disabled?
--- a/package.json
+++ b/package.json
@@ -166,7 +166,7 @@
   },
   "devDependencies": {
     "webpack-cli": "^3.3.0",
-    "@babel/plugin-transform-runtime": "^7.3.4"
+    "@babel/plugin-transform-runtime": "^7.4.0"
   },
   "resolutions": {
     "babel-core": "7.0.0-bridge.0"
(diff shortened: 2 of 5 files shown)


bwreid:
@Aeropio Thanks for the PR. In terms of adding a test, I would look at the `dyk_nominiation_monitor_spec.rb` file. This has a number of tests that should help point you in the right direction.

Aeropio:
@bwreid ok sure i will look into it

bwreid:
@Aeropio It looks like you're encountering a syntax error. I would try running the tests locally so you can more quickly troubleshoot it.

Additionally, we need to add another request as part of this feature change. I'll comment on issue #2635 directly.

Aeropio:
@bwreid I will check for the syntax error and update you

package.json
```diff
@@ -166,7 +166,7 @@
   },
   "devDependencies": {
     "webpack-cli": "^3.3.0",
-    "@babel/plugin-transform-runtime": "^7.3.4"
+    "@babel/plugin-transform-runtime": "^7.4.0"
```
bwreid: I don't think this should be needed as part of this Pull Request

app/models/alert.rb
```diff
@@ -118,6 +118,17 @@ def email_target_user
     update_attribute(:email_sent_at, Time.zone.now)
   end
 
+  def email_course_instructor
+    return if emails_disabled?
+    return if course.nil?
```
bwreid: What about combining this to:
```js
return if emails_disabled? || course.nil?
```

spec/lib/alerts/dyk_nomination_monitor_spec.rb
```diff
@@ -57,6 +62,13 @@ def mock_mailer
       expect(Alert.last.email_sent_at).not_to be_nil
     end
 
+    it 'emails course creator' do
+      create(:courses_user, user_id: course_creator.id, course_id: course.id, role: 1)
+      allow_any_instance_of(AlertMailer).to receive(:alert).and_return(mock_mailer)
+      described_class.create_alerts_for_course_articles
+      expect(Alert.last.email_sent_at).not_to be_nil
```
bwreid: A better check here might be to make sure that the person to whom the email was sent matches up with the `course_creator`. Right now, it looks like you're just checking to make sure an email was sent. I imagine you could either take a look at other properties on the `Alert.last` object or expect `AlertMailer.alert` method to receive the `course_creator`.

PR #2668, opened 2019-03-20
[WIP] Ticketing system

bwreid:
In progress


bwreid:
@ragesoss Would love to get your thoughts on this last commit.

I set it up so that we can receive emails but there are a couple logistical challenges of matching everything up:

1. If we just receive an email, we won't necessarily know what user it is from given that we don't have to store emails.
1. That means we likely won't know what class it's associated with.
1. We need some way to tie a forwarding email address (e.g. "wreid@wikiedu.org") with a User account on the dashboard. It might not be the case that these things match up.

What I did to start this process was to build a new SpecialUser type that is a list of Wikipedia Experts. Maybe we can make that array more complex than just a list of usernames and instead have, for example, a hash that maps forwarding emails to user accounts. Can you think of another way to tie these things together?

PR #2670, opened 2019-03-21
[WIP]Change-Subject-field-to-a-multi-select-dropdown-#2499

neozenweb:
This PR adds a multiple select to the "Subject" field on the course creator form, replacing the text input.
multi-select-subjects


neozenweb:
@bwreid , One of the Travis build seems to fail. How should I fix this problem? Thanks.

bwreid:
@neozenweb These errors are a bit odd. I'm going to re-run the build and we can see if anything changes. In the meantime, are you able to get the tests running successfully locally?

It still looks like this PR includes unnecessary file changes. The PR should only include the files necessary to solve this feature. There are a number of changes here that don't relate to the overall feature. You'll need to remove these from the PR first before we are able to merge.

neozenweb:
I rebased the code to master branch and just changed the one
file app/assets/javascripts/components/course_creator/course_form.jsx  so I
don't know where those changes are coming from. I also have run the jest
tests and the local testing successfully. Thanks.

PR #2671, opened 2019-03-22
multi-select#2499

neozenweb:
The subject field was a text input field for course create form. This is changed to a multiple select field. 
multi-select-subjects

### DIFF
--- a/app/assets/javascripts/components/course_creator/course_form.jsx
+++ b/app/assets/javascripts/components/course_creator/course_form.jsx
@@ -1,16 +1,15 @@
 import React from 'react';
+import CreatableSelect from 'react-select/lib/Creatable';
 import TextAreaInput from '../common/text_area_input.jsx';
 import CreatableInput from '../common/creatable_input.jsx';
 import TextInput from '../common/text_input.jsx';
 import DatePicker from '../common/date_picker.jsx';
-import MultiSelectField from '../common/multi_select_field.jsx';
 import CourseLevelSelector from './course_level_selector.jsx';
 
 const CourseForm = ({ courseFormClass, course_utils, string_prefix, updateCourseAction, course, showEventDates, showEventDatesState, updateCoursePrivacy, showTimeValues, updateCourseDateAction, courseDateUtils, eventClass, roleDescription, defaultCourse }) => {
   const dateProps = courseDateUtils.dateProps(course);
   let term;
   let subject;
-  let newsubject;
   let expectedStudents;
   let courseLevel;
   const subjects = [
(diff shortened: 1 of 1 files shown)


ragesoss:
This looks right from the screenshot, but it needs to use the full list of disciplines from the wizard yml file.

The easiest way to fix the merge conflict might be to start a new branch and copy over the changes, as we just merged some significant refactoring of CourseCreator.

neozenweb:
@ragesoss  @bwreid  For parsing the .yml file, can I use this https://www.npmjs.com/package/read-yaml      If ok, then should it be included in the package.json too? Thanks.

neozenweb:
I tried getting the data from the courses.rb using the 118n.t(courses.subject), but there is no data. It just shows the string "Subject".

bwreid:
This seems ok to me. @ragesoss, do you have another idea on how @neozenweb could pull a dynamic list of subjects?

ragesoss:
@bwreid  @neozenweb I've been thinking about it, and nothing appropriately simple comes to mind. **The best option I can think of is duplicating that list manually, with comments in the .yml file and in the frontend component noting that they are duplicated and will need to be updated in sync.**

Here are the ways I could think of to do it dynamically:
* parsing it server-side and sending it to the client as needed (either embedding it in the HTML for /course_creator or fetching from a JSON endpoint, like we do with the rest of the wizard data already);
* significantly refactoring how wizard data is delivered, so that data for all possible wizard choices is fetched at the beginning of course creator, and that list gets pulled out of the data and parsed client-side;
* parsing it at build time with a custom `gulp` task or similar, to generate a javascript file with the list of subjects.

The first and third introduce enough extra complexity to the data flow that I don't think it's worth what we gain in DRY-ness.

The second one is the way that seems like the best design, but it would involve an enormous refactor so that course_creator and the wizard are not totally separate things, and React maintains state through the transition from course_creator to wizard. Currently, we do a redirect and then load the page from scratch when the course creator is submitted, to transition from the creator (during which there is no server-side course record) to the wizard (which starts from an already-created course). That's an awkward bit of architecture already. It would be nice to do at some point, but a wildly disproportionate amount of work for the the task at hand.

neozenweb:
@ragesoss @bwreid  Thanks for the inputs and thoughts about populating the data.

 I have been going through some links provided by @bwreid ,  about the model-view-controller aspects of the application. I came across "all_courses_csv" in config/routes.rb . This gives a csv with all courses data(unfortunately no 'subject'). I think this route corresponds to app/presenters/courses_presenter.rb . With my limited knowledge about the application, just wanted to get your views if these can somehow be used to populate the subject field. Thanks.

bwreid:
@neozenweb Given what Sage said above, it sounds like his thought is just to duplicate the subjects. Clever sleuthing, though!

Thanks for your patience on this. Let's add the subjects manually, fix the merge conflict, and then we can get to adding this into the code base.  :)

ragesoss:

PR #2689, opened 2019-03-27
Search suggestions added#649 WIP

hjhimanshu01:
## What this PR does
Creates a search suggestion using MediaWiki API calls

## Questions
While making API calls , data retrieval takes some seconds and fetching that data while updating react state causes react to store previous data retrieved while fetching , that creates inconsistency sometimes

## Pending Work
Flow of the code of article_finder is becoming hard to get hold off , I'm thinking to document it ?

### DIFF
--- a/app/assets/javascripts/components/students/assign_button.jsx
+++ b/app/assets/javascripts/components/students/assign_button.jsx
@@ -41,7 +41,8 @@ const AssignButton = createReactClass({
       showOptions: false,
       language: this.props.course.home_wiki.language,
       project: this.props.course.home_wiki.project,
-      title: ''
+      title: '',
+      data: false
     });
   },
 
@@ -57,7 +58,6 @@ const AssignButton = createReactClass({
     return e.stopPropagation();
   },
 
-
   handleShowOptions(e) {
     e.preventDefault();
     return this.setState(
@@ -70,12 +70,35 @@ const AssignButton = createReactClass({
     const assignment = CourseUtils.articleFromTitleInput(title);
     const language = assignment.language ? assignment.language : this.state.language;
     const project = assignment.project ? assignment.project : this.state.project;
+    this.fetchSuggestionsForArticle(title);
     return this.setState({
       title: assignment.title,
       project,
       language
     });
   },
+
+  fetchSuggestionsForArticle(search_term) {
+    const URL_to_fetch = `https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&utf8=1&srsearch=${encodeURIComponent(search_term)}`;
+    $.ajax({
+      dataType: 'jsonp',
+      url: URL_to_fetch,
+      success: (data) => {
+        this.setState({
+          data: data,
+        });
+      },
+      error: (jqXHR, exception) => this.showException(jqXHR, exception)
+    });
+  },
+  TitleChangeSuggestionInput(e) {
+    this.setState({
+      title: e.target.value,
+      data: false
+    }
+    );
+  },
+
   handleChangeLanguage(val) {
     return this.setState(
       { language: val.value });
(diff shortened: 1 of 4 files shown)


ragesoss:
It looks like this is adding the search suggestions to ArticleFinder, rather than to AssignButton. The latter is where this will be particularly useful. I'm not sure it makes sense in ArticleFinder, where we care less about finding an exact title match and more about finding all articles that match a keyword search that may or may not be one of the suggestions that comes back from search autocomplete API.

hjhimanshu01:
@ragesoss , Yes , before starting the issue , @bwreid and I had a talk on slack from where to start this issue and reached on a conclusion that it would be safe to start from ArticleFinder , though the same functionality could be added directly to assignButton , if that's the case as I had planned accordingly , would stash changes from articleFinder and make those in assignButton

ragesoss:
Cool. If you got the basic concept working, it should be easy to port over to different input fields. It's the kind of thing that might be relevant in multiple places, so the more you can isolate the main functionality from the specific use case, the better.

hjhimanshu01:
@ragesoss , couldn't completely isolate it but used the code from which I had already written in assignButton , almost the same , though i'm not clear as to where to put stylings , currently in _article_finder.styl .

ragesoss:
Definitely, `_article_finder.styl` is not the place for it. `_inputs.styl` maybe?

hjhimanshu01:
Makes sense! will do that! :)

app/assets/javascripts/components/students/assign_button.jsx
```diff
@@ -70,12 +70,35 @@ const AssignButton = createReactClass({
...
     return this.setState({
       title: assignment.title,
       project,
       language
     });
   },
+
+  fetchSuggestionsForArticle(search_term) {
+    const URL_to_fetch = `https://en.wikipedia.org/w/api.php?action=query&format=json&list=search&utf8=1&srsearch=${encodeURIComponent(search_term)}`;
+    $.ajax({
```
bwreid: It might be nice to separate this API call out to be in a different file. Additionally, can we use `cross-fetch` instead of `$.ajax`?

app/assets/javascripts/components/students/assign_button.jsx
```diff
@@ -70,12 +70,35 @@ const AssignButton = createReactClass({
...
+      url: URL_to_fetch,
+      success: (data) => {
+        this.setState({
+          data: data,
+        });
+      },
+      error: (jqXHR, exception) => this.showException(jqXHR, exception)
+    });
+  },
+  TitleChangeSuggestionInput(e) {
```
bwreid: Don't think this needs a capital letter to start it off.  :)

app/assets/javascripts/components/students/assign_button.jsx
hjhimanshu01: I had tried using cross-fetch earlier but it had issue due to CORS policy , I wasn't sure how to solve it , tried stack overflow and other stuff , changed headers but still wasn't able to get the desired data , so did it with $.ajax , any ideas on how to solve that?

shivanshbindal9: Try using `https://en.wikipedia.org/w/api.php?origin=*&action=query&format=json&list=search&utf8=1&srsearch=${encodeURIComponent(search_term)}`
I think this might work

hjhimanshu01: That "origin=*" basically is allowing requests from all origins? Am I right?

shivanshbindal9: Yeah! 
For more details, you can go through https://www.mediawiki.org/wiki/API:Cross-site_requests

hjhimanshu01: Thanks @shivanshbindal9 :) will try it !

ragesoss:

hjhimanshu01:
@ragesoss , sure would do it , just a little occupied right now ! :)

PR #2690, opened 2019-03-28
.ajax converted to fetch in course_actions.js 

hardy1334:
In this **.ajax** implementation has been converted to **fetch**  implementation as per issue #2015 in the **course_actions.js**

### DIFF
--- a/app/assets/javascripts/actions/course_actions.js
+++ b/app/assets/javascripts/actions/course_actions.js
@@ -77,10 +77,10 @@ export const updateClonedCourse = (course, courseSlug, newSlug) => (dispatch) =>
 
 const needsUpdatePromise = (courseSlug) => {
   return API.fetch(courseSlug, 'needs_update')
-  .then(data => {
+  .then((data) => {
       return data;
   })
-  .catch(err => {
+  .catch((err) => {
       return err;
   });
 };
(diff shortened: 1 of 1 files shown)


hardy1334:

ragesoss:
@hardy1334 have a look at extra details I just added to the issue description.

hardy1334:
@bwreid  @ragesoss  can u have a look at it

ragesoss:
It looks like the credentials inclusion (which I added details about in the issue description) is still not included.

hardy1334:
@ragesoss  I guess its there in **fetch** function in **api.js** file, as I'm calling fetch function in **course_actions.js**

ragesoss:
I get an error when I use the 'schedule data update' button with this change. (It looks like this particular feature lacks any client-side test, so it got through the test suite anyway.)

hardy1334:
okay,can you guide me a bit how to handle that

ragesoss:
You should be able to recreate the error by going to a course that has a campaign and is already ended, and then clicking 'schedule data update'. The resulting error in the console should give you info to start debugging.

hardy1334:
okay, thanks a lot

hardy1334:
@ragesoss  can u please have a look at this

Issue #2511, opened 2019-02-21
Redesign of Diff Viewer


bwreid:
![image](https://user-images.githubusercontent.com/1316902/53186854-b6e08c00-35b6-11e9-8d3e-70f0bf7cb1f5.png)

We should redesign the Diff Viewer a bit as it's currently somewhat cluttered. Some general notes on how to improve this:

- The article title does not stand out in any way
- The three buttons on the right-hand side of the modal are a bit of a mess
- The section for adding a Salesforce record isn't differentiated from the rest of the navigation

Let's improve this as much as possible so it's fun to use!

ananyaarun:
@bwreid , i would like to take this up. Can u give me more info on how to reproduce this ?

bwreid:
@ananyaarun Probably the easiest way to get this view is to run `bundle exec rake dev:populate`. That will set up three courses for you.

Once you've run that, go to the "Ragesoss" test course:
http://localhost:3000/courses/test/Ragesoss_(test)

Click the "Article" sub navigation and then click the +/- button under "Assessment tools:"
![image](https://user-images.githubusercontent.com/1316902/53752428-6cd39200-3e63-11e9-8224-f53e5a371380.png)

That will bring up the diff viewer. Alternatively, you can follow [the instructions here.](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/user_roles.md#populating-a-course-with-data)

Let me know if you need additional help!

bwreid:
@mariahajmal If you don't hear from @ananyaarun in the next day or so, please feel free to work on this.

hardy1334:
@mariahajmal  @ananyaarun @bwreid   if u r not working on this shall i take this issue

ragesoss:
The latest version improves on the title, and moves the previous/next buttons next to each other:
![Screenshot from 2019-03-29 19-08-57](https://user-images.githubusercontent.com/31733278/55239896-a37fab00-525d-11e9-860e-306de4c8fd4d.png)

Remaining areas for improvement include:
* clustered buttons at the top left... we should probably just get rid of the 'how did it work' link, as it's not a new feature by any means
* layout when there is nothing on one of the columns
* Salesforce section (admin-only) layout
* placement and display styling of the edit summary and timestampe
* timestamp jumps position after diff loads

ragesoss:
@Kirsty21 yes, it's still open. Go for it. :-)

[...]

PR #2697, opened 2019-03-29
Diff Viewer Redeseigned

hardy1334:
In this I did some changes in designing the **diff viewer** for issue #2511 

**Earlier:**

Screenshot from 2019-03-29 20-05-20

**Now:**

Screenshot from 2019-03-29 19-08-57

### DIFF
--- a/app/assets/javascripts/components/revisions/diff_viewer.jsx
+++ b/app/assets/javascripts/components/revisions/diff_viewer.jsx
@@ -206,7 +206,7 @@ const DiffViewer = createReactClass({
     return (
       <button
         onClick={this.showPreviousArticle}
-        className="button dark small"
+        className="button shift-previous dark small"
       >
         {I18n.t('articles.previous')}
       </button>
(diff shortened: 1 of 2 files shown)


hardy1334:

ragesoss:
This looks like a nice improvement. With the title styling, I think we can cut out the "article title" label.

hardy1334:
@ragesoss  ok,I'll remove article title from it

hardy1334:
**Current version:**

Screenshot from 2019-03-29 22-20-40

hardy1334:
@ragesoss  now is it fine ?

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -206,7 +206,7 @@ const DiffViewer = createReactClass({
     return (
       <button
         onClick={this.showPreviousArticle}
-        className="button dark small"
+        className="button shift-previous dark small"
```
ragesoss: Instead of introducing a new utility class here, we can re-use one of the many ways already in the CSS for getting the spacing to appear. The best one I can find is to add the `pull-right` class to this button, and to add the `margin` class to the first button.

hardy1334:
@ragesoss  Is it good to go now ?

Issue #2384, opened 2019-01-16
Add warning when instructors change course dates, to discourage re-using an old course


ragesoss:
This would only be relevant for Wiki Education, but we want to head off instructors from changing the course dates, removing their studenrts, and re-using the same course page for a subsequent term.

They should still be able to change the dates on an approved course that hasn't been closed yet, but they should see some kind of messaging to let them know start a new course rather than change the dates to a whole different term.

PR #2698, opened 2019-03-30
Alert added when Instructor tries to change dates of expired course

hardy1334:
For the issue #2384 In this I have added option for alert whenever Instructor tries to change dates of expired course and as he clicks on **Event end** button to change dates, he will be shown this alert to discourage him from changing dates of expired courses.
This alert is only shown when Instructor tries to change dates of expired course.

Screenshot from 2019-03-31 02-45-25


hardy1334:
@bwreid  @ragesoss  can either of u have a look at this

hardy1334:
@ragesoss yeah,you are right that thing which u just told is awesome

ragesoss:

PR #2730, opened 2019-04-06
Adding Vikidia to wikis [WIP]

ferdi2005:
Vikidia (vikidia.org) is a growing wiki dedicated to childrens. It's doing many project with school and the dashboard would be very useful for them, so why don't add vikidia to wikis to help them managing projects.

### DIFF
--- a/app/models/wiki.rb
+++ b/app/models/wiki.rb
@@ -33,6 +33,7 @@ class Wiki < ApplicationRecord
     wikiversity
     wikivoyage
     wiktionary
+    vikidia
   ].freeze
   validates_inclusion_of :project, in: PROJECTS
 


ragesoss:
Supporting Vikidia would require a lot more changes, because it's not a wiki that has its revision data in the wmflabs.org 'replica' database, which is where the Dashboard pulls all the edit and article data from. Basically, we'd need to add code to fetch article and revision data directly from the MediaWiki API for wikis that aren't available through the replica DB, and add handling multiple places to switch to that method for Vikidia.

ragesoss:
Hmm... it would also mean figuring out a way to handle the usernames, since they aren't part of the Wikimedia unified login system and might overlap with Wikimedia accounts but be different users.

The easiest path I could think of would be to add code to support direct MediaWiki queries for revision data, but then also adding configuration to allow a whole separate instance of the Dashboard configured to track an arbitrary MediaWiki site, and have a whole different Dashboard for Vikidia. This would be possible, but a lot of work. I don't think it's practical to add Vikidia support to P&E Dashboard.

ferdi2005:
I'll try to work on that, on a separate dashboard only for Vikidia.

ragesoss:
Awesome! Let me know if you have questions. I can send you an invite to our dev Slack if you want.

ferdi2005:
Absolutely yes!

ragesoss:
Closing this for now, but please open a new one or reopen this one when the time comes.

PR #2746, opened 2019-04-11
Use Delivered status on Tickets

bwreid:
example

Adds delivered status to individual messages and integrates database changes from TicketDispenser.


ragesoss:
LGTM!

PR #2748, opened 2019-04-11
Salesforce credentials

ragesoss:
## What this PR does
Adds a frontend for super admins to update Salesforce password and security token stored in a new Setting record, and uses that Setting record to override the values from application.yml when making Salesforce API calls.

sf

### DIFF
--- a/app/assets/javascripts/components/settings/settings_handler.jsx
+++ b/app/assets/javascripts/components/settings/settings_handler.jsx
@@ -36,22 +36,22 @@ export const SettingsHandler = createReactClass({
     if (Features.wikiEd) {
       salesforce = (
         <React.Fragment>
-          <h1 className="mx2" style={{ display: 'inline-block', maring: 0 }}>Other Settings</h1>
+          <h1 className="mx2">Other Settings</h1>
           <hr />
-          <h2 className="mx2" style={{ display: 'inline-block', maring: 0 }}>Salesforce</h2>
+          <h2 className="mx2">Salesforce</h2>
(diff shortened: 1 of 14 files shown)


app/assets/javascripts/components/settings/views/salesforce_credentials_form.jsx
```diff
@@ -0,0 +1,57 @@
...
+      <tr>
+        <td>
+          <form onSubmit={this.handleSubmit}>
+            <TextInput
+              id="salesforce_password"
+              editable
+              onChange={this.handleChange}
+              value={this.state.salesforce_password}
+              value_key="salesforce_password"
+              type="text"
```
bwreid: Should this be a password field?

app/assets/javascripts/components/settings/settings_handler.jsx
```diff
@@ -31,6 +32,17 @@ export const SettingsHandler = createReactClass({
   },
 
   render() {
+    let salesforce;
+    if (Features.wikiEd) {
+      salesforce = (
+        <React.Fragment>
+          <h1 className="mx2" style={{ display: 'inline-block', maring: 0 }}>Other Settings</h1>
+          <hr />
+          <h2 className="mx2" style={{ display: 'inline-block', maring: 0 }}>Salesforce</h2>
```
bwreid: Margin?

bwreid: Also, these would ideally be a class, as well.

PR #2749, opened 2019-04-11
Ticket coverage

ragesoss:
This covers most (maybe all if I'm lucky) of the uncovered Ruby code.


spec/controllers/tickets_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    it 'renders the main tickets dashboard URL' do
+      get '/tickets/dashboard'
+      expect(response.status).to eq(200)
+      expect(response.body).to include("id='react_root'")
+    end
+
+    it 'renders an arbitrary tickets path' do
+      get '/tickets/foo'
+      expect(response.status).to eq(200)
+      expect(response.body).to include("id='react_root'")
```
bwreid: This is a bit of an odd test. We're basically just checking that this route exists via React Router, right? If so, it may be useful to just rename this to something more like "has a tickets 'show' path" or something more direct.

ragesoss: Yeah, I guess testing for a real show path makes more sense.

PR #2751, opened 2019-04-12
Basic ticket filtering, following the example of Uploads

ragesoss:
This adds a barebones implementation of filters based on owner. It will take some refactoring to add support for multiple different kinds of filters, and we'll also want to add default filters to start off with.

### DIFF
--- a/app/assets/javascripts/selectors/index.js
+++ b/app/assets/javascripts/selectors/index.js
@@ -231,3 +231,9 @@ export const getFilteredTickets = createSelector(
     return _.filter(tickets.all, ticket => ownerIds.includes(ticket.owner.id));
   }
 );
+
+export const getTicketsById = createSelector(
+  [getTickets], (tickets) => {
+    return tickets.all.reduce((acc, ticket) => ({ ...acc, [ticket.id]: ticket }), {});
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/selectors/index.js
```diff
@@ -221,3 +223,17 @@ export const editPermissions = createSelector(
     return user.isAdmin || !course.closed;
   }
 );
+
+export const getFilteredTickets = createSelector(
+  [getTickets], (tickets) => {
+    const ownerIds = tickets.filters.map(filter => filter.value);
+    if (ownerIds.length === 0) { return tickets.all; }
```
bwreid: What about just:
```js
if (!ownerIds.length) { return tickets.all; }
```

app/assets/javascripts/selectors/index.js
```diff
@@ -221,3 +223,17 @@ export const editPermissions = createSelector(
...
+    return _.filter(tickets.all, ticket => ownerIds.includes(ticket.owner.id));
```
bwreid: Could we just use the regular `.filter()` function here?

PR #2761, opened 2019-04-15
Add ticket filtering by status, set default filters

ragesoss:
This adds a second selector for filtering by ticket status, and sets the default filters we want, on the same criteria that would make the bell ring: open tickets that are either owned by you or unassigned.

Filters!

### DIFF
--- a/app/assets/javascripts/actions/tickets_actions.js
+++ b/app/assets/javascripts/actions/tickets_actions.js
@@ -130,4 +130,5 @@ export const deleteTicket = id => async (dispatch) => {
   dispatch({ type: DELETE_TICKET, id });
 };
 
-export const setTicketFilters = filters => ({ type: FILTER_TICKETS, filters });
+export const setTicketOwnersFilter = filters => ({ type: FILTER_TICKETS, filters: { owners: filters } });
+export const setTicketStatusesFilter = filters => ({ type: FILTER_TICKETS, filters: { statuses: filters } });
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/actions/tickets_actions.js
```diff
@@ -130,4 +132,19 @@ export const deleteTicket = id => async (dispatch) => {
...
+export const setTicketStatusesFilter = filters => ({ type: FILTER_TICKETS, filters: { statuses: filters } });
+
+export const setInitialTicketFilters = () => (dispatch, getState) => {
+  // Open tickets only
+  dispatch(setTicketStatusesFilter([{ value: TICKET_STATUS_OPEN, label: STATUSES[TICKET_STATUS_OPEN] }]));
+
+  // Owned by current user, or no one
+  const state = getState();
+  const currentUserId = state.currentUserFromHtml.id;
+  const currentUser = state.admins.find(admin => admin[1] === currentUserId);
```
bwreid: This is a little confusing on first read. It might be better as:
```js
const currentUser = state.admins.find(([ _username, id ]) => id === currentUserId);
```

You could go even further and apply destructuring to remove the next line:
```js
const [ label, value ] = state.admins.find(([ _username, id ]) => id === currentUserId);
```

ragesoss: Looks good. Done.

PR #2763, opened 2019-04-15
Add sender_email, subject support to tickets

ragesoss:
This adds extra info to the `details` field of Ticket messages, and uses that in the Ticket Dashboard interface

ticket email, subject

ticket show email, subject

### DIFF
--- a/app/assets/javascripts/components/tickets/reply.jsx
+++ b/app/assets/javascripts/components/tickets/reply.jsx
@@ -11,9 +11,20 @@ export const Reply = ({ message }) => {
   const failedTime = moment(message.details.delivery_failed).format('YYYY/MM/DD h:mm a');
   const failed = `Failed on ${failedTime}`;
 
+  let subject;
+  if (message.details.subject) {
+    subject = (
+      <React.Fragment>
+        <h4 className=".message-subject">{ message.details.subject }</h4>
+        <hr />
+      </React.Fragment>
+    );
+  }
+
   return (
     <React.Fragment>
       <section className="module mb0 mt0">
+        {subject}
         <div className="plaintext" dangerouslySetInnerHTML={{ __html: linkifyHtml(message.content) }} />
       </section>
       <aside className="reply-details">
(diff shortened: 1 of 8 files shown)


ragesoss:
@bwreid I'm going ahead with pushing this now, but please do review it when you get a chance.

app/assets/javascripts/components/tickets/reply.jsx
```diff
@@ -11,9 +11,20 @@ export const Reply = ({ message }) => {
   const failedTime = moment(message.details.delivery_failed).format('YYYY/MM/DD h:mm a');
   const failed = `Failed on ${failedTime}`;
 
+  let subject;
+  if (message.details.subject) {
+    subject = (
+      <React.Fragment>
+        <h4 className=".message-subject">{ message.details.subject }</h4>
```
bwreid: Did you mean for this to be `.message-subject` as opposed to `message-subject`?

app/assets/javascripts/components/tickets/tickets_table_row.jsx
```diff
@@ -7,10 +7,13 @@ import TicketOwnerHandler from './ticket_owner_handler';
 const TicketsTableRow = ({ ticket }) => {
   return (
     <tr className={ticket.status === 0 ? 'table-row--faded' : ''}>
+      <td className="w10">
+        {ticket.sender.real_name || ticket.sender.username || ticket.sender_email || 'Unknown User Record' }
```
bwreid: I wonder if we should extract this out, at this point. We could define a variable above and still leave the default below.
```js
const { sender, sender_email } = ticket;
const user = sender.real_name || sender.username || sender_email;
// ...
{ user || 'Unknown User Record' }
```

app/assets/javascripts/components/tickets/reply.jsx
ragesoss: No. Don't need the class at all, actually. Whoops.

PR #2767, opened 2019-04-16
Allow for forwarded emails to attempt to pull correct sender

bwreid:
If an email is sent and includes the specified forwarding email as a CC, attempt to search through the message to find the sender.

### DIFF
--- a/lib/email_processor.rb
+++ b/lib/email_processor.rb
@@ -13,10 +13,10 @@ class EmailProcessor
     dispense_or_thread_ticket
   end
 
-  def email_addresses_from_email
+  def retrieve_forwarder_email
     raw_body = @email.raw_body
     expression = /\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{0,63}\b/i
-    raw_body.scan(expression)
+    raw_body.scan(expression).first
   end
 
   private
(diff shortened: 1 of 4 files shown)


spec/lib/email_processor_spec.rb
```diff
@@ -90,5 +89,43 @@
 
       expect(ticket.messages.length).to eq(2)
     end
+
+    it 'should assign forwarded emails to the original sender' do
+      student
+      body = <<~EXAMPLE
+        Example email test\r\n\r\n---------- Forwarded message ---------\r\nFrom:
+        Student <student@email.com>\r\nDate: Mon, Apr 8, 2019 at 3:42 PM\r\n
+        Subject: Help!\r\nTo: <staff@email.com>\r\n\r\n\r\nHelp message\r\n
```
ragesoss: If `staff@email.com` is also a user, it'll probably be the earlier one in the database, so we need to be sure that we don't match that as the sender.

lib/email_processor.rb
```diff
@@ -3,24 +3,42 @@
...
+    define_owner
+    define_sender
+    define_content_and_reference_id
+    dispense_or_thread_ticket
+  end
+
+  def retrieve_forwarder_email
+    raw_body = @email.raw_body
+    expression = /\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{0,63}\b/i
+    raw_body.scan(expression).first
```
ragesoss: This will still scan the entire body for all matches... I think `raw_body.match(expression)` (or, if you want to get fancy, `raw_body[expression]`) would be the more efficient way, although it's trivial.

PR #2771, opened 2019-04-17
Minor changes to ticket response email

bwreid:
Minor change to an email that goes out when an admin responds to a ticket. Removes the `visibility: hidden;` CSS in favor of just changing the text to white. Also reorients some of the hidden text.


ragesoss:
lgtm

PR #2789, opened 2019-04-22
Feature/ignore tickets with code

bwreid:
Skip over emails that include the `TICKET_IGNORE_CODE`. Only looks at `@email.body` which means only the most recent email should be affected.

### DIFF
--- a/config/application.example.yml
+++ b/config/application.example.yml
@@ -106,8 +106,11 @@
 # Default sender for emails
   SENDER_EMAIL_ADDRESS: 'test@example.com'
 
+# ==== Ticketing System Environment Variables
 # Email at which to receive forwarded ticket emails
   TICKET_FORWARDING_DOMAIN: 'wikiedu.org'
+# Code to include if emails should be ignored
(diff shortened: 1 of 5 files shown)


config/application.example.yml
```diff
@@ -106,8 +106,11 @@
 # Default sender for emails
   SENDER_EMAIL_ADDRESS: 'test@example.com'
 
+# ==== Ticketing System Environment Variables
 # Email at which to receive forwarded ticket emails
   TICKET_FORWARDING_DOMAIN: 'wikiedu.org'
+# Code to include if emails should be ignored
+  TICKET_IGNORE_CODE: 'ticket_ignore_code'
```
ragesoss: Why not just hard-code the ticket ignore code?

I could imagine a point in the future where we'd want this to be configurable, for example, with a standard scheme that embeds a string representing the email type in that hidden section — `dashboard.wikiedu.org_email_type:overdue_training_alert` or the like — and then using a Setting to hold the list of all email types to ignore.

Meanwhile, maybe there's no need to go through ENV.

PR #2790, opened 2019-04-22
Retrieve course by slug if unable to find at first

bwreid:
If we're unable to assign the ticket to a course, try searching through the email for a course slug.

### DIFF
--- a/lib/email_processor.rb
+++ b/lib/email_processor.rb
@@ -32,7 +32,7 @@ class EmailProcessor
   def email_can_be_ignored?
     @email.body.include?('ignore_creating_dashboard_ticket')
   end
-  
+
   # This regex helps look for a course slug inside of text
   # (?<=\/courses\/) - Starts with `/courses/`
   # [^?\/\s] - Any non-whitespace character excluding a slash or question mark
@@ -58,11 +58,8 @@ class EmailProcessor
   end
 
   def define_course
-    @course = @sender.courses.last if @sender
-    if @course.nil?
-      slug = retrieve_course_slug_by_url
-      @course = Course.find_by(slug: slug) if slug
-    end
+    @course = Course.find_by(slug: retrieve_course_slug_by_url)
+    @course ||= @sender.courses.last if @sender
   end
 
   def from_signature(from)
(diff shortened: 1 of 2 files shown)


spec/lib/email_processor_spec.rb
```diff
@@ -166,4 +187,25 @@
       expect(described_class.new(email).retrieve_forwarder_email).to eq(expected_result)
     end
   end
+
+  describe '#retrieve_course_by_url' do
+    it 'should return the first slug from a message' do
+      body = <<~EXAMPLE
+        Example email test\r\n\r\nhttps://example.org/courses/example/slug
```
ragesoss: We should test also that the right slug is found for a url like `https://dashboard.wikiedu.org/courses/School_of_the_Art_Institute_of_Chicago/New_Histories_of_Chicago_Media_(Spring_2019)/articles/edited?showArticle=51050930`

lib/email_processor.rb
```diff
@@ -19,8 +20,23 @@ def retrieve_forwarder_email
...
+    raw_body = @email.raw_body
+    expression = course_slug_pattern
+    raw_body[expression]
+  end
+
   private
 
+  # This regex helps look for a course slug inside of text
+  # (?<=\/courses\/) - Starts with `/courses/`
+  # [^?\/\s] - Any non-whitespace character excluding a slash
```
ragesoss:
```suggestion
  # [^?\/\s] - Any non-whitespace character excluding a slash or question mark
```

lib/email_processor.rb
```diff
@@ -19,8 +20,23 @@ def retrieve_forwarder_email
...
+  # \/ - A slash
+  # [^?\/\s] - Any non-whitespace character excluding a slash
```
ragesoss:
```suggestion
  # [^?\/\s] - Any non-whitespace character excluding a slash or question mark
```

lib/email_processor.rb
```diff
@@ -38,6 +54,10 @@ def define_sender
 
   def define_course
     @course = @sender.courses.last if @sender
+    if @course.nil?
+      slug = retrieve_course_slug_by_url
+      @course = Course.find_by(slug: slug) if slug
```
ragesoss: Course slug will never be nil, so you could just do `@course = Course.find_by(slug: retrieve_course_slug_by_url)`.

This is probabably a more reliable matcher than sender's last course, too. Sometimes an instructor has several simultaneous courses.

I suggest:

```ruby
@course = Course.find_by(slug: retrieve_course_slug_by_url)
@course ||= @sender.courses.last if @sender
```

PR #2796, opened 2019-04-23
Fallback to email in the event the original sender is not a user

bwreid:
If a recipient can't be found, fall back to using the sender email.

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -10,10 +10,11 @@ class TicketsController < ApplicationController
 
   def notify
     message = TicketDispenser::Message.find(notification_params[:message_id])
-    sender = User.find(notification_params[:sender_id])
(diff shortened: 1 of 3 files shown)


ragesoss:
LGTM

app/controllers/tickets_controller.rb
```diff
@@ -10,10 +10,11 @@ def dashboard
 
   def notify
     message = TicketDispenser::Message.find(notification_params[:message_id])
-    sender = User.find(notification_params[:sender_id])
+    sender_email = message.details[:sender_email]
+    sender = User.find(notification_params[:sender_id]) || sender_email
     ticket = message.ticket
     course = ticket.project
-    recipient = ticket.reply_to
+    recipient = ticket.reply_to || User.new(email: sender_email)
```
ragesoss: Nice. I would not have thought to do it like that.

PR #2799, opened 2019-04-24
Add Email Owner button

bwreid:
This adds a button titled "Email Ticket Owner" which will email the owner of a ticket with a message.

image

Right now a notification is shared at the top of the screen but is red -- how can we modify this so that it doesn't look like an error?

### DIFF
--- a/app/assets/javascripts/actions/tickets_actions.js
+++ b/app/assets/javascripts/actions/tickets_actions.js
@@ -36,7 +36,8 @@ export const notifyOfMessage = body => async (dispatch) => {
         type: ADD_NOTIFICATION,
         notification: {
           message: 'Email was sent to the owner.',
-          status: 'success'
+          type: 'success',
+          closable: true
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/actions/tickets_actions.js
```diff
@@ -12,10 +12,40 @@ import {
...
+    });
+    if (!response.ok) {
+      const json = await response.json();
+      dispatch({ type: API_FAIL, data: { statusText: json.message } });
+    } else {
+      dispatch({
+        type: ADD_NOTIFICATION,
+        notification: {
+          message: 'Email was sent to the owner.',
+          status: 'success'
```
ragesoss: I think `type: 'success'` will fix the color.

ragesoss: Probably also we want `closable: true`.

PR #2804, opened 2019-04-25
Feature/ticket notes

bwreid:
This adds the ability to store private notes on the Ticket Dashboard. I'm not sure I'm really in love with the design or anything...

image

### DIFF
--- a/app/controllers/tickets_controller.rb
+++ b/app/controllers/tickets_controller.rb
@@ -20,11 +20,11 @@ class TicketsController < ApplicationController
 
   def send_ticket_notification(owner=false)
     message = TicketDispenser::Message.find(notification_params[:message_id])
-    sender_email = message.details[:sender_email]
+    sender_email = message.ticket.messages.first.details[:sender_email]
     sender = User.find(notification_params[:sender_id]) || sender_email
     ticket = message.ticket
     course = ticket.project
-    recipient = owner ? ticket.owner : ticket.reply_to || User.new(email: sender_email)
+    recipient = owner ? ticket.owner : (ticket.reply_to || User.new(email: sender_email))
 
     TicketNotificationMailer.notify_of_message(course, message, recipient, sender)
 
--- a/app/helpers/tickets_helper.rb
+++ b/app/helpers/tickets_helper.rb
@@ -5,4 +5,16 @@ module TicketsHelper
   def sender_for_message(message)
     message.sender&.real_name || message.sender&.username || message.details[:sender_email]
   end
+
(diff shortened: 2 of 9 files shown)


app/controllers/tickets_controller.rb
```diff
@@ -20,11 +20,11 @@ def notify_owner
 
   def send_ticket_notification(owner=false)
     message = TicketDispenser::Message.find(notification_params[:message_id])
-    sender_email = message.details[:sender_email]
+    sender_email = message.ticket.messages.first.details[:sender_email]
```
bwreid: That error you were getting related to the `to` column has to do with this snippet. `:sender_email` is only stored on the first message. When we send a reply, it's not certain that we will have access to the sender's email. This walks back up to the first message and grabs the sender email from there.

Alternatively, what about sending just the first message from the front end? I think it might make the front-end code a bit more complex but it'll make less database calls.

ragesoss: Hmm... this seems fine. This particular DB call is going to be trivial and fast.

app/helpers/tickets_helper.rb
```diff
@@ -5,4 +5,16 @@ module TicketsHelper
   def sender_for_message(message)
     message.sender&.real_name || message.sender&.username || message.details[:sender_email]
   end
+
+  def successful_replies_in_reverse(_ticket, recipient)
+    successful_messages = @ticket.messages.reject do |message|
+      message.details[:delivery_failed]
```
ragesoss: As we saw yesterday, a message can be both successful and have a failed delivery, so we should not reject a failed one if it also succeeded.

app/helpers/tickets_helper.rb
```diff
@@ -5,4 +5,16 @@ module TicketsHelper
...
+    end
+
+    reversed_messages = successful_messages.reverse[1..-1]
+
+    reversed_messages.select do |message|
+      recipient.admin? || message.kind == TicketDispenser::Message::Kinds::REPLY
```
ragesoss: This would be nicer if we could ask `message` more directly: `message.reply?`.

app/helpers/tickets_helper.rb
bwreid: That's true. I don't think we could do a `.select` for `:delivered` though as that would exclude the first message. What about:

```ruby
message.details[:delivered].nil? && message.details[:delivery_failed]
```

ragesoss: Yeah, that sounds right.

PR #2823, opened 2019-04-30
Make multiple requests to load tickets

bwreid:
This PR separates a single call to the API for all tickets to segmented calls to the API. In total, four calls are made for tickets within 15 days of each other each.

This PR includes pagination.

pagination

This PR will need [this PR from TicketDispenser](https://github.com/WikiEducationFoundation/TicketDispenser/pull/7) to work correctly.

### DIFF
--- a/app/assets/javascripts/actions/tickets_actions.js
+++ b/app/assets/javascripts/actions/tickets_actions.js
@@ -14,7 +14,9 @@ import {
 import { STATUSES } from '../components/tickets/util';
 import { API_FAIL } from '../constants/api';
 import { ADD_NOTIFICATION } from '../constants';
+
 import fetch from 'cross-fetch';
+import moment from 'moment';
 
 const getCsrf = () => document.querySelector("meta[name='csrf-token']").getAttribute('content');
 
@@ -102,7 +104,6 @@ export const createReply = (body, status, bcc_to_salesforce) => async (dispatch)
   dispatch({ type: CREATE_REPLY });
 };
 
-
 export const readAllMessages = ticket => async (dispatch) => {
   const unreadMessages = ticket.messages.some(message => !message.read);
   if (!unreadMessages) return false;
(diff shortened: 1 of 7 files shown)


ragesoss:
eslint error is breaking the build.

bwreid:
Woops. Thought I had fixed that already. New version just pushed up.

app/assets/javascripts/components/tickets/tickets_handler.jsx
```diff
@@ -53,9 +65,12 @@ export class TicketsHandler extends React.Component {
       }
     };
 
+    const TICKETS_PER_PAGE = 10;
+
     const elements = this.props.filteredTickets.map(ticket => (
       <Row key={ticket.id} ticket={ticket} />
-    ));
+    )).slice(this.state.page, this.state.page + TICKETS_PER_PAGE);
+    const pagesLength = Math.floor(elements.length / TICKETS_PER_PAGE) + 1;
```
ragesoss: This should be done based on the full filteredTickets array, not just the slice of elements on the current page. I think it also shows an extra page when the ticket count is a multiple of ten. Maybe should be `(count - 1) / TICKETS_PER_PAGE` ?

app/assets/javascripts/actions/tickets_actions.js
```diff
@@ -120,13 +121,40 @@ export const readAllMessages = ticket => async (dispatch) => {
...
+  });
+};
+
+// Fetch as many tickets as possible
 export const fetchTickets = () => async (dispatch) => {
   dispatch({ type: FETCH_TICKETS });
 
-  const response = await fetch('/td/tickets');
-  const json = await response.json();
+  const segments = dateSegments();
```
ragesoss: Locally, I have 107 tickets, I end up with more than 400 in `state.tickets.all`. Also seeing some duplicate key warnings in some cases, so it looks like each ticket is ending up represented multiple times in the set.

ragesoss: Oh, of course, that's because I haven't updated TicketDispenser.

ragesoss:
@bwreid please review the last change I just pushed up. Does the change to the slice params look right?

bwreid:
This looks right to me!

ragesoss:
One thing I ran into before I had the TicketDispenser code updated is duplicate keys, which will throw warnings and potentially cause problems on the frontend. The frontend assumes that each independent request for tickets will return non-overlapping results. Better handling of arbitrary, possibly duplicate sets of tickets may be something we need to handle later on, if we add more/different options for extra queries.

PR #2832, opened 2019-05-06
Better handling of plagiarism emails for Scholars & Scientists

ragesoss:
Bug fix for when instructor and Wikipedia Expert are both staff and both have greeter: true. Also adds support for multiple Wikipedia Experts in the same course.


bwreid:
This looks good to me!

PR #2834, opened 2019-05-06
Add alerts for edits to Good or Featured articles

ragesoss:
Add alerts for whenever students edit Good or Featured articles, using the same strategy as DiscretionarySanctionsMonitor. All Good and Featured articles are in the respective category, although there are about 30,000 GAs and over 5,000 FAs.

Using CategoryImporter for these large categories makes it take ~30 seconds, so it will lengthen the `ConstantUpdate` cycle by that much.

These particular alerts share a lot of logic with the discretionary sanctions ones in particular and with plenty of similarity to several other alert monitors, but there are just enough differences among these monitors that I don't think it quite makes sense to refactor out the common logic. They also have a good track record of not requiring any changes, and would be as likely as not to move in different directions if they did change, so I think this duplication is an okay strategy here.

### DIFF
--- a/app/models/alert.rb
+++ b/app/models/alert.rb
@@ -42,6 +42,7 @@ class Alert < ApplicationRecord
     DYKNominationAlert
     FirstEnrolledStudentAlert
     GANominationAlert
+    HighQualityArticleEditAlert
     NeedHelpAlert
     NoEnrolledStudentsAlert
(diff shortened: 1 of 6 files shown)


bwreid:
Based on the other Monitors, this looks reasonable to me.

Issue #2855, opened 2019-05-10
NoMethodError: undefined method `details' for nil:NilClass


ragesoss:
https://sentry2.wikiedu.org/wikiedu/production/issues/3045/

```
NameError: NameError
  active_model/serializer.rb:110:in `block in attribute'
    raise NameError
  active_model/serializer.rb:456:in `attributes'
    _fast_attributes
  active_model/serializer.rb:480:in `_serializable_hash'
    attributes
  active_model/serializer.rb:359:in `serializable_hash'
    _serializable_hash
  active_model/serializer.rb:343:in `as_json'
    hash.merge!(root => serializable_hash)
...
(167 additional frame(s) were not displayed)

NoMethodError: undefined method `details' for nil:NilClass
  config/initializers/ticket_dispenser.rb:10:in `sender'
    return { email: message.details[:sender_email] } if user.nil?
...
(170 additional frame(s) were not displayed)

NoMethodError: undefined method `details' for nil:NilClass
```

PR #2856, opened 2019-05-10
Update TD initializer

bwreid:
Should fix Issue #2855.

If this error occurred, then we must have a ticket with no messages in Production.


ragesoss:
@bwreid see the Sentry details. It came from deleting a ticket.

bwreid:
Ah, that makes sense. Just tried deleting locally and this still fixed that error from showing up in my logs.

PR #2867, opened 2019-05-14
Resources tab

ragesoss:
Adds a Resources tab, and removes the Training link from the main nav when you're on a Wiki Education course page.

It will embed any Timeline block called 'Additional Resources' onto the Resources tab, and will also embed the wizard-generated block of subject-specific handouts (based on block title).

once more

### DIFF
--- a/app/assets/javascripts/components/resources/handouts.jsx
+++ b/app/assets/javascripts/components/resources/handouts.jsx
@@ -7,13 +7,16 @@ const handouts = [
   { name: 'Illustrating Wikipedia', link: 'https://wikiedu.org/illustratingwikipedia' }
 ];
 
+const HANDOUTS_BLOCK_KIND = 4;
+
+
 const Handouts = ({ blocks, trainingLibrarySlug }) => {
   let topicGuides;
-  const topicGuidesBlock = blocks.find(block => block.title.match(/writing articles in your topic area/));
-  if (topicGuidesBlock) {
+  const topicGuidesBlocks = blocks.filter(block => block.kind === HANDOUTS_BLOCK_KIND);
+  if (topicGuidesBlocks) {
     topicGuides = (
       <>
-        <Block block={topicGuidesBlock} trainingLibrarySlug={trainingLibrarySlug} />
+        {topicGuidesBlocks.map(block => <Block key={block.id} block={block} trainingLibrarySlug={trainingLibrarySlug} />)}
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/components/resources/handouts.jsx
```diff
@@ -0,0 +1,40 @@
...
+  { name: 'Evaluating Wikipedia', link: 'https://wikiedu.org/evaluatingwikipedia' },
+  { name: 'Illustrating Wikipedia', link: 'https://wikiedu.org/illustratingwikipedia' }
+];
+
+const Handouts = ({ blocks, trainingLibrarySlug }) => {
+  let topicGuides;
+  const topicGuidesBlock = blocks.find(block => block.title.match(/writing articles in your topic area/));
+  if (topicGuidesBlock) {
+    topicGuides = (
+      <>
```
bwreid: What's happening here?

app/assets/javascripts/components/resources/resources.jsx
```diff
@@ -0,0 +1,71 @@
...
+import Handouts from './handouts';
+
+const Resources = ({ weeks, current_user, course }) => {
+  const trainingLibrarySlug = course.training_library_slug;
+  let instructorModulesLink;
+  if (current_user.isInstructor && Features.wikiEd) {
+    instructorModulesLink = <CourseLink to={'/training/instructors'} className="button pull-right">Instructor orientation modules</CourseLink>;
+  }
+  const blocks = _.flatten(weeks.map(week => week.blocks));
+  const modules = _.compact(_.flatten(blocks.map(block => block.training_modules)));
```
bwreid: Something like [the flow()](https://lodash.com/docs/4.17.11#flow) method might be useful here. Alternatively, just breaking this up into more variables might make it easier to read.

app/assets/javascripts/components/resources/handouts.jsx
ragesoss: This is syntax for React fragments that I learned about from an error message today: `<>...</>` :boom:

app/assets/javascripts/components/resources/resources.jsx
ragesoss: How's this? https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2867/commits/8e1eb3291061ed624c6970d9c39ff939534ceefd

bwreid: Nice! I'm into this.  :)

PR #2877, opened 2019-05-18
Clone Course for Programs & Events Dashboard

bwreid:
Adds additional places for a course to be cloned.

1. Clone the course from a course's page
  image

1. Clone the course from the campaign list
  image

1. From the campaign home page, allow for users to clone a course


app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -191,7 +191,7 @@ const AvailableActions = createReactClass({
     }
 
     // If the user is an admin and the course is both published and a Wiki-Ed course.
```
ragesoss:
```suggestion
    // If the user is an admin and the course is published
```

ragesoss:
LGTM

PR #2878, opened 2019-05-18
Allow campaign organizers to edit courses in the campaign

ragesoss:
That was a lot simpler than I expected.


ragesoss:
No, nothing comes to mind.

PR #2880, opened 2019-05-18
Refactor Gulp imports

takidelfin:
## What this PR does
Changes
```js
import gulp from 'gulp';
```
to suitable *`named import(s)`*

## Screenshots
Before:
n/a

After:
n/a
| `gulp`  | module imports | named imports |
| ------- | --- | --- |
| `build` | 80s | 51s |

(maybe it is a coincidence, but I ran this several times to check. The difference is always around 30sec on my computer)
(theoretically named imports should be faster than module imports)


ragesoss:
Sweet!

PR #2882, opened 2019-05-18
Show WikiData names if possible

bwreid:
Before:

image

After:

image

To Do:
- Add names on dropdown


ragesoss:
Nice! Looks good to me.

Fetching the labels happens via the `fetchRevisions` and `fetchArticles` actions after the results come back from the Dashboard server, so I expect the ones that aren't showing up are assigned Wikidata items that haven't been edited by the course. We could similarly add a wikidata label callback to `fetchAssignments`.

PR #2889, opened 2019-05-19
Rename and remove duplicated span.created_at

### LINKED ISSUE #2870: When sorting programs by date, show creation date
### Current Behavior:
When browsing https://outreachdashboard.wmflabs.org/explore, you can sort by campaign date, however the creation date is not shown, making it more difficult to judge which campaigns are the ones I'm looking for.

### Desired Behavior:
When sorting by campaign date (or perhaps just generally), it would be useful to show the campaign creation date.

takidelfin:
## What this PR does
Removes `span.creation-date.hidden` and renames `span.created_at` to `span.creation-date`


ragesoss:
Merged already

PR #2922, opened 2019-05-31
A Wiki Selector Component

AmitJoki:
This combines both language and project into a singular searchable component that works for both single-wiki and multi-wiki.

Before:
image

After:
image

### DIFF
--- a/app/assets/javascripts/components/common/wiki_select.jsx
+++ b/app/assets/javascripts/components/common/wiki_select.jsx
@@ -4,17 +4,6 @@ import AsyncSelect from 'react-select/lib/Async';
 import PropTypes from 'prop-types';
 
 const options = [];
-const languages = JSON.parse(WikiLanguages);
-const projects = JSON.parse(WikiProjects);
-
-for (let i = 0; i < languages.length; i += 1) {
-  for (let j = 0; j < projects.length; j += 1) {
-    const language = languages[i];
-    const project = projects[j];
-    options.push({ value: { language, project }, label: `${language}.${project}.org` });
-  }
-}
-
 /**
  *  A Wiki selector component that has both single and multi mode with searchable wikis.
  */
@@ -39,6 +28,19 @@ const WikiSelect = createReactClass({
   },
 
   render() {
+    if (options.length == 0) {
+      // cache the options so it doesn't run on every render
+      const languages = JSON.parse(WikiLanguages);
+      const projects = JSON.parse(WikiProjects);
+      for (let i = 0; i < languages.length; i += 1) {
+        for (let j = 0; j < projects.length; j += 1) {
+          const language = languages[i];
+          const project = projects[j];
+          options.push({ value: { language, project }, label: `${language}.${project}.org` });
+        }
+      }
+    }
+
     let wikis = [];
     if (this.props.wikis) {
       wikis = this.props.wikis.map((wiki) => {
--- a/app/assets/javascripts/components/common/wiki_select.jsx
+++ b/app/assets/javascripts/components/common/wiki_select.jsx
@@ -22,7 +22,7 @@ const WikiSelect = createReactClass({
      */
     styles: PropTypes.object,
     /**
-     *  An array of { language, project }. Only used if multi = true
+     *  An array of { language, project }
      */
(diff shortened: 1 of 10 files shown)


AmitJoki:
Yes. We should be using this component everywhere where the split components under `<fieldset>` is used. I saw the usage in Article Finder. But I am waiting to get a final call on this. There are some performance issues which I'd like to sort out. Once it is done, I will replace the one in Article Finder too.

app/assets/javascripts/components/common/wiki_select.jsx
```diff
@@ -0,0 +1,51 @@
...
+  propTypes: {
+    multi: PropTypes.bool,
+    onChange: PropTypes.func,
+    styles: PropTypes.object,
+    defaultLanguage: PropTypes.string,
+    defaultValue: PropTypes.string
+  },
+
+  render() {
+    const options = [];
```
ragesoss: These options can be a constant that gets set once outside of render?

AmitJoki:
@ragesoss I've cracked this :smile: There's an option to load asynchronously. I used that to only load the filtered wikis after 2 characters are typed in. There's no lag whatsoever now.

ragesoss:
@AmitJoki this is nice! A few further bits:
* There's one other case where you cannot select the right wiki: (www.)wikisource.org, which is multilingual wikisource (and exists in addition to all the language-specific versions like es.wikisource.org)
* I think maybe the 'type 2 more characters' hints should be removed. The current behavior of letting you just start typing when you focus the input, but you can see the current value before you start typing, should make it very clear how to use it.

app/assets/javascripts/components/common/wiki_select.jsx
```diff
@@ -0,0 +1,98 @@
...
+      // cache the options so it doesn't run on every render
+      const languages = JSON.parse(WikiLanguages);
+      const projects = JSON.parse(WikiProjects).filter(proj => proj !== 'wikidata');
+      for (let i = 0; i < languages.length; i += 1) {
+        for (let j = 0; j < projects.length; j += 1) {
+          const language = languages[i];
+          const project = projects[j];
+          options.push({ value: { language, project }, label: `${language}.${project}.org` });
+        }
+      }
```
bwreid: A couple things about this!

1. Sage and I were discussing and we think actually _all_ of this could separated out to a different file somewhere else. None of this code depends upon the context of the WikiSelect component, so it'd be worthwhile to just move it elsewhere and import it into this file.

1. Typically convention is to use native array methods and more "explanatory" methods and fall back to using `for` loops when you need to for a specific reason (e.g. performance). Unless you think the performance here could get particularly bad, it may be worthwhile to refactor to something like:
```js
const options = languages.map(language => {
  return projects.map(project => {
    const value = { language, project };
    const label = `${language}.${project}.org`;
    return { value, label }
  })
}).reduce((a, b) => a.concat(b)) // Flatten
```

AmitJoki: Actually, the WikiLanguages and WikiProjects global variable that gets defined in a HAML file is the reason I put them inside this component as refactoring this code outside of the component made a smoketest fail.

ragesoss:
Maybe you can add those globals to the test helper file.

app/assets/javascripts/utils/wiki_options.js
```diff
@@ -1,44 +1,21 @@
...
     return { value, label };
   })
 ).reduce((a, b) => a.concat(b));
+
 // Wikidata is multilingual with English as the default language and therefore has
 // a custom label so it is more intuitive.
 WIKI_OPTIONS.unshift({ value: { language: 'en', project: 'wikidata' }, label: 'www.wikidata.org' });
+
 // Wikisource has a standalone www.wikisource.org in addition to language based sites like es.wikisource.org
 WIKI_OPTIONS.unshift({ value: { language: 'en', project: 'wikisource' }, label: 'www.wikisource.org' });
```
ragesoss: This should be language `www` or language `null`. I'm not sure which, perhaps either will work. `en.wikisource` is another valid project, and this is separate from that.

ragesoss:
That fixed the jest specs. Yay!

PR #2925, opened 2019-05-31
Add 'Articles Edited' to Explore campaign list, improve query performance

ragesoss:
This adds a new column to the list of campaigns, Articles Edited, since this is a stat that one of the core Wiki Education program metrics (as opposed to Articles Created).

It also refactors how the counts get generated, so that a single SQL query most of the data that is summed across courses in one go.


ragesoss:
merged

PR #2936, opened 2019-06-05
Add previous Features  for each revision

khyatisoneji:


app/models/course_data/courses_users.rb
```diff
@@ -117,6 +118,13 @@ def character_sum(revisions, namespace)
       .sum(:characters) || 0
   end
 
+  def refs_sum_ms(revisions)
```
ragesoss: This method has the same name as the attribute, which I think will cause problems.

I also think we should avoid the shorthand "refs" and instead call the attribute something like `references_added`.

app/models/wiki_content/revision.rb
```diff
@@ -66,4 +67,10 @@ def plagiarism_report_link
     return unless ithenticate_id
     "/recent-activity/plagiarism/report?ithenticate_id=#{ithenticate_id}"
   end
+
+  def ref_tags_added
+    current_refs = self.features && self.features["feature.wikitext.revision.ref_tags"] || 0
```
ragesoss: The guard here against `nil` features shouldn't be necessary, since that attribute is serialized to hash. If the value is `NULL` in the database, it will serialize as an empty hash.

`self.` is also not needed here. The Revision instance will be the default receiver of a method call, so we only need `self.` in cases where it overlaps with a Ruby keyword or similar edge cases (or when it's important for clarity).

ragesoss:
Looks like you're picking up the Rails stuff quickly!

I suggest that we break up the two schema changes into separate PRs. First, one that adds the `features_previous` column to Revision and starts importing that data. Then once that is in place, we can add the CoursesUsers changes on top of that.

Revisions is a very large table, so that particular change I want to make sure we test out on a large database before we deploy. (@bwreid, I'm going to recruit your help for that: we should copy the production database over to staging, and then run that migration via the normal travis deployment process. The thing we want to make sure of is that the migration doesn't take too long, because if it does, the travis build times out after 10 minutes of no output, and will cut the connection and break the migration and deployment. So if it's anywhere close to that, we'll want to run the migration manually when it comes time to deploy to both P&E and Wiki Education production. Depending on how long it takes, we may also need to pause the update processes first.)

ragesoss:
For tests, the starting point will be to add some that exercise the basic flow of importing features for several revisions and returning the expected values for `references_added`.

`revision_score_importer_spec.rb` is a relevant example to start from.

db/schema.rb
```diff
@@ -237,6 +237,7 @@
     t.string "real_name"
     t.string "role_description"
     t.integer "total_uploads"
+    t.integer "refs_sum_ms", default: 0
```
ragesoss: This is left over from the other now-removed migration.

spec/models/revision_spec.rb
```diff
@@ -25,7 +25,90 @@
...
+               features_previous: {
+                 refs_tags_key => 17
+               })
+      end
+
+      it 'should return positive value' do
+        val = Revision.find_by(mw_rev_id: 870348507)
+        ref = val.features[refs_tags_key]
+        ref_previous = val.features_previous[refs_tags_key]
+        expect(val.references_added).to eq(ref - ref_previous)
```
ragesoss: I would do this with explicit values; we can see easily that the test includes 22 refs now vs 17 previously, so instead of doing `ref - ref_previous` in the test, it ought to just hard-code `5`.

app/models/wiki_content/revision.rb
```diff
@@ -66,4 +67,11 @@ def plagiarism_report_link
     return unless ithenticate_id
     "/recent-activity/plagiarism/report?ithenticate_id=#{ithenticate_id}"
   end
+
+  def references_added
+    current_refs = features['feature.wikitext.revision.ref_tags']
+    prev_refs = features_previous['feature.wikitext.revision.ref_tags'] || 0
+    return 0 if current_refs.nil? || current_refs < prev_refs
```
ragesoss: Negative value for references added is okay; it should return the negative number instead of zero in that case.

ragesoss:
This looks good to me. @bwreid can you review it and, when you have a chance, test out the migration as described above?

Once the migration is complete, this version shouldn't have any significant performance impact on the system (at least, I don't think), as it's just saving additional data for the same requests that are already happening.

For the next stage of it, fetching `features_previous` for every revision in the system, we'll need to look at how it affects production performance in terms of how long the update cycle takes; it might be fine, or it might require some work to speed up the bottlenecks.

PR #2943, opened 2019-06-05
Feature/my articles redesign

bwreid:
## What this PR does

Redesigns the My Articles section of the course page.

## Screenshots
Before:
image

image

After:
image

image

Would appreciate your thoughts in particular on making sure that the [sandbox is set to the correct user.](https://github.com/WikiEducationFoundation/WikiEduDashboard/compare/feature/my-articles-redesign?expand=1#diff-ecc3a4952d5d4317aaf4a588d93d917aR34)

### DIFF
--- a/app/assets/javascripts/components/overview/my_articles.jsx
+++ b/app/assets/javascripts/components/overview/my_articles.jsx
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router-dom';
 import AssignCell from '../students/assign_cell.jsx';
-import MyAssignment from './my_assignment.jsx';
+import MyAssignmentsList from './my_assignments_list.jsx';
 import { fetchAssignments } from '../../actions/assignment_actions';
 import { getFiltered } from '../../utils/model_utils';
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/overview/my_articles.jsx
```diff
@@ -26,27 +26,38 @@ export const MyArticles = createReactClass({
...
-          last={i === assignmentCount - 1}
-          current_user={this.props.current_user}
-          wikidataLabels={this.props.wikidataLabels}
-        />
-      );
+    const assignmentsWithSandboxes = assignments.map((article) => {
+      if (article.role === 0) return article;
+      const related = assignments.find((assignment) => {
+        return assignment.article_id === article.article_id
+          && assignment.user_id
```
bwreid: Remove this?

app/assets/javascripts/components/overview/my_articles.jsx
```diff
@@ -26,27 +26,38 @@ export const MyArticles = createReactClass({
...
+          && assignment.user_id !== user_id;
+      });
+      return {
+        ...article,
+        sandbox_url: `https://en.wikipedia.org/wiki/User:${related.username}/sandbox`
```
bwreid: Change this to dynamically change to the right article project/language.

app/assets/javascripts/components/overview/my_articles.jsx
bwreid: I remember what this was for. If we don't include this check, it will find the first assignment even if the `user_id` is `undefined`.

ragesoss: Ah. Carry on, then. This will get replaced with assignment status at some point soon-ish anyway.

PR #2950, opened 2019-06-06
Process all revisions of an article

khyatisoneji:


ragesoss:
This should be rebased on master, or cherry-picked to a fresh branch, so we can see just the new changes without the ones that were already merged.

PR #2957, opened 2019-06-09
Course Wikis [WIP]

AmitJoki:
## What this PR does
It creates the Model, View and Controller for CourseWikis which will be eventually used for multi-wiki support and the front-end component to back it up.
### Course Creator
image
### Details
image
### Article Finder
image

### Course Creator
image
### Details
image
### Article Finder
image


app/models/course_wiki.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class CourseWiki < ApplicationRecord
```
ragesoss: The naming convention in other join tables in the app is to pluralize both models: `CoursesWikis`.

app/models/course.rb
```diff
@@ -122,6 +122,8 @@ class Course < ApplicationRecord
 
   has_many :tags, dependent: :destroy
 
+  has_many :course_wikis
```
ragesoss: A course should also have many wikis through courses_wikis. That will simplify the controller as well; you should find the course and then do `course.wikis`.

app/controllers/course_wikis_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class CourseWikisController < ApplicationController
+  def update
+    course_id = params[:course_id]
+    wikis = course_wiki_params[:wikis]
+    CourseWiki.where(course_id: course_id).delete_all
```
ragesoss: I don't like this. It makes `created_at` and `updated_at` mostly useless. Records should only be deleted in the case where that wiki got removed.

app/controllers/course_wikis_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+    # clear existing course wikis to update with new ones
+    wikis.each do |wiki|
+      wiki_id = Wiki.get_or_create(language: wiki[:language], project: wiki[:project]).id
+      CourseWiki.create(course_id: course_id, wiki_id: wiki_id)
+    end
+    render json: {}, status: :ok
+  end
+
+  def wikis
```
ragesoss: I guess this is intended to be fetched by the frontend, similar to wht we do with `tags` and `campaigns` right now?

The alternative would be to fold this — as well as the update action — into the course update. The question is whether we want changing the wikis to be independent, or wrapped into the data flow for changing other course details (like we currently do with home_wiki). Since it will be a multi-select and not the atomic add/remove actions like with `tags` and `campaigns`, I think the preferable user experience, at least, will be waiting until the `Save` button is clicked to persist the change in wikis... although that could be done with two separate requests, I guess.

On the one hand, course update is already pretty complicated, but on the other hand, it's the one place to go for most changes in course options.

@bwreid what do you think of having a separate controller for CoursesWikis versus rolling wikis into the CoursesController#show and #update actions?

app/controllers/course_wikis_controller.rb
AmitJoki: Got it. I will refactor it so.

app/controllers/course_wikis_controller.rb
AmitJoki: Yes, the controller currently is modelled keeping in mind the "Save" is pressed and the wikis are handled. Once we reach a consensus, I will refactor it to suit our needs.

bwreid: Thanks for your patience on this. @AmitJoki My understanding is that we are accessing this information in a limited number of places, correct? Assuming this is true, I do think it would make the most amount of sense to roll this up into the CoursesController. Batching the update is likely a better pattern. 

But, are there cases where we are retrieving this information where we solely want the CourseWiki data? If so, we may want to think differently -- or leave the `show` action here.

app/controllers/courses_controller.rb
```diff
@@ -185,6 +188,11 @@ def set_course
     protect_privacy
   end
 
+  def set_courses_wikis
```
ragesoss: I don't think this is necessary; you can get the wikis from the `course`, just with `@course.wikis` (and Rails will cache that on the Course object so that if you need to call it more than once, it won't result in any extra queries).

db/schema.rb
```diff
@@ -176,6 +176,13 @@
     t.index ["user_id"], name: "index_commons_uploads_on_user_id"
   end
 
+  create_table "course_wikis", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
```
ragesoss: This is the old name, and shouldn't be in the diff.

db/schema.rb
```diff
@@ -242,6 +249,13 @@
     t.index ["user_id"], name: "index_courses_users_on_user_id"
   end
 
+  create_table "courses_wikis", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+    t.integer "course_id"
+    t.integer "wiki_id"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
```
ragesoss: To make the common queries as fast as possible, this should be indexed by course, and probably by wiki as well.

ragesoss: This should also include a uniqueness constraint: we don't want duplicate CoursesWikis records for the same course+wiki combination.

app/controllers/courses_controller.rb
```diff
@@ -237,6 +245,34 @@ def wiki_params
       .permit(:language, :project)
   end
 
+  def courses_wikis_params
+    params
+      .require(:course)
+      .permit(wikis: [:language, :project])
+  end
+
+  def update_courses_wikis
```
ragesoss: I think this method can be simplified a lot. We can lean on Rails for this kind of thing, because the CoursesWikis table represents a "has many, through" relationship. Ideally, we shouldn't need to get to the level ids.

For example, `course.wikis.delete(wiki)` and `course.wikis.push(wiki)` will add or remove a CoursesWikis record for provided `wiki`.

app/assets/javascripts/components/common/wiki_select.jsx
```diff
@@ -33,21 +33,36 @@ const WikiSelect = createReactClass({
     let wikis = [];
     if (this.props.wikis) {
       wikis = this.props.wikis.map((wiki) => {
+        wiki.language = wiki.language || 'www'; // for multilingual wikis language is null
         return {
-          value: wiki,
+          value: JSON.stringify(wiki),
```
ragesoss: What's `stringify` for here?

app/assets/javascripts/components/common/wiki_select.jsx
```diff
@@ -33,21 +33,36 @@ const WikiSelect = createReactClass({
...
+          wiki = wiki.map((w) => { return { value: JSON.parse(w.value), label: w.label }; });
+        } else {
+          const value = JSON.parse(wiki.value);
+          wiki = { label: wiki.label, value };
+        }
+        this.props.onChange(wiki);
+      } else {
+        this.props.onChange(this.props.multi ? [] : {});
+      }
+    };
```
ragesoss: This seems a little convoluted; is this because of the errors with removing the last selected value from a multi-select? If so, is there a simpler way to handle it?

lib/tasks/courses_wikis.rake
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+namespace :courses_wikis do
+  desc 'Creates CoursesWikis association for existing courses through revisions'
+  task migrate: :environment do
+    Rails.logger.debug 'Creating CoursesWikis associations'
+    Course.all.each do |course|
+      home_wiki_id = course.home_wiki.id # Home Wiki is already tracked
+      wiki_ids = course.revisions.pluck(Arel.sql('DISTINCT wiki_id')).uniq - [home_wiki_id]
```
ragesoss: I think this should be done with the same algorithm we use already for wiki_ids: home_wiki, plus the wikis for the course's assignments.

ragesoss:
After using this a bit and thinking on it, I've changed my mind about the behavior from earlier where the options don't load until you start typing, as some people may get confused if they just try to use the pulldown to find their wiki, with only the first 10 options present. If we load 10 options after the user types at least one character, it should be intuitive enough that they need to type to narrow the options further.

app/assets/javascripts/components/common/wiki_select.jsx
AmitJoki: React Select uses the value of value attribute as the key while rendering. Before the value was `wiki` an object, so the key was `wiki.toString()` which is always `[object Object]` so there were multiple collisions. JSON.stringify makes sure the value is a unique string and can also be parsed back to get the object.

app/assets/javascripts/components/common/wiki_select.jsx
AmitJoki: I think this is simple enough. The if block is parsing the JSON back to the object, the reasons of why we do this is given in my previous comment.

lib/tasks/courses_wikis.rake
AmitJoki: I had modified wiki_ids implementation that time so used the code from it and I didn't wanted home_wiki to be a part of courses_wikis. Now all this changes and I will edit it so.

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -86,6 +86,31 @@ const Details = createReactClass({
...
+    const prev_wiki = { ...this.props.course.home_wiki };
+    const updatedCourse = this.props.course;
+    updatedCourse.home_wiki.language = wiki.language;
+    updatedCourse.home_wiki.project = wiki.project;
+    if (!updatedCourse.wikis.filter(w => w.language === wiki.language && w.project === wiki.project).length) {
+      updatedCourse.wikis = [wiki, ...updatedCourse.wikis];
+    } else {
+      updatedCourse.wikis = [...updatedCourse.wikis];
+    }
+    // Remove automatically added previous wiki
```
ragesoss: It looks like this is duplicated from CourseForm. Maybe the pairing of home_wiki and tracked_wikis can be extracted into one component that can be used in both Details and CourseForm.

app/assets/javascripts/reducers/course.js
```diff
@@ -29,10 +29,15 @@ const initialState = {
   end: null,
   timeline_start: null,
   timeline_end: null,
+  language: 'en',
+  project: 'wikipedia',
```
ragesoss: Why do we need this data duplicated between with the home_wiki?

app/assets/javascripts/reducers/course.js
```diff
@@ -29,10 +29,15 @@ const initialState = {
...
+  home_wiki: { language: 'en', project: 'wikipedia' },
```
ragesoss: You can define a `defaultWiki` const above here, and then use it in both home_wiki and `wikis`.

app/controllers/courses_controller.rb
```diff
@@ -16,6 +16,8 @@ class CoursesController < ApplicationController
   before_action :require_permissions, only: %i[notify_untrained
                                                delete_all_weeks]
 
+  after_action :ensure_home_wiki_in_courses_wikis, only: %i[create update]
```
ragesoss: Using `after_action` here may be part of the problem you're having, as I think that happens after the view gets rendered, so the JSON representation of the course that gets passed back to the client won't reflect any changes that come from that callback. I suggest calling it explicitly from `update_courses_wikis` instead of using a callback.

app/controllers/courses_controller.rb
```diff
@@ -312,6 +341,14 @@ def verify_edit_credentials
     yield
   end
 
+  # Makes sure that the home wiki
+  # is always a part of courses wikis.
+  def ensure_home_wiki_in_courses_wikis
```
ragesoss: I was misunderstanding the place you had implemented this validation; I thought it was done on the model, with a `before_save` callback on the model. Doing it here doesn't give us much extra assurance beyond the frontend validation; it won't save us from mistakes elsewhere in the code that create or update a course but fail to include the home_wiki.

app/assets/javascripts/reducers/course.js
AmitJoki: Because in Course Creator when we change the home wiki, we need to add the current home wiki and remove the previous home wiki. In this case previous home wiki would be undefined as language and project didn't have a value initially.

app/controllers/courses_controller.rb
AmitJoki: Hm that's a valid point. I'll patch it up.

app/controllers/courses_controller.rb
AmitJoki: We don't use the changes from ater_action but yes, I will include it in update_courses_wikis.

app/controllers/courses_controller.rb
```diff
@@ -237,6 +239,32 @@ def wiki_params
...
   end
 
+  def courses_wikis_params
+    params
+      .require(:course)
+      .permit(wikis: [:language, :project])
+  end
+
+  def update_courses_wikis
+    @course.ensure_home_wiki_in_courses_wikis
```
ragesoss: This should not be necessary, as this method is only called immediately after a course record was saved anyway, and now we have that `before_save` hook.

app/models/course.rb
```diff
@@ -56,6 +56,8 @@
 
 #= Course model
 class Course < ApplicationRecord
+  before_save { ensure_home_wiki_in_courses_wikis }
```
ragesoss: This sould be in the `callbacks` section, and should be done with a symbol instead of a block to match the other callbacks.

app/models/course.rb
```diff
@@ -417,6 +422,13 @@ def reorder_weeks
     end
   end
 
+  # Makes sure that the home wiki
+  # is always a part of courses wikis.
+  def ensure_home_wiki_in_courses_wikis
+    return if nil?
```
ragesoss: What's this for? The implicit `self` receiving `.nil?` would be a `Course` object, so it would never be nil, right?

spec/fixtures/wikis.yml
```diff
@@ -11,3 +11,13 @@ default:
...
+
+fr:
+  id: 3
+  language: fr
+  project: wikipedia
+
+de:
+  id: 4
+  language: de
+  project: wikipedia
```
ragesoss: I would rather use `stub_wiki_validation` in more places, instead of adding additional wikis to the fixtures.

spec/controllers/courses_controller_spec.rb
```diff
@@ -140,6 +140,25 @@
...
+      put "/courses/#{course_params[:slug]}", params: params, as: :json
+      expect(course.wikis.count).to eq(3)
+
+      course_params[:wikis].pop
+      put "/courses/#{course_params[:slug]}", params: params, as: :json
+      expect(course.wikis.count).to eq(2)
+
+      course_params[:wikis][0][:language] = 'fr'
+      put "/courses/#{course_params[:slug]}", params: params, as: :json
+      expect(course.wikis.count).to eq(2)
```
ragesoss: This is replacing one wiki with another?

spec/controllers/courses_controller_spec.rb
```diff
@@ -140,6 +140,25 @@
       end
     end
 
+    it 'updates the wikis' do
+      course_params[:wikis] = [{ language: 'de', project: 'wikipedia' }]
```
ragesoss: The idea here is that this is testing the server-side handling of an update where the wikis don't include the home_wiki? So even though only one wiki was received, we expect the home wiki to also be there? If so, that's not obvious enough from the test itself.

I recommend either adding comments, or separating these into four different tests with titles that explain exactly what each one is trying to test for.

spec/features/course_cloning_spec.rb
```diff
@@ -104,5 +105,6 @@
     expect(new_course.submitted).to eq(false)
     expect(new_course.user_count).to be_zero
     expect(new_course.article_count).to be_zero
+    expect(new_course.wikis.count).to eq(1)
```
ragesoss: I think it would be better to extend this test so that the original course has more than one wiki, and make sure the set of tracked wikis is cloned as expected.

spec/support/request_helpers.rb
```diff
@@ -267,7 +267,8 @@ def stub_wiki_validation
       'en.wikinews.org',
       'pl.wikiquote.org',
       'de.wikiversity.org',
-      'commons.wikimedia.org'
+      'commons.wikimedia.org',
+      'en.wikipedia.org'
```
ragesoss: This should not be necessary, as en.wikipedia.org is the one wiki that was already present in the fixtures, so nothing in the test suite should attempt to validate it.

lib/course_clone_manager.rb
```diff
@@ -34,6 +29,21 @@ def clone!
 
   private
 
+  def clone_relevant_information
```
ragesoss: I don't like this refactor, because not all the steps copied over are really cloning information.

If this is just to satisfy rubocop by not having an extra line, maybe `clear_meeting_days_and_due_dates` could be moved to the end of `duplicate_timeline`, as it's basically part of that step conceptually.

app/controllers/courses_controller.rb
```diff
@@ -237,6 +239,32 @@ def wiki_params
...
+    multi_wikis = courses_wikis_params[:wikis]
+    return if multi_wikis.nil?
+
+    # Used to differentiate deleted wikis
+    old_wikis = @course.wikis.to_a
```
ragesoss: You don't need to convert an ActiveRecord relation to an array in most cases; you can do all the same things with the relation itself, including add or subject another relation or array to get an array back.

app/controllers/courses_controller.rb
```diff
@@ -237,6 +239,32 @@ def wiki_params
...
+
+    # Create an association for each new wiki
+    multi_wikis.each do |wiki|
+      new_wiki = Wiki.get_or_create(language: wiki[:language], project: wiki[:project])
+      new_wikis.push(new_wiki)
+      @course.wikis.push(new_wiki) unless old_wikis.include? new_wiki
+    end
+
+    # Delete removed wikis except home wiki
+    @course.wikis.delete(old_wikis - new_wikis - [@course.home_wiki])
```
ragesoss: There's probably a simpler way to do the adding and removing. ActiveRecord has a lot of support for join models like `CoursesWikis`, such that you can do something like `@course.wikis = set_of_wikis`, and it create or delete records as needed to make that happen.

ragesoss: You can also do `@course.update(wikis: set_of_wikis)` and will do the same thing.

So maybe take the wiki params, map them to a set of actual Wiki records with `Wiki.get_or_create`, and then do the update.

ragesoss: Hmm... thinking more on this along with the `ensure_home_wiki_in_courses_wikis`, here's what I think should happen...

Since changing the home_wiki or any other wikis will only happen through `CoursesController#update`, and you already have some special handling of it, the actual logic for updating the wikis could be moved to `Course` itself, with a new method like `#update_wikis`... which could do a `update(wikis: updated_wikis)` followed by `ensure_home_wiki_in_courses_wikis`... and then `ensure_home_wiki_in_courses_wikis` could just be a `before_create` callback instead of `before_save`. That way it would not add any extra queries to the updates that happen in other contexts.

lib/course_clone_manager.rb
AmitJoki: Yes this was indeed just to satisfy the rubocop. I will fix this up.

spec/support/request_helpers.rb
AmitJoki: It is needed as ensure_wiki_exists calls wiki_api which sends a query, which is stubbed by stub_wiki_validation and hence en.wikipedia.org needs to exist there, else VCR error comes up.

spec/fixtures/wikis.yml
AmitJoki: I've used these now in course_cloning_spec.rb

app/models/course.rb
AmitJoki: Yup! I migrated this code from controller and forgot this bit.

spec/support/request_helpers.rb
ragesoss: That doesn't sound right. `ensure_wiki_exists` only gets called if you try to find a wiki that isn't already in the DB, but the fixture file gets loaded directly when the test environment boots, and bypasses that for the wikis in the fixture file.

lib/tasks/courses_wikis.rake
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+namespace :courses_wikis do
+  desc 'Creates CoursesWikis association for existing courses through revisions'
+  task migrate: :environment do
+    Rails.logger.debug 'Creating CoursesWikis associations'
+    Course.all.each do |course|
+      course.wikis.push(Wiki.find(course.wiki_ids))
```
ragesoss: This needs to be idempotent. Currently, it will fail (because of the uniqueness constraint) if it is run when there are already some CoursesWikis in the DB or if it is run more than once.

ragesoss:
A few further changes:

Instead of switching on whether it's Wiki Education or P&E Dashboard, we need to make the switch more detailed. I guess, basically, we should make admins always be able to see/change the home wiki and tracked wikis.

On P&E Dashboard, the Details section should also list the tracked wikis outside of 'Edit Details' mode.

app/controllers/courses_controller.rb
```diff
@@ -237,6 +239,21 @@ def wiki_params
...
 
+  def courses_wikis_params
+    params
+      .require(:course)
+      .permit(wikis: [:language, :project])
+  end
+
+  def update_courses_wikis
+    multi_wikis = courses_wikis_params[:wikis]
+    return if multi_wikis.nil?
```
ragesoss: In what circumstances would mulit_wikis be nil?

app/models/course.rb
```diff
@@ -417,6 +426,13 @@ def reorder_weeks
     end
   end
 
+  # Makes sure that the home wiki
+  # is always a part of courses wikis.
+  def ensure_home_wiki_in_courses_wikis
+    return if home_wiki.nil?
```
ragesoss: home_wiki shouldn't ever be nil, I don't think. Is this necessary?

app/models/courses_wikis.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
ragesoss: Needs `annotate` info.

ragesoss:
Is it possible to remove the `x` from the wiki selector component, for the home wiki only? That way, it will be clear to users that the home wiki cannot be removed. The current UI, where you can click it but it does nothing, will be a little bit confusing for users.

ragesoss:
Wes and I reviewed this together, and also deployed and ran the migrate task on dashboard-testing.wikiedu.org. We opend a few issues we noticed along the way that are mostly independent of this PR. Take a look in particular at #3025, which you might take on at some point to reduce duplication and clean up some of the code from this PR.

The migrate task took about 12 minutes on a recent copy of the P&E Dashboard database, because of how wiki_ids is querying the Revisions table for each course. The Wiki Education Dashboard one will take maybe twice as long, most likely. That's fine, although definitely too long to include as part of the travis deployment (as it will cause the build to time out and then cut the connection midway through).

This is very close to being ready! Great work.

app/models/course.rb
AmitJoki: It is nil in the before_create callback because, it is before creation and there isn't a home_wiki yet.

app/controllers/courses_controller.rb
AmitJoki: The course_params have partial parameters and wikis may not always be present. It was the case in courses_controller_spec.rb

AmitJoki:
Fixes #3025 among other things. The failing spec: admin_role_spec.rb is passing on my local machine. May be restart the build?

ragesoss:
The list of wikis should be rendered in a similar style to the other data in Details when it's not in edit mode, not as an input. Also, the home wiki can be hidden in this case, since it's duplicate data from the tracked wikis.

view mode

app/assets/stylesheets/modules/_course.styl
```diff
@@ -205,3 +205,9 @@ rating(color)
 
     .details-form
       display block
+
+.multi-wiki [class$='multiValue']:first-of-type svg
```
ragesoss: Nice solution. I guess react-select doesn't support per-option configuration of whether to show the delection UI?

ragesoss: Can you get rid of the hover styling as well for this?

Also, I think it would be good to hide the dropdown for the home_wiki until the user begins typing, since we don't show any options in the dropdown before then.

no svg but still hover

app/controllers/courses_controller.rb
```diff
@@ -236,6 +236,7 @@ def initial_campaign_params
   def wiki_params
     params
       .require(:course)
+      .fetch(:home_wiki, {})
```
ragesoss: Nice!

PR #2968, opened 2019-06-10
Disable W3C requests for Selenium

bwreid:
## What this PR does
This should fix the errors we're receiving on Travis CI. Ideally, we would not need to disable this but, eh.

To locally upgrade your version of Chrome, you should be able to run `chromedriver-update`. You can also specify a version number if you choose.


ragesoss:
Is there related documentation that you can link to in the commit message? Otherwise, :shipit:

bwreid:
@ragesoss Ah, thank you for the reminder. I'll add the Stackoverflow answer I found that led me to the right solution and then :shipit: ✨

PR #2970, opened 2019-06-11
add references sum in courses_users

khyatisoneji:


khyatisoneji:
Here we want to store reference sum in courses_users model. Just like we are storing the character count for each courses_users relationship.
This will help us in calculating the references added by an individual faster than going through all revisions of the user

bwreid:
This looks great to me, Khyati

app/models/course_data/courses_users.rb
```diff
@@ -118,6 +119,13 @@ def character_sum(revisions, namespace)
       .sum(:characters) || 0
   end
 
+  def references_sum(revisions)
+    revisions
+      .where(articles: { namespace: Article::Namespaces::MAINSPACE, deleted: false })
+      .to_a
+      .sum(&:references_added) || 0
```
ragesoss: I think both the `.to_a` and the `|| 0` are not necessary.

Even though `revisions` is an ActiveRecord relation (rather than an array), the `.sum` method will fall back to the array method if it receives a block argument rather than a column name: https://github.com/rails/rails/blob/b9ca94caea2ca6a6cc09abaffaad67b447134079/activerecord/lib/active_record/relation/calculations.rb#L86

`Array#sum` defaults to 0 for an empty array.

app/models/course_data/courses_users.rb
```diff
@@ -101,6 +101,7 @@ def update_cache
     self.character_sum_ms = character_sum(revisions, Article::Namespaces::MAINSPACE)
     self.character_sum_us = character_sum(revisions, Article::Namespaces::USER)
     self.character_sum_draft = character_sum(revisions, Article::Namespaces::DRAFT)
+    self.refs_sum_ms = references_sum(revisions)
```
ragesoss: You should run `annotate` so that the comments at the top of the file get updated with the new attribute as well.

ragesoss:
Looks like this branch got messy with that merge. I suggest starting a new branch and using cherry-pick to apply your the courses_users changes.

PR #2971, opened 2019-06-11
process all revisions in ores score importer

khyatisoneji:


khyatisoneji:
As part of this pull request we want to process all revisions for calculating wp10_previous and features_previous values. 
Earlier we were only calculating these values for first revisions

lib/importers/ores_scores_before_and_after_importer.rb
```diff
@@ -20,15 +20,15 @@ def initialize(articles_courses:)
...
     @articles_courses.each do |ac|
-      first_revs << ac.all_revisions.where(wiki: wiki).order('date ASC').first
+      all_revs << ac.all_revisions.where(wiki: wiki)
       last_revs << ac.all_revisions.where(wiki: wiki).order('date ASC').last
     end
 
-    first_revs.select! { |rev| rev.present? && rev.wp10_previous.nil? }
-    RevisionScoreImporter.new(wiki.language).update_previous_wp10_scores first_revs
+    all_revs.select! { |rev| rev.present? && rev.wp10_previous.nil? }
+    RevisionScoreImporter.new(wiki.language).update_previous_wp10_scores all_revs
```
bwreid: Does this work? It looks as though with the changes above we're now adding a series of collections to `all_revs`. That is, I imagine `all_revs` is going to look like:
```
[ [ Revisions ], [ Revisions ], [ Revisions ] ]
```
This would be a different format than the `update_previous_wp10_scores` method has taken before.

ragesoss:
I think that instead of changing `OresScoresBeforeAndAfterImporter`, the way we should do this is to get rid of `OresScoresBeforeAndAfterImporter` and change `RevisionScoreImporter` to pull in this data for all revisions. See `RevisionScoreImporter#update_all_revision_scores_for_articles`.

PR #3004, opened 2019-06-20
Updating scores for all revisions

khyatisoneji:
Updated the #update_all_revision_scores_for_articles to take all revisions to update wp10_previous and features_previous values using #update_previous_wp10_scores function

### DIFF
--- a/daily_update_spec.rb
+++ b/daily_update_spec.rb
@@ -1,0 +1,69 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require "#{Rails.root}/lib/data_cycle/daily_update"
+
+describe DailyUpdate do
+  before do
+    create(:course, start: '2015-03-20', end: 1.month.from_now,
+                    flags: { salesforce_id: 'a0f1a9063a1Wyad' })
+    old_course = create(:course, slug: 'old', start: '2015-03-20', end: '2015-04-20',
+                                 flags: { salesforce_id: 'b0f1a9063a1Wyad' })
+    old_course.campaigns << Campaign.first
+  end
+
+  describe 'on initialization' do
+    it 'calls lots of update routines' do
+      VCR.use_cassette 'revision_scores/deleted_revision' do
+        expect(UserImporter).to receive(:update_users)
+        expect(AssignedArticleImporter).to receive(:import_articles_for_assignments)
+        expect(ArticlesCoursesCleaner).to receive(:rebuild_articles_courses)
+        expect(RatingImporter).to receive(:update_all_ratings)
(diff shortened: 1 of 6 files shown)


daily_update_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+  context 'when a pid file is present' do
+    it 'deletes the pid file for a non-running process' do
+      allow_any_instance_of(described_class).to receive(:create_pid_file)
+      allow_any_instance_of(described_class).to receive(:run_update)
+      File.open('tmp/batch_update_daily.pid', 'w') { |f| f.puts '123456789' }
+      described_class.new
+      expect(File.exist?('tmp/batch_update_daily.pid')).to eq(false)
+    end
+  end
+end
```
ragesoss: Looks like this file got accidentally copied to a new location.

revision_score_importer_spec.rb
```diff
@@ -0,0 +1,194 @@
...
+    end
+  end
+
+  context 'for a wiki without the articlequality model' do
+    it 'raises an error' do
+      expect { described_class.new('zh').update_revision_scores }
+        .to raise_error(RevisionScoreImporter::InvalidWikiError)
+    end
+  end
+end
```
ragesoss: This file also.

spec/lib/data_cycle/daily_update_spec.rb
```diff
@@ -14,20 +14,39 @@
...
-      expect(OresScoresBeforeAndAfterImporter).to receive(:import_all)
-      expect(UploadImporter).to receive(:find_deleted_files)
-      expect_any_instance_of(OverdueTrainingAlertManager).to receive(:create_alerts)
-      expect(PushCourseToSalesforce).to receive(:new)
-      expect(UpdateCourseFromSalesforce).to receive(:new)
-      expect(Raven).to receive(:capture_message).and_call_original
-      update = described_class.new
-      sentry_logs = update.instance_variable_get(:@sentry_logs)
-      expect(sentry_logs.grep(/Pushing course data to Salesforce/).any?).to eq(true)
+      VCR.use_cassette 'revision_scores/deleted_revision' do
```
ragesoss: It's good to test for the behavior of revision score importer, but the `DailyUpdate` spec is not the right place to do that.

Ideally, DailyUpdate's own spec is just testing the behavior of `DailyUpdate` itself, and won't break because of implementation details of `RevisionScoreImporter`... and `RevisionScoreImporter`'s specs should break when its behavior changes, rather than relying on a test of a different class to check some of its behavior.

PR #3008, opened 2019-06-22
Fix the race condition

AmitJoki:
## What this PR does
Fixes the race condition between Overview Handler and Timeline Handler.

### DIFF
--- a/app/views/courses/course.json.jbuilder
+++ b/app/views/courses/course.json.jbuilder
@@ -37,6 +37,8 @@ json.course do
   json.syllabus @course.syllabus.url if @course.syllabus.file?
   json.updates average_delay: @course.flags['average_update_delay'],
                last_update: @course.flags['update_logs']&.values&.last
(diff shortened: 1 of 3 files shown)


app/views/courses/course.json.jbuilder
```diff
@@ -37,6 +37,8 @@ json.course do
   json.syllabus @course.syllabus.url if @course.syllabus.file?
   json.updates average_delay: @course.flags['average_update_delay'],
                last_update: @course.flags['update_logs']&.values&.last
+  json.loading false
```
ragesoss: This is a strictly frontend concern, and should be handled completely in the reducer rather than here. That might just mean setting it in RECEIVE_COURSE, maybe RECEIVE_COURSE_CLONE, and maybe CREATED_COURSE... although I think for the last one, at least, it might not be necessary.

ragesoss:
merged

PR #3010, opened 2019-06-24
Added References Count

khyatisoneji:
Added References added to 
URL:  /explore, /campaigns , /courses, /articles.

Screenshots:
	Added references count to following pages: 

a. Programs and events dashboard homepage
Screenshot from 2019-06-26 18-55-37

b.  Campaign overview
image

c. Campaign programs
image

d. Campaign articles
image

e. Campaign users
image

f. Courses home
image

g. Courses students
image

h. Courses articles
image

i. User profile
image


ragesoss:
@khyatisoneji please add screenshots!

app/models/wiki_content/revision.rb
```diff
@@ -71,6 +71,6 @@ def plagiarism_report_link
   def references_added
     current_refs = features['feature.wikitext.revision.ref_tags'] || 0
     prev_refs = features_previous['feature.wikitext.revision.ref_tags'] || 0
-    current_refs - prev_refs
+    self.references = current_refs - prev_refs
```
ragesoss: Per my note on the migration, I think sticking with the previous implementation here is better.

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -56,7 +60,7 @@ const Article = createReactClass({
           </div>
         </td>
         <td className="desktop-only-tc">{this.props.article.character_sum}</td>
-        <td className="desktop-only-tc">{this.props.article.references_count}</td>
+        <td className="desktop-only-tc">{refCount}</td>
```
ragesoss: This works, but since `0` is falsy in javascript, you could take a shortcut here and do `this.props.article.references_count || ''`.

app/assets/javascripts/components/revisions/revision.jsx
```diff
@@ -24,7 +24,7 @@ const Revision = ({ revision, index, wikidataLabel, course, setSelectedIndex, la
       </td>
       <td className="desktop-only-tc">{revision.revisor}</td>
       <td className="desktop-only-tc">{revision.characters}</td>
-      <td className="desktop-only-tc">{revision.references}</td>
+      <td className="desktop-only-tc">{revision.features['feature.wikitext.revision.ref_tags'] || 0 - revision.features_previous['feature.wikitext.revision.ref_tags'] || 0}</td>
```
ragesoss: This probably makes sense to do either on the server, at the point of serializing the revisions to json, or if it's to be done client-side, do it in a selector.

I think server-side is preferable, as you can rely on the `Revisions` method that you already wrote.

config/locales/en.yml
```diff
@@ -169,6 +169,7 @@ en:
       list: "%{class}-class — a list rather than a conventional article"
       ?: Unrated — an article that hasn't yet been rated on Wikipedia's quality scale
       does_not_exist: This article does not exist on Wikipedia
+    references_doc: It is the number of citations added in the article.This data comes for the ORES article quality model, and is only available for some languages.
```
ragesoss: "This is the number of reference tags added to articles, and can include multiple references to the same source. The data comes from the ORES article quality model, and is only available for some languages."

ragesoss:
i18n problem with this message: 
missing translatoin

It's looking for the courses_generic variant of a message; it can probably just use a single version for all cases instead of applying a course prefix.

ragesoss:
@khyatisoneji can rebase this or merge `master` into it?

The schema.rb file should use the highest value for the schema version (which means, `rake db:migrate` will apply all un-applied migrations up to and including the one specified, so your changes will be applied on the next deploy).

The courses_spec.rb had a little refactor that you'll need to merge with.

The one other thing to change is to use an abbreviation for the references count when it gets very high, like we do for other stats like Total Edits:

appreviated value for refs added

Here, the rendered value should be something like "3.34K" instead of "3344", using the same conversion as the other stats.

PR #3020, opened 2019-06-26
Test tracking of wikis from CoursesWikis.

AmitJoki:
## What this PR does
Tests whether #2957 is working alright after migration of course.wiki_ids to course.wikis


AmitJoki:
@ragesoss @bwreid after the deployment, this PR actually uses course.wikis instead of wiki_ids, so when testing the #2957 you can may be include these commits as well.

PR #3031, opened 2019-06-29
Added References Added to Course and campaign stats

khyatisoneji:
Updated CSV files:

Campaign overview,
Course overview,
Course edits,
Course editor,
Course articles

### DIFF
--- a/lib/analytics/course_articles_csv_builder.rb
+++ b/lib/analytics/course_articles_csv_builder.rb
@@ -31,18 +31,12 @@ class CourseArticlesCsvBuilder
     @course.all_revisions.includes(article: :wiki).map do |edit|
       article_edits = @articles_edited[edit.article_id] || new_article_entry(edit)
       article_edits[:characters][edit.mw_rev_id] = edit.characters
-      references_update(article_edits, edit)
+      article_edits[:references] = edit.references_added
       article_edits[:new_article] = true if edit.new_article
       # highest view count of all revisions for this article is the total for the article
       article_edits[:views] = edit.views if edit.views > article_edits[:views]
       @articles_edited[edit.article_id] = article_edits
     end
-  end
-
-  def references_update(article_edits, edit)
-    current_refs = edit.features['feature.wikitext.revision.ref_tags'] || 0
-    prev_refs = edit.features_previous['feature.wikitext.revision.ref_tags'] || 0
-    article_edits[:references] = current_refs - prev_refs
   end
 
   def new_article_entry(edit)
@@ -51,8 +45,7 @@ class CourseArticlesCsvBuilder
       new_article: false,
       views: 0,
       characters: {},
-      features: {},
-      features_previous: {},
+      references: {},
       title: article.title,
       namespace: article.namespace,
       url: article.url,
--- a/lib/analytics/course_articles_csv_builder.rb
+++ b/lib/analytics/course_articles_csv_builder.rb
@@ -31,11 +31,18 @@ class CourseArticlesCsvBuilder
     @course.all_revisions.includes(article: :wiki).map do |edit|
       article_edits = @articles_edited[edit.article_id] || new_article_entry(edit)
       article_edits[:characters][edit.mw_rev_id] = edit.characters
+      references_update(article_edits, edit)
       article_edits[:new_article] = true if edit.new_article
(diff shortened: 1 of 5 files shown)


khyatisoneji:
Tests would break, as it does not have the changes made in courses_users, articles, courses, articles_courses to add references_count

ragesoss:
You should be able to rebase this now.

lib/analytics/course_edits_csv_builder.rb
```diff
@@ -37,8 +38,14 @@ def row(revision)
...
-    row << revision.characters
+    add_character_references(revision, row)
     row << revision.new_article
     row << revision.system
   end
 end
+
+def add_character_references(revision, row)
+  row << revision.characters
+  row << revision.features['feature.wikitext.revision.ref_tags'] || 0 -
```
ragesoss: Use `revision.reference_added` here instead of reimplementing the the calculation.

lib/analytics/course_articles_csv_builder.rb
```diff
@@ -31,19 +31,28 @@ def set_articles_edited
     @course.all_revisions.includes(article: :wiki).map do |edit|
       article_edits = @articles_edited[edit.article_id] || new_article_entry(edit)
       article_edits[:characters][edit.mw_rev_id] = edit.characters
+      references_update(article_edits, edit)
```
ragesoss: This should also be done with `edit.references_added` instead of reimplementing the calculation.

This is a confusing section of code, but it will be easier to follow if the assignment is done inline like with views and characters, like `article_edits[:references] = edit.references_added`.

lib/analytics/course_edits_csv_builder.rb
```diff
@@ -37,8 +38,13 @@ def row(revision)
     row << revision.article.full_title
     row << revision.url
     row << revision.user.username
-    row << revision.characters
+    add_character_references(revision, row)
```
ragesoss: I think this is clearer as keeping one line per piece of data. Doing just these two as their own method makes it harder to follow what's going on, and those two items are any more closely related than some of the other items.

khyatisoneji: Rubocop was giving warning on Assignment Branch Assignment
So had to make a function and add it

ragesoss: I think disabling that rubocop rule for the method is a better option for cases like this.

The idea behind the rule is to avoid methods that have too much complexity and are hard to understand, but in cases like this where we have a really clear and repetitive pattern — add something to a row, once for each column — it's more clear and simple to break the rule than try to work around it.

khyatisoneji: Guess, it was giving that warning when was calculating it using features and features_previous and not by calling function

lib/analytics/course_articles_csv_builder.rb
```diff
@@ -87,6 +90,12 @@ def character_sum(article_data)
     end
   end
 
+  def references_count(article_data)
+    article_data[:references] do |sum, references|
+      references ? sum + references : sum
```
ragesoss: I don't understand what's going on here. What does `article_data[:references]` look like, and where does the `sum` come in?

khyatisoneji: Sorry, was just checking it, forgot to remove it

lib/analytics/course_articles_csv_builder.rb
```diff
@@ -31,6 +31,7 @@ def set_articles_edited
     @course.all_revisions.includes(article: :wiki).map do |edit|
       article_edits = @articles_edited[edit.article_id] || new_article_entry(edit)
       article_edits[:characters][edit.mw_rev_id] = edit.characters
+      article_edits[:references] = edit.references_added
```
ragesoss: Whoops... I merged this a little too quickly. This will overwrite the previous value, right? Instead of providing a total across all edits?

ragesoss: I'll poke at it today and do a fix if necessary.

khyatisoneji:
Yes!
I forgot why I was doing the sum in the first place, I thought it would
just show the total current number of ref tags so I removed it, but it
slipped my mind that it's calculating the references added per edit/revision
Sorry about that 😅

PR #3050, opened 2019-07-03
Issue:2530  Added Radio button for Academic System with Semester Default

khyatisoneji:
Made changes in Course_form.jsx to add Academic System 

**Screenshots**

**Before**:
**Course Form:**
image

**Details:**
image

**After**:
**Course Form:**
image

**Details:**
image

### DIFF
--- a/app/assets/javascripts/components/common/academic_system.jsx
+++ b/app/assets/javascripts/components/common/academic_system.jsx
@@ -23,11 +23,12 @@ const AcademicSystem = createReactClass({
     const academic_system = [];
     for (i = 0; i < options.length; i += 1) {
       academic_system.push(
-        <label key={options[i]}>
+        <label className="radio-inline">
           <input
             type="radio"
             name="academic_system"
             value={options[i]}
+            style={{ display: 'inline-block', width: '30px' }}
             defaultChecked={this.state.selectedOption === options[i]}
             onChange={this.handleOptionChange}
           />
--- a/app/assets/javascripts/components/common/academic_system.jsx
+++ b/app/assets/javascripts/components/common/academic_system.jsx
@@ -5,6 +5,7 @@ const createReactClass = require('create-react-class');
 const AcademicSystem = createReactClass({
 
   getInitialState: function () {
+    this.props.updateCourseProps({ academic_system: this.props.value || 'Semester' });
     return {
       selectedOption: this.props.value || 'Semester'
     };
--- a/app/assets/javascripts/components/common/academic_system.jsx
+++ b/app/assets/javascripts/components/common/academic_system.jsx
@@ -5,9 +5,9 @@ const createReactClass = require('create-react-class');
 const AcademicSystem = createReactClass({
 
   getInitialState: function () {
-    this.props.updateCourseProps({ academic_system: this.props.value || 'Semester' });
+    this.props.updateCourseProps({ academic_system: this.props.value || 'semester' });
     return {
-      selectedOption: this.props.value || 'Semester'
+      selectedOption: this.props.value || 'semester'
     };
   },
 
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/components/course_creator/course_form.jsx
```diff
@@ -7,6 +7,36 @@ import CourseUtils from '../../utils/course_utils.js';
 import selectStyles from '../../styles/select';
 import WikiSelect from '../common/wiki_select.jsx';
 
+const createReactClass = require('create-react-class');
+
+const Application = createReactClass({
```
ragesoss: This needs a better name, specific enough that seeing only the component name is enough that I know what it is for. It should probably be in a separate file as well.

app/assets/javascripts/components/common/academic_system.jsx
```diff
@@ -0,0 +1,33 @@
...
+  handleChange: function () {
+    this.props.onChange(this.props.myValue);
+  },
+
+  render: function () {
+    return (
+      <div>
+        <label>
+          <input type = "radio" name="academic_system" defaultChecked={this.state.selectedOption === 'Semester'} onChange={this.handleOptionChange}/>
+          Semester
```
ragesoss: This should probably have an `Other` option as well. I suggest putting the options in an array, and generating the inputs dynamically, so that it will be easy to edit the options later.

ragesoss:

This doesn't yet propagate to the server or return to the client, right?

khyatisoneji:
Not yet
Have to add another column specifying which system does the school/university follows

ragesoss:
Okay. I think that should be done as a flag, rather than a new database
column.

-Sage

khyatisoneji:
It depends 
I have a doubt 
Should this detail about the system followed be shown along with other details of course 
If so then we will need to add another column if not then can do it with flag

khyatisoneji:
@ragesoss can you please check this? I think this is good to go

ragesoss:
Thanks @khyatisoneji.

The data round-trip looks like it's mostly working. It looks like the flag remains `nil` until it gets changed to `Quarter` (and then will take the value `Semester`) if changed back. But once set to a value, it behaves as expected.

The one key change before this could be deployed is to add it to the CourseCreator form (only for ClassroomProgramCourse). The main point where we want to collect this information is when the course is being set up (and it's unlikely that it will need to be edited afterwards, except in rare cases).

ragesoss:
One more tweak that didn't occur to me until now... It should probably have options 'semester', 'quarter', and 'other'. There are some 'trimester' systems that wouldn't be either of the first two... and probably other rare systems as well.

Also, I think lower case is the most appropriate format.

ragesoss:
The choice of system still shows up in Edit Details even when it's not a ClassroomProgramCourse.

bwreid:
@khyatisoneji This looks good to me! I will likely leave this up for a bit before merging until Sage returns, but I think it's good to go.  :)

khyatisoneji:
Sure! Tysm 🙂

app/assets/javascripts/components/user_profiles/course_details.jsx
```diff
@@ -23,6 +23,10 @@ const CourseDetails = createReactClass({
               <div className="course-details_title">{I18n.t('courses.term')}</div>
               <div className="course-details_value">{course.course_term}</div>
             </div>
+            <div className="col">
+              <div className="course-details_title">{I18n.t('courses.academic_system')}</div>
+              <div className="course-details_value">{course.academic_system}</div>
```
ragesoss: This data doesn't come from course.json.jbuilder, so academic system is not included.

I don't think this is especially useful info for the user profile page, though, so rather than add it to the json view this uses, we should just cut it.

ragesoss: I'm adding this change and will merge this stuff shortly.

ragesoss:
merged.

PR #3051, opened 2019-07-03
Add wikidata namespaces support

AmitJoki:
## What this PR does
Adds the wikidata namespaces support.

After:

image


ragesoss:
Does that screensot include Q and P entries?

This looks good, but I think one further change would help a lot, which is to put the title (without namespace) in parentheses in cases where it's showing the label instead of the Q/P number. Probably it could be styled similarly to the de-emphasized look on Wikidata pages.

prop and label
title and label

AmitJoki:
Yes the screenshot includes both P and Q entries. The Mapillary ID is P and the rest is Q mostly.

PR #3052, opened 2019-07-03
Issue: 2977 Made Revisions csv file for campaign and added mw_rev_id

khyatisoneji:
Made changes in course_csv_builder.rb and campaign_csv_builder.csv to create a revision CSV file for campaigns
Also, made some frontend changes to add revisions link to download revision CSV file for campaign

### DIFF
--- a/lib/analytics/course_csv_builder.rb
+++ b/lib/analytics/course_csv_builder.rb
@@ -25,7 +25,6 @@ class CourseCsvBuilder
     articles_created
     bytes_added
     total_edits
-    revision_id
     mainspace_edits
     article_talk_edits
     userspace_edits
@@ -52,7 +51,10 @@ class CourseCsvBuilder
     row << @course.article_count
     row << @course.new_article_count
     row << @course.character_sum
-    revisions_rows(row)
+    row << @course.revision_count
+    row << revisions_by_namespace(Article::Namespaces::MAINSPACE)
+    row << revisions_by_namespace(Article::Namespaces::TALK)
+    row << revisions_by_namespace(Article::Namespaces::USER)
     row << @course.view_sum
     row << @course.upload_count
     row << @course.uploads_in_use_count
@@ -71,14 +73,6 @@ class CourseCsvBuilder
     else
       CSV_HEADERS
     end
-  end
-
-  def revisions_rows(row)
-    row << @course.revision_count
-    row << revisions_id
-    row << revisions_by_namespace(Article::Namespaces::MAINSPACE)
-    row << revisions_by_namespace(Article::Namespaces::TALK)
-    row << revisions_by_namespace(Article::Namespaces::USER)
   end
 
   def generate_csv
@@ -105,10 +99,6 @@ class CourseCsvBuilder
     # An editor counts as retained if they make at least one revision 7 or more
     # days after_the end of the course.
     new_editors.joins(:revisions).where('revisions.date > ?', @course.end + 7.days).distinct
-  end
-
-  def revisions_id
-    @course.revisions.map(&:mw_rev_id)
   end
 
   def revisions_by_namespace(namespace)
--- a/lib/analytics/course_csv_builder.rb
+++ b/lib/analytics/course_csv_builder.rb
@@ -25,6 +25,7 @@ class CourseCsvBuilder
     articles_created
     bytes_added
     total_edits
+    revision_id
     mainspace_edits
     article_talk_edits
     userspace_edits
@@ -51,10 +52,7 @@ class CourseCsvBuilder
     row << @course.article_count
     row << @course.new_article_count
     row << @course.character_sum
-    row << @course.revision_count
-    row << revisions_by_namespace(Article::Namespaces::MAINSPACE)
-    row << revisions_by_namespace(Article::Namespaces::TALK)
-    row << revisions_by_namespace(Article::Namespaces::USER)
+    revisions_rows(row)
     row << @course.view_sum
     row << @course.upload_count
     row << @course.uploads_in_use_count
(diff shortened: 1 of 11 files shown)


lib/analytics/course_csv_builder.rb
```diff
@@ -75,6 +73,14 @@ def headers
     end
   end
 
+  def revisions_rows(row)
+    row << @course.revision_count
+    row << revisions_id
```
ragesoss: This is, basically, putting all the revision ids as an array into one cell of the CSV? It looks like this also would add this to the other CSVs besides the revision one?

I'm not sure, but I think the revisions list is a different enough set of data that it might make more sense to create a new, dedicate CSV builder file instead of adapting this one to the new use. This one is designed to output one row per course, whereas I think a revisions CSV would ideally be one row per revision.

khyatisoneji: Are you suggesting to make a new file name course_revisions_csv_builder.rb? and fetch revision data for campaign revision CSV file from this file as we do for articles?

ragesoss: Following the pattern for articles is the best option, I think.

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -0,0 +1,92 @@
...
+    end
+  end
+
+  def set_revisions
+    @new_revisions = {}
+    @course.all_revisions.includes(article: :wiki).map do |edit|
+      revision_edits = @new_revisions[edit.article_id] || new_revision(edit)
+      revision_edits[:mw_rev_id] = edit.mw_rev_id
+      revision_edits[:mw_page_id] = edit.mw_page_id
+      revision_edits[:wiki_id] = edit.wiki_id
```
ragesoss: Using `wiki_id` is a problem, as there's no way to see which wiki that ID is for. I think it needs to use `edit.wiki.domain` instead.

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  def revisions_rows
+    @new_revisions.values.map do |revision_data|
+      build_row(revision_data)
+    end
+  end
+
+  def set_revisions
+    @new_revisions = {}
+    @course.all_revisions.includes(article: :wiki).map do |edit|
```
ragesoss: This should probably be `includes(:wiki, :article)`. The output should include the title and namespace of the article, as well as the domain (rather than the Dashboard-specific ID) of the wiki.

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -0,0 +1,92 @@
...
+    deleted
+    wp10
+    wp10_previous
+  ].freeze
+
+  def build_row(revision_data)
+    row = [revision_data[:mw_rev_id]]
+    row << revision_data[:page_id]
+    row << revision_data[:wiki_id]
+    row << revision_data[:wiki_domain]
```
ragesoss: Looks like this was intended to be included, but isn't being set anywhere?

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -24,11 +24,11 @@ def revisions_rows
 
   def set_revisions
     @new_revisions = {}
-    @course.all_revisions.includes(article: :wiki).map do |edit|
+    @course.all_revisions.includes(:wiki, :article).map do |edit|
       revision_edits = @new_revisions[edit.article_id] || new_revision(edit)
       revision_edits[:mw_rev_id] = edit.mw_rev_id
       revision_edits[:mw_page_id] = edit.mw_page_id
-      revision_edits[:wiki_id] = edit.wiki_id
+      revision_edits[:wiki_id] = edit.wiki.domain
```
ragesoss: I think `wiki` rather than `wiki_id` is the right name for this.

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -0,0 +1,91 @@
...
+    article_id
+    pageviews
+    deleted
+    wp10
+    wp10_previous
+  ].freeze
+
+  def build_row(revision_data)
+    row = [revision_data[:mw_rev_id]]
+    row << revision_data[:page_id]
```
ragesoss: This should be `:mw_page_id`. It's coming up blank right now.

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -0,0 +1,91 @@
...
+    page_id
+    wiki
+    characters_added
+    references_added
+    new
+    article_id
+    pageviews
+    deleted
+    wp10
+    wp10_previous
```
ragesoss: I think this should include article title and the user's username, in addition to these things.

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -0,0 +1,91 @@
...
+  CSV_HEADERS = %w[
+    revision_id
+    page_id
+    wiki
+    characters_added
+    references_added
+    new
+    article_id
+    pageviews
+    deleted
```
ragesoss: pageview and deleted are switched in `build_row`.

lib/analytics/course_revisions_csv_builder.rb
```diff
@@ -0,0 +1,103 @@
...
+      deleted: edit.deleted,
+      wp10: {},
+      wp10_previous: {},
+      article_id: edit.article_id
+    }
+  end
+
+  def update_title_username(revision_edits, edit)
+    revision_edits[:title] = edit.article.title
+    revision_edits[:username] = edit.user.username
```
ragesoss: This is making an extra query for every revision, to fetch the user and read the username.

Adding `:user` to the `includes` on line 27 should this N+1 problem.

ragesoss:
squashed/merged in a new branch.

PR #3068, opened 2019-07-08
Issue: 2622 Import ORES data for wikidata revisions 

khyatisoneji:
Updated ores_api.rb file to import ores data for wikidata revisions and also updated ores_api_spec.rb to run tests for wikidata revisions


ragesoss:
This looks great!

lib/ores_api.rb
```diff
@@ -41,7 +41,12 @@ class InvalidProjectError < StandardError
   private
 
   def query_url(rev_ids)
-    base_url = "/v3/scores/#{@project_code}/?models=articlequality&features&revids="
+    base_url =
```
ragesoss: Instead of an `if/else` here, I think it would make sense to set the model at the same time you set the project code, and then interpolate it here.

khyatisoneji: I have added another variable named `project_model` just like `project_code` for a different model

PR #3072, opened 2019-07-08
Feature/my articles update

bwreid:
## What this PR does

This PR does a few things:
- Refactors AssignCell
- Builds on the redesign of MyArticles section with a Remove button
- When assigning an article on the MyArticles section, the articles are pulled from the Available Articles that have been assigned by the instructor
- When deciding an article to review on the MyArticles section, the articles are pulled from what articles are currently being worked on by other students
- As an instructor, when assigning articles to students, the list shows both what articles are assigned and what articles are available to assign to a student.
- As an instructor, when assigning articles _to review_ to students, the list shows both what articles are currently assigned to _other_ students and what articles that student is already reviewing
- Change Assignment buttons and titles to be clearer

### Student View

**Adding an article**
image

**Reviewing an article**
image

### Instructor View

**Read-only view on /students**
image

**Edit view for assigned articles on /students**
image

**Edit view for reviewing articles on /students**
image

**Assignment buttons and titles**
image
image

### Student View

**Adding an article**
image

**Reviewing an article**
image

_Note: This surfaced a bug for me where if someone is reviewing a user's article, that same article shows up to be reviewed._

### Instructor View

**Read-only view on /students**
image

**Edit view for assigned articles on /students**
image

**Edit view for reviewing articles on /students**
image

**Assignment buttons and titles**
image
image
image


ragesoss:
Do we need some kind of affordance for long lists of available articles, or will it be usable now even if the list is (say) 20 or 100 articles long?

bwreid:
@ragesoss It will be usable but not ideal. It will work how it currently works which is that the list just gets really exceptionally long and stretches the page vertically. I can add on scrollbars according to a certain if you think that might be best?

ragesoss:
No, as long as it's usable, I think it's an improvement on the current UX and we can ship it. I guess I think scrollbars would probably be a little better, but maybe as we move forward wil think of something even better.

PR #3081, opened 2019-07-11
Added References Graph for articles

khyatisoneji:
**Screenshots:**

**Before:**

**ORES:**
image

**Article Development:**
image

**After:**

**ORES:**
image

**Article Development:**
image

### DIFF
--- a/app/assets/javascripts/components/articles/article_graphs.jsx
+++ b/app/assets/javascripts/components/articles/article_graphs.jsx
@@ -3,6 +3,7 @@ import createReactClass from 'create-react-class';
 import PropTypes from 'prop-types';
 import OnClickOutside from 'react-onclickoutside';
 import Wp10Graph from './wp10_graph.jsx';
+import ReferencesGraph from './references_graph.jsx';
 import EditSizeGraph from './edit_size_graph.jsx';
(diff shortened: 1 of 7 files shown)


khyatisoneji:
If you feel, updating ORES graph for a range of references added values is not necessary/important, I can undo those changes

ragesoss:
I think these would work better as line graphs, rather than step graphs.

khyatisoneji:
Made condition for small values( < 8) take value 8

ragesoss:
Cool!

I don't think the control for changing the interpolation is useful.

Other than that, this looks great!

ragesoss:
This looks good, but I noticed that it is a lot slower to switch between graphs than it was before. Do you know what is bogging down performance?

khyatisoneji:
No, it's working fine with me

ragesoss:
How many revisions are in the article(s) you've tested it with? It might be something that only bogs down when the revision count gets higher.

khyatisoneji:
Not more than 10 revisions

khyatisoneji:
Removed unnecessary code

ragesoss:
I couldn't ever figure out why this had a performance problem for me. This work will probably be useful if someone takes this up again in the future.

PR #3092, opened 2019-07-11
Wikidata labels fix

bwreid:


app/assets/javascripts/actions/wikidata_actions.js
```diff
@@ -27,9 +27,22 @@ const fetchWikidataLabelsPromise = (qNumbers) => {
...
+// This takes a Wikidata page title and checks whether it looks like
+// an *entity*, which could be in the Property (P), Lexeme (L), or main (Q) namespace.
+// Entities have labels, while other pages are conventional wiki pages.
+// Requesting the labels for an entitiy that doesn't exist will result in an API error,
+// so this is used to filter out edits to (for example) WikiProject pages.
+const isEntityTitle = (title) => {
+  if (!title.match(/:/)) { return true; } // mainspace, aka Q item
+  if (title.match(/(Property|Lexeme):/)) { return true; } // Property or Lexeme namespace, aka P or L
+  return false;
+};
```
bwreid: What about:
```js
const isEntityTitle = (title) => {
  const isQItem = !title.match(/:/);
  const isPropertyOrLexeme = title.match(/(Property|Lexeme):/);
  return isQItem || isPropertyOrLexeme;
}
```

ragesoss: I guess we actually want `test` instead of `match`, so that it will still return a boolean.

This version is more clear, but also means running the second test on every title, even though 95% will be mainspace. But this shouldn't ever be a performance bottleneck, so clarity wins.

PR #3100, opened 2019-07-12
Add Find Articles button to MyArticles section

bwreid:
## What this PR does

This PR adds a *Search Wikipedia for an Article* button which links to the Article Finder in the MyArticles under assignments.

## Screenshots

image


ragesoss:
Nice! It also shows up on the Available Articles tab multiline version of AssignButton, which I think is okay.

Issue #2762, opened 2019-04-15
Optionally limit or exclude articles/categories from user contributions


bwreid:
This is tangentially related to Issue #2129.

We want a way to be able to either limit or exclude specific articles and/or categories from user contributions for a course. For example, the following scenarios should be possible:

* A course should be able to select specific articles and only edits to those articles will be counted towards the course's progress
* A course should be able to select specific articles and any edits to those articles will be excluded from the overall course's progress

These changes could be enacted on the course page or a new page, if appropriate.

ragesoss:
I think this is done.

PR #3102, opened 2019-07-13
Enable article excluding

AmitJoki:
## What this PR does
First pass at resolving #2762. This PR contains only the backend and non-UI stuff.


ragesoss:
Cool! I like where it's going with the UI.

I think, rather than doing it based on Articles, this will need to be done via ArticlesCourses. We don't want one course's action to un-track an article to affect another course editing that same article. Or perhaps there's another way that I haven't thought of.

We'll also need to use the 'untracked' status for a particular one to be reflected in the course and user stats (via ArticlesCourses via CoursesUsers).

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -49,6 +49,15 @@ const Article = createReactClass({
     const formattedTitle = CourseUtils.formattedArticleTitle(this.props.article, this.props.course.home_wiki, this.props.wikidataLabel);
     const historyUrl = `${this.props.article.url}?action=history`;
 
+    let tracked;
+    if (this.props.current_user.isInstructor || this.props.current_user.admin) {
```
bwreid: I think just `current_user.isAdvancedRole` might work here as well.

AmitJoki: Yup. It does. Thanks!

AmitJoki:
@ragesoss this PR only sets the ability to track/untrack an article. The actual implementation that will make use of the scope and the suggestions you made, I am thinking of opening another PR. Is that OK or should I push commits into this one?

ragesoss:
The easiest way would be to implement the back end portion only - just DB and model scopes and update cycle - as one PR, then the frontend concerns as another. We could merge back end by itself, but not this frontend stuff.

ragesoss:
By easiest, I mean easiest to review. :)

app/models/course.rb
```diff
@@ -84,16 +84,27 @@ class Course < ApplicationRecord
   # :revisions and :all_revisions have the same default implementation,
   # but a course type may override :revisions.
   has_many(:revisions, lambda do |course|
```
ragesoss: Changing the base `revisions` scope seems like a riskier and more disruptive strategy to me. I think you probably want to go the opposite route and add `tracked_revisions` as a new scope that builds on top of revisions.

ragesoss: Ideally, that would also avoid the need to change any existing specs. Ideally, this new feature wouldn't break any specs, since it will only change the behavior of the system in cases where there are ArticlesCourses that have been explicitly marked for exclusion.

app/models/course_data/articles_courses.rb
```diff
@@ -49,7 +49,7 @@ def new_article
   end
 
   def live_manual_revisions
-    course.revisions.live.where(article_id: article_id)
+    course.tracked_revisions.live.where(article_id: article_id)
```
ragesoss: This one should still use `revisions` instead of `tracked_revisions`, I think. We mark an article as untracked on teh ArticlesCourses record itself, but there's no reason not to update the data on the AC record, even if it's being excluded from other stats.

ragesoss:
The first failing test, for CourseStatistics, looks like it's a byproduct of how the test is written: it directly creates the records that were needed to make the test pass, but it skips creating other records (the ArticlesCourses records in particular) that were not necessary.

I think if you tweak the test to also generate do `ArticlesCourses.update_from_course` before the CoursesUsers step, then it will be fixed.

ragesoss:
Similarly, the RevisionImporter spec is testing a case that would pass if the ArticlesCourses had been generated.

This makes me thing again that maybe the data model should be tweaked, though. Like, ideally, it should be that ArticlesCourses could _exclude_ articles from the tracked_revisions scope, but wouldn't be necessary to _include_ them. Is there a clean way to do that?

ragesoss:
I think the other spec I'd really like to see for this is a more high-level one that tries to replicate the whole cycle:

1. Create a course with a user who has edited several articles during the time period.
2. Run `UpdateCourseStats` on the course to generate stats, CoursesUsers, and ArticlesCourses.
3. Run some expectations on the stats.
4. Update one of the ArticlesCourses, then run `UpdateCourseStats` again.
5. Run more expectations on the now-lower (but not zero) stats.

spec/models/course_spec.rb
```diff
@@ -846,4 +847,48 @@
       expect(course.pages_edited).not_to include(article2)
     end
   end
+
+  describe '#tracked_revisions' do
```
ragesoss: I think this makes sense as its own file rather than added onto the already-huge course spec, especially since it's pretty independent of the rest of the specs here.

spec/models/course_spec.rb
```diff
@@ -846,4 +847,48 @@
...
+    let(:course) { create(:course, start: 1.year.ago) }
+    let(:user) { create(:user, id: 3, username: 'Amit') }
+    let(:article) { create(:article, namespace: 0, deleted: false) }
+    let(:article2) { create(:article, namespace: 0, deleted: false) }
+
+    before do
+      create(:revision, date: course.start + 1.minute,
```
ragesoss: I think, rather than creating the revisions explicitly, this test should choose a user and course date range such that the update will create those articles and revisions and we're relying completely on the update flow to generate the article, revision and course stats data.

spec/lib/tracked_revisions_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_dependency "#{Rails.root}/lib/analytics/course_statistics.rb"
+
+describe 'TrackedRevisions' do
+  let(:course) { create(:course, start: '30-05-2019'.to_date, end: '31-07-2019'.to_date) }
+  let(:user) { create(:user, username: 'Textorus') }
```
ragesoss: I recommend choosing a user who is less active than this one, and using a date farther in the past.

One consideration is that the course update process pulls all the revisions from the course start date through the present, so over time the query could get slower if it's a very active user. The total edits over all time are also a factor in how quickly the queries go.

spec/lib/tracked_revisions_spec.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_dependency "#{Rails.root}/lib/analytics/course_statistics.rb"
+
+describe 'TrackedRevisions' do
```
ragesoss: This could be named more clearly, perhaps `Course#tracked_revisions` since that's the exact scope it's mainly testing.

Probably this can go in `spec/models` in that case.

spec/lib/tracked_revisions_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  before do
+    course.students << user
+  end
+
+  it 'fetches the course stats for only tracked articles' do
+    VCR.use_cassette 'course_upload_importer/Textorus' do
+      UpdateCourseStats.new(course)
+    end
+    subject = CourseStatistics.new([course.id]).report_statistics
+    expect(subject[:articles_edited]).to eq(151)
```
ragesoss: This is too specific for such a large number of articles; if there are pages or edits that get deleted sometime in the future, the spec will break.

I think exact numbers are okay if there are far fewer edits involved, though. I would suggest an example user and time range where there are just a few articles edited. Here's perhaps a good one (from one our our courses last summer): https://en.wikipedia.org/wiki/Special:Contributions/Hbultra

If you use the article title find find the right article and ArticlesCourses records to untrack, the test will be very easy to understand (and to compare expected results by viewing the contributions history).

spec/lib/tracked_revisions_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  before do
+    course.students << user
+  end
+
+  it 'fetches the course stats for only tracked articles' do
+    VCR.use_cassette 'course_upload_importer/Textorus' do
+      UpdateCourseStats.new(course)
+    end
+    subject = CourseStatistics.new([course.id]).report_statistics
```
ragesoss: I don't think this needs to go through CourseStatistics.

The main things I'd like to see tested are a) the values that end up on the `Course` and `CoursesUsers` records, and b) the revisions (probably just the count) included in the `tracked_revisions` scope. If you test those before and after the untracking and re-updating, that will be a nice test I think.

spec/models/tracked_revisions_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+    expect(course.revisions.count).to eq(course.tracked_revisions.count)
+    untracked_id = Article.find_by(title: 'Afterload').id
+    course.articles_courses.where(article_id: untracked_id).update(tracked: false)
+    VCR.use_cassette 'course_upload_importer/Hbultra' do
+      UpdateCourseStats.new(course)
+    end
+    expect(course.tracked_revisions.count).to be < course.revisions.count
+  end
+end
```
ragesoss: This looks likes a solid basic approach to me.

I'd like a little more detail, with some expectations set for the `course` data itself as well as the CoursesUsers and ArticlesCourses records before and after the untracking.

ragesoss:
This looks looks good to me, although I've seen it enough times that I'm going to be blind to anything I've missed so far. @bwreid and @khyatisoneji, could you two have a look, and ask about anything that's unclear about what's going on?

db/schema.rb
```diff
@@ -47,6 +46,7 @@
     t.date "average_views_updated_at"
     t.integer "wiki_id"
     t.integer "mw_page_id"
+    t.boolean "tracked", default: true
```
bwreid: Is this supposed to be on the "articles" table as well?

bwreid:
Hi @AmitJoki! I'm having a hard time reproducing this. I'm trying the following:
```ruby
slug = 'Course/Slug' # some slug for a course
Course.find_by(slug: slug).articles_courses.update_all(tracked: false)
```
Then, I'm going to that course and going to the `/manual_update` path. I don't see any of the revisions disappear or stats change. Do you have any ideas on how I can test this?

lib/course_cache_manager.rb
```diff
@@ -67,7 +67,7 @@ def update_trained_count
   end
 
   def update_revision_count
-    @course.revision_count = @course.revisions.live.size
+    @course.revision_count = @course.tracked_revisions.live.size
```
ragesoss: There's more in this file that needs to be adjusted to ensure that only tracked articles are contributing to the stats.

The main before/after test should look at every one of these that are based on articles edited: `revision_count`, `character_sum_ms`, `view_sum`, `references_count`, `article_count`, `new_article_count`.

db/schema.rb
AmitJoki: No. Making an article tracked will make it so for every otther course irrespective of whether the instructor wants this article to be tracked or not. So it is only available on ArticlesCourses where it can be namespaced along with a course id so other courses can still make the decision of whether they want the article to be tracked or not.

app/models/course.rb
```diff
@@ -263,6 +263,10 @@ def training_module_ids
                                   .collect(&:training_module_ids).flatten
   end
 
+  def tracked_revisions
+    revisions.where.not(article_id: articles_courses.not_tracked.pluck(:article_id))
```
khyatisoneji: It's a Lil confusing for me, there is a field named `tracked`, why not directly use the `tracked` and get `tracked revisions` directly?

AmitJoki: It is because articles courses isn't the one source of truth and so we exclude the untracked articles which are definitely available on articles_courses rather than including the other way around.

ragesoss: @khyatisoneji this way, the default before any ArticlesCourses have been created is that all revisions will be tracked. ArticlesCourses get generated from the revisions, and we want to make sure there are no unexpected missing stats.

AmitJoki:
@bwreid may be try it out again. Might have been because the CourseCacheManager had some migration yet to do. I've done that now and hopefully you should be seeing changes in the stats.

bwreid:
Thanks @AmitJoki. It looks pretty good! Here's a course I have with tracked articles:
image
And then here it is after setting those articles to `tracked: false`:
image
It's hard to tell at a quick glance what those edits are, but it looks like the uploads to commons are still pulling through. @ragesoss is this to be expected?

ragesoss:
@bwreid the commons uploads are behaving as expected, yes. Those are tracked in a whole separate table, CommonsUploads, so they don't touch the articles, articles_courses, or revisions tables.

The 24 total edits is (I assume) a wrinkel I had not considered. That total edit count is for all edits the Dashboard knows about, not just mainspace article edits that correspond to ArticlesCourses. So un-tracking all the ArticlesCourses only excludes the mainspace edits from the count. I think this is okay; it should make sense for the most common use cases where someone just wants to exclude one particular article from the stats. 'Total edits' is not an especially useful metric anyway, although this will likely frustrate people in a few edge cases where they want to exclude edits by active users in other namespaces as well.

ragesoss:
I tested this out now as well (which I'd neglected previously), and it seems to be working smoothly. Anything else you'd like to see from it, or test on it, @bwreid?

bwreid:
@ragesoss Nope, other than that this seems good to me!

ragesoss:
Great. @AmitJoki my plan is to merge and deploy this tomorrow, and then monitor carefully for any significant performance impacts resulting from the new joins on crucial revisions queries.

AmitJoki:
@ragesoss thanks! I am eager to get this monkey off my back, has been with me for too long :smile:

PR #3110, opened 2019-07-15
Add `sandbox_url` to Assignments

bwreid:
# What this PR does

At the moment, this PR:
- Adds `sandbox_url` to Assignments
- Before an Assignment is saved, creates a relevant `sandbox_url`
- If that assignment already exists in the same course, use the same `sandbox_url` for the previous assignment
- Adds a script that will backfill `sandbox_urls` for past Assignments

### DIFF
--- a/app/models/course_data/assignment.rb
+++ b/app/models/course_data/assignment.rb
@@ -74,10 +74,9 @@ class Assignment < ApplicationRecord
   end
 
   def set_sandbox_url
-    return unless wiki && user
+    return unless user
     # If the sandbox already exists, use that URL instead
-    existing = Assignment.where(course: course, article: article,
-                                wiki: wiki)
+    existing = Assignment.where(course: course, article_title: article_title, wiki: wiki)
                          .where.not(user: user).first
     if existing
       self.sandbox_url = existing.sandbox_url
--- a/app/models/course_data/assignment.rb
+++ b/app/models/course_data/assignment.rb
@@ -75,7 +75,15 @@ class Assignment < ApplicationRecord
 
   def set_sandbox_url
     return unless wiki && user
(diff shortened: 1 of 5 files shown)


app/models/course_data/assignment.rb
```diff
@@ -70,4 +72,18 @@ def set_defaults_and_normalize
     return if article_title.nil?
     self.article_title = ArticleUtils.format_article_title(article_title, wiki)
   end
+
+  def set_sandbox_url
+    return unless wiki && user
```
ragesoss: Is it possible to have an Assignment without a wiki?

app/models/course_data/assignment.rb
```diff
@@ -70,4 +72,18 @@ def set_defaults_and_normalize
...
+    # If the sandbox already exists, use that URL instead
+    existing = Assignment.where(course: course, article: article,
+                                wiki: wiki)
+                         .where.not(user: user).first
```
ragesoss: This won't match for cases where the article doesn't exist, so the existing assignment has an `article_title` that is the same instead.

lib/tasks/assignment.rake
```diff
@@ -0,0 +1,30 @@
...
+  "#{base_url}/User:#{user.username}/sandbox"
+end
+
+namespace :assignment do
+  # This task will assign sandbox_urls to all valid assignments
+  # that were created before August 2019
+  desc 'Give assignments default sandbox_url'
+  task set_sandbox_url: :environment do
+    Rails.logger.debug 'Setting sandbox_url for all assignments'
+    Assignment.where(created_at: '2019-08-01'.to_date).each do |assignment|
```
ragesoss: I think this only matches records created at exactly the start of August.

PR #3111, opened 2019-07-15
Issue: 2557 Added delete and refresh cached button for ores cached data

khyatisoneji:


app/assets/javascripts/components/articles/course_ores_plot.jsx
```diff
@@ -59,6 +80,10 @@ const CourseOresPlot = createReactClass({
         return (
           <div className="ores-plot">
             <CourseQualityProgressGraph graphid={'vega-graph-ores-plot'} graphWidth={1000} graphHeight={200} articleData={this.state.articleData} />
+            <div>
+              <button className="button ghost stacked right" onClick={this.delete}>Delete Cached Data</button>
+              <button className="button ghost stacked right" style = {{ marginRight: 20 }} onClick={this.refresh}>Refresh Cached Data</button>
```
ragesoss: I don't think there's any need to do these separately; the only use case for deleting the data is so that you can see a newer version of it, so I think just the 'refresh' version should be included.

ragesoss:
Sorry, didn't realize this until I tested it out visually...

The 'refresh' button should not be rendered unless the plot is already open. It's only relevant for when the user is viewing the plot and realizes it's not up-to-date, and it isn't necessarily clear which data it's talking about if it's in the initial state when the plot has not been rendered.

Also, this should probably be a subtle link, perhaps in parentheses at the end of the paragraph explaining the graph, instead of a prominent button.

khyatisoneji:
I tired putting `refresh cached data button` when the graph is rendered but wasn't able to render refreshed graph as they have the same URL, still working on that

khyatisoneji:
Not, sure this will work but it is able to refresh cached data and allows to refresh only one time at a time
Until scrolled back and then again hit refresh

Also not sure about loading being displayed

Will change button to link once refresh button works perfectly

bwreid:
@khyatisoneji, it looks like the refresh button is working as intended. Whenever it is clicked, I see a request going to the backend and updating the `refreshedData` key. What part of it is not working as you'd like?

Although, it doesn't _look_ like anything is happening. If you want the graph to disappear and then reappear (like it does the first time), I believe you just need to set `refreshedData: null` before you fetch the file.

khyatisoneji:
I tried making refreshedData value `null` but it wasn't working, I was doing it in a wrong way so the `refreshedData` was not set to `null`. I was directly doing `this.setState.refreshedData = null` instead of `this.setState({ articleData: null })`

app/controllers/ores_plot_controller.rb
```diff
@@ -10,6 +10,16 @@ def course_plot
     render json: json_data
   end
 
+  def delete_ores_data
```
ragesoss: This should be a private method. Other than that, I think this is ready to ship.

PR #3120, opened 2019-07-16
Update references for wikidata revisions

khyatisoneji:

### DIFF
--- a/app/models/wiki_content/revision.rb
+++ b/app/models/wiki_content/revision.rb
@@ -68,30 +68,23 @@ class Revision < ApplicationRecord
     "/recent-activity/plagiarism/report?ithenticate_id=#{ithenticate_id}"
   end
 
-  def wikidata_ref_tags
-    features['feature.len(<datasource.wikidatawiki.revision.references>)']
-  end
-
-  def wikidata_ref_tags_previous
-    return 0 if new_article
-    features_previous['feature.len(<datasource.wikidatawiki.revision.references>)']
-  end
-
   def ref_tags
-    features['feature.wikitext.revision.ref_tags']
+    return features['feature.wikitext.revision.ref_tags'] if
+      features['feature.wikitext.revision.ref_tags']
+    return features['feature.len(<datasource.wikidatawiki.revision.references>)'] if
+      features['feature.len(<datasource.wikidatawiki.revision.references>)']
   end
 
   def ref_tags_previous
     return 0 if new_article
-    features_previous['feature.wikitext.revision.ref_tags']
+    return features_previous['feature.wikitext.revision.ref_tags'] if
+      features_previous['feature.wikitext.revision.ref_tags']
+    return features_previous['feature.len(<datasource.wikidatawiki.revision.references>)'] if
+      features_previous['feature.len(<datasource.wikidatawiki.revision.references>)']
   end
 
   def references_added
-    return ref_tags - ref_tags_previous if ref_tags && ref_tags_previous
-
-    return wikidata_ref_tags - wikidata_ref_tags_previous if
-      wikidata_ref_tags && wikidata_ref_tags_previous
-
-    return 0
+    return 0 unless ref_tags_previous && ref_tags
+    ref_tags - ref_tags_previous
   end
 end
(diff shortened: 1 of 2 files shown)


app/models/wiki_content/revision.rb
```diff
@@ -68,6 +68,15 @@ def plagiarism_report_link
     "/recent-activity/plagiarism/report?ithenticate_id=#{ithenticate_id}"
   end
 
+  def wikidata_ref_tags
+    features['feature.len(<datasource.wikidatawiki.revision.references>)']
+  end
+
+  def wikidata_ref_tags_previous
+    return 0 if new_article
+    features_previous['feature.len(<datasource.wikidatawiki.revision.references>)']
```
ragesoss: I think this would work better if you incorporated it into `ref_tags` instead of having a parallel method dedicated to wikidata.

app/models/wiki_content/revision.rb
```diff
@@ -69,12 +69,18 @@ def plagiarism_report_link
   end
 
   def ref_tags
-    features['feature.wikitext.revision.ref_tags']
+    return features['feature.wikitext.revision.ref_tags'] if
```
ragesoss: This could instead be something like `features['feature.wikitext.revision.ref_tags'] || features['feature.len(<datasource.wikidatawiki.revision.references>)']`.

Also, the key for each feature could probably be turned into a constant to make it clearer:

`features[WIKITEXT_REF_TAGS] || features[WIKIDATA_REFERENCES]`

app/models/wiki_content/revision.rb
```diff
@@ -43,6 +43,9 @@ class Revision < ApplicationRecord
 
   include ArticleHelper
 
+  WIKITEXT_REF_TAGS = 'feature.wikitext.revision.ref_tags'
+  WIKIDATA_REFERENCES = 'feature.len(<datasource.wikidatawiki.revision.references>)'
```
ragesoss: As a matter of style, since these are only used within a single method, I would usually move these constant assignments right above the method. It doesn't matter much either way, but that will be a little more consistent with the rest of the codebase.

spec/models/revision_spec.rb
```diff
@@ -28,6 +28,7 @@
 describe Revision, type: :model do
   describe '#references_added' do
     let(:refs_tags_key) { 'feature.wikitext.revision.ref_tags' }
+    let(:wiki_refs_tags_key) { 'feature.len(<datasource.wikidatawiki.revision.references>)' }
```
ragesoss: This name is misleading, as it's specifically Wikidata.

spec/models/revision_spec.rb
```diff
@@ -57,11 +58,21 @@
                features: {
                  refs_tags_key => 10
                })
+        create(:revision,
+               mw_rev_id: mw_rev_id + 1,
+               article_id: 90010260,
+               mw_page_id: 90010260,
+               new_article: true,
+               features: {
+                 wiki_refs_tags_key => 10
```
ragesoss: This is testing the Wikidata version, but the context says Wikipedia. This may lead to confusion later on.

I recommend being a little more explicit, either assigning a name or or using `lets` to make it clear which is a Wikidata revision and which is a Wikipedia revision. I also think this is more likely to be a stable test if it's done with a Wikidata example fully — including using a real mw_page_id from Wikidata, and setting the wiki for the revision to be Wikidata as well.

khyatisoneji: Had to use VCR, as was throwing `VCR::Errors::UnhandledHTTPRequestError:`
Also using VCR in just `it` block where revisions were created in `before` block was throwing the same error

spec/models/revision_spec.rb
```diff
@@ -28,27 +28,38 @@
 describe Revision, type: :model do
   describe '#references_added' do
     let(:refs_tags_key) { 'feature.wikitext.revision.ref_tags' }
+    let(:wikidata_refs_tags_key) { 'feature.len(<datasource.wikidatawiki.revision.references>)' }
+    let(:wiki) { create(:wiki, project: 'wikidata', language: 'en') }
```
ragesoss: This should be named more explicitly, since the tests care about this being a `wikidata` example rather than an arbitrary wiki.

PR #3128, opened 2019-07-19
Feature/peer review count

bwreid:
## What this PR does

Updates the MyArticles component to display the amount of Peer Reviews a student should choose.

## Screenshots

image
image

### DIFF
--- a/lib/wizard_timeline_manager.rb
+++ b/lib/wizard_timeline_manager.rb
@@ -43,7 +43,7 @@ class WizardTimelineManager
 
     # Save any tags that have been generated from this Wizard output
     add_tags
-  
+
     # Save any flags that correspond to specific wizard choices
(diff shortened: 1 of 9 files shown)


lib/wizard_timeline_manager.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+# rubocop:disable Metrics/ClassLength
```
ragesoss: For the other classes that exceed the rubocop length, I've done the disabling in `.rubocop.yml` instead, so that there's a central place to see all the over-length classes.

bwreid: Ah nice! I will check that out and change. 👍

ragesoss:
Looks great!

PR #3129, opened 2019-07-19
Senty error fixes

ragesoss:
This is an assortment of fixes for infrequent errors I noticed browsing Sentry, and some related refactoring I noticed while working on the fixes.


ragesoss:
Difference places use `course_id` to mean either the actual ID or the slug, and it's confusing. This was a pattern from the early origins of the app, and it's been a low-grade pain point for a long time. Ultimately, we should name the data 'course_slug' whenever it's the slug, and id when it's the id. This chips away at the problem by fixing it in a couple of relatively isolated places.

PR #3143, opened 2019-07-24
Fixes #3024

### LINKED ISSUE #3024: Do not require resetState to be passed to component
# Current Behavior:

Currently, the `EditableRedux` component requires that `resetState()` is passed in, otherwise it will fail as it's a function. This leads to some weird requirements on the components that use it, as if no `resetState()` function makes sense, an empty function will need to be passed in.

### Desired Behavior:

Instead, in the `EditableRedux` component, we should change the `cancelChanges()` method so that it checks to see if `resetState` is defined and a function. If so, it will invoke it. Otherwise, it will do nothing.

### Related Files & Code:

* [EditableRedux: cancelChanges()](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/components/high_order/editable_redux.jsx#L37-L41)

AmitJoki:
resetState is now optional.

### DIFF
--- a/app/assets/javascripts/components/high_order/editable_redux.jsx
+++ b/app/assets/javascripts/components/high_order/editable_redux.jsx
@@ -23,7 +23,7 @@ const EditableRedux = (Component, Label) => {
       course_id: PropTypes.any,
       current_user: PropTypes.object,
       editable: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


bwreid:
This looks good to me! Thank you.

PR #3148, opened 2019-07-29
[WIP] Fetching articles' title using PSID

khyatisoneji:
**Screenshoots:**

**_Before:_**
image

**_After:_**
image


spec/lib/importers/category_importer_spec.rb
```diff
@@ -5,9 +5,22 @@
 
 describe CategoryImporter do
   let(:wiki) { Wiki.default_wiki }
-  let(:subject) { described_class.new(wiki).page_titles_for_category(category, depth) }
+
+  describe '.page_titles_for_psid' do
+    let(:subject) { described_class.new(wiki).page_titles_for_psid(psid) }
+
+    let(:psid) { 9964305 }
```
ragesoss: Document this, since it's a specific query that the test relies on.

lib/importers/category_importer.rb
```diff
@@ -22,6 +23,16 @@ def page_titles_for_category(category, depth=0, namespace=nil)
     page_properties_for_category(category, 'title', depth, namespace)
   end
 
+  def page_titles_for_psid(psid)
```
ragesoss: I don't think this makes sense as part of CategoryImporter, as it's doing something fundamentally different from CategoryImporter, which is all about getting data from a specific named category via MediaWiki.

Although PetScan does use categories itself, from the Dashboard's perspective it's a whole different source of data.

It's simple enough, though, that you might just add this method into `PetScanApi` directly.

PR #3156, opened 2019-07-30
Find articles using PSID

khyatisoneji:
## Screenshots
Before:
image

After:
image


ragesoss:
This is cool! It looks like it retains the results from a previous query and adds the new results, instead of clearing the old before loading the new.

Also, with PetScan queries, `relevance index` is not an accurate description; that value should probably just not be shown for that column when it's a petscan query, because the ordering of petscan results is by MediaWiki page ID, so unlike the search API queries, results being at the top isn't related to relevance.

khyatisoneji:
I made a few changes and it is a lot quicker than it was and also now it doesn't throw `Error: 429`
But this still requires some changes, like the page is slower to scroll down
Also, those conditions are placed for fetching details of articles because it sometimes throws  `Cannot set the value of undefined`  for those I'm directly setting their state as `received`

ragesoss:
This is most of the way there, but I don't think it's a super high priority so I'm going to close this. If we return to it, this will be a solid starting point to finish it up.

PR #3157, opened 2019-07-30
Add PSID to category

khyatisoneji:


ragesoss:
This looks good. You should add back the test as well.

khyatisoneji:
The test was for `Category_importer`, I'll add the tests once I'm done with adding PSID to `Category`

ragesoss:
I tried this out, and it's working really nicely. I like how little code it actually took to implement!

It looks like there are some missing interface messages. Once those are fixed up, I think this is ready to merge.

missing

There are some follow-up improvements that could make this more useful, especially around surfacing the status of a given category, especially showing at the frontend whether it's already been imported and how many articles it has. If you want to dive into that, please do. Otherwise, I'll add an issue so we can come back to it another time.

PR #3163, opened 2019-07-31
[WIP] Feature/complete assignment flags

bwreid:


lib/assignment_pipeline.rb
```diff
@@ -0,0 +1,68 @@
...
+    PEER_REVIEW_COMPLETED = 'peer_review_completed'
+  end
+
+  REVIEW_STATUS_PIPELINE = [
+    ReviewStatuses::PEER_REVIEWER_NEEDED,
+    ReviewStatuses::PEER_REVIEW_STARTED,
+    ReviewStatuses::PEER_REVIEW_COMPLETED
+  ]
+
+  def initialize(flags:)
```
ragesoss: Is the idea that `Assignment` will use this class by sending its own flags in, and then it will delegate things like `assignment_status` to this?

lib/assignment_pipeline.rb
```diff
@@ -0,0 +1,68 @@
...
+  def assignment_status
+    last_status(@flags[:assignment_status]) || ASSIGNMENT_STATUS_PIPELINE.first
+  end
+
+  def next_assignment_status
+    next_status(ASSIGNMENT_STATUS_PIPELINE, assignment_status)
+  end
+
+  def increment_assignment_status
+    @flags[:assignment_status] ||= []
```
ragesoss: `@flags` here is going to be from an `Assignment` instance? So this is mutating the flags object that got passed in, but leaving it up to `Assignment` to then save it? Might be better to pass in the whole assignment to this class so that the flags can be both changed and persisted together, or alternatively, don't do any mutations of `@flags` and leave that to the code that calls this class.

lib/assignment_pipeline.rb
```diff
@@ -0,0 +1,68 @@
...
+    @flags[:assignment_status] << [ next_assignment_status, Time.now ]
```
ragesoss: This potentially allows any number of calls to `increment_assignment_status`, and it will fill the array with `[nil, <time>]` once it reaches the last status, I think.

If it's going to be a strictly linear status pipeline, I don't think there's a need to store every status change; just keeping the current status and its timestamp should be a good representation of the state.

app/models/course_data/assignment.rb
```diff
@@ -65,6 +68,14 @@ def sibling_assignments
       .where.not(user: user_id)
   end
 
+  def status
+    AssignmentPipeline.new(assignment: self).status
```
ragesoss: Better to just inflate one `AssignmentPipeline` per `Assignment`, I think, and re-use it across methods.

lib/assignment_pipeline.rb
```diff
@@ -0,0 +1,56 @@
...
+    ReviewStatuses::PEER_REVIEW_COMPLETED
+  ].freeze
+
+  def initialize(assignment:)
+    @assignment = assignment
+    @flags = assignment.flags
+
+    role = Assignment::ROLE_NAMES[assignment.role]
+    editing = role == 'Editing'
+    @key = editing ? :assignment : :review
```
ragesoss: This is a lot of work to dance through the constants, and still ends up using multiple magic strings.

Maybe define `editing?` and/or `reviewing?` on `Assignment`? Or maybe define `role_name`. The role name could also be used as the key.

app/models/course_data/assignment.rb
bwreid: What do you mean by "inflate?" Something like the following in each?
```ruby
def status
  @pipeline ||= AssignmentPipeline.new(assignment: self)
  @pipeline.status
end
```

PR #3175, opened 2019-08-05
Exclude article UI

AmitJoki:
## What this PR does
First pass at resolving #2762 This PR contains the UI changes.
## Screenshots
Before:
image

After:
image

Also, for non advanced roles, the tracking toggle is greyed out and disabled thus still showing if the article is tracked or not to the commonfolk.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -175,7 +175,7 @@ en:
     show_current_version_with_authorship_highlighting: 'Current Version w/ Authorship Highlighting'
     title: Title
     tracked: Tracked
-    tracked_doc: The status of whether or not an article has been tracked. The course stats depend only on tracked articles so articles deemed not relevant to the course can be untracked here and the stats will reflect that the next time the course is updated.
+    tracked_doc: Only tracked articles contribute to the the statistics for this program. When articles are marked as untracked, they will be removed from the statistics during the next data update.
     title_example: Article title
     tools: Assessment tools
     view: View Article
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -175,6 +175,7 @@ en:
     show_current_version_with_authorship_highlighting: 'Current Version w/ Authorship Highlighting'
     title: Title
     tracked: Tracked
+    tracked_doc: The status of whether or not an article has been tracked. The course stats depend only on tracked articles so articles deemed not relevant to the course can be untracked here and the stats will reflect that the next time the course is updated.
     title_example: Article title
     tools: Assessment tools
     view: View Article
--- a/spec/features/course_page_spec.rb
+++ b/spec/features/course_page_spec.rb
@@ -426,4 +426,26 @@ describe 'the course page', type: :feature, js: true do
       end
     end
   end
+
+  describe 'articles tracking' do
+    it 'does not allow articles to be marked for tracking by students' do
+      js_visit "/courses/#{Course.last.slug}/articles"
+      expect(first('.tracking').find('input').disabled?).to eq(true)
+    end
+
+    it 'does allows articles to be marked for tracking by instructors/admin' do
+      login_as(admin)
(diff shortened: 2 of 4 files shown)


ragesoss:
Can this be rebased so that it's only the UI changes, instead of including the already-merged backend changes as well?

ragesoss:
This works really nicely. I think it needs a few further refinements to be sufficiently intuitive...

Users should be able to filter based on tracked status (just like with the current option of 'new, existing, new and existing') if there are both tracked and untracked articles. The default should be to only show tracked articles.

If the user is not able to edit the tracked status, then we could also hide the the tracked toggle altogether, except for users who can use those toggles.

A documentation tooltip on the 'Tracked' column header would also be helpful, explaining what it means to un-track an article and that the stats will reflect it the next time the course is updated.

AmitJoki:
@ragesoss have rebased on master and added the tracked docs and made thee tracking component visible only to those who can use it.

For the filter, I will branch off from this branch to keep things separate.

config/locales/en.yml
```diff
@@ -175,6 +175,7 @@ en:
     show_current_version_with_authorship_highlighting: 'Current Version w/ Authorship Highlighting'
     title: Title
     tracked: Tracked
+    tracked_doc: The status of whether or not an article has been tracked. The course stats depend only on tracked articles so articles deemed not relevant to the course can be untracked here and the stats will reflect that the next time the course is updated.
```
ragesoss: `Only tracked articles contribute to the the statistics for this program. When articles are marked as untracked, they will be removed from the statistics during the next data update.`

spec/features/course_page_spec.rb
```diff
@@ -426,4 +426,61 @@ def js_visit(path, count=3)
...
+      login_as(admin)
+      js_visit "/courses/#{course.slug}/articles"
+      expect(course.tracked_revisions.count).to eq(course.revisions.count)
+      first('.tracking').click
+      expect(course.tracked_revisions.count).to be < course.revisions.count
+    end
+
+    it 'is not shown for certain course types like ClassroomProgramCourse' do
+      login_as(admin)
+      js_visit "/courses/#{Course.first.slug}/articles"
```
ragesoss: This first course is different from `course`?

This is confusing, because it's not explicit that `Course.first` is the the one from the `before` block and is therefore ClassroomProgramCourse type.

I think a better option would be to add `let(:classroom_program_course)` that explicitly creates a course of that type, parallel to the editathon in every other respect.

spec/features/course_page_spec.rb
```diff
@@ -426,4 +426,61 @@ def js_visit(path, count=3)
...
+             start: '10-07-2019'.to_date,
+             end: '10-08-2019'.to_date,
+             timeline_start: course_start.to_date,
+             timeline_end: course_end.to_date,
+             school: 'This university.foo.bar',
+             expected_students: 1,
+             term: 'term 2015',
+             home_wiki_id: home_wiki.id,
+             description: 'This is a great course',
+             flags: update_logs)
```
ragesoss: The more you can rely on the factory defaults, the shorter and easier to understand this test will be. You might be able to just do `create(:editathon)` and have this pass.

spec/features/course_page_spec.rb
```diff
@@ -426,4 +426,61 @@ def js_visit(path, count=3)
...
+    end
+    let(:user) { create(:user, username: 'Hbultra') }
+
+    before do
+      course.students << user
+      VCR.use_cassette 'course_upload_importer/Hbultra' do
```
ragesoss: Because this spec is aimed at the frontend rather than the tracked data model itself, doing this network-reliant data update during the spec is not ideal. It will be slower and more fragile than it needs to be, and it doesn't test anything related to the data updates that isn't tested elsewhere already. I'd rather see a tracked article added to the course explicitly in order to set up the right situation where the filters and toggles appear.

PR #3176, opened 2019-08-05
Using references_added function for user's ref count

khyatisoneji:


ragesoss:
@khyatisoneji please have a look at the test that failed.

spec/presenters/individual_statistics_presenter_spec.rb
```diff
@@ -40,7 +37,7 @@
         expect(course2.revisions.count).to eq(1)
         expect(subject.individual_article_views).to eq(100)
         expect(subject.individual_character_count).to eq(100)
-        expect(subject.individual_references_count).to eq(5)
+        expect(subject.individual_references_count).to eq(22)
```
ragesoss: Why didn't this work the way it was before?

khyatisoneji: Here the example taken is for `newarticle`, so it'll return only current `ref_tags`, earlier I was just directly just subtracting, now I'm calling the function

So, I was made changes accordingly for `newarticle` revision example

ragesoss: Ah! That makes sense, thanks.

PR #3185, opened 2019-08-08
Make the search UI align good on all screens and widths

### LINKED ISSUE #2886: Improve alignment of ArticleFinder wiki selector
In a wide enough window the wiki selector aligns with the right side of the input in ArticleFinder.

In a narrower window, it starts to shift out of alignment:

alignment shift

AmitJoki:
Fixes #2886 

Before:

image

After:

image

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -148,8 +148,8 @@ const ArticleFinder = createReactClass({
         placeholder={I18n.t('article_finder.search_placeholder')}
         onKeyDown={this.onKeyDown}
         ref="searchbox"
-        child={searchButton}
-      />);
+      >{searchButton}
+      </TextInput>);
 
     const searchType = (
       <div>
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -377,8 +377,8 @@ const ArticleFinder = createReactClass({
     );
     if (this.state.showLanguageAndWikiSelectors) {
       options = (
-        <div>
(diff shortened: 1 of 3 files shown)


ragesoss:
@bwreid Visually, this looks good to me. Any objections to the implementation, adding a child component prop to work around the HTML/CSS structure?

bwreid:
@AmitJoki As @ragesoss alluded to, I think passing through the Search button as a prop is a bit odd. Since we're not using the `child` prop anywhere else at the moment, can we just pass in a prop like `hasSearchButton` or `showSearchButton` that toggles whether or not that specific button shows? That seems like it'd make a bit more sense at the moment.

Otherwise, this looks great!

AmitJoki:
I felt that CSS was more like hard coded and wasn't adaptive or responsive
enough. I haven't seen flex display used in much places so I thought it was
odd.

Also I thought a button and the text input coming under one form group made
more semantic sense. So I thought of KISS and without any CSS at all it got
aligned perfectly well in all widths which is what we are aiming for.

Does that make sense? If we are to fix it via CSS, it will be convoluted I
think.

There's already some of the conditional renders that TextInput does. So I
thought it would be okay given it changes nothing at all.

AmitJoki:
Also this allows TextInput to be generic as it is now and yet be flexible
for cases like these.

If we add hasSearch props that will make it very specific.

bwreid:
@AmitJoki How about doing something like this instead?
https://stackoverflow.com/questions/25797048/how-to-pass-in-a-react-component-into-another-react-component-to-transclude-the

This is a bit more React-like and keeps the same idea.

bwreid:
I think this ends up being much better. Thanks @AmitJoki! No more comments from me, @ragesoss

PR #3195, opened 2019-08-11
Tracking filter

AmitJoki:
Adds a intuitive filter based on an article's tracked status and works in real time.


ragesoss:
This works really nicely! I'm very excited about rolling this out, and it's very close to ready.

I left my feedback for making the tests a little cleaner and for copyediting the interface message, in the other PR.

@bwreid could you give this a review? If you don't have any further changes you'd like to see, I think we can ship this tomorrow.

bwreid:
@AmitJoki @ragesoss This looks good to me!

PR #3202, opened 2019-08-14
Add articles count in category row

khyatisoneji:
**Screenshots**
**Before:**
image

**After:**
image


ragesoss:
This looks solid so far. We could merge it now, but I'll wait until you add the follow-up improvements we discussed today (adding `updated_at`... which ideally should also include a tooltip in the header explaining that the system will take a bit of time to import the list of articles included).

One other tweak that could improve this further would be to add URL support, so that clicking the title would take the user to the relevant place, either the (non-auto-run) PetScan page, or the category or template page on Wikipedia.

ragesoss:
This is excellent. The links should open in new tabs, other than that I think it's good to go.

PR #3205, opened 2019-08-15
Make slug editable along with validation

AmitJoki:
Before:

image

After:

image

### DIFF
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -156,9 +156,11 @@ class Campaign < ApplicationRecord
   end
 
   def set_slug
-    campaign_slug = slug.present? ? slug : title
-    # Strip everything but unicode letters and digits, and convert spaces to underscores.
-    self.slug = campaign_slug.downcase.gsub(/[^\p{L}0-9 ]/, '').tr(' ', '_')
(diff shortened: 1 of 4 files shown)


app/models/campaign.rb
```diff
@@ -156,8 +156,9 @@ def valid_start_and_end_dates?
   end
 
   def set_slug
+    campaign_slug = slug.present? ? slug : title
     # Strip everything but unicode letters and digits, and convert spaces to underscores.
-    self.slug = title.downcase.gsub(/[^\p{L}0-9 ]/, '').tr(' ', '_') unless slug.present?
+    self.slug = campaign_slug.downcase.gsub(/[^\p{L}0-9 ]/, '').tr(' ', '_')
```
ragesoss: Looks like this change is causing some test failures, probably in connection with the default campaign in the fixture.

AmitJoki: It is passing now. Should only normalize if it is not already valid.

bwreid:
@AmitJoki This looks good to me. We'll probably wait to officially merge this until Sage returns, but I don't believe any further work is required. Nice job!

AmitJoki:
@ragesoss can we get this merged?

PR #3265, opened 2019-09-06
Sidekiq 6.0 with systemd process management

ragesoss:
## What this PR does
Upgrades Sidekiq to version 6.0, documents the server-side configuration necessary to switch to using systemd services to replace sidekiq process daemonization, and replacing capistrano-sidekiq with explicit tasks to quiet, stop and start sidekiq processes during deployment.

## Notes
This will mean that sidekiq processes that die unexpectedly should automatically get restarted by systemd, sidekiq processes will also restart if the system reboots... removing a few of the rare gotchas for administering production servers.

Sidekiq 6.0 requires redis-server 4 or higher; I've already upgraded the three live dashboard (Wiki Education Dashboard, Programs & Events Dashboard, and staging) from redis 3 to 5 (which works fine with Sidekiq 5 already).

The systemd services have been enabled and tested (but not started) on each of the production servers. Before deployment, we should quiet the processes on the production servers and wait for them to finish their jobs, then stop them manually and start the systemd services. After that, they'll automatically get restarted during each deployment (unless we opt to skip restarting, which is now supported as well).


ragesoss:
I've switched all the processes to systemd services for P&E and Wiki Education production, and just merged to production which should deploy and restart them again.

There's one more problem to solve for P&E, which is that the non-root deployment user can't run `systemctl` commands.

PR #3281, opened 2019-09-10
Revision scores import flow

ragesoss:
This moves the ORES score importing into the course data update cycle, rather than the independent ConstantUpdate cycle, so that references added stats get updated accurately at the same time as the rest of the stats.


ragesoss:
The first commit had one failing test, which is fixed by the second commit. The second build is failing for unrelated reasons, I think.

PR #3328, opened 2019-09-26
Upgrade react-dropzone to version from 7 to 10

prionkor:
## What this PR does
`react-dropzone` upgraded to `10.X.X`

The project build was successful, all tests passed. Need someone to test the UI uploading process.


ragesoss:
Not sure what happened with the original build, but looks like this has conflicts now so if we try this upgrade again we should start fresh.

PR #3360, opened 2019-10-03
Add defaults for language and project to assignments

bwreid:
## What this PR does
This should solve the issue of wikidata articles being incorrectly formatted.

Once merged, we should also run the following on the backend:
```
Assignment.where('sandbox_url LIKE ?', 'https://.wikidata%').each { |a| a.save }
```


ragesoss:
LGTM, once rubocop is pacified.

PR #3361, opened 2019-10-03
Fix assign my self an article button

### LINKED ISSUE #3101: AssignButton does not respond to first click user navigates away and back while it is open
This seems like a conflict between local and redux state for the popover button. It might only happen in cases like the new link button within the open state of AssignButton, where the link is in the popover and so does not trigger the click-on-outside callback to close it (and update the state to closed in the UI reducer).

Kartik1397:
## Fixes assign button response to first click when user navigates away and back while it's open
Fixes #3101 
**Before:**
1. In the MyArticles section, click 'Assign myself an article'
2. Within the popover at the bottom, click 'Search Wikipedia for an Article'
3. Navigate back to the MyArticles tab
4. Click 'Assign myself an article' again. The first click does not open the popover, (I think) because the 'ui' reducer state has it marked as open (and so the click sets it to closed).

**After:**
1. In the MyArticles section, click 'Assign myself an article'
2. Within the popover at the bottom, click 'Search Wikipedia for an Article'
3. Navigate back to the MyArticles tab
4. Click 'Assign myself an article' again. popover will open.

### DIFF
--- a/app/assets/javascripts/components/students/assign_button.jsx
+++ b/app/assets/javascripts/components/students/assign_button.jsx
@@ -193,12 +193,12 @@ const EditButton = ({
   );
 };
 
-const FindArticles = ({ course }) => {
+const FindArticles = ({ course, open }) => {
   return (
     <tr className="assignment find-articles-section">
       <td>
         <Link to={`/courses/${course.slug}/article_finder`}>
-          <button className="button border small link">
+          <button className="button border small link" onClick={open}>
             Search Wikipedia for an Article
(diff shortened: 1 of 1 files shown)


ragesoss:
Nice and simple! Well done. I'll merge assuming the tests pass.

ragesoss:
@bwreid is this going to conflict with your branch?

bwreid:
@ragesoss If it does, it should be minor. I can rebase after this is merged in.

PR #3372, opened 2019-10-06
Makes a popover in a modal responsive.

codingmamakaz:
## What this PR does
Makes a popover in a modal responsive.

## Screenshots 
Before:

Screen Shot 2019-10-06 at 11 14 23 AM

After:

Screen Shot 2019-10-06 at 11 14 54 AM
Screen Shot 2019-10-06 at 11 15 13 AM

Hi! I pushed a PR earlier and realized I added schema file with my commit. I created another PR to keep things clean. This is my first PR. Rspec is not passing, but all I touched was CSS, so I'm assuming I can ignore it. Npm tests are passing. I will appreciate any feedback!

### DIFF
--- a/app/assets/stylesheets/modules/_course.styl
+++ b/app/assets/stylesheets/modules/_course.styl
@@ -172,6 +172,7 @@ rating(color)
   .details-form
     display flex
 
+<<<<<<< HEAD
   .pop__container
     .pop.open
       position relative
@@ -183,6 +184,8 @@ rating(color)
         justify-content center
          
 
+=======
+>>>>>>> parent of 7e55a95b7... WIP trying to make pop container responsive
 @media only screen and (max-width: 920px)
   .course_main,
   .campaign_main
@@ -206,7 +209,6 @@ rating(color)
       .course-details_value
         font-size 15px
 
-
   .modal-course-details
     top 100px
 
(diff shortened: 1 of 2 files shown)


ragesoss:
Thanks! It looks like this is headed in the right direction.

The "+" button looks like it gets shifted to the next line on the left, in the 'after' screenshot, and the arrow that is supposed to connect the popover to the "+" button isn't lined up with that button.

Also, there's still a schema.rb change included in this PR. This is basically a benign change, since the schema.rb gets regenerated from your local database whenever you run migrations but the format and defaults can change between Rails versions and/or based on the local database environment. But it should not be included in the PR. What I usually do is `git stash` immediately after running migrations, unless I'm working on something that is supposed to change the schema.

ragesoss:
The easiest way, since there are just a few changes, will be to make a new branch from `master`, then copy back in the stylesheet changes and make anew commit with just those.

codingmamakaz:
@ragesoss Would this work?

Full screen
Screen Shot 2019-10-07 at 3 16 54 PM

Width 1270
Screen Shot 2019-10-07 at 3 17 04 PM

Mobile
Screen Shot 2019-10-07 at 3 16 29 PM

codingmamakaz:
I started a new branch,, so it does not include the schema this time. I will open a new PR and close this one if these changes work.

ragesoss:
The latest screenshots look good. Is that username icon something your browser adds? I've never seen that before.

codingmamakaz:
It must be. I didn't add any in the code. Okay, I will open a new PR with the new changes.

PR #3375, opened 2019-10-07
Makes a popover in a modal responsive

codingmamakaz:
## What this PR does
Makes a popover in a modal responsive.
## Screenshots
Before:
Screen Shot 2019-10-07 at 3 44 36 PM
After:
Full screen
Screen Shot 2019-10-07 at 3 16 54 PM

Width 1270px
Screen Shot 2019-10-07 at 3 17 04 PM

Mobile
Screen Shot 2019-10-07 at 3 16 29 PM

### DIFF
--- a/app/assets/stylesheets/modules/_course.styl
+++ b/app/assets/stylesheets/modules/_course.styl
@@ -172,6 +172,16 @@ rating(color)
   .details-form
     display flex
 
+  .pop__container
+  //.pop.open
+    //position relative
+    //width 100%
+
+    form
+      display flex
+      align-items center
+      justify-content center
+
 @media only screen and (max-width: 920px)
   .course_main,
   .campaign_main
@@ -205,6 +215,11 @@ rating(color)
 
     .details-form
       display block
+
+    .pop__container
+      .pop.open
+        form
+          flex-direction  column  
 
 .multi-wiki [class$='multiValue']:first-of-type svg
   display: none;
--- a/app/assets/stylesheets/modules/_course.styl
+++ b/app/assets/stylesheets/modules/_course.styl
@@ -173,10 +173,6 @@ rating(color)
     display flex
 
   .pop__container
-  //.pop.open
-    //position relative
-    //width 100%
-
     form
       display flex
       align-items center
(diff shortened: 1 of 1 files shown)


khyatisoneji:
It looks great!.  But I think you should add some padding between the textbox and button

codingmamakaz:
Screen Shot 2019-10-08 at 7 22 13 AM
Screen Shot 2019-10-08 at 7 22 48 AM

How about this?

ragesoss:
I'm not sure why it failed. It might have been a failure to fetch one of the dependencies; looks unrelated to your code. I've restarted the build, which will probably pass if it was just an intermittent build bug.

codingmamakaz:
Ok. Thank you! It looks like it passed this time. Is there a way for us to restart the buid if it fails again in the future or is it for admin only?

ragesoss:
For re-running a build without adding a new commit, it's admin-only.

ragesoss:
This looks like a definite improvement. There are some intermediate widths where it still goes off the screen, though.

intermediate width

I think the vertical layout would work fine without being responsive; using the vertical will still look fine when the page is full-width. I recommend just using the vertical layout always.

codingmamakaz:
@ragesoss, can you tell me at which width it goes off the screen? I can't reproduce it in my browser. Would you prefer the popover to stay vertical even when the main form goes responsive at 920px?

ragesoss:
If it's just a little over 1270 for me, in Firefox:

over 1270

Yeah, I think just having it be vertical at all page widths would be a good solution.

ragesoss:
@codingmamakaz are you still working on this? The styling you did is good, and I'm eager to get it merged. It just needs to be redone as universal styling instead of based on screen width, and then it will be good to go.

Manasa2850:
@ragesoss I'll work on this. Could you let me know what exactly the issue is?

ragesoss:
@Manasa2850 the remaining thing to do is to update the change in this PR so that instead of switching to the vertical layout (see the 1270px screenshot from the first comment) only when the screen is narrow, it should use that layout all the time.

PR #3376, opened 2019-10-08
Correctly display articles for Review an article button

bwreid:
## What this PR does
This should fix an issue where completely new articles were showing up as available when clicking the "Review an article" button in the My Articles section. The problem is fixed by relying on `article_title` as the unique value as opposed to `artice_id`.

### DIFF
--- a/app/assets/javascripts/components/util/helpers.js
+++ b/app/assets/javascripts/components/util/helpers.js
@@ -34,7 +34,6 @@ export const groupByAssignmentType = (assignments, user_id) => {
 
   // To find articles that are able to be reviewed...
   const pluckArticleId = pluck('article_id');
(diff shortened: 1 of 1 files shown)


ragesoss:
I don't think article title guaranteed to be unique; you could have the same title on different wikis.

bwreid:
@ragesoss Ah, ok. I'll return to `article_id` then.

ragesoss:
I thought multiple articles with no ID is the source of a bug. Can you use title + wiki?

bwreid:
@ragesoss That _or_ carve out an exception for when `article_id` is `undefined`. I'll look at both and try and implement whatever seems cleaner.  :)  The previous exception wasn't working wholly but maybe there's a clever way to fix that.

ragesoss:
Looks good to me!

PR #3379, opened 2019-10-08
Feature/training module assignments

bwreid:
## What this PR does

Adds the `kind` attribute to some training modules that will allow for setting them as 'exercise' or 'discussion'. Additionally, update the frontend timeline to distinguish between these types.

## Screenshots
**Before:**
image

**After:**
image

### DIFF
--- a/lib/alerts/overdue_training_alert_manager.rb
+++ b/lib/alerts/overdue_training_alert_manager.rb
@@ -29,7 +29,7 @@ class OverdueTrainingAlertManager
     course.training_modules.each do |training_module|
       due_date_manager = TrainingModuleDueDateManager.new(course: course, user: student,
                                                           training_module: training_module)
-      overdue = true if due_date_manager.overdue? && training_module.is_training?
+      overdue = true if due_date_manager.overdue? && training_module.training?
       status[training_module.slug] = { due_date: due_date_manager.computed_due_date,
                                        status: due_date_manager.deadline_status,
                                        progress: due_date_manager.module_progress }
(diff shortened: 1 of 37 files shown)


ragesoss:
Yay!

It looks like 'Exercise' gets its own heading, but Discussion keeps the 'Training' heading? I like the idea of blocks separating these out with each getting its own heading (even if we don't actually have any default blocks that would include all three at once.)

The 'Exercise:' and 'Discussion:' prefixes seem unnecessary since the type is indicated in the heading.

bwreid:
Screenshot updated. I also made a change so that the mid-title (e.g. "Training," "Exercise") will be hidden if it's the only assignment in the block.

ragesoss:
Looks like there's the `--` placeholder for the Discussion? That should be treated like Exercise, with no due date shown.

I think we'll also need to update the overdue training alert generation to make sure we don't send emails about overdue training for the new module types.

ragesoss:
We also need to make sure that the new module types don't contribute to counting students as `overdue`. I think this would just involve adding a guard clause to `TrainingModuleDueDateManager#overdue?` so that never returns true for the other types.

ragesoss:
I still get the 'Training Assignment (due date)' warning for Exercise and Discussion modules on the index pages:

due date

I think that should be removed for these new types.

lib/alerts/overdue_training_alert_manager.rb
```diff
@@ -29,7 +29,7 @@ def manage_alert(course, student)
     course.training_modules.each do |training_module|
       due_date_manager = TrainingModuleDueDateManager.new(course: course, user: student,
                                                           training_module: training_module)
-      overdue = true if due_date_manager.overdue?
+      overdue = true if due_date_manager.overdue? && training_module.is_training?
```
ragesoss: Maybe instead, it should do `next if training_module.is_training?` at the top of the `each` block. That way, no status will get inserted into the `status` hash for the other module types. `status` here ends up getting converted to a table in the email, so we probably don't want to include the non-trainings at all: https://dashboard-testing.wikiedu.org/rails/mailers/overdue_training_alert_mailer/message_to_student

ragesoss:
LGTM!

PR #3380, opened 2019-10-09
Training modules users fixup

ragesoss:
This is part one out of two.

The controller change should prevent creation of new records that have `user_id: 0`.

The cleanup script removes those records, along with the extra copies of records for the same user+module, which we need to do before adding an uniqueness constraint to the database. It will still be possible to create new duplicates in the meantime, though, so this will need to be run right before part two gets deployed with the schema change.


ragesoss:
@bwreid a double-check of the database cleanup code would be especially helpful.

PR #3383, opened 2019-10-10
Import average views of articles during course data update

ragesoss:
This lays the groundwork for switching to average views as the basis for pageview numbers, so that we can show estimated pageviews more immediately and update them indefinitely for older courses, and we won't have the current unsustainable view import bottleneck.

First step: deploy this, which will start importing Article data.

Second step: replace the revision-based view summing with estimates that are based on average views and time since first edit.


bwreid:
Is this class (`AverageViewsImporter`) being used by something else in our code base? It doesn't look like it has a connection to any other file.

ragesoss:
No, it doesn't. It's been around for a while because it used to be part of an on-demand category fetcher that would find the average views of each article in a category... which was superseded by ArticleFinder (which does basically the same thing, client-side). I never deleted it, though, because switching to using average views for the view stats has been on the the backburner for a long while and I thought it would likely come in handy.

PR #3396, opened 2019-10-14
Add MyArticles section to Articles main page

bwreid:
## What this PR does

This adds the MyArticles section to the Articles main page. It essentially works as intended but there's a couple notes I have:

1. Is this entirely necessary? There is a notification that should show up each time someone assigns an article to themselves.
2. When all Assigned Articles are gone, we currently have it redirect to the "Assigned Articles" page. This makes it a bit confusing when selecting the last article. Is there somewhere else this should redirect?

## Screenshots
image

### DIFF
--- a/app/assets/javascripts/components/articles/available_articles.jsx
+++ b/app/assets/javascripts/components/articles/available_articles.jsx
@@ -7,6 +7,7 @@ import { Link } from 'react-router-dom';
 import AssignCell from '../students/assign_cell.jsx';
 import ConnectedAvailableArticle from './available_article.jsx';
 import AvailableArticlesList from '../articles/available_articles_list.jsx';
+import MyArticlesContainer from '../overview/my_articles/containers';
 import { ASSIGNED_ROLE } from '../../constants';
 
 const AvailableArticles = createReactClass({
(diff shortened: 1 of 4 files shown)


ragesoss:
It's probably not *entirely* necessary, but LiAnna said she saw more than one person at the editathon confused about the interaction — what had just happened, and what they should do next ­— after they chose an article.

Maybe an alternative would be to make the notification harder to miss, and/or more explicit?

As for the redirection, I didn't realize it did that. Maybe just don't redirect it? I think we hide the tab link that takes you to the Available Articles if you're a student and there are none available; that seems good enough.

app/assets/javascripts/components/articles/available_articles.jsx
```diff
@@ -64,6 +65,13 @@ const AvailableArticles = createReactClass({
     }
 
     const showAvailableArticles = elements.length > 0 || this.props.current_user.isAdvancedRole;
+    const showMyArticlesSection = this.props.current_user.isStudent;
```
ragesoss: I was thinking this could be shown only once there is at least one assigned article. That way, the list of available articles is still the main focus until you've chosen an article, and then MyArticles appears basically just to visualize "where" the article you chose went.

ragesoss:
I like it! I think this interaction will have little room for leaving people confused.

Issue #2888, opened 2019-05-19
Extract wiki language/project selectors into a reusable component


ragesoss:
We have several places in the UI that allow the user to select a wiki. The most relevant ones are ArticleFinder and AssignButton, which both have a show/hide link for changing the wiki. Similar but not quite the same is the home wiki selector in CourseCreator and Details.

At least the ArticleFinder and AssignButton ones, and possibly the others as well, should be extracted into a reusable component (or perhaps several components).

AmitJoki:
I think this can be closed now.

xiaobaioo7:
@ragesoss Can I take this one?

ragesoss:
Sure, go for it.

xiaobaioo7:
@ragesoss I started looking into the 2 components "ArticleFineder" and "AssignButton", but I am not sure about what these 2 components do. Can you give a brief explanation on what they do? Or a quick walkthrough or steps to get to these components will be really appreciated.

bwreid:
Hi @xiaobaioo7! Thanks for the question.

The `ArticleFinder` component is just what it sounds like -- it manages the view that makes it easier for students and instructors to find available articles. In your screenshot above, your such on "Buddha" resulted in multiple articles that include "Buddha" in the title.

The `AssignButton` component is actually a bit different than what you've circled on the screenshot above. In the above screenshot, clicking the button "Add as an available article" will add the chosen article as a potential option for students to pick. You can see that list under the "Articles" tab.

The link that allows you to change the Wiki is in the above screenshot you've posted underneath the Search Bar where it says "en.wikipedia.org". You can also see it in the below screenshot under the My Articles section. This view is only available to you as a student/editor. This is also the `AssignButton` component.
![image](https://user-images.githubusercontent.com/1316902/66770605-eb1dfb00-ee6c-11e9-84de-47ae576c6f8e.png)
Does this answer some of your questions?

xiaobaioo7:
Hi @bwreid, thank you for the information! This certainly helps 👍 

Correct me if I am wrong, the request here is to extract the piece like "en.wikipedia.org (Change)" from the mentioned components to its own reusable component, is that correct? 

As a side note (which may not relate to this issue), I encountered some issue when I clicked on the button "Add as available article" from the article finder page. I think I might know what's causing this issue, I created a separate issue to track this #3395 .

bwreid:
Thanks for the additional issue, @xiaobaioo7 

[...]

PR #3402, opened 2019-10-15
Update component ArticleFinder to reuse component SelectedWikiOption

xiaobaioo7:
## What this PR does
Update component `ArticleFinder` to use existing component `SelectedWikiOption`. This is to consolidate the component for selecting wiki and show the selection. The behavior in `ArticleFinder` is slightly different from `SelectedWikiOption`, where it includes the UI to hide the wiki selection.

With the changes, the UI in component `SelectedWikiOption` should remain the same. The UI in component `ArticleFinder` is changed to remove UI for "hiding the wiki selection" when it's shown.

Issue #2888 

## Screenshots
UI for component `ArticleFinder`.
Before:
image

After:
image

The issue also mentioned possibly updating components `CourseCreator` and `Details`. This PR does not include changes for these components. Will look into these next.

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -4,8 +4,7 @@ import { connect } from 'react-redux';
 import InputRange from 'react-input-range';
 import _ from 'lodash';
 import qs from 'query-string';
-import selectStyles from '../../styles/select';
-import WikiSelect from '../common/wiki_select.jsx';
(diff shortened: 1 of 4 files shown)


ragesoss:
Thanks! I don't think the 'Hide' option is needed. The idea is to minimize the switcher by default since in most cases, the currently selected wiki is the right one. But if someone does choose to change it, there's little point to minimizing it again afterwards.

xiaobaioo7:
Okay, that sounds good. I will update the `SelectedWikiOption` to remove the 'Hide' option.

PR #3410, opened 2019-10-17
Feature/complete training modules

bwreid:
## What this PR does
This PR performs a few updates:
1. Add flags to the Training Module table
1. Training Modules of the "Exercise" kind can now be marked as "Complete" or "Incomplete" manually by the student

## Screenshots

Until exercises have been read through (all slides completed), nothing will show.
image

After exercises have been read through, a button will appear which will allow for someone to mark as complete.
Oct-17-2019 13-59-42

As of **October 22nd**, the due date is listed next to exercises and a disabled button is shown if the module has not been read.
image

If `enableAdvancedFeatures` is _not turned on_, the view will look like the following:
image

Still to be completed:
- [x] Update/check styling for all pages where Training Modules appear (e.g. Home, Resources)
- [x] Include more tests
- [x] What should we do for discussions?
- [x] What changes should we include in particular for the resources page? (See below)


ragesoss:
This looks great. For the Resources tab, I guess the first thing that comes to mind would be just to add additional sections the exercises and discussions, and treat those the same as the categories we already use.

bwreid:
@ragesoss A couple of questions for you:

1. Right now I am not allowing for Discussions to be marked as In/Complete. Any disagreement on that?
1. Is the styling unclear at the moment as to why some buttons are showing and others are not? I don't know if we need something in place of button for exercises that have yet to be completed...

image

ragesoss:
I think the Discussion handling is good.

I'd say it'll probably be clearer if we render a disabled 'Mark Complete' button, maybe with a tooltip as well to explain that they need to View the instructions (and follow them) before marking it as complete.

ragesoss:
This looks great. I tested it with and without the ENABLE_ADVANCED_FEATURES flag, and I can't think of anything that should stop merging and deploying this now.

PR #3424, opened 2019-10-24
[WIP] Display Exercises on Course Home page

bwreid:
## What this PR does

This begins the work for showing exercises on the course home page.


ragesoss:
The remaining exercise link doesn't work; it's missing the `/courses/` part.

bwreid:
@ragesoss Added the `exact` keyword and it should be working as intended now.  :)

ragesoss:
Cool, looks good to me.

One minor improvement would be to use an anchor link to scroll down to the exercises. There's probably not quite enough content in Resources to make Exercises appear at the top for most screen sizes, but at least that will ensure that the exercises section is above the fold.

bwreid:
Ah, good call. I'm using `center` as a reference for where to scroll to as the navbar tends to hide the section heading if `top` is used. That should make it a bit more apparent where to look.

PR #3425, opened 2019-10-25
Added pagination for upload API

### LINKED ISSUE #2996: [Not urgent] Add Pagination in Uploads API.
in Some courses like https://outreachdashboard.wmflabs.org/courses/WMFr/EDITATHONS_2017/uploads.json ,the uploads API is taking a lot of time to give response [this course has around 500 images]. We can implement pagination in the API to give results by using parameter page_number.

ANURADHAJHA99:
It solves the issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/2996. 
1. Modified the files `courses-controller.rb` and `uploads.json.jbuilder`
2. closes #2996

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -113,6 +113,7 @@ class CoursesController < ApplicationController
 
   def uploads
     set_course
+    set_page
   end
 
   ##########################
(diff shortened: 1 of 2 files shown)


ragesoss:
It looks like that code in the jbuilder template is incorrect and causes an error. Please test out the uploads endpoint locally to make sure it works after the change; you can also run the same test that broke in the travis-ci build to make sure it doesn't interfere with the current functionality.

PR #3436, opened 2019-10-29
Changed Subject field to a multi-select dropdown Issue:  #2499

puthalalitha:
## What this PR does
The purpose of this pull request is to change Subject field to a multi-select dropdown menu in the course creation flow.

## Screenshots
Before:
Screen Shot 2019-10-29 at 12 15 34 PM

After:
Screen Shot 2019-10-27 at 9 19 43 PM

None

### DIFF
--- a/app/assets/javascripts/components/course_creator/course_subject_selector.jsx
+++ b/app/assets/javascripts/components/course_creator/course_subject_selector.jsx
@@ -12,13 +12,13 @@ const CourseSubjectSelector = ({ updateCourse }) => {
       .then(resp => resp.json())
       .then((data) => {
         const jsonObj = data;
-        const jsonObjects = jsonObj[5].options;
+        const subjectOptions = jsonObj[5].options;
 
-        Object.keys(jsonObjects).forEach((key) => {
-          if (jsonObjects.hasOwnProperty(key)) {
+        Object.keys(subjectOptions).forEach((key) => {
+          if (subjectOptions.hasOwnProperty(key)) {
             opts.push({
-              value: jsonObjects[key].title,
-              label: jsonObjects[key].title
+              value: subjectOptions[key].title,
+              label: subjectOptions[key].title
             });
           }
         });
@@ -47,5 +47,7 @@ CourseSubjectSelector.propTypes = {
 };
 
 export default CourseSubjectSelector;
+
+
 
 
--- a/app/assets/javascripts/components/course_creator/course_subject_selector.jsx
+++ b/app/assets/javascripts/components/course_creator/course_subject_selector.jsx
@@ -6,14 +6,18 @@ const CourseSubjectSelector = ({ updateCourse }) => {
   const [options, setOptions] = useState();
 
   useEffect(() => {
+    // Will contain an array of subjects in the required object format
     const opts = [];
 
     fetch('/wizards/researchwrite.json')
       .then(resp => resp.json())
       .then((data) => {
         const jsonObj = data;
+        // Exctracting the 5th index of the researchwrite.json file that contains all the subjects
+        // Index number may change in the future if additional information is added to researchwrite.json.  
+        // Please update as needed
         const subjectOptions = jsonObj[5].options;
-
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/course_creator/course_subject_selector.jsx
```diff
@@ -0,0 +1,55 @@
...
+  let selectedOption = { value: subject, label: subject };
+
+  const handleChange = option => {
+    const courseSubject = option.value;
+    selectedOption = option;
+    updateCourse('subject', courseSubject);
+  };
+
+  useEffect(() => {
+    let opts = [];
```
khyatisoneji: use const

ragesoss:
Note that `react-select` has a config prop you can use to allow users to add an option to the list.

puthalalitha:
User can create their own subject.

ragesoss:
It looks like this doesn't persist the subject when you create the course now, although I'm not sure why.

This should not be a multi-select field; we want to collect just one Subject, but it can either be one from the list or a user-provided entry. It should behave just like the existing 'Your role in the course' field, which offers two options in the dropdown but you can also type in something different.

puthalalitha:
@ragesoss , I now made it so that user can only select or create one subject, exactly like in 'Your role in the course' field. 
Regarding persistence, where/ how should the subject persist? Do you mean step 8 of the course creation process?

ragesoss:
By persist, I mean once the course has been created, the 'subject' should be saved in the database, and should be part of the attributes that are shown when you use the Edit Details form for the course.

When I tested this yesterday, after creating the course, the Subject attribute was empty.

puthalalitha:
@ragesoss , As you said the subject is saved in the database and after creating the course. When I enter into the Edit Details I am able to see the subject attribute.

app/assets/javascripts/components/course_creator/course_subject_selector.jsx
```diff
@@ -0,0 +1,51 @@
...
+  const [options, setOptions] = useState();
+
+  useEffect(() => {
+    const opts = [];
+
+    fetch('/wizards/researchwrite.json')
+      .then(resp => resp.json())
+      .then((data) => {
+        const jsonObj = data;
+        const jsonObjects = jsonObj[5].options;
```
ragesoss: This needs to be documented, as it's not going to be obvious what is going on here and how it is coupled to the researchwrite.yml file; anything that addes to content above the subject choices will require changing the index here.

ragesoss: `jsonObjects` is not a very informative variable name here. Maybe `courseSubjects` or `subjectOptions`?

puthalalitha: Replaced the variable `jsonObjects` into `subjectOptions`.

ragesoss: Great. Just needs some inline comments explaining what's going on, especially with `jsonObj[5].options`, and then it will be ready to merge.

puthalalitha: Thank you I will provide inline comments.

app/assets/javascripts/components/course_creator/course_subject_selector.jsx
```diff
@@ -0,0 +1,57 @@
...
+
+    fetch('/wizards/researchwrite.json')
+      .then(resp => resp.json())
+      .then((data) => {
+        const jsonObj = data;
+        // Exctracting the 5th index of the researchwrite.json file that contains all the subjects
+        // Index number may change in the future if additional information is added to researchwrite.json.  
+        // Please update as needed
+        const subjectOptions = jsonObj[5].options;
+        // itarating through all the subjects and copying to the opts array
```
ragesoss: This comment isn't necessary; in general, we only need comments that explain _why_ we're doing something, not _what_ we're doing, unless it's a particularly complicated, confusing, or unusual code pattern.

PR #3437, opened 2019-10-30
Refactor MyArticles section to use @components

bwreid:
A refactor to use the shortened `@components` and `~` in the MyArticles section, where it is much needed.


ragesoss:
Nice.

Issue #1088, opened 2016-12-19
Diff Viewer shows only one column, on left, if all changes are new paragraphs


ragesoss:
When there are no 'minus' paragraphs to show, then the whole table shifts so that added content is on the left.

As of now, that's the case with the Dodd-Frank article here: https://dashboard.wikiedu.org/courses/Northeastern/Advanced_Writing_in_the_Business_Administration_Professions_-_online_%28Fall_2016%29/articles

![screen shot 2016-12-19 at 1 02 19 pm](https://cloud.githubusercontent.com/assets/848483/21324151/70fd4e9c-c5d5-11e6-9b39-a475c00abe3a.png)

dnahol:
Hi, @ragesoss! I'm new to this repo. I'm assuming you would like the 'plus' columns to always be on the right, as they are here?

<img width="1197" alt="Screen Shot 2019-11-05 at 6 26 00 AM" src="https://user-images.githubusercontent.com/2976514/68216194-628b0a00-ff95-11e9-9794-9f2bd9862594.png">

dnahol:
Apparently changing the colspan from 2 to 4 of the .empty-diff element on the left changes the issue. Does that look like an acceptable solution that won't break anything else?

Screenshot in Firefox Devtools
<img width="1160" alt="Screen Shot 2019-11-05 at 7 24 26 AM" src="https://user-images.githubusercontent.com/2976514/68221010-94a06a00-ff9d-11e9-85f9-60666cda26ec.png">

dnahol:
@ragesoss what do you think?

dnahol:
@ragesoss  @khyatisoneji Should we close this issue?

khyatisoneji:
Yes

PR #3472, opened 2019-11-09
add ref to diff, resize empty diff col, wrap Loading element in tbody, rename certain react components UNSAFE_

dnahol:

## What this PR does

- Address issue  #1088 Diff Viewer shows only one column, on left, if all changes are new paragraphs 
- Fixes a bug within the diff table

Files changed: 
app/assets/javascripts/components/revisions/diff_viewer.jsx
- Add ref to diff:
  The Dangerously Set HTML from the Wikipedia Diff URL did not previously have a ref. I added one, including a setDiff method for it.

- Resize empty diff column: 
  When the table cells, td elements, are empty, they previously lost their width. This is what was causing issue #1088. This PR sets the first empty cell to 50% width in the table, so that both (-) and (+) columns can take up half the table width.

- Wrap Loading and (-) diff marker elements in tbody:
  There were previously errors caused by the <Loading/> and (-) diff marker elements because they were within a table without being wrapped in tbody. This PR wraps both in tbody, tr, and td, fixing the errors.

## Screenshots
Before:
Screen Shot 2019-11-05 at 6 39 10 AM

After:
Screen Shot 2019-11-08 at 10 27 33 PM

I learned how to use dangerouslySetInnerHTML in React, how to add refs to elements using JSX in React components, and that divs within tables must be wrapped in a tbody, tr, and td.

### DIFF
--- a/app/assets/javascripts/components/activity/activity_table.jsx
+++ b/app/assets/javascripts/components/activity/activity_table.jsx
@@ -29,7 +29,7 @@ const ActivityTable = createReactClass({
     };
   },
 
-  componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
     if (nextProps.activity !== this.state.activity) {
       this.setState({ activity: nextProps.activity });
     }
--- a/app/assets/javascripts/components/activity/did_you_know_handler.jsx
+++ b/app/assets/javascripts/components/activity/did_you_know_handler.jsx
@@ -23,13 +23,13 @@ const DidYouKnowHandler = createReactClass({
     loading: PropTypes.bool
   },
 
-  componentWillMount() {
-    this.props.fetchDYKArticles();
-  },
-
   setCourseScope(e) {
     const scoped = e.target.checked;
     this.props.fetchDYKArticles({ scoped });
+  },
+
+  UNSAFE_componentWillMount() {
+    this.props.fetchDYKArticles();
   },
 
   render() {
--- a/app/assets/javascripts/components/activity/plagiarism_handler.jsx
+++ b/app/assets/javascripts/components/activity/plagiarism_handler.jsx
@@ -24,13 +24,13 @@ const PlagiarismHandler = createReactClass({
     loading: PropTypes.bool
   },
 
-  componentWillMount() {
-    return this.props.fetchSuspectedPlagiarism();
-  },
-
   setCourseScope(e) {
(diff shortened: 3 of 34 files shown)


khyatisoneji:
I tested it out, it works smoothly, great work!

khyatisoneji:
As mentioned by @ragesoss, it still gives the warning to convert components( like componentWillMount, componentWillReceiveProps) to use new lifecycle methods

Let us know if you need any help

dnahol:
Hey @khyatisoneji, thanks for looking it over. I looked at the errors and the React documentation. I tried changing `componentWillReceiveProps` in `app/assets/javascripts/components/revisions/diff-viewer.jsx` to `componentDidUpdate` and I am left with diffs that won't load. I tried the same for `app/assets/javascripts/components/common/article-viewer.jsx` with the same result.

I think that properly updating all the soon-to-be-deprecated `componentWillMount` and `componentWillReceiveProps` methods is going to be a separate refactoring using the new React version way of doing things. 

What do you think about me opening a second issue to work on this?

In the meantime, I ran this command from the documentation to rename such methods in `app/assets/javascripts`:

`npx react-codemod rename-unsafe-lifecycles` 

`componentWillReceiveProps()` -> `UNSAFE_componentWillReceiveProps()`
`componentWillMount()` -> `UNSAFE_componentWillMount()`

khyatisoneji:
You can open another PR to remove all warnings, this PR would only involve changing only a few files
(article_handler.jsx, article_finder.jsx, nav.jsx, diff_viewer.jsx, article_viewer.jsx)

khyatisoneji:
The errors would be shown in the terminal(only for frontend code), or you can run tests locally to check for any errors

dnahol:
Hi @khyatisoneji. Thank you, I hadn't noticed the linter errors- I think I created them by running `npx react-codemod rename-unsafe-lifecycles` in my previous commit to fix the React lifecycle warnings. 

I fixed the 18 linter errors and squashed the commit.

ragesoss:
Cool! This looks like a nice improvement. I'm going to merge it now.

One way to improve it a little more will be to handle the Loading state better. After this change, the loading animation includes a border, which is a little distracting. Ideally, the loading animation should span the entire DiffViewer element instead of just the left portion.

dnahol:
Hey @ragesoss I see what you mean. I tried a couple of things- no luck yet. I'll open a separate issue to deal with this.

PR #3474, opened 2019-11-10
Convert `/alerts_list` to React

chrisnorwood:
## What this PR does
fix for WikiEducationFoundation#1803

* Refactors AlertsList component into seperate CampaignAlerts and AdminAlerts components with shared functionality
* Allows for asynchronous 'resolving' for AdminAlerts
* Campaign alerts functionality not changed

Notes: In `/alerts_list` the alert-type links to the "alert_lists/:id"/show page, similarly to the previous version, though the resolve button now handles requests asynchronously.

## Screenshots
Before:
1803-before

After:
1803-after

I notice that the alert types listed in the `AlertsHandler` React component's ALERTS constant differ from the list of types defined within the `Alert` rails model's frozen-constant ALERT_TYPES, or PUBLIC_ALERT_TYPES even, for that matter. (For example, HighQualityArticleEditAlert is not available for filtering in the React component). I did not modify those in this PR, as it felt outside of this issue's scope.  Which alert types belong in the React component's searchable list?  Should these be modified in another PR perhaps?


chrisnorwood:
I see all the failing tests are coming from the alerts_list_controller_spec. I see that most of the failures are due to React rendering in the HTML response that is returned to RSpec.  Furthermore, previously, the page could filter results with query parameters.  Functionality has changed now a bit, and filtering happens within a React component, without query strings (on that note, though, the API endpoint will still filter the JSON results exactly as before if query strings are used. (I could pass this through to the API call within the react component if desired, though that seems a bit hacky...)

Do you have any suggestions on how to reconcile these modifications with the current controller test?

chrisnorwood:
I just tried to fix the conflicts due to the recent PR that was merged to master, and I got super confused.  Now somehow all these new commits were introduced, yet the conflict still exists.  Any advise on how to clean this up/get back to a "non-conflicting/ready-to-be reviewed" state?

ragesoss:
Nice work!

Regarding the tests, I think your strategy was a good one: switching to json endpoints and looking for the right content in the json data instead of the html.

The `type` param support for alerts is intended for use by `fetchOnboardingAlert`, which is used on course pages; although it was technically possible to use a URL param to filter based on type for `/alerts_list`, that isn't a feature we need to preserve.

I see you moved the default selections to the new campaign-specific alerts component, so that we keep the current behavior of setting the default filter for that list, but not doing any filtering by default for the admin list — which is great. The list of possible filter options in AlertHandler is also specific to the campaign use case, as there are additional alert types (the non-public ones) that can also appear in the admin list. Maybe the right approach here would be to just populate a list of options for the filters based on all the alert types that are received; for the admin view, they could use the type as the label, rather than the slightly-more-human-friendly labels that get defined AlertHandler currently.

ragesoss:
The easiest way to sort out the git history here is probably to cut a new branch from the latest master, then cherry-pick your commits onto it.

The other possible approach would be to run `git rebase master` (after getting master up to date), although that might be slightly more complicated. Either way, you'll likely to need to manually resolve the merge conflict in AlertsHandler; I think you can just keep your version of the file, as it removes the method that got renamed to UNSAFE_ anyway.

chrisnorwood:
Ok, thanks for that advice!!  I'll do that now, and re-incorporate the modifications into the new branch.  I'll rename the two ComponentWillMount's into ComponentDidMount while I'm at it, and make those couple filtering fixes based on your advise.

Perhaps, it will make the most sense to just make a new PR after that?

ragesoss:
Sure, a new PR is fine.

Issue #3482, opened 2019-11-14
Loading element in diff_viewer should take up whole table width


dnahol:
### Current Behavior:
From @ragesoss comment on #3472
> Cool! This looks like a nice improvement. I'm going to merge it now.
> 
> One way to improve it a little more will be to handle the Loading state better. After this change, the loading animation includes a border, which is a little distracting. 

### Desired Behavior:
From @ragesoss comment on #3472 
> Ideally, the loading animation should span the entire DiffViewer element instead of just the left portion.

### Related Files & Code:
app/assets/javascripts/components/revisions/diff_viewer.jsx

ishantgupta777:
@ragesoss  I would like to work on this issue.

DevanshKariya:
@ragesoss is this issue still open? I would like to work on it

ragesoss:
@DevanshKariya yes, go for it.

DevanshKariya:
@ragesoss Can you give some more info on this

ragesoss:
Hm... it looks like this got fixed at some point, as there's no border on the DiffViewer loading element.

DevanshKariya:
ok😅

PR #3483, opened 2019-11-14
Set diff table width to 100%

dnahol:
## What this PR does
As explained in Issue #3482 
The diff table width in diff viewer is set to 100% so that the Loading element animation can take up the full width of the diff table.

## Screenshots
Before:
loading element diff_viewer issue

After:
loading element diff_viewer PR

### DIFF
--- a/app/assets/stylesheets/modules/_diff_viewer.styl
+++ b/app/assets/stylesheets/modules/_diff_viewer.styl
@@ -23,6 +23,9 @@
   overflow-y: scroll
   overflow-x: hidden
 
+.diff-viewer-scrollbox > table
+  width 100%
+  
 .diff-viewer-footer
   border-top 1px solid mischka !important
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -346,7 +346,7 @@ const DiffViewer = createReactClass({
           <div>
             <div className="diff-viewer-scrollbox">
               <strong>{salesforceButtons}</strong>
-              <table>
+              <table style={{ width: '100%' }}>
```
ragesoss: I'd prefer this be done via the stylesheet, rather than setting the style at the component level.

dnahol: Ah good point. Fixed!

Issue #3478, opened 2019-11-12
Refactor componentWillMount and componentWillReceiveProps


dnahol:
### What is happening?

componentWillMount and componentWillReceiveProps were producing warnings in the browser console, saying that use of the soon-to-be-depricated componentWillMount and componentWillReceiveProps has to be refactored and phased out. 

Commit [a0321daf12d62479d84eff5729dafce4089111bf](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/3472/commits/a0321daf12d62479d84eff5729dafce4089111bf)
in PR #3472 renames these methods to the supported UNSAFE_ versions. 34 occurrences were replaced. 
 
The next step is to refactor these methods. 

### To Reproduce

1. Open dev code of project
2. Search whole project for UNSAFE_
3. See the React methods labeled UNSAFE_

### Expected behavior
A full refactor of the UNSAFE_ methods in accordance with the new React documentation.

### System Information

 - OS: MAC OS High Sierra Version 10.13.6
 - Browser Chrome Version 78.0.3904.70

### Additional context
From comment in PR #3472 
(This PR doesn't solve the issue, the conversation here explains the context)

> Hey @khyatisoneji, thanks for looking it over. I did notice the pre-existing warnings. Here is what I found. I am opening a separate issue to track that work: 
> 
> I looked at the errors and the React documentation. I tried changing `componentWillReceiveProps` in `app/assets/javascripts/components/revisions/diff-viewer.jsx` to `componentDidUpdate` and I am left with diffs that won't load. I tried the same for `app/assets/javascripts/components/common/article-viewer.jsx` with the same result.
> 
> I think that properly updating all the soon-to-be-deprecated `componentWillMount` and `componentWillReceiveProps` methods is going to be a separate refactoring using the new React version way of doing things. 
> 
> In the meantime, I ran this command from the documentation to rename such methods in `app/assets/javascripts`:
> 
> `npx react-codemod rename-unsafe-lifecycles` 
> 
> `componentWillReceiveProps()` -> `UNSAFE_componentWillReceiveProps()`
> `componentWillMount()` -> `UNSAFE_componentWillMount()`
> 
> 34 changes were made.
> 
> From [React Blog: Update on Async Rendering](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html)
> 
> > **Gradual Migration Path**
> > 
> > React follows semantic versioning, so this change will be gradual. Our current plan is:
> > 
[...]

PR #3498, opened 2019-11-15
Activity: refactor UNSAFE react methods

dnahol:
# What this PR does
One of several PRs coming to address Issue #3478.

Refactored any files in app/assets/javascripts/components/activity
to use the new react lifecycle react methods and stop using soon-to-be-deprecated UNSAFE_ react methods.

No bugs found from this change. Please let me know if you find any.

### Note:
Got rid of chunks of code in `app/assets/javascripts/components/activity/activity_table.jsx`.

Entirely removed `UNSAFE_componentWillReceiveProps`
and replaced instances of `this.state.activity` with `this.props.activity`

### Why the big change?
Use of `UNSAFE_componentWillReceiveProps` was only being used to keep `state.activity` matching `props.activity`. Since `activity` was already a prop, there was no need to keep `state.activity` as well. The React documentation calls it an "anti-pattern" that makes the application more bloated than necessary. 

[React: Why You Probably Don't Need Derived State](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state)

### DIFF
--- a/app/assets/javascripts/components/activity/activity_table.jsx
+++ b/app/assets/javascripts/components/activity/activity_table.jsx
@@ -23,18 +23,6 @@ const ActivityTable = createReactClass({
     toggleDrawer: PropTypes.func
   },
 
-  getInitialState() {
-    return {
-      activity: this.props.activity
-    };
-  },
-
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (nextProps.activity !== this.state.activity) {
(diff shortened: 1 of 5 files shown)


ragesoss:
Nice. Not sure why that was using state in the first place, but it looks like the origin goes way way way back to when we were using `.cjsx`. I'll test this out manually, but this looks like a solid refactor.

PR #3504, opened 2019-11-15
Swap out instances of fetch with a request helper method

bwreid:
## What this PR does

This PR standardizes the use of the `fetch` method to pull from the `cross-fetch` library. It also creates a helper `request` function.


ragesoss:
Looks like this fails with a CORS error for a request that goes to Wikimedia servers rather than the dashboard app server.

ragesoss:
LGTM

Issue #2842, opened 2019-05-08
Removing a course participant doesn't cause template to be removed from wiki


urbanecm:
If you remove a course participant, the participation template (for cs.wikipedia, it's Template:Nástěnka/Účastník) isn't removed from user page and user talk page.

xiaobaioo7:
@ragesoss can I work on this one? Noted this issue is added to "Prague Hackathon" project, not sure if this issue is still relevant, please advise.

ragesoss:
@xiaobaioo7 yes, go for it. It's something we identified as an issue during that Hackathon, but didn't get to it then.

xiaobaioo7:
Hi, I started looking into this issue and have a few questions.

The issue mentioned the problem with user page. If I understand this correctly, these are the steps I followed to reproduce the problem:

1. I removed an editor A from a program B
2. I went to the user page for the removed editor A. But I don't see any indication of the program B on the user page. 

I am not sure if this issue still exists for the user page, or I could be missing something. Would you take a look or guide me through the repro steps?

Also, the issue mentioned the problem with user talk page, but I can't seem to locate the page. Can you help guide me through the repro steps?
p.s.: I found somewhere in the code that points to a user talk url like `https://${language}.${project}.org/wiki/User_talk:${username}`. Is this the user talk page?

ragesoss:
Yes, you're on the right track.

These steps won't let you reproduce the problem locally, though, unless you your development environment configured to enable wiki edits.

For example, here's the user page for a student in one of the current course: https://en.wikipedia.org/wiki/User:Jfitz19

That template pointing to the course got added as one of the actions fired off by `EnrollInCourseWorker`, specifically, `WikiCourseEdits#enroll_in_course`.

The idea here would be to have a similar job for cleanup when a user leaves a course, which would search for and remove the course template from the userpage.

Yes, that's a user talk page at the end of your post. For example, https://en.wikipedia.org/wiki/User_talk:Jfitz19

xiaobaioo7:
Thanks @ragesoss for the detailed explanation! 

You mentioned that this cannot be reproduced locally unless the development environment is configured to enable wiki edits. Can you shed some light on how to do that? Is that a configuration at the application level or at the wiki user level?

[...]

PR #3505, opened 2019-11-16
[WIP] Remove templates from user (and talk) pages when disenrolled

xiaobaioo7:
## What this PR does
When user is dis-enrolled from course, it makes updates to user page and user talk page to remove existing enrollment-in-course templates (if any). This intends to revert the addition to the pages when enrolling the user in course.

Issue #2842 

What are you thoughts about testing this?


xiaobaioo7:
Hmm.. have been fixing the build, it looks like the build now has 2 blocking issues with class length over sized. How would you approach that?
image

ragesoss:
That can be handled editing `.rubocop.yml` to either change the limit or exclude those files from the class length rule... or by refactoring some of the functionality of those large classes into new, if you see a way to do that cleanly.

xiaobaioo7:
Thanks @ragesoss , I would think refactoring the large classes are the ultimate goal (and this is what rubocop would be used for). I would like to take another look into the effort. But for now, I temporarily excluded the 2 large classes from `.rubocop.yml`.

xiaobaioo7:
About the 2 large classes, here are some of my thoughts on refactoring them:

- `UsersController`: we may extract the logic for user enrollment and dis-enrollment into a separate controller something like `UserEnrollmentsController`
- `WikiCourseEdits`: how about extracting the enrollment related logic into a separate class like `WikiCourseEnrollmentEdits`?

ragesoss:
Thanks @xiaobaioo7! I think those sound like the right approaches to refactoring this.

xiaobaioo7:
Thanks @ragesoss ! With an attempt to separate the enrollment logic from the 2 large classes into new ones, it seems to be passing the class max length rule from rubocop. In addition to the refactoring, I made some minor changes. Please see my inline comments.

lib/wiki_course_edits.rb
```diff
@@ -204,7 +145,7 @@ def homewiki_assignments_grouped_by_article
   end
 
   def update_assignments_for_article(title:, assignments_for_same_article:)
-    return if WikiApi.new(@home_wiki).redirect?(title)
+    return if @wiki_api.redirect?(title)
```
xiaobaioo7: Consider to use the variable from the constructor instead.

lib/wiki_course_edits.rb
```diff
@@ -101,7 +85,6 @@ def remove_assignment(assignment:)
 
   private
 
-  # rubocop:disable Metrics/CyclomaticComplexity
```
xiaobaioo7: Will this work still by removing the comment for rubocop at this line and one below to disable/enable Metrics/CyclomaticComplexity?
Running rubocop locally with the comments in place seems to produce offenses.

ragesoss:
This looks like it was pretty close to being complete, but I must not have looked at it again after the refactor because it was still marked as [WIP]. This will probably be useful for someone else if they take up the issue.

PR #3506, opened 2019-11-16
Article Finder: refactor UNSAFE react methods

dnahol:
# What this PR does
One of several PRs coming to address Issue #3478.

Refactored any files in app/assets/javascripts/components/article_finder
to use the new react lifecycle react methods and stop using soon-to-be-deprecated UNSAFE_ react methods.

No bugs found from this change. Please let me know if you find any.

### Note:
Larger refactor in `app/assets/javascripts/components/article_finder/article_finder_row.jsx`.

Entirely removed `UNSAFE_componentWillReceiveProps`
and replaced custom async function wrapped around an `await` 
which sets `state.isLoading` once `await` is resolved

### Why this change?
The new React lifecycle method to help replace `UNSAFE_componentWillReceiveProps` is `getDerivedStateFromProps`. However, the React documentation strongly advises against using `getDerivedStateFromProps` if it can be avoided, as it makes the whole component and whole app more complicated.

This is what I did to avoid using `getDerivedStateFromProps`:

In this component, `UNSAFE_componentWillReceiveProps` was used to update `state.isLoading` once a change in assignment is loaded. So I wrote a custom `async` function to do assignment updates and also update the update state as a side effect, once the await that stores the response resolves. 

@ragesoss @khyatisoneji please let me know what you think of this refactor ^ 

[React: Why You Probably Don't Need Derived State](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#when-to-use-derived-state)

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -36,6 +36,16 @@ const ArticleFinder = createReactClass({
     };
   },
 
+  componentDidMount() {
+    if (window.location.search.substring(1)) {
+      this.getParamsURL();
+    }
+    if (this.props.course_id && this.props.loadingAssignments) {
+      this.props.fetchAssignments(this.props.course_id);
+    }
+    return this.updateFields('home_wiki', this.props.course.home_wiki);
+  },
+
   componentWillUnmount() {
     return this.props.resetArticleFinder();
   },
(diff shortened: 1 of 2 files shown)


ragesoss:
It looks like this changes the behavior of the 'Add as available article' button when using the article finder for a specific course (eg, `localhost:3000/courses/test/Short_test_(test)/article_finder` ). When clicked, it goes to the disabled state, but when the request resolves, it goes back to the enabled state (and clicking it again will send the 'add assignment' post again, which will result in an error from the attempt to duplicate it). That doesn't happen with the pre-refactor version.

(I also noticed a bug already present, which is that already-available articles have the button enabled if you refresh the article finder page and do the query again.)

dnahol:
Hi @ragesoss thank you for catching this! I just got home from a daytrip- will look at this right away.

ragesoss:
Shashwat's PR worked through the remaining issues with ArticleFinder.

PR #3516, opened 2019-11-20
Fix Article Finder Row button toggle

### LINKED ISSUE #3515: Article Finder Row buttons not toggling
`localhost:3000/courses/test/Short_test_(test)/article_finder`

### What is happening?

Buttons are not toggling and they are remaining disabled once article assigned, until reload.

In `article_finder_row.jsx`:

1. `this.props.assigned` is always `undefined` regardless of whether an article is assigned or not. 

2. According to the existing (pre-refactor) code, buttons in article rows should not be disabled when assigned; they should be toggled from `Assign` to `Unassign`.

### To Reproduce

1. Go to `localhost:3000/courses/test/Short_test_(test)/article_finder`
2. Search for an term in the search bar
3. Assign an article using the button
4. Instead of toggling to 'Unassign', the button will become disabled
5. Reload the page. Now all buttons will be 'Assign' and enabled, but still none will be toggled.

### Expected behavior
Buttons in article_finder_row should be toggling from 'Assign' to 'Unassign' when clicked. They  should only be disabled while an article is loading, then enabled again when article is done loading.

### Screenshots
ArticleFinderToggle Issue

### System Information

 - OS: MacOS High Sierra 10.13.6 (17G7024)
 - Browser Chrome 
 - Version 78.0.3904.97 (Official Build) (64-bit)

### Additional context
Bug found by @ragesoss and @dnahol while working on #3478 React Refactor, PR #3506 Article Finder

dnahol:
## What this PR does
Fixes #3515 Article Finder Row buttons not toggling
Affects behavior in #3506 Article Finder: refactor UNSAFE react methods

## Screenshots
Before:
ArticleFinderToggle Issue

After:
ArticleFinderToggle PR

### `language` and `project` search fields:
Had to remove `language` and `project` search fields to get results from searching this.props.assignments array. This seems to produce the desired result, but let me know if you find otherwise. 

### `STUDENT_ROLE`:
Still need to test the fix in the `STUDENT_ROLE` case. @ragesoss @khyatisoneji How can I do that? Also in general how do you want `STUDENT_ROLE` article finder and assignments to behave differently from non `STUDENT_ROLE` ones?

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -279,8 +279,8 @@ const ArticleFinder = createReactClass({
       const elements = _.map(this.props.articles, (article, title) => {
         let assignment;
         if (this.props.course_id) {
-          if (this.props.current_user.isNonstudent) {
-            assignment = _.find(this.props.assignments, { article_title: title, user_id: null, language: this.props.home_wiki.language, project: this.props.home_wiki.project });
+          if (this.props.current_user.role !== STUDENT_ROLE) {
+            assignment = _.find(this.props.assignments, { article_title: title, user_id: null });
           } else if (this.props.current_user.role === STUDENT_ROLE) {
             assignment = _.find(this.props.assignments, { article_title: title, user_id: this.props.current_user.id, language: this.props.home_wiki.language, project: this.props.home_wiki.project });
           }


ragesoss:
Nice job pinpointing the problem.

If the user is in the student role, this should assign the article to that user. If it's a non-student course role — `isNonstudent` means admin, or instructor, or one of several other roles, as opposed to no role in the course at all — then choosing an article from Article Finder adds it to the set of 'Available Articles' for the course, which are `Assignment` records with the course that don't have a user attached.

I think there are edge cases where removing the language/project from the filter will result in buggy behavior. In particular, there is a mix of assignments from different wikis (such the article about the same person in several different languages) then this would incorrectly show the article in another language as already selected if one with the same title was already selected for a different language.

If I recall correctly, the system doesn't send the language/project explicitly for assignments that are on the course's home_wiki, only for ones that are on a different wiki.

PR #3520, opened 2019-11-22
Fixes Popover Inputbox Focus

### LINKED ISSUE #3501: When 'add/remove students' button is clicked, username input should get focus
On the 'Students' tab of a course page, admins and instructors have an 'Add/Remove Students' button that opens a popover form where usernames of new students can be added.

This input box should grab focus when the popover opens, so that a user can immediately begin typing a username.

`enroll_button.jsx` is the place to start.

Kartik1397:
Fixes #3501 
Input box grab focus when 'Add/Remove Students' button of the 'Student' tabs of a course page is clicked. 

## Screenshots
Before:
captured (7)

After:
captured (6)

### DIFF
--- a/app/assets/javascripts/components/students/enroll_button.jsx
+++ b/app/assets/javascripts/components/students/enroll_button.jsx
@@ -192,7 +192,19 @@ const EnrollButton = createReactClass({
     const buttonText = this.props.inline ? '+' : CourseUtils.i18n('enrollment', this.props.course.string_prefix);
 
     // Remove this check when we re-enable adding users by username
-    const button = <button className={buttonClass} onClick={this.props.open}>{buttonText}</button>;
+    const button = (
+      <button
+        className={buttonClass}
+        onClick={() => {
+          this.props.open();
+          setTimeout(() => {
+            this.refs.username.focus();
+          }, 125);
+        }}
+      >
+        {buttonText}
+      </button>
+    );
 
     return (
       <div className="pop__container" onClick={this.stop}>


ragesoss:
@bwreid is SetTimeout a good way to do this, or can you think of a more React-y way to grab focus when the popover is opened?

bwreid:
I tried a few things to improve this but was unable to get it to work. I think the issue may be [described here](https://stackoverflow.com/questions/35522220/react-ref-with-focus-doesnt-work-without-settimeout-my-example). With that said, I was unable to actively trigger the `.focus()` even after the popup has opened. Is anything else besides React managing this popup that you know of?

ragesoss:
@bwreid the thing managing the popup is the PopoverExpandable higher order component (the state of which is part of the redux `ui` state). The whole thing should be rendered already before and after clicking; the transition just flips the `is_open` prop which in turn sets the `open` CSS class on the `Popover` component that is part of the rendered result.

bwreid:
@ragesoss Right, I saw this chain of events. So, I tried the following:

- Moving the `.focus()` call to various life cycle methods (e.g. `componentDidMount`, `componentWillMount`, `componentDidReceiveProps`)
- Moving the `.focus()` call to right above the render

That last option doesn't seem great but I would expect it to work -- except it doesn't. If you want to pair on this a bit more, I'm happy to demonstrate what I tried.

ragesoss:
@Kartik1397 any ideas why the setTimeout is necessary and we can't get the focus any other way?

Kartik1397:
I think popover is not visible at time of calling `.focus()`

PR #3531, opened 2019-11-25
Remove new line when checking whether or not to post wiki edits

bwreid:
This just removes the `"\n"` at the end of the template when making edits for a user getting enrolled in a course.


ragesoss:
We still want to post the newline, as it avoids messing up pages where there is already content present (such as another template, or the first message on someone's talk page).

I think instead of removing it from the template, we want to remove it from what we're matching against in the `include?` check.

PR #3559, opened 2019-12-05
Change passcode detail to be a link - #2520

RyanHiller:
## What this PR does
This is a fix for Issue #2520 

## Screenshots
Before:
Before

After:
After

The link in the above example would lead to http://localhost:3000/courses/MSU_Denver/Example/enroll/rwsijmf
This does not include the copy-to-clipboard button like what was talked about in the issue comments.


RyanHiller:
Sorry about that, format of the URL should be good to go now.

ragesoss:
@RyanHiller thanks!

PR #3556, opened 2019-12-05
Pull hook

minhducsun2002:
## What this PR does
Adding a opt-in `post-merge` hook that refreshes all dependencies when changes to `package.json` or `Gemfile` get merged.

PR #3563, opened 2019-12-06
Pull hook

minhducsun2002:
Rewrite of #3556.


ragesoss:
I tested this out and it works nicely. Have a look before I merge it, @bwreid ?

bwreid:
Unfortunately `realpath` doesn't work for me on Mac.  :\  Can something similar be done with `pwd`? For example:
```
SCRIPT_ABSOLUTE_PATH=$(pwd -P $SCRIPT_PATH)
```

minhducsun2002:
`readlink` should be portable.

ragesoss:
I haven't had a chance to check it on Mac yet, but it already worked for Linux and this seems like it ought to work for Mac as well. I'll verify next week before merging it.

minhducsun2002:
Well I checked with a Hackintosh user and looks like `readlink` is available. It should work.

bwreid:
I'm sorry, I should've gotten to this on Friday.

`readlink` still doesn't perform correctly for me. The command is available but I get the following output:

```
$ ./.git-hooks/install-pull-update-deps
readlink: illegal option -- m
usage: readlink [-n] [file ...]
This will overwrite your current post-merge hook with /hooks/pull-update-deps.
cp: /../.git/hooks/post-merge: No such file or directory
```

.git-hooks/install-pull-update-deps
```diff
@@ -1,8 +1,37 @@
...
+            link=$t
+        fi
+    done
+    realpath="$PWD/$(basename "$link")"
+    cd "$ourpwd"
+    echo "$realpath"
+}
+
+# ===========
+
```
ragesoss: This seems like a lot of code to keep in the repo for such a narrow use case. I think a better approach would be to just document how to install the git hook manually by copying it to the correct location, without including a script to install it.

minhducsun2002:
Done.

ragesoss:
This cool with you @bwreid ?

bwreid:
Yep, this works. Thanks for the changes!

PR #3574, opened 2019-12-10
Added styles to article viewer

### LINKED ISSUE #3507: ArticleViewer should add appropriate styling for Wikidata
Wikidata items look terrible in ArticleViewer:

article viewer wikidata

We should add basic styles so that a Wikidata item looks similar in ArticleViewer to the way it looks on Wikidata (eg, https://www.wikidata.org/wiki/Q975 )

The place to start is probably `_article_viewer.styl`.

n3rsti:
### UI Before changes:
image

### UI After changes:
image

I did some changes to article viewer styles. It's not completed yet, but I wanted to open this PR to commit changes and you can let me know if something is wrong with UI.

### DIFF
--- a/app/assets/stylesheets/modules/_article_viewer.styl
+++ b/app/assets/stylesheets/modules/_article_viewer.styl
@@ -155,28 +155,54 @@
     vertical-align super
     font-size smaller
 
-.wikibase-statementgroupview-property-label
-  padding 30px 10px
-  width 100%
-  margin-right 70px
+.wikibase-statementgroupview-property-label 
+	padding 30px 10px
+	width 100%
+	margin-right 70px
 
-.wikibase-statementgroupview-property
-  background-color #eaecf0
+.wikibase-statementgroupview-property 
+	background-color #eaecf0
+	width 15em
 
-.wikibase-statementgroupview
-  background-color #fff
-  border 1px solid #c8ccd1
-  margin-bottom 15px
-  display flex
+.wikibase-statementgroupview 
+	background-color #fff
+	border 1px solid #c8ccd1
+	margin-bottom 15px
+	display flex
 
-.wb-normal
-  margin-left 15px
-  border-bottom 1px solid #eaecf0
-  width 100%
-  padding 15px 5px
+.wb-normal, .wb-preferred 
+	border-bottom 1px solid #eaecf0
+	width 100%
+	padding 15px 25px
 
-.wikibase-statementlistview
-  width -webkit-fill-available
+.wikibase-statementlistview 
+	width -webkit-fill-available
 
-.wikibase-statementview-mainsnak-container
-  margin-right 30px
+.wikibase-statementview-mainsnak-container 
+	margin-right 30px
+
+.wikibase-snakview 
+	display flex
+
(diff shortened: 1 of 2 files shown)


ragesoss:
This looks like a great start.

n3rsti:
## I added styles to references div. Now it looks like that
image

I wasn't sure if I can change html so I didn't add any "reference arrow down" with collapsing list of references. I think it looks pretty much like WikiData site.

I didn't find option to create my own article or use already existing from wikidata in local development project, but I tested it in chrome dev tools on website you linked in Issue #3507 and i think it works fine.

Please let me know if there is something to change.

n3rsti:
I pushed schema.db by accident, but I will delete this file now.

app/assets/stylesheets/modules/_article_viewer.styl
```diff
@@ -153,3 +154,55 @@
   sup
     vertical-align super
     font-size smaller
+
+.wikibase-statementgroupview-property-label
```
AmitJoki: Can the classes be a bit succinct while still conveying the meaning?

AmitJoki: Sorry. I thought you were adding these classes, whereas you're only styling the existing markup.

AmitJoki:
This looks neat!

n3rsti:
Is it working fine? :)

ragesoss:
This is definitely a lot better than the unstyled version we had before. It would look much better if the sitelinks section was moved to the right column like it is on Wikidata. Here's what it looks like now, at the end of the page:

sitelinks

n3rsti:
Ok, thanks for feedback. I will fix it.

n3rsti:
@ragesoss I can try to do that, but I see that article viewer window is smaller then whole wikidata page and main content (statements) will be tight and it can overflow on smaller devices than FHD. Wouldn't it be better if I keep sitelinks on bottom and style them to look like wikidata sitelinks?

ragesoss:
@n3rsti that's a good point. On Wikidata, it shifts the position depending on screen width, but at the width of ArticleViewer, it does show up underneath.

Still, it should style the sitelinks as on Wikidata.

n3rsti:
Ok. I tried to fix it like that. Sitelinks and Statements titles are not on the same height because info above the Statements is always different for every article and this info is in the main content div and I could do the height calculation in Javascript, but not in css (i think). If screen width is small, sitelinks are going under the Statements width flex-wrap. Is it fine?
image

ragesoss:
Yeah, that screenshot looks very good.

n3rsti:
Ok, I will fix overflow bugs and push changes.

n3rsti:
Should it be responsive for smaller devices like ip5s? I've already done rwd for width to about 700-800px depending on article, but rwd on smaller devices would need many breakpoints and styles.

ragesoss:
I think it's fine if it goes to 700-800px. There are plenty of bigger problems with smaller sizes.

n3rsti:
I added sitelinks to the right side.

On smaller devices (less than 1300px):
image

On bigger devices (more than 1300px):
image

PR #3579, opened 2019-12-11
[wip] Feature/instructor editors view

bwreid:
This is a refactor and feature add to enable a more robust instructor view for the students page. Currently a work in progress.


bwreid:
This PR is ready for review. The test that keeps failing does not fail locally, so I'm not sure what exactly the issue is.

ragesoss:
This looks great!

A few minor tweaks:
* "Select Students to assign and view article assignments" should be 'Select a student', so it doesn't imply that you can select more than one at once.
* The whole table row of each assignment in the Assignments view registers as clickable, even though only the links have relevant click behavior.
* In the drawers for recent revisions and for trainings, there is extra whitespace after the purple border. I think this is because of whatever hacky CSS workaround was there from when I first added the training completion to the drawers.

white space

bwreid:
@ragesoss Thanks for the review! These changes should be implemented.

ragesoss:
This looks fabulous. On the code side of things, the organization looks great. The only thing that occurs to me is that it might be nice to pull template names (like the preload bibliography template) out into explicit constants instead of parts of bigger url fragments... one constants file that includes the names of all the templates that get used by frontend code will surely come in handy at some point.

Looks like the border is still 1px off:
one pixel off

ragesoss:
@bwreid I'm going to go ahead and merge and deploy this. Whee!

Issue #2990, opened 2019-06-17
Incomplete state of Articles Edited list is not obvious, when there are over 500 articles


ragesoss:
A course with more than 500 articles will only load the first 500 initially, and loading more is done with the 'See More' button at the end of the list. This is confusing if you're expecting to see a certain article in the top several (sorted by content added, or resorted some other way).

In cases where only a partial list of articles has been loaded, the count of loaded articles versus the total, along with the button to load more, should appear at the top of the list.

thefifthisa:
Hey @ragesoss! Can I give this a try?

ragesoss:
@thefifthisa go for it! Let us know if you have any questions.

prionkor:
Any urls for such long courses? Helps checking the issue quickly.

ragesoss:
@prionkor here's an example that has more than 500: https://outreachdashboard.wmflabs.org/courses/NeuroMat/Wikidata_Lab_XI_-_Estruturando_projetos/articles/edited

prionkor:
Few observations: 

1. If "load more" button is moved to the top then when people goes to bottom they will have to scroll up again to load again.
2. If "load more" at the top then it will be problematic if articles started to load at the bottom. Users wouldn't know.

**Suggestion**

1. Pagination: Paginate after 10 (or more) articles in the list. Load on pagination click.
2. Scroll to new results onload: If pagination is not a option, users is needed to scroll to the result after new items are loaded. 
3. Need a loading state for the "See more" button.

Let me know what you think.

ragesoss:
I think a loading state for the 'see more' button, combined with duplicating the button on top and bottom of the list, would be a relatively simple way to start.

bharathjoshi:
Hey @ragesoss ! I have been on this issue for quite a time, found a way to duplicate the button but didn't get the essence of "loading state" for the see more button.
Another question I have is, how can I get the number of total articles ( in the context of adding the count of loaded vs count of total articles). The code has been using fetchArticles where we are giving the id and then the number. Not really being able to understand how the API works. Any insights?

ragesoss:
@bharathjoshi we currently don't have anything that shows the total number, except that in most cases it will match the `articles_edited` value from the `course` state.

Currently, we determine whether or not to show the button via the `limitReached` value from the articles reducer, based on whether the number of articles received is less than the limit of the last fetchArticles query.

[...]

PR #3591, opened 2019-12-18
Add showing xx of xx  in articles edited list

abhishekbvs:
## What this PR does
Fixes issue #2990. Added showing x of y where x is the no of articles loaded at the instant and y is the total no of edited articles. Shifted See More button from the bottom to top of the table.

## Screenshots
Before:
DeepinScreenshot_select-area_20191218210148

After:
DeepinScreenshot_select-area_20191218205102

### DIFF
--- a/app/assets/javascripts/components/articles/article_list.jsx
+++ b/app/assets/javascripts/components/articles/article_list.jsx
@@ -215,14 +215,20 @@ const ArticleList = createReactClass({
       </div>
     );
 
-    let showMoreButton;
+    let showMore;
     if (!this.props.limitReached) {
-      showMoreButton = <div><button className="button ghost stacked right" onClick={this.showMore}>{I18n.t('articles.see_more')}</button></div>;
+      showMore = (
+        <div className="see-more">
+          <button className="button ghost" onClick={this.showMore}>{I18n.t('articles.see_more')}</button>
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/articles/article_list.jsx
```diff
@@ -215,14 +215,20 @@ const ArticleList = createReactClass({
       </div>
     );
 
-    let showMoreButton;
+    let showMore;
     if (!this.props.limitReached) {
-      showMoreButton = <div><button className="button ghost stacked right" onClick={this.showMore}>{I18n.t('articles.see_more')}</button></div>;
+      showMore = (
```
ragesoss: I suggest a more explicit name like `showMoreSection` so that this won't get confused with the function (this.seeMore) that has the same name.

config/locales/en.yml
```diff
@@ -180,6 +180,7 @@ en:
       ?: Unrated — an article that hasn't yet been rated on Wikipedia's quality scale
       does_not_exist: This article does not exist on Wikipedia
     see_more: See more
+    showing_of: 'Showing %{count} of %{total}'
```
ragesoss: I suggest a more explanatory name for this interface message, like `articles_shown`. 'showing_of' is likely to cause confusion.

PR #3592, opened 2019-12-18
Adapt CourseStats for Wikidata and other non-Wikipedia courses

ragesoss:
This is the start of making CourseStats more flexible based on the course stats present and the home wiki. It hides 'bytes added' for Wikidata courses, and updates the labels for Wikidata and for other non-Wikipedia courses, and also hides the uploads count unless there is at least one upload.

## Screenshots
For a Wikidata course...

Before:
old course stats

After:
Screenshot from 2019-12-18 10-08-14

For an empty course:

Before:
empty course before

After:
empty course

### DIFF
--- a/app/assets/javascripts/components/overview/course_stats.jsx
+++ b/app/assets/javascripts/components/overview/course_stats.jsx
@@ -47,9 +47,7 @@ const CourseStats = ({ course }) => {
         <small>{I18n.t('metrics.word_count')}</small>
       </div>
     );
-  } else if (isWikidata) {
-    // noop
-  } else {
+  } else if (!isWikidata) {
     contentCount = (
       <div className="stat-display__stat" id="bytes-added">
         <div className={valueClass('word_count')}>{course.character_sum_human}</div>
--- a/app/assets/javascripts/components/overview/course_stats.jsx
+++ b/app/assets/javascripts/components/overview/course_stats.jsx
@@ -29,6 +29,16 @@ const CourseStats = ({ course }) => {
   let trainedTooltip;
   let uploadCount;
 
+  let articlesCreated;
+  if (course.created_count !== '0') {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/overview/course_stats.jsx
```diff
@@ -3,14 +3,41 @@ import PropTypes from 'prop-types';
...
+    editedLabel = I18n.t('metrics.articles_edited_generic');
+  }
+
+  let pageviews;
   let infoImg;
   let trainedTooltip;
+  let uploadCount;
+
+  let articlesCreated;
+  if (course.created_count !== '0') {
```
bwreid: Hm. Would this get the same idea across? Just a suggestion.
```js
const count = Number(course.created_count)
if (count > 0) { // or, (count !== 0)
// ...
}
```

app/assets/javascripts/components/overview/course_stats.jsx
```diff
@@ -20,6 +47,8 @@ const CourseStats = ({ course }) => {
         <small>{I18n.t('metrics.word_count')}</small>
       </div>
     );
+  } else if (isWikidata) {
```
bwreid: This seems like an odd move. Can we rephrase this into an affirmative `if/else` statement? For example:
```js
else if (!isWikidata) {
```

app/assets/javascripts/components/overview/course_stats.jsx
ragesoss: That works mostly, except sometimes created_count could be something like '1.2k'.

bwreid: Ah, that makes sense. Ok, well, enh. Thanks for clarifying.

PR #3594, opened 2019-12-18
Bug/progress tracker ui fixes

bwreid:
## What this PR does

There were two small bugs I found while working on the instructor view:
1. The navigation bar styling had changed so that the text of the first step appeared on two lines
1. When marking a review as completing, the last step viewable was skipped

This updates the styling to fix the first problem and adds an additional status to the AssignmentPipeline to solve the second problem.

After merging and deploying this change, we may need to update existing review Assignments to have the correct status. We can do so with something like this:
```ruby
all_assignments = Assignment.where(role: Assignment::Roles::REVIEWING_ROLE)
assignments = all_assignments.reject { |a| a.flags.empty? }
assignments.each do |assignment|
  status = AssignmentPipeline::ReviewStatuses::READING_ARTICLE
  assignment.assignment_pipeline.update_status(status)
end
```

## Screenshots
**Realigned Navigation Bar**
image

**Last Step on Reviews now Accessible**
image


ragesoss:
In production, we don't yet have any assignments that have any flags, so no update should be required there.

PR #3620, opened 2019-12-28
made changes to troubleshooting.md

Manasa2850:
## What this PR does
< Changes made to troubleshooting.md. After trying out several times and reading online, I have found out that different cli and local versions of gulp do not affect the dev environment. I have removed the statement which says they must be the same in troubleshooting.md>
GulpVersions


ragesoss:
Thanks! There are two extra changes that should not be included in the PR:

The schema.rb change appears to be a correct change, but should be done in a separate commit. I guess that line was added on accident from an intermediate version of a PR by @AmitJoki .

The package-lock file is not necessary; we use yarn.lock instead.

Manasa2850:
I'll make the necessary changes and resend the pr

PR #3633, opened 2020-01-01
changed troubleshooting.md

Manasa2850:
## What this PR does
<< Changes made to troubleshooting.md. After trying out several times and reading online, I have found out that different cli and local versions of gulp do not affect the dev environment. I have removed the statement which says they must be the same in troubleshooting.md>
GulpVersions
>


Manasa2850:
@ragesoss could you let me know what the error is? I just changed troubleshooting.md

Issue #2099, opened 2018-10-12
Long page titles break the layout of UploadViewer


ragesoss:
Wrapping the page title should fix it.
![long article names in upload viewer](https://user-images.githubusercontent.com/848483/46887580-de5a6780-ce12-11e8-9fc2-e04ad2d7535d.png)

DK308:
Hello is this issue available?

ragesoss:
@DK308 yes, if you'd like to work on it, please go ahead.

DK308:
Ok, I m going to solve this issue :)

ragesoss:
@snehasi yes, please do! Sorry for the late reply; I've been on vacation and neglecting my inbox.

utkarsh-raj:
Are you still working on the issue @snehasi ? If not, I would like to take up the issue @ragesoss ?

twisted-sres:
Hey, I am an outreachy applicant, can I take this up?

helenatxu:
Hello, I am an applicant for the Outreachy internship, can I take this issue for my first contribution?

mariahajmal:
@ragesoss  Hi, Is this fixed? If not, can you please assign it to me

kevinejiro:
Hi @ragesoss   This issue doesn't seem to be taken, can I work on it?

ragesoss:
I believe this is partially fixed, but there there are some layout problems in some cases still. Feel free to explore it and make any improvements you can.

Manasa2850:
@ragesoss can I work on this?

ragesoss:
Yes, go for it.

Manasa2850:
@ragesoss where exactly is this upload viewer? I created a campaign but I'm not quite sure how to proceed.

ragesoss:
@Manasa2850 this can be seen on the Uploads tab of a course page (if the course has any uploads).

Try creating some example data (see the "Populate example data" section of setup.md) and then visiting http://localhost:3000/courses/Uffizi/WDG_-_AF_2018_Florence/uploads

shipcy:
Can I take this issue?

ragesoss:
@shipcy this one is already done.

PR #3636, opened 2020-01-02
changed layout of uploads viewer

Manasa2850:
## What this PR does
< fixed issue #2099  >

## Screenshots
Before:
< 
Before
 >

After:
<
After
>

### DIFF
--- a/app/assets/javascripts/components/uploads/upload_viewer.jsx
+++ b/app/assets/javascripts/components/uploads/upload_viewer.jsx
@@ -74,8 +74,8 @@ const UploadViewer = createReactClass({
       usageTableElements = globalUsage.map((usage, index) => {
           return (
             <tr key={usage.url}>
-              <td>{usage.wiki}</td>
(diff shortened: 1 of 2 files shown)


ragesoss:
@Manasa2850 This is a solid improvement. I can merge it now, but I noticed one thing that could be further tweaked. The UploadViewer's table of file usage renders differently depending on whether the Uploads tab is in 'list view' or not.

In list view, the table cells have borders (which is the intended design) while in the other views they don't. This is because UploadViewer component itself is inside the gallery table which changes layouts via CSS.

List view:
list

Non-list:
non-list

ragesoss:
@Manasa2850 let me know if you'd like to work on that table borders thing as part of this PR. If not, I'll merge it and open a new issue.

Manasa2850:
@ragesoss I'm currently trying to solve the mobile layout errors. So you can create a new issue. I'll work on it once I finish resolving the mobile layout errors.

PR #3644, opened 2020-01-07
changed banner styling for mobile

Manasa2850:
## What this PR does
< fixes issue #927  >

## Screenshots
Before:
< 
Before
 >

After:
< 
After>

< Please ignore the previous pr as it was wrong. This should work fine. Let me know if I need to make any changes. I have tested it in iPhone 6/7/8 Plus. >

### DIFF
--- a/app/assets/stylesheets/modules/_modules.styl
+++ b/app/assets/stylesheets/modules/_modules.styl
@@ -98,4 +98,4 @@
     margin-top -30px
     +below(desktop)
       margin-top: 0px
-
+      margin-bottom: 350px


ragesoss:
I don't think this is a good approach, as it adds a whole lot of empty space in the case where there are no banners. Ideally, this would be done in a way that only shift things down as much as the banners need.

Manasa2850:
Yup agreed. I'll try to make the changes. What is the best way to go about this? Should I make changes for each thing in the navbar like home, timeline, students etc or is there a generalized method to solve this?

Manasa2850:
@ragesoss yup agreed. I'll try to make those changes.

Issue #3648, opened 2020-01-08
disordered timeline page in mobile phones


Manasa2850:
This is how the timeline page for each course looks in mobile phones.
![Before](https://user-images.githubusercontent.com/43354059/71980512-3e260500-3246-11ea-917c-dc932222fc3c.jpeg)

kunalthedev:
@ragesoss  To solve this, should i just add some relevant margin's in CSS or something else?

ragesoss:
I think the ideal solution to this would be based on using the media queries system that we already use to provide mobile-friendly layouts. (Look for `+above(` and `+below(` in the .styl files.) The desktop layout should remain the same, but whatever adjustments need to be make to improve the mobile layout should apply at lower widths. It will be easiest to understand and modify later if, where possible, it uses the same CSS properties that are already being used to style the classes of a given component, just with different values for the mobile version.

kunalthedev:
@ragesoss this can be closed now

PR #3649, opened 2020-01-09
changed banner styling for the timeline page

Manasa2850:
## What this PR does
< solves issue #3648  >

## Screenshots
Before:
< 
Before
 >

After:
< 
WhatsApp Image 2020-01-10 at 12 52 01 AM
>

### DIFF
--- a/app/assets/javascripts/components/timeline/timeline.jsx
+++ b/app/assets/javascripts/components/timeline/timeline.jsx
@@ -64,12 +64,12 @@ const Timeline = createReactClass({
     return week.blocks;
   },
 
-  submit(e) {
+  /* submit(e) {
     e.preventDefault();
     if (!confirm(I18n.t('courses.warn_mirrored'))) { return; }
     this.props.updateCourse({ submitted: true });
     return this.props.persistCourse(this.props.course.slug);
-  },
+  }, */
 
   usingCustomTitles() {
     return this.props.weeks.some(week => week.title);
--- a/app/assets/javascripts/components/timeline/timeline.jsx
+++ b/app/assets/javascripts/components/timeline/timeline.jsx
@@ -64,13 +64,6 @@ const Timeline = createReactClass({
     return week.blocks;
   },
 
-  /* submit(e) {
-    e.preventDefault();
-    if (!confirm(I18n.t('courses.warn_mirrored'))) { return; }
-    this.props.updateCourse({ submitted: true });
-    return this.props.persistCourse(this.props.course.slug);
-  }, */
-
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -64,6 +64,13 @@ const Timeline = createReactClass({
     return week.blocks;
   },
 
+  submit(e) {
```
ragesoss: I don't understand what's going on with all the changes to this file. It doesn't look like this `submit` function is used anywhere.

Can you explain more fully what the strategy of this change is, and clean up any changes to this file that aren't necessary?

Manasa2850: @ragesoss sorry my bad. There's some unwanted code in the file. I'll remove that and make a commit

Manasa2850: And about the other changes, basically what I've done is I check whether there is a banner or not and do the styling accordingly. I got the if conditions from course_alerts.jsx

ragesoss: Thanks for the explanation! That helps me understand it.

Manasa2850: Is what I have done fine?

Manasa2850:
@ragesoss I wasn't able to solve it just with CSS. 
So ok, I'll try creating a new state element and work with it!

Manasa2850:
@ragesoss this is the best I could do. I tried this method but it didn't work 
1) tried using **'document.getElementsByclassName('notification').offsetHeight; '** but it returned undefined for the height

ragesoss:
Thanks for exploring this @Manasa2850. I'm pretty sure that something simpler and more general is possible, so I'm not going to merge a fix that introduces so much code to work around one specific situation. It looks like the current state of it is duplicating the logic of to show that specific banner, but there could be other large banners being rendered from the Notifications component that cause the same problems.

The `getElementsByClassName` function will return an array of elements with that class name; `offsetHeight` is undefined for the array, but will work for an individual element. It looks like the notification you're testing against is actually not from the Notifications component but from the CourseAlerts component. I think the place to start would be something like `document.getElementsByClassName('course-alerts')[0].offsetHeight`

ragesoss:
(Or add an ID to that CourseAlerts element, since there should only be one.)

ragesoss:
I'm going to close this for now. Please feel free to reopen if you take it up again.

PR #3664, opened 2020-01-13
Fix Training Module Selection of a Block

### LINKED ISSUE #3653: Editing the training modules of a block is buggy when multiple types are present
If there is more than one type of module present (training / exercise / discussion) then in edit mode, there are two inputs for adding / removing modules.

1. Both have the same label (unlike in view mode).
2. All the modules are available for each input, not just the corresponding type.
3. Adding or removing modules from the second input and saving just deletes all the modules that are not in the first input.
4. Some modules (ones with individual vs group) have the same name and there's no way to tell which you've selected.

Kartik1397:
Fixes #3653

## What this PR does
In edit mode all training modules are displayed in one input. 

## Screenshots
before:
Screenshot from 2020-01-15 16-44-43
after:
Screenshot from 2020-01-15 16-42-50


ragesoss:
@Kartik1397 please include before/after screenshots.

ragesoss:
This works nicely. I think the 'view' mode portion should be refactored into a separate component, as it's a lot of conditional code in the `Block` component, and that logic is specific to rendering the training modules when they are not being edited.

Kartik1397:
Ok, I will create seperate component `TrainingModulesViewMode` from `Block` component.

app/assets/javascripts/components/timeline/TrainingModules/TrainingModulesViewMode.jsx
```diff
@@ -0,0 +1,88 @@
...
+    displayName: 'TrainingModulesViewMode',
+
+    propTypes: {
+        block: PropTypes.object,
+        all_training_modules: PropTypes.array,
+        trainingLibrarySlug: PropTypes.string.isRequired,
+        editable: PropTypes.bool,
+      },
+
+      passedUpdateBlock(selectedIds) {
```
ragesoss: Is this update code necessary in view mode? If so, it can probably be passed from the parent component instead of duplicated here. But I suspect it can be removed altogether, since the view-mode components shouldn't need to fire any update actions.

Kartik1397: You are right, this is unnecessary.

ragesoss:
Nice job @Kartik1397! Thanks!

Issue #1318, opened 2017-06-30
Disordered Article page when in mobile


lopezabegaillyn:
The navbar is blocking some parts of the div of the articles.

lopezabegaillyn:
<img width="660" alt="screen shot 2017-06-30 at 10 04 01 am" src="https://user-images.githubusercontent.com/29187834/27718154-8c6d3f9a-5d7c-11e7-9e0b-fdf028edbfb6.png">

ragesoss:
Thanks! @sejalkhatri is working on a mobile layout cleanup soon.

Vinit-Dantkale:
Hey @ragesoss , is this issue solved or can I work on it?

ragesoss:
@Vinit-Dantkale yes, go for it.

Vinit-Dantkale:
Hey, just one small check, Im seeing this UI on localhost for art+feminism course. It looks bit different from image above. Also,I can't find **Programming** course, I only have **art+feminism** and **Default campaign**. Is the below image same page as you are looking
![image](https://user-images.githubusercontent.com/32016324/71303136-70c4a680-23da-11ea-8b60-0319b0de6215.png)

shipcy:
Can I work on it?

ragesoss:
@shipcy you can work on any issue that's open, just please focus on one at a time, leave a message to say which one you're working on, and follow up if you stop work on it.

Vinit-Dantkale:
Hey, I'm so sorry for responding late.The status so far, I tried fixing some distortion in **uploads** page. 
![image](https://user-images.githubusercontent.com/32016324/72354115-2d283880-370b-11ea-8782-4d1a22c9cc88.png)

ragesoss:
@Vinit-Dantkale that looks like a nice improvment. Do you have an open PR? This looks like it might overlap with #3661

Vinit-Dantkale:
Oh @ragesoss , I didn't realize that. Should I open a PR? will follow your word.

ragesoss:
@Vinit-Dantkale please do.

Vinit-Dantkale:
> @Vinit-Dantkale please do.
Hey @ragesoss, I opened a PR #3670

Vinit-Dantkale:
Hi, I attempted to fix some distortion for articles section for tablets and below
[...]

PR #3670, opened 2020-01-15
Fixing view-buttons distortion

Vinit-Dantkale:
## What this PR does
This PR attempts to fix part of #1318. The uploads page had **view-buttons** div distorted for smaller screens. Added some **media queries** to prevent distortion.

## Screenshots
Before(left) / After(Right)
image

As this PR is part of bigger issue, next would like to fix distortion of **articles** page.

### DIFF
--- a/app/assets/stylesheets/modules/_uploads.styl
+++ b/app/assets/stylesheets/modules/_uploads.styl
@@ -90,16 +90,27 @@
     .info
       text-align left
 
-.view-buttons
-  position absolute
-  top 20px
-  right 80px
+@media (min-width: 721px)
+  .view-buttons
+    position absolute
+    top 20px
+    right 80px
 
-  button
-    padding 0
-    width 40px
-    height 40px
-    border-radius 50%
+    button
+      padding 0
+      width 40px
+      height 40px
+      border-radius 50%
+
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/modules/_uploads.styl
```diff
@@ -90,16 +90,27 @@
...
-  position absolute
-  top 20px
-  right 80px
-
-  button
-    padding 0
-    width 40px
-    height 40px
-    border-radius 50%
+@media (min-width: 721px)
```
ragesoss: Is this particular breakpoint essential? We have three other breakpoints that get used elsewhere — 400, 920, and 1200 — that are defined in `_variables.styl` and get used in many stylesheets (eg, see the `+above(desktop)` and `+below(desktop)` parts in `home.styl`. If we can use the same named breakpoints here for consisency, that would be ideal.

Vinit-Dantkale: Actually, I was checking for different devices and small width of browser tab, though `+below(desktop)` looked better choice and as per format when I checked other files. I made a commit with less repetition and no regression in desktop view.

PR #3661, opened 2020-01-13
changed uploads page for mobile phones

Manasa2850:
## What this PR does
< fixes #3660 >

## Screenshots
Before: 
before

After(in mobile): 
WhatsApp Image 2020-01-15 at 7 15 58 PM(1)

After(in desktop):
WhatsApp Image 2020-01-15 at 7 15 58 PM

 
## Open questions and concerns
< I have selected max-width as 800px so that it works fine even in devices like iPad and Kindle Fire.>

PR #3678, opened 2020-01-16
changed styling of uploads page

### LINKED ISSUE #3660: disordered uploads page in mobile phones
WhatsApp Image 2020-01-13 at 11 24 42 PM

Manasa2850:
## What this PR does 
< Fixes #3660 >

## Screenshots
Same as the ones in https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/3661

### DIFF
--- a/app/assets/stylesheets/modules/_selects.styl
+++ b/app/assets/stylesheets/modules/_selects.styl
@@ -40,6 +40,7 @@
     background url('../images/sort.svg') no-repeat 48% white
     -webkit-appearance none
     text-indent -200px
+    margin-top -10px
 
 .select-with-button
   width 400px
--- a/app/assets/stylesheets/modules/_selects.styl
+++ b/app/assets/stylesheets/modules/_selects.styl
@@ -32,23 +32,26 @@
   right 100px
   top 20px
 
-.sort-select
-  width 60px
-  &.inset
-    right 191px
-  select.sorts
-    background url('../images/sort.svg') no-repeat 48% white
-    -webkit-appearance none
-    text-indent -200px
++above(desktop)
+ .sort-select
+   width 60px
+   &.inset
+     right 191px
+   select.sorts
+     background url('../images/sort.svg') no-repeat 48% white
+     -webkit-appearance none
+     text-indent -200px
 
-.sort-select1
-  width 60px
-  &.inset
-    right 191px
-  select.sorts
-    background url('../images/sort.svg') no-repeat 48% white
-    -webkit-appearance none
-    text-indent -200px
++below(desktop)
+ .sort-select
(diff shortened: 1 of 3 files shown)


Manasa2850:
@ragesoss this is fine right?

ragesoss:
@Manasa2850 yes, thanks! I'll probably be able to test it out and either merge or give feedback next week.

Manasa2850:
@ragesoss sure! 
I haven't made any new changes. It's just the ones I had done previously.

Manasa2850:
@ragesoss I'm currently resolving pr 3649. Let me know if I need to make any changes to this as well.

app/assets/stylesheets/modules/_selects.styl
```diff
@@ -40,6 +40,7 @@
     background url('../images/sort.svg') no-repeat 48% white
     -webkit-appearance none
     text-indent -200px
+    margin-top -10px
```
ragesoss: This is going to affect every sort-select, not just the one in the screenshot.

I think this adjustment either needs to be done in a way that just affects that uploads page, or in a way that address the same issue for every view that includes a sort-select without any code specific to uploads.

Manasa2850: @ragesoss can I create a new className element for just the uploads page and make styling changes in that?

ragesoss: No, that will lead to more styling trouble later one, if/when we try to adjust the styles of those selects and it leaves just one of them unchanged. Doing it in a way that would be beneficial for all the sort-selects is much preferable to a fix that just deals with this one.

Manasa2850: I had a look at most of the sort-select buttons and it's only in the mobile view that there are layout issues. So I'll add **margin-top -10px** just to the mobile version.
Is that fine @ragesoss ?

Manasa2850: Applying that code will not create any disorientation in other pages that have sort-select

ragesoss: Okay. Can you include screenshots of before / after for both the uploads page and one of the other pages with a sort-select?

Manasa2850:
Desktop (before)
desktop(before)

Desktop (after)
desktop(after)

Mobile (before)
mobile(before)

Mobile (after)
mobile(after)

Other page with a sort-select (before)
otherpage(before)

Other page with a sort-select (after)
other page(after)

Manasa2850:
@ragesoss do I need to make any changes to this? If not, I'll start working on pr #3375

ragesoss:
This looks good.

PR #3699, opened 2020-01-23
Makes a popover in a modal responsive

Manasa2850:
## What this PR does
< vertical layout is present at all times >

Credits for initial code : @codingmamakaz

## Screenshots
Before:
Before

After:
After

### DIFF
--- a/app/assets/stylesheets/modules/_course.styl
+++ b/app/assets/stylesheets/modules/_course.styl
@@ -172,6 +172,12 @@ rating(color)
   .details-form
     display flex
 
+.pop__container
+  form
+    display flex
+    align-items center
+    justify-content center
+
 @media only screen and (max-width: 920px)
   .course_main,
   .campaign_main
@@ -211,3 +217,12 @@ rating(color)
 
 .multi-wiki input[readonly]
   display: none
+
+.modal-course-details
+    .pop__container
+        .pop.open
+          form
+            flex-direction  column
+            input
+              width 200px
+              margin-bottom 10px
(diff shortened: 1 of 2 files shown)


ragesoss:
Hmm... I think this will need to be scoped specifically to that one popover.

It causes a regression in others, like the AssignButton one, where the `en.wikipedia.org (change)` part gets shifted to the same line as the input, instead of below it like it should be.

regression

ragesoss:
Also, please add credit for the person whose work you're starting from in the commit message.

Manasa2850:
@ragesoss sorry I'll add the credits immediately.

Manasa2850:
@ragesoss sure, I'll make changes to just that one popover.

ragesoss:
@Manasa2850 are you still working on this?

Manasa2850:
@ragesoss yes I'm working on this. Sorry I was a little busy because of my tests.

Manasa2850:
I will try to resolve this by tomorrow night.

Manasa2850:
@ragesoss I've created another class 'modal-course-details1' which can be called whenever there is an issue with the popover. Hope this is fine

Manasa2850:
@ragesoss I've solved it specifically for that one popover. Hope it's fine

app/assets/stylesheets/modules/_course.styl
```diff
@@ -210,4 +210,17 @@ rating(color)
...
\ No newline at end of file
+  display: none
+
+.modal-course-details
+    .pop__container
+        form
+          display flex
+          align-items center
+          justify-content center
+        .pop.open
```
ragesoss: Why is this styling only in the `open` mode?

It looks like this is causing a problem with the transition animation: the modal shifts out of column format for a brief flash as the popover closes.

Manasa2850:
@ragesoss is it fine now?

ragesoss:
Looks good to me.

PR #3703, opened 2020-01-27
Replaced jquery to vanilla js in nav.jsx

### LINKED ISSUE #3675: replace jquery with vanilla JS in `nav.jsx`
The navbar component, `nav.jsx`, uses global jquery (`$`) to extract data from the hmtl and to find the window dimensions. These things can and should be done with vanilla JavaScript instead, as part of the gradual effort to remove the need for jquery in the codebase.

pranaykgupta:
##  What this PR does
Closes #3675 
Used vanilla js instead of jquery to extract data from the HTML and to find the window dimensions in the navbar component, nav.jsx;

## Screenshots
Before:
N/A

After:
N/A

### DIFF
--- a/app/assets/javascripts/components/nav/nav.jsx
+++ b/app/assets/javascripts/components/nav/nav.jsx
@@ -11,22 +11,23 @@ const Nav = createReactClass({
   displayName: 'Nav',
 
   getInitialState() {
-    const rootUrl = $('#nav_root').data('rooturl');
-    const logoPath = $('#nav_root').data('logopath');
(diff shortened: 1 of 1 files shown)


pranaykgupta:
can anyone help me with the CI checks

ragesoss:
@pranaykgupta it looks like the changes in your PR are resulting in some different routes; the failures mostly seem to be routing errors on the Rails side of things, meaning probably that requests are going to different endpoints than they were before.

PR #3709, opened 2020-01-29
Add aria-label for article viewer open/close buttons

### LINKED ISSUE #2490: ArticleViewer open/close buttons unlabeled
The ArticleViewer's open and close buttons are currently not accessible in that they have no labels. When used by a screenreader, it is difficult to understand what each button is.

We need to add some kind of label to the ArticleViewer's open and close buttons. I believe the Uploads tab on the course page has some examples of visual buttons that contain accessible text.

ihsavru:
Fixes #2490

Added `aria-label` attribute to buttons which have icons as labels.
Added `aria-describedby` attribute to button with text labels.


ragesoss:
@ihsavru lovely to see you in the pulls!

LGTM!

Issue #3715, opened 2020-01-30
AssignButton popover goes off the screen when title is too long


ragesoss:
For example:

![wikiedu_offscreen_assignment](https://user-images.githubusercontent.com/848483/73482531-c33aa080-4352-11ea-85fe-be1d483cf6f8.png)

Perhaps wrapping the title to multiple lines will fix it, or maybe some other CSS fix is the right solution.

mattwarren10:
It looks like this issue is already resolved. Taking a look at these images, we find that the popover is accessed from a different button and in a different location.

Here is the old one:
<img width="1142" alt="Screen Shot 2020-02-01 at 9 01 40 PM" src="https://user-images.githubusercontent.com/13757305/74886626-a6d4c700-5346-11ea-9733-7956e0d63852.png">

Here is the new one:
<img width="1131" alt="Screen Shot 2020-02-19 at 6 34 25 PM" src="https://user-images.githubusercontent.com/13757305/74886642-b3591f80-5346-11ea-8ef6-bbd6355133a8.png">

Let me know if there was a pr that fixed this issue.

ragesoss:
Yes, you're right. The issue became irrelevant because of the major layout change from https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/3579

Issue #2012, opened 2018-09-25
Re-rendering of editors table in chrome


psinghal20:
The Editors table for any course seems to re-render again and again when scrolled to bottom in chrome. This behavior was not observed in Firefox. The attached GIF shows the mentioned behavior for both browsers. It was most probably introduced when the sticky headers were introduced.

ragesoss:
@psinghal20 this didn't come through as a GIF, just a single frame. I haven't been able to replicate the issue described. What version of Chrome did you see this in?

ragesoss:
I did some work that reducers unnecessary rendering of the students components, but I don't know if that was related to the issue you observed or not. Anyhow, if you poke at this again, be sure to pull in the latest first.

psinghal20:
@ragesoss Sorry for that! I thought I used a GIF. The issue still persists, I am attaching a video link here as it is much more visible in a video. https://drive.google.com/open?id=1OR6ct5-BDS1geJLd2DeqONp9OqFiz9hQ

PR #3719, opened 2020-02-02
Fix AssignButton popover to word-wrap text at bottom-#3715

mattwarren10:
## What this PR does
Add simple word-wrap that breaks at each word to the assign button popover: issue #3715

## Screenshots
Before:
assign-button-popover-#3715

After:
Screen Shot 2020-02-01 at 9 01 40 PM


bwreid:
@mattwarren10 This looks great to me! Very small thing, but can you change the added CSS to be in the same format as the rest of the file? I.e. no `:` or `;`.

Thank you!

mattwarren10:
@bwreid check it again

bwreid:
Taking a look at this again, this now leads to a different problem in that any article with a space will be shortened so that the box will appear very small.
image
This isn't too terrible, but ideally this box would:
* Have a standard width to show most articles
* In the event that an article name goes over, it carries to the next line

I believe this could be achieved by setting a `width` and `max-width` for one of the popover elements. Don't spend too long on this! If you get stuck, let me know and I'll take a crack at it.

Thanks for your hard work!

bwreid:
@mattwarren10 Hi Matt -- looks like you got some odd stuff merged in. I roll back your changes to before the merge and then remake just _your_ changes. If that's proving tricky, just let me know.  :)

ragesoss:
As you noted, the main issue got solved along the way with the layout change to the Students tab.

There are still some edge cases where the popover listing the assigned articles can go offscreen, but they seem more related to #2012 than anything else.

PR #3733, opened 2020-02-09
Added shortened footnote template to references added

### LINKED ISSUE #3276: Include shortened footnotes in references added counts
The 'references added' feature uses ORES feature data to compare the number of references before and after a given edit. Specifically, it uses `feature.wikitext.revision.ref_tags`. This excludes refs that use the `{{sfn}}` (shortened footnote) template, which will undercount some articles or count some edits as removing refs when they are changing the ref format.

ORES now includes the shortened footnote count (`feature.enwiki.revision.shortened_footnote_templates`), so we should add that in to our calculations.

khyatisoneji:
Fixes #3276

### DIFF
--- a/app/models/wiki_content/revision.rb
+++ b/app/models/wiki_content/revision.rb
@@ -73,14 +73,16 @@ class Revision < ApplicationRecord
   WIKI_SHORTENED_REF_TAGS = 'feature.enwiki.revision.shortened_footnote_templates'
 
   def ref_tags
-    (features[WIKITEXT_REF_TAGS] || features[WIKIDATA_REFERENCES]) +
-      features[WIKI_SHORTENED_REF_TAGS] || 0
+    return nil if features.empty?
+    (features[WIKITEXT_REF_TAGS] || features[WIKIDATA_REFERENCES] || 0) +
+      (features[WIKI_SHORTENED_REF_TAGS] || 0)
   end
 
   def ref_tags_previous
     return 0 if new_article
-    (features_previous[WIKITEXT_REF_TAGS] || features_previous[WIKIDATA_REFERENCES]) +
-      features_previous[WIKI_SHORTENED_REF_TAGS] || 0
+    return nil if features_previous.empty?
+    (features_previous[WIKITEXT_REF_TAGS] || features_previous[WIKIDATA_REFERENCES] || 0) +
+      (features_previous[WIKI_SHORTENED_REF_TAGS] || 0)
   end
 
   def references_added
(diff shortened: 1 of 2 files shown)


khyatisoneji:
I have added a test to check if it works as intended for now

@ragesoss can you give me some `revision ids` for which `feature.enwiki.revision.shortened_footnote_templates` is not zero

ragesoss:
Here is an example:

https://en.wikipedia.org/w/index.php?title=Cymbeline&oldid=902872698
https://ores.wmflabs.org/v3/scores/enwiki/?models=articlequality&revids=902872698&features

ragesoss:
Also, rebasing should fix the failing spec.

spec/models/revision_spec.rb
```diff
@@ -29,6 +29,7 @@
   describe '#references_added' do
     let(:refs_tags_key) { 'feature.wikitext.revision.ref_tags' }
     let(:wikidata_refs_tags_key) { 'feature.len(<datasource.wikidatawiki.revision.references>)' }
+    let(:shootened_refs_tags_key) { 'feature.enwiki.revision.shortened_footnote_templates' }
```
ragesoss: Typo: `shootened` (also below)

app/models/wiki_content/revision.rb
```diff
@@ -70,14 +70,19 @@ def plagiarism_report_link
 
   WIKITEXT_REF_TAGS = 'feature.wikitext.revision.ref_tags'
   WIKIDATA_REFERENCES = 'feature.len(<datasource.wikidatawiki.revision.references>)'
+  WIKI_SHORTENED_REF_TAGS = 'feature.enwiki.revision.shortened_footnote_templates'
 
   def ref_tags
-    features[WIKITEXT_REF_TAGS] || features[WIKIDATA_REFERENCES]
+    return nil if features.empty?
+    (features[WIKITEXT_REF_TAGS] || features[WIKIDATA_REFERENCES] || 0) +
```
ragesoss: This is complicated enough that I think it makes sense to extract it to a method that can be used for both `features` and `features_previous`, instead of duplicating it.

ragesoss:
This looks good. I think it'll be ready to merge once the duplicate calculation is extracted.

khyatisoneji:
IDK if this is the correct way but it is easy to understand
let me know if you have a nice approach

ragesoss:
This still duplicates the logic.

I was thinking something more like:

```ruby
def reference_count(ores_features)
  (ores_features[WIKITEXT_REF_TAGS] || ores_features[WIKIDATA_REFERENCES] || 0) +
      (ores_features[WIKI_SHORTENED_REF_TAGS] || 0)
end
```

Then you'd use that method once with `features`, and once with `features_previous`.

khyatisoneji:
But this logic will not work for this test case:
`not a new article but features_previous` is not available / empty`

ragesoss:
@khyatisoneji you could still use whatever guard statements are needed to keep the functionality the same. The idea is just to isolate the part of the code that turns an ORES data hash into an integer representing the number of references, and not duplicate that.

ragesoss:
Looks like there's a behavior change that's causing the course page specs to fail.

PR #3734, opened 2020-02-09
Added Page Pile option for category tracking

### LINKED ISSUE #3357: Add Pagepile option for Category tracking
Pagepile is a tool for representing a static list of articles — both existing and not — on a single wiki: https://tools.wmflabs.org/pagepile/

This could be a very similar option to the PetScan one for a Category record on the Dashboard, except that instead of needing to be updated regularly, it would just need to be imported once since it is a static list. That would make for a convenient way to track progress on creating missing articles as well.

khyatisoneji:
## What this PR does
Fixes #3357 

## Screenshots
After:
image

image


lib/pagepile_api.rb
```diff
@@ -0,0 +1,34 @@
...
+  rescue StandardError => e
+    raise e unless Errno::EHOSTUNREACH.include?(e.class)
+    return {}
+  end
+
+  def page_titles_for_pileid(pileid)
+    titles_response = get_titles(pileid)
+    return [] if titles_response.empty?
+    titles = titles_response['pages']
+    binding.pry
```
ragesoss: This is breaking the build.

khyatisoneji: Yes I wanted to ask about this, what if the id entered has no results

khyatisoneji: For the test case, I have tested for a id which gives results, what be the behavior when there is no ouput? Show a error sign saying no results found?

ragesoss: I don't know! I suggest testing with an invalid ID, like some very high number, and see what happens.

ragesoss:
If you rebase, the build will hopefully be fixed.

khyatisoneji:
It returns `nil` whereas in site (page pile) it says `*Fatal error`*
So I was asking if here too show error or just no record found for this
pile id

ragesoss:
Hmm... I guess showing 'no record found' makes the most sense to me.

PR #3744, opened 2020-02-13
Navbar

puthalalitha:
Creates a React component Campaign navbar.

 Before
Home tab
 
Screen Shot 2020-02-18 at 12 46 04 PM

ORES tab
Screen Shot 2020-02-18 at 12 46 41 PM

Alerts tab
Screen Shot 2020-02-18 at 12 47 06 PM

After
Home tab
Screen Shot 2020-02-18 at 12 51 12 PM

ORES tab
Screen Shot 2020-02-18 at 12 52 02 PM

Alerts tab
Screen Shot 2020-02-18 at 12 52 30 PM


puthalalitha:
error  RECEIVE_ARTICLE_FEEDBACK not found in '../constants' 
error  POST_USER_FEEDBACK not found in '../constants'  
error  DELETE_USER_FEEDBACK not found in '../constants'  these errors are appearing in the terminal. Please check the PR. Isolated the navbar part.  Still, the campaign navbar is not rendering. These errors are causing the problem. I am not sure.

ragesoss:
Looks like the navbar is not rendering because of the the guard statement in the campaign.jsx `render`. It's rendering just an empty div, rather than the navbar.

ragesoss:
The reason that's happening is because the show.json.jbuilder file doesn't include the extra data that you added in the other branch. You should add that back in (as the extension of `show` data is also what Glory needs for the Android app, and we'll try to get this shipped next week).

ragesoss:
For this first minimal version, let's only have the changes appear on the two React-based tabs: ores_plot and alerts. The other tabs should use the server-rendered nav for now, instead of the React one. The reason for this is because of the `download stats` modal on the overview, which using a new react_root on that tab will interfere with until we have the whole overview tab ported to React.

app/assets/javascripts/constants/index.js
```diff
@@ -12,7 +13,6 @@ export * from './course';
 export * from './course_alerts';
 export * from './did_you_know';
 export * from './exercises';
-export * from './feedback';
```
ragesoss: Looks like this is the cause of the `gulp` errors.

puthalalitha:
I added the missing line in the constants/index.js file. I added the new data to show.json.jbuilder file.  It looks like the data is not coming. Still, DevTools are failing.  Please let me know what is the problem.

ragesoss:
I don't think the DevTools warnings are related to your code; I'm not sure what changed, but it's affecting the development build on other branches as well. You can ignore that for now.

Are there errors related to the json data not coming through? Are the necessary instance variables being set in the controller?

puthalalitha:
I am checking those values again.

puthalalitha:
Setting this value `@is_admin = current_user@.admin? ` in the overview method in the campaign_controller.rb giving the problem.  Now it is receiving its data.

app/views/campaigns/overview.html.haml
```diff
@@ -1,6 +1,7 @@
 - content_for :before_title, "#{@campaign.title} Overview  — "
```
ragesoss: Let's not make any changes to overview.html.haml for this PR; it'll just affect the ores_plot and alerts tabs.

app/views/campaigns/edit.html.haml
```diff
@@ -139,7 +140,15 @@
           %p.rails_editable-content
             = preserve @campaign.template_description
           = text_area(:campaign, :template_description, class: 'rails_editable-input', maxlength: 65535)
-
+      - if (current_user&.admin?)
```
ragesoss: This change is good, I think, and we can keep it independent of the React conversion of the overview tab.

app/assets/javascripts/components/common/campaign_navbar.jsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <div className="campaign-nav__wrapper">
+      <div className="campaign_navigation">
+        <div className="container">
+          <a className="nav__item">
+            <h2 className="title">Campaign: {campaign.title}</h2>
+          </a>
+          <nav>
+            <div className="nav__item" id="overview-link">
+              <p><NavLink activeClassName="active" to={`/campaigns/${campaign.slug}/overview`}>Home</NavLink></p>
```
ragesoss: This one shold be an anchor tag for now; we'll just use NavLink for the tabs that are being converted.

app/assets/javascripts/components/util/routes.jsx
```diff
@@ -23,15 +24,16 @@ const routes = (
     <Route path="/courses/:course_school/:course_title" component={Course} />
     <Route path="/course_creator" component={ConnectedCourseCreator} />
     <Route path="/users/:username" component={UserProfile} />
-    <Route path="/campaigns/:campaign_slug/overview" component={CampaignOverviewHandler} />
+    {/* <Route path="/campaigns/:campaign_slug/overview" component={CampaignOverviewHandler} /> */}
     <Route path="/campaigns/:campaign_slug/alerts" component={CampaignAlerts} />
-    <Route path="/campaigns/:campaign_slug/ores_plot" component={CampaignOresPlot} />
+    {/* <Route path="/campaigns/:campaign_slug/ores_plot" component={CampaignOresPlot} /> */}
```
ragesoss: These can be deleted rather than just commented out... once we have the new versions working, no reason to save the old code.

app/views/campaigns/show.json.jbuilder
```diff
@@ -9,5 +9,31 @@ if @campaign
...
+    json.course_string_prefix @presenter.course_string_prefix 
+    json.show_the_create_course_button Features.open_course_creation? && current_user&.admin?
+    json.editable @editable
+    json.register_accounts @campaign.register_accounts
+    json.current_user_admin @is_admin
+    json.requested_accounts_any @campaign.requested_accounts.any?
+    json.organizers_any @campaign.organizers.any?
+    json.organizers @campaign.organizers
+    json.current_user current_user
+    json.template_description_present @campaign.template_description.present?
```
ragesoss: I think everything from current_user_admin down through template_description_present should be excluded for now. Some of this I think will ultimately be better handled on the frontend.

We will need `organizers` later on, but we shouldn't pass the entire User record for each one (as this code would do), just probably the usernames.

ragesoss:
Please include screenshots of before/after for the two affected tabs as well.

app/views/campaigns/edit.html.haml
puthalalitha: What do you mean by "we can keep it independent of the React conversion of the overview tab"? Does that mean I can not remove the code line number 41 to 49 in the overview.html.haml  that relates to the Disable/ Enable account request button?

app/views/campaigns/overview.html.haml
puthalalitha: I took out a few lines of the code, line number 41 to 49 from overview.html.haml file that relates to the button Disable/Enable account request.  I will add those lines.

app/views/campaigns/edit.html.haml
ragesoss: Yes, exactly.

app/assets/javascripts/components/common/campaign_navbar.jsx
puthalalitha: I changed the Home tab, NavLink to the anchor tag.

app/assets/javascripts/components/util/routes.jsx
puthalalitha: I removed those commented lines.

app/views/campaigns/show.json.jbuilder
puthalalitha: I removed the lines from show.json.jbuilder file 29 to 34.

app/views/campaigns/edit.html.haml
puthalalitha: I added that code to overview.html.haml file. So I can keep those lines that belong
 to Disable/Enable account request button in the edit.html.haml file?

puthalalitha:
Added screen shots and I realized that the Alerts tab is not rendering from react component. I am trying to figure out the reason for it.

app/assets/javascripts/components/campaign/campaign.jsx
```diff
@@ -33,8 +33,8 @@ export const Campaign = createReactClass({
           campaign={this.props.campaign}
         />
         <Switch>
-          <Route exact path={`/campaigns/${this.props.campaign.slug}/ores_plot`} component={CampaignOresPlot} />
-          <Route exact path={`/campaigns/${this.props.campaign.slug}/alerts`} render={() => <CampaignAlerts campaign_slug={this.props.campaign.slug} />} />
+          <Route exact path="/campaigns/:campaign_slug/ores_plot" render={() => <CampaignOresPlot campaign_slug={this.props.campaign.slug} />} />
```
ragesoss: I think you can just do `component=` here like before.

app/assets/javascripts/components/util/routes.jsx
```diff
@@ -23,15 +21,13 @@ const routes = (
     <Route path="/courses/:course_school/:course_title" component={Course} />
     <Route path="/course_creator" component={ConnectedCourseCreator} />
     <Route path="/users/:username" component={UserProfile} />
-    <Route path="/campaigns/:campaign_slug/overview" component={CampaignOverviewHandler} />
```
ragesoss: Leave this overview route in place, as it is needed for rendering the 'Download stats' button and we don't yet have the new React version of the overview ready.

app/assets/javascripts/components/campaign/campaign_ores_plot.jsx
```diff
@@ -37,7 +38,7 @@ const CampaignOresPlot = createReactClass({
 
   fetchFilePath() {
     $.ajax({
-      url: `/campaigns/${this.props.match.params.campaign_slug}/ores_data.json`,
+      url: '/campaigns/:campaign_slug/ores_data.json',
```
ragesoss: This format only works as part of the `path` for a `Route` component; it uses pattern matching to find what the campaign slug is in the current path, and passes that as part of the `match` prop to the component it renders, if the route gets rendered via `component=` rather than `render=`.

Here, you can get the slug from the campaign prop instead, since that's how you've set it up so far. `/campaigns/${this.props.campaign.slug}/ores_data.json`

app/assets/javascripts/components/campaign/campaign_ores_plot.jsx
```diff
@@ -65,7 +66,14 @@ const CampaignOresPlot = createReactClass({
       }
       return <div>No Structural Completeness data available</div>;
     }
-    return (<button className="button small" onClick={this.show}>Change in Structural Completeness</button>);
+    return (
+      <div className="container">
+        <section className="overview container">
+          <CampaignStats campaign={this.props.campaign} />
```
ragesoss: This should be rendered all the time, not just before the `show` state is activated. I think it makes more sense to put this into campaign.jsx and have it appear on the Alerts tab as well, instead of rendering it from campaign_ores_plot.jsx.

app/assets/javascripts/components/campaign/campaign.jsx
```diff
@@ -0,0 +1,52 @@
...
+      return <div />;
+    }
+    return (
+      <div>
+        <CampaignNavbar
+          campaign={this.props.campaign}
+        />
+        <Switch>
+          <Route exact path="/campaigns/:campaign_slug/ores_plot" render={() => <CampaignOresPlot campaign={this.props.campaign} />} />
+          <Route exact path="/campaigns/:campaign_slug/alerts" render={() => <CampaignAlerts campaign_slug={this.props.campaign.slug} />} />
```
ragesoss: The error in the alert tab is from not receiving the `match` prop as expected. You can fix that by using `component=` instead of a render function here:
`<Route exact path="/campaigns/:campaign_slug/alerts" component={CampaignAlerts} />`

app/assets/javascripts/components/campaign/campaign.jsx
```diff
@@ -0,0 +1,52 @@
...
+    match: PropTypes.object,
+  },
+
+  componentDidMount() {
+    const campaignSlug = this.props.match.params.slug;
+    return this.props.getCampaign(campaignSlug);
+  },
+
+  render() {
+    if (!this.props.campaign.uploads_in_use_count) {
```
ragesoss: I believe the specs are still failing because of this... if the uploads_in_use_count value is 0, that's still a falsy value so this will skip rendering even after it receives the 0 value from the server.

You can do something like `if (this.props.campaign.uploads_in_use_count === '')` instead, so that will only skip rendering in the case of the initial empty string state, and any integer received from the server will allow it to render.

(A better approach here would be to add a `loading` attribute to the reducer, with `true` as the default, and then set it to false once the campaign data is received. But I think keeping it simpler for now is fine.)

app/assets/javascripts/actions/campaign1_actions.js
```diff
@@ -0,0 +1,33 @@
+import {
```
ragesoss: This file needs a better name. For now, I suggest `campaign_view_actions.js`.

`campaign_actions.js` is ultimately probably the right name, but we'd first need to rename the current file of that name to something like `course_campaign_actions`, since it's actions related to the campaigns of a particular course rather than the campaign itself.

puthalalitha: I going to change that file name.

app/assets/javascripts/components/campaign/campaign.jsx
khyatisoneji: I'd rather suggest using the `render` as doing it in component way gives returns different props, 
she can follow same method/ procedure used in `course.jsx` that is either defined some variable which contains required props and pass them

ragesoss: I'm fine with either way. The component way has the main advantage that it retains the interface for `CampaignAlerts` without requiring any changes, but the render way is more explicit and perhaps easier to understand in terms of data flow.

app/assets/javascripts/components/alerts/campaign_alerts.jsx
```diff
@@ -19,6 +19,7 @@ class CampaignAlerts extends React.Component {
   }
 
   getCampaignSlug() {
+    console.log(this.props.match.campaign_slug);
```
ragesoss: Please remove the console log.

ragesoss:
I squashed and merged this in a new branch. Yay!

PR #3753, opened 2020-02-14
Added Page Pile option for category tracking

### LINKED ISSUE #3357: Add Pagepile option for Category tracking

khyatisoneji:
Fixes #3357 

image


ragesoss:
This works fine as long as the wiki of the pile matches the one set from the Dashboard, but it doesn't handle cross-wiki piles. I think the wiki of the pile should be extracted from the pagepile API rather than use the one that the user submitted with (which will usually just be the default), and the Category record should be updated with that wiki. Ideally, the UI for adding a PagePile ID would not include the wiki chooser at all. (I've just realize the same issue probably affects the PetScan option as well.)

config/locales/en.yml
```diff
@@ -323,24 +320,28 @@ en:
   categories:
     add_category: Add category
     add_psid: Add PetScan PSID
+    add_pileid: Add PagePile id
     add_template: Add template
     add_this_category: Add this category
     add_this_psid: Add this PSID
+    add_this_pileid: Add this Page pile id
```
ragesoss: Use consistent terminology for PagePile ID in all interface message. I think "PagePile ID" is the best option.

khyatisoneji:
Can you please point to part where fetching according to wiki is done
I'm not able to find/ understand how result is fetched according to wiki

ragesoss:
In the UI when you add a PSID / PagePile ID / Category, the wiki selector is below the input.

Whichever wiki is set there will be the one that the `Category` record uses for determining which articles are part of the set. The `Category` record stores an array of titles, and then finds `Article` records that match the title and wiki when determining whether any given edited article is within the scope.

For PagePile and PSID, the wiki is part of the data associated with a given ID. In the current implementation, we're only getting the set of titles from PagePile or PetScan, so basically it won't work as expected if the wiki happens to be different than the one in the `Category` record.

khyatisoneji:
You're talking about the part of when we added `PSID` option in article finder and then fetching accordingly in that case wiki doesn't come into picture for `PSID`, directly fetching `articles titile` and other data from the json data received.

ragesoss:
@khyatisoneji article_finder_actions doesn't relate to category tracking, unless I'm missing something.

khyatisoneji:
All the code to fetch/ receive article titles/details  for any category/keyword is in `article_finder_actions.js`

ragesoss:
`article_finder_actions.js` has code for queries that happen directly between the client and mediawiki, for exploring categories via ArticleFinder. But for `Category` records on the server, it uses `CategoryImporter` instead. Whichever kind of `Category` record it is (whether actual wiki categories, or petscan, or template, or pagepile), the basic way it works is the same: the dashboard gets a set of page titles and saves that on the `Category` record, and then to see if an edit is in that set, it does a query for all the article ids that match the titles for whatever wiki the `Category` record has (which is determined when the user adds the category, based on the wiki chooser UI).

So `title_list_from_wiki` currently uses the wiki for the `category` and `template` source types, but it just assumes the wiki matches for `petscan` and `pagepile`. And if it does not match, then it will take that list of titles from petscan/pagepile, but the `article_ids` method will look for articles from the dashboard database on whatever wiki is set on the record, not whatever wiki the actual petscan or pagepile was for. So I think, instead, we need to remove the wiki chooser UI from petscan and pagepile queries, and then set the wiki on the `Category` record based on whatever wiki the petscan/pagepile query says.

khyatisoneji:
I have removed to change `wiki` in UI for `PagePile`, json received from `PagePile` has `wiki` in it whereas in `PetScan` there is no data of `wiki`, so i have just made changes in `article_ids` function for `PagePile`

khyatisoneji:
Added description in popover
image

I'm not sure if this is the way you wanted to display

khyatisoneji:
Removed the `change wiki` option for `PetScan` 
Also, `PetScan` test is failing with this error:
image
 I'm not sure how to fix this

ragesoss:
It should probably wrap so the popover isn't so wide, but otherwise, yes, I think that's a good way to display it. Maybe use a max-width for a css class for popover explanation?

ragesoss:
I'm not sure about the spec failure. Did that pass before and it fails now? Or is this the one marked `pending` that you're trying to fix?

khyatisoneji:
It was showing `pending` earlier now it shows  `*F`

spec/models/category_spec.rb
```diff
@@ -40,7 +40,7 @@
     context 'for psid-source Category' do
       let(:category) { create(:category, name: 9964305, source: 'psid') }
       let(:course) { create(:course) }
-      let!(:article) { create(:article, title: 'A cappella') }
+      let!(:article) { create(:article, title: 'A cappella', id: 2411) }
```
khyatisoneji: I added that because, I saw that here it was taking id as `100001974`, whereas the id is `2411`
I saw in `PetScan result`, the id for the article `A cappella` is `2411`

ragesoss: The id of an Article record won't match the id from petscan. There's a different id in Article for the mediawiki page id.

khyatisoneji: So how to test in that case?

ragesoss: For testing this, I wouldn't worry about Article records at all. The way you can find out if feature worked is to check for the expected title in the Category's `article_titles`. If it's there, then if/when the corresponding `A

khyatisoneji: Okay, I got it that, have to `wiki` from `PetScan` and just use the `Article.where(namespace: 0, wiki_id: wiki_id, title: article_titles).pluck(:id)`

But I'm not sure how to get the `Wiki_id` from the `Wiki`

ragesoss: If you're going the route we discussed of just warning people that they need to use petscan with the right wiki, then you won't need to worry about extracting the wiki from petscan data.

config/locales/en.yml
ragesoss: Please update all the messages to "PagePile ID".

spec/models/category_spec.rb
```diff
@@ -61,6 +61,32 @@
       end
     end
 
+    context 'for pileid-source Category' do
+      let(:category) { create(:category, name: 28301, source: 'pileid') }
```
ragesoss: Describe this example — what all is in this Page Pile? — and include a link to the pile in a comment.

Probably the name should be a string, since that's how it gets saved. It works the same either way, but that will help other developers recognize that the `Category` records have string values and not integers for `name`.

spec/models/category_spec.rb
```diff
@@ -61,6 +61,32 @@
...
+      let(:course) { create(:course, id: 28301) }
```
ragesoss: This doesn't seem necessary. Setting the ID on the `Course` to be the same as the pileid is potentially going to cause confusion, and the course ID isn't relevant to what you're testing.

spec/models/category_spec.rb
```diff
@@ -61,6 +61,32 @@
...
+        expect(described_class.last.article_titles).to be_empty
+
+        VCR.use_cassette 'categories' do
+          described_class.refresh_categories_for(Course.find_by(id: 28301))
+          expect(described_class.last.article_titles).not_to be_empty
+          expect(described_class.last.article_ids).to include(article.id)
+        end
+      end
+
+      it 'return empty list' do
```
ragesoss: I'm unclear what is being tested here. It looks like Course 1234 doesn't exist during the test.

spec/models/category_spec.rb
ragesoss: This comment is still relevant. A comment should describe PagePile 28301, so we have some idea of why 'America' is an expected result. The comment should also link to that pile, so we can see what else is in the pile and more easily make adjustments to the test later if we need to.

PR #3761, opened 2020-02-18
Update box-shadow for student drawer

bwreid:
#### What this PR does

This is a minor change that changes the look of the revision box. It removes a small sliver of white at the bottom of the drawer.

##### Before

image

##### After

image


ragesoss:
Works for me. The one other place I can think of that this could affect is the 'recent activity' admin view (`/recent-activity/recent-edits`), which uses a drawer as well (and which for some reason doesn't currently have that same white sliver), but I don't think this will have any negative effects there (and it's a page that sees very little use anyway).

Issue #3746, opened 2020-02-13
Instruction box for students/articles view seems like a drag-and-drop target


ragesoss:
The dotted line border is often used to indicate a drop zone, so some users will attempt to drag a student into that box rather than click on the name.

Maybe a solid border or a background color instead?

![drag and drop](https://user-images.githubusercontent.com/848483/74483952-7683b900-4e6c-11ea-9a0b-1f3f64810453.png)

Issue #3745, opened 2020-02-13
In students/articles view, clicking a name near the end of a long list doesn't change visible content


ragesoss:
When the list of students is long enough that you must scroll down, clicking on one of those students results in an update of the content at the top of the page, but you have to scroll back up to actually see the details for the student you just clicked.

Maybe the students list sidebar should be in its own scrollable container, separate for the scroll position of the page?

PR #3763, opened 2020-02-18
Minor visual updates for students/articles page

bwreid:
## What this PR does

- Add scrollbar for students
- Add dark background for No Student Selected
- Update title for students/articles page

This should solve issues #3745 and #3746 

## Screenshots
Before:

image

After:

image

image

image

### DIFF
--- a/app/assets/javascripts/components/students/containers/Articles.jsx
+++ b/app/assets/javascripts/components/students/containers/Articles.jsx
@@ -36,7 +36,7 @@ export class Articles extends React.Component {
             <StudentsSubNavigation
               course={course}
               current_user={current_user}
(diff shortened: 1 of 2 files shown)


ragesoss:
Nice. Is it possible to remove the non-functional scrollbar when there aren't enough students for it to actually scroll? It's not a big deal eithe way, as the bar-less scrollbar actually looks like it's just part of the visual design to connect the student boxes, but I think it'd be a tiny bit nicer without it.

bwreid:
I was looking in to this and I didn't immediately come across a solution. I can check a bit more or I can make another issue out of it.

ragesoss:
Sure, opening an issue sounds good to me.

Issue #2360, opened 2019-01-07
When a timeline start date is Sunday, the previous week shows up as an empty week 1


ragesoss:

Ace-Bansal:
@ragesoss  I'd like to work on this.

tonynguyen111997:
Hi @Ace-Bansal, I noticed that this issue has been open for a month now and I couldn't find a PR on it. Are you still interested in working on this bug? If not I'd like to attempt to tackle it.

ragesoss:
@tonynguyen111997 go for it. Let us know if you have any questions.

Manasa2850:
@ragesoss can I work on this?

ragesoss:
Go for it.

Manasa2850:
@ragesoss could you please let me know where I can find the files related to this issue?

ragesoss:
The place to start looking is `timeline_handler.jsx`. Most of the logic related to dates on the timeline is in `CourseDateUtils`, so that's where you'll probably end up.

Manasa2850:
Ok. I'll have a look at it.

abdelquodr:
Hi @ragesoss. Is anyone working on this ticket?

ragesoss:
@abdelquodr no, no one is working on it. since it's been a few years, i recommend starting by confirming that the issue is still happening.

nexpectArpit:
@ragesoss working on this!

Issue #618, opened 2016-01-19
First week, with no meeting days, is still a Week with blocks on the timeline


ragesoss:
See https://dashboard-testing.wikiedu.org/courses/test/wiki_output_%28test%29/timeline

The course starts 2016-01-01 (a Friday) and meets on Tuesdays. The first week of content in the timeline (ie, the first non-blackout week) should therefore be the week of 2016-01-03. Instead, the frontend starts the content from the week of 2015-12-27 even though there are no meeting days that week.

Fixing this would be a good opportunity to also move the rest of the timeline date logic to the server side. That should be a little easier since this: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/127c2caf636a90ed901ff4b474b258e201cc4a9d

With the new CourseMeetingsManager#meeting_dates_of(week) method, we could send all the week meeting dates as part of the data for each week.

ragesoss:
It appears that CourseDateUtils.weekMeetings does not do quite the same thing as CourseMeetingsManager.week_meetings, in particular for the case where the first partial calendar week does not include any of the normal meeting days.

PR #3771, opened 2020-02-20
courses starting on sunday do not have empty first week

Manasa2850:
## What this PR does
< Resolves issue #2360  >

## Screenshots
Before: 
Timeline_Sunday(before)
 
After:
Timeline_Sunday

I've made the first day of the week as Sunday irrespective of which day the course starts on. The timeline is looking fine for all courses irrespective of which day the course starts on,  in my local machine. Please let me know if I need to make some changes.

### DIFF
--- a/app/assets/javascripts/utils/course_date_utils.js
+++ b/app/assets/javascripts/utils/course_date_utils.js
@@ -129,13 +129,8 @@ const CourseDateUtils = {
     __range__(0, (courseWeeks - 1), true).forEach((week) => {
       weekStart = moment(recurrence.startDate()).startOf('week').add(week, 'weeks');
 
-      // Account for the first partial week, which may not have 7 days.
       let firstDayOfWeek;
-      if (week === 0) {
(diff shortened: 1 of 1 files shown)


ragesoss:
Hmm... it looks like the current logic was put in here: https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/618

I think this change would re-introduce the problem from that issue, where there is an empty week that shouldn't be present in cases where the start date of the course is after any meeting dates that would be in the first week.

Manasa2850:
@ragesoss I created a course which starts on Friday(28th Feb) and meets on Tuesday as mentioned in #618. What should the first week look like ideally?

Manasa2850:
It should start from 1st March (Sunday) right?
I'll make the necessary changes.

ragesoss:
That sounds right.

Manasa2850:
@ragesoss I have my exams coming up so I'll continue working on this issue next week.

ragesoss:

Issue #3596, opened 2019-12-19
Course nav bar  and Articles tab should change "Articles" label depending on course type


ragesoss:
### Current Behavior:
The navbar always says 'Articles', and that tab's header says 'Articles Edited' (and Assigned Articles, and Available Articles).

### Desired Behavior:
For courses with wikidata as the home wiki, these Article labels should instead say "Items".

For courses with other non-Wikipedia home wikis, the labels should say "Pages".

### Additional context:
We already do similar switching of labels for the Students tab and other interface messages using `course.string_prefix`. The same strategy probably makes sense here. We made need a third prefix for wikidata.

Muskan-Mittal:
Hi, can I take this issue?

noordean:
I would like to work on this @ragesoss . Can I go ahead?

ragesoss:
@noordean go for it

sastava007:
> I would like to work on this @ragesoss . Can I go ahead?

@noordean Any progress?

noordean:
@sastava007 Please go ahead. I have been pretty busy lately.

PR #3772, opened 2020-02-20
Developed a prefix scheme to display different label based on home wi…

sastava007:

## What this PR does
This displays different labels on course navbar based on home wiki data. 
This solves the issue #3596 

## Screenshots
Before:
Screenshot from 2020-02-21 01-53-49

After:
3596_after

Have created a prefix scheme to display different labels based on home wiki data further, I'll fix the labels in sub-navigation displaying articles Available/ Assigned and Edited in upcoming PRs.

### DIFF
--- a/app/assets/javascripts/components/common/course_navbar.jsx
+++ b/app/assets/javascripts/components/common/course_navbar.jsx
@@ -84,12 +84,6 @@ const CourseNavbar = ({ course, location, currentUser, courseLink }) => {
     );
   }
 
-  // ///////////////////////
-  // Prefix for home wiki /
-  // //////////////////////
-  let string_article_prefix = course.home_wiki.project === 'wikidata' ? 'wikidata' : 'others';
-  string_article_prefix = `${course.string_prefix}.${string_article_prefix}`;
-
   return (
     <div className="container">
       {courseLinkElement}
@@ -102,7 +96,7 @@ const CourseNavbar = ({ course, location, currentUser, courseLink }) => {
           <p><NavLink to={studentsLink} activeClassName="active">{CourseUtils.i18n('students_short', course.string_prefix)}</NavLink></p>
         </div>
         <div className="nav__item" id="articles-link">
-          <p><NavLink to={articlesLink} activeClassName="active">{CourseUtils.i18n('articles_short', string_article_prefix)}</NavLink></p>
+          <p><NavLink to={articlesLink} activeClassName="active">{CourseUtils.i18n('articles_short', course.wiki_prefix)}</NavLink></p>
         </div>
         <div className="nav__item" id="uploads-link">
           <p><NavLink to={uploadsLink} activeClassName="active">{I18n.t('uploads.label')}</NavLink></p>
--- a/app/assets/javascripts/components/common/course_navbar.jsx
+++ b/app/assets/javascripts/components/common/course_navbar.jsx
@@ -84,6 +84,12 @@ const CourseNavbar = ({ course, location, currentUser, courseLink }) => {
     );
   }
 
+  // ///////////////////////
+  // Prefix for home wiki /
+  // //////////////////////
+  let string_article_prefix = course.home_wiki.project === 'wikidata' ? 'wikidata' : 'others';
+  string_article_prefix = `${course.string_prefix}.${string_article_prefix}`;
+
   return (
     <div className="container">
       {courseLinkElement}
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -1106,9 +1086,6 @@ en:
     enroll_confirmation: Are you sure you want to add %{username}?
     enroll_url: 'Users may join by visiting this URL:'
     enrolled_success: '%{username} was added successfully!'
-    exercise_modules_completed:
```
ragesoss: Looks like you've deleted a lot of interface messages on accident.

ragesoss: I guess this may be related to starting from an older version of the code, and then overwriting the changes the happened in the meantime with your version of en.yml. I recommend starting from the latest version and re-adding your new i18n messages.

app/assets/javascripts/components/common/course_navbar.jsx
```diff
@@ -84,6 +84,12 @@ const CourseNavbar = ({ course, location, currentUser, courseLink }) => {
     );
   }
 
+  // ///////////////////////
+  // Prefix for home wiki /
+  // //////////////////////
+  let string_article_prefix = course.home_wiki.project === 'wikidata' ? 'wikidata' : 'others';
+  string_article_prefix = `${course.string_prefix}.${string_article_prefix}`;
```
ragesoss: Interesting. With this approach, we end up with two different branches for any given label: course type (course string prefix), and wiki (article prefix).

I think the terminology we use in place of "Article" will be independent of course type, so branching on course string prefix results in duplication.

I think a better approach would be make sure that all the article labels are not in the courses/courses_generic message trees, and instead move any you find from there to `articles` message tree, and then have alternative branches for `pages` and `items`.

The `article_string_prefix` is probably a property that should come from the `Wiki` model and be passed from the server to the frontend.

config/locales/en.yml
sastava007: This time I made my changes over latest `en.yml` file.

app/assets/javascripts/components/common/course_navbar.jsx
sastava007: I have created a wiki_prefix in the backend instead of initializing itself in course_navbar component as I was doing last time. And using that data I'm fetching labels under Articles which will be independent of course type.
@ragesoss Please review.

app/views/courses/course.json.jbuilder
```diff
@@ -27,6 +27,7 @@ json.course do
   json.published CampaignsCourses.exists?(course_id: @course.id)
   json.closed @course.closed?
   json.enroll_url "#{request.base_url}#{course_slug_path(@course.slug)}/enroll/"
+  json.wiki_prefix @course.home_wiki.project == 'wikidata' ? 'articles.items' : 'articles.pages'
```
ragesoss: I think it makes the most sense to make `string_prefix` a method on the `Wiki` model, in parallel to the `string_prefix` method of `Course`. This way, we can use the same logic for server-rendered i18n messages as well, and easily include the prefix data in other endpoints besides course.json.

(Then you'd do something like `json.wiki_string_prefix @course.home_wiki.string_prefix` here.)

config/locales/en.yml
```diff
@@ -218,6 +218,10 @@ en:
       usually indicates that an article follow Wikipedia's conventions for the
       structure of a good article, but it does not take into account the content
       itself. (These scores are only available for English Wikipedia articles.)
+    items:
```
ragesoss: There are three different possible names for "Article" which we want to support:
* "Article" for wikipedia
* "Item" for wikidata
* "Page" for anything else

I think instead of having these be within the `articles` i18n branch, we should follow the same pattern as with `courses` and `courses_generic`. To keep them near each other in the file and match the pattern of `courses_generic`, this might mean the new branch keys are `articles_wikidata` and `articles_generic`. So we'd add a new set of messages as the same level as the `articles` branch, and switch which branch to use based on the wiki prefix.

ragesoss:
This is coming along. There's a lot going on with the i18n scheme, and it wasn't obvious to me at the start what the right way to do it would be be, so thanks for exploring this!

sastava007:
So you want me to create two branches `articles_wikidata` for articles having wikidata as home wiki and `articles_generic` for all remainings?
But then how we'll differentiate the wikipedia from others, do I need to create some kind of sub-branch under `articles_generic` for handling wikipedia and everything else?

@ragesoss

sastava007:
> Yes, a new method, probably called `string_prefix`, should be added to the model. It would be an instance method.

Have created this method in wiki.rb what else I need to change to get the data into props?
image

ragesoss:
Something like `json.wiki_string_prefix @course.home_wiki.string_prefix` in the jbuilder file would add it to the data being received by the frontend.

There's no need to define it as an empty string first, or even use a variable. Ruby will implicitly return the last statement evaluated, so:

```ruby
case project
when 'wikipedia'
  'articles'
when 'wikidata'
  'articles_wikidata'
else
  'articles_generic'
end
```

sastava007:
@ragesoss 
I have shifted the `string_prefix` from json view template to wiki.rb model and tested for all three use cases (i.e program with homewiki as `wikipedia`/`wikidata`/ everything remaining )it's working fine for all of them.

app/models/wiki.rb
```diff
@@ -120,6 +120,17 @@ def course_prefix
     edit_templates['course_prefix']
   end
 
+  def string_prefix
+    case project
+      when 'wikipedia'
```
ragesoss: This isn't standard Ruby formatting. Try running `bundle exec rubocop` or configure your editor to do rubocop syntax highlighting.

ragesoss:
Nice! This works as expected, and it should be easy to extend the pattern to the other interface messages.

Issue #3643, opened 2020-01-06
In Uploads tab, UploadViewer file usage table displays differently depending on view mode


ragesoss:
### What is happening?
UploadViewer is the component to see details about a single file from the Uploads tab. The Uploads tab has three layout options: gallery, list, and grid. In list mode, the file usage table inherits some of the styling of the Uploads list itself (which is also a table), so the usage table has borders around the cells. In gallery and grid modes, the borders are gone.

### To Reproduce

Create a course with uploads, including some that are in use on Wikipedia articles. (The `populate` script should create such a course.)

Go to the Uploads tab, switch to list view, sort the list by usage count so you can find an upload that is in use, then switch to a different layout and view the same image.

### Expected behavior
The file usage table should have consistent styling regardless of Uploads layout. The borders are preferable to the borderless version (although further design changes to UploadViewer are welcome).

### Screenshots
In list layout:
![list](https://user-images.githubusercontent.com/848483/71853422-3d4b7280-3090-11ea-970a-767321911970.png)

In the other layouts:
![non-list](https://user-images.githubusercontent.com/848483/71853423-3d4b7280-3090-11ea-9a18-7c18bb4a5195.png)

shipcy:
Can I take this issue?

ragesoss:
yes, please do.

Harshwp:
Hello! I'd like to work on this

Tripathi-Vaibhav:
Hi! Can you share the link for its codebase ?

ragesoss:
@Tripathi-Vaibhav this comes from the `uploads` components in `app/assets/javascripts/uploads` and from the corresponding stylesheet code, probably in `app/assets/stylesheets/modules/_uploads.styl`. The `user_roles.md` documentation file has details about how to set up your environment to be able to test this out.

soumyaa1804:
@Tripathi-Vaibhav are you working on this issue? If not, I would like to work on it.

sastava007:
Sorry, @soumyaa1804 . But I'm already working on this. 
Pardon me I haven't left any comment on it, hope you understand.

Tripathi-Vaibhav:
No

Tripathi-Vaibhav:
No Soumya Sharma

soumyaa1804:
Okay, cool! @sastava007 @Tripathi-Vaibhav

tichnas:
Is this issue still open?

[...]

PR #3785, opened 2020-02-24
Fixed upload viewer bug

Manasa2850:
## What this PR does
Solves issue #3643 
The tables have borders even in the gallery and tile view.

GIF
uploads

### DIFF
--- a/app/assets/stylesheets/modules/_uploads.styl
+++ b/app/assets/stylesheets/modules/_uploads.styl
@@ -158,6 +158,6 @@
 .view-file-details
   border 1px solid #ddd
 
-.row-details
-  border-bottom 1px solid #ddd
-  padding 10px
+  .row-details
+    border-bottom 1px solid #ddd
+    padding 10px
--- a/app/assets/stylesheets/modules/_uploads.styl
+++ b/app/assets/stylesheets/modules/_uploads.styl
@@ -155,9 +155,9 @@
     .upload-viewer-button
       margin 10px
 
-.view_fileDetails
+.view-file-details
   border 1px solid #ddd
 
-.row_details
+.row-details
   border-bottom 1px solid #ddd
   padding 10px
--- a/app/assets/stylesheets/modules/_uploads.styl
+++ b/app/assets/stylesheets/modules/_uploads.styl
@@ -139,8 +139,7 @@
 
     table, td, th
       word-wrap break-word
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/modules/_uploads.styl
```diff
@@ -155,3 +154,10 @@
 
     .upload-viewer-button
       margin 10px
+
+.view_table
+  border 1px solid #ddd
+
+.element
```
ragesoss: This is not a very informative class name. Is there another term that would convey the intended specific use?

app/assets/stylesheets/modules/_uploads.styl
```diff
@@ -155,3 +154,10 @@
 
     .upload-viewer-button
       margin 10px
+
+.view_fileDetails
```
ragesoss: To follow the conventions of our css class names, this should probably be `view-file-details`, and `row-details`.

These should also not be top-level rules, so that we don't accidentally use the same class name somewhere else. It looks like these two classes would only be used for child elements of `upload-viewer`, so they should be scoped to only apply to child elements, like the rest of the classes in this file.

Manasa2850: Sorry about this @ragesoss . I'll definitely keep this in mind.

app/assets/stylesheets/modules/_uploads.styl
```diff
@@ -155,3 +154,10 @@
 
     .upload-viewer-button
       margin 10px
+
+.view-file-details
+  border 1px solid #ddd
+
+.row-details
```
ragesoss: This rule should be scoped to children of `view-file-details`. It's a class name that is pretty general, so we want to make sure it doesn't collide with the same name in some other file if it's a top-level rule.

Manasa2850: @ragesoss sorry I didn't get you. So what should I change the name of `view-file-details` to?

ragesoss: That name is fine. You should change the scope of `.row-details` so that it's a rule that only applies to children of `view-file-details` elements. ie, indent that rule so that it's a sub-rule of `view-file-details`.

ragesoss:
Looks good! Thanks!

PR #3787, opened 2020-02-24
changed the :slug to :campaign_slug in the routes.jsx

puthalalitha:
Fixed the  :slug  to /:campaign_slug  in the routes.jsx
 
ORES tab before:
Screen Shot 2020-02-24 at 10 57 40 AM

Alerts tab before:
Screen Shot 2020-02-24 at 10 59 02 AM

After:
ORES tab After:
Screen Shot 2020-02-24 at 10 57 40 AM

Alerts tab After:
Screen Shot 2020-02-24 at 10 57 55 AM

### DIFF
--- a/app/assets/javascripts/components/alerts/campaign_alerts.jsx
+++ b/app/assets/javascripts/components/alerts/campaign_alerts.jsx
@@ -19,7 +19,7 @@ class CampaignAlerts extends React.Component {
   }
 
   getCampaignSlug() {
-    return `${this.props.match.params.campaign_slug}`;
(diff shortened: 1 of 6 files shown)


ragesoss:
This change should just require edits to two lines: the routes.jsx, and the place in `campaign.jsx` where the slug is extracted from the `match` param`. Everything else should work without further changes, I think. I'd rather not switch to `render=` for the routes within `campaign.jsx` unless there's a good reason for it.

puthalalitha:
I did the changes as you suggested and pushed the code.

PR #3800, opened 2020-02-26
Add label for training module selector

### LINKED ISSUE #3791: When editing a timeline Block, the training module selector should label each option with its type
Currently, when you're adding a training module to a block, it's not clear which ones are trainings vs exercises vs discussions. The names of the modules don't make this clear in some cases, because there are cases where a training covers basically the topic as the exercise (but one is about how to do the task in general, and the other is about what specifically is required as an assigned exercise).

A good solution would add a label to the end of each title to specify which type of module it is.

module select

Kartik1397:
## What this PR does
Fixes #3791 
## Screenshots
Before:

before

After:

after


ragesoss:
@Kartik1397 I haven't tested it yet, but on a quick review the code and screenshots look great! The labels should use i18n instead of being hard-coded in English.

Kartik1397:
label's string are now updated with i18n

ragesoss:
Great work, thanks!

PR #3802, opened 2020-02-26
Created Home tab component in React

puthalalitha:
The Home tab is rendering from the server before now it is rendering from React component.
Before:
Screen Shot 2020-02-26 at 1 56 13 PM

After:
 
 
Screen Shot 2020-02-26 at 3 09 38 PM


app/assets/javascripts/components/alerts/campaign_alerts.jsx
```diff
@@ -19,11 +19,10 @@ class CampaignAlerts extends React.Component {
   }
 
   getCampaignSlug() {
-    return `${this.props.match.params.campaign_slug}`;
+    return `${this.props.campaign_slug}`;
```
ragesoss: This change probably shouldn't include any changes to other tabs, since we have those working already... just changes isolated to the Home tab.

puthalalitha: It looks like git rebase command didn't fix all the changes. Now I fixed all those changes and pushed the code.

ragesoss:
@puthalalitha it looks like this is getting closer!

The first failing spec, ./spec/controllers/campaigns_controller_spec.rb:405, is a controller spec that was written to look for the text of the campaign description in the HTML. In other words, it's a spec that is particular to the server-rendered version of the page. It should be removed, and replaced with a browser spec that confirms that the same thing still renders in the browser (even though that text is not part of the HTML now).

The other failing spec is a browser spec. See if you can pinpoint the reason why it's failing, as it might point to a difference in behavior between the server-rendered version and your new React version.

PR #3803, opened 2020-02-26
Jquery uls to webpack config#2312

### LINKED ISSUE #2312: Move jquery-uls to the webpack config

soumyaa1804:
## What this PR does
Fixes #2312  

## Screenshots
After changes been implemented. (nav bar is rendering)
image

Though it is working fine locally without any errors in the console, the `rspec spec/features` test failed on the command line.


soumyaa1804:
I should have removed vendor from minify.js.

soumyaa1804:
I tried to reproduce the failure `campaign overview page as a campaign organizer campaign details shows add organizers button and title field when in edit mode` and found that I am not getting add organizer option in the edit mode in the campaign page. The '+' is itself not rendering and the save button is not saving the changes. I am trying to fix it.

soumyaa1804:
@ragesoss, can you please help? Are the failures because of my changes or because of some previous merges as I can see similar errors in **pv-912**'s PR also.

bwreid:
@soumyaa1804 Hi there! Sometimes the feature tests just fail, but you can check this by running the specs on your own machine. If you click on the "Details" link, it'll bring you to Travis CI where you can look for the errors.

I've copied below the failed tests. Do these tests pass locally on your own machine?
```
Failed examples:

rspec ./spec/features/campaign_overview_spec.rb:168 # campaign overview page as a campaign organizer campaign details shows add organizers button and title field when in edit mode

rspec ./spec/features/campaign_overview_spec.rb:174 # campaign overview page as a campaign organizer campaign details updates the campaign when you click save

rspec ./spec/features/campaign_overview_spec.rb:183 # campaign overview page as a campaign organizer campaign details shows an error if you try to add a nonexistent user as an organizer

rspec ./spec/features/campaign_overview_spec.rb:201 # campaign overview page as a campaign organizer campaign details start and end dates shows an error for invalid dates

rspec ./spec/features/campaign_overview_spec.rb:211 # campaign overview page as a campaign organizer campaign details start and end dates updates the date fields properly, and unsets if #use_dates is unchecked
```

soumyaa1804:
Yes, they failed locally also. And they are true as I am able to reproduce the errors pointed out here locally. Like, the save button is not working and 'add organizer' option is not rendering on my system also.

ragesoss:
Those features are ones that depend on jquery; `app/assets/javascripts/campaigns.js` is a jquery-based bunch of code that adds interactivity to the campaign overview page. It sounds like either jquery is not available, or maybe it's loading in a different order than it did before.

soumyaa1804:
Okay, will try to fix it.

soumyaa1804:
I created an entry point for jquery in the webpack.js and the reference to its output file in _head.html.haml. The save button and the add organizer options are working fine now in the edit campaign page. Means, jquery is available to `app/assets/javascripts/campaigns.js`. 

image

But as the error shows it is unable to recognize `$` in `settings_actions.js`. Is there any issue with the ordering of jquery entry file or in loading jquery output file to `_head.html.haml`?
I don't know why it builds two jquery output files in `public/assets/javascripts`, `jquery.4151fc9295f7d667af49.js` and `jquery.4a9783bdb68264021ac8.js`
I am new to webpack and still learning. Please help.

soumyaa1804:
I am getting these errors in the console on the `\surveys` page with one survey.

image

ragesoss:
Hmm... I don't know. It looks like those two functions are from jquery plugins (slick-carousel and jquery-ui/widgets/tabs) that are imported by the respective files. Whether that's because of a change in the jquery version or because jquery isn't available in the environment in the same way, I'm not sure.

I also don't know why there would be two jquery outputs.

ragesoss:
In the travis build, I see just one jquery file in the output (plus the corresponding source map).

The settings_actions.js error may be related to the JavaScript test setup, if you're only seein that in the `jest` tests and not when running the app. `testHelper.js` includes code for making jquery available in the jest test environment, which might need to be adjusted.

soumyaa1804:
> I also don't know why there would be two jquery outputs.

My bad! I loaded gulp again and it showed only one output file.

soumyaa1804:
> It looks like those two functions are from jquery plugins (slick-carousel and jquery-ui/widgets/tabs) that are imported by the respective files.

In the [gulpfile.babel.js](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/gulpfile.babel.js#L8-L33) file, when we remove jquery-uls and add it to webpack the order of jquery and then stylesheet got disturbed. And we need to maintain the order of task in gulp so is it possible to bring task one by one from gulp to webpack if they are interdependent ? I am unable to come up with any explaination or way to solve this issue. Can you please guide further?

soumyaa1804:
I checked testHelper,js but that is importing jquery directly from node_modules So it should not be a reason jest failure.

gulp/tasks/webpack.js
```diff
@@ -78,7 +79,6 @@ function startWebpack(cb) {
       ]
     },
     externals: {
-      jquery: 'jQuery',
```
ragesoss: Maybe the problem is related to this. Within a lot of the other javascript files, we currently assume that global jquery is available. With this change, it might be disrupting the other files besides the jquery one and how jquery is made available to them.

ragesoss:
This got taken care of as part of the complete switch off of webpack that Amit wrapped up in May.

PR #3804, opened 2020-02-26
Make jquery global using webpack and remove jquery-uls

### LINKED ISSUE #2312: Move jquery-uls to the webpack config

pv-912:
Fixes #2312 
## What this PR does
- Add jquery as a global variable using webpack
- Removed building jquery-uls.js from gulp task


pv-912:
Didn't see that [PR](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/3803) is already opened for this issue just few hours ago. Please give priority to that one.

pv-912:
@ragesoss I tried to debug the error `campaign overview page as a campaign organizer campaign details shows add organizers button and title field when in edit mode`. There I found that in [app/assets/javascripts/campaigns.js](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/campaigns.js) on line 14 `.on('editable:edit')` is not triggering after clicking edit button due to which further nested functions aren't executing.

However, on clicking the edit button the [function](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/aca2e0e09ac87b9601da43129f6669928f416bb2/app/assets/javascripts/utils/editable.js#L39) in `/app/assets/javascripts/utils/editable.js`  to add class `rails_editable-editing`  is working fine.

Can you please help me out why this is happening?

ragesoss:
This got taken care of as part of the complete switch off of webpack that Amit wrapped up in May.

PR #3814, opened 2020-02-27
Fixed -Log in with Wikipedia- text alignment.

### LINKED ISSUE #3807: Disaligned 'Log in with Wikipedia' text on hover.
### What is happening?
The 'Log in with Wikipedia' should appear in the same line.

### To Reproduce

1. Go to application main page
2. Hover on 'Log in'

### Expected behavior
I guess it would be better to have all of it in the same line.

### Screenshots

Screenshot from 2020-02-28 00-50-45

### System Information

 - OS: Linux 18.04
 - Browser : Chrome
 - Latest Version

shashwatkathuria:
## What this PR does
Align the 'Log in with Wikipedia' text correctly. Fixes #3807 

## Screenshots
Before:
Screenshot from 2020-02-28 00-50-45

After:
Screenshot from 2020-02-28 02-16-11

### DIFF
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -110,15 +110,14 @@ nav.top-nav
     margin-bottom: -1px;
     display: inline-block;
     font-size 16px
-    vertical-align middle
+    vertical-align top
 
   .expand
-    overflow hidden
     display inline-block
     width 0
     transition all 250ms ease-in-out
     opacity 0
-    vertical-align middle
+    vertical-align top
     white-space nowrap
     text-align left
 
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -118,7 +118,7 @@ nav.top-nav
     width 0
     transition all 250ms ease-in-out
     opacity 0
-    vertical-align top
+    vertical-align middle
     white-space nowrap
     text-align left
 
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -113,6 +113,7 @@ nav.top-nav
     vertical-align top
 
   .expand
+    overflow hidden
     display inline-block
     width 0
     transition all 250ms ease-in-out
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -117,7 +117,7 @@ nav.top-nav
     width 0
     transition all 250ms ease-in-out
     opacity 0
-    vertical-align top
+    vertical-align inherit
     white-space nowrap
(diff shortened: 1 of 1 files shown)


ragesoss:
Using `vertical-align: middle` results in misalignment on dashboard.wikiedu.org:
vertical allgn middle

The bug you're trying to solve also does not show up on dashboard.wikiedu.org. (You should be able to replicate this with `wiki_education: 'true'` in `application.yml`.)

I'm not sure what is causing the difference in alignment between the two configuration. One guess is that this difference could be due to the language picker (which is not included by default on dashboard.wikiedu.org, but is on outreachdashboard.wmflabs.org). Another guess is that it's font-related, as the wiki_education config uses web fonts while the other does not.

shashwatkathuria:
Okay! I'll try to see what can be done.

shashwatkathuria:
@ragesoss Does any of these look good as a work around?

Peek 2020-02-29 11-21

Peek 2020-02-29 11-22

ragesoss:
Hmm... I think the stylized W is important to privide a connection between that link and Wikipedia; I'd rather have a fix for the alignment that keeps the current design.

shashwatkathuria:
Okay! I'll try once more to keep the alignment.

shashwatkathuria:
Screenshot from 2020-03-04 02-06-25
Screenshot from 2020-03-04 02-13-07

@ragesoss This is as close as I could get. I hope its better than before?

ragesoss:
For me (in Firefox) this change pushes it down so that the underline goes all the way to the bottom of the navbar.

mouseover

shashwatkathuria:
Okay! I'll try to experiment with Firefox also (was using Chrome till now).

shashwatkathuria:
Screenshot from 2020-03-14 02-35-50
Screenshot from 2020-03-14 02-36-27

shashwatkathuria:
@ragesoss Please review this once ^^

shashwatkathuria:
@ragesoss Please have a look

ragesoss:
This introduces some glitchy behavior when you mouse over the right side.
glitch

shashwatkathuria:
@ragesoss Please have a look.

shashwatkathuria:
Peek 2020-03-17 17-53
Peek 2020-03-17 17-54
Peek 2020-03-17 17-55
Peek 2020-03-17 17-56

ragesoss:
It seems like the root of the problem isn't the `align` property. The real issue, I think, is the distortion that gets introduced when the language picker is present. With the latest change, the positioning is better when the language picker is present, but worse when it is not.

Maybe the optimal solution is to tweak the language picker instead, so that it doesn't cause any change to the dimensions of the navbar.

shashwatkathuria:
Hmmmm......Maybe, I'll try to remove the language picker (for experimenting)! I'll reply in a few hours.

PR #3821, opened 2020-02-28
campaign_editable.jsx: Add isMulti for campaign selection

### LINKED ISSUE #2498: Select component for setting the campaign does not use isMulti
The implementation of the [campaign selection](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/components/overview/campaign_editable.jsx#L94-L103) appears like a multi-pick option but functionally acts like a dropdown in that you can only select one item at a time. 
dropdown
Is the intention here to make it look like you can type in it? It is possible to type in dropdowns as well.

ihsavru:
Fix #2498

### DIFF
--- a/app/assets/javascripts/components/overview/campaign_editable.jsx
+++ b/app/assets/javascripts/components/overview/campaign_editable.jsx
@@ -25,7 +25,9 @@ const CampaignEditable = createReactClass({
   },
 
   getInitialState() {
-    return {};
+    return {
+      selectedCampaigns: []
+    };
   },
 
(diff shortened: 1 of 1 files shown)


ragesoss:
This lets you add multiple campaigns at once, but upon clicking 'Add', only the first selected campaign gets removed from the select; the others remain, even though they were actually added.

keeps multiple

ihsavru:
@ragesoss fixed

ragesoss:
Thanks!

Issue #3822, opened 2020-02-28
Requested account creation display error.


shashwatkathuria:
### What is happening?
The requested account page is rendering JSON directly instead of processing and then displaying to the user.

### To Reproduce

   1. Create a request account enabled course
   2. Deliberately request an invalid account (or even a valid account).
   3. Open course requests
   4. Click on 'Create these accounts'
   5. See error

### Expected behavior
The output should be formatted.

### Screenshots
![Screenshot from 2020-02-28 21-26-32](https://user-images.githubusercontent.com/37243661/75564312-e94a8200-5a71-11ea-97ee-73221dedc58d.png)
![Screenshot from 2020-02-28 21-26-37](https://user-images.githubusercontent.com/37243661/75564326-eea7cc80-5a71-11ea-98e0-63bd2fa56942.png)

### System Information

 - OS: Ubuntu 18.04
 - Browser Chrome
 - Version

### Additional context
This is also mentioned in the TODO Comments of the appropriate controllers.

PR #3823, opened 2020-02-28
Requested account creation status display.#3822

shashwatkathuria:
## What this PR does
Solves #3822 . Displayes the status of requested accounts creation properly instead of unprocessed JSON Ruby Hash earlier.

## Screenshots
Before:
Screenshot from 2020-02-28 21-26-37

After:
Screenshot from 2020-02-29 01-23-47

### DIFF
--- a/app/controllers/requested_accounts_controller.rb
+++ b/app/controllers/requested_accounts_controller.rb
@@ -89,16 +89,14 @@ class RequestedAccountsController < ApplicationController
   def create_account(requested_account)
     creation_attempt = CreateRequestedAccount.new(requested_account, current_user)
     result = creation_attempt.result
-    if result[:failure]
-      result = CreateRequestedAccount.format_failure_requested_account_result(result)
-      return result
-    elsif result[:success]
+    result[:result_description] = creation_attempt.result_description(result)
+    if result[:success]
       # If it was successful, enroll the user in the course
       user = creation_attempt.user
       JoinCourse.new(course: requested_account.course, user: user,
                      role: CoursesUsers::Roles::STUDENT_ROLE)
-      return result
     end
+    return result
   end
 
   def check_creation_permissions
--- a/app/controllers/requested_accounts_controller.rb
+++ b/app/controllers/requested_accounts_controller.rb
@@ -89,7 +89,6 @@ class RequestedAccountsController < ApplicationController
   def create_account(requested_account)
     creation_attempt = CreateRequestedAccount.new(requested_account, current_user)
     result = creation_attempt.result
-    result[:result_description] = creation_attempt.result_description(result)
     if result[:success]
       # If it was successful, enroll the user in the course
       user = creation_attempt.user
--- a/app/controllers/requested_accounts_controller.rb
+++ b/app/controllers/requested_accounts_controller.rb
@@ -90,8 +90,7 @@ class RequestedAccountsController < ApplicationController
     creation_attempt = CreateRequestedAccount.new(requested_account, current_user)
     result = creation_attempt.result
     if result[:failure]
-      result[:failure] = result[:failure].sub('response: {}', '')
-      result[:failure] = result[:failure].sub('https://en.wikipedia.org', '')
+      result = CreateRequestedAccount.format_failure_requested_account_result(result)
       return result
     elsif result[:success]
       # If it was successful, enroll the user in the course
--- a/app/controllers/requested_accounts_controller.rb
+++ b/app/controllers/requested_accounts_controller.rb
@@ -25,7 +25,6 @@ class RequestedAccountsController < ApplicationController
       render json: { message: I18n.t('courses.new_account_submitted') }
       return
     end
-    
     result = create_account(@requested)
     if result[:success]
       render json: { message: result.values.first }
@@ -91,8 +90,8 @@ class RequestedAccountsController < ApplicationController
     creation_attempt = CreateRequestedAccount.new(requested_account, current_user)
     result = creation_attempt.result
     if result[:failure]
-      result[:failure] = result[:failure].sub("response: {}", "")
-      result[:failure] = result[:failure].sub("https://en.wikipedia.org", "")
+      result[:failure] = result[:failure].sub('response: {}', '')
+      result[:failure] = result[:failure].sub('https://en.wikipedia.org', '')
       return result
     elsif result[:success]
       # If it was successful, enroll the user in the course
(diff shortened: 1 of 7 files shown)


ragesoss:
Nice! A few notes:
* Remove the space before the colon in the 'Failure' and 'Success' labels.
* Since the code for turning a failure response hash into user-friendly text is used in two different places, it should be extracted and shared between them rather than duplicated. It might be appropriate to do that in `CreateRequestedAccount` or it might be better to do it in a separate helper file. Doing it in CreateRequestedAccount would be what I'd try first, because we want to avoid coupling the implementation of CreateRequestedAccount with the formatting code, which relies on English Wikipedia being the wiki where accounts are created; if that changed, then the hard-coded en.wiki URL would cause problems.
* "Failure" and "Success" should use i18n rather than hard-coded English.

shashwatkathuria:
@ragesoss I am confused whether i should add requested account status in I18n-en to both campaigns and courses, or something more general. Done with both for now.

app/services/create_requested_account.rb
```diff
@@ -26,6 +26,12 @@ def initialize(requested_account, creator)
     process_request(@creator, @creation_reason)
   end
 
+  def self.format_failure_requested_account_result(result)
```
ragesoss: This should be done as an instance method rather than a class method, so you can call something like `creation_attempt.result_description` to get the message that will be shown on the frontend. That will let you simplify the code in each of the controllers as well, since either way, `creation_attempt.result_description` will provide the message to show and you'll just need to use a conditional to determine whether to do the `JoinCourse` part.

config/locales/en.yml
```diff
@@ -626,6 +629,9 @@ en:
       one: '%{count} requested account pending.'
       other: '%{count} requested accounts pending.'
     requested_accounts_alert_view: 'View'
+    requested_account_status:
```
ragesoss: Doing this just one place and re-using the same i18n message for both paths is better.

shashwatkathuria: Done!

app/services/create_requested_account.rb
shashwatkathuria: Done!

app/controllers/requested_accounts_controller.rb
```diff
@@ -91,13 +89,14 @@ def all_requested_accounts
...
     creation_attempt = CreateRequestedAccount.new(requested_account, current_user)
     result = creation_attempt.result
-    return result unless result[:success]
-    # If it was successful, enroll the user in the course
-    user = creation_attempt.user
-    JoinCourse.new(course: requested_account.course,
-                   user: user,
-                   role: CoursesUsers::Roles::STUDENT_ROLE)
-    result
+    result[:result_description] = creation_attempt.result_description(result)
```
ragesoss: This is still some unnecessary work... you could set the `:result_description` from within `CreateRequestedAccount` right after the response is received from the server, so that calling `result = creation_attempt.result` will mean it already includes the description.

shashwatkathuria: Done!

ragesoss:
Thanks! I think this could be merged now, but if you'd like to polish it up a little further I have a few more suggestions.

1. The specs should be updated with this new behavior where applicable. In particular, I think new expectations could be added to the existing tests of `CreateRequestedAccount` to check that the expected description is part of the result, for both the success and failure states.
2. There room to pare down the new code in `CreateRequestedAccount` further, as there's not really a need to pass `result` as an argument`. It's already available as an instance variable, so instead of "@result[:result_description] = result_description(@result)" it could be just a call to `set_result_description`.
3. The I18n messages should take the result text as a variable rather than concatenate the success/failure plus `:` plus the result. That way it's not a "lego" message (ie, built from small pieces), and translators are free to incorprate the result however makes sense, without having to work around the placement of the colon.

shashwatkathuria:
@ragesoss Please check. If it's fine then I'll go forward with coding the tests part.

spec/services/create_requested_account_spec.rb
```diff
@@ -34,20 +34,23 @@
     stub_account_creation(wiki: home_wiki)
     allow(UserImporter).to receive(:new_from_username).and_return(user)
     expect(subject.result[:success]).not_to be_nil
+    expect(subject.result[:result_description]).not_to be_nil
```
ragesoss: This could be better. Instead of just *not nil*, it should look for some specific text that will give us a clear indication that it worked as expected, such as looking for the the string "Success:" or "Failure:" within the description.

shashwatkathuria: Done!

app/services/create_requested_account.rb
```diff
@@ -26,6 +26,19 @@ def initialize(requested_account, creator)
     process_request(@creator, @creation_reason)
   end
 
+  def set_result_description
+    if @result[:failure]
+      @result[:result_description] = @result[:failure].sub('response: {}', '')
```
ragesoss: This is still a bit confusing, because you're writing to `@result[:result_description]` twice, first with an intermediate string and then with the final i18n message.

Better to use a temporary variable here. Why are we substituting the URL? The name of the temporary variable should give some indication of the purpose of this step.

app/services/create_requested_account.rb
```diff
@@ -26,6 +26,19 @@ def initialize(requested_account, creator)
...
+                                                      .sub('https://en.wikipedia.org', '')
+      @result[:result_description] = I18n.t('users.requested_account_status.failure_message',
+                                            result_description: @result[:result_description])
+    elsif @result[:success]
+      @result[:result_description] = @result[:success]
```
ragesoss: Same as above, writing twice to `@result[:result_description]` is a little confusing. Here, you can just skip this steppass `@result[:success]` to `I18n.t` .

shashwatkathuria: Done!

app/services/create_requested_account.rb
shashwatkathuria: Done!
The url and empty response substrings are removed from the received message from the server so as to display them in user friendly format. Otherwise they would look like the failure string in the Before image of the pull request (Image in the starting).

ragesoss:
With the refactoring of `CreateRequestedAccount`, it looks like now there's no need to reorganize the control flow in the controllers. I'd rather stick with the guard statement approach (return results unless success, then do the enrollment), since it makes the code a little more linear.

shashwatkathuria:
Okay! I have changed the controllers to the way they were before (with the guard statement approach).

ragesoss:
@shashwatkathuria nice work, thanks for persevering!

Issue #1940, opened 2018-07-05
i18n-js locale files should be fingerprinted


ragesoss:
Our asset build process adds fingerprints to the compiled javascript entries, by using `webpack-manifest-plugin`. (See `gulp/tasks/webpack.js`)

However, the i18n locale files are generated a different way (via rake task, called from `gulp/tasks/i18n.js`). They are not fingerprinted, so users often have cached locale files that do not include new messages or translations.

One way or another, we should add fingerprinting to the locale files. This might happen alongside changing to the `chunkhash` option for webpack, so that we don't force users to fetch unchanged assets each new deployment.

This might be useful: https://www.npmjs.com/package/webpack-rails-i18n-js-plugin

ragesoss:
Fixed with #4042

We're not actually fingerprinting them, but we use the old strategy of including the hash as a parameter to make sure browsers pull in a new copy when it changes.

PR #3826, opened 2020-02-29
transfer i18n-js export task from gulp to webpack

pv-912:
## What this PR does
Remove the i18n-js export task from the gulp pipeline and add it to the webpack which will execute it before build start.


ragesoss:
Cool! Looks like the build is passing. I'm dubious about WebpackShellPlugin, though. It hasn't been updated in 3 years and people are forking it to fix bugs. Is there a reason this plugin in needed instead of using a built-in option for running a shell task like `child_process`?

Is there an issue for this? Does it get us closer to a solution for #1940 ?

pv-912:
> Is there a reason this plugin in needed instead of using a built-in option for running a shell task like `child_process`?

I will try to do it using `child_process`. Also, I have found a way to this on [stackoverfolw](https://stackoverflow.com/a/49786887/8111649) and will try that.

pv-912:
@ragesoss However `webpackShellPlugin` also does the same thing as we can see [here](https://github.com/1337programming/webpack-shell-plugin/blob/master/src/webpack-shell-plugin.js) like uses the `child_process` and executes it before or after the build started.

ragesoss:
If it's the same thing and can be done with a built-in node command, I'd rather avoid adding an abandoned dependency.

ragesoss:
This is now taken care of in `prepare.js`.

Issue #3500, opened 2019-11-15
Add a way to verify that students really took training modules


ragesoss:
Instructors want to be able to confirm that students read the training modules and completed the on-wiki exercises, rather than just clicking through to the end and registering with the dashboard as having completed them.

How to do this? A few possibilities:

* Show instructors the time it took between starting and completing each module (which can be calculated from the TrainingModulesUsers record's timestamps).
* Add limits to the training module progress update flow so that too-rapid advancing through slides won't count for completion (either frontend or backend limits).
* For modules with that have on-wiki edits, check for edits to the target pages and do not register modules as complete unless such edits have been made. Perhaps, don't allow progress past the slide that launches them until such edits are made.

shashwatkathuria:
@ragesoss How do I disable the frontend functionality for admin?? More specifically, how do I know that an admin is logged in (in javascript)? I have implemented the frontend functionality, with only the admin functionality remaining (disabling frontend limit for admin).

fillingtothemomo:
@ragesoss I would like to work on this issue, what deliverables did the previous PR fail to address that I should keep in mind?

ragesoss:
@fillingtothemomo the last bullet in the issue description is the remaining thing we could do. It's a fairly complicated feature to implement, though.

prathamVaidya:
@ragesoss Is this slide an example of on-wiki edits : [https://outreachdashboard.wmflabs.org/training/editing-wikipedia/editing-basics/about-sandboxes](https://outreachdashboard.wmflabs.org/training/editing-wikipedia/editing-basics/about-sandboxes)

![image](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/61202986/2d4d2f34-076c-4077-99fc-39a67db8ee70)


The button redirects user to Special sandbox page (https://en.wikipedia.org/wiki/Special:MyPage/sandbox)[https://en.wikipedia.org/wiki/Special:MyPage/sandbox] 

we can track if user has added an edit by using the wiki api to check last edit timestamp of 
https://en.wikipedia.org/wiki/User:<UserName>/sandbox 

For example : https://en.wikipedia.org/wiki/User:Prathamistesting/sandbox

Is this approach correct?

ragesoss:
@prathamVaidya I think for this particular example, we don't want to do anything with the generic `/sandbox` page, but for others like `/be_bold`, yes, that I think that is the right kind of thing.

PR #3828, opened 2020-02-29
Verify usage/completion of training modules by students (enhancement) #3500

shashwatkathuria:
## What this PR does
Solves #3500. Adds time fields to relevant data json fields which are rendered in the training module status of the user.

## Screenshots
Before:
Screenshot from 2020-03-01 03-11-58
Screenshot from 2020-03-01 03-08-56
Screenshot from 2020-03-01 03-09-13

After:
Screenshot from 2020-03-03 02-20-09
Screenshot from 2020-03-03 02-20-30
Screenshot from 2020-03-03 02-20-32

@ragesoss I am going to add more features you mentioned in #3500 in this same PR.
Following are my concerns and plans : 
1. I am planning on adding frontend button disabling for certain seconds before 'Next Page' can be clicked.
2. Can you please guide me how to check for edits to the target pages for modules with on-wiki edits ?? I want to implement this also if possible.


app/views/training_status/show.json.jbuilder
```diff
@@ -9,5 +9,13 @@ json.course do
     json.kind training_module.kind
     json.status training_progress_manager.status
     json.completion_date training_progress_manager.completion_date
+    tmu = TrainingModulesUsers.find_by(user_id: @user.id, training_module_id: training_module.id)
+    if training_progress_manager.completion_date
+      json.completion_time Time.at(tmu.completed_at - tmu.created_at)
+                               .utc
+                               .strftime('%H ' + I18n.t('users.training_module_time_field.hours')\
+                               + ' %M ' + I18n.t('users.training_module_time_field.minutes')\
+                               + ' %S ' + I18n.t('users.training_module_time_field.seconds'))
```
ragesoss: This should probably be handled on the frontend. If `completion_time` is passed as a number (in seconds), then we can use `moment` to translate that into a human-friendly representation.

Doing it as three parts like this is going to lead to i18n problems with some languages, as the relationship between the hours, minutes and seconds strings is hard-coded (so right-to-left languages, and possibly others, will probably not get the right formatting).

app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
```diff
@@ -13,7 +13,9 @@ export const TrainingModuleRows = ({ trainingModules }) => {
     if (trainingModule.completion_date) {
       moduleStatus = (
         <span className="completed">
-          Completed at {moment(trainingModule.completion_date).format('YYYY-MM-DD   h:mm A')}
+          Completed at {moment(trainingModule.completion_date).format('YYYY-MM-DD   h:mm A')}.
+          <br/>
+          Time taken to complete {trainingModule.completion_time}.
```
ragesoss: This should be done with `i18n`. (So should the "Completed at" part that is already present.)

Maybe "Completion time:" instead of "Time taken to complete", just to be a like more concise.

ragesoss:
Thanks! I think disabling the buttons without a minimum time should probably be done in a different PR. One consideration there is that the rate-limit should be disabled for an admin, and also shouldn't stop a user from quickly moving through slides they've already visited... only if it's the furthest slide they've reached should the limit affect them.

As for checking for on-wiki edits, that's moderately complicated. Here's one example:
https://dashboard.wikiedu.org/training/students/choose-topic-exercise/present-articles

That exercise instructs the student to follow a link to a user sandbox page like "User:Ragesoss/Choose an Article". So to find out whether `Ragesoss` has made that edit, you would look for a `Revision` by Ragesoss made to an `Article` in the `USER` namespace titled "Ragesoss/Choose an Article".  Combining that smoothly with the user experience is a big project, though.

shashwatkathuria:
Updated After Screenshots:
Screenshot from 2020-03-03 02-20-09
Screenshot from 2020-03-03 02-20-30
Screenshot from 2020-03-03 02-20-32

shashwatkathuria:
Okay @ragesoss , I'll come up with the frontend part soon in a different PR. I'll also try to look into on-wiki edits.

app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
shashwatkathuria: Done! Please have a look over it.

app/views/training_status/show.json.jbuilder
shashwatkathuria: Done! Please have a look over it.

shashwatkathuria:
Please ignore this PR.

PR #3837, opened 2020-03-02
Verify training modules usage enhancement#3500

### LINKED ISSUE #3500: Add a way to verify that students really took training modules

shashwatkathuria:
## What this PR does
Fixes #3500. Adds time fields to relevant data json fields which are rendered in the training module status of the user.

## Screenshots
Before:
Screenshot from 2020-03-25 00-59-11
Screenshot from 2020-03-01 03-08-56
Screenshot from 2020-03-01 03-09-13

After:
Screenshot from 2020-03-25 00-57-59
Screenshot from 2020-03-03 02-20-30
Screenshot from 2020-03-03 02-20-32

## Note

Continuation of #3828 . Some bad rebasing occurred in 3828 branch, new branch with required changes created.

### DIFF
--- a/app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
+++ b/app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
@@ -13,7 +13,9 @@ export const TrainingModuleRows = ({ trainingModules }) => {
     if (trainingModule.completion_date) {
       moduleStatus = (
         <span className="completed">
-          Completed at {moment(trainingModule.completion_date).format('YYYY-MM-DD   h:mm A')}
+          {I18n.t('training_status.completed_at')}: {moment(trainingModule.completion_date).format('YYYY-MM-DD   h:mm A')}
+          <br/>
+          {I18n.t('training_status.completion_time')}: {moment.utc(trainingModule.completion_time * 1000).format('HH:mm:ss')}
         </span>
       );
     } else {
--- a/app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
+++ b/app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
@@ -11,17 +11,15 @@ export const TrainingModuleRows = ({ trainingModules }) => {
   return trainings.map((trainingModule) => {
     let moduleStatus;
     if (trainingModule.completion_date) {
-      let completionTime;
+      let completionTime = '';
       if (trainingModule.completion_time <= 60 * 60) {
-        completionTime = moment.utc(trainingModule.completion_time * 1000).format('HH:mm:ss');
-      } else {
-        completionTime = I18n.t('training_status.sufficient_time_allocation');
+        completionTime = `${I18n.t('training_status.completion_time')}: ${moment.utc(trainingModule.completion_time * 1000).format('HH:mm:ss')}`;
       }
       moduleStatus = (
         <span className="completed">
           {I18n.t('training_status.completed_at')}: {moment(trainingModule.completion_date).format('YYYY-MM-DD   h:mm A')}
           <br/>
-          {I18n.t('training_status.completion_time')}: {completionTime}
+          {completionTime}
         </span>
       );
     } else {
--- a/app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
+++ b/app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
@@ -11,11 +11,17 @@ export const TrainingModuleRows = ({ trainingModules }) => {
   return trainings.map((trainingModule) => {
     let moduleStatus;
     if (trainingModule.completion_date) {
+      let completionTime;
+      if (trainingModule.completion_time <= 60 * 60) {
+        completionTime = moment.utc(trainingModule.completion_time * 1000).format('HH:mm:ss');
+      } else {
+        completionTime = I18n.t('training_status.sufficient_time_allocation');
+      }
       moduleStatus = (
         <span className="completed">
           {I18n.t('training_status.completed_at')}: {moment(trainingModule.completion_date).format('YYYY-MM-DD   h:mm A')}
           <br/>
-          {I18n.t('training_status.completion_time')}: {moment.utc(trainingModule.completion_time * 1000).format('HH:mm:ss')}
+          {I18n.t('training_status.completion_time')}: {completionTime}
         </span>
       );
     } else {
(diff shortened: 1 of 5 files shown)


shashwatkathuria:
@ragesoss Please have a look. This is a continuation of a previous pull request I closed due to a bad git rebase.

app/views/training_status/show.json.jbuilder
```diff
@@ -2,12 +2,14 @@
...
+    tp_manager = TrainingProgressManager.new(@user, training_module)
 
     json.id training_module.id
     json.module_name training_module.name
     json.kind training_module.kind
-    json.status training_progress_manager.status
-    json.completion_date training_progress_manager.completion_date
+    json.status tp_manager.status
+    json.completion_date tp_manager.completion_date
+    tmu = TrainingModulesUsers.find_by(user_id: @user.id, training_module_id: training_module.id)
```
ragesoss: This is probably an extra query that doesn't need to happen; the `TrainingProgressManager` already finds the `TrainingModulesUsers` when it is initialized. You could add a `completion_time` method to `TrainingProgressManager`.

app/views/training_status/show.json.jbuilder
```diff
@@ -2,12 +2,14 @@
 
 json.course do
   json.training_modules @assigned_training_modules do |training_module|
-    training_progress_manager = TrainingProgressManager.new(@user, training_module)
+    tp_manager = TrainingProgressManager.new(@user, training_module)
```
ragesoss: I guess this was just to satify rubocop with the long `completion_time` line? That shouldn't be a problem if the calculation gets moved to a `TrainingProgressManager` instance method.

app/views/training_status/user.json.jbuilder
```diff
@@ -10,6 +10,7 @@ json.user do
       tp_manager = TrainingProgressManager.new(@user, training_module, training_module_user: tmu)
       json.status tp_manager.status
       json.completion_date tp_manager.completion_date
+      json.completion_time tmu.completed_at - tmu.created_at if tp_manager.completion_date
```
ragesoss: Same thing here: use TrainingProgressManager for the calculation so this doesn't need to be duplicated.

app/views/training_status/show.json.jbuilder
shashwatkathuria: Yes. Okay I'll move it there.

app/views/training_status/user.json.jbuilder
shashwatkathuria: Okay!

app/views/training_status/show.json.jbuilder
shashwatkathuria: Okay!

shashwatkathuria:
@ragesoss Please have a look

lib/training_progress_manager.rb
```diff
@@ -74,6 +74,10 @@ def module_progress
     module_completed? ? I18n.t('training_status.completed') : completing
   end
 
+  def completion_time
+    return @tmu.completed_at - @tmu.created_at if completion_date
```
ragesoss: This will be slightly clearer if you use `module_completed?`

I recommend doing it in two lines with a guard clause and then the calculation. That makes it a little more clear what the possible return values are (nil, or a duration).

app/views/training_status/show.json.jbuilder
ragesoss: Please change it back to `training_progress_manager` as well.

lib/training_progress_manager.rb
shashwatkathuria: Done! @ragesoss Please have a look.

app/views/training_status/show.json.jbuilder
ragesoss: @shashwatkathuria ping

shashwatkathuria: Done! @ragesoss

ragesoss:
This doesn't handle situations where there's a long time between starting a finishing a training module very well. Times greater than one day are left out entirely (so starting a 1pm on one day and finishing a 1:01pm the next day will show that it took just 1 minute to complete).

I think maybe it should only show any duration at all if the total time is less than an hour. It's really only relevant for cases where the time it took to complete the module is very short; instructors will want to know that the student probably did not spend the time to actually read through the module.

shashwatkathuria:
@ragesoss I have made the changes. Please check.

app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
```diff
@@ -11,9 +11,17 @@ export const TrainingModuleRows = ({ trainingModules }) => {
   return trainings.map((trainingModule) => {
     let moduleStatus;
     if (trainingModule.completion_date) {
+      let completionTime;
+      if (trainingModule.completion_time <= 60 * 60) {
+        completionTime = moment.utc(trainingModule.completion_time * 1000).format('HH:mm:ss');
+      } else {
+        completionTime = I18n.t('training_status.sufficient_time_allocation');
```
ragesoss: I think it should not show this message. Taking a long time between starting and finishing doesn't necessarily mean any significant time was spent on each slide; it could be that they just started it, left it, and then rushed through it later. If the total completion time is very short, that's a meaningful bit of data that we want to show, but if it's long, it's not a clear enough signal to draw any conclusions from.

shashwatkathuria: @ragesoss Okay! I just removed the message if it is more than one hour.

config/locales/en.yml
```diff
@@ -1364,6 +1364,9 @@ en:
   training_status:
     completed: "Completed"
     continue: "Continue"
+    completed_at: "Completed at"
+    completion_time: "Completion time"
+    sufficient_time_allocation: "Sufficient time allocated"
```
ragesoss: No need for this now.

shashwatkathuria: Okay, removed

ragesoss:
Trying this out again, I think it's not clear enough how to read the completion time. It's not going to be obvious to some people that it's supposed to mean Hours:Minutes:Seconds (as opposed to Days:Hours:Minutes or something else).

Is there another `moment` format that we can use that will make this clearer?

shashwatkathuria:
@ragesoss I can write it like 02 Hours 10 Minutes 30 Seconds if you like..Or 02 Hr 10 Min 30 Sec

ragesoss:
Hr Min Sec is good, I think.

shashwatkathuria:
Should I write I18n strings for these short forms?

ragesoss:
If moment doesn't have that option built-in, then yes.

shashwatkathuria:
@ragesoss Please check. I removed Hours because anyways its always 00.

PR #3852, opened 2020-03-09
[WIP] Feature/bad work flag

bwreid:
## What this PR does

This PR implements a feature for individuals to mark a particular article as having "bad work." This notification will then go to a Wikipedia Expert to address with the instructor and/or related students.


ragesoss:
I think this is pretty much ready to go, and we can deploy it today. I think all the TODO items have been addressed now; there's a thank-you message after submission, only advanced roles will see the button, and it creates an unassigned ticket if there's no Wiki Expert to assign it to.

I found a bug testing it out this morning: if there's a slow response from the server to clicking the 'Notify Wiki Expert' button, the user may click it a second time before the 'Thank You' appears, and this will create a duplicate ticket. The button should go into a disabled state after the first click while the request is resolving.

Issue #3851, opened 2020-03-09
Add the number of results


rajat2502:
> In the Search field on the explore page, the result can also have the feature of the total number of results related to that particular search.

### Current Behavior:
> Currently, It only shows "Results for ${SeachItem}"

### Desired Behavior:
>Add the Total number of results found to it.
It should be like "Total ${number} Results for ${SearchItem}.

### Additional context:
> 
![image](https://user-images.githubusercontent.com/42200276/76242699-24606880-625d-11ea-9b55-ae78818b229d.png)

ClintonEnyinna:
I'm an Outreachy applicant and I'm interested in working on this issue. Please can it be assigned to me?

ragesoss:
Hi. We don't formally assign issues for this project, but feel free to work on this. You should work on one issue at a time, and leave a note if you stop working on it.

ClintonEnyinna:
I created a pull request for the issue

YanaNirvana:
Sorry if this sounds obvious:  The merged PR looks like it resolves this issue... is there a reason for that this is still open?  / Are there additional work to be done?

ragesoss:
Nope, the automatic closing just didn't trigger when the PR closed.

PR #3855, opened 2020-03-11
Add the number of results #3851

ClintonEnyinna:
#3855  What this PR does
It adds the total number of searches found and pluralizes the word 'Result' accordingly (1 Result, 2 Results).

## Screenshots
Before:
Result for: 'js'
Screenshot from 2020-03-10 22-43-06

After:
1 Result for: 'Art'
Screenshot from 2020-03-11 00-04-59

Any feedback is highly appreciated

### DIFF
--- a/My Survey-results2020-03-11.csv
+++ b/My Survey-results2020-03-11.csv
@@ -1,0 +1,1 @@
+username,course,Q7: Long Text Question
--- a/My Survey-results2020-03-11.csv
+++ b/My Survey-results2020-03-11.csv
@@ -1,1 +1,0 @@
-username,course,Q7: Long Text Question
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -106,6 +106,10 @@ class ApplicationController < ActionController::Base
     current_user&.admin?
   end
 
+  def count_and_pluralize(results)
+    "#{results.count} #{(results.count > 1 ? 'Results' : 'Result')}"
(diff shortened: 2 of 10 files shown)


My Survey-results2020-03-11.csv
```diff
@@ -0,0 +1 @@
+username,course,Q7: Long Text Question
```
ragesoss: This and the similar files below are temporary files generated by tests and should not be included.

app/controllers/application_controller.rb
```diff
@@ -106,6 +106,10 @@ def can_administer?
     current_user&.admin?
   end
 
+  def count_and_pluralize(results)
```
ragesoss: Pluralization should be handled by the I18n system rather than a custom method. See, for example, the `requested_accounts_alert` messages in `en.yml` and how they are used.

app/controllers/campaigns_controller.rb
```diff
@@ -106,6 +108,10 @@ def programs
     set_page
     set_presenter
     @search_terms = params[:courses_query]
+    if @search_terms.present?
+      @results = @presenter.search_courses(@search_terms)
+      @num_of_results = @results.count
```
ragesoss: I don't think having a separate instance variable for the count is necessary. You can just use `results.count` or similar in the views.

ragesoss:
@ClintonEnyinna thanks!

PR #3865, opened 2020-03-12
Refactoring and improving spec coverage for Ruby code

ragesoss:
This fixes all the line coverage gaps in the rspec test suite, except for the PetScan API which has an in-progress PR. It also includes a small refactor of Assignment and its relationship to AssignmentPipeline, so that the latter is isolated as an implementation detail of Assignment.


app/models/course_data/assignment.rb
```diff
@@ -40,6 +40,10 @@ class Assignment < ApplicationRecord
 
   serialize :flags, Hash
 
+  delegate :status, to: :assignment_pipeline
+  delegate :update_status, to: :assignment_pipeline
+  delegate :all_statuses, to: :assignment_pipeline
```
bwreid: I didn't know this was a thing. This is great!

PR #3867, opened 2020-03-16
Counting only those revisions (course revision count) tracked by course.

### LINKED ISSUE #3647: Course revision count should only include revisions from tracked wikis
Currently, if there are other edits by the same user on an untracked wiki (typically, because they are being tracked for a different course), those revisions are counted in the total, even though only tracked wikis are counted for the other stats.

shashwatkathuria:
## What this PR does
Fixes #3647 

### Before:
Screenshot from 2020-03-16 14-39-02
Screenshot from 2020-03-16 14-39-22
Screenshot from 2020-03-16 14-39-42

### After:
Screenshot from 2020-03-16 14-27-44
Screenshot from 2020-03-16 14-26-53
Screenshot from 2020-03-16 14-27-17

@ragesoss Do I be have to make the same changes to the revision count of Course + User join model at app/models/course_data/courses_users.rb:109 or leave it as it is?

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -265,7 +265,7 @@ class Course < ApplicationRecord
 
   def tracked_revisions
     revisions.where.not(article_id: articles_courses.not_tracked.pluck(:article_id))
-             .where('revisions.wiki_id in (?)', wiki_ids)
+             .where(wiki_id: wiki_ids)
   end
 
   def tracked_articles
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -265,6 +265,7 @@ class Course < ApplicationRecord
 
   def tracked_revisions
     revisions.where.not(article_id: articles_courses.not_tracked.pluck(:article_id))
+             .where('revisions.wiki_id in (?)', wiki_ids)
   end
 
   def tracked_articles
--- a/spec/models/tracked_revisions_spec.rb
+++ b/spec/models/tracked_revisions_spec.rb
@@ -5,22 +5,34 @@ require_dependency "#{Rails.root}/lib/analytics/course_statistics.rb"
 
 describe 'Course#tracked_revisions' do
   let(:course) { create(:course, start: '10-07-2019'.to_date, end: '10-08-2019'.to_date) }
+  let(:different_wiki_course) do
+    create(:course,
+           title: 'Science',
+           school: 'WinterSchool',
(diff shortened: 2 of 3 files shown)


ragesoss:
@shashwatkathuria it looks like the code on CoursesUsers line 109 is coming from `live_revisions` which in turn comes from `Course#tracked_revisions`, so if you make the change to the `Course` method rather than the one in `CourseCacheManager`, it should cover both cases.

shashwatkathuria:
@ragesoss I knew that but didn't really do that, because tracked_revisions was being used at many places and that might break the code other than the course revision count part. Anyways, now that you have indicated, I will go forward with it.

app/models/course.rb
```diff
@@ -265,6 +265,7 @@ def training_module_ids
 
   def tracked_revisions
     revisions.where.not(article_id: articles_courses.not_tracked.pluck(:article_id))
+             .where('revisions.wiki_id in (?)', wiki_ids)
```
ragesoss: This should be done the more Rails-y way, `.where(wiki_id: wiki_ids)`

shashwatkathuria: Done!

ragesoss:
Cool. This looks like it will work, and it doesn't break any existing tests (yay!).

The PR should include new test(s) to exercise the change (and let us catch any later changes that would undo the fix).

spec/models/tracked_revisions_spec.rb
```diff
@@ -5,22 +5,34 @@
...
     VCR.use_cassette 'course_upload_importer/Hbultra' do
       UpdateCourseStats.new(course)
+      UpdateCourseStats.new(different_wiki_course)
     end
 
     # Check if the data is cached
     old_character_sum = course.character_sum
     old_course = course.dup
     expect(course.revision_count).to eq(old_course.revision_count)
+    expect(different_wiki_course.revision_count).to eq(0)
```
ragesoss: Does this test fail if you undo the change to `course.rb`?

shashwatkathuria: Yes it gives the following failure error
Expected :0
Got: 41

shashwatkathuria: Screenshot from 2020-03-17 22-40-50
Screenshot from 2020-03-17 22-41-04

PR #3869, opened 2020-03-16
Course specific activity information#3847

### LINKED ISSUE #3847: Activity information not course specific
### What is happening?
The 'Activity' Tab for a specific course gives information about user's whole activity rather than only course specific activity.

### To Reproduce

   1. Make 2 new basic programs as an instructor. Make a student join both of them.
   2. Assign the student one article in each course.
   3. Make edits to both articles and also do manual update (/manual_update) for pulling the revisions.
   4. See 'Activity' tab for each course. 
   5. See error

### Expected behavior
Only the content related to the specific course should be shown in the 'Activity' Tab, because the Activity Tab is of the specific course.

### Screenshots
Screenshot from 2020-03-06 02-04-08
Screenshot from 2020-03-06 02-04-16
Screenshot from 2020-03-06 02-04-21
Screenshot from 2020-03-06 02-04-26

### System Information

 - OS: Linux Ubuntu 18.04
 - Browser Chrome
 - Latest Version

shashwatkathuria:
## What this PR does
Fixes #3847 

## Screenshots
Peek 2020-03-18 16-30

### DIFF
--- a/app/assets/javascripts/components/revisions/revisions_handler.jsx
+++ b/app/assets/javascripts/components/revisions/revisions_handler.jsx
@@ -27,7 +27,7 @@ const RevisionHandler = createReactClass({
 
   UNSAFE_componentWillMount() {
     if (this.props.loadingRevisions) {
-      this.props.fetchRevisions(this.props.course_id, this.props.limit, this.state.courseSpecificBool);
+      this.props.fetchRevisions(this.props.course_id, this.props.limit);
     }
   },
 
@@ -36,11 +36,10 @@ const RevisionHandler = createReactClass({
   },
 
   showMore() {
-    return this.props.fetchRevisions(this.props.course_id, this.props.limit + 100, this.state.courseSpecificBool);
+    return this.props.fetchRevisions(this.props.course_id, this.props.limit + 100);
   },
 
   toggleCourseSpecificBool() {
-    this.props.fetchRevisions(this.props.course_id, this.props.limit, !this.state.courseSpecificBool);
     this.setState({
       courseSpecificBool: !this.state.courseSpecificBool,
     });
@@ -76,7 +75,7 @@ const RevisionHandler = createReactClass({
           </div>
         </div>
         <RevisionList
-          revisions={this.props.revisions}
+          revisions={(this.state.courseSpecificBool) ? this.props.revisions.filter((revision) => { return revision.course_specific; }) : this.props.revisions}
           course={this.props.course}
           sortBy={this.props.sortRevisions}
           wikidataLabels={this.props.wikidataLabels}
--- a/app/assets/javascripts/components/revisions/revisions_handler.jsx
+++ b/app/assets/javascripts/components/revisions/revisions_handler.jsx
@@ -15,6 +15,7 @@ const RevisionHandler = createReactClass({
     limitReached: PropTypes.bool,
     limit: PropTypes.number,
     revisions: PropTypes.array,
+    course_specific_revision_ids: PropTypes.array,
     wikidataLabels: PropTypes.object,
     loadingRevisions: PropTypes.bool
   },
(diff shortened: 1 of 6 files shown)


ragesoss:
This is not the approach to implementation that I would prefer; refetching for each toggle and doing the determination of which revisions are in-scope server-side is going to be slower than doing the filtering client-side (probably via a selector, following the pattern of other places where we do similar things with data that lives in the redux store).

If there's an efficient query for it, one option might be to include a property for each revision of whether it's course-specific or not. Then the filtering client-side would be very simple. Otherwise, the more complicated option for filtering client-side would be to use the data in the `articles` store as the basis for selecting.

shashwatkathuria:
@ragesoss I have shifted the filtering to client side and now need some more clarification regarding the Course type(as indicated by you in the issue comments), i.e, which type of revisions are course specific and for what articles. 

I want to know how should I handle the articles whose activity counts as course specific for each Course type (because that might be different for every course type). Should I do it with the articles in article.json endpoint only? Or should I work with the articles reducer at client side instead of server side filtering (if that could cover all course types)?

ragesoss:
I'm not sure, but my guess is that it's possible to adjust the revisions endpoint so that it includes the `course_specific` property for eafch revision, without major performance implications. That route will make client-side filtering very easy, without having to worry about the articles reducer.

shashwatkathuria:
@ragesoss I have done that only in the commits after review, what I need to clarify is what are the articles/assignments/etc that are course specific for different course types and how should I deal with them.

ragesoss:
@shashwatkathuria the course `tracked_revisions` scope should be the root source of which revisions are course specific and which are not.

shashwatkathuria:
The articles for the following are at : http://localhost:3000/courses/QCA/Brisbane_QCA_ArtandFeminism_2018/articles/edited

Peek 2020-03-18 16-30

app/views/courses/revisions.json.jbuilder
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 json.course do
+  course_specific_revision_ids = @course.tracked_revisions.pluck(:id)
```
ragesoss: This could be a significant performance problem for a course with huge numbers of revisions. Is there a way to figure this out as part of the revisions query on the following line, putting that work onto the database instead of Ruby?

shashwatkathuria: I'll try to find a way and then reply back!

shashwatkathuria: @ragesoss 
I tried to experiment what could be done regarding this, so I thought of using the Rails - and & operator for the intersection and difference of the two queries(all revisions and tracked_revisions), but that would make the queries complex and I also couldn't make sense of how they could possibly help us in improving in the performance problem, so what I now did was send all the course_specific_revision_ids as data in the json sent and the filter out everything on the client side itself, and also limited the number of tracked_revisions (which will be atmost the variable limit, because atmost every revision will be a tracked revision).

app/assets/javascripts/components/revisions/revisions_handler.jsx
```diff
@@ -54,7 +76,7 @@ const RevisionHandler = createReactClass({
           </div>
         </div>
         <RevisionList
-          revisions={this.props.revisions}
+          revisions={(this.state.courseSpecificBool) ? this.props.revisions.filter((revision) => { return this.props.course_specific_revision_ids.includes(revision.id); }) : this.props.revisions}
```
ragesoss, line 79: I don't like doing the filtering here. This should probably be implemented as a `selector`; see `ArticlesHandler` for an example, where the state of the currently selected filters is used to do the filtering in `mapStateToProps`.

In addition to following the pattern of filtering elsewhere in the system, this approach also provides automatic caching of the filtered results; otherwise, every time one of the props for this component changes (which may happen a lot with wikidataLabels, in particular, as they get fetched after the rest of the data has been received) that `filter()` call will run again. Probably not a huge performance issue, but it might matter.

shashwatkathuria: Done in new PR #4002

PR #3871, opened 2020-03-17
Switch to 'Words Added' for Students and Articles

ragesoss:
We convert characters/bytes to words to give course-level estimates of content added, based on a value estimate for English Wikipedia. We want to do this also at the level of individual articles and editors as well, but without disrupting stats for other wikis where we don't have an estimate of bytes per word.

This patch addresses the issue by adding a `bytes_per_word` method for Wiki, which will return nil for any wiki that we don't have an estimate for (ie, everything but en.wikipedia.org). The strategy here is to pass that value for the home wiki of the course to the frontend, and then switch between words and bytes based whether a `home_wiki_bytes_per_word` value is received. This way, it should be easy to extend support for other wikis if/when we have bytes-per-word estimates for them.

### Before
chars added

chars per student

### After
words added

words per student

### DIFF
--- a/app/assets/javascripts/components/students/shared/StudentList/Student/Student.jsx
+++ b/app/assets/javascripts/components/students/shared/StudentList/Student/Student.jsx
@@ -131,6 +131,24 @@ const Student = createReactClass({
       );
     }
 
+    let contentAdded;
+    if (course.home_wiki_bytes_per_word) {
+      const mainspaceWords = Math.round(student.character_sum_ms / course.home_wiki_bytes_per_word);
+      const userspaceWords = Math.round(student.character_sum_us / course.home_wiki_bytes_per_word);
+      const draftWords = Math.round(student.character_sum_draft / course.home_wiki_bytes_per_word);
+      contentAdded = (
+        <td className="desktop-only-tc">
+          {mainspaceWords} | {userspaceWords} | {draftWords}
+        </td>
+      );
+    } else {
+      contentAdded = (
+        <td className="desktop-only-tc">
+          {student.character_sum_ms} | {student.character_sum_us} | {student.character_sum_draft}
+        </td>
+      );
+    }
+
     const uploadsLink = `/courses/${course.slug}/uploads`;
 
     return (
@@ -163,9 +181,7 @@ const Student = createReactClass({
           {reviewButton}
         </td>
         {recentRevisions}
-        <td className="desktop-only-tc">
-          {student.character_sum_ms} | {student.character_sum_us} | {student.character_sum_draft}
-        </td>
+        {contentAdded}
         <td className="desktop-only-tc">
           {student.references_count}
         </td>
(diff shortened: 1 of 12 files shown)


lib/word_count.rb
```diff
@@ -11,9 +11,12 @@ class WordCount
...
-  CHARACTERS_PER_WORD = 5.175
+  HALFAK_EN_WIKI_ESTIMATE = 5.175
+  BYTES_PER_WORD = {
+    'en.wikipedia.org' => HALFAK_EN_WIKI_ESTIMATE
+  }.freeze
 
   def self.from_characters(characters)
-    (characters / CHARACTERS_PER_WORD).to_i
+    (characters / HALFAK_EN_WIKI_ESTIMATE).to_i
   end
```
bwreid: Would it make sense to perform this operation on the front end instead of the back end? This is strictly a display concern, right?

app/assets/javascripts/components/students/shared/StudentList/StudentList.jsx
```diff
@@ -78,12 +95,7 @@ export const StudentList = (props) => {
       sortable: true,
       info_key: 'users.revisions_doc'
     },
-    character_sum_ms: {
-      label: I18n.t('users.chars_added'),
-      desktop_only: true,
-      sortable: true,
-      info_key: 'users.character_doc'
-    },
+    character_sum_ms: contentAddedKey,
```
bwreid: I think this ends up being a bit confusing given that all of these declarations are all over the case. Maybe it makes sense to make a separate `.js` file that just includes the keys, or to somehow otherwise "chunk" the relevant key code together.

app/assets/javascripts/components/students/shared/StudentList/Student/Student.jsx
```diff
@@ -131,6 +131,24 @@ const Student = createReactClass({
...
+          {mainspaceWords} | {userspaceWords} | {draftWords}
+        </td>
+      );
+    } else {
+      contentAdded = (
+        <td className="desktop-only-tc">
+          {student.character_sum_ms} | {student.character_sum_us} | {student.character_sum_draft}
+        </td>
+      );
+    }
```
bwreid: This could be a separate component!  :)

app/assets/javascripts/components/students/shared/StudentList/Student/Student.jsx
```diff
@@ -131,6 +131,24 @@ const Student = createReactClass({
       );
     }
 
+    let contentAdded;
+    if (course.home_wiki_bytes_per_word) {
+      const mainspaceWords = Math.round(student.character_sum_ms / course.home_wiki_bytes_per_word);
+      const userspaceWords = Math.round(student.character_sum_us / course.home_wiki_bytes_per_word);
+      const draftWords = Math.round(student.character_sum_draft / course.home_wiki_bytes_per_word);
```
bwreid: Maybe we can make a utility class / file that encapsulates all of this math? That way it can be used elsewhere more easily.

lib/word_count.rb
ragesoss: Yes, but some of the places that include this are server-rendered.

ragesoss:
@bwreid how's this?

PR #3875, opened 2020-03-18
Deleting all open requested accounts on course deletion.

### LINKED ISSUE #3873: Requested Accounts page returns 500 error if a course with a requested account got deleted
This cropped up on P&E Dashboard:

```
NoMethodError: undefined method `title' for nil:NilClass
  from app/controllers/requested_accounts_controller.rb:88:in `block in all_requested_accounts'
```

Presumably, a course for an open account request got deleted.

shashwatkathuria:
## What this PR does
Fixes #3873 

This PR fixes the error by deleting all open requested accounts as they are linked to a course which is itself deleted.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -72,10 +72,7 @@ class Course < ApplicationRecord
   has_many :staff, -> { where('courses_users.role = 4') },
            through: :courses_users, source: :user
   has_many :survey_notifications, dependent: :destroy
-  has_many :requested_accounts,
-           class_name: 'RequestedAccount',
-           foreign_key: 'course_id',
-           dependent: :destroy
+  has_many :requested_accounts, dependent: :destroy
   has_many :tickets,
            class_name: 'TicketDispenser::Ticket',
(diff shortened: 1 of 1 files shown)


app/models/course.rb
```diff
@@ -72,7 +72,10 @@ class Course < ApplicationRecord
   has_many :staff, -> { where('courses_users.role = 4') },
            through: :courses_users, source: :user
   has_many :survey_notifications, dependent: :destroy
-  has_many :requested_accounts
+  has_many :requested_accounts,
+           class_name: 'RequestedAccount',
+           foreign_key: 'course_id',
```
ragesoss: I don't think the `foreign_key` and `class_name` specifications are necessary, as these classes follow the default Rails naming conventions. It's only necessary for cases like `tickets` below, where the class name and/or foreign key column name are different from the defaults that Rails infers.

shashwatkathuria: Oops! Sorry for that :)

PR #3879, opened 2020-03-23
Bundle size analysis

AmitJoki:
Added a webpack config file and modified existing code to be able to generate static bundle analysis from various tools as listed in https://webpack.js.org/guides/code-splitting/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ Session.vim
 1
 .byebug_history
 .ruby-gemset
+stats.json
--- a/docs/tools.md
+++ b/docs/tools.md
@@ -7,3 +7,18 @@ You'll need access to a Sentry server to use this functionality; add the Sentry 
 
 #### Analytics: [Piwik](https://github.com/piwik/piwik)
 You need access to a Piwik server to use this functionality; add the url and project id to `config/piwik.yml`.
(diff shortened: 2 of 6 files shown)


ragesoss:
What does the output look like, and how do we generate it? (The latter question should be addressed by finding a relevant place in the docs to add details.)

AmitJoki:
@ragesoss I've updated the documentation in the `tools.md` Let me know if that's alright.

ragesoss:
It looks like this bundle analysis config file mirrors a lot of the config code from various gulp/webpack files. Does that mean it's pretty coupled to the build config and will need to be updated a lot as the build config changes?

AmitJoki:
@ragesoss yes. Earlier there was a separate `webpack.config.js` but I guess that's no longer the case. I'll make it so that the gulp/webpack imports the one in the root directory.

AmitJoki:
@ragesoss I have decoupled the config file and the webpack task file. Now, any changes that needs to be done can be made in the `webpack.config.js`

ragesoss:
LGTM. Any concerns about this, @bwreid?

bwreid:
This looks great, and really neat! Thanks for this.  :)

PR #3881, opened 2020-03-23
Exercise completion course specific#3680

### LINKED ISSUE #3680: Marking an exercise as complete isn't course-specific
Unlike with completing a training module, if a user completes an exercise during one course and then takes another course that also uses that exercise, we should not use the completion of the exercise from the earlier course to count... exercise completion, unlike training module completion, is a course-specific concept.

One possible solution would be to update how the data for marking completion gets stored on a TrainingModulesUsers record: the flags field should record completion on a course-by-course basis rather than as a singular yes/no status.

shashwatkathuria:
## What this PR does
Fixes #3680 .Marks an exercise course specific.

## Screenshots
Before:
Peek 2020-03-23 19-54

After:
Peek 2020-03-23 19-52

### DIFF
--- a/app/models/user_data/training_modules_users.rb
+++ b/app/models/user_data/training_modules_users.rb
@@ -29,8 +29,9 @@ class TrainingModulesUsers < ApplicationRecord
     training_progress_manager.slide_further_than_previous?(slide_slug, last_slide_completed)
   end
 
-  def mark_completion(value=true)
-    flags[:marked_complete] = value
+  def mark_completion(value=true, course_id=nil)
+    flags[:marked_complete] = value unless training_module.exercise?
+    flags[course_id] = { marked_complete: value }
   end
 
   private
--- a/app/models/user_data/training_modules_users.rb
+++ b/app/models/user_data/training_modules_users.rb
@@ -30,7 +30,6 @@ class TrainingModulesUsers < ApplicationRecord
   end
 
   def mark_completion(value=true, course_id=nil)
-    flags[:marked_complete] = value unless training_module.exercise?
     flags[course_id] = { marked_complete: value }
   end
 
--- a/app/views/courses/_block.json.jbuilder
+++ b/app/views/courses/_block.json.jbuilder
@@ -19,7 +19,11 @@ if block.training_modules.any?
     json.due_date due_date_manager.computed_due_date.strftime('%Y/%m/%d')
     json.overdue due_date_manager.overdue?
     json.deadline_status due_date_manager.deadline_status
-    json.flags due_date_manager.flags
+    if tm.exercise? && !due_date_manager.flags.nil?
+      json.flags due_date_manager.flags[course.id] || {}
+    else
+      json.flags due_date_manager.flags
+    end
     json.block_id block.id
   end
 end
--- a/app/views/courses/_block.json.jbuilder
+++ b/app/views/courses/_block.json.jbuilder
@@ -19,11 +19,7 @@ if block.training_modules.any?
     json.due_date due_date_manager.computed_due_date.strftime('%Y/%m/%d')
     json.overdue due_date_manager.overdue?
     json.deadline_status due_date_manager.deadline_status
-    if tm.exercise? && !due_date_manager.flags.nil?
-      json.flags due_date_manager.flags[course.id] || {}
-    else
-      json.flags due_date_manager.flags
-    end
+    json.flags due_date_manager.flags(course.id)
     json.block_id block.id
   end
 end
--- a/lib/course_training_progress_manager.rb
+++ b/lib/course_training_progress_manager.rb
@@ -111,10 +111,9 @@ class CourseTrainingProgressManager
       .where(user_id: @user.id)
       .where(training_module_id: exercise_modules_for_course)
       .count do |tmu|
-        course_flags_hash = tmu.flags[@course.id] || tmu.flags
-        if course_flags_hash[:marked_complete] == true
-          TrainingModule.find(tmu.training_module_id).exercise?
-        end
+        flags_hash = tmu.flags
+        course_flags_hash = flags_hash[@course.id] || flags_hash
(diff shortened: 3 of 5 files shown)


shashwatkathuria:
@ragesoss Please have a look. The tests are passing but the PR shows a yellow button on top.

app/views/courses/_block.json.jbuilder
```diff
@@ -19,7 +19,11 @@ if block.training_modules.any?
     json.due_date due_date_manager.computed_due_date.strftime('%Y/%m/%d')
     json.overdue due_date_manager.overdue?
     json.deadline_status due_date_manager.deadline_status
-    json.flags due_date_manager.flags
+    if tm.exercise? && !due_date_manager.flags.nil?
```
ragesoss: I don't like having this logic in the view. I think it would be better to change the due date manager `flags` method so that it takes an option course_id param.

ragesoss:
I think this makes sense, and it seems appropriate simple in implementation.

How will existing records be affected? For example, cases where there is already a record with where the exercise was marked complete (without a course id). Will those now show up as incomplete, and have to be re-marked?

app/views/courses/_block.json.jbuilder
shashwatkathuria: Okay! I will move the logic inside due date manager flags method.

shashwatkathuria:
@ragesoss For that, the line <pre>return flags_hash[course_id] || flags_hash</pre> returns flags_hash in that case, i.e., the untouched modules will stay the way they were earlier.

lib/training_module_due_date_manager.rb
```diff
@@ -47,8 +47,13 @@ def module_progress
     progress_manager.module_progress
   end
 
-  def flags
-    @tmu&.flags
+  def flags(course_id)
+    flags_hash = @tmu&.flags
+    if @training_module.exercise? && !flags_hash.nil?
```
ragesoss: Instead of `!flags_hash.nil?` you can use `flags_hash.present?`

shashwatkathuria: Done @ragesoss

ragesoss:
@bwreid can you review this if you have a chance? It makes sense to me and I'm happy with the code, but I haven't testing it out manually yet.

app/models/user_data/training_modules_users.rb
```diff
@@ -29,8 +29,9 @@ def furthest_slide?(slide_slug)
     training_progress_manager.slide_further_than_previous?(slide_slug, last_slide_completed)
   end
 
-  def mark_completion(value=true)
-    flags[:marked_complete] = value
+  def mark_completion(value=true, course_id=nil)
+    flags[:marked_complete] = value unless training_module.exercise?
+    flags[course_id] = { marked_complete: value }
```
bwreid: So, does this mean in the event of a training module (i.e. `training_module.training?`) we'll end up with both a key for `:marked_complete` and a key that represents the course id? Won't this end up being duplicate information?

ragesoss: We don't use `mark_completion` for anything but exercise modules anyway, do we?

I had noticed this before but then forgot to mention it... as far as I can tell, the line 33 code could just be left out entirely.

bwreid: If that's what we're doing, then we'll want to change the code I quoted above and will need to write some code that updates our existing training modules.

ragesoss:
Ah, good catch @bwreid. I didn't pick up that query as a critical coupling point.

app/models/user_data/training_modules_users.rb
shashwatkathuria: @bwreid 
Thanks for pointing out, I'll change the code quoted above.

Regarding updating the existing training modules:
The working of the training modules ( training_module.training? ) stays as it was earlier, but the working of exercises( training_module.exercise? ) changes. The working of the existing exercises will be determined by the earlier hash values <code>flags_hash[:marked_complete]</code> but if the users now click on the mark complete or mark incomplete button, then for those exercises the new nested course hash is created, and from then on, their working will be determined by <code>flags_hash[course_id][:marked_complete]</code>. 
I am not able understand what to update in the existing training modules?

shashwatkathuria: @ragesoss Then I'll go forward will removing that line

shashwatkathuria: @bwreid @ragesoss Please have a look at the changes.

lib/course_training_progress_manager.rb
```diff
@@ -110,9 +110,11 @@ def completed_exercise_modules_for_user_and_course
     TrainingModulesUsers
       .where(user_id: @user.id)
       .where(training_module_id: exercise_modules_for_course)
-      .where(flags: { marked_complete: true })
       .count do |tmu|
-        TrainingModule.find(tmu.training_module_id).exercise?
+        course_flags_hash = tmu.flags[@course.id] || tmu.flags
+        if course_flags_hash[:marked_complete] == true
+          TrainingModule.find(tmu.training_module_id).exercise?
+        end
```
bwreid: This does seem to solve the `:marked_complete` issue! Well done.  :)

I would ask for a couple changes to this though:

1. We should query our database as little as possible. Instead of using `.find()` twice, we should store the value of that result in a variable. We also can probably get to the training module _through_ the `tmu` variable (e.g. `tmu.training_module`

2. We should be more explicit with returning `true` and `false`. The else case for the `if` conditional is implicit as opposed to explicit. To put that another way, we should feel confident about what value is actually being returned. In this case, we're either returning the result of `.exercise?` or... whatever is stored inside of `course_flags_hash`?

Great work, overall!

shashwatkathuria: Thanks @bwreid ! I have made the changes, please have a look.

PR #3907, opened 2020-03-26
Fixed 'Edit' Link navigation from surveys.

### LINKED ISSUE #3905: Links to "Edit" a survey Question Group do different things in different places
If you're editing surveys and want to change a Question Group, there are two places you can do that from: the Question Groups list, which has an 'Edit' button for each group, and the Survey 'edit' view, which includes an 'Edit' link for each group included in that survey.

Clicking 'Edit' from the Question Groups index view opens an edit view includes both the questions that are part of the group and also the settings for the group itself (name, tags, etc).

edit question group good

The 'Edit' link from the Survey only has the questions, and is less useful. It should behave the same way as the other 'Edit' link.

question group edit bad

Also, the 'preview question group' link only appears in the view that lacks the other details.

shashwatkathuria:
## What this PR does
Fixes #3905 

### Before:
Peek 2020-03-26 20-02
### After:
Peek 2020-03-26 20-01

### DIFF
--- a/app/views/rapidfire/questions/_question_group_list.html.haml
+++ b/app/views/rapidfire/questions/_question_group_list.html.haml
@@ -4,6 +4,8 @@
     = link_to "Add New Question", new_question_group_question_path(@question_group), :class => 'button dark button--no-margin'
   - if @questions.length > 1
     .survey__admin-tools
+      = link_to "Preview Question Group", new_question_group_answer_group_path(@question_group) + '?preview'
+      %br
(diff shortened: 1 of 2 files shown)


ragesoss:
This is going to be worked on by a team applying for Rails Girls Summer of Code... sorry, I should have mentioned that when I created the issue. I'm going to close this for now, but if they don't end up completing it we can reopen this. (It's too easy for us to learn anything more about your skills anyway.)

shashwatkathuria:
@ragesoss Oops, okay :)

PR #3910, opened 2020-03-26
Make training progress queries more efficient

ragesoss:
This fixes a number of N+1 queries in the training module progress code.

ActiveRecord can handle the relationship between a TrainingModulesUsers record and the corresponding TrainingModule better than the manual caching we had in TrainingModulesUsers, which required an extra query for every TrainingModulesUsers record instead of allowing the database to load them all at once.

We can avoid individually loading each TrainingModule for a course with a separate query, when trying to sort the trainings from the exercises, and instead collect all the modules for a course and reuse them for exercises vs trainings.

For the completion data for an individual user in a course, loading the training_modules as part of the TrainingModulesUsers query means we don't have to load each module individually to see if it's an exercise or training.

### DIFF
--- a/app/models/user_data/training_modules_users.rb
+++ b/app/models/user_data/training_modules_users.rb
@@ -16,13 +16,9 @@ require_dependency "#{Rails.root}/lib/training_progress_manager"
 
 class TrainingModulesUsers < ApplicationRecord
   belongs_to :user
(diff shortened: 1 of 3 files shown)


bwreid:
Never mind, just tried it out. Looks good to me!  :)

PR #3911, opened 2020-03-26
Check that TrainingModuleUser has a created_at before checking completion

bwreid:
I think this should be the fix, but _I have questions._

1. There are a _number_ of training module records in our existing database that do not have a `:created_at` date. How did this happen?
1. I tried creating some tests for this, but was unable to do so in part because creating a TrainingModule is a bit more difficult in tests, but also because there is a database failure on creating a TrainingModule with a `nil` value for `:created_at`. What would be a way to test for this then?
1. If we ever find `created_at: nil`, should we correct it somehow? Or, update our existing database to correct it?

### DIFF
--- a/lib/training_progress_manager.rb
+++ b/lib/training_progress_manager.rb
@@ -75,7 +75,7 @@ class TrainingProgressManager
   end
 
   def completion_time
-    return unless module_completed?
+    return unless module_completed? && @tmu.created_at.present?
     @tmu.completed_at - @tmu.created_at
   end
 


ragesoss:
That... shouldn't be possible.

How many TrainingModulesUsers records do we have with this problem? When were the corresponding User records created?

If they are all old, I think we should fix the records and leave the code alone.

ragesoss:
Oh... I see...

https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/58e38cb5e25776383908fd9211efb540f65926f4

The table was there for a long while before timestamps were added, so all the early created before then didn't include them.

In that case, I guess this is probably the better solution vs. fabricating `created_at` data.

Issue #3914, opened 2020-03-27
Update MacOS setup script


swathi-0901:
### What is happening?
There is a problem in the MacOSX setup script in line 105

### To Reproduce
just run the setup.py file in mac osx.
Setup is aborting just because of this line.

### Expected behavior
Setup should not be aborted and it should continue.

### Screenshots
![Screen Shot 2020-03-27 at 2 46 10 AM](https://user-images.githubusercontent.com/43814493/77718921-19058f00-700a-11ea-8871-61f3389b64ea.png)

### System Information

 - OS: Mac OSX

ragesoss:
@swathi-0901 did the rest of the setup go smoothly once you fixed that bit of the script?

swathi-0901:
Yes, Rest of the script worked fine for me :)

swathi-0901:
> @swathi-0901 did the rest of the setup go smoothly once you fixed that bit of the script?



> @swathi-0901 did the rest of the setup go smoothly once you fixed that bit of the script?

Yes, Rest of the script worked fine for me :)

PR #3915, opened 2020-03-27
Fixes issue with Mac-setup script

swathi-0901:
## What this PR does
#3914, Corrected the syntax in the mac-setup script which was causing trouble for MacOS X


ragesoss:
Thanks! That was a recently added bit of the script that we hadn't tested on MacOS.

PR #3930, opened 2020-04-01
Revert "Merge pull request #3879 from AmitJoki/WebpackAnalysis"

bwreid:
This reverts commit a98131e43daf7ca6c9530284af26a87a95e62411, reversing
changes made to a40b8890e86937d0567a62cb1bce0b6898bdfbf2.

@AmitJoki, I encountered some problems that were solved by reverting this commit. In particular, working on the front end became more difficult. The build would not load quickly. Instead, I would need to stop the `gulp build` process and restart it.

This seems like a great addition, so please try fixing this up so we can add it back!


AmitJoki:
@bwreid um, if you see the changes, it is nothing more than a refactor. I have not changed any code *per se* just refactored the config into a separate file and imported it in the gulp/webpack.

Can you give me the steps to reproduce this so I can see what's wrong?

bwreid:
@AmitJoki I agree; I wouldn't have expected this to cause any issues.

Can you give it a try yourself and let me know if you encounter similar issues? To reproduce, make the changes in this PR and then run gulp using the `gulp` command alongside your server. Try making changes to the front end code and see how quickly gulp picks up those changes.

For me, it quickly became significantly slower and/or stopped re-rendering altogether.

PR #3935, opened 2020-04-02
Feature/students tab refactor

bwreid:
## What this PR does

This PR joins together the Exercises & Trainings page with the Articles & Assignments page. A single page now shows Exercises, Training Modules, Article Assignments, and Revisions. This should affect all course types.

Features can be tested on the [staging server](https://dashboard-testing.wikiedu.org/courses/WikiEdu/Improved_Instructor_Dashboard_(Term)/home).

## Screenshots

**Before:**
image

**After:**
image


ragesoss:
Nice!

I think we should probably put Assigned (and Reviewing) over the trainings & exercises segment, but this looks great.

ragesoss:
Sweeeet.

A few things I noticed trying it out:

Trainings/Exercises vs Revisions are in the wrong spots.

backwards

Trainings/Exercises should show up if none are assigned.

no trainings or exercises

Clicking the username from the Students Overview is glitchy. First it switches to the Assignments & Exercises view briefly, then it reloads the page on the user profile page. Also, navigating 'back' in the browser from there results in a 404 on the URL that ought to be the A&S view of that student.

Maybe the best option here, since we know that instructors are likely to click on the username as the first thing to try, would be to have the username act just like a non-link click on the student row, and then add an additional link on the A&S view (either a new link labeled 'profile', or the username itself) that provides a way to reach the user profile page.

bwreid:
Thanks, this is helpful! I'll address those bugs and limit the clicking ability for the Overview page. I'm guessing that title switch happened during movement to I18n.

Also, you said:
> Trainings/Exercises should show up if none are assigned.
But I'm guessing you meant should *not* show up?

ragesoss:
> But I'm guessing you meant should not show up?

Yes, indeed.

PR #3944, opened 2020-04-07
Modifies 'new' article conditions.

### LINKED ISSUE #3941: Articles that start as sandboxes aren't counted as 'new' if the first edit is a template added by another editor
For example, this article was originally a sandbox, and the template was added by _User:Elysia (Wiki Ed)_ https://en.wikipedia.org/wiki/Dental_aerosol

It does not appear as 'new': https://dashboard.wikiedu.org/courses/University_of_Cincinnati/Occupational_Epidemiology_2020_(Spring_2020)/articles/edited

Elysia's edit was subsequently deleted, but this does not trigger recategorizing the article as new.

Ideally, the check for whether to call an article `new` would be based on the first non-automated edit being by a student, rather than just the first edit.

shashwatkathuria:
## What this PR does
Fixes #3941 
Counts an article as new <code>new_article</code> if it is created during the course, or its revisions(by any user, not only restricted to students) include <code>system: true</code> or <code>new_article: true</code>.

### DIFF
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -57,6 +57,10 @@ class ArticlesCourses < ApplicationRecord
     course.all_revisions.where(article_id: article_id)
   end
 
+  def article_revisions
+    article.revisions.where('date >= ?', course.start).where('date <= ?', course.end)
+  end
+
   def update_cache
     revisions = live_manual_revisions
 
@@ -67,9 +71,6 @@ class ArticlesCourses < ApplicationRecord
     # We use the 'all_revisions' scope so that the dashboard system edits that
     # create sandboxes are not excluded, since those are often wind up being the
     # first edit of a mainspace article's revision history
-    article_revisions = article.revisions
-                               .where('date >= ?', course.start)
-                               .where('date <= ?', course.end)
     self.new_article = all_revisions.exists?(new_article: true) ||
                        article_revisions.exists?(new_article: true, system: true)
     save
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -70,10 +70,8 @@ class ArticlesCourses < ApplicationRecord
     article_revisions = article.revisions
                                .where('date >= ?', course.start)
                                .where('date <= ?', course.end)
-    article_created_at = article.created_at
-    created_during_course = article_created_at > course.start && article_created_at < course.end
-    self.new_article = created_during_course ||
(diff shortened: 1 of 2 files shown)


app/models/course_data/articles_courses.rb
```diff
@@ -67,8 +67,13 @@ def update_cache
     # We use the 'all_revisions' scope so that the dashboard system edits that
     # create sandboxes are not excluded, since those are often wind up being the
     # first edit of a mainspace article's revision history
-    self.new_article = all_revisions.exists?(new_article: true)
-
+    article_revisions = article.revisions
+                               .where('date >= ?', course.start)
+                               .where('date <= ?', course.end)
+    article_created_at = article.created_at
```
ragesoss, line 73: The `created_at` value for an `Article` is the timestamp of when the record in the Dashboard database got created, not when the actual Wikipedia page was created. A timestamp of a revision with `new_article: true` is the only marker we have for when an article was created.

I think the two conditions where we can mark it 'new' are a) the way we determined it previously, or b) if there exists a revision for that article that is both `system: true` and `new_article: true` and that revision happened during the course.

shashwatkathuria: Okay! So I completely removed the condition involving <code>article.created_at</code> as the work is itself going to be done by the revisions.

shashwatkathuria: So now it only checks the following conditions:
1. Course revision by students for that article with<code>new_article: true</code> (Earlier way)
2. Any article revision(by anyone, not necessarily students) during course dates having <code>system: true</code> and <code>new_article: true</code>

app/models/course_data/articles_courses.rb
```diff
@@ -67,8 +67,11 @@ def update_cache
     # We use the 'all_revisions' scope so that the dashboard system edits that
     # create sandboxes are not excluded, since those are often wind up being the
     # first edit of a mainspace article's revision history
-    self.new_article = all_revisions.exists?(new_article: true)
-
+    article_revisions = article.revisions
```
ragesoss, line 70: This should probably be moved to a separate method, so that this code doesn't need to be run unless the first condition is false.

ragesoss:
I think this makes sense, but this needs to include some tests that exercise the new behavior.

The example article in the issue might be a good basis for it.

app/models/course_data/articles_courses.rb
shashwatkathuria: Done!

shashwatkathuria:
Added the tests!

PR #3952, opened 2020-04-08
Display whether or not a training module is late and due date

bwreid:
## What this PR does

This PR adds the due date to Training Modules in the same way it is displayed for Exercises. Additionally, it adds some "Late" text to the training modules to identify whether or not a module was completed after the deadline.

## Screenshots

image


ragesoss:
Looks good. I think the 'late' note for a completed module should show up after the competion date, not the completion time.

bwreid:
image

Do you think something like the first option or the second option, here?

ragesoss:
I like having it in the status column rather than next to the due date, and that lets the gray text contrast with the green without being in-your-face-danger red.

I meant the first of these rather than the second.

78814425-73e0a400-7983-11ea-8da2-f0257eafb7ef

bwreid:
Updated with new screenshot!

PR #3953, opened 2020-04-08
Discretionary sanctions monitor assignment based#3943

### LINKED ISSUE #3943: DiscretionarySanctionsMonitor should create alerts based on Assignments (in addition to edited articles)
DiscretionarySanctionsMonitor checks for articles that have been edited by students (represented by ArticlesCourses records) that are under 'discretionary sanctions' on Wikipedia, meaning they are controversial topics and come under extra scrutiny. It creates alerts to notify Wiki Experts when such an article is edited.

In addition to alerting us when such an article is edited, it would be a nice improvement to also alert us when such an article is chosen as an assignment by a student. The alert messages should differentiate between edited articles and (merely) assigned articles.

shashwatkathuria:
## What this PR does
Fixes #3943 . Adds a new feature (enhancement) of adding alert when a discretionary sanctions article is chosen as an article by a student.
## Screenshots
After:
Screenshot from 2020-04-09 01-34-31
Screenshot from 2020-04-09 01-05-37
Screenshot from 2020-04-09 01-05-43

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,6 +27,7 @@ Layout/LineLength:
 Metrics/ClassLength:
   Max: 127 # We should bring this down, ideally to the default of 100
   Exclude: # These are too big, and should be shrunk if feasible.
+    - 'app/models/alert.rb'
     - 'app/models/course.rb'
     - 'app/controllers/surveys_controller.rb'
     - 'app/controllers/campaigns_controller.rb'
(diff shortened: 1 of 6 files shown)


ragesoss:
This looks solid to me, but I checked in production and we have 810 existing assignments for articles under discretionary sanctions, which means that when this gets deployed it will immediately create 810 Alerts, mostly for courses that are long-ended.

I think the right fix will be to limit the assignment query to assignments for current courses (`Assignment.where(course: Course.current)`).

shashwatkathuria:
@ragesoss Done! Had to change the start and end date of the tests otherwise 5 tests of 7 were failing.

PR #3963, opened 2020-04-14
The current loaded spec version of capistrano is returned to lock as a string#3955

### LINKED ISSUE #3955: Build passes but deployment fails if Capistrano version is updated without changing deploy.rb
Our Capistrano version is locked using a config line at the beginning of `deploy.rb`, which makes Capistrano stop deploying if a different version is being used. However, the automated pull requests to bump the version of Capistrano don't change the `deploy.rb`, so we often end up merging one of these changes without updating `deploy.rb`, and then the builds fail until we notice (which might not be immediately, because the builds still pass even though the after-build deployments from travis-ci fail).

shashwatkathuria:
## What this PR does
Fixes #3955 

## Screenshots
Screenshot from 2020-04-14 13-41-49

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -1,4 +1,5 @@
-lock Gem.loaded_specs['capistrano'].version.to_s
+# config valid only for current version of Capistrano
+lock '3.13.0'
 
 set :application, 'wiki_edu_dashboard'
 set :repo_url, 'git@github.com:WikiEducationFoundation/WikiEduDashboard.git'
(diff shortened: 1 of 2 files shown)


ragesoss:
@shashwatkathuria does this use the version from the Gemfile.lock, or does it basically just look for whatever version is available in the environment and then "lock" to that, bypassing the point of the Capistrano 'lock' feature?

Maybe instead of bypassing the lock — which could be dangerous, allowing Capistrano to attempt a deployment with an old version that isn't compatible — we can add a spec that makes sure the lock config is in sync with the Gemfile.lock.

shashwatkathuria:
@ragesoss Done! (For reference : ) It looks for whatever version available, so I added a test for checking that the hard-coded version in `deploy.rb` is in sync with the `Gem.loaded_spec['capistrano'].version` which will always yield the latest version in Travis CI.

PR #3970, opened 2020-04-15
Wiki course edits replace default sandbox template#1600

### LINKED ISSUE #1600: WikiCourseEdits#enroll_in_course should replace the default sandbox template if present
`WikiCourseEdits#enroll_in_course` makes several edits, including one edit that adds a custom template to the user's "/sandbox" user subpage. This is done to preempt the default template that gets added if the user clicks the Sandbox link from their personal tools on Wikipedia (and which steers users to Articles for Creation, which we want to spare them from).

Improve this method so that, if the default sandbox template is already present, it will be replaced instead of just adding the custom template to the page.

shashwatkathuria:
## What this PR does
Fixes #1600 . Adds new service for the functionality.

### DIFF
--- a/app/services/add_sandbox_template.rb
+++ b/app/services/add_sandbox_template.rb
@@ -12,7 +12,10 @@ class AddSandboxTemplate
     @wiki_editor = WikiEdits.new(home_wiki)
     @wiki_api = WikiApi.new(home_wiki)
     @initial_page_content = @wiki_api.get_page_content(@sandbox)
+    add_template
   end
+
+  private
 
   def add_template
     # Never double-post the sandbox template
--- a/app/services/add_sandbox_template.rb
+++ b/app/services/add_sandbox_template.rb
@@ -4,14 +4,14 @@ require_dependency "#{Rails.root}/lib/wiki_api"
 require_dependency "#{Rails.root}/lib/wiki_edits"
 
 class AddSandboxTemplate
+  DEFAULT_TEMPLATE = '{{user sandbox}}'
   def initialize(home_wiki:, sandbox:, sandbox_template:, current_user:)
     @current_user = current_user
     @sandbox = sandbox
-    @default_template = '{{user sandbox}}'
     @sandbox_template = sandbox_template
     @wiki_editor = WikiEdits.new(home_wiki)
-    @wiki_api = WikiApi.new(home_wiki)
-    @initial_page_content = @wiki_api.get_page_content(@sandbox)
+    wiki_api = WikiApi.new(home_wiki)
+    @initial_page_content = wiki_api.get_page_content(@sandbox)
     add_template
   end
 
@@ -20,7 +20,11 @@ class AddSandboxTemplate
   def add_template
     # Never double-post the sandbox template
     return if sandbox_template_present?
-    default_template_present? ? replace_default_with_sandbox_template : add_sandbox_template
+    if default_template_present?
+      replace_default_with_sandbox_template
+    else
+      add_sandbox_template
+    end
   end
 
   def sandbox_template_present?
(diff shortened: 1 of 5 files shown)


app/services/add_sandbox_template.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/wiki_api"
+require_dependency "#{Rails.root}/lib/wiki_edits"
+
+class AddSandboxTemplate
+  def initialize(home_wiki:, sandbox:, sandbox_template:, current_user:)
+    @current_user = current_user
+    @sandbox = sandbox
+    @default_template = '{{user sandbox}}'
```
ragesoss, line 10: This can be a constant rather than an instance variable.

app/services/add_sandbox_template.rb
```diff
@@ -0,0 +1,45 @@
...
+    @sandbox_template = sandbox_template
+    @wiki_editor = WikiEdits.new(home_wiki)
+    @wiki_api = WikiApi.new(home_wiki)
```
ragesoss, line 13: If this is only used in `initialize`, it should be a local variable rather than an instance variable. In general, the fewer instance variables a class has, the easier it is to understand how it works and what the variables are for.

spec/services/add_sandbox_template_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+describe AddSandboxTemplate do
+  let(:course) { create(:course, home_wiki_id: 1) }
+  let(:user) { create(:user) }
+  let(:enrolling_user) { create(:user, username: 'Belajane41') }
+  let(:no_template) { 'No default or sandbox template present.' }
+  let(:sandbox) { "User:#{enrolling_user.username}/sandbox" }
+
+  before do
+    @sandbox_template = "{{#{ENV['dashboard_url']} sandbox}}"
+    @default_template = '{{user sandbox}}'
```
ragesoss, line 14: These can both be done with `let` also.

spec/services/greet_ungreeted_students_spec.rb
```diff
@@ -20,7 +20,8 @@
     # New section for the welcome message
     expect_any_instance_of(WikiEdits).to receive(:add_new_section).and_call_original
     # Templates are added to the top of user page, talk page, sandbox
-    expect_any_instance_of(WikiEdits).to receive(:add_to_page_top).thrice.and_call_original
+    expect(AddSandboxTemplate).to receive(:new)
```
ragesoss, line 23: Does this still pass if you use `.and_call_original`? If so, that would be better. This test stubs the requests themselves, so it's designed for all the code up to the actual network requests to run.

app/services/add_sandbox_template.rb
```diff
@@ -0,0 +1,45 @@
...
+    @initial_page_content = @wiki_api.get_page_content(@sandbox)
+    add_template
+  end
+
+  private
+
+  def add_template
+    # Never double-post the sandbox template
+    return if sandbox_template_present?
+    default_template_present? ? replace_default_with_sandbox_template : add_sandbox_template
```
ragesoss, line 23: There are three possibly paths here, depending on the state of the sandbox page: do nothing, add the template, replace the default template.

This will be clearer if all three paths are treated in parallel ways, either with a case statement or an if/elsif/else.

app/services/add_sandbox_template.rb
shashwatkathuria: Done!

app/services/add_sandbox_template.rb
shashwatkathuria: Done!

spec/services/add_sandbox_template_spec.rb
shashwatkathuria: Done!

spec/services/greet_ungreeted_students_spec.rb
shashwatkathuria: Nope, it fails when I use `.and_call_original`, it is having problems with the new `WikiEdits` instance I make there.

app/services/add_sandbox_template.rb
shashwatkathuria: Done!

app/services/add_sandbox_template.rb
```diff
@@ -0,0 +1,49 @@
...
+    wiki_api = WikiApi.new(home_wiki)
+    @initial_page_content = wiki_api.get_page_content(@sandbox)
+    add_template
+  end
+
+  private
+
+  def add_template
+    # Never double-post the sandbox template
+    return if sandbox_template_present?
```
ragesoss, line 22: To really make this an equal branch, it should be
```ruby
if sandbox_template_present?
  return
elsif ...
else ...
end
```

That way, it's super clear that one (and only one) of the three options should happen.

shashwatkathuria: Done!

PR #3975, opened 2020-04-16
Rubocop new pending cops bulk configuration#3961

### LINKED ISSUE #3961: Enable pending new cops in bulk.
### What is happening?
Every time <code>rubocop</code> is bumped (updated) some new cops are introduced, which leads to warnings like the ones given in the picture. A new option of <code>NewCops</code> listed at https://docs.rubocop.org/en/latest/versioning/ would solve the issue, but the feature is yet unreleased so we'll have to wait a bit until the feature is released.
Screenshot from 2020-04-14 07-16-28

(3rd Line # 7850)
Screenshot from 2020-04-14 02-35-43

shashwatkathuria:
## What this PR does
Fixes #3961 . Also enables some pending previous cops which were having violations and fixes their violations.


app/models/training_slide.rb
```diff
@@ -62,7 +62,7 @@ def self.inflate(all_content, slug, wiki_page = nil)
     slide.save
     slide
   rescue StandardError => e
-    puts "There's a problem with file '#{slug}'"
```
ragesoss, line 65: Let's not change this; it's easier to notice when this particular problem happens than if it's done with the logger, I think.

app/helpers/campaign_helper.rb
```diff
@@ -15,6 +15,6 @@ def nav_link(link_text, link_path)
   def html_from_markdown(markdown)
     return unless markdown
     converter = Redcarpet::Markdown.new(Redcarpet::Render::HTML)
-    raw converter.render(markdown)
+    sanitize converter.render(markdown)
```
ragesoss, line 18: I'm not sure what the difference here is or whether this could cause bugs. This change shouldn't be part of this PR.

ragesoss:
The refactoring changes here look good. Please undo the functional changes to logging and HTML rendering.

app/helpers/campaign_helper.rb
shashwatkathuria: Okay, I've removed it

app/models/training_slide.rb
shashwatkathuria: Okay, I've undone these changes.

PR #3987, opened 2020-04-21
Renamed UNSAFE_ComponentWillMount methods to ComponentDidMount in React.

### LINKED ISSUE #3478: Refactor componentWillMount and componentWillReceiveProps

shashwatkathuria:
## What this PR does
Fixes #3478 . Renames `UNSAFE` methods in `React`

### DIFF
--- a/app/assets/javascripts/components/alerts/admin_alerts.jsx
+++ b/app/assets/javascripts/components/alerts/admin_alerts.jsx
@@ -6,7 +6,7 @@ import AlertsHandler from './alerts_handler.jsx';
 import { fetchAdminAlerts } from '../../actions/alert_actions';
 
 class AdminAlerts extends React.Component {
-  UNSAFE_componentWillMount() {
+  componentDidMount() {
     // This adds ALL alerts to the state, to be used in AlertsHandler
     this.props.fetchAdminAlerts();
   }
--- a/app/assets/javascripts/components/alerts/campaign_alerts.jsx
+++ b/app/assets/javascripts/components/alerts/campaign_alerts.jsx
@@ -18,14 +18,14 @@ class CampaignAlerts extends React.Component {
     this.getCampaignSlug = this.getCampaignSlug.bind(this);
   }
 
-  getCampaignSlug() {
-    return `${this.props.match.params.campaign_slug}`;
-  }
-
-  UNSAFE_componentWillMount() {
+  componentDidMount() {
     // This adds the specific campaign alerts to the state, to be used in AlertsHandler
     this.props.fetchCampaignAlerts(this.getCampaignSlug());
     this.props.filterAlerts(this.state.defaultFilters);
+  }
+
+  getCampaignSlug() {
+    return `${this.props.match.params.campaign_slug}`;
   }
 
   render() {
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -36,6 +36,16 @@ const ArticleFinder = createReactClass({
     };
   },
 
+  componentDidMount() {
+    if (window.location.search.substring(1)) {
+      this.getParamsURL();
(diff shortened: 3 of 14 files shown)


ragesoss:
@bwreid This looks all good to me. Give it a quick look?

shashwatkathuria:
@bwreid 
After making changes in this PR, I got to know that the `MultiSelectField` is not working as expected, and after some console logging, got to know that the value of `this.state.value` which is set to `this.props.selected` (only initially) is not getting updated. The following code snippet solves the problem but I wanted some help in when should I stop the updation process, otherwise it ends up in an infinite loop, as is generally when we use `setState` inside of `componentDidUpdate`, which can be solved by carefully adding a condition(Line 25) on when to update.
Peek 2020-04-22 01-57
Peek 2020-04-22 01-58
Screenshot from 2020-04-22 03-18-27

ragesoss:
This uncovers some awkward state/props interactions with `MultiSelectField`, but those can perhaps be fixed separately.

PR #3989, opened 2020-04-21
Feature/user ids on articles courses

bwreid:
## What this PR does

This PR adds a new row on the Students > Articles & Exercises page that displays articles that the user has edited but is not assigned to the user. It does so by adding a column to the **ArticlesCourses** model that stores User IDs from associated revisions.

## Screenshots

image

### DIFF
--- a/app/views/courses/articles.json.jbuilder
+++ b/app/views/courses/articles.json.jbuilder
@@ -11,6 +11,6 @@ json.course do
     json.url article.url
     json.rating_num rating_priority(article.rating)
     json.pretty_rating rating_display(article.rating)
-    json.user_ids ac.details['user_ids']
+    json.user_ids ac.user_ids
   end
 end
--- a/app/views/courses/articles.json.jbuilder
+++ b/app/views/courses/articles.json.jbuilder
@@ -11,5 +11,6 @@ json.course do
     json.url article.url
     json.rating_num rating_priority(article.rating)
     json.pretty_rating rating_display(article.rating)
+    json.user_ids ac.details['user_ids']
   end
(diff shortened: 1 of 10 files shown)


app/views/courses/articles.json.jbuilder
```diff
@@ -11,5 +11,6 @@ json.course do
     json.url article.url
     json.rating_num rating_priority(article.rating)
     json.pretty_rating rating_display(article.rating)
+    json.user_ids ac.details['user_ids']
```
ragesoss, line 14: I think implementing it this way can result in this coming through as `null` sometimes. The AC records get created before the `update_cache` happens, this will not be an array (even an empy one) until the first time the AC record gets an `update_cache` cycle.

That could be worked around either by the new column explicitly `user_ids` and serialize it to an array, or with a method to on AC to handle transforming the `details` into an array of user ids.

bwreid: Your second sentence is missing... something. I tried to set details to be a default column but ran across an error in doing so. Maybe instead I can set up a method that is `user_ids` and pull from `details['user_ids']` or set to an array?

ragesoss: I meant...
> The AC records get created before the update_cache happens, **so** this will not be an array (even an empy one) until the first time the AC record gets an update_cache cycle.

And also...
> That could be worked around either by **naming** the new column explicitly user_ids and serialize it to an array, or with a method to on AC to handle transforming the details into an array of user ids.

ragesoss:
LGTM

PR #4015, opened 2020-05-05
Fix debian setup shell script

jonathan-richer:
## What this PR does
It fixes a typo in the `setup/deb-setup.sh` file preventing debian-based system to run it properly.
and it changes a path of a .rb file which is 

## Screenshots
Before:
image

After fixing the typo:
image
I didn't know that the filepaths given in the shell script are relative to the python file that calls it.

I have read the [CONTRIBUTING.md](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/CONTRIBUTING.md#pull-request-process) and I know that I'm maybe not conform to the instructions, however I believe that this small fix, given that it's not on the project's content, didn't need to be rebased, and that it didn't need another branch only for this one line. If it causes problem, tell me and I will do what's needed to conform to every points of the [CONTRIBUTING.md](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/CONTRIBUTING.md#pull-request-process)

### DIFF
--- a/setup/deb-setup.sh
+++ b/setup/deb-setup.sh
@@ -117,7 +117,7 @@ else
 fi
 
 printf '[*] Updating bundler... \n'
-if ruby should_update_rubygems.rb
+if ruby setup/should_update_rubygems.rb; then
   gem update --system
   printf '[+] Bundler updated\n'
 else


ragesoss:
Thanks! This seems good to me.

ragesoss:
It looks like there's another bug in the debian script, which someone else hit today.... there's a missing `; then` at the end of line 120.

jonathan-richer:
@ragesoss This is one of the 2 fixes, it needed a `;` and a `then` at the end, but on the same line, with the fix, it couldn't find a `.rb` script, which I fixed too.

PR #4016, opened 2020-05-05
Remove usage deprecated component will receive props#3478

### LINKED ISSUE #3478: Refactor componentWillMount and componentWillReceiveProps

shashwatkathuria:
## What this PR does
Fixes #3478 . Uses `componentDidUpdate` or/and `getDerivedStateFromProps` as required in the scenario. The method being removed is `UNSAFE_componentWillReceiveProps` .


app/assets/javascripts/components/article_finder/article_finder_row.jsx
```diff
@@ -14,8 +14,9 @@ const ArticleFinderRow = createReactClass({
     };
   },
 
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (this.state.isLoading && (this.props.assignment !== nextProps.assignment)) {
+  componentDidUpdate(prevProps, prevState) {
```
ragesoss, line 17: This would benefit from a comment explaining how we're using `isLoading` for two different situations and so need to do this in a weird way that violates the usual best practices.

app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -67,11 +67,9 @@ const CourseCreator = createReactClass({
   },
 
   UNSAFE_componentWillReceiveProps(nextProps) {
-    if (nextProps.course.school !== '' && nextProps.course.title !== '') {
-      this.state.tempCourseId = CourseUtils.generateTempId(nextProps.course);
-    } else {
-      this.state.tempCourseId = '';
-    }
+    this.setState({
```
ragesoss, line 70: I think this may result in behavior change when creating a new course.

The idea is that we don't want the current version of the course slug (ie, the tempCourseId) to show up in the CourseCreator modal at the bottom until after the user has added something for both school and title.

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -37,6 +37,18 @@ const DiffViewer = createReactClass({
     };
   },
 
+  // When 'show' is clicked, this component may or may not already have
+  // users data (a list of usernames) in its props. If it does, then 'show' will
+  // fetch the MediaWiki user ids, which are used for coloration. Those can't be
+  // fetched until the usernames are available, so 'show' will fetch the usernames
+  // first in that case. In that case, componentDidUpdate fetches the
+  // user ids as soon as usernames are avaialable.
+  componentDidUpdate(prevProps, prevState) {
```
ragesoss, line 46: This looks fine, but did you test it out manually on the Articles tab of a course with edited articles, as well as the Students tab?

This is behavior that is pretty important, but I don't think it's tested in any of feature specs.

app/assets/javascripts/components/course_creator/course_creator.jsx
ragesoss: Nevermind! I see that you moved that behavior into CourseUtils. Good!

app/assets/javascripts/training/components/quiz.jsx
```diff
@@ -29,10 +29,6 @@ const Quiz = createReactClass({
     return this.setState({ selectedAnswerId: e.currentTarget.getAttribute('data-answer-id') });
   },
 
-  UNSAFE_componentWillReceiveProps(newProps) {
```
ragesoss, line 32: Remind me what's up with this file. I'm confused about how we're using state vs. props for keeping track of the selected answer. Was this bit of copying props over to state just unnecessary?

Are there better names we could use than `setAnswer` vs `setSelectedAnswer` to make it clear what's going on?

shashwatkathuria: Okay...So through `setAnswer`, we first update the choice user has chosen from the `e.currentTarget`. This update is done through setState (ofc), which means that the state is updated successfully. Now that this part is done as and when the user wants to change the option, now the user clicks on `CHECK ANSWER` which calls verifyAnswer which in turn calls `setSelectedAnswer`. Now, what `setSelectedAnswer` does is, that it updates the redux state through the action appropriately so that the user is able to go to the next slide if the option chosen is the same as `state.currentSlide.assessment.correct_answer_id`. `this.props.selectedAnswerId` is actually nothing in the process I just elaborated and so always comes out to be `undefined`. This is why I removed it.

app/assets/javascripts/components/revisions/diff_viewer.jsx
shashwatkathuria: Yes, I did check the `diff_viewer.jsx` manually (the popup that shows the differences in the revised version form the previous one). I'll submit its feature spec in a new PR.

app/assets/javascripts/components/article_finder/article_finder_row.jsx
shashwatkathuria: Done!

ragesoss:
@AmitJoki could you give this a review? I've gone through it a few times and don't notice anything else that I can suggest.

app/assets/javascripts/components/common/calendar.jsx
```diff
@@ -23,6 +23,14 @@ const Calendar = createReactClass({
     updateCourse: PropTypes.func.isRequired
   },
 
+  statics: {
```
AmitJoki, line 26: If we won't be using any more static methods, it is better to go with `static getDerivedStateFromProps`, keeps the code concise.

app/assets/javascripts/components/common/calendar.jsx
```diff
@@ -34,12 +42,6 @@ const Calendar = createReactClass({
     }
     return true;
   },
-  UNSAFE_componentWillReceiveProps(nextProps) {
-    if (nextProps.course.start !== moment(this.state.initialMonth).format('YYYY-MM-DD')) {
```
AmitJoki, line 38: is this condition not needed?

AmitJoki:
I have left a comment @ragesoss . Also I think checkAnswer can be a bit ambiguous. How about `reviewAnswer`?

app/assets/javascripts/components/common/calendar.jsx
shashwatkathuria: I talked to @ragesoss in Slack about this. Firstly the statement is being compared in a wrong way, i.e., ISO String to yyyy-mm-dd (ISO String contains both date and time). Secondly, when I corrected this comparison and shifted it to `getDerivedStateFromProps`, @ragesoss told me to leave this statement as we were calling `moment` twice in this and this is not really necessary and to leave this condition altogether.

PR #4021, opened 2020-05-07
Added 'macos' keyword option to setup script 

kimberleejohnson:
## What this PR does
When I ran the setup.py script, I kept receiving the error that my operating system is not supported. I'm running macOS Catalina. 

I ran the line 45 of setup.py in my python shell, and confirmed that -1 was the value returned. It seems like platform.system() returns 'Darwin', but platform.platform() returned 'macOS'. 

So, I added an or statement to the script to also check for 'darwin' *or* 'mac' to run osx_setup(). 

## Screenshots
Before, showing my operating system not supported: 
before_operating_system_not_supported

Confirming what platform and system I ran from the shell: 
platform_check

After the *or* addition, system found: 
after_system_found

There might be other, better workarounds for this, but this change made the script work for me! Shout with any feedback, and thank you!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,8 @@ if platform.platform().lower().find('ubuntu') != -1 \
     deb_setup()
 elif platform.platform().lower().find('fedora') != -1:
     dnf_setup()
-elif platform.platform().lower().find('darwin') != -1:
+elif platform.platform().lower().find('darwin') != -1 \
+        or platform.platform().lower().find('mac') != -1: 
     osx_setup()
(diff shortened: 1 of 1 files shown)


kimberleejohnson:
I see that this failed tests and I will look into it!

ragesoss:
Those failing tests are definitely unrelated to your PR. Those are cases where an external API behaved badly during the test run.

kimberleejohnson:
> Those failing tests are definitely unrelated to your PR. Those are cases where an external API behaved badly during the test run.

Thank you so much for confirming!

Issue #3829, opened 2020-03-01
Scrolling Effect Breaking in Firefox for resources and profile section


hardy1334:
In Firefox, when we scroll down the page the page starts blinking, for some amount of time. I have attached a video to have a look at the issue. This behaviour is happening when we visit resources section in course and when a user visits his profile.

https://drive.google.com/file/d/1Y99xcthP9hw3XWqrch06JiwbQ9G1furq/view

ragesoss:
Thanks! This is very similar to #2012

It think it's fundamentally the same issue on the Resources tab of a course page — it comes from the use of the `Affix` component to fix the navbar in position at the top of the window as the user scrolls down — but it's a little different for profile pages.

It looks like here, it's done via the jquery code in `app/assets/javascripts/utils/users_profile.js`.

In both cases, the bug occurs when the size of the page content is just large enough that it requires scrolling to see the bottom of the page, but once you've scrolled far enough to activate the fixed header, the whole page now fits in the window and scroll position changes and the header goes back to unfixed mode, and keeps switching back and forth.

For the user profile variant of the bug, I think the solution (to start with at least) is to finish converting the profile page to React, and replacing the jquery with the same `Affix` strategy used in course pages... then we can find a fix by changing how `Affix` works that will solve it both places.

hardy1334:
@ragesoss True, I also agree for the same

PR #4028, opened 2020-05-11
Fix jump in page content when scrolling

jproft:
## What this PR does
Fixes a content jump due to the sticky nav being removed from the flow of the page when its position is set to fixed. The fix is achieved by giving the sticky nav wrapper the same height as the sticky nav itself.

## Screenshots
Before:

Jittering on scroll:
before_1

Visible content jump when scrolling:
before_2

After:

No longer jittering (can actually scroll to the bottom of the page!):
after_1

No content jump when scrolling:
after_2


ragesoss:
This is wonderful! That's been an annoying little bug for a long time: #2012 (and related, which this probably unblocks a fix to, #3829)

PR #4037, opened 2020-05-14
Convert _question_group.html.erb to HAML

jproft:
## What this PR does
This PR converts the last remaining ERB file in this repository to HAML.

I discovered that the reason this file had to be converted manually to HAML (see PR #1067) is the conditional adding and closing of `div`s -- the existing automatic ERB to HAML converters just can't handle that. The consequence is that there are three places in the HAML file where I had to inject HTML. Not the prettiest solution, but additional refactoring elsewhere will be required if a cleaner solution is desired.


ragesoss:
Nice! I looked at the HTML from before and after this change for the test survey in `spec/features/surveys_spec.rb` and they are identical. That survey involves grouped questions, so any changes to those lines that required manual conversion breaks the test (and breaks the whole survey UI).

I'm glad to have the last bits of ERB gone!

I don't think it's worthwhile to invest in refactoring this. The current approach to rendering a survey, which involves putting the entire survey in the HTML but then creating a one-by-one question UI on top of that using jquery (839 lines of `app/assets/javascripts/surveys/modules/Survey.js`), is confusing and very difficult to work with. If and when we tackle overhauling the survey code, I think we'll want to rewrite it to use React for the UI, putting the survey content into JSON rather than relying on server-generated partials.

PR #4038, opened 2020-05-16
Migrate to Webpack-only configuration

AmitJoki:
I'll be removing the gulp tasks and replace them with the Webpack equivalent.


webpack.dev.js
```diff
@@ -0,0 +1,117 @@
...
+
+const outputPath = path.resolve(`${config.outputPath}/${config.jsDirectory}`);
+const jsSource = `./${config.sourcePath}/${config.jsDirectory}`;
+const cleanPaths = [
+  `${config.outputPath}/fonts/*`,
+  `${config.outputPath}/images/*`,
+  `${config.outputPath}/stylesheets/*`,
+  `${config.outputPath}/javascripts/*`
+];
+const entries = {
```
ragesoss, line 17: What can be done to avoid duplicating code across the three different webpack config files?

webpack.config.js
```diff
@@ -0,0 +1,115 @@
...
+          loader: 'babel-loader',
+          query: {
+            cacheDirectory: true
+          }
+        }
+      },
+      {
+        test: /\.jsx?$/,
+        exclude: [/vendor/, /node_modules(?!\/striptags)/],
+        loader: 'eslint-loader',
```
ragesoss, line 101: What does this do? I assume the version bump of eslint was required for this?

webpack.dev.js
AmitJoki: we can export them from a single file and import them.

webpack.config.js
AmitJoki: gulp used to do the linting. Now Webpack does.

webpack.config.js
```diff
@@ -0,0 +1,115 @@
...
+const wpPlugins = [];
+const absoluteOutputPath = path.join(process.cwd(), config.outputPath);
+const jqueryUlsPath = './node_modules/jquery/dist/jquery.min.js';
+const mode = 'production';
+
+wpPlugins.push(new CleanWebpackPlugin({
+  cleanOnceBeforeBuildPatterns: cleanPaths.map(relative => path.join(process.cwd(), relative))
+}));
+
+wpPlugins.push({
```
ragesoss, line 39: Can you add some comments to explain the strategy here for the I18n pipeline?

Also, is there a simple way to add fingerprinting? A common problem is new interface messages get deployed and users get the new main.js build that uses them, but they see 'message missing' until they clear their cached I18n files since they don't get fingerprinted.

ragesoss:
Looks like this approximately doubles the build time (54 seconds for `gulp build` vs 114 seconds for `build` plus 7 seconds for `gulp`).

Any ideas why? I'm not especially worried about build time, but the more we know about what contributes to long build times, the better.

AmitJoki:
gulp has parallel execution which is not the case with Webpack I am assuming.

webpack.config.js
AmitJoki: Sure. Though I don't think i18n is part of the main bundle.

AmitJoki:
@ragesoss where did you get the time? I checked it out on Travis CI. The `yarn gulp build` is taking more than `110` seconds to complete.

The `yarn build` is taking just `32` seconds.

AmitJoki:
This is on master branch.
https://travis-ci.org/github/WikiEducationFoundation/WikiEduDashboard/jobs/688589512#L3418

This is on this branch:
https://travis-ci.org/github/WikiEducationFoundation/WikiEduDashboard/jobs/688889379#L3430

ragesoss:
The `gulp` integration is buggy in dev/watch mode. After the initial build, when I save a new JS change, I get something like this:

```
[09:37:21] Requiring external module @babel/register
[09:37:22] Using gulpfile ~/WikiEduDashboard/gulpfile.babel.js
[09:37:22] Starting 'dev'...
[09:37:22] Starting 'stylesheets'...
[09:37:24] Finished 'stylesheets' after 1.98 s
[09:37:24] Starting 'watch-css'...

... Uhoh. Got error listen EADDRINUSE: address already in use :::35729 ...
Error: listen EADDRINUSE: address already in use :::35729
    at Server.setupListenHandle [as _listen2] (net.js:1259:14)
    at listenInCluster (net.js:1307:12)
    at Server.listen (net.js:1395:7)
    at Server.listen (/home/sage/WikiEduDashboard/node_modules/tiny-lr/src/server.js:264:19)
    at Function.listen (/home/sage/WikiEduDashboard/node_modules/gulp-livereload/index.js:97:18)
    at /home/sage/WikiEduDashboard/gulp/tasks/watch-css.js:13:22
    at taskWrapper (/home/sage/WikiEduDashboard/node_modules/undertaker/lib/set-task.js:13:15)
    at bound (domain.js:425:14)
    at runBound (domain.js:438:12)
    at asyncRunner (/home/sage/WikiEduDashboard/node_modules/async-done/index.js:55:18)

You already have a server listening on 35729
You should stop it and try again.
```

I guess it ought to do the `gulp` the first time, and not again after every emit? Also, can `gulp` be started in parallel before the first emit? On my machine, it takes about 5 seconds to emit the webpack assets, and then 2 seconds to compile the stylesheets. Those are indepedent, so it ought to be possible to do them at the same time.

ragesoss:
A few notes:
* With this change, there's no linting of the JS files that aren't part of the build — eg, the webpack config itself (which has some eslint violations in the current form). What's a good way to fix that?
* With this change, javascript is not emitted at all when there are linting errors, and the dev environment stops working until listing errors are fixed. Previously, webpack would still emit an updated working JS file even if a linting error was introduced, but you would see the error/warning in the gulp terminal. Can it be configured to still build with eslint errors?

webpack.config.js
```diff
@@ -0,0 +1,220 @@
...
+  raven: [`${jsSource}/raven.js`],
+  styleguide: [`${jsSource}/styleguide/styleguide.jsx`],
+  survey: [`${jsSource}/surveys/survey.js`],
+  survey_admin: [`${jsSource}/surveys/survey-admin.js`],
+  survey_results: [`${jsSource}/surveys/survey-results.jsx`],
+  campaigns: [`${jsSource}/campaigns.js`],
+  charts: [`${jsSource}/charts.js`],
+  tinymce: [`${jsSource}/tinymce.js`],
+  embed_course_stats: [`${jsSource}/embed_course_stats.js`],
+  main_css: [`${cssSource}/main.styl`],
```
ragesoss, line 38: Is there a good way to avoid the arbitrary `_css` in the filenames?

ragesoss:
This seems to be working very smoothly now. I'm slightly annoyed about the name collision in the entries that makes us add `_css` to the stylesheet filenames, but I can live with that if there's no clean way to avoid it.

PR #4039, opened 2020-05-18
(Old) Tracking course update errors downwards

shashwatkathuria:
## Note
Final cleaner version of this PR at #4076


lib/ores_api.rb
```diff
@@ -20,17 +20,22 @@ def self.valid_wiki?(wiki)
...
     raise InvalidProjectError unless OresApi.valid_wiki?(wiki)
     @project_code = wiki.project == 'wikidata' ? 'wikidata' + 'wiki' : wiki.language + 'wiki'
     @project_model = wiki.project == 'wikidata' ? 'itemquality' : 'articlequality'
+    @course = course
   end
 
   def get_revision_data(rev_ids)
-    response = ores_server.get query_url(rev_ids)
+    server_conn = ores_server
+    url_prefix = server_conn.url_prefix
```
ragesoss, line 32: what's this?

lib/wiki_api.rb
```diff
@@ -139,4 +142,13 @@ def typical_errors
      Faraday::ConnectionFailed,
      MediawikiApi::HttpError]
   end
+
+  def mediawiki_request_url
+    @mediawiki.instance_variable_get(:@conn).instance_variable_get(:@url_prefix).to_s
+  end
+
+  def save_course_error_record(type_of_error, url)
```
ragesoss, line 150: It doesn't necessarily need to be done now... the more time you have to think about the ideal way to do it, the better... but it might make sense to extract the error-recording code as a `module` and including it in the API classes that would use it.

lib/ores_api.rb
shashwatkathuria: What this is doing is extracting the prefix of the url of `Faraday::Connection` object. I couldn't really make sense of how to get the full url like `https://ores.wikimedia.org/v3/scores/wikidatawiki/?models=itemquality&features&revids=1136533131|1136536523` . The `server_conn.url_prefix.to_s` in this case is basically `https://ores.wikimedia.org/` and the `url_query` part is `/v3/scores/wikidatawiki/?models=itemquality&features&revids=1136533131|1136536523`. I had to use these variables in the recue block below to pass on the full url to `save_course_error_record`.

lib/wiki_api.rb
shashwatkathuria: Okay, I'll shift the code there. Also, it was the same in the three APIs so I anyways had to refactor it somewhere.

lib/ores_api.rb
shashwatkathuria: I also have the same doubt in `def mediawiki` in `WikiApi` on what is the correct way to get the url of the `MediawikiApi::Client` object `@mediawiki`

lib/errors/course_error_records.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module Errors
```
ragesoss, line 3: Why is this a module nested within a module?

lib/errors/course_error_records.rb
```diff
@@ -0,0 +1,10 @@
...
+  module CourseErrorRecords
```
ragesoss, line 4: Following the convention in the (limited) amount of non-Helper modules in the code base, consider naming this after what it does: `CourseUpdateErrorHandling`.

lib/ores_api.rb
ragesoss: Since this prefix is the same thing that you pass when setting up the Faraday object in the `ores_server` method, this will be simpler to just extract as a constant and re-use, without getting involved in additional methods on the connection object.

lib/wiki_api.rb
```diff
@@ -139,4 +150,8 @@ def typical_errors
      Faraday::ConnectionFailed,
      MediawikiApi::HttpError]
   end
+
+  def mediawiki_request_url
+    @mediawiki.instance_variable_get(:@conn).instance_variable_get(:@url_prefix).to_s
```
ragesoss, line 155: Is this the full URL, or just the domain?

I guess that the `api_url` is the same domain plus some extra parts of the path?

In this case, we're reaching rather far into the MediawikiApi gem to try to find a URL, when we never dealt with a URL directly in the first place, instead leaving the construction of a URL from the query params completely up to gem. We should probably instead save the same set of data that we currently ship off to Sentry: the action, query and api_url.

Probably that means adding another column to error record table, for miscellaneous details.

lib/errors/course_error_records.rb
shashwatkathuria: I started with a non-nested module, but I could find less examples and got influenced by the convention followed by the other modules in `lib/errors` so I did this. I'll remove the outer module.

lib/errors/course_error_records.rb
shashwatkathuria: Done!

lib/ores_api.rb
shashwatkathuria: Okay..Done!

lib/wiki_api.rb
shashwatkathuria: In this case, I actually have to verify whether or not the full URL is always there or not. During development when I raise the error after `@mediawiki.send(action, query)`, I get the full url correctly. But that raising of error is done after the request is correctly completed. On the other hand, the testing method(which uses stubbing and raises errors) gives only the domain. I actually have to see what actually happens in the real update scenario.

shashwatkathuria: Okay, I will add a `miscellaneous` column and save the details there.

lib/ores_api.rb
```diff
@@ -57,8 +66,31 @@ def query_url(rev_ids)
   end
 
   def ores_server
-    conn = Faraday.new(url: 'https://ores.wikimedia.org')
+    conn = Faraday.new(url: ORES_SERVER_URL)
     conn.headers['User-Agent'] = ENV['dashboard_url'] + ' ' + Rails.env
     conn
   end
+
+  def report_exception(error, url, level: 'error', sentry_tag_uuid: nil)
```
ragesoss, line 74: Why is all this error-handling code in OresApi instead of the error-handling module?

shashwatkathuria: I refactored the error handling code in the three APIs to `ErrorHandling` module.

AmitJoki:
Looks good!

lib/replica.rb
```diff
@@ -207,11 +210,17 @@ def compile_revision_ids_query(revisions)
...
   TYPICAL_ERRORS = [Errno::ETIMEDOUT, Net::ReadTimeout, Errno::ECONNREFUSED,
                     Oj::ParseError].freeze
-  def report_exception(error, endpoint, query, level='error')
-    level = 'warning' if TYPICAL_ERRORS.include?(error.class)
-    Raven.capture_exception error, level: level, extra: {
-      query: query, endpoint: endpoint, language: @wiki.language, project: @wiki.project
-    }
-    return nil
+
+  def invoke_error_handling_tasks(error, endpoint, query: nil, url: nil, response: nil)
```
ragesoss, line 214: I think most of this can and should be moved into the error handling module. The closer we get to uniform handling of API errors, instead of treating each different API in a unique way, the easier it will be to extend this strategy to additional places.

There's a lot of duplication between this and OresApi.

lib/wiki_api.rb
```diff
@@ -97,46 +101,36 @@ def mediawiki(action, query)
...
-    [Faraday::TimeoutError,
-     Faraday::ConnectionFailed,
-     MediawikiApi::HttpError]
+  TYPICAL_ERRORS = [Faraday::TimeoutError,
+                    Faraday::ConnectionFailed,
+                    MediawikiApi::HttpError,
+                    MediawikiApi::ApiError].freeze
+
+  def invoke_error_handling_tasks(error, action, query)
+    Rails.logger.info "Caught #{error}"
```
ragesoss, line 131: This type of logging should definitely be in the error handling module and not the API class.

lib/wiki_api.rb
```diff
@@ -97,46 +101,36 @@ def mediawiki(action, query)
...
-    retry if tries >= 0
-    Raven.capture_exception e, level: 'warning'
-    return nil # Do not return a Raven object
+    # Raise unknown errors.
+    # Continue for typical errors so that the request can be retried, but wait
+    # a short bit in the case of 429 — too many request — errors.
+    sleep 1 if too_many_requests?(e)
+    retry unless tries.zero?
+    invoke_error_handling_tasks(e, action, query)
+    raise e unless TYPICAL_ERRORS.include?(e.class)
```
ragesoss, line 112: I think the raising of unexpected errors could also be done in the error-handling module; this would probably mean passing in the relevant set of `TYPICAL_ERRORS` as an argument in the error handling code (or getting it dynamically from the class that the error handling method is being invoked from).

db/migrate/20200508211543_create_course_error_records.rb
```diff
@@ -0,0 +1,10 @@
+class CreateCourseErrorRecords < ActiveRecord::Migration[6.0]
+  def change
+    create_table :course_error_records do |t|
+      t.string :type_of_error
+      t.text :api_call_url
+
+      t.timestamps
```
ragesoss, line 7: Can you put all these into one migration? For a new table with no existing data to worry about, doing all the changes at once makes it easier to make sense of.

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/errors/course_update_error_handling"
+
+module ErrorHandling
+  include CourseUpdateErrorHandling
+
+  def perform_error_handling_tasks(error, level, extra, course, optional_params)
+    return report_exception_sentry(error, level, extra) unless course.present?
```
ragesoss, line 9: Doing this twice, either at the beginning or end, is a confusing control flow.

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,25 @@
...
+    sentry_tag_uuid = SecureRandom.uuid
```
ragesoss, line 10: I suggest using a uuid for every error, and then only doing the save_course_error_record if if the course is present.

lib/replica.rb
shashwatkathuria: WikiApi is not using this code but the OresApi and Replica are. So I refactored this into ErrorHandling module and added comment regarding the same.

lib/wiki_api.rb
shashwatkathuria: Done! Earlier only WikiApi was logging the error, so I thought it might be specific to it.

lib/errors/error_handling.rb
shashwatkathuria: I have refactored the course update part into `CourseUpdateErrorHandling` module. Also moved its sentry error reportinng part separately there.

lib/errors/error_handling.rb
shashwatkathuria: I have done something different, please have a look at the `def perform_error_handling` in `ErrorHandling` and `def perform_course_error_handling` in `CourseUpdateErrorHandling`. What I have done is actually momved all the part relating to `uuid` in `CourseUpdateErrorHandling` itself so that the control flow in the methods is not confusing as well as in the relevant modules.

lib/wiki_api.rb
shashwatkathuria: Done!

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def report_exception_sentry(error, level, extra)
+    Raven.capture_exception error, level: level, extra: extra
+  end
+
+  # Building optional params for Replica and OresApi which require
+  # response body for Oj::ParseError to be saved in CourseErrorRecord
+  # WikiApi does not have Oj::ParseError as a typical error, which is
+  # why it does not use this method
+  def build_optional_params(course, error, url, response_body)
```
ragesoss, line 35: I don't like coupling this to the two classes that would use "optional_params"; it means that ErrorHandling and Replica each have to "know something" about the internals of the other.

Instead, I recommend making all the data needed for this (url, response body) into optional named arguments of `perform_error_handling`, so that the only thing other classes need to "know" is the method signature of `perform_error_handling`.

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/errors/course_update_error_handling"
+
+module ErrorHandling
+  include CourseUpdateErrorHandling
+
+  def perform_error_handling(error, typical_errors, extra, course, optional_params)
```
ragesoss, line 8: This is probalby going to be easier to keep track of with keyword args.

lib/replica.rb
```diff
@@ -207,11 +211,10 @@ def compile_revision_ids_query(revisions)
...
-    level = 'warning' if TYPICAL_ERRORS.include?(error.class)
-    Raven.capture_exception error, level: level, extra: {
-      query: query, endpoint: endpoint, language: @wiki.language, project: @wiki.project
-    }
-    return nil
+
+  def invoke_error_handling(error, endpoint, query: nil, url: nil, response_body: nil)
+    extra = { query: query, endpoint: endpoint, language: @wiki.language, project: @wiki.project }
+    optional_params = build_optional_params(@course, error, url, response_body)
+    perform_error_handling(error, TYPICAL_ERRORS, extra, @course, optional_params)
```
ragesoss, line 218: This would definitely be nicer if the module could find TYPICAL_ERRORS for whichever class it has been included in, automatically, instead of need it as an explicit argument.

db/migrate/20200508211543_create_course_error_records.rb
shashwatkathuria: Done!

lib/replica.rb
shashwatkathuria: Done!

lib/errors/error_handling.rb
shashwatkathuria: Done!

lib/errors/error_handling.rb
shashwatkathuria: In `perform_error_handling`, the number of arguments were getting out of hand(around 7), so what I did was added an attribute `optional_params[:build]`, passed on `optional_params` to `CourseUpdateErrorHandling` module and then executed `build_optional_params` if `optional_params[:build]` was `true`. Also, I moved the method `build_optional_params` to the relevant module `CourseUpdateErrorHandling` and called that method in `save_course_course_record` where it was ultimately required.

lib/errors/course_update_error_handling.rb
```diff
@@ -20,4 +23,21 @@ def report_course_exception_sentry(error, level, extra, sentry_tag_uuid)
...
   end
+
+  # Building optional params for Replica and OresApi which require
+  # response body for Oj::ParseError to be saved in CourseErrorRecord
+  # WikiApi does not have Oj::ParseError as a typical error, which is
+  # why it does not use this method
+  def build_optional_params(error, optional_params)
+    # Does processing only if optional_params[:build] is true, which is
+    # true for Replica and OresApi
+    return optional_params unless optional_params[:build] == true
```
AmitJoki, line 34: You don't need to equate it with true. You can just do `unless optional_params[:build]`

AmitJoki:
I think the name `optional_params` can be a bit ambiguous. It can mean that it can be left out and it wouldn't change anything but I see that it is passed almost everywhere, which means it isn't optional.

I would suggest renaming it to just `params` or something which doesn't make it look like it could be left out.

Also I see lots of parameters being passed down several methods. I think it would be better if for instance of having `(error:, level:, extra:, course:, optional_params:)` as the function sigature, you could abstract these into a class, say, `ErrorRecord` which would have these fields and then you can pass that object around instead,, which, in my opinion would look cleaner.

AmitJoki:
I think the name `optional_params` can be a bit ambiguous. It can mean that it can be left out and it wouldn't change anything but I see that it is passed almost everywhere, which means it isn't optional.

I would suggest renaming it to just `params` or something which doesn't make it look like it could be left out.

Also I see lots of parameters being passed down several methods. I think it would be better if for instance instead of having `(error:, level:, extra:, course:, optional_params:)` as the function sigature, you could abstract these into a class, say, `ErrorRecord`  which would have these fields and then you can pass that object around instead,, which, in my opinion would look cleaner.

app/services/update_course_stats.rb
```diff
@@ -7,11 +7,17 @@
...
 
 #= Pulls in new revisions for a single course and updates the corresponding records
 class UpdateCourseStats
+  include CourseUpdateHelper
+
+  attr_reader :course, :sentry_tag_uuid
+
   def initialize(course, full: false)
     @course = course
+    @sentry_tag_uuid = generate_sentry_tag_uuid
```
ragesoss, line 20: This could be done in the module, so that all we need to do is include the module and then the update object will be able to respond to `:sentry_tag_uuid`.

lib/errors/course_update_helper.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+module CourseUpdateHelper
+  def generate_sentry_tag_uuid
+    SecureRandom.uuid
```
ragesoss, line 5: I suggest instead:

```
def sentry_tag_uuid
  @sentry_tag_uuid ||= SecureRandom.uuid
end
```

lib/errors/error_record.rb
```diff
@@ -0,0 +1,29 @@
...
+  attr_writer :level
+
+  def initialize(error, sentry_extra, update_course_stats)
+    @error = error
+    @sentry_extra = sentry_extra
+    @update_course_stats = update_course_stats
+    @level = nil
+  end
+
+  def course
```
ragesoss, line 14: I think we don't need to care about a `course` here at all now. Basically, it should be the job of the update service to figure out what error reporting behavior is wanted... and the pattern ought to apply to any kind of update, not just an `UpdateCourseStats`. We'll just start with the course updates, but this pattern may be useful for other updates as well.

lib/importers/course_upload_importer.rb
```diff
@@ -10,10 +10,11 @@ def self.update_courses(courses)
     end
   end
 
-  def initialize(course)
-    @course = course
-    @start = course.start
-    @end = course.end + Course::UPDATE_LENGTH
+  def initialize(update_cs)
```
ragesoss, line 13: I think this and the rest of the importers that currently take a `course` as their main argument should not be changed to take an update service; the update service should be an additional, optional keyword argument.

The idea here is that these importers are mainly concerned about a specific course and its data, and the optional context of the update service is there just for the purpose of logging.

lib/errors/course_update_helper.rb
shashwatkathuria: Done!

app/services/update_course_stats.rb
shashwatkathuria: Done!

lib/errors/error_record.rb
shashwatkathuria: I have made `ErrorRecord` more general and now the specific tasks of error handling is a part of the update object itself.

lib/course_revision_updater.rb
```diff
@@ -7,21 +7,21 @@ class CourseRevisionUpdater
...
+  def self.import_revisions(course, all_time:, update_cs: nil)
+    course = course
     return if course.students.empty?
-    new(update_cs).update_revisions_for_relevant_wikis(all_time)
+    new(course, update_cs: update_cs).update_revisions_for_relevant_wikis(all_time)
     ArticlesCourses.update_from_course(course)
   end
 
-  def initialize(update_cs)
+  def initialize(course, update_cs: nil)
```
ragesoss, line 17: All the cases of `update_cs` need a better name.

lib/errors/course_update_error_handling.rb
```diff
@@ -1,13 +1,24 @@
...
-                              course: error_record.course_slug
+                              course_update_id: sentry_tag_uuid,
+                              course: @course.slug
                             })
   end
+
+  def update_course_flags
+    @course.flags[:errors] ||= {}
+    @course.flags[:errors][sentry_tag_uuid] ||= 0
+    @course.flags[:errors][sentry_tag_uuid] += 1
```
ragesoss, line 22: I think we need to log the error counts and the sentry uuid as part of the update logging we already do (including only keeping the last 10 updates) instead of putting it in a separate `errors` key.

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,28 @@
...
+module ErrorHandling
+  def perform_error_handling(error_record)
+    Rails.logger.info "Caught #{error_record.error}"
+    error_record.level = error_level(error_record.error)
+    if error_record.specific_error_tasks?
+      error_record.perform_error_handling
+    else
+      report_exception_sentry(error_record)
+    end
+    return nil
```
AmitJoki, line 12: Is this `return nil` needed?

AmitJoki: Nevermind, I guess you're explicitly stating that this doesn't return anything.

lib/errors/course_update_error_handling.rb
shashwatkathuria: Okay, I forgot about that 😅

lib/course_revision_updater.rb
shashwatkathuria: I have renamed all of them to `update_object`

lib/errors/error_handling.rb
shashwatkathuria: Yes, you got it right, so that it doesn't return a `Raven` object. Earlier `WikiApi` and `Replica` both required this behaviour, now only `Replica` requires it. I thought it would be better to be explicit about it.

lib/errors/course_update_error_handling.rb
```diff
@@ -0,0 +1,26 @@
...
+    @error_count = error_count + 1
+    report_course_exception_sentry(error_record)
+  end
+
+  def report_course_exception_sentry(error_record)
+    Raven.capture_exception(error_record.error,
+                            level: error_record.level,
+                            extra: error_record.sentry_extra,
+                            tags: {
+                              course_update_id: sentry_tag_uuid,
```
ragesoss, line 22: Make this `update_service_id` so that it can be used for other types of updates as well.

lib/errors/course_update_error_handling.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module CourseUpdateErrorHandling
```
ragesoss, line 3: Maybe call this `UpdateServiceErrorLogging` This doesn't do any actual handling of errors, in terms of flow control... its only job is to make sure the errors get logged appropriately.

lib/errors/course_update_error_handling.rb
```diff
@@ -0,0 +1,26 @@
...
+  def sentry_tag_uuid
+    @sentry_tag_uuid ||= SecureRandom.uuid
+  end
+
+  def error_count
+    @error_count ||= 0
+  end
+
+  def perform_error_handling(error_record)
```
ragesoss, line 12: I'd call this `log_error`.

shashwatkathuria: Done!

lib/errors/course_update_error_handling.rb
shashwatkathuria: Done!

lib/errors/course_update_error_handling.rb
shashwatkathuria: Done!

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module ErrorHandling
```
ragesoss, line 3: This is specifically about handling API errors, so I would call it that: `ApiErrorHandling`.

shashwatkathuria: Done!

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,28 @@
...
+  def log_error(error_record)
+    Rails.logger.info "Caught #{error_record.error}"
+    error_record.level = error_level(error_record.error)
+    if error_record.specific_error_logging?
```
ragesoss: I really don't like the level of indirection happening with this module vs ErrorRecord vs the update object.

I'm not convinced that `ErrorRecord` gets us anything. I think the ideal situation is that with this module, there's basically just a single method that each API class needs to know (`log_error`). If we need to create an `ErrorRecord` (which I'm still open to), that could be one inside this method, instead of in the API-specific code that gets duplicated across API files.

lib/errors/error_handling.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def error_level(error)
+    self.class.const_get(:TYPICAL_ERRORS).include?(error.class) ? 'warning' : 'error'
+  end
+
+  def raise_unexpected_error(error)
+    raise error if self.class.const_get(:TYPICAL_ERRORS).exclude?(error.class)
+  end
+
+  def report_exception_sentry(error_record)
```
ragesoss: This duplicates what's going on in `UpdateServiceErrorLogging`. We should have only one method responsible for sending errors to Sentry, whether or not there is an update service or course to tag it with.

lib/ores_api.rb
```diff
@@ -57,8 +68,15 @@ def query_url(rev_ids)
   end
 
   def ores_server
-    conn = Faraday.new(url: 'https://ores.wikimedia.org')
+    conn = Faraday.new(url: ORES_SERVER_URL)
     conn.headers['User-Agent'] = ENV['dashboard_url'] + ' ' + Rails.env
     conn
   end
+
+  def invoke_error_handling(error, url, response_body)
```
ragesoss: This method is still very similar across APIs. I think it could be done such that a single method (maybe rename `log_error` to `handle_api_error`) with optional keyword arguments is used across all these APIs, and none of them define their own `invoke_error_handling` method, they just call the one provided by the module with whatever arguments are relevant for this particular API.

lib/errors/error_handling.rb
shashwatkathuria: Now that the number of arguments are only three, `error`, `sentry_extra` and `update_service`, they can be passed into this method easily. After passing the parameters, I guess `ErrorRecord` won't be of as much help as it was earlier, should I remove it or not?

lib/ores_api.rb
shashwatkathuria: The reasons why I have made this is for passing of arguments. `sentry_extra` is different for all the three APIs, and also `invoke_error_handling` calls in Replica pass different arguments in 2 places. I think moving the method would have the same effect, and there could be problems with passing of the high number of optional arguments

lib/errors/error_handling.rb
shashwatkathuria: Okay, I'll do that

lib/ores_api.rb
shashwatkathuria: Or I could do it similar to how`UpdateLogger.update_course` is called? (Talking about the way `times` argument is passed there)

ragesoss: Yes, if all the extra optional data is an arbitrary hash of stuff to pass in to sentry, collecting it all up as an implicit single argument hash (like with `UpdateLogger.update_course`) makes perfect sense.

lib/errors/api_error_handling.rb
```diff
@@ -1,14 +1,17 @@
...
+    sentry_tags = nil
+    if update_service.present?
+      update_service.update_error_stats
+      sentry_tags = update_service.sentry_tags
     end
+    Raven.capture_exception(error,
+                            level: error_level(error),
+                            extra: sentry_extra,
+                            tags: sentry_tags)
     return nil
```
ragesoss, line 15: Looks like this now skips the `raise_unexpected_error`?

shashwatkathuria: raise_unexpected_error is not required by Replica so I have called it separately(fid the same way earlier too)

ragesoss: Oh, I see... that gets done in the inidividual API classes...

ragesoss: I think it makes would makes sense to get rid of that raising behavior altogether. Part of the point here is that we want to make errors easier for end-users to find out about, and if we raise an error and kill the update service process, then the error won't get logged in the course flags and will be harder to find. If we needed to retain the current behavior, where some files can raise an unexpected error but some cannot (eg, `Replica`), I think would make sense to configure whether or not to raise the errors using a configuration constant in each class (`RAISE_UNEXPECTED_ERRORS`). But I think removing that behavior makes more sense.

PR #4042, opened 2020-05-19
Adds md5 fingerprint for i18n

AmitJoki:


ragesoss:
This works perfectly. The md5 digest is fast enough that it won't have any appreciable affect on server load (my system can do that digest about 10,000 times per second), and it works the same in the the dev environment (where the i18n files get regenerated on pageload if the translation files changed).

ragesoss:
This works perfectly. The md5 digest is fast enough that it won't have any appreciable affect on server load (my system can do that digest about 10,000 times per second), and it works the same in the the dev environment (where the i18n files get regenerated on pageload if the translation files changed).

app/views/shared/_head.html.haml
```diff
@@ -11,7 +11,7 @@
 = content_for :head
 
 = javascript_include_tag '/assets/javascripts/i18n'
-= javascript_include_tag '/assets/javascripts/i18n/' + I18n.locale.to_s
+= javascript_include_tag '/assets/javascripts/i18n/' + I18n.locale.to_s + "?v=#{i18n_md5_hash(I18n.locale.to_s)}"
```
ragesoss, line 14: We use this in one other view as well, the stats embedding view that you created a while ago.

Maybe the whole thing should be turned into a helper method like `i18n_javascript_tag` so we can use the same pattern both places without duplication.

AmitJoki:
I have rebased this PR @ragesoss

PR #4046, opened 2020-05-21
Difference viewer spec.

shashwatkathuria:
## What this PR does
Feature spec for difference viewer.

### DIFF
--- a/spec/features/diff_viewer_spec.rb
+++ b/spec/features/diff_viewer_spec.rb
@@ -3,30 +3,33 @@
 require 'rails_helper'
 
 describe 'difference viewer', type: :feature, js: true do
-  let(:course) { create(:course, start: '2013-10-01', end: '2014-10-01') }
-  let(:user) { create(:user, username: '66.45.13.253') }
+  let(:course) { create(:course, start: '2013-02-07', end: '2014-01-01') }
+  let(:user) { create(:user, username: 'Greatgavini') }
 
-  it 'checks if renders details properly' do
+  before do
     create(:courses_user, course_id: course.id, user_id: user.id)
 
     # Updating course
-    login_as user
     VCR.use_cassette 'diff_viewer/course_update' do
-      visit "/courses/#{course.slug}/manual_update"
+      UpdateCourseStats.new(course, full: true)
     end
+  end
 
-    # User has 6 contributions since 24th Oct 2013
-    # https://en.wikipedia.org/wiki/Special:Contributions/66.45.13.253
+  it 'checks if renders details properly' do
     visit "/courses/#{course.slug}/activity"
-    expect(page).to have_css('button.icon-diff-viewer', count: 6)
 
-    # Checking length of difference, difference of characters (which in this case
-    # is ' lilah', should appear only once) and a small portion of unchanged text,
-    # which should appear two times (i.e., in both the columns of diff viewer)
-    # https://en.wikipedia.org/w/index.php?title=Heaven%27s_Gate_(religious_group)&diff=prev&oldid=578607606
+    # User has 3 contributions since 7th Feb 2013
+    # https://en.wikipedia.org/wiki/Special:Contributions/Greatgavini
+    expect(page).to have_css('button.icon-diff-viewer', count: 3)
+
+    # Checking length of difference, a small portion of changed text (which in this case
+    # is 'They see ash run into', should appear only once) and a small portion of unchanged text
+    # (which in this case is 'Although Misty is the leader'), which should appear two
+    # times (i.e., in both the columns of diff viewer)
+    # https://en.wikipedia.org/w/index.php?diff=prev&oldid=554568302
     all('button.icon-diff-viewer').last.click
-    expect(page).to have_content('6 Chars Added')
-    expect(page).to have_content(' lilah').once
-    expect(page).to have_content('Jacques Vallée').twice
+    expect(page).to have_content('-1299 Chars Added')
+    expect(page).to have_content('They see ash run into').once
+    expect(page).to have_content('Although Misty is the leader').twice
   end
 end
--- a/spec/features/diff_viewer_spec.rb
+++ b/spec/features/diff_viewer_spec.rb
@@ -1,47 +1,32 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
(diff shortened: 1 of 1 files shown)


spec/features/diff_viewer_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+        build(:user, username: 'I enjoy sandwiches')
+      ]
+      rev_start = 2016_04_17_003430
+      rev_end = 2016_04_18_003430
+
+      response = Replica.new(en_wiki).get_revisions(all_users, rev_start, rev_end)
+
+      response[article.mw_page_id.to_s]['revisions'].each_with_index do |revision, index|
+        create(:revision,
+               id: index,
```
ragesoss, line 27: Don't hard-code record IDs in specs unless it's really necessary.

spec/features/diff_viewer_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+               mw_rev_id: revision['mw_rev_id'],
+               date: revision['date'],
+               characters: revision['characters'].to_i,
+               wiki_id: revision['wiki_id'],
+               article_id: article.id,
+               new_article: revision['new_article'],
+               system: revision['system'],
+               deleted: false,
+               user_id: user.id)
+      rescue ActiveRecord::RecordNotUnique
```
ragesoss, line 37: This should never be necessary in a spec.

spec/features/diff_viewer_spec.rb
shashwatkathuria: Okay..Removed

spec/features/diff_viewer_spec.rb
shashwatkathuria: Okay..Used `uniq` instead...I was actually getting duplicate values in the response.

spec/features/diff_viewer_spec.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require "#{Rails.root}/lib/replica"
+
+describe 'diff viewer', type: :feature, js: true do
+  let(:en_wiki) { Wiki.get_or_create(language: 'en', project: 'wikipedia') }
+  let(:course) { create(:course, id: 1, start: '2013-01-01', end: '2014-01-02') }
```
ragesoss, line 8: No need for the course ID here. In most cases, you shouldn't need any hard-coded record IDs at all in a spec.

ragesoss: What's with these course dates? They don't seems to have anything to do with the revisions you're looking for in the spec.

spec/features/diff_viewer_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+  let(:user) { create(:user, id: 1, username: 'I enjoy sandwiches') }
```
ragesoss, line 9: It would be useful to add some documentation to explain that what you're trying to do here... looking for actual revisions made by this user to a specific page, in particular https://en.wikipedia.org/w/index.php?title=Periodontium&diff=715718156&oldid=714831980&diffmode=source

spec/features/diff_viewer_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+      login_as user
+      all_users = [
+        build(:user, username: 'I enjoy sandwiches')
+      ]
+      rev_start = 2016_04_17_003430
+      rev_end = 2016_04_18_003430
+
+      response = Replica.new(en_wiki).get_revisions(all_users, rev_start, rev_end)
+
+      response[article.mw_page_id.to_s]['revisions'].uniq.each do |revision|
```
ragesoss, line 25: It looks like you're basically recreating revisions that would get created via a course update process normally. This is confusing and couples this test far too tightly to `Replica` and the revision importer, which it has nothing to do with.

I suggest either just hard-coding the data you care about so that the test doesn't need to refetch it from Replica each time, or setting up this test so that you can do a UpdateCourseStats and it will pull in the revisions you need based on the dates and the enrolled user.

If you go the latter route, this could be a good starting point for other tests of the more interactive frontend features as well.

spec/features/diff_viewer_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+               user_id: user.id)
+      end
+
+      create(:courses_user, id: 1, course_id: 1, user_id: 1)
+      create(:articles_course, id: 1, course: course, article: article, tracked: true)
+
+      visit "/courses/#{Course.first.slug}/activity"
+      expect(page).to have_css('button.icon-diff-viewer', count: 3)
+      all('button.icon-diff-viewer').last.click
+      expect(page).to have_content '-69 Chars Added'
```
ragesoss, line 44: If I understand this test correctly, this `-69 Chars Added` is what renders immediately in the DiffViewer, even before the full diff from the Wikipedia API is received.

You should also add an expectation that looks for the actual contents of the diff.

shashwatkathuria: Done!

spec/features/diff_viewer_spec.rb
shashwatkathuria: Did thorugh the `UpdateCourseStats` route !

spec/features/diff_viewer_spec.rb
shashwatkathuria: Done!

spec/features/diff_viewer_spec.rb
shashwatkathuria: Removed the unnecessary stuff

ragesoss:
This looks better.

Using an IP editor is not ideal, as it's not currently possible to add an IP user as a student. Some of the system works just fine by putting an IP address as the username, but it will throw an error via the frontend if you try this because it confirms that an account exists before adding the user.

I'd also suggest not using `/manual_update` via the browser. Instead, you can run `UpdateCourseStats` in a `before` block, just to keep the setup more isolated from the feature being tested.

shashwatkathuria:
Okay, I have made the required changes.

shashwatkathuria:
Okay, I have made the required changes.

ragesoss:
Looks great. This spec takes about 10 second with an empty cassette (3 seconds with a recorded one), but I think that's okay.

PR #4052, opened 2020-05-28
Fixes javascript cross site warnings.

shashwatkathuria:
## What this PR does
There are 2 ways to make an API call to another MediaWiki website: 
- JSONP
- CORS

JSONP is more of a hack which turns data into scripts to bypass the cross domain restrictions, it can also cause Cross-Site Scripting issues if the site to which request is being made is compromised.

In CORS, new HTTP headers provide a way to request remote URLs only when they have permission. The server responds appropriately whether or not it wants credentials, after which the actual request is made. In the MediaWiki API case, `origin=*` (unauthenticated requests) responds with `Access-Control-Allow-Credentials: false` , meaning that it will process the request as if logged out.

Also one thing is to be kept in mind that JSONP is an old method now and CORS is a new modern solution. JSONP works with very old browsers nicely whereas CORS can cause some issues. CORS works well with the modern browsers.

Referencing #3998 
Referencing [https://www.mediawiki.org/wiki/API:Cross-site_requests#Unauthenticated_CORS_Requests](https://www.mediawiki.org/wiki/API:Cross-site_requests#Unauthenticated_CORS_Requests)

The error in our case was
Screenshot from 2020-05-28 03-50-24

### DIFF
--- a/app/assets/javascripts/actions/new_account_actions.js
+++ b/app/assets/javascripts/actions/new_account_actions.js
@@ -5,8 +5,7 @@ import API from '../utils/api.js';
 const _checkAvailability = (newAccount) => {
   return new Promise((res, rej) =>
     $.ajax({
-      dataType: 'jsonp',
-      url: `https://meta.wikimedia.org/w/api.php?action=query&list=users&ususers=${newAccount.username}&usprop=cancreate&format=json`,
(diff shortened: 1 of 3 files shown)


ragesoss:
LGTM

PR #4054, opened 2020-05-29
add parallel webpack

AmitJoki:


ragesoss:
This is is a major improvement of the build speed. Now JS changes take half a second or less. (It varies pretty widely; seems like there's some speedup after the first several save/emit cycles.

CSS changes are also faster, now taking under 2 seconds.

When I save a change to a .styl file, the terminal output incorrectly describes the build as .js instead of .css. Ie, when I save a stylesheet change, it says:
```
Finished building training.[chunkhash].js within 1.031 seconds
```

prepare.js
```diff
@@ -0,0 +1,23 @@
...
+  from: `${config.sourcePath}/${config.fontsDirectory}`,
+  to: `${config.outputPath}/${config.fontsDirectory}`,
+}, {
+  from: './node_modules/tinymce/skins',
+  to: `${config.outputPath}/${config.jsDirectory}/skins`,
+}];
+
+copyPaths.forEach(entry =>
+  execSync(`cp -r ${entry.from} ${entry.to}`));
+
```
ragesoss, line 23: Some terminal output for this would be nice, just to note that the task completed and give a quick description of what it's for.

AmitJoki: Sure!

webpack.multi.js
```diff
@@ -0,0 +1,181 @@
...
+];
+
+const cssEntries = [
+  { main: [`${cssSource}/main.styl`] },
+  { styleguide: [`${cssSource}/styleguide.styl`] },
+  { surveys: [`${cssSource}/surveys.styl`] },
+  { training: [`${cssSource}/training.styl`] },
+];
+
+const cssConfig = env => ({
```
ragesoss, line 37: This file has a lot going on, in terms of both composing the separate sets of entries for parallel processing and also configuring sets of plugins for CSS vs JS.

Can you refactor this to separate the concerns more fully? Ideally, there would be one file that we can look at to understand the JS configuration, another for the CSS configuration, and this file would just be the glue to make them work with parallel-webpack.

AmitJoki:
@ragesoss the css-loader does emit a trivial javascript file which is used when we do 

```
import 'path/to/stylesheet.css';
```

within JS files. There's no way to disable that feature for now.

I have filed an issue in the corresponding loader. But unfortunately we will have to make do with it for now.

ragesoss:
Okay... is there something else that can be printed to the console to make it clear that a css file (also) got emitted?

AmitJoki:
@ragesoss I have made it so that the extension of the js files emitted is `.styl` so we can differentiate it.

AmitJoki:
And I have added new stats for the `prepare.js`. Check it out.

ragesoss:
This is very nice. I'm going to merge now. I still think it would make sense to refactor the configs, per my note above, but that could be done in a later PR.

Is there any way to get rid of the `[chunkhash]` in the output, or have it use the actual hash? Right now it says ` Finished building main.[chunkhash].js within 7.832 seconds`.

AmitJoki:
ah I didn't see your comment on the refactor. I just saw it now. I'll do it in a separate PR later.

PR #4055, opened 2020-05-31
changed Ruby version from 2.5.0 to 2.6.6 in gemfile to make Oj and Bo…

PANKAJ-DAHIYA123:
…otsnap gems compatible in windows platform


ragesoss:
We've done the upgrade to Ruby 2.7.1 now, which hopefully unblocks Windows compatibility.

PR #4059, opened 2020-06-02
Lodash es

AmitJoki:
Migrates from CommonJS `lodash` to ES Module `lodash-es`.


ragesoss:
It will be good to include a bundle size before/after for each of these, so we can see how much of an effect it's having. (I assume since this is just adding lodash-es but not completely removing lodash, there's no improvement yet?)

AmitJoki:
@ragesoss yes. Also I haven't completely replaced all of them. Once I completely remove it, I will compare the stats with master and with this branch and put that info here.

AmitJoki:
We have reduced roughly 50% of the lodash bundle size from **42 KB to 21 KB.**

package.json
```diff
@@ -156,5 +137,6 @@
   },
   "resolutions": {
     "babel-core": "7.0.0-bridge.0"
-  }
+  },
+  "sideEffects": false
```
ragesoss, line 141: Please explain what this does (ideally, in an updated commit message).

webpack.config.js
```diff
@@ -51,6 +52,12 @@ module.exports = (env) => {
     path: ['^.*css.*\\.js$']
   }));
 
+  plugins.push(new LodashModuleReplacementPlugin({
```
ragesoss, line 55: Please explain what these do (ideally, in an updated commit message).

PR #4069, opened 2020-06-14
Migrating away from momentjs

AmitJoki:


ragesoss:
@AmitJoki this is a dead end, right? Could you open an issue summarizing the blockers for replacing momentjs as we currently use it?

AmitJoki:
The main motive to migrate to `dayjs` was the fluff and the unneeded locale files that was bloating the size of momentjs. Both the fluff and the locale files have been removed and moreover `charts.js` and `moment.recur` have `moment` as their dependency as well so this PR was abandoned.

PR #4076, opened 2020-06-16
Tracking course update errors downwards

shashwatkathuria:
## What this PR does
Implementing tracking of course update errors by passing update service `UpdateCourseStats` downwards to APIs and keeping track of errors with the help of Sentry and error count with the help of `UpdateLogger` and `course.flags`.

### DIFF
--- a/lib/data_cycle/update_logger.rb
+++ b/lib/data_cycle/update_logger.rb
@@ -4,9 +4,9 @@ class UpdateLogger
   ################
   # Entry points #
   ################
-  def self.update_course(course, times)
+  def self.update_course(course, hash_info)
     logs = course.flags['update_logs']
-    new_logs = new(logs).update(times)
+    new_logs = new(logs).update(hash_info)
     course.flags['update_logs'] = new_logs
     course.flags['average_update_delay'] = average_delay(new_logs)
     course.save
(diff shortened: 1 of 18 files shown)


lib/data_cycle/update_logger.rb
```diff
@@ -4,9 +4,9 @@ class UpdateLogger
   ################
   # Entry points #
   ################
-  def self.update_course(course, times)
+  def self.update_course(course, hash_info)
```
ragesoss, line 7: Can you think of a better name than `hash_info`? I tried to leave a comment before but it might not have gone through...

I think you could call this `new_logs`, and then rename the intermediate variable below to `updated_logs`. That makes it a little less generic and reduces the chance for someone getting confused about what this is supposed to do.

lib/errors/api_error_handling.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module ApiErrorHandling
+  def log_error(error, update_service: nil, sentry_extra:)
+    Rails.logger.info "Caught #{error}"
+    sentry_tags = nil
+    if update_service.present?
+      update_service.update_error_stats
```
ragesoss, line 8: This behavior isn't being tested at all (and shows up as uncovered code after a spec run).

You should add appropriate tests, perhaps to `update_course_stats_spec.rb`, to verify that it does what we expect in terms of posting something to Sentry and including the right error count in the `flags` update logs.

`constant_update_spec.rb` also has some relevant examples, in terms of interacting with Sentry behavior in a test.

lib/data_cycle/update_logger.rb
shashwatkathuria: Done!

lib/errors/api_error_handling.rb
shashwatkathuria: Done!

PR #4080, opened 2020-06-19
Course update errors UI

shashwatkathuria:
## What this PR does
UI implementation for #4076 

### If errors exist - Outreach Basic Course
Peek 2020-06-30 02-02

### If errors exist - Wiki Education Article Scoped Program
Peek 2020-06-30 02-05

### No error
Peek 2020-06-30 02-04

### DIFF
--- a/app/assets/javascripts/components/overview/statistics_update_info.jsx
+++ b/app/assets/javascripts/components/overview/statistics_update_info.jsx
@@ -110,7 +110,7 @@ const StatisticsUpdateInfo = createReactClass({
       );
     }
 
-    // Render update time information along with 'See More' button to open modal
+    // Render update time information along with 'See More' link to open modal
     return (
       <div className="course-data-update pull-right mb2">
         <small>
--- a/app/assets/javascripts/components/overview/statistics_update_info.jsx
+++ b/app/assets/javascripts/components/overview/statistics_update_info.jsx
@@ -17,33 +17,35 @@ const StatisticsUpdateInfo = createReactClass({
     };
   },
 
-  getUpdateTimesMessage() {
+  getUpdateTimesInformation() {
     const course = this.props.course;
-
     const lastUpdate = course.updates.last_update.end_time;
     const lastUpdateMoment = moment.utc(lastUpdate);
     const averageDelay = course.updates.average_delay;
     let lastUpdateMessage = '';
+    let nextUpdateMessage = '';
+    let isNextUpdateAfter = null;
+
     if (lastUpdate) {
       lastUpdateMessage = `${I18n.t('metrics.last_update')}: ${lastUpdateMoment.fromNow()}`;
-    }
-
-    const nextUpdateExpectedTime = lastUpdateMoment.add(averageDelay, 'seconds');
-    let nextUpdateMessage = '';
-    if (nextUpdateExpectedTime.isAfter()) {
+      const nextUpdateExpectedTime = lastUpdateMoment.add(averageDelay, 'seconds');
+      isNextUpdateAfter = nextUpdateExpectedTime.isAfter();
       nextUpdateMessage = `${I18n.t('metrics.next_update')}: ${nextUpdateExpectedTime.fromNow()}`;
     }
 
-    return `${lastUpdateMessage}. ${nextUpdateMessage} `;
+    return [lastUpdateMessage, nextUpdateMessage, isNextUpdateAfter];
+  },
+
+  getTotalNumberOfUpdates() {
+    const updateNumbers = Object.keys(this.props.course.flags.update_logs);
+
+    return updateNumbers[updateNumbers.length - 1];
   },
 
   getCourseUpdateErrorMessage() {
-    let courseUpdateErrorMessage = '';
     const errorCount = this.props.course.updates.last_update.error_count;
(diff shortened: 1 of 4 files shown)


ragesoss:
That 'See More' link is way too intrusive. It should just be part of the text, which should remain in one line as with the current version.

current stats UI

For the modal, I suggest an opaque white background and smaller size.

config/locales/en.yml
```diff
@@ -865,9 +865,6 @@ en:
     label: Course Activity
     last_update: Last statistics update
     next_update: Next estimated update
-    update_statistics_link: See More
-    close_modal: Close
-    error_count_message: We encountered %{error_count} errors while updating your course.
```
ragesoss: "The dashboard encounted X error(s) duing the most recent update."

app/assets/javascripts/components/overview/statistics_update_info.jsx
```diff
@@ -1,27 +1,124 @@
...
+      return (
+        <div className="pull-right mb2">
+          <small>{ updateTimesMessage }</small>
+        </div>
+      );
+    }
+
+    // If there are errors
+
+    // Render Modal
```
ragesoss, line 80: There's too much going on in this component. The modal should be extracted as a separate component.

config/locales/en.yml
```diff
@@ -873,19 +873,36 @@ en:
     articles_edited: Articles Edited
     articles_edited_generic: Pages Edited
     articles_edited_wikidata: Items Edited
+    article_scoped_program_info:
+        Your course is an 'Article Scoped Program', in which only assigned articles and categories
```
ragesoss, line 877: This is an 'Article Scoped Program', which only counts edits to assigned articles and categories.

config/locales/en.yml
```diff
@@ -873,19 +873,36 @@ en:
...
     revisions: Recent Edits
     references_count: References Added
     references_doc: This is the number of reference tags added to articles, and can include multiple references to the same source. The data comes from the ORES article quality model, and is only available for some languages.
+    replag_info: >
+        One possibility of missing data might be replag. Sometimes our replica databases 
+        lag a bit behind the production databases. It might be possible that your course 
+        has not updated properly due to the lag. You can click on the below link to check 
+        the status of our replica databases. Slice s1-s8 refers to the production masters 
+        and the lag time is shown in both, seconds and time.
+    replag_link: Database Status
```
ragesoss, line 901: 'Check replag'

config/locales/en.yml
```diff
@@ -873,19 +873,36 @@ en:
...
     tags: Tags
+    total_updates: Total number of updates till now
+    update_status_heading: Update status
+    update_statistics_link: See More
```
ragesoss, line 905: 'See more' (lower case m)

config/locales/en.yml
```diff
@@ -873,19 +873,36 @@ en:
...
+    error_count_message: We encountered %{error_count} errors while updating your course.
     last_reviewed: Last Reviewed
+    missing_data_heading: Missing or unexpected results?
+    outreach_help: Need Help? Refer to 'Report a problem' in the navigation bar.
     per_user: '%{number}/user'
     revisions: Recent Edits
     references_count: References Added
     references_doc: This is the number of reference tags added to articles, and can include multiple references to the same source. The data comes from the ORES article quality model, and is only available for some languages.
+    replag_info: >
+        One possibility of missing data might be replag. Sometimes our replica databases
```
ragesoss, line 896: The Dashboard retrieves edit data from WMF Cloud replica databases. Sometimes these databases are missing the most recent edits because of replication lag (replag). When this is the case, the missing edits will show up later on.

config/locales/en.yml
shashwatkathuria: Done!

app/assets/javascripts/components/overview/statistics_update_info.jsx
shashwatkathuria: Done!

config/locales/en.yml
shashwatkathuria: Done!

config/locales/en.yml
shashwatkathuria: Done!

config/locales/en.yml
shashwatkathuria: Done!

config/locales/en.yml
shashwatkathuria: Done!

ragesoss:
On a wider screen , the modal is much too large.

wider

Can the scrollbar on the right of the modal be removed (at least, if it's not sized such that scrolling is needed)?

I think the 'missing data' section will be clearer if it is formatted as bullets, like:

## Missing or unexpected results?

Common issues include:
* replag explanation
* article scoped program explanation

shashwatkathuria:
I have removed the `Modal` and defined a different css class, as it was not really helping now and the css was getting different. The scrolling part doesn't exist now, and the user can scroll down the page even if the modal is bigger in height than the screen, which wasn't possible earlier when I was using `Modal` which was kind of stuck if no scroll bar was there in such a situation, though that could be easily solved by `overflow-y: auto`.

PR #4083, opened 2020-06-23
remove locale files

AmitJoki:
Momentjs by default, includes nearly 193 locale files. Those are only used if we use the methods:
```
moment.locale()
moment.tz()
```
or if we import certain locales in the codebase, like, if we were to import Russian locale, it would be
```
require('moment/locales/ru')
```
The locale files combine to form nearly 200KB of the bundle size. We don't use the locale files at all in the codebase, so we wouldn't need them. This PR excludes such unnused locale files, thereby bringing 250KB MomentJS bundle to ~54KB.

### DIFF
--- a/js.webpack.js
+++ b/js.webpack.js
@@ -84,8 +84,13 @@ module.exports = (env) => {
     ],
     optimization: {
       splitChunks: {
-        chunks: chunk => !/tinymce|charts/.test(chunk.name),
-        name: 'vendors'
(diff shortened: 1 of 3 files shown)


ragesoss:
Can you explain in the PR description a bit more about exactly what this does why it won't have any effects on the behavior of the system?

AmitJoki:
@ragesoss updated the PR description.

PR #4088, opened 2020-06-28
dynamic load on DOMContentLoaded

AmitJoki:
The `DOMContentReady` listener in `main.js` is supposed to load a bunch of dependencies upon page load. But what currently happens is, Webpack just sees the `require('...')` and includes them right in the `main.js`

So, the event handler is noop. This PR rectifies that so `main.js` actually dynamically loads the dependencies when the actual DOMContentReady event is fired.

This drastically reduces the amount of resources that is loaded initially.

### DIFF
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -11,13 +11,17 @@ require('location-origin');
 require('@rails/ujs').start(); // Enables rails-ujs, which adds JavaScript enhancement to some Rails views
 
 document.addEventListener('DOMContentLoaded', () => {
-  window.I18n = require('i18n-js');
-  window.List = require('list.js'); // List is used for sorting tables outside of React
-  require('./utils/course.js'); // This adds jquery features for some views outside of React
-  // This is the main React entry point. It renders the navbar throughout the app, and
-  // renders other components depending on the route.
-  require('./components/app.jsx');
-  require('events').EventEmitter.defaultMaxListeners = 30;
-  require('./utils/editable.js');
-  require('./utils/users_profile.js');
+  require.ensure(
+    ['i18n-js', 'list.js', './utils/course.js', './components/app.jsx', 'events', './utils/editable.js', './utils/users_profile.js'],
+    (require) => {
+      window.I18n = require('i18n-js');
+      window.List = require('list.js'); // List is used for sorting tables outside of React
+      require('./utils/course.js'); // This adds jquery features for some views outside of React
+      // This is the main React entry point. It renders the navbar throughout the app, and
+      // renders other components depending on the route.
+      require('./components/app.jsx');
+      require('events').EventEmitter.defaultMaxListeners = 30;
+      require('./utils/editable.js');
+      require('./utils/users_profile.js');
+    });
 });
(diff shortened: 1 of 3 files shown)


ragesoss:
Looks like this breaks survey-admin.js?

The failing specs are all related to survey admin features.

AmitJoki:
Yes probably some race condition I guess. I will look into it.

ragesoss:
Can you explain more fully how this will change the behavior when a page is loaded?

Is `require.ensure()` being used because we can't use `import`?

I looked at the documentation, but it's confusing to me: https://webpack.js.org/api/module-methods/#requireensure

I don't think I understand the what 'bundle' means vs. 'entry', with webpack.

AmitJoki:
The code before did not actually load those dependencies upon the firing of the DOMContentLoaded event. So

```
document.addEventListener('DOMContentLoaded', () => {
  window.I18n = require('i18n-js');
  window.List = require('list.js'); // List is used for sorting tables outside of React
  require('./utils/course.js'); // This adds jquery features for some views outside of React
  // This is the main React entry point. It renders the navbar throughout the app, and
  // renders other components depending on the route.
  require('./components/app.jsx');
  require('events').EventEmitter.defaultMaxListeners = 30;
  require('./utils/editable.js');
  require('./utils/users_profile.js');
});
```

was essentially this:

```
window.I18n = require('i18n-js');
window.List = require('list.js'); // List is used for sorting tables outside of React
require('./utils/course.js'); // This adds jquery features for some views outside of React
// This is the main React entry point. It renders the navbar throughout the app, and
// renders other components depending on the route.
require('./components/app.jsx');
require('events').EventEmitter.defaultMaxListeners = 30;
require('./utils/editable.js');
require('./utils/users_profile.js');
```

Which made main.js bundle quite heavy as it included every one of the above dependency in itself. The event listener was a no-op.

Now, what we have is one that actually works. `main.js` is very small now and only contains the loader code. Basically the dependencies inside the `require.ensure` are bundled into a separate file and `main.js` just has the code to load it. It uses promises.

If you see the Networks tab, you can see main.js is just around 3 KB. The required dependencies are now actually loaded when the DOMContentLoaded event is fired.

AmitJoki:
The `require.ensure` is basically import. ESLint complains about `import` keyword so we would have to have a `babel-eslint` eslint parser which would recognize the `import` keyword.

ragesoss:
Ah, so this results in an additional javascript file (`12.[hash].js`) which basically contains all of the non-vendor code that was previously part of main.js. So this is the first stage of webpack code-splitting, and we can take it further by using require.ensure (or import, with the plugin for eslint support) elsewhere in the main.js dependency tree to make it split that 12.js file up further, and make it not even load some of the component code unless/until a user needs it.

Am I understanding that correctly?

ragesoss:
What are the downsides to adding babel-eslint, aside from just adding an additional dependency? It looks like it's well-supported and considered part of babel core, so I don't object to adding the dependency if you think it makes sense otherwise.

AmitJoki:
@ragesoss I didn't do extensive testing but the builds seemed slower with babel-eslint.  And yes, that's pretty much the gist of it.

PR #4099, opened 2020-07-01
[WIP] Orphan lock course update

shashwatkathuria:
## What this PR does
Removing orphan locks occuring in `CourseDataUpdateWorkers`

### Expectation: To display next estimated only modal, message about successful update, Last 10 updates summary
Screenshot from 2020-07-16 21-52-06
Screenshot from 2020-07-16 21-52-12

### Expectation: To display next estimated in both places, message about update errors, Last 10 updates summary
Screenshot from 2020-07-16 21-50-58
Screenshot from 2020-07-16 21-51-02

### DIFF
--- a/app/services/check_course_jobs.rb
+++ b/app/services/check_course_jobs.rb
@@ -76,16 +76,15 @@ class CheckCourseJobs
     # Extracting the latest update end time by getting the last element
     # in the values of update logs which are filtered by no orphan lock logs
     last_update_times_log = update_logs&.values
-                                       &.select { |element| element['orphan_lock_failure'].nil? }
+                                       &.select { |element| element[:orphan_lock_failure].nil? }
                                        &.last
-    last_end_time = last_update_times_log['end_time']
 
-    # If there is no log having update times means all are orphan lock logs
-    return true unless last_end_time.present?
+    # If we cannot find a log having update times means all are orphan lock logs
+    return true unless last_update_times_log.present?
 
-    # Return true only if the last update time is bigger than last update time + one day,
+    # Return true only if the last update end time is bigger than last update time + one day,
     # as currently most update jobs would end in a few minutes at max
-    Time.zone.now > last_end_time + 1.day
+    Time.zone.now > last_update_times_log['end_time'] + 1.day
   end
 
   def find_queued_job
--- a/app/services/check_course_jobs.rb
+++ b/app/services/check_course_jobs.rb
@@ -6,9 +6,27 @@ require_dependency "#{Rails.root}/lib/data_cycle/course_queue_sorting"
 class CheckCourseJobs
   include CourseQueueSorting
 
+  def self.remove_orphan_locks(courses_to_update)
+    orphan_lock_courses = []
+    courses_to_update.each do |course|
+      check_course_job = new(course)
+      is_removed = check_course_job.delete_orphan_lock
+      orphan_lock_courses << check_course_job.sentry_extra if is_removed
+    end
+
+    orphan_lock_count = orphan_lock_courses.length
+    return unless orphan_lock_count.positive?
+
+    Raven.capture_message("#{orphan_lock_count} Orphan lock(s) removed",
+                          level: 'warn',
+                          extra: { courses: orphan_lock_courses })
+    return nil
+  end
+
   def initialize(course)
     @course = course
     @course_id = course.id
+    @queue = queue_for(course)
   end
 
   def health_report
@@ -30,7 +48,7 @@ class CheckCourseJobs
   def expected_digest
     hash = {
       'class' => 'CourseDataUpdateWorker',
-      'queue' => queue_for(@course),
+      'queue' => @queue,
       'unique_args' => [@course_id]
     }.to_json
     digest = OpenSSL::Digest::MD5.hexdigest hash
@@ -46,5 +64,17 @@ class CheckCourseJobs
     end
 
     return nil
+  end
+
+  def delete_orphan_lock
+    if find_job.nil? && SidekiqUniqueJobs::Digests.all.include?(expected_digest)
+      delete_unique_lock
+      return true
+    end
+    return false
+  end
+
+  def sentry_extra
+    { course: @course.slug, queue: @queue }
   end
 end
--- a/app/services/check_course_jobs.rb
+++ b/app/services/check_course_jobs.rb
@@ -56,11 +56,14 @@ class CheckCourseJobs
   end
 
   def find_job
-    Sidekiq::Queue.all.each do |queue|
-      queue.each do |job|
-        next unless job.klass == 'CourseDataUpdateWorker'
-        return job if job.args == [@course_id]
-      end
+    Sidekiq::Queue.new(@queue).each do |job|
+      next unless job.klass == 'CourseDataUpdateWorker'
+      return job if job.args == [@course_id]
+    end
+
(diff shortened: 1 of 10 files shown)


app/services/check_course_jobs.rb
```diff
@@ -30,21 +48,43 @@ def delete_unique_lock
...
+      'queue' => @queue,
       'unique_args' => [@course_id]
     }.to_json
     digest = OpenSSL::Digest::MD5.hexdigest hash
     "uniquejobs:#{digest}"
   end
 
-  def find_job
+  def find_job # rubocop:disable Metrics/CyclomaticComplexity
+    Sidekiq::ScheduledSet.new.select do |retri|
```
ragesoss, line 59: What is this doing?

app/services/check_course_jobs.rb
```diff
@@ -30,21 +48,43 @@ def delete_unique_lock
...
-      'queue' => queue_for(@course),
+      'queue' => @queue,
       'unique_args' => [@course_id]
     }.to_json
     digest = OpenSSL::Digest::MD5.hexdigest hash
     "uniquejobs:#{digest}"
   end
 
-  def find_job
+  def find_job # rubocop:disable Metrics/CyclomaticComplexity
```
ragesoss, line 58: Break this up into smaller methods instead off disabling the complexity cop. That will also make the code more self-documenting, if you choose good method names.

app/services/check_course_jobs.rb
```diff
@@ -30,21 +48,43 @@ def delete_unique_lock
...
+    Sidekiq::Workers.new.each do |_process_id, _thread_id, work|
+      next unless work['payload']['class'] == 'CourseDataUpdateWorker'
+      return work if work['payload']['args'] == [@course_id]
+    end
+
     return nil
   end
+
+  def delete_orphan_lock
+    if find_job.nil? && SidekiqUniqueJobs::Digests.all.include?(expected_digest)
```
ragesoss, line 80: The check for whether it's locked can be extracted to its own method, and you can add a method like `job_exists?` that does `find_job.nil?`.

app/services/check_course_jobs.rb
```diff
@@ -6,9 +6,27 @@
 class CheckCourseJobs
   include CourseQueueSorting
 
+  def self.remove_orphan_locks(courses_to_update)
```
ragesoss, line 9: This is a good time to refactor this class as cleanly as you can. The `health_report` bit can be deleted, and you should figure out which method(s) are the public interface vs which are just going to be called internally, and use `private` to divide them.

lib/data_cycle/schedule_course_updates.rb
```diff
@@ -36,7 +37,11 @@ def run_update
 
   def enqueue_course_updates
     log_message "Ready to update #{Course.ready_for_update.count} courses"
-    Course.ready_for_update.each do |course|
+
+    courses_to_update = Course.ready_for_update
+    CheckCourseJobs.remove_orphan_locks(courses_to_update)
```
ragesoss, line 42: What is the advantage of doing it with all courses in a batch, instead of one-by-one inside the block that also schedules each update?

app/services/check_course_jobs.rb
```diff
@@ -6,9 +6,27 @@
...
+    courses_to_update.each do |course|
+      check_course_job = new(course)
+      is_removed = check_course_job.delete_orphan_lock
+      orphan_lock_courses << check_course_job.sentry_extra if is_removed
+    end
+
+    orphan_lock_count = orphan_lock_courses.length
+    return unless orphan_lock_count.positive?
+
+    Raven.capture_message("#{orphan_lock_count} Orphan lock(s) removed",
```
ragesoss, line 20: This is fine, but I would also like to have logging on the individaul course record — probably using the same data structure as we already do for logging the completion time of each successful update. So, an orphan lock should be treated like an update that happened but failed, and it should increment count of total updates for a course accordingly.

ragesoss:
I think the most useful spec here will be a test of ScheduleCourseUpdates. In particular, I think there are three useful scenarios to test:

* course has a job already enqueued; no change in queue from before and after running ScheduleCourseUpdates
* course has no job enqueued and no lock; queue is empty before, and has one job after (and it's the right kind of job)
* course has no job enqueued but does have a lock; queue is empty before, has one job after, and the course.flags has a record of a failed update.

AmitJoki:
The code looks straight-forward but for the race condition, I think we could have the `find_job` be retried for a threshold say 10 times with an interval of 500ms.

This way, if even after 10 tries, there's no job, we can delete the orphan. Otherwise, if a job was indeed present, then find_job must have found the job. The thresholds can be thought out a bit but this approach could possibly rectfiy the race condition situation.

app/services/check_course_jobs.rb
shashwatkathuria: I am basically searching for scheduled jobs here, though it doesn't seem like the use of `CourseDataUpdateWorker` is scheduled. I am still trying to understand what is going on regarding the issue I mentioned on slack, the use of this is still somewhat unclear to me. [https://github.com/mperham/sidekiq/wiki/API#named-queues](https://github.com/mperham/sidekiq/wiki/API#named-queues)

app/services/check_course_jobs.rb
shashwatkathuria: Done!

app/services/check_course_jobs.rb
shashwatkathuria: Done!

app/services/check_course_jobs.rb
shashwatkathuria: Done!

lib/data_cycle/schedule_course_updates.rb
shashwatkathuria: There are 2 reasons why I am doing it before the update block
- I did not like the idea of removing orphan locks, and executing jobs one by one as that would be kind of messy. I wanted to move forward with executing the jobs only when the previous "mistakes"(orphan locks) were solved.
- It is at times difficult for debugging.

app/services/check_course_jobs.rb
shashwatkathuria: Done!

app/services/check_course_jobs.rb
ragesoss: I think this is exclusively for "Scheduled Jobs", which are jobs that sidekiq will run at a particular time. See https://github.com/mperham/sidekiq/wiki/Scheduled-Jobs

We don't use time-based scheduling for anything at this point, so I think you can remove this.

shashwatkathuria: Yes, removed

lib/data_cycle/check_course_jobs_logging.rb
```diff
@@ -0,0 +1,20 @@
...
+module CheckCourseJobsLogging
+  def log_orphan_record
+    Raven.capture_message('Orphan lock removed',
+                          level: 'warn',
+                          extra: sentry_extra)
+    return nil
+  end
+
+  def log_previous_failed_update
+    UpdateLogger.update_course(@course, 'orphan_lock_failure' => true)
```
ragesoss, line 14: I think it would be better to keep the timestamp for this. It could just be the timestamp instead of `true`.

shashwatkathuria: Done!

PR #4101, opened 2020-07-03
Dynamic tinymce

AmitJoki:
Imports Tinymce only when needed, aka, when `<TextAreaInput wysiwyg={true} />`

This PR also adds babel-eslint which lets us use dynamic `import()` statements.

`import()` doesn't work for tinymce since the tinymce loads the skins relative to the current URL which leads to 404/500.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,7 @@
 .idea/
 .vscode/
 /dump.rdb
-
+/public/skins
 *.rbc
 capybara-*.html
 .rspec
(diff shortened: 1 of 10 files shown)


ragesoss:
This still loads tinymce when a user visits a course timeline, even if the user is logged out and cannot edit (and so has no use for tinymce). It also appears to make a separate request for tinymce for every single block on the timeline.

tinymce

AmitJoki:
@ragesoss I fixed those, now it only loads once and only if the user is logged in.

But there are unrelated errors here in Travis something which I haven't touched upon at all.

PR #4103, opened 2020-07-11
Course specific activity improving query#3847

### LINKED ISSUE #3847: Activity information not course specific

shashwatkathuria:
## What this PR does
Fixes #3847 Improves the way course scoped revisions are received by sending all of them if an additional parameter is true. Earlier, the ids were sent, which were used to kind of filter or intersect with the revisions which are there, leading to weird behaviour if there are no recent tracked scope (course scoped) revisions by keep clicking on 'See More' to get the tracked revisions to appear after a few clicks depending upon the course.

### Rendering course scoped only when user goes there(talking about only the initial fetch)
Peek 2020-07-16 02-37

### Different 'See More' limits for both
Peek 2020-07-16 02-34


app/views/courses/revisions.json.jbuilder
```diff
@@ -1,11 +1,15 @@
 # frozen_string_literal: true
 
 json.course do
-  json.course_specific_revision_ids @course.tracked_revisions
-                                           .order(date: :desc).limit(@limit).pluck(:id)
-  json.revisions(@course.recent_revisions
-                        .eager_load(:user, :wiki).includes(article: :wiki)
-                        .order(date: :desc).limit(@limit)) do |rev|
+  revisions = if @course_scoped
+                @course.tracked_revisions.order(date: :desc).limit(@limit)
```
ragesoss, line 5: I think you want to use the same `eager_load` in the query here; I believe that prevents N+1 queries when rendering the json for each individual revision.

shashwatkathuria: Done!

app/assets/javascripts/components/revisions/revisions_handler.jsx
```diff
@@ -3,41 +3,44 @@ import createReactClass from 'create-react-class';
...
+  getInitialState() {
+    return {
+      isCourseScoped: false
+    };
+  },
+
   componentDidMount() {
     if (this.props.loadingRevisions) {
       this.props.fetchRevisions(this.props.course_id, this.props.limit);
+      this.props.fetchRevisions(this.props.course_id, this.props.courseScopedLimit, true);
```
ragesoss, line 34: This should only be fetched upon request, when a user switches to course-scoped; otherwise, we're doing a lot of extra queries that in most cases will not be used.

ragesoss:
Aside from the too-eager fetching I noted, this works great!

I tried it out locally with the [example course](https://outreachdashboard.wmflabs.org/courses/Museu_Paulista/Wikiconcurso_do_Museu_Paulista_2020/home) and it works just like it ought to; I think this will be very useful the organizer of that program, among others.

app/assets/javascripts/components/revisions/revisions_handler.jsx
shashwatkathuria: Done! `componentDidUpdate()` was making the most sense for me to do this, so I implemented it there

shashwatkathuria:
Happy to know this!!

app/assets/javascripts/components/revisions/revisions_handler.jsx
ragesoss: Hmm... not sure that's the best place for it. Why not have it be part of action from clicking the button?

shashwatkathuria: Done!

ragesoss:
This works nicely!

There's a repeated query, but I think I can take care of that...

One possible area for further improvement: the `loading` and `courseRevisionsLoading` state names are somewhat misleading, as the revisions aren't necessarily being loaded now, once the component is mounted... they will be not present and also not currently loading, until the button is clicked.

For clarity, I would change both of these bits of state to `loaded` instead of `loading` (and reverse boolean) so that we don't get confused about the meaning here. We change the state when a given set of revisions gets received, so at that point we can say the revisions are loaded.

Another option is to make the state tracking more complete by keeping track of `REQUEST_REVISIONS` (etc) as additional actions so that the state will also prevent a request from being repeated while the initial one is still resolving.

I'll go ahead an merge now. If you'd like to do the above refactoring in a separate PR, go for it. Otherwise, I'm okay leaving it as-is for now.

PR #4113, opened 2020-07-21
Surfacing to UI future course update information and some css fixes

shashwatkathuria:
## What this PR does
Surfaces following information to UI:
- Future update date
- Whether or not one-time additional update is scheduled
- Also, includes some CSS fixes
### Before
Screenshot from 2020-07-22 01-36-25
Screenshot from 2020-07-22 01-36-12

#### Regularly updated
Screenshot from 2020-08-01 00-29-31
#### No longer updated, but one-time update scheduled
Screenshot from 2020-08-01 00-30-47
#### No longer updated
Screenshot from 2020-08-01 00-29-43

### Before and After
Peek 2020-07-31 03-56
Peek 2020-08-01 01-57
### Before and After
Screenshot from 2020-08-01 01-35-08
Peek 2020-08-01 01-54
### Before and After
Peek 2020-08-01 01-32
Screenshot from 2020-08-01 01-52-08

### DIFF
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -323,7 +323,9 @@ nav.ham-nav
       text-decoration none
       text-align center
       +above(desktop)
-        max-width 370px
+        max-width 290px
+        @media only screen and (min-width: 1000px)
+          max-width 370px
         margin-right 20px
         padding-right 40px
         text-align left
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -295,7 +295,7 @@ nav.ham-nav
 .course-nav__wrapper,
 .campaign-nav__wrapper
   height $nav_height
-  +below(desktop)
+  +below(desktop - 1)
     margin-top 50px
 
 .course_navigation,
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -250,6 +250,11 @@ class Course < ApplicationRecord
     flags[:closed_date].present?
   end
 
+  def current_and_future?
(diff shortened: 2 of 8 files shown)


app/models/course.rb
```diff
@@ -285,6 +290,10 @@ def category_article_ids
     categories.inject([]) { |ids, cat| ids + cat.article_ids }
   end
 
+  def update_in_future?
```
ragesoss, line 293: Instead of this, I think it would be more useful to define `updates_until` or something like that, which is the date that updates will continue until. This will be much more useful on the frontend, so people know exactly what to expect in terms of continued updates.

Then on the frontend, you can change the message depending on whether `updates_until` is in the future or not, and you can also separately show a message when a non-updating course has been scheduled for a single update (ie, `needs_update: true`).

shashwatkathuria: Done!

app/assets/stylesheets/modules/_nav.styl
```diff
@@ -295,7 +295,7 @@ nav.ham-nav
 .course-nav__wrapper,
 .campaign-nav__wrapper
   height $nav_height
-  +below(desktop)
+  +below(desktop - 1)
```
ragesoss, line 298: What's going on with the handful of 1-pixel changes?

shashwatkathuria: This is a borderline case, where the margin should be applied starting from 919px, so that is why the `-1` . At 920px, the margin gets applied but the mobile layout mode hasn't yet been applied

ragesoss: Okay. A separate PR to clean up the hard-coding would be great.

PR #4114, opened 2020-07-21
Improving Activity Tab.

shashwatkathuria:
Using 'loaded' instead of 'loading' to avoid misleading meaning and added loading spinner as done in other tabs also.
## What this PR does
Fixes last comment mentioned in #4103 .
- Adds loading spinner for each revisions request (all or course scoped)
- Renamed `loading` to `loaded` in both, all scope and course scoped. The misleading meaning regarding `loading` refers to that when the component is mounted, the all scope revisions are loading and then we get the revisions, but, the course scoped revisions are neither loading nor the course scoped revisions are there, hence the misleading meaning of `loading` here.
- Prevents another request by not showing the 'See More' until the previous one has resolved

### Before (Slow Internet Speed Example on this [example course](https://outreachdashboard.wmflabs.org/courses/Museu_Paulista/Wikiconcurso_do_Museu_Paulista_2020/activity))
Peek 2020-07-22 03-59

### Before
Peek 2020-07-23 02-26

### After
Peek 2020-07-23 02-12

### After (Example on a course with no revisions - Showing no recent activity messsage only when sure, i.e., revisions are loaded)
Peek 2020-07-23 02-17


ragesoss:
This looks great!

PR #4120, opened 2020-07-27
Moving updation of categories to UpdateCourseStats

shashwatkathuria:
## What this PR does
Removes the daily workers which update the categories daily for current courses and move that behaviour to UpdateCourseStats on a per course basis instead and doing per category updation per day.

NOTE: Consider branch name as moving-updation-categories-to-updatecoursestats, article status updation will be moved in a separate PR


app/services/update_course_stats.rb
```diff
@@ -82,6 +84,11 @@ def log_update_progress(step)
                           extra: { logs: @sentry_logs }
   end
 
+  def should_update_categories?
+    previous_categories_update = @course.flags['previous_categories_update']
```
ragesoss, line 88: I don't think this is the right approach. The decision of whether to update a category should be based on `Category#updated_at` (and probably doesn't need to happen inside `UpdateCourseStats`.
`Category.refresh_categories_for` should apply a query to `Course.categories` to find the ones that were updated more than a day ago, and just update those ones.

One additional consideration is to make sure that `updated_at` gets changed when this happens, even if the actual contents of the category didn't change. This might or might not require [`touch`](https://apidock.com/rails/ActiveRecord/Timestamp/touch).

shashwatkathuria: Done, used `touch` to update `updated_at` timestamp

shashwatkathuria:
I thought that `needs_update` condition was not required anymore so I removed it, now that the category updates have been moved from `daily_update` in batches to `UpdateCourseStats` individually and happen according to each course category with the required time condition, so they will anyways get updated whenever required in an update cycle. Is removing `needs_update` correct?

app/services/update_course_stats.rb
ragesoss: This method can be deleted now, right?

ragesoss:
Yes, that's correct; there's nothing we need to special-case for `needs_update` now.

app/services/update_course_stats.rb
shashwatkathuria: Yes, deleted....forgot to do that

Issue #4048, opened 2020-05-26
ArticleViewer authorship highlighting does not work for user with `+` in username


ragesoss:
### What is happening?
In ArticleViewer, a user named `Latinx+art38` does not have their edits highlighted, even though the whocolor data includes text attributed to that user.

This is probably related to the Dashboard showing that username as `Latinx art38`, because at some point it gets interpreted as a URL and the plus sign gets turned into a space.

The student's username is show correctly elsewhere in the Dashboard (such as on the Students page), but not at the bottom of ArticleViewer.

### To Reproduce

View this article with authorship highlighting:

https://dashboard.wikiedu.org/courses/Fort_Lewis_College/ENG_177_(Spring_2020)/articles/edited?showArticle=52231339

### Expected behavior

Some text, such as the text added [here](https://en.wikipedia.org/w/index.php?title=Josefina_Niggli&diff=945214708&oldid=929279441&diffmode=source), should be highlighted.

devil-cyber:
can I work on this? I am new to open source

ragesoss:
@devil-cyber go for it. This is a good first issue. :-)

If you'd like to join our slack, send me an email.

If you have any questions about how to get set up or how to replicate the issue locally, you can ask here or on slack.

devil-cyber:
I have sent you the mail please check out @ragesoss

ragesoss:
@devil-cyber that email didn't come through. Can you try to sage@wikiedu.org?

devil-cyber:
how to replicate the issue locally?will you please provide me some insight
of this issue?

ragesoss:
Once you have your dev environment set up, you can replicate this by creating a course that matches the start and end dates of the one linked in the description, then add `Latinx+art38` as a student. Then you can do a data update on your course (see [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/user_roles.md#populating-a-course-with-data)), and then you should be able to find that same article in your course.

devil-cyber:
ok I am going to work on this

BaileyMcKelway:
Got my dev environment setup and was going to take a look into this if it is still open. 👍

BaileyMcKelway:
#4133

BaileyMcKelway:
#4133

BaileyMcKelway:
Squashed my commits and fixed the prettier issue on this pull request.
#4133

PR #4133, opened 2020-08-03
fixed author highlighting

BaileyMcKelway:
Seems like Eslint is adding some inconsequential changes.

I added encodeURIComponent to the users name before it is added to the URL this prevents the "+" from being translated into a " ". This is in file common/ArticleViwer/URLBuilder.js

Then I decode the user name in common/ArticleViewer/container/ArticleViewer.jsx


BaileyMcKelway:
#4048

ragesoss:
@BaileyMcKelway nice! This makes sense to me. Can you include the before/after screenshot?

The eslint changes are fine, but they should be moved to a separate commit so that it's easier to see the functional changes on their own.

Also, please exclude the schema.rb changes from the PR.

BaileyMcKelway:
Prev
Update

ragesoss:
Looks good to me!

PR #4139, opened 2020-08-04
Moving article status updation to UpdateCourseStats

shashwatkathuria:
## What this PR does
Moving article status updation to UpdateCourseStats.


ragesoss:
I think this needs to use the same strategy as the category updates, and only check the status for articles that haven't been checked for more than a day.

This is one of the most time-consuming data update processes, which is why it was in the daily_update. It's also not something that changes very often for any given article; the idea is that article status needs to be updated to reflect when articles get moved or deleted, but those are relatively rare occurrences.

I think we should probably also just do away with the threads altogether, when switching to the course-by-course approach. Doing it as part of the course updates means that there's already some concurrency based on having multiple course updates happening at once time, and the extra threading — especially just with two threads per worker — doesn't provide much benefit compared to the risk of running out of memory network connections (which I believe is what had been happening a few months ago before I changed the concurrency from 10 to 2).

shashwatkathuria:
@ragesoss Done!

lib/article_status_manager.rb
```diff
@@ -12,25 +12,20 @@ def initialize(wiki = nil)
...
-        course_batch.each do |course|
-          update_article_status_for_course(course)
-        end
-      end
-    end
-    threads.each(&:join)
-  end
-
   def self.update_article_status_for_course(course)
     Wiki.all.each do |wiki|
```
ragesoss, line 16: It looks like this strategy of going through every wiki for every course is left over from before Amit implemented `CoursesWikis`. This should be `course.wikis.each` instead; we only ever count contributions to tracked wikis for each course, so there's no need to deal with pages edited on other wikis.

lib/article_status_manager.rb
```diff
@@ -12,25 +12,20 @@ def initialize(wiki = nil)
...
       next unless course.pages_edited.exists?(wiki_id: wiki.id)
-      course.pages_edited.where(wiki_id: wiki.id).find_in_batches do |article_batch|
-        new(wiki).update_status(article_batch)
+      # Updating only those articles which are updated more than  1 day ago
+      course.pages_edited
+            .where(wiki_id: wiki.id)
+            .where('articles.updated_at < ? OR articles.created_at = articles.updated_at',
```
ragesoss, line 21: I don't think the `created_at` part is necessary here. Unlike a `Category`, where it gets created in a state of incomplete info (ie, there's no list of articles for it at the time it gets created), an `Article` record gets created in an up-to-date state at the time the first revisions to that article get imported. So if it was created less than a day ago (and updated only at that time) then it still doesn't need to be updated during this process.

lib/article_status_manager.rb
```diff
@@ -12,25 +12,20 @@ def initialize(wiki = nil)
...
+                   1.day.ago)
+            .in_batches do |article_batch|
+        # Using in_batches so that the update_at of all articles in the batch can be
+        # excuted in a single query, otherwise if we use find_in_batches, query for
+        # each article for updating the same would be required
+        new(wiki).update_status(article_batch.to_a)
```
ragesoss, line 27: Why is `.to_a` being used here?

shashwatkathuria: I thought of giving the same type of input to update_status as before, i.e., `array` (returned by `find_in_batches`), but it seems like `ActiveRecord::Relation` object (returned by `in_batches`) also works the same as array in this case. Removed `to_a`.

lib/article_status_manager.rb
shashwatkathuria: Done!

lib/article_status_manager.rb
shashwatkathuria: Done!

PR #4154, opened 2020-08-11
Tracking category talk page categories and talk template categories

### LINKED ISSUE #3762: Tracking a Category doesn't work for talk page categories (or for talk templates outside of English Wikipedia)
For example, this category appears on article talk pages: https://pt.wikipedia.org/wiki/Categoria:!Artigos_de_import%C3%A2ncia_1_sobre_Museu_Paulista

The MediaWiki API returns page titles for a `categorymembers` query with the namespace prefix included, but the prefix will vary by language.

I think the strategy to fix this for both category- and template-type Category records could be similar to the one taken already in `TransclusionImporter` but modified to take into account the namespace of each result. The results for both a transcludedin and categorymembers query can include the namespace in addition to the title of each result: https://en.wikipedia.org/w/api.php?action=query&prop=transcludedin&titles=Template:Cn

For any page outside mainspace (namespace 0), there will be a namespace prefix at the start of the page title, and it will end in a colon (`:`). So to extract the titles without the prefix, we would remove everything up to the first colon for any pages that aren't in namespace 0.

The code for processing the API results to get the titles without namespaces could probably be shared between `TransclusionImporter` and `CategoryImporter`. (This will probably require some refactoring and extension of CategoryImporter, since it sometimes gets used by code that expects the titles to include prefixes, I think. Maybe a new method like mainspace_page_titles_for_category could be added, to be used by `Category#title_list_from_wiki`.)

shashwatkathuria:
## What this PR does
Fixes #3762

Putting here the difference between `properties` and `data`:
Data is the result which comes from the API, refers to the items inside "transcludein"
[https://en.wikipedia.org/w/api.php?action=query&prop=transcludedin&titles=Template:Cn](https://en.wikipedia.org/w/api.php?action=query&prop=transcludedin&titles=Template:Cn)
`{ "pageid": 12345, "ns": 0, "title": "ABC Title"}`

<del>The difference I made here was that I made `property` parameter(which was used earlier) optional for `CategoryImporter`, so that we can get only the `"title"` keys from the result (or any other `property`). But for removing prefixes we want to know the `"ns"` (namespace) also, which is why I added support for getting all the key/value pairs in json for the page. </del>

All data passes though, filtering done by CategoryUtils.

Removing prefixes requires us to know the namespace, which if 0, should be as they are, and when not 0, the prefix (for e.g., `Talk:`) should be removed.

### DIFF
--- a/lib/category_utils.rb
+++ b/lib/category_utils.rb
@@ -8,6 +8,7 @@ class CategoryUtils
   #     "ns": 0,
   #     "title": "Andorra"
   # }
+
   # Removing prefixes for pages outside mainspace
   # Everything till (and including) the first semicolon(:) is removed
   def self.get_titles_without_prefixes(pages)
@@ -15,5 +16,9 @@ class CategoryUtils
       title = page['title']
       page['ns'] == Article::Namespaces::MAINSPACE ? title : title[(title.index(':') + 1)..-1]
     end
+  end
+
+  def self.get_titles(pages)
+    pages.map { |page| page['title'] }
   end
 end
--- a/lib/category_utils.rb
+++ b/lib/category_utils.rb
@@ -1,6 +1,13 @@
 # frozen_string_literal: true
 
 class CategoryUtils
+  # API Documentation: https://en.wikipedia.org/w/api.php?action=query&prop=transcludedin&titles=Template:Cn
+  # A page object here looks like:
+  # {
+  #     "pageid": 600,
+  #     "ns": 0,
+  #     "title": "Andorra"
+  # }
   # Removing prefixes for pages outside mainspace
   # Everything till (and including) the first semicolon(:) is removed
   def self.get_titles_without_prefixes(pages)
--- a/lib/category_utils.rb
+++ b/lib/category_utils.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+class CategoryUtils
+  def self.get_titles_without_prefixes(pages)
+    pages.map do |page|
+      title = page['title']
+      page['ns'] == Article::Namespaces::MAINSPACE ? title : title[(title.index(':') + 1)..-1]
+    end
+  end
+end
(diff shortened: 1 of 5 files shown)


lib/category_utils.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class CategoryUtils
+  # Removing prefixes for pages outside mainspace
+  # Everything till (and including) the first semicolon(:) is removed
+  def self.get_titles_without_prefixes(pages)
```
ragesoss, line 6: It would be useful to document in a comment what a `page` object looks like here and why. (ie, I guess it's a hash that includes at least `title` and `ns` keys, which you can get from a `list=categorymembers` query from the MediaWiki API)

lib/importers/category_importer.rb
```diff
@@ -18,42 +19,49 @@ def initialize(wiki, opts={})
...
-    property_values = []
+  def get_category_member_data(query, property)
+    data_values = []
     continue = true
     until continue.nil?
       cat_response = WikiApi.new(@wiki).query query
       page_data = cat_response.data['categorymembers']
-      page_data.each { |page| property_values << page[property] }
+      page_data.each do |page|
+        data = property.present? ? page[property] : page
```
ragesoss, line 52: This is a bit confusing. In what situation will the requested property no be present?

lib/category_utils.rb
shashwatkathuria: Yes, documented it in a comment

lib/importers/category_importer.rb
shashwatkathuria: Earlier property data was getting filtered here itself, and with that I had to combine an implementation which required the full page data (to know the namespace and accordingly remove the prefix to fix the bug), but all of this was getting unnecessarily confusing. Now all the data received gets returned and whatever property filtering required is done in `CategoryUtils`, as filtering that data shouldn't really be the job of `get_category_member_data`.

Issue #4023, opened 2020-05-08
In the Students details view, ArticleViewer only shows edits by assigned group members


ragesoss:
#3143 # What is happening?
ArticleViewer with authorship highlighting is available from both the Students tab details view, and from the Articles tab.

On the Articles tab, it will show authorship data for any student in the course who has edited that article. For example: https://dashboard.wikiedu.org/courses/Georgetown_University/Culture_Medicine_and_Gender_(Spring_2020)/articles/edited?showArticle=175357

On the Student tab details view, it only shows the authorship for the set of students who are assigned to that articles. For example, here's it's missing Jane4dc:
https://dashboard.wikiedu.org/courses/Georgetown_University/Culture_Medicine_and_Gender_(Spring_2020)/students/articles/Hadleyggeorgetown?showArticle=175357

### Expected behavior
ArticleViewer should show all the students who edited a given article, even when viewed from the Students tab assignment list.

h-ranjan1110:
@ragesoss the links that you gave are taking much time to open. like it loads for 2-3 minutes and then stops. my internet connection is good, other sites work properly. can you please see what is this issue.

rahahahat:
Has this issue been resolved, because i can see jane4dc in the Students tab details view.

ragesoss:
In this example, I'm not sure. jane4dc was added as a group member, so we can't tell from that link whether this is still a bug. @BaileyMcKelway has been working on improvements to ArticleViewer, but I'm not sure whether they changed this behavior.

You should be able to set up the same situation locally by adding those same users to an example course (with the start/end dates from that course), doing a course stats update to pull in their edits, and assigning all the student except jane4dc to that article.

rahahahat:
Alright thanks, i'll try to reproduce the bug. However, would i need to set-up OAuth for this functionality?

ragesoss:
No, the default OAuth credentials allow you to sign into the dev instance using any Wikipedia account, which is all you'll need.

Setting up your own OAuth credentials is only required to test OAuth edit actions.

rahahahat:
Hello @ragesoss, 
I have somewhat replicated the course structure, added 4 students (editors) : `Jane4dc` `Amm567` `Ashiannaj` `Hadleyggeorgetown` and pulled their edits. Afterwards, I added the `Postpartum depression` article and assigned it to all except `Jane4dc`.
 This is the behaviour I got from the article viewer: 
[...]

PR #4245, opened 2020-10-13
Fix bug mentioned in Issue #4023, related to articleViewer [WIP]

rahahahat:
## What this PR does
This pull request fixes the bug mentioned in issue #4023, wherein, when accessed through the `Students` tab the `articleViewer` container did not display all the editors that have edited a particular article, and only displayed those editors that were assigned to that article.

## Screenshots
Before:
94878069-3ea9a000-0454-11eb-9c99-0d57562878ed

After:
Screenshot from 2020-10-13 00-54-08 (1)

The approach i chose to solve this bug was to replicate the process of acquiring editors when the `articleViewer` was accessed through the editors tab, as that was the only reference i had to a bug free implementation. My approach involved invoking the `fetchArticleDetails` function with relative arguments to update the `redux-store` with the `articleDetails` field, using which i extract the missing users. However, since my approach involves altering the `redux-store`, I was wondering if there was another approach to this bug.

### DIFF
--- a/app/assets/javascripts/components/common/get_help_button.jsx
+++ b/app/assets/javascripts/components/common/get_help_button.jsx
@@ -203,7 +203,7 @@ const GetHelpButton = createReactClass({
           <form target="_blank" action="/faq" acceptCharset="UTF-8" method="get">
             <input name="utf8" type="hidden" defaultValue="✓" />
             <input name="source" type="hidden" defaultValue="get_help_button" />
-            <input type="text" name="search" id="search" defaultValue="" placeholder="Search Help Forum" />
+            <input type="text" name="search" id="get_help_search" defaultValue="" placeholder="Search Help Forum" />
             <button type="submit">
               <i className="icon icon-search" />
             </button>
--- a/app/assets/javascripts/components/nav/hamburger_menu.jsx
+++ b/app/assets/javascripts/components/nav/hamburger_menu.jsx
@@ -69,7 +69,7 @@ const HamburgerMenu = createReactClass({
           <li>
             <form className="top-nav__faq-search" target="_blank" action="/faq" acceptCharset="UTF-8" method="get">
               <input name="utf8" type="hidden" defaultValue="✓" />
-              <input type="text" name="search" id="search" defaultValue="" placeholder={I18n.t('application.search')} />
+              <input type="text" name="search" id="hamburger_search" defaultValue="" placeholder={I18n.t('application.search')} />
               <input name="source" type="hidden" defaultValue="nav_ask_form" />
               <button type="submit">
                 <i className="icon icon-search" />
--- a/app/assets/javascripts/components/nav/nav.jsx
+++ b/app/assets/javascripts/components/nav/nav.jsx
@@ -114,7 +114,7 @@ const Nav = createReactClass({
           <div className="top-nav__faq-search">
             <form target="_blank" action="/faq" acceptCharset="UTF-8" method="get">
               <input name="utf8" type="hidden" defaultValue="✓" />
-              <input type="text" name="search" id="search" defaultValue="" placeholder={I18n.t('application.search')} />
+              <input type="text" name="search" id="nav_search" defaultValue="" placeholder={I18n.t('application.search')} />
               <input name="source" type="hidden" defaultValue="nav_ask_form" />
               <button type="submit">
                 <i className="icon icon-search" />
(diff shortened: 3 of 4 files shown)


ragesoss:
Can you explain what you mean by 'altering the redux-store'?

This looks like a relatively clean implementation. I think it would benefit from separating out the different roles of assigned users vs. editors more, so instead of taking a union of them to produce a single `users` state, keep those pieces of state separate... `assignedUsers` and `editors`. I think that would allow you to reduce the complexity of some of the conditionals you've added.

rahahahat:
So, the current fix, updates the redux state through the function `fetchArticleDetails` and then passes on the changes (`articleDetails`) as props to the articleViewer.  Also, since I am passing in a prop called `assigned` in `Assignment.jsx` to the `ArticleViewer.jsx`, that makes this fix just restriced to the `ArticleViewer` when viewed from the Students Tab and only works if the students have any assignments. In my first commit, the fix was very generic but caused unncessary state updates whenever the `articleViewer` container was invoked anywhere in the code, but it did update the redux state with `articleDetails` everytime.
Please point me towards which implementation you'd prefer. Also, i'll make the `users` state changes. Thanks alot for helping.

ragesoss:
I think using `fetchArticleDetails` is the right approach. That means we get two separate pieces of information passed into the ArticleViewer component: the set of users from the course who have edited the article (which comes from fetchArticleDetails), and the set of assigned users. If these are kept separate for as long as possible in the data flow, it will make it easier to follow what's happening and to make changes. They only really need to be combined for rendering.

rahahahat:
Thanks you very much for the help, i will get on it immediately.

rahahahat:
@ragesoss I have simplified the conditional logic and divided `users` into `assignedUsers` and `editors`. Please review the changes. Thanks.

app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
```diff
@@ -48,9 +50,15 @@ export class ArticleViewer extends React.Component {
     this.showBadArticleAlert = this.showBadArticleAlert.bind(this);
     this.submitBadWorkAlert = this.submitBadWorkAlert.bind(this);
     this.isWhocolorLang = this.isWhocolorLang.bind(this);
+    this.updateStateEditors = this.updateStateEditors.bind(this);
   }
 
   componentDidMount() {
+    if (this.props.assigned) {
+      if (!has(this.props.articleDetails, this.props.article.id) && isEmpty(this.props.fetchArticleDetails)) {
```
ragesoss, line 58: This is confusing to me. What are the situations this code is trying to deal with? In particular, what does `isEmpty(this.props.fetchArticleDetails)` do?

rahahahat: Hi, seems that there has been typographical error in the logic, i'll fix is immediately.

app/assets/javascripts/components/students/components/Articles/SelectedStudent/AssignmentsList/Assignment/Assignment.jsx
```diff
@@ -51,6 +51,7 @@ export const Assignment = ({ assignment, course, current_user, fetchArticleDetai
           fetchArticleDetails={fetchArticleDetails}
           users={users}
           showOnMount={showArticleId === article.id}
+          assigned={true}
```
ragesoss, line 54: I think rather than using `assigned` as a boolean here, we want to basically pass the assigned editors as a different prop instead of `users`. Then for `users`, we can use the same strategy as as in `article.jsx` to set `users`, and that prop will represent all the editors who have made edits. This way, the data flow is not conditional, and an `assignedUsers` prop is just used to add additional usernames to the legend.

ragesoss (commented):
I think this can be fixed without nearly as many conditionals.

app/assets/javascripts/components/students/components/Articles/SelectedStudent/AssignmentsList/Assignment/Assignment.jsx
rahahahat: Thanks alot. i'll get working on it.

rahahahat:
> I think this can be fixed without nearly as many conditionals.
Will look into this as well

rahahahat:
@ragesoss 
Due to some problmes with my laptop, i lost all my files, when i finally managed to get everything working again, i had problems rebasing my branch, so i decided to delete my old branch and pull request. I have opened a new pull request #4264 regarding the same issue. Sorry for the inconvenience.

Issue #4217, opened 2020-09-28
Editors CSV should include number of articles edited, and number of articles created, for each user


ragesoss:
### Current Behavior:

The CSV of editor data for a course, which can be downloaded via the 'Download stats' modal, includes the following columns: `username	enrollment_timestamp	registered_at	revisions_during_project	mainspace_bytes_added	userpace_bytes_added	draft_space_bytes_added	references_added	registered_during_project`

### Desired Behavior:
Users have requested data on how many articles each editor has edited, and how many articles each editor has created. Columns for those values should be added to the CSV output.

### Related Files & Code:

`lib/analytics/course_students_csv_builder.rb`

RuthNjeri:
@ragesoss I would like to pick this issue, please assign it to me?

ragesoss:
Go for it. We don't generally assign issues; just leave another message here if you're no longer working on it.

RuthNjeri:
Oh alright @ragesoss 

Based on the CoursesUsers model, it seems that the columns to use are:
1. total_uploads for `how many articles each editor has created`
2. revision_count for `how many articles each editor has edited`
Is this correct?

Also, when I access the 'Download stats' model, I get another pop up, 
![Screenshot 2020-10-12 at 18 52 26](https://user-images.githubusercontent.com/7622875/95766611-2e978900-0cbc-11eb-9484-f82315de6158.png)

Which one is related to the CV with the columns related to this issue?

RuthNjeri:
I have downloaded each CSV above and can't find one with these headers `username enrollment_timestamp registered_at revisions_during_project mainspace_bytes_added userpace_bytes_added draft_space_bytes_added references_added registered_during_project` I am testing it locally using the dummy data generated..

ragesoss:
No, neither of the metrics for this issue correspond directly to columns in the CoursesUsers model.

total_uploads is the number of files uploaded to Wikimedia Commons (tracked by CommonsUpload records).

revision_count is the number of edits the Dashboard has recorded for the user during the course (tracked by Revision records). The number of unique articles (unique pages in namespace 0) edited is different.

The popup you have in the screenshot is from a Campaign (which is a collection of Courses). You should probably start from a Course, which has a similar popup, although I think the solution for this issue will affect the CSVs for both Courses and Campaigns.

RuthNjeri:
So far, what I know is that I should use the courses_users in determining what articles the course_user has edited and created.

[...]

PR #4248, opened 2020-10-14
Add total articles created and edited in CSV

RuthNjeri:
## What this PR does
- Adds the total number of articles edited or created by editors in the editors CSV
- The [issue](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/4217)

## Screenshots
Before:
Screenshot 2020-10-14 at 19 35 38

After:
Screenshot 2020-10-14 at 19 29 49

- The build is failing because of  [50 and 51](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4248/files#diff-6c15f159a652c018e981d3e09825dbf1b0623f0b08b4573c4b4f66345659a31fR50-R51) in the CSV builder. 
- I can create a separate method to fix this unless we don't need to worry about this issue in favour of readability.

### DIFF
--- a/lib/analytics/course_students_csv_builder.rb
+++ b/lib/analytics/course_students_csv_builder.rb
@@ -5,6 +5,9 @@ require 'csv'
 class CourseStudentsCsvBuilder
   def initialize(course)
     @course = course
+    @editors = @course.articles_courses.pluck(:user_ids).flatten
+    new_articles = @course.articles_courses.where(new_article: 1)
+    @new_articles_revisions = Revision.where(article_id: new_articles, new_article: 1)
   end
 
   def generate_csv
@@ -31,6 +34,8 @@ class CourseStudentsCsvBuilder
     draft_space_bytes_added
     references_added
     registered_during_project
+    total_articles_edited
+    total_articles_created
   ].freeze
   def row(courses_user)
     row = [courses_user.user.username]
@@ -42,9 +47,21 @@ class CourseStudentsCsvBuilder
     row << courses_user.character_sum_draft
     row << courses_user.references_count
     row << newbie?(courses_user.user)
+    row << total_articles_edited(courses_user.id)
+    row << total_articles_created(courses_user.id)
   end
 
   def newbie?(user)
     (@course.start..@course.end).cover? user.registered_at
+  end
+
+  def total_articles_edited(id)
+    total_edits = @editors.count(id)
+    @editors.delete(id)
+    total_edits
+  end
+
+  def total_articles_created(id)
+    @new_articles_revisions.where(user_id: id).count
   end
 end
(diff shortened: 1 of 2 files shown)


lib/analytics/course_students_csv_builder.rb
```diff
@@ -42,9 +47,21 @@ def row(courses_user)
...
   end
+
+  def total_articles_edited(id)
+    total_edits = @editors.count(id)
+    @editors.delete(id)
+    total_edits
+  end
+
+  def total_articles_created(id)
+    @new_articles_revisions.where(user_id: id).count
```
ragesoss, line 65: This will count all new articles that the Dashboard knows about, not just ones created during the course.

lib/analytics/course_students_csv_builder.rb
```diff
@@ -5,6 +5,9 @@
 class CourseStudentsCsvBuilder
   def initialize(course)
     @course = course
+    @editors = @course.articles_courses.pluck(:user_ids).flatten
+    new_articles = @course.articles_courses.where(new_article: 1)
+    @new_articles_revisions = Revision.where(article_id: new_articles)
```
ragesoss, line 10: This is likely to be a very expensive query on a production server, which will have millions of Revisions and possibly tens or hundreds of thousands of new_article revisions.

RuthNjeri: Noted, I'll try and find another way..

lib/analytics/course_students_csv_builder.rb
RuthNjeri: I thought it would only affect articles created in the course, because this query
`new_articles = @course.articles_courses.where(new_article: 1)` gets all the new articles within the course...

Then `@new_articles_revisions = Revision.where(article_id: new_articles)` uses the new_articles(course's articles) and ignores other articles.

RuthNjeri: I think it will count all the revisions a user has made.. because I did not limit the `@new_articles_revisions` to search for `new_article` as well...

Once I correct the query above.. this should be fine..

lib/analytics/course_students_csv_builder.rb
RuthNjeri: What if, I limit the scope of the query to also include new articles, for instance 
@new_articles_revisions = Revision.where(article_id: new_articles, new_article: 1)

That way, it does not retrieve all the revisions of an article, just the first one for each article..

RuthNjeri:
Hi @ragesoss I made a small change of adding the new note attribute in the revision query.. I am not sure if this would make an impact in the change.. 
Alternatively, I was thinking of maybe adding a column on the article courses that specifies the user id that created the course because it's difficult to determine which user created the course in the list of user_ids..

RuthNjeri:
I'll no longer be able to work on this due to other commitments.

ragesoss:
Thanks for letting me know, @RuthNjeri!

Apologies I couldn't review it sooner.

PR #4261, opened 2020-10-22
Make 'Available Articles' work for group assignments

ragesoss:
This adds new behavior to Available Articles for courses that select to work in groups: an Available Article won't leave the list after it is selected, and will remain so tha another classmate can choose the same one.


ragesoss:
TODO: make sure already-assigned articles don't show up as available for a given user

ragesoss:
merged.

PR #4264, opened 2020-10-24
Fix bug mentioned in Issue #4023, related to articleViewer

rahahahat:
## What this PR does
This pull request fixes the bug mentioned in issue #4023, wherein, when accessed through the Students tab the articleViewer container did not display all the editors that have edited a particular article, and only displayed those editors that were assigned to that article. 

## Screenshots
Before:

After:

This bug was addressed in a previous Pull Request #4245, However due to some errors related to rebasing, i decided to delete my old branch and opened a new pull request.

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
@@ -61,13 +61,26 @@ export class ArticleViewer extends React.Component {
   // fetch the MediaWiki user ids, which are used for coloration. Those can't be
   // fetched until the usernames are available, so 'show' will fetch the usernames
   // first in that case. In that case, componentDidUpdate fetches the
-  // user ids as soon as usernames are avaialable.
+  // user ids as soon as usernames are avaialable. However, the articleViewer will 
+  // contain an extra prop "assignedUsers" when viewed from the Students/Editors tab
+  // which holds all the users extracted from assigned articles in addition to 
+  // the "users" prop, which in this case contains all the usernames that have edited
+  // the article but have not been assigned to it.
   componentDidUpdate(prevProps, prevState) {
+    // checks for the availability of 'assignedUsers' and 'users' props.
+    // This condition will only pass when articleViewer is accessed from the Students/Editors tab 
     if (this.props.assignedUsers && this.props.users) {
+    // checks if usersIds required for coloration have already been fetched or not.
+    // if not, userIds are fetched by passing in an array that contains both 'assignedUsers'
+    // and 'users' as an argument to the fetchUserIds function.
       if (!this.state.userIdsFetched) {
         this.fetchUserIds(union(this.props.assignedUsers, this.props.users));
       }
-    } else if (!prevProps.users && this.props.users) {
+    }
+    // this condition will pass whenever 'assignedUsers' props is not available
+    // and the fetchUserIds inside this branch is invoked when 'users', if not available
+    // previously, have been fetched and passed down as props.
+    else if (!prevProps.users && this.props.users) {
         if (!prevState.userIdsFetched) {
           this.fetchUserIds(this.props.users);
       }
--- a/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
@@ -4,7 +4,7 @@ import { connect } from 'react-redux';
 import OnClickOutside from 'react-onclickoutside';
 
 // Utilities
-import { forEach } from 'lodash-es';
+import { forEach, union } from 'lodash-es';
 import { trunc } from '~/app/assets/javascripts/utils/strings';
 
 // Components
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
```diff
@@ -63,9 +63,13 @@ export class ArticleViewer extends React.Component {
   // first in that case. In that case, componentDidUpdate fetches the
   // user ids as soon as usernames are avaialable.
   componentDidUpdate(prevProps, prevState) {
-    if (!prevProps.users && this.props.users) {
-      if (!prevState.userIdsFetched) {
-        this.fetchUserIds(this.props.users);
+    if (this.props.assignedUsers && this.props.users) {
```
ragesoss, line 66: It would be helpful to have comments explaining these two different branches where `fetchUserIds` can happen and why it needs to be branched like this. From a user perspective, what are the different situations where each of these branches occurs?

rahahahat (commented):
Hello @ragesoss, apologies for the late reply. I have added some comments that explain the different situations where the branches could occur. Please review them.
Thank you very much.

rahahahat:
@ragesoss Hello, im getting this error in Travis-ci, while all i did was add comments. Could you please check it.
```
 1) RevisionScoreImporter#fetch_ores_data_for_revision_id for Wikidata revisions returns a hash with features
2681     Failure/Error: expect(subject[:rating]).to eq('E')
2682     
2683       expected: "E"
2684            got: "D"
2685     
2686       (compared using ==)
2687     # ./spec/lib/importers/revision_score_importer_spec.rb:185:in `block (5 levels) in <top (required)>'
2688     # ./spec/lib/importers/revision_score_importer_spec.rb:183:in `block (4 levels) in <top (required)>'
```

ragesoss:
That's an unrelated error because of some changes in a Wikimedia API. I've got a fix coming soon.

app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
```diff
@@ -63,9 +63,21 @@ export class ArticleViewer extends React.Component {
...
-    if (!prevProps.users && this.props.users) {
-      if (!prevState.userIdsFetched) {
-        this.fetchUserIds(this.props.users);
+    // checks for the availability of 'assignedUsers' and 'users' props.
+    // This condition will only pass when articleViewer is accessed from the Students/Editors tab
+    if (this.props.assignedUsers && this.props.users) {
+    // checks if usersIds required for coloration have already been fetched or not.
+    // if not, userIds are fetched by passing in an array that contains both 'assignedUsers'
+    // and 'users' as an argument to the fetchUserIds function.
+      if (!this.state.userIdsFetched) {
```
ragesoss, line 72: Why does this switch on `state.userIdsFetched` but the other branch switches on `prevState.userIdsFetched`?

It seems like any time `prevState.userIdsFetched` is true, `state.userIdsFetched` will also be true, since it just goes from false to true, not the other way around.

If that's the case, I suggest using a guard clause at the start so that less nesting and repetition is required.

app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
```diff
@@ -63,9 +63,21 @@ export class ArticleViewer extends React.Component {
   // first in that case. In that case, componentDidUpdate fetches the
   // user ids as soon as usernames are avaialable.
   componentDidUpdate(prevProps, prevState) {
-    if (!prevProps.users && this.props.users) {
-      if (!prevState.userIdsFetched) {
-        this.fetchUserIds(this.props.users);
+    // checks for the availability of 'assignedUsers' and 'users' props.
```
ragesoss, line 66: These comments are helpful, but I think they might point the way to some further simplification that is possible.

Here's my understanding:

Wherever `ArticleViewer` is used, the `users` prop represents all the relevant users who have edited the article, and it won't be passed initially...it is only passed once fetchArticleDetails returns data for which users in the course have edited the article. So any time `prevProps.users` is empty but `props.users` is not empty, it means it's time to fetch user Ids if they haven't been fetched yet.

This logic holds whether `assignedUsers` is passed or not, so we could get rid of the `else` branch altogether by using something that works whether or not `assignedUsers` is there, like `union(this.props.assignedUsers || [] , this.props.users)`

rahahahat: Thanks alot for the input, i have improved upon the solution by keeping all these things in mind, just waiting for travis-ci to finish.

rahahahat: Hello @ragesoss, i have updated the solution, please review it.

app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
```diff
@@ -116,7 +124,13 @@ export class ArticleViewer extends React.Component {
...
       this.props.fetchArticleDetails();
     } else if (!this.state.userIdsFetched && !this.props.showArticleFinder) {
-      this.fetchUserIds(this.props.users);
+      // if articleViewer is accessed through Students/Editors tab, a combination
+      // of both assignedUsers and users will be passed to the fetchUserIds function
+      // when show is clicked and the conditions have been satisfied. Otherwise
+      // whenever the articleViewer is accessed through any other tab, e.g Articles tab,
+      // the fetchUserIds function will be invoked with users because
+      // assignedUsers would be undefined.
+      this.fetchUserIds(union(this.props.assignedUsers || [], this.props.users));
```
ragesoss, line 133: I think this is just about ready!

Since this is the key strategy for making the set of users consistent and it gets used in two different places, I think it would make sense to refactor this further. The arguments being passed to `fetchUserIds()` are just derived from props, so we don't actually need to pass them at all; constructing the array of users from the props could be done within the `fetchUserIds()` function, or a new function could be extracted which basically just does `this.fetchUserIds(union(this.props.assignedUsers || [], this.props.users));`. Either way, this would reduce duplication and give you only one place that requires documentation of what's going on.

rahahahat:
@ragesoss 
Just wanted to say thank you for helping me fix the bug. I've really learned alot.

ragesoss:
Great! Thanks for tackling it!

Issue #4022, opened 2020-05-08
articles assigned to other students in the class don't show up as 'other edited articles'


ragesoss:
### Current Behavior:
The details view on the Students tab shows all the assigned articles for the selected student, as well as the articles they have signed up to peer review, and a list of other unassigned articles they've made edits to.

However, if a student has edited an article that is not assigned to them but is assigned to a classmate, it won't show up at all on this view.

For example, this student has edited "Eugenics in the United States", which is assigned to a classmate:
![assigned](https://user-images.githubusercontent.com/848483/81452147-f30bf800-913a-11ea-82f2-e1fad8b2fbfb.png)

### Desired Behavior:
The Students details view should show every mainspace article that student has edited, either in the assigned articles list or in the 'other articles edited' list. An article being assigned to a classmate shouldn't affect this.

BaileyMcKelway:
Looking into this!

rahahahat:
I would like to work on this bug.

Bam92:
Hi @rahahahat!

Are you already working on this issue? If not, let me know.

rahahahat:
@Bam92 working on it.

rahahahat:
@Bam92 progress has been a bit slower bit definitely working on it

rahahahat:
@ragesoss should the fix take into consideration all available articles as well all those articles that have been assigned to students, or only those that have been assigned to other classmates but have been edited by some who aren't assigned to them.

ragesoss:
Basically, every article edited by a student that shows up in the Articles tab should also be shown in the Students view when looking at that student. So if it's not one of that student's assigned articles, it should show up in the 'other articles edited' section. That includes both classmates' assignments and available articles.

rahahahat:
> Basically, every article edited by a student that shows up in the Articles tab should also be shown in the Students view when looking at that student. So if it's not one of that student's assigned articles, it should show up in the 'other articles edited' section. That includes both classmates' assignments and available articles.

thanks for the input, i'll get to work

rahahahat:
@ragesoss Apologies for the late update, i've been working on the bug wanted to know whether an assigned article should come up under other edited articles as well?

ragesoss:
No, it should not if it's assigned to that user. If it's only assigned to
someone else, it should show up.

rahahahat:
@ragesoss I have opened a pull request. Please, review it.

rahahahat:
[...]

PR #4284, opened 2020-11-10
Solves Issue #4022 "articles assigned to other students in the class don't show up as other edited articles".

rahahahat:
## What this PR does
This pull request resolves Issue #4022, wherein articles assigned to other students in the class don't show up as 'other edited articles'.

## Screenshots
Before:

After:

## Outline of the solution
This Bug was present in the `setUnassignedArticles.js` file. The current implementation of the `setUnassignedArticles` function filtered out any article that was edited by an unassigned editor if the same article was present in the assignments array either as an assigned article to a classmate or as an available article yet to be assigned. While solving this bug I came across five cases which needed to be handled because the assignmets array could contain multiple assignments having the same article and differing only in assignees or existing as an assignment yet to be assigned to an editor. Here are some `console.log` statements that depict the same:

**The five cases to be handled are:**
**1)** Edited article isn't availabe as an assignment.
**2)** Edited article is available as an assignment but hasn't been assigned to any editor yet.
**3)** Edited article is not available as an assignment but has been assigned to some editors but not all who have edited the article.
**4)** Edited article has been assigned to user/users but is also available as an assignment to other users.
**5)** Edited article has been assigned to all users who have edited the article.

My approach towads the solving the bug includes editing the reducer function of  `assignments.reduce()` such that if multiple assignments had the same `article/articleId` and an assignement with the same `article/articleId` was already present in the `accumulator`, then only the editor assigned to that assignement would be added to the `assignees` property of the article object in the accumulator. If the assignment had no assigned editor then a `null` was added to `assignees`. At a time there can only be 1 unassigned assignment with a particular article. WikiEduDashboard prevents having multiple unassigned assignments with the same article, so there can only be 1 null entry in the `assignees` array.

The assignments array is then reduced to an object - `assignmentsByArticleId` that is sorted/grouped by articleIds, each id referencing its corresponding article.

After this the solution checks if any articles in `assignmentsByArticleId` and "articles edited by an editor" are common and have the same editor. If present those articles are not included. if not they are included in the final array.

### DIFF
--- a/app/assets/javascripts/components/students/components/Articles/SelectedStudent/SelectedStudent.jsx
+++ b/app/assets/javascripts/components/students/components/Articles/SelectedStudent/SelectedStudent.jsx
@@ -11,7 +11,7 @@ import EditedUnassignedArticles from './EditedUnassignedArticles/EditedUnassigne
 
 // Utils
 import { processAssignments } from '@components/overview/my_articles/utils/processAssignments';
-import setUnassignedArticles from '@components/students/utils/setUnassignedArticles';
+import setOtherEditedArticles from '@components/students/utils/setOtherEditedArticles';
 
 export const SelectedStudent = ({
   groupedArticles, assignments, course, current_user, fetchArticleDetails,
@@ -21,7 +21,7 @@ export const SelectedStudent = ({
   const {
     assigned, reviewing
   } = processAssignments({ assignments, course, current_user: selected });
-  const unassigned = setUnassignedArticles(groupedArticles, assignments, selected);
+  const otherEditedArticles = setOtherEditedArticles(groupedArticles, assignments, selected);
   const showArticleId = Number(location.search.split('showArticle=')[1]);
 
   return (
@@ -78,9 +78,9 @@ export const SelectedStudent = ({
       }
 
       {
-        !!unassigned.length && (
+        !!otherEditedArticles.length && (
           <EditedUnassignedArticles
-            articles={unassigned}
+            articles={otherEditedArticles}
             course={course}
             user={selected}
             current_user={current_user}
--- a/app/assets/javascripts/components/students/utils/groupArticlesCoursesByUserId.js
+++ b/app/assets/javascripts/components/students/utils/groupArticlesCoursesByUserId.js
@@ -9,7 +9,6 @@ export default (articles) => {
         acc[id] = [article];
       }
     });
-
     return acc;
   }, {});
 };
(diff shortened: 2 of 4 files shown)


ragesoss:
This makes sense and works as expected. Thanks for the thorough explanation of both the possible situations to account for as well as how the code itself works.

The the thing I'd suggest is improving the name of the file and the variables... this isn't really about unassigned articles, it's about edited-but-unassigned articles (aka, 'other edited articles'). Changing names in both this file and `SelectedStudent.jsx` where it's called would make it easier to understand what this code is intended to do.

rahahahat:
@ragesoss thanks for the review, i'll change the names immediately.

rahahahat:
@ragesoss Hello, I have Improved the variable names. Please review it.

rahahahat:
Getting CORS errors on the latest Travis-CI check: 
```
 1) Training 'editing-basics' module lets the user go from start to finish
     Got 2 failures from failure aggregation block "javascript errrors".
     # ./spec/rails_helper.rb:109:in `block (2 levels) in <top (required)>
     1.1) Failure/Error: expect(error.level).not_to eq('SEVERE'), error.message
            https://www.youtube-nocookie.com/embed/bfxjrtot1M8 - Access to XMLHttpRequest at 'https://googleads.g.doubleclick.net/pagead/id' from origin 'https://www.youtube-nocookie.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
          # ./spec/rails_helper.rb:115:in `block (4 levels) in <top (required)>'
          # ./spec/rails_helper.rb:110:in `each'
          # ./spec/rails_helper.rb:110:in `block (3 levels) in <top (required)>'
     1.2) Failure/Error: expect(error.level).not_to eq('SEVERE'), error.message
            https://www.youtube-nocookie.com/embed/BU5Ga_Yz8ZU - Access to XMLHttpRequest at 'https://googleads.g.doubleclick.net/pagead/id' from origin 'https://www.youtube-nocookie.com' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
          # ./spec/rails_helper.rb:115:in `block (4 levels) in <top (required)>'
          # ./spec/rails_helper.rb:110:in `each'
          # ./spec/rails_helper.rb:110:in `block (3 levels) in <top (required)>'
```

rahahahat:
@ragesoss The build has passed all tests, please review the changes.

PR #4312, opened 2020-12-30
Include number of articles edited, and number of articles created, for each user in Editors CSV

Gchahm:
## What this PR does
This PR implements the desired behavior requested in issue #4217

## Screenshots
Before:
image

After:
image

The solution for both created and edited was based off how course.new_article_count is updated in course_cache_manager (line 82).

For the edit part I've unscoped the filter on new_article.

I've chosen to hit the database only twice to get the values for all employees and then loop through the result for each employee as I assume this will be faster than 2 SQL queries per employee, but I'm not sure this is correct.

### DIFF
--- a/lib/analytics/course_students_csv_builder.rb
+++ b/lib/analytics/course_students_csv_builder.rb
@@ -5,6 +5,8 @@ require 'csv'
 class CourseStudentsCsvBuilder
   def initialize(course)
     @course = course
+    @created_articles = course.new_articles.select(:article_id, :user_ids).to_a
+    @edited_articles = course.new_articles.select(:article_id, :user_ids).unscope(where: :new_article).to_a
   end
 
   def generate_csv
@@ -31,6 +33,8 @@ class CourseStudentsCsvBuilder
     draft_space_bytes_added
     references_added
     registered_during_project
+    total_articles_created
+    total_articles_edited
   ].freeze
   def row(courses_user)
     row = [courses_user.user.username]
@@ -42,9 +46,19 @@ class CourseStudentsCsvBuilder
     row << courses_user.character_sum_draft
     row << courses_user.references_count
     row << newbie?(courses_user.user)
+    row << total_articles_created(courses_user.user_id)
+    row << total_articles_edited(courses_user.user_id)
   end
 
   def newbie?(user)
     (@course.start..@course.end).cover? user.registered_at
+  end
+
+  def total_articles_created(user_id)
+    @created_articles.count {|a| a[:user_ids].include?(user_id)}
+  end
+
+  def total_articles_edited(user_id)
+    @edited_articles.count {|a| a[:user_ids].include?(user_id)}
   end
 end
--- a/lib/analytics/course_students_csv_builder.rb
+++ b/lib/analytics/course_students_csv_builder.rb
@@ -7,10 +7,11 @@ class CourseStudentsCsvBuilder
     @course = course
     @created_articles = Hash.new(0)
     @edited_articles = Hash.new(0)
-    @revisions = @course.all_revisions.where(new_article: true).pluck(:article_id, :user_id).to_h
   end
 
   def generate_csv
+    @new_article_revisions = @course.all_revisions.where(new_article: true)
+                                    .pluck(:article_id, :user_id).to_h
     populate_created_articles
     populate_edited_articles
     csv_data = [CSV_HEADERS]
@@ -34,7 +35,7 @@ class CourseStudentsCsvBuilder
   end
 
   def article_creator?(article_id, user_id)
-    @revisions[article_id] == user_id
+    @new_article_revisions[article_id] == user_id
   end
 
   def populate_edited_articles
@@ -61,6 +62,7 @@ class CourseStudentsCsvBuilder
     total_articles_created
     total_articles_edited
   ].freeze
+  # rubocop:disable Metrics/AbcSize
   def row(courses_user)
     row = [courses_user.user.username]
     row << courses_user.created_at
@@ -74,6 +76,7 @@ class CourseStudentsCsvBuilder
     row << @created_articles[courses_user.user_id]
     row << @edited_articles[courses_user.user_id]
   end
+  # rubocop:enable Metrics/AbcSize
 
   def newbie?(user)
     (@course.start..@course.end).cover? user.registered_at
--- a/lib/analytics/course_students_csv_builder.rb
+++ b/lib/analytics/course_students_csv_builder.rb
@@ -5,11 +5,14 @@ require 'csv'
 class CourseStudentsCsvBuilder
   def initialize(course)
     @course = course
-    @created_articles = course.new_articles.select(:article_id, :user_ids).to_a
-    @edited_articles = course.new_articles.select(:article_id, :user_ids).unscope(where: :new_article).to_a
+    @created_articles = Hash.new(0)
+    @edited_articles = Hash.new(0)
+    @revisions = @course.all_revisions.where(new_article: true).pluck(:article_id, :user_id).to_h
   end
 
   def generate_csv
+    populate_created_articles
+    populate_edited_articles
     csv_data = [CSV_HEADERS]
     courses_users.each do |courses_user|
       csv_data << row(courses_user)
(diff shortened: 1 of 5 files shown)


lib/analytics/course_students_csv_builder.rb
```diff
@@ -5,6 +5,8 @@
 class CourseStudentsCsvBuilder
   def initialize(course)
     @course = course
+    @created_articles = course.new_articles.select(:article_id, :user_ids).to_a
+    @edited_articles = course.new_articles.select(:article_id, :user_ids).unscope(where: :new_article).to_a
```
ragesoss, line 9: If a user creates a new article, it should get counted as both an edited article and a new article. It looks like this `unscope` clause would make it exclude new articles from the edited count.

spec/lib/analytics/course_students_csv_builder_spec.rb
```diff
@@ -5,14 +5,40 @@
...
+
+    expected_result = {
+      'first_user' => { created: '1', updated: '1' },
+      'second_user' => { created: '0', updated: '1' },
+      'third_user' => { created: '0', updated: '0' }
+    }
+
+    lines.each do |line|
+      columns = line.split(',')
+      user_result = expected_result[columns[0]]
```
ragesoss, line 37: It took me a little while to understand what this is trying to do, and in particular, to realize that `columns[0]` is supposed to be the username, which is then used to look up the expected results for that user in the `expected_results` hash.

I think there's room to refactor this so that it's easier to follow.

lib/analytics/course_students_csv_builder.rb
Gchahm: Unscope will remove the new_artricle = true from the query, and will cause both new  / edits to be counted. This is being validaded by the changes in the test file.

see SQL output from rails console

image

lib/analytics/course_students_csv_builder.rb
```diff
@@ -42,9 +46,19 @@ def row(courses_user)
...
+    row << total_articles_created(courses_user.user_id)
+    row << total_articles_edited(courses_user.user_id)
   end
 
   def newbie?(user)
     (@course.start..@course.end).cover? user.registered_at
   end
+
+  def total_articles_created(user_id)
+    @created_articles.count {|a| a[:user_ids].include?(user_id)}
```
ragesoss, line 58: This needs to account for a situation where a newly created article was edited by multiple users in the course. In that situation, only the user who actually created it should have it counted as a created article, while everyone who edited it should have it counted as an edited article.

lib/analytics/course_students_csv_builder.rb
ragesoss: Ah! I missed that it was starting from the `new_articles` query. I think this is confusing, and probably shouldn't be don't starting from `course.new_articles`.

I'm immediately sure why the `new_articles` method in `Course` is as complex as it is (in particular, I'm not sure why it's using a join to the articles table), but it might make sense to make a parallel `edited_articles` method in `Course` that doesn't add the `new_article` scope in the first place, rather than unscoping the existing method.

ragesoss: Hmm... after digging into it, I'm pretty sure that join is unnecessary, and `Course#new_articles` can be simpler... it's also confusing that it's named `new_articles` but it actually returns `ArticlesCourses` objects and not `Article` objects.

I'll work on a refactor of the relevant `Course` methods today that will pave the way for updating your PR.

lib/analytics/course_students_csv_builder.rb
```diff
@@ -17,6 +22,28 @@ def generate_csv
...
+  def populate_created_articles
+    # A user has created an article during the course if
+    # the user is in the user_ids of new_articles_courses
+    # has a revision with new_article: true for that article
+    @course.new_articles_courses.pluck(:article_id, :user_ids).each do |article_id, user_ids|
+      user_ids.each do |user_id|
+        @created_articles[user_id] += 1 if article_creator?(article_id, user_id)
+      end
+    end
+  end
```
Gchahm, line 34: This goes through the results of new articles and checks which user in the user_ids is the creator by looking at the revisions of the course.
at the end of this method the created_articles should be a hash with user_id as keys and the count of created articles as values.

Gchahm, line 34: This goes through the results of new articles and checks which user in the user_ids is the creator by looking at the revisions of the course.
at the end of this method the created_articles should be a hash with user_id as keys and the count of created articles as values.

lib/analytics/course_students_csv_builder.rb
```diff
@@ -17,6 +22,28 @@ def generate_csv
...
+  def article_creator?(article_id, user_id)
+    @revisions[article_id] == user_id
+  end
+
+  def populate_edited_articles
+    # A user has edited an article if the user is in the user_ids list of edited_articles_courses
+    @course.edited_articles_courses.pluck(:article_id, :user_ids).each do |_article_id, user_ids|
+      user_ids.each { |user_id| @edited_articles[user_id] += 1 }
+    end
+  end
```
Gchahm, line 45: Similar to the populate edit articles but in this case the user gets a count for every line it's included.
I've decided to change to these populated methods because this way it iterates through the SQL result only once instead of once per user as it was being done previously.

lib/analytics/course_students_csv_builder.rb
```diff
@@ -17,6 +22,28 @@ def generate_csv
...
+
+  def article_creator?(article_id, user_id)
+    @revisions[article_id] == user_id
+  end
+
+  def populate_edited_articles
+    # A user has edited an article if the user is in the user_ids list of edited_articles_courses
+    @course.edited_articles_courses.pluck(:article_id, :user_ids).each do |_article_id, user_ids|
```
ragesoss, line 42: Is it necessary to pluck the article_id here?

lib/analytics/course_students_csv_builder.rb
```diff
@@ -5,9 +5,14 @@
 class CourseStudentsCsvBuilder
   def initialize(course)
     @course = course
+    @created_articles = Hash.new(0)
+    @edited_articles = Hash.new(0)
+    @revisions = @course.all_revisions.where(new_article: true).pluck(:article_id, :user_id).to_h
```
ragesoss, line 10: This looks like an efficient way to find the user who creates each new article. Name the instance variable more clearly, eg, `@new_article_revisions`.

lib/analytics/course_students_csv_builder.rb
Gchahm: I think so because pluck is using select DISTINCT so if I leave out article_id it'll return just the ammount of combinations of user id's lists

image

lib/analytics/course_students_csv_builder.rb
Gchahm: I'll get this done, should I also move this to generate_csv method? So all the computation happens there instead of some happening in new?

lib/analytics/course_students_csv_builder.rb
ragesoss: cool

lib/analytics/course_students_csv_builder.rb
ragesoss: Sure, seems reasonable. Either way would be fine with me.

lib/analytics/course_students_csv_builder.rb
```diff
@@ -42,6 +71,8 @@ def row(courses_user)
     row << courses_user.character_sum_draft
     row << courses_user.references_count
     row << newbie?(courses_user.user)
+    row << @created_articles[courses_user.user_id]
+    row << @edited_articles[courses_user.user_id]
```
Gchahm, line 75: rubocop is complaining the size of this method is too high 

Metrics/AbcSize: Assignment Branch Condition size for row is too high. [<1, 26, 0> 26.02/23]

I suppose I should supress this somehow or just ignore?

ragesoss: yeah, you can suppress it by putting comments before and after the method:

`# rubocop:disable Metrics/AbcSize`
`# rubocop:enable Metrics/AbcSize`

For a method like this where it's following a very clear pattern many times, that complexity metric isn't a useful indicator that the code is going to be hard to understand or change.

Issue #4314, opened 2020-12-31
There should be a way to see the problems for an individual course: Alerts, possible plagiarism, and deleted articles


ragesoss:
### Current Behavior:
We generate a lot of different information about problems that come up during a course, including `Alert` records, automatically detected edits that might be plagiarism, and deleted revisions and articles. There's no UI to see this stuff, however.

### Desired Behavior:
This probably makes sense as a new tab of the Course component, maybe as a subtab of Activity, and could be implemented in several stages:

1. Show the `Alerts` for a course. We already have components to do this, which are used in the Campaign Alerts tab and the admin-only `alerts_list` page. This should work just like the Campaign Alerts page, except showing only the alerts for that course. Admins should get all alerts, while other users should only get public alerts.
2. Show edits that might be plagiarism, ie, course revisions that have an `ithenticate_id` attribute. For each one, it should include a link to the revision, a link to the article, and (only for admins or instructors) a link to the plagiarism report.
3. Deleted revisions and articles

Gchahm:
Ok I've done my initial check for the first part and here's what I think needs to be done for it.
### Rails
1. add method alert in courses_controller (similar to the existing campains_controller method)
2. create the routes
3. add view / json builder for the method here the json builder will be different from what already exists in campains because it needs to return course.alerts for admins and course.public_alerts for non admins

### React
I'm not so sure of this yet but I suppose the steps are something like
1. add new tab item for courses alert
2. load alerts from the new method in redux and let the container do the rest (?)

I'll start working on this today can you assign this to me please?

ragesoss:
Yep, that sounds like the right shape of things!

Gchahm:
Hey Ros,

I can see potential for some refactor on the [alert_actions](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/actions/alert_actions.js)

Basically for this work I'll need to add 2 new functions in this file **fetchCourseAlertsPromise** and **fetchCourseAlerts**

They will be pretty similar to **fetchCampaignAlertsPromise** and **fetchCampaignAlerts** the only real difference would be the url they access.

So the first part would be to create a function **requestAlerts** like the below and have all the promisses including **fetchAdminAlertsPromise** to call this

```js
const requestAlerts = (url) => {
  return request(url, {
    credentials: 'include'
  }).then(fetchResponseToJSON)
    .catch((error) => {
      logErrorMessage(error);
      return error;
[...]

PR #4320, opened 2021-01-07
Course alerts Component Implementation for issue #4314 (part 1)

Gchahm:
## What this PR does
Implement the first part of desired behavior if issue #4314

## Screenshots
Before:
image

After:
image
image

- There was already a CourseAlert component, so the new Component was called CourseAlertList.
- The CampaignAlerts component came with 2 default filters that dind't make sense and are not even selectable from when trying to add a new filter so did not use any default filter for the new component.
- The alert_action was refactored to be more DRY and there might be two unused functions that can be removed of the file as well submitBadWorkAlert, submitNeedHelpAlert
- I will add tests in the next commit

### DIFF
--- a/app/assets/javascripts/components/activity/activity_handler.jsx
+++ b/app/assets/javascripts/components/activity/activity_handler.jsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import createReactClass from 'create-react-class';
+import { withRouter } from 'react-router';
+import { Redirect, Route, Switch } from 'react-router-dom';
+
+import SubNavigation from '../common/sub_navigation.jsx';
+import CourseAlertsList from '../alerts/course_alerts_list';
+import RevisionHandler from '../revisions/revisions_handler';
+
+export const ActivityHandler = createReactClass({
+  displayName: 'ActivityHandler',
+
+  propTypes: {
+    course_id: PropTypes.string,
+    current_user: PropTypes.object,
+    course: PropTypes.object,
+  },
+
+  render() {
+    const links = [
+      {
+        href: `/courses/${this.props.course.slug}/activity/recent`,
+        text: I18n.t('application.recent_activity')
+      },
+      {
+        href: `/courses/${this.props.course.slug}/activity/alerts`,
+        text: I18n.t('courses.alerts')
+      }
+    ];
+
+    return (
+      <div className="activity-handler">
+        <SubNavigation links={links} />
+        <Switch>
+          <Route exact path="/courses/:course_school/:course_title/activity/recent" render={() => <RevisionHandler {...this.props}/>} />
+          <Route exact path="/courses/:course_school/:course_title/activity/alerts" render={() => <CourseAlertsList {...this.props} />} />
+          <Redirect to={{ pathname: '/courses/:course_school/:course_title/activity/recent' }}/>
+        </Switch>
+      </div>
+    );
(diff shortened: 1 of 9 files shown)


ragesoss:
Nice! Rather than a new tab dedicated just to this, I think this should become a sub-tab of the Activity tab. This is similar to how we have different views within the Articles tab, where you can switch between Articles Edited, Assigned Articles, and Available Articles.

Gchahm:
I've updated the pull request now, I'll be working on adding tests tomorrow.

app/assets/javascripts/components/common/course_navbar.jsx
```diff
@@ -58,6 +58,7 @@ const CourseNavbar = ({ course, location, currentUser, courseLink }) => {
   const articlesLink = `${courseLink}/articles`;
   const uploadsLink = `${courseLink}/uploads`;
   const activityLink = `${courseLink}/activity`;
+  const alertsLink = `${courseLink}/alerts`;
```
Gchahm, line 62: I forgot to remove this on the last commit will get it in the next

ragesoss (changes requested):
This looks great. On trying it out, I'm realizing that the amount of info that's available from the alerts list is pretty limited, and doesn't translate that well from how it's used in the other two locations. From `alerts_list`, it links to a view where an admin can see all the details for a given alert. From the campaign alerts tab, the default filters are for alert types that are pretty self-explanatory for the people who visit that page, and knowing the course and/or article for each alert is all you really need to see.

In the context of an individual course, where the people seeing these alerts can be an instructor or student in the class, some of the alerts that will show up may not make much sense without extra context and details (such as a link to a 'show' view for each alert, similar to how the `alerts_list` works) as well as user-facing definitions of what each type of alert means.

However, this will be useful for admins immediately. I think we should make the link to this view in the `Activity` tab only show up for admins, at this point. (It would still work for others, it just wouldn't be discoverable.) Then later on we can rework the list to get it into a shape that's more useful than confusing for instructors and others.

app/controllers/courses_controller.rb
```diff
@@ -117,6 +117,11 @@ def uploads
     set_course
   end
 
+  def alerts
+    set_course
+    @alerts = current_user.roles(@course)[:admin] ? @course.alerts : @course.public_alerts
```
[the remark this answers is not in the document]
Gchahm: I've added a new commit for that.

Gchahm:
I've set the links variable to only include alerts when the user is admin, so the alerts button won't appear for normal users.

app/assets/javascripts/components/activity/activity_handler.jsx
```diff
@@ -0,0 +1,48 @@
...
+  propTypes: {
+    course_id: PropTypes.string,
+    current_user: PropTypes.object,
+    course: PropTypes.object,
+  },
+
+  render() {
+    const links = [
+      {
+        href: `/courses/${this.props.course.slug}/activity/recent`,
```
ragesoss, line 23: Having just one link present doesn't make much sense. I suggest instead only rendering the `SubNavigation` component at all if the user is an admin.

Gchahm: That makes sense, I'll get it sorted.

Gchahm: I've changed this now.

PR #4329, opened 2021-01-19
Course plagiarism report

Gchahm:
## What this PR does
Implements the second part of issue #4314

## Screenshots
Before:
image

After:
image

The new component is very similar to the existing recent activity component but with only the relevant information to this feature.
The json builder currently contains much more info than required but I'll wait for a review to update it just in case there is more to be added in the front end.

### DIFF
--- a/app/assets/javascripts/components/suspected_plagiarism/suspected_plagiarism_handler.jsx
+++ b/app/assets/javascripts/components/suspected_plagiarism/suspected_plagiarism_handler.jsx
@@ -1,0 +1,57 @@
+import React from 'react';
+import createReactClass from 'create-react-class';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import {
+  fetchSuspectedCoursePlagiarism,
+  sortSuspectedPlagiarism
+} from '../../actions/suspected_plagiarism_actions';
+import SuspectedPlagiarismList from './suspected_plagiarism_list';
+
+  const PossiblePlagiarismHandler = createReactClass({
+  displayName: 'PossiblePlagiarismHandler',
+
+  propTypes: {
+    course_id: PropTypes.string,
+    course: PropTypes.object,
+    sort: PropTypes.object,
+    suspectedPlagiarism: PropTypes.array,
+    loading: PropTypes.bool,
+    fetchSuspectedCoursePlagiarism: PropTypes.func,
+    sortSuspectedPlagiarism: PropTypes.func,
+  },
+
+
+  componentDidMount() {
+    if (this.props.loading) {
+      this.props.fetchSuspectedCoursePlagiarism(this.props.course_id);
+    }
+  },
+
+  render() {
+    return (
+      <div id="possible-plagiarism">
+        <SuspectedPlagiarismList
+          revisions={this.props.suspectedPlagiarism}
+          course={this.props.course}
+          loaded={!this.props.loading}
+          sortBy={this.props.sortSuspectedPlagiarism}
+          sort={this.props.sort}
+        />
+      </div>
+    );
+  }
+});
+
+const mapStateToProps = state => ({
+  suspectedPlagiarism: state.suspectedPlagiarism.revisions,
+  loading: state.suspectedPlagiarism.loading,
+  sort: state.suspectedPlagiarism.sort,
+});
+
+const mapDispatchToProps = {
+  fetchSuspectedCoursePlagiarism,
+  sortSuspectedPlagiarism
+};
+
+export default connect(mapStateToProps, mapDispatchToProps)(PossiblePlagiarismHandler);
--- a/app/assets/javascripts/components/suspected_plagiarism/suspected_plagiarism_handler.jsx
+++ b/app/assets/javascripts/components/suspected_plagiarism/suspected_plagiarism_handler.jsx
@@ -8,7 +8,7 @@ import {
 } from '../../actions/suspected_plagiarism_actions';
 import SuspectedPlagiarismList from './suspected_plagiarism_list';
 
-  const PossiblePlagiarismHandler = createReactClass({
+const PossiblePlagiarismHandler = createReactClass({
   displayName: 'PossiblePlagiarismHandler',
 
   propTypes: {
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/components/suspected_plagiarism/suspected_plagiarism_handler.jsx
```diff
@@ -0,0 +1,57 @@
...
+import createReactClass from 'create-react-class';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import {
+  fetchSuspectedCoursePlagiarism,
+  sortSuspectedPlagiarism
+} from '../../actions/suspected_plagiarism_actions';
+import SuspectedPlagiarismList from './suspected_plagiarism_list';
+
+  const PossiblePlagiarismHandler = createReactClass({
```
ragesoss, line 11: I had totally forgotten that we already had a view for showing recent possible plagiarism across the whole system. This could probably be done by re-using the existing component (which shows up at `/recent-activity/plagiarism`), although making a new component as you've done is also fine.

app/controllers/courses_controller.rb
```diff
@@ -122,6 +122,12 @@ def alerts
     @alerts = current_user&.admin? ? @course.alerts : @course.public_alerts
   end
 
+  def suspected_plagiarism
+    set_course
+    @revisions = @course.suspected_plagiarism
+    @show_report = current_user&.admin?
```
ragesoss, line 128: This should be available to both admins and course instructors. You can do that with `current_user&.can_edit?(@course)`; if a user is allowed to edit a course, it they are either an admin or an instructor for the particular course.

ragesoss (changes requested):
I haven't tested it locally yet, but this looks solid. I'll test it once the permissions tweak is implemented.

app/assets/javascripts/components/suspected_plagiarism/suspected_plagiarism_handler.jsx
Gchahm: I tried to use that component but it used a drawer in the list to show / hide all courses the revision belongs to which wouldn't make much sense in this case. The alternative would be refactor the code to only render the drawer if a flag is passed to it, (or maybe it does make sense to show the drawer). I don't mind going this route if that's a better fit.

ragesoss:
This works smoothly. Trying it out, I realize there are a couple of additional things it needs before we can deploy it:

* The 'none' message needs to be specific. It currently says "This course has no recent Wikipedia editing activity.", which might be confusing.
* It needs some way of presenting an explanation of how to interpret this list of "possible plagiarism". We currently have some explanation in the emails that instructors get whenever a new possible plagiarism revision is detected: https://dashboard-testing.wikiedu.org/rails/mailers/suspected_plagiarism_mailer/content_expert_email . Maybe a good way to handle this could be to add a header (like the `Recent Activity` header in that subtab) and use an info mouseover for some text adapted from that email (which can be found in the codebase at `/app/views/suspected_plagiarism_mailer/content_expert_email.hml.haml`)

config/locales/qqq.yml
```diff
@@ -626,6 +626,10 @@ qqq:
     no_edits: Message stating there are no recent edits.
     plagiarism_report: Label for link to a plagiarism report (Wiki Ed only)
     possible_plagiarism: Label for Possible Plagiarism tab (Wiki Ed only)
+    plagiarism_explanation: |-
+    Message that explains that the report might be a false positive.
```
ragesoss, line 630: This needs to be indented. I think that's the cause of the test failures.

config/locales/en.yml
```diff
@@ -944,6 +944,11 @@ en:
     no_edits: There are no recent edits.
     plagiarism_report: Plagiarism Report
     possible_plagiarism: Possible Plagiarism
+    plagiarism_explanation: >
+      These reports are not 100% accurate. For example, a direct quote with proper attribution
```
ragesoss, line 948: I suggest adding this at the beginning: "Wikipedia's plagiarism detection system uses Turnitin/iThenticate to compare new edits with published text."

app/controllers/courses_controller.rb
```diff
@@ -122,6 +122,12 @@ def alerts
     @alerts = current_user&.admin? ? @course.alerts : @course.public_alerts
   end
 
+  def suspected_plagiarism
```
ragesoss, line 125: This method isn't exercised in the specs. I think the ideal thing here would be a new feature spec that loads the activity tab and then clicks the subnav link and looks for a result in the table, since that will also touch your frontend changes. You can see an example of a similar minimal feature spec in `spec/features/article_viewer_spec.rb`.

Gchahm: I've added a new feature to test this, one thing that I couldn't understand though is how to test admin / normal user? It seems it's loging as admin by default.

app/views/courses/suspected_plagiarism.json.jbuilder
```diff
@@ -0,0 +1,13 @@
...
+
+json.revisions @revisions.includes(:article, :user) do |rev|
+  json.call(rev, :url, :user_id, :mw_rev_id, :mw_page_id, :wiki)
+  json.key rev.id
+  json.datetime rev.date
+  json.article_url rev.article.url
+  json.title rev.article.full_title
+  json.references_added rev.references_added
+  json.username rev.user.username
+  json.report_url @show_report ? nil : rev.plagiarism_report_link
```
ragesoss, line 11: This is backwards: it excludes the report for admins/instructors and includes it for others.

spec/features/suspected_plagiarism_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+  before do
+    course.campaigns << Campaign.first
+    create(:courses_user, course: course, user: user)
+    create(:revision, article: article, user: user, date: '2019-01-01')
+    create(:articles_course, course: course, article: article, user_ids: [user.id])
+  end
+
+  it 'should show an empty list when there are no possible plagiarism' do
+    visit "/courses/#{course.slug}/activity/plagiarism"
+    within(:css, '.possible_plagiarism', wait: 20) do
```
ragesoss, line 18: Looking for specific elements based on their css and xpath selectors is okay, but when it's possible to do without them, it makes for a spec that is more readable, more focused on a user perspective, and less likely to break when the related interface components get changed.

In this case, instead of looking for a table with a single entry, I suggest looking for the unique text of the 'no suspected plagiarism' message.

ragesoss (commented):
These tests are being run as a user who isn't logged in. You can find an example of logging in as an admin in `article_finder_spec.rb`. There's a factory to create an admin user for tests, and the `login_as` method is provided in the test environment by the Devise authentication framework. (The `stub_oauth_edit` helper method may be necessary to account for the `verify_edit_credentials` method of `courses_controller`, when an admin or instructor views a course page.)

ragesoss:
Looks great!

PR #4335, opened 2021-02-03
Update Dockerfile

ramushetty:
Error: While trying to setup project using Docker 
Your Ruby version is 2.5.1, but your Gemfile specified 2.7.1
can be resolved by changing the version in the Docker file to **FROM ruby:2.7.1**


ragesoss:
I think the Docker setup has more problems than just the Ruby version, but obviously this won't hurt.

Issue #4332, opened 2021-01-21
Trailing whitespace in i18n messages


Nikerabbit:
### What is happening?
I18n messages cannot have trailing whitespace in translatewiki.net.

### To Reproduce

List of messages:
```
Wikimedia:Wikiedudashboard-courses.campaigns/en
'Campaigns: '                          | 'Campaigns:'                          

Wikimedia:Wikiedudashboard-courses.rating feedback/en
'Predicted rating for the article is:  | 'Predicted rating for the article is:'
'                                      |                                       

Wikimedia:Wikiedudashboard-courses.tags/en
'Tags: '                               | 'Tags:'                               

Wikimedia:Wikiedudashboard-courses generic.campaigns/en
'Campaigns: '                          | 'Campaigns:'                          

Wikimedia:Wikiedudashboard-timeline.empty week 1/en
'To get started, '                     | 'To get started,'                     

Wikimedia:Wikiedudashboard-timeline.empty week 3/en
' or '                                 | ' or'                                 

Wikimedia:Wikiedudashboard-users.course passcode/en
'Passcode: '                           | 'Passcode:'                           

Wikimedia:Wikiedudashboard-users.edits by/en
'Edits by: '                           | 'Edits by:'                           

Wikimedia:Wikiedudashboard-users.my assigned/en
'Assigned: '                           | 'Assigned:'                           

Wikimedia:Wikiedudashboard-users.my reviewing/en
'Reviewing: '                          | 'Reviewing:'  
```

### Expected behavior
If trailing whitespace is not significant, it should be removed from the message to not cause issues at translatewiki.net
If trailing whitespace is significant, a workaround must be used (such as encoding the space, or using a marker that will be stripped, or avoiding constructing sentences from pieces).

ragesoss:
This should be fixed now with @shailesh1028 's PR.

PR #4343, opened 2021-02-07
FIXED : Trailing whitespace in i18n messages

shailesh1028:
As per the error defined in the issue: Trailing whitespace in i18n messages #4332, I have made some changes in config/locales/en.yml.
Earlier                                 Now
'Campaigns: '        |        'Campaigns:'                          
'Predicted rating for the article is:  | 'Predicted rating for the article is:'
'Tags: '                   |           'Tags:'                               
'Campaigns: '        |        'Campaigns:'                          
'To get started, '    |       'To get started,'                     
' or '                       |           'or'                                 
'Passcode: '          |        'Passcode:'                           
'Edits by: '             |         'Edits by:'                           
'Assigned: '           |         'Assigned:'                           
'Reviewing: '         |         'Reviewing:'  
Please review the changes.

### DIFF
--- a/app/assets/javascripts/components/common/AssignCell/AssignCell.jsx
+++ b/app/assets/javascripts/components/common/AssignCell/AssignCell.jsx
@@ -20,7 +20,7 @@ const ArticleLink = ({ content, href, prefix }) => {
 
   return (
     <span>
-      { prefix }{ content }
+      { prefix }&nbsp;{ content }
     </span>
   );
 };
--- a/app/assets/javascripts/components/common/article_viewer_legend.jsx
+++ b/app/assets/javascripts/components/common/article_viewer_legend.jsx
@@ -63,7 +63,7 @@ const ArticleViewerLegend = ({ article, users, colors, status, allUsers, failure
 
   return (
     <div className="user-legend-wrap">
-      <div className="user-legend">{I18n.t('users.edits_by')} </div>
+      <div className="user-legend">{I18n.t('users.edits_by')}&nbsp;</div>
       {userLinks}
       {usersStatus}
     </div>
--- a/app/assets/javascripts/components/common/enroll_button.jsx
+++ b/app/assets/javascripts/components/common/enroll_button.jsx
@@ -152,7 +152,7 @@ const EnrollButton = createReactClass({
       editRows.push(
         <tr className="edit" key="enroll_students">
           <td>
-            <p>{I18n.t('users.course_passcode')}<b>{this.props.course.passcode}</b></p>
+            <p>{I18n.t('users.course_passcode')}&nbsp;<b>{this.props.course.passcode}</b></p>
             <p>{I18n.t('users.enroll_url')}</p>
             <input type="text" readOnly={true} value={enrollUrl} style={{ width: '100%' }} />
             {massEnrollmentLink}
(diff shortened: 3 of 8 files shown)


ragesoss:
This removes the whitespace, but it will make these places in the UI appear without necessary whitespace now. For example, The passcode listing in the Details box will say `Passcode:abcde` instead of `Passcode: abdce`.

In addition to changing the en.yml, we'll need to find where each of those messages are used and update the components to render appropriately without relying on trailing whitespace in the i18n message itself.

shailesh1028:
Basically, the whitespace should not be contained in the message but it should be present while rendering the components. Passcode:abcde should look like Passcode: abcde.[message: value]. This means that the whitespace should be there in the value part depending on whether the value is empty or not. So I have to refactor it where the messages are used in the components. Am I right?
Or do I need to remove the whitespace in the message itself, wherever they are used ... say in language.js files present in public/asset/javascript/i18n files
Please let me know !!

ragesoss:
Yes, refactor how the messages are used in components. The i18n js files in
public/assets are generated from the yml locale files, and shouldn't be
edited directly. (They also are not under version control.)

shailesh1028:
Hi @rageross, I got delayed by an uncalled event. Now I have edited the components such that [message: value] will be shown. But I had one doubt i.e in case if the value is empty then it should show [message: ] or [message:]. 
should a single space be shown if the field is empty?
Or do I need to add a condition in rendering the components if the values are empty? 
I feel that some of the above values won't be empty eg: Passcode, Assigned, Reviewing.
Please guide me !! Thanks

ragesoss:
If the value is empty, then it doesn't matter whether the single space is included or not. These should all be implemented so that they work if the value is empty. At least some of them, like passcode, are sometimes empty.

ragesoss:
Sorry, I missed that you updated this @shailesh1028. Looks like there's one failing test related to your change, and it's a case where there was not any whitespace previously (I think).

If you update the spec, I think that should be fine.

It looks like this changes 10 messages, but it only changes 6 places in the frontend where messages are used. Are there more components that need to be updated?

shailesh1028:
@ragesoss  I have made the changes in the frontend for the following messages:
1) 'Predicted rating for the article is: | 'Predicted rating for the article is:'
2) 'To get started, ' | 'To get started,'
3) ' or ' | 'or'
4) 'Passcode: ' | 'Passcode:'
5)'Edits by: ' | 'Edits by:'
6)'Assigned: ' | 'Assigned:'
7) 'Reviewing: ' | 'Reviewing:'
6 and 7 had been rendered by a common element, so it counts for one change. This is how only 6 changes were visible in the code.
Apart from these changes, for the remaining parts, the code already took care of the whitespace problem. No need to add more whitespace in those cases.
I hope this explains your doubt. Please review it.

ragesoss:
@shailesh1028 thanks, that's helpful! Can you update the failing spec?

shailesh1028:
@ragesoss  Do you want me to resolve the errors occurring during the build of Travis-CI?
If so, then I have checked and it's occurring because of this eslint-loader package. I checked we are using "eslint-loader": "^4.0.2" and "babel-eslint": "^10.1.0" which are good enough to use. 
I checked the lines at which it is throwing errors, and those lines of code are not connected with the changes I did. I could not understand why it failed. Could you help me out?

ragesoss:
I'm referring to this spec which is failing because of a (minor) change introduced in this PR: https://travis-ci.org/github/WikiEducationFoundation/WikiEduDashboard/jobs/759100319#L3025

It's a feature spec that looks for a specific string, and now that string has a whitespace difference, causing the spec to fail.

shailesh1028:
@ragesoss  I did as per what you explained. The build is also working fine. Please review it.

Issue #4351, opened 2021-02-11
Typo in /setup/deb-setup.sh


Electron-2002:
### What is happening?
There is a word extra in the line 'Your developmental environment setup is complete. If `you` there are any errors, please refer to the docs for manual installation, or ask for help.'

### To Reproduce

Line 175 of  '/setup/deb-setup.sh

### Expected behavior

The line should be: 'Your developmental environment setup is complete. If there are any errors, please refer to the docs for manual installation, or ask for help.'

Removing the word 'you'

### System Information

 - OS: [e.g. iOS] PopOS (Apt)
 - Browser [e.g. chrome, safari] Chrome

Electron-2002:
Can I take up this issue?

PR #4352, opened 2021-02-11
Removed grammatical error from deb-setup.sh

Electron-2002:
## What this PR does
Fixes the typo in the file `deb-setup.sh` by removing the word **you**.

Solves #4351


ragesoss (changes requested):
Thanks! It would also be good to change 'developmental to 'development'.

Electron-2002:
> Thanks! It would also be good to change 'developmental to 'development'.

Sure

Issue #4157, opened 2020-08-12
The sandbox URL for an Assignment should be updateable


ragesoss:
### Current Behavior:
When a user gets an Assignment (ie, lists the article they will work on), the dashboard sets as sandbox url for that assignment. The normal behavior is that the sandbox is a userspace subpage that takes the name of the assigned topic. For example, if I (User:Ragesoss) get assigned the article "Selfie" on en.wikipedia.org, then the sandbox URL will be `https://en.wikipedia.org/wiki/User:Ragesoss/Selfie`

If a classmate gets assigned to the same article that another student is already assigned to, their assignment will have the same sandbox URL as the first one, so that they can work together in the same sandbox.

However, sometimes this isn't where the students want to draft their work. Common patterns include:
* a user starts their draft in another location before adding the Assignment on the dashboard
* a user manually begins working in their generic sandbox page, because it is linked from the top of the page on Wikipedia
* users assigned to the same article title are working independently and need separate sandboxes

### Desired Behavior:
The dashboard should have a way for a student or instructor (or admin) to update their sandbox URL to an arbitrary URL.

rahahahat:
@rageross, I would like to work on this issue. However, i do have a few questions as i'm fairly new to rails.
1) For updating the sandbox url, do you specifically mean like an input, which allows the user to edit the `User:"new-url"` part of the url, for allotment of a separate sandbox?
2) Also, is this issue in the same context as #4276, If so should I look at the same files for a starting point. If not, could you please direct me towards a starting point.
Thanks alot.

ragesoss:
1. Yes, something like that.
2. Yes, this is a closely related issue, although I don't thin they directly overlap. On the Rails side of things, the relevant places to look are the `Assignment` model `AssignmentsController`, and perhaps (for example code that interacts with Assignment records and the sandbox_url field) `AssignmentManager`.

rahahahat:
@ragesoss Thanks, i'll get working.

rahahahat:
[...]

PR #4355, opened 2021-02-16
[WIP] The sandbox URL for an Assignment should be updateable - #4157

rahahahat:
## What this PR does
This PR allows an instructor or student to edit an assignment's sandbox URL.

For the frontend part I have defined a new component called `EditSandboxUrl` and attached it to the `AssignmentLinks` component. I have defined an action in `assignments_action.js` which makes a call to the a newly defined route `/assignments/:id/:new_url/update_sandbox_url` which points to a `PATCH` method `update_sandbox_url` in the `Assignments_controller.rb` file. The backend changes aren't completely implemented.
I was wondering how should we go about updating the sandbox url because updating the `User:` part of the sandbox url (`https://en.wikipedia.org/wiki/User:"username"/"article_name"`) could lead to duplication errors.


rahahahat:
@ragesoss any reviews?

ragesoss:
I'll have a look next week.

rahahahat:
@ragesoss sounds good, thanks alot.

db/schema.rb
```diff
@@ -47,7 +47,6 @@
     t.date "average_views_updated_at"
     t.integer "wiki_id"
     t.integer "mw_page_id"
-    t.boolean "tracked", default: true
```
ragesoss, line 50: The changes to schema.rb are not related to the PR and should not be included. (They are artifacts of the local settings at the time the migrations are run; unless we are making changing to the database tables, we usually avoid any changes to the version of schema.rb in the repo.)

ragesoss (commented):
I don't think we need to be concerned about duplication or already-existing pages for this feature... in many cases, the purpose of changing it will be to make the 'sandbox draft' link point to an already-existing page where the student is already underway with creating a draft.

The basic strategy you describe for the implementation so far makes sense. I think for the UI, the 'edit' link needs to be smaller so that it's more clear that it's not just another link but is something related to the sandbox draft link. The 'Change' link for the wiki language and project selector (eg, at https://dashboard.wikiedu.org/article_finder ) might be a good model to follow.

db/schema.rb
rahahahat: > I don't think we need to be concerned about duplication or already-existing pages for this feature... in many cases, the purpose of changing it will be to make the 'sandbox draft' link point to an already-existing page where the student is already underway with creating a draft.
> 
> The basic strategy you describe for the implementation so far makes sense. I think for the UI, the 'edit' link needs to be smaller so that it's more clear that it's not just another link but is something related to the sandbox draft link. The 'Change' link for the wiki language and project selector (eg, at https://dashboard.wikiedu.org/article_finder ) might be a good model to follow.

Thanks alot for the review, I'll get working on it immediately.

rahahahat: > The changes to schema.rb are not related to the PR and should not be included. (They are artifacts of the local settings at the time the migrations are run; unless we are making changing to the database tables, we usually avoid any changes to the version of schema.rb in the repo.)

I'll fix this immediately. I always make sure to filter out all git noise. Sorry this slipped through.

rahahahat:
@ragesoss Hello, I have updated the PR with a working fix. Please review it. Thanks alot.

ragesoss:
@rahahahat I think I missed your request for a review. Sorry! If you're still interested in finishing this, it looks like it needs to be rebased and updated a little bit to account for a change to the SandboxLink component.

rahahahat:
@ragesoss Sorry i missed you comment. I am busy with exams till 8th june. However I'll be happy to work on this issues after my exams are finshed.

ragesoss:
@rahahahat great! good luck on your exams!

PR #4360, opened 2021-02-24
Fixed dotted outline styling on navbar links

UmerMIB:
## What this PR does
Removes outline on focus on the navbar components

## Screenshots
Before:
Screenshot from 2021-02-25 04-22-09

After:
Screenshot from 2021-02-25 04-22-44

This feature has been tested both on chrome and firefox

### DIFF
--- a/app/assets/stylesheets/modules/_nav.styl
+++ b/app/assets/stylesheets/modules/_nav.styl
@@ -26,9 +26,9 @@ nav.top-nav
     font-size 15px
     &:hover, &.active
       color $brand_primary
-      border none !important
+      border none 
       border-bottom 3px solid $brand_primary
-      outline none !important
+      outline none 
 
 
   .container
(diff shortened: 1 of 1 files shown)


UmerMIB:
I have tested it on ubuntu 20.04 chrome and mozilla

app/assets/stylesheets/modules/_nav.styl
```diff
@@ -26,7 +26,9 @@ nav.top-nav
     font-size 15px
     &:hover, &.active
       color $brand_primary
+      border none !important
```
ragesoss, line 29: I'd like to avoid using `!important` as much as possible, as it can make further CSS refactoring and style changes more difficult.

UmerMIB: Let me remove it and push it again

ragesoss:
Hmm... these outlines only show up in the `/training` view, because the training pages use a different stylesheet from the rest of the app. Both stylesheets (main.styl and training.styl) include the `_nav.styl` file you've changed here. Why does the training view include the outlines but the main style does not? If you can find the root cause of the difference — what is preventing outlines in the main.styl file? — we might (or might not) be able to find a better way to fix the problem

UmerMIB:
Sure let me check

PR #4362, opened 2021-03-01
Add X-Request-Start header to Apache

nateberkopec:
Required for New Relic's request queueing calculation to work

I'm not really an Apache wizard - I think this is the correct way to load mod_headers but I'm not sure.

### DIFF
--- a/server_config/apache2.conf
+++ b/server_config/apache2.conf
@@ -231,6 +231,6 @@ LoadModule passenger_module /usr/local/rvm/gems/ruby-2.7.1/gems/passenger-6.0.6/
 # Add header to incoming requests, timestamping them with time since the epoch in microseconds
 # This is required for New Relic's request queueing calculation
 <IfModule !mod_headers.c>
-  LoadModule headers_module modules/mod_headers.so
+  LoadModule headers_module /usr/lib/apache2/modules/mod_headers.so
 </IfModule>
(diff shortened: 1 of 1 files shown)


ragesoss:
I tried this on dashboard-testing.wikiedu.org and Apache fails to start:
`Syntax error on line 234 of /etc/apache2/apache2.conf: Cannot load modules/mod_headers.so into server: /etc/apache2/modules/mod_headers.so: cannot open shared object file: No such file or directory`

ragesoss:
The modules are in `mods-available`. The header directive works after I enabled the headers mod via `a2enmod headers`.

ragesoss:
In debian, looks like loading the module via the apache2.conf would look like `LoadModule headers_module /usr/lib/apache2/modules/mod_headers.so`. That's the contents of `/etc/apache2/mods-available/headers.load`.

nateberkopec:
Fixed, LMK if you want me to squash or if you can squash with the big green button 👍

ragesoss:
I'll push up the actual change I made on the production and staging servers, which is just the final line. Loading the module is taken care of via a2enmod, so I don't think we need anything else in apache2.conf.

ragesoss:
Done with 4547b2cb218a8e387f0fb944259241c77e84dd31

PR #4368, opened 2021-03-05
Fix HTML by creating a valid head tag.

nateberkopec:
+ Adds a head tag
+ Moves vega javascript in the code so that it can render in this new setup
+ Div is not a valid tag inside of a head tag, so we change the wikilanguage/project definitions to a meta tag.

I double checked and `WikiLanguages` and `WikiProjects` are still defined correctly.

### DIFF
--- a/app/views/layouts/application.html.haml
+++ b/app/views/layouts/application.html.haml
@@ -1,13 +1,13 @@
 - content_for :head do
   = dashboard_stylesheet_tag("main")
 
-- content_for :javascripts do
+- content_for :javascripts do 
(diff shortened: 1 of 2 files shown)


app/views/layouts/application.html.haml
```diff
@@ -1,13 +1,13 @@
 - content_for :head do
   = dashboard_stylesheet_tag("main")
 
-- content_for :javascripts do
+- content_for :javascripts do
```
nateberkopec, line 4: ack, 1 sec

PR #4375, opened 2021-03-11
Speed up UpdateCourseStats by 12-30%

nateberkopec:
## What this PR does

As identified earlier, UpdateCourseStats is a major generator of load on Wiki Ed servers. Speeding it up has the potential to decrease load on Wiki Ed servers, or at least to increase throughput of the CourseUpdateWorker job so that courses can be updated more frequently.

In addition, this pull request greatly reduces database load by translating many N+1s into single queries, primarily updating or inserting many rows at once.

There are three improvements:

1. Improving concurrency of average view update by using a `FixedThreadPool`.
2. Fixing an N+1 with Articles/Revisions/Users, and batching them into bigger updates.
3. Fixing an N+1 with ArticleCourses, which went to the Revisions table multiple times (now it only goes once).

## Results

Speed increase of 12-30% depending on scenario. Peak memory usage is unaffected.

Before:
```
### GC state
{:count=>57, :heap_allocated_pages=>1526, :heap_sorted_length=>1526, :heap_allocatable_pages=>0, :heap_available_slots=>621987, :heap_live_slots=>621336, :heap_free_slots=>651, :heap_final_slots=>0, :heap_marked_slots=>473255, :heap_eden_pages=>1526, :heap_tomb_pages=>0, :total_allocated_pages=>1526, :total_freed_pages=>0, :total_allocated_objects=>3899135, :total_freed_objects=>3277799, :malloc_increase_bytes=>156704, :malloc_increase_bytes_limit=>22404975, :minor_gc_count=>48, :major_gc_count=>9, :compact_count=>0, :remembered_wb_unprotected_objects=>5018, :remembered_wb_unprotected_objects_limit=>7568, :old_objects=>464894, :old_objects_limit=>725014, :oldmalloc_increase_bytes=>6939056, :oldmalloc_increase_bytes_limit=>22765743}
### Cold Database
.......... 26.341452   1.252287  27.593739 (195.683865)
### Hot Database
..........  7.663174   0.415003   8.078177 ( 24.650678)
```

After:

```
### GC state
{:count=>54, :heap_allocated_pages=>1537, :heap_sorted_length=>1537, :heap_allocatable_pages=>0, :heap_available_slots=>626487, :heap_live_slots=>625600, :heap_free_slots=>887, :heap_final_slots=>0, :heap_marked_slots=>456201, :heap_eden_pages=>1537, :heap_tomb_pages=>0, :total_allocated_pages=>1537, :total_freed_pages=>0, :total_allocated_objects=>3541208, :total_freed_objects=>2915608, :malloc_increase_bytes=>211984, :malloc_increase_bytes_limit=>16777216, :minor_gc_count=>45, :major_gc_count=>9, :compact_count=>0, :remembered_wb_unprotected_objects=>4622, :remembered_wb_unprotected_objects_limit=>7740, :old_objects=>448328, :old_objects_limit=>729816, :oldmalloc_increase_bytes=>6810416, :oldmalloc_increase_bytes_limit=>22765743}
### Cold Database
.......... 21.863503   0.948639  22.812142 (136.959312)
### Hot Database
..........  6.106491   0.249173   6.355664 ( 21.941323)
```

Speedscope profiler output afterward:

Screen Shot 2021-03-11 at 3 38 11 PM

After this PR, this task spends ~10% of total on importing revisions, 60% on revision score updating, 20% on average views, 5% updating caches and 5% on other.

Further improvements to this job would probably be:

1. Moving the revision score updating into its own job and queue. This queue would be serviced by a single Sidekiq process with `concurrency: 4`. This will ensure maximum throughput while not exceeding the ORES API concurrency limit of 4.
2. Doing the same with average views updating, though probably with far more aggressive concurrency.
3. Refactoring this process into many smaller jobs, which then updates the cache at the end. Perhaps a Sidekiq Batch? This would allow us to parallelize parts of the task that can be parallelized. Multinode is probably a higher priority than this though, because for now, at least on global, CPU is maxed out, so parallelizing does not buy additional throughput.

### DIFF
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -82,7 +82,7 @@ class ArticlesCourses < ApplicationRecord
   def views_since_earliest_revision(revisions)
     return if revisions.blank?
     return if article.average_views.nil?
-    days = (Time.now.utc.to_date - revisions.min { |a,b| a.date <=> b.date }.date.to_date).to_i
+    days = (Time.now.utc.to_date - revisions.min_by(&:date).date.to_date).to_i
     days * article.average_views
   end
 
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -64,9 +64,9 @@ class ArticlesCourses < ApplicationRecord
   end
 
   def update_cache
-    revisions = live_manual_revisions
+    revisions = live_manual_revisions.load
 
-    self.character_sum = revisions.where('characters >= 0').sum(:characters)
+    self.character_sum = revisions.select { |r| r.characters >= 0 }.sum(&:characters)
     self.references_count = revisions.sum(&:references_added)
     self.view_count = views_since_earliest_revision(revisions)
     self.user_ids = associated_user_ids(revisions)
@@ -80,14 +80,14 @@ class ArticlesCourses < ApplicationRecord
   end
 
   def views_since_earliest_revision(revisions)
-    return if revisions.empty?
+    return if revisions.blank?
     return if article.average_views.nil?
-    days = (Time.now.utc.to_date - revisions.order('date ASC').first.date.to_date).to_i
+    days = (Time.now.utc.to_date - revisions.min { |a,b| a.date <=> b.date }.date.to_date).to_i
     days * article.average_views
   end
 
   def associated_user_ids(revisions)
-    return [] if revisions.empty?
+    return [] if revisions.blank?
     revisions.map(&:user_id).compact.uniq
   end
 
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -66,7 +66,7 @@ class ArticlesCourses < ApplicationRecord
   def update_cache
     revisions = live_manual_revisions.load
 
-    self.character_sum = revisions.select { |r| r.characters >= 0 }.sum(&:characters)
+    self.character_sum = revisions.sum { |r| r.characters >= 0 ? r.characters : 0 }
     self.references_count = revisions.sum(&:references_added)
     self.view_count = views_since_earliest_revision(revisions)
     self.user_ids = associated_user_ids(revisions)
--- a/lib/importers/average_views_importer.rb
+++ b/lib/importers/average_views_importer.rb
@@ -11,35 +11,28 @@ class AverageViewsImporter
     update_average_views(to_update)
   end
 
-  # We get some 429 / too many requests errors with 8 per batch.
-  ARTICLES_PER_BATCH = 3
+  # We get some 429 / too many requests errors with 8
+  MAX_HTTP_CONCURRENCY = 3
   def self.update_average_views(articles)
-    article_batches = articles.includes(:wiki).each_slice(ARTICLES_PER_BATCH)
-    article_batches.each do |batch|
-      update_average_views_for_batch batch
+    pool = Concurrent::FixedThreadPool.new(MAX_HTTP_CONCURRENCY)
+    average_views = Concurrent::Hash.new
+    time = Time.zone.today
+
+    # Get the average views data and put it into a concurrency-safe datastructure
+    articles.includes(:wiki).each do |article|
+      pool.post { update_average_views_for_batch(article, average_views, time) }
     end
+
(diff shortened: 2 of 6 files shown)


db/migrate/20210310195204_add_unique_index_to_articles.rb
```diff
@@ -0,0 +1,5 @@
+class AddUniqueIndexToArticles < ActiveRecord::Migration[6.0]
+  def change
+    add_index :articles, [:mw_page_id, :wiki_id], unique: true
```
ragesoss, line 3: It looks like we have many violations of this constraint in Wiki Ed production:
```
Article.all.group(:mw_page_id, :wiki_id).having('count(*) > 1').count.count
 => 3674 
```

I believe this largely or exclusively represent cases where all but one of the duplicate records are marked as `deleted: true`. This is part of what `ArticleStatusManager` does.

nateberkopec: This isn't the most critical part of the PR, so I'll think about what I want to do here.

ragesoss: Okay. This constraint makes a lot of sense, it's just related to some of the harder parts of the system to reason about. It's been a little while since I've taken a shot at it; we might be able to find a way to make it work.

ragesoss: Also, from global production:
```
Article.all.group(:mw_page_id, :wiki_id).having('count(*) > 1').cou
nt.count
 => 16655 
```

lib/importers/average_views_importer.rb
```diff
@@ -11,35 +11,28 @@ def self.update_outdated_average_views(articles)
...
 
-  def self.save_updated_average_views(articles, average_views, average_views_updated_at)
-    Article.transaction do
-      articles.each do |article|
-        article.average_views_updated_at = average_views_updated_at
-        article.average_views = average_views[article.id]
-        article.save
-      end
-    end
+  def self.update_average_views_for_batch(article, average_views, time)
```
ragesoss, line 32: Not a batch. `update_average_views_for_article` ?

app/models/course_data/articles_courses.rb
```diff
@@ -64,9 +64,9 @@ def article_revisions
   end
 
   def update_cache
-    revisions = live_manual_revisions
+    revisions = live_manual_revisions.load
 
-    self.character_sum = revisions.where('characters >= 0').sum(:characters)
+    self.character_sum = revisions.select { |r| r.characters >= 0 }.sum(&:characters)
```
ragesoss, line 69: All of these changes to the cache-updating code go against my incorrect intuitions about what is faster. I've taken a few passes at trying to speed this up before, but usually working from the assumption of 'the database is faster than ruby, so do as much ordering and selecting in the database as you can.

My takeaway is that — even if doing a single calculation in the database is faster — ruby Enumerable operations might be faster than going back to the database multiple times.

nateberkopec: The critical line is the next one - `revisions.sum(&:references_added)`. Because we have to do this, that means `revisions` _must_ be loaded. So, if we have to load it, just load it earlier and do all of the other calculations in-memory. 

If we _didn't_ need to load all `revisions`, doing orders/filtering in the database would probably be faster.

db/migrate/20210310195204_add_unique_index_to_articles.rb
nateberkopec: So, if this is true, how did the code work before? 

Before it was calling 

```ruby 
article = Article.find_by(mw_page_id: article_data['article']['mw_page_id'], wiki_id: @wiki.id)
```

The generated query is:

```
SELECT `articles`.* FROM `articles` WHERE `articles`.`mw_page_id` = 1 AND `articles`.`wiki_id` = 1 LIMIT 1
```

That means that we were selecting a random article out of the set matching these attributes.

app/models/course_data/articles_courses.rb
nateberkopec: Someone on Twitter pointed out this might be a good point to use window aggregate functions. Maybe next time, I think this Ruby version is fast enough.

PR #4378, opened 2021-03-12
Update schema.rb

nateberkopec:
This is what is output when running rake db:migrate from a blank DB.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema.define(version: 2020_10_07_191933) do
 
-  create_table "alerts", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+  create_table "alerts", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4", force: :cascade do |t|
     t.integer "course_id"
     t.integer "user_id"
     t.integer "article_id"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -47,12 +47,11 @@
     t.date "average_views_updated_at"
     t.integer "wiki_id"
     t.integer "mw_page_id"
-    t.boolean "tracked", default: true
```
nateberkopec, line 50: I'm a little confused by this removal, as the migration exists for it.

db/schema.rb
```diff
@@ -260,21 +259,21 @@
     t.index ["wiki_id"], name: "index_courses_wikis_on_wiki_id"
   end
 
-  create_table "faqs", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci", force: :cascade do |t|
```
nateberkopec, line 263: note the collation removal here

db/schema.rb
```diff
@@ -498,7 +497,7 @@
     t.index ["user_id", "training_module_id"], name: "index_training_modules_users_on_user_id_and_training_module_id", unique: true
   end
 
-  create_table "training_slides", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci", force: :cascade do |t|
+  create_table "training_slides", options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci ROW_FORMAT=DYNAMIC", force: :cascade do |t|
```
nateberkopec, line 500: unsure where `ROW_FORMAT` came from

db/schema.rb
```diff
@@ -510,7 +509,7 @@
     t.string "slug"
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
-    t.index ["slug"], name: "index_training_slides_on_slug", unique: true
+    t.index ["slug"], name: "index_training_slides_on_slug", unique: true, length: 191
```
nateberkopec, line 512: This one appears to be correct, as ConvertTrainingModulesToUtf8Mb4 contains this constraint.

db/schema.rb
nateberkopec: Issue resolved: the original migration said the column was type `bool`, which is not a valid type.

PR #4419, opened 2021-04-15
Hashistack

nateberkopec:


config/environments/production.rb
```diff
@@ -35,10 +35,11 @@
   # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for nginx
 
   # Force all access to the app over SSL, use Strict-Transport-Security, and use secure cookies.
-  config.force_ssl = true
+  config.force_ssl = false
```
ragesoss, line 38: I don't think I can merge this until we're actually ready to switch over.

ragesoss:
merged in https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4431

Issue #4549, opened 2021-07-06
Timeline Week date range range meeting days UI is confusing


ragesoss:
### What is happening?
A Week on the Timeline shows a date range that represents the start to end of that week (currently locked to a Sunday-to-Saturday configuration), followed by a list of the days of the week on which the course is scheduled to meet for that week (based on weekdays setting combined with the meeting day exceptions defined for that course).

Especially if there is only one meeting day set, it's easy to misinterpret the meeting day as being connected to the end of the date range... which can be very confusing, as the last day of range is a Saturday while the meeting day is probably not a Saturday.

For example:

![week 1 thursday](https://user-images.githubusercontent.com/848483/124651370-ad86b000-de4f-11eb-9e15-6f2615330b7a.png)

Here, the first week spans August 22 through August 28, but the only actual meeting day is Thursday, August 26.

### Expected behavior

The visual design should more clearly convey meeting days (possibly including the dates of each meeting day) as separate information from the date range of the week itself.

PR #4643, opened 2021-10-18
Add dates to meeting days in the timeline

1v4n4:
This PR is a solution for the issue  Timeline Week date range meeting days UI is confusing #4549. 
The aim was a better display of the week's date ranges and meeting dates. 

This was the previous state:
timeline-before

Now, the range is in the same line with the week's title, and meeting days are in the next line in day/date format.

Timeline:
three

Home page:
two

Timeline with first empty week:
three


1v4n4:
I agree it looks bad. My thought was to put a range in the same line as a Week,  and meetings I the next line, something like:

Week 3 (01/02 - 01/08)
Meetings: Thu - 01/02, Fri - 01/47

With range font smaller than Week.

Is that enough?

ragesoss:
This looks good. How about formatting it "Meetings: Monday (01/23), Wednesday (01/25)"?

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -228,52 +228,53 @@ const Timeline = createReactClass({
...
-              onBlockDrag={this._handleBlockDrag}
-              weeksBeforeTimeline={weeksBeforeTimeline}
-              trainingLibrarySlug={this.props.course.training_library_slug}
-              current_user={this.props.current_user}
-            />
-          </div>
-        )
-        );
+        )        );
+        j += 1;
```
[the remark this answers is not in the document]
1v4n4: I kind of overlooked the previous suggestion. Now I did this.props.week_meetings[i] && this.props.week_meetings[i].length === 0, it looks nice, thanks. I hope it will work

app/assets/javascripts/components/timeline/timeline_handler.jsx
```diff
@@ -92,7 +92,6 @@ const TimelineHandler = createReactClass({
       key: 'wizard_handler',
       course: this.props.course,
       weeks: this.props.weeks,
-      week_meetings: weekMeetings,
```
ragesoss, line 95: This just isn't used by any of the components that receive `courseProps`?

1v4n4: I would say no. If I'm not mistaken, it goes only to the Meetings and Wizard and there is no use of week_meetings. I tried to double-check and now I can't see the point of passing meetings to that two components either.

ragesoss:
This is ready to merge. Can you update the initial post to add screenshots from the final version?

1v4n4:
Wow, sure! Should I remove the meetings from courseProps too?

ragesoss:
I think that's better done as a separate PR.

1v4n4:
Done

Issue #4526, opened 2021-06-10
Wiki select for assigning articles lets you choose untracked wikis


ragesoss:
See https://phabricator.wikimedia.org/T284757

The tracked wikis for a course are the only ones that should show up when you are selecting the language/project via the `SelectedWikiOption` component, if it's on the context of a course.

This probably requires adding an optional prop of wiki options to `WikiSelect`, and if that prop is provided, use that instead of the full default list of `WIKI_OPTIONS` from `utils/wiki_options.js`.

PR #4660, opened 2021-11-03
Select only tracked wikis

1v4n4:
This PR should solve issue #4526 (Wiki select for assigning articles lets you choose untracked wikis). While adding articles and assigning students to articles, all Wikis list was displayed instead of tracked Wikis list.

I added a new prop to WikiSelect, and now this component can take a list of all or tracked Wikis or something else, if needed.

I needed to use some functions a few times, so I moved them from the WikiSelect component to the wiki_utils file in utils and created tests for them. 

I updated the test for the successful creation of an assignment to an article to work with tracked Wikis only.

Tracked Wikis lists (before here was all Wikis list):
tracked
tracked1
tracked2

All Wikis lists (same as before):
all
all1

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder.jsx
@@ -17,7 +17,7 @@ import { fetchCategoryResults, fetchKeywordResults, updateFields, sortArticleFin
 import { fetchAssignments, addAssignment, deleteAssignment } from '../../actions/assignment_actions.js';
 import { getFilteredArticleFinder } from '../../selectors';
 
-import { formatOption } from '../../utils/wiki_utils';
+import { trackedWikisMaker } from '../../utils/wiki_utils';
 
 const ArticleFinder = createReactClass({
   getDefaultProps() {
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -374,14 +376,25 @@ const ArticleFinder = createReactClass({
       );
     }
 
+    let trackedWikis = [];
```
ragesoss, line 379: Is there any reason for duplicating this code for generating `trackedWikis` from `course.wikis` across both components that do this, rather than extracting it to `wiki_utils`?

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -45,38 +45,6 @@
     end
   end
 
-  it 'creates a valid assignment from an article and an alternative project and language' do
```
ragesoss, line 48: Rather than removing this test, I'd prefer that it be updated to work with the new behavior by adding es.wikisource as a tracked wiki (but not the home wiki) for the test `course`.

app/assets/javascripts/components/article_finder/article_finder.jsx
1v4n4: none, I'll fix that

spec/features/multiwiki_assignment_spec.rb
1v4n4: got it!

Issue #4347, opened 2021-02-10
On P&E Dashboard, update info should show up before the first update


ragesoss:
### Current Behavior:
Once a course has gone through its first stats update, the 'Last statistics update' detail shows up beneath the stats on the Home tab, with a 'see more' modal that includes additional information about the update process and things that may go wrong with it.

However, between the start of a course and its first update, there's no indication of how long you may need to to wait before an update, and this is a common source of questions and confusion — especially for long courses that initially get sorted into the 'long_update' queue and may take several days before they are first updated.

### Desired Behavior
An update info message should show up even before the first update, and the 'see more' modal should provide info about roughly when to expect an update.

gauransh7:
I had some doubts regarding this issue - 
1. What should be the update info message that we need to show?
2. I ran the `updateCourseStats` function manually to update activities of my course, but how should we provide info about roughly when to expect an update.

ragesoss:
The update message should be something like: "The Dashboard hasn't imported any data for this program yet. It has been added to an update queue that currently takes about [X time] for a program to receive its first update."

The actual time to show there should be based on the Sidekiq queue latency for whatever queue it is sorted into initially.

See `CheckCourseJobs` for related code. Basically, you'll need to figure out the right way to use `Sidekiq::Queue` to find the current latency for a particular queue.

1v4n4:
I'd like to try to figure out this issue if no one's on it

ragesoss:
@1v4n4 great! Go for it.

PR #4679, opened 2021-11-18
Add first update info

1v4n4:
The solution for an issue 'On P&E Dashboard, update info should show up before the first update #4347'.

I added info about course updates before any of them are made. 

Before the course is enqueued for an update, a message is displayed and this is the only case with no link for statistic updates modal
no updates

After the course is enqueued for the first time, it will receive the first update flags
first_update

This info is used to calculate the first update time after the course is enqueued for update: 
up

The update messages and statistic update modal is adapted for a few different possible situations.

- The course is enqueued, but the update is not done yet.
  The update is expected:
msg
msg1

  The update is late:
msg10
msg11

- The courses enqueued for an update before implementing this fix, will not receive first update flags.

-  When the course once successfully updated, info will be displayed the same as before.

 - If the course was en enqueued for the first update before implementing this fix, but that update failed, this will be the info:
msg100
msg101

Also, the modal message about the next update is changed if that update takes longer than expected:
m

  Previously, it was like this:
m1


1v4n4:
Hi @ragesoss, this is far from over, but can you, please, check if it makes sense?

The first update is going to be in a short or medium update queue, based on this code:
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/9311fc6a69a610f64ced182e7776b8b880f48c2b/lib/data_cycle/course_queue_sorting.rb#L3-L11

https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/9311fc6a69a610f64ced182e7776b8b880f48c2b/lib/data_cycle/course_queue_sorting.rb#L37-L44

So, I made an endpoint (controller and route) for fetching data about that two queues.
On the front end, I stored data in the redux store (is that overkill?) and made a statistic component and modal able to receive that info. 
The next step is to work on update msgs.

ragesoss:
@1v4n4 nice job working your way through how all these parts of the codebase connect!

I think this approach makes sense and would work, but I think there's a much simpler one possible. Since each course is only ever going to get a first update once, I suggest storing the relevant queue latency on the `Course` record itself at the time the course is first added to a queue. We already store and update info about update times in `Course#flags`, but currently we only do that after updates. We could add a step to `ScheduleCourseUpdates` to store the following info for a course that hasn't yet gone through any updates: `enqueued_at`, `queue_name`, `queue_latency`.

Then, rather than adding new endpoints and additional fetches, it would just be a matter of updating what the two update-related React components render based on data they receive in `course.flags`.

This has the benefit, as well, of letting us say something very concrete about the update status of each course (ie, it's been added to the update queue).

The main situations to handle are:
1. A course will happen in the future. It won't be enqueued until date X (ie, `Course#start`)
2. A course has started, but has not yet been enqueued. When things are working normally, this period should only last a few minutes.
3. A course has been enqueued for an update, and the queue latency at the time it was added to the queue should give an accurate estimate for when to expect the first update.

1v4n4:
Got it!

1v4n4:
@ragesoss 
I added this first_update info in flags. I'm not using queue_name for anything, should I remove it?
first_update

I can fit the all first update info in the existing two messages without showing the modal:
Before enqueued:
beforeenqueueing
After enqueued:
afterenqueued
In that case, a link for the modal would be shown after the first update:
modallink

What do you think about this solution?

ragesoss (changes requested):
This looks like it's on the right track.

I suggest this for the 'first update' message:
* `This program has been added to the update queue. Estimated statistics update: X`

app/assets/javascripts/components/overview/statistics_update_info.jsx
```diff
@@ -19,24 +19,35 @@ const StatisticsUpdateInfo = createReactClass({
 
   getUpdateTimesInformation() {
     const course = this.props.course;
-    const updateTimesLogs = Object.values(course.flags.update_logs).filter(log => log.end_time !== undefined);
+    let lastSuccessfulUpdateMessage = '';
+            let nextUpdateMessage = '';
+    let isNextUpdateAfter = false;
+    if (course.updates.last_update === null) {
+      lastSuccessfulUpdateMessage = "The Dashboard hasn't imported any data for this program yet";
```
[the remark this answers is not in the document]
1v4n4: Sure, I just put that message to check the output

app/assets/javascripts/components/overview/statistics_update_info.jsx
```diff
@@ -19,24 +19,35 @@ const StatisticsUpdateInfo = createReactClass({
 
   getUpdateTimesInformation() {
```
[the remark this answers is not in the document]
1v4n4: Got it!

lib/data_cycle/schedule_course_updates.rb
```diff
@@ -40,6 +40,13 @@ def enqueue_course_updates
 
     courses_to_update.each do |course|
       CourseDataUpdateWorker.update_course(course_id: course.id, queue: queue_for(course))
+
+      # if course isn't updated before, add first update flags
+      flags = course.flags
+      next if flags[:first_update]
+      first_update = first_update_flags(course)
+      course.flags[:first_update] = first_update
```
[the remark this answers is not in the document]
1v4n4: Got it

app/assets/javascripts/components/overview/statistics_update_info.jsx
```diff
@@ -19,27 +19,47 @@ const StatisticsUpdateInfo = createReactClass({
...
+        lastSuccessfulUpdateMessage = `${I18n.t('metrics.last_update')}: ${lastSuccessfulUpdateMoment.fromNow()}`;
+        const nextUpdateExpectedTime = lastSuccessfulUpdateMoment.add(averageDelay, 'seconds');
+        isNextUpdateAfter = nextUpdateExpectedTime.isAfter();
+        nextUpdateMessage = `${I18n.t('metrics.next_update')}: ${nextUpdateExpectedTime.fromNow()}`;
+      }
+    }
     return [lastSuccessfulUpdateMessage, nextUpdateMessage, isNextUpdateAfter];
   },
 
+  firstUpdateTime() {
```
ragesoss, line 48: I suggest isolating these even further. Neither of these functions has any side effects in terms of React, so there's no need for them to be part of the component if they take the `course` object as their argument. I recommend either moving them to a separate file entirely (as you did previously with  wiki_utils.js) or moving them out of the component as standalone `const` functions outside of the React class (which you can see an example of in `alerts_handler.jsx`).

config/locales/en.yml
```diff
@@ -878,6 +873,9 @@ en:
     label: Course Activity
     last_update: Last statistics update
     next_update: Next estimated update
+    no_update: The Dashboard hasn't imported any data for this program yet
```
ragesoss, line 876: This is a complete sentence and should end in a period.

config/locales/en.yml
```diff
@@ -878,6 +873,9 @@ en:
...
+    enqueued_update: This program has been added to the update queue
```
ragesoss, line 877: This should end in a period.

config/locales/en.yml
```diff
@@ -878,6 +873,9 @@ en:
...
+    first_update: Estimated statistics update
```
ragesoss, line 878: In thinking this through, I realized that this language doesn't quite accurately describe the situation... this time estimate is based on the queue latency, which means it's an estimate of how long until the system *starts* processing the course for a stats update. That's in contrast to the way that it estimates update time after the first update, which is based on the average amount of time between previous updates (so, it's an estimate of when the next update will complete).

For most first updates, this is not a significant difference, but some (if they involve large numbers of edits and/or highly active users) will actually take a long time (minutes to, in extreme cases, hours) to perform the update. Maybe "Estimated start of statistics update" would reflect this better? Or if you can think of any better language, please suggest it.

config/locales/en.yml
1v4n4: Dots are here. I can remove them to yml, but I would need to change other statistic messages too.
https://github.com/1v4n4/WikiEduDashboard/blob/0606c62b796d9c7450f416a8118b33c5f46cea76/app/assets/javascripts/components/overview/statistics_update_info.jsx#L88-L92

ragesoss: Ah. Changing all of them would be good. Each complete message should include its own punctuation in the i18n file. If we rely on string substitution to combine messages with punctuation, it can result in garbled messages in languages where the English punctuation isn't relevant, or in right-to-left languages where the punctuation might end up on the wrong end of the message.

1v4n4:
Done!

app/assets/javascripts/components/overview/statistics_update_info.jsx
```diff
@@ -49,37 +55,38 @@ const StatisticsUpdateInfo = createReactClass({
...
-          getUpdateTimesInformation={this.getUpdateTimesInformation}
+          updateTimesInformation={updateTimesInformation}
           toggleModal={this.toggleModal}
         />
       );
     }
 
-    const updateTimesInformation = this.getUpdateTimesInformation();
     let updateTimesMessage = '';
     if (updateTimesInformation !== null) {
```
ragesoss, line 75: Under what circumstances could `updateTimesInformation` be `null`?

ragesoss (commented):
This looks good to me overall. I'm fine with merging it as-is, once I have a chance to test it out locally (probably tomorrow).

However, there is room to further refactor `getUpdateTimesInformation()` if you want to. I think might make sense to extract that entire function to the `utils` file. The function itself is easier to read with the `first...` and `last...` parts extracted, but it's still got quite a lot of complexity in terms of `if/else` and nesting. You might be able to find a way to reorganize and/or comment it to make it easier to understand. You might also consider returning an object with the three update-related values as properties, instead of an array, so that their names will provide clear indicators of which value is which. That would go along with refactoring StatisticsUpdateModal, since it currently expects the data as an array; you currently can't understand `updateTimesInformation` in the modal without referring to the other file to see how it's generated and which index of the array holds which information.

app/assets/javascripts/components/overview/statistics_update_info.jsx
1v4n4: I don't think it is possible. It wasn't my code and it was no problem for me, so I left it. I'll remove it and see what is happening

1v4n4:
I extracted everything I could. There is some heavy code in Utils but the component is clean.
I passed two variables to modal instead of an array, there is no use for all elements of the array in the modal.
Also, I fixed this:
modal1
This will not be shown if time is off, same as the message in the StatisticUpdateInfo component

ragesoss:
Great. I think having some info in the modal about when the update was expected, even if that time already passed, is useful. Ideally, the message would just change slightly to account for that situation, but I'm okay with leaving it like "Next estimated update: an hour ago" rather than removing it.

1v4n4:
Ok, let's go for ideally, what should I put into that message?

ragesoss:
I think something like this:

"A new update was expected: an hour ago. This may mean that the update is taking longer than expected, or may indicate a problem that is blocking updates."

ragesoss (changes requested):
Testing this out locally, I get an error when viewing a course that was already in my development environment but hasn't been scheduled for any updates yet, when I click 'See more' to activate the modal. (`Cannot read properties of null (reading 'add')`)

Looks like the utils functions need to handle clases where some of the possible relevant properties of `flags` are missing.

app/assets/javascripts/components/overview/statistics_update_info.jsx
```diff
@@ -49,33 +31,30 @@ const StatisticsUpdateInfo = createReactClass({
...
-    }
+    const updateTimesMessage = isNextUpdateAfter ? `${lastUpdateMessage} ${nextUpdateMessage} ` : `${lastUpdateMessage} `;
 
-    // Render update time information along with 'See More' link to open modal
+    // Render update time information and if some updates were made a 'See More' link to open modal
     return (
       <div className="statistics-update-info pull-right mb2">
         <small>
-          { updateTimesMessage }<a onClick={this.toggleModal} href="#">{I18n.t('metrics.update_statistics_link')}</a>
+          {updateTimesMessage} {course.updates.last_update !== null && <a onClick={this.toggleModal} href="#">{I18n.t('metrics.update_statistics_link')}</a>}
```
1v4n4, line 57: I thought I made the modal link impossible to show before the first update with this.

Anyway, I realized that modal can be useful even if there are no updates yet, at least for the late update info, so I put it back and made it work without updating data.
A link for modal is not shown before enqueued, it could work with as it is, but it will bring no extra information.

It is a bit tricky to replicate enqueued, but not updated situation and I hope I didn't miss anything.

Also, I extracted some code from modal to Utils.

lib/data_cycle/schedule_course_updates.rb
```diff
@@ -40,6 +40,12 @@ def enqueue_course_updates
 
     courses_to_update.each do |course|
       CourseDataUpdateWorker.update_course(course_id: course.id, queue: queue_for(course))
+
+      # if course isn't updated before, add first update flags
+      next if course.flags[:first_update]
```
ragesoss, line 45: Sorry, I didn't notice this before... I think this should also skip setting the :first_update flag the course already has update logs from previous updates. Otherwise, existing courses in production will have inaccurate "first update" timestamps that aren't actually from the when they were first updated.

app/assets/javascripts/components/overview/statistics_update_info.jsx
```diff
@@ -49,33 +31,30 @@ const StatisticsUpdateInfo = createReactClass({
...
-    }
+    const updateTimesMessage = isNextUpdateAfter ? `${lastUpdateMessage} ${nextUpdateMessage} ` : `${lastUpdateMessage} `;
 
-    // Render update time information along with 'See More' link to open modal
+    // Render update time information and if some updates were made a 'See More' link to open modal
     return (
       <div className="statistics-update-info pull-right mb2">
         <small>
-          { updateTimesMessage }<a onClick={this.toggleModal} href="#">{I18n.t('metrics.update_statistics_link')}</a>
+          {updateTimesMessage} {course.flags.first_update && <a onClick={this.toggleModal} href="#">{I18n.t('metrics.update_statistics_link')}</a>}
```
ragesoss, line 57: With this guard statement, the 'see more' link does not show up for any courses already in the system before the 'first_update' flag feature gets deployed. Instead, the modal (or getUpdateTimesArray) should be updated to be able to handle any shape of `course` object.

1v4n4: oh, sure. Wasn't thinking about consequences

ragesoss:
I'm still getting an error when testing this out locally. In particular, I have a course in my local environment where the `flags` looks like this:

```
flags: {
    academic_system: 'semester',
    format: 'In-person',
    timeline_enabled: true,
    wiki_edits_enabled: true,
    online_volunteers_enabled: false,
    edit_settings: {
      wiki_course_page_enabled: true,
      assignment_edits_enabled: false,
      enrollment_edits_enabled: true
    },
    peer_review_count: 2,
    stay_in_sandbox: false,
    update_logs: {
      '1': {
        orphan_lock_failure: true
      }
    },
    average_update_delay: null
```

So it has `update_logs`, but no successful updates.

I think this is something that the util should handle.

Issue #4680, opened 2021-11-18
Silent javascript error when a course search returns no results


ragesoss:
### What is happening?

If you go to /explore and do a course search that returns no results, there is a JS error like this:

```
templater.js:37 Uncaught TypeError: Cannot read properties of undefined (reading 'childNodes')
    at n.getItemSource (templater.js:37)
    at new n (templater.js:6)
    at t.exports (templater.js:173)
    at Object.start (index.js:53)
    at new t.exports (index.js:261)
    at HTMLDocument.<anonymous> (course.js:21)
    at e (jquery.min.js:2)
    at t (jquery.min.js:2)
    at r (helpers.js:72)
```

It looks like this is because the jQuery code in `course.js` doesn't account for an empty table of results.

It has no visible effects, it just generates error logs that get sent to Sentry.

The fix should probably be done by adjusting the JavaScript (rather than tweaking the results HTML to meet the expectations of the JS).

raffifu:
The problem happen because there are two section with `id='course'`
- 1st Section: show search result
- 2nd Section: show campaign course

It will throw an error when one of these element not contain class 'list' used in`list.js` dependencies. in this case, happen when return no results from search

The Best Practice is to change the `id` value from search section. Because as i know `id` params must unique (only one for one html page). then `course.js` must updated to recognize new `id` value

The second option (easy way) is to edit `_search.html.haml` and then delete `#course` if none result from search

I can work on this issue, but it will take a little time to setup the environment (because i'm new here)

ragesoss:
@raffifu go for it. If you run into any problems setting up the environment, give me a ping. (If you like, you can also send me your email address and I can invite you to the project Slack.)

raffifu:
Thank you for inviting me, this is my public email `ah2n85yos@relay.firefox.com` (If you can't invite me with these email, i will send my actual email). 

Anyway, i'm facing problem when setup dev env with docker. I can't open `/explore` because always redirect to `/`. I've copied `application.example.yml` and `database.example.yml`
![image](https://user-images.githubusercontent.com/22138274/143686562-2a082f3f-c16c-4d14-b718-82e1de5db50b.png)
This is the log from console [pastebin](https://pastebin.com/DRzLdyEN)

ragesoss:
You need to create a default campaign, a Campaign record with a slug that
matches the default_campaign value from application.yml.

ragesoss:
[...]

PR #4682, opened 2021-11-28
Silent javascript error when a course search returns no results

raffifu:

## What this PR does
This commit solve #4680 

To fix silent error, I change `id value from courses to course_results` in `_search.html.haml` And add logic for new id in `course.js`. So after the PR applied there is will be different `id value` in one html page.

- The code in `course.js` is redundant because section `course` and `course_results` absolutely doing same logic (see line 21 and 34)

### DIFF
--- a/app/assets/javascripts/utils/course.js
+++ b/app/assets/javascripts/utils/course.js
@@ -27,6 +27,19 @@ $(() => {
     });
   }
 
+  // Course Results sorting
+  // only sort if there are tables to sort
+  let courseResultList;
+  if ($('#course_results table').length) {
+    courseResultList = new List('course_results', {
+      page: 500,
+      valueNames: [
+        'title', 'school', 'revisions', 'characters', 'references', 'average-words', 'views',
+        'reviewed', 'students', 'creation-date', 'ungreeted', 'untrained'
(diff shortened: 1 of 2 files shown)


ragesoss:
This looks good to me. I'll test it out tomorrow.

Your open question is a good one. I'm okay with the duplication, in this case, because it's very easy to understand, and in the longer run, we're more likely to replace the jquery + server-side HTML with React than we are to keep duplicating it further. Removing jquery as a dependency is a long-term goal.

raffifu:
Thank you for your response. I'm interested with the idea to replace jQuery with React. is there any project board for this goal? I have experience using react and want to contribute

ragesoss:
@raffifu there's no project board for it, but I can create an issue.

I should clarify that replacing the HTML table views (which use jquery and the list.js pacakge, in the courses.js util file, to provide table sorting functionality) with React is partially independent of removing the jquery dependency.

We use jquery's ajax for many of the frontend data fetching requests from React components, and possibly we have a few other uses of jquery as well. (It might also be possible to retain the server-rendered HTML table views while finding a lighter-weight way to implement table sorting without using jquery.)

Issue #4681, opened 2021-11-27
Improve input matching in WikiSelect component


zblace:
### Current Behavior:
There is no way to set some of Wikipedia instances as main (home) ones.

### Desired Behavior:
Option to select HR, SH and BS Wikipedia if not ALL?

### Related Files & Code:
I have no code overview or skill toknow how this is managed.

### Additional context:
https://outreachdashboard.wmflabs.org/

ragesoss:
You can set any of these as the home wiki. Each course/event can only have one home wiki; additional wikis should be added to the Tracked Wikis field.

For `hr` and `sh`, you may need to type out most of the domain before it shows up as an option in the dropdown, because other language codes that come earlier include those letters (ksh, chr). That can be confusing. An improvement would be to do the filtering based on matching the domain from the beginning, instead of matching anywhere the input is found, so that typing `sh.wiki` would not return matches for `ksh.wikipedia.org` and so on.

PR #4688, opened 2021-11-30
Improve input matching in WikiSelect

1v4n4:
Solution for issue  'Improve input matching in WikiSelect component #4681'

For optimizing results in the WikiSelect component I added two types of matching input.

If the input is less than three characters, it will be matched from the beginning of the string. That would enable easy access to some Wikis 
match1
match

If the input is more than three characters, it will be matched anywhere in the string. That way results like Wikidata wouldn't be missed.
match2a
match1b

### DIFF
--- a/app/assets/javascripts/components/common/wiki_select.jsx
+++ b/app/assets/javascripts/components/common/wiki_select.jsx
@@ -106,8 +106,8 @@ const WikiSelect = createReactClass({
     };
 
     const options = this.props.options;
-    // Two types of matching for optimizing results.
-    // If the input is less than four characters, it will be matched from the beginning of the string.
+
+    // If the input is less than three characters, it will be matched from the beginning of the string.
 
     const filterOptionsLong = function (val) {
       return options.filter(wiki =>
@@ -115,16 +115,16 @@ const WikiSelect = createReactClass({
       ).slice(0, 10); // limit the options for better performance
     };
 
-  // If the input is less than four characters, it will be matched anywhere in the string
+  // If the input is at least three characters, it will be matched anywhere in the string
 
     const filterOptionsShort = function (val) {
       return options.filter(wiki =>
         wiki.label.toLowerCase().startsWith(val.toLowerCase())
-      ).slice(0, 10); // limit the options for better performance
+      ).slice(0, 10);
     };
 
     const loadOptions = function (inputValue, callback) {
-      if (inputValue.length < 4) {
+      if (inputValue.length < 3) {
         callback(filterOptionsShort(inputValue));
       }
       callback(filterOptionsLong(inputValue));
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/common/wiki_select.jsx
```diff
@@ -107,7 +107,7 @@ const WikiSelect = createReactClass({
     const options = this.props.options;
     const filterOptions = function (val) {
       return options.filter(wiki =>
-        wiki.label.toLowerCase().includes(val.toLowerCase())
+        wiki.label.toLowerCase().startsWith(val.toLowerCase())
```
ragesoss, line 110: One wrinkle I didn't consider previously: how will this work when you want to select "wikidata" as the wiki?

app/assets/javascripts/components/common/wiki_select.jsx
```diff
@@ -104,15 +104,30 @@ const WikiSelect = createReactClass({
...
+
+    const filterOptionsShort = function (val) {
+      return options.filter(wiki =>
+        wiki.label.toLowerCase().startsWith(val.toLowerCase())
+      ).slice(0, 10); // limit the options for better performance
+    };
+
     const loadOptions = function (inputValue, callback) {
-      callback(filterOptions(inputValue));
+      if (inputValue.length < 4) {
```
ragesoss, line 127: I like this strategy! Since the project domains all start with "wiki", this should accurately filter to an appropriate set of wikis in almost all cases. The one thing I could think of is that, for wikidata (and three other unique projects), we put them at the top of the list currently, so that when you type `w`you see `www.wikidata`, `www.wikisource`, `commons`, and `incubator` at the top. However, `w` is a valid language code, so we don't want to special-case that.

But since the problem we're solving for is language-code collisions where a two letter code is part of a three letter code (but not the first part), I think we could safely move the "short" cutoff to `< 3`.

app/assets/javascripts/components/common/wiki_select.jsx
```diff
@@ -104,15 +104,30 @@ const WikiSelect = createReactClass({
...
+    // Two types of matching for optimizing results.
+    // If the input is less than four characters, it will be matched from the beginning of the string.
+
+    const filterOptionsLong = function (val) {
       return options.filter(wiki =>
         wiki.label.toLowerCase().includes(val.toLowerCase())
       ).slice(0, 10); // limit the options for better performance
     };
 
+  // If the input is less than four characters, it will be matched anywhere in the string
```
ragesoss, line 118: This comment should be "at least four" (or three).

app/assets/javascripts/components/common/wiki_select.jsx
1v4n4: :D

1v4n4:
I changed it for less than three. As you said, it solves the problem. And I wouldn't like for users who are looking for Wikidata or Wikisource to give up before the fourth letter. 
I'll check the other PR tomorrow (my tomorrow).

Issue #4677, opened 2021-11-16
Add Alerts for "de-userfying" edits


ragesoss:
When new users move a sandbox into mainspace on Wikipedia, the edits are automatically tagged `de-userfying`. Here's a recent changes feed of such edits on English Wikipedia: https://en.wikipedia.org/w/index.php?hidebots=1&hidecategorization=1&hideWikibase=1&tagfilter=de-userfying&limit=500&days=30&title=Special:RecentChanges&urlversion=2

The Wiki Education version of the Dashboard should monitor recent de-userfying edits, match the usernames to users in current Courses, and generate Alerts that send email to Wiki Expert (aka content expert) for the course.

### Related Files & Code:

This is similar to other feeds that we monitor and generate alerts for via `UpdateCycleAlertGenerator`, and should follow the pattern the other alert manager or monitor files.

The outline for completing this is something like:

* build an equivalent API query that returns a feed of edits tagged `de-userfying`.
* turn that query into a call to `WikiApi`
* build a monitor class that uses that query to generate a set of recent de-userfying edits (including the username, article, and revision id for each one)
* write an ActiveRecord query to get all the usernames of students in current Courses.
* filter the edits list to find any edits made by the Users in current Courses.
* for each of the matched edits, create a new Alert and send the email for that Alert, unless an Alert for that user/article/course combination already exists.
* write tests that verify the monitor works as expected.

cyrillefr:
Hello @ragesoss, can i work on this one ?

ragesoss:
@cyrillefr sure! let me know if you have any questions.

cyrillefr:
Hello @ragesoss , I am almost done.
Though, I do not know how to link the article title I get from the api call with a model/table.
In my dev environment, I have assigned a student to a wikipedia article, hoping to populate the courses_users.assigned_article_title field to get a match, but it did not change.
I have a new line in Assignment and Article though.
Which model/join should I use to match the article title ?

ragesoss:
Hmm... without seeing your code, I'm not sure exactly what you're trying to do. If you open a work-in-progress PR, I can take a look.

I don't think `Assignment` will need to be part this feature. `Article#title` is a relevant attribute for matching titles you get from the API, although I think in some cases, the Dashboard won't yet have an `Article` record that corresponds to the newly de-userfied page.

cyrillefr:
Hello @ragesoss ,
I just opened a [WIP] PR.
[...]

PR #4691, opened 2021-12-06
Add Alerts for "de-userfying" edits

cyrillefr:
## What this PR does
References issue #4677 (Add Alerts for "de-userfying")
This PR to create and handle a new "de-userfying" alert

By: 
- adding a new AlertMonitor object
- adding a new sub type of Alert
- adding the matching specs
- adding to the list of alerts
- adding to the list of mailer previews

### DIFF
--- a/lib/alerts/de_userfying_edit_alert_monitor.rb
+++ b/lib/alerts/de_userfying_edit_alert_monitor.rb
@@ -1,0 +1,72 @@
+# frozen_string_literal: true
+
+# This class identifies articles that have been moved from
+# their use space by users
+# Tag: de-userfying
+# Cf. https://en.wikipedia.org/wiki/MediaWiki:Tag-de-userfying
+class DeUserfyingEditAlertMonitor
+  def self.create_alerts_for_deuserfying_edits
+    new.create_alerts
+  end
+
+  def initialize
+    @wiki = Wiki.find_by(language: 'en', project: 'wikipedia')
+  end
+
+  def create_alerts
+    student_edits = edits_made_by_students(edits, current_students)
+    student_edits.each do |edit|
+      article = Article.find_by(mw_page_id: edit['pageid'])
+      user = User.find_by(username: edit['user'])
+      course_id = ArticlesCourses
+                  .where(article_id: article.id)
+                  .joins(course: [:users])
+                  .find_by(users: { id: user.id })
+                  .course_id
+      create_alert(user.id, course_id, article.id, edit['revid'])
+    end
+  end
+
+  def edits
+    api = WikiApi.new @wiki
+    query_params = {
+      list: 'recentchanges',
+      rcprop: 'title|ids|flags|user|tags|loginfo',
+      rctag: 'de-userfying',
+      rcshow: '!bot'
+    }
+    res = api.query(query_params)
+    res.data['recentchanges'].map { |change| change.slice('user', 'revid', 'pageid', 'logparams') }
+  end
+
+  # Those enrolled in at least one course. Multiple enrolled are counted only once.
+  def current_students
+    CoursesUsers
+      .select(:user_id, 'users.username')
+      .joins(:user)
+      .where(role: CoursesUsers::Roles::STUDENT_ROLE)
+      .distinct(:user_id)
+  end
+
+  def edits_made_by_students(edits, students)
+    usernames = students.map(&:username)
+    edits.filter { |edit| usernames.include?(edit['user']) }
+  end
+
+  def create_alert(user_id, course_id, article_id, revision_id)
+    return if alert_already_exists?(course_id, article_id, revision_id)
+    alert = Alert.create!(type: 'DeUserfyingAlert',
+                          user_id: user_id,
+                          course_id: course_id,
+                          article_id: article_id,
+                          revision_id: revision_id)
+    alert.email_content_expert
+  end
+
+  def alert_already_exists?(course_id, article_id, revision_id)
+    Alert.exists?(type: 'DeUserfyingAlert',
+                  course_id: course_id,
+                  article_id: article_id,
+                  revision_id: revision_id)
+  end
+end
--- a/lib/alerts/de_userfying_edit_alert_monitor.rb
+++ b/lib/alerts/de_userfying_edit_alert_monitor.rb
@@ -16,7 +16,7 @@ class DeUserfyingEditAlertMonitor
   def create_alerts
     student_edits = edits_made_by_students(edits, current_students)
     student_edits.each do |edit|
-      article = Article.find_by(mw_page_id: edit['pageid'])
+      article = article_by_mw_page_id(edit['pageid'])
       user = User.find_by(username: edit['user'])
       course_ids = courses_for_a_student(user.id)
       course_ids.each do |course_id|
@@ -25,6 +25,12 @@ class DeUserfyingEditAlertMonitor
     end
   end
 
+  # Query:
+  # https://en.wikipedia.org/w/api.php
+  # ?action=query&list=recentchanges&rcprop=title|ids|flags|user|tags|loginfo&rctag=de-userfying
+  # Will fetch last 10 edits (parameter rclimit default) on an approximatively
+  # 12 hours timespan.
+  # Cf. https://www.mediawiki.org/wiki/API:RecentChanges
   def edits
     api = WikiApi.new @wiki
     query_params = {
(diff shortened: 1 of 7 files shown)


ragesoss:
I'll try to test this out in more detail tomorrow. It looks like it should work. There are multiple possible ways of finding which course(s) are relevant for a given de-userfying edit.

I think the approach I would recommend would be to avoid going through ArticlesCourses at all, and just create an Alert record for every current course where that user is a student. That would avoid problems with the ArticlesCourses query coming up with no matching courses (which might happen if an ArticlesCourses record doesn't exist yet because this alert is being generated before the most recent edits and Article records get imported). It should still set an article_id, so it might need to import the Article if it doesn't already exist in the database.

ragesoss:
Looks like there's also a failing test related to this PR.

spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      edits = mntor.edits
+      students = mntor.current_students
+      expect(mntor.edits_made_by_students(edits, students).count).to eq editsfeed.count
+    end
+  end
+
+  describe '.create_alerts' do
+    before do
+      populate_content_expert
+      populate_articles_courses
```
ragesoss, line 64: What happens if neither the Articles nor ArticlesCourses records exist when the monitor runs?

lib/alerts/de_userfying_edit_alert_monitor.rb
```diff
@@ -0,0 +1,78 @@
...
+      course_ids = courses_for_a_student(user.id)
+      course_ids.each do |course_id|
+        create_alert(user.id, course_id, article.id, edit['revid'])
+      end
+    end
+  end
+
+  def edits
+    api = WikiApi.new @wiki
+    query_params = {
```
ragesoss, line 30: It would be helpful to include comments to characterize this query, including
* How many edits over roughly how much time result from this query (ie, it looks like there are something like ~20 de-userfying edits on English Wikipedia per day)
* The equivalent URL for doing the same query manually in a browser

spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require "#{Rails.root}/lib/alerts/de_userfying_edit_alert_monitor"
+
+def mock_mailer
+  OpenStruct.new(deliver_now: true)
+end
+
+describe DeUserfyingEditAlertMonitor do
```
ragesoss, line 10: In addition to this spec, it would be very useful to add a preview for this alert (in `alert_mailer_preview.rb`) so that we can easily see what an email for this alert type would look like.

cyrillefr: Done. There was also a mistake in the class name.

lib/alerts/de_userfying_edit_alert_monitor.rb
cyrillefr: Added these comments. Hope they are ok.

spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb
cyrillefr: Oops, I forgot to follow your advice not to use ArticlesCourses from the last conversation. (Actually, did look at it locally but then forgot :) )
I also add the case where article from the wiki not present in the dashboard DB.

lib/alerts/de_userfying_edit_alert_monitor.rb
ragesoss: Great. I suggest adding `limit: 100` to the parameters, as there could definitely be spikes in de-userfying edits where more than 10 happen in a short timeframe, so we that might miss some between ConstantUpdate runs.

spec/mailers/previews/alert_mailer_preview.rb
```diff
@@ -49,4 +57,9 @@ def example_alert(type: 'HighQualityArticleEditAlert')
     Alert.new(type: type, article: example_article,
               course: example_course, id: 9)
   end
+
+  def example_de_userfying_alert
+    Alert.new(type: 'DeUserfyingAlert', article: example_article,
+              course: example_course, id: 9, user: example_student)
```
ragesoss, line 63: This didn't work for me, because the `example_course` doesn't get created but the alert object includes a user... so I get an error when trying to construct the contribution URL for the user. I fixed it locally by using `Course.last` in place of `example_course` here.

lib/alerts/de_userfying_edit_alert_monitor.rb
```diff
@@ -0,0 +1,92 @@
...
+      .where(role: student, user_id: id)
+      .pluck(:course_id)
+  end
+
+  def article_by_mw_page_id(mw_page_id)
+    wiki_id = @wiki.id
+    unless Article.exists?(wiki_id: wiki_id, mw_page_id: mw_page_id)
+      ArticleImporter.new(@wiki).import_articles([mw_page_id])
+    end
+    Article.find_by(wiki_id: wiki_id, mw_page_id: mw_page_id, deleted: false)
```
ragesoss, line 90: Does this need to be `deleted: false`?

There could be cases where the article has already been deleted by the time the alert is generated (but it already got imported as an Article record during a course update process before it was deleted).

ragesoss:
Good work so far! This is close to being ready.

I tested it out and was able to a) add someone who recently made a de-userfying edit as a student, b) run the monitor, and c) generate an Alert as expected. Yay!

It errors in cases where there is no Article record, and the article has already been deleted when the monitor runs. This will be a relevant use case, as sometimes a student's article will get deleted because they moved it to mainspace inappropriately and someone noticed quickly and flagged it for removal. You can replicate that situation by finding one of the recent logs where the target article is a redlink and adding that user to a course in your development environment: https://en.wikipedia.org/w/index.php?hidebots=1&hidecategorization=1&hideWikibase=1&tagfilter=de-userfying&limit=500&days=30&title=Special:RecentChanges&urlversion=2

I think it would be useful to store several additional bits of data as a hash in the `details` column of the alert: the logid (which can be used to construct a link to the move log, eg https://en.wikipedia.org/w/index.php?title=Special:Log&logid=124896595), and the timestamp (which will require adding `timestamp` to the query params). That would make it easy to build a nicer custom email template for this alert type, which I think we'll want to do as a follow-on after this is live.

spec/mailers/previews/alert_mailer_preview.rb
cyrillefr: I cannot replicate that, is there something for me to do ?

ragesoss: Are you able to visit `/rails/mailers/alert_mailer/de_userfying_alert` without hitting an error?

A quick fix will just be to replace `example_course` with `Course.last` here, but I'm not sure about why it would work for you and not for me.

ragesoss:
`ArticleImporter.new(@wiki).import_articles([mw_page_id])` won't error, but if that page doesn't exist or has been deleted (which means it is still present in the Wikipedia database, but not viewable), then an `Article` record won't be created.

I think in a case like this, we still want to create an Alert, just with `article_id: nil`. In the `create_alert` call, `article&.id` would probably make this work.

lib/alerts/de_userfying_edit_alert_monitor.rb
cyrillefr: Added the 100 limit

spec/mailers/previews/alert_mailer_preview.rb
cyrillefr: Yes, there is no error showing up when accessing /rails/mailers/alert_mailer/de_userfying_alert
A bit strange.

lib/alerts/de_userfying_edit_alert_monitor.rb
cyrillefr: I copied the snippet from existing code with deleted: false. I removed it.

cyrillefr:
I updated the create_alert method to handle this case.

But ... I just realise I should have updated:
```ruby
def url
  article.url
end
```
to
```ruby
def url
  article&.url
end
```
in the `app/models/alert_types/de_userfying_alert`
I should add it in the next round if there is one.

ragesoss:
Cool. Please make that change to the `url` method. Everything else looks good to me at this point. I'll test it out again locally once you've added the `url` change.

ragesoss:
Nice work, thank you!

PR #4695, opened 2021-12-11
Add item terminology

1v4n4:
Items terminology added to the course/items tabs.

I created utils functions for returning items or articles messages, based on the course's home Wiki. I added Items related messages to en.yml.


1v4n4:
Idea is to shift messages between articles and items, with simple function

1v4n4:
The idea is to shift messages with a simple function. 
It would look like this (I didn't change everything yet, just marked stuff):

s1

The question is, if you agree with this approach, can I rearrange en.yml and put all article-related stuff together?

config/locales/en.yml
```diff
@@ -218,6 +218,83 @@ en:
       itself. (These scores are only available for English Wikipedia articles.)
     articles_short: Articles
 
+  items:
+    item_not_found: "This item has not been created yet."
```
ragesoss, line 222: Are all these used, or is this just the fill list of messages that you found that currently use "article" terminology?

Using `item_not_found` (and so on) instead of sticking with `article_not_found` could introduce some additional complexity. I recommend staying with the same keys for each message except for using the `items` message tree instead of `articles`, so that it would either be `articles.article_not_found` or `items.article_not_found`.

config/locales/en.yml
```diff
@@ -218,6 +218,83 @@ en:
...
+    item_development: (item development)
+    item_shown: "Showing %{count} of %{total}"
+    assigned: Assigned Items
+    available: Available Items
+    no_available: No Available Items
+    next: Next
```
ragesoss, line 228: It would be preferable not to duplicate messages that don't have any differences between the "Item" version and "Article" version. That's the pattern we've used for the `courses` vs. `courses_generic` message trees.

app/assets/javascripts/utils/articles_or_items_utils.js
```diff
@@ -0,0 +1,5 @@
+const getArticlesOrItems = (project) => {
```
ragesoss, line 1: Consider using the pattern from `CourseUtils.i18n`: you could have a function to which you pass in the message key along with the project. This would also let you more easily solve the problem of a fallback when the new `items` version isn't translated or perhaps doesn't exist; you can use the `default` option to fall back to the `articles` version when needed. (For users, that would mean seeing the "articles" version of a message in their own language rather than seeing the "items" version in English, if there's no translation yet available.)

1v4n4:
@ragesoss 
Can I use the same message for both of the courses?

courses:
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/f2e9930484cb9ad208e2a5e79f81e9b9c45c9df8/config/locales/en.yml#L394

courses_generic:
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/f2e9930484cb9ad208e2a5e79f81e9b9c45c9df8/config/locales/en.yml#L753

ragesoss:
Hmm... maybe we can find a way to avoid having both 'courses' and 'articles' in the same message here. "There are no edited articles yet." and "There are no edited items yet." would work.

1v4n4:
So far, I changed




Everything, except this messages:
4n

The next will be 
![5next](https://user-images.githubusercontent.com/65791349/146428857-b489ff87-648b-4c0d-8709-f8f3d7ba5782.png

app/assets/javascripts/utils/articles_or_items_utils.js
```diff
@@ -0,0 +1,14 @@
+
+export default class ArticlesOrItemsUtils {
+// Returns article or item, based on home Wiki.
+  static articlesOrItems(project) {
+    return project === 'wikidata' ? 'items' : 'articles';
+  }
+
+// Returns article or item message.
+  static articlesOrItemsI18n(messageKey, articlesOrItems, defaultarticlesOrItems = 'articles') {
```
ragesoss, line 9: I suggest having this take the `project` as a parameter instead of `articlesOrItems`, and use the articlesOrItems function internally, so that the components that use it only need one function rather than both.

I suggest calling the class `ArticleUtils` and the function just `i18n`, to make it a little more concise in all the places that will use it. I think `ArticleUtils.i18n` communicates well enough the purpose of providing appropriate terminology for pages of different types, in parallel to `CourseUtils.i18n`.

config/locales/en.yml
```diff
@@ -218,6 +221,34 @@ en:
...
+      Mainspace pages edited between the start and end dates are listed below.
+      The dashboard pulls in contributions from the home wiki of this program, along
+      with any other wikis for which there is an Assignment or an Available Item.
+    does_not_exist: The Item doesn't exist. Create a Sandbox page and assign yourself to that to receive suggestions.
+    edited: Items Edited
+    edited_none: There are no edited items yet.
+    rating_doc: The rating an item has been assigned by WikiProjects on Wikipedia. These ratings may corresponds to much older versions, so they are just a rough indicator of item maturity.
+    tracked_doc: Only tracked items contribute to the the statistics for this program. When items are marked as untracked, they will be removed from the statistics during the next data update.
+    view_doc: The estimated views to each item since it was first edited, through the most recent stats update. You can use pageviews.toolforge.org to see the exact numbers.
+    rating_docs:
```
ragesoss, line 239: These ratings are specific to Wikipedia, so I don't think any alternative message strings are needed.

All Wikidata items at this point (I think) will be unrated, but even better than translating the 'unrated' message will be (later) to not render any rating icon at all for Wikidata items.

app/assets/javascripts/components/assignments/assignment.jsx
```diff
@@ -53,6 +53,7 @@ const Assignment = createReactClass({
     if (reviewers.length) { reviewers.pop(); }
 
     let feedback;
+    console.log('in assignment', this.props.assignmentGroup[0]);
```
ragesoss, line 56: This should be removed.

ragesoss:
This looks good... I think we can merge it after the changes I suggested.

config/locales/en.yml
```diff
@@ -910,6 +938,9 @@ en:
     recent_updates_summary: Out of the last %{total} updates, %{failure_count} failed and %{error_count} encountered errors.
     revisions: Recent Edits
     references_count: References Added
+    references_doc_separated:
```
ragesoss, line 941: I missed these `separated` message keys earlier.

I think it's better to avoid renaming (or duplicating) existing messages, which will cause i18n regressions for other languages unless/until translators add the translations for the new keys.

I suggest, instead, adding the new message key like `references_doc_wikidata`, then use the `isWikidata` logic or similar to append to the key name if appropriate.

PR #4704, opened 2021-12-18
Remove ratings for Items Edited, Assigned and Available tabs

1v4n4:
Ratings are removed for Items, for desktop and mobile versions.

2m

I'm not sure what I did in the edited tab here, because I don't know how to add edited items, but it should be fine.

Articles are the same as before.
4a

### DIFF
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -44,7 +44,6 @@ const Article = createReactClass({
   render() {
     const ratingClass = `rating ${this.props.article.rating}`;
     const ratingMobileClass = `${ratingClass} tablet-only`;
-    const isArticle = project !== 'wikidata';
 
     // Uses Course Utils Helper
     const formattedTitle = CourseUtils.formattedArticleTitle(this.props.article, this.props.course.home_wiki, this.props.wikidataLabel);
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/articles/article.jsx
```diff
@@ -44,6 +44,7 @@ const Article = createReactClass({
   render() {
     const ratingClass = `rating ${this.props.article.rating}`;
     const ratingMobileClass = `${ratingClass} tablet-only`;
+    const isArticle = project !== 'wikidata';
```
ragesoss, line 47: Where does `project` come from?

app/assets/javascripts/components/articles/available_article.jsx
```diff
@@ -58,6 +58,7 @@ export const AvailableArticle = createReactClass({
     const ratingClass = `rating ${assignment.article_rating}`;
     const ratingMobileClass = `${ratingClass} tablet-only`;
     const articleLink = <a onClick={this.stop} href={article.url} target="_blank" className="inline">{article.formatted_title}</a>;
+    const isArticle = this.props.course.home_wiki.project !== 'wikidata';
```
ragesoss, line 61: Similarly to the Revision component, this can be done on a per-article basis with `assignment.project`, and should probably be done with `isWikipedia` rather than `isArticle`.

app/assets/javascripts/components/assignments/assignment.jsx
```diff
@@ -24,6 +24,8 @@ const Assignment = createReactClass({
     wikidataLabel: PropTypes.string
   },
   render() {
+    const isArticle = this.props.course.home_wiki.project !== 'wikidata';
```
ragesoss, line 27: Here also, get the project from the `article` (after `const article = ...` is defined).

app/assets/javascripts/components/articles/article.jsx
1v4n4: From the article (props).
https://github.com/1v4n4/WikiEduDashboard/blob/1b6b98351d3f2c06e753d3b402b320006837eb00/app/assets/javascripts/components/articles/article.jsx#L72-L74
That one is ok, right? I'll change all variables to isWikipedia.

ragesoss:
Looks to me, and works as expected locally.

PR #4705, opened 2021-12-18
Add Items terminology to Add Available Items components

1v4n4:
I added Items terminology to AddAvailableArticles and ArticleFinder components.
For the Wikidata course it looks like this:

If the wiki is changed to Wikipedia, terminology in this component will switch to Articles.
placeholder

For the other courses, it is the same as before.

except wiki is changed to Wikidata
switch1

Article Finder has Items terminology:
af

If Wiki in ArticleFinder is changed, whole terminology changes:
af1

If Wiki in AddAvailableWiki is changed, ArticleFinder will render appropriate articles or items version.


app/assets/javascripts/components/articles/add_available_articles.jsx
```diff
@@ -64,6 +65,7 @@ const AddAvailableArticles = createReactClass({
   },
 
   render() {
+    const articlesOrItems = ArticleUtils.articlesOrItems(this.props.course.home_wiki.project);
```
ragesoss, line 68: This should use the project from `props` rather than the home_wiki, so that it automatically changes terminology to match the selected wiki.

app/assets/javascripts/components/articles/add_available_articles.jsx
```diff
@@ -64,17 +65,26 @@ const AddAvailableArticles = createReactClass({
   },
 
   render() {
+    const isWikidata = this.props.project === 'wikidata';
     return (
       <div className="pop__padded-content">
-        <TextAreaInput
+        {isWikidata && <TextAreaInput
```
ragesoss, line 71: Is there a simpler way to do this, such as defining the placeholder before the `return`?

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
```diff
@@ -194,21 +193,24 @@ const EditButton = ({
   );
 };
 
-const FindArticles = ({ course, open }) => {
+const FindArticles = ({ course, open, project, language }) => {
+  let btnText = 'Search Wikipedia for an Article';
```
ragesoss, line 197: Ideally this should be moved to i18n.

app/assets/javascripts/components/articles/add_available_articles.jsx
1v4n4: I tried this:
placeholder

No good. Like a few other things, I could think of. This ugly stuff was the only thing that works for endless switching from wiki to data and reverse

ragesoss: Hmm... I'm not sure why that wouldn't work. The component obviously re-renders when the project changes, or else neither implementation would work.

I'll play around with it when I have a chance. This seems like it might be related to rendering shortcuts with TextAreaInput.

1v4n4: With previous code (and all others) everything is changing nicely, but the placeholder. It is switching only from wiki to data and only one time.

ragesoss: The simpler way works if you also change the `id` to correspond to the current state of `isWikidata`. That forces React to use two different DOM objects (and component instances) for the two versions of this `TextAreaInput` component, so it works around whatever is preventing it from re-rendering when the `placeholder` value changes.

I think the reason it was not re-rendering is that the TextAreaInput component is using the `InputHOC` higher-order-component, and that component implements a custom `shouldComponentUpdate` implementation that limits re-renders to `state` (not `prop`) changes in order to keep the input field performance fast enough that you don't wait for a character to render when you type it.

app/assets/javascripts/components/article_finder/article_finder.jsx
```diff
@@ -454,4 +459,7 @@ const mapDispatchToProps = {
   clearResults: clearResults,
 };
 
-export default connect(mapStateToProps, mapDispatchToProps)(ArticleFinder);
+export default compose(
```
ragesoss, line 462: What is this change for?

app/assets/javascripts/components/articles/add_available_articles.jsx
ragesoss: (Here's what I did for the id: `const inputId = isWikidata ? 'add_available_items' : 'add_available_articles';`)

app/assets/javascripts/components/article_finder/article_finder.jsx
1v4n4: I added withRouter to access location, because I'm passing project if changed in AssignButton for ArticleFinder to render right version.

https://github.com/1v4n4/WikiEduDashboard/blob/6f593ae734a857ae97fd94c60a4fcf7c4af26019/app/assets/javascripts/components/common/AssignCell/AssignButton.jsx#L202

ragesoss: I don't understand where `location` comes into it. It looks like the project comes through a prop.

1v4n4: In Article finder project is determined based on home_wiki. 
My logic was to pass it through link_to from AssignButton, so I could work with something other than a home wiki. To access that prop in Article Finder with this.props.location.project I needed withRouter. Is this making any sense?

ragesoss: Oh... is that because the prop is being passed through the LinkTo element?

1v4n4: right

1v4n4: Anyway. I just found another error to fix tomorro
w
w

app/assets/javascripts/components/articles/add_available_articles.jsx
1v4n4: Works great with the id. Thanks for the explanation, there is a lot things I didn't know

1v4n4:
I'm not sure what to do with the falling test (spec/lib/commons_spec.rb). It is ok locally, so I can't tell what is wrong with it. I also, don't know why this is happening (different test results). 
test

ragesoss:
That test failure is definitely unrelated to your change.

The failing test checks whether there are at least two pages of API results for pages that use a certain file on Wikimedia Commons. It uses the VCR library to save and re-use the actual request results, so if you ran this test before locally, it will still be using the results from the earlier request rather than hitting the Commons API again every time you run the test.

There are a few possible causes. It could be a temporary network problem which caused those API queries to fail while that particular test was running on travis-ci, or it could be that the API query being tested no longer returns results that make the test pass (either because there are now fewer than 500 global usages of the tested file, or because the API changed in some other way that makes it work differently). Given the fact that the same test failed on two builds in a row, it's probably not a network problem, which means we can probably replicate the failure locally by deleting the `fixtures` directory or the specific file corresponding to this test. (If you've never done so, it's interesting to look at the contents of one of these VCR files before you delete it.)

I can replicate this test failure locally, but I'm still not sure what exactly changed. But we can merge this PR anyway, since it's unrelated. Is there anything else you're changing, or is it ready?

1v4n4:
It's ready

PR #4706, opened 2021-12-18
Add item terminology to article finder WIP

1v4n4:
Items added to Article finder and ArticleFinderRow.

 
Articles are handled as usual.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -250,19 +250,27 @@ en:
     already_exists: This assignment already exists!
     title_too_large: The title shouldn't exceed 255 characters
     assignees: Assignee(s)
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -250,19 +250,27 @@ en:
     already_exists: This assignment already exists!
     title_too_large: The title shouldn't exceed 255 characters
     assignees: Assignee(s)
-    assign_other: Assign/remove an article
+    assign_other:
```
ragesoss, line 253: As in the other PR, let's keep these existing message keys as they are and continue using them for non-wikidata, and just add a new ones for wikidata terminology.

PR #4707, opened 2021-12-19
Remove ratings for items on Recent Activity tab and add item terminology

1v4n4:
Ratings for items are removed from the Recent Activity tab.

Articles are the same as before.

### DIFF
--- a/app/assets/javascripts/components/revisions/revision.jsx
+++ b/app/assets/javascripts/components/revisions/revision.jsx
@@ -9,20 +9,20 @@ const Revision = ({ revision, index, wikidataLabel, course, setSelectedIndex, la
   const ratingMobileClass = `${ratingClass} tablet-only`;
   const formattedTitle = CourseUtils.formattedArticleTitle({ title: revision.title, project: revision.wiki.project, language: revision.wiki.language }, course.home_wiki, wikidataLabel);
   const subtitle = wikidataLabel ? `(${CourseUtils.removeNamespace(revision.title)})` : '';
-  const isArticle = course.home_wiki.project !== 'wikidata';
+  const isWikipedia = revision.wiki.project === 'wikipedia';
 
   return (
     <tr className="revision">
       <td className="tooltip-trigger desktop-only-tc">
-        {isArticle && <p className="rating_num hidden">{revision.rating_num}</p>}
-        {isArticle && <div className={ratingClass}><p>{revision.pretty_rating || '-'}</p></div>}
-        {isArticle && <div className="tooltip dark">
+        {isWikipedia && <p className="rating_num hidden">{revision.rating_num}</p>}
+        {isWikipedia && <div className={ratingClass}><p>{revision.pretty_rating || '-'}</p></div>}
+        {isWikipedia && <div className="tooltip dark">
           <p>{I18n.t(`articles.rating_docs.${revision.rating || '?'}`, { class: revision.rating || '' })}</p>
           {/* eslint-disable-next-line */}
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/revisions/revision.jsx
```diff
@@ -9,17 +9,20 @@ const Revision = ({ revision, index, wikidataLabel, course, setSelectedIndex, la
   const ratingMobileClass = `${ratingClass} tablet-only`;
   const formattedTitle = CourseUtils.formattedArticleTitle({ title: revision.title, project: revision.wiki.project, language: revision.wiki.language }, course.home_wiki, wikidataLabel);
   const subtitle = wikidataLabel ? `(${CourseUtils.removeNamespace(revision.title)})` : '';
+  const isArticle = course.home_wiki.project !== 'wikidata';
```
ragesoss, line 12: I don't think switching this behavior on the course home wiki is the best option here. The revisions in the Activity view may include ones from wikis other than the home wiki; a course that tracks both Wikipedia and Wikidata could have a mix of edits from either project.

Since these ratings are specific to Wikipedia (and are also irrelevant for other non-Wikipedia wikis in addition to Wikidata), I think checking whether the revision is for Wikipedia is probably the best option (ie, `isWikipedia = revision.wiki.project == 'wikipedia`).

app/assets/javascripts/components/revisions/revision_list.jsx
```diff
@@ -62,7 +63,7 @@ const RevisionList = createReactClass({
       references: {
         label: I18n.t('revisions.references'),
         desktop_only: true,
-        info_key: 'metrics.references_doc'
+        info_key: `metrics.references_doc_separated.${ArticleUtils.articlesOrItems(this.props.course.home_wiki.project)}`
```
ragesoss, line 66: Per my comment on the earlier PR, I think this should be updated so that it continues to use `metrics.references_doc` for articles, and renames the items version something like `references_doc_wikidata`.

ragesoss (changes requested):
This works nicely. I can merge once the i18n key issue is fixed.

app/assets/javascripts/utils/article_utils.js
```diff
@@ -1,19 +1,23 @@
...
-  const articlesOrItems = ArticleUtils.articlesOrItems(project);
-  return ArticleUtils.articlesOrItemsMsg(messageKey, articlesOrItems);
-}
+  // Returns article or item message, based on home Wiki project.
+  static I18n(messageKey, project) {
+    const articlesOrItems = ArticleUtils.articlesOrItems(project);
+    return ArticleUtils.articlesOrItemsMsg(messageKey, articlesOrItems);
+  }
+
+  static chooseMsg(project, msg) {
```
ragesoss, line 20: Everything in this PR looks good to me now, except for the name of this function. Can you think of a name that more clearly communicates its intended use of possibly appending a project-specific suffix to a message key? (I think that, while the current implementation only knows about Wikidata, we could imagine the behavior changing later to add a different suffix for a different project.)

1v4n4: appendProjectSufix? But for Wikipedia nothing is appending...

ragesoss: Maybe just `projectSuffix`? That gets across the idea of varying the string based on the project, without implying that one will always be added.

1v4n4: Done!

Issue #4716, opened 2021-12-29
ArticleViewer tries to get data from wikis where WikiWho data isn't available


ragesoss:
For example, a French Wikipedia article listed on the Dashboard still tries to fetch WikiWho data for authorship highlighting, even though French isn't one of the languages that WikiWho supports. (Hopefully, we'll add support for more languages at some point, but meanwhile the Dashboard should be aware of which languages do have support and which don't, and reflect this in the UI.)

Florence-Njeri:
Hi @ragesoss , can I pick this task?

ragesoss:
@Florence-Njeri yes, go for it.

ragesoss:
Hi @Florence-Njeri! Are you still working on this? Let us know if you have any questions.

marklocklear:
Spent some time trying to understand this issue. I found `fetchParsedArticle()` in `app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx` and also `const supported = ['de', 'en', 'es', 'eu', 'tr'];` in `isWhocolorLang()` in this same file. Is this in the ballpark @ragesoss ?

ragesoss:
@marklocklear yes, it sounds like you're on the right track.

cyrillefr:
Hello @ragesoss , I have been playing a bit with this issue.

When displaying an article in French, I cannot see any sign of bug in the browser console (obviously no lines appear highlighted).
Furthermore, when looking in the network, I see a call to /en/whocolor/... for an article in English, but no such call for the one in french (which seems good to me).
What am i missing here? 
(Should articles that cannot be highlighted be excluded ?)

ragesoss:
hmm... I think I was mistaken, and this bug doesn't exist. I remember filing this bug, and I think I had noticed some kind of behavior like it describes, but there's code that was already there when I filed it that works to prevent making such useless requests. Sorry about that!

PR #4717, opened 2021-12-30
Add items terminology to editors tab

1v4n4:
If course's home Wiki is Wikidata, Items version of Editors Tabs is rendered.

1a
2a
4a

Before the same tab had only Articles terminology:

MyArticles feature is adapted for Wikidata courses, as MyItems.

Before, there was no that option.
5a

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -225,6 +225,7 @@ en:
 
   items:
     items: Items
+    article_not_found: "This item has not been created yet."
     assigned: Assigned Items
     assignments_none: There are no assigned items yet.
     available: Available Items
@@ -893,23 +894,29 @@ en:
   instructor_view:
     article_assignments: "%{prefix} Assignments"
     assignments_table:
-      article_name: Article Name
+      article_name: Name
       current_stage: Current Stage
       relevant_links: Relevant Links
     assigned_articles: Assigned Articles
+    assigned_articles_wikidata: Assigned Articles
     bad_work:
       learn_more: "If student work made the article worse, we encourage you to edit it yourself. Otherwise, let us know about the problems so that we can fix them later on."
+      learn_more_wikidata: "If student work made the item worse, we encourage you to edit it yourself. Otherwise, let us know about the problems so that we can fix them later on."
       learn_more_button: "Learn More"
       submit_issue: "What are the significant problems with this article?"
+      submit_issue_wikidata: "What are the significant problems with this item?"
       submit_issue_placeholder: "The more expert perspective and detail you can provide, the better."
       thank_you: "Thank you! Contact your Wiki Expert if you have additional questions."
     exercises_and_trainings: "%{prefix} Exercises & Trainings"
     no_assignments: "This student currently has no assigned articles. You can assign an article by using the buttons above."
+    no_assignments_wikidata: "This student currently has no assigned itemss. You can assign an item by using the buttons above."
     overview: "%{prefix} Overview"
     reviewing_articles: Reviewing Articles
+    reviewing_articles: Reviewing Items
     select_student:
       title: "Select %{prefix}"
       content: to assign and view article assignments
+      content_wikidata: to assign and view article assignments
 
   metrics:
     label: Course Activity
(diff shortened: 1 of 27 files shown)


config/locales/en.yml
```diff
@@ -1200,7 +1225,7 @@ en:
       increased page size. This number doesn't necessarily reflect the
       amount of new content the user added. Restoring removed content
       will also increase this total.
-    chars_added: Characters Added<br>(article | user | draft)
+    chars_added: Characters Added<br>(item | user | draft)
```
ragesoss, line 1228: It looks like this changes the default version rather than add a new variation for Wikidata.

config/locales/en.yml
```diff
@@ -1232,6 +1257,7 @@ en:
     no_articles_wikidata: No items
     no_revisions: This editor has made no revisions.
     no_highlighting: "No text from the current version of the article is attributed to this editor. Check the page history to see their edits."
+    no_highlighting_wikidata: "No text from the current version of the item is attributed to this editor. Check the page history to see their edits."
```
ragesoss, line 1260: This feature does not work for Wikidata anyway — see https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/4716 — so I don't think a variant message is necessary.

config/locales/en.yml
1v4n4: I thought that  ' characters added' is only for Wikidata, but it is for all non-Wikipedias, right? I'll change that.

ragesoss: Characters added can be for any project. For English Wikipedia specifically, we use 'words added' instead in some places (by converting a byte/character count to an estimated word count), but the data comes from basically just the difference in byte size between two revisions of a wiki page.

app/controllers/user_profiles_controller.rb
```diff
@@ -59,6 +60,12 @@ def delete_profile_image
     redirect_to controller: 'user_profiles', action: 'show', username: @user.username
   end
 
+  def max_project
```
ragesoss, line 63: This is interesting! I hadn't thought about doing anything with the user profile pages. I would prefer this feature — switching terminology on user profile pages — as a separate PR.

app/controllers/user_profiles_controller.rb
```diff
@@ -59,6 +60,12 @@ def delete_profile_image
...
+    projects_array = public_courses.map { |course| Wiki.find_by(id: course.home_wiki_id).project }
```
ragesoss, line 64: This is an N+1 query; it does a SELECT on the wikis table for every course. You can rewrite this so that all the wikis are loaded at once before the `map`.

ragesoss:
If you separate out the part that uses `max_project` and fix the linting error that is causing the build to fail, this will probably be okay to merge.

It's easier for me to handle large sets up independent updates like this if most of the commits have their own PR. Ideally, anything that can stand on its own as a ready-to-go improvement gets a PR, that one change request in one part of it doesn't hold up shipping another change. (Since these are already combined and most of them are doing similar things, it's fine to leave them together this time.)

app/controllers/user_profiles_controller.rb
1v4n4: Will do. 
And I have a question here. In case that there are more wikis with max occurrences, Wikipedia will be prioritized and articles version rendered.
If there is no Wikipedia, the first in the max array will render.
What do you think to check for Wikidata after Wikipedia (make Wikidata second priority) and render Items version if, for example, max array has Wikidata and other Wikis, but no Wikipedia?

ragesoss: That sounds reasonable, although perhaps not a high enough priority to worry about. I think this `max projects` approach should only be a temporary one. What this user profile page really wants is to show Wikidata stats separately from Wikipedia stats, rather than showing combined stats from multiple projects with just the labels changing based on which project is the most used.

1v4n4: if that's the case, no need for that

ragesoss:
This looks good. I'll test it out manually tomorrow and likely merge it. Could you include before/after screenshots?

1v4n4:
Sure.

Issue #4693, opened 2021-12-08
Generate an Alert when a WikiProject Medicine article is assigned for a course that doesn't have the medical content training module


ragesoss:
Most existing articles that relate to medicine — and which fall under the strict sourcing rules covered in [this training module](https://dashboard.wikiedu.org/training/students/editing-medical-topics) — have the WikiProject Medicine banner template on their talk page. In `wiki_education` mode only, we should generate an alert whenever one of these articles is assigned — but only if the course does not have the medical topics training module assigned.

The point is to catch cases where a class that wasn't expected to work on medical topics tries to do so, so that we can point them to the relevant training module or intervene some other way.

## Related Files & Code:

This is similar to other feeds that we monitor and generate alerts for via `UpdateCycleAlertGenerator`, and should follow the pattern the other alert manager or monitor files. `HighQualityArticleMonitor` is probably the most relevant one, as it does something very similar.

cyrillefr:
Hello @ragesoss, may I work on this one ?

ragesoss:
@cyrillefr yes! I was hoping you might take this one on.

cyrillefr:
Hi @ragesoss , I cannot find the way to assign a training module to a course.(wiki_education mode on and off, my perm is 1(admin)).

ragesoss:
To add a training module:

* Go to the Timeline tab of a course. If it doesn't have a Timeline tab, enable the timeline via Edit Details on the Home tab.
* Add a Week to the Timeline, then add a Block, and from the Block, select a training module. Then save it.
* If there are no training modules available to select, you need to load them. Make sure `wiki_education: 'true'` is set in `application.yml`, then visit `/reload_trainings?module=all`.

Alternatively, if you are already in `wiki_education` mode, when you create a new course, you'll be taken through the assignment wizard. Choose the 'research and write articles' path, and at the end it will generate a timeline automatically that includes many training modules (possibly including the medical topics module, depending on your choices in the assignment wizard).

cyrillefr:
Thank you. Regarding training modules, which I had not, I did TrainingBase.reload_all in the console. In both mode, guess I have them all by now(that was before your post). But I have the one about medical articles editing.

PR #4720, opened 2021-12-31
Generate an Alert when a WP Med article is assigned without proper training

cyrillefr:
## What this PR does

Generate an Alert when a WikiProject medicine article is assigned
for a course that doesn't have the medical content training module.

Issue: #4693 

How to find such WP med article ? Their talk page belongs to the
Category:All_WikiProject_Medicine_article category.
We query wp for that. After grabing the last assignments, we check
for med articles without the training.
For each case -> an alert is raised.

- add a new sub type of alert
- add to log in constant update
- add a lib monitor file
- the matching spec
- the matching preview
- add a factory to create a training module
- fix two unexpected spec bugs(training modules created but non deleted in a spec)
  that cause the temp creation of a training module to fail because of
  duplicate slug name in the med alert spec.


cyrillefr:
Some extra points I did:
While testing, I have found that 2 models(TrainingModule and TrainingLibrary) where not reseted between examples.
Which lead to the fail of the spec I have created, where I need to create a med training module.

This is because of the before(:all), which is practical because you do not need to run before each example.
Downside is it does not behave like before(:each), wich is cleaned after example is run.

So I added 2 fixes to clean after the example suite complete by adding 2 `after(:context)` in:
`spec/features/training_tool_spec.rb` and `spec/models/course_spec.rb`

Also: I have a few rubocop issues that will be corrected with asked corrections if any.
But, I have one Metrics/AbcSize issue with file `lib/data_cycle/update_cycle_alert_generator.rb`.
In the giant method `def generate_alerts`.
Could I add an exclusion of that file in the .rubocop.yml file?

lib/alerts/medicine_article_monitor.rb
```diff
@@ -0,0 +1,90 @@
...
+                  course_id: assignment.course_id)
+  end
+
+  def last_assigned_med_articles
+    last_assigned_articles
+      .select { |assignment| med_article?(assignment.article_title) }
+  end
+
+  def last_assigned_articles
+    Assignment.last NUMBER_OF_ARTICLES_TO_CHECK
```
ragesoss, line 51: I don't think looking at a fixed number of recent assignments is the right strategy here. There will be times when a large number of assignments are added between two consecutive `constant update` cycles.

We want to start generating alerts for new assignments, without adding a slew of useless alerts for courses from the past. Finding recent assignments based on creation date (eg, assignments created within the last day) is probably a good approach.

lib/alerts/medicine_article_monitor.rb
```diff
@@ -0,0 +1,90 @@
...
+  end
+
+  def med_training_for_course?(course_id)
```
ragesoss, line 54: This can be done a little bit more efficiently and concisely. The medical topics training actually has two variants for different kinds of courses, but these modules have hard-coded IDs (11, 48).

You can make a constant for the array of medical training IDs, and check whether either of them are found among the `training_module_ids` for a course (ie, `(Course.find(course_id).training_module_ids & MEDICAL_TRAINING_IDS).any?` or similar... that can probably be refactored a bit for clarity).

lib/alerts/medicine_article_monitor.rb
```diff
@@ -0,0 +1,90 @@
...
+  # No need to normalize article title
+  # "Talk:Colonic polypectomy" eq Talk:Colonic_polypectomy
+  def med_article?(title)
+    api = WikiApi.new @wiki
+    query_params = {
+      action: 'query',
+      prop: 'categories',
+      clcategories: WIKI_PROJECT_MEDICINE_CAT,
+      titles: "Talk:#{title}",
+      cllimit: 50,
```
ragesoss, line 79: Hmm... this approach will do a separate API query for every title being checked.

There are a few alternatives that I suspect could be faster.

You can add multiple titles to the same query, doing a search for all of the recent article titles at once. There might be cases where we search for many medical articles at once, though, so the limit of 50 might come into play. (It can probably be raised to 500, which would make it unlikely to be a problem, although still maybe possible.)

Alternatively, you could use the `Category` model to store and occasionally update the entire list of WikiProject Medicine articles. That would look something like `Category.find_or_create_by(source: 'category', name: 'All_WikiProject_Medicine_articles', wiki: <English Wikipedia>)`. This category could be updated (via `refresh_titles`) if it hasn't been updated within the last day. (It takes about 30 seconds to refresh a category that large, so we probably don't want to do it every update cycle.)

I think the `Category` approach is probably the most appropriate here.

spec/lib/alerts/medicine_article_monitor_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+end
+
+describe MedicineArticleMonitor do
+  let(:mntor) { described_class.new }
+  let(:course1) { create(:course, slug: 'slug-one') }
+  let(:course2) { create(:course, slug: 'slug-two') }
+  let(:student1) { create(:user, username: 'alice', email: 'student1@example.edu') }
+  let(:med_article) { create(:article, title: 'my med title 2', mw_page_id: 222) }
+  let(:content_expert) { create(:user, greeter: true) }
+  let(:med_training_module) { create(:training_module, slug: 'editing-medical-topics') }
```
ragesoss, line 17: I think it's better here to use the same `TrainingModule.load_all` method from the other specs for setting up the training module content for the tests rather than using a factory — especially per my comments above, since we should be specifying the relevant modules via their static IDs. That will also make it so that no changes to other tests are necessary.

lib/alerts/medicine_article_monitor.rb
cyrillefr: From reading from `config/schedule.yml`, I read that ConstantUpdateWorker is executed each 4 minutes.
This is why I have chosen a fixed number. 

If I take daily created assignements, there will be a lot of trying to create alerts.
Though I check that an alert does not exists before creating it.
I will do that: the last 24 hours.
But, to be sure, the cron run each 4 minutes ?

ragesoss: Yes, the cron runs every 4 minutes at most. It won't schedule a new one if the previous one is still running, though (because of the `lock: :until_executed` option for that worker, which is part of the `sidekiq_unique_jobs` plugin).

I think a shorter time than 24 hours would be okay, but I'm guessing that making it that long won't cause any problems and will have only a minor impact on execution time.

lib/alerts/medicine_article_monitor.rb
cyrillefr: I was a bit proud of that method, and this query :)
But, I see your point, I tried the `Category` thing.
The problem is; how to update the category ? We cannot put it in the constant_update.
Can it have its own worker in schedule.yml ? Or some other place ?

ragesoss: I think it can be done from this file, with something like `category.refresh_titles unless category.updated_at > 1.day.ago`. Updating that category is something that only this monitor cares about, so I think it would be unnecessarily complex to schedule that update separately.

cyrillefr: I have changed to your Category solution, with this refreshing strategy.

spec/lib/alerts/medicine_article_monitor_spec.rb
cyrillefr: I have reverted to that.

lib/alerts/medicine_article_monitor.rb
```diff
@@ -0,0 +1,89 @@
...
+  def last_refresh_too_late?
+    @med_category.updated_at < @assignment_time_for_refresh_titles
+  end
+
+  def no_article_in_med_category?
+    @med_category.article_titles.empty?
+  end
+
+  def med_category
+    prms = { name: WIKI_PROJECT_MEDICINE_CAT, source: 'category', wiki_id: 1 }
```
ragesoss, line 86: We can't assume that en.wiki has ID 1. Here's a strategy for finding it: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/lib/importers/rating_importer.rb#L72

lib/data_cycle/update_cycle_alert_generator.rb
```diff
@@ -43,6 +44,17 @@ def generate_alerts
 
     log_message 'Generate de-userfying edits alerts'
     DeUserfyingEditAlertMonitor.create_alerts_for_deuserfying_edits
+
+    generate_alert_for_med_articles
+  end
+
+  def generate_alert_for_med_articles
+    if Features.wiki_ed?
```
ragesoss, line 52: For consistency with most of the codebase, I'd prefer this guard as `return unless Features.wiki_ed?` rather than using an `if` block.

lib/alerts/medicine_article_monitor.rb
```diff
@@ -0,0 +1,89 @@
...
+
+  def initialize
+    @assignment_time_for_alert = 1.day.ago
+    @assignment_time_for_refresh_titles = 1.day.ago
+    @med_category = med_category
+  end
+
+  def create_alerts_for_no_med_training_for_course
+    last_assigned_med_articles.each do |assignment|
+      unless med_training_for_course?(assignment.course_id)
```
ragesoss, line 28: One inefficiency here is that it will run the `med_training_for_course?` query once per Assignment, even if all the assignments are for the same course.

It think this is okay; it's probably not a significant bottleneck, and the code is clear and isolated to this file. But if we wanted to make this a little faster, we could include the `Course` in the `last_assigned_articles` query, and make `med_training_for_course` into an instance method of `Course` (with its value cached in an instance variable the first time it gets checked). (Other strategies to work around this N+1 situation without changing `Course` are also possible.)

ragesoss: I'm not requesting any changes to this (just thinking it through in writing).

ragesoss:
I'll test it out locally, probably early next week. Code-wise, it looks good aside from the two small changes I noted.

ragesoss:
Nice work, thank you!

PR #4724, opened 2022-01-03
Add Items terminology to user profile

1v4n4:
Users profile tab is rendering based on max home wiki occurrences. 

If Wikidata is most frequent, Items terminology is displayed: 
profile

In other cases, it is Articles terminology as before:
profile1


1v4n4:
I tried to improve this method by searching only for max wiki(s).
def

ragesoss:
That looks reasonable. I think a slightly better way — more idiomatic Rails — would be to use `.includes(:home_wiki)` before the `.map`.

ragesoss:
This PR should only include the changes for the user profile page. You may need to cut a new branch starting from `master` and cherry-pick the relevant commits.

PR #4732, opened 2022-01-12
Add stats table WIP

1v4n4:
Added Statistic table

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -77,6 +77,7 @@ class Course < ApplicationRecord
            class_name: 'TicketDispenser::Ticket',
            foreign_key: 'project_id',
            dependent: :destroy
+  has_one :course_stat, class_name: 'CourseStat', dependent: :destroy
 
   #########################
   # Activity by the users #
--- a/app/services/update_wikidata_stats.rb
+++ b/app/services/update_wikidata_stats.rb
@@ -32,10 +32,9 @@ class UpdateWikidataStats
   def update_wikidata_stats
     return if course_revisions.empty?
     stats = WikidataSummaryParser.analyze_revisions(course_revisions)
-    course_stat = CourseStat.find_by(course_id: @course.id)
-                  || CourseStat.create(course_id: @course.id)
-    course_stat.stats_hash = stats
-    course_stat.save
+    crs_stat = CourseStat.find_by(course_id: @course.id) || CourseStat.create(course_id: @course.id)
+    crs_stat.stats_hash = stats
+    crs_stat.save
   end
 
   def wikidata_revisions
--- a/app/services/update_wikidata_stats.rb
+++ b/app/services/update_wikidata_stats.rb
@@ -5,15 +5,15 @@ require_dependency "#{Rails.root}/lib/wikidata_summary_parser"
 class UpdateWikidataStats
   def initialize(course)
     @course = course
-    update_summaries
-    update_wikidata_stats
+    import_summaries
+    update_wikidata_statistics
   end
 
   private
 
-  def update_summaries
-    return if wikidata_revisions.empty?
-    wikidata_revisions.find_in_batches do |rev_batch|
+  def import_summaries
+    return if wikidata_revisions_without_summaries.empty?
+    wikidata_revisions_without_summaries.find_in_batches do |rev_batch|
       rev_batch.each do |rev|
         summary = WikidataSummaryParser.fetch_summary(rev)
         next if summary.nil?
@@ -25,11 +25,12 @@ class UpdateWikidataStats
         end
       end
     end
+
   end
 
   # Get Wikidata stats based on revision's summaries
 
-  def update_wikidata_stats
+  def update_wikidata_statistics
     return if course_revisions.empty?
     stats = WikidataSummaryParser.analyze_revisions(course_revisions)
     crs_stat = CourseStat.find_by(course_id: @course.id) || CourseStat.create(course_id: @course.id)
@@ -37,8 +38,8 @@ class UpdateWikidataStats
     crs_stat.save
   end
 
-  def wikidata_revisions
-    @course.revisions.where(wiki: wikidata, summary: nil, deleted: false)
+  def wikidata_revisions_without_summaries
+    course_revisions.where(summary: nil)
   end
 
   def course_revisions
@@ -46,6 +47,6 @@ class UpdateWikidataStats
   end
 
   def wikidata
-    Wiki.get_or_create(language: nil, project: 'wikidata')
+    @wikidata ||= Wiki.get_or_create(language: nil, project: 'wikidata')
   end
 end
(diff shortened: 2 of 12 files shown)


app/models/course.rb
```diff
@@ -222,6 +224,7 @@ def self.ready_for_survey(opts)
   before_save :check_course_times
   before_save :set_needs_update
   after_create :ensure_home_wiki_in_courses_wikis
+  after_create :create_statistic
```
ragesoss, line 227: I don't think this is necessary. You'll need to account for already-existing courses that won't have a statistics record, anyway.

app/models/statistic.rb
```diff
@@ -0,0 +1,4 @@
+class Statistic < ApplicationRecord
```
ragesoss, line 1: I suggest calling this `CourseStatistics` or `CourseStats`, since it's specifically designed to hold stats for a Course.

ragesoss (commented):
I think you're on the right track!

app/models/statistic.rb
ragesoss: Also, it should be in `app/models/course_data/`.

db/migrate/20220112180923_change.rb
```diff
@@ -0,0 +1,5 @@
+class Change < ActiveRecord::Migration[6.1]
+  def change
+    rename_table :statistics, :course_stats
```
ragesoss, line 3: I'd rather have just one migration for this PR.

app/services/update_wikidata_stats.rb
```diff
@@ -0,0 +1,51 @@
...
+class UpdateWikidataStats
+  def initialize(course)
+    @course = course
+    update_summaries
+    update_wikidata_stats
+  end
+
+  private
+
+  def update_summaries
```
ragesoss, line 14: This method imports missing summaries rather than update any already-present ones. I suggest calling it `import_summaries`.

app/services/update_wikidata_stats.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def update_wikidata_stats
+    return if course_revisions.empty?
+    stats = WikidataSummaryParser.analyze_revisions(course_revisions)
+    crs_stat = CourseStat.find_by(course_id: @course.id) || CourseStat.create(course_id: @course.id)
+    crs_stat.stats_hash = stats
+    crs_stat.save
+  end
+
+  def wikidata_revisions
```
ragesoss, line 40: This name does not convey that it's only revisions without summaries. Maybe `wikidata_revisions_without_summaries`?

app/services/update_wikidata_stats.rb
```diff
@@ -0,0 +1,51 @@
...
+    @course.revisions.where(wiki: wikidata, summary: nil, deleted: false)
```
ragesoss, line 41: This could be done as`course_revisions.where(summary: nil)` to remove a little bit of repetition.

app/services/update_wikidata_stats.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def course_revisions
+    @course.revisions.where(wiki: wikidata, deleted: false)
+  end
+
+  def wikidata
+    Wiki.get_or_create(language: nil, project: 'wikidata')
```
ragesoss, line 49: You can cache this so that it only needs one database query no matter how many times `wikidata` gets used: `@wikidata ||= Wiki.get_or_create ...`

lib/data_cycle/schedule_course_updates.rb
```diff
@@ -3,17 +3,20 @@
...
   include BatchUpdateLogging
   include CourseQueueSorting
 
   def initialize
+    @updated_courses = []
     setup_logger
     return if updates_paused?
 
     run_update
+    wikidata_stats_updates
```
ragesoss, line 19: This strategy does not solve the problem of getting the wikidata updates in sync with the other course data updates. It will add all the CoursesDataUpdateWorker jobs to the sidekiq queue, and then add all the UpdateWikidataStatsWorker jobs to the queue, but then either all the first set of jobs will need to complete before the first wikidata job gets processed, or if there are multiple processes taking jobs from the queue, you could still end up with cases where the wikidata job runs before the other job for that course has completed.

I suggest, instead, that we start by including the wikidata update as *part of* the CourseDataUpdate for each course. You can do this while still keeping the worker framework by calling `UpdateWikidataStatsWorker.new.perform(course_id)`. Calling `perform` directly means it won't be enqueued, it will just execute that job inline. (This way, we can also take advantage of the logging framework and see the effects of this change in the course update logs in terms of how much time a given course takes to do this new stats processing.)

ragesoss:
This is looking great so far.

A good step at this point would be to write a test that runs a course data update and generates some wikidata stats for that course (ie, a course with a user who made a few wikidata edits during the course timeframe). That will be very useful in refactoring.

1v4n4:
@ragesoss I added some basic tests, not sure if that is enough, can you check that, please?

ragesoss:
Yes, those tests look like a great foundation to me.

lib/data_cycle/schedule_course_updates.rb
```diff
@@ -41,21 +43,32 @@ def enqueue_course_updates
     courses_to_update.each do |course|
       CourseDataUpdateWorker.update_course(course_id: course.id, queue: queue_for(course))
 
+      UpdateWikidataStatsWorker.new.perform(course) if wikidata(course)
```
ragesoss, line 46: Do this in UpdateCourseStats, rather than here. `ScheduleCourseUpdates` is designed to just put all the update jobs for current courses into the right queue, rather than do any of the actual data updates.

Doing it this way would mean that this process would first add a course CourseDataUpdateWorker job to the queue, then (before that job runs) it would do a `UpdateWikidataStatsWorker` job without going through the queue, and then it would move on to the next course to put one job in the queue and do the other in the current context.

Putting it into UpdateCourseStats means the existing update logging scheme can capture details about how long the wikidata part of it took for each course, and that time will contribute to the total average time of updates for that course (and so, it will affect which queue it should get sorted into).

app/controllers/courses_controller.rb
```diff
@@ -127,6 +127,11 @@ def alerts
     @alerts = current_user&.admin? ? @course.alerts : @course.public_alerts
   end
 
+  def course_stats
```
ragesoss, line 130: I think it's going to be easier to include these new stats as part of `show.json`, rather than having a separate route and separate request for them.

Also, I think we should finish and merge and deploy the database and update cycle changes (with or without sending the stats to the frontend in `show.json`) before adding any frontend changes to the PR.

What's left from your perspective before this would be ready to merge?

1v4n4: I reset the last commit and I can make that show.json fast. And the other stuff... Maybe to fix CVS to be made from table

ragesoss:
Okay! I think this is ready. I'm going to try to deploy it today.

PR #4733, opened 2022-01-12
Add random items terminology WIP

1v4n4:
Items terminology added to InstructorMessage, CourseStats, StatisticUpdateModal, Details

image(2)
image(1)

Screenshot from 2022-01-24 21-46-16
Screenshot from 2022-01-24 21-45-47

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -702,9 +702,13 @@ en:
       other: "%{count} requested accounts pending."
     requested_accounts_alert_view: "View"
     retain_available_articles: Retain Available Articles
+    retain_available_articles_wikidata: Retain Available Items
     retain_available_articles_info: >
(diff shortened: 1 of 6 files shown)


config/locales/en.yml
```diff
@@ -952,6 +953,7 @@ en:
     articles_edited_generic: Pages Edited
     articles_edited_generic_wikidata: Items Edited
     article_scoped_program_info: This is an 'Article Scoped Program', which only counts edits to assigned articles and categories.
+    article_scoped_program_info_wikidata: This is an 'Items Scoped Program', which only counts edits to assigned Items and categories.
```
ragesoss, line 956: This is the name of the program type. It's not a great name and might change, but we shouldn't make it dynamic based on wikidata vs wikipedia.

Issue #4650, opened 2021-10-25
SurveyResponseAlert emails should indicate which survey the alert is from


ragesoss:

cyrillefr:
Hi @ragesoss , Can I work on this one ?

ragesoss:
@cyrillefr sure. This is probably a pretty small issue.

ragesoss:
Yes.

cyrillefr:
Hello @ragesoss, I am facing difficulties completing this task.

I have looked at the DB to see how I could handle this issue. I have watched the associations between models, and may be I am missing something, but I cannot think of a way to get a relationship between an answer and a survey.

So I did a little test:
I went on /admin, then /surveys
- created 2 question groups (with 3 questions each)
- created 2 surveys
- assigned to each survey both question groups (no complains)

Then I went to http://localhost:3000/surveys/1/ to complete the survey #1:
every questions from both question groups = 6 answers to 6 questions

I have checked at http://localhost:3000/surveys/results
Each survey is 0% answered (which seems a bit strange btw)

I can download a CSV of survey #1.
BUT, I can also download survey #2, that is the same result as survey #1 (even though I did not attempt to take survey #2).

I think this is due to the models:
In the `surveys_question_groups` table(which is a join table), you can link a survey and question group. A question group may have surveys, and the other way around.

When you take(answer) a survey, you populate `rapidfire_answer_groups` and `rapidfire_answers` tables.

There is no direct link to a survey, but a link to question group. Which may have a collection of surveys. That means, there is no direct association between an answer and a survey, only between an answer and a question group. And a n-n relationship between question groups and surveys.

ragesoss:
Oh... interesting. I didn't realize that an answer can't be traced back unambiguously to the survey it came from. The survey code is quite a mess, both in terms of database design and UI.

Still, in practice, we don't re-use question groups between surveys; we clone them and use the new question group for a new survey. So I think you could work around this problem by checking whether the question group is only part of only one survey, and if so, infer that it's the right one.

Alternatively, the email could just put the name of the question group, which would also provide the necessary context in most cases.

PR #4749, opened 2022-01-26
Add Survey/Question group to survey response alert

cyrillefr:
Add survey(s)/question group to the SurveyResponseAlert email 
Issue #4650

- update survey response manager to include alert source
- add alert mailer to alerts list

### DIFF
--- a/lib/alerts/survey_response_alert_manager.rb
+++ b/lib/alerts/survey_response_alert_manager.rb
@@ -9,19 +9,31 @@ class SurveyResponseAlertManager
     @answers.each do |answer|
       next unless answer_meets_alert_conditions?(answer)
       next if alert_already_exists?(answer)
-      details = {
-        question: answer.question.question_text,
-        answer: answer.answer_text,
-        followup: answer.follow_up_answer_text
-      }
       alert = Alert.create(type: 'SurveyResponseAlert',
                            message: alert_message(answer),
                            user_id: answer.user.id,
                            subject_id: answer.question.id,
-                           details: details,
+                           details: details(answer),
                            target_user_id: alert_recipient&.id)
       alert.email_target_user
     end
+  end
+
+  def details(answer)
+    {
+      question: answer.question.question_text,
+      answer: answer.answer_text,
+      followup: answer.follow_up_answer_text,
+      source: self.class.source(answer)
+    }
+  end
+
+  def self.source(answer)
+    question_group = answer.answer_group.question_group
+    survey_names = question_group.surveys.map(&:name).join(' .')
+
+    format('Question group: %<qg>s, Survey(s): %<s>s ',
+           { qg: question_group.name, s: survey_names })
   end
 
   private
--- a/spec/mailers/previews/alert_mailer_preview.rb
+++ b/spec/mailers/previews/alert_mailer_preview.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "#{Rails.root}/lib/alerts/survey_response_alert_manager"
+
 #= Preview all emails at http://localhost:3000/rails/mailers/alert_mailer
 class AlertMailerPreview < ActionMailer::Preview
   def articles_for_deletion_alert
@@ -34,6 +36,10 @@ class AlertMailerPreview < ActionMailer::Preview
     AlertMailer.alert(example_no_med_training_for_course, example_user)
   end
 
+  def survey_response_alert
+    AlertMailer.alert(example_survey_response_alert, example_user)
+  end
+
   private
 
   def example_user
@@ -41,7 +47,7 @@ class AlertMailerPreview < ActionMailer::Preview
   end
 
   def example_student
-    User.new(email: 'nospam@nospam.com', username: 'nospam', permissions: 0)
+    User.new(email: 'nospam@nospam.com', username: 'Me_student', permissions: 0)
   end
 
   def example_course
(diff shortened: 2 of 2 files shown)


cyrillefr:
Hi @ragesoss , I had to create a few records to get the preview working. But I cannot clean the DB after (it is not test mode).
I find it a bit annoying, if you have an idea to work around it, i am open to all ideas :)

ragesoss:
I'll play around with it. I don't like all that code in the preview, but I don't know if there's a good way to get usable example records without saving them to the database.

If we can't find a good way to do it with non-persisted records, I think I'd rather use the strategy from (for example) NoEnrolledStudentsAlertMailerPreview, which is to just pass SurveyResponseAlert.last to the preview and let it error if there are no such alerts in the development database.

lib/alerts/survey_response_alert_manager.rb
```diff
@@ -9,21 +9,33 @@ def create_alerts
...
+      question: answer.question.question_text,
+      answer: answer.answer_text,
+      followup: answer.follow_up_answer_text,
+      source: self.class.source(answer)
+    }
+  end
+
+  def self.source(answer)
+    question_group = answer.answer_group.question_group
+    survey_names = question_group.surveys.map(&:name).join(' .')
```
ragesoss, line 33: The idea here is to show all the surveys that this alert might have been generated from?

ragesoss:
Maybe a good way to handle this would just be to include commented code in the preview file that someone can use to create an example alert and the associated records.

spec/mailers/previews/alert_mailer_preview.rb
```diff
@@ -73,4 +79,49 @@ def example_no_med_training_for_course
...
+    answer = Rapidfire::Answer
+             .new(answer_text: 'Physics',
+                  follow_up_answer_text: 'Really ?',
+                  answer_group: answer_group,
+                  question: question)
+    answer.user = example_student
+    answer
+  end
+
+  def clean_up_answer
```
ragesoss, line 122: This doesn't clean up the new Survey.

lib/alerts/survey_response_alert_manager.rb
```diff
@@ -9,21 +9,33 @@ def create_alerts
...
+  def details(answer)
+    {
+      question: answer.question.question_text,
+      answer: answer.answer_text,
+      followup: answer.follow_up_answer_text,
+      source: self.class.source(answer)
+    }
+  end
+
+  def self.source(answer)
```
ragesoss, line 31: Is this just a class method so that it can be called from the preview?

cyrillefr: Actually, no. Method was too long and rubocop raised an offense. Si I made a method of detail;

lib/alerts/survey_response_alert_manager.rb
cyrillefr: Yes. I wanted to display both Survey(s) and question group, to have all the information.

cyrillefr:
Hello @ragesoss.
The reason we have an issue lies here in
`app models alert_types survey_response_alert`:
```ruby
  def question
    Rapidfire::Question.find subject_id
  end
```
There is a db request, so, an error will be raised if no related survey (question/group/survey). 

You are right, there is too much code in the preview, furthermore, it creates a survey that is not deleted afterwards.
So in production, it can be a problem (if there is a survey with 99999 id).

You talked about letting some code in comments to help create a survey(and questions etc).
Another possibility could be to indicate in comments where to create a survey (the path), this could be shorter. 
Or maybe add a Survey/Question/Answers seeds for local testing (the same way there is some data to ingest to do some local tests).
And so, an error will be raised if not enough data in the db. And I will remove all the code that creates surveys/groups the clean-up etc...

So, should i make a huge comment or write the path to survey creation ?
The thing is: the survey will stay ... it is ok in dev mode, less in production. But if is ok in dev, then it should be ok in prod(at least till a real survey an alert are made).
Or maybe, there could be a "dummy" survey even in production ?

ragesoss:
The mailer previews aren't accessible in production anyway.

I think I'd prefer a comment with code that can be used in the console to create an example alert and the related survey/question/answer records.

lib/alerts/survey_response_alert_manager.rb
ragesoss: Doing this as a class method is an unusual way to handle it, unless you need to be able to call it from somewhere else without having an instance already. It would be more idiomatic Ruby to use a private instance method  here.

spec/mailers/previews/alert_mailer_preview.rb
```diff
@@ -73,4 +85,44 @@ def example_no_med_training_for_course
...
+  public
+
+  def create_example_survey_response_alert
+    answer = create_answer
+    details =
+      {
+        question: answer.question.question_text,
+        answer: answer.answer_text,
+        followup: answer.follow_up_answer_text,
+        source: SurveyResponseAlertManager.source(answer)
```
ragesoss, line 98: This probably needs to become `SurveyResponseAlertManager.new.source(answer)`.

cyrillefr: Oups oups oups ...
Corrected

spec/mailers/previews/alert_mailer_preview.rb
```diff
@@ -73,4 +85,44 @@ def example_no_med_training_for_course
               article: example_article,
               course: example_course)
   end
+
+  public
+
+  def create_example_survey_response_alert
```
ragesoss, line 91: This method and `create_answer` show up as preview links at `/rails/mailers`. I'll comment them out, and update the instructions.

ragesoss:
Thanks @cyrillefr !

Issue #4727, opened 2022-01-07
Configure tests to run on GitHub Actions instead of travis-ci


ragesoss:
### Current Behavior:
Our CI is configured to run on travis, via `.travis.yml`. However, Travis-CI now has limited support for open source projects, and switching to GitHub Actions is probably going to be a more reliable way to run the tests.

### Desired Behavior:

Add a GitHub Actions config to run the complete test suite.

### Related Files & Code:

Here's an example CI config for a Rails project: https://github.com/diaspora/diaspora/blob/develop/.github/workflows/ci.yml

RishiKumarRay:
I would like to work on this

ragesoss:
@RishiKumarRay please do! Give me a ping if you run into any problems.

aaditya9899:
Hey @ragesoss I too working on this issue and made some progress ....i want to know what this command does to be exact 
`sudo apt-add-repository -y ppa:rael-gc/rvm` ?

ragesoss:
@aaditya9899 that command is only necessary on WSL; it's for installing the right Ruby version. It looks like it gets configured directly via the `ruby` value in ci.yml.

aaditya9899:
hey @ragesoss I have written this much code till now , i know there also some mistakes and few more things to add , could you tell me in brief what i have to do next?
`name: CI

on: [push]

jobs:

  build:

      runs-on: '${{ matrix.os }}'
    strategy:
      matrix:
      os:
          - ubuntu-latest
        ruby:
          - 2.7.1
        
        node-version: [16.x,17.x]
        redis-server:
        mariadb:

          
        db:
          - mysql
          - mariadb
      
      env:
        DB: ${{ matrix.db }}
        RAILS_ENV: test


        services:
        - redis-server
        

      steps:
         if: matrix.os == 'Linux'
      - name: Updating packages
        run: sudo apt-get update
        
      - name: Install dependencies
        run: sudo apt-get install -y redis-server mariadb-server libmariadb-dev rvm nodejs npm pandoc

      - name: Installing Yarn Globally
        run: sudo npm install --global yarn

         - name: Cache multiple paths
        uses: actions/cache@v2
        with:
          path: |
          -
            vendor
            node_modules


      - name: Start MySQL
        run: sudo service mysql start

        if: matrix.db == 'mysql'
      - uses: actions/checkout@v2
      - uses: ruby/setup-ruby@v2.7.1
        with:
[...]

PR #4751, opened 2022-01-28
Add Wikidata stats 

1v4n4:
Wikidata statistics is displayed on the course overview page in separate component (WikidataOverviewStats), if that statistics exist:

Wikidata stats with info description:
stats

Wikidata stats without info description:
noinfo

### DIFF
--- a/app/assets/javascripts/components/overview/info_stats.jsx
+++ b/app/assets/javascripts/components/overview/info_stats.jsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const InfoStats = ({ info }) => {
+  let iStats;
+
+  if (typeof info === 'string') {
+      iStats = (
+        <p>{info}</p>);
+  } else {
+  iStats = (info.map((x) => {
+        return (
+          <div key={`${x[1]}`}>
+            <h4 className="stat-display__value">{x[0]}</h4>
+            <p>{x[1]}</p>
+          </div>);
+  })
+  );
+  }
+
+  return (
+    <div className="tooltip dark" id="upload-usage">
+      {iStats}
+    </div>
+  );
+};
+
+InfoStats.propTypes = {
+  info: PropTypes.oneOfType([
+    PropTypes.array,
+    PropTypes.string
+  ])
+};
+
+export default InfoStats;
--- a/app/assets/javascripts/components/overview/overview_stats.jsx
+++ b/app/assets/javascripts/components/overview/overview_stats.jsx
@@ -2,9 +2,9 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import CourseUtils from '../../utils/course_utils.js';
 import ArticleUtils from '../../utils/article_utils.js';
-import CourseStat from './course_stat';
+import OverviewStat from './overview_stat';
 
-const CourseStats = ({ course }) => {
+const OverviewStats = ({ course }) => {
   const isWikidata = course.home_wiki.project === 'wikidata';
   const isWikipedia = course.home_wiki.project === 'wikipedia';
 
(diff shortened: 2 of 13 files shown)


app/assets/javascripts/components/overview/info_stats.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const InfoStats = ({ info }) => {
+  let iStats;
+
+  if (typeof info === 'string') {
+      iStats = (
```
ragesoss, line 8: Looks like some inconsistent whitespace throughout this file.

1v4n4:
I'm not having Github actions here, what can I do about it?

ragesoss:
This happens every few weeks whenever we run out of travis-ci credits. They dole them out for free to open source projects, but in relatively small increments and we have to request more each time we run out. I've just made the request; it will probably be working again by tomorrow.

(At some point we can probably move our tests to GitHub Actions, per https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/4727 )

ragesoss:
We have credits again, so CI should run the next time you push a commit.

ragesoss:
This is great progress! I think, as a start, we should not display any of the wikidata stats when the number is 0. I've pinged Will for his thoughts... my initial reaction is that we might want to visually separate the wiki-specific stats from the existing top-level stats, so that it's easy to look at them as see "400 articles edited" refers to cumulative activity across multiple wikis, while "200 items created" is specific to Wikidata. Design-wise, I'm not sure how to accomplish that.

1v4n4:
I was working on a first issue.  
Wikidata stats displayed without zero values (full stats in a console):
sc(1)
scr(1)

app/assets/javascripts/components/overview/overview_stats.jsx
```diff
@@ -108,96 +108,67 @@ const OverviewStats = ({ course }) => {
...
+          key={label}
+          id={Object.keys(stat)[0].split(' ').join('-')}
+          className={valueClass(label)}
+          stat={Object.values(stat)[0]}
+          statMsg={I18n.t(`metrics.${label}`)}
+          info={infoStats}
+          infoId={`${label}-info`}
+        />;
+      }
+      return <div />;
```
ragesoss, line 167: This is why there are empty slots for the zero-value stats? I think it shouldn't leave empty space for these. That's similar to how the top-level stats already behave; only 4 stats are shown when a course doesn't have any activity yet (users, articles edited, words added, edit count), while others like upload count are only rendered if nonzero, and the overview stats component renders them evenly spread out no matter how many there are.

app/assets/javascripts/components/overview/overview_stats.jsx
```diff
@@ -108,96 +108,67 @@ const OverviewStats = ({ course }) => {
...
-    const labelsInfo = [[statistics['labels changed'], I18n.t('metrics.labels_changed')],
-    [statistics['labels removed'], I18n.t('metrics.labels_removed')]];
-    const mergedInfo = [[statistics['merged to'], I18n.t('metrics.merged_to')]];
-    const otherUpdatesInfo = [[statistics['qualifiers added'], I18n.t('metrics.qualifiers_added')],
-    [statistics['redirects created'], I18n.t('metrics.redirects_created')],
-    [statistics['restorations performed'], I18n.t('metrics.restorations_performed')],
-    [statistics['reverts performed'], I18n.t('metrics.reverts_performed')],
-    [statistics['no data'], I18n.t('metrics.no_data')]];
-
-    wikidataStats = (
```
ragesoss: The approach here is much easier to read and make sense of than `map` version. I would suggest pushing the code for whether or not to render a stat with value zero into the OverviewStat component. Maybe it could be controlled with another prop, like `renderZero`. This way, we can easily control which stats should still be shown even if the value is zero, and which ones to hide — for both the Wikidata stats and the prior stats like `refCount` and `uploadCount`.

1v4n4:
I added two props renderZero and renderZeroInfo, so OverviewStat and OverviewStatInfo can be displayed with zero values, or not (there are 3 options - neither, both, or just OverviewStat. 

This is example where renderZero is true and renderZeroInfo false (all zeros in info):



For all Wikidata stats renderZero is on false:


Wikipedia course with items and a lot of stats:

1v4n4:
Should I refactor OverviewStatsInfo to be able to display only non-zero values? That would require another prop.
Now it could be hidden only if all are zeros.

ragesoss:
Cool! I think we should schedule a meeting with Will to demo this and talk through what makes sense for each of the individual stats.

In the meantime, I think factoring out WikidataOverviewStats into its own component would be a good idea.

ragesoss:
Is this ready to go from your perspective, @1v4n4 ?

1v4n4:
I think so. It works and we can change the style, add descriptions, or whatever at any time.

ragesoss:
Let's gooooo!

PR #4760, opened 2022-02-04
Fix DOM nesting error in Uploads list

1v4n4:
UploadsList and List are refactored not to render no uploads message inside of tbody HTML element.

Before, in the List and Tile view, an HTML table was generated even if there were no updates with a message (in div tags) inside. Having a div as a child element of tbody was causing an error and sometimes tests were failing because of it (screenshot).
Screenshot from 2022-02-06 22-05-57

I removed the no-updates message from the table and added tests for no uploads situation.

The design is slightly changed and no uploads message looks the same for all three views.

After:
gallery
list
tile

Before:
gallery-before
list-before
tile-before

### DIFF
--- a/app/assets/javascripts/components/uploads/upload_list.jsx
+++ b/app/assets/javascripts/components/uploads/upload_list.jsx
@@ -73,17 +73,19 @@ const UploadList = createReactClass({
     }
 
     if (this.props.view === LIST_VIEW) {
-      uploadsView = (
-        <div className="list-view">
-          <List
-            elements={elements}
-            keys={keys}
-            table_key="uploads"
-            sortBy={this.props.sortBy}
-            none_message={I18n.t('courses_generic.uploads_none')}
-          />
-        </div>
-      );
+      if (elements.length && elements.length > 0) {
+        uploadsView = (
+          <div className="list-view">
+            <List
+              elements={elements}
+              keys={keys}
+              table_key="uploads"
+              sortBy={this.props.sortBy}
+            />
+          </div>);
+      } else {
+          uploadsView = (<div className="list-view">{noUploadsMessage}</div>);
+        }
     }
 
     if (this.props.view === TILE_VIEW) {
--- a/app/assets/javascripts/components/uploads/upload_list.jsx
+++ b/app/assets/javascripts/components/uploads/upload_list.jsx
@@ -65,15 +65,11 @@ const UploadList = createReactClass({
     let uploadsView;
 
     if (this.props.view === GALLERY_VIEW) {
-      if (elements.length === 0) {
-        uploadsView = (<div className="list-view">{noUploadsMessage}</div>);
-      } else {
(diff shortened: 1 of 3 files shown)


spec/rails_helper.rb
```diff
@@ -1,164 +1,584 @@
...
-                 .chrome(chromeOptions: { w3c: false })
-  options = Selenium::WebDriver::Chrome::Options.new(
-    args: %w[headless no-sandbox disable-gpu --window-size=1200,1200]
-  )
-  Capybara::Selenium::Driver.new(app,
-                                 browser: :chrome,
-                                 options: options,
-                                 clear_local_storage: false, # Persist local storage across tests
-                                 desired_capabilities: capabilities)
+require 'rails_helper'
```
ragesoss, line 3: Looks like the new test got copied over into the wrong file.

ragesoss:
It would be helpful if you include a more detailed description of the error this fixes and what caused it (either in a commit message, or in the PR description).

1v4n4:
I tried. Is this better?

1v4n4:
What do you think about this one? It is somewhere between those two approaches, I think it will solve the problem. Screenshots still apply.

app/assets/javascripts/components/uploads/upload_list.jsx
```diff
@@ -62,31 +65,34 @@ const UploadList = createReactClass({
...
+      } else {
+          uploadsView = (<div className="list-view">{noUploadsMessage}</div>);
+        }
     }
 
     if (this.props.view === TILE_VIEW) {
       uploadsView = (
         <div className="tile-view">
-          {elements}
+          {uploads.length > 0 && elements}
```
ragesoss, line 94: Do you prefer doing it this way over combining this to one statement so that it's clearer that it will show `elements` OR `noUploadsMessage`? For example, `uploads.length > 0 ? elements : noUploadsMessage`.

app/assets/javascripts/components/uploads/upload_list.jsx
```diff
@@ -62,31 +65,34 @@ const UploadList = createReactClass({
...
-          <List
-            elements={elements}
-            keys={keys}
-            table_key="uploads"
-            sortBy={this.props.sortBy}
-            none_message={I18n.t('courses_generic.uploads_none')}
-          />
-        </div>
-      );
+      if (elements.length && elements.length > 0) {
```
ragesoss, line 76: Are both of these checks necessary? I'd think either one would be sufficient. I can't think of a way that `elements.length` would be truthy but not greater than 0.

ragesoss:
This approach looks fine to me.

app/assets/javascripts/components/uploads/upload_list.jsx
1v4n4: oh, got it. I thought it will be undefined if elements is not an array

PR #4779, opened 2022-02-10
Add a GitHub Actions config to run the complete test suite 

aaditya9899:
## What this PR does
The use of travis has limited support and not so in use nowadays , that's why this will be a reliable approach to run tests.

## Screenshots

final commit

It was my first time that I used GitHub actions , infact created the whole pipeline for it with the help of @ragesoss so I am pretty happy that I learnt something new.


aaditya9899:
done, kindly review my pull request , the commits are very messy , it wont happen again as I have done all the things in browser itself so i was not able to stash them all under one commit but will ensure it from next time otherwise will do something to make it look cleaner if you say so.

ragesoss:
Thanks! I guess this repo isn't configured to use GitHub actions yet, but your fork is. What steps did you need to take to enable it? Can you link to the latest run on your fork?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+  # Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch:
+
+jobs:
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest 
+
+    steps:
+      - uses: actions/checkout@v2
```
ragesoss, line 20: What does this do? A comment would be helpful here.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+
+           
+      - name: Setup Node
+        uses: actions/setup-node@v1
+               
+      - run: 'export PATH=$PATH:`yarn global bin`'
+      
+      - run: export DB=test
+      
+      - run: export SENDER_EMAIL_ADDRESS='sender@wikiedu.org'
```
ragesoss, line 53: I'm almost certain this is not necessary (here or in the travis-ci config).

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+           ruby-version: 2.7.1
+           bundler-cache: true
+
+           
+      - name: Setup Node
+        uses: actions/setup-node@v1
+               
+      - run: 'export PATH=$PATH:`yarn global bin`'
+      
+      - run: export DB=test
```
ragesoss, line 51: This shouldn't be necessary, since the database name is coming from database.yml via the `db:create` command later on.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+      - uses: ruby/setup-ruby@v1    
+        with:
+           ruby-version: 2.7.1
+           bundler-cache: true
+
+           
+      - name: Setup Node
+        uses: actions/setup-node@v1
+               
+      - run: 'export PATH=$PATH:`yarn global bin`'
```
ragesoss, line 49: This is probably not necessary. I think it is specific to how we installed `yarn` on travis-ci.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+
+      - name:  Creating Test Database 
+        env:
+          RAILS_ENV: test
+          DATABASE_PORT: 3306
+        run: |
+            bin/rails db:create
+            bin/rails db:migrate RAILS_ENV=test
+      
+      - name: Running tests
```
ragesoss, line 76: It looks like this isn't running the the JS tests, which are run via `yarn test`.

Ideally, in terms of the output and how a test run shows up, we want to know which of the test suites and checks passed and which failed. Those consist of: JS tests (`yarn test`), Ruby tests (`bundle exec rspec ...`), Ruby linting (`bundle exec rubocop`) and JS linting (`yarn lint-non-build`). Maybe that means putting each one in a separate step?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+      
+      - run: export DB=test
+      
+      - run: export SENDER_EMAIL_ADDRESS='sender@wikiedu.org'
+      
+      - name: Installing Packages 
+        run: |
+            yarn install
+            gem install bundler
+            gem install rubocop
```
ragesoss, line 59: I'm not sure that this and the following are necessary, since we call rubocop via `bundle exec`... so it should be installed as part of the bundle, and there's no need to also install it globally. At least, I think so.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,86 @@
...
+            gem install rubocop-rspec-focused
+            bundle config set path 'vendor'
+            bundle install --jobs=3 --retry=3
+            
+      - run: mkdir tmp -m 777
+      - run: cp config/database.example.yml config/database.yml
```
ragesoss, line 65: Rather than changing the username/password in the example file to match what GitHub Actions expects, I think we should add a new file explicitly for this Actions workflow and use that instead of the example file.

.github/workflows/ci.yml
aaditya9899: yeah this action checks-out your repository under $GITHUB_WORKSPACE , so your workflow can access it.

aaditya9899: I will make the comment above that .

.github/workflows/ci.yml
aaditya9899: okay then i will be removing it and see if it works fine

.github/workflows/ci.yml
aaditya9899: ohh yeah absolutely , will be removing these two also.

.github/workflows/ci.yml
aaditya9899: yeah i thought so too but i was bit hesitant to remove it due to the travis file but will do it now thank.

.github/workflows/ci.yml
aaditya9899: oh no i forget that too add in my final commit i will be putting it under the test itself and will definitely work fine or will sperate few of them under linting  to make it look cleaner.

.github/workflows/ci.yml
aaditya9899: okay will try the test without these and see if this works or not but i remeber it was conflicting with the version of ruby and gems file that's why doing this made the dependencies work but i'll see to it

.github/workflows/ci.yml
aaditya9899: so like making a new file with the same content except username and password and then naming it something else?

.github/workflows/ci.yml
aaditya9899: this worked absolutely fine !!!!... Initially i was having trouble setting up correct ruby version that's why doing these bring the packages on even level but after installing correct ruby version , it installed the bundler and other gem files with same version from Gemfile.lock and also removing the above said things by you didn't make any difference in the workflow, thankx.

ragesoss:
> Thanks! I guess this repo isn't configured to use GitHub actions yet, but your fork is. What steps did you need to take to enable it? Can you link to the latest run on your fork?

@aaditya9899 did you see this?

aaditya9899:
> 
Here's my latest run
https://github.com/aaditya9899/WikiEduDashboard/actions/runs/1828236819/workflow

And it's very easy to set up:-
1. You have to delete `.travis.yml`  file
2. you have to create a folder in `.github` named **workflows**
3. under that , create a `ci.yml` file and paste my code in it.

That's it.

aaditya9899:
> > 
> 
> Here's my latest run https://github.com/aaditya9899/WikiEduDashboard/actions/runs/1828236819/workflow
> 
> And it's very easy to set up:-
> 
> 1. You have to delete `.travis.yml`  file
> 2. you have to create a folder in `.github` named **workflows**
> 3. under that , create a `ci.yml` file and paste my code in it.
> 
> That's it.

is there any issue merging my code like normally we do in github??

aaditya9899:
If that so, i can delete this repository and make the necessary changes again under one commit.

aaditya9899:
> > Thanks! I guess this repo isn't configured to use GitHub actions yet, but your fork is. What steps did you need to take to enable it? Can you link to the latest run on your fork?
> 
> @aaditya9899 did you see this?

oh no , I haven't seen that sorry :')

ragesoss:
Okay, so it sounds like if I just merge this, it should run the CI without any further changes.

aaditya9899:
> Okay, so it sounds like if I just merge this, it should run the CI without any further changes.

Yes absolutely

aaditya9899:
I don't like doing work like this with so much mess , but it was all done in the github itself and everytime I make changes it gets into another commit automatically but when i will work on the application code base , I will be creating a new branch and work on that .

ragesoss:
I'm looking at the test run you linked, and it looks like the tests only take 2 minutes to run — which is much shorter than I expect — and also, the coverage report at the end shows 31% Ruby coverage, compared to the expected value of ~100%. Not sure what is going on there.

ragesoss:
@aaditya9899 let's keep the .travis-ci.yml for this PR, so that we can keep using Travis until we're confident the GitHub Actions workflow is doing everything we want.

aaditya9899:
> @aaditya9899 let's keep the .travis-ci.yml for this PR, so that we can keep using Travis until we're confident the GitHub Actions workflow is doing everything we want.

yeah sure i think it will be best

aaditya9899:
> I'm looking at the test run you linked, and it looks like the tests only take 2 minutes to run — which is much shorter than I expect — and also, the coverage report at the end shows 31% Ruby coverage, compared to the expected value of ~100%. Not sure what is going on there.

oh wait i remember i actually made rspec run in **dry run** At first, it sounds a little bit weird, but yes, it’s possible to run tests without running them. It’s called a dry run, if you have a large tests codebase and you would like to see how many specs are there or how many are pending, you can perform a dry run that won’t trigger the tests, but the rspec will behave as we would run them , I did this because I have to check make workflow run faster as I was doing continuous changes to make all the thing work , let me remove it but it was taking too i think almost that alone process took more than 10 minutes , is it fine or is there any other way also to  make it faster but i dont know yet?

ragesoss:
@aaditya9899 the travis-ci tests take about 20-25 minutes, so I'd expect something similar here. That's mainly due to the number of rspec tests.

aaditya9899:
> tHub actions yet, but your fork is. What steps did you need to take to enable it? Can you link to the latest run on your fork?



> @aaditya9899 the travis-ci tests take about 20-25 minutes, so I'd expect something similar here. That's mainly due to the number of rspec tests.

hey the test stopped while running the complete rspec  , is this my fault or something already there in the codebase ?

here's the link:-

https://github.com/aaditya9899/WikiEduDashboard/runs/5160154077?check_suite_focus=true

ragesoss:
It looks like we need to be able to get pandoc in the github environment. Maybe there's an action that can do that.

I think keeping the fail-fast behavior is good.

ragesoss:
Probably this: https://github.com/r-lib/actions/tree/master/setup-pandoc

aaditya9899:
yes i just noticed pandoc there so i will be keeping the idea of system updates and 

> Probably this: https://github.com/r-lib/actions/tree/master/setup-pandoc

thank you i remember removing system updates and that from my initial code , should i also include ` sudo apt-get update && sudo apt-get dist-upgrade` again or just pandoc would be fine?

aaditya9899:
> yes i just noticed pandoc there so i will be keeping the idea of system updates and
> 
> > Probably this: https://github.com/r-lib/actions/tree/master/setup-pandoc
> 
> thank you i remember removing system updates and that from my initial code , should i also include ` sudo apt-get update && sudo apt-get dist-upgrade` again or just pandoc would be fine?

i think since its pulling ubuntu latest , it will be having the latest dependencies of the OS , what do you think?

ragesoss:
Hopefully just adding pandoc will be enough. I suspect there are some other things that will turn up from other tests as well.

aaditya9899:
> Hopefully just adding pandoc will be enough. I suspect there are some other things that will turn up from other tests as well.

that's why removed fail fast for now to run the complete test and get all the broken things at once then I will add it in my final commit if everything works out

aaditya9899:
> > Hopefully just adding pandoc will be enough. I suspect there are some other things that will turn up from other tests as well.
> 
> that's why removed fail fast for now to run the complete test and get all the broken things at once then I will add it in my final commit if everything works out

this is the result it produced :

**Finished in 23 minutes 29 seconds (files took 10.41 seconds to load)
1682 examples, 13 failures, 13 pending**

can you check what's the fault , here's the link:-

https://github.com/aaditya9899/WikiEduDashboard/runs/5160520606?check_suite_focus=true

aaditya9899:
> > > Hopefully just adding pandoc will be enough. I suspect there are some other things that will turn up from other tests as well.
> > 
> > 
> > that's why removed fail fast for now to run the complete test and get all the broken things at once then I will add it in my final commit if everything works out
> 
> this is the result it produced :
> 
> **Finished in 23 minutes 29 seconds (files took 10.41 seconds to load) 1682 examples, 13 failures, 13 pending**
> 
> can you check what's the fault , here's the link:-
> 
> https://github.com/aaditya9899/WikiEduDashboard/runs/5160520606?check_suite_focus=true

wait i have to add redis server also , is there any other thing missing?

ragesoss:
Looks like redis might be the last thing it needs.

aaditya9899:
> Looks like redis might be the last thing it needs.



> Looks like redis might be the last thing it needs.

Hey @ragesoss  adding fail fast as final commit and when you think this is ready , merge it with the main branch ,  will look for more issue of this project to work upon. Thankx, I literally learnt many many things from this and you also.

ragesoss:
Yay! It looks like this is working with the whole test suite now.

Next steps to improve it would be to find as many ways to cut down on the number of non-test steps, by combining them wherever possible. I'd like the JS tests, Ruby tests, JS linting, and Ruby linting to each have their own step... everything else is setup, and should be combined as much as possible.

aaditya9899:
> Yay! It looks like this is working with the whole test suite now.
> 
> Next steps to improve it would be to find as many ways to cut down on the number of non-test steps, by combining them wherever possible. I'd like the JS tests, Ruby tests, JS linting, and Ruby linting to each have their own step... everything else is setup, and should be combined as much as possible.

yeah sounds good as they are the only which will matter to the developers to see the bug efficiently , on it 👍

aaditya9899:
> > Yay! It looks like this is working with the whole test suite now.
> > Next steps to improve it would be to find as many ways to cut down on the number of non-test steps, by combining them wherever possible. I'd like the JS tests, Ruby tests, JS linting, and Ruby linting to each have their own step... everything else is setup, and should be combined as much as possible.
> 
> yeah sounds good as they are the only section which will matter to the developers to see the bug efficiently , on it 👍

Here's the link

https://github.com/aaditya9899/WikiEduDashboard/runs/5165990695?check_suite_focus=true

PR #4785, opened 2022-02-12
Fix number to human bug

1v4n4:
OverviewStat component is refactored to take strings as a prop in addition to numbers,  (because numbers greater than 1000 are transformed to strings by number_to_human in course.json).
Total revisions are added to stats_hash in CourseStat. If there are no total revisions (before the next update) that stat would not be displayed.
Number_to_human is added to stats_hash values in course.json.
As a result, top and Wikidata stats are displayed the same way in the course overview tab (first screenshot) and course.json (second screenshot).
tab
json


ragesoss:
Looks good to me!

PR #4790, opened 2022-02-15
Generate Wikidata stats CSV from CourseStat

1v4n4:
Since we have a CourseStat table, there is no need for CourseWikidataCsvBuilder to analyze revisions each time when building CSV. I removed redundant code and refractor CourseWikidataCsvBuilder to generate CSV from CourseStat record


ragesoss:
Easy peasy! I assume you've checked that this produces the same output before and after?

1v4n4:
> Easy peasy! I assume you've checked that this produces the same output before and after?

Even better - I screenshoted it.
Before:
oldcsv
After:
newcsv

ragesoss:
:chefs_kiss:

1v4n4:
Is it OK for no-Wikidata stats CSV to look like this?
nostats

Before it was list of all categories with zeros. I can make it the same, but I think I will be adding more code for no good reason.

PR #4792, opened 2022-02-16
Add unknown to other wikidata updates

1v4n4:
The Sum of other and unknown updates is displayed as Other updates in the course.json and WikidataOverviewStats component.
I added a views helper method for adding unknown to other updates and removing from stats hash. The plan is to use that method for Wikidata stats in campaigns also.

The info is removed from Redirects created.

After:
other
otherjson
Before:
otherjsonbefore
otherbefore

### DIFF
--- a/app/assets/javascripts/components/overview/wikidata_overview_stats.jsx
+++ b/app/assets/javascripts/components/overview/wikidata_overview_stats.jsx
@@ -173,8 +173,6 @@ const WikidataOverviewStats = ({ statistics }) => {
               stat={statistics['redirects created']}
               statMsg={I18n.t('metrics.redirects_created')}
               renderZero={true}
-              info={I18n.t('metrics.redirects_info')}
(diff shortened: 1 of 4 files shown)


ragesoss:
I think these should be added together server-side so that the Rails version of number_to_human remains the only one we use. Otherwise, there will be inconsistent internationalization (since the Rails version automatically takes account of locale when producing the 'human' versions of numbers).

1v4n4:
Ok. So I should leave CourseStats as it is, but add unknown to other updates for course.json, right?

ragesoss:
Yes, exactly.

1v4n4:
This test is ok at my place and I don't think it is related to what I'm doing here.
test

ragesoss:
Yes, this was caused by wikiedu.org being down while the tests ran. I've restarted build, to make sure no later tests were broken.

PR #4796, opened 2022-02-18
Slide handler to FC

TheTrio:
## What this PR does
This PR converts the Training Slide Handler to a functional component. I've also replaced `connect` with the `useSelector` and `useDispatch` hooks. Similarly, I've removed `withRouter` and instead used the `useHistory` and `useParams` hooks. 

The title of the page now changes with each slide change. 
I'm still not entirely sure about which functions to place inside of the component and which ones outside. Currently, functions which dispatch actions have been included inside of the component.  

Also, since the thunk creator functions(`setSlideCompleted`, `setCurrentSlide`) and the functions which actually call them have the same name, I have simply added `_FC` in the name for those defined inside of the Functional Component. This wasn't a problem before since it was `this.setSlideCompleted` vs `this.props.setSlideCompleted`. Don't know if these are the best names but :shrug: 

And as always, I'm not entirely sure about the format of the title. Here's how it looks currently- 

titlesPreview

### DIFF
--- a/app/assets/javascripts/training/components/training_slide_handler.jsx
+++ b/app/assets/javascripts/training/components/training_slide_handler.jsx
@@ -88,6 +88,8 @@ const TrainingSlideHandler = () => {
     dispatch(setCurrentSlide(training.previousSlide.slug));
   };
 
+  // runs when the component is first rendered
+  // fetches the initial data and sets the base title
   useEffect(() => {
     setBaseTitle(document.title);
     const slideId = __guard__(routeParams, x => x.slide_id);
@@ -95,6 +97,9 @@ const TrainingSlideHandler = () => {
     dispatch(fetchTrainingModule({ module_id: moduleId(routeParams), slide_id: slideId, user_id: userId }));
   }, []);
 
+  // runs whenever the training from the redux store changes
+  // which means its run essentially when the user goes from one slide to another
+  // changes the page title according to the slide title and updates the event listener
   useEffect(() => {
     const handleKeyPress = (e) => {
       const navParams = { library_id: routeParams.library_id, module_id: routeParams.module_id };
@@ -118,7 +123,7 @@ const TrainingSlideHandler = () => {
     document.title = `${slideTitle} - ${baseTitle}`;
 
     return () => {
-      // cleanup
+      // cleanup. Removes the old event listener
       return window.removeEventListener('keyup', handleKeyPress);
     };
   }, [training]);
--- a/app/assets/javascripts/training/components/training_slide_handler.jsx
+++ b/app/assets/javascripts/training/components/training_slide_handler.jsx
@@ -1,281 +1,268 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import createReactClass from 'create-react-class';
-import { withRouter } from 'react-router';
+import React, { useEffect } from 'react';
+import { useHistory, useParams } from 'react-router';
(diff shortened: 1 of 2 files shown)


TheTrio:
Huh. I looked at the test which was failing and I can't really understand it. Here's what it has

```js
const TestLink = mount(
    <Provider store={store}>
      <MemoryRouter initialEntries={['/training/foo/bar/kittens']}>
        <Route
          exact
          path="/training/:library_id/:module_id/:slide_id"
          render={({ match }) => (
            <TrainingSlideHandler
              loading={false}
            >
              <SlideLink
                slideId="foobar"
                buttonText="Next Page"
                disabled={false}
                button={true}
                onClick={jest.fn()}
                params={match.params}
              />
            </TrainingSlideHandler>
          )}
        />
      </MemoryRouter>
    </Provider>
  );
```
But afaik, even before, the `TrainingSlideHandler` component didn't take any props - it is dependent solely on `training` which it gets from the Redux store. This is how its called in the `training_app.jsx` file - without any props. 

```js
const TrainingApp = () => (
  <div>
    <Switch>
      <Route exact path="/training/:library_id/:module_id" component={TrainingModuleHandler} />
      <Route exact path="/training/:library_id/:module_id/:slide_id" component={TrainingSlideHandler} />
    </Switch>
  </div>
);
```
So what is the test exactly trying to do? Since it was passing before, I assume I messed up something but I can't find it.

TheTrio:
Have fixed the test. Turns out it was merely a coincidence that it was passing before. For example, I changed the button text from "Next Page" to "Click Here"

```js
<TrainingSlideHandler
  loading={false}
>
  <SlideLink
    slideId="foobar"
    buttonText="Click Here"
    disabled={false}
    button={true}
    onClick={jest.fn()}
    params={match.params}
  />
</TrainingSlideHandler>
```

Did the same in the tests. 

```js
it('renders correct text', () => {
  expect(domBtn.text()).toEqual('Click Here');
});
```

The test failed. 

image

The "Next Page" was merely a default. In fact, as I said earlier, `TrainingSlideHandler` doesn't take any props at all - so everything passed to it was essentially ignored.

ragesoss:
I tested this out locally, and it seems to work perfectly. This component is too complex for me to really think the changes through in terms of code — and it's doing some things we've never done in functional components before, such as replacing `connect` with `useDispatch` and `useSelector`. But as far as I can tell, it's doing everything the right way.

I think it would benefit from some comments to explain what each of the `useEffect` calls is for and when it fires.

TheTrio:
> I think it would benefit from some comments to explain what each of the useEffect calls is for and when it fires.

Have added some comments. Hopefully it explains the purpose of the changes well enough.

app/assets/javascripts/training/components/training_slide_handler.jsx
```diff
@@ -118,7 +123,7 @@ const TrainingSlideHandler = () => {
     document.title = `${slideTitle} - ${baseTitle}`;
 
     return () => {
-      // cleanup
+      // cleanup. Removes the old event listener
```
ragesoss, line 126: Does this cleanup run every time the reduce store changes, or only when the component unmounts?

TheTrio: Everytime, since the needs the updated value of `training`. When the component unmounts it removes the event listener for the last time.

Issue #4621, opened 2021-10-01
Milestones component doesn't always number weeks the same way as Timeline


ragesoss:
### What is happening?
The Timeline for a course lists all the Weeks for the course, and also accounts for weeks where the course is not meeting and weeks before the start of the Wikipedia assignment. For example, a Timeline might start with Week 2 (because the course `timeline_start` is a week after the course `start`, and then have an empty vacation week, following by Week 4, even though there are only two Week records for the course.

The Timeline component handles all of this correctly. However, the Milestones component used on the Overview tab does not account for empty vacation weeks, so the week numbering for where the Milestone blocks occur can be different from the week numbering on the Timeline.

### To Reproduce

1. Create course with a timeline, add time between the course start and the timeline start, and set the meeting days so that there is a week with no meetings at some point after the timeline start.
2. Add a few Milestone blocks throughout the timeline.
3. Look at the week numbering on the Timeline tab vs. the Milestones component of the Overview tab.

### Expected behavior
All the week labels should match between tabs.

### Additional context
The right way to fix this is probably to create a canonical representation of the timeline — including the dates, week labels and empty weeks — outside of the individual components that process the timeline. This probably means implementing a selector that combines `course` and `weeks` data using the same approach currently found in the `timeline.jsx` component. Then this selector can be used in both the Timeline and Milestones components.

subhangi2731:
hello i am an outreachy applicant. I am interested to contribute. can u please guide me here and can I take up this issue

ragesoss:
@subhangi2731 yes, you can work on this. It's at the more complex end of the ones I've marked newcomer friendly, as it involves a refactor of multiple components and the use of Redux and selectors.

ragesoss:
@subhangi2731 have you reproduced the problem locally yet?

vaidehi44:
@ragesoss, I was able to reproduce this bug. Can I work on it?

ragesoss:
yes, go for it.

vaidehi44:
Hi @ragesoss, I got the reason why this is happening. I want to know what you mean by 'selector' in the description of issue.

vaidehi44:
> Hi @ragesoss, I got the reason why this is happening. I want to know what you mean by 'selector' in the description of issue.

Okay, I found the selectors folder which lists all the selectors. So, got an idea of implementing it.

vaidehi44:
Hey @ragesoss. I have a opened a PR. Please have a look.

[...]

PR #4797, opened 2022-02-19
Make week numbers in milestone and timeline sections same [WIP]

vaidehi44:
## What this PR does
This PR tries to resolve this issue #4621. 

Currently, the weeks in timeline section and overview tab don't show same order/number. The overview tab shows weeks which have milestone blocks. Hence, the milestones component is not correctly numbering the weeks.

The milestones component filters milestone blocks from `weeks` derived from `getWeeksArray` selector. Further, it uses `order` property to number weeks. However, the array returned by this selector only contains non-empty weeks (weeks with scheduled meetings). And, the `order` property also does not take account of this. Hence, I have defined a new selector `getAllWeeksArray` which returns an array of all the weeks including empty ones. Each week object contains keys: `empty` and `weekNumber`. If the week is non-empty, it also includes the other content being returned by `getWeeksArray` selector.

### Timeline Section: Milestone blocks defined in week 3 and 8.
week3_cropped
week8_cropped

### Milestones in overview tab:
Before: Milestone tab numbering - 2 and 5.
image

After: Milestone tab numbering - 3 and 8.
image

- Right now, I have just made a simple selector which groups empty and non-empty weeks in proper order. It is helping in determining if a week is empty or not, and giving its exact number through `weekNumber`. I was able to combine all these weeks using the `weekMeetings` method of `CourseDateUtils`, I am not sure if its the only correct way to do it.  Also, I think the selector can be improved and would like to get your guidance in that case.
- I am not understanding why in the first place we haven't designed a selector which would also consider empty weeks. Or, why haven't we maintained empty and non-empty weeks and their properties together.
- I am not sure if the `weekIsCompleted` method of Milestone component is working right. As you can see in the screenshot, in overview tab, it shows week 3 complete, but week 3 is still the ongoing week.


ragesoss:
Looks like you may need to edit one or more of the JS tests to account for the new `allWeeks` prop.

vaidehi44:
> Looks like you may need to edit one or more of the JS tests to account for the new `allWeeks` prop.

Yeah. Will do it.

vaidehi44:
@ragesoss, I have tried to fix some errors raised from the build. Now, I am stuck at Ruby tests. Seems like it is raised by `timeline_rearranging_spec.rb`. Can you please help?

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -271,7 +273,71 @@ const Timeline = createReactClass({
...
+              editableBlockIds={this.props.editableBlockIds}
+              edit_permissions={this.props.edit_permissions}
+              saveBlockChanges={this.props.saveBlockChanges}
+              setBlockEditable={this.props.setBlockEditable}
+              cancelBlockEditable={this.props.cancelBlockEditable}
+              updateBlock={this.props.updateBlock}
+              addBlock={this.props.addBlock}
+              deleteBlock={this.props.deleteBlock}
+              saveGlobalChanges={this.props.saveGlobalChanges}
+              canBlockMoveUp={this._canBlockMoveUp.bind(this, week, (week.order - 1))}
```
vaidehi44, line 327: May be, there is something wrong which I have done here. But, I can't figure it out.

vaidehi44:
May be, there is something wrong which I have done here. But, I don't exactly know.
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/dc73305f091bf8949673a63c80a8c2782be704df/app/assets/javascripts/components/timeline/timeline.jsx#L327-L328

vaidehi44:
Hey @ragesoss, it has passed all the checks now. The error had occurred because the numbering of `order` prop of `week`, in `timeline_rearranging_spec`, was zero-indexed while normally it starts from 1. And, in the new code, I am using `order` to decide if blocks can be rearranged or not (in `canBlockMoveUp` and `canBlockMoveDown`). Hence, I made the change in spec file. Hope it is correct.

ragesoss:
@vaidehi44 thanks! I'm planning to have a look at this soon.

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -208,6 +208,8 @@ const Timeline = createReactClass({
     // week meetings, which indicates a blackout week. Then insert the week itself.
     // The index 'i' represents the zero-index week number; both empty and non-empty
     // weeks are included in this numbering scheme.
```
ragesoss, line 210: These block of comments probably needs to be updated?

ragesoss: Actually, it probably needs to be moved to the selector.

vaidehi44: Okay. Will correct it right way.

ragesoss:
I want to test this out locally and study the code a bit more, but this looks good.

I think there are some possible follow-up changes that should be done after this:
* Rather than loop through allWeeks and have an `if/else` check to insert either a Week or an EmptyWeek, this could probably now be rewritten so that every week in allWeeks renders one Week component, and if it's empty then the `Week` component is responsible for rendering an EmptyWeek instead of the its normal output.
* The number of props required for `Week` can probably be reduced. At the least, anything that is a property of the `week` or `course` props should be removable.
* the `ThisWeek` component still has its own implementation for figuring out which week is "this week", and it basically duplicates the empty week logic that got moved to the selector. It could be rewritten to use allWeeks as well.

vaidehi44:
> I want to test this out locally and study the code a bit more, but this looks good.
> 
> I think there are some possible follow-up changes that should be done after this:
> 
> * Rather than loop through allWeeks and have an `if/else` check to insert either a Week or an EmptyWeek, this could probably now be rewritten so that every week in allWeeks renders one Week component, and if it's empty then the `Week` component is responsible for rendering an EmptyWeek instead of the its normal output.
> * The number of props required for `Week` can probably be reduced. At the least, anything that is a property of the `week` or `course` props should be removable.
> * the `ThisWeek` component still has its own implementation for figuring out which week is "this week", and it basically duplicates the empty week logic that got moved to the selector. It could be rewritten to use allWeeks as well.

So, some doubts about the follow-up changes which you have suggested:

- If we have to enable `Week` component to render empty weeks too, then `EmptyWeek` component should be entirely removed, right?
- About reducing the props in `Week` component; current props like `index`, `blocks` which can be obtained from `week` prop should be removed and, props like `timeline_start`, `timeline_end` which can be obtained from `course` prop should be removed, right? If yes, then will have to include `course` prop in `Week` as it is not present currently, right?

ragesoss:
I'm not suggesting that these need to be done as part of this PR, just noting the areas for further improvement so I don't forget them.

> If we have to enable Week component to render empty weeks too, then EmptyWeek component should be entirely removed, right?

What I meant was that the EmptyWeek component is returned by the Week component, when appropriate.

> About reducing the props in Week component; current props like index, blocks which can be obtained from week prop should be removed and, props like timeline_start, timeline_end which can be obtained from course prop should be removed, right? If yes, then will have to include course prop in Week as it is not present currently, right?

Yes.

vaidehi44:
@ragesoss, I am happy that it is working as expected. If you think that `allWeeks` property is fine and nothing more needs to be changed or added, then its completed from my side too.
I just feel that the `weekIsCompleted` method of Milestone component is not working right. The method indicates if a milestone week is completed or not in the overview component. Have also mentioned it in  'Open Questions and concerns'. Will confirm it again and let you know.

ragesoss:
Ah, right. I forgot about that weekIsCompleted issue you mentioned and did not test for that. If you're having trouble figuring out why it's not behaving as expected, one good strategy could be to write a test for the expected behavior and then dig in to figure out why it doesn't match that.

vaidehi44:
Okay. Would like to try testing. But, I don't have much idea about how to writes a test. Can you provide a starting point or an example.

ragesoss:
In this case, since the behavior you want to test is based on the week data you pass to `Milestones` and the simple `weekIsCompleted` function, I suggest going with a plain JS test. The best starting point for an example, and possibly even the place where you want to write the test(s), is `course_date_utils.spec.js`. You can recreate the `week` and `currentWeek` values that are used in the Milestones component, and just test those in isolation. It's probably something related to how `currentWeek` gets calculated, or perhaps something related to `week.order`.

vaidehi44:
Hi @ragesoss. 
I was able to figure out the problem with milestone's `weekIsCompleted` method. Originally, it was using `week.order` and `currentWeek` to compare the order. `currentWeek` is derived from `CourseDateUtils`'s `currentWeekOrder`. But, `currentWeekOrder` has nothing to do with `week.order` prop, as it counts the current week from `timelineStart`. Whereas, `order` prop of `week` is only applicable for non-empty weeks and won't account for the empty ones.
So, a better and simple way is to just compare it (`currentWeek`) with our new `weekNumber` property. This prop also numbers the weeks from `timelineStart` (which is the start of assignment timeline). Hence, this solves the issue.

ragesoss:
Great. is `currentWeekOrder` calculated the same way as `week.weekNumber` in terms of whether numbering starts with 0 or 1?

vaidehi44:
> Great. is `currentWeekOrder` calculated the same way as `week.weekNumber` in terms of whether numbering starts with 0 or 1?

Yeah. They both start with 1. `currentWeekOrder` adds 1 to `currentWeekIndex` inorder to start from 1. And, `weekNumber` is already indexed starting from 1.

PR #4802, opened 2022-02-23
React campaign top stats 

1v4n4:
This is switching the campaign's top stats (and navbar) to React.
I used the same component as for the course's top stats, moved React root element at the top of all tabs, and refactored Campaign.jsx to handle some more components. 
Rails navbar is replaced with React's component for all tabs.

Before:
before1
before2

After:
after1
after3
after4
after5
after6

The programs tab is now consistent with others.
Before:
before3
After:
programs

Download stats modal button is pinned to the lower part of the page by position absolute. If there are any buttons in the sidebar above, its position is corrected. It is responsive.
The page when user not logged in:
modal

I fixed some random layout problems:
scshot
scshot1
After:
after7
after8


1v4n4:
Top stats for the campaign: 
stats

 React top stats for the campaign with OverviewStats components (I still need to put them into the right place):
statsreact

app/assets/javascripts/components/campaign/campaign.jsx
```diff
@@ -29,24 +26,18 @@ export const Campaign = createReactClass({
...
-      return <CampaignOverviewHandler {...this.props} />;
-    }
+
     return (
-      <div>
-        <CampaignNavbar
-          campaign={this.props.campaign}
-        />
+      <div className="container">
+        <CampaignOverviewHandler {...this.props} />
```
ragesoss, line 32: In terms of naming conventions, the 'overview handler' should be the component that specifically handles the `overview` tab, as opposed to shared component that handles the stats for all the tabs.

app/assets/javascripts/components/campaign/campaign_overview_handler.jsx
```diff
@@ -1,8 +1,30 @@
...
+  const location = useLocation();
+  let statsModal;
+  if (location.pathname.match(/overview/)) {
+    statsModal = (
+      <div className="stats-download-modal">
+        <CampaignStatsDownloadModal {...props} />;
+      </div>);
+  }
+  let navBar;
+  if (location.pathname.match(/ores_plot/) || location.pathname.match(/alerts/)) {
```
ragesoss, line 17: I don't like this. Rendering the CampaignNavbar (and the CampaignStats) should be the responsibility of the top `Campaign` component, rather than the overview handler.

app/assets/javascripts/components/campaign/campaign_overview_handler.jsx
```diff
@@ -1,8 +1,30 @@
...
+import CampaignNavbar from '../common/campaign_navbar';
 
 const CampaignOverviewHandler = (props) => {
-  return <CampaignStatsDownloadModal {...props} />;
+  const location = useLocation();
+  let statsModal;
+  if (location.pathname.match(/overview/)) {
+    statsModal = (
+      <div className="stats-download-modal">
+        <CampaignStatsDownloadModal {...props} />;
```
ragesoss, line 13: I see this extra semi-colon in the screenshot.

1v4n4:
I think I made all changes, it looks much better now, I updated screenshots

app/assets/javascripts/components/campaign/campaign.jsx
```diff
@@ -30,13 +29,24 @@ export const Campaign = createReactClass({
...
+          <CampaignStats campaign={this.props.campaign} />
+          <div className="stats-download-modal">
+            <CampaignOverviewHandler {...this.props} />
+          </div>
+        </div>);
     }
+
+    let navBar;
+  if (window.location.href.match(/ores_plot/) || window.location.href.match(/alerts/)) {
+    navBar = (<CampaignNavbar
```
ragesoss, line 43: Why not render the navbar for the overview tab?

1v4n4: Navbar was rendered only for ores and Alerts, those two tabs are completely in React. For the rest of the tabs it comes from Rails views, this one is for an overview:
navbar.
I tried not to mess with stuff other than stats, but I could try to render React navbar for all tabs

ragesoss: It doesn't need to be for all tabs right now, but I think it should be for the overview tab.

app/assets/javascripts/components/alerts/campaign_alerts.jsx
```diff
@@ -19,6 +19,20 @@ class CampaignAlerts extends React.Component {
   }
 
   componentDidMount() {
+    // This clears Rails parts of the previous pages, when changing Campagn tabs
```
ragesoss, line 22: What happens when you navigate to Alerts and then back to Overview?

ragesoss: If that's a problem, consider hiding things with css (display: none) and then removing that css when they need to be shown again.

1v4n4: That works fine. The only problem was when navigating from the React-Rails tab to React tab. Rails parts just stay under, so I used this to clear them if any. 
Navigating between two Rails-React tabs and React only to Rails-React was ok all the time.

ragesoss:
Okay, I think this is a fine next step. Is there an easy way to fix the layout changes? I'd rather not add the extra space between the overview stats and the server-rendered elements, along with the extra space between the server-rendered elements and the 'Download Stats' button.

1v4n4:
Not at the moment, but I'll think of something by tomorrow.

ragesoss:
Looks good to me. Is this ready to be merged?

1v4n4:
Wait a bit. I'm not happy with these 'Download stats'. This is a solution for none or 3 buttons (create program, edit...) buttons. I realized that it could be 0 to 4 of them, and in that case, this wouldn't fit. I'm trying to calculate distance based on the number of buttons.

1v4n4:
I think it is done now. It was a lot of looking in almost the same layouts (tabs), I may have missed something.

app/assets/javascripts/components/campaign/campaign.jsx
```diff
@@ -29,24 +28,53 @@ export const Campaign = createReactClass({
...
+        );
     }
+
+    if (window.location.href.match(/programs/)) {
+      return (
+        <div>
+          <CampaignNavbar
+            campaign={this.props.campaign}
+          />
+          <header className="main-page">
```
ragesoss, line 52: Putting this stats in a `header` element is the reason there's extra space above and below the stats on this tab. I think if you remove this, all of the tabs will be consistent.

ragesoss:
Aside from the stray `console.log`s and the `header` issue, this looks good to me.

1v4n4:
@ragesoss what about this? 
students-editors

ragesoss:
> @ragesoss what about this? students-editors

It would be nice to fix that. Looks like `overview/_nav.html.haml` handles these correctly (and the server-rendered Students/Editors table does as well, in `users.html.haml`) but all the tab labels in the React navbar are hard-coded English and don't vary based on `campaign.course_string_prefix` like they should.

1v4n4:
> > @ragesoss what about this? students-editors
> 
> It would be nice to fix that. Looks like `overview/_nav.html.haml` handles these correctly (and the server-rendered Students/Editors table does as well, in `users.html.haml`) but all the tab labels in the React navbar are hard-coded English and don't vary based on `campaign.course_string_prefix` like they should.

I just opened navbar and I saw the depth of the problem. I'll fix all of that.

Issue #4345, opened 2021-02-09
Wikidata stats should be download option for an entire campaign


ragesoss:
### Current Behavior:
For courses that are tracking Wikidata, the 'download stats' modal has an option to download Wikidata-specific stats. However, there's no way to download the aggregate Wikidata stats for an entire campaign (except by downloading the Wikidata stats for each course and adding them together yourself).

### Desired Behavior:
The 'Download stats' modal on a campaign Home tab should provide a Wikidata option.

### Related Files & Code:
* `CourseWikidataCsvBuilder`
* `CampaignStatsDownloadModal`
* `CampaignsController`

### Additional context:
Follow the pattern used by other campaign stats data flows: build on top of the course CSV builder to compile data for all the courses in a campaign.

cyrillefr:
Hello @ragesoss , may I work on this one ?

ragesoss:
@1v4n4 would you like to let cyrillefr take this on? (Ivana is working on Wikidata stats for her Outreachy internship, and might otherwise take this on pretty soon.)

1v4n4:
I was thinking to add Wikidata stats to a campaign.json as a next step. I think that would  impact this issue. Otherwise, I don't mind.

ragesoss:
Cool. I think doing it in campaign.json will basically be independent of this task. What impact do you foresee?

1v4n4:
well, i thought about it and it wouldn't. except it will generate the same data as csv

ragesoss:
Great. Go for it, @cyrillefr .

PR #4803, opened 2022-02-24
Download Wikidata stats for a campaign

cyrillefr:
Adresses issue #4345 : Wikidata stats should be download option for an entire campaign

When wiki_education is set to true in application.yml, there is a new link to download wikidata statistics.
There is currently a similar link, but only for courses.
The csv data is an aggregate of the data by courses.

Commit message:
- Add link to wikidata stats download to UI
- Back end code (controller + lib)
- Refactoring CourseWikidataCsvBuilder to use with multiple courses
  in order to also use it with a Campaign.
- Small refactoring in worker(rubocop)
- Add locales in en and fr
- Add route
- Add tests: units, controllers
- Add factory for CourseStat

### DIFF
--- a/app/assets/javascripts/components/campaign/campaign_stats_download_modal.jsx
+++ b/app/assets/javascripts/components/campaign/campaign_stats_download_modal.jsx
@@ -11,6 +11,20 @@ const CampaignStatsDownloadModal = ({ match }) => {
   const editorsLink = `/campaigns/${campaignSlug}/students.csv`;
   const editorsByCourseLink = `/campaigns/${campaignSlug}/students.csv?course=true`;
   const instructorsLink = `/campaigns/${campaignSlug}/instructors.csv?course=true`;
+  const wikidataLink = `/campaigns/${campaignSlug}/wikidata.csv`;
+
+  let wikidataCsvLink;
+  if (Features.wikiEd) {
+    wikidataCsvLink = (
+      <>
+        <hr />
+        <p>
+          <a href={wikidataLink} className="button right">{I18n.t('campaign.data_wikidata')}</a>
+          {I18n.t('campaign.data_wikidata_info')}
+        </p>
+      </>
+    );
+  }
 
   if (!show) {
     return (
@@ -52,6 +66,7 @@ const CampaignStatsDownloadModal = ({ match }) => {
         <a href={instructorsLink} className="button right">{I18n.t('campaign.data_instructors')}</a>
         {I18n.t('campaign.data_instructors_info')}
       </p>
+      {wikidataCsvLink}
     </div>
   );
 };
--- a/app/assets/javascripts/components/campaign/campaign_stats_download_modal.jsx
+++ b/app/assets/javascripts/components/campaign/campaign_stats_download_modal.jsx
@@ -13,18 +13,6 @@ const CampaignStatsDownloadModal = ({ match }) => {
   const instructorsLink = `/campaigns/${campaignSlug}/instructors.csv?course=true`;
   const wikidataLink = `/campaigns/${campaignSlug}/wikidata.csv`;
 
-  let wikidataCsvLink;
-  if (Features.wikiEd) {
-    wikidataCsvLink = (
-      <>
-        <hr />
-        <p>
-          <a href={wikidataLink} className="button right">{I18n.t('campaign.data_wikidata')}</a>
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/components/campaign/campaign_stats_download_modal.jsx
```diff
@@ -11,6 +11,20 @@ const CampaignStatsDownloadModal = ({ match }) => {
   const editorsLink = `/campaigns/${campaignSlug}/students.csv`;
   const editorsByCourseLink = `/campaigns/${campaignSlug}/students.csv?course=true`;
   const instructorsLink = `/campaigns/${campaignSlug}/instructors.csv?course=true`;
+  const wikidataLink = `/campaigns/${campaignSlug}/wikidata.csv`;
+
+  let wikidataCsvLink;
+  if (Features.wikiEd) {
```
ragesoss, line 17: There's no need to limit this based on `Features.wikiEd`.

lib/analytics/campaign_csv_builder.rb
```diff
@@ -44,6 +45,11 @@ def revisions_to_csv
     CSV.generate { |csv| csv_data.each { |line| csv << line } }
   end
 
+  def wikidata_to_csv
+    courses = @campaign.courses.joins(:course_stat)
+    CourseWikidataCsvBuilder.new(courses).generate_csv
```
ragesoss, line 50: I'd prefer this to follow the same pattern as the other methods like `articles_to_csv`, building an aggregate CSV by looping through the courses in the campaign, and adding a row of wikidata stats for each course.

Making `CourseWikidataCsvBuilder` handle the cases of one as well as multiple courses will make this harder to reason about and change, I think.

The problem with that approach is that, unlike the other CSV builders, the output for the Wikidata one isn't a format that is suitable to that aggregation strategy. I think the best way around that will be to change the output format of the Wikidata CSV builder. Currently, it uses headers of `revision_type, count` with one row per revision type. The alternative would be to have a column for each revision type, so that the campaign version could have one row per course.

cyrillefr: I went with the join so I do not have to loop, but I can change to the `articles_to_csv` pattern.
I also will make 2 methods: one `CourseWikidataCsvBuilder` for courses and another one for campaign.

So, if I understand well, your format proposal for campaign is:
claims created, claims changed, claims removed, etc. course // header
4,2,1, xx,x,"This course name"  // lines of data for one course of a campaign
2,4,5, xx, xxx "This other course name" // same

Also, do you want me to change `CourseWikidataCsvBuilder` so that it will be the same format, but with just one line ?

app/assets/javascripts/components/campaign/campaign_stats_download_modal.jsx
cyrillefr: I did that, because there is this limit in the course version (`app/assets/javascripts/components/overview/course_stats_download_modal.jsx`
So I implied it would be the same for a campaign. I will remove it.

lib/analytics/campaign_csv_builder.rb
ragesoss: Yes, that's correct, except that `course name` should be the first column.

app/assets/javascripts/components/campaign/campaign_stats_download_modal.jsx
ragesoss: Ah! We should change that limitation on CourseStatsDownloadModal... now it should check for whether the `course` object has a `course_stats` property. @1v4n4's project made the wikidata stats data available without the Wiki Ed restriction.

lib/analytics/campaign_csv_builder.rb
```diff
@@ -44,6 +45,25 @@ def revisions_to_csv
...
+    end
+
+    csv_data << sum_wiki_columns(csv_data) if courses.any?
+
+    CSV.generate { |csv| csv_data.each { |line| csv << line } }
+  end
+
+  def sum_wiki_columns(csv_data)
+    # Skip 1st header row + 1st column course name
+    csv_data[1..].transpose[1..].map(&:sum).unshift('Total')
```
ragesoss, line 64: This is clever. We don't include a `total` line in any of the other CSVs, but I'm okay with it.

Issue #4685, opened 2021-11-29
Filter settings in Articles Edited view of course page should be reflected in URL


ragesoss:
The Articles Edited list can be filtered by wiki, new vs existing article, and tracked vs untracked.

It would be useful if these filter settings (if not default) were reflected in URL parameters, so that users could share a link to the exact filtered view they see. (ie, https://commons.wikimedia.org/wiki/File:Screenshot_Programs_Events_Dashboard.png)

See https://meta.wikimedia.org/w/index.php?title=Talk:Programs_%26_Events_Dashboard&oldid=22391543#adding_unique_URL_for_each_view_in_statistics

TheTrio:
Was thinking about trying this. How should I import the articles data so I could try this out locally? Is [this](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/user_roles.md#copy-a-course-from-production) enough?

ragesoss:
Yes, that is probably a good way to do it.

PR #4804, opened 2022-02-24
added URL feedback on filter change

TheTrio:
## What this PR does

Addresses #4685. It introduces 3 new search query params. 

| Key | Possible Values | Default |
| --------------- | --------------- | --------------- |
| wiki | `all` and other wikis(like `en.wikipedia`) | `all` |
| tracked | `tracked`, `untracked` and `both` | `tracked` |
| newness | `new`, `existing` and  `both` | `both` |

For example, the URL for this 
image

will be `/courses/Southwestern_University/Biochemistry_(Fall)/articles/edited?newness=existing&tracked=tracked&wiki=en.wikipedia`

As the user changes the filters, the URL is updated automatically, making it easier to share specific views.

Currently, the filters are not rendered if there are only 1 type of edits. For example, if there exists only one wiki, or if there are no untracked edits, their respective filters are not visible.

However, the search params don't have any such restriction.If you add `&tracked=untracked` to a course without any untracked edits, it will show an empty list. Since the filters are not visible, the user won't have any way to change the filter(other than updating the URL manually of course)

I'm not sure if I should bother with this particular case, since the user will have to manually edit the URL to get an empty view in the first place.


ragesoss:
Cool! This looks reasonable. I'll test it out once the build is passing.

TheTrio:
As discussed, have deleted the test file. Will try extending the tests in `spec/features/course_page_spec.rb` to check for the search params in another PR.

ragesoss:
This works fine, and I can merge it. One problem I noticed while testing it out is that filter state is persisted if you navigate to a different tab and then navigate back, but when you navigate back, the filter dropdowns don't reflect the filter state; they show the default value even though the articles list is being filtered differently. I'll open a separate issue for that.

ragesoss:
https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/4805

Issue #4805, opened 2022-02-24
In Articles Edited view, filter settings don't match UI after navigating away and back


ragesoss:
### What is happening?
The Articles Edited view on a course page — eg, https://dashboard.wikiedu.org/courses/Brigham_Young_University/The_United_States_Since_1877_(Winter_2021)/articles/edited — lets you filter the list. (There are up to three filter parameters, but just one in that example.) If you change one of the default settings (for example, filter the example to only show `New` articles), then you navigate to a different tab of the course, then back to the Articles tab, the filter selector shows the default filter values (`New and existing`) even though the list is still filtered with the previously-set parameters (`New`).

### Expected behavior
When you navigate away and then return to the Articles tab, the filter selectors should reflect the currently-active filters.

### Additional context
This probably needs to be solved by pushing more state into Redux.

TheTrio:
Am working on this.

PR #4807, opened 2022-02-25
Made the filter selectors in sync with the redux store

TheTrio:
## What this PR does

Addresses #4805 

For convenience, I have changed the default value of `wikiFilter` and `newnessFilter` to the default values in the drop downs instead of `null`. For example, `newnessFilter` now has a default value of `both` since that matches with the dropdown and thus makes it easier to map. 

For `wikiFilter`, the new default value is `{project:"all"}`. This is also mostly for convenience - it means I don't have to have another if condition to check whether its `null`

Now if the user visits `/courses/Southwestern_University/Biochemistry_(Fall)/articles/edited`, the url is changed to `/courses/Southwestern_University/Biochemistry_(Fall)/articles/edited?newness=both&tracked=tracked&wiki=all` since those are the default values of the redux store. 

If the user changes a filter and goes to another tab, and comes back, the URL is changed appropriately. 

gif

### DIFF
--- a/app/assets/javascripts/components/articles/article_list.jsx
+++ b/app/assets/javascripts/components/articles/article_list.jsx
@@ -12,6 +12,8 @@ import articleListKeys from './article_list_keys';
 import ArticleUtils from '../../utils/article_utils.js';
 import { parse, stringify } from 'query-string';
 
+const defaults_params = { wiki: 'all', tracked: 'tracked', newness: 'both' };
+
 const ArticleList = createReactClass({
   displayName: 'ArticleList',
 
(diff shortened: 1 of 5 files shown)


ragesoss:
This looks really good. Would it be difficult to skip adding default URL params? Like, I'd prefer the params not to appear if a user is navigating from tab to tab without interacting with the filters.

TheTrio:
I guess I could just add some if checks and only if the value is different from the default value does it add it to the URL. How does that sound?

ragesoss:
Yes, that sounds like the ideal behavior. Unnecessary parameters in the URLs can make an app feel cluttered, in my opinion.

TheTrio:
This should be enough I guess.

ragesoss:
Nice! This works perfectly.

PR #4811, opened 2022-02-27
Fixed urls and errors on settings page

TheTrio:
## What this PR does

This PR does multiple things. 

~~Firstly, it converts the ajax calls with the fetch API. This wasn't required really, but I was in the middle of doing exactly this when I found the other bugs  so :shrug:~~ Nvm ajax was being mocked in the tests. Prolly best to do this later lol

Secondly, it fixes the URLs of the requests, which were nonexistent, resulting in the following error 

image

And thirdly, I've removed the undefined property access. It occurs when the user is not a super admin and tries to add an admin, special user, etc. 

ksnip_20220228-004835

Now, as is expected, a notification pops up, showing the error message. 

ksnip_20220228-005211

### DIFF
--- a/app/assets/javascripts/actions/settings_actions.js
+++ b/app/assets/javascripts/actions/settings_actions.js
@@ -152,10 +152,12 @@ export const upgradeSpecialUser = (username, position) => (dispatch) => {
           submitting: false
         },
       });
-
+      if (response.responseJSON === undefined) {
+        response.responseJSON = { message: response.responseText };
+      }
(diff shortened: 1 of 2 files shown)


TheTrio:
So I believe the problem was that when the request was unauthorized, the server sent back just plain text. However, when some other sort of error occurred - like the user not existing - it sent back JSON. The error notifications should now work on both these cases.

TheTrio:
The error should now show the username of the missing user 

image

Issue #4684, opened 2021-11-29
Replace current usage of jquery and remove it as a dependency


ragesoss:
We have a number of places, both in React views and server-rendered HTML views, that use jquery in various ways.

Both to shrink the bundle size and to make it easier to understand the frontend code without being familiar with jquery conventions, it would be nice to remove the jquery dependency.

Possibly among other things, this will require:
* Replacing the many cases in React action files and `api.js` that use jquery `ajax` to make requests
* Re-implementing the current server-rendered list views that use jquery and the `list.js` package to provide table sorting functionality. (See the `utils/course.js` file.) These views could be converted to React components, or (maybe) the sorting functionality could be rewritten to bypass the use of jquery.

TheTrio:
Am working on this

kunalthedev:
I guess only `Survey.js` and `SurveyAdmin.js` are left

sylvia-boyani:
Hey working on this.
Removing jquery from the remaining files, and finally remove it entirely.

sunkireddy-Barath:
Hi @ragesoss , i would like to work on this issue, Could you please assign this task to me?

ragesoss:
@sunkireddy-Barath we don't assign issues, but you're welcome to work on it. I strong recommend working in small pieces.

sunkireddy-Barath:
Thank you for the clarification. I understand that issues are not assigned
in this repository. I will look for a suitable "good first issue" and start
working on it, focusing on small changes as you suggested. I will open a PR
once I complete the changes.

Thanks for the guidance 😊

SUNKIREDDY BARATH,
Student @ CIT,
Chennai-69

GonzaFP:
@ragesoss Since this issue is still open, I wonder what files are still using ajax. I see the ones mentioned have been worked on.
I also see @Sahethi worked on the sorting functionality.

Formasitchijoh:
Hi @GonzaFP  here are some of the files still using jQuery 

`utils/editable.js `— a jQuery plugin that turns text blocks into inputs
`/app/assets/javascripts/surveys/modules/Survey.js` — uses a jQuery object for slider manipulation
`/app/assets/javascripts/surveys/modules/SurveyAdmin.js `— imports jquery-ui sortable and tabs widgets

GonzaFP:
@Formasitchijoh  Thank you. I will work on this after #5486

PR #4818, opened 2022-03-01
Convert ajax to fetch

TheTrio:
## What this PR does

Addresses #4684 

The coverage report indicates that there are currently 12 requests which haven't been tested. 

1. [`fetchWikidataLabelsPromise`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4818/files#diff-7db6e393e6f74dbb08ce544acb55415ad34735fc73d58a5da18b9ba53fee4886)
2. [`article_graphs.jsx`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4818/files#diff-4fd0bc428b65148d07c6dbae3f39b909dcb827e52951dd7e4a2685cb0cd0f3c0)
3. [`updateSalesforceCredentialsPromise`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4818/files#diff-7ced3f635bf09bb5c531ecd01903b936ca5143a641f9dcfed64f8ab520940354)
4. [`grantAdminPromise`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4818/files#diff-7ced3f635bf09bb5c531ecd01903b936ca5143a641f9dcfed64f8ab520940354)
5. `createCustomFeedback`, `destroyCustomFeedback`, `greetStudents`, `dismissNotification`, `createBadWorkAlert`, `requestNewAccount`, `enableAccountRequests`, and `updateSalesforceRecord` all in [`api.js`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4818/commits/36d0df0f52bae1750ff88e5414879ea57e7cd6d7#diff-fe52ddf568c7cec6e7f56bc61c464f5b923852485bb723f0bb3fc3328c06e34e)

Apart from these, I haven't been able to test whether the `SentryLogger` works as expected. Its used in 2 places - `saveTimeline` and `saveCourse`, both of them in `api.js`

### DIFF
--- a/app/assets/javascripts/actions/wikidata_actions.js
+++ b/app/assets/javascripts/actions/wikidata_actions.js
@@ -2,28 +2,27 @@ import { chunk, map, join, filter } from 'lodash-es';
 import * as types from '../constants';
 import logErrorMessage from '../utils/log_error_message';
 import CourseUtils from '../utils/course_utils';
+import request from '../utils/request';
+
 
 const wikidataApiBase = 'https://www.wikidata.org/w/api.php?action=wbgetentities&format=json&origin=*';
 
-const fetchWikidataLabelsPromise = (qNumbers) => {
+const fetchWikidataLabelsPromise = async (qNumbers) => {
   const idsParam = join(qNumbers, '|');
-  return new Promise((res, rej) => {
-    return $.ajax({
-      url: wikidataApiBase,
-      data: {
-        ids: idsParam,
-        props: 'labels',
-        languages: `${I18n.locale}|en`
-      },
-      success: (data) => {
-        return res(data);
-      },
+  const response = await request(wikidataApiBase, {
+    method: 'POST',
+    body: JSON.stringify({
+      ids: idsParam,
+      props: 'labels',
+      languages: `${I18n.locale}|en`
     })
-      .fail((obj) => {
-        logErrorMessage(obj);
-        return rej(obj);
-      });
   });
+  if (!response.ok) {
+    logErrorMessage(response);
+    const data = await response.json();
+    throw data;
+  }
+  return response.json();
 };
 
 // This takes a Wikidata page title and checks whether it looks like
--- a/app/assets/javascripts/actions/wikidata_actions.js
+++ b/app/assets/javascripts/actions/wikidata_actions.js
@@ -3,20 +3,18 @@ import * as types from '../constants';
 import logErrorMessage from '../utils/log_error_message';
 import CourseUtils from '../utils/course_utils';
 import request from '../utils/request';
-
+import { stringify } from 'query-string';
 
 const wikidataApiBase = 'https://www.wikidata.org/w/api.php?action=wbgetentities&format=json&origin=*';
 
 const fetchWikidataLabelsPromise = async (qNumbers) => {
   const idsParam = join(qNumbers, '|');
-  const response = await request(wikidataApiBase, {
-    method: 'POST',
-    body: JSON.stringify({
-      ids: idsParam,
-      props: 'labels',
-      languages: `${I18n.locale}|en`
-    })
-  });
+  const query = {
+    ids: idsParam,
+    props: 'labels',
+    languages: `${I18n.locale}|en`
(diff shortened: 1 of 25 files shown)


TheTrio:
Not sure why that failed. I haven't changed anything related to syllabus upload. In fact, when I run the test locally it works fine 
image

TheTrio:
@ragesoss I think that covers everything for this PR

ragesoss:
Were you able to manually test all the ones that aren't covered in the rspec suite?

I'm not worried about the SentryLogger part. The actual network calls related to that are all handled by the Sentry SDK (via calls to Sentry.captureMessage) and the SentryLogger object is just used to hold data that we want to log. So as long as the requests that use Sentry are themselves working, I expect the Sentry part to be fine. (Breaking those would also be very low impact.)

TheTrio:
I've tested `updateSalesforceCredentialsPromise` and `updateSalesforceCredentialsPromise` and they seem to work fine. As for others, like the feedback one, they are visible only if certain conditions are satisfied, so testing them will require a bit more effort. Haven't gotten around to doing that yet.

app/assets/javascripts/actions/wikidata_actions.js
```diff
@@ -2,28 +2,27 @@ import { chunk, map, join, filter } from 'lodash-es';
 import * as types from '../constants';
 import logErrorMessage from '../utils/log_error_message';
 import CourseUtils from '../utils/course_utils';
+import request from '../utils/request';
+
 
 const wikidataApiBase = 'https://www.wikidata.org/w/api.php?action=wbgetentities&format=json&origin=*';
 
-const fetchWikidataLabelsPromise = (qNumbers) => {
+const fetchWikidataLabelsPromise = async (qNumbers) => {
```
ragesoss, line 10: This appears to be broken. The `body` params are not being included.

I wrote a test that uses this (and breaks with your change): https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/b9866a6071a842e10574c706364545ae9c064ba7

TheTrio: Should be fixed now

TheTrio:
I think I have found another problem. 

Currently, the `request` function adds a `X-CSRF-Token` only if the path is relative. However, it defines `relative` to mean not satisfying the regex `/http(s?)/g`. This works fine, until you have a request which has the word `http` or `https` in it. 

For example, the delete assignment request has a query parameter `sandboxUrl` which is set to `https%3A%2F%2Fwww.wikidata.org%2Fwiki%2FUser%3ADumbPotato21%2FQ16060473` or whatever the URL of the user is. 

image

This results in an unauthorized error 

image

To fix this, I've redefined it to be a relative path only if the URL doesn't _**start**_ with `http` or `https`. Hopefully this doesn't break anything else although I can't see how it would.

ragesoss:
nice catch. how did you discover that problem?

TheTrio:
Had no idea how to reproduce that wiki data bug so was just looking around, pressing buttons randomly. Found that I couldn't remove an item once I assigned it to a user. It was working on master so I knew it was a change I made that borked it. After that, it was just comparing the requests made by this branch and master.

TheTrio:
I don't think this is related. It runs fine locally. 

image

ragesoss:
@TheTrio Now that this is live, I've noticed a problem I missed before... when there is a server error, it returns HTML even for a json request. This used to result in an error notification like "Internal Server Error", but now instead it shows a JSON parsing error (because there's another error in the JS now when it tries to parse the HTML response as JSON).

For example, try adding `raise StandardError` to the first line of `CoursesController#update`, then editing the Description of a course and saving it.

Want to take a shot at fixing that? Any of the routes could potentially return an HTML error response.

TheTrio:
If I have understood you correctly, all I would need to do is check if the response from the server is valid json. If it isn't, just show a generic error message(ie, internal server error)

Is that right?

TheTrio:
Yeah that makes a lot more sense. I'll try this tomorrow.

PR #4825, opened 2022-03-08
To integrate unused babel dependency

Sahethi:
## What this PR does
In this PR, I have included babel as dependency cause it was unused and created a `src` folder with `example.js` to transpile and create the output in `app/example.js`

Hey, for the current GSoC project, I was planning to include the unused babel dependency for JS and JSX files. Most of the `.js` files in `app` uses ES6, hence thought this would be a good task to work on in order to support browsers which support ES5.  
Please do let me know if I should proceed to do so further for the files in `app` and use `babel/preset-react` for react components as well, or drop the unused dependency, cause i have found quite a few.


ragesoss:
The project does use babel already; it's configured via `babel.config.js` and is used by the Webpack compilation process. (If you rename that babel config file, then run `yarn start`, it will fail to compile when it hits JSX syntax.)

Sahethi:
Ohh yes, I just tried to do that and my code did fail to compile. So I think the only necessary thing was updating the dependencies.
Do you think it would be a better task to focus on upgrading Webpack here?

ragesoss:
Upgrading babel and the plugins to the latest versions is generally useful, and sometimes necessary to complete other upgrades. I'm not sure whether babel is currently blocking anything, though. My guess is that it's not a blocker for the Webpack upgrade. But I'd be happy to merge a PR that just upgrades the babel packages.

Sahethi:
Hey, that sounds good, I'll get rid of the `example.js` files and the `src` folder and submit a PR with updated babel packages.

Sahethi:
Hey also, `@babel/polyfill` package is deprecated, does that need to be dealt with as well?

ragesoss:
I expect so. Why was it deprecated and what advice have the maintainers provided related to the deprecation?

Sahethi:
So, I checked the babel [website](https://babeljs.io/docs/en/babel-polyfill.html) and it states 

> As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions):
`import "core-js/stable";`
`import "regenerator-runtime/runtime";`

Sahethi:
> So, I checked the babel [website](https://babeljs.io/docs/en/babel-polyfill.html) and it states
> 
> > As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions):
> > `import "core-js/stable";`
> > `import "regenerator-runtime/runtime";`

Hence according to the docs, I think one needs to import the individual polyfills manually. Can this be a potential issue to work on?

ragesoss:
Yes. I'm not sure where those would need to be imported or whether they are necessary, but I look forward to hearing about it from you!

Sahethi:
> Yes. I'm not sure where those would need to be imported or whether they are necessary, but I look forward to hearing about it from you!

Sure thing! Thank you!

PR #4829, opened 2022-03-08
Updating babel packages

Sahethi:
## What this PR does
Updated the babel dependencies to the latest versions. As mentioned in #4825


ragesoss:
This needs to also include the corresponding changing to `yarn.lock`. (Those will happen if you run `yarn` after changing `package.json`.)

Sahethi:
@ragesoss Hey I updated `yarn.lock`. I actually was getting errors cause I was directly running `yarn start` and it got resolved by running `yarn` causing the respective changes.Thanks for letting me know!

PR #4830, opened 2022-03-09
Fixed JSON parsing error

TheTrio:
If `response.ok` is `false`, instead of trying to parse the response as JSON, the promise functions now hand over that job to `handleErrorNotification` in `notifications.js`

Before: 

image

After: 

image

### DIFF
--- a/app/assets/javascripts/actions/campaign_actions.js
+++ b/app/assets/javascripts/actions/campaign_actions.js
@@ -6,8 +6,9 @@ const fetchCampaignsPromise = async (courseId) => {
   const response = await request(`/courses/${courseId}/campaigns.json`);
   if (!response.ok) {
     logErrorMessage(response);
-    const data = await response.json();
-    throw data;
+    const data = await response.text();
+      response.responseText = data;
+      throw response;
   }
   return response.json();
 };
@@ -34,8 +35,9 @@ const removeCampaignsPromise = async (courseId, campaignId) => {
   });
   if (!response.ok) {
     logErrorMessage(response);
-    const data = await response.json();
-    throw data;
+    const data = await response.text();
+    response.responseText = data;
+    throw response;
   }
   return response.json();
 };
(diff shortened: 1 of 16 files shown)


app/assets/javascripts/actions/campaign_actions.js
```diff
@@ -6,8 +6,9 @@ const fetchCampaignsPromise = async (courseId) => {
   const response = await request(`/courses/${courseId}/campaigns.json`);
   if (!response.ok) {
     logErrorMessage(response);
-    const data = await response.json();
-    throw data;
+    const data = await response.text();
```
ragesoss, line 9: What happens if the response to one of these is actually a JSON error? Will it still correctly handle it as JSON in notifications.js?

TheTrio: Yup. For example, go to `/settings` and try to add a non existent user as an admin. This requests returns a JSON object with a key `message`. The notification should still show up correctly, since `handleErrorNotification` does try to parse the json and check if a message property exists.

ragesoss: Does `.text()` retain the other parts of the response that `notifications.js` has code for handling? It's designed to look at `responseJSON.error`, `responseText.message`, `errors`, `message`, or `statusText`, so that it can find the best error message to show regardless of what kind of error response it receives. (This is important because the system handles most expected errors by returning JSON responses, but unexpected and unhandled server errors may return Rails error HTML or even an error straight from Puma/Passenger/Nginx/Apache, depending on just what went wrong.)

TheTrio: I believe `responseJSON` was a property added by `ajax` (it converted the response to JSON automatically) so that doesn't really exist now. If there are routes which don't send an error back with the `message` property, then yes the current method would simply fall back to the `statusText`, ignoring others. I think one way to handle this would be to add a `responseJSON` property in  `handleErrorNotification` itself, if it is indeed valid json. This way `.error` would be preserved. 

As for the others, `responseText.message` doesn't exist since `responseText` is a string. The function checks for the message property by first parsing it into JSON though, so I don't think that's a problem. As for `data.errors`, I'm not sure. I guess in this case it expects `data` to be an object, with an `errors` property. Currently however, we're always sending text. One way to handle this would be to check for `.errors` after parsing it into JSON. 

Although I would mention that in the places where `request` was already being used, only the `statusText` was being sent to the dispatcher. For example, see `fetchFeedback` in `api.js`

PR #4841, opened 2022-03-10
Updated eslint configuration

TheTrio:
# What it does

This PR attempts to remove outdated and deprecated plugins used for `eslint` Firstly, it removes [`eslint-loader`](https://www.npmjs.com/package/eslint-loader), which hasn't been updated for the last two years. The obvious alternative was [`eslint-webpack-plugin`](https://webpack.js.org/plugins/eslint-webpack-plugin/). 

However, the latest version of that package doesn't support webpack 4. For the time being therefore, I've gone with version 2. When we do upgrade to webpack 5, transitioning to version 3 of the plugin shouldn't be a major issue. 

Another old package was [`babel-eslint`](https://www.npmjs.com/package/babel-eslint), which is a custom parser we use to enable using `import` and `export` everywhere in a file(see `/app/assets/javascripts/main.js`) This is to allow lazy loading of modules, reducing bundle size. The default eslint parser [doesn't allow this](https://stackoverflow.com/questions/39158552/ignore-eslint-error-import-and-export-may-only-appear-at-the-top-level)

Since this package was also not maintained anymore, [`@babel/eslint-parser`](https://www.npmjs.com/package/@babel/eslint-parser) was chosen. 

I've also added the new packages as dev dependencies. The earlier ones were dependencies. I'm not sure if this was done on purpose - if it was, I will add them back as normal dependencies. 

As for the configuration, it is mostly the same as before. One thing to note is that as far as I can tell, `eslint-webpack-plugin` doesn't have an option for the specifying parser like `eslint-loader` did. So that configuration has now been moved to the `.eslintrc` file

One side effect of this is that the editor no longer complains about the imports in `/app/assets/javascripts/main.js`

Another change is that while `eslint-loader` stored the cache files in `./node_modules/.cache/eslint-loader`, the webpack plugin does so in `./.eslintcache`. This can obviously be changed, if needed.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -96,6 +96,7 @@
     "jquery": true,
     "browser": true
   },
+  "parser": "@babel/eslint-parser",
   "settings": {
     "import/resolver": {
       "babel-plugin-root-import": [
--- a/js.webpack.js
+++ b/js.webpack.js
@@ -2,6 +2,7 @@ const path = require('path');
 const config = require('./config');
 const ManifestPlugin = require('webpack-manifest-plugin');
 const LodashModuleReplacementPlugin = require('lodash-webpack-plugin');
+const ESLintPlugin = require('eslint-webpack-plugin');
 const MomentLocalesPlugin = require('moment-locales-webpack-plugin');
 const webpack = require('webpack');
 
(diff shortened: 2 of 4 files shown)


ragesoss:
Can you add some more explanation about what these configuration changes mean?

A narrative about the things you tried to get to this and the problems you ran into along the way — along with links to any documentation that you used — would also be appropriate.

TheTrio:
Edited.

ragesoss:
That's excellent, thanks!

PR #4850, opened 2022-03-11
Add tests for filters

TheTrio:
# What this PR does

#4804 added URL feedback for the various filters in the Articles tab. This PR adds some tests for the same. 

I'm wasn't sure what was the best way to organize the tests so I have put everything under just one describe.

### DIFF
--- a/spec/features/articles_edited_spec.rb
+++ b/spec/features/articles_edited_spec.rb
@@ -29,7 +29,7 @@ describe 'Articles Edited view', type: :feature, js: true do
     create(:revision, article: article2, user: user, date: '2018-01-01')
     create(:revision, article: article3, user: user, date: '2020-04-01')
     create(:revision, article: article4, user: user, date: '2020-03-01')
-    create(:revision, article: article5, user: user, date: '202-05-01')
+    create(:revision, article: article5, user: user, date: '2020-05-01')
 
     create(:articles_course, course: course, article: article1, user_ids: [user.id])
     create(:articles_course, course: course, article: article2, user_ids: [user.id], new_article: true)
--- a/spec/features/articles_edited_spec.rb
+++ b/spec/features/articles_edited_spec.rb
@@ -1,23 +1,132 @@
 # frozen_string_literal: true
-
 require 'rails_helper'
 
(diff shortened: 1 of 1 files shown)


spec/features/articles_edited_spec.rb
```diff
@@ -1,23 +1,132 @@
...
+  let(:es_wiktionary) { create(:wiki, language: 'es', project: 'wiktionary') }
+  let(:fr_wikidata) { create(:wiki, language: 'fr', project: 'wikidata') }
+  
   before do
+    stub_wiki_validation
+    course = create(:course,
+      slug: slug,
+      start: '2017-01-01',
+      end: '2021-01-01',
+      home_wiki_id: home_wiki.id
```
ragesoss, line 18: In most cases, you can and should be a little more concise in the tests by passing the object instead of the idea for `create`. In this case, `home_wiki: home_wiki`. The same goes for the `wiki` of each article below.

TheTrio: Thanks! Have changed this

TheTrio:
Do you think these tests are enough or should I test all possible values for each filter?

spec/features/articles_edited_spec.rb
```diff
@@ -1,23 +1,164 @@
...
 
+slug = 'foo/bar'
+
 describe 'Articles Edited view', type: :feature, js: true do
-  let(:course) { create(:course, start: '2017-01-01', end: '2020-01-01') }
   let(:user) { create(:user, username: 'Ragesoss') }
-  let(:article) { create(:article, title: 'Nancy_Tuana') }
+  let(:home_wiki) { Wiki.get_or_create language: 'en', project: 'wikipedia' }
+  let(:es_wiktionary) { create(:wiki, language: 'es', project: 'wiktionary') }
+  let(:fr_wikidata) { create(:wiki, language: 'fr', project: 'wikidata') }
```
ragesoss, line 10: There is no `fr.wikidata`. Wikidata is multilingual project, and within the dashboard, the language attribute for it will be `null`. See `wiki.rb#81`.

ragesoss:
Much better!

PR #4869, opened 2022-03-19
Update react router to v6

TheTrio:
# What this PR does

This PR attempts to update the codebase to use v6 of React Router

The [official migration guide](https://reactrouter.com/docs/en/v6/upgrading/v5) is pretty good and I would recommend reading it. However, I will mention some changes that I've had to make which were not part of the migration guide. 

## Absolute vs Relative

Currently, we use absolute paths everywhere. For example, see `app/assets/javascripts/components/course/course.jsx` 

```jsx
<Switch>
    <Route exact path="/courses/:course_school/:course_title/overview" .../>
    ...
</Switch>
```

Since v6 however, [Routers use **only** relative paths](https://github.com/remix-run/react-router/issues/8035). Since this `course.jsx` is inside of the another router(see `app/assets/javascripts/components/util/routes.jsx`)

```jsx
<Switch>
    <Route path="/courses/:course_school/:course_title" >
        <Course />
    </Route>
</Switch>
```

Thus, since v6, the all Routes inside `Course` will become relative to the path `/courses/:course_school/:course_title` This is why the new path is just `"overview"` instead of `"/courses/:course_school/:course_title/overview"`

```jsx
<Switch>
    <Route path="overview" .../>
    ...
</Switch>
```

## Hooks

React Router v6 only supports using hooks to get the route information. The `withRouter` function which enabled class based components to use routers before v6 has been removed. Fortunately, making a [wrapper component is pretty straightforward](https://reactrouter.com/docs/en/v6/faq#what-happened-to-withrouter-i-need-it). 

There still are some problems though, although they are probably manageable. 

1. `useNavigate` will give a warning if you use it outside `useEffect`. Since class based components don't have `useEffect`, we will get this warning until we convert those components to functional ones.
2. Navigating the virtual DOM(via the React dev tools extension for example) will be harder because there will be dummy components created by `withRouter` which essentially do nothing. 
3. Tests will no longer be able to set state manually. This is because the component is a functional one now(because of the wrapper) and there's no `setState` function for it. Fortunately, there's only one test I found which did this and removing that call didn't affect the outcome, so all good. Although, as the documentation for `setState` notes, this function should be [used sparingly anyway](https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/setState.html#setstatenextstate-callback--self)

## Exact Routes

Since v6, all routes are exact. That means a `<Route path="/foo" .../>` will **NOT** match `/foo/bar`. We used this functionality in a number of places, and those routes now need to be changed to `/foo/*` to be "not exact"

We also had a number of exact routes(`<Route exact path="/foo" .../>`). The exact prop for them has now been removed since it is the default. 

## Standalone Routes

Since v6, all Routes **MUST** be inside of a `<Routes>` component. Currently, we have the `Nav` component(see `app/assets/javascripts/components/nav/nav.jsx`) however, which uses location but is not inside a Router. In fact, its not even inside the main react root. To fix this, I've simply replaced React Router's location with the `window.location` global. Since the Nav component's use of location is fairly simple(it just reads the current URL), this shouldn't be a problem. 

## Browser Router

Since the `Nav` component doesn't use location anymore, we can also remove the custom implementation of History we used before alongside it([which the docs say we shouldn't use unless necessary](https://reactrouterdotcom.fly.dev/docs/en/v6/api#router)). Instead, I have switched to using the builtin `BrowserRouter`.

## Mutable history

Since v6, the mutable history object is gone. This means that the hooks work as expected, rerendering the component when the URL changes. However, I couldn't get this to work, probably because the component was class based. Therefore, I have switched to using the global `window.location` (See `app/assets/javascripts/components/articles/article_list.jsx#102`)

Again, this shouldn't be an issue, but probably an indication that we should convert this component to a functional one as well. 

I'm probably missing a lot of stuff but I these are the things which I was stuck at.


app/assets/javascripts/components/course/course.jsx
```diff
@@ -127,24 +127,55 @@ export const Course = createReactClass({
         <div className="course_main container">
           <Confirm />
           {enrollCard}
-          <Switch>
-            <Route exact path="/courses/:course_school/:course_title" render={() => <OverviewHandler {...courseProps} />} />
-            <Route exact path="/courses/:course_school/:course_title/home" render={() => <OverviewHandler {...courseProps} />} />
+          <Routes>
+            <Route
```
ragesoss, line 131: I think this will be easier to read if each Route is on a single line, like before.

app/assets/javascripts/components/util/history.js
```diff
@@ -1,26 +0,0 @@
-import { createBrowserHistory } from 'history';
-
-// Custom history component handles scroll position for back button, hashes,
```
ragesoss, line 3: Did you find the situations where this was relevant, and test whether scroll position is still handled well during navigation without this scrollIntoView trigger?

TheTrio: ah, I didn't actually see that. I'll check.

ragesoss:
This break browser 'back' navigation in several cases. It looks like it's related to cases where the route gets rewritten when the component loads. For example, if you start from the Home tab of a course then navigate to `students`, the route immediately gets rewritten to `students/overview`. Clicking 'back' from there appears to just reload the Students overview, rather than taking you back to the Home tab. I guess this is because it goes back to `students` and then immediately routes you to `students/overview` again.

The Articles and Activity tabs behave similarly.

ragesoss:
Cool. That fixed it.

app/assets/javascripts/components/util/withRouter.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import {
+  useLocation,
+  useNavigate,
+  useParams,
+} from 'react-router-dom';
+
+function withRouter(Component) {
```
ragesoss, line 8: A comment explaining the reason for this new wrapper component, and when (if ever) it should be used with new components, would be helpful. Is this just something we need for wrapping class components, and can get rid when wrapped components get converted to functional components?

TheTrio: Done!

PR #4880, opened 2022-03-23
Sorting functionality rewritten in utils/courses.js to bypass the use of jQuery. 

Sahethi:
## What this PR does?
This PR addresses issue #4684.

The aim is to further facilitate the removal of jQuery as a dependency. Here the file `utils/courses.js` extensively uses jQuery but vanilla javascript can perform the same functionalities. 

The changes were made by checking out the resources like [You Might Not Need jQuery](https://youmightnotneedjquery.com/) and [You Don't Need jQuery](https://github.com/nefe/You-Dont-Need-jQuery).

- `$() => {}` binds the function to be executed when DOM has finished loading has been replaced with `document.onreadystatechange() => {}`, here `onreadystatechange` event is fired on a HTML document when the load state of the page's content has changed
- `$('selector')` is replaced by either  `document.querySelector` or `document.querySelectorAll`. 
- But it should be noted that `document.querySelectorAll` returns all matched elements as NodeList. Hence in cases where `addEventListener` is added `forEach` has been implemented for the same because one has to loop through the same in order to add the listener. 
- Another change would be in `navigator.platform` this is a deprecated feature and has been replaced by `navigator.userAgentData.platform` in accordance with the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform).

## Element which is affected
In order to access the courses table, one has to make `wiki-education: 'true'` in `application.yml` and then it can be viewed on the link `localhost:3000/campaigns/miscellenea/programs`


ragesoss:
Cool! Did you test this on the tables where it gets used (to make sure the sorting features still work)? I'm not sure whether List.js itself also has jquery dependencies, or if it's self-contained and we were just using jquery to load it.

Sahethi:
Yup, I have tested everything on the front end for the courses table. I think I'll mention in the PR which element to access as well. Also, `list.js` seems to not have jQuery dependency according to the npm [docs](https://www.npmjs.com/package/list.js?ref=vanillalist).

Sahethi:
@ragesoss hey the check seemed to have failed and the error I saw was `Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')` and on reading a few articles, it was mentioned that the script tag should be at the bottom of the body. I'm not sure what to do here, I thought the elements didn't get loaded initially hence I changed the `document.onreadystatechange()` to `window.onload()` but the error seems to persist.

ragesoss:
Do you see similar errors in the browser console when you navigate around the dashboard locally?

Sahethi:
I can't see it in my browser console, I think I'll add an if condition to make sure the state is ready. and try again

TheTrio:
The problem is probably caused by the fact that the script is included globally. It runs everywhere - not just on the pages where the sorting is required. 

Have never used jquery but I suppose what's probably happening is that jquery doesn't run the `on` function if the DOM element doesn't exist(which it does not on a lot of pages). The `querySelector` function does no such thing and you end up trying to add an event listener to a null object.

Sahethi:
@TheTrio Hey, thanks for the info! I checked the readyState only cause it was implemented likewise in jQuery. I seemed to have figured out the DOMContentLoaded this with a few conditions. 

@ragesoss The failures I'm getting as of now is cause it's not able to find `revisions` in the `select dropdown` not too sure about that cause it works for `title, school` etc.

ragesoss:
This all looks good @Sahethi. While testing it, I noticed a few small problems with the sorting, which you could fix in a follow-up PR if you want to. (If you're already planning to work on something different, that's fine too.)

In particular, most of the tables start out by indicated that they are sorted by the first column, even if that's not the case. For example, the campaign articles list is sorted by characters added, but the UI initially indicates that it's being sorted by the title.

Also, on the Students table for a campaign can be sorted by edit count, but the first click on that column sorted it from low to high; it should be reversed, as it's more useful to focus on the users with high edit counts.

Sahethi:
> This all looks good @Sahethi. While testing it, I noticed a few small problems with the sorting, which you could fix in a follow-up PR if you want to. (If you're already planning to work on something different, that's fine too.)
> 
> In particular, most of the tables start out by indicated that they are sorted by the first column, even if that's not the case. For example, the campaign articles list is sorted by characters added, but the UI initially indicates that it's being sorted by the title.
> 
> Also, on the Students table for a campaign can be sorted by edit count, but the first click on that column sorted it from low to high; it should be reversed, as it's more useful to focus on the users with high edit counts.
> 
> (These problems were present already, and are unrelated to your changes.)

Hey! Thanks for reviewing! I'll make up a follow-up PR regarding the same as you mentioned and start working on it along with the other issue I'm dealing with.

PR #4884, opened 2022-03-24
Integration with Wikimedia Event Center

ragesoss:
This adds an API to allow syncing courses with Wikimedia Event Center. For courses with sync enabled, editors can only be added/removed via Event Center.

This version should be sufficient to allow for testing, once the Wikimedia Campaigns team is ready to do so.


ragesoss:
TODO:
* document how to generate an appropriate secret
* deploy to staging, generate a secret for staging, share with WMF for testing
* review UI to check for additional things that need to be adjusted when sync is enabled

PR #4886, opened 2022-03-27
Switch to Media Wiki API for fetching recently edited articles

### LINKED ISSUE #4392: Recent revision activity for a course should be fetched directly from mediawiki
As part of an effort to move `revisions` data out of the Dashboard database, we need to replace all the ways we currently use the `revisions` table to server data to the frontend.

### Current Behavior:
Visiting the `activity` tab of a course fetches the `revisions.json` route for a course, which provides a list of recent revisions from the Dashboard's database.

### Desired Behavior:
Instead of loading recent revisions from the Dashboard, the frontend should use the list of users and the date range to query relevant wikis for revisions.

(As a first stage, we can still continue using the Dashboard to provide a list of "course specific" revisions if requested.)

TheTrio:
# What this PR does

Attempts to fix #4392 

Currently, the references added are not being fetched. 

Apart from that, everything does seem to work all right. I'll add more details soon.

### DIFF
--- a/app/assets/javascripts/actions/revisions_actions.js
+++ b/app/assets/javascripts/actions/revisions_actions.js
@@ -9,9 +9,145 @@ import {
 import { fetchWikidataLabelsForRevisions } from './wikidata_actions';
 import logErrorMessage from '../utils/log_error_message';
 import request from '../utils/request';
+import API from '../utils/api';
+import moment from 'moment';
+import { stringify } from 'query-string';
+import { PageAssessmentGrades, PageAssessmentSupportedWiki } from '../utils/article_finder_language_mappings';
 
-const fetchRevisionsPromise = async (courseId, limit, isCourseScoped) => {
-  const response = await request(`/courses/${courseId}/revisions.json?limit=${limit}&course_scoped=${isCourseScoped}`);
+const fetchClassFromRevisions = async (wiki, API_URL, revisions) => {
+  // remove duplicates -> each article occurs only once after this
+  const uniqueArticles = [...new Set(revisions.map(revision => revision.title))];
+  const titles = uniqueArticles.join('|');
+
+  const params = {
+    action: 'query',
+    format: 'json',
+    titles,
+    prop: 'pageassessments',
+    pasubprojects: false,
+    palimit: 50
+  };
+  const response = await request(`${API_URL}?${stringify(params)}&origin=*`);
+  const ratings = (await response.json()).query.pages;
+
+
+  /* eslint-disable no-restricted-syntax */
+  for (const revision of revisions) {
+    // if pageassessments exists
+    if (ratings[revision.pageid].pageassessments) {
+      // pick the first key of the object pageassessments
+      let rating;
+      for (const value of Object.values(ratings[revision.pageid].pageassessments)) {
+        if (value.class) {
+          rating = value.class;
+          break;
+        }
+      }
+      if (rating) {
+        const mapping = PageAssessmentGrades[wiki.project][wiki.language][rating];
+        if (mapping) {
+          revision.rating_num = mapping.score;
+          revision.pretty_rating = mapping.pretty;
+          revision.rating = mapping.class;
+        }
+      }
+    }
+  }
+  /* eslint-enable no-restricted-syntax */
+};
+const fetchRevisionsFromWiki = async (wiki, usernames, start_time, end_time) => {
+  const params = {
+    action: 'query',
+    format: 'json',
+    list: 'usercontribs',
+    ucuser: usernames,
+    ucprop: 'ids|title|sizediff|timestamp',
+    uclimit: 50,
+    ucend: start_time,
+    ucstart: end_time,
+    ucdir: 'older',
+  };
+
+  let prefix;
+  if (wiki.language) {
+    prefix = `https://${wiki.language}.${wiki.project}.org`;
+  } else {
+    prefix = `https://${wiki.project}.org`;
+  }
+
+  const API_URL = `${prefix}/w/api.php`;
+
+  const response = await request(`${API_URL}?${stringify(params)}&origin=*`);
+  const json = await response.json();
+  const revisions = json.query.usercontribs;
+  /* eslint-disable no-restricted-syntax */
+
+  for (const revision of revisions) {
+    revision.wiki = wiki;
+    const diff_params = {
+      title: revision.title,
+      diff: 'prev',
+      oldid: revision.parentid
+    };
+
+    // url for the diff
+    revision.url = `${prefix}/w/index.php?${stringify(diff_params)}`;
+
+    // main article url - we use stringify to ensure that its encoded properly
+    revision.article_url = `${prefix}/w/index.php?${stringify({ title: revision.title })}`;
+
+    // to maintain the old structure of the revision object
+    revision.characters = revision.sizediff;
+    revision.mw_rev_id = revision.revid;
+    revision.id = revision.revid;
+    revision.revisor = revision.user;
+    revision.date = revision.timestamp;
+    revision.mw_page_id = revision.pageid;
+  }
+  /* eslint-enable no-restricted-syntax */
+
+
+  if (PageAssessmentSupportedWiki?.[wiki.project]?.[wiki.language]) {
+    // the wiki supports page assessments
+    await fetchClassFromRevisions(wiki, API_URL, revisions);
+  }
+
+  return revisions;
+};
+
+const fetchRevisionsFromUsers = async (course, users) => {
+  const usernames = users.map(user => user.username).join('|');
+
+  // Converting to ISO 8601 format
+  // the Media Wiki API doesn't accept fractional seconds. This gets rid of that
+  const start_time = moment.utc(course.timeline_start).format();
+  const end_time = moment.utc(course.timeline_end).format();
+
+  const revisions = [];
+
+  /* eslint-disable no-restricted-syntax */
+  for (const wiki of course.wikis) {
+    /* eslint-disable no-await-in-loop */
+    const items = await fetchRevisionsFromWiki(wiki, usernames, start_time, end_time);
+    revisions.push(...items);
+    /* eslint-enable no-await-in-loop */
+  }
+  /* eslint-enable no-restricted-syntax */
+
+  revisions.sort((revision1, revision2) => {
+    const date1 = new Date(revision1.date);
+    const date2 = new Date(revision2.date);
+    return date2.getTime() - date1.getTime();
+  });
+  return revisions;
+};
+
+const fetchRevisionsPromise = async (course, limit, isCourseScoped, users) => {
+  if (!isCourseScoped) {
+    course.revisions = await fetchRevisionsFromUsers(course, users);
+    return { course };
+  }
+  const response = await request(`/courses/${course.slug}/revisions.json?limit=${limit}&course_scoped=${isCourseScoped}`);
   if (!response.ok) {
     logErrorMessage(response);
     const data = await response.text();
@@ -21,7 +157,8 @@ const fetchRevisionsPromise = async (courseId, limit, isCourseScoped) => {
   return response.json();
 };
 
-export const fetchRevisions = (courseId, limit, isCourseScoped = false) => (dispatch) => {
+export const fetchRevisions = (course, limit, isCourseScoped = false) => async (dispatch, getState) => {
+  const courseId = course.slug;
   let actionType;
   if (isCourseScoped) {
     actionType = RECEIVE_COURSE_SCOPED_REVISIONS;
@@ -30,8 +167,22 @@ export const fetchRevisions = (courseId, limit, isCourseScoped = false) => (disp
     actionType = RECEIVE_REVISIONS;
     dispatch({ type: REVISIONS_LOADING });
   }
+  const state = getState();
+  let users;
+  if (state.users.users.length) {
+    // the users list has already been fetched
+    users = state.users.users;
+  } else {
+    try {
+      const response = await API.fetch(courseId, 'users');
+      users = response.course.users;
+    } catch (e) {
+      dispatch({ type: API_FAIL, data: e });
+      return;
+    }
+  }
   return (
-    fetchRevisionsPromise(courseId, limit, isCourseScoped)
+    fetchRevisionsPromise(course, limit, isCourseScoped, users)
       .then((resp) => {
         dispatch({
           type: actionType,
--- a/app/assets/javascripts/actions/revisions_actions.js
+++ b/app/assets/javascripts/actions/revisions_actions.js
@@ -13,6 +13,7 @@ import API from '../utils/api';
 import moment from 'moment';
 import { stringify } from 'query-string';
 import { PageAssessmentGrades, PageAssessmentSupportedWiki } from '../utils/article_finder_language_mappings';
+import { url } from '../utils/wiki_utils';
 
 const fetchClassFromRevisions = async (wiki, API_URL, revisions) => {
   // remove duplicates -> each article occurs only once after this
@@ -55,7 +56,7 @@ const fetchClassFromRevisions = async (wiki, API_URL, revisions) => {
   }
   /* eslint-enable no-restricted-syntax */
 };
-const fetchRevisionsFromWiki = async (wiki, usernames, start_time, end_time) => {
+const fetchRevisionsFromWiki = async (wiki, usernames, start_time, end_time, prevContinueToken) => {
   const params = {
     action: 'query',
     format: 'json',
@@ -67,6 +68,12 @@ const fetchRevisionsFromWiki = async (wiki, usernames, start_time, end_time) => 
     ucstart: end_time,
     ucdir: 'older',
   };
+
+  if (prevContinueToken) {
+    // if token exists, add it.
+    params.uccontinue = prevContinueToken.uccontinue;
+    params.continue = prevContinueToken.continue;
+  }
 
   let prefix;
   if (wiki.language) {
@@ -80,6 +87,7 @@ const fetchRevisionsFromWiki = async (wiki, usernames, start_time, end_time) => 
   const response = await request(`${API_URL}?${stringify(params)}&origin=*`);
   const json = await response.json();
   const revisions = json.query.usercontribs;
+  const continueToken = json.continue;
   /* eslint-disable no-restricted-syntax */
 
   for (const revision of revisions) {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/course/course.jsx
```diff
@@ -76,7 +76,7 @@ export const Course = withRouter(createReactClass({
   render() {
     const courseSlug = this.getCourseSlug();
     const course = this.props.course;
-    if (!courseSlug || !course || !course.home_wiki || course.title === '' || !this.props.users.length) { return <div />; }
```
ragesoss: This doesn't seem right. Does this mean the whole Course component will just return an empty div if there are no users in the course?

app/assets/javascripts/actions/revisions_actions.js
```diff
@@ -9,9 +11,349 @@ import {
...
+      continueToken = json.continue;
+    } else {
+      hasMore = false;
+    }
+  }
+  return allData;
+};
+const fetchAllRevisions = async (API_URL, days, usernames, wiki, course_start, last_date) => {
+  let ucend;
+  if (moment(last_date).subtract(days, 'days').isBefore(course_start)) {
```
ragesoss, line 51: If we can avoid using `moment` here, I think that would be preferable. All the dates in `course` are UTC and that's also what the MediaWiki API uses. I'd like, eventually, to replace `moment` with a lighter library.

app/assets/javascripts/actions/revisions_actions.js
```diff
@@ -9,9 +11,349 @@ import {
...
 import logErrorMessage from '../utils/log_error_message';
 import request from '../utils/request';
+import moment from 'moment';
+import { stringify } from 'query-string';
+import { ORESSupportedWiki, PageAssessmentGrades, PageAssessmentSupportedWiki } from '../utils/article_finder_language_mappings';
+import { url } from '../utils/wiki_utils';
+import { chunk, flatten } from 'lodash-es';
+import { queryUrl } from '../utils/article_finder_utils';
+
+const fetchAll = async (API_URL, params, continue_str) => {
```
ragesoss, line 21: It might make sense to put all this new functionality into a separate actions file, so that we have one file for interacting with revisions from the Dashboard, and another for getting revisions and related data from MediaWiki.

spec/features/course_page_spec.rb
```diff
@@ -472,8 +475,11 @@ def js_visit(path, count=3)
 
   describe 'activity view' do
     it 'displays a list of edits' do
-      js_visit "/courses/#{slug}/activity"
-      expect(page).to have_content 'Article 1'
+      Capybara.using_wait_time 10 do
+        js_visit "/courses/#{slug}/activity"
+        expect(page).to have_content('Bareilly College').twice
+        expect(page).to have_content('Hartmann College', count: 3)
```
ragesoss, line 481: Is this going to break if/when that user has more recent edits to different pages?

app/assets/javascripts/actions/revisions_actions.js
```diff
@@ -9,9 +11,349 @@ import {
...
+    };
+    const response = await request(`${API_URL}?${stringify(params)}&origin=*`);
+    const pages = (await response.json())?.query?.pages;
+    if (pages) { allRatings.push(pages); }
+  }
+  if (!allRatings) {
+    // no ratings found
+    return;
+  }
+  // merge list of objects of ratings into a single object
```
ragesoss, line 108: I think all this processing of the results should be extracted into utility functions, with a clear explanation in comments of what shape of data it expects and what shape of data it outputs.

app/assets/javascripts/actions/revisions_actions.js
TheTrio: hmm. I guess I could, although it wouldn't be as concise as this. Though I guess I could create a helper function to extract the manual date handling.

spec/features/course_page_spec.rb
ragesoss: Just put an expectation of at least X revisions appearing, without looking for any specific ones. That way it will verify that it pulls in revisions, but it will always work regardless of what new revisions happen.

ragesoss:
For the 'too many revisions' problem with long-ago revisions, I think it's fine to set a cut-off for both how far back it will try to fetch revisions (say, 5 years) and for how many revisions to retrieve before giving up on a `continue` (say, 1000). I'd rather have edge cases just not work or provide incomplete results, versus potentially fetching huge numbers of revisions that take a long time and/or spam the API with hundreds of requests.

app/assets/javascripts/reducers/revisions.js
```diff
@@ -76,7 +76,15 @@ export default function revisions(state = initialState, action) {
     case SORT_REVISIONS: {
       const absolute = action.key === 'characters';
       const desc = action.key === state.sort.sortKey;
-      const sortedRevisions = sortByKey(state.revisions, action.key, null, desc, absolute);
+      // some values like references and rating num are fetched independently of revisions
+      // to reduce waiting time
+      let mapTo;
+      if (action.key === 'references_added') {
```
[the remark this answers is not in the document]
ragesoss: I don't have a clear picture of why the key function is necessary. What is different about this situation that we can't just pass the key to the sort function like with all the other sorts that happen currently?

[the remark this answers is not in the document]
TheTrio: Since the page assessments and references added are fetched after the revisions have been rendered, they exist as different redux states. Before this, all the properties were part of the revision object - now, that's not the case. The references added state for example is an object with a key of the revision id and the value of how many references were added.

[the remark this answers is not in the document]
ragesoss: Ah, I see. Why not add them to the `revision` objects in redux whenever they get received?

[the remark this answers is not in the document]
TheTrio: I didn't want to mutate the revisions array since that could lead to more problems. Say if the user fetches revisions and they're loaded in. If page assessments comes in first,  sets `state.revisions.revisions` to this a new array(which has the class info), then this change would be overridden by the references request, which would do the same - set `state.revision.revisions` to the its version of the array. Not to mention that if the user presses see more before either of them can come in, they will have a different version of the revisions array. 

I just felt a lot of things could go wrong, although I assume most of the times they won't. This way it doesn't matter at all in which order the requests are completed since all 3 requests would work on 3 different redux states.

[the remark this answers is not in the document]
ragesoss: Can't it be done in a way that doesn't risk any data being overwritten? The assessments received redux action should only add the new data to the relevant revisions, not replace or remove any other attributes.

Like, first you fetch revisions and have an array like `[{ id: 123, title: 'abc' }]`, and then you receive page assessment data and it becomes `[{ id: 123, title: 'abc', class: 'c' }]`, and then you receive ORES data and it becomes `[{ id: 123, title: 'abc', class: 'c', references_added: 2 }]`. It should be possible to write reducer functions such that the order of receiving data doesn't matter.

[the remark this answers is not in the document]
ragesoss: This kind of situation is where redux really shows its value, in my opinion. We do something similar in the `article_finder.js` reducer.

TheTrio:
Some changes

1. Revisions older than 5 years won't be fetched
2. The max time period will be 2 years
3. If the number of revisions in a period exceeds 1000, the request aborts(thus skipping all continues) to prevent the API from getting spammed. 

I've also refactored the code, which (hopefully) makes it more readable and easier to follow. And I have written 2 new reducers as you suggested, and they have eliminated the need for the other 2 redux state variables.

app/assets/javascripts/utils/media_wiki_references_utils.js
```diff
@@ -0,0 +1,79 @@
...
+    // wiki does not support ORES
+    return;
+  }
+  let suffix;
+  if (wiki.project === 'wikidata') {
+    suffix = 'wikidatawiki';
+  } else {
+    suffix = `${wiki.language}wiki`;
+  }
+  const API_URL = `http://ores.wikimedia.org/v3/scores/${suffix}`;
```
ragesoss, line 31: This should be `https:`

app/assets/javascripts/utils/media_wiki_references_utils.js
```diff
@@ -0,0 +1,79 @@
...
+
+
+  const promises = [];
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const revid_chunk of chunks) {
+    const params = {
+      revids: revid_chunk.join('|'),
+      features: true,
+      models: 'articlequality'
```
ragesoss, line 43: For wikidata, the model is called `itemquality` instead. Querying for `articlequality` results in a 404.

app/assets/javascripts/actions/media_wiki_revisions_actions.js
```diff
@@ -0,0 +1,24 @@
...
+import { getWikiMap } from '../utils/revision_utils';
+import { fetchReferencesAdded } from '../utils/media_wiki_references_utils';
+
+export const fetchRevisionsAndReferences = async (revisions, dispatch) => {
+  const wikiMap = getWikiMap(revisions);
+  fetchClassFromRevisions(wikiMap).then((assessments) => {
+    // eslint-disable-next-line no-console
+    console.log('Successfully fetched all page assessment information');
+    return dispatch({
+      type: 'RECEIVE_ASSESSMENTS',
```
ragesoss, line 11: These types should be imported as constants, to follow the convention of the rest of the redux actions.

app/assets/javascripts/actions/revisions_actions.js
```diff
@@ -9,9 +10,25 @@ import {
 import { fetchWikidataLabelsForRevisions } from './wikidata_actions';
 import logErrorMessage from '../utils/log_error_message';
 import request from '../utils/request';
+import { fetchRevisionsFromUsers } from '../utils/mediawiki_revisions_utils';
+import { fetchRevisionsAndReferences } from './media_wiki_revisions_actions';
+import { sortRevisionsByDate } from '../utils/revision_utils';
+
+const fetchRevisionsPromise = async (course, limit, isCourseScoped, users, last_date, lastRevisions, dispatch) => {
```
ragesoss, line 17: I think this makes the isCourseScoped vs normal requests so different that they shouldn't be handled in the same function. It will be easier to reason about if there is a separate function for course-scoped revision requests. Probably at the component level, it should just call a different action (`fetchRevisions` vs `fetchCourseScopedRevisions`) so that the data flows are totally separate.

app/assets/javascripts/utils/media_wiki_classes.js
```diff
@@ -0,0 +1,83 @@
...
+    const pages = (await response.json())?.query?.pages;
+    if (pages) { allRatings.push(pages); }
+  }
+  if (!allRatings) {
+    // no ratings found
+    return;
+  }
+  // merge list of objects of ratings into a single object
+  const ratings = Object.assign({}, ...allRatings);
+  const assessments = {};
```
ragesoss, line 43: The data transformation part of this function should be extracted to a separate function.

app/assets/javascripts/utils/media_wiki_classes.js
```diff
@@ -0,0 +1,83 @@
...
+  // eslint-disable-next-line no-console
+  console.log(`Fetching page assessments from ${wiki_url}`);
+
+  const uniqueArticles = [...new Set(revisionsOfWiki.filter(revision => revision.ns === 0).map(revision => revision.title))];
+  const allRatings = [];
+  const chunks = chunk(uniqueArticles, 50);
+
+  // eslint-disable-next-line no-restricted-syntax
+  for (const uniqueArticlesChunk of chunks) {
+    const params = {
```
ragesoss, line 25: The code for making an individual request can probably be extracted to a separate function here.

app/assets/javascripts/components/revisions/diff_viewer.jsx
```diff
@@ -5,6 +5,7 @@ import jQuery from 'jquery';
 import moment from 'moment';
 import SalesforceMediaButtons from '../articles/salesforce_media_buttons.jsx';
 import Loading from '../common/loading.jsx';
+import { url } from '../../utils/wiki_utils';
```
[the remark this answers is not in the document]
TheTrio: Well, its not related to this PR, but it does fixes a problem - the wikidata base url is `https://www.wikidata.org` The function earlier evaluation it to `https://wikidata.org` Clicking on the diff viewer for a wikidata item earlier lead to a 301 (Moved Permanently) error. 

If you think this change should be made in another PR, I'm fine with that as well.

app/assets/javascripts/utils/media_wiki_classes.js
TheTrio: Well, its not related to this PR, but it does fixes a problem - the wikidata base url is https://www.wikidata.org The function earlier evaluation it to https://wikidata.org Clicking on the diff viewer for a wikidata item earlier lead to a 301 (Moved Permanently) error.

If you think this change should be made in another PR, I'm fine with that as well.

app/assets/javascripts/utils/mediawiki_revisions_utils.js
```diff
@@ -0,0 +1,119 @@
...
+import { stringify } from 'query-string';
+import { url } from './wiki_utils';
+import moment from 'moment';
+
+export const fetchRevisionsFromUsers = async (course, users, days, last_date) => {
+  const usernames = users.map(user => user.username);
+
+  let revisions = [];
+  const wikiPromises = [];
+  /* eslint-disable no-restricted-syntax */
```
ragesoss, line 17: three times

app/assets/javascripts/components/activity/activity_handler.jsx
```diff
@@ -41,7 +41,7 @@ export const ActivityHandler = createReactClass({
       <div className="activity-handler">
         <SubNavigation links={links}/>
         <Routes>
-          <Route path="recent" element={<RevisionHandler {...this.props}/>} />
+          {this.props.usersLoaded && <Route path="recent" element={<RevisionHandler {...this.props}/>} />}
```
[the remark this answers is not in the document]
TheTrio: Since the revisions are fetched as soon as the `RevisionHandler` component is mounted, the users must be fetched before that happens. I could use a `componentDidUpdate` to run the code only when the users were fetched, but the comment in `app/assets/javascripts/components/revisions/revisions_handler.jsx` says that placing the code here is intentional, so this just felt like an easier solution.

app/assets/javascripts/utils/revision_utils.js
```diff
@@ -0,0 +1,85 @@
...
+    if (json.continue) {
+      continueToken = json.continue;
+    } else {
+      hasMore = false;
+    }
+  }
+  return allData;
+};
+
+// uses the ORES API to get the number of references in a particular revision
```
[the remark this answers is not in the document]
TheTrio: Yup, will change this comment.

app/assets/javascripts/components/activity/activity_handler.jsx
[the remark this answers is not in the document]
ragesoss: okay. I'm not really clear on the logic, but I'm sure it made sense at the time. The user experience is reasonable for when the `recent` route gets loaded directly (it first renders the nav, then renders the Recent Activity table in the 'loading' state once the users are received, I guess).

TheTrio:
Apart from the refactoring, I have changed the following 

1. Now, at max 200 revisions are rendered on the first load. The number of fetched revisions remains unchanged. 
2. When the user clicks on show more and there are revisions already fetched but not rendered, they are added to the table. If not, new revisions are fetched. The max number of revisions which can be added to the table when the user clicks "Show More" is 50. 
3. If the user sorts the table before clicking on show more, the table will now maintain its order - that is, if the table was sorted by characters before(ascending), the bottom most revision of the table after clicking show more will still be the one with the most characters.

The exception to this is that if the user doesn't change the default order of the table or keeps it at date(descending). In both these cases, the revisions will be added as it is - no sorting is done. 

If however the order is date(ascending), then the new revisions will be added at the top of the table(to maintain the order). There is currently no obvious indication to the user that anything is being added at all. 

I guess one solution would be to simply scroll up to the top of the page. But I'm not exactly sure what the best experience would be.

ragesoss:
I'm testing it out with a copy of a pretty active course, https://outreachdashboard.wmflabs.org/courses/CBA_Unisa_e_GWMAB/CBA_UNISA_-_GWMAB_Datathon_per_Wikigap_2022_(marzo_2022)

In this case, the user experience is decent, but it fires off enough requests that it quickly runs into 429 errors for the ORES API.

I think this needs a concurrency limit. You can find an implementation for limiting concurrent requests in `article_finder_actions.js`. The same strategy might be good here.

app/assets/javascripts/utils/media_wiki_references_utils.js
```diff
@@ -4,6 +4,10 @@ import { queryUrl } from './article_finder_utils';
 import { chunk } from 'lodash-es';
 import { getReferencesCount } from './revision_utils';
 
+import promiseLimit from 'promise-limit';
+
+const limit = promiseLimit(10);
```
ragesoss, line 9: Did you test this? 10 might be too many.

TheTrio: It did work for me, although one request did fail the second time I ran it. Do I just set it to 5?

ragesoss: yeah, I think 5 is probably good.

app/assets/javascripts/utils/media_wiki_references_utils.js
```diff
@@ -0,0 +1,78 @@
...
+  let suffix;
+  if (wiki.project === 'wikidata') {
+    suffix = 'wikidatawiki';
+    models = 'itemquality';
+  } else {
+    suffix = `${wiki.language}wiki`;
+    models = 'articlequality';
+  }
+  const API_URL = `https://ores.wikimedia.org/v3/scores/${suffix}`;
+  const revids = revisions.filter(revision => revision.ns === 0).map(revision => `${revision.parentid}|${revision.revid}`);
```
ragesoss, line 40: This and the following `chunks` could use some additional comments to explain what's going on with the magic numbers and the `parentid` values.

app/assets/javascripts/utils/media_wiki_references_utils.js
```diff
@@ -0,0 +1,78 @@
...
+    });
+  }))).filter(item => item).map(data => data?.[suffix]?.scores);
+
+  // merge the array of objects into one object
+  const combinedObject = Object.assign({}, ...values);
+
+  const referencesAdded = {};
+
+  for (const revision of revisions) {
+    const references = getReferencesCount(combinedObject?.[revision.revid], wiki.project === 'wikidata');
```
ragesoss, line 61: The code in this loop should probably be extracted to a function, and tested (along with the `getReferencesCount` function). This seems like critical logic in terms of determining whether we have enough data to determine the reference count for a particular revision.

TheTrio:
> It should only fetch activity for students, not for users in other roles.

I thought this was the expected behavior :sweat_smile: 

Have fixed this

ragesoss:
This is great work! Thanks @TheTrio!

TheTrio:
:partying_face: Thanks for perfecting it! Couldn't have caught most of the bugs myself.

Issue #4510, opened 2021-05-24
Replace react-motion with react-spring


ragesoss:
We currently use the unmaintained `react-motion` library to provide animations for when a user is reordering blocks in a course Timeline. (See `week.jsx`.) It throws warnings right now, and is likely to stop working with a future version of React.

It looks like `react-spring` provides the same sort of functionality — it's similarly based on the `spring` concept — and it is actively maintained.

Convert `week.jsx` to use `react-spring` instead of `react-motion`.

TheTrio:
This can be closed as well.

PR #4897, opened 2022-03-31
Migrating to react-spring from react-motion

Sahethi:
## What this PR does?
Addresses the issue #4510. In `app/assets/javascripts/components/timeline/week.jsx` the `Motion` tag was replaced with `Spring` tag according to [docs](https://react-spring.io/components/spring). `defaultStyle` indicated the `from` and `style` indicates `to`. The values of [220, 30] is for stiffness and damping of the animation which is currently not considered as a priority as it can be added in later. 

References to the animation we want to perform with `react-spring` would be [this](https://codesandbox.io/s/react-spring-draggable-list-wfwv3?file=/src/components/DraggableList.tsx).

## Screenshots
Current Issue: Here, I have 3 blocks, in which all of them are collapsing on top of each other.
image

## Additional Info
Another method, I've tried.
```
if (this.props.reorderable) {
    const orderableBlock = (value) => {
        const rounded = Math.round(value.y);
        const animating = rounded !== i * 75;
        const willChange = animating ? 'top' : 'initial';
        const blockLineStyle = {
            top: rounded,
            position: 'absolute',
            width: '100%',
            left: 0,
            willChange,
            marginLeft: 0
        };
        return (
            <Spring key={block.id} marginTop={`${i * 75}px`}>
                {styles => {
                    <a.li style={{ ...blockLineStyle, ...styles }}>
                        <OrderableBlock
                            block={block}
                            canDrag={true}
                            animating={animating}
                            onDrag={this.props.onBlockDrag.bind(null, i)}
                            onMoveUp={this.props.onMoveBlockUp.bind(null, block.id)}
                            onMoveDown={this.props.onMoveBlockDown.bind(null, block.id)}
                            disableDown={!this.props.canBlockMoveDown(block, i)}
                            disableUp={!this.props.canBlockMoveUp(block, i)}
                            index={i}
                            title={block.title}
                            kind={[I18n.t('timeline.block_in_class'), I18n.t('timeline.block_assignment'), I18n.t('timeline.block_milestone'), I18n.t('timeline.block_custom')][block.kind]}
                        />
                    </a.li>
                }}
            </Spring>
        );
    };
    return orderableBlock
}
```

Here, I felt like it would be easier to manage if one abstracts everything in `OderableBlock` component. But it didn't work as it wasn't in a react fragment. Here 2 important things to note:
1. Any element receiving an animated property must be like `<a.li></a.li>` with `import {a, Spring} from "@react-spring/web"`
2. The `<Spring></Spring>` element provides an object to its children to access, that's why `<Spring prop={i * 10}>{styles => styles.prop}</Spring>` (Then one can work on stiffness properties)


ragesoss:
Thanks for working on this Sahethi!

Sahethi:
> Thanks for working on this Sahethi!

No problem, thank you! @ragesoss

PR #4905, opened 2022-04-05
ArticleFinder fetch labels for Wikidata items

### LINKED ISSUE #2887: ArticleFinder should fetch labels for Wikidata items
ArticleFinder can be used for Wikidata, but the results are only Q numbers.

It should show the labels instead or in addition to Q numbers.

sanyud24:
## What this PR does
Closes #2887

## Screenshots
Before:
image

After:
image


ragesoss:
Nice! This would be better if the Title column showed the Q number in parentheses, like we currently do with the Activity tab (see for example https://outreachdashboard.wmflabs.org/courses/AACC/Art_and_Feminism_(March_1-March_31)/activity/recent near the bottom).

app/assets/javascripts/reducers/article_finder.js
```diff
@@ -140,6 +140,18 @@ export default function articleFinder(state = initialState, action) {
         lastRelevanceIndex: state.lastRelevanceIndex + 50,
       };
     }
+    case RECEIVE_ARTICLE_TITLE: {
```
ragesoss, line 143: Can this instead be done with the existing `RECEIVE_WIKIDATA_LABELS` action? Using the same data flow as we do with other places that use Wikidata labels would keep things as simple as possible.

sanyud24: @ragesoss You mean reusing the same constant to denote the action type in the `article_finder` reducer? Or reusing the `wikidata_labels` reducer (though it doesn't sound possible to do)?

ragesoss: I mean re-using the reducer. I think it should be possible, as the strategy is to replace titles with labels (when available) in the individual components. See how they are used in ArticleList and article.jsx, for example.

sanyud24: @ragesoss done, reusing the same reducer now. Had to change the `getFilteredArticleFinder` selector so it can mutate the article title from the updated labels

app/assets/javascripts/selectors/index.js
```diff
@@ -139,8 +140,10 @@ export const getFilteredAlerts = createSelector(
 );
 
 export const getFilteredArticleFinder = createSelector(
-  [getArticleFinderState], (articleFinder) => {
-    return pickBy(articleFinder.articles, (article) => {
+  [getArticleFinderState, getWikidataLabelsState], (articleFinder, wikidataLabels) => {
+    return pickBy(articleFinder.articles, (article, key) => {
+      if (wikidataLabels.labels && !!wikidataLabels.labels[key]) article.displayTitle = `${wikidataLabels.labels[key]} (${key})`;
```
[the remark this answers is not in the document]
sanyud24: @ragesoss fixed now

ragesoss:
Nice! This could be improved slightly if the "Label (Q-number)" part of each row used the same styling as in `revision.jsx`.

app/assets/javascripts/actions/article_finder_action.js
```diff
@@ -218,6 +220,7 @@ export const fetchKeywordResults = (keyword, home_wiki, offset = 0, continueResu
     return data.query.search;
   })
   .then((articles) => {
+    if (home_wiki.project === 'wikidata') fetchWikidataLabels(articles, dispatch);
```
ragesoss, line 223: Do the `articles` here have a `project` property? If so, this conditional could be replaced with a call to `fetchWikidataLabelsForArticles` (which already handles filtering for wikidata).

sanyud24: @ragesoss No, it doesn't have that property from the API call.

ragesoss:
@sanyud24 would you like to make the improvement I suggested for the styling? If not, I think I could merge it any time now, but if you'd like to make that change, I'll wait until it's included.

sanyud24:
> @sanyud24 would you like to make the improvement I suggested for the styling? If not, I think I could merge it any time now, but if you'd like to make that change, I'll wait until it's included.

No I don't wish to work on it , u can merge it

Issue #4561, opened 2021-07-26
Improve the course approval user interface


ragesoss:
### Current Behavior:
When Wiki Education staff review and approve a submitted course, they must:

* Click 'Edit Details'
* Add the program manager in the staff role
* Add the wiki expert in the staff role
* (sometimes) Add additional tags to the course
* Select the campaign corresponding to the current or upcoming term and add it to the course

### Desired Behavior:

Instead of these separate steps (and process that isn't explicit in the UI), when a course is submitted there should be a single dedicated interface for an admin to efficiently do all the steps of course approval.

ragesoss:
It should be easy within this approval workflow to add additional campaigns as well. There should probably be a default campaign, which is the upcoming campaign as defined by the "recruiting term" setting, but we often need to add additional campaigns at the point of approval as well.

vaidehi44:
@ragesoss, can you mention steps to reproduce this? I am not able to get to a point where I can approve submitted courses.

vaidehi44:
Okay. And, how can I login as Admin?

vaidehi44:
This seems interesting to me @ragesoss. Can I work on this?

ragesoss:
@vaidehi44 sure, go for it.

vaidehi44:
Hey @ragesoss. A few questions after looking at the present UI (haven't actually seen the code yet) -

- After clicking on 'View Course Page' the wiki education staff reviewer gets the same interface as it is when some instructor creates a course. This gives the reviewer ability to even edit the course page. Is this deliberately done? If not, then I think first step would be to present the course details in such a way that the reviewer can only read/review it and not edit. Then, we should have a separate section which the reviewer easily identifies as their area of adding program manager, wiki expert, etc.
- Who is the program manager? How does the wiki education staff identify them?

ragesoss:
The program manager is typically also the person who is reviewing a submitted course page (Helaine — https://wikiedu.org/about-us). Within the Dashboard, they are identified by having the `classroom_program_manager` Special User role (via /settings). The review sometimes needs to make edits to a course before approving it, and in general, an admin should be able to edit a course no matter what state it's in with respect to approval. But I think a separate section — perhaps an extra component that shows up at the top, with a form for submitting all the approval-related updates at once — would make sense.

vaidehi44:
Okay. Will work on the new section then and let you know.

vaidehi44:
[...]

PR #4913, opened 2022-04-06
Course approval UI for admins

vaidehi44:

## What this PR does
This PR tries to improve the UI/UX for course approval process for admins. (#4561)
## Screenshots
Before:

image

After:
A new course approval form with prefilled dropdowns.
image

## Some pending tasks -
- Include "Tags" field in the form. Make "Campaigns" and "Tags" multi-select dropdowns.
- Right now, the form opens up for admin users every time they click on the "Edit Details" button. So, this needs to be corrected and the form needs to be displayed only when a campaign is not submitted for the course.
- Improve overall UI of the form. (remove extra space between dropdowns, move "Approve Course" button to the bottom right, etc)


vaidehi44:
@ragesoss, some things are still needed to be worked upon. I have mentioned them in the PR. Please have a look at current code and let me know of your review.

ragesoss:
@vaidehi44 thanks! It looks from the screenshot like this form is a modal that covers up course content, and/or is only rendered when in Edit Details mode?

When it shows up, it shouldn't cover any other page content or prevent interaction with the rest of the page. It should probably basically go in the same place as `EnrollCard` in `course.jsx`.

vaidehi44:
Yes @ragesoss, right now it is opening up in the edit details mode for admins. I think, rendering it in the same place as enroll card would be a good idea. 
Also, the notification which prompts admin to approve a course has a `Home` button. So, rendering the form just below it in the `Home` tab will be a good thing.
image

vaidehi44:
@ragesoss, I saw a cool "lodash" method - `difference`, used in the selectors file. I was wondering if I can use it in my .jsx file. I think, its just a javascript library, but still wanted to ask first if I can use it.

ragesoss:
> @ragesoss, I saw a cool "lodash" method - `difference`, used in the selectors file. I was wondering if I can use it in my .jsx file. I think, its just a javascript library, but still wanted to ask first if I can use it.

Yes, you can.

app/assets/javascripts/components/overview/course_approval.jsx
```diff
@@ -0,0 +1,243 @@
...
+import Select from 'react-select';
+import selectStyles from '../../styles/single_select';
+import { fetchSpecialUsers } from '../../actions/settings_actions';
+import { fetchAllCampaigns, addCampaign } from '../../actions/campaign_actions';
+import { addUser } from '../../actions/user_actions';
+import { getCourseApprovalStaff } from '../../selectors';
+import { STAFF_ROLE } from '../../constants';
+
+
+const CourseApproval = createReactClass({
```
ragesoss, line 15: I'd prefer new components to be functional components, if possible. That will make it easier to follow future developments in React, I think.

vaidehi44: Okay

vaidehi44:
@ragesoss, I have moved the form from edit mode page to the overview page. This is how its looking - 

ezgif com-gif-maker

vaidehi44:
I am working on turning the react class component into functional one. Other than that, I think, there's one thing left - to infer default campaign from course dates. Here is the comment you had made about the same -

image

ragesoss:
The demo video looks great! Does it still have an issue with requiring a page refresh after submitting the form?

vaidehi44:
> The demo video looks great! Does it still have an issue with requiring a page refresh after submitting the form?

@ragesoss Yeah, sometimes I get Internal Server Error. In that case, the changes only reflect after I refresh the page.

ezgif com-gif-maker (1)

But, sometimes it works perfectly. As you can see in the previous video, it got immediately updated and the form also vanished after it was submitted. So, I think, its not something specific to the form component. I even face this issue when I update users/campaigns on the edit details page.

ragesoss:
Okay. Can you replicate the error? Is it based on specific usernames or other values being submitted? You can check the `guard` console to see what is causing the server error.

vaidehi44:
I don't think this happens specifically, but sometimes I do get 500 errors unnecessarily. It works fine again when I restart the server. I am not familiar with guard console. Will take a look at it.

ragesoss:
`guard` is just a watcher process for the rails server, in development. It is mostly the same as `rails s` (the typical command for running a rails development server), but it automatically restarts the server any time you save a change to ruby code so that your browser hits the latest version.

The terminal where you are running the rails server (either `guard` or `rails s`) is where you can see the stack trace and debugging details for the Rails side of things, so any time you get a 500 server error, that's where you should look for the details.

ragesoss:
Are these problems that could also happen in production, for example if a course gets approved, then the campaign is removed, then the approval form gets submitted again?

vaidehi44:
I didn't get the question actually. The form would only appear if `course.published` is false. Then, it would only get submitted if the  admin clicks on the approve button.

ragesoss:
But then if you remove the campaign from the course, it will again be `course.published` false. So it's possible to submit the form more than once for the same course, correct?

vaidehi44:
> But then if you remove the campaign from the course, it will again be `course.published` false. So it's possible to submit the form more than once for the same course, correct?

Yeah, right.

vaidehi44:
I think, this is what we had discussed on slack - the situation where this form should appear is when an admin is viewing the course page, the course is submitted, but it doesn't have a campaign. This is the same situation where we currently render a notice at the top of the page.

ragesoss:
It doesn't cause any errors if you're just adding the campaign. But if it's trying to add a user who is actually already enrolled (ie, the wiki expert or program manager) then adding the same user again will result in an error. If that's what happened when you hit the errors locally, that's something that could also happen in production if the form doesn't have handling for that situation (ie, it should not try to add a user who is already in the course).

vaidehi44:
Also, if there was any error due to repeated user being added, it would have been something like this - User already enrolled.... But, here I had got Internal Server Error. I think, that too was due to some mistake from my side. Can you test this once? I think this should work fine now. If there are any errors, I will look into them.

ragesoss:
Cool. I will test this out soon. Do try to trigger the server error again, though so you can see what it is/was.

vaidehi44:
Okay, alright.

ragesoss:
This works really nicely! You're still planning to implement inferring the initial selected campaign from the course dates, correct?

vaidehi44:
Yeah, I am working on inferring the default selected campaign from course dates. Do you have any other suggestions?

ragesoss:
No, I think this will be ready to go once that feature is added. I'll do a little more testing at that point to make sure I didn't miss any problems.

ragesoss:
It would also be good to add a feature spec to exercise this new functionality in the test suite, but I can merge it without that.

vaidehi44:
> I am working on turning the react class component into functional one. Other than that, I think, there's one thing left - to infer default campaign from course dates. Here is the comment you had made about the same -
> 
> image

I have some questions about this -

- I am confused about this `term` property. It is a string and I think it is set by the instructors. So, not sure what values it can have ... like `spring_2023` , `fall_2023` as you mentioned.
- Will we also use the `start` and `end` properties for course and campaigns to infer the appropriate campaign?

vaidehi44:
I am just confused whether it is to be done using `term` or course dates or mixture of both.

ragesoss:
As you note, the `term` for a course is set by the instructor, and shouldn't be used for this feature. I wrong "default to `spring_2023` term", but I meant campaign instead. We typically create campaigns that correspond to `spring`, `summer`, and `fall` terms each year, and assign the appropriate campaign to each course in our Wikipedia Student Program, but which campaign to assign is basically just a function of the start and end dates (mainly the start date).

vaidehi44:
Okay got it. But, what if there's no such campaign for a course?

ragesoss:
In that case, it should not select any campaign by default (allowing the admin to choose any existing campaign). In most cases, if there's no appropriate campaign that already exists, the admin would need to create the campaign before returning to approve the course.

vaidehi44:
Also, from the schema, I think, there's no `term` defined for a campaign. So, should I infer the `spring`, `summer`, `fall` terms from the campaign title?

ragesoss:
You should infer it from the campaign `slug`.

vaidehi44:
okay, thanks!

vaidehi44:
@ragesoss, I wanted some help in mapping course dates/months with term/season.

vaidehi44:
@ragesoss, I have pushed the code for inferring default selected campaign. But, here I am stuck on a failing test for a long time. Actually, I did some changes in the rails code in order to get `slug` in `all_campaigns`. And, the failure is probably due to it. But I also updated the `.spec.jsx` file with new structure(in my last commit), still I am getting the same error. Can you please look into this and point out what I am missing here.

ragesoss:
The failing tests are JS tests that don't interact with the Ruby code at all. It looks like something related to the redux state for `campaigns` is what is causing the test failure. (There might also be problems in the Ruby tests, but those did not get run because a JS test failed.)

app/assets/javascripts/reducers/campaigns.js
```diff
@@ -26,7 +26,7 @@ export default function campaigns(state = initialState, action) {
     case RECEIVE_ALL_CAMPAIGNS: {
       const newState = {
         ...state,
-        all_campaigns: action.data.values
+        all_campaigns: action.data.campaigns
```
ragesoss, line 29: This change is almost certainly what is causing the tests to fail.

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -492,72 +492,74 @@ const Details = createReactClass({
     }
 
     const shared = (
```
ragesoss, line 494: Why are there changes to `details.jsx`?

app/assets/javascripts/components/course/course_approval.jsx
```diff
@@ -0,0 +1,354 @@
...
+      return {
+        value: user.username,
+        label: `${user.username} (${user.realname})`
+      };
+    });
+    return options;
+  };
+
+  const setDefaultCampaign = () => {
+    // Try to infer the best suitable campaign from the start date of course.
```
ragesoss, line 78: All this logic should be extracted to a separate file as a pure function, so that we can write simple tests for how a start date gets converted into a campaign slug.

app/assets/javascripts/components/course/course.jsx
```diff
@@ -56,6 +57,17 @@ export const Course = withRouter(createReactClass({
     return `${course_school}/${course_title}`;
   },
 
+  showCourseApprovalForm() {
```
ragesoss, line 60: This would benefit from a comment explaining what situations we intend to show the form.

app/assets/javascripts/components/overview/details.jsx
vaidehi44: As initially I was rendering the form in the edit mode, I was working with `details.jsx`. But, I had removed that part of code. May be some changes still persist. Will correct it.

app/assets/javascripts/reducers/campaigns.js
vaidehi44: Yeah, right. But, I thought by just correcting this in the `all_campaigns` variable in the `campaign_editable.spec.jsx` would work. But it says that `all_campaigns` is undefined. So, I am still trying to figure out a way to resolve this.

ragesoss: If you're looking for ways to debug this and understand what is happening with Redux state, I suggest (a) adding lots of console logging to see what the data looks like at each place in the flow, and (b) installing ReduxDevTools if you haven't already, to be able to inspect the Redux state at any time and see all the redux actions that transformed it.

vaidehi44: Have resolved this one. Will push the code soon.

app/assets/javascripts/components/course/course_approval.jsx
vaidehi44: @ragesoss, is it okay to place this function inside `utils` folder in `app/assets/javascripts/components/course/` ?

ragesoss: That sounds like the right place for it.

app/assets/javascripts/reducers/campaigns.js
vaidehi44: Basically, here the issue was with the `all_campaigns` state in `campaign_expandable.spec.jsx`. `allCampaigns` was being passed as a prop in `campaignEditable` component. But that was not even required as it is not a defined prop for it. I just correctly set the `all_campaigns` array in the initial state and the error was gone.

vaidehi44:
Hi @ragesoss, I have resolved the js test error and have also done the changes you told in the review. Other than that, I did some minor additions like disabling the approve button if a campaign is not selected. I think its ready for thorough review from your side.
Let me know if I need to correct or update anything.

ragesoss:
Great! I'll review this soon.

ragesoss:
The user experience for this is all set, and I haven't noticed any problems with the code. Nice work!

In testing it locally, I noticed one time (but could not replicate it) that it threw a console warning about useState called for an unmounted component — I think after I made some changes to the form and then submitted it. So I think there's some cleanup in the hooks configuration that needs to be added, but I'm not sure where.

I'd like this to include a few tests:
* One feature of the approval workflow, to test using the Approve Course button and ensuring that it adds the users and campaigns and tags as expected.
* A set of tests for `inferDefaultCampaign` to provide examples of dates and the expected default campaign for each. (These will be really helpful for if/when we change that date behavior, to document what it's supposed to do, since it has some potentially counter-intuitive behavior like having a December 2021 start date mean `spring_2022`.)

vaidehi44:
Okay. I think I also got the unmounted state error, will look into it. Will also start working on the tests you mentioned.

ragesoss:
@vaidehi44 are you still working on the tests for this, and were you able to replicate that error?

vaidehi44:
@ragesoss, I could replicate the error and have also fixed it. But, I am stuck on the tests.

ragesoss:
@vaidehi44 what was the cause of that error, and what fixed it?

test/utils/infer_default_campaign.spec.js
```diff
@@ -0,0 +1,61 @@
...
+
+describe('No matching campaign', () => {
+  test('returns null', () => {
+    start_date = moment([2019, 2, 5]);
+    const result = inferDefaultCampaign(campaigns, start_date);
+    expect(result).toBe(null);
+  });
+});
+
+describe('Matching campaigns available', () => {
```
ragesoss, line 19: This is a nice set of tests.

vaidehi44:
> @vaidehi44 what was the cause of that error, and what fixed it?

@ragesoss it seems that the reason was state not being cleared before component unmouting. So, I used a cleanup function in useEffect hook. Here - [5517a30](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4913/commits/5517a30e41a527f53ddd2be6e46f09e82ea7892c)

test/utils/infer_default_campaign.spec.js
vaidehi44: Okay. What about the feature tests? Are they fine?

ragesoss: Yes, those look good as well.

vaidehi44:
Thanks @ragesoss for merging this. Took a lot of time to complete it :).

vaidehi44:
> Thanks for sticking with it! Our staff is really excited about it; I deployed it right after merging it yesterday.

Happy to hear this. :)

Issue #4911, opened 2022-04-06
Create a way for admins to copy courses from a production instance of the Dashboard


ragesoss:
We have a script for copying the core data of one program on a public instance of the Dashboard to another instance: `setup/copy_course_from_production.rb`. However this can only be used manually from a rails console.

A good step forward would be to add a controller and very simple frontend for this functionality, so that it's really easy for a developer to copy a course from production for testing.

This will also require a thorough refactor of the script into something that looks more like one of the `app/services` files, along with comprehensive tests for the controller and service.

One bug to fix and test for is the conversion of the course `flags` has. We use integer keys within the `update_logs` portion of the flags, but these get converted into strings when the go through JSON. (And then, this causes an error if you run an update on the same course.)

anirudh-t-d:
Hi @ragesoss I would like to start working on this. But I am new to this, so it might take me a bit of time.

enelesmai:
Hello @ragesoss do you think I can work on this issue? Or should I take another one? I'm an applicant for outreachy program

enelesmai:
> Hi @ragesoss I would like to start working on this. But I am new to this, so it might take me a bit of time.

Sorry I didn't saw your message and started to work on it, but we can collaborate and work in the same branch if you want. Maybe I can guide you :)

enelesmai:
> > Hi @ragesoss I would like to start working on this. But I am new to this, so it might take me a bit of time.
> 
> I'm sorry @anirudh-t-d I didn't saw your message before and started to work on it, but we can collaborate and work in the same branch if you want. Maybe I can guide you :)

anirudh-t-d:
It's ok @enelesmai I couldn't start to work on it as I had to take care of my Mum. You can work on it. I hope things will get better by tomorrow and work on another issue. And @ragesoss can you please add me to the slack channel?

Kingstalux:
Hello @ragesoss , since this issue has not yet been resolved can I start working on it too?

marklocklear:
In case others stumble across this and its not clear, there is a PR addressing this issue here https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4914

gabina:
Can I work on this? I see there is already a (closed) PR for this, maybe I can take it as a guide.

ragesoss:
Go for it!

PR #4914, opened 2022-04-07
Create a way for admins to copy courses from a production instance of the Dashboard

enelesmai:

Issue: #4911 

## What this PR does
We have a script for copying the core data of one program on a public instance of the Dashboard to another instance: `setup/copy_course_from_production.rb`. However this can only be used manually from a rails console.

A good step forward would be to add a controller and very simple frontend for this functionality, so that it's really easy for a developer to copy a course from production for testing.

This will also require a thorough refactor of the script into something that looks more like one of the `app/services `files, along with comprehensive tests for the controller and service.

## Screenshots
Before:
image

After:
image

image

** Usage:

Copy the url to the text box:
image

Result:
image

image

* Run added test cases
**`rspec ./spec/services/copy_course_from_production_spec.rb`
**`rspec ./spec/controllers/copy_course_from_production_controller_spec.rb`

An url example is added to test. VCR.use_cassette is used to perform HTTP request. The test cases expect to validate the information is loaded from production and copied to the local db for test.

>>

### DIFF
--- a/app/controllers/copy_course_from_production_controller.rb
+++ b/app/controllers/copy_course_from_production_controller.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+#= Controller for copying course from production to test
+class CopyCourseFromProductionController < ApplicationController
+    respond_to :html
+    before_action :require_permissions
+
+    def index
+        @created_course = ""
+    end
+
+    def create
+        copied_course = CopyCourseFromProduction.new(params[:url])
+        pp 'Course created!'
+        pp "http://localhost:3000/courses/#{copied_course.course_slug}"
+        redirect_to copy_course_from_production_path, 
+            locals: { created_course: copied_course.course_slug },
+            notice: "Course #{copied_course.course_slug} was created."
+    end
+end
--- a/app/controllers/copy_course_from_production_controller.rb
+++ b/app/controllers/copy_course_from_production_controller.rb
@@ -11,10 +11,9 @@ class CopyCourseFromProductionController < ApplicationController
 
   def copy
     copied_course = CopyCourseFromProduction.new(params[:url])
-    pp 'Course created!'
     pp "http://localhost:3000/courses/#{copied_course.course_slug}"
     redirect_to copy_course_from_production_path,
-                locals: { created_course: copied_course.course_slug },
-                notice: "Course #{copied_course.course_slug} was created."
+                locals: { copied_course: copied_course.course_slug },
+                notice: "Course #{copied_course.title} was created. http://localhost:3000/courses/#{copied_course.course_slug}"
   end
 end
--- a/app/controllers/copy_course_from_production_controller.rb
+++ b/app/controllers/copy_course_from_production_controller.rb
@@ -11,7 +11,6 @@ class CopyCourseFromProductionController < ApplicationController
 
   def copy
     copied_course = CopyCourseFromProduction.new(params[:url])
-    pp "http://localhost:3000/courses/#{copied_course.course_slug}"
     redirect_to copy_course_from_production_path,
                 locals: { copied_course: copied_course.course_slug },
                 notice: "Course #{copied_course.title} was created. http://localhost:3000/courses/#{copied_course.course_slug}"
--- a/app/controllers/copy_course_from_production_controller.rb
+++ b/app/controllers/copy_course_from_production_controller.rb
@@ -2,19 +2,19 @@
 
 #= Controller for copying course from production to test
 class CopyCourseFromProductionController < ApplicationController
-    respond_to :html
-    before_action :require_permissions
+  respond_to :html
+  before_action :require_permissions
 
-    def index
-        @created_course = ""
-    end
+  def index
+    @created_course = ''
+  end
 
-    def create
-        copied_course = CopyCourseFromProduction.new(params[:url])
-        pp 'Course created!'
-        pp "http://localhost:3000/courses/#{copied_course.course_slug}"
-        redirect_to copy_course_from_production_path, 
-            locals: { created_course: copied_course.course_slug },
-            notice: "Course #{copied_course.course_slug} was created."
-    end
+  def copy
+    copied_course = CopyCourseFromProduction.new(params[:url])
+    pp 'Course created!'
+    pp "http://localhost:3000/courses/#{copied_course.course_slug}"
+    redirect_to copy_course_from_production_path,
+                locals: { created_course: copied_course.course_slug },
+                notice: "Course #{copied_course.course_slug} was created."
+  end
 end
--- a/app/controllers/copy_course_from_production_controller.rb
+++ b/app/controllers/copy_course_from_production_controller.rb
@@ -6,7 +6,7 @@ class CopyCourseFromProductionController < ApplicationController
   before_action :require_admin_permissions
 
   def index
-    
+    @copied_course = nil
   end
 
   def copy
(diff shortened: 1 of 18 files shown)


enelesmai:
I have a question here @ragesoss, for the frontend is it ok only adding a textbox input for the url? and then a button to execute the service? or do you suggest something else?

ragesoss:
Yes, a simple text input is fine.

enelesmai:
Hello @ragesoss  I have one more question: Where should I put the link to access to this page?

ragesoss:
I guess it should link from /admin .

enelesmai:
> I guess it should link from /admin .

I was exploring and found that each course have a list of actions, and there is a similar functionality called ´Clone course´. I think I can add the option to copy the course on this list, and do the necessary changes so we don't have to use the url to copy the course and so this functionality will be easier to use. What do you think about it?

image

ragesoss:
What you're showing in this screenshot is the 'Actions' buttons for an
individual course. To use this, you would need to first set up a course
that matched one on production. It will be much easier to use a URL to copy
a course, without already having a course page for it in your dev
environment.

-Sage

enelesmai:
> 

That's right I was thinking about that, it was just an idea but it will need a lot of changes like the way we get the course data and the validation to show this option only to admins and to production courses, in addition to it could be used in production only. So I'm just adding the new page, hehe. ✌

enelesmai:
Hello @ragesoss, I opened this PR for review

app/controllers/copy_course_from_production_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  respond_to :html
+  before_action :require_admin_permissions
+
+  def index
+    @copied_course = nil
+  end
+
+  def copy
+    copied_course = CopyCourseFromProduction.new(params[:url])
+    pp "http://localhost:3000/courses/#{copied_course.course_slug}"
```
ragesoss, line 14: We shouldn't have any `pp` output in the production rails code.

app/views/admin/index.html.haml
```diff
@@ -20,6 +20,8 @@
       %a{href: '/sidekiq'} Job queue
     %li
       %a{href: 'https://dashboard-testing.wikiedu.org/rails/mailers'} Email templates
+    %li
+      %a{href: '/course_copy_prod'} Copy course from production
```
ragesoss, line 24: Probably a better label for this will be "Copy course from another Dashboard server"

spec/controllers/copy_course_from_production_controller_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        context 'when the user is an admin' do
+            before do
+                allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(admin)
+            end
+
+            it 'copies the course' do
+                subject
+                expect(response.status).to eq(200)
+                expect(Course.exists?(:title => existent_prod_course)).to eq(true)
```
ragesoss, line 23: This doesn't make sense, as `existent_prod_course` is a URL string, not a course title (which in this case, is "Invertebrate Zoology").

ragesoss: This probably also needs to stub `CopyCourseFromProduction`, as we don't want to hit a production server more than once at most in the test suite. (We don't currently hit production dashboard servers at all in the test suite, and doing so makes me just a little bit uneasy, but I think it makes sense for a test of the service itself.)

spec/services/copy_course_from_production_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+describe CopyCourseFromProduction do
+  let(:existent_prod_course) { 'https://dashboard.wikiedu.org/courses/University_of_California_Merced/Feminism,_Handmaids_and_Wild_Seeds_(Spring_2021)' }
+
+  context 'with courses in production' do
+    let(:subject) do
+      described_class.new(existent_prod_course)
+    end
+
+    it 'copy course to dev env' do
```
ragesoss, line 13: This should test more than just the returned `result`. It should verify that examples of each of the important types of data got copied over as well: students, instructors, and assignments.

app/controllers/copy_course_from_production_controller.rb
enelesmai: Removed here [ecd6486](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4914/commits/ecd6486288eafd4dff4518664b57ded7d7360153)

app/views/admin/index.html.haml
enelesmai: Updated here [82d04a6](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4914/commits/82d04a6c1dbb8c370b5f958e0134a994b43b98ef)

spec/controllers/copy_course_from_production_controller_spec.rb
enelesmai: I did the change using slug property to check the existence [d75a150](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4914/commits/d75a1504d60105e3a1ad5da485c735e90c16cd37)

enelesmai: I moved the test case to check service only from CopyCourseFromProduction [d49f462](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4914/commits/d49f46200acd3cf91310105c9ff01dd8c7c77940)

spec/services/copy_course_from_production_spec.rb
enelesmai: Do you think this kind of examples are enough? [6e16b75](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4914/commits/6e16b75e308a948ef00753b658878b351947d7b6)

ragesoss: Looks good.

spec/services/copy_course_from_production_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+describe CopyCourseFromProduction do
+  let(:url_base) { 'https://dashboard.wikiedu.org/courses/' }
+  let(:existent_prod_course_slug) do
+    'University_of_South_Carolina/Invertebrate_Zoology_(Spring_2022)'
+  end
+
+  context 'with courses in production' do
+    let(:subject) do
+      described_class.new(url_base + existent_prod_course_slug)
```
ragesoss, line 13: Since these tests make external web requests from the server, you need to use VCR to allow those requests to go through in the test environment.

enelesmai: @ragesoss I just realize about this comment, in the morning I added de VCR.use_cassette block to each test case, do you think is ok there?

ragesoss: The way you did it should be fine.

ragesoss:
Are these specs passing for you locally?

enelesmai:
> Are these specs passing for you locally?
I had my environment kind of broken but right now I already fixed it, I added some fixes but the specs are not showing anything 🤔

image

ragesoss:
>  I had my environment kind of broken but right now I already fixed it, I added some fixes but the specs are not showing anything thinking
> 
> image

Hmm... weird. I haven't tried running these locally yet, but they have been failing in CI. But I would expect them to return something when you run them locally, just like they do on CI.

enelesmai:
Only one try more, I'm still fixing my rspec installation 😰

ragesoss:
@enelesmai the one failing test is unrelated to this PR, I think. Is this PR ready to go from your perspective, or are there other changes you're still working on?

PR #4917, opened 2022-04-07
Scroll to top when a student is selected in the Students Tab

TheTrio:
This also fixes the issue where the selected user wasn't being highlighted. 

I've had to duplicate a route for `studentSelection` since that component is rendered both when a user isn't selected and when it is.

### DIFF
--- a/app/assets/javascripts/components/util/ScrollToTop.jsx
+++ b/app/assets/javascripts/components/util/ScrollToTop.jsx
@@ -1,0 +1,12 @@
+import { useEffect } from 'react';
+
+const ScrollToTop = () => {
+  useEffect(() => {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/util/ScrollToTop.jsx
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from 'react';
+
+const ScrollToTop = () => {
```
ragesoss, line 3: This one only scrolls to top when it mounts, right? As opposed to the `ScrollToTop` example from the docs that scrolls any time the route changes? If so, using the same naming as the docs probably makes sense for clarity.

Issue #4912, opened 2022-04-06
In dev build, console is flooded with errors related to source maps


ragesoss:
### What is happening?

When the dev build of the assets (`yarn start`) are being used, when you load the dashboard locally, the console is flooded with errors like this:

Chrome: `DevTools failed to load source map: Could not load content for webpack:///node_modules/react-day-picker/build/index.js.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME`

Firefox: `Source map error: TypeError: LocaleUtils.js.map is not a valid URL.
Resource URL: webpack:///./node_modules/react-day-picker/build/LocaleUtils.js?
Source Map URL: LocaleUtils.js.map[Learn More]`

These are likely related to our webpack configuration along with the Redux DevTools and/or React Development Tools browser plugins, but it would be nice to fix whatever is going wrong.

PR #4918, opened 2022-04-08
fix source maps in development

TheTrio:
## What this PR does

Addresses #4912 

Relevant: [Devtool - webpack](https://v4.webpack.js.org/configuration/devtool/)

## Screenshots
Before:
ksnip_20220408-112707

After:

ksnip_20220408-112803


ragesoss:
Thanks! Does this significantly change compile or recompile time? Now that I think of it, compile time may have been the reason why we used the `eval` option there. What other options are there, and what is their significance?

ragesoss:
I followed the link and found most of the answers... I'll test how much this changes compile times.

ragesoss:
Seems to make only a small difference (from 8 seconds to 9 seconds for `yarn start` JS compilation, no major change in recompile time of about half a second).

PR #4922, opened 2022-04-10
disable WebpackRTLPlugin in development

TheTrio:
This provides some performance improvements 

# Before
ksnip_20220410-222750

# After
ksnip_20220410-222657

### DIFF
--- a/css.webpack.js
+++ b/css.webpack.js
@@ -72,7 +72,7 @@ module.exports = (env) => {
       }),
       // generates a RTL version of the emitted CSS files
       // this is done only in production
-      ...(env.production ? [new WebpackRTLPlugin({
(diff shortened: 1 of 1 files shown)


TheTrio:
Not sure why the build is failing - this shouldn't affect the production build at all. In fact, the test runs file on my system - 
image

The first test isn't even related to my change so :confused:

ragesoss:
I think one of the test failures was a random one unrelated to the change, but the `update_locle_spec.rb` one seems like it might be connected to this change. I haven't seen that one fail before, and it relates to generated CSS files.

ragesoss:
In particular, that failing test is for changing to an RTL language and verifying that i18n is working.

Issue #4916, opened 2022-04-07
Update article_finder_actions.js and related files with accurate variable names


ragesoss:
I noticed while reviewing #4905 that in `article_finder_actions.js` and related files that use it, `fetchKeywordResults` and some other functions take a `home_wiki` parameter. However, wiki object passed to it are not necessarily the home wiki for a course. The variable names are misleading, and this code will be easier to understand and modify if variable and prop names are updated where appropriate. `home_wiki` should only be used in situations where it actually is the home_wiki property of a course. Otherwise, `wiki` will be more general and accurate.

anirudh-t-d:
@ragesoss I would like to work on this

PR #4926, opened 2022-04-11
Renames home_wiki in article finder functions and corresponding files

Sahethi:
## What this PR does 
Aims to close the issue #4916
Changes were made in `article_finder.jsx` and `article_finder_action.js`

- In `article_finder_action.js`, the name of the variable was `home_wiki` which is not accurate cause it's not always the case that `home_wiki` of the course is passed to it.
- In the case of `article_finder.jsx` where the elements are being rendered, functions are imported from `article_finder_action.js` and here `home_wiki` of courses is passed in `updateFields()` where it refers the same by `this.props.course.home_wiki` and in few functions like `fetchCategoryResults()` & `fetchKeywordResults()`.
- And in cases where updateFields is being passed `wiki`, you can see that it takes its value parameter not based on `courses.home_wiki`

 
I have renamed `home_wiki` to `wiki` where I thought `home_wiki` of `courses` wasn't being referred to. Please review and let me know.


ragesoss:
I'm going to hold off on reviewing or merging this until #4905 is finished, so it doesn't cause any conflicts.

Sahethi:
> I'm going to hold off on reviewing or merging this until #4905 is finished, so it doesn't cause any conflicts.

Okay no problem 👍

ragesoss:
@Sahethi okay, it's ready for you to update to reflect the newly merged #4905. :)

ragesoss:
It would be useful to include a note (in the PR description) about where in the code the dividing line is between when it's a home_wiki and when it's just a wiki. As in, where are the places that a course home_wiki actually does get passed to one of these functions?

That will help us think more clearly about whether this change makes sense and why a variable might be `home_wiki` elsewhere but then become just `wiki` when passed to one of these functions.

Sahethi:
Yes, I'll do that @ragesoss

Sahethi:
@ragesoss Heyy, are the changes good enough?

Issue #4873, opened 2022-03-21
Cannot directly load student view for username that include a period


ragesoss:
The server returns a 404 for this URL: https://dashboard.wikiedu.org/courses/University_of_Ottawa/CMN2160C_(Winter)/students/articles/Heba.Aweiwi

Other usernames in the same course work fine: https://dashboard.wikiedu.org/courses/University_of_Ottawa/CMN2160C_(Winter)/students/articles/Nabila%20Bashamuher

Navigating to that user from an already-loaded course page (such as the second link) also work fine, so the problem is likely with the Rails route.

enelesmai:
Hello @ragesoss I would like to work on this issue, I will start the research

ragesoss:
@enelesmai how is it going? Give me a ping if you need any help.

buildwithallan:
Hi @ragesoss I will like to work on this if it's still available.

ragesoss:
@buildwithallan sure. can you summarize what you've researched and what ideas you have so far?

buildwithallan:
@ragesoss  Okay. I read that if you want to support a url parameter more complex than an id number, you may run into trouble with the parser if the value contains a period. Anything following a period will be assumed to be a format(i.e. json)

If you want to support a url parameter more complex than an id number, you may run into trouble with the parser if the value contains a period. Anything following a period will be assumed to be a format (i.e. json, xml). For it to work, you will have to add a constraint to broaden the accepted input.

Example: `resources :users, constraints: { id: /.*/ } 
`

ragesoss:
Good, it sounds like you are on the right track. A good next step would be to write some tests for the examples in this issue. Doing that may help you pin down which route lacks the `constraints` it needs to work in this situation.

ragesoss:
You can see that we already use this `constraints: { id: /.*/ }` approach for many routes in `routes.rb`, so it might be easy to fix once you find the correct route.

buildwithallan:
@ragesoss  Sure I will do that.

zishugshan:
initially :

```js
get 'courses/:school/:titleterm(/:_subpage(/:_subsubpage(/:_subsubsubpage)))' => 'courses#show',
        :as => 'show',
        constraints: {
          school: /[^\/]*/,
          titleterm: /[^\/]*/
}
```

mysuggestion :
```js
[...]

PR #4927, opened 2022-04-11
Load student view for username that include a  period #4873

buildwithallan:
## What this PR does
This PR is trying to fix issue #4873

Before: The server returns 404 for this url: http://localhost:3000/courses/UCC/University_of_Cape_Coast_Art_and_Feminism_2022_(2022)/students/articles/Allan.Aikins

After: The server returns 200 for this url: http://localhost:3000/courses/UCC/University_of_Cape_Coast_Art_and_Feminism_2022_(2022)/students/articles/Allan.Aikins

### DIFF
--- a/spec/features/students_page_spec.rb
+++ b/spec/features/students_page_spec.rb
@@ -35,7 +35,7 @@ describe 'Students Page', type: :feature, js: true do
                           real_name: 'Mr. Tester',
                           trained: true)
 
-    @user1 = create(:user, username: 'Allan.Aikins',
+    @user_with_period = create(:user, username: 'Allan.Aikins',
                           real_name: 'Allan Aikins',
                           trained: true)
 
@@ -44,6 +44,12 @@ describe 'Students Page', type: :feature, js: true do
            course_id: @course.id,
            user_id: @user.id,
            real_name: @user.real_name)
+
+    create(:courses_user,
+           id: 2,
+           course_id: @course.id,
+           user_id: @user_with_period.id,
+           real_name: @user_with_period.real_name)
 
     article = create(:article,
                      id: 1,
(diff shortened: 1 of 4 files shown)


ragesoss:
This PR should not include any changes to the schema.rb or Gemfile.lock files.

spec/features/students_page_spec.rb
```diff
@@ -68,6 +72,23 @@ def js_visit(path)
     expect(page).to have_content @user.username
   end
 
+
+  describe 'students view for user names' do
+    let(:user) { create(:user) }
+
+    it 'loads students view for user names without period' do
+       js_visit "/courses/#{@course.slug}/students/articles/#{@user.username}"
```
ragesoss, line 80: Is it necessary to use `js_visit` and the `sleep 1` after it, or will these tests pass with just `visit`?

This is one of the only tests that has hard-coded `sleep` instead of relying on Capybara to automatically wait until the looked-for elements are on screen, and it might not be necessary anymore.

spec/features/students_page_spec.rb
```diff
@@ -35,6 +35,10 @@ def js_visit(path)
                           real_name: 'Mr. Tester',
                           trained: true)
 
+    @user1 = create(:user, username: 'Allan.Aikins',
```
ragesoss, line 38: This would be better if the variable reflected the way it's being used, like `@user_with_period`.

spec/features/students_page_spec.rb
buildwithallan: > This PR should not include any changes to the schema.rb or Gemfile.lock files.

okay

buildwithallan: > Is it necessary to use `js_visit` and the `sleep 1` after it, or will these tests pass with just `visit`?
> 
> This is one of the only tests that has hard-coded `sleep` instead of relying on Capybara to automatically wait until the looked-for elements are on screen, and it might not be necessary anymore.

Thanks for the info. I am changing it

spec/features/students_page_spec.rb
buildwithallan: Something like this? ` @user1 = create(:user, username: @user_with_period)` ?

ragesoss: No, like `@user_with_period = create(...`

PR #4934, opened 2022-04-14
Add progress bar for references and page assessments in activity tab

TheTrio:
## What this PR does
Adds a progress bar in the activity page for when the references and page assessments are being loaded. 

## Screenshots
demo

image


app/assets/javascripts/components/revisions/revisions_handler.jsx
```diff
@@ -115,6 +124,15 @@ const RevisionHandler = createReactClass({
         />
         {!loaded && <Loading/>}
         {loaded && showMoreButton}
+        {loaded && metaDataLoading && (
+          <div id="progress" >
```
ragesoss, line 128: You should extract this as a new component that takes the loading message as its prop, so that we can easily re-use this in other places too.

ragesoss:
This works really smoothly.

PR #4935, opened 2022-04-14
Extracted enroll to another controller

TheTrio:
## What this PR does

Addresses #1949

I think this is functionally equivalent, but I don't know if I'm following the right patterns. Followed [Getting Started with Rails](https://guides.rubyonrails.org/getting_started.html) to get a basic idea of everything and then some googling.


ragesoss:
This looks great. The tests are passing, so I'm pretty confident it didn't break anything. I didn't realize how little would be left in USersController.

Issue #715, opened 2016-03-11
In Students, Articles and Activity views, 'words added' should be used instead of chars.


ragesoss:
Fix this should also make #159 irrelevant.

ragesoss:
@aSquare14 this issue wasn't clear enough. In order to switch from characters added to words added, we also need to be able to convert the values from characters to words. We do this for English via `lib/word_count.rb`.

aSquare14:
I still don't get it ! Characters to words means like converting an array of characters to a string ? Is that what it means ?

ragesoss:
We store data about the wiki edits that people make in terms of characters, because that's how mediawiki does it. For example, [this edit](https://en.wikipedia.org/w/index.php?title=Okenia_japonica&diff=775856563&oldid=775777495) added 12 characters to the article.

However, it's easier to understand those stats in terms of "words added". So we convert from characters to words using an average characters-per-word value, and say that 12 characters is ~ 2 words of content added.

bwreid:
It might help here to look at the `word_count.rb` file in implementing this feature.

zishugshan:
hey @ragesoss 
I sort it out , I tested it in my localhost, it working fine .

PR #4936, opened 2022-04-15
Num of char to num word human readable form

zishugshan:
## It is solving the issue :
 https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/715

## What this PR does
- [x] num of character is not relevant many time.
- [x] so converting number of char to number of word is needed.
- [x] I changed the num of char to num of words.

## Screenshots
Before:
Screenshot 2022-04-15 at 7 30 25 PM

After:
Screenshot 2022-04-15 at 6 03 30 PM

### DIFF
--- a/app/views/campaigns/_article_row.html.haml
+++ b/app/views/campaigns/_article_row.html.haml
@@ -3,7 +3,7 @@
     = link_to article.full_title, article.url
     = t('articles.deleted') if article.deleted
   %td.char_added
-    = number_to_human article_course.character_sum/7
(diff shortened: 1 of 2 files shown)


zishugshan:
hey @ragesoss 
please review this change);

ragesoss:
Per my comment on the other PR, I don't think this is a viable way to fix the "char" issue, because the conversion rate varies from wiki to wiki and we don't have data for any other wikis but English Wikipedia.

PR #4937, opened 2022-04-15
View count human readable form

zishugshan:
## It is solving the issue :
 in campaign article page View count is converted into human readable form  ==> (47997 == 47.997k)

## What this PR does
- [x] big numbers  of views is not relevant here.
- [x] so converting big number of views to  human readable form.
- [x] I changed the number of views to human readable form-->(47997 == 47.997k)

## Screenshots
Before:
Screenshot 2022-04-15 at 7 30 25 PM

After:
Screenshot 2022-04-16 at 10 34 43 PM

### DIFF
--- a/app/views/campaigns/_article_row.html.haml
+++ b/app/views/campaigns/_article_row.html.haml
@@ -3,7 +3,7 @@
     = link_to article.full_title, article.url
     = t('articles.deleted') if article.deleted
   %td.char_added
-    = number_to_human article_course.character_sum/7
+    = article_course.character_sum
   %td.references
     = article_course.references_count
   %td.views
--- a/app/views/campaigns/_article_row.html.haml
+++ b/app/views/campaigns/_article_row.html.haml
@@ -3,11 +3,11 @@
     = link_to article.full_title, article.url
     = t('articles.deleted') if article.deleted
   %td.char_added
-    = article_course.character_sum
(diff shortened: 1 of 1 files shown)


zishugshan:
hey @ragesoss 
please review this change is this good to go

ragesoss:
Unfortunately, we can't fix this so simply, as the conversion between bytes/characters and words is language-dependent. The 5.175 value is an estimate specific to English Wikipedia, but the Campaign Articles list can include a mixture of articles from different language Wikipedias and other non-Wikipedia wikis.

ragesoss:
It looks like you've added the 'view count' change on top of the change from your other PR. They should be independent.

zishugshan:
> It looks like you've added the 'view count' change on top of the change from your other PR. They should be independent.

okey I will make it independent

zishugshan:
> Unfortunately, we can't fix this so simply, as the conversion between bytes/characters and words is language-dependent. The 5.175 value is an estimate specific to English Wikipedia, but the Campaign Articles list can include a mixture of articles from different language Wikipedias and other non-Wikipedia wikis.

How should I go about it , any further suggestion , how can I solve this problem.

zishugshan:
I made 

> It looks like you've added the 'view count' change on top of the change from your other PR. They should be independent.

I made the desired changes in the PR please review it

Issue #3259, opened 2019-09-05
There should be a way to disable student emails for an individual course


ragesoss:
Occasionally an instructor wants to stop any reminder emails for trainings to be sent to students.

We could perhaps implement it as a course flag, which would get checked before the training reminder emails go out.

marklocklear:
A couple of notes on this one:

- add `disable_training_emails` boolean flag to course.rb model
- add logic to `overdue_training_alert.rb` with something like:
```
def send_email
    return if emails_disabled?
    return if course.disable_training_emails?
    return if opted_out?
    OverdueTrainingAlertMailer.send_email(self)
  end
```

- Add drop-down selection to Details page of a course allowing the instructor to disable training emails for students

@ragesoss other alerts/mailers I should consider for this feature?

ragesoss:
I think think there are other alerts that should be wrapped in as well (so, the flag and selector shouldn't be specific to training emails).

When an instructor wants to turn off emails for their students, they typically want to try to ensure that NO email from the dashboard reaches their students at all. The relevant one that I spot from `/rails/mailers` is: `High Quality Article Assignment Mailer`. I imagine there would be others in the future; we don't currently send very many emails to students, but I think we will add more over time.

I'm not sure what the right way is to capture in code the idea of "this set of emails should not go out to students, if their course has student emails disabled".

marklocklear:
Ah, I see. Something more abstract would be adding a boolean `disabled` field to the alert model. Each course would have an instance of each of its alerts (OverdueTrainingAlert, ProductiveCourseAlert...etc) then you could enable/disable each of those alerts if you like. Default would be true, then in each of the alert models (ie class OverdueTrainingAlert < Alert) we could call `return if self.disabled?` in the `send_email` method.

Just FYI I do see:

``` 
  def emails_disabled?
    ENV["#{self.class}_emails_disabled"] == 'true'
  end
```

...in alert.rb, but this is system-level email disabling for all courses?

ragesoss:
Correct, that is code you found is for system-level disabling of emails for an alert type.

The `flags` hash for an individual course is probably the right place for this to live. I don't think we want to disable alert-types one-by-one at the course level, just one toggle for whether students should receive any course-related emails or not.

marklocklear:
[...]

PR #4945, opened 2022-04-21
Disable email for course

marklocklear:
## What this PR does
https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/3259 Wiring that adds a flag to course (serialized) for disabling student email for an individual course and also toggle in course details page.

## Screenshots
After:

image

Only added a single check for the flag in app/models/alert_types/overdue_training_alert.rb but should also add to other alerts where we would want this to apply

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -591,7 +591,7 @@ en:
     delete_suggestion: Delete
     delete_course_instructions: You may delete this page if all campaigns are removed first.
     disable_student_emails: Disable Student Emails
-    disable_student_emails_tooltip: Disabling student email will stop all email being sent to students associated with this course.
+    disable_student_emails_tooltip: Disabling student email will stop some notification emails from being sent to students for this course.
     dismiss_survey: Dismiss
     dismiss_survey_confirm: Are you sure you want to permanently dismiss this survey?
     download_stats_data: Download stats
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -590,6 +590,8 @@ en:
     delete_course: Delete course
     delete_suggestion: Delete
     delete_course_instructions: You may delete this page if all campaigns are removed first.
+    disable_student_emails: Disable Student Emails
+    disable_student_emails_tooltip: Disabling student email will stop all email being sent to students associated with this course.
     dismiss_survey: Dismiss
     dismiss_survey_confirm: Are you sure you want to permanently dismiss this survey?
     download_stats_data: Download stats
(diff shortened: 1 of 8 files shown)


ragesoss:
Thanks @marklocklear! I expect to review this week after next, as I'll be on vacation next week.

ragesoss:
The code for this looks good. It needs a test of the toggling functionality, either by adding an example to `courses_controller_spec.rb` (following the existing examples for "toggling timeline"), or by adding a browser-based test that exercises the feature.

marklocklear:
Added tests here 49bbf46

spec/controllers/courses_controller_spec.rb
```diff
@@ -218,6 +218,22 @@
       end
     end
 
+    describe 'toggling disable student emails' do
+      it 'sets the disable student emails flag to true' do
+        expect(course.flags[:disable_student_emails]).to be_nil
+        params = { id: course.slug, course: { disable_student_emails: true } }
+        put "/courses/#{course.slug}", params: params, as: :json
+        expect(course.reload.flags[:disable_student_emails]).to eq(true)
```
ragesoss, line 226: It would be slightly better if these tests use `course.disable_student_emails?` to rather than `course.flags`. That way, how `.disable_student_emails?` works is an implementation detail that could change without affecting the tests.

ragesoss:
With the above minor changes, this will be ready to merge. Thanks!

config/locales/en.yml
```diff
@@ -590,6 +590,8 @@ en:
     delete_course: Delete course
     delete_suggestion: Delete
     delete_course_instructions: You may delete this page if all campaigns are removed first.
+    disable_student_emails: Disable Student Emails
+    disable_student_emails_tooltip: Disabling student email will stop all email being sent to students associated with this course.
```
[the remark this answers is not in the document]
marklocklear: updated with https://github.com/marklocklear/WikiEduDashboard/commit/7dc9274b7214c23e2c5f61228dc2af7d6dce5d96

spec/controllers/courses_controller_spec.rb
marklocklear: Makes sense. Updated with https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/a77ee30ca19af83b737810da49082f42d643f08c Note I changed the expected initial state from nil to false. Let me know if that's not right.

ragesoss: Perfect. As a setting, it only makes sense for it to be either true or false and to default to false, so that's how I would have done it too.

PR #4946, opened 2022-04-21
Implement client side fetching for course scoped revisions

TheTrio:
# What this PR does

#4886 implemented client side fetching for revisions which weren't course scoped. This PR attempts to do the same for course scoped revisions.

I haven't tested it on courses with a large number of tracked articles but it does seem to work for the dummy courses that I've setup. 

Presently, I've just set the article limit to 500 but I'm not sure if that's reasonable or not. 

demo

~~One possible issue might be that currently the revisions are filtered after being fetched. For example, say we fetch 50 revisions but none of them are tracked. Until the user clicks on show more, new revisions won't be fetched and the user will see an empty list~~

~~Essentially what it boils down to is that unlike for non course scoped revisions(where we keep fetching until we get 50 revisions), there's no such minimum value for course scoped revisions~~

This has now been changed so that the filtering happens **while** the articles are being fetched. This means that we keep fetching until we find 50 tracked revisions or go beyond the course start date.


ragesoss:
@TheTrio I'll review this probably week after next, as I'm on vacation next week.

app/assets/javascripts/actions/revisions_actions.js
```diff
@@ -18,15 +21,16 @@ import { sortRevisionsByDate } from '../utils/revision_utils';
 import { INCREASE_LIMIT } from '../constants/revisions';
 import { STUDENT_ROLE } from '../constants/user_roles';
 
-const fetchCourseScopedRevisionsPromise = async (course, limit) => {
-  const response = await request(`/courses/${course.slug}/revisions.json?limit=${limit}&course_scoped=true`);
+const fetchAllArticles = async (course) => {
+  const response = await request(`/courses/${course.slug}/articles.json?limit=500`);
```
ragesoss, line 25: I think this limit is inevitably going to be confusing in some edge cases with large numbers of articles.

I think perhaps the solution could be to disable course-specific edits altogether for courses with too many articles to safely do this. My suggestion is to keep the limit, but don't show the `Course Specific` button at all if the course's article count is over 500 (which you can check via the `course` object without needing to query articles.json).

TheTrio: Hmm I guess that's one way to do it. Do you think the limit should be increased though (from the current 500)?

ragesoss: I think that limit is fine.

app/assets/javascripts/components/revisions/revisions_handler.jsx
```diff
@@ -113,12 +113,15 @@ const RevisionHandler = createReactClass({
...
+
+    // we only fetch articles data for a max of 500 articles(for course specific revisions).
+    // If there are more than 500 articles, the toggle button is not shown
     const revisionFilterButton = <div><button className="button ghost stacked right" onClick={this.toggleCourseSpecific}>{this.revisionFilterButtonText()}</button></div>;
     return (
       <div id="revisions">
         <div className="section-header">
           <h3>{I18n.t('application.recent_activity')}</h3>
-          {revisionFilterButton}
+          {this.props.course.article_count <= 500 && revisionFilterButton}
```
ragesoss, line 124: This should be a named constant, so that we make sure that the limit here always matches the fetch limit.

Issue #4970, opened 2022-05-11
Articles Edited should have an option to show deleted articles


ragesoss:
### Current Behavior:

The Articles tab of a course page shows (by default) all the mainspace articles edited by a course, but if an article is subsequently deleted, it will disappear from the list. This is a common source of confusion, with instructors asking where an article they saw before went.

### Desired Behavior:

Articles that got deleted (or that the Dashboard mistakenly thinks are deleted) should be shown on Articles Edited, with some kind of visual indication of their deleted status.

PR #4975, opened 2022-05-15
Show deleted articles

TheTrio:
## What this PR does
Address #4970

### DIFF
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -44,6 +44,8 @@ const Article = createReactClass({
   render() {
     const ratingClass = `rating ${this.props.article.rating}`;
     const ratingMobileClass = `${ratingClass} tablet-only`;
+    const isDeleted = this.props.article.deleted;
+    const deletedMessage = I18n.t('articles.deleted_message');
 
     // Uses Course Utils Helper
     const formattedTitle = CourseUtils.formattedArticleTitle(this.props.article, this.props.course.home_wiki, this.props.wikidataLabel);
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/articles/article.jsx
```diff
@@ -44,6 +44,8 @@ const Article = createReactClass({
   render() {
     const ratingClass = `rating ${this.props.article.rating}`;
     const ratingMobileClass = `${ratingClass} tablet-only`;
+    const isDeleted = this.props.article.deleted;
+    const deletedMessage = I18n.t('articles.deleted_message');
```
ragesoss, line 48: Since this never varies and is only rarely used, it might make sense to implement this as a constant outside of the class. That way, it will only call I18n.t once, instead of once for every time an Article renders.

I doubt this one thing makes any practical performance difference, but I do think that overall, rendering large numbers of Article components can cause real performance problems.

It also just streamlines the logic slightly, as generating the `deletedMessage` from i18n is not really part of the core task of rendering this component.

app/views/courses/articles.json.jbuilder
```diff
@@ -1,10 +1,10 @@
 # frozen_string_literal: true
 
 json.course do
-  json.articles @course.articles_courses.live.includes(article: :wiki).limit(@limit) do |ac|
+  json.articles @course.articles_courses.all.includes(article: :wiki).limit(@limit) do |ac|
```
ragesoss, line 4: Is `.all` necessary, or can this just be `@course.articles_courses.includes...` ?

config/locales/en.yml
```diff
@@ -224,6 +224,8 @@ en:
       structure of a good article, but it does not take into account the content
       itself. (These scores are only available for English Wikipedia articles.)
     articles_short: Articles
+    deleted_message: >
+      This article might have been deleted - the dashboard can sometimes get out of sync, and incorrectly mark an article as deleted
```
ragesoss, line 228: "According to the Dashboard's data, this page has been deleted. (In some cases, page moves, history merges or other actions can result in an article being incorrectly listed as deleted. Deleted articles may also be later undeleted. For articles that are not actually deleted, or have been subsequently restored, the Dashboard will usually reflect this within one or two days.)"

ragesoss:
This looks pretty good. I suggest that, instead of the current "history" and "article development" links below the title, a deleted article should link to the page logs. For example, here are the public logs for the deleted page "Law of Nine": https://en.wikipedia.org/wiki/Special:Log?type=&user=&page=Law+of+Nine&wpdate=&tagfilter=&subtype=

config/locales/en.yml
```diff
@@ -226,7 +226,7 @@ en:
     articles_short: Articles
     deleted_message: >
       According to the Dashboard's data, this page has been deleted. (In some cases, page moves, history merges or other actions can result in an article being incorrectly listed as deleted. Deleted articles may also be later undeleted. For articles that are not actually deleted, or have been subsequently restored, the Dashboard will usually reflect this within one or two days.)
-
+    page_logs: (page logs)
```
ragesoss, line 229: It would be good to include a matching entry in the documentation (`qqq`) locale file, to explain that this particular message is a link to a MediaWiki logs for a particular page title. That will help translators understand the context.

TheTrio: Done.

PR #4982, opened 2022-05-23
Transition to webpack 5

TheTrio:
# What this PR does

Upgrades webpack to version 5

# Changes

### 1. Switch to relative paths for css assets

This is because absolute paths are now relative to the root of the project. 

### 2. Using asset modules instead of `url-loader`

[`url-loader`](https://v4.webpack.js.org/loaders/url-loader/) has been deprecated for Webpack 5 in favor of [asset modules](https://webpack.js.org/guides/asset-modules/). There are multiple asset types and I have gone with the [general asset type](https://webpack.js.org/guides/asset-modules/#general-asset-type). From the documentation - 

> Now webpack will automatically choose between resource and inline by following a default condition: a file with size less than 8kb will be treated as a inline module type and resource module type otherwise.

### 3. Single manifest file

Before, two manifest files were being created - one for css and another for JS. I've combined them into one `manifests.json` file so that the manifest plugin is only executed once. 

### 4. Replace `webpack-rtl-plugin` with `rtlcss-webpack-plugin`

[`webpack-rtl-plugin`](https://github.com/romainberger/webpack-rtl-plugin) was last updated on Mar 15, 2019. Despite there being a [PR which adds support for Webpack 5](https://github.com/romainberger/webpack-rtl-plugin/pull/37), it hasn't been merged for over 8 months. So its been replaced with [`rtlcss-webpack-plugin`](https://github.com/wix/rtlcss-webpack-plugin/) which looks to be maintained more actively. 

### 5. Add CssMinimizerWebpackPlugin to minify css

For some reason,`webpack-rtl-plugin` also minified the css files - both the RTL and non RTL ones. Since we no longer use that plugin, we need to minify the css with  [`css-minimizer-webpack-plugin`](https://github.com/webpack-contrib/css-minimizer-webpack-plugin) 

### 6. Remove ExcludeAssetsPlugin

This plugin was used to clean up the empty JS files which were being generated by Webpack. With v5, Webpack no longer generates these files, so there was no need to use this plugin. 

### 7. Remove Parallel Webpack

[Parallel Webpack](https://github.com/trivago/parallel-webpack) was used to speed up the build process by spreading the workload across multiple processors. The repo for this project has been archived and it does not support webpack 5. [Webpack doesn't have native support for parallelism](https://github.com/webpack/webpack/issues/14386), but that might change in the future. 

### 8. Use webpack dev server by default

Earlier, this was only being used with `yarn hot`. I couldn't get `yarn hot` to work so it might have been broken for a while. Regardless, `yarn start` now uses webpack dev server. The files are written to the disk presently, but I will try to use the server directly. I haven't been been able to make that work so far. The switch to webpack dev server also means that the webpage reloads automatically on code changes. It also shows warnings and errors inside of the browser. 

### 9. Fix regex for coverage

The regex which split `main.js` into multiple modules broke because webpack uses arrow functions instead of the function keyword. [I've changed that so that](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4982/commits/3a75ee282aceab7bbfc662ed3272564ac343149c) `yarn coverage` still works as expected. 
 
There are some other configuration changes but these are the major ones I could think of. 

# Performance

These numbers are obviously not scientific but the time taken is consistent from the testing I've done.

| command  | cache | Webpack 5 | Webpack 4 |
| --------  | ------------------- | --------------------- | --------------------- |
| `yarn start` | no | 70s | 40s |
| `yarn start` | yes | 10s | 10s |
| `yarn build` | no | 101s | 75s |
| `yarn build` | yes | 41s | 25s |

For rebuilds, it takes about 2000ms the first time and the subsequent rebuilds are much faster, around 800ms. This is probably due to the source map configuration. Unfortunately, if we want the right line numbers in the error messages, I think this is the best that can be done. More details can be found [here](https://webpack.js.org/configuration/devtool/)


ragesoss:
Super helpful summary! Can you provide before/after numbers (ie, v4 vs v5) for the `yarn` commands?

TheTrio:
Edited. I think the main reason for the performance difference is because of parallel webpack running things in, well, parallel.

TheTrio:
Here are the new numbers

| command  | cache | Webpack 5 | Webpack 4 |
| --------  | ------------------- | --------------------- | --------------------- |
| `yarn start` | no | 20s | 40s |
| `yarn start` | yes | 10s | 10s |
| `yarn build` | no | 58s | 75s |
| `yarn build` | yes | 41s | 25s |

It looks like ESLint is the culprit here. I've enabled `lintDirtyModulesOnly` which caches dirty modules only. So while the initial build times are faster, during development, when you edit the file for the first time, it will take time - 3-7 seconds. Subsequent edits will be faster and as the cache builds up, it will drop to around 1000ms. 

Essentially what we were doing during the build process(building up the eslint cache) is now being done during development. This slows things initially.

TheTrio:
Since most editors and IDEs run ESlint in the background anyways, I've added a new command - `start:no-lint`. This disables the ESlint plugin - which means that even the first edit is compiled instantly. 

`yarn start` is slow on the first edit, but after that(as the cache builds up), it should be almost as fast as `yarn start:no-lint`

css.webpack.js
```diff
@@ -57,48 +35,29 @@ module.exports = (env) => {
...
-          NODE_ENV: JSON.stringify(mode),
-        },
-      })
     ],
-    watch: env.watch_js,
-    devtool: env.development ? 'eval' : 'source-map',
+    optimization: {
+      minimizer: [
+        '...',
+        new CssMinimizerPlugin(),
```
dhruvdutt, line 58: Explore if we can pass any options that might be helpful.

webpack.config.js
```diff
@@ -32,11 +22,32 @@ module.exports = (env) => {
...
+    optimization: {
+      ...jsConfig.optimization,
+      ...cssConfig.optimization,
+    }
   };
+  if (env.development) {
+    output.devServer = {
+      devMiddleware: {
+        publicPath: path.join(__dirname, '/public'),
+        writeToDisk: true,
```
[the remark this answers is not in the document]
dhruvdutt: currently has an issue with dynamic import bundles if we switch to in-memory

js.webpack.js
```diff
@@ -23,39 +22,41 @@ const entry = {
 };
 
 module.exports = (env) => {
-  const doHot = env.development && !env.watch_js;
```
dhruvdutt, line 26: Need to check how this works with Ruby server and enable hot module replacement.

js.webpack.js
```diff
@@ -23,39 +22,41 @@ const entry = {
...
-  const mode = env.development ? 'development' : 'production';
-  const outputPath = doHot
-    ? path.resolve(appRoot, `${config.outputPath}/${config.jsDirectory}`)
-    : path.resolve(`${config.outputPath}/${config.jsDirectory}`);
-  const devtool = env.coverage ? 'cheap-module-source-map' : 'cheap-eval-source-map';
+  const outputPath = path.resolve(appRoot, `${config.outputPath}/${config.jsDirectory}`);
```
dhruvdutt, line 25:
```suggestion
  const outputPath = path.resolve(`${config.outputPath}/${config.jsDirectory}`);
```

js.webpack.js
```diff
@@ -23,39 +22,41 @@ const entry = {
...
+
+  let devtool;
```
dhruvdutt, line 27:
```suggestion
  let devtool = 'eval-cheap-source-map';
```

lib/tasks/coverage.rake
```diff
@@ -14,7 +14,7 @@ namespace :generate do
       funs.each do |fun|
         counter += 1
         text = text.sub(fun, "*/module_#{counter}") # Replaces the function with module_{number}
-        fun = fun.gsub(%r/\/[*]{3}\/ \(function/, "var module_#{counter} = (function")
+        fun = fun.gsub(%r/\/[*]{3}\/ \(/, "var module_#{counter} = (")
```
dhruvdutt, line 17: @ragesoss Need to validate if `coverage.rake` file is working as expected since the output in `public/assets/javascripts/main.js` might have changed from webpack 4 to 5.

ragesoss: Yep, I tried it out locally and the coverage.rake is still working.

css.webpack.js
```diff
@@ -57,48 +35,29 @@ module.exports = (env) => {
...
-          if (/rtl-.*\.css$/.test(file.path)) {
-            file.name = `rtl-${file.name}`;
-          }
-          return file;
-        }
-      }),
-      // node environment
-      new webpack.DefinePlugin({
-        'process.env': {
-          NODE_ENV: JSON.stringify(mode),
```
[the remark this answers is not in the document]
TheTrio: [It is](https://webpack.js.org/plugins/define-plugin/) but it looks like since Webpack 4 defining `mode` should set `process.env.NODE_ENV` [automatically](https://webpack.js.org/guides/production/#specify-the-mode) - so this might've been redundant as well.

css.webpack.js
```diff
@@ -57,48 +35,29 @@ module.exports = (env) => {
...
-          },
+          type: 'asset'
         }
       ]
     },
     plugins: [
       // extracts CSS to a separate file
       new MiniCssExtractPlugin({
-        filename: env.development ? '../stylesheets/[name].css' : '../stylesheets/[name].[hash].css',
+        filename: env.development ? '../stylesheets/[name].css' : '../stylesheets/[name].[contenthash].css',
```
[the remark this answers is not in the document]
TheTrio: Looks like the supported substitutions are `hash`, `contenthash` and `chunkhash`. The reason I went with `contenthash` was because the [webpack 5 migration guide recommended it](https://webpack.js.org/migrate/5/#clean-up-configuration)

css.webpack.js
```diff
@@ -57,48 +35,29 @@ module.exports = (env) => {
...
       }),
       // generates a RTL version of the emitted CSS files
-      // this is done only in production
-      ...(env.production || env.coverage ? [new WebpackRTLPlugin({
-        filename: '../stylesheets/rtl-[name].[contenthash].css'
+      // this is done only in production/coverage
+      ...(env.production || env.coverage ? [new RtlCssPlugin({
+        filename: '../stylesheets/rtl-[name].[fullhash].css'
```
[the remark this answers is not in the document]
TheTrio: The `rtl-css-plugin` still has some deprecation warnings when using `contenthash`. Which is why I used `fullhash` here.

ragesoss:
@TheTrio @dhruvdutt let me know when both of you are happy with this PR. I can merge it as soon as you both think it's ready.

TheTrio:
Yeah having just one webpack config would definitely make things simpler. 

Apart from that, I don't think there's anything major left to do in this PR. There's still some stuff I want to explore - like HMR, speeding up eslint and switching to memory mode for the dev server. 

But I think those can be tracked independently of this PR.

css.webpack.js
dhruvdutt: @TheTrio Were you able to explore this?

dhruvdutt:
@TheTrio Can you please create issues for things we're going to be looking into separately?
- HMR
- eslint optimizations
- Webpack dev server in-memory chunks
- Vendor chunking

css.webpack.js
TheTrio: The defaults were good enough in my opinion so there wasn't much to configure. As for CSS Nano options, auto prefixing was the only one that felt necessary to enable but then I found out that stylus loader already does that. 

So no, not much there I think. Do you have some specific configuration in mind?

Issue #3210, opened 2019-08-16
Users should be able to update their email via their user profile page


ragesoss:
### Current Behavior:
The only way to edit your email is to go through `/onboarding` again, which is not discoverable.

### Desired Behavior:
Email should be one of the editable fields in the user profile edit view.

katherine95:
@ragesoss I would like to take this up

ragesoss:
Great! Let us know if you run into any trouble.

Manasa2850:
@ragesoss I would like to work on this

marklocklear:
@ragesoss looking at this one. Just so I'm clear, we are wanting to edit the user model (@user.email) and email preferences in the user profile (user_profile.email_preferences)?

ragesoss:
@marklocklear I think I just had in mind updating @user.email.

The `user_profile.email_preferences` is, if I remember correctly, designed to let a user follow a link in a particular email and opt out of an further emails of that kind, but it's used very minimally and I don't think we need to be able to see or edit that from the profile page.

PR #4988, opened 2022-05-26
Allow email edit

marklocklear:
## What this PR does
Addresses https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/3210#issuecomment-1122880124

## Screenshots
Allows the user to edit their email address

image

Nothing at this time.


spec/controllers/user_profiles_controller_spec.rb
```diff
@@ -187,11 +191,21 @@
...
         post route, params: { username: user.username,
+                              email: { email: user.email },
                               user_profile: { id: profile.id,
                                               user_id: profile.user_id,
                                               image_file_link: file_link } }
         expect(user.user_profile.image_file_link).to eq(file_link)
       end
+
+      it 'updates users email' do
+        user.email = 'fake_email@gmail.com'
```
ragesoss, line 202: This doesn't look like it tests that an update happened. It just sets a user's email, then posts that same email, then checks that the user's email is still the one that was set. It should pass even if the `post` is omitted.

ragesoss:
Can you align the email input with the other two below it, by making the icon a little bigger?

spec/controllers/user_profiles_controller_spec.rb
marklocklear: Little help here? My spidey sense was tingling a bit when I wrote it but I took another stab at it and no cigar https://github.com/marklocklear/WikiEduDashboard/commit/3d5ed8b26cae4a2a3eb4c5de8ac4092e6ba20da5. If you look at the parameters in the console they are:

```Parameters: {"utf8"=>"✓", "authenticity_token"=>"YTVgxykvtLLFfRstHvrIpHu2ElRjPPzS5WTb6y+eGcmKzEzcfQwPzlNBxQ6MAjD0WIQwge0KNO24RZzN5/MP9A==", "user_profile"=>{"image_file_link"=>"", "bio"=>"Here is my bio~ test", "location"=>"Prospect NC", "institution"=>"uncp"}, "email"=>{"email"=>"marklocklear@extension1.org"}, "username"=>"Marklocklear"}```

...which is why I feel like `email: { email: email }` should work.

ragesoss: The way that rspec works is that the first time you call `user` in the test, run the `let` block and execute the code `create(:user)`, which creates a database User record and returns a User object for that record, and then every subsequent time you call `user`, it's going to return that same object, which is still in memory, without touching the database. So in this first line, you get a User object and then set the email (but, importantly, you don't set that new value in the database). Then in the `post` line, you hit the controller which updates the User record in the database with whatever value you passed (in this case, the same value that you set without persisting to the database in the first line). The controller action will persist that value to the database, so the end state of the database should be the same as if you had done `user.update(email: 'fake_email@gmail.com'). However, changing the database via that `post` does not affect the in-memory `user` object that you you are using here to test. For the `expect` line at the end, it's reading the value of `user.email` from that same in-memory object, but it does not know or care whether the database record for that user has changed in the meantime.

The way to fix it is to:
1. explicitly set an email value in the `let(:user)` block.
2. pass a different string as the email value in the `post` call
3. verify that the database got updated with the new value, which you can do with `user.reload.email` in the `expect` at the end. (Calling `reload` on an ActiveRecord object will refresh the attributes from the database and return the same object.)

ragesoss:
> OK, fixed this with [marklocklear@69376d0](https://github.com/marklocklear/WikiEduDashboard/commit/69376d0badb9c2abda8c136036952cf4b96138da) icon is still a bit small, but it was the best I could do with out bringing in a new svg file.

Thanks! Can you update the screenshots? As long as the new size makes the inputs aligned, I'm fine with it.

spec/controllers/user_profiles_controller_spec.rb
marklocklear: Great explanation thank you! Updated this here https://github.com/marklocklear/WikiEduDashboard/commit/93e8662f6f284502f2c2cfbc38478e659424aaa5

ragesoss:
Thanks. I'd like the alignment of those inputs to be as perfect as possible.

ragesoss:
I cherry-picked these commits and merged this feature. The layout and styling code for this is kind of a mess (not your fault) so I decided it's good enough as-is.

Thanks @marklocklear!

(FYI, for future reference, it's much preferable if you can rebase your branch rather than merge newer commits from upstream as you did here. See [docs/dev_process.md](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/dev_process.md))

Issue #4961, opened 2022-05-03
Add an easy way to see a student's edits in a particular namespace


ragesoss:
### Current Behavior:

The Assignments & Articles view of the Students shows a list of recent edits, a list of assigned and other edited articles, and links to a) an on-wiki list of the user's sandboxes, and b) the on-wiki contribution history.

Sometimes, though, instructors want to check for specific kinds of edits, for example, edits to Talk pages to find article critiques that have been posted there.

### Desired Behavior:

The Students tab should make it easy find specific kinds of edits. Maybe could best be done with a filter option for the Revisions list. 

### Notes

The example use case might be ideally be solved through adding the ability for the Dashboard to keep track of where each exercise should be completed and whether or not edits to that page have been made, so that instead of a more generic filter for talk page edits, the instructor could easily find the specific page where the student posted work for that particular exercise.

However, more powerful filtering of revisions (for both the Activity tab and the Students tab) would still be useful as well.

vaidehi44:
@ragesoss, the links which you mentioned - 
>a) an on-wiki list of the user's sandboxes, and b) the on-wiki contribution history. 

Are you referring to these links?

![wikiedu](https://user-images.githubusercontent.com/56690165/168664305-28dd2462-fb58-4e52-b986-4740eedfee02.png)

ragesoss:
Yes, exactly.

vaidehi44:
Okk...I am not getting where the filter option would go. You said to have a filter option for Revision list. Does it mean for the `Revisions` table. 
![image](https://user-images.githubusercontent.com/56690165/168667071-74090bb6-fdc0-4dc0-9a94-dbf86efb0da4.png)

Or, should it filter the results for entire page (the `Assignments & Exericises` tab)?

vaidehi44:
Okay. But that wouldn't affect the above links, right. So, we'll have to add a section, probably a column ,in the table for links corresponding to the filter?

ragesoss:
Something like that. I think the same kind of UI above the table that we use on the Articles tab — for example https://outreachdashboard.wmflabs.org/courses/Wikimedia_Ghana_Entertainment_Community/3Music_Awards/articles/edited?newness=new&tracked=both — would probably work.

vaidehi44:
[...]

PR #4989, opened 2022-05-26
Filter to see students edits in particular namespace

vaidehi44:
## What this PR does
This PR tries to resolve this issue - #4961

## Screenshots
Before:
image

After:
image


vaidehi44:
@ragesoss, I have provided options to filter for article, user and talk namespace. Let me know if it's okay, or should add more.

app/assets/javascripts/components/students/components/Articles/SelectedStudent/RevisionsList/StudentRevisionsList.jsx
```diff
@@ -67,10 +90,27 @@ export class StudentRevisionsList extends React.Component {
...
     const keys = { recent_revisions, character_sum_ms, references_count, total_uploads };
 
+    const filterLabel = <b>Namespace Filter:</b>;
+    const filterRevisions = (
+      <select
+        className="filter-revisions"
+        value={this.state.namespace}
+        onChange={this.onNamespaceChange}
+      >
+        <option value="all">All</option>
```
ragesoss, line 100: The option text for each of these should use I18n.

app/assets/javascripts/components/students/components/Articles/SelectedStudent/RevisionsList/StudentRevisionsList.jsx
```diff
@@ -14,12 +14,26 @@ export class StudentRevisionsList extends React.Component {
...
     };
 
     this.toggleDrawer = this.toggleDrawer.bind(this);
   }
 
+  onNamespaceChange = (e) => {
+    return this.setState({ namespace: e.target.value });
+  };
+
+  namespaceToId(nm) {
```
ragesoss, line 28: This is is likely a generally useful mapping that we may want to extend later on, and may want to use in other places at some point.

I suggest extracting it to `article_utils.js`.

ragesoss:
This works smoothly when used as intended. Nice work!

There are two UX issues I noticed that might be confusing in some cases.

1. If the filter selector is used when the revisions drawer is collapsed (as it is by default) then there's no visible change when a filter option is selected. It would be better if, when the drawer is closed, changing the filter selection would open the drawer.
2. The 'no revisions' message is slightly misleading when a filter is selected such that no revisions are shown: "This editor has made no revisions.". If there are revisions in other namespaces, it's not the case that the user has made no revisions (or even no revisions in the selected namespace), just that of the 10 most recent revisions, none of them are in that namespace. Maybe the message should just be "(no revisions)".

vaidehi44:
Okay @ragesoss. About your second point; could a more detailed message be used, like 'There are no recent edits in this namespace'? Also, the original message, 'This editor has made no revisions.', should still show if there are no revisions for `All` option, right?

ragesoss:
@vaidehi44 those kinds of variable messages would be maybe be a tiny bit better for users, but I think the simplicity of using the same message in all situations of an empty table would be a little better because it won't add code complexity and making the message vary by filter might make the code harder to work with later on.

vaidehi44:
Okay, will use "(no revisions)" for all the cases.

ragesoss:
Perfect!

PR #4992, opened 2022-06-02
Switch to react spring

TheTrio:
Addresses #4510 

This also introduces a new drop target so that you can drag items into an empty week. 
 
Before: 

before

After: 

drag


app/assets/javascripts/components/timeline/DummyBlock.jsx
```diff
@@ -0,0 +1,46 @@
...
+          onMoveUpWeek(_item.item.id);
+        }
+      },
+      collect: monitor => ({
+        isOverCurrent: monitor.isOver({ shallow: true }),
+      }),
+    }),
+    [],
+  );
+  const text = 'Empty week. Drag items here ';
```
ragesoss, line 25: Use I18n.

ragesoss:
This works perfectly! I'm really glad you implemented the DummyWeek. The lack of a drag target in empty weeks has been a source of occasional confusion for a long time... always just infrequent enough for me not to prioritize fixing it.

TheTrio:
So I found a couple of issues with the current implementation. 

1. If you move across the empty week and then come from the opposite direction, the drop doesn't move the block correctly. 
2. It only worked if the dropped block was adjacent to the empty week. This case can only be hit if for some reason you drag the block from the side.
 
bugs

I've fixed this now

fixed

ragesoss:
Nice catch. I probably wouldn't have found out about that problem for a long time.

For the empty block message, let's go with "Empty week. You can move a block from another week here."

Issue #5005, opened 2022-06-10
Switch to in-memory mode for the dev server


TheTrio:
Writing files to the disk is a slow process - serving the files from memory is a lot more faster and efficient.

Switching to in-memory mode should be straightforward but there are some issues related to [dynamic imports](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#dynamic_imports).

To see for yourself, set `hot_loading` to `true` in `config/application.yml` and disable `writeToDisk` in `webpack.config.js`

Issue #5003, opened 2022-06-10
Add support for HMR(Hot Module Replacement)


TheTrio:
Presently, when you run start the dev server using `yarn start` and make some changes, the browser window reloads when those changes are compiled. This means that you lose all your existing state(like filled in forms, or data stored in the Redux store)

[Hot Module Replacement](https://webpack.js.org/guides/hot-module-replacement/) allows code to be updated at runtime without the need for a full refresh. The [webpack dev server enables HMR by default](https://webpack.js.org/configuration/dev-server/#devserverhot). However, HMR doesn't actually work - the page always does a full refresh on recompilation, possibly because of how this interacts with the Rails server.

Look into how to make HMR play nicely with the Rails.

PR #5013, opened 2022-06-16
add support for HMR

TheTrio:
Addresses #5003 and #5005 

This PR attempts to add support for HMR along with in memory mode for Webpack. 

It sets up a proxy so that all requests to Rails under `/assets` are forwarded to the Webpack dev server.  The dev server holds these files in memory instead of writing them to the disk - which should make things a bit faster. 

I've also had to split up `app.jsx` into multiple parts to ensure that the Redux store remains persistent across hot reloads. 

~~There are still some issues - for example, pages which load TinyMCE will not be hot reloaded - but for the most part, HMR should work as expected.~~ TinyMCE should no longer causes any issues. See below for details.

## How to Enable

To enable it, make sure your `application.yml` has the following options set

```
DISABLE_SENTRY: "true"
hot_loading: "true"
```

After that, you can run either `yarn hot` or `yarn hot:no-lint` to start the dev server. Then start the Rails server as usual.

Now you can edit the React code and have it reflected in the browser without needing a refresh. The progress(and any errors) can be viewed in the browser console.

hmr

If the code results in an error(runtime or compile time), its shown in the browser as an overlay.

error

Across these "refreshes", the Redux state stays persistent. However, if you edit one of the reducer files, the page will hard reload.


ragesoss:
Are the un-covered lines in `application_helper.rb` now dead code?

TheTrio:
No, I don't think so. I think the reason they're not being executed now is I've added an additional check for `Rails.env.development`. 

I found it annoying to have to disable `hot_loading` everytime I wanted to run a test. So I made sure hot loading only affected when you're in development.

TheTrio:
With this, TinyMCE is now being bundled with Webpack directly, instead of us specifying the file manually in the config file.

This means two things - 

1. HMR works since TinyMCE is now part of the dependency chain
2. We no longer need to insert the script tag manually in the HTML.

The comments in `app/assets/javascripts/components/common/text_area_input.jsx` say that dynamic imports didn't work earlier. I am not really sure what changed - maybe Webpack 4 had some issues with it or maybe there was some conflicting config option. 

Regardless, dynamic imports work as expected now.

ragesoss:
> > On my computer, it takes about two seconds to compile a change, and about three seconds total before the changed module re-renders. This feels a bit slow. Do you know why it takes 2 seconds to compile a trivial change to a single file?
> 
> Did you try `yarn hot:no-lint`? That should skip the linting and should be a bit faster. Also, subsequent changes should take less time as the cache builds up.

Do you think there's a good reason to leave linting enabled by default for `yarn hot`?

Seems like it needs to warm up with a few recompiles when linting is disabled, but it gets the compile time down to about half a second, which makes HMR a lot more fun to use.

TheTrio:
> Do you think there's a good reason to leave linting enabled by default for yarn hot?

Not really. Linting is enabled for `yarn start` which is why I enabled it for `yarn hot`. Disabling it by default would give a better experience though. What do you think?

config/initializers/dev_server.rb
```diff
@@ -0,0 +1,4 @@
+require_dependency "#{Rails.root}/lib/dev_server_proxy"
+if Features.hot_loading? && Rails.env.development?
+  Rails.application.config.middleware.use WebpackDevServerProxy, dev_server_host: "localhost:8080"
```
ragesoss, line 3: so... what if, instead of switching on Rails.env.development, this middleware just varied the host based on environment? Like, it uses localhost:3000 in test and localhist:8080 in dev?

app/helpers/application_helper.rb
```diff
@@ -37,7 +37,9 @@ def hot_javascript_tag(filename)
   # :nocov:
 
   def hot_javascript_path(filename)
-    return "http://localhost:8080/#{filename}.js" if Features.hot_loading?
```
ragesoss, line 40: Would this work as `/assets/javascripts/#{filename}.js` (which would delegate the domain to the proxy middleware)?

TheTrio: ah, I think I misunderstood you. Did you mean to replace this with `/assets/javascripts/#{filename}.js`?

ragesoss: Yeah, that that's what I meant.

TheTrio:
Couple of changes:

1. `yarn hot` will disable linting by default. 
2. Setting `hot_loading` to `"true"` in `application.yml` will be enough to enable HMR

lib/dev_server_proxy.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+class WebpackDevServerProxy < Rack::Proxy
+  def initialize(app = nil, opts = {})
+    super
+    @dev_server_host = opts[:dev_server_host]
+  end
+
+  def perform_request(env)
```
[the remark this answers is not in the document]
TheTrio: I found a similar example here - https://github.com/ncr/rack-proxy#rails-middleware-example

I think `@app.call` is the default behavior. And the `super` call is for when we need to forward the request to the Dev Server.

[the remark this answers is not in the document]
ragesoss: Can you add comments about what this file is for, how it gets loaded, and what example it's following?

I think this file should probably be moved to a location that more clearly communicates the idea that this is just a file for dev/test environments and isn't used in production. Maybe `lib/development/dev_server_proxy.rb`.

Also, you should add a simplecov filter for this file in `spec/spec_helper.rb`.

PR #5018, opened 2022-06-17
Update coverage documentation

TheTrio:
This also deletes `lib/tasks/instrument_js.rake`.


TheTrio:
Don't think the test failure is related.

PR #5039, opened 2022-07-09
convert course lists to JSX

TheTrio:
This PR attempts to convert several of the courses lists to react components. This includes the lists at the following URLs

1.  `/explore`
2. `/active_courses`
3. `/courses_by_wiki`

For `/explore` and `/active_courses`, the entire page is rendered in react. This includes the search functionality on `/explore`

For `/courses_by_wiki`, the header stats are part of the haml file. 

There still remain 4 courses tables which are rendered on the server. They are:-

1. `app/views/campaigns/programs.html.haml`
2. `app/views/dashboard/_admin_courses.html.haml`
3. `app/views/unsubmitted_courses/index.html.haml`
4. `app/views/dashboard/index.html.haml`

Converting them will require a bit more effort and since this PR is already quite large, I think I'll take a stab at these in another PR.


TheTrio:
# Explore Before

https://user-images.githubusercontent.com/10794178/178753498-417308b8-7886-42d9-8459-09f318a7b953.mp4

# Explore After

https://user-images.githubusercontent.com/10794178/178753405-ea99bba4-838f-4d15-b3f0-e03ed9dad8a3.mp4

# Active Courses Before

https://user-images.githubusercontent.com/10794178/178753749-be093497-2c7c-4b6c-b8a6-cfe3fb5ff9df.mp4

# Active Courses After

https://user-images.githubusercontent.com/10794178/178753791-a33d5101-6c20-4009-a657-135feee20c02.mp4

# Courses By Wiki Before

https://user-images.githubusercontent.com/10794178/178753899-57fe1aeb-5aef-4892-ba4d-7960d3e9562b.mp4

# Courses By Wiki After

https://user-images.githubusercontent.com/10794178/178753976-2e74c640-ab6e-4f1e-9f10-238804401653.mp4

app/assets/javascripts/actions/course_actions.js
```diff
@@ -1,10 +1,14 @@
 import {
   ADD_NOTIFICATION, API_FAIL, UPDATE_COURSE, RECEIVE_COURSE, RECEIVE_COURSE_UPDATE,
   PERSISTED_COURSE, DISMISS_SURVEY_NOTIFICATION, TOGGLE_EDITING_SYLLABUS,
-  START_SYLLABUS_UPLOAD, SYLLABUS_UPLOAD_SUCCESS, LINKED_TO_SALESFORCE, COURSE_SLUG_EXISTS
+  START_SYLLABUS_UPLOAD, SYLLABUS_UPLOAD_SUCCESS, LINKED_TO_SALESFORCE, COURSE_SLUG_EXISTS, RECEIVE_COURSE_SEARCH_RESULTS, SORT_COURSE_SEARCH_RESULTS, FETCH_COURSE_SEARCH_RESULTS, RECEIVE_ACTIVE_COURSES, SORT_ACTIVE_COURSES,
+  RECEIVE_CAMPAIGN_ACTIVE_COURSES
 } from '../constants';
+import { RECEIVE_WIKI_COURSES, SORT_WIKI_COURSES } from '../constants/wiki_courses';
```
ragesoss, line 7: Can't these also be imported straight from the `constants` index like the others?

app/assets/javascripts/actions/course_actions.js
```diff
@@ -1,10 +1,14 @@
...
 import API from '../utils/api.js';
 import CourseUtils from '../utils/course_utils';
+import request from '../utils/request';
+import { url } from '../utils/wiki_utils';
```
ragesoss, line 11: This function would benefit from a renaming (eg, `wikiDomain`).

app/assets/javascripts/reducers/active_courses.js
```diff
@@ -0,0 +1,42 @@
...
+    case RECEIVE_ACTIVE_COURSES:
+    case RECEIVE_CAMPAIGN_ACTIVE_COURSES: {
+      return {
+        ...state,
+        courses: action.data.courses,
+        isLoaded: true,
+      };
+    }
+    case SORT_ACTIVE_COURSES: {
+      const desc = action.key === state.sort.sortKey;
```
ragesoss, line 28: As we discussed, this should follow the example of the `articles` reducer for handling default sort order for columns where `desc` should be the default.

app/assets/javascripts/reducers/course_search_results.js
```diff
@@ -0,0 +1,44 @@
...
+    case RECEIVE_COURSE_SEARCH_RESULTS: {
+      const newState = {
+        ...state,
+        results: action.data.courses,
+        loaded: true
+      };
+      return newState;
+    }
+    case SORT_COURSE_SEARCH_RESULTS: {
+      const desc = action.key === state.sort.sortKey;
```
ragesoss, line 30: Same here, follow the `articles` reducer as the example.

app/assets/javascripts/reducers/index.js
```diff
@@ -80,7 +83,10 @@ const reducer = combineReducers({
   users,
   validations,
   wikidataLabels,
-  wizard
+  wizard,
+  course_search_results,
+  active_courses,
+  wiki_courses
```
ragesoss, line 89: Put these in alphabetical order.

app/assets/javascripts/utils/model_utils.js
```diff
@@ -32,6 +32,9 @@ export const sortByKey = (models, sortKey, previousKey = null, desc = false, abs
   // null. The desired order requires the null values to be in the end instead
   // of beginning.
   const sortFunc = (model) => {
+    if (typeof model[sortKey] === 'string') {
```
ragesoss, line 35: A comment to explain the purpose would be helpful here.

app/controllers/active_courses_controller.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 
 class ActiveCoursesController < ApplicationController
-  respond_to :html
+  respond_to :json, :html
 
   def index
```
ragesoss, line 6: It seems like this is being overloaded for HTML vs JSON, with the same instance variables being set up (and the same queries being run) for both.

Would this make more sense with a different controller action for the html index versus the JSON active courses list?

app/controllers/courses_by_wiki_controller.rb
```diff
@@ -1,14 +1,15 @@
...
-  respond_to :html
+  respond_to :html, :json
 
   before_action :set_wiki
 
   def show
     @courses_list = Course.where(home_wiki: @wiki).order(id: :desc)
     @courses_list = @courses_list.where("year(created_at) = #{params[:year].to_i}") if params[:year]
     @presenter = CoursesPresenter.new(current_user: current_user, courses_list: @courses_list)
+    @courses = @presenter.courses
```
ragesoss, line 12: As above, I think re-using the same action for both HTML and JSON will result in duplicate queries being run.

app/controllers/active_courses_controller.rb
TheTrio: Yeah this is something I changed in [321f966](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5039/commits/321f966ea4ee2641400b4daaea692d213bffb5c2)

I need the presenter to get the course string prefix. In this case though, I suppose that will always be equal to `Features.default_course_string_prefix`? If so, I could just set it to that.

app/views/explore/index.html.haml
```diff
@@ -1,18 +1,5 @@
 - content_for :after_title, " - #{t(Features.wiki_ed? ? 'application.explore' : 'courses_generic.explore')}"
-
-%header.main-page
-  .container
-    %h1=  ENV['dashboard_title']
-
+:javascript
+  Features.course_string_prefix = "#{@presenter.course_string_prefix}"
```
ragesoss, line 3: I guess this is the only dynamic part of the explore page now? It looks like initializing a CoursesPresenter is going to also do a query for the campaign courses, even if it's not going to need it (as with serving this HTML route), so it might make sense to just use `Features.default_course_string_prefix` instead of the presenter. I think in this context it will be the same.

app/views/courses_by_wiki/show.html.haml
```diff
@@ -7,58 +7,7 @@
   .container
     = render 'courses/header_stats', presenter: @presenter
 
+:javascript
+  Features.course_string_prefix = "#{@presenter.course_string_prefix}"
```
ragesoss, line 11: I think this can be `Features.default_course_string_prefix` as well.

app/views/active_courses/index.html.haml
```diff
@@ -1,61 +1,4 @@
-- content_for :after_title, " - Active Courses"
-
-%header.main-page
-  .container
-    %h1= ENV['dashboard_title']
-
+:javascript
+  Features.course_string_prefix = "#{@presenter.course_string_prefix}"
```
ragesoss, line 2: Same here, I think. Replace presenter with `Features.default_course_string_prefix`.

spec/features/explore_page_spec.rb
```diff
@@ -63,83 +44,131 @@
...
+        course_id: @course_active_name_z.id,
+        campaign_id: Campaign.default_campaign.id
+      )
+      CampaignsCourses.create(
+        course_id: @course_inactive.id,
+        campaign_id: Campaign.default_campaign.id
+      )
+    end
 
-    it 'allows navigation to a campaign page' do
```
ragesoss, line 132: Does this not work anymore?

spec/features/explore_page_spec.rb
```diff
@@ -63,83 +44,131 @@
...
+
+      it 'for everyone else' do
+        visit '/explore'
+        expect(page).to have_content(@course_active_name_a.title)
+        expect(page).to have_content(@course_active_name_z.title)
+        expect(page).not_to have_content(@course_inactive.title)
+      end
     end
 
-    it 'allows navigation to a course page' do
```
ragesoss, line 138: Does this not work anymore?

app/controllers/active_courses_controller.rb
ragesoss: Yeah, for /explore and /active_courses, I think these should always be equal to the default.

For an individual campaign, it could vary based on the campaign's default course type, but that could be done just by loading the campaign itself and doing `campaign.course_string_prefix` instead of going through an otherwise-unnecessary presenter.

ragesoss:
Overall, the code looks good. I'll test it out once you've implemented the changes we talked about today.

spec/features/explore_page_spec.rb
TheTrio: ah I think I removed this accidentally. Will add back.

app/views/explore/index.html.haml
TheTrio: I thought the value might be different for different campaigns. I suppose that case might not be relevant for these templates.

ragesoss:
There's a styling bug with the light gray text, which is too close to the main text.
layout bug with gray text

ragesoss:
The 'All Courses' link on `/explore` points to `/campaigns`.

TheTrio:
Have fixed the styling. This is how it looks now

image

Have corrected the all courses link as well.

PR #5050, opened 2022-07-15
split bundles based on routes

TheTrio:
This is [one of the recommendations on the official documentation](https://reactjs.org/docs/code-splitting.html#route-based-code-splitting). It means that the user only downloads the code of the routes they actually visit.

I've also had to change the default export in the course creator file because dynamic imports don't support named exports.


ragesoss:
Do you notice any differences in user experience with this change, locally?

What user experience changes should we expect in production (where some pages may load more slowly than similar ones in your development environment do)?

PR #5051, opened 2022-07-15
Track and display stats for namespaces

vaidehi44:
## What this PR does
The PR is an enhancement for tracking namespace specific stats on dashboard.

## Screenshots

image

image

* Currently, the namespace stats attributes are not appropriate for all the namespaces. They are probably suitable for namespace similar to `mainspace` like `cookbook`, `wikijunior`.


vaidehi44:
@ragesoss, here are some of my queries, which you can probably answer after going through the pr.

* In `wiki.rb`, I have defined a mapping of projects and their namespace ids. I am not sure if its the right place and way to define it. But I wanted to store this mapping somewhere on the backend side. Was also confused if it can be defined in  `courses_namespaces.rb`.
* Also,  now `CoursesNamespaces` is related to `CoursesWikis`, and it just has a namespace field. Is there a need of different model for this? Can we just have a `namespaces`  (array/hash of ns ids) column in `CoursesWikis` then?
* And, while selecting the namespaces to be tracked, in Tracked Namespaces dropdown, I am automatically adding the mainspaces for all the tracked wikis. May be this is not desired, but don't know why I had included such feature. I guess as we show the mainspace stats aggregatively for all tracked wikis on the overview page, thats why I had done this. But this would mean that instructor would always get mainspace stats seperately for all the tracked wikis. So, I think this would be wrong. Let me know if I should remove this thing or change it.

app/assets/javascripts/components/campaign/campaign.jsx
```diff
@@ -47,7 +47,9 @@ export const Campaign = createReactClass({
         <div className="container campaign_main">
           <section className="overview container">
             <CampaignStats campaign={this.props.campaign} />
-            {this.props.campaign.course_stats && <WikidataOverviewStats
+            {this.props.campaign.course_stats &&
```
TheTrio, line 50: This property should be included in the reducer `app/assets/javascripts/reducers/campaign.js`

app/helpers/course_helper.rb
```diff
@@ -29,14 +29,25 @@ def course_i18n(message_key, course = nil)
     I18n.t("#{string_prefix}.#{message_key}")
   end
 
-  def format_wikidata_stats(course_stats)
```
TheTrio, line 32: This function is still referenced in `app/views/campaigns/show.json.jbuilder`

app/assets/javascripts/components/overview/namespace_select.jsx
```diff
@@ -0,0 +1,134 @@
...
+      });
+      if (ns_obj !== undefined && ns_obj !== null) {
+        namespaces = ns_obj.namespaces;
+        if (ArrayUtils.hasObject(namespaces, main_ns_id)) return { wiki, namespaces };
+        namespaces.unshift(main_ns_id);
+        return { wiki, namespaces };
+      }
+      return { wiki, namespaces };
+    });
+    updateNamespaces(new_namespaces);
```
TheTrio, line 62: I think it might be better to include this logic in the course reducer with a new action - say `UPDATE_NAMESPACES`. Will help simplify this file.

app/assets/javascripts/components/overview/namespace_select.jsx
```diff
@@ -0,0 +1,134 @@
...
+    const new_namespaces = tracked_wikis.map((wiki) => {
+      wiki.language = wiki.language || 'www'; // for multilingual wikis, language is null
+      let namespaces = [0];
+      const main_ns_id = 0;
+      const ns_obj = tracked_namespaces.find((obj) => {
+        return JSON.stringify(obj.wiki) === JSON.stringify(wiki);
+      });
+      if (ns_obj !== undefined && ns_obj !== null) {
+        namespaces = ns_obj.namespaces;
+        if (ArrayUtils.hasObject(namespaces, main_ns_id)) return { wiki, namespaces };
```
TheTrio, line 56: Is there a reason to use `ArrayUtils.hasObject` instead of just `namespaces.includes`?

app/assets/javascripts/components/overview/namespace_select.jsx
```diff
@@ -0,0 +1,134 @@
...
+        namespaces.unshift(main_ns_id);
+        return { wiki, namespaces };
```
TheTrio, line 58: This return is not required.

app/assets/javascripts/components/common/wiki_overview_stats.jsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const hasWikidataStats = wikidataStats !== null;
+  const isWikidata = (hasWikidataStats && currentTabId === 0);
+
+  let statsData = [];
+  if (hasWikidataStats) {
+    statsData.push({ id: 0, tabTitle: 'Wikidata', data: wikidataStats });
+  }
+
+  let i = hasWikidataStats ? 1 : 0;
```
ragesoss, line 40: I guess `i` here is an index for which tab of stats it is? This would benefit from a more meaningful variable name, or a comment explaining how this iterator variable should be understood.

app/assets/javascripts/components/common/wiki_overview_stats.jsx
```diff
@@ -0,0 +1,89 @@
...
+import PropTypes from 'prop-types';
+
+import WikiOverviewTabs from './OverviewStats/wiki_overview_tabs';
+import NamespaceOverviewStats from './OverviewStats/namespace_overview_stats';
+import WikidataOverviewStats from './wikidata_overview_stats';
+
+import ArticleUtils from '../../utils/article_utils';
+
+
+const WikiOverviewStats = ({ wikidataStats, wikiNamespaceStats }) => {
```
ragesoss, line 11: I think this component's name could be improved, to more clearly indicate what it's for (displaying and interacting with a set of tabs that show narrow wiki- or namespace-specific stats).

app/assets/javascripts/utils/article_utils.js
```diff
@@ -30,4 +30,39 @@ export default class ArticleUtils {
     };
     return mapping[namespace];
   }
+
+  static NamespaceTitleFromId = {
```
ragesoss, line 34: If this mapping is duplicated in Ruby, both places should include a comment pointing to the other one so that we are less likely to update one without updating the other.

app/controllers/courses_controller.rb
```diff
@@ -269,6 +271,29 @@ def update_courses_wikis
     @course.update_wikis(new_wikis)
   end
 
+  def courses_namespaces_params
+    params
+      .require(:course)
+      .permit(namespaces: [wiki: [:language, :project], namespaces: []])
+  end
+
+  def update_courses_namespaces
```
ragesoss, line 280: This is a very complex method. Can you find ways to simplify it?

app/controllers/courses_controller.rb
```diff
@@ -269,6 +271,29 @@ def update_courses_wikis
...
+    wiki_ns_obj = courses_namespaces_params[:namespaces]
+    wiki_ns_obj.each do |obj|
```
ragesoss, line 282: What kind of object is this? There's almost always a better variable name than `obj`.

app/models/wiki.rb
```diff
@@ -63,6 +63,19 @@ class Wiki < ApplicationRecord
     'wikisource' => 'wikisource.org'
   }.freeze
 
+  PROJECTS_NAMESPACES = {
```
ragesoss, line 66: A comment should document where these namespaces are described/listed.

app/services/update_namespace_stats.rb
```diff
@@ -0,0 +1,88 @@
...
+          'user_count': user_count,
+          'word_count': word_count,
+          'references_count': references_count,
+          'views_count': views_count
+        }
+      end
+      stats[@wiki.domain] = wiki_stats
+    end
+    
+    crs_stat = CourseStat.find_by(course_id: @course.id) || CourseStat.create(course_id: @course.id)
```
ragesoss, line 33: I prefer avoiding abbreviations like this in variable names. Better to use a longer name and spell out the words more explicitly, to make it easier to read and figure out what's going on.

app/models/course.rb
```diff
@@ -282,6 +284,17 @@ def tracked_revisions
              .where(wiki_id: wiki_ids)
   end
 
+  def tracked_namespaces
```
ragesoss, line 287: I don't like this data structure very much. At least on the Ruby side of things, I think it will be easier to work with the `courses_namespaces` objects directly.

For example, in `UpdateNamespaceStats#update_stats`, you don't need to have nested loops if you do something like:
```
@course.courses_namespaces.each do |c_ns|
  wiki = c_ns.wiki
  namespace = c_ns.namespace
  ...
end

Similarly, you can probably serialize the courses_namespaces objects more directly, instead of doing this somewhat complex mapping.

File: app/models/courses_namespaces.rb [hunk 14]
```diff
@@ -0,0 +1,3 @@
+class CoursesNamespaces < ApplicationRecord
```
ragesoss, line 1: The more I review this, the more I think this object needs a different name. It's fundamentally a record that specifies both a wiki and a namespace for a particular course (not just a namespace for the course).

Maybe `CourseWikiNamespaces` or even `TrackedNamespaces`?

The idea is that this class (like all of the ActiveRecord classes in this app) is not just an interface for database data, but it's also a tool for helping us think clearly about the data, packaging up a set of data or relationships between data so that we don't have to think about more specific data structures (eg, an array of wiki ids and namespace integers) when we can avoid it.

app/models/course_data/articles_courses.rb
```diff
@@ -100,7 +101,7 @@ def self.update_all_caches(articles_courses)
 
   def self.update_from_course(course)
     course_article_ids = course.articles.where(wiki: course.wikis).pluck(:id)
-    revision_article_ids = get_mainspace_revisions(course.revisions).distinct.pluck(:article_id)
+    revision_article_ids = course.revisions.distinct.pluck(:article_id)
```
ragesoss, line 104: Does this change mean that ArticlesCourses records will be created for all namespaces for which revisions exist, regardless of whether it's a tracked namespace?

Maybe there's a way to use `revisions` scopes to get just the revisions from the namespaces we care about? Or maybe this needs to be done by looping through wikis/namespaces? I think we'll need to avoid creating ArticlesCourses records for non-tracked namespaces.

app/services/update_namespace_stats.rb
```diff
@@ -0,0 +1,88 @@
...
+    crs_stat.stats_hash['namespace_stats'] = {}
```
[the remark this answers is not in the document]
vaidehi44: Yeah, the structure is right.
Okay, if we keep a parallel structure, we'll still need to specify the WIKI_DATA_STATS and namespace specific stats for wikidata inside it?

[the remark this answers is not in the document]
vaidehi44: So, in order to be more explicit, I has used the key `namespace_stats`. 
Is it okay, if we have a parallel structure with top-level keys as wiki domains, and then something like `namespace_overview_stats` and `wiki_overview_stats` keys. Ofcourse, as of now, `wiki_overview_stats` would only exist for wikidata.

app/models/courses_namespaces.rb
vaidehi44: okay, I think `CourseWikiNamespaces` is good, should I go ahead with it?

app/assets/javascripts/components/common/wiki_overview_stats.jsx
ragesoss: `OverviewStatsTabs`?

"overview" identifies it a component of the Overview tab, and "stats tabs" identifies it as the tabbed stats that show up separately from the static overall stats at the top.

app/assets/javascripts/styles/select.js
```diff
@@ -25,13 +25,15 @@ const selectStyles = {
...
     backgroundColor: '#F2F2F2'
   }),
-  multiValueRemove: base => ({
-    ...base,
-    color: '#676EB4',
-    backgroundColor: '#F2F2F2',
-    borderRadius: 0,
-    '&:hover': { backgroundColor: '#D9D9D9', color: '#676EB4' }
-  })
+  multiValueRemove: (base, props) => (
```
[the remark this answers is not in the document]
ragesoss: I think we'll want to (ideally, eventually) support removing mainspace from the set of tracked namespaces, but I think for now, treating it as an automatically handled default that can't be disabled would be fine. I don't think Mainspace needs to show up on this picker (for now) and shouldn't have any CourseWikiNamespace records.

app/services/update_namespace_stats.rb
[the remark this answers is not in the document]
ragesoss: I'm not sure what the best structure would be.

I think we want to design the data structure in a way that makes it easy to think about how the stats will be used... namely, different sets of stats become different tabs of a stats component.

The one that matches that most would be one top-level key for each tab of stats. So maybe something like this?

```
{
  'www.wikidata.org' => WIKIDATA_STATS,
  'en.wikipedia.org-namespace-0' => ENWIKI_MAINSPACE_STATS,
  'en.wikibooks.org-namespace-102' => WIKIBOOKS_COOKBOOT_STATS
}
```

At least as far as we've identified so far, there aren't any features where it would be important to have different namespaces of the same wiki collected together for stats, separately from additional namespaces for other wikis for the same course.

app/models/courses_namespaces.rb
ragesoss: sure

ragesoss:
PR 1 should just have frontend changes for displaying stats. Basically, it should just be able to handle a set of manually created namespace stats in the CourseStats object for a course. The UI for selecting namespaces in edit mode would be part of PR 3.

app/assets/javascripts/components/common/wiki_overview_stats.jsx
vaidehi44: Alright.

PR #5053, opened 2022-07-17
remove moment recur as a dependency

TheTrio:
This is WIP


ragesoss:
Were the course_date_utils tests useful?

app/assets/javascripts/components/common/date_picker.jsx
```diff
@@ -92,7 +92,7 @@ const DatePicker = createReactClass({
    * @return {String} formatted date
    */
   getFormattedDateTime() {
-    return CourseDateUtils.formattedDateTime(this.getDate(), this.props.showTime);
+    return CourseDateUtils.formattedDateTime(this.getDate().toDate(), this.props.showTime);
```
ragesoss, line 95: Why is this done here instead of within `getDate()`?

app/assets/javascripts/utils/course_date_utils.js
```diff
@@ -21,59 +32,53 @@ const CourseDateUtils = {
...
       timeline_start: {
         minDate: startDate,
-        maxDate: moment(course.timeline_end, 'YYYY-MM-DD')
+        maxDate: parseISO(course.timeline_end)
       },
       timeline_end: {
-        minDate: moment(course.timeline_start, 'YYYY-MM-DD'),
-        maxDate: moment(course.end, 'YYYY-MM-DD')
+        minDate: parseISO(course.timeline_start),
+        maxDate: parseISO(course.end)
```
ragesoss, line 57: Does `parseISO(datetime)` do the same thing as moment(datetime, 'YYYY-MM-DD'), or is this changing the format of these `dateProps` values?

TheTrio: The format is not being changed - `moment(datetime, 'YYYY-MM-DD')` simply parsed the given string(`datetime`) according to the given format(`YYYY-MM-DD`) to return a moment date object. 

`parseISO` will fail if the date can't be parsed according to the given format.

app/assets/javascripts/utils/course_date_utils.js
```diff
@@ -21,59 +20,53 @@ const CourseDateUtils = {
...
     }
-    if (moment(updatedCourse.timeline_start, 'YYYY-MM-DD').isAfter(updatedCourse.timeline_end, 'YYYY-MM-DD') && valueKey !== 'timeline_end') {
+    if (isAfter(toDate(updatedCourse.timeline_start), toDate(updatedCourse.timeline_end)) && valueKey !== 'timeline_end') {
       updatedCourse.timeline_end = updatedCourse.timeline_start;
     }
-    if (moment(updatedCourse.timeline_end, 'YYYY-MM-DD').isAfter(updatedCourse.end, 'YYYY-MM-DD') && valueKey !== 'end') {
+    if (isAfter(toDate(updatedCourse.timeline_end), toDate(updatedCourse.end)) && valueKey !== 'end') {
       updatedCourse.end = updatedCourse.timeline_end;
     }
-    if (moment(updatedCourse.start, 'YYYY-MM-DD').isAfter(updatedCourse.timeline_start, 'YYYY-MM-DD') && valueKey !== 'timeline_start') {
```
ragesoss, line 73: All the other parts of this are being converted to not use `moment` but this one is just being removed. Does that change the behavior?

app/assets/javascripts/components/common/date_picker.jsx
TheTrio: The rest of the file still uses moment js so `getDate()` still returns a moment date object.

app/assets/javascripts/utils/course_date_utils.js
```diff
@@ -93,49 +86,42 @@ const CourseDateUtils = {
...
-    weekEnd.day(6);
-    let weekStart = recurrence.startDate();
-    const firstWeekStart = recurrence.startDate().day();
-    weekStart.day(0);
-    const courseWeeks = Math.ceil(weekEnd.diff(weekStart, 'weeks', true));
-    if (!recurrence.rules || recurrence.rules[0].measure !== 'daysOfWeek') {
-      return [];
-    }
-
+  weekMeetings(course, exceptions) {
```
ragesoss, line 133: Can you explain more about the change to this function and why the first argument (and the `meetings` function) are not needed?

app/assets/javascripts/utils/course_date_utils.js
ragesoss: Yeah, it looks like that's the case. I see I added it in https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/2b9ae07a8764eba7c84ec7d44da78c7a2d46e86e but I guess that duplicated part must have been useless. I can't think of any way this would change behavior.

PR #5064, opened 2022-07-21
Frontend changes for showing tabbed course overview stats

vaidehi44:
## What this PR does
This PR contains frontend changes corresponding to tabbed overview stats presented on course overview page.

The structure for course_stats object is -

```
{
  'www.wikidata.org' => WIKIDATA_STATS,
  'en.wikipedia.org-namespace-0' => ENWIKI_MAINSPACE_STATS,
  'en.wikibooks.org-namespace-102' => WIKIBOOKS_COOKBOOT_STATS
} 
```

## Screenshots
image
image


ragesoss:
I think there's some room for design improvements for the tabs. I'm still unsure whether showing all the tabs like this would be better than have a dropdown to select between them, but here's the design (by Toqa, a designer I hired to work on it a few months ago) that we have for it:

Final 1

Final 2

app/assets/javascripts/components/common/OverviewStats/namespace_overview_stats.jsx
```diff
@@ -0,0 +1,65 @@
...
+      />
+      <OverviewStat
+        id="total-edits"
+        className="stat-display__value"
+        stat = {statistics.revision_count}
+        statMsg={I18n.t('metrics.edit_count_description')}
+        renderZero={true}
+      />
+      <OverviewStat
+        id="student-editors"
```
ragesoss, line 30: I don't think the user count needs to be in any namespace-specific stats.

app/assets/javascripts/components/common/overview_stats_tabs.jsx
```diff
@@ -0,0 +1,99 @@
...
+import NamespaceOverviewStats from './OverviewStats/namespace_overview_stats';
+import WikidataOverviewStats from './wikidata_overview_stats';
+
+import ArticleUtils from '../../utils/article_utils';
+
+
+const OverviewStatsTabs = ({ statistics }) => {
+  const [currentTabId, setCurrentTabId] = useState(0);
+
+  const getTabTitle = (ns_id, wiki) => {
```
ragesoss, line 14: These two functions should probably be in a utils file rather than in this component. Tests for them would also be good.

app/assets/javascripts/components/common/overview_stats_tabs.jsx
```diff
@@ -0,0 +1,99 @@
...
+    return setCurrentTabId(Number(e.currentTarget.id));
+  };
+
+  const hasWikidataStats = () => {
+    if (statistics['www.wikidata.org']) return true;
+    return false;
+  };
+
+  const statsDataList = [];
+  // if there are wikidata overview stats, they should be displayed on first tab
```
ragesoss, line 38: This and the forEach loop are more data manipulation than I'd like to see. If possible, it would be best to do this in a way that doesn't special-case Wikidata. (I don't think it necessarily needs to be first.)

app/assets/javascripts/components/students/components/Articles/SelectedStudent/RevisionsList/StudentRevisionsList.jsx
```diff
@@ -8,15 +8,14 @@ import StudentRevisionRow from './StudentRevisionRow';
...
-import ArticleUtils from '~/app/assets/javascripts/utils/article_utils.js';
 import studentListKeys from '@components/students/shared/StudentList/student_list_keys.js';
 
 export class StudentRevisionsList extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
       isOpen: false,
-      namespace: 'all'
+      namespace: -1 // show all namespace revisions by default
```
ragesoss, line 18: For a code comprehensibility perspective, I don't think replacing strings with numbers here is helpful. Do the changes here have any effect on behavior? If they are unrelated to displaying stats on the Overview tab, they probably shouldn't be included in this PR.

ragesoss:
I think the namespace titles displayed in the tabs could be something like: `en.wikibooks.org - Cookbook`, `en.wikipedia.org - Mainspace` and `www.wikidata.org`.

It would be good to treat the titles in parallel, so that whatever parsing of the stats hash key is used, it works the same way for keys that include a namespace and those that don't. (ie, it should be able to handle `www.wikidata.org`, `www.wikidata.org-namespace-120`, `en.wikibooks.org`, and `en.wikibooks.org-102`.)

app/assets/javascripts/components/students/components/Articles/SelectedStudent/RevisionsList/StudentRevisionsList.jsx
vaidehi44: These changes are related to the namespace filter of revisions table.
Actually, earlier this code used `namespaceToId` function from ArticleUtils. But, now that I have defined new mapping function there (`NamespaceIdMapping`), that function would have become its duplicate. Hence, some changes reflected here.
Should I open another pr then?

vaidehi44:
> I think there's some room for design improvements for the tabs. I'm still unsure whether showing all the tabs like this would be better than have a dropdown to select between them, but here's the design (by Toqa, a designer I hired to work on it a few months ago) that we have for it:
> 
> Final 1
> 
> Final 2

I guess, right now the only reason for not having tabs would be because they can extend horizontally beyond page. But, if it is handled and made scrollable or something like that, I would prefer tabs than dropdowns :).

vaidehi44:
Should I also modify the UI for the aggregated mainspace stats at the top, according to the design?

app/assets/javascripts/components/common/OverviewStats/namespace_overview_stats.jsx
vaidehi44: Won't it give insight of how many students worked on particular tracked namespace?

ragesoss:
No, don't change the style of the overview stats.

vaidehi44:
> It would be good to treat the titles in parallel, so that whatever parsing of the stats hash key is used, it works the same way for keys that include a namespace and those that don't. (ie, it should be able to handle `www.wikidata.org`, `www.wikidata.org-namespace-120`, `en.wikibooks.org`, and `en.wikibooks.org-102`.)

I didn't get this. Should the namespace titles and stats title be same? Or there shouldn't be any stats title?

ragesoss:
maybe, but I don't think that's something that is a priority to display.
although it will be easy to change later.

ragesoss:
the domain, along with the namespace if present, should be a sufficient
title in all cases.

app/assets/javascripts/components/common/overview_stats_tabs.jsx
vaidehi44: Also, I got this conceptual question. For each tab we have a different content rendered by different component. So, is it better to create this components, store and use them (like I am doing now), or is it better to initialize and render them instantaneously when a tab is selected?

I think in our case, the compoents are not that heavy and so they can probably be stored before hand.

vaidehi44:
@ragesoss I have pushed most of the changes suggested except tabs UI. I feel that the tabs design which you have shared is too ordinary :sweat_smile:. So, I tried out some stuff myself. Do you think, we can may be try something from these, or anything else?

Screenshot (28)

Screenshot (29)


Screenshot (38)

ragesoss:
I like the third one!

app/assets/javascripts/components/common/overview_stats_tabs.jsx
```diff
@@ -0,0 +1,64 @@
...
+        active={currentTabId === index}
+      />
+    );
+    index += 1;
+  });
+
+  const title = statsList[currentTabId].statsTitle;
+  const data = statsList[currentTabId].statsData;
+  const content = (title === 'www.wikidata.org')
+    ? <WikidataOverviewStats statistics={data} isCourseOverview={true}/>
```
ragesoss, line 42: It might make sense here to implement another simple component (eg, `OverviewStatsTab` (singular)), so that this tabs component has a simple job: render all the tabs and render the content for the current tab. The new component would be in charge of which component to render for the current tab content.

app/assets/javascripts/components/common/wikidata_overview_stats.jsx
```diff
@@ -3,10 +3,16 @@ import PropTypes from 'prop-types';
 import OverviewStat from './OverviewStats/overview_stat';
 import I18n from 'i18n-js';
 
-const WikidataOverviewStats = ({ statistics }) => {
+const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
+  let containerClass = 'wikidata-stats-container';
+  let title = 'Wikidata stats';
```
ragesoss, line 8: It looks like this title isn't used in the the latest screenshots you shared?

app/assets/javascripts/components/students/components/Articles/SelectedStudent/RevisionsList/StudentRevisionsList.jsx
ragesoss: I think this is okay, except... is there any reason to use `-1` in place of `all`? The other integer values all map to a mainspace, but this one does not and `-1` not as easy to interpret as `all`.

spec/features/wikidata_course_spec.rb
```diff
@@ -23,7 +23,7 @@
 
   it 'shows Wikidata stats on the Home tab' do
     visit "/courses/#{course.slug}"
-    expect(page).to have_content('Wikidata stats')
+    expect(page).to have_content('www.wikidata.org')
```
ragesoss, line 26: I'd like this PR to include a ruby feature spec for having multiple tabs and interacting with them. This file could be renamed `detailed course stats` or similar, and then the spec could be basically the same as this Wikidata one except with multiple tabs worth of data in the CoursesStats record.

The spec should test clicking through different tabs and seeing the different stats render.

app/assets/javascripts/components/students/components/Articles/SelectedStudent/RevisionsList/StudentRevisionsList.jsx
vaidehi44: Actually, I don't know what number to use for 'all', because for other options it would be their namespace id. May be, we can still use the strings - 'all', 'article', 'talk', etc.

ragesoss: I think a mix of numbers (for actual namespaces) and a string for `all` would be okay. But also, using strings for all of the values seems find to me.

app/assets/javascripts/components/common/overview_stats_tabs.jsx
vaidehi44: Well, I have already created a component named `OverviewStatsTab` which renders a tab.
May be we can name it `OverviewStatsContent` or `OverviewStatsData` or anything like that. And, you are saying that the logic of identifying the content (whether it is wikidata or namespace stats) according to active tab, should be done in this component? Instead of having `statsList`, we should use this new component?

app/assets/javascripts/components/common/wikidata_overview_stats.jsx
vaidehi44: Yes, it is used. In the last one. Here, the title isn't "Wikidata stats", but "www.wikidata.org". I thought that this would be in accordance with other namespace stats and their title. So, "Wikidata Stats" would be only shown for campaign overview stats. Is it okay?

spec/features/wikidata_course_spec.rb
vaidehi44: okay!

app/assets/javascripts/components/common/overview_stats_tabs.jsx
ragesoss: You're right, `OverviewStatsTab` would be a bad name for it, and the component you have now called that is appropriately named. `OverviewStatsContent` sounds good.

And yes, I think having this component identify wikidata vs namespace stats should be done in that new component.

vaidehi44: @ragesoss, I have created the `OverviewStatsContent` component. Please check it along with `OverviewStatsTabs` to confirm if that's okay.

ragesoss: yes, it looks good to me.

app/presenters/courses_presenter.rb
```diff
@@ -151,9 +151,9 @@ def trained_percent
   end
 
   def wikidata_stats
-    stats ||= courses.joins(:course_stat).where.not(course_stats: nil).map do |course|
+    stats ||= courses.joins(:course_stat).where.not(course_stats: nil).map { |course|
       course.course_stat.stats_hash['www.wikidata.org']
-    end
+    }.compact
```
ragesoss, line 156: I think rubocop will complain about using braces for a multi-line block instead of do/end. You can still chain the do/end syntax with `end.compact`.

vaidehi44: ohkk, thanks.

ragesoss:
@vaidehi44 can you include before/after screenshots for the case of just Wikidata stats?

app/presenters/courses_presenter.rb
```diff
@@ -151,9 +151,9 @@ def trained_percent
   end
 
   def wikidata_stats
-    stats ||= courses.joins(:course_stat).where.not(course_stats: nil).map { |course|
+    stats ||= courses.joins(:course_stat).where.not(course_stats: nil).filter_map do |course|
```
ragesoss, line 154: nice

vaidehi44:
> @vaidehi44 can you include before/after screenshots for the case of just Wikidata stats?

### Before

image

### After

image

ragesoss:
Okay! I think this is ready to merge. Anything you can think of that should be done before that?

vaidehi44:
@ragesoss, I think, one thing which needs to be done is to handle overflow of tabs if there are more number of tabs. Of course, this can be done later too as now in prod there won't be any tabs.

ragesoss:
Good point. Do you have a plan for that? I'm fine with either merging this now and handling tab overflow in a separate PR, or including it in this one.

vaidehi44:
I have some ideas, but not a perfect plan. So, I think I can open a seperate a PR for this. We can may be try different things there.

ragesoss:
I think either way is fine. Usually I prefer letting each component decide
what to render, rather than passing components as props for others to
render.

PR #5075, opened 2022-07-27
Add service to update namespace stats.

vaidehi44:
## What this PR does
Defines a service to generate/update namespace stats to CourseStats.

## Steps to reproduce for cookbook namespace

1.  Copy/ Import the course -https://outreachdashboard.wmflabs.org/courses/Edriiic/Nigerian_Cuisine
2.  In rails console, 
```
course = Course.find_by(title: "Nigerian Cuisine");
wiki = course.wikis.find_by(project: "wikibooks");
```
3.  Update revisions with `UpdateCourseStats.new(course)`
4.  Update namespace stats with `UpdateWikiNamespaceStats.new(course, wiki, 102)`

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -283,7 +283,7 @@ class Course < ApplicationRecord
   end
 
   def tracked_namespaces
-    return ['0']
+    return [0]
   end
 
   def suspected_plagiarism
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -282,6 +282,10 @@ class Course < ApplicationRecord
              .where(wiki_id: wiki_ids)
   end
 
+  def tracked_namespaces
+    return ['0']
+  end
+
   def suspected_plagiarism
     revisions.suspected_plagiarism
   end
--- a/app/services/update_wiki_namespace_stats.rb
+++ b/app/services/update_wiki_namespace_stats.rb
@@ -11,13 +11,13 @@ class UpdateWikiNamespaceStats
   def update_stats
     stat_key = "#{@wiki.domain}-namespace-#{@namespace}"
     stats = {
-      'edited_count': edited_articles_count,
-      'new_count': new_articles_count,
-      'revision_count': revision_count,
-      'user_count': user_count,
-      'word_count': word_count,
-      'reference_count': reference_count,
-      'view_count': view_count
+      edited_count: edited_articles_count,
+      new_count: new_articles_count,
+      revision_count: revision_count,
+      user_count: user_count,
+      word_count: word_count,
+      reference_count: reference_count,
+      view_count: view_count
     }
     course_stats = CourseStat.find_or_create_by(course_id: @course.id)
     course_stats.stats_hash[stat_key] = stats
@@ -27,20 +27,20 @@ class UpdateWikiNamespaceStats
   # live, tracked course revisions filtered by wiki and namespace
   def revisions_filtered_by_wiki_namespace
     @course.revisions.where.not(article_id: @course.articles_courses.not_tracked.pluck(:article_id))
-                     .where(wiki_id: @wiki.id)
-                     .namespace(@namespace).live
+           .where(wiki_id: @wiki.id)
+           .namespace(@namespace).live
   end
 
   # live, tracked articles filtered by wiki and namespace
   def articles_filtered_by_wiki_namespace
     @course.articles_courses.joins(:article).where(articles: { wiki: @wiki, namespace: @namespace })
-                                            .tracked.live
+           .tracked.live
   end
 
   def edited_articles_count
     articles = articles_filtered_by_wiki_namespace
     articles.count
-	end
+  end
 
   def new_articles_count
     articles = articles_filtered_by_wiki_namespace
--- a/app/services/update_wiki_namespace_stats.rb
+++ b/app/services/update_wiki_namespace_stats.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+class UpdateWikiNamespaceStats
+  def initialize(course, wiki, namespace)
+    @course = course
+    @wiki = wiki
+    @namespace = namespace
+    update_stats
+  end
+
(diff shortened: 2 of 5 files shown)


app/services/update_wiki_namespace_stats.rb
```diff
@@ -0,0 +1,75 @@
...
+  # live, tracked course revisions filtered by wiki and namespace
+  def revisions_filtered_by_wiki_namespace
+    @course.revisions.where.not(article_id: @course.articles_courses.not_tracked.pluck(:article_id))
+                     .where(wiki_id: @wiki.id)
+                     .namespace(@namespace).live
+  end
+
+  # live, tracked articles filtered by wiki and namespace
+  def articles_filtered_by_wiki_namespace
+    @course.articles_courses.joins(:article).where(articles: { wiki: @wiki, namespace: @namespace })
```
ragesoss, line 36: This doesn't include generating ArticlesCourses for non-mainspace articles, correct? So this will always be zero for this PR?

ragesoss:
What additional changes would be necessary to ensure that the correct stats are being generated (via an rspec test)?

I guess generating ArticlesCourses for non-mainspace articles is the main thing?

app/services/update_wiki_namespace_stats.rb
vaidehi44: Oh right. I missed it. Will correct it.

vaidehi44: @ragesoss, right now tracked_namespaces are not known, so should I enable article_courses for all namespaces?

ragesoss: no, i don't think so. I think it would be better to include a minimal version of tracked_namespaces in this PR such that it could be deployed without changing current behavior, but would still allow for testing.

One approach would be to define `tracked_namespaces` for `course.rb` but just have it return `[0]` (ie, hard-code the current behavior of mainspace being the only tracked namespace for any course), and then in the test you can stub that method to override it and add cookbook. This would be the simplest way i could think of for this PR, although it would mean a small amount of rewriting for the next stage.

Another approach would be include the new database table in this PR, so that tracked_namespaces could be determined from the database (which will be necessary for the later PRs anyway).

vaidehi44: But, are you going to deploy it right away? I mean before we have `tracked_namespaces` from `CourseWikiNamespaces` ?

ragesoss: Maybe! Part of the point of the separate PRs is to make it possible to deploy before everything else is ready.

ragesoss:
Now with the `tracked_namespaces`, you should be able to write a test for generating ArticlesCourses records and calculating stats for a Cookbook course. This will probably entail something like `allow_any_instance_of(Course).to receive(:tracked_namespacse).and_return([0, 102])` or similar.

app/models/course.rb
```diff
@@ -282,6 +282,10 @@ def tracked_revisions
              .where(wiki_id: wiki_ids)
   end
 
+  def tracked_namespaces
+    return ['0']
```
ragesoss, line 286: Should this be an integer instead of a string?

vaidehi44: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/dbf2c87e75c2cbfeafb552986eff16ea563f62a3/app/models/course_data/articles_courses.rb#L131-L133
Yeah, but in articles_courses it was a string, so I kept it a string.

ragesoss: Ah. I don't think there's any reason for it to be a string, as namespace is an integer in the database.

vaidehi44: okay.

spec/services/update_wiki_namespace_stats_spec.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe UpdateWikiNamespaceStats do
```
ragesoss, line 5: This look solid!

spec/services/update_wiki_namespace_stats_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+  it 'updates wiki-namespace stats with appropriate keys' do
+    stats = course.course_stat.stats_hash['en.wikibooks.org-namespace-102']
+
+    expect(stats).to have_key(:edited_count)
+    expect(stats).to have_key(:new_count)
+    expect(stats).to have_key(:revision_count)
+    expect(stats).to have_key(:user_count)
+    expect(stats).to have_key(:word_count)
+    expect(stats).to have_key(:reference_count)
+    expect(stats).to have_key(:view_count)
```
ragesoss, line 52: What values do these have? Does this test exercise the generation of non-zero values for all of these stats? (I'm guessing no for the view count, and yes for the most of the rest?)

vaidehi44: Actually, I haven't checked for the values.
So, should I get the count for each attribute and write a test for it?

vaidehi44: Also, I wasn't sure if I should include the test where it checks for each attribute in key. Should I keep it?

ragesoss: Yes. We might find that such values can vary a little bit over time (if any of the edits get deleted at some point) but for now, I think it's better to have a little extra test focus on verifying that these stats are actually being counted the way we think they should be.

ragesoss: > Also, I wasn't sure if I should include the test where it checks for each attribute in key. Should I keep it?

Yes, I think that's fine. It makes it a little more explicit what shape we expect this data to take.

vaidehi44: @ragesoss, have updated the test.

PR #5093, opened 2022-08-03
Update react dnd

TheTrio:
This upgrades [`react-dnd`](https://react-dnd.github.io/react-dnd/) to v16. From this version, the library only supports functional components via hooks. I've rewritten the HOC to use hooks instead.

With this, `vendors.js` goes from `1.35MB` to `1.32MB`


ragesoss:
Are there any differences in behavior?

TheTrio:
None that I've noticed so far.

ragesoss:
Excellent choice to extract the drag hook. This works smoothly; I also don't notice any problems or differences with the timeline drag-and-drop behavior.

PR #5098, opened 2022-08-05
reduce usage of JQuery

TheTrio:
This PR attempts to remove all the instances of JQuery from all routes except for

1. the campaigns route
2. the survey route

JQuery is only downloaded when the user visits these routes.

Also, instead of including `jquery-ui` in the vendors bundle, I've imported it dynamically so that it only gets loaded when its required.

This reduces `vendors.js` from `1.32MB` to `1.27MB`


app/helpers/application_helper.rb
```diff
@@ -24,7 +24,11 @@ def dashboard_stylesheet_tag(filename)
   end
 
   def hot_javascript_tag(filename)
-    javascript_include_tag hot_javascript_path(filename)
+    js_path = hot_javascript_path(filename)
+    if (filename == 'campaigns') || (filename.include? 'survey')
```
ragesoss, line 28: any way to avoid this?

TheTrio: One way would to be to have do this all manually in all the haml files which need JQuery. Or I suppose I could include the JQuery from the JS itself. What do you think?

app/assets/javascripts/components/timeline/block.jsx
```diff
@@ -37,7 +37,7 @@ const Block = createReactClass({
   },
 
   passedUpdateBlock(selectedIds) {
-    const newBlock = $.extend(true, {}, this.props.block);
+    const newBlock = structuredClone(this.props.block);
```
ragesoss, line 40: Do we have polyfills to ensure that this works across browsers? Browser support for this just dates to early 2022, and elsewhere in the codebase we use `Object.assign` for this kind of thing.

app/assets/javascripts/components/wizard/option.jsx
```diff
@@ -16,13 +16,18 @@ const Option = createReactClass({
     selectWizardOption: PropTypes.func.isRequired
   },
 
+  getInitialState() {
+    this.descriptionRef = React.createRef(null);
```
ragesoss, line 20: Is `getInitialState` the best place for this, or should it be in a different lifecycle function?

app/assets/stylesheets/modules/_user_profiles.styl
```diff
@@ -216,3 +216,13 @@
...
+
+html
+	scroll-behavior smooth
+	overflow scroll
+	scroll-padding-top 100px
+
+#userprofile_navbar
+	position sticky
+	top 50px
+	z-index 10
```
ragesoss, line 228: What are these changes for?

app/assets/stylesheets/modules/_wizard.styl
```diff
@@ -193,12 +193,18 @@ wizard-width = 860px
...
         width 100%
+        transition all .2s ease-in-out
+        opacity 0
         &.open
           border-color mischka
+      .wizard__option__description.open
+        height 100%
+        margin-top 18px
+        padding 15px 0 0
+        opacity 1
```
ragesoss, line 206: What are these changes for?

app/helpers/application_helper.rb
ragesoss: I think including the jquery from the campaigns and surveys js files makes the most sense, just in terms of making sure the js files are the places that keep the details about what they require to work properly.

app/assets/javascripts/components/timeline/block.jsx
TheTrio: Yeah `Object.assign` would be simpler but it doesn't clone the object recursively(which the JQuery version does)

`core-js` does have polyfills for [`structuredClone`](https://github.com/zloirock/core-js#structuredclone) so it shouldn't be an issue.

app/assets/javascripts/components/wizard/option.jsx
TheTrio: This should be done in the constructor but because we're not using ES6 classes that doesn't exist. In [the docs](https://reactjs.org/docs/react-without-es6.html#setting-the-initial-state), they used `getInitialState` in place of the constructor so that's what I went with.

ragesoss: Cool.

app/assets/stylesheets/modules/_user_profiles.styl
TheTrio: I've removed [`users_profile.js`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/2f6fdea8e9c41698591754eb8c158f27955b63fc/app/assets/javascripts/utils/users_profile.js) and implemented the functionality using CSS instead.

app/assets/javascripts/components/timeline/block.jsx
ragesoss: Hmm... do we even need deep cloning here? I can't think of a reason we would.

app/assets/stylesheets/modules/_wizard.styl
TheTrio: This used to use JQuery's `slideToggle` to implement the closing and opening animation. This doesn't have a direct alternative in JS so I used some css animations instead.

app/assets/stylesheets/modules/_user_profiles.styl
ragesoss: Whoops, I missed that. :+1:

app/assets/javascripts/components/timeline/block.jsx
TheTrio: No idea :shrug: I went with it just to be safe. Should I just use `Object.assign`?

ragesoss: I think so. I'm pretty confident there's no reason why a shallow clone wouldn't be fine here. The `block` object is not that complicated, and here we're ultimately just replacing a block (as identified by its `id` attribute) in the redux state with a version of it that has been modified in an explicit way.

PR #5113, opened 2022-08-15
UI code for selecting namespaces for tracking

vaidehi44:
## What this PR does
The PR contains frontend code for selecting/tracking namespaces for a course.

## Screenshots
image

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -88,7 +88,8 @@
     "SurveyDetails": true,
     "Features": true,
     "WikiLanguages": true,
-    "WikiProjects": true
+    "WikiProjects": true,
+    "ProjectNamespaces": true
   },
   "env": {
     "jest": true,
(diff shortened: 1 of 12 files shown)


ragesoss:
I think this needs a UI message to explain the intended use / effect of Tracked Namespaces... namely, that pages in the tracked namespaces are the ones that contribute to the topline "pages edited", "words added" and similar stats, so you should only track the namespaces that your program is focused on improving.

ragesoss:
On a first survey of the code, this looks good! It's so much easier to see what's going on an why with this smaller PR, versus one that also includes backend changes.

vaidehi44:
> I think this needs a UI message to explain the intended use / effect of Tracked Namespaces... namely, that pages in the tracked namespaces are the ones that contribute to the topline "pages edited", "words added" and similar stats, so you should only track the namespaces that your program is focused on improving.

Okk ... but I didn't get what you meant by **topline** "pages eited"....

vaidehi44:
Also, should the message use tooltip? I guess as this is a new feature, it should be displayed more explicitly.

ragesoss:
I think a tooltip is okay, unless you can think of a better way to do it.

PR #5115, opened 2022-08-15
Backend code for updating namespaces to database

vaidehi44:
## What this PR does
This PR contains Rails code to receive the namespaces from React and save it appropriately to the CourseWikiNamespaces table.


app/models/course_wiki_namespaces.rb
```diff
@@ -0,0 +1,3 @@
+class CourseWikiNamespaces < ApplicationRecord
```
ragesoss, line 1: You should run `annotate` to generate the comments showing the columns of this table.

spec/factories/course_wiki_namespaces.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :course_wiki_namespaces do
+    namespace { 1 }
+    courses_wikis { nil }
```
ragesoss, line 4: Maybe this is related to InvalidForeignKey problem? There shouldn't be any CourseWikiNamespaces records that don't have a CoursesWikis parent record, correct?

app/controllers/courses_controller.rb
```diff
@@ -278,6 +280,27 @@ def update_courses_wikis
...
+  end
+
+  def update_course_wiki_namespaces
+    namespaces = course_wiki_namespaces_params
+
+    @course.courses_wikis.each do |course_wiki|
+      wiki_domain = course_wiki.wiki.domain
+      cw_ns = []
+      namespaces.each do |ns|
+        ns_wiki_domain = ns.split('-', 2)[0]
```
ragesoss, line 294: This code would be easier to understand with some comments showing what shape `namespaces` takes.

ragesoss:
This PR should include a change to `course#namespaces`, preserving the current behavior of making namespace 0 the default but using CourseWikiNamespaces if available.

vaidehi44:
> This PR should include a change to `course#namespaces`, preserving the current behavior of making namespace 0 the default but using CourseWikiNamespaces if available.

Okk, so I should update the current `tracked_namespaces` function from course.rb, right?

spec/factories/course_wiki_namespaces.rb
vaidehi44: Not sure about this. This is just a spec file definition, can it affect actual database processes? And, the issue gets resolved when I use `foregin_key : { on_delete: cascade }` in migration file. May be I am missing something or doing something wrong.

ragesoss:
> > This PR should include a change to `course#namespaces`, preserving the current behavior of making namespace 0 the default but using CourseWikiNamespaces if available.
> 
> Okk, so I should update the current `tracked_namespaces` function from course.rb, right?

Yes.

spec/factories/course_wiki_namespaces.rb
ragesoss: I think I misunderstood; I thought this PR still had that InvalidForeignKey error you mentioned. If the current migration in PR seems to work properly, that's fine.

ragesoss:
Looks like something related to the courses controller isn't in a working state yet. Can it be made compatible independently of the frontend PR?

vaidehi44:
> Looks like something related to the courses controller isn't in a working state yet. Can it be made compatible independently of the frontend PR?

May be it is because it expects `namespaces` param coming from frontend.

vaidehi44:
> Looks like something related to the courses controller isn't in a working state yet. Can it be made compatible independently of the frontend PR?

@ragesoss, are you getting any kind of error when courses_controller executes?
And, how can I make it independent of frontend changes? As I had asked earlier, could I temporarily add a pseudo namespace param for every incoming request/course object?

ragesoss:
> how can I make it independent of frontend changes?

One way would be to make the controller that handles the namespaces treat that set of data as optional, so that hitting the `CoursesController#update` action will continue to work with the current frontend code, and simply skip the path for updating namespaces when no namespace data is received.

The controller behavior could still be tested using a controller spec (as opposed to a browser-based spec that relies on the frontend).

app/controllers/courses_controller.rb
```diff
@@ -278,6 +280,29 @@ def update_courses_wikis
     @course.update_wikis(new_wikis)
   end
 
+  def course_wiki_namespaces_params
+    params.require(:course).permit(namespaces: [])[:namespaces]
+  end
+
+  def update_course_wiki_namespaces
+    namespaces = course_wiki_namespaces_params
```
ragesoss, line 288: Adding a guard clause here (`return unless namespaces.present?` or similar) might be all that's needed to make this PR independent of frontend changes and able to be merged and deployed separately.

vaidehi44: Quite surprisingly for me, even without the guard clause, it worked just fine. The controller did not raise any error.
Should I still add a guard clause here?

vaidehi44: Okay....the clause is anyways necessary as some spec tests are failing...have added the guard clause now.

ragesoss:
This still doesn't include the change to `course#tracked_namespaces` that we discussed.

vaidehi44:
Oh yeah, I guess I forgot to push them. Let me push the changes.

app/models/course_data/articles_courses.rb
```diff
@@ -132,7 +131,15 @@ def self.destroy_invalid_records(course, valid_article_ids)
     end
   end
 
-  def self.get_revisions_by_namespaces(namespaces, revisions)
-    revisions.joins(:article).where(articles: { namespace: namespaces })
+  def self.get_revisions_by_namespaces(course)
+    # Select all mainspace article revisions
```
vaidehi44, line 135: @ragesoss, is this fine?

app/models/course.rb
```diff
@@ -284,10 +284,6 @@ def tracked_revisions
              .where(wiki_id: wiki_ids)
   end
 
-  def tracked_namespaces
```
ragesoss, line 285: Rather than removing this, I think it should be replaced with an implementation that tries to return an array based on the course_wiki_namespaces and falls back to `[0]` if there are none. We want to allow for tracking an alternative namespace _instead of_ mainspace.

app/models/course_data/articles_courses.rb
ragesoss: See my comment on `#tracked_namespaces`.

app/models/course.rb
vaidehi44: Oh...so you mean we won't be storing all mainspace articles by default unless user has no tracked namespaces ?

vaidehi44: I thought the default behavior of including all the mainspace articles was still there...

ragesoss: I think that would be best.

vaidehi44:
@ragesoss, I have done some changes to `course.rb` and `articles_courses.rb`. As tracked namespaces would depend on both wiki and namespace, it can't simply be a numerical array. So, I have defined a wiki-namespace object there,

Also, currently some topline stats, they are still showing stats corresponding to all mainspace articles only. Like total edits, words added, references added, etc. So, I guess, that also needs to be corrected when tracked_namespaces is not null.

app/models/course.rb
```diff
@@ -283,7 +285,11 @@ def tracked_revisions
   end
 
   def tracked_namespaces
-    return [0]
+    course_wiki_namespaces.map do |course_wiki_ns|
+      wiki = course_wiki_ns.courses_wikis.wiki
+      namespace = course_wiki_ns.namespace
+      { :wiki => wiki, :namespace => namespace }
```
ragesoss, line 291: I think it will be important for the fallback/default behavior of tracking mainspace to be defined here, rather than (for example)  in ArticlesCourses#get_revisions_by_namespaces. Otherwise, we may end up needing to account for it in multiple places independently.

As you pointed out, I didn't account for this needing to return an array wiki+namespace combinations rather than just an array of namespace integers.

Here's how I think it should work: for each tracked wiki, if there is a course_wiki_namespace record, add a hash for that Wiki+Namespace, and otherwise, include a hash for Wiki+Mainspace.

vaidehi44: @ragesoss, I have updated it accordingly.

ragesoss:
@vaidehi44 Looks like `UpdateWikiNamespaceStats` spec needs to be updated to use use the CourseWikiNamespaces instead of `allow_any_instance_of(Course).to receive(:tracked_namespaces).and_return([0, 102])`.

I was able to get it passing with some adjustments to that spec, like this:

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe UpdateWikiNamespaceStats do
  let(:course) { create(:course, start: Date.new(2022, 8, 1), end: Date.new(2022, 8, 2)) }
  let(:wikibooks) { Wiki.get_or_create(language: 'en', project: 'wikibooks') }
  let(:cookbook_ns) { 102 }
  let(:user1) { create(:user, username: 'Jamzze') } # user with cookbook edits
  let(:user2) { create(:user, username: 'FieldMarine') } # user with other namespace edits
  let(:cookbook_course_wiki) { create(:courses_wikis, course: course, wiki: wikibooks) }

  before do
    stub_wiki_validation

    create(:course_wiki_namespaces, courses_wikis: cookbook_course_wiki, namespace: 0)
    create(:course_wiki_namespaces, courses_wikis: cookbook_course_wiki, namespace: cookbook_ns)
    JoinCourse.new(course:, user: user1, role: 0)
    JoinCourse.new(course:, user: user2, role: 0)
    VCR.use_cassette 'course_update' do
      UpdateCourseStats.new(course)
    end
    described_class.new(course, wikibooks, cookbook_ns)
  end

...
```

app/models/course_data/articles_courses.rb
```diff
@@ -132,7 +131,12 @@ def self.destroy_invalid_records(course, valid_article_ids)
...
 
-  def self.get_revisions_by_namespaces(namespaces, revisions)
-    revisions.joins(:article).where(articles: { namespace: namespaces })
+  def self.get_revisions_by_namespaces(course)
+    # Return revisions corresponding to tracked wikis and namespaces
+    course.tracked_namespaces.map do |wiki_ns|
+      wiki = wiki_ns[:wiki]
+      namespace = wiki_ns[:namespace]
+      course.revisions.joins(:article).where(articles: { wiki: wiki, namespace: namespace })
+    end.flatten
```
ragesoss, line 140: I think this is going to be a problem, because it's creating an array of all the revisions in a course. It's likely to run the system out of memory when it does this for a course with a very large number of revisions. We only actually need the `article_id` values, so it would be better to change this method to return an array of ids directly; a large array of integers takes up a trivial amount of memory compared to the same number of ActiveRecord objects.

vaidehi44: Oh...right, got it.

vaidehi44:
@ragesoss, I hope now it is correct. Haven't really tested it though.

app/models/course_data/articles_courses.rb
ragesoss: I'm not certain, but I don't think so. An ActiveRecord Relation object doesn't create an array of ActiveRecord objects, unless it gets converted to an array or iterated through. So in the current version, that Relation object gets `.distinct.pluck(:article_id)` called on it, which just returns the array of ids without individually creating an Article or Revision object for each record. Using Enumerable iterator methods `map` and/or `flatten`, on the other hand, do create an array of such objects.

app/models/course_data/articles_courses.rb
```diff
@@ -131,12 +131,12 @@ def self.destroy_invalid_records(course, valid_article_ids)
...
-  def self.get_revisions_by_namespaces(course)
-    # Return revisions corresponding to tracked wikis and namespaces
+  def self.article_ids_by_namespaces(course)
+    # Return article ids from revisions corresponding to tracked wikis and namespaces
     course.tracked_namespaces.map do |wiki_ns|
       wiki = wiki_ns[:wiki]
       namespace = wiki_ns[:namespace]
       course.revisions.joins(:article).where(articles: { wiki: wiki, namespace: namespace })
-    end.flatten
+    end.flatten.pluck(:article_id)
```
ragesoss, line 140: I think doing it this way will still load all the ActiveRecord objects into an array, I think (because of `.flatten`). I suggest building an array of ids more explicitly, by initializing an empty array and then adding ids to on each pass through the `tracked_namespaces` loop.

ragesoss:
I'll be ready to merge this after a) the further adjustment to ArticlesCourses I suggested in my last code comment, and b) the rubocop violations are fixed.

vaidehi44:
@ragesoss, it seems that there's only one failing test now. I tried to debug, but couldn't exactly figure out what's the issue here. Currently, I am extremely busy, so you might resolve it earlier. Other than this, I guess the piece of code which actually calls the UpdateWikiNamespaceStats service (in UpdateCourseStats) is still not pushed. I think, I have it in my stashed codes, so I can push that here or in another pr.

ragesoss:
That's a surprising test failure; I don't see how it would be related to your PR (but also not sure how it can be failing at all). Adding the code that calls the service would be great. I can take it from there, if you'd like.

ragesoss:
That failing test was because of a change I made the other day.

vaidehi44:
Okay...nice. I think, other than the other tests would be failing as the frontend is yet not linked to backend code, right?

vaidehi44:
Oh yeah, I just thought that that might be the case because the last commit didn't pass all the CI tests. But, I guess its the same test which is failing.

PR #5119, opened 2022-08-18
update I18n to version 4

TheTrio:
There doesn't seem a way to export JS files now. Instead, the exported files are JSON which must be fetched independently. Because of this, I18n is no longer an external script but is instead handled by webpack.

I've only tried this on the home page but I assume its similar elsewhere so should hopefully work there as well.


ragesoss:
Do you know what's causing the spec failures? Those aren't the tests I would have expected to fail for an i18n change.

PR #5127, opened 2022-08-24
fix user profile page

TheTrio:
I didn't notice that the user profile page depended on the Editable JS plugin. This plugin was already being loaded on the campaigns page.

Loading it on the profile page fixes the issue where in clicking the "Edit Details" button simply reloaded the page.


ragesoss:
@TheTrio can you rebase this? I've added a feature spec that would have caught the regression.

TheTrio:
Just did.

Issue #5094, opened 2022-08-03
`/training` should explain how to load training content if none is available


ragesoss:
### Current Behavior:

When a developer visits `/training` in their development environment, no training content is shown by default. Training content is loaded into the database — either quickly from .yml files in the `training_content` directory if it's in `wiki_education` mode, or slowly by pulling in wiki page content and translations from meta.wikimedia.org if it's not in `wiki_education` mode — when you trigger a training update by visiting `/reload_trainings?module=all` (or in the case of pulling from MetaWiki, `?module=<slug>` to import a single module).

### Desired Behavior:

When there are no training modules already loaded in the database, visiting `/training` should display some documentation explaining how to load training content (or maybe even a button link to the reload_trainings URL).

cyrillefr:
Hello @ragesoss , I would like to work on that one.

I would go for a link to the reload_trainings URL.
But no idea of the proper text to go along.

ragesoss:
There should be two variants, one for if the app is in `wiki_education` mode and one if not.

`wiki_education` mode:

> There are no TrainingLibrary records in the database. Click here to load training data from the `training_content/` yaml files.

non-wiki_education mode:

> There are no TrainingLibrary records in the database.

PR #5160, opened 2022-09-26
/training should display some documentation

cyrillefr:
- altered view to handle the no library case
- added localized text
- added tests for the 2 modes

When there is no library in the DB, some text is displayed stating there is no library yet.
In wiki_education mode, a link to /reload_trainings?module=all is also displayed.

Issue #5094

### DIFF
--- a/app/views/training/index.html.haml
+++ b/app/views/training/index.html.haml
@@ -3,27 +3,20 @@
   %h1 Training Libraries
   %ul.training-libraries.no-bullets.no-margin
     - defocus_class = @focused_library_slug ? 'training-library-defocus' : ''
-    - unless @libraries.empty?
-      - @libraries.each do |library|
-        - next if library.exclude_from_index?
-        - focus_class = @focused_library_slug == library.slug ? 'training-library-focus' : defocus_class
-        %li{class: "training-libraries__individual-library no-left-margin #{focus_class}"}
-          %a.action-card.action-card-index{href: "/training/#{library.slug}"}
-            %header.action-card-header
-              %h3.action-card-title
-                = library.translated_name
-              %span.icon-container
-                %i.action-card-icon.icon.icon-rt_arrow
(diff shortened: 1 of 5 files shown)


app/views/training/index.html.haml
```diff
@@ -3,20 +3,27 @@
...
-            %span.icon-container
-              %i.action-card-icon.icon.icon-rt_arrow
-        .action-card-text
-          %h3 Included Modules:
-          %ul
-            - library.categories.pluck('modules').flatten.each do |training_module|
-              %li
-                %a{href: "/training/#{library.slug}/#{training_module['slug']}", target: "_blank"}
-                  = training_module['name']
+    - unless @libraries.empty?
```
ragesoss, line 6: I think this is best done in a way that doesn't introduce such a large `if/else` branch in the template, because it is not trivial to skim this file and immediately see what the branch is for.

Instead, I suggest handling this within the controller, using `render(plain: ...) if ...`. That would contain the logic for this change to a single line of code.

ragesoss:
Thanks! Apologies for the two week wait before I could look at this; I just got back from vacation and I'm now catching up. :-)

app/views/training/index.html.haml
cyrillefr: Ok. As I needed to display a link in one case, I intended to make a method in the controller to be called like
render :html no_training
Problem is message contains a link, so a need to render html and not plain/text. Message display if use html_safe, but htat triggers a warning. I could have use sanitize, but I would have been forced to include ViewHelper ... which is not at its place in a controller.
So I made a new view instead, so that there is only one more line at the end of the index method.

cyrillefr:
No Problem, vacations are good :)

app/views/training/index.html.haml
ragesoss: These seems like a good solution to me.

Issue #4340, opened 2021-02-06
Error logging during course data updates should be extended


ragesoss:
### Current Behavior:
We have error logging that keeps track of the number of errors as well as Sentry IDs for problems that happen during part of the course data update process (which gets handled by passing the `update_service` object down to the API calls where data gets fetched, and then using the `ApiErrorHandling` module to log any errors.

This strategy hasn't been implemented for all of the steps of `UpdateCourseStats`, just the first few.

### Desired Behavior:

The error logging strategy should cover all the APIs called during the course stats update process.

### Related Files & Code:

* `app/services/update_course_stats.rb`

panktishah62:
@ragesoss  I am a new comer here, and after exploring the project a bit. I would like to work on it. Please can you add me to your slack channel where other technical conversations take place. Thank you !

rohitgeddam:
@panktishah62 are you still working on this? If u need some help, I would be happy to help you!

ragesoss:
> @ragesoss I am a new comer here, and after exploring the project a bit. I would like to work on it. Please can you add me to your slack channel where other technical conversations take place. Thank you !

@panktishah62 sorry, I missed this initially. I can send you an invite if you send me your email address. Either post it here or send to sage at wikiedu.org

aaditya9899:
hey @ragesoss I think I can work on this simultaneously with the previous one I mentioned so can I or should I wait for pull request then process.

ragesoss:
@aaditya9899 starting on this now is fine. Thanks.

sharon-kathambi:
Hi Sage, decided to start on this one, as i wait for your response on the other issue.

sharon-kathambi:
For the APIs, i can see the mediawiki,oreo and replica api, all have the loggers,Using Sentry and Rails.logger. Could you shed somelight on which steps on updateCourseStats that dont have the error handling

ragesoss:
(Adding it to just one step at a time is probably the best approach.)

sharon-kathambi:
Hi Sage, kindly review my PR to see if i am headed the right way
https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5174

wulanseruniati:
[...]

PR #5174, opened 2022-10-18
Fixes Error logging during course data updates should be extended issue #4340

sharon-kathambi:
## What this PR does
< It adds error logging to the updateCourseStats steps that do not have the error logging >

< The error logging uses Sentry to store the logs and Rails.logger >

### DIFF
--- a/app/services/update_course_stats.rb
+++ b/app/services/update_course_stats.rb
@@ -59,7 +59,7 @@ class UpdateCourseStats
   end
 
   def update_article_status
-    ArticleStatusManager.update_article_status_for_course(@course)
+    ArticleStatusManager.update_article_status_for_course(@course, update_service: self)
     log_update_progress :article_status_updated
   end
 
--- a/app/services/update_course_stats.rb
+++ b/app/services/update_course_stats.rb
@@ -54,7 +54,7 @@ class UpdateCourseStats
   end
 
   def update_categories
-    Category.refresh_categories_for(@course)
+    Category.refresh_categories_for(@course, update_service: self)
     log_update_progress :categories_updated
   end
 
@@ -64,7 +64,7 @@ class UpdateCourseStats
   end
 
   def update_average_pageviews
-    AverageViewsImporter.update_outdated_average_views(@course.articles)
+    AverageViewsImporter.update_outdated_average_views(@course.articles, update_service: self)
     log_update_progress :average_pageviews_updated
   end
 
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,15 +11,15 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.0].define(version: 2022_08_15_143028) do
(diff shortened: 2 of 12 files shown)


app/services/update_course_stats.rb
```diff
@@ -54,7 +54,7 @@ def fetch_data
   end
 
   def update_categories
-    Category.refresh_categories_for(@course)
+    Category.refresh_categories_for(@course, update_service: self)
```
ragesoss, line 57: Doing just this won't add logging. The idea is to pass the `UpdateCourseStats` instance to whatever code is being called to do work, and then update that code so that it uses the `UpdateCourseStats` instance whenever it sends error reports to Sentry — in particular, when there are errors during external network requests (API calls). See how it gets passed along to `CourseUploadImporter` then to `Commons` then to `WikiApi`, then to the `log_error` method from the `ApiErrorHandling`, for example.

To do the same thing here, you'd need to pass the service instance along to each of the classes that may make network requests (and thus, are likely to be a source of errors), starting with `CategoryImporter` (and then `WikiApi`, which is already instrumented to handle the update_service paramenter).

app/services/update_course_stats.rb
```diff
@@ -64,7 +64,7 @@ def update_article_status
   end
 
   def update_average_pageviews
-    AverageViewsImporter.update_outdated_average_views(@course.articles)
+    AverageViewsImporter.update_outdated_average_views(@course.articles, update_service: self)
```
ragesoss, line 67: I suggest starting with just one case (either this one, or the first branch of `Category#title_list_from_wiki` as I outlined above).

db/schema.rb
```diff
@@ -11,15 +11,15 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.0].define(version: 2022_08_15_143028) do
-  create_table "alerts", id: :integer, charset: "utf8mb4", force: :cascade do |t|
```
ragesoss, line 14: These changes to schema.rb should not be included in your PR.

ragesoss (changes requested):

app/services/update_course_stats.rb
sharon-kathambi: Hi Sage.
I have passed them on the ArticleStatusManager and averageViewImporter. Kindly review to see if I am on the right track

app/services/update_course_stats.rb
sharon-kathambi: Okay, noted

db/schema.rb
sharon-kathambi: okay

sharon-kathambi:
Hi Sage, My work got a bit jumbled up and bulky,  as i was doing all steps on one branch. I've decided to open different branches for each step...hence i'll have different PRs for each step.

sharon-kathambi:
https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5181

Issue #4940, opened 2022-04-18
Add a way to reach specific training modules and slides via ID numbers


ragesoss:
For the training system on Programs & Events Dashboard, we use wiki pages on Meta to specify training content that gets imported into the Dashboard's database, including unique IDs as the primary keys. However, there's no way to load a module index or an individual slide by ID, only by slug. Adding a similar redirect functionality to translate a URL with module/slide IDs into a working training URL would be useful.

For context, see https://meta.wikimedia.org/w/index.php?title=User_talk:Sage_(Wiki_Ed)&oldid=23169149#Wikidata_training_modules?

sharon-kathambi:
Hi @ragesoss  Can I work on this one?

ragesoss:
sure, please do.

sharon-kathambi:
Okay, i have a question, you would like the ID of the articles to be stored on the URL as well?

ragesoss:
I'm not sure what you mean.

sharon-kathambi:
In the training modules controller, we find modules by slug.. is the issue supposed to find training modules by ID? because when you click on a training module it redirects you to the specific module...and that's the functionality we want...or what exactly is the issue?

ragesoss:
The idea is, if you know the ID of a training module or slide — because you're editing the source on Meta, eg, slide 10815 https://meta.wikimedia.org/wiki/Training_modules/dashboard/slides/10815-using-a-talk-page — you should be able to use that to reach the slide on the Dashboard. This could be one similarly to the `find_course` route: `/find_training_slide/10815` could redirect to the URL of that slide ( https://outreachdashboard.wmflabs.org/training/editing-wikipedia/editing-basics/using-a-talk-page ). And similarly, `/find_training_module/:id` could do the same thing for finding the index URL for that TrainingModule.

gabina:
Hey @ragesoss, I am an outreachy applicant. I would like to apply for the Wikimedia Project "Improve how Wiki Education Dashboard counts references added".
I was reading the description and the discussion for this issue, and I think it's a good first step to start contributing. WDYT?
I already have my development environment working. Do you mind sending me a Slack invite?

ragesoss:
[...]

PR #5175, opened 2022-10-18
Add a way to reach specific training modules and slides via ID numbers issue#4940

sharon-kathambi:
## What this PR does
< It adds a way to reach specific training modules and slides via ID numbers 
issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/4940 
>

### DIFF
--- a/app/controllers/training_modules_controller.rb
+++ b/app/controllers/training_modules_controller.rb
@@ -14,14 +14,24 @@ class TrainingModulesController < ApplicationController
  
   #new code
   def find
-    set_training_library
+   # set_training_library
     @find_training_module ||= if is_number?(params[:module_id])
       TrainingModule.find_by(id: params[:module_id])
     else
       TrainingModule.find_by(slug: params[:module_id])
     end
-    redirect_to "/training/#{@library.slug}/#{@find_training_module.slug}"
+    find_corresponding_library
+    redirect_to "/training/#{@find_training_library.slug}/#{@find_training_module.slug}"
 
+  end
+
+  def find_corresponding_library
+    @find_training_library ||= if is_number?(params[:module_id])
+    #@find_training_library = TrainingLibrary.all.select(:categories).select(:modules).where("slug = ?", params[:module_id])
+      TrainingLibrary.all.where("categories.modules.id LIKE ?", params[:module_id])
+    else
+      TrainingLibrary.all.where("categories.modules.slug LIKE ?", params[:module_id]) 
+    end
   end
 
     private
--- a/app/controllers/training_modules_controller.rb
+++ b/app/controllers/training_modules_controller.rb
@@ -12,23 +12,22 @@ class TrainingModulesController < ApplicationController
     @training_module = TrainingModule.find_by(slug: params[:module_id])
   end
  
-  #new code
   def find_module
     @find_training_module ||= if is_number?(params[:module_id])
-      TrainingModule.find_by(id: params[:module_id])
-    else
-      TrainingModule.find_by(slug: params[:module_id])
-    end
+                                TrainingModule.find_by(id: params[:module_id])
+                              else
+                                TrainingModule.find_by(slug: params[:module_id])
+                              end
     find_corresponding_library
     redirect_to "/training/#{@library_slug}/#{@find_training_module.slug}"
   end
 
   def find_slide
-    @find_training_slide ||= if is_number?(params[:slide_id])
-      TrainingSlide.find_by(id: params[:slide_id])
-    else
-      TrainingSlide.find_by(slug: params[:slide_id])
-    end
+    @find_training_slide ||=  if is_number?(params[:slide_id])
+                                TrainingSlide.find_by(id: params[:slide_id])
+                              else
+                                TrainingSlide.find_by(slug: params[:slide_id])
+                              end
     find_corresponding_module
     find_corresponding_library
     redirect_to "/training/#{@library_slug}/#{@module_slug}/#{@find_training_slide.slug}"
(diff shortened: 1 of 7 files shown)


sharon-kathambi:
Hi Sage, how can i connect wiki_page to the dashboard code?
Screenshot from 2022-10-18 16-12-50
I tried editing the wiki_page by adding a 'to view the slide on dashboard, click here", but i am not sure how to connect the wiki page url to the code....

ragesoss:
@sharon-kathambi you don't need to make any edits on Meta-Wiki.

Once this feature is live on the Dashboard, we can edit one of the templates used on the Meta-Wiki source pages (like the page in your screenshot) so that such a link would appear on every page. However, it might be good to ensure that, on the Dashboard side of things, it can handle a parameter that includes both slug and ID. (For example, `/find_training_slide/10815-using-a-talk-page` would work in addition to `/find_training_slide/10815` or `/find_training_slide/using-a-talk-page`.)

app/controllers/training_modules_users_controller.rb
```diff
@@ -29,6 +29,14 @@ def mark_exercise_complete
     render 'courses/_block', locals: { block:, course: block.course }
   end
 
+  def find_slide
+    @dashboard_url = 'https://outreachdashboard.wmflabs.org/'
```
ragesoss, line 33: There's no need to make this an absolute link, and it shouldn't be hard-coded to `outreachdashboard.wmflabs.org` (as this feature will also be on `dashboard.wikiedu.org`).

app/controllers/training_modules_users_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 class TrainingModulesUsersController < ApplicationController
   respond_to :json
-  before_action :require_signed_in, only: [:create_or_update, :mark_exercise_complete]
+  before_action :require_signed_in, only: [:create_or_update, :mark_exercise_complete, :find_slide]
```
ragesoss, line 5: There's no need to call `require_signed_in` for this controller action. That would limit it to logged-in users, but we don't need to keep logged out users from using it as well.

app/controllers/training_modules_users_controller.rb
```diff
@@ -29,6 +29,14 @@ def mark_exercise_complete
     render 'courses/_block', locals: { block:, course: block.course }
   end
 
+  def find_slide
+    @dashboard_url = 'https://outreachdashboard.wmflabs.org/'
+    library_id = @training_library.id
+    module_id = @training_module.id
```
ragesoss, line 35: `@training_library` and `@training_module` will both be `nil` unless they get set first.

app/controllers/training_modules_users_controller.rb
```diff
@@ -29,6 +29,14 @@ def mark_exercise_complete
...
+    @slide = TrainingSlide.find(params[:id])
+    redirect_to '#{dashboard_url}/training/#{library_id}/#{module_id}/#{slide_id}'
```
ragesoss, line 37: This should be redirecting to the current URL for the slide, such as `/training/editing-wikipedia/editing-basics/using-a-talk-page`. To do that, we'll need to find the library slug, module slug, and slide slug.

ragesoss (changes requested):
This will also need one or more accompanying tests, probably by adding to `spec/features/training_tool_spec.rb`.

sharon-kathambi:
Decided to change my approach, what if i toggle between the id and slug?

app/controllers/training_modules_users_controller.rb
sharon-kathambi: Okay, i have made the change

ragesoss:
> Decided to change my approach, what if i toggle between the id and slug?

Can you explain what you mean in more detail?

ragesoss:
I think the best way to make sure this is working as intended will be to write a `feature` spec for it, so that we have a test that exercises the new route.

sharon-kathambi:
Hi Sage, sorry i have been MIA..i have been unwell. I am back now.
I wrote a find_training_module/:id that directs the module to the '/training/students/#{training_module.slug}' path...I have written some regex as well..that handles a parameter that takes both id and slug for the training_module

i think this approach works...let me know

app/controllers/training_modules_controller.rb
```diff
@@ -11,4 +11,20 @@ def index
...
-end
+ 
+  #new code
+  def find
+    @find_training_module ||= if is_number?(params[:module_id])
+      TrainingModule.find_by(id: params[:module_id])
+    else
+      TrainingModule.find_by(slug: params[:module_id])
+    end
+    redirect_to "/training/students/#{@find_training_module.slug}"
```
ragesoss, line 22: `students` here is the path for a specific `TrainingLibrary` with slug `students`. Although it will still render no matter which  value is given for the library slug, it won't accurately reflect the context for the given module.

I think it's important to find the corresponding library as well and reflect that in the URL.

The same concept is even more important for `TrainingSlide`, because a slide URL won't actually render unless it includes the corresponding `TrainingModule` slug as well. (This is because the last layer of routing for training slides is handled on the frontend, and the slide content for the entire module gets loaded in order to display a single slide, so that navigating between slides is instantaneous because the whole module is already loaded.)

sharon-kathambi: Do you have to be logged in as a user for the trainingSlide redirect feature to work?

ragesoss: No, it shouldn't require being logged in.

sharon-kathambi: I have added the corresponding library.slug as well as module slug for the find_training_slide
it works on my end...
would something like this work?

ragesoss: I don't think so. It needs to be able to find the right redirect using only one id parameter, so to find the library that a module is in, you will likely need to load all the libraries and find the first one that includes the module's slug as a value within its `categories` data. You can see what this data structure looks like in the files in `training_content/wiki_ed/libraries`.

I think one way to go about this is to add an instead method to the `TrainingModule` file to find and return the corresponding library. (And the same approach should work for `TrainingSlide` to find the corresponding module.)

sharon-kathambi: def find_corresponding_library
    @find_training_library ||= if is_number?(params[:module_id])
 TrainingLibrary.all.where("categories.modules.id LIKE ?", params[:module_id])
    else
      TrainingLibrary.all.where("categories.modules.slug LIKE ?", params[:module_id]) 
    end
  end

trying to fetch module slug from the trainingLibrary...but i keep getting slug not found
This error...
[2022-10-28 11:34:39.655 FATAL] NoMethodError (undefined method `slug' for #<TrainingLibrary::ActiveRecord_Relation:0x00007fe39d9dd590>):

Is modules an array inside category?and is category an array column?

sharon-kathambi: This should ideally work?

ragesoss: Both of those ActiveRecord queries use `where`, so they are returning a Relation (which works similarly to an Array), but you want to find an individual record. I think you can use `find_by` instead to have it return a single record.

sharon-kathambi: Okay. I actually worked on it a few days ago, can you review it kindly?

spec/features/training_tool_spec.rb
```diff
@@ -70,6 +70,7 @@
       login_as(user, scope: :user)
     end
 
+
```
ragesoss, line 73: Looks like you removed the spec you had added?

This definitely needs a spec or two to exercise the feature.

sharon-kathambi: Was getting some errors with the tests, i am rewriting them.

Issue #4978, opened 2022-05-16
Add a way for users to trigger a username update for a renamed user


ragesoss:
### Current Behavior:

If a user gets their account renamed on Wikipedia, the Dashboard does not automatically detect this. That means, after the name change, the Dashboard continues to look for new revisions by the old username.

The username gets updated on the Dashboard if the user logs out (of the Dashboard) and then back in again, because the OAuth login flow for an unknown username includes a step checking whether the same global user ID is already associated with a different (previous) username. See `UserImporter.new_from_username`, and specifically, `.update_username_for_global_id`.

### Desired Behavior:

When an instructor knows that their student has changed usernames, they (or Wiki Education staff) should be able to easily trigger a username update, without needing to pass along instructions to the student to log out and back in again.

I think this could be done via a simple Rails form that allows a user to enter the new username of a user who changed usernames; upon submission, it would do the same thing as the login process, calling `UserImporter.update_username_for_global_id` for that username. The new username would then automatically be reflected on the student's course page, and the Dashboard would resume pulling in new edits for that user during the regular update cycle (along with any edits that were missed since the name change)

Once such a form is live, we can create an FAQ entry to point instructors to it.

iLynette:
Hi @ragesoss I would like to work on this issue. Can I?

ragesoss:
Yes. (Please stick to one issue at a time, though.)

iLynette:
Got it, thank you.)

tnwoodard:
Hi @ragesoss and @iLynette. I'd like to try to fix this issue if it is still open. Thanks!

ragesoss:
@iLynette thanks. please post on slack if you'd like any help with the setup process.

tnwoodard:
Ok I understand. That sounds good. @ragesoss and @iLynette I will check back in a day or so, and if it is still open I'll ask again. Thanks.

PR #5179, opened 2022-10-19
Attempted to address Issue #4978

tnwoodard:
I installed simple forms to generate a form in the FAQ section for updating student roster.

## What this PR does
The purpose of this PR is to address Issue #4978  

## Screenshots
Before:
No UI available. This is a new enhancement.

After:

FAQs with link to update roster
Form to update roster with update user clicked

I have some questions and concerns listed below:
1. Where is the information going?  Once the "Update User" button is clicked, the new username will be added where?
2. Is there more to be done with `UserImporter.update_username_for_global_id`?
3. Currently there is a link to update the user on the FAQ page ('/faq'). It would be nice to add it to Activity section or somewhere else that is easy to find.
4. Eventually, the form will need some style.

Thank you,
Tempestt


config/initializers/simple_form_bootstrap.rb
```diff
@@ -0,0 +1,440 @@
+# frozen_string_literal: true
```
ragesoss, line 1: I don't think adding this is necessary. To make the feature as simple and maintainable as possible, it should use existing UI patterns and follow the conventions of the rest of the codebase; I don't want to add a large amount of initializer code like this that I don't fully understand.

config/locales/en.yml
```diff
@@ -59,7 +59,7 @@ en:
     log_out: Log out
     my_dashboard: My Dashboard
     recent_activity: Recent Activity
-    report_problem: Report a problem
+    report_problem:
```
ragesoss, line 62: Looks like this was inadvertant?

app/views/faq/index.html.haml
```diff
@@ -24,3 +24,5 @@
         All questions
       - else
         %a.topic-link{href: "/faq?all=true"} All questions
+        %br
+        %a.topic-link{href: "/update_roster"} Update student roster
```
ragesoss, line 28: I don't think we need a hard-coded link to this from the FAQ index; linking to it from an individual FAQ is something we can do from the frontend (by creating or editing a relevant FAQ entry) after the form is deployed.

app/controllers/roster_controller.rb
```diff
@@ -0,0 +1,13 @@
+class RosterController < ApplicationController
```
ragesoss, line 1: I don't think "roster" is the right name to use here. This controller's job will be to trigger a username update, which is independent of which courses that user is participating in. (We also don't use the term "roster" anywhere else in the system, so it's likely to be confusing.)

app/controllers/roster_controller.rb
```diff
@@ -0,0 +1,13 @@
...
+  before_action :require_signed_in
+  respond_to :html, :json
+  def update_roster
```
ragesoss, line 4: Where is the code that actually updates the `User` record?

ragesoss:
This should include a Feature spec to verify that it works as intended, updating a `User` record with that user's current username if the Dashboard currently has an outdated username for that account.

tnwoodard:
Thanks for the feedback, Sage. I will work on this to address your
feedback.


-Tempestt

config/initializers/simple_form_bootstrap.rb
tnwoodard: Ok, I will remove Bootstrap UI.

app/views/faq/index.html.haml
tnwoodard: Oh ok. Do you have a preference of where the form should be located?

config/locales/en.yml
tnwoodard: Yes. I'm unsure how it happened, but I added it back.

app/controllers/roster_controller.rb
tnwoodard: It will go in the enrollment_controller.rb file.

app/views/faq/index.html.haml
ragesoss: Somewhere similar, perhaps `/update_username`, would make sense.

tnwoodard:
Hi Sage,

I tried to address the comments with `update_username` and creating a Feature spec. I learned about test specs in my bootcamp program, but this is my first test spec. I really appreciate your feedback. Thank you.

app/views/faq/index.html.haml
```diff
@@ -19,8 +19,3 @@
         = topic.name
       - else
         %a.topic-link{href: "/faq?topic=#{topic.slug}"}= topic.name
-    %hr
```
ragesoss, line 22: It looks like you've removed this by accident.

PR #5181, opened 2022-10-21
Passing down update service to Update article status path issue #4340

sharon-kathambi:
## What this PR does
< This PR breakdowns the issue #4340 
Passing down update service to one of the updateCourseStats steps - updateArticlestatus >

### DIFF
--- a/My Survey-results2022-10-21.csv
+++ b/My Survey-results2022-10-21.csv
@@ -1,0 +1,1 @@
+username,course,Q7: Long Text Question
--- a/Question#7-results2022-10-21.csv
+++ b/Question#7-results2022-10-21.csv
@@ -1,0 +1,2 @@
+Question,Grouped Matrix Question,Answer,Follow Up Question,Follow Up Answer,Username,User Email
+Long Text Question,"",hello world,,yes,Ragesock,instructor@school.edu
--- a/app/services/update_course_stats.rb
+++ b/app/services/update_course_stats.rb
@@ -17,8 +17,8 @@ class UpdateCourseStats
 
   def initialize(course, full: false)
     @course = course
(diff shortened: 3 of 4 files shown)


My Survey-results2022-10-21.csv
```diff
@@ -0,0 +1 @@
+username,course,Q7: Long Text Question
```
ragesoss, line 1: This is a file generated during a test run that didn't get deleted. (This can happen sometimes when a test is stopped before the cleanup routine that should delete this file gets run.)

Question#7-results2022-10-21.csv
```diff
@@ -0,0 +1,2 @@
+Question,Grouped Matrix Question,Answer,Follow Up Question,Follow Up Answer,Username,User Email
+Long Text Question,"",hello world,,yes,Ragesock,instructor@school.edu
```
ragesoss, line 2: This is also a test-generated file that should not be included.

app/services/update_course_stats.rb
```diff
@@ -59,7 +59,7 @@ def update_categories
   end
 
   def update_article_status
-    ArticleStatusManager.update_article_status_for_course(@course)
+    ArticleStatusManager.update_article_status_for_course(@course, update_service: self)
```
ragesoss, line 62: Note this this isn't `ArticleStatusManager.new` (which corresponds to the `initialize` class method that you changed), it's a class method which also needs to be modified to accept an additional parameter. If you run the tests for the files you're working on locally, you can get quick feedback about problems like this.

sylvia-boyani:
When running any commands Sharon you should add the '--no test framework' to avoid tests generated by default when the commands runs. And maybe write your own tests later if it is easier for you.

Issue #5007, opened 2022-06-15
Training resources should have a search feature


ragesoss:
### Current Behavior:
There are lots of training modules about specific topics and specific task, but it's no always obvious where to find the training content about the topic you are looking for. Browsing the modules to find exactly what you are looking for is tedious.

### Desired Behavior:
There should be a search bar that can help you find training content related to arbitrary keywords or phrases, searching the content, titles, and descriptions of training slides, modules and libraries.

marklocklear:
How do we set up /training in our local environment? When I got to http://localhost:3000/training I get a blank page.

ragesoss:
In `wiki_education` mode, you can load the full set of training modules from .yml files by visiting `/reload_trainings?module=all`.

It would be a nice improvement to add a message or link on how to do this that would display when there are no training modules in the database.

ragesoss:
I opened a separate issue for the latter: https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5094

cyrillefr:
Hello @ragesoss , can I work on this one ?

ragesoss:
Sure!

cyrillefr:
Is there a pattern from an existing feature you want me to copy/get inspiration of ?
(/explore : there is an input box to look for courses, there is also a FAQ and I surely misses other ones).

Also, I have spotted a bug in `/reload_trainings?module=all` and possibly another one when modules are displayed.

cyrillefr:
Hello @ragesoss,
regarding the courses search module, front-end is in React. The current code for listing training modules is in html/haml. I can mimick the back-end code of courses search, but for the front-end part, I guess it is better to stick with back-end html.

Regarding  the bug(?) I found out: on my local machine `/reload_trainings?module=all` causes a 500 error due to a `ActiveRecord::RecordNotUnique` at line `app/models/training_slide.rb:62`
It happens in the method `self.inflate` at the `slide.save` fragment. 

But the real problem is the first line `slide = TrainingSlide.find_or_initialize_by(id: all_content['id'])`
The population of the TrainingSlide object is made via the `training_content/wiki_ed` (slides and modules).
Id and slug are based on file names. 
Populating the db means ingesting those files and creating the matching records in DB with id & slug red.
But ...
As it appears there is on my dev DB some TrainingSlide objects with -another- id than the ones red from the files.
So, another object is created with another id aaand the same slug.
There is a unique index on the slug.
[...]

PR #5190, opened 2022-10-28
Training resources should have a search feature

cyrillefr:
- added a lib to handle search
 - extracted previous query from controller
 - localization
 - front: search bar
 - corresponding tests(lib + features)

## What this PR does
At `/training`, a simple search bar is added, it searches in the TrainingSlide objects content for a keyword.
If found, it then displays the matching modules, the same clickable links  as usual(no search).

Issue #5007

### DIFF
--- a/app/views/training/index.html.haml
+++ b/app/views/training/index.html.haml
@@ -4,7 +4,7 @@
   .search-bar{style: 'position:relative'}
     = form_tag("/training", method: "get") do
       = text_field_tag(:search_training, '', placeholder: t('training.search_training_resources'), style: 'width: 100%; height: 3rem;font-size: 15px')
-      %button{type: 'submit', style: 'position: absolute; right: 20px; top: 10px'}
+      %button{type: 'submit', id: 'training_search_button', style: 'position: absolute; right: 20px; top: 10px'}
         %i.icon.icon-search
   - if @search
     = render 'search_results'
--- a/app/views/training/index.html.haml
+++ b/app/views/training/index.html.haml
@@ -1,6 +1,11 @@
 - content_for :before_title, "Training Libraries - "
 .container.training
   %h1 Training Libraries
+  .search-bar{style: 'position:relative'}
+    = form_tag("/training", method: "get") do
+      = text_field_tag(:search_training, '', placeholder: t('training.search_training_resources'), style: 'width: 100%; height: 3rem;font-size: 15px')
+      %button{type: 'submit', name: 'glass_training', style: 'position: absolute; right: 20px'}
+        %i.icon.icon-search
   %ul.training-libraries.no-bullets.no-margin
     - defocus_class = @focused_library_slug ? 'training-library-defocus' : ''
     - @libraries.each do |library|
--- a/app/views/training/index.html.haml
+++ b/app/views/training/index.html.haml
@@ -4,7 +4,7 @@
   .search-bar{style: 'position:relative'}
     = form_tag("/training", method: "get") do
       = text_field_tag(:search_training, '', placeholder: t('training.search_training_resources'), style: 'width: 100%; height: 3rem;font-size: 15px')
-      %button{type: 'submit', name: 'glass_training', style: 'position: absolute; right: 20px; top: 10px'}
+      %button{type: 'submit', style: 'position: absolute; right: 20px; top: 10px'}
         %i.icon.icon-search
   - if @search
     = render 'search_results'
--- a/app/views/training/index.html.haml
+++ b/app/views/training/index.html.haml
@@ -4,24 +4,9 @@
   .search-bar{style: 'position:relative'}
     = form_tag("/training", method: "get") do
       = text_field_tag(:search_training, '', placeholder: t('training.search_training_resources'), style: 'width: 100%; height: 3rem;font-size: 15px')
-      %button{type: 'submit', name: 'glass_training', style: 'position: absolute; right: 20px'}
+      %button{type: 'submit', name: 'glass_training', style: 'position: absolute; right: 20px; top: 10px'}
         %i.icon.icon-search
-  %ul.training-libraries.no-bullets.no-margin
-    - defocus_class = @focused_library_slug ? 'training-library-defocus' : ''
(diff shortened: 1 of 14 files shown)


ragesoss:
Thanks! This is a great start. After testing it out, I have some suggestions for a better user experience.

* I think, instead of modules being the unit of content returned by the search, it should be slides. The most common use case is that a user is looking for the specific slide that covers topic X. Ideally, this would include both a link to the slide title (and maybe a link to the module it's part of) and also an excerpt of the slide content that includes the search term.
* The search bar styling is off. This is probably because the /training pages use a different stylesheet, and some of the `.search-bar` and `input` styles from the main stylesheet aren't part of this one. Compare the search bar on /explore, which is well-styled.
*

cyrillefr:
Hello @ragesoss , I think there might be an issue with the first bullet point.
If I somewhat provide/display a direct link to a training slide, it will update the training all the way to this slide.
That is: 
let's say on the module Wikipedia policies, I have watched till the second slide(2 - Five pillars).
If I then visit "13 - Notability", then all slides from 2 to 13 will be displayed as seen, and from 5% of the training, there will be 59% done. Without ever have studied slides 3-12.
So, I do not know if we should display direct links to training slides.(or may be you are talking of another link).
May be adding under the link to the module: an excerpt of the slide content with the search word highlighted + name of the slide

ragesoss:
That is a very good point. I think I'm okay with letting users skip to the end of a module via search results, though.

Perhaps a good solution (not necessarily as part of this PR) would be to disable progress/completion tracking whenever a module is accessed via search results, or started from any slide later than immediately after the user's previous progress.

cyrillefr:
Hello @ragesoss, I have made some changes.
Tell me if you like the new listing, I hope it is similar with the original one. I can make some other changes.(I realise there is no message if no match).
Also, regarding the CSS, I have found a padding of 10 for the container class caused the glass magnifier to sometimes not be horizontally aligned with the search box. I have found that there were some extra styling for the training(comparing with /explore), so I just made a tiny modification with this padding.

lib/training/training_resource_query_object.rb
```diff
@@ -0,0 +1,73 @@
...
+    slide_paths_with_excerpt(slugs)
+  end
+
+  private
+
+  def add_libr_and_mod_slugs(slugs)
+    slugs.each do |slug|
+      slug[:module], slug[:module_name] =
+        module_of_a_slide(slug[:slide]).then { |tm| [tm.slug, tm.name] }
+      slug[:library] = library_of_a_module(slug[:module]).slug
```
ragesoss, line 36: I'm getting errors in some cases from this, when no library is found.

lib/utils/string_utils.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class StringUtils
+  class << self
+    def excerpt(text, kword, length)
+      txt = text.tr "\n", ''
+      idx = txt.downcase.index(/#{kword}/i)
+      pad = (length / 2) - 3
+      tmp = "...#{txt[[idx - pad, 0].max..idx + pad - 1]}..."
```
ragesoss, line 9: This is another place where some searches throw an error, when `idx` is nil.

This whole method is also hard to understand on a quick read. What does `idx` mean here?

app/views/training/index.html.haml
```diff
@@ -1,22 +1,12 @@
...
-          %h3 Included Modules:
-          %ul
-            - library.categories.pluck('modules').flatten.each do |training_module|
-              %li
-                %a{href: "/training/#{library.slug}/#{training_module['slug']}", target: "_blank"}
-                  = training_module['name']
+  .search-bar{style: 'position:relative'}
+    = form_tag("/training", method: "get") do
+      = text_field_tag(:search_training, '', placeholder: t('training.search_training_resources'), style: 'width: 100%; height: 3rem;font-size: 15px')
+      %button{type: 'submit', name: 'glass_training', style: 'position: absolute; right: 20px; top: 10px'}
```
ragesoss, line 7: Why is this called 'glass_training'?

ragesoss (commented):
It's possible to have training modules that are not part of a Library, and these don't behave well. (These are, for example, older modules that we no longer actively use and don't want people to discover while browsing, but we want links from old courses to still work.)

Training paths still work even if there is no library, though, so a placeholder string like 'none' could be used.  For example, this link works: http://localhost:3000/training/none/sandboxes

The current behavior is instead to link to something like `http://localhost:3000/training//sandboxes` (which does not work, because of the double slashes).

ragesoss:
Design-wise, I like it!

lib/training/training_resource_query_object.rb
cyrillefr: Aaaah, I was not aware of that. I thought each module belonged to a library. Thank you for the clarification.
Just to be sure: do I have to exclude those slides(belonging to a module without library) from the results ?

ragesoss: Hmm... I think including them in the results would be fine (the links just need to be munged to avoid broken paths).

app/views/training/index.html.haml
ragesoss: Okay. I think `query` or `search` would be a better name, so that it's obvious the resulting URL what that parameter is for.

lib/training/training_resource_query_object.rb
cyrillefr: I have added some "sanitazation" (?). I realised that the search was made on text inside tags like img and href and afterwards were displayed (eg search on 'wikipedia' displayed results on url containing wikipedia ). Also I strip markdown link too.

lib/utils/string_utils.rb
cyrillefr: I rewrote a bit StringUtils, added some checks + tests.

app/views/training/index.html.haml
cyrillefr: We did not really need a name here after all :). An id is better as it do not appears in the ur.l

ragesoss:
Looks good! Thank you!

Issue #5165, opened 2022-10-10
Lists of students, articles, assignments can get out of date when tab is left open


ragesoss:
### What is happening?

If a user keeps a Dashboard window open (for example, perpetually throughout a course) without reloading, then the lists of users, articles and assignments are not refreshed from the server. This can lead to confusion wherein a student joins but is not shown to the instructor because they've only been navigating from tab-to-tab of the course page without a full page reload.

### Expected behavior

When a user navigates away from one of these tabs and then back to it, the key data associated with it should be refreshed from the server. However, we should make sure that we aren't double-fetching the same data when a page is initially loaded.

iLynette:
Hi, @ragesoss I would like to work on this issue. Can I?

ragesoss:
This one is probably too complex for a good first issue.

iLynette:
Alright.)

TheTrio:
One quick and simple way to solve this would be to refresh the data every few minutes - say something like this in `app/assets/javascripts/components/students/containers/StudentsTabHandler.jsx`

```jsx
...
componentDidMount() {
  this.interval = setInterval(() => {
    this.props.fetchUsers(this.props.course.slug);
  }, 1000 * 60);
},

componentWillUnmount() {
  clearInterval(this.interval);
},
...
```

Another way might be to refresh the data only when the user switches tabs. Something like this

```
let firstMount = true;
const StudentsTabHandler = createReactClass({
  displayName: 'StudentsTabHandler',
  ...
  componentDidMount() {
    if (!firstMount) {
      this.props.fetchUsers(this.props.course.slug);
    } 
      firstMount = false;
      // don't fetch since this is the first mount
    }
  },
```

This however doesn't work if the user is on the same tab for a long time - so I guess a combination of both of these approaches would provide the best experience. What do you think?

PR #5228, opened 2022-12-26
refresh stale data on tab switch

TheTrio:
Addresses #5165 

This works by re-fetching data when the user navigates from tab to tab. However, to prevent too many requests, the previously fetched data must be at least a minute old.

To test this out, go to the course page. In another window, make some changes to the course - like adding a new user. In the old window, navigate between the tabs and after a minute, when you go back to the student tab, the data will be re-fetched. 

There are some improvements but I'm not sure if they are worth pursuing - 

1. Data could be re-fetched automatically after some point of time when on the same tab. With the current implementation, if you're on the Student's tab and a new student is added, you won't notice it until you move away or come back(or directly click on the student tab - which refreshes it the same way) Like I wrote in the issue, it would be easy enough to have a `setTimeout` run every now and then and refresh the data. 
2. Have some sort of "notification" so that the user knows that the data has updated. 

Do let me know what you think of these ideas.

### DIFF
--- a/app/assets/javascripts/components/util/routes.jsx
+++ b/app/assets/javascripts/components/util/routes.jsx
@@ -33,6 +33,7 @@ const routes = () => {
     lastRequestAssignmentTimestamp: useSelector(state => state.assignments.lastRequestTimestamp)
   };
 
+  // this is called whenever the route changes
   useEffect(() => {
     refreshData(location, refreshArgs, dispatch);
   }, [location.pathname]);
--- a/app/assets/javascripts/components/util/routes.jsx
+++ b/app/assets/javascripts/components/util/routes.jsx
@@ -33,7 +33,17 @@ const routes = () => {
     lastRequestAssignmentTimestamp: useSelector(state => state.assignments.lastRequestTimestamp)
   };
 
+
   // this is called whenever the route changes
+  // it ensures that the user isn't displayed stale/outdated data for a route
+  // they previously visited(and thus had already loaded the data for in the Redux store)
+  // this fires the refreshData function, which does a series of URL checks and checks to see if
+  // the data is stale, and if so, refresh it by triggering the appropriate Redux action
+  // see refresh.js for more details on how this works
+
+  // the reason this exists here rather than in the individual components is to make the
+  // entire thing more centralized and easier to maintain. Besides, the fact that the
+  // this component deals with just the routes means that unnecessary re-renders are avoided
   useEffect(() => {
     refreshData(location, refreshArgs, dispatch);
   }, [location.pathname]);
--- a/app/assets/javascripts/components/util/routes.jsx
+++ b/app/assets/javascripts/components/util/routes.jsx
@@ -28,7 +28,7 @@ const routes = () => {
   const refreshArgs = {
     lastUserRequestTimestamp: useSelector(state => state.users.lastRequestTimestamp),
     courseSlug: useSelector(state => state.course.slug),
-    limit: useSelector(state => state.articles.limit),
+    articlesLimit: useSelector(state => state.articles.limit),
     lastRequestArticleTimestamp: useSelector(state => state.articles.lastRequestTimestamp),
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/components/util/routes.jsx
```diff
@@ -20,32 +22,48 @@ const TrainingApp = lazy(() => import('../../training/components/training_app.js
...
-      <Route path="/explore" element={<Explore dashboardTitle={window.dashboardTitle}/>} />
-      <Route path="/active_courses" element={<ActiveCoursesHandler dashboardTitle={window.dashboardTitle}/>}/>
-      <Route path="/courses_by_wiki/:wiki_url" element={<CoursesByWikiHandler />}/>
+const routes = () => {
+  const location = useLocation();
+  const dispatch = useDispatch();
+  const refreshArgs = {
+    lastUserRequestTimestamp: useSelector(state => state.users.lastRequestTimestamp),
+    courseSlug: useSelector(state => state.course.slug),
+    limit: useSelector(state => state.articles.limit),
```
ragesoss, line 31: I know `articles` is the only thing we use limit for, but I think there's still potential for confusion later on, with this called `limit` instead of `articleLimit`. Now that this bundle of args is shared across several different types of requests, a more explicit name would help.

yarn.lock
```diff
@@ -8670,7 +8670,7 @@ __metadata:
   resolution: "list.js@https://github.com/WikiEducationFoundation/list.js.git#commit=3be71e3414d60e8a40b41a0ca8d582f2f40f9308"
   dependencies:
     string-natural-compare: ^2.0.2
-  checksum: b853ac7141ebd1bf530ca6104950c0ec0f43b1b4d791db8488ba4523011c72301f62e344e1b8aaf060f69fbd93207826ab4ece0a40c99f6c1c929af417e2f9ec
+  checksum: ea0e976ceae4ddc2b7c78ec41214ceabc4e65368261cc635324b1ef0b7fb67cfca0de6f9af0eb6399da08f31480a5804a2927496ea3b238544b6317f2ad0f687
```
ragesoss, line 8673: What is with these changes?

ragesoss (commented):
I haven't tested it yet, but I think this looks good to me architecture-wise.

yarn.lock
TheTrio: Huh didn't notice that. I ran `yarn` which afaik shouldn't change the lockfile. 

Will revert these changes.

app/assets/javascripts/components/util/routes.jsx
```diff
@@ -20,32 +22,48 @@ const TrainingApp = lazy(() => import('../../training/components/training_app.js
...
+    lastRequestArticleTimestamp: useSelector(state => state.articles.lastRequestTimestamp),
+    lastRequestAssignmentTimestamp: useSelector(state => state.assignments.lastRequestTimestamp)
+  };
 
-      {/* this prevents the "route not found" warning for pages which are server rendered */}
-      <Route path="*" element={<div style={{ display: 'none' }}/>} />
-    </Routes>
-  </Suspense>
-);
+  useEffect(() => {
```
ragesoss, line 36: Whoops, I think the comment I meant to leave here got lost...

I think this should have some comments explaining what this useEffect is for and when the function itself will fire (separately from when `refreshData` will do something, which is based on timestamps).

TheTrio: It'll fire when there is a change to the route - I've added a comment saying that. I'm not sure there's more to it than that :sweat_smile:

ragesoss: Sorry, I should have been more clear. I think the comment should explain the strategy here: firing the effect whenever the route changes, but using logic within refreshData to determine whether to actually make the request... since this is a strategy we'll probably want to extend to other similar requests as well, by adding more to the refreshData logic.

This file is a pretty high level one in the component hierarchy, so being a little more explicit about what job it's doing will be particularly useful for newcomers to the codebase.

ragesoss:
Works just as expected. :-)

PR #5232, opened 2023-01-05
Issue #5195 Added global_id column next to username column to CSV export of editor data

kunalthedev:
Issue #5195
Updated CSV Export of data , added a global_id column next to username column in CSV export
Old CSV:-
image

New CSV:-
image

### DIFF
--- a/spec/lib/analytics/course_students_csv_builder_spec.rb
+++ b/spec/lib/analytics/course_students_csv_builder_spec.rb
@@ -42,12 +42,12 @@ describe CourseStudentsCsvBuilder do
       columns = line.split(',')
       username = columns[0]
       user_result = expected_result[username]
-      # column 9 is 'registered_during_project', which should be true
-      expect(columns[8]).to eq('true')
-      # column 10 is 'total_articles_created'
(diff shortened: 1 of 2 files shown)


spec/lib/analytics/course_students_csv_builder_spec.rb
```diff
@@ -42,12 +42,12 @@
...
       username = columns[0]
       user_result = expected_result[username]
-      # column 9 is 'registered_during_project', which should be true
-      expect(columns[8]).to eq('true')
-      # column 10 is 'total_articles_created'
-      expect(columns[9]).to eq(user_result[:articles_created])
-      # column 11 is 'total_articles_edited'
-      expect(columns[10]).to eq(user_result[:articles_updated])
+      # column 10 is 'registered_during_project', which should be true
+      expect(columns[9]).to eq('true')
```
ragesoss, line 46: Ideally, this test would be fixed in a way that won't require renumbering these indices if another column gets added. I'm okay merging it as-is, since this problem was already present, but making the test less brittle would be a nice next step if you'd like to try it. Let me know either way.

Issue #4859, opened 2022-03-15
Create an Alert when an active course Timeline gets removed


ragesoss:
### Current Behavior:
On dashboard.wikiedu.org, a course must have a Timeline before it gets submitted, and Wiki Education staff will only approve courses with detailed timelines that include the standard set of training modules (among other things that we check for).

However, after course approval, an instructor can still modify their Timeline. In rare cases, they may remove the standard Timeline content entirely. We don't have a way of detecting cases like this automatically, but it's important to know about them quickly so that we can contact the instructor.

### Desired Behavior:
When an instructor blanks the Timeline or updates to include no training modules, the Dashboard should generate an Alert and email the staff (ie, `Alert#email_course_admins`).

This should only apply for `wiki_education` mode, and only for the ClassroomProgramCourse type.

shekinahmanyi:
Hi @ragesoss please i will like to work on this issue

ragesoss:
@shekinahmanyi hi! are you working on this?

shekinahmanyi:
> @shekinahmanyi hi! are you working on this?

Yes @ragesoss

ragesoss:
Great! Let me know if you have any questions or get stuck with anything.

shekinahmanyi:
> Great! Let me know if you have any questions or get stuck with anything.

Ok thanks!
I'll do so

marklocklear:
@ragesoss How can we tell if a course is approved? Is the **approved?** method in **course.rb** sufficient?

ragesoss:
@marklocklear yes, for an individual Course, `#approved?` is the way to tell. (We don't have scope for getting the collection of all approved courses, but the same logic could be turned into a scope if needed.)

marklocklear:
@ragesoss here is a first stab at this https://github.com/marklocklear/WikiEduDashboard/tree/timeline_removal_alert. Let me know if I'm headed down the right path. Also probably need more logic in the **deleted_timeline?** method in deleted_timeline_alert_manager.rb.

Also unclear about what email template gets called from **alert_mailer.rb**; like what is **@alert.message**?

ragesoss:
Thanks @marklocklear. I think following the pattern of the other alert managers is probably not the right way to do this, because the situation we're trying to detect may look like this:

* Course is approved with a normal timeline
* Later, instructors removes all content from the timeline (either using the 'delete timeline' button or by deleting each week or block manually)
[...]

PR #5233, opened 2023-01-09
Trying issue #4859 Sending alert notification when admin tries to del…

kunalthedev:
Workflow goes like

Whenever any active timeline gets removed, `CheckTimeline` helper function gets invoked which checks the following factors:-
If the coourse is approved but doesn't have any timeline then it creates and alert.
Mailer Preview:-
image
Admin Alert List:-
image
Alert Description:-
image


app/models/alert_types/deleted_timeline_alert.rb
```diff
@@ -0,0 +1,42 @@
...
+    course.url
+  end
+
+  def resolvable?
+    !resolved
+  end
+
+  def resolve_explanation
+    <<~EXPLANATION
+      It appears you have removed either the timeline, week or block from your course.
```
ragesoss, line 39: This alert should be going to Wiki Education staff, rather than the user who did the removal.

app/models/alert_types/deleted_timeline_alert.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def url
+    course.url
+  end
+
+  def resolvable?
+    !resolved
+  end
+
+  def resolve_explanation
```
ragesoss, line 37: This `resolve_explanation` is not intended as the main content of the alert, and doesn't need to duplicate the `message` that gets set when the Alert is created.

ragesoss:
I see some problems with the code, but I think my first suggestion would be to work on 1) finding a way to manually test the feature locally (and fix any problems you find), and 2) writing a feature spec that does the same thing.

lib/alerts/course_alert_manager.rb
```diff
@@ -58,6 +60,10 @@ def create_deleted_uploads_alerts
     DeletedUploadsAlertManager.new(@courses_to_check).create_alerts
   end
 
+  def create_deleted_timeline_alerts
+    DeletedTimelineNotification.new.create_alerts
```
ragesoss, line 64: This will error, because `new` is being called without a required argument.

ragesoss:
The build is failing now because of linting violations. Please have a look at the RuboCop violations and fix those.

app/controllers/courses_controller.rb
```diff
@@ -197,6 +197,9 @@ def notify_untrained
   def delete_all_weeks
     @course = find_course_by_slug(params[:id])
     @course.weeks.destroy_all
+    if @course.approved?
```
ragesoss, line 200: These can be removed in the controllers, since there's now a check for this in the service object.

app/controllers/courses_controller.rb
```diff
@@ -197,6 +197,9 @@ def notify_untrained
   def delete_all_weeks
     @course = find_course_by_slug(params[:id])
     @course.weeks.destroy_all
+    if @course.approved? 
+      DeletedTimelineAlertManager.new(course)
```
ragesoss, line 201: I think this needs a better name than `DeletedTimelineAlertManager`, one that more clearly communicates that it works like a service object (it does a single task). Maybe something like `CheckTimeline` would be a good name, as that's the core idea, to check the timeline after it's been modified and see if anything needs to be done (eg, create an alert).

ragesoss:
Please include a screenshot of what the Alert looks like. (A screenshot from the rails mailer preview will be fine.)

lib/alerts/deleted_timeline_alert_manager.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class DeletedTimelineAlertManager
+  def initialize(course)
+    @course = course
+    create_alerts
+  end
+
+  def create_alerts
+    return unless @course.approved?
+    # it will not create an alert if course doesn't have any training module
```
ragesoss, line 11: This comment has it backwards.

spec/controllers/blocks_controller_spec.rb
```diff
@@ -4,17 +4,33 @@
...
+        expect(Alert.count).to eq(0)
+      end
+    end
+    context 'when course is approved' do
+      it 'create an alert and destroys the block' do
+        course_approved
+        expect(Block.count).to eq(1)
+        delete "/blocks/#{block.id}", params: { id: block.id, format: :json }
+        expect(Block.count).to eq(0)
+        expect(Alert.count).to eq(1)
```
ragesoss, line 32: This should be a little more specific; no reason not to check for the specific alert type, rather than just the count of the superclass Alert.

lib/alerts/check_timeline_manager.rb
```diff
@@ -11,7 +11,7 @@ def create_alerts
     # if course doesn't have any training module, it will not create an alert
     return if @course.training_module_ids.any?
     alert = Alert.create(type: 'CheckTimeline', course_id: @course.id,
-                         message: 'It appears you have removed either the timeline, week or block from your course.')
+                         message: 'It appears you have removed timeline from your course.')
```
ragesoss, line 14: Per my earlier comment, I think this message should be reworded, as it's a message for admins rather than for the user who deleted the timeline. It's also not completely accurate, as deletion of the timeline is not the only condition that will trigger an alert.

I suggest moving the message itself into the alert subtype class file.

kunalthedev: So i don't need to add message while creating alert right?. As the message in the alert subtype class would be showed up. And what do you suggest as changes in message?

kunalthedev: Do i need to add message here? [](https://user-images.githubusercontent.com/72390769/213884087-6650b2ae-8e29-48af-8d95-2260fddae8c3.png)

ragesoss: Adding the message when creating the alert is okay. I'm suggesting that the message string be defined in the class file, rather than here.

app/controllers/blocks_controller.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
-
+require "#{Rails.root}/lib/alerts/check_timeline_manager"
 class BlocksController < ApplicationController
   respond_to :json
 
   def destroy
+    course = Block.find(params[:id]).course
```
ragesoss, line 7: With this data flow, you're doing the same `find` twice. This should be done with as few and as efficient SQL queries as possible.

One alternative would be to do it like this:
`block = Block.find(params(:id).destroy`
`course = block.course`

This destroys the block first (ie, deletes the record from the database), but keeps the object as a variable so that you can still use it to find the corresponding course.

app/controllers/weeks_controller.rb
```diff
@@ -1,10 +1,12 @@
 # frozen_string_literal: true
-
+require "#{Rails.root}/lib/alerts/check_timeline_manager"
 class WeeksController < ApplicationController
   respond_to :json
 
   def destroy
+    course = Week.find(params[:id]).course
```
ragesoss, line 7: As above, this should be done without unnecessary duplication of queries.

lib/alerts/check_timeline_manager.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class CheckTimelineManager
+  def initialize(course)
+    @course = course
+    create_alerts
+  end
+
+  def create_alerts
```
ragesoss, line 9: This should only create one (unresolved) alert for a given course.

I think this can be a `resolvable` alert, such that if an admin marks it as resolved (meaning the Timeline has been fixed so that it's now okay and includes training modules) then a second alert could be created.

Some of the other alert manager files have examples you can follow for how to implement this.

spec/controllers/blocks_controller_spec.rb
ragesoss: Looks like you may have missed this comment from last week. ^

kunalthedev: > This should be a little more specific; no reason not to check for the specific alert type, rather than just the count of the superclass Alert.

@ragesoss what do you mean by this. I wasn't able to understand properly

kunalthedev: > > This should be a little more specific; no reason not to check for the specific alert type, rather than just the count of the superclass Alert.
> 
> @ragesoss what do you mean by this. I wasn't able to understand properly

Do you want me to check for Alert.type to be equal to CheckTimeline?

app/models/alert_types/check_timeline.rb
```diff
@@ -0,0 +1,47 @@
...
+#  updated_at     :datetime         not null
+#  message        :text(65535)
+#  target_user_id :integer
+#  subject_id     :integer
+#  resolved       :boolean          default(FALSE)
+#  details        :text(65535)
+#
+
+# Alert for a course whose timeline is deleted
+class CheckTimeline < Alert
```
ragesoss, line 24: For consistency with other alert types, this should be `CheckTimelineAlert`

spec/controllers/blocks_controller_spec.rb
ragesoss: You could do `CheckTimelineAlert.count` here.

app/models/alert_types/check_timeline.rb
kunalthedev: Apart from this anything needs to be fixed?

spec/controllers/blocks_controller_spec.rb
kunalthedev: @ragesoss instead of that, i did Alert.first type. eq('CheckTimeline') . And it's working you can see my last commit. Should I change it to CheckTimelineAlert.count?

ragesoss: This is fine. I think it's a little better not to rely on database order (ie, via `.first`) in tests, but it's a minor thing.

lib/alerts/check_timeline_alert_manager.rb
```diff
@@ -0,0 +1,23 @@
...
+  def initialize(course)
+    @course = course
+    create_alerts
+  end
+
+  def create_alerts
+    return unless @course.approved?
+    # if course doesn't have any training module, it will not create an alert
+    return if @course.training_module_ids.any?
+    return if alert_already_exists?(@course.id)
```
ragesoss, line 13: There's no need to pass the id as an argument to this method, since the instance variable `@course` is automatically available to every instance method.

lib/alerts/check_timeline_alert_manager.rb
```diff
@@ -0,0 +1,23 @@
...
+    alert = Alert.create(type: 'CheckTimelineAlert',
+                         course_id: @course.id,
```
ragesoss, line 15: It's slightly more elegant and Rails-y to do this with `course: @course` instead of `course_id: @course.id`. (They do the same thing.

lib/alerts/check_timeline_alert_manager.rb
```diff
@@ -0,0 +1,23 @@
...
+                         message: CheckTimelineAlert.default_message)
+    alert.email_content_expert
+  end
+
+  def alert_already_exists?(id)
+    CheckTimelineAlert.exists?(course_id: id, resolved: false)
```
ragesoss, line 21: As above, this can be `course: @course`.

spec/controllers/weeks_controller_spec.rb
```diff
@@ -4,17 +4,34 @@
...
+      end
+    end
+
+    context 'when course is approved' do
+      it 'create an alert and destroys the week' do
+        course_approved
+        expect(Week.count).to eq(1)
+        delete "/weeks/#{week.id}", params: { id: week.id, format: :json }
+        expect(Week.count).to eq(0)
+        expect(Alert.count).to eq(1)
```
ragesoss, line 33: This should be `CheckTimelineAlert.count` as well.

spec/lib/alerts/check_timeline_alert_manager_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+      expect(Alert.count).to eq(0)
+    end
+  end
+
+  context 'when course is approved but has 0 training modules' do
+    it 'create an alert' do
+      adding_blocks_without_training_module
+      course_approved
+      subject
+      expect(Alert.count).to eq(1)
```
ragesoss, line 40: This should be `CheckTimelineAlert.count` as well.

lib/alerts/check_timeline_alert_manager.rb
kunalthedev: @ragesoss I've updated everything mentioned. And Now it's not generating duplicate alert when there is already an existing alert. You can have a look and try it on your local

app/models/alert_types/check_timeline_alert.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def resolve_explanation
+    <<~EXPLANATION
+      Resolving this alert means the course has some timeline. A new
+      alert will be generated if it again nominated or tagged for timeline deletion.
+    EXPLANATION
+  end
+
+  def self.default_message
+    'All the training modules have been removed from the course'
```
ragesoss, line 45: Add a period to the end of this message.

app/models/alert_types/check_timeline_alert.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def resolvable?
+    !resolved
+  end
+
+  def resolve_explanation
+    <<~EXPLANATION
+      Resolving this alert means the course has some timeline. A new
+      alert will be generated if it again nominated or tagged for timeline deletion.
```
ragesoss, line 40: "Resolving this alert means the course has at least one training module again. A new alert will be generated if the last timeline block with a training gets removed."

ragesoss (changes requested):
A couple of text changes, and this should be ready to go.

PR #5243, opened 2023-01-21
Add pagination to articles page

TheTrio:
Looks like this

Recording 2023-01-20 at 19 23 49

The data fetching logic remains the same - we could fetch the data in pages too but its probably not worth the complexity.


app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div style={{
+      display: 'flex',
+      justifyContent: 'space-between',
+      alignItems: 'center',
+    }}
+    >
+      <ReactPaginate
+        pageCount={totalPages}
+        nextLabel="Next"
```
ragesoss, line 24: These labels needs to be done with i18n.

ragesoss:
I tested this locally, and it has some confusing behavior. Whenever all the articles for a course have been fetched (either by clicking 'See More', or because the course has less than 500 articles so they all get fetched initially) then the page controls don't show up. For a course with more than 100 but less than 500, it doesn't show up at all. For one with more than 500, it disappears once the last set of articles is fetched.

TheTrio:
Have addressed these issues. I had moved the "show More" button into the same component as the page controls and so the earlier logic of hiding that button caused the page controls to disappear as well when all the articles were fetched.

ragesoss:
This works smoothly now. I could merge it as-is, but I think there are a couple of UI refinements that would be worthwhile.

1. The 'showing 100 of X' message is useful, even when it's showing all of the articles already (eg, '12 of 12').
2. That message ought to be moved right, inline with the Show More button when it's present.
3. It would be useful if it showed which slice is being displayed (eg, 'Showing 101-200 of 1.5k') when applicable
4.  It would be nice if it worked with the filters, so that the number of pages and the 'showing' message both updated based on which filters are selected. Currently, if you set the filters to exclude some of the articles, then go to the last page, it will be empty. For the 'showing' message, it ought to ideally including both the number of filtered articles and the total without filtering.

Let me know if you want to make some or all of these changes.

TheTrio:
Thanks for those ideas! 

This is how it looks now

Recording 2023-01-27 at 02 36 03

ragesoss:
Looks great. I notice it shows  2300 hundred as the final range. Ideally that would show the actual number instead (the exact version of 2.27k).

PR #5244, opened 2023-01-23
Added an alert which pops up when students go through training very fast

Chaitanyarai899:
## What this PR does
If a student is going through training very quickly and hitting the Next Page button(in training slides) repeatedly in short intervals of time, an alert will pop up. This alert will emphasize on the importance of learning the training content thoroughly.
More specifically, if a student hits the Next Page button 4 times in an interval of 10 seconds, then the alert will pop up. However, the alert will popup only once per module. This feature only works in student training library.
No additional libraries have been used.

## Screenshots
Before:

https://user-images.githubusercontent.com/95732261/214013208-f053085d-3d18-4f2e-ac45-716807c7c327.mp4

After:

https://user-images.githubusercontent.com/95732261/214906833-e91a0259-edba-452f-b271-7c18ed949916.mp4


ragesoss:
Neat!

I'd much prefer that this be done without adding a new library.

I think maybe only showing it once (per module) would be best. I think we'll also want to restrict this feature to modules in the Students training library, so that instructors won't see it if they are going quickly through the instructor-focused modules. I guess the most flexible way to do this would be to configure this feature to be enabled at the level of each individual TrainingModule (via a value in the `TrainingModules#settings` hash; see `training_content/wiki_ed/modules/evaluate-wikipedia-exercise.yml` for the place where we use that settings hash for something else for one module).

Does it work only based on clicking 'next', or will it also trigger when slides are advanced quickly via the right arrow key?

Chaitanyarai899:
Yes, it will get triggered if slides are advanced quickly using the right arrow key.
Okay! I will rebuild the component without using any new libraries and will try to restrict the feature only to student training modules. Also, it will be only shown once(per module of course).
Thank you very much for the instructions!

Chaitanyarai899:
I remade the component without adding any new library. The alert popup only gets triggered in the students training library and not in the  instructor-focused modules. 
The alert will be triggered only once per module not more than that!

app/assets/javascripts/training/components/Alert.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { useState } from 'react';
```
kunalthedev, line 2: You can do import React , {useState} from 'react'. I think that's one of the error in failing build

Chaitanyarai899: Thank you very much for the suggestion!

kunalthedev: You can have a look at the other fails too

Chaitanyarai899: Yes! I am fixing the indentation errors and spacing issues as well. Thanks!

ragesoss:
Thanks for keeping at this! Have you been running the feature test locally?

Chaitanyarai899:
@ragesoss Yes, but it shows more errors than there are when I run the tests locally. It takes a lot of time and shows different unrelated errors. 
I run the tests using `rspec spec/features/training_tool_spec.rb`  and I get different errors every time I run the tests.
Sometimes it says web driver not found, some times it doesn't.
The only actual error is an intercepted click. I am trying to modify the test to check if the alert box is being displayed and click the close button if it exists. 
Actually, I had no experience with testing and had ongoing exams. So, it is taking a lot more time than expected. If it fails this time as well then I'll try running the tests in github codespaces.
I am very sorry for the inconvenience caused.

ragesoss:
No problem! If you'd like help debugging your local tests, post some details in Slack.

app/assets/javascripts/training/components/training_slide_handler.jsx
```diff
@@ -76,11 +87,27 @@ const TrainingSlideHandler = () => {
...
 
+
+  const [isShown, setIsShown] = useState(false);
+
+  // this function checks whether alert should be shown or not.
+  const Alerthandler = () => {
+    if (routeParams.library_id === 'students') {
+    count += 1;
+    const clickingTime = new Date().getTime() - enteringTime;
+    if (count > 3 && clickingTime < 10000 && nooftimes < 2) {
```
ragesoss, line 98: There are a lot of unnamed numbers here. The clickingTime comparison value, in particular, should be a well-named constant, and the other two values should probably also be constants with names that explain what they are for (eg, maxTooFastAlerts or similar).

app/assets/javascripts/training/components/Alert.jsx
```diff
@@ -0,0 +1,23 @@
+import React, { useState } from 'react';
+
+export default function Alert() {
```
ragesoss, line 3: This file and component needs a better name. It's not a general alert (which I would expect to take props to control what message it shows), it's a very specific one with a message built in.

app/assets/stylesheets/training_modules/_Alert.styl
```diff
@@ -0,0 +1,53 @@
+.backdrop
```
ragesoss, line 1: This isn't a very specific class name.

app/assets/javascripts/training/components/training_slide_handler.jsx
Chaitanyarai899: Thank you for your input, I'll fix this asap.

spec/features/training_tool_spec.rb
```diff
@@ -264,6 +264,11 @@ def check_slide_contents(slide, slide_number, slide_count)
 end
 
 def proceed_to_next_slide
+  if page.has_selector?('.alert-box-container', wait: 10)
+    within('.alert-box-container') do
+      page.execute_script("$('.alert-box-container').hide()")
```
ragesoss, line 269: I'd rather avoid introducing new cases of `execute_script`. Can this just click 'CLOSE' instead?

app/assets/javascripts/training/components/training_slide_handler.jsx
```diff
@@ -76,11 +90,27 @@ const TrainingSlideHandler = () => {
     }));
   };
 
+
+  const [isShown, setIsShown] = useState(false);
+
+  // this function checks whether fastTrainingAlert should be shown or not.
+  const fast_Training_Alert_handler = () => {
```
ragesoss, line 97: This function name should be in camelCase to match the style of the rest of our JS.

app/assets/javascripts/training/components/training_slide_handler.jsx
```diff
@@ -59,13 +62,24 @@ const getSlideInfo = (training, locale) => {
 
 const keys = { rightKey: 39, leftKey: 37 };
 
+
+// helper variables for Alerthandler function.
```
ragesoss, line 66: I don't like how much of this logic is getting added to the TrainingSlideHandler, especially considering that it's spread out ins two different sections. The handler file is already too complex.

Can this be done in a way where all or almost all of the logic for whether to display the modal is contained within the FastTrainingAlert file instead?

Chaitanyarai899: Done! I was successful in moving the fastTrainingAlertFunction as well as all the helper variables to the FastTrainingAlert file. I have only called the function in TrainingSlideHandler and used one only more line for the Usestate hook as there is no other way around. Now, TrainingSlideHandler has barely three lines of code added.

spec/features/training_tool_spec.rb
Chaitanyarai899: Done! I've removed additional script and used .click for achieving the same.

app/assets/javascripts/training/components/fast_training_alert.jsx
```diff
@@ -0,0 +1,44 @@
...
+        </div>
+      )};
+    </>
+  );
+}
+
+let count = 0;
+let nooftimes = 0;
+const enteringTime = new Date().getTime();
+const max_no_times_alert_shown = 1;
```
ragesoss, line 28: Since these three constants are truly just constant integers, make them ALL_CAPS for consistency with the rest of the codebase. Also, include an inline comment for the 'min time' with the value in seconds.

Chaitanyarai899: Done!

PR #5247, opened 2023-01-25
Tickets should be searchable by course #5224

### LINKED ISSUE #5224: Tickets should be searchable by Course
When viewing a course, or viewing a single ticket, it would be really useful for an admin to be able to view all the other tickets associated with the same course. For a ticket (`TicketDispenser::Ticket`), the `project_id` is the ID of the associated course (and calling `Ticket#project` will return a `Course` object).

Ideally, this would be implemented as an additional ticket search mode, taking a course slug as input, then loading the associated course record and its ID and finding all tickets with the matching `project_id`. Then we could easily add links that include the course slug as a query parameter, both from the ticket `show` view and from a course page, which would load all that course's ticket's when followed.

cyrillefr:
## What this PR does
This PR aims to resolve #5224 .

That is : 
 - add a new option for searching tickets: by course (by course slug)
 - also add a link from a course to search for all tickets related to that very course (for admins)
 - also add a link from a ticket detail to search for all other tickets related to the same course that this very ticket. 

Click one of this 2 new links display the dashboard with a search done and the matching tickets listed.

### DIFF
--- a/app/assets/javascripts/components/tickets/sidebar.jsx
+++ b/app/assets/javascripts/components/tickets/sidebar.jsx
@@ -48,6 +48,11 @@ export class Sidebar extends React.Component {
               : 'Course Unknown'
           }
         </section>
+        <section className="course-name">
+          <Link target="_blank" to={`/tickets/dashboard?search_by_course=${ticket.project.slug}`}>
+            Search all tickets for course: {ticket.project.title}
+          </Link>
+        </section>
         <section className="course-user-details">
           {
             ticket.project.id && ticket.sender.username
--- a/app/assets/javascripts/components/tickets/sidebar.jsx
+++ b/app/assets/javascripts/components/tickets/sidebar.jsx
@@ -48,7 +48,7 @@ export class Sidebar extends React.Component {
               : 'Course Unknown'
           }
         </section>
-        <section className="course-name">
+        <section className="related-tickets">
           <Link target="_blank" to={`/tickets/dashboard?search_by_course=${ticket.project.slug}`}>
             Search all tickets for course: {ticket.project.title}
           </Link>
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/components/tickets/sidebar.jsx
```diff
@@ -48,6 +48,11 @@ export class Sidebar extends React.Component {
               : 'Course Unknown'
           }
         </section>
+        <section className="related-tickets">
+          <Link target="_blank" to={`/tickets/dashboard?search_by_course=${ticket.project.slug}`}>
+            Search all tickets for course: {ticket.project.title}
```
ragesoss, line 53: Since you've already created an i18n message for this, you may as well use it here.

cyrillefr: You will say that I am nitpicking :), but I did not use i18n here because there were none, so I guess I had to blend with the existing code.

Also, the i18 string I have added is : "Search all tickets for this course" and is used for the other place.
The string here is without "this".
We would have "Search all tickets for this course: course XXX".
It seems a bit of a repetition.
However, English is not my mother tongue :) and I'll make the change.

ragesoss: You're right, it's a bit different. In this case, I think I'd rather have a reusable i18n that works for both places, rather than the perfect message that avoids repetition. But I think 'this course' or 'course' are not needed here anyway. One alternative would be 'Search all tickets for: XXX' (which I'm fine with not using i18n for). That would make the sidebar a little less cluttered too.

ragesoss:
Thank you! This is going to be really useful for my Wiki Education colleagues.

PR #5249, opened 2023-02-01
add print button to article viewer

### LINKED ISSUE #4313: Article Viewer should have a Print feature
The Article Viewer's authorship highlighting is useful for evaluating and sharing the changes made to a Wikipedia article, and some instructors want to be able to print this or save it as a PDF (so that, for example, they can annotate it by hand, or simply save a record of it).

Using built-in browser print features doesn't provide a usable print version, because of the modal layout of the Article Viewer over the rest of the page. The default print version also doesn't render the authorship highlighting.

We could probably implement a printable version by:
* Using a popup window with just the article viewer HTML
* Using a stylesheet designed for print to style the popup window (or adding print styling to the main stylesheet)

TheTrio:
Fixes #4313 

This is how it looks

https://user-images.githubusercontent.com/10794178/216028491-c87ebaf2-0b5d-449f-9a89-46f801dc44c3.mp4

We will need to print the legend as well to make sense of the highlighting but I wanted to know how you felt about this implementation.

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/components/Footer.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/components/Footer.jsx
@@ -63,7 +63,7 @@ export const Footer = ({
         }}
         onClick={printArticleViewer}
       >
-        Print
+        {I18n.t('application.print')}
       </button>
     </div>
   );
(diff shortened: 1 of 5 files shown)


ragesoss:
This looks great! I suggest putting the legend at the top for printing. Also, you should move all the printing logic to a separate file.

ragesoss:
This works nicely, including for cases with many students. Can you remove the background color?

TheTrio:
The background of the page?

ragesoss:
The background color is light gray, rather than white, so printing will use a lot of extra toner and will also be less readable than it should be. I think just explicitly setting the background color to white for the print version would do the trick.

Issue #1760, opened 2018-03-07
Provide a useful error message when course home wiki is changed to an invalid wiki


ragesoss:
Currently, the message is just 'Internal Server Error'. It should be something more useful.

You can replicate this issue as follows:
1. create a new program
2. click edit details, and change the language and wiki; you have to change it to one that doesn't exist.
for example, pick language `egl` and project `wikinews`

The option to change Home Wiki Language and Home Wiki Project in the Edit Details view will only show up if your dev environment is configured in the non-Wiki Education setting, which means that you must edit application.yml and change the `dashboard_url` value (to anything but `dashboard.wikiedu.org`).

What happens on the server is that, if a Wiki record doesn't already exist, it tries to create one. But first, it pings the server of that wiki to confirm that it exists. There is no egl.wikinews.org (but it's a valid language, probably for wikipedia) so throwing the error is correct and should be shown to the user.
but the message should be understandable, informing the user that the requested wiki doesn't exist.

![wiki error](https://user-images.githubusercontent.com/848483/40197411-3a941a24-59c8-11e8-88b7-4bdbfecd8fd8.png)

Note: When this gets fixed, we need to close this one as well: https://phabricator.wikimedia.org/T258320

mtrivera:
I'm interested in working on this, do you have any resources for Rails? I have done some basic Ruby before.

psinghal20:
Hi @mtrivera , in my opinion the rails documentation is one of the best resource. If you have a basic understanding of the MVC architecture, it will be much easier for you to work with rails.

mtrivera:
@psinghal20 Thanks, I will look into it.

mtrivera:
Having issues with my dev environment setup. If someone else wants to work on this, go ahead.

RyanHiller:
Hey @ragesoss a group of students and myself were hoping to tackle this issue as a starting point for contributing to the repo. Is this issue still open? I got my dev environment set up however I am not seeing a language field in order to recreate the error. Not sure if this is due to changes in the project or if I am not recreating the program correctly. Any help would be appreciated.

![Screenshot_2019-10-22 Test2 — Programs and Events Dashboard](https://user-images.githubusercontent.com/855005/67314405-d3c0bc80-f4c1-11e9-8fd4-3e4351e0ef45.png)

ragesoss:
The interface changed so that now the language is selected as part of the home_wiki or tracked_wikis field.

[...]

PR #5251, opened 2023-02-02
error message when home wiki is changed to invalid

kunalthedev:
### **Issue #1760**
Current workflow goes like
initially it goes through `get_or_create` if provided wiki doesn't exist then it tries to create a new one in db, to create new wiki for db, we have 2 checks. 

1. ensuring valid project(which just checks if the project name is mentioned in the provided array or not)

2. ensuring wiki exists(which pings the given server using language and project name provided ex lang = "egl' and project name = "wikidata")

in the above example, since, it doesn't exist so, it will create , for that it will first go through first check which will pass since the language and project provided is valid but for the 2nd check it will fail since wiki doesn't exist. 

Old error:- 
image

Updated error
Screenshot from 2023-02-04 00-59-09

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -49,6 +49,8 @@ class CoursesController < ApplicationController
     ensure_passcode_set
     UpdateCourseWorker.schedule_edits(course: @course, editing_user: current_user)
     render json: { course: @course }
+  rescue Wiki::InvalidWikiError => e
+    render json: { errors: e, message: I18n.t('courses.error.invalid_language_or_project') }, status: :not_found
   end
 
   def destroy
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -50,7 +50,10 @@ class CoursesController < ApplicationController
     UpdateCourseWorker.schedule_edits(course: @course, editing_user: current_user)
     render json: { course: @course }
   rescue Wiki::InvalidWikiError => e
-    render json: { errors: e, message: I18n.t('courses.error.invalid_language_or_project') },
+    language = params.dig(:course, :home_wiki, :language)
+    project = params.dig(:course, :home_wiki, :project)
+    message = I18n.t('courses.error.invalid_wiki', language:, project:)
+    render json: { errors: e, message: },
            status: :not_found
   end
 
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -50,9 +50,8 @@ class CoursesController < ApplicationController
     UpdateCourseWorker.schedule_edits(course: @course, editing_user: current_user)
     render json: { course: @course }
   rescue Wiki::InvalidWikiError => e
-    language = params.dig(:course, :home_wiki, :language)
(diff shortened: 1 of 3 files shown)


ragesoss:
Cool! I think we can do better with the error message. It ought to say something like "Cannot connect to egl.wikinews.org. It may not be a valid wiki."

kunalthedev:
> Cool! I think we can do better with the error message. It ought to say something like "Cannot connect to egl.wikinews.org. It may not be a valid wiki."

Done @ragesoss

app/controllers/courses_controller.rb
```diff
@@ -49,6 +49,12 @@ def update
     ensure_passcode_set
     UpdateCourseWorker.schedule_edits(course: @course, editing_user: current_user)
     render json: { course: @course }
+  rescue Wiki::InvalidWikiError => e
+    language = params.dig(:course, :home_wiki, :language)
```
ragesoss, line 53: Hmm... doing it this way will result in an incorrect error message if, rather the the home wiki, the invalid wiki comes from the set of additional tracked wikis. Rather than digging for the parameter from the request, I think the correct behavior should be to get the invalid domain from the error object. (This might require a change to how the error in generated in wiki.rb.)

kunalthedev: Hmm. So you mean i should pass incorrect domain through the error obj in wiki.rb and should try to fetch that obj in here.?

ragesoss: Yes. The error object is available already (`e`) but I'm not sure whether it currently includes the project and language for the invalid wiki.

kunalthedev: > Yes. The error object is available already (`e`) but I'm not sure whether it currently includes the project and language for the invalid wiki.

@ragesoss I've Done this please check my approach once. is this what you were talking about?

ragesoss (changes requested):

app/models/wiki.rb
```diff
@@ -177,17 +177,27 @@ def ensure_valid_project
...
-    raise InvalidWikiError if site_info.nil?
+    raise InvalidWikiError.new(project, language) if site_info.nil?
     servername = site_info.data.dig('general', 'servername')
-    raise InvalidWikiError unless base_url == "https://#{servername}"
+    raise InvalidWikiError.new(project, language) unless base_url == "https://#{servername}"
   end
 
-  class InvalidWikiError < StandardError; end
+  class InvalidWikiError < StandardError
+    def initialize(project, language)
```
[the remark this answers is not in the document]
kunalthedev: How can i use predefined `domain_name` method inside InvalidWikiError. I wasn't able to do that. ? 
Instead , I can pass domain name to the initialization.

ragesoss:
Can you update the screenshot?

kunalthedev:
> Can you update the screenshot?

Done @ragesoss

ragesoss:
Great. This looks good. I'll have one more look through the code today and test it out locally, and then it should be good to go.

kunalthedev:
> Great. This looks good. I'll have one more look through the code today and test it out locally, and then it should be good to go.

Cool!

ragesoss (approved):

PR #5254, opened 2023-02-08
cache i18n files

TheTrio:
This speeds up the build process substantially.

https://user-images.githubusercontent.com/10794178/217521049-8129bbdc-a125-4484-a7c0-6586c383797c.mp4

### DIFF
--- a/prepare.js
+++ b/prepare.js
@@ -6,19 +6,48 @@ function elapsed(hrtime) {
   return seconds;
 }
 
+function getHashOfI18nFiles() {
+  return execSync('cat config/locales/* | sha1sum').toString().split('-')[0].trim();
+}
+
 let start = process.hrtime();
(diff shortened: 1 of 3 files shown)


prepare.js
```diff
@@ -6,19 +6,48 @@ function elapsed(hrtime) {
...
+
 let start = process.hrtime();
 const execSync = require('child_process').execSync;
+const exec = require('child_process').exec;
+
 const config = require('./config');
+const fs = require('fs');
+
+let fullRebuild = false;
+try {
```
ragesoss, line 21: Can you add comments explaining how this works? The code is pretty clear, but this is the kind of things that developers will appreciate not needing to read any code to understand.

TheTrio:
How does this look like?

PR #5262, opened 2023-02-18
Addedview and controller for updating username

kunalthedev:
## What this PR does
With reference to issue #4978 .
This PR creates a way for user to add his/her username and trigger update username withoug logging out.

Error message when wrong username is entered
image

Message when right username is entered
image

### DIFF
--- a/app/assets/stylesheets/modules/_notifications.styl
+++ b/app/assets/stylesheets/modules/_notifications.styl
@@ -32,7 +32,7 @@
   position relative
   z-index 9999
   width 100%
-  text-align left
+  text-align center
   background-color warning
   color #fff
   > .container
--- a/app/assets/stylesheets/modules/_notifications.styl
+++ b/app/assets/stylesheets/modules/_notifications.styl
@@ -32,7 +32,7 @@
   position relative
   z-index 9999
   width 100%
-  text-align center
+  text-align left
   background-color warning
   color #fff
   > .container
@@ -42,3 +42,7 @@
         cursor pointer
         color smoke
         outline none
+
+.notice--error
+  @extends .notice
+  text-align center
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -3,16 +3,15 @@
 require_dependency "#{Rails.root}/lib/importers/user_importer"
 
 class UpdateUsernameController < ApplicationController
-    before_action :require_signed_in
-    respond_to :html, :json
-    def index
-        @user = User
-    end
+  before_action :require_signed_in
+  respond_to :html, :json
+  def index
+    @user = User
+  end
 
-    def update
-        @username = params['username']
-        puts @username
-        UserImporter.update_username_for_global_id(@username)
-        redirect_to '/'
-    end
+  def update
+    @username = params['username']
+    UserImporter.update_username_for_global_id(@username)
+    redirect_to '/'
+  end
 end
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -13,16 +13,14 @@ class UpdateUsernameController < ApplicationController
     @username = params['username']
     if @username.blank?
       flash[:error] = t('update_username.empty_username')
-      redirect_to '/update_username'
     else
       user = UserImporter.update_username_for_changed_metawiki_usernames(@username)
       if user.nil?
         flash[:error] = t('update_username.not_found')
-        redirect_to '/update_username'
       else
         flash[:notice] = t('update_username.username_updated')
-        redirect_to '/update_username'
       end
     end
+    redirect_to '/update_username'
   end
 end
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -11,7 +11,18 @@ class UpdateUsernameController < ApplicationController
 
   def update
     @username = params['username']
-    UserImporter.update_username_for_changed_metawiki_usernames(@username)
-    redirect_to '/'
+    if @username.blank?
+      flash[:error] = t('update_username.empty_username')
+      redirect_to '/update_username'
+    else
+      user = UserImporter.update_username_for_changed_metawiki_usernames(@username)
+      if user.nil?
+        flash[:error] = t('update_username.not_found')
+        redirect_to '/update_username'
+      else
+        flash[:notice] = t('update_username.username_updated')
+        redirect_to '/update_username'
+      end
+    end
   end
 end
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/importers/user_importer"
+
+class UpdateUsernameController < ApplicationController
+    before_action :require_signed_in
+    respond_to :html, :json
+    def index
+        @user = User
+    end
+
+    def update
+        @username = params['username']
+        puts @username
+        UserImporter.update_username_for_global_id(@username)
+        redirect_to '/'
+    end
+end
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -10,12 +10,12 @@ class UpdateUsernameController < ApplicationController
   def update
     username = UserImporter.sanitize_username params['username']
     user = UserImporter.update_username_for_global_id username
+    redirect_to '/update_username'
     return if username.blank?
     if user.nil?
       flash[:error] = t('update_username.not_found')
     else
       flash[:notice] = t('update_username.username_updated')
     end
-    redirect_to '/update_username'
   end
 end
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -5,8 +5,7 @@ require_dependency "#{Rails.root}/lib/importers/user_importer"
 class UpdateUsernameController < ApplicationController
   before_action :require_signed_in
   respond_to :html, :json
-  def index
-  end
+  def index; end
 
   def update
     username = UserImporter.sanitize_username params['username']
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -11,15 +11,13 @@ class UpdateUsernameController < ApplicationController
 
   def update
     @username = params['username']
+    user = UserImporter.update_username_for_changed_metawiki_usernames(@username)
     if @username.blank?
       flash[:error] = t('update_username.empty_username')
+    elsif user.nil?
+      flash[:error] = t('update_username.not_found')
     else
-      user = UserImporter.update_username_for_changed_metawiki_usernames(@username)
-      if user.nil?
-        flash[:error] = t('update_username.not_found')
-      else
-        flash[:notice] = t('update_username.username_updated')
-      end
+      flash[:notice] = t('update_username.username_updated')
     end
     redirect_to '/update_username'
   end
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -11,7 +11,7 @@ class UpdateUsernameController < ApplicationController
 
   def update
     @username = params['username']
-    UserImporter.update_username_for_global_id(@username)
+    UserImporter.update_username_for_changed_metawiki_usernames(@username)
     redirect_to '/'
   end
 end
--- a/app/controllers/update_username_controller.rb
+++ b/app/controllers/update_username_controller.rb
@@ -6,15 +6,13 @@ class UpdateUsernameController < ApplicationController
   before_action :require_signed_in
   respond_to :html, :json
   def index
-    @user = User
   end
 
   def update
-    @username = params['username']
-    user = UserImporter.update_username_for_changed_metawiki_usernames(@username)
-    if @username.blank?
-      flash[:error] = t('update_username.empty_username')
-    elsif user.nil?
+    username = UserImporter.sanitize_username params['username']
+    user = UserImporter.update_username_for_global_id username
+    return if username.blank?
+    if user.nil?
       flash[:error] = t('update_username.not_found')
     else
       flash[:notice] = t('update_username.username_updated')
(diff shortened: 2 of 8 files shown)


spec/features/update_username_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    it 'visits and enter updated username' do
+      VCR.use_cassette('update_username') do
+        create(:user, id: 500, username: 'Ragesoss')
+        new_username = 'New Username'
+        User.find(500).update_attribute(:username, new_username)
+        visit '/update_username'
+        fill_in('username', with: new_username)
+        click_button 'Update Username'
+        expect(User.find(500).username).to eq(new_username)
```
ragesoss, line 26: This test doesn't make sense. The form should update the username to whatever their current username is according to Wikimedia servers, based on their global_id. But in this example, the user does not have a global_id and the test looks for the same username ("New Username") that user 500 already had before using the form.

spec/features/update_username_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+
+describe 'Update username controller', type: :feature, js: true do
+  let!(:user) { create(:user) }
+
+  before do
+    login_as user
+  end
+
+  context 'Update Username page' do
+    it 'checks if page is visited' do
```
ragesoss, line 13: This is not necessary.

app/views/update_username/index.html.haml
```diff
@@ -0,0 +1,6 @@
+.container
+  %h2= 'Enter Username to update'
```
ragesoss, line 2: All the text should be done via the i18n system. I think the messages can be improved as well:

"Update a changed username" is a better header. The input label can be just "Username", and the submit label can be "Update". The submit link should be styled as a button.

I think a fuller explanation right above the form would also be useful: "If a user has changed the username for their Wikimedia account, enter it here to update the username on the Dashboard."

Will this also work if you enter the new username, or only if you enter the old one?

spec/features/update_username_spec.rb
kunalthedev: Sorry i forgot to remove it

app/views/update_username/index.html.haml
kunalthedev: Accordint to the code, you've to enter your updated Wikimedia username, then it searches for the global_id associated with it, then it updates your username correspond to that global_id in DB.

spec/features/update_username_spec.rb
kunalthedev: So What i'll do is i'll create a new user and assign some global_id to it as well then I'll update its username using update_attribute. then I'll check. Will that work?

ragesoss: You can use the example in user_importer_spec.rb line 149 as an example.

app/views/update_username/index.html.haml
ragesoss: Okay, so it should explicitly say "New username", and the explanation should say "If a user has changed the username for their Wikimedia account, enter the new username here to update it on the Dashboard."

spec/features/update_username_spec.rb
kunalthedev: @ragesoss I've updated it in that way which you mentioned

app/controllers/update_username_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/importers/user_importer"
+
+class UpdateUsernameController < ApplicationController
+  before_action :require_signed_in
+  respond_to :html, :json
+  def index
+    @user = User
```
ragesoss, line 9: I don't think this does anything.

kunalthedev: Yes but do i need to initialize anything in the index method(controller)? To make the page render?

kunalthedev: I'll probably try to remove it and see if everything works properly. Else @ragesoss just let me know where I need to add link to this page?

app/controllers/update_username_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def update
+    @username = params['username']
+    UserImporter.update_username_for_global_id(@username)
+    redirect_to '/'
```
ragesoss, line 15: Rather than redirect to the home page, I think this ought to redirect back to update username index, with a confirmation message (or a failure message, if the user couldn't be found).

app/controllers/update_username_controller.rb
ragesoss: You should keep the index method, it just doesn't need to set any instance variables. `@user = User` is just setting the instance variable `@user` to point to the class `User` (rather than an instance of that class). That's almost never something we'd want to do, as it's likely to cause confusion; everywhere else in the codebase, an instance variable with that name would be used for a single user record, not the class itself.

kunalthedev: Ok I'll just keep empty method

app/controllers/update_username_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+class UpdateUsernameController < ApplicationController
+  before_action :require_signed_in
+  respond_to :html, :json
+  def index
+    @user = User
+  end
+
+  def update
+    @username = params['username']
+    UserImporter.update_username_for_global_id(@username)
```
ragesoss, line 14: Upon further reflection, I think this should not be calling `UserImporter.update_username_for_global_id` without sanitizing the user input first. I think that will require some refactoring of UserImporter, because the sanitization we want is in the first part of the .new_from_username method, lines 32 through 48. That could be extracted into its own method and used here.

config/locales/en.yml
```diff
@@ -1285,6 +1285,13 @@ en:
     no_training_library_records_non_wiki_ed_mode: There are no TrainingLibrary records in the database.
     search_training_resources: Search for training resources
     no_training_resource_match_your_search: No training resource matches your search
+
+  update_username:
+    header: Update a changed username
+    info: If a user has changed the username for their Wikimedia account, enter the new username here to update it on the Dashboard.
+    label: New Username
```
ragesoss, line 1292: No need to capitalize `Username` here.

spec/features/update_username_spec.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'Update username controller', type: :feature, js: true do
+  let!(:user) { create(:user, id: 500, username: 'Temp') }
```
ragesoss, line 6: This user is only used for logging in; there's no need to specify an ID or username, and there's no need to use `let!` instead of `let`, since it gets called in the `before` block.

spec/features/update_username_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  before do
+    login_as user
+  end
+
+  context 'Update Username page' do
+    it 'visits and enter updated username' do
+      VCR.use_cassette('update_username') do
+        create(:user, id: 1, username: 'Old Username', global_id: '14093230')
```
ragesoss, line 15: It's better to avoid setting arbitrary values that don't matter to the test, in this case `id`. You can find the user in the `expect` statement below via global_id instead.

app/controllers/update_username_controller.rb
ragesoss: I think the only place that should link to this page would be the `/admin` index, in the 'Useful links' list. We'll add another link from an FAQ, but that content is not part of the codebase itself.

config/routes.rb
```diff
@@ -447,4 +447,8 @@
   match '/422', to: 'errors#unprocessable', via: :all
   match '/599', to: 'errors#login_error', via: :all
   match '/500', to: 'errors#internal_server_error', via: :all
+
+  #Enhancements
```
ragesoss, line 451: This is not a helpful comment. I suggest moving these near conceptually similar routes, such as the `# Users` section at line 62.

app/controllers/update_username_controller.rb
kunalthedev: Okk

spec/features/update_username_spec.rb
kunalthedev: @ragesoss if i dont specify ID and username here, then it was creating default user which was with username 'Ragesock' and same global_id which i need to use in below test case, So, it was giving DiplicateEntry Error

ragesoss: You can specify a username without an `id`. There's no need for any explicit `id` values in this test file.

lib/importers/user_importer.rb
```diff
@@ -75,6 +75,30 @@ def self.update_users(users=nil, wiki=nil)
     end
   end
 
+  # There are some users who updated their username through MetaWiki, and
+  # to get their updated username get reflected in the dahsboard, they need to relogin.
+  # So, this method will update the username without making the user relogin
+  def self.update_username_for_changed_metawiki_usernames(username)
```
ragesoss, line 81: Don't duplicate all this code for sanitizing input. Extract it to a new method, and then call that method from `new_from_username`.

spec/features/update_username_spec.rb
```diff
@@ -12,12 +12,12 @@
   context 'Update Username page' do
     it 'visits and enter updated username' do
       VCR.use_cassette('update_username') do
-        create(:user, id: 1, username: 'Old Username', global_id: '14093230')
+        create(:user, username: 'Old Username', global_id: '14093230')
```
ragesoss, line 15: Include a comment noting that this global ID belongs to user 'Ragesock', since that isn't obvious just from reading this file.

lib/importers/user_importer.rb
kunalthedev: Oh, ya sorry I'll do it

spec/features/update_username_spec.rb
kunalthedev: Okie

lib/importers/user_importer.rb
```diff
@@ -86,6 +76,27 @@ def self.update_user_from_auth(user, auth)
                 wiki_secret: auth.credentials.secret)
   end
 
+  def self.username_sanitization(username)
```
ragesoss, line 79: Methods are usually named for what they do or what they return. In this case, I think `sanitize_username` is a good name.

app/controllers/update_username_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+require_dependency "#{Rails.root}/lib/importers/user_importer"
+
+class UpdateUsernameController < ApplicationController
+  before_action :require_signed_in
+  respond_to :html, :json
+  def index
+    @user = User
+  end
+
+  def update
```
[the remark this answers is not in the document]
kunalthedev: Check now if it's fine

app/controllers/update_username_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/importers/user_importer"
+
+class UpdateUsernameController < ApplicationController
+  before_action :require_signed_in
+  respond_to :html, :json
+  def index
+    @user = User
```
ragesoss, line 9: You forgot to remove this instance variable to make this an empty method.

spec/features/update_username_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        expect(page).to have_content(I18n.t('update_username.not_found'))
+      end
+    end
+
+    it 'and do not enter username' do
+      VCR.use_cassette('update_username') do
+        visit '/update_username'
+        fill_in('username', with: '')
+        click_button 'update_username'
+        expect(page).to have_content(I18n.t('update_username.empty_username'))
```
ragesoss, line 42: I think this behavior is not necessary. I'd rather it just redirect without a message if the form is submitted without a username.

Ideally, the button would not be enabled unless there is text in the input.

app/controllers/update_username_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def update
+    @username = params['username']
```
ragesoss, line 13: I think this would be a little clearer if this controller explicitly handled the sanitization, like this:

```ruby
username = UserImporter.sanitize_username params['username']
user = UserImporter.update_username_for_global_id username
```

That way, there will be no doubt that input sanitization is happening if someone reads this file but not UserImporter.

Also, I don't think the `username` variable needs to be an instance variable, as it's not relied upon by any other methods in this controller and it's not used by the render template.

ragesoss:
This works as expected, and I think it will be fine to merge once the changes I just requested are implemented.

However, there is some room for improvement on the UI, if you'd like to put in a little more work on it:

* The header would look better with a little bit of space between it and the nav bar. (I'm pretty sure you can find an appropriate CSS class used elsewhere that can do this.)
* The error notice is left-aligned, while the success notice is center-aligned. Making them both center-aligned would be best.

app/assets/stylesheets/modules/_notifications.styl
```diff
@@ -32,7 +32,7 @@
   position relative
   z-index 9999
   width 100%
-  text-align left
+  text-align center
```
ragesoss, line 35: You can't be confident that this change won't mess up other places in the UI that use this class.

I suggest, instead, that you either add an existing class to the html that will take precedence, or add a new class and use it.

app/controllers/update_username_controller.rb
```diff
@@ -6,15 +6,13 @@ class UpdateUsernameController < ApplicationController
...
 
   def update
-    @username = params['username']
-    user = UserImporter.update_username_for_changed_metawiki_usernames(@username)
-    if @username.blank?
-      flash[:error] = t('update_username.empty_username')
-    elsif user.nil?
+    username = UserImporter.sanitize_username params['username']
+    user = UserImporter.update_username_for_global_id username
+    return if username.blank?
```
ragesoss, line 14: I think still needs to redirect before returning, or it will error.

kunalthedev: You mean i should redirect first then i should do other things like checking conditions and flashing messages?

ragesoss: I think that would work, since every path through this method should redirect the same way.

kunalthedev: image


Naah, Doing this is giving internal server error.
the way i am doing is not giving error though

kunalthedev: Cool i've made required changes

PR #5269, opened 2023-03-01
improve survey population script

TheTrio:
This attempts to make the survey script more realistic in terms on the survey it generates.

Among other things, this PR adds the following

1. creates 100 users
2. creates 100 courses
3. each course has 20 users
4. creates 2 question groups
5. each question group has 50 questions
6. each question has 100 responses
7. creates 1 survey assignment
8. the survey assignment has ~1500 survey notifications

### DIFF
--- a/setup/populate_surveys.rb
+++ b/setup/populate_surveys.rb
@@ -1,18 +1,76 @@
 # frozen_string_literal: true
 
+def populate_courses
+  clear_courses
+  populate_users
+  
+  n_courses = ENV['courses'] || 100
+  to_insert = []
+  n_courses.to_i.times do |i|
+    to_insert << {
+      title: "generated course",
+      slug: "course-#{i}",
+    }
+  end
+  Course.insert_all(to_insert)
+  courses_users = []
+  to_insert.each_with_index do |course, i|
+    course_id = Course.find_by(slug: "course-#{i}").id
+    1..20.times do |i|
+      user_id = User.where(email: "generated user").order("RAND()").first.id
+      courses_users << {
+        course_id: course_id,
+        user_id: user_id,
+        role: 0
+      }
+    end
+  end
+  CoursesUsers.insert_all(courses_users)
+end
+
+def populate_users
+  clear_users
+  n_users = ENV['users'] || 100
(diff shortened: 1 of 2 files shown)


setup/populate_surveys.rb
```diff
@@ -1,19 +1,77 @@
...
+    to_insert << {
+      title: "generated course",
+      slug: "course-#{i}",
+    }
+  end
+  Course.insert_all(to_insert)
+  courses_users = []
+  to_insert.each_with_index do |course, i|
+    course_id = Course.find_by(slug: "course-#{i}").id
+    1..20.times do |i|
```
ragesoss, line 20: A more efficient and Rails-y way to do this would be to fetch the random set of users in one query and then iterate through them. Since you just created the courses and this is a dev environment where we can assume that no other database transactions are happening at the same time, you could just do `Course.last(n_courses)` and iterate through them.then you don't need to worry about an iterator either.

Also, you have a nested loop here, but both the inner and outer loop use the same index variable. In terms of scoping this is not a problem, but it's good practice to always use separate names (eg, i, j, m, n) in cases like this for readability.

setup/populate_surveys.rb
```diff
@@ -62,28 +159,66 @@ def populate_survey_answers
   Rapidfire::Answer.insert_all(to_create)
 end
 
-def clear_survey_answers
-  question_group = Rapidfire::QuestionGroup.find_by(name: "Populated Survey")
-  return unless question_group
+def clear_survey_answers 
+  question_group_1 = Rapidfire::QuestionGroup.find_by(name: "generated group 1")
+  question_group_2 = Rapidfire::QuestionGroup.find_by(name: "generated group 2")
```
ragesoss, line 164: This pattern, creating several variables that are defined separately and have indexes in their names, is something of a code smell. For both readability and extensibility, you could refactor this to use an iterator of some kind, which would make it trivial to change the number of question groups from 2 to something larger later on.

TheTrio:
Have made a few changes.

ragesoss:
Looks good, and works smoothly.

PR #5273, opened 2023-03-04
further optimize survey results

TheTrio:
This does the following optimizations:

1. caches the users for the entire survey rather than per question group
2. computes number of answers per question eagerly so that no queries are needed while rendering the template
3. attach question to answer object to avoid extra queries while rendering the template

## Setup

Simply run `rake dev:populate_surveys` and visit `/surveys` and find the survey named "generated survey". 

## Before ~ 13 seconds

image

## After ~ 7 seconds

image

### DIFF
--- a/app/controllers/surveys_controller.rb
+++ b/app/controllers/surveys_controller.rb
@@ -150,6 +150,7 @@ class SurveysController < ApplicationController
   def set_question_groups
     @question_groups = Rapidfire::QuestionGroup.all
     @surveys_question_groups = SurveysQuestionGroup.by_position(params[:id])
+    @survey_user_cache = {}
   end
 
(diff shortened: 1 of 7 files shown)


ragesoss:
Nice. I haven't looked at the code yet, but it looks like you've eliminated an N+1 that was getting called each loop through the template. There must be one more N+1 that is being executed the first time that template renders, with the 178 queries. Do you know what query or queries are contributing to that?

TheTrio:
The queries are for `tags`, `campaigns` and `courses` - all of which are based on the `course_id`. They probably can be reduced to fewer queries but that would increase the complexity of the code quite a bit and so I plan on going down that path only when there isn't anything else obvious left to optimize.

PR #5275, opened 2023-03-06
Replace ajax with fetch

Aminehassou:
With reference to #2015 
## What this PR does
Replaces the remaining ajax calls in the codebase with fetch API. Most of the replaced ajax calls here are related to the survey system. Specifically `Survey.js` and `SurveyAdmin.js`

There's one fetch call that is causing odd behaviour but doesn't seem to affect user experience. 

When submitting a survey as a student, a POST call is made to `/surveys/rapidfire/question_groups/[id]/answer_groups`. This call returns status code `302 found.` This causes the POST call to be redirected to `surveys/rapidfire/question_groups` which returns status code `401 unauthorized`. However, the actual survey is still submitted and recorded, with the UI still behaving normally.

See the screenshots below for more details.

**Initial fetch call:**
image

**Redirected to:**
image

### DIFF
--- a/app/assets/javascripts/surveys/modules/SurveyAdmin.js
+++ b/app/assets/javascripts/surveys/modules/SurveyAdmin.js
@@ -176,12 +176,18 @@ const SurveyAdmin = {
   },
 
   getQuestion(id) {
-    return $.ajax({
-      url: `/surveys/question_group_question/${id}`,
-      method: 'get',
-      dataType: 'json',
-      contentType: 'application/json',
+    return fetch(`/surveys/question_group_question/${id}`, {
+      method: 'GET',
+      credentials: 'include',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRF-Token': SurveyAdmin.$csrf_token.content
+      },
       success: $.proxy(this, 'handleConditionalQuestionSelect')
+    }).then((response) => {
+      if (response.ok) {
+        $.proxy(this, 'handleConditionalQuestionSelect');
+      }
     });
   },
 
--- a/app/assets/javascripts/surveys/modules/SurveyAdmin.js
+++ b/app/assets/javascripts/surveys/modules/SurveyAdmin.js
@@ -176,6 +176,7 @@ const SurveyAdmin = {
   },
 
   getQuestion(id) {
+    const that = this; // This is done to store the correct "this" pointer to the object while inside the promise
     return fetch(`/surveys/question_group_question/${id}`, {
       method: 'GET',
       credentials: 'include',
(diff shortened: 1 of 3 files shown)


ragesoss:
Is this 401 only happening after the conversion to `fetch`? Is it because ajax has automatic redirect-following behavior and `fetch` does not? Can the request be fixed so that no redirect is needed?

Aminehassou:
Yes, with ajax the 401 does not happen, I can add `redirect: 'manual'` to fetch and that could possibly stop the fetch from auto-redirecting. I'll look into it.

Aminehassou:
Ok I added `redirect: 'manual'`. Now it cancels the redirect and shows this in devtools:
image
The survey submission is recorded fine and there's no unnecessary unauthorized call to the other route.

ragesoss:
I'd really like to know what is triggering this redirect the first place and what purpose it was supposed to serve, before I'll feel safe merging it.

Aminehassou:
> I'd really like to know what is triggering this redirect the first place and what purpose it was supposed to serve, before I'll feel safe merging it.

It looks like the issue was related to the rapidfire survey controller returning an html view instead of the json view when using fetch. This was happening because the fetch request header was missing `Accept: 'application/json'`. After adding that into the headers. The redirect no longer happens.
image

ragesoss:
Nice! I suggest opening a separate PR for the controller fix, and then cleaning up this one to be as concise as possible.

app/assets/javascripts/surveys/modules/SurveyAdmin.js
```diff
@@ -172,12 +176,18 @@ const SurveyAdmin = {
...
+      method: 'GET',
+      credentials: 'include',
+      headers: {
+        'Content-Type': 'application/json',
+        'X-CSRF-Token': SurveyAdmin.$csrf_token.content
+      },
       success: $.proxy(this, 'handleConditionalQuestionSelect')
+    }).then((response) => {
+      if (response.ok) {
+        $.proxy(this, 'handleConditionalQuestionSelect');
```
ragesoss, line 189: Does this still work even though the initial request isn't using jquery anymore?

Aminehassou: You're right, this was not working, I pushed a commit that fixes this (also replaced the use of $.proxy here since it's deprecated)

ragesoss:
@Aminehassou can you document what you've manually tested, in terms of the requests being modified here? Code-wise, it looks good to me currently.

Aminehassou:
Sure, here's each request that was modified and the manual tests I did @ragesoss 

`Route: /surveys/question_position`
`Method: PUT`
I tested this by creating a bunch of sample question groups, then reordering the questions, and making sure everything saves properly

`Route: /surveys/update_question_group_position`
`Method: POST`
I tested this by creating a bunch of sample surveys, and updating the question group positions related to them.

`Route: /surveys/question_group_question/{id}`
`Method: GET`
I tested this by attempting to create conditional questions and making sure they behaved properly

`Route: /survey_notification`
`Method: PUT`
I tested this by answering surveys as an instructor and surveys as a student

_This route is passed in through form data, so it's not explicitly written in the code_
`Route: /surveys/rapidfire/question_groups/{id}/answer_groups`
`Method: POST`
Tested submitting surveys and made sure the survey data actually gets recorded

I also made sure to test the calls before converting to fetch and fter, just to make sure there isn't any odd behavior.

PR #5276, opened 2023-03-06
Clone available Articles alongside course 

### LINKED ISSUE #4331: When cloning a course, instructor should have the option to copy Available Articles list
### Current Behavior:
Cloning a course creates a copy of an old course's timeline, but with no Available Articles.

### Desired Behavior:
Some instructors want to re-use the set of available articles from a previous course. When they clone a course, they should have an option to copy over the available articles as part of the cloning process.

### Related Files & Code:

* Backend: `course_clone_controller.rb` and  `course_clone_manager.rb`
* Frontend: `course_clone_button.jsx` and/or `course_cloned_modal.jsx` and/or `reuse_existing_course.jsx`

kunalthedev:
#Fixes #4331 
## What this PR does
When an instructor tries to clone a course then, he should be displayed an option to copy over the available articles as part of the cloning process.

## ScreenRecordings
[Screencast from 09-03-23 01:43:20 AM IST.webm](https://user-images.githubusercontent.com/72390769/223839086-a52170f4-0a28-484b-b963-b50eb803f248.webm)


ragesoss:
This looks like it works along the lines of what I was imagining. Nice!

I think it would be better if the checkbox appeared below the dropdown instead of pushing the button around when it appears/disappears.

kunalthedev:
> This looks like it works along the lines of what I was imagining. Nice!
> 
> I think it would be better if the checkbox appeared below the dropdown instead of pushing the button around when it appears/disappears.

See i've updated the video

ragesoss:
Can you get the buttons to remain in place, instead of shift vertically?

kunalthedev:
> Can you get the buttons to remain in place, instead of shift vertically?
In that case the height of the button would be increased

kunalthedev:
> Can you get the buttons to remain in place, instead of shift vertically?

In that case height of the button would be increased in accordance with the height of flexbox

spec/features/course_cloning_spec.rb
```diff
@@ -118,13 +118,14 @@
   end
 
   it 'copies relevant attributes of an existing course with assignments' do
+    pending 'This sometimes fails for unknown reasons.'
```
ragesoss, line 121: I'd like to avoid this if at all possible. How often does it fail, and what is the reason?

ragesoss:
I think it would look a little better with just a bit of extra space between the course select and the checkbox.

spec/features/course_cloning_spec.rb
kunalthedev: So, I've tried running this test 8-10 times and it worked fine so i am pushing it removing pending mark. But my last pr failed because of this only so i am not sure about this. And i am attaching error log too
Screenshot from 2023-03-08 23-39-38
So, what happens is that when you select a course to clone from the dropdown, and click on "clone this course" button, it sometimes show some other thing rather than showing "Update Details for cloned Course"

lib/course_clone_manager.rb
```diff
@@ -11,7 +11,26 @@ def initialize(course, user, campaign_slug=nil)
...
     @clone = @course.dup
+    course_clone
+    return @clone
+  # If a course with the new slug already exists — an incomplete clone of the
+  # same course — then return the previously-created clone.
+  rescue ActiveRecord::RecordNotUnique
+    return Course.find_by(slug: @clone.slug)
+  end
 
+  def clone_with_assignment!
```
ragesoss, line 22: Similarly, instead of a new method that basically duplicates the existing one, I think this would be best implemented by extending the existing methods. In this case, you could add a parameter to `clone!` for when the assignments should be copied, and make it default to the current behavior.

app/assets/javascripts/selectors/index.js
```diff
@@ -138,6 +139,12 @@ export const getCloneableCourses = createSelector(
   }
 );
 
+export const getCoursesWithoutUsers = createSelector(
```
ragesoss, line 142: This name does not make sense.

app/assets/javascripts/utils/api.js
```diff
@@ -179,6 +179,21 @@ const API = {
     return response.json();
   },
 
+  async cloneCourseWithAssignment(id, campaign) {
```
ragesoss, line 182: Again, I think there's no need to create a new method; the old one should be able accommodate both behaviors.

spec/features/course_cloning_spec.rb
ragesoss: This kind of error suggests that the page is in a different state based perhaps on the window size in the test environment, or similar. The 'found 1 time non-visible' suggests that the expected text is getting covered up by something, either because of the window size or because of an open dropdown menu, notification message, or similar.

kunalthedev: > This kind of error suggests that the page is in a different state based perhaps on the window size in the test environment, or similar. The 'found 1 time non-visible' suggests that the expected text is getting covered up by something, either because of the window size or because of an open dropdown menu, notification message, or similar.

Sometimes it's even 'not visible' like i am updating screenshot below
Screenshot from 2023-03-09 00-36-14
Although after my recent commit it didn't threw any error

kunalthedev: I removed the translation thing from feature spec though because i thought it's useless to check for that text. as we are already checking for checkbox.

kunalthedev: See @ragesoss it failed again due to that same error. and both of the test cases of this file are failing sometime that's why the other case is pending i guess. Although i did ran the same test around 10times in my local and it was working fine. But it just fails sometimes

kunalthedev: if you rerun the CI then it might pass too

app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -310,11 +323,18 @@ const CourseCreator = createReactClass({
...
-    const options = this.props.cloneableCourses.map((course, i) => <option key={i} data-id-key={course.id}>{course.title}</option>);
+    const options = this.props.cloneableCourses.map((course, i) => <option key={i} data-id-key={course.id} value={course.slug}>{course.title}</option>);
     const selectClassName = `select-container ${selectClass}`;
     const eventFormClass = this.state.showEventDates ? '' : 'hidden';
     const eventClass = `${eventFormClass}`;
-    const reuseCourseSelect = <select id="reuse-existing-course-select" ref={(dropdown) => { this.courseSelect = dropdown; }}>{options}</select>;
+    const reuseCourseSelect = <select id="reuse-existing-course-select" ref={(dropdown) => { this.courseSelect = dropdown; }} onChange={this.change}>{options}</select>;
+
+    let showCheckbox;
+    if (this.props.assignmentsWithoutUsers.length > 0) {
```
ragesoss, line 333: Is `this.props.assignmentsWithoutUsers` guaranteed to always be set, so that calling `.length` on it will never throw an error?

app/assets/javascripts/selectors/index.js
```diff
@@ -138,6 +139,12 @@ export const getCloneableCourses = createSelector(
   }
 );
 
+export const getAvailableArticles = createSelector(
```
ragesoss, line 142: A comment explaining the definition of Available Articles would be useful here.

app/controllers/course_clone_controller.rb
```diff
@@ -29,4 +31,8 @@ def check_permission
   def clone_params
     params.permit(:campaign_slug)
   end
+
+  def copy_assignments
```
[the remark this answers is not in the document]
kunalthedev: is assignment_params a good name?

lib/course_clone_manager.rb
```diff
@@ -3,10 +3,11 @@
 
 #= Procedures for creating a duplicate of an existing course for reuse
 class CourseCloneManager
-  def initialize(course, user, campaign_slug=nil)
+  def initialize(course, user, clone_assignments, campaign_slug=nil)
```
ragesoss, line 6: I think this would benefit from converting the positional arguments to named arguments. That will make it more clear what all the parameters are, whenever this method is called.

app/controllers/course_clone_controller.rb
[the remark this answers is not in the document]
ragesoss: I don't think that's explicit enough. Better to use a longer name that indicates exactly what the method is for.

app/assets/javascripts/components/course_creator/course_creator.jsx
kunalthedev: Just to make sure it never throws an error I've added a condition to only check `.length` if `this.props.assignmentsWithoutUsers` is present

ragesoss: if there's no logical way for it to be unset, I don't think that check should be added. As along as the initial state will render without depending on a successful fetch of assignment data first, it should be fine.

app/assets/javascripts/components/course_creator/course_creator.jsx
```diff
@@ -66,6 +68,13 @@ const CourseCreator = createReactClass({
       this.props.fetchCampaign(campaignParam);
     }
     this.props.fetchCoursesForUser(window.currentUser.id);
+    },
+
+  setCopyCourseAssignments(e) {
+    const scoped = e.target.checked;
```
ragesoss, line 74: Naming this constant doesn't make the code any more readable. I suggest just using `e.target.checked` directly in the setState call.

app/assets/javascripts/components/course_creator/course_creator.jsx
kunalthedev: But for just a case API crashes it's better to have that check there.

app/controllers/course_clone_controller.rb
```diff
@@ -29,4 +31,8 @@ def check_permission
   def clone_params
     params.permit(:campaign_slug)
   end
+
+  def clone_assignments_params
+    params.permit(:copy_assignments)
```
ragesoss, line 36: Since this is also a clone param, and you're digging into it for the `:copy_assignments` value anyway, I think this should be done by extending `clone_params` to permit the additional value, rather than using a separate method.

app/assets/javascripts/components/course_creator/course_creator.jsx
ragesoss: Okay. A better solution than adding the check here will be to make sure the initial state is set to an empty array or similar.

kunalthedev: So that is already there in global state, so i can remoive that extra check and it should not produce any error

kunalthedev: > So i just realised that empty array is already there in global state, so i can remoive that extra check and it should not produce any error

PR #5289, opened 2023-03-11
Added area-label to Button without text to make it readable by screen readers

### LINKED ISSUE #2531: Label remove button for students
image

The remove button (-) for students is unlabeled and therefore inaccessible by screen readers. Look for `removeButton` in AssignButton, EnrollButton, TagEditable, and CampaignEditable and add functionality so that these buttons describe what they do.

kunalthedev:
## What this PR does
Fixes #2531

So, from a [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) doc i found that button label can be accessed via screenreader using `area-label` attribute of button.

### DIFF
--- a/app/assets/javascripts/components/common/enroll_button.jsx
+++ b/app/assets/javascripts/components/common/enroll_button.jsx
@@ -121,7 +121,7 @@ const EnrollButton = createReactClass({
       let removeButton;
       if (this.props.role !== 1 || this.props.users.length >= 2 || this.props.current_user.admin) {
         removeButton = (
-          <button className="button border plus" onClick={this.unenroll.bind(this, user.id)}>-</button>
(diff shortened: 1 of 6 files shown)


ragesoss:
The idea behind this issue is not to change the visual UI, but instead to add a label so that the button is accessible via screen reader.

kunalthedev:
> The idea behind this issue is not to change the visual UI, but instead to add a label so that the button is accessible via screen reader.

Ohh so you mean to add area-label attribute? ?

ragesoss:
Something like that.

kunalthedev:
> Per discussion, the visual UI should not be changed.

Yes I'll update it soon

Issue #1038, opened 2016-11-23
Replace all javascript confirm and alert instances with Confirm and Notification components


ragesoss:
We have several places in the React frontend that use the native browser JavaScript functions `confirm()` and `alert()`. These don't match the styling of the rest of the dashboard, and the confirm dialog can also cause problems related to browsers (ie, Firefox) that detect multiple uses of it in a row and provide the option to "don't show anymore"... which if clicked, will make it impossible to confirm any more actions. (One example of this is removing Available Articles. See https://phabricator.wikimedia.org/T292078 )

Instead of the native functions, we can use the `initiateConfirm` action for confirmation dialogs, and the `addNotification` action for alerts.

lopezabegaillyn:
Hi @ragesoss , can you explain this further? Thank you so much. :)

Sahethi:
Hello! While working on this issue I think addNotification is not doing the same job as alert cause I ran into this problem. 
So when trying to delete a campaign, the prompt asks you to enter the title and will give you an alert in case of an empty answer.  Something like this 
<img width="1440" alt="Screenshot 2022-01-18 at 9 02 07 PM" src="https://user-images.githubusercontent.com/40771676/149967672-8877ea7e-d6dd-48cf-8bc4-b4143c470cad.png">
<img width="1440" alt="Screenshot 2022-01-18 at 9 02 14 PM" src="https://user-images.githubusercontent.com/40771676/149967690-2baa27c2-6837-40f5-8754-3506c5dc24bd.png">


But on replacing **alert** with **addNotification** in **app/assets/javascripts/campaigns.js** Line No. 8
The website seems to not check whether the prompt was empty and deletes the campaign anyways. 

Is this an issue that happens on replacing the native functions?

ragesoss:
The ones that require entering text probably can't be replaced without some changes to existing components or creation a of new component.

I suggest starting with other alerts/confirms if there are any.

Sahethi:
Okay I will look into that, thank you!

PR #5292, opened 2023-03-12
Fix issue 1038

lavish123409:
With Reference to https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1038

## What this PR does
This PR replaces the javascript confirm() instance with Confirm component, that is, initiateConfirm() function in the block.jsx file.

## Screenshots
Before:
image

After:
image


ragesoss:
You'll need to update the timeline editing spec to work with the new UI. It's failing because it's written specifically to interact with the native confirm currently.

ragesoss:
Looks good to me!

PR #5302, opened 2023-03-13
Removed unwanted dependencies

kunalthedev:
## What this PR does
Removing dependencies which are no longer used in project.
This removes `mkdirp` , `jsx-require-extension` and `run-sequence`

`mkdirp` was added 7 years ago by @ragesoss and it was mainly used in `.coffee` files, since we are using ReactJS now, so we no longer need this [commit](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/1bbdacd32d44239c3bc1b37a8cededa8997446a9)

`jsx-require-extension` was also added 7 years ago by @ragesoss and was used in testing of `.coffee` files [commit](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/37d14ea729be1f8ca05d52556ea9299338a161ee)

`run-sequence` was added 9 years ago  and was used i`.coffee` files [commit](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/d3bfec332de3948bbc155b17e39b7a1c776b1a7c)


ragesoss:
Ideally, explanations like this should be in the commit messages... the more detail included, the easier it is to understand a change when browsing code with a git-enabled code editor.

kunalthedev:
> Ideally, explanations like this should be in the commit messages... the more detail included, the easier it is to understand a change when browsing code with a git-enabled code editor.

So, what should I modify now. Was it a thing to note for future PR?

ragesoss:
It's possible to update commit messages (using `git commit --amend`), or you can just take that advice for future work.

kunalthedev:
> It's possible to update commit messages (using `git commit --amend`), or you can just take that advice for future work.

Ohh, Thanks i just updated it using that.

Issue #5297, opened 2023-03-13
Many rubocop linting rules are disabled. Choose one, enable it, and fix all violations


ragesoss:
The `.rubocop.yml` includes a long list of 'Temporary exceptions'. Each of this is a style rule that we've disabled, because when it appeared in a new release of Rubocop, the Dashboard codebase had one or more violations, and no attempt has yet been made to evaluate that rule and either fix existing violations or make an explicit decision that we don't want to use that rule.

Choose a single rule, enable it by deleting it's entry `.rubocop.yml`, run `bundle exec rubocop` to identify all violations, and then either fix those violations or write up an explanation of why the rule should become a permanent exception.

sharon-odhiambo:
HI @ragesoss , I'd like to work on this issue. KIndly assign it to me.

sharon-odhiambo:
Also, is one restricting to fixing one linting rule or I can do multiple?

ragesoss:
@sharon-odhiambo we don't assign issues, but you're welcome to work on this. Please do one rule at a time (in separate PRs), but if you think it's a useful learning exercise for you to more than one, you're welcome to do so.

sharon-odhiambo:
HI @ragesoss . Understood and thanks.

sharon-odhiambo:
Hi, I'll be working on RSpec/BeNil and Rails/I18nLocaleTexts.

kessio:
hi, @ragesoss I have noticed there are some rubocop rules once enabled, I can use autocorrect to correct the files. Can I enable that, and autocorrect the files with errors?

ragesoss:
(Copying over from chat in case it's helpful for someone else...)

Yes, but please 1) inspect all the changes to make sure they make sense, and 2) only do this for one rule at a time. Be sure to explain as much as possible about the purpose of the rule in your commit message, and explain why you think it's good for it to be enabled rather than permanently disabled in your PR description.

akanshajais:
Hey @ragesoss , I am akansha jaiswal an outreachy applicant I am commenting to record my contribution for outreachy. I read about problem and I found I would choose 'Style/GuardClause' rule to identify violations as GuardClause returns early from a method if the condition is met which helps to reduce the level of nesting and can make code easier to read but in some cases, using a guardclause can make the code less readablet too , Particularly if the condition that triggers the general clause is complex or if the method has a lot of branching logic. 

[...]

PR #5308, opened 2023-03-15
Enable Rails/I18nLocaleTexts and translate the admin page flash notice

sharon-odhiambo:

## What this PR does
In this PR, I enabled the Rails/I18nLocaleTexts linting rule to disallow untranslated versions of flash notices. On this I updated the "You don't have access to that page" translation to different languages on the admin page to allow all users understand the warning and proceed with the necessary authentication and authorization
It's in line with this issue  - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5297

## Screenshots
Before:
Screenshot from 2023-03-15 16-42-38

After:
Screenshot from 2023-03-15 16-43-18

 I'm wondering whether we can have a gem that utilizes translatewiki.net to automatically translate the remaining flash notices to the other languages


ragesoss:
This is definitely a rule we want to add to the permanent exceptions; we exclusively use human translations from translatewiki.net, and as long as every message is in the en.yml, there will be fallback messages for any languages that have some missing translations.

sharon-odhiambo:
Okay, lemme do that.

ragesoss:
This PR should only disable that rule. I don't want to include any of the machine translations.

sharon-odhiambo:
Okay let me revert the commit. Thank you for the feedback.

Issue #5316, opened 2023-03-19
Courses > Institution > Program institution name overflows when too long


Aminehassou:
### What is happening?
If the institution name is too long. It overflows in multiple areas on the `/courses/[institution]/[program]` program page

### To Reproduce

1. Create an institution with a very long name
2. Access a program page that is being run by that institution

### Expected behavior
The institution name wraps correctly and doesn't go past its container

### Screenshots
![image](https://user-images.githubusercontent.com/39999898/226148794-7cc80c9c-7df2-4895-8cb7-291ee6d8bfa0.png)

### System Information

 - OS: Ubuntu 22.04.1 LTS
 - Browser: Google Chrome 110.0.5481.77 (Official Build) (64-bit)

PR #5317, opened 2023-03-19
Fixed institution and course text overflow issue

Aminehassou:
With reference to #5316
## What this PR does
This PR fixes the institution and course text overflow issues by replacing the underscores with spaces
## Screenshots
Before:
image

After:
image


Aminehassou:
> I think these things ought to be fixed with CSS. The approach of replacing underscores is a very narrow one that could result in same issue for long words that don't include underscores, and it also adds complexity to the JS which I'd like to avoid as much as possible.

I've fixed the issue using CSS instead of JS. Here's what the page looks like now with the new css fix:
image

Issue #5034, opened 2022-07-07
Detect and surface large pageview spikes


ragesoss:
Dashboard users — instructors, editors, Wiki Education staff — like to know when content they worked on becomes suddenly very relevant to the public and results in large spikes in the number of pageviews. These typically represent success stories, where users created or improved content that then became highly relevant to the public (eg, because the topic become relevant to an important current event).

The Dashboard should be able to detect and log this kind of pageview spike.

The place to do this would probably be during the import of pageviews to update an Article's average pageviews, since daily pageviews over a moderately long period of time are fetched (and then averaged) during this process. It might make sense to do this after addressing https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/4370 since a good solution to that will probably involve extending the period over which pageviews are fetched.

vaidehi44:
@ragesoss, since you suggested to leave the issue referenced here (#4370), should I also drop this one?

ragesoss:
Hmm... I'm not sure. If you can think of an efficient way to do this one without relying on #4370, I'm open to it.

vaidehi44:
Okk ... I guess here too best way would be to introduce `earliest_edit` and `avgerage_page_views` in `ArticlesCourses`. Because will have to keep a track of average views to compare it when the count rises suddenly. What do you think?

vaidehi44:
And, also what should be the threshold or the criteria to decide whether there has been significant rise in views? Like, if the views increase 5 fold or 10 fold ... anything like that?

ragesoss:
Yes, I think that approach would work well, putting both of those values into ArticlesCourses. That will make it easy to if any single day is many times above the average, during the process for updating that average.

I think 5-fold is a good starting point... maybe 5-fold increase and minimum spike of 100 views in a day. Those can be tweaked, depending on how often we find it happening.

vaidehi44:
@ragesoss, I am little confused in where the code for updating the `earliest_edit` and `average_page_views` fields will fall. I think, for `earliest_edit` there should be an instance method which would be called from `update_cache` method of `ArticlesCourses`.

And, what I get is that the cache of `ArticlesCourses` is being updated from `UpdateCourseStats` which falls under the `constant_update` category through `ScheduleCourseUpdatesWorker`. So, I guess it is being updated every 5 minutes.

[...]

PR #5318, opened 2023-03-19
Check and notify the spikes in pageviews of articles courses

vaidehi44:
## What this PR does
It is a PR for issue #5034 

As discussed, new fields - `earliest_edit`, `average_pageviews`, `views_updated_at` have been added to the `ArticlesCourses`. `average_pageviews` would get updated weekly and, the current criteria for detecting the spike is - 5 fold increase in the average value and, minimum of 100 views in a day.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -134,7 +134,7 @@ driver_hosts = (
 VCR.configure { |config| config.ignore_hosts(*driver_hosts) }
 
 VCR.configure do |c|
-  c.allow_http_connections_when_no_cassette = false
+  c.allow_http_connections_when_no_cassette = true
   c.cassette_library_dir = 'fixtures/vcr_cassettes'
   c.hook_into :webmock # or :fakeweb
   c.default_cassette_options = { record: :new_episodes }
--- a/app/mailers/pageview_spike_mailer.rb
+++ b/app/mailers/pageview_spike_mailer.rb
@@ -1,0 +1,18 @@
+# frozen_string_literal: true
+
+class PageviewSpikeMailer < ApplicationMailer
+	def self.send_spike_alert_email(article_course)
+		return unless Features.email?
+		wiki_ed_staff = article_course.course.courses_users.where(role: CoursesUsers::Roles::WIKI_ED_STAFF_ROLE)
+		email(article_course, wiki_ed_staff).deliver_now
+	end
+
+	def email(article_course, staff)
+		@course = article_course.course
+		@article = article_course.article
+		@course_link = "https://#{ENV['dashboard_url']}/courses/#{@course.slug}"
+		@articles_link = "#{@course_link}/articles"
+		mail(to: staff.map(&:email),
+		     subject: "Notable spike in the pageviews count of article.")
+	end
+end
--- a/app/mailers/pageview_spike_mailer.rb
+++ b/app/mailers/pageview_spike_mailer.rb
@@ -1,18 +1,21 @@
 # frozen_string_literal: true
 
 class PageviewSpikeMailer < ApplicationMailer
-	def self.send_spike_alert_email(article_course)
-		return unless Features.email?
-		wiki_ed_staff = article_course.course.courses_users.where(role: CoursesUsers::Roles::WIKI_ED_STAFF_ROLE)
-		email(article_course, wiki_ed_staff).deliver_now
-	end
+  def self.send_spike_alert_email(article_course)
+    return unless Features.email?
+    wiki_ed_staff = article_course
+                    .course
+                    .courses_users
+                    .where(role: CoursesUsers::Roles::WIKI_ED_STAFF_ROLE)
+    email(article_course, wiki_ed_staff).deliver_now
+  end
 
-	def email(article_course, staff)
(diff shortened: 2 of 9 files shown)


vaidehi44:
@ragesoss, I have implemented the basic methods which would update the new fields and check the spike in views. Haven't implemented the alert mail part yet. Let me know, if this looks good.

vaidehi44:
@ragesoss, the failing tests are due to some http requests which VCR is unable to handle. But, when I checked one of the GET requests in browser, it worked and returned the response. What could be the issue here?

ragesoss:
I'm not sure. When I saw them fail the first time, I assumed that the pageview API was just down at the time the tests ran, but it failed a second time, so I'm not sure...

One possibility is that, with your changes, the test suite now triggers rate limiting from the Wikimedia pageviews API.

vaidehi44:
Ohkk, what do you mean by rate limiting? And, is the current approach fine, or should I make any changes?

ragesoss:
I haven't reviewed it yet, but I will do so as as soon as I can, likely tomorrow.

Rate limiting means that the API only allows a certain number of requests from a single IP over a certain time period. Typically, once that rate is exceeded, the server will return the '429' response code indicating 'too many requests'. I'm not sure if that's what is going on, but it's one possibility... especially if your new tests pull significantly more data from the pageviews API than the test suite did previously.

ragesoss:
I think this approach makes sense, but I'm pretty sure the tests are indeed failing because of rate limiting. With this change, every test that involved a course update now also involves pulling pageview data. I'm also concerned about performance impacts, and I would want to limit this new functionality to Wiki Education mode at least at first.

I suggest keeping the pageview updates separate from the rest of the `update_cache` procedure. That will make it easier to control when to pull in pageview data, in both the test environment and production. Perhaps UpdateCourseStats is the place to explicitly invoke the pageviews update for the ArticlesCourses records.

vaidehi44:
Okk...got it.

vaidehi44:
@ragesoss, seems like the failing tests were not due to rate limiting. I changed a setting in VCR configuration - made `allow_http_connections_when_no_cassette` true. I guess this was the reason it was failing, because the average views request for pageviews was made, but (may be) it was never invoked through a `VCR.use_cassette` command.
So, with this change now only two tests are failing... which most probably should be due to some other reason.

vaidehi44:
Also, I realized one more thing which could make the computation for pageviews much optimal. 
In the new approach, since we are already storing `earliest_edit`, `views_updated_at` and `average_pageviews`, we won't be required to fetch the entire pageviews data for a `article_course` since the earliest_edit. We could do this -

1. Get total pageviews till the time it was last updated by doing this - `average_pageviews`*(`views_updated_at` - `earliest_edit`).
2. Get the pageviews count for last 7 days by fetching the data with start_date = `views_updated_at`, and end_date = `current_date`
3. Get total days by (`current_date` - `earliest_edit`)
4. Get the average pageviews by adding 1 and 2 and then dividing with total days from 3.

This way, we would need to just fetch the pageviews data for last 7 days for any `article_course`.

ragesoss:
That sounds like a reasonable optimization. Do you know how much of a performance impact it makes, roughly? I don't know how much the pageview API cares about the size of the date range, in terms of response times.

spec/spec_helper.rb
```diff
@@ -134,7 +134,7 @@
 VCR.configure { |config| config.ignore_hosts(*driver_hosts) }
 
 VCR.configure do |c|
-  c.allow_http_connections_when_no_cassette = false
+  c.allow_http_connections_when_no_cassette = true
```
ragesoss, line 137: Okay! So, this means that the failures before were just from the Pageviews fetching code eating the errors that VCR threw. We don't want to keep this setting enabled, though. Using mocks might still be a good strategy for the tests, or perhaps we can work around this in the test environment by making the pageview fetching behavior dependent on a setting, so that it skips doing it by default when updating ArticlesCourses records.

ragesoss:
The remaining failures (articles_courses_spec)  are just from tests that hard-code arbitrary pageview values that the test doesn't expect to be updated.

ragesoss:
I think just running it in `since earliest edit` mode the first time a given record needs to be updated is fine.

vaidehi44:
@ragesoss I have pushed the code of new alert mailer. The preview looks like this. I have used some part from the `active_course_mailer`.

vaidehi44:
> I suggest keeping the pageview updates separate from the rest of the `update_cache` procedure. That will make it easier to control when to pull in pageview data, in both the test environment and production. Perhaps UpdateCourseStats is the place to explicitly invoke the pageviews update for the ArticlesCourses records.

But still the pageview data fetching code is in `update_cache` of articles_courses. (I think, you had given the above suggestion because we thought tests were failing due to rate limiting. )Should I move it to seperate method in `UpdateCourseStats`?

ragesoss:
@vaidehi44 sorry to take so long to get back to you. Yes, I think moving the new calculations to a separate method would be best in UpdateCourseStats would be best.

vaidehi44:
No issues @ragesoss :). Will make the changes soon.

Issue #5293, opened 2023-03-13
There is no easy way to navigate to an individual FAQ question


ragesoss:
The FAQ system will show a set of search results when you do a search from the nav bar. However, it's a little tedious if you want to first search, and then get the URL for an individual entry that shows up in the results.

There should be some kind of link from each search result to the individual entry.

saha23s:
Hi @ragesoss I want to work on this issue - Can you guide me a little more about what skills should be necessary for fixing this? Currently, my faq page looks like this. 
![image](https://user-images.githubusercontent.com/93749279/226130440-28cf9797-38d4-4594-a219-bea1be2f3cc9.png)

ragesoss:
You should start by adding some FAQ entries. You can do this from the `/admin` page (which you need to log in and give your account admin permissions to access... see `docs/user_roles.md`).

saha23s:
@ragesoss I could successfully set up the entries. First of all, I am curious why there's no nav item called "help" or "FAQs" in the navbar where I can click and see all the FAQs. Because as a user I was confused about how to access faqs.
Besides, can you clarify the expected behavior when you mentioned "There should be some kind of link from each search result to the individual entry"
Also, I want to continue working on this issue!

Titilola2:
@ragesoss , I'd like to know if I can work on this big or only one person is allowed to work on it at a time?

saha23s:
@ragesoss If I understand your explanation correctly, then it means I will aim to create an individual page for every FAQ entry when a user clicks on a FAQ they are interested in. Right now, toggling is the way to see the details. Let me know if I am thinking it right.
![image](https://user-images.githubusercontent.com/93749279/227572915-b87babdd-b37d-489b-8e30-6c1b12652813.png)

ragesoss:
@saha23s the individual pages already exist, they just aren't easy to navigate to. As an admin, the way I currently do it is to use the `edit` link next to the title of an entry, and then remove `/edit` from the end of the url.

One possible way to do this would be to simply turn the title itself into a link. Another way would be to add a 'permalink' link or similar, next to each one.

PR #5327, opened 2023-03-27
AddingFAQLinkToTitle

saha23s:
Issue: https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5293 

## What this PR does
It turns the FAQ title into a link so that users can access the individual FAQ page by clicking on it.
It also removes the accordion behavior while clicking on the title.

Before:

image

After:

image
image

### DIFF
--- a/app/views/faq/_faq.html.haml
+++ b/app/views/faq/_faq.html.haml
@@ -1,8 +1,10 @@
 %div{class: "faq #{expanded ? 'expanded' : 'collapsed'}", id: "faq-#{faq.id}"}
   .faq-header{onclick: "toggleAccordian('faq-#{faq.id}')"}
-    %h2.faq-title= link_to faq.title, faq_path(faq.id), onclick: "event.stopPropagation();"
-    - if current_user&.admin?
-      .edit-link= link_to 'edit', edit_faq_path(faq)
+    - if params[:action] == 'show' # check if current page is the show page
+      %h2.faq-title= faq.title
+    - else
+      %h2.faq-title= link_to faq.title, faq_path(faq.id), onclick: "event.stopPropagation();"
+    .edit-link= link_to 'edit', edit_faq_path(faq)
     .icon.icon-arrow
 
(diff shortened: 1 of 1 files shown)


ragesoss:
Thanks! Can you add before/after screenshots to the PR description?

saha23s:
Hi, @ragesoss I just added them!

ragesoss:
Thanks! This works.

One thing I noticed is that when you click one of these title links from a list of multiple questions, clicking the title activates the javascript to expand the answer accordion, so you briefly see the expanded version on the same page before the new page loads. If you can find a way to stop that from happening, that would be a nice enhancement.

saha23s:
@ragesoss  I have removed toggling functionality from both the faq and show pages. But can you confirm if the individual FAQ page is looking fine? Because the edit link for the admin in the individual FAQ page is below the title now. Will that be a problem?

ragesoss:
The accordion behavior is still important, it just shouldn't be triggered when you click the link. Clicking elsewhere (such as the region near the accordion icon, or anywhere in that header that isn't the link itself, should still cause the accordion section to toggle open.

saha23s:
@ragesoss I have updated the PR so that the toggling functionality is not removed and the accordion behavior is not triggered while clicking on the title.

ragesoss:
That fixed the accordion issue effectively. Now I've noticed another thing that should be improved... the title link should not be there when viewing the individual FAQ (since it just links back to the same page you're already on).

saha23s:
@ragesoss I removed the link from the show page.

ragesoss:
That was quick! LGTM.

saha23s:
@ragesoss sorry! I think I mistakenly removed "if current_user&.admin?" while I added a new if else statement for the title above the edit link for the faq element which I think is important! Everyone should not access the edit link. I just noticed it. How is it possible to add that right now?

ragesoss:
Oops, I missed that! You can open a new PR to fix it.

PR #5328, opened 2023-03-27
Use mediawiki logs API for marking articles as deleted

TheTrio:
I'll update this with details later.

### DIFF
--- a/app/workers/deleted_articles_worker.rb
+++ b/app/workers/deleted_articles_worker.rb
@@ -7,7 +7,7 @@ class DeletedArticlesWorker
     last_run = Setting.find_or_create_by(key: 'deleted_articles')
     last_timestamp = last_run.value['timestamp'] || 1.week.ago.to_i
     Wiki.all.each do |wiki|
-      articles = Article.where(wiki:, deleted: false)
+      articles = Article.where(wiki:)
       wiki_api = WikiApi.new(wiki)
       deleted_articles_ids = {}
       restored_articles_ids = {}
--- a/app/workers/deleted_articles_worker.rb
+++ b/app/workers/deleted_articles_worker.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+class DeletedArticlesWorker
+  include Sidekiq::Worker
+  sidekiq_options lock: :until_executed
+
+  def perform
+    last_run = Setting.find_or_create_by(key: 'deleted_articles')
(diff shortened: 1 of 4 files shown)


lib/article_status_manager.rb
```diff
@@ -48,11 +48,7 @@ def update_status(articles)
     # This happens in situations such as history merges.
     # If articles move in between title/namespace updates and mw_page_id updates,
     # then it's possible to end up with inconsistent data.
-    update_article_ids @deleted_page_ids
```
ragesoss, line 51: I forget... what's the context for the PR? Note that this file does a whole lot of fiddly things to ensure as consistent state as possible between the Dashboard and Wikipedia, including handling for deletion, undeletion, and ID changes on the Wikipedia side (which can happen in various situations that we don't try to model on the Dashboard side, such as when the history of two pages get merged into one article).

TheTrio: Yes, for now I have simply commented that out but I plan on calling this function from my worker as well.

The context is a `FIXME` which says that instead of relying on Replica, an article should be marked deleted if there's a deletion log.

ragesoss: cool. this general approach looks good.

app/workers/deleted_articles_worker.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+class DeletedArticlesWorker
+  include Sidekiq::Worker
+  sidekiq_options lock: :until_executed
+
+  def perform
```
ragesoss, line 6: I prefer to keep workers very small, and keep all the business logic in a separate file.

lib/wiki_api.rb
```diff
@@ -77,6 +77,34 @@ def get_article_rating(titles)
     ArticleRatingExtractor.new(pages).ratings
   end
 
+  def deleted_logs(last_run_timestamp)
```
ragesoss, line 80: I think this should live outside of WikiApi. It might be a good opportunity to refactor it so that doing a query that uses `continue` doesn't require calling a private method directly.

TheTrio: So I should make `fetch_all` public and move these methods to a different(new?) file?

TheTrio:
So it turns out, fixing this involves a lot more effort than I anticipated and would likely break some functionality while also being much harder to test as extensively as the current implementation is right now.

Closing this for now

Issue #5298, opened 2023-03-13
Document usability problems for screen readers


ragesoss:
The Dashboard has some users — including an admin — who are blind and use screen readers.

Choose a screen reader from the following:
* VoiceOver on Mac OS
* VoiceOver on iOS
* JAWS on Windows (proprietary, costs money... but it might be possible to get a free education license)
* NVDA on Windows
* ORCA on Linux
* ChromeVox (Chrome browser plugin, cross-platform)

Use that screen reader (this may take some research and/or practice, to learn to use it), and navigate through either a Dashboard development environment or the live outreachdashboard.wmflabs.org site, and document any accessibility problems you find: pages that are harder than they should be to navigate via keyboard commands and screen reader nav tools, buttons and links that are not interpreted correctly via screen reader, and anything else that makes the Dashboard harder to use than it ought to be for screen reader users. You can either document problems in comments below, or open a new issue for each individual problem you find.

RedWing0021:
Hello,
I would like to work on this issue. Please assign this to me if no one is working on it.

ragesoss:
@RedWing0021 we don't assign issues, but you're welcome to work on this.

AkankshaAI:
@ragesoss Since no is contributing to this issue ,I would like to contribute and work on this issue

AkankshaAI:
Im learning to use it, using NVDA for windows but finding a bit of difficulty. Can you guide a bit for resources

Aminehassou:
I added a possible fix to an accessibility issue. I tested using ChromeVox so I'm unsure if it works using other screen readers

QuantumByteMaster:
@ragesoss  i would like to contribute on this issue . I will be using ORCA on linux.

crocmons:
@ragesoss I would like to contribute to this issue . I will choose ChromeVox (Chrome browser plugin, cross-platform)

Jay2006sawant:
@ragesoss I would like to contribute on this issue

Issue #2505, opened 2019-02-19
Improve accessibility for adding training modules on a block


bwreid:
Currently, when a user adds training modules to a block, they show up in a multi-select.
![image](https://user-images.githubusercontent.com/1316902/53052584-f347a700-3453-11e9-9cd8-c6ad39a10640.png)

However, the close buttons for each individual item are not accessible. We need to label these so a screen reader can read off individual items for removal.

ragesoss:
I note that the new react-select release that I just merged has several a11y fixes. Once it's deployed, we should see if this is already fixed for Helaine.

https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/2528

ragesoss:
Nope, the latest release didn't change it for Helaine.

PR #5332, opened 2023-03-30
Make remove module icon for adding training modules readable by screen readers

Aminehassou:
With reference to #2505 and partially #5298
## What this PR does
This PR adds an aria-label to the cross icon for removing existing training modules. Allowing for easy readability for screen readers (tested on ChromeVox screen reader)
## Screenshots
**Before**
image

**After**
image


Aminehassou:
The check failed and I'm not sure why, seems to be a ruby error but I only modified the javascript?

ragesoss:
Can you add before/after screenshots?

ragesoss:
I tested this locally, and it works fine, and actually looks better as well with the larger X in the 'remove' component.

Aminehassou:
Added before/after screenshots.

Issue #5320, opened 2023-03-21
Create an alert when an edited article is moved out of mainspace


ragesoss:
### Current Behavior:

During the course update process, ArticleStatusManager is responsible for updating the status of Article records associated with the course, handling cases where the article got deleted, moved to a different namespace, or was modified such that it has a new MediaWiki ID.

### Desired Behavior:

When the namespace of an article has changed from mainspace to a different namespace (typically, User or Draft), ArticleStatusManager should trigger creation of an alert (an Alert record of a new type) that notifies the Wiki Expert for the the course associated with that article.

This way, we can more quickly look into cases where student work got moved out of mainspace, as this may indicate problems that Wiki Education staff need to help with.

### Additional context:

This probably requires some refactoring of ArticleStatusManager so that the `course` record can easily be passed on along as needed, along with a new alert type (see `app/models/alert.rb` and `app/models/alert_types`) and an alert manager responsible for generating the alert.

sharon-odhiambo:
Hi, I'll be attempting on this issue.

ragesoss:
@sharon-odhiambo you're welcome to give it a shot, but note that this isn't marked as a newcomer issue. I'd like to ship this within the next week or two, so let me know how progress on it goes; if you get stuck, I'll likely implement it myself for the sake of getting it shipped.

sharon-odhiambo:
Hi @ragesoss , sure let me attempt and let you know by end of this week.

sharon-odhiambo:
Hi @ragesoss , I had promised to share my progress on this by end of this week but am just starting to understand this issue now. Kindly allow me to do an update beginning next week and you can guide on whether to proceed or pick another issue. Thank you.

sharon-odhiambo:
Hi @ragesoss. Can you help explain how I can be able to see the current records being shown by the ArticleStatusManager during the course update process? I have tried to try working on the alert but I don't know how to test it.

sharon-odhiambo:
Hi @ragesoss , I tried to work on this issue. However, I've been stuck on testing how the alerts work. I tried to see how the articles-for-deletion work to help understand this. This is my [PR-draft](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5329)
Kindly let me know whether this is leading somewhere or not. 
I'm able to create the articles and edit them but the editing doesn't show on my local environment dashboard.

sharon-odhiambo:
[...]

PR #5338, opened 2023-03-31
Create an alert when an article is moved from mainspace

sharon-odhiambo:

## What this PR does
This PR creates an alert for when an article associated with a specific course is moved from the mainspace to a different namespace like draft when the student plans to work on it or user when student is/has already worked on it.

## Screenshots
Before:

After:


ragesoss:
Nothing special needs to be done regarding frontend. Creating the Alert record is enough to make it appear in the couple of alert list views that exist in the system. As we do with some other alert types, the way to make the alert reach people is via email... in this case, probably the standard 'email wiki experts' method (I forget the actual name). So after creating it, we'd immediately call that email method. There are a number of well-factored alert manager classes that use this pattern.

sharon-odhiambo:
Okay, I've updated the view but still unable to get any alert. Could you check and help me how to proceed?

sharon-odhiambo:
Hi @ragesoss , could you kindly provide feedback on this when you can?

ragesoss:
> Hi @ragesoss , could you kindly provide feedback on this when you can?
I need some more details about what you've tried and what you're stuck on. What exactly isn't working the way you expect it to? Writing tests for it may help you pin down what's going on.

sharon-odhiambo:
So far I created the new alert for the namespace-change where I'm hoping the email_content-expert method sends an alert when a change it occurs, I added it to the alert types and is calling the new alert in the Article Status Manager. I'm hoping when a user moves an article from the mainspace an alert is sent to the expert's email. I'm logged in as the facilitator so I was hoping when I a change occurs I should receive an alert.

ragesoss:
You should be able to find the alert by going to the course page -> Activity tab > Alerts.

sharon-odhiambo:
Oh okay. Thank you! Let me check it.

app/controllers/courses_controller.rb
```diff
@@ -143,6 +143,11 @@ def suspected_plagiarism
     @show_report = current_user&.can_edit?(@course)
   end
 
+  def article_namespace_change
```
ragesoss, line 146: What is this for?

ragesoss:
I'm not sure this is the right strategy. As suggested in https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5320 , I think ArticleStatusManager should be the entry point for alert generation, as that's when the system first gets the data that shows a namespace change.

sharon-odhiambo:
Let me try to redo it in a few and you can advice on whether to continue in that direction. I think I did include it in the CourseAlertManager which is confusing even to myself.

sharon-odhiambo:
Hi @ragesoss , so I've tried to refactor this again. I've created a new alert type model and added it in the alert rb alert types. I've then created an alert in the article_namespace_change_alert_manager, with alert type and message, and then called the email content expert method. I've then called this method in the ArticleStatusManager where we are checking if a page got moved. Kindly let me know how this is.

sharon-odhiambo:
I've also been going through the messages in the slack group and getting a clearer understanding which makes me wonder whether the intended email recipient which I'm thinking is the facilitator is a content_expert in this case?

ragesoss:
> I've also been going through the messages in the slack group and getting a clearer understanding which makes me wonder whether the intended email recipient which I'm thinking is the facilitator is a content_expert in this case?

Initially, at least, the intended recipient will be the content_expert. We might expand that later on, but we want to get a clear idea of how often it happens and develop the right messages for editors, first.

lib/article_status_manager.rb
```diff
@@ -45,6 +46,7 @@ def update_status(articles)
 
     # First we find any pages that just moved, and update title and namespace.
     update_title_and_namespace @synced_articles
+    ArticleChangeNamespaceAlertManager.create_alerts_for_article_namespace_change
```
ragesoss, line 49: The idea here is to only do this for the specific articles where we needed to change the namespace. It looks like the current version of this PR doesn't account for detecting a namespace change.

sharon-odhiambo: Oh yeah. Is it advisable to do that in the alert manager or here in the ArticleStatusManager?

sharon-odhiambo: Like this recent change?

sharon-odhiambo: Oh okay. Lemme update

lib/article_status_manager.rb
```diff
@@ -110,6 +109,10 @@ def update_title_and_namespace(synced_articles)
                         deleted: false)
         # Find corresponding Assignment records and update the titles
         AssignmentUpdater.update_assignments_for_article(article)
+
```
ragesoss, line 112: Please test this locally. I don't think it will be working as written, as it does not pass the article to the manager.

Also, it will need to compare the old namespace and the new namespace (from `article` vs `article_data`), because this path also handles cases where only the title but not the namespace changed, or where the namespace changed but it was not in mainspace either before or after.

sharon-odhiambo: Uhm so ideally an old namespace belonging to the article and a new namespace after the update?
Also the build is failing due to a rubocop lint on method name which I wonder why it started failing just now but the method name hasn't changed.

sharon-odhiambo: Also how do I move an article page from mainspace to a diferent namespace on the app? So far I tried to move one but I honestly can't remember how I did it.

ragesoss: The easiest way for testing purposes is to use the rails console; you can do `article.update(namespace: 2)` or similar.

sharon-odhiambo: Oh right, thanks for this! Lemme get to it again.

PR #5340, opened 2023-04-03
Fix violations of the Rails/SkipsModelValidations cop

cadenforrest:
## What this PR does
This PR is a part of https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5297. It should fix all violations of the `Rails/SkipsModelValidations` cop, and remove the rule from the temporary exceptions list in `rubocop.yml`. We allow exceptions to this rule in-line when we are bulk updating potentially many rows on a table and don't care about skipping callbacks. 

My thought is that we should enforce the rule, but allow exceptions in-line for when we know what we are doing. It pollutes the code with `rubocop:enable`/`rubocop:disable` comments, but it helps contributors and reviewers not skip callbacks they were intending to run. How do we feel about that?

### DIFF
--- a/spec/controllers/courses_controller_spec.rb
+++ b/spec/controllers/courses_controller_spec.rb
@@ -209,7 +209,11 @@ describe CoursesController, type: :request do
     context 'setting passcode' do
       let(:course) { create(:course, slug: slug_params) }
 
-      before { course.update(passcode: nil) }
+      before do
+        # skip validations here - some course types allow nil passcodes, some dont
+        course.passcode = nil
+        course.save(validate: false)
+      end
 
       it 'sets randomly if it is nil and not in params' do
         params = { id: course.slug, course: { title: 'foo' } }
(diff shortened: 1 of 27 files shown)


spec/controllers/courses_controller_spec.rb
```diff
@@ -209,7 +209,11 @@
     context 'setting passcode' do
       let(:course) { create(:course, slug: slug_params) }
 
-      before { course.update(passcode: nil) }
+      before do
+        # skip validations here - some course types allow nil passcodes, some dont
+        course.passcode = nil
+        course.save(validate: false)
+      end
```
cadenforrest, line 216: Bumping for the reviewer to confirm this is OK 👍

ragesoss:
Hmm... not sure how I feel about it. I guess it makes the callback skipping explicit everywhere it happens, so this is probably better than just leaving things as-as.

PR #5342, opened 2023-04-03
Ugraded React-select to ^5.7.2

kunalthedev:
## What this PR does
This PR upgrades dependency [react-select](https://www.npmjs.com/package/react-select?activeTab=readme) to the latest version. To make sure it doesn't break anything i looked at the documentation and compared it with the current usage and I found out that there is no change.
Although there are a lot of components using react-select, I've uploaded few screenshots to show that everything is working fine. 

## Screenshots

Screenshot from 2023-04-04 00-42-13
image
Screenshot from 2023-04-04 00-40-50

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "react-router-dom": "^6.2.2",
     "react-router-hash-link": "^1.2.2",
     "react-scroll": "^1.8.1",
-    "react-select": "^3.1.1",
+    "react-select": "^5.7.2",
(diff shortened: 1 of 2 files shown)


kunalthedev:
> Can you include `before` screenshots as well? I want to understand any styling changes introduced by this update.

Done

ragesoss:
Cool. So it looks like there's no visible UI changes. Is that correct from what you've observed?

PR #5350, opened 2023-04-06
disable register button when user is blocked from creating an account

### LINKED ISSUE #4187: When visiting an enroll link, IP blocks that prevent account creation should be detected
### Current Behavior:
If a user's IP address is blocked, then when they begin from a Dashboard enroll link and try to create a Wikipedia account, they will see a somewhat confusing error message on Wikipedia with no obvious instructions for working around it.

### Desired Behavior:
Visiting an enroll link while logged out on the Dashboard is most often done by users who do not yet have a Wikipedia account, so this would be a good time — before sending them over to Wikipedia to create an account — to provide advice for getting an account when your current IP is blocked.

When a user visits an enroll link, the user's browser should make a query to the Wikipedia API to determine whether their IP is blocked (eg, by doing a username availability check for an arbitrary nonexistent but valid username). If they are blocked, disable the 'create account' button and provide information about working around this: by accessing the internet from a different network, or by asking the instructor for assistance.

### Related Files & Code:

`EnrollCard` and `NewAccountActions` are the most relevant places to start looking.

TheTrio:
Fixes #4187 

Looks like this

Screenshot_20230406_234003

### DIFF
--- a/app/assets/javascripts/components/enroll/new_account_button.jsx
+++ b/app/assets/javascripts/components/enroll/new_account_button.jsx
@@ -56,7 +56,7 @@ const NewAccountButton = createReactClass({
             <img className="info-img" src="/assets/images/info.svg" alt="tooltip default logo" />
             <div className="tooltip dark large">
               <p>
-                Your IP address has been blocked from creating an account. Try connecting from a different network or contact your instructor for more information.
+                {I18n.t('error.ip_blocked')}
               </p>
             </div>
           </div>
--- a/app/assets/javascripts/components/enroll/new_account_button.jsx
+++ b/app/assets/javascripts/components/enroll/new_account_button.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import createReactClass from 'create-react-class';
 import PropTypes from 'prop-types';
 import NewAccountModal from './new_account_modal.jsx';
+import { canUserCreateAccount } from '@components/util/helpers';
 
 const NewAccountButton = createReactClass({
   displayName: 'NewAccountButton',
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/enroll/new_account_button.jsx
```diff
@@ -32,9 +39,29 @@ const NewAccountButton = createReactClass({
...
+          >
+            <i className="icon icon-wiki-logo" />
+            {I18n.t('application.sign_up_extended')}
+          </a>
+          {this.state.disabled && (
+          <div className="tooltip-trigger tooltip-small">
+            <img className="info-img" src="/assets/images/info.svg" alt="tooltip default logo" />
+            <div className="tooltip dark large">
+              <p>
+                Your IP address has been blocked from creating an account. Try connecting from a different network or contact your instructor for more information.
```
ragesoss, line 59: "You cannot currently create an account from this device. This my happen if your IP address is blocked by Wikipedia or has already reached the daily limit for new accounts. Try connecting from a different network."

app/assets/javascripts/components/util/helpers.js
```diff
@@ -102,3 +102,11 @@ export const selectUserByUsernameParam = (users, usernameParam) => {
   selectedUser = users.find(({ username }) => username.replace('?', '%3F') === usernameParam);
   return selectedUser;
 };
+
+export const canUserCreateAccount = async () => {
+  const response = await fetch('https://en.wikipedia.org/w/api.php?action=query&meta=userinfo&uiprop=cancreateaccount&format=json&origin=*');
+  const data = await response.json();
+
+  // cancreateaccounterror is present if the user cannot create an account
+  return !data.query.userinfo.cancreateaccounterror;
```
ragesoss, line 111: It would be good to document what a typical error response looks like here. In the future, we my want to handle multiple different error types and/or display the error message that mediawiki returns.

PR #5354, opened 2023-04-08
Fix HMR for i18n

TheTrio:
This brings i18n under the control of webpack - which is what I should've done in the first place. 

Bringing it under webpack means that HMR once again works for components which use i18n(which most of them do)


TheTrio:
HMR for components which use i18n is currently broken. 

To try it out, switch to master, run `yarn hot` (after enabling it in `application.yml`) and start the rails server. 

Visit any course page and make a simple edit to `app/assets/javascripts/components/course/course.jsx` - say add a `h1` element like so 

image

the page won't update. If you open the devtools in the browser, you'll find this error - 
image

This error occurs because `i18n` is currently outside the control of webpack - its included from inside Rails(see `app/views/shared/_head.html.haml`)

This PR attempts to fix this. Repeat the same steps on this branch and you shall see the changes instantly.

ragesoss:
I deployed this to the staging server, and it broke all the i18n.

staging i18n

ragesoss:
Per discussion on slack, and the problem I ran into with it not properly rebuilding the cache when I deployed this to staging, looks like this isn't worth the extra thing to think about.

Issue #5343, opened 2023-04-03
Default sorting of Students list is not ideal, and resorting state is not retained when navigating between tabs


ragesoss:
### What is happening?

On the Students (aka Editors) tab of a course page, students are listed by default in the order in which they joined the Course (I think). They can be resorted using either the sorting dropdown menu at the top of the Students list, or by clicking one of the headers in the Students tab. (Some sorting methods, such as sorting by last name, are only available via the dropdown.)

Sorting by last name and falling back to username would be a better default.

If a user re-sorts the list, by last name for example, then it retains that sorting while the user navigates between tabs... until a data refresh occurs (which happens when you change course tabs and it's been more than a minute since the last fetch), and then the users go back to default sorting.

If a user has set a value for sorting the users, then this sorting should be re-applied after user data gets refreshed.

### Additional context

The core of this bug is a Redux problem, and can be thought of primarily in terms of redux actions and state. In particular, the RECEIVE_USERS and SORT_USERS actions are the relevant ones here.

However, there's a related code quality problem that will need to be fixed first in order to leave us with a pure redux state-management problem, which is that the current user sorting feature offers options of 'first name' and 'last name', but these don't come directly from the server and are instead set client-side in components via the `setRealName` functions that are present in StudentExerciseRow and StudentRow components. These functions are actually mutating the Redux state without using a reducer, which means that the sorting options mostly work, but aren't easy to reason about. So I think the first thing will be to move the setting of those first and last name values into the reducer path for RECEIVE_USERS instead of the current situation where it happens during rendering as a side effect. Once that's fixed, it should be relatively easy from there to make sorting behave properly in conjunction with re-fetching of user data.

sharon-odhiambo:
HI @ragesoss , a quick one on this. Just clarifying whether it's the course page or the campaign page?

ragesoss:
Course page.

sharon-odhiambo:
This is what I'm getting on the course page
![Screenshot from 2023-04-11 10-30-26](https://user-images.githubusercontent.com/88780311/231196102-5828dbde-e302-4c40-b76c-f0d3f777b6df.png)
And this on the campaign page
![Screenshot from 2023-04-11 10-30-56](https://user-images.githubusercontent.com/88780311/231196222-94110868-782d-4632-bc2e-f8e199a9737b.png)

sharon-odhiambo:
[...]

PR #5360, opened 2023-04-12
Change course students default sorting to username

sharon-odhiambo:

## What this PR does
This PR adds a default sorting on the students tab in the Course Page to username. This helps with a better user experience especially when the facilitator wants to find the students by their names.
This PR is solving this [issue](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5343)

## Screenshots
Before:
Screenshot from 2023-04-12 08-22-10

After:
Screenshot from 2023-04-12 08-23-06
[Screencast from 04-12-2023 08:39:50 AM.webm](https://user-images.githubusercontent.com/88780311/231461218-81feb648-a404-4663-9319-4f5e76140973.webm)

### DIFF
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -4,9 +4,8 @@ import { sortByKey } from '../utils/model_utils';
 const initialState = {
   users: [],
   sort: {
-    sortKey: null,
+    sortKey: 'real_name',
     key: null,
-    initialSortKey: 'real_name'
   },
   isLoaded: false,
   lastRequestTimestamp: 0 // UNIX timestamp of last request - in milliseconds
@@ -25,17 +24,12 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-      const sortKey = state.sort.sortKey || state.sort.initialSortKey;
-      const sortedUsers = sortByKey(action.data.course.users, sortKey, 'username');
+      const sortedUsers = sortByKey(action.data.course.users, state.sort.sortKey, 'username');
       return {
         ...state,
         users: sortedUsers.newModels,
         isLoaded: true,
         lastRequestTimestamp: Date.now(),
-        sort: {
-          ...state.sort,
-          sortKey
-        }
       };
     }
 
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -24,7 +24,7 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-      const sortedUsers = sortByKey(action.data.course.users, state.sort.sortKey, 'username');
+      const sortedUsers = sortByKey(action.data.course.users, state.sort.sortKey, 'username', null, false, false, true);
       return {
         ...state,
         users: sortedUsers.newModels,
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -6,8 +6,8 @@ const initialState = {
   sort: {
     sortKey: null,
     key: null,
-    defaultKey: 'username',
-    defaultDirection: 'desc'
+    initialSortKey: 'real_name',
+    initialSortDirection: 'asc'
   },
   isLoaded: false,
   lastRequestTimestamp: 0 // UNIX timestamp of last request - in milliseconds
@@ -26,7 +26,7 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-        const sortedUsers = sortByKey(action.data.course.users, state.sort.defaultKey, null, false, false, false);
+     const sortedUsers = sortByKey(action.data.course.users, state.sort.initialSortKey, 'username' || state.sort.sortKey, null, false, false, false);
       return {
         ...state,
         users: sortedUsers.newModels,
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -6,6 +6,7 @@ const initialState = {
   sort: {
     sortKey: null,
     key: null,
+    initialSortKey: 'real_name'
   },
   isLoaded: false,
   lastRequestTimestamp: 0 // UNIX timestamp of last request - in milliseconds
@@ -24,8 +25,7 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-      const initialSortKey = 'real_name';
-      const sortKey = state.sort.sortKey || initialSortKey;
+      const sortKey = state.sort.sortKey || state.sort.initialSortKey;
       const sortedUsers = sortByKey(action.data.course.users, sortKey, 'username');
       return {
         ...state,
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -26,7 +26,7 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-        const sortedUsers = sortByKey(action.data.course.users, state.sort.defaultKey, null, false, false, true);
+        const sortedUsers = sortByKey(action.data.course.users, state.sort.defaultKey, null, false, false, false);
       return {
         ...state,
         users: sortedUsers.newModels,
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -6,8 +6,6 @@ const initialState = {
   sort: {
     sortKey: null,
     key: null,
-    initialSortKey: 'real_name',
-    initialSortDirection: 'asc'
   },
   isLoaded: false,
   lastRequestTimestamp: 0 // UNIX timestamp of last request - in milliseconds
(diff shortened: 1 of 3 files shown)


sharon-odhiambo:
HI @ragesoss , could you kindly give feedback on this?

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +25,16 @@ const SORT_DESCENDING = {
...
 export default function users(state = initialState, action) {
   switch (action.type) {
-    case RECEIVE_USERS: return {
-      ...state,
-      users: action.data.course.users,
-      isLoaded: true,
-      lastRequestTimestamp: Date.now()
-    };
+    case RECEIVE_USERS: {
+        const sortedUsers = sortByKey(action.data.course.users, state.sort.defaultKey, null, false, false, false);
```
ragesoss, line 29: Using `defaultKey` here every time means it will still override the user-selected sort mode each time users are received. Instead, you could set `username` as the initial sortKey and then sort by the current sortKey when users are received.

sharon-odhiambo: Yeah sure. Thank you. Let me update this.

sharon-odhiambo:
Hi @ragesoss , I updated this and added a screenshot in the PR of the test.

sharon-odhiambo:
Hi @ragesoss , I also made updates on this and would appreciate a review when you can.

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +25,16 @@ const SORT_DESCENDING = {
...
 export default function users(state = initialState, action) {
   switch (action.type) {
-    case RECEIVE_USERS: return {
-      ...state,
-      users: action.data.course.users,
-      isLoaded: true,
-      lastRequestTimestamp: Date.now()
-    };
+    case RECEIVE_USERS: {
+     const sortedUsers = sortByKey(action.data.course.users, state.sort.initialSortKey, 'username' || state.sort.sortKey, null, false, false, false);
```
ragesoss, line 29: Why are there so many arguments for `sortByKey` here, when there are fewer in the SORT_USERS action below?

ragesoss: It looks like this still sorts by initialSortKey every time?

ragesoss: It might be helpful to document the arguments or introduce constants to name them.

In this case, it looks like sortByKey already has logic for handling a `refresh` sort, where the sorting need not consider the previous key.

sharon-odhiambo: Oh for the RECEIVE_USERS I added the default sort direction which I guess are not that necessary. I think I can do away with them and remain with only what's making a difference. 
I think it sorts by initialSortKey every time a specific key is not selected.

app/assets/javascripts/reducers/users.js
```diff
@@ -6,6 +6,8 @@ const initialState = {
   sort: {
     sortKey: null,
     key: null,
+    initialSortKey: 'real_name',
```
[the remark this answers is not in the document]
sharon-odhiambo: Uhm, I think I can just set the default value in the reducer instead.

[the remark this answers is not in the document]
ragesoss: I don't think that helps solve the problem here, which is that user-selected sorting gets overridden when RECEIVE_USERS happens (again, after the initial action followed by a sorting action that the user triggered). you can just set the initial value of  `sort.sortKey`, and then use that (which may have changed based on a subsequent SORT_USERS action) every time within the RECEIVE_USERS reducer.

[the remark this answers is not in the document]
sharon-odhiambo: Would it be better if I did the default sorting in the SORT_USERS instead? To avoid the overriding by the RECEIVE_USERS?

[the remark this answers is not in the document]
ragesoss: I don't think I understand what you're suggesting. Your current implementation already does default sorting in RECEIVE_USERS.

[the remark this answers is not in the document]
sharon-odhiambo: Yes. I was wondering whether I should move the default sorting to SORT_USERS instead to avoid the overidding. But also isn't RECEIVE_USERS only called on refresh? Which then I think should reset the sorting to default? 
I think I'm a bit confused, I hope my communication is clear.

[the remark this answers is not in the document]
sharon-odhiambo: Screenshot from 2023-04-13 13-41-04
Something like this?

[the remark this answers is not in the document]
ragesoss: There are two situations where RECEIVE_USERS action happens:
1. when a course page is rendered initially, before the client has received any data about the users in the course. No sorting initiated by the user can take place before the initial user data is received. 
2. when at least 1 minute has passed and the user navigates between tabs, and the client automatically requests a refresh of user data. this is the case where we don't want the sorting to get reset. there might be additional users who were not part of the course previously, and so this might result in changes to the user list... but it should not change the sorting, including in the case where the user selected their preferred non-default sorting.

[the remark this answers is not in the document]
sharon-odhiambo: Okay, this is clearer now. Let me update it. Thank you.

sharon-odhiambo:
Hi @ragesoss , I updated the code to this. Could you please check and give feedback.

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +23,22 @@ const SORT_DESCENDING = {
...
   switch (action.type) {
-    case RECEIVE_USERS: return {
-      ...state,
-      users: action.data.course.users,
-      isLoaded: true,
-      lastRequestTimestamp: Date.now()
-    };
+    case RECEIVE_USERS: {
+      const initialSortKey = 'real_name';
+      const sortKey = state.sort.sortKey || initialSortKey;
```
ragesoss, line 28: Is there any reason the `real_name` value can't simply be set in the `initialState`?

sharon-odhiambo: Updated.

app/assets/javascripts/reducers/users.js
```diff
@@ -6,6 +6,7 @@ const initialState = {
   sort: {
     sortKey: null,
     key: null,
+    initialSortKey: 'real_name'
```
ragesoss, line 9: Can this value not just be used for sortKey on line 7? then there's no need for conditionals within the reducer.

sharon-odhiambo: Hmm, I didn't think of it this way. This definitely makes more sense and much cleaner code! Tried and it works!

sharon-odhiambo:
Hi @ragesoss , updated it. I hope it's fine now?

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +23,16 @@ const SORT_DESCENDING = {
...
 export default function users(state = initialState, action) {
   switch (action.type) {
-    case RECEIVE_USERS: return {
-      ...state,
-      users: action.data.course.users,
-      isLoaded: true,
-      lastRequestTimestamp: Date.now()
-    };
+    case RECEIVE_USERS: {
+      const sortedUsers = sortByKey(action.data.course.users, state.sort.sortKey, 'username');
```
ragesoss, line 27: Per my earlier comment, I think this is where we probably want to use the `refresh` parameter of `sortByKey`. Otherwise, if the list gets sorted by `username` by the user, then when new user data is received, it will resort it by `username` but it will take `username` as the previous sort key as well, so it will reverse the sorting. I guess that could still result in changing the order of the list if the user has already sorted by something is a non-default direction, though, because sortByKey is implicitly written to assume that it's just acting on the already-received set of models and re-sorting them. hmm... not entirely sure what the best way around that issue is. but I think i'm okay with ignoring that problem, and just going with the `refresh` route so that in most cases it will behave as expected.

sharon-odhiambo: Hi @ragesoss ,I tried adding the refresh parameter here but didn't quite see much change. Maybe it will need a change in the utils in the sortByKey function but I think the current issue only affects username and probably ignoring it won't be a bad idea?
Kindly let me know.

ragesoss: I think the refresh parameter should definitely be used here (and there is no need to provide the previous key parameter, which can just be null).

sharon-odhiambo: Okay, lemme implement it.

sharon-odhiambo: Updated it.

app/assets/javascripts/reducers/users.js
```diff
@@ -24,7 +24,7 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-      const sortedUsers = sortByKey(action.data.course.users, state.sort.sortKey, 'username');
+      const sortedUsers = sortByKey(action.data.course.users, state.sort.sortKey, 'username', null, false, false, true);
```
ragesoss, line 27: You forgot to remove the 'username', so there are too many params.

sharon-odhiambo: Oh yeah. Removed it.

sharon-odhiambo: Also thanks for the continous feedback you've provided on this. Learnt a lot!

sharon-odhiambo:
Hi @ragesoss , is there anything else I need to do on this issue or it's ready?

ragesoss:
I think it's ready, but I would like to test it manually before merging. I
probably won't do that until next week (as I'm traveling this week).

ragesoss:
It looks like this may have introduced a test failure that will require a slight adjustment to the instructor role spec.

sharon-odhiambo:
Hi @ragesoss, lemme check it out.

sharon-odhiambo:
Hi @ragesoss , updated it. Did you test it locally?

ragesoss:
not yet! i will soon.

ragesoss:
Testing this out, I'm realizing that there are some problems with how we're processing names that make this a little tricky to fully solve without fixing those existing problems first. In particular, we're setting first_name and last_name in components, and in doing so, mutate the Redux `users` state without going through a reducer. I think I should document the problem and how to fix, in the issue description, and then we'll be in a better position to take another shot at this PR.

ragesoss:
Okay, I've updated the issue description. I'm closing this PR, per my previous comment. Thanks for taking a shot at this and helping to uncover the bigger problem that I didn't realize was there.

Issue #4283, opened 2020-11-09
Add stats for edits in the Lexeme namespace on Wikidata


ragesoss:
### Current Behavior:
Lexeme edits are imported, but they do not contribute to either the top-level stats for Wikidata courses, or show up in the CSV stat downloads specific to Wikidata.

### Desired Behavior:

1. A summary of Lexeme contributions should be in the wikidata CSV data.
2. Lexemes should be shown alongside Items as a top-level stat (including lexemes created, lexemes edited) in Wikidata courses.

cyrillefr:
#4999 brought me here :D
Hello @ragesoss,
Can I work on this issue ?

Are we talking about these stats ?
http://localhost:3000/courses/Stanford_Law_School/Advanced_Legal_Research_Winter_2020_(Winter) (course)
Then click on Edits data ? -> CSV generated

How can I import/reimport new `www.wikidata.org` stats for a course ?

Have you got courses from dashboard that I can import that have articles within the lexema namespace ?

cyrillefr:
Hello @ragesoss, 
May I work on this issue ?

ragesoss:
@cyrillefr sure, go for it! :-)

ragesoss:
The easiest way to get a useful dev environment for this is probably just to look at recent changes in the Lexeme namespace, choose a few usernames that are making edits, and add those to a course on your machine: https://www.wikidata.org/wiki/Special:RecentChanges?hidebots=1&hidecategorization=1&namespace=146&limit=50&days=7&urlversion=2

cyrillefr:
Hello @ragesoss ,

Thanks for the tip. I have now some data from the Lexeme namespace.

But I would need clarification about where to code exactly.

I have spotted 3 places about Wikidata
- http://localhost:3000/campaigns/miscellanea/programs
    There is a "wikidata stats" with some data under the label
- http://localhost:3000/campaigns/miscellanea/wikidata.csv
   Some CSV data processed there
- http://localhost:3000/courses/Stanford_Law_School/Advanced_Legal_Research_Winter_2020_(Winter)
  On course homepage, there is a "Download stats" button then a "Wikidata stats" button

So, where are the 1. and the 2. ?

ragesoss:
[...]

PR #5365, opened 2023-04-14
Add stats for edits in the lexeme namespace on wikidata

cyrillefr:
## What this PR does
Issue   #4283 (Add stats for edits in the Lexeme namespace on Wikidata) 

- new locales
- add a header for the csv generation + processing of a new string in revisions summary
- add data at the Front End: top of Course/Campaign page

### DIFF
--- a/lib/wikidata_summary_parser.rb
+++ b/lib/wikidata_summary_parser.rb
@@ -139,7 +139,8 @@ class WikidataSummaryParser
   end
 
   def created_item?
-    @summary.include? 'wbeditentity-create'
+    @summary.include?('wbeditentity-create') &&
+      @summary.exclude?('wbeditentity-create-lexeme')
   end
 
   def created_lexeme_item?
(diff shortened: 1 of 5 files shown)


lib/wikidata_summary_parser.rb
```diff
@@ -140,6 +142,10 @@ def created_item?
     @summary.include? 'wbeditentity-create'
   end
 
+  def created_lexeme_item?
+    @summary.include? 'wbeditentity-create-lexeme'
```
ragesoss, line 146: This string includes the substring `wbeditentity-create`, so we've already been miscounting lexeme creations as item creation. With this update, now a lexeme creation gets counted as both item creation and lexeme creation. I guess the matcher for `created_item?` needs to be updated to make sure it excludes lexemes.

cyrillefr: Ooups. You are right. I will update the `created_item?` method.

ragesoss:
Thanks!

PR #5396, opened 2023-05-16
add scoping methods to course creator

### LINKED ISSUE #4658: Improve UX for Article Scoped Programs
It should be intuitive to set up a program that will track a specific scope of articles, using one or more of the scoping methods already available (assignments, categories, templates, PetScan, PagePile).

The user interface should explicitly note the methods for scoping a course at the point of choosing the program type, it should guide users who have created an ArticleScopedProgram that doesn't have any in-scope articles yet on how to set up the desired scoping, and it should make the current scoping settings more obvious (eg, as part of the Home tab) when viewing such a program.

TheTrio:
Partially resolves #4658 

This adds a new segment to the course creation modal which allows users to track a specific set of articles more easily by configuring the 4 scoping methods - PetScan, PagePile, Templates and Categories - in the course creation modal itself.

This is how it looks in action 

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/10794178/351bbdd7-f3cf-47e2-ace0-84042675fa40


ragesoss:
I'll look at it in more detail later this week, but on a first pass, this looks great. I surprised it didn't break any course creation specs; it should probably include a feature spec that exercises most of the new functionality.

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
       participate_in_campaign: Participate in a Campaign
       save_cloned_course: Save
       start_date: Start of activity tracking
+      scoping_methods:
+        scoping_methods: Scoping Methods
+        configure: Configure Scoping Methods
+        about: >
+          This is what lets the dashboard know which articles you wish to track. The articles tracked contribute to the course statistics and are displayed on the articles tab. Rather than adding each article you wish to track individually, you can add filters to track several of them at once.
```
ragesoss, line 893: "For an Article Scoped Program, the Dashboard will only track edits to the articles that you have configured it for. After your page has been created, you can add individual articles to track assigning articles to individual editors or by adding them as Available Articles. For larger sets of articles, you can use any of these tools to define articles you want to track."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+          
+          Say you only want to track pages created from the “Historic sites templates” or pages which belong to the category “Algebra” but not “Geometry”. With the correct scoping method, doing this is incredibly easy
+        categories: Categories
+        categories_short_desc: Categories are intended to group together pages on similar subjects. Categories are normally found at the bottom of an article page.
```
ragesoss, line 897: "The Category feature on a wiki is used to collect related articles. You can use the name of the Category to tell the Dashboard to track all the articles in that Category."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+        categories_desc: >
+          Categories are intended to group together pages on similar subjects. Categories help readers to find, and navigate around, a subject area, to see pages sorted by title, and to thus find article relationships.
```
ragesoss, line 899: "Tracking a Category means that the Dashboard will count edits to any articles in that Category. By default (depth 0), only the articles that are directly in that Category will be tracked. You can increase the depth value (up to a maximum of 3) if you want to include articles that are in sub-categories of that Category. Be sure to browse the Category first, to make sure the sub-Categories match what you intend to track."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+          
+          Say you only want to track pages created from the “Historic sites templates” or pages which belong to the category “Algebra” but not “Geometry”. With the correct scoping method, doing this is incredibly easy
+        categories: Categories
+        categories_short_desc: Categories are intended to group together pages on similar subjects. Categories are normally found at the bottom of an article page.
+        categories_desc: >
+          Categories are intended to group together pages on similar subjects. Categories help readers to find, and navigate around, a subject area, to see pages sorted by title, and to thus find article relationships. 
+
+          You can enter the categories whose pages you wish to track. If the depth is set to 0, only the direct pages will be tracked. 
+        templates: Templates
+        templates_short_desc: Templates are pages intended to be included in other pages. Templates are normally found at the top of an article page.
```
ragesoss, line 903: "You can track articles based on a Template that they contain, which works whether the Template is used on article itself, or on the corresponding Talk page — such as WikiProject templates."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+        petscan: PetScan
+        petscan_short_desc: PetScan is a tool that allows you to search for pages based on a wide range of criteria - pages in a category, pages that belong to a WikiProject, pages that have a certain template, etc
+        pagepile: PagePile
+        pagepile_short_desc: PagePile is a tool that allows you to create a list of pages which doesn't change over time and is identified by a unique ID
+        templates_desc: >
+          A template is a Wikipedia page created to be included in other pages. Templates usually contain repetitive material that might need to show up on a larger number of articles or pages. They are commonly used for boilerplate messages, standardized warnings or notices, infoboxes, navigational boxes, and similar purposes.
```
ragesoss, line 909: "Templates are used for a wide variety of purposes on wiki pages, but some are specific to a topic. Navigation templates, WikiProject templates, and topic-specific Infobox templates are particularly useful when using a Template to configure an Article Scoped Program."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+        categories: Categories
+        categories_short_desc: Categories are intended to group together pages on similar subjects. Categories are normally found at the bottom of an article page.
+        categories_desc: >
+          Categories are intended to group together pages on similar subjects. Categories help readers to find, and navigate around, a subject area, to see pages sorted by title, and to thus find article relationships. 
+
+          You can enter the categories whose pages you wish to track. If the depth is set to 0, only the direct pages will be tracked. 
+        templates: Templates
+        templates_short_desc: Templates are pages intended to be included in other pages. Templates are normally found at the top of an article page.
+        petscan: PetScan
+        petscan_short_desc: PetScan is a tool that allows you to search for pages based on a wide range of criteria - pages in a category, pages that belong to a WikiProject, pages that have a certain template, etc
```
ragesoss, line 905: "PetScan is a tool that allows you to query wiki pages based on a wide range of criteria — including categories, templates, wikilinks, and much more. You can use a PetScan query ID to configure a set of articles to track on the Dashboard."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+        pagepile: PagePile
+        pagepile_short_desc: PagePile is a tool that allows you to create a list of pages which doesn't change over time and is identified by a unique ID
+        templates_desc: >
+          A template is a Wikipedia page created to be included in other pages. Templates usually contain repetitive material that might need to show up on a larger number of articles or pages. They are commonly used for boilerplate messages, standardized warnings or notices, infoboxes, navigational boxes, and similar purposes.
+        pagepile_desc: >
+          PagePile is a tool developed by the Wikimedia Foundation for the Wikimedia projects, including Wikipedia. It allows users to collect and organize a group of web page links into a single bundle called a "pile." These piles can then be easily shared with others, enabling collaborative work and information sharing.
```
ragesoss, line 911: "PagePile is a tool on Wikimedia Cloud.  It allows users to collect and organize a group of wiki pages into a single bundle called a "pile." These piles can then be easily shared with others, enabling collaborative work and information sharing.
 
In contrast to PetScan, the articles in a pile will not change over time, even when categories, templates or other data used to create that set of articles have changed."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+
+          PagePile serves as a convenient way to gather and present a curated set of Wikipedia articles or other web pages related to a specific topic or theme. It can be used for various purposes, such as creating reading lists, compiling research materials, organizing articles for a project, or simply bookmarking a collection of pages for personal reference
+        petscan_desc: >
+          PetScan is another tool developed by the Wikimedia Foundation that allows users to generate custom lists of Wikipedia pages based on specific criteria. It is a powerful tool for retrieving and filtering information from Wikipedia's vast database of articles.
```
ragesoss, line 915: "PetScan is a powerful but complex tool for querying a set of wiki pages based on a wide variety of criteria and data sources. You can use PetScan directly to create the query for a set of articles you want to track, which gives you access to the full set of options to target your query very precisely. Alternatively, you can create a PetScan query from some of the most common options directly from the Dashboard.

PetScan saves the query parameters and provides an ID for that query. The results of that query may change over time, for example when new articles are added to category that the query includes. The Dashboard will periodically refresh the query results for a given ID, and will update its statistics based on the latest set of articles that PetScan produces."

config/locales/en.yml
```diff
@@ -885,6 +886,45 @@ en:
...
+        petscan_have_psid_title: I have a PetScan ID
+        petscan_not_have_psid_title: I don't have a PetScan ID
+        petscan_have_psid_desc: >
+          Use this if you have already built a PetScan query and have the ID for it. You can find the ID in the URL of the PetScan query. For example, if the URL of the PetScan query is https://petscan.wmflabs.org/?psid=1234567, then the ID is 1234567.
+
+          If you don't have a PetScan ID, you can build a query using the PetScan tool and then come back here to enter the ID.
+        petscan_not_have_psid_desc: >
+          Use this if you don't have a PetScan ID and wish to use a simplified builder to create it. This lacks some of the advanced features of the PetScan tool, but is easier to use and you don't have to leave the dashboard to build the query!
+        petscan_create_psid: Create a PetScan ID using the official PetScan tool
```
ragesoss, line 926: What is this used for?

TheTrio: image

in here

ragesoss: I don't see "Create a PetScan ID using the official PetScan tool" anywhere in that screenshot.

app/assets/javascripts/components/common/ScopingMethods/autocomplete_categories_input.jsx
```diff
@@ -0,0 +1,44 @@
...
+  return (
+    <>
+      <label htmlFor="categories">{label} </label>
+      <AsyncSelect
+        loadOptions={loadOptions}
+        placeholder={I18n.t('courses_generic.creator.scoping_methods.start_typing_to_search')}
+        isMulti
+        id="categories"
+        onChange={updateCategories}
+        noOptionsMessage={() => 'No categories found'}
```
ragesoss, line 37: I notice that it's a little wonky as I start typing something, where I may see results from the first few characters, then 'No categories found' as I type more, then showing results again as I get close to the complete input. For example (on English Wikipedia):

femi -> one result
femin -> no results
feminism -> several results

It looks like it basically only matches complete words. That's fine behavior, but I wonder the 'No categories found' message might be slightly confusing.

As I've played around with it and thought through this comment, I can't think of a better way to do it. I don't think removing would be an improvement. This is fine as-is, unless you can think of a way to reduce the potential for confusion.

config/locales/en.yml
```diff
@@ -885,6 +886,48 @@ en:
       participate_in_campaign: Participate in a Campaign
       save_cloned_course: Save
       start_date: Start of activity tracking
+      scoping_methods:
+        scoping_methods: Scoping Methods
+        configure: Configure Scoping Methods
+        about: >
+          For an Article Scoped Program, the Dashboard will only track edits to the articles that you have configured it for. After your page has been created, you can add individual articles to track assigning articles to individual editors or by adding them as Available Articles. For larger sets of articles, you can use any of these tools to define articles you want to track
```
ragesoss, line 893: Add a period at the end of this sentence.

config/locales/en.yml
```diff
@@ -885,6 +886,48 @@ en:
...
+        categories: Categories
+        categories_short_desc: The Category feature on a wiki is used to collect related articles. You can use the name of the Category to tell the Dashboard to track all the articles in that Category
```
ragesoss, line 895: Period at the end.

config/locales/en.yml
```diff
@@ -885,6 +886,48 @@ en:
...
+        categories_desc: >
+          Tracking a Category means that the Dashboard will count edits to any articles in that Category. By default (depth 0), only the articles that are directly in that Category will be tracked. 
+          
+          You can increase the depth value (up to a maximum of 3) if you want to include articles that are in sub-categories of that Category. Be sure to browse the Category first, to make sure the sub-Categories match what you intend to track
+        templates: Templates
+        templates_short_desc: You can track articles based on a Template that they contain, which works whether the Template is used on article itself, or on the corresponding Talk page — such as WikiProject templates.
+        petscan: PetScan
+        petscan_short_desc: PetScan is a tool that allows you to query wiki pages based on a wide range of criteria — including categories, templates, wikilinks, and much more. You can use a PetScan query ID to configure a set of articles to track on the Dashboard
```
ragesoss, line 903: Period at the end.

config/locales/en.yml
```diff
@@ -885,6 +886,48 @@ en:
...
+        pagepile: PagePile
+        pagepile_short_desc: PagePile is a tool that allows you to create a list of pages which doesn't change over time and is identified by a unique ID
+        templates_desc: >
+          Templates are used for a wide variety of purposes on wiki pages, but some are specific to a topic. Navigation templates, WikiProject templates, and topic-specific Infobox templates are particularly useful when using a Template to configure an Article Scoped Program
```
ragesoss, line 907: Period at the end.

config/locales/en.yml
```diff
@@ -885,6 +886,48 @@ en:
...
+        pagepile_desc: >
+          PagePile is a tool on Wikimedia Cloud. It allows users to collect and organize a group of wiki pages into a single bundle called a "pile." These piles can then be easily shared with others, enabling collaborative work and information sharing.
+
+          In contrast to PetScan, the articles in a pile will not change over time, even when categories, templates or other data used to create that set of articles have changed
```
ragesoss, line 911: Period at the end.

app/assets/javascripts/components/course_creator/scoping_methods/pagepile_scoping.jsx
```diff
@@ -0,0 +1,54 @@
...
+      <label htmlFor="pagepile-ids">Enter PagePile IDs</label>
+      <CreatableSelect
+        inputValue={inputValue}
+        isClearable
+        isMulti
+        menuIsOpen={false}
+        onChange={ids => dispatch({ type: UPDATE_PAGEPILE_IDS, ids })}
+        onInputChange={newValue => !isNaN(newValue) && setInputValue(newValue)}
+        onKeyDown={handleKeyDown}
+        placeholder="Type something and press enter..."
```
ragesoss, line 45: All these English placeholders, labels, etc, should be done with i18n... although that can be in a later PR if you don't want to implement it now.

More importantly, I think the requirement to press enter is going to cause some confusion (both in this component and probably the other ones). A user may enter a value (in this case, a PagePile ID) and then click 'Next' without first pressing enter. In that case, their input is lost but the new view is shown so the user may not realize it. (It looks like the input state is lost, as the input is empty if I click Clicking 'next' should automatically do the equivalent of pressing 'enter', if there is input that hasn't yet been converted into a multi-input value.

app/assets/javascripts/components/course_creator/scoping_methods/petscan_query_builder.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import CategoryAutoCompleteInput from '../../common/ScopingMethods/autocomplete_categories_input';
+import { UPDATE_EXCLUDE_CATEGORIES_PETSCAN, UPDATE_EXCLUDE_TEMPLATES_PETSCAN, UPDATE_INCLUDE_CATEGORIES_PETSCAN, UPDATE_INCLUDE_TEMPLATES_PETSCAN } from '../../../constants/scoping_methods';
+import TemplatesAutoCompleteInput from '../../common/ScopingMethods/autocomplete_templates_input';
+import { useSelector } from 'react-redux';
+
+const PetScanQueryBuilder = () => {
```
ragesoss, line 7: I testing this with some values I expected to work, but the resulting query has 0 results.

I entered "Wikipedia FemTechNet" a category to include, and "WikiProject Feminism" as a template to include.

I can enter the same values as a Category and a Template separately (not via PetScan) and those categories show a non-zero article count. The Category actually only includes pages in the Wikipedia namespace, while the template is a talk page template. But I would at least have expected the PetScan query to return results based on the template.

I note that the PetScan UI shows 'Intersection' rather than 'Union' for the category part, when I view that query. And I note that only mainspace (the unlabeled one) is selected on the 'Page properties' tab, which is possibly why the template didn't return any results.

I think we need to play around with more combinations of PetScan queries to see where it's likely to go wrong compared with user expectations. Maybe the example I tried could be fixed by include additional options (such as more namespaces)?

TheTrio: have pushed a fix for this

app/assets/javascripts/components/common/ScopingMethods/autocomplete_categories_input.jsx
TheTrio: have pushed a fix for this

ragesoss:
The default depth for Categories (contradicting the explanation) is still 1.

ragesoss:
I tested out the PetScan again, trying to combine a (talk) template with a category, and it returns 0 results... while I can adjust the query afterwards to use only the category *or* only the template, and either one returns results. (The new default of selecting Talk namespace makes the template one work.)

I tried changing the 'Intersection' default on the Categories tab to 'Union', but it still returns no results when there is also template on the other tab. I think we should drop the petscan builder and just let users get their own PSID.

config/locales/en.yml
TheTrio: Ah you were referring to that - its here 
image

TheTrio:
Have got rid of the custom petscan builder.

lib/course_creation_manager.rb
```diff
@@ -115,4 +120,39 @@ def add_instructor_to_course
...
+
+    add_categories_to_course @scoping_methods[:categories] if @scoping_methods[:categories]
+    add_templates_to_course @scoping_methods[:templates] if @scoping_methods[:templates]
+    add_page_pile_to_course @scoping_methods[:pagepile] if @scoping_methods[:pagepile]
+    add_petscan_to_course @scoping_methods[:petscan] if @scoping_methods[:petscan]
+  end
+
+  def add_categories_to_course(category_params)
+    depth = category_params[:depth]
+    create_all_category(category_params[:tracked], depth, 'category')
```
ragesoss, line 144: For each of this `add_X_to_course` methods, is it possible for the sub-params like `category_params[:tracked]` to be `nil` or to throw a NoMethodError for `[ ]`, even though the param itself made it through the guard statement of `@scoping_methods[:categories]`?

ragesoss:
This is going to help a lot, I think! Great work so far.

PR #5397, opened 2023-05-18
Converting CampaignAlerts class component to functional component

Aminehassou:
With reference to #5393 
## What this PR does
Converts campaign_alerts.jsx into a functional component. Also uses the newer redux hooks
**Affected Pages:** `/campaigns/[campaign_title]/alerts`
## Videos
Before:

[before refactor assignment alerts.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/aaab269b-0ab8-4a7c-be52-99b2ee1a577a)

After:

[after refactor assignment alerts.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/2fe75429-4b0e-4a4c-9c8d-8d76f6698fce)

## Questions and Concerns
Unrelated question: I noticed that the "course" column in the alerts table is always highlighted white even when sorting by date or other columns. Is this intended design behaviour or is the white highlight supposed to change based on the column that we're sorting by?

### DIFF
--- a/app/assets/javascripts/components/alerts/admin_alerts.jsx
+++ b/app/assets/javascripts/components/alerts/admin_alerts.jsx
@@ -1,5 +1,4 @@
 import React, { useEffect } from 'react';
-import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
 
 import AlertsHandler from './alerts_handler.jsx';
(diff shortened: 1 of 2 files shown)


ragesoss:
That highlighted column header is a bug.

Aminehassou:
Should I create an issue for it? I could work on it too.

ragesoss:
Sure, please do.

PR #5404, opened 2023-05-22
Converting StudentRevisionsList and Assignment class components into functional components 

Aminehassou:
With reference to #5393 

## What this PR does
Converts `StudentRevisionsList.jsx` and `Assignment.jsx` into functional components. 
**Affected Pages:**

- `/courses/[institution_id]/[course_id]/students/articles/[student_username]` (for `StudentRevisionsList.jsx`)
- `/courses/[institution_id]/[course_id]` (for `Assignment.jsx`, only if it's an edu course)

## Videos
Before `StudentRevisionsList.jsx`

[before refactor studentRevisionList.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/63c2a365-47ba-46b5-a9b5-5482fe4c39ed)

After `StudentRevisionsList.jsx`

[after refactor studentRevisionList.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/8880a635-01c3-442c-915b-5dc9be69d7ca)

Before `Assignment.jsx`

[before refactor assignment.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/2fb4b0cf-99ff-43d3-8717-626fe5f6a8cd)

After `Assignment.jsx`

[after refactor assignment.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/77252d0f-8e1c-4b35-a3ca-2fb7618d7628)

## Redux Hooks Refactor
When refactoring `Assignment.jsx` to use redux hooks. I noticed that the the dispatched actions weren't actually being used in Assignment. They were used in Assignment's child components. So I removed the use of redux in Assignment and instead only imported the actions in the child components that were using them directly. Then I called useDispatch in those child components to actually dispatch the actions.

## Test Changes
I made some changes to the tests to account for the use of redux hooks. I also had to update two test snapshots because they were no longer valid due to me using `mount()` instead of `shallow()`

### DIFF
--- a/app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/MarkAsIncompleteButton.jsx
+++ b/app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/MarkAsIncompleteButton.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
-import { updateAssignmentStatus, fetchAssignments } from '../../../../../../../../../actions/assignment_actions';
+import { updateAssignmentStatus, fetchAssignments } from '@actions/assignment_actions';
 
 const update = ({ assignment, courseSlug, dispatch }) => async () => {
   const statuses = assignment.assignment_all_statuses;
--- a/app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/MarkAsIncompleteButton.jsx
+++ b/app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/MarkAsIncompleteButton.jsx
@@ -1,36 +1,33 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useDispatch } from 'react-redux';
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/MarkAsIncompleteButton.jsx
```diff
@@ -1,36 +1,33 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { useDispatch } from 'react-redux';
+import { updateAssignmentStatus, fetchAssignments } from '../../../../../../../../../actions/assignment_actions';
```
ragesoss, line 4: This much relative directory navigation is very hard to understand visually. Can you use an `@` path shortcut instead (as defined in jsconfig.json and used for example in `article.jsx`)?

Aminehassou: Okay. I added an actions path and replaced my imports with it. You're right, the imports look much nicer now.

Issue #5078, opened 2022-07-28
Rails mailer previews should be available in production


ragesoss:
### Current Behavior:

Wiki Education Dashboard sends out a variety of emails in various circumstances. However, these previews are not available in production — only in staging and development environments. Also, the staging environment — https://dashboard-testing.wikiedu.org/rails/mailers — is not necessarily up-to-date, not all the previews work (because some previews expect example records to be available in the database), and some could reveal private information if enabled on production (because, for example, they use the most recent record of a given type that's in the database).

### Desired Behavior:

Mailer previews should be available in production, and should work either without relying on live records from the database or should ensure that only public information is used to generate the previews.

ragesoss:
See the Course Approval Mailer preview for examples of how to work around mailers that expect records to behave like production records in the database; `define_singleton_method` can work around these cases to avoid using any production records in the previews.

cyrillefr:
Hello @ragesoss, may I work on this one ?

Btw, the Unsubmitted Course Alert Mailer does not work(but I fixed it so the check may stay :) ).

ragesoss:
@cyrillefr sure! I know I worked on some of these last year, and there might be some that are already okay and just need to be double-checked.

PR #5415, opened 2023-06-01
Fix 500 Error on Unsubmitted Course Alert Mailer Preview

cyrillefr:
## What this PR does
Fix error when previewing Unsubmitted Course Alert Mailer Preview
Task of #5078 

## Screenshots
Before:
[https://dashboard-testing.wikiedu.org/rails/mailers/unsubmitted_course_alert_mailer/message_to_instructor](url)

After:
mailer_2

### DIFF
--- a/config/locales/be-tarask.yml
+++ b/config/locales/be-tarask.yml
@@ -69,6 +69,7 @@ be-tarask:
     opt_out: Не
     opt_yes: так
     opt_no: не
+    print: Друкаваць
   article_statuses:
     not_yet_started: Пачатак працы
     in_progress: Работа ў пясочніцы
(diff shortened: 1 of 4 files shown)


spec/mailers/previews/unsubmitted_course_alert_mailer_preview.rb
```diff
@@ -17,4 +18,12 @@ def alert
   def example_user
     User.new(email: 'sage@example.com', username: 'Ragesoss')
   end
+
+  def create_classroom_program_manger
+    unless SpecialUsers.classroom_program_manager
+      SpecialUsers.define_singleton_method(:classroom_program_manager) do
```
ragesoss, line 24: This could possibly cause problems in production, I think, because the singleton method would stay in place and affect subsequent requests. It wouldn't hit this code in Wiki Education production, but it would on Programs & Events Dashboard where no classroom_program_manager is assigned.

Maybe instead, modify the mailer so that it can still work when classroom_program_manager is `nil`.

ragesoss: If the Rails mailer can handle `reply_to: nil`, then you could just use safe navigation. That user isn't actually the sender, just the reply-to email, so leaving out the reply-to would fall back to the default of replying to the sender. (In production, the sender is dashboard@dashboard.wikiedu.org .)

Issue #5199, opened 2022-11-14
Add a button to for instructors to add all students' talk pages to their watchlist


ragesoss:
### Desired Behavior:
Some instructors want to use their Wikipedia watchlists to keep an eye on the Talk pages of all their students. There's no automatic way to add all your students to your watchlist, though. As an instructor, I want to be to click a button on my Course page to have the Dashboard automatically add all the User pages of my students' to my watchlist.

(This is at the harder end of a 'newcomer friendly' task. It involves a fair amount of work, from adding a frontend UI to a Rails Controller to some logic in turning a set of Users into a set of user page titles, to editing a user's watchlist through the MediaWiki API, to writing tests for the new feature. However, each of these parts on their own should be relatively straightforward to implement by following examples of related functionality that already exists. The hard parts will be finding all the relevant examples to follow and understanding enough about the app to understand the use case. Feel free to ask for pointers for any of those hard parts.)

twinsfan421:
I can work on this

ragesoss:
> I can work on this

Great! Let me know if you'd like to join our Slack to chat about anything related.

twinsfan421:
> > I can work on this
> 
> Great! Let me know if you'd like to join our Slack to chat about anything related.

Yes, joining your slack channel sounds good. Thanks

ragesoss:
YGM

twinsfan421:
@ragesoss would you like the add all students/editors to watchlist button on the courses home page, maybe under "Actions"?  or under the Editors page(students/overview)?  Also, for the "editing a user's watchlist through the MediaWiki API", is there a relevant example for that you recommend?   Thanks

ragesoss:
Good questions. I think `students/overview` is probably the best place.

For editing the watchlist, the closest things are probably the methods that call `api_post` in `lib/wiki_edits.rb`. I expect this will look similar, in terms of figuring out the right parameters to post to the general MediaWiki API in order to make watchlist changes. See https://en.wikipedia.org/w/api.php?action=help&modules=watch

For testing this locally and changing your own watchlist via API, you might also need to register your own OAuth consumer. See `docs/oauthmd`.

Chaitanyarai899:
Hello! I would like to work on this enhancement! I hope that's ok.

reachaadrika:
@ragesoss  Hey can I work on this issue , I have to create a dev environment on my system and after that I will start working on this !

ragesoss:
[...]

PR #5416, opened 2023-06-02
Add button to add all students' talk pages to instructor's watchlist …

Abishekcs:
with reference to #5199

## What this PR does
This pull request adds a new feature to the Dashboard that allows instructors to easily add all their students' talk pages to their watchlist on Wikipedia. By clicking a button on the Course page, instructors can automatically subscribe to the Talk pages of all their students, enabling them to stay updated on discussions.
Affected Pages: /students/overview (admin-only)

## Screenshots
Before:

[Before.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/7f18dd91-4aee-4e64-a31d-1559890bd386)

After:
(Updated)

[After .webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/bd825586-403a-4d81-ab24-1d31990ee506)


app/controllers/courses_controller.rb
```diff
@@ -147,6 +149,13 @@ def suspected_plagiarism
   # User-initiated actions #
   ##########################
 
+  # /courses/:slug/students/add_to_watchlist
```
ragesoss, line 152: I would rather have this action in a separate sub-controller, similar to the `Courses::SyllabusesController`. The main Courses controller is already responsible for too many things.

lib/course_students.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+class CourseStudents
```
ragesoss, line 2: Rather than a separate class for creating an array of userpage names, I think this would make more sense as a method on CoursesUsers, so that you could call `.userpage`. We actually have some closely related methods already, spread across User and CoursesUsers, so it would make sense do some refactoring to minimize duplicated logic.

app/controllers/courses_controller.rb
Abishekcs: Okay, I will make the necessary changes

lib/course_students.rb
Abishekcs: okay

lib/watchlist_edits.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+class WatchlistEdits
```
ragesoss, line 3: Ideally, you want to re-use whatever logic you can from WikiEdits, and just include the unique logic related to watchlists in the new file.

app/assets/javascripts/components/students/containers/Overview.jsx
```diff
@@ -46,6 +47,7 @@ export class Overview extends React.Component {
         }
 
         <RandomPeerAssignButton {...this.props} />
+        { current_user.isAdvancedRole ? (<AddToWatchlistButton slug={course.slug} prefix={prefix} />) : null }
```
ragesoss, line 50: It would be better to have this aligned next to, rather than beneath, the other button.

Abishekcs: @ragesoss for this should I just apply some CSS to make them aligned next to each other?

Abishekcs:
Hi @ragesoss , I have made the updates on this and also updated the screenshot, would appreciate a review when you can.

Abishekcs:
Hi @ragesoss , I have made the updates on this and also updated the screenshot, would appreciate a review when you can.

PR #5417, opened 2023-06-02
Fix error on Unsubmitted course alert mailer preview

cyrillefr:
## What this PR does
Fix error when previewing Unsubmitted Course Alert Mailer Preview
Task of #5078 

Allow a fallback in case of nil value

Due to some git manipulation, PR #5415 has been involuntary closed.

### DIFF
--- a/app/views/unsubmitted_course_alert_mailer/email.html.haml
+++ b/app/views/unsubmitted_course_alert_mailer/email.html.haml
@@ -22,4 +22,4 @@
                 Best,
                 %br
                 %em
-                  =@classroom_program_manager&.username || 'The system'
+                  =@classroom_program_manager&.username || 'Classroom Program Manager'
--- a/app/views/unsubmitted_course_alert_mailer/email.html.haml
+++ b/app/views/unsubmitted_course_alert_mailer/email.html.haml
@@ -22,4 +22,4 @@
                 Best,
                 %br
                 %em
-                  =@classroom_program_manager.username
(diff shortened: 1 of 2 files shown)


cyrillefr:
Hello @ragesoss, sorry for the inconvenience.

I decided to hard reset the code I did in the first place (because it had to be reverted entirely), but then my branch appeared empty (it was) and github closed the PR.
So I opened this new one.

Regarding the code, I follow what you have said in the previous conversation: a reply il + safe conversation.
However, as the template needs a program manager, there is a need for a signature, always.
So, in case of nil, I have thought of a generic one, I have pick 'The system' not knowing what could be used.
This is only to show how it works, I will update with the correct one you will chose.

app/views/unsubmitted_course_alert_mailer/email.html.haml
```diff
@@ -22,4 +22,4 @@
                 Best,
                 %br
                 %em
-                  =@classroom_program_manager.username
+                  =@classroom_program_manager&.username || 'The system'
```
ragesoss, line 25: I guess "Classroom Program Manager" is the best option here. That would be a usable fallback if for some reason we removed the special user in production.

ragesoss:
No problem about the second PR. It's not significantly inconvenient for me.

app/views/unsubmitted_course_alert_mailer/email.html.haml
cyrillefr: Done :)

PR #5421, opened 2023-06-06
Add watchlist button

Abishekcs:
... watchlist #5199 

## What this PR does

This pull request adds a new feature to the Dashboard that allows instructors to easily add all their students' Talk pages to their watchlist on Wikipedia. By clicking a button on their Course page, instructors can automate the process of monitoring their students' discussions by adding their students talk page to their watchlist on Wikipedia.

## Screenshots
Before:

[Before.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/b4a216cd-2ec3-42bc-be29-42b5bf365bc6)

After:
(Updated)

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/f3c42fb1-fe3d-4626-a781-dca0d6af7c6d)


Abishekcs:
@ragesoss, I had to delete the previous pull request because I made some mistakes while rebasing the add-watchlist-button branch. Sorry for the trouble. The current pull request includes the following changes: 1) Refactoring of UI, Controller, and Classes.

app/controllers/courses/watchlist_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+require_dependency "#{Rails.root}/lib/watchlist_edits"
+
+class Courses::WatchlistController < ApplicationController
+  include CourseHelper
+
+  # /courses/:slug/students/add_to_watchlist
+  def add_to_watchlist
+    course = find_course_by_slug(params[:slug])
+    student_names = CoursesUsers.new.user_page(course.students)
+    WatchlistEdits.new(course.home_wiki, student_names).oauth_credentials_valid?(current_user)
```
ragesoss, line 12: This is not a great interface, as it does not communicate what it's going to do.

Here are some interfaces that would make sense to me:

`WatchlistEdits.new(wiki, current_user).add_to_watchlist(array_of_pagenames)`

`WatchlistEdits.new(wiki, current_user).watch_userpages(array_of_users)`

`WatchUserpages.new(wiki:, current_user:, users:)`

lib/watchlist_edits.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/wiki_edits"
+class WatchlistEdits < WikiEdits
+  def initialize(wiki = nil, my_array = nil)
+    super(wiki)
+    @my_array = my_array || []
+  end
+
+  def oauth_credentials_valid?(current_user)
```
ragesoss, line 10: A method like this, by convention, should answer the question it asks. I would expect this simply to return `true` or `false` without any actions that aren't necessary to answer that question. The method name does not give any hint that it actually also executes adding to the watchlist.

lib/watchlist_edits.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/wiki_edits"
+class WatchlistEdits < WikiEdits
+  def initialize(wiki = nil, my_array = nil)
+    super(wiki)
+    @my_array = my_array || []
```
ragesoss, line 7: This is not a useful variable name.

ragesoss (changes requested):
This is coming along.

app/controllers/courses/watchlist_controller.rb
Abishekcs: Okay

Abishekcs:
Hi @ragesoss , I have made updates to this and would appreciate a review when you can.

ragesoss:
Just wanted to let you know that reviewing this updated PR is on my todo list, but I probably won't get to it until next week. Thanks!

Abishekcs:
okay :)

app/models/course_data/courses_users.rb
```diff
@@ -144,6 +144,10 @@ def cleanup
     CourseCleanupManager.new(course, user).cleanup_articles
   end
 
+  def user_page(students)
```
ragesoss, line 147: I think it makes more sense to implement this as a User instance method that only cares about the single User record and returns the userpage for the corresponding user. Mapping that into an array can be done concisely from WatchlistEdits (eg, ` users.map(&:userpage)`).

ragesoss:
Overall, this code looks like what I'd expect for this feature! :+1: 

Have you tested it manually with a live connection to Wikipedia, to confirm that it successfully adds pages to your watchlist?

Abishekcs:
Yes, I have with my own OAuth consumer `Miscellaneous activity (Edit your watchlist)`

app/models/course_data/courses_users.rb
Abishekcs: Okay

lib/watchlist_edits.rb
```diff
@@ -0,0 +1,42 @@
...
+    return { status: 'token retrieval failed' } unless retrieve_tokens
+
+    data = build_watchlist_data(users)
+    @access_token.post(@wiki.api_url.to_s, data)
+  end
+
+  private
+
+  # Build the data hash for adding pages to the watchlist
+  def build_watchlist_data(users)
```
ragesoss, line 33: This method name could be improved. I suggest "watch_action_parameters".

lib/watchlist_edits.rb
```diff
@@ -0,0 +1,42 @@
...
+
+require_dependency "#{Rails.root}/lib/wiki_edits"
+class WatchlistEdits < WikiEdits
+  def initialize(wiki = nil, current_user= nil)
+    super(wiki)
+    @current_user = current_user
+  end
+
+  # Retrieves OAuth tokens for watchlist edits if the credentials are valid
+  def retrieve_tokens
```
ragesoss, line 11: Since `watch_userpages` is the only method called by other code, this should be a private method.

test/components/students/Add_To_Watchlist_Button.spec.jsx
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
```
ragesoss, line 1: You can remove this file. We've just updated to React 18, and removed all the enzyme tests.

lib/watchlist_edits.rb
Abishekcs: Okay

ragesoss:
I've tested this out now, and confirmed it works to add pages to the watchlist. Nice work!

A few things:
* There is no feedback on whether or not it worked successfully. I had to do some debugging to figure out why it wasn't working for me at first... it was because I was using an OAuth consumer that didn't have the right to edit the watchlist. It should show an error message if the API request goes through but returns an error, and a confirmation message if the request goes through and returns a success response.
* This feature will only be enabled for dashboard.wikiedu.org, not outreachdashboard.wmflabs.org, so it will need to be configured to only show the button when `Features.wikiEd` is true. (Search the codebase for `Features.wikiEd` to find examples of this kind of thing.)
* This feature isn't compatible with the current OAuth consumer we have deployed in production, so before this could be enabled in production I will need to get an updated consumer approved, then find a time to deploy that consumer and reset all login tokens to force users to log in again with the new consumer. The easiest way forward will be to just wait to merge this until I've taken those steps. (No changes to the PR are necessary for this.)

PR #5428, opened 2023-06-09
Fix 500 error on overdue training mailer alert

cyrillefr:
## What this PR does
Fix 500 error on overdue training mailer alert

Task of #5078 
## Screenshots
Before:
See [https://dashboard-testing.wikiedu.org/rails/mailers/overdue_training_alert_mailer/message_to_student](url)

After:
mailer_6

### DIFF
--- a/spec/mailers/previews/overdue_training_alert_mailer_preview.rb
+++ b/spec/mailers/previews/overdue_training_alert_mailer_preview.rb
@@ -20,7 +20,7 @@ class OverdueTrainingAlertMailerPreview < ActionMailer::Preview
 
   def example_alert
     OverdueTrainingAlert.new(user: example_profile.user,
-                             course: Course.first,
+                             course: Course.nonprivate.last,
                              details: example_alert_details)
   end
 
--- a/spec/mailers/previews/overdue_training_alert_mailer_preview.rb
+++ b/spec/mailers/previews/overdue_training_alert_mailer_preview.rb
@@ -20,7 +20,7 @@ class OverdueTrainingAlertMailerPreview < ActionMailer::Preview
 
   def example_alert
     OverdueTrainingAlert.new(user: example_profile.user,
-                             course: ClassroomProgramCourse.current.nonprivate.last,
+                             course: Course.first,
                              details: example_alert_details)
   end
 


cyrillefr:
Hello @ragesoss ,

I have various questions about the "level" of mocking
I do not know exactly where would be the "limit".
Here, I replaced ClassroomProgramCourse.current.nonprivate.last by Course.first, assuming that either on
staging or production, there will be at least one course.
Locally it works(without the 500 error).

On this very case, there are also some TrainingModule in the view. 
By default, this is empty, but I do not know for example if there are some on staging.
I do not have access to stagging, I can only make strong hypothesis on what causes errors.
If there is non TrainingModule in the staging DB, this PR won't work

There are some data retrieved from Course object (ex in high quality) that work on staging).

Should I mock Course + TrainingModule/Library here ?
Should I mock simple Course access in other tasks ? (the Course.find and likes, The Course.new can work anywhere)

spec/mailers/previews/overdue_training_alert_mailer_preview.rb
```diff
@@ -20,7 +20,7 @@ def example_alert_details
 
   def example_alert
     OverdueTrainingAlert.new(user: example_profile.user,
-                             course: ClassroomProgramCourse.current.nonprivate.last,
+                             course: Course.first,
```
ragesoss, line 23: I think `Course.current.nonprivate.last` or `Course.nonprivate.last` would be better than `Course.first`. The latter should be just as reliable in production as using `Course.first`, as there will definitely be a least one nonprivate course on either of the production dashboards, while the first course might be private or more likely a test course or otherwise atypical one.

ragesoss:
Ideally, this preview should reliably include multiple training modules. The training modules are all public, and some exist on staging (https://dashboard-testing.wikiedu.org/training). Mocking a course so that it will behave like a real one with training modules is an acceptable strategy. Alternatively, you could query for a real nonprivate course that includes at least two training modules. If it's relatively simple to use the mocking strategy, I think that might be better because it will be less likely to fail in a development environment with few or no courses.

Here's the stack trace for the error on staging:

```
NoMethodError: undefined method `slug' for nil:NilClass

    "https://#{ENV['dashboard_url']}/courses/#{course.slug}"
                                                     ^^^^^
  from app/models/alert_types/overdue_training_alert.rb:32:in `url'
  from app/mailers/overdue_training_alert_mailer.rb:16:in `email'
```

spec/mailers/previews/overdue_training_alert_mailer_preview.rb
cyrillefr: Thank you, I will do that.

Issue #5315, opened 2023-03-19
Error when editing cloned conditional questions in certain situations


Aminehassou:
### What is happening?
When attempting to edit a cloned question that is also a conditional, an error is returned if the original question that the conditional is based on has been deleted.

### To Reproduce

Make sure you're logged in and have administrator permissions
1. Go to `/surveys`
2. Click on the "Question Groups" tab
3. Click "New Question Group"
4. Create the question group (group details are irrelevant)
5. Edit the question group, create two questions, one question (q1-cond) should be a conditional based on the other question (q1)
6. Clone the conditional question (clone q1-cond)
7. Delete the initial question (delete q1)
8. Try editing the conditional question or its clone (edit q1-cond or clone of q1-cond)

### Expected behavior
The edit page loads and the user can edit the question as they please

### Videos
[Screencast from 03-19-2023 01:21:01 AM.webm](https://user-images.githubusercontent.com/39999898/226147036-94191ac5-a768-4acc-b2d2-a1472e0ed7a1.webm)

### System Information

 - OS: Ubuntu 22.04.1 LTS
 - Browser: Google Chrome 110.0.5481.77 (Official Build) (64-bit)

ragesoss:
Very nice bug report @Aminehassou !

leevic31:
I would like to work on this

Abishekcs:
@ragesoss  i can work on this

Abishekcs:
Update on this bug -
1) The bug occurs when editing a conditional question, `whether it is a clone or not`.
2) If the original question, on which the conditional logic is based, is deleted.
3) The bug triggers an error when attempting to edit the conditional question, but only if there are two questions remaining in the question group after the deletion.

[Edit Error.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/62c02a3b-2347-430a-bdfa-5fa443da77dc)

Abishekcs:
<b>Cause of the error -</b>

1) When a "Conditional Question" is created, the model object or record of the "Conditional Question" includes an attribute called `conditionals`, which stores the `question ID` of the original question it is based upon. 

![Screenshot from 2023-06-18 22-19-04](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/b3796ef4-a838-41c7-b10d-6367efee5075)


[...]

PR #5439, opened 2023-06-21
Fixed error when editing cloned conditional questions after deletion of base question

Abishekcs:
This pull request addresses issue #5315.

## What this PR does

This pull request addresses two issues related to conditional question editing 

1. The first issue is related to the error that occurs when attempting to edit a question, whether it's cloned or not, that is also a conditional, after the original question it is based on has been deleted. To fix this issue, I have implemented a `check` for the existence of the "Original Base Question" before allowing the editing of cloned conditional questions. If the base question is not found, it simply returns an empty value (''), and the edit page will load without any errors.

2. The second issue is regarding the self-reference problem in the Show if field. Previously, it was possible to set a question as conditionally based upon itself, which doesn't make sense logically and  it also lead to error when an attempt is made to edit that question again. To fix this issue, an additional SQL `WHERE` clause was added to exclude the currently added or edited question from the list of possible conditional references in the `Show if` field.

## Screenshots

Before:

Before

[Before.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/a7b14078-a779-41bb-9ccb-18942e4fb25e)

After:

After

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/8172bc4a-695c-43ac-9571-f784ebe74aaa)


config/initializers/surveys.rb
```diff
@@ -108,8 +107,12 @@ def set_tags
     scope :course_data_questions, ->{where("course_data_type <> ''")}
     serialize :alert_conditions, Hash
 
-    def self.for_conditionals
-      where("conditionals IS NULL OR conditionals = ''")
+    def self.for_conditionals(question_id)
+      unless question_id.nil?
+        where.not(id: question_id).where("conditionals IS NULL OR conditionals = ''")
```
ragesoss, line 112: Would it work to just use `where.not(id: question_id)` for every query? If question_id is `nil`, it should have the same behavior, I think.

app/helpers/surveys_helper.rb
```diff
@@ -119,6 +119,14 @@ def question_group_locals(surveys_question_group, index, total, is_results_view:
 
   def question_conditional_string(question)
     return '' if question.nil?
+    unless question.conditionals.nil?
```
ragesoss, line 122: I think this could be refactored to be more concise with less branching. The idea here is that we want to validate the conditional string, and ensure that the question it refers to actually exists. The best way to do that might be to make a separate method that takes a conditional string and returns `true` or `false` based on whether it references a valid conditional question. With a well-named method, the purpose of this check will also be clearer.

Abishekcs: Okay

config/initializers/surveys.rb
ragesoss: I meant, I think it would work to use the entire thing, `where.not(id: question_id).where("conditionals IS NULL OR conditionals = ''")`, for every query. That ought to work even when `question_id` is nil.

Abishekcs: Got it, yes it will work :)

Aminehassou (approved):
LGTM! The failing test is unrelated to the changes you made.

Issue #5438, opened 2023-06-20
On Timeline tab, 'Add Week' button is disabled when there is room for another week


ragesoss:
### What is happening?

When building a Timeline from scratch, you must use the 'Add Week' button to add weeks to the timeline, and then add Blocks to those weeks. The course dates — in particular, the Assignment Start and Assignment End — determine how many weeks a course timeline can hold. However, at least for short timelines that span just a few weeks from Assignment Start to Assignment End, the 'Add Week' button is disabled even when there is at least (part of) one week available at the end.

In the example from this screenshot (below) that I made while replicating this bug locally, the Assignment End is July 6, while the last week shown on the Timeline ends July 1. There is room for another week (which would only go to July 6, rather than July 8), but the 'Add Week' button is disabled. I'm guessing this is because the logic for calculating whether to disable that button is different from the rest of places where we handle timeline dates, because I can adjust the Assignment End, add a week, and then change it back, and this works as it should (without displaying the warning that the timeline extends beyond the assignment end date).

![disabled Add Week](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/848483/8c3c112d-d2bf-4baa-a07a-b4b46680085e)

Abishekcs:
I can work on this

Abishekcs:
I can work on this

PR #5448, opened 2023-07-07
Fixed Add Week button logic to account for available space on the timeline

Abishekcs:
…timeline #5438

## What this PR does

Bug Fix:

 * Updated 'Add Week' button logic to account for available space on the timeline

Solution:
* Made a modification to the code by specifying the roundingMethod option as `ceil` in the `differenceInWeeks` function. This change will ensure that fractional values are rounded up instead of being truncated to zero. By using 'ceil' as the rounding method, the calculation will accurately determine the number of weeks between the end and start dates. This should address the issue where the 'Add Week' button was incorrectly disabled when there was some available space on the timeline.
* Refrence - https://date-fns.org/v2.30.0/docs/differenceInWeeks

## Screenshots

Before:

[Before.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/bf20969c-aac9-44cc-98b9-5cfc5f15b159)

After:

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/425fb461-b980-4d1a-92e6-86ec22dcc5f3)

### DIFF
--- a/app/assets/javascripts/components/timeline/timeline.jsx
+++ b/app/assets/javascripts/components/timeline/timeline.jsx
@@ -364,7 +364,7 @@ const Timeline = createReactClass({
 
       const start = toDate(this.props.course.timeline_start);
       const end = toDate(this.props.course.timeline_end);
-      const weeksDiff = differenceInWeeks(end, start);
+      const weeksDiff = differenceInWeeks(end, start, { roundingMethod: 'ceil' });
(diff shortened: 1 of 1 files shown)


ragesoss:
A simple fix, nice! Thanks for documenting this well and explaining the change.

Issue #5446, opened 2023-07-04
Improve pagination controls on Articles tab


ragesoss:
Currently, if there are no edited articles, the Articles tab still renders the Prev and Next buttons, and says "Showing 1 - 0 of 0".

It also throws a console warning: `(react-paginate): The forcePage prop provided is greater than the maximum page index from pageCount prop (0 > -1).`

In the case of zero articles, it would be best to simply not render the controls or the "Showing" message.

For larger numbers, the "Showing message" is based on the total number of articles, and if a filter is used (eg, only showing New articles), the total does not update. This can be confusing because you can reach the final page but the "Showing" message implies that there ought to be more pages.

subhamkumarr:
I would like to work on this please assign it to me

PR #5450, opened 2023-07-08
Improve pagination controls on Articles tab

Abishekcs:
#5446

## What this PR does
Fixes:
1. When there are no edited articles, the Prev and Next buttons are now properly hidden.
2. The "Showing" message now updates correctly when a filter is applied.

## Screenshots
Before:

Before

[Before.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/fae6c885-560d-457c-996f-cf038af03fe3)

After:

After

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/5bed3efd-60ea-4815-8c89-aec172418a34)

### DIFF
--- a/app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
+++ b/app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
@@ -19,7 +19,8 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
   const filteredArticles = useSelector(getArticlesByTrackedStatus);
   const currentPage = useSelector(state => state.articles.currentPage);
   const pageRange = getPageRange(currentPage, filteredArticles.length);
-  const totalEditedArticles = useSelector(state => state.course.edited_count);
+  const totalEditedArticles = getPageRange(currentPage, filteredArticles.length).split('-')[1];
+
 
   // this is for when a filter is applied and the number of pages changes
   // we need to reset the page to the first page
@@ -31,7 +32,7 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
 
   return (
     <div id="articles-view-controls" className={limitReached ? 'hidden-see-more-btn' : ''}>
-      <ReactPaginate
+      { totalPages ? (<ReactPaginate
         pageCount={totalPages}
         nextLabel={nextLabel}
         previousLabel={previousLabel}
@@ -39,7 +40,7 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
         containerClassName={'pagination'}
         onPageChange={handlePageChange}
         forcePage={currentPage - 1}
-      />
+      />) : null }
       {!limitReached
         && (
           <button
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
```diff
@@ -19,7 +19,8 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
   const filteredArticles = useSelector(getArticlesByTrackedStatus);
   const currentPage = useSelector(state => state.articles.currentPage);
   const pageRange = getPageRange(currentPage, filteredArticles.length);
-  const totalEditedArticles = useSelector(state => state.course.edited_count);
+  const totalEditedArticles = getPageRange(currentPage, filteredArticles.length).split('-')[1];
```
ragesoss, line 22: I think this has the potential to be confusing, because the variable name in this case seems to be the total number of filtered articles rather than the total of all articles edited. Maybe a change to the variable name will make it more clear?

Also, I'm confused about the implementation here. Is there an easier way to get the right value than splitting the string?

Abishekcs: Okay i will change the variable name and as for the implementation, I have to check i am not sure if there is any other way to get the total number of filtered articles.

Abishekcs: I have changed the implementation too.

app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
```diff
@@ -19,7 +19,7 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
   const filteredArticles = useSelector(getArticlesByTrackedStatus);
   const currentPage = useSelector(state => state.articles.currentPage);
   const pageRange = getPageRange(currentPage, filteredArticles.length);
-  const totalEditedArticles = getPageRange(currentPage, filteredArticles.length).split('-')[1];
+  const totalFilteredArticles = Math.min(filteredArticles.length, currentPage * ARTICLES_PER_PAGE);
```
ragesoss, line 22: Why is this not just filteredArticles.length? In what circumstance would currentPage * ARTICLES_PER_PAGE be smaller than the length of `filteredArticles`?

Abishekcs: In case if there are more than 100 ` filteredArticles` for a particular filter and since only 100 articles can be shown per page only.
Also, i think i made a mistake here because then only  `ARTICLES_PER_PAGE` will be enough instead of  `currentPage * ARTICLES_PER_PAGE`. 
Then it will simply be  - `Math.min(filteredArticles.length,  ARTICLES_PER_PAGE)`.

ragesoss:
I'll have another look at this early next week. thanks @Abishekcs !

Abishekcs:
Okay :+1:  ty, @ragesoss

Abishekcs:
Okay, I will make the changes

Abishekcs:
@ragesoss , I have made update to this and would appreciate a review when you can. I think it's working as intended now as i have removed the `Math.min` an kept only the `filteredArtciles.length` for the `totalFilteredArticles`.
Screenshots:
1)
Screenshot from 2023-07-25 14-34-50
2)
Screenshot from 2023-07-25 14-34-32
3)
Screenshot from 2023-07-25 14-34-15
Screenshot from 2023-07-25 14-34-03

PR #5451, opened 2023-07-09
add tests for new scoping methods in course creation modal

TheTrio:
title

### DIFF
--- a/spec/features/scoping_methods_course_creation_spec.rb
+++ b/spec/features/scoping_methods_course_creation_spec.rb
@@ -14,6 +14,7 @@ describe 'Course creation for Article Scoped Programs', type: :feature, js: true
                                     completed_at: Time.zone.now)
     allow(Features).to receive(:wiki_ed?).and_return(false)
     allow(Features).to receive(:open_course_creation?).and_return(true)
+    allow(Features).to receive(:default_course_string_prefix).and_return('courses_generic')
     login_as(user, scope: :user)
 
     visit root_path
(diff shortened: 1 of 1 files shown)


spec/features/scoping_methods_course_creation_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+                  id: 1,
+                  permissions: User::Permissions::INSTRUCTOR)
+    create(:training_modules_users, user_id: user.id,
+                                    training_module_id: 3,
+                                    completed_at: Time.zone.now)
+    allow(Features).to receive(:wiki_ed?).and_return(false)
+    login_as(user, scope: :user)
+
+    visit root_path
+    click_link 'Create Course'
```
ragesoss, line 19: Where is this link? In P&E mode, the button on the home page should be called "Create an Independent Program" (which is also what the modal's header says once the course creator launches), right?

ragesoss:
I'm quite confused about why this would be failing, as this passes locally for me too.

ragesoss:
Okay, I figured it out. In the test environment, `Features.open_course_creation?` is returning false. This is because of how it's implemented in `Features`, which doesn't rely directly on `Features.wiki_ed?`. (I had to disable headless mode for the tests to see what was going on.)

If you stub `open_course_creation?` (and then adjust the test to work locally), I think it should be in working order.

spec/features/scoping_methods_course_creation_spec.rb
TheTrio: fixed this

PR #5466, opened 2023-07-28
Integration of Wikidata-Diff-Analyzer Gem

saha23s:
## What this PR does
This PR integrates https://rubygems.org/gems/wikidata-diff-analyzer gem in the Dashboard.

Approach:
Currently uses both the edit summaries(previous revisions)  and output of the gem (new revisions)
Using the summary field of the revision object to save stats from the gem after serializing

### DIFF
--- a/app/models/wiki_content/revision.rb
+++ b/app/models/wiki_content/revision.rb
@@ -86,4 +86,15 @@ class Revision < ApplicationRecord
     return 0 unless references_count(features) && references_count(features_previous)
     references_count(features) - references_count(features_previous)
   end
+
+  def edit_summary
+    return summary unless diff_stats # If diff_stats is nil, then summary is the edit summary
+    nil # Otherwise, it's diff_stats and returns nil
+  end
+
+  def diff_stats
+    JSON.parse(summary) if summary.present? && summary.start_with?('{', '[')
+  rescue JSON::ParserError
+    nil # Return nil if parsing fails (i.e., not diff_stats)
+  end
 end
--- a/app/models/wiki_content/revision.rb
+++ b/app/models/wiki_content/revision.rb
@@ -87,11 +87,20 @@ class Revision < ApplicationRecord
     references_count(features) - references_count(features_previous)
   end
 
+  # Generally, the summary field captured edit summary comment of an edit until August 2023
+  # This code is for a switch to save diff_stats instead(stats hash created from wikidata-diff-analyzer gem)
+  # used in update_wikidata_stats.rb
+  # These two methods find out if the content in summary field is a stats hash or not
+  # Edits made before August 2023 will be handled through WikidataSummaryParser and
+  # the later edits will be handled by stats collected from wikidata-diff-analyzer
+
   def edit_summary
     return summary unless diff_stats # If diff_stats is nil, then summary is the edit summary
     nil # Otherwise, it's diff_stats and returns nil
   end
 
+  # This function parses the serialized stats saved in the summary field, in case of any errors
+  # it returns nil meaning the field contains an edit summary
   def diff_stats
     JSON.parse(summary) if summary.present? && summary.start_with?('{', '[')
   rescue JSON::ParserError
--- a/app/models/wiki_content/revision.rb
+++ b/app/models/wiki_content/revision.rb
@@ -24,7 +24,7 @@
 #  features_previous :text(65535)
 #  summary           :text(65535)
 #
-
+require 'json'
 #= Revision model
 class Revision < ApplicationRecord
   belongs_to :user
--- a/app/models/wiki_content/revision.rb
+++ b/app/models/wiki_content/revision.rb
@@ -88,9 +88,8 @@ class Revision < ApplicationRecord
   end
 
   # Generally, the summary field captured edit summary comment of an edit until August 2023
-  # This code is for a switch to save diff_stats instead(stats hash created from wikidata-diff-analyzer gem)
-  # used in update_wikidata_stats.rb
-  # These two methods find out if the content in summary field is a stats hash or not
+  # This code is for a switch to save diff_stats instead(output hash generated from wikidata-diff-analyzer gem)
+  # These two methods find out if the content in summary field is a stats_hash or not
   # Edits made before August 2023 will be handled through WikidataSummaryParser and
   # the later edits will be handled by stats collected from wikidata-diff-analyzer
 
--- a/app/models/wiki_content/revision.rb
+++ b/app/models/wiki_content/revision.rb
@@ -88,7 +88,8 @@ class Revision < ApplicationRecord
   end
 
   # Generally, the summary field captured edit summary comment of an edit until August 2023
-  # This code is for a switch to save diff_stats instead(output hash generated from wikidata-diff-analyzer gem)
+  # This code is for a switch to save diff_stats instead(output hash generated from
+  # wikidata-diff-analyzer gem)
   # These two methods find out if the content in summary field is a stats_hash or not
   # Edits made before August 2023 will be handled through WikidataSummaryParser and
   # the later edits will be handled by stats collected from wikidata-diff-analyzer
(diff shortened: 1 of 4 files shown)


app/services/update_wikidata_stats.rb
```diff
@@ -21,10 +46,117 @@ def import_summaries
...
+    # crs_stat.save
+    stats_hash = crs_stat.stats_hash || {}
+
+    # Initialize arrays to store revisions with edit summaries and serialized stats
+    revisions_with_summary = []
+    revisions_with_serialized_stats = []
+
+    # Divide revisions based on edit summaries or serialized stats
+    course_revisions.each do |revision|
+      if revision.summary.present? && revision.summary.start_with?('{', '[')
```
ragesoss, line 60: I think that knowledge about how to handle the `summary` field should be part of the Revision object.

I would recommend implementing a pair of methods for Revision, like `edit_summary` and `diff_stats`, and `edit_summary` would return an edit summary if that's what is present, and otherwise return `nil`, while `diff_stats` would return the already already-parsed stats if present and otherwise return `nil`.

I think this `starts_with` check is going to work for all edits to Wikidata Item namespace (as well as Lexeme, etc), but there's a chance it could fail for an edit in userspace or another conventional wiki namespace where edit summaries are not created automatically (so someone could make an edit that actually starts with one of these characters even though it's just a string and parsing it would through an error).

saha23s:
I have moved the edit_summary and diff_stats check in the revision object, and uodated the main logical check based on if the summary is parsable or not. Is there any possibility that the edit summary is in complete JSON format and parsable? In that case, should I check more specifically the structure of the diff_stats because the structure would be always same? @ragesoss

saha23s:
@ragesoss currently, I am doing /manual_update for a course url to run UpdateCourseStats which is taking a long time, what's the efficient way to test if my code is successful at doing what it should do?

saha23s:
another question is related to `total revisions`. Currently, in wikidata summary parser, `stats['total revisions'] = stats.values.sum`. Will I do it for serialised stats as well? And will the total edits count be same as total revisons count on the page?

saha23s:
Local (Total Revision: 161)
image
Production (Total Revision: 112)
image
@ragesoss

ragesoss:
> I have moved the edit_summary and diff_stats check in the revision object, and uodated the main logical check based on if the summary is parsable or not. Is there any possibility that the edit summary is in complete JSON format and parsable? In that case, should I check more specifically the structure of the diff_stats because the structure would be always same? @ragesoss

It might be theoretically possible, but the way to reliably prevent such a situation would probably be to filter the revisions by wiki namespace before trying to do any parsing of summaries. At the least, no edits in the namespaces that you built the gem to analyze ought to have free-text edit summaries.

ragesoss:
> @ragesoss currently, I am doing /manual_update for a course url to run UpdateCourseStats which is taking a long time, what's the efficient way to test if my code is successful at doing what it should do?

There's no good way around that, except to choose example courses small enough that it doesn't take too too long to run an update.

ragesoss:
> another question is related to `total revisions`. Currently, in wikidata summary parser, `stats['total revisions'] = stats.values.sum`. Will I do it for serialised stats as well? And will the total edits count be same as total revisons count on the page?

Yes, you should include a total revisions count as well, although summing the values from the individual categories won't work anymore since now an edit could be counted towards more than one. The easier way would be to just count the number of revisions at the time they are passed in to the gem, and return that count. Or any similar way that relies on the number of revisions.

saha23s:
@ragesoss as I have tested with the courses, I think my PR is ready to be merged if wikidata-diff-analyzer is doing its job.

Can you suggest more ways to test out before merging? Should I write RSpec test cases for anything?

Also here are the differences between production and local.

Local:
image

Production:
image

There are differences in a couple of areas. Did you mention a way to figure out why the differences 
somehow breaking into the revisions in smaller chunks in our meeting? I don't remember exactly. Let me know.

app/services/update_wikidata_stats.rb
```diff
@@ -1,32 +1,132 @@
 # frozen_string_literal: true
-
 require_dependency "#{Rails.root}/lib/wikidata_summary_parser"
 require_dependency "#{Rails.root}/lib/importers/wikidata_summary_importer"
+# require the installed wikidata-diff-analyzer gem
+require 'wikidata-diff-analyzer'
 
 class UpdateWikidataStats
+  STATS_CLASSIFICATION = [
+    'claims created', 'claims removed', 'claims changed',
```
ragesoss, line 9: This will be easier to read if there's one entry per line (and you can also organize them into sections via comments). The closer this list matches up with the UI for how we display this data, the easier it will be to maintain both (as we potentially add new stats in the future).

app/models/wiki_content/revision.rb
```diff
@@ -86,4 +86,15 @@ def references_added
     return 0 unless references_count(features) && references_count(features_previous)
     references_count(features) - references_count(features_previous)
   end
+
+  def edit_summary
```
ragesoss, line 90: I suggest adding a longer comment to this file explaining the switch from edit summaries to diff stats (and the approximate date of that switchover), along with the strategy of using the `summary` field to store diff stats. This is important context that will be helpful for developers who work on anything related to this in the future.

app/services/update_wikidata_stats.rb
```diff
@@ -1,32 +1,132 @@
...
     @course = course
-    import_summaries
+    # import_summaries
+    update_summary_with_stats
     update_wikidata_statistics
   end
 
   private
 
+  # When summary is nil, instead of fetching edit summaries for each revision, I want to use
```
ragesoss, line 37: Comments should not be written in first-person, unless you're signing them to explicitly indicate who "I" is. (In contrast, I think first-person comments are fine in commit descriptions, as those are explicit signed and it's useful to understand the personal perspective of the committer.)

app/services/update_wikidata_stats.rb
```diff
@@ -1,32 +1,132 @@
...
+    revision_ids = wikidata_revisions_without_summaries.pluck(:mw_rev_id)
+    analyzed_revisions = WikidataDiffAnalyzer.analyze(revision_ids)[:diffs]
+
+    revision_ids.each do |rev_id|
+      individual_stat = analyzed_revisions[rev_id]
+      # Serialize the individual_stat to JSON format
+      serialized_stat = individual_stat.to_json
+
+      # Find the revision object by mw_rev_id
+      revision = course_revisions.find_by(mw_rev_id: rev_id)
```
ragesoss, line 55: I think doing a 'find' and 'update' on each revision may be inefficient. Since we've already loaded all the revisions in `wikidata_revisions_without_summaries`, and since the `analyzed_revisions` hash has the IDs as keys already, I think it makes more sense to loop through `wikidata_revisions_without_summaries` and look up the relevant data for each Revision in `analyzed_revisions` via the mw_rev_id. That would cut out the `find_by` database query, and if you wrap the whole loop in `Revision.transaction` (see other examples of `.transaction` in this codebase), it will do all the updates as one database transaction.

ragesoss:
I'm excited about this, and I think the relative simplicity of the PR is an indication that this was probably a good strategy. I will test it out manually with a few courses. I left some refactoring and cleanup comments, but overall this looks like it will work. Great job!

app/services/update_wikidata_stats.rb
saha23s: @ragesoss Currently the order of the entries is based on the structure of the output of wikidata-diff-analyzer because that helps while making the stats hash according to the same order. I have tried to add comments and make each entry in one line to make it easier to handle.

ragesoss:
Can you provide updated screenshots? Rather than comparing between production and your version, it might be better to compare between local on master versus local with your version. That should clear up the view count differences and such. (Also, I think you fixed some things since the screenshots, so I'm not sure what is currently different in terms of the statistics.)

saha23s:
That's a great idea. I would do that.

For http://localhost:3000/courses/Wiki_Education/Wikidata_Institute-May-2023_(Spring_2023)

My Branch:
image

Master Branch: (after dropping and creating the new database)
image

For http://localhost:3000/courses/Wiki_Education/Wikidata_Institute-May-2023_(Spring_2023)

My Branch: 
If I run manual update on completely newly created course on my branch I get this
image

Master Branch:
image

I think my code is doing good job

ragesoss:
Yeah, it looks like it!

I'm satisfied. I'll have another look and do a little more testing, but I think it's ready to merge and see how it performs in production.

PR #5470, opened 2023-08-01
Add fallback check for when no contributions are highlighted in ArticleViewer

### LINKED ISSUE #5461: Add fallback check for when no contributions are highlighted in ArticleViewer
### Current Behavior:
The ArticleViewer's authorship highlighting feature shows which parts of the current version of an article were originally added by course participants. This is reliable for highlighting prose contributions, but typically does not work for cases where the wikitext is exclusively templates (including ref templates). I think this is because the WhoColor system works by adding span tags around each token in the wikitext to embed authorship metadata, then sends it through Wikipedia for rendering to HTML, but templates and other specially-rendered markup elements strip such tags from the rendered output.

Here's an example where two users made contributions (both of which are still present in the live article as of July 26, 2023): https://dashboard.wikiedu.org/courses/UCSF/Foundations_II_(Summer_2023)/articles/edited?showArticle=52655504

In this example, user CarUCSF2025 added a reference, but the ArticleViewer detects no contributions.

### Desired Behavior:

Since the Dashboard already knows which users made edits to an article (at least, when ArticleViewer is used in the context of the Articles tab of a course), we could do a fallback check against the WikiWho API (for example https://wikiwho-api.wmcloud.org/en/api/v1.0.0-beta/latest_rev_content/Atrial_fibrillation/?o_rev_id=true&editor=true&token_id=true&out=true&in=true )to get the wikitext metadata of the current revision to see if a user who doesn't have any attributed HTML contributions actually does still have extant wikitext contributions or not. Then, instead of the current behavior where that user just doesn't get a highlight color, we could show a mouseover or similar message for that user to indicate the situation, eg, "There are contributions from this user in the latest version of the article, but they cannot be highlighted. This usually means that the contributions consist of references or templates, which can't be accurately attributed in HTML."

We only want to make an extra request like that when the first WhoColor request doesn't return any attributed content for one of the course participants who edited that article.

Abishekcs:
resolves #5461 

## What this PR does
* It Adds a fallback check for when no contributions are highlighted in ArticleViewer.
* **Bug-Fix** - It also Fixes a `CSS bug` in the ArticleViewer  when a `popover` is supposed to be shown for when the contributions made by the editor is not  `highlighted `  or  when the editor doesn't have a extant wikitext contributions in the `wikitext metadata` of the current version.

## Changes Made
- **article_viewer.jsx :**
 a) The primary changes in this file are related to the addition of the `fallback mechanism` for users whose contributions could not be highlighted. 
 - **article_viewer_api.js :**
    a) In this file, the method `fetchWikitextMetaData` has been added to the ArticleViewerAPI class. This method is responsible for fetching the wikitext metadata.
   
- **article_viewer_legend.jsx** :
     a) The unhighlightedContributors prop has been added to the ArticleViewerLegend component.
     b) CSS has been added `popup-style` to fix the `Bug` related to the `popup`.

## Screenshots
Before:
a) Related to the Bug: 

[Before Popover Bugfix.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/c39f16f5-1fea-4afa-91e3-32ce369390a9)

b) Before Adding a fallback check

[Before(Enhancement).webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/dca20267-a590-4cc1-ba90-4696d441b47f)

After:
a) After Bugfix

[After Bugfix.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/235146c6-78a8-4dae-92d8-704002882b31)

b)After Adding a fallback check

[After-enhancement.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/e42dde0b-9d99-4627-9409-8042fd947266)

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
@@ -52,6 +52,7 @@ const ArticleViewer = ({ showOnMount, users, showArticleFinder, showButtonLabel,
   const [userIdsFetched, setUserIdsFetched] = useState(false);
   const [whoColorHtml, setWhoColorHtml] = useState(null);
   const [parsedArticle, setParsedArticle] = useState(null);
+  const [unhighlightedContributors, setUnhighlightedContributors] = useState([]);
 
   const dispatch = useDispatch();
   const ref = useRef();
@@ -170,9 +171,29 @@ const ArticleViewer = ({ showOnMount, users, showArticleFinder, showButtonLabel,
       const styledAuthorSpan = `<span title="${user.name}" class="editor-token token-editor-${user.userid} ${colorClass}`;
       const authorSpanMatcher = new RegExp(`<span class="editor-token token-editor-${user.userid}`, 'g');
       html = html.replace(authorSpanMatcher, styledAuthorSpan);
-      if (prevHtml !== html) user.activeRevision = true;
+      if (prevHtml !== html) {
+          user.activeRevision = true;
+      } else {
+        // If highlighting failed, verify if the user's has contributions in the article using checkFallback.
+        checkFallback(user);
+      }
     });
     setHighlightedHtml(html);
+  };
+
+  const checkFallback = (user) => {
+    const builder = new URLBuilder({ article: article });
+    const api = new ArticleViewerAPI({ builder });
+    api.fetchWikitextMetaData()
+       .then((response) => {
+        const { tokensForRevision } = response;
+        const foundToken = tokensForRevision.find(token => token.editor = user.userid);
+        // If the user's revision is found (i.e., the user has contributions in the article),
+        // update the unhighlightedContributors state by adding the user's userid to the array.
+        if (foundToken) {
+          setUnhighlightedContributors(x => [...x, user.userid]);
+      }
+    });
   };
 
   const fetchParsedArticle = () => {
@@ -302,6 +323,7 @@ const ArticleViewer = ({ showOnMount, users, showArticleFinder, showButtonLabel,
           showArticleFinder={showArticleFinder}
           whoColorFailed={whoColorFailed}
           users={usersState}
+          unhighlightedContributors={unhighlightedContributors}
         />
       </div>
     </div>
--- a/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
@@ -52,7 +52,7 @@ const ArticleViewer = ({ showOnMount, users, showArticleFinder, showButtonLabel,
   const [userIdsFetched, setUserIdsFetched] = useState(false);
   const [whoColorHtml, setWhoColorHtml] = useState(null);
   const [parsedArticle, setParsedArticle] = useState(null);
-  const [unhighlightedContributors, setUnhighlightedContributors] = useState([]);
+  const [unhighlightedEditors, setUnhighlightedEditors] = useState([]);
 
   const dispatch = useDispatch();
   const ref = useRef();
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
```diff
@@ -170,11 +171,31 @@ const ArticleViewer = ({ showOnMount, users, showArticleFinder, showButtonLabel,
...
+          user.activeRevision = true;
+      } else {
+        // If highlighting failed, verify if the user's has contributions in the article using checkFallback.
+        checkFallback(user);
+      }
     });
     setHighlightedHtml(html);
   };
 
+  const checkFallback = (user) => {
```
ragesoss, line 184: As currently implemented, it looks like this would be run multiple times if multiple users edited the article but had no highlighted contributions. The wikitext metadata should only need to be fetched once at most.

app/assets/javascripts/components/common/article_viewer_legend.jsx
```diff
@@ -25,23 +24,28 @@ const ArticleViewerLegend = ({ article, users, colors, status, allUsers, failure
 
       setUserLinks(users.map((user, i) => {
         let res;
+        // The 'unhighlightedContributors' array keeps track of the userids of users whose contributions
+        // were not successfully highlighted in the article viewer.
+        const isUserHovered = unhighlightedContributors.find(x => x === user.userid);
```
ragesoss, line 29: It looks like this constant is actually keeping track of whether the user had any contributions attributed in the `unhighlightedContributors` array. I don't think `isUserHovered` is a good name for that.

app/assets/javascripts/components/common/article_viewer_legend.jsx
```diff
@@ -25,23 +24,28 @@ const ArticleViewerLegend = ({ article, users, colors, status, allUsers, failure
...
         const userLink = UserUtils.userTalkUrl(user.name, article.language, article.project);
         const fullUserRecord = allUsers.find(_user => _user.username === user.name);
         const realName = fullUserRecord && fullUserRecord.real_name;
         if (status === 'loading') {
           res = <div key={`legend-${user.name}`} className={'user-legend'}><a href={userLink} title={realName} target="_blank">{user.name}</a></div>;
         } else if (user.activeRevision === true) {
           res = <div key={`legend-${user.name}`} className={`user-legend ${colors[i]}`}><a href={userLink} title={realName} target="_blank">{user.name}</a><img className="user-legend-hover" style={{ color: 'transparent' }} src="/assets/images/arrow.svg" alt="scroll to users revisions" width="30px" height="20px" onClick={() => Scroller.scrollTo(user.name, scrollBox)} /></div >;
+        } else if (isUserHovered) {
```
ragesoss, line 37: This will be false while the fallback request is being made, so during that request, a user will have the no_highlighting message even if they actually have contributions and are about to get the contributionsNotHighlighted message. Ideally, instead of that, it would wait until the fallback request is completed before adding either message. I'm okay with cutting that corner if it's hard to implement, though. Shouldn't be too much of a problem for users.

app/assets/javascripts/components/common/ArticleViewer/containers/ArticleViewer.jsx
ragesoss: This function name is kind of vague. Can you think of a more precise one? A comment explaining the purpose of this would also be useful.

app/assets/javascripts/components/common/article_viewer_legend.jsx
Abishekcs: Okay will change variable name

Abishekcs:
Changes:

*  Changed the  `checkFallback ` function  name to `usersContributionExists` and
* Made the `Wikitextmetadata `to be fetched only once for multiple users of the same article.
* Changed  `isUserHovered` variable name to `UnhighlightedContributions`.
* Made the `HighlightedMessage` appear after only loading  of the authorship data is ready  depending on the `unhighlightedEditors` and ` highlightedHtml` in the " footer component".  So, that only the  `contributionsNotHighlighted`  or `no_highlighting` message appears for the user.

PR #5471, opened 2023-08-01
Improve UI for modifying scoping methods after course has been created

TheTrio:
This unifies the interface used for interacting with the various scoping methods when the course is being created and after it has been created.

Looks like

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/10794178/de2bf5a2-d3dd-4d19-b5e2-c55d1b33703b


ragesoss:
Code-wise, this looks good to me. Please add a before/after screenshot or video, and fix the linting error.

PR #5472, opened 2023-08-01
Fix Default Sorting of Students List and Retain Sorting State

### LINKED ISSUE #5343: Default sorting of Students list is not ideal, and resorting state is not retained when navigating between tabs

Abishekcs:
resolves #5343 

### What this PR does
<hr>

**FixBug:**

On the Students (aka Editors) tab of a course page, the default sorting of students was not ideal, as they were listed based on the order of joining the course. The default sorting method was not considered optimal, and the sorting state was not retained when users navigated between tabs and data refresh occurs (which happens when you change course tabs and it's been more than a minute since the last fetch).

**Solution:**

* Updated the default sorting to be based on the student's  `last names` when receiving users' data for the first time.
* Implemented sorting retention when users manually re-sort the list. The sorting state is now preserved when users navigate between tabs or when data gets refreshed.

**Changes Made:**

* Modified the users reducer `user.js` to handle sorting of students based on `last name` when receiving users' data for the first time.
* If the `state.users` array is empty, check if any user has a `real_name` in the received data. If so, transform the user list to include `first_name` and `last_name` properties based on `real_name`.
* If `state.users` is empty and no user has `real_name`, sort the user list by `username`.
* Retained sorting state by using the existing users in the state in the Redux store when applicable.

### **Screenshots**

Before:

[Before.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/efe7e22e-7c72-469a-954f-14fd4c8fd619)

After: 

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/7ccf84ef-e58d-4712-b7d8-e31d16ce2dee)

### DIFF
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -1,5 +1,5 @@
 import { RECEIVE_USERS, SORT_USERS, ADD_USER, REMOVE_USER } from '../constants';
-import { sortByKey } from '../utils/model_utils';
+import { sortByKey, transformUsers } from '../utils/model_utils';
 
 const initialState = {
   users: [],
@@ -33,12 +33,9 @@ export default function users(state = initialState, action) {
 
        // Check if the user list is empty in the state and if any user in the action data has 'real_name'
        if (!state.users.length && action.data.course.users.some(user => user.real_name)) {
-         // Perform data transformation on the user list: Split 'real_name' into 'first_name' and the rest of the name ('rest')
-         user_list = action.data.course.users.map((user) => {
-           const [first_name, ...rest] = (user.real_name?.trim().toLowerCase() || '').split(' ');
-            // Return the user object with updated 'first_name' and 'last_name' properties
-           return { ...user, first_name, last_name: rest.join(' ') };
-         });
+        // If any users have 'real_name', transform the 'real_name' into separate
+        // 'first_name' and 'last_name' properties and update the user list
+        user_list = transformUsers(action.data.course.users);
        } else if (!state?.users.length) {
          // If there are no users in the state and none of the users in the action data have 'real_name',
         // set the sorting key to 'username'
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -24,7 +24,7 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-       // Transform the 'real_name' for users in 'action' into separate 'first_name'
+      // Transform the 'real_name' for users in 'action' into separate 'first_name'
       // and 'last_name' properties if 'real_name' is available by using transformUsers.
       let user_list = transformUsers(action.data.course.users);
 
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -6,7 +6,7 @@ const initialState = {
   sort: {
     sortKey: null,
     key: null,
-    previousKey: null,
+    previousSortKey: null,
   },
   isLoaded: false,
   lastRequestTimestamp: 0 // UNIX timestamp of last request - in milliseconds
@@ -67,7 +67,7 @@ export default function users(state = initialState, action) {
           sortKey: sorted.newKey,
           key: action.key,
           // store the previous sortKey to use if the instructor switches between tabs and comes back to the students/editors tab
-          previousKey: state.sort.sortKey,
+          previousSortKey: state.sort.sortKey,
         }
       };
     }
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -24,8 +24,8 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-      // Transform the 'real_name' in 'action' into separate 'first_name' and 'last_name' properties
-      // if 'real_name' is available by using transformUsers.
+       // Transform the 'real_name' for users in 'action' into separate 'first_name'
+      // and 'last_name' properties if 'real_name' is available by using transformUsers.
       let user_list = transformUsers(action.data.course.users);
 
       // Get the sorting 'key' if available from Redux store or else use last_name or
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -24,7 +24,7 @@ const SORT_DESCENDING = {
 export default function users(state = initialState, action) {
   switch (action.type) {
     case RECEIVE_USERS: {
-      // Get the sorting key from if available from Redux store or else use last_name
+      // Get the sorting key if available from Redux store or else use last_name
       let sort_key = state.sort.key || 'last_name';
       let previousKey = null;
       // Initialize a variable to hold the user list.
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -6,7 +6,6 @@ const initialState = {
   sort: {
     sortKey: null,
     key: null,
-    previousSortKey: null,
   },
   isLoaded: false,
   lastRequestTimestamp: 0 // UNIX timestamp of last request - in milliseconds
@@ -27,7 +26,7 @@ export default function users(state = initialState, action) {
     case RECEIVE_USERS: {
       // Get the sorting key from if available from Redux store or else use last_name
       let sort_key = state.sort.key || 'last_name';
-
+      let previousKey = null;
       // Initialize a variable to hold the user list.
       let user_list = action.data.course.users;
 
@@ -41,8 +40,10 @@ export default function users(state = initialState, action) {
           sort_key = 'username';
         }
 
+        if (state.sort.key) { previousKey = state.sort.sortKey ? null : sort_key; }
+
         // Sort the 'user_list' array based on the 'sort_key'
-        user_list = sortByKey(user_list, sort_key, state.sort.previousSortKey, SORT_DESCENDING[sort_key]);
+        user_list = sortByKey(user_list, sort_key, previousKey, SORT_DESCENDING[sort_key]);
     return {
       ...state,
       users: user_list.newModels, // Update 'users' with the sorted user list.
@@ -66,8 +67,6 @@ export default function users(state = initialState, action) {
         sort: {
           sortKey: sorted.newKey,
           key: action.key,
-          // store the previous sortKey to use if the instructor switches between tabs and comes back to the students/editors tab
-          previousSortKey: state.sort.sortKey,
         }
       };
     }
--- a/app/assets/javascripts/reducers/users.js
+++ b/app/assets/javascripts/reducers/users.js
@@ -5,7 +5,7 @@ const initialState = {
   users: [],
   sort: {
     sortKey: null,
-    key: null,
+    key: 'last_name',
   },
   isLoaded: false,
   lastRequestTimestamp: 0 // UNIX timestamp of last request - in milliseconds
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +23,38 @@ const SORT_DESCENDING = {
...
+    case RECEIVE_USERS: {
+      // Get the sorting key from the Redux store
+      let sort_key = state.sort.key;
+
+       // Initialize a variable to hold the user list. If there are existing users in the state,
+      //  use that list; otherwise, use the user list from the action data.
+      let user_list = state.users.length ? state.users : action.data.course.users;
+
+       // Check if the user list is empty in the state and if any user in the action data has 'real_name'
+       if (!state.users.length && action.data.course.users.some(user => user.real_name)) {
```
ragesoss, line 35: I think this transformation ought to be extracted to a separate pure function (which can live in model_utils, along with the sortByKey function).

In addition to making the reducer logic easier to understand, this would let us test the sorting behavior in a more isolated way, which I think is important for this kind of thing where we may discover sorting edge cases over time that we need to handle and test for regressions.

Abishekcs: Okay sure :)

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +23,35 @@ const SORT_DESCENDING = {
...
 export default function users(state = initialState, action) {
   switch (action.type) {
-    case RECEIVE_USERS: return {
+    case RECEIVE_USERS: {
+      // Get the sorting key from the Redux store
+      let sort_key = state.sort.key;
+
+       // Initialize a variable to hold the user list. If there are existing users in the state,
+      //  use that list; otherwise, use the user list from the action data.
+      let user_list = state.users.length ? state.users : action.data.course.users;
```
TheTrio, line 32: Are you sure this is correct? Why would you want to read the users from the state when we are receiving users?

TheTrio:
I'm really confused as to why the code has a bunch of uses of `state.users.length`. Currently(that is, on production), we read only from the action data - which makes sense since this is the reducer for receiving users. Why should the existing state be involved at all?

Abishekcs:
> I'm really confused as to why the code has a bunch of uses of `state.users`. Currently(that is, on production), we read only from the action data - which makes sense since this is the reducer for receiving users. Why should the existing state be involved at all?

hmm Since "The `state.users` will help retain the sorting order when instructors sort the user list, either by using the sorting dropdown menu at the top of the Students list or by clicking on one of the headers in the Students tab. If the instructor chooses to sort explicitly and then switches between tabs, the same sorted list will be retained. This is done using the list already available in the previous users' state. If the instructor didn't explicitly sort the list, `action.data.course.users` will be sufficient. This is because the user list in the state will be empty, and `action.data.course.users` will be chosen by default. While the sorting can also be retained using the data from `action.data.course.users`, I thought it's better to use the sorted user list from the state. This avoids resorting and utilizes the sorting state already stored in the `usersState`."

TheTrio:
I still don't understand. `state.users` is not necessary equal to `action.data.course.users`. That's the reason the refetching logic exists in the first place - to fetch newer data. If I understand your code correctly, if there's an update and the server sends back the updated users list, it would simply be ignored - because you give priority to the users list present in the cache.

Abishekcs:
> I still don't understand. `state.users` is not necessary equal to `action.data.course.users`. That's the reason the refetching logic exists in the first place - to fetch newer data. If I understand your code correctly, if there's an update and the server sends back the updated users list, it would simply be ignored - because you give priority to the users list present in the redux store.

Oh ty for pointing that out i guess then i have to add an condition to check if there's an update and then use the `action.data.course.users` instead of the `state.users`. I think then it will work properly?

TheTrio:
I think simply sorting the data present in the action is fine. There's no need to over-optimize this. Just make use of the sort key - that should be equal to the property the user used.

Abishekcs:
Okay that Sounds much better will do that.

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +24,32 @@ const SORT_DESCENDING = {
 
 export default function users(state = initialState, action) {
   switch (action.type) {
-    case RECEIVE_USERS: return {
+    case RECEIVE_USERS: {
+      // Get the sorting key from if available from Redux store or else use last_name
+      let sort_key = state.sort.key || 'last_name';
```
TheTrio, line 29: where is this being used? If `state.sort.key` is null, we set `sort_key` to `username` [here](https://github.com/Abishekcs/WikiEduDashboard/blob/FixSortingStudents/app/assets/javascripts/reducers/users.js#L41) What's the use of setting it to `last_name` here?

app/assets/javascripts/reducers/users.js
```diff
@@ -1,11 +1,12 @@
 import { RECEIVE_USERS, SORT_USERS, ADD_USER, REMOVE_USER } from '../constants';
-import { sortByKey } from '../utils/model_utils';
+import { sortByKey, transformUsers } from '../utils/model_utils';
 
 const initialState = {
   users: [],
   sort: {
     sortKey: null,
     key: null,
+    previousSortKey: null,
```
TheTrio, line 9: What's this for? `sortKey` should be enough

app/assets/javascripts/reducers/users.js
```diff
@@ -44,7 +65,9 @@ export default function users(state = initialState, action) {
         users: sorted.newModels,
         sort: {
           sortKey: sorted.newKey,
-          key: action.key
+          key: action.key,
+          // store the previous sortKey to use if the instructor switches between tabs and comes back to the students/editors tab
```
TheTrio, line 69: why is this required?

TheTrio (changes requested):

app/assets/javascripts/reducers/users.js
TheTrio, line 29: setting it to `last_name` gets overridden by the assignment at L41 - what's it for?

app/assets/javascripts/reducers/users.js
```diff
@@ -44,7 +65,9 @@ export default function users(state = initialState, action) {
...
+          previousSortKey: state.sort.sortKey,
```
TheTrio, line 70: I'm not sure why this is required. Why isn't `sortKey` enough?

Abishekcs:
> this isn't required - `storyKey` should be enough
>I'm not sure why this is required. Why isn't sortKey enough?

:sweat_smile: I tried to use the `sortKey` but if any one sort by clicking (First Click) one of the headers in the Students tab. Then again sort by by clicking (Second Click) the same header then it will reverse the sorting and return `sort.newKey` as null.
Which is not a problem but if the user switches between tabs after `reverse sorting` and comes back to the students/editors tab then there is no way to know the old value of `sort.newKey` since its set to null and we won't know if the `sorting was done in reverse`. That's why i used `previousSortKey` to know the previous sort key in case of `RECEIVE_USERS`.

[2023-08-09 11-15-02.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/619d0d6e-9200-4e7f-b443-f7ddf7fe7c12)

app/assets/javascripts/reducers/users.js
Abishekcs: It does , but in case `if` (L35) statements is true which is if even one user have `real_name`  then it wont be overridden. As for the `last_name` it is used if  any user have real_name or else sorting will fall back to `username`.

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +23,34 @@ const SORT_DESCENDING = {
...
+       if (user_list.some(user => user.real_name)) {
+         // If any users have 'real_name', transform the 'real_name' into separate
+         // 'first_name' and 'last_name' properties and update the user list
+         user_list = transformUsers(user_list);
+        } else if (!state.sort.key) {
+          // If there are no users with real_name and key in the store is null then set sort_key by username
+          sort_key = 'username';
+        }
+
+        if (state.sort.key) { previousKey = state.sort.sortKey ? null : sort_key; }
```
TheTrio, line 43: can you add a comment here explaining why/how this works?

Abishekcs: Okay sure

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +23,45 @@ const SORT_DESCENDING = {
...
+      */
+      let previousKey = null;
+      // Initialize a variable to hold the user list.
+      let user_list = action.data.course.users;
+
+      // Check if any user in the user_list has 'real_name'
+      if (user_list.some(user => user.real_name)) {
+        // If any users have 'real_name', transform the 'real_name' into separate
+        // 'first_name' and 'last_name' properties and update the user list
+        user_list = transformUsers(user_list);
```
TheTrio, line 42: why is this if check even required? If `real_name` doesn't exist, `transformUsers` shouldn't change anything.

TheTrio (approved):
LGTM, left a minor comment

app/assets/javascripts/utils/model_utils.js
```diff
@@ -61,3 +61,13 @@ export const sortByKey = (models, sortKey, previousKey = null, desc = false, abs
   }
   return { newModels, newKey };
 };
+
+// transformUsers: Transforms user objects by splitting 'real_name' into 'first_name' and 'last_name'
+export const transformUsers = (users) => {
```
ragesoss, line 66: I suggest adding tests for this function, so we can show what it puts out for various types of real names as well as users without a real_name.

It should probably also return the original user object rather than a new one in cases where no real_name is present.

Abishekcs: Okay i will add the test

app/assets/javascripts/reducers/users.js
```diff
@@ -23,12 +23,42 @@ const SORT_DESCENDING = {
 
 export default function users(state = initialState, action) {
   switch (action.type) {
-    case RECEIVE_USERS: return {
+    case RECEIVE_USERS: {
+      // Get the sorting key if available from Redux store or else use last_name
```
ragesoss, line 27: The sorting logic for this is still quite hard to understand. That's mostly because of how we don't actually store sorting direction in state, we just compare the new sorting from the SORT_USERS action with the not-yet-updated state.

I think this needs to be refactored to make it easier to follow, and to be more explicit about whether to do reverse sorting.

I suggest:
* handling `user_list` first.
* next, determine the sort key (`sort.key` or `last_name` or `username`).
* next, determine whether the previous sorting was a reversed sort or not, via an explicitly named boolean. (It was reversed if `sort.key` is different from `sort.sortKey`, I think.
* finally, do the sortByKey operation and use the `isReversed` (or whatever name) boolean to decide whether to pass the sort key or `null` as the `previousKey` argument. It will do reverse sorting if the same value is passed for both sortKey and previousKey.

Abishekcs: Okay

PR #5482, opened 2023-08-11
Add multi-search functionality to tickets system

### LINKED ISSUE #5481: Ticket system search feature should allow for combining multiple types of search
### Current Behavior:
The Ticket system has a search feature that supports search by Subject, Email/Username, Body, and Course. However, it's not possibly to combine search types, for example to search for tickets with "plagiarism" in the subject that were for a specific user.

### Desired Behavior:
Allow searching by multiple modes at once to narrow down the search more specifically.

Aminehassou:
closes #5481 
## What this PR does
This PR adds the ability to utilize multiple search types at the same time when searching for tickets. Now users should be able to use all 4 search types at the same time.

I've gone ahead and removed the select fields that were being used before. Instead, we now have 4 inputs fields and 2 buttons, the "clear search" buttons clears the current search and makes the filters available to the user again. The "search tickets" button queries the tickets and returns the result.

I have also added a few UI tweaks such as making the loading animation only take up the tickets table and not the entire page. And making the "search tickets" button greyed out when all the input fields are empty

## Videos
Before:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/c91716ae-91c3-4326-b3d9-e59b475c8f94

After:
Main search functionality:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/a1d045e2-db40-4a74-b11f-463b8e278177

Clearing search and getting all course tickets:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/39999898/8842593e-ab54-425f-8eff-90301f3b3670

### DIFF
--- a/lib/tickets/ticket_query_object.rb
+++ b/lib/tickets/ticket_query_object.rb
@@ -3,60 +3,50 @@
 class TicketQueryObject
   def initialize(params)
     @search = params[:search]
-    @what = params[:what]
+    @whats = params[:what]
     @offset = params[:offset] || 0
     @limit = params[:limit] || 100
   end
 
   def search
-    case @what
-    when 'by_email_or_username'
-      search_by_username_or_by_email
-    when 'in_subject'
-      search_by_subject
-    when 'in_content'
-      search_by_content
-    when 'by_course'
-      search_by_course
+    query = tickets
+
+    @whats.each do |what|
+      case what
+      when 'by_email_or_username'
+        query = query.merge(search_by_username_or_by_email)
+      when 'in_subject'
+        query = query.merge(search_by_subject)
+      when 'in_content'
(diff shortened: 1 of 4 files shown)


lib/tickets/ticket_query_object.rb
```diff
@@ -2,61 +2,58 @@
...
-    when 'in_content'
-      search_by_content
-    when 'by_course'
-      search_by_course
+    query = tickets
+
+    @whats.each do |key, value|
+      case key
+      when :by_email_or_username
+        query = query.merge(search_by_username_or_by_email(value))
```
ragesoss, line 22: This is an interesting strategy, using `merge` to combine the query parameters. Nice!

PR #5495, opened 2023-10-04
System for discouraging assignment of specific articles

### LINKED ISSUE #5477: Create a system for discouraging assignment of specific articles
### Current Behavior:
Some specific Wikipedia articles — like [[Boba]] — are subject to repeated assignment by students, but rarely lead to significant improvement. We don't have a way to discourage a student from choosing one of these articles to work on, and it annoys the editors of such articles.

### Desired Behavior:
We need a way that both Wiki Education staff and other Wikipedia editors can mark a specific articles such that it either cannot be assigned for a Wiki Education course at all, or in a way that a student will be actively discouraged from choosing that article (but can still choose to work on it if they acknowledge the warning first).

### Additional context:

One way to implement this would be to use one or two hidden categories on Wikipedia, and have the Dashboard collect the articles in those categories and check against them whenever a new assignment is created.

Abishekcs:
closes #5477

## What this PR does

This pull request addresses the issue documented in #5477. It introduces a feature that displays a `warning message` to students and administrators when they select a discouraged article/assignment.

**Affected Pages:**

* `/courses/[institution_name]/[course_name]/articles/available` 
* `/courses/[institution_name]/[course_name]/article_finder` 
* `/courses/[institution_name]/[course_name]/students/articles/[user]`
* `/courses/[institution_name]/[course_name/home`

## Screenshots
**Before:**

For Admin :

[Before(Admin).webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/2e01f444-41b9-4ee5-95fe-5bbeae92786c)

For Student :

[Before(Student).webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/6cbe2048-0bdf-4383-87fa-3c4070e1f214)

**After:**

For Admin:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/7cbdafcf-bd7f-4e83-9216-6d895e83bca3

For Student:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/36dcf871-7ef5-429f-8df5-5d1a73e28258

**Note:**

For testing a category from wikipedia set `blocked_assignment_category: [Category Name]`  in `application.yml` (used in assignment_manager.rb) and also run `dailyupdate`


app/assets/javascripts/components/article_finder/article_finder_row.jsx
```diff
@@ -64,6 +67,18 @@ const ArticleFinderRow = createReactClass({
     return this.props.deleteAssignment(assignment);
   },
 
+  async isCategoryMember(userId = null) {
```
pppery, line 70: This should not be called "isCategoryMember" - that implies it returns a boolean rather than doing things.

app/assets/javascripts/components/articles/add_available_articles.jsx
```diff
@@ -33,6 +36,18 @@ const AddAvailableArticles = createReactClass({
     this.props.open();
   },
 
+  async handleClick() {
```
pppery, line 39: Duplicate code. Should be written out only once if possible.

app/assets/javascripts/components/articles/available_article.jsx
```diff
@@ -1,28 +1,39 @@
...
     const successNotification = {
       message: I18n.t('assignments.article', { title }),
       closable: true,
       type: 'success'
     };
-
-    return dispatch(claimAssignment(assignmentObj, successNotification));
+    if ((await API.isCategoryMember(title))[0] !== null) {
+      // need to add to en.yml
+      const confirmMessage = 'Warning: Assigning a Discouraged Article. You are attempting to assign an article that has been marked as discouraged. Please confirm if you want to proceed.';
```
pppery, line 29: Triplicate code?

app/assets/javascripts/components/article_finder/article_finder_row.jsx
Abishekcs: okay i will make the changes

app/assets/javascripts/components/articles/add_available_articles.jsx
Abishekcs: okay ty for pointing this out

app/assets/javascripts/components/articles/available_article.jsx
Abishekcs: @pppery Sorry, but I didn't exactly understand what you meant by 'triplicate code' here. 😅

pppery: Sorry for missing this comment. It means the same thing as "duplicate code", except you have three copies rather than two.

Abishekcs: Okay, got it. ty  :+1:

pppery (commented):
I see nothing obviously wrong with this code, but I'm not an expert here.

Abishekcs:
> I see nothing obviously wrong with this code, but I'm not an expert here.

okay np.

Abishekcs:
@ragesoss, your review would be greatly appreciated whenever you have the time.

ragesoss:
Thanks for taking on such a big issue @Abishekcs!

I think I would prefer a somewhat different approach. The first major suggestion I have is to rely on the existing `Category` table rather than create a new record type in the database. These records are already designed to query Wikipedia for the titles in a category and save them, so I don't think there will be a need for any database changes.

I also think that the majority of the logic for this feature should be on the server.  `AssignmentManager` is the natural place to handle checking for whether a new assignment is one that ought to be blocked or requires a warning confirmation.

I suggest breaking it up into two parts, beginning with implementing a system that just completely prevents creating assignments that are in a certain category. This would be relatively easy to handle in terms of data flow: AssignmentManager should check against the category, raise a specific error (similar to the existing DuplicateAssignmentError), and then AssignmentsController should rescue that error and return a good error message and error status to the frontend. (The Category will need to be routinely refreshed, which can be done via a worker that gets scheduled via `DailyUpdate`.)

A follow-up could then extend it to handle articles that are discouraged but not prevented altogether. We could do this either by adding a way to acknowledge the warning and resubmit from the frontend, or by triggering an Alert when a discouraged articles is assigned and then sending an email to the user and/or the instructor with details about discouraged articles and why they should consider choosing a different article instead. Hmmm... after thinking through it, I guess the frontend approach will be better there, because one goal would be to prevent the noise of adding additional templates to the talk page of the article (even if they later get removed).

Abishekcs:
Sure, I will make the changes you have suggested. Thank you for the review .

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
```diff
@@ -306,16 +322,17 @@ const AssignButton = ({ course, role, course_id, wikidataLabels = {}, hideAssign
...
-    title.split('\n').filter(Boolean).forEach((assignment_title) => {
-      const assignment = {
-        title: decodeURIComponent(assignment_title).trim(),
-        project: project,
-        language: language,
-        course_slug: course.slug,
-        role: role
-      };
+    const assignArticle = () => {
+      title.split('\n').filter(Boolean).forEach(async (assignment_title) => {
```
TheTrio, line 328: why is this async?

app/assets/javascripts/utils/api.js
```diff
@@ -600,6 +598,33 @@ const API = {
         label,
       };
     });
+  },
+
+  async checkArticleInWikiCategory(articles) {
+    const wiki_articles_title = Array.isArray(articles) ? articles : [articles];
```
TheTrio, line 604: Ideally, the caller should always pass in an array. Would remove the need for this check

app/assets/javascripts/utils/api.js
```diff
@@ -600,6 +598,33 @@ const API = {
...
+            }
+            else {
+             return Promise.reject({ statusText: 'Request failed' });
+            }
+          });
+        } catch (error) {
+          logErrorMessage(response);
+        }
+      }));
+    return category_members.filter(result => result !== null);
```
TheTrio, line 627: might as well use `await` if we're doing async stuff. Would be a lot cleaner.

app/assets/javascripts/utils/api.js
```diff
@@ -600,6 +598,33 @@ const API = {
...
+
+  async checkArticleInWikiCategory(articles) {
+    const wiki_articles_title = Array.isArray(articles) ? articles : [articles];
+    const category_members = await Promise.all(wiki_articles_title.map(async (title) => {
+      try {
+        return await request(`discouraged/${title}`, {
+          method: 'GET',
+          headers: {
+            'Content-Type': 'application/json',
+          },
```
TheTrio, line 611: is this not the default?

app/models/wikipedia_category_member.rb
```diff
@@ -0,0 +1,57 @@
...
+    # Find the members to be removed (members in the database but not in the new data)
+    members_to_remove = existing_category_members - category_members
+
+    # Remove members that are not in the new data
+    WikipediaCategoryMember.where(category_member: members_to_remove).delete_all
+
+    # Add new members to the database
+    members_to_add.each do |category_member|
+      WikipediaCategoryMember.create(category_member:)
+    end
```
TheTrio, line 55: Might want to optimize this to use a single query if possible, although this might be redundant once you change the schema to the one suggested by @ragesoss

app/assets/javascripts/components/articles/add_available_articles.jsx
```diff
@@ -60,8 +63,29 @@ const AddAvailableArticles = createReactClass({
...
-      .then(() => this.resetInput());
+
+    // Check if any article/assignment title is under a particular wikipedia category
+    const categoryMembers = await API.checkArticleInWikiCategory(assignments.map(assignment => assignment.title));
+    const isArticleInCategory = categoryMembers.length > 0;
+
+    const addAssignment = () => {
+      return this.chainSubmissions(assignments)
+        .then(() => this.resetInput());
+    };
```
TheTrio, line 74: use async await

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
Abishekcs: Actually :sweat_smile: i was using async for this line  -
`const categoryMember = await API.checkArticleInWikiCategory(title.split('\n').map(item => item.trim()).filter(Boolean));`

TheTrio: Ah, missed that - the formatting makes it hard to read. There's still an extra async in `async (assignment_title)`

Also, could you fix the indentation for this?

Abishekcs: Okay, i will make the changes ty for the review

Abishekcs:
@ragesoss , I have moved the logic for discouraged assignments to `AssignmentManager` and am now utilizing the existing `category`. I would appreciate a review whenever you have the time.

app/assets/javascripts/utils/api.js
```diff
@@ -600,6 +598,28 @@ const API = {
...
+          }
+        } else {
+          throw new Error('Request failed');
+        }
+      } catch (error) {
+        logErrorMessage(error);
+      }
+    }
+
+    return category_members;
```
TheTrio, line 622: is there any reason this can't be parallelised? Currently, only request is made at a time even though all the requests are independent of each other.

Abishekcs: It can be parallelised 😅. Additionally, I am considering sending a single POST request to the server, passing the entire `wiki_articles_title` array in the request body?

lib/assignment_manager.rb
```diff
@@ -59,6 +59,16 @@ def claim_assignment(claimed_assignment)
     end
   end
 
+  def self.check_wiki_edu_discouraged_article(title:)
+    category = Category.find_by('article_titles LIKE ? AND name = ?', "%#{title}%",
```
ragesoss, line 63: It looks like this is looking for a Category named 'Wiki_Edu_Discouraged_Articles', and also matching based on article title?

lib/assignment_manager.rb
```diff
@@ -59,6 +59,16 @@ def claim_assignment(claimed_assignment)
...
+                                'Wiki_Edu_Discouraged_Articles')
+
+    if category&.source == ENV['blocked_assignment_category']
```
ragesoss, line 66: `Category#source` is for indicating which type of category record it is (ie, where it pulls the article title data from). In this case, the `Category` record should have `category` as the source (because it should use a Wikipedia category), and the `name` is the name of the Wikipedia category (which should come from configuration in application.yml).

lib/wiki_discouraged_article_manager.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+require_dependency "#{Rails.root}/lib/importers/category_importer"
+
+class WikiDiscouragedArticleManager
+  def retrieve_wiki_edu_discouraged_articles
+    category_titles = [ENV['blocked_assignment_category'],
+                       ENV['warning_assignment_category']]
```
ragesoss, line 7: Again, I think this will be better if we cut out the `warning` feature for this PR and just do the blocked assignments as an initial implementation. (That will make it much easier for me to review, easier to write tests for, and that's the feature I know that we will want to use immediately.)

ragesoss (changes requested):

lib/wiki_discouraged_article_manager.rb
Abishekcs: Okay, i will remove the `warning` feature.

lib/assignment_manager.rb
Abishekcs: Okay

app/assets/javascripts/utils/api.js
TheTrio: that sounds better

Abishekcs:
@ragesoss, I have removed the 'warning' feature for this PR and kept the feature for 'blocked assignment.' Currently, I am writing test code for this PR.

ragesoss:
@Abishekcs sounds good. Give me a ping when the tests are done. I will likely have some time to review it next week if it's ready then.

Abishekcs:
@ragesoss the test code is done for this PR. I would appreciate a review whenever you have the time.

ragesoss:
it's on my todo list.

app/controllers/assignments_controller.rb
```diff
@@ -25,10 +25,13 @@ def destroy
   def create
     check_permissions(assignment_params[:user_id].to_i)
     set_wiki { return }
-
+    check_wiki_edu_discouraged_article
```
ragesoss, line 28: I think this ought to be done a little differently; rather than separately and explicitly calling for this check within the controller, the check against the discouraged titles could be done as part of `AssignmentManager#create_assignment`. This would make it parallel to how we use `AssignmentManager::DuplicateAssignmentError`.

In addition to being simpler, this would also ensure that we do the same `set_clean_title` flow to get the title into the right form before checking it.

lib/data_cycle/daily_update.rb
```diff
@@ -33,6 +34,7 @@ def run_update
     update_users
     update_commons_uploads
     update_article_data
+    update_wiki_discouraged_article
```
ragesoss, line 37: I guess this should also use `if Features.wiki_ed?`, as we only anticipate using this feature on Wiki Education Dashboard.

lib/wiki_discouraged_article_manager.rb
```diff
@@ -0,0 +1,35 @@
...
+    category_title = ENV['blocked_assignment_category']
+    discouraged_articles = CategoryImporter.new(Wiki.default_wiki)
+                                           .page_titles_for_category("Category:#{category_title}")
+    update_or_save_discouraged_articles(discouraged_articles, category_title)
+  end
+
+  private
+
+  def update_or_save_discouraged_articles(discouraged_articles, category_title)
+    existing_article = Category.find_by(name: category_title)
```
ragesoss, line 15: This could find the wrong category, because `name` is not a unique identifier for a Category record.  The combination of name, wiki, source, and depth is unique.

I suggest using `Category.find_or_create_by`, which will let you simplify the rest of this method.

lib/wiki_discouraged_article_manager.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+require_dependency "#{Rails.root}/lib/importers/category_importer"
+
+class WikiDiscouragedArticleManager
+  def retrieve_wiki_edu_discouraged_articles
+    category_title = ENV['blocked_assignment_category']
+    discouraged_articles = CategoryImporter.new(Wiki.default_wiki)
```
ragesoss, line 7: Instead of calling `CategoryImporter` directly to get the titles, you can just find the `Category` record and then use `#refresh_titles` to do this.

ragesoss:
@Abishekcs it looks like the one failing spec is related to assignments, and it's not one I've seen fail before, so it might be related to your changes. Have you had a look at it and run that spec locally?

Abishekcs:
@ragesoss tried running the code locally, still failing, but can't figure out the issue. Any ideas or suggestions 😅?

Screenshot from 2023-12-05 12-06-30
Screenshot from 2023-12-05 12-06-22
Screenshot from 2023-12-05 12-06-10
Screenshot from 2023-12-05 12-06-04

ragesoss:
I fixed it by updated the selenium-webdriver gem, and I've pushed that update. If you rebase and `bundle install`, hopefully the tests will work properly for you now.

Abishekcs:
> I fixed it by updated the selenium-webdriver gem, and I've pushed that update. If you rebase and `bundle install`, hopefully the tests will work properly for you now.

okay i will give it a try

Abishekcs:
> I fixed it by updated the selenium-webdriver gem, and I've pushed that update. If you rebase and `bundle install`, hopefully the tests will work properly for you now.

@ragesoss the test is working properly now.
For this, Failed examples:
`rspec ./spec/lib/importers/revision_score_importer_spec.rb:205 # RevisionScoreImporter.update_revision_scores_for_all_wikis imports data and calcuates an article completeness score for available wikis` its woking fine locally

ragesoss:
Great. That failure is unrelated to your PR, so the tests seem to be in order. I will continue trying out and reviewing it soon. Thanks!

lib/assignment_manager.rb
```diff
@@ -104,6 +105,14 @@ def set_article_from_database
     @article = exact_title_matches.first
   end
 
+  def check_wiki_edu_discouraged_article
+    category = Category.where('name = ? AND article_titles LIKE ?',
```
ragesoss, line 109: I think the better way to do this is to find the category first, and then check whether the title is included in the catetgory's array of titles. Using LIKE might result in mismatches, such as blocking the assignment of "Animal" when "Animal rights" is on the blocklist.

ragesoss:
I tested this out, and it works the way I would expect. With the additional changes I requested, this will be ready to merge.

Abishekcs:
> I tested this out, and it works the way I would expect. With the additional changes I requested, this will be ready to merge.

 @ragesoss , I have made the requested changes for this.

Abishekcs:
@ragesoss any update on this PR😅 ?

ragesoss:
@Abishekcs yes, i'm happy with the PR and it is ready to merge, but I have not yet introduced this new feature to the Wikipedia community or finalized the category name to so, so I don't want to merge and deploy it yet. Doing that introduction is near the top of my todo list now that I am back from vacation; I hope to get it started today.

ragesoss:
The on-wiki discussion about this has not really come to a conclusion, in particular with regards to the category name or how the community wants to use it... but I'd like to go ahead and merge it and deploy it so that it's ready whenever we do come to consensus. So... what happens if this gets deployed but there is no `blocked_assignment_category` value set?

ragesoss:
Yes, that's what I mean. Is it safe to deploy it without such a value, so that it just would not do anything? If not, can it be easily adjusted to behave like that?

Abishekcs:
@ragesoss, `ENV['blocked_assignment_category']` is used in two places:

a) In `AssignmentManager`, it is used to find a category by name. It simply returns `null` if there is no value set for `ENV['blocked_assignment_category']` when I test it locally.

b) The second place is in `WikiDiscouragedArticleManager`, which is used in `dailyupdate`. I think for the second part, I will add an if statement to check if `ENV` is set. Since it might cause trouble with other daily updates.

Abishekcs:
test 
Ignore the `check ENV` it was just for testing locally
log

@ragesoss  i have added the if statement for ` WikiDiscouragedArticleManager`  and for `AssignmentManager` there is already an if statement to check category is present or not.

ragesoss:
Great, thanks!

lib/wiki_discouraged_article_manager.rb
```diff
@@ -9,12 +9,18 @@ def retrieve_wiki_edu_discouraged_articles
...
 
   def update_or_create_discouraged_articles
-    Category.find_or_create_by(
-      name: ENV['blocked_assignment_category'],
-      wiki_id: en_wiki.id,
-      source: 'category',
-      depth: 0
-    ).refresh_titles
+    blocked_assignment_category = ENV['blocked_assignment_category']
+
```
ragesoss, line 13: I'd prefer a `return unless` guard clause here, rather than the branching `if` statement.

ragesoss:
Looks like your new change broke the tests. The fix probably involves adding a new entry in `config/environments/test.rb`.

Abishekcs:
> Looks like your new change broke the tests. The fix probably involves adding a new entry in `config/environments/test.rb`.

It should pass now  😅 added the entry for test ENV.

ragesoss:
Thanks @Abishekcs! This is an exciting feature, and I'm looking forward to seeing it in action.

Abishekcs:
> Thanks @Abishekcs! This is an exciting feature, and I'm looking forward to seeing it in action.

Great! 👍

PR #5498, opened 2023-10-04
refactor: article graph to use functional component

justiceotuya:

## What this PR does
refactor: article_graph class component to use functional component from this Issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5393

## Screenshots

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/17678464/f0a2de60-c8d3-49f9-b9ed-c4d6f83f4df1


Aminehassou:
Could you add a reference to the relevant issue in your PR? (it makes it easier to track which components have been refactored if each PR references back to issue #5393)

justiceotuya:
> Could you add a reference to the relevant issue in your PR? (it makes it easier to track which components have been refactored if each PR references back to issue #5393)

done

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,148 @@
...
-
-  componentDidUpdate(_, prevState) {
-    if (this.state.showGraph && !prevState.showGraph) {
-      // Add event listener when the component is visible
-      document.addEventListener('mousedown', this.handleClickOutside);
+  useEffect(() => {
+    if (showGraph) {
+      document.addEventListener('mousedown', handleClickOutside);
+    } else {
+      document.removeEventListener('mousedown', handleClickOutside);
```
TheTrio, line 20: this won't actually work because `handleClickOutside` is defined within this function itself - which means that it will be recreated everytime the component rerenders. The reason it worked earlier(with the class component) was because only the render method is called during rerenders and so `this.handleClickOutside` never changed

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,148 @@
...
+
+  function handleRadioChange(event) {
+    setSelectedRadio(event.currentTarget.value);
+  }
+
+  function graphId() {
+    return `vega-graph-${article_id}`;
+  }
+
+  const dataIncludesWp10 = articleData && articleData[0].wp10;
```
TheTrio, line 65: use optional chaining here

app/assets/javascripts/components/articles/article_graphs.jsx
justiceotuya: so i dont understand when you say it wont work. are you saying that because of performance reason we should only create the `handleClickOutside` once through out the lifecycle of the component?

because yes it does work and I can click outside to close the graph

@TheTrio what do you think the solution should be?

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,148 @@
...
+      <EditSizeGraph
+        graphid={graphId()}
+        graphWidth={500}
+        graphHeight={300}
+        articleData={articleData}
+      />
+    );
+  }, [articleData, dataIncludesWp10, selectedRadio]);
+
+  const radioInput = useMemo(() => {
```
TheTrio, line 93: there's no need for the `useMemo` - there's no network call involved here

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,148 @@
...
-});
+  }, [articleData, dataIncludesWp10, selectedRadio]);
+
+  const editSize = useMemo(() => {
+    if (dataIncludesWp10 || !articleData) {
+      return null;
+    }
+
+    return <p>{I18n.t('articles.edit_size')}</p>;
+  }, [articleData, dataIncludesWp10]);
```
TheTrio, line 130: same here - why the `useMemo?`

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,148 @@
...
+  }
+
+  function graphId() {
+    return `vega-graph-${article_id}`;
+  }
+
+  const dataIncludesWp10 = articleData && articleData[0].wp10;
+
+  const graph = useMemo(() => {
+    if (!articleData) {
```
TheTrio, line 68: no need for the `useMemo`

app/assets/javascripts/components/articles/article_graphs.jsx
TheTrio: it will work - the problem is that it won't remove the old event listener. You can verify this by putting a console.log in `handleClickOutside` and then going through this cycle of hiding and showing the graph. You will get an extra console log every time since the old listeners aren't cleaned up

app/assets/javascripts/components/articles/article_graphs.jsx
justiceotuya: We need the use memo here as building the graph I based on the result gotten from calling getData, so the graph changes based on the result from calling getData. adding articleData as array dependency in the memo call allows the graph to be rendered with the new article data @TheTrio

app/assets/javascripts/components/articles/article_graphs.jsx
justiceotuya: removed

app/assets/javascripts/components/articles/article_graphs.jsx
justiceotuya: removed

app/assets/javascripts/components/articles/article_graphs.jsx
justiceotuya: done

app/assets/javascripts/components/articles/article_graphs.jsx
TheTrio: removing the useMemo still does that - it will get called every time state changes. Not sure I follow your logic here

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,153 @@
...
+
+  const dataIncludesWp10 = articleData?.[0]?.wp10;
 
-  handleClickOutside(event) {
-    const element = this.ref.current;
-    if (element && !element.contains(event.target)) {
-      this.hideGraph();
+  const graph = () => {
+    if (!articleData) {
+      return <Loading />;
```
TheTrio, line 72: The pattern we use in the rest of the codebase is doing something like

```js
let graph;
if(...){
  graph = ...
}else{
  graph = ...
}
```

I think that's preferable to having an unnecessary function call

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,153 @@
...
+        graphWidth={500}
+        graphHeight={300}
+        articleData={articleData}
+      />
+    );
+  };
+
+  const radioInput = () => {
+    if (!dataIncludesWp10 || !articleData) {
+      return null;
```
TheTrio, line 98: similar to https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5498/files#r1365943358

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,153 @@
...
+    setShowGraph(false);
+  }
+
+  function handleRadioChange(event) {
+    setSelectedRadio(event.currentTarget.value);
+  }
+
+  function graphId() {
+    return `vega-graph-${article_id}`;
+  }
```
TheTrio, line 66: why not make this a constant instead of a function?

TheTrio:
@justiceotuya could you resolve conversations as you go along addressing them? Makes it easier to track changes

justiceotuya:
> @justiceotuya could you resolve conversations as you go along addressing them? Makes it easier to track changes

Resolved and fixed @TheTrio

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,148 @@
...
         </div>
-      </a>
+      </div>
     );
   }
-});
+
+  let editSize;
+  if (dataIncludesWp10 || !articleData) {
+    editSize = null;
```
TheTrio, line 126: this is not required

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,148 @@
...
+        graphHeight={300}
+        articleData={articleData}
+      />
+    );
+  }
+
+  let radioInput;
+
+  if (!dataIncludesWp10 || !articleData) {
+    radioInput = null;
```
TheTrio, line 94: this isn't required either

TheTrio: why is this marked as resolved?

app/assets/javascripts/components/articles/article_graphs.jsx
TheTrio: same here — this hasn’t been fixed and yet the conversation is marked as resolved.

justiceotuya: unresolved and fixed @TheTrio

app/assets/javascripts/components/articles/article_graphs.jsx
justiceotuya: unresolved and fixed @TheTrio

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -1,170 +1,137 @@
...
+        articleData={articleData}
+      />
     );
   }
-});
+
+  let radioInput = (
+      <div>
+        <div className="input-row">
+          <input
```
TheTrio, line 94: missing a check for `articleData[0].wp10`

TheTrio (approved):
LGTM

Issue #5492, opened 2023-09-29
Use LiftWing instead of ORES in ArticleFinder


ragesoss:
We need to use the new LiftWing API instead of the now-disabled ORES API, when fetching article quality data from the React frontend.

See `article_finder_actions.js`.

The migration from ORES to LiftWing was recently completed for server-side uses; see `lift_wing_api.rb` for some details on usage, as well as https://wikitech.wikimedia.org/wiki/Machine_Learning/LiftWing/Usage

ragesoss:
> hi @ragesoss can you please help on how i can start contributing in these issues, how to setup project locally

Hi @codebreaker0001 . Instructions for setting up the project locally are in https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/setup.md

sharon-kathambi:
Hi @ragesoss, I would like to attempt this issue

justiceotuya:
@ragesoss how do we get to know which model is available? i have tried converting to Liftwing and I am getting this error

https://api.wikimedia.org/service/lw/inference/v1/models/enwiki-articlequality:predict?models=wp10&revids=1175222458%7C1178717228%7C1173765883%7C908481996%7C1177872320%7C1114447757%7C1022916502%7C1167068313%7C1178263693%7C1162049025&callback=jsonp_1696603447179_28957

```
{
  "error": "Model with name enwiki-articlequality:predict does not exist."
}
```

justiceotuya:
i have constructed the poat request but it I am getting an API cors error @ragesoss

justiceotuya:
@ragesoss https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5501

PR #5501, opened 2023-10-07
Liftwing conversion

justiceotuya:

## What this PR does
This MR fixes this issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5492 where Liftwing Api is converted from the previous Ores implementation

## Screenshots
Before:

Screenshot 2023-10-07 at 4 53 56 PM

After:

Screenshot 2023-10-07 at 4 50 37 PM

- learnt how to read and understand a ruby class
- learnt how to read data in a ruby class from the front end and how to also send data back to the frontend

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -13,7 +13,7 @@ Rails.application.routes.draw do
   # intercepting the click and issuing a post request. Omniauth login is post-only.
   get 'users/auth/mediawiki', to: redirect('/')
   devise_for :users, controllers: { omniauth_callbacks: 'omniauth_callbacks' }
-  
+
   devise_scope :user do
     # OmniAuth may fall back to :new_user_session when the OAuth flow fails.
     # So, we treat it as a login error.
@@ -203,6 +203,7 @@ Rails.application.routes.draw do
        constraints: { course_id: /.*/ }
 
   get 'revisions' => 'revisions#index'
+  get 'revisions/show' => 'revisions#show'
 
   get 'articles/article_data' => 'articles#article_data'
   get 'articles/details' => 'articles#details'
@@ -302,6 +303,7 @@ Rails.application.routes.draw do
 
   # Revision Feedback
   get '/revision_feedback' => 'revision_feedback#index'
+  # post '/revision_feedback' => 'revision_feedback#index'
 
   # Wizard
   get 'wizards' => 'wizard#wizard_index'
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -302,7 +302,6 @@ Rails.application.routes.draw do
 
   # Revision Feedback
   get '/revision_feedback' => 'revision_feedback#index'
-  # post '/revision_feedback' => 'revision_feedback#index'
 
   # Wizard
   get 'wizards' => 'wizard#wizard_index'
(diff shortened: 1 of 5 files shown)


ragesoss:
I'd strongly prefer that the article quality data be fetched directly from the client, rather than hitting a Rails controller and having the server make the LiftWing requests.

If you ran into trouble with making the requests via JavaScript, post the details and maybe we can figure out how to fix it.

justiceotuya:
ok. on it

justiceotuya:
@ragesoss i am having issues sending data to the api.wikimedia endpoint to get the lifting data
This is my curl request. it works on the terminal but gets a COR error while using the browser and that was why I resorted in going via the server route.

one solution to this is to 
- use a proxy solution like ngrok (this means when testing ngrok must be used)
- whitelist localhost to be able to hit and get data from the api.wikimedia service

```bash
curl --location 'https://api.wikimedia.org/service/lw/inference/v1/models/enwiki-articlequality:predict' \
--header 'Content-Type: application/json' \
--data '{
    "rev_id": 1162049025
}'
```


attached is a screenshot of the error

SCR-20231009-roqy

ragesoss:
Can you post the code for the request that resulted in that error?

justiceotuya:
> Can you post the code for the request that resulted in that error?

Here  @ragesoss  https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5504

ragesoss:
Try adding `&origin=*` to the end of the URL: https://www.mediawiki.org/wiki/API:Cross-site_requests#CORS_usage

justiceotuya:
> Try adding `&origin=*` to the end of the URL: https://www.mediawiki.org/wiki/API:Cross-site_requests#CORS_usage

@ragesoss  same issue. i used

`https://api.wikimedia.org/service/lw/inference/v1/models/enwiki-articlequality:predict&origin=*`

also

`https://api.wikimedia.org/service/lw/inference/v1/models/enwiki-articlequality:predict&origin=*`

The error still persists

image

image

ragesoss:
The Wikimedia ML team is looking into making this work: https://phabricator.wikimedia.org/T348511

elukey:
@justiceotuya @ragesoss Hi! A fix has been deployed to support the Content-Type header in CORS policies, could you please re-test?

justiceotuya:
@ragesoss fixed

config/routes.rb
```diff
@@ -302,6 +302,7 @@
 
   # Revision Feedback
   get '/revision_feedback' => 'revision_feedback#index'
+  # post '/revision_feedback' => 'revision_feedback#index'
```
ragesoss, line 305: This is not needed.

ragesoss:
It looks like the spec that tests Article Finder is still failing.

ragesoss:
It looks like this is basically working, although in testing it locally I'm getting some intermittent errors that are not handled well. (On the LiftWing side, I saw one instance of a 504 error, which caused a red error notification but showed no useful user-facing information. There are also some errors that sometimes happen with the pageviews API, which are probably related to #5508. I'll open a new issue for these once this is merged.)

It looks like this PR combines a lot of code reformatting with the actual changes to move to LiftWing, which makes it a lot harder for me to review. Could you isolate the necessary changes without any formatting changes alongside it, so that I can think through the specific LiftWing-related code before merging? Thanks!

PR #5502, opened 2023-10-08
Add a way to reach specific training modules and slides via ID numbers

### LINKED ISSUE #4940: Add a way to reach specific training modules and slides via ID numbers

gabina:
## What this PR does
The purpose of this PR is to add:
- A way to reach a training module by their id.
- A way to reach an individual slide by their id.

In order to do that, two new routes were added (`find_training_module` and  `find_training_slide`), as long as their corresponding controllers and specs.

The logic behind the controllers is to redirect the user to the URL for the given training module (`/training/library-slug/module-slug`) or to the individual slide (`/training/library-slug/module-slug/slide-slug`).

Closes #4940

## Screenshots
After this PR, users are redirected to the training module or slide site.

[Screencast from 08-10-23 23:50:00.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/8717242/5889fb40-dc36-4016-8175-f747d5b4e8b2)


spec/factories/training_slides.rb
```diff
@@ -19,5 +19,8 @@
 
 FactoryBot.define do
   factory :training_slide do
+    id { 456875}
+    title { 'How to create a slide' }
+    slug { 'how-to-create-a-slide' }
   end
 end
```
gabina, line 26: Added this because I want to test the behavior for an orphan slide (one that doesn't belong to any module).

gabina:
Thanks for approving the workflow. It looks like the new `TrainingController` specs I added passed. However, CI failed. By taking a look at the failed examples I think it's not related to my changes.  
```
Failed examples:

rspec ./spec/features/article_finder_spec.rb:17 # article finder performs searches and returns results
rspec ./spec/lib/lift_wing_api_spec.rb:10 # LiftWingApi returns the basically the same data as OresApi
rspec ./spec/lib/lift_wing_api_spec.rb:20 # LiftWingApi handles TextDeleted revs similarly
rspec ./spec/lib/ores_api_spec.rb:29 # OresApi#get_revision_data fetches json from ores.wikimedia.org for wikipedia
rspec ./spec/lib/ores_api_spec.rb:37 # OresApi#get_revision_data fetches json from ores.wikimedia.org for wikidata
rspec ./spec/lib/ores_api_spec.rb:46 # OresApi#get_revision_data handles many revisions per request for wikipedia
rspec ./spec/lib/ores_api_spec.rb:53 # OresApi#get_revision_data handles many revisions per request for wikidata
rspec ./spec/lib/replica_spec.rb:88 # Replica API requests returns a list of existing articles
```

app/controllers/training_controller.rb
```diff
@@ -59,6 +60,15 @@ def reload
     render plain: e.message
   end
 
+  def find_slide
+    training_slide = TrainingSlide.find(params[:slide_id])
+    training_module = find_module_from_slide_slug(training_slide.slug)
+    raise ActionController::RoutingError, 'module not found' unless training_module
+    training_library = find_library_from_module_slug(training_module.slug)
+    raise ActionController::RoutingError, 'library not found' unless training_library
```
ragesoss, line 68: It's possible to have a module that is not in any library, and it would render if you visited a path that had any valid training library slug. (You can view a module from one library even if you substitute the slug from another library.)

It would be good to find the library as you do here, but if no library is found, it would be better to use TrainingLibrary.first as a fallback instead of throwing a 404.

app/helpers/training_helper.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+#= Helpers for training views
+module TrainingHelper
+  # Given a module slug, it returns the first library that has a
+  # category including that module slug. It returns nil if no such
+  # library is found.
+  def find_library_from_module_slug(module_slug)
```
ragesoss, line 8: I think both of these helper methods would make more sense as instance methods of `TrainingSlide` and `TrainingModule` respectively.

app/helpers/training_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+      library.categories.each do |category|
+        next unless category.key?('modules')
+        category['modules'].each do |mod|
+          return library if mod['slug'] == module_slug
+        end
+      end
+    end
+  end
+
+  # Given a slide slug, it returns the first module including it.
```
ragesoss, line 19: As a TrainingSlide instance method, you could implement this as 

```
TrainingModule.all.detect { |tm| tm.slide_slugs.include? slug }
```

ragesoss:
This looks great overall.

app/helpers/training_helper.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+#= Helpers for training views
+module TrainingHelper
+  # Given a module slug, it returns the first library that has a
+  # category including that module slug. It returns nil if no such
+  # library is found.
+  def find_library_from_module_slug(module_slug)
+    TrainingLibrary.all.find_each do |library|
```
ragesoss, line 9: For this method, I would suggest implementing TrainingLibrary#training_module_slugs (for example:
```
  def training_module_slugs
    categories.map do |cat|
      cat['modules'].map { |mod| mod['slug'] }
    end.flatten
  end
```
)

and then using that to implement TrainingModule#library:

```
TrainingLibrary.all.detect { |tl| tl.training_module_slugs.include? slug }
```

app/controllers/training_controller.rb
gabina: hmm interesting. That behavior is not so intuitive I guess.

I changed the logic for the controllers and updated specs on fbb894f8765a8f9a1537ed3ea056bdcdcc714eea

Note that the new definitions may result in failure if no library exists. In such a scenario, calling `TrainingLibrary.first` would return `nil`, causing an error when attempting to access the slug.
Not sure if a world with no libraries makes sense, or it is not worth considering this case and we can suppose `TrainingLibrary.first` is always well defined. 
It looks like modules and slides render even using a non-existing library slug, so we could use a literal string
`'mylibrary'` as default library slug value if no libraries at all.
What's your opinion here?

app/helpers/training_helper.rb
gabina: Much more idiomatic, I like it!

app/models/training_module.rb
```diff
@@ -153,5 +153,17 @@ def sandbox_location
...
+  # It returns nil if no such library is found.
+  def find_library_by_slug
+    TrainingLibrary.all.detect { |tl| tl.training_module_slugs.include? slug }
+  end
+
+  # Returns a specific training library for the module,
+  # or a default library if it is not found.
+  def find_or_default_library
+    find_library_by_slug || TrainingLibrary.first
+  end
```
gabina, line 166: I created this method only to avoid code duplication in controllers.

Alternatively, I could define just something like this: 
```
  # Returns the first library that has a category including the module slug,
  # or a default library if it is not found.
  def find_library_by_slug_or_default
    TrainingLibrary.all.detect(-> { TrainingLibrary.first }) { |tl| tl.training_module_slugs.include? slug }
  end
```

ragesoss:
Nice work!

PR #5503, opened 2023-10-08
Add fon language code for Fon Wikipedia

Mahuton:
This has been launched recently.
See: https://incubator.wikimedia.org/wiki/Incubator:Site_creation_log


Mahuton:
@ragesoss hi,
do you think we can have this approved soon?
Thanks

ragesoss:
Thanks! I'll deploy it within the next few days.

PR #5507, opened 2023-10-13
Remove course template from user page and user talk page when user disenrolls

### LINKED ISSUE #2842: Removing a course participant doesn't cause template to be removed from wiki

gabina:
## What this PR does
The purpose of this PR is to add a way to remove course templates from user page and user talk page when user gets dis-enrolled from a specific course.

When a user enrolls in a course, a template pointing to the course is added to the user page and the user talk page (example user page with participating template: https://en.wikipedia.org/wiki/User:Jfitz19)

This PR adds a new `DisenrollFromCourseWorker` worker to remove course templates from user page and user talk page. It basically "reverts" what the `EnrollInCourseWorker` does (except for the user's sandbox action).

Now, when a user gets removed from a course, the `EnrollmentController#remove` schedules a job for the  `DisenrollFromCourseWorker` to remove the participating templates from user page and user talk page.

Closes #2842

### DIFF
--- a/app/controllers/users/enrollment_controller.rb
+++ b/app/controllers/users/enrollment_controller.rb
@@ -119,11 +119,17 @@ class Users::EnrollmentController < ApplicationController
       return
     end
 
-    remove_assignment_templates
+    remove_assignment_and_enrollment_templates
+
     @course_user.destroy # destroying the course_user also destroys associated Assignments.
 
     render 'users', formats: :json
     update_course_page_and_assignment_talk_templates
+  end
+
+  def remove_assignment_and_enrollment_templates
+    remove_assignment_templates
+    make_disenrollment_edits
   end
 
   # If the user has Assignments, update article talk pages to remove them from
@@ -134,6 +140,15 @@ class Users::EnrollmentController < ApplicationController
       WikiCourseEdits.new(action: :remove_assignment, course: @course,
                           current_user:, assignment:)
     end
+  end
+
+  # Remove enrollment templates from user page and user talk page.
+  def make_disenrollment_edits
+    return unless student_role?
+    # for students only, remove templates from userpage and user talk page
+    DisenrollFromCourseWorker.schedule_edits(course: @course,
+                                             editing_user: current_user,
+                                             disenrolling_user: @user)
   end
 
   ##################
--- a/app/controllers/users/enrollment_controller.rb
+++ b/app/controllers/users/enrollment_controller.rb
@@ -104,22 +104,12 @@ class Users::EnrollmentController < ApplicationController
   def remove
     set_course_and_user
     return if @user.nil?
-    # For events controlled by Event Center, only non-student roles
-    # can be changed on the Dashboard. Student role is handled
-    # via WikimediaEventCenterController.
-    if @course.controlled_by_event_center? && student_role?
-      render json: { message: I18n.t('courses.controlled_by_event_center') }, status: :unauthorized
-      return
-    end
 
-    @course_user = CoursesUsers.find_by(user: @user, course: @course,
-                                        role: enroll_params[:role])
-    if @course_user.nil? # This will happen if the user was already removed.
-      render 'users', formats: :json
-      return
-    end
+    ensure_role_is_authorized { return }
+    ensure_course_user_exists { return }
 
-    remove_assignment_and_enrollment_templates
+    remove_assignment_templates
+    make_disenrollment_edits
 
     @course_user.destroy # destroying the course_user also destroys associated Assignments.
 
@@ -127,9 +117,21 @@ class Users::EnrollmentController < ApplicationController
     update_course_page_and_assignment_talk_templates
   end
 
-  def remove_assignment_and_enrollment_templates
-    remove_assignment_templates
-    make_disenrollment_edits
+  # For events controlled by Event Center, only non-student roles
+  # can be changed on the Dashboard. Student role is handled
+  # via WikimediaEventCenterController.
+  def ensure_role_is_authorized
+    return unless @course.controlled_by_event_center? && student_role?
+    render json: { message: I18n.t('courses.controlled_by_event_center') }, status: :unauthorized
+    yield
+  end
+
+  def ensure_course_user_exists
+    @course_user = CoursesUsers.find_by(user: @user, course: @course,
+                                        role: enroll_params[:role])
+    return unless  @course_user.nil? # This will happen if the user was already removed.
+    render 'users', formats: :json
+    yield
   end
 
   # If the user has Assignments, update article talk pages to remove them from
--- a/lib/wiki_course_edits.rb
+++ b/lib/wiki_course_edits.rb
@@ -188,7 +188,7 @@ class WikiCourseEdits
     # If content to remove does not exist on initial page, then there is nothing
     # to remove
     return unless initial_page_content&.include?(content)
-    new_page_content = initial_page_content.gsub(content, '')
+    new_page_content = initial_page_content.gsub(/#{Regexp.quote(content)}(\n)?/, '')
     @wiki_editor.post_whole_page(@current_user, page_title, new_page_content, summary)
   end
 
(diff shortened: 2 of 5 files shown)


app/controllers/users/enrollment_controller.rb
```diff
@@ -119,13 +119,19 @@ def remove
...
     @course_user.destroy # destroying the course_user also destroys associated Assignments.
 
     render 'users', formats: :json
     update_course_page_and_assignment_talk_templates
   end
 
+  def remove_assignment_and_enrollment_templates
+    remove_assignment_templates
+    make_disenrollment_edits
+  end
```
gabina, line 133: Created this because got an "Method has too many lines" offense from rubocop

lib/wiki_course_edits.rb
```diff
@@ -70,6 +70,25 @@ def enroll_in_course(enrolling_user:)
...
+    user_page = "User:#{@disenrolling_user.username}"
+    summary = generator.disenrollment_summary
+    remove_content_from_page(user_page, template, summary)
+
+    # Remove existing template from the user's talk page
+    talk_template = generator.enrollment_talk_template
+    talk_page = "User_talk:#{disenrolling_user.username}"
+    talk_summary = "removing {{#{template_name(@templates, 'user_talk')}}}"
+    remove_content_from_page(talk_page, talk_template, talk_summary)
+  end
```
gabina, line 90: Sanity check: should I do something with the template in sandbox to revert the `add_template_to_sandbox` action?

app/controllers/users/enrollment_controller.rb
ragesoss: I think a better way to reorganize this code would be to extract some of the setup and guard code at the start of the method. Ideally, the code in a method is at roughly the same level of abstraction, and the most detailed bits are the first ones to extract.

Extracting these two lines does not really make the method any easier to understand, which is the purpose behind that rubocop rule.

lib/wiki_course_edits.rb
ragesoss: I don't think that's necessary, as that's not a place other editors will look to understand the context for the un-enrolled student's edits.

app/controllers/users/enrollment_controller.rb
gabina: Makes sense, I see your point. I tried to do that on 99392335e6128fb183c5a182d25d3c536deb9b7d

lib/wiki_course_edits.rb
gabina: I think I edited the message adding a second question after you answered, so just checking you saw both questions.

gabina:
Note: I was able to enable wiki edits on my local environment, and I reproduced this by adding and then removing my own user to a course. The history of how my user page and user talk page was modified can be check on https://en.wikipedia.org/w/index.php?title=User:Gabinaluz&action=history.

I didn't feel very confident reproducing that for fear of unintentional changes, so I didn't try more than that.

gabina:
CI failed with the following errors:
```
Finished in 31 minutes 9 seconds (files took 12.19 seconds to load)
1825 examples, 4 failures, 17 pending

Failed examples:

rspec ./spec/features/article_finder_spec.rb:17 # article finder performs searches and returns results
rspec ./spec/lib/wiki_pageviews_spec.rb:72 # WikiPageviews.views_for_article for an unviewed article returns an empty hash
rspec ./spec/lib/wiki_pageviews_spec.rb:151 # WikiPageviews.average_views_for_article for an article that does not exist returns 0
rspec ./spec/lib/wiki_pageviews_spec.rb:164 # WikiPageviews.average_views_for_article for an article that exist but has no view data returns 0
```
Will check if something is related to my changes.

ragesoss:
> Note: I was able to enable wiki edits on my local environment, and I reproduced this by adding and then removing my own user to a course. The history of how my user page and user talk page was modified can be check on https://en.wikipedia.org/w/index.php?title=User:Gabinaluz&action=history.
> 
> I didn't feel very confident reproducing that for fear of unintentional changes, so I didn't try more than that.

It looks like the edit summaries were buggy: `User has disenrolled in [[]]`. Do you know why that is?

lib/wiki_course_edits.rb
ragesoss: Ah, I missed the second question. I think yes, checking `enrollment_edits_enabled?` for this action makes sense.

gabina: Added the check and spec for that on 00cf010f45cfc3db2c60db59fc01c3e63189de8f

gabina:
> It looks like the edit summaries were buggy: `User has disenrolled in [[]]`. Do you know why that is?

Oh good point.
The `disenrollment_summary` uses the `course.wiki_title` to build the summary, which needs the `wiki_course_page_enabled?` to return `true` to be other than `nil`. Due to the fact that on my local environment the `submitted` field for the course example was set to 0, `course.wiki_title` ended up being `nil`. I was playing around modifying the settings on my local environment and could get something like:
- User has enrolled in [Project:Outreach_Dashboard/Stanford_Law_School/Advanced_Legal_Research_Winter_2020_(Winter)](https://en.wikipedia.org/w/index.php?title=Wikipedia:Outreach_Dashboard/Stanford_Law_School/Advanced_Legal_Research_Winter_2020_(Winter)&action=edit&redlink=1).
- User has disenrolled in [Project:Outreach_Dashboard/Stanford_Law_School/Advanced_Legal_Research_Winter_2020_(Winter)](https://en.wikipedia.org/w/index.php?title=Wikipedia:Outreach_Dashboard/Stanford_Law_School/Advanced_Legal_Research_Winter_2020_(Winter)&action=edit&redlink=1).

You can check that on https://en.wikipedia.org/w/index.php?title=User:Gabinaluz&action=history
BTW, the site for the course doesn't exist (it was automatically created by my action, but I removed it). 

On the other hand, while taking a look at your point, I noticed a small bug in the disenrollment action, that wasn't caught by the tests. If you [compare the last two versions](https://en.wikipedia.org/w/index.php?title=User%3AGabinaluz&diff=1180488710&oldid=1180488539), you can notice that an extra new line is added when removing the course template, which is wrong. I modified the tests to catch the error, and then fixed it using a regular expression on 48d7a55c38dc2f4b8ac1355cd3c8897254a2110d

lib/wiki_course_edits.rb
```diff
@@ -163,6 +182,16 @@ def add_template_to_sandbox
                            sandbox_template:, current_user: @current_user)
   end
 
+  def remove_content_from_page(page_title, content, summary)
+    initial_page_content = @wiki_api.get_page_content(page_title)
+
+    # If content to remove does not exist on initial page, then there is nothing
+    # to remove
+    return unless initial_page_content&.include?(content)
+    new_page_content = initial_page_content.gsub(/#{Regexp.quote(content)}(\n)?/, '')
```
gabina, line 191: I added an optional new line in the regexp, but maybe we can use just something like `#{Regexp.quote(content)}\n`, as the new line should be always added when writing the course template

ragesoss: I guess the newline should be removed if present, but if there is not a newline, it should still remove the template.

gabina: Makes sense.  Then it looks like using `/#{Regexp.quote(content)}(\n)?/` (as it's now) is the right way. If the new line after the content exists, it will match the content and the new line. If the new line doesn't exist, it will match only the content.

Let me know if there is something for me to review. Otherwise, I understand I already addressed all the suggestions/questions.

gabina:
I sent a request to join to wiki-education organization for Sentry because I cannot see the issue in Sentry

ragesoss:
@gabina that error is definitely not related to your change; I haven't deployed it to Programs & Events Dashboard yet.

PR #5513, opened 2023-10-18
Fix failing specs due to new wikipageviews API behavior when no data available

### LINKED ISSUE #5508: Update WikiPageviews and spec for new API behavior
We have a handful of specs in `wiki_pageviews_spec.rb` that recently started failing. It looks like this is because the behavior of the API has changed when handling requests for which no data is available, so our code needs to be updated to handle the current behavior.

The easiest way to gets started will be to run that spec, then figure out what changes need to be made to make the specs pass.

gabina:
## What this PR does
The behavior of the wikipageviews API has changed when handling requests for which no data is available, causing some of our specs to fail. This PR starts managing this new behavior to fix our specs.
It basically adds a new `no_data_available_response?` method to identify these specific responses.
Read more details in the proper issue.

Closes #5508

### DIFF
--- a/lib/wiki_pageviews.rb
+++ b/lib/wiki_pageviews.rb
@@ -99,23 +99,15 @@ class WikiPageviews
     return unless response
     data = Utils.parse_json(response)
     return data['items'] if data['items']
-    # As of October 2017, the data type is https://www.mediawiki.org/wiki/HyperSwitch/errors/not_found
-    return no_results if %r{errors/not_found}.match?(data['type'])
-    return no_results if no_data_available_response?(data)
+
+    # Since the API experienced some changes in the response when handling requests for which no
+    # data is available, we decided to rely only on the data status being 404 to return no results
+    return no_results if data['status'] == 404
     raise PageviewApiError, response
   end
 
   def no_results
     {}
-  end
-
-  # As of October 2023, we started to see 404 not found responses with about:blank type
-  # and a specific detail when handling requests for which no data is available
-  def no_data_available_response?(response)
-    no_data_avialable_detail = 'The date(s) you used are valid, but we either do not have data '\
-                               'for those date(s), or the project you asked for is not loaded yet.'\
-                               ' Please check documentation for more information.'
-    response['status'] == 404 && response['detail'] == no_data_avialable_detail
   end
 
   def wiki_url_param
--- a/lib/wiki_pageviews.rb
+++ b/lib/wiki_pageviews.rb
@@ -101,11 +101,21 @@ class WikiPageviews
     return data['items'] if data['items']
     # As of October 2017, the data type is https://www.mediawiki.org/wiki/HyperSwitch/errors/not_found
     return no_results if %r{errors/not_found}.match?(data['type'])
+    return no_results if no_data_available_response?(data)
     raise PageviewApiError, response
   end
 
   def no_results
     {}
+  end
+
+  # As of October 2023, we started to see 404 not found responses with about:blank type
+  # and a specific detail when handling requests for which no data is available
+  def no_data_available_response?(response)
+    no_data_avialable_detail = 'The date(s) you used are valid, but we either do not have data '\
+                               'for those date(s), or the project you asked for is not loaded yet.'\
+                               ' Please check documentation for more information.'
+    response['status'] == 404 && response['detail'] == no_data_avialable_detail
   end
 
   def wiki_url_param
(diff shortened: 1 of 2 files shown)


gabina:
CI failed, will check 
```
1826 examples, 2 failures, 17 pending

Failed examples:

rspec ./spec/features/namespace_stats_spec.rb:27 # Namespace-specific stats generates and renders stats for Cookbook namespace on en.wikibooks.org
rspec ./spec/lib/importers/revision_score_importer_spec.rb:205 # RevisionScoreImporter.update_revision_scores_for_all_wikis imports data and calcuates an article completeness score for available wikis
```

spec/lib/wiki_pageviews_spec.rb
```diff
@@ -161,6 +161,8 @@
       let(:title) { 'Voyages,_aventures_et_combats/Chapitre_18' }
       let(:subject) { described_class.new(article).average_views }
 
+      before { travel_to Date.new(2023, 10, 18) }
+
```
gabina, line 165: I had to add this because it looks like I visited the site yesterday so I broke the spec. 
I noticed that because the build for PR #5498 failed with:
```
  3) WikiPageviews.average_views_for_article for an article that exist but has no view data returns 0
     Failure/Error: expect(subject).to eq(0)

       expected: 0
            got: 1.0
```

This way it should be more robust.

gabina:
@ragesoss this PR is ready for review ( I don't think I can set you as reviewer)

lib/wiki_pageviews.rb
```diff
@@ -101,13 +101,23 @@ def parse_results(response)
...
   end
 
   def no_results
     {}
   end
 
+  # As of October 2023, we started to see 404 not found responses with about:blank type
+  # and a specific detail when handling requests for which no data is available
+  def no_data_available_response?(response)
+    no_data_avialable_detail = 'The date(s) you used are valid, but we either do not have data '\
```
ragesoss, line 115: I think relying on the exact 'detail' message here is going to be too fragile. Is this always the message when type is `about:blank`? I think the `type` is less likely to change than the text of the 'detail' message, which could get updated wording at any point.

lib/wiki_pageviews.rb
```diff
@@ -101,13 +101,23 @@ def parse_results(response)
     return data['items'] if data['items']
     # As of October 2017, the data type is https://www.mediawiki.org/wiki/HyperSwitch/errors/not_found
     return no_results if %r{errors/not_found}.match?(data['type'])
```
ragesoss, line 103: Are there still also requests that return this error type, or did all 404's become `about:blank`?

ragesoss: Maybe it would be better to just rely on the 404 status and ignore `type` altogether? I imagine that any case of a 404 should be treated the same way by the Dashboard, and it looks like that's the only error mode that we've ever handled anyway.

ragesoss (commented):
Do you think the `about:blank` behavior for the `type` is a bug?

lib/wiki_pageviews.rb
gabina: At first glance, I thought that `about:blank`as `type` was a kind of bug on the API side, as I didn't find any specific documentation for that and it wasn't pretty meaningful for me. That's why I decided to rely on the `detail` instead, I thought that maybe the `type`  would change in the near future. However, I think we can rely on the `status` as you mentioned in your other comment.

lib/wiki_pageviews.rb
gabina: hmm good point. I wasn't able to reproduce any request returning `https://www.mediawiki.org/wiki/HyperSwitch/errors/not_found` as `type`. If I understand correctly, [this request](https://wikimedia.org/api/rest_v1/metrics/pageviews/per-article/en.wikipedia/all-access/user/Selfie/daily/2015010100/2015010200) should be one of those that returned that type. So I think that your point about only relying on the 404 status makes total sense. Done on 41a142faab1455171a62a114fd1b9c8a63fb8381. Two specs are failing but I guess those are flaky tests?

spec/lib/wiki_pageviews_spec.rb
```diff
@@ -47,14 +47,12 @@
...
 
         it 'does not raise an error' do
-          stub_request(:any, /.*wikimedia.org.*/)
-            .to_return(
-              status: 404,
-              body: '{"type":"https://mediawiki.org/wiki/HyperSwitch/errors/not_found"}'
-            )
-          expect { subject }.not_to raise_error
+          VCR.use_cassette 'wiki_pageviews/views_for_article' do
+            expect { subject }.not_to raise_error
```
gabina, line 54: I don't think that stubbing the request makes sense now. Instead of that, I set start and end date out of the "allowed data range"

ragesoss:
LGTM!

Issue #5486, opened 2023-08-22
Add 'Get Help' button to training modules


ragesoss:
### Current Behavior:

For logged in users, in `wiki_education` mode, the 'Get Help' button appears in the Course navbar (whenever a user is on a course page). However, when a student launches a training module from their course page, the Course navbar is not part of the training interface and there is no 'Get Help' button.

### Desired Behavior:

For logged in users, a 'Get Help' button should also appear when going through training modules.

### Additional context:

Currently, the training navbar is rendered server-side (while the main navbar at the top is rendered in React on the `nav_root` element, and the training slides are rendered in React below the training navbar on the `react_root` element).

The best option will probably be to refactor the training pages so that the training navbar also renders in React.

The behavior of the 'Get Help' button might need to be a little different, because on Course pages, it shows different options depending on the staff roles assigned to the course. It should probably just default to sending alerts to the Wiki Expert for the user's course (which could either be inferred server-side based on the user and which courses they are part of, or a course ID could be supplied). One way to get the relevant course would be to extract the slug from the `data-return-to` property of the React root (which is included in the slide_view html).

mapra99:
@ragesoss hey! I'd like to look into this

justiceotuya:
@ragesoss how do we navigate to the course page. cant seem to see that. also I have a training page that is not populated. is this the intended behaviour?

<img width="1800" alt="image" src="https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/17678464/95555726-5d2e-4b23-938d-9d4e467cef10">

ragesoss:
@justiceotuya see https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/user_roles.md#loading-training-modules

To enable the Get Help button on a course page, you also need to be in `wiki_education: true` mode (via application.yml).

sharon-odhiambo:
Hello, I'll be working on this issue.

Ritesh17rb:
Has this issue been resolved @ragesoss ?

ragesoss:
@Ritesh17rb no, this is still open.

om-chauhan1:
Working on this

JiyaGupta-cs:
@ragesoss I guess the ```data-return-to``` way is not correct for all the possibilities, since if the user doesnt get to that training via any course then it won't include the course,
[...]

PR #5519, opened 2023-10-23
Add Get Help Button to Training Pages

sharon-odhiambo:
## What this PR does
Whenever a user is logged in wiki-education mode, they are able to access the get help button. The button is however not available when the user launches the training modules from the course page. This PR solves that problem:
Refactoring the training libraries page from the server side rendering by refactoring the index.html.erb to a React component
Refactoring a single library page from the server side to the client side
Adding the Get help button to a single library page and training module slides
Modifying the Get help button to default it to notifying the expert whenever a user clicks the get help button from the training pages.

This PR addresses issue [5486](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5486)

## Screenshots
Before:

After:


sharon-odhiambo:
Hi @ragesoss ,
 after some time I've managed to make some progress on this page. However, I'm not able to get all the data when I'm on training pages but when I maneuvre to other pages I have the data.
If you are able to assist on this I'll appreciate. I'm still debuggin in the meantime. Kindly let me know if I'm on the right direction with this issue.

sharon-odhiambo:
Hi @ragesoss . I have the data in the local. The one I meant was for instance when I'm on training page and I want to access the state, I kind of find empty keys without their values or null values but when I move to like a course page, I have the data. It's probably how I'm passing data and all. 
Thanks for the help. Yes I agree that a PR for just the conversion would be easier before I do the Get Help. Lemme make a separate one. Thank you!

ragesoss:
@sharon-odhiambo I see. That data gets loaded into the Redux store on course pages because it gets fetched by the Course component when it mounts, and/or because it's data embedded in the HTML of course pages that gets parsed by React. Most of that data is specific to the individual course.

I think the strategy here should be:
1. Re-implement navigation in React, which may or may not require fetching or embedding additional data for use via Redux, but if it requires any, it should be pretty minimal, and perhaps would be best done by embedding it in the HTML. Search the codebase for ` #react_root{data:` to find examples of data embedded in the HTML that is intended for use in React.
2. Update the GetHelp component so that it can still function with as little data as possible. In particular, it will probably need to be able to work without access to a `course` object, because we don't necessarily know what course is relevant when a user is on a training page. (If they are only in a single course, we could infer that it's the relevant course, but for a user in multiple courses or no courses at all, we still want this work.)
3. Add an endpoint to fetch whatever additional data is needed, if it can't be added to the HTML. (There might not be any additional data needed that can't be included the HTML.)

PR #5523, opened 2023-10-24
refactor training pages to allow rendering on client-side instead of …

sharon-odhiambo:
## What this PR does
Whenever a user is logged in wiki-education mode, they are able to access the get help button. The button is however not available when the user launches the training modules from the course page. This PR solves that problem:
Refactoring the training libraries page from the server side rendering by refactoring the index.html.erb to a React component
Refactoring a single library page from the server side to the client side

This PR addresses issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5486

Screenshots

After:
This video shows the affected pages functionalities and visual appearance.
[Screencast from 10-24-2023 02:09:35 AM.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/88780311/f6c4f8ce-ebc6-4554-a211-d79e9ae07dde)


sharon-odhiambo:
Hi @ragesoss . This PR is ready for review.

ragesoss:
Thanks! This is a deeper React conversion than what I was suggesting — it converts the entire Training and Module views, rather than just implementing the nav in React and leaving the rest as server-rendered.

This is potentially good, and it looks like you've basically got it working!

One thing I notice as I try it out is that the 'No Libraries' message flashes briefly whenever you load the training index. The first render of any of these pages should avoid flashing any temporary messages like that.

app/assets/javascripts/actions/training_actions.js
```diff
@@ -2,12 +2,67 @@ import { extend } from 'lodash-es';
...
 
+export const fetchTrainingLibraries = () => async (dispatch) => {
+  try {
+    const response = await request('/training.json');
+    if (!response.ok) {
+      const data = await response.text();
+      response.responseText = data;
+      throw response;
+    }
+    const data = await response.json();
```
TheTrio, line 20: this can be extracted to a function — its used in a few places even before this PR

app/assets/javascripts/training/components/search_results.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+const SearchResults = ({ slides }) => {
+  if (slides && slides.length > 0) {
+    return (
+      <ul className="training-libraries no-bullets no-margin action-card-text">
+        {slides.map((slide, index) => (
+          <li key={index}>
+            <a href={`https://dashboard.wikiedu.org/training/${slide.path}`} target="_blank">
```
TheTrio, line 9: don't think this is correct? The root url will not always be `https://dashboard.wikiedu.org/`

app/assets/javascripts/actions/training_actions.js
sharon-odhiambo: Hi @TheTrio , sure I can do that.

app/assets/javascripts/training/components/search_results.jsx
sharon-odhiambo: Hi, while I didn't make any changes on this and not very sure, I think it will always be that url because the slides controller that has the data is using the TrainingSlide model.

sharon-odhiambo:
> Thanks! This is a deeper React conversion than what I was suggesting — it converts the entire Training and Module views, rather than just implementing the nav in React and leaving the rest as server-rendered.
> 
> This is potentially good, and it looks like you've basically got it working!
> 
> One thing I notice as I try it out is that the 'No Libraries' message flashes briefly whenever you load the training index. The first render of any of these pages should avoid flashing any temporary messages like that.

Hi @ragesoss . Sure, lemme check this out. And thank you for the feedback.

app/assets/javascripts/training/components/training_module_handler.jsx
```diff
@@ -39,24 +39,12 @@ const TrainingModuleHandler = createReactClass({
       );
     }
     );
-    let moduleSource;
-    if (this.props.training.module.wiki_page) {
-      moduleSource = (
-        <div className="training-module-source">
-          <a href={`https://meta.wikimedia.org/wiki/${this.props.training.module.wiki_page}`} target="_blank">{I18n.t('training.view_module_source')}</a>
-          <br />
-          <a href={`/reload_trainings?module=${this.props.training.module.slug}`}>{I18n.t('training.reload_from_source')}</a>
```
ragesoss: I think these messages are still relevant, they just shouldn't be shown until after the data has been loaded. If any empty response has already been received from the server, at that point we should show relevant 'missing' messages.

sharon-odhiambo:
Hi @ragesoss, so I was working on getting the module progress and I think this is ready now. However, I have 3 specs failing, in relation to the flashing messages. Initially the messages were only showing when @libraries.empty? is truthy. Currently though, that also seems to be truthy while the data is loading. Can you kindly assist on how I can counter that? I've moved the messages rendering to the controller as before but the specs are still failing.

ragesoss:
Hmm... I'm not sure. On the server, @libraries ought to be set by the time it renders, as Rails will wait until the data is loaded in that controller code before moving on. However, I think it would be ideal to move the message to React, and just not show it until after the data has been fetched and received... then only show it at that point if an empty array was received for Libraries.

ragesoss:
Hmm... I'm not sure what is going wrong. I suggest running the failing test locally with headless mode turned off (by removing the `headless` arg in rails_helper.rb).

ragesoss:
@sharon-odhiambo what's the status of this? it looks like some of the training-specific tests are still failing; were they passing for you locally? anything you are stuck with that i can try to get you unstuck with?

sharon-odhiambo:
Hi @ragesoss , the tests weren't passing locally when I run them but when I check the page when my database is empty the messages appear for both wiki edu and non-wiki edu mode. I'm not sure how to go about it from here.
I'll appreciate any help.

ragesoss:
Sorry, I didn't get a chance to dive into this to provide advice to get you unstuck. I'm going to close this now, as it overlaps with some work we're doing with the trainings currently, so it will be best to start fresh if/when someone returns to it.

PR #5528, opened 2023-11-07
Add admin view to copy courses from production instance of dashboard (#5528)

### LINKED ISSUE #4911: Create a way for admins to copy courses from a production instance of the Dashboard

gabina:
## What this PR does
This PR creates a new admin view to dump course data from a production dashboard to the local environment.
In order to do this a new copy course service was created, based on the already existing `setup/copy_course_from_production.rb` command.

The PR also fixes an existing bug on the command related to `update_logs` keys being parsed as string instead of integers.

Closes #4911

## Screenshots
Before:
No admin view

After:
Example of success and failure:

[example.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/8717242/f880547a-7707-4029-8583-9e0165cd5cf1)

There are a lot of duplicated code in `setup/copy_course_from_production.rb` and the new copy course service. We should either delete the command or make a refactor on it to use the sevice.
It also looks like there is some code duplication in `setup/populate_dashboard.rb` that we may want to refactor.


app/views/admin/index.html.haml
```diff
@@ -14,21 +14,23 @@
   %h2 Useful links
```
gabina, line 14: I ordered the links alphabetically

setup/copy_course_from_production.rb
```diff
@@ -50,3 +51,8 @@ def make_copy_of(url)
   pp "http://localhost:3000/courses/#{course.slug}"
 end
 
+# When parsing update_logs from flags, keys are set as strings instead of integers
+# This causes problems, so we need to force the keys to be integers.
+def fix_update_logs_parsing(update_logs)
+  update_logs.transform_keys(&:to_i)
+end
```
gabina, line 58: There is a lot of code duplication, do we want to keep this command?

gabina:
@ragesoss this PR is ready for review whenever you have some free time. Failing tests don't seem to be related to the PR changes.

setup/copy_course_from_production.rb
ragesoss: No, this file should be removed.

app/services/copy_course.rb
```diff
@@ -0,0 +1,98 @@
...
+  def copy_main_course_data
+    # Extract the attributes we want to copy
+    params_to_copy = %w[school title term description start end subject slug timeline_start
+                        timeline_end type flags]
+    copied_data = {}
+    params_to_copy.each { |p| copied_data[p] = @course_data[p] }
+    @home_wiki = Wiki.get_or_create(language: @course_data['home_wiki']['language'],
+                                    project: @course_data['home_wiki']['project'])
+    copied_data['home_wiki_id'] = @home_wiki.id
+    copied_data['passcode'] = 'passcode' # set an arbitrary passcode
```
ragesoss, line 33: Since we could potential use this in production to copy from one server to another, I think it will be better to use a random passcode via `GeneratePasscode.call`.

app/views/admin/index.html.haml
```diff
@@ -14,21 +14,23 @@
...
   %ul
     %li
-      %a{href: '/article_finder'} Article Finder
+      %a{href: '/faq/new'} Add new FAQ
     %li
-      %a{href: '/usage'} Usage
+      %a{href: '/article_finder'} Article Finder
     %li
-      %a{href: '/sidekiq'} Job queue
+      %a{href: '/copy_course'} Copy Course from Another Dashboard Server
```
ragesoss, line 21: 'Copy course from another server'. No need for extra capitalization, and I think 'Dashboard' is redundant.

app/views/shared/_flash.html.haml
```diff
@@ -1,5 +1,5 @@
 - if flash[:notice]
-  .notification= flash[:notice]
+  .notification= flash[:notice].html_safe
```
ragesoss, line 2: What is this change for?

ragesoss:
This looks great overall.

app/views/shared/_flash.html.haml
gabina: oh I cherry-picked this from the [previous PR](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/4914/files#diff-62220853e559babd01ed5e0d5717a56f1ab4f0a8d1f0d8907e48dbed27266501). 
We are using `flash[:notice]` for rendering the success message in the controller, and that message has a link to the new course (which is added using `<a>` tag). It looks like we need to add the `html_safe` to indicate that the string should be considered safe HTML content and should not be escaped when rendered in a view. Without the `html_safe`, the link doesn't render properly.

app/views/admin/index.html.haml
gabina: Done on 048de6856118b649e1f5d97e595e75c18fdaf3fa

app/services/copy_course.rb
gabina: Done on 65188dad85135cd0f29871dc2195194f4d13eb81

setup/copy_course_from_production.rb
gabina: Done on fff45d60a71334b38b007bfaab1c18b73c09c91d

Issue #5532, opened 2023-11-14
Create an Independent Program - quits when trying to select hours/minutes before the date.


ujjwalpathaak:
### What is happening?
When trying to create a new program and selecting dates for start/end of activity tracking etc... when not done in CHRONOLOGICAL ORDER quits the process and returns to main screen.

### To Reproduce

> Steps to reproduce the behavior:
> 1. Go to 'Dashboard'
> 2. Click on 'Create an Independent Program'
> 3. Click on 'Basic Program'
> 4. Fill in details. Click Next
> 5. Try selecting minutes before selecting the hours or hours before selecting the date.
> 6. See error - it would quit the process

![image](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/92005669/86233b1f-b870-4451-b55a-f894f8a8d685)

### Expected behavior
> The other fields should be disabled while the prerequisite isn't met

ujjwalpathaak:
Can I work on this issue? I have figured out the problem and thinking about a solution for the same.

ujjwalpathaak:
Hi @ragesoss can you have a look at this and provide any feedback necessary

ragesoss:
Nice job spotting this bug! I'm surprised we never caught it before.

PR #5533, opened 2023-11-14
Fix/create an independent program quits

ujjwalpathaak:
## What this PR does
#5532 - If you try to set the time before picking a date, the process will stop and take you back to the main screen. This happens because, without a date, the function can't check the time format properly.

To make things smoother for users, the time fields will be inactive until you choose a date. This way, you won't be able to set the time until you've selected a date first. Once you pick a date, then you can set the time. This should make the whole process more intuitive.

## Screenshots
Before:
image

After:
image

### DIFF
--- a/config/locales/se.yml
+++ b/config/locales/se.yml
@@ -73,6 +73,9 @@ se:
     cancel: Gaskkalduhte
     creator:
       save_cloned_course: Vurke
+    student_editors: Geavaheaddjit
+    students: Geavaheaddjit
+    students_short: Geavaheaddjit
   editable:
     cancel: Gaskkalduhte
     save: Vurke
@@ -106,6 +109,7 @@ se:
   uploads:
     file_name: Fiilanamma
   users:
+    editors: Geavaheddjiid mearri
     first_name: Ovdanamma
     last_name: Goargu
     loading: Viežžamin geavaheddjiid...


ragesoss:
Looks like there are some linting errors.

I tested it out manually, and it looks like this will work.

ujjwalpathaak:
I have resolved the linting issues which where mentioned.
image

ujjwalpathaak:
@ragesoss I have made the changes.

ragesoss:
It looks like there might be a behavior change that is affecting one or two the tests. Can you check whether the two feature tests that failed in CI also fail locally for you? If so, you may need to adjust the tests to work with the new datetime behavior.

ujjwalpathaak:
Hello, @ragesoss I encountered several issues with my previous approach due to how date input was being handled.

I've taken a different approach. Now, there's a reducer call to notification.js, and an error message is prompted, saying to enter the date before the time.

While my yarn test cases are passing successfully, I'm feeling a bit uncertain about the rspec test cases. I am new to unit testing and Ruby.

According to me, the test cases affected by my changes are passing, but it would be great if I could take a look at the CI to confirm whether any further adjustments are needed.

image

ujjwalpathaak:
Hello @ragesoss all the checks have passed for the last commit. Can you please have a look at my approach?

app/helpers/application_helper.rb
```diff
@@ -41,7 +41,11 @@ def fingerprinted(path, filename, file_prefix = nil)
   def css_fingerprinted(filename)
     manifest_path = "#{Rails.root}/public/assets/javascripts/manifest.json"
     manifest = Oj.load(File.read(File.expand_path(manifest_path, __FILE__)))
-    manifest[filename].split('/').last
+    if manifest[filename]
```
ragesoss, line 44: This seems unrelated. What is this change for?

ragesoss:
I think the error notice is a little too aggressive. A better option might be to automatically set the date to today if the user clicks the time picker before the date is set.

app/helpers/application_helper.rb
ujjwalpathaak: I'll stash that change, it was for one of the test failing locally for me but i guess it’s unnecessary and was probably a local issue

ujjwalpathaak:
@ragesoss Hey i guess i made a mistake. Can you please have a look and please resolve this?

ragesoss:
Can you explain a little more? What should I have a look at?

ujjwalpathaak:
I was trying to make some changes in my fork but I ended up force merging this pull request and now it shows it's merged and closed.

image
image

ujjwalpathaak:
> I think the error notice is a little too aggressive. A better option might be to automatically set the date to today if the user clicks the time picker before the date is set.

So should I raise a new pull request with these changes?

ragesoss:
Yes, that's fine.

Issue #4466, opened 2021-05-04
Standarize the usage of icons in the UI


ragesoss:
Icons are implemented in different ways throughout the app. There's an 'icon font' (`app/assets/fonts`) which hasn't been updated in a long while but it still used for some icons, via CSS classes. There are also a variety of SVG icons (`app/assets/images`) which are used as background images for CSS classes, and there are some SVG icons defined directly in JSX components (eg, `enroll_card.jsx`).

Identify the best pattern for handling icons throughout the app and standardize it so that all icons are implemented the same way.

### Related Files & Code:

* `app/assets/stylesheets/_icons.styl`

aaditya9899:
we can use fontawesome for that i think so?

aaditya9899:
yeah I thought so while implementing fontawesome but installing it will bring all the icons from the library which might reduce page efficiency and better alternative might be using react-icons library as it will just import the necessary icons and that too from different websites!! 

here's the link , do tell me how this sounds to you.

[https://react-icons.github.io/react-icons]

aaditya9899:
yeah I get to know while implementing fontawesome that installing it will bring all the icons from the library which will reduce page performance and better alternative might be using react-icons library as it will just import only the necessary icons we use in the page and also provide one additional functionality of choosing icons from different library as well like fontAwesome, Feather, Bootstrap etc !!!

here's the link , do tell me how this sounds to you.

https://react-icons.github.io/react-icons

ragesoss:
We want to keep the current icons we're using, and just standardize how they are implemented. It looks like `react-icons` is a library for importing from an existing collection of icons, not taking custom icons and integrating them into an app.

aaditya9899:
Oh okay understood i will tell you if i start working on this one.

prashulkale:
if you allow me to contribute on the feature, I will update it with REMIX ICON

ragesoss:
I don't know anything about Remix Icon. If it can be used for any icon assets (so that we can continue using the current icons we have, and just standardize them via this library), then I'm open to it. I suggest starting with a demo showing how it would be implemented for a single icon, one of the ones we already use.

ujjwalpathaak:
There are ways the icons are being used as per the issue - 
**1. icons directly mentioned in JSX components -** identify JSX components to find instances where icons are directly mentioned. Convert these icons to SVG files and use them just like other SVGs being used.
[...]

PR #5538, opened 2023-11-28
Issue4466/inline svg to svg files and trying out font files conversion

ujjwalpathaak:
reference #4466 

## Changes
Inline SVGs were being used in the files -
```
app/assets/javascripts/components/common/notifications.jsx
```
```
app/assets/javascripts/components/enroll/enroll_card.jsx
```
It was basically a simmilar and a smaller version of the close "X" SVG previously being used so I made a new file ``` app/assets/images/search_icon.svg ``` and set it as a background in ``` _icons.style``` for the class ```.icon-close-small```.

This is the way majority of icons are currently being used - in svg format and being set as a background for different classes and the classes are then being used in jsx components.

## Screenshots
Before:
image
After:
image

## Font Files approach
I converted the ```icons.ttf``` to **.sdf** format, then I was able to export the SVGs of the icons being used using FontForge software.

Now i have all the SVGs that were being used like this - 
```
@font-face {
  font-family: "icons";
  src: url('../fonts/icons.eot');
  src: url('../fonts/icons.eot?#iefix') format('eot'),
    url('../fonts/icons.woff') format('woff'),
    url('../fonts/icons.ttf') format('truetype');
  font-weight: normal;
  font-style: normal;
}

.icon:before {
  display: inline-block;
  font-family: "icons";
  font-style: normal;
  font-weight: normal;
  line-height: 1;
  -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale;
}

.icon-search:before { content: "\EA08" }
```
So to standardize the process I used the search icon SVG in "icon-search" class replacing the content attribute to background-image and then used that class in JSX file.

Like this other icons can also be converted from font files to svgs *this was done for testing purpose*


ujjwalpathaak:
@ragesoss Please have a look at this and let me know if any changes are necessary.

Issue #4234, opened 2020-10-09
FAQ loads without default questions


rajat2502:
> Thanks for reporting a problem! Please delete any line that begins with a '>' (including this one) and replace with your own content.

### What is happening?
> The FAQ page loads without any default questions if the search icon in the navbar is clicked without any search text.

### To Reproduce

> Steps to reproduce the behavior:
> 1. Go to 'https://dashboard.wikiedu.org/'
> 2. Click on 'Search icon'
> 3. FAQs page will open in a new tab without any questions

### Expected behavior
> Show the default questions or the top questions.

### Screenshots
> Here is a video showing the same. 
https://dashboard.wikiedu.org/

dldisha:
I'll work on this issue.

prathamVaidya:
I think this Issue can be closed. Looks like UI has already been updated.

prathamVaidya:
Also @ragesoss  can you send me a slack invite. Can't find a untouched issue to work on. 
Email: iamprathamvaidya@gmail.com

ragesoss:
@prathamVaidya this issue is still present. Compare:

* https://dashboard.wikiedu.org/faq - shows default questions
* https://dashboard.wikiedu.org/faq?utf8=%E2%9C%93&search=&source=nav_ask_form - no default questions shown for empty search

I've sent a slack invite.

prathamVaidya:
Ok, I can work on this issue.

PR #5539, opened 2023-11-29
Fixed Bug #4234

prathamVaidya:

## What this PR does
Fix Issue #4234 : FAQ loads without default questions 

## Screenshots
Before:
image

After:
image

Changes:

- Added Empty String check for `query` and set `Nil` if its empty. This will Line 10 `@topic_slug` assignment and `DEFAULT_TOPIC` will be assigned to `@topic_slug`.

### DIFF
--- a/app/controllers/faq_controller.rb
+++ b/app/controllers/faq_controller.rb
@@ -6,7 +6,7 @@ class FaqController < ApplicationController
   DEFAULT_TOPIC = 'top'
 
   def index
-    @query = params[:search]
+    @query = (params[:search] == "") ? nil : params[:search]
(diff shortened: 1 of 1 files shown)


app/controllers/faq_controller.rb
```diff
@@ -6,7 +6,7 @@ class FaqController < ApplicationController
   DEFAULT_TOPIC = 'top'
 
   def index
-    @query = params[:search]
+    @query = (params[:search] == "") ? nil : params[:search]
```
ragesoss, line 9: I think this might be easier to understand like `@query = params[:search] if params[:search].present?`

prathamVaidya: Done @ragesoss

PR #5540, opened 2023-11-30
Improvements in debian setup script #1709

prathamVaidya:

## What this PR does
Improvements in Debian Setup Script. As discussed in Issue #1709.

## Changes

1. Added a fallback package `libmariadb-dev` to support latest debian versions where alternate package `libmariadbclient-dev` was not available by default respositories.
2. Improved Bash Function output_line :  used `%s` formatting over direct concatenation so that special characters like `%` can also be printed without conflicting with formatters.
3. Optimized the execution of MySQL commands by removing the redundant `-p` flag. Since `sudo` is used to execute SQL commands with root privileges, there is no need for the additional `-p` flag, avoiding repetitive password prompts.

1. Was the `-p` flag in `mysql` command serving any purpose that I missed?
2. Should I also update other os scripts as well?

### DIFF
--- a/setup/deb-setup.sh
+++ b/setup/deb-setup.sh
@@ -131,7 +131,7 @@ fi
 printf "[*] Creating Databases... \n"
 echo "CREATE DATABASE IF NOT EXISTS dashboard DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_general_ci;
       CREATE DATABASE IF NOT EXISTS dashboard_testing DEFAULT CHARACTER SET utf8 DEFAULT COLLATE utf8_general_ci;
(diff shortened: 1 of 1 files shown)


prathamVaidya:
@ragesoss can you review?

ragesoss:
LGTM

ragesoss:
The revision importer spec has been very flaky since I updated it to loop through a fairly large set of languages.

prathamVaidya:
Okay, should i fix the same debian script issues for other os script as well? I might not be able to test them but the changes are minimal and should not break anything.

ragesoss:
If you can't test, better to leave them as they are I think.

Issue #5524, opened 2023-10-24
Admins should be able to email the instructors of a course


ragesoss:
### Current Behavior:

When a Wiki Education staff members wants to contact all the instructors of a course, the easiest way is currently to copy each of their email address and use a separate email client. It would be more convenient to be able to send an email message directly from the course page.

### Desired Behavior:
There should be a button in Actions that opens a text box and allows an admin to write a message that will be emailed to all the instructors of the course. The email should look good, with minimal formatting using similar styling to existing emails sent by the Dashboard, and should have a reply-to for the admin who sends the message.

tushar-glitch:
Hi @ragesoss, looks like an interesting one to solve! Can I work on this issue?

ragesoss:
@tushar-glitch sure go for it.

prathamVaidya:
@ragesoss are editors and instructors same? What page do you want the Button to send notification to instructors. I was thinking maybe here (if editor means instructors):

![image](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/61202986/c8c1ea82-f9ff-460d-a3cf-60d98c267d48)

ragesoss:
@prathamVaidya there are two main production contexts: Programs & Events Dashboard ( outreachdashboard.wmflabs.org, corresponding to the `wiki_education: false` setting) and Wiki Education Dashboard (dashboard.wikiedu.org, corresponding to `wiki_education: true`). There is some different terminology between them.

'Editor' on P&E Dashboard is the same as 'Student' in Wiki Education Dashboard (role 0 in a CoursesUsers record), while 'Instructor' is the same as 'Facilitator' (role 1 in a CoursesUsers record).

For this issue, the feature is only intended for Wiki Education Dashboard, and the Home tab (probably in the Actions section) is where I think the button should go.

ragesoss:
(I think that FAQ is still accurate; it is specifically for Wiki Education Dashboard.)

prathamVaidya:
Thanks @ragesoss, Please check the PR.

Brajesh1213:
I would like to work on this issue may i

ragesoss:
@Brajesh1213 there is already an open PR that I have not reviewed yet. Issue #5530 is a good one. Please only work on one at a time.

prathamVaidya:
Issue can be closed. Feature added in https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5545

PR #5545, opened 2023-12-05
feat: Implement email notifications for instructors

prathamVaidya:
## What this PR does

This pull request addresses the issue #5524 by introducing a new feature that allows Wiki Education staff members to easily send email messages to all instructors of a course directly from the course page. The new feature includes the following changes:

- [x] Created Instructor Notification Mailer
- [x] Created new Alert Type `InstructorNotificationAlert`
- [x] Added Test and Preview for Instructor Notification Mailer
- [x] Email Template with minimal formatting, using a similar style to existing Dashboard emails.
- [x] Added a button in the 'Actions' section for sending email messages to instructors.
- [x] Implemented a text box for admins to compose the email message.
- [x] Included a 'reply-to' functionality for the admin who sends the message.

## Screenshots
Before:
Screenshot 2024-01-10 at 15-44-57 Maths

After:
Screenshot 2024-01-10 at 15-45-18 Maths

Screenshot 2024-01-10 at 15-45-32 Maths

Current Email Template Preview:

Screenshot from 2024-01-10 15-52-14

None


prathamVaidya:
@ragesoss Please give clarity on following points:

1. From user, currently set to `SpecialUsers.classroom_program_manager`
2. BCC users, currently set to `@alert.bcc_to_salesforce_email`
3. This is the minimal HTML template right now. Let me know any changes.

image

4. Should I push modified `schema.rb` and newly created migration file as well?

prathamVaidya:
Also will it be any better if we just make the feature more versatile. Allowing Admins to choose from the available audience like students, instructors and any other role before sending a notification.

ragesoss:
> From user, currently set to SpecialUsers.classroom_program_manager

This should be whichever admin is writing the message. In this case, I think that should be handled with the `user` column of the Alert record.

ragesoss:
It looks like you have not implemented the flow for creating the Alert from the frontend and then sending the email? I'm guessing that will involve a new Rails controller, along with some frontend code for posting to that controller.

ragesoss:
I think ideally, there shouldn't be any text for the subject or body of the email that is hard-coded. Both the subject line and contents should be specified by the admin sending it.

prathamVaidya:
> I think ideally, there shouldn't be any text for the subject or body of the email that is hard-coded. Both the subject line and contents should be specified by the admin sending it.

Sure, but if we'll directly take subject and body from the admin. That will open a possibility that admin will have access to send completely modified email to the course participants. Like they can also send an email for an scam or impersonating officials. I am just discussing about a possibility, otherwise it really depends on how much we trust the Admins.

prathamVaidya:
> > I think ideally, there shouldn't be any text for the subject or body of the email that is hard-coded. Both the subject line and contents should be specified by the admin sending it.
> 
> Sure, but if we'll directly take subject and body from the admin. That will open a possibility that admin will have access to send completely modified email to the course participants. Like they can also send an email for an scam or impersonating officials. I am just discussing about a possibility, otherwise it really depends on how much we trust the Admins.

@ragesoss The feature is ready. 

However, right now the subject is hard coded and message is added in the pre-defined email template. As per your preference you don't want it to be hard coded. I can do that but currently I am using `Alert` Modal to achieve this feature and it does not have subject field. Now because `Alert` modal is used by many other types I just wanted to confirm if you want me to proceed with modification of `Alert` because it might open a vulnerability (Discussed in the previous comment).

ragesoss:
Thanks! I'm not worried about admin access as a security issue; admins are few and highly trusted, and already have access to things that would cause major problems if abused.

I think the subject could be stored in the Alert table without any modifications at the database level. The `details` column is a serialized hash that can be used to store arbitrary information such as a subject string. See `BlockedEditsAlert`, for example, which saves data about a Wikipedia block in the `details` field and retrieves it by digging into the hash.

prathamVaidya:
Sure I will update it to use a custom subject. One last thing about email body: should I use the message and add in the email template which includes footer and header or take entire email body from frontend only.

ragesoss:
You should use a template that has a footer and header and uses the same styling as other mailers.

prathamVaidya:
As per your comments, I have updated the code to use custom subject utilizing `details` field. Let me know if you have any suggestions or improvements.

Also, because I have added a new Alert type. There is a new migration file that i have committed. I am still unsure if I should push the updated `schema.rb` or not.

image

ragesoss:
There should be no need to include a migration or schema change. The `Alert` table uses the Rails 'Single Table Inheritance' feature, so we can add subtypes of Alert without making any schema changes.

prathamVaidya:
Okay @ragesoss . I have removed the migration file.

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -191,6 +192,12 @@ const AvailableActions = createReactClass({
       ));
     }
 
+    if (user.admin) {
```
ragesoss, line 195: This should only appear in Wiki Education mode (when `Features.wikiEd` is truthy).

app/assets/javascripts/components/overview/notify_instructors_button.jsx
```diff
@@ -0,0 +1,120 @@
...
+};
+
+const initialState = {
+  subject: '',
+  message: '',
+  status: 'DEFAULT', // DEFAULT, PENDING, FAILED
+  error: null,
+};
+
+export default function NotifyInstructorsButton(props) {
```
ragesoss, line 23: For style consistency, I'd prefer to use ES6 arrow functions for functional components, rather than traditional function syntax.

app/assets/javascripts/components/overview/notify_instructors_button.jsx
```diff
@@ -0,0 +1,120 @@
...
+  const dispatch = useDispatch();
+
+  const [modalVisibility, setModalVisibility] = useState(false);
+  const [notification, setNotification] = useState(initialState);
+
+  const createNotificationApi = () => {
```
ragesoss, line 29: The  preferred convention in this codebase is to keep API interaction code separate from components, in an appropriate file in `actions` (like with the confirm_actions.js file used in this component).

app/assets/javascripts/utils/api.js
```diff
@@ -600,7 +600,23 @@ const API = {
         label,
       };
     });
-  }
+  },
+
+  async createInstructorNotificationAlert(courseId, subject, message) {
```
ragesoss, line 605: I don't want new actions added to api.js, as it is already much too big. We've been gradually moving things out of this file, and I don't want anything new added to it. This along with the other API request/response code should be moved to a new file.

app/mailers/instructor_notification_mailer.rb
```diff
@@ -0,0 +1,27 @@
...
+class InstructorNotificationMailer < ApplicationMailer
+  def self.send_email(alert)
+    return unless Features.email?
+    email(alert).deliver_now
+  end
+
+  def email(alert)
+    @alert = alert
+    set_email_parameters
+    params = { to: @instructors.pluck(:email),
```
ragesoss, line 12: This should probably BCC to the sender, to make it easy to know that it went through.

app/models/alert.rb
```diff
@@ -97,8 +98,11 @@ class Alert < ApplicationRecord
     UnsubmittedCourseAlert
     UntrainedStudentsAlert
     SandboxedCourseMainspaceAlert
+    InstructorNotificationAlert
   ].freeze
 
+  # InstructorNotificationAlert public or private?
```
ragesoss, line 104: Private

config/routes.rb
```diff
@@ -218,6 +218,9 @@
       get 'resolve'
       put 'resolve'
     end
+    collection do
```
ragesoss, line 221: What does `collection` do versus `member`? This is a bit of Rails I'm not familiar with.

ragesoss (commented):
Nice work! This looks broadly like what I would expect, code-wise. Once the comments are addressed, I'll test it out locally.

config/routes.rb
prathamVaidya: `member` always create routes that require `id` to perform some action. It usually used for a route that actions on a unique instance.

`collection` is used for performing operations over a group or where there is no ID required. Such as in this case where we need to create Alert and there is no ID yet.

In collection our route becomes `/notify_instructors`, and if `member`  route would have become `/notify_instructors/:id`

prathamVaidya:
@ragesoss As per your feedback I have done the following changes: 

- Separated API Logic outside the component
- Separated Alert API outside the api.js to separate file `alert_utils.js` (I also propose, it will better if we have all API code in separate directory called `apis` or services rather than mixed `utils` directory) 
- Added Check for WikiEd Truthy
- Use Arrow Function
- Added BCC (sender's email)
- `InstructorNotificationAlert` to private

Also I am directly using dispatch in the React Component. Let me know if you want me to separate that as well as functions.

ragesoss:
I tried this out and it works nicely! I noticed a few language and email layout issues that I'd like to have updated, and then it will be ready to merge and deploy.

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
       notification_message: Take our survey for this course.
       link: Take Survey
 
+  course_instructor_notification:
+    notify_instructors: Notify Instructors
```
ragesoss, line 826: Let's call this 'Email instructors'

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
...
+    notify_instructors_feature_description: Send notification alerts to Instructors through emails
```
ragesoss, line 827: 'send a message to instructors via email'

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
...
+    send_notification_button_text: Send Notification
```
ragesoss, line 828: 'Send'

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
...
+    notification_sent_success: "Notification sent to all Instructors of Course %{courseTitle}"
```
ragesoss, line 829: "Message sent to all instructors of %{courseTitle}"

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
...
+    notification_send_error_no_internet: "Unable to send request. Check if you are connected to the Internet."
+    notification_send_error_server: "Server Error: Unable to send notification (%{status} %{statusText})"
+    notification_empty_message: "Subject or Message cannot be empty"
+    confirm_send_notification: "Are you sure you want to send the notification to all the Instructors?"
```
ragesoss, line 833: "Send to all instructors for this course?"

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
...
+    write_subject: "Write Subject"
```
ragesoss, line 834: "subject"

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
...
+    write_message_placeholder: "Write Message Here (Header and footer will be automatically included)"
```
ragesoss, line 835: "message"

config/locales/en.yml
```diff
@@ -822,6 +822,19 @@ en:
...
+    sending_notification: "Sending Notification to Instructors"
```
ragesoss, line 836: "Sending"

app/views/instructor_notification_mailer/email.html.haml
```diff
@@ -0,0 +1,26 @@
+%link{rel: 'stylesheet', href:'/mailer.css'}
+%table.row
+  %tbody
+    %tr
+      %th
+        %table
+          %tr
+            %td.main-content
+              %h1.headline
```
ragesoss, line 9: This is included as the subject line, so it should not be repeated

app/views/instructor_notification_mailer/email.html.haml
```diff
@@ -0,0 +1,26 @@
...
+                = "#{@subject}"
+              %p.paragraph
+                = "Course: "
+                %a{href: @course_link}= @course.title
```
ragesoss, line 13: This should go in a separate section after the main content.

app/views/instructor_notification_mailer/email.html.haml
```diff
@@ -0,0 +1,26 @@
...
+              %p.paragraph
+                = "Dear Instructors,"
```
ragesoss, line 15: Remove this.

app/views/instructor_notification_mailer/email.html.haml
ragesoss: See below for how I think makes sense to include the link.

app/views/instructor_notification_mailer/email.html.haml
```diff
@@ -0,0 +1,26 @@
...
+              %p.paragraph
+                = "#{@message}"
+
+        %table
+          %tr
+            %td.main-content
+              %p.paragraph
+                Best regards,
```
ragesoss, line 23: I'd like this to say: "This is a message from Wiki Education staff for your course [course title linking to course_link]."

It should be styled differently as well, probably in a smaller font, to make it more clear which parts of the email are a human-written message and which parts are from a template.

ragesoss (changes requested):
This looks great. Just some language and styling tweaks and it will be ready to go.

app/views/instructor_notification_mailer/email.html.haml
prathamVaidya: Variant 1: Only Text Size Smaller
image

Variant 2: Added a `.footer` class in stylesheet changing size, color, font.
image

Variant 3: Made main message black and increase `font-weight` to normal.  
image

let me know which one of these satisfies the theme @ragesoss

ragesoss: I think the .footer class one looks the best. I suggest using that one, plus adding additional spacing above and below the main message.

Also, there's no need for the 'Best regard, The Wiki Education team'.

The admin can optionally include a personal closing line, but I would rather not hard-code one.

ragesoss:
There's also a failing spec that is connect to this PR.

app/views/instructor_notification_mailer/email.html.haml
prathamVaidya: Done!
image

ragesoss: Perfect, this looks great.

prathamVaidya:
Added a required method `main_subject` that caused the spec to fail. I'll check again after few mins if anything fails this time and fix it today only.

prathamVaidya:
The spec that I added is now passing but as you are aware, the workflow script still always fails.  
I want to fix this behavior. 
One of the reason i found might be because `feedback_form_spec` is failing.
Should I create a new issue and try to fix this or whats the correct way to fix this issue? @ragesoss

ragesoss:
@prathamVaidya yes, feel free to create an issue for that flaky spec. I think that particular one is failing sometimes because it relies on an API that sometimes is not available (the one that results in the NoMethodError on the Dashboard side), so the right solution is probably to create a new request stub in request_helpers.rb and use that for the spec.

Issue #5546, opened 2023-12-05
Overlapping of campaign_navbar at the desktop view on language selector


Sahityaaryan:
### What is happening?
 Overlapping of campaign_navbar at the desktop view over the language selector at the campaign page

### Expected behavior
 The language selector should be over the campaign navbar

### Screenshots
![image](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/115429795/d1f4acfb-5aeb-4af3-b548-249b62336964)

Sahityaaryan:
@ragesoss verify it and I will make a PR.

ragesoss:
Yeah, I've seen this happen before. The sticky navbars are tricky to make play nicely with the main navbar. A PR is welcome!

Sahityaaryan:
Sure

himadeepthi1:
Hello ! @ragesoss , can I work on this ?

ragesoss:
We already have an open PR for this, which I have not reviewed yet but likely fixes the issue.

himadeepthi1:
OK , then I will look into other issues , @ragesoss  please send slack invite, it would be really helpful

himadeepthi1:
My email address is himadeepthisayani@gmail.com

himadeepthi1:
Hi , @ragesoss , can you please send me slack invite, my email address is himadeepthisayani@gmail.com

PR #5550, opened 2023-12-11
 Overlapping of campaign_navbar at the desktop view on language selector

Sahityaaryan:
## What this PR does

This is to fix the issue #5546
Overlapping of campaign_navbar at the desktop view on the language selector

## Screenshots
Before:
image

After:

image


ragesoss:
This looks like a fine fix that is unlikely to cause unintended problems in any other parts of the UI.

It looks like it introduces some trailing whitespace in two lines. Fix that, and it should be ready to merge.

ragesoss:
@PR4NJ41 yes, please do

PR #5554, opened 2023-12-15
Added test (confirm.spec.js )for one of the Redux Reducers  (confirm.js)

shruti862:
Addresses issue #2082


ragesoss:
Please exclude the changes to schema.rb

shruti862:
@ragesoss  I excluded all the changes made to schema.rb                                                                                            Please review it again

ragesoss:
There are still changes to schema.rb. This PR should have only 1 file changed. I recommending copying your new test file, then resetting your git state to match upstream, and re-adding the file in a new commit. ( We have some git documentation that may be helpful: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/git.md )

Issue #5551, opened 2023-12-13
Comment for UntrainedStudentsAlert is wrong


pppery:
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/dc125aaccb2e2e5b69b5c7174ab8154e2a40e2dd/app/models/alert_types/untrained_students_alert.rb#L23

This seems to have been copied from somewhere else and doesn't make sense here

nitish-singh07:
I am interested in working on this issue.

PR #5555, opened 2023-12-16
Update untrained_students_alert.rb

nitish-singh07:
Address issue NO #5551 

Changes - Correct the Comment for UntrainedStudentsAlert.

### DIFF
--- a/app/models/alert_types/untrained_students_alert.rb
+++ b/app/models/alert_types/untrained_students_alert.rb
@@ -20,7 +20,7 @@
 #  details        :text(65535)
 #
 
-# Alert for a course that has no enrolled students after it is underway
+# Receive an alert when a course has no enrolled students even after it has started.
 class UntrainedStudentsAlert < Alert
   def main_subject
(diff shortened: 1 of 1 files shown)


app/models/alert_types/untrained_students_alert.rb
```diff
@@ -20,7 +20,7 @@
 #  details        :text(65535)
 #
 
-# Alert for a course that has no enrolled students after it is underway
+# Receive an alert when a course has no enrolled students even after it has started.
```
ragesoss, line 23: This is also incorrect. This kind of Alert indicates that a significant portion of enrolled students have not completed their assigned training modules.

nitish-singh07: Can you explain me little bit . what is should be?

ragesoss: It should be something like 'Alert for when the students in a course are behind schedule on their assigned training modules.'

PR #5556, opened 2023-12-16
Added test (confirm.spec.js) for Redux Reducer file( confirm.js)

shruti862:
Addresses issue #2082


shruti862:
@ragesoss ,Would you let me know what's the error behind the failure of this pull request. How can I get the error resolved?

ragesoss:
It looks like something related to the browser in the CI environment has changed, and that's causing many ruby tests to fail. It's definitely unrelated to your PR.

PR #5561, opened 2023-12-18
fixed an error in js code of  assets/javascript/campaign.js file

shruti862:
With previous code I was getting this error (screenshot attached) after clicking on Create Campaign button
error
 Now with the changes I did in the campain.js file ,no longer error exits.

### DIFF
--- a/app/assets/javascripts/campaigns.js
+++ b/app/assets/javascripts/campaigns.js
@@ -93,6 +93,6 @@ window.onload = () => {
   });
 
   if (createModalWrapper?.classList.contains('show-create-modal')) {
-    createCampaignButton.click();
+    createCampaignButton?.click();
   }
 };


ragesoss:
Thank you @shruti862!

This is a very small change that looks like it will be safe to merge, but I'd like to understand more about why the error occurred in the first place.

I assume from the URL in your screenshot that you clicked the 'Create a New Campaign' button on the `/explore` page (as opposed to the one on `/campaigns` which does not add a parameter to the URL). I haven't been able to replicate the error either locally or in production, so I'm curious about what's happening differently in your dev environment.

shruti862:
@ragesoss 
I just went to 'Find Programs' tab (or clicking 'Explore' button on homepage) which led me to /explore page   where after clicking on 'Create a New Campaign' Button this error window pop up .

ragesoss:
Thanks. Do you have an idea why this is happening for you, but not in production or for me locally? The intended behavior when clicking that button is that the modal for starting the creation of a program will pop up immediately; does happen for you, or do you have to click the create button on the `/campaigns` page after clicking the one on `/explore`? (Needing to click again is what I would expect to happen in cases where the optional chaining is relevant.)

shruti862:
@ragesoss I don't know why this error is only occuring in  my local system. I am attaching video clip of how it does happen.
[Uploading Ubuntu [Running] - Oracle VM VirtualBox 2024-01-08 23-37-32.zip…]()

ragesoss:
@shruti862 i'm closing this because i could not replicate the bug and I haven't heard back about what exactly triggers it, but please let me know if you return to it and we can re-open it.

Issue #5547, opened 2023-12-05
Improve how Wiki Education Dashboard counts references added


gabina:
This issue is the main issue for the "Improve how Wiki Education Dashboard counts references added" Outreachy (round 27) project.

Timeline project can be found [here](https://phabricator.wikimedia.org/T352177).
### Current Behavior:
The Dashboard currently gets statistics on references added by fetching the features data supplied by Wikimedia's article quality machine learning models, and comparing the values of one revision with values for the previous revision. This reference counting method is constrained by the availability of article quality machine learning models, which are not accessible for the majority of Wikipedia language editions, including the Spanish language version. That means that not every language has references added metric available.

### Desired Behavior:
References added metric should be available for every language version of Wikipedia. One high priority is to enable reference counting for Spanish Wikipedia, in support of the Spanish #1Lib1Ref campaign.

### Main tasks:
Note: the following tasks are high-level tasks just to keep a general idea of the project progress. More details can be added on the fly.

- [x]  Research the current method for calculating references added.
- [x] Evaluate which API we will use to get the references added number for all the languages  (`misalignment.wmcloud.org/api/v1/quality-revid-features` API is one option, building our own API for this particular case is another one).
- [x] Build our own API to get references and host it in Toolforge (as a conclusion of the previous step).
- [x] Determine where to store data from the new API. Handle database migration if necessary. **Note**: pull request was closed because we decided to use the existing `features` and `features_previous` fields. (PR #5567)
- [x] Develop a streamlined method for making requests to the new API (similar to `LiftWingApi` class).  (PR #5565)
- [x] Implement a way to use the newly created class to import and store data in the database  (similar to `RevisionScoreImporter`).For this, we did the following things:
    -  Refactor the existing `RevisionScoreImporter` (PR #5578).
    - Create a new intermediate class to handle API requests (PR #5594). 
    - Update the `Revision::references_count` method to use the new  data populated in `features` (or `features_previous`) from the reference-counter API (PR #5615)
- [x] Make a [performance report](https://yummy-mercury-c00.notion.site/Performance-report-3f999043f16040f89cb936ad90fc41a1) for the new score import process.
- [x] Make a [quality report](https://yummy-mercury-c00.notion.site/Quality-Report-8137be99d8c9479eac84c1fe9fbe1586) for the new score import process.
- [x] Integrate all changes.
- [x] Clean up code (PR #5675).

akshayw1:
Can I work on this issue?

[...]

PR #5567, opened 2023-12-28
[Outreachy Round 27] Add `references` and `references_previous` fields to Revisions table

gabina:
## What this PR does

The purpose of this PR is to add two new (integer) fields to the `Revisions` table:
- `references`:  to store the number of references existing in a given revision id (`AddReferencesToRevision` class)
- `references_previous`: to store the number of references existing in the previous revision id (`AddReferencesPreviousToRevision` class)

This PR also provides a new regenerated `erd.pdf` file, and updates annotations on revision models and factories to reflect the changes.

Changes in a`pp/models/wiki_content/revision.rb`, `spec/models/revision_spec.rb`, and `spec/factories/revisions.rb` were automatically done by running `annotate --models`.

- Notice that these new two fields will replace `features` and `features_previous` ones. We will need to remove them in the future  (as well as all code related to them).
- `db/schema.rb` was automatically modified when running the migration locally, but manually modified after that, as some changes looked incorrect.


ragesoss:
@gabina If possible, I would like to re-use the existing fields rather than adding these ones and removing the `features` and `features_previous` ones. Those `features` columns are serialized hashes, so they are flexible enough to hold the references counts for new data while co-existing with the features data from already-present revisions in the production database.

gabina:
> @gabina If possible, I would like to re-use the existing fields rather than adding these ones and removing the `features` and `features_previous` ones. Those `features` columns are serialized hashes, so they are flexible enough to hold the references counts for new data while co-existing with the features data from already-present revisions in the production database.

Yes, sure. In that case, I'm closing this PR as it won't be necessary.

PR #5565, opened 2023-12-26
[WIP] [Outreachy Round 27] Create ReferenceCounterApi class to hit new Reference Counter API

gabina:
## What this PR does
This PR is part of the "Improve how Wiki Education Dashboard counts references added" project (read issue #5547).
It adds a new class: `ReferenceCounterApi`. This new class makes requests to the [Toolforge Reference Counter API](https://toolsadmin.wikimedia.org/tools/id/reference-counter).
It also adds some basic specs for it.
## Open questions and concerns
< anything you learned that you want to share, or questions you're wondering about related to this PR >

PR #5578, opened 2024-01-09
[Outreachy Round 27] Refactor `LiftWingApi`  and `RevisionScoreImporter`

gabina:
## What this PR does

It makes some refactors on `LiftWingApi`  and `RevisionScoreImporter` classes.

The `RevisionScoreImporter` is responsible for populating fields such as `wp10`, `features`, `wp10_previous`, `features_previous`, and `deleted` in the revisions. Currently, this involves using the LiftWing API to retrieve data. However, as part of the "Improve how Wiki Education Dashboard counts references added" project, we also need to use the new reference-counter API (see #5565).

The primary idea is to maintain the `RevisionScoreImporter` as the class responsible for populating all revision score fields, regardless of whether the values come from the LiftWing API or the Toolforge reference-counter API. This approach offers benefits, as there is valuable logic that should be shared in the population process, irrespective of the data source. For example, the `get_parent_revisions` method retrieves parent revision IDs, necessary for updating both `wp10_previous` (from the LiftWing API) and `features_previous` (from the Toolforge reference-counter API).

Before this PR, the importer contained a lot of logic specific to the articlequality model data and how to process it (such as calculating the `wp10` value from the scores).

The purpose of this refactor is to move elements too specific to the LiftWing API into the existing `LiftWingApi` class. This way, `RevisionScoreImporter` becomes a class with the logic to populate fields from both `LiftWingApi` and `ReferenceCounterApi`.

### DIFF
--- a/lib/importers/revision_score_importer.rb
+++ b/lib/importers/revision_score_importer.rb
@@ -58,10 +58,6 @@ class RevisionScoreImporter
   ##################
   private
 
-  def model_key
-    @model_key ||= @wiki.project == 'wikidata' ? 'itemquality' : 'articlequality'
-  end
-
   def get_and_save_scores(rev_batch)
     scores = @lift_wing_api.get_revision_data rev_batch.map(&:mw_rev_id)
     save_scores(scores)
@@ -89,10 +85,10 @@ class RevisionScoreImporter
   def save_parent_scores(parent_revisions, scores)
     parent_revisions.each do |mw_rev_id, parent_id|
       next unless scores.key? parent_id
-      article_completeness = scores.dig(parent_id, 'wp10')
+      wp10_previous = scores.dig(parent_id, 'wp10')
       features_previous = scores.dig(parent_id, 'features')
       Revision.find_by(mw_rev_id: mw_rev_id.to_i, wiki: @wiki)
-              .update(wp10_previous: article_completeness, features_previous:)
+              .update(wp10_previous:, features_previous:)
     end
   end
 
--- a/lib/importers/revision_score_importer.rb
+++ b/lib/importers/revision_score_importer.rb
@@ -39,10 +39,9 @@ class RevisionScoreImporter
 
   # assumes a mediawiki rev_id from the correct Wikipedia
   def fetch_ores_data_for_revision_id(rev_id)
-    articlequality_data = @lift_wing_api.get_revision_data([rev_id])
-    features = articlequality_data.dig(wiki_key, 'scores', rev_id.to_s, model_key, 'features')
-    rating = articlequality_data.dig(wiki_key, 'scores', rev_id.to_s, model_key,
-                                     'score', 'prediction')
+    result = @lift_wing_api.get_single_revision_parsed_data(rev_id)
+    features = result.dig('features')
+    rating = result.dig('prediction')
     return { features:, rating: }
   end
 
@@ -69,39 +68,31 @@ class RevisionScoreImporter
   ##################
   private
 
-  # The top-level key representing the wiki in LiftWing data
-  def wiki_key
-    # This assumes the project is Wikipedia, which is true for all wikis with the articlequality
-    # or the language is nil, which is the case for Wikidata.
-    @wiki_key ||= "#{@wiki.language || @wiki.project}wiki"
-  end
-
   def model_key
     @model_key ||= @wiki.project == 'wikidata' ? 'itemquality' : 'articlequality'
   end
 
   def get_and_save_scores(rev_batch)
-    scores_data = @lift_wing_api.get_revision_data rev_batch.map(&:mw_rev_id)
-    scores = scores_data.dig(wiki_key, 'scores') || {}
-    save_scores(scores)
+    rev_batch.each do |rev|
+      result = @lift_wing_api.get_single_revision_parsed_data(rev.mw_rev_id)
+      save_score(rev.mw_rev_id, result)
+    end
   end
 
   def get_and_save_previous_scores(rev_batch)
     parent_revisions = get_parent_revisions(rev_batch)
     return unless parent_revisions&.any?
-    parent_quality_data = @lift_wing_api.get_revision_data parent_revisions.values.map(&:to_i)
-    scores = parent_quality_data.dig(wiki_key, 'scores') || {}
-    save_parent_scores(parent_revisions, scores)
+    parent_revisions.each do |rev_id, parent_id|
+      parent_score = @lift_wing_api.get_single_revision_parsed_data(parent_id)
+      save_parent_score(rev_id, parent_score)
+    end
   end
 
-  def save_parent_scores(parent_revisions, scores)
-    parent_revisions.each do |mw_rev_id, parent_id|
-      next unless scores.key? parent_id
-      article_completeness = weighted_mean_score(scores[parent_id])
-      features_previous = scores[parent_id]&.dig(model_key, 'features')
-      Revision.find_by(mw_rev_id: mw_rev_id.to_i, wiki: @wiki)
-              .update(wp10_previous: article_completeness, features_previous:)
-    end
+  def save_parent_score(rev_id, parent_score)
+    return unless parent_score
+    Revision.find_by(mw_rev_id: rev_id.to_i, wiki: @wiki)
+            .update(wp10_previous: parent_score.dig('wp10'),
+                    features_previous: parent_score.dig('features'))
   end
 
   def mainspace_userspace_and_draft_revisions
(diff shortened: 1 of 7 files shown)


spec/services/update_course_stats_spec.rb
```diff
@@ -106,14 +106,14 @@
         subject
       end
       sentry_tag_uuid = subject.sentry_tag_uuid
-      expect(course.flags['update_logs'][1]['error_count']).to eq 8
+      expect(course.flags['update_logs'][1]['error_count']).to eq 328
```
gabina, line 109: Before this PR, errors were logged at the batch level, resulting in a single Sentry log per batch, not per revision ID. For the specified course, the `update_revision_scores_for_course` method updates a total of 328 revisions (167 revisions + 161 parent revisions). Since all LiftWing requests are intentionally set to fail, there are 328 different Sentry logs.

Prior to these changes, there were only 8 logs because logs were at the batch level, and there were 8 batches: (161/50 + 1 = 4 batches) + (167/50 + 1 = 4 batches) = 8 batches.

I think that logging at revision level instead of batch level makes sense, because we're making one request to the liftwing API per revision id, so requests are completely individual. If we think that this could mean an overwhelming number of Sentry logs, I can make changes on the way errors are logged.

ragesoss: It's a significant risk to go from batches to individual revisions for error logging. If there is a systematic problem where every request is failing, this could quickly saturate our Sentry event quota and result in further monitoring becoming unavailable.

It's good to include the counts in the Course update logs, as long as we're not actually sending that many separate Sentry events as well (if the same thing happened in production).

gabina: Gotcha I rolled back some changes in order to keep logging errors at the batch level (50 revisions). Done mainly on c490de07f3fde4022a85e9e065b1ed629f2ee7c1. I think I should do something similar for the new `ReferenceCounterApi` class.

> It's good to include the counts in the Course update logs

That can be done, but we should implement some modifications. As I rolled back the changes, we're once again invoking `ApiErrorHandling.log_error` per batch, which, in turn, calls `UpdateServiceErrorHelper.update_error_stats` and increments `@error_count` by 1 each time. Despite passing the actual number of errors in the batch to `ApiErrorHandling.log_error`, we're not consistently utilizing that number to update `@error_count`, and it looks like that's the number we use for the 'error_count' course flag. I don't think that's in the scope of this PR, I could do it in a different one if we think it's important.

ragesoss: It's not very important, but if you have a clear picture in your mind of what the problem is, please open an issue to document it.

gabina: Created issue #5580

lib/importers/revision_score_importer.rb
```diff
@@ -69,38 +58,37 @@ def update_previous_revision_scores
...
   end
 
+  def save_scores(scores)
+    scores.each do |mw_rev_id, score|
+      revision = Revision.find_by(mw_rev_id: mw_rev_id.to_i, wiki_id: @wiki.id)
+      next unless revision
+      revision.wp10 = score.dig('wp10')
+      revision.features = score.dig('features')
+      # only modify the existing deleted value if revision was deleted
+      revision.deleted = true if score.dig('deleted')
```
gabina, line 80: The spec modified in c04587283aa60c5f66a2b495b737bc704b8426de was to be sure that `revision.deleted` is not updated if `score.dig('deleted')` isn't set to true. If `score.dig('deleted')` is nil (because `score` is the empty hash {}), then `revision.deleted` is not updated.

lib/lift_wing_api.rb
```diff
@@ -22,31 +22,51 @@ def self.valid_wiki?(wiki)
...
 
+  # Given an array of revision ids, it returns an array with useful metrics for those
+  # revision ids.
+  # Format result example:
+  # { 'rev_id0' => { 'wp10' => 0.2915228958136511656e2, 'features' => features_value,
+  #                 'deleted' => false, 'prediction' => 'Stub' }
+  #   ...
+  #   'rev_idn' => { 'wp10' => 0.285936675221734978e2, 'features' => features_value,
+  #                 'deleted' => false, 'prediction' => 'D' }
+  # }
```
gabina, line 38: I added this as I don't think there is any other way to specify types in ruby

lib/lift_wing_api.rb
```diff
@@ -2,9 +2,9 @@
 
 require_dependency "#{Rails.root}/lib/errors/api_error_handling"
 
-# Gets data from Lift Wing
+# Gets and processes data from Lift Wing
 # https://wikitech.wikimedia.org/wiki/Machine_Learning/LiftWing
-class LiftWingApi
+class LiftWingApi # rubocop:disable Metrics/ClassLength
```
gabina, line 7: I had to disable the rule because the class is too long.
One option could be to move the `WEIGHTING` constants outside the class? Other thing that would take more effort would be to split the `LfitWingApi` class in two different classes: one to get data form the api, and other one to process it (calculating the `wp10` for example).

lib/lift_wing_api.rb
```diff
@@ -22,31 +22,51 @@ def self.valid_wiki?(wiki)
...
 
     log_error_batch(rev_ids)
 
     return results
   end
 
-  def get_single_revision_data(rev_id)
+  private
+
+  # Returns a has with wp10, features, deleted, and prediction, or empty hash if
```
gabina, line 52: typo here: Returns a hash

lib/lift_wing_api.rb
```diff
@@ -22,31 +22,51 @@ def self.valid_wiki?(wiki)
...
+  # there is an error.
+  def get_single_revision_parsed_data(rev_id)
     body = { rev_id:, extended_output: true }.to_json
     response = lift_wing_server.post(quality_query_url, body)
     parsed_response = Oj.load(response.body)
 
-    return equivalent_ores_error_response(rev_id, parsed_response) if parsed_response.key? 'error'
+    # If the responses contains an error, do not try to calculate wp10 or features.
```
gabina, line 59: typo here: If the responses contain

spec/lib/importers/revision_score_importer_spec.rb
```diff
@@ -57,22 +57,6 @@
     end
   end
 
-  it 'saves wp10 scores by article' do
-    VCR.use_cassette 'revision_scores/by_article' do
-      described_class.update_revision_scores_for_all_wikis
```
gabina, line 62: Removed all the specs calling `update_revision_scores_for_all_wikis` because removed the method.

lib/lift_wing_api.rb
```diff
@@ -72,36 +101,103 @@ def lift_wing_server
...
-        } } } } }
-  end
-
   # TODO: monitor production for errors, understand them, put benign ones here
   TYPICAL_ERRORS = [].freeze
 
   def log_error_batch(rev_ids)
     return if @errors.empty?
 
     log_error(@errors.first, update_service: @update_service,
```
gabina, line 110: I understand that we always log the first error, even when there might be multiple distinct errors. I guess we work under the assumption that if several requests fail, it's likely for the same reason, which may hold true in most cases. However, I suggest resetting the `@errors` array in `get_revision_data`. Without this adjustment, if there are two batches of requests, each with a single request failing, we would log the first error twice and miss the second one entirely.

gabina: Done on cde5e1d182f2b302b043cd8a6b538d240fc138da

lib/lift_wing_api.rb
ragesoss: The better place to do this is `.rubocop.yml` so that we have a central list of all the too-big classes.

ragesoss:
Overall this looks solid. I left a couple of suggestions; other than that, it looks ready to me. @Aminehassou would you like to have a look as well?

spec/lib/lift_wing_api_spec.rb
```diff
@@ -70,15 +75,24 @@
     it 'handles timeout errors' do
       stub_request(:any, 'https://api.wikimedia.org')
         .to_raise(Errno::ETIMEDOUT)
-      expect_any_instance_of(described_class).to receive(:log_error).once
+      expect(lift_wing_api_class_en_wiki).to receive(:log_error).once
```
gabina, line 78: Changed this to avoid using `expect_any_instance_of` here, as the spec then makes statements on `subject0`, which is a particular instance. I don't like making expectations to any instance and to a particular instance in the same spec, I think it's confusing.

lib/lift_wing_api.rb
```diff
@@ -72,36 +103,103 @@ def lift_wing_server
...
+    'sv' => SVWIKI_WEIGHTING,
+    'nl' => NLWIKI_WEIGHTING,
+    'pt' => PTWIKI_WEIGHTING
+  }.freeze
+
+  def weighting
+    @weighting ||= WEIGHTING_BY_LANGUAGE[@wiki.language]
+  end
+
+  def weighted_mean_score(score)
```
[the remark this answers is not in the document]
gabina: I like the idea! Added the new module on d0dfb96349ee564529ea94b2331a08fea71cd649, and modified the `LiftWingApi` class to include the module on 18c7a404fe4c14577c04b0d56edefeb448362703.
Didn't add specific specs for the new module because it's only used in the `LiftWingApi`, and the `LiftWingApi` 
 specs already check that the weighted mean score makes sense.

Issue #279, opened 2015-08-08
Milestones section should show dates in addition to week numbers.


ragesoss:
On the overview page, the dates of the highlighted Timeline details are important information.

ragesoss:
This should be pretty easy after the refactor of how we handle timeline weeks to use shared redux state for Timeline and Milestones.

HARSHITHASALADI1:
Could you please assign me this task?

fillingtothemomo:
Hey @ragesoss  I would like to work on this issue, can you elaborate on how to reproduce this?

ragesoss:
@fillingtothemomo To reproduce this, put your dev environment in `wiki_education: true` mode, then create a course and go through the assignment wizard. At the end, the course will have content on the Timeline tab, as well as a Milestones section at the lower right of the Home tab. (For example: https://dashboard.wikiedu.org/courses/University_of_Pennsylvania/Medical_missionaries_to_Community_Partners_(Fall_2023)/home )

The Milestones section of the Home tab is based on the same data as the Timeline (it represents the Milestone type blocks from the Timeline), but it does not include the dates associated with those blocks.

(Please only work on one issue at a time.)

MDCODE247:
Hello @ragesoss I'm picking this up and working on it

shishiro26:
Hiii @ragesoss, what's the status of this ? I would like to raise a contribution..

shishiro26:
Hiii @ragesoss, what's the status of this ? I would like to raise a contribution.

ragesoss:
@shishiro26 this is still open, and you're welcome to give it a shot.

GonzaFP:
@ragesoss is there anymore we can add to this? I see the dates have been added.

ragesoss:
> [@ragesoss](https://github.com/ragesoss) is there anymore we can add to this? I see the dates have been added.

Good catch. This is done.

PR #5581, opened 2024-01-11
renders dates for milestones in home tab

fillingtothemomo:
## What this PR does
It addresses the issue #279  Milestones section should show dates in addition to week numbers. 
I declared two functions getWeekStartDate and getWeekEndDate in my `app/assets/javascripts/utils/course_date_utils.js` that gets the start and end date of each week and they are used in `app/assets/javascripts/components/overview/milestones.jsx` to render the dates of each milestone.

## Screenshots
Before:
image

After:
image

### DIFF
--- a/app/assets/javascripts/components/overview/milestones.jsx
+++ b/app/assets/javascripts/components/overview/milestones.jsx
@@ -4,20 +4,19 @@ import createReactClass from 'create-react-class';
 import { filter } from 'lodash-es';
 
 import CourseDateUtils from '../../utils/course_date_utils';
-import DateCalculator from '../../utils/date_calculator.js';
+import DateCalculator from '../../utils/date_calculator'; // Import DateCalculator
 
 const md = require('../../utils/markdown_it.js').default();
-
 
 const Milestones = createReactClass({
   displayName: I18n.t('blocks.milestones.title'),
 
   propTypes: {
     timelineStart: PropTypes.string.isRequired,
-    timelineEnd: PropTypes.string,
+    timelineEnd: PropTypes.string.isRequired, // Add timelineEnd prop
     weeks: PropTypes.array.isRequired,
     allWeeks: PropTypes.array.isRequired,
-    course: PropTypes.object.isRequired,
+    course: PropTypes.object.isRequired
   },
 
   milestoneBlockType: 2,
@@ -27,8 +26,6 @@ const Milestones = createReactClass({
   },
 
   render() {
-    const dateCalc = new DateCalculator(
-    this.props.timelineStart, this.props.timelineEnd, this.props.index, { zeroIndexed: false });
     const currentWeek = CourseDateUtils.currentWeekOrder(this.props.timelineStart);
     const weekNumberOffset = CourseDateUtils.weeksBeforeTimeline(this.props.course);
     const blocks = [];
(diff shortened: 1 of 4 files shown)


fillingtothemomo:
@ragesoss I have made the PR for this issue please review it and suggest changes, if any.

ragesoss:
This is likely to include bugs related to how weeks in the Timeline get turned into dates. A good solution will involve using the same code to calculate the Timeline dates as to calculate the dates for the Milestone component.

In addition to using the same code to calculate dates for both places, it will be useful to test this out manually (and include in the screenshots or video for the PR) by looking at a complex timeline that include empty weeks (weeks with holidays marked for each day where the course normally meets) so that you can see the behavior of when the week numbers don't match up to the actual calendar weeks spanned.

fillingtothemomo:
> This is likely to include bugs related to how weeks in the Timeline get turned into dates. A good solution will involve using the same code to calculate the Timeline dates as to calculate the dates for the Milestone component.
> 
> In addition to using the same code to calculate dates for both places, it will be useful to test this out manually (and include in the screenshots or video for the PR) by looking at a complex timeline that include empty weeks (weeks with holidays marked for each day where the course normally meets) so that you can see the behavior of when the week numbers don't match up to the actual calendar weeks spanned.

image
WIKI1
@ragesoss I created a course in which I put the week1 as holiday and this is the produced behavior, do I have to make any changes to it? I also refactored the code.
image

ragesoss:
Hmm... I would check whether blackout dates that are not the first week behave differently, as that's where I would expect bugs to be the most likely.

However, I think this still exacerbates the problem of complex date code being used independently in multiple places. Rather than having multiple implementations of the code necessary to calculate the dates, I think the right way to fix this will be to either move the calculations higher up in the render chain so that the same results can be passed to both the Timeline and Milestones components, or extract the necessary code to a single function that accepts the same arguments in each place. (I realize that's a substantial refactor that would require a lot more work than the implementation you've done, but I think that's what is necessary to fix this one properly.)

fillingtothemomo:
@ragesoss the checks fails for master too so what changes should I make to my code?

ragesoss:
@fillingtothemomo no, if there a test this is failing on your PR that also fails on master, that is probably unrelated. I'm going to try to fix the build on master this week.

ragesoss:
@fillingtothemomo per my previous January 18 comment, I'm not satisfied with the way this uses separate code to calculate the dates just for the Milestones component. I think it needs to be done in such a way that the same function and same data gets used for dates on both the Milestones component and for the dates on the relevant parts of the Timeline tab.

fillingtothemomo:
> @fillingtothemomo per my previous January 18 comment, I'm not satisfied with the way this uses separate code to calculate the dates just for the Milestones component. I think it needs to be done in such a way that the same function and same data gets used for dates on both the Milestones component and for the dates on the relevant parts of the Timeline tab.

I did change the logic of fetching milestone dates in the second commit , might have missed pinging for review , please review the code and suggest changes.

app/assets/javascripts/utils/course_date_utils.js
```diff
@@ -127,6 +127,14 @@ const CourseDateUtils = {
     const timelineStart = startOfWeek(toDate(course.timeline_start));
     return differenceInWeeks(timelineStart, courseStart);
   },
+  getWeekStartDate(week, timelineStart) {
```
ragesoss, line 130: It looks like these functions are not used?

fillingtothemomo: Yes, these are related to the initial commit and this redundant. I will remove these.

ragesoss: Thanks. The issue from my January 18 comment remains: this implements new code to calculate the dates within the Milestone component, rather than refactoring how and where the dates are calculated so that the same calculation can be used for both the Milestones and the Timeline.

fillingtothemomo: @ragesoss I used the calculating timeline dates code logic in week.jsx 
image
and implemented dateCalc in milestones.jsx, should I just implement a dateCalc function in dateCalculator.jsx and use that to calculate dates for both?

fillingtothemomo: @ragesoss A similar logic is being used in week.jsx as well as empty_week.jsx so a function being implemented as a util would make more sense to be in sync for all files. What do you think?

fillingtothemomo:
@ragesoss  Please review the new changes.

app/assets/javascripts/components/overview/milestones.jsx
```diff
@@ -31,6 +33,12 @@ const Milestones = createReactClass({
     this.props.allWeeks.map((week) => {
       if (week.empty) return null;
 
+      const datesStr = DateCalculator.calculateDates(
+        this.props.timelineStart, this.props.timelineEnd, this.props.index, { zeroIndexed: false }
```
ragesoss, line 37: Why is this one zeroIndexed `false` while the other usage is `true`?

Issue #5525, opened 2023-11-01
Simplify Program deletion process


Chioma-data-analytics:
## What is happening.
The process for deleting a program involves multiple steps and navigation between pages. This could be confusing for users. 

### To Reproduce
1.	On My dashboard, Find and click the campaign name that contains the program to delete.
2.	From the list of programs, click on the program name 
3.	Click "remove" button
4.	“Ok” a confirmation dialog to confirm deletion
5.	Return to My dashboard
6.	Click on the program's name
7.	Scroll  to Action Section and Click "delete program"
8.	Type the name of the program in the prompt to confirm deletion, Click “Ok”  

### Expected behavior
I'm suggesting implementing a "Delete" or "Remove" button directly on the program within the user's dashboard "My dashboard" or end the process in the campaign page with just few clicks as shown below.
1.     On my dashboard, Find and click the campaign name that contains the program to delete.
2.	From the list of programs, click on the program name 
3.	Click "remove" button
4.	“Ok” a confirmation dialog to confirm deletion

ragesoss:
I think you're right, 3 buttons would be needed, and you are correct about the 2-campaigns situation. I think the labels also need to be more explicit. In particular 'remove' should probably become 'Remove from Campaign'.

Obviously, only admins should see the delete button.

PR #5583, opened 2024-01-13
#5525 - Simplify Program Deletion Process

ujjwalpathaak:
refs - #5525 

image
```/home/ujjwalpathaak/WikiEduDashboard/app/views/campaigns/_course_row.html.haml``` - this page has actions mentioned for clone and remove

```
 - if @presenter&.can_remove_course?
      %a.course-link{:href => "#{course_slug_path(course.slug)}"}
        = form_for(@campaign, url: remove_course_campaign_path(@campaign.slug, course_id: course.id), method: :put, id: "remove_course-#{course.id}", html: { class: 'remove-program-form' }) do
          = hidden_field_tag('course_title', course.title, id: "course_title-#{course.id}")
          %button.button.danger.remove-course{'data-id' => course.id, 'data-title' => course.title, 'data-campaign-title' => @campaign.title}
            = t('assignments.remove')
```
This is the code for remove button - if i can get a prop to check if **user is allowed to delete the course + if no of campaigns = 1**
just like it's done here - ```/home/ujjwalpathaak/WikiEduDashboard/app/assets/javascripts/components/overview/available_actions.jsx```

```
 delete() {
    // The action is only available once a course has been removed from all campaigns.
    if (this.props.course.published) {
      return alert(I18n.t('courses.delete_course_instructions'));
    }

    const enteredTitle = prompt(I18n.t('courses.confirm_course_deletion', { title: this.props.course.title }));
    if (enteredTitle.trim() === this.props.course.title.trim()) {
      return this.props.deleteCourse(this.props.course.slug);
    } else if (enteredTitle) {
      return alert(I18n.t('courses.confirm_course_deletion_failed', { title: enteredTitle }));
    }
  },

  if ((user.isAdvancedRole || user.admin) && (!course.published || !Features.wikiEd)) {
        controls.push((
          <div title={I18n.t('courses.delete_course_instructions')} key="delete" className="available-action">
            <button className="button danger" onClick={this.delete}>
              {CourseUtils.i18n('delete_course', course.string_prefix)}
            </button>
          </div>
        ));
      }
```

I can replicate the functinianlity and add the button in the haml file -- so the user can delete the programs directly from campaign's list of programs.

### CODE REVIEW

**Comment by ujjwalpathaak** [comment]
@ragesoss let me know the changes.

This is how the delete process is working rn
[Screencast from 13-01-24 08:57:37 PM IST.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/92005669/6b7c8c65-c3d6-4661-a9c6-4e38b81cd483)

- I think if only 1 campaign is there then delete button should be visible otherwise remove button only.
- This snippet needs to be updated with correct permissions. maybe?  - ```presenters/courses_presenter.rb```
```  
     def can_delete_course?
        @can_remove ||= current_user&.admin? || campaign_organizer?
     end
```
- I need to change the behaviour -- after the title is correctly entered and course is deleted it's going to the response .json file.

**Comment by ujjwalpathaak** [comment]
@ragesoss LMKYT regarding the commit

**Comment by ujjwalpathaak** [comment]
@ragesoss I think this PR is ready for review.

File: app/assets/javascripts/components/overview/available_actions.jsx [hunk 1]
```diff
@@ -72,7 +72,7 @@ const AvailableActions = createReactClass({
     }
 
     const enteredTitle = prompt(I18n.t('courses.confirm_course_deletion', { title: this.props.course.title }));
-    if (enteredTitle.trim() === this.props.course.title.trim()) {
+    if (enteredTitle !== null && enteredTitle.trim() === this.props.course.title.trim()) {
```
ragesoss, line 75: Why is this necessary? Are there cases where the trimmed title from the props could be null?

app/assets/javascripts/utils/course.js
```diff
@@ -91,6 +91,24 @@ document.onreadystatechange = () => {
     });
   }
 
+  const y = document.getElementsByClassName('delete-course-prerequisite')[0];
```
ragesoss, line 94: Please use descriptive variable names rather than generic ones like `y` and `z`.

app/presenters/courses_presenter.rb
```diff
@@ -59,6 +59,10 @@ def can_remove_course?
     @can_remove ||= current_user&.admin? || campaign_organizer?
   end
 
+  def can_delete_course?
+    @can_remove ||= current_user&.admin? || campaign_organizer?
```
ragesoss, line 63: Is this variable name a mistake? It looks like this is just a complete duplicate of `can_remove_course?`

ragesoss: I think a little more details will be required to determine whether a course can be deleted. For any given course, it may be in multiple campaigns, but normally only one of the facilitators/instructors for the program itself is allowed to delete a course.

I think it also makes sense to allow an campaign organizer or admin to delete a course that is in campaign... but only if it is not in any other campaigns. Otherwise, it should probably just remove the campaign and report that the course could not be deleted because it is part of another campaign.

config/locales/en.yml
```diff
@@ -293,6 +294,7 @@ en:
     confirm_add_available: Are you sure you want to add %{title}?
     confirm_addition: Are you sure you want to assign %{title} to %{username}?
     confirm_deletion: Are you sure you want to delete this assignment?
+    delete: Delete
```
ragesoss, line 297: Perhaps 'Remove and Delete' would be a better title.

app/controllers/courses_controller.rb
```diff
@@ -59,9 +59,21 @@ def update
 
   def destroy
     validate
-    DeleteCourseWorker.schedule_deletion(course: @course, current_user:)
-    render json: { success: true }
-  end
+    if params.key?(:campaign)
```
ragesoss, line 62: Can the flow control be simplified here? I don't like having two separate `if` blocks within the method.

app/controllers/courses_controller.rb
```diff
@@ -59,9 +59,21 @@ def update
...
+      flash[:notice] = t(message, title: @course.title,
+                                  campaign_title: params[:campaign_title])
+    end
+    DeleteCourseWorker.schedule_deletion(course: @course, current_user: current_user)
+    if params.key?(:campaign)
+      redirect_to programs_campaign_path(params[:slug])
+    else
+      render json: { success: true }
+    end
+  end
```
ragesoss, line 76: This controller is already too large, and rather than make it even more complex, I think this is a good opportunity to move the destroy method to a child controller following the pattern of the other ones in `app/controllers/courses/`.

ragesoss (changes requested):

app/assets/javascripts/components/overview/available_actions.jsx
ujjwalpathaak: it is not visible in prod but when i cancel the prompt while it is empty in dev it shows an error.

[Screencast from 14-02-24 12:28:01 PM IST.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/92005669/d5f34d71-d5f1-498c-a1c9-0f04c3670494)

app/assets/javascripts/utils/course.js
ujjwalpathaak: > WikiEduDashboard/app/assets/javascripts/utils/course.js   ---   Line: 80
```
  // for use on campaign/programs page
  const x = document.querySelectorAll('.remove-course');
  for (let i = 0; i < x.length; i += 1) {
    x[i]?.addEventListener('click', (e) => {
        const confirmed = window.confirm(I18n.t('campaign.confirm_course_removal', {
          title: e.target.dataset.title,
          campaign_title: e.target.dataset.campaignTitle
        }));
        if (!confirmed) {
          e.preventDefault();
        }
    });
  }
```

Since variable x was already in use for a similar function within the file, I went ahead and used variables y and z to maintain uniformity for new functions. I'll update the y and z variables for better clearity. should i do the same for x?

ujjwalpathaak: I have changed the names for all three variables - 
x -> removeCourseBtn
y -> deleteCoursePrerequisiteBtn
z ->  deleteCourseBtn

app/controllers/courses_controller.rb
ujjwalpathaak:
```
  def destroy
    validate
    if params.key?(:campaign)
      remove_course_from_campaign
    else
      DeleteCourseWorker.schedule_deletion(course: @course, current_user: current_user)
      render json: { success: true }
    end
  end
  
  def remove_course_from_campaign
    campaigns_course = CampaignsCourses.find_by(course_id: @course.id, campaign_id: params[:campaign_id])
    result = campaigns_course.destroy
    message = result ? 'campaign.course_removed' : 'campaign.course_already_removed'
    flash[:notice] = t(message, title: @course.title, campaign_title: params[:campaign_title])
    DeleteCourseWorker.schedule_deletion(course: @course, current_user: current_user)
    redirect_to programs_campaign_path(params[:slug])
  end
```

made a new function `remove_course_from_campaign` to **increase readability** and **simplified the workflow**

app/presenters/courses_presenter.rb
ujjwalpathaak: I have updated `@can_remove` to `@can_delete`.

Still it's unclear to me what permission changes are required. I'll give you a little rundown, and then can you please mention the changes once again?

### Workflow 
The Remove and Delete button is only visible to a user if 
```
  def can_delete_course?
    @can_delete ||= current_user&.admin? || campaign_organizer?
  end
```
is true.


When clicked on the Remove and Delete button. 
image

It checks if the program is only present in one campaign.

If yes. it removes and deletes the program.

If not? then it shows - 
image

### I set the same level of permission to - _ability to delete a program_ as the permissions given to -  _ability to remove a program from campaign_.

### But the **program can only be deleted** if it is not present in any other campaign.

ujjwalpathaak:
@ragesoss I have made a few changes, but i need help in one of the changes required in `app/presenters/courses_presenter.rb`. LMKWYT.

app/presenters/courses_presenter.rb
ragesoss: This flow makes sense to me.

app/assets/javascripts/components/overview/available_actions.jsx
ragesoss: Cool, that makes sense. Please add a comment noting what situations these guard checks are for.

app/presenters/courses_presenter.rb
ragesoss: But, if a user tries to 'Remove and Delete' and it is in another campaign, it should still remove this campaign and the notification should reflect that.

app/controllers/courses_controller.rb
ragesoss: This flow is better, but there is duplication of the `schedule_deletion` part and the remove_course_from_campaign method does more than just remove the campaign, it also queues the course for deletion.

I think both versions of the delete action should be moved to a new child controller.

ujjwalpathaak:
Hi @ragesoss, I have made the necessary changes.
- Made a child controller
- Simplified the workflow + a tiny change in the workflow
- Added a comment 

## 1) Course with only 1 campaign
### Program 24 is only in one campaign, so after clicking "Remove and Delete" it removes the program also delete it, to notify that action it provokes a notification. -- 
``` course_removed_and_deleted: '"%{title}" has been deleted and removed from the "%{campaign_title}" campaign' ```

[Screencast from 18-02-24 03:20:00 PM IST.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/92005669/f77cc5f0-23a6-4e4b-9798-c61ef5d90be9)

## 2) Course with 1+ campaigns
### Program 23 is in more than one campaign, so after clicking "Remove and Delete" it removes the program but dosent delete it, to notify that action it provokes a notification. --
``` course_removed_but_not_deleted: '"%{title}" has been removed from the "%{campaign_title}" but has not been deleted as it is present in other campaigns as well' ```

[Screencast from 18-02-24 03:19:37 PM IST.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/92005669/6f1e93dc-d32d-4abc-abfa-d25b3a4c19ce)

File: app/assets/stylesheets/_fonts.styl [hunk 7]
```diff
@@ -19,4 +19,7 @@
       color zircon
 
 .red
-  color warning
\ No newline at end of file
+  color warning
+
+.delete-course-prerequisite
+  margin 0px 24px 12px 24px !important;
```
ragesoss, line 25: Why is this in `_fonts`?

Can it be implemented without using `!important` (for example by using a more specific selector based on the classes of parent elements of the one you are styling)?

app/views/campaigns/_course_row.html.haml
```diff
@@ -57,5 +57,18 @@
       %a.course-link{:href => "#{course_slug_path(course.slug)}"}
         = form_for(@campaign, url: remove_course_campaign_path(@campaign.slug, course_id: course.id), method: :put, id: "remove_course-#{course.id}", html: { class: 'remove-program-form' }) do
           = hidden_field_tag('course_title', course.title, id: "course_title-#{course.id}")
-          %button.button.danger.remove-course{'data-id' => course.id, 'data-title' => course.title, 'data-campaign-title' => @campaign.title}
+          %button.button.danger.remove-course{ 'data-id' => course.id, 'data-title' => course.title, 'data-campaign-title' => @campaign.title, title: t('campaign.remove_course_tooltip') }
             = t('assignments.remove')
+    - if @presenter&.can_delete_course?
+      - if course.campaigns.size > 1
```
ragesoss, line 63: I would like to avoid needing to query every course for the number of campaigns, as this creates an "N+1" query with an additional database query for every course on the page. (If we can know it can't be deleted already at the time of rendering, then we shouldn't show the delete button.)

I think the right way to do this is to render the same button format for each course, and just detect whether or not it can be deleted during the request to the deletion controller.

app/presenters/courses_presenter.rb
```diff
@@ -59,6 +59,10 @@ def can_remove_course?
     @can_remove ||= current_user&.admin? || campaign_organizer?
   end
 
+  def can_delete_course?
```
ragesoss, line 62: As this will behave the same way as `can_remove?`, I suggest using `alias_method` along with a comment, rather than repeating the implementation.

ragesoss (changes requested):

ujjwalpathaak:
@ragesoss i have made the changes

PR #5592, opened 2024-01-23
Resolved fetching revisions from wikisource.org

### LINKED ISSUE #5530: Activity tab does cannot fetch revisions from wikisource.org
### What is happening?
The Activity tab of this course does not show any activity: https://outreachdashboard.wmflabs.org/courses/Igbo_Wikimedians_User_Group/Introduction_to_Igbo_Wikisource/activity/recent

There are JS errors in the console; it looks like CORS is stopping the requests to the Wikisource API.

### Additional context

wikisource.org is a separate wiki (for multilingual source texts) from the individual language versions of Wikisource like en.wikisource.org. In contrast to `www.wikidata.org`, the canonical URL for this wiki does not include the `www.` subdomain. However, the requests from the Activity tab are going to www.wikisource.org rather than just wikisource.org. It's like that the CORS errors are because of this.

om-chauhan1:
## What this PR does
Fixes #5530 
This PR resolves the CORS error that is occurring while fetching revisions from wikisource.org

## Screenshots
Before:

Before

After:

After

### DIFF
--- a/app/assets/javascripts/utils/wiki_utils.js
+++ b/app/assets/javascripts/utils/wiki_utils.js
@@ -1,8 +1,7 @@
 import ArticleUtils from './article_utils';
 
 const toWikiDomain = (wiki) => {
-  const subdomain = wiki.language || 'www';
-  return `${subdomain}.${wiki.project}.org`;
+  return `${wiki.project}.org`;
 };
 
 const formatOption = (wiki) => {
--- a/app/assets/javascripts/utils/wiki_utils.js
+++ b/app/assets/javascripts/utils/wiki_utils.js
@@ -1,12 +1,9 @@
 import ArticleUtils from './article_utils';
 
 const toWikiDomain = (wiki) => {
-  const subdomain = wiki.language;
-  let url = `${wiki.project}.org`;
-  if (subdomain) {
-    url = `${subdomain}.${wiki.project}.org`;
-  }
-  return `${url}`;
+  const language = (wiki.language) ? `${wiki.language}.` : 'www.';
+  const subdomain = (wiki.project === 'wikisource' && !wiki.language) ? '' : language;
+  return `${subdomain}${wiki.project}.org`;
 };
 
 const formatOption = (wiki) => {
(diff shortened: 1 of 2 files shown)


ragesoss:
Probably comes from a call to `Wiki#domain`, as defined in `wiki.rb`. If a change to the wiki object is making it to the backend, that's probably a mistake; I would expect this issue to be limited to data and code on the JS side of the app.

om-chauhan1:
So, anything else that needs to be done in this issue from my end ?

om-chauhan1:
Okay, all js tests are passing.

ragesoss:
I'm not sure. If it's throwing a new error on the back end, then it's probably an unintended side effect of the frontend change you made that is causing unexpected data to reach the backend.

This should probably also include a test that exercises the now-working fetching of wikisource.org.

Now that I think about it again, I'm not certain that a correct solution won't involve a backend change, for example, to ensure that the `domain` itself is passed along for each `Wiki` that the frontend deals with. It might be that we need to use a domain supplied by the backend for making queries, rather than using the toWikiDomain function. Either than, or the function will need to deal specifically with the case of wikidata.org.

om-chauhan1:
@ragesoss i resolved all those previous errors just my modifying the logic of toWikiDomain as you expected earlier that this issue is only related to js side of app . Now only one test is failing but this test is also failing before doing any changes in  code (like in master also).


After Doing Changes :

Screenshot from 2024-01-24 16-38-07


In Master Also :

image

app/assets/javascripts/utils/wiki_utils.js
```diff
@@ -1,8 +1,9 @@
 import ArticleUtils from './article_utils';
 
 const toWikiDomain = (wiki) => {
-  const subdomain = wiki.language || 'www';
-  return `${subdomain}.${wiki.project}.org`;
+  const language = (wiki.language) ? `${wiki.language}.` : 'www.';
+  const subdomain = (wiki.project === 'wikisource' && !wiki.language) ? '' : language;
```
ragesoss, line 5: Include a comment here to explain why we have to special-case wikisource.org.

ragesoss:
Great. Sounds like the failing test is unrelated; I'll look into that. I would also like this PR to include a new test that would have caught the bug before the toWikiDomain change.

om-chauhan1:
Sure, I have added a new js test which will take care of previous bug in "toWikiDomain" function.


## Screenshot


New test with old function
image

New test after modifying function
image

ragesoss:
Looks good, thanks!

PR #5595, opened 2024-01-25
fixes 500 link limit issue

### LINKED ISSUE #5588: Fix 500-link limit for loading training slides from wiki

fillingtothemomo:
fixes #5588

## What this PR does
It fixes the 500 page link limit for the training modules.
## Screenshots
Before:
image

After:
image

### DIFF
--- a/lib/training/wiki_training_loader.rb
+++ b/lib/training/wiki_training_loader.rb
@@ -43,12 +43,12 @@ class WikiTrainingLoader
   def add_trainings_to_collection(wiki_page)
     content = new_from_wiki_page(wiki_page)
     unless content&.valid?
-      Sentry.capture_message 'Invalid wiki training content',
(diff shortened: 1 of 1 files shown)


fillingtothemomo:
@ragesoss The CI build keeps failing for this pr well as as earlier pr i made, what is the issue in my code i am not able to understand

ragesoss:
I tried this locally and got:
```
NoMethodError: undefined method `dig' for nil:NilClass

        current_links = response.dig('pages', @wiki_base_page, 'links') || []
```

fillingtothemomo:
@ragesoss this PR is faulty, I forgot to close it as I implemented continue assuming that the api response has a continue field to fetch more results however on inspecting the api response, there is no continue field so I am stumped as to how to apply pagination for more than 500 links.

Issue #5569, opened 2024-01-01
On the P&E Dashboard "Articles" tab of a Program add  informative text to the "See more" button on hover 


Ibom99:
### Current Behavior:
Currently on the Dashboard before a user can see or navigate through large sets of articles available in a program using pagination, the "See more" button needs to be selected before the pagination reflects the following number of subset articles (500 more articles). After the subset of 500 articles is reflected in the pagination and the user still wants to navigate through more articles, they need to select the "See more" button again, before the pagination reflects the next set of articles and so on till all articles have loaded (the "See more" button disappears, once the full set of articles have loaded) and are reflected in the pagination.

### Desired Behavior:
On the "Articles" tab of a program the "See more" button on hover should have informative text, informing the user of the relationship between the button and the pagination beside it reflecting more loaded articles. 
POSSIBLE INFORMATIVE TEXT ="Click "See more" to load the next subset of 500 articles and continue the seamless navigation through the articles with the pagination on the left."

### Additional context:
![Screenshot (88)](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/104465078/4879f5bf-5549-4422-b499-1dc7fd936179)

In the image below the "See more" button has been clicked on. The pagination range on the left increases and the numbers under the "See more" button increase as well. This shows that the next subset of articles have loaded and the pagination on the left reflects the loaded subset of articles.

![Screenshot (89)](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/104465078/67d4f176-7440-4d22-9114-890a4d1234c8)

Ibom99:
Hi @ragesoss @Aminehassou what file in the codebase can this feature be implemented?

ragesoss:
@Ibom99 the 'Articles Edited' view is from the `ArticleList` component (`article_list.jsx`), and the pagination UI is the `PaginatedArticleControls` component (`PaginatedArticleControls.jsx`) within the list component.

MrBlackGhostt:
@Ibom99 i want to contribute on this can i work on it

ragesoss:
Sure. I think the text will need some changes, but if you open a PR that will make it easy to discuss the wording.

PR #5597, opened 2024-01-27
Added Informative Text to 'See More' Button on Hover in P&E Dashboard 'Articles' Tab of a Program

PR4NJ41:
## What this PR does
Resolves issue #5569 

## Screenshots
Before:
Screenshot 2024-01-27 at 3 46 45 PM

After:
Screenshot 2024-01-27 at 3 24 19 PM

### DIFF
--- a/app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
+++ b/app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
@@ -42,12 +42,17 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
       />) : null }
       {!limitReached
         && (
-          <button
-            style={{ width: 'max-content', height: 'max-content' }}
-            className="button ghost articles-see-more-btn" onClick={showMore}
-          >
-            {I18n.t('articles.see_more')}
-          </button>
+          <div className={'tooltip-trigger'}>
+            <button
+              style={{ width: 'max-content', height: 'max-content' }}
+              className="button ghost articles-see-more-btn" onClick={showMore}
+            >
+              {I18n.t('articles.see_more')}
+            </button>
+            <div className="tooltip tooltip-center tooltip-see-more dark large">
+              <p style={{ padding: '0px', textWrap: 'balance' }}>Click &apos;See more&apos; to load the next subset of 500 articles and continue the seamless navigation through the articles with the pagination on the left.</p>
+            </div>
+          </div>
         )
       }
       <p className="articles-shown-label">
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
```diff
@@ -42,12 +42,17 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
...
-          </button>
+          <div className={'tooltip-trigger'}>
+            <button
+              style={{ width: 'max-content', height: 'max-content' }}
+              className="button ghost articles-see-more-btn" onClick={showMore}
+            >
+              {I18n.t('articles.see_more')}
+            </button>
+            <div className="tooltip tooltip-center tooltip-see-more dark large">
+              <p style={{ padding: '0px', textWrap: 'balance' }}>Click &apos;See more&apos; to load the next subset of 500 articles and continue the seamless navigation through the articles with the pagination on the left.</p>
```
ragesoss, line 53: This should be implemented in the I18n system so that it can be translated.

ragesoss: @PR4NJ41 you should add it in `en.yml` and ideally in `qqq.yml` as well. The latter is the documentation file, intended to describe the purpose and usage of the message, for translators, so they have enough context to translate it well. All the other locale files get handled via the volunteer translator community at translatewiki.net.

config/locales/en.yml
```diff
@@ -206,6 +206,7 @@ en:
       ?: Unrated — an article that hasn't yet been rated on Wikipedia's quality scale
       does_not_exist: This article does not exist on Wikipedia
     see_more: See more
+    see_more_tooltip: Click 'See more' to load the next subset of 500 articles and continue the seamless navigation through the articles with the pagination on the left.
```
ragesoss, line 209: "Only the first 500 edited articles are loaded initially. Click to load additional articles."

config/locales/qqq.yml
```diff
@@ -186,6 +186,7 @@ qqq:
       does_not_exist: Explanation of the Ø  symbol, which indicates than the article
         does not exist on Wikipedia
     see_more: Message indicating that there are more articles to load
+    see_more_tooltip: Message displayed when clicking see more button to load next subsets of 500 articles
```
ragesoss, line 189: Tooltip message shown when hovering on "See More" button for loading additional articles

ragesoss (commented):
Update the messages and this should be good to go.

PR #5599, opened 2024-01-27
Solved bug of next estimated update

### LINKED ISSUE #5489: Course update info is misleading for non-current courses
### What is happening?
On course pages, a small message below the header stats shows info about the update process: when the last update was completed, and (if it's in the future) when the next update is expected.

For a course that ended and stopped receiving updates, but then was updated via a manual update or via the 'Schedule Data Update' button, after the update is complete, the update info says something like "Next estimated update: in 6 days." This happens because it takes the average time between updates for the last 10 updates to estimate the time of next update, but for a course that has ended and is no longer in the update window (by default, 30 days after the end of the course, based on a setting in application.yml), this is misleading because no further updates are expected.

om-chauhan1:
## What this PR does
Fixes #5489 
This PR solves the a bug of next estimated update which is misleading previously.
 I have created the same scenario as mentioned in the issue(as shown in before screenshot) and then added the logic of checking whether the calculated estimated date is within the range of 30 days (after course end date) or not.

## Screenshots
Before:

Before

After:

After

### DIFF
--- a/app/assets/javascripts/utils/statistic_update_info_utils.js
+++ b/app/assets/javascripts/utils/statistic_update_info_utils.js
@@ -14,6 +14,10 @@ const lastSuccessfulUpdateMoment = (update_logs) => {
   return new Date(lastSuccessfulUpdateTime);
 };
 
+const isNextUpdateAfterUpdatesEnd = (nextUpdateExpectedTime, updatesEndMoment) => {
+  return isAfter(nextUpdateExpectedTime, new Date()) && isAfter(updatesEndMoment, new Date());
+};
+
 const getLastUpdateMessage = (course) => {
   let lastUpdateMessage = '';
   let nextUpdateMessage = '';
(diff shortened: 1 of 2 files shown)


om-chauhan1:
@ragesoss Sir, this test is also failing prior to the changes done in this pr i.e in master branch also.

app/assets/javascripts/utils/statistic_update_info_utils.js
```diff
@@ -19,11 +19,12 @@ const getLastUpdateMessage = (course) => {
   let nextUpdateMessage = '';
   let isNextUpdateAfter = false;
   const lastUpdateMoment = lastSuccessfulUpdateMoment(course.flags.update_logs);
+  const updatesEndMoment = toDate(course.update_until);
   if (lastUpdateMoment) {
     const averageDelay = course.updates.average_delay ?? 0;
     lastUpdateMessage = `${I18n.t('metrics.last_update')}: ${formatDistanceToNow(lastUpdateMoment, { addSuffix: true })}.`;
     const nextUpdateExpectedTime = addSeconds(lastUpdateMoment, averageDelay);
-    isNextUpdateAfter = isAfter(nextUpdateExpectedTime, new Date());
+    isNextUpdateAfter = (isAfter(nextUpdateExpectedTime, new Date()) && isAfter(updatesEndMoment, new Date()));
```
ragesoss, line 27: This works as expected, but it's not going to be obvious to future coders what the purpose of this line is, and the logic is duplicated in the other method.

Can the check against the end-of-updates date be extracted to another method so there's less duplication and there's a method name that makes the purpose more clear?

ragesoss (changes requested):

app/assets/javascripts/utils/statistic_update_info_utils.js
om-chauhan1: sure, i have extracted it to another method.

PR #5628, opened 2024-02-06
Made sandbox URL for an Assignment updateable.

### LINKED ISSUE #4157: The sandbox URL for an Assignment should be updateable

om-chauhan1:
## What this PR does
Fixes #4157 
This pr adds up the functionality of updating existing sandbox url to new url. 

## Screenshots
Before:

Before

After:

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/c4082338-ebc0-471d-82b0-22305e2dcda5)

### DIFF
--- a/app/assets/javascripts/components/common/AssignmentLinks/EditSandboxUrl.jsx
+++ b/app/assets/javascripts/components/common/AssignmentLinks/EditSandboxUrl.jsx
@@ -12,7 +12,7 @@ constructor(props) {
   this.handleNewUrlChange = this.handleNewUrlChange.bind(this);
   this.submit = this.submit.bind(this);
   this.state = {
-    newUrl: '',
+    newUrl: props.assignment.sandbox_url,
   };
 }
 
@@ -35,7 +35,7 @@ toggle(e) {
   e.preventDefault();
   this.props.open();
   this.setState({
-    newUrl: ''
+    newUrl: this.props.assignment.sandbox_url
   });
 }
 
@@ -64,10 +64,11 @@ render() {
   );
   if (is_open) {
     return (
-      <div className="pop__container edit_sandBox" onClick={this.stop}>
+      <div className="pop__container" onClick={this.stop}>
         <Popover
           is_open={is_open}
           edit_row={editRow}
+          right={true}
         />
       </div>
     );
--- a/app/assets/javascripts/components/common/AssignmentLinks/EditSandboxUrl.jsx
+++ b/app/assets/javascripts/components/common/AssignmentLinks/EditSandboxUrl.jsx
@@ -1,0 +1,89 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import Popover from '../../common/popover';
+import EditSandboxUrlInput from '../../assignments/edit_sandbox_url_input';
+import PopoverExpandable from '../../high_order/popover_expandable';
+import { updateSandboxUrl } from '../../../actions/assignment_actions';
+
+export class EditSandboxUrl extends React.Component {
+constructor(props) {
+  super(props);
+  this.toggle = this.toggle.bind(this);
+  this.handleNewUrlChange = this.handleNewUrlChange.bind(this);
+  this.submit = this.submit.bind(this);
+  this.state = {
+    newUrl: '',
+  };
+}
+
+getKey() {
+  let tag = 'open_edit_sandbox_url_input_';
+  return tag += this.props.assignment.id;
+}
+
+handleNewUrlChange(e) {
+  this.setState({
+    newUrl: e.target.value,
+  });
+}
+
+stop(e) {
+  e.stopPropagation();
+}
+
+toggle(e) {
+  e.preventDefault();
+  this.props.open();
+  this.setState({
+    newUrl: ''
+  });
+}
+
+submit(e) {
+  if (this.state.newUrl !== '') {
+    const { assignment } = this.props;
+    this.toggle(e);
+    this.props.updateSandboxUrl(assignment, this.state.newUrl);
+  } else {
+    e.preventDefault();
+  }
+}
+
+render() {
+  const { is_open } = this.props;
+  const editRow = (
+    <tr className="edit">
+      <td>
+        <EditSandboxUrlInput
+          submit={this.submit}
+          onChange={this.handleNewUrlChange}
+          value={this.state.newUrl}
+        />
+      </td>
+    </tr>
+  );
+  if (is_open) {
+    return (
+      <div className="pop__container edit_sandBox" onClick={this.stop}>
+        <Popover
+          is_open={is_open}
+          edit_row={editRow}
+        />
+      </div>
+    );
+  }
+    return (
+      <>
+        <a href="" target="_blank" onClick={this.toggle} className="change-sandbox-url">
+          ({I18n.t('assignments.change_sandbox_url')})
+        </a>
+      </>
+    );
+  }
+}
+
+const mapDispatchToProps = {
+  updateSandboxUrl,
+};
+
+export default connect(null, mapDispatchToProps)(PopoverExpandable(EditSandboxUrl));
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+# rubocop:disable Metrics/ClassLength
 
 require 'uri'
 require_dependency "#{Rails.root}/lib/assignment_manager"
@@ -91,16 +92,20 @@ class AssignmentsController < ApplicationController
 
   # This method validates the existance of the new username
   # which will be used to update the sandbox url of an existing assignment.
+  # rubocop:disable Style/HashSyntax
   def validate_new_sandbox_url_username(username = nil, assignment = nil)
     user = User.find_by(username: username)
-    if !user
-      render json: { errors: assignment.errors, message: t('assignments.no_user_exists', username: username) },
-             status: :not_found
-    else
+    if user
       assignment.update_sandbox_url(username)
+
       render partial: 'updated_assignment', locals: { assignment: assignment }
+    else
+      render json: { errors: assignment.errors,
+             message: t('assignments.no_user_exists', username: username) },
+             status: :not_found
     end
   end
+  # rubocop:enable Style/HashSyntax
 
   def update_onwiki_course_and_assignments
     UpdateAssignmentsWorker.schedule_edits(course: @course, editing_user: current_user)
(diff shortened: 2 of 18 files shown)


om-chauhan1:
Out of the 3 failing tests :

Two tests are passing locally :

image


And the third one is failing in master also

image

ragesoss:
This looks very cool! I will test it out and provide feedback as soon as I have a chance, hopefully by the end of this week.

app/models/course_data/assignment.rb
```diff
@@ -105,6 +108,19 @@ def peer_review_pagename
     "#{sandbox_pagename}/#{user.username}_Peer_Review"
   end
 
+  def update_sandbox_url(new_url)
+    return unless new_url
+    return unless user
+    # creates and initializes instance variable (edit_sandbox_url)
+    # used to bypass the before_save (set_sandbox_url) method.
+    @edit_sandbox_url = true
```
ragesoss, line 116: This should not be necessary; as long as the `sandbox_url` is already present, `set_sandbox_url` will not change it because of the `return if sandbox_url.present?` line.

app/models/course_data/assignment.rb
```diff
@@ -105,6 +108,19 @@ def peer_review_pagename
...
+    language = wiki.language || 'www'
+    project = wiki.project || 'wikipedia'
+    base_url = "https://#{language}.#{project}.org/wiki"
+    new_sandbox_url = "#{base_url}/User:#{new_url}/#{article_title}"
```
ragesoss, line 120: I don't think we want to assume that `new_url` is a username. I think the expected behavior would be to take a complete URL, perhaps validate it to make sure it's from the expected wiki, and then save it. It might or might not include the same article_title as the Assignment record.

app/controllers/assignments_controller.rb
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+# rubocop:disable Metrics/ClassLength
```
ragesoss, line 2: The preferred way to do this is via `.rubocop.yml`. Even better would be to extract the business logic to a separate file, keep the controller action as simple as possible, and avoid passing the limit. I'm not sure how close to the limit it is already, though.

app/controllers/assignments_controller.rb
```diff
@@ -77,8 +78,33 @@ def update_status
...
+    @new_url_username = params[:new_url]
+    @assignment = Assignment.find(params[:id])
+    validate_new_sandbox_url_username(@new_url_username, @assignment)
+  end
+
   private
 
+  # This method validates the existance of the new username
+  # which will be used to update the sandbox url of an existing assignment.
+  # rubocop:disable Style/HashSyntax
```
ragesoss, line 93: I'd rather not disable this rule, for codebase consistency.

ragesoss (changes requested):
I haven't tested the frontend behavior yet, but this looks reasonable overall.

om-chauhan1:
@ragesoss Kindly review whenever you are free.

ragesoss:
Yeah, if you still can't get it under the limit, add it as an exception in .rubocop.yml

om-chauhan1:
@ragesoss 
Now the flow is i am taking complete url as input then checking conditions accordingly like : 
1. Whether current_user is permitted to do so or not.
2. Whether the provided url is valid or not.
3. Whether the provided url belong to same wiki or not.
If every thing is fine then current_user is allowed to update its url.
According to this implementation user can easily customize its new url **including article title.**
In this updated code i have also taken care of both "Not disabling any rubocop rule" as well as translations.

Video :

[Updated.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/8638c6b8-2d6a-45ee-961f-05e4a3254b4a)

app/models/course_data/assignment.rb
om-chauhan1: sure

om-chauhan1:
@ragesoss  Ready for review, whenever you are free kindly have a look

ragesoss:
It's on my todo list, but it will probably not be until next week.

om-chauhan1:
sure, no problem

app/services/sandbox_url_updator.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
```
ragesoss, line 1: This needs some tests, with different example URLs that should vs should not validate.

I tested it out locally, and it throws an invalid URL error when I didn't expect it to, for an English Wikipedia Assignment trying to updated it to the URL "https://en.wikipedia.org/wiki/User:Sage_(Wiki_Ed)/test"

app/services/sandbox_url_updator.rb
```diff
@@ -0,0 +1,38 @@
...
+  end
+
+  def update
+    validate_new_url
+    @assignment.update_sandbox_url(@new_url)
+    return { partial: 'updated_assignment', locals: { assignment: @assignment } }
+  rescue StandardError => e
+    case e
+    when SandboxUrlUpdator::InvalidUrlError
+      { json: { errors: e, message: e.message }, status: :bad_request }
```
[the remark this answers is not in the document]
om-chauhan1: sure, instead of status text i am now displaying the exact error message(made for user).

app/services/sandbox_url_updator.rb
om-chauhan1: Now i have fixed the regex pattern matching as previously it was not expecting semicolon after wiki/  . Adding more tests like with spaces, without spaces, with special characters etc in url.

om-chauhan1:
@ragesoss  done the changes

ragesoss:
I tested this out again, and I get an error trying to change a URL:

"There was an error: Cannot update sandbox url. No user with the username Sage_(Wiki_Ed) exists." I first tried this as the instructor using the account "Sage (Wiki Ed)".

I am able to change it when I use other URLs with a different username.

ragesoss:
I think it ought to prepopulate the edit field with the current URL.

app/services/sandbox_url_updator.rb
```diff
@@ -0,0 +1,43 @@
...
+    # Handle invalid url
+    raise InvalidUrlError, I18n.t('assignments.invalid_url', url: @new_url) unless new_url_match
+    # Handle mismatched wiki
+    new_language, new_project, new_username = new_url_match.captures
+    unless existing_language == new_language && existing_project == new_project
+      raise MismatchedWikiError, I18n.t('assignments.mismatched_wiki', url: @new_url)
+    end
+    # Handle user not exist
+    new_user = User.find_by(username: new_username)
+    raise UserNotExist, I18n.t('assignments.no_user_exists', username: new_username) unless new_user
```
ragesoss, line 41: I'm not sure why this raised an error in my testing even for a user that does exist, but I don't think a check for a user is necessary or helpful anyway. I would say checking that it is a non-empty title that includes the `User:` prefix is good enough.

config/locales/en.yml
```diff
@@ -298,6 +298,8 @@ en:
     confirm_deletion: Are you sure you want to delete this assignment?
     delete: Remove and Delete
     editors: Assigned to
+    change_sandbox_url: Change
```
ragesoss, line 301: 'Change Sandbox' I think is better here.

config/locales/en.yml
```diff
@@ -298,6 +298,8 @@ en:
...
+    edit_sandbox_url_placeholder: Edit URL
```
ragesoss, line 302: 'Sandbox URL'

om-chauhan1: okay

app/services/sandbox_url_updator.rb
om-chauhan1: sure, removed that check and done the same

app/assets/javascripts/components/common/AssignmentLinks/EditSandboxUrl.jsx
```diff
@@ -0,0 +1,90 @@
...
+          is_open={is_open}
+          edit_row={editRow}
+          right={true}
+        />
+      </div>
+    );
+  }
+    return (
+      <>
+        <a href="" target="_blank" onClick={this.toggle} className="change-sandbox-url">
```
ragesoss, line 78: It's a little confusing that this is implemented as a link rather than a button, as mousing over it will make it seem like a normal link to the current page.

I would also be good to have a confirmation notification upon successfully changing the URL.

config/locales/en.yml
```diff
@@ -298,6 +298,7 @@ en:
     confirm_deletion: Are you sure you want to delete this assignment?
     delete: Remove and Delete
     editors: Assigned to
+    change_sandbox_url: Change Sandbox
```
ragesoss, line 301: On further reflection, I think this is better as lower case: 'change sandbox', so that feels a little more differentiated from the sandbox link itself.

ragesoss (changes requested):
This is in good shape, but I found a few more things I'd like to adjust before merging it.

app/assets/javascripts/components/common/AssignmentLinks/EditSandboxUrl.jsx
om-chauhan1: okay, to remove that confusion I have converted it to an action button. And also displaying notification, upon successful updation of the sandbox url.

config/locales/en.yml
om-chauhan1: lower case for button also ?

om-chauhan1:
@ragesoss  how is it now ?
After2

om-chauhan1:
this test is failing on master also

ragesoss:
Yeah, I think that works.

PR #5633, opened 2024-02-08
Update default campaign automatically (#5082)

### LINKED ISSUE #5082: Default campaign should be determined automatically
### Current Behavior:

The default term, which is used for the (logged out) home page and the explore page, is set via `/settings` and needs to be changed twice a year for Wiki Education Dashboard. We routinely create spring and fall terms for each year, but we must also update the default term whenever it's time to change terms.

### Desired Behavior:

The default term should be automatically updated at standard dates each year (with a fallback in case the expected new default term does not exist as a Campaign).

gabina:
## What this PR does
This PR schedules an automatic job that runs at the beginning of the spring and fall in northern hemisphere (please double-check these dates):
- March, 21st
- September, 21st

The job performs a default campaign update using `DefaultCampaignUpdate` class. First of all, it determines the current term (for example, 'spring_2024' or 'fall_2024'). If the current term already exists as a campaign (this is manually created each year according to the issue description), then the current term is set as the default campaign.
Otherwise, default campaign is not updated. 

Notice that the update is only done when `wiki_education` feature is set to true, as I understand the scope of the issue is for Wiki Education Dashboard.

Closes #5082

- Double-check the spring and fall start dates.
- Determine if there is any fallback that we want to use when the current term doesn't exist as a campaign.


gabina:
Failing specs are not related to these changes.

ragesoss:
Code and tests look great! I'll give it a spin soon.

spec/workers/sidekiq_cron_jobs_spec.rb
```diff
@@ -33,7 +33,7 @@
   context 'when WikiEd Feature disabled' do
     before { allow(Features).to receive(:wiki_ed?).and_return(false) }
 
-    it 'run default campaign update' do
+    it 'do not run default campaign update' do
```
gabina, line 36: I just fixed this, since the description was incorrectly saying "run default campaign update", when what I wanted to specified was the opposite.

lib/deafult_campaign_update.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
```
ragesoss, line 1: Filename is misspelled.

ragesoss:
We don't actually want to switch these based on the current season, but rather based on the academic calendar where 'spring' is the semester that starts in January. I think the right dates for this feature are January 1 through June 30 for spring, and July 1 through December 31 for fall.

lib/deafult_campaign_update.rb
gabina: Fixed in fd9a70f9c5cded399d1a84ae6c04271312717bd8

gabina:
> We don't actually want to switch these based on the current season, but rather based on the academic calendar where 'spring' is the semester that starts in January. I think the right dates for this feature are January 1 through June 30 for spring, and July 1 through December 31 for fall.

oh that makes sense! I changed the dates on commit 927f8072c3583407f22aab62746d0339f192b4e7, and added some comments to clarify that dates are related to spring/fall semesters based on academic calendar.

gabina:
I closed this PR by mistake and I don't think I'm able to reopen it now. But it should be re-opened, as it's ready for review and merge.

PR #5635, opened 2024-02-09
[WIP] Refactor  `DefaultCampaignForm` component to functional component

gabina:
## What this PR does
Refactor `DefaultCampaignForm` component to functional component.

## Screenshots
Before:

After:

### DIFF
--- a/app/assets/javascripts/components/settings/views/default_campaign_form.jsx
+++ b/app/assets/javascripts/components/settings/views/default_campaign_form.jsx
@@ -1,48 +1,36 @@
-import createReactClass from 'create-react-class';
-import PropTypes from 'prop-types';
-import React from 'react';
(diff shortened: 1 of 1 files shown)


gabina:
> @gabina There is already existing PR #5620 related to this which was opened a week earlier and yet to be merged. All checks are passing on it.

oops, you're right, thank you! I will close this.
I think there is no easy way to check if there is already an open pull request for a given component? Maybe a good idea would be to put the component name on the PR name/description. I actually looked  for pull requests matching "DefaultCampaignForm" before starting this and found nothing. Now I noticed that you're writing the file name (which is usually different from the component name).  Anyway, hopefully in the next few weeks there won't be as many open pull requests and it will be easier to know.

Issue #5636, opened 2024-02-10
Bug: Checkbox state not being preserved in Quality Checklist


PR4NJ41:
### What is happening?
Upon clicking on the "Quality Checklist" button on the homepage of the Course, instructors are experiencing an issue where the checkboxes marked as selected do not persist after closing and reopening the component. 
The component is reopening as expected, but the state of checkboxes is not being preserved. This results in checkboxes losing their state (e.g. checked or unchecked) when the component is reopened, leading to a poor user experience.

### To Reproduce

1. Navigate to the homepage of the Course.
2. Click on the "Quality Checklist" button.
3. Mark checkboxes as selected.
4. Save the changes.
5. Close the quality checklist.
6. Reopen the quality checklist for the same article.

### Current Behavior:
Upon reopening the quality checklist, the previously marked checkboxes are lost, and the state is not persisted.

### Expected Behavior:
The checkboxes marked as selected should persist after saving and reopening the quality checklist.

### Screenshots

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/96368921/04c611f1-5d7a-48da-abd4-40ff8fa8f87f

PR #5637, opened 2024-02-10
Fixed State Persistance of Quality Checklist

PR4NJ41:
## What this PR does
This PR resolves State persistance issue of checkboxes in Quality Checklist.
Fixes issue #5636 

## Preview
Before:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/96368921/9fecfbcc-902f-4a80-a486-bdb4e008e2f1

After:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/96368921/7342bf6a-7a60-4833-b559-bad9730e8d76

### DIFF
--- a/app/assets/javascripts/components/common/final_article_checklist.jsx
+++ b/app/assets/javascripts/components/common/final_article_checklist.jsx
@@ -2,14 +2,36 @@ import React, { useState } from 'react';
 import useOutsideClick from '../../hooks/useOutsideClick';
 
 const FinalArticleChecklist = () => {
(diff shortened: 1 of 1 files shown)


ragesoss:
Thanks! I'll test this soon, but it's worth noting that this will only solve persistence during the lifetime of the React app, so refreshing or navigating away will still result in loss of state. That's okay, though, I don't think this is a feature worth trying to persist on the backend, so this approach is good enough for me.

PR #5638, opened 2024-02-10
enhancement_#3775/ask-instructor-whether-they-have-TAs

### LINKED ISSUE #3775: During Wiki Education course creation, ask instructor whether they have TAs
We want to:
a) ask instructors during course creation whether they have TA support; and
b) send reminder emails after a week if no additional instructor (ie, the TA) has been added to the course yet, for courses that do have a TA.

The first step (this issue) is to collect that information. It should probably be stored in the form of a tag.

After that, we can create a new alert type and alert manager to trigger emails to be sent out.

ujjwalpathaak:
This PR is to resolve #3775 
### Before:
Screenshot from 2024-02-10 15-38-38
### After:
Screenshot from 2024-02-10 15-34-39
### Before:
Screenshot from 2024-02-10 15-42-08
### After:
Screenshot from 2024-02-10 15-26-33


ujjwalpathaak:
Hey @ragesoss, I have added checkboxes to ask if there will be a TA in the program -- it is getting updated in the course object.

LMKWYT. if this is okay I can move to the secound step.

> The first step (this issue) is to collect that information. It should probably be stored in the form of a tag.
> After that, we can create a new alert type and alert manager to trigger emails to be sent out.

https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/3775#issue-568642166

ujjwalpathaak:
Hello @ragesoss, I have added ```ta_support``` in course's flag. 
Course with ta_support checkbox checked
image
Course with ta_support checkbox un-checked
for no ta_support -> just like academic_system the value for is kept as ```empty```
image

LMKWYT, what changees should be made or I can start looking into the email sending feature

ragesoss:
Did you look into implementing it with a Tag instead of via course.flags? I think that's the best way to hold this particular piece of information, as it reflects a choice made during course creation that won't be directly editable afterwards but it will be possible for admins to manually modify it (by adding or removing a Tag) if it is necessary.

ujjwalpathaak:
yes I did try implementing it by using tags, but i thought storing this info in the same table as the course could also be a better solution.
Also pls correct me if I am wrong I might have misinterpreted what you mean by tags -
I should add a new entry into the Tags table with every new course created? with reference to that course id? Just like it is been done for first_time_instructor & returning_instructor

ragesoss:
Yes, similar to `first_time_instructor` and `returning_instructor` is what I had in mind.

ujjwalpathaak:
I'll change the implementation to using tags 
image

ujjwalpathaak:
Program 1 -- checkbox **_ticked_** for ta support
Program 2 -- checkbox **_NOT ticked_** for ta support

questions - what should be the key for ```ta_support```, or i should keep it as NULL like ```cloneable```
image


**courses table**
image
image

**tags table**
image

ujjwalpathaak:
@ragesoss PR is using tags to store the info now -- tests are passing now, let me know if this is what you had in mind.

ujjwalpathaak:
Hello, I'm currently working on `create_no_ta_support_alerts` I am taking help from `create_no_students_alerts` & `create_first_student_alerts` workflow.

As I'm still navigating through the process of setting up the email workflow, I have a couple of questions:
- Is my approach correct till now?
-  how do I test this functionality locally for myself?

ragesoss:
With a quick skim, it looks like you're on the right track.

For testing, you can a) create an email preview, and b) create tests similar to other alerts that use emails to test that everything up through the mailer action is working.

To actually test the full functionally locally, you would need to integrate with an email sending service, but I don't think that's necessary if it follows all the same patterns as similar emails. (I have this set up locally when I need to test such things, but I often don't use it if I've done all the other partial tests and it works very similarly to other emails.)

ujjwalpathaak:
Hey, I was working to get the preview working and setting up the spec. Spec is passing and preview is working fine.
Let me know what all changes are required? Also if I missed something that needs attention.

image
image

ujjwalpathaak:
I am basically -
1) Getting all the courses which have tags marked with ta_support.
2) Iterating for each course.
3) Getting all the users with ```Roles::INSTRUCTOR_ROLE``` in that course from the table ```courses_users```.
4) If the course is 7 days old && ```instructors.size <= 1``` I am checking for an Alert already created.
5) If not then create one and begin the process of sending an email.

ragesoss:
That looks about right! I'll suggest some email text and review the code this week.

app/assets/javascripts/components/course_creator/course_form.jsx
```diff
@@ -17,7 +17,11 @@ const CourseForm = (props) => {
     props.updateCourseProps({ private: isPrivate });
     props.updateCourseAction('private', isPrivate);
   };
-
+  const updateTaSupport = (e) => {
+    const ta_support = e.target.checked;
+    props.updateCourseProps({ ta_support: ta_support });
```
ragesoss, line 22: I'm not really clear on why we have these two separate functions, updateCourseProps and updateCourseAction. It seems like some components use one, some use the other, and only one uses both. I suspect that the one using both is not a great pattern to follow. What is your understanding of these two functions and what they are for?

Also, the less login that lives in this file, the better; I think this new input should be extracted to its own component like some of the other ones in the CourseForm.

app/views/no_ta_enrolled_alert_mailer/email.html.haml
```diff
@@ -0,0 +1,34 @@
...
+
+%table.row
+  %tbody
+    %tr
+      %th
+        %table
+          %tr
+            %td.main-content
+              %p.paragraph
+                = "Hello #{@greeted_users},"
```
ragesoss, line 11: Hi [users].

You indicated that your course has an additional instructor or TA support. It looks like you haven't added anyone else in the instructor role for your dashboard.wikiedu.org course page yet.

Please follow these steps to add an additional instructor:

* Navigate to your [course page](link to course page].
* Click 'Edit Details'.
* Click the '+' icon next to your name in the 'Instructors'.
* Enter the name, username and role (eg, "TA") and click 'Enroll'.

Feel free to get in touch with us if you have any questions.

Best regards,
[name]

config/locales/en.yml
```diff
@@ -540,6 +540,7 @@ en:
         Please include enough detail to give other Wikipedia editors a sense of
         context for the contributions your students will be making.
       course_private: Private program (viewable only by admins and facilitators)
+      course_ta_support: Will this program have more than one instructor?
```
ragesoss, line 543: Will this course have any additional instructors or TAs?

lib/course_creation_manager.rb
```diff
@@ -9,8 +9,9 @@ class CourseCreationManager
 
   # rubocop:disable Metrics/ParameterLists
   def initialize(course_params, wiki_params, scoping_methods, initial_campaign_params,
-                 instructor_role_description, current_user)
+                 instructor_role_description, current_user, ta_support = false)
```
ragesoss, line 12: Is a default to `false` necessary? If this only gets called from the controller, it should always have the parameter included, so leaving off the default value will help us catch any cases where it's being called differently.

ragesoss (changes requested):

app/assets/javascripts/components/course_creator/course_form.jsx
ujjwalpathaak: ```props.updateCourseAction('private', isPrivate);``` in ```course_form.jsx``` is a function passed down by ```course_creator.jsx``` which isn't actually updating the props on it's own. My guess is that the name has "action" in it to tell that it triggers an action to update the prop.

```updateCourseAction``` is a prop name given to  ```updateCourse``` function while passing it down from ```course_creator.jsx``` to ```course_form.jsx```

```
  updateCourse(key, value) {
    this.props.updateCourse({ [key]: value });
    if (includes(['title', 'school', 'term'], key)) {
      return this.props.setValid('exists');
    }
  },
```
Here you can see it basically just adds a check to call setValid and other than that it is just doing the same thing as ```props.updateCourseProps({ ta_support: ta_support });``` does, by converting the input values to a ```key: value``` pair.

As you can see in line **427 ./course_creator.jsx**. ```updateCourseProps``` is just a prop name given to ```this.props.updateCourse```

File: app/views/no_ta_enrolled_alert_mailer/email.html.haml (hunk 2 above)
Comment (ujjwalpathaak) [reply app/views/no_ta_enrolled_alert_mailer/email.html.haml:11]: image

File: app/assets/javascripts/components/course_creator/course_form.jsx (hunk 1 above)
Comment (ujjwalpathaak) [reply app/assets/javascripts/components/course_creator/course_form.jsx:22]: Made another component ```course_checkbox.jsx``` and moved both `ta_support checkbox` and `privacy checkbox` to it

**Comment by ujjwalpathaak** [comment]
I have made all the changes that were requested. let me know if anything else needs to changed.
Should I try to remove the use of updateCourseAction ?

File: app/assets/javascripts/components/course_creator/course_form.jsx (hunk 1 above)
Comment (ragesoss) [reply app/assets/javascripts/components/course_creator/course_form.jsx:23]: So `updateCourseProps` is basically just a wrapper around `updateCourse` which also something related to validation for the title, school and term inputs, while `updateCourseAction` is just the plain `updateCourse` redux action. The title/school/term validation is not relevant for either the new checkbox or the existing one that you've refactored, so I think you can use only simpler one (`updateCourseAction`). Definitely don't use both.

File: app/assets/javascripts/components/course_creator/course_form.jsx [hunk 5]
```diff
@@ -17,7 +18,11 @@ const CourseForm = (props) => {
     props.updateCourseProps({ private: isPrivate });
     props.updateCourseAction('private', isPrivate);
   };
-
+  const updateTaSupport = (e) => {
```
ragesoss, line 21: This update logic looks like it's exactly the same for both checkboxes, just based on the prop key for one versus the other.  It can probably be moved to the `CourseCheckbox` component.

app/assets/javascripts/components/course_creator/course_form.jsx
```diff
@@ -113,7 +119,16 @@ const CourseForm = (props) => {
         </span>
       </div>
     );
-
+    taSupportCheckboxEdu = (
+      <CourseCheckbox
+        id="course_ta_support"
+        label="course_ta_support"
```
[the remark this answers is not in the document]
ujjwalpathaak: I did this to have seperate props for ```htmlFor``` and ```id``` -- if in future any use case for this component wanted to have different values in id and label

changed the component to - 
```
import React from 'react';

const CourseCheckbox = (props) => {
  // if no label is provided, use the checkbox_id as the label
  const htmlFor = props.label || props.checkbox_id;
    return (
      <div className="form-group">
        <label htmlFor={htmlFor}>
          <input
            id={props.checkbox_id}
            type="checkbox"
            value={props.value}
            onChange={props.onChange}
            checked={props.checked}
          />
          {props.text}
        </label>
      </div>
    );
};

export default CourseCheckbox;
```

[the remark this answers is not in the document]
ragesoss: My understanding is that `htmlFor` always takes the ID of the element that the label is for, so I cannot think of a reason why they would ever need to be different. The actual label text is `props.text`.

ujjwalpathaak:
Now using a new prop ```checkboxFor``` to be used for-
- id
- htmlFor
- key

app/views/no_ta_enrolled_alert_mailer/email.html.haml
```diff
@@ -0,0 +1,34 @@
...
+                  Feel free to get in touch with us if you have any questions.
+            %td.expander
+        %table
+          %tr
+            %td.main-content
+              %p.paragraph
+                Best regards,
+                %br
+                %em
+                  = @alert.from_user&.real_name || 'Wiki Ed'
```
ragesoss, line 34: 'Wiki Education' rather than 'Wiki Ed'

lib/alerts/course_alert_manager.rb
```diff
@@ -28,12 +30,17 @@ def self.generate_course_alerts
 
   def initialize
     @courses_to_check = Course.strictly_current.where(withdrawn: false)
+    @tagged_courses_to_check = Tag.courses_tagged_with('ta_support')
```
ragesoss, line 33: Because this set of courses is specific to the new feature rather than used across multiple alert managers, I think this is better done within the manager. Also, you could scope this with `.current_and_future` to ensure that only ongoing courses that aren't already over can get the alerts.

ragesoss (changes requested):
This looks like it's very close.

A few things:

1. This should only be present in Wiki Education mode. The 'Create an Independent Program' form should be unchanged.
2. The checkbox should appear after the 'Your role in the course' input on the lower right.

ujjwalpathaak:
#### Hey, what do you think? which one looks better? I'll keep it and push the changes. ( placement of course Id )
image
image

ragesoss:
The first one definitely. Also, let's remove the word 'any' from the text so that it can fit on a single line.

ujjwalpathaak:
Then I think I have included all changes in the latest commit.

ragesoss:
@ujjwalpathaak remove 'any' from the text.

ujjwalpathaak:
Yes I forgot to push the changes that's why I deleted the last comment.

ujjwalpathaak:
@ragesoss sorry about commenting before pushing the actual changes. I have made the needed commit.

ujjwalpathaak:
image

PR #5649, opened 2024-02-13
Fixed replica_spec.rb and replica.rb

### LINKED ISSUE #5647: Refactor and fix replica_spec 'API Requests > returns a list of existing articles'
### What is happening?
In `spec/lib/replica_spec.rb`, the 'returns a list of existing articles' (currently starting on line 88) does too many things and is unreliable.

See the TODO comment for more details. It should be broken up into a separate test for each article and should more explicitly look for the expected namespaces of each article.

It's also currently broken (and marked pending), so along the way to fixing it you will likely figure out which article title is currently causing a problem and fix that.

om-chauhan1:
## What this PR does
Fixes #5647 
This pr resolve the issue in the test case 'returns a list of existing articles' spec in spec/lib/replica_spec.rb. The original test was performing multiple tasks and was unreliable. To rectify this, I modularized the test into smaller tests, each focusing on individual articles. 

## Screenshots
Before:

image

After:

image

### DIFF
--- a/lib/replica.rb
+++ b/lib/replica.rb
@@ -144,7 +144,8 @@ class Replica
     return if response.body.empty?
     parsed = Oj.load(response.body)
     raise unless parsed['success']
-    parsed['data']
+    filtered_data = parsed['data'].select { |article| data.include?(article['page_title']) }
+    return filtered_data
   rescue StandardError => e
     tries -= 1
     sleep 2 && retry unless tries.zero?
--- a/lib/replica.rb
+++ b/lib/replica.rb
@@ -144,8 +144,7 @@ class Replica
     return if response.body.empty?
     parsed = Oj.load(response.body)
     raise unless parsed['success']
-    filtered_data = parsed['data'].select { |article| data.include?(article['page_title']) }
-    return filtered_data
+    parsed['data']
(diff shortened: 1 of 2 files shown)


lib/replica.rb
```diff
@@ -144,7 +144,8 @@ def api_post(endpoint, key, data)
     return if response.body.empty?
     parsed = Oj.load(response.body)
     raise unless parsed['success']
-    parsed['data']
+    filtered_data = parsed['data'].select { |article| data.include?(article['page_title']) }
+    return filtered_data
```
ragesoss, line 148: Why is this change to replica.rb included?

om-chauhan1: Not required.
Initially, due to the fact that i was seeing previous vcr response again and again, i thought that the api is not filtering out data with the provided tittle. 
Now when i changed the vcr cassette name for recording response, got to know api is working fine as expected.

om-chauhan1:
Found that now article "Autism" does not exist in namespace 118.
Now this test is working fine.

om-chauhan1:
Failing tests are flaky.

ragesoss:
Looks good to me

Issue #4933, opened 2022-04-13
Add 'News' feature to the Dashboard


ragesoss:
### Current Behavior:

Currently, there's no good way to keep users informed of recent changes and other important news/information related to the Dashboard. We occasionally use the `sitenotice` feature to add a banner about particularly important announcements, but this shows up on every page all the time and cannot be dismissed, so it's not suitable for something like "Zoom office hours for Dashboard users will happen on such-and-such date" (except for in the few days immediately before).

The closest alternative is to post on the News section of the Meta page: https://meta.wikimedia.org/wiki/Programs_%26_Events_Dashboard#News_and_recent_changes

However, that isn't easy to follow or discover. 

### Desired Behavior:

There should be a News feature in the navbar that lets users quickly see a small feed of recent news items, along with a way for admins to edit the news items. Ideally, it might also show highlighting when the news has been updated recently, so that it's easy to discover a new update.

### Related Files & Code:

One possible way to build this would be to make the news content come from a special FAQ record, so that the editing/updating part of it is basically free since we've already implemented the FAQ system that can be edited by admins.

Another possibility would be to build it on top of the training module system, where the news content is basically a special TrainingSlide with content — and translations — pulled from Meta.

vaidehi44:
For this feature, I guess a simple pop up kind of thing, like the one which exists for `Get Help` button, would be enough, @ragesoss ?

ragesoss:
Yes, something like that might work well... although layout would probably be a little different, to flexibly accommodate a lot or a little text. The admin notification bell might also be a good pattern to follow, in terms of the non-popped-up state... I think the ideal situation would be to show some kind of subtle indicator for unread news, which would use a cookie (with some kind of hash indicating the version, so that updating the news content would automatically make it re-activate) to remember that a given user has already looked at it.

rajat99-default:
I would like to work on this issue
also I need some guidance on how to do this because this can be my first open source contribution

ragesoss:
@rajat99-default I think the other issue you posted on will be a better first contribution. If you'd like an invite to our slack channel, please let me know your email address.

rajat99-default:
My email is 
rajatsujane6@gmail.com

Abishekcs:
@ragesoss i can work on this issue

ragesoss:
> @ragesoss i can work on this issue

Great!

Abishekcs:
[...]

PR #5650, opened 2024-02-13
Add admin-only Notes feature for courses

### LINKED ISSUE #5526: Add admin-only Notes feature for courses
### Current Behavior:
There is no way within the Dashboard for Wiki Education admins to record notes about courses they are supporting.

### Desired Behavior:

There should be a way for an admin to edit a Notes field related to a course, see that existing notes are already present from the Home tab.

### Related Files & Code:

It think the best option for implementing this would be to add a new CourseNote model and table, and use a single editable field to record the notes for a given course.

Abishekcs:
closes #5526

## What this PR does

Enable only `Admins`  to `create, edit, read and  delete` Course Notes

## Screenshots
Before:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/46e933b0-01de-4326-8b18-4c5d9e4ab1e8

After:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/a5515635-9191-4400-a410-f152316c5077


Abishekcs:
@ragesoss  could you please review my code whenever you are free? Still working on the test code for this feature. Thanks!

ragesoss:
When I create a note, I get this error:

```
TypeError
Cannot read properties of undefined (reading 'id')
Call Stack
 _callee4$
  app/assets/javascripts/actions/course_notes_action.js:205:29
 tryCatch
  app/assets/javascripts/actions/course_notes_action.js:29:2404
 Generator.eval [as _invoke]
  app/assets/javascripts/actions/course_notes_action.js:29:1964
 Generator.eval [as next]
  app/assets/javascripts/actions/course_notes_action.js:29:3255
 asyncGeneratorStep
  app/assets/javascripts/actions/course_notes_action.js:31:103
 _next
  app/assets/javascripts/actions/course_notes_action.js:33:194
```

ragesoss:
Actually, that error happens after the server returns a 422. It looks like this is because 'edited by' is blank.

But the frontend should render an error notification if there's an error response returned by the server.

app/controllers/course_notes_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  def update
+    if @course_note.update_note(course_note_params)
+      render json: { success: true }
+    else
+      render json: { error: 'Failed to update course note' }, status: :unprocessable_entity
+    end
+  end
+
+  def create
+    note_details = CourseNote.new.create_new_note(course_note_params)
```
ragesoss, line 24: This is not the idiomatic Rails way to create a new record. If you do `CourseNote.create(attrributes)` instead, it will return a CourseNote note object which you can check for validation errors (by calling `.errors` on it). Or you can do `CourseNote.create!` and it will raise those errors directly so that you can rescue them and return an appropriately detailed error response.

ragesoss:
It works as long as I'm also enrolled in the Wiki Ed Staff role, but it does not work if I'm an admin who is not part of the course.

Abishekcs:
@ragesoss, once I've addressed all these issues, I'll let you know. Thank you for the review!

Abishekcs:
> It works as long as I'm also enrolled in the Wiki Ed Staff role, but it does not work if I'm an admin who is not part of the course.

😅 It was working for me. I will figure out what's the problem. Ty for pointing it out!

Abishekcs:
> 😅 It was working for me. I will figure out what's the problem. Ty for pointing it out!

I just found the problem: if the user is enrolled in the Wiki Ed Staff, `currentUser.username` is available. However, if the user is only an admin, `currentUser.username` is not available

Abishekcs:
> It works as long as I'm also enrolled in the Wiki Ed Staff role, but it does not work if I'm an admin who is not part of the course.

@ragesoss, It wasn't working because if the `user` is only an ` Admin and not a Wiki Ed Staff`,  then the Admin `username`  is not available in the `current_user` which is passed as "prop" from  `admin_quick_action`.

So, I am getting the admin name from `nav_root` which I'm not sure if it is a good approach 😅 Your suggestion would be greatly appreciated on this. Ty!

Below, is the code from `notes_handler.jsx` which make sure to add the `admin`  username if not available:

```
// If user is Admin and not Wiki Ed Staff, fetch Admin username from 'nav_root'.
// If user is Admin and Wiki Ed Staff, username is already in CurrentUser.

  if (!currentUser.username) {
    const { username } = document.getElementById('nav_root').dataset;
    currentUser = { ...currentUser, username };
  }
```

ragesoss:
I suggest that you don't bother with getting the username on the client side at all; you can get the current user server-side with `current_user`.

Abishekcs:
> > I suggest that you don't bother with getting the username on the client side at all; you can get the current user server-side with `current_user`.
> 
> @ragesoss, I've implemented the changes, and it should now work for cases where the user is an admin but not part of the course.

@ragesoss any update on this PR😅

ragesoss:
Thanks for the reminder. I hope to review this today or tomorrow.

ragesoss:
This works pretty smoothly. I think there's some room for improvement with the UI:

1. When viewing one note, then going back to the index and clicking a different note, the previous note renders briefly before it refreshes to the new one.
2. It's confusing to use the 'X' icon to return to the index when viewing. I guess that should be a 'back' button. However, I don't really like that the whole index and header disappear when entering view mode for a single note. Maybe instead of changing to a view of just a single note, clicking on a note could trigger an expanded accordion section to view that note within the table? Or maybe there's some other design that would feel more intuitive.
3. The 'Edit Notes' button isn't a great label, as that's where you have to click to either create a new note or delete an existing one... but not to edit an existing one.

Abishekcs:
> This works pretty smoothly. I think there's some room for improvement with the UI:
> 
> 1. When viewing one note, then going back to the index and clicking a different note, the previous note renders briefly before it refreshes to the new one.
> 2. It's confusing to use the 'X' icon to return to the index when viewing. I guess that should be a 'back' button. However, I don't really like that the whole index and header disappear when entering view mode for a single note. Maybe instead of changing to a view of just a single note, clicking on a note could trigger an expanded accordion section to view that note within the table? Or maybe there's some other design that would feel more intuitive.
> 3. The 'Edit Notes' button isn't a great label, as that's where you have to click to either create a new note or delete an existing one... but not to edit an existing one.
> 
Ty for the review @ragesoss I will make the required changes needed to improve the UI😁

ragesoss:
> > 2. Maybe instead of changing to a view of just a single note, clicking on a note could trigger an expanded accordion section to view that note within the table? Or maybe there's some other design that would feel more intuitive.
> 
> @ragesoss I am also thinking of an expanded accordion for creating new notes within the same table instead of opening a new table. Or would it be better if creating a note opens a new table entirely?

I think it will be very rare for a course to have more than one or two notes, so I think keeping the same table and creating a new note within it as you propose will be a good option.

Abishekcs:
> > > 2. Maybe instead of changing to a view of just a single note, clicking on a note could trigger an expanded accordion section to view that note within the table? Or maybe there's some other design that would feel more intuitive.
> > 
> > 
> > @ragesoss I am also thinking of an expanded accordion for creating new notes within the same table instead of opening a new table. Or would it be better if creating a note opens a new table entirely?
> 
> I think it will be very rare for a course to have more than one or two notes, so I think keeping the same table and creating a new note within it as you propose will be a good option.

Okay, great. I will make the changes to the UI according to this.

Abishekcs:
@ragesoss, I've made some changes to the UI. Below is a recording showcasing the updates. It's still a **`work in progress`**. If this looks okay to you, I'll proceed with this UI. For creating notes, I'll use something similar to the UI of [ News Feature](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/4933#issuecomment-1954197170) . Please  let me know if any adjustments are needed. Additionally, I think with this UI change, the number of lines of code will also reduce significantly.


https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/09b7b7bd-0c04-465c-9364-4b2ef18cfa39

ragesoss:
@Abishekcs yeah, I like this!

Abishekcs:
@ragesoss any update on this PR?

ragesoss:
@Abishekcs not yet, but it's on my todo list!

ragesoss:
This looks good, thanks!

Issue #5645, opened 2024-02-12
In development environment, there are warnings and errors when starting `guard`


ragesoss:
### What is happening?
When I start `guard` for a local development server, I get the following output:

```
Warning: you have a Gemfile, but you're not using bundler or RUBYGEMS_GEMDEPS
11:03:12 - INFO -
> [#] Guard here! It looks like your project has a Gemfile, yet you are running
> [#] `guard` outside of Bundler. If this is your intent, feel free to ignore this
> [#] message. Otherwise, consider using `bundle exec guard` to ensure your
> [#] dependencies are loaded correctly.
> [#] (You can run `guard` with --no-bundler-warning to get rid of this message.)
11:03:13 - INFO - Bundle already up-to-date
11:03:13 - INFO - [Guard::Rails] will start the default web server on port 3000 in development.
11:03:13 - INFO - Starting Rails...

=> Booting Puma
=> Rails 7.0.7 application starting in development 
=> Run `bin/rails server --help` for more startup options
Redis#sadd will always return an Integer in Redis 5.0.0. Use Redis#sadd? instead.(called from: /home/sage/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sidekiq-cron-1.7.0/lib/sidekiq/cron/job.rb:468:in `block in save')
2024-02-12T19:03:14.435Z pid=1422661 tid=ug71 INFO: Cron Jobs - added job with name: schedule_course_updates
Redis#sadd will always return an Integer in Redis 5.0.0. Use Redis#sadd? instead.(called from: /home/sage/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sidekiq-cron-1.7.0/lib/sidekiq/cron/job.rb:468:in `block in save')
2024-02-12T19:03:14.436Z pid=1422661 tid=ug71 INFO: Cron Jobs - added job with name: update_constantly
Redis#sadd will always return an Integer in Redis 5.0.0. Use Redis#sadd? instead.(called from: /home/sage/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/sidekiq-cron-1.7.0/lib/sidekiq/cron/job.rb:468:in `block in save')
2024-02-12T19:03:14.436Z pid=1422661 tid=ug71 INFO: Cron Jobs - added job with name: update_daily
[...]

PR #5651, opened 2024-02-13
Fixing warnings when starting `Guard`

PR4NJ41:
## What this PR does
Bumped `sidekiq-cron` from v1.7.0 to v1.12.0.

This will fix warnings caused by it.

Addresses issue #5645 

## Screenshots
Before:
Before

After:
After

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,7 +20,7 @@ gem 'paper_trail' # Save histories of record changes related to surveys
 gem "kt-paperclip" # used by Course and UserProfile for file attachments.
 gem 'sidekiq' # Framework for running background worker jobs
 gem 'sidekiq-unique-jobs' # Plugin to prevent duplicate jobs in the sidekiq queue
-gem 'sidekiq-cron' # Plugin for cron-style recurring jobs in Sidekiq
+gem 'sidekiq-cron', '~> 1.12.0' # Plugin for cron-style recurring jobs in Sidekiq
 gem 'dalli' # Caching
 gem 'connection_pool'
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -20,7 +20,7 @@ gem 'paper_trail' # Save histories of record changes related to surveys
 gem "kt-paperclip" # used by Course and UserProfile for file attachments.
 gem 'sidekiq' # Framework for running background worker jobs
 gem 'sidekiq-unique-jobs' # Plugin to prevent duplicate jobs in the sidekiq queue
-gem 'sidekiq-cron' # Plugin for cron-style recurring jobs in Sidekiq
+gem 'sidekiq-cron', '~> 1.12.0' # Plugin for cron-style recurring jobs in Sidekiq
```
ragesoss, line 23: This is not necessary, as the lock file will ensure that the new version gets used. Typically we include as few explicit version constraints in the Gemfile as possible.

ragesoss:
Please include a link to the sidekiq-cron changelog and a summary of what changed between these versions, so we know what to look out for in terms of bugs or regressions that this update might introduce.

Gemfile
ragesoss: The updated Gemfile.lock will ensure that 1.12.0 is used instead. To update the lock file, you need to start by either making a change like this to the Gemfile, or removing the entries for that gem from the lock file and then re-installing the gems to regenerate the lock file, or using bundler from the command line to update a particular gem. But if you use the Gemfile method like you did, it's fine to then remove the version constraint after you update the lock file.

Issue #5652, opened 2024-02-13
assigned_articles_spec.rb is flaky


ragesoss:
### What is happening?

The 'Assigned Articles' feature spec fails intermittently on CI. For example: https://github.com/WikiEducationFoundation/WikiEduDashboard/actions/runs/7889907818/job/21530874633

```
  1) Assigned Articles view lets users submit feedback about articles
     Got 0 failures and 2 other errors:

     1.1) Failure/Error: find('textarea.feedback-form').fill_in with: 'This is a great article!'

          Capybara::ElementNotFound:
            Unable to find css "textarea.feedback-form"
          # ./spec/features/assigned_articles_spec.rb:26:in `block (3 levels) in <top (required)>'
          # ./spec/features/assigned_articles_spec.rb:22:in `block (2 levels) in <top (required)>'

     1.2) Failure/Error: Capybara::Screenshot.new.screenshot_and_save_page if example.exception

          NoMethodError:
            undefined method `new' for Capybara::Screenshot:Module

                Capybara::Screenshot.new.screenshot_and_save_page if example.exception
                                    ^^^^
          # ./spec/rails_helper.rb:101:in `block (2 levels) in <top (required)>'
```

Update the test to pass reliably, and describe what made the test unreliable if you can.

It may be useful to remove the `--headless=new` parameter from the Capybara config in rails_helper.rb so that you can watch the test run in the browser. (Be sure not to include that change to rails_helper in the PR.)

ragesoss:
Yes, I opened a separate issue for the Capybara screenshot error.

ragesoss:
This seems related to too-slow network requests that rely on external services. I think the right fix is probably to identify the slow request, and then stub it (with a new method in request_helpers.rb) so that the test can run quickly without actually hitting the slow API.

nexpectArpit:
@ragesoss is this issue still persist?

ragesoss:
@nexpectArpit I'm not sure. You can browse the recent builds to see if it comes up in any of the ones that failed.

PR #5654, opened 2024-02-13
Fixed Rspec Checks Failure of `assigned_articles_spec.rb`

PR4NJ41:
## What this PR does
This PR fixes Rspec checks failure which were failing due to `assigned_articles_spec.rb` file.

Addresses Issue #5652 

## Screenshots
Before:

After:
Screenshot 2024-02-14 at 12 10 47 AM

### DIFF
--- a/spec/features/assigned_articles_spec.rb
+++ b/spec/features/assigned_articles_spec.rb
@@ -21,12 +21,12 @@ describe 'Assigned Articles view', type: :feature, js: true do
     # we need to use VCR to avoid getting stopped by WebMock
     VCR.use_cassette('assigned_articles_view') do
       visit "/courses/#{course.slug}/articles/assigned"
-      expect(page).to have_content('Nancy Tuana')
+      page.has_content?('Nancy Tuana')
       find('a', text: 'Feedback').click
-      find('textarea.feedback-form').fill_in with: 'This is a great article!'
+      find('textarea.feedback-form').fill_in(with: 'This is a great article!')
       click_button 'Add Suggestion'
       find('a', text: 'Delete').click
       expect(page).not_to have_content('This is a great article!')
-    end
+    end    
   end
 end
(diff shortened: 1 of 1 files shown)


spec/features/assigned_articles_spec.rb
```diff
@@ -21,9 +21,9 @@
     # we need to use VCR to avoid getting stopped by WebMock
     VCR.use_cassette('assigned_articles_view') do
       visit "/courses/#{course.slug}/articles/assigned"
-      expect(page).to have_content('Nancy Tuana')
+      page.has_content?('Nancy Tuana')
```
ragesoss, line 24: This looks like it simply removes the check for the expected article, so that if `.has_content?` returns `false` it doesn't matter and the test continues... whereas the `expect` line also relies on Capybara to wait a certain amount of time and continue checking for the content, so it ensure that the assigned articles list has rendered before moving on. I don't think will make it more reliable.

ragesoss:
The test usually passes, so if you weren't able to replicate the failure locally, I suspect that this won't improve the test.

spec/features/assigned_articles_spec.rb
```diff
@@ -21,7 +21,7 @@
     # we need to use VCR to avoid getting stopped by WebMock
     VCR.use_cassette('assigned_articles_view') do
       visit "/courses/#{course.slug}/articles/assigned"
-      page.has_content?('Nancy Tuana')
+      expect(page).to have_content('Nancy Tuana', wait:20)
```
ragesoss, line 24: I'm not certain about how `page.has_content?` works, but `have_content` automatically waits and continues checking the page (up to a certain amount of time) until the expected content is found, so there's no need for the `wait` argument here.

ragesoss:
Note that this spec still failed on CI for this PR.

ragesoss:
Hmm... the default wait time is pretty long; what makes you think that waiting even longer is the solution? Why does it sometimes take that long to load? My guess was that when it fails, it's not because of being slow to render, but that something else went wrong such as a request that failed or didn't return the expected data. But I guess that `manual_update` can take a while. The solution might instead be to adjust the test setup so that a potentially slow manual update is not needed.

ragesoss:
That 'manual_update' speculation is relevant for another flaky spec, but not this one. But it still seems related to too-slow network requests that rely on external services. I think the right fix is probably to identify the slow request, and then stub it (with a new method in `request_helpers.rb`) so that the test can run quickly without actually hitting the slow API.

Issue #5653, opened 2024-02-13
Capybara screenshot feature in tests is broken


ragesoss:
### What is happening?

When a feature test fails, the configuration in rails_helper.rb is used to attempt to make a screenshot (which can be helpful for diagnosing a failing browser-based feature spec). The Capybara screenshot interface seems to have changed, so the call for creating a screenshot throws a NoMethodError. For example, the second part of this failure output:

```
  1) Assigned Articles view lets users submit feedback about articles
     Got 0 failures and 2 other errors:

     1.1) Failure/Error: find('textarea.feedback-form').fill_in with: 'This is a great article!'

          Capybara::ElementNotFound:
            Unable to find css "textarea.feedback-form"
          # ./spec/features/assigned_articles_spec.rb:26:in `block (3 levels) in <top (required)>'
          # ./spec/features/assigned_articles_spec.rb:22:in `block (2 levels) in <top (required)>'

     1.2) Failure/Error: Capybara::Screenshot.new.screenshot_and_save_page if example.exception

          NoMethodError:
            undefined method `new' for Capybara::Screenshot:Module

                Capybara::Screenshot.new.screenshot_and_save_page if example.exception
                                    ^^^^
          # ./spec/rails_helper.rb:101:in `block (2 levels) in <top (required)>'
```

Fix this so that screenshots are properly generated when a feature spec fails.

PR #5657, opened 2024-02-14
Fixed Capybara screenshot feature in broken tests

atulyadav745:
## What this PR does
This PR addresses the issue #5653  where, in the event of a failing feature test, the code in rails_helper.rb attempts to capture a screenshot using Capybara. However, due to changes in the Capybara screenshot interface, the method call for creating a screenshot results in a NoMethodError. This PR resolves the issue by updating the method call to match the changed Capybara screenshot interface, ensuring that screenshots are successfully generated when a feature spec fails.

 @ragesoss  please review this pull request and let me know if there are any issues that need to be addressed before it can be merged, or if it is ready to proceed.
Thank You

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -98,7 +98,7 @@ RSpec.configure do |config|
     # them.
     # Instead, we clear and print any after-success error
     # logs in the `before` block above.
-    Capybara::Screenshot.new.screenshot_and_save_page if example.exception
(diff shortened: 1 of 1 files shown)


ragesoss:
@atulyadav745 did you test this locally? Please post a screenshot from test. (You could do this by adding an explicit error to the middle of a feature spec and then running it.)

PR #5662, opened 2024-02-15
Fixes 500 links loading issue

### LINKED ISSUE #5588: Fix 500-link limit for loading training slides from wiki

fillingtothemomo:
fixes #5588 

## What this PR does
Implements continue in  wiki_source_pages method in `lib/training/wiki_training_loader.rb` which is responsible for fetching the links and earlier it was only fetching 500 links and now it fetches till continue parameter becomes nil.
## Screenshots
Before:
image
After:
image

### DIFF
--- a/lib/training/wiki_training_loader.rb
+++ b/lib/training/wiki_training_loader.rb
@@ -91,18 +91,25 @@ class WikiTrainingLoader
     full_content
   end
 
-  # Gets a list of page titles linked from the base page
-  def wiki_source_pages
-    # To handle more than 500 pages linked from the source page,
-    # we'll need to update this to use 'continue'.
-    query_params = { prop: 'links', titles: @wiki_base_page, pllimit: 500 }
-    response = WikiApi.new(MetaWiki.new).query(query_params)
-    begin
-      response.data['pages'].values[0]['links'].map { |page| page['title'] }
-    rescue StandardError
-      raise InvalidWikiContentError, "could not get links from '#{@wiki_base_page}'"
+# Gets a list of page titles linked from the base page
+def wiki_source_pages
+  # To handle more than 500 pages linked from the source page,
+  # we'll need to update this to use 'continue'.
+  query_params = { prop: 'links', titles: @wiki_base_page, pllimit: 500 }
+  response = WikiApi.new(MetaWiki.new).query(query_params)
+  @query.merge! @continue unless @continue.nil?
+  begin
+    page_titles = response.data['pages'].values[0]['links'].map { |page| page['title'] }
+    while response['continue']
+      query_params[:plcontinue] = response['continue']['plcontinue']
+      response = WikiApi.new(MetaWiki.new).query(query_params)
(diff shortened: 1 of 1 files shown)


fillingtothemomo:
@ragesoss I realized the issue in the previous PR as I was extracting continue parameter from response.data whereas it should have been extracted from response only. I believe now the modules are being loaded fine. Please review the PR.

lib/training/wiki_training_loader.rb
```diff
@@ -91,18 +91,25 @@ def training_content_and_translations(content:, base_page:, wikitext:)
...
-    query_params = { prop: 'links', titles: @wiki_base_page, pllimit: 500 }
-    response = WikiApi.new(MetaWiki.new).query(query_params)
-    begin
-      response.data['pages'].values[0]['links'].map { |page| page['title'] }
-    rescue StandardError
-      raise InvalidWikiContentError, "could not get links from '#{@wiki_base_page}'"
+# Gets a list of page titles linked from the base page
+def wiki_source_pages
+  # To handle more than 500 pages linked from the source page,
+  # we'll need to update this to use 'continue'.
```
gabina, line 97: If this PR implements the use of 'continue' then you should remove/update this comment

gabina, line 97: If this PR implements the use of 'continue' then you should remove/update this comment

lib/training/wiki_training_loader.rb
```diff
@@ -91,18 +91,25 @@ def training_content_and_translations(content:, base_page:, wikitext:)
...
+  query_params = { prop: 'links', titles: @wiki_base_page, pllimit: 500 }
+  response = WikiApi.new(MetaWiki.new).query(query_params)
+  @query.merge! @continue unless @continue.nil?
```
gabina, line 100: What is the sense of this line? If `@continue` is not previously defined then it will be always `nil`?

gabina: @fillingtothemomo my previous comment was in relation to the following line:
`@query.merge! @continue unless @continue.nil?`

what I meant is that if `@continue` is not previously defined, then that line has no effect.

fillingtothemomo: @gabina I looked into continue implementation as in transclusion importer and I didnt see any explicit initialisation of @continue , can you elaborate on that?

gabina: Could you please explain me what the following line does?
`@query.merge! @continue unless @continue.nil?`

gabina: And what is the content of  `@continue`?

fillingtothemomo: It's nil currently.  I tried with ` @continue = response['continue'] ` as well so it wouldn't be null, but didn't succeed

gabina: Ok, you have to fix that. What the answer looks like? That should help you understand why `@continue = response['continue']` doesn't work.

PR #5665, opened 2024-02-19
Revise reducer feedback.js and add a test

zhjiang1103:
## What this PR does
< This PR revises the feedback.js file. It creates a new custom array for newState[action.assignmentId], including the existing items (if any) and the new feedback item. This ensures that the state is not mutated directly, avoiding the "Cannot add property 0, object is not extensible" error.  This PR also adds a test file for reducer feedback.js>

## Screenshots
Before:
Screen Shot 2024-02-18 at 11 41 11 PM

#2082


ragesoss:
This PR includes significant changes to the JS dependency versions in package-lock.json. Are those changes necessary for the PR? It looks like you've added the `jest-canvas-mock` library but I don't see it being imported anywhere.

zhjiang1103:
Thank you for pointing out the issue. I believe the change is a result of the updates to the Node version. I have reverted the change and created a new pull request for the files I modified to avoid any confusion. Please refer to this new pull request for the updated changes. PR: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5683

PR #5667, opened 2024-02-20
Made 'impact_stats' editable from /settings

### LINKED ISSUE #5658: Make the 'impact stats' from `_wiki_ed_impact.html.haml` editable from `/settings`
Periodically, we update the stats shown on the logged-out home page of dashboard.wikiedu.org.

Currently, this requires a commit like https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/cf238a39ff2e0733f1bfbf7d501ed0434a4837e4 followed by redeploying the Dashboard.

Instead, those values should be persisted in an editable form so that we can update them via the frontend in `/settings`. (This requires super-admin permission (`permissions: 3`) on your User record to access the settings page.) The data can probably live in a new Setting record.

It should be done with caching so that no database queries are required when rendering the impact template (unless the cache has expired), and updating the values from the settings menu should also expire the cache.

om-chauhan1:
## What this PR does
Fixes #5658 
This PR allow users with super admin permission to update/edit the 'impact stats' of  _wiki_ed_impact.html.haml through settings page.

## Screenshots/Videos
Before:

Before

After:

[After(Compressed).webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/18df97d4-74d6-40e1-aa2d-2949b8b73e2a)

### DIFF
--- a/app/assets/javascripts/components/settings/settings_handler.jsx
+++ b/app/assets/javascripts/components/settings/settings_handler.jsx
@@ -40,21 +40,24 @@ export const SettingsHandler = createReactClass({
   },
 
   render() {
-    const otherSettings = (
-      <React.Fragment>
-        <h1 className="mx2 mt4">Other settings</h1>
-        <hr />
-        <h2 className="mx2">Update</h2>
-        <UpdateImpactStats />
-        <br /> <br />
-        <h2 className="mx2">Salesforce</h2>
-        <UpdateSalesforceCredentials />
-        <br /> <br />
-        <CourseCreationSettings settings={this.props.courseCreation}/>
-        <br />
-        <DefaultCampaignSetting defaultCampaign={this.props.defaultCampaign}/>
-      </React.Fragment>
-    );
+    let otherSettings;
+    if (Features.wikiEd) {
+      otherSettings = (
+        <React.Fragment>
+          <h1 className="mx2 mt4">Other settings</h1>
+          <hr />
+          <h2 className="mx2">Update</h2>
+          <UpdateImpactStats />
+          <br /> <br />
+          <h2 className="mx2">Salesforce</h2>
+          <UpdateSalesforceCredentials />
+          <br /> <br />
+          <CourseCreationSettings settings={this.props.courseCreation}/>
+          <br />
+          <DefaultCampaignSetting defaultCampaign={this.props.defaultCampaign}/>
+        </React.Fragment>
+      );
+    }
     return (
       <div id="settings" className="mt4 container">
         <Notifications />
--- a/app/assets/javascripts/components/settings/settings_handler.jsx
+++ b/app/assets/javascripts/components/settings/settings_handler.jsx
@@ -12,6 +12,7 @@ import SpecialUserList from './special_users_list';
 import UpdateSalesforceCredentials from './views/update_salesforce_credentials';
 import CourseCreationSettings from './course_creation_settings';
 import DefaultCampaignSetting from './default_campaign_setting';
+import UpdateImpactStats from './views/update_impact_stats';
 
 export const SettingsHandler = createReactClass({
   propTypes: {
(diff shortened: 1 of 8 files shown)


om-chauhan1:
### Working of Caching

[Working of Cache(Compressed).webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/d6207a17-2c1c-41dc-9e04-04e2a3412d04)

om-chauhan1:
@ragesoss  Sir, the only thing i am concerned about is that after merging this code, initially we need to update the current impact_stats through /setting page. Once the entry is created in db then everything will work smoothly, but before that it won't display that "log out" page which render "_wiki_ed_impact.html.haml" partial bcz things are not hard coded there.

app/controllers/home_controller.rb
```diff
@@ -8,5 +8,26 @@ class HomeController < ApplicationController
...
+        total_pages: '361,000',
+        volumes: '77',
+        article_views: '438',
+        universities: '800'
+      }
+    end
+  end
+
+  def fetch_statistics
+    Rails.cache.fetch('impact_stats', expires_in: 12.hours) do
```
ragesoss, line 28: Why is the cache set to expire in 12 hours? Is there anything that can change without going through the cache deletion upon updating the stats?

om-chauhan1: i will remove this expiry, as every change in these stats automatically trigger cache deletion.

app/controllers/home_controller.rb
```diff
@@ -8,5 +8,26 @@ class HomeController < ApplicationController
   def index
     campaign_slug = CampaignsPresenter.default_campaign_slug
     @presenter = CoursesPresenter.new(current_user:, campaign_param: campaign_slug)
+    @stats = fetch_statistics
+    if @stats.nil?
+      # Dummy data when stats are not available initially or while running rspec tests
+      @stats = {
```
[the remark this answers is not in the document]
om-chauhan1: okay, now i am returning an empty hash for tests and initial deployment.

app/views/home/_wiki_ed_impact.html.haml
```diff
@@ -27,36 +27,42 @@
...
+          million. That's six times larger than the population of France.
 
 %section.universities
   .container
     %h2
       Wiki Education staff have supported instructors and students in more than
-      %b 800 universities
-      across the United States and Canada
+      %b= @stats['universities']
+      universities across the United States and Canada
```
ragesoss, line 65: "universities" should also remain in bold.

app/views/home/_wiki_ed_impact.html.haml
```diff
@@ -27,36 +27,42 @@
...
+          The number of words students have donated to the world. That’s 
+          = @stats['total_pages']
+          pages – the equivalent of 
+          = @stats['volumes']
+          volumes of a printed encyclopedia, shared with everyone, for free.
 
       .col
         %h3 Article views, in a single term
-        %h4 438 Million
+        %h4= @stats['article_views']
```
ragesoss, line 54: This should probably in `article_views_in_millions`, because it's used twice but once is capitalized and once is not. In the current version, including "Millions" in the value results in "438 Million million" below.

app/assets/javascripts/components/settings/settings_handler.jsx
```diff
@@ -43,11 +44,14 @@ export const SettingsHandler = createReactClass({
     if (Features.wikiEd) {
       otherSettings = (
         <React.Fragment>
-          <h1 className="mx2">Other settings</h1>
+          <h1 className="mx2 mt4">Other settings</h1>
           <hr />
+          <h2 className="mx2">Update</h2>
```
ragesoss, line 49: header should be "Impact Stats" here.

ragesoss (changes requested):
Overall this looks good. A few tweaks and it will be ready to go.

app/views/home/_wiki_ed_impact.html.haml
om-chauhan1: okay, 'article_views' now only hold the value in number eg: 438 which will remain same for both and i have appended Million and million after it accordingly.

app/views/home/_wiki_ed_impact.html.haml
om-chauhan1: done

app/assets/javascripts/components/settings/settings_handler.jsx
om-chauhan1: done

app/assets/javascripts/components/settings/views/impact_stats_form.jsx
```diff
@@ -0,0 +1,106 @@
...
+              label="Volumes"
+            />
+            <TextInput
+              id="article_views"
+              editable
+              onChange={handleChange}
+              value={stats.article_views}
+              value_key="article_views"
+              type="text"
+              label="Article Views in a single term"
```
ragesoss, line 84: This label as well as the 'added words' label should indicate "(in millions)" or similar so that we don't get mixed up.

om-chauhan1: done

Issue #5543, opened 2023-12-02
Error creating course when no campaign exists 


prathamVaidya:
### What is happening?
If no campaigns are created and user tries to to create a course. This server throws a 500 Error in response with following log:

```
[2023-12-02 14:21:36.347 FATAL] ActiveRecord::AssociationTypeMismatch (Campaign(#117300) expected, got nil which is an instance of NilClass(#40)):
``` 

This is understandable that campaign is a requirement to create a course but logic should be added to handle the case and throw error message indicating the real reason.

### To Reproduce

 1. Go to '[...](http://localhost:3000/course_creator)'. Make sure no campaigns are created
 2. Fill the required fields and Click on 'Create My Program'.
 3. Red Error toast will show saying 'There was an error: Internal server error'  

### Expected behavior
Proper Logic should be added to handle the case and show appropriate error message to fix this error.

### Screenshots

![image](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/61202986/c89b7cd9-0a5f-4d6e-b5ce-61af94cf93c6)

### Additional context
While this is a minor and unimportant issue for production use cases as there will always be a campaign, it can create a confusion for beginners trying to setup or try the project.

The check for campaign could be added to `course_creation_manager.rb` or `courses_controller.rb`.

prathamVaidya:
While setting up the project  `rake dev:populate` can be used to create example courses to avoid creating courses manually.
Refer: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/user_roles.md#creating-a-courseprogram

RachitPrakash:
Hii can you please explain the problem a little more elaborately as I am not able to get to this issue.

prathamVaidya:
@ragesoss what do you think about the issue? Should it be fixed?

ragesoss:
@prathamVaidya thanks! i think fixing it would be helpful for newcomers trying to get started with code contributions.

I'm not sure what the best strategy to fix it would be. One possibility would be to create a new error type that gets caught only in development mode to display the solution (eg, to run the dev populate task). We do this already for some types of development-specific errors; see `lib/errors/rescue_development_errors.rb`, which could potentially be extended to cover this.

PR #5678, opened 2024-02-27
fix: Issue #5543 Handle case if no campaign is available when creating program

prathamVaidya:
## What this PR does

Fixes Issue #5543 

If no campaigns are created and user tries to create a course. This server throws a 500 Error in response with following log:

```
[2023-12-02 14:21:36.347 FATAL] ActiveRecord::AssociationTypeMismatch (Campaign(#117300) expected, got nil which is an instance of NilClass(#40)):
```
## Screenshots

Before:

image

After:

image

## Solution 
 
Added a missing case. This allows program to be created without a campaign. If this should not be the case in `wiki_education=false` then, let me know I'll update the case to throw relevant error message.

### DIFF
--- a/lib/course_creation_manager.rb
+++ b/lib/course_creation_manager.rb
@@ -99,11 +99,12 @@ class CourseCreationManager
     return unless Features.open_course_creation?
 
     @overrides[:campaigns] = if @initial_campaign_params.present?
-                               # rubocop:disable Layout/LineLength
-                               [Campaign.find_by(id: @initial_campaign_params[:initial_campaign_id])]
-                               # rubocop:enable Layout/LineLength
+                               [Campaign.find_by(
+                                 id: @initial_campaign_params[:initial_campaign_id]
+                               )]
                              else
-                               [Campaign.default_campaign]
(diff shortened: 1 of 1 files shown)


lib/course_creation_manager.rb
```diff
@@ -99,11 +99,12 @@ def set_initial_campaign
...
-                               # rubocop:disable Layout/LineLength
-                               [Campaign.find_by(id: @initial_campaign_params[:initial_campaign_id])]
-                               # rubocop:enable Layout/LineLength
+                               [Campaign.find_by(
+                                 id: @initial_campaign_params[:initial_campaign_id]
+                               )]
                              else
-                               [Campaign.default_campaign]
+                               default_campaign = Campaign.default_campaign
+                               default_campaign ? [default_campaign] : []
```
ragesoss, line 107: Maybe this can be more concise, like `[Campaign.default_camaign].compact`

ragesoss:
This looks fine to me once the linting error is fixed.

ragesoss:
Yeah, that was a linting error that I missed before merging another PR yesterday.

PR #5682, opened 2024-02-28
[Outreachy Round 27] Stop storing liftwing features for non-wikidata wikis.

gabina:
## What this PR does

Before this PR, wikis supported by liftwing and reference-counter API had both API responses stored in the `features`/`features_previous` fields.
After this PR, only wikidata wikis (just supported by the liftwing API) have liftwing features stored in the `features`/`features_previous` fields. Other non-wikidata wikis, like `fr.wikipedia` or `es.wiktionary ` will only store reference-counter response in their `features`/`features_previous` fields.

**Context on why we're changing this**
The revision score importing process is conducted through automatic jobs as part of the course updates. The RevisionScoreImporter class is responsible for selecting "unscored" revisions and querying the API to populate the revision score fields (such as features, wp10, etc.). It determines if a revision is "unscored" by checking if the features or features_previous fields are nil.
Based on my understanding, previously, if the process of querying the LiftWing API failed for any reason, the features field remained nil. Consequently, the RevisionScoreImporter would attempt to populate that field during the next run, as the revision would still be considered "unscored".
Now, with the introduction of two APIs (LiftWing and reference-counter), and storing values from both in the features field, this behavior has changed. For instance, if a LiftWing API request fails unexpectedly, the features field will not be nil because it will contain the response from the reference-counter API. As a result, during the subsequent course update run, the revision score importer will not attempt to query the LiftWing API again to complete the features field.
Although we have implemented a retry strategy when querying APIs, prolonged downtime of one API could lead to many revisions remaining without complete data.

Liftwing features keeps being stored in the `features`/`features_previous` fields for wikidata wikis because they're not supported by the new reference-counter API, so references have to be calculated through liftwing features. However, I'm not sure if wikidata uses a completely different approach to count references. In that case, maybe wikidata doesn't need features at all.

### DIFF
--- a/app/controllers/revision_feedback_controller.rb
+++ b/app/controllers/revision_feedback_controller.rb
@@ -1,16 +1,16 @@
 # frozen_string_literal: true
 
 require_dependency "#{Rails.root}/lib/revision_feedback_service"
-require_dependency "#{Rails.root}/lib/importers/revision_score_importer"
+require_dependency "#{Rails.root}/lib/lift_wing_api"
 
 class RevisionFeedbackController < ApplicationController
   def index
     set_latest_revision_id
     return if @rev_id.nil?
-    revision_data = RevisionScoreImporter.new.fetch_data_for_revision_id(@rev_id)
-    @feedback = RevisionFeedbackService.new(revision_data[:features]).feedback
+    revision_data = LiftWingApi.new(@wiki).get_revision_data([@rev_id])[@rev_id.to_s]
+    @feedback = RevisionFeedbackService.new(revision_data['features']).feedback
     @user_feedback = Assignment.find(params['assignment_id']).assignment_suggestions
-    @rating = revision_data[:rating]
+    @rating = revision_data['prediction']
   end
 
   private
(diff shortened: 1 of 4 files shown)


app/controllers/revision_feedback_controller.rb
```diff
@@ -1,16 +1,16 @@
...
   def index
     set_latest_revision_id
     return if @rev_id.nil?
-    revision_data = RevisionScoreImporter.new.fetch_data_for_revision_id(@rev_id)
-    @feedback = RevisionFeedbackService.new(revision_data[:features]).feedback
+    revision_data = LiftWingApi.new(@wiki).get_revision_data([@rev_id])[@rev_id.to_s]
+    @feedback = RevisionFeedbackService.new(revision_data['features']).feedback
     @user_feedback = Assignment.find(params['assignment_id']).assignment_suggestions
-    @rating = revision_data[:rating]
+    @rating = revision_data['prediction']
```
gabina, line 13: I think this is the easiest way to fix that for now. Instead of using the `RevisionScoreImporter`, it uses the `LiftWingApi` directly. This is because the `RevisionScoreImporter` won't return the LiftWing features or the predictions for `en.wikipedia` (and this controller ony works for `en.wikipedia`).

I can still create an issue to move this to the frontend if we think that's better.

PR #5686, opened 2024-02-28
Fixed Broken Layout of Article Viewer User Legend

### LINKED ISSUE #5681: Layout is broken for Article Viewer user legend
In Article Viewer, the user legend at the bottom has a broken layout. (I'm not sure when this regression appeared, but you can see how it's supposed to be in older screenshots and videos, such as https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5249 )

Here's how it currently looks after attribution data is loaded:

user legend

There is extra highlighted space to right of the username, and it's floating center instead of being aligned to the left.

It's even worse while attribution data is still loading:
user legend loading

The 'loading' message and animations should all be on one line.

PR4NJ41:
## What this PR does
Fixes Broken Layout of Article Viewer User Legend.
Also fixes location of Icon.

Fixes #5681 

## Preview
Before:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/96368921/7e15dea1-a5f8-40bf-bbb2-92a969a2f01a

After:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/96368921/5e8de26e-5ab5-48d8-9ed9-a7dbd14d7b30

### DIFF
--- a/app/assets/stylesheets/modules/_article_viewer.styl
+++ b/app/assets/stylesheets/modules/_article_viewer.styl
@@ -36,6 +36,10 @@
   position: relative
   color: white
 
+.tooltip-indicator-article-viewer
+  background: url('../images/circle_q_mark_icons.svg') center no-repeat
+  padding: 8.5px;
+
 
 .article-alert
   align-items flex-start
@@ -76,13 +80,15 @@
 .user-legend-wrap
   width 90%
   background white
-
-.user-legend
+  
+.user-legend-article-viewer
   margin 10px
   border-radius 15px
-  padding 2px 0px 2px 0px
   display flex
   align-items center
+
+.user-legend-name
+  padding 0px 0px 0px 10px
 
 .user-legend-hover:hover
   background-color rgba(0, 0, 0, 0.2)
--- a/app/assets/stylesheets/modules/_article_viewer.styl
+++ b/app/assets/stylesheets/modules/_article_viewer.styl
@@ -81,7 +81,7 @@
   width 90%
   background white
   
-.user-legend-article-viewer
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/modules/_article_viewer.styl
```diff
@@ -76,14 +80,16 @@
 .user-legend-wrap
   width 90%
   background white
-
-.user-legend
+  
+.user-legend-article-viewer
```
ragesoss, line 84: Is renaming this necessary? It seems like a pretty wordy class name.

ragesoss: Okay. Maybe instead, make both of them more specific, for example by making it apply only to child components of `.article-viewer`? Or call it `article-viewer-legend` instead?

Ideally, we don't want any CSS in specific modules like this from affecting other modules, so adding more top-level rules is dangerous. I guess it was the recent change to the diff_viewer version that caused this problem?

ragesoss:
Yeah, having two top-level user-legend classes was the error. They should either be nested so they only apply to their respective components, or both of them should be renamed more specifically, like `article-viewer-legend` and `diff-viewer-legend`.

ragesoss:
> @ragesoss should i also update `article_viewer_spec.rb` accordingly ?

yes.

ragesoss:
It looks like that spec will definitely be coupled to the change of CSS class name, so even if it has failed randomly on master, you'll probably need to change the spec for this PR. Run it locally and confirm that you have it passing, and then we'll have a clearer idea of whether there is another problem as well.

PR #5687, opened 2024-02-29
Fixes date a11y issues

### LINKED ISSUE #616: calendar is not accessible via screenreader
There are a couple of a11y issues with the calendar:
- Dates do not behave as buttons for a screen reader — the `role="button"` attribute should solve that
- Dates do not convey their state via screen reader — the `aria-state="[true/false/mixed]"` attribute will probably be the right way to solve this, along with a screen-reader-accessible explanation that "mixed" state means an unselected but not blacked out day (and false would indicate a blacked out day).

fillingtothemomo:
## What this PR does
Fixes #616 
Adds the required roles to the calendar dates
# Before
image

# After 
image

### DIFF
--- a/app/views/styleguide/_form_controls.html.haml
+++ b/app/views/styleguide/_form_controls.html.haml
@@ -70,67 +70,67 @@
 
 %h5 Calendar
 #calendar
-.DayPicker
-  .DayPicker-wrapper
-    .DayPicker-NavBar
-      %span.DayPicker-NavButton.DayPicker-NavButton--prev
-      %span.DayPicker-NavButton.DayPicker-NavButton--next
-    .DayPicker-Months
-      .DayPicker-Month{:role => "grid"}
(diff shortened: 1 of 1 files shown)


ragesoss:
Note that this only affects the styleguide. Our actual calendar in the user-facing UI is rendered in React.

fillingtothemomo:
> Note that this only affects the styleguide. Our actual calendar in the user-facing UI is rendered in React.

Do I need to make changes to the react ui as well for a11y issue?
Also the ruby lint check fail is not related to this PR

ragesoss:
I can merge this one separately, it just won't close the issue. Can you include a before/after screenshot of the calendar on the styleguide?

fillingtothemomo:
> I can merge this one separately, it just won't close the issue. Can you include a before/after screenshot of the calendar on the styleguide?

I updated the screenshots

PR #5688, opened 2024-02-29
Implemented Native confirm component instead of JavaScript confirm instance

### LINKED ISSUE #1038: Replace all javascript confirm and alert instances with Confirm and Notification components

PR4NJ41:
## What this PR does
- Implemented native confirm components instead of JavaScript confirm instances for assignments deletion.
- resolves #1038

## Screenshots
Before:
Before

After:
After

### DIFF
--- a/app/assets/javascripts/components/articles/available_article.jsx
+++ b/app/assets/javascripts/components/articles/available_article.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch } from 'react-redux';
-
+import { initiateConfirm } from '~/app/assets/javascripts/actions/confirm_actions';
 import CourseUtils from '../../utils/course_utils.js';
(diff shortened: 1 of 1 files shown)


ragesoss:
Thanks! Does this feature get exercised in the test suite?

ragesoss:
You should change that spec and make sure it passes locally.

Issue #5685, opened 2024-02-28
In Edit Details form, all inputs should have accessible labels


ragesoss:
In the Edit Details view, some of the inputs, such as School, Title and Term, work smoothly with screen readers. I think this is because they use a `<label>` tag with a `for` attribute connecting them to the input.

Some of the others, including each of the selector-based inputs, do not work as well with screen readers, possibly because the labels are implemented in simple `<span>` tags instead. Improve the input component(s) so that they use accessible labels.

brendagold:
I want to pickup this issue. I am finding it hard to locate it in the codebase. Can you direct me, please?

PR #5692, opened 2024-03-01
Fixed screen reader accessibility in edit details view

atulyadav745:
## What this PR does
This PR fixes the issue #5685 
In Edit Details view some inputs including selector-based inputs,which do not work as well with screen readers, have been added label for each which makes it accessible to screen reader

@ragesoss Please review the PR and provide your valuable feedback if any changes are required in this.
Thank you


ragesoss:
Have you tested this? It doesn't look like it would work to me.

Issue #5700, opened 2024-03-07
'Email Instructors' feature should have 'BCC to Salesforce' option just like Ticket replies


ragesoss:
We recently added an option to email instructors from the course page. It would be very useful to also add 'BCC to Salesforce' to this by default, as that will help us keep track of communication we do with a given instructor more easily.

(It should work the same way as in the Ticket system, where 'BBC to Salesforce' is checked by default in the UI but can be unchecked to disable it for an individual message.)

psankhe28:
@ragesoss Can i work on this issue ?

ragesoss:
@psankhe28 sure. look at the way this works on the Ticket system (`/tickets/dashboard`) for the example to follow.

psankhe28:
Can you please confirm if the file where the changes need to be done is `/app/mailers/instructor_notification_mailer.rb`. I am not sure if I have found the right file.

ragesoss:
@psankhe28 yes, that's one place that will need to be changed. it should be optional (enabled by default), though, just like on tickets.

psankhe28:
> @psankhe28 yes, that's one place that will need to be changed. it should be optional (enabled by default), though, just like on tickets.

Okay, thanks!

psankhe28:
Hello @ragesoss, I apologize for my recent inactivity as I had been occupied with mid-semester exams. Could you please guide me on accessing a course page? I've encountered an issue where I created a program using one account but couldn't find the option to join it using another account, particularly in sending emails. Can you assist me with this matter?

psankhe28:
Thanks, I will try this.

psankhe28:
I've created one user as an instructor who has created a course, and another user as an admin. However, I'm unable to locate any options for sending emails. I understand this might be a lot to ask, but could you please assist me? Thank you!

ragesoss:
@psankhe28 sure. This feature is only enabled in Wiki Education mode, and I think you are not using that mode. Change the `wiki_education` value in `application.yml` and see if that gives you access to the feature.

psankhe28:
It worked, thanks!

prathamVaidya:
Looks like no one is active. I'll take the task.

Formasitchijoh:
@ragesoss can i take up this issue,

ragesoss:
I lost track of this and I can't tell why the build failed. I recommend starting from @prathamVaidya's open PR and seeing if it is in working shape.

[...]

PR #5708, opened 2024-03-13
Added bcc to salesforce

psankhe28:
## What this PR does
It adds BCC to salesforce by default
Fixes issue #5700


ragesoss:
This does not add any way to enable/disable the BCC. It needs to include a UI as well.

psankhe28:
> This does not add any way to enable/disable the BCC. It needs to include a UI as well.

Apologies, I wasn't aware that the UI needed to be included. I will make sure to add the UI component.

psankhe28:
By default
image

If unchecked
image

psankhe28:
@ragesoss , please review this.

app/mailers/instructor_notification_mailer.rb
```diff
@@ -15,6 +15,10 @@ def email(alert)
     unless @alert.sender_email.nil?
       params[:bcc] = @alert.sender_email # sender_email gives user email of the sender
     end
+    if bcc_to_salesforce && !@alert.sender_email.nil?
+      params[:bcc] ||= []
+      params[:bcc] << ENV['SALESFORCE_BCC_EMAIL']
```
ragesoss, line 20: I suggest mirroring the implementation in ticket_notification_mailer.

ragesoss (commented):
I think the layout has room for improvement. It would be better to put the BCC checkbox at the lower right, like with the Ticket reply UI:

BCC

app/mailers/instructor_notification_mailer.rb
psankhe28: I have made the code changes, can you please have a look at it?

PR #5709, opened 2024-03-17
Make sitenotice configurable

### LINKED ISSUE #5699: Make sitenotice configurable from Settings page
Currently, we can enable a sitenotice banner via adding a `sitenotice` entry to application.yml and then restarting Rails. If there's a syntax error (like unbalanced quote marks) then this brings the site down.

It would be much nicer if the sitenotice could be handled completely from the frontend, via the Settings page (/settings), and implemented in a way such that it's never going to bring the site down if there's mistake in the input. It needs to support HTML, as we often include a link in a sitenotice message. It should also be done in a way that uses efficient caching and cache breaking, likely following the pattern we are now using for the stats shown in on the logged-out home page in wiki_education mode via HomeController and updated via Settings.

om-chauhan1:
## What this PR does
Fixes #5699 
This PR made sitenotice configurable from frontend via Settings Page.
Implemented in such a way that :
1. It won't bring the site down if there is any error in input, 
2. Support HTML 
3. Efficient use of caching is done to optimize performance and reduce back-end calls. 

### Before:

Before

### After:

[Screencast from 2024-03-17 15-39-31.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/54506971-e854-4296-89a0-7e6f01d32b89)


app/controllers/settings_controller.rb
```diff
@@ -120,8 +121,43 @@ def update_impact_stats
...
+      ENV['sitenotice'] = ''
+      message = 'Site Notice Disabled Successfully.'
+    end
+    render json: { message: }, status: :ok
+  end
+
   private
 
+  def set_site_notice
+    ENV['sitenotice'] = current_site_notice['message']
```
ragesoss, line 151: Will this strategy retain the site notice across application or server restarts? I was expecting this to require some minor changes for how the message gets served to the frontend, such as going through the `Features` presenter in `_flash.htmlhaml` instead of using the ENV.

ragesoss (commented):
Looks like the existing feature specs related to the sitenotice are failing.

om-chauhan1:
yes sir, working on this

app/controllers/settings_controller.rb
om-chauhan1: you are right it won't retain site notice when server restarts. Now i have changed the approach using Features presenter instead of ENV to provide sitenotice from cache/database to _flash.htmlhaml

PR #5711, opened 2024-03-19
Improved layout of "subject specific guides" block in Timeline

### LINKED ISSUE #5707: Improve layout of "subject specific guides" block in Timeline

guides links

PR4NJ41:
## What this PR does
Improved the layout of "subject specific guides" block in Timeline.

fixes #5707 

## Screenshots
Before:
Before

After:
After

### DIFF
--- a/app/assets/javascripts/components/timeline/timeline.jsx
+++ b/app/assets/javascripts/components/timeline/timeline.jsx
@@ -181,6 +181,21 @@ const Timeline = createReactClass({
   },
 
   render() {
+    const blocks = document.querySelectorAll('.block__block-content');
+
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -181,6 +181,21 @@ const Timeline = createReactClass({
   },
 
   render() {
+    const blocks = document.querySelectorAll('.block__block-content');
```
ragesoss, line 184: Yeah, I think I'd prefer the other method that just applies going forward, rather than a using JS to insert the styling.

PR #5712, opened 2024-03-19
Improved the layout of "subject specific guides" block in Timeline.

### LINKED ISSUE #5707: Improve layout of "subject specific guides" block in Timeline

PR4NJ41:
## What this PR does
Improved the layout of "subject specific guides" block in Timeline.

This is the other method which adds div with classname on block creation. This is only added to the handout section. Hence this will be added to only newly created courses.

fixes #5707 

### Before:
Before
Screenshot 2024-03-21 at 1 35 51 AM
Screenshot 2024-03-20 at 4 12 04 AM

### After:
After
Screenshot 2024-03-21 at 1 35 34 AM
Screenshot 2024-03-20 at 4 12 12 AM

### DIFF
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -87,6 +87,9 @@
 
 .block__block-content
   color $text_dk
+  display flex
+  flex-wrap wrap
+
 
 .block.block--orderable
   display flex
@@ -533,6 +536,7 @@ a.timeline-exercise
     float right
 
 a.handout-link
+  margin-right 8px
   color #6a6a6a
   border solid 1px #ddd
   border-radius 3px
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -540,3 +540,10 @@ a.handout-link
   padding 5px
   &:hover, &:focus
     border solid 1px $brand_primary
+
+.handout-link-container
+  display flex
+  flex-wrap wrap
+  margin-top 10px
+  > p
+    margin-right 8px
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -540,10 +540,3 @@ a.handout-link
   padding 5px
   &:hover, &:focus
     border solid 1px $brand_primary
-
-.handout-link-container
-  display flex
-  flex-wrap wrap
-  margin-top 10px
-  > p
-    margin-right 8px
--- a/app/assets/javascripts/components/common/text_area_input.jsx
+++ b/app/assets/javascripts/components/common/text_area_input.jsx
@@ -90,7 +90,7 @@ const TextAreaInput = createReactClass({
               plugins: 'lists link code',
               toolbar: [
                 'undo redo | styleselect | bold italic',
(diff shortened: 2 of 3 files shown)


PR4NJ41:
@ragesoss Please review this.

app/assets/stylesheets/modules/_timeline.styl
```diff
@@ -539,4 +539,11 @@ a.handout-link
   text-decoration none
   padding 5px
   &:hover, &:focus
-    border solid 1px $brand_primary
\ No newline at end of file
+    border solid 1px $brand_primary
+
+.handout-link-container
+  display flex
+  flex-wrap wrap
```
ragesoss, line 546: I think this would work fine without the new div if you add the `display` and `flex-wrap` to `.block__block-content` and then add `margin-right` to `a.handout-link`.

I don't think adding those things to the block content is likely to cause any problems.

The margin-top is a little nicer than the current layout, but it mainly matters for a user who has access to the [Edit] button of the block. When that button is not there, extra space doesn't make much difference anyway, I think. And either way, it's a minor thing that I think we can leave off.

PR4NJ41: @ragesoss I already tried that before these two solutions but .block__block-content is being used in other places also like week 1. when i add flex in .block__block-content it sets the direction to row so other things gets changed. I will attach a Screenshot later. 

Yes, We can remove margin-top

PR4NJ41:
@ragesoss this might create unwanted styling changes. Also Some styling like `align center` of the edit menu won't work.

Before:
Screenshot 2024-03-20 at 4 12 04 AM

After:
Screenshot 2024-03-20 at 4 12 12 AM

app/assets/stylesheets/modules/_timeline.styl
PR4NJ41: @ragesoss I already tried that before these two solutions but .block__block-content is being used in other places also like week 1. when i add flex in .block__block-content it sets the direction to row so other things might gets changed. I will attach a Screenshot later.

Yes, We can remove margin-top

ragesoss:
Hmm... I actually like the bullet styling better. I don't think there is much if any custom content that uses the `center` formatting. I'm okay with some minor changes, as long as it doesn't break anything in a noticeable way.

PR4NJ41:
> I don't think there is much if any custom content that uses the `center` formatting.

By `center` I mean the formatting as shown in the video.

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/96368921/ad724f98-aff5-4374-8de1-b7ffc99a77ce

PR4NJ41:
@ragesoss I have made the suggested changes. Please review. Thanks!

ragesoss:
> By `center` I mean the formatting as shown in the video.

That's what I thought you meant. I think this is fine. I think you can remove the `aligncenter` center option by editing the tinyMCE config in text_area_input.jsx.

PR4NJ41:
@ragesoss should I remove `alignRight` also as it won't be useful just like `aligncenter`

PR4NJ41:
@ragesoss  I have made the changes.

PR4NJ41:
@ragesoss I have updates the screenshots. Thanks!

PR4NJ41:
@ragesoss I have updated the screenshots. Thanks!

PR #5718, opened 2024-03-22
feat: validate on-wiki edits in training module

prathamVaidya:
## What this PR does

Fix 3rd task of #3500
 
**Issue:** For training modules that have on-wiki edits, check for edits to the target pages and do not register modules as complete unless such edits have been made. Shouldn't allow progress past the slide that launches them until such edits are made.

## Solution: 

Added functionality to check if a wiki edit button exists in slide content with a valid wiki URL then enable Wiki Edit Assignment mode which checks that if a user is logged in then do not allow them to move to the next slide until they have made an edit on the wiki page. Uses the MediaWiki API to query the last revision author.

These are the constraints:

- user should be logged in
- user should make the wiki edit with the same Wikipedia account.
- last edit on the page should be from the user

## Screenshots

image

Check Demo :

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/61202986/0fae78bc-71f5-4065-8546-b2643b8cfc4e

Additionally, to detect the wiki edit button in the slide, I have added the query selector `a.wiki-edit-assignment`. This is to ensure that accidentally some other tag pointing to the wiki page is not used by this feature. This class name should be updated on the wiki page of the training module. 

The current regex to validate a wiki URL only allows the `en.wikipedia.org` domain. Let me know if I should remove it.


prathamVaidya:
@ragesoss please review

ragesoss:
This is pretty different from what I had in mind. The problem the issue is trying to solve is that we have Exercise modules where the user is expected to read through the instructions, then do some on-wiki edits to actually perform the exercise, then use a 'Mark Complete' button the Dashboard top indicate that they did it... but many students click the 'Mark Complete' button without having actually followed the instructions and done the wiki edits. So we want to validate the on-wiki edits on the backend, and only allow associated exercises to be marked complete if edits to the relevant pages were made.

prathamVaidya:
oh got it. 

So as per my understanding this is an example off exercise module

image

The assignment has multiple stages [ "not_yet_started", "in_progress", "ready_for_review", "ready_for_mainspace", "assignment_completed" ] and the task is to validate the edits associated with every status that has been made on the backend when user clicks "Mark Complete".

So in these 4 stages, is this how the validations should be done? 

Bibliography : no validation
Sandbox : check if edits in Sandbox link 
Expand your Draft : check if edits in Sandbox link
Move your work (ready_for_mainspace): check if edits in main article

Also, should the logic only check last edit or is there any other criteria?

ragesoss:
I think this is an issue I need to rethink more deeply and write up a fuller description for at some point.

PR #5721, opened 2024-03-23
Filter workflow steps depending on file changes

### LINKED ISSUE #5720: Optimizing Workflow and Rspec Test Suite
### What is happening?

The `ci.yml` workflow file for Continuous Integration (CI) serves the purpose of executing tests and identifying linting errors. Currently, this workflow typically consumes approximately 35 minutes to execute. An average of 30 minutes is used by the RSpec test suite. We can perform multiple optimizations on the workflow and rspec test suite to reduce this time and improve the overall development process.

### 1. Separate workflow steps depending on file changes

There is no need to run RSpec tests if only a javascript file or some `.md` file has changed.  We can filter the steps based on file 
changes using [dorny/paths-filter@v3](https://github.com/dorny/paths-filter) workflow action. I have opened a PR for this.

### 2. Running Steps Concurrently

While executing parallel steps is not supported by Github, a custom script can be written to run some steps parallel to reduce workflow time. [Refer to this Discussion](https://github.com/orgs/community/discussions/26291#discussioncomment-3251245)

## Possible RSpec Optimizations:

These are time-consuming but can be slowly started

- Separate Unit Tests and Integration Tests. Write logic to run UTs on every push, pull_request, and Integration Tests less often.
- Use Mock instead of db call
- try reducing time for the top slowest tests.

prathamVaidya:
## What this PR does

Fixes #5720 

This is an optimization on the CI workflow `ci.yml`. By filtering some steps based on file changes we can reduce the workflow execution time.

**For Example:** If only a javascript file is changed then there is no need to run Ruby tests.

## Changes

- I have updated the ci.yml to use [dorny/paths-filter@v3](https://github.com/dorny/paths-filter) github action.
- Created file-filters.yml that will have all the filter tags that can be used to check if a step should run or not.

Current `file-filters.yml` contents are:

```
backend:
      - '!app/assets/**'
      - '!app/views/**'
      - '!public/**'
      - '!training_content/**'
      - '!.github/**'

```

Here a backend filter is defined. The codebase has ruby files everywhere so I have defined the backend is everything other than frontend code, `.github` and `training_content`. The more definite this list will be the better it is. Let me know any other paths that needed to be defined here. (One can be all the markdowns `*.md`)

Ruby tests that took 30 mins will be skipped on all these folders. I have taken inspiration from how [sentry repository used this](https://github.com/getsentry/sentry/blob/2ebe01feab863d89aa7564e6d243b6d80c230ddc/.github/workflows/backend.yml#L36).

I have tested the workflow on my fork [at this branch](https://github.com/prathamVaidya/WikiEduDashboard/tree/optimize-workflow).

Here are some examples of successful workflows:

On Push -> https://github.com/prathamVaidya/WikiEduDashboard/actions/runs/8403231749
On Pull Request -> https://github.com/prathamVaidya/WikiEduDashboard/actions/runs/8403275731 (It failed because it ran flaky RSpecs which it should have because there was a modified ruby file)

@ragesoss Let me know if there is a breaking change associated with this feature. Especially on my assumption that the files in `training_content`, `app/assets` do not affect the ruby tests.

### DIFF
--- a/.github/file-filters.yml
+++ b/.github/file-filters.yml
@@ -1,0 +1,6 @@
+backend:
+      - '!app/assets/**'
+      - '!app/views/**'
+      - '!public/**'
+      - '!training_content/**'
+      - '!.github/**'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,38 +6,45 @@ on: ['push', 'pull_request', 'workflow_dispatch']
 jobs:
   build:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-latest
 
     steps:
       # checks-out your repository under $GITHUB_WORKSPACE , so that workflow can access it.
       - name: checkout
         uses: actions/checkout@v2
-      
+
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+              backend:
+                  - '!(app/assets|app/views|public|training_content)/**'
+
       - name: install Pandoc
         uses: r-lib/actions/setup-pandoc@v2
         with:
           pandoc-version: '2.9.2.1'
-      
+
       - name: setup Redis
         uses: supercharge/redis-github-action@1.4.0
-      
+
       # Setting up MariaDB based upon 'config/database.example.yml'
       - name: setup MariaDB server
-        uses: getong/mariadb-action@v1.1  
+        uses: getong/mariadb-action@v1.1
         with:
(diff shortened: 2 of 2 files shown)


ragesoss:
Many of the ruby tests are browser-based feature specs, which test against both the JS frontend and the Ruby backend. I prefer to just run the entire test suite for every change, except perhaps changes that are exclusively to docs.

ragesoss:
Let's close it, thanks.

PR #5723, opened 2024-03-24
Feature 5693 - Training Data in Editors CSV

### LINKED ISSUE #5693: Editors data CSV for a course should include data about training modules completed.
Instructors would find it useful to be able to download data about which students have completed which trainings. I think the easiest way to fit this in would be to include some data about completed training modules in the 'Editors' CSV.

prathamVaidya:
## What this PR does

Fixes #5693 

It populates the training data in the editor's CSV file.

Added three fields in the last:

- completed_training_modules_count
- assigned_training_modules_count
- completed_training_modules

image

### DIFF
--- a/lib/analytics/course_students_csv_builder.rb
+++ b/lib/analytics/course_students_csv_builder.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_dependency "#{Rails.root}/lib/course_training_progress_manager"
 require 'csv'
 
 class CourseStudentsCsvBuilder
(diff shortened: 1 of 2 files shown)


prathamVaidya:
@ragesoss I am getting "Class has too many lines" error. I am not sure how can I reduce lines in this class as nothing seems to be inheritable and no methods can be separated further.

image

ragesoss:
You can add an exception for this file in `.rubocop.yml`, or you could find a way to extract some of the code in that file to another file. I think most of the code that doesn't have to do directly with the user and their progress (like getting the array of exercise modules vs non-exercise modules) would make sense to live in its own file, perhaps implemented via delegation in `course.rb` similar to how that file uses `:trained_students_manager`.

ragesoss:
Looks like a failing test is connected to your change.

prathamVaidya:
> Looks like a failing test is connected to your change.

Fixed it

prathamVaidya:
Finally, No errors

PR #5735, opened 2024-03-29
Added new eslint rule to disallow string literals

prathamVaidya:
## What this PR does

This PR added a new ESLint rule to disallow string literals. This will help users to prefer I18n use over string literals to promote localization.

https://github.com/edvardchen/eslint-plugin-i18next

## Screenshots

image

There are two modes for it:

- `jsx-only` : validates the JSX attributes as well
- `all` : validates all literal strings

Currently its set to `jsx-only`. Let me know if you want to change.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -80,7 +80,7 @@
     "import/no-useless-path-segments": 0,
     "react/no-this-in-sfc": 0,
     "react/jsx-props-no-multi-spaces": 0,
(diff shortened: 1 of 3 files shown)


prathamVaidya:
There are 306 places where string literals are directly used. I have updated the rule to only throw a warning. Fixing these can be a `good-first-issue` for beginners.

image

PR #5741, opened 2024-03-31
Improved Article Finder Searchbar with added Autocomplete

### LINKED ISSUE #649: Article title fields should autocomplete
Do a prefix (or better) search against the wiki API to populate autocomplete.

prathamVaidya:
## What this PR does

Closes #649 

This PR improves the Article Finder Searchbar UI and adds autocomplete functionality to the finder.

## Screenshots

Before:

image

After:

image

[screen-recording-2024-03-31-19 56.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/61202986/98ab5f70-ffbb-43bf-85e6-4e5f4340f294)

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
@@ -56,6 +56,7 @@ function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
     <div className={searchClass}>
       <input
         type="text"
+        id="article-searchbar"
         placeholder={I18n.t('article_finder.search_placeholder')}
         onChange={inputChangeHandler}
         value={value}
--- a/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
@@ -40,7 +40,7 @@ function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
 
   const onKeyDownHandler = (e) => {
     // Search on Enter
-    if (e.keyCode === 13) {
+    if (e.key === 'Enter') {
       searchHandler();
     }
   };
@@ -72,8 +72,8 @@ function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
       </button>
       <div className="autocomplete">
         {
-          suggestions.map((sug, index) => {
-            return <div key={`suggestion-${index}`} className="autocomplete-item" onClick={() => autoCompleteClickHandler(sug)}> {sug} </div>;
+          suggestions.map((sug) => {
+            return <div key={sug} className="autocomplete-item" onClick={() => autoCompleteClickHandler(sug)}> {sug} </div>;
           })
         }
       </div>
--- a/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
@@ -1,0 +1,83 @@
+import React, { useCallback, useState } from 'react';
+import { debounce } from 'lodash';
+import { fetchArticleAutocompleteResults } from '../../utils/article_finder_utils';
+
+// Controls Search bar and autocomplete functionality
+
+function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/utils/article_finder_utils.js
```diff
@@ -90,3 +90,30 @@ export const extractClassGrade = (pageAssessments) => {
...
+  };
+};
+
+
+const mediawikiApiBase = (language, project) => {
+  if (project === 'wikidata') {
+    return `https://${project}.org/w/api.php?action=query&format=json&origin=*`;
+  }
+  return `https://${language}.${project}.org/w/api.php?action=query&format=json&origin=*`;
+};
```
TheTrio, line 111: There already exists a utility for this, best to use it https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/15788d822067e2ef9da6c09a34f67a4946e056f2/app/assets/javascripts/utils/wiki_utils.js#L2-L8

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
```diff
@@ -0,0 +1,84 @@
...
+        isAutocompleteLoading && <div className="loader"><div className="loading__spinner"/></div>
+      }
+
+      <button onClick={searchHandler} disabled={disabled}>
+        {I18n.t('article_finder.search')}
+      </button>
+      <div className="autocomplete">
+        {
+          suggestions.map((sug, index) => {
+            return <div key={`suggestion-${index}`} className="autocomplete-item" onClick={() => autoCompleteClickHandler(sug)}> {sug} </div>;
```
TheTrio, line 76: Best to avoid using the index for the key, see https://react.dev/learn/rendering-lists#why-does-react-need-keys

app/assets/javascripts/utils/article_finder_utils.js
```diff
@@ -90,3 +90,30 @@ export const extractClassGrade = (pageAssessments) => {
...
+
+export const fetchArticleAutocompleteResults = async (keyword, wiki) => {
+  const query = keywordAutocompleteGenerator(keyword);
+
+  return queryUrl(mediawikiApiBase(wiki.language, wiki.project), query).then((data) => {
+    return data.query.search.map(item => item.title);
+  });
+};
```
TheTrio, line 119: Why not use `await` if the function is marked as `async`?

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const searchHandler = () => {
+    if (disabled || value.trim() === '') return;
+    setSuggestions([]);
+    onSearch(value);
+  };
+
+  const onKeyDownHandler = (e) => {
+    // Search on Enter
+    if (e.keyCode === 13) {
```
TheTrio, line 43: Using the [key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) property might help the readability.

app/assets/javascripts/utils/article_finder_utils.js
prathamVaidya: Using await and return is redundant because then it will resolve the promise and again convert it into a promise because an async function returns a promise anyways.
See:
https://stackoverflow.com/a/44806250

app/assets/javascripts/utils/article_finder_utils.js
prathamVaidya: Thanks! I'll use that

app/assets/javascripts/utils/article_finder_utils.js
TheTrio: I meant in place of the `then`, the return is fine.

prathamVaidya: If you meant the `then` I used on `queryUrl` then Yes I do have used await when I called this method but I have used `then` trying to abstract the API logic and its returned structure. If I remove it then every where I use it I need to know the underlying structure of the response JSON.

TheTrio:
Haven't tested out the PR locally, but the code looks good to me.

– Shashwat  (He/him)

ragesoss:
Nice, this works pretty smoothly!

ragesoss:
thanks much for the code review @TheTrio !

PR #5753, opened 2024-04-05
Converted upload_viewer.jsx to functional component and used i18n for string literals

om-chauhan1:
## What this PR does
This PR converted the upload_viewer.jsx from class to functional component and used i18n for string literals.

## Videos

Before:

Before

After:

After

### DIFF
--- a/app/assets/javascripts/components/uploads/upload_viewer.jsx
+++ b/app/assets/javascripts/components/uploads/upload_viewer.jsx
@@ -28,7 +28,7 @@ const UploadViewer = ({ closeUploadViewer, upload, imageFile }) => {
     }
   }, [uploadMetadata, upload.id, loadingViews]);
 
-  const handleGetFileViews = files => {
+  const handleGetFileViews = (files) => {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/uploads/upload_viewer.jsx
```diff
@@ -1,181 +1,160 @@
...
+          <tbody>
+            {usageTableElements}
+          </tbody>
+        </table>
+      </div>
+    );
+  }
+  let categoriesList = [];
+  let categories;
+  forEach(get(metadata, 'categories', []), (category) => {
```
AmitJoki, line 93: I think we can slowly migrate away from using lodash functions for things which we can natively do in JS.

```
forEach(get(metadata, 'categories', []), (category) => {
```

would be just

```
(metadata.categories ?? []).forEach((category) => {
```

om-chauhan1: sure

PR #5762, opened 2024-04-10
Fix slow + flaky namespaces_stat spec

### LINKED ISSUE #5655: `namespace_stats_spec.rb` feature spec is flaky
Here's a recent CI failure:

```
2) Namespace-specific stats generates and renders stats for Cookbook namespace on en.wikibooks.org
     Got 0 failures and 2 other errors:

     2.1) Failure/Error: visit "/courses/#{course.slug}/manual_update"

          Net::ReadTimeout:
            Net::ReadTimeout with #<Socket:(closed)>
          # ./spec/features/namespace_stats_spec.rb:29:in `block (3 levels) in <top (required)>'
          # ./spec/features/namespace_stats_spec.rb:28:in `block (2 levels) in <top (required)>'

```

This might be a kind of failure related to the amount of calls to production Wikimedia APIs that get made by the test suite, or it might be a random network failure. Review recent failed CI runs to see how often this failure comes up in feature specs, and whether it happens in the same test often or whether it happens across multiple different tests.

It might make sense to address with some kind of retry behavior across the test suite specifically for network timeouts, or we might need to look for another way to make it more reliable.

### LINKED ISSUE #5660: `namespace_stats_spec.rb` is slow and flaky
The feature spec `namespace_stats_spec.rb` takes too long to run (eg, 39 seconds for me locally, after deleting VCR fixtures) and sometimes fails on CI because of a net timeout error. This is likely because of all the network requests that happen during the `manual_update` that happens at the start of the test. (It runs very quickly if the network requests are already recorded in a VCR fixture from an earlier test run.)

Since this is a frontend test intended to make sure the namespace stats are rendered properly, and we're confident that the backend side of pulling in data and generating records for namespace stats is okay, the best option to fix this test would probably be to generate appropriate namespace data directly during test setup instead of relying on pulling data from production servers during the test via `manual_update`.

cyrillefr:
## What this PR does
This PR aims to close #5660 by 

 - stubbing UpdateCourseStats to avoid real http calls
 - building test course stats to get data
 - removing now useless VCR

I think it might closes #5655 as well

### DIFF
--- a/spec/features/namespace_stats_spec.rb
+++ b/spec/features/namespace_stats_spec.rb
@@ -22,22 +22,23 @@ describe 'Namespace-specific stats', type: :feature, js: true do
     JoinCourse.new(course:, user: user1, role: 0)
     JoinCourse.new(course:, user: user2, role: 0)
     login_as superadmin
+    allow(UpdateCourseStats).to receive(:new)
+    create(:course_stats,
+           stats_hash: { 'en.wikibooks.org-namespace-0':
+    { edited_count: 1, new_count: 2, revision_count: 3,
+      user_count: 4, word_count: 5, reference_count: 6, view_count: 7 },
+   'en.wikibooks.org-namespace-102':
+    { edited_count: 2, new_count: 4, revision_count: 4, user_count: 14,
+      word_count: 15, reference_count: 16, view_count: 17 } }, course_id: course.id)
   end
 
   it 'generates and renders stats for Cookbook namespace on en.wikibooks.org' do
-    pending 'Manual update sometimes takes too long and times out'
+    visit "/courses/#{course.slug}/manual_update"
+    expect(page).to have_content 'en.wikibooks.org - Mainspace'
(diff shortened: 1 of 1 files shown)


cyrillefr:
Hello @ragesoss , 

I have got this Sentry related error:

`(Sentry).capture_message("Non-200 response hitting references counter API", {:extra=>{:content=>{"description"=>"mwapi error: permissiondenied - You don't have permission to vie...e_code=>"en", :project_code=>"wikipedia", :rev_id=>708326238, :status_code=>403}, :level=>"warning"})`

It happens because the code in `lib/reference_counter_api.rb` has been commented due to commit 6f40ec12b4

I guess it is better to make the line 29 `it 'logs the message if response is not 200 OK'` pending and I wait and rebase :)

cyrillefr:
Better now after rebase :)

ragesoss:
Looks good to me!

PR #5774, opened 2024-04-14
added: bcc to salesforce by default

### LINKED ISSUE #5700: 'Email Instructors' feature should have 'BCC to Salesforce' option just like Ticket replies

prathamVaidya:
## What this PR does
Fixes #5700 

- Added salesforce email to the BCC list by default.

## Screenshots

image


ragesoss:
See the clarification I added to the issue description; this needs a UI component as well and the option to disable it.

PR #5775, opened 2024-04-15
Added 'Get Help' button to training modules

### LINKED ISSUE #5486: Add 'Get Help' button to training modules

omChauhanDev:
## What this PR does
Fixes #5486 
This pr mainly aims to show the 'Get Help' button on training modules. The button's purpose is to send alerts to the Wiki Expert for the user's course.

Tasks :

- [x] Refactor the training pages such that the training navbar will render in React 

- [x] For logged-in users, in wiki_education mode, the 'Get Help' button should appear when going through training modules.

- [x] Defined behaviour of this button on the training navbar.


om-chauhan1:
@ragesoss Sir, based on the issue details, this 'Get Help' button on the training navbar should default to sending alerts to the Wiki Expert for the user's course. My question is about clarifying whether "Wiki Expert" refers to the **wikipediaHelpUser**, **programHelpUser**, **dashboardHelpUser,** or any other term I might be missing.

Till now, this button has the course.id from where the user reached the training module.

ragesoss:
@om-chauhan1 it refers to wikipediaHelpUser.

om-chauhan1:
Before :
Before


After :

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/0fc6fb86-ac89-48e6-971c-d72d5d50ebe6)

app/assets/javascripts/training/components/navbar_training.jsx
AmitJoki, line 1: Can be just:

```
import React from 'react';

const NavbarTraining = ({ navBreadcrumb }) => {
  return (
    <div className="container">
      {navBreadCrumb && <ol className="breadcrumbs" dangerouslySetInnerHTML={{ __html: navBreadcrumb }} />}
    </div>
  );
};

export default NavbarTraining;
```

app/assets/javascripts/training/components/training_slide_handler.jsx
```diff
@@ -97,10 +110,15 @@ const TrainingSlideHandler = () => {
   // fetches the initial data and sets the base title
   useEffect(() => {
     setBaseTitle(document.title);
+    setNavBreadcrumb(document.getElementById('react_root').getAttribute('data-breadcrumbs'));
```
AmitJoki, line 113: This code has been repeated 3 times. Maybe turn this into a function call?

om-chauhan1: okay done, thanks

app/assets/javascripts/training/components/navbar_training.jsx
om-chauhan1: sure, done

app/assets/javascripts/components/common/course_navbar.jsx
```diff
@@ -66,7 +66,7 @@ const CourseNavbar = ({ course, location, currentUser, courseLink }) => {
   if (Features.enableGetHelpButton) {
     getHelp = (
       <div className="nav__button" id="get-help-button">
-        <GetHelpButton course={course} currentUser={currentUser} key="get_help" />
+        <GetHelpButton course={course} currentUser={currentUser} key="get_help" courseSlug={courseSlug} onTrainingPage={false}/>
```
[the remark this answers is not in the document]
om-chauhan1: okay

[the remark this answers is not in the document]
om-chauhan1: okay

ragesoss:
Testing this out, I only see part of the expected content in the popover.

get help


Also, I think the color of the button needs to be changed in this context. It does not look good against the green.

ragesoss:
Unfortunately I never got around to reviewing this before it went stale with conflicts.

PR #5782, opened 2024-04-20
Refactored revision_list.jsx from class to functional component

om-chauhan1:
## What this PR does
This PR converted the revision_list.jsx from class to functional component.

## Screenshots

Before:

Before

After:

After


om-chauhan1:
The failing test is passing locally.

PR #5783, opened 2024-04-20
Fix Issue 5700 : Added 'BCC to Salesforce' in Email Instructors

### LINKED ISSUE #5700: 'Email Instructors' feature should have 'BCC to Salesforce' option just like Ticket replies

prathamVaidya:
## What this PR does

Fixes #5700 'Email Instructors' feature should have 'BCC to Salesforce' which is true by default.

- [x] Added BCC To salesforce option.
- [x] Added rspec test for `#notify_instructors` controller function.

## Screenshots

image


app/assets/javascripts/components/overview/notify_instructors_button.jsx
```diff
@@ -50,16 +51,30 @@ const NotifyInstructorsButton = (props) => {
...
-            onChange={(_, value) => {
-              dispatch({ type: ALERT_INSTRUCTOR_UPDATE_MESSAGE, payload: value });
-            }}
-            value_key="notification_message"
-            editable
-            placeholder={I18n.t('course_instructor_notification.write_message_placeholder')}
-          />
+          <div className="form-group">
+            <TextAreaInput
+              id={'notification-message'}
```
AmitJoki, line 56: No need to use `{}` for literals. `id='notification-message'` would suffice.

app/mailers/instructor_notification_mailer.rb
```diff
@@ -1,20 +1,23 @@
...
     @alert = alert
     set_email_parameters
     params = { to: @instructors.pluck(:email),
-               subject: @alert.subject }
+               subject: @alert.subject,
+               bcc: [
+                 @alert.sender_email,
+                 bcc_to_salesforce ? @alert.bcc_to_salesforce_email : nil
+               ].compact }
+    p params
```
ragesoss, line 18: Leftover debugging?

prathamVaidya: Oops!

prathamVaidya:
workflow failure looks unrelated

Formasitchijoh:
@ragesoss  can i continue with this task

PR #5785, opened 2024-04-21
refactor: text_input.jsx to functional component

prathamVaidya:
## What this PR does

Part of https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5393 (Component refactor for React class components)

Refactors Class component to a function component for `text_input.jsx`

## Screenshots

image

### DIFF
--- a/app/assets/javascripts/components/common/text_input.jsx
+++ b/app/assets/javascripts/components/common/text_input.jsx
@@ -1,116 +1,138 @@
-import React from 'react';
-import createReactClass from 'create-react-class';
+import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
 import InputHOC from '../high_order/input_hoc.jsx';
 import Conditional from '../high_order/conditional.jsx';
 
-const TextInput = createReactClass({
-  displayName: 'TextInput',
+const TextInput = ({
+  value,
+  value_key,
+  editable,
+  label,
+  placeholder,
+  spacer,
+  valueClass,
+  p_tag_classname,
+  inline,
+  type,
+  max,
+  maxLength,
+  focus,
+  onBlur,
+  onClick,
+  append,
+  onKeyDown,
+  _value,
+  invalid,
+  onChange,
+  onFocus,
+  id,
+  children
+}) => {
+  const inputRef = useRef(null);
 
-  propTypes: {
-    value: PropTypes.any,
-    value_key: PropTypes.string,
-    editable: PropTypes.bool,
-    label: PropTypes.string,
-    placeholder: PropTypes.string,
-    spacer: PropTypes.string,
-    valueClass: PropTypes.string,
-    p_tag_classname: PropTypes.string,
-    inline: PropTypes.bool,
-    type: PropTypes.string,
-    max: PropTypes.string,
-    maxLength: PropTypes.string,
-    focus: PropTypes.func,
-    onBlur: PropTypes.func,
-    onClick: PropTypes.func,
-    append: PropTypes.node,
-    onKeyDown: PropTypes.func,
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/common/text_input.jsx
```diff
@@ -1,116 +1,138 @@
...
-      );
+    // The default maximum length of 75 ensures that the slug field
+    // of a course, which combines three TextInput values, will not exceed
+    // the maximum string size of 255.
+    const usedMaxLength = maxLength || '75';
+    const inputElement = (
+      <input
+        id={id}
+        className={className}
+        value={_value !== undefined ? _value : value || ''}
```
AmitJoki, line 73: This can be simplified

```
{_value ?? (value || '')}

File: app/assets/javascripts/components/common/text_input.jsx [hunk 2]
```diff
@@ -1,116 +1,138 @@
...
+  let labelContent;
+  const usedSpacer = spacer || ': ';
 
-  dateChange(date) {
-    const value = date ? date.format('YYYY-MM-DD') : '';
-    return this.props.onChange({ target: { value } });
-  },
+  if (label) {
+    labelContent = label + usedSpacer;
+  }
```
AmitJoki, line 43: Can be simplified:

```
const labelContent = label ? `${label}${spacer || ': '}` : undefined;
```

app/assets/javascripts/components/common/text_input.jsx
```diff
@@ -1,116 +1,138 @@
...
 
-  render() {
-    let label;
-    const spacer = this.props.spacer || ': ';
+  let usedValueClass = 'text-input-component__value ';
+  if (valueClass) {
+    usedValueClass += valueClass;
+  }
```
AmitJoki, line 48: Can be simplified:

```
const usedValueClass = `text-input-component__value ${valueClass}`;
```

app/assets/javascripts/components/common/text_input.jsx
```diff
@@ -1,116 +1,138 @@
...
 
-    if (this.props.label) {
-      label = this.props.label + spacer;
+  if (editable) {
+    let labelClass = '';
+    let inputClass = inline ? 'inline' : '';
+    if (invalid) {
+      labelClass += 'red';
+      inputClass += ' invalid';
```
AmitJoki, line 55: Can be simplified:

```
const labelClass = invalid ? 'red' : '';
const inputClass =`${inline ? 'inline' : ''} ${invalid ? 'invalid' : ''}`
```

This way one does not have to keep in mind trailing whitespaces required before concatenating another class.

For inputClass there's only two variables for now and the above string interpolation is fine. But if there were to be more variables then one can do:

```
const inputClass = [
  inline ? 'inline' : '',
  invalid ? 'invalid' : '',
  // other conditional class names
].join(' ');
```

app/assets/javascripts/components/common/text_input.jsx
prathamVaidya: Sure, also added the undefined case

```jsx
  const usedValueClass = `text-input-component__value ${valueClass ?? ''}`;
```

app/assets/javascripts/components/common/text_input.jsx
prathamVaidya: Done, I didn't touch those because I was under the impression to avoid ternary operators especially in react because they make the code more unreadable.

app/assets/javascripts/components/common/text_input.jsx
prathamVaidya: Done, Thanks for reviewing @AmitJoki

PR #5794, opened 2024-04-25
Refactored course_alert.jsx from class to functional component

om-chauhan1:
## What this PR does
This PR converted the course_alert.jsx from class to functional component.

## Screenshots

Before:

Before

After:

After


om-chauhan1:
All these failing test cases are passing locally.

Issue #5776, opened 2024-04-15
Improve BlockedUserAlert emails


ragesoss:
### Current Behavior:
Wiki Education staff get alerts when a student's account gets blocked.

### Desired Behavior:
If the block happens while the student is enrolled in a still-ongoing course, the student and instructor(s) should be emailed as well, with a full explanatory message in the email.

cyrillefr:
Hello @ragesoss ,

May I work on this issue ?

ragesoss:
Sure!

cyrillefr:
Hello @ragesoss ,
I need clarification(s) :)

Currently, when creating a new Alert, there is only one course associated.
So, there is a mail sent to the content_expert of the last course of the student. (the one(s) who are greater).

But to be blocked is not associated with one particular course though I understand the mail sent to the last course content expert -> someone will get notice of the blocking.

So now, my guess is there should be the same number of Alerts that the number of ongoing courses the student is following. (3 ongoing courses -> 3 Alerts -> mail to 3 groups of instructors).

Also, should I then keep the mail to the content expert of the last course ? (`def email_content_expert` in `app/modales/alert.rb`. Are the 'greeter' the Wiki Education staff you are referring to ?

Do the mail sent to the student should be simple like 'You have been blocked in Wikipedia' or be implemented as an AlertMail or a regular mail ?

ragesoss:
I think keeping the current BlockedUserMonitor logic basically unchanged is fine. There might be a few rare cases where a student is in more than one current course, in which case the instructors for any courses except the latest one would not get an email, but I think that's okay. In practice, that will almost never happen.

Yes, the 'greeter' is the Wiki Education staff I was referring to.

If you implement an email with a simple message (but using the styling of the other emails that go to non-admins, like basically any email besides the AlertMailer one), I can expand it with additional text.

PR #5795, opened 2024-04-25
Improve BlockedUserAlert emails

cyrillefr:
## What this PR does
Closes issue #5776 

On one Alert, need to email to instructors + blocked student
-edited-
- a new mailer + new template
- a new method in BlockUserAlert to mails to a set of concerned people
     which replace the current one in `blocked_user_monitor.rb`
- the corresponding spec
- a new line in preview mailer

I have added a template that might not be exactly polished but is a good start.
To be expanded by me or others.

### DIFF
--- a/app/mailers/blocked_user_alert_mailer.rb
+++ b/app/mailers/blocked_user_alert_mailer.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+class BlockedUserAlertMailer < ApplicationMailer
+  def self.send_mails_to_concerned(alert)
+    return unless Features.email?
+    email(alert).deliver_now
+  end
+
+  def email(alert)
+    @alert = alert
+    set_recipients
+    return if @recipients.empty?
+    params = { to: @recipients,
+             subject: @alert.main_subject }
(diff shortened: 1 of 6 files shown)


app/mailers/alert_mailer.rb
```diff
@@ -19,4 +19,21 @@ def alert(alert, recipient)
     params[:reply_to] = @alert.reply_to unless @alert.reply_to.nil?
     mail(params)
   end
+
+  def self.send_msg_to_student(alert, recipient)
```
ragesoss, line 23: I think adding the student and all the instructors to the recipients, rather than sending a separate email to each, is the best way to do this. That way, it can be the beginning of a conversation among the student, the instructors, and the Wiki Education staff member.

cyrillefr: Aaah. I have done a bit complicated code I guess. I will redo it, and will rebase since many things will disappear :)

cyrillefr:
I have updated the first comment. I don't know if it is visible or not. 
Now, it 's clear :)

ragesoss:
The code of this looks fine to me, but I need to come up with better email content before merging it. I'm working with my team on figuring that part out.

PR #5797, opened 2024-04-26
Refactored add_category_button from class to functional component

omChauhanDev:
## What this PR does
This PR converted the add_category_button.jsx from class to functional component.

## Videos

Before:

[Before.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/67263734-e14b-484f-9ac7-6489446ed280)

After:

[After.webm](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/6017b340-87db-431d-9c76-0e3017ae1a0d)

### DIFF
--- a/app/assets/javascripts/components/categories/add_category_button.jsx
+++ b/app/assets/javascripts/components/categories/add_category_button.jsx
@@ -1,11 +1,9 @@
-import createReactClass from 'create-react-class';
-import PropTypes from 'prop-types';
-import { connect } from 'react-redux';
 import React from 'react';
+import PropTypes from 'prop-types';
+import { useSelector, useDispatch } from 'react-redux';
 import { initiateConfirm } from '../../actions/confirm_actions';
 import Popover from '../common/popover.jsx';
-import Expandable from '../high_order/expandable.jsx';
-
+import useExpandablePopover from '../../hooks/useExpandablePopover';
 import CategoriesScoping from '../course_creator/scoping_methods/categories_scoping';
 import PagePileScoping from '../course_creator/scoping_methods/pagepile_scoping';
 import PetScanScoping from '../course_creator/scoping_methods/petscan_scoping';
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/components/categories/add_category_button.jsx
```diff
@@ -1,132 +1,122 @@
...
+    description = (
+      <p>
+        {I18n.t('categories.psid_description')}
+      </p>
+    );
+  }
 
-    const buttonText = I18n.t(`categories.add_${this.props.source}`);
-    const showButton = <button className={`${className}`} onClick={this.props.open}>{buttonText}</button>;
+  const permitted = true;
```
AmitJoki, line 60: `permitted` is always `true` and is not reassigned. This means the following `if (permitted)` always executes. This variable and the ensuing if block nesting should either be removed or check if the `permitted` needs to be modified based on something rather than always being `true`.

app/assets/javascripts/components/categories/add_category_button.jsx
```diff
@@ -1,132 +1,122 @@
...
-      } else {
-        inputField = <TemplatesScoping />;
-      }
-      editRow = (
-        <tr className="edit">
-          <td>
-            <form
-              onSubmit={this.addCategory} style={{
+  const buttonText = I18n.t(`categories.add_${source}`);
+  const showButton = <button className={`${className}`} onClick={open}>{buttonText}</button>;
```
AmitJoki, line 65:
```
className={className}
```

No need for string interpolation here.

ragesoss (approved):
Refactor looks clean and follows the established pattern used by the eight other components already on `useExpandablePopover` in master. Both prior review comments from @AmitJoki are addressed: the unused `permitted` flag and its dead `if (permitted)` wrapper are gone, and the `className` template literal is gone too. The bonus i18n extraction (PSID/PagePile description strings into `en.yml`) is a nice cleanup.

Tests run on the rebased branch:
- `yarn build` — passes (no new warnings on the changed file)
- `spec/features/category_scopes_spec.rb` — 7/7 pass (covers category + template flows end-to-end, including the confirm-and-close path)
- An exploratory feature spec written during review to cover the PSID and PagePile popovers (which `category_scopes_spec.rb` does not exercise) — 4/4 pass

Behavioral note for the reviewer's awareness, not a blocker: the old `onConfirm` called `open(null)` to close the popover after confirm; the new `onConfirm` does not. The hook's outside-click handler covers this in practice (clicking "OK" on the confirm dialog is outside `pop__container`), and this matches the convention in `enroll_button.jsx` and the other migrated components.

Two optional follow-up suggestions, neither blocking:

1. `let editRow = null;` (line 65) is initialized to `null` and then unconditionally reassigned a few lines later. With `permitted` removed, it can simplify to `const editRow = (...)`.

2. Adopting the exploratory spec below would close a real coverage gap — `category_scopes_spec.rb` exercises only `category` and `template` sources, so PSID and PagePile popover paths are untested. The spec was written and verified during review:

   ```ruby
   # spec/features/add_category_button_popovers_spec.rb
   require 'rails_helper'

   describe 'AddCategoryButton popover variants', js: true do
     let(:course) { create(:course, type: 'ArticleScopedProgram') }
     let(:user) { create(:user) }

     before do
       JoinCourse.new(course:, user:, role: CoursesUsers::Roles::INSTRUCTOR_ROLE)
       login_as user
       stub_oauth_edit
       visit "/courses/#{course.slug}/articles"
     end

     it 'shows the PetScan description when opening the PSID popover' do
       click_button 'Add PetScan PSID'
       expect(page).to have_content(/PetScan queries only work with the Dashboard one wiki at a time/)
     end

     it 'shows the PagePile description when opening the PagePile popover' do
       click_button 'Add PagePile id'
       expect(page).to have_content(/Make sure the PagePile.s wiki is one of the tracked wikis/)
     end

     it 'does not show a description for the category or template popovers' do
       click_button 'Add category'
       expect(page).not_to have_content(/PetScan queries only work/)
       expect(page).not_to have_content(/Make sure the PagePile/)
       find('h3', text: 'Tracked Categories').click

       click_button 'Add template'
       expect(page).not_to have_content(/PetScan queries only work/)
       expect(page).not_to have_content(/Make sure the PagePile/)
     end

     it 'closes a popover when another category source button is clicked' do
       click_button 'Add PetScan PSID'
       expect(page).to have_content(/PetScan queries only work/)
       click_button 'Add PagePile id'
       expect(page).to have_content(/Make sure the PagePile/)
       expect(page).not_to have_content(/PetScan queries only work/)
     end
   end
   ```

   This spec was AI-generated during review and verified to pass against the rebased branch; treat it as a starting point rather than as authoritative.

*Drafted in a Claude Code session (~25 min, ~20 interactions). Sage directed the review and approved the comment before posting. The diff, the existing `category_scopes_spec.rb` run, and the exploratory spec run were all verified in the session; the AI-generated exploratory spec was reviewed and run by Sage but not independently audited line-by-line.*

PR #5805, opened 2024-05-16
Add Redirect For Requests For The Current Term Of Alerts

### LINKED ISSUE #5800: Add redirect for JSON and HTML requests for the current term of Alerts
### Current Behavior:

There is a public feed of Alerts for each campaign, available as both JSON and a web page:

* https://dashboard.wikiedu.org/campaigns/spring_2024/alerts.json
* https://dashboard.wikiedu.org/campaigns/spring_2024/alerts

However, the place to go for the current term changes every semester.

### Desired Behavior:

Add a redirect path so that bookmarks and/or software can reliably access the alerts for the current default term (`Campaign.default_campaign`).

### Note

This was requested by User:Mathglot on en.wikipedia.

cyrillefr:
## What this PR does
This PR aims to close #5800 

- One can now bookmark:
  - `/campaigns/current/alerts` and/or
  - `/campaigns/current/alerts.json`

With a new route rule to go to the new `current_alerts` method in the `Campaign` controller.
Inspired by the `alerts` method from same controller.

The `show` method had to be modified because React calls some `campaign/slug.json` after populating with some
other calls. Whether in html or json format.
So the `current` has to be interpolated to the real current Campaign
Current is the same as default

I believe one could add also this bookmark: `http://domain/campaigns/current/programs` quite easily


ragesoss:
What if we try to make a campaign with the slug `current`? I guess that should be prevented with a validation in `campaign.rb`, or some other way.

cyrillefr:
Hello @ragesoss ,

I need to check about the validation.
In case it is possible, this record should be updated with the values of the current campaign. Or may be some kind of link if it is possible.
I will investigate and in case post some code.

cyrillefr:
Hello @ragesoss ,

One can `dup` a Campaign object and give it the `current` slug (and also title). But such an object has no courses.
There is a need to do some other works to get the copy to link to associated objects of the current Campaign.

But also, how to keep these 2 objects in synch ? 

We would need to add a hook on save/update on Campaign, check if it is about the current campaign, and if so, proceed to update current.slug Campaign. In my opinion, this is a bit heavy.

ragesoss:
@cyrillefr i was thinking that we should just use a validation on the Campaign model to make sure we never have an actual campaign with the slug `current`. Then way you've implemented it in the controller will work fine.

cyrillefr:
Aaaah :)
I added the validation.

cyrillefr:
Checks failed at linting stage. But for a source file version that is not part of my branch ....
File `spec/controllers/articles_controller_spec.rb` that have a line too long.
Version of my branch is older and does not have this modification. 
A bit strange

ragesoss:
My fault. I added that linting violation while fixing an urgent but last week. Don't worry about that failure.

PR #5810, opened 2024-06-03
Added functionality for creating new Training Libraries

omChauhanDev:
## What this PR does
This pr added functionality for creating new training libraries through dashboard.

Tasks :

- [x] Implemented Validations at frontend.

- [x] Applied access rights : Only for admin.

- [x] Displaying success, custom error notifications to user.

### Before:
image

#### 1. Button on main training page
image

#### 2. Filling up form
image

#### 3. Successfully created library
image

#### 4. First Library
image

#### 4. Validations at front-end
image

#### 5. Custom Error Message 
image

### DIFF
--- a/app/assets/javascripts/actions/training_modification_actions.js
+++ b/app/assets/javascripts/actions/training_modification_actions.js
@@ -1,0 +1,35 @@
+import { API_FAIL } from '../constants/api';
+import { addNotification } from '../actions/notification_actions';
+import logErrorMessage from '../utils/log_error_message';
+import request from '../utils/request';
+
+const createLibraryPromise = async (library, setSubmitting, modalHandler) => {
+    const response = await request('/training_library', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(library),
+    });
+    setSubmitting(false);
+    modalHandler(false);
+    if (!response.ok) {
+      logErrorMessage(response);
+      const data = await response.text();
+      response.responseText = data;
+      throw response;
+    }
+    return response.json();
+  };
+
+export const createLibrary = (library, setSubmitting, modalHandler) => (dispatch) => {
+    return createLibraryPromise(library, setSubmitting, modalHandler)
+    .then(() => {
+        dispatch(addNotification({
+            type: 'success',
+            message: 'Library Created Successfully.',
+            closable: true
+        }));
+    })
+    .catch(error => dispatch({ type: API_FAIL, data: error }));
+};
--- a/app/assets/javascripts/training/components/modals/create_library.jsx
+++ b/app/assets/javascripts/training/components/modals/create_library.jsx
@@ -1,0 +1,123 @@
+import React, { useState, useEffect } from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import Modal from '../../../components/common/modal.jsx';
+import TextInput from '../../../components/common/text_input.jsx';
+import TextAreaInput from '../../../components/common/text_area_input.jsx';
+import { createLibrary } from '../../../actions/training_modification_actions.js';
+import { setValid, setInvalid, activateValidations, resetValidations } from '../../../actions/validation_actions.js';
+import { firstValidationErrorMessage } from '../../../selectors';
+
+const CreateLibrary = (props) => {
+  const [submitting, setSubmitting] = useState(false);
+  const [library, setLibrary] = useState({
+    name: '',
+    slug: '',
+    introduction: ''
+  });
+  const firstErrorMessage = useSelector(state => firstValidationErrorMessage(state));
+  const dispatch = useDispatch();
+
+  const handleInputChange = (key, value) => {
+    setLibrary(prevLibrary => ({
+      ...prevLibrary,
+      [key]: value
+    }));
+  };
+
+  useEffect(() => {
+    dispatch(resetValidations());
+  }, []);
+
+  const validateFields = () => {
+    let valid = true;
+    const message = I18n.t('training.validation_message');
+    if (library.name.trim() === '') {
+      dispatch(setInvalid('name', message));
+      valid = false;
+    } else {
+      dispatch(setValid('name'));
+    }
+
+    if (library.slug.trim() === '') {
+      dispatch(setInvalid('slug', message));
+      valid = false;
+    } else {
+      dispatch(setValid('slug'));
+    }
+
+    if (library.introduction.trim() === '') {
+      dispatch(setInvalid('introduction', message));
+      valid = false;
+    } else {
+      dispatch(setValid('introduction'));
+    }
+
+    return valid;
+  };
+
+  const submitHandler = () => {
+    dispatch(activateValidations());
+
+    if (validateFields()) {
+      setSubmitting(true);
+      dispatch(createLibrary(library, setSubmitting, props.modalHandler));
+    }
+  };
+
+  let formStyle;
+  if (submitting) {
+    formStyle = { pointerEvents: 'none', opacity: '0.5' };
+  }
+
+  return (
+    <>
+      <Modal>
+        <div className="container">
+          <div className="wizard__panel active training_modal" style={formStyle}>
+            <h3>{I18n.t('training.create_library')}</h3>
+            <p>{I18n.t('training.create_library_msg')}</p>
+            <div className="column">
+              <TextInput
+                id="name"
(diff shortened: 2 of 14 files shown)


app/assets/javascripts/actions/training_modification_actions.js
```diff
@@ -0,0 +1,35 @@
...
+const createLibraryPromise = async (library, setSubmitting, modalHandler) => {
+    const response = await request('/training_library', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify(library),
+    });
+    setSubmitting(false);
+    modalHandler(false);
```
TheTrio, line 15: `modalHandler` [doesn't take any arguments](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5810/files#diff-ebc286207cbdcfed824d6b79fd1d3086737bcd6cc94880d79a31181968f20943R11-R14) — I think this is in error?

app/assets/javascripts/training/components/modals/create_library.jsx
```diff
@@ -0,0 +1,123 @@
...
+  };
+
+  useEffect(() => {
+    dispatch(resetValidations());
+  }, []);
+
+  const validateFields = () => {
+    let valid = true;
+    const message = I18n.t('training.validation_message');
+    if (library.name.trim() === '') {
```
TheTrio, line 34: `!library.name.trim()`

TheTrio: same for the others. nit, obviously

app/assets/javascripts/training/components/modals/create_library.jsx
```diff
@@ -0,0 +1,123 @@
...
+                onChange={handleInputChange}
+                value={library.slug}
+                value_key="slug"
+                required
+                editable
+                label={I18n.t('training.library_slug')}
+                placeholder={`${I18n.t('training.enter')} ${I18n.t('training.library_slug')}`}
+              />
+              <button className="button light" onClick={props.modalHandler}>{I18n.t('training.cancel')}</button>
+              <span className="validation-error"> &nbsp; {firstErrorMessage || '\xa0'}</span>
```
TheTrio, line 101: What's the `\xa0` for?

app/assets/javascripts/actions/training_modification_actions.js
omChauhanDev: yes, now i made it correct

app/assets/javascripts/training/components/modals/create_library.jsx
omChauhanDev: sure

app/assets/javascripts/training/components/modals/create_library.jsx
omChauhanDev: **\xa0:** This is the non-breaking space character. It's used here to ensure that there is always at least one character's worth of space in the span, even if 'firstErrorMessage' is empty. This avoids any layout issues that might occur if the span were completely empty.
**Non-Breaking Space:** Unlike a regular space character, a non-breaking space won't cause line breaks. This makes it useful in situations where you want to ensure that spacing doesn't affect the flow of text or layout in unintended ways

app/assets/javascripts/training/components/training_library_handler.jsx
```diff
@@ -0,0 +1,29 @@
...
+import Notifications from '../../components/common/notifications.jsx';
+import { getCurrentUser } from '../../selectors';
+
+const TrainingLibraryHandler = () => {
+  const [showModal, setShowModal] = useState(false);
+  const currentUser = useSelector(state => getCurrentUser(state));
+
+  const modalHandler = () => {
+    setShowModal(!showModal);
+  };
```
TheTrio, line 13: `toggleModal` would be a better name here I think

TheTrio:
I think having a UI similar to this(which doesn't lose all the info you entered if the library with the same name already exists) would be better.

image

ragesoss:
Yes, it would be nice to follow the same patter as the course creation modal, and to validate from the form so that errors can be fixed without starting over.

app/assets/javascripts/training/components/training_library_handler.jsx
omChauhanDev: sure

omChauhanDev:
okay, working on this

omChauhanDev:
I have updated the validation pattern, kindly check this out :

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/119471995/421aad36-6537-41e9-b553-076e857b2480

app/controllers/training_library_controller.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class TrainingLibraryController < ApplicationController
+  def create
+    training_library = TrainingLibrary.new(training_library_params)
+    if training_library.save
+      render json: { status: 'success', data: training_library }, status: :created
+    else
+      render json: { status: 'error', message: training_library.errors.full_messages },
```
TheTrio, line 9: why's this called `message` if its an array of messages?

app/assets/javascripts/actions/training_modification_actions.js
```diff
@@ -0,0 +1,54 @@
...
+  messages.forEach((message) => {
+    const lowercaseMessage = message.toLowerCase();
+    if (lowercaseMessage.includes('name')) {
+      dispatch(setInvalid('name', message));
+    } else if (lowercaseMessage.includes('slug')) {
+      dispatch(setInvalid('slug', message));
+    } else if (lowercaseMessage.includes('introduction')) {
+      dispatch(setInvalid('introduction', message));
+    } else {
+      dispatch({ type: API_FAIL, data: error });
```
TheTrio, line 18: We probably should only dispatch an `API_FAIL` event only once — so I recommend switching to a `for..of` loop and doing an early return in the `else` block.

app/assets/javascripts/actions/training_modification_actions.js
```diff
@@ -0,0 +1,54 @@
...
+    }
+  });
+};
+
+const createLibraryPromise = async (library, setSubmitting) => {
+    const response = await request('/training_library', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
```
TheTrio, line 28: This is the default, so we can remove this

app/models/training_library.rb
```diff
@@ -23,8 +23,9 @@ class TrainingLibrary < ApplicationRecord
   serialize :translations, Hash
 
   validates_uniqueness_of :slug, case_sensitive: false
+  validates_uniqueness_of :name, case_sensitive: false
```
TheTrio, line 26: why was this added?

app/controllers/training_library_controller.rb
omChauhanDev: i have renamed it to 'errorMessages'

app/assets/javascripts/actions/training_modification_actions.js
omChauhanDev: okay

app/assets/javascripts/actions/training_modification_actions.js
omChauhanDev: done

app/models/training_library.rb
omChauhanDev: in order to make sure that two libraries won't contain same name.

image

app/assets/javascripts/actions/training_modification_actions.js
TheTrio: yes, this works too. We actually just disable that rule whenever we want to use `for...of`

ragesoss:
I tested this out, and it's working pretty smoothly. One improvement would be to refresh the page after creating a new Library or Category, as the new one doesn't appear until after a page refresh.

omChauhanDev:
Hi @ragesoss and @TheTrio,

I wanted to bring this to your attention. I have implemented 'switching between different modes' using session storage and have also modified all the relevant test cases. I'm considering pushing this commit in a separate PR to maintain a clear separation of concerns.

Please let me know if you have any feedback.

ragesoss:
@omChauhanDev i created this branch with the earlier commits: https://github.com/WikiEducationFoundation/WikiEduDashboard/tree/Om-Training

You should open the separate PR against that branch, for dealing with merge conflicts.

PR #5813, opened 2024-06-03
Implemented Timeline Copying Feature for Admin Users

PR4NJ41:
## What this PR does
The PR comprises methods that facilitate the Timeline copying from Wiki-Edu dashboard to P&E dashboard and vice versa.

## Screenshots
Before:
Before

After:
After

### DIFF
--- a/app/services/copy_course.rb
+++ b/app/services/copy_course.rb
@@ -26,7 +26,7 @@ class CopyCourse
   def copy_main_course_data
     # Extract the attributes we want to copy
     params_to_copy = %w[school title term description start end subject slug timeline_start
-                        timeline_end type flags created_at weekdays]
+                        timeline_end type flags weekdays]
     copied_data = {}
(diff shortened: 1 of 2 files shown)


app/services/copy_course.rb
```diff
@@ -24,7 +26,7 @@ def make_copy
   def copy_main_course_data
     # Extract the attributes we want to copy
     params_to_copy = %w[school title term description start end subject slug timeline_start
-                        timeline_end type flags]
+                        timeline_end type flags created_at weekdays]
```
ragesoss, line 29: I don't think we want to copy `created_at`. Better to use the timestamp from when the copying happened.

PR4NJ41: @ragesoss
Sure, I have removed `created_at`. Now, it will automatically take the current date when the copy of the course is made.

PR #5816, opened 2024-06-06
Fix Rubocop RedundantAssignment offenses #5297

NateWebber:
## What this PR does
This pull request addresses part of issue #5297, namely the rubocop rule [Style/RedundantAssignment](https://docs.rubocop.org/rubocop/cops_style.html#styleredundantassignment). When I enabled the rule there were a total of 32 offenses, which are now fixed. 

I noticed that Style/RedundantReturn is in the permanent exceptions, and I think that there could be a similar argument made for making this rule an exception as well (and thus this PR unnecessary). Some methods that literally only assign and then return a variable are pretty clearly better off following this rule, whereas others that are more complicated might be more readable before the change. In lib/wikitext.rb see #mediawiki_to_markdown for an example of the former, and #assignments_to_wikilinks for an example of the latter.

I personally think the code is mostly improved by adhering to this rule but I'd appreciate feedback from someone more familiar with this codebase. It was a quick fix so no harm either way.

### DIFF
--- a/lib/pagepile_api.rb
+++ b/lib/pagepile_api.rb
@@ -12,7 +12,8 @@ class PagePileApi
 
     update_language_and_project
 
-    @pile_data['pages']
+    titles = @pile_data['pages']
+    titles
   end
 
   ###################
--- a/lib/pagepile_api.rb
+++ b/lib/pagepile_api.rb
@@ -12,8 +12,7 @@ class PagePileApi
 
     update_language_and_project
 
-    titles = @pile_data['pages']
-    titles
+    @pile_data['pages']
   end
 
   ###################
--- a/lib/wiki_assignment_output.rb
+++ b/lib/wiki_assignment_output.rb
@@ -129,7 +129,8 @@ class WikiAssignmentOutput
     return page_content if page_content.include?(new_tag_with_signature)
 
     # Otherwise, we need to add the tag to the right place.
-    insert_tag_into_talk_page(page_content, new_tag_with_signature)
+    page_content = insert_tag_into_talk_page(page_content, new_tag_with_signature)
+    page_content
   end
 
   def remove_assignment_tag(page_content, existing_tag, header)
--- a/lib/wiki_assignment_output.rb
+++ b/lib/wiki_assignment_output.rb
@@ -47,8 +47,7 @@ class WikiAssignmentOutput
     # This is to avoid creating a new empty talk page.
     return nil if @assignments.empty? && initial_page_content.empty?
 
-    page_content = build_assignment_page_content(assignments_tag, initial_page_content)
-    page_content
+    build_assignment_page_content(assignments_tag, initial_page_content)
   end
 
   ###################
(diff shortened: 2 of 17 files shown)


NateWebber:
The above push was just squashing the commit where I undid some unintentional changes to unrelated files into the main commit

ragesoss:
Thanks @NateWebber! This is a very nice PR, and I appreciate the analysis you included. Browsing through all of the rule violations, I think my preference would be to move this rule to the permanent exceptions list. In some cases, the variable names are doing the work of explaining what kind of thing is being returned, or making a series of transformations more parallel and easy to read. But some of them are better off with the refactors you've done as well. I'll comment on the individual ones I would like to keep as-is.

lib/pagepile_api.rb
```diff
@@ -12,8 +12,7 @@ def page_titles_for_pileid
 
     update_language_and_project
 
-    titles = @pile_data['pages']
-    titles
+    @pile_data['pages']
```
ragesoss, line 15: Keep this as-is, since it communicates the meaning of the data as a set of page titles.

lib/wiki_assignment_output.rb
```diff
@@ -130,8 +129,7 @@ def replace_or_add_assignment_tag(page_content, existing_tag, new_tag)
     return page_content if page_content.include?(new_tag_with_signature)
 
     # Otherwise, we need to add the tag to the right place.
-    page_content = insert_tag_into_talk_page(page_content, new_tag_with_signature)
-    page_content
+    insert_tag_into_talk_page(page_content, new_tag_with_signature)
```
ragesoss, line 132: Keep this-as, since it makes it clear it more clear that the final return value is a further iteration of the `page_content` that is being modified throughout the method.

lib/wikitext.rb
```diff
@@ -33,21 +33,18 @@ def self.markdown_to_mediawiki(item)
     wikitext = replace_code_with_nowiki(wikitext)
     wikitext = reformat_image_links(wikitext)
     wikitext = replace_at_sign_with_template(wikitext)
-    wikitext = reformat_links(wikitext)
-    wikitext
+    reformat_links(wikitext)
```
ragesoss, line 36: Keep as-is.

lib/wikitext.rb
```diff
@@ -33,21 +33,18 @@ def self.markdown_to_mediawiki(item)
...
   end
 
   def self.html_to_mediawiki(item)
     wikitext = PandocRuby.convert(item, from: :html, to: :mediawiki)
     wikitext = replace_code_with_nowiki(wikitext)
     wikitext = replace_at_sign_with_template(wikitext)
-    wikitext = reformat_links(wikitext)
-    wikitext
+    reformat_links(wikitext)
```
ragesoss, line 43: Keep as-is.

lib/wikitext.rb
```diff
@@ -64,17 +61,15 @@ def self.replace_code_with_nowiki(text)
...
   end
 
   def self.assignments_to_wikilinks(assignments, home_wiki)
     return '' if assignments.blank?
     formatted_titles = assignments.map do |assignment|
       format_assignment_title(assignment, home_wiki)
     end
-    wikitext = '[[' + formatted_titles.join(']], [[') + ']]'
-    wikitext
+    '[[' + formatted_titles.join(']], [[') + ']]'
```
ragesoss, line 72: Keep as-is.

NateWebber:
Hi, thanks for the feedback! I figured something like this would ultimately end up making the most sense. I've reverted the methods you indicated and added the rule to the permanent exceptions.

ragesoss:
LGTM!

Issue #5798, opened 2024-04-29
ArticleViewer legend usernames are badly formatted when there are many of them


ragesoss:
This is a regression; the footer previously handled long lists pretty well, including cases where the list needs to move to a second row.

![Screenshot from 2024-04-29 11-10-56](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/848483/24270739-f8b7-424e-a542-e796cf940af3)

Here's the live example: https://dashboard.wikiedu.org/courses/The_Ohio_State_University/Classical_Archaeology_(Spring_2024)/students/articles/SaxicolousArtume?showArticle=673815983

PR #5818, opened 2024-06-07
Fix ArticleViewer legend usernames #5798

Abishekcs:
#5798

### **What this PR does**

This PR fixes a formatting issue in the ArticleViewer component where usernames in the legend were poorly formatted when there were many of them.

### **Cause of the Bug:**

The specific point where the bug was introduced is still unknown. I've checked previous commits where this issue was supposedly resolved, but the problem still persists in my local development environment.

### **Fix for the Bug:**

1. **Flex-wrap Property:**
   - The issue is resolved by using the `flex-wrap: wrap` property in the `.user-legend-wrap` class of `_diff_viewer.styl`.

2. **Width Adjustments:**
   - Increased the width of the first flex container in the article footer to allow more usernames to fit on a single line. The second flex container, which holds three fixed-size buttons, remains unaffected by this change. If additional buttons are added in the future, the width property of `.user-legend-wrap` may need further adjustment, as it is specified in `rem`. This approach was chosen because the three buttons have fixed sizes and will not change.

3. **UI Consistency:**
   - Fixing the `article-viewer-legend` with `flex-wrap` introduced `minor(bug)` changes in the `diff-viewer-legend` UI. These were addressed by removing some `inline styles` from `diff_viewer.jsx`, ensuring that `finalDate` and `characterCount` appear in their respective rows instead of a single row.

**Note:**

The existing `.user-legend-wrap` CSS class, which was modified to fix the bug, is only used in two places: `article-viewer-legend and diff-viewer-legend`.

### **Before:**

Article Viewer Legend:

article legend

Diff Viewer Legend (First Revision):

diff legend (First Revision)

Diff Viewer Legend: 

diff legend

### **After:**

Article Viewer Legend:

article legend

Diff Viewer Legend (First Revision):

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/15f8b3b9-9bc8-4d6b-86bb-0d8563179cbb

Diff Viewer Legend: 

diff legend

### DIFF
--- a/app/assets/javascripts/components/revisions/diff_viewer.jsx
+++ b/app/assets/javascripts/components/revisions/diff_viewer.jsx
@@ -280,14 +280,14 @@ const DiffViewer = createReactClass({
     if (!this.props.first_revision) {
       formatedDate = formatDateWithTime(this.props.revision.date);
       editDate = I18n.t('revisions.edited_on', { edit_date: formatedDate });
-      finalDate = <div className="diff-viewer-legend" style={{ justifyContent: 'center' }}>{editDate}</div>;
-      charactersCount = <div className="diff-viewer-legend" style={{ justifyContent: 'flex-end' }}>{this.props.revision.characters} {I18n.t('revisions.chars_added')}</div>;
+      finalDate = <div className="diff-viewer-legend" style={{ width: '66%' }}>{editDate}</div>;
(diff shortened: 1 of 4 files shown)


ragesoss:
Awesome, thanks!

PR #5819, opened 2024-06-07
Regenerate Gemfile.lock file to fix data rearchitecture project deployment

gabina:
## What this PR does
This PR re-generates the `Gemfile.lock` file to fix the data rearchitecture project deployment. I wasn't able to make the Passenger server work because of the following error:
`Could not spawn process for application /var/www/dashboard/current: The application encountered the following error: Could not find mini_portile2-2.8.2 in any of the sources (Bundler::GemNotFound)`
It looks like the `mini_portile2` gem is not longer necessary for `nokogiri`, so running `bundle install `  didn't install the `mini_portile2` gem. However, at some point Passenger got confused and thought that needed the missing `mini_portile2` gem, causing the error mentioned above.

After spending several hours looking at this, the easiest way I found to solve this was the following:
- Remove the existing  `Gemfile.lock`
- Run `bundle install`
- Check the new  `Gemfile.lock` file doesn't include the `mini_portile2` gem

I'm not sure if this change is safe to be applied to the master branch, as it updates a lot of gems, but I think it's OK to be merged on my `data-rearchitecture-for-dashboard` branch.

This PR also adds the `PASSENGER_INSTANCE_REGISTRY_DIR` environment variable to the data-rearchitecture-project-test stage, because otherwise I got the error:

> ERROR: Phusion Passenger doesn't seem to be running. If you are sure that it is running, then the causes of this problem could be:
> 
> You customized the instance registry directory using Apache's PassengerInstanceRegistryDir option, Nginx's passenger_instance_registry_dir option, or Phusion Passenger Standalone's --instance-registry-dir command line argument. If so, please set the environment variable PASSENGER_INSTANCE_REGISTRY_DIR to that directory and run passenger-status again.
> The instance directory has been removed by an operating system background service. Please set a different instance registry directory using Apache's PassengerInstanceRegistryDir option, Nginx's passenger_instance_registry_dir option, or Phusion Passenger Standalone's --instance-registry-dir command line argument.


gabina:
well, it looks like it breaks a lot of tests. Will try to apply the changes only for `nokogiri`

PR #5821, opened 2024-06-09
[BUG] Fix YAML keys for future_updates_remaining

### LINKED ISSUE #5820: Translation issues with future_updates_remaining
### What is happening?

Some i18n strings are not properly processed at translatewiki.net.

```
future_updates_remaining:
      true: The stats will continue to be updated until %{date}.
      false: The stats for this program are no longer being updated regularly.
```

True and false are booleans in yaml and at least our yaml parser (phpyaml) fails to handle them properly in array keys. This means these strings don't get proper translations from translatewiki.net.

Can you add quotes around the keys or use different keys?

24archit:
FIXED #5820

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1082,8 +1082,8 @@ en:
     edit_count_description: Total Edits
     error_count_message: The dashboard encountered %{error_count} error(s) during the most recent update.
     future_updates_remaining:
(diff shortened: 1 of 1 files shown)


ragesoss:
This will also require changing the spot where these messages are used. I'll add a commit that does this and then merge it.

ragesoss:
Thanks @24archit !

PR #5826, opened 2024-06-14
[WIP] Data rearchitecure proof of concept

gabina:
## What this PR does
This PR makes a small proof of concept for the data re-architecture project. Basically:

- It adds a new `RevisionDataManager` class that given a course, a wiki and a specific period of time (in the future, a timeslice), it fetches revision and revision scores from different APIs, and returns an array with all that data.
- It adds the client of the  `RevisionDataManager` class: `UpdateCourseStatsTimeslice` class that updates timeslices from this revision data in RAM (not commited yet).
- It implements the `update_cache_from_revisions` method for the `ArticleCourseTimeslice`, which updates an  `ArticleCourseTimeslice` record with the new revisions in memory.
- It updates the existing `update_cache` method for the `ArticleCourse`, to update caches based on the `ArticleCourseTimeslice` records instead of the revision table (`ArticlesCourses.update_cache_from_timeslices`).

This PR adds specs for all the new behaviors.
There are a lot of TODO comments that will be handled in the future.
A similar process should be done for `CourseWikiTimeslice` and `CourseUserWikiTimeslice` timeslices.

For `ArticlesCourses` records, now `view_count` is calculated  based on the first non-empty article course timeslice record `start` date (instead of using the first revision). We estimate the first non-empty record checking `user_ids` field is not null. This should lead to the same results.

### DIFF
--- a/app/models/article_course_timeslice.rb
+++ b/app/models/article_course_timeslice.rb
@@ -15,8 +15,9 @@
 #  updated_at        :datetime         not null
 #
 class ArticleCourseTimeslice < ApplicationRecord
-  belongs_to :articles_courses
+  belongs_to :articles_courses, foreign_key: 'article_course_id'
 
+  scope :non_empty, -> { where.not(user_ids: nil) }
   serialize :user_ids, Array # This text field only stores user ids as text
 
   ####################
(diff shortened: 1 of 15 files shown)


gabina:
RevisionScoreImporter specs are failing, I think I understand why. Will fix it on Monday.

gabina:
The following specs are failing but they're not related to my changes. I think both specs were already fixed in master. I'll sync my own `data-rearchitecture-for-dashboard` branch with master in a next PR.
```
rspec ./spec/lib/replica_spec.rb:151 # Replica API requests functions identically on non-English wikis
rspec ./spec/lib/replica_spec.rb:98 # Replica API requests Article retrieval when fetching existing articles returns article "Autism"
```

PR #5830, opened 2024-06-16
Refactor: Migrated from Class-Based Component to Functional Component of StudentTabHandler[WIP]

codic-yeeshu:
## What this PR does
This pull request refactors the StudentsTabHandler component from a class-based component to a functional component.
I am a first-time contributor to Wikimedia. If there are any issues with my pull request, please provide feedback, and I will address them promptly.

before change 

https://github.com/user-attachments/assets/1b1213d9-b7f6-4807-a13b-272f68ceeea0

after first week

https://github.com/user-attachments/assets/27ec813d-a586-48ab-a44c-8c9fdfe04e29

### DIFF
--- a/app/assets/javascripts/components/students/containers/StudentsTabHandler.jsx
+++ b/app/assets/javascripts/components/students/containers/StudentsTabHandler.jsx
@@ -1,5 +1,4 @@
-import React from 'react';
-import createReactClass from 'create-react-class';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Navigate, Route, Routes } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


ragesoss:

ragesoss:
Closing this without merging. The class-to-functional refactor preserves the JSX shape but disconnects the component from Redux dispatch in several places. References to `fetchAssignments`, `fetchArticles`, `notifyOverdue`, and `sortUsers` inside the function body resolve to the module-level imports rather than the dispatch-bound versions injected by `connect()` — those props are not destructured from the function arguments. Calling an action creator directly returns the action object; it never dispatches.

Concrete consequences:

- `useEffect` (`StudentsTabHandler.jsx:30-32`) calls `delayFetchAssignmentsAndArticles({ fetchAssignments, fetchArticles }, …)`. The helper at `app/assets/javascripts/components/util/helpers.js:5-18` reads `props.loadingArticles`, `props.loadingAssignments`, `props.course_id`, and `props.limit`, all undefined in the passed object — so the early-return path fires and `setLoading(false)` runs immediately without any fetch.
- `notify()` (lines 34-38) calls the imported `notifyOverdue`, so confirming the overdue prompt is a no-op.
- `sortSelect()` (line 40) calls the imported `sortUsers`, so the sort dropdown does nothing. The same imported `sortUsers` is forwarded to `<Overview>` and `<Articles>` via the spread `props` object, so the same regression propagates to children.
- `propTypes` (lines 75-76) declares `notifyOverdue` and `fetchArticles` as required, but the function never reads either from props.

Given the PR's age and that a correct redo would touch roughly the same lines, closing rather than asking for revisions. For future class-to-functional refactors of `connect()`-wrapped components, the safe pattern is to destructure every dispatch-bound prop from the function arguments (or rename the imports to avoid shadowing), so the bound versions are what actually execute.

*Drafted in a Claude Code session (~10 min, ~4 interactions). Sage directed the review. Findings were verified by reading the merged file, `delayFetchAssignmentsAndArticles` in `helpers.js`, and the children that consume `sortUsers` and related props; no JS specs were run (none cover this component).*

PR #5836, opened 2024-06-19
Refactor : Converted timeline_handler.jsx from class based components to Functional Components

Prince0906:
## What this PR does
Converted timeline_handler.jsx from class based to functional components
## Screenshots
Before:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/145124801/7fa51c00-d612-46e5-a75d-dc9a44cdb43e

After:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/145124801/3bbdab01-3cbf-420e-8bb4-9f411db72b7f

This is my first time contrubuting to Wikimedia.So,if I have done anything wrong in my PR,give feedback and I will try to rectify that.

### DIFF
--- a/app/assets/javascripts/components/timeline/timeline_handler.jsx
+++ b/app/assets/javascripts/components/timeline/timeline_handler.jsx
@@ -1,4 +1,4 @@
-//Import necessary hooks (useState, useEffect) and useNavigate from react-router-dom
+// Import necessary hooks (useState, useEffect) and useNavigate from react-router-dom
 import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
--- a/app/assets/javascripts/components/timeline/timeline_handler.jsx
+++ b/app/assets/javascripts/components/timeline/timeline_handler.jsx
@@ -1,9 +1,7 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import { connect } from 'react-redux';
-import createReactClass from 'create-react-class';
 import PropTypes from 'prop-types';
 import TransitionGroup from '../common/css_transition_group';
-import withRouter from '../util/withRouter';
 import { Route, Routes } from 'react-router-dom';
 import Timeline from './timeline.jsx';
 // import Grading from './grading.jsx';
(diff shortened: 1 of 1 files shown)


Prince0906:
@ragesoss Can you review the changes in the code? I am also converting another file from class-based to functional components. If you find anything unclear or poorly coded, I will modify it and apply the same improvements to the other files.

Prince0906:
Sir, What will be the process for merging this to main branch or is there anything wrong in the file?

ragesoss:
The before/after videos you've included don't really show the functionality of the TimelineHandler component, because there is no timeline content. To test the component, you need to create a timeline. This is most easily done if you switch to `wiki_education: true` mode in application.yml, and then create a new course.

Prince0906:
> The before/after videos you've included don't really show the functionality of the TimelineHandler component, because there is no timeline content. To test the component, you need to create a timeline. This is most easily done if you switch to `wiki_education: true` mode in application.yml, and then create a new course.

Ok, tested it using that, now you can check the screen recording also, has updated it.

app/assets/javascripts/components/timeline/timeline_handler.jsx
```diff
@@ -19,79 +18,65 @@ import {
...
+// Replace state variables with useState hooks
+const TimelineHandler = (props) => {
+  const [reorderable, setReorderable] = useState(false);
+  const [editableTitles, setEditableTitles] = useState(false);
+
+// Replace componentDidMount with useEffect hook
+  useEffect(() => {
+    document.title = `${props.course.title} - ${I18n.t('courses.timeline_link')}`;
+    props.fetchAllTrainingModules();
+  }, [props.course.title, props.fetchAllTrainingModules]);
```
ragesoss, line 32: What is the purpose of passing `props.fetchAllTrainingModules` here?

app/assets/javascripts/components/timeline/timeline_handler.jsx
```diff
@@ -19,79 +18,65 @@ import {
...
+
+  const saveTimeline = () => {
+    setReorderable(false);
+    setEditableTitles(false);
+    const toSave = { weeks: props.weeks };
+    props.persistTimeline(toSave, props.course_id);
+  };
+
+// Replace render method with a separate renderTimelineHandler function
+  const renderTimelineHandler = () => {
```
ragesoss, line 69: Why implement this as a function instead of just returning a JSX node from main functional component?

app/assets/javascripts/components/timeline/timeline_handler.jsx
Prince0906: I am using fetchAllTrainingModules in the useEffect hook to ensure all training modules are loaded when the component initializes. It's included in the dependency array to follow React's best practices, though in practice, this function rarely changes. This approach ensures we have all necessary data available from the start.

app/assets/javascripts/components/timeline/timeline_handler.jsx
Prince0906: yaa, we don't really need a separate function here. I think to simplify this by moving the logic directly into the main component's return statement

Prince0906: > Why implement this as a function instead of just returning a JSX node from main functional component?

I had done that change , you can preview it

PR #5839, opened 2024-06-20
Feature/switch edit view mode in training

omChauhanDev:
## What this PR does
This pr added functionality for switching between Edit and View mode, implemented using 'session storage'.

Tasks :

- [x] Made switch to toggle between Edit and View mode,
- [x] Modified existing test cases accordingly.

### Before:
Main-Training-Page(After)

### After:
image


omChauhanDev:
@ragesoss and @TheTrio 
Kindly have a look.

app/controllers/training_controller.rb
```diff
@@ -68,8 +69,26 @@ def find_slide
     redirect_to "/training/#{training_library.slug}/#{training_module.slug}/#{training_slide.slug}"
   end
 
+  def fetch_training_mode
+    render json: { training_mode: current_training_mode }, status: :ok
```
ragesoss, line 73: Per our discussion today, I think the training mode can be included in the training layout haml file as an attribute of the react root, so that making a separate call just for this piece of data is not necessary.

omChauhanDev: sure, i have done it

PR #5840, opened 2024-06-21
Consistent spelling of username in translatable message

amire80:
Everywhere in MediaWiki, and elsewhere in this app, it's written without a space.


ragesoss:
Thanks @amire80!

PR #5845, opened 2024-06-22
migrated class based components to modern react.

FreakForever:
refactor : converted class based to functional components in details.jsx.


ragesoss:
Please use a more specific PR title, and include before/after screenshots.

PR #5847, opened 2024-06-23
Refactor: Migrate GetHelpButton from Class-Based to Functional Component

codic-yeeshu:
This PR refactors the GetHelpButton component from a class-based component to a functional component. 
Please let me know if there is anything missing as I am new to open source.

before changes

https://github.com/user-attachments/assets/9f1c10d4-3c50-4b51-9d19-3321539e3089

after changes

https://github.com/user-attachments/assets/2a1390a0-79c8-4753-a69e-34fc0cd592e1

### DIFF
--- a/app/assets/javascripts/components/common/get_help_button.jsx
+++ b/app/assets/javascripts/components/common/get_help_button.jsx
@@ -1,5 +1,4 @@
-import React from 'react';
-import createReactClass from 'create-react-class';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { bindActionCreators } from 'redux';
 import { connect } from 'react-redux';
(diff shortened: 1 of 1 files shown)


codic-yeeshu:
@ragesoss sir, while migrating the code to functional components, I encountered a issue, there are two functions named getKey and stop which are not being used in this component so they are causing eslint error. Additionally, another error occurred due to  missing I18n translations because in the original code they don't have used it. Please guide me on that.

ragesoss:
The eslint errors are breaking the build, and you should identify how to fix them. If the functions are not being used now, and they were before, then removing the unused functions should be part of the refactor.

The i18n warnings will not cause the build to fail, and you can ignore those.

Issue #5865, opened 2024-06-28
Revisit specs for `CoursesUsers` in `spec/models/courses_users_spec.rb`


gabina:
### What is happening?
It's not clear (for me) what the "updates only updates cache from tracked revisions" spec in `spec/models/courses_users_spec.rb` is trying to test.

First of all, the spec tries to get `CoursesUsers` record with id = 1 doing `described_class.update_all_caches(described_class.where(id: 1))`. I wonder if we can be sure that the id of the recently created `CoursesUsers` record is 1 (at least, locally) becuase we're not sepcifying the course user id when creating it.

Supposing that we force the `CoursesUsers` id to be 1 during creation, the other point is that we're doing expectations over `course_user` (where `course_user = courses_user`). According to what I see when printing it, it looks like `course_user` is not being updated as expected (even though the print happens after the `update_all_caches` call). However, it gets updated if we reload it.

```
puts course_user.inspect
#<CoursesUsers id: 1, created_at: "2024-06-28 16:59:24.000000000 +0000", updated_at: "2024-06-28 16:59:24.000000000 +0000", course_id: 290, user_id: 267, character_sum_ms: 0, character_sum_us: 0, revision_count: 0, assigned_article_title: "Selfie", role: 0, recent_revisions: 0, character_sum_draft: 0, real_name: "John Smith", role_description: nil, total_uploads: nil, references_count: 0>
puts course_user.reload.inspect
#<CoursesUsers id: 1, created_at: "2024-06-28 16:59:24.000000000 +0000", updated_at: "2024-06-28 16:59:24.000000000 +0000", course_id: 290, user_id: 267, character_sum_ms: 0, character_sum_us: 600, revision_count: 3, assigned_article_title: "Selfie", role: 0, recent_revisions: 0, character_sum_draft: 700, real_name: "John Smith", role_description: nil, total_uploads: 0, references_count: 0>
```
The point is that if we're only un-tracking the `ArticlesCourses` record for `article` (`ArticlesCourses.first.update(tracked: false)`), why would we `expect(course_user.revision_count).to eq(0)`? I think revisions for `talk_page`, `sandbox` and `draft` should be counted and we should `expect(course_user.revision_count).to eq(3)`.

### Additional context
[...]

PR #5866, opened 2024-06-28
Implement `CourseUserWikiTimeslice.update_cache_from_revisions`

gabina:
## What this PR does
This PR implements the `update_cache_from_revisions` instance method for `CourseUserWikiTimeslice` and specs for it.
The behavior for the new method and its specs are based on the `update_cache` instance method for `CoursesUsers`.

In addition, this PR implements class method `update_all_caches_from_timeslices` for `CoursesUsers`, which involves calculating caches for `CoursesUsers` records based on the existing timeslices for them.

Finally, it updates the existing `UpdateCourseStatsTimeslice` class to update caches for `CourseUserWikiTimeslice` and `CoursesUsers`. I adds specs for it.

- As part of this PR, opened issue #[5865](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5865).
- `update_cache_from_revisions` assumes that revisions are already filtered by that specific course user wiki (similar approach I followed for `ArticleCourseWikiTimeslice.update_cache_from_revisions`). We could add some checks later.
- I had to replace a lot of `where` queries. We may want to make some performance improvements in the future.
- `CourseUserWikiTimeslice.total_uploads` is not really implemented yet. We're counting all the uploads for the course user in every run (it's not aggregated by wiki/dates).

### DIFF
--- a/spec/models/article_course_timeslice_spec.rb
+++ b/spec/models/article_course_timeslice_spec.rb
@@ -63,7 +63,7 @@ describe ArticleCourseTimeslice, type: :model do
   end
   let(:subject) { article_course_timeslice.update_cache_from_revisions revisions }
 
-  describe '.update_cache_from_revisions' do
+  describe '#update_cache_from_revisions' do
     it 'updates cache correctly' do
       expect(article_course_timeslice.character_sum).to eq(100)
       expect(article_course_timeslice.references_count).to eq(3)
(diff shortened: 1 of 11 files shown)


spec/models/article_course_timeslice_spec.rb
```diff
@@ -63,7 +63,7 @@
   end
   let(:subject) { article_course_timeslice.update_cache_from_revisions revisions }
 
-  describe '.update_cache_from_revisions' do
+  describe '#update_cache_from_revisions' do
```
gabina, line 66: Not related to this PR. Just noticed that I didn't follow the convention.

PR #5867, opened 2024-06-30
Integrate Redux Toolkit and optimize store configuration

Abishekcs:
# Redux Toolkit Integration

## What this PR does
This PR updates our Redux store configuration to use Redux Toolkit, simplifying our setup and leveraging built-in optimizations.

## Changes
- Replace `createStore` with `configureStore` from Redux Toolkit
- Remove explicit `thunk` middleware import (now included by default)
- Simplify middleware configuration
- Add conditional Redux DevTools enablement
- Remove manual compose enhancers setup

For now, I have simply replaced `createStore` with `configureStore`. This follows the suggestion from the Redux migration guide on the official Redux website, which states:

> "You should always start by replacing the legacy `createStore` call with `configureStore`. This is a one-time step, and all of the existing reducers and middleware will continue to work as-is. `configureStore includes development-mode checks for common mistakes like accidental mutations and non-serializable values`, so having those in place will help identify any areas of the codebase where those mistakes are happening."

For more details, refer to the Redux documentation:
https://redux.js.org/usage/migrating-to-modern-redux#modernizing-redux-logic-with-redux-toolkit

### a) Immutable and Serializable Checks
I have currently set `immutableCheck` and `serializableCheck` (provided by Redux Toolkit) to `false`. This decision was made because state mutations and non-serializable values were detected in our existing code, as shown in the screenshots below:

Screenshot from 2024-06-30 15-51-05
Screenshot from 2024-06-30 16-20-11
Screenshot from 2024-06-30 16-24-16

I chose this approach for the following reasons:
1. It allows us to complete the migration to Redux Toolkit without immediately addressing these issues.
2. With these checks disabled, there are no errors or warnings, and the `dashboard functions as intended`.
3. The development-mode checks provided by Redux Toolkit are temporarily disabled to prevent disruption to our current workflow.

### b) Addressing Mutation and Serialization Issues
An important question to consider is whether we should address these mutation and serialization issues now or in the future, especially given the large number of reducers and actions that need to be checked for state mutations.

For example, we can enable mutation checks when resolving or detecting these issues by setting `enableMutationChecks` to `true`. Then, pick an existing slice reducer and its associated actions, and ensure there is no state mutation (RTK will automatically detect it). Replace them with RTK's `createSlice`. Repeat this process for one reducer at a time. After resolving all the state mutations, we can then `permanently re-enable these checks` i.e (development-mode checks for common mistakes like accidental mutations and non-serializable values).

### c) Redux DevTools in Production
I have also disabled Redux DevTools for production, but I'm not sure if this will be helpful.

Any suggestions regarding these three issues would be greatly appreciated. Thank you.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2022,6 +2022,26 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@reduxjs/toolkit@npm:^2.2.6":
+  version: 2.2.6
+  resolution: "@reduxjs/toolkit@npm:2.2.6"
+  dependencies:
+    immer: ^10.0.3
+    redux: ^5.0.1
+    redux-thunk: ^3.1.0
+    reselect: ^5.1.0
+  peerDependencies:
+    react: ^16.9.0 || ^17.0.0 || ^18
+    react-redux: ^7.2.1 || ^8.1.3 || ^9.0.0
+  peerDependenciesMeta:
+    react:
+      optional: true
+    react-redux:
+      optional: true
+  checksum: 9cd02df5c7f2a40ce9ea41c74a7684a9be2b4adf75c2ad897375fb64b57c039e4fa7f2b74c86588434f795f3ab90b71a14ab6768ace8fd5c6d0153a6ed93ba83
+  languageName: node
+  linkType: hard
+
 "@sentry/browser@npm:^6.18.1":
   version: 6.19.7
   resolution: "@sentry/browser@npm:6.19.7"
@@ -2765,6 +2785,7 @@ __metadata:
     "@babel/register": ^7.18.9
     "@pmmmwh/react-refresh-webpack-plugin": ^0.5.7
     "@rails/ujs": ^7.0.3-1
+    "@reduxjs/toolkit": ^2.2.6
     "@sentry/browser": ^6.18.1
     "@tinymce/tinymce-react": ^3.12.6
     "@wojtekmaj/enzyme-adapter-react-17": ^0.6.6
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -10654,6 +10691,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"redux@npm:^5.0.1":
```
ragesoss, line 10694: This adds redux 5 without also removing redux 4. Is that the expected behavior for this first stage, or should we expect to only use a single version of redux.

ragesoss:
I am not familiar with Redux Toolkit, but it sounds like a good idea to use it and then find and fix the reducers that have state mutations.

yarn.lock
Abishekcs: > This adds redux 5 without also removing redux 4. Is that the expected behavior for this first stage, or should we expect to only use a single version of redux.

We need to keep both versions for now.
The expected behavior when adding Redux Toolkit is indeed to have Redux 5 added without immediately removing Redux 4. However, the next step in the migration process will be to remove the direct dependency on Redux 4 (and other packages like redux-thunk and reselect) from the package.json, as they're now provided by Redux Toolkit. This should be done only after we have `fully migrated to Modern Redux`.

Source of Information: https://redux.js.org/tutorials/fundamentals/part-8-modern-redux#package-cleanup

Abishekcs:
> I am not familiar with Redux Toolkit, but it sounds like a good idea to use it and then find and fix the reducers that have state mutations.

Great! 😁 I am still going through the official[ Redux docs](https://redux.js.org/) and [style guide](https://redux.js.org/style-guide/) for modern Redux. I will try to fix those two mutations as shown in the screenshot of this PR. I will also try to replace the old reducers and actions with Redux Toolkit's createSlice.

PR #5880, opened 2024-07-03
Added functionality for handling duplicate slugs when copying courses

PR4NJ41:
## What this PR does
This PR adds the functionality of handling duplicate slugs when copying courses across servers. 
This works for both the servers.

## Screenshots
Before:
Screenshot 2024-07-03 at 9 17 09 PM

After:

No Such Error Now

### DIFF
--- a/app/services/copy_course.rb
+++ b/app/services/copy_course.rb
@@ -47,7 +47,7 @@ class CopyCourse
   end
 
   def modify_course_slug
-    @course_data['term'] = 'CLONED FROM ' + @course_data['term']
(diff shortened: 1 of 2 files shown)


app/services/copy_course.rb
```diff
@@ -43,6 +46,14 @@ def copy_main_course_data
     @course = Course.create!(copied_data)
   end
 
+  def modify_course_slug
+    @course_data['term'] = 'CLONED FROM ' + @course_data['term']
```
ragesoss, line 50: Let's make this 'COPIED FROM ' instead

PR4NJ41: Done

PR #5887, opened 2024-07-05
Fix: Adjust pagination controls for single-page results and edge cases

### LINKED ISSUE #5873: Articles tab of course page shows 'Previous' and 'Next' buttons when there is only 1 page
### What is happening?

The pagination controls for the Articles Edited list are rendering even when there is only a single page of results. This is a regression, as it used to work properly.

For example: https://dashboard.wikiedu.org/courses/Temple_University/The_Destruction_of_Images_(Spring_2024)/articles/edited?

### Expected behavior

The pagination controls should only render if there is more than 1 page of results.

Abishekcs:
closes #5873 
## What this PR does

This PR addresses the issue where the pagination controls (i.e., 'Previous' and 'Next' buttons) are displayed even when there is only a single page of results in the Articles Edited list.

Additionally, this PR enhances the pagination controls by:

Hiding the 'Previous' button when the user is on the first page.
Hiding the 'Next' button when the user is on the last page

## Changes Made

Updated the controls to ensure 'Previous' is hidden on the first page and 'Next' is hidden on the last page.

## Screenshots
Before:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/3a5fc8ae-6383-4c7a-b7fa-40aeb6e93e1c

After:

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/74126284/31778d4a-d912-4d82-a818-e6e852e9dcf4

### DIFF
--- a/app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
+++ b/app/assets/javascripts/components/articles/PaginatedArticleControls.jsx
@@ -20,6 +20,9 @@ export const PaginatedArticleControls = ({ showMore, limitReached }) => {
   const currentPage = useSelector(state => state.articles.currentPage);
   const pageRange = getPageRange(currentPage, filteredArticles.length);
   const totalFilteredArticles = filteredArticles.length;
+  const hideNextButton = (totalPages === currentPage) ? 'hide-next-btn' : '';
+  const hidePrevButton = (currentPage === 1) ? 'hide-prev-btn' : '';
+  const hasLimitReached = limitReached ? 'hidden-see-more-btn' : '';
 
   // this is for when a filter is applied and the number of pages changes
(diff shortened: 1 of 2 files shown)


ragesoss:
This looks fine, and I'm going to merge it. But I'm curious... did you figure out what caused the regression? It used to work like this without using CSS to hide it.

Abishekcs:
> This looks fine, and I'm going to merge it. But I'm curious... did you figure out what caused the regression? It used to work like this without using CSS to hide it.

No, I couldn't figure out what caused the regression 😅. Also, not sure, but as far as I went through the docs, the react-paginate library doesn't have any built-in feature to hide the prev & next buttons. If I find out what caused the regression, I will resolve it and remove the CSS in a future PR.

PR #5893, opened 2024-07-10
[Data rearchitecture project] Add logic to create articles courses records from revisions in RAM

gabina:
## What this PR does
This PR implements `update_from_course_revisions` class method to create `ArticlesCourses` records from revisions in RAM. It's based on the behavior for the existing `update_from_course` class method.

- TODO: the  `update_from_course_revisions` method only creates new `ArticlesCourses` records. We still have to add logic to remove all the `ArticlesCourses` that do not correspond to course revisions. That may happen if the course dates changed, so some revisions are no longer part of the course. Also remove records for articles that aren't on a tracked wiki. See the full `update_from_course`  definition.
- TODO: revisit view counts calculation when `TimesliceManager` is implemented.

### DIFF
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -19,7 +19,7 @@
 #= ArticlesCourses is a join model between Article and Course.
 #= It represents a mainspace Wikipedia article that has been worked on by a
 #= student in a course.
-class ArticlesCourses < ApplicationRecord
+class ArticlesCourses < ApplicationRecord # rubocop:disable Metrics/ClassLength
   belongs_to :article
   belongs_to :course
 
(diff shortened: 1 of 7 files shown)


app/models/course_data/articles_courses.rb
```diff
@@ -19,7 +19,7 @@
 #= ArticlesCourses is a join model between Article and Course.
 #= It represents a mainspace Wikipedia article that has been worked on by a
 #= student in a course.
-class ArticlesCourses < ApplicationRecord
+class ArticlesCourses < ApplicationRecord # rubocop:disable Metrics/ClassLength
```
gabina, line 22: Doesnt make sense to make a refactor to fix this because this class has a lot of duplicated code that will be deleted in the future

gabina:
Well.. the following specs are failing:

```
rspec ./spec/features/account_requests_spec.rb:17 # Account requests can be enabled by the course facilitator
rspec ./spec/services/update_wiki_namespace_stats_spec.rb:31 # UpdateWikiNamespaceStats adds articles with only tracked namespaces to article_courses
rspec ./spec/services/update_wiki_namespace_stats_spec.rb:67 # UpdateWikiNamespaceStats updates the wiki-namespace stats correctly

```

It's suspicious because `UpdateWikiNamespaceStats` specs are related to UpdateCourseStats class. However, I think I didn't really change anything in `UpdateCourseStats` (only created a new `UpdateCourseStatsTimeslice` class). In addition, specs are passing locally for me and I found that those specs also failed in master [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/actions/runs/9833380295/job/27143516438) . So I guess it's transient and not related to my changes.

PR #5895, opened 2024-07-12
[Data rearchitecture project] Implement timeslice manager

gabina:
## What this PR does
This PR implements (a first version of) the `TimesliceManager` class and specs for it.
The `TimesliceManager` class takes care of creating the timeslices records (`article_course_timeslices`, `course_user_wiki_timeslices` and `course_wiki_timeslices`) for a given course.

After thinking about some alternatives, the approach taken is the one whose logic keeps simplest (I _think_). **Timeslices records are added at once for the entire course duration** (that means, from `course.start` to `course.end`), as article course records are created, new course users join or new course wiki are tracked.
The class provides three instance methods:
1. `create_timeslices_for_new_article_course_records`: given a collection of (new) `ArticlesCourses`, it creates article course timeslices records for each of them. This method should be called right after the  `ArticlesCourses` creation, likely as part of `ArticlesCourses.update_from_course` - the point where `ArticlesCourses` are created.
2. `create_timeslices_for_new_course_user_records`: given a collection of (new) `CoursesUsers`, it creates course user wiki timeslices records for each of them (for every wiki). This method should be called right after the  `CoursesUsers` creation, likely as part of `JoinCourse` class - which I _think_ is the only place where `CoursesUsers` are created (to double-check). 
3. `create_timeslices_for_new_course_wiki_records`: given a collection of (new) `CoursesWikis`, it creates _a)_ course wiki timeslices records for each of them and _b)_ course user wiki timeslices for all existing course user records for the new wikis. This method should be called right after the  `CoursesWikis` creation, likely as part of `Course.update_wikis` - which I _think_ is the only place where `CoursesWikis` are created (to double-check). 

**Advantages of this approach**
- The logic keeps pretty simple.
- Creating several records at once (in batches) is performant.
- Keep synchrony between records should be easy. I mean, it's easy to guarantee that every article course record has all the timeslices records that needs at the moment.

**Disadvantages of this approach**
- It could involve a lot of creation/deletion batch of records in some cases. Examples: course start/end dates change a lot, tracked/untracked wikis are modified frequently. 

**Note:** having the complete universe of timeslices created previously **is not mandatory**. We could work in the future with an approach which creates timeslices on demand. But I was a bit stuck and to start working with timeslices, it was easier for me to start with the simplest operation: creating the complete universe of empty timeslices.

TODO: this first version doesn't take _deletions_ into account. Cases to handle in the next PR:

- Course start/end changes.
- Tracked wikis removal.
- Courses users removal.
- Articles courses updates (see `ArticlesCourses.destroy_invalid_records`).


lib/timeslice_manager.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+#= Adds new ArticleCourseTimeslice, CourseUserWikiTimeslice and CourseWikiTimeslice records.
+class TimesliceManager
```
gabina, line 4: This is the main file.

spec/lib/timeslice_manager_spec.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require "#{Rails.root}/lib/timeslice_manager"
+
+describe TimesliceManager do
```
gabina, line 6: Specs file.

gabina:
I don't like these specs failing, but I can't reproduce the error locally, so I'm not sure if they are related to my changes.
```
Failures:

  1) UpdateWikiNamespaceStats adds articles with only tracked namespaces to article_courses
     Failure/Error: expect(fetched_namespaces).to include(0, 2, 3, 102)
       expected [] to include 0, 2, 3, and 102
     # ./spec/services/update_wiki_namespace_stats_spec.rb:37:in `block (2 levels) in <top (required)>'

  2) UpdateWikiNamespaceStats updates the wiki-namespace stats correctly
     Failure/Error: expect(stats[:edited_count]).to eq 1

       expected: 1
            got: 0

       (compared using ==)
     # ./spec/services/update_wiki_namespace_stats_spec.rb:70:in `block (2 levels) in <top (required)>'
```

ragesoss:
I'm rerunning the build, let's see if it's failing consistently.

PR #5897, opened 2024-07-15
Disable editing of title, school  and term/when fields after event sync

### LINKED ISSUE #5811: The slug of a course can change even after it is synced to Event Center
If the slug changes, it breaks integration with Event Center. See https://phabricator.wikimedia.org/T366462

We should make sure the title/school/term fields are not editable for an event that is synced.

Abishekcs:
closes #5811 

## What this PR does

Prevent changes to the title, school and term/when fields for courses that are already linked to an event to maintain integration stability.

## Screenshots

Before:

https://github.com/user-attachments/assets/c33d4c73-b4c4-4b4a-b8f3-e792e1e8f021

After:

https://github.com/user-attachments/assets/beac4e2b-e96a-467a-a551-c09e031aff66


ragesoss:
Nice!

PR #5903, opened 2024-07-21
[Data rearchitecture] Use timeslice manager to create empty timeslices

gabina:
## What this PR does
This PR uses the `TimesliceManager` class to create empty timeslices and updates `UpdateCourseStatsTimeslice` class to modify the right timeslice.

- Create article course timeslices as part of `update_from_course_revisions` in `ArticlesCourses` using the `TimesliceManagar` class.
- Create course user wiki timeslices when new user joins a course using the `TimesliceManagar` class.
- Create course wiki timeslices and course user wiki timeslices when new course wikis are added to a course using the `TimesliceManagar` class.
- Make `UpdateCourseStatsTimeslice` class update the right timeslice (given the revision date). **Only works for daily timeslices.**
- Schema changes: add `last_mw_rev_datetime` field to `course_wiki_timeslices`. Add unique index for `course_wiki_timeslices`.
- Make `UpdateCourseStatsTimeslice`  fetch revisions based on `last_mw_rev_datetime` and course end date.
- Update `populate_dashboard` script to create course user wiki and course user timeslices.
- TODO: We should fetch data even after the course end to calculate retention. However, right now this causes problems due to lack of timeslices for those days.
- TODO: make `UpdateCourseStatsTimeslice` work independently on the timeslice duration. Right now only works for daily timeslices.


gabina:
Failing specs are not related to PR changes.

Issue #5901, opened 2024-07-18
Prevent `CourseCreationManager` from creating `CoursesUsers` if it does not create such a course


gabina:
### What is happening?
It looks like the `CourseCreationManager` can create courses users records even though the course itself is not created due to the lack of certain course params.

### To Reproduce
This issue is not easily reproducible via the web-app, as the course creation form is robust. However, the `CoursesController` spec reveals this weird behavior.

> Steps to reproduce the behavior:
> 1. Run `rspec spec/controllers/courses_controller_spec.rb`
> 2. The case `CoursesController#create setting slug from school/title/term not all slug params present does not set slug (and does not create course)` doesn't create the course record but still creates the `CoursesUsers` record for the instructor with `course_id` set to `nil`. 
Example: `#<CoursesUsers id: 707, created_at: "2024-07-18 22:45:13.000000000 +0000", updated_at: "2024-07-18 22:45:13.000000000 +0000", course_id: nil, user_id: 745, character_sum_ms: 0, character_sum_us: 0, revision_count: 0, assigned_article_title: nil, role: 1, recent_revisions: 0, character_sum_draft: 0, real_name: nil, role_description: nil, total_uploads: nil, references_count: 0>`

### Expected behavior
`CourseCreationManager.create` should not produce any side effect if the course is not successfully created. 
Specifically, methods such as `add_instructor_to_course` and `add_tags_to_course` should not be invoked if any `before_save` callback on the course model fails.

PR #5913, opened 2024-07-27
Prevent CourseCreationManager side effects if it does not create a course

Kerman07:
Closes: #5901 

## What this PR does
If the CourseCreationManager.create method fails to create a Course don't execute the actions where that course is expected.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -35,10 +35,9 @@ class CoursesController < ApplicationController
       return
     end
     @course = course_creation_manager.create
-    update_courses_wikis
-    update_course_wiki_namespaces
-    update_academic_system
-    update_course_format
+    # return early if the course was not persisted to the db
+    return if @course.id.nil?
+    handle_post_course_creation_updates
   end
 
   def update
@@ -362,6 +361,13 @@ class CoursesController < ApplicationController
       }
       @course.save
     end
+  end
+
+  def handle_post_course_creation_updates
+    update_courses_wikis
+    update_course_wiki_namespaces
+    update_academic_system
+    update_course_format
   end
 
   def course_params
--- a/lib/course_creation_manager.rb
+++ b/lib/course_creation_manager.rb
@@ -40,13 +40,17 @@ class CourseCreationManager
   end
 
   def create
+    # Returns the course object, either as an in-memory object if save fails,
+    # or as a persisted object if save is successful.
     set_passcode
     set_course_type
     set_initial_campaign
-    @course = Course.create(@course_params.merge(@overrides))
-    add_instructor_to_course
-    add_tags_to_course
-    process_experiments
+    @course = Course.new(@course_params.merge(@overrides))
+    if @course.save
+      add_instructor_to_course
+      add_tags_to_course
+      process_experiments
+    end
     @course
   end
 


gabina:
Thanks for taking this issue. Please, notice that the following spec is failing:

```
1) CoursesController#create setting slug from school/title/term not all slug params present does not set slug (and does not create course)
     Failure/Error:
       @course.courses_wikis.each do |course_wiki|
         wiki_domain = course_wiki.wiki.domain
         course_wiki_namespaces = []
         namespaces.each do |ns|
           ns_wiki_domain = ns.split('-', 2)[0]
           next if ns_wiki_domain != wiki_domain
           ns_id = ns.split('-', 3)[2].to_i
           cw_ns = course_wiki.course_wiki_namespaces.find_or_create_by(namespace: ns_id)
           course_wiki_namespaces << cw_ns
         end

     NoMethodError:
       undefined method `courses_wikis' for nil:NilClass

           @course.courses_wikis.each do |course_wiki|
                  ^^^^^^^^^^^^^^
     # ./app/controllers/courses_controller.rb:295:in `update_course_wiki_namespaces'
     # ./app/controllers/courses_controller.rb:39:in `create'
     # ./spec/controllers/courses_controller_spec.rb:384:in `block (5 levels) in <top (required)>'
```

This is probably because before your changes the `CourseCreationManager.create` method was returning always a new instance of `Course` class (maybe not persisted in the database), and now it is returning `nil` when the course is not created. You should be able to run these specs locally on your computer and fix them.

gabina:
Some specs keep failing. Please try to fix them and tag me when ready for review.

Kerman07:
@gabina Not sure if the current failing specs have anything to do with the change.

app/views/courses/create.json.jbuilder
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 json.course do
-  json.call(@course, :title, :description,
-            :start, :end, :school, :term, :subject, :slug)
+  if @course
+    json.call(@course, :title, :description,
+              :start, :end, :school, :term, :subject, :slug)
+  end
```
gabina, line 7: I do understand why you do this, but I would prefer not to change what the course creation endpoint renders at all, as it is difficult to be 100% sure that this will not cause any problems.

I think one option is to make `CourseCreationManager.create` method keep returning a course even though persistence to the database fails (as it's working now), but skipping the side effect actions. We can still determine if the course was created based on the presence of  `@course.id`.
It would be helpful  to add some comments to the `CourseCreationManager.create` method, clarifying that it returns either a newly created course object or a course object with `nil` for the `id` if creation fails.

I understand if you think this proposed solution seems less clean, but I think the key point is to minimize changes.

gabina:
> @gabina Not sure if the current failing specs have anything to do with the change.

Nope, the `wikidata_course_spec` failing is not related to your changes.

Don't forget to [run rubocop locally](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/CONTRIBUTING.md#code-style) on the changed files to be sure there are no offenses.

app/views/courses/create.json.jbuilder
Kerman07: @gabina Ok, I've changed it back to returning the course no matter what happened. Added a comment too.

gabina: Thanks!
I think it’s a good idea to keep the premature `return` you had included in `CoursesController.create` if `@course` was not saved to skip calling `update_courses_wikis`, `update_course_wiki_namespaces`, etc. You can check if @course.id is nil or not to determine this. Do you mind adding it? This should be ready to go after that.

Kerman07: @gabina I don't see where you would return early, I thought we are still returning the course no matter if the save succeeded or not.

gabina: hmm I cannot see your previous commits because you force-pushed. Right now the only change is in `CourseCreationManager.create`. That method is called in `CoursesController.create`. I think we could follow your previous proposal to return earlier if the course was not persisted to the database correctly and make something like:

```
  def create
    require_signed_in

    course_creation_manager = CourseCreationManager.new(course_params, wiki_params,
                                                        params[:course][:scoping_methods],
                                                        initial_campaign_params,
                                                        instructor_role_description, current_user,
                                                        params[:course][:ta_support])
    unless course_creation_manager.valid?
      render json: { message: course_creation_manager.invalid_reason },
             status: :not_found
      return
    end
    @course = course_creation_manager.create
    # return early if the course was not persisted to the db
    return if @course.id.nil?
    update_courses_wikis
    update_course_wiki_namespaces
    update_academic_system
    update_course_format
  end
```

I didn't test it, but I think that doing something like that would make sense. WDYT?

Kerman07: @gabina Yes, that makes sense. I've added that change to the PR.

gabina:
There is a ruby linting error because the method has too many lines. Try to encapsulate some part of it in a private method to fix it and then run [rubocop](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/CONTRIBUTING.md#code-style) locally to be sure it gets fixed. Let me know if you have any trouble.
```
app/controllers/courses_controller.rb:24:3: C: Metrics/AbcSize: Assignment Branch Condition size for create is too high. [<2, 23, 2> 23.17/23]
  def create ...
  ^^^^^^^^^^
app/controllers/courses_controller.rb:24:3: C: Metrics/MethodLength: Method has too many lines. [17/16]
  def create ...
  ^^^^^^^^^^
```

gabina:
Thanks for taking this. I think this PR is ready to go. However, I cannot merge or approve this PR. We'll have to wait for Sage for that.

ragesoss:
Thanks @Kerman07 and @gabina!

PR #5914, opened 2024-07-30
5. Feature/adding slide in training modules

omChauhanDev:
## What this PR does
This pr help user in adding slides to training modules via dashboard.

The logic for adding slides to training modules is dipicted by this flow diagram :
Logical Diagram final

### Before:
Before

#### After:
1. Add slide button
After

2. Modal for adding slide
1  Fillign form

4. Slide added 
2  Slided added

3. Validations : 
 
 - [x] V1 : One slug cannot be taken by two training web pages 
image
 
 - [x] V2 : Multiple addition of same slide is not allowed 
image
 
 - [x] V3 : No field should be empty  
 image


ragesoss:
This works as expected for me. I think it needs to handle the wiki pages a little more flexibly, accept either just the page title or the complete URL, and extracting the page title if a full URL is submitted.

PR #5920, opened 2024-08-02
[Data rearchitecture] Allow deleting tracked course wikis

gabina:
## What this PR does
This PR implements the ability to remove tracked wikis for a course.

When a new tracked wiki is added, we create the given timeslices (already implemented before this PR).

When existing tracked wikis get removed for a course, we want to:
1. Remove existing timeslices for those wikis. This is done through the new `TimesliceManager.delete_timeslices_for_deleted_course_wikis` (commit d943064313035db381b436ea954da58d9bb2bc57)
2. Remove articles courses records for those wikis. This is done through the new `ArticlesCoursesCleanerTimeslice` class (commit 12e024f8b2935ab13087f3691d88a62ed548c562)
3. Update course cache.

These things need to be done through a worker in an asynchronous way, because they can potentially take some time, and we don't want to leave the web-app user waiting. See new `CourseWikiUpdaterWorker` class.

### Delete `es.wikipedia.org`
Before deleting  `es.wikipedia.org` there is one article course for that wiki and 62 edited articles in total.
When deleting the es.wikipedia wiki, an asynchronous task is queued.

[delete_es_wikipedia.webm](https://github.com/user-attachments/assets/6d25147c-c669-40a6-b390-bd8cf22f5b1d)

### Re-add `es.wikipedia.org` once deleted
After the asynchronous task runs, there are 61 edited articles in total for the course, because the one associated to the es.wikipedia was removed. Then, we re-add it through the web-app.

[re_add_es_wikipedia.webm](https://github.com/user-attachments/assets/95fa05ee-eb7b-43ac-bc08-09ed8c83f510)

### Next course update runs
After the next `UpdateCourseStatsTimeslice` task runs again, the es.wikipedia data is ingested again, and the course has 62 articles in total.

image

Note: Article records not used in any other course should be deleted too, but I don't think there is any nice way to do that (it would involve non course-scoped queries).

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -333,12 +333,17 @@ class Course < ApplicationRecord
   def update_wikis(updated_wikis)
     existing_wiki_ids = courses_wikis.map(&:wiki_id)
     new_wikis = updated_wikis.reject { |wiki| existing_wiki_ids.include?(wiki.id) }
-    deleted_wiki_ids = existing_wiki_ids.reject do |wiki_id|
-      updated_wikis.map(&:id).include?(wiki_id)
-    end
+    deleted_wiki_ids = existing_wiki_ids - updated_wikis.pluck(:id) - [home_wiki_id]
+
     update(wikis: updated_wikis)
     ensure_home_wiki_in_courses_wikis
-    TimesliceManager.new(self).update_wikis(new_wikis, deleted_wiki_ids)
+    # Create course wiki timeslice records for new wikis synchronously
+    TimesliceManager.new(self).create_timeslices_for_new_course_wiki_records new_wikis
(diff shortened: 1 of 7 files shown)


gabina:
Failed examples are not related to my changes.

```
rspec ./spec/features/wikidata_course_spec.rb:36 # A Wikidata course loads the labels for Wikidata Q items
rspec ./spec/services/update_wiki_namespace_stats_spec.rb:31 # UpdateWikiNamespaceStats adds articles with only tracked namespaces to article_courses
rspec ./spec/services/update_wiki_namespace_stats_spec.rb:67 # UpdateWikiNamespaceStats updates the wiki-namespace stats correctly
```

PR #5923, opened 2024-08-09
[Data rearchitecture] Fix  `update_last_mw_rev_datetime` logic. 

gabina:
## What this PR does
Update the logic behind `update_last_mw_rev_datetime`. Now `last_mw_rev_datetime` field in `CourseWikiTimeslice` records has the datetime for the latest revision fetched for that timeslice period. The `last_mw_rev_datetime` field is `nil` if no revision was fetched for that period, either because the period hasn't been processed yet or because there were no revisions during that time.


gabina:
`rspec ./spec/features/language_switcher_spec.rb:29 # language_switcher user logged out fallbacks to en for locales with incomplete translations` started to fail on this branch and I should investigate why.

PR #5929, opened 2024-08-15
Fix: Automatically Refresh 'Updated At' Timestamp When Course Stats Are Refetched

### LINKED ISSUE #5925: On course pages, 'Updated at' does not automatically refresh when stats are re-fetched
On the Home tab of a course page, the frontend will refetch course data once per minute. This results in automatically updating the course stats (which is nice during a live event, as leaving the page open lets you see the stats as they increase throughout the event). This is configured in the `details.jsx` component.

When the course data is refetched, however, it does not update the StatisticsUpdateInfo component, which displays the time of the last update and the next expected update... so the course stats may be actually refreshed and based on new update, while this component still says 'last update 1 day ago' or similar.

This is probably a pretty easy fix via one of the Redux reducer functions.

Abishekcs:
closes #5925 

## What this PR does

Fix: Automatically Refresh 'Updated At' Timestamp When Course Stats Are Refetched

## Cause of Bug

The `RECEIVE_COURSE_UPDATE` case in the `course reducer` was not taking in to account to update the update_logs field, even when the new data had updates for update_logs. This led to the old update_logs remaining unchanged in the Redux state.

## Screenshots

Before:

https://github.com/user-attachments/assets/5bd1ee30-1a24-45e6-9706-8bf76e21f2ea

After:

https://github.com/user-attachments/assets/1eed71e5-898a-41f4-8f6c-30d95cc23f49

### DIFF
--- a/app/assets/javascripts/reducers/course.js
+++ b/app/assets/javascripts/reducers/course.js
@@ -52,7 +52,15 @@ export default function course(state = initialState, action) {
       const courseData = action.data.course;
       const newStats = CourseUtils.newCourseStats(state, courseData);
       const newKeys = CourseUtils.courseStatsToUpdate(courseData, newStats);
-      return { ...state, ...newKeys, newStats };
+
+      return {
+        ...state,
+        ...newKeys,
+        newStats,
+        flags: {
+          ...courseData.flags,
+        },
+      };
     }
     case PERSISTED_COURSE:
       return { ...state, ...action.data.course };


app/assets/javascripts/reducers/course.js
```diff
@@ -52,7 +52,17 @@ export default function course(state = initialState, action) {
...
       const newStats = CourseUtils.newCourseStats(state, courseData);
       const newKeys = CourseUtils.courseStatsToUpdate(courseData, newStats);
-      return { ...state, ...newKeys, newStats };
+      const refreshedLogs = CourseUtils.refreshCourseUpdatedAtTimestamp(state, courseData);
+
+      return {
+        ...state,
+        ...newKeys,
+        newStats,
+        flags: {
```
ragesoss, line 61: This seems more complicated than it needs to be. Maybe it would be better to simply replace the entire `flags`?

I think the reason we limit the other changes to just the stats is to avoid overwriting unsaved changes that the user might be making to the course details. But I think the flags data can't be directly changed on the frontend, so it should be safe to replace the entire flags object with whatever is received from the server.

Abishekcs: > Maybe it would be better to simply replace the entire `flags`?

Okay, I will implement this. It sounds much simpler.

ragesoss: Thanks! I assume you confirmed it works like this?

Abishekcs: > Thanks! I assume you confirmed it works like this?

Yes, it works properly. I have attached a screen recording below.


https://github.com/user-attachments/assets/4cd2c741-7e1b-44b1-94b2-388e5eb7b7a5

PR #5938, opened 2024-08-23
Optimize course data refetch: Trigger only on Dashboard tab focus

### LINKED ISSUE #5926: Only re-fetch course data when Dashboard tab is focused
A related issue: #5925

### Current Behavior:
In `details.jsx`, the frontend will refetch course data every minute as long as the Details component is mounted (ie, when the Home tab of a course page is open).

### Desired Behavior:
To minimize useless API calls and server load, this refetch behavior should only be active while the dashboard is in a focused tab. There's little point in refetching the data if it's not currently visible to the user.

Abishekcs:
closes #5926

## What this PR does

Only re-fetch course data when Dashboard tab is focused.

## Screenshots

Before (Tab in-active for 3 minutes ):

https://github.com/user-attachments/assets/100aaa99-1f0f-466d-b59c-afe0a0df6496

After (Tab in-active for 3 minutes):

https://github.com/user-attachments/assets/8cf3d288-96a3-4a54-856e-22f0a2355aa3

### DIFF
--- a/app/assets/javascripts/components/overview/details.jsx
+++ b/app/assets/javascripts/components/overview/details.jsx
@@ -75,6 +75,7 @@ const Details = createReactClass({
       clearInterval(this.timeout); // End it
     }
   },
+
   updateDetails(valueKey, value) {
     const updatedCourse = this.props.course;
     updatedCourse[valueKey] = value;
@@ -127,7 +128,9 @@ const Details = createReactClass({
 
   poll() {
     return setInterval(() => {
-      if (this.state.updateCount > MAX_UPDATE_COUNT) { return; }
+      const documentVisible = document.hidden;
+
+      if (documentVisible || (this.state.updateCount > MAX_UPDATE_COUNT)) { return; }
       if (!this.props.editable) {
         this.props.refetchCourse(this.props.course.slug);
         this.setState({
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/components/overview/details.jsx
```diff
@@ -127,7 +128,9 @@ const Details = createReactClass({
 
   poll() {
     return setInterval(() => {
-      if (this.state.updateCount > MAX_UPDATE_COUNT) { return; }
+      const documentVisible = document.hidden;
```
ragesoss, line 131: This is confusing. I would expect `document.hidden` to be true when the page is not focused, but `documentVisible` suggests the opposite. Is this a mistake, or a quirk of `document`? If it's the latter, I suggest using a different constant name and/or a comment to explain how it works.

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -127,7 +128,9 @@ const Details = createReactClass({
...
+
+      if (documentVisible || (this.state.updateCount > MAX_UPDATE_COUNT)) { return; }
```
ragesoss, line 133: Oh, I see, it's part of the guard clause, so it will return without re-fetching if the document is not focused. In that case, maybe rename it `documentNotFocused` or `documentHidden`?

PR #5939, opened 2024-08-26
Fix disruptive pile-up of success notifications

### LINKED ISSUE #5927: When adding many users to a course, the pile-up of success notifications is disruptive
Some kinds of notifications on course pages can accumulate and block the view of content. One common type that has this behavior is the success notifications from adding a new editor to a course, and each one must be dismissed individually (or it will block content, as they stack up).

I'm not sure what the best solution to this is. One good one might be to automatically dismiss earlier notifications of the same type, so that only a certain number (eg, 3) will be present. I'd guess that this could mostly or completely be implemented within the notifications reducer.

Abishekcs:
closes #5927

## What this PR does

Implement max limit of 3 notifications per type (error/success) to prevent pile-up of `success/error` notifications.

## Screenshots

**Before:**

For Error Notification:

https://github.com/user-attachments/assets/90423a6e-42aa-4ad8-b6fe-77d1cb9705a1

For Success Notification:

https://github.com/user-attachments/assets/8960cb33-4e98-47c3-91d8-54c4ff8c224e

**After:**

For Error Notification:

https://github.com/user-attachments/assets/b36aa13f-eb2c-4dfe-a0ef-d56557178a21

For Success Notification:

https://github.com/user-attachments/assets/d9ebc4d8-8a65-4297-9dab-e1350caf0141

### DIFF
--- a/app/assets/javascripts/reducers/notifications.js
+++ b/app/assets/javascripts/reducers/notifications.js
@@ -41,9 +41,27 @@ const handleErrorNotification = function (data) {
 export default function notifications(state = initialState, action) {
   switch (action.type) {
     case ADD_NOTIFICATION: {
-      const newState = [...state];
-      newState.push(action.notification);
-      return newState;
+      const newState = [...state, action.notification];
+
+      // Set the maximum number of notifications allowed for each type
+      const maxNotifications = 3;
+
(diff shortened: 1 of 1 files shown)


ragesoss:
nice!

PR #5944, opened 2024-08-27
[Data rearchitecture] No longer create course user wiki timeslices for non-students

gabina:
## What this PR does
This PR fixes a bug that caused duplicated course user wiki timeslices for the same course, wiki, user, and dates. The issue occurred when there were multiple course user rows for the same user ID but with different roles. This duplication led to repeated references, revision counts, etc., for the user. The solution is straightforward: we no longer create course user wiki timeslices for non-student course users (those with a role different from 0).

I was able to reproduce the error and test the fix locally.

Note: Some changes in commit 548061cd7fffb87347e2f3e56a97ce17b58f8b3c address a previous bug (a missing reload) that became apparent only when adding the` .where(role: CoursesUsers::Roles::STUDENT_ROLE)` query in the timeslice manager.

### DIFF
--- a/app/services/join_course.rb
+++ b/app/services/join_course.rb
@@ -89,7 +89,10 @@ class JoinCourse
     # Do not try to create course user wiki timeslices if there is no course id
     # This should never happen in production but we need this check because otherwise
     # there are specs that fail.
-    create_course_user_wiki_timeslices course_user unless course_user.course_id.nil?
+    # Only create course user wiki timeslices for students.
+    if student_role? && course_user.course_id.present?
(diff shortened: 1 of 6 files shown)


gabina:
Failing spec doesn't seem related to changes in this PR:

```
  1) Student users visiting the ?enroll=passcode url works even if a student has never logged in before
     Failure/Error: click_link 'Students'

     Capybara::ElementNotFound:
       Unable to find link "Students"
     # ./spec/features/student_role_spec.rb:231:in `block (3 levels) in <top (required)>'
```

PR #5947, opened 2024-09-02
[Data rearchitecture] support full timeslice updates

gabina:
## What this PR does
This PR provides an easy way to make a full course stats timeslice update. Manually set `needs_update` course field to true and then use the `manual_update_timeslice` route. This will fetch revisions from the course start and overwrite timeslices.

### DIFF
--- a/app/services/update_course_stats_timeslice.rb
+++ b/app/services/update_course_stats_timeslice.rb
@@ -22,11 +22,11 @@ class UpdateCourseStatsTimeslice
     # it could be because the dates or other paramters were just changed.
     # In that case, do a full update rather than just fetching the most
     # recent revisions.
-    # @full_update = full || @course.needs_update
+    @full_update = @course.needs_update
 
     @start_time = Time.zone.now
     import_uploads
-    timeslice_errors = UpdateCourseWikiTimeslices.new(@course).run
+    timeslice_errors = UpdateCourseWikiTimeslices.new(@course).run(all_time: @full_update)
     update_categories
     update_article_status if should_update_article_status?
     update_average_pageviews
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -18,11 +18,19 @@ class UpdateCourseWikiTimeslices
   end
 
   def run(all_time:)
+    clean_timeslices if all_time
     fetch_data_and_process_timeslices_for_every_wiki(all_time)
     error_count
   end
 
   private
+
+  # Delete and recreate timeslices
+  def clean_timeslices
+    wiki_ids = @course.wikis.pluck(:wiki_id)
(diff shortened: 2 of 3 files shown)


gabina:
Failing specs don't seem related to these changes but I should take a look at them:

```
1) LiftWingApi#get_revision_data fetches json from api.wikimedia.org for wikipedia
     Failure/Error: expect(subject0.dig('829840085', 'wp10').to_f).to eq(29.15228958136511656)

       expected: 29.152289581365117
            got: 29.15228958136513

       (compared using ==)
     # ./spec/lib/lift_wing_api_spec.rb:43:in `block (4 levels) in <top (required)>'
     # ./spec/lib/lift_wing_api_spec.rb:35:in `block (3 levels) in <top (required)>'

  2) RevisionDataManager#fetch_revision_data_for_course fetches all the revisions that occurred during the given period of time
     Failure/Error: expect(revisions[0].wp10).to eq(18.287608774878528)

       expected: 18.287608774878528
            got: 18.28760877487853

       (compared using ==)
     # ./spec/lib/revision_data_manager_spec.rb:39:in `block (4 levels) in <top (required)>'
     # ./spec/lib/revision_data_manager_spec.rb:29:in `block (3 levels) in <top (required)>'

  3) RevisionScoreApiHandler when the wiki is available only for LiftWing API #get_revision_data returns completed scores if retrieves data without errors
     Failure/Error: expect(subject.dig('144495298', 'features')).to be_a(Hash)
       expected nil to be a kind of Hash
     # ./spec/lib/revision_score_api_handler_spec.rb:99:in `block (5 levels) in <top (required)>'
     # ./spec/lib/revision_score_api_handler_spec.rb:87:in `block (4 levels) in <top (required)>'

```

ragesoss:
I think those failing specs are the result of changed rounding behavior in LiftWing. A different set of specs broke in the same way a few weeks ago, which I fixed like this: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/000b98b98a47d34d018f7ab2d75fa0cc44af623c

I'm guessing that LiftWing was caching old values until now, which is why these specs didn't also break at the same time.

gabina:
> I think those failing specs are the result of changed rounding behavior in LiftWing. A different set of specs broke in the same way a few weeks ago, which I fixed like this: [000b98b](https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/000b98b98a47d34d018f7ab2d75fa0cc44af623c)
> 
> I'm guessing that LiftWing was caching old values until now, which is why these specs didn't also break at the same time.

Interesting. The specs are not failing consistently (for example, the specs for PR #5950 passed after this fail), so maybe there is some cache process that changes rounding behavior someway. Anyway, I don't think that number of decimal places makes a lot of sense in a test. I'll open a PR in master to fix them.

PR #5951, opened 2024-09-05
Fix admin notes accessibility

### LINKED ISSUE #5941: Admin Notes feature is not accessible via screen reader
Here's the bug report from someone who uses a screen reader:

"I just tried to create a note on a dashboard course page and not sure if it's an accessibility issue but I didn't see how to actually input any text. I clicked on the admin notes button and didn't see where to write anything. I saw text that said create new note but it wasn't anything I could click on."

(These sorts of issues can often be fixed by using patterns from other parts of the Dashboard UI, typically by changing to a more accessible HTML tag and/or adding appropriate metadata to the element.)

Abishekcs:
closes #5941 

## What this PR does

- Fixes  accessibility issue in the Admin Note feature for screen readers.

- Other minor Fix:
  - Fixed an issue where the note input for title and text would mistakenly close when the user clicked the "Post Note" button, if one or both input areas were empty.
  - Fixed an issue where note creation did not close automatically if it was opened when the admin note panel was closed.

## Cause of Bug

The cause of the bug was that elements meant to be button tags were incorrectly implemented as span tags.😅
  

 

## Screenshots

Before:

https://github.com/user-attachments/assets/e82a5a8c-67bd-4e2e-9bc3-d00edbcbbd77

Before (Other minor Fix):

https://github.com/user-attachments/assets/3d2825c6-4249-45bf-a1ac-369d3d132ae1

After:

https://github.com/user-attachments/assets/78f2870c-5d89-418d-80d4-9aa127777e7e

After  (Other minor Fix):

https://github.com/user-attachments/assets/559ab1b0-12a2-4804-bd39-980277f41ad6

`Note:`

`Screen reader used:` Orca Screen Reader (Default Settings)

`Operating system:` Ubuntu 22.04.4 LTS

`Broswer`: Version 127.0.6533.99 (Official Build) (64-bit)


config/locales/en.yml
```diff
@@ -1838,3 +1838,21 @@ en:
     admin:
       button_text: "Admin Notes"
       header_text: "Admin Notes Panel"
+    screen_reader:
```
ragesoss, line 1841: Since these are all specific to the admin notes, maybe this set of messages should be nested under `admin`? Also, `aria_label` or `aria` might be a better key than `screen_reader`, to more precisely indicate how these messages are used.

Abishekcs: Thank you for the review! I will make the necessary changes.

Issue #5952, opened 2024-09-05
Improve `tagged_courses` view


ragesoss:
The 'tagged_courses' view is available to admins. In `wiki_education` mode, it is used by Wiki Education staff to see a list of courses that share a given `Tag`.

Currently, the tagged_courses view recycles the view templates for the campaign programs view. It should probably be forked into a separate set of view templates, before implementing each of these. (Forking the templates should be done in the same PR as the first item. Additional PRs can address the other items afterwards.)

- [ ]  Add a column to display which Wiki Expert(s) is assigned to each course. (In implementing this, take care to avoid N+1 queries.)
- [ ] Add a stats summary to the top of the page, in the same way as on the Campaign views. This will require changes to the React app; try to make the necessary changes as minimal as possible.
- [ ] Add an option to download a CSV with equivalent data to the table in the tagged_courses view.

Abishekcs:
I'm working on this 😃.

ragesoss:
I think a button with a text label would be better here.

ragesoss:
This is an admin-only view, so being verbose and clear is much preferred. Admins easily get used to the amount of visual information, and filter out what they don't need, but for infrequently-used features like this, having a clear text label will be really useful for the times when it does get used.

ragesoss:
(Also, a button without a new icon helps keep the code concise.)

PR #5959, opened 2024-09-13
Improve 'tagged_courses' view and add Wiki Expert column

Abishekcs:
#5952

## What this PR does

This PR forks the `tagged_courses view` from the existing campaign programs view to allow for further customizations specific to tagged courses. The main feature added in this PR is a new column that displays the `Wiki Expert(s)` assigned to each course and to avoid N+1 queries on how Wiki Experts are retrieved for each course.

Addresses the `first point` of the issue regarding the addition of a Wiki Expert column in the tagged_courses view. This change is part of a larger enhancement of the tagged_courses view used by Wiki Education admin to manage courses based on tags.

## Screenshots

Before:

https://github.com/user-attachments/assets/d49ad091-f206-47d4-a4a7-f4118a0ab23c

After:

https://github.com/user-attachments/assets/6377a14d-25f2-42f5-bcc1-0213faaa48e8

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -36,7 +36,6 @@ Metrics/ClassLength:
     - 'app/controllers/courses_controller.rb'
     - 'app/controllers/users_controller.rb'
     - 'app/controllers/assignments_controller.rb'
-    - 'app/presenters/courses_presenter.rb'
     - 'lib/course_creation_manager.rb'
     - 'lib/importers/revision_score_importer.rb'
     - 'lib/wizard_timeline_manager.rb'
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -36,6 +36,7 @@ Metrics/ClassLength:
     - 'app/controllers/courses_controller.rb'
     - 'app/controllers/users_controller.rb'
     - 'app/controllers/assignments_controller.rb'
+    - 'app/presenters/courses_presenter.rb'
     - 'lib/course_creation_manager.rb'
     - 'lib/importers/revision_score_importer.rb'
     - 'lib/wizard_timeline_manager.rb'
--- a/app/controllers/tagged_courses_controller.rb
+++ b/app/controllers/tagged_courses_controller.rb
@@ -9,7 +9,6 @@ class TaggedCoursesController < ApplicationController
   def articles
     set_page
     set_courses_and_presenter
-    render 'tagged_courses/articles'
   end
 
   def alerts
@@ -26,7 +25,6 @@ class TaggedCoursesController < ApplicationController
     set_page
     set_courses_and_presenter
     load_wiki_experts
-    render 'tagged_courses/programs'
   end
 
   private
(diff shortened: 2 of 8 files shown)


app/presenters/courses_presenter.rb
```diff
@@ -164,4 +165,26 @@ def wikidata_stats
   def creation_date
     I18n.l campaign.created_at.to_date
   end
+
+  # Returns the wiki expert username for the given course, if available
+  def assigned_wiki_expert_for_course(course_id)
```
ragesoss, line 170: I'd prefer for the new code for this *not* to live in CoursesPresenter. Since this is only being used for tagged courses, it could be done within the controller.

ragesoss:
Overall, this looks good. Happy to merge once the new setup code is moved out of the presenter.

Abishekcs:
I have made the necessary changes and moved the logic to load the wiki expert to the controller.


https://github.com/user-attachments/assets/901990dd-7ef2-4d01-a98c-64fa271c024f

.rubocop.yml
```diff
@@ -36,6 +36,7 @@ Metrics/ClassLength:
     - 'app/controllers/courses_controller.rb'
     - 'app/controllers/users_controller.rb'
     - 'app/controllers/assignments_controller.rb'
+    - 'app/presenters/courses_presenter.rb'
```
ragesoss, line 39: This is no longer necessary, correct?

Abishekcs: > This is no longer necessary, correct?

Yes, I have just removed it just now.

app/controllers/tagged_courses_controller.rb
```diff
@@ -41,9 +42,21 @@ def set_tag
...
     @courses = Tag.courses_tagged_with(@tag)
-    # This fake Campaign is so that we can reuse campaigns views
-    @campaign = Campaign.new(slug: @tag, title: "Tag: #{@tag}")
     @presenter = CoursesPresenter.new(current_user:, tag: @tag,
                                       courses_list: @courses, page: @page)
   end
+
+  # Loads CoursesUsers records with role 4 and filters by wiki experts, avoiding N+1 queries
+  def load_wiki_experts
+    return @wiki_experts if @wiki_experts # Avoid re-loading if already loaded
```
ragesoss, line 51: In what situation will @wiki_experts already be loaded?

app/controllers/tagged_courses_controller.rb
```diff
@@ -41,9 +42,21 @@ def set_tag
...
+
+    course_ids = @courses.pluck(:id)
+    wiki_experts_set = SpecialUsers.special_users[:wikipedia_experts]&.to_set
+
+    @wiki_experts = CoursesUsers
+                    .where(course_id: course_ids, role: 4)
+                    .includes(:user)
+                    .select { |course_user| wiki_experts_set.include?(course_user.user.username) }
```
ragesoss, line 59: I think this could be done a little bit more efficiently, by using SpecialUsers.wikipedia_experts to the user records.

There's also no need to pluck IDs for the courses; you can instead do `CoursesUsers.where(course: @courses, user: SpecialUsers.wikipedia_experts, role: ROLE)`

ragesoss (changes requested):
This looks pretty solid, but I thin it can be refactored to be a little cleaner, per inline comments.

app/controllers/tagged_courses_controller.rb
Abishekcs: Actually, this wasn't required. I was using it to test something and forgot to remove it.

Abishekcs:
@ragesoss, I have made the required changes. Could you please take a look? Thank you again for your review!"


https://github.com/user-attachments/assets/d426a667-ef60-483a-a36b-461812c26a5a

PR #5966, opened 2024-09-25
Add Test  for the assignments reducer

Formasitchijoh:
## What this PR does

This PR addresses [#2082](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/2082) by adding comprehensive tests for the assignments reducer.

## Files changed

-  Created a new file test/reducers/assignments.spec.js that covers a wide range of cases, including adding, updating, and deleting assignments. These tests ensure that all cases are handled correctly without breaking existing functionality.

- Made adjustments to the assignments.js reducer to properly handle the case where an assignment being updated is not found in the state

## Test
where ever the assignment reducer was used it was  effectively tested so that is does not break existing code.

@ragesoss, @abhishekbvs , @gabina ,   When you have a chance, could you please review this? I'm happy to make any adjustments and ready to move on to the next reducers. or any other pressing issue,


Formasitchijoh:
@ragesoss, I’d like to clarify something regarding the third test in this case. When multiple dispatches occur with the same data, should the sortKey be set to null for subsequent dispatches?

Currently, the sortByKey function nullifies the sortKey if it matches the previous key, which prevents it from being retained in the state. Should the sortKey remain null for these repeated dispatches, or should it persist for further operations?

Formasitchijoh:
Hello @ragesoss  just want to check in on this PR, Please let me know if there’s anything else I can do to help with the review process or if any changes are needed

ragesoss:
> Hello @ragesoss just want to check in on this PR, Please let me know if there’s anything else I can do to help with the review process or if any changes are needed

Thanks! I'll take a look at it today.

ragesoss:
> @ragesoss, I’d like to clarify something regarding the third test in this case. When multiple dispatches occur with the same data, should the sortKey be set to null for subsequent dispatches?
> 
> Currently, the sortByKey function nullifies the sortKey if it matches the previous key, which prevents it from being retained in the state. Should the sortKey remain null for these repeated dispatches, or should it persist for further operations?

Great question. The `sortByKey` function is a utility function we use for all the frontend sorting operations that we do on data corresponds to database records. It can take a sort key, along with a fallback default sort key. We use the sortKey behavior so that you can use UI controls to sort by a property and then reverse the sorting by passing the same key again. However, in this particular reducer, it only gets the RECEIVE_ASSIGNMENTS action with data from a single server endpoint, which does not have a sortKey, so the value is actually `null` in all cases for this reducer.

app/assets/javascripts/reducers/assignments.js
```diff
@@ -35,6 +42,10 @@ export default function assignments(state = initialState, action) {
     }
     case UPDATE_ASSIGNMENT: {
       const updatedAssignment = action.data.assignment;
+      const assignmentExists = state.assignments.some(assignment => assignment.id === updatedAssignment.id);
```
ragesoss, line 45: I think this may introduce a behavior change.

The UPDATE_ASSIGNMENT action is used for two things: "claiming" an assignment that is on the Available Articles list, and updating the sandbox url for an already-claimed assignment. The latter ought to have the same assignment ID, but the former may return a new different assignment ID. On the server, AssignmentManager uses the course's "retain available articles" setting to determine whether to update the existing assignment to associate it with a user (thereby moving it from the Available Articles list to the Assigned Articles list), or to create a duplicate to associated with the user while leaving the existing Assignment record unchanged (and still available for another user to choose from the Available Articles list).

So if the updatedAssignment is a newly-created duplicate, it will have an ID that is not already present in the state.

Formasitchijoh: Thank you for the clarification! I will update this and let you know

Formasitchijoh:
@ragesoss  I have updated the changes you requested, and also made some changes to the  alert_actions.spec.js test because it was failing, I would appreciate your feedback on this change and how to improve it.

PR #5969, opened 2024-10-01
Implement Alert  for Wiki Expert when a course has students joining before the start date

### LINKED ISSUE #5963: Generate an Alert and send it to Wiki Expert when a course has students joining before the start date
In some cases, students join a course and begin making edits before the listed 'Start' date of the course. Most of the time, that means that we should adjust the start date to match reality more closely and ensure that the early edits made by students are counted properly.

The easiest way to catch these cases will be to generate an instance of a new Alert type when the first student joins a course but the course hasn't started yet. The Alert should email the wiki experts for that course, so they can handle the situation.

### Additional context:
This should only apply to ClassroomProgramCourse type courses.

Abishekcs:
closes #5963

## What this PR does

This PR generates an alert and sends an email notification to the respective Wiki Experts when students join a `ClassroomProgramCourse` before its official start date.

## Screenshots

Screenshot from 2024-11-16 21-14-32


app/controllers/self_enrollment_controller.rb
```diff
@@ -39,6 +41,12 @@ def enroll_self
 
   private
 
+  def check_early_student_join
+    return unless @course.is_a?(ClassroomProgramCourse) && @course.start.to_date > Time.zone.today
+
+    WikiExpertNotificationAlert.new(course: @course)&.send_email
```
ragesoss, line 47: I think we also want to avoid  sending an unlimited number of these alerts, if many students join at once.

One single alert per course is probably fine, so I suggest an additional guard  clause that checks for an existing alert, before creating a new one.

ragesoss (changes requested):
Apologies for the delay in reviewing this. Overall, it looks good. Just a few adjustments needed, and then I'll test it locally.

app/controllers/self_enrollment_controller.rb
Abishekcs: Okay

Abishekcs:
> Apologies for the delay in reviewing this. Overall, it looks good. Just a few adjustments needed, and then I'll test it locally.

No problem😃! I've made the required changes. Could you kindly review them when you have time? Thank you!

Screenshot from 2024-11-15 18-46-26

app/models/alert_types/wiki_expert_notification_alert.rb
```diff
@@ -0,0 +1,43 @@
...
+#  email_sent_at  :datetime
+#  created_at     :datetime         not null
+#  updated_at     :datetime         not null
+#  message        :text(65535)
+#  target_user_id :integer
+#  subject_id     :integer
+#  resolved       :boolean          default(FALSE)
+#  details        :text(65535)
+
+class WikiExpertNotificationAlert < Alert
```
ragesoss, line 22: I should have noticed this in my first pass, but... this is not a good name for this alert. It should be specific to the situation it's designed around. I suggest EarlyEnrollmentAlert (and a similar name for the mailer).

app/models/alert_types/wiki_expert_notification_alert.rb
```diff
@@ -0,0 +1,43 @@
...
+    WikiExpertNotificationMailer.send_email(self)
+    update(email_sent_at: Time.zone.now)
+  end
+
+  def url
+    course_url
+  end
+
+  def main_subject
+    "Alert: A Student Joined Early in Course #{course.title}"
```
ragesoss, line 41: I think "EarlyEnrollmentAlert: #{course.slug}" would be good for the subject line of the email.

Abishekcs:
Screenshot from 2024-11-16 21-14-32

app/controllers/self_enrollment_controller.rb
```diff
@@ -39,6 +41,13 @@ def enroll_self
 
   private
 
+  def check_early_student_join
+    return unless @course.is_a?(ClassroomProgramCourse) && @course.start.to_date > Time.zone.today
+    return if Alert.exists?(course_id: @course.course_id, type: 'EarlyEnrollmentAlert')
+
+    EarlyEnrollmentAlert.new(course: @course)&.send_email
```
ragesoss, line 48: I think using `.new` will initialize the object, but will not save it in the database (so we'd still send out a new email each time it happens). This should probably be `.create` instead.

Abishekcs: I have made the required changes and also fixed a typo error. Thank you for the code review 😃.

Also, I didn’t know why .new was working for me locally 😅, so I’m leaving an explanation below in case it’s needed to clarify why it was working previously 😅.

> I initially thought .new was saving the object to the database, but I didn’t understand why it seemed to work that way. After some research, I realized it was actually because of the .update method.
> 
> Here’s why it worked and why it might not always work:
> 
> If the model doesn’t have strict validations, Rails allows .update to save the object, which makes it look like .new is saving it. However, this isn't how Rails is designed to work, and it might break if validations or other changes are added in the future.

For .new i.e saved by .update later on (Previous Implementation):





For .create and .update (Current Implementation):

ragesoss: Ah, that makes sense, it was updating (and saving for the first time) when it sends the email, for the email_sent_at timestamp.

PR #5972, opened 2024-10-02
Enable javascript style linting rules 

claire256:
https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1408#issue-259228085

## What this PR does
The pull request enables some of the javascript ESLint rules that were disabled. The ESLint worked on include Indent , no-use-before-define, Object-shorthand etc

Before:
The javascript ESLint were not functioning and commented out.

After:
The ESLint were re-enabled and the violations in the codebase fixed.


ragesoss:
For changes like this, I would prefer each rule be separated into its own PR.

ragesoss:
However, no need to redo this PR. The changes look safe, and if it passes CI I can merge it.

ragesoss:
Looks like there are a number of linting errors with this change.

claire256:
Hello @ragesoss , i didn't separate the rules because in the issue you mentioned that it should be a few rules per pr and one rule per commit, but i get you.

ragesoss:
Thanks. (I updated the issue description to indicate that I'd prefer one-rule-per-PR; I guess my preferences have changed since I wrote it initially!)

claire256:
Hi @ragesoss 
- I agree with the one rule per PR.
- So i have run lint-non-build, coverage etc but i'm not getting any linting errors.
- I don't seem to have the file that is failing the CI build here.
So what do you suggest i should do?

ragesoss:
I don't know why `yarn lint-non-build` isn't working locally; it doesn't seem to do anything for me either.

Which file are you talking about? There are multiple files that are triggering the errors in CI, but they are files in the repo.

I suggest closing this PR, opening a new one with just one rule (not the indent rule, which I think we should probably not enable) and fixing all the violations.

Issue #5948, opened 2024-09-03
User#can_edit? may prevent editing by a user who has multiple roles in a course


ragesoss:
### What is happening?
The facilitator for this event, who is also participating as an editor, reports an permissions error when trying to edit the tracked wikis: https://outreachdashboard.wmflabs.org/courses/Wikimedians_of_Kerala/Wiki_Loves_Onam_2024/home

I think this is because of the implementation of `User#role`, which looks for a single CoursesUsers record, so if the wrong one is found the user will be treated as a 'student' rather than 'instructor'.

### Additional context

A fix for this should include a unit test for `#can_edit?` that fails with the current implementation but passes with a fixed implementation.

TunrayoIlawole:
I am currently trying to set up my development environment @ragesoss. I will get started on the issue as soon as i get it running

TunrayoIlawole:
Hi @ragesoss , please I am having trouble finding where the unit tests are. Can you help point me to the directory?

Formasitchijoh:
Hi @TunrayoIlawole which unit test are you looking for?

TunrayoIlawole:
Thank you @Formasitchijoh

crocmons:
I would like to work on this! @ragesoss

ragesoss:
@crocmons someone else is actively working on this already. I recommend choosing a different issue.

PR #5976, opened 2024-10-03
User#can_edit? fix

TunrayoIlawole:
## What this PR does
This PR addresses #5948 by ensuring users with multiple roles are correctly assigned editing permissions.

## More details on what was done

- Updated User#role method to properly handle cases where a user has multiple roles in a course.
- The method now checks if the user has any editing role and returns the appropriate role.
- Added unit test to verify that users with both campus volunteer (non-editing role) and instructor roles are correctly recognized as having editing privileges.

### DIFF
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -49,7 +49,7 @@ class AssignmentsController < ApplicationController
 
   # Select an Available Article as a new Assignment for a user
   def claim
-    @claimed_assignment = Assignment.find(params[:id])
+    @claimed_assignment = Assignment.find(params[:assignment_id])
     @course = @claimed_assignment.course
     check_permissions(assignment_params[:user_id].to_i)
     check_participation # prevents a user from a different course claiming an assignment
--- a/app/controllers/assignments_controller.rb
+++ b/app/controllers/assignments_controller.rb
@@ -49,7 +49,7 @@ class AssignmentsController < ApplicationController
 
   # Select an Available Article as a new Assignment for a user
   def claim
-    @claimed_assignment = Assignment.find(params[:assignment_id])
+    @claimed_assignment = Assignment.find(params[:id])
     @course = @claimed_assignment.course
     check_permissions(assignment_params[:user_id].to_i)
     check_participation # prevents a user from a different course claiming an assignment
@@ -67,8 +67,9 @@ class AssignmentsController < ApplicationController
   end
 
   def update_status
-    check_permissions(assignment_params[:user_id].to_i)
     @assignment = Assignment.find(assignment_params[:id])
+    @course = @assignment.course
+    check_permissions(assignment_params[:user_id].to_i)
 
     if assignment_params[:status]
       @assignment.update_status(assignment_params[:status])
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -421,7 +421,7 @@ class CoursesController < ApplicationController
   def protect_privacy
     return unless @course.private
     # Admins and enrolled users have non-visitor roles
-    return if current_user&.can_view?(@course)
+    return if current_user&.nonvisitor?(@course)
     raise ActionController::RoutingError, 'not found'
   end
 
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -421,7 +421,7 @@ class CoursesController < ApplicationController
   def protect_privacy
     return unless @course.private
     # Admins and enrolled users have non-visitor roles
-    return if current_user && current_user.role(@course) != CoursesUsers::Roles::VISITOR_ROLE
+    return if current_user&.can_view?(@course)
     raise ActionController::RoutingError, 'not found'
   end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -180,6 +180,8 @@ class User < ApplicationRecord
       if course_user.any? { |cu| EDITING_ROLES.include?(cu.role) }
         # return the editing role
         return course_user.find { |cu| EDITING_ROLES.include?(cu.role) }.role
+      else
+        return course_user.order('role DESC').first.role
       end
     end
 
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -164,14 +164,12 @@ class User < ApplicationRecord
 
   def nonvisitor?(course)
     return true if admin?
-    roles = course_roles(course)
-    roles.size > 1 || (roles.size == 1 && roles.first != CoursesUsers::Roles::VISITOR_ROLE)
+    !course_roles(course).empty?
   end
 
   # returns an array of roles a user has in a given course
   def course_roles(course)
-    user_course_roles = course.courses_users.where(user_id: id).order('role DESC').pluck(:role)
-    return user_course_roles
+    course.courses_users.where(user_id: id).order('role DESC').pluck(:role)
   end
 
   def highest_role(course)
@@ -182,16 +180,11 @@ class User < ApplicationRecord
     roles.first
   end
 
-  def editing_role?(course)
-    # Check if the user has an editing role in the course
-    CoursesUsers.exists?(course:, user: self, role: EDITING_ROLES)
-  end
-
   EDITING_ROLES = [CoursesUsers::Roles::INSTRUCTOR_ROLE,
                    CoursesUsers::Roles::WIKI_ED_STAFF_ROLE].freeze
   def can_edit?(course)
     return true if admin?
-    return true if editing_role?(course)
+    return true if course_roles(course).any? { |role| EDITING_ROLES.include?(role) }
     return true if campaign_organizer?(course)
     false
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -173,9 +173,10 @@ class User < ApplicationRecord
   end
 
   def highest_role(course)
+    return CoursesUsers::Roles::INSTRUCTOR_ROLE if admin?
+
     roles = course_roles(course)
 
-    return CoursesUsers::Roles::INSTRUCTOR_ROLE if admin?
     return CoursesUsers::Roles::VISITOR_ROLE if roles.empty?
 
     roles.first
@@ -188,10 +189,6 @@ class User < ApplicationRecord
     return true if course_roles(course).any? { |role| EDITING_ROLES.include?(role) }
     return true if campaign_organizer?(course)
     false
-  end
-
-  def can_view?(course)
-    nonvisitor?(course)
   end
 
   def campaign_organizer?(course)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -170,8 +170,16 @@ class User < ApplicationRecord
 
   # returns an array of roles a user has in a given course
   def course_roles(course)
-    user_course_roles = course.courses_users.where(user_id: id).pluck(:role)
+    user_course_roles = course.courses_users.where(user_id: id).order('role DESC').pluck(:role)
     return user_course_roles
+  end
+
+  def highest_role(course)
+    roles = course_roles(course)
+
+    return CoursesUsers::Roles::VISITOR_ROLE if roles.empty?
+
+    roles.first
   end
 
   def editing_role?(course)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -172,8 +172,16 @@ class User < ApplicationRecord
     # Give admins the instructor permissions.
     return CoursesUsers::Roles::INSTRUCTOR_ROLE if admin?
 
-    course_user = course.courses_users.where(user_id: id).order('role DESC').first
-    return course_user.role unless course_user.nil?
+    # Get the courses_users record for the user in the course
+    course_user = course.courses_users.where(user_id: id)
+
+    if course_user.exists?
+      # check if the user has any editing role
+      if course_user.any? { |cu| EDITING_ROLES.include?(cu.role) }
+        # return the editing role
+        return course_user.find { |cu| EDITING_ROLES.include?(cu.role) }.role
+      end
+    end
 
     # User is in visitor role, if no other role found.
     CoursesUsers::Roles::VISITOR_ROLE
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -163,20 +163,15 @@ class User < ApplicationRecord
   end
 
   def nonvisitor?(course)
-    role(course) != CoursesUsers::Roles::VISITOR_ROLE
+    return true if admin?
+    roles = course_roles(course)
+    roles.size > 1 || (roles.size == 1 && roles.first != CoursesUsers::Roles::VISITOR_ROLE)
   end
 
-  def role(course)
-    # If this is a new course, grant permissions.
-    return CoursesUsers::Roles::INSTRUCTOR_ROLE if course.nil?
-    # Give admins the instructor permissions.
-    return CoursesUsers::Roles::INSTRUCTOR_ROLE if admin?
-
-    course_user = course.courses_users.where(user_id: id).order('role DESC').first
-    return course_user.role unless course_user.nil?
-
-    # User is in visitor role, if no other role found.
-    CoursesUsers::Roles::VISITOR_ROLE
+  # returns an array of roles a user has in a given course
+  def course_roles(course)
+    user_course_roles = course.courses_users.where(user_id: id).pluck(:role)
+    return user_course_roles
   end
 
   def editing_role?(course)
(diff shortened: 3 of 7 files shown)


ragesoss:
I think this would be better solved by implementing a different method to check for which roles a user has for a course. The User#role method is designed (and named) to return a single user role, but for can_edit? we want to consider all the roles a user has for a course, and check whether any of them are editing roles.

Maybe something like this would work in place of the call to `User#role`: `return true if CoursesUsers.exists?(course: course, user: self, role: EDITING_ROLES)`

TunrayoIlawole:
Thank you @ragesoss . Noted, i will make the update

TunrayoIlawole:
Hi, @ragesoss . I created the new method to check whether any of the roles a user has for a course is an editing role and called that in `can_edit`

ragesoss:
Please check the build log; it looks like this has broken something.

TunrayoIlawole:
Hi @ragesoss, I noticed that in cases where the course is nill, the call to User#role returns an Instructor role which ultimately returns a true value, bypassing the need to check User#campaign_organizer?

For my new method, should i follow a similar approach and return true if the course is nill, since an instructor role is considered an editing role? Or would it be more appropriate to implement a check that returns early in User#campaign_organizer?

I’d appreciate your advice on the best way to handle this case to ensure it aligns with the project's standards. Thank you.

ragesoss:
The behavior of returning the instructor role for cases where the course is nil is confusing, and I'm not exactly sure how we use that. Based on the comment, it seems to be related to the course creation process, before the CoursesUsers record for the instructor has been created, because the course itself as not been saved in the database. But the best thing to do might be to identify exactly where we use that behavior (for example, by throwing an error instead of returning the instructor role, and seeing which tests fail) and handle that differently, so that `role` can be a little more intuitive.

TunrayoIlawole:
Hi, @ragesoss, thank you for your feedback. The behaviour is not really relied on, except in determining whether a user has an editing role. 

The main issue with assigning an Instructor role to a user that is not associated with any course (i.e course is nil) is that it automatically gives the user an editing role. This prevented the error-prone User#campaign_organizer?, which currently does not handle nil values for course and therefore fails, from being called.

This is the code that is failing. `course.campaigns` throws an error because course is nil
```
  def campaign_organizer?(course)
    CampaignsUsers.exists?(user: self, campaign: course.campaigns,
                           role: CampaignsUsers::Roles::ORGANIZER_ROLE)
  end
```

The pipeline is failing because User#campaign_organizer? is now being called as a result of the changes i made to the User#can_edit code to no longer rely on User#role for determining whether a user has an editing role or not.

I propose:
- Returning false in User#campaign_organizer? if the course is nill
- Assigning a default, non-editing role to a user that is not associated with a course or raising an error . Either of these methods will require significant changes to the logic and tests.

Please let me know what you think, @ragesoss

ragesoss:
Which tests fail when you throw an error from the course.nil? check?

TunrayoIlawole:
Hi, @ragesoss , the following tests fail when i throw an error from the nil check:

- [User#role grants instructor permission for a user creating a new course](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/models/user_spec.rb#L55)
- [AssignmentsController PATCH #update_status when a status param is provided renders a 200](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/controllers/assignments_controller_spec.rb#L422)
- [AssignmentsController PATCH #update_status when no status param is provided renders a 422](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/controllers/assignments_controller_spec.rb#L432)
- [AssignmentsController PATCH #update_sanbox_url updating sandbox url with valid urls update sandbox url successfully with example 1](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/controllers/assignments_controller_spec.rb#L452)
- [AssignmentsController PATCH #update_sanbox_url updating sandbox url with valid urls update sandbox url successfully with example 2](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/controllers/assignments_controller_spec.rb#L459)
- [AssignmentsController PATCH #update_sanbox_url updating sandbox url with valid format but belongs to different wiki does not update url and send response with status: unprocessable entity](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/controllers/assignments_controller_spec.rb#L475)
- [AssignmentsController PATCH #update_sanbox_url updating sandbox url with invalid url format does not update url and send response with status: bad request example 1](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/controllers/assignments_controller_spec.rb#L490)
- [AssignmentsController PATCH #update_sanbox_url updating sandbox url with invalid url format does not update url and send response with status: bad request example 2](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/controllers/assignments_controller_spec.rb#L498)


There are other tests that fail for both implementations(i.e with error thrown and with no error thrown), such as:
[updating locale via URL updates the locale and redirects to the home page](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/features/update_locale_spec.rb#L5)

and other errors from Capybara

ragesoss:
Hmm... that's weird. I would not expect `course` to be `nil` in those AssignmentsController tests, as creating the course record is part of the setup. Did you configure it to only throw an error when `course` is `nil`, or every time that line is reached?

TunrayoIlawole:
Hi @ragesoss , sorry i should have clarified. The AssignmentsController tests are failing because User#campaign_organizer? is now being called in the code. It was not called before because User#can_edit? automatically returned true as a result of the instructor role being returned from User#role.

So i belive the only test that is directly failing as a result of the error that i threw in User#role is [User#role grants instructor permission for a user creating a new course](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/models/user_spec.rb#L55)

ragesoss:
Okay. I'm still not clear on why the assignment tests rely on the instructor role behavior. Is it because the test setup is incorrect?

TunrayoIlawole:
They do not rely on the instructor role @ragesoss . Apologies for the confusion. They are failing because of a missing nil check in User#campaign_organizer. The reason why I kept referencing the AssignmentsController tests is because I wanted to give some context as to why they were not failing with the previous implementation.

ragesoss:
Okay. So it sounds like, as suspected, the only test that directly relies on the instructor role-for-nil-course behavior is the one related to creating a new course, but that behavior was also masking either a bug with the campaign_organizer check, or a bug with test setup related to it. Before adding a new nil check, we want to understand whether any actual valid requests can have a nil value, or whether it's just something that is missing from the test setup but in actual usage it could never be nil.

The other thing to look into is how `role` gets called when creating a new course, as we may want to find a different way of handling it.

TunrayoIlawole:
Hi @ragesoss , thank you for your feedback.

From my investigation, i can see that User#role is only called in the application to verify edit permissions and [confirm access to a course](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/app/controllers/courses_controller.rb#L424).

When a course is created by a user, that user is automatically assigned an instructor role. Also, the course is returned whether it was successfully persisted in the database or not. I believe this is the reason why User#role was set to return an instructor role for an empty course.

So the idea is whenever User#role (User = current user i.e the user that created the course) is called with a nil course value, it automatically returns the instructor role so as to avoid not getting any result [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/app/models/user.rb#L175) in case the course has not been saved to the database.

But technically, i do not think any actual valid requests can have a nil course value since a course object is returned whether the course was successfully persisted or not, so User#role will always be called with a course object.

`role` does not get called when creating a new course.

Also, i think the description of this [test](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/df2fac6f923e61f78282b0d3ceb67546cb169e7b/spec/models/user_spec.rb#L56) might be misleading because User#role only determines the role of a user in a given course, it has nothing to do with course creation.

Please let me know what you think.

ragesoss:
Interesting. So it sounds like we should probably remove that behavior altogether, for handling a nil course.

I think implementation of `User#role` is very confusing, as it is really more about permissions than roles. Since it is only called from a few places, I think the best thing would be to refactor it from `#role` to `#course_roles`, make that method simply return an array roles the user has in the course, and move the logic for handling permissions for admins and campaign organizers over to the methods that are explicitly about permission (`User#can_edit?` and maybe a new `User#can_view?` for use in `protect_privacy`.)

TunrayoIlawole:
Noted @ragesoss , I will work on that.

The AssignmentsController tests are also failing because for some reason, when User#can_edit is called in AssignmentsController#check_permissions, an empty course is passed. I am looking into that

TunrayoIlawole:
Hi @ragesoss , I discovered that most of the AssignmentControllers tests that are failing are failing due to incomplete request parameters being passed. 

But i am still having an issue with two of the tests: [1](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/887d8da25c76b672366bdd30f9ccca4464fa095d/spec/controllers/assignments_controller_spec.rb#L422) & [2](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/887d8da25c76b672366bdd30f9ccca4464fa095d/spec/controllers/assignments_controller_spec.rb#L432) and it has to do with not calling `set_course` before executing `update_status` - [see code](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/887d8da25c76b672366bdd30f9ccca4464fa095d/app/controllers/assignments_controller.rb#L12)

Is there a reason for that behaviour?

ragesoss:
The JS action that hits the update_status endpoint does not include any course identifier, so I think set_course would fail. This comes from `updateAssignmentStatus` in the assignment_actions.js file.

ragesoss:
Probably the right approach is to set the course via the assignment (as with `#destroy`) before performing the permissions check.

ragesoss:
@TunrayoIlawole yes

TunrayoIlawole:
Hi @ragesoss , here are all the changes I made to the code:

- created a method User#editing_role? to check if a user has any editing role in a course 
- created a method to return an array of all the courses a user has in a course
- created User#can_view method
- updated User#nonvisitor? to return true if a user has any role other than a visitor role
- updated User#can_see_real_names? to return true if a user has any role that is included in the `REAL_NAME_ROLES` object
- created tests for all new methods
- updated `request_params` in some assignment controller tests to include the `course_slug` so that the course can be set correctly via AssignmentsController#set_course
- set course before checking permissions in AssignmentsController#update_status and AssignmentsController#claim

app/controllers/assignments_controller.rb
```diff
@@ -49,7 +49,7 @@ def assign_reviewers_randomly
 
   # Select an Available Article as a new Assignment for a user
   def claim
-    @claimed_assignment = Assignment.find(params[:assignment_id])
+    @claimed_assignment = Assignment.find(params[:id])
```
ragesoss, line 52: Did you run the tests locally? This is a change I would expect to break things, as the parameter for this action from the `routes.rb` code is passed as `assignment_id`.

spec/models/user_spec.rb
```diff
@@ -76,36 +62,23 @@
...
     end
 
-    it 'returns -1 for a user who is not part of the course' do
-      course = create(:course)
-      user = create(:user)
-      role = user.role(course)
-      expect(role).to eq(-1)
+    it 'returns an empty array when the user has no roles in the course' do
+      course = create(:course,
+                      id: 1)
```
ragesoss, line 77: We generally want to avoid setting `id` in artificial ways like this for tests. Especially when using the same value for different purposes, it can be confusing and also makes it easier to miss bugs.

spec/models/user_spec.rb
```diff
@@ -140,6 +113,156 @@
...
+      create(:courses_user,
+             course_id: 1,
+             user_id: 1,
+             role: 1) # instructor
+
+      # User is also a campus volunteer
+      create(:courses_user,
+             course_id: 1,
+             user_id: 1,
+             role: 2) # campus volunteer
```
ragesoss, line 132: Use the existing constants rather than magic numbers for this (which also eliminates the need for a comment).

ragesoss:
I think that strategy makes sense. I would suggest naming these methods very clearly: have `#user_roles` return all roles, and have `#highest_role` return the highest role, and any additional functionality required for a specific reason should have its own method building on top of those (or implement additional logic independently).

My priority is to make the public methods on this widely-used class as simple and self-explanatory as possible.

app/controllers/courses_controller.rb
```diff
@@ -421,7 +421,7 @@ def verify_edit_credentials
   def protect_privacy
     return unless @course.private
     # Admins and enrolled users have non-visitor roles
-    return if current_user && current_user.role(@course) != CoursesUsers::Roles::VISITOR_ROLE
+    return if current_user&.can_view?(@course)
```
ragesoss, line 424: Why not just use `nonvisitor?`

app/models/user.rb
```diff
@@ -163,31 +163,43 @@ def course_student?
...
-    role(course) != CoursesUsers::Roles::VISITOR_ROLE
+    return true if admin?
+    roles = course_roles(course)
+    roles.size > 1 || (roles.size == 1 && roles.first != CoursesUsers::Roles::VISITOR_ROLE)
+  end
+
+  # returns an array of roles a user has in a given course
+  def course_roles(course)
+    user_course_roles = course.courses_users.where(user_id: id).order('role DESC').pluck(:role)
+    return user_course_roles
```
ragesoss, line 174: Better to make this a one line method without assigning it to a variable or using explicit return.

app/models/user.rb
```diff
@@ -163,31 +163,43 @@ def course_student?
   end
 
   def nonvisitor?(course)
-    role(course) != CoursesUsers::Roles::VISITOR_ROLE
+    return true if admin?
+    roles = course_roles(course)
+    roles.size > 1 || (roles.size == 1 && roles.first != CoursesUsers::Roles::VISITOR_ROLE)
```
ragesoss, line 168: The visitor role does not exist as a CoursesUsers record, so it should never be among the roles returned by `course_roles`

app/models/user.rb
```diff
@@ -163,31 +163,43 @@ def course_student?
...
+    # Check if the user has an editing role in the course
+    CoursesUsers.exists?(course:, user: self, role: EDITING_ROLES)
   end
 
   EDITING_ROLES = [CoursesUsers::Roles::INSTRUCTOR_ROLE,
                    CoursesUsers::Roles::WIKI_ED_STAFF_ROLE].freeze
   def can_edit?(course)
     return true if admin?
-    return true if EDITING_ROLES.include? role(course)
+    return true if editing_role?(course)
```
ragesoss, line 194: Since this concept of editing roles is only used here, I think a better option would be to take the intersection of course_roles and editing roles, like `(EDITING_ROLES & course_roles(course)).any?`, or the same `any?` pattern you are implementing in `can_see_real_names?`. No need for a separate method.

app/models/user.rb
TunrayoIlawole: You are right, @ragesoss . I guess a better approach will be to check that the `course_roles` array is not empty, right?

ragesoss: yeah. course_role().empty? should work.

ragesoss:
I suspect that these two failing tests are related to the changes; they seem like tests that would be affected by 'can edit' checks.

TunrayoIlawole:
Thank you, @ragesoss . Update made

app/models/user.rb
```diff
@@ -175,6 +175,7 @@ def course_roles(course)
   def highest_role(course)
     roles = course_roles(course)
 
+    return CoursesUsers::Roles::INSTRUCTOR_ROLE if admin?
```
ragesoss, line 178: This should be done before the query for roles, as it may replace the need to make a database query.

app/models/user.rb
```diff
@@ -163,31 +163,37 @@ def course_student?
...
                    CoursesUsers::Roles::WIKI_ED_STAFF_ROLE].freeze
   def can_edit?(course)
     return true if admin?
-    return true if EDITING_ROLES.include? role(course)
+    return true if course_roles(course).any? { |role| EDITING_ROLES.include?(role) }
     return true if campaign_organizer?(course)
     false
   end
 
+  def can_view?(course)
```
ragesoss, line 193: This is not needed, correct? (The tests can just be changed to `nonvisitor?`.)

TunrayoIlawole: Update made and tested @ragesoss

PR #5978, opened 2024-10-03
Add Comprehensive Tests for Campaigns Reducer

Formasitchijoh:
## What this PR does
This PR addresses https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/2082 by adding comprehensive tests for the campaigns reducer to ensure its functionality and correctness. The tests cover various scenarios, including adding campaigns, receiving course campaigns, receiving all campaigns, and sorting campaigns. This enhances the test coverage and verifies that the reducer behaves as expected
## Files changed
Created a new file test/reducers/campaigns.spec.js that includes tests for all relevant action types and edge cases related to the campaigns reducer

## Open Questions
@ragesoss I'm unclear about the DELETE_CAMPAIGN action in the reducer. Since it shares case handling with ADD_CAMPAIGN and RECEIVE_COURSE_CAMPAIGNS, does it imply that deletion requires providing an updated list of campaigns, or is there a mechanism for marking a campaign as deleted, like setting its published status to false?

### DIFF
--- a/app/assets/javascripts/reducers/campaigns.js
+++ b/app/assets/javascripts/reducers/campaigns.js
@@ -33,11 +33,7 @@ const initialState = {
 
 export default function campaigns(state = initialState, action) {
   switch (action.type) {
-    case ADD_CAMPAIGN: {
-      const newCampaign = action.data.course.campaigns;
-      const updatedCampaign = [...state.campaigns, newCampaign];
-      return { ...state, campaigns: updatedCampaign, isLoaded: true };
-    }
+    case ADD_CAMPAIGN:
     case DELETE_CAMPAIGN:
     case RECEIVE_COURSE_CAMPAIGNS: {
       const newState = {
--- a/app/assets/javascripts/reducers/campaigns.js
+++ b/app/assets/javascripts/reducers/campaigns.js
@@ -33,7 +33,11 @@ const initialState = {
 
 export default function campaigns(state = initialState, action) {
   switch (action.type) {
-    case ADD_CAMPAIGN:
+    case ADD_CAMPAIGN: {
+      const newCampaign = action.data.course.campaigns;
+      const updatedCampaign = [...state.campaigns, newCampaign];
+      return { ...state, campaigns: updatedCampaign, isLoaded: true };
+    }
     case DELETE_CAMPAIGN:
     case RECEIVE_COURSE_CAMPAIGNS: {
       const newState = {
--- a/test/reducers/campaigns.spec.js
+++ b/test/reducers/campaigns.spec.js
@@ -1,0 +1,95 @@
+import deepFreeze from 'deep-freeze';
+import campaigns from '../../app/assets/javascripts/reducers/campaigns';
+import {
+    RECEIVE_COURSE_CAMPAIGNS,
+    RECEIVE_ALL_CAMPAIGNS,
+    ADD_CAMPAIGN,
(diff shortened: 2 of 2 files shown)


test/reducers/campaigns.spec.js
```diff
@@ -0,0 +1,95 @@
...
+        expect(newState.isLoaded).toBe(false);
+    });
+
+    test('should and and return campaigns data with ADD_CAMPAIGN ans set isLoaded to true', () => {
+        const initialState = { campaigns: [], all_campaigns: [], isLoaded: false };
+        deepFreeze(initialState);
+
+        const mockedAction = {
+          type: ADD_CAMPAIGN,
+          data: { course: { campaigns: campaignsArray } }
```
ragesoss, line 33: This test doesn't reflect the actual usage of the ADD_CAMPAIGN action, as it is only used to add a single campaign per action, but this test goes from zero to two campaigns. (This is still theoretically possible, as another campaign could have been added independently, without this browser store knowing about it, so that adding one campaign would result in a response that includes more than one new campaign. This test can be kept, but it ought to have a comment explaining that sort of situation.)

Formasitchijoh:
> The action name is perhaps not clear enough, but DELETE_CAMPAIGN is the action used to represent removing a campaign from a course (ie, deleting a CampaignsCourses record). On the server, it hits the CoursesController#list action, which handles both adding and deleting campaigns, and returns an updated list of the campaigns for that course in either case.

Thank you for the clarification, i will take a take an in depth look at the controller to understand how to improve the test

app/assets/javascripts/reducers/campaigns.js
```diff
@@ -33,7 +33,11 @@ const initialState = {
 
 export default function campaigns(state = initialState, action) {
   switch (action.type) {
-    case ADD_CAMPAIGN:
+    case ADD_CAMPAIGN: {
+      const newCampaign = action.data.course.campaigns;
+      const updatedCampaign = [...state.campaigns, newCampaign];
+      return { ...state, campaigns: updatedCampaign, isLoaded: true };
+    }
```
ragesoss, line 40: What is the benefit of making this into a separate case?

Formasitchijoh: @ragesoss I had some misunderstandings regarding the definition and use of this reducer. but i have resolved and removed this separate case

test/reducers/campaigns.spec.js
```diff
@@ -0,0 +1,104 @@
...
+  });
+
+  test('should add and return campaigns data with ADD_CAMPAIGN and set isLoaded to true', () => {
+    const initialState = { campaigns: [], all_campaigns: [], isLoaded: false };
+    deepFreeze(initialState);
+
+
+    // Review needed: The implementation of the ADD_CAMPAIGN case in the reducer
+    // is unclear regarding whether it should replace the existing campaigns or
+    // append to them. The expected behavior of the state update should be clarified.
```
ragesoss, line 35: The intended behavior is to return the up-to-date campaigns list in response to adding a new campaign, so replacing the existing campaigns is the expected behavior.

Formasitchijoh: I understand now, @ragesoss  is this ready for or i should remove this comment

ragesoss: Please remove the comment.

Formasitchijoh: @ragesoss I have removed it

PR #5979, opened 2024-10-03
Remove unused confirm_remove_special code

### LINKED ISSUE #5844: The message settings.special_users.new.confirm_remove_special_user appears to be unused
### What is happening?
I noticed that the message settings.special_users.new.confirm_remove_special_user may be problematic: it looks like it's supposed to have a parameter (probably a username), but it doesn't.

However, when I searched for it in the code, I found nothing. It appears to be used only in translation files and not in code.

If it's not actually used, it should be removed from the English messages file.

osujipius:

## What this PR does
Cleans up unused code in the codebase
Fixes #5844


config/locales/en.yml
```diff
@@ -1276,7 +1276,7 @@ en:
         form_label: Grant Special User privileges
         form_placeholder: Username
         confirm_add_special_user: "Add Special User"
-        confirm_remove_special_user: "Remove Special User privileges for"
+        _user: "Remove Special User privileges for"
```
ragesoss, line 1279: What is this for?

osujipius: Looks like it is unused as well @ragesoss

ragesoss: Oh, I guess it was a typo and you intended to remove the whole line?

osujipius: > Oh, I guess it was a typo and you intended to remove the whole line?

Exactly!
Removed it completely now

ragesoss:
Please exclude formatting changes (such as replacing one type of quote mark with another) to any of the locale files except for `en` and `qqq`, as these files are maintained via translatewiki.net.

config/locales/et.yml
```diff
@@ -1300,4 +1299,4 @@ et:
     assignment_status: Õppeülesanne (%{status})
   multi_wiki:
     selector_placeholder: Alusta vikidomeeni sisestamist
-...
+...
```
ragesoss, line 1302: Please exclude the newline changes at the end of the locale files.

db/schema.rb
```diff
@@ -11,15 +11,25 @@
 # It's strongly recommended that you check this file into your version control system.
 
 ActiveRecord::Schema[7.0].define(version: 2024_05_15_163420) do
-  create_table "alerts", id: :integer, charset: "utf8mb4", force: :cascade do |t|
+  create_table "admin_course_notes", charset: "utf8mb3", collation: "utf8mb3_general_ci", force: :cascade do |t|
```
ragesoss, line 14: No schema or yarn.lock changes should be included.

config/locales/et.yml
osujipius: I think this is as a result of taking out the unused code some lines above or do you mean the "..." should be removed entirely?

ragesoss: No, this is likely a result of your text editor automatically remove the invisible newline character at the end of the file.

When you run `git diff`, the only changes should be the lines you removed.

Issue #5701, opened 2024-03-11
Calender not rendering correctly in date-picker in Course Creator


PR4NJ41:
### What is happening?
In `date-picker` dates are not rendering correctly. The disabled dates must render whenever we are choosing the dates. But they are rendering later.

### To Reproduce

In the course creator make changes as shown in the video

### Expected behavior
The disabled dates must render whenever we are choosing the dates.

### Screenshots

https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/96368921/f4da0cc8-95a8-4c58-a288-79a43c170256

claire256:
Hello @ragesoss, i'm taking up this issue.

PR #5981, opened 2024-10-05
Date Calender not rendering correctly when creating a course.

claire256:
issue - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5701#issue-2179974510

## What this PR does
This PR is fixing the disable issue that was occurring when creating an Independent Program.

## Screenshots
Before:
[Screencast from 05-10-24 19:26:15.webm](https://github.com/user-attachments/assets/097f6d36-2717-4e5b-9de0-af7f1751e4c7)

After:
[Screencast from 24-10-24 21:05:37.webm](https://github.com/user-attachments/assets/66643f1c-dd7e-41bc-b4f8-ab290d22bec9)

### DIFF
--- a/app/assets/javascripts/components/course_creator/course_dates.jsx
+++ b/app/assets/javascripts/components/course_creator/course_dates.jsx
@@ -1,12 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import DatePicker from '../common/date_picker.jsx';
 import CourseUtils from '../../utils/course_utils.js';
 import CourseDateUtils from '../../utils/course_date_utils.js';
 
 const CourseDates = (props) => {
+  const [hasUpDatedDateProps, setHasUpDatedDateProps] = useState(false);
+
   const updateCourseDates = (key, value) => {
     const updatedCourse = CourseDateUtils.updateCourseDates(props.course, key, value);
     props.updateCourseProps(updatedCourse);
+    setHasUpDatedDateProps(true);
   };
   const dateProps = CourseDateUtils.dateProps(props.course);
   const timeZoneMessage = (
@@ -87,6 +90,7 @@ const CourseDates = (props) => {
         enabled={!!props.course.start}
         isClearable={false}
         showTime={props.showTimeValues}
+        rerenderHoc={hasUpDatedDateProps}
       />
       {timelineText}
       {timelineStart}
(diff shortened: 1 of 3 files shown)


ragesoss:
Hmm...

This looks like an improvement. The system is supposed to enforce that event start/end dates are not outside of the range of activity tracking dates, but it looks like this only applies that to the start date, not the end date.

app/assets/javascripts/components/course_creator/course_dates.jsx
```diff
@@ -1,12 +1,15 @@
-import React from 'react';
+import React, { useState } from 'react';
 import DatePicker from '../common/date_picker.jsx';
 import CourseUtils from '../../utils/course_utils.js';
 import CourseDateUtils from '../../utils/course_date_utils.js';
 
 const CourseDates = (props) => {
+  const [hasUpDatedDateProps, setHasUpDatedDateProps] = useState(false);
```
ragesoss, line 7: I think these should use `Updated` rather than `UpDated`.

claire256: Okay

claire256:
Alright @ragesoss , Let me make the necessary changes.

claire256:
Hello @ragesoss , could you have a look at this again. I have also been able to disable the end dates inputs.

ragesoss:
Looks like the build is failing related to your changes.

app/assets/javascripts/utils/course_date_utils.js
```diff
@@ -50,7 +50,7 @@ const CourseDateUtils = {
       },
       timeline_start: {
         minDate: startDate,
-        maxDate: parseISO(course.timeline_end)
+        maxDate: parseISO(course.end)
```
ragesoss, line 53: I don't think this is a change we want to make. The timeline start should always be before the timeline end, and the timeline end should always be before the course end.

claire256: I reverted the changes so that i could debug why the CI workflow is failing. But i'm planning on putting them back.

claire256: > I don't think this is a change we want to make. The timeline start should always be before the timeline end, and the timeline end should always be before the course end.

Concerning this, a user will start by selecting the timeline start and at that point we are not sure of what date the user will select for timeline end. This means we may not be able to validate the max date for timeline start. Is this what is expected?

ragesoss:
Thanks! Can you update the 'after' video based on the current implementation?

claire256:
Hello @ragesoss, I have updated the 'after' video, you can have a look.

PR #5983, opened 2024-10-06
Add ?teach parameter to see instructor version of homepage regardless of user role (#5812)

### LINKED ISSUE #5812: add a parameter to override the lack of 'instructor' role to let a user create a course
We encourage people who have taken a Scholars & Scientists course with Wiki Education to then teach in our Wikipedia Student Program. If a user first signed in to the Dashboard via an enrollment link, then the 'onboarding' flow does not let them indicate they are an instructor, and their user permissions will be `0`. That means they won't be shown the link to the instructor orientation or the 'create a course' button on the Dashboard home page.

We should add a parameter (such as ?teach) that overrides this to show the instructor version of the landing page, so that fewer users get stuck and need to ask about how to reset their instructor status (via /onboarding) before creating a course.

Abishekcs:
#5812

## What this PR does

Introduces a new feature to force the instructor view of the homepage for users using ?teach param, regardless of their role

## Screenshots
Before:

https://github.com/user-attachments/assets/e4e279d3-b8f6-4d2a-910a-96265a83d3f4

After:

https://github.com/user-attachments/assets/da3cd9fc-cf06-4252-b2eb-91d2dbfd5e67

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -14,7 +14,8 @@ class DashboardController < ApplicationController
     set_blog_posts if Features.wiki_ed?
     set_admin_courses_if_admin
     @pres = DashboardPresenter.new(current_courses, past_courses,
-                                   @submitted, @strictly_current, current_user)
+                                   @submitted, @strictly_current,
+                                   current_user, params.key?(:teach))
   end
 
   def my_account
--- a/app/presenters/dashboard_presenter.rb
+++ b/app/presenters/dashboard_presenter.rb
@@ -8,16 +8,17 @@ class DashboardPresenter
 
   ORIENTATION_ID = 3
 
-  def initialize(current, past, submitted, strictly_current, current_user)
+  def initialize(current, past, submitted, strictly_current, current_user, force_instructor_view) # rubocop:disable Metrics/ParameterLists
     @current = current
     @past = past
     @submitted = submitted
     @strictly_current = strictly_current
     @current_user = current_user
+    @force_instructor_view = force_instructor_view
   end
 
   def instructor?
-    current_user.instructor_permissions?
+    @force_instructor_view || current_user.instructor_permissions?
   end
 
   def admin?
@@ -80,7 +81,7 @@ class DashboardPresenter
 
   # Show explore button for non instructors/admins
   def show_explore_button?
-    current_user.permissions == User::Permissions::NONE
+    !instructor? && current_user.permissions == User::Permissions::NONE
   end
 
   # Get the url path for orientation


ragesoss:
This is different than what I had in mind. Rather than using the ?teach parameter for enrollment or onboarding links, I'd like the homepage itself to use that parameter to render the instructor version regardless of user role. The idea would be that we could include links to `https://dashboard.wikiedu.org/?teach` that would let a user get access to the instructor view of the My Dashboard page (if logged in).

I think we could also optionally also automatically set the role from student to instructor,  but the place to do that would be upon completion of the instructor orientation training module.

app/controllers/training_modules_users_controller.rb
```diff
@@ -31,6 +32,38 @@ def mark_exercise_complete
 
   private
 
+  def instructor_training_complete?
```
ragesoss, line 35: This feature should not be in the same PR. I'd prefer to first merge the simplest possible version that will address the issue.

app/controllers/dashboard_controller.rb
```diff
@@ -15,6 +15,7 @@ def index
     set_admin_courses_if_admin
     @pres = DashboardPresenter.new(current_courses, past_courses,
                                    @submitted, @strictly_current, current_user)
+    @instructor_view = params.key?(:teach) and user_signed_in?
```
ragesoss, line 18: I think the better way to handle this is inside the DashboardPresenter, by passing the 'teach' parameter to the presenter.

Using a separate instance variable makes it more complicated to get all the correct behaviors (such as showing the link to instructor orientation but not the create course button, if the user has not already completed that training module). With the presenter approach, I think there will be no changes needed in the view templates.

app/controllers/training_modules_users_controller.rb
Abishekcs: Okay

app/controllers/dashboard_controller.rb
Abishekcs: Okay, I will make the required changes. Thank you for the review

app/controllers/dashboard_controller.rb
```diff
@@ -46,6 +47,10 @@ def current_courses
     current_user.courses.current_and_future
   end
 
+  def force_instructor_view?
+    params.key?(:teach) && user_signed_in?
```
ragesoss, line 51: I don't think `user_signed_in?` is necessary, as a logged out user gets redirected to HomeController#index.

Abishekcs: Okay

Abishekcs: I've removed both user_signed_in and force_instructor_view, so I can now pass the teach parameter directly.

PR #5985, opened 2024-10-06
Fix failing  tests for the alert_actions

Formasitchijoh:
## What this PR does
This pull request fixes the failing test for the alert actions in the Redux store. The test was previously failing due to an assertion error where the expected and received states did not match## Screenshots

Before:
Screenshot 2024-10-03 at 03 31 54

After:
< add a screenshot of the UI af
Screenshot 2024-10-06 at 20 14 58
ter your change >

@ragesoss , @Abishekcs  Could it be that some promises were not being resolved, leading to excessive waits and causing the 'call retries were exceeded' error? I opted to use async/await to ensure that actions complete in the intended order. Does this approach improve the readability and reliability of the tests?

### DIFF
--- a/test/actions/alert_actions.spec.js
+++ b/test/actions/alert_actions.spec.js
@@ -4,16 +4,20 @@ import '../testHelper';
 describe('AlertActions', () => {
   const initialState = { submitting: false, created: false };
   const expectedState = { submitting: false, created: true };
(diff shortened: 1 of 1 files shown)


ragesoss:
Odd that this was failing locally, but not on CI. It must be something related to behavior that differs between different node environments, or something else particular to the testing environment.

PR #5987, opened 2024-10-08
test: add RSpec unit test for WizardController#wizard_index

Solobarine:
## Summary
This Pull Request adds a unit test for the `wizard_index` action in `WizardController`.

## Test Behaviour
- The test mocks the YAML content loaded from the `wizard_index.yml` file located in the Rails config directory.
- It verifies that:
  - The action successfully reads the YAML file.
  - The content is correctly converted to JSON and returned in the response.
  - The response status is `200 OK`.
  - The response content type is `application/json`.
  - The response encoding is`utf-8`.
  - The response body matches the expected YAML content that is parsed into a JSON object.
  
  ## Notes
This test ensures that the `wizard_index` action behaves as expected when rendering wizard configuration content in JSON format.

### DIFF
--- a/app/views/faq_topics/edit.html.haml
+++ b/app/views/faq_topics/edit.html.haml
@@ -1,4 +1,4 @@
-- content_for :before_title, "Edit FAQ Topic - #{@topic.title}"
+- content_for :before_title, "Edit FAQ Topic - #{@topic.slug}"
 .container
   %h2= "Editing topic: #{@topic.slug}"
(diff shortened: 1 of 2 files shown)


ragesoss:
Although we don't have a unit test for this specific controller method, we have feature tests that rely on it working correctly (the course creation feature specs). Unless there is a coverage gap that represents untested aspects of this controller, I think the unit test may be unnecessary.

Solobarine:
Ok. I look for areas with less test coverage and add tests to some of them.
Thanks @ragesoss.

PR #5988, opened 2024-10-08
Enable react/jsx-equals-spacing javascript linting rule

claire256:
issue- https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1408#issue-259228085

## What this PR does
This PR enables the react/jsx-equals-spacing JavaScript linting rule.

I learned that you could run ' yarn coverage' in order to pick all the linting errors instead of only running 'yarn lint-non-build'.


.eslintrc
```diff
@@ -29,7 +29,7 @@
     # Rules below here should be enabled once violations are fixed.
     "react/forbid-prop-types": 0,
     "react/no-string-refs": 0,
-    "react/jsx-equals-spacing": 0,
+    "react/jsx-equals-spacing": [2, "never"],
```
ragesoss, line 32: This is a rule that is enabled by default, so rather than changing the configuration, the entry for it should simply be removed so that we rely on the defaults. The goal here is to get as close to a default configuration as possible, rather than explicitly configure the disabled rules.

claire256: Okay, noted.

PR #5993, opened 2024-10-09
Enable object-short-hand linting rule

claire256:
issue - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/1408#issue-259228085

## What this PR does
This PR is enabling the  Object-Short-hand JavaScript linting rule. 
Object-Short-hand require or disallow method and property shorthand syntax for object literals. In this PR i'm only enforcing the use of shorthand syntax on only 'properties' defined where the key name matches name of the assigned variable. 

The default rule (always), would apply to all 'methods' (including generators) defined in object literals and any 'properties'

PROPERTIES
Example ;
    
 *before:
              var foo = {
                    x: x,
                    y: y,
                    z: z,
                };
 
 *after
             var foo = {x, y, z};    

METHODS       
Example ;
    
 *before:
              var foo = {
                    a: function() {},
                    b: function() {}
                    };

 *after
     var foo = {
              a() {},
              b() {}
             };          
      
#Why
I did not enable the default rule (leaving out the methods), because i felt it would bring violations that may be confusing to some people, so i felt it was not necessary.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,7 +46,6 @@
     "import/no-named-as-default": 0,
     "no-underscore-dangle": 0,
     "no-empty-pattern": 0,
-    "object-shorthand": ["error", "always"],
     "no-useless-computed-key": 0,
     "newline-per-chained-call": 0,
     "class-methods-use-this": 0,
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,7 +46,7 @@
     "import/no-named-as-default": 0,
     "no-underscore-dangle": 0,
     "no-empty-pattern": 0,
-    "object-shorthand": 2,
+    "object-shorthand": ["error", "properties"],
     "no-useless-computed-key": 0,
     "newline-per-chained-call": 0,
     "class-methods-use-this": 0,
--- a/.eslintrc
+++ b/.eslintrc
@@ -46,7 +46,7 @@
     "import/no-named-as-default": 0,
     "no-underscore-dangle": 0,
     "no-empty-pattern": 0,
-    "object-shorthand": ["error", "properties"],
+    "object-shorthand": ["error", "always"],
     "no-useless-computed-key": 0,
     "newline-per-chained-call": 0,
     "class-methods-use-this": 0,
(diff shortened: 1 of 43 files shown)


claire256:
What do you think about this linting rule? @ragesoss

.eslintrc
```diff
@@ -46,7 +46,7 @@
     "import/no-named-as-default": 0,
     "no-underscore-dangle": 0,
     "no-empty-pattern": 0,
-    "object-shorthand": 0,
+    "object-shorthand": 2,
```
ragesoss, line 49: Is `2` the default? We want to enable rules by removing their entries from the dot file, rather than hard-coding their settings.

ragesoss:
I'm fine with enabling this rule.

.eslintrc
claire256: 'Always' is the default, but i have changed it to only 'properties'

ragesoss:
> I did not enable the default rule because this would bring complicated violations which could also result in changing the purpose of the code.

Can you explain more? I don't know what the default rule is or what kinds of violations it would trigger.

claire256:
I have explained more. @ragesoss

ragesoss:
Thanks. How many violations of the default rule would there be?

claire256:
They're (93) for default and (90) for properties only. @ragesoss

ragesoss:
@claire256 so there are only three cases where the methods would need to be changed to fit the rule? If that's the case, I'd like to see what it would look like to make those changes.

claire256:
Hello @ragesoss, I have enabled all the default changes.

.eslintrc
```diff
@@ -46,7 +46,7 @@
     "import/no-named-as-default": 0,
     "no-underscore-dangle": 0,
     "no-empty-pattern": 0,
-    "object-shorthand": 0,
+    "object-shorthand": ["error", "always"],
```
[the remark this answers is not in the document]
claire256: Alright.

PR #5996, opened 2024-10-12
Fix how to calculate error_count in flags['update_logs'] for courses 

arafats1:
This PR addresses issue #5580 by ensuring that the `error_count` in `flags['update_logs']` properly tracks the number of actual errors during the course update process.

- Modified `UpdateServiceErrorHelper.update_error_stats` to accept an optional new_errors_count parameter. This parameter should default to 1 if not provided.
- Modified` ApiErrorHandling.log_error` to pass the correct error count using the sentry_extra[:new_errors_count] value (if available).
- Added isolated tests for `ApiErrorHandling` focusing on the log_error method to ensure it accurately increments the error count based on various scenarios.
- Added isolated tests for `update_service_error_helper_spec.rb` to validate the functionality of the update_error_stats method.


EmmanuelEmp:
Hey @arafats1, I think changes in your "schema.rb" file are auto-generated on setup. And should not be included in your commits.

arafats1:
> Hey @arafats1, I think changes in your "schema.rb" file are auto-generated on setup. And should not be included in your commits.

Hi @EmmanuelEmp, Oh yes, I missed that, should I run `rake erd orientation=vertical` and `annotate`, Will that rectify it? Kindly guide me, thank you.

gabina:
Please avoid including changes to unrelated files like `.gitignore`, `db/schema.rb`, or `yarn.lock` in this PR. Kindly add a commit to revert changes in those files.

ragesoss:
In addition to the cleanup of unrelated file changes, please include a detailed explanation of how you've tested this change. With this sort of change, it is not easy to verify the correctness just by looking at the code.

spec/services/update_course_stats_spec.rb
```diff
@@ -78,6 +78,23 @@
       create(:courses_user, course_id: course.id, user_id: user.id)
     end
 
+    it 'tracks individual errors instead of batch count' do
+  
+      allow(Sentry).to receive(:capture_exception)
+      allow_any_instance_of(LiftWingApi).to receive(:get_single_revision_parsed_data).and_raise(Faraday::ConnectionFailed)
+    
+      rev_ids = [1, 2, 3, 4]
+      subject.get_revision_data(rev_ids)  # Assuming `subject` is an instance of LiftWingApi
```
ragesoss, line 87: subject is defined in line 7; it's in instance of `UpdateCourseStats` initialized with a course.

ragesoss: Did you run this test locally? I would expect it to fail with a NoMethodError here, because the subject is not a LiftWingApi instance.

arafats1: > Did you run this test locally? I would expect it to fail with a NoMethodError here, because the subject is not a LiftWingApi instance.

Yes It threw that error, i reverted changes in that `update_course_stats_spec.rb `file , trying to test with the default expected 8 errors but I am getting more than that when i run the test
```
UpdateCourseStats sentry course update error tracking tracks update errors properly in LiftWing
     Failure/Error: expect(course.flags['update_logs'][1]['error_count']).to eq 8
     
       expected: 8
            got: 336
```
Iam thinking there might be an issue with how errors are being counted in the batch processing logic but have not found it yet. Because the error_count is being incremented by a much larger value than expected. In the default test, expected error_count is to be 8, but it turned out to be 336.

Kindly guide me

arafats1: > Did you run this test locally? I would expect it to fail with a NoMethodError here, because the subject is not a LiftWingApi instance.

Yes It threw that error, i reverted changes in that `update_course_stats_spec.rb `file , trying to test with the default expected 8 errors but I am getting more than that when i run the test
```
UpdateCourseStats sentry course update error tracking tracks update errors properly in LiftWing
     Failure/Error: expect(course.flags['update_logs'][1]['error_count']).to eq 8
     
       expected: 8
            got: 336
```
Iam thinking there might be an issue with how errors are being counted in the batch processing logic but have not found it yet. Because the error_count is being incremented by a much larger value than expected. In the default test, expected error_count is to be 8, but it turned out to be 336.

Kindly guide me

ragesoss: The other errors are likely from web requests that other parts of the full `UpdateCourseStats` routine, which fail because VCR prevents from them running. Try adding some `puts` statements inside the error logging method to see what is coming through.

arafats1:
Yes, after using` puts` logging, the other errors are from web requests being prevented. The WebMock exceptions I'm encountering seem to be affecting multiple revision IDs. The log indicates that the tests are attempting to make HTTP requests, which are being intercepted and blocked by WebMock.

At the start of the logs
start

At the end 
end

ragesoss:
That's the expected behavior, as we only allow web requests within explicit VCR blocks from the test suite.

In this case, I think the best strategy would be to test the changed functionality in a more isolated way, without relying on UpdateCourseStats, since your test was specific to the LiftWingApi.

arafats1:
I took

> That's the expected behavior, as we only allow web requests within explicit VCR blocks from the test suite.
> 
> In this case, I think the best strategy would be to test the changed functionality in a more isolated way, without relying on UpdateCourseStats, since your test was specific to the LiftWingApi.

I took into consideration isolated testing, I reverted back to the master branch and did the specific changes as I tested them independently. I did not make any changes to the `LiftWingApi` class. Given that the LiftWingApi class already implements error handling and calls log_error with the necessary context, it doesn't seem to require changes for the error_count calculation in flags['update_logs'].

Based on the provided information, I made and tested changes in the `UpdateServiceErrorHelper` and `ApiErrorHandling` classes
Reference

For changes in `UpdateServiceErrorHelper` :
I added isolated tests for the module to validate the functionality of the update_error_stats method. (file: `update_service_error_helper_spec.rb`)

The tests cover the following scenarios:
Default Increment: The first test verifies that calling update_error_stats without any arguments correctly increments the error_count by 1.

Custom Increment: The second test checks that when a specific number of errors is provided as an argument, update_error_stats increments the error_count by that exact number.

Multiple Increments: The third test confirms that the error_count can be incremented multiple times with different values, ensuring cumulative updates work as intended.
Test-Update

For changes in `ApiErrorHandling`:
I also added isolated tests for this module focusing on the log_error method to ensure it accurately increments the error count based on various scenarios. (file: `api_error_handling_helper_spec.rb`)

The tests include the following:
Default Increment: The first test verifies that calling log_error without a new_errors_count argument correctly increments the error_count by 1.

Custom Increment: The second test checks that when a specific new_errors_count is provided in the sentry_extra argument, log_error increments the error_count by that exact number.

Missing Argument Handling: The third test confirms that if the sentry_extra argument is missing, the method defaults to incrementing the error_count by 1.
Service-error-spec

I pushed the changes with the test files for review, let me know if I should remove the files and re-commit.

ragesoss:
There are linting errors in the build.

arafats1:
> There are linting errors in the build.

Let me check them out.

arafats1:
> There are linting errors in the build.

Fixed the lining errors.

ragesoss:
Nice job. I like the way you implemented the tests for this, mirroring how we actually want to use the logging module.

lib/errors/api_error_handling.rb
```diff
@@ -5,7 +5,8 @@ def log_error(error, update_service: nil, sentry_extra: nil)
     Rails.logger.info "Caught #{error}"
     sentry_tags = nil
     if update_service.present?
-      update_service.update_error_stats
+      new_errors_count = sentry_extra&.dig(:new_errors_count) || 1
```
ragesoss, line 8: We don't actually use this anywhere, correct? So this will not change the current behavior of the related bug until we also new_errors_count within the code that is collating errors?

arafats1: Yes, that is correct. 
This change will not affect the current behavior until we also implement `new_errors_count` when reporting errors in places where errors are batched or where multiple errors may occur. 
This will enable the log_error method to increment the error count based on the actual number of errors, instead of defaulting to 1.

arafats1:
> Nice job. I like the way you implemented the tests for this, mirroring how we actually want to use the logging module.

Thank you!!!

PR #6003, opened 2024-10-16
Feat: Improved Training Slide Navigation Button

EmmanuelEmp:
## What this PR does
This PR updates the training slide navigation by improving the styling(text and icon) of the `Previous` button to match the `Next` button.

## Screenshots
Before:
Previous_slide_nav

After:
New_slide_nav


ragesoss:
I don't think this is a change I want to make. The main UI flow will be to proceed forward through the module, so the more subtle 'Previous' link compared to the prominent 'next' link is intentional.

EmmanuelEmp:
Interesting, I understand the UI/UX need to emphasize the 'Next' button above the 'Previous'.

Issue #6012, opened 2024-10-28
On Programs & Events Dashboard, allow deleting a program before removing campaigns


ragesoss:
The requirement to remove campaigns before deleting a course is the behavior we want in `wiki_education` mode, but it's confusing and unnecessary to require it for pages on Programs & Events Dashboard.

Implementing this will resolve https://phabricator.wikimedia.org/T378013

empty-codes:
I would like to work on this!

PR #6015, opened 2024-10-29
Feature: Allows Programs to be Deleted before Removing Campaigns on Programs & Events Dashboard #6012

empty-codes:
## What this PR does
Solves Issue #6012 and https://phabricator.wikimedia.org/T378013 by removing the requirement to remove campaigns before deleting a program is the behavior on Programs & Events Dashboard.

## Changes Made:
1. When a user attempts to delete a program with the 'Delete Program' button:
-If in the Wiki Ed environment or if the program has no associated campaigns, it is deleted immediately using the normal `deleteCourse` actions.
-If in a non-Wiki Ed environment i.e on Program and Events Dashboard and the program is linked to campaigns, it removes the program from the associated campaign(s) before deletion, using new `removeAndDeleteCourse` actions.

This makes deleting courses more flexible by eliminating the previous need to first remove all campaign associations manually.
It however retains the requirement to remove campaigns before deleting a course if in wiki_education mode.

2. Alerts and hover-activated titles showing course delete instructions removed on the Program & Events Dashboard

## Screenshots:

Before:
Could not delete programs that were associated with campaigns, no `DeleteCourseWorker` jobs seen in sidekiq

before alert, tooltip

After:
Can now delete programs that are associated with campaigns, DeleteCourseWorker jobs seen in sidekiq and sentry

after working now

Evidence of Deletion in Sidekiq:
world reg cropped

Evidence of Deletion in Sentry:
image

### DIFF
--- a/app/assets/javascripts/components/overview/available_actions.jsx
+++ b/app/assets/javascripts/components/overview/available_actions.jsx
@@ -58,8 +58,8 @@ const AvailableActions = ({ course, current_user, updateCourse, courseCreationNo
     const enteredTitle = prompt(I18n.t('courses.confirm_course_deletion', { title: course.title }));
     // Check if enteredTitle is not null before calling trim.
     if (enteredTitle !== null && enteredTitle.trim() === course.title.trim()) {
-      // If in Wiki Ed or course has no campaigns, delete the course directly; otherwise, remove from campaign first.
-      if (Features.wikiEd || !course.campaigns || course.campaigns.length === 0) {
+      // If course has no campaigns, delete the course directly; otherwise, remove from campaign first.
+      if (!course.campaigns || course.campaigns.length === 0) {
         return dispatch(deleteCourse(courseSlug));
       }
       const campaign = course.campaigns[0];
--- a/app/assets/javascripts/components/overview/available_actions.jsx
+++ b/app/assets/javascripts/components/overview/available_actions.jsx
@@ -54,11 +54,6 @@ const AvailableActions = ({ course, current_user, updateCourse, courseCreationNo
   };
 
   const deleteCourseFunc = () => {
-    // The action is only available once a course has been removed from all campaigns.
-    if (course.published) {
-      return alert(I18n.t('courses.delete_course_instructions'));
-    }
-
     const enteredTitle = prompt(I18n.t('courses.confirm_course_deletion', { title: course.title }));
     // Check if enteredTitle is not null before calling trim.
     if (enteredTitle !== null && enteredTitle.trim() === course.title.trim()) {
@@ -112,7 +107,7 @@ const AvailableActions = ({ course, current_user, updateCourse, courseCreationNo
     // so that users can see the instructions for how to enable deletion.
     if ((user.isAdvancedRole || user.admin) && (!course.published || !Features.wikiEd)) {
       controls.push((
-        <div title={I18n.t('courses.delete_course_instructions')} key="delete" className="available-action">
+        <div title={Features.wikiEd ? I18n.t('courses.delete_course_instructions') : undefined} key="delete" className="available-action">
           <button className="button danger" onClick={deleteCourseFunc}>
             {CourseUtils.i18n('delete_course', course.string_prefix)}
           </button>
--- a/app/assets/javascripts/components/overview/available_actions.jsx
+++ b/app/assets/javascripts/components/overview/available_actions.jsx
@@ -11,7 +11,7 @@ import CourseStatsDownloadModal from './course_stats_download_modal.jsx';
 import EmbedStatsButton from './embed_stats_button.jsx';
 import CloneCourseButton from './clone_course_button.jsx';
 import { enableAccountRequests } from '../../actions/new_account_actions.js';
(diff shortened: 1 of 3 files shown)


empty-codes:
@ragesoss What I have done currently is to remove the alert and also stop the instructions from showing when you hover over the delete button on P&E.
I want to confirm if I am to change the logic of the 'Delete program' button itself or this is enough. Thank you

ragesoss:
Yes, I think the right way is to change the logic to work the same way as the 'remove and delete' button on the campaign page: it should first remove the campaigns and then delete the course.

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -108,11 +112,9 @@ const AvailableActions = ({ course, current_user, updateCourse, courseCreationNo
       controls.push((<div key="search" className="available-action"><a href={`/tickets/dashboard?search_by_course=${course.slug}`} className="button">{I18n.t('courses.search_all_tickets_for_this_course')}</a></div>));
     }
     // If course is not published, show the 'delete' button to instructors and admins.
-    // Show a disabled version of it on P&E Dashboard even if a course is published,
-    // so that users can see the instructions for how to enable deletion.
     if ((user.isAdvancedRole || user.admin) && (!course.published || !Features.wikiEd)) {
       controls.push((
-        <div title={I18n.t('courses.delete_course_instructions')} key="delete" className="available-action">
+        <div title={Features.wikiEd ? I18n.t('courses.delete_course_instructions') : undefined} key="delete" className="available-action">
```
ragesoss, line 117: What's this change for?

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -54,15 +54,19 @@ const AvailableActions = ({ course, current_user, updateCourse, courseCreationNo
   };
 
   const deleteCourseFunc = () => {
-    // The action is only available once a course has been removed from all campaigns.
-    if (course.published) {
```
ragesoss, line 58: I think the easiest way to handle this would be to keep the alert but only if course.published AND Features.wikiEd.

app/assets/javascripts/components/overview/available_actions.jsx
empty-codes: From my understanding the old if statement translates as: 

Assuming the user has an advanced role or is an admin:

1. **If the course is not published:**
-For both WikiEd and (P&E), the title attribute will be set to the instructions, which display on hover.
2. **If the course is published:**
-In WikiEd mode, the button will not be displayed at all, because the condition fails due to !Features.wikiEd being false.
-In P&E mode, the title attribute will be set to the instructions, which display on hover.

Since the desired behaviour is to stop the "You may delete this page if all campaigns are removed first. " instructions from showing in P&E, the added condition only sets the title to the instructions if in WikiEd mode.

So basically, it only displays on hover if the course is not published and in WikiEd mode.

app/assets/javascripts/components/overview/available_actions.jsx
empty-codes: I think that due to this if statement: `if ((user.isAdvancedRole || user.admin) && (!course.published || !Features.wikiEd))` , when Features.wikiEd is true, the delete button is only rendered if the course is not published.  If the course is published, the delete button won't render at all, and therefore, deleteCourseFunc won’t be triggered. 

It worked previously because the button was rendered in P&E mode whether the course is published or not, so adding the ` if (course.published)` allowed the alert to show only if in P&E mode AND course.published.
 
But I also understand that we can keep the alert as a form of two-factor authentication for course deletion. Should I still keep it?

ragesoss: We should keep the alert that requires the user to enter the title of the course, I think. But if there was an additional one that was just informational, that's not necessary.

empty-codes: Yes🫡, the title confirmation prompt is still present (it's shown in the after screenshot  in the PR description). It was the additional one removed.

app/assets/javascripts/components/overview/available_actions.jsx
```diff
@@ -54,15 +54,19 @@ const AvailableActions = ({ course, current_user, updateCourse, courseCreationNo
...
-      return alert(I18n.t('courses.delete_course_instructions'));
-    }
-
+    const courseSlug = course.slug;
     const enteredTitle = prompt(I18n.t('courses.confirm_course_deletion', { title: course.title }));
     // Check if enteredTitle is not null before calling trim.
     if (enteredTitle !== null && enteredTitle.trim() === course.title.trim()) {
-      return dispatch(deleteCourse(course.slug));
+      // If in Wiki Ed or course has no campaigns, delete the course directly; otherwise, remove from campaign first.
+      if (Features.wikiEd || !course.campaigns || course.campaigns.length === 0) {
```
ragesoss, line 62: Is the check for Features.wikiEd here redundant, since it will only render the button in the first place if the campaigns length is zero (because that means it is not published)?

empty-codes: That is true! Another oversight, since published campaigns are automatically added to Miscellanea. I'll remove that now

Issue #5837, opened 2024-06-20
On P&E Dashboard, 'When' field validation is overzealous


ragesoss:
Per bug report on Meta: "If you empty the “when” field, this will not be saved and will result in an error "This field has invalid characters". knowing that I don't want to make this modification because it impacts the event title which in turn causes a change to the event link I want to add in my event tool"

On Wiki Education Dashboard, 'term' is a required field, but we don't require it on P&E Dashboard so it should be possible to blank that field.

akshaya9999:
@ragesoss Just to be clear, you want the institution box in the attached ss to be not mandatory?
![image](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/116485510/2730e08a-bcc4-4cb8-8da6-d84bdc19afbf)

ragesoss:
@akshaya9999 correct. But only for the 'Create an Independent Program' version, which is used in `wiki_education: 'false'`  mode (via `application.yml`). It should still be required in the `wiki_education: 'true'` case.

Formasitchijoh:
working on this

Formasitchijoh:
@ragesoss I have a quick question on this, task, since the institution field in the P&E dashboard is used to construct the slug or the event title which in turn causes a change to the event link, should there be a generic name used in this scenario or the title should be constructed with just the  title in case the institution is not provided

ragesoss:
Hmm... I must have been confused in my earlier comment. The 'institution' field does need to remain mandatory. It's just the 'when' (aka 'term') field that we want to relax the validation for.

Formasitchijoh:
Alright thank you for the clarification

PR #6017, opened 2024-10-29
Fix: Relax the validation on term (when) field on create a course form

Formasitchijoh:
## What this PR does
This PR addresses  issue #5837 where the "Term" field in the Wiki Education Dashboard's course creation form. Previously, the validation for this field was overly strict. This update relaxes the validation criteria
## Screenshots
Before:

https://github.com/user-attachments/assets/cc7c0157-3f74-4c01-9cd9-67ce361f6d4a

After:

https://github.com/user-attachments/assets/ab169d8e-8253-4aa0-af2e-d33a145ae8e8

### DIFF
--- a/app/assets/javascripts/components/overview/course_cloned_modal.jsx
+++ b/app/assets/javascripts/components/overview/course_cloned_modal.jsx
@@ -176,6 +176,7 @@ const CourseClonedModal = createReactClass({
 
     const dateProps = CourseDateUtils.dateProps(this.state.course);
     const saveDisabled = this.saveEnabled() ? '' : 'disabled';
+    const isRequiredTermField = this.props.course.type === 'ClassroomProgramCourse';
 
     // Form components that are conditional on course type
     let expectedStudents;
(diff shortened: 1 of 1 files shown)


ragesoss:
It looks like this change would modify the validation behavior of both Programs & Events Dashboard and Wiki Education Dashboard? We still want to validate the presence of a `term` in wiki_education mode.

Formasitchijoh:
@ragesoss i have checked and the term field is only available for the Wiki Education Dashboard, not in the Programs and Events Dashboard form, please clarify me if i am wrong

Formasitchijoh:
@ragesoss i have updated this PR, sorry i miss understood the task, so I realized the field you were refering to in this case is the when field when copying  a course from an external source in the P&E dashboard, 

I have relaxed the validation by removing the required on the input field, i retained the slug validation to ensure if a user decided to put in a value it should be valid. and the validation of the Wiki education dashboard is maintained

please i would appreciate you feedback

app/assets/javascripts/components/overview/course_cloned_modal.jsx
```diff
@@ -339,7 +339,6 @@ const CourseClonedModal = createReactClass({
           onChange={this.updateCourse}
           value={this.state.course.term}
           value_key="term"
-          required={true}
```
ragesoss, line 342: We do still want to require it for `wiki_education` mode.

Formasitchijoh: Alright i have updated that

Issue #651, opened 2016-02-05
Frontend day calculations for week meetings does not exclude days after end date


ragesoss:

Alone-Y154:
@ragesoss , I would like to work on this issue

Nikuunj:
Hi @ragesoss,

I hope this message finds you well. I'm interested in joining the WikiEducationFoundation Slack channel to stay informed and contribute to the discussions. Could you please add me to the channel?

Looking forward to being a part of the community.

DanAletor:
Hi @ragesoss, I would like to work on this issue

shishiro26:
Hii @ragesoss, what's the status of this ? I would like to raise a  contribution

ragesoss:
@shishiro26 it's still open. You're welcome to work on it.

shishiro26:
@ragesoss, could you provide some examples for this? I'm unable to reproduce the issue. I tried cloning a course, but that doesn't seem to be the correct example I should be working with. I've created several courses, but I'm still unable to reproduce the issue. The photos I included in my PR are related to the course I cloned, but it seems that's not the right example to use, especially since the Travis CI build is also failing.

ragesoss:
Here's what it looks like:

![last week](https://github.com/user-attachments/assets/7bd5b00a-eb6e-4e83-9f6b-fcd032807d21)


The timeline includes weeks that extend to the end date. The last week, as shown in the week header, doesn't go a full week, but only 03/30 - 04/01 (the end date). However, the meeting days are set to every day of the week, and these include days after the end date (Wednesday, Thursday, Friday and Saturday) instead of the expected behavior of showing only the meeting days up to the end date.

You should be able to recreate this locally by setting up a course with a timeline, then adding content to the timeline until it extends to the final week (with appropriate settings for meeting days and an end date that occurs in the middle of that week.

nexpectArpit:
@ragesoss is this issue open for contribution?

PR #6020, opened 2024-10-29
feat(milestones):add dates to milestones section

shishiro26:
## What this PR does

This PR addresses the issue of missing date information in the Milestones section on the Home tab. Previously, while the Milestones section displayed the week numbers derived from the Timeline data, it did not include the corresponding start and end dates for each milestone block.

## Screenshots
Before:
image

After:
Screenshot from 2024-10-29 23-55-16

#279


ragesoss:
Can you add some screenshots that show that the dates are the same in the Milestones vs the Timeline and This Week?

The tricky part of this particular issue is to guarantee that we're using the same algorithm to calculate the dates everywhere they are shown — ideally, only calculating them once, or if done multiple times, then using the same function with the same inputs.

shishiro26:
> Can you add some screenshots that show that the dates are the same in the Milestones vs the Timeline and This Week?

Timeline:
image
Milestone:
Screenshot from 2024-10-30 00-18-26

>The tricky part of this particular issue is to guarantee that we're using the same algorithm to calculate the dates everywhere they are shown — ideally, only calculating them once, or if done multiple times, then using the same function with the same inputs.

I am using the same `DateCalculator` that is used in the Timeline section.

app/assets/javascripts/components/overview/milestones.jsx
```diff
@@ -28,9 +30,11 @@ const Milestones = createReactClass({
     const weekNumberOffset = CourseDateUtils.weeksBeforeTimeline(this.props.course);
     const blocks = [];
 
-    this.props.allWeeks.map((week) => {
+    this.props.allWeeks.map((week,navIndex) => {
```
ragesoss, line 33: Where does this come from? In Timeline.jsx, it the result of some complex code in the render cycle for handling empty weeks.

ragesoss:
You've shown the first week, but it's important to check for how it handles empty weeks as well.

app/assets/javascripts/components/overview/milestones.jsx
shishiro26: The `navIndex` is generated by the `map` function, which helps identify unique elements in the array. Additionally, the `DateCalculator` requires the `navIndex` prop for its calculations. Upon cross-checking in the Timeline page, I confirmed that the `navIndex` is simply the index number generated by the `map` function.

shishiro26:
> You've shown the first week, but it's important to check for how it handles empty weeks as well.

I apologize, but I did not check how the implementation handles empty weeks, as every week currently contains assignments. Could you please provide me with an example of what an empty week looks like?

ragesoss:
Empty weeks can come at the beginning of a course, by including a gap between course start and timeline_start, and also in the middle of the course by using the calendar (via Edit Course Dates on the Timeline tab) to create weeks with no meeting days.

shishiro26:
> You've shown the first week, but it's important to check for how it handles empty weeks as well.

In the Milestones section, empty weeks are not displayed 

Milestones: 
image

Timeline:
image


I also tried using another course for this: [University of Pennsylvania Medical Missionaries to Community Partners (Fall 2023)](https://dashboard.wikiedu.org/courses/University_of_Pennsylvania/Medical_missionaries_to_Community_Partners_(Fall_2023)/home).

ragesoss:
cool. this looks like it's working properly. do you see a good way to avoid repeating the date calculations so that both the timeline and the milestones component can use the same value without a risk that they will diverge with later changes to the timeline component? i think it's dangerous to have these hard-to-understand parallel implementations.

shishiro26:
> cool. this looks like it's working properly. do you see a good way to avoid repeating the date calculations so that both the timeline and the milestones component can use the same value without a risk that they will diverge with later changes to the timeline component? i think it's dangerous to have these hard-to-understand parallel implementations.

Yes, but it's only two lines of code. I hope there won't be any divergence in the future since I guess it’s a straightforward implementation.
  const dateCalc=new DateCalculator(this.props.timelineStart, this.props.timelineEnd,navIndex,{
        zeroIndexed:true
      });

ragesoss:
@shishiro26 I think the ideal strategy would be to do whatever data manipulation and calculation is necessary in a `selector` that operates on data from the redux store, so that the same selector (which can also memo-ize the result) is used in both cases. However, that likely requires a significant refactor of the Timeline component. I think I'm okay with this current implementation that you have done, but it should include some explanatory comments to note how this implementation is coupled to the code in the Timeline component and that we need to make sure both have the same results and account for empty weeks the same way.

ragesoss:
@shishiro26 I think it's wise to do the minimum amount of refactoring necessary to unify this particular calculation, and if you notice anything along the way that you think ought to be refactored, make a note of it.

shishiro26:
@ragesoss, I will ensure that I do the minimum necessary refactoring and update the commit accordingly. I will also make sure to note any important refactoring opportunities if they arise.

shishiro26:
@ragesoss I have updated the code. Could you please check if this aligns with what we discussed?

app/assets/javascripts/components/overview/milestones.jsx
```diff
@@ -4,6 +4,7 @@ import createReactClass from 'create-react-class';
 import { filter } from 'lodash-es';
 
 import CourseDateUtils from '../../utils/course_date_utils';
+import DateCalculator from '../../utils/date_calculator';
```
ragesoss, line 7: Is this still needed?

app/assets/javascripts/components/overview/milestones.jsx
```diff
@@ -28,19 +31,26 @@ const Milestones = createReactClass({
...
     const blocks = [];
 
-    this.props.allWeeks.map((week) => {
+    
+    this.props.allWeeks.map((week,navIndex) => {
       if (week.empty) return null;
-
       const milestoneBlocks = filter(week.blocks, block => block.kind === this.milestoneBlockType);
       return milestoneBlocks.map((block) => {
+        if (
```
ragesoss, line 39: It looks we're doing something similar in both this component and the Timeline, checking for whether to map a particular block to `null`. Is this something that can become shared logic in the selector, or do they handle these cases differently?

app/assets/javascripts/components/overview/milestones.jsx
```diff
@@ -28,19 +31,26 @@ const Milestones = createReactClass({
...
+          !this.props.weekDates[navIndex]?.start || 
+          !this.props.weekDates[navIndex]?.end || 
+          this.props.weekDates[navIndex]?.start > this.props.weekDates[navIndex]?.end
```
ragesoss, line 42: This is quite verbose, but it's not obvious what this if statement is for. What situation(s) for a block are being checked for here?

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -388,16 +389,24 @@ const Timeline = createReactClass({
...
 
-      const dateCalc = new DateCalculator(this.props.course.timeline_start, this.props.course.timeline_end, navIndex, { zeroIndexed: true });
       const navWeekKey = `week-${navIndex}`;
       const navWeekLink = `#week-${navIndex + 1 + weeksBeforeTimeline}`;
+      const startDate = this.props.weekDates[navIndex]?.start;
+      const endDate = this.props.weekDates[navIndex]?.end;
+
+      if (
+        !startDate || 
+        !endDate
```
ragesoss, line 399: Why were these checks not necessary in the current implementation?

app/assets/javascripts/components/overview/milestones.jsx
shishiro26: This check was added to prevent any date from showing after the timeline's end date in the milestone.

app/assets/javascripts/components/overview/milestones.jsx
shishiro26: Nope,I forgot to remove the DateCalculator. I'll go ahead and remove it now, sorry about that.

app/assets/javascripts/components/overview/milestones.jsx
shishiro26: Actually, the problem I encountered is that in the selector, I needed the index for the weeks to display, so I used `weeks` in the `const getWeeks = state => state.timeline.weeks;` function to memoize all the results. However, the issue is that the `allWeekDates` being sent from `getAllWeekDates(state)` contains weeks that are after the timeline end date, but the `getWeeks` function doesn't include them. 

To handle this, when creating the `DateCalculator` instance:

```javascript
const dateCalc = new DateCalculator(course.timeline_start, course.timeline_end, index, {
  zeroIndexed: true
});
```

I found that the last indexes didn't have any dates, causing it to be empty when trying to render in the milestones and timeline. To fix this, I added a check for that in both files.

app/assets/javascripts/components/overview/milestones.jsx
ragesoss: Hmm... I don't think that' necessary. The Timeline can handle dates that are after the end date (in which case, they get highlighted in red). I think in the Milestones component, it would be better to show dates that stretch beyond the end date, when that happens.

shishiro26: Okay, I will update the part where the deadline extends. Should I display the extended deadline within the milestone like in the below image, or should I show it after the timeline's end date?
image

ragesoss: I don't think any special formatting is needed in the Milestones component.

shishiro26: Yes, I have updated it. Could you please check it?

ragesoss:
Build is currently failing because of JS linting errors.

shishiro26:
I would like to do this since I have been using Docker. Could you provide me with the steps to run lint in Docker?

ragesoss:
You're kind of on your own with that, as we don't have up-to-date docs for docker, and I don't use it.

shishiro26:
I have resolved all the linting issues, @ragesoss.

shishiro26:
@ragesoss I came across an issue while fixing the lint errors. I saw the line lint-non-build": "eslint 'test/**/*.{jsx,js}' '*.js'" and I’m not sure if this is the only line responsible for linting, as I couldn't find any other run functions in the codebase. Last time, when I encountered lint issues, I fixed everything explicitly. Could you guide me on what I need to do for this?

ragesoss:
The failing tests seem to be related to this change.

shishiro26:
I hope these new changes work now, @ragesoss! 🤞🤞

app/assets/javascripts/components/timeline/timeline.jsx
```diff
@@ -397,7 +398,7 @@ const Timeline = createReactClass({
       if (usingCustomTitles) {
         let navTitle = '';
         if (weekInfo.emptyWeek) {
-          const datesStr = `${dateCalc.start()} - ${dateCalc.end()}`;
+          const datesStr = `${weekDates[navIndex].start} - ${weekDates[navIndex].end}`;
```
ragesoss, line 401: I note that his branch of code is being switched to use weekDates, but another branch below uses dateCalc. Is that the intended/correct behavior?

shishiro26: Thank you for pointing this out! I don't recall the exact branch in question, but this is the updated and correct code. The change to weekDates is intentional and reflects the latest update to the logic.

ragesoss: What I mean is, is it intention to switch from dateCalc to weekDates in only one branch, or should it be done in both?

shishiro26: It should be implemented in only one branch, not both.

ragesoss: Okay. Why?

shishiro26: The other branch is the older one, and I have been working on different changes there. In this branch, I’ve completely rewritten the code, while the other branch continues to focus on the previous changes or issues

ragesoss: I feel a little lost here. Why can't the calls to dateCalc in the `else` branch below also be switched to use weekDates?

shishiro26: Apologies for any confusion on my end. I currently have two branches in development: one focused on resolving the current issue, and the other working on issue #651. The second branch is entirely separate, and I need to make some changes to it. As a result, I've converted that branch into a draft and have been focusing solely on the first branch for now. Are we on the same page, or is there still some misunderstanding on my part?

shishiro26: Apologies for any confusion on my end. I currently have two branches in development: one focused on resolving the current issue, and the other working on issue #651. The second branch is entirely separate, and I need to make some changes to it. As a result, I've converted that branch into a draft and have been focusing solely on the first branch for now. Are we on the same page, or is there still some misunderstanding on my part? This is why I've added the weekDates to this branch and not the other branch .

ragesoss: Oh! Sorry, I meant logical branch of the code, as in the `if/else` statement that starts at line 399. This PR changes the `if` portion but does not change the `else` portion, and I think it should probably change both.

shishiro26: Thanks for the clarification! I’ve updated both the if and else parts as suggested and pushed the changes. Let me know if anything else is needed! 😅👍

app/assets/javascripts/components/course/course.jsx
```diff
@@ -27,6 +27,7 @@ import { getStudentCount, getCurrentUser, getWeeksArray } from '../../selectors'
 import ActivityHandler from '../activity/activity_handler';
 import CourseApproval from './course_approval';
 
+
```
ragesoss, line 30: Please remove this.

app/assets/javascripts/selectors/index.js
```diff
@@ -1,11 +1,28 @@
 import { createSelector } from 'reselect';
-import { sortBy, difference, uniq, filter, includes, pickBy, find } from 'lodash-es';
+import {
```
ragesoss, line 2: I'd prefer not to include all these formatting changes alongside the new feature.

app/assets/javascripts/selectors/index.js
```diff
@@ -326,34 +401,56 @@ export const firstValidationErrorMessage = createSelector(
...
+    (acc, ticket) => ({ ...acc, [ticket.id]: ticket }),
+    {}
+  );
+});
+
+export const getAllWeekDates = createSelector(
+  [getAllWeeksArray, getCourse],
+  (weeksArray, course) => {
+    return weeksArray.map((_, index) => {
+      const DateCalc = new DateCalculator(
```
ragesoss, line 442: By convention, this DateCalc object should start with a lower-case letter.

ragesoss:
I am happy with this implementation, and I'm ready to merge it after cleanup of the minor issues in my latest comments.

Issue #6024, opened 2024-10-31
Add Tooltip for Article Views Metric


empty-codes:
### Current Behavior:
While there are tooltips for other metrics like "Views" that explain their estimated nature, there is currently no tooltip for Article Views 

![image](https://github.com/user-attachments/assets/7d5b0f04-bc5a-4bb8-a106-266d9a4f41f8)

### Desired Behavior:
Add a tooltip to provide users with context about how the numbers are calculated

empty-codes:
I am working on it!

PR #6025, opened 2024-10-31
Feature: Add tooltip to Article Views metrics on relevant pages #6024

empty-codes:
## What this PR does
Resolves Issue #6024 by adding a tooltip for Article Views to provide users with context about how the numbers are calculated.

Note: I added the tooltip  only to pages where 'References Added' also has a tooltip and the `metrics.view_count_description` specifically is used

## Screenshots
Before:
image

For wikidata:
image

After:
image

For wikidata:

image

### DIFF
--- a/app/assets/javascripts/components/overview/overview_stats.jsx
+++ b/app/assets/javascripts/components/overview/overview_stats.jsx
@@ -136,6 +136,8 @@ const OverviewStats = ({ course }) => {
         stat={course.view_count}
         statMsg={I18n.t(`metrics.${ArticleUtils.projectSuffix(course.home_wiki.project, 'view_count_description')}`)}
         renderZero={false}
+        info={I18n.t(`metrics.${ArticleUtils.projectSuffix(course.home_wiki.project, 'view_count_doc')}`)}
+        infoId="view-count-info"
       />
       {uploadCount}
     </div>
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/components/overview/overview_stats.jsx
```diff
@@ -136,6 +136,8 @@ const OverviewStats = ({ course }) => {
         stat={course.view_count}
         statMsg={I18n.t(`metrics.${ArticleUtils.projectSuffix(course.home_wiki.project, 'view_count_description')}`)}
         renderZero={false}
+        info={I18n.t(`metrics.${ArticleUtils.projectSuffix(course.home_wiki.project, 'view_count_doc')}`)}
```
ragesoss, line 139: This seems like it will break for Wikidata, since there's no corresponding message for when that suffix is present?

empty-codes: That's true. I can create a:

view_count_doc_wikidata: This is the estimated number of views based on a 30-day average for each item, through the most recent stats update. Views may decrease if the updated average is lower than previous counts.

Is this text fine?

ragesoss:
Looks good. Can you include screenshots of both variants of the message?

empty-codes:
> Looks good. Can you include screenshots of both variants of the message?

Done!

PR #6027, opened 2024-11-02
Fix: Failing test in pull request  #5783

Formasitchijoh:
## What this PR does
This PR fixes the failing test from pull request #5783, and update functionality of sending  instructor alerts. All necessary tests have been written to cover various scenarios and edge cases.

## ScreenShot of Email Sent
Screenshot 2024-11-19 at 05 56 07

Screenshot 2024-11-19 at 05 57 14

@ragesoss please can you review this when you are chanced


app/mailers/instructor_notification_mailer.rb
```diff
@@ -2,27 +2,28 @@
...
   def email(alert)
     @alert = alert
     set_email_parameters
     params = { to: @instructors.pluck(:email),
-               subject: @alert.subject }
-    params[:reply_to] = @alert.sender_email unless @alert.sender_email.nil?
-    unless @alert.sender_email.nil?
-      params[:bcc] = @alert.sender_email # sender_email gives user email of the sender
+               subject: @alert[:details][:subject] }
+    unless @alert[:details][:bcc_to_salesforce] == false
```
ragesoss, line 14: I suggest using `dig` here rather than a series of `[]` calls, so that it will handle it gracefully if this method is called on an alert record that lacks the expected bbc_to_salesforce field (such as an older record from before the PR).

app/mailers/instructor_notification_mailer.rb
```diff
@@ -2,27 +2,28 @@
...
+      params[:bcc] = @alert.bcc_to_salesforce_email
     end
+    params[:reply_to] = @alert.sender_email unless @alert.sender_email.nil?
+    Rails.logger.info "confirm : #{params.inspect}"
```
ragesoss, line 18: Be sure to remove logging that was used for testing.

spec/mailers/previews/instructor_notification_preview.rb
```diff
@@ -3,7 +3,7 @@
 
 class InstructorNotificationPreview < ActionMailer::Preview
   def message_to_instructors
-    InstructorNotificationMailer.email(example_alert)
+    InstructorNotificationMailer.email(example_alert, true)
```
ragesoss, line 6: It looks like the intention was to add a second parameter to the mailer's email method, but it looks like this wasn't implemented. (And the fact that no test for it failed suggests that it's not being covered in the tests.) I would suggest using a named parameter for this, so that it's obvious what `true` means in this context. Or maybe this shouldn't have an extra parameter since the logic for it is implemented within the email method, and this preview method is just broken.

app/mailers/instructor_notification_mailer.rb
Formasitchijoh: Alright, i surely missed that i will , update this

app/mailers/instructor_notification_mailer.rb
Formasitchijoh: yes, i will, thank you 🙏

spec/mailers/previews/instructor_notification_preview.rb
Formasitchijoh: Your right, thank you @ragesoss  for pointing this out, I think this is where my blocker has been, where the test I have been writing for this feature doesn't pass 😀,  since i was working and trying to fix the failed test in @prathamVaidya  PR,

This PR is still a WIP, as test to cover the feature has not yet been written.

ragesoss:
Thanks! Can you rebase this so that in only includes the changes relevant to the new feature?

Formasitchijoh:
> Thanks! Can you rebase this so that in only includes the changes relevant to the new feature?

@ragesoss i've rebased to the feature changes

Formasitchijoh:
Hello @ragesoss 
Just checking  in regarding this task. Is there anything specific I need to do to complete it or any additional updates required from my end?

ragesoss:
I still need to manually test this. However, I note that this PR includes an unexpected removal of a large number of messages from en.yml. Perhaps in needs to be rebased?

Formasitchijoh:
> I still need to manually test this. However, I note that this PR includes an unexpected removal of a large number of messages from en.yml. Perhaps in needs to be rebased?

Alright updating that

ragesoss:
It looks like you've squashed the commits from @prathamVaidya's PR into the first commit, but it also includes additional changes that you've made? I'd prefer to either leave the commits from the other PR as separate commits, or squash them into a single one that represents only those commits, and add any additional changes in new commits.

config/routes.rb
```diff
@@ -235,6 +235,8 @@
     end
     collection do
       post 'notify_instructors'
+      post 'reply'
```
ragesoss, line 238: I'm confused about this. Is it used by the frontend?

Formasitchijoh:
> It looks like you've squashed the commits from @prathamVaidya's PR into the first commit, but it also includes additional changes that you've made? I'd prefer to either leave the commits from the other PR as separate commits, or squash them into a single one that represents only those commits, and add any additional changes in new commits.

Alright I will do that

config/routes.rb
Formasitchijoh: Is was meant for replying to an email, since i was referencing the ticket system, but I have removed it as the front end does not use it currently

spec/mailers/previews/instructor_notification_preview.rb
ragesoss: Are you still working on addressing this?

Formasitchijoh:
Could i start with something else while you are reviewing this

ragesoss:
> Could i start with something else while you are reviewing this

Yes

ragesoss:
Looks good! Thanks @prathamVaidya and @Formasitchijoh !

PR #6040, opened 2024-11-20
chore: remove explicit devTools config, use default behavior

Abishekcs:
## What this PR does

Removes the explicit devTools configuration from the Redux store, as [configureStore enables it by default.
](https://redux.js.org/tutorials/essentials/part-3-data-flow#creating-the-store)

Screenshot from 2024-11-20 13-27-33

Before:

https://github.com/user-attachments/assets/4c3a2dc6-83f1-4380-8f37-209fbddadaa8

After:

https://github.com/user-attachments/assets/7c0044bb-b8c8-4543-b454-bd78ed34d0ee

### DIFF
--- a/app/assets/javascripts/components/util/create_store.js
+++ b/app/assets/javascripts/components/util/create_store.js
@@ -43,8 +43,6 @@ export const getStore = () => {
       immutableCheck: enableMutationChecks,
       serializableCheck: enableMutationChecks,
     }),
-    // Enable Redux DevTools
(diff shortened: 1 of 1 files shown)


ragesoss:
Did you confirm that this works in the version of Redux we're using?

Abishekcs:
> Did you confirm that this works in the version of Redux we're using?

I did. I will attach a screen recording showing the behavior before and after.

PR #6047, opened 2024-11-26
Fix campaigns list rendering above recent news section  issue on index page when navigating back via cancel button

Abishekcs:
## What this PR does

This pull request fixes an issue where the campaigns list sometimes rendered incorrectly above the Recent News section when navigating back via the cancel button. The campaigns list now consistently appears below Recent News.

## Screenshots

Before:

https://github.com/user-attachments/assets/9900e65a-7b4b-40d5-b0d6-7d19f4a2487e

After:

https://github.com/user-attachments/assets/ce412455-3a45-4e0a-afaf-a949278eaff6

## Cause of Bug :

The issue occurred because, in the HAML file, the HTML for the campaigns list was placed above the HTML for the Recent News section. This behavior was observed when the request path included /course_creator.

### DIFF
--- a/app/views/dashboard/index.html.haml
+++ b/app/views/dashboard/index.html.haml
@@ -1,11 +1,5 @@
 - content_for :after_title, " - #{t("application.my_dashboard")}"
 
--# this is the react root when the route is /course_creator
-- if request.path.include?('/course_creator')
-  #react_root{data: {default_course_type: @pres.default_course_type,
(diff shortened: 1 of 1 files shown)


Abishekcs:
> except when you navigate back this route via a cancel button

@ragesoss Just to be clear, 😅 it should render when we navigate via the cancel button, but it should appear below the Recent News section?

PR #6054, opened 2024-12-08
Refactor components to use navigation utility hook for student details view

### LINKED ISSUE #6045: Refactor common logic for navigating to the single-student view within a Course page
We have at least two places that use username links to navigate to the individual student view of the Students tab of a course page. See https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6042#pullrequestreview-2459488373

We also have username links in the MyArticles section and other places that show the group members for an assigned article. The ones that link to user profile pages (/users/USERNAME) should link to the individual student view within the course instead.

Abishekcs:
closes #6045

## What this PR does

This pull request introduces the usage of the `useNavigationsUtils hook` in multiple components to enable navigation to the student details view. The hook provides the `openStudentDetailsView function`, which constructs the URL for the single student view page within a course page and navigates to it.

## Screenshots
Before:

https://github.com/user-attachments/assets/bb774d48-39df-45e0-8916-ef4939abdb61

After:

https://github.com/user-attachments/assets/69577c08-3314-4cf3-a962-d0af3f197528

Additionally, the `ReviewerLink` component, which is supposed to navigate to user profile pages `(/users/USERNAME)`, does not seem to be used anywhere in the codebase (though I’m not 100% sure). However, based on a search through the code, the only place it's mentioned is within the `AssignmentsLinks` component, where it is currently commented out, as shown below:

`// const reviewers = <ReviewerLink key={`reviewers-${id}`} reviewers={assignment.reviewers} />;`

### DIFF
--- a/app/assets/javascripts/components/common/AssignmentLinks/AssignmentLinks.jsx
+++ b/app/assets/javascripts/components/common/AssignmentLinks/AssignmentLinks.jsx
@@ -59,9 +59,9 @@ const AssignmentLinks = ({ assignment, courseType, user, course, project, editMo
   let groupMembers;
   if (editors) {
     if (role === ASSIGNED_ROLE) {
-      groupMembers = <GroupMembersLink members={editors} />;
+      groupMembers = <GroupMembersLink members={editors} course={course} />;
     } else {
-      groupMembers = <EditorLink key={`editor-${id}`} editors={editors} />;
+      groupMembers = <EditorLink key={`editor-${id}`} editors={editors} course={course} />;
     }
   }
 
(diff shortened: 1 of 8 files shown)


ragesoss:
Hmm... looks like you're correct about ReviewerLink. I think the implementation changed as it was being worked on initially, and GroupMembersLink does what that was initially designed to do. Feel free to open a PR removing that commented-out code along with the unused component.

This implementation of the utility hook looks reasonable. I'll test it out soon.

Abishekcs:
> Is it possible to make them behave more like links, with cursor changing upon mouseover, and ideally displaying the target URL as well, like a browser does with conventional links?

I believe it will be possible, I will give it a try.😄

Abishekcs:
image


https://github.com/user-attachments/assets/561164e0-00fd-459a-aa56-dbf7c5a1ee75


I tried this approach as shown in the screenshot by using the `href`, and it works. However, with the student/editor table, it won't be possible since those are `td` tags."

Wrapping each child of the `td tags` (table data) with an anchor tag will work, but that will leave the `tr tag` (table row) without a clickable link (which is the current implementation).

ragesoss:
Okay. I guess this is a broader issue with how we use the `td` tags.

PR #6055, opened 2024-12-09
Improve/Fix Accessibility for WeekdayPicker Component for Screen reader

### LINKED ISSUE #6051: WeekdayPicker state is not fully accessible via screen reader
When using a screen reader and selecting or unselecting the meeting weekdays, the screen reader does not detect/read out the updated state.

Abishekcs:
closes #6051

## What this PR does

This PR fixes the accessibility of the Calendar and WeekdayPicker components by introducing the `aria-label` and removing `aria-pressed`. And prop `ariaModifier` which was not being used was replaced by  a simple hardcoded string which is used to detect if a "weekday" is selected.

<hr/>

## Cause of the Bug:

The `ariaModifier` prop (of type `PropTypes.string`) required by the WeekdayPicker component was not being passed from the Calendar component or any of its parent components. This prop is essential for determining whether a button has been pressed and updating the `aria-pressed` attribute accordingly. As a result, the `aria-pressed` attribute remained `false` even when the button was pressed because it depended on the correct value of `ariaModifier.` Which was not passed in the first place itself.

<hr/>

## Screenshots

Before:

https://github.com/user-attachments/assets/624f4bbf-298b-4b27-9c6c-64f7376e69fd

After:

https://github.com/user-attachments/assets/1068fde8-25a6-4252-8a63-fb5120ddec8a

I applied CSS to hide the `aria-live` div (weekday_picker.jsx), using a style similar to that already used for hiding admin notes. Although the two elements use different class names, the CSS properties are the same. I considered using a single shared CSS class for both cases but was uncertain about the best file to place this shared CSS in.

### DIFF
--- a/app/assets/javascripts/components/common/weekday_picker.jsx
+++ b/app/assets/javascripts/components/common/weekday_picker.jsx
@@ -25,7 +25,6 @@ const WeekdayPicker = ({
   style,
   tabIndex,
 
-  ariaModifier,
   modifiers,
 
   locale,
@@ -145,7 +144,7 @@ const WeekdayPicker = ({
 
     dayClassName += customModifiers.map(modifier => ` ${dayClassName}--${modifier}`).join('');
 
-    const ariaSelected = customModifiers.indexOf(ariaModifier) > -1;
+    const ariaSelected = customModifiers.indexOf('selected') > -1;
 
     let tabIndexValue = null;
     if (onWeekdayClick) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/common/calendar.jsx
```diff
@@ -151,6 +152,7 @@ const Calendar = createReactClass({
 
     const onWeekdayClick = this.props.editable ? this.selectWeekday : null;
     const onDayClick = this.props.editable ? this.selectDay : null;
+    const ariaModifier = this.props.useAriaLabel ? 'selected' : null;
```
ragesoss, line 155: Why is it necessary to make changes outside of the WeekdayPicker component for this?

Abishekcs: Actually, I reviewed the code again and realized this change is not necessary. I have removed the existing `ariaModifier` since the `WeekdayPickerComponent` was not using it at all, and the Calendar component was not passing it either. I replaced it with the hardcoded string `selected`.

Additionally, I removed `aria-pressed` and replaced it with `aria-label`  & `aria-live`. 

Also, since the WeekdayPickerComponent is only used by the Calendar component, this change should not cause any issues.

Can you please review the updated code whenever you have time? Thank you 😃

Issue #5549, opened 2023-12-10
Restore removed self-assigned articles to the pool of available articles 


Ibom99:
### Current Behavior:
At the moment, the programs and event dashboard doesn't move self-assigned articles back to the pool of available articles when remove is selected, it deletes the article completely from the program.

### Desired Behavior:
The self-assigned articles when removed should return to the list of available articles it should not be deleted from the program entirely. 

### Additional context:
1. Self-Assigned Article

![Screenshot (70)](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/104465078/a1b9a0f9-7095-40e0-909b-0df03d2f894d)

2. After selecting remove 

![Screenshot (71)](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/104465078/17cf54b1-2926-43c7-94aa-33cbe16ba03f)

3. Article not in the list of the available articles
![Screenshot (72)](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/104465078/c16e82e4-caf9-4d83-8d6f-6320cea04aee)

ragesoss:
This feature would touch a lot of files, mostly involving the `Assignment` model, along with related controllers and front-end components.

fillingtothemomo:
@ragesoss can I take this issue up?

ragesoss:
@fillingtothemomo yes, but I think this is a fairly complicated issue that, and it may take a number of iterations to figure out exactly what the implementation should look like.

fillingtothemomo:
@ragesoss Can you suggest some newcomer friendly issues for me to work on?

ragesoss:
@fillingtothemomo I think you should give #5489 a try. It looks like the other person isn't active on it anymore.

ragesoss:
A few clarifications of this issue:

1. This should account for the possibility that the removed article is already present as an Available Article (and should simply remove the article, in that case).
2. This new behavior should only apply to courses that are using Available Articles in the first place, but we don't actually have a good way to detect that option right now. Basically, we want this new behavior only for a course where the instructor chose the 'Instructor prepares a list' option in the assignment wizard. I think a good way to address this would be to add `choose_articles_from_list` as a tag in the `researchwrite/wizard.yml` file so that the courses where this should apply will have the `choose_articles_from_list` Tag. Then the behavior can be controlled on the backend in the controller by determining whether the course has that tag.

shishiro26:
[...]

PR #6060, opened 2024-12-12
fix:restored removed article to available articles

shishiro26:

## What this PR does
This PR restores both self-assigned and admin-assigned articles back to the pool of available articles.
I have written a test case for the controller to ensure that the changes work as expected.

## Screenshots
After:
https://github.com/user-attachments/assets/db77ecb3-734d-4693-94d2-cfd235297c90

With this change, when removing an article, a dialog appears with the message: *"Are you sure you want to delete this program?"* 
- Do we need to modify the dialog message to reflect the new functionality? 

#5549


shishiro26:
In this PR, I have added Ruby test cases for the `unclaim` method in the `AssignmentsController`. 


The scenarios covered are as follows:
1. **When the article is assigned to the user by the admin**:  
   - The admin removes the assignment.
   - https://github.com/user-attachments/assets/7bbd0d4e-bac7-4f56-9066-af09b595dc84
2. **When the article is assigned to the user by the admin**:  
   - The user removes the assignment.
   -  https://github.com/user-attachments/assets/392911db-789f-44dc-81c7-84f73232d7a6
3. **When the article is self-assigned**:  
   - Either the user or the admin removes the self-assigned article.
   - https://github.com/user-attachments/assets/ea029f97-58a5-455c-996d-4d46ee9ee9a2

shishiro26:
@ragesoss  the test is passing locally in my system 
image

shishiro26:
@ragesoss, I'm not entirely sure, but the multiwiki_assignment_spec.rb test seems to be failing after the latest commit merge. Additionally, the account_requests_spec.rb test is failing both on my local setup and in the master branch. However, this failure appears to be unrelated to the changes I made.

ragesoss:
@shishiro26 yes, i'm not sure why it started failing, but i'm going to look into it.

ragesoss:
Does this implementation differentiate articles that started as Available Articles versus ones that the student chose via title without starting as an Available Article? I think we need this to only apply for ones that started as Available Articles.

shishiro26:
I’m not sure if you’re referring to this one, sir

https://github.com/user-attachments/assets/d07b9a8b-1284-459d-8ba4-cfd91b9d3aa1


If so, then yes, it doesn’t differentiate between the two. It removes both from the assigned ones and keeps them in the available articles, sir.

shishiro26:
Sir, I would like to know if there is any user-specific association, such as a user_id or something similar, that tracks whether a particular person created or added an article. Is there an existing reference to identify the user who created or added the article? From the logs I’ve seen, there are references to users, but they seem to be related to assignments rather than article creation. If such an association doesn’t currently exist, would it be possible for me to add a user_id field to the articles table and link it to the user who created or added the article?

ragesoss:
I think you could use `Assignment#flags` to keep track of whether an Assignment record began as an Available Article.

shishiro26:
Sir, I believe there is no change in the flags field, as it remains NULL in both cases in the SQL database. There appears to be no update to Assignment.flags.

ragesoss:
Right, you would need to implement that.

shishiro26:
okay, sir! 😄

shishiro26:
Could you please review this, sir? This is just a rough implementation. If this method looks fine, the `unclaim` method (which removes the article if it is not in the available articles pool, and if it is in the pool, the assignment is simply unclaimed and added back to the available articles), and if the `assignment#flags` are set correctly, I will refine the code and update the RSpec tests for the final version.

ragesoss:
I will plan to review this next week.

shishiro26:
Okay sir

ragesoss:
This looks like it's going in the right direction, but from my reading of the code, I think it doesn't correctly the initial setting of flags. An "Available Articles" is an Assignment record created by the instructor of a course (or an admin) that is not linked to a user. These are the ones that are available as a list for students to select as their assignment. In contrast, a newly created assignment that is linked to a user, which happens when a student enters an article title directly rather than selecting from the Available Articles list, does not start as an Available Article. Those two situations are what the flag should keep track of: whether a new Assignment started as Available Article or not.

I think feature specs should be extended to test for the new behavior. One should test for the presence of the flag when an instructor adds an Available Article, and the absence when a student adds an assigned article directly. Another should test for a student claiming and then unclaiming an assignment, verifying that the article returns to the Available list. These might be possible to test via adjusting or extending existing tests that interact with assignment features, such as my_articles_spec.rb and instructor_role_spec.rb

shishiro26:
Thank you for the clarification. I realized I had overlooked this situation earlier. When adding assignments, I mistakenly assumed they were directly linked to articles rather than being part of the assignments workflow. Upon reviewing, I noticed that when adding an available_article, the user_id is set to nil and it gets added to the assignments. I’ll address this issue and ensure the flags are set correctly to distinguish between "Available Articles" and directly assigned articles.

Additionally, I’ll review and update the tests to align with the expected behavior. I now have a clearer understanding of the logic, but I’ll double-check the code and adjust or extend the relevant tests before committing the changes.

shishiro26:
-  When a user tries to assign themselves an available article and then unclaims it: 
    https://github.com/user-attachments/assets/eaffb8cf-0cc5-46e7-8942-d0109021609c

-  When a user tries to assign themselves an unavailable article and then unclaims it:
   https://github.com/user-attachments/assets/f9827066-5aff-4f80-9a3a-690e2dcb5b4f

-  When an admin assigns a user an available article and the user (or admin) unclaims it:
    https://github.com/user-attachments/assets/35ac1113-8658-49c9-9d45-83758ea3c40e

-  When a user tries to assign an unavailable article to another user:
  https://github.com/user-attachments/assets/bcc4857e-8c61-4d90-bbdb-e29b7aa4e9a9

 I've also updated the test cases to reflect these scenarios. However, I am unsure if I should update the flags in the `create_assignment` within the `assignment_controller_spec.rb`. Could you clarify this?

ragesoss:
> I am unsure if I should update the flags in the `create_assignment` within the `assignment_controller_spec.rb`. Could you clarify this?

I'm not sure what you're referring to here; I don't see `create_assignment` in that spec. However, it would make sense to ensure that spec tests the creation of assignment both with and without a user, and verify the flag is present as expected in the case of creating an assignment without a user. I think that both adding Available Articles and adding user-assigned Assignments both go through the #create method of that controller.

shishiro26:
Sorry sir , I was actually referring to the `create` method in the `assignments_controller_spec.rb`. In the context where the article does not exist, it means the article is being created by the user through a search. So, the assignment should have the article, but the `user_id` should be `nil`. However, I am seeing that the `user_id` is being set to a value, which shouldn't happen.

image

shishiro26:
Sorry, sir. Yes, I'm referring to the create method in the assignments_controller_spec.rb. However, I’m facing an issue where the flags value isn't appearing in any of the contexts in the create test. Even though I’ve set the flags value, I’m unable to see it; instead, it shows as an empty JSON object ({}) in every context of the test. Could you please help me figure out why this is happening?

shishiro26:
I've only added the test for available articles, while the tests for user-assigned assignments are already present, which are not covered in this test, sir.

shishiro26:
Could you please take a look at this, @ragesoss?

ragesoss:
@shishiro26 it's on my todo list, but I likely won't be able to review it until next week. @Abishekcs: if you're interested in having a look, I think your feedback would be helpful in the meantime.

Abishekcs:
> @shishiro26 it's on my todo list, but I likely won't be able to review it until next week. @Abishekcs: if you're interested in having a look, I think your feedback would be helpful in the meantime.

Sure, I will look into it.😄

app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Header.jsx
```diff
@@ -61,7 +66,7 @@ export const Header = ({
         isEnglishWikipedia={isEnglishWikipedia({ assignment, course })}
         isClassroomProgram={isClassroomProgram(course)}
         isComplete={isComplete}
-        unassign={unassign({ assignment, course, deleteAssignment, dispatch })}
+        unassign={unassign({ assignment, course, unclaimAssignment, dispatch })}
```
Abishekcs, line 69: There's no need to pass `unclaimAssignment` as a parameter since it is already imported at the top of the file. Additionally, you can call `useDispatch` (dispatch) directly inside the `unassign` function, as dispatch is not used anywhere else in this file. Moving useDispatch into the function will help eliminate any potential confusion for anyone new to the codebase about whether dispatch is being used elsewhere in this file.

app/views/assignments/_assignment.json.jbuilder
```diff
@@ -4,7 +4,7 @@ json.call(assignment, :id, :user_id, :article_id, :role, :article_url, :sandbox_
           :draft_sandbox_status, :bibliography_sandbox_status, :peer_review_sandbox_status)
 json.assignment_id assignment.id
 json.article_title assignment.article_title.tr('_', ' ')
-
+json.flags assignment.flags
```
Abishekcs, line 7: You can move this within the `json.call(assignment ...)`?

Abishekcs:
> ## Open questions and concerns
> With this change, when removing an article, a dialog appears with the message: _"Are you sure you want to delete this program?"_
> 
> * Do we need to modify the dialog message to reflect the new functionality?

How about adding a "notification message" after successfully unclaiming an assignment that belonged to available articles? 

By dispatching an `ADD_NOTIFICATION` action in the `unclaimAssignment` thunk. This way, the one who removes the article will be clearly notified if an article has been returned to the available pool and don't always have to check the if the article was successfully added back to the available articles.

app/assets/javascripts/actions/assignment_actions.js
```diff
@@ -78,6 +78,12 @@ export const claimAssignment = (assignment, successNotification) => (dispatch) =
     .catch(response => dispatch({ type: types.API_FAIL, data: response }));
 };
 
+export const unclaimAssignment = assignment => (dispatch) => {
```
Abishekcs, line 81: Hmm, I also think a comment would be nice to explain that it not only unclaims an assignment but also returns it to the list of available articles.

Abishekcs:
> I believe this new implementation works well. When an assignment is unclaimed, it won't display any dialog box or confirmation message. Instead, it will directly unclaim the assignment and show a notification.
> image
> 

Probably should keep the existing dialog box and confirmation message. Just add the Notification after the user has confirmed his choice.

app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Header.jsx
```diff
@@ -33,8 +33,15 @@ const isClassroomProgram = course => (course.type === 'ClassroomProgramCourse');
 const unassign = ({ assignment, course, dispatch }) => {
   const body = { course_slug: course.slug, ...assignment };
   const confirmMessage = I18n.t('assignments.confirm_deletion');
+  const successNotification = {
```
Abishekcs, line 36: Moving the successNotification definition inside the unclaimAssignment function would be better. Since this notification is specifically tied to the success state of unclaiming an article, it makes more sense to define it within the function itself rather than passing it as a parameter.

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
```diff
@@ -363,6 +363,14 @@ const AssignButton = ({ course, role, course_id, wikidataLabels = {}, hideAssign
 
   const unassign = (assignment) => {
     const confirmMessage = I18n.t('assignments.confirm_deletion');
+    const successNotification = {
```
Abishekcs, line 366: Same for this too.

app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Header.jsx
shishiro26: Yes, but I noticed that in other parts of the code, the notification is also being sent as a parameter. So, I kept it that way and did not change it.

app/assets/javascripts/actions/assignment_actions.js
```diff
@@ -78,6 +78,21 @@ export const claimAssignment = (assignment, successNotification) => (dispatch) =
     .catch(response => dispatch({ type: types.API_FAIL, data: response }));
 };
 
+// This moves the assigned articles to the 'available articles' section if they are already listed there.
+export const unclaimAssignment = (assignment, successNotification) => (dispatch) => {
+  return API.unclaimAssignment(assignment)
+    .then((resp) => {
+      if (resp.assignment) {
+        if (successNotification) { dispatch(addNotification(successNotification)); }
```
Abishekcs, line 86: You can remove the redundant `if (successNotification)` check.

app/assets/javascripts/actions/assignment_actions.js
```diff
@@ -78,6 +78,21 @@ export const claimAssignment = (assignment, successNotification) => (dispatch) =
     .catch(response => dispatch({ type: types.API_FAIL, data: response }));
 };
 
+// This moves the assigned articles to the 'available articles' section if they are already listed there.
+export const unclaimAssignment = (assignment, successNotification) => (dispatch) => {
+  return API.unclaimAssignment(assignment)
+    .then((resp) => {
+      if (resp.assignment) {
```
Abishekcs, line 85: Additionally, I am not sure but using `assignment` to check for successful unclaim may not be the most reliable indicator. Instead,  consider using `flags` or checking if `user_id` is null/undefined, as these would more explicitly confirm that the article was successfully unclaimed.

app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Header.jsx
Abishekcs: > Yes, but I noticed that in other parts of the code, the notification is also being sent as a parameter. So, I kept it that way and did not change it.

ohh but still it would be nice to move as much as logic to redux the less parameter to pass better.

app/assets/javascripts/actions/assignment_actions.js
shishiro26: Yuppp

Abishekcs:
@shishiro26, LGTM. I have tested this, and it works as expected. However, it still needs a final review from @ragesoss.
Also, please upload the 'before' screenshots.

ragesoss:
Hmm...

This set of changes touches more of the stack than I think it needs to. I suspect it could be implemented more simply if the handling of the flag and whether to delete or unclaim an Assignment happened entirely on the server. The `destroy` action of the controller could use the flag to determine whether to unclaim or destroy.

A possibly tricky bit with this approach would be to ensure that frontend state is correct afterwards (so that the available article is present again, if it was unclaimed rather than destroy). But I think that is better than bringing the "delete vs unclaim" logic all the way through to the frontend components.

ragesoss:
One additional complication that I had forgotten about is the "Retain Available Articles" setting, which makes an Available Article stay on the list and creates a new copy for the student when it is selected. This situation may already be handled correctly, but just in case: when that setting is in use for the course, choosing from the list should create a new Assignment that does not have the flag (because otherwise, unclaiming it would result in adding a second copy of the same Assignment to the Available Articles list).

shishiro26:
I think I can implement this in the delete request by adding a new parameter to specify whether the action is to "unclaim" or "delete." Based on this, the `destroy` method can include a check to decide whether to unclaim or delete the assignment.  

```ruby
  def destroy
    set_assignment { return }
    @course = @assignment.course
    check_permissions(@assignment.user_id)

    if assignment_params[:action_type] == 'unclaim'
      @assignment.update(user_id: nil)
    else
      remove_assignment_template
      @assignment.destroy
    end

    update_onwiki_course_and_assignments
    render json: { assignmentId: @assignment.id }
  end

```

Something like this should work to achieve the desired functionality.

  > One additional complication that I had forgotten about is the "Retain Available Articles" setting, which ensures an Available Article remains on the list and creates a new copy for the student when selected.
  
I haven’t tested the 'Retain Available Articles' functionality yet and am not familiar with it. Could you let me know where to find this feature?

ragesoss:
In Wiki Education mode, as an admin you can control that setting from the Details component of the course overview tab.

shishiro26:
- Admin side for deleting and unclaiming an assignment:  
https://github.com/user-attachments/assets/58889d3f-ba66-4258-8dc5-7996d2a74dbe

- User side for deleting an assignment:
https://github.com/user-attachments/assets/d9be1bf0-aec9-46dd-9d00-34fab332d923

ragesoss:
I would still like for this to be reworked to involve only backend changes, if possible, or at least as minimal frontend changes as possible.

shishiro26:
sir, could you please tell me  specific areas in the frontend where further optimization or minimization is possible?

ragesoss:
> sir, could you please tell me specific areas in the frontend where further optimization or minimization is possible?

This is what I was suggesting in my comment last week: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6060#issuecomment-2605542605

In particular, I'm suggesting that you attempt to implement the change by using only the current `destroy` action, so that there would be no need to make any changes to the frontend as part of this PR. Removing an assignment in the UI should not behave any differently for a retained assignment versus a non-retained one; in both cases, it would send the same request to the server.

shishiro26:
>so that there would be no need to make any changes to the frontend as part of this PR

Sir, in the frontend, the same delete API request is being used in multiple places where available articles are being removed. To differentiate between unclaim requests and delete requests, I added a new action_type. While the delete action_type isn't explicitly used, I included it to avoid potential errors from undefined values. However, the functionality works fine with or without the delete action_type. I was concerned that leaving it undefined might cause unexpected issues, so I decided to add it.

>Removing an assignment in the UI should not behave any differently for a retained assignment versus a non-retained one; in both cases, it would send the same request to the server.

Sir, I’m not sure what changes you are referring to, but I didn’t notice any UI changes when I remove either retained or non-retained articles. If a non-retained article is removed, only a notification is shown. If that’s the UI change you mean, I can remove it or add a notification for retained articles as well. For both retained and non-retained articles, the same request is sent to the server. The only difference is that for non-retained articles, an additional parameter (action_type="unclaim") is added in the url for the available articles.

ragesoss:
I don't think differentiating between an `unclaim` and a `delete` action type on the frontend is necessary. The Assignment record on the backend will already have all the information needed to determine whether to unclaim or delete the record.

shishiro26:
In this new approach, sir, I'm ensuring that the assignment_id and user_id are present, and confirming that the available_articles flags are set. This will help in determining whether to unclaim or delete the record based on the parameters information.

shishiro26:
@ragesoss is this okay or do I need to make any improvements

app/assets/javascripts/actions/assignment_actions.js
```diff
@@ -53,7 +53,19 @@ export const randomPeerAssignments = randomAssignments => (dispatch) => {
...
   return API.deleteAssignment(assignment)
-    .then(resp => dispatch({ type: types.DELETE_ASSIGNMENT, data: resp }))
+    .then((resp) => {
+      if (assignment.assignment_id && assignment.user_id !== null && assignment.flags.available_article) {
+        const successNotification = {
+          message: 'The assignment has been successfully moved to the available articles list.',
+          closable: true,
+          type: 'success'
+        };
+        dispatch(addNotification(successNotification));
```
ragesoss, line 63: I don't think this notification is necessary.

lib/assignment_manager.rb
```diff
@@ -113,6 +114,11 @@ def check_wiki_edu_discouraged_article
     end
   end
 
+  def set_flags
```
ragesoss, line 117: By convention, a method called `set_flags` should do set the value, rather than be used for its return value.

I suggest calling it `set_available_article_flag`, and have it set a value for `@flags`, rather than doing `flags = set_flags`.

shishiro26: Yuppp got it sir

shishiro26: Sir, are there any docs available outlining the specific conventions to be followed in the repo?

ragesoss: This particular thing is just a ruby convention, along with my personal taste.

shishiro26:
Just checking in to see if there are any blockers for merging this branch.

ragesoss:
This works smoothly! Nice work.

shishiro26:
@ragesoss Thank you for merging the PR!. 🙌
Would you be able to suggest some more issues that I could work on?

Issue #6057, opened 2024-12-10
On Programs & Events Dashboard user profile pages, links to courses should not use education-specific terminology


ragesoss:
The user profile page lists courses/programs/events that a user has participated in, eg: https://outreachdashboard.wmflabs.org/users/Ragesoss

Each link is labeled "View Course Page". That terminology is appropriate for Wiki Education Dashboard, but Programs & Events Dashboard should use a more generic label, perhaps simply "View Page".

sumana-2705:
hello @ragesoss 

I'm Angajala Sumana Sree pursuing my masters in IIT(BHU). I want to fix this word in dashboard, Can I work on this issue?

ragesoss:
@sumana-2705 go for it.

bhushan354:
Hello @ragesoss ,
I’ve been working on this issue and, since I have my local repo and database set up, I’ve pushed a fix to update the label. 
Thanks !

bhushan354:
@ragesoss ,
https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6061
Take a look !

sumana-2705:
Hello @ragesoss 

I guess @bhushan354  has opened a valid PR to solve the issue, I will find another one to contribute. Thank you

PR #6061, opened 2024-12-13
Generic label for Programs & Events Dashboard

bhushan354:

## What this PR does
This PR fixes : https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6057
Update link label from 'View Course Page' to 'View Page'

## Screenshots
Before:
image

After:
image

### DIFF
--- a/app/assets/javascripts/components/user_profiles/course_details.jsx
+++ b/app/assets/javascripts/components/user_profiles/course_details.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const CourseDetails = ({ courses }) => {
-  const labelKey = window.Features.wiki_ed ? 'courses.view_page' : 'courses_generic.view_page';
+  const labelKey = Features.wikiEd ? 'courses.view_page' : 'courses_generic.view_page';
   const elements = courses.map((course) => {
     return (
       <a className="course" key={`${course.course_slug}-${course.user_role}`} href={`/courses/${course.course_slug}`}>
--- a/app/assets/javascripts/components/user_profiles/course_details.jsx
+++ b/app/assets/javascripts/components/user_profiles/course_details.jsx
@@ -2,10 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const CourseDetails = ({ courses }) => {
+  const labelKey = window.Features.wiki_ed ? 'courses.view_page' : 'courses_generic.view_page';
   const elements = courses.map((course) => {
     return (
       <a className="course" key={`${course.course_slug}-${course.user_role}`} href={`/courses/${course.course_slug}`}>
-        <div className="button border">{I18n.t('courses.view_page')}</div>
+        <div className="button border">{I18n.t(labelKey)}</div>
         <div className="course-title">{course.course_title}</div>
         <div className="course-details">
           <div className="col">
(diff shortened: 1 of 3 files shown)


ragesoss:
It looks like this just changes it in all cases. Instead, it needs to switch behavior either based on Course type or based on `Features.wiki_ed?`. The locale actually already has an entry for the `courses_generic` version ('View Program Page'), so the necessary change is to use the correct one on Programs & Events Dashboard configuration without affecting the Wiki Education Dashboard configuration.

bhushan354:
@ragesoss Thanks for the feedback! I’ll update the PR to handle the behavior based on Features.wiki_ed? as you suggested.

Also, should I update `courses_generic : view_page: 'View Program Page'` to ` view_page: 'View Page'` for the Programs & Events Dashboard, or should it remain as is?

bhushan354:
Hi @ragesoss ,
image

This is the programme and event dashboard right ?

ragesoss:
@bhushan354 "Programs & Events Dashboard" is outreachdashboard.wmflabs.org, one of two production instances of the Dashboard. The other is Wiki Education Dashboard dashboard.wikiedu.org. The two sites have many behavior differences, mainly controlled by the `wiki_education` setting in `application.yml` (which in turn controls `Features.wiki_ed?`, which is used in the Ruby environment to decide which behavior to use, and `Features.wikiEd` in the JavaScript environment). So in a dev environment, you need to set that value to `false` to use Programs & Events Dashboard mode.

template.erb
```diff
@@ -1 +1,2 @@
 window.stores = <%= JSON.generate(translations) %>;
+window.Features = { wiki_ed: <%= Features.wiki_ed?.to_json %> };
```
ragesoss, line 2: This is normally handled in the shared `_head` haml template. I'm not sure what this .erb file is for but I don't think we use it in production. @TheTrio do you remember what this erb file is for? Something needed for Hot Module Reloading from https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5354 ? Or maybe it's just an accident and not used for anything?

app/assets/javascripts/components/user_profiles/course_details.jsx
```diff
@@ -2,10 +2,11 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 const CourseDetails = ({ courses }) => {
+  const labelKey = window.Features.wiki_ed ? 'courses.view_page' : 'courses_generic.view_page';
```
ragesoss, line 5: `window` shouldn't be necessary here, as we normally call `Features.wikiEd` anywhere in the JS that needs to switch behavior based on this setting.

ragesoss: Note that we have a helper function that is built to do this kind of switching: `CourseUtils.i18n`. It might be that the necessary data is not directly available, but you should check whether the course_string_prefix is part of the JS state already, which will let you use that utility function instead of implementing behavior switch here.

bhushan354: I have removed that. Thank you for helping me understand this!

bhushan354:
> @bhushan354 "Programs & Events Dashboard" is outreachdashboard.wmflabs.org, one of two production instances of the Dashboard. The other is Wiki Education Dashboard dashboard.wikiedu.org. The two sites have many behavior differences, mainly controlled by the `wiki_education` setting in `application.yml` (which in turn controls `Features.wiki_ed?`, which is used in the Ruby environment to decide which behavior to use, and `Features.wikiEd` in the JavaScript environment). So in a dev environment, you need to set that value to `false` to use Programs & Events Dashboard mode.

Since the application.yml file has already been set to false, I made the changes while recompiling the webpack. Please let me know if any further enhancements are needed. Thanks !

PR #6062, opened 2024-12-13
System Status and Performance Info UI

empty-codes:
## What this PR does
Documents progress in building a user-facing UI to present system status and performance information for the dashboard

Current Look:

https://github.com/user-attachments/assets/05c13562-20e3-41f5-af33-c5d9cfca2e1e

### DIFF
--- a/app/services/get_system_metrics.rb
+++ b/app/services/get_system_metrics.rb
@@ -66,7 +66,15 @@ class GetSystemMetrics
     remaining_seconds = (seconds % unit).to_i
     result = "#{main_unit} #{main_unit_name}#{'s' unless main_unit == 1}"
     if remaining_seconds.positive?
-      sub_unit_value = (remaining_seconds / (unit / 60)).to_i
+      sub_unit, sub_unit_name = case main_unit_name
+                                when 'day'
+                                  [3600, 'hour']
+                                when 'hour'
+                                  [60, 'minute']
+                                else
+                                  [1, 'second']
+                                end
+      sub_unit_value = (remaining_seconds / sub_unit).to_i
       result += " #{sub_unit_value} #{sub_unit_name}#{'s' unless sub_unit_value == 1}"
     end
     result
--- a/app/services/get_system_metrics.rb
+++ b/app/services/get_system_metrics.rb
@@ -2,51 +2,73 @@
 
 require 'sidekiq/api'
 
-class SystemMetricsService
+class GetSystemMetrics
   def initialize
     @queues = YAML.load_file('config/sidekiq.yml')[:queues]
+                  .reject { |queue_name| queue_name == 'very_long_update' }
     fetch_sidekiq_stats
   end
 
   def fetch_sidekiq_stats
     stats = Sidekiq::Stats.new
-
-    @processed_jobs_stat = stats.processed
-    @failed_jobs_stat = stats.failed # Failed Job Retention
-    @enqueued_jobs_stat = stats.enqueued
-    @scheduled_jobs_stat = stats.scheduled_size
-
-    # also allow stats from selected dates to be viewed
-    # using Stats::History
+    {
+      enqueued_jobs: stats.enqueued,
+      active_jobs: stats.processes_size
+    }
   end
 
-  # def fetch_system_health_metrics
-  # end
-
-  # def fetch_reliability_metrics
-  # end
-
-  # def fetch_availability_metrics
-  # end
-
-  # def fetch_processing_efficiency_metrics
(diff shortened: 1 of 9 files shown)


app/services/get_system_metrics.rb
```diff
@@ -0,0 +1,52 @@
...
+    @processed_jobs_stat = stats.processed
+    @failed_jobs_stat = stats.failed # Failed Job Retention
+    @enqueued_jobs_stat = stats.enqueued
+    @scheduled_jobs_stat = stats.scheduled_size
+
+    # also allow stats from selected dates to be viewed
+    # using Stats::History
+  end
+
+  # def fetch_system_health_metrics
```
ragesoss, line 23: My personal preference is to avoid leaving this kind of thing in the code, if it's not going to be part of a minimal initial implementation that we can merge and deploy. It think a first version that includes only sidekiq stats will already be useful, so we should probably plan on rolling that out before implementing any additional not-yet-defined metrics.

ragesoss:
You don't need to merge from `master` unless there is a conflict. If you want to make sure it's based on up-to-date code, you can rebase instead. With the merge, it makes for a very noisy set of changes that includes many things unrelated to the PR.

empty-codes:
> You don't need to merge from `master` unless there is a conflict. If you want to make sure it's based on up-to-date code, you can rebase instead. With the merge, it makes for a very noisy set of changes that includes many things unrelated to the PR.

Sorry about that, I've fixed it now

app/views/system_status/index.html.haml
```diff
@@ -0,0 +1,66 @@
+.container.queues
+  .module
+    .section-header
+      %h3 Queues Overview
```
ragesoss, line 4: Use i18n for all the text.

empty-codes: Done! Any other changes?

ragesoss: Not at the moment! Is this a complete MVP? If so, mark it ready for review and I can probably review it tomorrow. Also @Formasitchijoh please do a code review if you have a chance.

ragesoss:
This is still a work-in-progress, or should I review it for initial deployment? If it's not ready yet, what changes remain before it can be used?

empty-codes:
> This is still a work-in-progress, or should I review it for initial deployment? If it's not ready yet, what changes remain before it can be used?

I'd say it's ready. Please review it, I'll remove the WIP in the title

app/services/get_system_metrics.rb
```diff
@@ -0,0 +1,96 @@
+# frozen_string_literal: true
+
+require 'sidekiq/api'
+
+class GetSystemMetrics
```
ragesoss, line 5: I think `SystemMetrics` is a better name for this class than `GetSystemMetrics`.

app/services/get_system_metrics.rb
```diff
@@ -0,0 +1,96 @@
...
+  def initialize
+    @queues = YAML.load_file('config/sidekiq.yml')[:queues]
+                  .reject { |queue_name| queue_name == 'very_long_update' }
```
ragesoss, line 8: A code comment explaining the exclusion of this queue would be helpful.

app/services/get_system_metrics.rb
```diff
@@ -0,0 +1,96 @@
...
+    {
+      name: queue.name,
+      size: queue.size,
+      status: get_queue_status(queue.name, queue.latency),
+      latency: convert_latency(queue.latency)
+    }
+  end
+
+  def get_queue_status(queue_name, latency)
+    latency_thresholds = {
```
ragesoss, line 52: These should thresholds are constants, and should probably be implemented explicitly a constant object outside of the method.

ragesoss (commented):
This looks solid, overall. I think it'll be ready to merge after these minor changes.

app/services/system_metrics.rb
```diff
@@ -2,8 +2,18 @@
...
-class GetSystemMetrics
+class SystemMetrics
+  LATENCY_THRESHOLDS = {
+    'default' => 1,
+    'short_update' => 2.hours,
+    'medium_update' => 12.hours,
+    'long_update' => 1.day,
+    'daily_update' => 1.day,
+    'constant_update' => 15.minutes
+}.freeze
```
ragesoss, line 13: This is probably a linting violation, it should be indented the same as the constant name.

I usually also place a constant like this, that is used by one specific method, right above the method that uses it. That makes it a little easier to browse the code and see what the constant is for.

PR #6064, opened 2024-12-14
 Fix Broken Conditionals When Cloning Survey Question Groups

### LINKED ISSUE #2418: Cloning survey question groups results in broken conditionals
When a question group gets cloned, the conditionals that refer to other questions in the same question group get copied over unchanged, so the question they depend on isn't part of the new cloned question group.

We should either disable cloning of question groups, or make sure the conditions get updated appropriately during cloning and also add some validation for the conditionals within a question group to make sure they all refer to extant questions in that group.

Abishekcs:
closes #2418

## What this PR does

Fixes an issue where conditionals in cloned question groups pointed to non-existent questions from the original group by updating them to reference corresponding cloned questions and adding validations to ensure conditionals are valid.

## Cause of the Bug:

The root cause of the issue lies in the cloning process. When a question group is cloned:

- All questions within the group are duplicated, generating new records with new IDs.
- Conditionals are copied directly from the original questions without updating the referenced question IDs to match the new cloned group.
- As a result, conditionals in the cloned group refer to question IDs from the original group, causing broken or invalid behavior in `initConditionals() in SurveyAdmin.js` the line of code which lead to the error is mentioned below.

```
 return $row.find('select').val(`${question_id}`).trigger('change'); 
```

## Changes Implemented

**1.** **Conditional Mapping During Cloning:**

- When a question group is cloned, a mapping is created between the original questions and their cloned counterparts.
- Conditionals in the cloned questions are updated to refer to the corresponding cloned question IDs instead of the original IDs.

**2.** **Validation for Conditionals:**

- Added validation to ensure all conditionals within a question group point to questions that exist within the same group.
- This prevents broken or invalid conditionals from being saved.

**3.** **Fallback Handling:**

- If an original question referenced in a conditional cannot be matched to a cloned question, the conditional is cleared to prevent invalid references.

## Test Description

This test verifies that conditional question groups are correctly cloned, including their conditional logic.

#### Setup:
- Creates a new question group with a conditional follow-up question.
- Configures conditional logic to show the follow-up question only when a specific answer is selected for the first question.

#### Clone Operation:
- Clones the created question group using the "Clone" action.

#### Assertions:
- Ensures the cloned question group exists with the correct number of questions.
- Confirms the conditional logic in the cloned group correctly points to the corresponding cloned questions and values.

This ensures that the cloning functionality preserves all related conditional configurations.

## Screenshots
Before:

https://github.com/user-attachments/assets/0af8d059-22ad-4204-b0b0-cae15dfef81d

After:

https://github.com/user-attachments/assets/a18e45ab-a206-4698-b330-7bc2be7edace

- When a question group is deleted, the questions belonging to that group still exist in the database. Shouldn't these questions also be deleted ?

https://github.com/user-attachments/assets/6a369517-5c7e-49af-8eb1-e40bfc122d4a

- When a question used by a conditional question is deleted, the conditionals column of the conditional question is not updated. It still references the deleted question instead of being set to NULL. 

https://github.com/user-attachments/assets/10639cdb-aa24-41e6-bd83-82559124daae

### DIFF
--- a/app/assets/javascripts/surveys/modules/SurveyAdmin.js
+++ b/app/assets/javascripts/surveys/modules/SurveyAdmin.js
@@ -182,7 +182,7 @@ const SurveyAdmin = {
       credentials: 'include',
       headers: {
         'Content-Type': 'application/json',
-        'X-CSRF-Token': SurveyAdmin.$csrf_token.content
+        'X-CSRF-Token': SurveyAdmin.$csrf_token?.content
       },
     }).then(response => response.json())
       .then((data) => {
--- a/app/controllers/surveys_controller.rb
+++ b/app/controllers/surveys_controller.rb
@@ -119,9 +119,13 @@ class SurveysController < ApplicationController
   end
 
   def clone_question_group
-    clone = Rapidfire::QuestionGroup.find(params[:id]).deep_clone include: [:questions]
-    clone.name = "#{clone.name} (Copy)"
-    clone.save
+    # Fetch the original question group and its questions
+    @original_group = Rapidfire::QuestionGroup.includes(:questions).find(params[:id])
+    @clone_group = @original_group.deep_clone include: [:questions]
+    @clone_group.name = "#{@clone_group.name} (Copy)"
+    @clone_group.save
+
+    update_cloned_conditionals
     redirect_to rapidfire.question_groups_path
   end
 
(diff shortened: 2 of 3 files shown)


ragesoss:
Nice!

I think it would be good to extend the survey_admin_spec so that it includes this behavior (in a way that would break the test if we undid the controller changes).

I think both of the additional concerns you identify are valid, although I don't think they've caused any problems in practice.

Abishekcs:
> I think it would be good to extend the survey_admin_spec so that it includes this behavior (in a way that would break the test if we undid the controller changes).

Great! I will write the test code for it too.😄

> I think both of the additional concerns you identify are valid, although I don't think they've caused any problems in practice.

Okay.

Abishekcs:
@ragesoss,

I’ve extended the test for `survey_admin_spec`.

While writing the test, I noticed three "FIXME" tests that were commented out because they were failing. Interestingly, the test I was working on for the current PR was also failing for the same reason as at least one of those "FIXME" tests.

The issue was due to not using optional chaining (`?.`) when accessing the `content` property of a potentially non-existent CSRF token object. This caused a JavaScript error:
`Cannot read property 'content' of undefined.`

The error would occur when the token wasn’t immediately available during test execution, leading to failures.

The problematic code is in `SurveyAdmin.js`  (line 185):

```
getQuestion(id) {
  const that = this; // Storing the correct "this" reference for use inside the promise
  return fetch(`/surveys/question_group_question/${id}`, {
    method: 'GET',
    credentials: 'include',
    headers: {
      'Content-Type': 'application/json',
      'X-CSRF-Token': SurveyAdmin.$csrf_token.content // Issue: No optional chaining used here
    },
  })
  .then(response => response.json())
  .then((data) => {
    SurveyAdmin.handleConditionalQuestionSelect.call(that, data);
  });
}
```

The failure occurred because optional chaining (`?.`) was not used in `SurveyAdmin.$csrf_token.content`.

## Screen recording:


https://github.com/user-attachments/assets/8ec19f57-6189-430b-8b5a-d46207aaa756




## Capybara Screenshots (Just adding for reference in case required):
FirstQuestionGroup Edit Page
QuestionGroupPageAfterClone
CloneGroupQuestionEdit Page
Edit 2nd conditional question of the second cloned question group

Abishekcs:
Also, I have tested one of the three FIXME tests, and it is now passing. I have not tried the other two FIXME tests yet. If those also pass, I will open a separate PR for them.

PR #6065, opened 2024-12-17
Fix: failing test in account_requests_spec 

Formasitchijoh:
## What this PR does
This PR fixes the failing test in account_requests_spec by ensuring that the confirmation modal is visible and the "OK" button is rendered before interacting with it.

#### Key Changes:
- Added checks to ensure the modal is visible.
- Confirmed the "OK" button is available before the test interacts with it.

The initial failure could not be reproduced locally. However, a thorough analysis of the modal's implementation and the associated test suggests that the error is likely caused by the "OK" button not being available at the time the test is executed.

#### Files Changed:
-  account_request_spec.rb
## Screenshots
Before:(Failure on the CI log)
Screenshot 2024-12-17 at 12 13 15
After:
Screenshot 2024-12-17 at 12 12 44

### DIFF
--- a/spec/features/account_requests_spec.rb
+++ b/spec/features/account_requests_spec.rb
@@ -18,7 +18,10 @@ describe 'Account requests', type: :feature, js: true do
     login_as(instructor)
     visit "/courses/#{course.slug}"
     click_button 'Enable account requests'
-    click_button 'OK'
+    expect(page).to have_selector('.confirm-modal-overlay')
+    within('.confirm-modal') do
+      click_button 'OK'
+    end
     expect(page).to have_content('Account request generation enabled')
   end
 


Formasitchijoh:
> Interesting! I assumed this failure was because of the API call to check whether the username was available, which I guessed was failing sometimes because it was happening from GitHub servers affected by a rangeblock.
> 
> How confident are you that this fixes the issue? Was it failing before because the 'OK' button was disabled, or not appearing at all? If it wasn't appearing at all, I suspect that this won't fix it, because `click_button` is one of the Capybara actions that automatically includes waiting behavior.

This particular failure was with respect to requesting to  Enable an account, specifically this test was what  is failing
`  it 'can be enabled by the course facilitator' do
    login_as(instructor)
    click_button 'Enable account requests'
      click_button 'OK'
    expect(page).to have_content('Account request generation enabled')
  end`
  
  It test the Confirm Modal, that has the OK button,  this modal is rendered using Javascript,  So from the error, the button cannot be found, meaning the modal is not rendered, and the part that says, `a  button that is not disabled`, could be because there are other buttons on the page

ragesoss:
That makes sense as a possibility, but I don't know which other (disabled) OK button it would be. I'll merge this, and if it fails again at some point, we can revisit it.

Formasitchijoh:
> Interesting! I assumed this failure was because of the API call to check whether the username was available, which I guessed was failing sometimes because it was happening from GitHub servers affected by a rangeblock.
> 
> How confident are you that this fixes the issue? Was it failing before because the 'OK' button was disabled, or not appearing at all? If it wasn't appearing at all, I suspect that this won't fix it, because `click_button` is one of the Capybara actions that automatically includes waiting behavior.

I’m fairly confident this will address the issue, as the 'OK' button is not disabled. My assumption is that Capybara might have been interacting with the wrong button since the page contains multiple buttons. By explicitly ensuring the modal is visible and rendered before interacting, this fix should reduce ambiguity.

That said, I couldn’t reproduce the failure locally, so I’m addressing this somewhat blindly. If the issue persists, it might require further debugging to identify the exact conditions causing the test to fail.

Formasitchijoh:
> That makes sense as a possibility, but I don't know which other (disabled) OK button it would be. I'll merge this, and if it fails again at some point, we can revisit it.

Okay then

PR #6068, opened 2024-12-18
Implement pagination with 'continue' to load all linked pages

### LINKED ISSUE #5588: Fix 500-link limit for loading training slides from wiki

bhushan354:

## What this PR does
fixes https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5588

## Screenshots
Before:

After:

### DIFF
--- a/lib/training/wiki_training_loader.rb
+++ b/lib/training/wiki_training_loader.rb
@@ -93,15 +93,23 @@ class WikiTrainingLoader
 
   # Gets a list of page titles linked from the base page
   def wiki_source_pages
-    # To handle more than 500 pages linked from the source page,
-    # we'll need to update this to use 'continue'.
+    source_pages = []
     query_params = { prop: 'links', titles: @wiki_base_page, pllimit: 500 }
-    response = WikiApi.new(MetaWiki.new).query(query_params)
-    begin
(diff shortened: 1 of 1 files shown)


bhushan354:
Screenshot from 2024-12-18 14-08-58
12de0f4be5a)

bhushan354:
Screenshot from 2024-12-18 14-08-58

Screenshot from 2024-12-18 14-10-52

slug `editing-basics was` successfully deleted previously but they still again appear in console

bhushan354:
@ragesoss can you review this ?

ragesoss:
Thanks for the ping. It's on my todo list.

ragesoss:
You'll definitely want to implement a test for this. I think the API always includes a continue param of some sort, if not all the results in included in the response. The limit won't necessarily be 500; that depends on a mediawiki setting that might change, as well as the identify of the client making the request (as making a request while logged in with some accounts will allow for larger limits). But I think 500 is the current limit for a logged-0ut request (which is what the dashboard uses).

I think the slides index currently has more than 500 links, so you can write  test around that: https://meta.wikimedia.org/wiki/Training_modules/dashboard/slides

I suggest making a dedicated method to do the link fetching, so that you can test that in isolation, without needing to do anything with training records in the test.

PR #6072, opened 2024-12-20
fix(DiffViewer & ArticleViewer): handle moved articles using MediaWik…

### LINKED ISSUE #6056: DiffViewer breaks if article has been moved
The DiffViewer tool is available on the Articles Edited view of a course page (as well as a few other places).

If an article gets subsequently moved to a new title, and the Dashboard hasn't updated that article to reflect the move, then attempting to use the DiffViewer results in JS console error (and the DiffViewer doesn't render).

The DiffViewer should be updated to handle this situation. Ideally, it should be able to follow redirects and still show diffs for moved articles, but just showing a message like 'could not find article' would be okay (and probably necessary for the case where an article was deleted or moved without a redirect).

Abishekcs:
closes #6056

## What this PR does

This PR addresses an issue where the `DiffViewer` and `ArticleViewer` components did not handle moved articles correctly, resulting in `errors` for `DiffViewer`  and `Could not fetch authorship data: Requested page not found. [404]` for `ArticleViewer`.

## Fix
Check the current article title by fetching it from the Wikipedia API using the article's page ID if first and last revision are undefined and update the title and URL for the respective article.

## Screenshots
Before:

https://github.com/user-attachments/assets/1b2e1c42-9af7-4b89-9e6b-431a378c1c0e

After:

https://github.com/user-attachments/assets/65390cf7-3604-46fa-8e66-c62637138aee


Abishekcs:
I don't know why the build is failing for the test article_viewer_spec.rb. It’s passing for me locally.😅 


https://github.com/user-attachments/assets/9ee72c61-e68c-481a-aaef-12ccf81da13c

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -14,7 +14,7 @@ const Article = ({ article, index, course, fetchArticleDetails, updateArticleTra
 
   const fetchMissingArticleDetails = () => {
     if (!articleDetails) {
-      fetchArticleDetails(article.id, course.id);
+      fetchArticleDetails(article.id, course.id, article.title, article.mw_page_id);
```
ragesoss, line 17: I think this gets called from several places, including Assignment.jsx which you haven't changed. Did you check them all to verify that this doesn't break anything?

app/views/courses/articles.json.jbuilder
```diff
@@ -12,5 +12,6 @@ json.course do
     json.rating_num rating_priority(article.rating)
     json.pretty_rating rating_display(article.rating)
     json.user_ids ac.user_ids
+    json.mw_page_id article.mw_page_id
```
ragesoss, line 15: This could probably be done within the `json.call(article ...)` instead of on its own line? I guess `url` and `user_ids` could be consolidated as well?

app/assets/javascripts/actions/article_actions.js
```diff
@@ -3,19 +3,20 @@ import API from '../utils/api.js';
...
+// action object, it returns a function that takes the store dispatch function —
 // which Thunk automatically provides — and can then dispatch a series of plain
 // actions to be handled by the store.
 // This is how actions with side effects — such as API calls — are handled in
 // Redux.
-export function fetchArticleDetails(articleId, courseId) {
-  return function (dispatch) {
+export function fetchArticleDetails(articleId, courseId, articleTitle, article_mw_page_id) {
+  return async function (dispatch) {
+    const crossCheckedArticleTitle = await dispatch(crossCheckArticleTitle(articleId, articleTitle, article_mw_page_id));
```
ragesoss, line 13: Hmm... this will add an extra network call to every time we use fetchArticleDetails, even though in most cases the page has not moved.

Is it possible to implement this strategy only for cases where the query for article data doesn't work?

Abishekcs: > Is it possible to implement this strategy only for cases where the query for article data doesn't work?

Yes, I made the changes to  call `crossCheckedArticle` only if the last and first revisions are undefined, indicating that the article title has been moved.

app/views/courses/articles.json.jbuilder
Abishekcs: Okay.

app/assets/javascripts/components/articles/article.jsx
Abishekcs: > Did you check everything to ensure this doesn't break anything?

No 😅, I forgot to check.
Now, I have updated fetchArticleDetails to fetch the article title and the MW page ID directly from the Redux store using the existing article ID via getState. This means other places using fetchArticleDetails no longer need to pass the title and page ID.

Abishekcs:
https://github.com/user-attachments/assets/565f0db3-d5f0-4ee7-9429-70caf228a0b0

PR #6075, opened 2024-12-20
[FIX]: Database Connection Issue by Hostname Resolution During Setup

### LINKED ISSUE #6074: Database Connection Error: During Setup using Docker
### Current Behavior:
During the setup process using Docker, an issue occurs while populating the database. It results in a database connection error, leading to the failure of the setup process.

**Screenshots-**

image

Steps to Reproduce:

- Clone the repository and follow the setup instructions on a Fedora system.
- Attempt to run the Docker Compose setup.
- Try populating the database with the command `rake dev:populate`
- Observe the database connection error

### Desired Behavior:
It should have successfully populated the database

### Environment:
**Operating System: Fedora**

JiyaGupta-cs:
## What this PR does
This pull request addresses an issue encountered during the setup process on Fedora using Docker, where the database connection fails due to hostname resolution issues. The error occurs when the system is unable to resolve the required hostname, leading to the failure of the setup process.

Fixes #6074

### Changes Introduced:

Added a script that automatically resolves the hostname by ensuring proper network mapping, allowing the database connection to be established successfully.
This script ensures that the system can correctly resolve the hostname and connect to the database without errors.
Can be invoked by ```bash ./update_hosts.sh```

### Explanation:
The issue arises because the necessary host entries are not present in the ```/etc/hosts``` file
When the system attempts to resolve an address, it fails to locate the corresponding hostname in the local resolution file and subsequently defaults to querying the network gateway to determine if the address exists.
To address this, I have implemented a script to automate the addition of host entries to the ```/etc/hosts file```. This ensures that the hostname resolution process can efficiently locate and map the desired addresses locally without needing to fallback to external network queries.
The current Docker Compose file lacks the necessary configurations to locate the mappings effectively.

### Key Cases Handled:

- **Container Running:**  
  If the Docker container is up, the script updates the `/etc/hosts` file with the new IP address, ensuring correct hostname resolution.

- **Container Restarted:**  
  If the Docker container is restarted and the IP address changes, the script checks for the old address and updates it with the new IP address in `/etc/hosts`.

- **Container Down:**  
  If the Docker container is down, the `update_hosts.sh` script removes the corresponding address from the `/etc/hosts` file to prevent stale entries.

- **No Duplicate Entries:**  
  The script ensures that there are no duplicate IP addresses for the same hostname in the `/etc/hosts` file, maintaining clean mappings.

### Testing:

Verified that running the script ensures successful hostname resolution.
Confirmed that the database connection is now established properly, and the setup process completes without errors.

### Remark
This PR solves the database connection issue for Fedora

### Other Questions:
This also needs updation in the Docker docs
So should I make that also ?

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -177,7 +177,7 @@ class CoursesController < ApplicationController
   def manual_update
     require_super_admin_permissions
     @course = find_course_by_slug(params[:id])
-    UpdateCourseStats.new(@course, full: true)
+    UpdateCourseStats.new(@course)
     redirect_to "/courses/#{@course.slug}"
   end
 


JiyaGupta-cs:
Video Demonstration of using with and without script

Uploading VID_20241221_034417_795.mp4…

JiyaGupta-cs:
Video Demonstration of setting up with and without script 

https://github.com/user-attachments/assets/41e46325-87b7-4302-b408-a93beb65e3ce

JiyaGupta-cs:
> Yes, please update the docker.md file along with this. I don't use Docker for my dev environment, but this looks like a reasonable script.

Sure, I will update that also.

JiyaGupta-cs:
I closed this by mistake
Will open a new one for the same issue @ragesoss

PR #6080, opened 2024-12-29
Add student program endpoint

Aminehassou:
## What this PR does
This PR adds new JSON endpoints for the dashboard. These endpoints are needed to add new functionality to the impact-visualizer project.

**The added endpoints:**
`/courses/classroom_program_students.json` (returns all students and the courses they belong to from the Wikipedia Student Program)
`/courses/classroom_program_students_and_instructors.json`  (returns all students, instructors and the courses they belong to from the Wikipedia Student Program)
`/courses/fellows_cohort_students.json` (returns all students and the courses they belong to from the Scholars and Scientists Program)
`/courses/fellows_cohort_students_and_instructors.json`  (returns all students, instructors and the courses they belong to from the Scholars and Scientists Program)

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -139,10 +139,28 @@ class CoursesController < ApplicationController
     @alerts = current_user&.admin? ? @course.alerts : @course.public_alerts
   end
 
-  def approved_classroom_courses_json
-    courses = Course.approved_classroom_courses_with_users
+  def classroom_program_students_json
+    courses = Course.classroom_program_students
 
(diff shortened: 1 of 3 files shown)


ragesoss:
Can you include examples of what the JSON output for each of these new endpoints looks like?

Aminehassou:
> Can you include examples of what the JSON output for each of these new endpoints looks like?

I made a quick change to filter out some unnecessary data, this is what the json data looks like now:
image

the returned data structure for the "and_instructors" endpoints are the same except for an added "instructors" field that has the same user fields as the "students" field

PR #6081, opened 2024-12-29
@fix/mock flaky test

Formasitchijoh:
## What this PR does
This PR addresses flaky tests that are caused by interactions with external api's like LiftWing and the Reference Counter API 
#### Changes Made:
- Mocked API responses for LiftWingApi, referenceCounterApi, and revision_score_api by stubbing the request using WebMock and returning the expected data this helps to resolve flaky test failures caused by external dependencies.
- Explicitly verified actual error counts in the `update_course_stats` test to prevent reliance on static values and ensure accurate error reporting 
- Prevent retries during testing in the LiftWing class by using a RETRY-COUNT constant that is stub in the test using `stub_const` to prevent multiple retries in test mode

These changes ensure more stable and reliable test execution by isolating tests from external API calls and improving the validation logic.


lib/lift_wing_api.rb
```diff
@@ -60,7 +60,7 @@ def get_revision_data(rev_ids)
   # Returns a hash with wp10, features, deleted, and prediction, or empty hash if
   # there is an error.
   def get_single_revision_parsed_data(rev_id)
-    tries ||= 5
+    tries = Rails.env.test? ? 1 : 5
```
ragesoss, line 63: I strongly prefer not to have environment-based switching in normal code like this.

One alternative strategy is to make the initial `tries` value come from a constant, and mock that constant within the tests as needed.

Also, I think this will break the intended behavior unless we keep the `||=` operator. Otherwise, it will retry endlessly, as `tries` gets updated to the initial value on every retry.

spec/features/syllabus_upload_spec.rb
```diff
@@ -28,7 +28,8 @@
       visit "/courses/#{course.slug}?syllabus_upload=true"
       expect(page).to have_content course.title
       expect(page).to have_content 'Syllabus'
-      click_button 'edit'
+      expect(page).to have_selector('.module.course-description.course__syllabus-upload__inner')
+      click_button 'edit' if page.has_button?('edit')
```
ragesoss, line 32: Can you link the CI log from where this spec failed? It's not clear to me why this change will help make it more reliable.

Formasitchijoh: Here is the build
https://github.com/WikiEducationFoundation/WikiEduDashboard/actions/runs/11393351404/job/31969011021

spec/support/request_helpers.rb
```diff
@@ -561,4 +561,64 @@ def stub_course
     stub_timeline
     stub_users
   end
+
+  def stub_lift_wing_response
```
ragesoss, line 565: The idea with all the stub methods in `request_helpers` is to use `stub_request` to replace real web requests to specific domains or URLs, not to directly return data.

Formasitchijoh: I understood what the stubbing required,  but the data returned from these api calls are large, so i did not want to fill the entire file with those data,

ragesoss: I smaller response is fine, but anything in this file should still rely on `stub_request` to do its work.

Formasitchijoh: Alright, I'll update that and push

spec/features/syllabus_upload_spec.rb
ragesoss: That doesn't provide any clues as to why the edit button didn't appear, and I expect this to fail in the following `find` line instead, if the same situation happens after this change.

I suggest removing this change from this PR, as I don't think we have any good clues about what is going wrong.

This is also a very low-priority test (as we basically never use this feature), so marking it pending would be an acceptable fallback.

spec/lib/revision_score_api_handler_spec.rb
```diff
@@ -119,21 +133,23 @@
   end
 
   context 'when the wiki is available only for reference-counter API' do
-    before { stub_wiki_validation }
+    before do
+      stub_wiki_validation
+      allow_any_instance_of(Wiki).to receive(:ensure_wiki_exists).and_return(true)
```
ragesoss, line 138: This should not be necessary, as stub_wiki_validation ought to take take of it unless there is a wiki being created in the test that isn't included amongst the ones stubbed by that method.

Formasitchijoh: No there isn't,  I will remove this ` allow_any_instance_of(Wiki).to receive(:ensure_wiki_exists).and_return(true)` and test with the `stub_wiki_validation` and ensure it works as expected

lib/lift_wing_api.rb
```diff
@@ -20,6 +20,8 @@ class LiftWingApi
   # All the wikis with an articlequality model as of 2023-06-28
   # https://wikitech.wikimedia.org/wiki/Machine_Learning/LiftWingq
   AVAILABLE_WIKIPEDIAS = %w[en eu fa fr gl nl pt ru sv tr uk].freeze
+  # config/initializers/retry_config.rb
+  RETRY_COUNT = Rails.env.test? ? 1 : 5
```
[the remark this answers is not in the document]
Formasitchijoh: Alright, I will fix that

spec/features/syllabus_upload_spec.rb
ragesoss: I suggest removing the changes to syllabus_upload_spec, as I'm not convinced they do anything and they are independent of the other changes in this PR (which look good to me).

ragesoss:
This looks like a good set of changes. Please clean up the git commit history and update the description to reflect only the changes that are part of the current version.

Formasitchijoh:
> This looks like a good set of changes. Please clean up the git commit history and update the description to reflect only the changes that are part of the current version.

I will update that, I just have a blocker which I have not been able to fix, with respect to the multiwiki_assignment_spec failure , I wish to ask if i could put as a Pending test temporarily while I try to investigate the cause or refactor the test

ragesoss:
Yes, you can do make the multiwiki_assignment_spec Pending, but please do that in a separate PR.

Formasitchijoh:
@ragesoss I have resolved the changes you requested here and cleaned the commit history

ragesoss:
The git history still has the syllabus revert and such. My usually strategy for case like this would be to take all the changes, move to a fresh branch, and then make the changes in one or more new commits, with each commit making sense (and leaving the code in working condition) on its own.

Formasitchijoh:
> The git history still has the syllabus revert and such. My usually strategy for case like this would be to take all the changes, move to a fresh branch, and then make the changes in one or more new commits, with each commit making sense (and leaving the code in working condition) on its own.

Alright I will do just that, I was trying to  keep the reviews to requested so, you can see it's been done

PR #4381, opened 2021-03-12
Add a unique index on mwpageid and wikiid for non-deleted rows

nateberkopec:
```
[4] pry(main)> Article.first
[2021-03-12 15:03:57.667 DEBUG] Article Load (0.4ms)  SELECT `articles`.* FROM `articles` ORDER BY `articles`.`id` ASC LIMIT 1
=> #<Article:0x00007fd3f775d258
 id: 2,
 title: "Bar/Discussioni/ArtAndFeminism_2018_a_Firenze",
 created_at: Fri, 12 Mar 2021 21:51:10 UTC +00:00,
 updated_at: Fri, 12 Mar 2021 21:51:10 UTC +00:00,
 views_updated_at: nil,
 namespace: 4,
 rating: nil,
 rating_updated_at: nil,
 deleted: false,
 language: nil,
 average_views: nil,
 average_views_updated_at: nil,
 wiki_id: 1,
 mw_page_id: 6887755,
 index_hash: "68877551">
[5] pry(main)> Article.first.update(deleted:true)
[2021-03-12 15:04:08.987 DEBUG] Article Load (0.5ms)  SELECT `articles`.* FROM `articles` ORDER BY `articles`.`id` ASC LIMIT 1
[2021-03-12 15:04:08.991 DEBUG]  (0.2ms)  BEGIN
[2021-03-12 15:04:08.991 DEBUG] Article Update (0.3ms)  UPDATE `articles` SET `articles`.`deleted` = TRUE, `articles`.`updated_at` = '2021-03-12 22:04:08' WHERE `articles`.`id` = 2
[2021-03-12 15:04:08.992 DEBUG]  (0.6ms)  COMMIT
=> true
[6] pry(main)> Article.first
[2021-03-12 15:04:11.336 DEBUG] Article Load (0.3ms)  SELECT `articles`.* FROM `articles` ORDER BY `articles`.`id` ASC LIMIT 1
=> #<Article:0x00007fd3ef71c350
 id: 2,
 title: "Bar/Discussioni/ArtAndFeminism_2018_a_Firenze",
 created_at: Fri, 12 Mar 2021 21:51:10 UTC +00:00,
 updated_at: Fri, 12 Mar 2021 22:04:08 UTC +00:00,
 views_updated_at: nil,
 namespace: 4,
 rating: nil,
 rating_updated_at: nil,
 deleted: true,
 language: nil,
 average_views: nil,
 average_views_updated_at: nil,
 wiki_id: 1,
[...]

PR #6083, opened 2025-01-02
[Data rearchitecture] Implement article status manager for timeslices

gabina:
## What this PR does
This PR implements a new `ArticleStatusManagerTimeslice` class that "imitates" the process previously done by `ArticleStatusManager` class.

### Analysis of the ArticleStatusManager class (revisions verison)
The article status manager has several steps. Conceptually, `ArticleStatusManager` modifies articles and revisions. Both entities are course agnostic, i.e., they are not associated with a particular course. This is a fundamental difference with the new system because although the entity article was not modified, the revisions entity no longer exists. Instead, 
we have articles course timeslices, which are entities associated to a specific course. This is something to pay special attention to, because if several courses share the same article, article course timeslices should be updated for all courses that use that article. 

### New implementation
Due to the fundamental differences between the systems, it is not possible to make a direct translation from one implementation to the other. It is necessary to redesign the behavior of the class thinking about the new properties that the timeslices system presents.

As part of this PR, we modified the logic for article course timeslice creation. Before this PR, timeslices were created only for articles courses (articles edited by the course users in the tracked namespaces). Now, timeslices for all articles with at least one revision are created, but we only create article course records for those that are pertinent to the course. This is to have data for articles that, for example, are created in draft but then move to the main namespace.

The new `ArticleStatusManagerTimeslice` class does the following:

1. Update articles that moves or gets deleted/undeleted. It basically iterates over article course timeslices to sync articles. This updates title, `namespace`, `deleted` and `mw_page_id` fields.
2. Reset articles according to their new status:
    - Articles that were deleted or untracked. These are articles that were either deleted or moved to a namespace not traceable by the course. Such articles should be excluded from course statistics.
   - Articles that were restored or re-tracked: These are articles that were either undeleted or moved to a namespace relevant to the course. Such articles should be included in course statistics.

It looks like there are two main use cases for the `ArticleStatusManager` class. The usual one is through the periodic course update. The other one is through the cleanup scripts: `docs/cleanup_scripts/duplicate_articles.rb` and `docs/cleanup_scripts/duplicate_mw_page_id_handling.rb`. For the latter case, `update_status` with a single article is invoked, and it's kind of a special case. I'm not entirely sure if those cleanup scripts are still used, as it seems that we added a db restriction in the Articles table (see #4381). If the scripts are no longer used then we can simplify the new implementation.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -332,6 +332,13 @@ class Course < ApplicationRecord
     categories.inject([]) { |ids, cat| ids + cat.article_ids }
   end
 
+  def articles_from_timeslices(wiki_id)
+    Article.joins(:article_course_timeslices)
+           .where(article_course_timeslices: { course_id: id })
+           .where(wiki_id:)
+           .distinct
+  end
+
   def update_until
     self.end + UPDATE_LENGTH
   end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -332,6 +332,9 @@ class Course < ApplicationRecord
     categories.inject([]) { |ids, cat| ids + cat.article_ids }
   end
 
+  # Retrieve articles based on the existing article course timeslices.
+  # This includes both tracked and untracked articles, such as those that
+  # don't belong to a tracked namespace.
   def articles_from_timeslices(wiki_id)
     Article.joins(:article_course_timeslices)
            .where(article_course_timeslices: { course_id: id })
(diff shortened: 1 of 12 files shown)


gabina:
> Hello @gabina the failures in your build has been fixed, and merged into the master branch, Could you update your branch with the latest changes.

Thank you for letting me know. I'll sync with master in a different PR to avoid making such a large PR.

PR #6086, opened 2025-01-03
Fix: failure in Student_role_spec

Formasitchijoh:
## What this PR does
This PR resolves the recent failure in the  multiwiki_assignment_spec and student_role_spec failure by fixing useNavigationUtils access and updating course creation date
## Screenshots
Before:
Screenshot 2025-01-03 at 17 06 32

Screenshot 2025-01-03 at 17 05 28

Screenshot 2025-01-03 at 17 06 53

### DIFF
--- a/spec/features/student_role_spec.rb
+++ b/spec/features/student_role_spec.rb
@@ -15,8 +15,8 @@ describe 'Student users', type: :feature, js: true do
            slug: 'University/An_Example_Course_(Term)',
            submitted: true,
            passcode: 'passcode',
-           start: '2015-01-01'.to_date,
-           end: '2025-01-01'.to_date)
+           start: '2025-01-01'.to_date,
+           end: 1.month.from_now)
   end
   let!(:editathon) do
     create(:editathon,
@@ -26,8 +26,8 @@ describe 'Student users', type: :feature, js: true do
            slug: 'University/An_Example_Editathon_(Term)',
            submitted: true,
            passcode: '',
-           start: '2015-01-01'.to_date,
-           end: '2025-01-01'.to_date)
+           start: '2025-01-01'.to_date,
+           end: 1.month.from_now)
   end
 
   before do


ragesoss:
I see that exported method from useNavigationUtils has a different name than the file itself — this was like an accident on the part of @Abishekcs — but I don't understand why that should matter, as all files that import it use the corresponding method and file names. This spec isn't failing every time.

spec/features/student_role_spec.rb
```diff
@@ -15,8 +15,8 @@
            slug: 'University/An_Example_Course_(Term)',
            submitted: true,
            passcode: 'passcode',
-           start: '2015-01-01'.to_date,
-           end: '2025-01-01'.to_date)
+           start: '2025-01-01'.to_date,
+           end: '2025-03-01'.to_date)
```
ragesoss, line 19: A course cannot be joined after it has ended, so I think just the end date is relevant here. This appears to be unrelated to the multiwiki assignment spec; I imagine this just started failing at the beginning of the year.

A better fix will be to only change the end date, and make it dynamic (like `1.month.from_now`) so that it will never break again because of the same issue.

This is best done in a separate PR, since it's not about the multiwiki bit.

Formasitchijoh:
> I see that exported method from useNavigationUtils has a different name than the file itself — this was like an accident on the part of @Abishekcs — but I don't understand why that should matter, as all files that import it use the corresponding method and file names. This spec isn't failing every time.

Recently, for the last five or more builds, the test suite has been failing, including the multiwiki_assignment_spec, with no test passing even locally.

The primary issue was an unresolved access to the useNavigationUtils method, which caused the Webpack bundle to throw an error, preventing the tests from progressing.

Additionally, the course's start and end dates were problematic. The enrollment date for the course had already passed, making the "Join Course" button and other related buttons or links inaccessible to students.

spec/features/student_role_spec.rb
Formasitchijoh: The multiwiki_assignement_spec passed when i merge from  the master branch, I will just have to validate if this build passes

Formasitchijoh: I decided to fix both the failures in this PR because they both have been occurring along side for the recent failed builds. do you think i should separate them

ragesoss:
I haven't seen any of the cases where unresolved access to useNavigationUtils was a problem on CI builds. If this was happening locally, you might have had a problem with the local assets, which probably got fixed when you rebuilt the assets during this set of changes.

spec/features/student_role_spec.rb
ragesoss: Yes, I think a separate PR just for the student_role_spec is appropriate. We have a clear understanding of this new failure and the fix should be easy.

Formasitchijoh:
> I haven't seen any of the cases where unresolved access to useNavigationUtils was a problem on CI builds. If this was happening locally, you might have had a problem with the local assets, which probably got fixed when you rebuilt the assets during this set of changes.

Alright, but the new failure is with respect to some Javascript vega-embed.min.js file, currently checking what that is for

ragesoss:
That's a file that gets served externally (see _vega.html.haml), so either a CDN failure (for Wiki Education mode) or a problem with toolforge.org (for Programs & Events mode) would make that file fail to load.

Formasitchijoh:
It seems to be with the toolforge   http://wikiedudashboard.toolforge.org/vega-embed.min.js
Screenshot 2025-01-03 at 18 16 38

ragesoss:
Is that usually the same error we see with the account requests spec? I haven't noticed it before, but I haven't paid attention to it.

Formasitchijoh:
In the Github Worflow , it is possible to be able to trigger a build to rerun apart from making a PR, but i haven't been able to see that when i check the failed build in the Actions tab

Formasitchijoh:
> Is that usually the same error we see with the account requests spec? I haven't noticed it before, but I haven't paid attention to it.

No this isn't the error with the account request spec was with respect to ElementNotFound for the Ok Button

ragesoss:
Okay. I suspect this was a temporary disruption of toolforge, and is probably not worth focusing on currently.

ragesoss:
> In the Github Worflow , it is possible to be able to trigger a build to rerun apart from making a PR, but i haven't been able to see that when i check the failed build in the Actions tab

It probably requires more permissions than you have for this repo. Ping me on slack if you want me to trigger a re-run of any build.

Formasitchijoh:
Alright I will try, rerunning again

Formasitchijoh:
> > In the Github Worflow , it is possible to be able to trigger a build to rerun apart from making a PR, but i haven't been able to see that when i check the failed build in the Actions tab
> 
> It probably requires more permissions than you have for this repo. Ping me on slack if you want me to trigger a re-run of any build.

Okay then thank you

Formasitchijoh:
@ragesoss  Please could you check this out when you are chanced, so everyone could have the updated changes

ragesoss:
This still contains commits from two unrelated things. I'd like a separate PR that just includes the student_role_spec change. I don't think the useNavigationUtils changes do anything.

Formasitchijoh:
Alright since this PR contains just the changes for the useNavigation i will revert it and keep the student_role_spec and update the PR name

ragesoss:
Sounds good. Please clean up the commit history as well; this just needs one commit.

PR #6088, opened 2025-01-04
Fix: multiwiki_assignment_spec failure

Formasitchijoh:
## What this PR does
This PR addresses the recent failure in the multiwiki_assignment_spec where Capybara was unable to find the button "Assign all" because it possible was not available. The issue is resolved by utilizing Capybara's default wait mechanism, ensuring that the button is present and enabled before interacting with it.

#### Notes:
The other errors that occurred along with this issue, such as the missing content ("Heyder Cansa" and "Wp/kiu/Hey"), have not occurred recently.
This improvement is due to the increased Capybara wait time, which accounted for the longer period required for the element to become available, thus preventing interaction failures.

Screenshot 2025-01-04 at 05 21 09

### DIFF
--- a/spec/features/multiwiki_assignment_spec.rb
+++ b/spec/features/multiwiki_assignment_spec.rb
@@ -60,8 +60,9 @@ describe 'multiwiki assignments', type: :feature, js: true do
         first('textarea').set(
           "Terre\nhttps://fr.wikipedia.org/wiki/Anglais"
         )
+        expect(page).to have_button('Assign all')
+        click_button 'Assign all'
       end
-      click_button 'Assign all'
 
       visit "/courses/#{course.slug}/students/articles"
       first('.student-selection .student').click


Formasitchijoh:
This failure is due to the issue in the student_role_spec which i have fixed [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6086), after merging that i will rebuild this and it should work

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -60,8 +60,9 @@
         first('textarea').set(
           "Terre\nhttps://fr.wikipedia.org/wiki/Anglais"
         )
+        expect(page).to have_button('Assign all')
+        click_button 'Assign all'
```
ragesoss, line 64: I don't think this will make a difference, as click_button already incorporates Capybara 'wait' behavior. If the button renders in time, it will work either way, and if it takes more than the wait time to render, it will fail at the `expect(page)...`

Formasitchijoh: Your right, I thought ensuring the button is  found within the  #user  element, 
when i tested locally running Capybara in headless mode i realized the Assign all button appeared and close very fast

PR #6092, opened 2025-01-05
Sync troubleshooting.md with setup.md

chrisnicollo:
Currently, the [prerequisites section](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/setup.md#prerequisite) of `setup.md` says to use Node version 14 or 16. However, the [common setup issues section](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/troubleshooting.md#other-common-setup-issues) of `troubleshooting.md` says to use Node 10 or below. After having run into issues with using Node 10, it seemed to me that that information was out of date.

In this PR, I change `troubleshooting.md` to state the same Node versions as `setup.md`.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -24,7 +24,7 @@ For Windows, the easiest way to get an environment set up is to [use Ubuntu with
 ## Prerequisite
 There are some basic requirements for the script to work:
 - git (to clone the repository)
-- node version 14 or 16
+- node version 14 or newer
 - python 3.5 or newer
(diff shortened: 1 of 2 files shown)


ragesoss:
I think setup.md is out of date as well. We are using version 14 for the tests (configured in `.github/workflows/ci.yml`) but I use node 20 locally, and the current LTS version is 22. As far as I know, any version that is 14 or higher should work.

PR #6093, opened 2025-01-05
Fix Rubocop Style/GuardClause Violations

chrisnicollo:
As stated in issue #5297, there are currently too many global exemptions to rubocop rules. One of them is Style/GuardClause.

In this PR, I removed that exemption and replaced violations with guard clauses where necessary.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -123,8 +123,6 @@ Rails/FilePath:
   Enabled: false
 Performance/UnfreezeString:
   Enabled: false
-Style/GuardClause:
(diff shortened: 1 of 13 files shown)


ragesoss:
These look like nice style improvements to me.

In principle I don't like adding exceptions just for the sake of avoiding the violation of another rule, and I think there might be some refactoring that could make that code a little more readable and satisfy both rules together. But that might be more trouble than it's worth, and I'm fine with leaving the inline exceptions in place at this point.

PR #6103, opened 2025-01-08
Fix Conditional Question Display State Issues in Surveys

### LINKED ISSUE #2190: Fix handling of conditional survey questions when condition changes
In the survey system, there is an option for conditional questions that are only shown if previous questions are answered in a specific way (which is used for follow-up questions that only apply to some answers).

Previously, there was a bug (#1263) such that answering in that specific way, then proceeding to the conditional followup question, then backing up and changing the answer to the earlier question to one that does not trigger the followup, would put the survey into an unusable state with no questions or controls visible.

Now (pending the merge of https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/9f374a937e7ea693d3939b5c31d8957384f5d0e6), that function has been disable, such that once a conditional followup question is enabled, it won't be disabled even if the previous answer changes. Instead, it should remove the followup question as intended, but without breaking the survey flow.

The code that controls survey navigation is in `app/assets/javascripts/surveys/modules/Survey.js`, and is mostly implemented with jQuery and the slick-carousel library. The commented-out `resetConditionalGroupChildren` function is the one that handles the situation where a conditional followup should be removed. The spec `spec/features/survey_bugs_spec.rb` tests the current behavior, and could be modified slightly to test the intended behavior.

Abishekcs:
closed #2190

## What this PR does

This PR fixes an issue where conditional questions in a survey were being displayed incorrectly. Now, conditional questions are shown only if the preceding question is answered with an option that triggers the conditional question. If the triggering option is not selected, the conditional question will not be displayed.

## Cause of the Bug

When the answer to the trigger question was changed, the `resetConditionalQuestion function` [detached](https://api.jquery.com/detach/) the conditional question's DOM element but did not remove it from the `Slick slide` managed by the [Slick library](https://kenwheeler.github.io/slick/). This caused Slick to display a detached DOM element, leading to inconsistent behavior.

**Note (Not Related to the Bug)**
The DOM for the conditional question was being added to the `Slick slide` using the `activateConditionalQuestion function.

` $slider.slick('slickAdd', $question, parentIndex);`

## Fix

The fix ensures that when the answer to a trigger question changes, the conditional follow-up question is properly removed. This is done by updating the Slick slider's internal state before performing DOM manipulations (e.g., using [jQuery .detach()](https://api.jquery.com/detach/)). This restores the intended flow of the survey.

## Test

- Removed few lines of test code in survey_bugs_spec.rb  that was no longer required due to the current fix.
- Verified that the conditional question is skipped correctly without breaking the survey flow.

## Screenshot:

Before:

https://github.com/user-attachments/assets/f84bc42b-c48d-443a-8ed7-595a8c304dd8

After:

https://github.com/user-attachments/assets/0dfb5040-9ff1-4d3b-962c-0cb0267d3501

Additionally, when attempting to preview any of the question groups in the "Question Groups" tab, an empty page appears upon clicking the Preview Question Group button, as shown below.

https://github.com/user-attachments/assets/903b644c-b907-4978-b7f9-4b278f6787db

### DIFF
--- a/app/assets/javascripts/surveys/modules/Survey.js
+++ b/app/assets/javascripts/surveys/modules/Survey.js
@@ -698,50 +698,45 @@ const Survey = {
     }
   },
 
-  // FIXME: This is supposed to remove a conditional question from
-  // the flow if the condition that it depends on has changed.
-  // However, when this happens it leaves the survey in a state
-  // with no visible questions and no way to proceed.
-  // Disabling this feature means that, once inserted, a conditional
-  // question will not be removed, but that's better than a broken survey.
-  resetConditionalGroupChildren(/* conditionalGroup */) {
-    // const { children, currentAnswers } = conditionalGroup;
+  // To remove a conditional question from the flow if the condition that it depends on has changed.
+  resetConditionalGroupChildren(conditionalGroup) {
+    const { children, currentAnswers } = conditionalGroup;
 
-    // if ((typeof currentAnswers !== 'undefined' && currentAnswers !== null) && currentAnswers.length) {
-    //   const excludeFromReset = [];
-    //   currentAnswers.forEach((a) => { excludeFromReset.push(a); });
-    //   children.forEach((question) => {
-    //     const $question = $(question);
-    //     let string;
-    //     if ($question.data('conditional-question')) {
-    //       string = $question.data('conditional-question');
-    //     } else {
-    //       string = $question.find('[data-conditional-question]').data('conditional-question');
-    //     }
-    //     const { value } = Utils.parseConditionalString(string);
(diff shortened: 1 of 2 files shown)


ragesoss:
Regarding the open question section... that question group preview feature used to work, I think. not sure what's on with it, but feel free to open an issue. We don't use that, and I'd be okay with just removing the feature, but understanding why it doesn't work would be ideal.

Abishekcs:
> Regarding the open question section... that question group preview feature used to work, I think. not sure what's on with it, but feel free to open an issue. We don't use that, and I'd be okay with just removing the feature, but understanding why it doesn't work would be ideal.

Okay

PR #6110, opened 2025-01-10
Fix: Added rapid: false parameter to textarea set method 

Formasitchijoh:
## What this PR does
This PR aims to solve the issue that occurs where the multiwiki-assignment-spec fails intermittently this is mostly due to from the observation from the screen shot that was gotten from the CI build

usually if fails with the following error

Assign All button not found this is because

The Capybara defaulut for the set parameter of an input fiels is set to true meaning is runs in a  faster "rapid" mode, which causes a race condition if the imput fiels has to process the input passed to generate the final test which is the case here as mentioned in this discussion , so from the screen shot we see the text passed to the input field is this `Terre\nhttps://fr.wikipedia.org/wiki/Anglais` but when the test was ran it had just this `Terrais` which is just a single word with no new line so the `Assign all ` text is not displayed cause it appreas just when a new line is added with another text, so the text fails because the typying speed for capybara is too fast since we have some Javascript interactions on form inputs

The other failure which had to do with the text `Heyder Cansa` and `Wp/kiu/Hey` not found, the same applies it to it as above simce the typing speed is fast the  interactions with the javascript is not complte to fully rendere the correct content hence these text cannot be found on the screen

### DIFF
--- a/spec/features/multiwiki_assignment_spec.rb
+++ b/spec/features/multiwiki_assignment_spec.rb
@@ -48,8 +48,6 @@ describe 'multiwiki assignments', type: :feature, js: true do
   end
 
   it 'creates valid assignments from multiple article titles' do
-    pending 'Fails in CI caused by Assign all button not found'
-
     VCR.use_cassette 'multiwiki_assignment' do
       visit "/courses/#{course.slug}/students/articles"
       first('.student-selection .student').click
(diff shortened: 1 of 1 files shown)


ragesoss:
Interesting! Either `#first` changed behavior at some point, or I was just mistaken about it not using wait behavior.

PR #6122, opened 2025-01-14
Update kt-paperclip to version 7.2.2

Formasitchijoh:
## What this PR does
This PR updates the kt-paperclip version from `7.2.0` to `7.2.2`, along with it's dependencies,  in an attempt to fix the failure that occurs with the `Identify command not found for ImageMagick` that is used by kt-paperclip


ragesoss:
I think we determined that this didn't fix the ImageMagick error, but we may as well make these upgrades as they don't seem to break anything.

PR #6124, opened 2025-01-15
[Data rearchitecture] Implement recent revisions for timeslices

gabina:
## What this PR does
This PR implements a new `RevisionStatTimeslice` class to replace the existing `RevisionStat` timeslice. Since raw revisions are no longer available, we calculate recent revisions based on the latest timeslices.

- For course-level recent revisions, we use the most recent `CourseWikiTimeslices` for the course. As we cannot assume a fixed timeslice duration, we compute a daily ratio of revisions (revisions per day) and multiply it by the `REVISION_TIMEFRAME` to determine the total count.
- For user-level recent revisions, we rely on the latest `CourseUserWikiTimeslices` for the course and user. This approach differs from the original implementation as it only considers revisions relevant to the course. In contrast, the previous method included all user edits, even those from namespaces like user talk, which are not directly pertinent to the course.

## Screenshots
Before:
Recent revisions was 0

After:
Recent revisions for course
image

Recent revision for editors
image

As noted, the recent revision counts generated by the timeslice-based approach are not identical to the previous implementation. In particular, the user-level recent revision counts differ, leading to some discrepancies in the web application. For example:
image

To address this, we may want to consider updating the tooltip to provide additional clarity regarding the changes in how revisions are calculated.

### DIFF
--- a/app/models/course_data/courses_users.rb
+++ b/app/models/course_data/courses_users.rb
@@ -144,7 +144,8 @@ class CoursesUsers < ApplicationRecord
     update_values_from_timeslices
 
     # recent_revisions field doesn't belong to timeslices
-    self.recent_revisions = RevisionStat.recent_revisions_for_courses_user(self).count
+    self.recent_revisions = RevisionStatTimeslice.new(course)
+                                                 .recent_revisions_for_courses_user(self)
     # assigned_article_title field doesn't belong to timeslices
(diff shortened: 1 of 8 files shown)


Formasitchijoh:
@gabina  a fix has been made to the failure you are having in your build. Please Could you update your branch with master

Issue #5713, opened 2024-03-19
Dialog word present in sidebar of styleguide


psankhe28:
### What is happening?
The dialog is absent, yet the word "dialog" is displayed in the sidebar.

### To Reproduce

> Steps to reproduce the behavior:
> 1. Go to 'http://localhost:3000/styleguide'
> 2. Click on 'dialogs.'
> 3. See error

### Expected behavior

Either include a dialog or eliminate it from the sidebar.

### Screenshots

![image](https://github.com/WikiEducationFoundation/WikiEduDashboard/assets/84843461/7587468c-66f0-4dd7-b16a-978224d7b64b)

### System Information

 - OS: Windows
 - Browser Chrome

adydams:
@ragesoss I am working on this

ujjwaldubey1:
@adydams are still working on it ,or Can I strt working on it ?

Jay2006sawant:
@ragesoss I would like to work on this issue.

ragesoss:
@Jay2006sawant I expect it's still open; you can check if the issue is still present in your dev environment.

Jay2006sawant:
@ragesoss I have verified the issue in my development environment, and it persists. I would like to take on this task and work towards resolving it.

ragesoss:
Go for it!

Jay2006sawant:
@ragesoss Could you please review this [PR](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6125)

Jay2006sawant:
@ragesoss could you please review this [PR](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6126#issue-2789346186)

PR #6126, opened 2025-01-15
Bugfix: Remove Unused Dialog Link from Navigation

Jay2006sawant:
With reference to #5713 

## What this PR does
This PR addresses the issue of the "Dialogs" link in the navigation sidebar of the Style Guide, which currently points to an incomplete or missing section.

## Screenshots

Before:
Screenshot 2025-01-15 114705
After:
Screenshot 2025-01-15 110625

Should the "Dialogs" section be completely removed until dialogs are implemented, or should we keep the placeholder content for future reference?

Are there any existing dialog components in the codebase that should be showcased in this section? If so, guidance on identifying them would be helpful.

### DIFF
--- a/app/views/styleguide/_nav.html.haml
+++ b/app/views/styleguide/_nav.html.haml
@@ -9,8 +9,7 @@
     %a{href: "#buttons"}Buttons
   %li
     %a{href: "#formcontrols"}Form Controls
-  %li
-    %a{href: "#forms"}Forms
+ 
   %li
     %a{href: "#dialogs"}Dialogs
   %li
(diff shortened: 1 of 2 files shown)


app/views/styleguide/_nav.html.haml
```diff
@@ -9,8 +9,7 @@
     %a{href: "#buttons"}Buttons
   %li
     %a{href: "#formcontrols"}Form Controls
-  %li
-    %a{href: "#forms"}Forms
```
Abishekcs, line 13: This is removing Forms link not the dialog.

PR #6127, opened 2025-01-15
Bugfix: Remove Unused Dialog Link from Navigation

Jay2006sawant:
With refrence to #5713 
## What this PR does
This PR addresses the issue of the "Dialogs" link in the navigation sidebar of the Style Guide, which currently points to an incomplete or missing section.

## Screenshots
Before:

Screenshot 2025-01-15 114705

After:
Screenshot 2025-01-15 110625

Should the "Dialogs" section be completely removed until dialogs are implemented, or should we keep the placeholder content for future reference?

Are there any existing dialog components in the codebase that should be showcased in this section? If so, guidance on identifying them would be helpful.

### DIFF
--- a/app/views/styleguide/_dialogs.html.haml
+++ b/app/views/styleguide/_dialogs.html.haml
@@ -1,2 +1,2 @@
 .sg-note
-  TODO: wizard, unsure of where else
+Dialogs are not implemented yet. This section will be updated once dialogs are added.
--- a/app/views/styleguide/_dialogs.html.haml
+++ b/app/views/styleguide/_dialogs.html.haml
@@ -1,2 +1,3 @@
 .sg-note
-Dialogs are not implemented yet. This section will be updated once dialogs are added.
+TODO: wizard, unsure of where else
+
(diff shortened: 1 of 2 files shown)


Jay2006sawant:
@ragesoss could you please review this PR

app/views/styleguide/_dialogs.html.haml
```diff
@@ -1,2 +1,2 @@
 .sg-note
-  TODO: wizard, unsure of where else
+Dialogs are not implemented yet. This section will be updated once dialogs are added.
```
ragesoss, line 2: I don't think this is necessary or useful.

Jay2006sawant: @ragesoss Thank you for your feedback. Should I remove the statement and revert it to its previous version?

ragesoss: yes

app/views/styleguide/_dialogs.html.haml
```diff
@@ -1,2 +1,3 @@
 .sg-note
-  TODO: wizard, unsure of where else
+TODO: wizard, unsure of where else
```
ragesoss, line 2: This line should not be changed at all. I recommend cleaning up the PR to include only one commit with the change to the other file.

PR #6128, opened 2025-01-15
Fix: Improve error handling when `get_page_content` returns `nil` when called by `WikiCourseEdits`

empty-codes:
## What this PR does
Improves error handling when `get_page_content` returns `nil` when called by `WikiCourseEdits`:
- Creates a custom `NilPageContentError` with a (more) descriptive error message
for example, "Failed to fetch content for User:Belajane41/sandbox. Initial page content is nil."

- Refactors `initialize` method of `AddSandboxTemplate`, `WikiCourseEdits`'s `add_template_to_user_page`, and `add_template_to_user_talk_page` methods to guard for and raise a `NilPageContentError` if `initial_page_content` is `nil`:

```
exception = Errors::PageContentErrors::NilPageContentError.new(user_page / talk_page / @sandbox)
raise exception if initial_page_content.nil?
```

- Adds test to ensure this expected behaviour is enforced in both classes
#### Before:
`add_sandbox_template_spec.rb`:
image

`wiki_course_edits_spec.rb`:
image

#### After:
`add_sandbox_template_spec.rb`:
image

`wiki_course_edits_spec.rb`:
image

- Adds `NilPageContentError` error type to sentry config's `excluded_exceptions` so occurences of it are not sent to Sentry since we know that retry via Sidekiq will take care of it. See [Sentry docs](https://docs.sentry.io/platforms/ruby/configuration/options/#excluded-exceptions) and [code implementation](https://github.com/getsentry/sentry-ruby/blob/a1d294128e3d0fa3f5ca80d53285ebf0a59bec8b/sentry-rails/lib/sentry/rails/configuration.rb#L15).

---

Note: The error is referenced with its full namespace path `Errors::PageContentErrors::NilPageContentError`  when used/called because the other option of:
```
require_dependency "#{Rails.root}/lib/errors/page_content_errors"
include Errors::PageContentErrors

#then calling it like
exception = NilPageContentError.new(talk_page)
raise exception if initial_page_content.nil?

```
results in NameErrors like:
```
 NameError:
       uninitialized constant WikiCourseEdits::PageContentErrors

           exception = PageContentErrors::NilPageContentError.new(user_page)
                       ^^^^^^^^^^^^^^^^^
```
due to the concerned classes being in `/services` and `/libs` + auto loading / eager loading / Zeitwerk configs. Alternatively could just be a skill issue on my part :-

### DIFF
--- a/lib/errors/page_content_errors.rb
+++ b/lib/errors/page_content_errors.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+module Errors
+  module PageContentErrors
+    class NilPageContentError < StandardError
+      def initialize(page)
+        msg = "Failed to fetch content for #{page}. Initial page content is nil."
+        super(msg)
+      end
+    end
+  end
+end
--- a/app/services/add_sandbox_template.rb
+++ b/app/services/add_sandbox_template.rb
@@ -10,12 +10,19 @@ class AddSandboxTemplate
     @sandbox = sandbox
     @sandbox_template = sandbox_template
     @wiki_editor = WikiEdits.new(home_wiki)
-    wiki_api = WikiApi.new(home_wiki)
-    @initial_page_content = wiki_api.get_page_content(@sandbox)
+    @wiki_api = WikiApi.new(home_wiki)
+    initial_page_content
     add_template
   end
 
   private
+
+  def initial_page_content
+    @initial_page_content = @wiki_api.get_page_content(@sandbox)
+    exception = Errors::PageContentErrors::NilPageContentError.new(@sandbox)
+    raise exception if @initial_page_content.nil?
+    @initial_page_content
+  end
 
   def add_template
     # Never double-post the sandbox template
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,4 +1,9 @@
+# frozen_string_literal: true
+
 Sentry.init do |config|
   config.dsn = ENV['sentry_dsn']
   config.enabled_environments = %w[development staging production test]
+
+  # Appends to IGNORE_DEFAULT array of exception classes that should never be sent to Sentry
+  config.excluded_exceptions += ['Errors::PageContentErrors::NilPageContentError']
 end
--- a/lib/wiki_course_edits.rb
+++ b/lib/wiki_course_edits.rb
@@ -154,6 +154,10 @@ class WikiCourseEdits
 
     # Never double-post the enrollment template
     initial_page_content = @wiki_api.get_page_content(user_page)
+
+    exception = Errors::PageContentErrors::NilPageContentError.new(user_page)
+    raise exception if initial_page_content.nil?
+
     return if initial_page_content.include?(template)
 
     summary = @generator.enrollment_summary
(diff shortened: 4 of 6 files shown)


lib/errors/page_content_errors.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module Errors
+  module PageContentErrors
+    class NilPageContentError < StandardError
```
ragesoss, line 5: I think this may be a confusing way of naming and handling this error.

We have a pretty good idea that this error is specific to WikiApi when it attempts to fetch page content and fails to do so, and it's not a 404.

After looking over the different places in the code that use WikiApi#get_page_content, I think none of them are written in a way that explicitly handles this kind of failure adequately, as the current behavior generally treats an empty string (returned from a 404) the same way as `nil`, so we may be missing other problems that we don't know about.

My current thinking is that we should actually raise an error from WikiApi rather than return `nil`, and then figure out the next steps from there.

empty-codes:
```
 # Returns nil if it cannot get any info from the wiki, but returns
  # empty string if it's a 404 because the page is a redlink.
  def get_page_content(page_title)
    response = mediawiki('get_wikitext', page_title)
    case response&.status
    when 200
      response.body.force_encoding('UTF-8')
    when 404
      ''
    end
  end
```
Should i add a case for the error being 429 and let it raise an error from there, or a general else case and raise the error from there? or both?
 I could create an error class specific to 429 errors since that's what we assume is happening and raise it when the status is 429, then raise the more general NilPageContentError (renamed) for the else case.
 
 In that case would the nil guards still be needed in the code that use WikiApi#get_page_content, since when errors are raised, I assume the entire method execution will immediately stop.

ragesoss: I think we could just throw the errors from here, and decide later whether to add error handling at any of the specific places in the code that are affected by it.

The trick here is that there are some 429 errors that we are intentionally basically ignoring, but not ones that use .get_page_content. Ideally, we'd be raising those errors based on status within the method that actually makes the call, so that would wouldn't need parallel handling at every method that share `mediawiki`. But 429 is the only kind of error that `mediawiki` already handles, so I don't think you'll be able to detect when it's a 429 here unless you modify `mediawiki`, for example, by adding an optional parameter for raising an error rather than returning nil. I guess that's probably the right strategy here.

empty-codes: Got it, I'll revert the changes except the custom error class and add this to wiki_api:
```
else
      exception = Errors::PageContentErrors::NilPageContentError.new(page_title, response&.status)
      raise exception
    end
```
so the new message will be: `"Failed to fetch content for #{page} with response status: #{status}."`

For the renaming of the error, I can as well remove the module and just have a direct class like `Errors::PageFetchError`. I'm not too clear on what a good name is though, so your suggestions are appreciated.

About `mediawiki`, I took a look at the code in the separate repo and I ended up confusing myself a bit as to why 429 errors would occur here given that other 429 instances usually log a `MediawikiApi::HttpError: unexpected HTTP response (429)` at this line : `@mediawiki.send(action, query)` so maybe it's other error codes causing this?

ragesoss: Hmm... I'm not sure.

For the error, since this is a type of error we expect to only originate from WikiApi, my suggestion would be to define it within WikiApi rather than in the separate module.

empty-codes: Created a new PR and branch for it: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6158

PR #6137, opened 2025-01-17
Fixed : On-wiki training_slide with quiz doesn't include complete question prompt

### LINKED ISSUE #5823: On-wiki training slide with quiz doesn't include complete question prompt
Here's a quiz slide: https://outreachdashboard.wmflabs.org/training/editing-wikipedia/plagiarism/close-paraphrasing-quiz

The wiki source page includes a quoted passage which does not show up on the Dashboard: https://meta.wikimedia.org/wiki/Training_modules/dashboard/slides/11511-close-paraphrasing-quiz

This may be a bug in WikiSlideParser, or somewhere else along the path to loading training content from a wiki page.

omChauhanDev:
## What this PR does
Fixes #5823

This PR improves quiz question formatting by:
- Adding multi-line support for quiz content parsing
- Removing unwanted wiki markup (colons and triple quotes) from rendered content
- Preserving proper line breaks and paragraph formatting

## Technical Changes
- Updated `template_parameter_value` in `WikiSlideParser` to handle multi-line content
- Modified regex pattern to safely remove wiki formatting characters
- Added markdown processing for quiz questions in the frontend Quiz component

## Screenshots
Before:
Before

After:
After

### DIFF
--- a/lib/training/wiki_slide_parser.rb
+++ b/lib/training/wiki_slide_parser.rb
@@ -216,9 +216,13 @@ class WikiSlideParser
   end
 
   def template_parameter_value(template, parameter)
-    # Extract value from something like:
-    # | parameter_name = value
-    match = template.match(/\|\s*#{parameter}\s*=\s*(?<value>.*)/)
-    match && match['value']
+    # Extract value from the template
+    match = template.match(/\|\s*#{parameter}\s*=\s*(?<value>.*?)(?=\||\Z)/m)
+
+    # Check if match is present and value is not nil
+    value = match && match['value']
+
+    # Remove undesired formatting and whitespace, only if value is not nil
+    value ? value.gsub(/(^:|''')/, '').strip : nil
   end
 end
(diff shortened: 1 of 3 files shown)


lib/training/wiki_slide_parser.rb
```diff
@@ -216,9 +216,13 @@ def module_hash(template, module_number)
   end
 
   def template_parameter_value(template, parameter)
-    # Extract value from something like:
-    # | parameter_name = value
-    match = template.match(/\|\s*#{parameter}\s*=\s*(?<value>.*)/)
-    match && match['value']
+    # Extract value from the template
+    match = template.match(/\|\s*#{parameter}\s*=\s*(?<value>.*?)(?=\||\Z)/m)
```
ragesoss, line 220: This is a complicated regex, so it should include comments to explain what each part is doing.

Also, the example from the comment should not be removed. If necessary, update that example or add a second one for the case this is dealing with.

ragesoss (commented):
This should include a new test in wiki_slide_parser_spec.rb, which fails without the update to WikiSlideParser but passes with it. The exact slide content from the bug report can probably be used for this.

omChauhanDev:
## Screenshots (for test case)
Before updating WikiSlideParser:
Before

After updating WikiSlideParser:
After



## Testing
- Added new quiz with multi-line content
- Verified formatting in both backend parsing and frontend display
- Confirmed existing quiz functionality remains unchanged

PR #6144, opened 2025-01-20
Fix State Mutation in Timeline Reducer and Refactor TimelineHandler for Immutable Week Title Updates

Abishekcs:
## Note

Previously, I considered opening this as two separate PRs (#6138 and #6143). However, later i realized both require a common component change, specifically the TimelineHandler component.

For testing please set `enableMutationChecks` to true in create_store.js

## What this PR does
- Fixes state immutability issues in the weeks property of the timeline slice done by `getWeeksArray` selector.
- Fixes mutation in timeline reducer(s) which are handling the action `RESET_TITLES` & `UPDATE_TITLE`

## Cause of Bug (For `getWeeksArray` selector)
 - The bug lies in the direct modification of the `newWeek object`, which is a reference to the `weeks` property of the 
Redux state (Timeline Slice).

https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/88a89ecc3e5bac1092b875977d16f4438d7abd32/app/assets/javascripts/selectors/index.js#L257

## Fix  (For `getWeeksArray` selector)
- Resolved by using `spread operator` since we are modifying only the top level properties of the object.

## Cause of Bug (For `RESET_TITLES` & `UPDATE_TITLE`)

For this finding the cause of bug was easy as the error message exaclty indicated where mutation was as shown below:

**For** `RESET_TITLES`:

RESET_TITLES

**For** `UPDATE_TITLE` 

UPDATE_TITLE

The spread operator doesn’t work as expected with deeply nested objects because it copies the values of the top-level properties of weeks to new addresses/references, but the inner objects retain the same addresses/references which are the one's updated by mutation.

## Fix (For `RESET_TITLES` & `UPDATE_TITLE`)
Resolved by using [`Immer Produce`.](https://immerjs.github.io/immer/produce/)

## Why the changes in TimelineHandler component?

Previously, the `getWeeksArray` selector was mutating the `weeks` property of the `timeline slice` directly. This meant that dispatching `SAVED_TIMELINE` from the `saveTimeline` function immediately after dispatching `RESET_TITLES` from the `_resetTitles` function worked without issues because both were referencing the same mutated weeks array from `getWeeksArray` selector.

So, after updating immutable way immediately dispatching won't work as the existing weeks in the TimlineHandler from selector doesn't have the same reference.

**Little bit detail about redux for future reference:**
 - **In Redux**:
      - **Mutation**: Changes are immediate since we were previously directly modifying the same object reference - 
            any code holding that reference sees the changes instantly.
      - **Immutable**: Creates a new object reference, requiring React/Redux to process and batch the state update, 
            making the changes not immediately available. Also, after resetting the timeline week titles, the saveTimeline 
            was called immediately, leading to an API call right away. During this time, Redux had not yet received the 
           new updates.
      - **Source**: [Immutability in React and Redux: The Complete Guide](https://daveceddia.com/react-redux-immutability-guide/) 

## Screenshots (For `getWeeksArray` selector)
Before:

https://github.com/user-attachments/assets/9037304e-38b5-45c0-9993-f352f7743332

After:

https://github.com/user-attachments/assets/d70cd94f-6c7c-471b-9c5a-ab17e0938931

## Screenshots (For `RESET_TITLES` & `UPDATE_TITLE`)
Before:

https://github.com/user-attachments/assets/9803c068-e403-4c84-8ae0-0bc86f835301

After:

https://github.com/user-attachments/assets/f8751d9e-d901-4263-99a8-0ed18839d7a6

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2817,6 +2817,7 @@ __metadata:
     eslint-webpack-plugin: ^3.1.1
     fetch-jsonp: ^1.2.1
     i18n-js: ^4.2.2
+    immer: ^10.1.1
     jeet: ^7.2.0
     jest: ^26.0.1
     jquery: ^3.5.1
@@ -6866,7 +6867,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"immer@npm:^10.0.3":
+"immer@npm:^10.0.3, immer@npm:^10.1.1":
   version: 10.1.1
   resolution: "immer@npm:10.1.1"
   checksum: 07c67970b7d22aded73607193d84861bf786f07d47f7d7c98bb10016c7a88f6654ad78ae1e220b3c623695b133aabbf24f5eb8d9e8060cff11e89ccd81c9c10b
--- a/app/assets/javascripts/components/timeline/timeline_handler.jsx
+++ b/app/assets/javascripts/components/timeline/timeline_handler.jsx
@@ -1,5 +1,5 @@
 // Import necessary hooks (useState, useEffect) and useNavigate from react-router-dom
-import React, { useEffect, useState } from 'react';
+import React, { useEffect, useState, useRef } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import TransitionGroup from '../common/css_transition_group';
@@ -25,11 +25,20 @@ const TimelineHandler = (props) => {
   const [reorderable, setReorderable] = useState(false);
   const [editableTitles, setEditableTitles] = useState(false);
 
-// Replace componentDidMount with useEffect hook
+  const resetWeekTitles = useRef(false);
+
+  // Replace componentDidMount with useEffect hook
   useEffect(() => {
     document.title = `${props.course.title} - ${I18n.t('courses.timeline_link')}`;
     props.fetchAllTrainingModules();
   }, [props.course.title, props.fetchAllTrainingModules]);
+
+  useEffect(() => {
+    if (resetWeekTitles.current) {
+      saveTimeline();
+      resetWeekTitles.current = false;
+    }
+  }, [props.weeks]);
 
 // Convert class methods to regular functions within the component
   const _cancelBlockEditable = (blockId) => {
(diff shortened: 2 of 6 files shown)


Abishekcs:
@ragesoss would you be open to using [Immer](https://immerjs.github.io/immer/) for immutability updates? Redux Toolkit has Immer built-in, but we'd need to refactor our reducer file to use it.

ragesoss:
@Abishekcs sure, seems reasonable.

ragesoss:
This seems fine to me, but since it's introducing a new tool, the PR should probably include some explanation of it in frontend.md.

Also, as part of the PR description, please note how adding this package affect the bundle size.

yarn.lock
```diff
@@ -6866,7 +6867,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"immer@npm:^10.0.3":
+"immer@npm:^10.0.3, immer@npm:^10.1.1":
```
ragesoss, line 6870: Since it looks like we're already using a different version of immer, is it possible to just use the newer version without the older one (possibly by upgrading whatever dependency was already requiring it)?

Abishekcs:
> This seems fine to me, but since it's introducing a new tool, the PR should probably include some explanation of it in frontend.md.
> 
> Also, as part of the PR description, please note how adding this package affect the bundle size.

Okay

yarn.lock
Abishekcs: @ragesoss i think both version ranges (^10.0.3 and ^10.1.1) are being satisfied by a single installation of version 10.1.1.  as shown in yarn .
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/8e6a3a267e4295783b7792c41fa539429567f088/yarn.lock#L6871

So there's only one version installed, and it satisfies both requirements.
And, while installing Redux Toolkit (RTK) in [PR #5867 ](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/5867), it brought in immer as a dependency since RTK uses immer internally for its immutable state updates.

ragesoss: Great. So it shouldn't affect the bundle size either.

Abishekcs: > Great. So it shouldn't affect the bundle size either.

Yes.

Abishekcs:
@ragesoss i have updated the frontend.md.

a) There, is one link  i.e Thinking in React  which leads to page not found.
b) Also, the doc needs to be updated to reflect the recent changes in react and redux. I will make those in seperate PR related only to the docs.

ragesoss:
Thanks! This looks good to me; I'll merge once the tests pass.

PR #6145, opened 2025-01-21
[Data rearchitecture] Track errors when getting scores

gabina:
## What this PR does
This PR marks timeslices as 'needs_udpate' when requests to LiftWing or ref-counter APIs fail.

The `RevisionScoreImporter` class is in charge of importing revision scoring data from Lift Wing and reference-counter APIs.  In the revision version, this importer is used as part of the `UpdateCourseStats`. Right after importing revisions, we import revision scoring data for all those revisions that are missing them. That means that we calculate “unscored revisions” for the entire course each time `UpdateCourseStats` executes. The advantage of this is that if a Lift Wing API or reference-counter API call fails due to a transient error, the next time the update runs for that course, we will retry to get the scores, and all the caches will be recalculated.

In the timeslice version, we have a `RevisionDataManager` class that fetches new revisions and revision scoring data for them. Right after fetching this, we process it and calculate timeslice cache that we persist to the db. If the the Lift Wing API or reference-counter API call fails due to a transient error, we’ll miss scores for that timeslice, probably without even noticing it. A consequence of this could be that the added references are left at 0, because they could not have been correctly retrieved.

This PR introduces an initial simple strategy for handling errors during score fetching. If an error occurs, it is propagated to the associated Revision record. This allows the timeslice to update its value while keeping the `needs_update` flag set to true.

## Screenshots
Before:

After:

### DIFF
--- a/app/models/course_wiki_timeslice.rb
+++ b/app/models/course_wiki_timeslice.rb
@@ -94,7 +94,7 @@ class CourseWikiTimeslice < ApplicationRecord
     update_uploads_in_use_count
     update_upload_usages_count
     update_stats
-    self.needs_update = false
+    update_needs_update
     save
   end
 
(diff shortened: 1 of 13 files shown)


gabina:
Failing spec isn't related to these changes.

PR #6146, opened 2025-01-21
Fix: Prevent occurrences of `TypeError`s

empty-codes:
This PR refactors code to prevent occurrences of:

- `TypeError: undefined is not an object (evaluating 'window.navigator.userAgentData.platform')` in `course.js` by adding optional chaining operators to handle when `window.navigator.userAgentData` is `undefined`.

- `TypeError: Cannot read properties of undefined (reading 'childNodes')` in `course.js` by Introducing a helper function to check table validity as an extension of  the `document.querySelectorAll('#users table').length` check. It now checks that at least one table is present and the table(s) has a body and rows`.

 - `TypeError: Cannot read properties of undefined (reading 'style')` in `list.jsx` by ensuring `floatingHeaderRow` exists before accessing its `style` property.
 
- `TypeError: Cannot read properties of null (reading 'cloneNode')` in `article_viewer.js` by ensuring `.user-legend-wrap` exists before attempting to clone it.

- `Error: Objects are not valid as a React child (found: object with keys {error}). If you meant to render a collection of children, use an array instead.` in `notifications.js` by ensuring `notification.message` is always a string when it is passed to the React component for rendering.

- `NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.` and `NotFoundError: The object can not be found here.` in:
   -  `training_slide_handler.jsx` by wrapping `<FastTrainingAlert/>` in a `<div>` element. 
   -  in `tag_list.jsx` by wrapping the string `I18n.t('courses.none')` in a `<span>` element.
 The error happens when google translate interferes with the DOM and then the DOM element's native removeChild method crashes.

- `TypeError: Cannot read properties of undefined (reading 'setUploadFilters')` in `StudentRevisionRow.jsx` by removing `this` as it is undefined in functional components and importing `setUploadFilters` directly.

- `TypeError :Cannot read properties of undefined (reading 'wp10')`, `TypeError: d[0] is undefined` and `TypeError: undefined is not an object (evaluating 'd[0].wp10')` by adding checks to ensure `articleData` is a non-empty array

- `TypeError: Cannot read properties of undefined (reading 'coordinates')` and `TypeError: Cannot read properties of undefined (reading 'filteredParagraphs')` by adding optional chaining and fallback values when `this.scrollObject[name]` is undefined


test/reducers/notifications.spec.js
```diff
@@ -135,12 +135,13 @@ describe('notifications reducer', () => {
...
+    // when data.responseText is invalid, notification.message is assigned
+    // the data object as the last fallback. This is however serialized into
+    // a string as React cannot render objects.
     const notification = {
       closable: true,
       type: 'error',
-      message: {
-        responseText: '{"message":"Example notification status"'
-      }
+      message: '{"responseText":"{\\"message\\":\\"Example notification status\\""}',
```
empty-codes, line 144: The first run failed because my branch was not up to date and the recently added `notification.spec.js` was not present so the tests passed when I ran it locally. 
I updated my branch and also updated this failing test to expect a string and not an object because it was based on the error-ing code that causes this `Error: Objects are not valid as a React child (found: object with keys {error}). If you meant to render a collection of children, use an array instead.` in `notifications.js`.

ragesoss:
LGTM!

PR #6151, opened 2025-01-23
Optimized training_tool_spec

Formasitchijoh:
## What this PR does
This PR Optimizes training_tool_spec runtime by removing unnecessary waits and using a module with fewer slides, to reduce the build time of this test group 

As of the screenshot below it takes 2min: 25sec  to run this test group it started `Test suite started at: 2025-01-24 07:54:19 UTC` and ended  `Test suite ended at: 2025-01-24 07:56:34 UTC`

After Screen Shot

Screenshot 2025-01-24 at 13 36 23

## Notes
- This optimization is evident in the amount of time it takes to complete this build `31min `

### DIFF
--- a/spec/features/training_tool_spec.rb
+++ b/spec/features/training_tool_spec.rb
@@ -2,11 +2,11 @@
 
 require 'rails_helper'
 
-DESIRED_TRAINING_MODULES = [{ slug: 'evaluating-articles' }].freeze
+DESIRED_TRAINING_MODULES = [{ slug: 'sandboxes-talk-watchlists' }].freeze
 
 describe 'Training', type: :feature, js: true do
   let(:user) { create(:user, id: 1) }
-  let(:module_2) { TrainingModule.find_by(slug: 'evaluating-articles') }
+  let(:module_2) { TrainingModule.find_by(slug: 'sandboxes-talk-watchlists') }
   let(:new_instructor_orientation_module) do
     TrainingModule.find_by(slug: 'new-instructor-orientation')
   end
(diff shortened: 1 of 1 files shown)


ragesoss:
Nice. Can you provide some data about which segments of this test file take how much time? 8 minutes for this is still really long, and I don't think it used to be nearly as long as this.

Formasitchijoh:
> Nice. Can you provide some data about which segments of this test file take how much time? 8 minutes for this is still really long, and I don't think it used to be nearly as long as this.

Loading the data for the modules seems to take very long, when I  ran it locally with my connection disconnected it seemed to pass and took about 2m, when when I reconnected to the internet it took this 8/10 min

I have a question in the TrainingBase class which loads the modules and slides from  either from the  `WikiTrainingLoader` or the `YamlTrainingLoader` 
does it mean the  `WikiTrainingLoader` is the remote server?

ragesoss:
Is that training loading happening in Wiki Education mode, or Peony mode? WikiTrainingLoader is used in Peony mode, and takes much longer. Ideally, we should find a way to still test that loading from wiki works, process as few modules and slides as possible.

Formasitchijoh:
Currently my local is set to the Wiki Education mode, 
is it possible to switch mode when the test is running ?

ragesoss:
overall, the code at the top of config/environments/test.rb controls the mode for the test suite. individual tests sometimes override that, usually via `allow(Features).to receive(:wiki_ed?).and_return(true)`.

Formasitchijoh:
> overall, the code at the top of config/environments/test.rb controls the mode for the test suite. individual tests sometimes override that, usually via `allow(Features).to receive(:wiki_ed?).and_return(true)`.

Alright I see, but this test does not override this  `allow(Features).to receive(:wiki_ed?).and_return(true)`

ragesoss:
Hmm... maybe that can simply be changed. There are two places in `training_tool_spec` that explicitly deal with `allow(Features)`, and those are related to specific handling of one config versus the other. Everything else in this spec seems (on a quick skim) like it should not depend on which mode is active.

Formasitchijoh:
> allow(Features)

Yeah switching the does not seem to change anything much,  I am currently adding some logs to each describe block to give the exact time each block takes

ragesoss:
Are you still working this one, or is it ready to merge?

Formasitchijoh:
There are some logs I added  I will need to  remove that and ping you.

Formasitchijoh:
@ragesoss  I think this is ready to be merged

PR #6153, opened 2025-01-24
Fix: Failure in account_request and assigned_article spec

Formasitchijoh:
## What this PR does

This PR fixes recent failures in the account_request and assigned_article spec 

### Changes made
- Added account request spec to pending
-  included fixtures for dynamically loading assigned articles, reducing interaction with external api which takes time and fails  intermittently


ragesoss:
I'm okay with starting to selectively save some of the VCR fixtures for specific flaky APIs, but I'd like them to be clearly labeled and differentiated, as we don't want to add them all by default.

.gitignore
```diff
@@ -3,7 +3,6 @@
 /config/newrelic.yml
 /config/secrets.yml
 /dockerAuth.json
-/fixtures
```
ragesoss, line 6: I suggest that you do something like add `!/fixtures/committed` and use that for all the fixtures that will be saved in git, so that other fixtures that we don't need to save don't get added by accident. That will make it easy within the test files to change a fixture so that it gets saved, by adding `committed/` to the start of the VCR cassette name.

Formasitchijoh: I think that makes sense , I will update that

ragesoss:
LGTM!

PR #6147, opened 2025-01-22
Refactor alert monitors to remove Revisions table dependency

Abishekcs:
#6141

## What this PR does

This PR updates the logic for the following alerts to eliminate their reliance on the deprecated `Revisions` table:

- ArticlesForDeletionAlert
- GANominationAlert
- DYKNominationAlert

PR #6156, opened 2025-01-24
Refactor monitors for `HighQualityArticleEditAlert` and `DiscretionarySanctionsEditAlert`.

### LINKED ISSUE #6141: Rewrite some alert monitor logic to work without using Revisions table
### Context:
The system has several alerts to send emails to editors or admins based on certain criteria (see class `Alert` and all its sub-classes).

Some alerts use `Revisions` table to retrieve some useful data. As we're close to stop using that table at all, we should replace the places where data from that table is used.

### Changes required
The monitors for the alerts `ArticlesForDeletionAlert`, `GANominationAlert` and `DYKNominationAlert` use the revision table to determine the user id when creating alerts. See `create_alert` definition for `ArticlesForDeletionMonitor`, `GANominationMonitor` and `DYKNominationMonitor`.  One option could be to determine the user id based on the `user_ids` field in the `ArticlesCourses` record.

On the other hand, monitors for `DiscretionarySanctionsEditAlert` and `HighQualityArticleEditAlert` also use the revision id to decide if the alert was resolved. We would probably want to make an API call to determine this.. See `create_alert` definition for `DiscretionarySanctionsMonitor` and `HighQualityArticleMonitor`.

### Additional context:
`ContinuedCourseActivityAlert` has already been rewritten in PR #6132 and merged to the `data-rearchitecture-for-dashboard`

gabina:
## What this PR does
This PR refactors both `DiscretionarySanctionsMonitor` and  `HighQualityArticleMonitor` to work without relying on revisions in the data base. 
The refactor involves two main things:
1. The user id assigned to the alert is taken from `user_ids` `ArticlesCourses` field instead of using the revision (similar to PR #6147)
2. `resolved_edit_alert_covers_latest_revision?` definition changes. To decide whether to create a new alert when a resolved alert already exists, now we query the API to retrieve all edits made to the specified page after the resolved alert's creation date. A new alert is created if any of those edits were made by a course student (I didn't find any way to retrieve revisions for a given page id made by a list of users, the `rvuser` parameter allows a single user according to [docs](https://www.mediawiki.org/wiki/API:Revisions)). Note that we only hit edits made within the period from the `creation date of the last resolved alert` to the `course end date` (would we prefer to use `course end date + UPDATE_LENGTH`?). 

Closes #6141 

1. Now alert record is created with a nil `revision_id`.
2. I made an effort to mock the API requests in the specs instead of using VCR cassette, but I didn't find a way to make the `MediawikiApi::Response`  double instance work when implementing `continue`. 
3. This PR has a lot of  duplicate code. However, because of how the alerts are implemented, I didn't find a good place to put common code. A reasonable thing to do would be to move the logic for retrieving revisions made to a given article to the `WikiAPI` class. However, the current implementation is interrupting requests if it finds a positive result, to avoid making unnecessary continue requests (considering that the previous implementation did not make any requests because it used the persisted revisions in the database). Therefore IMHO it has quite a lot of use-specific logic to move it to the `WikiApi` class.

### DIFF
--- a/lib/alerts/discretionary_sanctions_monitor.rb
+++ b/lib/alerts/discretionary_sanctions_monitor.rb
@@ -57,15 +57,11 @@ class DiscretionarySanctionsMonitor
 
   def create_edit_alert(articles_course)
     return if unresolved_edit_alert_already_exists?(articles_course)
-    revisions = articles_course.course.revisions.where(article_id: articles_course.article_id)
-    last_revision = revisions.last
-    return if resolved_edit_alert_covers_latest_revision?(articles_course, last_revision)
-    first_revision = revisions.first
+    return if resolved_edit_alert_covers_latest_revision?(articles_course)
     alert = Alert.create!(type: 'DiscretionarySanctionsEditAlert',
                           article_id: articles_course.article_id,
-                          user_id: first_revision&.user_id,
-                          course_id: articles_course.course_id,
-                          revision_id: first_revision&.id)
+                          user_id: articles_course&.user_ids&.first,
+                          course_id: articles_course.course_id)
     alert.email_content_expert
   end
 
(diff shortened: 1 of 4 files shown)


lib/alerts/discretionary_sanctions_monitor.rb
```diff
@@ -90,12 +86,60 @@ def unresolved_assignment_alert_already_exists?(assignments_course)
...
+  # [last resolved alert, course end]
+  def query_params(course, page_id, last_resolved_date)
+    {
+      action: 'query',
+      prop: 'revisions',
+      pageids: page_id,
+      rvend: last_resolved_date.strftime('%Y%m%d%H%M%S'),
+      rvstart: course.end.strftime('%Y%m%d%H%M%S'),
+      rvdir: 'older', # List newest first. rvstart has to be later than rvend.
+      rvmax: 500
```
gabina: `rvmax` is not a valid parameter. I corrected it to `rvlimit` in 28c94964beaa737fd7530c901c34e10ccac6ffdd

spec/lib/alerts/high_quality_article_monitor_spec.rb
```diff
@@ -49,9 +49,7 @@
     end
 
     it 'creates Alert records for edited Good articles' do
-      VCR.use_cassette 'high_quality' do
-        described_class.create_alerts_for_course_articles
-      end
+      described_class.create_alerts_for_course_articles
```
gabina, line 52: I don't think we're doing any request in these specs so I deleted the use of VCR cassette.

PR #6158, opened 2025-01-27
Raises a custom `PageFetchError` when the `response.status `is not 200 or 404

empty-codes:
## What this PR does
Raises a custom PageFetchError error when the response.status is not 200 or 404:

- Creates a custom `PageFetchError` with a descriptive error message
- Adds test to ensure this expected behaviour is enforced

## Screenshots
Before:
before pagefetchtest

After:
after pagefetchtest

### DIFF
--- a/spec/lib/wiki_api_spec.rb
+++ b/spec/lib/wiki_api_spec.rb
@@ -63,6 +63,21 @@ describe WikiApi do
         expect(response).not_to be_blank
       end
     end
+
+    it 'raises a PageFetchError for unexpected response statuses' do
+      # some possible status codes
+      unexpected_statuses = [401, 403, 406, 408, 409, 413, 414, 429]
+      unexpected_statuses.each do |status_code|
+        allow_any_instance_of(described_class).to receive(:mediawiki)
+          .and_return(OpenStruct.new(status: status_code))
+        expect do
+          described_class.new.get_page_content('SomePage')
+        end.to raise_error(
+          WikiApi::PageFetchError,
+          /Failed to fetch content for SomePage with response status: #{status_code}/
+        )
+      end
+    end
   end
 
   describe '#fetch_all' do
--- a/spec/lib/wiki_api_spec.rb
+++ b/spec/lib/wiki_api_spec.rb
@@ -11,28 +11,40 @@ describe WikiApi do
 
     it 'handles mediawiki 503 errors gracefully' do
       stub_wikipedia_503_error
-      expect(subject).to eq(nil)
+      expect { subject }.to raise_error(
+        WikiApi::PageFetchError,
+        /Failed to fetch content for Ragesoss with response status: 503/
+      )
     end
 
     it 'handles timeout errors gracefully' do
       allow_any_instance_of(MediawikiApi::Client).to receive(:send)
         .and_raise(Faraday::TimeoutError)
       expect_any_instance_of(described_class).to receive(:log_error).once
-      expect(subject).to eq(nil)
+      expect { subject }.to raise_error(
+        WikiApi::PageFetchError,
+        /Failed to fetch content for Ragesoss with response status: ./
+      )
     end
 
     it 'handles API errors gracefully' do
       allow_any_instance_of(MediawikiApi::Client).to receive(:send)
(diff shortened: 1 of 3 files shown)


ragesoss:
Looks good. I'll merge once the build is finished.

empty-codes:
> Looks good. I'll merge once the build is finished.

The first check failed because I forgot to update the expectations of existing tests which I've done with the second commit. The failure in the second check is not because of my changes however.

ragesoss:
It looks like the last error is a flaky test, but the first three for enrollment controller are probably tests that were inadvertently dependent on the previous behavior of not throwing an error.

empty-codes:
> It looks like the last error is a flaky test, but the first three for enrollment controller are probably tests that were inadvertently dependent on the previous behavior of not throwing an error.

I'll take a look at it now

empty-codes:
> It looks like the last error is a flaky test, but the first three for enrollment controller are probably tests that were inadvertently dependent on the previous behavior of not throwing an error.

They all throw this error:
```
Failure/Error: raise PageFetchError.new(page_title, response&.status)

     WikiApi::PageFetchError:
       Failed to fetch content for User:Ragesock with response status: .
     # ./lib/wiki_api.rb:37:in `get_page_content'
     # ./lib/wiki_course_edits.rb:186:in `remove_content_from_page'
     # ./lib/wiki_course_edits.rb:83:in `disenroll_from_course'
     # ./lib/wiki_course_edits.rb:26:in `initialize'
```

In `enrollment_controller`'s  `remove_assignment_templates` method, when it calls `WikiCourseEdits.new` which also calls `remove_content_from_page` in the disenrollment process, 
the `get_page_content` raises the `PageFetchError`  because the response status is `.`  (cause it's a test) and that meets the else condition so, 
it causes the `remove_assignment_templates` method to exit immediately without completing the rest of the logic (e.g., removing templates from all assignments).

Possible fixes:
- add a case for `.` to return nil in the get_page_content
```
when .
    nil
```
- remove the else case and add a case for any Integer not 404 or 200 to raise the error instead
```
when Integer 
    raise PageFetchError.new(page_title, status)
```
- change the test behaviour /expectations

ragesoss:
Hmm... Why is the response "."?

empty-codes:
> Hmm... Why is the response "."?

I thought the status returned `.` but apparently its just cause `.` in the regex matches any single character including an empty string, so for tests that use mocks to simulate errors and there's no HTTP response, meaning there's no status code / status, the `response.status` is then empty / an empty string, which then falls into the `else` case and the error raises puts `.` as the placeholder.

ragesoss:
It's probably something related to the way the test is written, where previously we could just proceed without worrying about how that fetch content call worked. @Formasitchijoh maybe you can spot the issue?

spec/lib/wiki_api_spec.rb
```diff
@@ -11,28 +11,40 @@ class UnexpectedError < StandardError; end
...
     end
 
     it 'handles timeout errors gracefully' do
       allow_any_instance_of(MediawikiApi::Client).to receive(:send)
         .and_raise(Faraday::TimeoutError)
       expect_any_instance_of(described_class).to receive(:log_error).once
-      expect(subject).to eq(nil)
+      expect { subject }.to raise_error(
+        WikiApi::PageFetchError,
+        /Failed to fetch content for Ragesoss with response status: ./
```
empty-codes, line 26: That's true, the same behaviour happens here too: subject is `described_class.new.get_page_content('Ragesoss')`  when I added the else case, it failed with these errors so I changed them

empty-codes:
Update:
- The response.status was `"."` because of the full stop in the message string, instead of a `" "` empty string. This is because `response.status` is `nil` but when `nil` is interpolated into a string, it gets converted to an empty string. 
**Solution**: changed `#{status}.` to `#{status.inspect}`
- The `enrollment_controller_spec` was failing because it is a request spec and mocks the whole controller action so the `get_page_content` was mocked without a response making response.status `nil`. 
**Solution**: explicitly mocked the `get_page_content` method to return a valid response object with a status code (in this case `200`) instead of `nil`

ragesoss:
nice work

PR #6165, opened 2025-01-31
Fix: #6087 Display specific error message for course slug collision

### LINKED ISSUE #6087: Error message from course slug collision does not explain the problem
### What is happening?

ichandrasharma:
## What this PR does
Fixes #6087

This PR addresses the issue where a generic "Internal Server Error" message was displayed when a user attempted to change a course slug to one that already existed.

Key changes in code : 
Backend (Rails): Modified the Rails controller to send a JSON response containing a specific error message when a course slug collision occurs.  This message clarifies the reason for the error to the user.

Frontend: Updated the frontend code to correctly extract and display the specific error message received from the backend.  This replaces the generic "Internal Server Error" message with a user-friendly message indicating the slug collision.

## Screenshots
Before:
Course titles can be identical for different organizations or institutions, but must be unique within the same organization or institution.  This is because course titles are used to dynamically generate URLs (slugs), and duplicate titles within the same organization would create URL conflicts.
look: below
Screenshot (70)

and this is an error occurring not showing proper message only showing Internal Server Error
Screenshot (69)

After:
Showing proper message
Screenshot (71)

And should be a different title to avoid conflict in the URL
Screenshot (72)

## Learning
I took 3+ days to understand the codebase, and then I solved this

### DIFF
--- a/app/assets/javascripts/actions/course_actions.js
+++ b/app/assets/javascripts/actions/course_actions.js
@@ -1,30 +1,48 @@
 import {
-  ADD_NOTIFICATION, API_FAIL, UPDATE_COURSE, RECEIVE_COURSE, RECEIVE_COURSE_UPDATE,
-  PERSISTED_COURSE, DISMISS_SURVEY_NOTIFICATION, TOGGLE_EDITING_SYLLABUS,
-  START_SYLLABUS_UPLOAD, SYLLABUS_UPLOAD_SUCCESS, LINKED_TO_SALESFORCE, COURSE_SLUG_EXISTS, RECEIVE_COURSE_SEARCH_RESULTS, SORT_COURSE_SEARCH_RESULTS, FETCH_COURSE_SEARCH_RESULTS, RECEIVE_ACTIVE_COURSES, SORT_ACTIVE_COURSES,
-  RECEIVE_CAMPAIGN_ACTIVE_COURSES, RECEIVE_WIKI_COURSES, SORT_WIKI_COURSES
-} from '../constants/index';
-import API from '../utils/api.js';
-import CourseUtils from '../utils/course_utils';
-import request from '../utils/request';
-import { toWikiDomain } from '../utils/wiki_utils';
+  ADD_NOTIFICATION,
+  API_FAIL,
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/actions/course_actions.js
```diff
@@ -1,30 +1,48 @@
...
-  ADD_NOTIFICATION, API_FAIL, UPDATE_COURSE, RECEIVE_COURSE, RECEIVE_COURSE_UPDATE,
-  PERSISTED_COURSE, DISMISS_SURVEY_NOTIFICATION, TOGGLE_EDITING_SYLLABUS,
-  START_SYLLABUS_UPLOAD, SYLLABUS_UPLOAD_SUCCESS, LINKED_TO_SALESFORCE, COURSE_SLUG_EXISTS, RECEIVE_COURSE_SEARCH_RESULTS, SORT_COURSE_SEARCH_RESULTS, FETCH_COURSE_SEARCH_RESULTS, RECEIVE_ACTIVE_COURSES, SORT_ACTIVE_COURSES,
-  RECEIVE_CAMPAIGN_ACTIVE_COURSES, RECEIVE_WIKI_COURSES, SORT_WIKI_COURSES
-} from '../constants/index';
-import API from '../utils/api.js';
-import CourseUtils from '../utils/course_utils';
-import request from '../utils/request';
-import { toWikiDomain } from '../utils/wiki_utils';
+  ADD_NOTIFICATION,
```
ragesoss, line 2: Please keep functional changes separate from formatting changes. I don't mind changes that apply this sort of formatting change, but they should be done in a separate PR with a clear description of how they were done (eg, automatically via a linter in VSCode, or whatever). It's much harder to think through and review functional changes when they are mixed in with major formatting changes.

ragesoss:
Thanks! For the error message, ideally it should display just the message itself, without the json formatting.

app/assets/javascripts/actions/course_actions.js
ichandrasharma: ok i will off my automatically linter then I change a pr then and anything should be done

PR #6171, opened 2025-02-03
`DuplicateArticleDeleter`: fix existing bug to keep the most recently created article instead of the oldest one

gabina:
## What this PR does
This PR modifies the article deletion criteria in the `DuplicateArticleDeleter` class to be consistent with the comment:
```
  # Delete all articles with the given title
  # and namespace except for the most recently created
```
 
According to [docs](https://apidock.com/rails/ActiveRecord/QueryMethods/order), the default ordering for ActiveRecord is ascending, meaning that  `keeper = articles.first` ends up with the oldest article in `keeper`. This PR changes that to keep the last one.

Note that there were a lot of changes in this class (and its specs) along the time so I don't know how this should really work.
At some point we kept the last article instead of the first one (see commit 8ea90c5 where that was changes). In addition, we also forced the order to be descending (see commit ba8ff9c).

I'm changing the behavior to be consistent with the comment.

I need to replicate the behavior of `DuplicateArticleDeleter` for timeslices, so I want to clarify the exact criteria we should follow.

### DIFF
--- a/lib/duplicate_article_deleter.rb
+++ b/lib/duplicate_article_deleter.rb
@@ -49,7 +49,8 @@ class DuplicateArticleDeleter
   def delete_duplicates(title, ns)
     articles = Article.where(title:, namespace: ns, wiki_id: @wiki.id, deleted: false)
                       .order(:created_at)
-    keeper = articles.first
+    # Default order is ascendent, so we want to keep the last article
+    keeper = articles.last
     return [] if keeper.nil?
 
     # Here we must verify that the titles match, since searching by title is case-insensitive.
--- a/spec/lib/duplicate_article_deleter_spec.rb
+++ b/spec/lib/duplicate_article_deleter_spec.rb
@@ -31,26 +31,26 @@ describe DuplicateArticleDeleter do
                        deleted: true)
     end
 
-    it 'marks one deleted when there are two ids for one page' do
-      first = create(:article,
-                     id: 2262715,
-                     title: 'Kostanay',
-                     namespace: 0,
-                     created_at: 1.day.from_now)
-      second = create(:article,
-                      id: 46349871,
-                      title: 'Kostanay',
-                      namespace: 0,
-                      created_at: 1.day.ago)
-      described_class.new.resolve_duplicates([first])
-      undeleted = Article.where(
+    it 'marks oldest one deleted when there are two ids for one page' do
+      new_article = create(:article,
+                           id: 2262715,
+                           title: 'Kostanay',
+                           namespace: 0,
+                           created_at: 1.day.from_now)
+      duplicate_article = create(:article,
+                                 id: 46349871,
+                                 title: 'Kostanay',
+                                 namespace: 0,
+                                 created_at: 1.day.ago)
+      described_class.new.resolve_duplicates([new_article])
+      deleted = Article.where(
         title: 'Kostanay',
         namespace: 0,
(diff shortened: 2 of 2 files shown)


spec/lib/duplicate_article_deleter_spec.rb
```diff
@@ -31,26 +31,26 @@
...
-      first = create(:article,
-                     id: 2262715,
-                     title: 'Kostanay',
-                     namespace: 0,
-                     created_at: 1.day.from_now)
-      second = create(:article,
-                      id: 46349871,
-                      title: 'Kostanay',
-                      namespace: 0,
-                      created_at: 1.day.ago)
```
gabina, line 44: Note also that specs were passing because they were inconsistent with the comment: article called `first` was created tomorrow, while article called `second` was created yesterday. The "undeleted" was `second`, meaning that the oldest one was kept.

lib/duplicate_article_deleter.rb
```diff
@@ -49,7 +49,8 @@ def delete_duplicates_in(article_group)
   def delete_duplicates(title, ns)
     articles = Article.where(title:, namespace: ns, wiki_id: @wiki.id, deleted: false)
                       .order(:created_at)
-    keeper = articles.first
+    # Default order is ascendent, so we want to keep the last article
+    keeper = articles.last
```
gabina, line 53: According to [docs](https://apidock.com/rails/ActiveRecord/QueryMethods/order), the default ordering for `ActiveRecord` is ascending, meaning that `keeper = articles.first` ends up with the oldest article in keeper. This PR changes that to keep the last one.

lib/duplicate_article_deleter.rb
```diff
@@ -49,7 +49,8 @@ def delete_duplicates_in(article_group)
...
     return [] if keeper.nil?
 
     # Here we must verify that the titles match, since searching by title is case-insensitive.
```
gabina, line 56: I understand that this is an effort to avoid deleting articles that aren’t true duplicates but appear as such due to the database being case-insensitive for the `title` field. However, this mechanism isn’t perfect—if there are two articles named 'SalamanderMan08', but keeper is 'salamanderMan08', duplicates won’t be resolved. I’m not sure how critical this is, but I can put some effort into improving it if needed.

gabina:
Failed examples are not related to these changes:

```
rspec ./spec/lib/alerts/high_quality_article_monitor_spec.rb:51 # HighQualityArticleMonitor.create_alerts_for_course_articles creates Alert records for edited Good articles
rspec ./spec/lib/alerts/high_quality_article_monitor_spec.rb:117 # HighQualityArticleMonitor.create_alerts_for_course_articles does create second Alert if the first alert is resolved and there are later assignments
```

gabina:
@ragesoss I would like to know your thoughts on this when you get the chance.

lib/duplicate_article_deleter.rb
ragesoss: One of the constraints of Wikipedia article titles and most other Wikimedia wikis — but not Wiktionary — is that titles always start with a capital letter. So at least on Wikipedia, we shouldn't expect to have an article named `salamanderMan08`. The edge case is definitely possible, but I don't think it needs to be fixed now.

ragesoss:
This change is fine with me. I can't think of any reason why it would particularly matter to keep the older or newer one.

PR #6172, opened 2025-02-03
Fix: issue with course_assignment returning an empty object 

Formasitchijoh:
## What this PR does
This PR provides a quick fix for the recent [CI build error](https://github.com/WikiEducationFoundation/WikiEduDashboard/actions/runs/13120619294/job/36605410577), where the HighQualityArticleAssignmentAlert count was not as expected. The issue was due to the following:

- The `create_alerts_from_page_titles` method in `HighQualityArticleMonitor` filters assignments based on the current active course.
- However, this was returning an empty object for `course_assignments` because the end date specified in the test for the course had already passed.
- As a result, no assignment alerts were created since there were no active courses.
## Fix
The end date in the test has been updated to be dynamic, ending 30 days from the current time. This ensures that the course remains active during the test.

 Before ScreenShot 

Screenshot 2025-02-03 at 21 31 54

After ScreenShot

Screenshot 2025-02-03 at 21 29 31

@gabina I wish to confirm if this is the expected behaviour of this test. @ragesoss  could you review this when you are chanced

### DIFF
--- a/spec/lib/alerts/high_quality_article_monitor_spec.rb
+++ b/spec/lib/alerts/high_quality_article_monitor_spec.rb
@@ -5,7 +5,7 @@ require "#{Rails.root}/lib/alerts/high_quality_article_monitor"
 
 describe HighQualityArticleMonitor do
   describe '.create_alerts_for_course_articles' do
-    let(:course) { create(:course, start: '2024-01-01', end: '2025-01-01') }
+    let(:course) { create(:course, start: '2024-01-01', end: 30.days.from_now) }
     let(:student) { create(:user, username: 'Leemyongpak', email: 'learn@school.edu') }
     let(:instructor) { create(:user, username: 'instructor', email: 'teach@school.edu') }
     let!(:courses_user) do
--- a/spec/lib/alerts/high_quality_article_monitor_spec.rb
+++ b/spec/lib/alerts/high_quality_article_monitor_spec.rb
@@ -5,7 +5,11 @@ require "#{Rails.root}/lib/alerts/high_quality_article_monitor"
 
 describe HighQualityArticleMonitor do
   describe '.create_alerts_for_course_articles' do
-    let(:course) { create(:course, start: '2024-01-01', end: 30.days.from_now) }
+    let(:course) do
+      travel_to Date.new(2025, 1, 10) do
+        create(:course, start: '2024-01-01', end: Date.new(2025, 1, 10))
+      end
+    end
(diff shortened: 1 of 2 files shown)


gabina:
uups, sorry! It looks like I introduced a time bomb in those specs. I think the easiest way to fix them is to add something like `travel_to Date.new(2025, 1, 10)`  to ensure the course is always considered current, while keeping the fixed dates. I believe the dates were chosen specifically to get some particular API responses. We may need to do something similar for `DiscretionarySanctionsMonitor`.

Formasitchijoh:
Alright, I will update that, do you think I should update the `DiscretionarySactionsMonitor` in this same PR  since they are related or I should do it in a separate PR

gabina:
> do you think I should update the DiscretionarySactionsMonitor in this same PR since they are related or I should do it in a separate PR

Same PR is fine, as you mentioned changes are related

spec/lib/alerts/high_quality_article_monitor_spec.rb
```diff
@@ -5,7 +5,11 @@
 
 describe HighQualityArticleMonitor do
   describe '.create_alerts_for_course_articles' do
-    let(:course) { create(:course, start: '2024-01-01', end: '2025-01-01') }
+    let(:course) do
+      travel_to Date.new(2025, 1, 10) do
+        create(:course, start: '2024-01-01', end: Date.new(2025, 1, 10))
```
gabina, line 10: hmm this changes the end date from 2025-01-01 to 2025-01-10, but the specs are green so I'm good with that.

ragesoss:
nice work @Formasitchijoh

gabina:
It looks like the `HighQualityArticleMonitor` specs are failing again. I suspect that `travel_to` isn’t actually being applied when `Course.current` is calculated. The tests may have only passed because the course’s end date was updated from 2025-01-01 to 2025-01-10, making it "current" again.

I'm not familiar with the syntax used here, but I guess that the time travel is only applied when the course is created. 
```
let(:course) do
  travel_to Date.new(2025, 1, 10) do
    create(:course, start: '2024-01-01', end: Date.new(2025, 1, 10))
  end
end
```
A better fix would be to move `travel_to Date.new(2025, 1, 10)` into the `before` block, ensuring that all code runs as if it's 2025-01-10.

ragesoss:
@Formasitchijoh ^

Formasitchijoh:
@gabina I will take a look at this, didn't sorry didn't see this earlier

PR #6174, opened 2025-02-04
[WIP]: Caching slow test on github CI pipeline

Formasitchijoh:
## What this PR does
This PR is a work in progress to reduce the build time on the CI pipeline by caching the response for slow test and running any other test when the test file changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,26 @@ jobs:
               sudo ln -s $(which magick) /usr/bin/$cmd
             fi
           done
+      
+      # Cache for Slow Tests (cached folder)
+      - name: Cache VCR cassettes (Slow Tests)
+        uses: actions/cache@v3
+        with:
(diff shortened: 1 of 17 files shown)


ragesoss:
This seems promising. Is it possible to test how much it speeds things up, without merging it? If we use this approach, we'll also need some documentation of the caching strategy and instructions for breaking the cache.

Formasitchijoh:
> This seems promising. Is it possible to test how much it speeds things up, without merging it? If we use this approach, we'll also need some documentation of the caching strategy and instructions for breaking the cache.

@ragesoss  Thanks for the feedback! Just to clarify, are you asking if I can measure the speed improvement by running the CI pipeline on my branch before merging? My plan is to:

- Run the CI on my branch and compare the total build time before and after caching.
- Track how long the slow test takes with and without caching.
- Run the pipeline multiple times to see if the caching consistently speeds things up.

ragesoss:
Right!

PR #6184, opened 2025-02-07
Fix: Include require for histogram_plotter.rb in spec file

Formasitchijoh:
## What this PR does
This PR adds a require statement for histogram_plotter.rb in the relevant spec file to ensure all dependencies are properly loaded during test execution. This prevents potential NameError or missing constant issues when running the test suite.

Changes:
Added `require "#{Rails.root}/lib/analytics/histogram_plotter.rb"` in the spec file

Don't know how this worked on CI , but it failed locally which is expected

### DIFF
--- a/spec/lib/analytics/histogram_plotter_spec.rb
+++ b/spec/lib/analytics/histogram_plotter_spec.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'rails_helper'
+require "#{Rails.root}/lib/analytics/histogram_plotter.rb"
 
 describe HistogramPlotter do
   let(:course) do


ragesoss:
My guess for why this failed locally but passed on CI is that it is gets loaded via indirect requirements by other spec files, so in CI (where the specs run in a deterministic order) it is always loaded before this spec gets run, but it may fail when it runs on its own locally.

We used to have a lot of cases where that sort of thing happened, and we fixed most of them but this must have slipped through.

Formasitchijoh:
Wow it is nice how the CI handle loading those requirements before running the test. and Yeah i am sure this one was missed

PR #6189, opened 2025-02-10
[Data rearchitecture] fix bug when creating timeslices

gabina:
## What this PR does
This PR fixes a bug when creating timeslices due to a change in the start date. It worked fine for daily timeslices, but failed for 10-day timeslices. Now it should work fine no matter the tiemslice duration.

In addition, it stubs the `TIMESLICE_DURATION` env variable to be one day (86400 seconds) in several specs so they work no matter how the envvar is set.

### DIFF
--- a/app/services/update_timeslices_course_date.rb
+++ b/app/services/update_timeslices_course_date.rb
@@ -29,6 +29,9 @@ class UpdateTimeslicesCourseDate
 
     return unless drastic_change
 
+    Rails.logger.info "UpdateTimeslicesCourseDate: Course: #{@course.slug}\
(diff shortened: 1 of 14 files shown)


gabina:
Failing specs not related to these changes
```
Failed examples:

rspec ./spec/lib/alerts/high_quality_article_monitor_spec.rb:55 # HighQualityArticleMonitor.create_alerts_for_course_articles creates Alert records for edited Good articles
rspec ./spec/lib/alerts/high_quality_article_monitor_spec.rb:121 # HighQualityArticleMonitor.create_alerts_for_course_articles does create second Alert if the first alert is resolved and there are later assignments
```

PR #6190, opened 2025-02-10
Fix: Issue with high_quality_article_monitor_spec

Formasitchijoh:
## What this PR does
This PR addresses the recent [CI build failure](https://github.com/WikiEducationFoundation/WikiEduDashboard/actions/runs/13120619294/job/36605410577) and improves upon the previous [fix](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6172), which was inefficient.

### Issue with the Previous Fix:
- The **`travel_to`** helper method was only applied when creating the course, meaning that subsequent interactions in the test used the **current system time** instead of the **intended past time**. This caused failures when the test logic compared dates, as the current time was beyond the course end date. 

- The previous fix also relied on` Date.new`, which does not account for time zones, making it ineffective for this scenario.

### Improvements in This Fix:

- Replaced Date.new with Time.zone.local  which is Rails-specific and handle time zones .
- Moved `travel_to` into the before block to ensure that all test interactions occur within the correct time frame.
- Ensured proper reset to the system time after the test completes using  `travel_back`   method .

Screenshot 2025-02-10 at 21 24 06

This fix travels to a time between the **start and end date** to  ensures that all test blocks operate within the intended date context, leading to more reliable and consistent test execution.

A detail explanation of how these helpers work is found [here](https://api.rubyonrails.org/v5.2.1/classes/ActiveSupport/Testing/TimeHelpers.html)

@gabina , your assumption about how travel_to was used previously is correct. Thank you for your input

### DIFF
--- a/spec/lib/alerts/discretionary_sanctions_monitor_spec.rb
+++ b/spec/lib/alerts/discretionary_sanctions_monitor_spec.rb
@@ -9,11 +9,8 @@ end
 
 describe DiscretionarySanctionsMonitor do
   describe '.create_alerts_for_course_articles' do
-    let(:course) do
-      travel_to Date.new(2025, 1, 20) do
-        create(:course, start: '2024-12-10', end: Date.new(2025, 1, 20))
-      end
-    end
+    let(:course) { create(:course, start: '2024-12-10', end: '2025-01-20') }
+
     let(:student) { create(:user, username: 'Gelasin') }
     let!(:courses_user) do
       create(:courses_user, user_id: student.id,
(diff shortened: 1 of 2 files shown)


spec/lib/alerts/discretionary_sanctions_monitor_spec.rb
```diff
@@ -43,13 +40,18 @@ def mock_mailer
...
       allow_any_instance_of(CategoryImporter).to receive(:page_titles_for_category)
         .with('Category:Wikipedia pages about contentious topics', 1)
         .and_return(['Talk:1948 war',
                      'Talk:Ahmed Mohamed clock incident',
                      'Talk:Armenian Genocide denial'])
     end
 
+    after do
+      travel_back
+    end
```
gabina, line 53: Is this really necessary? Or is it just a good practice?

According to the `travel_to` docs:

> Changes current time to the given time by stubbing `Time.now`, `Time.new`, `Date.today`, and `DateTime.now` to return the time or date passed into this method. The stubs are automatically removed at the end of the test.

So I understand that it is not strictly necessary but perhaps it is considered a good practice?

Formasitchijoh: Yeah you are right, just wanted to make sure the time is reset back

PR #6193, opened 2025-02-11
CI: Add GitHub Action for caching and restoring VCR cassettes

Formasitchijoh:
## What this PR does
This PR introduces a GitHub Action to cache and restore VCR cassettes, improving test performance by avoiding redundant external requests.

### Changes Made:
- Added a GitHub Action step to cache fixtures/vcr_cassettes.
- Configured the cache with a key (vcr-cached-v1) and a restore key (vcr-cached-) to reuse existing caches when available.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,9 +86,17 @@ jobs:
         with:
           path: node_modules/.cache/
           key: ${{ runner.os }}-webpack
+
+      # Create cache or restore cache for Tests
+      - name: Cache VCR cassettes
+        uses: actions/cache@v3
+        with:
+          path: fixtures/vcr_cassettes
+          key: vcr-cached-v1
+          restore-keys: |
+            vcr-cached-
           
       #Testing the pushed code onwards
-      
       - name: JavaScript test suite 
         run: |
             yarn install --immutable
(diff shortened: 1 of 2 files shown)


Formasitchijoh:
@ragesoss I think this is good for review, I have researched and optimized it better from the previous implementation

Formasitchijoh:
The second commit is a fix to an issue I encountered when I ran the test locally, because the text was returned in an HTML tag, just added it here to ensure the build time actually reduced after the initial cache

ragesoss:
Nicely done, thanks!

gabina:
Hello! CI is failing for my PR #6417. The spec doesn't fail locally. I suspect the problem is that the request failed transiently when saving VCR files. I understand they're cached. Is it OK to go to github actions -> [caches](https://github.com/WikiEducationFoundation/WikiEduDashboard/actions/caches) and delete `vcr-cached-v1 ` that was cached 5 months ago? My hope is that by doing that, the requests will actually run again for the specs and this time there will be no transient errors, and my specs will pass.

```
Failures:

  1) UpdateCourseWikiTimeslices when a full update is required sets needs_update to false if update is successful
     Failure/Error: expect(course.course_wiki_timeslices.where(needs_update: true).count).to eq(0)

       expected: 0
            got: 2

       (compared using ==)
     # ./spec/services/update_course_wiki_timeslices_spec.rb:290:in `block (3 levels) in <top (required)>'
```

ragesoss:
@gabina it should be fine to delete the caches at any time; it ought to just mean a slower next CI run when the cache is missing. Do you have access to do that?

gabina:
Perfect, thanks! I wanted to check just in case. For the future, it seems that I do have access to delete the caches, I guess because I am a "member".

PR #6195, opened 2025-02-12
Data rearchitecture for dashboard

ragesoss:


app/controllers/courses_controller.rb
```diff
@@ -235,6 +235,13 @@ def manual_update
     redirect_to "/courses/#{@course.slug}"
   end
 
+  def manual_update_timeslice
```
ragesoss, line 238: I guess this was added just to make it easy to compare updates via the old method versus the new one?

During the cleanup phase, we'll want to remove this method and make `manual_update` use the timeslice approach.

app/models/course_types/article_scoped_program.rb
```diff
@@ -76,4 +76,26 @@ def multiple_roles_allowed?
   def passcode_required?
     false
   end
+
+  def filter_revisions(revisions)
+    filtered_data = revisions.select do |_, details|
+      wiki = Wiki.find(details['article']['wiki_id'])
```
ragesoss, line 82: This seems like it might do a lot of duplicate Wiki queries. It looks like this is only ever called for one wiki at a time, via RevisionDataManager, so maybe the wiki can just be passed in as an argument?

app/models/course_types/visiting_scholarship.rb
```diff
@@ -75,4 +75,22 @@ def use_start_and_end_times
   def multiple_roles_allowed?
     true
   end
+
+  def filter_revisions(revisions)
```
ragesoss, line 79: I would prefer not duplicate this method across multiple course types. I guess that the different implementations of `scoped_articles_titles` is the only difference, so perhaps this method can be extracted somewhere else?

lib/revision_data_manager.rb
```diff
@@ -0,0 +1,177 @@
...
+          mw_rev_id: rev_data['mw_rev_id'],
+          date: rev_data['date'],
+          characters: rev_data['characters'],
+          article_id: articles.nil? ? nil : articles[mw_page_id],
+          mw_page_id:,
+          user_id: users[rev_data['username']],
+          new_article: string_to_boolean(rev_data['new_article']),
+          system: string_to_boolean(rev_data['system']),
+          wiki_id: rev_data['wiki_id'],
+          views: scoped_revision?(scoped_sub_data, mw_page_id, rev_data['mw_rev_id'])
```
ragesoss, line 161: Why is the `views` field used for this? If it's just a hack so that the Revision model can be used as the in-memory representation of a revision without saving it to the database, since we weren't using the `views` field anymore for its original purpose, we could put `attr_accessor :scoped` into `Revision`, and then `Revision.new(scoped: value)` will create a Revision object that with an accessible `scoped` value that just lives in memory.

lib/revision_stat_timeslice.rb
```diff
@@ -0,0 +1,52 @@
+# frozen_string_literal: true
+
+#= Provides a count of recent revisions by a user(s)
+class RevisionStatTimeslice
+  REVISION_TIMEFRAME = 7
```
ragesoss, line 5: This constant name should make it clear that the value is in days.

ragesoss (commented):
Just a few little things I spotted

app/controllers/courses_controller.rb
gabina: > I guess this was added just to make it easy to compare updates via the old method versus the new one?

Yes, exactly. There are two routes: `/manual_update` and `/manual_update_timeslice` to easily compare metrics.

app/models/course_types/article_scoped_program.rb
gabina: Yes, I think you're right.

lib/revision_stat_timeslice.rb
gabina: Done on 96ea93e2d5428f25bb77528195fd7f537cdf1c0e

app/models/course_types/article_scoped_program.rb
gabina: Done on 14357c06fa0809d6b9c16a6b3d819aa5285b77a1

lib/revision_data_manager.rb
gabina: Yes, it's a hack as you mentioned.

I used the `attr_accessor` strategy in [c9510a5](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6195/commits/c9510a50c8354a8f1b0c546ee08e1c2ee2db7901)

In addition, I followed the same approach for `error` field (I was using `ithenticate_id`) in [66f123c](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6195/commits/66f123c52c9ffa9d3d6e876e0d033042873a483f)

app/models/course_types/visiting_scholarship.rb
ragesoss: I agree that the parent class should avoid doing any query for `filter_revisions`. We could think of `scoped_article_titles` as a concept that is independent of course type, so that the method is available as a shared implementation on the parent class, but is only used by the course types that do filtering of edits based on scope. Alternatively, it could be implemented as a module that gets included in the course types that use it.

I don't feel strongly about where the shared code should live, I just think it's too complex to duplicate without risk of accidentally diverging later on.

gabina: I added a `CustomRevisionFilter` module in commit 55cb666f4556d3bf32d2ce59e256fa9ad9214959. I'm not sure it was the most elegant way to do it, but I couldn't think of any better ways and I didn't want to spend more time on it. It's already deployed and running in data-rearchitecture instance.

ragesoss: I looked at it; no objections from me.

app/controllers/courses_controller.rb
gabina: This was done in #6308

Issue #5689, opened 2024-02-29
Add option for making calendar / timeline weeks go Monday - Sunday instead of Sunday - Saturday


ragesoss:
We get periodic requests from instructors to have the dates work on a Monday through Sunday basis. I suspect that this would not be a welcome change for some people if done universally, so we'd probably need to add it as an option that can be switched back and forth for an individual course. It probably involves a lot of tweaks to the frontend date code, and a bit on the backend as well.

ragesoss:
Sure. This will likely need some very thorough tests.

shishiro26:
I  would like to work on this, @ragesoss.

ragesoss:
@shishiro26 i suspect that I lost track of the above PR because the build was failing and I forgot to check back, but you can probably use @PR4NJ41's work as a starting point. I suggest starting by carefully looking at the changes there until you understand them, and them re-implement them starting from the current code (to avoid the conflicts from that branch).

ragesoss:
How are you thinking about cloned courses in relation to this feature?

shishiro26:
When cloning courses, if a course originally runs from Monday to Sunday, the cloned course will follow the same schedule. I believe that, by default, some flags in the course are automatically set based on the original. Based on this, I will either keep the Monday-to-Sunday flag schedule or add a dropdown option to choose between Sunday-to-Saturday or Monday-to-Sunday.

ragesoss:
Ah. I think if a course has been modified to use the non-default week boundaries, then a cloned course should keep that same non-default setting.

shishiro26:
Yupp got it

PR #6196, opened 2025-02-13
add is_monday_start option to course creation and cloning.

shishiro26:

#5689

## What this PR does
This PR adds the `is_monday_start` flag to the course model, allowing the system to track whether a course follows a Monday–Sunday or a Sunday–Saturday schedule.

## Screenshots
Before:
https://github.com/user-attachments/assets/eb5b308f-9f44-424d-8c24-6d05ed832123
After:

course creation 
https://github.com/user-attachments/assets/c972ef0b-ba9b-4ab9-beb6-b8afea4e3a86

course cloning 

https://github.com/user-attachments/assets/d966ed36-bb5a-4a1f-bc05-4882749bcb97

### DIFF
--- a/app/assets/javascripts/components/course_creator/course_dates.jsx
+++ b/app/assets/javascripts/components/course_creator/course_dates.jsx
@@ -11,6 +11,12 @@ const CourseDates = (props) => {
     props.updateCourseProps(updatedCourse);
     setHasUpdatedDateProps(true);
   };
+
+  const handleCalendarStartDayChange = (e) => {
+    const value = e.target.value === '1';
+    props.updateCourseProps({ is_monday_start: value });
+    props.updateCourseAction('is_monday_start', value);
+  };
   const dateProps = CourseDateUtils.dateProps(props.course);
   const timeZoneMessage = (
     <p className="form-help-text">
@@ -34,6 +40,7 @@ const CourseDates = (props) => {
         id="course_timeline_start"
         onChange={updateCourseDates}
         value={props.course.timeline_start}
+        isMondayStart={props.course.is_monday_start}
         value_key="timeline_start"
         editable
         label={CourseUtils.i18n('creator.assignment_start', props.stringPrefix)}
@@ -51,6 +58,7 @@ const CourseDates = (props) => {
         id="course_timeline_end"
         onChange={updateCourseDates}
         value={props.course.timeline_end}
+        isMondayStart={props.course.is_monday_start}
         value_key="timeline_end"
         editable
         label={CourseUtils.i18n('creator.assignment_end', props.stringPrefix)}
@@ -64,11 +72,26 @@ const CourseDates = (props) => {
       />
     );
   }
+
+  const calendarDropdown = (
+    <div>
+      <p><strong>Select a Day for the calendar:</strong></p>
+      <select onChange={handleCalendarStartDayChange} value={props.course.is_monday_start ? '1' : '0'}>
+        <option value="none" disabled>Select Option</option>
+        <option value="0">Sunday - Saturday</option>
+        <option value="1">Monday - Sunday</option>
+      </select>
+    </div>
+  );
+
   return (
     <div className={props.courseDateClass}>
       <p>{CourseUtils.i18n('creator.course_dates_info', props.stringPrefix)}</p>
+      {calendarDropdown}
+      {/*  The key ensures the component re-renders when the is_monday_start value changes */}
       <DatePicker
         id="course_start"
+        key={`course_start_${props.course.is_monday_start}`}
         onChange={updateCourseDates}
         value={props.course.start}
         value_key="start"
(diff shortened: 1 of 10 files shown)


shishiro26:
@ragesoss please can you review this PR

shishiro26:
Hey @ragesoss, just following up on this PR. Let me know if you need any changes. Thanks!

ragesoss:
Thanks! It's still on my todo list, but I'm further behind than I'd like to be on some other projects. Hoping to get to your open PRs next week.

app/controllers/courses_controller.rb
```diff
@@ -418,6 +418,10 @@ def update_timeslice_duration
     @course.save
   end
 
+  def update_course_calendar_view
```
ragesoss, line 421: This is not a very precise name for a method that is exclusively concerned with the week-start flag.

app/assets/javascripts/components/course_creator/course_dates.jsx
```diff
@@ -64,11 +72,26 @@ const CourseDates = (props) => {
       />
     );
   }
+
+  const calendarDropdown = (
+    <div>
+      <p><strong>Select a Day for the calendar:</strong></p>
```
ragesoss, line 78: Use i18n for all interface messages.

ragesoss: I think maybe 'Weeks start on:' should be the message here, and the values should just be `Sunday` and `Monday`.

app/assets/javascripts/components/overview/course_cloned_modal.jsx
```diff
@@ -153,11 +153,14 @@ const CourseClonedModal = createReactClass({
     return true;
   },
 
+
   render() {
     const i18nPrefix = this.props.course.string_prefix;
     let buttonClass = 'button dark';
     buttonClass += this.state.isPersisting ? ' working' : '';
 
+    console.log('I am in this ', this.props.course);
```
ragesoss, line 162: Remove

ragesoss (commented):
The code here looks reasonable. I definitely want tests to be included as well. I was expecting this to require some backend changes related to how dates are handled, like due dates for training modules in BlockDateManager. Did you check the behavior of that to make sure it doesn't require any changes?

app/assets/javascripts/components/course_creator/course_dates.jsx
shishiro26: For the translations, do I only need to update the en.yml file, or do I need to update all the other language files as well?

ragesoss: The en.yml file, and ideally the qqq.yml file as well. The latter is not a language, but is used for English descriptions of what the message is for and how it is used, so that translators have additional context.

shishiro26:
> The code here looks reasonable. I definitely want tests to be included as well. I was expecting this to require some backend changes related to how dates are handled, like due dates for training modules in BlockDateManager. Did you check the behavior of that to make sure it doesn't require any changes?

When you mention tests, are you referring to the frontend select button tests? I haven’t looked into `BlockDateManager` yet, but I will check it out and work on migrating the handling of weekday dates to the backend."

ragesoss:
Yes, I think at minimum this should include a feature spec that uses the feature from an instructor perspective.

shishiro26:
I tried writing test cases for this, but I'm not sure why they keep failing with the error **"Course already exists."** Even after changing the course name, term, and other details, the issue persists. No matter what I try, it doesn't seem to work. Could you help me with this?  

image

ragesoss:
@shishiro26 can you post the test code that fails with the screenshot you posted?

shishiro26:
I initially added some of my code, but it failed. Later, I removed it and tested again, but it still failed.

ragesoss:
@shishiro26 is this happening for the test case you wrote, or is that happening on another unrelated test? Without seeing what code was running for the test, it's hard to guess what might be wrong.

shishiro26:
I haven't written any new test code. I just ran the existing code, and it started failing in my local environment right away.

ragesoss:
Which test?

ragesoss:
Can you post the stack traces from the failures? That will show exactly which line of the test it failed at.

shishiro26:
https://gist.github.com/shishiro26/fcb6ea1c8876d2211e997b290bb90a88

ragesoss:
Ah, so it looks like there are javascript errors  causing the problem.  Try running `yarn build` to use the production version of the assets and see if the errors change.

shishiro26:
yupp thank you ragesoss it worked now

shishiro26:
@ragesoss I'm not sure why, but the `is_monday_start` flag is not being stored in the SQL database. It worked when I tested it, but now it doesn't seem to be working.

PR #6199, opened 2025-02-17
[Data rearchitecture] Address sage reviews

gabina:
## What this PR does
This PR addresses Sage's comments in PR #6195.

In addition, it adds `error` revision field as `attr_accessor :error` instead of using `ithenticate_id` field. See PR #6145 for tracking errors when getting scores.

### DIFF
--- a/lib/importers/revision_score_importer.rb
+++ b/lib/importers/revision_score_importer.rb
@@ -190,16 +190,13 @@ class RevisionScoreImporter
     rev.features = rev_scores['features']
     rev.wp10 = rev_scores['wp10']
     rev.deleted = rev_scores['deleted'] # double check if this is a boolean
-    # TODO: use another field for this
-    rev.ithenticate_id = 1 if rev_scores['error'] # tocuh ithenticate_id if there was an error
+    rev.error = rev_scores['error']
   end
 
   def update_previous_scores(rev, parent_rev_scores)
     rev.wp10_previous = parent_rev_scores['wp10']
     rev.features_previous = parent_rev_scores['features']
-    # TODO: use another field for this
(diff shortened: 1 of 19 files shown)


Formasitchijoh:
@gabina seems you haven't updated your branch with my latest fix for the error, possibly the reason you're still having the same error

lib/importers/revision_score_importer.rb
```diff
@@ -190,16 +190,13 @@ def update_scores(rev, rev_scores)
...
+    rev.error = rev_scores['error']
   end
 
   def update_previous_scores(rev, parent_rev_scores)
     rev.wp10_previous = parent_rev_scores['wp10']
     rev.features_previous = parent_rev_scores['features']
-    # TODO: use another field for this
-    # tocuh ithenticate_id if there was an error
-    rev.ithenticate_id = 1 if parent_rev_scores['error']
+    rev.error = parent_rev_scores['error']
```
ragesoss, line 199: If the `update_scores` had an error, would this would overwrite it, including overwriting it with `nil` if the `parent_rev_scores` does not have an error?

gabina: hmm yeah, nice find, looks like I introduced a bug with this change, and the specs weren't good enough to catch it. I will open a workaround for this, should be quick

gabina: Fixed in commit [d0fd09f](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6195/commits/d0fd09f6a54f71f138fba06d38645d4bf36d89f0) (trying to minimize code changes). I'm deploying to data-rearchitecture instance now so it stays running all night before deployment to production.

PR #6201, opened 2025-02-18
LTIAAS Integration

Cvmcosta:
## What this PR does
This PR implements an integration with the LTIAAS API. It lays the groundwork for building a full LTI integration.

Additional work is required in the form of custom pages to where we can redirect the user, as well as error handling for the custom exceptions defined in this PR


app/models/lti_context.rb
```diff
@@ -0,0 +1,16 @@
...
+# Table name: lti_contexts
+#
+#  id                 :integer          not null, primary key
+#  user_id            :integer          not null, foreign_key
+#  user_lti_id        :string(255)      not null
+#  context_id         :string(255)      not null
+#  lms_id             :string(255)      not null
+#  lms_family         :string(255)
+#
+
```
ragesoss, line 13: Comments defining the last four attributes here would be helpful, including how they translate to fields in a Canvas instance.

ragesoss:
Code-wise, this looks great.

I think this will be safe to merge and deploy without doing any configuration (in which case it might throw benign errors if traffic hits the `lti` route, but there won't be any such traffic anyway). Is that right from your perspective?

PR #6208, opened 2025-02-24
fix: mark survey conditional questional required or not

### LINKED ISSUE #935: cannot make a conditional question required

JiyaGupta-cs:
## Description
When creating a survey with conditional questions, it was unable to mark these questions as required or not.
Even when the "Question is Required" checkbox was selected in the form, conditional questions were always saved with the ```presence``` validation rule set to "0". This was happening because of hardcoded logic in the ```to_question_params``` method that forced conditional questions to be optional.

Fixes #935

Based on the logs, when submitted the form with marked conditional question as required, means setting answer_presence to "1" , i.e., question should be required, but in the actual SQL INSERT statement, the presence value is being saved as "0", i.e., not required.
```
Parameters: {"question"=>{"type"=>"Rapidfire::Questions::Checkbox", "answer_presence"=>"1", ...}
```
```
INSERT INTO `rapidfire_questions` (...) VALUES (..., '---\n:presence: \'0\'\n:grouped: \'0\'\n...
```

## Fix
Modified the ```to_question_params``` method to respect the user's choice for required fields, even when the question has conditionals.
This change allows survey creators to mark conditional questions as required, giving them more flexibility in designing survey logic.

## Screenshots

After:

image

### DIFF
--- a/config/initializers/surveys.rb
+++ b/config/initializers/surveys.rb
@@ -234,7 +234,7 @@ Rails.application.config.to_prepare do
         :placeholder_text => placeholder_text,
         :track_sentiment => track_sentiment,
         :validation_rules => {
-          :presence => !conditionals.nil? && !conditionals.empty? ? "0" : answer_presence,
+          :presence => answer_presence,
           :grouped => answer_grouped,
           :grouped_question => answer_grouped_question,
           :minimum  => answer_minimum_length,


ragesoss:
Thanks! Did you verify that a required conditional question does not break the survey if the condition for showing that question is not met?

JiyaGupta-cs:
[Screencast from 2025-02-25 01-00-41.webm](https://github.com/user-attachments/assets/cca48de2-fcf8-4a58-9029-3d3f1f8ad235)

Yup, I tried creating questions in all of the ways, conditional required, conditional not required, unconditional required, unconditional not required

JiyaGupta-cs:
And in addition to it, apart from this issue of marking the unconditional question required, I also noticed that
if the user selected A then he sees the conditional question matching to case A, but if he returns back and mark B in place of A, then he sees the conditional question mapped to B as well as A. Ideally only B should have been visible. 
Need to open an issue for it.

PR #6210, opened 2025-02-24
Fixes `414 request-uri too long` for `import_articles_by_title` by using a POST request .2

empty-codes:
## What this PR does
Fixes `414 request-uri too long` for `import_articles_by_title` by using a POST request to send the query parameters in the `request.body` instead of a GET request:
- Adds `kn.wikipedia.org` to list of stubbed wiki validations
- Adds 3 API post methods, `api_post`, `do_post` and `too_many_requests` for creating and sending the post requests
- Refactors `import_articles_by_title` since response is now a hash
- Adds a test that confirms `414 request-uri too long` no longer occurs and a new vcr-cassette `some_article_titles.yml`

Adds `TYPICAL_ERRORS` array and safe navigation operator `&.` to `results = response&.dig('query', 'pages')` to prevent:
- `NameError: uninitialized constant ArticleImporter::TYPICAL_ERRORS` when `ApiErrorHandling` sets the error level
- `NoMethodError: undefined method dig for nil:NilClass results = response.dig('query', 'pages')` when response is nil ( for example when methods that call `import_articles_by_title` are mocked, the api_post response becomes `nil`)

The idea of `api_post` and `do_post` was gotten from a similar implementation in [replica.rb](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/60ffa77191a16d2e3986ca12cc72938d9bb63c91/lib/replica.rb#L140) and I tried to follow the flow of wiki_api.rb too.

## Screenshots
Before:
image

After:
image
image

### DIFF
--- a/lib/importers/article_importer.rb
+++ b/lib/importers/article_importer.rb
@@ -34,7 +34,7 @@ class ArticleImporter
         prop: 'info',
         titles: some_article_titles.join('|')
       }
-      response = api_post(@wiki, params)
+      response = api_post(@wiki.api_url, params)
       results = response.dig('query', 'pages')
       next if results.blank?
       import_articles_from_title_query(results)
@@ -71,9 +71,9 @@ class ArticleImporter
   # API methods #
   ###############
 
-  def api_post(wiki, params)
+  def api_post(api_url, params)
     tries ||= 3
-    response = do_post(wiki, params)
+    response = do_post(api_url, params)
     return unless response
     parsed_body = JSON.parse(response.body)
     return if parsed_body.empty?
@@ -83,12 +83,12 @@ class ArticleImporter
     sleep 1 if too_many_requests?(e)
     retry unless tries.zero?
     log_error(e, update_service: @update_service,
-              sentry_extra: { params:, api_url: @api_url })
+              sentry_extra: { params:, api_url: })
     return nil
   end
 
-  def do_post(wiki, params)
-    uri = URI(wiki.api_url)
+  def do_post(api_url, params)
+    uri = URI(api_url)
     https = Net::HTTP.new(uri.host, uri.port)
     https.use_ssl = true
 
--- a/lib/importers/article_importer.rb
+++ b/lib/importers/article_importer.rb
@@ -1,9 +1,12 @@
 # frozen_string_literal: true
 
 require_dependency "#{Rails.root}/lib/replica"
+require_dependency "#{Rails.root}/lib/errors/api_error_handling"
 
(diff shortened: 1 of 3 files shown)


lib/importers/article_importer.rb
```diff
@@ -60,4 +66,44 @@ def import_articles_from_title_query(results)
     end
     Article.import articles, on_duplicate_key_update: [:title, :namespace]
   end
+
+  ###############
+  # API methods #
```
ragesoss, line 71: This is all about interacting with the MediaWiki API, independently of ArticleImporter, so it probably should not be part of this class. It should be either in WikiApi or in a new WikiPostApi.

lib/importers/article_importer.rb
```diff
@@ -60,4 +66,44 @@ def import_articles_from_title_query(results)
...
+    sleep 1 if too_many_requests?(e)
+    retry unless tries.zero?
+    log_error(e, update_service: @update_service,
+              sentry_extra: { params:, api_url: })
+    return nil
+  end
+
+  def do_post(api_url, params)
+    uri = URI(api_url)
+    https = Net::HTTP.new(uri.host, uri.port)
```
ragesoss, line 92: Did you test whether you can do a post request with the mediawiki_api gem?

It is not well-documented, but it looks like it will support it as a parameter: https://github.com/wikimedia/mediawiki-ruby-api/blob/master/lib/mediawiki_api/client.rb#L157

I think that means you can add `http_method: :post` as a query parameter, and it will make it as a POST request instead of a GET. If that works, maybe just adding the extra parameter will make it work without re-implementing any API methods.

spec/lib/importers/article_importer_spec.rb
```diff
@@ -67,5 +68,50 @@
         expect(Article.find_by(title: 'Istanbul').mw_page_id).to eq(3391396)
       end
     end
+
+    it 'returns the correct number of titles without raising error 414' do
```
ragesoss, line 72: This looks like a great test. It should include a comment explaining in a little more detail what the test is doing and why the list of titles is what it is.

lib/importers/article_importer.rb
empty-codes: That was the first way I tried and it did not work so I used this way instead. However, your comment prompted me to retry and turns out I simply implemented it wrongly the first time. I was able to get it to work by simply adding the extra parameter (although I had to create a separate `meta` method in `wiki_api` since it kept error-ing). I created a new PR to that effect: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6212

PR #6211, opened 2025-02-24
fix style/optionalBooleanParam cop 

manbhav234:

## What this PR does
This PR attempts to make progress on the issue #5297 by removing the Style/optionalBooleanParameters cop from `.rubocop.yml` and fix the offenses generated by it. This style changes the definition of default boolean parameters inside function definitions from `func_name(bool_val = true`) to `func_name(bool_val: true)`.

## Screenshots
No UI changes made.

I have only commit changes to function definitions. Even though this fixes all the offenses generated by removing this cop, it's better to add such definition to function calls, i.e, `func_name(true)` to `func_name(bool_val: true)` so that it is clear for what variable the boolean value is being passed for. However, adding this would require to scan through the codebase and find the exact function calls for those functions which is a little tedious especially for the "email" function as there are many email functions defined in different classes. What are your thoughts on this ?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -171,8 +171,6 @@ Style/DateTime:
   Enabled: false
 Rails/ContentTag:
   Enabled: false
-Style/OptionalBooleanParameter:
-  Enabled: false
 Style/ExplicitBlockArgument:
   Enabled: false
 Style/KeywordParametersOrder:
--- a/app/controllers/surveys_controller.rb
+++ b/app/controllers/surveys_controller.rb
@@ -186,7 +186,7 @@ class SurveysController < ApplicationController
     return false
   end
 
-  def user_is_assigned_to_survey(return_notification = false)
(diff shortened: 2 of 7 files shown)


ragesoss:
The email functions are, for the most part, using the Rails mailer conventions, so I'm not too worried about those ones. Meeting the requirements of that linting rule is good enough for me, I think.

ragesoss:
The changes look fine to me. If the build passes, this should be ready to merge.

ragesoss:
Looks like this broke a lot of things.

manbhav234:
> Looks like this broke a lot of things.

Yeah the issue is the functions now expects keyword arguments instead of positional arguments. So just passing `func(true)` in place of `func(param_value: true)` causes errors which says '1 argument given (positional) but 0 required'. Shall I then try to fix these errors or something else needs to be done regarding this rule ?

PR #6212, opened 2025-02-25
Fix: Prevents `414 request-uri too long` for `import_articles_by_title` by using a POST request

empty-codes:
## What this PR does
Fixes 414 request-uri too long for import_articles_by_title by using a POST request instead of a GET request:
- Adds kn.wikipedia.org to list of stubbed wiki validations
- Adds `meta` method in wiki_api to prevent:
```
      Failure/Error:
       def query(query_parameters = {}, http_method: :get)
         mediawiki('query', query_parameters.merge(http_method: http_method))

     ArgumentError:
       unknown keyword: :meta
     # ./lib/wiki_api.rb:23:in `query'
     # ./app/models/wiki.rb:188:in `ensure_wiki_exists'
```

The former code  ` site_info = WikiApi.new(self).query(meta: :siteinfo)` in `wiki.rb`'s[ `ensure_wiki_exists` method](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/d0d5411c34cebc2c8de858e354e6229a1717458c/app/models/wiki.rb#L188) worked because the `query` method is a general entry point for all API requests and the `mediawiki` gem automatically processes `meta` requests since it is a valid method. My change to wiki_api's query method however bypasses the gem's automatic routing of `meta: :siteinfo` causing the error. 

The solution was implementing a `meta` method in the  `WikiApi` class that directly calls the [`meta` method](https://github.com/wikimedia/mediawiki-ruby-api/blob/84a1c3598524b048ab2b0cadf8ddbdbb07999114/lib/mediawiki_api/client.rb#L145) from the `mediawiki-api` gem.

- Adds a test that confirms `414 request-uri too long no longer occurs` and a new vcr-cassette `some_article_titles.yml`

## Screenshots
Before:
image

After:
image
image

### DIFF
--- a/app/models/wiki.rb
+++ b/app/models/wiki.rb
@@ -185,7 +185,7 @@ class Wiki < ApplicationRecord
 
   def ensure_wiki_exists
     return if errors.any? # Skip this check if the wiki had a validation error.
-    site_info = WikiApi.new(self).query(meta: :siteinfo)
+    site_info = WikiApi.new(self).meta(:siteinfo)
     raise InvalidWikiError, domain if site_info.nil?
     servername = site_info.data.dig('general', 'servername')
     raise InvalidWikiError, domain unless base_url == "https://#{servername}"
--- a/lib/importers/article_importer.rb
+++ b/lib/importers/article_importer.rb
@@ -26,7 +26,7 @@ class ArticleImporter
     # 40 is too much for some languages, such as bn.wipedia.org
     titles.each_slice(30) do |some_article_titles|
       query = { prop: 'info', titles: some_article_titles }
-      response = WikiApi.new(@wiki).query(query)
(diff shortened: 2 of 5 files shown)


empty-codes:
> ## Suspect Issues
> This pull request was deployed and Sentry observed the following issues:
> 
> * ‼️ **MediawikiApi::HttpError: unexpected HTTP response (429)** `Sidekiq/CourseDataUpdateWorker` [View Issue](https://wiki-education.sentry.io/issues/6332476291/?referrer=github-pr-bot)
> 
> Did you find this useful? React with a 👍 or 👎

I just checked and it's the same usual `429` error that occurs so it wasn't caused by this. I guess sentry is notifying because it's the first occurence after I changed `query`

PR #6213, opened 2025-02-25
Fix Redux mutation issue in AdminQuickActions

Abishekcs:
## What this PR does

Replaces direct state mutation with a Redux dispatch call.

## Cause of Bug

- Directly mutating the `course` object inside the component.

## Fix of Bug

- Used `dispatch(updateCourse({ ...course, last_reviewed: { ... } }))`.

## Screenshots
Before:

https://github.com/user-attachments/assets/f65fb809-a6f3-463e-aa8b-68dcae79d0a3

After:

https://github.com/user-attachments/assets/c20a4a2d-a195-49ff-9374-54e9c303c4cb


ragesoss:
Thanks @Abishekcs!

PR #6217, opened 2025-02-25
Fix: Handle Parent Conditional Change for Survey

### LINKED ISSUE #6216: [BUG]: Conditional Questions still persisting even when answers change and are no longer selected
### To Reproduce

> 1. Go to Surveys
> 2. Create a Survey with Multiple Conditional Questions tied to different options of the Parent Question
> 3. Observe the issue

### Screenshots

https://github.com/user-attachments/assets/fe1f1823-0b6e-46eb-8ad6-cc979da4b693

JiyaGupta-cs:
## What this PR does
This PR fixes an issue where conditional questions remain visible even after the user changes their selection in a survey. Previously, when a user selected an option that triggered a conditional question, then switched to a different option, the previously shown conditional question would still be displayed.
The bug was in the ```handleParentConditionalChange``` function which manages which conditional questions should be displayed based on selected answers. The function had several issues:

- Incorrect array manipulation when removing previous answers from the ```currentAnswers``` array
- Logic to identify which conditional questions to show/hide wasn't properly checking for previously selected options

Fixes #6216

## Screenshots
Before:

https://github.com/user-attachments/assets/fe1f1823-0b6e-46eb-8ad6-cc979da4b693

After:

[Screencast from 2025-02-26 01-07-14.webm](https://github.com/user-attachments/assets/c43af4c4-07f4-4a38-8d77-e0c4c47dd6b1)

### DIFF
--- a/app/assets/javascripts/surveys/modules/Survey.js
+++ b/app/assets/javascripts/surveys/modules/Survey.js
@@ -314,13 +314,13 @@ const Survey = {
             answerGroup[answerId] = val;
           }
         } else if (value !== '0') { // Multi-Select (Checkbox)
-            if (typeof answerGroup[answerId] !== 'undefined') {
-              answerGroup[answerId][answerKey].push('0');
-              answerGroup[answerId][answerKey].push(value);
-            } else {
-              answerText[answerKey] = ['0', value];
-              answerGroup[answerId] = answerText;
-            }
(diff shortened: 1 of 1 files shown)


ragesoss:
Sorry for taking so long before providing feedback!

Can you implement it without any unrelated formatting changes, or separate the formatting changes into a separate commit? That will make it much easier for me to think through the fix.

PR #6218, opened 2025-02-25
fix: eslint errors in admin quick actions

JiyaGupta-cs:
## What this PR does
It fixes the ES Lint errors which were seen after merging #6213

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -563,6 +563,7 @@ en:
         Your course has been cloned, including the elements of the timeline
         (weeks and blocks). Be sure to add the dates, which are not carried over.
         Has anything else about your course changed? Feel free to update it now.
+      closed_on: This course was closed on
       copy_successful_details: >
         Your course has been copied, including the elements of the timeline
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -563,6 +563,7 @@ en:
         Your course has been cloned, including the elements of the timeline
         (weeks and blocks). Be sure to add the dates, which are not carried over.
         Has anything else about your course changed? Feel free to update it now.
+      closed_on: This course was closed on
```
ragesoss, line 566: We should only use complete messages in the locale files, rather than fragments like this that interact grammatically with other content before or after. (These are "lego" messages which need to be assembled from pieces, and they cause problems for translators.)

ragesoss: We have a lot of these extant in the codebase. Fixing them is good when it's convenient and makes sense, but I think we need to find a pattern for disabling the linting rule when using i18n is not needed... which is true for any feature that is admin-facing and is only used in Wiki Education mode (like this one).

JiyaGupta-cs: Hmm...need to find an alternative

PR #6227, opened 2025-02-27
Improve method signature for WikiApi with http_method support

ragesoss:


lib/alerts/continued_course_activity_alert_manager.rb
```diff
@@ -33,7 +33,8 @@ def significant_activity_after_course_end?(course)
   def count_revisions_for_wiki(course, wiki)
     # 50 is the max users for query
     course.students.pluck(:username).in_groups_of(40, false).sum do |usernames|
-      response = WikiApi.new(wiki).query(query(course, usernames))
+      post_course_revisions_query = query(course, usernames)
+      response = WikiApi.new(wiki).query post_course_revisions_query
```
gabina, line 37: This is called `post_course_revisions_query`, but what guarantees that this request will be a post request instead of a get request? Shouldn't we set the`http_method: :post` in the parameters?

ragesoss: In this context, I meant "query for post-course revisions", with "post" meaning "after". But I should not have included the change to this file anyway. I was seeing whether this change would fix it without any changes to WikiApi, and forgot to revert it. Updated version leaves your code alone.

Issue #6191, opened 2025-02-10
Merge and deploy re-architecture branch


ragesoss:
We are nearing completion of the data re-architecture work: https://github.com/WikiEducationFoundation/WikiEduDashboard/tree/data-rearchitecture-for-dashboard

This is the largest change to the update system we've ever made, so we should do as much as we can to roll it out safely and reversibly.
 
## Preparation for dashboard.wikiedu.org rollout

- [ ] Test migrations against a recent copy of p&e database.
  NOTE: Modification to existing ArticlesCourses tab is the only migration to worry about. P&E database has 15.5 million ArticlesCourses rows.
- [ ] Open PR to merge data-rearchitecture-for-dashboard
- [ ] Sage reviews PR
- [ ] Estimate database storage requirements for high-activity courses, old vs new system
- [ ] Write message for users about changes and where to report problems

## dashboard.wikiedu.org rollout

- [ ] Add message about changes
- [ ] Pause course updates, pause constant updates, wait for workers to complete current updates
- [ ] Save a server snapshot
- [ ] Clear the course update queues
- [ ] Merge and deploy. (If migration will be slow, run migration manually before Capistrano deployment.)
- [ ] manually add current term courses, ongoing Scholars & Scientists courses, to a queue, enable worker for that queue
- [ ] wait for that queue to process all current courses
- [ ] Monitor for problems for at least several days

## outreachdashboard.wmflabs.org rollout

- [ ] Add message about changes
- [ ] Pause course updates, pause constant updates, wait for workers to complete current updates
- [ ] Save a database dump
- [ ] Clear the course update queues
- [ ] Run migration manually, server-side.
- [ ] Merge and deploy to web server
- [ ] manually add current courses, ordered from least to most revisions, to appropriate queues
- [ ] update sidekiq process servers, enable queue workers one at a time, monitoring for problems

## After rollout

- [ ] monitor database usage
- [ ] monitor server resource utilization
- [ ] monitor Sentry logs
- [ ] begin cleanup of now-unused code

gabina:
# Estimate database storage requirements for high-activity courses, old vs new system #

There are currently two heavy ingested courses that I'm using to estimate database storage requirements.

## 1. Monthly Outstanding Editor Recognition

[Data-rearchitecture link](https://data-rearchitecture-test.wmcloud.org/courses/Wikimedia_Nigeria/Monthly_Outstanding_Editor_Recognition_(Februar))
ID: 10002
[...]

PR #6235, opened 2025-03-04
[Data rearchitecture] Fix bug for scoped articles courses

gabina:
## What this PR does
This PR avoids running the article status manager for `ArticleScopedProgram` and `VisitingScholarship` courses. It was created after debugging course Visiting Scholars at Northeastern University - Rosiestep
 in Wiki Edu Dashboard (see [this](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6191#issuecomment-2695217137) comment).
 
 ### Explanation
 This is the documentation about `ArticleStatusManagerTimeslice`:
```
#= Updates articles to reflect deletion and page moves on Wikipedia
# This class is responsible for two main things (we may separate it in the future).
# - Iterate over article course timeslices to sync articles. This updates title,
# namespace, deleted and mw_page_id fields.
# - Reset articles according to their new status:
#   * Articles that were deleted or untracked. These are articles that were either
#   deleted or moved to a namespace not traceable by the course. Such articles
#   should be excluded from course statistics.
#   * Articles that were restored or re-tracked: These are articles that were
#   either undeleted or moved to a namespace relevant to the course. Such articles
#   should be included in course statistics.
```
The problem with this logic for `ArticleScopedProgram` or `VisitingScholarship` courses, is that it "thinks" that all articles that have an ACT record but do not have an article course record (because they're non-scoped articles) were "re-tracked". That's because those articles are indeed in a tracked wiki and namespace. As no article course record exists for them, the manager "thinks" they were recently moved to a tracked wiki namespace. Then, it marks all timeslices associated to them to be reprocessed.

Note that `UpdateTimeslicesScopedArticle` class takes care of checking if a scoped article was deleted from the list (for example because an assignment is deleted), or if a new one was added (for example, because a category was added). However, this class is not a replacement for the article status manager, since the article status manager checks deletion/movements of the article itself. As we don't expect wikipedia scoped articles to be deleted or moved to a different namespace frequently, we consider OK to skip the article status manager.

### DIFF
--- a/app/services/update_timeslices_course_date.rb
+++ b/app/services/update_timeslices_course_date.rb
@@ -34,7 +34,8 @@ class UpdateTimeslicesCourseDate
 
     remove_timeslices_prior_to_start_date
     remove_timeslices_after_end_date
-    @timeslice_manager.create_timeslices_for_new_course_wiki_records(@course.wikis)
+    @timeslice_manager.create_timeslices_for_new_course_wiki_records(@course.wikis,
+                                                                     needs_update: true)
   end
 
   def update_timeslices_if_start_date_changed
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -346,8 +346,8 @@ class Course < ApplicationRecord
     end.flatten
   end
 
-  def article_status_manager_enabled?
-    true
+  def only_scoped_articles_course?
+    false
   end
 
   # The default implemention retrieves all the revisions.
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -346,6 +346,10 @@ class Course < ApplicationRecord
     end.flatten
   end
 
+  def article_status_manager_enabled?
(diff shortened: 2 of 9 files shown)


app/services/update_timeslices_course_date.rb
```diff
@@ -34,7 +34,8 @@ def update_timeslices_if_drastic_change
 
     remove_timeslices_prior_to_start_date
     remove_timeslices_after_end_date
-    @timeslice_manager.create_timeslices_for_new_course_wiki_records(@course.wikis)
+    @timeslice_manager.create_timeslices_for_new_course_wiki_records(@course.wikis,
+                                                                     needs_update: true)
```
gabina, line 38: This change is not related to this PR. It's just another bug that I found when debugging course Visiting Scholars at Northeastern University

PR #6242, opened 2025-03-06
[Data rearchitecture] Add first_revision field to article course timeslices

gabina:
## What this PR does
This PR adds a new `first_revision` ACT field to make existing `first_revision` article course field depend only on it.

In order to merge and deploy this we should:

1. Manually run the migration to add new `first_revision` field to ACT.
2. Run a script to manually populate all the new `first_revision` fields. The basic idea will be, if the `revision_count` field is 0, then keep the `first_revision` as nil. If the `revision_count` is different from 0, then set `first_revision` to start field.
3. Deploy.

### Explanation
While the solution we implemented in PR https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6230 works fine for most cases, there are some specific edge cases where it could have an incorrect value. For example:
1- The student who made the first revision for a given article is removed from the course. While some CWTs will be marked as needs_update, the first_revision field in AC won't be updated (assuming that other students have also updated the article; otherwise, the AC is deleted).
2- The start date for a course changes to a later date. While all timeslices for dates before the new course start date are deleted, ACs with edits also after the new course start date persist. This means that the first_revision field for those ACs will reflect a date that is before the new course start date.

These are the two cases where we'll have an incorrect first_revision date.

A better approach is to make the first_revision field work similarly to the other AC metrics—that is, by depending only on the ACT values. To do that, the only way is to add a first_revision field to the ACT table and calculate the AC first_revision field based on the first_revision value of the earliest (min) ACT. If we do that, removing a user and their corresponding ACTs won't cause issues, because the AC first_revision will update to the first_revision from the new minimum ACT. This PR implements this approach

Note: commit f68c37604d03114c1764fd03aab9de184dcc341e is not related to these changes. It's bug that I found when working on this so I fix it.


ragesoss:
Code looks good to me.

docs/cleanup_scripts/populate_act_first_revision.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+# This script populates first_revision article course timeslice field
+
+require_dependency "#{Rails.root}/app/models/article_course_timeslice"
+
+ArticleCourseTimeslice.where.not(revision_count: 0).in_batches do |act_batch|
+  # rubocop:disable Rails/SkipsModelValidations
+  act_batch.update_all('first_revision = start')
```
ragesoss, line 8: I don't think I've ever done an update_all like this with a string argument, but I assume you've tested this.

gabina: Yes, I ran it locally. In data-rearchitecture I ran the raw SQL query because I have problems to open a rails console:

```
UPDATE article_course_timeslices
SET first_revision = start
WHERE revision_count != 0
```

ragesoss: Cool. I usually can't open a rails console in a new production server until I jump through some hoops related to installing binstubs, but I always forget exactly what commands are needed (and forget to add them to the docs).

PR #6243, opened 2025-03-06
[Data-rearchitecture] Force revisions to be associated to non-deleted articles

gabina:
## What this PR does
This PR ensures the recently fetched revisions are associated to a non-deleted article. It's a quick try to fix the bug that it's causing [all these sentry logs](https://wiki-education.sentry.io/issues/6338758408/events/?project=6251155&referrer=issue-stream&statsPeriod=90d&stream_index=0).

What if there are no non-deleted article for a given `mw_page_id`?


ragesoss:
> What if there are no non-deleted articles for a given mw_page_id?
ArticleImporter was called on the same data immediately before this, so any articles represented in that data should have been created in undeleted state if they didn't exist already.

PR #6244, opened 2025-03-06
[WIP] Switch to using authenticated mediawiki requests to increase rate limits

empty-codes:
## What this PR does
Resources + Documentation used:
https://www.mediawiki.org/wiki/API:Login
https://www.mediawiki.org/wiki/Manual:Bot_passwords
https://en.wikipedia.org/wiki/Wikipedia:Using_AWB_with_2FA
https://github.com/ragesoss/mediawiki-ruby-api/blob/master/lib/mediawiki_api/client.rb

I created a bot for testing and added `bot_username:` and `bot_password:` to my `application.yml`

- Modifies `api_client` to store clients (i.e the api_urls) as a class variable
using a class level variable

- Adds `ensure_logged_in` method that uses the `MediawikiApi::Client` methods `logged_in`, `get_token` and `log_in` to fetch login tokens and then use that to log in.
- Stubs the `log_in` and `get_token` methods for the purpose of tests

Note; I ran the `wiki_api_spec` tests and they passed, however we expect this to break a lot more tests so the ci/cd build won't pass yet + I need to write test cases to ensure the login process works fine and it is handled appropriately when it does not.

The tests are a bit tricky because any class that uses `WikiApi`'s `mediawiki` method now has to be authenticated which translates to the two extra requests. Now in specs that use VCR_cassettes, these 2 requests will be added to the cassette file unless stubbed. Just running the test suite however won't tell you all the specs this could happen in so we can stub it, since the number of requests in / order of requests is not being tested...

Another thing I noticed is the `stub_wiki_validation` is used 82 times in 53 files so i could use that to infer where to put the `stub_log_in` too.

## Screenshots
Before:

After:

### DIFF
--- a/lib/wiki_api.rb
+++ b/lib/wiki_api.rb
@@ -8,6 +8,9 @@ require_dependency "#{Rails.root}/lib/errors/api_error_handling"
 #= This class is for getting data directly from the MediaWiki API.
 class WikiApi
   include ApiErrorHandling
+  def self.clients
+    @clients ||= {} # Class-level hash to persist clients (different api urls)
+  end
 
   def initialize(wiki = nil, update_service = nil)
     wiki ||= Wiki.default_wiki
@@ -107,7 +110,7 @@ class WikiApi
 
   def mediawiki(action, query)
     tries ||= 3
-    @mediawiki = api_client
+    @mediawiki = ensure_logged_in(api_client)
     @mediawiki.send(action, query)
   rescue StandardError => e
     tries -= 1
@@ -121,7 +124,22 @@ class WikiApi
   end
 
   def api_client
-    MediawikiApi::Client.new @api_url
+    self.class.clients[@api_url] ||= MediawikiApi::Client.new(@api_url)
+  end
+
+  def ensure_logged_in(client)
+    return client if client.logged_in?
+    login_tries = 2
+    begin
+      login_token = client.send(:get_token, 'login') # get_token is a protected method
+      client.log_in(ENV['bot_username'], ENV['bot_password'], login_token)
+      return client
+    rescue MediawikiApi::LoginError => e
+      login_tries -= 1
+      raise e if login_tries.zero?
+      client = MediawikiApi::Client.new(@api_url) # Reset client
+      retry
+    end
   end
 
   def too_many_requests?(e)
--- a/spec/support/request_helpers.rb
+++ b/spec/support/request_helpers.rb
@@ -253,47 +253,63 @@ module RequestHelpers
       .to_return(status: 503, body: '', headers: {})
   end
 
-  def stub_wiki_validation
-    wikis = [
-      'incubator.wikimedia.org',
-      'es.wikipedia.org',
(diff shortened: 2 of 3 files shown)


lib/wiki_api.rb
```diff
@@ -121,7 +124,22 @@ def mediawiki(action, query)
   end
 
   def api_client
-    MediawikiApi::Client.new @api_url
+    self.class.clients[@api_url] ||= MediawikiApi::Client.new(@api_url)
+  end
+
+  def ensure_logged_in(client)
```
empty-codes, line 130: `MediawikiApi::Client`'s
 - logged_in method: https://github.com/ragesoss/mediawiki-ruby-api/blob/f91ca9e7940a5ad22282d5fe714fbf2160815be2/lib/mediawiki_api/client.rb#L17
- get_token method: https://github.com/ragesoss/mediawiki-ruby-api/blob/f91ca9e7940a5ad22282d5fe714fbf2160815be2/lib/mediawiki_api/client.rb#L188
- log_in method: https://github.com/ragesoss/mediawiki-ruby-api/blob/f91ca9e7940a5ad22282d5fe714fbf2160815be2/lib/mediawiki_api/client.rb#L122

spec/support/request_helpers.rb
```diff
@@ -253,50 +253,66 @@ def stub_commons_503_error
...
     end
   end
 
+  def stub_get_token(wiki)
+    fake_tokens = '{"query":{"tokens":{"logintoken":"faketoken+\\\\"}}}'
+    stub_request(:get, "https://#{wiki}/w/api.php?action=query&format=json&meta=tokens&type=login")
+      .to_return(status: 200, body: fake_tokens, headers: {})
+  end
+
+  def stub_log_in
```
empty-codes, line 306: I chose to use this approach as opposed to simply passing in a `wiki` as a parameter to the stub method because it is simpler, than having to call the method multiple times for different wikis in a test. Also these 2 stubs are similar to the `stub_wiki_validation` stub since you need to get tokens and login for each distinct wiki

lib/wiki_api.rb
```diff
@@ -121,7 +124,22 @@ def mediawiki(action, query)
...
+    return client if client.logged_in?
+    login_tries = 2
+    begin
+      login_token = client.send(:get_token, 'login') # get_token is a protected method
```
empty-codes, line 134: The login_token is gotten separately instead of just using `action=login` because that behaviour is deprecated:

```
'{"warnings":{"main":{"*":"Subscribe to the mediawiki-api-announce mailing
        list at <https://lists.wikimedia.org/postorius/lists/mediawiki-api-announce.lists.wikimedia.org/>
        for notice of API deprecations and breaking changes. Use [[Special:ApiFeatureUsage]]
        to see usage of deprecated features by your application."},"login":{"*":"Fetching
        a token via \"action=login\" is deprecated. Use \"action=query&meta=tokens&type=login\"
        instead."}},"login":{"result":"NeedToken","token":"+\\"}}'
```

PR #6250, opened 2025-03-11
[Data-rearchitecture] Update assignments right after importing articles from revisions

gabina:
## What this PR does
This PR forces a call to `update_assignments_for_article` right after importing articles. This is because importing an article could involve updating the title for an existing article, which could leave the assignment for such an article with an inconsistent title.

This was done specifically for course [Visiting Scholars at Northeastern University - Rosiestep](https://dashboard.wikiedu.org/courses/Northeastern_University/Visiting_Scholars_at_Northeastern_University_-_Rosiestep_(2017)/home), which keeps reprocessing 60 timeslices forever.

### DIFF
--- a/lib/importers/article_importer.rb
+++ b/lib/importers/article_importer.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_dependency "#{Rails.root}/lib/assignment_updater"
 require_dependency "#{Rails.root}/lib/replica"
 
 #= Imports articles from Wikipedia into the dashboard database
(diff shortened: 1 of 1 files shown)


lib/importers/article_importer.rb
```diff
@@ -40,9 +41,20 @@ def import_articles_by_title(titles)
...
   def import_articles_from_revision_data(data)
     # We rely on the unique index here, mw_page_id and wiki_id
     Article.import data, on_duplicate_key_update: [:title, :namespace]
+
+    # Recover recently imported articles
+    imported = data.map { |d| [d['mw_page_id'], d['wiki_id']] }
+    # Get the articles with specific mw_page_id and wiki_id
+    articles = imported.each_slice(1000).flat_map do |batch|
+      Article.where(batch.map { '(mw_page_id = ? AND wiki_id = ?)' }.join(' OR '), *batch.flatten)
+    end
```
gabina, line 54: This is not nice, but I think there is no better way to match both `mw_page_id` and `wiki_id` for a list of articles? Finding by `index-hash` could avoid modifying the assignments for deleted articles (which I'm not entirely sure if it's important or not though).

ragesoss: I don't like this very much; it seems like potentially a whole lot of database work that will happen every time we import data, regardless of how many assignments we're dealing with.

Perhaps instead, we can pass in an array of all the mw_page_ids of the assignments for whichever course we're updating, and then only call AssignmentUpdater for articles with those ids?

PR #6251, opened 2025-03-11
Update assignment titles during Article import

ragesoss:
This is intended to address the issue of timeslices requiring reprocessing when an assignment's article_title doesn't match the current title of the Article.


app/models/course.rb
```diff
@@ -577,6 +577,16 @@ def no_sandboxes?
     flags[:no_sandboxes].present?
   end
 
+  def assigned_articles
+    @assigned_articles ||= Article.where(id: assignments.pluck(:article_id)).to_a
```
gabina, line 581: we should use `assigned_article_ids` definition instead of `assignments.pluck(:article_id)`

ragesoss: fixed.

app/models/course_types/custom_revision_filter.rb
```diff
@@ -7,7 +7,12 @@ def filter_revisions(wiki, revisions)
     filtered_data = revisions.select do |_, details|
       article_title = details['article']['title']
       formatted_article_title = ArticleUtils.format_article_title(article_title, wiki)
-      scoped_article_titles.include?(formatted_article_title)
+      # Normally, scoped_article_titles will include all the in-scope articles
+      # but if the title of an assigned article has changed, we still want to process
+      # edits to that article so that we can update the Assignment#article_title.
+      # A secondary check against the mw_page_ids for assigned articles covers that.
+      scoped_article_titles.include?(formatted_article_title) ||
+        assigned_article_page_ids(wiki).include?(details['article']['mw_page_id'].to_i)
```
gabina, line 15: I think we could still have a "forever reprocessing" problem here if there are two articles with different `mw_page_id` but same article title. Because I think this code would accept both articles (because both of them match at least the title). But then, the logic on `UpdateTimeslicesScopedArticle` only looks at `scoped_article_ids`, and will mark as "needs_update" ACT associated to an article with an article course record but without having article id in `scoped_article_ids`.
However, I'm not sure how often this is the case. So maybe we can deploy this, see what percentage of “forever reprocessing” is solved and then re-analyze by adding more cases.

ragesoss: I'm not sure whether it's possible to have two articles with the same title but different mw_page_id if we're looking at a single wiki, and at a batch of revisions that we just fetched, I guess unless there are two different namespaces represented... but I that case the titles from this revisions data would include the namespace prefix if it's not mainspace?

gabina: I have this in my DB after adding assignments for course 11402. The namespace are `main` and `talk`. I don't think we create articles courses for talk namespaces unless it's a specific tracked namespace, so I don't think it will be an issue at least for this specific case.

```
id      |title                              |created_at         |updated_at         |views_updated_at|namespace|rating|rating_updated_at  |deleted|language|average_views|average_views_updated_at|wiki_id|mw_page_id|index_hash|
--------+-----------------------------------+-------------------+-------------------+----------------+---------+------+-------------------+-------+--------+-------------+------------------------+-------+----------+----------+
  125755|I.M.E._Blandin                     |2025-03-03 20:30:18|2025-03-11 17:27:46|                |        0|      |2025-03-03 20:30:18|      0|        |     0.142857|              2025-03-10|      1|  77462360|77462360-1|
51149528|I._M._E._Blandin                   |2025-03-10 15:53:29|2025-03-11 17:27:46|                |        0|      |                   |      0|        |      1.93878|              2025-03-11|      1|  75411769|75411769-1|
51149529|I._M._E._Blandin                   |2025-03-10 15:53:29|2025-03-11 17:27:46|                |        1|      |                   |      0|        |             |                        |      1|  75411998|75411998-1|
51149530|I.M.E._Blandin                     |2025-03-10 15:53:29|2025-03-11 17:27:46|                |        1|      |                   |      0|        |             |                        |      1|  77462361|77462361-1|
```

Issue #6246, opened 2025-03-07
`page_present?` method does not handle when `page_info` is `nil`


empty-codes:
### What is happening?
The [page_present? method](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/7dc281e57fe9adf1b0f691d802e7b92272c71402/app/services/check_assignment_status.rb#L84) throws a `NoMethodError` when `page_info` is `nil`.

Relevant Stack Trace:

```
NoMethodError: undefined method `dig' for nil:NilClass

    page_info.dig('pages', '-1', 'missing').nil?
             ^^^^
  from app/services/check_assignment_status.rb:85:in `page_present?'
  from app/services/check_assignment_status.rb:70:in `page_status'
  from app/services/check_assignment_status.rb:59:in `block in update_status'
  from app/services/check_assignment_status.rb:57:in `each'
  from app/services/check_assignment_status.rb:57:in `update_status'
  from app/services/check_assignment_status.rb:32:in `initialize'
  from app/services/check_assignment_status.rb:12:in `new'
```

### To Reproduce
1. Call page_present? with nil

### Expected behavior
Instead of raising an error, the method should handle `nil` safely and return false.

evangelineMavrommati:
I'm taking a look at this

PR #6254, opened 2025-03-12
Secure CheckAssignmentStatus#page_present? against nil parameters

evangelineMavrommati:
## What this PR does
If the `page_info` param was `nil`, this method would throw

```
NoMethodError: undefined method `dig' for nil:NilClass
```

Now, it will instead early return `false`

See issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6246

please feel free to nit

### DIFF
--- a/spec/services/check_assignment_status_spec.rb
+++ b/spec/services/check_assignment_status_spec.rb
@@ -104,4 +104,13 @@ describe CheckAssignmentStatus do
       expect(assignment_2.reload.peer_review_sandbox_status).to eq('exists_in_userspace')
     end
   end
+
+  describe '#page_present?' do
+    let(:sandbox_url) { 'https://en.wikipedia.org/wiki/User:Ragesock/student_sandbox' }
+    context 'when there is no page_info' do
+      it 'should return false' do
+        expect(subject.page_present?(nil)).to eq(false)
+      end
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


spec/services/check_assignment_status_spec.rb
```diff
@@ -104,4 +104,13 @@
       expect(assignment_2.reload.peer_review_sandbox_status).to eq('exists_in_userspace')
     end
   end
+
+  describe '#page_present?' do
+    let(:sandbox_url) { 'https://en.wikipedia.org/wiki/User:Ragesock/student_sandbox' }
```
empty-codes, line 109: Hi, this is looking great 👏 Just one small suggestion—right now, the test isn't strict enough. Since the key behavior we’re testing is how the method handles `nil`, it would be better to explicitly check that by either making `WikiApi#get_page_info` return `nil` or calling `page_present?` directly with `nil`. Also, the test currently ensures there’s no error, which is good, but it would be even stronger if it also checked that the method returns `false`.

evangelineMavrommati: Thank you for the feedback, I will fix this spec today 🙏

evangelineMavrommati: updated!

empty-codes: Looks good! Thanks 😊. @ragesoss will review/merge it when he has the time.

Issue #6225, opened 2025-02-27
StudentGreetingChecker repeats greeter queries over and over


ragesoss:
StudentGreetingChecker makes the same username pluck query every time through `update_greeting_status`. The usernames should only be plucked once for each time `StudentGreetingChecker` is run. An instance variable with the usernames should fix it.

evangelineMavrommati:
Taking a look at this

lupamo:
Hi, @ragesoss I think you forgot to close this issue, it seemed to have been worked on already.

PR #6256, opened 2025-03-13
Set username as instance var on StudentGreetingChecker

evangelineMavrommati:
## What this PR does
Only pull greeter usernames once for every time `StudentGreetingChecker` is initialized.

See issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6225

As always, please feel free to nit

### DIFF
--- a/lib/student_greeting_checker.rb
+++ b/lib/student_greeting_checker.rb
@@ -7,7 +7,7 @@ class StudentGreetingChecker
   end
 
   def initialize
-    @greeters = User.where(greeter: true)
+    @greeter_usernames = User.where(greeter: true).pluck(:username)
   end
 
   def check_all_ungreeted_students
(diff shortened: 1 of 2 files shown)


lib/student_greeting_checker.rb
```diff
@@ -30,12 +30,13 @@ def initialize(student, wiki, greeters)
       @student = student
       @wiki = wiki
       @all_greeters = greeters
+      @username = greeters.pluck(:username)
```
ragesoss, line 33: This is an array of possibly many usernames, so the variable name should not be singular.

ragesoss: This `Check` class gets initialized once for every student being checked, so I don't think this solves the problem; it will still run a new pluck query for every student checked.

spec/lib/student_greeting_checker_spec.rb
```diff
@@ -58,6 +58,16 @@
           expect(checker.send(:talk_page_blank?)).to eq(true)
         end
       end
+
+      describe '#update_greeting_status' do
+        it 'should not try and pluck username from greeters' do
```
ragesoss, line 63: This is not a useful test, as it is about a specific implementation rather than the expected behavior of the system.

ragesoss (changes requested):

lib/student_greeting_checker.rb
evangelineMavrommati: ah, gotcha - I misunderstood your instance variable ask. It should be an instance variable on `StudentGreetingChecker`. Could then pass that as a param to `Check`. Would that work instead?

evangelineMavrommati:
@ragesoss able to take another look?

lib/student_greeting_checker.rb
```diff
@@ -20,22 +21,23 @@ def check_all_ungreeted_students
...
   def check(student, wiki)
-    Check.new(student, wiki, @greeters).update_greeting_status
+    Check.new(student, wiki, @greeters, @usernames).update_greeting_status
   end
 
   # Checks a single student to see if they are actually ungreeted, and updates
   # their greeted attribute if so.
   class Check
-    def initialize(student, wiki, greeters)
+    def initialize(student, wiki, greeters, usernames)
```
ragesoss, line 30: Is `greeters` still used here? Perhaps `greeters` should become `greeter_usernames`? I suspect we only need one instance variable (passed along to the `Check` class) for the greeters for this whole file.

evangelineMavrommati: updated!

ragesoss:
LGTM!

PR #6260, opened 2025-03-18
Try to speed up queries in alert monitors

gabina:
## What this PR does
This PR tries to speed up the queries done on some monitors. We found the following slow query in our logs:

`SELECT articles_courses.* FROM articles_courses INNER JOIN articles ON articles.id = articles_courses.article_id WHERE articles.title IN ('1946_Manchester_Borough_Council_election', ...) AND articles.wiki_id = 1;`

This same query seems to be done in different monitors: `ArticlesForDeletionMonitor`, `GANominationMonitor` and `HighQualityArticleMonitor`.

After some quick checks, the strategy of this PR is to split the original query by doing this:

1. `SELECT articles.id FROM articles WHERE title IN (...) and wiki_id =...;`  67 rows in set (6.270 sec)
2. `SELECT articles_courses WHERE article_id in (...);` 20 rows in set (0.003 sec)

Instead of:

1. `SELECT articles_courses.* FROM articles_courses INNER JOIN articles ON articles.id = articles_courses.article_id WHERE articles.title IN (...) AND articles.wiki_id = 1;` 20 rows in set (49.777 sec)

Note that similar queries are also done in:
-  `DYKNominationMonitor` (but a specific course is set)
```
    course_articles = ArticlesCourses.joins(:article)
                                     .where(course: Course.current)
                                     .where(articles: { title: @page_titles, wiki_id: @wiki.id }) 
```

### DIFF
--- a/lib/alerts/articles_for_deletion_monitor.rb
+++ b/lib/alerts/articles_for_deletion_monitor.rb
@@ -54,11 +54,11 @@ class ArticlesForDeletionMonitor
     find_proposed_deletions
     find_candidates_for_speedy_deletion
     normalize_titles
+    set_article_ids
   end
 
   def create_alerts_from_page_titles
-    course_articles = ArticlesCourses.joins(:article)
-                                     .where(articles: { title: @page_titles, wiki_id: @wiki.id })
+    course_articles = ArticlesCourses.where(article_id: @article_ids)
     course_articles.each do |articles_course|
       create_alert(articles_course)
     end
@@ -100,6 +100,10 @@ class ArticlesForDeletionMonitor
     end
     @page_titles.compact!
     @page_titles.uniq!
+  end
+
+  def set_article_ids
+    @article_ids = Article.where(title: @page_titles, wiki_id: @wiki.id).pluck(:id)
   end
 
   def create_alert(articles_course)
--- a/lib/alerts/discretionary_sanctions_monitor.rb
+++ b/lib/alerts/discretionary_sanctions_monitor.rb
@@ -15,11 +15,11 @@ class DiscretionarySanctionsMonitor
     find_pages_tagged_for_discretionary_sanctions
     extract_page_titles_from_talk_titles
     normalize_titles
+    set_article_ids
   end
 
   def create_alerts_from_page_titles
-    course_articles = ArticlesCourses.joins(:article)
-                                     .where(articles: { title: @page_titles, wiki_id: @wiki.id })
+    course_articles = ArticlesCourses.where(article_id: @article_ids)
     course_assignments = Assignment.joins(:article)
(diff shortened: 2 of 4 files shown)


gabina:
Failing specs don't seem to be related to these changes

```
rspec ./spec/features/campaign_overview_spec.rb:211 # campaign overview page as a campaign organizer campaign details start and end dates updates the date fields properly, and unsets if #use_dates is unchecked
rspec ./spec/features/faq_spec.rb:55 # FAQs NEW page redirects to SHOW after creating
rspec ./spec/features/survey_admin_spec.rb:231 # Survey Administration an admin correctly clones question groups with conditional questions
rspec ./spec/features/update_username_spec.rb:13 # Update username controller it visits /update_username and enter updated username
```

Abishekcs:
> SELECT articles_courses.* FROM articles_courses INNER JOIN articles ON articles.id = articles_courses.article_id WHERE articles.title IN ('1946_Manchester_Borough_Council_election', ...) AND articles.wiki_id = 1;
SELECT articles_courses.* 
FROM articles_courses INNER JOIN articles 
ON articles.id = articles_courses.article_id 
WHERE articles.title IN ('Lucia_Casalini_Torelli') 
AND articles.wiki_id = 2;

Hi @Gabina, I was just looking through and had a quick question—would it be possible to include namespace in the WHERE clause of this SQL query?

gabina:
Hey @Abishekcs! I don't understand if your question is related to this PR. You mentioned these two queries:

```
SELECT articles_courses.*
FROM articles_courses INNER JOIN articles
ON articles.id = articles_courses.article_id
WHERE articles.title IN ('1946_Manchester_Borough_Council_election', ...)
AND articles.wiki_id = 1;
```

```
SELECT articles_courses.*
FROM articles_courses INNER JOIN articles
ON articles.id = articles_courses.article_id
WHERE articles.title IN ('Lucia_Casalini_Torelli')
AND articles.wiki_id = 2;
```
But I think this PR replaced those queries with different ones (no inner join is done now I think?).

Could you elaborate on where your question comes from, please?

Abishekcs:
Screenshot from 2025-03-28 02-31-07


> Could you elaborate on where your question comes from, please?

because i found there is muti index on articles for `namespace wikied title`

gabina:
> So, i was thinking whether we can utilize this three together for search since they already exists as multi index was just curious😁

hmm so I think I understand your point but this PR is already closed and we took another strategy (splitting the query in two parts). In addition, we should have to check for this case if we are interested only in some particular namespace (so we can fix `namespace` to 1 for example) or if we want to retrieve articles no matter the `namespace`. I don't remember how these alerts pretend to work.

But I guess you have a point and that strategy may be taken into account to improve the performance of other queries.

Abishekcs:
Okay, got it! Thank you for your response.

PR #6261, opened 2025-03-18
Fix: Recent failures in the CI build

Formasitchijoh:
## What this PR does
This PR solves the recent failures in the CI build for the following test 

- [ ] campaign_overview_spec.rb

### DIFF
--- a/spec/features/campaign_overview_spec.rb
+++ b/spec/features/campaign_overview_spec.rb
@@ -20,6 +20,14 @@ describe 'campaign overview page', type: :feature, js: true do
            description: 'This is the best campaign')
   end
 
+  before do
+    travel_to Time.zone.local(2016, 0o1, 0o5, 0o1, 0o4, 44)
+  end
+
+  after do
+    travel_back
+  end
+
   describe 'header' do
     before do
       campaign_two = create(:campaign_two)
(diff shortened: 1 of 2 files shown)


Formasitchijoh:
> Thanks! Do you know why this one is failing? I'd prefer to adjust it some other way than to use `travel_to`, if possible.

I realized the issue was with the date here 
` expect(campaign.reload.start).to eq(Time.zone.parse('2016-01-10 00:00:00'))`
which returned `nil`
let me capture a screenshot of the error

Formasitchijoh:
I have ran the other tests
- [ ] faq_spec.rb:55 
- [ ] survey_admin_spec.rb:231
- [ ] update_username_spec.rb:13 
they all pass locally but for all the failures

ragesoss:
Hmm... that's weird! I wonder why it returns `nil`. These campaign-related ones pass locally for me as well.

Formasitchijoh:
> Hmm... that's weird! I wonder why it returns `nil`. These campaign-related ones pass locally for me as well.

currently sending from my mobile phone, I have power outage since yesterday, I will resume this once it is back

ragesoss:
I think all these failing tests were from race conditions between the browser and the Rails backend. Check out the changes I committed today.

Issue #6236, opened 2025-03-04
For courses tracking multiple wikis, link to global edits tool instead of home wiki contributions page


ragesoss:
See https://phabricator.wikimedia.org/T387895

betty-alagwu:
@ragesoss would love to work on this #outreachy

PR #6264, opened 2025-03-21
 Improvement/change edits link to global edits

betty-alagwu:
#6236 

## What this PR does
This PR modifies the "Edits" link to "Global edits" on Campaign Dashboard application

## Screenshots
Before:
Screenshot 2025-03-21 at 7 39 40 PM

After:
Screenshot 2025-03-21 at 7 41 46 PM

Video with latest changes (showing condition)

https://github.com/user-attachments/assets/7b61eb48-b55b-41f7-82ec-dc8a4e1a3b31

### DIFF
--- a/app/assets/javascripts/components/students/shared/StudentList/Student/Student.jsx
+++ b/app/assets/javascripts/components/students/shared/StudentList/Student/Student.jsx
@@ -116,7 +116,7 @@ const Student = createReactClass({
           <div className="sandbox-link">
             <a onClick={this.stop} href={student.sandbox_url} target="_blank">{I18n.t('users.sandboxes')}</a>
             &nbsp;
-            <a onClick={this.stop} href={student.contribution_url} target="_blank">{I18n.t('users.edits')}</a>
+            <a onClick={this.stop} href={student.global_contribution_url} target="_blank">{I18n.t('users.edits')}</a>
           </div>
           <ExerciseProgressDescription student={student} />
           <TrainingProgressDescription student={student} />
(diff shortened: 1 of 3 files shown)


ragesoss:
Thanks! We only want to make this change for courses where there is more than one wiki, though.

I suggest keeping the backend changes you've implemented, and add fronted code to switch between the current  edits link and the new one, depending on whether the course has more than one wiki.

betty-alagwu:
> Thanks! We only want to make this change for courses where there is more than one wiki, though.
> 
> I suggest keeping the backend changes you've implemented, and add fronted code to switch between the current edits link and the new one, depending on whether the course has more than one wiki.

Got it! I'll keep the backend changes and update the frontend to conditionally switch the edits link based on the number of wikis in a course. Thanks for the clarification—I’ll make the changes and push an update soon!

ragesoss:
Thanks! Can you include `after` screenshots demonstrating both variations of the link?

betty-alagwu:
@ragesoss I updated with a video. Thank you for the reminder.

ragesoss:
@betty-alagwu I merged and deployed this, and then got confused about whether it was working, until I realized that this PR only changed one of three places that use the contributions link. The other ones both apply when viewing an individual editor, like https://outreachdashboard.wmflabs.org/courses/Europlanet_asbl/Planetary_Scientists_Edit-a-thon_2025/students/articles/DiBs97

The 'edits' link there (from the Articles/SelectedStudent/Header.jsx component) should do the same thing, and so should "View full contribution history" link that appears at the bottom when you expand the Revisions drawer.

betty-alagwu:
@ragesoss Thank you, I'll create a follow-up pull request.

PR #6279, opened 2025-03-29
Optimize Survey Recent Responses fetching to prevent N+1 queries

### LINKED ISSUE #984: Survey results page is super slow to load
For admins viewing survey results, it takes a very long time — over 30 seconds — to view results from a survey with many questions and many responses.

Improve the performance of this page by a factor of 5 or more.

Abishekcs:
Found this while trying to resolve #984

## What this PR does

- Added `.includes(:question_group, :user)` to preload related models, preventing unnecessary queries `(1+N)`.

- Replaced `each` loop with `.map` to make array transformation **more efficient**.

- Used `.limit(100)` instead of .first(100) to ensure filtering happens at the database level.

##  Why This Change?

- Previously, fetching `AnswerGroup` records triggered  `additional queries` to load users individually `(User.find)`.
- Now, all required users and question groups are `loaded in bulk`, reducing additional queries

## Screenshots
**Before:**

https://github.com/user-attachments/assets/6ac67a05-33e4-4f60-b273-a311968704ba

**After:**

https://github.com/user-attachments/assets/179e4a02-a239-406c-8e98-73b03d130944

## Benchmark

https://github.com/user-attachments/assets/4dcb30b1-2d26-4a04-9be7-c30c0f4a621d

## Guard Logs

**Before**:

Before

**After**:

After

### DIFF
--- a/app/controllers/survey_responses_controller.rb
+++ b/app/controllers/survey_responses_controller.rb
@@ -5,11 +5,12 @@ class SurveyResponsesController < ApplicationController
   layout 'surveys'
 
   def index
-    @responses = []
-    Rapidfire::AnswerGroup.order(created_at: :desc).first(100).each do |answer_group|
-      @responses << { user: User.find(answer_group.user_id),
-                      answer_group:,
-                      question_group: answer_group.question_group }
+    answer_groups = Rapidfire::AnswerGroup.includes(:question_group, :user)
+                                          .order(created_at: :desc)
+                                          .limit(100)
+
+    @responses = answer_groups.map do |answer_group|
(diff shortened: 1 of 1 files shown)


ragesoss:
This looks like a fine change, and I'm happy to merge it. This is not the route that #984 is about though. The slow route is `SurveysController#results` (both HTML and CSV versions), which you can reach via the 'Results' link for an individual survey from `/surveys/results`.

Abishekcs:
> This looks like a fine change, and I'm happy to merge it. This is not the route that #984 is about though. The slow route is `SurveysController#results` (both HTML and CSV versions), which you can reach via the 'Results' link for an individual survey from `/surveys/results`.

Okay, thank you!

PR #6281, opened 2025-04-01
Fix Redux state mutation in `CourseTypeSelector` and `YesNoSelector` component

Abishekcs:
## What this PR does

This PR fixes all the Redux state mutations caused by the options shown below in the screenshot.

Screenshot from 2025-04-01 20-49-26

## Fix for the mutation

This fix is similar to other Redux state mutation fixes in the codebase, i.e., creating a new reference for the object before modifying its value.

## Screenshots

Before:

https://github.com/user-attachments/assets/7ddeceee-4e6d-4139-b7de-040bfa45cd82

After:

https://github.com/user-attachments/assets/c4987c92-94fb-4158-9e19-fc8415d2fe3d

@ragesoss, I’m not sure why, but the Rack Mini Profiler on the course page appears briefly and then disappears immediately. This only happens on the course page. Do you have any idea what might be causing this? Also, could this be an issue specific to my local development? I’m not sure.

https://github.com/user-attachments/assets/6435b231-dd60-4cfa-8ce2-926fbe535c31

### DIFF
--- a/app/assets/javascripts/components/overview/course_type_selector.jsx
+++ b/app/assets/javascripts/components/overview/course_type_selector.jsx
@@ -20,19 +20,16 @@ const CourseTypeSelector = (props) => {
   const [selectedOption, setSelectedOption] = useState({ value: props.course.type, label: _getFormattedCourseType(props.course.type) });
 
   const _handleChange = (e) => {
-    const course = props.course;
     const courseType = e.value;
-    course.type = courseType;
+    let { timeline_start, timeline_end } = props.course;
     setSelectedOption(e);
-    if (courseType === 'ClassroomProgramCourse' || course.timeline_enabled) {
-      if (!course.timeline_start) {
-        course.timeline_start = course.start;
-      }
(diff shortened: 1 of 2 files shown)


Abishekcs:
> Instead of implementing these changes in the components, would it be feasible to prevent mutations within the reducer? That seems like potentially a better strategy to guarantee that we don't get more mutations in new components later on.

Yes, that should definitely be possible! However, we need to ensure a few things:

- Identify all the components that are using the updateCourseAction.

- Ensure that these components don’t mutate values directly. Instead, they should pass the updated values to the Redux action, where the immutable update is handled properly.

Abishekcs:
Also, today I was testing out [eslint-plugin-react-compiler](https://react.dev/learn/react-compiler#installing-eslint-plugin-react-compiler) for my local development. This helps in identifying mutations and, in some cases, unnecessary re-renders or when memoization is needed. But, need to cross check since `it's still in beta`


Screenshot from 2025-04-01 23-31-49

Issue #6224, opened 2025-02-27
CheckAssignmentStatus is slow and inefficient


ragesoss:
During the ConstantUpdate process, the `CheckAssignmentStatus` takes quite a while in Wiki Education, and makes a whole lot more database queries than are probably necessary... basically a constant stream of Wiki load, User load, and Assignment exists? queries.

It also makes a separate query to mediawiki for every page it checks.

There's probably a lot of efficiency to be gained by handling this process better in terms of batching both database queries and external API usage.

syedbarimanjan:
@ragesoss I would like to work on this.

betty-alagwu:
@syedbarimanjan can you please confirm if you're still working on this, I've got an idea on how to resolve this issue.

betty-alagwu:
@ragesoss I'll be taking this on

Ajay-singh1:
@ragesoss If this one is solved we can close the issue.

PR #6282, opened 2025-04-01
Improvement/update performance of assignment status checker

betty-alagwu:
#6224

## What this PR does
**Performance Optimization**: Replaces N individual DB updates with a single transactional batch update.

## Screenshots
Before:
```rb
Course.current.each do |course|
      course.assignments.where.not(user: nil).each do |assignment|
        new(assignment)`
 ```       
Previously, We were looping through each of the `Course` and `assignment` and also initializing each assignment

After:
```rb
assignments = Assignment.joins(:course).merge(Course.current).where.not(user: nil)
```

We're making a single database call to get the assignments in the new update. 

Before:
```rb
def page_status(page_info)
    return AssignmentPipeline::SandboxStatuses::DOES_NOT_EXIST unless page_present?(page_info)

    case page_info['pages'].values.first['ns']
    when Article::Namespaces::USER
      AssignmentPipeline::SandboxStatuses::EXISTS_IN_USERSPACE
    when Article::Namespaces::DRAFT
      AssignmentPipeline::SandboxStatuses::EXISTS_IN_DRAFT_SPACE
    when Article::Namespaces::MAINSPACE
      AssignmentPipeline::SandboxStatuses::EXISTS_IN_MAINSPACE
    else
      AssignmentPipeline::SandboxStatuses::EXISTS_ELSEWHERE
end
```
This function checks the namespace of a single page and assigns a status based on predefined conditions.

It only processes one page at a time, making it inefficient when handling multiple pages.

The lookup is done using page_info['pages'].values.first, which assumes only one page exists per call.

After:
```rb
def process_batch(entries, batch, page_infos)
      return unless page_infos && page_infos['pages']

      pages_by_normalized_title = build_pages_by_normalized_title(page_infos)

      Assignment.transaction do
        update_assignments_for_batch(entries, batch, pages_by_normalized_title)
      end
end


```

Batch Processing: Instead of handling pages one at a time, it now processes an entire batch in a single database transaction.
Caching Data: Instead of repeatedly checking page_infos, it builds a hash (pages_by_normalized_title) for faster lookups.

Efficient Database Updates: Uses Assignment.transaction to minimize DB writes, improving performance when updating multiple pages at once.

### DIFF
--- a/app/services/check_assignment_status.rb
+++ b/app/services/check_assignment_status.rb
@@ -57,22 +57,32 @@ class CheckAssignmentStatus
     def process_batch(entries, batch, page_infos)
       return unless page_infos && page_infos['pages']
 
-      # Convert pages hash to array maintaining API response order
-      pages_array = page_infos['pages'].values.sort_by { |page| page['index'] }
+      pages_by_normalized_title = {}
+      page_infos['pages'].each_value do |page|
+        normalized_title = page['title'].tr(' ', '_')
 
-      batch.each_with_index do |pagename, index|
-        # Get page data by index instead of searching by title
-        page_data = pages_array[index]
-        status = if page_data
-                  status_from_namespace(page_data['ns'])
-                else
-                  AssignmentPipeline::SandboxStatuses::DOES_NOT_EXIST
-                end
+        page['present'] = !page.key?('missing')
 
-        entries_for_pagename = entries.select { |e| e[:pagename] == pagename }
+        pages_by_normalized_title[normalized_title] = page
+      end
 
-        entries_for_pagename.each do |entry|
-          entry[:assignment].update_sandbox_status(entry[:key], status)
+      Assignment.transaction do
+        batch.each do |pagename|
+          # Get page data by index instead of searching by title
+          normalized_pagename = pagename.tr(' ', '_')
+          page_data = pages_by_normalized_title[normalized_pagename]
+
+          status = if page_data
+                    status_from_namespace(page_data['ns'], page_data['present'])
+                  else
+                    AssignmentPipeline::SandboxStatuses::DOES_NOT_EXIST
+                  end
+
+          entries_for_pagename = entries.select { |e| e[:pagename] == pagename }
+
+          entries_for_pagename.each do |entry|
+            entry[:assignment].update_sandbox_status(entry[:key], status)
+          end
         end
       end
     end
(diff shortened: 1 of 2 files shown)


betty-alagwu:
@ragesoss, kindly review this PR when you get the chance.

app/services/check_assignment_status.rb
```diff
@@ -1,88 +1,126 @@
...
-    Course.current.each do |course|
-      course.assignments.where.not(user: nil).each do |assignment|
-        new(assignment)
+    assignments = Assignment.joins(:course).merge(Course.current).where.not(user: nil)
+    sandboxes_to_check = collect_sandboxes(assignments)
+
+    process_sandboxes(sandboxes_to_check)
+  end
+
+  class << self
```
ragesoss, line 15: I'd prefer not to use `class << self` here, as the extra nesting makes the code a little harder to read (since we don't use this format very many other places in the codebase).

app/services/check_assignment_status.rb
```diff
@@ -1,88 +1,126 @@
...
+      end
+    end
+
+    def update_entries_for_pagename(entries, pagename, status)
+      entries.select { |e| e[:pagename] == pagename }.each do |entry|
+        entry[:assignment].update_sandbox_status(entry[:key], status)
+      end
+    end
+
+    def find_page_data(page_infos, pagename)
```
ragesoss, line 107: Comments showing what the page_infos data looks like would be useful here, since the code is doing nontrivial things to search through a specific expected data structure.

app/services/check_assignment_status.rb
ragesoss: I think the code would be a little simpler if instances were still used, but instead of going one assignment at a time, it went one batch at a time, so that the each instance is expected to make one API request and the page_infos for that request can be an instance variable.

ragesoss (commented):
Overall, this looks like solid work. Can you include before/after details of the database queries and the network requests?

betty-alagwu:
@ragesoss in the screenshot below, the optimization we made for the API call is:

Before, we were making 1 API call for each assignment in the database. After, we're making 1 API for all assignments in the database

Before: 
Screenshot 2025-05-02 at 8 47 59 PM

After:

Screenshot 2025-05-02 at 8 43 42 PM

ragesoss:
Nice work! This is much faster!

Abdulwahab03:
Hi @ragesoss is this still open, I would like to take up this, is it still open

Issue #2460, opened 2019-02-11
Query to fetch user uploads in UserProfilesController#stats is slow for users with many uploads


ragesoss:
Part of the query for user stats involves fetching the latest 20 uploads for that user. This is extremely slow for a user with lots and lots of uploads (eg, User:Pharos).

My guess is that it's because of the ordering, which makes it miss the index. Might be solveable by adding an index for the date of upload.

marklocklear:
How can I reproduce this locally? Any way to copy user data/contributions to my local environment?

ragesoss:
One way would be to create a course locally, and set the start and end dates very wide (eg, 2014 - 2022), then add 'Pharos' as a participant and do a data update.

You could perhaps get a more realistically full database by copying each of the courses that user has participated in:
https://outreachdashboard.wmflabs.org/users/Pharos

I'm not sure the description is right, as this user doesn't actually have a large number of uploads personally. (The large upload number is from events he has led.) But in any case, the user stats route is very slow for this user.

marklocklear:
I'm getting an "Internal Server Error" when trying to create a program. Any ideas? Here is what I am seeing in the console:

```
app/controllers/courses_controller.rb:33:in `create'
[2022-03-03 15:26:40.521 INFO ] Processing by ErrorsController#internal_server_error as JSON
[...]

PR #6284, opened 2025-04-03
Fix: Optimize upload usage count retrieval 

oree-xx:
#2460

## What this PR does
This PR will optimize the upload usage count logic.

1. Reduced N+1 queries within each course:

- Original: For each course, it ran course.uploads.where(user_id: @user.id) which loads all upload objects

- Updated: Uses a single direct query with pluck for each course, only fetching the needed id and usage_count

2. Early termination:

- Original: No early exit check

- Updated: return if student_courses.empty? avoids unnecessary work if there are no courses

3. Eager loading of course data:

- Original: Potentially makes repeated queries to load course date information

- Updated: Loads all course IDs, start dates, and end dates in a single query with pluck(:id, :start, :end) 

4. More efficient data structure usage:

- Original: Loads full ActiveRecord objects for every upload

- Updated: Only extracts the specific fields needed (id, usage_count), reducing memory usage 

5. Database-level filtering:

- Original: Loads data into Ruby and then filters

- Updated: Pushes more filtering to the database level with clear date range conditions

The optimized version maintains the same logic while being more performant.

## Screenshots
Before:
Screenshot 2025-04-03 at 19 05 32

After:
Screenshot 2025-04-03 at 20 41 48

### DIFF
--- a/app/presenters/individual_statistics_presenter.rb
+++ b/app/presenters/individual_statistics_presenter.rb
@@ -97,16 +97,22 @@ class IndividualStatisticsPresenter
   def set_upload_usage_counts
     @upload_usage_counts = {}
     
-    # Get all course IDs in a single query
-    course_ids = individual_courses.pluck(:id)
-    return if course_ids.empty?
+    # Get all student courses with their date ranges
+    student_courses = individual_courses
+                      .includes(:courses_users)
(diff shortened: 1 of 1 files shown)


oree-xx:
@ragesoss I don't think the failing check is related to this fix

ragesoss:
LGTM!

PR #6287, opened 2025-04-05
 Optimize data loading for list of Articles edited in a Campaign, shown in `Articles Edited` tab

Abishekcs:
**WORK IN PROGRESS**

## What this PR does (**TO BE UPDATED**)

## Screenshots (**TO BE UPDATED**)
Before:

After:
## Open questions and concerns (**TO BE UPDATED**)

### DIFF
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -61,19 +61,71 @@ class ArticlesCourses < ApplicationRecord # rubocop:disable Metrics/ClassLength
   end
 
   def live_manual_revisions
+    if !fetched_live_manual_revisions && fetched_course_article_ids
+      revisions = course.revisions.live.where(article_id: fetched_course_article_ids)
+      return self.class.store_live_manual_revisions(revisions)
+    end
+
     course.revisions.live.where(article_id:)
   end
 
   def all_revisions
+    if !fetched_all_revisions && fetched_course_article_ids
+      revisions = course.all_revisions.where(article_id: fetched_course_article_ids)
+      self.class.store_all_revisions(revisions)
+      return
+    end
+
     course.all_revisions.where(article_id:)
   end
 
   def article_revisions
-    article.revisions.where('date >= ?', course.start).where('date <= ?', course.end)
+    if !fetched_revisions_by_article && fetched_course_article_ids
+      revisions = Revision.where(article_id: fetched_course_article_ids)
+                          .where(date: course.start..course.end)
+
+      self.class.store_revisions_by_article(revisions)
+      return
+    end
+
+    article.revisions
+           .where('date >= ?', course.start)
+           .where('date <= ?', course.end)
   end
 
-  def update_cache
-    revisions = live_manual_revisions.load
+  def fetched_live_manual_revisions
+    self.class.live_manual_revisions
+  end
+
+  def fetched_course_article_ids
+    self.class.course_article_ids
+  end
+
+  def fetched_all_revisions
+    self.class.all_revisions
+  end
+
+  def fetched_revisions_by_article
+    self.class.revisions_by_article
+  end
+
+  def update_cache # rubocop:disable Metrics
+    # Fetch relevant revisions:
+    # a) Use previously fetched live manual revisions if course article IDs were stored in advance
+    # b) Otherwise, load revisions for the current article only
+    revisions = if fetched_live_manual_revisions
+                  fetched_live_manual_revisions.select { |r| r.article_id == article_id }
+                else
+                  live_manual_revisions.load
+                end
+
+    # Load all revisions only if we have previously stored all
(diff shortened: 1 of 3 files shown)


Abishekcs:
@ragesoss Whenever you have time, please take a look at this PR. Thank you!

Currently, there were three spec failures, but I think neither is related to the changes introduced in this PR:
Test

a) A fix for article__viewer_spec.rb has already been opened by @Formasitchijoh in PR #6288.
b) I attempted a fix for assigned_articles_spec.rb, which is included in this PR. However, I'm not entirely sure if it's the correct approach to fix this spec.

ragesoss:
Okay! I was holding off because it's set as draft. Is it no longer a draft, and ready to merge from your perspective?

ragesoss:
This is about articles for a Course, not a Campaign, correct? (Title had me confused.)

Abishekcs:
> Okay! I was holding off because it's set as draft. Is it no longer a draft, and ready to merge from your perspective?

yes

ragesoss:
Oh, this is interesting. So, the N+1 problem is related to `ArticlesCourses#update_cache` method, which gets called whenever one of the stats fields is missing. So, for example, if no view counts have yet been populated for any of the AC records for a course, then every article will undergo the cache update during the request for the articles data. Normally, these cache updates happen during the background update cycle, rather than during web requests, but this seems like a pretty bad situation for any production situation where it would happen.

However, during the update cycle we now use `update_cache_from_timeslices` rather than `update_cache`, and we have plans to remove the old `update_cache` or replace it with the timeslice-based implementation. (This is on @gabina's agenda.)

In the new system, we don't do any revisions queries to get view counts, as we have stopped save Revision records in the database. However, we still rely on ac.view_count values for old records that haven't been updated in the new system. See https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/46b72c48b7d134227bcc69dc740b66a4f5e5e657

I think next steps here should be to clean up all the uses of the old `update_cache`, perhaps by changed all the methods that use `update_cache`. We can also remove the :view_count attribute from articles.json (in the `json.call(ac ...) line` as it is already not functional since the view_count gets overwritten with the new implementation with `json.view_count calculate_view_count(ac.first_revision, article.average_views, ac.view_count)`. I think that we could maybe just move the `calculate_view_count` implementation to `AriclesCourses#view_count`; accessing that field should not be able to trigger any cache updates, as we no longer want to rely on the AC cache for the values (instead using the timestamp of first revision and the Article record's average views).

gabina:
hmm I see the problem here. I don't think that running `update_cache` (or even `update_cache_from_timeslices`) if one of the stats fields is missing makes sense. I guess that if any statistics are missing, it's probably because the timeslices (or the revisions) are not yet complete, so updating caches won't fix anything?

ragesoss:
Yeah, I think removing those update_cache calls probably makes sense. I don't remember why they were added originally.

Abishekcs:
> Yeah, I think removing those update_cache calls probably makes sense. I don't remember why they were added originally.

Okay, great! Then I’ll go ahead and remove the update_cache calls.

Abishekcs:
> I think next steps here should be to clean up all the uses of the old `update_cache`, perhaps by changed all the methods that use `update_cache`

@ragesoss, should I open a separate PR for this, or implement it in this one?

ragesoss:
@Abishekcs I think let's hold off on this; Gabina is about to start removing the code for the old update methods that are dead or nearly-dead code, and I expect that she'll clean this up along the way. After that, we can see if any N+1 issues around this remain or not.

Abishekcs:
Okay, I’ll try to work on improving the performance of Surveys#results next.

gabina:
FYI, I included removing `update_cache` method and calls in issue #6303

app/models/course_data/articles_courses.rb
```diff
@@ -41,6 +41,15 @@ class ArticlesCourses < ApplicationRecord
...
   # Instance methods #
   ####################
+  def view_count
+    # view_count AC Field is no longer used in the timeslice system
+    # however, this is a hack to display article views for historical courses
+    # that will not receive a new update
+    return self[:view_count] if self[:first_revision].nil? || self[:average_views].nil?
+    days = (Time.now.utc.to_date - self[first_revision.to_date]).to_i
+    (days * self[:average_views]).to_i
+  end
```
ragesoss, line 51: The current implementation of `calculate_view_count` uses the Article's `average_views`, and we don't have an `average_views` column for ArticlesCourses. Am I missing something about how this works?

Abishekcs: To be honest, I'm a bit confused about this now. I'll go through it once again an update afterward. Thank you for the review!

Abishekcs:
Closing this PR as it is not required anymore.

Issue #6267, opened 2025-03-23
Improve performance for slow queries


gabina:
We enabled the slow query log feature on the `outreachdashboard` server to identify and optimize queries that take excessively long to execute. Currently, the log captures queries that run for more than one minute. 
The plan is to select a query, analyze it, optimize its performance, and submit a pull request with the improvements.

- [X] **Alert monitor queries.** 
```
SELECT articles_courses.* FROM articles_courses INNER JOIN articles ON articles.id = articles_courses.article_id WHERE articles.title IN ('1946_Manchester_Borough_Council_election', ...) AND articles.wiki_id = 1;
```
Done on PR #6260
- [x] **`AverageViewsImporter#update_outdated_average_views` query**
```
# Time: 250319 19:09:17
# User@Host: outreachdashboard[outreachdashboard] @ peony-sidekiq-medium.globaleducation.eqiad1.wikimedia.cloud [172.16.6.194]
# Thread_id: 627  Schema: dashboard  QC_hit: No
# Query_time: 159.906835  Lock_time: 0.000077  Rows_sent: 85  Rows_examined: 15843645
# Rows_affected: 0  Bytes_sent: 10693
SET timestamp=1742411357;
SELECT DISTINCT `articles`.* FROM `articles` INNER JOIN `articles_courses` ON `articles`.`id` = `articles_courses`.`article_id` WHERE `articles_courses`.`course_id` = 22647 AND (`articles`.`average_views_updated_at` IS NULL OR average_views_updated_at < '2025-03-05 19:09:17.908465') ORDER BY `articles`.`id` ASC LIMIT 200;
```
Done on PR #6274

- [ ] **This query seems to be to calculate all the data for the Miscellanea campaign. We should better investigate where it comes from.**
```
# Time: 250324 18:07:04
# User@Host: outreachdashboard[outreachdashboard] @ peony-web.globaleducation.eqiad1.wikimedia.cloud [172.16.6.223]
# Thread_id: 1191  Schema: dashboard  QC_hit: No
# Query_time: 164.954211  Lock_time: 0.000501  Rows_sent: 122535  Rows_examined: 688342
# Rows_affected: 0  Bytes_sent: 396367704
SET timestamp=1742839624;
[...]

PR #6297, opened 2025-04-24
[Data rearchitecture] Re-implement students CSV report without revisions

gabina:
## What this PR does
This PR re-implements `CourseStudentsCsvBuilder` based on article course timeslices instead of using raw revisions. Note that this new implementation is not compatible with historical courses that never ran a timeslice update, meaning that old courses will have empty `total_articles_created` column for CSV reports. We could have both versions coexisting and decide which one use based on the last course update, but I don't think that makes sense if the plan is to remove the revisions table at some point.

It deals with "Download editors data in CSV format" slow query mentioned in https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6267

The CSV report contains the same information as before. The only column that was re-implemented is `total_articles_created` (the one depending on raw revisions). To calculate this column we used to retrieve all course revisions with `new` set to true to identify the article creator id, and use that to count the number of articles created by students. Without using revisions, we had two options: 1) delete `total_articles_created` column or 2) find a way to deduce the article creator for new articles. This PR implements option 2) using a small hack in `ArticleCourseTimeslices`: as ACT records have a `user_ids` array field with a list of all user ids that edited a given article in timeslice, we know that for ACT records with `new_article` true, the article creator id is one of the ids in `user_ids` field. This PR implements a small change in `associated_user_ids` method to guarantee that the article creator is always the first user id in the `user_ids` field. Note that this is done when we still have revisions in RAM. In addition, this PR adds a new ACT `article_creator` method that returns the first user id in `user_ids` when it corresponds to the article creator or nil otherwise (for example, if the timeslice is not for a new article). A new `ArticlesCourses#article_creator` method was also added to be able to retrieve article creator from a given article course. This way the logic of this hack is more encapsulated.

Note that already populated ACT records don't have this guarantee, so `article_creator` makes the best effort to return the user id of the article creator and returns `nil` if that's not possible. Because of that, some report rows may have `total_articles_created` incorrectly set to 0 because it wasn't possible to deduce the article creator in some cases.
Other thing to notice is that `ArticlesCourses#user_ids` method doesn't have the guarantee that the first user id in `user_ids` is the article creator (we could try to do that if we want).

### Example:

[Education and Social Change in the Black Diaspora](https://dashboard.wikiedu.org/courses/Rutgers_University-Newark/Education_and_Social_Change_in_the_Black_Diaspora_(Fall_2023)/home) course.

**Before**: locally generated after running `manual_update`: 
[Rutgers_University-Newark_Education_and_Social_Change_in_the_Black_Diaspora_Revisions_(April_2025)-editors-2025-04-24.csv](https://github.com/user-attachments/files/19898551/Rutgers_University-Newark_Education_and_Social_Change_in_the_Black_Diaspora_Revisions_.April_2025.-editors-2025-04-24.csv)

**After**: locally generated after running `manual_update_timeslice`:
[Rutgers_University-Newark_Education_and_Social_Change_in_the_Black_Diaspora_Timeslices_(April_2025)-editors-2025-04-24.csv](https://github.com/user-attachments/files/19898600/Rutgers_University-Newark_Education_and_Social_Change_in_the_Black_Diaspora_Timeslices_.April_2025.-editors-2025-04-24.csv)

### DIFF
--- a/app/models/article_course_timeslice.rb
+++ b/app/models/article_course_timeslice.rb
@@ -89,6 +89,11 @@ class ArticleCourseTimeslice < ApplicationRecord
     save
   end
 
+  def article_creator
+    return nil unless new_article
+    user_ids.first
+  end
+
   private
 
   # Returns an array containing all the user ids in the given revisions.
--- a/app/models/article_course_timeslice.rb
+++ b/app/models/article_course_timeslice.rb
@@ -89,7 +89,7 @@ class ArticleCourseTimeslice < ApplicationRecord
     save
   end
 
-  # Returns the id of the user that created the article for and ACT record with
+  # Returns the id of the user that created the article for an ACT record with
   # new_article set to true that was populated after CREATOR_FIRST_DEPLOY_DATE.
   def article_creator
     return nil unless new_article && creator_first_ensured?
--- a/app/models/article_course_timeslice.rb
+++ b/app/models/article_course_timeslice.rb
@@ -89,12 +89,22 @@ class ArticleCourseTimeslice < ApplicationRecord
     save
   end
 
+  # Returns the id of the user that created the article for and ACT record with
+  # new_article set to true that was populated after CREATOR_FIRST_DEPLOY_DATE.
   def article_creator
-    return nil unless new_article
+    return nil unless new_article && creator_first_ensured?
     user_ids.first
   end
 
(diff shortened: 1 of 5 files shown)


gabina:
Closing in favor of #6299

PR #6300, opened 2025-04-29
[Data rearchitecture] Add warning when downloading stats for historical courses

gabina:
## What this PR does
This PR adds a warning message to the `CourseStatsDownloadModal` to alert users that want to download CSV reports that never ran a timeslice update.

## Screenshots
Before:
image

After:
image

TODO: revisit `was_course_ever_updated?` method definition. I created this method to identify if the one running is the first course update. However, now I note that it returned `true` for every historical course, because they did have `update_logs` flag for the old system. That makes me think that `was_course_ever_updated?` uses weren't really necessary in `PrepareTimeslices` and `UpdateTimeslicesCourseUser`, but I'll improve this in a separate PR.

### DIFF
--- a/app/assets/javascripts/components/overview/course_stats_download_modal.jsx
+++ b/app/assets/javascripts/components/overview/course_stats_download_modal.jsx
@@ -35,10 +35,16 @@ const CourseStatsDownloadModal = ({ course }) => {
     );
   }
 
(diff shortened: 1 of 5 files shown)


gabina:
@ragesoss I think this is ready for review

Issue #5296, opened 2023-03-13
Mac OS setup script is out of date


ragesoss:
The main python setup script (setup.py) calls different setup scripts, depending on which platform it's being run on. For Mac OS, it calls `setup/macOS-setup.sh`. This script is out of date, and may not successfully result in a working Dashboard development environment without some manual steps.

Run the script on (relatively clean) Mac OS system, post a comment in this issue for anything that you find does not work correctly on the first try without any extra steps, and then update the script to make it work correctly.

lixelamos:
@ragesoss im working on this

sonalyadav1:
"Hello @ragesoss @lixelamos 

I noticed your previous comment on this issue and wanted to check in to see if you’re still actively working on it. If not, I’d be glad to pick it up from here and continue with any progress you may have made. Please let me know—thank you!"

trevorhauter:
@ragesoss @lixelamos @sonalyadav1 I'm assuming that nobody is currently looking at this, so I'm going to begin work on this but feel free to ping me to tell me otherwise! :-)

trevorhauter:
Testing on a fresh macOS installation on an ARM based machine, the first error I'm seeing is
```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /Users/trevorhauter/.rvm/gems/ruby-3.1.2/gems/x25519-1.0.10/ext/x25519_precomputed
/Users/trevorhauter/.rvm/rubies/ruby-3.1.2/bin/ruby -I /Users/trevorhauter/.rvm/rubies/ruby-3.1.2/lib/ruby/3.1.0 -r ./siteconf20250330-91075-cukp7y.rb extconf.rb
mkmf.rb can't find header files for ruby at /Users/trevorhauter/.rvm/rubies/ruby-3.1.2/lib/ruby/include/ruby.h

You might have to install separate package for the ruby development
environment, ruby-dev or ruby-devel for example.
```
It's expecting my header file to be here `/Users/trevorhauter/.rvm/rubies/ruby-3.1.2/lib/ruby/include/ruby.h`, but, the header file is here `/Users/trevorhauter/.rvm/rubies/ruby-3.1.2/include/ruby-3.1.0/ruby.h` 

I'm not sure if this is out of scope or not, since having ruby installed is a prereq to the setup script, but it seems likely that an extra step  is required to get this working.

[...]

PR #6309, opened 2025-05-03
Fix: macOS setup script

trevorhauter:
This is a small change to bring the macOS setup script up to date: https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/5296

On a fresh installation of macOS I am now able to get a working dashboard using the setup script on this branch!

## What this PR does
- Add compatibility flag to bundler in macOS setup script
    - This adds a compatibility flag to resolve errors like the following
        - ```sh
            compiling selector.c selector.c:301:26: error: incompatible function pointer types passing 'VALUE (*)(VALUE *)' (aka 'unsigned long (*)(unsigned long *)') to parameter of type 'VALUE (*)(VALUE)' (aka 'unsigned long (*)(unsigned long)') [-Wincompatible-function-pointer-types]```
        - Source: https://stackoverflow.com/questions/78313570/unable-to-bundle-install-nio4r-on-a-new-mac-m3

- Unset environment variable `SDKROOT`, which is set by apples xcode tools when the setup script is invoked from python3. This is a bug that exists in our current version of Ruby. Ruby version 3.2.1 is the first version released with a fix for this bug, but this work around suffices for now!
    - sources:
        - https://github.com/rbenv/ruby-build/discussions/2123
        - https://bugs.ruby-lang.org/issues/19403
        - https://stackoverflow.com/questions/46377667/docker-for-mac-mkmf-rb-cant-find-header-files-for-ruby

### DIFF
--- a/setup/macOS-setup.sh
+++ b/setup/macOS-setup.sh
@@ -106,6 +106,7 @@ else
 fi
 
 printf '[*] Installing Gems... \n'
+output_line "bundle config build.nio4r --with-cflags='-Wno-incompatible-pointer-types'"
 output_line "bundle install" && print_success "${CLEAR_LINE}[+] Gems installed\n"
 
 printf '[*] Checking for application configurations... \n'
--- a/setup/macOS-setup.sh
+++ b/setup/macOS-setup.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+unset SDKROOT
+
 . "$(dirname "$0")/color-helpers.sh"
 
 clear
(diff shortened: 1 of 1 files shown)


ragesoss:
Thanks @trevorhauter! I haven't tested this, but it looks reasonable to me. I'm happy to merge this now, but it might be better to add additional commits as you make progress, and merge them altogether. If that's what you'd like to do, you can turn this into a draft until it includes all the fixes you want to me. Let me know which way to go.

trevorhauter:
@ragesoss Thanks for taking a look! Good shout, I think it does make sense to ship everything together. I've converted this back to a draft!

ragesoss:
Thanks! These changes look reasonable; since you were able to use the script on a fresh macOS, I'm cool with merging it.

PR #6315, opened 2025-05-06
[Data rearchitecture] Update MonthlyReport and CourseStatistics classes to work without revisions table

### LINKED ISSUE #6311: [Data-rearchitecture] Reimplement `MonthlyReport` and `CourseStatistics` classes to work without revisions table
As part of the clean-up process, we noticed (thanks to failing specs) that there are still some parts of the code that depend on Revisions table and need to be re-implemented.

- [x] `AnalyticsController#monthly_report`: it uses `MonthlyReport` class.
- [x] `AnalyticsController#campaign_stats` and `AnalyticsController#campaign_intersection`: they use `CourseStatistics` class

Both classes have to be updated.

gabina:
## What this PR does
This PR updates (hopefully) the last two classes that still depended on revisions table:
- `MonthlyReport` 
- `CourseStatistics`

Both clases are used through the analytics controller.

Closes #6311

### After
I don't put a "before" image because the numbers are pretty different given all the local courses I have on my computer (and I don't want to wipe my db).

image

## Screenshots CourseStatistics
Note that `revisions after = (revisions before) * 2`. This is because without having `mw_page_id` we have no way to determine if we're counting the same revisions for different courses. This behavior is the same as we see in `references_added` . While we see 28 `references_added`, there were only 14 references added in the campaign (but two courses have the same exact new references). The duplication of references added is a preexisting problem since they were calculated based on `courses_users` and not based on revisions (see commit 4f5444a). Same thing happens for characters added/words added.

### After
image

### Before
image

Note that while `MonthlyReport.run` class method has `opts={}` parameter, we never used it?

There is room for improvement in the `MonthlyReport` specs, however, since there were no previous specs to adapt, I simply created some basic tests.

There is also room for improvement in the `CourseStatistics` specs, making notorious the unexpected behaviors in terms of duplicated metrics mentioned in this PR.

This PR adds a new Course method called `tracked_article_course_timeslices`. I did not do an exhaustive search in the code for places where it might be possible to use this method instead of calculating it manually.

This PR deletes dead code in `AnalyticsController`.

### DIFF
--- a/app/controllers/analytics_controller.rb
+++ b/app/controllers/analytics_controller.rb
@@ -28,8 +28,6 @@ class AnalyticsController < ApplicationController
   def results
     if params[:monthly_report]
       monthly_report
-    elsif params[:campaign_stats]
-      campaign_stats
     elsif params[:campaign_intersection]
       campaign_intersection
     end
@@ -107,16 +105,6 @@ class AnalyticsController < ApplicationController
 
   def monthly_report
     @monthly_report = MonthlyReport.run
-  end
-
-  def campaign_stats
-    @campaign_stats = {}
-    @articles_edited = {}
-    Campaign.all.each do |campaign|
-      course_ids = campaign.courses.pluck(:id)
-      stats = CourseStatistics.new(course_ids, campaign: campaign.slug)
-      @campaign_stats.merge! stats.report_statistics
-    end
   end
 
   def campaign_intersection
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -334,7 +334,12 @@ class Course < ApplicationRecord
   end
 
   def tracked_article_course_timeslices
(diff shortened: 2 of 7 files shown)


app/controllers/analytics_controller.rb
```diff
@@ -28,8 +28,6 @@ def index; end
   def results
     if params[:monthly_report]
       monthly_report
-    elsif params[:campaign_stats]
-      campaign_stats
     elsif params[:campaign_intersection]
       campaign_intersection
```
gabina, line 32: The `elseif` could be removed and only keep an `else` clause (since the only option if `params[:monthly_report]` is false is that `params[:campaign_intersection]` is true). However, removing the `elseif` would make the `params[:campaign_intersection]` less explicit.

gabina:

PR #6319, opened 2025-05-13
[WIP] [Data-rearchitecture] Remove dependency on UpdateWikidataStats in AnalyticsController spec

### LINKED ISSUE #6305: Make `AnalyticsController#course_wikidata_csv` spec not dependent on `UpdateWikidataStats` class
`UpdateWikidataStats` class will be deleted soon since it's an old class that works with revisions in the db (something that is no longer supported). The class was already replaced by `UpdateWikidataStatsTimeslice`  (see PR #6059). Because of that, we have to update `AnalyticsController#course_wikidata_csv` spec so that it no longer uses the old class . See `spec/controllers/analytics_controller_spec.rb` file).

This issue is related to #6303.

gabina:
## What this PR does
Closes #6305

### DIFF
--- a/spec/controllers/analytics_controller_spec.rb
+++ b/spec/controllers/analytics_controller_spec.rb
@@ -105,13 +105,7 @@ describe AnalyticsController, type: :request do
   end
 
   describe '#course_wikidata_csv' do
-    let(:wikidata) { Wiki.get_or_create(language: nil, project: 'wikidata') }
-    let(:course) { create(:course, slug: 'foo/bar_(baz)', home_wiki: wikidata) }
-
-    before do
-      stub_wiki_validation
-      UpdateWikidataStats.new(course)
-    end
+    let(:course) { create(:course, slug: 'foo/bar_(baz)') }
 
     it 'returns a CSV' do
       get '/course_wikidata_csv', params: { course: course.slug }


spec/controllers/analytics_controller_spec.rb
```diff
@@ -105,13 +105,7 @@
...
 
   describe '#course_wikidata_csv' do
-    let(:wikidata) { Wiki.get_or_create(language: nil, project: 'wikidata') }
-    let(:course) { create(:course, slug: 'foo/bar_(baz)', home_wiki: wikidata) }
-
-    before do
-      stub_wiki_validation
-      UpdateWikidataStats.new(course)
-    end
+    let(:course) { create(:course, slug: 'foo/bar_(baz)') }
```
gabina, line 108: There is no a trivial way to replace the use of  `UpdateWikidataStats.new(course)` with the new `UpdateWikidataStatsTimeslice`. However, I think there is no need for calling  `UpdateWikidataStatsTimeslice` class for this spec, as the only thing that it checks it's that the body includes 'total revisions', which apparently is handled by `CourseWikidataCsvBuilder`, independently on the values of the CSV. In addition, it looks like it doesn't matter if the course has `wikidata` as a tracked wiki, since the headers are always included in the CSV.

PR #6304, opened 2025-05-01
[Data rearchitecture] Clean up unused old classes - Part 1

gabina:
## What this PR does
Work in progress. This PR handles a good number of points mentioned in issue #6303 :

- Delete `IndividualStatisticsPresenter` class as this was replaced by `IndividualStatisticsTimeslicePresenter`. Used for user profile.
- Delete `RevisionStat` class as this was replaced by `RevisionStatTimeslice`. Used in `CoursesUsers#update_cache` and `CourseCacheManager#update_recent_revision_count`.
- Delete `ArticlesCoursesCleaner` class as this was replaced by `ArticlesCoursesCleanerTimeslice`.
- Delete old `article.rake` task and its specs because it depends on  `ArticlesCoursesCleaner`. See commit [07ade1b](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6304/commits/07ade1b7623b225fd0a828a167ab7588e74a264e).
- Delete `ArticleStatusManager` class as this was replaced by `ArticleStatusManagerTimeslice`. It's still used in `DuplicatePageIdHandler` class, `duplicate_articles` cleanup script and `UpdateCourseStats#update_article_status`.
- Delete `DuplicatePageIdHandler` class and `duplicate_articles` cleanup script. See commit  
- Delete `UpdateWikidataStats` class as this was replaced by `UpdateWikidataStatsTimeslice`.
- Delete `UpdateWikidataStatsWorker` as it only worked for `UpdateWikidataStats`.
- ~Comment `AnalyticsController` spec that depends on `UpdateWikidataStats`. See issue #6305 .~ Edit: PR for that issue was already merged.
- Delete `UpdateWikiNamespaceStats` class as this was replaced by `UpdateWikiNamespaceStatsTimeslice`.
- Delete `RevisionImporter` class. It's used in `CourseRevisionUpdater` class.
- Delete `ModifiedRevisionsManager` class.
- Delete `UpdateCourseStats` class as this was replaced by `UpdateCourseStatsTimeslice`.
- Delete `DuplicateArticleDeleter#resolve_duplicates` method and `ModifiedRevisionsManager` dependency in `CourseRevisionUpdater` (commit 176a696f2faa0c6801ad99354c7c2d9881729a2c). Delete specs for `DuplicateArticleDeleter#resolve_duplicates` (commit 3bb22431e7df6e2ce6e213e47db4f497d2d1315d)
[...]

PR #6324, opened 2025-05-16
[Data rearchitecture] Clean up unused old classes - Part 2

gabina:
## What this PR does
This PR starts from PR #6304 and continue the cleaning-up process.

- Remove unused `Course.update_all_caches` class method.
- Remove `Course.update_cache` and `CourseCacheManager.update_cache` methods.
- Remove `ArticlesCourses.update_all_caches` class method and specs for it.
- Remove `ArticlesCourses.update_from_course` class method and specs for it.
- Remove `ArticlesCourses#update_cache` instance method.
- Remove revision scopes from `Course` class
- Remove revision scopes from `ArticleScopedProgram` and `VisitingScholarship` classes
- Remove revision scopes from `User` class
- Remove revision scopes from `Article` class
- Remove `all_revisions` from `ArticlesCourses` class
- Remove revision-dependent scripts: `docs/analytics_scripts/campaign_edit_tracking.rb`, `docs/analytics_scripts/ores_changes.rb` and `docs/cleanup_scripts/fix_crosswiki_revisions.rb`
- Remove unused methods from `RevisionScoreImporter`: `mainspace_userspace_and_draft_revisions`, `unscored_revisions` and `unscored_previous_revisions`.
- Define sandboxes course scope through ACT instead of revisions (commit 5268c27bde6fc360a52f6aba99d09049f4c96f50)
- Define new `course_user_wiki_timeslices` user scope to handle duplicate users (commit b309eaa4f991438436250d1639403bd3e2ec71d9)
- Comment lines depending on revisions in `lib/analytics/ores_diff_csv_builder.rb` (commit fa387a2104907f60430c4e4aea06feadcb6edb87). The ores plot was already disabled from frontend (PR #6328), and there is an open issue for re-implementing (issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6327). I didn't remove all the code associated to it because I think it will be re-used when re-implementing.


gabina:

app/models/course_data/articles_courses.rb
```diff
@@ -42,54 +42,21 @@ class ArticlesCourses < ApplicationRecord # rubocop:disable Metrics/ClassLength
   # Instance methods #
   ####################
   def view_count
-    update_cache unless self[:view_count]
     self[:view_count]
```
ragesoss: I think these methods can be deleted entirely, although it's also fine to leave them. Now that they simply return the value from the database, the default Rails behavior will work.

gabina: ah yes, I see. Although it looks like we repeat that pattern sometimes, for example for  `new_article`, or even for `assigned_article_title` in courses_users.

```
  def new_article
    self[:new_article]
  end
```

Do you want me to remove all those methods?

ragesoss: Sure, I think it makes sense to remove them.

ragesoss:
This all looks good to me.

app/models/course_data/articles_courses.rb
gabina: Done on 310aa2343259edf3a44e742594a6f0c364297800

Issue #6327, opened 2025-05-17
Ores plot is empty in timeslices system


gabina:
### What is happening?
While we're importing scores and calculating `wp10` using the `weighted_mean_score` definition through the `RevisionScoreImporter` class, we're not actually persisting that value in any place. That means that all the places where `wp10` is used are not working as expected on the new system.

In particular, `ores_data.json` returns zeroed `ores_before` and `ores_after` values for ores plot.

![Image](https://github.com/user-attachments/assets/203de073-f2f6-4b93-bfd9-b074be8494e4)

### To Reproduce

> Steps to reproduce the behavior:
> 1. Go to [250_by_2026 campaign](https://dashboard.wikiedu.org/campaigns/250_by_2026/overview)
> 2. Click on ORES tab
> 3.  Click on "Changes in Structural Completeness"
> 4. See the chart with empty mean
![Image](https://github.com/user-attachments/assets/3e130147-e2ed-4b97-9abe-9bc779990da4)

### Expected behavior
See a complete ores plot

![Image](https://github.com/user-attachments/assets/85a4fcd1-f3cd-465d-8ae1-bbb77aecd6c3)

### Additional context
This bug comes from the migration to the new timeslices system. Now that we have no more revisions in the database, the value of `wp10` is not persisted.

ragesoss:
I guess we should disable this feature for now, but I think it will be feasible to re-implement at some point. It's not a high priority.

gabina:
I'll try to disable the feature by setting `ORESSupportedWiki` to no projects to avoid commenting/removing code that we may need when re-implementing. Does it make sense for you?

PR #6328, opened 2025-05-19
Disable ores plot feature temporarily

gabina:
## What this PR does
This PR disables ores plot feature for every campaign and course language/project until re-implementing the feature for the new system. Read issue #6327

## Screenshots
Before:
**Ores plot for campaigns**
campaign_before

**Ores plot for courses**
course_before

After:
**Ores plot for campaigns**
campaign_after

**Ores plot for courses**
course_after


gabina:
@ragesoss I think this is ready for review. It makes the minimum changes to disable the ores feature in the frontend side. I also forced the failing spec to pass.

PR #6329, opened 2025-05-20
[WIP]: Improve performance of slow routes: Optimize survey results data loading

### LINKED ISSUE #6316: Improve survey results page load time
**Related to** #984

This performance issue appears to be caused by the following factors:

---

### 1. Deep Eager Loading in `result.html.haml`

```ruby
@surveys_question_groups.includes(rapidfire_question_group: { questions: { answers: { answer_group: :user }}})
```

This query likely causes performance issues due to excessive nesting in the eager loading. The depth of associations can lead to large and complex SQL queries, high memory usage, and significant overhead when Active Record builds the object graph. Splitting this query into smaller, more targeted queries may help reduce memory usage and improve load times.

**2. N+1 Query Pattern**

There are potential N+1 queries being triggered in the following methods:

- **surveys_helper.rb**
```ruby
def question_group_locals(surveys_question_group, index, total, is_results_view:)
```

- **surveys.rb**
```ruby
def notification(survey_id, curr_user = nil)
```
These methods may be indirectly triggering multiple queries per iteration when accessing associated records.

Abishekcs:
Closes #6316

# Survey Results Performance Optimization

## Overview
This PR significantly improves the performance of survey results pages, reducing load times by less than <1000ms for `radio, checkbox, select and rangeinput` question type and less than <2000ms for `text and long` question type  through database query optimization and data loading refactoring.

### Performance Bottleneck
The survey results pages were experiencing severe performance issues due to inefficient data loading patterns. The primary culprit was this problematic query:

```ruby
@surveys_question_groups.includes(rapidfire_question_group: { questions: { answers: { answer_group: :user}}})
```

**Issues with the original approach:**
- **Deep nested includes**: Created complex `joins across 5+ tables` simultaneously this approach works great for small   
data set of responses but if not it completely breaks down in speed for  larger data set.
- **Over-fetching**: Loaded entire model objects when only specific columns were needed
- **Memory intensive**: Large nested object graphs consumed excessive memory
- **N+1 queries**: Despite includes, some queries still triggered additional database calls
- **Inefficient data structure**: Data wasn't properly indexed for view consumption

### New Data Loading Strategy
Replaced the monolithic query with a strategic, step-by-step data preparation approach:

```ruby
def prepare_survey_data
  load_question_groups_for_survey    # Load question groups with position order
  load_questions                     # Index questions by ID for fast lookup
  load_and_index_answers_with_counts # Group answers and calculate counts
  build_answer_group_builders        # Initialize form builders
  build_answer_group_user_cache      # Cache users to prevent N+1
  survey_courses_cache              # Attach course data efficiently
end
```

#### 1. **Selective Column Fetching** 
- **Example**: `SELECT id, username FROM users` and `SELECT id, title FROM courses`
- **Impact**: Reduced data transfer and memory usage

#### 2. **Data Indexing**  (Few examples)
```ruby
# Questions indexed by ID for O(1) lookup
@rapidfire_questions_by_id = {}

# Answers grouped by question ID
@rapidfire_answers_by_question_id = Hash.new { |h, k| h[k] = [] }

# Answer counts per question
@question_answers_count = Hash.new(0)
```

#### 3. **N+1 Query Prevention** 
- Built user cache with preloaded data
- Eliminated redundant database calls in view rendering
- Used `find_each` for memory-efficient iteration over large datasets

#### 4. **Course Data Caching**
- Dynamically attaches course data to users via singleton methods
- Eager loads campaigns and tags in a single query
- Provides easy access pattern: `user.survey_course`

## Screnshots:

**Before:**

Web_Photo_Editor

https://github.com/user-attachments/assets/47098473-828b-4d2e-ad41-79917e9eec0a

**After:**

`Note: If I don't use a screen recording tool like OBS, the page usually loads in under 1 second.`

Web_Photo_Editor

https://github.com/user-attachments/assets/1b3e5c4d-937f-4aa2-af09-9dd0ea971612

### DIFF
--- a/app/controllers/surveys_controller.rb
+++ b/app/controllers/surveys_controller.rb
@@ -18,7 +18,7 @@ class SurveysController < ApplicationController
     results
   ]
   before_action :ensure_logged_in
-  before_action :set_question_groups, only: %i[
+  before_action :prepare_survey_data, only: %i[
     show
     edit
     edit_question_groups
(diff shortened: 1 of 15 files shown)


app/controllers/surveys_controller.rb
```diff
@@ -18,7 +18,7 @@ class SurveysController < ApplicationController
     results
   ]
   before_action :ensure_logged_in
-  before_action :set_question_groups, only: %i[
+  before_action :prepare_survey_data, only: %i[
```
ragesoss: Since this action applies to `show`, does that mean it will do extra work related to answer counts (and possibly other things) when loading a survey that a user is taking (rather than when an admin is navigating the survey admin UI)?

Abishekcs: Yes, I just tested it. I think an if check should work for this solution?

Abishekcs: The current implementation of the  Programs & Events Dashboard—not this PR—takes around 2000ms for the `show` action. The version in this PR also takes roughly the same time (i.e., 2000ms). I think adding those if/unless checks might significantly `decrease` the load time of `show` as well.

ragesoss: rather than if/unless checks, would be possible to break `prepare_survey_data` into two separate methods, and have both of them run for the admin-facing actions but only one run for `show`? That would make it easier to understand, when looking at the controller.

ragesoss:
On a first pass, this looks like a good set of performance optimizations, but it's a lot to think through at once. However, since it is focused on the admin results views, I think it's relatively low risk despite the large number of files involved.

Especially with this performance work, I think it's a good idea to make each PR as small as possible, so if there are multiple  optimizations that you want to implement, isolating each one to its own PR is ideal. (I'm fine with using this PR, although if separating it out into smaller PRs for each of the optimizations would be easy for you, that's cool too.)

Abishekcs:
> I think it's a good idea to make each PR as small as possible

That works great for me too! I’ll go through the code once again and make improvements wherever possible

app/controllers/surveys_controller.rb
Abishekcs: > rather than if/unless checks, would be possible to break `prepare_survey_data` into two separate methods, and have both of them run for the admin-facing actions but only one run for `show`? That would make it easier to understand, when looking at the controller.

l will take another look at the code with these points in mind and share an update once I’ve have gone  thorough it again.

ragesoss: if/unless checks are acceptable, but hopefully not necessary with the right refactoring.

Abishekcs:
> would be possible to break `prepare_survey_data` into two separate methods, and have both of them run for the admin-facing actions but only one run for `show`? That would make it easier to understand, when looking at the controller.

Hi @ragesoss

I’ve refactored the `prepare_survey_data` method to use a `case` statement, so it now loads only the necessary data depending on the context (`admin`, `show`, `edit`, or `edit_question_groups`). This should also add in improve performance and clarity.

Below are the **before and after** screen recordings for each view. I used OBS to record, which adds a bit of overhead, so I’ve also included **screenshots** for a more accurate comparison of response times.!


### **Before**:
---


**_show view:_**
---


Before show view



https://github.com/user-attachments/assets/17ab9e0c-e7f1-4de6-9b80-80ab512f2f5e

**_results view_**
---

Before admin results view


https://github.com/user-attachments/assets/3f4ca740-20bb-47c6-b3f1-6f7b510594cb



**_edit view_**
---

Before edit view


https://github.com/user-attachments/assets/11dd5896-902e-4c8a-bd2e-b2ad7f3af6bb



**_edit_question_groups view_**
---


edit_question_group_view



https://github.com/user-attachments/assets/195d8ea0-1d1f-4823-9044-fbc1e4c95b3a



### **After**:
---

**_show view:_**
---

show view

[https://github.com/user-attachments/assets/0de14116-d12e-4614-adc9-dd3d000e9360](https://github.com/user-attachments/assets/0de14116-d12e-4614-adc9-dd3d000e9360)

**_results view_**
---

results admin view

[https://github.com/user-attachments/assets/4700f077-330a-4383-990f-8d997eac98d6](https://github.com/user-attachments/assets/4700f077-330a-4383-990f-8d997eac98d6)

**_edit view_**
---

edit view

[https://github.com/user-attachments/assets/740c85c4-dee2-4eb7-ac18-3670a955aece](https://github.com/user-attachments/assets/740c85c4-dee2-4eb7-ac18-3670a955aece)

**_edit_question_groups view_**
---

edit\_question\_groups view

[https://github.com/user-attachments/assets/b81334d5-fca7-474d-997d-8153d466484b](https://github.com/user-attachments/assets/b81334d5-fca7-474d-997d-8153d466484b)

---

app/controllers/surveys_controller.rb
```diff
@@ -151,10 +151,143 @@ def set_survey
...
   end
 
-  def set_question_groups
-    @question_groups = Rapidfire::QuestionGroup.all
+  # Loads and caches survey data based on the specific action's requirements.
+  # Optimizes performance by loading only necessary data for each action:
+  # - 'show': Basic data for survey display (avoids expensive answer processing)
+  # - 'results': Full dataset including answer counts and user data for admin analysis
+  # - 'edit'/'edit_question_groups': Minimal data for survey editing interface
+  def prepare_survey_data
```
ragesoss: Nice job separating these based on which actions need which data. I think it makes more sense to refactor them a bit further so that there are several different `before_action` methods that each get called from the actions that require them, rather than using a singe method that uses `case` to change behavior for different actions. The `before_action` Rails method (with `only:` and `except:` options) is designed for exactly this sort of situation.

app/views/rapidfire/answer_groups/new.html.haml
```diff
@@ -4,7 +4,7 @@
       Preview Mode
       = link_to t('question_groups.edit_question_group'), edit_question_group_path(@question_group), class: "button"
 .loading{"data-survey-form-container" => ""}
-  = render partial: "surveys/question_group", locals: { |
+  = render partial: "surveys/preview_question_group", locals: { |
```
ragesoss, line 7: Is this correct? It looks like this part of the template is not dependent on the `preview` param.

Abishekcs: hmm 😅 if i understand correctly this is for  the route as shown below

https://github.com/user-attachments/assets/a3e7d4d3-e58e-407e-a4f7-221909406c95

Abishekcs: I created a new file:
`app/views/surveys/_preview_question_group.html.haml`

This was to avoid changing the existing `surveys/question_group` partial . I had added some logic there for showing survey results to admins, but that ended up breaking the preview view. So instead of trying to handle both cases in one file (which made things complicated), I made a copy without the results logic just for the preview. It was the easier for me that way  😅

app/controllers/surveys_controller.rb
Abishekcs: Okay!

Abishekcs: > I think it makes more sense to refactor ...

Pushed the necessary code for this just now.

ragesoss:
@Formasitchijoh would you like to take a pass at reviewing this one?

Abishekcs:
Closing this PR as this PR as been split into #6375, #6383, #6385 and #6386

Issue #6321, opened 2025-05-15
On Wiki Education Dashboard, require both semester and year in Term field for new courses and cloned courses


ragesoss:
Many courses get created with an ambiguous value for the 'Term' field, such as "summer".

The validations used in the CourseCreator and the CourseClonedModal should prevent saving a course if the term field does not include at least two digits (ie, a value that likely indicates the year), and the error message should include an example like "Spring 2025".

The placeholder text should be changed from "Term" to "Term (including year)".

swayam-agrahari:
Hello, I would like to work on this issue.

ragesoss:
@swayam-agrahari go for it!

swayam-agrahari:
How should I handle the placeholder for the word as it is not present in any of the localization files ?

ragesoss:
You can add a new entry in the en.yml file with the new placeholder text, as well as an entry in the qqq.yml file to a description of what that message is used for.

ragesoss:
Resolved with https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6344

PR #6336, opened 2025-05-28
[WIP] Enforce Year Requirement in Term Field for Course Creation and Cloning 

swayam-agrahari:
## What this PR does
This pull request adds a validation rule to ensure that the **Term** field includes a year (at least two digits) when creating or cloning a course. This aims to fix [#6321](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6321), which aims to prevent ambiguous term values such as "Summer" without a year.

- Adds frontend validation in both `CourseCreator` and `CourseClonedModal`.
- Updates placeholder text from `"Term"` to `"Term (including year)"`.
- Shows an inline error message prompting for values like `"Spring 2025"`.

## Screenshots
Before:
Screenshot From 2025-05-28 20-56-51

After:
Screenshot From 2025-05-28 20-54-42


ragesoss:
It looks like the tests are likely failing because of the changes you made to the course form. You should be able run those failing tests locally as well, and if you remove the 'headless' option in the Capybara config in `rails_helper.rb`, you'll be able to watch the tests run in the browser. You'll likely need to make some adjustments to the test to match the new version of the form.

PR #6340, opened 2025-06-01
Minor: Skip redundant API calls on /survey routes to improve minor survey page performance

Abishekcs:
Found this while working on PR #6329 

## What this PR does

This PR disables the API call to `/requested_accounts.json`  and `/faq/Programs_&_Events_Dashboard_News/handle_special_faq_query` when the current route starts with `/survey`. These routes are admin-only and are structured with full-page reloads for each tab, which was causing the API request to be triggered repeatedly—even though most of  the time data is not needed on these pages.

### Reason for Change
- The `/requested_accounts.json` and `/faq/Programs_&_Events_Dashboard_News/handle_special_faq_query` endpoint provides account requests and unread news notifications count, which are mostly relevant outside of  the survey context.
- Admins can still access updated data by navigating to the homepage or other relevant views.
- Preventing this call on `/survey` pages improves performance `(not dramatically noticeable, but both endpoints consistently add ~100ms each)` and reduces unnecessary backend traffic.
- Avoids polluting logs and metrics with redundant, contextually irrelevant requests.

### Implementation Details
- Added a pathname check using `location.pathname.startsWith('/survey')` to conditionally skip the request.
- Added a comment explaining the logic for future contributors.

No functionality is lost for admins, and the user experience remains fully intact.

## Screenshots
Before:

https://github.com/user-attachments/assets/b31be780-7092-4ce7-9168-e3c27473cc2c

After:

https://github.com/user-attachments/assets/7daf2222-9154-43f7-b279-2de493589578

### DIFF
--- a/app/assets/javascripts/components/nav/news/news_nav_icon.jsx
+++ b/app/assets/javascripts/components/nav/news/news_nav_icon.jsx
@@ -23,7 +23,14 @@ const NewsNavIcon = ({ setIsOpen }) => {
   useEffect(() => {
     const fetchNewsData = async () => {
       try {
-        const newsContentList = await API.fetchNews(newsType);
+        let newsContentList = [];
+
+        // Skip fetching account requests if on a survey page.
+        // Survey tabs cause full page reloads, which would trigger unnecessary API calls.
+        // Admins can view updated requests/notifications from the homepage instead.
+        if (!location.pathname.startsWith('/survey')) {
+          newsContentList = await API.fetchNews(newsType);
+        }
 
         // Retrieve the last fetch timestamp from the cookie
         const userLastFetchTimestamp = Cookies.get(`lastFetchTimestamp_${newsType}`);
--- a/app/assets/javascripts/components/nav/notifications_bell.jsx
+++ b/app/assets/javascripts/components/nav/notifications_bell.jsx
@@ -15,10 +15,15 @@ const NotificationsBell = () => {
         .catch(err => err);
     }
 
-    request('/requested_accounts.json')
-      .then(res => res.json())
-      .then(({ requested_accounts }) => setHasRequestedAccounts(requested_accounts))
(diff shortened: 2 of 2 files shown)


app/assets/javascripts/components/nav/news/news_nav_icon.jsx
```diff
@@ -23,7 +23,14 @@ const NewsNavIcon = ({ setIsOpen }) => {
   useEffect(() => {
     const fetchNewsData = async () => {
       try {
-        const newsContentList = await API.fetchNews(newsType);
```
ragesoss, line 26: Hmm... This seems like a very particular set of fixes.

I think the underlying issue is that the behavior for fetching news and items for the notification bell are not dynamic enough, but I wonder if there's a better place to make this sort of change, or a broader strategy to make the requests only when needed. @Formasitchijoh any ideas?

Abishekcs: How about setting a timer so that if a request is made within 5 seconds (like when switching between tabs), it doesn't sent a request? RTK Query provides this feature, but I don't think it's necessary to use RTK Query for such a small change.

ragesoss: Hmm... maybe.

Since this is very minor from a performance perspective, I suggest that we open a pair of issue for improving the behavior of each of these API calls to figure out when it's appropriate to make the call vs. not.

For the notification bell, there's a related issue to include as well: for the open tickets notification, it needs to re-fetch that data whenever the user interacts with the status of a ticket (for example, after replying to a ticket, if it was that user's last open ticket, then the red icon should disappear, but currently it remains in place until a refresh, even when the page navigates back to the ticket index).

Abishekcs: Okay, I’ll look into this further and provide an update. Closing this PR for now.

PR #6343, opened 2025-06-04
[WIP]: Optimize Survey#to_csv by Eliminating N+1 Queries and Adding Foreign Key to SurveyNotification

Abishekcs:
TO BE UPDATED

## What this PR does

## Screenshots
Before:

After:


Abishekcs:
@ragesoss @Formasitchijoh This PR is ready for review. Please take a look whenever you have time. Thank you!

app/models/surveys/survey.rb
```diff
@@ -45,22 +45,29 @@ def status
...
         csv << response_row
       end
     end
   end
 
   private
 
+  def prepare_data_for_csv
+    respondents
+    course_for
```
ragesoss: These method names are not very clear about what they are for or what they do.

ragesoss:
Where is the N+1 this fixes, and what do the N+1 queries look like?

This seems like quite a bit of complex code, so I suspect it can be refined to be more concise.

Abishekcs:
> Where is the N+1 this fixes, and what do the N+1 queries look like?


In `course_for`


Screenshot from 2025-06-04 21-56-32




Screenshot from 2025-06-04 21-55-24


https://github.com/user-attachments/assets/825a14c6-fdf1-478a-b939-030163c5a8b2

app/models/surveys/survey.rb
Abishekcs: Okay, I’ll do that. Thank you for the review!

Abishekcs:
> This seems like quite a bit of complex code, so I suspect it can be refined to be more concise.

Okay

app/models/surveys/survey.rb
```diff
@@ -86,13 +93,54 @@ def response(user)
...
+
+  def load_courses_by_id(course_ids)
+    Course.where(id: course_ids).select(:id, :slug).index_by(&:id)
+  end
+
+  def assign_survey_courses_to_users(notifications_by_user, courses_by_id)
+    notifications_by_user.each do |user_id, notification|
+      user = @users_by_id[user_id]
+      course = courses_by_id[notification.course_id] || fallback_course_for_user(user_id)
+      user.define_singleton_method(:survey_course) { course }
```
ragesoss: I think I understand what is going on here with `define_singleton_method`, but there must be a more elegant way to insert the course into the `notifications_by_user` data structure.

Some comments explaining the structure of this object (probably above `fetch_notifications_by_user`, since that's where the structure is created) would be helpful.

Abishekcs: Okay got it. Thank you!

app/models/surveys/survey.rb
```diff
@@ -86,13 +93,54 @@ def response(user)
...
+  def course_for
+    load_survey_assignment_ids
+    notifications_by_user = fetch_notifications_by_user
+    course_ids = notifications_by_user.values.map(&:course_id).uniq
+    courses_by_id = load_courses_by_id(course_ids)
+    assign_survey_courses_to_users(notifications_by_user, courses_by_id)
+  end
+
+  def load_survey_assignment_ids
+    @survey_assignment_ids = SurveyAssignment.where(survey_id: id).pluck(:id)
```
ragesoss: Since this in a `Survey` instance variable, the idiomatic Rails way to do this to use conditional assignment, something like...

```ruby
def survey_assignment_ids
  @survey_assignment_ids ||= survey_assignments.pluck(:id)
end
```

Then you can just call it from whichever methods need the ids (in place of the instance variable), without having to explicitly call `load_survey_assignment_ids`

Abishekcs: Okay, thank you! I was a bit confused about that—this really clears things up.

app/models/surveys/survey.rb
Formasitchijoh: Maybe using hash map to associate each `user_id` with the appropriate `course`, and then accessing the course for each user in the `to_cvs`

Abishekcs:
### Update Changes

- **Data Preparation Refactor:**
  -  Introduced `prepare_data_for_csv`, which initializes all required data before generating the CSV, avoiding repeated DB  queries.

  -  Replaced unclear method names like `respondents` and `course_for` with explicit, purpose-driven methods like 
     `survey_responding_users_by_id` and `course_slug_for_user.`

- **Improved Rails Conventions:**
   - Updated `survey_assignment_ids` to use conditional memoization following idiomatic Rails style for instance variable caching

- **More Efficient Course Mappings:**
  -  Replaced dynamic method definitions (`define_singleton_method`) with a cleaner approach using hash maps.
  -  Constructed a course slug lookup using `course_slug_for_user` and handled missing courses with a fallback (`fallback_course_slug_for_user`).

- **Commentary & Clarity:**
  -  Added inline comments explaining data structures like `notifications_by_user_id`, making it easier to understand object shapes and intentions.

### `prepare_data_for_csv` Call Flow

This method orchestrates the loading of all necessary data for efficient CSV generation:

1. **`survey_assignment_ids`**  
   Loads all survey assignment IDs for the current survey instance.

2. **`survey_responding_users_by_id`**  
   Fetches users who responded to the survey, indexed by their user ID.

3. **`notifications_by_user_id`**  
   Retrieves the primary `SurveyNotification` for each responding user.

4. **`courses_by_id`**  
   Loads course records relevant to the notifications.  
   This involves two sub-steps:
   
   4.1. **`notification_course_ids`**  
   Extracts unique course IDs from the notifications.
   
   4.2. **`load_courses_by_id(course_ids)`**  
   Fetches course details and indexes them by ID.


### **Peformance Metric:**


> **Note:** The screenshot below highlights the log for  
> `response_row = [user.username, course_slug_for_user(id)]`  
> while intentionally omitting `response(user)`, as that part is **not addressed in this PR**.  
> 
> The `response(user)` method still contains multiple N+1 queries, which remain a **major performance bottleneck** in the CSV export process which will be resolved in a separate PR.

**Before:**

Before Code
Before Screenshot


**After:**
After Code
After Code

ragesoss:
For this, I'd like you to go through it with same strategy I encouraged for the other PR: see if you can find good-enough simpler optimizations that are based around just tweaking existing queries with `includes` before turning to the kinds of complex queries that return raw data structures. I'd really like to see each of these open PRs be as small as possible.

Abishekcs:
> For this, I'd like you to go through it with same strategy I encouraged for the other PR: see if you can find good-enough simpler optimizations that are based around just tweaking existing queries with `includes` before turning to the kinds of complex queries that return raw data structures. I'd really like to see each of these open PRs be as small as possible.

Okay

Abishekcs:
### Default one

Deafult one (Main branch)


### implemented initially

Initial Implementaion


### Latest Version

Latest Version

ragesoss:
Great. The complex implementation was a little faster, and significantly reduced allocations, but this one gets most of the performance gain without adding to code complexity.

PR #6344, opened 2025-06-04
Adds Year Requirement in Term Field for Course Creation and Cloning 

swayam-agrahari:
## What this PR does
This pull request adds a validation rule to ensure that the **Term** field includes a year (at least two digits) when creating or cloning a course. This aims to fix [#6321](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6321), which aims to prevent ambiguous term values such as "Summer" without a year.

- Adds frontend validation in both `CourseCreator` and `CourseClonedModal`.
- Updates placeholder text from `"Term"` to `"Term (including year)"`.
- Shows an inline error message prompting for values like `"Spring 2025"`.
- Updates the spec tests for such validation

## Screenshots
Before:
Screenshot From 2025-05-28 20-56-51

After:
Screenshot From 2025-05-28 20-54-42

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -44,6 +44,7 @@
 # Author: Wizardeck
 # Author: Yllelder
 # Author: YoViajo
+# Author: Yupik
 ---
 es:
   number:
@@ -1084,6 +1085,9 @@ es:
       update_featured_campaigns_button: Actualizar campañas destacadas
       add_button: Añadir campaña
       remove_button: Eliminar campaña
+    special_users:
+      new:
+        form_placeholder: Nombre de usuario
   suggestions:
     editing: Editar sugerencias
     suggestion_docs:
--- a/config/locales/nb.yml
+++ b/config/locales/nb.yml
@@ -828,6 +828,8 @@ nb:
     included_modules: Inkluderte moduler
     view_module_source: vis modulkilde
     reload_from_source: last inn igjen fra kilde
+  update_username:
+    label: Nytt brukernavn
   uploads:
     credit: Kreditering
     file_name: Filnavn
--- a/config/locales/sl.yml
+++ b/config/locales/sl.yml
@@ -272,6 +272,7 @@ sl:
     add_this_template: Dodaj predloge
     confirm_add_n_scoping_method: Želite dodati metod(e) obsega %{n}?
     name: Ime
+    category_name: Ime kategorije
     template_name: Ime predloge
     none: Ni sledenih kategorij.
     remove: Odstrani
@@ -501,6 +502,7 @@ sl:
     all: Vsi
     project: Projekt
     template: Predloga
+    category: Kategorija
     page: Stran
     translation: Prevod
     query: Poizvedba
--- a/config/locales/sv.yml
+++ b/config/locales/sv.yml
@@ -13,6 +13,7 @@
 # Author: Sturban
 # Author: Umeaboy
 # Author: WikiPhoenix
+# Author: Yupik
 ---
 sv:
   number:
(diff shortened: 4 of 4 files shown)


swayam-agrahari:
Hello, @ragesoss Please take a look at this pr and let me know if updation is required. Thanks!

ragesoss:
@swayam-agrahari thanks! I'll likely review this next week.

app/assets/javascripts/components/overview/course_cloned_modal.jsx
```diff
@@ -125,10 +125,20 @@ const CourseClonedModal = createReactClass({
...
+    this.props.setInvalid('term', I18n.t('courses.creator.field_invalid_term'));
+    return false;
+  }
+  return true;
+},
 
   saveCourse() {
     this.props.activateValidations();
-    if (this.props.isValid) {
+    const termValid = this.termIncludesYear();
```
ragesoss: Can this be implemented in a way that goes through `activateValidations()`? Doing it separately from the rest of the validations seems likely to cause confusion.

Imvedansh:
@swayam-agrahari Are you still working on this?

app/assets/javascripts/components/overview/course_cloned_modal.jsx
swayam-agrahari: Hey @ragesoss, sorry for the delay—my exams started and I completely lost track of this. I’ve added the term validation directly via the validation prop on the <TextInput> for the term field. This way, it validates on user input and shows an error immediately. It doesn’t hook into activateValidations() though, so it won’t trigger unless the field is touched. I tried wiring it into the Redux validation flow, but it was tricky since activateValidations() doesn’t run custom logic per field yet.

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -127,6 +127,20 @@ const Details = createReactClass({
     this.props.updateCourse({ ...this.props.course, namespaces });
   },
 
+  termValidation(value) {
```
ragesoss: Looks like this one isn't limited to ClassroomProgramCourse?

I think think we could leave the `Details` component unchanged, as we really only need to enforce this at the time of course creation. It's very rare to change the term value after that.

swayam-agrahari:
@ragesoss , I have made the suggested changes. Please let me know if anything else is required.

swayam-agrahari:
@ragesoss , hello please take a look at this pr when you are free. Thank you.

ragesoss:
Thanks! I expect to review it this week.

ragesoss:
Thanks again! I tested it out today.

The validation error message shows up immediately when a user starts typing. I think it's better for it to only show up when the user leaves the field, so that it doesn't distract when they are in the middle of typing.

I think it  needs to validate for a 4-digit year, so for consistency, since that's the usual format.

Your implementation doesn't prevent proceeding via 'Next' or creating a course with an invalid term. It should prevent a user from proceeding if the field doesn't pass the year requirement. (I also note that other fields stop you at 'Next' if missing, but even in the current UI, the term field doesn't get validated until you try to create the course and you must then click 'Back' to reach it).

ragesoss:
Hey @swayam-agrahari! Are you still interested in working on this? I think it's close.

oree-xx:
Hello @swayam-agrahari nice work so far, are you still working on it or i could continue, kindly let me know.
cc: @ragesoss

swayam-agrahari:
Heyy @oree-xx , you can go ahead with this. Thanks!

oree-xx:
Alright thank youu @swayam-agrahari

PR #6348, opened 2025-06-09
Fix N+1 query issues on survey assignments index view

Abishekcs:
## What this PR does

Fixes N+1 query issues on survey assignments index view by using `.includes`

**a)  Where is the N+1 this fixes**

https://github.com/user-attachments/assets/32130701-10ca-454c-8d33-146eb0b8d8c1

**b) What do the N+1 queries look like**

https://github.com/user-attachments/assets/35e22076-a1d1-4108-b5fb-9358ce3792d3

## Performance & Resource Usage Average

| Metric               | Default | Initial Implementaion | Latest Version |
|----------------------|-------------|----------------|---------------------|
| **Total Time (ms)**  | 186.1       | **66.7**       | 128.6               |
| **View Time (ms)**   | 138.3       | **27.9**       | 99.4                |
| **AR Time (ms)**     | 42.5        | **22.4**       | 24.8                |
| **Allocations**      | 85494.3     | **20625.2**    | 50934               |

Screenshot from 2025-06-11 18-46-11

**Default Version**

Screenshot from 2025-06-11 18-50-04

**Initial Implementaion**

After

**Latest Version**

Latest Version

## Screenshots (Default & Latest Version)
Before:

https://github.com/user-attachments/assets/de528e88-9186-481a-95b7-2974f88f2f43

Before

Screenshot from 2025-06-09 15-59-38

After:

https://github.com/user-attachments/assets/8b43303f-8191-42fb-9336-35d00d4144f1

Latest Version

Screenshot from 2025-06-11 19-19-26

### DIFF
--- a/app/controllers/survey_assignments_controller.rb
+++ b/app/controllers/survey_assignments_controller.rb
@@ -16,7 +16,8 @@ class SurveyAssignmentsController < ApplicationController
   # GET /survey_assignments
   # GET /survey_assignments.json
   def index
-    @survey_assignments = SurveyAssignment.all
+    @survey_assignments = SurveyAssignment.includes(:campaigns, :survey, :survey_notifications,
+                                                    :courses).all
   end
 
   # GET /survey_assignments/1
--- a/app/helpers/surveys_analytics_helper.rb
+++ b/app/helpers/surveys_analytics_helper.rb
@@ -38,14 +38,13 @@ module SurveysAnalyticsHelper
   end
 
   def assignment_response(survey_assignment)
-    completed = survey_assignment.survey_notifications.completed.length
+    completed = survey_assignment.survey_notifications.count(&:completed)
     notified = survey_assignment.survey_notifications.length
     response_summary_string(completed, notified)
   end
 
   def assignment_dismissal(survey_assignment)
-    dismissed = survey_assignment.survey_notifications
-                                 .where(dismissed: true, completed: false).length
+    dismissed = survey_assignment.survey_notifications.count { |n| !n.completed && n.dismissed }
     notified = survey_assignment.survey_notifications.length
     response_summary_string(dismissed, notified)
   end
--- a/app/views/survey_assignments/_survey_assignment.html.haml
+++ b/app/views/survey_assignments/_survey_assignment.html.haml
@@ -1,6 +1,6 @@
 .survey__admin-row.survey__admin-row--assignment
   .survey__admin-row__title
-    .contextual Target Campaign#{survey_assignment.campaigns.count > 1 ? 's' : ''}:
+    .contextual Target Campaign#{survey_assignment.campaigns.size > 1 ? 's' : ''}:
     %strong= survey_assignment.campaigns.pluck(:title).join(', ')
     .contextual Target User:
     %span.emphasis= user_role(survey_assignment)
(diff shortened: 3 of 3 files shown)


app/controllers/survey_assignments_controller.rb
```diff
@@ -77,6 +76,81 @@ def send_test_email
...
+
+  # Loads campaign titles related to survey assignments
+  # @return [Hash{Integer => Array<String>}] survey_assignment_id => [campaign_title, ...]
+  # Example: {6=>["2025 Campaign", "Default Campaign"], 5=>["2025 Campaign"]}
+  def load_survey_campaigns_title_by_id
+    Campaign.joins(:campaigns_survey_assignments)
+            .where(campaigns_survey_assignments: { survey_assignment_id: @survey_assignment_ids })
+            .pluck('campaigns_survey_assignments.survey_assignment_id', 'campaigns.title')
+            .group_by(&:first)
+            .transform_values { |arr| arr.map(&:last).uniq }
```
ragesoss: I'm not necessarily against this kind of strategy — creating a data structure that has the required data by writing a query that plucks the minimum data needed — but in most cases, I'd rather have a more idiomatic Rails solution if possible, using `includes` to load whatever records are need to avoid N+1 but relying on ActiveRecord to organize and access the data without rolling our own data structures. These are typically not maximally efficient — especially in terms of allocations — but if the N+1 is fixed, usually the additional performance gains are not worth the tradeoff of harder-to-read and harder-to-edit code.

Whenever I find an N+1 issue, I can almost always find a Rails-y solution, and doing anything more adjusting the main query is a last resort.

Abishekcs: Okay, I’ll make the required changes. Thank you for the review!

app/controllers/survey_assignments_controller.rb
```diff
@@ -10,14 +10,13 @@ class SurveyAssignmentsController < ApplicationController
...
   before_action :set_survey_assignment, only: %i[show edit update destroy send_test_email]
   before_action :set_survey_assignment_options, only: %i[new edit update]
+  before_action :prepare_survey_assignment_view_data, only: %i[index]
   layout 'surveys'
   include SurveyAssignmentsHelper
 
   # GET /survey_assignments
   # GET /survey_assignments.json
-  def index
-    @survey_assignments = SurveyAssignment.all
```
ragesoss, line 19: My starting point would be here. The N+1 issues are based around iterating through the SurveyAssignments, because we only load the assignments here, but the first thing to try would just adding `includes`  to try to load more data in the initial query.

That might not get us where we need to be, but it might fix the worst of it.

In production, this route is pretty slow, taking about 5 to 6 seconds to render with about 40 SurveyAssignment records, and I assume that render time is basically scaling with the number of SurveyAssignments.

~40 is not so many that we can't tolerate some N+1 queries, if they are fast. So for this, I would try to find a balance... look for the simplest way to reduce the number of queries required, and see how that improves performance before moving on to a fuller but more complex solution that fully handles the N+1 issues.

The idea here is that Rails default patterns are not necessarily the most efficient, but Rails performance is usually good enough if we avoid the most egregious N+1 patterns.

Abishekcs: So, from what I understand: start with the Rails way. If performance is good, keep it—there's no need to over-optimize. But if there's not much improvement, then look into other approaches that go beyond the standard Rails way.

ragesoss: exactly!

ragesoss:
This looks great!

PR #6354, opened 2025-06-10
[Data-rearchitecture] Change behavior `UpdateCourseWikiTimeslices`  to make full updates work progressively

gabina:
## What this PR does
This PR changes behavior in `UpdateCourseWikiTimeslices` class when a full update is required. Now, the timeslices are re-created with `needs_update` set to true, and the `needs_update`  course flagis set to false right after doing that. This is to make the full update process able to work progressively. It also adds basic specs for it.


gabina:

PR #6357, opened 2025-06-12
[Data-rearchitecture] Improve `UpdateTimeslicesCourseUser` logic to prevent re-marking users as newly added

gabina:
## What this PR does
This PR updates `created_at` field for courses users right after considered them as newly added (and set timeslices to reprocess) in `UpdateTimeslicesCourseUser`, so they're not considered new users again in the next update if the process dies unexpectedly.

### Context
I was reproducing locally the update for course [Coordinate Me 2025 LV](https://outreachdashboard.wmflabs.org/courses/Wikidata/Coordinate_Me_2025_LV_(2025)). The addition of new course users with so many revisions triggers a re-processing of every timeslice in the course. While I don't think there is so much we can do around that, I think we can do make an improvement on the process. As it is working now, a new user is identified by comparing the `created_at` field for the course user and the `start_time` for the last successful update. If the user was created after the last `start_time`, then it's a new user. The thing is that if the update dies in the middle of the process, the user is considered new in the next update, setting as `needs_update` some timeslices that maybe were already re-processed in the previous update that didn't finish successfully. We could make a change so that, once the timeslices were set to `needs_update`, those users are no longer considered new (regardless of whether the update finishes successfully or dies in the middle). That way, in the next update, the remaining timeslices will be reprocessed until it eventually finishes successfully.

### DIFF
--- a/app/services/update_timeslices_course_user.rb
+++ b/app/services/update_timeslices_course_user.rb
@@ -28,9 +28,9 @@ class UpdateTimeslicesCourseUser
 
     # Users that were added after the last course update start are considered new
     # It's not safe to rely on new_user_ids = current_user_ids - processed_users
-    course_update_start = @course.flags['update_logs'].values.last['start_time']
+    @course_update_start = @course.flags['update_logs'].values.last['start_time']
     new_user_ids = @course.students.where('courses_users.created_at >= ?',
-                                          course_update_start).pluck(:id)
+                                          @course_update_start).pluck(:id)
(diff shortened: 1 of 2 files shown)


gabina:
@ragesoss let me know what you think about this

ragesoss:
This looks good to me.

ragesoss:
Is this ready to go from your perspective @gabina ?

gabina:
Yes, I think this is ready to go

PR #6358, opened 2025-06-13
 Optimize Question Group Rendering and Eliminate N+1 Queries

Abishekcs:
## What this PR does

This PR improves the performance and maintainability of the admin survey question group interface by:

- Eliminating N+1 queries across question group rendering.

- Preloading and grouping related data (Survey, SurveyAssignment, Rapidfire::Question, and PaperTrail::Version) for efficient access in views.

- Refactoring helper methods to accept preloaded data and reduce query load during rendering.

## Changes

- **View (`index.html.haml`)**
   - Preloaded question group-related associations: survey mappings, surveys with assignments, related questions, and version history.
   - Passed preloaded data to partials to reduce query repetition.

- **Partial (`_question_group.html.haml`)**
   - Switched to using preloaded data for questions, status, and author display.
   - Ensured fallback behavior remains consistent.

- **Helper (`SurveysAnalyticsHelper`)**
  - Added `question_group_survey_author` to retrieve author from a version object.
  - Updated `question_group_status` to accept grouped objects instead of fetching them dynamically.

https://github.com/user-attachments/assets/af514b95-f7b3-4719-a706-018193e841eb

https://github.com/user-attachments/assets/cb49f42c-2ae4-462e-8d36-a45019623d45

## Rails Log Performance Comparison (Average)

| Metric                | Before (N+1)      | After (Optimized) | Improvement         |
| --------------------- | ----------------- | ----------------- | ------------------- |
| **Avg Response Time** | ~218 ms           | ~63 ms            | ~71% faster         |
| **Avg DB Time**       | ~82 ms            | ~18 ms            | ~78% reduction      |
| **Avg Allocations**   | ~68,000+ objects  | ~27,000 objects   | ~60% fewer objects  |

Screenshot from 2025-06-13 15-05-18

## Screenshots
Before:

https://github.com/user-attachments/assets/c17f2ac2-0da6-46b9-8938-bf0396be6f51

Before Screenshot

Before Rails Log

After:

After Rails Log

After Screenshot

https://github.com/user-attachments/assets/0ba4d319-71aa-48de-856a-7eade0c5d3a5

### DIFF
--- a/app/views/rapidfire/question_groups/index.html.haml
+++ b/app/views/rapidfire/question_groups/index.html.haml
@@ -2,6 +2,21 @@
 
 = render 'admin_header'
 .container.survey__edit
+  - # Extract all question group IDs from @question_groups
+  - question_groups_id = @question_groups.pluck(:id)
+  
+  - # Group survey-question group mappings by question group ID
+  - @survey_question_groups = SurveysQuestionGroup.where(rapidfire_question_group_id: question_groups_id).group_by(&:rapidfire_question_group_id)
+  
+  - # Fetch surveys linked to question groups, preload assignments, and group by survey ID
+  - @surveys = Survey.includes(:survey_assignments).where(id: @survey_question_groups.values.flatten.map(&:survey_id).uniq).group_by(&:id)
+  
+  - # Group all questions by their question group ID
+  - @rapidfire_questions = Rapidfire::Question.where(question_group_id: question_groups_id).group_by(&:question_group_id) 
+  
+  - # Get the latest version for each question group using PaperTrail versioning
+  - @latest_question_groups_version = PaperTrail::Version.where(item_type: 'Rapidfire::QuestionGroup', item_id: question_groups_id).group_by(&:item_id).transform_values(&:last)
+
   - if @question_groups.length > 0
     %ul.list.survey__admin__row-container= render partial: "question_group", collection: @question_groups.reverse
   - else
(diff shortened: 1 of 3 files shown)


app/views/rapidfire/question_groups/index.html.haml
```diff
@@ -2,6 +2,21 @@
 
 = render 'admin_header'
 .container.survey__edit
+  - # Extract all question group IDs from @question_groups
```
ragesoss: Is this haml template used only for the index view?

The normal Rails pattern here would be to set all the necessary instance variables in the controller action, rather than the view. Ideally, all the queries for loading data get completed in the controller before the view template begins to render (although we have many situations where that's not the case). That makes it much easier to reason about  when looking at the controller.

In this case, if it's the gem's default question groups index action that we're using, then overwriting it in the surveys initializer may be the best option.

Abishekcs: > Is this haml template used only for the index view?

Yes, i think so.

Abishekcs: > In this case, if it's the gem's default question groups index action that we're using, then overwriting it in the surveys initializer may be the best option.

Okay.

Abishekcs: > In this case, if it's the gem's default question groups index action that we're using, then overwriting it in the surveys initializer may be the best option.

@ragesoss, I've made the requested changes, whenever you have time, could you please take another look? Thank you again for your review!

ragesoss:
This looks good to me. I will merge once the tests pass.

PR #6362, opened 2025-06-16
Optimize response method by preloading and grouping Rapidfire answers csv Download (PART 3)

Abishekcs:
This PR is dependent on  #6360 (merged) and  #6361(merged)

## What this PR does

This PR improves the `response` method performance by introducing an answer preloading mechanism via the `prepare_rapidfire_answers` method.

### Key Improvements:

* **Bulk preloading:** Answers are fetched in a single query using pre-collected `answer_group_ids` and `question_ids`.
* **Grouped access:** Answers are grouped by `[answer_group_id, question_id]` to avoid lookup overhead during iteration.
* **N+1 elimination:** Removes per-question, per-user database queries from the response generation loop.

Previously, every call to `response(user)` would issue multiple `Answer.where` calls — one for each question per user. This update consolidates those into a single query and enables fast in-memory access, greatly improving performance on views or reports involving many users.

Screenshot from 2025-06-16 17-13-11

Screenshot from 2025-06-16 16-41-47

## Screenshots from rails log for CSV Download

**Before:**

Before Logs

**After:** 

After Logs

## Screenshots

**Before:**

https://github.com/user-attachments/assets/5c44eb5e-8cdc-478a-8854-7dd5836d1bc4

**After:**

https://github.com/user-attachments/assets/65250489-29ab-4ef3-a966-c63ffa8d6606

### DIFF
--- a/app/models/surveys/survey.rb
+++ b/app/models/surveys/survey.rb
@@ -93,11 +93,33 @@ class Survey < ApplicationRecord
     end
   end
 
-  def response(user)
+  def prepare_rapidfire_answers
+    @answers ||= begin
+      answer_group_ids = answer_groups_by_user_id.values.flatten.map(&:id)
+      question_ids = questions_with_separate_followups.map { |q| q[:question].id }
+
+      Rapidfire::Answer.where(
+        answer_group_id: answer_group_ids,
+        question_id: question_ids
+      )
+    end
+
+    @answers_by_group_and_question ||= @answers.group_by do |answer|
+      [answer.answer_group_id, answer.question_id]
+    end
+  end
+
+  def response(user) # rubocop:disable Metrics/CyclomaticComplexity
     answer_group_ids = answer_groups_by_user_id[user.id].map(&:id)
+    prepare_rapidfire_answers
+
     questions_with_separate_followups.map do |question_hash|
-      answer = Rapidfire::Answer.where(answer_group_id: answer_group_ids,
-                                       question_id: question_hash[:question].id).first
+      question_id = question_hash[:question].id
+
+      answer = answer_group_ids.map do |group_id|
+        @answers_by_group_and_question[[group_id, question_id]]&.first
+      end.find(&:present?)
+
       question_hash[:followup] ? answer&.follow_up_answer_text : answer&.answer_text
     end
   end
--- a/app/models/surveys/survey.rb
+++ b/app/models/surveys/survey.rb
@@ -109,6 +109,11 @@ class Survey < ApplicationRecord
     end
   end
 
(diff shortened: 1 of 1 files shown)


Abishekcs:
@ragesoss This PR is ready for review.
Below are the CSV files generated by the main branch and this PR.
Since the data is quite large, I used https://www.textcompare.org/csv/ to verify that the results match and are correct.

[CSV WITH N+1.csv](https://github.com/user-attachments/files/20760952/CSV.WITH.N%2B1.csv)
[CSV WITHOUT N+1.csv](https://github.com/user-attachments/files/20760954/CSV.WITHOUT.N%2B1.csv)

app/models/surveys/survey.rb
```diff
@@ -93,11 +93,33 @@ def answer_groups_by_user_id
...
+        question_id: question_ids
+      )
+    end
+
+    @answers_by_group_and_question ||= @answers.group_by do |answer|
+      [answer.answer_group_id, answer.question_id]
+    end
+  end
+
+  def response(user) # rubocop:disable Metrics/CyclomaticComplexity
```
ragesoss, line 117: Hmm... I assume that the issue here is that `response` gets called once for every user who submitted a survey response, so this change works by fetching data for the entire survey first time that `response` gets called, and then each subsequent time `prepare_rapidfire_answers` doesn't need to do any queries because the data is already fetched. However, this means that if there's any case where we are just calling `response` for a single user, it would be fetching a lot more data than it needs for just that user's response.

Are we using the `response` method anywhere except the CSV downloads? If not, I guess this is okay, but it should include a comment to warn people about the method being designed around loading data for the entire survey and it may be inefficient if used for a single user.

Abishekcs: > **Are we using the `response` method anywhere except the CSV downloads?**

I searched using VS Code, and I didn’t find any other usage of the `response` method besides the survey CSV download.

> **However, this means that if there's any case where we are just calling `response` for a single user, it would be fetching a lot more data than it needs for just that user's response.**

I don’t think that will be an issue right now. The `response` method is only called through the `respondents` method inside `to_csv`.
Since `prepare_rapidfire_answers` uses `answer_groups_by_user_id`, which gets its user IDs from `respondents`, it will only fetch data relevant to those users. So, it shouldn't load unnecessary data.


> **If not, I guess this is okay, but it should include a comment to warn people about the method being designed around loading data for the entire survey and it may be inefficient if used for a single user.**

I’ll add a comment now, since with these changes, the `response` method isn't meant to be used independently — it should only be called through `to_csv`.
Thanks for the review!

PR #6366, opened 2025-06-17
Optimize admin course loading for P&E Homepage by conditionally eager-loading students only for Wiki Ed

Abishekcs:
## What this PR does

This PR refines the `set_admin_courses_if_admin` method to avoid eager loading student associations unless necessary.

- Prevents unnecessary eager loading when student associations aren't used.
- Improves performance for admin users in non-Wiki Ed contexts.
- Keeps code aligned with actual data needs based on feature flags.

### Before
`current_user.courses.strictly_current.includes(:students)` was always called for admin users, even when student data wasn’t needed.

### After
Student data is now eager-loaded **only** if `Features.wiki_ed?` is enabled. In other cases, we load only the courses, reducing database overhead.

## .includes(:students) was only using Wik Edu Dashboard and not by P&E as shown in the screenhot below of `_admin_courses.html.haml`

Screenshot from 2025-06-17 17-51-40

## Screenshots
**Before:**

Before Rails includes

Before

**After:**

After

After not using includes

### DIFF
--- a/app/controllers/dashboard_controller.rb
+++ b/app/controllers/dashboard_controller.rb
@@ -40,7 +40,12 @@ class DashboardController < ApplicationController
 
     return unless current_user.admin?
     @submitted = Course.submitted_but_unapproved
-    @strictly_current = current_user.courses.strictly_current.includes(:students)
+    @strictly_current = if Features.wiki_ed?
+                          # Reduce number of queries to show ungreeted student count using includes(:students) # rubocop:disable Layout/LineLength
+                          current_user.courses.strictly_current.includes(:students)
+                        else
+                          current_user.courses.strictly_current
+                        end
   end
 
   def current_courses
--- a/spec/controllers/dashboard_controller_spec.rb
+++ b/spec/controllers/dashboard_controller_spec.rb
@@ -55,6 +55,20 @@ describe DashboardController, type: :request do
         expect(assigns(:blog_posts)).to eq([])
       end
     end
+
+    context 'when Wiki Ed feature is disabled' do
+      before do
+        allow(Features).to receive(:wiki_ed?).and_return(false)
+        allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(admin)
+        create(:courses_user, user_id: admin.id, course_id: course.id)
+      end
+
+      it 'does not eager-load students for strictly_current courses' do
+        get '/course_creator'
+        expect(assigns(:pres)).to be_a(DashboardPresenter)
+        expect(assigns(:pres).past.count).to eq(1)
+      end
+    end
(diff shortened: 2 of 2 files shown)


app/controllers/dashboard_controller.rb
```diff
@@ -40,7 +40,11 @@ def set_admin_courses_if_admin
 
     return unless current_user.admin?
     @submitted = Course.submitted_but_unapproved
-    @strictly_current = current_user.courses.strictly_current.includes(:students)
+    @strictly_current = if Features.wiki_ed?
```
ragesoss, line 43: This seems reasonable, but it's not obvious from the controller why the query is different based on the Features flag. A comment explaining how the Wiki Ed behavior differs would be helpful here.

Abishekcs: Okay

ragesoss: This comment is a good start. Can you edit it to use plainer language and make it clearer? In the simplest way you can come up with, it should answer the question of why it `includes(:students)` for Wiki Ed only.

Abishekcs: How about - Reduce number of queries to show ungreeted student count using includes(:students) ?

ragesoss: Works for me.

Abishekcs: Okay i have pushed the latest changes.

ragesoss:
Great. Do you know which tests exercise this code? I'd like to be confident that if either branch had a bug that generated an error, we'd know from the test suite.

Abishekcs:
> Great. Do you know which tests exercise this code? I'd like to be confident that if either branch had a bug that generated an error, we'd know from the test suite.

spec/controllers/dashboard_controller_spec.rb

ragesoss:
> > Great. Do you know which tests exercise this code? I'd like to be confident that if either branch had a bug that generated an error, we'd know from the test suite.
> 
> spec/controllers/dashboard_controller_spec.rb

And if either branch has a bug (eg, if you just add a `raise StandardError`) it will break one of the tests?

Abishekcs:
> And if either branch has a bug (eg, if you just add a `raise StandardError`) it will break one of the tests?

It does when i add `raise Standard Error` but only for `if` since i think the test case only test for `wiki ed`. I wil share the screenrecording

ragesoss:
Okay, in that case, you should add a test case for the non-wiki_ed version.

Abishekcs:
https://github.com/user-attachments/assets/5b9c11f2-027c-4f50-b309-53af27e4465a

PR #6367, opened 2025-06-17
Extract survey results logic into separate SurveyResultsController

Abishekcs:
## What this PR does

This PR introduces a refactor that moves the `results` action from `SurveysController` into a new `SurveyResultsController`. The goal is to isolate logic related to survey results, reducing complexity in the main controller and setting the stage for further enhancements (performance improvements #6329).

## Screenshots

Before:

https://github.com/user-attachments/assets/ccf741aa-060d-4ef8-a61c-8eabdab8d57f

After:

https://github.com/user-attachments/assets/840a8f38-7948-492a-9e02-fc1c6e1670cc

### DIFF
--- a/app/controllers/survey_results_controller.rb
+++ b/app/controllers/survey_results_controller.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+class SurveyResultsController < SurveysController
+  layout 'surveys'
+
+  before_action :set_survey
+  before_action :set_question_groups
+
+  def results
+    protect_confidentiality { return }
+    respond_to do |format|
+      format.html
+      format.csv do
+        filename = "#{@survey.name}-results#{Time.zone.today}.csv"
+        send_data @survey.to_csv, filename:
+      end
+    end
+  end
+
+  private
+
+  def set_question_groups
+    super
+    @survey_user_cache = {}
+  end
+end
--- a/app/controllers/surveys_controller.rb
+++ b/app/controllers/surveys_controller.rb
@@ -15,14 +15,12 @@ class SurveysController < ApplicationController
     destroy
     edit_question_groups
     course_select
-    results
   ]
   before_action :ensure_logged_in
   before_action :set_question_groups, only: %i[
     show
     edit
     edit_question_groups
-    results
   ]
   before_action :check_if_closed, only: [:show]
   before_action :set_notification, only: [:show]
@@ -39,17 +37,6 @@ class SurveysController < ApplicationController
 
   def results_index
     @surveys = Survey.all
-  end
-
-  def results
-    protect_confidentiality { return }
-    respond_to do |format|
-      format.html
-      format.csv do
-        filename = "#{@survey.name}-results#{Time.zone.today}.csv"
-        send_data @survey.to_csv, filename:
-      end
-    end
   end
 
   # GET /surveys/1
@@ -155,7 +142,6 @@ class SurveysController < ApplicationController
   def set_question_groups
     @question_groups = Rapidfire::QuestionGroup.all
     @surveys_question_groups = SurveysQuestionGroup.by_position(params[:id])
-    @survey_user_cache = {}
   end
 
   # This removes the question groups that do not apply to the course, because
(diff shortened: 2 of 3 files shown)


app/views/layouts/survey_results.html.haml
```diff
@@ -0,0 +1,25 @@
+- content_for :head do
```
ragesoss: It shouldn't be necessary to create any new haml views. If there is a view that was only being used for the results, it can be renamed. If a view is used by both the results and another action, the controller action can use `render [tempate]` to render the view that has a different filename than the default matching the action name.

app/controllers/surveys_controller.rb
```diff
@@ -41,17 +39,6 @@ def results_index
     @surveys = Survey.all
   end
 
-  def results
-    protect_confidentiality { return }
```
ragesoss, line 45: This method is only used for the results action, so the definition can be removed from this controller as well.

app/controllers/survey_results_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      format.csv do
+        filename = "#{@survey.name}-results#{Time.zone.today}.csv"
+        send_data @survey.to_csv, filename:
+      end
+    end
+  end
+
+  private
+
+  def set_survey
```
ragesoss: If this and other methods should be shared with the other survey controller, you may be able to make this controller a child class of the main survey controller (as we do with `Courses::DeleteFromCampaignController`).

We don't want to duplicate code between these files for the refactor.

app/views/layouts/survey_results.html.haml
Abishekcs: Okay

app/controllers/survey_results_controller.rb
Abishekcs: Okay

app/controllers/surveys_controller.rb
Abishekcs: I think i have removed this one already  😅

Abishekcs:
@ragesoss I’ve made the required changes. Could you please review this PR whenever you have a time? Thank you!

app/views/survey_results/_question_group.html.haml
```diff
@@ -0,0 +1,38 @@
+- content_for :after_title, ' – Question Groups'
```
ragesoss: Is this new file necessary?

app/views/survey_results/_results_admin_header.html.haml
```diff
@@ -0,0 +1,8 @@
+- content_for :after_title, ' – Survey Results'
```
ragesoss: What header was being rendered before?

app/views/survey_results/_question_group.html.haml
Abishekcs: Not really it just omits someparts that is not required by survey results.

app/views/survey_results/_results_admin_header.html.haml
Abishekcs: If i understood correctly it is being used by `survey_results/results.html.haml`

app/views/survey_results/_question_group.html.haml
ragesoss: In that case, I think it would be better to simplify this refactor by continuing to use the same template that was previously used to render the question groups.

Abishekcs: The highlighted code in the screenshot shows the lines that were omitted in the copied file from the original


https://github.com/user-attachments/assets/66a236ab-c70c-4b08-8de6-d97579ce0b91

app/views/survey_results/_results_admin_header.html.haml
Abishekcs: > If i understood correctly it is being used by `survey_results/results.html.haml`

Nevemind this  I have changed the code just now will push the changes.

Abishekcs:
Hi @ragesoss, I've made the required changes and tried to keep them as minimal as possible. Please let me know if any further adjustments are needed. Thanks again for your helpful review!

Abishekcs:
I don't think the failing test are related to the changes made in this PR. Test seems to pass locally

Screenshot from 2025-06-20 00-01-48

ragesoss:
Looks great!

Issue #6291, opened 2025-04-15
Add support for courses with timelines but without set meeting days


ragesoss:
Currently, using the timeline basically requires having meeting days set, as weeks with no meeting days are treated as holiday weeks and timeline blocks don't get inserted in those weeks.

Some courses are completely asynchronous, so they have a start and end and weekly timeline content still makes sense, but they have no set meeting dates at all. The Dashboard should handle this situation, likely by making it optional to include any meeting dates at all, and for that situation, basically treat the timeline as if every day of every week is a meeting date (but with no display of those days on the timeline weeks).

Formasitchijoh:
working on this

PR #6374, opened 2025-06-20
Add Support for No meeting days in Course creation

Formasitchijoh:
## What this PR does
This PR solve issue #6291 , by adding support for courses with timelines but without set meeting days, this is achieved by adding a field(**`no_meeting_days`**)  to the course table that keeps track of the status of the course if it has set meeting days or not and using it to compute the timeline appropriately. 

## Screenshots
After
## Course with no meeting days sets
https://www.loom.com/share/e0d9517577d34afe8535d4c1c62c228c?sid=88da0d2e-cbaa-4a4d-b9f6-e7ee4ae7664b

## Course with meeting days set 
https://www.loom.com/share/566c2eab6c0c4ea394e90773b276b8cc?sid=887877d2-c000-4dbb-a3e6-56923459f584


Formasitchijoh:
@ragesoss I think this  ready now

ragesoss:
This looks solid. The code is very clean, and from the demo video, it looks like it works as intended.

I think it the UI for the dates panel needs improvement.

The toggle for this feature should probably be just above the weekday selector (in the same section), and ideally it should give a clear visual indication of disabling (or hiding) the weekday and calendar selector when it's active.

Suggestion for the text: "Do not include meeting days or holidays/breaks in the course timeline. (Note: If you don't include meeting days and your course includes vacation weeks, you may need to manually adjust the timeline to account for them.)"

Formasitchijoh:
> This looks solid. The code is very clean, and from the demo video, it looks like it works as intended.
> 
> I think it the UI for the dates panel needs improvement.
> 
> The toggle for this feature should probably be just above the weekday selector (in the same section), and ideally it should give a clear visual indication of disabling (or hiding) the weekday and calendar selector when it's active.
> 
> Suggestion for the text: "Do not include meeting days or holidays/breaks in the course timeline. (Note: If you don't include meeting days and your course includes vacation weeks, you may need to manually adjust the timeline to account for them.)"

So If I am getting your right it should be a toggle button that hide the weekday and calendar selector when it's active.

ragesoss:
I think so. That would make it more clear that, if they aren't using meeting days, the rest of the inputs on that panel are not applicable.

Formasitchijoh:
> I think so. That would make it more clear that, if they aren't using meeting days, the rest of the inputs on that panel are not applicable.

That will be better, I will make the change

Formasitchijoh:
@ragesoss I updated the feedback

ragesoss:
One minor issue: when enabled, there's an extra horizontal rule that should be removed:

extra line

The new section looks good, but it makes this wizard panel taller than I'd like. I think it might be a good idea to change the layout of the two date entry sections so that the inputs are on the same line.

Testing this out locally, I found that the Dates modal doesn't include the 'no meetings' toggle, so if that flag is enabled, the Dates panel is broken. The user sees the Meeting Days and Calendar components, but they don't function correctly. (It looks like it's still possible to click and activate both components, even though they don't highlight whatever you click.)

Formasitchijoh:
> One minor issue: when enabled, there's an extra horizontal rule that should be removed:
> 
> extra line
> 
So if i am getting you right the input field of the start and end should be flexed horizontal instead of vertical as they currently are to reduce the height of the form
> The new section looks good, but it makes this wizard panel taller than I'd like. I think it might be a good idea to change the layout of the two date entry sections so that the inputs are on the same line.
> 

I realized on editing the course  Dates where the no_meeting_days is true the toggle button does appear is this what you are referring to
> Testing this out locally, I found that the Dates modal doesn't include the 'no meetings' toggle, so if that flag is enabled, the Dates panel is broken. The user sees the Meeting Days and Calendar components, but they don't function correctly. (It looks like it's still possible to click and activate both components, even though they don't highlight whatever you click.)

ragesoss:
> the toggle button does appear
 
You mean doesn't appear? Yes, that's what I mean.

Formasitchijoh:
Yes I meant, it does not appear

Formasitchijoh:
@ragesoss  I updated the design to look like this is this okay 

## Form Panel
Screenshot 2025-06-27 at 12 27 12 PM


## Course Dates Modal
Screenshot 2025-06-27 at 12 30 33 PM

ragesoss:
yeah, that looks good!

Formasitchijoh:
@ragesoss the build is complete

ragesoss:
great! i think it's ready, but i'll do some hopefully final testing early next week.

ragesoss:
The functionality is solid. I'm not quite satisfied with the wording and the way this feature gets presented during the course creation process; I am going to talk about it with other Wiki Education staff.

I think toggle needs a little more/different explanation and labeling, but I'm not sure what a good concise wording is. I'll figure that out soon.

I also think it might make sense to actually make 'no-dates' the default, and switch the toggle to *enable* dates, but I want to discuss that with others first.

If we do keep the current default of using dates, and keep the toggle as a 'disable' toggle, it might be best to move the toggle and the weekday and calendar to a separate wizard panel, but I think it would make sense to open a separate issue for that after merging.

Formasitchijoh:
Okay I can make any changes from your feedback

> The functionality is solid. I'm not quite satisfied with the wording and the way this feature gets presented during the course creation process; I am going to talk about it with other Wiki Education staff.

ragesoss:
Okay!

After talking with Helaine... let's make no-dates the default for new courses (and have it represented by the 'off' position of the toggle). Obviously this should not affect existing courses that were created before the feature gets deployed.

Label: `Specify course meeting days, holidays and breaks?`

Updated note: `Note: If you don't add meeting days and your course has break weeks, you may need to manually adjust the timeline to account for them.`

ragesoss:
@Formasitchijoh please return to this when you get a chance.

Formasitchijoh:
> Okay!
> 
> After talking with Helaine... let's make no-dates the default for new courses (and have it represented by the 'off' position of the toggle). Obviously this should not affect existing courses that were created before the feature gets deployed.
> 
> Label: `Specify course meeting days, holidays and breaks?`
> 
> Updated note: `Note: If you don't add meeting days and your course has break weeks, you may need to manually adjust the timeline to account for them.`

Hi @ragesoss  is this the only changes I should fix up here,

ragesoss:
yes, plus the merge conflict.

ragesoss:
@Formasitchijoh that, plus fixing the conflicts to make it merge-able. Thank you! This is still relevant, and shipping it soon would be great.

Formasitchijoh:
> @Formasitchijoh that, plus fixing the conflicts to make it merge-able. Thank you! This is still relevant, and shipping it soon would be great.

Currently testing it out, will  push it soon!

PR #6375, opened 2025-06-22
Optimize survey results data loading (PART 1): Enrich answers with user and course data while eliminating N+1 queries

Abishekcs:
**Note: This is the first PR which breaks down #6329**

## What this PR does

This PR refactors how survey answers are processed and enriched in order to:
- Attach associated questions to each answer
- Collect and preload related users
- Dynamically attach each user's related course via `SurveyNotification`
- Preload associated `campaigns` and `tags` for each course
- Replace the legacy `survey_user_cache` approach

---

Previously, rendering answer-related data introduced **severe N+1 query problems**, including:

- One query per course to fetch `campaigns` and `tags`
- One query per user to retrieve completed `survey_notifications`
- Multiple queries per answer when calling `answer.course(@survey.id, user)`

This PR eliminates those N+1s by:
- Collecting user IDs in one pass
- Batch-loading users and their course info
- Eager-loading `campaigns` and `tags` with `.includes`
- Caching course information via a dynamically defined method: `user.survey_course`

---

- Added a new method: `enrich_answers_with_users_and_courses` in `SurveysHelper`
- Introduced helper methods:
  - `store_user_ids`
  - `store_users_by_id`
  - `user_survey_courses`
- Updated `question_group_locals` to invoke this enrichment process
- Updated `QuestionResultsHelper` and all relevant views to replace `survey_user_cache` with `@users_by_id`
- Removed now-unnecessary instance variables and logic in `SurveyResultsController` and `_question_results.html.haml`

---

This refactor prevents N+1 queries like:

- `SELECT * FROM courses WHERE id = ... LIMIT 1`
- `SELECT * FROM tags WHERE course_id = ...`
- `SELECT * FROM survey_notifications WHERE user_id = ...`
- `SELECT * FROM campaigns_courses WHERE course_id = ...`

All of these are now executed as **batch queries**, resulting in faster page loads and reduced DB load.

Where is the N+1 this fixes

https://github.com/user-attachments/assets/fe6253c8-8028-4317-bb90-6a6350096423

## Screenshots

Before:

Before

After:

After

### DIFF
--- a/app/helpers/surveys_helper.rb
+++ b/app/helpers/surveys_helper.rb
@@ -92,7 +92,7 @@ module SurveysHelper
     @questions = surveys_question_group
                  .rapidfire_question_group
                  .questions
-                 .includes(answers: { user: :survey_notifications })
+                 .includes(answers: :user)
 
     @id_to_question = {}
     @questions.each do |question|
@@ -103,10 +103,7 @@ module SurveysHelper
                               .where(question_id: @questions.pluck(:id))
                               .group(:question_id).count
 
-    @answer_group_builder.answers.each do |answer|
-      question = @id_to_question[answer.question_id]
-      answer.question = question
-    end
+    enrich_answers_with_users_and_courses
 
     return { question_group: @question_group,
       answer_group_builder: @answer_group_builder,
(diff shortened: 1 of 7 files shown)


app/helpers/surveys_helper.rb
```diff
@@ -83,7 +83,7 @@ def grouped_question(answer)
     answer.question.validation_rules[:grouped_question]
   end
 
-  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Metrics/MethodLength,Lint/MissingCopEnableDirective
   def question_group_locals(surveys_question_group, index, total, is_results_view:)
```
ragesoss, line 87: I note that this helper is used in both the results view, and also in the survey/show template. Will this affect the behavior of the `show` route?

Abishekcs: No, this PR only deals with users and their notifications, which is how we get the user course data. The `show` route doesn't need any of that.

However, I checked the main branch and saw that it's already loading all users and survey notifications even for the `show` view—where they're not actually needed. So, to stay consistent, and to keep the PR small i kept it the same in this PR. I'll open a separate PR later to avoid loading users and notifications unnecessarily when rendering the `show` view.

Abishekcs: **Show route**


https://github.com/user-attachments/assets/b455ef71-5d8f-4231-be1d-5f5da4673a40


Screenshot from 2025-06-24 14-21-50

ragesoss: Okay. But This is adding more data fetching within the `question_group_locals` which is currently used by the `show` route, correct? So this will add even more unnecessary processing until the other PR is added?

Abishekcs: Actually the load time is same when for `show view` compared with the main branch(since the main branch also loads those data not required by show) i can add that commit in this PR it's just few if/unless statements?

Abishekcs: I can upload a screenshot for the load time of the `show page` for the main and this PR?

ragesoss: Yeah, if it's simple, please add it to this PR.

Abishekcs: It's same most of the time

**Main Branch**

Main Branch


**Current PR**

Current PR

Abishekcs: > Yeah, if it's simple, please add it to this PR.

Okay

Abishekcs:
@ragesoss I've pushed a commit to conditionally load users based on is_results_view.
This change has also reduced the load time for the survey show view.

Screenshot from 2025-06-24 20-26-17

ragesoss:
Okay! I'll merge this once the tests pass.

Abishekcs:
Fixing linting errors.

PR #6376, opened 2025-06-23
Optimize article deduplication with pluck and composite index

Abishekcs:
## What this PR does

This refactor optimizes the `articles_grouped_by_title_and_namespace` method by:

* Replacing Ruby-level `map` calls with SQL `.pluck` for extracting `title` and `namespace` directly.
* Leveraging the existing composite index on `(namespace, wiki_id, title)` for faster lookups.

## Screenshots

Before:

Screenshot from 2025-06-24 00-11-11

After:

Screenshot from 2025-06-24 00-44-26

### DIFF
--- a/lib/duplicate_article_deleter.rb
+++ b/lib/duplicate_article_deleter.rb
@@ -34,9 +34,10 @@ class DuplicateArticleDeleter
   private
 
   def articles_grouped_by_title_and_namespace(articles)
-    articles ||= Article.where(deleted: false, wiki_id: @wiki.id)
-    titles = articles.map(&:title)
-    Article.where(title: titles, wiki_id: @wiki.id).group(%w[title namespace]).count
+    Article.where(namespace: articles.pluck(:namespace).uniq,
+                  title: articles.pluck(:title),
+                  wiki_id: @wiki.id)
+           .group(%w[title namespace]).count
   end
 
   def delete_duplicates_in(article_group)


db/schema.rb
```diff
@@ -78,6 +78,7 @@
     t.index ["index_hash"], name: "index_articles_on_index_hash", unique: true
     t.index ["mw_page_id"], name: "index_articles_on_mw_page_id"
     t.index ["namespace", "wiki_id", "title"], name: "index_articles_on_namespace_and_wiki_id_and_title"
+    t.index ["title", "wiki_id", "namespace"], name: "index_articles_on_title_and_wiki_id_and_namespace"
```
ragesoss: This looks like a duplicate index, just with a different order. This shouldn't be necessary, right?

Abishekcs: yes but can we drop the old composite index on (namespace, wiki_id, title)? From what I can tell, the places where it’s used (I found around 4 using VSCode search) could easily switch to the new (title, wiki_id, namespace) index instead. The new one actually fits better with how we’re querying things now on other parts of the codebase too?

ragesoss: What is the functional difference between them?

Abishekcs:
```ruby
# Old index
t.index ["namespace", "wiki_id", "title"], name: "index_articles_on_namespace_and_wiki_id_and_title"

# New index
t.index ["title", "wiki_id", "namespace"], name: "index_articles_on_title_and_wiki_id_and_namespace"
```

The main difference is the **order of the columns**. In databases, composite indexes follow a *left-to-right-no-skipping* rule — we can't skip parts of the index. So the usefulness of the index depends on how the query is written.

For example:

* `.where(namespace: ..., wiki_id: ...)`
  → will work with the **old index**.

* `.where(title: ..., wiki_id: ...)`
  → won’t use the old index because the order doesn’t match — but the **new index** works great here.

---

We already have a few places in the codebase where the order is `title` → `wiki_id`, which means they can’t use the old index effectively.

For instance:

```ruby
def delete_duplicates(title, ns)
  articles = Article.where(title: title, wiki_id: @wiki.id, namespace: ns, deleted: false)
                    .order(:created_at)
```

Or:

```ruby
# somewhere else
def set_article_ids
  @article_ids = Article.where(title: @page_titles, wiki_id: @wiki.id).pluck(:id)
end
```

These queries won’t benefit from the `namespace, wiki_id, title` index — but **do benefit** from the new `title, wiki_id, namespace` index.

Abishekcs: There are four places shown by vscode which uses namspace but i am not sure how accurate searching like this will be

Screenshot from 2025-06-23 23-06-03

Abishekcs: And  removing the old composite index we have to make sure it doesnt degrade the existing querys using it.

ragesoss: Okay. This is, I think the largest table in the database (except the deprecated revisions table), so I'm wary of making changes to it for performance reasons (aka, worried that adding the index might take a long time and require downtime). At least some of the queries can likely be rewritten use the order of the existing index, right?

Can the articles_grouped_by_title_and_namespace be rewritten to use the existing index? We could pluck the namespaces from the `articles` (along with the titles), and then only search for duplicates in those namespaces... which in most cases will just be namespace 0.

Abishekcs: > At least some of the queries can likely be rewritten use the order of the existing index, right?

Yes

Abishekcs: > Can the articles_grouped_by_title_and_namespace be rewritten to use the existing index? We could pluck the namespaces from the `articles` (along with the titles), and then only search for duplicates in those namespaces... which in most cases will just be namespace 0.

Okay i will try this.

Abishekcs:
@ragesoss I’ve made the requested changes. Whenever you have time, could you please take another look at the code? If everything looks good, I’ll go ahead and update the PR description with the performance details.

Abishekcs:
> update the PR description with the performance details.

Done

lib/duplicate_article_deleter.rb
```diff
@@ -34,9 +34,10 @@ def resolve_duplicates_for_timeslices(articles)
   private
 
   def articles_grouped_by_title_and_namespace(articles)
-    articles ||= Article.where(deleted: false, wiki_id: @wiki.id)
-    titles = articles.map(&:title)
-    Article.where(title: titles, wiki_id: @wiki.id).group(%w[title namespace]).count
+    Article.where(namespace: articles.pluck(:namespace),
```
ragesoss: For the namespace, probably want to deduplicate it? Eg, `.pluck(:namespace).uniq`?

Might not matter performance-wise, but it makes for prettier SQL.

Abishekcs: Okay

ragesoss:
Thanks! Is the performance any difference compared to the version you had before with the replaced index?

Abishekcs:
> Thanks! Is the performance any difference compared to the version you had before with the replaced index?

No its the same.

PR #6382, opened 2025-06-25
Refactor how we handle non-scoped/scoped revisions in RevisionDataManager

gabina:
## What this PR does

This PR refactors how we handle non-scoped/scoped revisions. Before this PR, we maintained both hashes in memory (one with all the revisions, one with only scoped revisions). When creating revisions, we used to set the scoped field based on a private `RevisionDataManager` method called `scoped_revision?`, that basically checked if the given revision is in the scoped revisions hash.

After this PR, instead of having two hashes in memory, we only keep one (with all revisions), but adding a new `scoped` field inside the article data. Example:
```
[
        [ "55345266",
        { 
                "article"=>{"mw_page_id"=>"55345266", "title"=>"Ragesoss/citing_sources", "namespace"=>"2", "wiki_id"=>1, "scoped"=>true},
                "revisions"=>
                [
                        {"mw_rev_id"=>"849116430", "date"=>Fri, 06 Jul 2018 16:58:15 +0000, "characters"=>"569", "mw_page_id"=>"55345266", "username"=>"Ragesoss", "new_article"=>"false", "system"=>"false", "wiki_id"=>1}, 
                        {"mw_rev_id"=>"849116480", "date"=>Fri, 06 Jul 2018 16:58:46 +0000, "characters"=>"17", "mw_page_id"=>"55345266", "username"=>"Ragesoss", "new_article"=>"false", "system"=>"false", "wiki_id"=>1}, 
                        {"mw_rev_id"=>"849116533", "date"=>Fri, 06 Jul 2018 16:59:17 +0000, "characters"=>"19", "mw_page_id"=>"55345266", "username"=>"Ragesoss", "new_article"=>"false", "system"=>"false", "wiki_id"=>1}, 
                        {"mw_rev_id"=>"849116572", "date"=>Fri, 06 Jul 2018 16:59:37 +0000, "characters"=>"6", "mw_page_id"=>"55345266", "username"=>"Ragesoss", "new_article"=>"false", "system"=>"false", "wiki_id"=>1}
                ]
        }
        ]
]
 
```     

This reduces the memory used for allocating revisions, although for many courses it should not be such a large amount of memory (since they are unscored revisions). However, I believe it was an inefficient approach. These changes also mean that the `CustomRevisionFilter` module does not longer need to know about the complex structure of these hashes, keeping the logic more encapsulated.


gabina:

PR #6384, opened 2025-06-25
[WIP] Remove ApplicationRecord inheritance for Revisions

gabina:
## What this PR does
WIP

### DIFF
--- a/app/models/wiki_content/revision_on_memory.rb
+++ b/app/models/wiki_content/revision_on_memory.rb
@@ -1,15 +1,43 @@
 # frozen_string_literal: true
 
+#  id                :integer          not null, primary key
+#  characters        :integer          default(0)
+#  created_at        :datetime
+#  updated_at        :datetime
+#  user_id           :integer
+#  article_id        :integer
+#  views             :bigint           default(0)
+#  date              :datetime
+#  new_article       :boolean          default(FALSE)
+#  deleted           :boolean          default(FALSE)
+#  wp10              :float(24)
+#  wp10_previous     :float(24)
+#  system            :boolean          default(FALSE)
+#  ithenticate_id    :integer
+#  wiki_id           :integer
+#  mw_rev_id         :integer
+#  mw_page_id        :integer
+#  features          :text(65535)
+#  features_previous :text(65535)
+#  summary           :text(65535)
+
 class RevisionOnMemory
   include ActiveModel::Model
+  include ActiveModel::Attributes
   # Helps with importing data
   alias_attribute :rev_id, :mw_rev_id
 
   validates :mw_page_id, presence: true
   validates :mw_rev_id, presence: true
 
-  attr_accessor :mw_rev_id, :date, :characters, :article_id, :mw_page_id, :user_id, :new_article,
-                :system, :wiki_id, :scoped, :features, :features_previous, :summary, :error
+  attribute :mw_rev_id, :integer
+  attribute :features, default: -> { {} }
+  attribute :features_previous, default: -> { {} }
+  attribute :summary, default: -> { {} }
+
+  attr_accessor :date, :characters, :article_id, :mw_page_id, :user_id, :new_article,
+                :system, :wiki_id, :scoped, :error,
+                :wp10, :wp10_previous, :deleted
 
   ####################
   # Instance methods #
--- a/app/models/wiki_content/revision_on_memory.rb
+++ b/app/models/wiki_content/revision_on_memory.rb
@@ -34,8 +34,9 @@ class RevisionOnMemory
   attribute :features, default: -> { {} }
   attribute :features_previous, default: -> { {} }
   attribute :summary, default: -> { {} }
+  attribute :characters, :integer, default: 0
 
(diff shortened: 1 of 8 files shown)


app/services/update_wikidata_stats_timeslice.rb
```diff
@@ -77,7 +77,7 @@ def initialize(course)
   # Returns the updated array.
   def update_revisions_with_stats(revisions)
     # We will only use the diff stats for in-scope revisions, and this is very slow.
-    revision_ids = revisions.select(&:scoped).pluck(:mw_rev_id)
+    revision_ids = revisions.select(&:scoped).map(&:mw_rev_id)
```
gabina, line 80: pluck only works for active records

app/models/alert.rb
```diff
@@ -118,6 +117,15 @@ def user_contributions_url
     courses_user&.contribution_url
   end
 
+  # Returns the web diff url for the revision, e.g.,
+  # https://en.wikipedia.org/w/index.php?title=Eva_Hesse&diff=prev&oldid=655980945
+  # Note: this will break the url for old alerts
+  def revision_url
+    return if article.nil? || revision_id.nil?
+    title = article.escaped_full_title
+    "#{article.wiki.base_url}/w/index.php?title=#{title}&diff=prev&oldid=#{revision_id}"
```
gabina, line 126: Something that took my attention is that `revision_id` field contains internal revision ids in some cases and mediawiki revision ids in other cases. Example:

- Alert id 79361. Article id 52464555 (Mohammad_Reza_Zahedi). Revision id 4347612. Type `DiscretionarySanctionsEditAlert`. The url would fail in this case because the mediawiki revision id is [1217351095](https://en.wikipedia.org/w/index.php?title=Mohammad_Reza_Zahedi&oldid=1217351095).
- Alert id 80419. Article id 773674490 (Suzan_Daniel). Revision id [1277251385](https://en.wikipedia.org/w/index.php?title=Suzan_Daniel&oldid=1277251385). Type `DeUserfyingAlert`. The url would fail in this case because `revision_id` is a real wikimedia revision id.

gabina: Notice that all the alerts created with `revision_id` different from null right now are using media wiki revision ids, as we refactored them to work without revisions table. However, some historical alerts may have an incorrect revision url.

ragesoss: Noted. I think this is okay.

gabina:

app/models/wiki_content/revision_on_memory.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+class RevisionOnMemory
+  include ActiveModel::Model
+  include ActiveModel::Attributes
```
ragesoss, line 7: It would be helpful to include some explanation of what features from these two modules that we are relying on, as later on we might want to try to remove them if the required methods are narrow enough.

gabina:
> This looks great. I note that there's one other stray bit of dead code related to the dyk_eligible json template: fetchDykArticles in api.js. May as well remove that now as well.

Good find. Done in af95b0608b53138ad908c51a0a4921f45b32747a

app/models/wiki_content/revision_on_memory.rb
gabina: Good point. In fact, I just replaced the `ActiveModel::Model` inclusion to `ActiveModel::Validations` since it looks more specific. I think we only need this to define validations and attributes. Let's see if specs pass. BTW, I'm not sure if validations and attribute aliases are really necessary now, but I'm afraid of removing them and breaking something. I mean, I suspect we'd even be okay with just this:

```
class RevisionOnMemory
  # Include this to define attributes
  include ActiveModel::Attributes

  attribute :mw_rev_id, :integer
  attribute :features, default: -> { {} }
  attribute :features_previous, default: -> { {} }
  attribute :characters, :integer, default: 0

  attr_accessor :date, :article_id, :mw_page_id, :user_id, :new_article,
                :system, :wiki_id, :scoped, :error, :summary,
                :wp10, :wp10_previous, :deleted
```

Check commit 53b265aaf4fc48892c468c49e5dab2f6cb5b1f32.

ragesoss: yeah, whether and how we were using validations is the main thing I was wondering about. It's fine to be conservative about not breaking anything for the initial implementation, but we have at least one test (I think?) that goes through the entire stats generation process, so it should be relatively safe to remove anything that doesn't break the tests.

gabina: Well.. specs did their. It looks like we do need the `ActiveModel::Model` inclusion because it adds the initializer.

As regards the validations, although we have specs, I was afraid that it would be used if some API response did not contain a field because of an error or something like that. However, it seems that such validations were added many years ago (commit e57e2bafc58cc69f496ff3791511872344e207bb) after refactoring revisions, articles, or wikis. 

I'm re-including `ActiveModel::Model` but deleting the validations. Let's see if specs pass. I'm keeping attribute aliases. Commit b46090f5c94fc252d501b24be07704b2e8bb5746

PR #6389, opened 2025-07-01
Improve error handling when refreshing categories

gabina:
## What this PR does
This PR improves error handling when refreshing categories. It was created after some problems we had with petscan API (see PR #6387) but it works for categories no matter the source.
- On one hand, we were worried about importing categories with such a huge number of articles that they cannot be saved in the `article_titles` field (which is `mediumtext`, that means 16,777,215 chars). The discussed approach to "deal" with this situation - before migrating the db to change the field size- is to detect these cases and mark the category with error (by changing the source field to `previous_source + 'Error'`). This PR adds `rescue ActiveRecord::ValueTooLong` statement. That's the error caused when trying to run `save` when the title array is too long. Now, when that error is identified, we only modify the source, avoiding future refreshing for that category.
- On the other hand, this PR adds an empty `rescue StandardError` statement and ensures Petscan class raises the error if the given API fails. This is to prevent the `article_titles` field to be cleared when some unexpected error happens hitting the API. `TYPICAL_ERRORS` remains existing because `ApiErrorHandling` class needs it to choose the log level. A similar thing was done for `PagePileApi`.

In addition, this PR makes frontend changes to show errors for categories with too many articles to the user know what's going on.

### How this was reproduced locally
We're being proactive in handling categories with too many articles to fit in `article_titles`. Because of that, I didn't have any specific category to reproduce the error. I played around with petscan to find a query returning something obviously greater than `mediumtext`, but when I added many levels of depth to the category (to make it have more articles) the query never finished. So I used a "normal" petscan category and manually enlarged that array (I concatenated it several times with itself). When I did that, the first error I found was the following:
```
ActiveRecord::ConnectionNotEstablished: MySQL client is not connected
from /home/gabina/.rvm/gems/ruby-3.1.2/gems/mysql2-0.5.4/lib/mysql2/client.rb:148:in _query'
Caused by Mysql2::Error: MySQL client is not connected
from /home/gabina/.rvm/gems/ruby-3.1.2/gems/mysql2-0.5.4/lib/mysql2/client.rb:148:in _query'
Caused by ActiveRecord::StatementInvalid: Mysql2::Error::ConnectionError: Got a packet bigger than 'max_allowed_packet' bytes
from /home/gabina/.rvm/gems/ruby-3.1.2/gems/mysql2-0.5.4/lib/mysql2/client.rb:148:in _query'
Caused by Mysql2::Error::ConnectionError: Got a packet bigger than 'max_allowed_packet' bytes
from /home/gabina/.rvm/gems/ruby-3.1.2/gems/mysql2-0.5.4/lib/mysql2/client.rb:148:in _query'
```
I had `max_allowed_packet` variable set to 16,777,216 bytes locally, so I had to change it to the production value, which is 1,073,741,824 bytes. Once I updated that, I found the real `ActiveRecord::ValueTooLong` error.

### Before
image

### After
image

### DIFF
--- a/app/assets/javascripts/components/categories/category.jsx
+++ b/app/assets/javascripts/components/categories/category.jsx
@@ -19,16 +19,34 @@ const Category = ({ course, category, remove, editable }) => {
   const catName = CourseUtils.formattedCategoryName(category, course.home_wiki);
   let depth;
   let link;
-  if (category.source === 'category') {
+  if (category.source.includes('category')) {
     depth = category.depth;
   }
-  if (category.source === 'psid') {
+  if (category.source.includes('psid')) {
     link = `https://petscan.wmcloud.org/?psid=${category.name}`;
-  } else if (category.source === 'pileid') {
+  } else if (category.source.includes('pileid')) {
     link = `https://pagepile.toolforge.org/api.php?id=${category.name}&action=get_data`;
   } else {
     link = `https://${course.home_wiki.language}.${course.home_wiki.project}.org/wiki/${catName}`;
   }
+  const errored = category.source.includes('Error');
+
+  const cateNameLink = (
+    <a target="_blank" href={link}>
+      {catName}
+    </a>
+  );
+
+  const cateNameLinkWarning = (
+    <div className="tooltip-trigger">
+      {cateNameLink}
+      <span className="tooltip-indicator" />
+      <div className="tooltip dark">
+        <p>{I18n.t('categories.category_error')}</p>
+      </div>
+    </div>
+  );
+
   const lastUpdate = toDate(category.updated_at);
   let lastUpdateMessage;
   const beenUpdated = !isSameDay(toDate(category.created_at), lastUpdate);
@@ -41,11 +59,9 @@ const Category = ({ course, category, remove, editable }) => {
   }
   return (
     <>
-      <tr>
+      <tr className={errored ? 'warning-background' : undefined}>
         <td>
-          <a target="_blank" href={link}>
-            {catName}
-          </a>
+          {errored ? cateNameLinkWarning : cateNameLink}
         </td>
         <td>{depth}</td>
         <td>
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/components/categories/category.jsx
```diff
@@ -19,16 +19,34 @@ const Category = ({ course, category, remove, editable }) => {
   const catName = CourseUtils.formattedCategoryName(category, course.home_wiki);
   let depth;
   let link;
-  if (category.source === 'category') {
+  if (category.source.includes('category')) {
```
gabina, line 22: Using `includes` now just in case the source was updated to `categoryError`

app/models/wiki_content/category.rb
```diff
@@ -66,6 +73,8 @@ def self.refresh_categories_for(course, update_service: nil)
   end
 
   def refresh_titles(update_service: nil)
+    # Do not try to refresh titles for error sources
+    return unless VALID_SOURCES.include? source
```
gabina, line 77: The safety guard is here because this is the only method that calls private method `title_list_from_wiki`  (which doesn't  handle error sources).

app/models/wiki_content/category.rb
```diff
@@ -75,6 +84,15 @@ def refresh_titles(update_service: nil)
...
     # rubocop:enable Rails/SkipsModelValidations
+  rescue ActiveRecord::ValueTooLong
+    # This means we couldn't save the article titles in the db because they're too long.
+    # Revert article_titles change
+    self.article_titles = attribute_was(:article_titles)
+    # Update source to indicate it's a source with error
+    self.source = source + 'Error'
+    save
+  rescue StandardError
+    # If something went wrong when refresing titles, we don't want to replace article_titles field.
```
gabina, line 95: `ActiveRecord::ValueTooLong` is expected to be raised by the save call.

`StandardError` is expected if the API fails while refreshing the category.

gabina:

ragesoss:
This looks great, thanks!

PR #6390, opened 2025-07-02
Fully Utilise composite index of Articles while joining ArticlesCourses with Articles in DYKNominationMonitor#create_alerts_from_page_titles

Abishekcs:
## What this PR does

DYKNominationMonitor#create_alerts_from_page_titles previously didn't use  composite index for Article model while ArticlesCourses model was being joined with Article model leading to inefficient SQL operation.With this changes i.e including namespace also in the where clause of Article it will fully utilise existing composite index improving query time significantly.

## Screenshots

Before:

After:


lib/alerts/d_y_k_nomination_monitor.rb
```diff
@@ -17,9 +17,10 @@ def initialize
   end
 
   def create_alerts_from_page_titles
+    namespace = Article::Namespaces::MAINSPACE
     course_articles = ArticlesCourses.joins(:article)
                                      .where(course: Course.current)
-                                     .where(articles: { title: @page_titles, wiki_id: @wiki.id })
+                                     .where(articles: { title: @page_titles, wiki_id: @wiki.id, namespace: }) # rubocop:disable Layout/LineLength
```
ragesoss, line 23: Does the order not matter here on the Rails side?, as long each consecutive element of the index is part of the query?

Abishekcs: order wont matter rails adjust it as long as all the index are used.

PR #6391, opened 2025-07-02
Part 1: Remove redundant index on course_id from assignments table

Abishekcs:
Part 1 #6391
Part 2 #6392
Part 3 #6393
Part 4 #6394
Part 5 #6395
Part 6 #6396
Part 7 (FINAL) #6397

## What this PR does

This PR removes the `index_assignments_on_course_id` from the `assignments` table.

The index was redundant due to the presence of a more selective composite index: `index_assignments_on_course_id_and_user_id`, which already covers the same leading column (`course_id`). Removing this standalone index helps reduce duplication and optimizes index maintenance during write operations.

* Removed: `index_assignments_on_course_id`

No application logic is affected by this change.

### DIFF
--- a/db/migrate/20250702174753_remove_index_assignments_on_course_id.rb
+++ b/db/migrate/20250702174753_remove_index_assignments_on_course_id.rb
@@ -1,0 +1,5 @@
+class RemoveIndexAssignmentsOnCourseId < ActiveRecord::Migration[7.0]
+  def change
+    remove_index :assignments, name: "index_assignments_on_course_id"
+  end
+end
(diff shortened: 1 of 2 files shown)


ragesoss:
Thanks. I'm not going to merge this now, but I'll try to do so at a time when I'm ready to deal with migrations.

Abishekcs:
@ragesoss Do you think it would be useful to merge this PR `(1-7)`? We can also close it, since I don't think merging it will make any noticeable difference either way.

ragesoss:
Yeah, I guess we should close them. We can also reimplement them pretty easily later on.

Issue #5956, opened 2024-09-10
Prevent duplicate  `articles_courses` rows.


gabina:
### What is happening?
While we're using [sidekiq-unique-jobs](https://github.com/mhenrixon/sidekiq-unique-jobs) gem to add unique constraints to sidekiq jobs, and we currently set `sidekiq_options lock: :until_executed` in `CourseDataUpdateWorker` worker to avoid  duplicate jobs (with the same arguments and job class), it's still possible to end up with duplicate `articles_courses` rows.
One possible scenario for this is when running a manual update course stats process through the `manual_update` route. As this update runs synchronously there is no guarantee that a queued `UpdateCourseStats` job will run simultaneously.

### To Reproduce

> Steps to reproduce the behavior:
As the problem is related to a race condition, it's not trivial to reproduce. However, this is one scenario that may show the problem:
> 1. Start the app locally.
> 2. Be sure `sidekiq` is up (`bundle exec sidekiq start`).
> 3. Download a course locally (you can do this through the "Copy course from another server" admin view).
> 4. Add the course to default campaign.
> 5. Run a manual update using the `manual_update` route (you need to be an admin for that).

If an update for that course was scheduled and run at the same time than your synchronous update, then duplicate `articles_courses` rows may be created. 

### Expected behavior
We should avoid having duplicate `articles_courses` rows (i.e., multiple rows with the same `course_id` and `article_id`). This can be prevented by adding a unique constraint on (`course_id`, `article_id`) in the `articles_courses` table. Proper tests should be done before migrating `articles_courses` table.

Formasitchijoh:
Hello @ragesoss  i would like to work on this

Formasitchijoh:
Hello @ragesoss  working on this

cr3ativ3cod3r:
I am working on this issue

ragesoss:
@cr3ativ3cod3r this is probably not a great issue for a new contributor, as it will require some manual testing against a realistically-sized database in a production-like environment.

PR #6401, opened 2025-07-07
Optimize Article Batch Performance by Removing distinct Dependency

Abishekcs:
## Note (IMPORTANT)

The optimization assumes that the `ArticlesCourses` table does **not** contain duplicate `(course_id, article_id)` pairs. If such duplicates exist, the behavior may be incorrect due to missing deduplication in SQL.

If deduplication is still needed, maybe we can shift that logic to Ruby (e.g., using `.uniq`) after querying — this allows us to retain the fast, index-only query path without `DISTINCT`.

## Example from production log

**1)** [Example from production slow log.txt](https://github.com/user-attachments/files/21101241/Example.from.production.slow.log.txt)

**2)** [Time: 250629  3:48:06.txt](https://github.com/user-attachments/files/21321430/Time.250629.3.48.06.txt)

Screenshot from 2025-07-07 16-36-13

Screenshot from 2025-07-18 23-53-59

## What this PR does

This PR introduces a new `has_many :all_articles` association on the `Course` model that bypasses the `distinct` constraint on the existing `articles` association.

Using the original `@course.articles` with `.in_batches` caused ActiveRecord to generate `SELECT DISTINCT` queries. This triggered the use of **temporary tables** and **increased query latency** significantly.

**Before (with `DISTINCT`):**

* Queries used: `Using temporary`
* Execution times: \~30ms per batch
* Rows scanned: often 15K+ per 1000 actual rows needed

**After (with `JOIN` only):**

* Queries use: `Using index`, no temporary tables
* Execution times: \~2–7ms per batch
* Rows scanned: exactly 1000 (as expected)

## Screenshots

**Before:**

Before 1

Before 1

**After:**

Screenshot from 2025-07-19 00-14-06

Before 2

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -89,6 +89,7 @@ class Course < ApplicationRecord
 
   has_many :articles_courses, class_name: 'ArticlesCourses', dependent: :destroy
   has_many :articles, -> { distinct }, through: :articles_courses
+  has_many :all_articles, through: :articles_courses, source: :article
   has_many :assignments, dependent: :destroy
 
   has_many :categories_courses, class_name: 'CategoriesCourses', dependent: :destroy
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -88,8 +88,7 @@ class Course < ApplicationRecord
   end, through: :students)
 
   has_many :articles_courses, class_name: 'ArticlesCourses', dependent: :destroy
-  has_many :articles, -> { distinct }, through: :articles_courses
-  has_many :all_articles, through: :articles_courses, source: :article
+  has_many :articles, through: :articles_courses
   has_many :assignments, dependent: :destroy
 
   has_many :categories_courses, class_name: 'CategoriesCourses', dependent: :destroy
--- a/lib/articles_courses_cleaner_timeslice.rb
+++ b/lib/articles_courses_cleaner_timeslice.rb
@@ -159,8 +159,6 @@ class ArticlesCoursesCleanerTimeslice # rubocop:disable Metrics/ClassLength
   end
 
   def reset_articles_in_untracked_namespaces
-    # Get relevant article IDs for the course
-
     @course.all_articles.in_batches do |article_batch|
       tracked = @course.tracked_namespaces.each.flat_map do |wiki_ns|
         wiki_id = wiki_ns[:wiki].id
--- a/lib/articles_courses_cleaner_timeslice.rb
+++ b/lib/articles_courses_cleaner_timeslice.rb
@@ -160,9 +160,8 @@ class ArticlesCoursesCleanerTimeslice # rubocop:disable Metrics/ClassLength
 
   def reset_articles_in_untracked_namespaces
     # Get relevant article IDs for the course
-    article_ids = ArticlesCourses.where(course_id: @course.id).pluck(:article_id)
 
-    Article.where(id: article_ids).in_batches do |article_batch|
+    @course.all_articles.in_batches do |article_batch|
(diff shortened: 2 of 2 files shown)


ragesoss:
I think we've designed ArticlesCourses and the related processes around the implicit assumption that there is just one record for any course_id-article_id pair, but we don't have a database constraint on that. I can't think of any situations where duplicate records might get created, but it might be worth checking whether any such cases exist in production.

Can you provide a reasonably efficient query that I could use to check whether there any such duplicates in production?

Abishekcs:
> Can you provide a reasonably efficient query that I could use to check whether there any such duplicates in production?

```sql

SELECT course_id, article_id, COUNT(*) AS occurrences
FROM articles_courses
GROUP BY course_id, article_id
HAVING COUNT(*) > 1;


```

I think this should work and is also efficient

Screenshot from 2025-07-21 23-05-44

ragesoss:
There are 400k instances of duplication on Programs & Events Dashboard (and 8 on Wiki Education Dashboard, all from old courses). There are many from there recent course with the super-active bot account JarBot: https://outreachdashboard.wmflabs.org/courses/Levant/Editing_and_Contributing_Training_-_University_Students/students/overview

I suspect that these are an artifact of update processes that died before finishing, but it looks like we don't have a process that cleans up the duplicates (as that example course has been undergoing successful updates recently).

I think we should put this PR in statis and open an issue about the duplicate records. @gabina any thoughts?

ragesoss:
(Query took 5 minutes and 12 seconds in P&E production.)

gabina:
I ran the following query and found that the last duplicate article course record for course 31480 was created on 2025-04-02. I think that makes sense because the course was created on 2025-03-29 and it is likely that it was still being processed frequently during the first few days of April.

```
SELECT course_id, article_id, COUNT(*) AS occurrences, max(created_at), min(created_at)
FROM articles_courses
WHERE course_id = 31480
GROUP BY course_id, article_id
HAVING COUNT(*) > 1
ORDER by 4 ASC;
```

I also ran the following query and found that the maximum difference between the creation of a record and its duplicate is 1 second.

```
SELECT course_id, article_id, first_created, second_created, TIMESTAMPDIFF(SECOND, second_created, first_created)
FROM
(SELECT course_id, article_id, COUNT(*) AS occurrences, max(created_at) first_created, min(created_at) second_created
FROM articles_courses
WHERE course_id = 31480
GROUP BY course_id, article_id
HAVING COUNT(*) > 1) t
ORDER by 5 ASC;
```

I think everything points out that there were two or three process running at the same time (issue #5956).

In order to fix this we should:
1- Find a way to avoid having new duplicate `articles_courses` rows. I guess the easiest way is to add a unique constraint on (course_id, article_id) in the `articles_courses` table. 
2- Remove existing duplicate records. We should create a one-off script for this. Although I would like to test this locally first, I don't think removing duplicates is tricky. I can't think of any reason it would matter whether we delete the first, last, or even the middle record (in case of three duplicates), since they should be identical except for id (which I believe is never used), `created_at` and `updated_at` (which are very similar or even equal). Note that ACT points to `article_id` ,not to `articles_courses_id`.

I can start by reproducing the error locally, and writing and testing the one-off script to fix it. Adding the unique constraint would involve some test in the db to be sure it's reasonable (`articles_courses` table has 17,496,073 rows). I think we already ran a migration over `articles_courses` table when deploying data-rearchitecture and it worked smoothly.

gabina:
Opened PRs #6432 and #6433 to delete duplicates through a one-off script and to add a database unique constraint to prevent new ArticlesCourses duplication. After deploying and running migration, I think we would be OK to remove the distinct clause.

PR #6402, opened 2025-07-08
Optimize campaign articles to use deferred joins and structured data loading

Abishekcs:
## References

- [Deferred joins (Planetscale)](https://planetscale.com/learn/courses/mysql-for-developers/examples/deferred-joins)
- [Efficient MySQL pagination using deferred joins -  Aaron](https://aaronfrancis.com/2022/efficient-mysql-pagination-using-deferred-joins-15d0de14)
- [Deferred Join: A Deep Dive](https://hackmysql.com/deferred-join-deep-dive/)

## What this PR does

Refactors the campaign articles functionality to use deferred joins and structured data loading, significantly improving performance for large datasets while maintaining the same user experience and improving page load time.

The previous implementation had performance issues when dealing with large numbers of articles, particularly during pagination and sorting operations. The query approach was causing slow page loads for campaigns with extensive article collections.

## Solution
Implemented a deferred join pattern with structured data separation:

#### 1. **Deferred Join Pattern**
- Replaced direct query with `ranked_articles_subquery` that pre-filters and ranks articles
- Uses `INNER JOIN` with the ranked subquery for optimal performance
- Maintains conditional sorting logic based on article count

#### 2. **Structured Data Loading**
- `campaign_articles` now returns a hash with three separate collections:
  - `articles_courses`: Paginated articles_courses data
  - `courses`: Indexed courses by ID for efficient lookup
  - `articles`: Indexed articles by ID for efficient lookup

#### 3. **Custom Pagination**
- Implements manual pagination using `WillPaginate::Collection`
- Proper offset/limit handling with `total_articles_count`
- Maintains existing pagination interface in views

#### 4. **Optimized Data Fetching**
- Uses `.select()` to load only required fields
- Implements instance variable caching for expensive operations
- Separates concerns with dedicated methods for each data type

### Performance Improvements
- **Query Optimization**: Deferred join reduces query complexity for large datasets
- **Memory Efficiency**: Loads only necessary fields and data structures
- **Caching**: Instance variables prevent redundant database calls
- **Scalability**: Handles large article counts without performance degradation

### View Updates
Updated templates to work with the new data structure:
- `articles.html.haml`: Uses structured data from the hash
- `_article_row.html.haml`: Accepts pre-loaded course and article objects
- Maintains existing UI/UX with no user-facing changes

### Method Organization
The refactored presenter follows a logical data flow:
1. **Main interface**: `campaign_articles`
2. **Data sources**: `campaigns_courses_ids`, `load_courses`
3. **Query builders**: `ranked_articles_subquery`, `articles_courses_scope`
4. **Data fetchers**: `fetched_articles`
5. **Pagination**: `paginated_articles_courses`, `total_articles_count`
6. **Helpers**: `per_page`, `current_page`, `offset`

### Database Impact
- Reduces complex joins through deferred pattern
- Maintains existing indexes and relationships
- No schema changes required

## Testing
- Verify pagination works correctly for large datasets
- Confirm sorting behavior matches previous implementation
- Test performance with campaigns containing 1000+ articles
- Ensure view rendering remains consistent

## Screenshots  `(WITH SORTING)`

**Before:**

**`PAGE 1`**  **(4841 ms)**

Screenshot from 2025-07-08 22-25-49

**`PAGE 1197`**  **(5870 ms)**

Screenshot from 2025-07-08 22-27-28

**After:** 

**`PAGE 1`**  **(732 ms)**

Screenshot from 2025-07-08 22-20-37

**`PAGE 1197`**  **(831 ms)**

Screenshot from 2025-07-08 22-22-26

## Screenshots  `(WITHOUT SORTING)`

**BEFORE**

**`PAGE 1`**  **(430 ms)**

Screenshot from 2025-07-09 17-52-56

**`PAGE 1197`** **(2068 ms)**

Screenshot from 2025-07-09 17-50-41

**AFTER**

**`PAGE 1`**  **(307 ms)**

Screenshot from 2025-07-09 17-26-46

**`PAGE 1197`**  **(268 ms)**

Screenshot from 2025-07-09 17-28-31

### DIFF
--- a/app/presenters/courses_presenter.rb
+++ b/app/presenters/courses_presenter.rb
@@ -2,6 +2,7 @@
 
 require_dependency "#{Rails.root}/lib/word_count"
 require_dependency "#{Rails.root}/lib/analytics/histogram_plotter"
+require_dependency "#{Rails.root}/app/presenters/query/ranked_articles_courses_query"
 
 #= Presenter for courses / campaign view
 class CoursesPresenter
@@ -43,7 +44,7 @@ class CoursesPresenter
 
     {
       articles_courses: paginated_articles_courses,
-      courses: load_courses.index_by(&:id),
+      courses: course_ids_and_slugs.index_by(&:id),
       articles: fetched_articles
     }
   end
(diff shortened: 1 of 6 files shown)


app/presenters/courses_presenter.rb
```diff
@@ -41,13 +41,78 @@ def campaign
...
+    end
+  end
+
+  # Count total articles for pagination metadata
+  def total_articles_count
+    ArticlesCourses.where(course_id: load_courses.map(&:id), tracked: true).count
+  end
+
+  # Number of items per page
+  def per_page
```
ragesoss: Use a constant.

app/presenters/courses_presenter.rb
```diff
@@ -41,13 +41,78 @@ def campaign
...
+  # Create deferred join scope using ranked subquery for performance
+  def articles_courses_scope
+    @articles_courses_scope ||= ArticlesCourses
+                                .joins("INNER JOIN (#{ranked_articles_subquery.to_sql}) AS ranked_articles USING (id)") # rubocop:disable Layout/LineLength
+                                .select(:article_id, :course_id, :character_sum, :references_count, :view_count) # rubocop:disable Layout/LineLength
+  end
+
+  # Fetch and index articles by ID for efficient lookup
+  def fetched_articles
+    @fetched_articles ||= begin
```
ragesoss: I'd prefer to avoid doing this with a `begin` block. Instead, you can start the method with something like `return @fetched_articles unless @fetched_articles.nil?`

app/presenters/courses_presenter.rb
```diff
@@ -41,13 +41,78 @@ def campaign
...
+    }
+  end
+
+  # Fetch course IDs associated with the campaign
+  def campaigns_courses_ids
+    @campaigns_courses_ids ||= CampaignsCourses.where(campaign_id: campaign.id).select(:course_id).pluck(:course_id) # rubocop:disable Layout/LineLength
+  end
+
+  # Load only needed course data, grouped by ID
+  def load_courses
```
ragesoss: In idiomatic Rails code, a method with a name like would load data but not be used for its return value. A better method name would be one that describes what it returns, perhaps `course_ids_and_slugs`.

ragesoss:
This significantly increases the size and complexity of CoursesPresenter. Is there a clean way to extract the code for the main complex query to a separate file?

Abishekcs:
> This significantly increases the size and complexity of CoursesPresenter. Is there a clean way to extract the code for the main complex query to a separate file?

I can try to do that .

ragesoss:
This looks good to me. It still adds a lot of code to CoursesPresenter, which makes it larger and more complex than I'd like... especially because it is used in several different contexts but the new code is only for use with campaign views. Most of the logic was shared between the different uses before, but this change should prompt us to think about how to better organize the code so that we don't have one file pulled in multiple directions, perhaps by extracting the shared code between the use cases and implementing a separate CampaignCoursesPresenter that includes the campaign-specific parts. However, that's not necessary immediately, and I'm fine with merging in the current state.

@Formasitchijoh anything you spot that should be improved before merging?

Abishekcs:
> how to better organize the code

I'll take a look and see what I can do, now that I'm more familiar with this part of the codebase. I'll handle it in a separate PR. Thank you for the review.

Abishekcs:
### **UPDATE:**

**This PR not only resolves [Sentry#29f69f95](https://wiki-education.sentry.io/issues/6388168027/?project=6269916&query=is%3Aunresolved&referrer=issue-stream&stream_index=0), but also addresses a slow query from the MySQL slow log that I came across today (shown below). This issue is likely resolved by the newly introduced [`paginated_articles_courses`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6402/files#diff-e132e74b4f6657fb776d13c16588ee540fdf249594d26e586fa66947083156e0R88) method in this PR.**

[select_count_distinct_articles_courses_id.txt](https://github.com/user-attachments/files/21349964/select_count_distinct_articles_courses_id.txt)


### Screenshots:

`a) MYSQL SLOW LOG QUERY`

Screenshot from 2025-07-21 20-10-34


`b) Screenshot from my local development log`

Screenshot from 2025-07-21 20-15-48


`c) Source code responsible for the slowquery Highlighted`  (views/campaigns/articles.html.haml [Line 64])

Screenshot from 2025-07-21 20-19-38

PR #6404, opened 2025-07-11
[WIP]: Optimize article_course_timeslices queries by adding composite index on course_id and tracked

Abishekcs:
## What this PR does

This PR adds a composite index on the `article_course_timeslices table` for the `(course_id, tracked)` columns to improve the performance of queries that filter by both fields.

## Screenshot From Production Log

Screenshot from 2025-07-11 20-48-18

## Screenshot

**Code generating this query**

Screenshot from 2025-07-11 21-22-14

**Before:**

Screenshot from 2025-07-11 21-11-46

**After:**

Screenshot from 2025-07-11 21-05-48

### DIFF
--- a/db/migrate/20250711134708_add_index_to_article_course_timeslices_on_course_id_and_tracked.rb
+++ b/db/migrate/20250711134708_add_index_to_article_course_timeslices_on_course_id_and_tracked.rb
@@ -1,0 +1,5 @@
+class AddIndexToArticleCourseTimeslicesOnCourseIdAndTracked < ActiveRecord::Migration[7.0]
+  def change
+    add_index :article_course_timeslices, [:course_id, :tracked], name: :article_course_timeslice_by_course_id_and_tracked
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2024_05_15_163420) do
+ActiveRecord::Schema[7.0].define(version: 2025_07_11_134708) do
   create_table "alerts", id: :integer, charset: "utf8mb4", force: :cascade do |t|
     t.integer "course_id"
     t.integer "user_id"
@@ -47,6 +47,7 @@ ActiveRecord::Schema[7.0].define(version: 2024_05_15_163420) do
     t.datetime "updated_at", null: false
     t.datetime "first_revision"
     t.index ["article_id", "course_id", "start", "end"], name: "article_course_timeslice_by_article_course_start_and_end", unique: true
+    t.index ["course_id", "tracked"], name: "article_course_timeslice_by_course_id_and_tracked"
     t.index ["course_id", "updated_at", "article_id"], name: "article_course_timeslice_by_updated_at"
   end
 


Abishekcs:
Hi @gabina, could you please review this PR when you have a moment? Thank you so much!

@ragesoss, I believe the failing tests may not be related to this PR. Thank you!

gabina:
This is an interesting case.

When I was working on the `data-rearchitecture` instance I created this same index (PR #6039) to try to speed up pre-update process but I removed it 3 months later because "it wasn't used according to index stats" (PR #6198). Apparently, _I enabled `userstat` and checked `SHOW INDEX_STATISTICS;`. `article_course_timeslice_by_tracked` index didn't seem to be very useful._ TBH, I don't remeber a lot about that .

I think creating indexes over low cardinality columns is usually not recommended, but for composite indexes is OK (examples: [this stackoverflow post](https://stackoverflow.com/questions/61027237/does-it-make-sense-to-make-a-composite-index-of-2-columns-when-one-of-the-column), this [MariaDB video](https://www.youtube.com/watch?v=x-5hDEvQhuM)). In addition, most of the rows should have `tracked` = true, so I guess the index would work for searching `tracked` = false.

----------------------------------------
However, I'm not sure if adding that index is the best approach here.

The query you mentioned runs on every update:
`has_untracked_timeslices = @course.article_course_timeslices.exists?(tracked: false)`

If there are untracked timeslices, then we also run this second query:
```
@article_course_timeslices_to_retrack = ArticleCourseTimeslice
                                            .for_course_and_article(@course, article_ids)
                                            .where(tracked: false)
```

Do you know if that second query is also slow? Maybe it's not slow because it uses the `article_course_timeslice_by_article_course_start_and_end` index?

If the second query is not slow, then maybe we can remove the check, because I understand that the first query is only done to avoid the retracking step if there are no untracked articles (but maybe that query itself is slower than the retracking step?)
```
    # Most of the time there are no untracked timeslices, so we can skip the retrack step
    has_untracked_timeslices = @course.article_course_timeslices.exists?(tracked: false)
    return unless has_untracked_timeslices
```

And do something like this:

```
  def retrack(article_ids)
    return if article_ids.empty?

    @article_course_timeslices_to_retrack = ArticleCourseTimeslice
                                            .for_course_and_article(@course, article_ids)
                                            .where(tracked: false)
    # Only reprocess those non-empty timeslices
    non_empty = @article_course_timeslices_to_retrack.where.not(user_ids: nil)

    # We can skip the retrack step if there are no non-empty articles to retrack
    return if non_empty.empty?

    # Mark course wiki timeslices that needs to be re-processed
    @timeslice_cleaner.reset_timeslices_that_need_update_from_article_timeslices(non_empty)
    retrack_timeslices
  end
```

WDYT?

Abishekcs:
@gabina Thank you  for the review.

> If the second query is not slow, then maybe we can remove the check, because I understand that the first query is only done to avoid the retracking step if there are no untracked articles (but maybe that query itself is slower than the retracking step?)

I don't think we can remove the check because the second query is much slower than the first one, as shown in the screenshot below. I'm looking for another solution will update on it if I find something better

Screenshot from 2025-07-14 12-42-56

Abishekcs:
> (examples: [this stackoverflow post](https://stackoverflow.com/questions/61027237/does-it-make-sense-to-make-a-composite-index-of-2-columns-when-one-of-the-column), this [MariaDB video](https://www.youtube.com/watch?v=x-5hDEvQhuM))

Thank you — these two resources were very helpful. In case you're interested, I found another explanation of the same concept: https://planetscale.com/learn/courses/mysql-for-developers/indexes/index-selectivity

gabina:
> I don't think we can remove the check because the second query is much slower than the first one, as shown in the screenshot below.

Interesting, I guess the `article_course_timeslice_by_article_course_start_and_end` index is not used then.

Abishekcs:
> Interesting, I guess the `article_course_timeslice_by_article_course_start_and_end` index is not used then

Actually, it uses the `article_course_timeslice_by_article_course_start_and_end` index

Screenshot from 2025-07-14 18-33-56


It uses `Extra: Using index condition; Using where`. This is already very optimized. We can avoid the `Using where` 
by adding composite index on `(course_id, tracked, article_id);`  but i don't think that will be very helpful. It could have been helpful if only `(course_id, tracked, article_id);` i.e the first two column of the composite index, was used by the check part of the code 

```
has_untracked_timeslices = @course.article_course_timeslices.exists?(tracked: false)

```

Abishekcs:
> Actually, it uses the `article_course_timeslice_by_article_course_start_and_end` index

I Found one more interesting thing  for this code 

Screenshot from 2025-07-14 19-36-50


If When the `IN (...)`  list is very long i.e  (hundreds or thousands of values) then it won't use the index as shown below

Screenshot from 2025-07-14 19-39-27

Abishekcs:
Closing this one i have few more ideas i want to try but probably in a new PR

Issue #6369, opened 2025-06-17
ArticleViewer should follow redirects if assigned article is a redirect


ragesoss:
In some cases, an assigned article is currently a redirect to a different title. (In some cases, I think the Dashboard's update system may detect cases where a redirect is assigned, and update the assigned title to the redirect target, but that isn't always the case.)

Currently, in a case like this, the ArticleViewer just loads the redirect itself. Instead, it should recognize the redirect and load the redirect target page instead.

"Selfies" is an example redirect page that you can assign for testing this behavior: https://en.wikipedia.org/w/index.php?title=Selfies&redirect=no

Imvedansh:
@ragesoss I think its the issue when it uses the parameter redirect=no, apart from that i tested cases are likely to work, though when uses this parameter it return the unexpected responses. WDYT?

Nooraw:
Hello @ragesoss , I am an Outreachy contributor and would like to take on this task if its still available. Thanks!

Abishekcs:
@Nooraw feel free to take on this issue

PR #6406, opened 2025-07-12
fix to redirect_parameter in ArticleViewer

Imvedansh:

## What this PR does
#6369 -Fixes
Problem: URLs with redirect=no were showing redirect pages instead of actual content
Solution:
1. ArticleViewerAPI.js - MAIN FIX
What Changed: Modified the fetchParsedArticle method
Why: This is where the actual Wikipedia API call happens. Added logic to:

Detect if Wikipedia returns a redirect response
Automatically make a second API call to get the target article
Return the actual article content instead of redirect content

Specific Change: Added redirect detection and handling in the .then() chain

2. URLBuilder.js - SUPPORTING CHANGE
What Changed: Added &redirects=true parameter to the Wikipedia API URL
Why: This tells Wikipedia's API to include redirect information in the response, so we can detect when an article is a redirect
Specific Change: In parsedArticleURL() method, changed the query string to include redirect info

##Result:

"Selfies" URLs now automatically redirect to "Selfie" content
Clean URLs stored in database
No more redirect pages in ArticleViewer
Test passes 

### Before

https://github.com/user-attachments/assets/10d73dde-4e7f-4f99-b81d-cefb7d7ce83b

### After

https://github.com/user-attachments/assets/0ba889f5-cb97-4627-b6f1-01c5da6767f7

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/utils/ArticleViewerAPI.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/utils/ArticleViewerAPI.js
@@ -76,6 +76,31 @@ export class ArticleViewerAPI {
     }).then(response => this.__handleFetchResponse(response))
       .then((response) => {
         if (response.error) throw new Error(this.__setException({ status: 404 }));
+        // Check if this is a redirect page
+        if (response.parse.redirects && response.parse.redirects.length > 0) {
+          const redirectTarget = response.parse.redirects[response.parse.redirects.length - 1].to;
+          // Make a new request for the redirect target
+          const redirectUrl = this.builder.parsedArticleURL().replace(
+            `page=${encodeURIComponent(this.builder.article.title)}`,
+            `page=${encodeURIComponent(redirectTarget)}`
+          );
+
+          return fetch(`${redirectUrl}&origin=*`, {
+            headers: {
+              'Content-Type': 'application/javascript'
+            }
+          }).then(redirectRes => this.__handleFetchResponse(redirectRes))
+            .then((redirectResponse) => {
+              if (redirectResponse.error) throw new Error(this.__setException({ status: 404 }));
+              return {
+                articlePageId: redirectResponse.parse.pageid,
+                fetched: true,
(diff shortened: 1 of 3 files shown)


ragesoss:
Thanks! Can you include a before/after screenshot or video demonstrating the changed behavior?

Imvedansh:
Sure let me attach video of before vs after, that would me more clear

Imvedansh:
Tested with other instances as well to ensure everything works smoothly

Imvedansh:
https://github.com/user-attachments/assets/a93e2695-e770-43aa-a687-2d9f63b9d185

As well , refer above , shared Before Vs After

ragesoss:
Thanks! It looks like this breaks one of the Javascript tests.

ragesoss:
This issue is specifically about the ArticleViewer component. In the 'Selfies' example, you can go to the Editors tab, then click your username and click the page icon to see this:

article viewer redirect

(The ArticleViewer is also used on the Articles tab for the table of edited articles, but it works the same way in both situations.)

ragesoss:
Thanks @Imvedansh! Can you update the PR description with a video demo for ArticleViewer?

app/assets/javascripts/components/common/ArticleViewer/utils/ArticleViewerAPI.js
```diff
@@ -76,6 +76,31 @@ export class ArticleViewerAPI {
     }).then(response => this.__handleFetchResponse(response))
       .then((response) => {
         if (response.error) throw new Error(this.__setException({ status: 404 }));
+        // Check if this is a redirect page
```
ragesoss, line 79: Does any of this make sense to extract to a separate method, to keep `fetchParsedArticle` from becoming harder to read and understand?

Imvedansh: Its not completed yet,
Im clossing it for now ,Will open it again when done

PR #6408, opened 2025-07-14
Auto focus empty string validation

Imvedansh:
## What this PR does
-Fixes->6407
-This PR adds autofocus to the ArticleFinderSearchBar input field when the component mounts. It improves the user experience by allowing users to start typing immediately without needing to click the input box.

-This addresses the issue where the search input did not receive focus automatically . A friction point for users, especially those who rely on keyboard navigation.

## Screenshots
Before:

https://github.com/user-attachments/assets/d1a89164-0a22-4243-99cd-a4736fbc871b

After:

https://github.com/user-attachments/assets/f7a6e1b1-f469-4228-b75f-7e72c3f7946c

##How to fix it
1)Added a ref (searchInputRef) to directly access the search input element.
2)Used useEffect to auto-focus the input field when the component mounts.
3)Included a slight setTimeout to ensure smooth DOM rendering before focusing.
4)Attached the ref to the <input> tag for focus control.
5)If input already has a value, the search is triggered automatically on mount.
6)Introduced a basic error state when user tries searching with empty input.
7)Highlighted the input and refocused it to guide the user.

### DIFF
--- a/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
+++ b/app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
@@ -1,28 +1,54 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useState, useEffect, useRef } from 'react';
 import { debounce } from 'lodash';
 import { fetchArticleAutocompleteResults } from '../../utils/article_finder_utils';
 
 // Controls Search bar and autocomplete functionality
-
 function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
   const [suggestions, setSuggestions] = useState([]);
   const [isAutocompleteLoading, setAutocompleteLoading] = useState(false);
+  const [showEmptySearchError, setShowEmptySearchError] = useState(false);
+  const searchInputRef = useRef(null);
+
   let searchClass = 'article-finder-search-bar';
 
   if (suggestions.length > 0) {
     searchClass += ' autocomplete-on';
   }
 
-  const _getSuggestionsApi = useCallback(debounce(async (q) => {
-    setAutocompleteLoading(true);
-    const results = await fetchArticleAutocompleteResults(q, wiki).catch(() => []);
-    setAutocompleteLoading(false);
-    setSuggestions(results);
-  }, 500), []);
+  // Auto-focus on the search input when component mounts
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      if (searchInputRef.current) {
+        searchInputRef.current.focus();
+
+        // Trigger search if input already has a value
+        if (value.trim() !== '') {
+          onSearch(value.trim());
+        }
+      }
+    }, 100);
+    return () => clearTimeout(timer);
+  }, []);
+
+  // Hide error message when user starts typing
+  useEffect(() => {
+    if (value.trim() !== '' && showEmptySearchError) {
+      setShowEmptySearchError(false);
+    }
+  }, [value, showEmptySearchError]);
+
+  const _getSuggestionsApi = useCallback(
+    debounce(async (q) => {
+      setAutocompleteLoading(true);
+      const results = await fetchArticleAutocompleteResults(q, wiki).catch(() => []);
+      setAutocompleteLoading(false);
+      setSuggestions(results);
+    }, 500),
+    []
+  );
 
   const inputChangeHandler = (e) => {
     onChange(e.target.value);
-    // autocomplete disabled while disabled (during loading etc.)
     if (!disabled) {
       if (e.target.value === '') {
         setSuggestions([]);
@@ -33,13 +59,20 @@ function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
   };
 
   const searchHandler = () => {
-    if (disabled || value.trim() === '') return;
+    if (disabled) return;
+    if (value.trim() === '') {
+      setShowEmptySearchError(true);
+      if (searchInputRef.current) {
+        searchInputRef.current.focus();
+      }
+      return;
+    }
+    setShowEmptySearchError(false);
     setSuggestions([]);
     onSearch(value);
   };
 
   const onKeyDownHandler = (e) => {
-    // Search on Enter
     if (e.key === 'Enter') {
       searchHandler();
     }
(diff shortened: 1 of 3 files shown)


Imvedansh:
@ragesoss Looks good?

ragesoss:
thanks @Imvedansh! we'll try to get it reviewed this week.

Imvedansh:
This build failed is not related to the PR in my knowledge

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
```diff
@@ -1,28 +1,54 @@
...
-    setAutocompleteLoading(false);
-    setSuggestions(results);
-  }, 500), []);
+  // Auto-focus on the search input when component mounts
+  useEffect(() => {
+    const timer = setTimeout(() => {
+      if (searchInputRef.current) {
+        searchInputRef.current.focus();
+
+        // Trigger search if input already has a value
```
Abishekcs: Are you sure the trigger search is working as intended? From what I understand, it should initiate a search on component mounts, but only if the input field already contains a value.

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
```diff
@@ -1,28 +1,54 @@
...
+        if (value.trim() !== '') {
+          onSearch(value.trim());
+        }
+      }
+    }, 100);
+    return () => clearTimeout(timer);
+  }, []);
+
+  // Hide error message when user starts typing
```
Abishekcs: What kind of error message is this hiding can you show screenshot or recording ?

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
```diff
@@ -33,23 +59,30 @@ function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
   };
 
   const searchHandler = () => {
-    if (disabled || value.trim() === '') return;
+    if (disabled) return;
+    if (value.trim() === '') {
+      setShowEmptySearchError(true);
+      if (searchInputRef.current) {
+        searchInputRef.current.focus();
```
Abishekcs: hmm I am not sure but when I tested this with an empty search, the input element still appeared focused even with this line of code commented out `if (searchInputRef.current) { ...`. So, I’m wondering — wouldn’t the input remain focused since it was already set by the useEffect on mount? Or is this line handling some specific edge case I might be missing?

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
Imvedansh: Thanks for bringing my attention to it,
actually trigger search here is unreasonable ,initially thought of adding it as a feature but it Unnecessary network requests on every component mount as well users  didn't explicitly request a search it happen automatically, so kind of not needed, I forgot to remove it  before pushing

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
Imvedansh: So i was thinking of adding the warning in placeholder or maybe adding alert as if user hit empty search (showEmptySearchError label) but that's not related to this PR and should be done separately, so i ll draft another PR with proper direction for that

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
Imvedansh: This is also about error handling for empty string search ,So when a user clicks the search button with empty input, the focus might shift away from the input field. Re-focusing it after showing the error allows users to immediately start typing without having to click back on the input.
I ll remove both these code residue and Draft separate PR for it.

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
```diff
@@ -61,21 +75,29 @@ function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
         onChange={inputChangeHandler}
         value={value}
         onKeyDown={onKeyDownHandler}
+        ref={searchInputRef}
       />
 
-      {
-        isAutocompleteLoading && <div className="loader"><div className="loading__spinner"/></div>
-      }
+      {isAutocompleteLoading && (
```
Abishekcs: About this section: since the changes mainly break the JSX into multiple lines without adding new logic, I’m wondering if this refactor is strictly necessary. The original version seemed concise and readable enough as it was. Personally, I find the single-line format clearer here.What do you think?

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
```diff
@@ -61,21 +75,29 @@ function ArticleFinderSearchBar({ value, onChange, onSearch, disabled, wiki }) {
...
         {I18n.t('article_finder.search')}
       </button>
+
       <div className="autocomplete">
-        {
-          suggestions.map((sug) => {
-            return <div key={sug} className="autocomplete-item" onClick={() => autoCompleteClickHandler(sug)}> {sug} </div>;
-          })
-        }
+        {suggestions.map(sug => (
```
Abishekcs: Same for this.

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
Abishekcs: Okay great. Whenever you open a new PR for the changes you mentioned. Please make sure to add a Screenrecording for the new features you added.Thank you!

Imvedansh: Absolutely,Thank you much

app/assets/javascripts/components/article_finder/article_finder_search_bar.jsx
Imvedansh: yes , you are right, will revert it by the final commit,Thanks!!

Imvedansh:
Build failure is not related to this PR in my knowledge,
am i missing something?

Abishekcs:
> Build failure is not related to this PR in my knowledge,
> am i missing something?

No its not related to your PR

Abishekcs:
@ragesoss This PR looks good to me! I have tested it out.

@Imvedansh Just one last thing before this PR gets merged i think  you should clean up the git history for this PR? Since it’s a small change, I think having just a single commit would be ideal.

PR #6410, opened 2025-07-15
Optimize article_ids query by batching titles to prevent large IN clause slowdowns

Abishekcs:
## Production Log (First Occurence is in the month of June)

Screenshot from 2025-07-15 19-59-03

## What this PR does

Improves the performance of the `article_ids` method in the `Category` model by introducing batching on the `article_titles` array when querying `Article` records.

Previously, the query used a single large `IN` clause:
```ruby
@article_ids ||= Article.where(namespace: 0, wiki_id:, title: article_titles).pluck(:id)
````

When `article_titles` contained thousands of entries (e.g., 1000+), this resulted in:

* Very large SQL `IN (...)` clauses
* Query planner inefficiencies
* Memory-intensive materialization in MySQL
* Increased `Lock_time` and `Query_time` in production logs

We now batch the `article_titles` into slices of 500:

```ruby
@article_ids ||= article_titles.each_slice(500).flat_map do |titles_batch|
  Article.where(namespace: 0, wiki_id: wiki_id, title: titles_batch).pluck(:id)
end
```

*  **Smaller IN clauses** reduce MySQL's parsing and planning overhead
*  **Index usage is optimized** (`range` instead of `ALL`)
*  **Avoids materializing large subqueries**, saving memory
*  **Improves overall query time and reduces locking**
*  **EXPLAIN plans confirm efficient access paths** with `Using index` and `range` types

## Screenshots

**Before:** `1000 at once`

```
type: ALL
Subquery materialization
```

Screenshot from 2025-07-15 21-37-08

**After:** `500 in batches`

```
type: range
key: index_articles_on_namespace_and_wiki_id_and_title
Extra: Using where; Using index
```

Screenshot from 2025-07-15 21-36-37

## **Benchmark Timing Breakdown (10 Runs per Method)**

# Single

| Run | Method          | User (s)   | System (s) | Total (s)  | Real (s)    |
| --- | --------------- | ---------- | ---------- | ---------- | ----------- |
| 1   | Single Query    | 2.3580     | 0.8188     | 3.1768     | 23.4926     |
| 2   | Single Query    | 2.2588     | 0.7845     | 3.0433     | 22.9931     |
| 3   | Single Query    | 2.1097     | 0.7519     | 2.8616     | 23.3423     |
| 4   | Single Query    | 2.0646     | 0.6570     | 2.7216     | 23.4222     |
| 5   | Single Query    | 2.1569     | 0.6499     | 2.8068     | 23.7333     |
| 6   | Single Query    | 2.0162     | 0.5740     | 2.5902     | 23.2345     |
| 7   | Single Query    | 2.1789     | 0.6311     | 2.8100     | 23.5475     |
| 8   | Single Query    | 2.2262     | 0.6690     | 2.8952     | 23.5060     |
| 9   | Single Query    | 2.4627     | 0.8091     | 3.2719     | 23.5800     |
| 10  | Single Query    | 2.2506     | 0.7278     | 2.9785     | 25.1082     |
|     | **Avg (Set 1)** | **2.2087** | **0.7072** | **2.9159** | **23.6960** |

---

# Batch

| Run | Method          | User (s)   | System (s) | Total (s)  | Real (s)    |
| --- | --------------- | ---------- | ---------- | ---------- | ----------- |
| 1   | Batched Query   | 3.8946     | 0.2639     | 4.1585     | 11.0029     |
| 2   | Batched Query   | 4.2386     | 0.3201     | 4.5587     | 10.9594     |
| 3   | Batched Query   | 4.0917     | 0.2321     | 4.3238     | 10.7075     |
| 4   | Batched Query   | 4.1782     | 0.3019     | 4.4801     | 10.9013     |
| 5   | Batched Query   | 4.0295     | 0.2899     | 4.3194     | 10.5783     |
| 6   | Batched Query   | 4.1714     | 0.3165     | 4.4879     | 11.0138     |
| 7   | Batched Query   | 3.9317     | 0.2578     | 4.1895     | 9.3219      |
| 8   | Batched Query   | 4.1886     | 0.3093     | 4.4978     | 11.7283     |
| 9   | Batched Query   | 4.0898     | 0.2790     | 4.3688     | 10.9055     |
| 10  | Batched Query   | 4.1529     | 0.3004     | 4.4534     | 10.9656     |
|     | **Avg (Set 2)** | **4.1967** | **0.2571** | **4.4538** | **10.8085** |

---

## **Benchmark Comparison: Single Query vs. Batched Query (2000-title slices)**

**Test Conditions:**

* Dataset: 782,495 unique titles
* Queries: 10 runs per method
* Metric: Time measured using Ruby’s `Benchmark.measure` (user, system, total, real)

---

### **Average Time per Query (10 runs)**

| Metric           | Single Query      | Batched Query (2000-title) | Less time required by          |
| ---------------- | ----------------- | -------------------------- | --------------- |
| **User**      | 2.21 seconds      | 4.20 seconds               | Single Query    |
| **System**    | 0.71 seconds      | 0.26 seconds               | **Batched**     |
| **Total CPU** | 2.92 seconds      | 4.45 seconds               | Single Query    |
| **Real Time** | **23.70 seconds** | **10.81 seconds**          | **Batched**  |

## [BenchMark Memory](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6410#issuecomment-3084221686)

### DIFF
--- a/app/models/wiki_content/category.rb
+++ b/app/models/wiki_content/category.rb
@@ -78,7 +78,14 @@ class Category < ApplicationRecord
   end
 
   def article_ids
-    @article_ids ||= Article.where(namespace: 0, wiki_id:, title: article_titles).pluck(:id)
+    return @article_ids if @article_ids.present?
+
+    @article_ids = []
+    article_titles.each_slice(500) do |titles_batch|
+      @article_ids.concat(Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id))
+    end
+
+    @article_ids.uniq! || @article_ids
   end
 
   def name_with_prefix
--- a/app/models/wiki_content/category.rb
+++ b/app/models/wiki_content/category.rb
@@ -81,7 +81,7 @@ class Category < ApplicationRecord
     return @article_ids if @article_ids.present?
 
     @article_ids = []
-    article_titles.each_slice(2000) do |titles_batch|
+    article_titles.each_slice(5000) do |titles_batch|
       @article_ids.concat(Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id))
     end
 
--- a/app/models/wiki_content/category.rb
+++ b/app/models/wiki_content/category.rb
@@ -81,11 +81,11 @@ class Category < ApplicationRecord
     return @article_ids if @article_ids.present?
 
     @article_ids = []
-    article_titles.each_slice(500) do |titles_batch|
+    article_titles.each_slice(2000) do |titles_batch|
       @article_ids.concat(Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id))
     end
 
-    @article_ids.uniq! || @article_ids
+    @article_ids
   end
 
   def name_with_prefix


gabina:
I like this idea. Are you estimating previous query execution time vs current query execution time to see the improvement? What about memory?

Abishekcs:
> Are you estimating previous query execution time vs current query execution time to see the improvement?

Yes these is one of the ways i am trying to measure the improvement.

Screenshot from 2025-07-16 15-03-30


**this is the benchmark code:** 

```ruby

  def article_ids # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
    require 'benchmark'

    # First approach - Single query
    articles_titles = Article.all.pluck(:title)

    single_query_time = Benchmark.measure do
      @article_ids = Article.where(namespace: 0, wiki_id:, title: articles_titles).pluck(:id)
    end

    # Second approach - Batched queries
    @article_ids = nil # Reset for fair comparison

    batched_query_time = Benchmark.measure do
      @article_ids = articles_titles.each_slice(500).flat_map do |titles_batch|
        Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id)
      end
    end

    puts 'Single Query Results:'
    puts single_query_time
    puts "Article IDs found: #{@article_ids.count}"
    puts '---'

    puts 'Batched Query Results:'
    puts batched_query_time
    puts "Article IDs found: #{@article_ids.count}"
    puts '---'

    # Compare the results
    puts 'Performance Comparison:'
    puts "Single query time: #{single_query_time.real.round(4)}s"
    puts "Batched query time: #{batched_query_time.real.round(4)}s"
    puts "Improvement: #{((single_query_time.real - batched_query_time.real) / single_query_time.real * 100).round(2)}%"
  end


```

Abishekcs:
> Are you estimating previous query execution time vs current query execution time to see the improvement?

Yes these is one of the ways i am trying to measure the improvement. (all the below three screenshot where tested again dataset of - `155841 articles`). I think larger the dataset the more the perfomance improvement is.


Comparsion 3
Comparsion 2
Comparsion 1


`This is the benchmark code`: 

```ruby

def article_ids # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
    require 'benchmark'

    # Capture titles once and remove duplicates to ensure consistency
    articles_titles = Article.all.pluck(:title).uniq.freeze
    puts "Testing with #{articles_titles.count} unique titles"

    # First approach - Single query
    single_query_time = Benchmark.measure do
      @article_ids1 = Article.where(namespace: 0, wiki_id:, title: articles_titles).pluck(:id)
    end

    # Second approach - Batched queries
    batched_query_time = Benchmark.measure do
      @article_ids2 = articles_titles.each_slice(500).flat_map do |titles_batch|
        Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id)
      end.uniq # Extra safety to remove any potential duplicates
    end

    # Verify results are identical before showing performance
    results_identical = @article_ids1.sort == @article_ids2.sort

    puts 'Single Query Results:'
    puts single_query_time
    puts "Article IDs found: #{@article_ids1.count}"
    puts '---'

    puts 'Batched Query Results:'
    puts batched_query_time
    puts "Article IDs found: #{@article_ids2.count}"
    puts '---'

    # Results validation
    if results_identical
      puts 'Results are identical - performance comparison is valid'
    else
      puts "Results differ - there's a logical error!"
      puts "Single query unique IDs: #{@article_ids1.uniq.count}"
      puts "Batched query unique IDs: #{@article_ids2.uniq.count}"
      puts "Difference: #{(@article_ids1 - @article_ids2).count} IDs missing from batched"
      return # Don't show performance if results differ
    end

    # Performance comparison (only if results match)
    puts 'Performance Comparison:'
    puts "Single query time: #{single_query_time.real.round(4)}s"
    puts "Batched query time: #{batched_query_time.real.round(4)}s"

    if batched_query_time.real < single_query_time.real
      improvement = ((single_query_time.real - batched_query_time.real) / single_query_time.real * 100).round(2)
      puts "Improvement: #{improvement}% (batched is faster)"
    else
      regression = ((batched_query_time.real - single_query_time.real) / single_query_time.real * 100).round(2)
      puts "Regression: #{regression}% (single query is faster)"
    end
  end

```

> What about memory?


**Original:** Database must materialize entire item subquery in memory

**Batched:** Maximum 500 items in memory at any time, with garbage collection between batches

Abishekcs:
I am trying to write a benchamark to test memory consumption too.

Abishekcs:
### **Update:**

I was able to reduce the memory consumption too with recent changes to the code:

``` ruby

  def article_ids
    return @article_ids if @article_ids.present?

    @article_ids = []
    article_titles.each_slice(500) do |titles_batch|
      @article_ids.concat(Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id))
    end

    @article_ids.uniq! || @article_ids
  end

```
**This code save memory by:** 

- concat appends directly to a single array (@article_ids).

- Only one main array is used throughout — no need for flattening or reallocation.

- concat is in-place and uses Array method under the hood, which is efficient.

- .uniq! mutates the array once `(i,e in the original variable it self)`, instead of copying.


### Benchmark Screenshot (Time + Memory)


NEW 1
New 3
New 2
Screenshot from 2025-07-17 00-16-08


### Benchmark (Code)


```ruby

 def article_ids # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
    require 'benchmark'

    # Simulated large input i.e take all the data available in my local db its around 155481
    articles_titles = Article.all.pluck(:title)

    puts "Benchmarking with #{articles_titles.size} titles..."

    results = {}

    ### Original (Single Query) ###
    GC.start(full_mark: true, immediate_sweep: true)
    ObjectSpace.trace_object_allocations_start

    memory_before = ObjectSpace.memsize_of_all
    time_single = Benchmark.realtime do
      @article_ids1 = Article.where(namespace: 0, wiki_id:, title: articles_titles).pluck(:id)
    end
    memory_after = ObjectSpace.memsize_of_all

    results[:single] = {
      time: time_single.round(3),
      memory_mb: ((memory_after - memory_before) / 1024.0 / 1024.0).round(2),
      count: @article_ids1.size
    }

    ### Batched Query ###
    GC.start(full_mark: true, immediate_sweep: true)
    ObjectSpace.trace_object_allocations_start

    memory_before = ObjectSpace.memsize_of_all
    time_batched = Benchmark.realtime do
      @article_ids2 = []
        articles_titles.each_slice(500) do |batch|
          @article_ids2.concat(Article.where(namespace: 0, wiki_id:, title: batch).pluck(:id))
        end
      @article_ids2.uniq!
    end
    memory_after = ObjectSpace.memsize_of_all

    results[:batched] = {
      time: time_batched.round(3),
      memory_mb: ((memory_after - memory_before) / 1024.0 / 1024.0).round(2),
      count: @article_ids2.size
    }

    ### Output ###
    puts "\n=== Results ==="
    results.each do |key, data|
      puts "#{key.capitalize} Query:"
      puts "  Time:        #{data[:time]}s"
      puts "  Memory used: #{data[:memory_mb]} MB"
      puts "  IDs fetched: #{data[:count]}"
    end

    if results[:batched][:time] < results[:single][:time]
      gain = ((results[:single][:time] - results[:batched][:time]) / results[:single][:time] * 100).round(2)
        puts "Batching is faster by #{gain}%"
    else
      loss = ((results[:batched][:time] - results[:single][:time]) / results[:batched][:time] * 100).round(2)
        puts "Batching is slower by #{loss}%"
    end
  end


```

gabina:
This sounds exciting! I don't have a lot of experience running benchmarks, but I have some concerns about whether the comparisons are fair.

As regards the first benchmark you ran to measure time, notice that `Benchmark.measure` reports four times: 1) user CPU time, 2) system CPU time, 3) the sum of the user and system CPU times, and 4) the elapsed real time.

CPU time and elapsed real time (that includes waiting time) are greater in single query results than in batched query results. However, user CPU time is grater in batched query results than in  single query results.
image

Although you ran the queries three times and results seem similar, I would run the query at least 10 times for each one and then show averages times for each different time.

In addition, I would run the single query in a different script from the batched query. I think there is a possibility that something is left in cache after the first query, and then the second one does not run under the same conditions. Do you get the same time values if you run first the batched query and then the single one?

I'm not super familiar with `GC` and `ObjectSpace` but I have the same concern on the memory benchmark, are you sure that something is not cached when running the first query that does affect the performance when running the second one? I would try to separate the benchmarks for each query into different scripts.

Abishekcs:
> Do you get the same time values if you run first the batched query and then the single one?

Yes.

> are you sure that something is not cached when running the first query that does affect the performance when running the second one?

hmm i am pretty sure the order of execution doesn't affect each other. But i will run them separately to be sure.

The results are given below `It seems running a batch of 2000 seems to give better result than batch of 500`. Also, i have tried to increase the datset of `Articles` from `150,000` to `782,495`


## **Benchmark Timing Breakdown (10 Runs per Method)**

# Single

| Run | Method          | User (s)   | System (s) | Total (s)  | Real (s)    |
| --- | --------------- | ---------- | ---------- | ---------- | ----------- |
| 1   | Single Query    | 2.3580     | 0.8188     | 3.1768     | 23.4926     |
| 2   | Single Query    | 2.2588     | 0.7845     | 3.0433     | 22.9931     |
| 3   | Single Query    | 2.1097     | 0.7519     | 2.8616     | 23.3423     |
| 4   | Single Query    | 2.0646     | 0.6570     | 2.7216     | 23.4222     |
| 5   | Single Query    | 2.1569     | 0.6499     | 2.8068     | 23.7333     |
| 6   | Single Query    | 2.0162     | 0.5740     | 2.5902     | 23.2345     |
| 7   | Single Query    | 2.1789     | 0.6311     | 2.8100     | 23.5475     |
| 8   | Single Query    | 2.2262     | 0.6690     | 2.8952     | 23.5060     |
| 9   | Single Query    | 2.4627     | 0.8091     | 3.2719     | 23.5800     |
| 10  | Single Query    | 2.2506     | 0.7278     | 2.9785     | 25.1082     |
|     | **Avg (Set 1)** | **2.2087** | **0.7072** | **2.9159** | **23.6960** |

---

# Batch

| Run | Method          | User (s)   | System (s) | Total (s)  | Real (s)    |
| --- | --------------- | ---------- | ---------- | ---------- | ----------- |
| 1   | Batched Query   | 3.8946     | 0.2639     | 4.1585     | 11.0029     |
| 2   | Batched Query   | 4.2386     | 0.3201     | 4.5587     | 10.9594     |
| 3   | Batched Query   | 4.0917     | 0.2321     | 4.3238     | 10.7075     |
| 4   | Batched Query   | 4.1782     | 0.3019     | 4.4801     | 10.9013     |
| 5   | Batched Query   | 4.0295     | 0.2899     | 4.3194     | 10.5783     |
| 6   | Batched Query   | 4.1714     | 0.3165     | 4.4879     | 11.0138     |
| 7   | Batched Query   | 3.9317     | 0.2578     | 4.1895     | 9.3219      |
| 8   | Batched Query   | 4.1886     | 0.3093     | 4.4978     | 11.7283     |
| 9   | Batched Query   | 4.0898     | 0.2790     | 4.3688     | 10.9055     |
| 10  | Batched Query   | 4.1529     | 0.3004     | 4.4534     | 10.9656     |
|     | **Avg (Set 2)** | **4.1967** | **0.2571** | **4.4538** | **10.8085** |


---

## **Benchmark Comparison: Single Query vs. Batched Query (2000-title slices)**

**Test Conditions:**

* Dataset: 782,495 unique titles
* Queries: 10 runs per method
* Metric: Time measured using Ruby’s `Benchmark.measure` (user, system, total, real)

---

### **Average Time per Query (10 runs)**

| Metric           | Single Query      | Batched Query (2000-title) | Less time required by          |
| ---------------- | ----------------- | -------------------------- | --------------- |
| **User**      | 2.21 seconds      | 4.20 seconds               | Single Query    |
| **System**    | 0.71 seconds      | 0.26 seconds               | **Batched**     |
| **Total CPU** | 2.92 seconds      | 4.45 seconds               | Single Query    |
| **Real Time** | **23.70 seconds** | **10.81 seconds**          | **Batched**  |

Abishekcs:
###  Benchmark: Memory Usage

For memory benchmarking, I ran each method only once, as memory consumption remained consistent across repeated runs. Therefore, averaging was not necessary.

**Observation:**  
Batched queries appear to consume **more memory** than single queries, as shown below:


## Single Query:

Single Memory Consumption


``` ruby

  def article_ids # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
    require 'benchmark'
    require 'memory_profiler'

    # Simulated large input i.e take all the data available in my local db
    articles_titles = Article.all.pluck(:title).first(790000).uniq.freeze

    puts "Benchmarking with #{articles_titles.size} titles..."

    results = {}

    ### Original (Single Query) ###
    GC.start(full_mark: true, immediate_sweep: true)

    time_single = Benchmark.realtime do
      report_single = MemoryProfiler.report do
        @article_ids1 = Article.where(namespace: 0, wiki_id:, title: articles_titles).pluck(:id)
      end
      results[:single_report] = report_single
    end

    results[:single] = {
      time: time_single.round(3),
      count: @article_ids1.size
    }

    puts "\n=== Memory Report: Single Query ==="
    results[:single_report].pretty_print

```


## Batch Query (Batch of 2000):


Batch Memory Consumption


``` ruby

  def article_ids # rubocop:disable Metrics/MethodLength,Metrics/AbcSize
    require 'benchmark'
    require 'memory_profiler'

    # Simulated large input i.e take all the data available in my local db
    articles_titles = Article.all.pluck(:title).first(790000).uniq.freeze

    puts "Benchmarking with #{articles_titles.size} titles..."

    results = {}

    ## Batched Query ###
    GC.start(full_mark: true, immediate_sweep: true)

    time_batched = Benchmark.realtime do
      report_batched = MemoryProfiler.report do
        @article_ids = []
        articles_titles.each_slice(2000) do |titles_batch|
          @article_ids.concat(Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id))
        end

        @article_ids.uniq!
      end
      results[:batched_report] = report_batched
    end

    results[:batched] = {
      time: time_batched.round(3),
      count: @article_ids.size
    }

    puts "\n=== Memory Report: Batched Query ==="
    results[:batched_report].pretty_print
  end


```

gabina:
hmm OK, so I understand that these changes improve time in a considerable way (real time gets reduced by less than half). However, memory allocation is not reduced. Is that so?

I have some doubts. Are you using `uniq!` because `article_titles` may have duplicate titles? Or just as a way of memory optimization? Asking because having duplicate titles in `article_titles` is a bad thing that we should address. In fact, I just noticed that `scoped_article_titles` might contain duplicate articles and that's something we definitely want to fix considering `scoped_article_titles` is one of the most memory-intensive resources (see new issue #6415).

I understand you did the things mentioned below to try to reduce memory. if those changes do not mean a significant reduction in memory, I would avoid making them as they make the implementation more complex.

> I was able to reduce the memory consumption too with recent changes to the code:
> 
>   def article_ids
>     return @article_ids if @article_ids.present?
> 
>     @article_ids = []
>     article_titles.each_slice(500) do |titles_batch|
>       @article_ids.concat(Article.where(namespace: 0, wiki_id:, title: titles_batch).pluck(:id))
>     end
> 
>     @article_ids.uniq! || @article_ids
>   end
> This code save memory by:
> 
> concat appends directly to a single array (@article_ids).
> 
> Only one main array is used throughout — no need for flattening or reallocation.
> 
> concat is in-place and uses Array method under the hood, which is efficient.
> 
> .uniq! mutates the array once (i,e in the original variable it self), instead of copying.

Abishekcs:
> However, memory allocation is not reduced. Is that so?

Yes.

> I have some doubts. Are you using `uniq!` because `article_titles` may have duplicate titles? Or just as a way of memory optimization?

Actually, I’m using `uniq!` to avoid duplicate `article_id`s that can result from batching and to save little bit of memory too. Let me explain with an example:

Suppose we have an array of article titles:

```ruby
article_titles = ["Apple", "Orange", "Football", "GitHub", "Dota 2", "Apple"]
```

Notice that `"Apple"` appears twice — at index 0 and 5. If we run this array in **batches of 3**, the first batch will include `"Apple"` once, and the second batch will include it again.

As a result, `"Apple"` will be queried twice across different batches, and its `article_id` will appear multiple times in the result.

So, to ensure we don’t store duplicate IDs when batching, i was uing  `.uniq!` at the end to de-duplicate them.

In contrast, with a **single query**, the entire array is passed in one `WHERE IN (article_titles)` clause, and i think the database handles duplicates internally — returning each matching article ID only once.

Abishekcs:
However, the original source of the `article_titles` array comes from the code below. While testing, I needed a large dataset, so I was fetching article titles directly from the database using .all.

If this existing code (shown in the below screenshot) ensures that the `article_titles` array only contains **unique titles**, then technically, I can remove the call to `.uniq!` later in the batching logic — since there would be no duplicates to worry about.

However, if there's any chance the titles fetched are not unique, it's still safer to keep `.uniq!` to prevent duplicate `article_id`s from being included.

Screenshot from 2025-07-17 22-58-13

gabina:
Gotcha. Thank you very much for your detailed response!

> If this existing code (shown in the below screenshot) ensures that the article_titles array only contains unique titles, then technically, I can remove the call to .uniq! later in the batching logic — since there would be no duplicates to worry about.
However, if there's any chance the titles fetched are not unique, it's still safer to keep .uniq! to prevent duplicate article_ids from being included.

I think it's always better to fix the problem at its root. If we suspect there are duplicate titles in the `article_titles` category field (I don't think there are since they come from an API, but it may be worth double-checking), then we should use `uniq` before saving the titles to the database. That’s preferable to saving duplicates in the db and relying on `uniq` later when mapping them to article IDs just to save a bit of memory.
The `article_titles` category field is one of the most memory-intensive resources when fetched, so we definitely want to avoid duplicates there.
I'm okay with removing the `uniq` from the `article_ids` method. I'm adding a note in issue #6415 to ensure title uniqueness in the `article_titles` category field.

Does it sound reasonable for you?

Abishekcs:
> (I don't think there are since they come from an API, but it may be worth double-checking)

I will check it once to be sure.



> I'm okay with removing the `uniq` from the `article_ids` method.

Okay got it.



> Does it sound reasonable for you?

Yes. Thank you for the review.

ragesoss:
@Abishekcs is the benchmark data in the opening comment based on the the current version of the PR?

@gabina this looks good to you now?

Abishekcs:
> is the benchmark data in the opening comment based on the the current version of the PR?

this is the latest benchmark 

https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6410#issuecomment-3084151938.  for the current PR and the same benchmark  is shown in the PR description.

Also, after [removing  unique in this commit](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6410/commits/df2f3048ba41a60d25f3694572af55aae50af25d) there is not much difference in the benchmark in terms of memory and speed too.

ragesoss:
The speed up seems significant. The memory difference seems small enough that I don't think it's likely to be relevant for us. Looks like it's about 13MB more allocated and 0.16MB more retained, so it's unlikely to significantly affect the memory pressure.

Abishekcs:
Also, I forgot to mention: I changed the batch size from 500 to 2000 because with a batch size of 500, the performance improved by about 20–30%. But with a batch size of 2000, the improvement is consistently over 50%.

I didn't went beyond 2000.

ragesoss:
@Abishekcs interesting. i wonder if optimal batch size is related to how sparse the Articles table is relatively to the title lists. For large categories, I suspect most of the titles will not have Article records. I'm interested to see a comparison with several additional batch sizes.

Abishekcs:
> I'm interested to see a comparison with several additional batch sizes.

Okay I will try with different batches

gabina:
> @gabina this looks good to you now?

Yes, changes are OK from my perspective.

As regards what we have been discussing, just to be extra clear, `article_ids` method may return duplicate ids with these changes. Beyond unnecessary memory usage, this doesn't cause any functionality issues, since `article_ids` is only used to define `scoped_article_ids`, which is then used in IN clauses, where duplicates don't affect the results. Once I handle issue #6415, we’ll be guaranteed that `article_ids` no longer contains duplicates.

Abishekcs:
I will also post the benchmark with different batches soon.

Abishekcs:
[Benchmark.txt](https://github.com/user-attachments/files/21386954/Benchmark.txt) (Text File of the benchmark)

**Note:**

**The [previous benchmark](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6410#issuecomment-3084151938) was conducted on a dataset of 756,000 records, whereas the current benchmark utilizes a significantly larger dataset of 1,756,965 records.**

**I attempted to run a single query against the full dataset of 1,756,965 records, but encountered a "Server has gone away" error, indicating that the query likely exceeded server limits or resource thresholds.**

Screenshot from 2025-07-23 17-42-41

So, I was able to run single query for maximum data set of `780,000`

Screenshot from 2025-07-23 18-09-24




### **Average Timing by Batch Size**

- **Data Set Size:**  `1,756,965`

| **Batch Size** | **User (s)** | **System (s)** | **Total (s)** | **Real (s)** |
| -------------- | -----------: | -------------: | ------------: | -----------: |
| 500            |      12.4353 |         0.9883 |       13.4230 |      30.9571 |
| 1000           |      10.4699 |         0.6922 |       11.1621 |      26.6449 |
| 2000           |       9.2426 |         0.5693 |        9.8119 |      23.5579 |
| 3000           |       8.5110 |         0.4915 |        9.0025 |      22.5729 |
| 4000           |       8.1602 |         0.4543 |        8.6145 |      19.8409 |
| 5000           |       7.5619 |         0.4230 |        7.9849 |      17.7325 |
| 6000           |       7.4183 |         0.4196 |        7.8379 |      17.2583 |
| 7000           |       7.3496 |         0.4043 |        7.7539 |      16.9953 |
| 8000           |       7.0449 |         0.3844 |        7.4292 |      16.4889 |
| 9000           |       7.0824 |         0.3856 |        7.4681 |      16.6109 |
| 10000          |       6.8437 |         0.3685 |        7.2122 |      16.0460 |


-   `10000` yields the **lowest real execution time** of **16.046 seconds**, along with the lowest CPU usage metrics across the board.

- There's a consistent drop in all time metrics as the batch size increases — particularly steep improvements from 500 to 4000. Gains taper after batch 8000, indicating **diminishing returns**.



### Code


- **Batch**

```ruby

   def article_ids
      require 'benchmark'

      articles_titles = Article.all.pluck(:title)

        # Second approach - Batched queries
      batched_query_time = Benchmark.measure do
        @article_ids2 = []
        articles_titles.each_slice(10000) do |batch|
            @article_ids2.concat(Article.where(namespace: 0, wiki_id:, title: batch).pluck(:id))
        end
      end

      puts 'Batched Query Results:'
      puts batched_query_time
      puts "Benchmarking with #{articles_titles.size} titles..."
   end
   
```

- **Single**

```ruby

  def article_ids
    require 'benchmark'

       articles_titles = Article.all.pluck(:title).first(780000)
       single_query_time = Benchmark.measure do
         @article_ids1 = Article.where(namespace: 0, wiki_id:, title: articles_titles).pluck(:id)
       end

       puts 'Single Query Results'
       puts single_query_time
       puts "Benchmarking with #{articles_titles.size} titles..."
  end
  
```

ragesoss:
Excellent. Perhaps 5000 is a good batch size to go with?

Abishekcs:
> Perhaps 5000 is a good batch size to go with?

Okay, 5,000 seems like a good choice as it sits around the middle of all the batch sizes tested, both in terms of size and performance. I’ll make a commit to change the batch size.

PR #6417, opened 2025-07-18
Do not process the current timeslice twice in the same course update

gabina:
## What this PR does
This PR makes two basic things:
### 1- Fix a bug related to reprocessing and `last_mw_rev_datetime`.

The ingestion process for the timeslice system starts fetching data from the latest timeslice (i.e., the one with the greatest `start` field) where `last_mw_rev_datetime` is not nil. This is implemented in the `get_ingestion_start_time_for_wiki` method.
The process relies on the guarantee that all previous timeslices (those with an earlier `start` field) have already been processed - even if they contain zero revisions.

However, when reprocessing a random timeslice (for example, because a new user was added), we may update the `last_mw_rev_datetime` field for that timeslice. This can break the guarantee mentioned above. For instance, if the latest timeslice with `last_mw_rev_datetime != nil` has `start: 2025-07-10`, and we reprocess a timeslice with `start: 2025-07-18`, then the next time we invoke `get_ingestion_start_time_for_wiki`, it would return 2025-07-18. As a result, the timeslices between 2025-07-10 and 2025-07-18 would never be processed.

To fix this bug, we now reprocess a timeslice only if it's a "future" timeslice - that is, only if its `start` field is later than the ingestion start time.

Commit: bb815e4fc1f3cd3be8d878a897065194c994257d

### 2- Avoid processing the current timeslice if it has already been reprocessed in the same course update round.
Instead of saving only the count of reprocessed timeslices, we now store the `wiki_id` and `start` time for each timeslice reprocessed during the update round. With that information, we're able to identify which timeslices were already reprocessed for a given course. Before processing a timeslice, we check whether it was already reprocessed based on its `wiki_id` and `start` time. If it was, we skip it.

Commit: c01b6b4c5f2e7d0a21698fbc85ec5691180fee29


gabina:
There is a failing spec. This doesn't fail locally. I suspect the problem is that the request failed transiently when saving VCR files.
```
Failures:

  1) UpdateCourseWikiTimeslices when a full update is required sets needs_update to false if update is successful
     Failure/Error: expect(course.course_wiki_timeslices.where(needs_update: true).count).to eq(0)

       expected: 0
            got: 2

       (compared using ==)
     # ./spec/services/update_course_wiki_timeslices_spec.rb:290:in `block (3 levels) in <top (required)>'
```

ragesoss:
I've deleted that cache

gabina:
Interesting. The spec failed again but now for different revisions (only one timeslice with error instead of two). 
```
Failures:

  1) UpdateCourseWikiTimeslices when a full update is required sets needs_update to false if update is successful
     Failure/Error: expect(course.course_wiki_timeslices.where(needs_update: true).count).to eq(0)

       expected: 0
            got: 1

       (compared using ==)
     # ./spec/services/update_course_wiki_timeslices_spec.rb:290:in `block (3 levels) in <top (required)>'
```

There are several spec files using the same `course_update` VCR cassette. From what I've seen locally, it looks like some requests are mixed up, so I'm decoupling the VCR cassettes.

PR #6423, opened 2025-07-23
[WIP] Move course csv generation to background workers

### LINKED ISSUE #6414: Generate course CSV stats report asynchronously
### What is happening?
Course CSV stats report may be hard to run in a web request, as they involve heavy queries. For huge courses, trying to generate these reports may cause a dashboard downtime.

### To Reproduce
Please don't reproduce this in production

### Possible solution
We should move course CSV stats generation to background workers similar to what we did in #4376

### Additional context
I was able to reproduce this in production for course 31480 [Editing and Contributing Training - University Students](https://outreachdashboard.wmflabs.org/courses/Levant/Editing_and_Contributing_Training_-_University_Students)

gabina:
## What this PR does
WIP.

Closes #6414 
## Screenshots
Before:

After:

### DIFF
--- a/app/controllers/analytics_controller.rb
+++ b/app/controllers/analytics_controller.rb
@@ -48,8 +48,7 @@ class AnalyticsController < ApplicationController
   end
 
   def course_csv
-    send_data CourseCsvBuilder.new(@course, per_wiki: true).generate_csv,
-              filename: "#{@course.slug}-#{Time.zone.today}.csv"
+    csv_of('overview')
   end
 
   def tagged_courses_csv
@@ -59,23 +58,19 @@ class AnalyticsController < ApplicationController
   end
 
   def course_uploads_csv
-    send_data CourseUploadsCsvBuilder.new(@course).generate_csv,
-              filename: "#{@course.slug}-uploads-#{Time.zone.today}.csv"
+    csv_of('uploads')
   end
 
   def course_students_csv
-    send_data CourseStudentsCsvBuilder.new(@course).generate_csv,
-              filename: "#{@course.slug}-editors-#{Time.zone.today}.csv"
+    csv_of('editors')
   end
 
   def course_articles_csv
-    send_data CourseArticlesCsvBuilder.new(@course).generate_csv,
-              filename: "#{@course.slug}-articles-#{Time.zone.today}.csv"
+    csv_of('articles')
   end
 
   def course_wikidata_csv
-    send_data CourseWikidataCsvBuilder.new(@course).generate_csv,
-              filename: "#{@course.slug}-wikidata-#{Time.zone.today}.csv"
+    csv_of('wikidata')
   end
 
   def all_courses_csv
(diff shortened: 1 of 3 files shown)


gabina:
This PR follows the approach used in #4376. There is a bit of code duplication. For example, `CourseCsvWorker` is pretty similar to `CampaignCsvWorker`. The logic in `AnalyticsController#csv_of` is almost the same as in `CampaignsController#csv_of`.
On the other hand, `CsvCleanupWorker` is re-used since all the files are saved in the same folder: `CampaignsController::CSV_PATH`.

What are your thoughts about creating a single class that takes care of all CSV reports generated in the background instead of having them split up between `AnalyticsController` (course CSVs) and `CampaignsController` (campaigns CSV)? Note that `AnalyticsController` class has also the `all_courses_csv` which seems to be a campaign report, so logic encapsulation is not pretty clear.

Edit: I tagged a wrong user (sage)

ragesoss:
I think moving them all to a single controller is a good idea.

lib/analytics/campaign_csv_builder.rb
```diff
@@ -3,7 +3,6 @@
 require 'csv'
 require_dependency "#{Rails.root}/lib/analytics/course_csv_builder"
 require_dependency "#{Rails.root}/lib/analytics/course_articles_csv_builder"
-require_dependency "#{Rails.root}/app/workers/campaign_csv_worker"
```
gabina, line 6: I don't think this dependency was necessary?

app/controllers/reports_controller.rb
```diff
@@ -0,0 +1,109 @@
...
+
+  # Builds the filename for a report of the given type, based on wether @course is defined
+  # or @campaign is defined
+  def build_filename(type)
+    # Filename does not have to contain '/' char because it's interpreted as a route
+    return "#{@course.slug}-#{type}-#{Time.zone.today}.csv".tr('/', '-') if course_report?(type)
+
+    include_course_segment = csv_params[:course] ? '-with_courses' : ''
+    "#{@campaign.slug}-#{type}#{include_course_segment}-#{Time.zone.today}.csv".tr('/', '-')
+  end
```
gabina, line 95: Note that `.tr('/', '-')` is necessary because the `/` character is interpreted as a path separator, which would cause new folders to be created. To reuse `CsvCleanupWorker` as-is and avoid introducing the need for additional testing, I forced the files to be created directly under `public#{CSV_PATH}`, without any nested folder structure.

app/workers/report_csv_worker.rb
```diff
@@ -0,0 +1,70 @@
...
+  private
+
+  def write_csv(filename, data)
+    FileUtils.mkdir_p "public#{ReportsController::CSV_PATH}"
+    File.write "public#{ReportsController::CSV_PATH}/#{filename}", data
+  end
+
+  def course_report?(type)
+    type.start_with?('course')
+  end
```
gabina, line 74: I had to duplicate this method in `ReportCsvWorker` and `ReportsController`

lib/analytics/campaign_csv_builder.rb
ragesoss: Left over from previous refactor, I guess.

ragesoss:
LGTM!

PR #6424, opened 2025-07-24
[WIP]: Optimize/campaign articles deferred pagination

Abishekcs:

## What this PR does

## Screenshots
Before:

After:


ragesoss:
Thanks! In cases like this, I find it helpful to have an even more detailed commit  message to explain what problem the eager-loading solves (ie, specifically what went wrong with reference to the commit that broke eager loading but wasn't detected until it was deployed to production).

I guess the reason to do this in the CI config instead of the `test` environment is because it will make it slower to run individual tests in a development situation (eg, where `guard` is automatically running related tests every time a file is saved). That might also merit a mention in the commit message.

app/presenters/courses_presenter.rb
```diff
@@ -65,7 +64,7 @@ def course_ids_and_slugs
   def articles_courses_scope
     return @articles_courses_scope unless @articles_courses_scope.nil?
 
-    @articles_courses_scope = RankedArticlesCoursesQuery.new(
+    @articles_courses_scope = Query::RankedArticlesCoursesQuery.new(
```
ragesoss, line 67: Is Query already an existing class (like, from Rails)?

Alternatively, is there another way besides namespacing the classname to avoid eager load errors?

If this is the only way, it's fine, but I don't like it.

Abishekcs: > Is Query already an existing class (like, from Rails)?

I am not sure but if there was i think Rails will throw an error?

Abishekcs: > Alternatively, is there another way besides namespacing the classname to avoid eager load errors?

Another way will be to remove query/ folder and move it's file into a top-level directory like:
app/presenters/ranked_articles_courses_query.rb


Will this approach be better?

ragesoss: Yeah, I guess so. No other option to give the eager loader the right hint?

ragesoss: The namespace is slightly better than moving the file, I think.

Abishekcs: > The namespace is slightly better than moving the file, I think.

Okay.

Abishekcs:
@ragesoss 

The main functionality of the PR remains the same as before — aside from the namespace fix and the new CI check for eager loading, there are no changes to the core logic.

The only updates are in the two new commits below:

* [ `2837576`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6424/commits/283757683230ef1ba3470aedd6e263ab2b54a3da)
*  [`cfe9470`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6424/commits/cfe9470dfdb91e9dcb5c3690e7a03fed422bd039)

Whenever you're free, please take another look at this PR thank you so much!

PR #6428, opened 2025-07-26
[WIP] Auto-hide Wikidata stats section when all values are zero

### LINKED ISSUE #6413: Toggle to hide "Wikidata stats" section in campaigns
### Current Behavior:
In a campaign, a section called "Wikidata stats" is displayed bellow the common stats. Even if Wikidata was not worked on in any of the campaign's edit-a-thons.

### Desired Behavior:
A way to toggle it so that it does not appear.

Imvedansh:
## What this PR does
Fixes #6413 
Auto-hides the Wikidata stats section when all statistics are zero, preventing unnecessary UI clutter.
### Problem
The Wikidata stats section was displayed even when all values were 0, taking up screen space and providing no value to users (as shown in campaigns like Wiki Loves Sport 2025).

### Solution
- Added `allStatsAreZero()` helper function to check if all Wikidata statistics are zero or undefined
- Component now returns `null` when all stats are zero, hiding the entire section
- Maintains existing functionality when any stat has a value > 0
- CSV download behavior remains unchanged as requested
- - Fixed ESLint warnings by replacing hardcoded strings with I18n translations for "Descriptions", "Aliases", and "Other" labels

###  Files Changed
- `app/assets/javascripts/components/common/wikidata_overview_stats.jsx`

### Testing
- Test with campaigns that have zero Wikidata activity (should hide the section)
- Test with campaigns that have Wikidata activity (should display normally)
- Test campaigns without Wikidata tracking (should remain unchanged)

### DIFF
--- a/app/assets/javascripts/components/common/wikidata_overview_stats.jsx
+++ b/app/assets/javascripts/components/common/wikidata_overview_stats.jsx
@@ -150,7 +150,7 @@ const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
           </div>
         </div>
         <div className="stat-display__row">
-          <h5 className="stats-label">Descriptions</h5>
+          <h5 className="stats-label">{I18n.t('metrics.descriptions')}</h5>
           <div className="stat-display__value-group">
             <OverviewStat
               id="descriptions-added"
@@ -176,7 +176,7 @@ const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
           </div>
         </div>
         <div className="stat-display__row">
-          <h5 className="stats-label">Aliases</h5>
+          <h5 className="stats-label">{I18n.t('metrics.aliases')}</h5>
           <div className="stat-display__value-group">
             <OverviewStat
               id="aliases-added"
@@ -202,7 +202,7 @@ const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
           </div>
         </div>
         <div className="stat-display__row double-row">
-          <h5 className="stats-label">Other</h5>
+          <h5 className="stats-label">{I18n.t('metrics.other')}</h5>
           <div className="stat-display__value-group double">
             <OverviewStat
               id="qualifiers-added"
--- a/app/assets/javascripts/components/common/wikidata_overview_stats.jsx
+++ b/app/assets/javascripts/components/common/wikidata_overview_stats.jsx
@@ -3,6 +3,45 @@ import PropTypes from 'prop-types';
 import OverviewStat from './OverviewStats/overview_stat';
 
 const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
+  // Helper function to check if all stats are zero
+  const allStatsAreZero = () => {
+    const statsToCheck = [
+      'total revisions',
+      'merged to',
+      'interwiki links added',
+      'items created',
+      'items cleared',
+      'claims created',
+      'claims changed',
+      'claims removed',
+      'labels added',
+      'labels changed',
+      'labels removed',
+      'descriptions added',
+      'descriptions changed',
+      'descriptions removed',
+      'aliases added',
+      'aliases changed',
+      'aliases removed',
+      'qualifiers added',
+      'references added',
+      'redirects created',
+      'reverts performed',
+      'restorations performed',
+      'other updates',
+      'lexeme items created'
+    ];
+
+    return statsToCheck.every(statKey =>
+      !statistics[statKey] || statistics[statKey] === 0
+    );
+  };
+
+  // If all stats are zero, don't render the component
+  if (allStatsAreZero()) {
+    return null;
+  }
+
   let containerClass = 'wikidata-stats-container';
   let title = 'Wikidata stats';
   if (isCourseOverview) {
@@ -26,14 +65,14 @@ const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
             <OverviewStat
               id="merged"
               className="stat-display__value-small"
-              stat = {statistics['merged to']}
+              stat={statistics['merged to']}
               statMsg={I18n.t('metrics.merged')}
               renderZero={true}
             />
             <OverviewStat
               id="interwiki-links"
               className="stat-display__value-small"
-              stat = {statistics['interwiki links added']}
+              stat={statistics['interwiki links added']}
               statMsg={I18n.t('metrics.interwiki_links_added')}
               renderZero={true}
             />
@@ -227,7 +266,7 @@ const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
         </div>
       </div>
     </div>
-    );
+  );
 };
 
 WikidataOverviewStats.propTypes = {
@@ -235,4 +274,4 @@ WikidataOverviewStats.propTypes = {
   isCourseOverview: PropTypes.bool
 };
 
-  export default WikidataOverviewStats;
+export default WikidataOverviewStats;


Imvedansh:
Screenshot 2025-07-26 152449

app/assets/javascripts/components/common/wikidata_overview_stats.jsx
```diff
@@ -3,6 +3,45 @@ import PropTypes from 'prop-types';
 import OverviewStat from './OverviewStats/overview_stat';
 
 const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
+  // Helper function to check if all stats are zero
+  const allStatsAreZero = () => {
```
Imvedansh, line 7: helper function to check if all Wikidata statistics are zero

Imvedansh:
This is ready for a review!!

app/assets/javascripts/components/common/wikidata_overview_stats.jsx
```diff
@@ -3,6 +3,45 @@ import PropTypes from 'prop-types';
...
+    const statsToCheck = [
```
Abishekcs: Since you mentioned the required stats for Wikidata in `statsToCheck`, I assume there are additional statistical values in the `statistics` object beyond those listed in `statsToCheck`?

Imvedansh: Yes, the statsToCheck array includes all the statistical keys that are rendered in this component. If there are any other they are not getting render here.
Are you suggesting that i missed some?

Abishekcs: > Are you suggesting that i missed some?

No

Imvedansh: > Since you mentioned the required stats for Wikidata in `statsToCheck`, I assume there are additional statistical values in the `statistics` object beyond those listed in `statsToCheck`?

I included all parameter's which are rendered, if there are additional keys in the statistics object, they can be safely ignored because they are not used in rendering this WikidataOverviewStats component.
WDYT?

Imvedansh:
@Abishekcs What do you think about this?, Id really appreciate your views

Abishekcs:
> @Abishekcs What do you think about this?, Id really appreciate your views

Is there an easier way to implement this besides using a `helper function to check if all stats are zero?` A helper function is great, but is there another way that doesn't involve checking each stat individually? If not it's okay.

Imvedansh:
> Is there an easier way to implement this besides using a `helper function to check if all stats are zero?` A helper function is great, but is there another way that doesn't involve checking each stat individually? If not it's okay.

Sure, umm like for instance  Object.values() with every() but it will checks all properties in statistics object, not just the ones we render,which is not ideal  like
 ` allStatsAreZero = () => {
  return Object.values(statistics).every(value => !value || value === 0);
};`
but with `helper function`
-it Only checks stats that are actually rendered in this component
-also Won't break if the statistics object gets additional properties in the future
-and Clear what stats affect the visibility logic.
Ultimately this is most suitable ,surely in future can try other approach if needed

Abishekcs:
> Ultimately this is most suitable ,surely in future can try other approach if needed

ok

Imvedansh:
@Abishekcs Ik this is not the place but i haven't receive the invite for slack yet, are there any other medium to discuss issue and ask questions?

Abishekcs:
> Ik this is not the place but i haven't receive the invite for slack yet, are there any other medium to discuss issue and ask questions?

hmm no i think slack and github are the only place.

Abishekcs:
> haven't receive the invite for slack yet

Just a quick update on the Slack invite — I believe it might take another week, as ragesoss usually handles those and he may not be available until next week.

Abishekcs:
until then you can post your question on the github issue or pr you are working.

ragesoss:
@Abishekcs are you happy with the current state of this?

Abishekcs:
> @Abishekcs are you happy with the current state of this?

Yes, beside from the `Helper function` i guess there is no other way to do this.

app/assets/javascripts/components/common/wikidata_overview_stats.jsx
```diff
@@ -3,6 +3,45 @@ import PropTypes from 'prop-types';
...
+      'aliases changed',
+      'aliases removed',
+      'qualifiers added',
+      'references added',
+      'redirects created',
+      'reverts performed',
+      'restorations performed',
+      'other updates',
+      'lexeme items created'
+    ];
```
ragesoss: This seems like it's duplicating a list of keys that might change over time, and then this list might not match.

Is there a good way to do this check without relying on an explicit list of keys? Can we just loop through all the keys that are present and check them for non-zeroes?

Imvedansh: I understand the concern about maintenance, but I chose the explicit key list approach for several important reasons for instance
-we only want to hide the component when these particular statistics are zero.
-If the backend ever adds metadata fields, configuration flags, or non-statistic data to the statistics object, the generic Object.keys() approach could break unexpectedly. 
-being explicit about which fields we're checking makes the code more predictable and easier to debug, WDYT?
alternatively could have easily 
` return Object.keys(statistics).every(key => 
    !statistics[key] || statistics[key] === 0
  );`

ragesoss: This component only receives an object of Wikidata stats specifically, if we add new keys to the object later on, they will likely correspond to new Wikidata stats that might be displayed, with corresponding changes to rendered component. I think the `Object.keys` approach is fine here.

Imvedansh: I see. Though previously I think I was just being too cautious, and the resulting changes were unnecessary. Sure, let’s stick with looping through all keys.

Imvedansh: @ragesoss WDYT? , My bad bout the delay

ragesoss: These seems good. I will test it this week or next. Thanks!

app/assets/javascripts/components/common/wikidata_overview_stats.jsx
```diff
@@ -163,7 +175,7 @@ const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
           </div>
         </div>
         <div className="stat-display__row double-row">
-          <h5 className="stats-label">Other</h5>
+          <h5 className="stats-label">{I18n.t('metrics.other')}</h5>
```
ragesoss: This should be 'metrics.other_updates'.

Imvedansh: Got it! Will update to `metrics.other_updates`. Thanks for the correction.

Imvedansh: @ragesoss Im not completely positive but i dont think this build fail is related to PR

Imvedansh: @ragesoss No rush, but whenever you get time, could you let me know the changes I need to make? Looking forward to it.

ragesoss: @Imvedansh I left a comment on the issue thread... this works correctly as far as I can tell, but I couldn't replicate the bug it's supposed to solve, so I'm not actually sure this change the behavior of the system at all.

ragesoss:
I'm not sure this makes any functional difference, as I was not able to replicate the issue. We can revisit this if we learn more from the person who reported the issue, but meanwhile I'm going to close this PR. Thanks for your efforts @Imvedansh!

PR #6435, opened 2025-07-31
Conditional Questions Persisting When Parent Selection Changes in Surveys 

Imvedansh:
### Fixes: Conditional Questions Persisting When Parent Selection Changes
 Fixes issue #6216  

### Problem
When users changed their selection on a parent question with conditional logic, previously shown conditional questions would remain visible alongside the new conditional questions. For example:
- User selects Option A → Conditional Question A appears
- User changes to Option B → Both Conditional Question A and B are visible (BUG)

###  Root Cause
The `handleParentConditionalChange()` function used complex logic to selectively reset conditional questions, which failed to properly hide previously shown conditionals when the parent selection changed.

## Solution
Simplified the conditional reset logic with a "reset-first" approach

### Key Changes:
1. **Always reset all conditionals first** before showing new ones
2. **Simplified `resetConditionalGroupChildren()`** to always reset all children instead of selective exclusion

###  Files Modified:
- `app/assets/javascripts/surveys/modules/Survey.js`
  - Modified `handleParentConditionalChange()` method
  - Simplified `resetConditionalGroupChildren()` method

### Before

https://github.com/user-attachments/assets/f4064a21-20ae-4721-9081-721709d94946

### After

https://github.com/user-attachments/assets/fac1d671-4053-4d9d-8956-db2046c3aede

### DIFF
--- a/app/assets/javascripts/surveys/modules/Survey.js
+++ b/app/assets/javascripts/surveys/modules/Survey.js
@@ -314,13 +314,13 @@ const Survey = {
             answerGroup[answerId] = val;
           }
         } else if (value !== '0') { // Multi-Select (Checkbox)
-           if (typeof answerGroup[answerId] !== 'undefined') {
-             answerGroup[answerId][answerKey].push('0');
-             answerGroup[answerId][answerKey].push(value);
-           } else {
-             answerText[answerKey] = ['0', value];
-             answerGroup[answerId] = answerText;
-           }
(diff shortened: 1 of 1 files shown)


Imvedansh:
I dont think this build failure is related to the PR

Imvedansh:
Build is failing when im trying to remove some trailing spaces

PR #6437, opened 2025-08-01
Conditional Questions Persisting When Parent Selection Changes in Surveys

Imvedansh:
### Fixes: Conditional Questions Persisting When Parent Selection Changes
Fixes issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6216

### Problem
When users changed their selection on a parent question with conditional logic, previously shown conditional questions would remain visible alongside the new conditional questions. For example:

User selects Option A → Conditional Question A appears
User changes to Option B → Both Conditional Question A and B are visible (BUG)

### Root Cause
The handleParentConditionalChange() function used complex logic to selectively reset conditional questions, which failed to properly hide previously shown conditionals when the parent selection changed.

### Solution
Simplified the conditional reset logic with a "reset-first" approach

### Key Changes:
Always reset all conditionals first before showing new ones
Simplified resetConditionalGroupChildren() to always reset all children instead of selective exclusion

### Files Modified:
app/assets/javascripts/surveys/modules/Survey.js
Modified handleParentConditionalChange() method
Simplified resetConditionalGroupChildren() method

### Before

https://github.com/user-attachments/assets/9b26849c-acbc-4752-8abc-5e08218a364f

### After

https://github.com/user-attachments/assets/8401eb41-cd82-4cfd-84c5-c17395a48bff

### DIFF
--- a/app/assets/javascripts/surveys/modules/Survey.js
+++ b/app/assets/javascripts/surveys/modules/Survey.js
@@ -627,48 +627,36 @@ const Survey = {
   },
 
   handleParentConditionalChange(value, conditionalGroup, $parent) {
-    let { currentAnswers } = conditionalGroup;
     let conditional;
-    // let resetQuestions = false;
 
-    if (Array.isArray(value)) {
-      // Check if empty
-      if (value.length === 0 && currentAnswers) {
-        conditionalGroup.currentAnswers = [];
-      }
-
-      // Check if conditional was present and is no longer
-      currentAnswers.forEach((a) => {
-        if (value.indexOf(a) === -1) {
-          const index = currentAnswers.indexOf(a);
-          if (currentAnswers.length === 1) {
-            currentAnswers = [];
-          } else {
-            currentAnswers = currentAnswers.slice(index, index + 1);
-          }
-        }
-      });
-      // Check if value matches a conditional question
-      value.forEach((v) => {
-        if (conditionalGroup[v] !== undefined
-            && currentAnswers.indexOf(v) === -1) {
-          conditional = conditionalGroup[v];
-          currentAnswers.push(v);
-          conditionalGroup.currentAnswers = currentAnswers;
-        }
-      });
-
-      if (currentAnswers.length === 0) {
-        conditionalGroup.currentAnswers = [];
-      }
-    } else {
-      conditional = conditionalGroup[value];
-    }
-
+    // Always reset all conditional questions first to clear any previously shown conditionals
     this.resetConditionalGroupChildren(conditionalGroup);
 
-    if (typeof conditional !== 'undefined' && conditional !== null) {
-      this.activateConditionalQuestion($(conditional), $parent);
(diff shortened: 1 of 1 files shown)


Imvedansh:
I hope this is helpful!

Abishekcs:
I will try to review it.

Imvedansh:
> I will try to review it.

Thanks, that would be great

Abishekcs:
I have tested this PR it seems to fix the bug. But there is one more bug which exists in the main branch and even in this PR. I have attached a screen recording below on how to reproduce it, If you can fix that too will be great. There might be a solution that fixes both of this at the same time.

https://github.com/user-attachments/assets/cacf2fdd-20f4-44ea-ad87-393355411b1b

Imvedansh:
> I have tested this PR it seems to fix the bug. But there is one more bug which exists in the main branch and even in this PR. I have attached a screen recording below on how to reproduce it, If you can fix that too will be great. There might be a solution that fixes both of this at the same time.
> 
>  Bug.Survey.mp4

Thanks for reviewing,Sure let me resolve it , most likely to be fixed with same solution

Imvedansh:
@Abishekcs I encounter more similar issue in surveys, i would take some more time to fix and come up with the solution, what do you suggest should i raise  a Seprate PR for these fixes or combine within this one?

Abishekcs:
> what do you suggest should i raise a Seprate PR for these fixes or combine within this one?

Separate and smaller PR is best.

Imvedansh:
Sure, I’ll keep the current fixes within this PR and raise another one as a progression of this. Given the current PR description, the changes have been made and are working fine. I’ll raise a Part-2 PR separately with the new findings and your suggested bug fix.
IG then this PR is done ?

Abishekcs:
> IG then this PR is done ?

Feel free to work on the other PR. If there are any more changes required in this PR i will let you know.

Imvedansh:
> > IG then this PR is done ?
> 
> Feel free to work on the other PR. If there are any more changes required in this PR i will let you know.

Great!!

Abishekcs:
@Imvedansh i forgot to ask. But i hope PR-2 will be independent of this PR (Like not dependent on this PR)?

Imvedansh:
@Abishekcs I haven't resolve all but i think they will be based on this PR, This PR have chnges which will affect the functionality ,
For instance the bug you suggested is reproduced in main as well but this PR brings us closer to fix.
If you are not sure with this PR, i can try for completely independent PR as well but i think some changes would be common even then.

ragesoss:
@Abishekcs are you happy with this PR in its current state?

Abishekcs:
> @Abishekcs are you happy with this PR in its current state?

Yes. I have tested this one out and it works properly.

Abishekcs:
> If you are not sure with this PR, i can try for completely independent PR as well but i think some changes would be common even then.

It's okay

PR #6451, opened 2025-08-12
Refactor CoursesPresenter to use shared campaign_articles_count helper

Abishekcs:
Found in Sentry Transaction Summary [#a480922522d3 ](https://wiki-education.sentry.io/insights/summary/trace/a480922522d340f38599e585f62a405c/?breakdown=none&fov=0%2C503.5791015625&node=span-ae33988a45739df5&project=6269916&query=&showTransactions=p95&sort=-transaction.duration&source=performance_transaction_summary&tab=events&timestamp=1754953168&transaction=CampaignsController%23articles)of **CampaignsController#articles**

## What this PR does

Refactors `CoursesPresenter` to introduce a new `campaign_articles_count` method that:
- Centralizes logic for counting tracked campaign articles
- Uses memoization to avoid repeated queries within a request
- Ensures both `too_many_articles?` and pagination metadata use the same count logic

### Changes
- Removed `total_articles_count` method
- Added `campaign_articles_count` with `tracked: true` filter
- Updated `paginated_articles_courses` and `too_many_articles?` to use the new helper

### Benefits
- Keeps counting logic DRY and consistent
- Ensures the tracked filter is applied consistently
- Reduces duplicate query overhead with memoization

No change to end-user behavior, Except for reducing Query time to count the number of articles which was required for pagination.

## Screenshots

Screenshot from 2025-08-13 01-52-25

**Before:**

SCR 1

---
**After:**

Screenshot from 2025-08-13 01-44-13

### DIFF
--- a/app/presenters/courses_presenter.rb
+++ b/app/presenters/courses_presenter.rb
@@ -85,14 +85,9 @@ class CoursesPresenter
 
   # Create paginated collection for articles_courses data
   def paginated_articles_courses
-    WillPaginate::Collection.create(current_page, PER_PAGE, total_articles_count) do |pager|
+    WillPaginate::Collection.create(current_page, PER_PAGE, campaign_articles_count) do |pager|
       pager.replace(articles_courses_scope)
     end
-  end
-
-  # Count total articles for pagination metadata
-  def total_articles_count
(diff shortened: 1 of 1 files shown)


ragesoss:
Looks good! This is probably better logic for as well.

Abishekcs:
Also, the query time most the time is same with `tracked: true` and without `tracked:true`

PR #6457, opened 2025-09-02
[WIP] Update timeslices marked as `needs_update` when changing timeslice duration

gabina:
## What this PR does
This PR updates the behavior of `UpdateTimeslicesCourseWiki`.
Previously, `UpdateTimeslicesCourseWiki.update_timeslices_durations` deleted and recreated unprocessed timeslices if their duration had changed, but it ignored timeslices marked as `needs_update`.
With this PR, timeslices marked as `needs_update` are also deleted and recreated with the new duration.

This matters because timeslices with a large number of revisions (e.g., more than 100K) are often marked as `needs_update` due to transient API failures when fetching scores. Having such large timeslices repeatedly reprocessed in every course update involves a lot of latency in the queue. By splitting them into smaller timeslices, we increase the likelihood that they are successfully updated.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -46,6 +46,7 @@ Metrics/ClassLength:
     - 'lib/training/wiki_slide_parser.rb'
     - 'lib/wiki_assignment_output.rb'
     - 'lib/course_training_progress_manager.rb'
+    - 'lib/timeslice_cleaner.rb'
     - 'lib/timeslice_manager.rb'
 
 Metrics/AbcSize:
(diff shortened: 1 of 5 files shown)


gabina:
Once this is deployed, we can manually update timeslice duration for course [32963](https://outreachdashboard.wmflabs.org/courses/Wikimedia_Indonesia/Geodatathon_Agustus_2025/home) from default (1 day) to 1 hour.  Since it’s only a 3-day course, using hourly timeslices is not a big deal, and it gives us an easy way to see how a course looks with smaller timeslices (of course the difference won’t be noticeable without checking the DB).

PR #6460, opened 2025-09-08
Fix article views metric for campaign view.

gabina:
## What this PR does
This PR fixes a minor bug related to the article views metric. On the `campaigns/articles` route, we were displaying the raw  `articles_courses.view_count` value. Since the article views metric implementation has changed, we can’t rely on that value anymore. The correct approach is to use the `calculate_view_count` helper (read method docs).

## Screenshots
Before:
before_article_views

After:
after_article_views

### DIFF
--- a/spec/lib/importers/average_views_importer_spec.rb
+++ b/spec/lib/importers/average_views_importer_spec.rb
@@ -7,13 +7,14 @@ describe AverageViewsImporter do
   let(:article) { create(:article, title: 'Selfie') }
   let(:course) { create(:course) }
 
-  let!(:articles_course) do
+  let(:articles_course) do
     create(:articles_course, course:, article_id: article.id, average_views: 1,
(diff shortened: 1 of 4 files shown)


spec/lib/importers/average_views_importer_spec.rb
```diff
@@ -7,13 +7,14 @@
...
-  let!(:articles_course) do
+  let(:articles_course) do
     create(:articles_course, course:, article_id: article.id, average_views: 1,
     average_views_updated_at: 1.day.ago, first_revision: 10.days.ago)
   end
 
   before do
     travel_to Date.new(2025, 8, 28)
+    articles_course
   end
```
gabina, line 18: Note this fix is not related to this PR. The thing here is that is we need the test to travel in time first and then load `articles_course`.

PR #6461, opened 2025-09-10
Implement strategy to automatically split timeslices if there are more than REVISION_THRESHOLD

gabina:
## What this PR does
This PR implements the adaptive timeslice splitting strategy through a new `SplitTimeslices` class. The algorithm starts with a default timeslice (usually, 1-day timeslices), fetches revisions for that period and, if the timeslice exceeds the threshold (currently set to 10K revisions), recursively splits it until all timeslices are within limits. Note that this algorithm may fetch revisions twice, but only when a timeslice needs to be split. The flow is: first fetch revisions; if they fit within the timeslice, fetch their scores. Otherwise, split the timeslice and fetch revisions again for each new timeslice.

### Changes by file

- `lib/timeslice_manager.rb`: a new `maybe_create_course_wiki_timeslice` method was introduced to ensure the existence of a single CWT. This is needed because splitting a timeslice requires deleting the existing timeslice from A to B and creating two new ones: one from A to C and another from C to B. Commit 8925f3aa23c67268d9a67f3ea3c61a32abafe454.

- `lib/revision_data_manager.rb`: small refactor was made to make `fetch_score_data_for_course` method independent of `@articles`. This is necessary because `@articles` is set during `fetch_revision_data_for_course`, and we want to decouple `fetch_score_data_for_course` from it to avoid refetching revisions twice. Commit 0e3a767b8c8a62c80cb307c38ffe80b03d32f94c.

- `lib/course_revision_updater.rb`: deleted `fetch_data_for_course_wiki` method, which used to fetch both revisions and scores, and replaced it with two new methods: `fetch_revisions_for_course_wiki` and `fetch_scores_for_revisions`. The goal is to decouple fetching revisions from fetching scores, so revisions are not refetched twice for every timeslice. Now we first fetch revisions; if they fit within the timeslice, we then fetch their scores. Otherwise, we split the timeslice and fetch revisions again, ensuring refetching only happens when the timeslice is actually split. Commits 

- `app/services/update_course_wiki_timeslices.rb`: the new recursive `handle_timeslice` method implements the adaptive timeslice splitting strategy through the `SplitTimeslices` class.

- `app/services/split_timeslice.rb`: this new file contains the logic for splitting timeslices. It determines whether a timeslice needs to be split for a given wiki and date range, ensuring unnecessary timeslices are deleted, and necessary ones are created.

I chose to use fixtures to compare timeslices to make it easier to test all relevant fields.


spec/services/update_course_stats_spec.rb
```diff
@@ -173,7 +173,7 @@
       end
       sentry_tag_uuid = subject.sentry_tag_uuid
       # one error for each timeslice that tried to update
-      expect(course.flags['update_logs'][1]['error_count']).to eq 5
+      expect(course.flags['update_logs'][1]['error_count']).to eq 10
```
gabina: Note that now we fetch revisions twice: one to count if it fits in the timeslice, and another one with scores.

gabina:

ragesoss:
Thanks! I'll review this next week.

app/services/split_timeslices.rb
```diff
@@ -0,0 +1,167 @@
...
+  private
+
+  def too_many_revisions?(wiki)
+    @revisions[wiki][:revisions].count(&:scoped) > REVISION_THRESHOLD
+  end
+
+  # A timeslice is not split if its duration is an odd number of seconds.
+  # This is because the revisions API expects start and end times in
+  # YYYY-MM-DD HH:MM:SS format, which does not allow fractions of a second.
+  def splittable?(start_date, end_date)
```
ragesoss: This odd restriction seems... odd. Are there situations where a default-sized timeslice is not splittable because of this? It seems like we could still split a timeslice with odd seconds, just by putting the odd second into one or the other arbitrarily.

spec/services/update_course_stats_spec.rb
ragesoss: Can you think of a good way to avoid this? Splitting timeslices should be relatively rare, so it would be nice if we only had to refetch in cases where a split is needed.

app/services/split_timeslice.rb
```diff
@@ -0,0 +1,167 @@
...
+require_dependency "#{Rails.root}/lib/timeslice_cleaner"
+require_dependency "#{Rails.root}/lib/data_cycle/update_debugger"
+require_dependency "#{Rails.root}/lib/revision_scanner"
+
+#= Implements the adaptive timeslice splitting strategy.
+# The algorithm starts with a default timeslice, fetches revisions for that period
+# and, if the timeslice exceeds the threshold, recursively splits it until all
+# timeslices are within limits.
+class SplitTimeslices
+  def initialize(course, update_service: nil)
```
ragesoss: How practical would it be to have a `SplitTimeslice` service that acts on a single timeslice (perhaps by extracting logic from this, or perhaps to be called directly when we find a timeslice that needs to be split)?

I think it would be easier to understand and maintain if the logic for handling a single too-large timeslice was isolated to its own class. (But I'm okay with this current approach if you think it's better.)

app/controllers/courses_controller.rb
```diff
@@ -231,7 +231,7 @@ def tag
   def manual_update
     require_super_admin_permissions
     @course = find_course_by_slug(params[:id])
-    UpdateCourseStatsTimeslice.new(@course)
+    UpdateCourseStats.new(@course)
```
ragesoss, line 234: I'm glad you are making this change. If it's not too much work, this would be better done in a separate PR, but if it's too tedious, I'm okay with keeping it combined with the splitting PR.

lib/course_revision_updater.rb
```diff
@@ -80,13 +94,14 @@ def new_revisions?(revisions, wiki, timeslice_start)
...
                                    .first
     if timeslice.nil?
-      # This scenario is unexpected, so we log the message to understand why this happens.
+      # This can occur while a timeslice is being split and hasn’t been created yet.
+      # In that case, we treat it as new data only if at least one revision is fetched.
       Sentry.capture_message 'No timeslice found for revision date',
                              level: 'warning',
                              extra: { course_name: @course.slug,
                                        wiki: wiki.id,
                                        date: timeslice_start }
```
gabina, line 103: I’d keep this Sentry message for about a week to monitor when timeslices get split in production. It shouldn’t generate a large amount of logs.

app/services/update_course_wiki_timeslices.rb
```diff
@@ -56,139 +54,48 @@ def fetch_data_and_process_timeslices_for_every_wiki(all_time)
...
-      log_processing(wiki, start_date, end_date)
-
-      fetch_data(wiki, start_date, end_date, only_new: true)
-      # Only process timeslices if there is new data
-      process_timeslices(wiki) if new_data?(wiki)
-      @processed_timeslices_count += 1
+      next if timeslice_reprocessed?(wiki.id, t.start)
+
+      ActiveRecord::Base.transaction do
+        processed = @splitter.handle(wiki, t.start, t.end, only_new: true)
```
ragesoss: I think `UpdateCourseWikiTimeslices` and `SplitTimeslice` are too tightly coupled. Ideally, I think `SplitTimeslice` would only be responsible for splitting a timeslice, not for fetching, calculating and storing stats, or updating other records. Is that practical?

gabina: Yup, I agree with your feedback. I tried to decouple `SplitTimeslice` from `UpdateCourseWikiTimeslices` in commit [a7c11d2](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6461/commits/a7c11d2ade6f1adf8ee0e1459ba70697a8e9304b) . Now `SplitTimeslice` is only responsible for determining whether a timeslice needs to be split for a given wiki and date range, ensuring unnecessary timeslices are deleted, and necessary ones are created. Data fetching and timeslice processing remain the responsibility of `UpdateCourseWikiTimeslices`. Note that the recursive algorithm now resides in `UpdateCourseWikiTimeslices`, since determining when to stop splitting requires fetching the revisions first
If you're OK with these changes, then I'll fix the specs for both classes.

ragesoss: Nice. This helps a lot. Go ahead with the specs.

gabina: The specs for class `UpdateCourseWikiTimeslices` could be improved in the future, as there are too many and perhaps some could be deleted. But I think we're ready to go now.

lib/revision_data_manager.rb
```diff
@@ -175,8 +175,11 @@ def create_revision(rev_data, article_data, users, articles)
   # only_scoped_articles_course? set to true).
   # Returns [scoped_revisions_in_spaces, non_scoped_revisions_or_out_spaces]
   def partition_revisions
+    articles = Article.where(wiki_id: @wiki.id, deleted: false,
```
ragesoss, line 178: I think it's worth clarifying in comments at the top about how this class is supposed to be used, in terms of the life cycle of a single instance, and how @revisions and @articles are used over that lifecycle (which may involve calls to each of the three non-private instance methods). It's confusing that one method fetches data to populate @revisions, and another takes revisions as an argument and sets @revisions based on that. Similarly, it's not clear from the code what the difference is between @articles and articles right here. (You addressed the reason for this change in the PR description, but perhaps more specific variable names help.)

PR #6466, opened 2025-09-29
Mark timeslices as needs update if rolling back

gabina:
## What this PR does
This PR starts from #6461. Only the last two commits are relevant to this PR:

- 3248978afe2adc27893a9ceaef02285491005511
- 9cbe98a521a30550a1edf14c6adeb9f8e0b06a1d

The basic idea is to mark timeslices as `needs_update` if something goes wrong during the update and the db transaction needs to be roll-backed. 

After analyzing `Mysql2::Error: Field 'article_id' doesn't have a default value` sentry [log](https://wiki-education.sentry.io/issues/6382949119/events/?project=6269916&query=&referrer=issue-stream), I noticed that we need to explicitly mark a failed timeslice as `needs_update`, otherwise it may not be retried for processing.

The improvement on the log (including start and end dates) would also help to better understand the error and reproduce it.

### DIFF
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -61,12 +61,14 @@ class UpdateCourseWikiTimeslices
       # If the timeslice was reprocessed in this update, then skip it
       next if timeslice_reprocessed?(wiki.id, t.start)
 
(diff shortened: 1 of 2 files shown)


gabina:
@ragesoss it's ready for review

Issue #6468, opened 2025-10-04
Slack call to action doesn't work if user is signed into multiple gmail accounts (firefox)


trevorhauter:
### What is happening?
When a user is signed into more than 1 gmail account on firefox and clicks on the slack call to action in the githubs readme, it downloads a random file and doesn't open gmail.

### To Reproduce
1. Open firefox
2. Be authenticated on multiple gmail accounts
3. Go to https://github.com/WikiEducationFoundation/WikiEduDashboard?tab=readme-ov-file#contribute
4. Click on "Join slack now"!
5. You should observe a random file being downloaded

### Expected behavior
I expected my gmail to open up with a prompt to email Sage Ross to request access to Slack.

### System Information

 - OS: macOS
 - Browser Firefox
 - Version 143.0.3

trevorhauter:
Resolved in https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6469

PR #6469, opened 2025-10-04
Update slack CTA from gmail specific to generic mailto link

trevorhauter:
## What this PR does
- Issue - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6468
- Updates the slack call to action link to be a generic `mailto` link as opposed to a gmail specific link


trevorhauter:
Actually, I'm realizing this should be a `mailto`. Not everybody uses gmail. `mailto` is more dynamic and seems like a better solution to me, but I'm happy to be wrong.

PR #6473, opened 2025-10-08
Enable additional javascript style linting rules (Enable react/jsx-equals-spacing ESLint rule and fix violations)

Vanessa082:
## What this PR does
This PR is a partial Fix to issue #1408. It enables the previously disabled ESLint rule `react/jsx-equals-spacing` and fixes the code violation in the code base.
## Screenshots
Before:
image

After:
Screenshot from 2025-10-08 18-37-44


Vanessa082:
Hi @ragesoss, I have some build errors here that I do not seem to understand. Please, I would appreciate your guidance and explanation so as to work on it.

gabina:
The CI failure isn’t related to your changes , it looks like we have some broken specs. JS test suite and linting passed, so your PR should be fine from a CI perspective.

Vanessa082:
> The CI failure isn’t related to your changes , it looks like we have some broken specs. JS test suite and linting passed, so your PR should be fine from a CI perspective.

Ok thank you @gabina that means I can continue working 😃 
You do have much understanding here I will really love to understand what is going on here.
I guess with time I will

ragesoss:
This looks safe to me! Thanks!

PR #6474, opened 2025-10-08
fix: enabled and add explicit type for all buttons (react/button-has-type)

Vanessa082:
## What this PR does
This  pull request partially resolves issue #1408 by enabling and setting an explicit type for every button.


Abishekcs:
@Vanessa082, would it be possible to break this PR into smaller PRs?

ragesoss:
This touches too many files, and it's a very low priority. I don't want to review or merge it at this time, since I'm not completely sure that this won't have effects for users. I merged your other linting PR @Vanessa082. I suggest moving on to something different next.

PR #6480, opened 2025-10-10
Enhance Year Requirement in Term Field for Course Creation and Cloning Contd

oree-xx:
This PR is a continuation of the work on this PR [6344](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6344). This PR enhances the term field validation for course creation and cloning to require a 4-digit year and improves the user experience by validating only whenusers leave the field.

  Implemented the following changes:
  1. Enhanced Year Validation
    - Changed validation regex from /\d{2,}/ to /\d{4}/ to require exactly 4 digits
    - Ensures terms like "Spring 2025" are valid, but "Spring 25" are not
  2. Improved Validation Timing
    - Added validateOnBlur prop to InputHOC to control when validation occurs
    - Term fields now validate only when users leave the field, not while typing
    - Prevents distracting error messages during input
  3. Added Next Button Validation
    - Added term validation to titleSubjectAndDescriptionAreValid() in CourseCreator
    - Users cannot proceed to the next step without a valid term
    - Prevents the issue where validation only occurred at final course creation
    
    
    Before:

https://github.com/user-attachments/assets/7da5bb9d-ec83-4209-9da4-5049bfa4b7a7

After: 

https://github.com/user-attachments/assets/1385e5ff-89b6-4233-9c24-71d4a60eb7e0

### DIFF
--- a/app/assets/javascripts/components/course_creator/course_form.jsx
+++ b/app/assets/javascripts/components/course_creator/course_form.jsx
@@ -237,6 +237,7 @@ const CourseForm = (props) => {
 
   return (
     <div className={props.courseFormClass}>
+      {props.firstErrorMessage && <div className="warning"><p style={{ color: '#7c1c1c' }}>{props.firstErrorMessage}</p></div>}
       <div className="column">
         {campaign}
         <TextInput
@@ -277,8 +278,6 @@ const CourseForm = (props) => {
           {backOrCancelButton}
           <p className="tempEduCourseIdText">
             {props.tempCourseId || '\xa0'}
-          &nbsp;
-            <span className="red">{props.firstErrorMessage || '\xa0'}</span>
           </p>
         </div>
 


app/assets/javascripts/components/course_creator/course_form.jsx
```diff
@@ -41,6 +41,22 @@ const CourseForm = (props) => {
     }
   }, []);
 
+  const termValidation = (value) => {
```
Abishekcs, line 44: Requires 2-space indentation style and also align the closing braces

Abishekcs:
LGTM!
I’ve tested it out and it works properly.

@oree-xx it would be great if you could add a before and after screenshot or a short screen recording.

oree-xx:
> LGTM! I’ve tested it out and it works properly.
> 
> @oree-xx it would be great if you could add a before and after screenshot or a short screen recording.

Thank youuu for the review @Abishekcs, will add the screen recording

config/locales/en.yml
```diff
@@ -617,6 +617,7 @@ en:
         to re-use, you can clone a copy of it with new dates. You can also
         use the clone option to create additional sections for a course you
         already set up.
+      next: Next
```
ragesoss: We have a few different 'Next' buttons, and it's likely to be button label that comes up again. I suggest adding this one with a less specific message key, so that it can serve as a general 'Next' button rather than one that is specific to one component.

oree-xx: I have updated that on my latest commit

oree-xx:
> > LGTM! I’ve tested it out and it works properly.
> > @oree-xx it would be great if you could add a before and after screenshot or a short screen recording.
> 
> Thank youuu for the review @Abishekcs, will add the screen recording

@Abishekcs I just added a screen recording of the before and after, kindly review, thank youuu✨

ragesoss:
The video looks good! One thing I notice is that the validation error message is broken across two lines. Can you find an easy way to make that look better? (If not, that would be a good thing to open a new issue for.)

Abishekcs:
> Can you find an easy way to make that look better?  (If not, that would be a good thing to open a new issue for.)

Just wanted to add something if the UI can be same as the course  homepage that will be great too i think as shown below
screenshot-2025-10-14_21-46-19

oree-xx:
> > Can you find an easy way to make that look better?  (If not, that would be a good thing to open a new issue for.)
> 
> Just wanted to add something if the UI can be same as the course homepage that will be great too i think as shown below 
CleanShot 2025-10-15 at 16 36 55@2x
implemented this in my last commit
@ragesoss @Abishekcs

ragesoss:
Great! Can you update the video to show the current version?

oree-xx:
> Great! Can you update the video to show the current version?

Updated the video. @ragesoss

oree-xx:
> > Great! Can you update the video to show the current version?
> 
> Updated the video. @ragesoss

@ragesoss is there anything else i need to do?

ragesoss:
This is solid! I'm going to merge it now.

One possible improvement for a followup PR would be to find a good way to avoid having the form change size (and change the position of the inputs) when the validation warning is added and removed.

oree-xx:
> This is solid! I'm going to merge it now.
> 
> One possible improvement for a followup PR would be to find a good way to avoid having the form change size (and change the position of the inputs) when the validation warning is added and removed.

Okay will look into that, thank youu

PR #6481, opened 2025-10-10
Add basic AI alert stats admin view

gabina:
## What this PR does
This PR adds a basic admin view to show some aggregated stats for AI edit alerts:
- General stats: number of students with multiple alerts (ie, unique non-nil values of Alert#prior_alert_id_for_user for the current-term alerts), number of same-page alerts (ie, unique non-nil values of alert#prior_alert_id_for_page), and number of alerts with feedback.
- Breakdown by page.
- Recently-submitted feedback
- Recent mainspace alerts (or unknown, which is currently how all mainspace pages will be classified, but there might be others included as well).
- Recent alerts with a `prior_alert_id_for_user`

## Screenshots
Before:
No admin view

After:
[Screencast from 2025-10-28 23-47-49.webm](https://github.com/user-attachments/assets/6446ba07-23e3-4455-adf4-7e2ef04c6331)

No sorting functionality is implemented.

### DIFF
--- a/app/views/admin/index.html.haml
+++ b/app/views/admin/index.html.haml
@@ -16,6 +16,8 @@
     %li
       %a{href: '/faq/new'} Add new FAQ
     %li
+      %a{href: '/ai_edit_alerts_stats'} AI alerts stats
+    %li
       %a{href: '/article_finder'} Article Finder
     %li
       %a{href: '/copy_course'} Copy course from another server


gabina:
@ragesoss this is the first version of this PR. I plan to make some implementation updates (a few renamings and code adjustments). But before that, I’d appreciate any feedback and suggestions on the aggregated metrics you’d like to see.

ragesoss:
The main way we are going to want to look at these is by term, so I think as a first implementation, the stats should be scoped to alerts from the courses in `Campaign.default_campaign` (and not by time period). Then the breakdown by page type will work fine as-is.

Other keys stats that we care about are: number of students with multiple alerts (ie, unique non-nil values of `Alert#prior_alert_id_for_user` for the current-term alerts), number of same-page alerts (ie, unique non-nil values of `alert#prior_alert_id_for_page`), and number of alerts with feedback.

ragesoss:
The next priorities (not necessarily in this order) are:

a) graphs to show alert trends over time (perhaps similar to https://impact-visualizer.wmcloud.org/topics/2#wp10 where different page types are represented by different colors, and the color band expands as alerts for that page type are created)

b) tables for listing the most recent alerts that might need special attention: recently-submitted feedback, recent mainspace alerts (or `unknown`, which is currently how all mainspace pages will be classified, but there might be others included as well), and recent alerts with a `prior_alert_id_for_user`. Such tables should have relevant links for each alert: the single-alert view, the course, the user, the revision diff link, and the pangram link.

gabina:
One followup question.

> tables for listing the most recent alerts that might need special attention: recently-submitted feedback

I don't think we have a time specific to when a followup was completed other than the `updated_at` alert field. Is that right? It looks like it's safe to rely on that field, since I don't think there is any other way we update the alert after creation. But it's something to take into account if at some point we add any other feature/data to the alert.

ragesoss:
good point. adding a slot for feedback timestamp when saving a feedback response will be a good idea.

gabina:
I updated the PR description and screenshots to only address b) in this PR. The graphs to show alert trends over time will be in a different PR.

gabina:
Failing specs seem transient. I think this is ready for a review @ragesoss 
```
2068 examples, 9 failures, 24 pending

Failed examples:

rspec ./spec/lib/importers/average_views_importer_spec.rb:25 # AverageViewsImporter.update_average_views saves average page views on ArticlesCourses records
rspec ./spec/lib/importers/average_views_importer_spec.rb:58 # AverageViewsImporter.update_outdated_average_views does not update recently-updated records
rspec ./spec/lib/wiki_pageviews_spec.rb:19 # WikiPageviews.average_views_from_date for a popular article returns the average page views
rspec ./spec/lib/wiki_pageviews_spec.rb:29 # WikiPageviews.average_views_from_date for an article with a slash in the title returns the average page views
rspec ./spec/lib/wiki_pageviews_spec.rb:39 # WikiPageviews.average_views_from_date for an article with an apostrophe in the title returns the average page views
rspec ./spec/lib/wiki_pageviews_spec.rb:49 # WikiPageviews.average_views_from_date for an article with quote marks in the title returns the average page views
rspec ./spec/lib/wiki_pageviews_spec.rb:59 # WikiPageviews.average_views_from_date for an article with unicode characters in the title returns the average page views
rspec ./spec/lib/wiki_pageviews_spec.rb:72 # WikiPageviews.average_views_from_date for a wikidata item returns the average page views
rspec ./spec/lib/wiki_pageviews_spec.rb:82 # WikiPageviews.average_views_from_date for an article that does not exist returns 0
```

PR #6485, opened 2025-10-11
Fix early `average view` updates for articles

### LINKED ISSUE #6476: Courses have 0 article views for 14 days
### What is happening?
It looks like a course may have 0 articles views for 2 weeks, when maybe its articles did receive views.

### Example
Course 19324 [Del_Mar_College/Texas_Government_Sec._A00_(Fall_2025)](https://dashboard.wikiedu.org/courses/Del_Mar_College/Texas_Government_Sec._A00_(Fall_2025)/home).  Start: 2025-08-26. End: 2025-12-10.

It has 7 articles edited by now, all of them with 0 average views.

```
+------------+----------------------------+---------------+--------------------------+
| article_id | first_revision             | average_views | average_views_updated_at |
+------------+----------------------------+---------------+--------------------------+
|   39206022 | 2025-10-08 00:33:41.000000 |             0 | 2025-10-08               |
|   50779985 | 2025-10-06 03:27:14.000000 |             0 | 2025-10-06               |
|   51046965 | 2025-10-08 16:02:55.000000 |             0 | 2025-10-08               |
|   51395971 | 2025-10-08 02:08:57.000000 |             0 | 2025-10-08               |
|   51637151 | 2025-10-05 02:52:36.000000 |             0 | 2025-10-05               |
|  974436086 | 2025-09-27 16:56:57.000000 |             0 | 2025-09-27               |
|  977409360 | 2025-10-08 14:26:24.000000 |             0 | 2025-10-08               |
+------------+----------------------------+---------------+--------------------------+
```

I think the issue is that the average views are being updated on the same day as the first revision. Because of that, when we call `average_views_from_date`, it returns 0 (see the `views_from_date` definition). Since we also update the `average_views_updated_at` field, we have to wait 2 weeks (see `DAYS_UNTIL_OUTDATED = 14`) before updating `average_views` again.

### Expected behavior
We should avoid having to wait 2 weeks to see article views. One option could be to skip updating the `average_views_updated_at` field when `calculate_average_views` returns 0 only because not enough time has passed since the `first_revision`.

### Screenshots
Image

Solobarine:
## What this PR does
This PR fixes the issue with courses having 0 views when their articles may have had views. This fix skips updating the `average_views_updated_at` when the `views_from_date` is 0. This makes it possible for views to be updated when they happen.

Closes #6476

### DIFF
--- a/lib/importers/average_views_importer.rb
+++ b/lib/importers/average_views_importer.rb
@@ -42,13 +42,12 @@ class AverageViewsImporter
   def self.update_average_views_for_article(article_course, average_views, time)
     can_update = (1.day.ago - article_course.first_revision) > 1.day
 
+    # If not enough time has passed since first revision, return
     return unless can_update
 
-    views_since_revision = WikiPageviews.new(article_course.article)
-                                        .average_views_from_date(article_course.first_revision)
-
     average_views[article_course.id] = {
-      average_views: views_since_revision,
+      average_views: WikiPageviews.new(article_course.article)
+                                  .average_views_from_date(article_course.first_revision),
       average_views_updated_at: time
     }
   end
--- a/spec/lib/importers/average_views_importer_spec.rb
+++ b/spec/lib/importers/average_views_importer_spec.rb
@@ -14,8 +14,8 @@ describe AverageViewsImporter do
   end
 
   let(:article_course_with_insufficient_rev_time) do
-    create(:articles_course, course:, article_id: article4.id, average_views: 10,
-           average_views_updated_at: 1.day.ago, first_revision: 1.day.ago)
+    create(:articles_course, course:, article_id: article4.id, average_views: nil,
+           average_views_updated_at: nil, first_revision: 1.day.ago)
   end
 
   before do
@@ -40,7 +40,7 @@ describe AverageViewsImporter do
       VCR.use_cassette 'average_views' do
         described_class.update_average_views(ArticlesCourses.all)
       end
-      expect(article_course_with_insufficient_rev_time.reload.average_views).to eq(10)
+      expect(article_course_with_insufficient_rev_time.reload.average_views).to eq(nil)
     end
   end
 


Abishekcs:
@Solobarine is the PR ready to be reviewed?

Solobarine:
Not yet, some tests are failing and they are a result of the changes I made.

Abishekcs:
okay

gabina:
I'd like to review this when it's ready. Just ping me

Solobarine:
Hi @gabina, the pull request is now available for review. I was waiting for the CI results.

lib/importers/average_views_importer.rb
```diff
@@ -40,9 +40,14 @@ def self.update_average_views(articles_courses)
...
+    views_since_revision = WikiPageviews.new(article_course.article)
+                                        .average_views_from_date(article_course.first_revision)
+
+    # Only update if there are views
+    return unless views_since_revision.positive?
+
     average_views[article_course.id] = {
-      average_views: WikiPageviews.new(article_course.article)
-                                  .average_views_from_date(article_course.first_revision),
+      average_views: views_since_revision,
```
gabina: As your PR description says, this strategy skips updating `average_views_updated_at` when `views_since_revision` is 0. Note that the average views being 0 is not necessarily wrong. There are two possible cases for that:

1. The article actually received 0 views.
2. Not enough time has passed since the first revision.

While case 1 is quite unlikely, I think we should only skip updating `average_views_updated_at` for case 2. One option could be to avoid calling `average_views_from_date` if not enough time has passed since the first revision, or to call `average_views_from_date` but return something that indicates that not enough time has passed.

spec/lib/importers/average_views_importer_spec.rb
```diff
@@ -61,7 +61,7 @@
       end
       expect(articles_course.reload.average_views).to eq(1)
       expect(article_course_never_updated.reload.average_views).to be > 1300
-      expect(article_course_very_old.reload.average_views).to eq(0)
+      expect(article_course_very_old.reload.average_views).to eq(1)
```
gabina: This spec is testing that _update_outdated_average_views does not update recently-updated records_. That should remain true even with your changes. Why did you need to update this?

Solobarine: Due to the change I made, the average_views for the `article_course` is never updated hence, the average_views remains 1.

lib/importers/average_views_importer.rb
Solobarine: Ok, that makes sense. So what minimum duration from the first revision should be considered enough. I'm thinking 3 or 5 days.

gabina: The only problem is when `start_date` > `end_date`. That means that "not enough time has passed"
```
  def views_from_date(start_date)
    end_date = 1.day.ago
    # Do not query views for an empty period of time
    return no_results if start_date.nil? || start_date > end_date
    url = query_url(start_date:, end_date:)
    parse_results(api_get(url))
  end
```

3 or 5 days are definitely enough time to calculate article views.

spec/lib/importers/average_views_importer_spec.rb
gabina: hmm, I see. But it should be updated, since those article views are already outdated (see `DAYS_UNTIL_OUTDATED = 14`). The only case when we don’t want to update them is if not enough time has passed since the first revision.

Solobarine: I resolved this in this current commit.

lib/importers/average_views_importer.rb
Solobarine: The issue with  implementing the condtion in `views_from_date` is that this function only retrieves  `views_data` and further modfication will detract from its purpose.

What I did was to add a time-based condition in the `self.update_average_views_for_article` function to serve as a filter for updating `articles_course` so `views_from_date` can maintain it's simplicity and purpose.

lib/importers/average_views_importer.rb
```diff
@@ -40,9 +40,15 @@ def self.update_average_views(articles_courses)
   end
 
   def self.update_average_views_for_article(article_course, average_views, time)
+    can_update = (Time.zone.now - article_course.first_revision) > 3.days
+
+    return unless can_update
+
+    views_since_revision = WikiPageviews.new(article_course.article)
+                                        .average_views_from_date(article_course.first_revision)
+
```
gabina: I think this strategy makes sense. However, I would define the time-based condition as:

`can_update = (1.day.ago - article_course.first_revision) > 1.day`

to stay consistent with what `views_from_date` currently does. If you look at that method, it queries views for the period [first_revision, 1.day.ago]. Once the difference between those dates is greater than one day, it means they’re actually different days and we’re querying views for a non-empty period.

Does it sound good to you?

spec/lib/importers/average_views_importer_spec.rb
```diff
@@ -27,6 +34,7 @@
         described_class.update_average_views(ArticlesCourses.all)
       end
       expect(articles_course.reload.average_views).to be > 50
+      expect(article_course_with_insufficient_rev_time.reload.average_views).to eq(10)
```
gabina: What are you trying to test here?

I understand that you’re trying to check that if not enough time has passed since the first revision, the `average_views` and `average_views_updated_at` fields are not updated.

In that case, I think it’s better to create another example case. Note that this example is ‘.update_average_views saves average page views on ArticlesCourses records’, but you want to test a different case: that it doesn’t save average page views if not enough time has passed since the first revision.

lib/importers/average_views_importer.rb
Solobarine: Yes.

spec/lib/importers/average_views_importer_spec.rb
Solobarine: True. I have created a separate test block for clarity

Solobarine:
Hello @gabina,

I've implemented the changes and it's ready for review.

lib/importers/average_views_importer.rb
```diff
@@ -40,9 +40,15 @@ def self.update_average_views(articles_courses)
   end
 
   def self.update_average_views_for_article(article_course, average_views, time)
+    can_update = (1.day.ago - article_course.first_revision) > 1.day
+
+    return unless can_update
```
gabina, line 46: It would be nice to add a small comment explaining why we may want to return prematurely

lib/importers/average_views_importer.rb
```diff
@@ -40,9 +40,15 @@ def self.update_average_views(articles_courses)
...
+
+    views_since_revision = WikiPageviews.new(article_course.article)
+                                        .average_views_from_date(article_course.first_revision)
+
     average_views[article_course.id] = {
-      average_views: WikiPageviews.new(article_course.article)
-                                  .average_views_from_date(article_course.first_revision),
+      average_views: views_since_revision,
```
gabina: Is there a specific reason why you extracted the average views into a `views_since_revision` variable? If not, I would revert the change, as it's always best to keep changes to a minimum in any PR.

spec/lib/importers/average_views_importer_spec.rb
```diff
@@ -28,6 +35,13 @@
       end
       expect(articles_course.reload.average_views).to be > 50
     end
+
+    it 'does not save average views when insufficient time has passed since first revision' do
+      VCR.use_cassette 'average_views' do
+        described_class.update_average_views(ArticlesCourses.all)
+      end
+      expect(article_course_with_insufficient_rev_time.reload.average_views).to eq(10)
```
gabina: The spec looks good! But in what case do you think an article course would have `average_views` set to 10 and a non-null `average_views_updated_at`, but not enough time has passed since the first revision? I think a more obvious case is when both `average_views` and `average_views_updated_at` are null. WDYT?

gabina (commented):
Nice work. I think we're pretty close!

spec/lib/importers/average_views_importer_spec.rb
Solobarine: True, I did not have a specific case in mind. I'll set those values to null.

Solobarine:
I resolved them. Is there anything you would like me to add?

gabina (approved):
Thanks! LGTM

PR #6487, opened 2025-10-13
Fix i18n warnings in CourseOresPlot component

oree-xx:
This PR resolves all i18n linting warnings in the `CourseOresPlot` component by
  replacing hardcoded strings with proper internationalization keys.

  ## Changes Made

  ### 1. Added translation keys to `config/locales/en.yml`:
  - `ores_plot_description` - Main graph description text
  - `ores_plot_description_link` - ORES FAQ URL
  - `ores_plot_no_data` - Message when no data is available
  - `ores_plot_refresh_data` - Refresh button text
  - `ores_plot_show_button` - Main button text

  ### 2. Updated `course_ores_plot.jsx`:
  - Replaced all hardcoded strings with `I18n.t()` calls
  - Handled HTML content by splitting the description text around "(ORES)" to
  properly insert the link
  - Maintained all existing functionality while making the component fully
  internationalizable

  ## Testing

  - [x] All eslint i18n warnings resolved (7 warnings fixed)
  - [x] Component renders correctly with translated text
  - [x] Links and click handlers remain functional

  ## Before
  WARNING in [eslint]
  /app/assets/javascripts/components/articles/course_ores_plot.jsx
     66:12   warning  Use I18n over string literals for localization
     68:84   warning  Use I18n over string literals for localization
     ... (7 warnings total)

  ## After
  ✅ No i18n warnings

### DIFF
--- a/app/assets/javascripts/components/articles/course_ores_plot.jsx
+++ b/app/assets/javascripts/components/articles/course_ores_plot.jsx
@@ -66,7 +66,7 @@ const CourseOresPlot = ({ course }) => {
         <p
           dangerouslySetInnerHTML={{
           __html: I18n.t('courses.ores_plot_description', {
-            ores_link: `<a href="${I18n.t('courses.ores_plot_description_link')}" target="_blank">ORES</a>`,
+            ores_link: '<a href="https://www.mediawiki.org/wiki/ORES/FAQ" target="_blank">ORES</a>',
             refresh_link: `<a href="#" class="refresh-link">${I18n.t('courses.ores_plot_refresh_data')}</a>`
           })
         }} onClick={(e) => {
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -782,7 +782,6 @@ en:
       rating is based on a machine learning project (%{ores_link})
       that estimates an article's quality rating based on the amount of
       prose, the number of wikilinks, images and section headers, and other features. (%{refresh_link})
-    ores_plot_description_link: "https://www.mediawiki.org/wiki/ORES/FAQ"
     ores_plot_no_data: No Structural Completeness data available
     ores_plot_refresh_data: Refresh Cached Data
     ores_plot_show_button: Change in Structural Completeness


app/assets/javascripts/components/articles/course_ores_plot.jsx
```diff
@@ -64,11 +64,10 @@ const CourseOresPlot = ({ course }) => {
       <div className="ores-plot">
         <CourseQualityProgressGraph graphid={'vega-graph-ores-plot'} graphWidth={1000} graphHeight={200} articleData={data} />
         <p>
-          This graph visualizes, in aggregate, how much articles developed from
-          before students first edited them until their most recent edits. The <em>Structural Completeness </em>
-          rating is based on a machine learning project (<a href="https://www.mediawiki.org/wiki/ORES/FAQ" target="_blank">ORES</a>)
-          that estimates an article&apos;s quality rating based on the amount of
-          prose, the number of wikilinks, images and section headers, and other features. (<a href="#" onClick={refreshHandler}>Refresh Cached Data</a>)
+          {I18n.t('courses.ores_plot_description').split('(ORES)')[0]}
```
ragesoss: This is not a safe way to handle an I18n message, as we don't know that the substring ORES will be present in every translation. Better to pass in the link in as a parameter to the message or find some other way to handle it.

oree-xx: Thanks for the review @ragesoss , i fixed it by passing the link as a parameter in my latest commit

config/locales/en.yml
```diff
@@ -776,6 +776,16 @@ en:
     online_volunteers_enabled: Online Volunteers enabled
     online_volunteers_tooltip: Enabling Online Volunteers allows user editors to join as a volunteer. Their edits will not be tracked.
     ores_plot: ORES
+    ores_plot_description: >
+      This graph visualizes, in aggregate, how much articles developed from
+      before students first edited them until their most recent edits. The Structural Completeness
+      rating is based on a machine learning project (%{ores_link})
+      that estimates an article's quality rating based on the amount of
+      prose, the number of wikilinks, images and section headers, and other features. (%{refresh_link})
+    ores_plot_description_link: "https://www.mediawiki.org/wiki/ORES/FAQ"
```
ragesoss: This link is better hard-coded, as we don't want translators to have to try to figure out how to translate it.

app/assets/javascripts/components/articles/course_ores_plot.jsx
```diff
@@ -63,13 +63,19 @@ const CourseOresPlot = ({ course }) => {
...
         <CourseQualityProgressGraph graphid={'vega-graph-ores-plot'} graphWidth={1000} graphHeight={200} articleData={data} />
-        <p>
-          This graph visualizes, in aggregate, how much articles developed from
-          before students first edited them until their most recent edits. The <em>Structural Completeness </em>
-          rating is based on a machine learning project (<a href="https://www.mediawiki.org/wiki/ORES/FAQ" target="_blank">ORES</a>)
-          that estimates an article&apos;s quality rating based on the amount of
-          prose, the number of wikilinks, images and section headers, and other features. (<a href="#" onClick={refreshHandler}>Refresh Cached Data</a>)
-        </p>
+        <p
+          dangerouslySetInnerHTML={{
```
ragesoss, line 67: This seems reasonable. I wish it was easier to understand, but I don't know of a better way to combine a link with an interface message in React.

ragesoss:
Thanks! Can you include a screenshot from the new version, so we can see that it's formatted well and doesn't change the UI?

app/assets/javascripts/components/articles/course_ores_plot.jsx
oree-xx: Yeah, I felt the same way. Initially, I used dangerouslySetInnerHTML since i18n-js only returns plain strings, but after rethinking it, a clearer option would be to split the translation like this:

```suggestion
          <p>
  {I18n.t('courses.ores_plot_description_intro')}
  {' '}
  (<a href="https://www.mediawiki.org/wiki/ORES/FAQ" target="_blank">{I18n.t('courses.ores_plot_link_text')}</a>)
  {' '}
  {I18n.t('courses.ores_plot_description_middle')}
  {' '}
  (<a href="#" onClick={refreshHandler}>{I18n.t('courses.ores_plot_refresh_data')}</a>)
</p>

```
```
 ores_plot_description_intro: >
      This graph visualizes, in aggregate, how much articles developed from
      before students first edited them until their most recent edits. The Structural Completeness
      rating is based on a machine learning project
    ores_plot_description_middle: >
      that estimates an article's quality rating based on the amount of
      prose, the number of wikilinks, images and section headers, and other features.
    ores_plot_link_text: ORES
```
This keeps it React-friendly and avoids HTML injection. What do you think?

ragesoss: Splitting the messages is not a good option, because it makes things very difficult for translators.

oree-xx:
> Thanks! Can you include a screenshot from the new version, so we can see that it's formatted well and doesn't change the UI?
CleanShot 2025-10-13 at 23 47 46@2x

app/assets/javascripts/components/articles/course_ores_plot.jsx
oree-xx: Okay got it!

PR #6484, opened 2025-10-11
Fix: Errors in eslintrc file

Ajay-singh1:
NOTE: Please review the pull request process before opening your first PR: https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/CONTRIBUTING.md#pull-request-process

## What this PR does
< describe the purpose of this pull request and note the issue it addresses >

This PR fixes all the errors thrown by VS Code in .eslintrc file.

## Screenshots
Before:
Screenshot From 2025-10-11 11-57-33

After:
Screenshot From 2025-10-11 11-58-59

## Open questions and concerns
< anything you learned that you want to share, or questions you're wondering about related to this PR >

PR #6488, opened 2025-10-13
Update .eslintrc file for proper comment 

Abdulwahab03:
Description: Replaced invalid ‘#’ comments in JSON file with ‘//’ to fix parsing error”

Problem Statement: The # character was used to add comments within a JSON file. However, the JSON format does not support the use of # for comments. As a result, this caused a “Colon expected” parsing error when the file was read or validated.

What I implemented: I removed all instances of the # character used for commenting and replaced them with //. This change ensures the file remains syntactically valid and can be correctly parsed without errors.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,8 +2,8 @@
   "root": true,
   "extends": "airbnb",
   "rules": {
-    # This makes it easier for Windows users to work on the project. Git automatically
-    # converts between linebreak styles to unix-style should end up in the repo.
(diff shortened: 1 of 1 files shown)


Abishekcs:
@Abdulwahab03 you already open a PR for this changes also one of the changes you made here is already done in a open PR #6484

Abdulwahab03:
Should I just close this pull request?

Abishekcs:
yes

PR #6489, opened 2025-10-14
Enabling js style lint rules

Abdulwahab03:
This pull request address this: #1408 by enabling and setting an explicit type for every button.
I tried to modify this PR #6474 which has too much by breaking it into smaller pieces and without touching many files kindly.


ragesoss:
I don't think we can change the .eslintrc for this rule without fixing all the violations. I'd accept a PR that fixes about this many instances of the violation, but without including the rule change. It should be a single commit, with no unrelated commits or changes except adding the type to buttons.

Abdulwahab03:
thank you @ragesoss, I think the best thing for me to do if I really want to work on this issue is to first close this PR, then fix as many instances of the violation as possible, without including the rule change in .eslintrc. This solution should work I think.

PR #6494, opened 2025-10-16
Create CourseUpdateController to handle course update.

Abdulwahab03:
This PR address this: Refactor CoursesController into multiple controllers #1417 

Problem Statement: CoursesController is rather large, because it is responsible for a bunch of different independent things.

Solution Implemented: I tried to split the "update" method from the CourseController, so that "update" will now have its own controller, by doing this the CourseController will have minimum methods inside it, also "update" method and all its relative private methods will be in a new controller. By doing all these our code will be very easy to maintain.

### DIFF
--- a/app/controllers/courses_update_controller.rb
+++ b/app/controllers/courses_update_controller.rb
@@ -1,0 +1,93 @@
+# frozen_string_literal: true
+
+require "#{Rails.root}/app/workers/update_course_worker"
+require "#{Rails.root}/app/workers/announce_course_worker"
+
+class CoursesUpdateController < ApplicationController
+  include CourseHelper
+  respond_to :html, :json
+
+  before_action :require_permissions, 
+                only: [:update, :validate]
+
+################
+  # CRUD methods #
+  ################
+
+  def update
+    validate
+    handle_course_announcement(@course.instructors.first)
+    slug_from_params if should_set_slug?
+    @course.update update_params
+    update_courses_wikis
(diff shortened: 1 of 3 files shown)


Abdulwahab03:
@ragesoss and @Abishekcs, kindly help me look into this PR.

Abishekcs:
@Abdulwahab03 there are specs in CI build failing due to the changes made in this PR

Abdulwahab03:
rspec error
@ragesoss, I tried running the test locally but failed too, can you kindly help me check it out, I used several hours solving it on my own the problem still persist, you can point out where the error from, then I will take it from there. thank you

Abishekcs:
did you run `bin/rails db:migrate RAILS_ENV=test` as shown in the screenshot ?

Abdulwahab03:
Hi @Abishekcs, thanks so much for your response, yes, I ran it this is what it brought in the picture below, and I don't want to complicate things more.
Error 2

Abishekcs:
But i am confused how where you able to run it before pushing your code. Since this test case was written by yourself `courses_update_controller_spec.rb` ?

Abishekcs:
Try to set `config.active_record.migration_error ` to **false** in `config/envrionments/development.rb`. Then run the test.

ragesoss:
If your test environment has migration-related problems, you can usually just reset that database, or drop and recreate it then run the migrations again.

PR #6496, opened 2025-10-18
Add type to button element in components

Abdulwahab03:
Problem Description:  There are button elements without type=button, and this behavior can change in ways we might not expect especially inside forms.

Solution Implemented: I added type="button" to the button elements in the components.

### DIFF
--- a/app/assets/javascripts/components/admin_notes/notes_panel.jsx
+++ b/app/assets/javascripts/components/admin_notes/notes_panel.jsx
@@ -110,6 +110,7 @@ const NotesPanel = () => {
       <div className="basic-modal">
         {/* Add a close button to the modal */}
         <button
+          type="button"
           onClick={closeModalAndHandleNoteCreation}
           aria-label={I18n.t('notes.admin.aria_label.close_admin')}
           className="pull-right article-viewer-button icon-close admin-focus-highlight"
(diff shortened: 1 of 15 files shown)


Abdulwahab03:
Hello @Abishekcs kindly help me look into this PR this is the Image of the warning I got while running it 
eslint i18 issue
could this be the problem why the build failed"

Abishekcs:
> could this be the problem why the build failed"

I don't think warning will lead to build failing.

ragesoss:
I think at least one of the failing specs is related to the changes here. If adding type results in changing a `link` to a `button` in the HTML, it may require changing the corresponding feature spec (for example by replacing a `click_link` with a `click_button`).

Abdulwahab03:
I will work on this early tomorrow; it's midnight here already  I will check and implement all your suggested changes @ragesoss

Abdulwahab03:
Hello @ragesoss, I tried to implement the change you suggested the last time by replacing click_link with a click_button, but the test still failed,  though I made several chnges in some files that day, but If I can be pointed to the particular change I made that actually introduced the failure I will be glad and take it up from there.

> I think at least one of the failing specs is related to the changes here. If adding type results in changing a `link` to a `button` in the HTML, it may require changing the corresponding feature spec (for example by replacing a `click_link` with a `click_button`).

ragesoss:
I recommend running the failing test locally and looking at the screenshots or disabling headless mode in `spec/rails_helper.rb`. You might be able to see exactly what's causing it to fail.

Issue #6292, opened 2025-04-15
Add option to toggle pageview stats between cumulative views and average views per day


ragesoss:
The Articles tab of a course page includes a column for pageviews, which are intended to represent the cumulative pageviews since the first tracked edit for each article. In some cases, especially when no significant views have accumulated yet, it's more useful to show the average daily views of each article so that you can spot the high-traffic ones even if they were only recently edited.

NOTE: This should not be implemented until after we update the handling of pageviews to be more accurate less susceptible to variation (by storing average daily views on ArticlesCourses records based on fetching total pageviews since first revision, and displaying an estimate based on extrapolating that value over the necessary amount of time).

Formasitchijoh:
@ragesoss  currently starting to work on this, WRT, this notes is it handled yet and is there anything more I need to know if I should want to handle it

> NOTE: This should not be implemented until after we update the handling of pageviews to be more accurate less susceptible to variation (by storing average daily views on ArticlesCourses records based on fetching total pageviews since first revision, and displaying an estimate based on extrapolating that value over the necessary amount of time).

oree-xx:
Hello @Formasitchijoh are you still working on this, want to take a look
cc: @ragesoss

ragesoss:
@oree-xx go for it.

PR #6497, opened 2025-10-20
Add pageview stats toggle functionality

oree-xx:
This PR will add a pageview stats toggle switch to allow users to switch between viewing cumulative pageviews and average daily pageviews for articles #6292 

- Implements Redux action and reducer to manage the display mode state
- Persists user preference to localStorage for consistency across sessions                                               
- Updates Article component to dynamically display pageviews based on selected mode 

Before

CleanShot 2025-10-24 at 10 54 35

After
https://github.com/user-attachments/assets/637f9961-7db2-4ab4-87b9-bdad101045cd

### DIFF
--- a/app/assets/javascripts/components/articles/article_list.jsx
+++ b/app/assets/javascripts/components/articles/article_list.jsx
@@ -158,9 +158,16 @@ const ArticleList = createReactClass({
     }
 
     // Add custom label for view_count column with inline toggle
+    const articlesOrItems = ArticleUtils.articlesOrItems(project);
     keys.view_count.label = (
       <div className="view-column-header">
-        <span>{I18n.t('metrics.view')}</span>
+        <span className="tooltip-trigger">
+          {I18n.t('metrics.view')}
+          <span className="tooltip-indicator-list" />
+          <div className="tooltip dark">
+            <p>{I18n.t(`${articlesOrItems}.view_doc`)}</p>
+          </div>
+        </span>
         <div className="view-mode-toggle">
           <button
             className={`view-mode-btn ${this.props.pageviewDisplayMode === 'cumulative' ? 'active' : ''}`}
@@ -177,6 +184,8 @@ const ArticleList = createReactClass({
         </div>
       </div>
     );
+    // Remove info_key since we're manually rendering the tooltip
+    delete keys.view_count.info_key;
 
     const sort = this.props.sort;
     if (sort.key) {
--- a/app/assets/stylesheets/modules/_articles.styl
+++ b/app/assets/stylesheets/modules/_articles.styl
@@ -28,7 +28,7 @@
 
     > span
       font-size 12px
-      font-weight 600
+      font-weight 800
 
     .view-mode-toggle
       display flex


oree-xx:
@Abishekcs kindly review
cc: @ragesoss

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
```
Abishekcs: You can use `useSelector` instead of `connect` since it's function component

ragesoss:
I think the UI for this needs to be closer to the column of views. Otherwise, it needs more of a label to show that it's specifically about the views.

app/assets/javascripts/components/articles/article.jsx
oree-xx: Thanks for this, I have updated it in my latest commit

oree-xx:
> I think the UI for this needs to be closer to the column of views. Otherwise, it needs more of a label to show that it's specifically about the views.

I have added a label, what do you think of the ui now. Updated the recording

ragesoss:
Hmm... UI is still not clear enough, I think. Using the 'on/off' style of toggle with the X is confusing, as it's a choice between two different views rather than a feature that is enabled and disabled. The also don't like how much space this takes up in the interface, and it's not obvious enough that everything from "Pageviews" to "Daily Avg" is part of the same UI element. (Maybe it needs a border?)

I think a toggle that is part of, or just above, the 'View' label at the top of the table column might be a better approach. Then the "pageviews" label wouldn't be needed.

oree-xx:
> Hmm... UI is still not clear enough, I think. Using the 'on/off' style of toggle with the X is confusing, as it's a choice between two different views rather than a feature that is enabled and disabled. The also don't like how much space this takes up in the interface, and it's not obvious enough that everything from "Pageviews" to "Daily Avg" is part of the same UI element. (Maybe it needs a border?)
> 
> I think a toggle that is part of, or just above, the 'View' label at the top of the table column might be a better approach. Then the "pageviews" label wouldn't be needed.

Oh okay, i was quite confused on how the ui should look, but this clears the confusion, will work on it, thanks

oree-xx:
@ragesoss I have updated the ui, you can check the updated screen recording to see how it looks now, thank you

ragesoss:
Oh, that looks much better! Can you move the '?' icon to beside the "Views" label instead of below the toggle?

oree-xx:
> Oh, that looks much better! Can you move the '?' icon to beside the "Views" label instead of below the toggle?

@ragesoss i have moved the '?' icon

ragesoss:
This looks really good. I will do code review and try it out locally this week.

ragesoss:
(If it's ready to go from your perspective, please update the PR title.)

oree-xx:
> This looks really good. I will do code review and try it out locally this week.

Thank you @ragesoss

oree-xx:
> (If it's ready to go from your perspective, please update the PR title.)

Done

PR #6498, opened 2025-10-21
Add type attribute to button element

Abdulwahab03:
Problem Description: There are button elements without type=button, and this behavior can change in ways we might not expect especially inside forms.

Solution Implemented: I added type="button" to the button elements in the components.


ragesoss:
This failing test may be related. If it changed a link into a button element in the HTML, it may require a change to the test.

Issue #6370, opened 2025-06-17
Add 'Download roster' option for course instructors


ragesoss:
Some instructors would find it useful to be able to download a spreadsheet of their students. The instructor who requested this asked for the following data:

* student's real name (ideally as surname, given name(s))
* student's username
* assigned article to edit
* assigned article for peer review

This is straightforward for cases where each student has only (at most) one assigned article and assigned peer review. I'm not sure what the best format design is for cases where there are multiple assignments per student, though, as the use case would involve sorting by assigned article to identify groups.

masembebrian:
assign me this issue please

Formasitchijoh:
Hi @masembebrian  issues are not assigned . Just indicate on the issue you are working on it so everyone knows

Also if you are new here try out issues tagged with new `newcomer friendly` to get you started rather than `enhancement `issues as this one

masembebrian:
noted

SanjanaSogimatt:
I am taking up this @ragesoss

SanjanaSogimatt:
@Abishekcs I wanted to know your thought on the plan before i actually start implementing this

### Objective:
Enable exporting student assignment data as CSV in both wide and detailed formats, leveraging existing reporting infrastructure.

**Backend Infrastructure**
- New route & controller action: Provides a dedicated endpoint for CSV export.
- CSV builder class (CourseStudentAssignmentsCsvBuilder): Encapsulates CSV generation logic for maintainability and reusability.
- Sidekiq worker update: Ensures async processing for large courses without blocking requests.

**CSV Builder Implementation**
- Wide format: One row per student, multiple assignments as columns → better overview for instructors.
- Detailed format: One row per assignment → suitable for granular analysis.
- Name formatting & role separation: Improves readability and distinguishes assignment types (edit vs peer review).

**Frontend Integration**
- Download buttons & localization: Makes the feature accessible and user-friendly in the UI, supports multiple languages.

**Testing & Edge Cases**
- Ensures correctness with sample data.
- Handles students with no assignments or multiple assignments → prevents errors and data inconsistencies.

### Data Flow:

### > Route → Controller → Worker → CSV Builder → Query → CSV Download

**Reasoning:**
- Follows existing reports infrastructure → consistency & easier maintenance.
- Async processing via Sidekiq → scalable for large courses.
- Wide and detailed formats → flexible use for instructors and admins.

Reference image - 

[...]

PR #6500, opened 2025-10-21
Add course students assignments CSV functionality

SanjanaSogimatt:
## What this PR does
- Introduced a new CSV report for student assignments, allowing download of data including real names, usernames, assigned articles, and peer review articles.
- Updated the ReportsController to handle the new CSV endpoint.
- Added corresponding routes and localization strings for the new feature.
- Enhanced the CourseStatsDownloadModal component to include a link for downloading the assignments CSV.

Closes - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6370

## Screenshots
Before:
Screenshot 2025-10-21 at 10 55 21 PM

After:
Screenshot 2025-10-21 at 10 43 29 PM

Sample output ( a basic one ):
```
real_name,username,assigned_article_title,peer_review_article_title
"",Camelia.boban,"",""
"",Kaspo,"",""
"",Manuelarosi,"",""
"",Alexmar983,"",""
"",Giaccai,"",""
"",LigaDue,"",""
"",Solelu,"",""
"",Nea.Lapini,"",""
"",Chiara.toti,"",""
"",MissDaae,"",""
"",Ciucia60,"",""
"",Lizwicks,"",""
"",Katy1q77,"",""
"",Mariannapetricelli,"",""
```

### DIFF
--- a/lib/analytics/course_students_assignments_csv_builder.rb
+++ b/lib/analytics/course_students_assignments_csv_builder.rb
@@ -95,6 +95,8 @@ class CourseStudentsAssignmentsCsvBuilder
                                    assigned_articles, reviewing_articles)
     process_assigned_articles(csv_data, real_name, username,
                               assigned_articles, reviewing_articles)
+    return if assigned_articles.count >= reviewing_articles.count
+
     create_reviewing_article_rows(csv_data, real_name, username,
                                   reviewing_articles, assigned_articles.count)
   end
@@ -115,8 +117,6 @@ class CourseStudentsAssignmentsCsvBuilder
   def create_reviewing_article_rows(csv_data, real_name, username,
                                     reviewing_articles, start_index = 0)
     # Create rows for reviewing articles starting from the specified index
-    return if start_index >= reviewing_articles.count
-
     reviewing_articles[start_index..].each do |reviewing_article|
       csv_data << [
         real_name,


Abishekcs:
It has `N+1` query should be a easy fix. `N+1` are most of the time not a big issue really, but sometimes for some `large course` it leads to performance problems that we can't predict when it might cause an issue.

n plus one

SanjanaSogimatt:
> It has `N+1` query should be a easy fix. `N+1` are most of the time not a big issue really, but sometimes for some `large course` it leads to performance problems that we can't predict when it might cause an issue.
> 
> n plus one

I didn't get it, what should happen?

lib/analytics/course_students_assignments_csv_builder.rb
```diff
@@ -0,0 +1,132 @@
...
+class CourseStudentsAssignmentsCsvBuilder
+  def initialize(course)
+    @course = course
+  end
+
+  def generate_csv
+    csv_data = [CSV_HEADERS]
+
+    students.each do |courses_user|
+      process_student_assignments(csv_data, courses_user)
```
Abishekcs: > I didn't get it, what should happen?

This line is probably the source of the N+1 problem  for each student, it’s making a separate query to the assignment table in the database.

It would be great if it could fetch all the details of all the assignments for all the students in a single query.

to put in simple words `you can think of the N+1 query problem as a situation where multiple round trips are made to the database, even though the same work could be done with just one round trip.`

SanjanaSogimatt: resolved it

Abishekcs:
This looks good. The current format works fine, though another option could be to show multiple assignments or reviews for a student in the same row, separated by commas but again that might be more messy.

@ragesoss any suggestion?

Abishekcs:
>Sample output ( a basic one ):

@SanjanaSogimatt can you upload a screenshot of the actual CSV format too?

SanjanaSogimatt:
> can you upload a screenshot of the actual CSV format too?

That is the actual format, Am I understanding something wrong? 🤔

Abishekcs:
I meant like this 

screenshot-2025-10-28_00-02-04


It's difficult to see how the acutal CSV will look like to the end user with only  `simple output` given in PR description

SanjanaSogimatt:
Screenshot 2025-10-28 at 11 19 13 AM
@Abishekcs

lib/analytics/course_students_assignments_csv_builder.rb
```diff
@@ -0,0 +1,140 @@
...
+      csv_data << [
+        real_name,
+        username,
+        assigned_article.article_title,
+        reviewing_article || ''
+      ]
+    end
+  end
+
+  def add_remaining_reviewing_articles(csv_data, real_name, username,
```
Abishekcs: These methods `add_remaining_reviewing_articles` and `create_reviewing_article_rows` have duplicate logic and can be combined into one i think. The difference is just the starting index (0 vs assigned_articles.count), which could be passed as an optional parameter.

lib/analytics/course_students_assignments_csv_builder.rb
```diff
@@ -0,0 +1,129 @@
...
+        assigned_article.article_title,
+        reviewing_article || ''
+      ]
+    end
+  end
+
+  def create_reviewing_article_rows(csv_data, real_name, username,
+                                    reviewing_articles, start_index = 0)
+    # Create rows for reviewing articles starting from the specified index
+    return if start_index >= reviewing_articles.count
```
Abishekcs: You could move the `return if start_index >= reviewing_articles.count` check to `create_assigned_article_rows` instead.
That way, we can skip calling `create_reviewing_article_rows` altogether when there are no remaining reviewing articles.

SanjanaSogimatt:
@Abishekcs can you review this?

Abishekcs:
I will try to review it within the next few days. Thank you!

Abishekcs:
This is looking good. By the way did you check the `guard logs` when we download `CSV` to make sure `N+1` was surely eliminated.It seems it is resolved by looking at the code,  just wanted to be sure.

PR #6504, opened 2025-10-22
Updated Calendar component.

Libonet:
## What this PR does
Part of #5393 (Component refactor for React class components)

Refactors the Calendar component into a functional component.

## Screenshots
Before:

https://github.com/user-attachments/assets/91e04102-9ff1-4c23-a0d9-80028ca1b36c

After:

https://github.com/user-attachments/assets/9391d926-818f-474e-a196-78e3ac914286

While doing this refactor I found that the calendar component wasn't displaying the selected days correctly, other than when selecting the days of the week.  
It was fixed in the refactor by returning a mutated course on the "selectDay" function.

### DIFF
--- a/app/assets/javascripts/components/overview/course_cloned_modal.jsx
+++ b/app/assets/javascripts/components/overview/course_cloned_modal.jsx
@@ -115,6 +115,9 @@ const CourseClonedModal = createReactClass({
     if (__guard__(updatedCourse.day_exceptions, x => x.length) > 0 || updatedCourse.no_day_exceptions) {
       this.props.setValid('holidays');
     }
+
+    this.setState({ course: updatedCourse });
+
     return this.props.updateCourse(updatedCourse);
   },
 


ragesoss:
Thanks! Can you separate out the bug fix from the refactoring? Also, a demo of the bug would be really helpful.

Libonet:
> Thanks! Can you separate out the bug fix from the refactoring? Also, a demo of the bug would be really helpful.

Sorry, I thought I had replied. Does this work as a demo of the bug?

https://github.com/user-attachments/assets/7108b183-c249-4e88-9325-59c44867cdf2

ragesoss:
@Abishekcs @empty-codes would either of you like to review this in the next couple of days? If not, let me know and I will.

app/assets/javascripts/components/common/calendar.jsx
```diff
@@ -1,199 +1,192 @@
...
-        const weekday = this.props.course.weekdays.charAt(weekNumber) === '1';
-        return inrange && exception && weekday;
       }
-    };
+      const toPass = course;
+      toPass.day_exceptions = exceptions.join(',');
+      toPass.no_day_exceptions = compact(exceptions).length === 0;
+      return updateCourse(toPass);
+    },
+    [course, updateCourse, inrange, CourseDateUtils]
```
empty-codes: Since `CourseDateUtils` is a static import, is there a need to include it in the dependency array? If i am assuming correctly, it doesn't change like the other dependencies.

app/assets/javascripts/components/common/calendar.jsx
```diff
@@ -1,199 +1,192 @@
...
+  }, [course.start]);
 
-  getInitialState() {
-    return ({ initialMonth: toDate(this.props.course.start) });
-  },
-  shouldComponentUpdate(nextProps) {
-    if (nextProps.course) {
-      const start = toDate(nextProps.course.start);
-      const end = toDate(nextProps.course.end);
-      return isValid(start) && isValid(end);
```
empty-codes, line 41: I also see that you removed the `isValid` check. Are you confident it's redundant? if it is, it is fine to skip it but if not you could reintroduce it since it's basically another safety net

app/assets/javascripts/components/common/calendar.jsx
Libonet: Yes, I don't think it should be there. My mistake

app/assets/javascripts/components/common/calendar.jsx
Libonet: I'm not completely confident it's redundant. I removed it because the functional component is always updated when the parent updates it. Should I put the checks in 'shouldComponentUpdate' inside a react memo?

Abishekcs:
> This is a clear improvement, good work! I left some comments but I would appreciate if @Abishekcs also takes a look at the code to see if i missed anything.

I can go through this code tomorrow

app/assets/javascripts/components/common/calendar.jsx
empty-codes: Oh okay, I see your point now. In the old class component, `isValid()` prevented the entire component from re-rendering when start or end dates were invalid. So it had like a dual purpose.
With your refactoring, rendering is already controlled by React’s default flow so we don’t need it for render prevention anymore.

If you were to add the `isValid` checks it would simply be in `inrange()` or wherever the start and end dates are used, not in a memo since it isn't really concerned with performance (anymore)

However, I see you added `if (!inrange(day)) { return; }` in `selectDay` (line 41) so that should be enough, you can leave this as is then 👍

Abishekcs:
I’m still going through the code since it’s used by a few parent components, so need to be careful. But while testing it out, it seems there’s a Redux state mutation happening.




https://github.com/user-attachments/assets/61041d39-5286-4df6-9e01-972eb3e72bae

app/assets/javascripts/components/common/calendar.jsx
```diff
@@ -1,196 +1,192 @@
...
-        if (!this.props.course.day_exceptions || !this.props.course.weekdays) { return false; }
-        const formatted = format(day, 'yyyyMMdd');
-        const inrange = this.inrange(day);
-        const exception = __in__(formatted, this.props.course.day_exceptions.split(','));
-        const weekNumber = getDay(day);
-        const weekday = this.props.course.weekdays.charAt(weekNumber) === '1';
-        return inrange && exception && weekday;
       }
-    };
+      const toPass = course;
```
Abishekcs: This might be the source of `state mutation` as `course` is a `prop` which is probably coming from `redux store`

Abishekcs: set `enableReduxSafetyChecks = true` in `create_store.js` to detect `state mutation`

Libonet: Is the problem with updating the selected days in the updateCourse then? This way of changing the course was also used when selecting weekdays

Abishekcs: Maybe, it's better to fix mutation in the refactor itself should require minor changes.

Libonet: When copying the course with updated fields, updateCourse doesn't seem to re-render the calendar.

https://github.com/user-attachments/assets/d57b0f8d-ace4-46a5-9434-8326587c1e73

I'm not sure where the problem with updating the selected days comes from then

Abishekcs: While testing did you set `enableReduxSafetyChecks = true` ?

Libonet: Yes. The error didn't popup for me, but it would make sense if it did. I don't know if I've made a mistake on my setup

Abishekcs: That's weird let me check if i can also find the source i will let you know.

Abishekcs: Okay, If we select `week` first only then the `state mutation` occurs otherwise it won't


https://github.com/user-attachments/assets/919ef7e8-4f67-4db4-a8f1-5ffc254867fb

Abishekcs: If you can fix it and also figure out why this `mutation` occurs only in this pattern that will be great.

Libonet:
After we stop mutating the course in 'selectDay' and 'selectWeekday' the calendar doesn't update. It seems the problem was coming from the parent component in this case.


https://github.com/user-attachments/assets/043faadb-2b58-47c6-b3ad-85f4bf9cd926

I've enabled the redux safety checks, but I'm not sure if they even triggered for me in the first place before this fix.

Abishekcs:
Okay, great! I will try to review this within the week. Thank you!

Abishekcs:
okay, i have tested it out no more `state mutation`.Code is looking good too.

Abishekcs:
> I've enabled the redux safety checks, but I'm not sure if they even triggered for me in the first place before this fix.

I don't know why it didn't trigger for you. But with your previous **commit** `Stop mutating passed course` it has been resolved.

Issue #6499, opened 2025-10-21
Improve usability of 'Change sandbox' feature and enable it for 'no-sandbox' courses


ragesoss:
Currently, the 'Change sandbox' button is available for each assigned article in the My Articles section of the course page, but it's not very self-explanatory how to use it.

It could be streamlined to just take a username (or extract a username from a sandbox URL), since the use case for it is that an assignment has a classmate's username and the student wants to use sandboxes under their own username instead. Some text instructions that describe what will change and/or how to fix the main use case (enter your own username).

It should also be enabled for 'no sandboxes' courses, since they still use a Bibliography and Outline sandbox.

SanjanaSogimatt:
Taking up this

SanjanaSogimatt:
@ragesoss @Abishekcs does this sound good?

### Frontend Changes

- Create username input component - Replace URL input with simple username field + preview
- Update EditSandboxUrl - Use new component with clear instructions
- Remove course restriction - Enable button for all course types (line 35 in Actions.jsx)


### **Backend Changes**

- Modify SandboxUrlUpdator - Accept username input and generate URLs
- Update assignment controller - Handle username instead of full URL

### **User Experience**
- Add instructions - "This changes sandbox from classmate's username to yours"
- Show preview - Display new URLs before confirming
- Add validation - Ensure valid username format

### Implementation:
- [x] Frontend UI (username input + preview)
- [x] Backend API (username → URL generation)
- [x] Add instructions and validation




Result: Simple username input that generates proper sandbox URLs with clear instructions, available for all course types.

ragesoss:
It might be best to implement this solely on the frontend, and still pass a URL to the backend. Otherwise, this seems like a reasonable plan. I'll probably have feedback about the instruction copy, but I'm not sure what exactly it should say yet.

PR #6508, opened 2025-10-24
Implement EditSandboxUsernameInput component and update sandbox URL handling

SanjanaSogimatt:
## What this PR does
- Added a new React component, EditSandboxUsernameInput, for editing sandbox usernames with validation and URL preview functionality.
- Updated the Actions component to enable sandbox URL editing for all courses, including those without sandboxes.
- Refactored EditSandboxUrl to use the new EditSandboxUsernameInput component.
- Enhanced the SandboxUrlUpdator service to process usernames and generate sandbox URLs accordingly.
- Added localization strings for username validation and instructions in the English locale file.
- Updated styles for the new input form to improve user experience.

Closes - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6499

## Screenshots
Before:
Screenshot 2025-10-24 at 4 11 26 PM

After:
Screenshot 2025-10-24 at 4 07 43 PM

### DIFF
--- a/app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/EditSandboxUrl.jsx
+++ b/app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/EditSandboxUrl.jsx
@@ -70,7 +70,7 @@ render() {
         <Popover
           is_open={is_open}
           edit_row={editRow}
-          styles={{ minWidth: '500px' }}
+          styles={{ width: '500px', tableLayout: 'fixed' }}
         />
       </div>
     );
--- a/app/assets/stylesheets/modules/_assignments.styl
+++ b/app/assets/stylesheets/modules/_assignments.styl
@@ -37,10 +37,14 @@ tr.assignment-section-header h3
   flex-direction column
 
 .edit_sandbox_url_input
-  min-width: 30rem;
+  width 100%
+  max-width 100%
+  box-sizing border-box
 
 .edit-sandbox-username-form
   padding 10px 0
+  max-width 460px
+  box-sizing border-box
 
   .instruction-text
     font-size 14px
@@ -71,6 +75,8 @@ tr.assignment-section-header h3
     padding 12px
     border 1px solid $border_lt
     margin-top 10px
+    max-width 460px
+    box-sizing border-box
 
     .preview-label
       font-weight 600
@@ -79,11 +85,15 @@ tr.assignment-section-header h3
 
     .preview-url
       margin-top 8px
+      width 100%
+      overflow hidden
       
       .url-type
         font-size 12px
         font-weight 600
         color $text
+        display block
+        margin-bottom 2px
 
       .url-value
         font-size 11px
@@ -92,7 +102,11 @@ tr.assignment-section-header h3
         font-family monospace
         padding 4px
         background-color white
-        margin-top 2px
+        display block
+        width 100%
+        box-sizing border-box
+        overflow-wrap break-word
+        white-space pre-wrap
 
   button[type="submit"]
     margin-top 10px


ragesoss:
This is a great start for the UI! I like the previews of new URLs. Does the `Sandbox` one get removed if the course is set to `no_sandboxes` mode? (It should.)

I think this is almost universally used to change the sandbox from a different username to your own, so it might make sense to offer a streamlined path to do that, but I think this is sufficiently self-explanatory that it will be easy for people to do what they intend.

app/assets/stylesheets/modules/_assignments.styl
```diff
@@ -37,4 +37,80 @@ tr.assignment-section-header h3
...
+    .input-label
+      font-weight 600
+      font-size 13px
+      color #333
+
+  .error-message
+    color #d33
+    font-size 13px
+    padding 8px 12px
+    background-color #fee
```
ragesoss: Ideally, CSS colors should use variables so that we have consistency in color usage across the UI.

app/assets/stylesheets/modules/_assignments.styl
```diff
@@ -37,4 +37,80 @@ tr.assignment-section-header h3
   flex-direction column
 
 .edit_sandbox_url_input
-  min-width: 30rem;
\ No newline at end of file
+  min-width: 30rem;
+
+.edit-sandbox-username-form
```
ragesoss, line 44: This is a lot of new CSS. Can it be streamlined? Adding new CSS is okay, but the less the better.

app/services/sandbox_url_updator.rb
```diff
@@ -23,9 +26,73 @@ def update
     end
   end
 
+  private
+
+  def process_input
```
ragesoss, line 28: I don't like how much custom code for username validation is happening here, especially since there is no guarantee that it will avoid problems and result in a valid sandbox URL.

I better strategy might be to find the user by username; if the user exists, that will definitely be a valid sandbox URL, and if the user does not exist, we should disallow using that sandbox anyway (even if it's technically a usable URL).

app/services/sandbox_url_updator.rb
```diff
@@ -23,9 +26,73 @@ def update
...
+    # Check if input is a full URL or just a username
+    if @input.start_with?('http://', 'https://')
```
ragesoss: Since the new frontend does not say full URLs are supported, I don't think we need to support them on the backend either.

app/services/sandbox_url_updator.rb
ragesoss: I think a lot of this new code could be removed or simplified if the `find-the-user` strategy is used here.

SanjanaSogimatt:
@ragesoss resolved all comments 🙌🏻

SanjanaSogimatt:
@Abishekcs @ragesoss  how do i resolve the CI build failing?

ragesoss:
This look like unrelated failures, either because of a temporary problem with the pageviews API or something that changed with the pageview data that we need to update the specs for. I'll restart the build here to see if it's resolved, but if not, we can review this PR and ignore the failing specs.

app/services/sandbox_url_updator.rb
```diff
@@ -23,9 +26,53 @@ def update
...
+    username = username.sub(/^User:/, '') if username.start_with?('User:')
+
+    # Find the user in the database
+    user = User.find_by(username:)
+    return user if user
+
+    raise UserNotFoundError, I18n.t('assignments.invalid_username', username:)
+  end
+
+  def generate_sandbox_url_for_user(user)
```
ragesoss: This is more complicated than necessary. Since we're taking valid usernames, rather than arbitrary URLs, generating the new URL can be done in basically the same way as the initial URL was generated, which is via `Assignment#default_sandbox_url`. That method could perhaps be modified to take an optional argument of the username.

ragesoss:
This is looking solid. A few UI improvements could be made...

* The side the popover changes as you type, because the width is determined by the longest URL. If there's an elegant way to avoid that problem, that would be helpful.
* It should not display the first link ('Sandbox') if a course has the `no sandboxes` setting enabled.
* The submit button should not be enabled if the URL has not been changed from its current value.

app/assets/javascripts/components/overview/my_articles/components/Categories/List/Assignment/Header/Actions/EditSandboxUrl.jsx
```diff
@@ -70,6 +70,7 @@ render() {
         <Popover
           is_open={is_open}
           edit_row={editRow}
+          styles={{ minWidth: '500px' }}
```
ragesoss: This doesn't seem to have had the intended effect. I still get the component changing width as I type, no matter how short the input.

ragesoss:
Thanks! Remaining issues:
* The width still changes as the input size changes
* The amount of new CSS is still larger than I'd like

It's an improvement over the current situation already, so I'm willing to merge it now and add an issue to clean up those issues later, but please let me know if you're still working on those things for this PR.

SanjanaSogimatt:
> Thanks! Remaining issues:
> 
> * The width still changes as the input size changes
> * The amount of new CSS is still larger than I'd like
> 
> It's an improvement over the current situation already, so I'm willing to merge it now and add an issue to clean up those issues later, but please let me know if you're still working on those things for this PR.

Ok, i'll make changes in this pr itself.

Issue #6502, opened 2025-10-21
Make it easier to generate a survey preview link


ragesoss:
When you want to send out a survey for testing and review, you have to first pick the course out of a dropdown of all courses and then proceed to the survey. Since the survey behavior can depend on which course you choose, it's important to choose an appropriate one in some cases.

It should be easier to get a default preview link for a current course, and to create a preview link for a specific course.

SanjanaSogimatt:
picking this up

SanjanaSogimatt:
### How I plan on doing this - 
**On the survey list page:**

- Change the current "Preview" link to include a default course slug
- Example: Instead of /surveys/1?preview, make it /surveys/1?preview&course_slug=some-default-course

**On the survey edit page:**

- Add a simple section that shows: "Preview URL: [copy button]"
- Let admin pick a course from dropdown, instantly generate the preview link
- Can copy/paste to share with reviewers

How does it sound? @ragesoss does it need any improvements?

ragesoss:
For the preview courses, I think the important aspect we usually care about is to generate appropriate URLs representing each of the relevant conditions set for the question groups. For example, a survey might have two question groups; one has the tag 'returning_instructor' and the other has the tag 'first_time_instructor'. In that case, we will want at least two different preview links, one for a course that has the returning instructor tag and another for a course that has the new instructor tag. Similarly, if there campaigns for the question groups, we will want courses that are in those campaigns (so that the previews will include those question groups).

This could get complicated, but I think as a start, just looking at the tags and finding a recent course with that tag for each tag would be a good way to do it.

PR #6509, opened 2025-10-25
Add survey preview links based on question group tags

SanjanaSogimatt:
## What this PR does
- Implemented a method to generate preview links for surveys based on associated question group tags.
- Updated the survey view to display a dropdown for multiple preview links or a single link if only one is available.
- Added helper methods to collect tags, find recent courses, and build preview link objects.

Closes - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6502

## Screenshots
Before:

https://github.com/user-attachments/assets/e4408883-bd5f-4827-a51f-c3ab835401b4

After:

https://github.com/user-attachments/assets/8a8662d9-5bd0-4fcb-bc7b-bc1f991fff39

In the above screen recording example - I created 2 tags one in the course and another in question group called as `test` and hence we can see the URL taking that in the course_slug, Please ignore the questions in the testing 😅

### DIFF
--- a/app/assets/stylesheets/survey_modules/survey__admin-row.styl
+++ b/app/assets/stylesheets/survey_modules/survey__admin-row.styl
@@ -103,7 +103,7 @@
 .surveys__admin-header--question-groups--edit
   margin-bottom 0 !important
 
-// Preview dropdown styles (accordion-style like FAQ)
+// Preview dropdown - extends existing .pop styles
 .preview-dropdown
   position relative
   display inline-block
@@ -120,19 +120,13 @@
   box-shadow 0 4px 12px rgba(0,0,0,0.15)
   border-radius 4px
   margin-top 4px
-  padding 8px 0
   z-index 1000
-  overflow hidden
 
 .preview-dropdown.collapsed .preview-dropdown__menu
-  max-height 0
-  padding 0
-  box-shadow none
-  transition max-height 0.2s, padding 0.2s
+  display none
 
 .preview-dropdown.expanded .preview-dropdown__menu
-  max-height 500px
-  transition max-height 0.3s
+  display block
 
 .preview-link-item
   padding 10px 16px
@@ -143,26 +137,21 @@
     background-color rgba(103, 110, 180, 0.05)
 
 .preview-link
-  display block
-  text-decoration none
   color $brand_primary
   font-weight 500
-  margin-bottom 4px
   &:hover
     text-decoration underline
 
 .preview-link-detail
-  display flex
-  flex-direction column
   margin-top 4px
   font-size 13px
-  color $text_lt
 
-.preview-link-detail .course-name
-  color $text_dk
-  font-weight 500
-  margin-bottom 2px
+  .course-name
+    color $text_dk
+    font-weight 500
 
-.preview-link-detail .question-groups
-  color $text_dk
-  font-size 12px
+  .question-groups
+    display block
+    color $text_dk
+    font-size 12px
+    margin-top 4px


app/views/surveys/_survey.html.haml
```diff
@@ -19,7 +19,20 @@
   .survey__admin-row__col.survey__admin-row__col--actions
     = link_to 'Edit', edit_survey_path(survey)
     - if survey.rapidfire_question_groups.length > 0
-      = link_to 'Preview', "#{survey_url(survey)}?preview"
+      - preview_links = survey_preview_links(survey)
+      - if preview_links.length == 1
+        = link_to 'Preview', preview_links.first[:url]
+      - else
```
Abishekcs, line 25: Can you attach a screenshot for the else part too.

SanjanaSogimatt: Screenshot 2025-10-27 at 7 22 56 PM
I added more css to make it better

Abishekcs: Okay make sense it was not working before since CSS was not added.

Can you make the `Shows:` text color a bit different? It’s hard to read because it almost blends into the background.

app/helpers/surveys_helper.rb
```diff
@@ -25,6 +25,14 @@ def survey_preview_url(survey)
     "#{survey_url(survey)}?preview"
   end
 
+  # Generate preview links for courses that match question group conditions
+  def survey_preview_links(survey)
+    tags_with_groups = collect_tags_from_question_groups(survey.rapidfire_question_groups)
+    links = generate_preview_links_for_tags(survey, tags_with_groups)
+    add_generic_preview_if_needed(survey, links)
```
Abishekcs: It might be cleaner to check whether links is present before calling `add_generic_preview_if_needed`, so you can avoid the extra method **(instead of checking in the method if link is present)** call when not needed.

SanjanaSogimatt: Thanks, it actually makes a diffrence

Abishekcs:
LGTM!

Abishekcs:
I had a question about this feature shouldn’t the preview link (URL)

```
http://localhost:3000/surveys/4?preview&course_slug=QCA/Brisbane_QCA_ArtandFeminism_2018
```
 contain information about the tag? So, that the user can know which kind of survey they are taking like `returning_instructor` or `first_time_instructor`

ragesoss:
> I had a question about this feature shouldn’t the preview link (URL)
> 
> ```
> http://localhost:3000/surveys/4?preview&course_slug=QCA/Brisbane_QCA_ArtandFeminism_2018
> ```
> 
> contain information about the tag? So, that the user can know which kind of survey they are taking like `returning_instructor` or `first_time_instructor`

I don't think that's necessary.

app/assets/stylesheets/survey_modules/survey__admin-row.styl
```diff
@@ -102,3 +102,57 @@
 
 .surveys__admin-header--question-groups--edit
   margin-bottom 0 !important
+
+// Preview dropdown styles
+.preview-dropdown
```
ragesoss, line 107: Should these styles be scoped more specifically? `.preview-dropdown` isn't a classname that is specific to `survey__admin-row`, so these should either be in a different file (if they likely to be generally useful in other contexts) or they should follow the naming convention of css classes that are specific to the filename.

ragesoss:
This is cool!

1. For the surveys without relevant tags, the preview links should also have a course by default.

2. This kind of mouseover-based dropdown is always a little annoying, as you have to move the mouse in a particular way to access the links. Something that you click to reveal and then it stays open until you click again is often a better UI. One simple example of this is elsewhere in the app is the FAQ accordion UI: https://dashboard.wikiedu.org/faq?topic=instructor_faq
Clicking the down arrow for an entry opens the section and the content remains visible until it is closed. This might be a better way to do it (and might also let you reduce the amount of new CSS).

3. If the dropdown UI is kept... if I slowly drag my mouse from the Preview box down to the dropdown, the dropdown disappears when the mouse is in the gap between the box and the dropdown. (When moving the mouse quickly, it reaches the dropdown before disappearing, so it works as intended in that case.)

SanjanaSogimatt:
> * This kind of mouseover-based dropdown is always a little annoying, as you have to move the mouse in a particular way to access the links. Something that you click to reveal and then it stays open until you click again is often a better UI. One simple example of this is elsewhere in the app is the FAQ accordion UI: https://dashboard.wikiedu.org/faq?topic=instructor_faq
>   Clicking the down arrow for an entry opens the section and the content remains visible until it is closed. This might be a better way to do it (and might also let you reduce the amount of new CSS).
> * If the dropdown UI is kept... if I slowly drag my mouse from the Preview box down to the dropdown, the dropdown disappears when the mouse is in the gap between the box and the dropdown. (When moving the mouse quickly, it reaches the dropdown before disappearing, so it works as intended in that case.)

1. I didn't get the first one, what's the default course? Like any specific condition 
2. Changed the ui, I think it's better now please take a look at code and lmk also updated the names

ragesoss:
> I didn't get the first one, what's the default course? Like any specific condition

Currently, the preview links that don't have your feature applied (because there are no tags) result in a preview link like `/surveys/50?preview` and then when you follow that link, you still have to select a course from the dropdown. If there are not courses or campaigns specified in the question groups, then it doesn't matter which course is selected (unless there are questions that let the user select from among the edited articles or student usernames). So, the idea would be to just include a course (for example, any course that has some edited articles) by default so that a course doesn't need to be selected by the user when previewing the survey.

ragesoss:
UI is better, thanks!

SanjanaSogimatt:
> Currently, the preview links that don't have your feature applied (because there are no tags) result in a preview link like `/surveys/50?preview` and then when you follow that link, you still have to select a course from the dropdown. If there are not courses or campaigns specified in the question groups, then it doesn't matter which course is selected (unless there are questions that let the user select from among the edited articles or student usernames). So, the idea would be to just include a course (for example, any course that has some edited articles) by default so that a course doesn't need to be selected by the user when previewing the survey.

lmk if the recent update works - I made couple of fallback states

ragesoss:
Thanks! This looks good to me! My remaining concern is the amount of new CSS. Can it be simplified, and/or adapted to re-use existing classes for most of the styling?

SanjanaSogimatt:
> Thanks! This looks good to me! My remaining concern is the amount of new CSS. Can it be simplified, and/or adapted to re-use existing classes for most of the styling?

I am trying 🥲

ragesoss:
Let me know when you've finished (and please include an updated screenshot, to make sure it reflects the latest styling).

SanjanaSogimatt:
> Let me know when you've finished (and please include an updated screenshot, to make sure it reflects the latest styling).

Screenshot 2025-11-04 at 9 45 01 PM

This is the updated one. I tired to reduce the code if I reduce more the style will be changed and it kinda looks ugly ngl when it's changed

ragesoss:
Okay. This works for me.

PR #6510, opened 2025-10-25
upgrade ruby version from 3.1.2 to 3.3.3

Abdulwahab03:
Description: This PR upgrade ruby version from 3.1.2 to 3.3.3

I made a proposal based on this implementation for Outreachy internship, here is the phrabricator for the proosal: https://phabricator.wikimedia.org/T408124

This is the outcome of running **bundle install** command after ruby was upgraded from 3.1.2 to 3.3.3
bundle install outcome. This allows to resolve all the dependencies and also to updates the Gemfile.lock

  db check bundle exec rails dbconsole This is the outcome after running **bundle exec rails db:check**  in order to verify that my database configuration and connection work properly.

outcome after running sidekick
this is the outcome after running **bundle exec sidekiq** to check that the sidekiq is working properly

Note: during the process of upgrading some gems got updated too from their old version here is the picture
updating bundle

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '3.3.3'
+ruby '3.1.2'
 
 ### Basic Framework
 gem 'rails', '7.0.7'
@@ -9,7 +9,6 @@ gem 'bootsnap', require: false # Makes rails boot faster via caching
 gem 'faker', require: false # Generates random data for example records
 gem 'figaro' # easy access to ENV variables. Deprecated.
 gem 'puma'
-gem 'logger', '~> 1.6'
 
 ### Database and caching
 gem 'mysql2' # MariaDB integration for ActiveRecord
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -142,7 +142,7 @@ GEM
       msgpack (~> 1.2)
     breadcrumbs_on_rails (4.1.0)
       railties (>= 5.0)
-    browser (6.2.0)
+    browser (5.3.1)
     builder (3.3.0)
     byebug (12.0.0)
     capistrano (3.19.2)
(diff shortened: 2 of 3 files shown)


Abdulwahab03:
though the failure claimed that my ruby version is 3.1.2 which I have upgraded already and set in my gemsfile, @ragesoss and @Abishekcs can you please help me look into this?"

Abdulwahab03:
rails incompatibility with ruby 3 3 3
I think the problem that caused the build failure is the incompatibility of rails 7.0.7 to newly upgraded ruby version which is ruby 3.3.3. If we can upgrade ruby too gradually from 7.0.7 to 7.1.3 and to finally 8.0.0 the problem should solve..  I want your suggestions @ragesoss, @empty-codes and @Abishekcs before proceeding to upgrade rails

Abishekcs:
Not sure but it might be  the case  that  you need the `ruby` version required by the current `rails` version and so on

Abdulwahab03:
It seems so @Abishekcs based on my findings Ruby 3.3.3 moved many built-in classes like Logger, OpenURI, Timeout, into separate default gems. Which make ruby 3.3.3 not to autoload Logger but Rails 7.0.7 requires Logger to be autoloaded by Ruby core. source from https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/. I will wait to hear from @ragesoss before upgrading the rails

Abishekcs:
What about schema.rb i think that's a mistake?

ragesoss:
This is probably too much to upgrade at once. The best way to do ruby upgrades is to first bump all the gems that need to be upgraded to remain compatible with the new ruby version, and then bump the ruby version separately (or with only the gem upgrades that can't be upgraded independently, if there are any).

Abdulwahab03:
thank you so much @ragesoss for the feedback, if I understand your comment well sir, you mean before upgrading the Ruby version, I will first need to check for some outdated gems in my current ruby version let's say 3.1.2 and upgrade them gradually maybe 10 per PRs for them to be compatible with the new Ruby version. Then after doing all these, I will now upgrade the Ruby version to the new version independently or with some gems that can't be upgraded independently without the new Ruby version right?

Abdulwahab03:
I will need to close this PR and start afresh.

Issue #6338, opened 2025-05-29
Article Viewer makes too many requests too frequently when wikiwho data is not available


ragesoss:
If a request for authorship data results in a message indicating that the data is not available yet on the wikiwho server, it currently will continue making requests indefinitely. It should have exponential backoff and/or a max number of re-requests before it gives up. (It used to stop after a certain number of requests, so later changes might have broken the intended failure handling.)

marshadkhn:
I searched the repo but couldn't find the client file that requests /wikiwho/authorship. Could you point me to the file(s) that make those requests (Article Viewer or related components)?

If there isn't a dedicated place yet, may I add a small utility (e.g. app/assets/javascripts/utils/fetch_with_retry.js or app/javascript/utils/fetch_with_retry.js depending on the build setup) that provides:
- exponential backoff with jitter
- a max retry limit (suggested default: maxRetries=6, baseDelay=1000ms, multiplier=2)
- AbortController support to stop retries on unmount
and then update the Article Viewer to use it and show a “data being prepared - try again later” message on final failure?

Thanks.

ragesoss:
I think these requests come from [`ArticleViewerAPI.js`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/components/common/ArticleViewer/utils/ArticleViewerAPI.js).

marshadkhn:
okay thankyou

PR #6511, opened 2025-10-26
Fix: Reduce frequent Article Viewer requests

marshadkhn:
### What this PR does
This PR fixes “Article Viewer makes too many requests too frequently when wikiwho data is not available” by adding proper retry/backoff and a cooldown, and by preventing a verify-and-refetch loop that caused repeated WhoColor requests even after failures.

- Addresses: Article Viewer re-requesting WhoColor indefinitely when the API reports data is not yet available.
- Scope: ArticleViewerAPI.js and a small guard in `ArticleViewer.jsx.`

- Impact:
  - Fewer network calls
  - Capped retries
  - No infinite loops
  - User sees a clear failure state if data isn’t ready.

**Background**
When the WhoColor endpoint returns a 200 with `success: false` (data not yet prepared), the viewer kept trying again quickly and could be re-triggered by the “verify title” path in the container, causing repeated requests. This PR restores and strengthens the intended failure handling.

**Changes in detail**
Exponential backoff with jitter and hard cap

- Location:` ArticleViewerAPI.fetchWhocolorHtml`

- Behavior:
  - Retries now use exponential delays with a small jitter: ~0s, ~1s, ~2s, ~4s, ~8s (capped per-attempt delay $\le$ 30s).
  - Attempts are capped at 5 (configurable in the function).
  - On terminal failure, preserves the existing error message format:
  - “Request failed after 5 attempts. …”

**Cooldown to avoid re-hammering the API when not ready**
- Location: `ArticleViewerAPI.js`

- Behavior:

- Adds an in-memory cooldown cache keyed by language|title|revision.
- After a terminal failure, subsequent calls for the same article+revision within 5 minutes are short-circuited locally and do not hit the network.
- Returns a friendly error:
  - “WhoColor temporarily unavailable; retry later (cooldown active).”

**Stop the verify-and-refetch loop in the container**

- Location: `ArticleViewer.jsx (catch handler for fetchWhocolorHtml)`

- Behavior:
  - On WhoColor unavailability errors (either “Request failed after N attempts …” or the cooldown message), we no longer set checkArticleTitle = true.
  - This prevents the cross-check-and-refetch path from re-triggering WhoColor during a not-ready window.
  - Other error types (like real 404/redirect issues) still allow a one-time title verification.

## Screenshots
Before:
NAN

After:
image


marshadkhn:
Issue number: #6338

Abishekcs:
Please remove [`schema.rb`](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6511/files#diff-cbf8b25d6d853acf58fa9057841f407d29ffe90649c75cf9e3f51b2d6e3aa1d3)

Abishekcs:
There are still changes made to files that are not required please check in `Files changed` tab

Abishekcs:
You can check locally what changes are made in a commit to make sure no unnecessary files are pushed

marshadkhn:
okay, I am a bit confuse here
I ignored the schema.db and there are 3 files showing there, please tell me which file i need to look

Abishekcs:
> I ignored the schema.db

a) Yes, but it was done in the actual `.gitignore` of the codebase file.

b) Also, since your latest changes shows that you have deleted `schema.rb` file itself which will be true locally too.

`Solution`:

a) Simplest fix is to `git reset` to the `commit ID` before your `first commit` and `git commit` only the files that are required then push force your changes.

marshadkhn:
now?
or should i send a fresh PR?

PR #6515, opened 2025-10-28
Upgraded Rails to version 7.1.4 and fixed compatibility issues

DanielChristLight1999:
## What this PR does
This is the first step of an upgrade to the rails version from **7.0.7 -> 7.1.4** which also includes the necessary updates that ensured compatibility with the changes in the new version.

### I made the following key changes:
- Upgraded Rails version from in the `Gemfile` from `7.0.7` to `7.1.4`.
- Removed the deprecated configuration `legacy_connection_handling` according to the [official rails upgrade documentation](https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-0-to-rails-7-1)
- Updated cache format version to:
```
config.active_support.cache_format_version = 7.1
```
to fix deprecation error from previous 6.1 version in `config/application.rb`
- Upgraded PaperTrail gem to version 16.0 for ActiveRecord 7.1 compatibility.
- Updated JavaScript import for `@rails/us` in `app/assets/javascripts/main.js` according to the new import syntax introduced in Rails 7.1.

Note: Some new files were automatically created after running the [update task](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#the-update-task) ` bin/rails app:update`. I ensure that no existing files were overwritten.


Abishekcs:
Please check `Files changed` tab carefully there are many unnecessary changes made like `schema.rb` and many more `files`

app/assets/javascripts/main.js
```diff
@@ -8,9 +8,12 @@
 import 'core-js/stable';
 import 'regenerator-runtime/runtime';
 import i18n from './i18n';
+import Rails from '@rails/ujs';
 
 require('location-origin');
-require('@rails/ujs').start(); // Enables rails-ujs, which adds JavaScript enhancement to some Rails views
+// require('@rails/ujs').start(); // Enables rails-ujs, which adds JavaScript enhancement to some Rails views
```
Abishekcs: Are you sure this is required or is it specific to your local devlepoment?

DanielChristLight1999: Oh yeah I ran into an error after the upgrade and found the fix to be from the official documentation for upgrading 7.0 to 7.1
You can find it in the link below
https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#import-syntax-from-@rails-ujs-is-modified

DanielChristLight1999:
Hello @Abishekcs thank you so much for the pointers. 
I have reversed all the changes that created all the unnecessary files.
I now understand that I shouldn't have run `bin/rails app:update` command as it only just created a bunch of files that are not necessary. I also now understand better that only strictly necessary files should be stages and pushed for a PR.

Abishekcs:
@DanielChristLight1999 CI build is failing

Issue #6294, opened 2025-04-15
Add option for maximum group size for assigned articles


ragesoss:
Courses should be able to specify a maximum group size (stored as a setting in flags), editable in the course Details section. By default, there should be no maximum size.

If a max size is set, a user-facing error response should be returned if a user tries to add an assignment that would increase the size of the group for that article beyond the maximum.

lupamo:
Hi @ragesoss I'll be working on this.

swayam-agrahari:
Hey, I would like to give this a try!

SanjanaSogimatt:
### How I plan on doing this -

### 1. Backend 

- Add max_group_size method to Course model to access flags[:max_group_size]
- Default: nil (unlimited group size)

**When creating an assignment:**
- Check if course has max group size set
- Count existing students assigned to the same article
- Raise MaxGroupSizeExceededError if limit would be exceeded
- Create new exception class MaxGroupSizeExceededError
- Return user-friendly error response (422 status)

### 2. Frontend 

- Add number input field in form for max group size
- Place alongside other course settings (sandbox toggles, etc.)
- Label: "Maximum group size (leave blank for unlimited)"
- Only visible to admins/instructors
- Handle the 422 error response when student tries to join full group
- Display error message to user explaining the limit


@ragesoss sounds good?

ragesoss:
Sounds good!

PR #6517, opened 2025-10-28
Add max group size functionality to assignments

SanjanaSogimatt:
## What this PR does
- Introduced a new flag for courses to set a maximum group size for assignments.
- Updated the `AssignmentsController` to handle `MaxGroupSizeExceededError`.
- Added a method in `AssignmentManager` to check the current group size against the maximum allowed.
- Enhanced the `CoursesController` to update assignment settings based on course flags.
- Implemented UI changes in `Details` component to allow input for maximum group size.
- Added tests to ensure proper handling of group size limits during assignment creation.

Closes - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6294

## Screenshots
Before:

https://github.com/user-attachments/assets/b1ea2994-3819-4617-8c78-696533a33e87

After:

https://github.com/user-attachments/assets/77294c2c-da35-461a-a9c7-f21f025ef6ce

Here, in the above example I am checking with 2 accounts, first profile I have already assigned myself an assignment 

**Edge case:** Already assigned assignments might cause issue because if 3 users are already working and if the max size is set as 2 then that's an issue

Possible solution (not implemented) - Only allow number of users at that moment and above. For example, if 3 users are assigned on a course then the min. user will be able to select will be 3 or more

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -430,10 +430,8 @@ class CoursesController < ApplicationController
 
   def update_assignment_settings
     max_group_size = params.dig(:course, :flags, :max_group_size)
-    peer_review_count = params.dig(:course, :flags, :peer_review_count)
 
     @course.flags[:max_group_size] = max_group_size.to_i if max_group_size.present?
-    @course.flags[:peer_review_count] = peer_review_count.to_i if peer_review_count.present?
 
     @course.save
   end


ragesoss:
Can you explain more about the edge case? If a course has a group of 3 already, and then a course gets updated to have a maximum group size of 2, what will happen?

I think it makes sense to use this feature to prevent group assignments by default for courses that are supposed to have the students work individually. (This can probably be implemented easily in `WizardTimelineManager#add_flags` based on the `working_individually` tag.)

I haven't reviewed the code yet, but in case you didn't already account for this: the group size limit should only apply to Assigned articles (ie, Assignment role 0, as opposed to role 1 which is for peer reviews). A limit for peer review groups would also makes sense, but that would need to be based on the `peer_review_count` flag (and probably doesn't need to be editable via `Details`).

app/assets/javascripts/components/overview/details.jsx
```diff
@@ -90,6 +90,16 @@ const Details = createReactClass({
     return this.props.updateCourse(updatedCourse);
   },
 
+  updateFlags(flagKey, value) {
```
Abishekcs, line 99: `updateFlags` probably leads to `state mutation` you can set `enableReduxSafetyChecks = true` in `create_store.js` then try to set `max_group_size` it will probably throw some type of `state mutation error`.

SanjanaSogimatt: Thanks I missed that

SanjanaSogimatt:
> Can you explain more about the edge case? If a course has a group of 3 already, and then a course gets updated to have a maximum group size of 2, what will happen?
> 
> I think it makes sense to use this feature to prevent group assignments by default for courses that are supposed to have the students work individually. (This can probably be implemented easily in `WizardTimelineManager#add_flags` based on the `working_individually` tag.)
> 
> I haven't reviewed the code yet, but in case you didn't already account for this: the group size limit should only apply to Assigned articles (ie, Assignment role 0, as opposed to role 1 which is for peer reviews). A limit for peer review groups would also makes sense, but that would need to be based on the `peer_review_count` flag (and probably doesn't need to be editable via `Details`).

Well, the case is contradicting the numbers don't line up. I haven't thought of a good solution for this. Also should i add max count it to the peer reviews as well?

ragesoss:
> Well, the case is contradicting the numbers don't line up. 

This just means there will be a group that has larger size than the new setting allows? So it wouldn't be able to get any larger, and if one user is removed, they couldn't be added back? If that's accurate, I think this is fine; that's the behavior I would expect.

> Also should i add max count it to the peer reviews as well?

I'd welcome an additional PR with that. It shouldn't be part of this one.

lib/assignment_manager.rb
```diff
@@ -8,10 +8,10 @@
 class AssignmentManager
   def initialize(course:, user_id: nil, wiki: nil, title: nil, role: nil)
     @course = course
-    @user_id = user_id
+    @user_id = user_id&.to_i
```
ragesoss, line 11: In what situation is this `&.to_i` relevant?

config/locales/en.yml
```diff
@@ -827,6 +828,8 @@ en:
     start: Start
     start_typing: Start typing
     stay_in_sandbox: Stay in sandbox
+    max_group_size: Maximum group size
+    max_group_size_tooltip: Max. number of students who can work on the same article. Leave blank for unlimited.
```
ragesoss: I think "Maximum" is better in this message.

config/locales/en.yml
```diff
@@ -827,6 +828,8 @@ en:
     start: Start
     start_typing: Start typing
     stay_in_sandbox: Stay in sandbox
+    max_group_size: Maximum group size
```
ragesoss: I think "Max group size" is better here.

ragesoss:
@Abishekcs did you have additional review you planned to do, or have you checked this to your satisfaction?

lib/assignment_manager.rb
SanjanaSogimatt: userId and role are always integer right, that's why but seems like it wasn't necessary

Abishekcs:
> @Abishekcs did you have additional review you planned to do, or have you checked this to your satisfaction?

My only concern was `redux mutation` which has been resolved. Rest of the code is okay.

lib/assignment_manager.rb
ragesoss: Given that it has never been needed until now, I think it's not necessary (unless you ran into an error related to it).

SanjanaSogimatt: To be very honest, I did ran into an error that's why i added that which i don't remember what it was and when you mentioned now again i tired it without that and it worked fine 🥲

ragesoss: Okay! Please remove it, and then this should be ready to go, pending final testing by me.

app/controllers/courses_controller.rb
```diff
@@ -427,6 +428,16 @@ def update_last_reviewed
     @course.save
   end
 
+  def update_assignment_settings
+    max_group_size = params.dig(:course, :flags, :max_group_size)
+    peer_review_count = params.dig(:course, :flags, :peer_review_count)
```
ragesoss: Does this do anything? I don't think we have anything for changing peer review count on the frontend.

ragesoss: @SanjanaSogimatt ^

PR #6519, opened 2025-10-28
Fix MySQL error: Field 'article_id' doesn't have a default value (#6470)

### LINKED ISSUE #6470: Research and fix ```Mysql2::Error: Field 'article_id' doesn't have a default value```
### What is happening?
We have several ```Mysql2::Error: Field 'article_id' doesn't have a default value``` logs in Sentry. The error comes from `Sidekiq/CourseDataUpdateWorker`. After deploying #6466 , we have wiki, start and end for the timeslice that produces the error in production. However, I wasn't able to reproduce it locally.

According to docs:
_Mysql2::Error: Field 'article_id' doesn't have a default value usually occurs under strict SQL mode when you try to insert a record into a table without specifying a value for a field that doesn't have a default value set and is not defined as nullable._

The only table where `article_id` is non-nullable is `article_course_timeslices`. Although I don’t have strong evidence (since I wasn’t able to reproduce the error), I suspect it originates from `ArticleCourseTimeslice#update_article_course_timeslices`, in a case where `article_id` is null.

```
      ac_timeslice = ArticleCourseTimeslice.find_or_create_by(course:,
                                                              article_id:,
                                                              start: timeslice.start,
                                                              end: timeslice.end)
```
The `article_id` comes from the array of `RevisionOnMemory`. It can be null if `create_revision` is called with articles set to nil (which shouldn’t happen when updating timeslices), or if the `mw_page_id` cannot be found (though I don’t understand how that could occur).

```
  # Creates a revision record for the given revision data.
  def create_revision(rev_data, article_data, users, articles)
    mw_page_id = rev_data['mw_page_id'].to_i
    RevisionOnMemory.new({
          mw_rev_id: rev_data['mw_rev_id'],
          date: rev_data['date'],
          characters: rev_data['characters'] || 0,
          article_id: articles.nil? ? nil : articles[mw_page_id],
          mw_page_id:,
          user_id: users[rev_data['username']],
          new_article: string_to_boolean(rev_data['new_article']),
          system: string_to_boolean(rev_data['system']),
          wiki_id: rev_data['wiki_id'],
          scoped: article_data['scoped']
        })
  end
```

### Example:
**Course**:: Florida_State_University/Afro-diasporic_Legacies_(Fall)
**wiki_id**: 1
**start**: 2025-10-03 00:00:00 UTC
**end**: 2025-10-04 00:00:00 UTC

That timeslice keeps reprocessing multiple times, but it seems to eventually get fixed since there are no new related logs in Sentry. I’m not sure whether this is connected to the course going through a full update.

I wasn't able to reproduce it locally even after enabling `STRICT_TRANS_TABLES` SQL mode.

### Additional context
According to Sentry, the first occurrence of the ```Field 'article_id' doesn't have a default value``` error on both servers was in release 8904d862751a (created on July 7). I believe PR #6400 was merged in that release, so I wonder if it might be related.

RajPrakash681:
## Problem Description
Fixes #6470 - CourseDataUpdateWorker was failing with `Mysql2::Error: Field 'article_id' doesn't have a default value` in MySQL strict mode when processing timeslices.

### Root Cause
The error occurs when `RevisionOnMemory` objects are created with `nil` article_id because the `mw_page_id` cannot be found in the articles dictionary during revision processing. This specifically affected the Florida_State_University/Afro-diasporic_Legacies_(Fall) course timeslice from 2025-10-03 to 2025-10-04.

## Files Modified
- `app/models/article_course_timeslice.rb`: Added validation and error handling
- `app/services/update_course_wiki_timeslices.rb`: Added filtering and exception handling  
- `lib/revision_data_manager.rb`: Added debug logging
- `spec/models/article_course_timeslice_spec.rb`: Added comprehensive tests
- `spec/services/update_course_wiki_timeslices_spec.rb`: Added edge case tests
- `spec/factories/article_course_timeslices.rb`: Fixed incomplete factory
- `Gemfile`: Updated Ruby version

The multi-layer approach ensures the system gracefully handles cases where article_id is nil while maintaining data integrity and providing comprehensive logging for debugging future issues.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '3.1.7'
+ruby '3.1.2'
 
 ### Basic Framework
 gem 'rails', '7.0.7'


RajPrakash681:
@ragesoss

Abishekcs:
> * Updated Ruby version

Why update ruby?

RajPrakash681:
> > * Updated Ruby version
> 
> Why update ruby?

extremely sorry for this... I updated Ruby locally while setting up the development environment and thought I should sync the Gemfile, but you're right that's not related to the MySQL fix. 

I'm still learning the best practices for contributing - should I revert that change to keep this PR focused only on issue #6470?

Abishekcs:
Yes, please try to resolve only one issue in a PR

RajPrakash681:
> Yes, please try to resolve only one issue in a PR

sure!

Abishekcs:
@RajPrakash681 you should create a different branch and then make a pull request you have made changes in the `master` branch itself

Abishekcs:
https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/CONTRIBUTING.md#before-opening-a-pr

PR #6522, opened 2025-10-29
Upgrade ruby version incrementally from 3.1.2 to 3.2.1

Abdulwahab03:
This PR upgraded Ruby version from 3.1.2 to 3.2.1 incrementally, I tried as much as possible to make sure small gems are updated at a time to avoid any issue that might cause unwanted Issue
Ruby was upgraded from 3.1.2 to 3.2.1 in Gemfile
Bundler gem also got updated from 2.3.22 to 2.7.2, making it 2 gems that was updated everything work perfectly here are the screen short
upgraded ruby -v
This shows when I ran **ruby -v** and **rails -v** command to confirm the version of ruby and rails in use also after the confirmation I ran **rails s** to run the server.
Wiki-dashboard 1 ------> outcome

The next PRs will be aimed to update the remaining gems gradually without any breakage


Abishekcs:
CI build is failing

Abdulwahab03:
@ragesoss and @Abishekcs , I tried to update the gems as small as possible but there are some failures in Ruby RSpec test.

ragesoss:
Try to figure out what is causing the test failures.

Abdulwahab03:
The build failure occurred due to some gems updated from their previous versions, I discovered that some gems do not need to be updated for them to work well with ruby 3.2.1 while some gems have other gems depending on them so updating them will break other gems that depend on them, these are compatibility issues I am trying to solve, I checked the build failure properly and saw what went wrong  and I'm trying to solve them.

PR #6526, opened 2025-10-31
i18n: Fixing eslint warnings caused by hard coded strings

DanielChristLight1999:
## What this PR does
This PR is for updating the hardcoded strings in campaign_ore_plot that causes eslint warnings. It should be the first of my attempts to fix as many significant warnings as I can.

### DIFF
--- a/app/assets/javascripts/components/campaign/campaign_ores_plot.jsx
+++ b/app/assets/javascripts/components/campaign/campaign_ores_plot.jsx
@@ -51,23 +51,17 @@ const CampaignOresPlot = (props) => {
       return (
         <div id="ores" className="ores-plot">
           <CourseQualityProgressGraph graphid={'vega-graph-ores-plot'} graphWidth={1000} graphHeight={400} articleData={filePath} />
-          <p>
-            This graph visualizes, in aggregate, how much articles developed from
-            before campaign participants first edited them until the most recent campagin edits. The <em>Structural Completeness </em>
-            rating is based on a machine learning project (<a href="https://www.mediawiki.org/wiki/ORES/FAQ" target="_blank">ORES</a>)
-            that estimates an article&apos;s quality rating based on the amount of
-            prose, the number of wikilinks, images and section headers, and other features.
-          </p>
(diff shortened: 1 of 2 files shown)


ragesoss:

DanielChristLight1999:
Hello @ragesoss,
I am not so sure what screenshots I might want to include here since the changes will not reflect in the frontend.
I only just fixed the no-literal-strings eslint warning in components/campaign/campaign_ores_plot.jsx.
I apologize for not making it clear in my PR description.

ragesoss:
The screenshot in this case should be of the UI that uses this message, so we can see if there are any unexpected formatting changes or anything like that.

DanielChristLight1999:
Hello @ragesoss, thank you so much for the review. Learned a valuable lesson. Taking the screenshot showed me the html didn't render properly. Below are the screenshots of before and after.

# The graph description:

## Before
Screenshot 2025-11-01 094009

## After
Screenshot 2025-11-01 104929


# The Button 

## Before
Screenshot 2025-11-01 115351

## After
Screenshot 2025-11-01 121740


# Findings

I discovered the following in my quest for the screenshots

1. There is no button or link that leads to ores_plot route anywhere. At least none I have seen.
2. All the plots keep looking the same which may mean an underlying error. I checked even the online version to comfirm this and it is still true.
3. I created a new campaign which should not have any plot and should according to the current code render "No Structural Completeness data available" when the button is clicked but that didn't happen. It instead rendered the graph description when the button is clicked. I have not been able to confirm this in the online version since I don't know how to simulate this without creating a real campaign online which may just be wrong and wasteful (I'm not sure).
I have attached a screen record to illustrate this finding. The code can be found here https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/app/assets/javascripts/components/campaign/campaign_ores_plot.jsx

https://github.com/user-attachments/assets/bece5a41-a15d-463a-9260-795b66a7c13d

ragesoss:
Thanks! I believe that we removed the links as a way to basically disable that feature, when we changed the way we handle article quality data and stopped storing individual revision. That's why the plot doesn't show anything useful. It's a feature we might or might not find a way to re-enable in the future.

But it looks like your i18n fix is in good shape!

PR #6535, opened 2025-11-01
Feat: Filter pageviews to count created articles only

### LINKED ISSUE #6534: Stats: Pageviews are inflated by edits to existing articles
**Context:** 
The current "Pageviews" statistic on the dashboard is very helpful, but it appears to aggregate views from all articles impacted by a program. This includes both newly created articles and existing articles that were edited.

**Current Behavior:** 
The "Pageviews" metric sums the views from:
1. Articles newly created by participants.
2. Existing articles that were edited by participants.

**Expected Behavior:** 
There should be a way to view a "Pageviews" statistic that is filtered to only include views from articles that were newly created by participants.

**Why this is important:** 
For organizations using the dashboard to track the impact of new content creation, the current, combined statistic is "inflated." It makes it difficult to isolate the specific impact of new articles versus the (often much larger) view counts from minor edits to high-traffic existing pages.

Providing a filtered "created articles only" pageview count would offer a much clearer and more accurate metric for this specific use case.

I am currently working on a patch to implement this filter and will be opening a Pull Request shortly to address this.

satrijo:
Closes #6534 

## What this PR does
This Pull Request modifies the main pageview statistics calculation. It filters the results to only include pageviews from articles newly created by program participants.

This changes the previous behavior, which included views from all edited articles (both new and existing).

## Screenshots
Before:
screenshot-2025-11-01_19-24-33

After:
image

### DIFF
--- a/app/assets/javascripts/components/overview/overview_stats.jsx
+++ b/app/assets/javascripts/components/overview/overview_stats.jsx
@@ -13,6 +13,34 @@ const OverviewStats = ({ course }) => {
     return course.newStats[stat] ? 'stat-display__value stat-change' : 'stat-display__value';
   };
 
+  const buildPageviewsBreakdown = (courseData) => {
+    // Check if breakdown data is available
+    const hasBreakdown = (courseData.view_count_created !== undefined
+      && courseData.view_count_created !== null
+      && courseData.view_count_existing !== undefined
+      && courseData.view_count_existing !== null);
+
+    // If breakdown data is not available, fall back to simple doc
+    if (!hasBreakdown) {
+      return I18n.t(`metrics.${ArticleUtils.projectSuffix(courseData.home_wiki.project, 'view_count_doc')}`);
+    }
+
+    // Build breakdown array using the readable values from backend
+    const breakdown = [
+      [
+        courseData.view_count_created,
+        I18n.t('metrics.pageviews_breakdown.new_articles')
+      ],
+      [
+        courseData.view_count_existing,
+        I18n.t('metrics.pageviews_breakdown.existing_articles')
+      ]
+    ];
+
+    // Return breakdown array (OverviewStatInfo will handle rendering)
+    return breakdown;
+  };
+
   let editedLabel;
   let createdLabel;
   if (isWikidata) {
@@ -140,7 +168,7 @@ const OverviewStats = ({ course }) => {
         stat={course.view_count}
         statMsg={I18n.t(`metrics.${ArticleUtils.projectSuffix(course.home_wiki.project, 'view_count_description')}`)}
         renderZero={false}
-        info={I18n.t(`metrics.${ArticleUtils.projectSuffix(course.home_wiki.project, 'view_count_doc')}`)}
+        info={buildPageviewsBreakdown(course)}
         infoId="view-count-info"
       />
       {uploadCount}
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -13,6 +13,7 @@
 #  term                  :string(255)
 #  character_sum         :integer          default(0)
 #  view_sum              :bigint           default(0)
+#  view_sum_created      :bigint           default(0)
 #  user_count            :integer          default(0)
 #  article_count         :integer          default(0)
 #  revision_count        :integer          default(0)
@@ -612,6 +613,27 @@ class Course < ApplicationRecord
 
   def update_cache_from_timeslices
     CourseCacheManager.new(self).update_cache_from_timeslices course_wiki_timeslices
+  end
+
+  # Returns the cached estimate of cumulative pageviews for articles that were
+  # created by participants in this course. This value is updated by
+  # CourseCacheManager during course cache updates, not calculated on-the-fly
+  # for performance reasons (courses can have up to 100K articles).
+  # The calculation mirrors view_sum but filters for tracked, live, created articles only.
+  #
+  # For courses that have been updated but view_sum_created is still 0 (likely
+  # because it wasn't populated during the first update after migration), we calculate
+  # on-the-fly. For historical courses that haven't been updated, we return the cached
+  # value to avoid performance issues.
+  def view_sum_created
+    cached_value = self[:view_sum_created] || 0
+    # If course has been updated but view_sum_created is 0 and we have new articles,
+    # calculate on-the-fly (this handles courses updated before view_sum_created was added)
+    if cached_value.zero? && was_course_ever_updated? && new_article_count.positive?
+      CourseCacheManager.calculate_view_sum_created_for_course(self)
+    else
+      cached_value
+    end
   end
 
   # Ensures weeks for a course have order 1..weeks.count
--- a/app/views/courses/course.json.jbuilder
+++ b/app/views/courses/course.json.jbuilder
@@ -54,6 +54,17 @@ json.course do
   json.word_count number_to_human @course.word_count
   json.references_count number_to_human @course.references_count
   json.view_count number_to_human @course.view_sum
+  # Only include breakdown if:
+  # 1. Course has been updated (timeslice_update_ran? or was_course_ever_updated?)
+  #    OR course has valid cached data (view_sum > 0 indicates it was updated at some point)
+  # 2. view_sum_created doesn't exceed view_sum (to avoid negative existing count)
+  # Note: view_sum_created can be 0 (meaning all views are from existing articles)
+  course_updated = @course.timeslice_update_ran? || @course.was_course_ever_updated?
+  has_valid_cache = @course.view_sum.positive? # If view_sum > 0, course was updated at some point
+  if (course_updated || has_valid_cache) && @course.view_sum_created <= @course.view_sum
+    json.view_count_created number_to_human @course.view_sum_created
+    json.view_count_existing number_to_human(@course.view_sum - @course.view_sum_created)
+  end
   json.character_sum_human number_to_human @course.character_sum
   json.syllabus @course.syllabus.url if @course.syllabus.file?
   json.updates average_delay: @course.flags['average_update_delay'],
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1211,8 +1211,11 @@ en:
       other: total usages across languages
     view_count_description: Article Views
     view_count_description_wikidata: Item Views
-    view_count_doc: This is the estimated number of views based on a 30-day average for each article, through the most recent stats update. Views may decrease if the updated average is lower than previous counts.
-    view_count_doc_wikidata: This is the estimated number of views based on a 30-day average for each item, through the most recent stats update. Views may decrease if the updated average is lower than previous counts.
+    view_count_doc: Estimated total views calculated from each article's first revision during this course up to the present. The daily average is computed from all daily view counts during this period, and the total is estimated as (days from first revision to present) × daily average. Each course calculates its own average for the same article based on when it was first edited in that course.
+    view_count_doc_wikidata: Estimated total views calculated from each item's first revision during this course up to the present. The daily average is computed from all daily view counts during this period, and the total is estimated as (days from first revision to present) × daily average. Each course calculates its own average for the same item based on when it was first edited in that course.
+    pageviews_breakdown:
+      new_articles: "from new articles"
+      existing_articles: "from existing articles"
     view_data_unavailable: No view data available
     view: Views
     wiki_ed_help: Use the 'Get Help' button to report a problem.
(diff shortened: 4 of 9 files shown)


Abishekcs:
@satrijo Some of the changes made are not related to this PR, including the commits, I think.

satrijo:
> @satrijo Some of the changes made are not related to this PR, including the commits, I think.

Hi @Abishekcs 

Thank you for the feedback. You were absolutely right. I had incorrectly merged master into this branch, which pulled in those unrelated commits.

I have just cleaned up the branch history. I've rebased it onto the latest master and force-pushed the changes to update this PR.

The commit history should now be clean and only contain the relevant commits for this issue.

Could you please take another look when you have a moment?

Thanks!

ragesoss:
We haven't had a significant amount of requests for this as a separate stat, so I'm not sure it makes sense to add it has a top-level stat in the UI like this. It might be more appropriate to add as additional detail that shows up upon mouseover, as we do with 'Commons Uploads' and 'Editors'.

satrijo:
> We haven't had a significant amount of requests for this as a separate stat, so I'm not sure it makes sense to add it has a top-level stat in the UI like this. It might be more appropriate to add as additional detail that shows up upon mouseover, as we do with 'Commons Uploads' and 'Editors'.

Hi, @ragesoss 

Thanks again for the excellent feedback. I've just pushed an update that implements your suggestion.
The main "Pageviews" stat now shows the original, combined total.

I've added the filtered "New Article Pageviews" count to a mouseover pop-up, following the pattern used by the "Editors" stat.
I've also updated the screenshots in the PR description to show the new behavior.

screenshot-2025-11-04_13-47-30


Could you please take another look when you have a chance?

Thanks!

app/views/courses/course.json.jbuilder
```diff
@@ -54,6 +54,10 @@ json.course do
   json.word_count number_to_human @course.word_count
   json.references_count number_to_human @course.references_count
   json.view_count number_to_human @course.view_sum
+  json.view_count_created number_to_human @course.view_sum_created
+  # Raw numbers for breakdown calculation
+  json.view_sum_raw @course.view_sum
+  json.view_sum_created_raw @course.view_sum_created
```
gabina: json field names are a bit confusing. If `view_count` and `view_sum` mean the same (which seems to be the case), I’d suggest keeping only `view_count` for consistency.

I noticed you added a `view_count_created` field to show the readable version of views for new articles, but it doesn’t look like it’s actually used anywhere, is that right?

I think understand why you’re sending the raw values, but IMHO it’d be fine to sacrifice a bit of accuracy here and send three readable values instead: `view_count`, `view_count_created`, and `view_count_existing` (where the last one could just be `view_count - view_count_created`). Views are always approximations anyway, and this approach would simplify the frontend logic and remove the need for that format handling there.

app/models/course.rb
```diff
@@ -614,6 +614,28 @@ def update_cache_from_timeslices
...
+      .sum(
+        Arel.sql(
+          <<~SQL.squish
+            FLOOR(
+              DATEDIFF(UTC_TIMESTAMP(), articles_courses.first_revision) * articles_courses.average_views
+            )
+          SQL
+        )
+      )
+  end
```
gabina: Will this query run every time `course.json` is requested? I have a concern about performance here. Some courses can have a large number of articles (potentially around 100K), so it would be nice to make sure this runs efficiently enough.

gabina:
It would be nice to add some backend specs and include a screenshot showing a non-empty example.

app/models/course.rb
satrijo: I added caching, similar to view_sum:

- Added a view_sum_created database column
- Updated CourseCacheManager to cache the value during course updates
- The query now runs only during background updates, not on every request

app/views/courses/course.json.jbuilder
satrijo: You were right, the JSON was confusing. I've simplified this based on your suggestion:

I removed the _raw value fields (view_sum_raw, view_sum_created_raw) from course.json.jbuilder.

I added the new pre-calculated, human-readable value view_count_existing directly from the backend.

The frontend (overview_stats.jsx) is now much simpler. It just displays the three values (view_count, view_count_created, view_count_existing) passed from the API and does no formatting or math for the tooltip.

satrijo:
> It would be nice to add some backend specs and include a screenshot showing a non-empty example.

This is done.

I've added 8 new backend specs in total:

- 2 specs in spec/lib/course_cache_manager_spec.rb to test the new caching logic.
- 6 specs in spec/models/course_spec.rb to test the view_sum_created method (covering the cache, the fallback logic, and edge cases).
- All tests are passing ✅.
- I have also updated the PR description with a new screenshot showing a non-empty example (815K total, broken down into 143K new / 672K existing).

Thanks again for the guidance! Let me know if this looks better.

app/models/course.rb
gabina: Ok, that approach involves a DB migration. I'm not sure if that's something we want to do considering we haven't had a significant amount of requests for this. I would wait for @ragesoss's opinion.

ragesoss: I'm not opposed it it, if the implementation overall is good.

gabina:
@satrijo please fix the linting and once CI is green, I'll review again

satrijo:
> @satrijo please fix the linting and once CI is green, I'll review again

Hi @gabina 

I've just pushed my final set of changes. This includes the new backend specs and all the linting fixes.

All CI checks (including specs and linting) are now passing (green) ✅.

Thanks!

app/assets/javascripts/components/overview/overview_stats.jsx
```diff
@@ -13,6 +13,34 @@ const OverviewStats = ({ course }) => {
...
   };
 
+  const buildPageviewsBreakdown = (courseData) => {
+    // Check if breakdown data is available
+    const hasBreakdown = (courseData.view_count_created !== undefined
+      && courseData.view_count_created !== null
+      && courseData.view_count_existing !== undefined
+      && courseData.view_count_existing !== null);
+
+    // If breakdown data is not available, fall back to simple doc
```
gabina, line 23: Do you mind explaining me in which cases the breakdown data wouldn't be available?

app/assets/javascripts/components/overview/overview_stats.jsx
```diff
@@ -13,6 +13,34 @@ const OverviewStats = ({ course }) => {
...
+    if (!hasBreakdown) {
+      return I18n.t(`metrics.${ArticleUtils.projectSuffix(courseData.home_wiki.project, 'view_count_doc')}`);
+    }
```
gabina, line 26: It seems like a strange design decision that in some cases you see a doc explaining how views are calculated, while in others you see only a breakdown without the explanation. However, as issue #6548 indicates, the view count doc is outdated, so maybe it’s fine not to show it.

app/models/course.rb
```diff
@@ -614,6 +615,25 @@ def update_cache_from_timeslices
...
+    cached_value = self[:view_sum_created] || 0
+    # If view_sum > 0 but view_sum_created is 0, it's likely the cache hasn't been
+    # populated yet (unless truly all views are from existing articles).
+    # Fallback to calculation in this case for backward compatibility.
+    if cached_value.zero? && view_sum.positive? && articles_courses.tracked.live.new_article.exists?
+      calculate_view_sum_created
+    else
+      cached_value
+    end
+  end
```
gabina, line 637: I still don't know if this is performant for historical courses with many articles. Maybe it's OK to just avoid showing a breakdown for historical courses where we never populated the new `view_sum_created` field.

Also, isn’t the number we show as “article views” (`view_count`) a static value from the database? If that’s the case, then for historical courses without recent updates, we won’t show an updated general view count. I don’t think it’s worth showing a dynamically calculated view count for new articles, since that would result in a mix of outdated and updated data.

Additionally, note that due to how we calculate `view_count_existing`, it could be a negative number if `view_sum` is quite old and we keep recalculating the current `view_sum_created` value:

`json.view_count_existing number_to_human(@course.view_sum - @course.view_sum_created)`

Let me kwnow if I'm missing something here.

app/models/course.rb
```diff
@@ -632,6 +652,24 @@ def ensure_home_wiki_in_courses_wikis
...
+        Arel.sql(
+          <<~SQL.squish
+            FLOOR(
+              DATEDIFF(UTC_TIMESTAMP(), articles_courses.first_revision) * articles_courses.average_views
+            )
+          SQL
+        )
+      )
+  end
+
```
gabina: Is this query the same as the one in `update_view_sum_created_based_on_first_revision`? If so, we should avoid the duplication.

config/locales/en.yml
```diff
@@ -1195,8 +1195,13 @@ en:
       other: total usages across languages
     view_count_description: Article Views
     view_count_description_wikidata: Item Views
+    view_count_created: "Pageviews (created articles)"
+    view_count_created_doc: "Estimated pageviews only for articles created by participants."
```
gabina: Where do these strings show up in the app? You should add a screenshot for them

app/assets/javascripts/components/overview/overview_stats.jsx
satrijo: Breakdown data (`view_count_created and view_count_existing`) is not available in the following cases:

- Historical courses that haven't been updated since the migration: Courses that existed before the view_sum_created column was added and haven't been updated since then will have view_sum_created = 0 (the default). For these courses, we intentionally don't calculate the breakdown on-the-fly to:

-- Avoid performance issues (courses can have 100K+ articles)

-- Maintain consistency (since view_sum is a static cached value, mixing it with a dynamically calculated view_sum_created would be inconsistent)

- Courses with no new articles: If a course truly has no new articles created by participants, view_sum_created will be 0, and the breakdown would show "0 from new articles" and all views "from existing articles". However, we still show the breakdown in this case.

- Data inconsistency edge case: If view_sum_created somehow exceeds view_sum (which shouldn't happen but could due to timing issues or data inconsistencies), we skip the breakdown to avoid showing negative values for "existing articles".

The frontend checks for undefined or null values, which will occur when the backend doesn't include these fields in the JSON response (see `course.json.jbuilder `lines 59-62).

app/assets/javascripts/components/overview/overview_stats.jsx
satrijo: You're right that this is inconsistent. The current implementation shows:

- **With breakdown**: Only the breakdown (no explanation of how views are calculated)
- **Without breakdown**: Only the doc (explanation of how views are calculated)

As you noted, issue #6548 indicates the view count doc is outdated. The breakdown provides more useful information (showing the split between new and existing articles), so showing only the breakdown when available seems reasonable. However, if you'd prefer, we could show both the breakdown AND the doc when breakdown is available.

app/models/course.rb
satrijo: You're absolutely right about the performance concerns. I've implemented a balanced approach:

- **Historical courses that haven't been updated**: Return the cached value (which defaults to 0) without on-the-fly calculation to avoid performance issues
- **Courses that have been updated but `view_sum_created` is still 0**: Calculate on-the-fly only if the course has been updated (`was_course_ever_updated?`) and has new articles. This handles courses that were updated before the `view_sum_created` column was added
- **Courses with recent updates**: Show the breakdown using cached values that were calculated during the update cycle

This approach:

- Avoids performance issues for historical courses that haven't been updated
- Provides accurate breakdowns for courses that have been updated but `view_sum_created` wasn't populated during the first update after migration
- Maintains consistency between `view_sum` and `view_sum_created` for courses that have been properly updated

app/models/course.rb
satrijo: I've refactored to eliminate the duplication:

- Created a shared class method `CourseCacheManager.calculate_view_sum_created_for_course(course)` that contains the query logic
- Both `CourseCacheManager#update_view_sum_created_based_on_first_revision` and the removed `Course#calculate_view_sum_created` now use this shared method
- This ensures consistency and makes maintenance easier

**Files changed:**

- `lib/course_cache_manager.rb`: Added shared class method `calculate_view_sum_created_for_course`
- `app/models/course.rb`: Removed duplicate `calculate_view_sum_created` method, now uses the shared method via fallback logic

satrijo:
Hi @gabina, just wanted to check in on this. I’ve resolved the bugs and ensured all tests are passing. Is there anything else needed from my side to move this forward? Thanks!

gabina:
Hello!

Do you mind responding to the review comments I left? If you already addressed a change, please mention that. If you decided not to address a change and have reasons for it, please explain those. That usually helps make the review process more fluent.

PR #6539, opened 2025-11-03
Improve page_type classifier: add draft and mainspace

gabina:
## What this PR does
This PR adds draft and mainspace as new page types:

- If the article title starts with "Draft:", then it's classified as draft
- If the article title does not contain a colon (:), is classified as mainspace

## Screenshots
Before:
image

After:
image

Namespace classifications like draft, mainspace, userspace may be implemented through the Article record associated to the alert `article_id`. That strategy should be 100% accurate.

### DIFF
--- a/app/models/alert_types/ai_edit_alert.rb
+++ b/app/models/alert_types/ai_edit_alert.rb
@@ -168,7 +168,7 @@ class AiEditAlert < Alert
     details[:prior_alert_for_user]
   end
 
-  def page_type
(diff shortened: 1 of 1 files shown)


ragesoss:
> Namespace classifications like draft, mainspace, userspace may be implemented through the Article record associated to the alert article_id. That strategy should be 100% accurate.

The only concern with that strategy is that it requires loading the Article record, so it might have performance implications. (This is probably not a significant concern, but if we do use that strategy, we'll need to account for it and avoid N+1 queries.)

Issue #6533, opened 2025-10-31
CheckRevisionWithPangram does not handle missing revisions gracefully


ragesoss:
### What is happening?
If a revision is missing on Wikipedia (eg, because it was deleted) then `CheckRevisionWithPangram` does not handle the resulting API error, and that job will get periodically retried (with progressively longer waits between retries) in Sidekiq.

### Expected behavior
If `CheckRevisionWithPangram` can't run for a particular revision that isn't available, it should handle it gracefully, logging the failure to Sentry in an  understandable way and not retrying.

Ajay-singh1:
Picking Up.

PR #6541, opened 2025-11-03
Stop Sidekiq from running when badrevid is returned by Wikipedia API

Ajay-singh1:

## What this PR does
This PR logs message to sentry when bad revision id is returned by the Wikipedia API and immediately stop Sidekiq to re-run the jobs.

Fixes: #6533

### DIFF
--- a/spec/services/check_revision_with_pangram_spec.rb
+++ b/spec/services/check_revision_with_pangram_spec.rb
@@ -83,21 +83,17 @@ describe CheckRevisionWithPangram do
   end
 
   context 'when the revision is missing or deleted' do
-    let(:missing_revision_id) { 123456789 }
+    let(:missing_revision_id) { 999999999999 }
 
     it 'logs a message to Sentry and exits gracefully' do
-      allow_any_instance_of(WikiApi).to receive(:query).and_return(
-        OpenStruct.new(data: { 'badrevids' => { missing_revision_id.to_s => {
-'revid' => missing_revision_id, 'missing' => ''
-} } })
-      )
-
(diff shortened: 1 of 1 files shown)


Ajay-singh1:
@ragesoss I can't reproduce the bug locally but I guess this thing might solve the bug I am not entirely sure.The approach is something like this if ```badrevid``` is returned by the Wikipedia API nothing gets returned to the method and the message is logged on to Sentry.

ragesoss:
You should be able to replicate it by writing tests that call `CheckRevisionWithPangram` with revision IDs that a) don't exist, or b) have been deleted. We have some existing specs that test other parts of system behavior for deleted revisions, like `reference_counter_api_spec.rb`, which you can probably use to find the id of a deleted revision to test against.

Ajay-singh1:
On it.

Ajay-singh1:
@ragesoss Up for review.

spec/services/check_revision_with_pangram_spec.rb
```diff
@@ -81,4 +81,23 @@
       expect(AiEditAlert.last.article_id).to eq(live_article.id)
     end
   end
+
+  context 'when the revision is missing or deleted' do
+    let(:missing_revision_id) { 123456789 }
+
+    it 'logs a message to Sentry and exits gracefully' do
+      allow_any_instance_of(WikiApi).to receive(:query).and_return(
```
ragesoss: I would rather use VCR here instead of mocking the API behavior.

Ajay-singh1:
Thanks @ragesoss. It is much cleaner now.

PR #6546, opened 2025-11-04
Fix: Fixed page view specs failure in wiki_pageviews_spec

### LINKED ISSUE #6542: Pageview-related specs fail intermittently on CI
Lately, we've been have a fair number of CI build failures because of specs that involve pageviews failing.

This is probably because of 429 errors or other issues related to API behavior. The specs should be made more robust, either by mocking the required API calls or adding the correct responses as cached fixtures that are committed to the repo. The latter is probably the better strategy in this case. (We do this already one fixture file; see https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6153.)

Here's an example of a build that failed because of the pageview specs: https://github.com/WikiEducationFoundation/WikiEduDashboard/actions/runs/18985962701/job/54229542619

DanielChristLight1999:
## What this PR does
This PR fixes the rspec failures in spec/lib/wiki_pageviews_spec.rb that occurs intermittently on CI due to API errors. I have cached the correct responses as fixtures in fixtures/vcr_cassetes/cached/wiki_pageviews

Fixes #6542

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@
 /config/newrelic.yml
 /config/secrets.yml
 /dockerAuth.json
-/fixtures
 /node_modules
 /js_coverage
 /vendor/*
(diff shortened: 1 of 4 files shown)


ragesoss:
Nice. This looks like it will work, thanks!

PR #6547, opened 2025-11-05
fixed only mainspace title should be allowed

### LINKED ISSUE #6531: When assigning an article for a ClassroomProgramCourse, only mainspace titles should be allowed
Currently, it's possible to add an assigned article that isn't in mainspace, and this breaks some of the features that expect a mainspace title (and is just not a supported way to do things).

At last for the ClassroomProgramCourse type, the assign UI for editing articles should check whether the input is in a different namespace or is otherwise not a valid mainspace title, and throw an error message rather than submitting such an assignment.

For example, entering a URL for a category page (or the full title including Category: prefix) should result in the error message: `https://en.wikipedia.org/wiki/Category:1993_in_sports_in_Alberta`

DanielChristLight1999:
## What this PR does
This PR fixes the issue of assigning articles that are not in mainspace. This fix checks if the title being added is valid and in the mainspace and if not raises an error.

Fixes #6531

## Screenshots
Before:

https://github.com/user-attachments/assets/f5920004-c8b3-42ea-b71a-aa28b835b53c

After:

https://github.com/user-attachments/assets/41914df6-24dc-4710-9af2-0a0697ac75ce

### DIFF
--- a/app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
+++ b/app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
@@ -14,6 +14,8 @@ import { ASSIGNED_ROLE, REVIEWING_ROLE } from '~/app/assets/javascripts/constant
 import SelectedWikiOption from '../selected_wiki_option';
 import { trackedWikisMaker } from '../../../utils/wiki_utils';
 import ArticleUtils from '../../../utils/article_utils';
+import { verifyMainSpaceArticle } from '@actions/article_actions.js';
+import { addNotification } from '@actions/notification_actions.js';
 
 
 // Helper Components
@@ -306,28 +308,44 @@ const AssignButton = ({ course, role, course_id, wikidataLabels = {}, hideAssign
     return dispatch(initiateConfirm({ confirmMessage, onConfirm, warningMessage }));
   };
 
-  const assign = (e) => {
+  const assign = async (e) => {
     e.preventDefault();
-    title.split('\n').filter(Boolean).forEach((assignment_title) => {
-      const assignment = {
-        title: decodeURIComponent(assignment_title).trim(),
-        project,
-        language,
-        course_slug: course.slug,
-        role
-      };
+    // for (const assignment_title of title.split('\n').filter(Boolean))
+    const lines = title.split('\n').filter(Boolean);
 
-      if (!assignment.title || assignment.title === 'undefined') return;
-      if (assignment.title.length > 255) {
-        // Title shouldn't exceed 255 chars to prevent mysql errors
-        return alert(I18n.t('assignments.title_too_large'));
-      }
-      const studentId = (student && student.id) || null;
-      dispatch(addAssignment({
-        ...assignment,
-        user_id: studentId
-      }));
-    });
+    await Promise.all(
+      lines.map(async (assignment_title) => {
+        const assignment = {
+          title: decodeURIComponent(assignment_title).trim(),
+          project,
+          language,
+          course_slug: course.slug,
+          role
+        };
+
+        if (!assignment.title || assignment.title === 'undefined') return;
+        if (assignment.title.length > 255) {
+          // Title shouldn't exceed 255 chars to prevent mysql errors
+          return alert(I18n.t('assignments.title_too_large'));
+        }
+        const studentId = (student && student.id) || null;
+
+        if (course.type === 'ClassroomProgramCourse') {
+          const result = await dispatch(verifyMainSpaceArticle(assignment.title, assignment.language, assignment.project));
+          if (!result?.valid) {
+            return dispatch(addNotification({
+              message: result?.error,
+              type: 'error',
+              closable: true
+            }));
+          }
+        }
+        dispatch(addAssignment({
+          ...assignment,
+          user_id: studentId
+        }));
+      })
+    );
   };
 
   const review = (e, assignment) => {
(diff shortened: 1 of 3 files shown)


Abishekcs:
Instead of `Window alert()` you can use existing `notifications.jsx`

DanielChristLight1999:
Oh awesome thank you @Abishekcs let me do that then.

DanielChristLight1999:
Hi @Abishekcs and @ragesoss 
The CI builds keep failing the same particular tests but I have run those tests locally with my changes and they pass. Any idea what I could do? Here is screenshot of the test on my computer.

Screenshot 2025-11-05 220053

ragesoss:
Hmm... not sure. Did you double check that you ran them after doing `yarn build` to compile the changed versions of the JS files? That would be one reason why they might pass locally but fail in CI.

DanielChristLight1999:
Okay so after following your advice @ragesoss, I found the solution and the real problem too. My local tests were passing because after switching from the master branch to the fixing branch, I didn't run yarn build and that makes the tests pass, making me think my code works perfectly when it actually doesn't.

Now for the problem, my code queries the wikipedia api with the title of the article, I then check if it's in mainspace and if it is not, it should return an error message and stop the assignment. This is what I think stops the tests from passing, since after the assign button is clicked on during the test, it doesn't run the the new api request I wrote for checking if an article is in mainspace.

My question now is if there is any other case like this perhaps an issue or PR from before I could learn the solution from. The only solution I can think of now is to somehow detect in my code when it is a test running and then just skip the article mainspace check if it is a test or we are in a test environment. I have tried using if (process.env.NODE_ENV === "test") but that doesn't work although I didn't exepect it too. I have attached a screenshot of the action I am using to do the check.
Any ideas? @ragesoss @Abishekcs 
Screenshot 2025-11-06 082418

ragesoss:
Those API calls should still be made in the tests. I suggest running the tests without headless mode, add some `sleep` statements to the test, and open the browser console while the tests are running, so you can debug exactly what's happening when the tests fail.

DanielChristLight1999:
Alright awesome. Thank you.

DanielChristLight1999:
Hello @ragesoss after following your suggestion, I confirmed that somehow the network doesn't show the api call happening. So I decided to write an endpoint in `controllers/lookups_controller.rb` ran the test and it all passed. I also created an rspec test for the new endpoint and have tested them to all pass.
In summary, I have made changes to the following:
- `controllers/lookups_controller.rb`: I added an endpoint that uses the WikiApi to validate an article exists in the mainspace.
- `config/routes.rb`: added the route to the endpoint
- `spec/controllers/lookups_controller_spec.rb`: created rspec tests for the new endpoint.
- `actions/article_actions.js`: added an actioin that calls the api endpoint and checks if the article is in mainspace.
- `components/common/AssignCell/AssignButton.jsx`: updated assign function to check if article is in mainspace first.
I've attached a screenshot of the new test passing. Please let me know if you need any other screenshots or information on my changes. Thank you :)

Screenshot 2025-11-07 003236

DanielChristLight1999:
Hello @ragesoss and @Abishekcs trust you are doing fine.
I think I am ready for review now.

ragesoss:
I don't want to have the API calls done server-side. It should be possible to make them work (and including in tests) without going through the rails server, along the lines that you tried initially.

DanielChristLight1999:
alright thank you for the review. I'm on it.

DanielChristLight1999:
Hello @ragesoss I have fixed the code successfully. After observing the tests run several times on the browser, I realized that the articles actually get assigned successfully during the tests but somehow capybara is so fast that it doesn't wait for the assignment to complete before it reads the UI. So I added some wait properties to the test expectations to allow capybara detect the assigned articles before concluding the test.

DanielChristLight1999:
> Hello @ragesoss I have fixed the code successfully. After observing the tests run several times on the browser, I realized that the articles actually get assigned successfully during the tests but somehow capybara is so fast that it doesn't wait for the assignment to complete before it reads the UI. So I added some wait properties to the test expectations to allow capybara detect the assigned articles before concluding the test.

In a search to minimize the number of changes made to absolutely necessary, I discovered that all I needed was to add a line that expects the assignment link to be visible after the assign button is clicked.

ragesoss:
Great! It can be tricky to understand how to structure capybara tests so that you don't have to rely on `sleep` statements.

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
```diff
@@ -306,28 +308,44 @@ const AssignButton = ({ course, role, course_id, wikidataLabels = {}, hideAssign
...
+
+        if (!assignment.title || assignment.title === 'undefined') return;
+        if (assignment.title.length > 255) {
+          // Title shouldn't exceed 255 chars to prevent mysql errors
+          return alert(I18n.t('assignments.title_too_large'));
+        }
+        const studentId = (student && student.id) || null;
+        const result = await dispatch(verifyMainSpaceArticle(assignment.title, assignment.language, assignment.project));
+
+        if (course.type === 'ClassroomProgramCourse') {
```
ragesoss, line 333: The check for mainspace-only should only be performed in the first place if it's a ClassroomProgramCourse. No behavior (including API calls) should be different in other cases.

DanielChristLight1999: Oh sorry. That was a real rookie mistake. Just fixed it now.

Formasitchijoh:
@DanielChristLight1999  I am looking at working on this [issue](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6551) but I see it depends on this task. Just wish to ask if it is ready to be merged or maybe you need some help

PR #6558, opened 2025-11-06
Add table of courses with more than 3 AI edit alerts to AI edit alerts view

gabina:
## What this PR does
Closes #6552 

The course table is created with courses from the default campaign that have more than three AI edit alerts (not necessarily recent alerts).

## Screenshots
Before:
image

After:
It includes the course table in the table of content and adds the course table.

image

PR #6559, opened 2025-11-07
Add sort functionality to AI admin view

gabina:
## What this PR does
This PR adds a minor sort functionality by "number of alerts in mainspace" and "number of users involved" columns in "Courses with AI alerts table" as it was requested [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6558#issuecomment-3499843599).

## AI usage
I used ChatGPT to have a first version of `CoursesWithAiAlertsList` component. I modified that first version to make it simpler. I explicitly asked it to create a component without using global Redux state, because I thought it was unnecessary, and that was how sorting was implemented with the `List` component in most of the app (according to what I understand).

Sometimes I also use ChatGPT to improve my PR descriptions, but I always write the description myself first and then ask to improve collocations or errors. I always try to keep my writing style.

## Screenshots
Before:
Just a table without interaction allowed.
image

After:
[Screencast from 2025-11-08 14-12-39.webm](https://github.com/user-attachments/assets/9cf54823-7a58-48fa-8fba-fe6cc3735d0e)

### DIFF
--- a/app/assets/javascripts/components/alerts/courses_with_ai_alerts_list.jsx
+++ b/app/assets/javascripts/components/alerts/courses_with_ai_alerts_list.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import List from '../common/list.jsx';
 
 const CoursesWithAiAlertsList = ({ stats }) => {
-  const [sortConfig, setSortConfig] = useState({ key: null, direction: 'asc' });
+  const [sortConfig, setSortConfig] = useState({ key: null, direction: null });
 
   const courses = Object.values(stats);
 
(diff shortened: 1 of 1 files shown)


ragesoss:
The first click should sort it high-to-low, as the most common use case will be to see the highest-count courses.

ragesoss:
The sorting has a weird bug. One course at the top isn't sorted correctly, and each time the table is re-sorted, a duplicate entry for it gets added. There are two courses in production with the same course title (but different slugs), and it's these courses that are being duplicated.

After first sort:
first sort

After many sorts:
sorting bug

Issue #6554, opened 2025-11-06
On Programs & Events Dashboard, user profile should link to global contributions.


ragesoss:
See https://phabricator.wikimedia.org/T387895#11348844

This only applies to Programs & Events Dashboard. In Wiki Education mode, the behavior should not change.

GitaHub71:
I'm trying this one

PR #6561, opened 2025-11-11
Add global contribution url to User Profile

GitaHub71:
Add global contribution to user profile on dashboard

## What this PR does
1. Add a parameter to contribution_link on users_helper to identify which request wants global user contribution.
2. Add a condition to call global_contribution_link
3. Modify Rspec according to current changes

## Screenshots
Before:
[Screencast from 11-11-25 16:35:31.webm](https://github.com/user-attachments/assets/d499ab68-625b-48ca-9030-fba1fafa55d7)

After:
[Screencast from 11-11-25 16:36:35.webm](https://github.com/user-attachments/assets/3af2febd-5c36-4200-8eb0-160ca4f9d3be)

Fixes: #6554

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -4,7 +4,11 @@
 module UsersHelper
   def contribution_link(courses_user, text = nil, css_class = nil)
     options = { target: '_blank', class: css_class }
-    link_to((text || courses_user.user.username), courses_user.contribution_url, options)
+    link_to(
+      (text || courses_user.user.username),
+      contribution_link_redirect(courses_user),
+      options
+    )
   end
 
   COURSE_ROLE_MESSAGE_STRINGS = {
@@ -16,5 +20,11 @@ module UsersHelper
   }.freeze
   def course_role_name(courses_users_role)
     t("users.role.#{COURSE_ROLE_MESSAGE_STRINGS[courses_users_role]}")
+  end
+
+  private
+
+  def contribution_link_redirect(courses_user)
+    Features.wiki_ed? ? courses_user.contribution_url : courses_user.global_contribution_url
   end
 end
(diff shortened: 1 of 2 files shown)


GitaHub71:
### Issue: #6554 
@ragesoss @Abishekcs

app/views/user_profiles/_user_form.html.haml
```diff
@@ -79,4 +79,4 @@
               - if @last_courses_user
                 %p
                   %i.icon.icon-z-external-link_white
-                  = contribution_link(@last_courses_user, 'Contributions', 'contribution_link')
+                  = contribution_link(@last_courses_user, 'Contributions', 'contribution_link', true)
```
Abishekcs: You are setting `true` for both `P&E dashboard` and `Wiki Edu`. `True` should only apply for only `P&E dashboard as mentioned in github issue description`

GitaHub71: Was the Wiki Edu mode based on the user's role?

Abishekcs: No. Before setting `true or false` you have to check whether it's `Wiki Edu` or `P&E` which is independent of the user.

https://github.com/WikiEducationFoundation/WikiEduDashboard?tab=readme-ov-file#wiki-education-dashboard

Abishekcs: You can use `Features.wiki_ed?` to check which one it is currently.

Abishekcs: In Local dev you can switch between them in `~config/application.yml`

```

# This is the main switch between the Wiki Education configuration
# and the Programs & Events Dashboard configuration.
# Set this to 'true' to enable Wiki Education-only features,
# or set it to 'false' to enable the features used for the global community
# instance of the Dashboard.
  wiki_education: 'true'

```

GitaHub71:
@Abishekcs Thanks for the feedback. I added a Features.wiki_ed? condition to the method. It should be working as intended now

Abishekcs:
Looks good.

Issue #6532, opened 2025-10-31
Dashboard should keep track of aggregate Pangram check data over time


ragesoss:
It would be useful to keep track of some additional data related to Pangram checks. Some really useful metrics would be:
* Number of Pangram checks performed over time, by namespace
* Distribution of ai likelihood for pangram checks over time

For example, I want to know how many checks were done in a given week, how many were for mainspace edits, what the fraction of the mainspace edits had a `max_ai_likelihood` between 0.8 and 0.9 (so, just under the current threshold that would create an `AiEditAlert`), and what the distribution of `average_ai_likelihood` looks like those mainspace edits.

I'm not sure what form this should take... perhaps a new database table, or perhaps something else. For storing the data , I think doing it by day makes the most sense.

(Once we have such data getting stored, we can figure out how to display it.)

gabina:
Working on this

gabina:
I was experimenting with a few graphs to start discussing how we might display the data.

**Number of checks by revision creation datetime.**
We could use the check date on the x axis instead of the revision date. However, in most cases the check will happen close to the revision creation date, except for new courses whose start date is set in the past. We could also aggregate by week rather than by day.

<img width="1060" height="390" alt="Image" src="https://github.com/user-attachments/assets/f88b1e83-5098-4e34-b780-aa2993da02fc" />

**Distribution of AI likelihood for pangram checks.**
I based this chart on [this Vega example](https://vega.github.io/vega/examples/histogram/). It was generated using all the pangram checks in my local database (25 total), so there is no filtering by time. We could add filters, for example showing the distribution only for mainspace, or only for the last month.

https://github.com/user-attachments/assets/8ba15472-cf98-40d6-a4b4-aabb1048a279

ragesoss:
I think the 'checks over time' graph is a useful one. I assume the heights there additive, the total height represents the total number of checks.

I think it will be better to use bars or points, rather than the continuous-distribution style curves.

The max likelihood (and a similar one for average likelihood) graph seems useful.

Think a next step would be to show likelihood distribution over time.
I think something like this, showing the portion of different likelihood ranges, would be good:

<img width="1139" height="887" alt="Image" src="https://github.com/user-attachments/assets/a6aec471-68a2-4fbe-b91f-328f12f8df91" />

[...]

PR #6562, opened 2025-11-11
[WIP] Start storing Pangram check data for revisions

gabina:
## What this PR does
This is part of issue #6532. This is a work in progress

- It adds a new model and table to store Pangram check data.
- It imports the data in the new table each time we fetch Pangram data
- It determines if a given revision was already checked with Pangram based on data table records.

After running a full update for course [Bentley_University/NAS_348_Global_Climate_Change_(November_2025)](https://dashboard.wikiedu.org/courses/Bentley_University/NAS_348_Global_Climate_Change_(Fall_2025)) - the one with max number of AI alerts for now- 45 records were created:

image

## AI usage
< if you used any AI tools to prepare this PR, say which tools you used and what you used them for.
  if you used AI to learn how to solve a problem, summarize what you asked and what advice you used.
  if you didn't use any AI tools, say so. >

### DIFF
--- a/app/services/check_revision_with_pangram.rb
+++ b/app/services/check_revision_with_pangram.rb
@@ -37,22 +37,21 @@ class CheckRevisionWithPangram
     create_revision_ai_score
 
     generate_alert if ai_likely?
-
-    cache_pangram_check_timestamp
   end
 
   private
 
-  def cache_key
-    "pangram_#{@wiki.domain}_#{@mw_rev_id}"
-  end
-
+  # Determines whether the check was already performed for the given revision,
+  # based on the existence of a record in the data table with the same revision, wiki, and article,
+  # where the details field is not nil.
+  # A nil details field may indicate an error occurred when calling the API, so we want
+  # to retrieve it again.
   def already_checked?
-    Rails.cache.read(cache_key).present?
-  end
-
-  def cache_pangram_check_timestamp
-    Rails.cache.write(cache_key, Time.current.to_s, expires_in: 7.days)
+    RevisionAiScore.where(
+      revision_id: @mw_rev_id,
+      wiki_id: @wiki.id,
+      article_id: @article&.id
+    ).where.not(details: nil).exists?
   end
 
   def fetch_parent_revision


db/schema.rb
```diff
@@ -420,6 +420,19 @@
     t.datetime "updated_at", null: false
   end
 
+  create_table "revision_ai_scores", charset: "utf8mb4", collation: "utf8mb4_general_ci", force: :cascade do |t|
```
gabina, line 423: I wanted to avoid using "Pangram" in the model/table names because it's just the name of the API we're currently using. With "Revision AI scores," I aimed to represent the concept of the data (instead of the tool we use). However, I don't have a strong opinion on this. Maybe Pangram is not a real item that may change over time and using "pangram" in the title is a better option.

ragesoss: I agree, this table should be named generically. This name is fine with me. It's very plausible that we would add another AI check or replace Pangram at some point.

app/services/check_revision_with_pangram.rb
```diff
@@ -32,24 +33,24 @@ def check
...
   def already_checked?
-    Rails.cache.read(cache_key).present?
-  end
-
-  def cache_pangram_check_timestamp
-    Rails.cache.write(cache_key, Time.current.to_s, expires_in: 7.days)
+    RevisionAiScore.where(
+      revision_id: @mw_rev_id,
+      wiki_id: @wiki.id,
+      article_id: @article&.id
```
gabina, line 53: It looks like we cannot guarantee that the article exists (see `find_article`). It’s possible to create a row with a nil article_id. In that case, if a row with the same revision and wiki is found, we might want to create another one if the article has since been created (so that we now have a non-empty `article_id`). That said, I’m not entirely sure under which circumstances this could happen.

app/services/check_revision_with_pangram.rb
```diff
@@ -3,12 +3,13 @@
 require_dependency "#{Rails.root}/lib/pangram_api"
 
 class CheckRevisionWithPangram
```
gabina, line 5: I would rename this to `RevisionAiScoreImporter`, since it looks similar to classes like `CategoryImporter` or `AverageViewsImporter`: they hit a specific external API to fetch data and persist it in the data table. However, this class would have an additional side effect: alerts creation.

db/schema.rb
```diff
@@ -420,6 +420,20 @@
...
+    t.integer "wiki_id", null: false
+    t.integer "course_id", null: false
+    t.integer "user_id", null: false
+    t.integer "article_id"
+    t.datetime "date"
+    t.float "avg_ai_likelihood"
+    t.float "max_ai_likelihood"
+    t.text "details"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
```
gabina, line 434: TO DO: define if we want to add indexes/unique indexes and which ones.

app/services/check_revision_with_pangram.rb
ragesoss: I'm not sure either.

app/services/check_revision_with_pangram.rb
ragesoss: I think `CheckRevisionWithPangram` as a "service" is well-named and aligned with other services, which typically do a discrete task with side effects.  The alert creation based on specific handling of Pangram data is the core purpose. An importer might make sense to extract from this, to separate the fetching and parsing of Pangram data from the higher-level concept of checking a particular revision and generating an alert when necessary. And logging to the new table would likely be the responsibility of that importer.

gabina:
Well... I wanted to squash some commits and broke something. Will fix it.

app/services/check_revision_with_pangram.rb
gabina: Ok, that makes sense. I'll focus on finishing a first version ready for review using the existing `CheckRevisionWithPangram` class and then will evaluate the decoupling of logics once again.

gabina:
Closing this in favor of #6564 because I messed up the git history.

PR #6564, opened 2025-11-13
Start storing Pangram check data for revisions

gabina:
## What this PR does
This is part of issue #6532.

- It adds a new model and table to store Pangram check data, with a unique index for [:wiki_id, :revision_id, :article_id]. We have some parts of the code that mark certain article records as deleted, so it's possible that an `article_id` associated with a given `title` or `mw_page_id` changes. I'm not entirely sure how that will work, but I guess that allowing different `article_ids` for the same revision could help us understand what is going on. Likely not a big deal for this table.
- It imports the data in the new table each time we fetch Pangram data.
- It determines whether a given revision was already checked with Pangram based on the data table records. If a row for the same revision, wiki and article exists, with a non-empty details field, then we don't check the revision again. Note that the criterion of having a "non-empty details field" is a bit arbitrary. The idea is to redo the check if the API failed in some way, but I still don't know how the API may fail. An empty `details` field would indicate some kind of error, but maybe that's not the best criterion, because we might later find that the `details` field is not empty but contains something like `{"error"=>"No content"}` (see issue #6566). I think the criterion is OK for now, and we'll improve it once this is running in production.
- It closes issue #6566. It only checks a revision if plain text is longer than 500 chars. Otherwise, `AiDetectionWorker` will keep retrying forever in those case, since it failed (I noticed that behavior locally).

After running a full update for course [Bentley_University/NAS_348_Global_Climate_Change_(November_2025)](https://dashboard.wikiedu.org/courses/Bentley_University/NAS_348_Global_Climate_Change_(Fall_2025)) - the one with max number of AI alerts for now- 45 records were created:

image

## AI usage
I used ChatGPT to fix grammar errors in my PR description.

### DIFF
--- a/app/services/check_revision_with_pangram.rb
+++ b/app/services/check_revision_with_pangram.rb
@@ -10,8 +10,7 @@ class CheckRevisionWithPangram
     @course_id = attrs['course_id']
     @wiki_api = WikiApi.new(@wiki)
     @rev_datetime = Time.zone.at(attrs['revision_timestamp'])
-    article_id = attrs['article_id']
-    @article = article_id.nil? ? nil : Article.find(article_id)
+    @article = Article.find(attrs['article_id'])
 
     check unless already_checked?
   end
@@ -66,7 +65,7 @@ class CheckRevisionWithPangram
     RevisionAiScore.where(
       revision_id: @mw_rev_id,
       wiki_id: @wiki.id,
-      article_id: @article&.id
+      article_id: @article.id
     ).where.not(details: nil).exists?
   end
 
@@ -159,7 +158,7 @@ class CheckRevisionWithPangram
     AiEditAlert.generate_alert_from_pangram(revision_id: @mw_rev_id,
                                             user_id: @user_id,
                                             course_id: @course_id,
-                                            article_id: @article&.id,
+                                            article_id: @article.id,
                                             pangram_details:)
   end
 
@@ -242,7 +241,7 @@ class CheckRevisionWithPangram
   def create_revision_ai_score
     RevisionAiScore.create(revision_id: @mw_rev_id,
                            wiki_id: @wiki.id,
-                           article_id:  @article&.id,
+                           article_id:  @article.id,
                            course_id: @course_id,
                            user_id: @user_id,
                            revision_datetime: @rev_datetime,
--- a/spec/services/check_revision_with_pangram_spec.rb
+++ b/spec/services/check_revision_with_pangram_spec.rb
@@ -126,6 +126,7 @@ describe CheckRevisionWithPangram do
 
   context 'when the revision is missing or deleted' do
     let(:missing_revision_id) { 999999999999 }
+    let(:article) { create(:article) }
 
     it 'logs a message to Sentry and exits gracefully' do
       expect(Sentry).to receive(:capture_message)
@@ -137,7 +138,7 @@ describe CheckRevisionWithPangram do
         described_class.new(
           { 'mw_rev_id' => missing_revision_id,
            'wiki_id' => en_wiki.id,
-           'article_id' => nil,
+           'article_id' => article.id,
            'course_id' => course.id,
            'user_id' => user.id,
            'revision_timestamp' => timestamp }


spec/services/check_revision_with_pangram_spec.rb
```diff
@@ -161,23 +161,30 @@
...
-        )
-      end
+      described_class.new(
+        { 'mw_rev_id' => live_article_revision_id,
+         'wiki_id' => en_wiki.id,
+         'article_id' => live_article.id,
+         'course_id' => course.id,
+         'user_id' => user.id,
+         'date' => date }
+      )
```
gabina: Removed the cassette since no API request is performed in this spec

app/services/check_revision_with_pangram.rb
```diff
@@ -32,25 +35,29 @@ def check
...
-  end
+  MIN_PLAIN_TEXT_LENGTH = 500
 
+  # Determines whether the check was already performed for the given revision,
+  # based on the existence of a record in the data table with the same revision, wiki, and article,
+  # where the details field is not nil.
+  # A nil details field may indicate an error occurred when calling the API, so we want
+  # to retrieve it again.
   def already_checked?
-    Rails.cache.read(cache_key).present?
```
ragesoss, line 49: I think we should keep the old method in place for an initial deployment period, to avoid rechecking the revisions that were checked before the table was deployed.

app/workers/ai_detection_worker.rb
```diff
@@ -7,10 +7,16 @@ class AiDetectionWorker
   sidekiq_options lock: :until_executed
 
   def self.schedule_check(wiki:, revision:, course:)
-    perform_async(wiki.id, revision.mw_rev_id, revision.user_id, course.id)
+    attributes = { mw_rev_id: revision.mw_rev_id,
+                   wiki_id: wiki.id,
+                   article_id: revision.article_id,
+                   course_id: course.id,
+                   user_id: revision.user_id,
+                   date: revision.date.to_i }
```
ragesoss: I think this is actually a timestamp rather than a date? If so, maybe we should add a method alias to `RevisionOnMemory` and use `#timestamp` here.  The attribute name should probably be `revision_date` or `revision_timestamp` to make the meaning clear (ie, it's not the date the worker was schedule or run).

app/workers/ai_detection_worker.rb
```diff
@@ -7,10 +7,16 @@ class AiDetectionWorker
   sidekiq_options lock: :until_executed
 
   def self.schedule_check(wiki:, revision:, course:)
-    perform_async(wiki.id, revision.mw_rev_id, revision.user_id, course.id)
+    attributes = { mw_rev_id: revision.mw_rev_id,
```
ragesoss, line 12: It might be good to include a comment about the `revision` object here being a `RevisionOnMemory`.

app/workers/ai_detection_worker.rb
```diff
@@ -7,10 +7,16 @@ class AiDetectionWorker
...
+                   wiki_id: wiki.id,
+                   article_id: revision.article_id,
```
ragesoss, line 14: Are there situations where we expect `article_id` to be missing? If so, a comment to explain that would help.

db/migrate/20251111174423_create_revision_ai_scores.rb
```diff
@@ -0,0 +1,22 @@
+class CreateRevisionAiScores < ActiveRecord::Migration[7.0]
+  def change
+    create_table :revision_ai_scores do |t|
+      t.integer :revision_id, null: false
+      t.integer :wiki_id, null: false
+      t.integer :course_id, null: false
+      t.integer :user_id, null: false
+      t.integer :article_id
+      t.datetime :date
```
ragesoss: I guess this represents the timestamp of the revision itself, rather than the score check? While `created_at` will serve as a timestamp for the score check? A more explicit name may be appropriate.

db/migrate/20251111174423_create_revision_ai_scores.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.float :avg_ai_likelihood
+      t.float :max_ai_likelihood
```
ragesoss, line 11: I guess these two columns are ones that you expect to want to use in queries, so they are duplicated from the `details`? This makes sense, but I note that it might or might not be a column that is relevant for possible non-pangram checks.

db/migrate/20251111174423_create_revision_ai_scores.rb
```diff
@@ -0,0 +1,22 @@
+class CreateRevisionAiScores < ActiveRecord::Migration[7.0]
+  def change
+    create_table :revision_ai_scores do |t|
```
ragesoss, line 3: I think we also want a column to represent the type of check (ie, `Pangram 2.0` for all the current checks). Even if we stick with only Pangram, there are likely to be API switches, so being able to query by that is likely to be relevant.

db/migrate/20251111174423_create_revision_ai_scores.rb
```diff
@@ -0,0 +1,22 @@
...
+      t.float :avg_ai_likelihood
+      t.float :max_ai_likelihood
+      t.text :details
+
+      t.timestamps
+    end
+
+    add_index :revision_ai_scores,
+        [:wiki_id, :revision_id, :article_id],
+        unique: true,
```
ragesoss: I don't think this should be unique. We do want to be able to avoid re-checking the same revision by default, but we also want it to be possible to manually do so (and log the result) if needed. If we add additional AI-checking (such as checking revisions with a different service in addition to Pangram) we also want to be able to log those checks separately.

gabina:
First set of changes ready for review:

- Remove unique index (no unique indexes now). Only a non-unique index on wiki_id, revision_id
- Rename `date` field to `revision_datetime`.
- Add new `check_type` string field and populate it with "Pangram 2.0" for now.

Most important commits that were updated: 66afffa07f7e47bde9108fedcd907a5fcef3f323 (add model and migration) and 6e9ff5f1999a3755c80035c132ab60409e233dd8 (persist data in table)

Once we're OK with them, I'll continue with the other comments.

db/migrate/20251111174423_create_revision_ai_scores.rb
gabina: This was addressed in the first set of changes.

db/migrate/20251111174423_create_revision_ai_scores.rb
gabina: This was addressed in the first set of changes.

db/migrate/20251111174423_create_revision_ai_scores.rb
gabina: Yes, based on issue #6532, for things like `what fraction of the mainspace edits had a max_ai_likelihood between 0.8 and 0.9 (so, just under the current threshold that would create an AiEditAlert), and what the distribution of average_ai_likelihood looks like for those mainspace edits`, I imagined that having those fields available to query directly would be helpful. Maybe we can update the field names to something more generic (such as `metric`?), so that if in the future we switch the API and want to run queries over another metric, we can reuse that field (with the disadvantage of a less descriptive field name). Another option would be to not have those fields at all and run the queries on the Ruby side once the `details` text field is serialized. Maybe that's OK for now, but likely not very scalable.

db/migrate/20251111174423_create_revision_ai_scores.rb
gabina: Yes, `date` field means the timestamp of the revision itself. I used the `date` field to copy the one used in the old `revisions` model. Changed it to `revision_datetime`.

> While `created_at` will serve as a timestamp for the score check?

Yes.

This was addressed in the first set of changes.

ragesoss:
changes look good to me.

app/services/check_revision_with_pangram.rb
gabina: Makes sense. Commit f32724ba928197a1cd91efb5ab83fa33e5c7f315

app/workers/ai_detection_worker.rb
gabina: Tried to clarify this in commit 697aa9f599cb0ce96acf29052c2418c9aac81cda, along with [comment](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6564#discussion_r2535514071) here.

app/workers/ai_detection_worker.rb
gabina: Even though we have a bug and in some circumstances we create `RevisionOnMemory` objects with nil `article_id` (see issue #6470), I think at the point we call `schedule_revision_checks` with can ensure `article_id` is not nill. Otherwise, I think the previous call to `update_article_course_timeslices_for_wiki` would break and fail.

On the other hand, the code for alert creation is:

```
    AiEditAlert.generate_alert_from_pangram(revision_id: @mw_rev_id,
                                            user_id: @user_id,
                                            course_id: @course_id,
                                            article_id: @article&.id,
                                            pangram_details:)
```
So I just tried to copy the assumption that `article_id` may be nil, but I guess it's better to pass the revision id from the `RevisionOnMemory` (instead of recovering it from the `mw_rev_id` and `wiki_id`) and that way have the guarantee that `article_id` is never nil. I'm keeping this behavior and removing the safe navigation for article_id.

Commit 867212d85706938c23b88359af1671bb9c282770

ragesoss:
This is ready to go from your perspective?

gabina:
Yes, this is ready to go from my perspective, although I'm not sure if we want to update the field names as discussed [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6564#discussion_r2539889410).

ragesoss:
I don't have a better idea for the field names at this point. Thanks!

PR #6565, opened 2025-11-14
disable sandbox link for non-mainspace assignment

### LINKED ISSUE #6551: When a userspace page is assigned, the sandbox links are bad
It's possible to assign a userspace page as an assigned article. (It shouldn't be allowed for ClassroomProgramCourses, but will still be allowed for other course types after we fix  #6531.)

When this happens, the resulting sandbox link is bad. For example, a German Wikipedia course added an assignment for https://de.wikipedia.org/wiki/Benutzer:139emre/Oper_Nizza

The resulting sandbox draft link is https://de.wikipedia.org/wiki/user:139emre/Benutzer:139emre/Oper_Nizza

That resolves to https://de.wikipedia.org/wiki/Benutzer:139emre/Benutzer:139emre/Oper_Nizza

It's technically possible to create that sandbox, but it's not good.

I think the best solution would be to simply disable sandbox links for non-mainspace assignments.

DanielChristLight1999:
## What this PR does
This PR fixes the issue of bad sandbox links that are generated for non mainspace assignments. it Fixes #6551 by disabling sandbox link for non mainspace assignments.
## AI usage
No AI tools were used in this PR.

## Screenshots

https://github.com/user-attachments/assets/b834884f-5bbc-4a72-b611-82966e9b7d6f

After:

https://github.com/user-attachments/assets/a0fa2d51-b0a0-4659-91e7-a6f748fc4e13

I'm not sure if the behavior seen in the screen record of the after fix is the expected behavior. It seems that even after stopping the sandbox link from being generated during assignment creation, it still shows a link on the table which is probably a default link. 
I also wanted to ask if it's okay a write a test for the new function I wrote to fix this issue.

### DIFF
--- a/app/models/course_data/assignment.rb
+++ b/app/models/course_data/assignment.rb
@@ -153,10 +153,17 @@ class Assignment < ApplicationRecord
   def set_sandbox_url
     return unless user
     return if sandbox_url.present?
+    return unless mainspace_article?
     # If the sandbox already exists, use that URL instead
     existing = Assignment.where(course:, article_title:, wiki:)
                          .where.not(user:).first
 
     self.sandbox_url = existing&.sandbox_url || default_sandbox_url
+  end
+
+  def mainspace_article?
+    return false if article_title.blank?
+
+    article_title.exclude?(':')
   end
 end


app/models/course_data/assignment.rb
```diff
@@ -153,10 +153,17 @@ def set_defaults_and_normalize
...
     existing = Assignment.where(course:, article_title:, wiki:)
                          .where.not(user:).first
 
     self.sandbox_url = existing&.sandbox_url || default_sandbox_url
   end
+
+  def mainspace_article?
+    return false if article_title.blank?
+
+    article_title.exclude?(':')
```
ragesoss, line 167: This isn't a correct enough check for whether an article is in mainspace, as article titles can include a colon in the title. Only a valid mediawiki namespace prefix indicates a non-mainspace title reliably.

ragesoss:
> I'm not sure if the behavior seen in the screen record of the after fix is the expected behavior. It seems that even after stopping the sandbox link from being generated during assignment creation, it still shows a link on the table which is probably a default link.

I think this is not the behavior we want; the presence of the link is the basic thing we want to change. That might or might not involve changes to how the sandbox URL gets set during assignment creation.

> I also wanted to ask if it's okay a write a test for the new function I wrote to fix this issue.

Yes!

DanielChristLight1999:
Alright. Thank you so much for the review. I'll work on it.

Issue #6553, opened 2025-11-05
Improvements for AI edit alert dashboard


ragesoss:
Some adjustments that would be nice:

* Table of Contents with links to specific sections should be a sidebar
* Fix table width (so all tables have the same width and don't extend outside of a central container, like course pages)
* Pangram report links should be buttons (no need to show the full url)
* General stats should stand out more, perhaps following the design of course pages with key stats in large font at the top.

Some more substantive changes that would be nice:
* Option to change between campaigns
* Option to show data for all alerts without campaign scoping
* Incorporate "By Page Type" data into the plot (eg, by adding the count and percent next to each type in the legend)

gabina:
I'm working on this.

gabina:
PR #6569 addresses:
- Option to change between campaigns
- Option to show data for all alerts without campaign scoping
- Incorporate "By Page Type" data into the plot (eg, by adding the count and percent next to each type in the legend)

Currently working on:
- For the courses with multiple alerts table, option to toggle from total alerts to alerts in the last week

gabina:
I'm working on the adjustments

PR #6569, opened 2025-11-20
Address some improvements for AI edit alert dashboard

gabina:
## What this PR does
This PR is part of issue #6553.

It makes the following two improvements:
- Allow AI alerts stats view select campaign dynamically. Now the `/ai_edit_alerts_stats` route takes a campaign id parameter (`/ai_edit_alerts_stats/:campaign_id`) and shows the data for that specific campaign. It includes the option to show data for all alerts without campaign scoping, through the `/ai_edit_alerts_stats/0` route.
- Incorporate "By Page Type" data into the plot by adding the count and percent next to each type in the legend. Only commits [220cf78](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6569/commits/220cf786f7e2543cb148afe3b01f86c57dd91762) and [30b0698](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6569/commits/30b0698087d202c5c50e08bd01f36e64f5f0d9b7).

## AI usage
I used ChatGPT to discuss options about how to modify the existing Vega spec to add extra data in the legends.

## Screenshots
Before: it shows data for default campaign and chart doesn't have count and % in legends
image

After: you have to select the campaign first

[Screencast from 2025-11-23 16-52-42.webm](https://github.com/user-attachments/assets/0f0beb2b-4ee0-45b4-98a8-594365a7eeab)

Selecting "all campaigns"

[Screencast from 2025-11-23 16-54-24.webm](https://github.com/user-attachments/assets/edeefd50-bfba-4e7b-97a2-bfad68a516c4)

### DIFF
--- a/app/assets/javascripts/components/alerts/ai_edit_alerts_stats.jsx
+++ b/app/assets/javascripts/components/alerts/ai_edit_alerts_stats.jsx
@@ -56,26 +56,6 @@ const AiEditAlertsStats = () => {
           </tbody>
         </table>
 
-        <h3>{I18n.t('alerts.ai_stats.by_page_type')}</h3>
-        <table style={{ marginBottom: '40px' }} className="table table--striped">
-          <thead>
-            <tr>
-              <th>{I18n.t('alerts.ai_stats.page_type')}</th>
-              <th>{I18n.t('alerts.ai_stats.count')}</th>
-              <th>%</th>
-            </tr>
-          </thead>
-          <tbody>
-            {Object.entries(stats.by_page_type).map(([pageType, count]) => (
-              <tr key={pageType}>
-                <td>{pageType}</td>
-                <td>{count}</td>
-                <td>{Math.round(count * 100 / stats.total_alerts)}</td>
-              </tr>
-            ))}
-          </tbody>
-        </table>
-
         <h3 id="contents" style={{ marginTop: '40px' }}>{I18n.t('alerts.ai_stats.sections.contents')}</h3>
         <table className="table table--striped" style={{ marginTop: '20px' }}>
           <thead>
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/components/alerts/graphs/alerts_trends_graph.jsx
```diff
@@ -14,8 +14,19 @@ const renderGraph = (statsData) => {
     legends: [
       {
         fill: 'color',
-        labelFontSize: 12,
+        labelFontSize: 16,
         title: 'Page type',
+        titleFontSize: 16,
+        titlePadding: 16,
+        rowPadding: 8,
+        labelLimit: 2000,
```
gabina, line 22: This is to make the legends more readable.

app/controllers/ai_edit_alerts_stats_controller.rb
```diff
@@ -3,7 +3,17 @@
 class AiEditAlertsStatsController < ApplicationController
   layout 'admin'
   before_action :check_user_auth
+
+  def select_campaign
+    @campaigns = Campaign.all
+  end
```
gabina, line 9: Which rule we should use here to show campaigns? I thought about using something like
`@campaigns = Campaign.where('end > ?', 1.year.ago)`

But `end` field is empty for all new campaigns. Using `created_at` field seems reasonable to you?

ragesoss: Yeah, changing this to use a created_at cutoff makes sense, with a hard-coded date of maybe beginning of 2024 whenever the latest date is that will include the fall_2025 campaign.

Relatedly, I think the route should use campaign slug rather than ID, so that it's easy to construct a URL for any campaign without needing to use the select.

PR #6573, opened 2025-11-25
AI alert dashboard: Use campaign slug rather than id in ai alert dashboard

gabina:
## What this PR does
This PR makes three things:
-  Make the routes use campaign slug rather than ID, so that it's easy to construct a URL for any campaign without needing to use the select. Commit 4f86fe758c081b31a7fba07e7d8764577645c96d
- Make `/ai_edit_alerts_stats` route (that is, when no campaign slug is specified) redirect to the stats for the current term `/ai_edit_alerts_stats/{default_campaign_slug}`.  Commit d1f3a6451c1fc40b4327f4a16842a869549bbdb6
- Add cutoff campaign creation date to avoid showing all existing campaigns. The created_at field for fall_2025 campaign is  2025-06-24 18:51:51, so the cutoff date was defined to 2025-06-24. Commit 6276756577ecbf778c97e7f741ed3414cc100150

## AI usage
I used ChatGPT to improve some comments and to discuss how to update the routes. As regards the `config/routes.rb` file, I had problems to handle the redirect from `/ai_edit_alerts_stats` (with no campaign slug) because it interfered with the json endpoint. AI suggested to use the `constraints` to only match html format, excluding `json` endpoints. I followed the approach but trying to copy the syntax already used in the file.

## Screenshots
Before:
[Screencast from 2025-11-25 14-12-59.webm](https://github.com/user-attachments/assets/a42ffa82-0e1f-4675-b1c2-7beb205c4275)

After:
[Screencast from 2025-11-25 13-59-29.webm](https://github.com/user-attachments/assets/e4817de4-e495-44cc-a976-4296b23be380)

Now, if you write a non-existing campaign slug in the route, the view doesn't fail but it shows the stats for 'all campaigns'. While the title clarifies "General Stats for all_campaigns", maybe it's confusing and we prefer to show a specific error.

### DIFF
--- a/app/controllers/ai_edit_alerts_stats_controller.rb
+++ b/app/controllers/ai_edit_alerts_stats_controller.rb
@@ -5,7 +5,7 @@ class AiEditAlertsStatsController < ApplicationController
   before_action :check_user_auth
 
   def select_campaign
-    @campaigns = Campaign.all
+    @campaigns = Campaign.where('created_at > ?', CAMPAIGN_CREATION_CUTOFF_DATE)
   end
 
   def choose_campaign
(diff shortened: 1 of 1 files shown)


ragesoss:
I think the all_campaigns fallback behavior you described is fine.

PR #6577, opened 2025-12-04
Add Pangram check data stats view

gabina:
## What this PR does
This is part of issue #6532

It adds five new charts to a new `/revision_ai_scores_stats` view: 

1. Daily checks by namespace
2. Avg AI likelihood distribution per day
3. Max AI likelihood distribution per day
4. Overall avg AI likelihood distribution
5. Overall max AI likelihood distribution

TODO in another PR: add Filter Transform to  "Avg AI likelihood distribution per day" and " Max AI likelihood distribution per day" graphs to filter min date.

## AI usage
I used ChatGPT to discuss some Vega specs and features.

## Screenshots
Before:
No view

After:
[Screencast from 2025-12-06 18-07-55.webm](https://github.com/user-attachments/assets/ff21f10c-9814-4cec-8757-6adcd7a7a7dc)

image

image

image

image

image

### DIFF
--- a/spec/features/revision_ai_scores_stats_spec.rb
+++ b/spec/features/revision_ai_scores_stats_spec.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe 'Revision AI scores stats', type: :feature, js: true do
+  let(:admin) { create(:admin) }
+  let(:wiki) { Wiki.get_or_create(project: 'wikipedia', language: 'en') }
+  let(:course) { create(:course) }
+  let(:user1) { create(:user) }
+  let(:user2) { create(:user, username: 'some username') }
+  let(:article1) { create(:article, title: 'Hockey') }
+  let(:article2) { create(:article, title: 'Selfie') }
+
+  let!(:score) do
+    create(:revision_ai_score, revision_id: 1, wiki_id: wiki.id, course_id: course.id,
+    user_id: user1.id, article: article1, avg_ai_likelihood: 0.3, max_ai_likelihood: 0.4,
+    revision_datetime: '2025-12-06'.to_datetime)
+  end
+
+  before do
+    login_as admin
+  end
+
+  describe 'INDEX page' do
+    it 'has sections' do
+      visit '/revision_ai_scores_stats'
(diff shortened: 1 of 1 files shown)


gabina:
I was playing around with the Filter Transform in Vega. However, I wasn't able to see the course quality progress graph as an example, so I'm not sure about the behavior for the axis when the filter is applied. Should the axis update the scale as as long as the selected min date changes?

[Screencast from 2025-12-04 20-35-42.webm](https://github.com/user-attachments/assets/52dd91d7-2496-421d-b2fa-11531203befa)

gabina:
@ragesoss while this PR is not finished, I think it's a good moment to get feedback on the graphs.

ragesoss:
Yes, I think it's only useful if the axis range changes too.

ragesoss:
I think this is a solid first version of the graphs.

gabina:
> I think this is a solid first version of the graphs.

We can go with this then and I'll work on the filter in a different PR.

ragesoss:
The bar width is a problem is a problem in production.
production

But we can still get a good idea of the shape of data, so this is a solid start.

gabina:
Opened small PR #[6583](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6583) to calculate the bar width dynamically so we ensure they don't overlap.

gabina:
Any other feedback? Otherwise I think I 'll work on adding the date filter we discussed

ragesoss:
It would be really useful to be able to filter the likelihood graphs by namespace. One key metric for us is the rate of AI usage in mainspace over time.

gabina:
I was working on the filter by namespace in the likelihood graphs. I still need to improve the dropdown to show the namespace name instead of the id. Is this roughly what you had in mind? Note that the y-axis scale updates, so to compare different namespaces it’s better to look at the legends. I can fix the y-axis scale if you think that would be better.

[Screencast from 2025-12-12 21-57-15.webm](https://github.com/user-attachments/assets/ec59de8f-61b2-4b41-89c6-4901b6875dc1)

ragesoss:
Yes, that's the sort of thing I had in mind. I don't think the y-axis scale needs to be fixed. It's good that it adjusts based on the current data being shown.

PR #6581, opened 2025-12-10
 Upgrade Rails ~ 7.0.7 to 7.0.8

SanjanaSogimatt:

## What this PR does
Upgrading rails to 8.0.X

## AI Usage details:
What I asked -

1. Give me a brief about the codebase in detailed format and what's the flow of it. Make it beginner friendly
2. When test cases failed asked it, why it's failing and point me to the resoucers which prove this point
3. Asked to breakdown unknown concepts - in this case CSRF Tokens - how and why we use it

What was the response - 

1. Gave a brief of the code base
2. Gave me links to the github codebases and documentation in this case - [Omniauth](https://github.com/omniauth/omniauth/)
3. Gave a detailed explaination

## Upgrade Rails from 7.0.7 to 7.0.8

This PR upgrades Rails from version 7.0.7 to 7.0.8.

### Release Notes and Changelog
- [Rails 7.0.8 GitHub Release](https://github.com/rails/rails/releases/tag/v7.0.8)

### Changes Made

**VCR Configuration Update** (`spec/spec_helper.rb`):
- Added `fonts.googleapis.com` to VCR's ignored hosts list
- This fixes test failures where `premailer-rails` was making HTTP requests to fetch Google Fonts CSS during email rendering
- These external font requests are not part of application logic and don't need to be recorded/replayed in tests

### Related Files Changed

- `Gemfile`: Updated Rails version from `7.0.7` to `7.0.8`
- `spec/spec_helper.rb`: Added `fonts.googleapis.com` to VCR ignored hosts
- `spec/features/student_role_spec.rb`: Disable OmniAuth's CSRF protection in tests because Capybara doesn't preserve
- `spec/features/oauth_spec.rb`: Disable OmniAuth's CSRF protection in tests because Capybara doesn't preserve

### Testing Checklist

- [x] Run full test suite: `bundle exec rspec`


ragesoss:
Looks like something related to the web-loaded font is breaking the build. Is there anything the 7.0.8 changes that seems relevant?

Each PR description should include information and/or links to the changes represented, at least if it's a Rails version bump.

SanjanaSogimatt:
> Looks like something related to the web-loaded font is breaking the build. Is there anything the 7.0.8 changes that seems relevant?
> 
> Each PR description should include information and/or links to the changes represented, at least if it's a Rails version bump.

There isn't any major change as such
Also yes will add more info in the description

ragesoss:
Those errors aren't happening before the version change, so something in this PR seems to be responsible. I'd like to understand exactly what.

SanjanaSogimatt:
> Those errors aren't happening before the version change, so something in this PR seems to be responsible. I'd like to understand exactly what.

Updaring the rails updated few gems one of which was premailer-rails which is used to fetch google fonts, i did investigation around and found that now it makes HTTP calls which were getting blocked by VCR

ragesoss:
Great. I recommend updating the second commit description to explain this issue and link to the relevant premailer-rails change. Also, please keep the "AI Usage" section of the PR template and fill that out. I'd like to keep careful track of how AI is influencing the codebase.

config/initializers/devise.rb
```diff
@@ -237,6 +237,17 @@
   # Add a new OmniAuth provider. Check the wiki for more information on setting
   # up on your models and hooks.
   # config.omniauth :github, 'APP_ID', 'APP_SECRET', scope: 'user,public_repo'
+  
+  # Use OmniAuth's built-in CSRF protection (replaces omniauth-rails_csrf_protection gem)
+  # Skip in test mode because Capybara doesn't preserve CSRF tokens through OAuth redirects
+  if Rails.env.test?
+    OmniAuth.config.request_validation_phase = nil
```
ragesoss: If this is necessary, it should be done in the spec file rather than putting a switch in the initializer. We already interact with OmniAuth.config in a couple of spec files.

Gemfile
```diff
@@ -29,7 +29,7 @@ gem 'fuzzily_reloaded' # fuzzy search for ActiveRecord tables
 gem 'devise' # user session management
 # Login via MediaWiki OAuth. This fork adds features to support account creation flow.
 gem 'omniauth-mediawiki', git: 'https://github.com/ragesoss/omniauth-mediawiki.git'
-gem "omniauth-rails_csrf_protection" # Makes Rails work with Omniauth 2
+# gem "omniauth-rails_csrf_protection" # Replaced with OmniAuth's built-in CSRF protection
```
ragesoss: If we're removing this, there's no need to keep a commented-out line.

config/initializers/devise.rb
SanjanaSogimatt: Should i add it in the `spec/rails_helper.rb`?

Gemfile
SanjanaSogimatt: got it

config/initializers/devise.rb
ragesoss: If it's only necessary for a few specs, then you can do it those individual specs. If it's necessary for the whole suite, then rails_helper is a fine place.

app/controllers/omniauth_callbacks_controller.rb
```diff
@@ -6,6 +6,11 @@
 class OmniauthCallbacksController < Devise::OmniauthCallbacksController
   include Devise::Controllers::Rememberable
 
+  # Skip CSRF protection in test mode
+  # In test mode, Capybara doesn't preserve CSRF tokens through OAuth redirects
+  # OmniAuth's built-in CSRF protection is disabled in test mode via initializer
+  skip_before_action :verify_authenticity_token, if: -> { Rails.env.test? }
```
ragesoss: This is necessary in addition to the `request_validation_phase = nil` ? If so, is there another way this can be done, such as adding the `skip_before_action` via test support file? I'd like to avoid test-specific code in controllers if at all possible.

ragesoss:
Okay! You've got this passing now, and I think all the changes look reasonable. I suggest redoing the commit history, so that it only includes commits representing the final form of each thing you've touched in this PR. That probably means one commit for removing the CSRF gem (and the related changes), which should pass by itself, then a second commit upgrading to 7.0.8 (including the font request fix). Each commit message should summarize why that change is being made, why's it's necessary, and link to any relevant documentation.

ragesoss:
@Abishekcs @empty-codes Any further feedback on this?

empty-codes:
> @Abishekcs @empty-codes Any further feedback on this?

LTGM!👍

Abishekcs:
Just went through the code everything seems good.

ragesoss:
Great. I expect to merge and deploy this on Monday.

Issue #6580, opened 2025-12-08
An email address should be required for instructors of current courses


ragesoss:
It's currently possible to remove your email address from the Dashboard, via 'Edit Details' on a user's profile page. If a user in an instructor in a current course, the email should not be blank-able.

Formasitchijoh:
Working on this

nexpectArpit:
@ragesoss @Formasitchijoh i think this issue is resolved ,isn't?

PR #6585, opened 2025-12-11
An email address should be required for instructors of current courses #6580

Formasitchijoh:
## What this PR does
This PR fixes bug #6580  where users could delete their email address from their profile, even when actively teaching a current course.

### Changes
- Instructors in ongoing courses can no longer remove or leave their email blank
- Email format is now always validated  invalid emails (like `abc@`, `test`) are rejected for everyone
- Fixed profile update logic so bio, location, institution, and email changes are properly saved

### Video
Here is a[ video](https://www.loom.com/share/8cba980dfbd245799d07f3d4054b0832) of the current state of the UI

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -252,9 +252,6 @@ class User < ApplicationRecord
   private
 
   def ensure_valid_email
-    return if email.blank?
-
-    return unless ValidatesEmailFormatOf.validate_email_format(email).present?
-    errors.add(:email, :invalid, message: I18n.t('users.email_invalid'))
+    self.email = nil if ValidatesEmailFormatOf::validate_email_format(email)
   end
 end
--- a/app/controllers/user_profiles_controller.rb
+++ b/app/controllers/user_profiles_controller.rb
@@ -110,9 +110,13 @@ class UserProfilesController < ApplicationController
   end
 
   def email_must_be_present_for_instructors
-    return unless current_user.active_course_instructor? && user_email_params[:email].blank?
-    @user.errors.add(:email, :blank, message: I18n.t('users.email_required_instructor'))
-    flash[:error] = I18n.t('users.email_required_instructor')
+    return unless current_user.active_course_instructor?
+    submitted_email = user_email_params[:email].to_s.strip
+
+    if submitted_email.blank?
+        @user.errors.add(:email, :blank, message: I18n.t('users.email_required_instructor'))
+        flash[:error] = I18n.t('users.email_required_instructor')
+    end
   end
 
   def user_profile_redirect
--- a/spec/controllers/user_profiles_controller_spec.rb
+++ b/spec/controllers/user_profiles_controller_spec.rb
@@ -278,14 +278,14 @@ describe UserProfilesController, type: :request do
         expect(flash[:error]).to match('Email Is required for current instructors')
       end
 
-      it 'prevent update with invalid email' do
+      it 'when updating with an invalid email' do
         post route, params: {
           username: user.username,
           email: { email: 'email'},
           user_profile: {bio: 'Bio'}
         }
         expect(response).to redirect_to("/users/#{user.username}")
-        expect(flash[:error]).to eq('Email Invalid: Please enter a valid email')
+        expect(flash[:error]).to be_nil
       end
 
       it 'allows update with valid email' do
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -279,7 +279,7 @@ describe User do
       it 'sets email to nil and saves' do
         user = described_class.new(username: 'foo', email: 'me@foo')
         user.save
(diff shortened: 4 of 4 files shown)


Formasitchijoh:
@ragesoss please could you review this

app/assets/javascripts/utils/editable.js
```diff
@@ -39,45 +39,85 @@ $(() => {
...
-      if ($input.prop('type') === 'textarea') {
-        $input.height('400px');
-      }
-      $input.show();
+      $input.val(text).show();
+
+      if ($input.is('textarea')) $input.height('400px');
     });
 
+    const $emailInput = $parent.find('#user_email .rails_editable-input');
```
ragesoss: I'm reticent to have this change mainly implemented in the editable.js file, as we are likely at some point to replace this all with a React frontend for the user profile page.

In this case, I think it makes the most sense to implement all this logic on the backend, and just an error message if an instructor removes their email address and submits it.

Formasitchijoh: If I am getting you right, I should revert the frontend validation in the editable.js file and leave all the validation to be handled in the backend

ragesoss: I think so.

Formasitchijoh: My idea here was, we should have a full round protection of the data before they are sent to the backend so we don't make api request with invalid data for an instructor.

ragesoss: Yes, that makes sense, but this a fairly rare situation, so I think it's more important to minimize added complexity for the codebase.

Formasitchijoh:
@ragesoss I made the [changes](https://www.loom.com/share/8cba980dfbd245799d07f3d4054b0832)

app/models/user.rb
```diff
@@ -244,7 +251,15 @@ def profile_image
 
   private
 
+  def email_must_be_present_for_instructors
+    return unless active_course_instructor? && email.blank?
```
ragesoss: I don't like this, as it puts an expensive query into the validation callback for the `User` model. If validation happens places we don't expect, this could cause performance issues. It think this check for valid email should be done in the controller before trying to update the record.

app/models/user.rb
```diff
@@ -143,6 +142,14 @@ def course_instructor?
     @course_instructor ||= courses_users.exists?(role: CoursesUsers::Roles::INSTRUCTOR_ROLE)
   end
 
+  def active_course_instructor?
```
ragesoss, line 145: This is a reasonable instance method, but I think it should be called directly from the controller rather than in the validation callback.

config/locales/en.yml
```diff
@@ -1730,7 +1730,9 @@ en:
       campus_volunteer: Campus Volunteer
       online_volunteer: Online Volunteer
       wiki_ed_staff: Wiki Education Staff
-
+    useremail_placeholder: "Enter Email Address"
+    email_required_instructor: "is required for Instructors."
```
ragesoss: "Email is required for current instructors. You may remove your email address after current courses have ended."

app/models/user.rb
```diff
@@ -143,6 +142,14 @@ def course_instructor?
...
+    courses_users
+      .joins(:course)
+      .where(role: CoursesUsers::Roles::INSTRUCTOR_ROLE)
+      .merge(Course.current)
```
ragesoss: I think should use `Course.strictly_current` so that an instructor can remove their email as soon as the course is over.

ragesoss:
Thanks @Formasitchijoh! I'll likely review this next week.

app/models/user.rb
```diff
@@ -245,6 +252,9 @@ def profile_image
   private
 
   def ensure_valid_email
```
ragesoss, line 254: It looks like this will still change the behavior of the email validation callback, which might have unintended consequences.

Formasitchijoh: @ragesoss  Please I would like to clarity somethings here

Initially  we have   ` self.email = nil if ValidatesEmailFormatOf::validate_email_format(email)` meaning that email is blank if it invalid. In the case for the active course users we don't want this.

Question:

- Should I keep the initial implementation ` self.email = nil if ValidatesEmailFormatOf::validate_email_format(email)`  and return an error only if the user is an `active_course_instructor`

ragesoss: I think the current implementation of the validation should remain the same. The extra check for if a user is an active instructor should only be applied to the controller. (This might mean there are other ways of removing an email address for an active instructor, but doing so via the controller is the only way I'm aware of that has actually happened.)

Formasitchijoh: Alright thank you!

Formasitchijoh:
The failure here is unrelated to this change!

ragesoss:
Thanks!

PR #6586, opened 2025-12-12
Revision AI scores stats: add filter by namespace

gabina:
## What this PR does
This PR makes two main things:
- Fix percentage calculations for daily checks by namespace and likelihood distribution per day graphs.
- Add new filter by namespace in likelihood per day graphs.

In addition, this PR changes the approach used by both `ScoresTrendsGraph` and `ScoresTrendsInBinsGraph`. Before this PR, the backend calculated detailed values, sent them to the frontend, and the graphs were built using all the data provided. After this PR, the backend only calculates a main “historical scores” array. Derived data (such as `countByNamespace`) is now computed in the frontend and used to render the graphs.

This change was needed because the legend values in `ScoresTrendsInBinsGraph` are now calculated dynamically (since they're refreshed every time a filter is applied). Having those stats pre-calculated in the backend no longer worked. To keep consistency, the same behavior was applied to the `ScoresTrendsGraph` component. That's why `ScoresTrendsGraph` and `ScoresTrendsInBinsGraph` components only have two props now: `id` and `statsData`.

## AI usage
I used chatGPT to improve my PR description. I wanted to use AI to add the filter through the Vega spec but didn't work because chatGPT confused different Vega versions (we're currently using an old version). Instead of that, reading the Vega docs were the best solution. 

## Screenshots (fix percentage)
Before:
image
After:
image

## Screenshots (new filter)
Before:
no filter

After:
[Screencast from 2025-12-14 22-44-20.webm](https://github.com/user-attachments/assets/19d4ebf5-1e60-4b9e-bda9-ad7fd11e2431)

### DIFF
--- a/app/controllers/revision_ai_scores_stats_controller.rb
+++ b/app/controllers/revision_ai_scores_stats_controller.rb
@@ -16,9 +16,6 @@ class RevisionAiScoresStatsController < ApplicationController
     set_scores
     set_avg_likelihoods
     set_max_likelihoods
-    set_count_by_namespace
-    set_count_by_max_likelihood
-    set_count_by_avg_likelihood
     set_historical_scores_by_namespaces
     set_historical_scores_by_max_values
     set_historical_scores_by_avg_values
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/utils/article_utils.js
```diff
@@ -30,17 +30,21 @@ export default class ArticleUtils {
     0: 'main',
     1: 'talk',
     2: 'user',
+    3: 'user_talk',
```
gabina, line 33: the filter breaks if the namespace is not found. I'm not sure if there is any reason why the namespaces for talk weren't here.

app/controllers/revision_ai_scores_stats_controller.rb
```diff
@@ -71,56 +71,65 @@ def set_count_by_avg_likelihood
...
-  # even if the count is zero for some combinations.
+  # The array covers the full date range and all namespaces, even if the count is zero
+  # for some combinations.
+  # The array also contains a bin field that always has the value -1, since it’s not
+  # really necessary in this case.
   def set_historical_scores_by_namespaces
     count_by_date_and_namespace = @scores.group_by do |s|
-      [s.revision_datetime.to_date, s.article.namespace]
+      # -1 represents an invalid bin
+      [s.revision_datetime.to_date, s.article.namespace, -1]
```
gabina, line 81: I made this decision (keeping an unused `bin` field) to reuse the `complete_hash` function and to maintain a generic hash format for all historical data (by namespace and by likelihood).

ragesoss:
LGTM!

Issue #6568, opened 2025-11-18
CSV generation should require login, and username should be included in job log


ragesoss:
We have a variety of CSV data files that users can download, for both campaigns and individual courses. These go through `ReportsController`, the files are generated asyncronously via `ReportCsvWorker`, and a scheduled `CsvCleanupWorker` job is added after report generation to delete the file later on.

In Programs & Events Dashboard production, we currently have more than 2800 scheduled jobs for CSV cleanup, meaning a very large number of CSVs was generated recently. This was likely an unknown bot/scraper of some sort, and that sort of large-scale data request may have caused database downtime recently.

We should make a few adjustments to the CSV generation system:

* Require the user to be logged in for every CSV generation request (which we currently do only for campaign data, not course data).
* Include the username of the requester as a parameter for CSV generation and cleanup jobs, so that we can easily see who requested large numbers of files if this sort of mass request happens again.

Abishekcs:
Working on it

PR #6591, opened 2025-12-16
Part 1(#6568): Add username tracking to Sidekiq job logs

Abishekcs:
**Issue** #6568

**Part 2** #6605

## Note

**Feature:** `Include the username of the requester as a parameter for CSV generation and cleanup jobs, so that we can easily see who requested large numbers of files if this sort of mass request happens again.` - **From** #6568

**There were two ways to implement this feature:**

**1)** Simply passing `username` to `ReportCsvWorker#perform` via  `ReportCsvWorker.generate_csv`  will work. Leading to showing `username` as one of the **arguments** of `Arguments column` in `Current jobs table`. It works but it still seems a weird place to get info about who **requested CSV**.

screenshot-2025-12-17_21-46-28

**2)** Other options was to use (Which is implemented in this PR) [Sidekiq and Request-Specific Context](https://www.mikeperham.com/2022/07/29/sidekiq-and-request-specific-context/) which uses **Rails**  `ActiveSupport::CurrentAttributes` and stores `Required Details` in  `Context Column` of `Current Jobs Table`.

screenshot-2025-12-17_21-48-19

## What this PR does

Implements `Sidekiq::CurrentAttributes.persist` which uses **Rails** `ActiveSupport::CurrentAttributes` to keep a record of `username`  of `user` who requested a particular CSV.

## Resources Referred

- [Sidekiq and Request-Specific Context](https://www.mikeperham.com/2022/07/29/sidekiq-and-request-specific-context/)
- [Current attributes in Rails](https://fullstackheroes.com/tutorials/rails/current-attributes/)
- [Sidekiq::CurrentAttributes.persist](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/middleware/current_attributes.rb)
- [ActiveSupport::CurrentAttributes](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/current_attributes.rb)

## AI usage

`List of Non-ASCII characters - Latin with diacritics` was generated by AI (Claude(First 4) & ChatGPT(Last 4)) 

## Screenshots
`Before:`

- **ReportCsvWorker**

screenshot-2025-12-16_23-53-31

- **CsvCleanupWorker**

screenshot-2025-12-16_23-54-48

`After:`

- **ReportCsvWorker**

BackGroundJob

- **CsvCleanupWorker**

Cleanup

### DIFF
--- a/spec/support/username_test_helper.rb
+++ b/spec/support/username_test_helper.rb
@@ -4,10 +4,10 @@ module UsernameTestHelper
   def self.test_usernames
     [
       # Non-ASCII characters - Latin with diacritics
-      'josé',
-      'françois',
-      'müller',
-      'søren',
+      'José',
+      'François',
+      'Müller',
+      'Søren',
       'Łukasz',
       'İbrahim',
       'Αλέξανδρος',


Abishekcs:
@ragesoss ready for your review.

ragesoss:
@Abishekcs thanks! I think it would be wise to include a test with this change, and use a set of usernames that include a variety of non-ASCII characters, to make sure no valid usernames would cause errors.

We have some similar tests of various types of usernames in `replica_spec.rb` and `revision_data_manager_spec.rb`. It might be good to create a spec helper function that collects an array of such usernames so that it's easy to loop through and test for bugs based on username characters.

Abishekcs:
Sure, thanks for the review. I’ll add the tests.

Abishekcs:
I have add the required test case. I will open a second PR which implements `Require the user to be logged in for every CSV generation request `.

Abishekcs:
`List of Non-ASCII characters - Latin with diacritics` where taken from `claude and chatgpt`.(Updated in PR description)

spec/support/username_test_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module UsernameTestHelper
+  def self.test_usernames
+    [
+      # Non-ASCII characters - Latin with diacritics
+      'josé',
```
ragesoss: In most or all cases, MediaWiki requires a capital letter to start a username. Ideally, these should all be valid Wikipedia usernames.

Abishekcs: I will change first four of them to capital letter that will do ?

config/initializers/sidekiq_unique_jobs.rb
```diff
@@ -1,3 +1,12 @@
+require "sidekiq/middleware/current_attributes"
+
+# Use to_prepare to ensure Current class is loaded
+Rails.application.config.to_prepare do
+  # Persist Current attributes across Sidekiq Jobs
+  # Source: https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/middleware/current_attributes.rb
+  Sidekiq::CurrentAttributes.persist(SidekiqJobContext)
```
ragesoss, line 7: This seems like a more complex way of adding the username to the job than I expected. Why not simply add the username as one of the arguments for `ReportCsvWorker`?

ragesoss: nevermind, you answered this in the PR description.

Abishekcs: Context column seemed the right place to put details like `username` which required using `Sidekiq::CurrentAttributes`.

BackGroundJob

spec/support/username_test_helper.rb
ragesoss: Yes.

ragesoss:
Cool! I was skeptical about the Context approach, but after looking at the Sidekiq documentation on it, I think it's the right strategy.

config/initializers/sidekiq_unique_jobs.rb
Abishekcs: Putting in `Arguments Column` is much simpler but seemed weird  since `username` was extra log details which seemed better in the context column.

PR #6592, opened 2025-12-16
Restore rails 7.0.8

SanjanaSogimatt:
## What this PR does
Restores the Rails 7.0.8 upgrade that was previously reverted due to Passenger spawning errors in production.

The issue was caused by Passenger activating Ruby default gems before Bundler during application boot, which resulted in a Gem::LoadError for a default gem. This did not surface in local development or CI but caused the app to fail during Passenger spawn.

This PR reapplies the Rails upgrade and adds documentation explaining this Passenger behavior, including the recommended configuration (PassengerPreloadBundler on) to prevent similar issues in the future.

## AI usage
I used ChatGPT to help reason through the Passenger spawning error and understand the interaction between Passenger, Bundler, and Ruby default gems. I asked for explanations of the error logs, why the issue occurred
only in production. The final fix and configuration changes were verified manually.

## Screenshots
Before:
Not applicable - this change affects application boot under Passenger and does not introduce user-facing UI changes. Before this fix, the application failed to start in production due to Passenger spawning errors.

After:
Not applicable - the application now boots successfully under Passenger after restoring the Rails upgrade and applying the documented configuration
No open questions at this time.

### DIFF
--- a/docs/upgrade_dependencies.md
+++ b/docs/upgrade_dependencies.md
@@ -39,6 +39,7 @@ LoadModule passenger_module /home/sage/.rvm/gems/ruby-3.1.2/gems/passenger-6.0.1
 <IfModule mod_passenger.c>
   PassengerRoot /home/sage/.rvm/gems/ruby-3.1.2/gems/passenger-6.0.14
   PassengerDefaultRuby /home/sage/.rvm/gems/ruby-3.1.2/wrappers/ruby
+  PassengerPreloadBundler on
 </IfModule>
 
 ```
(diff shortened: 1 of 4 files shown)


ragesoss:
The commit message incorrectly says 7.0.9.

Also, it would be good to link to the github thread comment where we found the solution. In the body of that second commit message would be where I would put it.

docs/upgrade_dependencies.md
```diff
@@ -49,6 +49,33 @@ On non-web application servers (ie, servers just running Sidekiq processes):
 * pull the latest code and run `bundle install` (with the new Ruby version)
 * restart the Sidekiq processes
 
+### Note:
```
ragesoss: Since this particular issue is unlike to happen again, I think this necessary. However, the example apache configs (like the one above) in doc files should all be updated to include the preload setting.

ragesoss:
This set of changes looks good. Can you clean up the git history? This should probably be two commits: one with the Passenger changes, and another restoring the 7.0.8 changes from the previous PR.

PR #6593, opened 2025-12-16
Conditionally hide Wikidata stats tab 

### LINKED ISSUE #6556: Wikidata stats are shown on courses when they should not be
The Wikidata stats component is designed to show detailed stats for Wikidata edits. There are two situations where we are showing Wikidata edits when we shouldn't:

* If all the stats are zero, the component should not be displayed.
* If wikidata is removed as a tracked wiki, wikidata stats should not be shown.

Here's an example where both situations are happening:

https://outreachdashboard.wmflabs.org/courses/University_of_Nottingham/Nottingham_Feminist_Archives

Image

This course initially tracked wikidata, so the initial updates created wikidata stats (in the course's CourseStat record). No edits were made on Wikidata, so all these stats are zero. Subsequently, Wikidata was removed as a tracked wiki, so even if the stats weren't zero they should still have been removed.

lakshita10341:
## What this PR does
Fix #6556 : Conditionally hide Wikidata stats tab 

## Screenshots
Before:
Screenshot from 2025-12-17 01-26-13

After:
with 0 stats:
Screenshot from 2025-12-17 01-47-36
with untracked wikidata:
Screenshot from 2025-12-17 01-47-19

### DIFF
--- a/spec/features/detailed_course_stats_spec.rb
+++ b/spec/features/detailed_course_stats_spec.rb
@@ -43,15 +43,15 @@ describe 'Detailed course overview stats', type: :feature, js: true do
     # Clicking other tabs renders the respective stats data
     page.find('.tab', text: 'www.wikidata.org').click
     expect(page.find('.tab.active')).to have_content('www.wikidata.org')
-    expect(page.find('.content-container .title')).not_to have_content('en.wikibooks.org - Cookbook')
+    expect(page.find('.content-container .title'))
+      .not_to have_content('en.wikibooks.org - Cookbook')
     expect(page.find('.content-container .title')).to have_content('www.wikidata.org')
     expect(page.find('.content-container')).to have_content("Claims\n35")
 
     page.find('.tab', text: 'en.wikipedia.org - Mainspace').click
     expect(page.find('.tab.active')).to have_content('en.wikipedia.org - Mainspace')
-    # rubocop:disable Layout/LineLength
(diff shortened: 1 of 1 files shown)


lakshita10341:
@ragesoss PTAL!

app/assets/javascripts/components/common/overview_stats_tabs.jsx
```diff
@@ -44,7 +50,10 @@ const OverviewStatsTabs = ({ course, statistics }) => {
     index += 1;
   });
 
-  const content = <OverviewStatsContent course={course} content={statsList[currentTabId]} />;
+  const selectedTab = statsList[currentTabId] || statsList[0];
```
ragesoss, line 53: What is this for? In what situation would it show something for statsList[0] but not statsList[currentTabId]?

ragesoss:
Looks like a related test is failing. You'll need to update that spec. I recommending making sure there is one spec that looks for the rendered stats for a case where that's expected, and another for the case where they are hidden by this new change. (The latter test should fail before the change, and pass after.)

app/assets/javascripts/components/common/overview_stats_tabs.jsx
lakshita10341: in case, if currentTabId stats got undefined, then falling back to statsList[0] and if it fails then falling back to return null to prevent page crash

ragesoss:
LGTM!

PR #6588, opened 2025-12-16
Add Sentry message to debug mysql error due to 'article_id' being null

gabina:
## What this PR does
This is to research issue #6470.
It adds a Sentry message to log `mw_page_ids` and `mw_rev_ids` when `article_id` is null in `update_article_course_timeslices`. According to what I understand, that's the function where the `Mysql2::Error: Field 'article_id' doesn't have a default value
` comes from.
## AI usage
No AI usage.

## Open questions and concerns
< anything you learned that you want to share, or questions you're wondering about related to this PR >

PR #6594, opened 2025-12-18
Try to fix `Mysql2::Error: Field 'article_id' doesn't have a default value`

gabina:
## What this PR does
This PR is part of issue #6470.

After deploying #6588 to wiki edu dashboard, I was able to debug some specific cases that trigger this error. As we were discussing in PR #6521,  the problem is related to the following workflow:

> 1.fetch_revision_data_for_course imports revision data.
> 2.Queries for deleted: false articles.
> 3.resolve_duplicate_articles may mark some as deleted.
> 4.The article dictionary is built from stale data.
> 
> ```
>     # Retrieve article records
>     articles = Article.where(wiki_id: @wiki.id, deleted: false,
>                              mw_page_id: article_attributes.map { |a| a['mw_page_id'] })
> 
>     resolve_duplicate_articles(articles)
> 
>     # Prep: get a user dictionary for all users referred to by revisions.
>     users = user_dict_from_sub_data(all_sub_data)
> 
>     # Now get all the revisions
>     # We need a slightly different article dictionary format here
>     article_dict = articles.each_with_object({}) { |a, memo| memo[a.mw_page_id] = a.id }
> ```
> The article query runs twice: once before resolving duplicate articles and once after that. If an article is marked as deleted, it won’t have an mw_page_id in the dict. However, I guess that means there’s a non-deleted article record with the same wiki id, namespace, and title (but a different mw_page_id), and maybe that’s the record that should have been marked as deleted.

# Example
Article link: https://en.wikipedia.org/wiki/He-Yin_Zhen or https://en.wikipedia.org/wiki/He_Zhen_(anarchist) (which is a redirect).
mw_page_id: 44151232

Article rows in production related to the article move:

```
MariaDB [dashboard]> SELECT * FROM articles WHERE wiki_id = 1 AND (mw_page_id = 44151232 OR title = 'He-Yin_Zhen');
+-----------+---------------------+---------------------+---------------------+------------------+-----------+--------+---------------------+---------+----------+---------------+--------------------------+---------+------------+------------+
| id        | title               | updated_at          | created_at          | views_updated_at | namespace | rating | rating_updated_at   | deleted | language | average_views | average_views_updated_at | wiki_id | mw_page_id | index_hash |
+-----------+---------------------+---------------------+---------------------+------------------+-----------+--------+---------------------+---------+----------+---------------+--------------------------+---------+------------+------------+
|  52565563 | He_Zhen_(anarchist) | 2022-03-13 11:37:08 | 2020-08-07 16:12:00 | NULL             |         0 | c      | 2022-03-13 11:37:08 |       0 | NULL     |       30.8367 | 2022-03-10               |       1 |   44151232 | 44151232-1 |
|  54949479 | He-Yin_Zhen         | 2025-12-18 02:41:28 | 2021-03-21 01:38:17 | NULL             |         1 | NULL   | NULL                |       0 | NULL     |          NULL | NULL                     |       1 |   48520836 | 48520836-1 |
| 171818780 | He-Yin_Zhen         | 2025-12-01 22:58:33 | 2022-01-11 03:29:22 | NULL             |         0 | c      | 2025-12-01 22:58:33 |       0 | NULL     |          NULL | NULL                     |       1 |   44423310 | 44423310-1 |
+-----------+---------------------+---------------------+---------------------+------------------+-----------+--------+---------------------+---------+----------+---------------+--------------------------+---------+------------+------------+

```

The article record with mw_page_id = 44151232 has an outdated title (`He_Zhen_(anarchist)`). When we process revisions like [1325758882](https://en.wikipedia.org/w/index.php?title=He-Yin_Zhen&oldid=1325758882), we re-import the article, updating the title to `He-Yin_Zhen`. However, when we call `resolve_duplicate_articles` for that title, namespace and wiki, we end up marking it as deleted because there is a newer article record (with a greater `created_at`) for the same title, namespace and wiki. That causes the RevisionOnMemory to be created with a nil article_id.

In an example like this one, I'm not sure resolving duplicate makes much sense, since we match revisions using `mw_page_id`. If I recall correctly, resolving duplicates by title, namespace and wiki is important for article scoped programs, where we may need to match revisions with articles based on titles. In any case, an easy way to fix this particular problem seems to be order duplicate articles by the `updated_at` field instead of `created_at`. That way, we ensure we don't mark an article as deleted if its title or namespace was recently updated. I'm not sure if there are any consequences to this change, but I think it’s worth trying.

## AI usage
No AI usage.

### DIFF
--- a/spec/lib/duplicate_article_deleter_spec.rb
+++ b/spec/lib/duplicate_article_deleter_spec.rb
@@ -36,17 +36,19 @@ describe DuplicateArticleDeleter do
              id: 2262715,
              title: 'Kostanay',
              namespace: 0,
-             created_at: 1.day.from_now)
+             created_at: 1.day.from_now,
+             updated_at: 1.day.from_now)
     end
     let!(:duplicate_article) do
       create(:article,
              id: 46349871,
              title: 'Kostanay',
              namespace: 0,
-             created_at: 1.day.ago)
+             created_at: 1.day.ago,
+             updated_at: 1.day.ago)
     end
 
-    it 'marks oldest one deleted when there are two ids for one page' do
+    it 'marks the one updated the longest time ago deleted when there are two ids for one page' do
       described_class.new.resolve_duplicates_for_timeslices([new_article])
       undeleted = Article.where(
         title: 'Kostanay',


gabina:
@ragesoss WDYT about this?

ragesoss:
LGTM

gabina:
It seems to have worked in wiki edu dashboard, since the last `Mysql2::Error: Field 'article_id' doesn't have a default value` log was 20 hours ago.

PR #6599, opened 2025-12-22
fix: post instructor userpage template on course approval

### LINKED ISSUE #6589: instructor userpage template should not be posted until course is approved
Currently, we create an on-wiki course page at the time a course is submitted, and also post a userpage template for the submitting instructor linking to that page. This can create a misleading impression that Wiki Education is supporting an event or course that does not get approved but happens independently of Wiki Education support. (This happened recently with an edit-a-thon, where the organizer first tried to set it up on Wiki Education Dashboard.)

The instructor's userpage template should instead get added at the time a course is approved.

lakshita10341:
## What this PR does
Fix #6589 :  changed the announcement timing from instructor course submission to course approval.

## AI usage
to verify the correctness of logic and writing tests to verify the changes.

I've verified the changes by running tests which executed successfully.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -279,8 +279,9 @@ class CoursesController < ApplicationController
   end
 
   def handle_course_announcement(instructor)
-    # Course announcements aren't particularly necessary, but we'll keep them on
-    # for Wiki Ed for now.
+    # Only send submission emails at submission time.
+    # Instructor userpage templates are now posted when the course is approved
+    # (via ListCourseManager) instead of at submission time.
     return unless Features.wiki_ed?
     newly_submitted = !@course.submitted? && course_params[:submitted] == true
     return unless newly_submitted
--- a/app/workers/list_course_worker.rb
+++ b/app/workers/list_course_worker.rb
@@ -1,0 +1,22 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/wiki_course_edits"
+
+class ListCourseWorker
+  include Sidekiq::Worker
+  sidekiq_options lock: :until_executed
+
+  def self.schedule_edits(course:, editing_user:, instructor:)
+    perform_async(course.id, editing_user.id, instructor.id)
+  end
+
+  def perform(course_id, editing_user_id, instructor_id)
+    course = Course.find(course_id)
+    editing_user = User.find(editing_user_id)
+    instructor = User.find(instructor_id)
+    WikiCourseEdits.new(action: 'add_course_template_to_instructor_userpage',
+                        course:,
+                        current_user: editing_user,
+                        instructor:)
+  end
+end
--- a/lib/list_course_manager.rb
+++ b/lib/list_course_manager.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_dependency "#{Rails.root}/app/workers/list_course_worker"
+
 #= Routines for adding or removing a course to/from a campaign
 class ListCourseManager
   def initialize(course, campaign)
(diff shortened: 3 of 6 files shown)


app/controllers/courses_controller.rb
```diff
@@ -279,18 +279,16 @@ def validate
...
+    # Instructor userpage templates are now posted when the course is approved
+    # (via ListCourseManager) instead of at submission time.
     return unless Features.wiki_ed?
     newly_submitted = !@course.submitted? && course_params[:submitted] == true
     return unless newly_submitted
     # Needs to be switched to submitted before the announcement edits are made
     @course.update(submitted: true)
     AddSubmittedTag.new(@course)
     CourseSubmissionMailerWorker.schedule_email(@course, instructor)
-    AnnounceCourseWorker.schedule_announcement(course: @course,
```
ragesoss, line 297: AnnounceCourseWorker makes two edits, one to a page for announcing new courses and the other to add a template to the instructor's talk page. We only want to change the timing of the template, not the new course announcement.

lakshita10341: I have refactored the code by making the logic accept an action argument. Now, the 'Announce Course' action is called at course submission, and the 'Userpage Template' action is posted at course approval.

ragesoss:
Thanks Lakshita! I think this is heading in the right direction. I think it would be best to make the two edits more fully independent, rather than using `AnnounceCourseWorker` in two different ways. Fundamentally, the `announce_course` course action is the only one that is part of the the named role of that worker, and we were previously just including the userpage template for convenience.

This might be a good opportunity to create a new job to represent async tasks for listing a course (eg, `ListCourseWorker`) to do the userpage template.

lakshita10341:
> Thanks Lakshita! I think this is heading in the right direction. I think it would be best to make the two edits more fully independent, rather than using `AnnounceCourseWorker` in two different ways. Fundamentally, the `announce_course` course action is the only one that is part of the the named role of that worker, and we were previously just including the userpage template for convenience.
> 
> This might be a good opportunity to create a new job to represent async tasks for listing a course (eg, `ListCourseWorker`) to do the userpage template.

Thanks @ragesoss for guidance. I agree that decoupling these tasks would be a cleaner approach. So I have refactored AnnounceCourseWorker to handle only course announcements and have created a new ListCourseWorker for updating userpage template. I guess this will resolve the issue and would be a cleaner approach.

app/workers/announce_course_worker.rb
```diff
@@ -14,7 +14,7 @@ def perform(course_id, editing_user_id, instructor_id)
     course = Course.find(course_id)
     editing_user = User.find(editing_user_id)
     instructor = User.find(instructor_id)
-    WikiCourseEdits.new(action: 'announce_course',
+    WikiCourseEdits.new(action: 'announce_course_on_announcement_page',
```
ragesoss: I think the existing action name (announce_course) is fine, it just needs to be updated to only do the one edit.

ragesoss: (I think the current implementation leaves `announce_course` as dead code?)

lakshita10341: Yes, correct! I have changed the action to `announce_course` and removed `add_course_template_to_instructor_userpage` action from it.

ragesoss:
In broad strokes, I think this looks good! I will take a closer look soon, and it should be ready to merge if I don't spot any problems.

ragesoss:
This looks good to me. Please clean up the git history; it should include only commits directly related to implementing the final shape of this changeset (either a single commit, or a set of commits that build on one another to progressively tell the story of this change).s

lakshita10341:
Hi @ragesoss, I have cleaned up the git history. PTAL if any changes are required.

PR #6607, opened 2025-12-31
Fix Failing ReportsController spec for unauthenticated users and fix linting offense in AiEditAlertMailer

Abishekcs:
## What this PR does

Test Case for `ReportsController#set_sidekiq_job_context when user is not signed in does not set the SidekiqJobContext username` was failing after **#6605** got merged as it added `authentication for all CSV report generation` which was not consider while adding this new test case in #6591

**Also, Fix linting offense in** `app/mailers/ai_edit_alert_mailer.rb`

## AI usage

No AI usage

## Screenshots

Before:

screenshot-2025-12-31_16-20-13

After:

screenshot-2025-12-23_17-16-53

### DIFF
--- a/app/mailers/ai_edit_alert_mailer.rb
+++ b/app/mailers/ai_edit_alert_mailer.rb
@@ -11,18 +11,7 @@ class AiEditAlertMailer < ApplicationMailer
     @course = @alert.course
     return unless @course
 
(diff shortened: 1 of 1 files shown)


ragesoss:
Thanks! I copied this fix over to a new commit.

PR #6612, opened 2026-01-04
feat: add "Update Scheduled" state to admin course actions

### LINKED ISSUE #6604: Admins need a way to schedule a 'full update' for any course
Courses sometimes have data problems where the easiest solution is to regenerate the stats by doing a full update (fetching and recalculating the stats for the entire timespan of the course). Admins should have a well-labeled and clear way to initiate this (and to see whether it's already been initiated) from the frontend.

lakshita10341:
## Fixes #6604 : Enable admin to schedule updates
This PR allows admins to schedule a full update of data for any course, independent of its completion status. It also improves the user experience by immediately disabling the button and changing the label to "Update Scheduled" when an update is pending.

## AI usage
used copilot to locate the correct file and understand the flow.

## Screenshots
Before:
image
(Earlier, with admin account)

After:
image 
image

image


app/assets/javascripts/actions/course_actions.js
```diff
@@ -99,7 +99,10 @@ const needsUpdateNotification = (response) => {
 export function needsUpdate(courseSlug) {
   return function (dispatch) {
     return needsUpdatePromise(courseSlug)
-      .then(resp => dispatch({ type: ADD_NOTIFICATION, notification: needsUpdateNotification(resp) }))
+      .then((resp) => {
+        dispatch({ type: ADD_NOTIFICATION, notification: needsUpdateNotification(resp) });
+        dispatch({ type: UPDATE_COURSE, course: { needs_update: true } });
```
ragesoss, line 104: This looks like it will work, but I think a more robust strategy would be to re-fetch the course.json (so that the `needs_update` value is derived from backend state).

ragesoss:
@gabina is setting `needs_update: true` sufficient to trigger rebuilding all the timeslice data for a course, or is there more than needs to be done?

gabina:
> @gabina is setting `needs_update: true` sufficient to trigger rebuilding all the timeslice data for a course, or is there more than needs to be done?

Setting `needs_update: true` is enough. Courses with `needs_update` set to `true` are picked by the `enqueue_course_updates` and queued for update (even if they're not ongoing courses). When  `UpdateCourseStats` finally runs, it recreates all the timeslices with `needs_update: 1`, since the `full_update` parameter is based on `needs_update` field.

ragesoss:
I'm going to go ahead and merge this now, and I'll open an issue for broader improvements to how we differentiate between courses that really need a full update versus a request to pull in new stats after the end of a course.

PR #6613, opened 2026-01-04
[WIP] Add backend system to automatic backup

gabina:
## What this PR does
This is a work in progress.

This PR adds a system to perform automatic database backups periodically. It is intended to work in the following way:
1. **On DB server.** Weekly, a cron job runs on the db server, in charge of performing the backup itself. The first task it performs is creating a new backup record with status set to `waiting`. This works as a way to let the backend app knowing that a backup is waiting to run.
2. **On app servers.** At this point, it's expected that some Sidekiq jobs are running and many others are queued. The  running jobs will continue until completion, but once they finish, new `CourseDataUpdateWorker` jobs will start and will be skipped due to the `BackupPauseMiddleware` added to the chain. At this stage, the app knows a backup is waiting to run. Other workers (for example `ConstantUpdateWorker`) will work as usual.
3.  **On DB server.** It checks the new `can_start_backup.json` endpoint to determine whether there are any  jobs currently running. At some point, no jobs should be running, since slow jobs are usually only `CourseDataUpdateWorker`, the other ones runs almost immediately. If the response is 503 server unavailable, it waits for a certain amount of time (to be defined) and retries until it receives a 200 OK response.
4. **On DB server.** It updates the data table record status to `running`.
5.  **On DB server.** It runs the backup itself.
6.  **On DB server.** It updates the data table record status to `finished`.
7. **On app servers.** Sidekiq jobs enqueued after this point will no longer be affected by the middleware, and the application returns to normal operation.

**Important notes**
- Queue latency would be missed with this approach.
- We have to rely on the sidekiq reaper to clean orphans locks from sidekiq-uniq-jobs gem. Since the `BackupPauseMiddleware` doesn't call `yield` for `CourseDataUpdateWorker` if backup is running, `peform` is never invoked, so the lock for that job becomes orphan.

**Conclusion**
A similar approach could be implemented through a new `BackupSkip` module so we don't need to rely on the sidekiq reaper and no orphan locks are generated.

## AI usage
< if you used any AI tools to prepare this PR, say which tools you used and what you used them for.
  if you used AI to learn how to solve a problem, summarize what you asked and what advice you used.
  if you didn't use any AI tools, say so. >

## Screenshots
Before:
< add a screenshot or screen recording of the UI before your change >

After:
< add a screenshot or screen recording of the UI after your change >
The rescheduling approach didn't work, since jobs were re-scheduled only once. Initially, all jobs went to "Scheduled Jobs" but then they disappear without running due to the unique-job gem. Since the middleware didn't run, then no job was re-scheduled.

### DIFF
--- a/app/workers/course_data_update_worker.rb
+++ b/app/workers/course_data_update_worker.rb
@@ -12,6 +12,10 @@ class CourseDataUpdateWorker
     CourseDataUpdateWorker.set(queue:).perform_async(course_id)
   end
 
+  def self.skippable_during_backup?
+    true
+  end
+
   def perform(course_id)
     course = Course.find(course_id)
     logger.info "Ignoring #{course.slug} update" if course.very_long_update?
--- a/config/initializers/sidekiq_backup.rb
+++ b/config/initializers/sidekiq_backup.rb
@@ -1,13 +1,14 @@
+# frozen_string_literal: true
 class BackupPauseMiddleware
-  RESCHEDULE_DELAY = 1.hour
-
   def call(worker, job, queue)
-    if Backup.in_process?
-      worker.class.perform_in(RESCHEDULE_DELAY, *job['args'])
-      return
-    end
+    return if skippable?(worker) && Backup.in_process?
 
     yield
+  end
+
+  def skippable?(worker)
+    worker.class.respond_to?(:skippable_during_backup?) &&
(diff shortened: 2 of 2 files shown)


gabina:
Closing this in favor of https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6625

Issue #6590, opened 2025-12-16
Add a convenient way to break the cache on-demand for campaign stats


ragesoss:
During the busiest period of each term, we like to check updated stats for the campaign (especially words added) several times per day. The current caching strategy means that these numbers only update once per day. (See the implementation in `CoursesPresenter#course_sums`.) This caching strategy was implemented to help with slow queries on Programs & Events Dashboard, but it's not a huge problem and would be fine to bypass manually when someone wants fresher stats.

It should be possible to break or bypass this cache when desired, either by just adding a parameter, or with some admin-only button to clear the cache.

Formasitchijoh:
Working on this

PR #6615, opened 2026-01-05
Check updated stats for the campaigns

Formasitchijoh:
## What this PR does
This PR resolves the task #6590  by adding a convenient way of clearing the cache of campaign stats and recalculating and fetching new stats

## AI usage
Used AI to determine the best approach for fixing the long paramater list error in the course presenter initializer

## Screenshots
Here is a video of the current implementation https://www.loom.com/share/f35cce5de7f34826b37ea196d38ebe7f

### DIFF
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -138,7 +138,7 @@ class Campaign < ApplicationRecord
   end
 
   def course_sums_cache_key
-    "#{slug}-course_sums_v2"
+    "#{slug}-course_sums"
   end
 
   def clear_course_sums_cache


Formasitchijoh:
@ragesoss Please could you review this

ragesoss:
This looks good! I don't think I like the button, though, as it adds more clutter to the UI for something that very few people will need. I think a good solution will be to remove the frontend changes, change the route to `/campaigns/:slug/refresh` and make the controller method handle the redirection. That way, you just have to know to add `/refresh` to the end in order to trigger a refresh.

app/controllers/campaigns_controller.rb
```diff
@@ -227,6 +227,19 @@ def current_term
...
   end
 
+  def refetch_stats
+    set_campaign
+    presenter = CoursesPresenter.new(
+      current_user: current_user,
+      campaign_param: @campaign.slug,
+      reset_cache: true
+    )
+    presenter.course_sums
```
ragesoss: Does calling this set the new cache? If not, there's no point in calling it here, as the redirect will handling rebuilding the cache and this presenter won't be used for anything.

It might not be ideal to have the presenter be responsible for clearing the cache; perhaps the `Campaign` model itself should be responsible for determining the cache key, and clearing it's own cache.

Formasitchijoh: Yes calling that clears the cache since the presenters `course_sums `function is  responsible for creating, storing and retrieving the cached stats, I though it wise to clear it if the status  of `reset_cache `is true before recalculating the new stats

Formasitchijoh: > It might not be ideal to have the presenter be responsible for clearing the cache; perhaps the `Campaign` model itself should be responsible for determining the cache key, and clearing it's own cache.

should I ensure the Campaign is the one responsible for clearing the cache

ragesoss: Yes, I think so. What I recommend is to implement two `Campaign` instance methods `:course_sums_cache_key` and `:clear_course_sums_cache`. Then the controller action should call the clear method (instead of using CoursesPresenter), and CoursesPresenter should call the cache_key method to determine the key to fetch (instead of the current implementation where the cache key is determined within the presenter code based on the campaign slug).

app/presenters/courses_presenter.rb
```diff
@@ -7,11 +7,11 @@
 class CoursesPresenter
   attr_reader :current_user, :campaign_param
 
-  def initialize(current_user:, campaign_param: nil, courses_list: nil, page: nil, tag: nil)
+  def initialize(current_user:, campaign_param: nil, courses_list: nil, **options)
```
Formasitchijoh, line 10: Left this implementation in case more than 5 parameters need to be passed to this class in the future. Without the `**options`  paramater, if more than 5 paramaters are passed,  it returns the warning
`Metrics/ParameterLists: Avoid parameter lists longer than 5 parameters.` which causes the build to fail when running linting using rubocop.

ragesoss: This is reasonable. The alternative would be to disable that rule for this particular method.

How did you test against regressions for the usage of the `page` and `tag` parameters? (Are these already used in tests in a way that would break if it was not compatible with the `**option` version?)

app/models/campaign.rb
```diff
@@ -137,6 +137,14 @@ def course_string_prefix
     @course_string_prefix ||= default_course_type.constantize.new.string_prefix
   end
 
+  def course_sums_cache_key
+    "#{slug}-course_sums_v2"
```
ragesoss: I'm not sure why we call this `v2`. Is there another place in the codebase that uses a similar `course_sums` cache key? If not, it might be a good opportunity to clean up the naming slightly and leave out the `_v2` at the end.

app/presenters/courses_presenter.rb
Formasitchijoh: Yes I tested for page in the campaign controller
The campaign_controller calls the `CoursePresenter `and pass it a `page attribute` in the `set_presenter `method which is used in a couple of methods in the controller like update, articles . I ran the test for the `campaign_controller_spec` which calls the `CoursePresenter `and all test passed 

Screenshot 2026-01-12 at 19 30 19

app/models/campaign.rb
Formasitchijoh: No other place uses the v2, I will clean up the naming

app/presenters/courses_presenter.rb
Formasitchijoh: There are no test for the tagged_controller that pass the tag parameter to the CoursePresenter, I can write to ensure all cases are met

PR #6621, opened 2026-01-07
Rails 7.1 Configuration Updates

SanjanaSogimatt:
## Overview
This PR updates application configuration files to be compatible with Rails 7.1, addressing deprecation warnings and ensuring proper cache format versioning.

### 1. Update `show_exceptions` Configuration in Test Environment

**File:** `config/environments/test.rb`

**Change:**
```ruby
### # Before
config.action_dispatch.show_exceptions = false

### After
### Rails 7.1: Use :none instead of false
config.action_dispatch.show_exceptions = :none
```

**Why:**
- Rails 7.1 deprecated the boolean `false` value in favor of the symbol `:none` for consistency with other Rails configuration values
- This change eliminates deprecation warnings: `Setting action_dispatch.show_exceptions to false is deprecated. Set to :none instead.`
- The `:none` value is semantically equivalent to `false` but follows Rails 7.1's standardized configuration pattern
- Prevents future breaking changes when the boolean value is removed

**References:**
- [Rails 7.1 Release Notes](https://github.com/rails/rails/releases/tag/v7.1.0)

---

### 2. Update Cache Store Configuration and Format Version

**File:** `config/application.rb`

**Changes:**

**2.1. Cache Store Pool Configuration**
```ruby
### Before
config.cache_store = :mem_cache_store, 'localhost', { pool_size: 5, ... }

### After
config.cache_store = :mem_cache_store, 'localhost', { pool: { size: 5 }, ... }
```

**Why:**
- Rails 7.1 changed the cache store configuration syntax
- The `pool_size` option is now nested under `pool: { size: ... }` for better structure and consistency
- This ensures the connection pool is properly configured for memcached

**2.2. Cache Format Version**
```ruby
# Added
config.active_support.cache_format_version = 7.1
```

**Why:**
- Rails 7.1 introduced a new cache format version (7.1) that is more efficient and secure
- Updating the cache format version ensures compatibility with Rails 7.1's caching improvements
- This prevents potential cache-related issues and ensures optimal performance
- The format version must be explicitly set when upgrading to maintain cache compatibility

**References:**
- [Rails 7.1 Cache Format Version](https://github.com/rails/rails/releases/tag/v7.1.0)
- [ActiveSupport Cache Store Configuration](https://api.rubyonrails.org/classes/ActiveSupport/Cache/Store.html)

---

### 3. Remove ActiveSupport Require from spec_helper.rb

**File:** `spec/spec_helper.rb`

**Change:**
```ruby
### # Before
require 'active_support'

### After
# active_support/testing/time_helpers moved to rails_helper.rb (after Rails loads)
```

**Why:**
- `active_support` should not be required in `spec_helper.rb` because it loads before Rails initializes
- ActiveSupport dependencies (like `TimeHelpers`) must be loaded after Rails is fully initialized
- This prevents initialization errors and ensures proper loading order
- The comment clarifies that time helpers are now properly configured in `rails_helper.rb` where Rails is already loaded

**Impact:**
- Prevents potential `NoMethodError` issues when ActiveSupport methods are called before Rails initialization
- Ensures test helpers are available at the correct time in the test lifecycle

---

### 4. Update Gemfile.lock Dependencies

**File:** `Gemfile.lock`

**Changes:**
- Updated `ffi` gem: Removed platform-specific build (`1.17.3-arm64-darwin` → `1.17.3`)
- Updated `glob` gem: `0.4.0` → `0.4.1`
- Updated `io-console` gem: `0.8.2` → `0.7.2`
- Updated `nokogiri` gem: Removed platform-specific build (`1.18.10-arm64-darwin` → `1.18.10`)
- Added `mini_portile2` gem: `2.8.9` (dependency for nokogiri)

**Why:**
- Rails 7.1 and updated dependencies require compatible gem versions
- Platform-specific builds are no longer needed as gems are now properly cross-platform compatible
- `mini_portile2` is now an explicit dependency for `nokogiri` compilation
- These updates ensure compatibility with Rails 7.1 and resolve potential dependency conflicts

**Impact:**
- Ensures all gems are compatible with Rails 7.1
- Resolves platform-specific build issues
- Maintains proper dependency resolution

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,14 @@ jobs:
         with:
            ruby-version: 3.1.2
            bundler-cache: true
+      
+      - name: Install system libraries for nokogiri
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libxml2-dev libxslt1-dev zlib1g-dev
+      
+      - name: Configure nokogiri to use system libraries
+        run: bundle config set --local build.nokogiri --use-system-libraries
      
       - name: setup Node
         uses: actions/setup-node@v3


lib/training_module_due_date_manager.rb
```diff
@@ -64,7 +69,7 @@ def blocks_with_training_modules_for_user
     return [] unless @user.present?
     Block.joins(week: { course: :courses_users })
          .where(courses_users: { user_id: @user.id, role: CoursesUsers::Roles::STUDENT_ROLE })
-         .where.not('training_module_ids = ?', [].to_yaml).includes(:week)
+         .where.not('training_module_ids = ?', [].to_yaml).includes(week: :course)
```
Abishekcs: Was this added to avoid `n+1` ?

SanjanaSogimatt: yes

app/helpers/application_helper.rb
```diff
@@ -71,4 +71,12 @@ def body_class(request)
     return 'survey-page' if survey_paths.include?(base_path)
     return 'fixed-nav'
   end
+
+  def page_class
```
Abishekcs: I think you can replace `page_class method` with `= hot_javascript_tag("survey") unless controller_name == 'survey_results' && action_name == 'results'` in `surveys.html.haml`

Abishekcs: Another good reason to remove `page_class` it eliminates confusion about why the method exists and where it's actually being used.

ragesoss: I agree, with such a narrow use case, better to do this explicitly in the one file that uses it instead of implementing a new helper method.

lib/training_module_due_date_manager.rb
```diff
@@ -34,7 +34,12 @@ def deadline_status
   # courses where module is assigned)
   def overall_due_date
     blocks = blocks_with_module_assigned(@training_module)
-    blocks.collect(&:calculated_due_date).min
+    return nil if blocks.empty?
```
Abishekcs: By the way changes in these file was made because test case was failing ?

app/controllers/requested_accounts_campaigns_controller.rb
```diff
@@ -11,12 +11,12 @@ class RequestedAccountsCampaignsController < ApplicationController
   # modal instead of redirecting to the mediawiki account creation flow.
   def enable_account_requests
     @campaign.update(register_accounts: true)
-    redirect_back(fallback_location: root_path)
+    redirect_back_or_to(root_path)
```
ragesoss, line 14: I think these changes to use `redirect_back_or_to` can be implemented in a separate PR before the upgrade. It looks like this method is available in 7.0: https://apidock.com/rails/v7.0.0/ActionController/Redirecting/redirect_back_or_to

app/controllers/settings_controller.rb
```diff
@@ -218,7 +218,7 @@ def attempt_special_user_downgrade
     # Check if the user already has the position
     unless SpecialUsers.is?(@user, @position)
       message = I18n.t(
-        'settings.special_users.new.already_is_not',
+        'settings.special_users.remove.already_is_not',
```
ragesoss: What's this for? Can it be implemented before the upgrade?

app/helpers/course_helper.rb
```diff
@@ -40,7 +40,7 @@ def format_course_stats(course_stats)
       end
       # convert stats to human readable values
       course_stats[wiki_ns_key].each do |stat_key, stat|
-        course_stats[wiki_ns_key][stat_key] = number_to_human(stat)
+        course_stats[wiki_ns_key][stat_key] = number_to_human(stat).to_s
```
ragesoss: What behavior change makes this necessary? I would expect `number_to_human` to already return a string, but if it's a case where now it returns `nil` when it didn't before (or something along those lines), the `.to_s` can probably be implemented separately before the upgrade.

app/presenters/courses_presenter.rb
ragesoss, line 10: There are changes from a merge, so they won't be part of a clean final upgrade PR. (The goal is to make it really easy to look at the final PR and see changes that are necessary during the upgrade, with no unrelated changes.)

app/services/system_metrics.rb
```diff
@@ -5,7 +5,8 @@
 class SystemMetrics
   def initialize
     # 'very_long_update' queue is excluded as it is intentionally never processed
-    @queues = YAML.load_file('config/sidekiq.yml')[:queues]
+    # Rails 7.1/Psych: Enable aliases to support YAML anchors (e.g., &defaults, *defaults)
+    @queues = YAML.load(File.read('config/sidekiq.yml'), aliases: true)[:queues]
```
ragesoss: Can this be done ahead of the upgrade?

app/views/layouts/surveys.html.haml
```diff
@@ -19,7 +19,7 @@
     = render "shared/foot"
     // survey.js sometimes interferes with the rendering of results.
     = javascript_include_tag '/assets/javascripts/jquery.min.js'
-    = hot_javascript_tag("survey") unless page_class == 'surveys results'
+    = hot_javascript_tag("survey") unless controller_name == 'survey_results' && action_name == 'results'
```
ragesoss, line 22: I think this can be done before the upgrade.

app/workers/ai_detection_worker.rb
```diff
@@ -9,12 +9,12 @@ class AiDetectionWorker
   def self.schedule_check(wiki:, revision:, course:)
     # revision object is a RevisionOnMemory and sidekiq good practices
     # suggest keeping job parameters small, simple, and JSON-compatible
-    attributes = { mw_rev_id: revision.mw_rev_id,
-                   wiki_id: wiki.id,
-                   article_id: revision.article_id,
-                   course_id: course.id,
-                   user_id: revision.user_id,
-                   revision_timestamp: revision.timestamp }
+    attributes = { 'mw_rev_id' => revision.mw_rev_id,
```
ragesoss: What makes this necessary?

config/environments/test.rb
```diff
@@ -76,7 +76,8 @@ def dump_js_coverage
   config.action_controller.perform_caching = false
 
   # Raise exceptions instead of rendering exception templates.
-  config.action_dispatch.show_exceptions = false
+  # Rails 7.1: Use :none instead of false
+  config.action_dispatch.show_exceptions = :none
```
ragesoss, line 80: Can this be done before the upgrade?

config/application.rb
```diff
@@ -54,12 +54,15 @@ class Application < Rails::Application
...
+
+    # Rails 7.1: Update cache format version from 6.1 to 7.1
+    config.active_support.cache_format_version = 7.1
 
     # Handle YAML safe loading of serialized Ruby objects
-    config.active_record.yaml_column_permitted_classes = [Symbol, BigDecimal, DateTime, Date, Time, ActiveSupport::TimeWithZone, ActiveSupport::TimeZone]
+    config.active_record.yaml_column_permitted_classes = [Symbol, BigDecimal, DateTime, Date, Time, ActiveSupport::TimeWithZone, ActiveSupport::TimeZone, ActiveSupport::SafeBuffer]
 
-    config.active_record.legacy_connection_handling = false
+    # legacy_connection_handling was removed in Rails 7.1
```
ragesoss: I think this is already deprecated in 7.0 and can be safely removed before the upgrade.

spec/lib/training_module_due_date_manager_spec.rb
```diff
@@ -191,9 +191,9 @@
                      .overall_due_date
     end
 
-    before(:all) { travel_to Date.new(2015, 8, 25) }
+    before(:each) { travel_to Date.new(2015, 8, 25) }
```
ragesoss: What's this for? These changes can likely be implemented before the upgrade.

app/workers/ai_detection_worker.rb
SanjanaSogimatt: And this was the reason the build was failing as this threw error

spec/spec_helper.rb
```diff
@@ -23,7 +23,7 @@
 require 'rspec/core' unless defined? RSpec.configure
 require 'webmock/rspec'
 require './spec/support/request_helpers'
-require 'active_support/testing/time_helpers'
+# active_support/testing/time_helpers moved to rails_helper.rb (after Rails loads)
```
ragesoss: Probably this can be done before the upgrade.

ragesoss:
Please make sure this starts from a clean version of `master`. It looks like it duplicates some of the changes I made to get io-console working.

spec/spec_helper.rb
ragesoss: This `require` just got added in the 7.1 PR. Was it not necessary here, since the time helpers were put in rails_helper?

In any case, I don't think we need to leave behind a comment, since it's unlikely to be relevant for future work.

PR #6622, opened 2026-01-07
Refactor serialization in multiple models to specify type explicitly

SanjanaSogimatt:
Updated the serialization of various attributes across multiple models to include the `type:` option for clarity and consistency. This change affects the following models: Alert, ArticleCourseTimeslice, CourseUserWikiTimeslice, CourseWikiTimeslice, Course, RevisionAiScore, Setting, TrainingLibrary, TrainingModule, TrainingSlide, ArticlesCourses, Assignment, Block, CourseStat, SurveyAssignment, TrainingModulesUsers, UserProfile, Category, Revision, and Rapidfire::Question.

## What this PR does

This PR addresses rubocop erros encountered during the Rails upgrade from 7.0.8 to 7.0.10, and prepares the codebase for Rails 7.1. The changes include:

RuboCop configuration incompatibilities
Addressed RuboCop errors caused by plugin extraction (rubocop-rspec-rails, rubocop-factory_bot, rubocop-capybara) and migrated configuration to the supported plugin-based format.

## AI usage

AI tools were used to assist with:

Understanding and interpreting Rails 7.1 deprecations and upgrade-related warning

###  Serialize Syntax Updates (Rails 7.1)

```



**Files**: Multiple model files including:
- `app/models/alert.rb`
- `app/models/course.rb`
- `app/models/article_course_timeslice.rb`
- `app/models/course_data/articles_courses.rb`
- `app/models/course_data/assignment.rb`
- `app/models/course_data/block.rb`
- `app/models/course_data/course_stat.rb`
- `app/models/course_user_wiki_timeslice.rb`
- `app/models/course_wiki_timeslice.rb`
- `app/models/revision_ai_score.rb`
- `app/models/setting.rb`
- `app/models/surveys/survey_assignment.rb`
- `app/models/training_library.rb`
- `app/models/training_module.rb`
- `app/models/training_slide.rb`
- `app/models/user_data/training_modules_users.rb`
- `app/models/user_data/user_profile.rb`
- `app/models/wiki_content/category.rb`
- `app/models/wiki_content/revision.rb`

**Change**:
```ruby
# Before
serialize :details, Hash
serialize :flags, Hash

# After
serialize :details, type: Hash
serialize :flags, type: Hash
```

**Why**: Rails 7.1 updated the `serialize` method to use keyword arguments (`type:`) instead of positional arguments for better clarity, consistency, and to support additional options.

**Documentation**:
[- [ActiveRecord::AttributeMethods::Serialization](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize)](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html#method-i-serialize)

### DIFF
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -38,7 +38,7 @@ class ArticlesCourses < ApplicationRecord
   scope :tracked, -> { where(tracked: true).distinct }
   scope :not_tracked, -> { where(tracked: false).distinct }
 
-  serialize :user_ids, type: Array, , default: [] # This text field only stores user ids as text
+  serialize :user_ids, type: Array, default: [] # This text field only stores user ids as text
 
   ####################
   # Instance methods #
--- a/app/models/surveys/survey_assignment.rb
+++ b/app/models/surveys/survey_assignment.rb
@@ -42,7 +42,7 @@ class SurveyAssignment < ApplicationRecord
   ###########################
   # Custom email attributes #
   ###########################
-  serialize :custom_email, type: Hash, , default: {}
+  serialize :custom_email, type: Hash, default: {}
 
   def custom_email_subject
     custom_email[:subject]
--- a/app/models/user_data/training_modules_users.rb
+++ b/app/models/user_data/training_modules_users.rb
@@ -18,7 +18,7 @@ class TrainingModulesUsers < ApplicationRecord
   belongs_to :user
   belongs_to :training_module
 
-  serialize :flags, type: Hash, , default: {}
+  serialize :flags, type: Hash, default: {}
 
   def furthest_slide?(slide_slug)
     return true if last_slide_completed.nil?
--- a/app/models/user_data/user_profile.rb
+++ b/app/models/user_data/user_profile.rb
@@ -20,7 +20,7 @@ class UserProfile < ApplicationRecord
   belongs_to :user
   has_attached_file :image, styles: { thumb: '150x150>' }
   validates_attachment_content_type :image, content_type: %r{\Aimage/.*\z}
-  serialize :email_preferences, type: Hash, , default: {}
+  serialize :email_preferences, type: Hash, default: {}
 
   def email_preferences_token
     set_email_preferences_token unless email_preferences.key?(:token)
--- a/app/models/wiki_content/category.rb
+++ b/app/models/wiki_content/category.rb
@@ -26,7 +26,7 @@ class Category < ApplicationRecord
   has_many :categories_courses, class_name: 'CategoriesCourses', dependent: :destroy
   has_many :courses, through: :categories_courses
 
-  serialize :article_titles, type: Array, , default: []
+  serialize :article_titles, type: Array, default: []
 
   validates :name, presence: true, length: { minimum: 1 }
   validates :name, numericality: { only_integer: true }, on: :create,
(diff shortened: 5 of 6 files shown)


ragesoss:
It's not necessary to merge from the main branch for a PR like this. I'd prefer you only include your own commits, and leave merging or rebasing to me during the review and merge process. (Please force-push this to only include the refactor commit.)

ragesoss:
Interesting. The failing specs seem to be caused by changed serialization behavior; the models are returning `nil` instead of an empty hash, when no value for that serialized attribute is in the database. Is that expected? Does that change with Rails 7.1, or is this strictly related to Rubocop? If it's just related to Rubocop, we should disable the rule.

SanjanaSogimatt:
> Interesting. The failing specs seem to be caused by changed serialization behavior; the models are returning `nil` instead of an empty hash, when no value for that serialized attribute is in the database. Is that expected? Does that change with Rails 7.1, or is this strictly related to Rubocop? If it's just related to Rubocop, we should disable the rule.

In Rails 7.1, `serialize :flags, type: Hash` returns nil when the database column is NULL. Previously, serialized attributes implicitly returned {}, which masked missing data. The failing specs relied on this implicit behavior, which no longer exists.

ragesoss:
Okay. So we need to find a clean way to re-implement the pre-7.1 behavior?

ragesoss:
Can you point to the relevant documentation or Rails changelog?

SanjanaSogimatt:
> In Rails 7.1, `serialize :flags, type: Hash` returns nil when the database column is NULL. Previously, serialized attributes implicitly returned {}, which masked missing data. The failing specs relied on this implicit behavior, which no longer exists.

https://api.rubyonrails.org/v7.1/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html

ragesoss:
I guess here? https://github.com/rails/rails/blob/7-1-stable/activerecord/CHANGELOG.md

Maybe this is relevant?

> The original default can be restored by setting `config.active_record.default_column_serializer = YAML`.

ragesoss:
Ah, so an alternative would be to add the appropriate `default` option for each instance? That might be better than overriding the new default serializer. I see that the default option didn't change from 7.0 to 7.1, but I guess that YAML already turned `nil` into an empty Hash or Array before needing to fall back to a default

SanjanaSogimatt:
> Ah, so an alternative would be to add the appropriate `default` option for each instance? That might be better than overriding the new default serializer. I see that the default option didn't change from 7.0 to 7.1, but I guess that YAML already turned `nil` into an empty Hash or Array before needing to fall back to a default

Another option is `   config.active_record.default_column_serializer = YAML` set global default ? Would that cause any problem?

ragesoss:
From the release notes, I see that they wanted to stop using YAML by default because it's less safe for serializing arbitrary content, so I think adding the `default` option to each case at the same time as adding the explicit `type:` parameter would be appropriate.

SanjanaSogimatt:
```
 def set_available_article_flag
    @flags = @user_id.nil? ? { available_article: true } : nil
  end
```
this from `lib/assignment_manager.rb` is setting the flags to nil

app/models/alert.rb
```diff
@@ -28,7 +28,7 @@ class Alert < ApplicationRecord
 
   include ArticleHelper
 
-  serialize :details, type: Hash
+  serialize :details, type: Hash, coder: YAML, default: {}
```
ragesoss: Was there a problem with leaving out the `coder` value? I thought setting `default` would probably keep the behavior we want, while still still switching to the new behavior we want. If there is no new default coder, we should use the `config` method from the changelog rather than explicitly setting the coder for each case.

PR #6625, opened 2026-01-08
[WIP] Add backend system to automatic backup with sleeping approach

gabina:
## What this PR does
This is a work in progress.

This PR adds a system to perform automatic database backups periodically. It is intended to work in the following way:
1. **On DB server.** Weekly, a cron job runs on the db server, in charge of performing the backup itself. The first task it performs is creating a new backup record with status set to `waiting`. This works as a way to let the backend app knowing that a backup is waiting to run.
2. **On app servers.** At this point, it's expected that some Sidekiq jobs are running and many others are queued. The  running jobs will continue until completion, but once they finish, new `CourseDataUpdateWorker` jobs will start and will be paused due to the `BackupPause` module that sleeps until no backup is waiting or running. At this stage, the app knows a backup is waiting to run. Other workers (for example `ConstantUpdateWorker`) will work as usual.
3.  **On DB server.** It checks the new `can_start_backup.json` endpoint to determine whether it's safe to start the backup.  The criterion is that a backup can only run if all currently running jobs started after the backup was scheduled. This ensures that, if any of those jobs are `CourseDataUpdateWorker` jobs, they are waiting for the backup to finish. 
If the response is 503 server unavailable, it waits for a certain amount of time (to be defined) and retries until it receives a 200 OK response.
4. **On DB server.** It updates the data table record status to `running`.
5.  **On DB server.** It runs the backup itself.
6.  **On DB server.** It updates the data table record status to `finished`.
7. **On app servers.** Sidekiq running jobs stop sleeping. The application returns to normal operation.

**Important notes**
-  `pause_until_no_backup` is called inside `UpdateCourseStats` instead of within `CourseDataUpdateWorker` to ensure it also runs when using the `/manual_update` route. Using a middleware instead of a module would mean that the `pause_until_no_backup` logic would not run for manual updates. This is likely not a big deal, since only admins can trigger them.
- Queue latency is not missed with this approach.
- Processes that run course updates (`peony-sidekiq`, `peony-sidekiq-3`, `peony-sidekiq-medium`) only runs course updates, so blocking the threads by sleeping is not a real problem, since there are no other jobs we want to run on those servers.
- We don't need to rely on the sidekiq reaper to clean orphans locks from sidekiq-uniq-jobs gem.

## AI usage
Used chatGPT to improve the PR description.

### DIFF
--- a/app/models/system/backup.rb
+++ b/app/models/system/backup.rb
@@ -16,6 +16,8 @@ class Backup < ApplicationRecord
   IN_PROCESS = %w[waiting running].freeze
 
   def self.current_backup
-    Backup.find_by(status: IN_PROCESS)
+    ActiveRecord::Base.uncached do
+      Backup.find_by(status: IN_PROCESS)
+    end
   end
 end
--- a/app/services/update_course_stats.rb
+++ b/app/services/update_course_stats.rb
@@ -11,6 +11,7 @@ require_dependency "#{Rails.root}/lib/importers/average_views_importer"
 require_dependency "#{Rails.root}/lib/errors/update_service_error_helper"
 require_dependency "#{Rails.root}/lib/data_cycle/course_queue_sorting"
 require_dependency "#{Rails.root}/lib/data_cycle/update_debugger"
+require_dependency "#{Rails.root}/lib/data_cycle/backup_pause"
 require_dependency "#{Rails.root}/app/services/update_course_wiki_timeslices"
 
 #= Updates stats for a specific course. This includes importing data (revisions, uploads,
@@ -19,8 +20,10 @@ require_dependency "#{Rails.root}/app/services/update_course_wiki_timeslices"
 class UpdateCourseStats
   include UpdateServiceErrorHelper
   include CourseQueueSorting
+  include BackupPause
 
-  def initialize(course)
+  def initialize(course) # rubocop:disable Metrics/MethodLength
+    pause_until_no_backup
     @course = course
     # If the upate was explicitly requested by a user,
     # it could be because the dates or other paramters were just changed.
--- a/lib/data_cycle/backup_pause.rb
+++ b/lib/data_cycle/backup_pause.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+module BackupPause
+  def pause_until_no_backup
+    sleep 60 until Backup.current_backup.nil?
+  end
+end


app/controllers/system/backups_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+
+      # There should always be a backup record before starting the backup.
+      return render_not_ready if current_backup.nil?
+
+      # The WorkSet stores the work being done by this Sidekiq cluster.
+      # This is live data that can change every millisecond.
+      # It's only safe to run a backup if all running jobs started
+      # after the backup was scheduled.
+      ready = Sidekiq::WorkSet.new.all? do |_process_id, _thread_id, work|
+        Time.zone.at(work['run_at']).to_datetime > current_backup.created_at + 1.minute
```
gabina, line 19: Didn't find detailed docs about `run_at` field. 

```
  #      # work is a `Sidekiq::Work` instance that has the following accessor methods.
  #      # [work.queue, work.run_at, work.payload]
```

However, I ran some tests locally and all makes me think that the `run_at` timestamp filed is the exact moment when the worker started to run.

app/models/system/backup.rb
```diff
@@ -0,0 +1,23 @@
...
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#
+class Backup < ApplicationRecord
+  IN_PROCESS = %w[waiting running].freeze
+
+  def self.current_backup
+    ActiveRecord::Base.uncached do
+      Backup.find_by(status: IN_PROCESS)
+    end
```
gabina, line 21: Added the `uncached` because otherwise it keeps sleeping more than necessary since it doesn't detect the back finished.

app/services/update_course_stats.rb
```diff
@@ -19,8 +20,10 @@
 class UpdateCourseStats
   include UpdateServiceErrorHelper
   include CourseQueueSorting
+  include BackupPause
 
-  def initialize(course)
+  def initialize(course) # rubocop:disable Metrics/MethodLength
```
gabina, line 25: I can find a better way for making this `initialize` method shorter if we want to go with this approach.

app/controllers/system/backups_controller.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+require 'sidekiq/api'
+
+module System
+  class BackupsController < ApplicationController
+    respond_to :json
+
+    def can_start_backup
```
gabina, line 8: This is the controller for the new `/system/can_start_backup.json` endpoint. The idea is for the backup script to make a request to it. While the response does not appear to expose confidential information, we may want to add some security layer here, although I am not yet sure which would be most appropriate.

gabina:
@ragesoss WDYT? I think this could be a good first version to deploy. It should be easy to test: we can manually create a backup record with "waiting" status and to see how new course updates (likely from the short queue) keep waiting until we change the record to "error" status.

As regards the https://github.com/kenaniah/sidekiq-status, it looks compatible, but I think it makes sense to have a first working version and then move to that approach, that would also be useful beyond the backup system.

app/controllers/system/backups_controller.rb
gabina: I forgot to remove the `+ 1.minute`. Will fix the commit.

gabina:
Closing this in favour of #6630 

## Some notes on why we didn't go with this approach:

For this approach, we consider that the only jobs that we want to pause for doing the backup are the course updates. Other jobs can operate normally.
We have to take a decision in two different places.
On one hand, a course update background job needs to decide if it can start the update normally or if it has to sleep. This can be decided based on the backup table records. If there is any row with "waiting" or "sleeping" status, then it has to sleep.
On the other hand, the db server needs to know if it's safe to start a backup in a given moment. When is it safe for the db server to start a backup? When all course updates running processes (if any) know there is a backup waiting, that means, they're sleeping. The way I found to determine all jobs are sleeping was by comparing the run_at field for every running sidekiq process and the scheduled_at field for the backup record. If the process started to run after the backup was scheduled, then it knows a backup has to run, so it's sleeping.

-------------------------------------------------------------------------
These two decisions have to be based on exactly the same criterion. Why?
Let's suppose that the db server thinks it's not safe to start a backup, but all the course updates running processes are already sleeping.
- The processes are already sleeping, so they will keep sleeping until the backup finishes.
- The backup won't start until the processes finishes (it doesn't know they're already sleeping).
This is a deadlock.
Let's suppose that the db server thinks it's safe to start a backup, but some course update is running normally. This is an undesirable situation, but there is no deadlock.
-------------------------------------------------------------------------
Original design:
1- On DB server. It creates a new backup record with "waiting" status and scheduled_at field set to "now" (when the record was created).
2- On sidekiq. Processes that has "run_at" field before the scheduled_at field will see the waiting record and will go to sleep. Other processes will run until completion.
3- On DB server. It asks if it's safe to run the backup. If all running processes started before scheduled_at field (they have run_at field before scheduled_at), then it's safe to run the backup. It changes the record status to "running"
4- On DB server. It runs the backup.
5- On DB server. It changes the record to "finished"
6- On sidekiq. Processes wake up.
What's the problem?
Sidekiq background jobs are not able to see themselves. They don't have access to the run_at field.
If, instead of checking the run_at field in step 2, the job only checks the existence of a backup record with "waiting"/"running" status, it could be a deadlock:
- The sidekiq job starts (internally it sets run_at).
- The backup record is created.
- The sidekiq job goes to sleep.
- The DB server never thinks it's safe to run the backup, because run_at is previous to scheduled_at.
-------------------------------------------------------------------------
Other approach:
Use the sidekiq-status gem as Sage suggested.
In a first basic approach, we could use a 'stage' custom field that indicates that the job is sleeping or is normally running (already passed through the decision about sleeping or not). That way, to decide if it's safe to run a backup we look at the stage custom fields. If any update course process already passed through the sleeping phase, then it's not safe (it's already updating the course). It should wait and ask again later.

PR #6626, opened 2026-01-08
Replace "<languages />" with nothing when parsing on-wiki slides

### LINKED ISSUE #6616: Replace "<languages />" with nothing when parsing on-wiki slides
People want translated pages on Meta-Wiki to have a languages tag. Right now you can see that literally at https://outreachdashboard.wmflabs.org/training/editathons/planning-leading-up-to-the-event/after-the-event

Rather than creating another awkward exception on Meta (the weird custom wikitext parser causes enough problems already), the dashboard could just be told to ignore that string ...

Formasitchijoh:
## What this PR does
This PR resolves #6616 by ensuring stray `<languages />` tags (and variants like `<languages/>` ) are removed during on-wiki slide parsing.

The bug occurred when a bare `<languages />` line survived the existing `<noinclude>` cleanup and became the extracted slide title. Adding a dedicated `remove_languages_tag` method explicitly strips these variants, preventing them from polluting the title while leaving legitimate translation markup on Meta-Wiki intact.The images shows the test for the example buggy wiki-text when the method is introduced and when it is commented out.

## AI usage
Used AI to explore and debug various scenarios where the `<languages />` tag could leak through the parser.

## Screenshots
Screenshot 2026-01-08 at 13 22 38

After:

Screenshot 2026-01-08 at 13 19 02

I will open a different PR wrt the duplicate heading as they are unrelated to this PR

### DIFF
--- a/spec/lib/training/wiki_slide_parser_spec.rb
+++ b/spec/lib/training/wiki_slide_parser_spec.rb
@@ -248,30 +248,32 @@ describe WikiSlideParser do
     WIKISLIDE
   end
 
-  let(:buggy_wikitext) do
+  # https://meta.wikimedia.org/w/index.php?title=Training_modules/dashboard/slides/12312-after-the-event&action=edit&oldid=29790789
+  let(:unbound_languages_tag) do
     <<~WIKISLIDE
-      <languages />
-      == After the event! ==
+      <languages/>
+      <translate>
+      ==After the event!== <!--T:1-->
 
+      <!--T:4-->
       Success! You made it through your editing event! There are a lot of opportunities for overlooking things or making mistakes, but you learned from them, and all in all it was a successful event! Now for a bit more communications and tracking!
 
+      <!--T:2-->
       Beyond running the event, there are a few steps that you may want to use to ensure that everyone (you, your partners and the participants), get the most from the editathon:
-
-      * Make sure to integrate any user names into the Programs and Events dashboard, so to generate metrics on the event. Note: if participants did not sign up with their user names on a public sign up page by either signing into the Programs and Events Dashboard and registering themselves, or listing their usernames on a public event page, it's important to get consent to publicly associate the editor's User ID with the event in order to protect privacy.
-
+      * Make sure to integrate any user names into the Programs and Events dashboard, so to generate metrics on the event. ''Note: if participants did not sign up with their user names on a public sign up page by either signing into the Programs and Events Dashboard and registering themselves, or listing their usernames on a public event page, it's important to get consent to publicly associate the editor's User ID with the event in order to protect privacy.''
       * Report outcomes and metrics to your participants and partners who supported the event. It may be useful to evaluate these outcomes in light of the goals you set at the beginning of the program. Reporting outcomes can help remind folks of the impact of the event, and that you still are interested in supporting their participation in the Wikimedia Community.
-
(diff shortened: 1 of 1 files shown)


ragesoss:
For the example from the original issue, why did the `languages` tag survive the `noinclude` cleanup? I couldn't spot what was buggy about that markup.

Formasitchijoh:
I think the reason the `<languages /> ` tag survived the `remove_noinclude ` cleanup in the original example is that, on the affected slide(s) at the time, it appeared as a bare/standalone line (e.g., just <languages /> at the top), without being wrapped in `<noinclude>...</noinclude>.`

The` remove_noinclude` method correctly removes full paired `<noinclude>…</noinclude>`  blocks (including the standard `<noinclude><languages /></noinclude>`  pattern seen in tests), but it doesn't touch content outside those tags. A stray bare tag would remain as the first non-blank line after cleaning, causing `WikiSlideParser#title`  to extract it literally and display `<languages/> `as the slide title.

Adding `remove_languages_tag`  provides a small, targeted defensive layer to strip any such leftovers (or variants like `<languages/> `or `<languages></languages>)`, without impacting properly wrapped cases or Meta's translation features.

ragesoss:
Okay. Can you add a link in the description to the specific version of the history of that example page that you think caused it? (It sounds like updating that module now should fix it in production, since the current version doesn't have the incompatible syntax, correct?)

Formasitchijoh:
The current raw Meta source looks clean 

The stray <languages /> (or variant) is possibly leaking through during parsing, possibly from subtle remnants in this slide or related module pages that survive` remove_noinclude.`

Re-exporting/updating the module via the usual script should indeed flush caches and fix it fully in production.

The `remove_languages_tag ` addition is just a lightweight defensive measure to handle any such strays robustly going forward, without depending only on perfect source cleanup.

Formasitchijoh:
> Okay. Can you add a link in the description to the specific version of the history of that example page that you think caused it? (It sounds like updating that module now should fix it in production, since the current version doesn't have the incompatible syntax, correct?)

@ragesoss Is it possible to update that module on the training slides in production. Because like you suggested the metawiki  syntax looks clean . Apologies for the late response

Screenshot 2026-01-09 at 16 36 30

ragesoss:
Done!

I'd still like to see which version in the page history had the bad syntax, as I don't want to merge this unless we know exactly what situation led to the problem in production.

Formasitchijoh:
> Done!
> 
> I'd still like to see which version in the page history had the bad syntax, as I don't want to merge this unless we know exactly what situation led to the problem in production.

Sure I will look at the revision history again to check

Formasitchijoh:
@ragesoss here is the buggy [revision](https://meta.wikimedia.org/w/index.php?title=Training_modules/dashboard/slides/12312-after-the-event&action=edit&oldid=29790789) that introduced the bug, Like the assumption I had , the `<languages />` tag was not found in a `<noinclude>…</noinclude> `  block so it was not cleaned up by the `remove_noinclude`  method 

Screenshot 2026-01-12 at 10 48 20

ragesoss:
@Formasitchijoh excellent. You should use that as the example in the spec (and maybe don't call it "buggy" since this will no longer cause any bugs after the change), and you should include a link to that revision in a comment.

Formasitchijoh:
> @Formasitchijoh excellent. You should use that as the example in the spec (and maybe don't call it "buggy" since this will no longer cause any bugs after the change), and you should include a link to that revision in a comment.

Alright I will do that

Issue #6579, opened 2025-12-08
Develop a robust data backup system for Programs & Events Dashboard


ragesoss:
Programs & Events Dashboard recently experienced a major database crash: https://phabricator.wikimedia.org/T411890

In this case, the database VM's filesystem broke, and also the attached storage volume that holds the database files had some corruption.

Since the rollout of the timeslice system, we're no longer worried about running out of storage on the Programs & Events Dashboard server, and after cleaning up the duplicated files used during the recovery process, we'd have plenty of extra storage such that we could hold a series of database dumps.

I think ideally, we'd want to do dumps while there are no course updates running, so I think we'd to build a system around something like:
* On a weekly basis, pause the sidekiq course update and other non-default workers, allow them to finish their current jobs, then do a database dump, and turn them back on.
* Automatically curate a set of database dumps (eg, keep one that is about 6 months old, one that is about a month old, and one that is at most a week old) to keep, and delete old ones as new ones get created.

gabina:
If this is a priority at any point, I'm interested in working on this

ragesoss:
@gabina feel free to work on this next.

PR #6630, opened 2026-01-13
Implement backup system using sidekiq-status gem

gabina:
## What this PR does
This PR adds a system to perform automatic database backups periodically using the [sidekiq-status gem](https://github.com/kenaniah/sidekiq-status).

See issue #6579.

## Intended flow
<table border="0">
 <tr>
    <td><b style="font-size:30px">DB Server</b></td>
    <td><b style="font-size:30px">App servers</b></td>
 </tr>
 <tr>
    <td> Weekly, a cron job runs on the db server, in charge of performing the backup itself. The first task it performs is creating a new backup record with status set to <b>waiting</b>. This works as a way to let the backend app knowing that a backup is waiting to run.
    </td>
    <td></td>
 </tr>
 <tr>
    <td></td>
    <td>From this point, all CourseDataUpdateWorker jobs that start the UpdateCourseStats will be paused sleeping until the backup record leaves the waiting/running status. Already running CourseDataUpdateWorker jobs will continue until completion.</td>
 </tr>
 <tr>
    <td> Some minutes later it checks the new `can_start_backup.json` endpoint to determine whether it's safe to start the backup.  The criterion is that a backup can only run if all currently running CourseDataUpdateWorker jobs are in  <b>sleeping </b> phase. 
If the response is 503 server unavailable, it waits for a certain amount of time (to be defined) and retries until it receives a 200 OK response. Once it gets the 200 OK response, it sets the backup status to <b>running</b>.
    </td>
    <td></td>
 </tr>
 <tr>
    <td>It runs the backup itself.</td>
    <td></td>
 </tr>
 <tr>
    <td>It updates the data table record status to <b>finished </b>.</td>
    <td></td>
 </tr>
 <tr>
    <td></td>
    <td>CourseDataUpdateWorker jobs leaves the sleeping phase and <b>woke up</b>. The application returns to normal operation.</td>
 </tr>
</table>

## Implementation details
The `phase` status for a sidekiq `CourseDataUpdateWorker` is stored using the sidekiq-status gem:

1. The first thing the `perform` method does is to store jid, worker, arguments and phase. The first phase is **initialization**. Saving jid, worker and arguments is not strictly necessary, it is supposed to do so automatically, but I noticed that sometimes they were not visible from the web interface (although they were when calling `Sidekiq::Status.get_all(job_id)`), so I forced it. That's something I still have to fully understand (I'm not sure how it'll behave in production).
2. Once the `UpdateCourseStats` is initialized, the first thing it does is to evaluate if it has to pause the job. If so, the phase is updated to **sleeping** and it goes to sleep. Periodically, it wakes up and evaluate if it's safe to continue the update or if it has to continue sleeping.
3. Once the backup finishes, the phase updates to **woke up** and the course update runs until completion.

**Notes**
-  `pause_until_no_backup` is called inside `UpdateCourseStats` instead of within `CourseDataUpdateWorker` to ensure it also runs when using the `/manual_update` route. Using a middleware instead of a module would mean that the `pause_until_no_backup` logic would not run for manual updates. This is likely not a big deal, since only admins can trigger them.
- Queue latency is not missed with this approach.
- Processes that run course updates (`peony-sidekiq`, `peony-sidekiq-3`, `peony-sidekiq-medium`) only runs course updates, so blocking the threads by sleeping is not a real problem, since there are no other jobs we want to run on those servers.
- We don't need to rely on the sidekiq reaper to clean orphans locks from sidekiq-uniq-jobs gem.

## AI usage
I used ChatGPT for:
- Improving PR description.
- Small styling updates to my hand-written code (mainly to make it more idiomatic).
- Discussing the problem overall, although the final solution is much more the result of back-and-forth between Sage and me (and between me and me-the-next-day, since I changed my mind several times :D), as well as reading documentation and blogs, than of ChatGPT itself. Example: even though I shared the detailed flow from the previous PR with ChatGPT, it didn't point out the deadlock at any point.

## Screenshots
Before: no status gem

After:
Example of a job currently sleeping:
image

Example of a job currently running the update:
image

The use of sidekiq-status gem may be extended in future work to track the number of timeslices that a job processed, etc, etc, etc.

### DIFF
--- a/spec/controllers/system/backups_controller_spec.rb
+++ b/spec/controllers/system/backups_controller_spec.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe System::BackupsController, type: :request do
+  let(:work) do
+    ['server-15ALC6:390735:407ff4a2b287',
+     '81ib',
+     { 'queue'=>'medium_update',
+     'payload'=>
+       { 'class'=>'CourseDataUpdateWorker',
+       'jid'=>'e118d576de83c8e44526bba7' },
+     'run_at'=>1769012340 }]
+  end
+
+  describe '#can_start_backup' do
+    context 'when no current backup' do
+      it 'returns 503 service unavailable' do
+        get '/system/can_start_backup.json'
+        expect(response.status).to eq(503)
+      end
+    end
+
+    context 'when backup is waiting' do
+      before { create(:backup, status: 'waiting') }
+
+      context 'when no jobs are running' do
+        it 'returns 200 OK' do
+          get '/system/can_start_backup.json'
+          expect(response.status).to eq(200)
+        end
+      end
+
+      context 'when all jobs are sleeping' do
+        before do
+          workset = instance_double(Sidekiq::WorkSet)
+          allow(Sidekiq::WorkSet).to receive(:new).and_return(workset)
+          allow(workset).to receive(:all?).and_return(true)
+        end
+
+        it 'returns 200 OK' do
+          get '/system/can_start_backup.json'
+          expect(response.status).to eq(200)
+        end
+      end
+
+      context 'when any job woke up' do
+        before do
+          workset = instance_double(Sidekiq::WorkSet)
+          allow(Sidekiq::WorkSet).to receive(:new).and_return(workset)
+          allow(workset).to receive(:all?).and_return(false)
+        end
+
+        it 'returns 503 service unavailable' do
+          get '/system/can_start_backup.json'
+          expect(response.status).to eq(503)
+        end
+      end
+    end
+  end
+end
--- a/app/services/log_sidekiq_status.rb
+++ b/app/services/log_sidekiq_status.rb
@@ -14,6 +14,8 @@ class LogSidekiqStatus
     @store.call phase: 'woke_up'
   end
 
+  private
+
   def pause
     @sleep_count += 1
     @store.call phase: "sleeping_#{@sleep_count}"
--- a/spec/services/log_sidekiq_status_spec.rb
+++ b/spec/services/log_sidekiq_status_spec.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+describe LogSidekiqStatus do
+  describe '#pause_until_no_backup' do
+    subject(:logger) { described_class.new(store) }
+
+    let(:store) { instance_double(Proc) }
+    let(:backup) { create(:backup, status: 'running') }
+
+    before do
+      stub_const('LogSidekiqStatus::SLEEP_TIME_IN_SECONDS', 0)
+      allow(store).to receive(:call)
+    end
+
+    describe '#pause_until_no_backup' do
+      context 'when a backup is running and then finishes' do
+        before do
+          allow(Backup).to receive(:current_backup)
+            .and_return(backup, nil)
+        end
+
+        it 'sleeps once and then wakes up' do
+          logger.pause_until_no_backup
+
+          expect(store).to have_received(:call).with(phase: 'sleeping_1')
+          expect(store).to have_received(:call).with(phase: 'woke_up')
+        end
+      end
+
+      context 'when a backup is running and multiple sleeps are required' do
+        before do
+          allow(Backup).to receive(:current_backup)
+            .and_return(backup, backup, nil)
+        end
+
+        it 'sleeps twice and then wakes up' do
+          logger.pause_until_no_backup
+
+          expect(store).to have_received(:call).with(phase: 'sleeping_1')
+          expect(store).to have_received(:call).with(phase: 'sleeping_2')
+          expect(store).to have_received(:call).with(phase: 'woke_up')
+        end
+      end
+
+      context 'when no backup is running/waiting' do
+        before do
+          allow(Backup).to receive(:current_backup).and_return(nil)
+        end
+
+        it 'does not sleep and immediately wakes up' do
+          logger.pause_until_no_backup
+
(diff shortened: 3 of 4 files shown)


config/initializers/sidekiq_status.rb
```diff
@@ -0,0 +1,10 @@
+expiration =  3.hours.to_i
```
gabina, line 1: According to docs:

_The expiration time set will be used as the [Redis expire time](https://redis.io/commands/expire), which is also known as the TTL (time to live). Once the expiration time has passed, all information about the job's status and any custom data stored via sidekiq-status will disappear.
It is advised that you set the expiration time greater than the amount of time required to complete the job._

It’s not clear from the description whether the timeout is refreshed when the status is updated or not. In particular, when it says “once the expiration time has passed,” it’s unclear from which point that is measured: from the creation of the sidekiq-status record or from its last update.

However, based on my local experiments, sidekiq status for "working" jobs does not appear to be cleaned up at any point. This differs from "completed" or "queued" jobs, where I can see the sidekiq-status data being deleted after the expiration time measured from creation.

The risk here is that, if sidekiq status were ever deleted for a "working" job, we would end up with an empty status and no way to determine whether the job is already sleeping.

Since I’m not completely sure how this behaves in all cases, a reasonable approach for now is to log a Sentry message when this happens (a CourseDataUpdateWorker with no sidekiq-status) and block the backup, assuming manual intervention would be required.

app/models/system/backup.rb
```diff
@@ -0,0 +1,23 @@
...
+#  created_at   :datetime         not null
+#  updated_at   :datetime         not null
+#
+class Backup < ApplicationRecord
+  IN_PROCESS = %w[waiting running].freeze
+
+  def self.current_backup
+    ActiveRecord::Base.uncached do
+      Backup.find_by(status: IN_PROCESS)
+    end
```
gabina, line 21: Added the `uncached` because otherwise it keeps sleeping more than necessary since it doesn't detect the backup finished.

app/services/update_course_stats.rb
```diff
@@ -20,7 +20,8 @@ class UpdateCourseStats
   include UpdateServiceErrorHelper
   include CourseQueueSorting
 
-  def initialize(course)
+  def initialize(course, sidekiq_status_logger: nil) # rubocop:disable Metrics/MethodLength
```
gabina, line 23: I can find a better way for making this initialize method shorter if we want to go with this approach.

app/controllers/system/backups_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+require 'sidekiq/api'
+
+module System
+  class BackupsController < ApplicationController
+    respond_to :json
+
+    COURSE_UPDATE_WORKER_CLASS = CourseDataUpdateWorker.name
+
+    def can_start_backup
```
gabina, line 10: This is the controller for the new `/system/can_start_backup.json` endpoint. The idea is for the backup script to make a request to it. While the response does not appear to expose confidential information, we may want to add some security layer here, although I am not yet sure which would be most appropriate.

gabina:
Well.. I think I'm finally happy with this. It's ready for a first review and feedback @ragesoss

app/services/update_course_stats.rb
ragesoss: I don't mind this for now. Refactoring this file at some point would be nice, but it's currently well-organized and easy to understand from my perspective.

config/initializers/sidekiq_status.rb
ragesoss: We know that some course update jobs can run for days, so we may need a strategy for dealing with those long-running jobs without waiting for completion. If TLL is reset at every status update — which I suspect is the case — that's probably not relevant here.

The other situation that I think we see in production, which may be more relevant for this status expiration, is when a sidekiq process dies or becomes unresponsive (eg, because it ran its server out of memory). In that case, the current behavior is that the process stops showing up on the Sidekiq UI, but I'm not sure what happens on on the Redis side of things in that situation. If this TTL controls how long Redis still reports a job as running even though the process stopped responding, we may want a shorter expiration to let the system continue as well as possible, or we may want a longer expiration so that it's easier to understand which process got stuck and when.

I think this is a fine place to start, since we don't know which situations it may be relevant for.

spec/controllers/system/backups_controller_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+          get '/system/can_start_backup.json'
+          expect(response.status).to eq(200)
+        end
+      end
+
+      context 'when all jobs are sleeping' do
+        before do
+          workset = instance_double(Sidekiq::WorkSet)
+          allow(Sidekiq::WorkSet).to receive(:new).and_return(workset)
+          allow(workset).to receive(:all?).and_return(true)
```
gabina, line 38: It would be good to avoid using `all?` in order to check the logic that determines whether jobs are sleeping or not, but I haven't found an easy way to do that.

gabina:
I think we can deploy this version. Everything should work exactly the same, except for the new sidekiq status user interface. Also, if we manually create a backup record in `waiting` status, we'll be able to see how jobs start sleeping in production.

ragesoss:
Great. I'm going to wait until next week to deploy it, as I'm traveling Thursday and Friday.

app/models/system/backup.rb
ragesoss: This is worth noting in code comment.

ragesoss:
This is deployed to Wiki Education Dashboard.

gabina:
Can you see something different in the `sidekiq` web interface?

ragesoss:
Yes, it has the Status tab now.

Issue #6555, opened 2025-11-06
Training slides based on on-wiki pages have stray curly braces


ragesoss:
Here are some examples with the stray `}}`:

* https://outreachdashboard.wmflabs.org/training/wikidata/introduction-to-wikidata/and-by-machines (in the image caption)
* https://outreachdashboard.wmflabs.org/training/wikidata/introduction-to-wikidata/wikidata-values-quiz (in the explanation for the final answer, which shows when selected)

I suspect this is being caused by how the dashboard handles `translate` tags from the wiki source pages. It should be fixable by making changes to WikiSlideParser. The above pages can be used to write tests. (Test-driven development would be a good strategy here.)

GitaHub71:
I'll try this one

GitaHub71:
@ragesoss I couldn't replicate this issue locally. How can I test this in my local environment?

ragesoss:
You have to switch to `wiki_education: false` mode, and then load the relevant training module content. See the 'Loading training modules' section of `docs/user_roles.md`.

GitaHub71:
@ragesoss I'm sorry I didn't write it clear enough. I already loaded the training module, even the specific module in this issue. But in my local development environment, it seemed normal, with wiki_edu or non wiki_edu mode.

<img width="1685" height="1001" alt="Image" src="https://github.com/user-attachments/assets/2d26c285-dbd4-4462-a982-177cb7d9bf6c" />

ragesoss:
That screenshot shows a training loaded from the .yml files, not from meta-wiki. You must load the specific training module you want to test, while in non-wiki-ed mode.

NIGHTFURY609:
Hey @ragesoss can I try this issue since the other person has not responded in a long time. can you assign me this issue, I was able to reproduce it working on it right now

ragesoss:
@NIGHTFURY609 you're welcome to work on this. We don't normally assign issues.

NIGHTFURY609:
HI @ragesoss , 
I was having problems with testing the fixes of this bug, even though tests are passing the corrected version is not being displayed in the page. Is this because the rendered content is coming directly from Wikimedia’s cached parsed output, not from a fresh parse using my updated code or is this an issue from my bug fix?

ragesoss:
[...]

PR #6634, opened 2026-01-15
Fix training slides based on on-wiki pages having stray curly braces

Formasitchijoh:
## What this PR does
This PR resolves the issue #6555  by improving the lookahead regex that is used for parsing and cleaning the on-wiki pages wikitext

### Cause of the bug
The extra curly braces were slipping through because the exiting regex lookahead `(?=\||\Z)` does not account for the template's closing brackets `}}.` In the input wikitext, for any final parameter the regex searches for the caption  or any final parameter and then uses a positive lookahead to stop when it hits either another pipe `(|)` or the end of the entire string `(\Z).`

**The Oversight:** Because there is no pipe following the caption, the regex continues capturing characters until it reaches the end of the string. The closing `}}` of the template is not a pipe and it is not the end of the string (since more text follows the template).hence the regex consumes the `}} ` as part of the caption's value because it's looking for a pipe that never comes before the template ends.

## AI usage
Used AI to understand the existing regex and how to improve on it

## Screenshots
Screenshot 2026-01-15 at 23 53 55

## Notes
I used a test driven development approach as suggested, I used the wikitext for the slide that have stray curly braces both for the normal text slide and question groups.

### DIFF
--- a/spec/lib/training/wiki_slide_parser_spec.rb
+++ b/spec/lib/training/wiki_slide_parser_spec.rb
@@ -278,6 +278,7 @@ describe WikiSlideParser do
   end
 
   # https://outreachdashboard.wmflabs.org/training/wikidata/introduction-to-wikidata/and-by-machines
+  # https://meta.wikimedia.org/w/index.php?title=Training_modules/dashboard/slides/12406-and-by-machines&oldid=29578464
   let(:onwiki_page_with_unmatched_curly_braces) do
     <<~WIKISLIDE
       <noinclude><languages/></noinclude>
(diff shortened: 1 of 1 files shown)


Formasitchijoh:
@ragesoss please could you review this

spec/lib/training/wiki_slide_parser_spec.rb
```diff
@@ -277,6 +277,51 @@
     WIKISLIDE
   end
 
+  # https://outreachdashboard.wmflabs.org/training/wikidata/introduction-to-wikidata/and-by-machines
```
ragesoss, line 280: In the comments, you should also include a link to the specific revision of the wiki source page where this wikitext came from.

ragesoss:
Nicely done! This will be ready to merge once you add the wikitext links.

spec/lib/training/wiki_slide_parser_spec.rb
Formasitchijoh: Okay I will add that

PR #6636, opened 2026-01-16
Update application configuration and test environment settings

SanjanaSogimatt:
# PR 1: Rails 7.1 Configuration Updates

## What this PR does
This PR updates the application configuration to be compatible with Rails 7.1. It removes deprecated configuration options and updates test environment settings to match Rails 7.1 requirements.

**Changes:**
- Removes `config.active_record.legacy_connection_handling` from `config/application.rb` (this option was removed in Rails 7.1) 
- Updates `config.action_dispatch.show_exceptions` in test environment from `false` to `:none` (Rails 7.1 syntax change)
- Moves `ActiveSupport::Testing::TimeHelpers` include from `spec/spec_helper.rb` to `rails_helper.rb` (Rails 7.1 best practice - helpers should be included after Rails loads)
- spec/lib/training_module_due_date_manager_spec.rb: Changes before(:all) to before(:each) and after(:all) to after(:each) for better test isolation

As per RuboCop RSpec docs, before(:each) { ... } has the same behavior as before { ... }. Omitting the argument uses the implicit style, where the hook runs before each example by default. This change aligns the spec with RuboCop’s recommended RSpec hook conventions and avoids unnecessary explicit arguments.

This PR addresses Rails 7.1 compatibility requirements and ensures the application follows Rails 7.1 configuration conventions.

Refrence docs - 
https://guides.rubyonrails.org/7_1_release_notes.html#active-record
https://guides.rubyonrails.org/7_1_release_notes.html#action-dispatch
https://github.com/rails/rails/releases/tag/v7.1.0
https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/HookArgument

## Screenshots
Before:
N/A - Configuration changes only, no UI impact

After:
N/A - Configuration changes only, no UI impact

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -59,8 +59,6 @@ module WikiEduDashboard
     # Handle YAML safe loading of serialized Ruby objects
     config.active_record.yaml_column_permitted_classes = [Symbol, BigDecimal, DateTime, Date, Time, ActiveSupport::TimeWithZone, ActiveSupport::TimeZone]
 
-    config.active_record.legacy_connection_handling = false
-
     config.action_dispatch.return_only_media_type_on_content_type = false
 
     config.middleware.insert_before 0, Rack::Cors do
--- a/spec/lib/training_module_due_date_manager_spec.rb
+++ b/spec/lib/training_module_due_date_manager_spec.rb
@@ -191,9 +191,9 @@ describe TrainingModuleDueDateManager do
                      .overall_due_date
     end
 
-    before(:all) { travel_to Date.new(2015, 8, 25) }
+    before { travel_to Date.new(2015, 8, 25) }
 
-    after(:all) { travel_back }
+    after { travel_back }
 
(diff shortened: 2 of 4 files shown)


config/application.rb
```diff
@@ -59,7 +59,7 @@ class Application < Rails::Application
     # Handle YAML safe loading of serialized Ruby objects
     config.active_record.yaml_column_permitted_classes = [Symbol, BigDecimal, DateTime, Date, Time, ActiveSupport::TimeWithZone, ActiveSupport::TimeZone]
 
-    config.active_record.legacy_connection_handling = false
+    # legacy_connection_handling was removed in Rails 7.1
```
ragesoss: I think this can simply be removed. No need to leave a comment in place.

ragesoss:
The changes look good. Please consolidate this into a single commit.

PR #6638, opened 2026-01-16
Refactor course helper and presenter for improved stat handling

SanjanaSogimatt:
# PR 3: Bug Fixes - Type conversion, association name, and view condition fixes

## What this PR does
This PR fixes several bugs across the application:

**Changes:**
1. **`app/presenters/courses_presenter.rb`**: Fixes incorrect association name from `course_stats` (plural) to `course_stat` (singular) in the `combined_wikidata_stats` method
2. **`app/views/layouts/surveys.html.haml`**: Fixes JavaScript loading condition by replacing `page_class == 'surveys results'` with explicit `controller_name == 'survey_results' && action_name == 'results'` check

These fixes address type conversion issues, incorrect association references, view rendering conditions, and test assertion failures.

## Screenshots
Before:
N/A - Bug fixes may not have visible UI impact, but could cause runtime errors or incorrect behavior

After:
N/A - Bug fixes ensure correct functionality

### DIFF
--- a/app/presenters/courses_presenter.rb
+++ b/app/presenters/courses_presenter.rb
@@ -244,7 +244,7 @@ class CoursesPresenter
   end
 
   def combined_wikidata_stats
-    @stats ||= courses.includes(:course_stat).where.not(course_stats: nil).filter_map do |course|
+    @stats ||= courses.includes(:course_stat).where.not(course_stat: nil).filter_map do |course|
       course.course_stat&.stats_hash&.[]('www.wikidata.org')
     end
 
--- a/app/views/layouts/surveys.html.haml
+++ b/app/views/layouts/surveys.html.haml
@@ -19,7 +19,7 @@
     = render "shared/foot"
     // survey.js sometimes interferes with the rendering of results.
     = javascript_include_tag '/assets/javascripts/jquery.min.js'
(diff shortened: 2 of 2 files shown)


app/helpers/course_helper.rb
```diff
@@ -40,7 +40,7 @@ def format_course_stats(course_stats)
       end
       # convert stats to human readable values
       course_stats[wiki_ns_key].each do |stat_key, stat|
-        course_stats[wiki_ns_key][stat_key] = number_to_human(stat)
+        course_stats[wiki_ns_key][stat_key] = number_to_human(stat).to_s
```
ragesoss: What situations is this relevant for? Are there specs that would break on Rails 7.1 without this change?

ragesoss:
It looks like the i18n changes are not included?

ragesoss:
Ah, I see that you undid the i18n change that was part of the first commit, because it broke as spec. Please clean up the commit history (either by replacing the two commits with a single one, or ideally making separate commits for each of the three independent changes).

app/helpers/course_helper.rb
SanjanaSogimatt: something was breaking fs that's why i added this, it was returning a hash instead of string but now when i am testing it, I don't think it's needed i'll remove it

PR #6639, opened 2026-01-16
Refactor TrainingModuleDueDateManager specs 

SanjanaSogimatt:
# PR 4: Test Improvements - TrainingModuleDueDateManager spec refactoring

## What this PR does
This PR improves test isolation and clarity in the `TrainingModuleDueDateManager` spec by changing from `before(:all)` to `before(:each)` hooks and adding detailed comments explaining the expected date calculations.

**Changes:**
- `spec/lib/training_module_due_date_manager_spec.rb`: 
  - Changes `before(:all)` to `before(:each)` and `after(:all)` to `after(:each)` for better test isolation
  - Adds detailed comments explaining the date calculations in two test cases:
    - Explains how `calculated_due_date` is computed when a block has no due date
    - Clarifies the expected minimum date calculation when both blocks have due dates

## Screenshots
Before:
N/A - Test-only changes, no UI impact

After:
N/A - Test-only changes, no UI impact

### DIFF
--- a/spec/lib/training_module_due_date_manager_spec.rb
+++ b/spec/lib/training_module_due_date_manager_spec.rb
@@ -191,9 +191,9 @@ describe TrainingModuleDueDateManager do
                      .overall_due_date
     end
 
-    before(:each) { travel_to Date.new(2015, 8, 25) }
+    before { travel_to Date.new(2015, 8, 25) }
 
-    after(:each) { travel_back }
+    after { travel_back }
 
     let!(:cu) { create(:courses_user, user_id: user&.id, course_id: course.id) }
 


spec/lib/training_module_due_date_manager_spec.rb
```diff
@@ -191,9 +191,9 @@
                      .overall_due_date
     end
 
-    before(:all) { travel_to Date.new(2015, 8, 25) }
+    before(:each) { travel_to Date.new(2015, 8, 25) }
```
ragesoss: Based on the rubocop docs, `before(:each) { ... }` should be the same behavior as simply `before { ... }`, as leaving out an argument is the "implicit" style where the `before` block is executed by default before each example: https://www.rubydoc.info/gems/rubocop-rspec/1.7.0/RuboCop/Cop/RSpec/HookArgument

Is that accurate? If so, I suggest using the implicit style (and then no change to the rubocop config will be necessary).

ragesoss:
The PR description or commit message should explain how this relates to the 7.1 upgrade.

PR #6640, opened 2026-01-16
Update Capistrano version to 3.20.0

SanjanaSogimatt:
# PR 5: Dependency Updates and Minor Fixes

## What this PR does
This PR updates the Capistrano deployment dependency and fixes a minor file formatting issue.

**Changes:**
- `config/deploy.rb`: Updates Capistrano version lock from `3.17.1` to `3.20.0` to get the latest bug fixes and improvements

These are minor maintenance updates that don't affect application functionality but keep dependencies up to date and maintain code style consistency.

## Screenshots
Before:
N/A - No UI or functional changes

After:
N/A - No UI or functional changes

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -117,7 +117,7 @@ GEM
       activerecord (>= 4.2)
     addressable (2.8.5)
       public_suffix (>= 2.0.2, < 6.0)
-    airbrussh (1.4.1)
+    airbrussh (1.6.0)
       sshkit (>= 1.6.1, != 1.7.0)
     annotate (3.2.0)
       activerecord (>= 3.2, < 8.0)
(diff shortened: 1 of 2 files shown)


spec/models/article_scoped_program_spec.rb
```diff
@@ -160,4 +160,4 @@
       expect(course.scoped_article?(wiki, 'Unassigned_article', 400)).to eq(false)
     end
   end
-end
+end
```
ragesoss: It looks like this has removed the newline, which caused the build to fail.

Ajay-singh1 (approved):
LGTM!

ragesoss:
Update the PR descrption and this should be good to go. Is this required for Rails 7.1? If so, why?

SanjanaSogimatt:
> Update the PR descrption and this should be good to go. Is this required for Rails 7.1? If so, why?

The `capistrano` got updated when rails was updated

ragesoss:
@SanjanaSogimatt the PR description still needs to be cleanup up, as no changes to the spec file are included.

PR #6641, opened 2026-01-17
Add backup scripts

gabina:
## What this PR does
This PR adds the back scripts that we expect to be run in the db server to execute the backup (issue https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6579).

- **Pre-backup tasks:** check there is enough free space, create backup folder for today, create **waiting** backup record, wait until all pertinent jobs are sleeping, update the backup record to **running**.

- **Backup tasks**: run the backup through the `mariadb-dump` command (previously known as `mysqldump`). 

- **Post back tasks**: update the backup record to **finished**.

## Free space check
**For peony db:**
- The volume has 500 GB.
- The current dashboard db (without revisions table) uses ~15GB.
- The plain logical dump is 12GB.
- The compressed logical dump is 1.5GB.
I set `MIN_FREE_SPACE` to 20 GB to ensure that even after decompressing the backup, we still have some free space

**For wiki edu db:**
I think the db uses ~3GB and we have more than 500 GB of free space, so setting `MIN_FREE_SPACE` to 5 GB seems reasonable.

## How to run the script (for both servers)
**1.** Create a system linux user called `dbbackup` :heavy_check_mark: 

`sudo adduser --system --group dbbackup --home dbbackup`

**2.** Add a sql user for it and grant the following permissions:  :heavy_check_mark: 
```
CREATE USER 'dbbackup'@'localhost' IDENTIFIED VIA unix_socket;

GRANT SELECT, SHOW VIEW, TRIGGER, LOCK TABLES ON dashboard.* TO 'dbbackup'@'localhost';

GRANT INSERT, SELECT, UPDATE ON dashboard.backups TO 'dbbackup'@'localhost';
```
**3.** Copy the script and queries to the server.  :heavy_check_mark: 

**4.** Update the script with correct `BACKUP_ROUTE`, `QUERY_ROUTE` and `JSON_ENDPOINT`.  :heavy_check_mark: 

**5.** Run `sudo -u dbbackup nano backup.sh`  :heavy_check_mark: 

## mariadb-backup vs mariadb-dump

There are (at least) two ways of running backups in our db server:

1. `mariadb-backup` (physical backup) and
2. `mariadb-dump` (logical backup). 

From[ mariadb docs](https://mariadb.com/docs/tools/mariadb-enterprise-operator/backup-and-restore/physical_backup):
### Physical backup
_A physical backup is a snapshot of the entire data directory `(/var/lib/mysql`), including all data files. This type of backup captures the exact state of the database at a specific point in time, allowing for quick restoration in case of data loss or corruption._
_Physical backups are the recommended method for backing up MariaDB databases, especially in production environments, as they are faster and more efficient than logical backups._

### Logical backup
_A logical backup is a backup that contains the logical structure of the database, such as tables, indexes, and data, rather than the physical storage format. It is created using [mariadb-dump](https://mariadb.com/docs/server/clients-and-utilities/backup-restore-and-import-clients/mariadb-dump), which generates SQL statements that can be used to recreate the database schema and populate it with data._
_Logical backups serve not just as a source of restoration, but also enable data mobility between MariaDB instances. These backups are called "logical" because they are independent from the MariaDB topology, as they only contain DDLs and INSERT statements to populate data._
_Although logical backups are a great fit for data mobility and migrations, they are not as efficient as [physical backups](https://mariadb.com/docs/tools/mariadb-enterprise-operator/backup-and-restore/physical_backup) for large databases. For this reason, physical backups are the recommended method for backing up MariaDB databases, especially in production environments._

### Research
A research was done before deciding which tool to use. There were two main aspects to consider:
I did a research before taking a decision about which one to use. There were two basic points to take into account:
- **Ability to run backups while the system is live**. The backup process must run without guaranteeing that no transactions occur at the same time. Although we stop course updates, which involve long transactions, short transactions may still run during the backup.
According to what I research, the time spent with locked tables with `mariadb-backup` is minimal (source: this official MariaDB [video](https://www.youtube.com/watch?v=xB4ImmmzXqU)). It waits until all queries complete before issuing the global lock (`FLUSH TABLES WITH READ LOCK`)  to prevent data from changing during the backup process. However, according to [How mariadb-backup Works](https://mariadb.com/docs/server/server-usage/backup-and-restore/mariadb-backup/how-mariadb-backup-works), writes are blocked only very briefly. That's why `mariadb-backup` is often recommended.
On the other side, mariadb-dump provides the `--single-transaction` option for InnoDB. This option starts a transaction before dumping and reads data from a consistent snapshot without locking the tables for extended periods, allowing concurrent reads and writes
- **Portability**. Logical backups are preferred when portability matters, since physical backups are not portable, even across different versions of the same MariaDB server. For this reason, we chose to use mariadb-dump.

## Logs
Successful example:
```
01262026 16:11:06 : Starting
01262026 16:11:06 : Waiting
01262026 16:11:06 : Ready
01262026 16:11:06 : Running backup
01262026 16:11:11 : Finishing
```

Error example:
```
01262026 16:14:14 : Starting
01262026 16:14:14 : Waiting
ERROR 1045 (28000): Access denied for user 'backup'@'localhost' (using password: YES)
01262026 16:14:14 : ERROR at line 57: command 'mysql < $QUERY_ROUTE/waiting.sql' exited with 1
ERROR 1045 (28000): Access denied for user 'backup'@'localhost' (using password: YES)

```

## TODO
Add automatic deletion for old backups once we created a new one.

## AI usage
Used ChatGPT for building the bash script, but always checking the manual for commands I didn't know.

### DIFF
--- a/config/backup/backup.sh
+++ b/config/backup/backup.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+set -euo pipefail
+
+BACKUP_ROUTE="ROUTE_TO_BACKUPS"
+QUERY_ROUTE="ROUTE_TO_QUERIES"
+JSON_ENDPOINT="https://outreachdashboard.wmflabs.org/system/can_start_backup.json"
+
+log() {
+  printf '%s : %s\n' "$(date '+%m%d%Y %T')" "$1" >> "$LOG_FILE"
+}
+
+on_error() {
+  local exit_code=$?
+  log "ERROR at line $1: command '$BASH_COMMAND' exited with $exit_code"
+
+  mysql < "$QUERY_ROUTE/failed.sql" || true
+
+  exit "$exit_code"
+}
+
+check_app_is_ready_for_backup() {
+  curl -s \
+  -o /dev/null  \
+  -w "%{http_code}" \
+  $JSON_ENDPOINT
+}
+
+wait_until_ready_for_backup() {
+  local status=$(check_app_is_ready_for_backup)
+
+  while [ $status != 200 ];
+  do
+      sleep 300
+      status=$(check_app_is_ready_for_backup)
+  done
+}
+
+trap 'on_error $LINENO' ERR
+
+# Set datetime
+printf -v datetime '%(%Y-%m-%d-%H:%M:%S)T' -1
+BACKUP_DIR="$BACKUP_ROUTE/$datetime"
+
+# Create backup folder
+mkdir -p $BACKUP_DIR
+
+LOG_FILE="$BACKUP_DIR/log-$datetime.log"
+exec >>"$LOG_FILE" 2>&1
+
+log "Starting"
+
+# Calculate min free space in bytes based on last backup
+LATEST_DIR=$(ls -1d "$BACKUP_ROUTE"/*/ | sort | tail -2 | head -1)
+LAST_BACKUP=$(ls "$LATEST_DIR"/*.sql.gz | head -1)
+
+LAST_BACKUP_SIZE=$(stat -c%s "$LAST_BACKUP")
+# Backups are stored in compressed form. We require 30x the size of the previous backup,
+# assuming the compressed version is about 10 times smaller 
+MIN_FREE_SPACE=$(( LAST_BACKUP_SIZE * 30 ))
+
+# Calculate free space in bytes
+FREE_SPACE=$(df -B 1 --output=avail $BACKUP_DIR | tail -1 | tr -dc '0-9')
+if [ $FREE_SPACE -lt $MIN_FREE_SPACE ]; then
+  log "Not enough free space. Aborting"
+  exit 1
+fi
+
+# Create waiting backup record
+mysql < $QUERY_ROUTE/waiting.sql
+
+# Sleep two minutes to guarantee that all processes see the new data table record
+sleep 120
+
+wait_until_ready_for_backup
+
+log "Ready"
+# Update waiting backup record to 'running'
+mysql < $QUERY_ROUTE/running.sql
+
+log "Running backup"
+# Run backup
+mariadb-dump --single-transaction --routines --databases dashboard | gzip > $BACKUP_DIR/dashboard-dump-$datetime.sql.gz
+
+log "Finishing"
+# Update running backup record to 'finished'
+mysql < $QUERY_ROUTE/finished.sql


config/backup/backup.sh
```diff
@@ -0,0 +1,80 @@
...
+
+wait_until_ready_for_backup
+
+log "Ready"
+# Update waiting backup record to 'running'
+mysql < $QUERY_ROUTE/running.sql
+
+log "Running backup"
+# Run backup
+mariadb-dump --single-transaction --routines --databases dashboard | gzip > $BACKUP_DIR/dashboard-dump-$datetime.sql.gz
```
gabina, line 83: Excluded `--triggers` from what's done in [T411890](https://phabricator.wikimedia.org/T411890) because that option is enabled by default.

config/backup/backup.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
+
+set -euo pipefail
```
ragesoss, line 3: I'm not familiar with this command. What is it for?

config/backup/backup.sh
```diff
@@ -0,0 +1,80 @@
...
+# Create backup folder
+mkdir -p $BACKUP_DIR
+
+LOG_FILE="$BACKUP_DIR/log-$datetime.log"
+exec >>"$LOG_FILE" 2>&1
+
+log "Starting"
+
+# Calculate free space in GB
+MIN_FREE_SPACE=20
```
ragesoss: This seems dangerous to hard-code, as database backup size will increase over time and could conceivably be larger than 20GB at some point. I guess this is okay, as it's not likely to be a problem before we run into other limitations, but I would feel a little better about it if it used some heuristic like '3x the size of the previous backup' as the minimum free space.

config/backup/backup.sh
gabina: It was an AI suggestion. It's a configuration to try to ensure that any error or unexpected behavior will be caught, which I think it's desirable in this case, since we want to be as sure as possible that if the backup ran and finished, everything went well.

From [this stackoverflow post](https://stackoverflow.com/questions/69570801/difference-of-e-u-and-o-pipefailin-bash-exit-status):

`set -e`: Exit immediately if a command exits with a non-zero status.

`set -u:` If you try to access an undefined variable, that is an error.

`set -o pipefail`: If any command in a pipeline returns a non-zero exit code, the return code of the entire pipeline is the exit code of the last failed command.

config/backup/backup.sh
gabina: Yeah, that makes sense. The only point is that the script currently stores the backup in compressed form, which I think is reasonable. Even having three times the size of the previous backup might still not be enough space to decompress it afterward, which would be a bad situation. Maybe we could require something like 30x the size of the previous backup, assuming the compressed version is about 10 times smaller than the uncompressed one?

ragesoss: Sure, that sounds reasonable.

ragesoss: Does the compression happen after the uncompressed file is generated? If so, we could potentially store some metadata about the backup, such as compressed and uncompressed size, in the Backup record in the database itself. Just an idea... and on second thought, probably that's more complicated than it needs to be. 30x the previous backup seems fine.

config/backup/backup.sh
```diff
@@ -0,0 +1,87 @@
...
+# Create backup folder
+mkdir -p $BACKUP_DIR
+
+LOG_FILE="$BACKUP_DIR/log-$datetime.log"
+exec >>"$LOG_FILE" 2>&1
+
+log "Starting"
+
+# Calculate min free space in bytes based on last backup
+LATEST_DIR=$(ls -1d "$BACKUP_ROUTE"/*/ | sort | tail -2 | head -1)
```
gabina, line 54: Note that the backup directories are timestamps so lexicographical sort = chronological sort. Most recent backup directory is always the one just created when running `mkdir -p $BACKUP_DIR`, that's why I used `| tail -2 | head -1` to get the previous one.

config/backup/backup.sh
gabina: Already updated.

ragesoss:
LGTM

gabina:
**For both db servers:**

Ran `$ sudo crontab -u dbbackup -e`. I had to install cron for db peony server.

Added `0 20 * * SUN /home/dbbackup/backup.sh`

Output:
```
no crontab for dbbackup - using an empty one
crontab: installing new crontab

```

In addition, I scheduled a backup for today to test it in both services. The backups ran successfully and I already deleted the crontab lines for it. :heavy_check_mark:

PR #6643, opened 2026-01-20
Update Article Views documentation to reflect new calculation method

### LINKED ISSUE #6548: Improve doc for "Article Views"  metric with new implementation
The article views doc says:
_This is the estimated number of views based on a 30-day average for each article, through the most recent stats update. Views may decrease if the updated average is lower than previous counts._

However, we updated the implementation for article views metrics in #6456. The new behavior is explained in that PR description.  We should update the text in the tooltip (`view_count_doc`), and check if it appears anywhere else in the dashboard.

Image

nexpectArpit:
- Changed from '30-day average' to 'daily average from first revision during course to present'
- Updated both view_count_doc and view_count_doc_wikidata entries
- Addresses issue #6548 following implementation changes in PR #6456

## What this PR does
Updates the tooltip documentation for Article Views metrics to reflect the new calculation method implemented in PR #6456.

## Changes Made

- Updated `view_count_doc` from "30-day average" to "daily average from first revision during course to present"
- Updated `view_count_doc_wikidata` with the same change for Wikidata items  
- Both changes reflect the new implementation that calculates views from the article's first revision during the course period up to the present

## Code Changes
**Before:**
`
view_count_doc: This is the estimated number of views based on a 30-day average for each article, through the most recent stats update. Views may decrease if the updated average is lower than previous counts.
`
**After:**
`
view_count_doc: This is the estimated number of views based on daily average from the article's first revision during the course up to the present, through the most recent stats update. Views may decrease if the updated average is lower than previous counts.
`

## Screenshots
No before/after screenshots are provided for the following reasons:
- Testing the tooltip requires Wikipedia OAuth authentication, which requires setting up a Wikipedia account and configuring OAuth credentials in the development environment, but in my case , i am not able to made the account, due to account errors from wikipedia side.
- The change can be verified by reviewing the locale file modifications shown in the code diff above

Closes #6548

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1224,8 +1224,8 @@ en:
       other: total usages across languages
     view_count_description: Article Views
     view_count_description_wikidata: Item Views
-    view_count_doc: This is the estimated number of views based on a 30-day average for each article, through the most recent stats update. Views may decrease if the updated average is lower than previous counts.
-    view_count_doc_wikidata: This is the estimated number of views based on a 30-day average for each item, through the most recent stats update. Views may decrease if the updated average is lower than previous counts.
+    view_count_doc: This is the estimated number of views based on the daily average from the article's first revision during the course up to the most recent stats update.
+    view_count_doc_wikidata: This is the estimated number of views based on the daily average from the item's first revision during the course up to the most recent stats update.
     view_data_unavailable: No view data available
     view: Views
     wiki_ed_help: Use the 'Get Help' button to report a problem.


gabina:
Screenshots are something definitely nice to have in a PR like this one. 

Maybe you can try to populate the dashboard with some example courses through `rake dev:populate` (read [doc](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/docs/setup.md#populate-example-data)).

If you have problems with setting up a Wikipedia account, you can post the details in the slack channel and someone will help you.

nexpectArpit:
@gabina i already posted the issue related to Wikipedia account on slack, waiting for resolution

gabina:
In the meantime I think that maybe you can populate the dashboard with some courses without setting up a Wikipedia account. I may be wrong, but at first glance I don't know why you would need to login for that.

nexpectArpit:
•before(image taken from depoyed site)
Screenshot 2026-01-22 at 2 45 42 AM

•After(image from my local)
Screenshot 2026-01-22 at 2 45 04 AM

@gabina please take a look and please suggest if any changes required

gabina:
Thanks for taking this!

I think that "Views may decrease if the updated average is lower than previous counts." is no longer true with the new implementation. I believe views will never decrease, so we may remove that part and use something like:

_This is the estimated number of views based on the daily average from the article’s first revision during the course up to the most recent stats update._

WDYT?

In addition, would you mind adding a screenshot of the new tooltip version for the campaign overview too?

nexpectArpit:
@gabina Thanks for the feedback! I agree that removing the "views may decrease" part makes sense with the new implementation.

Regarding the suggested text, I think it's a good improvement. However, I was wondering if we could make it a bit more explanatory for users? 
Something like:
"This shows the total estimated page views for articles edited as part of this course. The count includes all views from when participants first started editing each article up to our most recent data update. View statistics are refreshed regularly, so recent activity may take some time to appear."

I think ,this gives users a simple meaning of what they're looking at and why there might be delays, but let me know what you think!

gabina:
Thanks for the suggestion.

I'm not an expert on user interface, but I think we should keep the tooltip message short. I think the original idea was just to explain what that metric is. I'm not sure if it's the right place to say something like "View statistics are refreshed regularly, so recent activity may take some time to appear.", since all metrics (not only views) are refreshed regularly, and may experiment delays. I don’t think there is anything particular about views that justifies clarifying something that applies to how updates work in general. Unlike the “Views may decrease” note, which referred to something that only happened for that specific metric, other metrics never decreased. Does it make sense?

nexpectArpit:
@gabina  Yes, You're right about keeping it focused on what the metric represents rather than general system behaviour . 
I will go with the original text that you suggest "This is the estimated number of views based on the daily average from the article’s first revision during the course up to the most recent stats update."

nexpectArpit:
•before(image taken from depoyed site)
Screenshot 2026-01-22 at 11 56 55 PM

•After(image of program page from my local)
Screenshot 2026-01-22 at 11 47 08 PM

•After(image of compaign page from my local)
Screenshot 2026-01-22 at 11 46 51 PM

@gabina Please take a look!

nexpectArpit:
hey @gabina Just a reminder, please review the updated PR.

gabina:
This looks great. Do you mind cleaning the github history (to just keep the last commit)?

nexpectArpit:
@gabina Please have a look!

gabina:
LGTM!

@ragesoss this PR only updates the text description for the "article views" metric, since it was outdated after #6456.

Issue #6627, opened 2026-01-08
Implement OAuth2


ragesoss:
Wikipedia may discontinue support for OAuth1 at some point, and we rely on a stack of auth-related gems that isn't easy to customize or fix. (In particular, our current implementation of the login does not support setting a User-Agent, so we have to work around that to comply with the Wikimedia User Agent policy.)

This may be a good starting point for figuring out how to implement OAuth2: https://github.com/wikimedia/mediawiki-ruby-api?tab=readme-ov-file#080-2023-10-26

We may also need to adjust the way we interface with the API to make edits.

Abishekcs:
Seems interesting. I can work on this!

katara-Jayprakash:
sounds intresting! love to work on this issue @Abishekcs sir.

Abishekcs:
@ragesoss would you be open to using [ruby-oauth/oauth2](https://github.com/ruby-oauth/oauth2) to replace it with `omniauth-mediawiki` as it supports only **OAuth 1.0**.

Also, the recording below works with `devise + ruby-oauth/oauth2` and i think the upgrade will probably involve keeping both the **Old and New Implementation** to keep the PRs as small as possible until we can totally remove the old version/implementation.

https://github.com/user-attachments/assets/aa86d368-ae4f-4fde-8a7a-cd76b8d8fe3e

Abishekcs:
@katara-Jayprakash I don’t think this is a beginner-friendly issue. That said, if you feel comfortable working on it, please feel free to suggest any ideas on how we could implement it.

ragesoss:
That sounds reasonable. That gem looks like it's the same one that @abartov used in the example implementation linked from the Mediawiki API gem doc. (It looks the repo was moved from https://gitlab.com/oauth-xx/oauth2/ to https://github.com/ruby-oauth/oauth2

Abishekcs:
yes it's the same gem.

katara-Jayprakash:
Sorry, @Abishekcs sir. I was busy with my campus placements. I’d love to have a look at it in my free time, and if you’d like to work on this, please feel free to take it on.

Abishekcs:
started working on this again. Has been pending for too long. Sorry.

ragesoss:
No need to apologize! Thank you!

PR #6644, opened 2026-01-24
chore: upgrade mediawiki_api gem to 0.9.0

Abishekcs:
Required For #6627

## What this PR does

Upgrades `mediawiki_api` gem from 0.7.1 to 0.9.0 to add support for the `oauth_access_token` method, which is required for OAuth 2.0 authentication with Wikimedia's API.


ragesoss:
This is not necessary. The `Gemfile.lock` is already set to 0.9.0 for this gem. It was a recent change I made.

Abishekcs:
my mistake i didn't had latest changes in my branch

Issue #6557, opened 2025-11-06
Generate CSVs of all courses and instructors for admins


ragesoss:
Wikimedia Foundation uses data from Programs & Events Dashboard to understand program organizing across the Wikimedia ecosystem.

We have a script that produces data in their requested format: `docs/analytics_scripts/peony_programs_2024-2025.rb`. That script was written to pull data from a specific time period, but they would like to be able to pull data on-demand.

Based on that script, add a feature accessible only to admins to generate and download a CSV of all courses, using the same format as in that script, but with additional columns for course ID and creation timestamp (so that it's easy to isolate just the courses added since the last time you pull it).

ragesoss:
The script wasn't written with any performance considerations in mind, so you may need to adapt querying from what's in the script to avoid N+1 queries.

sharon-kathambi:
Hi @ragesoss  I’d like to take this up. A couple quick clarifiers:

1. By “on-demand,” do you mean an admin-only UI action (button/link) that triggers CSV generation + download?
2. Should the export be all courses by default, or do you want any filters in the UI (e.g., created_at since a date, start/end range, campaign)? If filters are desired, which are most important?
3. For performance: is it acceptable to generate synchronously (request → CSV) using batched queries/includes, or do you expect this to run as a background job with a “download when ready” flow?

ragesoss:
Hi Sharon! Nice to see you here again!

1. Yes.
2. All courses by default. I don't think this needs to be configurable with any options.
3. This should be done as a background job with the download-when-ready flow we use on other CSVs.

sharon-kathambi:
@ragesoss  Good to be back! I have sent an invite for slack.
Thank you for the clarifications, I will start on it.

ragesoss:
It looks like you already already have access to the slack.

PR #6647, opened 2026-01-27
Add admin-only background CSV export for all public courses and instructors

sharon-kathambi:
## What this PR does
This PR adds an admin-only CSV export that allows downloading a global list of all public courses and their instructors, without needing to select a specific campaign.
Addresses issue:  #[https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6557](6557)
#6557 

## Issue

The original CSV script:
- was run manually
- was not optimized for performance (N+1 queries)
Admins need a reliable way to pull tabular data for all public courses and instructors across campaigns, and to easily identify new courses since the last export.

### `ReportsController`

* Added a new **admin-only action** `all_courses_and_instructors_csv`.
* The action follows the existing CSV pattern:

  * checks if the CSV file already exists
  * redirects to the file if present
  * otherwise enqueues background generation and returns a “file is being generated” message.
* Ensures only **public courses** are included.
* No synchronous CSV generation in the controller.

### Routes

* Added a new route:
"all_courses_and_instructors_csv",
      to: "reports#all_courses_and_instructors_csv",

### `ReportCsvWorker`

* Extended the existing worker to support a new CSV type:
  **`all_courses_and_instructors`**
* CSV generation now:

  * preloads associations (`instructors`, `wikis`) to avoid N+1 queries
  * uses `find_each` for batch processing
  * writes output to `public/system/analytics`
* Added helper methods to keep the worker RuboCop-compliant and readable.
* Reuses existing cleanup behavior to delete generated CSVs after one week.

### No changes

* Existing campaign- and course-specific CSV exports are **unchanged**.
* No changes to the CSV builders used elsewhere.

## CSV format
Each course expands into multiple rows:
- one row per instructor
- one row per wiki

Columns:
Course ID | Created At | Slug | Title | Institution | Start | End | Facilitator | Wiki
Sample Generated CSV: [all-courses-and-instructors-2026-01-26.csv](https://github.com/user-attachments/files/24871071/all-courses-and-instructors-2026-01-26.csv)

## UI
The export is visible and accessible only to admins in the /usage page
If the CSV is not yet generated, the endpoint returns a “file is being generated” message.

## AI usage
To paraphrase the PR
Debug

## Screenshots
Before:
< add a screenshot or screen recording of the UI before your change >

After:
Screenshot 2026-01-27 at 01 29 45

Screenshot 2026-01-27 at 02 39 47


Abishekcs:
@sharon-kathambi `.yarn*, schema.db, .yarnrc.yml, package.json and yarn.lock ` are not related to your changes so you can remove them

app/workers/report_csv_worker.rb
```diff
@@ -17,12 +17,21 @@ def self.generate_csv(source:, filename:, type:, include_course:)
     perform_async(source.id, filename, type, include_course)
   end
 
+  def self.generate_all_courses_and_instructors_csv(filename:)
```
Abishekcs: You  can use the existing `generate_csv`. With proper method name in `reports_controller.rb` it should be clear that we are trying to generate all the csv.

app/workers/report_csv_worker.rb
```diff
@@ -72,4 +81,74 @@ def write_csv(filename, data)
...
+  def write_course_rows(csv, course)
+    base_row = base_course_row(course)
+    write_facilitator_rows(csv, base_row, course)
+    write_wiki_rows(csv, base_row, course)
+  end
+
+  def base_course_row(course)
+    [
+      course.id,
+      course.created_at&.utc&.iso8601,
```
Abishekcs: is utc and iso8691  necessary?

app/workers/report_csv_worker.rb
sharon-kathambi: I reused the existing generate_csv method as suggested. The only change I made was a small update to allow source to be nil for the global export case, since this CSV isn’t tied to a specific course or campaign.

app/workers/report_csv_worker.rb
sharon-kathambi: I used utc.iso8601 to make it easier to diff across exports. Would a normal .to_s suffice?

Abishekcs: No it's okay.

app/workers/report_csv_worker.rb
```diff
@@ -13,16 +13,23 @@ class ReportCsvWorker
   sidekiq_options lock: :until_executed
 
   # Generate the csv for the given source (course or campaign)
+  # if type is global, then can access source as nil
   def self.generate_csv(source:, filename:, type:, include_course:)
-    perform_async(source.id, filename, type, include_course)
+    id = source&.id
```
Abishekcs:
```suggestion
    perform_async(source&.id, filename, type, include_course)
```

Abishekcs:
@sharon-kathambi you should probably clean your git history too.  Refer to `PULL_REQUEST_TEMPLATE.md`

Abishekcs:
LGTM!

app/workers/report_csv_worker.rb
```diff
@@ -72,4 +78,74 @@ def write_csv(filename, data)
...
   end
+
+  def all_courses_and_instructors_to_csv
+    CSV.generate do |csv|
+      write_all_courses_headers(csv)
+      each_public_course do |course|
+        write_course_rows(csv, course)
+      end
+    end
+  end
```
gabina: My two cents here: it may be worthwhile to create a separate class that contains the logic for the new CSV report. Note that this class doesn't have the logic for any other report; it's just a kind of router that decides which builder to invoke. WDYT?

app/workers/report_csv_worker.rb
```diff
@@ -13,16 +13,22 @@ class ReportCsvWorker
...
+    data =
+      if type == 'all_courses_and_instructors'
+        all_courses_and_instructors_to_csv
+      elsif course_report?(type)
+        to_course_csv(type, id)
+      else
+        to_campaign_csv(type, id, include_course)
+      end
+
+    raise "CSV data was nil for type=#{type}" if data.nil?
```
gabina: What is the purpose of this `raise`?

app/workers/report_csv_worker.rb
sharon-kathambi: I see your point. Right now the worker is doing both dispatch and report-specific CSV construction, which isn’t ideal long-term. I kept the logic here initially to consolidate the reporting flow and reuse the existing background generation pattern, but I agree this report-specific logic could live in a dedicated builder. I will refactor accordingly.

app/workers/report_csv_worker.rb
sharon-kathambi: If `type` is passed incorrectly (e.g. due to a typo) or a new `type` is added later without being handled, the builder would return `nil`. The check ensures it does not silently generate an empty or invalid report. Or is it too strict..I can use logger instead or remove all together.

gabina: I understand your point, but I think we don’t usually use `raise` when we don’t expect any part of the code to handle that exception, which seems to be the case here, right?
Logging to Sentry might be a better option. However,  we should make sure that nil data is always an error and that we never expect a CSV report to legitimately be nil in some situations.

Also, which are the consequences for the user if we never call `write_csv`? Will the report be queued for regeneration every time the user wants to see if it has already been done?

sharon-kathambi: There isn’t a legitimate scenario where data should be `nil`,  each report type is expected to always return a CSV string, even if it contains only headers. If data is nil, that indicates a bug.

I’m fine switching this to logging to Sentry instead of raising if that’s the preferred pattern, but in that case we’d need some way to surface or track the failure. Otherwise, if `write_csv` is never called, the file won’t exist and the request will keep re-queueing the job, which would be a silent worst-case failure.

gabina: I agree that we should prevent the request from queueing the job indefinitely, although I understand that's completely independent of using `raise` or logging to Sentry.

sharon-kathambi: To keep this PR small, I’ll just log to Sentry if `data` is `nil`. This shouldn’t happen in normal use,  right now there’s only one case for this `type`, so a typo is very unlikely. The previous code didn’t guard for this - I mainly added the check to be safe if more types are added in the future.

I agree we should stop the request from queueing the job repeatedly, but that’s a separate concern and can be handled in a follow-up issue. What do you think?

ragesoss: I think `ReportCsvWorker` is already configured with a unique lock (`sidekiq_options lock: :until_executed`), so as long the parameters match, we shouldn't have to worry about adding multiple copies of the job into the queue.

sharon-kathambi: @ragesoss Good to know, that’s really helpful. Since the unique lock should prevent duplicate jobs for matching args, I’ve removed the Sentry log/raise for now.

sharon-kathambi:
Hey @ragesoss This PR is ready for review!

ragesoss:
This works smoothly, nice work! I messed up handling the merge conflict, but I'll get the sorted out and merge your changes separately from the this PR.

PR #6649, opened 2026-01-28
Upgrade Ruby version to 3.4.8 

SanjanaSogimatt:
## What this PR does

This PR upgrades the Ruby version from 3.1.2 to 3.4.8. This upgrade brings performance improvements, security patches, and access to the latest Ruby features. 

**Key changes:**
- Updated Ruby version requirement in `Gemfile` from `3.1.2` to `3.4.8`
- Added `csv` gem as an explicit dependency (required because CSV is no longer a default gem in Ruby 3.4+)
- Updated `Gemfile.lock` to reflect the new Ruby version and dependencies

**Why this upgrade:**
- Ruby 3.4.8 includes performance improvements and bug fixes
- Better compatibility with Rails 7.1.0
- Access to latest Ruby language features and improvements
- Security patches and stability improvements

**Breaking change note:**
In Ruby 3.4.0+, the `csv` standard library is no longer bundled as a default gem and must be explicitly added to the Gemfile. This PR adds it as a dependency to maintain compatibility with existing code that uses CSV functionality (e.g., `config/initializers/surveys.rb`).

Ref - https://www.ruby-lang.org/en/news/2025/12/17/ruby-3-4-8-released/

## AI usage

I used Cursor AI (Auto) to help with:
- Understanding the Ruby 3.4.8 upgrade requirements and breaking changes
- Diagnosing the CSV library loading error and identifying that it's no longer a default gem in Ruby 3.4+
- Resolving version manager conflicts between RVM and rbenv
- Verifying gem compatibility and installation

## Screenshots

**Before:**
- Ruby version: 3.1.2
- CSV library: Available as a default gem (no explicit dependency needed)

**After:**
- Ruby version: 3.4.8
- CSV library: Explicitly declared in Gemfile as `gem 'csv'`

**Note:** This is a backend upgrade with no visible UI changes. The upgrade maintains full backward compatibility for end users.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ gem 'faker', require: false # Generates random data for example records
 gem 'figaro' # easy access to ENV variables. Deprecated.
 gem 'puma'
 gem 'csv' # CSV library (required for Ruby 3.4+ as it's no longer a default gem)
+gem 'observer' # Observer library (required for Ruby 3.1+ as it's no longer in standard library)
 
 ### Database and caching
 gem 'mysql2' # MariaDB integration for ActiveRecord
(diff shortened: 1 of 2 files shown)


ragesoss:
Nice! It's great that this is passing with so few changes.

You should also include a commit that updates all the places in the docs and setup scripts that mention Ruby 3.1.2 specifically. (You don't need to change the cases where it's part of the path in example configurations like for Apache webserver, just the places in that relate to using and installing a specific Ruby version.

PR #6655, opened 2026-01-31
Feat/add settings for block bots

### LINKED ISSUE #6313: Add setting for disallowed usernames
Some courses on P&E Dashboard include bot accounts, which may have extremely high edit counts. Admins should be able to add high-edit bot usernames to a list of disallowed users, and prevent adding any of those users to a course.

lakshita10341:
## Fixes #6313 
This PR implements a new feature that allows super-admins to manage a list of "Disallowed Users" (e.g., bot accounts) who are strictly prevented from enrolling in any courses on the Dashboard. 

### Key Features:
1.  **Disallowed Users Management**: A new section in the Admin Settings allows adding and removing usernames.
2.  **User Existence and Super-admin Validation**: The system verifies that a username exists in the database before it can be added to the disallowed list and can only be added by super-admins.
3.  **Enrollment Blocking**: Both self-enrollment and instructor-led enrollment are automatically blocked for any user on the disallowed list, returning a clear error message.
4.  **High Edit Count Users Discovery**: A new "High Edit Count Users" helper section displays the top 30 users by total revision count across all courses, making it easy to identify and disallow potential bot accounts with a single click.

## AI usage
I used AI to:
- Write and fix RSpec tests to ensure full coverage of the new logic.
- Refine the UI layout and styling based on my feedback to match the existing design system.

## Screenshots/Video
[Screencast from 2026-02-01 00-12-27.webm](https://github.com/user-attachments/assets/df8c8938-b868-4067-9498-33e5e767b8a1)
image
image


app/assets/javascripts/actions/settings_actions.js
```diff
@@ -574,3 +575,134 @@ export const getSiteNotice = () => (dispatch) => {
   })
   .catch(data => dispatch({ type: API_FAIL, data }));
 };
+
+// Disallowed Users Actions
+const fetchDisallowedUsersPromise = async () => {
```
Abishekcs: I think all of this API endpoints can be moved to `api.js` under utils

lakshita10341: Hi @Abishekcs , just wanted to confirm, all the other apis in `settings_actions.js`  follow the same pattern (local promise functions). Moving only these new endpoints to `api.js`  would create inconsistency within the file, so should i proceed with moving them to `api.js`?

Abishekcs: hmm in most cases, when writing dispatch functions (i.e., actions), we don't want to worry about making API requests or handling the code for them. Newer versions of RTK follow this pattern as well (though we're on a much older version). It would be better to move the API logic to their respective files, abstracting them out so we don't have to worry about them in the dispatcher. Just make sure to add a comment for each function that makes an API request, indicating the purpose of each endpoint (i.e what resources are they providing).

ragesoss:
Thanks Lakshita!

I think I would prefer this be simplified a little bit. I don't think the 'high edit count' users table is necessary, and I worry that the query for it will impact performance of the settings page.

Can you remove the 'high edit count users' parts, and leave only the disallowed users portion?

lakshita10341:
> Thanks Lakshita!
> 
> I think I would prefer this be simplified a little bit. I don't think the 'high edit count' users table is necessary, and I worry that the query for it will impact performance of the settings page.
> 
> Can you remove the 'high edit count users' parts, and leave only the disallowed users portion?

@ragesoss  Done! I've removed the high edit count users section entirely. Only the disallowed users functionality remains now.

ragesoss:
This works smoothly! Nice work.

PR #6663, opened 2026-02-04
Fix: mainspace article validation for ClassroomProgramCour…

### LINKED ISSUE #6531: When assigning an article for a ClassroomProgramCourse, only mainspace titles should be allowed

Formasitchijoh:
## What this PR does
This PR resolves #6531 by cleaning up #6547 it adds a client-side mainspace article validation for ClassroomProgramCourse assignments for single and multi_wiki_assignment.This fix prevents broken assignments (e.g. Category:, Talk:, User: pages) that break stats, ArticleViewer, and impact tracking.

## AI usage
Used AI to understand the recursive chainSubmission method in the `add_available_article.jsx`  file  , Equally to catch edge cases for clearing  and displaying notification if an article title is invalid.

## Screenshots

https://www.loom.com/share/b7307c8073da4d00b438dac4ef544e31

@ragesoss should wikimedia incubator articles like `https://incubator.wikimedia.org/wiki/Wp/kiu/Heyder_Cansa ` be allowed or only mainspace articles are allowed for `ClassroomProgramCourses`

### DIFF
--- a/app/assets/javascripts/actions/article_actions.js
+++ b/app/assets/javascripts/actions/article_actions.js
@@ -112,15 +112,31 @@ export const crossCheckArticleTitle = (articleId, articleTitle, article_mw_page_
   };
 };
 
-export const verifyMainSpaceArticle = (articleTitle, language = 'en', project = 'wikipedia') => {
+/**
+ * Verifies if a given title is a valid mainspace page (existing or potential).
+ *
+ * Strategy:
+ * 1. Build the domain based on the wiki language/project provided by the UI.
+ * 2. Query MediaWiki API (action=query).
+ * 3. If 'invalid' is present: The title contains illegal characters.
+ * 4. Check 'ns' (namespace): Must be 0 for mainspace.
+ * 5. 'missing' is ALLOWED: This supports assigning articles that don't exist yet.
+ */
+
+export const verifyMainSpaceArticle = (articleTitle, language, project) => {
   return async (dispatch) => {
     try {
+      // Logic check: The UI should always provide these from the course settings or picker
+      if (!language || !project) {
+        return { valid: false, error: 'Language and project must be provided for article validation' };
+      }
       const title = articleTitle.trim();
 
       if (!title) {
         return { valid: false, error: 'Article title is required.' };
       }
 
+      // Handle Incubator and multi-domain logic
       let domain;
       if (project === 'wikimedia' && title.toLowerCase().startsWith('wp/')) {
         domain = 'incubator.wikimedia.org';
@@ -132,6 +148,7 @@ export const verifyMainSpaceArticle = (articleTitle, language = 'en', project = 
         domain = `${language}.${project}.org`;
       }
 
+      // Query article information from mediawiki based on the project home_wiki
       const apiUrl = `https://${domain}/w/api.php?action=query&titles=${encodeURIComponent(title)}&format=json&origin=*`;
 
       const response = await fetch(apiUrl);
@@ -140,13 +157,18 @@ export const verifyMainSpaceArticle = (articleTitle, language = 'en', project = 
         throw new Error(`API request failed with status ${response.status}`);
       }
       const data = await response.json();
+      // Extract the page information for the specific revision
       const pages = data?.query?.pages;
       const page = pages ? Object.values(pages)[0] : null;
 
-      if (!page || page.missing) {
-        return { valid: false, message: `The article does not exist on ${domain}` };
+      if (!page) return { valid: false, message: 'Unable to reach the Wiki API.' };
+
+      // Check for illegal characters (e.g. [ ] { } | )
+      if (page.invalid !== undefined) {
+        return { valid: false, message: `"${title}" is not a valid title format.` };
       }
 
+      // Block non-mainspace (Categories, User pages, etc.)
       if (page.ns !== 0) {
         return { valid: false, message: `${title} is not a mainspace article (only regular article pages are allowed)` };
       }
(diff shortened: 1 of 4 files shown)


ragesoss:
ClassroomProgramCourse is only used for Wiki Education Dashboard, and we don't have courses working on Incubator, so I think there's no need to handle that case.

app/assets/javascripts/actions/article_actions.js
```diff
@@ -112,6 +112,53 @@ export const crossCheckArticleTitle = (articleId, articleTitle, article_mw_page_
   };
 };
 
+export const verifyMainSpaceArticle = (articleTitle, language = 'en', project = 'wikipedia') => {
```
ragesoss: There's a lot going on here, so it would be good explain in comments what all this method is intended to do. What is the strategy for verifying that it's a mainspace article?

In what situation will the default language and project values be used? When a title in added in the UI, normally this should be treated as an article title on the home wiki of the course (which may not be English), unless the wiki picker has been used to set a different wiki. In either case, we ought to know enough to provide a language and project before calling this method.

app/assets/javascripts/actions/article_actions.js
```diff
@@ -112,6 +112,53 @@ export const crossCheckArticleTitle = (articleId, articleTitle, article_mw_page_
...
+
+      if (!response.ok) {
+        throw new Error(`API request failed with status ${response.status}`);
+      }
+      const data = await response.json();
+      const pages = data?.query?.pages;
+      const page = pages ? Object.values(pages)[0] : null;
+
+      if (!page || page.missing) {
+        return { valid: false, message: `The article does not exist on ${domain}` };
```
ragesoss: How does this work? We expect article titles that don't exist (and we just need to make sure they are valid potential mainspace titles), and I was able to add non-existent assigned articles without seeing this error message. Is there a particular situation where this should be triggered?

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -87,14 +88,16 @@
       expect(button).to have_content 'Assign/remove an article'
       button.click
       within('#users') do
-        first('input').set('https://wikisource.org/wiki/Heyder_Cansa', rapid: false)
+        first('input').set('https://en.wikisource.org/wiki/Alice%27s_Adventures_in_Wonderland',
```
ragesoss: I think the appropriate fix for these specs is to use a different course type, rather than change the examples. We do still want to be able to support the wikisource.org example and test for regressions, it's just not relevant for ClassroomProgramCourse.

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -124,4 +128,28 @@
...
+rapid: false)
+      end
+
+      click_button 'Assign'
+      sleep 1
+      visit "/courses/#{course.slug}/students/articles"
+      find('.student-selection .student').click
+
+      within('#users') do
+        expect(page).to have_no_content('Category:1993 in sports in Alberta')
```
ragesoss, line 228: It would be good to also check for the expected error message here. Simply checking for `have_no_content` can easily miss breaking changes.

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -124,4 +128,28 @@
...
+      find('.assign-button', text: 'Assign/remove an article').click
+
+      within('#users') do
+        find('input').set('https://en.wikipedia.org/wiki/Category:1993_in_sports_in_Alberta', 
+rapid: false)
+      end
+
+      click_button 'Assign'
+      sleep 1
+      visit "/courses/#{course.slug}/students/articles"
```
ragesoss: Try to write the test in a way that does not require a second call to `visit` to reload the page, which will make the `sleep` unnecessary. Instead, you can write an expectation based on some visible change you expect on the page after clicking `Assign`.

spec/features/multiwiki_assignment_spec.rb
```diff
@@ -36,6 +36,7 @@
         )
       end
       click_button 'Assign'
+      sleep 1
```
ragesoss: As much as possible, we want to avoid adding `sleep` to feature specs, as it adds up and makes the test suite take longer than necessary.

Tests that sometimes fail without `sleep` can usually be rewritten in a way to avoid it, such as writing an expectation that indicates the completion of whatever step needs to happen.

ragesoss:
I tested this locally, and it works nicely!

app/assets/javascripts/actions/article_actions.js
Formasitchijoh: Oh I think i missed that, Just to clarify

Remove the default `language` and `project`  in the method since each course will have a home_wiki it is tracking, hence should use that to verify the article

app/assets/javascripts/actions/article_actions.js
Formasitchijoh: No particular situation it should work if the article does not exist if the response object for the page has a missing attribute. Can you link the article you used to test this

ragesoss:
LGTM!

PR #6664, opened 2026-02-04
Fix NotSignedInError to redirect with UI instead of plain text

### LINKED ISSUE #6658: NotSignedInError results in plain "Please sign in" message with no UI for signing in
Any action that requires the user to be signed in will throw a `NotSignedInError`, which is handled via `lib/errors/rescue_errors.rb`.

Currently, it's a plain text response that simply says "Please sign in", but the navbar is not rendered so there's no way to log in. Ideally, doing something that triggers such an error would both give the user an obvious way to proceed to login, and also redirect to the original intended location after successful login.

One context where this is relevant is the followup questionnaire links in AI alert emails, so that could be a good example for building a test for, as part of implementing this.

nexpectArpit:
Fixes #6658

### What this PR does

This PR improves the experience when a signed-out user visits a protected page.

- Replaces the plain text **“Please sign in”** response with a proper **401 Unauthorized** page.
- Stores the intended destination in the session and automatically returns the user to that page after a successful Wikipedia login.
- Download Safety: Detects if the return path is a download-only endpoint (like /ungreeted) and redirects the user to the home page with a clear download link. This prevents the browser from getting stuck on the Wikipedia "Allow" page during silent file downloads.
- Security: Validates redirection paths to prevent open-redirect vulnerabilities.
- Tests: Adds unit tests for the redirection and security logic in ApplicationController.

### Screenshots

**Before**  
Screenshot 2026-02-05 at 3 58 15 AM

**After**  
Screenshot 2026-02-13 at 9 47 25 PM

@ragesoss

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,7 +27,27 @@ class ApplicationController < ActionController::Base
   end
 
   def after_sign_in_path_for(_resource_or_scope)
-    request.env['omniauth.origin'] || '/'
+    path = session.delete(:return_to) || request.env['omniauth.origin'] || '/'
+
+    # Security: ensure path is a relative path to prevent open redirects
+    begin
+      uri = URI.parse(path)
+      path = uri.path if uri.host
+      path = '/' unless path.start_with?('/')
+    rescue URI::InvalidURIError
+      path = '/'
+    end
+
+    # Special case for download/raw data paths:
+    # Sending users directly to these after login can be confusing (blank page or download starts).
+    # Instead, we send them to the home page with a clear success message and a link.
+    if path.include?('ungreeted') || path.include?('csv')
+      flash[:notice] = t('error.login_success_download_html',
+                         link: helpers.link_to(t('error.download_here'), path)).html_safe
+      return root_path
+    end
+
+    path
   end
 
   def check_onboarded
@@ -131,4 +151,6 @@ class ApplicationController < ActionController::Base
     return unless params[:locale]
     http_accept_language.user_preferred_languages.unshift(params[:locale])
   end
+
+
 end
--- a/app/views/errors/unauthorized.html.haml
+++ b/app/views/errors/unauthorized.html.haml
@@ -1,0 +1,15 @@
+- content_for :after_title, " - #{t('home.log_in')}"
+
+%section.unauthorized-error
+  .container.narrow
+    .error-card
+      %i.icon.icon-wiki-purple
+      %p.explanation= flash[:alert] || t('error_401.explanation')
+      
+      .actions
+        = link_to user_mediawiki_omniauth_authorize_path, method: :post, class: 'button dark auth big' do
+          %i.icon.icon-wiki-white
+          = t('home.log_in')
+        
+        %a.button.border.big(href="/")
+          = t('application.home')
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -192,9 +192,11 @@ describe ApplicationController do
     end
 
     context 'when user is not signed in' do
-      it 'returns an html 401' do
+      it 'returns an html 401 and stores the return path' do
         get :index
         expect(response.status).to eq(401)
+        expect(session[:return_to]).to eq('/anonymous')
+        expect(flash[:alert]).to eq('Please sign in.')
       end
 
       it 'returns a json 401' do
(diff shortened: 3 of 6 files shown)


nexpectArpit:
@ragesoss, I have considered the feedback of my copilot and made the changes. Kindly review it.

app/controllers/application_controller.rb
```diff
@@ -27,7 +27,24 @@ def after_sign_out_path_for(_resource_or_scope)
...
+    # Security: ensure path is a relative path to prevent open redirects
+    begin
+      uri = URI.parse(path)
+      path = uri.path if uri.host
+      path = '/' unless path.start_with?('/')
+    rescue URI::InvalidURIError
+      path = '/'
+    end
+
+    if path.include?('ungreeted') || path.include?('csv')
```
ragesoss, line 44: It's not clear to me what's going on here. Why are `ungreeted` and `csv` paths special-cased here?

db/schema.rb
```diff
@@ -10,16 +10,26 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2025_07_28_182700) do
-  create_table "alerts", id: :integer, charset: "utf8mb4", force: :cascade do |t|
+ActiveRecord::Schema[7.0].define(version: 2026_01_27_212149) do
```
ragesoss, line 13: Schema changes are not expected here.

ragesoss: It looks like our schema file is out of date and didn't get updated after recent migrations. But that's not relevant to this PR, so the schema change here should be excluded.

spec/controllers/application_controller_spec.rb
```diff
@@ -236,4 +238,43 @@ def index
...
+    it 'falls back to omniauth.origin' do
+      request.env['omniauth.origin'] = '/origin'
+      expect(controller.after_sign_in_path_for(user)).to eq('/origin')
+    end
+
+    it 'defaults to root path' do
+      expect(controller.after_sign_in_path_for(user)).to eq('/')
+    end
+
+    it 'validates return_to to prevent open redirects' do
```
ragesoss: I'm not clear on what situation is being tested here.

app/views/errors/unauthorized.html.haml
```diff
@@ -0,0 +1,14 @@
+- content_for :after_title, ' - Sign In Required'
+
+%header.main-page
+  .container
+    %h1= t('error_401.header', default: 'Sign In Required')
```
ragesoss: I think this is a pretty aggressive for a 'you need to sign in' message. It's fine for a 401 where the user is not signed in but not allowed to do what they tried to do, but I think for the specific case of needing to log in, the header is not necessary.

ragesoss (commented):
This looks like a reasonable approach overall.

app/controllers/application_controller.rb
nexpectArpit: It's just to make things less confusing. Since those links are only for downloading files or seeing raw data, sending a user straight there right after they log in can make it look like the site didn't load properly or just went to a blank page.

Sending them to the homepage with a clear 'Download here' link in the success message ensures they have successfully logged in and creates a smoother experience.

db/schema.rb
nexpectArpit: After updating the PR, I have reverted <unk> db/schema.rb<unk> to the expected version and made the necessary changes, thanks for pointing it out.

spec/controllers/application_controller_spec.rb
nexpectArpit: This test covers two simple cases around where a user is sent after login.
- First, it checks what happens if the stored return link points outside our site. In that case, the code ignores the unsafe part and sends the user to a safe internal page instead.
- Second, it checks the fallback behavior. If the return link is missing or broken, the app does not error out and redirects the user to the home page.

Actually I implemented these tests after considering feedback of co-pilot .
So the test makes sure the user always lands on a valid, safe page after login, regardless of what value ends up in the session.

app/views/errors/unauthorized.html.haml
nexpectArpit: i implemented  the changes, have a look!

app/controllers/application_controller.rb
ragesoss: Okay. That should be explained in a comment.

ragesoss:
Please update the screenshot and clean up the commit history.

nexpectArpit:
@ragesoss I cleaned the commit history and updated the Screenshot!

config/locales/ar.yml
```diff
@@ -1604,7 +1604,7 @@ ar:
       IP الخاص بك بواسطة Wikipedia أو وصل بالفعل إلى الحد اليومي للحسابات الجديدة.  حاول
       الاتصال من شبكة مختلفة.
   error_401:
-    header: غير مصرح. >:-(
```
ragesoss, line 1613: Rather than change this header message, I think it makes more sense to just not use the header in this particular case.  The header with the frown is still okay for other 401 situations. (When we do make changes to already-translated interface messages, we normally only change the English, as all the other locale files are basically maintained indirectly via translatewiki.net.)

ragesoss:
This looks like it has some unrelated changes, including CI and i18n. Those should be in separate PRs. This one should only include the changes directly related to the signin error.

nexpectArpit:
@ragesoss I noticed the build is failing on the test setup again. I  included those Redis changes in the first place because they seemed to be the only way to get the build to actually pass.
I've taken them out now to keep this PR focused strictly on the UI work as you asked, but now we're back to a failing build. If I move that setup fix to a separate PR, this UI one is going to stay 'red' for a while.

What’s the best way to move forward here? Should I put that fix back in just to get the green light on this PR, or are you okay with this one staying red while I handle the setup fix separately?

ragesoss:
@nexpectArpit i tried copying over those CI changes, but the build still failed. Not sure what's going on with that! But if you can get a PR that does fix CI, please do.

nexpectArpit:
@ragesoss I'm a bit confused to handle the PR though. just want a clarification like I can raise my original PR again, but it has everything together like the CI fix, the i18n changes, and the UI stuff. Would you prefer me to just raise that original one, or should I try to separate them? I'm just still a bit confused about the right way to do it when the changes are mixed like that.

Also, I wanted to check if you're seeing a lot of linting warnings on your side too? I'm getting about 220 warnings (mostly about 'literal strings') in my local setup, and I even see them when I'm on the master branch. I just want to be sure it's not some issue that is only happening on my side.

ragesoss:
I'd prefer separate PRs. The CI issue is not related to this, so it should be fixed on its own in a different PR.

The string literal warnings appear locally for all the JS components that have string literals instead of i18n messages. Fixes for these warnings in components that you're changing for a given issue is fine to include, but it should be limited to the components where other changes are required.

nexpectArpit:
@ragesoss I've raised a separate PR for the CI fix: #6674

There were actually two issues causing the CI failure:
- The Redis action was hitting Docker Hub rate limits
- The MariaDB action was using a Docker client version (1.40) that's too old for the current GitHub runners (minimum 1.44)

I replaced both with native GitHub Actions service containers. Once that PR is merged, the CI on this PR should pass!

ragesoss:
Great! You should rebase this branch so we can get a clean CI run.

nexpectArpit:
@ragesoss have a look.

Issue #6646, opened 2026-01-26
Site Notice feature is not user-friendly for admins


ragesoss:
Admins can configure a site notice banner for the Dashboard via `/settings`.

The input for this is too small to see it all at once, and there is no preview option. Implement a live preview and make it user-friendly to create or edit a site notice.

sharon-kathambi:
@ragesoss I'll take this up

sharon-kathambi:
Hey @ragesoss, just checking - the scope here is only to add a live preview, not to modify the input box itself, right? Something like this?

<img width="504" height="448" alt="Image" src="https://github.com/user-attachments/assets/dec1c800-cbea-410c-9acb-0ac13af49398" />

ragesoss:
@sharon-kathambi making the input box larger so that you see the full input is also part of the scope of this.

DevireddyHepsiba:
@ragesoss  sir ., i will take this up

sharon-kathambi:
@DevireddyHepsiba I have already worked on this..you can look for another issue

Worksalohy:
Hi @sharon-kathambi , are you still working on this issue?

sharon-kathambi:
@Worksalohy yes

ManishKumar081:
@sharon-kathambi any updates? Its been a while would be happy to help if required

sharon-kathambi:
Hi @ManishKumar081 thanks, I already implemented this.

nexpectArpit:
@sharon-kathambi are you still on that?

PR #6665, opened 2026-02-04
Add live preview to site notice settings

sharon-kathambi:
## What this PR does
This PR adds a live preview to the Site Notice configuration in /settings to make it easier for admins to create and edit site notices.
Addresses issue: #6646 

## Issue

- The input field was too small to properly view longer notices.
- There was no preview functionality 

## Changes Made
1. Replaced `TextInput `with `TextAreaInput`

- Allows better visibility of longer messages
- Enables auto-expand functionality
- Improves overall user experience when entering site notices
- Supports max length of 255 characters

2. Added `SiteNoticePreview` Component

- Displays a live preview of the site notice message
- Preview updates dynamically as the user types
- Preview is enabled only when the site notice is enabled
- Added break word functionality

3. State Sync Improvement

- Added `useEffect` to sync local state when `currentSiteNotice` changes.

4. Popover
- Set minimum width of popover to `800 px` for desktops and `75vw` for smaller screens

## AI usage
To paraphrase PR

## Screenshots
Before:
Screenshot 2026-02-04 at 15 46 46

After:

Screenshot 2026-02-17 at 14 48 42

Screenshot 2026-03-03 at 12 29 38

Screenshot 2026-03-03 at 12 36 15

### DIFF
--- a/app/assets/javascripts/components/settings/views/configure_site_notice.jsx
+++ b/app/assets/javascripts/components/settings/views/configure_site_notice.jsx
@@ -14,7 +14,14 @@ const ConfigureSiteNotice = (props) => {
   };
 
   const { isOpen, ref, open } = useExpandablePopover(getKey);
-  const form = <SiteNoticeForm handlePopoverClose={open} updateSiteNotice={updateSiteNotice} currentStatus={props.currentSiteNotice.status} />;
+  const form = (
+    <SiteNoticeForm
+      handlePopoverClose={open}
+      updateSiteNotice={updateSiteNotice}
+      currentStatus={props.currentSiteNotice.status}
+      currentSiteNotice={props.currentSiteNotice}
+    />
+  );
 
   const toggleHandler = (e) => {
     e.preventDefault();
@@ -42,6 +49,10 @@ const ConfigureSiteNotice = (props) => {
         is_open={isOpen}
         edit_row={form}
         right
+        styles={{
+          tableLayout: 'fixed',
+          width: 'min(800px, 75vw)',
+        }}
       />
     </div>
   );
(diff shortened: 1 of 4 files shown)


sharon-kathambi:
Hi @gabina @Abishekcs this PR is ready for review

Abishekcs:
https://github.com/user-attachments/assets/9859455d-6a02-49df-a58f-2ec49ac9cac0

Something is wrong with the styling as the box increases while typing.

ragesoss:
Thanks! This looks like a nice improvement already, and the code looks reasonable.

I think the whole popover ought to be wider by default, which will probably address the styling quirk that Abishek noted and will also make it feel more like a realistic preview of the full-width banner.

Also, please clean up the git history. It should probably all be consolidated into a single commit.

sharon-kathambi:
@Abishekcs @ragesoss okay, having a look at the changes requested.

sharon-kathambi:
> Thanks! This looks like a nice improvement already, and the code looks reasonable.
> 
> I think the whole popover ought to be wider by default, which will probably address the styling quirk that Abishek noted and will also make it feel more like a realistic preview of the full-width banner.
> 
> Also, please clean up the git history. It should probably all be consolidated into a single commit.

@ragesoss I’ve set a minimum width for the `popover` and added `word-breaking` to the preview section, since that was causing the popover to expand horizontally while typing (as @Abishekcs pointed out). I reused the existing `TextAreaInput `component with` rows={8}` instead of creating a separate component, to keep things consistent and avoid duplication, adjusting the width of that global text area input will be tricky.

ragesoss:
Thanks! Let me know when it's ready for review and I'll have another look.

sharon-kathambi:
@ragesoss PR is ready for review.

PR #6666, opened 2026-02-06
Fix: disable sandbox url and update button for userspace page

Formasitchijoh:
## What this PR does
This PR resolves https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6551 by disabling sandbox url and update button for userspace page.

## AI usage
No AI usage

## Screenshots

https://github.com/user-attachments/assets/8b8b872e-acf2-4bfb-beb2-9fcb985155c4

### DIFF
--- a/app/assets/javascripts/components/common/AssignmentLinks/AssignmentLinks.jsx
+++ b/app/assets/javascripts/components/common/AssignmentLinks/AssignmentLinks.jsx
@@ -11,7 +11,7 @@ import PeerReviewLink from './PeerReviewLink';
 import AllPeerReviewLinks from './AllPeerReviewLinks';
 import Separator from '@components/overview/my_articles/common/Separator.jsx';
 import ArticleUtils from '../../../utils/article_utils';
-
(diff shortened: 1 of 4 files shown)


ragesoss:
In this case (the example from the video), there should still be a link to the (user) page. This would be the same 'Article' link that is present for a mainspace page. And rather than leave a disabled "Sandbox Draft" text, that one should just be removed entirely.

Formasitchijoh:
> In this case (the example from the video), there should still be a link to the (user) page. This would be the same 'Article' link that is present for a mainspace page. And rather than leave a disabled "Sandbox Draft" text, that one should just be removed entirely.

Just to confirm my understanding:

For assignments in userspace (titles starting with User:), I should:
- Completely remove the "Sandbox Draft"  link/text (don't show it at all, even disabled)
- Keep showing the "Article" link, pointing directly to the userspace page (e.g. https://en.wikipedia.org/wiki/User:Formasit-chijoh) without any changes

I also removed the "Change Sandbox" button for these userspace cases. Does that align with what you want?

ragesoss:
Yes, that all sounds correct.

PR #6668, opened 2026-02-07
Make db transactions shorter

gabina:
## What this PR does
This PR tries to reduce the "Mysql2::Error::TimeoutError: Lock wait timeout exceeded; try restarting transaction" sentry logs.

### Problem
We have heavy parts of the code embedded in an `ActiveRecord::Base.transaction`.  While this was done to be safe in terms of guaranteeing db consistency during course updates, the code that was considered "atomic" was too big. It even includes API calls to fetch revisions and scores, which are known to be heavy and slow processes. That's not a good practice at all.

This PR updates the block of code that are enclosed with a db transaction, leaving parts of the code that fetch data out of them.

### Strategy
Remove the db transaction around `handle_timeslice` since it fetches transactions and scores. It still has the responsibility of logging the exception to Sentry if something goes wrong.

Now there are two shorter "cores" that we want to be atomic:
- `maybe_split`: we want a timeslice to be either completely split or not split at all. Leaving the "split process" halfway would result in an inconsistent state, since it involves removing the original timeslice and creating two new ones. In addition, new timeslices are now created with `needs_update` set to 1 to ensure that even if the process gets interrupted right after the timeslice is split, they will be processed in the next update.
- `update_timeslices`: to guarantee that a timeslice is processed completely.

## AI usage
No AI usage.

### DIFF
--- a/lib/course_revision_updater.rb
+++ b/lib/course_revision_updater.rb
@@ -94,14 +94,13 @@ class CourseRevisionUpdater
                                    .for_datetime(timeslice_start)
                                    .first
     if timeslice.nil?
-      # This can occur while a timeslice is being split and hasn’t been created yet.
-      # In that case, we treat it as new data only if at least one revision is fetched.
+      # This scenario is unexpected, so we log the message to understand why this happens.
       Sentry.capture_message 'No timeslice found for revision date',
-                             level: 'warning',
+                             level: 'error',
                              extra: { course_name: @course.slug,
                                        wiki: wiki.id,
                                        date: timeslice_start }
-      return !revision_count.zero?
+      return true
     end
 
     latest_revision = revisions.maximum(:date)


app/services/split_timeslice.rb
```diff
@@ -13,20 +13,26 @@ def initialize(course)
...
-    @timeslice_manager.maybe_create_course_wiki_timeslice(wiki.id, start_date, end_date)
-    return false, []
+    ActiveRecord::Base.transaction do
+      @revisions = revisions
+      if too_many_revisions?(wiki)
+        result = true, split_timeslice(wiki, start_date, end_date)
+      else
+        result = false, []
+      end
+      result
```
gabina, line 31: Note that in Rails 7, using `return` inside a transaction will roll the transaction back, so I avoided it.

lib/course_revision_updater.rb
```diff
@@ -94,14 +94,13 @@ def new_revisions?(revisions, wiki, timeslice_start)
...
-      # In that case, we treat it as new data only if at least one revision is fetched.
+      # This scenario is unexpected, so we log the message to understand why this happens.
       Sentry.capture_message 'No timeslice found for revision date',
-                             level: 'warning',
+                             level: 'error',
                              extra: { course_name: @course.slug,
                                        wiki: wiki.id,
                                        date: timeslice_start }
-      return !revision_count.zero?
+      return true
```
gabina, line 103: Undo what was done in [this commit ](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6461/changes/1a3a1c91cdef366681a2433ee52a922975731e2d#diff-91f34557e50b487cbe5731ea5d8ab12ab0f15ab0a1bef6a7e8069b680f6a4ffeR97).

Note that now timeslices are created immediately when `too_many_revisions?` is true, as part of `split_timeslice`:

```
    # Create new timeslices with needs_update set to 1
    @timeslice_manager.create_course_wiki_timeslice(wiki.id, start_date, midpoint)
    @timeslice_manager.create_course_wiki_timeslice(wiki.id, midpoint, end_date)
```
    
Therefore, now we never fetch revisions for a timeslice that has not yet been created.

ragesoss:
LGTM!

PR #6671, opened 2026-02-11
Fix interwiki assignments and user links 

### LINKED ISSUE #2491: Assignments in foreign wikis shows as interwiki in wikipage
Hi, 

I noticed that assignments in foreign wikis (in other wiki than course's home wiki) in courses with wiki page enabled shows as interwiki in the page. See https://cs.wikipedia.org/w/index.php?title=Wikipedie:N%C3%A1st%C4%9Bnka/Wikimedia_%C4%8CR/Workshop_psan%C3%AD_Wikipedie_-_zima_2019_(14-02-2019)&oldid=16952366 as an example (permanent link). Talien73's assignments are Pendon Museum@cswiki (shows correctly) and Slavonic Library in Prague@enwiki (doesn't show visually at all). If you examine the code, it's included as `[[En:Slavonic Library in Prague]]`. However, this is an (old school) method how to add an interwiki. If you want to link to a page at another wiki, the link must be prefixed with additional colon, so it will look like `[[:En:Slavonic Library in Prague]]`.

Thanks for fixing, 

Martin Urbanec
cs.wikipedia

nexpectArpit:
- Restored interwiki prefix parsing in AssignmentManager and models.
- Fixed Article link logic in AssignmentLinks to rely on article_id.
- Made user links in Assignments and Feedback interwiki-aware.
- Preserved raw input in AssignButton to prevent cursor jumps and prefix stripping.

 Closes #2491 

## Recording
Before:

https://github.com/user-attachments/assets/362dd2e8-9b42-4b77-b9ba-188022e740a0

After:

https://github.com/user-attachments/assets/b0a726d4-4109-479d-a5af-251c0c038a39

please review it @ragesoss @gabina @Abishekcs


app/models/wiki.rb
```diff
@@ -108,9 +108,51 @@ def self.language_for_multilingual(language:, project:)
...
+    'wikt' => 'wiktionary',
+    'q' => 'wikiquote',
+    'b' => 'wikibooks',
+    'n' => 'wikinews',
+    's' => 'wikisource',
+    'v' => 'wikiversity',
+    'voy' => 'wikivoyage',
+    'c' => 'wikimedia',
+    'wmf' => 'wikimedia',
+    'm' => 'metawikipedia'
```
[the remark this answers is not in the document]
nexpectArpit: Fixed! Mapped `m:` to `project: wikimedia, language: meta`

app/models/course_data/assignment.rb
```diff
@@ -147,6 +147,8 @@ def assignment_pipeline
   def set_defaults_and_normalize
     self.wiki_id ||= course.home_wiki.id
     return if article_title.nil?
+
+    self.article_title = CGI.unescape(article_title)
```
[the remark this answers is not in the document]
nexpectArpit: Fixed it, Replaced CGI.unescape with a conditional decoder that preserves +

app/assets/javascripts/reducers/revisions.js
```diff
@@ -42,16 +42,16 @@ const initialState = {
 };
 
 const isLimitReached = (course_start, last_date) => {
-  return isAfter(toDate(course_start), toDate(last_date)) || isBefore(toDate(last_date), subYears(new Date(), 5));
+  return isAfter(toDate(course_start), toDate(last_date, true)) || isBefore(toDate(last_date, true), subYears(new Date(), 5));
```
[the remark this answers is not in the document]
nexpectArpit: Added a null-date guard to isLimitReached

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
```diff
@@ -255,14 +256,14 @@ const AssignButton = ({ course, role, course_id, wikidataLabels = {}, hideAssign
...
-      articlesTitles.push(article.title);
+      articlesTitles.push(article.article_url ? article.title : articleTitle);
       articleLanguage = article.language;
       articleProject = article.project;
     });
 
-    setTitle(articlesTitles.join('\n'));
-    setProject(articleProject || project);
-    setLanguage(articleLanguage || language);
+    setTitle(text);
```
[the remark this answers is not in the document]
nexpectArpit: Fixed it

Abishekcs:
Ty @nexpectArpit i will try to review it within the next few days. Make sure to mark it ready for review when you are done with your final changes.

app/workers/ai_detection_worker.rb
```diff
@@ -7,18 +7,23 @@ class AiDetectionWorker
...
-    # suggest keeping job parameters small, simple, and JSON-compatible
-    attributes = { mw_rev_id: revision.mw_rev_id,
-                   wiki_id: wiki.id,
-                   article_id: revision.article_id,
-                   course_id: course.id,
-                   user_id: revision.user_id,
-                   revision_timestamp: revision.timestamp }
-    perform_async(attributes)
+    # Sidekiq good practices suggest keeping job parameters small, simple, and JSON-compatible.
+    # We pass IDs and integer timestamps instead of objects or hashes with symbol keys.
```
Abishekcs: this changes are not related to this PR?

nexpectArpit: Agree, but this was included because i encountered a test failure caused by this( `app/workers/ai_detection_worker.rb`) file during development and included the fix here to get CI passing. What do you suggest ?

Abishekcs: Please free to open a separate PR or issue for this if you want. And make sure to include the steps to reproduce the failure.

Abishekcs: And probably remove this part from this PR since it's unrelated.

nexpectArpit: I've reverted the file back so this PR stays focused on the interwiki assignment logic. And i think maybe we don't need to create the pr for the Cl failure issue, as this time it passed the tests. Any other feedback?

Abishekcs:
@nexpectArpit can you clean up your git commits

nexpectArpit:
@ragesoss @Abishekcs have a look.

nexpectArpit:
rebased with latest ruby version.

app/assets/javascripts/utils/course_utils.js
```diff
@@ -55,8 +56,83 @@ export default class CourseUtils {
...
+      }
+
+      const parts = normalizedInput.split(':');
+      if (parts.length >= 2) {
+        const prefix1 = parts[0].toLowerCase();
+        let project = null;
+        let language = null;
+        let title = null;
+
+        const projectMappings = {
```
Abishekcs, line 76: Where this key value pair picked from other part of the codebase or somewhere else ?

app/assets/javascripts/utils/course_utils.js
```diff
@@ -55,8 +56,83 @@ export default class CourseUtils {
...
+          c: 'wikimedia',
+          wmf: 'wikimedia',
+          m: 'wikimedia'
+        };
+
+        const languages = WikiLanguages ? JSON.parse(WikiLanguages) : [];
+        const projects = WikiProjects ? JSON.parse(WikiProjects) : [];
+
+        // Special case for Meta and Commons shorthands
+        if (prefix1 === 'm') {
```
Abishekcs, line 95: Can this `if else` be simplified ?

app/models/wiki.rb
```diff
@@ -108,9 +108,81 @@ def self.language_for_multilingual(language:, project:)
     language
   end
 
-  ####################
-  # Instance methods #
-  ####################
+  INTERWIKI_PREFIXES = {
```
Abishekcs, line 88: Where this key value pair picked from other part of the codebase or somewhere else ?

app/models/wiki.rb
```diff
@@ -108,9 +108,81 @@ def self.language_for_multilingual(language:, project:)
...
+    # Case 2: project shorthand or project name.
+    parse_project_prefix(prefix1, rest)
+  end
+
+  def self.parse_language_prefix(prefix1, rest)
+    # We exclude project names here so that "Wikipedia:Foo" isn't treated
+    # as an interwiki link with language "wikipedia".
+    return [rest, 'wikipedia', prefix1] if Wiki::LANGUAGES.include?(prefix1) && !Wiki::PROJECTS.include?(prefix1)
+  end
+  private_class_method :parse_language_prefix
```
Abishekcs: Won't using only private work ?

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
```diff
@@ -256,14 +257,14 @@ const AssignButton = ({ course, role, course_id, wikidataLabels = {}, hideAssign
...
+    setProject(articleProject || course.home_wiki.project);
+    setLanguage(articleLanguage || (course.home_wiki.language || 'www'));
```
Abishekcs, line 260: What is the purpose of setting this to www ?

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
```diff
@@ -256,14 +257,14 @@ const AssignButton = ({ course, role, course_id, wikidataLabels = {}, hideAssign
...
-      articlesTitles.push(article.title);
+      articlesTitles.push(article.article_url ? article.title : articleTitle);
       articleLanguage = article.language;
       articleProject = article.project;
     });
 
-    setTitle(articlesTitles.join('\n'));
-    setProject(articleProject || project);
-    setLanguage(articleLanguage || language);
+    setTitle(text);
```
Abishekcs, line 258: Why change it to text  ?

app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
```diff
@@ -10,11 +10,11 @@ export const TrainingModuleRows = ({ trainings }) => {
   return trainings.map((trainingModule) => {
     // due_date is not defined for the user profile page. This prevents the error from toDate if due_date is not defined
     const dueDate = toDate(trainingModule.due_date, !trainingModule.due_date);
-    const dueDateFormatted = format(dueDate, 'MMM do, yyyy');
-    const overdue = trainingModule.overdue || isBefore(
+    const dueDateFormatted = isValid(dueDate) ? format(dueDate, 'MMM do, yyyy') : null;
```
Abishekcs: What is the connection between Fixing interwiki assignments and user links. And the changes made in trainingModuleRows.jsx

app/assets/javascripts/reducers/revisions.js
```diff
@@ -42,16 +42,20 @@ const initialState = {
 };
 
 const isLimitReached = (course_start, last_date) => {
-  return isAfter(toDate(course_start), toDate(last_date)) || isBefore(toDate(last_date), subYears(new Date(), 5));
+  const lastDate = toDate(last_date, true);
```
Abishekcs: What is the connection between Fixing interwiki assignments and user links. And the changes made in revisions.js

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
nexpectArpit: The fix addresses a problem that arises when users mistakenly input raw URLs instead of the expected titles. The previous approach, which involved splitting the input on newlines using `articlesTitles.join('\n')` and then updating the state, sometimes altered the user's original format, leading to unwanted duplicate extensions. Now, by directly assigning the title state to the raw `e.target.value` from the text, we maintain the precise formatting of the user's input, whether typed or pasted, within the textarea. This change doesn't disrupt the underlying parsing logic.

app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
nexpectArpit: My mistake. While working on this issue, some commits got intermixed during a rebase on my side. I will revert this change so the PR stays focused on the intended fix.

app/assets/javascripts/reducers/revisions.js
nexpectArpit: same as above one, thanks for looking this

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
nexpectArpit: `www` acts as a necessary language fallback for multilingual projects that don't have a specific language code. For example, Wikidata and the multilingual version of Wikisource use www.wikidata.org and www.wikisource.org. If an assignment belongs to one of these projects and has no language, defaulting to www ensures the URL constructs correctly rather than resolving to undefined.wikidata.org or throwing an error.

app/models/wiki.rb
nexpectArpit: i think In Ruby, the standard private keyword only applies to instance methods. Because 
`parse_language_prefix`  is a class method (defined as `self.parse_language_prefix`), using the normal private keyword will silently fail to make it private. Calling `private_class_method` is the strictly correct Ruby syntax to make a class method private. am i thinking right?

app/models/wiki.rb
nexpectArpit: These are the standard Wikimedia interwiki prefixes used globally across Wikipedia to link to sister projects (e.g., w: for Wikipedia, q: for Wikiquote). They aren't picked from elsewhere in the codebase, but rather match the official Wikipedia definitions. I created the JS version (`projectMappings`) to mirror the Ruby version (`INTERWIKI_PREFIXES`) so that the parsing logic is perfectly consistent on both the frontend and backend.

app/assets/javascripts/utils/course_utils.js
nexpectArpit: I will refactor this block in updated pr

app/models/wiki.rb
Abishekcs: Okay. Then I think you can group private class method in a group.

Example:

```ruby
class Wiki

  def self.parse_interwiki_format
    parse_project_prefix
    parse_language_prefix
  end

  class << self

    private

      def parse_project_prefix
        puts 'parse_project_prefix'
      end
     
     def parse_language_prefix
       puts 'parse_language_prefix'
     end
  end
end

```

Abishekcs: Makes it's much easier to know which are the `Private Class Method`

nexpectArpit: i have refactored the code

nexpectArpit:
updated the pr

app/assets/javascripts/components/students/shared/StudentList/StudentDrawer/TrainingStatus/TrainingModuleRows.jsx
Abishekcs: Where this Files removed ?

app/assets/javascripts/reducers/revisions.js
Abishekcs: Same for this

Abishekcs:
What about `TrainingModuleRows.jsx` and `revisions.js` ?

nexpectArpit:
> What about `TrainingModuleRows.jsx` and `revisions.js` ?

i had things reverted in my local earlier , but somehow not pushed here, it should reflect now(the changes)

app/assets/javascripts/components/common/AssignCell/AssignButton.jsx
Abishekcs, line 258: I still have doubts about this. Because If I give input as `https://en.wikipedia.org/wiki/Dota_2` it will remain same from what I can say from looking at the code.

Whereas before this changes if the input was given `https://en.wikipedia.org/wiki/Dota_2` it became `Dota 2`

Abishekcs: Can you please cross check this ?

nexpectArpit: May you see now!

Abishekcs:
Thanks. I will have a look within the next few days.

ragesoss (changes requested):
Thanks for the thorough reworking and the new `parse_interwiki_format` specs.

**Bug: `feedback.jsx` sandbox URL is always `www.undefined.org`**

`app/assets/javascripts/components/common/feedback.jsx` reads `assignment.wiki`, but the assignments JSON builder (`app/views/assignments/_assignment.json.jbuilder`) emits `assignment.language` and `assignment.project` as top-level fields (and only when the assignment is on a non-home wiki). `toWikiDomain(assignment.wiki || {})` therefore always receives `{}`, producing `"www.undefined.org"` — both cross-wiki and same-wiki sandbox links are broken. The shape used in `assignment.jsx` (`assignment.language || course.home_wiki.language` with a similar project fallback) is the right pattern to mirror here; `course` or its home wiki will need to be threaded into `Feedback`.

**Duplicated interwiki parsing**

`Wiki.parse_interwiki_format` is invoked in both `AssignmentsController#set_new_assignment` and `AssignmentManager#set_clean_title`. When the controller path is hit, the manager re-parses an already-stripped title and no-ops — harmless but redundant. Consolidating in `AssignmentManager` would be cleaner; the `rescue Wiki::InvalidWikiError` in the controller could move up to the `create` action's existing rescue chain.

**Unreachable entries in `INTERWIKI_PREFIXES`**

`'m' => 'wikimedia'` and `'c' => 'wikimedia'` are never read, since `m:` and `c:` are special-cased earlier in `parse_interwiki_format`. `'wmf'` is reachable but currently routes through the `project == 'wikimedia'` fallback returning `language: nil`, which will fail `ensure_valid_project` validation.

**Collisions between `LANGUAGES` and project shorthands**

`Wiki::LANGUAGES` includes `w`, `commons`, `meta`, `incubator`, `minnan`, `simple`, `wikipedia`, `wiktionary`. Since `parse_language_prefix` runs first, `w:Foo` resolves to `['Foo', 'wikipedia', 'w']` (language `w`) rather than the `w:` project shorthand, and `commons:Foo` becomes `['Foo', 'wikipedia', 'commons']`. `Wikipedia:Foo` is correctly excluded via `!Wiki::PROJECTS.include?(prefix1)`. This is a pre-existing quirk in the LANGUAGES list but is newly surfaced by the interwiki logic.

**`Assignment#set_defaults_and_normalize` runs on every save**

The new `CGI.unescape` branch lives in a `before_validation` callback, so it runs on every save, not just for new records. Low practical risk since MediaWiki titles are already decoded, but gating on `new_record?` (or moving the logic into the controller/manager) would be safer.

**Spec gaps**

- No controller/request spec exercises the new `set_new_assignment` interwiki parsing path or the `Wiki::InvalidWikiError` rescue (e.g. POSTing `title: "en:Article"` or `title: "notalang:Foo"`).
- The JS `articleFromTitleInput` interwiki branches mirror the Ruby `parse_interwiki_format` but aren't covered by `test/utils/course_utils.spec.js`; regressions in the JS path would be invisible.
- The `Assignment#set_defaults_and_normalize` CGI.unescape branch is untested.

Verified locally: `wiki_spec.rb` (46 ex), `assignments_controller_spec.rb` + assignment lib/model specs (82 ex), `test/utils/course_utils.spec.js` (38 ex), and `multiwiki_assignment_spec.rb`/`instructor_role_spec.rb` all pass; `yarn build`, RuboCop, and ESLint are clean on the changed files.

*Drafted in a Claude Code session (~20 minutes, ~15 interactions). Sage directed the review, confirmed the findings, and reviewed the comment before posting. The `feedback.jsx` sandbox bug was traced from JS → jbuilder → `toWikiDomain` but not reproduced in a browser; the `LANGUAGES`/shorthand collision was not exercised with a failing test.*

nexpectArpit:
@Abishekcs @ragesoss please take  a look

ragesoss (approved):
(Superseded by a follow-up review from the same session; see the most recent approving review for the full content.)

ragesoss (approved):
(Superseded by a follow-up review from the same session; see the most recent approving review for the full content.)

ragesoss (approved):
The fix addresses the original issue (en.wikipedia article assigned to a non-English course parsing as an interwiki link rather than a homewiki title with a colon) and the substantive concerns raised in the Copilot review pass appear resolved in the current commits.

Verified during a Claude Code session against the merged PR branch:

- `Wiki.parse_interwiki_format` correctly handles `en:Category:Physics` (preserves the `Category:` namespace), rejects `Wikipedia:Help` as not an interwiki link, strips a leading colon, and resolves `wikt:fr:Word` and `m:`/`c:` shorthands.
- `+` in plain titles like `C++` is preserved through both `Assignment#set_defaults_and_normalize` and `AssignmentManager#set_clean_title` (the `gsub('+', '%2B')` guard before `CGI.unescape` does its job).
- `AssignmentsController#create` now rescues `Wiki::InvalidWikiError` and returns 422 rather than a 500.
- The JS interwiki branch was rewritten from regex to split-based parsing; the 2–3-letter regex limitation Copilot flagged is gone, and underscore normalisation is now applied to interwiki titles.

Six exploratory feature specs were added during review (typing `wikt:fr:Bonjour`, `de:Berlin`, `:en:Robot selfie`, `C++`, `m:Main_Page` into the assignment input, plus the new per-assignment user link pointing at the assignment wiki rather than the course home wiki). All six pass against the merged PR branch, and the existing `multiwiki_assignment` and `instructor_role` feature specs continue to pass with no regressions.

A few minor edge cases worth flagging — none blocking:

- `commons:File:Foo` and `meta:Foo` (long-form prefixes) fall through to `parse_language_prefix` because `commons` and `meta` are valid language codes, producing a non-existent wiki like `commons.wikipedia.org`. The new `InvalidWikiError` rescue catches this gracefully (422 instead of 500), and the wiki_spec documents the `commons:` case as expected behavior, so this is acceptable to live with — `c:` and `m:` are the more conventional shortcuts.
- `Actions.jsx` declares `course: PropTypes.object` (not required) but `Feedback.propTypes` declares it required. Minor inconsistency.
- The username link in `assignment.jsx` now points to the wiki where the assignment lives rather than the course home wiki. This is consistent with the spirit of the issue (foreign-wiki assignments should link to the foreign wiki), but it is a behavior change worth being aware of.

PR #6678, opened 2026-02-13
Support multi-wiki article quality ratings 

### LINKED ISSUE #1851: Extend article rating support to additional wikis
The PageAssessements API of Wikipedia is now enabled on Arabic and Turkish Wikipedias ([with more to come](https://phabricator.wikimedia.org/T120219)). RatingImporter should add support for those wikis.

This will probably make sense to implement after #1834 is complete.

nexpectArpit:
- Refactor RatingImporter to support on-demand rating updates for any wiki
- Expand ArticleHelper.default_class with international rating mappings (French, Turkish, Hungarian, Arabic, Russian)
- Add rating_class to JSON API views for standardized CSS styling
- Update React components to use rating_class for badge colors and tooltips
- Fix tooltip i18n lookups to use standardized rating keys

Closes #1851

### What this PR does
Previously, ratings were only fetched and displayed for English Wikipedia. This PR adds end-to-end support for French, Turkish, Hungarian, Arabic, and Russian wikis.

### AI usage
Used AI assistance to research the article quality rating systems across different Wikipedia languages (understanding what adq, teljes, م.مخ, etc. map to in each wiki's assessment scale) and to identify/fix the tooltip hover message issue where international rating keys were causing missing i18n translation errors.

### Before(Screenshot):
Screenshot 2026-02-14 at 12 58 23 AM

### After( Updated Screenshots):

- Articles edited
Screenshot 2026-03-23 at 3 53 45 AM

- Assigned Articles
Screenshot 2026-03-23 at 3 53 54 AM

- Find Articles
Screenshot 2026-03-23 at 3 51 04 AM

#### Learning:
i got to know about how the rating system works in wikimedia articles and their way to rate the articles

@ragesoss  @gabina @Abishekcs Please review it.

### DIFF
--- a/app/assets/javascripts/components/articles/available_article.jsx
+++ b/app/assets/javascripts/components/articles/available_article.jsx
@@ -91,7 +91,7 @@ export const AvailableArticle = ({ assignment, current_user, course, selectable 
         {isWikipedia && (
           <div className="tooltip dark">
             <p>
-              {I18n.t(`articles.rating_docs.${assignment.article_rating || '?'}`, { class: article.pretty_rating || assignment.article_rating || '' })}
+              {I18n.t(`articles.rating_docs.${assignment.article_rating || '?'}`, { class: assignment.article_rating || '' })}
             </p>
           </div>
         )}
--- a/app/assets/javascripts/components/assignments/assignment.jsx
+++ b/app/assets/javascripts/components/assignments/assignment.jsx
@@ -59,7 +59,7 @@ const Assignment = (props) => {
           {isWikipedia && <p className="rating_num hidden">{article.rating_num}</p>}
           {isWikipedia && <div className={ratingClass}><p>{article.pretty_rating || '-'}</p></div>}
           {isWikipedia && <div className="tooltip dark">
-            <p>{I18n.t(`articles.rating_docs.${article.rating || '?'}`, { class: article.pretty_rating || article.rating || '' })}</p>
+            <p>{I18n.t(`articles.rating_docs.${article.rating || '?'}`, { class: article.rating || '' })}</p>
             {/* eslint-disable-next-line */}
           </div>}
         </td>
--- a/app/assets/javascripts/utils/course_utils.js
+++ b/app/assets/javascripts/utils/course_utils.js
@@ -94,16 +94,16 @@ export default class CourseUtils {
 
     const indexphpFormatUrlParts = /([a-z-]+)\.(?:m\.)?(wik[a-z]+)\.org\/w\/index\.php\?title=([\w%]*)[^a-zA-Z0-9%](?:[^#]*)/.exec(articleTitle);
     if (indexphpFormatUrlParts) {
-      const title = decodeURIComponent(indexphpFormatUrlParts[3]).replace(/_/g, ' ');
-      const project = indexphpFormatUrlParts[2];
-      const language = indexphpFormatUrlParts[1];
-      return {
-        title,
-        project,
-        language,
-        article_url: articleTitle,
-      };
-    }
+        const title = decodeURIComponent(indexphpFormatUrlParts[3]).replace(/_/g, ' ');
+        const project = indexphpFormatUrlParts[2];
+        const language = indexphpFormatUrlParts[1];
+        return {
+          title,
+          project,
+          language,
+           article_url: articleTitle,
+        };
+      }
 
     return {
       title: articleTitleInput,


app/assets/javascripts/components/assignments/assignment.jsx
```diff
@@ -20,7 +20,7 @@ const Assignment = (props) => {
       article.formatted_title = CourseUtils.formattedArticleTitle(article, props.course.home_wiki, props.wikidataLabel);
     }
     const className = 'assignment';
-    const ratingClass = `rating ${article.rating}`;
+    const ratingClass = `rating ${article.rating_class || article.rating}`;
```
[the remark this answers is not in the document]
nexpectArpit: Updated to extract baseRating first and fall back to just 'rating' when both rating_class and rating are null/undefined

app/assets/javascripts/components/articles/available_article.jsx
```diff
@@ -49,7 +49,7 @@ export const AvailableArticle = ({ assignment, current_user, course, selectable
   const className = 'assignment';
 
   const article = CourseUtils.articleFromAssignment(assignment, course.home_wiki);
-  const ratingClass = `rating ${assignment.article_rating}`;
+  const ratingClass = `rating ${article.rating_class || assignment.article_rating}`;
```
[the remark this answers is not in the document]
nexpectArpit: fixed

app/helpers/article_helper.rb
```diff
@@ -39,13 +39,32 @@ def rating_display(rating)
...
+
   def default_class(rating)
-    # Handles the different article classes and returns a known article class
     return rating if %w[fa fl a ga b c start stub list].include? rating
-    return 'b' if rating.eql? 'bplus'
-    return 'a' if rating.eql? 'a/ga'
-    return 'list' if %w[al bl cl sl].include? rating
-    return nil
+    INTERNATIONAL_RATING_MAP[rating]
   end
```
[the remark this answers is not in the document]
nexpectArpit: Added unit tests for default_class

lib/importers/rating_importer.rb
```diff
@@ -37,7 +37,8 @@ def self.update_new_ratings
   end
 
   def self.update_rating_for_article(article)
-    return unless article.wiki_id == en_wiki.id # English Wikipedia only, see above.
+    # We support multi-wiki updates on-demand, but bulk updates are still
+    # limited to English Wikipedia for performance reasons.
```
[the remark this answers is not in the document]
nexpectArpit: implemented

gabina:
Hello! I took a first look at the PR. Thanks for working on this! I have a couple of questions.

The description says:

> Refactor `RatingImporter` to support on-demand rating updates for any wiki
> 
> This PR adds end-to-end support for French, Turkish, Hungarian, Arabic, and Russian wikis.

On the other side, the AI description says:

> This pull request extends article quality rating support from English Wikipedia to five additional language wikis: French, Turkish, Hungarian, Arabic, and Russian. The changes enable on-demand rating updates for any wiki while maintaining bulk update restrictions to English Wikipedia for performance reasons.

I don't think I fully understand what "on-demand" means in this context.

Is there any place where users can request adding rating scores for a specific wiki language, and it's expected that we add the wikis from which we import ratings based on that?
Or are we saying "on-demand" in the sense that a user manually creates an assignment?

Also, from the PR description, it's not clear to me whether this applies to _any wiki_ or only to the _specific five wikis mentioned_ (French, Turkish, Hungarian, Arabic, and Russian).

Based on what I see in the code, no wiki languages are hardcoded, so it looks like we try to fetch ratings for every article for which the user created an assignment, regardless of whether the wiki language is supported. Is that correct?

If that's the case, I'm not sure whether making an API call for a non-supported language makes sense. I understand that the rating import is only triggered when a user creates an assignment, so we don't have many requests per minute, but I'm still curious about your intention here.

nexpectArpit:
@gabina Thanks for review.



> Also, from the PR description, it's not clear to me whether this applies to _any wiki_ or only to the _specific five wikis mentioned_ (French, Turkish, Hungarian, Arabic, and Russian).

The description says 'any wiki' because the backend architecture was Generalized to allow multi-wiki imports. The AI mentions '5 wikis' because those are the ones where i completed the visual mapping. The description mismatch shows the shift from an English-only system to a multi-wiki capable system.

> I don't think I fully understand what "on-demand" means in this context.
> Is there any place where users can request adding rating scores for a specific wiki language, and it's expected that we add the wikis from which we import ratings based on that?
> Or are we saying "on-demand" in the sense that a user manually creates an assignment?

'On-demand' refers to rating updates triggered when a user manually creates an assignment or edit. Unlike the periodic bulk background jobs (which remain restricted to English Wikipedia for performance), this allows assigned articles on other wikis to get their ratings immediately.


> Based on what I see in the code, no wiki languages are hardcoded

[https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/lib/importers/rating_importer.rb#L39](link) here we can see that it is hardcoded to english only, 
and you are right, In this PR, I refactored the backend architecture to be generic so it can support any wiki, but the visual mappings (in ArticleHelper) are indeed currently limited to the specific five wikis mentioned.

Actually my goal was to move away from the hardcoded English-only check in `RatingImporter` (which originally blocked everything but en.wiki). 

However, I see your point about making redundant API calls for wikis we don't 'officially' support in the UI yet. for this I can add codes to restricts these on-demand API calls to only the wikis we have mappings for. Any other feedback?

gabina:
Thanks for the clarification!

> However, I see your point about making redundant API calls for wikis we don't 'officially' support in the UI yet. for this I can add codes to restricts these on-demand API calls to only the wikis we have mappings for.

Ok, so according to my understanding, there are two different things here:

- The **visual mappings** (in ArticleHelper). These are currently limited to the five specific wikis mentioned: French, Turkish, Hungarian, Arabic, and Russian. How did you choose those?
- The **PageAssessments API** of Wikipedia, which supports only a subset of projects/languages. Do you know which subset of projects/languages is supported? The [xtools assessments](https://xtools.wmflabs.org/api/project/assessments) endpoint mentions the following:
"ar.wikipedia.org", "en.wikipedia.org", "en.wikivoyage.org", "hu.wikipedia.org", "fr.wikipedia.org", "tr.wikipedia.org", "zh.wikipedia.org".
I’m not sure whether those are the only supported ones.

I guess the most thorough way to handle this would be to maintain an array of wiki languages currently supported by the PageAssessments API, define the visual mappings only for those, and make API calls to request ratings only for them.

Does that make sense to you?

lib/importers/rating_importer.rb
```diff
@@ -37,7 +37,15 @@ def self.update_new_ratings
...
+    # To avoid excessive load from frequent on-demand updates on non-English
+    # wikis, skip the update if the rating was refreshed very recently.
+    if article.wiki_id != en_wiki.id &&
+       article.rating_updated_at.present? &&
+       article.rating_updated_at > 5.minutes.ago
+      return
```
gabina: Why does `rating_updated_at > 5.minutes.ago` only apply to non-English Wikipedia articles?
In addition, I think the minimum amount of time should be a constant, similar to things like `RECENT_ALERTS_DAYS`.

spec/helpers/article_helper_spec.rb
```diff
@@ -45,4 +45,38 @@
...
+
+    it 'maps Turkish ratings correctly' do
+      expect(default_class('sm')).to eq('fa')
+      expect(default_class('km')).to eq('ga')
+      expect(default_class('taslak')).to eq('stub')
+    end
+
+    it 'maps Hungarian ratings correctly' do
+      expect(default_class('teljes')).to eq('b')
+    end
```
gabina, line 75: It looks like the mapping covers French, Turkish, Hungarian, Arabic, and Russian wikis, but the specs only mention French, Turkish and Hungarian. Are the other languages covered by another spec?

nexpectArpit:
> * The **visual mappings** (in ArticleHelper). These are currently limited to the five specific wikis mentioned: French, Turkish, Hungarian, Arabic, and Russian. How did you choose those?

I chose them because the codebase already had rating definitions for these wikis in [weighted_score_calculator.rb](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/lib/weighted_score_calculator.rb#L29-L47). French, Turkish, and Russian rating strings were already defined there for article quality scoring. I extended those same mappings to `ArticleHelper` for visual display. Hungarian and Arabic were included based on the original issue description, which specifically mentioned Arabic and Turkish as the first wikis where the PageAssessments API was enabled.


And 



> * The **PageAssessments API** of Wikipedia, which supports only a subset of projects/languages. Do you know which subset of projects/languages is supported? The [xtools assessments](https://xtools.wmflabs.org/api/project/assessments) endpoint mentions the following:
>   "ar.wikipedia.org", "en.wikipedia.org", "en.wikivoyage.org", "hu.wikipedia.org", "fr.wikipedia.org", "tr.wikipedia.org", "zh.wikipedia.org".
>   I’m not sure whether those are the only supported ones.


that's a very helpful insight. I did some investigative API calls and you're absolutely right, the `pageassessments `extension is not active on all wikis. For ex: while Russian Wikipedia (ru) has an established rating system, it doesn't currently support the pageassessments API property, so my initial inclusion of it was optimistic but technically incorrect for this specific API. And
>maintain an array of wiki languages currently supported by the PageAssessments API

Your suggestion of a whitelist is definitely the most applicable approach.

lib/importers/rating_importer.rb
nexpectArpit: To preserve the legacy behavior of English Wikipedia to avoid any risk of changing how the core wiki handles updates. And I completely agree with the "minimum amount of time should be a constant" part.

spec/helpers/article_helper_spec.rb
nexpectArpit: No, they aren't covered by another spec .Actually I only tested a few languages since the mapping is just a hash lookup and I figured testing a handful would be enough to prove it works. But yeah, makes sense to add specs for all of them so nothing slips through if someone changes the map later. I'll include that in my next update.

lib/importers/rating_importer.rb
nexpectArpit: I will update my PR to add this part.

nexpectArpit:
@gabina i updated the pr, have a look

gabina:
I think this PR is heading in the right direction!

While reviewing it, I found bug #6693 (not related to your changes). I’ve already opened PR #6694 to fix it. Please take a look at them when you have a chance.

While checking your changes on files `app/views/assignments/_assignment.json.jbuilder` and `app/views/courses/articles.json.jbuilder`, I wasn’t sure why we need to send both the raw article and the default class article to the frontend. That’s how I ended up discovering the issue mentioned above.

I think the best approach would be for you to build your PR on top of those changes. If you don’t want to wait for PR #6694 to be merged into master, you could manually cherry-pick its commits. I believe that would make your frontend changes in this PR much simpler.

gabina:
Something worth noting is that it looks like there is already support for quality classes in "Recent Activity", for [example](https://outreachdashboard.wmflabs.org/courses/les_sans_pagEs/S%C3%A9ance_%C3%A9valuation_LSP_2024-06-06_(jeudi_6_juin_2024)/activity/recent) for French. 

The mapping is in `app/assets/javascripts/utils/article_finder_language_mappings.js`

image

nexpectArpit:
> 
> I think the best approach would be for you to build your PR on top of those changes. If you don’t want to wait for PR #6694 to be merged into master, you could manually cherry-pick its commits. I believe that would make your frontend changes in this PR much simpler.


i saw your pr #6694 which suggest a cleaner approach for translation bug since it fixes the output right at the source in the jbuilder.  I'll wait for  your pr to get merged and then will rebase my branch on top of it. That way my PR stays focused on the international rating mapping in default_class, and your fix handles the jbuilder side.




> Something worth noting is that it looks like there is already support for quality classes in "Recent Activity", for [example](https://outreachdashboard.wmflabs.org/courses/les_sans_pagEs/S%C3%A9ance_%C3%A9valuation_LSP_2024-06-06_(jeudi_6_juin_2024)/activity/recent) for French.

Also noticed the French quality class support already working in Recent Activity via `article_finder_language_mappings.js` is a worth point. My backend `INTERNATIONAL_RATING_MAP` should complement that well for the server-rendered views.
Will update my PR once yours merge.

ragesoss:
@nexpectArpit #6694 is merged now.

nexpectArpit:
@ragesoss can you check the updated PR now!

app/helpers/article_helper.rb
```diff
@@ -39,13 +39,40 @@ def rating_display(rating)
     return rating[0] # use the first letter of the rating as the abbreviated version
   end
 
+  # Maps international rating strings to standardized article classes.
+  # Covers wikis with confirmed PageAssessments API support:
+  # Arabic, Chinese, French, Hungarian, Turkish.
+  INTERNATIONAL_RATING_MAP = {
```
ragesoss, line 51: The comment should include links to where these mappings come from, so that it will be easy for others to update them in the future.

app/assets/javascripts/utils/course_utils.js
```diff
@@ -122,6 +122,7 @@ export default class CourseUtils {
       rating: assignment.article_rating,
       rating_num: assignment.article_rating_num,
       pretty_rating: assignment.article_pretty_rating,
+      rating_class: assignment.article_rating_class,
```
gabina: I think I'm missing something here, what's `assignment.article_rating_class`?

nexpectArpit: @gabina  I've updated the PR to include the source mapping links in  `article_helper.rb` for the international ratings. While doing that, I also did a broad-level cleanup to remove all the obsolete `rating_class` and `article_rating_class` variables from both the Jbuilder files and the React components (`article.jsx` , `assignment.jsx` , etc.), since the API doesn't use them anymore.

nexpectArpit:
rebased with latest ruby version

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -24,7 +24,8 @@ const Article = ({ article, index, course, fetchArticleDetails, updateArticleTra
     setTracked(checked);
   };
 
-  const ratingClass = `rating ${article.rating}`;
+  const baseRating = article.rating;
+  const ratingClass = baseRating ? `rating ${baseRating}` : 'rating';
```
gabina: What is the purpose of this change?
There is already a class 'rating null' defined in `app/assets/stylesheets/modules/_course.styl`.

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -79,7 +80,7 @@ const Article = ({ article, index, course, fetchArticleDetails, updateArticleTra
         {isWikipedia && <div className="tooltip dark">
           <p>
             {
-              !isDeleted ? I18n.t(`articles.rating_docs.${article.rating || '?'}`, { class: article.rating || '' }) : deletedMessage
+              !isDeleted ? I18n.t(`articles.rating_docs.${article.rating || '?'}`, { class: article.pretty_rating || article.rating || '' }) : deletedMessage
```
gabina: I think it's better to keep the raw rating instead of the "pretty" version for the tooltip explanation, since the pretty version is only the first char for some rating class. That way we can differentiate between "start" and "stub", for example.

Stub:
Image


Start:
Image

app/assets/javascripts/components/articles/article.jsx
nexpectArpit: The original goal for that change was  to modernize the template string and avoid an ambiguous 'null' class in the DOM, which can sometimes be a sign of a missing data check. However, after your mention i saw that project's specific styling in `_course.styl ` has `.rating.null` which is a deliberate design pattern used for unrated articles . after your mention i think i need to go through the codebase again. Thanks

app/assets/javascripts/components/articles/article.jsx
nexpectArpit: i will add this point in latest pr,any other feedback!

app/assets/javascripts/components/articles/article.jsx
gabina: Note that the same point applies for the other files: `app/assets/javascripts/components/articles/available_article.jsx` and `app/assets/javascripts/components/assignments/assignment.jsx`

nexpectArpit: I have  reverted the `ratingClass` change in both `available_article.jsx` and `assignment.jsx`.

app/assets/javascripts/components/articles/article.jsx
gabina: Shouldn't this change apply to the other files too? `app/assets/javascripts/components/articles/available_article.jsx` and `app/assets/javascripts/components/assignments/assignment.jsx`? Please review all the files changed before setting the PR as ready for review, and make sure that only the necessary files have been modified.

app/assets/javascripts/utils/course_utils.js
```diff
@@ -94,16 +94,16 @@ export default class CourseUtils {
...
+      const title = decodeURIComponent(indexphpFormatUrlParts[3]).replace(/_/g, ' ');
+      const project = indexphpFormatUrlParts[2];
+      const language = indexphpFormatUrlParts[1];
+      return {
+        title,
+        project,
+        language,
+        article_url: articleTitle,
+      };
+    }
```
gabina: Is this change related to this PR? Are you changing anything other than the indentation?

nexpectArpit: i reverted them, they were unintentional identation changes

app/assets/javascripts/components/articles/article.jsx
nexpectArpit: modified the pr

gabina:
This looks good to me.

I'm still not sure about the need for `RATING_UPDATE_COOLDOWN`, since `update_rating_for_article` is only triggered by `create_assignment`. I don't know how frequent is to create different assignments for the same article in a period of 5 minutes. But I'm OK with that.

@nexpectArpit since there was a lot of back and forth,  I think you should update the screenshots with rating for articles for different wikis in the different sections of the web-app where we show that (not only `/articles/edited?` but also `/articles/assigned`, `/articles/available` and any other route you find). Just to be sure it's working as expected. After that, I think this will be ready to go.

nexpectArpit:
@gabina i updated the screenshots ,have a look. 
also note that While the `Revisions` tabs in my test environment  is currently empty (as there are no live tracked edits for this specific test course), I've verified that they use the same `article.rating` property and standardized badge logic as the Assigned tab. Additionally, the Article Finder results confirm that the cross-wiki assessment mapping is working correctly for live Wikipedia data, not just local database records. So i added the Article Finder Tab Screenshot (in french language wiki articles ).



> I'm still not sure about the need for `RATING_UPDATE_COOLDOWN`, since `update_rating_for_article` is only triggered by `create_assignment`. I don't know how frequent is to create different assignments for the same article in a period of 5 minutes

Agree with you that  it's a pretty narrow edge case since it's only triggered by `create_assignment`. that's why i have  kept it as a lightweight defensive safeguard for the rare scenario where multiple students might assign the same article in quick succession, avoiding redundant API calls.

ragesoss:
This looks good to me as well.

PR #6681, opened 2026-02-15
Fix Timeline Sunday Start bug 

### LINKED ISSUE #2360: When a timeline start date is Sunday, the previous week shows up as an empty week 1

nexpectArpit:
### What this PR does
This PR fixes a bug where a course starting on a Saturday with a timeline starting the following Sunday would display an empty "Week 1" and misnumber all subsequent weeks.

Closes #2360

### What this fixes

1. **Week content**: Empty leading weeks are skipped in `weekMeetings`
2. **Week numbering**:`weeksBeforeTimeline` now calculates offsets from the first real content week
3. **UI consistency** : Timeline and Home tabs show the same week numbers and date ranges
4. **Selectors**: `getAllWeekDates` now aligns date labels with meeting data

### Tests

- Added regression test `repro_2360.spec.js` to cover this scenario

### Screen Recording
Before:

https://github.com/user-attachments/assets/a21d0f78-14a6-43a8-a8f6-15bc79ab0bd6

After:

https://github.com/user-attachments/assets/7c37ec34-8884-49c7-8e05-9e08c151701f

@ragesoss Ready for review!!

### DIFF
--- a/spec/features/category_scopes_spec.rb
+++ b/spec/features/category_scopes_spec.rb
@@ -69,7 +69,7 @@ describe 'Tracked categories and templates', js: true do
     find(:css, '.multi-wiki-selector input').set('fr')
     find(:css, '.multi-wiki-selector div[class*="option"]', text: 'fr.wikipedia.org').click
 
-    find(:css, '#categories input').set('Apple ')
+    find(:css, '#categories input').set('Matériel Apple ')
     find(:css, '#categories div[class*="option"]', text: 'fr:Matériel Apple', exact_text: true).click # rubocop:disable Layout/LineLength
 
     click_button 'Add categories'
@@ -84,7 +84,7 @@ describe 'Tracked categories and templates', js: true do
 
     find(:css, '#templates input').set('Earth ')
     find(:css, '#templates div[class*="option"]', text: 'Earth mass').click
-    find(:css, '#templates input').set('Apple ')
+    find(:css, '#templates input').set('Apple Inc. ')
     find(:css, '#templates div[class*="option"]', text: 'en:Apple Inc.', exact_text: true).click
 
     click_button 'Add Templates'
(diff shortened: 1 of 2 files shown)


test/utils/repro_2360.spec.js
```diff
@@ -0,0 +1,48 @@
+import '../testHelper';
+import CourseDateUtils from '../../app/assets/javascripts/utils/course_date_utils';
+
+describe('Issue #2360: Sunday Timeline Start', () => {
```
ragesoss: To follow the conventions we use throughout the rest of the codebase, the filename and test descriptions should not be specific to an issue number. The normal convention would be to add these tests to the existing spec for CourseDateUtils (`course_date_utils.spec.js`).

ragesoss: You can include a link to the issue in a code comment.

test/utils/repro_2360.spec.js
```diff
@@ -0,0 +1,48 @@
...
+            start: '2026-02-15', // Same Sunday as timeline_start
+        };
+        const meetings = CourseDateUtils.weekMeetings(sameDayCourse, '');
+        const wbt = CourseDateUtils.weeksBeforeTimeline(sameDayCourse);
+
+        expect(wbt).toBe(0);
+        expect(meetings[0].length).toBeGreaterThan(0);
+    });
+
+    test('PROOF 5: What if course starts Saturday, timeline starts Sunday?', () => {
```
ragesoss: Do all these tests fail in the current implementation and pass after this PR? It would be good to specify in comments exactly which of these test represent behavior changes with your implementation of the bugfix.

test/utils/repro_2360.spec.js
nexpectArpit: sure

test/utils/repro_2360.spec.js
nexpectArpit: in latest pr i added the comment block above those test cases

nexpectArpit:
@ragesoss updated the pr.

PR #6614, opened 2026-01-05
Optimize `Generate CSV of all programs` generation

Abishekcs:
#6608

## What this PR does

Improve performance for `/all_courses_csv`

## Performance Comparsion

Request # | Views (ms) | ActiveRecord (ms) | Allocations
-- | -- | -- | --
BEFORE |   |   |  
1 | 1.7 | 866.3 | 668,266
2 | 1.9 | 885.2 | 667,841
3 | 1.6 | 901.3 | 667,781
4 | 1.4 | 882.8 | 667,763
5 | 1.4 | 877.5 | 678,668
6 | 1.7 | 919.2 | 680,436
7 | 1.8 | 875.6 | 678,631
8 | 1.3 | 908.4 | 669,499
9 | 1.3 | 908.4 | 669,499
10 | 1.5 | 958.0 | 683,439
BEFORE Average | 1.56 | 898.27 | 673,182
AFTER |   |   |  
1 | 0.9 | 30.8 | 21,540
2 | 0.8 | 29.5 | 21,378
3 | 0.8 | 30.5 | 21,375
4 | 0.8 | 30.7 | 21,942
5 | 0.7 | 28.9 | 21,914
6 | 0.9 | 29.5 | 21,376
7 | 0.8 | 42.6 | 21,914
8 | 0.8 | 42.4 | 21,913
9 | 1.0 | 31.2 | 21,914
10 | 1.0 | 30.7 | 21,946
AFTER Average | 0.85 | 32.68 | 21,721

</body></html>

## AI usage

No AI usage

## Does this performance optimization also includes removing N+1 ?

- Yes

## Where is the N+1 this fixes, and what do the N+1 queries look like

**a) Where is the N+1 this fixes**  `(course_csv_builder.rb)`

screenshot-2026-02-03_20-02-37

screenshot-2026-02-03_20-04-36

screenshot-2026-02-03_20-10-55

**b) what do the N+1 queries look like**

screenshot-2026-02-03_20-13-33

screenshot-2026-02-03_20-14-09

-  `Query For Three different namespaces i.e (0, 1, 2)`

screenshot-2026-02-03_20-15-19

screenshot-2026-02-03_20-16-07

screenshot-2026-02-03_20-16-33

## Screenshots
Before:

screenshot-2026-01-05_13-44-21

After:

CSV all generation

PR #6683, opened 2026-02-18
Reimplement retained new editors

### LINKED ISSUE #6345: Re-implement `retained_new_editors` column for course csv report
We may want to re-implement `retained_new_editors` column for csv, since it was deleted in second clean-up  PR #6324 (see commit 848748a3920152e045458da90fe53f66accaf01f). In the revisions system (when using the `revisions` table),  we used to keep ingesting revisions from editors after the course end date in order to have data for calculating "retained editors" (see PR #6132). In particular, the `CourseCsvBuilder` class had a `retained_new_editors` column with the following criteria: an editor counts as retained if they make at least one revision 7 or more.

Formasitchijoh:
## What this PR does
This PR re-implements the retained_new_editors metric within the CourseCsvBuilder and closes #6345 

The New Implementation, unlike the previous version does not rely on pre-calculated database revisions. Instead, it interacts directly with the `WikiApi` to track student activity after a course concludes. We now fetch live `usercontribs` from the MediaWiki API to identify edits made in the Mainspace (ns: 0) starting 7 days after the course end date. The column now reflects the distinct count of users who have made at least one qualifying contribution after the threshold, providing a more accurate measure of long-term editor retention.

## AI usage
Used AI for improving the test cases and debugging Issue with mocking wikiapi response data

## Open Questions
I have implemented this as a live API calculation within the CSV generation loop. While this works well for individual courses and small campaigns, it introduces a potential N+1 API bottleneck for large-scale campaigns (e.g., 50+ courses), which could lead to request timeouts or rate-limiting.

Should we consider adding a `retained_edits_count` column to the courses (or course_stats) table?. This would allow us to move the API logic into a background worker/cron job, making CSV generation instant and avoiding live network dependencies during the download request.
Alternatively: If current campaign sizes are generally small, we can keep the live implementation for now and transition to a cached column only if performance becomes an issue.

@gabina Could you take a look at this

### DIFF
--- a/lib/analytics/retained_new_editors_stats.rb
+++ b/lib/analytics/retained_new_editors_stats.rb
@@ -6,24 +6,27 @@ class RetainedNewEditorsStats
     @course = course
   end
 
-  def count
-    return 0 if @course.end.nil? || new_editors.empty?
+  def count 
 
-    threshold = (@course.end + DAYS_AFTER_END.days)
+    return 0 if new_editors.empty?
 
-    retained = 0
+    result = Rails.cache.fetch("retained_new_editors_#{@course.id}", expires_in: 7.days) do
 
-    # Current implementation is limited to home_wiki
-    # does not aggregate across all course wikis
+      retained = 0      
+      threshold = @course.end + DAYS_AFTER_END.days
 
-    wiki = @course.home_wiki || @course.wikis.first
+      # Current implementation is limited to home_wiki
+      # does not aggregate across all course wikis
+      wiki = @course.home_wiki
 
-    new_editors.pluck(:username).in_groups_of(40, false) do |batch|
-      next if batch.empty?
-      retained += count_retained_in_batch(batch, threshold, wiki)
+      new_editors.pluck(:username).in_groups_of(40, false) do |batch|
+        next if batch.empty?
+        retained += count_retained_in_batch(batch, threshold, wiki)
+      end
+      retained
     end
 
-    retained
+    result
   end
 
   private


gabina:
will review next week!

lib/analytics/retained_new_editors_stats.rb
```diff
@@ -0,0 +1,53 @@
+class RetainedNewEditorsStats
+  MINIMUM_EDITS = 1   # could be configurable
+  DAYS_AFTER_END = 7
+
+  def initialize(course)
+    @course = course
+  end
+
+  def count
+    return 0 if @course.end.nil? || new_editors.empty?
```
gabina: We don't have courses with a nil `end` field in our `courses` table. I don't think that's something reasonable to happen. I understand that you added this safety guard to prevent the report from failing if `end` is nil, but my opinion here is that having a course with a null `end` would be so weird that it's better for the report to fail than to just run it as if that were an expected situation.

lib/analytics/retained_new_editors_stats.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    threshold = (@course.end + DAYS_AFTER_END.days)
+
+    retained = 0
+
+    # Current implementation is limited to home_wiki
+    # does not aggregate across all course wikis
+
+    wiki = @course.home_wiki || @course.wikis.first
```
gabina: Similar comment here, we don't have courses without `home_wiki`.

gabina:
> I have implemented this as a live API calculation within the CSV generation loop. While this works well for individual courses and small campaigns, it introduces a potential N+1 API bottleneck for large-scale campaigns (e.g., 50+ courses), which could lead to request timeouts or rate-limiting.

hmm these reports are generated in the background, so it's usually not a big deal if they take some time to run. But performance is a problem. In fact, you can notice that there were some effort recently on improving the performance for the campaign CSV (see PR #6614).

What do you think could be a bottleneck here?

Formasitchijoh:
> What do you think could be a bottleneck here?

The main performance bottleneck is the repeated calls to the Wikipedia API (`list=usercontribs`) when calculating retained new editors.

- For small courses (~40–50 new editors): usually just 1 API call.
- For larger courses (300–600 new editors): 6–12+ API calls (batches of 50 usernames).
- Each call takes ~0.8–1.4 seconds on average, but can occasionally spike much higher (e.g., one batch took 23 seconds in logs due to network latency or Wikipedia-side delays).
- When generating a campaign CSV that includes many courses  especially if several have hundreds of new editors , the total time can easily reach several minutes and sometimes leads to timeouts.

To fix this without adding a new database column (which would bloat the `courses` table with a field that's not part of the core business logic and may not be queried frequently), I went with simple Rails caching:

- The result is cached per course using a key that includes the course ID, end date, and config values (`DAYS_AFTER_END`).
- Using `Rails.cache.fetch` with a 7-day expiration.
- The heavy Wikipedia API requests happen at most once per course (on cache miss).
- All subsequent generations (which are very common for reports) become near-instant.

This way:
- The high cost is amortized , then reused.
- We avoid unnecessary schema changes.

gabina:
LGTM!

I don’t know which kind of cache we use in production, and I’m not sure whether the TTL is longer than seven days. When we explicitly set `expires_in` in other parts of the code, it’s usually less than one week. That said, I’m OK with these changes.

PR #6686, opened 2026-02-19
Improve layout of subject-specific guides block in Timeline

### LINKED ISSUE #5707: Improve layout of "subject specific guides" block in Timeline

AarohiSharma5:
### **What this PR does?**

- Remove "< p >" wrapper from handout links in wizard_timeline_manager.rb
- Add margin-bottom and display inline-block to a.handout-link in _timeline.styl

Fixes #5707

### **Note:**
This works only for the new courses. 

### **Screenshots:**

**Before:**
before

**After:**
after

### DIFF
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -569,10 +569,12 @@ a.training-module
 
 a.handout-link
   margin-right 8px
+  margin-bottom 8px
   color #6a6a6a
   border solid 1px #ddd
   border-radius 3px
   text-decoration none
   padding 5px
+  display inline-block
   &:hover, &:focus
     border solid 1px $brand_primary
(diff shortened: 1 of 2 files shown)


AarohiSharma5:
Hi @ragesoss, can you please review this?

ragesoss:
Thanks @AarohiSharma5!

The key thing I'd like to account for is how this change will affect already-generated timeline blocks.

Can you include before/after screenshots of an example block where it was created with the current `<p>` tag format? If this change doesn't affect the layout of already-existing blocks, and fixes the formatting for newly-created ones, it should be good to go. (Please mark it ready for review once you've added those screenshots.)

AarohiSharma5:
Hi @ragesoss, thank you for the review!
I would like to extend my explanation on this issue. So, the change I made only accounts for the newly created courses. And, it doesn't change anything for the existing courses. So, the before and after screenshots I added earlier are of 2 different courses. The former includes the <p> tag and the later doesn't include it. 
The idea that got me to this solution is :
SCREENSHOT:
BEFORE(code):
code

Issue #6286, opened 2025-04-04
Stats for individual editors are incomplete


ragesoss:
A user is confused as to why the Revisions stats are showing as 0, despite significant edits in their contribution history: https://outreachdashboard.wmflabs.org/courses/CEPID_NeuroMat/Introdu%C3%A7%C3%A3o_ao_Jornalismo_Cient%C3%ADfico/students/articles/PEDRO%20HENRIQUE%20VISENTINI%20PANTAROTTO

In this case, the user has made mainspace edits in pt.wiktionary.org (which is one of the tracked wikis, but no the home wiki), so we should expect the recent edits and mainspace/userspace/draft stats to reflect these edits.

Also, there are no edited articles shown for this user, despite edits made to mainspace pages in pt.wiktionary.

betty-alagwu:
@ragesoss i'll be working on this

gabina:
I am catching up and reading the new issues and PRs. I would like to better understand this.

> A user is confused as to why the Revisions stats are showing as 0, despite significant edits in their contribution history: https://outreachdashboard.wmflabs.org/courses/CEPID_NeuroMat/Introdu%C3%A7%C3%A3o_ao_Jornalismo_Cient%C3%ADfico/students/articles/PEDRO%20HENRIQUE%20VISENTINI%20PANTAROTTO
In this case, the user has made mainspace edits in pt.wiktionary.org (which is one of the tracked wikis, but no the home wiki), so we should expect the recent edits and mainspace/userspace/draft stats to reflect these edits.

Right now I see this, but I assume that when the issue was created "Recent Edits" and "Characters Added (article | user | draft)" were set to 0, even though there were some user contributions listed?

![Image](https://github.com/user-attachments/assets/f4a98abb-a985-4bb4-81ad-598da097e396)

BTW, I don't see `pt.wiktionary.org` as a tracked wiki now, but I guess that it could have been removed recently.


> Also, there are no edited articles shown for this user, despite edits made to mainspace pages in pt.wiktionary.

Would you mind pointing me to where we show the "edited articles" for a user?

ragesoss:
@gabina there's a component for "Other Edited Articles" which should show edited articles for an individual user other than ones that are assigned. For example: https://dashboard.wikiedu.org/courses/Rutgers_Law_School/Advanced_Legal_Research_(Spring_2025)/students/articles/SmallAtLarge

[...]

PR #6691, opened 2026-02-23
Stats for individual editors are incomplete

Formasitchijoh:
## What this PR does
This PR addresses #6286 by ensuring that a student's "Other Edited Articles" are correctly displayed even when a course exceeds the default 500-article limit of articles.json.

**The New Implementation:**
Instead of relying solely on the global course-wide article lis which may be truncated due to pagination, this PR introduces a targeted fetch for individual students.

**Backend:** Updated the articles controller and ArticlesCourses model to support a user_id parameter, allowing the API to return all articles edited by a specific student regardless of the global limit.
**Frontend:** Added a studentArticles key to the Redux state to isolate student-specific data and avoid disrupting global article filters or pagination.
**Logic:** The SelectedStudent component now detects if a student has recorded edits (via character_sum stats) but no articles in the current local store. If this gap exists, it triggers a background fetch to pull the missing article metadata directly.
This ensures that "Recent Edits" and "Mainspace/Userspace/Draft" statistics always have matching article links in the UI, even in massive courses.

## AI usage
Used AI to  refine the useEffect logic to prevent infinite request loops during the targeted data fetch.

## Screenshots
Before:

https://github.com/user-attachments/assets/5001dea0-26b7-434b-98aa-a239bc7b4ac1

After:

https://github.com/user-attachments/assets/8818f7c8-be0e-4f2d-83c0-20499a70c1ba

### DIFF
--- a/app/assets/javascripts/components/students/components/Articles/SelectedStudent/EditedUnassignedArticles/EditedUnassignedArticles.jsx
+++ b/app/assets/javascripts/components/students/components/Articles/SelectedStudent/EditedUnassignedArticles/EditedUnassignedArticles.jsx
@@ -52,6 +52,7 @@ const EditedUnassignedArticles = ({
         table_key="users"
         stickyHeader={false}
         sortable={false}
+        none_message={I18n.t('users.more_articles')}
       />
       <div className="see-more">
         {!props.limitReached
--- a/app/assets/javascripts/components/students/components/Articles/SelectedStudent/SelectedStudent.jsx
+++ b/app/assets/javascripts/components/students/components/Articles/SelectedStudent/SelectedStudent.jsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
+import { useLocation, useParams, Navigate } from 'react-router-dom';
+import { useSelector } from 'react-redux';
 
 // Components
 import Header from './Header.jsx';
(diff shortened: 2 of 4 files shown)


app/controllers/courses_controller.rb
```diff
@@ -103,7 +103,15 @@ def course
 
   def articles
     set_course
-    set_limit
+    if params[:user_id].present?
+      @articles_courses = ArticlesCourses.search_by_course_and_user(@course, params[:user_id])
```
ragesoss: I think this is a major performance risk. The 500-article limit for the whole-course article fetching was necessary to prevent courses with huge numbers of edited articles from causing problems. (These include both backend performance impacts from the article queries, and frontend impacts from trying to render a very large number of articles in a React table.) In cases like that, it's common for a single user to have the majority of them.

Formasitchijoh: That’s a fair point. I hadn't considered cases where a single user's edits might exceed that 500-article threshold.
To avoid the performance risk while still fixing the issue where a student's edits aren't visible, I’ll take a different approach:

**Detection:** I'll check the student's character_sum. If it's non-zero but the otherEditedArticles list is empty, it indicates their edits haven't been fetched in the initial 500-article course batch.

**UI Action:** Instead of an automatic fetch of all their data, I'll display a message: 'This student has edits that are not yet loaded.'

**Consistency:**  I'll provide a button that triggers the existing fetchArticles logic to pull the next batch of 500.
This ensures we stay within the performance limits and only pull more data if the instructor explicitly chooses to 'See More' for that specific student. Does that sound like a safer middle ground?

Formasitchijoh: @ragesoss can you tell me what you think of this approach

ragesoss: Sounds good.

Formasitchijoh:
@ragesoss  could you take a look at this

ragesoss:
Looks reasonable to me.

PR #6695, opened 2026-02-25
Fix: Explain Error message from course slug collision

### LINKED ISSUE #6087: Error message from course slug collision does not explain the problem

Formasitchijoh:
## What this PR does
This PR resolves #6087 and implements feedback from #6290 by rescuing database-level uniqueness errors at the controller level. By using rescue_from `ActiveRecord::RecordNotUnique`, the implementation ensures that slug collisions are caught during both create and update actions without requiring extra database queries for validations.

When a collision occurs, the controller now returns a 409 Conflict status and a clear, internationalized error message. This replaces generic "Internal Server Error" responses and provides users with the specific conflicting slug sourced directly from the @course object so they can resolve the naming conflict by adjusting the title, school, or term

## AI usage
Used AI to optimize the rescue_from architecture and verify that @course.slug remains accessible for error reporting after a failed database transaction. And improve this PR message

## Screenshots

https://github.com/user-attachments/assets/dc73bdba-ca8a-42ab-848b-84b1c938480a

Screenshot 2026-03-17 at 17 53 08

### DIFF
--- a/app/assets/javascripts/utils/api.js
+++ b/app/assets/javascripts/utils/api.js
@@ -298,10 +298,14 @@ const API = {
       console.error('Couldn\'t save timeline!');
       SentryLogger.obj = this.obj;
       SentryLogger.status = this.status;
-      Sentry.captureMessage('saveTimeline failed', {
-        level: 'error',
-        extra: SentryLogger
-      });
+
+      // Ensure Sentry is defined to prevent failing silently
+      if (typeof Sentry !== 'undefined') {
+        Sentry.captureMessage('saveTimeline failed', {
+          level: 'error',
+          extra: SentryLogger
+        });
+      }
       response.responseText = data;
       throw response;
     }
@@ -328,10 +332,13 @@ const API = {
       this.status = response.statusText;
       SentryLogger.obj = this.obj;
       SentryLogger.status = this.status;
-      Sentry.captureMessage('saveCourse failed', {
-        level: 'error',
-        extra: SentryLogger
-      });
+
+      if (typeof Sentry !== 'undefined') {
+        Sentry.captureMessage('saveCourse failed', {
+          level: 'error',
+          extra: SentryLogger
+        });
+      }
       response.responseText = data;
       throw response;
     }
--- a/app/assets/javascripts/components/revision_ai_scores/graphs/scores_trends_graph.jsx
+++ b/app/assets/javascripts/components/revision_ai_scores/graphs/scores_trends_graph.jsx
@@ -88,7 +88,7 @@ const renderGraph = (id, statsData, pageTypes, labels, days) => {
       {
         scale: 'x',
         orient: 'bottom',
-        title: 'Revision creation date',
+        title: 'Score check date',
         format: '%b %d %Y',
         tickCount: 'day',
         labelAngle: -20,
--- a/app/assets/javascripts/components/revision_ai_scores/graphs/scores_trends_in_bins_graph.jsx
+++ b/app/assets/javascripts/components/revision_ai_scores/graphs/scores_trends_in_bins_graph.jsx
@@ -141,7 +141,7 @@ const renderGraph = (id, statsData, bins, labels, days, namespaces) => {
       {
         scale: 'x',
         orient: 'bottom',
-        title: 'Revision creation date',
+        title: 'Score check date',
         format: '%b %d %Y',
         tickCount: 'day',
         labelAngle: -20,
--- a/app/controllers/revision_ai_scores_stats_controller.rb
+++ b/app/controllers/revision_ai_scores_stats_controller.rb
@@ -42,7 +42,7 @@ class RevisionAiScoresStatsController < ApplicationController
   def set_historical_scores_by_namespaces
     count_by_date_and_namespace = @scores.group_by do |s|
       # -1 represents an invalid bin
-      [s.revision_datetime.to_date, s.article.namespace, -1]
+      [s.created_at.to_date, s.article.namespace, -1]
     end
                                     .transform_values(&:count)
 
@@ -56,7 +56,7 @@ class RevisionAiScoresStatsController < ApplicationController
   # for some combinations.
   def set_historical_scores_by_max_values
     count_by_date_ns_and_bin = @scores_with_likelihood.group_by do |s|
-      [s.revision_datetime.to_date, s.article.namespace, bin(s.max_ai_likelihood)]
+      [s.created_at.to_date, s.article.namespace, bin(s.max_ai_likelihood)]
     end
                                     .transform_values(&:count)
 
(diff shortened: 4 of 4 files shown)


app/assets/javascripts/components/overview/details.jsx
```diff
@@ -33,7 +33,6 @@ import NamespaceSelect from '../common/namespace_select.jsx';
 
 import EditableRedux from '../high_order/editable_redux.jsx';
 import TextInput from '../common/text_input.jsx';
-import Notifications from '../common/notifications.jsx';
```
ragesoss, line 36: Can you explain why this is being removed from the Details component?

app/controllers/courses_controller.rb
```diff
@@ -16,6 +16,8 @@ class CoursesController < ApplicationController
   respond_to :html, :json
   before_action :require_permissions, only: %i[notify_untrained
                                                delete_all_weeks]
+  # Catches the Database-level uniqueness constraint failure for all actions in this controller
+rescue_from ActiveRecord::RecordNotUnique, with: :record_not_unique_slug_error
```
ragesoss: Is this indented correctly? I would have expected it to be aligned with the other lines above.

app/controllers/courses_controller.rb
```diff
@@ -262,6 +264,7 @@ def delete_all_weeks
     render plain: '', status: :ok
   end
 
+
```
ragesoss: Extra space not needed

ragesoss:
This looks reasonable overall. Is there a more elegant way of doing it than implementing a new controller method and using a rescue? Ideally, the error itself would contain the relevant message, and the system would use that to automatically display it in the Notification in place of the default "Internal Server Error" without needing to rescue it in the controller.

app/assets/javascripts/components/overview/details.jsx
Formasitchijoh: I removed the local notification trigger from the Details component. Since Details is nested within `StudentHandler` and `Course`, having it there caused duplicate banners. I noticed the Course component (the top-level parent) already manages these notifications, so I’ve streamlined it to use that single source of truth. Let me know if I missed a specific edge case where a local notification is still required!
Screenshot 2026-03-03 at 20 41 54

Formasitchijoh:
> This looks reasonable overall. Is there a more elegant way of doing it than implementing a new controller method and using a rescue? Ideally, the error itself would contain the relevant message, and the system would use that to automatically display it in the Notification in place of the default "Internal Server Error" without needing to rescue it in the controller.

I considered using a Model validation, but I went with `rescue_from ActiveRecord::RecordNotUnique`  to avoid the extra database query that` validates_uniqueness_of` performs on every update.
By rescuing the database-level error directly, we ensure data integrity (handling the race condition) without the performance hit. I abstracted the rescue into the controller so that we can return a user-friendly 409 Conflict and a specific translation key instead of a generic 500 Internal Server Error. This seems to align with the [feedback](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6168#issuecomment-2701980271) you gave from the previous attempt on this issue.

Formasitchijoh:
@ragesoss is there any changes required here

ragesoss:
I'll have another look at it soon.

app/assets/javascripts/utils/api.js
```diff
@@ -314,10 +314,14 @@ const API = {
       this.status = response.statusText;
       SentryLogger.obj = this.obj;
       SentryLogger.status = this.status;
-      Sentry.captureMessage('saveCourse failed', {
-        level: 'error',
-        extra: SentryLogger
-      });
+      try {
```
ragesoss: Is this change just for easier debugging, or is it a necessary part of the improved error message behavior? If it's for debugging, please remove it from this PR (although I would be open to a separate PR, if there's a clear explanation for why it's helpful).

app/controllers/courses_controller.rb
ragesoss: I think I don't like doing this rescue at the controller level, and I worry that `ActiveRecord::RecordNotUnique` is too general and may result in an incorrect message if there is some other way of triggering a database uniqueness error that doesn't involve a course slug.

The one place we know this happens is via `#update`, so that might be the place to rescue like we do already with `InvalidWikiError`.

ragesoss: It might make sense to define and throw an explicit SlugNotUnique error from `Course` and propagate that instead of allowing the ActiveRecord error to propagate all the way to the controller.

We also have a pattern in place for rescuing specific error types that might occur across multiple controllers, which might be an appropriate thing to use here... see `lib/errors/rescue_errors.rb`.

Ultimately, any of these approaches are probably fine, but I'd like to avoid introducing new patterns to the codebase if possible.

ragesoss:
I tested this out locally, and it works as expected. Thanks for catching the duplication `Notifications` component as well. I think you're correct that it's completely unnecessary.

app/controllers/courses_controller.rb
Formasitchijoh: @ragesoss  
I see the` lib/errors/rescue_errors.rb`  is more global  which might be a bit disturbing to add a course specific `SlugNotUnique` error into it.  I think I will go with the approach to define and throw an explicit SlugNotUnique error from Course, and remove the general  ActiveRecord::RecordNotUnique from the controller

Also do you think it is better to rescue the error on the model validation or at the database level using the  ActiveRecord::RecordNotUnique

app/assets/javascripts/utils/api.js
Formasitchijoh: I remember is was part of fix cause there seemed to be a state miss match where the previous error showed in place of the current one. I will  check this again.

app/controllers/courses_controller.rb
ragesoss: I think relying on `ActiveRecord::RecordNotUnique` is probably the right way for the error cascade to begin.

app/assets/javascripts/utils/api.js
```diff
@@ -328,14 +328,10 @@ const API = {
       this.status = response.statusText;
       SentryLogger.obj = this.obj;
       SentryLogger.status = this.status;
```
Formasitchijoh, line 334: This fix was needed to ensure that when Sentry is undefined it does not stop the execution of the method. I opened a new PR [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6742) to solve this

lib/errors/rescue_errors.rb
```diff
@@ -76,6 +76,20 @@ def self.rescue_participating_user(base)
       end
     end
 
+    def self.rescue_slug_not_unique(base)
+      base.rescue_from ActiveRecord::RecordNotUnique do |e|
+         # Target the duplicate slug error
+        if e.message.include?('index_courses_on_slug')
```
Formasitchijoh, line 82: Specified the index of the slug to make sure we are capturing the error caused by duplicate course slug.  if any other operation  violates a uniqueness contraint it will be caught in the else clause

Formasitchijoh:
@ragesoss  Could you check this out, The history is a bit messy had to merge this [fix](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6742) I will clean it up once you are Satisfied with the fix

ragesoss:
@Formasitchijoh code-wise, this looks good to me. I assume you've confirmed that it still works the same way as when I tested it before, and the video is still an accurate depiction? If so, I'm ready to merge it once it's cleaned up.

Formasitchijoh:
> @Formasitchijoh code-wise, this looks good to me. I assume you've confirmed that it still works the same way as when I tested it before, and the video is still an accurate depiction? If so, I'm ready to merge it once it's cleaned up.

Yes it does added an Image of it along side

PR #6696, opened 2026-02-25
Fix stale sandbox URLs when user is renamed

### LINKED ISSUE #4276: Improve handling of sandbox URLs when a username changes
### Current Behavior:
When a username changes and that user has Assignments, the sandbox_url values will remain pointed to the userspace of the old username. This can affect both the user whose name changed, as well as classmates who are assigned to the same group or are peer reviewing that user's articles.

### Desired Behavior:
Upon username change:

* Identify which sandboxes corresponding to that student's Assignments, as well as classmates' peer reviews, have already been created in the old userspace, and perform a page move to bring them to the new userspace of the renamed user.
* Update the sandbox_url for each Assignment in the course that pointed to the old username, replacing it with the new one.

### Related Files & Code:

`UserImporter.from_omniauth` is where the system changes an updated username upon login. This is probably the place to begin.

nexpectArpit:
When a user's Wikipedia username changes (via UserImporter or direct update), the associated Assignment sandbox URLs were not updated. This left students and instructors with broken links pointing to the renamed user's old sandbox page.

### Fix:
Implemented an `after_save` callback in the `User` model that:
1. Detects if the `username` attribute has changed.
2. Iterates through the user's assignments.
3. Automatically updates the `sandbox_url` to match the new username.
4. Includes safety checks to only update URLs that match the default pattern (preserving any custom sandbox URLs manually set by users).

Closes #4276

#### Verification
- Added 3 new tests to `spec/models/user_spec.rb` covering the rename 
  scenario, custom URL protection, and no-op saves.
- Verified visually in a local environment that renames now trigger 
  automatic URL synchronization.

#### Before(Conceptual):
Before the fix, whenever a username changed, the sandbox URL stored in the assignment record became stale, pointing to the old username and creating a broken link.

#### After:
in this demonstration, I am showing the user `Sync_Renamed`. This user was just renamed from `Sync_User` to `Sync_Renamed`. As you can see in the student drawer, the sandbox link for `Final_Fix_Test` has automatically updated to point to the new username: `.../User:Sync_Renamed/...`
Screenshot 2026-02-25 at 11 43 24 PM

This proves that the new `after_save `callback is successfully detecting username changes and synchronizing all associated assignment sandbox URLs, ensuring that links remain functional after renames.

@ragesoss @gabina @Abishekcs Please review it.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -446,13 +446,17 @@ describe User do
                wiki:,
                article_title: 'Test_Article',
                role: Assignment::Roles::REVIEWING_ROLE,
-               sandbox_url: "#{wiki.base_url}/wiki/User:Original_Username/Test_Article/Classmate_Peer_Review")
+               sandbox_url: "#{wiki.base_url}/wiki/" \
+                            'User:Original_Username/Test_Article/' \
+                            'Classmate_Peer_Review')
       end
 
       it 'updates the sandbox_url to use the new username' do
         user.update!(username: 'New_Username')
         expect(review_assignment.reload.sandbox_url)
-          .to eq("#{wiki.base_url}/wiki/User:New_Username/Test_Article/Classmate_Peer_Review")
+          .to eq("#{wiki.base_url}/wiki/" \
+                 'User:New_Username/Test_Article/' \
+                 'Classmate_Peer_Review')
       end
 
       it 'does not update sandbox_url if the peer review sandbox already exists' do


app/models/user.rb
```diff
@@ -103,6 +104,21 @@ def self.search_by_real_name(real_name)
...
+    return unless old_username
+
+    assignments.each do |assignment|
+      next unless assignment.sandbox_url == assignment.default_sandbox_url(old_username)
+
+      assignment.update(sandbox_url: assignment.default_sandbox_url(new_username))
+    end
+  end
+
+  public
```
Ajay-singh1: Why this public classifier can you explain please?

app/models/user.rb
```diff
@@ -103,6 +104,21 @@ def self.search_by_real_name(real_name)
     User.where('lower(real_name) like ?', "%#{real_name}%")
   end
 
+  private
```
Ajay-singh1: You can group the private methods instead of duplicating I guess

nexpectArpit: Thanks for suggesting

app/models/user.rb
```diff
@@ -254,4 +255,15 @@ def profile_image
   def ensure_valid_email
     self.email = nil if ValidatesEmailFormatOf::validate_email_format(email)
   end
+
+  def update_assignment_sandbox_urls
+    old_username, new_username = saved_change_to_username
```
ragesoss: We haven't used this bit of Rails magic much before, so it would be good to explain in a comment how `saved_change_to_username` works and link to the Rails documentation for it.

app/models/user.rb
```diff
@@ -32,6 +32,7 @@
 class User < ApplicationRecord
   alias_attribute :wiki_id, :username
   before_validation :ensure_valid_email
+  after_save :update_assignment_sandbox_urls, if: :saved_change_to_username?
```
ragesoss, line 35: We generally try to minimize the use of callbacks like this, but I'm guessing this instance is pretty safe, as it shouldn't cause any extra database queries to check whether the username was changed, and that is a pretty rare occurrence.

app/models/user.rb
```diff
@@ -254,4 +255,15 @@ def profile_image
   def ensure_valid_email
     self.email = nil if ValidatesEmailFormatOf::validate_email_format(email)
   end
+
+  def update_assignment_sandbox_urls
```
ragesoss, line 262: A comment here should explain the logic, noting that we only want to update the sandbox URLs if the current URL is the default one corresponding to the old username.

app/models/user.rb
```diff
@@ -254,4 +255,15 @@ def profile_image
...
+    old_username, new_username = saved_change_to_username
+    return unless old_username
+
+    assignments.each do |assignment|
+      next unless assignment.sandbox_url == assignment.default_sandbox_url(old_username)
+
+      assignment.update(sandbox_url: assignment.default_sandbox_url(new_username))
```
ragesoss: There's an important situation that I think this doesn't account for: there may be multiple students assigned to the same article (as part of a group), whose sandboxes also point to the userspace of the now-renamed user. For those cases, the URLs for everyone in the group should be changed.

app/models/user.rb
ragesoss, line 262: One important consideration is whether or not the sandbox has already been created. If not, simply updating the assignment to use the new username is fine, but if the sandbox has already been created, then updating it without also moving the page itself will be disruptive, as a user may lose track of the old sandbox where they've already saved work.

ragesoss (changes requested):
I think as a first implementation, this should include a check for whether and of the the to-be-changed sandboxes already exists, and it can simply skip updating if so.

I guess the Assignment#flags data structure should include the necessary information for determining whether any of the sandboxes exist.

Because this is fairly complex logic, it might make sense to move the updating itself (based on the old and new usernames) into a separate `service` file.

app/models/user.rb
nexpectArpit: sure ,I will add a comment explaining that it returns an array of the [old, new] values and include a link to the [ActiveRecord::AttributeMethods::Dirty](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-saved_change_to_attribute) documentation for better refrence.

app/models/user.rb
nexpectArpit: I'll make sure to keep the logic within the callback as minimal as possible by moving the heavy lifting into a separate service class.

app/models/user.rb
nexpectArpit: will add.

app/models/user.rb
nexpectArpit: i will update the logic to search across all assignments in the relevant courses to find any that point to the old userspace URL, ensuring that group members don't get 'left behind' with a broken link.

app/models/user.rb
nexpectArpit: for that case i will add a check using `Assignment\#flags` (which tracks the on-wiki state) before applying any update. If the status is anything other than does_not_exist, it will skip the update entirely to avoid potential data loss.

nexpectArpit:
rebased with latest ruby version.

app/services/update_assignment_sandbox_urls.rb
```diff
@@ -0,0 +1,33 @@
...
+    # and ensures we don't miss assignments in courses where the user isn't joining records.
+    Assignment.where('sandbox_url LIKE ?', "%User:#{@old_username}/%").each do |assignment|
+      # 1. Safety Guard: Only update if the current URL matches the default pattern 
+      # for the old username. This ensures we don't overwrite manual customizations.
+      next unless assignment.sandbox_url == assignment.default_sandbox_url(@old_username)
+
+      # 2. Existence Check: Only update if the sandbox has not been created on the wiki yet.
+      # If work has already started (status is not 'does_not_exist'), we skip the update
+      # to avoid disrupting the user's workflow.
+      next unless assignment.draft_sandbox_status == 'does_not_exist'
```
ragesoss: This needs to check all the related sandbox statuses, not just the draft sandbox. For an assigned article, there are up the three different sandboxes (bibliography, outline and draft) where a user may have started doing relevant work, and if any of those exist, we don't want to change the URL or it will break a link. Similarly, if it's an assigned review, it shouldn't make any changes if that review sandbox has already been created. Think just looping through all four existence checks makes sense here.

spec/models/user_spec.rb
```diff
@@ -373,4 +373,68 @@
       expect(result).to eq([search_user, similar_search_user])
     end
   end
+
+  describe '#update_assignment_sandbox_urls' do
```
ragesoss, line 377: The tests should probably cover a REVIEWING_ROLE example.

The relevant situation is where a classmate has signed up to review an article assigned to "Original_Username", so user "Classmate" has a review assignment where the sandbox URL is "...User:Original_Username/Test_Article". If they already created that, none of the records should get updated, but if they haven't then that review assignment should also get updated (so that the new URL for their peer review is "...User:New_Username/Test_Article/Classmate_Peer_Review").

nexpectArpit: I've added a specific testing context for `REVIEWING_ROLE` in  `spec/models/user_spec.rb`.
The test verifies that when a user's username changes, peer review assignments (which belong to other reviewing users but point to the old username's space) are correctly updated. I also added logic in the service to properly append `_Peer_Review` dynamically when calculating expected default URLs for reviewing assignments.

app/services/update_assignment_sandbox_urls.rb
nexpectArpit: updated the `UpdateAssignmentSandboxUrls` service to explicitly check an array of all four statuses (`draft`, `bibliography`, `outline`, and `peer_review`)

ragesoss:
LGTM!

PR #6657, opened 2026-02-01
Update Ruby version to 3.4.8 

SanjanaSogimatt:
### What this PR does

This PR upgrades the Ruby version from 3.1.2 to 3.4.8 and updates the codebase to be compatible with Ruby 3.4+. It also refactors Sentry error logging to use the recommended block syntax for better context management.

### Key Changes:

1. **Ruby Version Upgrade**:
   - Upgraded Ruby from `3.1.2` to `3.4.8` in both `.ruby-version` and `Gemfile`
   - Updated CI workflow (`.github/workflows/ci.yml`) to use Ruby 3.4.8

2. **Required Gem Additions for Ruby 3.4+**:
   - Added `csv` gem (version 3.3.5) - CSV library is no longer a default gem in Ruby 3.4+
   - Added `observer` gem (version 0.1.2) - Observer library is no longer in the standard library since Ruby 3.1+

### Files Modified:
- `.github/workflows/ci.yml`
- `Gemfile`
- `Gemfile.lock`

## Why these changes were made

1. **Ruby 3.4.8 Upgrade**: Ruby 3.4.8 includes performance improvements, security fixes, and new features. This upgrade ensures the application benefits from the latest Ruby improvements while maintaining compatibility with the existing codebase.

2. **CSV and Observer Gems**: In Ruby 3.4+, the `csv` library is no longer a default gem and must be explicitly added. The `observer` library was removed from the standard library in Ruby 3.1+, so it must be added as a gem dependency.

## AI usage

I used Cursor AI (Claude Sonnet 4.5) to:
- Identify Ruby 3.4+ compatibility requirements (csv and observer gems)

The AI helped me understand:
- Ruby 3.4+ breaking changes and required gem additions

### External Documentation
- https://github.com/ruby/ruby/releases/tag/v3_4_8
- **[Ruby 3.4.0 Release Notes](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/)**: Official release notes for Ruby 3.4.0, detailing new features, performance improvements, and breaking changes.
- **[Ruby 3.4.0 Compatibility Guide](https://www.ruby-lang.org/en/news/2024/12/25/ruby-3-4-0-released/#compatibility-issues)**: Information about compatibility issues and migration notes when upgrading to Ruby 3.4.

PR #6699, opened 2026-02-27
Rails 8.1.2 Upgrade

SanjanaSogimatt:
### What this PR does

Updates Ruby and Rails and related dependencies in `Gemfile` and `Gemfile.lock`:

- **Ruby:** 3.1.2 → 3.4.8
- **Rails:** 7.1.0 → 8.1.2
- **New gems:**
  - `observer` — required for Ruby 3.1+ (removed from the standard library)
  - `csv` — added explicitly (standard library in Ruby but listed for consistency)
- **TicketDispenser:** Git dependency revision updated.
- **Gemfile.lock:** All transitive dependencies updated for Rails 8.1 and Ruby 3.4 (e.g. actionpack, activerecord, activesupport, mail, annotate, and new dependencies such as `useragent`, `securerandom`, `action_text-trix`, `logger`, etc.).

## Screenshots

**Before:** N/A — backend change only; no UI change.

**After:** N/A — backend change only; no UI change.

Will update the description


Gemfile
```diff
@@ -1,14 +1,16 @@
...
-gem 'rails', '7.1.0'
+gem 'rails', '8.1.0'
 gem 'jbuilder' # DSL for building JSON view templates
 gem 'haml-rails' # HTML template language, used instead of ERB
 gem 'bootsnap', require: false # Makes rails boot faster via caching
 gem 'faker', require: false # Generates random data for example records
 gem 'figaro' # easy access to ENV variables. Deprecated.
 gem 'puma'
+gem 'observer' # Observer library (required for Ruby 3.1+ as it's no longer in standard library)
+gem 'csv'
```
Abishekcs: Better add a small single line comment for this too as you did for `observer`

ragesoss:
Awesome that this is passing. This PR currently includes the serialization config changes that are now merged, so this PR should be cleaned up to exclude those.

I'm also interested in whether the Ruby version change can easily be separated from the Rails upgrade. Is our current Rails version (7.1.0) compatible with Ruby 3.4.8? If possible, I'd prefer to make those upgrades in production separate. Ruby upgrades require changes on the servers, and if something unexpected goes wrong, it's helpful to have only the Ruby changes to worry about, without a major Rails version bump alongside as a possible source of problems.

This set of changes will be relatively small, so I'm willing to keep them together if necessary (and I can test the deployment on the staging server before trying it with production). But if they can be easily separated, it would be better to upgrade Ruby first, then Rails.

spec/controllers/campaigns_controller_spec.rb
```diff
@@ -429,9 +429,8 @@
 
       context 'when campaign does not exist' do
         it 'raises routing error' do
-          expect do
-            get '/campaigns/non-existent-campaign/articles'
-          end.to raise_error(ActionController::RoutingError)
+          get '/campaigns/non-existent-campaign/articles'
+          expect(response.status).to eq(404)
```
ragesoss, line 433: It looks like these are all related to a change in how controller specs work. The PR description should explain this change and link to relevant upgrade docs.

ragesoss: Ah, I see this is all from building on top of the Rails 7.2 PR.

SanjanaSogimatt:
> I'm also interested in whether the Ruby version change can easily be separated from the Rails upgrade. Is our current Rails version (7.1.0) compatible with Ruby 3.4.8? If possible, I'd prefer to make those upgrades in production separate. Ruby upgrades require changes on the servers, and if something unexpected goes wrong, it's helpful to have only the Ruby changes to worry about, without a major Rails version bump alongside as a possible source of problems.

Yes, I did make just the Ruby changes in another PR [here](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6657) I made the same ruby upgrade here because Rails 8 and above requires Ruby version more than 3.2 so once that's merged I can reduce the PR size

PR #6702, opened 2026-02-28
[Backup] Add documentation

gabina:
## What this PR does
This PR adds backup documentation.

### DIFF
--- a/config/backup/backup.sh
+++ b/config/backup/backup.sh
@@ -4,7 +4,8 @@ set -euo pipefail
 
 BACKUP_ROUTE="ROUTE_TO_BACKUPS"
 QUERY_ROUTE="ROUTE_TO_QUERIES"
-JSON_ENDPOINT="https://outreachdashboard.wmflabs.org/system/can_start_backup.json"
+DASHBOARD_URL="URL_TO_DASHBOARD"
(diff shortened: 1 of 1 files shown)


server_config/SERVER_CONFIG.md
```diff
@@ -10,6 +10,46 @@ These files are the main web server configuration files:
 * `dashboard.conf` # apache config for dashboard.wikiedu.org (ie, production)
 * `dashboard-testing.conf` # apache config for dashboard-testing.wikiedu.org (ie, staging)
 
-## Database backup script
+## Database backup
```
ragesoss, line 13: This should also outline how the backup script is installed (eg, copying backup.sh to a location outside the git directory and modifying it with the relevant routes and endpoints? example cron command?) and should note where the backups will be found in our two production servers. (I found them yesterday, but had to hunt around a bit to figure out where they were.)

gabina: Yes, I agree. At first, I didn't do it because I felt it was something more "internal" than something to be in the public repo, but makes total sense. I added a section titled "Installation" trying to describe with some detail the idea of what I did, but without writing every single command I executed.

PR #6704, opened 2026-03-03
Fix: JS injection bug: escape username with .to_json in _head.html.haml

### LINKED ISSUE #6703: Username with paired quotation marks causes JS syntax error
Usernames that have paired internal double quote marks — like `Rage"theTest"Test` and `B"theSituation"H` — completely break the JS frontend of the Dashboard. If a user with such a username logs in, the Dashboard is unusable.

I think the problem originates from `app/views/shared/_head.html.haml` line 27, where a username string with quotes isn't properly escaped to carry through correctly.

Fix the bug, and include a test against regression using an example username that would trigger it.

ManishKumar081:
## What this PR does

Fixes a bug where usernames containing double quotes (e.g. `Rage"theTest"Test`) break the JS frontend when that user logs in.

In `app/views/shared/_head.html.haml` line 27, the username was interpolated directly into a JS string literal without escaping, causing a syntax error. Fixed by using `.to_json` which safely escapes all dangerous characters. 
Regression spec added.

Closes #6703 

## AI usage

Used Claude (Anthropic) to setup the local test environment and PR description( To follow the template)

## Screenshots

**Before:**
```javascript
currentUser = {
  id: "42",
  username: "Rage"theTest"Test"  // ← syntax error, breaks frontend
}
```

**After:**
```javascript
currentUser = {
  id: 42,
  username: "Rage\"theTest\"Test"  // ← valid JS
}
```

## Test Results:
image

Other fields in the same `:javascript` block (`default_course_string_prefix`, `SentryDsn`, etc.) use the same unsafe interpolation pattern and may be worth a follow-up audit.

### DIFF
--- a/app/views/shared/_head.html.haml
+++ b/app/views/shared/_head.html.haml
@@ -23,7 +23,7 @@
     availableLocales = #{I18n.available_locales.to_json.html_safe};
 
     currentUser = {
-      id: #{current_user&.id || '""'},
+      id: "#{current_user&.id}",
       username: #{(current_user&.username || '').to_json}
     }
 


spec/views/shared/_head.html.haml_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    allow(view).to receive(:i18n_javascript_tag).and_return('')
+    allow(view).to receive(:logo_favicon_tag).and_return('')
+    allow(view).to receive(:csrf_meta_tags).and_return('')
+    allow(view).to receive(:content_for).and_return('')
+    allow(view).to receive(:content_for?).and_return(false)
+  end
+
+  describe 'currentUser JS object' do
+    context 'when username contains double quotes' do
+      let(:user) { instance_double('User', id: 42, username: 'Rage"theTest"Test') }
```
ragesoss: This example should be added to `UsernameTestHelper`, and the spec should loop through all of them to makes sure each one works without causing problems.

ManishKumar081: I understand the need for looping through these. On it will reflect and update it

ManishKumar081:
the Test time seems to be significantly raised. I think the loop is causing the time to increase. I since we do want the loop it is not related,  thought to mention it.

image

spec/support/username_test_helper.rb
```diff
@@ -11,7 +11,14 @@ def self.test_usernames
       'Łukasz',
       'İbrahim',
       'Αλέξανδρος',
-      'Дмитрий'
+      'Дмитрий',
+      'Rage"theTest"Test',
+      'B"theSituation"H'
```
ragesoss: Only one of these should be included. They represent the same thing, so we don't gain any extra reliability by having both.

spec/views/shared/_head.html.haml_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    allow(view).to receive(:i18n_javascript_tag).and_return('')
+    allow(view).to receive(:logo_favicon_tag).and_return('')
+    allow(view).to receive(:csrf_meta_tags).and_return('')
+    allow(view).to receive(:content_for).and_return('')
+    allow(view).to receive(:content_for?).and_return(false)
+  end
+
+  describe 'currentUser JS object' do
+    context 'when username contains problematic characters' do
+      UsernameTestHelper::PROBLEMATIC_USERNAMES.each do |username|
```
ragesoss: I meant that all the test usernames should be looped through, not just the ones related to this issue... the idea here is to use a variety of usernames that might represent edge cases for string formatting, and test against each of them any time we are writing tests for username-related bugs.

spec/support/username_test_helper.rb
ManishKumar081: Done and  removed 'B"theSituation"H', keeping only 'Rage"theTest"Test' as a single representative example.

spec/views/shared/_head.html.haml_spec.rb
ManishKumar081: Fixed and now using UsernameTestHelper.test_usernames.each to loop through all edge-case usernames, not just the ones added for this issue.

ManishKumar081:
All 10 examples passing locally including the double-quote username and all non-ASCII edge cases from UsernameTestHelper.test_usernames:
image
Both review points addressed:
Removed 'B"theSituation"H' and keeping only 'Rage"theTest"Test' as a single representative example.
Switched to UsernameTestHelper.test_usernames.each to loop through all edge-case usernames, not just the ones added for this issue.

spec/views/shared/_head.html.haml_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    allow(view).to receive(:csrf_meta_tags).and_return('')
+    allow(view).to receive(:content_for).and_return('')
+    allow(view).to receive(:content_for?).and_return(false)
+  end
+
+  describe 'currentUser JS object' do
+    context 'when username contains problematic characters' do
+      UsernameTestHelper.test_usernames.each do |username|
+        it "escapes the username #{username.inspect} and produces valid JS" do
+          user = instance_double('User', id: 42, username: username)
```
ragesoss: The convention in this codebase would be to use FactoryBot to set up test objects, rather than `instance_double`. `build(:user, ...)` or `create(:user, ...)` in this case.

spec/views/shared/_head.html.haml_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe 'shared/_head', type: :view do
```
ragesoss: This doesn't match the way we set up `describe` blocks elsewhere in the test suite. It can probably just be `describe 'shared/_head ' ...`

spec/views/shared/_head.html.haml_spec.rb
ManishKumar081: Fixed in the latest commit and switched to FactoryBot.build(:user, username: username).

spec/views/shared/_head.html.haml_spec.rb
ManishKumar081: Fixed in the latest commit and changed to describe 'shared/_head', type: :view do.

ManishKumar081:
Squashed into a single clean commit. All 10 examples in our spec pass locally, including all non-ASCII usernames and Rage"theTest"Test.

image

ragesoss:
Looks good to me. I'll merge it once the build passes.

ragesoss:
@ManishKumar081 looks like there's a failing test, and it might be related to this change. Please check whether that spec passes for you locally.

```
rspec ./spec/features/language_switcher_spec.rb:16 # language_switcher user logged out switches to another language using URL param
```

I'll restart the build just in case it was an unrelated intermittent failure.

ManishKumar081:
Found the root cause of the language_switcher failure and language_picker.jsx checks currentUser.id !== '', so changing id from "" to null for logged-out users broke the language switcher. Fixed by keeping id as "" when no user is signed in. Both specs now pass locally.

app/views/shared/_head.html.haml
```diff
@@ -23,8 +23,8 @@
     availableLocales = #{I18n.available_locales.to_json.html_safe};
 
     currentUser = {
-      id: "#{current_user&.id || ''}",
-      username: "#{current_user&.username || ''}"
+      id: #{current_user&.id || '""'},
```
ragesoss: This is a little weird. Is it necessary to keep the quoted quotes, or will it work if it's just a simple empty string?

ManishKumar081: The quoted quotes are necessary to ensure a valid JS string when no user is logged in and without them, nil would render as nothing, producing invalid JS like id: ,

ManishKumar081:
I further simplified and changed to id: "#{current_user&.id}" which outputs "" for logged-out users and the actual id for logged-in users, keeping it consistent with the existing language_picker.jsx check.

PR #6707, opened 2026-03-04
Add integration with Originality AI detector

gabina:
## What this PR does
This PR integrates [Originality AI detector](https://originality.ai/) to the comparison AI tool admin view.

It creates a new `OriginalityApi` class that works similar to the `PangramApi`.

According to the [Originality docs](https://docs.originality.ai/originality-ai-api-v1#AI-Checker), it supports different models:
image

## Screenshots
Before:
image

After:
[Screencast from 2026-03-05 18-17-54.webm](https://github.com/user-attachments/assets/41c62ab1-f400-4905-befe-a396431983fb)

Originality results:
image

### DIFF
--- a/lib/originality_api.rb
+++ b/lib/originality_api.rb
@@ -44,6 +44,8 @@ class OriginalityApi
     @result = JSON.parse @response.body
   end
 
+  private
+
   def base_request_body(text)
     {
       check_ai: true,


lib/originality_api.rb
```diff
@@ -0,0 +1,60 @@
...
+  
+  API_URL = 'https://api.originality.ai/api/v3/scan'
+
+  # Originality Models descriptions:
+  # Very hard to bypass (higher False Positive rate).
+  # Use when you want to be certain AI was not used
+  TURBO_MODEL = 'turbo'
+  # Ideal for teachers and students.
+  # Accurate for STEM answers (Code and Formulas).
+  ACADEMIC_MODEL = 'academic'
```
gabina, line 17: Originality provides different models. I added support only for two of them.
Since for now we're using AI detectors mainly for EN Wikipedia, I discarded the multilingual language model. I also discarded the Lite model because  it's recommended when "Lightly AI-edited content is acceptable".

ragesoss: Great. I think adding support for the Lite model will be relevant as well, but I have some related notes for a next batch of work. I'll open an issue to collect them.

app/views/ai_tools/_originality.html.haml
```diff
@@ -0,0 +1,10 @@
+%p
+  Prediction:
+  - if originality_result['results']['ai']['classification']['AI'] == 1
+    AI written
+  - else
+    Human written
```
gabina, line 6: I added this `if` to print something more human friendly.

gabina: The raw classification is a classification hash like  `{'AI' -> 1, 'original' -> 0}`

app/views/ai_tools/_originality.html.haml
```diff
@@ -0,0 +1,10 @@
...
+%p AI confidence: #{originality_result['results']['ai']['confidence']['AI']}
+%p Human confidence: #{originality_result['results']['ai']['confidence']['Original']}
```
gabina, line 8: I understand this confidence field is the probability of each class (they add up to 1). I wanted to keep the word "confidence" because that's the name used by the API, but maybe we want to update it.

ragesoss: Following their terminology seems fine for now.

gabina:
There's some room for improving the visual side of this, but this is ready for you to take a first look @ragesoss . Once you're OK, I can add some basic specs for the `OriginalityApi`.

ragesoss:
Yeah, looks good to me!

gabina:
Well, I couldn't come up with good unit specs for `OriginalityApi`, because I was mocking a lot of things (I don't want to actually make the request in the spec) and I thought the spec was hard to understand. I also took a look at other similar classes to get some inspiration, and I don't think we have unit specs for them. I think we can leave the specs for the future. I'm open to suggestions if you have an opinion on how to add reasonable specs for it.

app/controllers/ai_tools_controller.rb
```diff
@@ -14,6 +15,8 @@ def compare_pangrams
     @plain = GetRevisionPlaintext.new(@rev_id, @wiki, diff_mode: @diff_mode, from_rev: @from_rev)
     @pangram_v2_result = PangramApi.v2.inference @plain.plain_text
     @pangram_v3_result = PangramApi.v3.inference @plain.plain_text
+    @originality_turbo_result = OriginalityApi.turbo.inference @plain.plain_text
```
ragesoss, line 18: I only used the term `inference` because that's the language from the Pangram API documentation, but now that we have a series of similar calls to multiple APIs, it might make sense to refactor these or use a clearer method name, like `ai_detection`. This is fine for now though.

ragesoss:
I implemented a mock API response for Pangram in `check_revision_with_pangram_spec.rb`.  I think a similar testing strategy is fine for other AI detection APIs, leaving the API class itself without unit tests, but testing other things around the expected shape of the response.

ragesoss:
LGTM!

PR #6708, opened 2026-03-05
feat: Add survey completion time tracking and display for analytics.

### LINKED ISSUE #6355: Survey system should record completion time
For analyzing the surveys and improving our usage of them, it would be nice to have data on how long surveys take for each submission.

lakshita10341:
## What this PR does

Adds survey completion duration tracking so admins can see how long users take to complete surveys.
Closes #6355 
**Problem:** The survey system captures what users answered but has no way to measure how long each submission takes. This makes it difficult to identify surveys that are too long or confusing.

**Solution:** A new `survey_completion_times` table records start/end timestamps for each survey submission. The frontend sends a POST when the user starts the survey, and a PUT when they submit. Duration is computed server-side.

### Changes

**Backend:**
- New migration, model (`SurveyCompletionTime`), and controller with `start`/`complete` actions
- New routes: `POST /survey/start` and `PUT /survey/complete`
- CSV export (`Survey#to_csv`) now includes `duration_seconds` as the last column

**Frontend:**
- `Survey.js` sends tracking requests on survey start and completion (fails silently to never block the survey)

**Admin views:**
- Survey list and results index show "Avg Duration" column
- Per-survey results page shows full stats: average, median, fastest, slowest, completion rate, and a time distribution chart

**Edge cases handled:**
- Old surveys show `--` for all stats (no backfill needed)
- Abandoned surveys (started but not finished) are excluded from stats

## AI usage

I used Gemini to help implement this feature. The AI was used for:
- Planning the database schema
- Writing the migration, RSpec model and controller tests
- Debugging Docker environment issues and data setup for local testing

## Screenshots

Before:
Screenshot from 2026-03-05 11-26-31

Screenshot from 2026-03-05 11-18-04
Screenshot from 2026-03-05 11-17-40

After:
Screenshot from 2026-03-05 10-57-21

Screenshot from 2026-03-05 10-58-07

Screenshot from 2026-03-05 10-58-52

Screenshot from 2026-03-05 10-59-17

##Open concerns
- Duration tracking only applies to surveys taken after this change is deployed. There is no backfill for historical survey submissions.


Abishekcs:
> The frontend sends a POST when the user starts the survey, and a PUT when they submit. Duration is computed server-side.

`The frontend sends a POST when the user starts the survey` - This will create a record in the DB right ? So what if the user lefts the survey before they submit.

I haven't look at the code yet so I had this question.

db/migrate/20260303160000_create_survey_completion_times.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class CreateSurveyCompletionTimes < ActiveRecord::Migration[7.0]
+  def change
+    create_table :survey_completion_times do |t|
+      t.integer :survey_id, null: false
+      t.integer :user_id, null: false
+      t.integer :survey_notification_id
+      t.datetime :started_at, null: false
+      t.datetime :completed_at
+      t.integer :duration_in_seconds
```
Abishekcs: `duration_in_seconds` is this required ?  Can't the time to calculate `SurveyCompletionTime` be obtained from `started_at and completed_at`

lakshita10341:
> > The frontend sends a POST when the user starts the survey, and a PUT when they submit. Duration is computed server-side.
> 
> `The frontend sends a POST when the user starts the survey` - This will create a record in the DB right ? So what if the user lefts the survey before they submit.
> 
> I haven't look at the code yet so I had this question.

Yes, this will create a record in db, with some starting time and 'nil' ending time if the user does not submit the survey. I added a `scope :completed, -> { where.not(completed_at: nil) }` to the model, so all of our duration statistics (average, median, fastest, etc.) explicitly filter out those abandoned surveys and only look at fully completed ones and additionally this also allow us to display the completion rate which we our showing at admin frontend.

db/migrate/20260303160000_create_survey_completion_times.rb
lakshita10341: yes, sorry i missed it. The duration column is not required, so i will remove it in next commit.

ragesoss:
I don't think it makes sense to add a separate table for survey completion time. We already have a table that represents a set of survey responses being submitted, which is `Rapidfire::AnswerGroup`. I think adding a `started_at` field to that table be sufficient; `created_at` represents the timestamp when the survey was completed.

lakshita10341:
> I don't think it makes sense to add a separate table for survey completion time. We already have a table that represents a set of survey responses being submitted, which is `Rapidfire::AnswerGroup`. I think adding a `started_at` field to that table be sufficient; `created_at` represents the timestamp when the survey was completed.

That is right, this approach will keep the database clean. But the main reason I initially considered a separate table was to track abandonment rates and avoid data duplication issues.

Since `AnswerGroup` records are only created after a successful submission, we wouldn't be able to see how many people started a survey but quit halfway through. The separate table allows us to calculate the Completion Rate analytics (starts vs. completions).

Also, since a Survey can contain multiple `QuestionGroups`, submitting a single survey often creates multiple AnswerGroup rows (one for each group). So, we need to duplicate the started_at timestamp across all of them.

Please let me know, if adding field to `Rapidfire::AnswerGroup` table is preferred approach, I'm happy to proceed with that!

ragesoss:
That's a good point about tracking abandonment rates, and I think I'm open to a new table if that's part of the design. The name of such a table should be clearer, as `survey_completion_times` implies it's only for completed surveys.

lakshita10341:
Will `SurveySession` work better?

ragesoss:
That sounds reasonable, although 'session' implies each record is limited to a single session, with different records if a user starts and abandons a survey, then later starts and completes it. Would that be how it works?

lakshita10341:
At every 'Start' attempt, there will be a new session record created, which lets us accurately track abandonment vs. completion rates. For the CSV export and duration analytics, we then only look at the successfully completed session.

Abishekcs:
Is index on all three required ?


``` ruby

add_index :survey_sessions, :survey_id
add_index :survey_sessions, :user_id
add_index :survey_sessions, %i[survey_id user_id]

```

lakshita10341:
> Is index on all three required ?
> 
> ```ruby
> add_index :survey_sessions, :survey_id
> add_index :survey_sessions, :user_id
> add_index :survey_sessions, %i[survey_id user_id]
> ```

Yeah, Since we have composite index on ` [:survey_id, :user_id]`,  we can remove the `survey_id` index. I'll update the code.

Abishekcs:
Can you also point to which part of the code is using this `composite index`

lakshita10341:
> Can you also point to which part of the code is using this `composite index`

```ruby
 add_index :survey_sessions, %i[survey_id user_id]
```
I am referring to this part of code

Abishekcs:
No I mean in the actual Survey codebase

Abishekcs:
That might be confusing what I just said. 

I mean which `Query` you wrote is utilizing this Composite Index of `survey_id` and `user_Id`.

Abishekcs:
One problem with adding indexes early is cardinality. Make sure you’ve checked in MySQL that the query planner is actually using both indexes.

lakshita10341:
> One problem with adding indexes early is cardinality. Make sure you’ve checked in MySQL that the query planner is actually using both indexes.

While the query is not using directly both the indexes, but this composite index still helps here:
```ruby
SurveySession.where(survey_id: id).index_by(&:user_id)
```
Although the `survey_id` has low cardinality, but this composite index is acting as covering index mapping directly the user id from index tree, making the lookup fast.

Abishekcs:
hmm can you show me the query planner using `EXPLAIN` or which ever you prefer

Abishekcs:
I don't think the `user_id` is being even used 

screenshot-2026-03-18_12-56-10

screenshot-2026-03-18_12-57-19

lakshita10341:
Sorry I was wrong. Since we are using `select *` the composite index here have no use. So we can remove this, and can keep index on `survey_id` only. Will it work?

Abishekcs:
> Sorry I was wrong. Since we are using `select *` the composite index here have no use. So we can remove this, and can keep index on `survey_id` only. Will it work?

keeping index on only `survey_id` make sense.

app/models/survey_session.rb
```diff
@@ -0,0 +1,28 @@
...
+#  started_at             :datetime         not null
+#  completed_at           :datetime
+#  created_at             :datetime         not null
+#  updated_at             :datetime         not null
+#
+
+class SurveySession < ApplicationRecord
+  belongs_to :survey
+  belongs_to :user
+  belongs_to :survey_notification, optional: true
```
Abishekcs, line 20: What is the purpose of `optional` here ?

db/migrate/20260303160000_create_survey_sessions.rb
```diff
@@ -0,0 +1,19 @@
...
+      t.integer :user_id, null: false
+      t.integer :survey_notification_id
+      t.datetime :started_at, null: false
+      t.datetime :completed_at
+
+
+      t.timestamps
+    end
+
+    add_index :survey_sessions, :user_id
```
Abishekcs: After removing it from `Composite Index` why add it separately again? Where is this being used ?

app/models/survey_session.rb
lakshita10341: If the user  takes a survey without clicking on the email link, then survey notification id will be null. That's why optional is used.

db/migrate/20260303160000_create_survey_sessions.rb
lakshita10341: Sorry, I meant to remove it alongside the composite index but it was an oversight. I've removed it now. I realized we can add it back if in future we need it.

Abishekcs:
Thanks for the changes. I will have a look within the next few days,

ragesoss (changes requested):
This adds `survey_sessions` for duration tracking plus admin UI for the resulting stats. The shape — a separate table to support abandonment-rate tracking — is a defensible expansion of issue #6355, and the in-PR conversation has already settled the table name and index decisions. Specs and lint are clean on the changed files.

A few items look worth resolving before merge.

### Likely bug: oldest completed session wins per user, not newest

In `app/models/surveys/survey.rb:139–145`:

```ruby
@completion_sessions_by_user_id ||= SurveySession
                                   .where(survey_id: id)
                                   .completed
                                   .order(created_at: :desc)
                                   .index_by(&:user_id)
```

`index_by` overwrites on duplicate keys (officially undefined, in practice last-write-wins). With `:desc` order the newest row is iterated first and the oldest last, so the hash stores the *oldest* completed session per user. A respondent who retook a survey would have their CSV `duration_seconds` reflect the first attempt rather than the most recent. Easy to verify in a console; the fix is either `.order(:created_at).index_by(&:user_id)` or `.order(created_at: :desc).group_by(&:user_id).transform_values(&:first)`.

### N+1 on the admin list pages

`_survey.html.haml:21` (admin survey list) and `_survey_results_index_item.html.haml:16` (admin results list) call `average_duration(survey)` per row. Each call runs `SurveySession.where(survey: survey).completed.map(&:duration_in_seconds)` — one query plus full Ruby materialization per row. With N surveys, that's N queries on every visit. A preloaded `{survey_id => avg_seconds}` hash, or `AVG(TIMESTAMPDIFF(SECOND, started_at, completed_at))` in SQL, would collapse this to a single query.

### Repeated identical scans on the per-survey results page

`results.html.haml:20–47` calls six helpers. Five of them (`average`, `median`, `fastest`, `slowest`, `distribution`) each re-execute the same `SurveySession.where(...).completed.map(&:duration_in_seconds)`. `completion_rate` adds two more counts. A single load with in-memory stats would be both cheaper and easier to extend. Lifting the math into a `SurveyDurationStats` service (`app/services/`, verb-noun convention, attr_reader, perform-on-init) would let the partials and the results page share one loaded set, and would move the logic out of the global helper namespace where it's currently mixed into every view.

### `#complete` is unauthenticated and unscoped

In `app/controllers/survey_sessions_controller.rb:14–18`:

- no `before_action :require_signed_in`,
- no check that the session belongs to `current_user`,
- `update!(completed_at: Time.zone.now)` overwrites unconditionally — a retried/double-fired PUT clobbers the original timestamp and inflates the recorded duration.

Anyone with a CSRF token can `PUT /survey/complete?tracking_id=N` and complete (or re-complete) another user's session. Blast radius is polluted analytics, not destructive, but both fixes are small:

```ruby
before_action :require_signed_in

def complete
  record = SurveySession.find_by(id: params[:tracking_id], user: current_user)
  return head :not_found unless record
  record.update!(completed_at: Time.zone.now) if record.completed_at.nil?
  render json: { duration_in_seconds: record.duration_in_seconds }
end
```

Adding `require_signed_in` would also have `#start` return 401 instead of the current 500 on `nil.id` for an unauthenticated request.

### Inline styles in `results.html.haml`

The new stats table and distribution-chart section (`results.html.haml:14–47`) is full of `style: '…'` strings. The rest of `app/views/surveys/`, including the two new `_survey*.html.haml` additions in this PR, uses BEM classes. The dynamic bar widths obviously need to stay inline, but the static styling (color, height, border-radius, table padding) belongs in a small `_survey_duration.styl` module under `app/assets/stylesheets/modules/` — these are auto-imported via the `@import "modules/*"` glob in `main.styl`.

### Smaller items

- **Frontend race.** `surveyStarted` fires `request('/survey/start')` and stores `trackingId` inside a `.then`. `submitAllQuestionGroups` later checks `if (this.trackingId)` and silently skips completion when the start request hasn't resolved yet. For the analytics-completeness goal of this PR that means fast users on slow networks silently disappear from stats. Worth either awaiting the start Promise on submit or buffering the completion call.
- **PR description is stale.** It still calls the table `survey_completion_times` (renamed to `survey_sessions` mid-review). Worth refreshing before merge.
- **No FK constraints** on `survey_id`, `user_id`, `survey_notification_id`. Matches existing project patterns, so flag-not-block.

### Test gaps worth filling

- No spec asserts on the new duration column in `Survey#to_csv`. `spec/features/surveys_spec.rb:250` already invokes `to_csv` end-to-end but only checks the header prefix and the username; one extra `expect(csv).to include('duration_seconds')` plus a value assertion would also catch the `index_by` ordering bug above.
- No spec for `SurveySessionsHelper` (~60 LOC, with non-trivial median branching and bucket thresholds).
- No spec covers unauth or cross-user `#complete`.

---

*Drafted in a Claude Code session (~25 min, ~12 interactions). Sage directed the review and read the full comment before posting. Specific findings — the `index_by` ordering, the N+1 patterns, the missing auth on `#complete`, the inline-style inconsistency — are based on code reading by Claude and have not been independently verified in a console. Author should treat them as hypotheses worth checking, not authoritative claims.*

lakshita10341:
> Likely bug: oldest completed session wins per user, not newest

Hi @ragesoss , I have a doubt regarding this, for the analytics purpose should we use the first completed version when the user has seen the first time survey or latest completed version.

ragesoss:
@lakshita10341 i think latest completed version makes sense. the system allows a user to resubmit a survey... and I think this results in overwriting the previous responses. If so, it would make sense to update completion time accordingly to match what the actual responses persisted.

PR #6709, opened 2026-03-05
Exclude pre-existing content from GetRevisionPlaintext diff output

### LINKED ISSUE #6706: GetRevisionPlaintext should exclude pre-existing content when it is independent of new content
We use `GetRevisionPlaintext` to isolate the prose content of Wikipedia edits and pages, before checking that text with an AI detector. (This is because Pangram, at least, seems to work best with strictly prose content.)

Currently, we include all the text in each "diff-addedline" element from an HTML diff table in the `GetRevisionPlaintext` output. The goal is to isolate newly added content, but to retain prose structure... so for example, if an edit changes an existing sentence or heavily copyedits an existing paragraph, we want to keep the whole new version of that sentence or paragraph, so that we are not sending sentence fragments to the AI detector.  However, in many cases, new content is strictly a prose addition to an existing paragraph, but the "diff-addedline" block includes the prior content as well.

Example: https://en.wikipedia.org/w/index.php?title=Third_place&diff=prev&oldid=1340536495

In this example, the new content added to the lead paragraph is strictly a prose addition, starting after a previous complete sentence:
 ```
Scholars have noted that online forums, social media platforms, and virtual communities can function as third places by facilitating informal interaction and community building without physical proximity. These digital environments often replicate characteristics such as regular participants, conversational exchange, and shared norms, though researchers debate whether they provide the same depth of social bonding as physical spaces.
```

Rather than returning the entire lead paragraph, in this case `GetRevisionPlaintext` output should begin from "Scholars have noted".

### PR DESCRIPTION
**Pull request by archietyagi100-tech**
## What this PR does

Updates the generate_wikitext_from_diff_table method to exclude pre-existing content when new content is an independent prose addition to an existing paragraph.

## How it works

For each diff row, the fix checks if diffchange elements exist in the added line. If removing the new text from the added line matches the deleted line exactly, only the new content is kept. For rewrites or complex edits, the full line is kept (current behavior).

## Tests

Added get_revision_plaintext_spec.rb with 9 test cases:
 Independent addition at end of paragraph
- Independent addition at beginning of paragraph
- Rewrite of existing content
- Entirely new paragraph
- Multiple rows with mixed changes
- Word change in middle of sentence
- New sentences with references
- Empty deleted line
- No diffchange markers

## AI usage

Used AI for understanding Ruby syntax and codebase structure.

Closes #6706

### DIFF
--- a/app/services/get_revision_plaintext.rb
+++ b/app/services/get_revision_plaintext.rb
@@ -77,13 +77,125 @@ class GetRevisionPlaintext
     @diff_table = resp.data['*']
   end
 
+  # Parses the diff table HTML and extracts only the
+  # changed wikitext, excluding pre-existing content
+  # when new text is an independent sentence addition.
+  # Parses the diff table HTML and extracts only the
+  # changed wikitext, excluding pre-existing content
+  # when new text is an independent sentence addition.
   def generate_wikitext_from_diff_table
     doc = Nokogiri::HTML(@diff_table)
-    # Collect all the '.diff-addedline' table cells.
-    # These represent all the sections of wikitext that were either added or changed,
-    # IE, the right side of a traditional wikitext diff table without the unchanged
-    # .diff-context cells.
-    @changed_wikitext = doc.css('.diff-addedline').map(&:text).reject(&:empty?).join("\n\n")
+    @changed_wikitext = extract_changed_parts(doc).reject(&:empty?).join("\n\n")
+  end
+
+  # Iterates over each <tr> in the diff table.
+  # See https://en.wikipedia.org/w/index.php?title=Third_place&diff=prev&oldid=1340536495
+  # The doc contains a table like:
+  #
+  #   <table>
+  #     <tr>
+  #       <td class="diff-deletedline diff-side-deleted">
+  #         <div>In sociology...refers to...</div>
+  #       </td>
+  #       <td class="diff-addedline diff-side-added">
+  #         <div>In sociology...refers to...
+  #         <ins class="diffchange">Scholars have
+  #         noted...</ins></div>
+  #       </td>
+  #     </tr>
+  #     <tr>
+  #       <td class="diff-empty"></td>
+  #       <td class="diff-addedline diff-side-added">
+  #         <div>== Criticism ==</div>
+  #       </td>
+  #     </tr>
+  #   </table>
+  #
+  # Returns an array of extracted text from each <tr>.
+  def extract_changed_parts(doc)
+    doc.css('tr').filter_map do |row|
+      extract_row_content(row)
+    end
+  end
+
+  # Extracts new content from a single diff <tr>.
+  # Example from Third_place diff (oldid=1340536495):
+  #
+  #   <td class="diff-deletedline diff-side-deleted">
+  #     <div>In sociology, the '''third place''' refers
+  #     to the social surroundings...</div>
+  #   </td>
+  #   <td class="diff-addedline diff-side-added">
+  #     <div>In sociology, the '''third place''' refers
+  #     to the social surroundings...
+  #     <ins class="diffchange">Scholars have noted that
+  #     online forums, social media platforms, and virtual
+  #     communities can function as third places...</ins>
+  #     </div>
+  #   </td>
+  #
+  # The old paragraph text is unchanged. The new sentences
+  # are inside <ins class="diffchange">. So this method
+  # returns only the <ins> content: "Scholars have noted..."
+  # For rows where text was rewritten (old != new without
+  # <ins>), the full .diff-addedline text is returned.
+  def extract_row_content(row)
+    added = row.at_css('.diff-addedline')
+    deleted = row.at_css('.diff-deletedline')
+    return nil if added.nil? || added.text.strip.empty?
+
+    ins_elements = added.css('ins.diffchange')
+
+    if independent_addition?(added, deleted, ins_elements)
+      ins_elements.map(&:text).join(' ')
+    else
+      added.text
+    end
+  end
+
+  # Returns true when the edit is a pure sentence-level
+  # addition. From the Third_place diff (oldid=1340536495):
+  #
+  #   <td class="diff-deletedline diff-side-deleted">
+  #     <div>In sociology...refers to...</div>
+  #   </td>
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by ArchieTyagi** [comment]
Thanks for the feedback I'll refactor the code into smaller methods and
rewrite the tests to use the public interface with stubbed API responses.
Will update the PR soon.

**Comment by ArchieTyagi** [comment]
Updated based on your feedback  refactored into smaller methods (extract_changed_parts, extract_row_content, independent_addition, text_without_ins) and rewrote tests to use the public interface with stubbed API responses. Ready for review

**Comment by ragesoss** [comment]
This looks like it's heading in the right direction!

I think each of the new methods would benefit from comments that have brief examples to illustrate the kind of structured text they are designed to operate on. We really want to optimize this class for clarity and understandability.

File: spec/services/get_revision_plaintext_spec.rb [hunk 1]
```diff
@@ -0,0 +1,197 @@
...
+require 'rails_helper'
+
+describe GetRevisionPlaintext do
+  let(:en_wiki) { Wiki.default_wiki }
+
+  def mock_response(data)
+    OpenStruct.new(data: data)
+  end
+
+  def stub_diff_api(diff_html)
```
ragesoss, line 12: I think this is a good strategy for isolating specific behaviors in the plaintext processing. I think it should also include at least one or two complex examples of real diffs where the API response is not stubbed (and VCR is used to record the real API responses).

ArchieTyagi:
Added comments with examples to each new method and a VCR based test using the real Third_place diff from the issue. 
All tests passing locally including rubocop.

app/services/get_revision_plaintext.rb
```diff
@@ -79,11 +79,68 @@ def fetch_diff_table
...
-    # These represent all the sections of wikitext that were either added or changed,
-    # IE, the right side of a traditional wikitext diff table without the unchanged
-    # .diff-context cells.
-    @changed_wikitext = doc.css('.diff-addedline').map(&:text).reject(&:empty?).join("\n\n")
+    @changed_wikitext = extract_changed_parts(doc).reject(&:empty?).join("\n\n")
+  end
+
+  # Iterates over diff table rows and extracts the relevant
+  # new content from each row.
+  # Example input: a diff table with rows like:
```
ragesoss: These examples don't seem to match up with the methods. For example, this method is designed to work on a diff table that includes multiple `tr` elements. The comment should use a real example, and include a link to the example diff on Wikipedia that it came from.

ArchieTyagi:
Updated comments to reference the real Third_place diff (oldid=1340536495) with links, showing how each method handles that specific example.

ragesoss:
Those comments still don't have the example details I'd like to see: the actual HTML structures from the example diff as they are transformed by each method.

For example `extract_row_content` takes a `row` argument that has `.diff-addedline`, `'.diff-deletedline`, and `ins.diffchange` elements within it. The comment should have an example row that is part the referenced real diff, and that contains all those elements.

ArchieTyagi:
Updated extract_row_content comment with actual HTML from the Third_place diff showing the real diff-deletedline, diff-addedline, and ins.diffchange structure.

ragesoss:
@archietyagi100-tech that looks good, and i like the way you've trimmed the extra text with ellipses but kept enough to clearly match with the example diff. Please do the same for the other methods. I want to be able to easily visualize the data structures that each one of these small methods works on.

ArchieTyagi:
Added real HTML examples from the Third_place diff to all methods, matching the same style as extract_row_content.

app/services/get_revision_plaintext.rb
```diff
@@ -79,11 +79,78 @@ def fetch_diff_table
...
+  # Determines whether the new content in a diff row is an
+  # independent prose addition — meaning the pre-existing text
+  # is unchanged and the new text is purely additive.
+  # Example where this returns true:
+  #   deleted: "Cats are mammals."
+  #   added:   "Cats are mammals. <ins>They are popular pets.</ins>"
+  # Example where this returns false (rewrite):
+  #   deleted: "The <del>building</del>."
+  #   added:   "The <ins>structure</ins>."
+  def independent_addition?(added, deleted, ins_elements)
```
ragesoss, line 172: This one is particularly important to have clear, realistic examples in the comments for: one where there non-independent additions, and one (the Third_place example) where it's independent additions.

ragesoss: I think this will also require some functional changes. The idea here is that we want to retain full sentences and avoid ending up with additions that are text fragments that don't make sense without the pre-existing text. For example, if a new phrase was added to the end of a sentence, that would be strictly show up an an `ins`, but we would want the resulting plaintext to include the entire sentence, not just the portion that was added.

This is a case that should include a real example in the tests.

ArchieTyagi: Both addressed in the latest push  independent_addition? now has real Third_place HTML examples for both cases (independent addition and non-independent). Also added a sentence-level check to keep full sentences when a phrase is added mid-sentence (starts lowercase), with a test for that case.

app/services/get_revision_plaintext.rb
```diff
@@ -77,13 +77,123 @@ def fetch_diff_table
...
+  # extracting mid-sentence fragments.
+  def independent_addition?(added, deleted, ins_elements)
+    return false unless ins_elements.any?
+    return false if deleted.nil? || deleted.text.strip.empty?
+    return false unless text_without_ins(added).strip == deleted.text.strip
+
+    # Only treat as independent if the new content starts
+    # a full sentence (avoids mid-sentence fragments)
+    first_ins_text = ins_elements.first.text.strip
+    first_ins_text.match?(/\A[A-Z]/)
```
ragesoss: Is this just checking for capitalization of the first letter, as a proxy for being the start of sentence?

I think we need to rely on punctuation as well.

ArchieTyagi: Fixed Now checking both preceding punctuation and capitalization  preceding.match?(/[.!?]['""]?\z/) && first_ins_text.match?(/\A[A-Z]/)

ragesoss:
Okay! I think this is a nice step forward, and will be in good shape for others to build upon as we add more requirements.

PR #6711, opened 2026-03-05
Batches multiple errors and limits Sentry samples for reference_count…

### LINKED ISSUE #5755: ReferenceCounterApi generates too many Sentry logs
For non-200 responses, we generate a Sentry log "Non-200 response hitting references counter API". When the system has a problem, as it did in late March, this floods Sentry with these logs and can quickly deplete our error logging quota (resulting in the dropping of logs until the next monthly billing cycle starts).

I guess we should either consolidate these logs better so that fewer separate events are logged (preferable) or disable sending this particular type of logs to Sentry.

Formasitchijoh:
## What this PR does
This PR resolves #5755 where the Sentry log flooding and quota depletion issue caused by non-200 responses from the reference-counter Toolforge API.  It refactors ReferenceCounterApi to handle batch failures.
Instead of triggering a separate Sentry warning for every failed revision ID which can reach hundreds per request—this implementation batches errors by HTTP status code.

**Key improvements include:**
`Batch Reporting:` Collects all API errors during the `get_number_of_references_from_revision_ids` loop and sends a single summary message to Sentry at the end of the batch.

`Quota Protection: ` Limits the errors array to a maximum of 5 examples per batch while still reporting the total_errors_count_in_batch to maintain visibility into the true scale of the failure.

`Sentry Fingerprinting:` Implements custom fingerprinting `['references-counter-api-error', status.to_s]` to ensure that multiple background jobs collapse into a single grouped issue in the Sentry UI, preventing dashboard noise.

`Data Integrity: `Ensures the method still returns nil for failed revisions, allowing the application to continue processing without a hard crash.

## AI usage
Used AI to  optimize the RSpec mocks to verify that the Sentry payload is strictly limited at 5 samples even when the "API flood" exceeds that limit.


ragesoss:
Hmm... I wish I had recorded more data about the flood of errors when that happened. We don't have access to Sentry logs that old, so I can't see exactly what was going on.

I see that we have logging of actual errors currently (in batches, via the rescue), but that at some point the logging of non-200 responses was commented out, so basically we are silently failing and returning `nil` for the reference count in cases where it's a non-200 response, and not capturing any data in Sentry about how often this happens.

Since we already have the batch-level error reporting in this class, I think it would make the most sense to extend the current error logging to handle those non-200 responses. So, instead of creating a new data structure (`@sentry_logs`), add an entry for each non-200 response into the same `@errors` data structure we're already using for actual errors.

@gabina what do you think?

gabina:
> Hmm... I wish I had recorded more data about the flood of errors when that happened. We don't have access to Sentry logs that old, so I can't see exactly what was going on.

Well, it looks like we have a 27GB-sized file called `uwsgi.log` in `/data/project/reference-counter` that we can check to find what happened two years ago.

Without checking that, according to what I remember, shortly after the new reference-counter API was up and running,  the queue got full and all the requests failed. As far as I remember, that happened two or three times, and it was fixed simply by restarting the server. I guess that the first time that error occurred, it exceeded the logging quota and Sage just commented out the logging for non-200 responses. 

I was checking the last 100000 lines of the `uwsgi.log` file to see which kind of non-200 responses we usually have:

**Classified as non-transient errors** in `ReferenceCounterApi` (we don't want to reprocess a timeslice for these kind of errors):
- **400 responses**. These are caused mainly for languages/projects that the reference-counter API currently doesn't support. For example, `gpe` language or `incubator` project. We could evaluate adding them, unsure if it makes sense.
- **403 responses**. These are caused mainly for deleted revisions. Example: [1343128406](https://en.wikipedia.org/?oldid=1343128406)
- **404 responses**. These are caused by revisions that are not found (example: [932566719](https://en.wikipedia.org/?oldid=932566719)) and other routes like `/robots.txt`.


**NOT classified as non-transient errors** (we retrieve timeslices with these errors)
- **500 responses**.
- **502 responses**.


> Since we already have the batch-level error reporting in this class, I think it would make the most sense to extend the current error logging to handle those non-200 responses. So, instead of creating a new data structure (@sentry_logs), add an entry for each non-200 response into the same @errors data structure we're already using for actual errors.

Yes, I agree that reusing the `@error` makes total sense.
I'm not sure about adding each non-200 response to `@error`. Maybe it's not worth to logging non-transient errors, since they're expected and there is not much we can do about them. It might be useful to add logs for 400 responses to detect missing languages/projects. But for 403 and 404 responses, maybe it's better to try to find a way to avoid fetching scores for deleted revisions (not as part of this issue, of course).
I guess that we can add logs for all kind of errors for now, and then see how we can fix some of them.

Formasitchijoh:
Thanks for the feedback 

To summarize the requested changes I'm planning to implement:

- Keep the current batching approach (collecting errors in batches of ~50–60) so we can estimate the scale of issues per status code (400, 500, etc.) without logging every single error individually  which would risk exceeding logging quotas.
- Use the `@error` data structure to accumulate errors during the batch, then send a single summarized Sentry message per batch.

Here's how I'm currently capturing the error :

```ruby
Sentry.capture_message(
  "Non-200 response from references counter API: #{status}",
  level: :warning,
  # Groups all errors with the same status code together
  fingerprint: ['references-counter-api-error', status.to_s],
  extra: {
    project_code:    @project_code,
    language_code:   @language_code,
    error_count:     data[:error_count],   # e.g. 53 — number of errors in this batch
    errors:          data[:errors]         # array of error details (e.g. short messages or codes)
  }
)
```
I'd really appreciate your thoughts on the extra structure, specifically

> 
> It might be useful to add logs for 400 responses to detect missing languages/projects. But for 403 and 404 responses, maybe it's better to try to find a way to avoid fetching scores for deleted revisions (not as part of this issue, of course).
> I guess that we can add logs for all kind of errors for now, and then see how we can fix some of them.

 And from  @gabina  suggestion Subsequenty I can open new PR's on how to  solve these error.

gabina:
Can't you re-use this method to log to Sentry? That way you don't need to redefine the extra structure. Note the `log_error` comes from the  `ApiErrorHandling` module.
```
  def log_error_batch(rev_ids)
    return if @errors.empty?

    log_error(@errors.first, update_service: @update_service,
    sentry_extra: { rev_ids:, project_code: @project_code,
    language_code: @language_code, error_count: @errors.count })
  end
```

That said, I like your batching per status code approach, but I think you should find a way to put all in `@errors` and then make a `log_error` call for each different kind of error. Does it make sense?

Formasitchijoh:
> Can't you re-use this method to log to Sentry? That way you don't need to redefine the extra structure. Note the `log_error` comes from the `ApiErrorHandling` module.
> That said, I like your batching per status code approach, but I think you should find a way to put all in `@errors` and then make a `log_error` call for each different kind of error. Does it make sense?

It does makes sense, and I think is a better way to go about it. I will work on the suggested feedback.

lib/reference_counter_api.rb
```diff
@@ -25,7 +25,7 @@ def initialize(wiki, update_service = nil)
     @language_code = wiki.language
     @update_service = update_service
     @errors = []
-    @sentry_logs = {}
+    @non_200_errors = {}
```
Formasitchijoh, line 28: I decided to keep a separate @sentry_logs hash rather than merging these into the flat @errors array for two reasons:

Grouping: @errors tracks individual transient exceptions (like timeouts). For `non-200 responses`,  we need to group by status code so that a batch of `403s` and` 400s` don't get mixed together in a single Sentry log.

Efficient Counting: The hash allows us to increment `error_count` and limit the errors samples to 5 per status code efficiently. If we used a flat array, we would have to loop and group them at the end anyway.

I've renamed the variable to @non_200_errors to better reflect that these are API-level errors being tracked alongside the connection-level `@errors`

Formasitchijoh:
@ragesoss I think this is ready

ragesoss:
It seems a little complicated for my taste, but I'm willing to give it a try.

PR #6204, opened 2025-02-21
Migrate Admin Notes CRUD from Redux to RTK Query & Set Up React-Redux Integration Testing

Abishekcs:
## What this PR does

This PR refactors the Admin Notes CRUD functionality by migrating from traditional Redux to [`RTK Query`](https://redux-toolkit.js.org/rtk-query/overview). It also sets up [`React-Redux integration testing`](https://testing-library.com/).

Also sets up testing for **React Testing Library** and [MSW](https://mswjs.io/), which is a **one-time setup** to ensure a consistent testing environment across the project.

**Node.js**: Changed Node version to 16 in `CI workflow` which resolves Jest test failures after adding @reduxjs/toolkit Query features. Test failures were caused by nullish coalescing assignment operator (??=) which isn't supported in Node.js 14 used by CI, while tests passed locally with newer Node versions.

**Minor BugFix:** Admin receiving redundant notifications for notes they’ve edited/updated.

**Link to Refrences & Blogs**
- [Writing Tests](https://redux.js.org/usage/writing-tests)
- [The Evolution of Redux Testing Approaches](https://blog.isquaredsoftware.com/2021/06/the-evolution-of-redux-testing-approaches/)
- [Kent C. Dodds: Testing Implementation Details](https://kentcdodds.com/blog/testing-implementation-details)
- [Kent C. Dodds: Stop mocking fetch](https://kentcdodds.com/blog/stop-mocking-fetch)
- [The tests of RTK query itself.](https://github.com/reduxjs/redux-toolkit/blob/64a30d83384d77bcbc59231fa32aa2f1acd67020/packages/toolkit/src/query/tests/buildHooks.test.tsx#L95-L114)

---

## RTK Query: 
With RTK Query, the Admin Notes component won't repeatedly call the API endpoint when it mounts and unmounts, as demonstrated in the before and after screenrecording under the `RTK Query DevTool header screenshot below`. The cache duration for the endpoint follows RTK Query's default setting, which is 60 seconds.

---

## **How RTK Query Stores Fetched Data**  

When data is fetched using **RTK Query**, it is automatically **stored and managed** by RTK in the **API slice** defined in the Redux store.

- **Data is stored in the cache** under the API slice name (Filename: apiSlice.js), which is defined in `createApi({ reducerPath: 'api', ...})`.  
- **Cached responses prevent unnecessary re-fetching**, reducing API calls and improving performance.  
- The default cache duration is **60 seconds**, but it can be customized using `keepUnusedDataFor`.  
[...]

PR #6713, opened 2026-03-07
Upgrade From Jest 26.0.1 to 30.0.0

Abishekcs:
## What this PR does

- Upgrades Jest to current version i.e `30.0.0` as  new version of Redux and RTK requires newer version of Jest and currently the dashboard is using `Jest 26.0.1` #6204. 

- This PR will remove the amount of `code that was added by me to maintain both version of Jest 26 and Jest 30 in PR #6204`. And will make upgrading `Redux` and adding `RTK` **Much Easier.** 

- Change Node.js version from 14 to 18 in CI as Jest 30 dropped support for older Node versions and requires Node 18 as the minimum.

- Add `jest-environment-jsdom@^30.2.0` to devDependencies. `Jest 30 removed jsdom` as the default test environment to  avoid performance overhead on projects that don't need it `(For more details check Under Performance header below)`

- Add `testEnvironment: 'jsdom'` to **jest.config.js** to restore browser-like environment for tests and added Jest 30's `globalsCleanup` to prevent global state from leaking between test files.

- Fix test matchers that no longer work in Jest 30.
  - `test/reducers/campaign.spec.js`
  - `test/utils/course_date_utils.spec.js`

## Refrences

- [Jest 30: Faster, Leaner, Better](https://jestjs.io/blog/2025/06/04/jest-30)
- [Globals cleanup between test files](https://jestjs.io/blog/2025/06/04/jest-30#globals-cleanup-between-test-files)
- [Jest 27: New Defaults for Jest, 2021 edition](https://jestjs.io/blog/2021/05/25/jest-27)

Node Js

- [testEnvironment [string]](https://jestjs.io/docs/configuration#testenvironment-string)

image

All of `Dashboard JS code` test cases depend on `jsdom` so we can't use `Node` as `testEnvrionment` which is the **default** now starting `Jest 27` as mentioned in **Jest Docs above**.  All test cases will fail if `jsdom` not used by `testEnvitonment` shown below.

image

## Performance

Jest 30 removed `jsdom` as the default test environment to eliminate performance overhead for projects that don't need it. Since all dashboard tests require a browser-like environment, we must explicitly opt in via `jest-environment-jsdom.` This means no performance improvement for us from that specific Jest 30 change, but also no regression the behavior is equivalent to what Jest 26 provided out of the box.

## AI usage

No AI usage.

## Screenshots

Before:

https://github.com/user-attachments/assets/22f1ad4f-25a9-4136-8871-210c1ab6a90c

After:

https://github.com/user-attachments/assets/2682ab37-e3c1-4a3b-bd6c-4e9c7e228ec1

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -3,6 +3,13 @@
 module.exports = {
   collectCoverage: false,
   testEnvironment: 'jsdom',
+
+  testEnvironmentOptions: {
+    // Prevent global state (e.g. window/global mutations) from leaking between test files
+    // https://jestjs.io/blog/2025/06/04/jest-30#globals-cleanup-between-test-files
+    globalsCleanup: 'on',
+  },
+
   roots: [
     'test',
     'app/assets/javascripts'


ragesoss:
This looks reasonable. Thanks @Abishekcs!

PR #6715, opened 2026-03-07
hamburger menu disappearance fix

### LINKED ISSUE #6714: Mobile hamburger menu disappeared on some tabs of campaign
On mobile devices, navigating to campaign tabs causes the mobile hamburger menu to disappear for some tabs like for  "Courses" . Instead, the desktop navigation links (Explore, My Dashboard, Language Switcher, Log out) are squeezed into the header.

### Steps to Reproduce
1. Open a Campaign's **Courses** (WikiEdu) or **Programs** (Outreach) tab on a mobile device or Chrome DevTools mobile view.
2. Look at the top navigation bar.
3. Observe the bug: The hamburger menu is gone for "courses" tab, replaced by squished desktop text links.
4. Navigate to the **Students** or **Alerts** tabs in the same campaign.
5. Observe the expected behavior: The hamburger menu works correctly.
 

### Screenshot

Image

nexpectArpit:
This PR resolves an issue where the mobile hamburger menu disappears on certain campaign tabs (`Courses/Programs` and `Articles`). Instead of the mobile menu, the desktop navigation links are squeezed into the header, causing layout overlapping.

Closes #6714

#### 1. Before (Program Page)

| | |
|---|---|
|  |  |

#### 2. Article Page

| Before | After Fix |
|---|---|
|  |  |

#### 3. Before (Courses Tab)

| Before | After Fix |
|---|---|
| Screenshot 2026-03-12 at 8 35 52 PM |  |

@ragesoss please review it

### DIFF
--- a/app/assets/javascripts/components/common/list.jsx
+++ b/app/assets/javascripts/components/common/list.jsx
@@ -137,9 +137,9 @@ const List = ({
     );
   }
   let fixedHeader;
-  let fixedArea;
+  let fixedArea = 'table-responsive';
   if (stickyHeader) {
-    fixedArea = 'persist-area';
+    fixedArea += ' persist-area';
     const fixheader = fixHeader === true ? 'floatingHeader' : '';
     fixedHeader = 'persist-header';
     fixedHeader += ` ${fixheader}`;
--- a/app/assets/stylesheets/modules/_tables.styl
+++ b/app/assets/stylesheets/modules/_tables.styl
@@ -196,6 +196,11 @@
   +below(tablet)
     width max-content
 
+.table-responsive
+  display block
+  overflow-x auto
+  -webkit-overflow-scrolling touch
+
 .revision-list-container
   +below(tablet)
     overflow-y: scroll
--- a/app/views/campaigns/articles.html.haml
+++ b/app/views/campaigns/articles.html.haml
@@ -25,7 +25,7 @@
           %option{rel: 'asc', value: 'course_title'}
             = t("#{@presenter.course_string_prefix}.courses")
 
-    %table.table.table--hoverable.table--sortable
+    %table.table.table--hoverable.table--sortable.table-responsive
       %thead
         %tr
           %th.sort.sortable{'data-default-order' => 'asc', 'data-sort' => 'title'}
--- a/app/views/campaigns/programs.html.haml
+++ b/app/views/campaigns/programs.html.haml
@@ -33,7 +33,7 @@
           %option{rel: 'asc', value: 'start-date'}
             = t("courses.start_date")
 
-    %table.table.table--hoverable.table--sortable
+    %table.table.table--hoverable.table--sortable.table-responsive
       %thead
         %tr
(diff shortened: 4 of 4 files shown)


ragesoss:
Please include before/after screenshots, and include screenshot examples for all three of the views/components being changed.

app/assets/stylesheets/modules/_tables.styl
```diff
@@ -196,6 +196,10 @@
   +below(tablet)
     width max-content
 
+.table-responsive
```
ragesoss, line 199: It looks like this change is intended to fix a mobile layout bug, but this CSS is not scoped to mobile. Should it be? If not, how have you tested for regressions / unwanted layout changes in non-mobile views?

nexpectArpit: Yes, it absolutely should be scoped to mobile. The viewport stretching bug is exclusively a mobile issue.
I will update the code in my latest commit ,so that the `overflow-x: auto` rule inside our `.table-responsive` class only activates `+below(tablet)`.

app/assets/javascripts/components/common/list.jsx
```diff
@@ -145,7 +145,7 @@ const List = ({
     fixedHeader += ` ${fixheader}`;
   }
   return (
-    <div className={fixedArea}>
+    <div className={`${fixedArea || ''} table-responsive`.trim()}>
```
ragesoss: What is this `.trim()` for?

app/views/campaigns/articles.html.haml
```diff
@@ -25,57 +25,58 @@
...
-    %table.table.table--hoverable.table--sortable
-      %thead
-        %tr
-          %th.sort.sortable{'data-default-order' => 'asc', 'data-sort' => 'title'}
-            = t('articles.title')
-            %span.sortable-indicator
-          %th.sort.sortable.desc{'data-default-order' => 'desc', 'data-sort' => 'char_added'}
-            .tooltip-trigger
-              = t('metrics.char_added')
+    .table-responsive
```
ragesoss: It it necessary to add this wrapping div, or could `.table-responsive` be added to the table element itself? (That would make this changeset a lot prettier, since only one line per file would need to be changed.)

app/assets/javascripts/components/common/list.jsx
ragesoss: Maybe instead of using trim, `fixedArea` can be created initially as `table-responsive`, and then ' persist-area' could be appended in the `if (stickyHeader)` block above?

nexpectArpit: > What is this `.trim()` for?

I was using .trim() defensively to prevent a leading space when fixedArea was empty, but initializing it properly as let fixedArea = 'table-responsive'; completely eliminates the need for string manipulation. I will refractor the logic in updated pr.

app/views/campaigns/articles.html.haml
nexpectArpit: implemented

ragesoss:
LGTM

PR #6717, opened 2026-03-09
Fix: Article development plot (ArticleGraphs) is empty in timeslice system

### LINKED ISSUE #6337: Article development plot (ArticleGraphs) is empty in timeslice system
### What is happening?
`articles/article_data.json?article_id=780686990` endpoint depends on raw revisions (see `app/views/articles/article_data.json.jbuilder` file), making the "Article development" graph empty in the new system.

### To Reproduce

> 1. Go to course [250 by 2026-3](https://dashboard.wikiedu.org/courses/Wiki_Education/250_by_2026-3_(Spring_2025))
> 2. Click on Articles tab
> 3.  Click on "Article development"
> 4. See the empty chart:

Image

### Expected behavior
See a complete article plot

Image

Image

### Additional context
This bug comes from the migration to the new timeslices system. Related to issue #

Formasitchijoh:
## What this PR does
This PR resolves #6337  by implementing a `revision_score` controller method. The new method accepts an article_id, looks up the associated course, fetches all revisions on that article between the course’s start and end dates and retrieves the `wp10` score for each revision using the `LiftWingApi`. The data is returned in the exact format that the `article_graph` requires for `plotting` revision scores. This functionality was moved to the server to prevent bloating the frontend in case of too many revisions and also ensure the data can be cached to reduce round trip time for the same resource. 

## Screenshots
Before: 

image

After:
https://www.loom.com/share/aeace9f1105d40758de474cbdc4859f1

@gabina Could you take a look when you are chanced

### DIFF
--- a/app/assets/javascripts/components/articles/article.jsx
+++ b/app/assets/javascripts/components/articles/article.jsx
@@ -7,6 +7,7 @@ import DiffViewer from '../revisions/diff_viewer.jsx';
 import Switch from 'react-switch';
 import { toWikiDomain } from '../../utils/wiki_utils.js';
 import { stringify } from 'query-string';
+import ArticleGraphs from './article_graphs.jsx';
 
 const Article = ({ article, index, course, fetchArticleDetails, updateArticleTrackedStatus, articleDetails, wikidataLabel,
   showOnMount, setSelectedIndex, lastIndex, selectedIndex, pageLogsMessage, deletedMessage, current_user }) => {
@@ -94,9 +95,7 @@ const Article = ({ article, index, course, fetchArticleDetails, updateArticleTra
           {!isDeleted
             ? (
               <small>
-                {/* Disabling article development plot feature until re-implementing. See issue #6337 */}
-                {/* <a href={historyUrl} target="_blank" className="inline">{I18n.t('articles.history')}</a> | <ArticleGraphs article={article} /> */}
-                <a href={historyUrl} target="_blank" className="inline">{I18n.t('articles.history')}</a>
+                <a href={historyUrl} target="_blank" className="inline">{I18n.t('articles.history')}</a> | <ArticleGraphs article={article} course_id={course.id}/>
               </small>
             )
             : (
--- a/app/assets/javascripts/components/articles/article_graphs.jsx
+++ b/app/assets/javascripts/components/articles/article_graphs.jsx
@@ -4,7 +4,7 @@ import EditSizeGraph from './edit_size_graph.jsx';
 import Loading from '../common/loading.jsx';
 import request from '../../utils/request.js';
 
-const ArticleGraphs = ({ article }) => {
+const ArticleGraphs = ({ article, course_id }) => {
   const { id: article_id } = article;
 
   const [showGraph, setShowGraph] = useState(false);
@@ -39,7 +39,7 @@ const ArticleGraphs = ({ article }) => {
     if (articleData) {
       return;
     }
-    const articledataUrl = `/articles/article_data.json?article_id=${article_id}`;
+    const articledataUrl = `/articles/${article_id}/revision_score?course_id=${course_id}`;
     request(articledataUrl)
       .then(resp => resp.json())
       .then((data) => {
--- a/app/controllers/articles_controller.rb
+++ b/app/controllers/articles_controller.rb
@@ -1,12 +1,19 @@
 # frozen_string_literal: true
+require_dependency "#{Rails.root}/lib/revision_score_manager"
 
 class ArticlesController < ApplicationController
   respond_to :json
-  before_action :set_course, except: :article_data
+  before_action :set_course
 
   # returns revision score data for vega graphs
-  def article_data
+  #  { rev_id: 123, characters: 1000, wp10: 0.5, date: '2021-01-01', username: 'Example' }
+  def revision_score
     @article = Article.find(params[:article_id])
+    rev_manager = RevisionScoreManager.new(@article, @course)
+    enrolled_usernames = @course.users.pluck(:username)
+    # root: false prevents wrap_parameters from wrapping the array under the
+    # controller name ("articles"), which would leak into the response.
+    render json: rev_manager.fetch_scored_revisions(enrolled_usernames), root: false
   end
 
   # returns details about how an article changed during a course
@@ -28,4 +35,5 @@ class ArticlesController < ApplicationController
   def set_course
     @course = Course.find(params[:course_id])
   end
+
 end
--- a/lib/revision_score_manager.rb
+++ b/lib/revision_score_manager.rb
@@ -1,0 +1,103 @@
+# frozen_string_literal: true
+require_dependency "#{Rails.root}/lib/lift_wing_api"
+
+# Handles fetching and scoring article revisions for course-specific graphs.
+class RevisionScoreManager
+  def initialize(article, course)
+    @article = article
+    @course = course
+    @wiki = article.wiki
+  end
+
+  MAX_REVISION_COUNT = 50
+
+  def fetch_scored_revisions(enrolled_usernames)
+    revisions = get_student_revisions(enrolled_usernames)
+    return [] if revisions.empty?
+    cached_scored_revisions(revisions.first(MAX_REVISION_COUNT))
+  end
+
+  private
+
+  # Use the ID of the most recent revision as the cache key
+  # If this ID changes, the cache automatically invalidates.
+  def cache_key(recent_revision_id)
+    "article_#{@course.id}_#{@article.mw_page_id}_#{recent_revision_id}"
+  end
+
+  # Caches scored revisions to avoid hitting LiftWing API repeatedly for the same data
+  def cached_scored_revisions(target_revisions)
+    Rails.cache.fetch(cache_key(target_revisions.first[:revid]), expires_in: 1.day) do
+      rev_ids = target_revisions.map { |r| r[:revid] }
+      scores = score_revisions(rev_ids)
+
+      target_revisions.map do |rev|
+        {
+          rev_id: rev[:revid],
+          characters: rev[:size],
+          wp10: scores.dig(rev[:revid].to_s, 'wp10'),
+          date: rev[:timestamp],
+          username: rev[:user]
+        }
+      end
+    end
+  end
+
+  # Filters the total revision to a manageable number for plotting.
+  # We explicitly take the most recent revisions (up to MAX_REVISION_COUNT)
+  # to ensure the graph remains performant and readable.
+  def get_student_revisions(enrolled_usernames)
+    params = query_params
+    student_revisions = []
+
+    loop do
+      response = WikiApi.new(@wiki).query(params)
+      page_data = response['query']['pages'][@article.mw_page_id.to_s]
+
+      process_revisions(page_data['revisions'], enrolled_usernames, student_revisions)
+
+      # Exits loop if there is no continue parameter in response or
+      # Stop immediately if we found 50 student revisions
+      break if student_revisions.size >= MAX_REVISION_COUNT || !response['continue']
+
+      # When the API returns a continue hash, pull out its
+      # `rvcontinue`/`continue` tokens and merge them into `params`.
+      # That mutates the query parameters so the next loop iteration
+      # will request the next page of revisions.
+      params.merge!(response['continue'].slice('rvcontinue', 'continue'))
+    end
+    student_revisions
+  end
+
+  def process_revisions(revisions, usernames, results)
+    (revisions || []).each do |r|
+      next unless usernames.include?(r['user'])
+
+      results << {
+        revid: r['revid'],
+        user: r['user'],
+        timestamp: r['timestamp'],
+        size: r['size']
+      }
+    end
+  end
+
+  def score_revisions(rev_ids)
+    LiftWingApi.new(@wiki).get_revision_data(rev_ids)
+  end
+
+  # Queries for revisions made within the course period
+  # [course.start, course.end]
+  def query_params
+    {
+      action: 'query',
+      prop: 'revisions',
+      pageids: @article.mw_page_id,
+      rvprop: 'user|ids|timestamp|size',
+      rvstart: @course.end.strftime('%Y%m%d%H%M%S'),
+      rvend: @course.start.strftime('%Y%m%d%H%M%S'),
+      rvdir: 'older', # List newest first. rvstart has to be later than rvend.
+      rvlimit: 500
+    }
+  end
+end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -246,7 +246,7 @@ Rails.application.routes.draw do
   post 'courses/:course_id/disable_timeline' => 'timeline#disable_timeline',
        constraints: { course_id: /.*/ }
 
-  get 'articles/article_data' => 'articles#article_data'
+  get 'articles/:article_id/revision_score' => 'articles#revision_score'
   get 'articles/details' => 'articles#details'
   post 'articles/status' => 'articles#update_tracked_status'
 
--- a/spec/controllers/articles_controller_spec.rb
+++ b/spec/controllers/articles_controller_spec.rb
@@ -6,7 +6,7 @@ describe ArticlesController, type: :request do
   let(:article) { create(:article) }
   let(:user) { create(:user) }
   let(:second_user) { create(:user, username: 'SecondUser') }
-  let(:course) { create(:course) }
+  let(:course) { create(:course, start: Time.zone.now, end: Time.zone.now + 1.week) }
 
   before do
     create(:courses_user, user_id: user.id, course_id: course.id)
@@ -15,9 +15,9 @@ describe ArticlesController, type: :request do
                              user_ids: [user.id, second_user.id])
   end
 
-  describe '#article_data' do
+  describe '#revision_score' do
     it 'sets the article from the id' do
-      get '/articles/article_data', params: { article_id: article.id, format: :json }
+      get "/articles/#{article.id}/revision_score", params: { course_id: course.id, format: :json }
       expect(assigns(:article)).to eq(article)
     end
   end
--- a/spec/lib/revision_score_manager_spec.rb
+++ b/spec/lib/revision_score_manager_spec.rb
@@ -1,0 +1,122 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_dependency "#{Rails.root}/lib/revision_score_manager"
+
+describe RevisionScoreManager do
+  let(:article) { create(:article, mw_page_id: 46721259) }
+  let(:course) { create(:course, start: '2023-01-01', end: '2023-02-01') }
+  let(:manager) { described_class.new(article, course) }
+
+  describe '#fetch_scored_revisions' do
+    let(:wiki_api_response) do
+      {
+        'query' => {
+          'pages' => {
+            '46721259' => {
+              'revisions' => [
+                { 'revid' => 1189_344_512, 'user' => 'RageSock',
+                  'timestamp' => '2023-01-10T12:00:00Z', 'size' => 42_100 },
+                { 'revid' => 1189_201_347, 'user' => 'WikiedStaff',
+                  'timestamp' => '2023-01-09T08:30:00Z', 'size' => 41_800 }
+              ]
+            }
+          }
+        }
+      }
+    end
+
+    let(:lift_wing_scores) do
+      {
+        '1189344512' => { 'wp10' => 0.5 }
+      }
+    end
+
+    before do
+      allow_any_instance_of(WikiApi).to receive(:query).and_return(wiki_api_response)
+      allow_any_instance_of(LiftWingApi).to receive(:get_revision_data).and_return(lift_wing_scores)
+    end
+
+    it 'returns formatted data only for enrolled users' do
+      enrolled_usernames = ['RageSock', 'MillerWon']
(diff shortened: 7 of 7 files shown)


app/controllers/articles_controller.rb
```diff
@@ -1,12 +1,31 @@
...
 
 class ArticlesController < ApplicationController
   respond_to :json
-  before_action :set_course, except: :article_data
+  # the revision_score action derives its course from the article itself,
+  # so it can't rely on a :course_id parameter.  skip the before_action
+  # here to avoid a RecordNotFound/404 when course_id is missing.
+  before_action :set_course, except: [:article_data, :revision_score]
 
   # returns revision score data for vega graphs
```
gabina, line 8: I think the comment should specify the format of the data it returns.

app/controllers/articles_controller.rb
```diff
@@ -1,12 +1,31 @@
...
-  def article_data
+  def revision_score # rubocop:disable Metrics/AbcSize
     @article = Article.find(params[:article_id])
+    @course = @article.courses.current.first
```
gabina: hmm why did you decide to retrieve the course from the article instead of sending the `course_id`? At first glance I would say that the same article may be updated for different courses?

app/controllers/articles_controller.rb
```diff
@@ -1,12 +1,31 @@
...
+    
+    revisions = get_revision_ids(@article.mw_page_id, @course.start, @course.end + 1.week)
+    rev_ids = revisions.map { |r| r[:revid] }
+    scores  = score_revisions(rev_ids, @article.wiki)
```
gabina: I think retrieving revisions and scores for the complete course period can take a long time if there are many revisions. I understand this is scoped to a specific article id, so I would check `articles_courses.revision_count` before fetching data. If that number is over certain threshold, then I would avoid fetching data.
Edit: this is assuming we want to restrict revisions only to users participating in the course.

app/controllers/articles_controller.rb
```diff
@@ -15,17 +34,47 @@ def details
...
   end
 
-  private
+  def get_revision_ids(page_id, start_date, end_date) # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
+    params = { action: 'query', prop: 'revisions',
+               pageids: page_id,
+               rvprop: 'user|ids|timestamp|size',
+               rvlimit: 500 }
+    params[:rvstart] = end_date.strftime('%Y%m%d%H%M%S')
+    params[:rvend] = start_date.strftime('%Y%m%d%H%M%S')
```
gabina: What is the reason why you decided to create the hash first and then add two more fields?

app/controllers/articles_controller.rb
```diff
@@ -15,17 +34,47 @@ def details
...
+              "user": r['user'], 
+              "timestamp": r['timestamp'], 
+              "size": r['size']}
+      end
+      
+      break unless response['continue']
+      params.merge!(response['continue'].slice('rvcontinue', 'continue'))
+    end
+    revisions
+  end
```
gabina: hmm this method is hard to read and have no documentation. In addition, while we're not great on scoping the API requests to some specific classes, I think that fetching revisions and scores in the `ArticlesController` is not the best design. I don't have a better class in mind, but I would put some effort on move this method to other place?
In addition, note that we make a similar request in `lib/alerts/discretionary_sanctions_monitor.rb`. Maybe you can follow the approach

app/controllers/articles_controller.rb
```diff
@@ -15,17 +34,47 @@ def details
...
+    params = { action: 'query', prop: 'revisions',
+               pageids: page_id,
+               rvprop: 'user|ids|timestamp|size',
+               rvlimit: 500 }
+    params[:rvstart] = end_date.strftime('%Y%m%d%H%M%S')
+    params[:rvend] = start_date.strftime('%Y%m%d%H%M%S')
 
+    revisions = []
+    loop do
+      @wiki = Wiki.find_by(language: @article.wiki.language, project: @article.wiki.project)
```
gabina: this could go outside the loop?

app/controllers/articles_controller.rb
gabina: Don't you need to restrict the revisions to the users enrolled in the course?

Formasitchijoh:
@gabina Thank you for the feedback, I have made the changes and improvement you mentioned

app/controllers/articles_controller.rb
Formasitchijoh: That was an oversight, I have ensured the course is gotten from the course_id passed to the endpoint

app/controllers/articles_controller.rb
Formasitchijoh: I think since we are showing the progression of the article over the duration of the course, pulling revisions for the enrolled users is neccesary i have made this update

app/controllers/articles_controller.rb
Formasitchijoh: I initially was checking for if the end_date was provided and hence needed to add these field after but I remember i don't need the check since a course will always have a start and end date

lib/revision_score_manager.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  # Returns true if the revision count is within a performant range.
+  def within_revision_limit?
+    article_course = ArticlesCourses.find_by(article_id: @article.id, course_id: @course.id)
+
+    # If no record exists, we allow the request (default to true)
+    return true if article_course.nil?
+    
+    count = article_course.try(:revision_count) || 0
+    count < 500
```
gabina: This 500 should be a constant. Beyond that, have you tested how long it takes to fetch 500 revisions and scores for all of them? I think you should look for examples, run benchmarks, and find a reasonable number. Maybe it's 500, but I highly doubt that would be a reasonable amount of time for a user to wait for a graph.

gabina: For example, the `recording2.mov` video shows that it takes at least 4 seconds to load the graph, how many revisions and scores have been fetched? I only see 7 points but I'm not sure if that reflects the number of revisions.

Formasitchijoh: The recording2.mov has just 7 revisions as the points indicates. I will update the PR description with the benchmarks for the different range of revisions and their scores

Formasitchijoh:
@gabina I  have made the changes and updated a video of the current implementation, currently the video was taken on a relatively slow internet speed(3Mbs)  for worst case scenario and how the caching works after the first display of the graph

app/controllers/articles_controller.rb
```diff
@@ -1,21 +1,27 @@
...
-  def article_data
-    @article = Article.find(params[:article_id])
-  end
+  #  { rev_id: 123, characters: 1000, wp10: 0.5, date: '2021-01-01', username: 'Example' }
+def revision_score
+  @article = Article.find(params[:article_id])
+  rev_manager = RevisionScoreManager.new(@article, @course)
+  enrolled_usernames = @course.users.pluck(:username)
+  render json: rev_manager.fetch_scored_revisions(enrolled_usernames)
+end
```
gabina: This seems to be incorrectly aligned?

app/controllers/articles_controller.rb
```diff
@@ -1,21 +1,27 @@
...
+
 
   # returns details about how an article changed during a course
   def details
     @article = Article.find(params[:article_id])
     article_course = ArticlesCourses.find_by(course: @course, article: @article, tracked: true)
     @editors = User.where(id: article_course&.user_ids)
   end
-
+
```
gabina: Remove this

lib/revision_score_manager.rb
```diff
@@ -0,0 +1,104 @@
...
+  def relevant_revisions(revisions)
+    within_revision_limit? ? revisions : revisions.first(MAX_REVISION_COUNT)
+  end
+  
+  # Returns true if the revision count is within a performant range.
+  def within_revision_limit?
+    count = ArticleCourseTimeslice.where(article_id: @article.id, course_id: @course.id)
+                                  .sum(:revision_count)
+    count <= MAX_REVISION_COUNT
+  end
```
gabina: I don't think doing this makes sense? At this point, you already fetched all the revisions through the API, so estimating the revision count for the article through the article course timeslices doesn't make sense here since you already have the real revisions.

Estimating the revision count for the article through the article course timeslices may be useful to decide whether to fetch revisions through the API. But at this point I think it's useless.

I think that you may use the estimation **before** fetching the revisions. If the article has 7 revisions during the course, then you can fetch them to make the graph. If the article has 700 revisions during the course, maybe you can just skip it. Does it make sense for you?

lib/revision_score_manager.rb
```diff
@@ -0,0 +1,104 @@
...
+      target_revisions.map do |rev|
+        {
+          rev_id: rev[:revid],
+          characters: rev[:size],
+          wp10: scores.dig(rev[:revid].to_s, 'wp10'),
+          date: rev[:timestamp],
+          username: rev[:user]
+        }
+      end
+    end
```
gabina: Maybe you can extract this into a different method to avoid disabling `Metrics/AbcSize,Metrics/MethodLength`

lib/revision_score_manager.rb
Formasitchijoh: > I don't think doing this makes sense? At this point, you already fetched all the revisions through the API, so estimating the revision count for the article through the article course timeslices doesn't make sense here since you already have the real revisions.
> 
This makes sense that was an oversight  I could just check the length from the fetched revisions from the API

> Estimating the revision count for the article through the article course timeslices may be useful to decide whether to fetch revisions through the API. But at this point I think it's useless.
> 
> I think that you may use the estimation **before** fetching the revisions. If the article has 7 revisions during the course, then you can fetch them to make the graph. If the article has 700 revisions during the course, maybe you can just skip it. Does it make sense for you?

Yes it perfectly does. But Ideally we do not want to show an empty graph ,And so my though process was Since fetching revisions does not affect the performance ( 500 rev takes ~1s) ,  if the revision count from the API or article course timeslices  is >= MAX_REVISION_COUNT  ie 700,  slice the most recent revisions which is (  MAX_REVISION_COUNT ) and plot. Does this sound reasonable ?

gabina: If you're going to fetch revisions no matter the article course timeslices count, then there's no point in doing the or:  if the revision count from the API **or** article course timeslices is >= MAX_REVISION_COUNT.

If you consider that fetching revisions does not affect the performance (which may be a bit tricky, because 500 rev takes ~1s but you don't know how many requests would need to do, but I guess we're OK most of the cases), then don't use the article course timeslices estimation at all, because you're doing a query to the db when you already fetched the real data.

If you're going to build a plot with only the most recent revisions, it should be explicit to avoid confusion.

Formasitchijoh: > If you're going to fetch revisions no matter the article course timeslices count, then there's no point in doing the or: if the revision count from the API **or** article course timeslices is >= MAX_REVISION_COUNT.
> 
I just updated the fix and removed this check

> If you consider that fetching revisions does not affect the performance (which may be a bit tricky, because 500 rev takes ~1s but you don't know how many requests would need to do, but I guess we're OK most of the cases), then don't use the article course timeslices estimation at all, because you're doing a query to the db when you already fetched the real data.
> 
I've removed the timeslice estimation logic as suggested to avoid the extra database query. Since the API returns up to 500 revisions per request, the performance impact of fetching real data is negligible compared to the scoring process. Slicing MAX_REVISION_COUNT from the filtered revisions ensures the graphs are based on the most accurate recent data

> If you're going to build a plot with only the most recent revisions, it should be explicit to avoid confusion.

To address the confusion regarding the revision limit, I have added a comment to the **relevant_revisions** method. to clarify this

Formasitchijoh:
@gabina I think this is ready for review.

lib/revision_score_manager.rb
```diff
@@ -0,0 +1,104 @@
...
+  end
+
+  # Filters the total revision to a manageable number for plotting.
+  # We explicitly take the most recent revisions (up to MAX_REVISION_COUNT)
+  # to ensure the graph remains performant and readable.
+  def relevant_revisions(revisions)
+    revisions.size > MAX_REVISION_COUNT ? revisions.first(MAX_REVISION_COUNT) : revisions
+  end
```
gabina: Can't you just return `revisions.first(MAX_REVISION_COUNT)` for all cases? If `revisions.size <= MAX_REVISION_COUNT`, then you would be returning revisions anyway. Let me know if I'm missing something here

lib/revision_score_manager.rb
gabina: > To address the confusion regarding the revision limit, I have added a comment to the relevant_revisions method. to clarify this

That makes sense, but I think that maybe we could add something for the user perspective. If the user is supposed to have to wait a certain amount of time, we could make that explicit instead of only showing "Loading..". Maybe adding something like "This can take a time...".  In addition, I think the plot title could mention that it shows the last N revisions... Maybe it's OK for this PR and you can open an issue for those docs for the user and adding them in other opportunity. WDYT?

lib/revision_score_manager.rb
Formasitchijoh: No you are not, I missed how the `.first` method works I have updated the method call.

Formasitchijoh:
@gabina Is there still something you would like me to fix here.

gabina:
Really sorry for my delay here, I wasn´t working for some weeks. There were several changes while I was out, and I would like to understand if some of them are related to this, as we made changes in LiftWingApi

Formasitchijoh:
> Really sorry for my delay here, I wasn´t working for some weeks. There were several changes while I was out, and I would like to understand if some of them are related to this, as we made changes in LiftWingApi

No Problem welcome back :) 

These changes here are still wrt what we discussed and some minor improvements I saw necessary to improve. It has nothing to do with the latest changes with LiftWingApi. 

I did see some work Sage was doing with the LiftwingApi and I do not know if it will affect these changes. I might need to check in with him.

ragesoss:
My LiftWing changes (I think) only affected the course update process, so on-demand requests through LiftWing (which I think are the idea with this PR) shouldn't be affected.

Formasitchijoh:
> My LiftWing changes (I think) only affected the course update process, so on-demand requests through LiftWing (which I think are the idea with this PR) shouldn't be affected.

Great, thanks for the clarification

gabina:
This LGTM. I'll do a review with Claude Code to see what happens.

app/assets/javascripts/components/articles/article.jsx
```diff
@@ -94,8 +95,7 @@ const Article = ({ article, index, course, fetchArticleDetails, updateArticleTra
           {!isDeleted
             ? (
               <small>
-                {/* Disabling article development plot feature until re-implementing. See issue #6337 */}
-                {/* <a href={historyUrl} target="_blank" className="inline">{I18n.t('articles.history')}</a> | <ArticleGraphs article={article} /> */}
+                <a href={historyUrl} target="_blank" className="inline">{I18n.t('articles.history')}</a> | <ArticleGraphs article={article} course_id={course.id}/>
                 <a href={historyUrl} target="_blank" className="inline">{I18n.t('articles.history')}</a>
```
gabina: From the Claude review: is it OK to have two "History" links?

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -35,15 +35,15 @@ const ArticleGraphs = ({ article }) => {
...
     if (articleData) {
       return;
     }
-    const articledataUrl = `/articles/article_data.json?article_id=${article_id}`;
+    const articledataUrl = `/articles/${article_id}/revision_score?course_id=${course_id}`;
     request(articledataUrl)
       .then(resp => resp.json())
       .then((data) => {
-        setArticleData(data);
+        setArticleData(data?.articles);
```
gabina: From Claude review:

The controller renders a plain JSON array:

`render json: rev_manager.fetch_scored_revisions(enrolled_usernames)`
  → [{rev_id: 1, characters: 100, wp10: 0.5, ...}, ...]

  But the frontend reads `.articles` from that response:

` setArticleData(data?.articles);`

Is that OK?

app/assets/javascripts/components/articles/article_graphs.jsx
```diff
@@ -35,15 +35,15 @@ const ArticleGraphs = ({ article }) => {
     };
   }, []);
 
-  function getData() {
+  async function getData() {
```
gabina: From Claude review:

Minor — `getData` declared `async` without using `await`
The `async` keyword is unused since `.then()` chains are used throughout. Either remove `async` or convert the body to use `await`.

gabina:
Claude Code also points out that the spec only mocks a single-page response with no continue key. A spec covering the multi-page continuation path would be useful.

I imagine that we several other places in the app where we don't have a spec for that continue loop. Undesirable but acceptable I guess.

app/assets/javascripts/components/articles/article_graphs.jsx
Formasitchijoh: Yes you are right , this was a bit of a hurdle but found why the array was wrapped in the articles key. The `wrap_parameters initializer (config/initializers/wrap_parameters.rb)` was causing the response array to be wrapped under the controller name articles. Fixed by adding `root: false` to the render call, which opts out of this wrapping. You can read more about it [here](https://api.rubyonrails.org/classes/ActionController/ParamsWrapper.html).

Formasitchijoh:
Hi @gabina , @ragesoss  I made the required fix and cleaned up the commit history with details about what changes were actually made. Can you take a look

app/assets/javascripts/components/articles/article_graphs.jsx
gabina: > The wrap_parameters initializer (config/initializers/wrap_parameters.rb) was causing the response array to be wrapped under the controller name articles.

Interesting, didn't know that!

So, as I understand it, by default all JSON responses rendered by our controllers are wrapped, right? If that's the case, I don't see a specific reason to change the default behavior for this particular case. WDYT?

I think Claude's comment comes from not knowing why the response was wrapped, but if wrapping is our default, I don't see that as a problem.

ragesoss (changes requested):
**A note on this feedback:** the following came out of a Claude Code review session. It isn't meant to be trusted or acted on automatically — it's offered to read, sanity-check, and use or discard on the author's judgment. The first finding was reproduced locally; the rest are observations that warrant independent verification.

**Re-enabled graph will return a 500 on wikis with no LiftWing articlequality model**

`RevisionScoreManager#score_revisions` calls `LiftWingApi.new(@wiki)`, whose constructor raises `InvalidProjectError` for any wiki outside `AVAILABLE_WIKIPEDIAS` (`en eu fa fr gl nl pt ru sv tr uk`). There's no guard, so the exception propagates uncaught through `ArticlesController#revision_score` and the request returns a 500.

The pre-timeslice implementation read pre-computed `revision.wp10` from the database, which was simply `nil` on unsupported wikis, so the graph degraded to edit-size only. The on-demand approach here removes that graceful fallback. On the Programs & Events Dashboard — which runs many language editions outside the supported set (Spanish, German, Italian, Arabic, etc.) — opening "Article development" on an article a student edited would hit this path.

Reproduced locally with a contrasting spec: an article on en.wikipedia returns scored revisions; the same setup on es.wikipedia raises `LiftWingApi::InvalidProjectError` at `lib/lift_wing_api.rb:37`, reached via `lib/revision_score_manager.rb:86`. A guard such as the following would restore the old graceful degradation:

```ruby
def score_revisions(rev_ids)
  return {} unless LiftWingApi.valid_wiki?(@wiki)
  LiftWingApi.new(@wiki).get_revision_data(rev_ids)
end
```

(with `wp10` then falling through as `nil`, which the frontend already handles via its `dataIncludesWp10` check). The AI-generated contrasting spec used to reproduce this can be shared if useful.

**Supported-wiki list is current — no change needed**

Separately checked whether `AVAILABLE_WIKIPEDIAS` is stale. It matches the current articlequality model set in [`wikimedia/articlequality/models`](https://github.com/wikimedia/articlequality/tree/master/models) and the [Wikitech LiftWing](https://wikitech.wikimedia.org/wiki/Machine_Learning/LiftWing) listing exactly. articlequality is a frozen, ORES-migrated model set that isn't being expanded, so the unsupported-language gap above is expected to persist rather than resolve on its own.

**Related robustness note**

`get_student_revisions` indexes `response['query']['pages'][…]` directly, but `WikiApi#query` returns `nil` after exhausting retries, and a missing/deleted page id won't appear under `pages`. Both cases would raise `NoMethodError` → 500. Lower priority than the above, but the same pattern of a live-API edge case surfacing to the user as a 500 where the DB-backed path did not.

**Minor**

The `root: false` comment in the controller attributes the bare-array response to opting out of `wrap_parameters`. `wrap_parameters` only affects incoming request params, not `render json:` output; with no ActiveModelSerializers and `include_root_in_json` off, `root:` is ignored on a plain array. The response is a bare array regardless — behavior is correct, but the comment's explanation may mislead later readers.

*Drafted in a Claude Code session (~25 min, ~3 interactions). Sage directed the investigation, reviewed all findings before posting, and had Claude Code verify the main finding locally — running the contrasting spec to reproduce the `InvalidProjectError` on es.wikipedia and cross-checking the articlequality list against the model repo and Wikitech. The minor and robustness notes were read but not each independently reproduced.*

ragesoss:
The data fetching changes are working well!

Testing this out locally, I'm struck by how bad these graphs are visually, and I don't want to re-enable them without significant design changes. I'm going to work on this.

Thanks for bring this PR so far @Formasitchijoh !

Issue #6705, opened 2026-03-03
GetRevisionPlaintext should exclude image captions


ragesoss:
We use `GetRevisionPlaintext` to isolate the prose content of Wikipedia edits and pages, before checking that text with an AI detector. (This is because Pangram, at least, seems to work best with strictly prose content.)

Currently, if an image is added as part of an edit, `GetRevisionPlaintext` output includes the image caption.

Example: https://en.wikipedia.org/w/index.php?title=Third_place&diff=prev&oldid=1340536495

Currently, the rest of the image markup is removed, but the caption remains: "Coworking Space in Berlin"

Instead, the entire image (including caption) should be removed.

sharon-kathambi:
Working on this.

ragesoss:
Great. The fix itself will probably be small and simple, but try to make the testing as elegant as you can. This is a service that we can write tests for pretty easily, and it's likely to become more complex over time, so good tests are critical.

sunkireddy-Barath:
Hi @ragesoss  I would like to work on this issue and open a PR shortly 
Thank You!

PR #6721, opened 2026-03-11
Fixed the getrevisionplaintext

sunkireddy-Barath:
What this PR does:

This PR fixes issue #6705: GetRevisionPlaintext should exclude image captions.

Previously, when a Wikipedia revision included an image, the service removed most of the image markup but the caption text still appeared in the extracted plaintext. Since captions are not part of the article prose, they should not be included in the text used for AI detection.

This change updates the service to remove common Wikipedia image containers (.thumb, figure, .mw-default-size) and image elements (img) before converting HTML to plaintext. This ensures that image captions are excluded from the extracted prose.

Tests were also added to verify that:
image captions are removed
figure-based images are removed
inline images are removed
surrounding prose text remains intact

AI usage
AI tools were used only to check for minor errors and verify the implementation. 
The solution, code changes, and tests were written and reviewed manually.

Open questions and concerns:
No open questions at the moment. I welcome any feedback or suggestions for improving the implementation or tests.

### DIFF
--- a/app/services/get_revision_plaintext.rb
+++ b/app/services/get_revision_plaintext.rb
@@ -77,13 +77,125 @@ class GetRevisionPlaintext
     @diff_table = resp.data['*']
   end
 
+  # Parses the diff table HTML and extracts only the
+  # changed wikitext, excluding pre-existing content
+  # when new text is an independent sentence addition.
+  # Parses the diff table HTML and extracts only the
+  # changed wikitext, excluding pre-existing content
+  # when new text is an independent sentence addition.
   def generate_wikitext_from_diff_table
     doc = Nokogiri::HTML(@diff_table)
-    # Collect all the '.diff-addedline' table cells.
-    # These represent all the sections of wikitext that were either added or changed,
-    # IE, the right side of a traditional wikitext diff table without the unchanged
-    # .diff-context cells.
-    @changed_wikitext = doc.css('.diff-addedline').map(&:text).reject(&:empty?).join("\n\n")
+    @changed_wikitext = extract_changed_parts(doc).reject(&:empty?).join("\n\n")
+  end
+
+  # Iterates over each <tr> in the diff table.
+  # See https://en.wikipedia.org/w/index.php?title=Third_place&diff=prev&oldid=1340536495
+  # The doc contains a table like:
+  #
+  #   <table>
+  #     <tr>
+  #       <td class="diff-deletedline diff-side-deleted">
+  #         <div>In sociology...refers to...</div>
+  #       </td>
+  #       <td class="diff-addedline diff-side-added">
+  #         <div>In sociology...refers to...
+  #         <ins class="diffchange">Scholars have
+  #         noted...</ins></div>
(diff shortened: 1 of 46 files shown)


sunkireddy-Barath:
Thank you for the suggestions #6705
After doing the changes  I will open an PR shortly

Abishekcs:
Follow the PR template.

Abishekcs:
It's troublesome to read PR description without no formatting.

sunkireddy-Barath:
@Abishekcs Sure , I will update the PR Template format
Thank You!

app/services/get_revision_plaintext.rb
```diff
@@ -104,12 +104,15 @@ def generate_plaintext_from_html
...
 
-  def remove_html_tables_and_citations(html)
+  def remove_non_prose_html_elements(html)
     doc = Nokogiri::HTML(html)
     doc.xpath('//table').each(&:remove) # Exclude tables, like infoboxes
     doc.xpath('//cite').each(&:remove) # Exclude `cite` tags, which usually appear at the end
     doc.css('.mw-cite-backlink').each(&:remove) # Exclude backlinks that precede `cite` tags.
 
+    # Exclude images and captions
+    doc.css('.thumb, figure, .mw-default-size, img').each(&:remove)
```
ragesoss: Is this the simplest and most general way to find the image elements? It has a number of selectors that might be too specific and miss different forms of image markup in articles. For example, `.thumb` might not be present and `.mw-default-size` might not be present.

spec/services/get_revision_plaintext_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  let(:wiki) { double('Wiki', id: 1) }
+  let(:mw_rev_id) { 12345 }
+  let(:wiki_api) { double('WikiApi') }
+
+  before do
+    allow(WikiApi).to receive(:new).with(wiki).and_return(wiki_api)
+    allow(Sentry).to receive(:capture_message)
+  end
+
+  describe '#plain_text' do
```
ragesoss, line 22: I'd prefer a testing strategy that sticks to the public interface for this service, without isolating private methods individually.

See my comments on the parallel PR: https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6709

sunkireddy-Barath:
Thanks for the feedback.

I've updated the implementation to use a simpler and more general approach for removing image elements, and I've also rewritten the tests to use the public interface of the service instead of calling private methods directly.

I've addressed the review comments and pushed the changes. Please take a look when you have time.

Thank You

ragesoss:
This looks good to me! I'm going to hold off on merging it, as I might want to handle a merge myself after https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6709 is ready. That one is a more complex change, so it will be easier to add this to the final version of the other PR than vice-versa.

sunkireddy-Barath:
Thank you for the feedback. Understood, I’ll wait until #6709 is finalized.

Issue #6724, opened 2026-03-11
AI detection comparison tool should work for an article URL


ragesoss:
The tool currently processes a diff URL, but it should also be able to handle an article URL without a specific revision or diff range specified. (I think the GetRevisionPlaintext is already set up to handle this.)

gabina:
According to what I understand, we want to support these three cases:

**1.** Diff mode with two revisions: https://en.wikipedia.org/w/index.php?title=Greater_Cooch_Behar_People%27s_Association&diff=1299571191&oldid=1299350801
**2.** Specific revision: https://en.wikipedia.org/w/index.php?title=Greater_Cooch_Behar_People%27s_Association&oldid=1299350679
**3.** Plain article: https://en.wikipedia.org/wiki/Greater_Cooch_Behar_People%27s_Association

In the AI detection comparison tool we currently support **1.**  and **2.**.

I'm currently working on supporting **3**. For that case it's possible to reuse `GetRevisionPlaintext`, but we have to provide the mw_rev_id  from the article title.

Let me know if there are any other cases we need to handle.

ragesoss:
For now, those are the cases I'm aware of.

I think another useful feature would be to add a text field and let you check arbitrary text rather than only wiki content. Not a high priority, but also probably easy to implement.

PR #6730, opened 2026-03-13
Support article urls in AI detection comparison tool

gabina:
## What this PR does
This PR is part of #6724.

It makes the AI detection comparison tool handle article URLs like:
https://en.wikipedia.org/wiki/Greater_Cooch_Behar_People%27s_Association

Note this kind of URLs doesn't have any revision id, so we need to manually fetch the latest revision to reuse the existing `GetRevisionPlaintext`.

### List of supported URL formats:
**Non-diff**

- https://en.wikipedia.org/w/index.php?title=List_of_the_busiest_airports_in_Malaysia&oldid=1276659876 
- https://en.wikipedia.org/w/index.php?oldid=1315039613
- https://en.wikipedia.org/wiki/Greater_Cooch_Behar_People%27s_Association (support added in this PR)

**Diff**

- https://en.wikipedia.org/w/index.php?title=Richard_G._F._Uniacke&diff=prev&oldid=936368512
- https://en.wikipedia.org/w/index.php?title=Richard_G._F._Uniacke&diff=1178859026&oldid=711811679
- https://en.wikipedia.org/w/index.php?title=List_of_hystricids&diff=1315039613
- https://en.wikipedia.org/w/index.php?diff=1315039613

## AI usage
No AI usage.

## Screenshots
Before:
[Screencast from 2026-03-13 00-32-48.webm](https://github.com/user-attachments/assets/b395d962-5f23-410f-89fb-5633eacdf2eb)

After:
[Screencast from 2026-03-13 00-27-30.webm](https://github.com/user-attachments/assets/a34e11b8-30ec-4be6-9bce-1bdac1abb377)

### DIFF
--- a/app/views/ai_tools/show.html.haml
+++ b/app/views/ai_tools/show.html.haml
@@ -9,7 +9,6 @@
       = form_with url: "/ai_tools/compare_pangrams", method: :post do |form|
         = label_tag :article_or_diff_url, 'Diff URL:'
         = text_field_tag :article_or_diff_url 
-        = hidden_field_tag :diff_mode, 'true'
         %hr
         %h4 Models
         - AiToolsController::MODELS_KEY.each do |key|


app/controllers/ai_tools_controller.rb
```diff
@@ -11,7 +11,6 @@ def show; end
 
   def compare_pangrams
     parse_url
-    @diff_mode = params[:diff_mode] == 'true'
```
gabina, line 14: Removed this since `@diff_mode` is based on the URL format.

app/controllers/ai_tools_controller.rb
```diff
@@ -63,9 +62,34 @@ def parse_url
...
+      revs = [parser.oldid, parser.diff].compact
+      @rev_id = revs.max
+      @from_rev = revs.min if revs.count == 2
+      @diff_mode = true
+    else
+      # If there is no oldid revision in the url, it means it's just an article url and we
+      # have to manually fetch the latest revision for it.
+      # Example: https://en.wikipedia.org/wiki/Greater_Cooch_Behar_People%27s_Association
+      @rev_id = latest_revision
+      @diff_mode = false
```
gabina, line 74: I considered using `@diff_mode = parser.oldid.present?` outside the `if` statement but I thought it was less clear.

app/controllers/ai_tools_controller.rb
```diff
@@ -63,9 +62,34 @@ def parse_url
...
+  end
+
+  # Query for the latest revision for a given article title
+  def query_params
+    {
+      action: 'query',
+      prop: 'revisions',
+      titles: CGI.unescape(@article_title),
+      rvprop: 'ids'
+    }
```
gabina, line 93: It makes me a little nervous to spread this kind of thing across different classes. Although all requests are made via `WikiApi` and are therefore easy to find in the code, I think that at some point we could do some refactoring to try to encapsulate them a little more. I left this request here because I don't know which class would be best to place it in.

ragesoss: Yeah, we have this kind of thing spread widely across the codebase, and it would make sense to encapsulate many of the interactions with WikiApi into dedicated classes.

My first thought would be to start with the queries that are related to article content: getting the wikicode for the current version of an article, getting the wikicode of a particular revision, getting the html for a an article or revision, and so on. There are many variations of that, but bringing them together would make it easier to refactor for clarity and hide the details of how to construct an arbitrary MediaWiki API query.

ragesoss:
There are a bunch of different URL formats for articles, old revisions of articles, and diffs.

These are all the same diff:
* https://en.wikipedia.org/w/index.php?title=List_of_hystricids&diff=prev&oldid=1315039613
* https://en.wikipedia.org/w/index.php?title=List_of_hystricids&diff=1315039613&oldid=1306839019
* https://en.wikipedia.org/w/index.php?title=List_of_hystricids&diff=1315039613
* https://en.wikipedia.org/w/index.php?diff=1315039613

These are the same non-diff:
* https://en.wikipedia.org/w/index.php?title=List_of_hystricids&oldid=1315039613
* https://en.wikipedia.org/w/index.php?oldid=1315039613

ragesoss:
If it can handle all those in terms of diff mode vs non-diff mode, the rest of it looks good to me.

gabina:
Yes, I've added support for all those versions. Right now, I'm adding some specs to make sure everything works as it should, since there are several formats and it can get confusing.

gabina:
I added some specs and tested all the supported different URL formats locally:

**Non-diff**

- https://en.wikipedia.org/w/index.php?title=List_of_the_busiest_airports_in_Malaysia&oldid=1276659876 
- https://en.wikipedia.org/w/index.php?oldid=1315039613
- https://en.wikipedia.org/wiki/Greater_Cooch_Behar_People%27s_Association (support added in this PR)

**Diff**

- https://en.wikipedia.org/w/index.php?title=Richard_G._F._Uniacke&diff=prev&oldid=936368512
- https://en.wikipedia.org/w/index.php?title=Richard_G._F._Uniacke&diff=1178859026&oldid=711811679
- https://en.wikipedia.org/w/index.php?title=List_of_hystricids&diff=1315039613
- https://en.wikipedia.org/w/index.php?diff=1315039613

Issue #6501, opened 2025-10-21
'Article-Scoped' should be an option for any course type


ragesoss:
`ArticleScopedProgram` is a course type, but article-scoping is a feature that is independent of other 'types' like Edit-a-thon or Individual Contributor.

The  ASP course type should be refactored to be a behavior that can be enabled or disabled for a course of any other type. Some course types (like ClassroomProgramCourse) should not display the option on the frontend.

bhushan354:
I am working on this

ragesoss:
> I am working on this

You're welcome to try it, but I'm not sure how complex of a change it will be... potentially, quite a lot of work will be be involved refactoring the relevant code.

lakshita10341:
@ragesoss, I would like to work on this issue. After looking into the codebase, I noticed that `ArticleScopedProgram` is nearly identical to `BasicCourse`, the only real differences are that it includes `CustomRevisionFilter` and overrides `only_scoped_articles_course?` to return `true`. So we can convert it  into a flag-based behavior rather than a separate course type. Here's my proposed approach:

1. Use the existing `flags` hash column to store a new `flags[:article_scoped]` boolean, so no schema migration is needed.

2. Update `only_scoped_articles_course?` in the base Course model to check this flag instead. Also move `scoped_article?` and `scoped_article_timeslices` from CustomRevisionFilter into the base Course model, guarded by the same flag.

3. Write a data migration to convert existing ArticleScopedProgram records to BasicCourse with `flags[:article_scoped] = true`, then remove the ArticleScopedProgram class.

4. The backend services that rely on `only_scoped_articles_course?` (update_course_stats, course_revision_updater, update_timeslices_scoped_article) won't need changes since they already depend on that method rather than the course type directly.

5. On the frontend, replace all `course.type === 'ArticleScopedProgram'` checks with a new `course.article_scoped` attribute exposed via the serializer. Remove ArticleScopedProgram from the course type selector and add an "Enable Article Scoping" toggle for eligible types (hidden for ClassroomProgramCourse).

If this approach looks good to you, I can start working on it!

ragesoss:
This sounds good overall. I think the migration is perhaps not strictly necessary, and might be better off as a followup PR once everything else is working. Then ArticleScopedProgram can retain a method that returns `true` instead of checking for a flag. That approach would also mean we could safely roll out the backend changes without any frontend support initially.

lakshita10341:
[...]

PR #6732, opened 2026-03-13
Add article_scoped flag and logic to base Course model

lakshita10341:
## What this PR does
This PR refactors "Article Scoping" from a distinct course type (`ArticleScopedProgram`) into a toggleable behavior that can be enabled for any course type via a flag. This is a backend-only change to prepare the system for the upcoming frontend toggle.
Closes 1st part of #6501 

**Key Backend Changes:**
- **Base Course Model:** Updated `only_scoped_articles_course?` to check for `flags[:article_scoped]`. It also implements the filtering logic for `scoped_article?` and `scoped_article_timeslices` directly in the base class, activated by this flag.
- **API Support:** Added `article_scoped` to `course.json.jbuilder` so the frontend can detect this setting regardless of the course type.
- **Testing:** Added `spec/models/article_scoped_flag_spec.rb` to verify the flag works for all course types.

## AI usage
I used AI for:
- Research the project architecture and identify relevant code paths for article scoping.
- Generate and run the RSpec test suite within the Docker environment to verify the new behavior and backward compatibility.

## Screenshots
N/A(This is the backend only change.  Existing `ArticleScopedProgram` and `VisitingScholarship` courses continue to work exactly as before by using their own method overrides.)

## Next Steps
1. **Frontend PR:** Add the "Enable Article Scoping" toggle to the course creator and update components (like `ArticlesHandler`) to use the new flag instead of checking the course type.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -345,6 +345,10 @@ class Course < ApplicationRecord
   # Checks if an article is in scope (assigned or in a tracked category).
   def scoped_article?(wiki, title, mw_page_id)
     return true unless only_scoped_articles_course?
+    # Normally, scoped_article_titles will include all the in-scope articles
+    # but if the title of an assigned article has changed, we still want to process
+    # edits to that article so that we can update the Assignment#article_title.
+    # A secondary check against the mw_page_ids for assigned articles covers that.
     scoped_article_titles(wiki).include?(title) ||
       assigned_article_page_ids(wiki).include?(mw_page_id)
   end


app/models/course.rb
```diff
@@ -337,20 +337,22 @@ def tracked_namespaces
...
-    false
+    flags[:article_scoped].present?
   end
 
-  # The default implementation retrieves always true.
-  # A course type may override this implementation.
-  def scoped_article?(_wiki, _title, _mw_page_id)
-    true
+  # Checks if an article is in scope (assigned or in a tracked category).
+  def scoped_article?(wiki, title, mw_page_id)
```
ragesoss, line 346: This should be implemented in a way that doesn't duplicate logic between flag-based courses and `ArticleScopedProgram` type courses.

app/models/course.rb
```diff
@@ -337,20 +337,22 @@ def tracked_namespaces
...
+    scoped_article_titles(wiki).include?(title) ||
+      assigned_article_page_ids(wiki).include?(mw_page_id)
   end
 
-  # The default implemention retrieves all the ACTs.
-  # A course type may override this implementation.
+  # Returns article course timeslices for scoped articles only.
   def scoped_article_timeslices
-    article_course_timeslices
+    return article_course_timeslices unless only_scoped_articles_course?
```
ragesoss, line 358: Same as above, this shouldn't duplicate any logic.

lakshita10341:
> This looks like it will be admirably simple!
> 
> I like that we currently have the important scoping behavior for `ArticleScopedProgram` encapsulated in a separate file (the `CustomRevisionFilter` module).
> 
> Is it possible to use that module for flag-based courses as well, such as by including the module for the individual object at the time it's created or afterwards?
> 
> In any case, I think we need to be using shared code for the class-based scoping and the flag-based scoping.
> 
> @gabina feel free to chime in if you see a better way to implement this.

I looked into `CustomRevisionFilter`, we can use the shared `CustomRevisionFilter` module instead of duplicating the logic. Here's how I'm planning to fix this:

**1. Add guards to `CustomRevisionFilter`** so it can be safely included in the base `Course` model:

```ruby
# app/models/course_types/custom_revision_filter.rb
module CustomRevisionFilter
  def scoped_article?(wiki, title, mw_page_id)
    return true unless only_scoped_articles_course?
    scoped_article_titles(wiki).include?(title) ||
      assigned_article_page_ids(wiki).include?(mw_page_id)
  end

  def scoped_article_timeslices
    return article_course_timeslices unless only_scoped_articles_course?
    article_course_timeslices.where(article_id: scoped_article_ids)
  end
end
```

**2. Include `CustomRevisionFilter` in the base `Course` model** and remove the duplicated methods from `course.rb`

This way `ArticleScopedProgram`, `VisitingScholarship`, and any flag-based course will all share the same filtering logic from one place. Does this approach look good to you?

gabina:
This is an interesting problem, I would say that we have to move from inheritance to composition.

> Is it possible to use that module for flag-based courses as well, such as by including the module for the individual object at the time it's created or afterwards?

That would be nice but I think that `include` works at the class level and the condition we need to evaluate works as the instance level (`flags[:article_scoped].present?`). At first glance I would say that's not possible to make some conditional inclusion like `include CustomRevisionFilter if only_scoped_articles_course?`, but it's worth looking into further.

I agree that ideally we should re-use the code defined in `CustomRevisionFilter` module, but maybe we have to unbound the instance methods defined in `CustomRevisionFilter` and call them as if they were from `Course` object? I can't think of a cleaner way to re-use them right now.

gabina:
Sorry @lakshita10341, I answered without reading your comment. I guess that approach makes sense, although it changes a bit the idea behind `CustomRevisionFilter` module. I think we should write a clear comment about why it works that way

ragesoss:
Thanks @gabina and @lakshita10341!

If the shared-code version means adding the guard clauses to those `CustomRevisionFilter` methods and including them in `Course`, then it probably just makes sense to put those methods in `Course` directly, since at that point they'll be written in a way that works for all course types, so not much point in having a module.

app/models/course.rb
lakshita10341: Just to confirm, to remove the duplication, should I delete the `CustomRevisionFilter` module entirely? 
Since it's only used by `VisitingScholarship` and `ArticleScopedProgram`, and those two methods (`scoped_article?` and `scoped_article_timeslices`) will now be defined in the base `Course` model, both of those course types can simply inherit them directly.
Will it work?

ragesoss: Yes, I think that's reasonable.

lakshita10341:
@ragesoss The PR is ready for review. PTAL when you get time!

ragesoss:
This looks good to me. @gabina any remaining concerns?

lakshita10341:
@gabina Could you please review and confirm if everything looks good for merge?

app/models/course.rb
```diff
@@ -337,20 +337,22 @@ def tracked_namespaces
...
 
-  # The default implementation retrieves always true.
-  # A course type may override this implementation.
-  def scoped_article?(_wiki, _title, _mw_page_id)
-    true
+  # Checks if an article is in scope (assigned or in a tracked category).
+  def scoped_article?(wiki, title, mw_page_id)
+    return true unless only_scoped_articles_course?
+    scoped_article_titles(wiki).include?(title) ||
+      assigned_article_page_ids(wiki).include?(mw_page_id)
```
gabina, line 353: I would like to keep the comment about why we use the `or`:

    # Normally, scoped_article_titles will include all the in-scope articles
    # but if the title of an assigned article has changed, we still want to process
    # edits to that article so that we can update the Assignment#article_title.
    # A secondary check against the mw_page_ids for assigned articles covers that.

gabina:
Sorry for the delay. LGTM! My only suggestion is to keep the comment explaining the logic for `scoped_article?` because I think we struggled a bit with that bug.

app/models/course.rb
lakshita10341: Okay, sure! I'll add it.

Issue #6736, opened 2026-03-16
Add default language for campaign


ragesoss:
See https://phabricator.wikimedia.org/T420090

costajohnt:
I'd like to work on this. Based on the Phabricator task (T420090), it looks like two things are needed:
1. Add a `default_language` field to campaigns (following the `default_course_type` pattern) and include it as `&locale=` in invite URLs
2. Preserve the `enroll` query parameter when users navigate/switch language

I'll start with Part 1 and can tackle Part 2 as a follow-up if you'd prefer to keep them separate. Let me know if I'm missing anything.

ragesoss:
@costajohnt thanks for working on this! I started reviewing this, and realized the there's some terminology mismatching in the Phabricator ticket. It says "campaign", but the actual examples are for a `Course`, not a `Campaign`.

I think the actual problem to solve is more like use a course's home wiki to set a default locale (assuming the user doesn't have a locale preference set already), instead of always using English as the default.

archietyagi100-tech:
@ragesoss I'd like to pick this up. From your last comment, it sounds like the fix should use the course's home wiki to set the default locale instead of always falling back to English  only when the user hasn't set a locale preference themselves. I'll dig into how the locale is currently being set and share my plan before writing any code. Let me know if I'm on the right track.

ragesoss:
Sure, give it a shot. Thanks @archietyagi100-tech

ArchieTyagi:
Hi @ragesoss, would appreciate your review on this whenever you get a chance. Happy to make any changes if needed. Thanks.

PR #6739, opened 2026-03-16
feat: add default_language to campaigns

costajohnt:
## What this PR does

Adds a `default_language` field to campaigns so campaign managers can set a default language for their campaign. This language is included as a `&locale=` parameter in enrollment links sent via email, ensuring participants from non-English campaigns land on the correct language version of the dashboard.

This addresses #6736 and covers Part 1 of [T420090](https://phabricator.wikimedia.org/T420090).

Changes made:
- New migration adding `default_language` string column to campaigns table
- Model validation against `I18n.available_locales`
- Language dropdown added to campaign create and edit forms
- Enrollment links in mailers now include `&locale=` when a campaign has a default language set
- `default_language` exposed in the campaign JSON API
- Model validation tests and mailer tests added

## AI usage

I used Claude Code to help scaffold the implementation by exploring the existing `default_course_type` pattern and identifying all the files that needed changes. The actual code changes, testing approach, and architectural decisions were guided by my understanding of the codebase from previous contributions. I also used it to help write the model validation and mailer tests.

## Screenshots

Here's the campaign edit form with the new Default Language field. The dropdown shows all available locales, and the current value is displayed inline when not editing.

Screenshot 2026-03-17 at 3 47 34 PM

Screenshot 2026-03-17 at 3 49 32 PM

Screenshot 2026-03-17 at 3 47 53 PM

This covers Part 1 from T420090. Part 2 (preserving the `enroll` query parameter during frontend navigation and language switching) can follow as a separate PR.

The language dropdown currently shows all 156+ available locales as raw locale codes (e.g., "nl", "es", "fr"). A future improvement could show human-readable language names instead.

### DIFF
--- a/app/models/campaign.rb
+++ b/app/models/campaign.rb
@@ -56,7 +56,8 @@ class Campaign < ApplicationRecord
     ArticleScopedProgram
   ].freeze
   validates :default_course_type, inclusion: { in: ALLOWED_TYPES }, allow_blank: true
-  validates :default_language, inclusion: { in: I18n.available_locales.map(&:to_s) }, allow_blank: true
+  validates :default_language, inclusion: { in: I18n.available_locales.map(&:to_s) },
+                              allow_blank: true
 
   before_save :set_default_times
 


db/migrate/20260316200000_add_default_language_to_campaigns.rb
```diff
@@ -0,0 +1,5 @@
+class AddDefaultLanguageToCampaigns < ActiveRecord::Migration[7.0]
```
ragesoss, line 1: This should also include the corresponding change to schema.rb

app/mailers/no_enrolled_students_alert_mailer.rb
```diff
@@ -22,7 +22,9 @@ def email(alert)
   def set_course_and_users
     @course = @alert.course
     @course_link = "https://#{ENV['dashboard_url']}/courses/#{@course.slug}"
+    locale_param = @course.campaigns.first&.default_language
```
ragesoss, line 25: I don't think any changes to the mailers should be included. Programs & Events Dashboard doesn't send any of these emails (only Wiki Education Dashboard), and Wiki Education Dashboard doesn't create courses from a Campaign.

ragesoss:
See my comment on the issue... I think this isn't actually the change we want to make. I apologize for not noticing that and addressing it when I opened the issue.

PR #6750, opened 2026-03-22
feat: add cumulative_diff_url method to ArticlesCourses

### LINKED ISSUE #6749: Cumulative diff URL should be available from an ArticlesCourses record
For our AI detection comparison tool, we use a diff URL. One typical use case is to look at the cumulative diff representing all the changes to an articles from before a student started editing it up through their last edit.

This is exactly what the frontend DiffViewer on the Articles tab of a course provides, but it figures out the diff ranges by making client-side queries to the wiki API to find the parent revision of the first edit by a student, and the latest edit by a student.

It would be convenient if we also had an easy way to generate those cumulative diffs server-side, so they could be included in alert emails and/or used for investigating large sets of edited articles for AI detection comparison via script. Currently, each `ArticleCourse` record includes a `first_revision` timestamp, but none of the relevant revision IDs. (Similarly, the `ArticleCourseTimeslice` records have a timestamp for the first revision within each timeslice, but nothing about the last known revision and no revision IDs.)

A way of generating a diff URL from an AC record that relies on making API queries would be an acceptable solution. At some point, we might want to make changes to the database and the timeslice processing so that we store all the necessary info in the database, but just being able to do a similar thing via API queries (like the client-side DiffViewer) would be the best way to start I think.

costajohnt:
## Summary

- Adds a `CumulativeDiffUrlBuilder` service that queries the MediaWiki API to find the earliest and latest revisions made by a course's editors within the course date range
- Adds a `cumulative_diff_url` instance method to `ArticlesCourses` that delegates to the builder
- Includes specs covering multi-user, single-user, empty users, missing page, nil API response, and new-article (parentid=0) cases

This enables cumulative diff URLs to be generated server-side for use in alert emails and batch scripting, without relying on the frontend `DiffViewer` component.

## Test plan

- [x] Verify `ArticlesCourses#cumulative_diff_url` returns a valid diff URL for a record with multiple editors
- [x] Verify it returns `nil` gracefully when there are no editors or the article has no revisions
- [x] Run `bundle exec rspec spec/lib/cumulative_diff_url_builder_spec.rb` (6/6 passing)

Closes #6749

### DIFF
--- a/spec/lib/cumulative_diff_url_builder_spec.rb
+++ b/spec/lib/cumulative_diff_url_builder_spec.rb
@@ -3,61 +3,112 @@
 require 'rails_helper'
 require "#{Rails.root}/lib/cumulative_diff_url_builder"
 
+# Examples in this spec use real users, articles, and dates from this ended course:
+# https://dashboard.wikiedu.org/courses/Aquinas_College/Advanced_Communication_Theory_(Fall_2025)
+# (start: 2025-08-18, end: 2025-12-10)
 describe CumulativeDiffUrlBuilder do
-  let(:article) { create(:article) }
-  let(:course) { create(:course, start: '2024-06-16', end: '2024-08-16') }
-  let(:user1) { create(:user, username: 'StudentA') }
-  let(:user2) { create(:user, username: 'StudentB') }
-  let(:articles_course) do
-    create(:articles_course, article:, course:, user_ids: [user1.id, user2.id])
-  end
-
-  before do
-    travel_to Date.new(2024, 7, 16)
-  end
-
-  after do
-    travel_back
-  end
+  let(:course) { create(:course, start: '2025-08-18', end: '2025-12-10') }
 
   describe '#url' do
-    context 'when both users have revisions' do
-      before do
-        allow_any_instance_of(WikiApi).to receive(:query).and_wrap_original do |_m, params|
-          response = instance_double(MediawikiApi::Response)
-          allow(response).to receive(:data).and_return(
-            build_response(params[:rvuser], params[:rvdir])
-          )
-          response
-        end
+    context 'when both editors contributed to the article' do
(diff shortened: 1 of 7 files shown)


spec/lib/cumulative_diff_url_builder_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+  let(:article) { create(:article) }
+  let(:course) { create(:course, start: '2024-06-16', end: '2024-08-16') }
+  let(:user1) { create(:user, username: 'StudentA') }
+  let(:user2) { create(:user, username: 'StudentB') }
+  let(:articles_course) do
+    create(:articles_course, article:, course:, user_ids: [user1.id, user2.id])
+  end
+
+  before do
+    travel_to Date.new(2024, 7, 16)
```
ragesoss: What's this for?

spec/lib/cumulative_diff_url_builder_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+  end
+
+  after do
+    travel_back
+  end
+
+  describe '#url' do
+    context 'when both users have revisions' do
```
ragesoss: These tests are not as easy to understand and extend as I'd like.

I think the tests for this should make real API requests and use VCR, rather than mocking a response.

The examples ought involve real users, articles and course dates. Since we already have a working implementation of this on the frontend, the examples can be pulled from browsing already-ended courses on the production dashboard, such as https://dashboard.wikiedu.org/courses/Aquinas_College/Advanced_Communication_Theory_(Fall_2025)/articles/edited?

ragesoss (commented):
apologies! i had drafted a review, and never submitted it. here are my comments from the last time i looked at it.

ragesoss:
LGTM

Issue #6727, opened 2026-03-12
Identify and fix a deprecation warning


ragesoss:
We have a number of deprecation warnings and other line noise in various aspects of the system. Find *one* such warning, identify the cause, leave a note about it in a comment, then find the simplest way to fix it.

(This is a newcomer-friendly issue intended to provide an easy way to get started working with the Dashboard codebase, which we'll keep open perpetually. Don't take on more than one or two warnings without discussing it first.)

Places to look include:

* Warnings during JavaScript compilation (seen in output of `yarn build`)
  * These are typically warnings for hard-coded user-facing text strings, indicating that I18n should be used instead. Search for an existing interface message in the `en.yml` locale, and use that if available. If no existing interface message is available, identify the appropriate section of the `en.yml` hierarchy to add a new message. Only include changes for a single component file, or a single message that is used across multiple components, in your PR.
* Warnings or console errors from the JavaScript test suite (seen in output of `yarn test`
* Warnings from the Rails environment (seen when starting a development server, or running rspec tests, or installing Ruby dependencies, or anywhere else)
* Anywhere else you notice a warning or unhelpful line noise

sunkireddy-Barath:
Hi @ragesoss , I am going to work on this issue and  I will open a PR shortly.

Thank You!

sunkireddy-Barath:
Hello @ragesoss , I  found the  deprecation in the i18n warning happens because the headers Training Status and Recent Uploads are directly written in user_training_status.jsx and user_uploads.jsx instead of using translation keys. When running yarn build, this gets flagged since those texts can’t be translated into other languages. To fix it, replace the hardcoded strings with I18n.t() and add the matching keys in the en.yml file.

So I am working on this issue after completing it I will open a PR.

Thank You!

ArchieTyagi:
@ragesoss I'd like to pick up a deprecation warning. I'll run yarn build and bundle exec rspec locally, find one that wasn't covered in #6774, and submit a fix.

sunkireddy-Barath:
@ragesoss,
I’m picking up the i18n issue for the edit button in block.jsx. I’ll work on it and submit a PR once it’s ready.

PR #6754, opened 2026-03-25
Fix hardcoded strings in embed_stats_button.jsx and add localized str

sunkireddy-Barath:
What this PR does:

This PR fixes an issue where some UI text in the EmbedStatsButton component was hardcoded in English and did not change with language selection.

Changes made:
Replaced hardcoded strings with I18n.t() in embed_stats_button.jsx
Added translation keys in en.yml and es.yml
Ensured button + modal text updates based on selected language

Fixed:#6727

AI usage:
I used an AI tool only to identify the source of the warnings. The actual implementation, code changes, and verification were done manually.

Before screenshort: 
The "Embed Course Stats" button and modal text were hardcoded in English, so they didn’t change when switching languages.

After screenshort: 
All text is now translated using I18n, so the button and modal update correctly based on the selected language.

Screenshot from 2026-03-25 17-43-34

Open questions and concerns:
No concerns. Follows existing I18n patterns and works correctly in both English and Spanish.

### DIFF
--- a/app/assets/javascripts/components/overview/embed_stats_button.jsx
+++ b/app/assets/javascripts/components/overview/embed_stats_button.jsx
@@ -8,7 +8,7 @@ const EmbedStatsButton = ({ title }) => {
   const copyToClipboard = (e) => {
     const el = e.target;
     navigator.clipboard.writeText(el.value).then(() => {
-      setStatus('Copied!');
+      setStatus(I18n.t('courses.copied'));
     });
   };
 
@@ -21,12 +21,12 @@ const EmbedStatsButton = ({ title }) => {
   return (
     <div className="basic-modal course-stats-download-modal embed_stats">
       <button onClick={() => setShow(false)} className="pull-right article-viewer-button icon-close" />
-      <h4>{I18n.t('courses.embed_course_stats_description')} <code>&lt;body&gt;</code></h4>
+      <h4>{I18n.t('courses.embed_course_stats_description')} <code>{I18n.t('courses.embed_course_stats_body')}</code></h4>
       <textarea
         id="embed"
         readOnly
         value={
-          `<a href="${location.href}">${title}</a><!-- This is optional -->
+          `<a href="${location.href}">${title}</a>${I18n.t('courses.embed_course_stats_optional')}
 <iframe src="${url}" style="width:100%;border:0px none transparent;"></iframe>`}
         onClick={copyToClipboard}
       />
--- a/app/assets/javascripts/components/overview/embed_stats_button.jsx
+++ b/app/assets/javascripts/components/overview/embed_stats_button.jsx
@@ -21,13 +21,11 @@ const EmbedStatsButton = ({ title }) => {
   return (
     <div className="basic-modal course-stats-download-modal embed_stats">
       <button onClick={() => setShow(false)} className="pull-right article-viewer-button icon-close" />
-      <h4>{I18n.t('courses.embed_course_stats_description')} <code>{I18n.t('courses.embed_course_stats_body')}</code></h4>
+      <h4>{I18n.t('courses.embed_course_stats_description')} <code>{I18n.t('courses.embed_course_stats_location')}</code></h4>
       <textarea
         id="embed"
         readOnly
-        value={
-          `<a href="${location.href}">${title}</a><!-- This is optional -->
-<iframe src="${url}" style="width:100%;border:0px none transparent;"></iframe>`} // eslint-disable-line i18next/no-literal-string
+        value={I18n.t('courses.embed_course_stats_body', { location: window.location.href, title, url })}
         onClick={copyToClipboard}
       />
       <small>{status}</small>
--- a/app/assets/javascripts/components/overview/embed_stats_button.jsx
+++ b/app/assets/javascripts/components/overview/embed_stats_button.jsx
@@ -26,8 +26,8 @@ const EmbedStatsButton = ({ title }) => {
         id="embed"
         readOnly
         value={
-          `<a href="${location.href}">${title}</a>${I18n.t('courses.embed_course_stats_optional')}
-<iframe src="${url}" style="width:100%;border:0px none transparent;"></iframe>`}
+          `<a href="${location.href}">${title}</a><!-- This is optional -->
+<iframe src="${url}" style="width:100%;border:0px none transparent;"></iframe>`} // eslint-disable-line i18next/no-literal-string
         onClick={copyToClipboard}
       />
       <small>{status}</small>
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -744,6 +744,9 @@ en:
     edit_course_dates: Edit Course Dates
     embed_course_stats: Embed Course Stats
     embed_course_stats_description: To embed the stats for this program in another website, copy and paste this code snippet into
+    embed_course_stats_body: " <body>"
+    embed_course_stats_optional: "<!-- This is optional -->"
+    copied: Copied!
     enable_account_requests: Enable account requests
     end: End
     ended: The course has ended.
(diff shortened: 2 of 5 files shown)


spec/rails_helper.rb
```diff
@@ -120,6 +120,8 @@
           # some specs test behavior for 4xx responses and other errors.
           # Don't fail on these.
           next if /Failed to load resource/.match?(error.message)
+          next if /WebSocket connection/i.match?(error.message)
```
ragesoss: This doesn't seem relevant to the PR.

config/locales/es.yml
```diff
@@ -683,6 +683,11 @@ es:
     dismiss_survey_confirm: ¿Estás seguro de que quieres descartar permanentemente
       esta encuesta?
     download_stats_data: Estadísticas de descargas
+    embed_course_stats: Incrustar estadísticas del programa
```
ragesoss: translations should not be included in the PR. we'll rely on translatewiki.net for them.

app/assets/javascripts/components/overview/embed_stats_button.jsx
```diff
@@ -21,12 +21,12 @@ const EmbedStatsButton = ({ title }) => {
   return (
     <div className="basic-modal course-stats-download-modal embed_stats">
       <button onClick={() => setShow(false)} className="pull-right article-viewer-button icon-close" />
-      <h4>{I18n.t('courses.embed_course_stats_description')} <code>&lt;body&gt;</code></h4>
+      <h4>{I18n.t('courses.embed_course_stats_description')} <code>{I18n.t('courses.embed_course_stats_body')}</code></h4>
       <textarea
         id="embed"
         readOnly
         value={
-          `<a href="${location.href}">${title}</a><!-- This is optional -->
```
ragesoss, line 29: This is a weird one, because it's an HTML comment meant to be ready by someone seeing and copying the raw HTML that they want to use to embed an iframe of stats on their own website.

If we're going to enable I18n for it, only the message itself, not the HTML comment markup, should be translated. But I think this one is better off not using I18n. If there's a way to disable the warning for this particular line, that would be better.

sunkireddy-Barath:
In the spec/rails_helper.rb file why i added that two line because 

The tests were failing because the browser was trying to connect to a development tool that isn't running during CI tests. These two lines tell the tests to ignore that harmless error so we can see the real test results.

Now i will resolve that issue 
Thank You!

sunkireddy-Barath:
Hi @ragesoss,
Could you please review the changes when you have a moment? If there’s anything I should improve or modify, 

Thank you!

config/locales/en.yml
```diff
@@ -743,6 +743,8 @@ en:
     edit_course_dates: Edit Course Dates
     embed_course_stats: Embed Course Stats
     embed_course_stats_description: To embed the stats for this program in another website, copy and paste this code snippet into
+    embed_course_stats_body: "<body>"
```
ragesoss: This one is HTML, so it shouldn't be translated. This would be another appropriate place to disable the warning inline.

ragesoss:
Only English messages should be added in most cases; we rely on translatewiki.net for translations. It's okay to include translations for a new message if you know the language, so (assuming you are fluent in Spanish) there's no need to remove them in this PR, but I'd generally prefer just adding `en.yml` (and `qqq.yml` if the context and meaning aren't self-explanatory) for new messages in most cases.

app/assets/javascripts/components/overview/embed_stats_button.jsx
```diff
@@ -21,13 +21,13 @@ const EmbedStatsButton = ({ title }) => {
...
       <button onClick={() => setShow(false)} className="pull-right article-viewer-button icon-close" />
-      <h4>{I18n.t('courses.embed_course_stats_description')} <code>&lt;body&gt;</code></h4>
+      <h4>{I18n.t('courses.embed_course_stats_description')} <code>{I18n.t('courses.embed_course_stats_body')}</code></h4>
       <textarea
         id="embed"
         readOnly
         value={
           `<a href="${location.href}">${title}</a><!-- This is optional -->
-<iframe src="${url}" style="width:100%;border:0px none transparent;"></iframe>`}
+<iframe src="${url}" style="width:100%;border:0px none transparent;"></iframe>`} // eslint-disable-line i18next/no-literal-string
```
ragesoss: Is this the right place? I thought the idea was to apply this to the body tag?

sunkireddy-Barath:
Hi @ragesoss,
I’ve moved the embed snippet into a translation key and updated the instruction to clearly indicate <body> as the paste location, while removing the hardcoded HTML and ESLint suppression. Please let me know if any further adjustments are needed.

ragesoss (commented):
A note on AI in review feedback: this comment was drafted with Claude Code. It is provisional and may misread context — read, judge, and use or ignore as appropriate.

Thanks for the iteration. Looking at the latest commit (`b783753b`), a few of the earlier review points still apply, and the new approach moves in the opposite direction from what the prior comments were pointing at.

**Direction of the latest commit.** Two prior comments specifically said the HTML pieces — the `<body>` literal and the inline HTML snippet (`<a>…<iframe>…<!-- This is optional -->`) — should *not* be put through I18n, and that disabling the lint warning inline would be the better approach. The latest commit goes the other way: it puts the entire HTML snippet into `embed_course_stats_body` and the literal `<body>` into `embed_course_stats_location`. Translators on translatewiki.net would then be asked to "translate" raw HTML, which is the outcome the earlier comments were trying to avoid.

**Wrong tool for the bypass.** The `# i18n-tasks-use` marker added to `embed_course_stats_body` belongs to the `i18n-tasks` Ruby gem, which doesn't scan JSX. The warning being silenced is from `eslint-plugin-i18next` (see `eslint.config.mjs`: `'i18next/no-literal-string': ['warn', ...]`). The pattern that does silence that rule is already used elsewhere in this repo, e.g. `app/assets/javascripts/training_module_composer/components/slide_editor.jsx:1`:

```js
/* eslint-disable i18next/no-literal-string */
```

A file-level disable at the top of `embed_stats_button.jsx`, combined with `I18n.t()` calls only for the actual user-facing strings (`Embed Course Stats`, the description, `Copied!`), would address the warning without putting HTML into the translation system.

**Sketch of an alternative.** Roughly:

```jsx
/* eslint-disable i18next/no-literal-string */
// ...
<h4>{I18n.t('courses.embed_course_stats_description')} <code>&lt;body&gt;</code></h4>
<textarea
  id="embed"
  readOnly
  value={
    `<a href="${location.href}">${title}</a><!-- This is optional -->
<iframe src="${url}" style="width:100%;border:0px none transparent;"></iframe>`}
  onClick={copyToClipboard}
/>
```

Only `courses.embed_course_stats`, `courses.embed_course_stats_description`, and `courses.copied` would be new translation keys; `embed_course_stats_body` and `embed_course_stats_location` could be removed.

**Other points still open / new:**

- `config/locales/es.yml` still has three unrelated edits (a quote-pair typo fix in `course_removed_but_not_deleted`, and deletions of `dismiss_survey_confirm` and `download_stats_data`). Per the earlier comment about relying on translatewiki.net for non-English translations, these can be dropped from the PR.
- `config/locales/qqq.yml` includes qqq descriptions for `edit_course_dates`, `embed_course_stats`, and `embed_course_stats_description`, which aren't keys this PR introduces — those can also be dropped.
- The newly-translated body snippet changes the actual rendered embed code from two lines (newline between the `<a>…<!-- -->` and the `<iframe>`) to one line (a single space). If the body translation is removed and the original template literal is restored, this goes away on its own.
- The branch is now far behind master; a rebase will be needed before this can merge.

*Drafted in a Claude Code session (~10 min, ~10 interactions). Sage directed the re-review, read the prior comment thread, and reviewed the full draft before posting. The code suggestions were proposed by Claude but not run locally — they should be tried before relying on them. (Comment written by Claude Code.)*

PR #6757, opened 2026-03-26
Allow Declining Submitted Courses Gracefully

### LINKED ISSUE #6728: Add a way for admins to archive submitted courses
In Wiki Education Dashboard (not Programs & Events Dashboard), the workflow for a new course is:

* Instructor creates course (and goes through assignment design wizard)
* After reviewing and making any desired changes, instructor submits course
* Wiki Education staff review submitted course, and:
  * approve it (adding it to a campaign, adding tags, adding the staff who will support that course); or
  * contact the instructor if changes should be made or there are concerns; or
  * unsubmit it if that course will definitely not be approved (for example, because it's not in the US or Canada, so it would need to be run without Wiki Education support, possibly using Programs & Events Dashboard)

For that last situation, unsubmitting it is a workaround for the lack of a dedicated way to 'archive' or 'decline' a course. The unsubmitted course will still trigger the creation of an `UnsubmittedCourseAlert` (via `UnsubmittedCourseAlertManager`) and send an email to the instructor, which often leads to that instructor resubmitting the course.

We need a way to 'decline' a submitted course, such that it won't trigger that alert. The 'declined' status could be represent by a course flag. The flag could be added via a frontend UI added to the `CourseApproval` component.

nexpectArpit:
This PR introduces a dedicated "Decline Course" workflow to address  the problem where administrators manually unsubmitting courses trigger an unending cycle of "Unsubmitted Course" alert emails. By utilizing a new persistent course flag (`declined: true`), administrators can now officially throw a course back to the instructor for draft rework, completely halting any automated nagging emails from the `UnsubmittedCourseAlertManager`

Closes #6728 

## Screenshots
Before:
Screenshot 2026-03-26 at 5 05 18 PM

After:

https://github.com/user-attachments/assets/a1cb3400-963e-4bb2-ab1e-b5ec58383232

@ragesoss please review it.

### DIFF
--- a/app/assets/stylesheets/modules/_notifications.styl
+++ b/app/assets/stylesheets/modules/_notifications.styl
@@ -30,7 +30,7 @@
       color #fff
 
   &.declined
-    background-color warning
+    // Default green background from .notification will be used
 
 .notice
   @extends .notification
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -976,7 +976,7 @@ en:
     word_count_doc: An estimate of the number of words added to mainspace articles by the course's students during the course term
     word_count_doc_wikidata: An estimate of the number of words added to mainspace items by the course's students during the course term
     yourcourses: Your Courses
-    declined_note: "This course has been declined by staff."
+    declined_note: "This course has been declined. Please contact Wiki Education staff if you have questions."
     decline_course: "Decline Course"
     decline_course_confirm: "Are you sure you want to decline this course? This will archive the submission and stop automatic unsubmitted alerts for this course."
     survey:


ragesoss:
This looks solid overall. The intention is to make 'decline' something separate from 'unsubmit', so that there's no prompt for the instructor to click 'submit' again after it's been declined. If it's headed back to the instructor to make changes, we'll still just unsubmit it so they can resubmit. Decline is intended for cases where it definitely won't be accepted, so the 'submit' banner should not appear and there should instead be a clear indication of the 'declined' state in the UI (probably via a similar banner to the submit one).

nexpectArpit:
@ragesoss i changed the logic. also added the updated recording, please re-review it

ragesoss:
I think the banner should not be red. Keeping the green color is fine.

Message should be "This course has been declined. Please contact Wiki Education staff if you have questions."

nexpectArpit:
Screenshot 2026-03-28 at 5 25 15 AM
have a look

ragesoss:
LGTM. Tested locally, and it seems to work just as expected.

PR #6758, opened 2026-03-26
Fix/image tooltip pluralization 6738

sunkireddy-Barath:
What this PR does:

This PR fixes a pluralization issue in the image usage count displayed on the Uploads tab.
Previously, when an image was used in a single article, it incorrectly displayed:
"Used in 1 articles"
This PR ensures correct grammatical handling by updating the implementation to display:
"Used in 1 article" (singular)
"Used in X articles" (plural)

Changes made:
Updated translation keys in config/locales/en.yml to support proper pluralization.
Modified upload.jsx to use I18n.t with the count parameter for automatic plural handling.
Ensured compatibility with existing i18n conventions in the codebase.

AI usage :
I used an AI tool only to identify the source of the warnings. The actual implementation, code changes, and verification were done manually.

Screenshots:
Before
Incorrect pluralization for singular case
Screenshot from 2026-03-26 22-15-45

After:
Correct singular and plural display
Screenshot from 2026-03-26 23-36-55

Open questions / concerns:
None. The change is scoped, minimal, and does not affect other parts of the system.

### DIFF
--- a/app/assets/javascripts/components/uploads/upload.jsx
+++ b/app/assets/javascripts/components/uploads/upload.jsx
@@ -51,7 +51,7 @@ const Upload = ({ upload, view, linkUsername }) => {
 
   let usage = '';
   if (upload.usage_count) {
-    usage = `${I18n.t('uploads.usage_count_gallery_tile', { usage_count: upload.usage_count })}`;
+    usage = `${I18n.t('uploads.usage_count_gallery_tile', { count: upload.usage_count })}`;
   }
 
   let uploadDivStyle;
@@ -68,7 +68,7 @@ const Upload = ({ upload, view, linkUsername }) => {
       <p className="tablet-only">
         <span>{upload.uploader}</span>
         <span>&nbsp;|&nbsp;</span>
-        <span>Usages: {upload.usage_count}</span>
+        <span>{I18n.t('uploads.usages')}: {upload.usage_count}</span>
       </p>
     );
   } else {
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1647,7 +1647,9 @@ en:
     uploaded_on: Uploaded On
     usage_count: Usage Count
     usage_doc: Number of Wikipedia articles that this image appears in.
-    usage_count_gallery_tile: Used in %{usage_count} articles
+    usage_count_gallery_tile:
+      one: "Used in 1 article"
+      other: "Used in %{count} articles"
     usages: Usages
     gallery_view: Gallery View
     list_view: List View
--- a/config/locales/qqq.yml
+++ b/config/locales/qqq.yml
@@ -696,8 +696,10 @@ qqq:
       only)
     show_courses: Checkbox for option to show user courses only (Wiki Ed only)
     uploaded_by: Label for the user who uploaded a file
-    usage_count: Label for the number of global usages of a file across Wikimedia
-      sites
+    usage_count: Label for the number of global usages of a file across Wikimedia projects
+    usage_count_gallery_tile:
+      one: "Label for when the media file is used in exactly one article"
+      other: "Label for when the media file is used in many articles. {{PLURAL:%1$d|1=Used in 1 article|Used in %1$d articles}}"
   revisions:
     chars_added: Label for counter of added chars
     class: |-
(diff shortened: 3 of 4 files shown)


sunkireddy-Barath:
Hello @ragesoss,
I accidentally closed the PR earlier, but everything is now fine. Could you please review the files and let me know if everything looks good or if any changes are needed?

Thank you!

ragesoss:
Thanks! Please clean up the git history; it should probably just include 1 commit. It would also be helpful to check out the related specs, and see if there's an easy way to adjust them to cover the new behavior. (I don't think adding a new test that increases the run time of the test suite makes sense for a small change like this, but if an existing test can be improved to cover this, that would be great.)

sunkireddy-Barath:
Thanks @ragesoss I’ll squash the commits into a single clean commit and check if an existing spec can be updated to cover this behavior.
Thank You!

ragesoss:
It doesn't look like the spec change interacts with this PR, is that correct?

I think the `course_page_spec.rb` 'uploads view' section is the right place to try to extend the specs to cover this. Probably the first spec in that example would be a good place for it. The tricky part will be to make Capybara do the mouseover behavior so the message will be shown, but I think that's possible using `.hover`. You can find an example of using that Capybara feature in the `timeline_editing_spec`.

sunkireddy-Barath:
Hi @ragesoss I have completed what you said and While CI testing It is failed due to this  category_scopes_spec.rb due to thsi issue the CI test is failed

Thank you!

ragesoss:
The category one is a flaky test, but the course_page_spec file you edited is also failing.

sunkireddy-Barath:
@ragesoss 
I’ve made the requested changes, and the updated file is now passing all CI tests successfully. Could you please take a moment to review the PR when you have time?

ragesoss (changes requested):
The core fix is correct — `uploads.usage_count_gallery_tile` now uses Rails i18n pluralization properly, and the `count` parameter in `upload.jsx` is the right approach. The hover-based spec testing singular/plural is also a good addition, as discussed.

A few things to address before this can be merged:

**Must fix:**
- **Remove `js_error_expected: true`** from the top-level `describe` block (line 33). This flag suppresses JS error detection for all 32 specs in the file. Verified that all specs pass without it — it's not needed and would mask real JS errors going forward.
- **Squash to a single commit**, as noted in the earlier review.
- **Remove unrelated spec changes.** The PR currently modifies many specs beyond the uploads section — replacing `sleep` with Capybara waits, switching `visit` to `js_visit`, adding selector guards before button clicks. These are reasonable improvements on their own, but they expand the scope well beyond the pluralization fix and make the PR harder to review. Please revert all `course_page_spec.rb` changes outside the `uploads view` describe block. (Some of these improvements may be adopted separately.)
- **Remove `visible: false` from the hover assertions.** The point of the hover test is to verify the tooltip becomes visible when the user hovers. Using `visible: false` bypasses that check — Capybara will find the element whether or not it's actually shown. Remove that option so the test confirms the tooltip is truly visible after hovering.

*Drafted in a Claude Code session (~15 min, ~10 interactions). Sage directed the review, verified the `js_error_expected` flag was unnecessary by checking spec results with and without it, and reviewed this comment before posting. Spec runs and lint checks were performed by Claude Code.*

ragesoss (commented):
Thanks for the cleanup — all four items from the previous round are addressed in ac02b327.

One additional issue surfaced when running the spec locally: hovering an `.upload` causes its `.info` block to switch from `display: none` to `display: flex` (defined in `app/assets/stylesheets/modules/_uploads.styl:40-44`), which exposes the file_name `<a>` link inside `.upload[role="button"]` to axe-core's nested-interactive rule. `expect(page).to be_axe_clean` then fails. The accessibility violation is pre-existing in the upload component — not introduced by this PR — but the spec's hover sequence triggers it.

This was resolved by moving the `be_axe_clean` assertion above the hover sequence. The PR commit plus that small follow-up fix have been cherry-picked to master with original authorship preserved, so this PR will close automatically when those land. The underlying nested-interactive issue in `.upload[role="button"]` will be fixed separately.

*Drafted in a Claude Code session (~45 min, ~25 interactions). Sage directed the review, picked the easy fix over the more ambitious one, and reviewed this comment before posting. The spec was run ~10 times locally to isolate the trigger and confirm the fix; linters passed; cherry-picks were applied to master.*

ragesoss:
Merged manually via cherry-pick (cbb6de1cf) plus a follow-up commit (d11c0eeb8) that reorders the axe-clean assertion. Closing this PR — the original commit's authorship is preserved on master. Thanks for the contribution.

Issue #6331, opened 2025-05-22
Inflated 'characters added' stat for article with deleted revisions


ragesoss:
See https://dashboard.wikiedu.org/courses/University_of_Florida/Discrimination_and_Health_(Spring_2025)/articles/edited

Student editor Rhitorical edited the article "Ball culture", beginning on February 28, 2025. The immediately preceding revision has been removed: https://en.wikipedia.org/w/index.php?title=Ball_culture&action=history

The Dashboard shows a `character sum` of 78018 (15076 words), which is basically the entire size of the page. It appears that the first revision has been credited complete to that editor, instead of just the 1,842 bytes actually added in that first revision.

lakshita10341:
Hi @ragesoss, I investigated this issue and found the root cause. The bug is in the WikiEduDashboardTools. In [`revisions.php`](https://github.com/WikiEducationFoundation/WikiEduDashboardTools/blob/master/public_html/revisions.php), the `byte_change` is calculated as:

```sql
CAST(c.rev_len AS SIGNED) - CAST(IFNULL(p.rev_len, 0) AS SIGNED) AS byte_change
```
When the parent revision is deleted, the `LEFT JOIN` returns `NULL` for the parent's length, and `IFNULL(..., 0)` treats it as 0. This causes `byte_change` to equal the entire current revision size instead of the actual difference  which is exactly what we're seeing here.

**Proposed fix:**
When the parent revision is deleted, find the next available ancestor revision that still exists and use its size as the baseline:

```sql
SELECT rev_len FROM revision 
WHERE rev_page = c.rev_page 
  AND rev_id < c.rev_id 
ORDER BY rev_id DESC 
LIMIT 1
```

This way, we get an accurate (or close to accurate) `byte_change` even when the immediate parent is deleted.
Please let me know if I should proceed with this fix!

ragesoss:
@lakshita10341 that sounds like it would work, although it might have significant performance implications, and would need some robust tests. If you're going to work on this, I think you'll need to get access to Toolforge so that you can build and test it against the real Replica database.

lakshita10341:
Hi @ragesoss , I have doubt about Toolforge, how to access it, I searched for it, but was not able to properly understand. Can you provide some documentation or something, so that i  can proceed with this issue solving.

ragesoss:
@lakshita10341 Here is the documentation for Toolforge: https://wikitech.wikimedia.org/wiki/Help:Toolforge

[...]

PR #6760, opened 2026-03-27
test: add regression test for suppressed parent revision bug

lakshita10341:
## What this PR does
This PR addresses the issue of inflated character count statistics on the dashboard (#6331 ). This typically occurs for edits where the parent revision has been suppressed. In these cases, the standard replica queries at `replica-revision-tools.wmcloud.org` fail to obtain the parent revision's size, leading to the entire page size being reported as the edit's character count.

Changes include:
- A new regression test in `spec/lib/replica_spec.rb` that verifies correct character count retrieval for a known problematic revision (mw_rev_id: 1278154303).
- A merged VCR cassette (`suppressed_parent_revision.yml`) containing both the buggy production response and the correct response from the fixed tool.
- An update to `REPLICA_TOOL_URL` in `lib/replica.rb` to point to a fixed Toolforge instance (`https://fix-stats.toolforge.org/`) that correctly calculates character deltas even when parent data is missing or suppressed.
- 
## AI usage
- Identify and fix syntax errors  in the newly added regression test.

## Screenshots
N/A(backend only test)

### DIFF
--- a/spec/lib/replica_spec.rb
+++ b/spec/lib/replica_spec.rb
@@ -339,7 +339,9 @@ describe Replica do
     it 'returns correct characters for a revision with a suppressed parent revision' do
       VCR.use_cassette 'cached/replica/suppressed_parent_revision' do
         user = build(:user, username: 'Rhitorical')
-        response = described_class.new(en_wiki).get_revisions([user], 2025_02_01_000000, 2025_06_01_000000)
+        response = described_class.new(en_wiki)
+                                  .get_revisions([user], 2025_02_01_000000,
+                                                 2025_06_01_000000)
         rev = response['9171866']['revisions'].find { |r| r['mw_rev_id'] == '1278154303' }
 
         expect(rev['characters'].to_i).to eq(1842)


ragesoss:
Looks like for testing this you'll need to replace the URL in some of the tests as well. (Refactoring the tests so that they work automatically for whichever URL `replica.rb` points to would be a nice improvement, but out of scope for this PR.)

lakshita10341:
This PR is for testing purpose only. We are not going to merge it, right? So can  I refactor all the tests in this PR so that these tests pass successfully, and we can verify the fix. Will it work? Or a new PR is needed which is URL independent?

ragesoss:
@lakshita10341 no need for a new PR, as indeed we're not going to merge it, but I would recommend just replacing all the URLs in the specs (rather than actually refactoring) for this PR. It's good to keep refactors totally separate from functional changes, so that we know what is breaking from a functional change vs what might be a broken refactoring.

lakshita10341:
@ragesoss I have changed the url in the tests, and all the tests are passing successfully now. PTAL when you get time.

ragesoss:
Great! I'm going to merge and deploy https://github.com/WikiEducationFoundation/WikiEduDashboardTools/pull/20

Please open a new PR that adds the new test (without the URL changes).

ragesoss:
Also, please don't add this VCR cassette as cached. It's a fast query, and we aren't caching any other replica requests, so I'd rather have the specs run against the live tool so that we can quickly spot problems if the response changes.

lakshita10341:
Okay, sure I'll add a new PR and will not add the VCR cassette as cache.

PR #6766, opened 2026-03-30
Add regression test for suppressed parent revision bug

lakshita10341:
## What this PR does
This PR addresses the issue of inflated character count statistics on the dashboard (#6331 ). This typically occurs for edits where the parent revision has been suppressed. In these cases, the standard replica queries at `replica-revision-tools.wmcloud.org` fail to obtain the parent revision's size, leading to the entire page size being reported as the edit's character count.

Changes include:
- A new regression test in `spec/lib/replica_spec.rb` that verifies correct character count retrieval for a known problematic revision (mw_rev_id: 1278154303).

## AI usage
- Identify the bug behind failing test cases.

## Screenshots
N/A(backend only test)

##Open questions and concerns
The tests in `replica_spec.rb` and `article_status_manager_timeslice_spec.rb` are  failing when I am testing on my local and also in my previous PR. The root cause seems to be that POST requests to the replica API endpoint  are returning `nil` instead of the expected article data.

I've found that adding the following line to the request in `lib/replica.rb` fixes the failures:
```ruby
req.add_field('Content-Type', 'application/x-www-form-urlencoded')
```
This line explicitly tells the server that we're sending a standard HTML form body, which allows it to correctly parse the article titles we're sending. Without it, the server doesn't understand the request and returns `nil`.

So I have a doubt, how these tests were passing previously successfully without adding this.

### DIFF
--- a/spec/lib/replica_spec.rb
+++ b/spec/lib/replica_spec.rb
@@ -333,4 +333,21 @@ describe Replica do
       expect(result).to be_empty
     end
   end
+
+  # Regression test for the suppressed parent revision bug.
+  describe 'suppressed parent revision' do
+    it 'returns correct characters for a revision with a suppressed parent revision' do
+      VCR.use_cassette 'replica/suppressed_parent_revision' do
+        user = build(:user, username: 'Rhitorical')
+        response = described_class.new(en_wiki)
+                                  .get_revisions([user], 2025_02_01_000000,
+                                                 2025_06_01_000000)
+        rev = response['9171866']['revisions'].find { |r| r['mw_rev_id'] == '1278154303' }
+
+        expect(rev['characters'].to_i).to eq(1842)
+      end
+    end
+  end
 end
+
+


ragesoss:
I've now updated the live version of the replica tool, so this test should pass. When I tested it locally after deploying, the new test passed without the `add_field`. Does it still fail for you, or was that only happening when you tested against your toolforge.org tool? If it's the latter, it may be something that is part of the shared toolforge infrastructure or the webservice system it uses to serve PHP tools; on https://replica-revision-tools.wmcloud.org/ we run the service on Apache instead of the `lighttpd` (see https://wikitech.wikimedia.org/wiki/Help:Toolforge/Web).

However, it seems like adding that field is benign and will be useful for future cases where we want to test against changes to the tool. If you want to do a PR to refactor the tests to have a single place where the tool URL is defined (in `replica.rb`), that would be a good place to also add the 'Content-Typle' field.

lakshita10341:
Actually the tests were still failing on local, I cleared all previous cassettes but still they are failing. If it passed for you, so may be there is some caching issue , I'll check once again in my local docker environment. 
And yes, I will create a new, separate PR for the refactor to define the tool URL in a single place. I'll get on that after I've sorted this out.

lakshita10341:
>However, it seems like adding that field is benign and will be useful for future cases where we want to test against changes to the tool. If you want to do a PR to refactor the tests to have a single place where the tool URL is defined (in replica.rb), that would be a good place to also add the 'Content-Type' field.

Hi @ragesoss, for refactoring , should I create a new issue or can I directly create a PR?

ragesoss:
@lakshita10341 you can create the PR directly.

PR #6792, opened 2026-04-09
Replace hardcoded strings with I18n in ArticleViewer components (#6727)

archietyagi100-tech:
## What this PR does

This PR fixes two i18next/no-literal-string warnings that were showing up during `yarn lint`. The issue was caused by hardcoded English strings in the ArticleViewer components.

The following updates were made:
- Replaced "Quality Problems?" in `BadWorkAlertButton.jsx` with `I18n.t('instructor_view.bad_work.quality_problems_button')`
- Replaced "Notify Wiki Expert" in `SubmitIssuePanel.jsx` with `I18n.t('instructor_view.bad_work.notify_wiki_expert')`
- Added both keys to `config/locales/en.yml` under the existing `instructor_view.bad_work` section

The UI looks the same, but the text now comes from localization files and can be translated into other languages.

Addresses #6727

AI usage:
 I used AI to understand how I18n.t() works in the project and to check the correct key path format under instructor_view.bad_work. The actual changes and file edits were done by me after running yarn lint locally to find the warnings.

Screenshots:
These buttons appear inside the Article Viewer when an instructor reviews a student's article. The UI looks identical before and after  the only change is that the text now comes from I18n localization files instead of hardcoded strings, enabling proper translation when the Dashboard is used in other languages.

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/components/BadWorkAlert/SubmitIssuePanel.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/components/BadWorkAlert/SubmitIssuePanel.jsx
@@ -31,7 +31,7 @@ export const SubmitIssuePanel = ({ handleChange, handleSubmit, isSubmitting, mes
           value_key="message"
         />
         <button className="button danger" type="submit" disabled={isSubmitting}>
-          Notify Wiki Expert
+          {I18n.t('instructor_view.bad_work.notify_wiki_expert')}
         </button>
       </form>
     </article>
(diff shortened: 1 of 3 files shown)


ragesoss:
@archietyagi100-tech please use the PR template and include before/after screenshots.

archietyagi100-tech:
@ragesoss Updated the description . Let me know if anything needs changing.

ragesoss (approved):
**Review summary**

The PR branch was checked out locally, merged with current master, assets were rebuilt, and the existing `article_viewer_spec.rb` was run. Both specs — including the one that clicks "Quality Problems?" and "Notify Wiki Expert" — pass, confirming the i18n keys resolve to the correct strings.

The code changes are correct: two hardcoded strings replaced with `I18n.t()` calls under the existing `instructor_view.bad_work` key path, matching the surrounding pattern in both JSX files.

One minor note: the PR description doesn't include screenshots, which the PR template requests. In this case the existing feature spec serves as verification that the UI renders correctly.

*Drafted in a Claude Code session (~10 min, ~8 interactions). Sage directed the review approach (merge with master, build assets, run existing specs headed), reviewed the spec output, and approved this comment. The i18n key correctness was verified by the passing spec, not by manual inspection.*

PR #6794, opened 2026-04-11
Enable RuboCop rules  and fix offenses

nexpectArpit:
part of #5297 

This PR enables 5 RuboCop rules:
- `Lint/SymbolConversion`, 
- `Style/RedundantStringEscape`, 
- `Style/RedundantConstantBase`, 
- `Lint/RedundantCopDisableDirective`, 
- `Style/RedundantParentheses`.
and fixes the offense

### Verification
All thses RuboCop rules pass , and 52 relevant RSpec tests pass with 0 failures under Ruby 3.4.8.

@ragesoss please review it.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -269,8 +269,8 @@ Style/MultipleComparison:
   Enabled: false
 Style/MinMaxComparison:
   Enabled: false
-Lint/SymbolConversion:
-  Enabled: false
+# Lint/SymbolConversion:
+#   Enabled: false
 Style/CollectionQuerying:
   Enabled: false
 Style/RedundantEach:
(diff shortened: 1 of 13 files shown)


ragesoss (changes requested):
**`.rubocop.yml`: delete rather than comment out**

The disabled rules should be deleted from `.rubocop.yml`, not commented out. RuboCop enables rules by default when they aren't listed, so the commented lines are just clutter. Same for `Style/RedundantParentheses` — it wasn't in the file before and doesn't need to be added as a comment.

**Don't enable `Style/RedundantParentheses`**

The parentheses in cases like `course_meetings_manager.rb:104` and `revision_data_manager.rb:200` add clarity for compound boolean expressions. Leave `Style/RedundantParentheses` disabled (keep the `Enabled: false` entry) and revert the parentheses removals.

The other four rule enables and their corresponding fixes look correct.

*Drafted in a Claude Code session (~5 min, 3 interactions). Sage directed the two specific requests after reviewing the diff summary; other observations were not independently verified.*

Issue #6726, opened 2026-03-12
Category feature spec fails intermittently


ragesoss:
In CI, I've noted this one failing sometimes:

`rspec ./spec/features/category_scopes_spec.rb:63 # Tracked categories and templates lets a facilitator add multiple categories from different wikis at once`

It typically fails with an error like this:

```
  1) Tracked categories and templates lets a facilitator add multiple categories from different wikis at once
     Failure/Error: find(:css, '#categories div[class*="option"]', text: 'fr:Matériel Apple', exact_text: true).click # rubocop:disable Layout/LineLength

     Capybara::ElementNotFound:
       Unable to find css "#categories div[class*=\"option\"]"
     # ./spec/features/category_scopes_spec.rb:73:in 'block (2 levels) in <top (required)>'
```

I'm not sure what exactly goes wrong; it might be the result of an API request that sometimes fails in CI, or it might be a race condition or something else.

Identify the cause of this test's intermittent failure, and update the test to make it reliable.

SanjanaSogimatt:
Picking up this

SanjanaSogimatt:
Update on this- Not able to firgure out exactly when the test fails, so will spend some more time to figure what's going wrong here

SanjanaSogimatt:
Ok a small update on this issue - (work in progess)
I tired `HEADED=true rspec spec/features/category_scopes_spec.rb` in master branch and this branch

On master all passed expect - 
```
Failures:

  1) Tracked categories and templates lets a facilitator add and remove a category
     Failure/Error: find(:css, '#categories div[class*="option"]', text: 'Earth sciences').click
     
     Selenium::WebDriver::Error::StaleElementReferenceError:
       stale element reference: stale element not found in the current frame
         (Session info: chrome=147.0.7727.56); For documentation on this error, please visit: https://www.selenium.dev/documentation/webdriver/troubleshooting/errors#stale-element-reference-exception
```
On this branch with changes and all of them passed

The original CI failure was intermittent missing French option in cross-wiki category selection.
Local headed runs on master also reveal a separate intermittent stale-element failure on category option clicks.
Both point to async react-select instability
@ragesoss

PR #6798, opened 2026-04-14
[WIP]: Add stub for French category search in category scopes spec

SanjanaSogimatt:
## Changes

Resolves - https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6726

- Added `stub_fr_category_search_for_materiel_apple` in `spec/features/category_scopes_spec.rb` to stub the browser-side French category lookup for `Matériel Apple`.
- Updated the flaky example (`adds multiple categories from different wikis`) to use this stub before interacting with the category autocomplete.
- Added explicit synchronization around wiki/category option rendering so the test waits for the expected UI state before clicking.

This makes the spec deterministic in CI by removing dependence on live Wikipedia search responses and reducing timing-related flakiness in async `AsyncSelect` interactions.

## What this PR does

This PR fixes an intermittent CI failure in:

- `spec/features/category_scopes_spec.rb:63`
- `Tracked categories and templates lets a facilitator add multiple categories from different wikis at once`

Previously, the test depended on a real network-backed category search (`fr:Matériel Apple`) and could fail when the async option list did not render in time or the external request behaved inconsistently. The PR replaces that external dependency with a targeted in-test stub and adds explicit waiting for the expected option, making the test reliable.

## AI usage

I used AI assistance to:
- analyze the likely cause of the flaky failure (async UI timing + external fetch dependency),
- inspect related frontend/spec code paths,
- propose and implement a targeted stubbing strategy for the French category search,

## Screenshots

Before:
- N/A (test/spec reliability change only; no user-facing UI change)

After:
- N/A (test/spec reliability change only; no user-facing UI change)

- The current fix stubs the specific French category search in this spec to keep behavior deterministic; this is intentionally narrow to avoid affecting unrelated requests.
- If similar flakes appear in other autocomplete feature specs, we may want a shared helper for deterministic search stubs.

### DIFF
--- a/spec/features/category_scopes_spec.rb
+++ b/spec/features/category_scopes_spec.rb
@@ -6,6 +6,38 @@ describe 'Tracked categories and templates', js: true do
   let(:course) { create(:course, type: 'ArticleScopedProgram') }
   let(:user) { create(:user) }
 
+  def stub_fr_category_search_for_materiel_apple
+    page.execute_script(<<~JS)
+      (function() {
+        if (window.__stubFrCategorySearchInstalled) return;
+        window.__stubFrCategorySearchInstalled = true;
+        const originalFetch = window.fetch.bind(window);
+        window.fetch = function(input, init) {
+          const url = (typeof input === 'string') ? input : input.url;
+          const isFrCategorySearch = url.includes('fr.wikipedia.org/w/api.php') &&
+            url.includes('list=search') &&
+            url.includes('srnamespace=14') &&
+            url.includes('Mat%C3%A9riel%20Apple');
+
+          if (isFrCategorySearch) {
+            return Promise.resolve(
+              new Response(JSON.stringify({
+                query: {
+                  search: [{ title: 'Category:Matériel Apple' }]
+                }
+              }), {
+                status: 200,
+                headers: { 'Content-Type': 'application/json' }
+              })
+            );
+          }
+
+          return originalFetch(input, init);
+        };
+      })();
+    JS
+  end
+
   before do
     JoinCourse.new(course:, user:, role: CoursesUsers::Roles::INSTRUCTOR_ROLE)
     login_as user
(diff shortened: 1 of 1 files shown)


ragesoss:
Thanks Sanjana! Which AI did you use? Can you share more details about how you used it and what your process was like?

SanjanaSogimatt:
> Thanks Sanjana! Which AI did you use? Can you share more details about how you used it and what your process was like?

Yes, I used Claude. Basically I prompted it, that this particular test case is failing randomly on CI but it works on local, so what can be the possible reason that this might be happening and state me the reason why this is happening. 
Before code modification I reviewed it and prompted why this particular change is being made and how will that affect the result.
One more thing is that, I asked it to spin up more agents to see if this particular modification will break existing system.

Response -
Screenshot 2026-04-15 at 1 58 22 PM

ragesoss:
Thanks. I've found that Claude Code can sometimes find the root cause for flaky tests, but that any given explanation shouldn't be trusted until you find a way to really verify that its explanation is correct. Debugging flaky tests is the thing I've spent the most time with Claude for so far, and it's very hit-or-miss, sometimes going down the path of hacky fixes that also don't really solve the problem.

For this one, the client-side category search is not something I'd thought of before, and it seems plausible, but for a PR like this, I would ideally want a detailed explanation of how you tested to confirm that the explanation is correct (eg, by running tests in HEADED mode and stubbing the that fetch so that it fails rather than succeeds, and reading the same failure point in the spec that we usually see when it fails).

Are you using Claude Code in a terminal in the Dashboard repo locally, or using it via the web? (I've started fleshing out the config in `.claude`, so I'm interested in feedback and ideas for improving that, if you're using it via terminal.)

SanjanaSogimatt:
> Thanks. I've found that Claude Code can sometimes find the root cause for flaky tests, but that any given explanation shouldn't be trusted until you find a way to really verify that its explanation is correct. Debugging flaky tests is the thing I've spent the most time with Claude for so far, and it's very hit-or-miss, sometimes going down the path of hacky fixes that also don't really solve the problem.

Yes, I totally get it. That's why this PR is still in work in progress. That I need to make sure I am getting it right. Will post an update if I find anything

I'm actually using Cursor as my IDE, so I'm accessing Claude through that rather than the terminal tool.

PR #6799, opened 2026-04-14
Enable RuboCop rules: HashEach, YodaCondition, ZeroLength, spacing

nexpectArpit:
fixes part of #5297 

This PR enabled 6 RuboCop rules :
- `Style/HashEachMethods`
- `Style/YodaCondition`
- `Style/EmptyStringInsideInterpolation`
- `Performance/StringIdentifierArgument`
- `Layout/SpaceAroundOperators`
- `Style/ZeroLengthPredicate`

### Verification
bundle exec rspec (modified files) → 124 examples, 0 failures
@ragesoss please review it.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -235,8 +235,6 @@ Rails/CompactBlank:
   Enabled: false
 Layout/FirstHashElementIndentation:
   Enabled: false
-Performance/StringIdentifierArgument:
-  Enabled: false
 Lint/UselessConstantScoping:
   Enabled: false
 Style/RedundantParentheses:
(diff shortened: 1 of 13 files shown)


ragesoss:
I'd prefer a separate commit for each rule, so that it's easy to see which rule is related to which change (and think through whether that rule makes sense to enforce or should become a permanent exception).

nexpectArpit:
> I'd prefer a separate commit for each rule, so that it's easy to see which rule is related to which change (and think through whether that rule makes sense to enforce or should become a permanent exception).

should i make a commit with its fixes for each rule and push in this same pr  and go on for all six in same way?, or for each commit , we have separate pr?

ragesoss:
all in one PR is okay in this case, i think, since it doesn't touch too many files.

nexpectArpit:
> all in one PR is okay in this case, i think, since it doesn't touch too many files.

i done that!

ragesoss:
Great! That was so much easier to evaluate. I'm okay with all these changes.

PR #6802, opened 2026-04-16
Implementing sticky privacy notice banner on Activity tab

### LINKED ISSUE #6689: Activity Tab continues to track and display user edits after Course End Date
### What is happening?
 Activity Tab continues to display users real-time Wikipedia contributions long after a course has officially ended. While the primary course statistics (Words Added, Articles Created) correctly stop at the "End Date," the Activity feed does not. Example program

https://outreachdashboard.wmflabs.org/courses/Cin%C3%A9ma_Municipal_Le_Cosmos/r%C3%A9alisatrices_et_actrices_afro-descendantes_germanophones_avec_Noircir_Wiki_(dimanche_9_novembre_2025)/activity/recent

The Problem:
Participants expect that once a course tracking period ends, the Dashboard will stop publicizing their wiki activity. Currently, the Activity tab fetches all recent edits via the API regardless of the course's timeframe. This creates a privacy concern where organizers/public can monitor a user's post-event activity indefinitely.

### Expected behavior
 The Activity feed should strictly stop displaying new contributions 7 days after the course end date to align with the standard data-fetching buffer

nexpectArpit:
fixes #6689 
### What this PR does
This PR addresses privacy concerns on the Activity tab for courses that have ended. It implements a sticky, full-width notification banner that informs users about the continued visibility of post-course edits and guides them toward the "Course Specific" filter for a restricted view

### AI usage
used to populate the data to understand the tracking behaviour of Contributions, yet the scope of this pr touches the UI

### Screenshots
Before:
Screenshot 2026-04-17 at 2 06 18 AM

After:
Screenshot 2026-06-06 at 11 51 53 AM

### Learnings
Got to know about the how tracking of contributions work

@ragesoss please review it and @Formasitchijoh any feedback welcome.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1359,6 +1359,8 @@ en:
     see_more: See more
     show_all: Show All
     show_course_specific: Show Course Specific
+    course_ended_notification: "This course due date has been passed, but the Recent Activity tab continues to show recent edits by enrolled users. You can use the 'Show Course Specific' filter to limit the view to activity that occurred during the course."
+    filter_for_course_edits: "Filter for Course Edits Only"
     show_only_new_editors: Show Only New Editors
 
   resources:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1359,8 +1359,7 @@ en:
     see_more: See more
     show_all: Show All
     show_course_specific: Show Course Specific
-    course_ended_notification: "This course due date has been passed, but the Recent Activity tab continues to show recent edits by enrolled users. You can use the 'Show Course Specific' filter to limit the view to activity that occurred during the course."
-    filter_for_course_edits: "Filter for Course Edits Only"
+    course_ended_notification: "This shows recent edits by enrolled users, but any edits after the 'end date' are not counted toward the stats or reflected on the Articles tab. You can use the 'Show Course Specific' filter to limit the view to activity that occurred before the end date."
     show_only_new_editors: Show Only New Editors
 
   resources:
--- a/app/assets/javascripts/components/activity/activity_handler.jsx
+++ b/app/assets/javascripts/components/activity/activity_handler.jsx
@@ -5,6 +5,7 @@ import { Navigate, Route, Routes } from 'react-router-dom';
 import SubNavigation from '../common/sub_navigation.jsx';
 import CourseAlertsList from '../alerts/course_alerts_list';
 import RevisionHandler from '../revisions/revisions_handler';
+import CourseDateUtils from '../../utils/course_date_utils';
 
 const ActivityHandler = (props) => {
   const { course, current_user } = props;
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -1359,6 +1359,8 @@ en:
     see_more: See more
     show_all: Show All
     show_course_specific: Show Course Specific
+    course_ended_notification: "This course due date has been passed, but the Recent Activity tab continues to show recent edits by enrolled users. You can use the 'Show Course Specific' filter to limit the view to activity that occurred during the course."
+    filter_for_course_edits: "Filter for Course Edits Only"
```
ragesoss: Looks like this is not used?

config/locales/en.yml
```diff
@@ -1359,6 +1359,8 @@ en:
     see_more: See more
     show_all: Show All
     show_course_specific: Show Course Specific
+    course_ended_notification: "This course due date has been passed, but the Recent Activity tab continues to show recent edits by enrolled users. You can use the 'Show Course Specific' filter to limit the view to activity that occurred during the course."
```
ragesoss: I propose this text:

This shows recent edits by enrolled users, but any edits after the 'end date' are not counted toward the stats or reflected on the Articles tab. You can use the 'Show Course Specific' filter to limit the view to activity that occurred before the end date.

ragesoss (commented):
Thanks for iterating on this. Confirming up front that the banner is the intended direction: the linked issue #6689 describes suppressing post-course edits outright, but the agreed fix is this informational banner pointing to the existing "Show Course Specific" filter, not hiding the feed. So the overall shape is right.

Two change requests, beyond the two inline notes (the unused `filter_for_course_edits` key and the notification copy):

**Show the banner as soon as the course ends, not after a delay.**
`isEndedTenDaysAgo` gates the banner on a 10-day window (the issue mentioned 7). A fixed delay isn't what's wanted here: edits made *soon after* the end date are exactly the ones most likely to confuse a viewer about why they aren't counted, so the notice is most useful immediately. `CourseDateUtils.isEnded(course)` already expresses "the course has ended" — using it lets the new `isEndedTenDaysAgo` helper be removed entirely.

**Move the banner styling out of the inline `style` object into a CSS class.**
Most of the inline declarations duplicate what the project already provides:
- The element already carries `className="notification"`, and `.notification` (`app/assets/stylesheets/modules/_notifications.styl`) already sets the green background, white bold text, and padding. `backgroundColor: '#40AD90'` is the `sprout` theme variable (`_variables.styl`) hardcoded as a literal.
- The inner `.container` already centers and pads via the global `.container` rule, so the inline `margin` / `maxWidth` / `padding` there are redundant.

The genuinely banner-specific bits — `position: sticky`, the top offset, `z-index`, left alignment, and the full-bleed margins/width — belong in a Stylus module (e.g. a `.notification--course-ended` modifier under `app/assets/stylesheets/modules/`, auto-imported via the `modules/*` glob), keeping hardcoded colors and layout values out of the JSX.

**Optional:** the new date logic has a natural test home in `test/utils/course_date_utils.spec.js`; a boundary example or two (just-ended vs. still-running) would lock in the behavior.

---
*Feedback drafted in a Claude Code session (~30 min, ~6 interactions). Sage directed the review and made the design calls — the banner approach and showing the notice as soon as the course ends. The code-reuse observations were verified against the repo: `filter_for_course_edits` is unreferenced anywhere, `#40AD90` is the `sprout` variable, and `.notification` / `.container` already supply the duplicated styling. The prose here is Claude-generated; treat the suggestions as provisional and weigh them with your own judgment.*

config/locales/en.yml
nexpectArpit: Ah! , that was leftover part of my earlier draft (where i had thought that we may need separate label.

nexpectArpit:
@ragesoss PTAL

ragesoss (approved):
Re-review of the fix commit. AI-driven review feedback below isn't meant to be trusted automatically — read it, sanity-check it, and use or ignore it on your own judgment.

Both points from the prior review are addressed: the unused `filter_for_course_edits` locale key is removed, and the banner text matches the proposed wording. The move from inline `style={{…}}` to the `.notification--course-ended` class (reusing the base `.notification` styling and the `sprout` token) is a clean improvement.

One behavior change worth flagging, since it isn't mentioned in the description: the gate moved from `isEndedTenDaysAgo` to `isEnded`, so the banner now appears as soon as the course end date passes rather than 10 days later. For an informational notice that seems reasonable — just confirming it's intended.

Minor: removing `isEndedTenDaysAgo` left a double blank line above `isEnded` in `course_date_utils.js` (lines 220–221). ESLint doesn't flag it, so purely cosmetic.

The banner itself has no spec coverage. A throwaway feature spec written during this review (by Claude Code, run and falsification-checked locally — inverting the assertions made both examples fail as expected) confirmed the banner renders on an ended course and is absent on an active one. Adding a spec along those lines would lock the behavior in. Build and ESLint are green on the changed files.

*Reviewed in a Claude Code session (~20 min, a handful of interactions). Sage directed the re-review, chose to add the exploratory spec, and approved this comment before posting; the local verification — `yarn build`, ESLint, and the falsification-tested feature spec — was run by Claude Code and reviewed by Sage at a high level rather than line-by-line.*

PR #6803, opened 2026-04-17
Use course home wiki language as default locale (#6736)

### LINKED ISSUE #6736: Add default language for campaign

ArchieTyagi:
## What this PR does

Fixes #6736. Right now the dashboard always falls back to English when a user doesn't have a locale preference set. This changes it so the course's home wiki language is used as the default instead.

For example, if a French professor creates a course on fr.wikipedia.org and shares the enrollment link with students, the dashboard will now show up in French instead of English  as long as the student hasn't already set their own language preference.

## How it works

Added `preferred_locale_from_course` to the locale chain in `ApplicationController#set_locale`. It checks if the current page is a course page (by looking for `school` and `titleterm` params), finds the course, and uses its home wiki language as the lowest-priority locale preference.

Priority order stays sensible:
1. URL param (`?locale=xx`)  highest
2. User's saved preference
3. Course home wiki language  new
4. Browser Accept-Language header
5. English  lowest

## Files changed

- `app/controllers/application_controller.rb`  added `preferred_locale_from_course` method and called it in `set_locale`
- `spec/controllers/application_controller_spec.rb`  added test for the new behavior

## AI usage

Used AI for tips on the approach, understanding the locale chain logic, and debugging when I got stuck. Tested on my local setup.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -135,11 +135,20 @@ class ApplicationController < ActionController::Base
   def set_locale
     # Saved user locale takes precedence over language preferences from HTTP headers.
     preferred_locale_from_user
-    # Param takes precedence over saved user local
+    # Param takes precedence over saved user locale.
     preferred_locale_from_param
 
     preferred_locale = http_accept_language.preferred_language_from I18n.available_locales
-    I18n.locale = preferred_locale || I18n.default_locale
+    I18n.locale = preferred_locale || locale_from_course || I18n.default_locale
+  end
+
+  def locale_from_course
+    return unless params[:school] && params[:titleterm]
+    slug = "#{params[:school]}/#{params[:titleterm]}"
+    course = Course.find_by(slug: slug)
+    lang = course&.home_wiki&.language
+    return unless lang && I18n.available_locales.include?(lang.to_sym)
+    lang.to_sym
   end
 
   def preferred_locale_from_user
@@ -151,6 +160,4 @@ class ApplicationController < ActionController::Base
     return unless params[:locale]
     http_accept_language.user_preferred_languages.unshift(params[:locale])
   end
-
-
 end
--- a/spec/controllers/application_controller_spec.rb
+++ b/spec/controllers/application_controller_spec.rb
@@ -237,6 +237,13 @@ describe ApplicationController do
       get :index, params: { locale: 'not-a-real-locale' }
       expect(I18n.locale).to eq(:en)
     end
+
+   it 'sets the locale from course home wiki language' do
+      wiki = Wiki.get_or_create(language: 'fr', project: 'wikipedia')
+      create(:course, slug: 'School/Course_(Term)', home_wiki: wiki, passcode: 'abc')
+      get :index, params: { school: 'School', titleterm: 'Course_(Term)' }
+      expect(I18n.locale).to eq(:fr)
+    end
   end
 
   describe '#after_sign_in_path_for' do


ArchieTyagi:
Hi @ragesoss, would appreciate your review on this whenever you get a chance. Happy to make any changes if needed. Thanks.

ragesoss (changes requested):
Thanks for the PR! Notes below are observations from a Claude Code review session — feedback to read, weigh, and act on or ignore based on your own judgment; not authoritative conclusions.

## Implemented priority order differs from the PR description

The PR description states the chain as: `URL param → user pref → course wiki lang → Accept-Language → English`. The actual chain at `app/controllers/application_controller.rb:142` is:

```ruby
preferred_locale = http_accept_language.preferred_language_from I18n.available_locales
I18n.locale = preferred_locale || locale_from_course || I18n.default_locale
```

Because `preferred_locale` already incorporates the request's `Accept-Language` header, the real order is: `URL param → user pref → Accept-Language → course wiki lang → English`. That means an English-locale browser visiting a course on fr.wikipedia still gets English — exactly the scenario the PR description's "French professor / French students" example is intended to fix.

An exploratory spec (AI-generated, run locally during review) confirms this:

```ruby
it 'uses Accept-Language even though the course is on fr.wikipedia' do
  request.env['HTTP_ACCEPT_LANGUAGE'] = 'en-US,en;q=0.9'
  get :index, params: { school: 'School', titleterm: 'Course_(Term)' }
  expect(I18n.locale).to eq(:en)   # passes — course wiki lang is ignored
end
```

Since `I18n.available_locales` is broad (~150 entries including `:de`, `:fr`, `:en`, etc.), almost any browser's default `Accept-Language` will match something, and the course-language fallback rarely fires in practice.

Two possible directions depending on intent:

- **Course language should beat the browser** (in line with the PR description): insert into the preferred-languages stack the way `preferred_locale_from_param` does — `unshift` it onto `http_accept_language.user_preferred_languages` before the Accept-Language parsing call.
- **Browser preference remains authoritative** (current behavior): worth aligning the PR description and possibly renaming the method so future readers aren't surprised.

A contrasting-cases spec covering the Accept-Language-vs-course case would pin this down either way.

## Linked issue is about campaigns, not courses

#6736 is titled "Add default language for campaign" and points to phab T420090. This PR adds the fallback to per-course pages (matched on \`school\`/\`titleterm\`), not campaigns. Worth confirming the scope is what was intended — a campaign-level default would key off a different param/route.

## Minor

- \`Course.find_by(slug: …)\` runs inside \`set_locale\`, a \`before_action\` on every request. The downstream course controllers will re-query the same course. Acceptable since \`slug\` is indexed, but a memoization helper could avoid the duplicate lookup if it shows up in profiling.
- The new spec only covers the happy path; a contrasting-cases test would lock in the priority semantics.
- The new \`it 'sets the locale from course home wiki language'\` line is indented 3 spaces instead of 4 (RuboCop didn't flag it, but it stands out).

---

*Drafted in a Claude Code session (~20 min, ~10 interactions). Sage directed the review and approved the comment before posting. Claude Code ran RuboCop, the existing \`application_controller_spec.rb\` (passes), and an AI-generated exploratory spec (\`spec/controllers/application_controller_locale_from_course_spec.rb\`) that reproduces the Accept-Language-vs-course-language priority finding; the spec was not reviewed line-by-line by Sage before posting. Other findings were derived from code reading and not independently verified.*

ArchieTyagi:
Sorry for the late reply @ragesoss  was busy with exams. Really appreciate the detailed review.
I went back and re read issue #6736 more carefully. I think I might have been confused by the wording  the title says "campaign" but the example URL the reporter gave is https://outreachdashboard.wmflabs.org/courses/test/nlwiki_test2/home, which is a course route, and the steps they describe all use course enroll links. So I'd implemented it for those /courses/:school/:titleterm pages thinking that's what they meant.
But I want to make sure I got that right before doing more. Could you let me know  is the scope I went with (course pages) okay, or should this actually target the Campaign model?
I couldn't open the Phabricator link (T420090) so I'm not sure if there's more context there.
On the priority order  yeah, you're right, the description I wrote doesn't match what the code does. I didn't realize preferred_language_from was already pulling in Accept-Language. I want to fix the code to match the description (course lang beats browser) since that's the actual point of the issue. I'll use the unshift approach you pointed to in preferred_locale_from_param.
I'll also add the contrasting spec and fix the indent.
Will hold off on pushing anything until I hear back on the scope question, in case it needs a bigger rewrite. Thanks again!

PR #6804, opened 2026-04-17
Refactor frontend components to use article_scoped flag

lakshita10341:
## What this PR does

This PR refactors the frontend "Article Scoping" logic to rely on a backend-provided `article_scoped` flag rather than the specific `ArticleScopedProgram` course type. 
This is part of  #6501 .

Key Changes:
- **`ArticlesHandler`**: The "Categories" tracking section now displays based on the `article_scoped` flag.
- **`StatisticsUpdateModal`**: The scoped-program information message now displays based on the `article_scoped` flag.
- **`CourseCreator`**: 
    - Updated the creator to set the `article_scoped` state property automatically when an "Article Scoped Program" type is selected.
    - Replaced the hard-coded type check for the "Scoping Methods" wizard step with a check for the `article_scoped` flag.
    
## AI usage
I used AI to verify if any other logic  needed to be updated to use the `article_scoped` flag.

## Screenshots
There are no frontend UI changes; the functionality remains exactly as it was before. This has been verified through manual testing of existing article-scoped programs.


lakshita10341:
@ragesoss  The PR is ready for review. PTAL when you get time.

ragesoss (changes requested):
Thanks for splitting this out as a first step per the plan on #6501 — the scope is nicely contained. A behavior change to flag before this lands, though:

The frontend `article_scoped` field is populated by the serializer from `@course.only_scoped_articles_course?` (`app/views/courses/course.json.jbuilder:34`), and **two** subclasses override that method to always return `true`: `ArticleScopedProgram` and `VisitingScholarship` (`app/models/course_types/visiting_scholarship.rb:82`). The previous string check (`course.type === 'ArticleScopedProgram'`) matched only ASP; `course.article_scoped` matches both.

As a result, this PR unintentionally changes two UIs for VisitingScholarship courses:

- `ArticlesHandler` now renders `CategoryHandler` for VS courses, where it was hidden before. The backend `CategoriesController` authorizes only on `can_edit?`, so category adds would succeed — but `VisitingScholarship#scoped_article_titles` returns only `assigned_article_titles`, so category-tracked articles wouldn't actually be scoped in.
- `StatisticsUpdateModal` shows the "article scoped program info" message for VS courses, whose text literally reads *"This is an 'Article Scoped Program'…"* (`config/locales/en.yml:1200`) — factually wrong on a VisitingScholarship.

The PR description says "functionality remains exactly as it was before," so this looks unintended. For this PR, it would be best to preserve current behavior (no UI changes for VS). Ideally VS would support the same scoping UIs eventually, but that's a separate conversation.

Two smaller notes:

- `course_creator.jsx`: `updateCourseType` sets `article_scoped: (value === 'ArticleScopedProgram')`, which creates a short-lived inconsistency for VS — the wizard sets `false`, but after save the serializer will return `true`. Doesn't affect the wizard itself (VS never entered the scoping modal), but worth noting while the model is being refactored.
- Nothing new in this PR is covered by specs. A small feature spec like the following would guard against regressions and document intended behavior:

```ruby
# spec/features/article_scoped_flag_ui_spec.rb
describe 'article_scoped flag UI behavior', js: true do
  let(:user) { create(:user) }
  before { stub_oauth_edit; login_as user }

  context 'BasicCourse with flags[:article_scoped] = true' do
    let(:course) do
      create(:basic_course, start: '2024-01-01', end: '2024-12-31',
                            flags: { article_scoped: true })
    end
    before { JoinCourse.new(course:, user:, role: CoursesUsers::Roles::INSTRUCTOR_ROLE) }

    it 'shows the Tracked Categories section on the Articles page' do
      visit "/courses/#{course.slug}/articles"
      expect(page).to have_content 'Tracked Categories'
    end
  end

  context 'plain BasicCourse (no flag)' do
    let(:course) { create(:basic_course, start: '2024-01-01', end: '2024-12-31') }
    before { JoinCourse.new(course:, user:, role: CoursesUsers::Roles::INSTRUCTOR_ROLE) }

    it 'does not show the Tracked Categories section' do
      visit "/courses/#{course.slug}/articles"
      expect(page).not_to have_content 'Tracked Categories'
    end
  end
end
```

This spec was written during review and passes against this PR (plus a third example confirming the VS behavior change above).

*Drafted in a Claude Code session (~30 min, ~10 interactions). Sage directed the triage and confirmed the VisitingScholarship finding before Claude ran specs and drafted this comment. Sage reviewed the full comment before posting. The exploratory spec was written by Claude, run locally, and copy-edited rather than line-by-line verified.*

lakshita10341:
Thanks for the review! Sorry, I missed the VisitingScholarship case. I'll fix the components and add the tests as suggested. Will push the fixes shortly.

ragesoss (approved):
The VisitingScholarship behavior change flagged in the previous review is resolved — `article_scoped_enabled?` correctly excludes VS while covering ArticleScopedProgram and flag-bearing courses, and the serializer-only placement keeps backend scoping logic untouched. The adopted specs all pass locally, as do the Course model/controller baselines.

One small gap in the new feature spec: the VisitingScholarship modal example (`article_scoped_flag_ui_spec.rb`, "does not show the article scoped program info message in the statistics modal") never opens the statistics modal, so the absence assertion passes vacuously — it would pass even without the fix. The modal only renders after clicking the "See more" button. Opening it first makes the example meaningful:

```ruby
it 'does not show the article scoped program info message in the statistics modal' do
  click_button 'See more'
  expect(page).to have_content 'Missing or unexpected results?'  # modal is open
  expect(page).not_to have_content 'Article Scoped Program'
end
```

This variant was run locally during review and passes (along with a contrasting flagged-BasicCourse case that asserts the message *is* present in the opened modal). With that adjustment this looks ready to merge.

*Drafted in a Claude Code session (~25 min, ~5 interactions). Claude verified the VS fix by running the PR's specs plus a modal-opening exploratory spec locally; Sage reviewed the findings and this comment before posting. The suggested spec code is AI-written and was run locally.*

PR #6805, opened 2026-04-21
Improve UI for updating impact stats and allow update by any admin

### LINKED ISSUE #6719: Improve UI for updating impact stats, allow any admin to do it
Currently, the 'impact stats' that are displayed on the home page when you're logged out (in `wiki_education` mode) are controlled by setting them in `/settings`.

However:
* The current values don't get loaded, so it's confusing to figure out what to change and you can't easily just change one value while leaving the others as they are
* Only super admins can save changes, but normal admins should be able to do so as well.

Formasitchijoh:
## What this PR does
Pre-populates the impact stats form with current saved values so admins can see what they're changing, and opens the update action to regular admins (not just super admins). Closes  #6719 

## AI usage
Used Claude Code to implement the solution and improve on the for the functionality

## Screenshots
https://www.loom.com/share/5d8664285f684c5bbd1475cc003fe432


ragesoss (changes requested):
Blocker: the `type="text"` → `type="number"` change on the form inputs breaks pre-population for the production values. We store numbers above a thousand as comma-formatted strings (`"5,000"`, `"20,000"`, etc.) and HTML `<input type="number">` rejects those as invalid and renders the input empty — so opening the form shows most fields blank.

The submit path is safe when the admin leaves a blank-looking field alone (React state still holds the original string), but the display is wrong for the values this PR is specifically trying to show. Worst case, an admin sees an apparently-empty `students` field and retypes `5000`, dropping the comma on the homepage.

Options for the fix:
- Revert to `type="text"` (smallest diff, matches current stored format)
- Use `type="text"` with `inputMode="numeric"` to keep a numeric mobile keyboard without the strict validation
- Keep `type="number"` but normalize strings in/out of the form — more invasive and changes the homepage rendering

Whichever path you take, it'd be great if the committed change included a feature spec that covers the round-trip with a comma-formatted value, since the existing controller specs don't catch this. A working version from a review session:

```ruby
# spec/features/impact_stats_spec.rb
require 'rails_helper'

describe 'impact stats', type: :feature, js: true do

  before do
    Setting.set_hash('impact_stats', 'wiki_edu_courses', 100)
    Setting.set_hash('impact_stats', 'students', '5,000')
    Setting.set_hash('impact_stats', 'worked_articles', '20,000')
    Setting.set_hash('impact_stats', 'added_words', 50)
    Setting.set_hash('impact_stats', 'total_pages', '250,000')
    Setting.set_hash('impact_stats', 'volumes', 500)
    Setting.set_hash('impact_stats', 'article_views', 800)
    Setting.set_hash('impact_stats', 'universities', 500)
    Rails.cache.delete('impact_stats')
  end

  it 'updates a stat via settings and shows the new value on the home page' do
    visit '/'
    expect(page).to have_content('500 universities')
    expect(page).to have_content('5,000')

    login_as(admin, scope: :user)
    visit '/settings'
    click_button I18n.t('settings.common_settings_components.buttons.update_impact_stats')
    expect(page).to have_field('universities', with: '500')
    expect(page).to have_field('students', with: '5,000')

    fill_in 'universities', with: '550'
    click_button I18n.t('application.submit')
    expect(page).to have_content('Impact Stats Updated Successfully.')

    logout(:user)
    visit '/'
    expect(page).to have_content('550 universities')
    expect(page).to have_content('5,000')
  end
end
```

On the current branch this spec fails at `have_field('students', with: '5,000')` — Capybara finds the field but the value is the empty string. After the input-type fix it passes, and it exercises the authorization change end-to-end as a regular admin along the way.

The authorization change itself is clean — `require_admin_permissions` stays as a class-wide `before_action`, only `update_impact_stats` is removed from the super-admin list, and the parameterized `[:admin, :super_admin]` controller specs cover both new and updated actions.

A few small things worth tidying while you're in there:

- `app/controllers/settings_controller.rb:11` — accidental double space before `:update_site_notice` in the `only:` list
- `config/routes.rb:54` — `get  '/settings/fetch_impact_stats'  => ...` has two doubled spaces, inconsistent with surrounding routes
- `SettingsController#current_impact_stats` (private) duplicates `HomeController#fetch_statistics` byte-for-byte — could be extracted if you feel like it

*Drafted in a Claude Code session (~75 min, many interactions). Sage set up an isolated playground environment to run specs locally, directed the review, confirmed via a feature spec that comma-formatted values fail to pre-populate under `type="number"`, verified that production uses comma formatting, and reviewed the full comment before posting. The suggested spec was AI-generated, run locally against the PR branch (where it fails as described), and proposed for adoption.*

ragesoss (approved):
Thanks — all the prior feedback items are addressed:

- Inputs reverted to `type="text"`, which restores correct pre-population for comma-formatted values like `"5,000"`.
- Both spacing nits fixed.
- Feature spec adopted; runs green locally (including headed) and exercises the comma round-trip plus the regular-admin authorization path end-to-end.

Approving on that basis.

One small thing for next time, no need to block on it: the `it` description picked up a double space (`'shows the  new value'`) — feel free to tidy in a follow-up.

*Drafted in a Claude Code session (~20 min, ~12 interactions). Sage checked out the branch, ran `yarn build`, the controller spec, the new feature spec headless and then headed in slow mode, and the reducer spec; reviewed the full delta diff against the prior feedback; and reviewed the comment before posting.*

PR #6809, opened 2026-04-24
Skip stale Backup rows so course updates don't hang

ragesoss:
It looks like something went wrong with the backup system this week.

A possible cause was a stalled (or simply very long-running) course update, which was running for several days straight without finishing. I think this left the status as 'waiting' so long that when I finally killed the stuck job as part of a deployment, the script on the database server that was supposed to handle starting the backup when the active jobs are done had died. (That's just a guess.)

The end result was backlog of about 5 days for the course update queues. (Thanks for flagging the problem @Formasitchijoh !). I updated the last 'backup' record to mark it as a failed update, and that got the queues unstuck — except for course 35819 (https://outreachdashboard.wmflabs.org/courses/Wiki_Update/From_Continet_To_Cup) which still has not completed a successful update. I'm working on getting to the bottom of why that course is stuck, but meanwhile, this is Claude's solution to make sure the backup system doesn't cause the whole update queue to stop working.

@gabina I'm going to merge this now, but feel free propose a different solution if you don't like this.

### DIFF
--- a/app/models/system/backup.rb
+++ b/app/models/system/backup.rb
@@ -14,12 +14,20 @@
 #
 class Backup < ApplicationRecord
   IN_PROCESS = %w[waiting running].freeze
+  # A healthy backup run touches its row within minutes. Anything older is an
+  # orphaned row left behind by a crashed backup.sh (e.g., host restart), and
+  # must not indefinitely block CourseDataUpdateWorker via
+  # LogSidekiqStatus#pause_until_no_backup.
+  FRESH_WINDOW = 2.hours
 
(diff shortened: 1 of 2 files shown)


gabina:
Interesting thing. At first glance, I think I would have put the time window in the script. Something like, if you created the backup row more than 2 hours ago and you couldn't start the backup, then just set the row as 'failed'. Sidekiq should go back to normal operation after that, and we would have the failed row to know that something went wrong. But I’ll think about it more carefully when I get back to work.

PR #6812, opened 2026-04-24
Don't use LiftWingApi during course updates; batch revision-counter requests

ragesoss:
LiftWing is by far the slowest part of the course update cycle for Wikipedia courses. Since deprecating the Revisions table, it looks like we don't actually use any of the data we were getting from LiftWing, except theoretically it could indicate a deleted revision... but that's redundant from Replica and the revision counter API, so we don't need that either.

This should dramatically speed up updates on the wikis where we were using it.

This also adds a switch to batching requests to the new batch-capable version of reference-counter, which is another major speedup to course updates.

@gabina am I missing something? I think this just slipped through the cracks as we gradually replaced or removed all the things we previously did with LiftWing data during the update cycle, and just never noticed that we no longer needed it.

### DIFF
--- a/app/models/wiki_content/revision_on_memory.rb
+++ b/app/models/wiki_content/revision_on_memory.rb
@@ -15,8 +15,7 @@ class RevisionOnMemory
   attribute :characters, :integer, default: 0
 
   attr_accessor :date, :article_id, :mw_page_id, :user_id, :new_article,
-                :system, :wiki_id, :scoped, :error, :summary,
-                :wp10, :wp10_previous, :deleted
+                :system, :wiki_id, :scoped, :error, :summary, :deleted
 
   ####################
   # Instance methods #
--- a/lib/analytics/ores_diff_csv_builder.rb
+++ b/lib/analytics/ores_diff_csv_builder.rb
@@ -51,8 +51,12 @@ class OresDiffCsvBuilder
     ]
   end
 
+  # Historically the wikis where Lift Wing provided ORES quality scores. The
+  # wp10 fields are TODO/disabled now, but the wiki filter still scopes the
+  # CSV to the same set of Wikipedias.
+  ORES_SUPPORTED_WIKIPEDIAS = %w[en eu fa fr gl nl pt ru sv tr uk].freeze
   def supported_wiki_ids
-    @ids ||= Wiki.where(language: LiftWingApi::AVAILABLE_WIKIPEDIAS,
+    @ids ||= Wiki.where(language: ORES_SUPPORTED_WIKIPEDIAS,
                         project: 'wikipedia').pluck(:id)
   end
 end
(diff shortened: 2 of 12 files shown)


gabina:
hmm this is pretty interesting and hard to believe that we never noticed this before?

As far as I remember,  we created the reference-counter API to support "references added" metric for those wikis not supported by LiftWing (for example, es.wikipedia). At that point we were using a lot of information from the LiftWing API. In particular, we used the `features` to calculate "references added" for available wikis. It looks like for wikis supporting "references added" by both APIs, we prioritized the references-counter API over the LiftWing features to calculate "references count". I'm not entirely sure what happened with Wikidata. I don't think reference-counter supports wikidata (maybe it does and we just never used it, but I guess that's not the case). In case that we're using the LiftWing features to calculate references-added for wikidata, then we would miss that. However, I think that we get that data for Wikidata from another different API. In that case, I think we're safe to remove the LiftWing calls.

ragesoss:
You were right to flag this — and the answer is yes, we're safe, because the Wikidata "references added" metric already routes around Lift Wing. After commit 70566fb11 ("Skip Lift Wing for Wikidata courses"), `RevisionOnMemory#references_added` (`app/models/wiki_content/revision_on_memory.rb:42-53`) takes an early return for Wikidata via `diff_stats`:

```ruby
def references_added
  stats = diff_stats
  if stats.is_a?(Hash) && stats.key?('added_references')
    return (stats['added_references'] || 0) - (stats['removed_references'] || 0)
  end
  …
end
```

The `summary` field that `diff_stats` parses is populated by `UpdateWikidataStatsTimeslice#update_revisions_with_stats` (`app/services/update_wikidata_stats_timeslice.rb:84`) using counts from `WikidataDiffAnalyzer.analyze` (the `wikidata-diff-analyzer` gem) — its `STATS_CLASSIFICATION` line 36 explicitly maps `'added_references' => 'references added'`. The call order in `update_course_wiki_timeslices.rb:113-116` runs the diff-analyzer step before timeslice aggregation, so `references_added` always sees the diff_stats branch for in-scope Wikidata revs.

So Lift Wing's `WIKIDATA_REFERENCES` feature was already being shadowed by the diff_stats branch since 70566fb11 — even though the fallback line was still in `references_count`, it wasn't being reached for Wikidata. Removing the Lift Wing fetch path is a no-op for this metric.

One marginal note for non-Wikidata wikis: the Lift Wing fallback in `references_count` (`WIKITEXT_REF_TAGS + WIKI_SHORTENED_REF_TAGS`) was a backup that only kicked in when ref-counter's `num_ref` came back nil (errors, suppressed content, outages). Those edge cases will now contribute 0 to `references_added` instead of a Lift Wing estimate. Given ref-counter supports every non-Wikidata wiki and the batched endpoint just landed (the follow-up commit on this branch switches us over) reducing transient-error surface a lot, this seems acceptable, but worth flagging.

PR #6816, opened 2026-04-29
Skip rescoring on warm updates with no actual new revisions

ragesoss:
This addresses what I think is the main source of update slowness in current courses, like many of the `long_update` courses on Peony.

I used Claude Code to benchmark some courses, particularly this one that was clogging the system until recent optimizations (skipping requests for unused LiftWing data, and making revision-counter handle batched requests): https://outreachdashboard.wmflabs.org/courses/Wiki_Update/From_Continet_To_Cup

Even when fully up-to-date with already-processed timeslices, this course still takes a while to update. This seems to be because of the mismatch between the `revision_count` we store on the CourseWikiTimeslice record and the revision count available to CourseRevisionUpdater when checking if there are any new revisions for a timeslice. Because there are some revisions that don't (and will never) get included in the `revision_count` in some timeslices — deleted or non-tracked ones — those timeslices get reprocessed every update without changing the result.

### DIFF
--- a/app/models/course_wiki_timeslice.rb
+++ b/app/models/course_wiki_timeslice.rb
@@ -17,6 +17,7 @@
 #  needs_update         :boolean          default(FALSE)
 #  created_at           :datetime         not null
 #  updated_at           :datetime         not null
+#  mw_rev_count         :integer          default(0)
 #
 class CourseWikiTimeslice < ApplicationRecord
   belongs_to :course
@@ -72,6 +73,7 @@ class CourseWikiTimeslice < ApplicationRecord
     update_character_sum
     update_references_count
     update_revision_count
+    update_mw_rev_count
     update_stats
     update_needs_update
     save
@@ -112,6 +114,13 @@ class CourseWikiTimeslice < ApplicationRecord
     end
 
     self.revision_count = tracked_revisions.count { |rev| !rev.deleted && !rev.system }
+  end
+
+  # Count of revisions as fetched from MediaWiki, excluding only system edits.
+  # Must mirror the same filter that CourseRevisionUpdater#new_revisions? applies
+  # to the live fetched revisions, so the two counts are comparable.
+  def update_mw_rev_count
+    self.mw_rev_count = @revisions.count { |rev| !rev.system }
   end
 
   def update_stats
--- a/db/migrate/20260428223445_add_mw_rev_count_to_course_wiki_timeslices.rb
+++ b/db/migrate/20260428223445_add_mw_rev_count_to_course_wiki_timeslices.rb
@@ -1,0 +1,5 @@
+class AddMwRevCountToCourseWikiTimeslices < ActiveRecord::Migration[7.0]
+  def change
+    add_column :course_wiki_timeslices, :mw_rev_count, :integer, default: 0
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2026_03_23_212149) do
+ActiveRecord::Schema[7.0].define(version: 2026_04_28_223445) do
   create_table "admin_course_notes", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.integer "courses_id"
     t.string "title"
@@ -268,6 +268,7 @@ ActiveRecord::Schema[7.0].define(version: 2026_03_23_212149) do
     t.boolean "needs_update", default: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "mw_rev_count", default: 0
     t.index ["course_id", "wiki_id", "start", "end"], name: "course_wiki_timeslice_by_course_wiki_start_and_end", unique: true
   end
 
--- a/lib/course_revision_updater.rb
+++ b/lib/course_revision_updater.rb
@@ -87,6 +87,14 @@ class CourseRevisionUpdater
 
   # Determines if there are new revisions, based on the number of revisions and the
   # last revision datetime.
+  #
+  # The count comparison is against `timeslice.mw_rev_count`, NOT
+  # `timeslice.revision_count`. The two columns apply different filters:
+  # `revision_count` excludes deleted revs and revs in `articles_courses.not_tracked`
+  # in addition to system revs, while this method (and `mw_rev_count`) excludes only
+  # system revs. Comparing against `revision_count` would falsely report new data
+  # every cycle for any timeslice that contains a deleted or not-tracked rev.
+  # See CourseWikiTimeslice#update_mw_rev_count.
   def new_revisions?(revisions, wiki, timeslice_start)
     live_revisions = revisions.reject(&:system)
     revision_count = live_revisions.count
@@ -104,6 +112,6 @@ class CourseRevisionUpdater
     end
 
     latest_revision = revisions.maximum(:date)
-    revision_count != timeslice.revision_count || latest_revision != timeslice.last_mw_rev_datetime
+    revision_count != timeslice.mw_rev_count || latest_revision != timeslice.last_mw_rev_datetime
   end
 end
--- a/spec/lib/course_revision_updater_spec.rb
+++ b/spec/lib/course_revision_updater_spec.rb
@@ -79,6 +79,7 @@ describe CourseRevisionUpdater do
         # complete the timeslice
         last_mw_rev_datetime = '2016-03-31 19:50:54'.to_datetime
         course.course_wiki_timeslices.update(revision_count: 3,
+                                             mw_rev_count: 3,
                                              last_mw_rev_datetime:)
         VCR.use_cassette 'course_revision_updater' do
           revisions = instance_class.fetch_revisions_for_course_wiki(wiki, start_date, end_date)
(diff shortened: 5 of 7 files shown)


gabina:
Interesting. Definitely, this PR has a point. 

> I used Claude Code to benchmark some courses, particularly this one that was clogging the system until recent optimizations (skipping requests for unused LiftWing data, and making revision-counter handle batched requests): https://outreachdashboard.wmflabs.org/courses/Wiki_Update/From_Continet_To_Cup

By taking a look at the last update logs for course 35819, it seems that it definitely got fixed after the optimizations, right?  I see that excepting the first update on 2026-04-30 that took ~2 hours because it processed 6 timeslices (I guess that it processed the ones for 2026-04-29 and 2026-04-30 for each wiki), updates took less than 10 minutes.
```
            "update_logs": {
                "5": {
                    "start_time": "2026-04-29T16:06:12.630+00:00",
                    "end_time": "2026-04-29T16:09:07.424+00:00",
                    "sentry_tag_uuid": "33256a9a-c5dc-43be-8a94-9783b0598719",
                    "error_count": 3,
                    "reference_counter_403_count": 4,
                    "processed": 3,
                    "reprocessed": 0
                },
                "6": {
                    "start_time": "2026-04-29T19:27:03.649+00:00",
                    "end_time": "2026-04-29T19:31:06.534+00:00",
                    "sentry_tag_uuid": "969d8876-6712-457b-b16d-2d952d5dcbef",
                    "error_count": 3,
                    "reference_counter_403_count": 4,
                    "processed": 3,
                    "reprocessed": 0
                },
                "7": {
                    "start_time": "2026-04-29T23:13:59.058+00:00",
                    "end_time": "2026-04-29T23:18:07.971+00:00",
                    "sentry_tag_uuid": "f33684de-cbc2-48c8-88f7-44a34e70ab9f",
                    "error_count": 0,
                    "reference_counter_403_count": 4,
                    "too_many_requests_count": 0,
                    "processed": 3,
                    "reprocessed": 0
                },
                "8": {
                    "start_time": "2026-04-30T09:27:07.424+00:00",
                    "end_time": "2026-04-30T11:42:58.176+00:00",
                    "sentry_tag_uuid": "28d2fef6-3a59-42dd-88b8-cbe772d4a63e",
                    "error_count": 0,
                    "reference_counter_403_count": 1,
                    "too_many_requests_count": 0,
                    "processed": 6,
                    "reprocessed": 0
                },
                "9": {
                    "start_time": "2026-04-30T21:54:18.810+00:00",
                    "end_time": "2026-04-30T21:59:08.658+00:00",
                    "sentry_tag_uuid": "05ffb102-1396-4833-8bbe-d9a699429fef",
                    "error_count": 0,
                    "reference_counter_403_count": 0,
                    "too_many_requests_count": 0,
                    "processed": 3,
                    "reprocessed": 0
                },
                "10": {
                    "start_time": "2026-05-01T00:57:20.315+00:00",
                    "end_time": "2026-05-01T01:02:56.714+00:00",
                    "sentry_tag_uuid": "60f42098-c96d-44b4-bffc-c728255f167b",
                    "error_count": 0,
                    "reference_counter_403_count": 0,
                    "too_many_requests_count": 0,
                    "processed": 6,
                    "reprocessed": 0
                },
                "11": {
                    "start_time": "2026-05-01T04:00:41.683+00:00",
                    "end_time": "2026-05-01T04:06:46.999+00:00",
                    "sentry_tag_uuid": "84f67810-adc3-47f1-8833-2c39a1730055",
                    "error_count": 0,
                    "reference_counter_403_count": 0,
                    "too_many_requests_count": 0,
                    "processed": 3,
                    "reprocessed": 0
                },
                "12": {
                    "start_time": "2026-05-01T08:19:41.486+00:00",
                    "end_time": "2026-05-01T08:25:18.636+00:00",
                    "sentry_tag_uuid": "2785c96e-8be6-466e-bfaa-27bd74cee257",
                    "error_count": 0,
                    "reference_counter_403_count": 0,
                    "too_many_requests_count": 0,
                    "processed": 3,
                    "reprocessed": 0
                },
                "13": {
                    "start_time": "2026-05-01T15:04:04.386+00:00",
                    "end_time": "2026-05-01T15:09:54.720+00:00",
                    "sentry_tag_uuid": "e021d9d7-7ea6-48e9-9635-d3bd0b025489",
                    "error_count": 0,
                    "reference_counter_403_count": 0,
                    "too_many_requests_count": 0,
                    "processed": 3,
                    "reprocessed": 0
                },
                "14": {
                    "start_time": "2026-05-02T03:24:50.710+00:00",
                    "end_time": "2026-05-02T03:34:14.302+00:00",
                    "sentry_tag_uuid": "494cb2a4-4f49-4ceb-bd80-91c0b3a7ea9d",
                    "error_count": 0,
                    "reference_counter_403_count": 0,
                    "too_many_requests_count": 0,
                    "processed": 6,
                    "reprocessed": 0
                }
```

> Even when fully up-to-date with already-processed timeslices, this course still takes a while to update.

I guess that I'm missing those updates that took a while.

> This seems to be because of the mismatch between the revision_count we store on the CourseWikiTimeslice record and the revision count available to CourseRevisionUpdater when checking if there are any new revisions for a timeslice. Because there are some revisions that don't (and will never) get included in the revision_count in some timeslices — deleted or non-tracked ones — those timeslices get reprocessed every update without changing the result.

This is a smart point.

The `new_revisions?` method was added as attempt at optimization. It's only pertinent when `only_new` is true. That means, when we're processing a timeslice — not reprocessing. The idea behind that is: if there were no errors when processing the current timeslice, then process it again only if there are new revisions.

The point you presented here is true: the optimization isn't perfect, because we're currently comparing two numbers that maybe never will get to be the same. That affects "current" timeslices (e.i, the last timeslice processed for a given wiki), because it could mean that we're always fetching scores for it, even if there are no new revisions.

I think that these changes would improve the optimization, ensuring that we only re-fetch scores if there are new revisions (even though this new revisions could be not scoped ones, but we have no way to know that at this point).

That said, according to what I understand, this only affects courses with a huge "current" timeslice that keeps being the last timeslice with data for a long time and have no new revisions. For example, I imagine that this could happen for a huge course that actually finished one day but have end_date set to one month later. So we keep fetching scores for the current timeslice for one month without new real data. I don't see this pattern for course 35819, since it seems to have data for every single day.

I'm not against merging this at all, since I think it could make an improvement. Just wondering why you think this is the main source of update slowness in current courses. At this moment I see course 35640 has been updating for more than a day, but that seems to be because a lot of timeslices got set as `needs_update` so they're getting reprocessed.

In addition to this, I think that probably having `REVISION_THRESHOLD` set to 10000 is not realistic, and we could put it on something much smaller.

ragesoss:
>  Just wondering why you think this is the main source of update slowness in current courses.

This was just a guess, extrapolated from noticing that one example course being pretty slow even after it had caught up. It might not be the main source of slowness overall.

ragesoss:
Thanks for digging into the logs — the 35819 update times confirm the "main source of update slowness" framing was overstated, and the slowness on 35640 (lots of `needs_update` reprocessing) is a different path that this PR doesn't touch.

A code-level walkthrough with Claude Code surfaced one related thing worth flagging.

**The new comparison silently misses ArticleScopedProgram / VisitingScholarship courses.** `new_revisions?` counts non-system revisions from the full fetched array (scoped + non-scoped). `update_mw_rev_count` runs after `CourseWikiTimeslice.update_course_wiki_timeslices` applies `select(&:scoped)`, so what gets persisted is the non-system count from the scoped subset only. For courses where `scoped_article?` is always `true` (the common case), the two are equal and the gate works. For only-scoped courses with any non-scoped revisions, `mw_rev_count` will be smaller than the next cycle's live count and `new_revisions?` keeps returning `true` — i.e. the optimization never fires. So the PR is correct for normal courses, but the optimization doesn't reach the case where the upstream `scoped` filter is doing the most work. Two options: also apply `.select(&:scoped)` in `new_revisions?` so the comparison stays apples-to-apples, or call out the scope as a known limitation.

*Drafted in a Claude Code session (~1hr, ~25 interactions). The `scoped`-filter mismatch was found by tracing the call path, not reproduced via a spec — an `ArticleScopedProgram` test with a non-scoped revision would confirm or refute. Sage reviewed before posting.*

ragesoss:
Migration safety check, while we're here: the shape is the safe one (additive, nullable column, literal default, no index, no separate backfill). Both production databases are on MariaDB versions with full `ALGORITHM=INSTANT` support — 10.5.15 on the Wiki Education host, 10.11.14 on Peony — so adding a nullable integer column with a literal default at the end of the table is a metadata-only change. Should run in milliseconds regardless of `course_wiki_timeslices` size, no table rewrite, brief MDL but nothing meaningful. Safe to ship inside a normal deploy.

(One nuance worth knowing for future: MariaDB caps the number of `INSTANT`-applied column changes per table before falling back to a rewrite. Not at risk here, but worth keeping in mind for tables that accumulate many migrations.)

*Follow-up to the prior comment in the same Claude Code session, after Sage confirmed the MariaDB versions on both production hosts. Sage reviewed before posting.*

gabina:
hmm I think I'm confused here. What's the specific case that this PR fixes then?

If `new_revisions?` is taken scoped and non-scoped revisions into account, but the previous `update_revision_count` field only counts scoped-revisions, can't we just update `new_revisions?` to filter by `scoped` and avoid having the new field?

By looking at `update_revision_count`, I see that we also filter revisions from non-tracked articles. Not sure how frequent is to have non-tracked articles, but I guess that we could even apply the same filter in `new_revisions?`.

```
  def update_revision_count
    excluded_article_ids = course.articles_courses.not_tracked.pluck(:article_id)
    tracked_revisions = @revisions.reject do |revision|
      excluded_article_ids.include?(revision.article_id)
    end

    self.revision_count = tracked_revisions.count { |rev| !rev.deleted && !rev.system }
  end
```

I'm not sure about the meaning of the `rev.deleted` filter here, isn't `deleted` something that came from LiftWing API?

ragesoss:
I think you're correct, it doesn't do anything currently. I thought that we also set `rev.deleted` via reference-counter, but that's not the case. But we do get that signal from reference-counter, so I think the LiftWing removal was a regression on how we account for deleted/suppressed revisions. I think we can fix that by setting rev.deleted based on reference-counter.

gabina:
May I work on a PR for that? I think it could force me to understand the current state of things

ragesoss:
Please do! I think a next step would also be to handle scoping, so that this is useful for article-scoped courses as well, but that should wait until the `rev.deleted` issue is fixed.

gabina:
See PR #6868

Am I OK to open a PR for updating `new_revisions?` to only count scoped revisions?

ragesoss:
> Am I OK to open a PR for updating `new_revisions?` to only count scoped revisions?

Yes

PR #6863, opened 2026-05-05
Add per-editor stats for Wikidata projects

### LINKED ISSUE #6795: Include full Wikidata metrics in Editor CSV export
### Current Behavior:
The Wikidata summary on the main campaign/program page shows detailed metrics (items created, labels added, descriptions, etc.).  However, while all these data are displayed in the main page of the dashboard, they cannot be associated with the specific users: when a coordinator downloads the Editors stats CSV, they can only see the number of references added, but not the other metrics, making it difficult to track individual student contributions for specific Wikidata tasks.

### Desired Behavior:
The Editors stats CSV export should be updated to include the same granular metrics available on the programs page. Specifically, each editor row should include columns for:

- Items created
- Labels added/changed/removed
- Descriptions added/changed/removed
- Aliases added/changed/removed
- Claims added/changed/removed
- Qualifiers added
- References added (ensure this aligns with existing data)

Alternatively, a separate "Wikidata Editor Stats" CSV could be generated if the main export becomes too cluttered.

Formasitchijoh:
## What this PR does
Closes  #6795 
Adds a separate Wikidata Editor Stats CSV download showing a per-editor breakdown of Wikidata activity for each enrolled student.

Stats are built per-user during the regular course update following the same approach as the existing course-level Wikidata CSV — diff output from the analyzer is stored per user in `CourseUserWikiTimeslice` during timeslice processing, and the CSV builder sums those across all timeslices to produce one row per editor.

A data accuracy guard blocks the download if the course has revisions processed before this feature was deployed, prompting the coordinator to request a full course update first rather than silently returning incomplete totals.

## AI usage
Used Claude Code to understand the data flow through the timeslice processing pipeline. Asked questions about how the existing Wikidata CSV is built and where per-user stats needed to be stored. The implementation, decisions, and code review were done by me

The Wikidata download buttons currently appear only when Wikidata is the course's home wiki. Should they also appear for courses that track Wikidata alongside other wikis (e.g. Wikipedia + Wikidata)?


ragesoss:
This is interesting. What is your take on the performance implications and storage implications of this? Adding a new stats hash for each user timeslice is something we ought to estimate the storage requirements for, for big Wikidata courses.

Formasitchijoh:
> This is interesting. What is your take on the performance implications and storage implications of this? Adding a new stats hash for each user timeslice is something we ought to estimate the storage requirements for, for big Wikidata courses.


For the performance implication, the WikidataDiffAnalyzer API call is made once per timeslice for all revisions together  unchanged from before. The per-user stats building (build_stats_from_revisions) is purely in-memory summation over already-analyzed revision objects (the diff JSON is already sitting on each revision in memory from that single API call), so no extra API calls are added regardless of how many editors a course has so no performance overhead.

On storage: the stats column is a text field storing a YAML-serialized hash of ~61 integer counters. Rows without Wikidata activity in that timeslice stay null, so only active user-timeslice pairs get populated. For a large Wikidata course with say 100 active editors each touching 20 timeslices, that's around 2,000 populated rows × ~1.5 KB ≈ ~3 MB of additional storage per course — modest in absolute terms, though it does scale with both editor count and course duration. We might consider deciding which fields are necessary for an editor-specific stats to reduce the size of the hash and in turn the storage footprint.

ragesoss (commented):
_AI-driven review feedback like this is meant to be read and weighed, not blindly acted on — feel free to use or ignore any item based on your own judgment._

The migration looks safe to ship inside a normal deploy: `add_column :stats, :text` is nullable with no default and no backfill, so it should be metadata-only `INSTANT` on MariaDB 10.3+ even on a large table. Worth a quick `SELECT VERSION();` confirmation on the Wiki Ed and Peony hosts.

A few observations from the code-review pass that may be worth addressing before merge:

1. **Operational rollout impact of the gate.** Every existing Wikidata course will show "request a full course update" until a coordinator triggers one, and for large/long Wikidata campaigns that's a non-trivial reprocess. The PR description acknowledges the design but it may be worth coordinating a proactive sweep of active Wikidata courses at deploy time rather than waiting for someone to download a CSV and bounce off the gate.

2. **Gate query loads and deserializes every matching row.** `wikidata_editor_stats_incomplete?` does `where(...).any? { |ts| ts.stats.exclude?('total revisions') }`, which fetches every Wikidata user-timeslice with revisions and YAML-parses each one. A SQL-side `where('stats IS NULL OR stats NOT LIKE ?', '%total revisions%')` (or an `IS NULL`-based check, see #5) would avoid materializing thousands of rows per download attempt.

3. **N+1 in `CourseWikidataEditorsCsvBuilder#user_wikidata_stats`.** One `CourseUserWikiTimeslice` query per student. For courses with hundreds of students this multiplies the per-download cost; prefetching all rows once and grouping by `user_id` would scale better.

4. **57-column CSV.** `CSV_HEADERS` pulls every value from `STATS_CLASSIFICATION`, including the "Not added yet" UI groups (lemmas, forms, senses, representations, glosses, formclaims, senseclaims) that aren't shown anywhere else in the dashboard. `CourseWikidataCsvBuilder` exposes a curated 28 columns. Pruning to the same set would reduce the per-row storage footprint as well as user-visible noise, and matches the suggestion in your own comment above.

5. **`'total revisions'` as a "stats were tracked" sentinel.** Indirect, and load-bearing for the gate. The column itself is `NULL` for un-reprocessed rows, so an `IS NULL` check (read via `read_attribute_before_type_cast` if needed, since `serialize` normalizes `nil` to `{}`) would be both more explicit and SQL-pushable.

6. **No spec for the gate logic in `ReportsController#course_wikidata_editors_csv`.** A 4-example exploratory spec was written during review to confirm the behavior; sharing it below in case it's useful as a starting point. (AI-generated; verified by running locally.)

7. **Open question in the PR description (multi-wiki Wikidata visibility).** Worth resolving before merge — current condition `course.course_stats && course.home_wiki.project === 'wikidata'` hides the buttons for any course that tracks Wikidata alongside another wiki.

<details>
<summary>Exploratory gate spec (passes against this branch)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe 'Wikidata editors CSV gate', type: :request do
  let(:wikidata) { Wiki.get_or_create(language: nil, project: 'wikidata') }
  let(:user) { create(:user) }
  let(:course) do
    create(:course, home_wiki: wikidata, slug: "foo/bar-#{SecureRandom.hex(4)}")
  end

  before do
    stub_wiki_validation
    login_as user
    allow(File).to receive(:exist?).and_call_original
    allow(File).to receive(:exist?).with(/course_wikidata_editors/).and_return(false)
  end

  context 'when there are no Wikidata user timeslices yet' do
    it 'returns the CSV without prompting for an update' do
      expect(CsvCleanupWorker).to receive(:perform_at)
      get '/course_wikidata_editors_csv', params: { course: course.slug }
      expect(response.body).to include('file is being generated')
    end
  end

  context 'when an existing timeslice has revisions but no stats hash' do
    before do
      create(:course_user_wiki_timeslice,
             course:, user:, wiki: wikidata,
             revision_count: 3, stats: {})
    end

    it 'blocks the download and asks for a full course update' do
      get '/course_wikidata_editors_csv', params: { course: course.slug }
      expect(response.body).to match(/full course update/i)
    end
  end

  context 'when all timeslices with revisions have new stats' do
    before do
      create(:course_user_wiki_timeslice,
             course:, user:, wiki: wikidata,
             revision_count: 2,
             stats: { 'total revisions' => 2, 'items created' => 1 })
    end

    it 'returns the CSV' do
      expect(CsvCleanupWorker).to receive(:perform_at)
      get '/course_wikidata_editors_csv', params: { course: course.slug }
      expect(response.body).to include('file is being generated')
    end
  end

  context 'when a timeslice with zero revisions has empty stats' do
    before do
      create(:course_user_wiki_timeslice,
             course:, user:, wiki: wikidata,
             revision_count: 0, stats: {})
      create(:course_user_wiki_timeslice,
             course:, user:, wiki: wikidata,
             revision_count: 5,
             stats: { 'total revisions' => 5 })
    end

    it 'is not blocked by the zero-revision timeslice' do
      expect(CsvCleanupWorker).to receive(:perform_at)
      get '/course_wikidata_editors_csv', params: { course: course.slug }
      expect(response.body).to include('file is being generated')
    end
  end
end
```
</details>

_Drafted in a Claude Code session (~20 min, ~10 interactions). Sage directed the review, approved the findings list, and reviewed the comment before posting. Verified locally: RuboCop and ESLint clean on changed files, the PR's own spec (5 examples) passes, existing controller + model specs (28 examples) pass, and the 4-example exploratory spec above passes against this branch. The exploratory spec is AI-generated; everything else is observation from code-reading and was spot-checked but not exhaustively verified._

_(Comment written by Claude Code.)_

Formasitchijoh:
> _AI-driven review feedback like this is meant to be read and weighed, not blindly acted on — feel free to use or ignore any item based on your own judgment._
> 
> The migration looks safe to ship inside a normal deploy: `add_column :stats, :text` is nullable with no default and no backfill, so it should be metadata-only `INSTANT` on MariaDB 10.3+ even on a large table. Worth a quick `SELECT VERSION();` confirmation on the Wiki Ed and Peony hosts.
> 
> A few observations from the code-review pass that may be worth addressing before merge:
> 
> 1. **Operational rollout impact of the gate.** Every existing Wikidata course will show "request a full course update" until a coordinator triggers one, and for large/long Wikidata campaigns that's a non-trivial reprocess. The PR description acknowledges the design but it may be worth coordinating a proactive sweep of active Wikidata courses at deploy time rather than waiting for someone to download a CSV and bounce off the gate.

@ragesoss  @claude  Worth noting for deployment: any existing Wikidata course with historical revisions processed before this feature was deployed will see a "request a full course update" message if a coordinator tries to download the per-editor Wikidata CSV. The course itself is unaffected  this only impacts that specific CSV export.

I am thinking, to avoid coordinators discovering this unexpectedly, it may be worth coordinating with program organizers in advance so they know to expect it when the feature is deployed.

Issue #6861, opened 2026-05-04
Fix backup script when previous backup file doesn't exist


gabina:
### What is happening?
For Peony, last successful backup was on 2026-04-12.

There were errors for generating backups after that, likely because we have some huge courses that take a long time to update.

Regardless of that, if the previous backup failed, the script will fail next time due to:

```
ls: cannot access '/srv/dumps/2026-04-26-20:00:01//*.sql.gz': No such file or directory
05032026 20:00:01 : ERROR at line 55: command 'LAST_BACKUP=$(ls "$LATEST_DIR"/*.sql.gz | head -1)' exited with 2
```
See `/srv/dumps/2026-04-26-20:00:01/log-2026-04-26-20\:00\:01.log`

PR #6864, opened 2026-05-05
Fix backup script when previous backup has no dump file

gabina:
## What this PR does

Fixes a bug where the backup script would crash when the most recent previous backup
directory exists but contains no `.sql.gz` file (because that prior run failed before
reaching the dump step). With `set -euo pipefail` active, the failing `ls *.sql.gz` glob
exited with code 2, causing the current run to abort too — compounding the outage.

The fix replaces the two-line directory lookup with a loop that walks backup directories
from newest to oldest, skips the just-created current directory, and finds the first
directory that actually contains a `.sql.gz` file. If no previous successful backup exists,
it logs a message and skips the free-space check rather than crashing.

Addresses #6861.

## AI usage

Claude Code was used throughout: it traced the root cause, designed and implemented the
fix, wrote the commit message, and drafted this PR description. The user identified the
issue, linked issue #6861, directed the work, and performed all manual testing.

The fix required two iterations: the first attempt was caught as incorrect through manual
testing of the failure scenario (a previous backup directory with no dump file). Claude Code
explained the root cause of the regression and implemented a corrected fix. An unrelated
accidental change was also caught during review and reverted.

This was a single session across two days (May 4–5, 2026). This PR description was drafted
using the Claude Code `/prepare-pr` skill.

## Screenshots

No UI changes.

None.

### DIFF
--- a/config/backup/backup.sh
+++ b/config/backup/backup.sh
@@ -51,20 +51,30 @@ exec >>"$LOG_FILE" 2>&1
 
 log "Starting"
 
-# Calculate min free space in bytes based on last backup
-LATEST_DIR=$(ls -1d "$BACKUP_ROUTE"/*/ | sort | tail -2 | head -1)
-LAST_BACKUP=$(ls "$LATEST_DIR"/*.sql.gz | head -1)
+# Calculate min free space in bytes based on last successful backup
+LAST_BACKUP=""
+while IFS= read -r dir; do
+  candidate=$(find "$dir" -maxdepth 1 -name "*.sql.gz" | head -1)
+  if [ -n "$candidate" ]; then
+    LAST_BACKUP="$candidate"
+    break
+  fi
+done < <(ls -1d "$BACKUP_ROUTE"/*/ | sort -r | tail -n +2)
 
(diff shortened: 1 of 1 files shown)


config/backup/backup.sh
```diff
@@ -51,20 +51,30 @@ exec >>"$LOG_FILE" 2>&1
...
-  exit 1
+# Calculate min free space in bytes based on last successful backup
+LAST_BACKUP=""
+while IFS= read -r dir; do
+  candidate=$(find "$dir" -maxdepth 1 -name "*.sql.gz" | head -1)
+  if [ -n "$candidate" ]; then
+    LAST_BACKUP="$candidate"
+    break
+  fi
+done < <(ls -1d "$BACKUP_ROUTE"/*/ | sort -r | tail -n +2)
```
gabina, line 62: I took a deep look at this to understand the code, since it wasn't obvious for me what's going on here.

gabina:
After this change I also want to open a PR to follow the same approach when deleting old backups: count real *.sql.gz files instead of folders.

ragesoss:
This looks good to me.

PR #6868, opened 2026-05-08
Propagate deleted flag from ReferenceCounterApi to RevisionScoreApiHandler

gabina:
## What this PR does

Fixes how `rev.deleted` is populated for non-Wikidata wikis by surfacing
the suppressed/deleted signal from `ReferenceCounterApi` all the way through
to `RevisionScoreApiHandler`, so that `update_revision_count` correctly
excludes deleted revisions.

**Commit 1 — spec improvements (`ReferenceCounterApi`):**
Adds a real deleted `es.wiktionary` revision (`6115106`) to the stub and
spec, so the test suite exercises the `{"num_ref": null, "error": "no
content"}` path the batch API returns for suppressed/deleted revisions.
Fixes `stub_reference_counter_batch_response` to emit that shape when
`num_ref` is `nil`, matching real API behaviour.

**Commit 2 — dead-code removal:**
Removes `non_transient_error?` and the `BAD_REQUEST`/`FORBIDDEN`/`NOT_FOUND`
constants. Those codes belonged to the old per-revision GET endpoint; the
batch POST endpoint returns only 200 (per-rev errors inline) or 400 for
malformed input. Removes the corresponding 403/404 stubs and consolidates
the non-200 specs into a single context that goes through the public
interface.

**Commit 3 — propagate `deleted` flag:**
`handle_forbidden` now returns `{ 'num_ref' => nil, 'deleted' => true }`
instead of `{ 'num_ref' => nil }`. This distinguishes suppressed/deleted
revisions from revisions missing from the response, and lets the flag flow
through `RevisionScoreApiHandler#complete_score` (which already calls
`score.fetch('deleted', false)`) so `rev.deleted` is correctly set for
non-Wikidata wikis.

**Commit 4 — remove unused fields from `RevisionScoreApiHandler`:**
Removes `wp10` and `prediction` from `complete_score` — both were always
nil since LiftWing was removed from the update path, and nothing downstream
reads them. Also removes the now-dead Wikidata branch from the `features`
construction, since `complete_score` is never reached for Wikidata revisions.

## AI usage

Developed with Claude Code (Sonnet 4.6) under the direction of @gabina.
The spec cleanup, analysis of the batch endpoint (confirming 403/404 are
impossible), and the `RevisionScoreApiHandler` spec additions were drafted
by Claude Code. The core production fixes (commits 3 and 4) were written
by the human; Claude Code reviewed correctness and verified no downstream
code reads the removed fields. All commit messages were drafted by Claude
Code. Work was done across two sessions on 2026-05-08.

This PR description was drafted using the Claude Code `/prepare-pr` skill.

## Screenshots

No UI changes.

During investigation we discovered the batch API returns two distinct error
strings: `"no content"` for suppressed/texthidden revisions and
`"not_found_or_hidden"` for rev_ids that never existed. The current code
only marks `"no content"` revisions as `deleted: true`, which is correct
semantically. The `"not_found_or_hidden"` case and the defensive `nil`
guard in `transform_entry` (now confirmed unreachable) can be addressed as
a follow-up.

### DIFF
--- a/lib/reference_counter_api.rb
+++ b/lib/reference_counter_api.rb
@@ -74,13 +74,7 @@ class ReferenceCounterApi
     return normalize_batch_results(rev_ids, parsed_response) if status == 200
 
     batch_non_200_response_log(status, { rev_ids:, content: parsed_response })
-
-    default = if non_transient_error?(status)
-                { 'num_ref' => nil }
-              else
-                { 'num_ref' => nil, 'error' => parsed_response }
-              end
-    rev_ids.to_h { |id| [id.to_s, default] }
+    rev_ids.to_h { |id| [id.to_s, { 'num_ref' => nil, 'error' => parsed_response }] }
   end
 
   def normalize_batch_results(rev_ids, parsed_response)
@@ -177,16 +171,6 @@ class ReferenceCounterApi
       },
       request: { open_timeout: OPEN_TIMEOUT, timeout: REQUEST_TIMEOUT }
     )
-  end
-
-  BAD_REQUEST = 400
-  FORBIDDEN = 403
-  NOT_FOUND = 404
-  # A bad request response indicates that the language and/or project is not supported.
-  # A forbidden response likely means we lack permission to access the revision,
-  # possibly because it was deleted or hidden.
-  def non_transient_error?(status)
-    [BAD_REQUEST, FORBIDDEN, NOT_FOUND].include? status
   end
 
   TYPICAL_ERRORS = [Faraday::TimeoutError,
--- a/lib/reference_counter_api.rb
+++ b/lib/reference_counter_api.rb
@@ -84,10 +84,14 @@ class ReferenceCounterApi
     end
   end
 
-  # Per-rev entry shape from the batch endpoint:
+  # Per-rev entry shape from the batch endpoint (input):
   #   { 'num_ref' => N }                                  # normal
   #   { 'num_ref' => nil, 'error' => 'no content' }       # suppressed/deleted
   #   nil                                                 # missing from response
+  # Transformed output:
+  #   { 'num_ref' => N }                                  # normal
+  #   { 'num_ref' => nil, 'deleted' => true }             # suppressed/deleted
+  #   { 'num_ref' => nil }                                # missing from response
   def transform_entry(entry)
     return { 'num_ref' => nil } if entry.nil?
     return handle_forbidden if suppressed_content?(entry)
@@ -103,7 +107,7 @@ class ReferenceCounterApi
   # surface a total per course update, and skip Sentry reporting.
   def handle_forbidden
     @update_service&.record_reference_counter_403
-    { 'num_ref' => nil }
+    { 'num_ref' => nil, 'deleted' => true }
   end
 
   def error_key(status)
--- a/lib/revision_score_api_handler.rb
+++ b/lib/revision_score_api_handler.rb
@@ -54,25 +54,16 @@ class RevisionScoreApiHandler
   def complete_score(score)
     completed_score = {}
 
-    # If the score already has error is key is because some API request failed some way
+    # If the score already has error key is because some API request failed some way
     completed_score['error'] = score.key?('error')
 
-    # Fetch the value for 'wp10' and 'prediction', or default to nil if not present.
-    completed_score['wp10'] = score.fetch('wp10', nil)
-    completed_score['prediction'] = score.fetch('prediction', nil)
     # Fetch the value for 'deleted, or default to 'false if not present.
     completed_score['deleted'] = score.fetch('deleted', false)
 
-    # Ensure 'features' has the correct value (a hash).
-    # For Wikidata, 'features' has to contain the LiftWing features.
+    # features field has to contain the reference-counter scores if
+    # different from nil. Otherwise, it should be the empty hash.
     completed_score['features'] =
-      if @wiki.project == 'wikidata'
-        score.fetch('features', {})
-      else
-        # For other wikis, 'features' has to contain the reference-counter scores if
-        # different from nil. Otherwise, it should be the empty hash.
-        score.fetch('num_ref').nil? ? {} : { 'num_ref' => score['num_ref'] }
-      end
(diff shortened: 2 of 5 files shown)


lib/reference_counter_api.rb
```diff
@@ -90,10 +84,14 @@ def normalize_batch_results(rev_ids, parsed_response)
...
+  # Per-rev entry shape from the batch endpoint (input):
   #   { 'num_ref' => N }                                  # normal
   #   { 'num_ref' => nil, 'error' => 'no content' }       # suppressed/deleted
   #   nil                                                 # missing from response
+  # Transformed output:
+  #   { 'num_ref' => N }                                  # normal
+  #   { 'num_ref' => nil, 'deleted' => true }             # suppressed/deleted
+  #   { 'num_ref' => nil }                                # missing from response
   def transform_entry(entry)
     return { 'num_ref' => nil } if entry.nil?
```
gabina, line 96: I wonder if missing entry from response is something expected in some case or something that was added as a defensive guard. If the latter, I would consider to delete it.

ragesoss: I suspect it's a defensive guard.

lib/reference_counter_api.rb
```diff
@@ -109,7 +107,7 @@ def suppressed_content?(entry)
   # surface a total per course update, and skip Sentry reporting.
   def handle_forbidden
     @update_service&.record_reference_counter_403
-    { 'num_ref' => nil }
+    { 'num_ref' => nil, 'deleted' => true }
```
gabina, line 110: Added the 'deleted' because otherwise wasn't possible to distinguish between the missing entry (which I understand it's not the same as a deleted revision) and the actual deleted revision.

gabina:
Rubocop didn't flag this, so I don't know how to catch it locally.
```
lib/revision_score_api_handler.rb:65: trailing whitespace.
+    completed_score['features'] = 
Error: Process completed with exit code 2.
```

lib/reference_counter_api.rb
gabina: What's your opinion on removing it? I think it's a bit confusing and it looks like we get `not_found_or_hidden` error if the revision is not found.

gabina:
@ragesoss do you mind taking a look at this? I think that the `RevisionScoreApiHandler` class could be entirely deleted. But we can leave that for later. This PR only removes the parts that I felt were causing more confusion (to me) by remaining.

ragesoss (commented):
Reviewed locally on `review/6868` with `master` merged in. Linters clean; the PR's own specs and the affected `revision_score_importer_spec` pass. The fix and rationale check out; a few documentation observations and an end-to-end spec are below.

## Verified

- **Bug fix end-to-end.** An exploratory spec was written at the importer level that exercises the existing `revision_scores/get_revision_scores` cassette (which already includes suppressed rev `708326238` returning `error: "no content"`). On this branch the rev arrives at `Revision#deleted = true`. Reverting `lib/reference_counter_api.rb` and `lib/revision_score_api_handler.rb` to master and rerunning the spec produces `deleted = false`, confirming both that the bug exists on master and that this PR fixes it.
- **`wp10` / `prediction` removal is safe.** Grepping `lib/`, `app/`, and `spec/` shows no live consumer of these fields out of `RevisionScoreApiHandler`'s output. `RevisionScoreImporter#update_scores` only reads `features`, `deleted`, and `error`.
- **Wikidata branch was unreachable.** `RevisionScoreApiHandler` is only constructed by `RevisionScoreImporter`. For wikidata, `maybe_get_reference_data` short-circuits to `{}` because `ReferenceCounterApi.valid_wiki?` returns false, so `complete_score` never sees a wikidata score. Removing the branch is a no-op.
- **Spec stubs now match the real API.** The old 403 / 404 stubs were teaching specs to expect HTTP statuses the batch endpoint does not produce. The new shape (`{ 'num_ref' => nil, 'error' => 'no content' }` inside a 200) matches the recorded cassettes.

## Minor — documentation drift

Three docstrings still describe the pre-PR response shape:

- `lib/revision_score_api_handler.rb:5-8` — class header references "Lift Wing API" as part of this code path; LiftWing is no longer reached here.
- `lib/revision_score_api_handler.rb:18-36` — `get_revision_data` docstring shows responses containing `"wp10"`, `"prediction"`, and a `"start"` key, none of which are produced anymore.
- `lib/importers/revision_score_importer.rb:6-8` — class header says it "populates wp10, wp10_previous, features, features_previous and deleted fields"; only the latter three are populated now.

Worth tightening these as part of this PR or a quick follow-up.

## Minor — naming observation

`ReferenceCounterApi#handle_forbidden` (and `record_reference_counter_403` on the update service) no longer correspond to a 403 status — both names date from the old per-rev GET endpoint. The method now handles per-rev suppressed-content entries inside a 200 response. Renaming would be clearer but propagates into `UpdateService`, so probably better as a follow-up than expanding this PR.

## Confirmed follow-up

The `"not_found_or_hidden"` case (rev `753277075` in the existing cassette) currently flows through as `deleted: false`, matching what the PR description calls out. No action here.

## Exploratory spec

End-to-end importer-level spec, sharing in case it's useful as a permanent regression test in `spec/lib/importers/revision_score_importer_spec.rb`:

```ruby
it 'marks suppressed-content revisions as deleted' do
  VCR.use_cassette 'revision_scores/get_revision_scores' do
    revisions = described_class.new.get_revision_scores(array_revisions)
    suppressed = revisions.find { |r| r.mw_rev_id == 708326238 }
    expect(suppressed.deleted).to eq(true)
  end
end
```

Written by Claude Code during the review and run locally. It uses the existing cassette without modification.

---

*Drafted in a Claude Code session (~3 interactions from Sage: scope, walkthrough, post). Sage directed the review and saw the spec pass on this branch and fail on master (rev `708326238`: expected `deleted=true`, got `false`), which is direct evidence the fix works. The stale-docstring scan, the `wp10`/`prediction` consumer grep, and the wikidata-unreachability reasoning were AI-performed and spot-checked but not independently re-verified by Sage. The exploratory spec was AI-written.*

ragesoss:
This looks good to me.

gabina:
I'll open another PR to cleaning up the comments.

PR #6870, opened 2026-05-11
Exclude non-scoped revisions from new_revisions? count

gabina:
## What this PR does

`CourseRevisionUpdater#new_revisions?` compares a live count of fetched revisions
against `timeslice.mw_rev_count` to decide whether a timeslice has new data. Before
this PR, the live count included non-scoped revisions, while `mw_rev_count` was
always built from scoped-only revisions. This asymmetry caused `new_revisions?` to
return `true` spuriously whenever non-scoped MediaWiki activity appeared in the
response, triggering unnecessary update cycles for `ArticleScopedProgram` courses.

The fix has two parts:

1. `CourseRevisionUpdater#new_revisions?` now excludes non-scoped revisions from
   the live count (in addition to system revisions), so both sides of the comparison
   apply the same filter.
2. `CourseWikiTimeslice#update_cache_from_revisions` and
   `CourseUserWikiTimeslice#update_cache_from_revisions` now filter `@revisions` to
   scoped-only internally, making the scoped filter self-contained in the model
   rather than at the call site. This is a readability improvement with no behaviour
   change for `mw_rev_count`.

New specs cover the fixed behaviour: `new_data` is false when only non-scoped
revisions differ, and true when the scoped count genuinely increases.

See PR #6816 and PR #6868
## AI usage

This PR was developed with Claude Code (Anthropic). The `new_revisions?` filter
change was authored by the human; Claude assisted with planning the approach,
proposed moving the `select(&:scoped)` filter into `update_cache_from_revisions`,
and wrote the new specs in `course_revision_updater_spec.rb`. Commit messages were
drafted by Claude Code. The work involved one planning session with back-and-forth
(~20 messages), resulting in 2 commits made within a single day (2026-05-11). This
PR description was drafted using the `/prepare-pr` Claude Code skill.

## Screenshots

No UI changes.

None.

### DIFF
--- a/app/models/course_user_wiki_timeslice.rb
+++ b/app/models/course_user_wiki_timeslice.rb
@@ -50,7 +50,7 @@ class CourseUserWikiTimeslice < ApplicationRecord
                        .for_revisions_between(rev_start, rev_end)
     timeslices.each do |timeslice|
       # Group revisions that belong to the timeslice
-      revisions_in_timeslice = revisions[:revisions].select(&:scoped).select do |revision|
+      revisions_in_timeslice = revisions[:revisions].select do |revision|
         timeslice.start <= revision.date && revision.date < timeslice.end
       end
       # Get or create article course timeslice based on course, article_id,
@@ -71,7 +71,8 @@ class CourseUserWikiTimeslice < ApplicationRecord
 
   # Assumes that the revisions are for their own course user wiki
   def update_cache_from_revisions(revisions)
-    @revisions = revisions
+    # Only work with scoped revisions
+    @revisions = revisions.select(&:scoped)
     @liverevisions = live_revisions
     tracked_namespace_revisions = live_revisions_in_tracked_namespaces
     update_character_sum(@liverevisions, tracked_namespace_revisions)
--- a/app/models/course_wiki_timeslice.rb
+++ b/app/models/course_wiki_timeslice.rb
@@ -53,7 +53,7 @@ class CourseWikiTimeslice < ApplicationRecord
                                                 .for_revisions_between(rev_start, rev_end)
     course_wiki_timeslices.each do |timeslice|
       # Group revisions that belong to the timeslice
-      revisions_in_timeslice = revisions[:revisions].select(&:scoped).select do |revision|
+      revisions_in_timeslice = revisions[:revisions].select do |revision|
         timeslice.start <= revision.date && revision.date < timeslice.end
       end
       # Update cache for CourseWikiTimeslice
@@ -67,7 +67,8 @@ class CourseWikiTimeslice < ApplicationRecord
 
   # Assumes that the revisions are for their own course wiki
   def update_cache_from_revisions(revisions)
-    @revisions = revisions
+    # Only work with scoped revisions
+    @revisions = revisions.select(&:scoped)
     @students = course.courses_users.where(role: CoursesUsers::Roles::STUDENT_ROLE)
 
     update_character_sum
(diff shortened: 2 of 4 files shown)


ragesoss (commented):
The fix and refactor look solid. Two notes flagged during review, neither blocking:

**1. Stale comment in `app/models/course_wiki_timeslice.rb:120`**

```ruby
# Count of revisions as fetched from MediaWiki, excluding only system edits.
# Must mirror the same filter that CourseRevisionUpdater#new_revisions? applies
# to the live fetched revisions, so the two counts are comparable.
def update_mw_rev_count
  self.mw_rev_count = @revisions.count { |rev| !rev.system }
end
```

Post-PR, `@revisions` is scoped-filtered inside `update_cache_from_revisions`, so the resulting `mw_rev_count` effectively excludes both system *and* non-scoped revisions. The "excluding only system edits" wording reads as if scoped revisions were still in the mix. Worth tightening to match the new code shape.

**2. Asymmetric refactor across the three timeslice models**

The PR moves `select(&:scoped)` into `update_cache_from_revisions` for `CourseWikiTimeslice` and `CourseUserWikiTimeslice`. `ArticleCourseTimeslice#update_cache_from_revisions` (`app/models/article_course_timeslice.rb:89`) is the third sibling with the same shape; neither its call site nor its method body filters `scoped`. For `ArticleScopedProgram` courses, that means non-scoped revisions can contribute to per-article `revision_count`/`character_sum`. This is pre-existing, not introduced here — but worth either (a) calling out in the commit/PR description that the third sibling was reviewed and intentionally left alone, or (b) opening a follow-up issue to investigate whether the same fix should apply there.

**Verified locally**

- Diff walkthrough confirms the refactor is behaviour-preserving for `mw_rev_count`: all consumers of `@revisions` inside `update_cache_from_revisions` on both modified models receive the same scoped-filtered set whether the filter was applied at the call site or inside the method.
- `bundle exec rspec spec/lib/course_revision_updater_spec.rb spec/models/course_wiki_timeslice_spec.rb spec/models/course_user_wiki_timeslice_spec.rb spec/models/article_course_timeslice_spec.rb` — all green.
- `bundle exec rubocop` on the four changed files — clean.

*Drafted in a Claude Code session (~15 min, ~10 interactions). Sage reviewed the full comment before posting. The behaviour-preservation argument was verified by reading the four caller methods, and the spec/lint runs above were executed locally; the `ArticleCourseTimeslice` semantics question was identified but not deeply investigated — treat it as a pointer for the author rather than a confirmed bug. (Comment written by Claude Code.)*

gabina:
hmm I'll update the comment in `update_mw_rev_count` because I agree that it's confusing

> The PR moves select(&:scoped) into update_cache_from_revisions for CourseWikiTimeslice and CourseUserWikiTimeslice. ArticleCourseTimeslice#update_cache_from_revisions (app/models/article_course_timeslice.rb:89) is the third sibling with the same shape; neither its call site nor its method body filters scoped. For ArticleScopedProgram courses, that means non-scoped revisions can contribute to per-article revision_count/character_sum. This is pre-existing, not introduced here — but worth either (a) calling out in the commit/PR description that the third sibling was reviewed and intentionally left alone, or (b) opening a follow-up issue to investigate whether the same fix should apply there.

This was always intentional. It was a strategy to deal with article scoped courses. Mainly with changes in categories. But we may want to review this at some point to try to find a better way because it's not working efficiently for those big ASC anyway.

gabina:
Ready for review now.

PR #6874, opened 2026-05-12
Fix wikidata stats wiped when reprocessing timeslice with no new revisions

### LINKED ISSUE #6860: Fetch wikidata stats when reprocessing timeslices even if no new revisions
### What is happening?
We only fetch wikidata stats for a timeslice if it has new revisions. That's  OK for most cases. However, if a timeslice is being reprocessed (i.e, `only_new` is false) and it has no new revisions compared to the current timeslice state, the wikidata stats are not fetched but timeslice is still reprocessed, so the Wikidata stats end up empty.

### To Reproduce

> 1. Copy course https://dashboard.wikiedu.org/courses/Wiki_Education/NYDCLC_Wikidata_Institute_(Fall_2021)/home
> 2. Run manual update
> 3. See wikidata stats
> 4. Set one non-empty timeslice as needs_update
> 5. Run manual update
> 6. See how wikidata stats changed

### Expected behavior
Wikidata stats should be fetched using the same criterion as timeslice processing:: `!only_new || new_data?(wiki)` .

At less I'm missing something, the fix seems pretty straightforward. However, it could be a nice opportunity to refactor those behaviors a bit to make it clearer.

### Screenshots
**After first course update there are wikidata stats.**

Image

**After setting timeslice as `needs_update` and updating course, wikidata stats are empty**

Image

### Additional context
I don't know if this bug is currently affecting courses, but we should definitely fix it.

I found this bug when reviewing PR #6816

gabina:
## What this PR does

Fixes a bug where Wikidata stats were wiped when a timeslice was reprocessed
with no new revisions. Closes #6860.

When a `CourseWikiTimeslice` with `needs_update: true` was reprocessed
(`only_new: false`) and no new revisions had been added since the last
processing, `process_timeslices` would run (clearing the stored stats) but
`maybe_fetch_wikidata_stats` would skip — because it only checked
`new_data?(wiki)` rather than the same condition used by `process_timeslices`:
`!only_new || new_data?(wiki)`. The result was a timeslice reprocessed with
empty Wikidata stats.

The fix aligns the guard condition in both places. A failing spec was written
first to demonstrate the bug, then the fix was applied.

The PR also includes a follow-up readability refactor of
`UpdateCourseWikiTimeslices#handle_timeslice`: its non-split branch is
extracted into `process_leaf_timeslice`, turning `handle_timeslice` into a
clean recursive dispatcher. The shared guard condition is also extracted into a
named predicate `should_update_timeslice?`, making it clear that the Wikidata
stats fetch and the timeslice cache update share the same criterion.

## AI usage

This PR was developed entirely with Claude Code (Anthropic) under the author's
direction. Claude Code read the relevant service and spec files, identified the
root cause, wrote the failing spec, proposed and applied the fix, and proposed
the refactor. The author directed each step, chose between design options (e.g.
which extraction approach to use for the refactor), and reviewed all changes
before committing. Both commit messages and this PR description were drafted by
Claude Code using the `/commit` and `/prepare-pr` skills.

Both commits were made within a single ~20-minute session. All work was driven
by the author; Claude Code produced no code autonomously.

## Screenshots

To reproduce, use the course https://dashboard.wikiedu.org/courses/Wiki_Education/NYDCLC_Wikidata_Institute_(Fall_2021)/home,
run a manual update, set one non-empty timeslice as `needs_update`, and run
the update again.

**Before (bug): Wikidata stats after setting a timeslice as `needs_update` and re-running the update — stats are empty.**

image

**After (fix): Wikidata stats are preserved after reprocessing.**

image

None.

### DIFF
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -112,7 +112,7 @@ class UpdateCourseWikiTimeslices
     else
       log_processing(wiki, start_date, end_date, only_new)
       add_scores(only_new:)
-      maybe_fetch_wikidata_stats(wiki)
+      maybe_fetch_wikidata_stats(wiki, only_new:)
       process_timeslices(wiki) if !only_new | new_data?(wiki)
       [start_date]
     end
@@ -143,8 +143,8 @@ class UpdateCourseWikiTimeslices
     [timeslice_end - 1.second, @course.end].min
   end
 
-  def maybe_fetch_wikidata_stats(wiki)
-    fetch_wikidata_stats(wiki) if wiki.project == 'wikidata' && new_data?(wiki)
+  def maybe_fetch_wikidata_stats(wiki, only_new:)
+    fetch_wikidata_stats(wiki) if wiki.project == 'wikidata' && (!only_new || new_data?(wiki))
   end
 
   def new_data?(wiki)
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -104,18 +104,22 @@ class UpdateCourseWikiTimeslices
     fetch_only_revisions(wiki, start_date, end_date)
     split, dates = @splitter.maybe_split(wiki, start_date, end_date, @revisions)
     if split
-      new_start = dates[0]
-      midpoint = dates[1]
-      new_end = dates[2]
-      handle_timeslice(wiki, new_start, midpoint,
-                       only_new:) + handle_timeslice(wiki, midpoint, new_end, only_new:)
+      new_start, midpoint, new_end = dates
+      handle_timeslice(wiki, new_start, midpoint, only_new:) +
+        handle_timeslice(wiki, midpoint, new_end, only_new:)
     else
-      log_processing(wiki, start_date, end_date, only_new)
-      add_scores(only_new:)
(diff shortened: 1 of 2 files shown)


ragesoss (approved):
The bug is well-characterized in the description, and the spec captures it cleanly. Running it against master's service code reproduces the failure (`update_revisions_with_stats` not called); running it against the PR passes. The refactor reads as a sensible follow-up — `process_leaf_timeslice` makes the recursive structure obvious, and `should_update_timeslice?` names what was previously a duplicated inline condition. Grouping the wikidata stats fetch and `process_timeslices` under a single `if` is also a nice clarity win, since the original `maybe_*` indirection was obscuring that the two share the same criterion.

Incidental observation: the refactor also normalizes `if !only_new | new_data?(wiki)` (bitwise `|`) to `||`. Same effect for boolean operands but more conventional.

*Reviewed in a Claude Code session (~10 min, ~10 interactions). Sage directed the review and read each finding; Claude Code verified the new spec independently by checking out master's `update_course_wiki_timeslices.rb` and confirming the spec fails there, then re-running against the PR to confirm it passes. RuboCop was run on both changed files (clean). The full spec file was not run end-to-end (the heavy VCR-backed examples would need cassette work that's out of scope); CI green was relied on for the rest.*

PR #6875, opened 2026-05-14
Simplify methods to update timeslices to handle a single timeslice

gabina:
## What this PR does

Simplifies the three timeslice update methods —
`CourseWikiTimeslice.update_course_wiki_timeslices`,
`CourseUserWikiTimeslice.update_course_user_wiki_timeslices`, and
`ArticleCourseTimeslice.update_article_course_timeslices` — by removing an
unnecessary loop-and-filter pattern. All three are always called with a period
that spans exactly one timeslice (the service layer sets `real_end =
timeslice_end - 1.second` to guarantee this), so the loop was dead
complexity. Each method now calls `.first` directly on the matched timeslices,
and the method comments document the single-timeslice contract.

A Sentry invariant guard was added to each method to log an error if more than
one timeslice is ever matched, making the assumption explicit and alerting us
if it is ever violated in production.

A separate commit also clarifies the spec coverage for `mw_rev_count` vs
`revision_count` in `course_wiki_timeslice_spec.rb`, introducing a non-scoped
revision to the fixture to make the difference between the two counters
explicit.

## AI usage

Claude Code (claude-sonnet-4-6) was used throughout this PR. The overall
approach and goal were defined by the human; Claude drafted the model changes
and Sentry guard patterns, iterated on RuboCop violations (extracting
`log_nil_article_id` into a private class method to avoid
`rubocop:disable`), and wrote all commit messages. The spec changes were
written by the human, with Claude adding the Sentry specs. Claude Code also
drafted this PR description using the `/prepare-pr` skill.

This was developed across 4 commits over roughly 25 hours (2026-05-14 to
2026-05-15), across two conversation sessions (the first ran out of context).
The human sent approximately 30–40 short messages across both sessions,
directing each step and confirming specs passed before each commit.

## Screenshots

No UI changes.

None.

### DIFF
--- a/app/models/article_course_timeslice.rb
+++ b/app/models/article_course_timeslice.rb
@@ -46,40 +46,40 @@ class ArticleCourseTimeslice < ApplicationRecord
   end
 
   # Given a course, an article, and a hash of revisions like the following:
-  # {:start=>"20160320", :end=>"20160401", :revisions=>[...]},
-  # updates the article course timeslices based on the revisions.
-  def self.update_article_course_timeslices(course, article_id, revisions) # rubocop:disable Metrics/AbcSize,Metrics/MethodLength
-    rev_start = revisions[:start]
-    rev_end = revisions[:end]
+  # {:start=>"20160320000000", :end=>"20160320235959", :revisions=>[...]},
+  # where start and end span a single timeslice period (end is 1 second
+  # before the next timeslice boundary), updates the article course timeslice.
+  def self.update_article_course_timeslices(course, article_id, revisions)
     wiki_id = revisions[:revisions].first.wiki_id
-    # For debugging purposes only. TODO: delete this sentry message
-    if article_id.nil?
-      Sentry.capture_message "Article id nil for course #{course.id}",
-                             level: 'error',
-                             extra: {
-                               wiki_id:,
-                               mw_page_ids: revisions[:revisions].map(&:mw_page_id),
-                               revision_ids: revisions[:revisions].map(&:mw_rev_id)
-                             }
-    end
-    # Timeslice dates to update are determined based on course wiki timeslices
+
+    # For debugging purposes only. TODO: delete this sentry message log
+    log_nil_article_id(course, article_id, wiki_id, revisions) if article_id.nil?
+
     timeslices = course.course_wiki_timeslices.where(wiki_id:)
-                       .for_revisions_between(rev_start, rev_end)
-    timeslices.each do |timeslice|
-      # Group revisions that belong to the timeslice
-      revisions_in_timeslice = revisions[:revisions].select do |revision|
-        timeslice.start <= revision.date && revision.date < timeslice.end
-      end
-      # Get or create article course timeslice based on course, article_id,
-      # timeslice.start and timeslice.end
-      ac_timeslice = ArticleCourseTimeslice.find_or_create_by(course:,
-                                                              article_id:,
-                                                              start: timeslice.start,
-                                                              end: timeslice.end)
-      # Update cache for ArticleCourseTimeslice
-      ac_timeslice.update_cache_from_revisions revisions_in_timeslice
+                       .for_revisions_between(revisions[:start], revisions[:end])
+    if timeslices.size > 1
+      message = "Multiple article course timeslices matched for course #{course.slug}"
+      Sentry.capture_message message,
+                             level: 'error',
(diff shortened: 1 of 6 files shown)


ragesoss (approved):
A note on how to use AI-driven review feedback: the observations below come from a Claude Code review session and shouldn't be automatically trusted or blindly acted on. Read them, use the judgment, ignore what isn't useful.

The single-timeslice contract was verified by tracing the call chain through `UpdateCourseWikiTimeslices#fetch_only_revisions` → `CourseRevisionUpdater#format_revision_response` → the three `update_*_timeslices` class methods. `real_end = [timeslice_end - 1.second, @course.end].min` combined with `for_revisions_between`'s `in_period.or(for_datetime(start)).or(for_datetime(end))` does match exactly the single timeslice row at every leaf of the split recursion. RuboCop and the three model specs are clean (18/18), and `update_course_wiki_timeslices_spec.rb` still passes (17/17).

A few observations to consider — none are blockers since the contract appears to hold:

**1. Behavior in the guard-fires case.** When `timeslices.size > 1` triggers the new Sentry alert, the code still proceeds with `timeslices.first.update_cache_from_revisions revisions[:revisions]`. The previous implementation distributed revisions across all matched timeslices by per-row date filtering. So if the invariant ever broke in production, the new code would over-count revisions in one row (whichever sorts first — no `ORDER BY` is applied) and silently leave the others un-updated. The Sentry alert means we'd notice quickly, but worth being conscious that the "loud failure" also has a data-correctness footprint, not just an inert assertion. One option, if a future incident shows this matters: fall back to the old per-timeslice filtering when `timeslices.size > 1`, so the alert fires *and* the data lands correctly.

**2. `timeslices.first` on an unscoped relation.** Corollary of #1: with no explicit ordering, which row sorts first in a violation is implementation-defined. Adding `.order(:start)` would make the "first" deterministic if #1 ever matters.

**3. `nil`-safety on `timeslices.first`.** In principle `timeslices.first` could be `nil` (e.g., concurrent timeslice deletion between split and process), causing a `NoMethodError`. The caller's `rescue StandardError` in `update_course_wiki_timeslices.rb` would catch it and re-mark the parent for update, so it's recoverable — but it's a place where the new code raises where the old loop silently no-op'd. Probably fine as-is.

The `log_nil_article_id` extraction to avoid the `rubocop:disable` pair is a clear win. The `mw_rev_count` vs `revision_count` spec clarification reads much better with the explicit `scoped: false` revision in the fixture.

*Drafted in a Claude Code session (~10 min, ~10 interactions). Sage directed the review, but the contract verification and finding analysis were primarily Claude's — Sage spot-read the findings and approved posting without independently re-tracing the call chain or re-running specs locally. No exploratory specs were written; the analysis rests on reading the existing call sites and running the PR's own specs.*

(Review written by Claude Code.)

gabina:
> 1. Behavior in the guard-fires case. When timeslices.size > 1 triggers the new Sentry alert, the code still proceeds with timeslices.first.update_cache_from_revisions revisions[:revisions]. The previous implementation distributed revisions across all matched timeslices by per-row date filtering. So if the invariant ever broke in production, the new code would over-count revisions in one row (whichever sorts first — no ORDER BY is applied) and silently leave the others un-updated. The Sentry alert means we'd notice quickly, but worth being conscious that the "loud failure" also has a data-correctness footprint, not just an inert assertion. One option, if a future incident shows this matters: fall back to the old per-timeslice filtering when timeslices.size > 1, so the alert fires and the data lands correctly.

I think that eventually we could stop using `for_revisions_between` and use a method that matches `start` and `end` more precisely, but didn't want to make both changes at the same time.

> 2. timeslices.first on an unscoped relation. Corollary of 1: with no explicit ordering, which row sorts first in a violation is implementation-defined. Adding .order(:start) would make the "first" deterministic if 1 ever matters.

That's true. But the Sentry alert exists to notice immediately if something is not working as expected, and in that case we should debug the course and see what was the problem.

> 3. nil-safety on timeslices.first. In principle timeslices.first could be nil (e.g., concurrent timeslice deletion between split and process), causing a NoMethodError. The caller's rescue StandardError in update_course_wiki_timeslices.rb would catch it and re-mark the parent for update, so it's recoverable — but it's a place where the new code raises where the old loop silently no-op'd. Probably fine as-is.

I'm OK with logging `NoMethodError` if there are no timeslices for the dates (which of course is a drastic error).

I think we're ready to go.

PR #6876, opened 2026-05-15
Escalate courses with consecutive unfinished updates to long_update queue

gabina:
## What this PR does

`CourseQueueSorting#queue_for` previously determined a course's Sidekiq queue
solely from the timing of its most recent *completed* updates (`update_logs`).
A course stuck on a heavy timeslice would keep accumulating entries in
`unfinished_update_logs` while remaining in `short_update` based on its past
fast completions — tying up short-queue workers and starving other courses.

This was observed in production with course 36303 [Campaña mujeres en tiempos
de guerra 2026](https://outreachdashboard.wmflabs.org/courses/Wikiesfera/Campa%C3%B1a_mujeres_en_tiempos_de_guerra_2026/home) (an article-scoped program):
three consecutive updates started but never finished. While the course seems to be
a small course, it has now 79,272 ACTs, which means at least 79,272 revisions.
The short queue was
restarted once, and it died again apparently because the same course was
re-enqueued into the same short queue.

This PR adds a heuristic: if a course has 2 or more consecutive unfinished
updates since its last successful `end_time`, it is escalated to `long_update`.
Once an update succeeds the counter resets and normal queue-assignment logic
(based on completed-update timings) resumes.

Old `unfinished_update_logs` entries that predate the last successful `end_time`
are excluded, so stale entries from prior incidents don't cause false positives.

## AI usage

Claude Code (claude-sonnet-4-6) was used to implement this change under the
author's direction. The author diagnosed the production incident and proposed the
heuristic; Claude Code translated it into code, wrote the spec, and drafted the
commit message. Iterations included: extracting the threshold into a named
constant (`UNFINISHED_UPDATES_LONG_QUEUE_THRESHOLD`) and refactoring an
overly-long guard line into a `too_many_consecutive_unfinished_updates?`
predicate. The author ran the specs and manually verified the behavior before
committing.

This was a single focused session resulting in one commit. This PR description
was drafted using the `/prepare-pr` Claude Code skill.

## Screenshots

No UI changes.

### DIFF
--- a/lib/data_cycle/course_queue_sorting.rb
+++ b/lib/data_cycle/course_queue_sorting.rb
@@ -1,9 +1,13 @@
 # frozen_string_literal: true
 
 module CourseQueueSorting
+  # Number of consecutive unfinished updates that triggers escalation to long_update queue
+  UNFINISHED_UPDATES_LONG_QUEUE_THRESHOLD = 2
+
   def queue_for(course)
     update_longest_update_time(course)
     return 'very_long_update' if course.very_long_update?
+    return 'long_update' if too_many_consecutive_unfinished_updates?(course)
 
     case longest_recent_update_time(course)
     when nil
@@ -15,6 +19,22 @@ module CourseQueueSorting
     when (601..) # more than 10 minutes
       'long_update'
     end
+  end
+
+  def too_many_consecutive_unfinished_updates?(course)
+    consecutive_unfinished_updates(course) >= UNFINISHED_UPDATES_LONG_QUEUE_THRESHOLD
+  end
+
+  # Returns the number of consecutive update failures since the last success.
+  # Old unfinished entries that predate the last successful end_time are excluded,
+  # since they're not evidence of a current problem.
+  def consecutive_unfinished_updates(course)
+    unfinished_logs = course.flags['unfinished_update_logs']
+    return 0 unless unfinished_logs.present?
+
+    finished_logs = course.flags['update_logs']
+    latest_end = finished_logs&.values&.filter_map { |log| log['end_time'].to_f }&.max || 0
+    unfinished_logs.values.count { |log| log['start_time'].to_f > latest_end }
   end
 
   def longest_recent_update_time(course)
--- a/spec/lib/data_cycle/course_queue_sorting_spec.rb
+++ b/spec/lib/data_cycle/course_queue_sorting_spec.rb
@@ -1,0 +1,109 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require "#{Rails.root}/lib/data_cycle/course_queue_sorting"
+
+describe CourseQueueSorting do
+  let(:including_class) { Class.new { include CourseQueueSorting } }
+  let(:subject) { including_class.new }
+
+  let(:fast_finished_log) do
+    { 1 => { 'start_time' => 10.seconds.ago, 'end_time' => 5.seconds.ago } }
+  end
+
+  describe '#queue_for' do
+    context 'when there are 2 or more consecutive unfinished updates' do
+      let(:course) do
+        create(:course, start: 1.day.ago, end: 2.months.from_now,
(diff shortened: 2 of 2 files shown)


gabina:
@ragesoss it looks like the short queue didn't actually die the second time, even though it wasn't visible on the Sidekiq dashboard for a while. However, course 36303 still doesn't process successfully: the last finished update was on 2026-05-14T13:18:56. After that, all attempts failed. I think merging this would work to automatically fix these cases.

```
        "unfinished_update_logs": {
                "1": {
                    "start_time": "2026-05-14T14:23:15.444+00:00"
                },
                "2": {
                    "start_time": "2026-05-14T21:10:56.770+00:00"
                },
                "3": {
                    "start_time": "2026-05-15T00:45:11.820+00:00"
                },
                "4": {
                    "start_time": "2026-05-15T06:37:53.868+00:00"
                }
            },
            "update_logs": {
                "34": {
                    "start_time": "2026-05-14T13:18:52.288+00:00",
                    "end_time": "2026-05-14T13:18:56.581+00:00",
                    "sentry_tag_uuid": "a8c0ede4-075a-4d5c-be93-26ee3ab3118b",
                    "error_count": 0,
                    "reference_counter_403_count": 0,
                    "too_many_requests_count": 0,
                    "processed": 14,
                    "reprocessed": 0
                }
            },
```

ragesoss (approved):
Approving — the heuristic is well-motivated, the code is straightforward, and the existing spec covers the core branches. The notes below are FYIs / possible follow-ups, not blockers.

(Heads-up on AI-assisted review: the following observations came out of a Claude Code review session. They're worth reading and using or ignoring on your own judgment, not treating as authoritative.)

**1. Optional spec addition — mixed stale + fresh entries.** The PR's spec covers all-stale and all-fresh `unfinished_update_logs`, but not a mix. A production-shaped course is likely to accumulate both over time (since `unfinished_update_logs` is only pruned when a matching `start_time` finishes). A case like:

```ruby
'update_logs' => {
  1 => { 'start_time' => 30.minutes.ago, 'end_time' => 25.minutes.ago }
},
'unfinished_update_logs' => {
  1 => { 'start_time' => 2.hours.ago },    # predates last success
  2 => { 'start_time' => 5.minutes.ago }   # current
}
```

…should count as 1 (not escalate). An exploratory spec written during review confirmed the current code handles this correctly, so this is really just locking the behavior in against future regressions.

**2. Cosmetic — `filter_map` does no filtering here.** In `consecutive_unfinished_updates`:

```ruby
latest_end = finished_logs&.values&.filter_map { |log| log['end_time'].to_f }&.max || 0
```

`to_f` always returns a Float (`nil.to_f == 0.0`), so `filter_map` is equivalent to `map`. Not a bug — `nil` end_times would produce `0.0` which `max` discards in favor of any real value — but the `filter_` prefix suggests defensive intent that isn't actually doing anything. Plain `map` reads more honestly.

**3. FYI on behavior — no further escalation.** A course that crashes twice on `short_update` is escalated to `long_update`; if it then crashes twice on `long_update`, it stays there (`very_long_update` only triggers via the existing `course.very_long_update?` flag, which is set elsewhere). For the production case (course 36303), that's exactly what's wanted — long_update workers have the right resource budget. But worth keeping in mind if a similar incident recurs on a course that's *also* too heavy for the long queue: this heuristic alone won't surface it. Possible future tweak: bump to `very_long_update` after N crashes on `long_update`. Not for this PR.

---

*Drafted in a Claude Code session (~45 minutes, ~15 interactions). Sage directed each phase, killed a runaway baseline spec mid-review, and read the full draft before posting. Claude Code wrote a temporary exploratory spec (5 cases) that confirmed item 1's behavior and that the production-realistic `DateTime` + YAML roundtrip shape works as intended; Sage didn't independently re-run those exploratory cases. Other findings are observations from code-reading and have not been independently verified at runtime.*

Issue #666, opened 2016-02-10
Student actions disappear in narrow viewport


adamwight:
See attachments:
![students_narrow](https://cloud.githubusercontent.com/assets/819437/12962148/07ebd41e-cff9-11e5-969d-ff684311a74a.png)
![students_wide](https://cloud.githubusercontent.com/assets/819437/12962149/07ef213c-cff9-11e5-8c75-3e8f9ed1b770.png)

PR #6880, opened 2026-05-19
Add advanced search and server-side sorting to campaign articles page

joserafael:
## What this PR does

This pull request introduces the Advanced Search functionality for the Campaign Articles view, matching the capabilities previously available on the programs view. It allows users to filter campaign articles using various metrics and parameters, significantly improving the discoverability of content within large campaigns.

## Key Changes

*   **Advanced Search Form:** Added an advanced search UI to `app/views/campaigns/articles.html.haml` featuring a toggleable form.
*   **New Filters:** Users can now filter articles by:
    *   Article Title
    *   Institution/School (Multi-select using TomSelect)
    *   Characters Added (Min/Max range)
    *   References Count (Min/Max range)
    *   Views (Min/Max range)
*   **Backend Query Logic:** Updated `RankedArticlesCoursesQuery`, `CampaignsController`, and `CoursesPresenter` to process and apply the new filter parameters effectively.
*   **JavaScript Refactoring:** Extracted inline scripts from the articles view into a dedicated, jQuery-free Webpack bundle (`campaigns_articles.js`). The view now cleanly includes this via `hot_javascript_tag 'campaigns_articles'`, bringing its architecture in line with `programs.html.haml`.
*   **Styling:** Added necessary styling for the grid layout of the advanced search inputs in `_campaign_articles.styl`.
*   **Tests:** Migrated and verified RSpec feature tests in `campaign_articles_spec.rb` to ensure search behavior and form visibility function correctly.

## AI usage
Gemini AI was used to write this description and fix the tests.

## Screenshots
Before:
Captura de Tela 2026-04-28 às 15 01 19

After:
Captura de Tela 2026-04-28 às 15 00 07

I will be attentive to any changes or suggestions


joserafael:
Hi @ragesoss pelase review. Thank you :-)

ragesoss:
AI-assisted review feedback follows — read it, sanity-check it against the code, and use or ignore it based on your own judgment. None of it should be acted on without verification.

This is the focused split-out of #6814, which is great — it's much easier to review now. A few items from the prior review on #6814 still apply, and one new one surfaced from the changes since:

**1. Stale PR-description claim about `open_details`**

The description says: *"Bug Fix: Addressed a minor issue in `CampaignsController#update` related to the `open_details` expansion logic."*

But this PR doesn't touch `@open_details`. The line at `campaigns_controller.rb:161` is unchanged from master (`.empty?`). That bullet looks like leftover text from #6814's description (where the value was inverted). Worth removing from the description to avoid confusion.

**2. Filter logic is still untested**

`spec/features/campaign_articles_spec.rb` only adds `?locale=en` to two existing examples and swaps an i18n key. Nothing in the PR exercises the new title, school, char/refs/views range filters, or the new `total_count` path in `paginated_articles_courses`. The filter logic in `Query::RankedArticlesCoursesQuery` (≈80 added lines of branching) has no spec coverage, and a regression like "school filter silently doesn't apply" would not be caught by CI. A minimal feature spec for each filter type would be enough.

**3. LIKE filters don't escape user-supplied `%` and `_`**

```ruby
q.joins(:article).where('articles.title LIKE ?', "%#{@article_title}%")
```

Same for `course_title`. A user typing `100%` or `foo_bar` will match more rows than they likely expect (any character / any character in those positions). The fix is `ActiveRecord::Base.sanitize_sql_like(@article_title)` before interpolating into the pattern. This was flagged on #6814 and still applies.

**4. Raw-SQL column interpolation in `Query::RankedArticlesCoursesQuery`**

In `apply_min_max`:

```ruby
q = q.where("#{column} >= ?", min_val) if min_val.present?
```

And in `order_clause_for_articles`:

```ruby
sql_column = column_map[column] || column     # falls back to user input
direction_sql = direction.upcase
order_parts << "#{sql_column} #{direction_sql}"
```

Both are gated today by `CampaignsController#set_sort` validating `@sort_column` against a whitelist and direction against `ASC/DESC`, and by `apply_min_max` only being called with hardcoded column strings. So **not actively exploitable**. But the query class is trusting that its callers validated the inputs — a future caller that misses that step exposes SQL injection (and `apply_min_max`'s contract isn't visible from the method signature). Two cheap hardenings:

- Whitelist `column` in `apply_min_max` against the three known column strings (or pass a symbolic key and look it up inside the query class).
- Drop the `|| column` fallback in `order_clause_for_articles`, or validate `column` is in `column_map` and raise/no-op otherwise. Also validate `direction.upcase` is `ASC`/`DESC` before interpolating.

**5. `course_title` filter is plumbed in but unreachable from this view**

`CoursesPresenter` reads `options[:course_title]` and passes it through to the query, but `CampaignsController#set_presenter` doesn't pass `params[:course_title]` and `articles.html.haml` has no input for it. Either intentional (mirroring the programs pattern for symmetry) or dead wiring. Worth a comment in code or removal.

**6. Minor: inline style in HAML**

```haml
.page-entries-info{style: 'text-align: center; color: #666; font-size: 14px; margin-top: 10px;'}
```

The PR adds a new `_campaign_articles.styl` module — this snippet belongs there too rather than in markup.

---

*This early-feedback comment was drafted in a Claude Code session (~25 min, ~10 interactions). Sage directed the triage and chose to send early feedback rather than running specs locally. Findings come from reading the PR diff alongside prior review history on #6814; the controller-side sort validation in `set_sort` was independently checked to verify the raw-SQL paths aren't actively exploitable. The LIKE-wildcard, missing-spec, stale-description, and inline-style observations were spot-checked but not all individually exercised against a running browser.*

ragesoss:
Follow-up after running the branch locally — same caveat as the prior comment: this is AI-assisted, sanity-check against the code before acting on any of it.

**LIKE-wildcard issue (item 3 above) is now reproducible.**

Wrote a small exploratory spec exercising each filter (title / school / char range / refs range / views range / filtered pagination). 7 of 9 examples pass against this branch; the two that fail are both LIKE-wildcard cases:

- searching `title=100%` matches an article titled *"100 dollars"* (the user's `%` is being interpreted as a SQL wildcard, so the pattern is `'%100%%'` and `LIKE` matches *anything containing "100" followed by anything*).
- searching `title=P_zza` matches *"Pizza"* (the user's `_` is being interpreted as the SQL single-character wildcard).

A minimal scaffold for reproducing / for the eventual fix:

```ruby
# spec/features/campaign_articles_filters_spec.rb
# (AI-generated during review; ran locally on this branch, 7/9 pass,
#  the two failures are the LIKE-wildcard cases above)

require 'rails_helper'

describe 'campaign articles advanced search filters', type: :feature, js: true do
  let(:campaign) { create(:campaign) }
  let(:course_a) { create(:course, slug: 'School_A/Course_A_(2024)', school: 'School A') }
  let(:course_b) { create(:course, slug: 'School_B/Course_B_(2024)', school: 'School B') }
  let(:pizza)    { create(:article, title: 'Pizza',       namespace: 0) }
  let(:pasta)    { create(:article, title: 'Pasta',       namespace: 0) }
  let(:cotton)   { create(:article, title: '100% Cotton', namespace: 0) }
  let(:dollars)  { create(:article, title: '100 dollars', namespace: 0) }

  before do
    create(:campaigns_course, campaign:, course: course_a)
    create(:campaigns_course, campaign:, course: course_b)
    create(:articles_course, course: course_a, article: pizza,
                             character_sum: 5000, references_count: 10, view_count: 1000)
    create(:articles_course, course: course_b, article: pasta,
                             character_sum: 200,  references_count: 1,  view_count: 50)
    create(:articles_course, course: course_a, article: cotton,
                             character_sum: 800,  references_count: 3,  view_count: 200)
    create(:articles_course, course: course_b, article: dollars,
                             character_sum: 400,  references_count: 2,  view_count: 80)
  end

  it 'treats user-entered % as a literal, not a SQL wildcard' do
    visit "/campaigns/#{campaign.slug}/articles?locale=en&title=100%25"
    expect(page).to have_content('100% Cotton')
    expect(page).not_to have_content('100 dollars')   # currently fails
  end

  it 'treats user-entered _ as a literal, not a SQL wildcard' do
    visit "/campaigns/#{campaign.slug}/articles?locale=en&title=P_zza"
    expect(page).not_to have_content('Pizza')         # currently fails
  end

  it 'school filter restricts to that school' do
    visit "/campaigns/#{campaign.slug}/articles?locale=en&school%5B%5D=School+A"
    expect(page).to have_content('Pizza')
    expect(page).not_to have_content('Pasta')
  end

  it 'char_added_from applies a min bound' do
    visit "/campaigns/#{campaign.slug}/articles?locale=en&char_added_from=1000"
    expect(page).to have_content('Pizza')
    expect(page).not_to have_content('Pasta')
  end

  # ... (similar for references / views / pagination total)
end
```

Feel free to lift or adapt any of that. The fix is `ActiveRecord::Base.sanitize_sql_like(...)` around `@article_title` (and `@course_title`) before building the `%...%` pattern.

**Two more small things noticed while running locally:**

- `app/assets/javascripts/campaigns_articles.js:1` sets `window.DISABLE_ARTICLES_LISTJS = true;`, but nothing else in the codebase reads that name. Looks like dead code from an earlier iteration.
- `apply_sorting` in `Query::RankedArticlesCoursesQuery` returns the query unchanged when `@too_many` is true, but `too_many_articles?` is computed against the unfiltered `campaign.articles_courses.count`. So a user on a 50K+ campaign who applies filters narrowing to 10 rows still gets an unsorted result. Minor edge case — probably worth either applying sort when the *filtered* count is small, or documenting the behavior.

---

*Drafted in a Claude Code session (~70 min, ~30 interactions total across both comments). Sage directed the full review-PR flow (orientation → triage → early feedback → branch checkout → build → exploratory specs → code review → follow-up). Locally verified: `yarn build` clean; rubocop/eslint clean on changed files; PR's own specs pass; the LIKE-wildcard failures above were reproduced via an exploratory feature spec. The `DISABLE_ARTICLES_LISTJS` finding is from a grep of the JS source. The exploratory spec itself was AI-generated and run once locally; not added to the codebase.*

ragesoss (changes requested):
The build failure on this PR is from a stale-rebase mismatch in `spec/features/assigned_articles_spec.rb`. The CI checks out the PR merged into the latest master, and on that merged tree the spec can't find its Feedback / Delete controls.

### What happened

This PR's commit *Re-enable assigned articles spec* (`a9cc0f50`) writes:

```ruby
within 'tr.assignment', text: 'Nancy Tuana' do
  find('a.button', text: /Feedback/i).click
end
…
within '.article-viewer.feedback' do
  find('a.button', text: /Delete/i).click
end
```

Those `a.button` selectors were correct against `app/assets/javascripts/components/common/feedback.jsx` as of when this branch was cut. But commit `f43d7e427` on master (*Enforce jsx-a11y/anchor-is-valid; convert button-shaped anchors*, 2026-05-13) converted three relevant controls — the Feedback button, the suggestions-feedback button, and the delete-suggestion button — from anchor-shaped buttons to real `<button>` elements. Once the merge into master happens at CI time, the selectors no longer match. (That commit's own message notes: *"Four feature specs that drove the converted controls via `click_link` were updated to use `click_button`."*)

So the failure is purely a test-vs-DOM mismatch — nothing wrong with the campaign-articles work itself.

### Suggested patch

In `spec/features/assigned_articles_spec.rb`, swap the two `find('a.button', …)` calls for `click_button` (matches both `<button>` and `<input type="submit">`, and Capybara already scopes it within the surrounding `within` block):

```diff
       within 'tr.assignment', text: 'Nancy Tuana' do
-        find('a.button', text: /Feedback/i).click
+        click_button('Feedback')
       end

       expect(page).to have_selector('textarea.feedback-form')
       find('textarea.feedback-form').fill_in with: 'This is a great article!'
       click_button 'Add Suggestion'

       # Scopes the delete action to the suggestions/feedback viewer modal
       within '.article-viewer.feedback' do
-        find('a.button', text: /Delete/i).click
+        click_button('Delete')
       end
```

After rebasing this branch onto current master and applying that change, the spec should pass (modulo the existing intermittent WikiApi/LiftWing concerns that motivated the original `pending`).

A separate code review of the rest of the PR will follow.

ragesoss:
A substantive review beyond the build-fix comment above — Claude-assisted, so treat findings as observations to consider, not directives.

### 1. Sort selection doesn't visibly reorder the page (regression)

The PR markets server-side sorting, and `RankedArticlesCoursesQuery#subquery` does build a correctly-ordered subquery. But `RankedArticlesCoursesQuery#scope` joins it via:

```ruby
ArticlesCourses
  .joins("INNER JOIN (#{subquery.to_sql}) AS ranked_articles USING (id)")
  .select(...)
```

The outer relation has no `ORDER BY`. MariaDB does not promise to preserve the subquery's ordering across the join, and in practice it returns rows in (essentially) PK order. The PR's own new unit spec captures this — three failing examples in `spec/presenters/query/ranked_articles_courses_query_spec.rb` (`apply_sorting` context). Output excerpts:

```
expected: [703, 705, 704, 706]
     got: [703, 704, 705, 706]
```

A small feature spec that asserts visible row order for default/title/views sorts also confirms it. All three variants return rows in identical insertion order:

```
expected: ["Alpha", "Mu", "Zeta", "Beta"]
     got: ["Zeta", "Alpha", "Mu", "Beta"]
```

This pattern exists on master too, but it's been masked by `lists.js` taking over sorting client-side. The PR sets `window.DISABLE_ARTICLES_LISTJS = true` in `campaigns_articles.js`, which removes that fallback for the campaign articles table. Net effect: clicking a column header now navigates with new sort params but the rendered rows reorder by **nothing**.

Two fixes that should work — either apply an `ORDER BY` to the outer relation (cheapest if the subquery is the deferred-join optimization step), or drop the deferred-join pattern and order in a single query.

The exploratory spec used to verify this:

```ruby
# spec/features/exploratory_campaign_articles_sort_order_spec.rb
require 'rails_helper'

describe 'campaign articles sort order', type: :feature, js: true do
  let(:campaign) { create(:campaign) }
  let(:course) { create(:course, slug: 'course-1', school: 'University A') }
  let!(:campaigns_course) do
    create(:campaigns_course, campaign_id: campaign.id, course_id: course.id)
  end

  before do
    [['Zeta', 100], ['Alpha', 500], ['Mu', 300], ['Beta', 50]].each do |title, chars|
      art = create(:article, title:)
      create(:articles_course, course:, article: art, character_sum: chars,
                               references_count: 1, view_count: chars)
    end
  end

  def visible_titles
    page.all('table.table tbody tr td.title a').map(&:text)
  end

  it 'orders rows by character_sum DESC by default' do
    visit "/campaigns/#{campaign.slug}/articles?locale=en"
    expect(visible_titles).to eq(%w[Alpha Mu Zeta Beta])
  end

  it 'orders rows by title ASC when sort=title direction=asc' do
    visit "/campaigns/#{campaign.slug}/articles?locale=en&sort=title&direction=asc"
    expect(visible_titles).to eq(%w[Alpha Beta Mu Zeta])
  end

  it 'orders rows by views ASC when sort=views direction=asc' do
    visit "/campaigns/#{campaign.slug}/articles?locale=en&sort=views&direction=asc"
    expect(visible_titles).to eq(%w[Beta Zeta Mu Alpha])
  end
end
```

This spec was written and run by Claude Code against both this branch and master; all three examples failed on both (master is masked by the lists.js client fallback in the real browser; the deferred-join order-loss is the underlying cause). Worth adopting it — the existing feature spec only checks page membership, never order.

### 2. Sort-column allowlist relaxed

Master's `RankedArticlesCoursesQuery` has:

```ruby
SORTABLE_COLUMNS = %w[title view_sum character_sum references_count].freeze
# falls back to DEFAULT_ORDER if @sort_column isn't whitelisted
```

The PR replaces this with:

```ruby
sql_column = column_map[column] || column
```

If a request arrives with `?sort=anything_else&direction=asc`, that bare word is concatenated into the `ORDER BY` string. Rails' `ActiveRecord::UnknownAttributeReference` guard catches obvious patterns like `(SELECT 1)` or `ASC;`, but a bare attacker-controlled column name (e.g. `?sort=secret_column`) passes the regex and either crashes on a non-existent column (500 error / DoS) or references the wrong table silently. Restoring an explicit allowlist (e.g. `return DEFAULT_ORDER unless column_map.key?(column)`) preserves the prior safety net.

### 3. Magic number `50000`

```ruby
def apply_sorting(query)
  return query if total_count > 50000
  ...
end
```

This replaces the prior `@too_many` mechanism (whose threshold is owned by `CoursesPresenter#too_many_articles?` and also drives the "list is too long to display" UI). Now there are two thresholds in two places. Either reuse `@too_many`, or hoist the constant and use it for both decisions.

### 4. PR re-enables `assigned_articles_spec.rb` — not in description

That re-enablement is what made CI fail (covered in the earlier build-fix comment). Worth either mentioning in the PR description or moving it to its own PR — the failure path looked like this PR's own work when it was actually an unrelated revival.

### 5. Smaller items

- **Inline styles in haml**: `.page-entries-info{style: 'text-align: center; color: #666; font-size: 14px; margin-top: 10px;'}` should live in the new `_campaign_articles.styl`.
- **Hardcoded `placeholder: 'Min'`/`'Max'`** in the form — not translated; consider adding strings under `form_search` or `metrics`.
- **`renderer: WillPaginate::LinkRenderer`** is the default — passing it explicitly is noise.
- **Param naming inconsistency**: the controller maps `params[:title]` → `articles_title:`, the presenter holds it as `@articles_title`, the query field is `article_title:` (singular). One concept, three names — pick one (`article_title` matches the existing translation key).
- **RuboCop**: 13 offenses on the changed Ruby files (one `Metrics/MethodLength` on `apply_text_filters` at 18/16 lines; the rest are `Layout/LineLength` in the specs). `bundle exec rubocop -A` cleans most.

### What looks good

- SQL wildcard escaping via `ActiveRecord::Base.sanitize_sql_like` on both title filters (good — the dedicated spec examples for `100%` and `P_zza` are tight).
- `school_options` reuses the existing presenter helper.
- Existing `campaigns_controller_spec.rb` and `courses_presenter_spec.rb` still pass (49 examples) after the changes.
- Build (`yarn build`) and ESLint both clean on the new JS bundle.

---

*Drafted in a Claude Code session (~30 min, ~25 interactions). Sage directed the investigation phases, but verification of the sort-doesn't-actually-sort regression came from a Claude-written exploratory feature spec (included above; run by Claude against both branches). The SQL/whitelist/threshold observations were not independently verified beyond reading the diff and inspecting generated SQL. Sage read the full draft before posting but did not re-verify each item individually.*

Issue #6550, opened 2025-11-05
Some course and timeline data isn't correctly transferred when copying a course from another server


ragesoss:
The "copy course from another server" feature (at `/copy_course`) has some glitches. The ones I'm aware are:

* If timeline content includes relative links, as it does for the  "Wikipedia assignment journal" assignment block that the assignment wizard can generate, those links may be broken on the new server.
* The `peer_review_count` flag setting doesn't not copy over.

lakshita10341:
working on it

lakshita10341:
Hi @ragesoss , I tried copying several courses, but was not able to find broken links, can you provide example of any course, so that i can be sure my changes are fixing it well.

ragesoss:
@lakshita10341 which courses did you copy? In particular, this issue happens when copying courses from dashboard.wikiedu.org to outreachdashboard.wmflabs.org (or, I would expect, a local dev server that is configured in `wiki_education: false` mode).

Ajay-singh1:
Picking this up :)

nexpectArpit:
@ragesoss can i work on this issue?

ragesoss:
@nexpectArpit sure

nexpectArpit:
@ragesoss while researching this issue, i found a 3rd glitch that is : 
while timeline tasks are copied , code manually copies the title and desc. but ignores  `due_date` and `points` while copying `Block` items for the timeline .

nexpectArpit:
@ragesoss i guess this issue work has been done!

PR #6890, opened 2026-06-06
 Fix glitches while copying course from another server

nexpectArpit:
#### What this PR does
This PR addresses three data-loss and formatting glitches that occur when copying a course from another server (e.g., from WikiEdu to Outreach Dashboard):
1. **Preserve `flags` when copying:**  The code now uses `.merge(timeline_enabled: true)` to ensure existing flags are safely preserved.
2. **Convert relative links to absolute links in timeline content:** The `CopyCourse` service now checks the block content and prepends the original server's `@host` to any relative `href="/..."` paths.
3. **Preserve custom assignment `due_date` and `points`:** When creating new blocks from the copied timeline data, the previous code mapped the title and content but ignored the `due_date` and `points` properties. These attributes are now explicitly mapped and saved on the new `Block` record to prevent data loss for custom assignments.

addresses #6550

### DIFF
--- a/app/controllers/copy_course_controller.rb
+++ b/app/controllers/copy_course_controller.rb
@@ -14,8 +14,8 @@ class CopyCourseController < ApplicationController
                   flash: { error: "Course not created: #{response[:error]}" })
     else
       course = response[:course]
+      course.flags = course.flags.merge(timeline_enabled: true)
       course.update(
-        flags: { timeline_enabled: true },
         cloned_status: 3,
         expected_students: course.expected_students || 0,
         term: ''
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -499,6 +499,14 @@ class Course < ApplicationRecord
     edit_settings['enrollment_edits_enabled']
   end
 
+  FEATURE_FLAG_KEYS = %w[
+    wiki_edits_enabled event_sync peer_review_count timeline_enabled
+    online_volunteers_enabled stay_in_sandbox no_sandboxes
+    retain_available_articles disable_student_emails review_bibliography
+    declined very_long_update max_group_size article_scoped
+    closed_date register_accounts
+  ].freeze
+
   def controlled_by_event_center?
     flags[:event_sync].present?
   end
--- a/app/services/copy_course.rb
+++ b/app/services/copy_course.rb
@@ -78,15 +78,7 @@ class CopyCourse # rubocop:disable Metrics/ClassLength
   def symbolize_feature_flags(flags_hash)
     return unless flags_hash.is_a?(Hash)
 
-    feature_flag_keys = %w[
-      wiki_edits_enabled event_sync peer_review_count timeline_enabled
-      online_volunteers_enabled stay_in_sandbox no_sandboxes
-      retain_available_articles disable_student_emails review_bibliography
-      declined very_long_update max_group_size article_scoped
-      closed_date register_accounts
-    ]
-
-    feature_flag_keys.each do |key|
+    Course::FEATURE_FLAG_KEYS.each do |key|
       if flags_hash.key?(key)
         flags_hash[key.to_sym] = flags_hash.delete(key)
       end
--- a/app/services/copy_course.rb
+++ b/app/services/copy_course.rb
@@ -41,6 +41,10 @@ class CopyCourse # rubocop:disable Metrics/ClassLength
       copied_data['flags']['update_logs'] =
         fix_update_logs_parsing(copied_data['flags']['update_logs'])
     end
+    # The flags hash is copied via JSON so all keys are strings.
+    # We must convert known feature flags to symbols because Course model
+    # readers (e.g. peer_review_count) look for symbol keys.
+    symbolize_feature_flags(copied_data['flags'])
     # Create the course
     @course = Course.create!(copied_data)
   end
@@ -69,6 +73,24 @@ class CopyCourse # rubocop:disable Metrics/ClassLength
   # This causes problems, so we need to force the keys to be integers.
   def fix_update_logs_parsing(update_logs)
     update_logs.transform_keys(&:to_i)
+  end
+
+  def symbolize_feature_flags(flags_hash)
+    return unless flags_hash.is_a?(Hash)
+
+    feature_flag_keys = %w[
+      wiki_edits_enabled event_sync peer_review_count timeline_enabled
+      online_volunteers_enabled stay_in_sandbox no_sandboxes
+      retain_available_articles disable_student_emails review_bibliography
+      declined very_long_update max_group_size article_scoped
+      closed_date register_accounts
+    ]
+
+    feature_flag_keys.each do |key|
+      if flags_hash.key?(key)
+        flags_hash[key.to_sym] = flags_hash.delete(key)
+      end
+    end
   end
 
   def add_tracked_wikis
--- a/app/services/copy_course.rb
+++ b/app/services/copy_course.rb
@@ -160,7 +160,8 @@ class CopyCourse # rubocop:disable Metrics/ClassLength
   def copy_blocks(week, blocks)
     blocks.each do |block_data|
       block = Block.create!(content: block_data['content'], title: block_data['title'],
-                            week_id: week.id, order: block_data['order'], kind: block_data['kind'])
+                            week_id: week.id, order: block_data['order'], kind: block_data['kind'],
+                            due_date: block_data['due_date'], points: block_data['points'])
       update_block_content(block, block_data)
     end
   end
@@ -180,6 +181,8 @@ class CopyCourse # rubocop:disable Metrics/ClassLength
     content_additions.reverse_each do |kind, addition|
       final_content = headings[kind] + addition + final_content unless addition.empty?
     end
+
+    final_content.gsub!('href="/', "href=\"#{@host}/") if final_content.present?
 
     block.update!(content: final_content)
   end
--- a/spec/services/copy_course_spec.rb
+++ b/spec/services/copy_course_spec.rb
@@ -291,4 +291,117 @@ describe CopyCourse do
       expect(block.content).to include('<h4 class="timeline-exercise">Exercise</h4>')
     end
   end
+
+  describe 'additional copy features' do
+    let(:slug) { 'University_of_South_Carolina/Invertebrate_Zoology_(COPIED_FROM_Spring_2022)' }
+    let(:course_url) { url_base + slug + '/course.json' }
+    let(:categories_url) { url_base + slug + '/categories.json' }
+    let(:users_url) { url_base + slug + '/users.json' }
+    let(:timeline_url) { url_base + slug + '/timeline.json' }
+    let(:training_modules_url) { @selected_dashboard + '/training_modules.json' }
+
+    let(:course_response_body) do
+      {
+        course: {
+          id: 15_907,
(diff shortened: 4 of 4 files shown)


ragesoss (changes requested):
Note on AI-assisted review: the feedback below was produced with Claude Code. It isn't meant to be trusted automatically or acted on blindly — read it, verify what matters, and use or discard each point based on your own judgment.

**Flag preservation doesn't reach the flags the issue names.**

The `.merge` change preserves the raw `flags` hash in storage, which is a real improvement for the string-keyed flags (`update_logs`, `average_update_delay`). But the feature flags called out in #6550 — `peer_review_count` in particular — are read through **symbol** keys on the model:

```ruby
def peer_review_count
  flags[:peer_review_count]
end

def online_volunteers_enabled?
  flags[:online_volunteers_enabled].present?
end
```

Flags copied through `course.json` arrive as **string** keys (JSON has no symbols), and `serialize :flags, type: Hash` has no indifferent access. So the copied flags are stored but the symbol-keyed readers can't see them. A spec written during review confirms it, against a copied course whose source flags were `{ "peer_review_count" => 2, "online_volunteers_enabled" => true }`:

```
copied_course.flags                         # => includes "peer_review_count" => 2   (stored)
copied_course.peer_review_count             # => nil                                 (not readable)
copied_course.online_volunteers_enabled?    # => false                               (not readable)
```

`timeline_enabled?` works only because the controller re-adds `:timeline_enabled` as a symbol. Every other symbol-keyed flag — `peer_review_count`, `online_volunteers_enabled`, `stay_in_sandbox`, etc. — remains dead after copy, and fails silently rather than raising. So the first item in the PR description isn't achieved end-to-end for the flag the issue is about.

Note that a blanket `deep_symbolize_keys` on the copied flags would break the string-keyed readers (`flags['update_logs']`, `flags['average_update_delay']`), so the fix needs to be targeted — symbolize the known feature-flag keys, or give the readers indifferent access. The choice is a maintainer call.

**Relative-link rewriting** works for the wizard-generated case and correctly leaves absolute `https://` hrefs alone (both verified). Minor edge cases, likely out of scope but worth a look: protocol-relative `href="//other.com"` becomes `href="https://host//other.com"`, and `src="/..."` (images) and single-quoted hrefs aren't handled.

**`due_date` / `points` preservation** works (verified).

**Tests:** none of the three fixes have spec coverage. An AI-generated exploratory spec covering all three (and reproducing the `peer_review_count` gap above) is posted as a follow-up comment to adopt — it was run locally during review; it has not been hand-audited line by line.

---
*Drafted in a Claude Code session (~20 min, ~7 interactions). Sage directed the review and the code walkthrough, and reviewed this comment before posting. The findings — the `peer_review_count` symbol/string key gap, the relative-link behavior, and `due_date`/`points` — were each reproduced via a new local spec run by Claude Code; Sage reviewed the results but did not independently re-run them. The exploratory spec is AI-generated.*

ragesoss:
Exploratory spec written during review, covering all three fixes and reproducing the `peer_review_count` gap from the review comment. It is AI-generated (Claude Code) and was run locally — the flag-preservation examples fail against the current branch, the relative-link and `due_date`/`points` examples pass. Drop it at `spec/services/copy_course_spec.rb` (or a new file) and adapt as needed.

```ruby
# frozen_string_literal: true

require 'rails_helper'

# Verifies the three claimed fixes:
#   1. flags preserved (not overwritten) when copying
#   2. relative href links rewritten to absolute against the source host
#   3. custom assignment due_date and points preserved on copied blocks
describe CopyCourse do

  let(:url_base) { 'https://dashboard.wikiedu.org/courses/' }
  let(:slug) { 'University_of_South_Carolina/Invertebrate_Zoology_(COPIED_FROM_Spring_2022)' }
  let(:host) { 'https://dashboard.wikiedu.org' }
  let(:course_url) { url_base + slug + '/course.json' }
  let(:categories_url) { url_base + slug + '/categories.json' }
  let(:users_url) { url_base + slug + '/users.json' }
  let(:timeline_url) { url_base + slug + '/timeline.json' }
  let(:training_modules_url) { host + '/training_modules.json' }

  # Source course flags as they appear in course.json: JSON has no symbols, so
  # peer_review_count arrives as a STRING key even though the model stored it
  # under a symbol key (flags[:peer_review_count]) on the source server.
  let(:course_response_body) do
    {
      course: {
        id: 15_907,
        slug: 'University_of_South_Carolina/Invertebrate_Zoology_(Spring_2022)',
        school: 'University of South Carolina',
        title: 'Invertebrate Zoology',
        term: 'Spring 2022',
        start: '2022-01-11T00:00:00.000Z',
        end: '2022-04-30T23:59:59.000Z',
        type: 'ClassroomProgramCourse',
        home_wiki: { id: 1, language: 'en', project: 'wikipedia' },
        flags: { 'peer_review_count' => 2, 'online_volunteers_enabled' => true },
        training_library_slug: 'students',
        wikis: [{ language: 'en', project: 'wikipedia' }]
      }
    }.to_json
  end

  let(:categories_response_body) { { course: { categories: [] } }.to_json }
  let(:users_response_body) { { course: { users: [] } }.to_json }

  # Block content with a relative link (as the assignment wizard generates for
  # the "Wikipedia assignment journal" block) plus a custom assignment block
  # carrying due_date and points.
  let(:timeline_response_body) do
    {
      course: {
        weeks: [
          {
            id: 57_366, order: 1, title: 'Week 1',
            blocks: [
              {
                id: 1, kind: 0, week_id: 57_366, order: 1,
                title: 'Journal',
                content: 'See <a href="/training/students/keeping-a-research-journal">the journal</a> ' \
                         'and <a href="https://en.wikipedia.org/wiki/Foo">Foo</a>.',
                due_date: nil, points: nil, training_module_ids: []
              },
              {
                id: 2, kind: 1, week_id: 57_366, order: 2,
                title: 'Custom assignment',
                content: 'Do the thing.',
                due_date: '2022-03-15', points: 25, training_module_ids: []
              }
            ]
          }
        ]
      }
    }.to_json
  end

  def stub_all_requests
    stub_request(:get, course_url).to_return(status: 200, body: course_response_body)
    stub_request(:get, categories_url).to_return(status: 200, body: categories_response_body)
    stub_request(:get, users_url).to_return(status: 200, body: users_response_body)
    stub_request(:get, timeline_url).to_return(status: 200, body: timeline_response_body)
    stub_request(:get, training_modules_url)
      .to_return(status: 200, body: { training_modules: [] }.to_json)
  end

  let(:copied_course) do
    described_class.new(url: url_base + slug, user_data: '0').make_copy
    Course.find_by(slug:)
  end

  describe 'flag preservation' do
    before { stub_all_requests }

    it 'keeps the raw flags hash from the source course' do
      expect(copied_course.flags).to include('peer_review_count' => 2)
    end

    it 'exposes peer_review_count through the model reader after copy' do
      # The reader is flags[:peer_review_count] (symbol). The copied flag is a
      # string key. This asserts the flag is actually usable, not just stored.
      expect(copied_course.peer_review_count).to eq(2)
    end

    it 'exposes online_volunteers_enabled? through the model reader after copy' do
      expect(copied_course.online_volunteers_enabled?).to eq(true)
    end
  end

  describe 'relative link rewriting' do
    before { stub_all_requests }

    it 'rewrites relative hrefs to absolute against the source host' do
      block = copied_course.weeks.first.blocks.find_by(title: 'Journal')
      expect(block.content).to include('href="https://dashboard.wikiedu.org/training/students/')
    end

    it 'leaves already-absolute hrefs untouched' do
      block = copied_course.weeks.first.blocks.find_by(title: 'Journal')
      expect(block.content).to include('href="https://en.wikipedia.org/wiki/Foo"')
    end
  end

  describe 'custom assignment fields' do
    before { stub_all_requests }

    it 'preserves due_date and points' do
      block = copied_course.weeks.first.blocks.find_by(title: 'Custom assignment')
      expect(block.due_date).to eq(Date.parse('2022-03-15'))
      expect(block.points).to eq(25)
    end
  end
end
```

ragesoss (changes requested):
Note on AI-assisted review: this follow-up was produced with Claude Code. It isn't meant to be trusted automatically or acted on blindly — read it, verify what matters, and use or discard each point based on your own judgment.

**The symbolization fix works.** Verified locally on the branch (merged with current master): `symbolize_feature_flags` makes the copied flags readable through the model readers — `peer_review_count`, `online_volunteers_enabled?`, etc. now return the source course's values after copy. The adopted specs pass (11 examples in `copy_course_spec.rb`, plus the controller specs), RuboCop is clean, and the allowlist was cross-checked against every `flags[:...]` read in `app/models/course.rb` — all 16 symbol-keyed feature flags are covered, and the omitted operational flags (`salesforce_id`, `timeslice_duration`, `no_max_users`, etc.) are ones a copied course should *not* inherit.

**One change requested before merge: move the flag-key list out of `CopyCourse`.**

As written, the list is a hardcoded local inside `CopyCourse#symbolize_feature_flags`, far from the flag readers in `Course`. When a future symbol-keyed feature flag is added to `Course`, nothing will prompt anyone to update this list — and the failure mode is the same silent one this PR fixes: the flag copies across as a dead string key and the reader quietly returns `nil`.

Suggested shape: define the list as a constant on `Course`, adjacent to the flag readers (e.g. `FEATURE_FLAG_KEYS` or similar), and have `CopyCourse` reference it. That way the list lives next to the code it must stay in sync with, and a new reader added there is much more likely to get a matching list entry in the same diff. Exact naming and placement are the author's call.

---
*Drafted in a Claude Code session (~20 min, ~4 interactions). Sage directed the re-review and reviewed this comment before posting. Verification was run by Claude Code locally: the PR branch was merged with master, the adopted specs and controller specs were run (all passing), and the allowlist was cross-checked against the `flags[:...]` reads in `course.rb`; Sage reviewed the results but did not independently re-run them.*

ragesoss (approved):
Note on AI-assisted review: this follow-up was produced with Claude Code. It isn't meant to be trusted automatically or acted on blindly — read it, verify what matters, and use or discard each point based on your own judgment.

**The remaining change from the prior review is addressed.** The feature-flag allowlist now lives as `Course::FEATURE_FLAG_KEYS`, defined directly above the flag readers in `course.rb`, and `CopyCourse#symbolize_feature_flags` references it. This puts the list next to the code it must stay in sync with, so a newly added symbol-keyed reader is much more likely to get a matching entry in the same diff — which was the concern.

Verified locally on the branch (merged with current master):
- The allowlist covers every symbol-keyed `flags[:...]` read in `course.rb` — all 16 — with none missing and none extra. The string-keyed flags (`update_logs`, `edit_settings`, `academic_system`, `format`) are correctly left out.
- The extraction is behavior-preserving: the key list is identical to the previous inline version.
- The PR's specs pass (15 examples, including the assertions that exercise `peer_review_count` and `online_volunteers_enabled?` through the model readers, not just storage), and RuboCop is clean on the changed files.

Approving.

Non-blocking, carried over from the first review and noted then as likely out of scope: relative-link rewriting still doesn't handle protocol-relative `href="//host"` (becomes `https://host//host`), `src="/..."` for images, or single-quoted hrefs. Reasonable to leave for a follow-up.

---
*Drafted in a Claude Code session (~10 min, ~6 interactions). Sage directed the re-review and reviewed this comment before posting. Verification was run by Claude Code locally: the PR branch was merged with master, the allowlist was cross-checked against the `flags[:...]` reads in `course.rb`, and the PR's specs and controller specs were run (15 passing, RuboCop clean); Sage reviewed the results but did not independently re-run them.*

PR #6891, opened 2026-06-06
Fix text overflow in Milestones sidebar and Timeline blocks

nexpectArpit:
#### Problem
While reproducing the bug #6550 during copying the course https://dashboard.wikiedu.org/courses/University_of_California_Santa_Cruz/Antarctic_Ecology,_Geology,_History,_and_Policy_(Spring_2026)/home , we can see the text overflow in Milestones section.

#### what does PR do
Added the `word-break: break-word` CSS property directly to the `.markdown` container inside `.milestones` (`_overview.styl`) and the `.block__block-content` container (`_timeline.styl`). 

## Screenshots
Before:
Screenshot 2026-06-06 at 4 02 24 PM

After:
Screenshot 2026-06-06 at 5 25 54 PM

### DIFF
--- a/app/assets/stylesheets/modules/_overview.styl
+++ b/app/assets/stylesheets/modules/_overview.styl
@@ -32,6 +32,7 @@
       opacity 0.5
   .markdown
     color body-color
+    word-break break-word
   .section-header:last-child
     margin-bottom 0
     hr
--- a/app/assets/stylesheets/modules/_overview.styl
+++ b/app/assets/stylesheets/modules/_overview.styl
@@ -32,7 +32,7 @@
       opacity 0.5
   .markdown
     color body-color
-    word-break break-word
+    overflow-wrap anywhere
   .section-header:last-child
     margin-bottom 0
     hr
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -87,7 +87,7 @@
 
 .block__block-content
   color $text_dk
-  word-break break-word
+  overflow-wrap anywhere
 
 .block.block--orderable
   display flex
--- a/app/assets/stylesheets/modules/_timeline.styl
+++ b/app/assets/stylesheets/modules/_timeline.styl
@@ -87,6 +87,7 @@
 
 .block__block-content
   color $text_dk
+  word-break break-word
 
 .block.block--orderable
   display flex


ragesoss (changes requested):
A note on this feedback: it comes from a Claude Code review session and is meant to be read and weighed, not automatically trusted or acted on. Use or discard based on your own judgment.

The fix is functionally correct — long unbreakable strings now wrap in both the Milestones sidebar and the timeline blocks. One observation worth verifying on the property choice:

`word-break: break-word` is a deprecated, non-standard value. Browsers implement it as a legacy alias for `word-break: normal` + `overflow-wrap: anywhere`. The standard property `overflow-wrap: anywhere` produces the identical result and is already the convention in this codebase — see `_assignments.styl:5`, `_students.styl:103`, and `_training_module_composer.styl:303`. `overflow-wrap` is used nearly everywhere in the stylesheets; this would be the only `word-break: break-word`.

Worth noting why a plain wrap rule wasn't already enough: `.overview .sidebar` already declares `overflow-wrap break-word` at `_overview.styl:6`, and `overflow-wrap` is inherited — so `.markdown` was receiving it yet still overflowing. The reason is that `.markdown` sits inside `.module__data`, a flex item of `.section-header { display: flex }`; flex items default to `min-width: auto` and won't shrink below their content's min-content width, which `overflow-wrap: break-word` does not reduce. `overflow-wrap: anywhere` (what `word-break: break-word` aliases to) does reduce it, which is why the chosen value works. Substituting `overflow-wrap: anywhere` keeps that behavior while staying standards-compliant and consistent with the rest of the stylesheets.

*Drafted in a Claude Code session (~10 min, a few interactions). Sage directed the review; the finding was derived from reading the stylesheets and component markup and was reviewed before posting, but was not verified by building/rendering locally. Comment written by Claude Code.*

ragesoss (approved):
A note on this feedback: it comes from a Claude Code review session and is meant to be read and weighed, not automatically trusted or acted on. Use or discard based on your own judgment.

The prior review's one request is addressed: both rules now use the standard `overflow-wrap: anywhere` instead of the deprecated `word-break: break-word`, consistent with the rest of the stylesheets. The change was built locally this time — `yarn build` compiles cleanly and the compiled CSS emits `overflow-wrap:anywhere` for both `.milestones .markdown` and `.block__block-content`. Looks good to merge.

*Drafted in a Claude Code session (~10 min, a few interactions). Sage directed the re-review. The fix was verified by building locally and inspecting the compiled CSS output, but not by rendering the overflow case in a browser. Comment written by Claude Code.*

PR #6893, opened 2026-06-08
Populates the registered_at field when a new user Joins a course

### LINKED ISSUE #6886: Registered_at field does not display for new users created during a Course
### What is happening?
When a user creates a new Wikimedia account during the course period, the `registered_at` field in the Students CSV export is empty for some users. This also affects the registered_during_project column, which depends on registered_at being populated to correctly identify editors who joined specifically for the program

### To Reproduce

- Create or open a course on the Dashboard
- Enroll a user who creates a new Wikimedia account during the course period
- Navigate to the course overview page and click "Download stats data"
- Download the Editors CSV
- Check the registered_at and registered_during_project columns for recently enrolled users

### Expected behavior
The registered_at column should be populated for all enrolled users, including those who created their accounts during the course period. The `registered_during_projec`t column should correctly reflect true for users whose registration date falls within the course dates.

Formasitchijoh:
## What this PR does
This PR resolve #6886  by  populating  the registered_at field immediately as a background job whenever a new User is created or joins a course

## AI usage
Used AI to debug why the registered_at field was not being populated and implementing the fix

### DIFF
--- a/app/workers/fetch_user_registration_worker.rb
+++ b/app/workers/fetch_user_registration_worker.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/importers/user_importer"
+
+class FetchUserRegistrationWorker
+  include Sidekiq::Worker
+  sidekiq_options lock: :until_executed
+
+  def perform(user_id)
+    user = User.find_by(id: user_id)
+    return unless user
+    UserImporter.update_users([user])
+  end
+end
--- a/lib/importers/user_importer.rb
+++ b/lib/importers/user_importer.rb
@@ -15,12 +15,15 @@ class UserImporter
   end
 
   def self.new_from_omniauth(auth)
-    User.create(
+    user = User.create(
       username: auth.info.name,
       global_id: auth.uid,
       wiki_token: auth.credentials.token,
       wiki_secret: auth.credentials.secret
     )
+    # Populates the registered_at field when ever a new user is created via omniauth
+    FetchUserRegistrationWorker.perform_async(user.id)
+    user
   end
 
   LTR_MARK = 8206.chr # left-to-right mark, Ruby character 8206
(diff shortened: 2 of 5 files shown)


ragesoss (changes requested):
This feedback comes from a Claude Code review session. It's offered as observations to read and weigh against your own judgment — not conclusions to action blindly.

The approach is sound: a deduplicated background job (`lock: :until_executed`, keyed on `user_id`) populates `registered_at` immediately, with the daily `update_users_worker` still acting as a backstop. CI is green, RuboCop is clean, and the affected importer / enrollment / `student_role` feature specs all still pass locally with the inline worker now firing. Two small items before merge, plus one optional spec suggestion.

**1. Typo — `lib/importers/user_importer.rb:23`.** "when ever" → "whenever".

**2. Comment overstates coverage — `lib/importers/user_importer.rb:51`.** It reads "Populates the registered_at field for any user enrolled or added into the course." Because of the early `return user if user` at line 33, the worker is only enqueued for users *newly created* on this path — an existing dashboard user with `registered_at: nil` who is re-enrolled won't trigger it (the daily job remains their route). Reword so the comment doesn't imply broader coverage than the code provides.

**3. (Optional) Pin the actual fix with a one-line spec extension.** The worker spec stubs `UserImporter.update_users`, so nothing currently asserts that `registered_at` ends up populated. This can be covered without a new spec or cassette: the existing `'creates a new user'` example in `spec/lib/importers/user_importer_spec.rb` already runs the inline worker against the `user/new_from_username` cassette, which returns a registration date. Adding a single assertion there —

```ruby
expect(user.reload.registered_at).not_to be_nil
```

— makes that example verify the end-to-end behavior this PR exists to deliver. (Confirmed passing locally against the current cassette.)

*Reviewed in a Claude Code session (~20 min, directed by Sage across a couple of interactions). Claude ran RuboCop and the worker, user-importer, enrollment-path, and `student_role` feature specs, and verified the suggested one-line assertion passes against the existing cassette. Sage directed the review and approved this comment before posting; findings were not otherwise independently re-verified by hand.*

PR #6899, opened 2026-06-11
[WIP] OAuth 2.0 Support

Abishekcs:
#6627

### Note
This is a very early draft for OAuth 2.0. I will add things incrementally. I'm opening this PR to keep everything on track.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,8 @@ gem 'omniauth-mediawiki', git: 'https://github.com/ragesoss/omniauth-mediawiki.g
 # Parses user agent strings to determine which browser is in use.
 # Used for browser support warnings.
 gem 'browser'
+# OAuth 2.0
+gem 'oauth2'
 
 ### Email
 gem 'validates_email_format_of' # Email format validation, used in User model
@@ -66,6 +68,8 @@ gem 'mediawiki_api',
 gem 'restforce' # Salesforce API access
 gem 'oj' # JSON Parsing library
 gem 'rss' # Standard RSS library
+
+
 
 ### Internationalization
 gem 'http_accept_language'
--- a/Gemfile
+++ b/Gemfile
@@ -35,6 +35,8 @@ gem 'omniauth-mediawiki', git: 'https://github.com/ragesoss/omniauth-mediawiki.g
 # Parses user agent strings to determine which browser is in use.
 # Used for browser support warnings.
 gem 'browser'
+# OAuth 2.0
+gem 'oauth2'
 
 ### Email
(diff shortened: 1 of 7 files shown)


Abishekcs:
@ragesoss  about `access` and `refresh` token. Should we store it in a separate table like `oauth_accounts`  with `refrence` to `user` table?

or I should look into something else.

ragesoss:
@Abishekcs i'm not sure. tell me more about what's different in terms of what we have to store.

Abishekcs:
In that case let me do some more research this weekend.

Every `resource` has different suggestion regarding storing the `access` and `refresh` token.

Abishekcs:
> @Abishekcs i'm not sure. tell me more about what's different in terms of what we have to store.

From what I have found:
- **`Access token`**, which allows an application (in this case, the server side of Wiki Edu) to access an API. They are short-**lived**, around a **few hours** I think it was 2 or 8 hours for **the** Wikimedia access token.
- **`Refresh token`** can be used to request a new `access token` when it expires, from **the** Authorization server (**e.g.**, Wikimedia).  **The** lifetime of **a** refresh token is usually **30 to 90 days**, so we can say **they're** long-lived.
 - We get both of **these tokens** when a user uses the `OAuth` login flow.
- Now, storing **the** `access token` **in the** DB is **kind of** bad because **it's** really short-lived, so instead we **store** the token server-side, in a session **store** like Redis (With TTL time to live matching the acces token expiration time).
- As for the `refresh token`, we have to store it in a DB column probably **in** the user table and also **encrypt** the token.

As for storing it client-side, that's a clear no due to risk of cross-site scripting, and it also doesn't match Wiki Edu as we make API calls through the server/backend to Wikimedia all the time.


I have no idea what is the lifetime of `refresh token` for wikimedia.

Abishekcs:
If I find some more useful information will update. Thanks.

Abishekcs:
@ragesoss any suggestion ?

ragesoss:
Hmm... I think a new column on `users` is probably fine here; I don't see any need for a new table for the refresh token.

Lifetime of refresh tokens is apparently 4 hours: https://www.mediawiki.org/wiki/Wikimedia_APIs/Authentication#5._Refresh_token

We already have Redis available, so that makes sense for storing the short-lived ones... although ideally we'd have a fallback that would still let a developer log in to a dev environment without spinning up Redis (which is the current situation).

Abishekcs:
> Lifetime of refresh tokens is apparently 4 hours:

I think you meant `access token` ? Refresh token is `365`

image

Abishekcs:
> without spinning up Redis (which is the current situation).

Okay sure

PR #6903, opened 2026-06-14
add system_stat and facilitator_stat table

lakshita10341:
## What this PR does

1. Adds the database schema for system_stats and facilitators_stats table as part of the GSoC 2026 "System-Wide Metrics & Data Downloads" project.
2. Adds background workers, cron scheduling and helper utilities to automate and manage stats calculations for the `system_stats` and `facilitator_stats` tables.

**Two new tables:**

1. `system_stats`: Daily snapshots of global metrics across all  non-private programs: total edits, article views, articles created,  new editors, active facilitators, per-wiki breakdown, etc.

2. `facilitator_stats` : Weekly snapshots of per-facilitator (organizer)  metrics: programs run, edits generated, new editors created, students  enrolled. Enables WMF to identify impactful organizers.

### System Stats Column Behavior
* **All-Time Cumulative**: `total_edits`, `total_article_views`, `total_articles_improved`, `total_articles_created`, `total_characters_added`, `new_editors_count`, and `new_editors_count_with_preregistration` represent cumulative all-time metrics.
* **Point-in-Time Snapshot**: `active_programs_count` and `active_facilitators_count` represent the counts of active programs and unique facilitators teaching them on that specific calendar day.
* **Archived Count**: `archived_programs_count` represents the total count of completed courses on that day.

### Facilitator Stats Column Behavior
* **All-Time Cumulative**: `total_programs_count`, `total_edits`, `total_students_count`, and `total_characters_added` represent all-time stats for the instructor.
* **Point-in-Time Snapshot**: `active_programs_count` represents the count of courses they are currently/strictly teaching today.
* **Trailing 12-Month Window**: `active_in_last_year` is a boolean flag indicating if the instructor has had any course end within the last 365 days (trailing year), rather than the calendar year.
* **New Editor Counts**: `new_editors_count` (registered strictly during their course) and `new_editors_count_with_preregistration` (registered up to 60 days before the course start) track new Wikipedia accounts created by students in their courses.

## AI usage
I used AI  throughout this PR for:
- writing migrations.
- Understanding how existing analytics code (CourseStatistics, RetainedNewEditorsStats) works.


db/migrate/20260613211819_create_system_stats.rb
```diff
@@ -0,0 +1,23 @@
...
+      t.integer  :archived_programs_count,    default: 0
+      t.integer  :new_editors_count,          default: 0  # registered during course
+      t.integer  :new_editors_count_with_preregistration, default: 0 # includes 60-day pre-window
+      t.integer  :active_facilitators_count,  default: 0
+      t.bigint   :total_characters_added,     default: 0
+      t.text     :wiki_stats                  # Per-wiki breakdown, serialized as Hash
+      t.timestamps
+    end
+
+    add_index :system_stats, :snapshot_date, unique: true
```
Abishekcs, line 21: It will be great if we can know the reason for adding indexes? Like are they really required and if yes,
did you tested it with a lots of data to make sure they are really needed.

Abishekcs:
Same for the rest of the newly added indexes.

lakshita10341:
Hi @Abishekcs ,MySQL requires indexes on foreign keys and unique constraints. So for now, all these indexes are applied.

Furthermore, I've tested the queries with or without indexes with 20k records, and find these:
*   **Query 1** (`where(snapshot_date: latest_date)`) 
    *   With Indexes: 0.91s | Without Indexes: 1.26s
    (The index on `snapshot_date` allows MySQL to instantly narrow down records to the target date instead of scanning the entire table, reducing query time by ~27%.)
    
*   **Query 2: Weekly Stat Upsert** (`upsert_all`) 
    *   With Indexes: 0.31s | Without Indexes: 0.21s
    (Without the unique index, MySQL performs a "blind insert" (just appending rows) without checking for conflicts, which is slightly faster (0.21s) but corrupts the database by inserting duplicate weekly snapshots.)

Abishekcs:
```
  add_index :facilitator_stats, %i[snapshot_date user_id], unique: true
  add_index :facilitator_stats, :snapshot_date
  add_index :facilitator_stats, :user_id
```
I am not sure how the Querys will look `(So, the below question I have might be wrong)` :

But can't the use of index  `add_index :facilitator_stats, :snapshot_date` be covered by `  add_index :facilitator_stats, %i[snapshot_date user_id], unique: true` ?

Abishekcs:
For the `facilitator_stats` table, are we planning to run queries like `WHERE user_id = ?`?

If so, keeping this index makes sense:

```ruby
add_index :facilitator_stats, :user_id
```

However, if the primary query pattern is:

```sql
WHERE snapshot_date = ? AND user_id = ?
```

then this composite index should be sufficient:

```ruby
add_index :facilitator_stats, %i[snapshot_date user_id], unique: true
```

lakshita10341:
> ```
>   add_index :facilitator_stats, %i[snapshot_date user_id], unique: true
>   add_index :facilitator_stats, :snapshot_date
>   add_index :facilitator_stats, :user_id
> ```
> 
> I am not sure how the Querys will look `(So, the below question I have might be wrong)` :
> 
> But can't the use of index `add_index :facilitator_stats, :snapshot_date` be covered by ` add_index :facilitator_stats, %i[snapshot_date user_id], unique: true` ?

Yes, we can do this. I'll remove this index on `snapshot_date`. Thanks!

lakshita10341:
> For the `facilitator_stats` table, are we planning to run queries like `WHERE user_id = ?`?
> 
> If so, keeping this index makes sense:
> 
> ```ruby
> add_index :facilitator_stats, :user_id
> ```
> 
> However, if the primary query pattern is:
> 
> ```sql
> WHERE snapshot_date = ? AND user_id = ?
> ```
> 
> then this composite index should be sufficient:
> 
> ```ruby
> add_index :facilitator_stats, %i[snapshot_date user_id], unique: true
> ```

I'm not sure whether we are going to use this type of queries or not. But can we remove the index on foreign key? The index was autogenerated when I ran `db:migrate` command in `schema.rb` .  For consistency, I added the index in migration file too.
When I tried to remove the index, I encountered this error:
```sql
Caused by:
Mysql2::Error: Cannot drop index 'index_facilitator_stats_on_user_id': needed in a foreign key constraint (Mysql2::Error)
```

Abishekcs:
If it was added automatically. We can leave them as it is.

Abishekcs:
Rest of the code is looking Good.

lakshita10341:
Thanks @Abishekcs  for the review! I am converting this PR to a draft for now. From here onwards, I will stack the subsequent task branches and target this branch for merges.

Abishekcs:
Great! If there are any new code changes you need us to look at, please let us know.

lakshita10341:
@ragesoss @Abishekcs The PR tests are failing during installation of `ImageBrick` with error:
```ruby
Attempting to retrieve from the cache: imagemagick-linux-July
Downloading magick from: https://imagemagick.org/archive/binaries/magick
Error: Unexpected HTTP response: 404
```
How should I proceed with this?

ragesoss:
i'm preparing a fix for that. (in case you're curious: https://github.com/WikiEducationFoundation/WikiEduDashboard/commit/70f5478886e2712cfe46747bdfe4671b51b0470d)

ragesoss:
The fixes are live, if you rebase or push new commits, CI should work now.

lakshita10341:
@ragesoss  The PR is ready to merge. PTAL when you get time.

ragesoss:
Does this include any frontend or user-accessible changes, or is it strictly the backend portion in preparation for the stats download / presentation UI?

ragesoss (commented):
**How to read this feedback:** the notes below were produced with Claude Code during review. They are observations to weigh with your own judgment, not authoritative conclusions — use, verify, or discard as you see fit.

Reviewed locally: RuboCop clean on all changed files; the PR's worker specs pass (20 examples); a model-scope spec written during review passes (9 examples). Migrations are additive `create_table`s on new tables — safe inside a normal deploy on both deployments.

**1. Facilitator worker swallows upsert failures (worth a second look).**
In `facilitator_stat_update_worker.rb`, the `rescue StandardError` around `upsert_all` logs to `Rails.logger` and lets the job complete successfully — so a failed weekly run gets no Sidekiq retry and no Sentry alert (rescued exceptions aren't auto-captured). For a once-a-week job, a silent failure could leave stale data for 7+ days unnoticed. `SystemStatUpdateWorker` takes the opposite approach (lets exceptions propagate → retry + Sentry). Consider making them consistent, and either re-raising or `Sentry.capture_exception(e)` in the rescue.

**2. Scheduling assumption on the P&E dashboard.**
`system_stat_update` is scheduled at 12:30 UTC, described as "1 hour after DailyUpdateWorker so course caches are fresh." On outreachdashboard the daily update frequently runs longer than an hour, so this snapshot may be computed against mid-update or stale course caches. Worth confirming the offset is large enough, or keying the run off update completion rather than a fixed clock offset.

**3. Factory `wiki_stats` shape does not match worker output.**
`spec/factories/system_stats.rb` emits a `'new_editors'` key per wiki, but `SystemStatUpdateWorker#compute_wiki_stats` emits `'new_editors_with_preregistration'` (and no `'new_editors'`). Anything later built against the factory's shape (serializer, frontend) would mismatch production data. Aligning the factory to the real shape now avoids that.

**4. `upsert`/`upsert_all` omit `unique_by` and batching.**
Both work on MariaDB via `ON DUPLICATE KEY UPDATE`, so production is fine. Two considerations: the existing codebase pattern (`ArticleCourseTimeslice.bulk_update_from_acuwt`) passes explicit `unique_by`/`update_only`, which is clearer and portable; and `FacilitatorStat.upsert_all(records)` sends every all-time facilitator in a single statement — that set only grows, so batching (e.g. `each_slice`) would keep the statement bounded.

**5. Test coverage gaps.**
The PR ships worker specs but no model specs; the `latest`/`current`/`recent_months`/`for_user` scopes and `wiki_stats` serialization are only exercised indirectly. A model spec covering these was written during review and can be shared/adopted if useful. Separately, the facilitator worker's rescue/error path (finding 1) has no coverage.

**6. Minor: duplicated new-editor SQL.**
The "registered during course" and 60-day pre-window queries are repeated across both workers and within each. Extracting the shared window logic (a small service or scope, per repo conventions) would reduce drift.

None of the above blocks merging; the code works and the worker-level behavior is well tested. Items 1 and 2 are the ones most worth the author's judgment.

*Reviewed in a Claude Code session (~20 min, 2 interactions). Sage directed the review and read this comment before posting. Findings were generated by Claude Code and checked by running the suite locally — RuboCop clean, the PR's 20 worker specs green, and 9 additional model-scope specs (AI-written) green — but were not each independently re-derived by the reviewer. The model-scope spec mentioned in item 5 is AI-generated and was run, not line-by-line reviewed.*

lakshita10341:
> Does this include any frontend or user-accessible changes, or is it strictly the backend portion in preparation for the stats download / presentation UI?

The PR contains only schema changes, stats computation and cron job scheduling.

lakshita10341:
> **4. `upsert`/`upsert_all` omit `unique_by` and batching.** Both work on MariaDB via `ON DUPLICATE KEY UPDATE`, so production is fine. Two considerations: the existing codebase pattern (`ArticleCourseTimeslice.bulk_update_from_acuwt`) passes explicit `unique_by`/`update_only`, which is clearer and portable; and `FacilitatorStat.upsert_all(records)` sends every all-time facilitator in a single statement — that set only grows, so batching (e.g. `each_slice`) would keep the statement bounded.

`unique_by` is raising an `arguement error`  Since we already have unique indexes in the schema for these columns, MySQL handles the conflict resolution automatically without needing unique_by.

ragesoss (commented):
**Re-review of the commits since the 2026-07-06 review** (`4a7b5be`, `e805ba5`).

The actionable findings from last time have been addressed:

- **Silent upsert failures (1):** the rescue now re-raises after logging, so a failed weekly run gets a Sidekiq retry and a Sentry alert instead of finishing green — matching `SystemStatUpdateWorker`. Verified locally with an added spec: `perform` propagates the exception and persists no partial record.
- **Factory `wiki_stats` shape (3):** the factory now emits `new_editors_with_preregistration`, matching `SystemStatUpdateWorker#compute_wiki_stats`. No `'new_editors'` references remain.
- **Batching (4):** `upsert_all` now runs in `each_slice(100)` batches, keeping the statement bounded as the facilitator set grows. Setting `unique_by` aside was the right call — Rails' `upsert_all` doesn't accept `:unique_by` on MySQL/MariaDB, and the schema's unique index handles conflict resolution. Nothing further needed.

Still open, all non-blocking and left to your judgment:

- **Scheduling offset (2):** `system_stat_update` still runs at a fixed 12:30 UTC. On outreachdashboard the daily update often runs longer than an hour, so the snapshot may be computed against a mid-update or stale course cache. Worth confirming the offset is large enough, or keying the run off update completion rather than the clock.
- **Model coverage (5):** no model specs yet — `latest` / `current` / `for_user` / `recent_months` / `for_date_range` / `active_facilitators` and the `wiki_stats` serialize round-trip are still exercised only indirectly. The error path is now covered by the spec below (AI-generated; run, not line-by-line reviewed) — adopt or replace with your own as you see fit.
- **Duplicated new-editor SQL (6):** the 60-day pre-registration window is still repeated across both workers; extracting it would reduce drift, but it's cosmetic.

No new issues in the delta.

<details><summary>Error-path spec used to verify finding 1</summary>

```ruby
describe FacilitatorStatUpdateWorker do
  # ... course + instructor set up via factories, Features.wiki_ed? => false ...

  describe '#perform error path' do
    it 're-raises when the upsert fails so Sidekiq retries and Sentry captures' do
      allow(FacilitatorStat).to receive(:upsert_all).and_raise(
        ActiveRecord::StatementInvalid, 'boom'
      )
      expect { described_class.new.perform }
        .to raise_error(ActiveRecord::StatementInvalid, /boom/)
    end

    it 'does not persist a partial record when the upsert fails' do
      allow(FacilitatorStat).to receive(:upsert_all).and_raise(
        ActiveRecord::StatementInvalid, 'boom'
      )
      expect { described_class.new.perform rescue nil }
        .not_to change(FacilitatorStat, :count)
    end
  end
end
```
</details>

*Re-reviewed in a Claude Code session (~15 min, one directing instruction). Sage directed the re-review and read this comment before posting. Verified locally: RuboCop clean on the changed files, the PR's 20 worker specs green, and two added error-path specs (AI-generated) confirming the worker re-raises upsert failures and persists no partial record. The remaining-issue notes were not each independently re-derived.*

Issue #6878, opened 2026-05-15
Improve Lexeme support for Wikidata stats


ragesoss:
Currently, our support for processing Lexeme edits is pretty minimal. Lexeme contributions are an area of high priority for Wikidata, so the Dashboard should do a better job of providing detailed stats about Lexeme-related contributions, like senses added and edited, glosses added and edited, and similar things.

This probably starts with WikidataDiffAnalyzer.

Formasitchijoh:
Working on this

Issue #6877, opened 2026-05-15
[DESIGN] Wikidata stats should emphasize the kinds of contributions that Wikidata wants most


ragesoss:
Currently, our wikidata stats component includes a wide array of stats, but without prioritizing one kind of contribution over another.

The Wikidata community and maintainers have some types of contributions they consider more valuable / higher priority. The Dashboard's stat displays should emphasize those things.

In particular:
* Adding references is the most valuable kind of addition
* Manual edits are more valuable than automated edits, generally
* Updating and adding statements is more important than adding labels and descriptions

(Worth noting: 'manual vs automated' cuts across the existing classification we do, but we can probably implement counting manual vs automated edits alongside the other classifications.)

Propose a high-fidelity design that emphasize the sorts of edits that Wikidata wants to prioritize.

SanjanaSogimatt:
Can I pick this up? Or is anyone else working on it. I'll share my plan on this

ragesoss:
Yeah, go for it! This intersects with work that the education hub team might take on, but I think the design side of it is definitely open at this point.

SanjanaSogimatt:
_AI is involved in making this plan_

Ok based on my understanding, I just populated my test DB with random stat data to see how it looks rn and this is the plan -

At Present, [wikidata_overview_stats.jsx](app/assets/javascripts/components/common/wikidata_overview_stats.jsx) shows ~25 metrics in a flat 4-column grid with equal visual weight

**The design goals here are based on what I know -**

- Make high-value work obvious at a glance (references + statement edits first).
- De-emphasize lower-priority edits (labels, descriptions, aliases) without hiding them.
- Prepare for manual vs automated as a cross-cutting dimension (Phase 2).
- Stay consistent with existing Dashboard patterns: [OverviewStat](app/assets/javascripts/components/common/OverviewStats/overview_stat.jsx), [_stats.styl](app/assets/stylesheets/modules/_stats.styl), brand primary color, tooltip info pattern.



Apply everywhere Wikidata stats appear: course Home tab (OverviewStatsTabs) and campaign overview (Campaign).

#### Proposed information architecture

Replace the flat grid with four visual tiers:

```mermaid
flowchart TB
  Hero["Tier 1: Hero — High-impact contributions"]
  Primary["Tier 2: Primary — Statement work detail"]
  Secondary["Tier 3: Secondary — Lexical & structural (collapsed)"]
[...]

PR #6904, opened 2026-06-15
Improve Lexeme support for Wikidata stats

Formasitchijoh:
## What this PR does
This addresses PART II of issue #6878
Adds backend mapping and frontend display for the 16 new Lexeme-specific stat keys introduced in the wikidata-diff-analyzer gem , specifically form references, form qualifiers, sense references, sense qualifiers, language, and lexical category. Also wires in the related Lexeme stats that were already tracked by the analyzer (lemmas, forms, form claims, senses, glosses, sense claims) but had never been shown in the UI. Addresses part II of .

Changes:
- `STATS_CLASSIFICATION` in `UpdateWikidataStatsTimeslice` extended with 16 new analyzer keys so they are summed and stored in `CourseStat`
- `wikidata_overview_stats.jsx` expanded with stat rows for all Lexeme sub-categories (language, lexical category, lemmas, forms, form claims, form references, form qualifiers, senses, glosses, sense claims, sense references, sense qualifiers)
- i18n labels added for all new section headers
- `Metrics/ClassLength` ceiling raised to 145 in `.rubocop.yml` to accommodate the larger `STATS_CLASSIFICATION` hash

## AI usage
No AI usage

## Screenshots

Screenshot 2026-06-15 at 16 20 34

Screenshot 2026-06-15 at 16 21 01

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -30,7 +30,7 @@ Layout/LineLength:
   Max: 100
 
 Metrics/ClassLength:
-  Max: 130 # We should bring this down, ideally to the default of 100
+  Max: 145 # We should bring this down, ideally to the default of 100
   Exclude: # These are too big, and should be shrunk if feasible.
     - 'app/models/alert.rb'
     - 'app/models/alert_types/ai_edit_alert.rb'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -31,7 +31,7 @@ GIT
 
 GIT
   remote: https://github.com/WikiEducationFoundation/wikidata-diff-analyzer
-  revision: 683c88e6744f6c95cc4653dd4a0df4e32a092c65
+  revision: 21a2914a32554b4372633463a34b2801c16c3311
   specs:
     wikidata-diff-analyzer (2.2.0)
       json (~> 2.1)
@@ -242,7 +242,7 @@ GEM
       railties (>= 5.0.0)
     faker (2.23.0)
       i18n (>= 1.8.11, < 2)
-    faraday (2.14.1)
+    faraday (2.14.2)
       faraday-net_http (>= 2.0, < 3.5)
       json
       logger
@@ -253,7 +253,7 @@ GEM
       faraday (>= 1, < 3)
     faraday-multipart (1.2.0)
       multipart-post (~> 2.0)
-    faraday-net_http (3.4.2)
+    faraday-net_http (3.4.4)
       net-http (~> 0.5)
     faraday-retry (2.4.0)
       faraday (~> 2.0)
@@ -331,7 +331,7 @@ GEM
     jbuilder (2.14.1)
       actionview (>= 7.0.0)
       activesupport (>= 7.0.0)
-    json (2.19.4)
+    json (2.19.9)
     jwt (2.5.0)
     kt-paperclip (7.2.2)
       activemodel (>= 4.2.0)
--- a/app/assets/javascripts/components/common/wikidata_overview_stats.jsx
+++ b/app/assets/javascripts/components/common/wikidata_overview_stats.jsx
@@ -232,6 +232,304 @@ const WikidataOverviewStats = ({ statistics, isCourseOverview }) => {
             />
           </div>
         </div>
+        <div className="stat-display__row">
+          <h5 className="stats-label">{I18n.t('metrics.language')}</h5>
+          <div className="stat-display__value-group">
+            <OverviewStat
+              id="language-added"
+              className="stat-display__value-small"
+              stat={statistics['language added']}
+              statMsg={I18n.t('metrics.added')}
+              renderZero={true}
+            />
+            <OverviewStat
+              id="language-changed"
+              className="stat-display__value-small"
+              stat={statistics['language changed']}
+              statMsg={I18n.t('metrics.changed')}
+              renderZero={true}
+            />
+          </div>
+        </div>
+        <div className="stat-display__row">
+          <h5 className="stats-label">{I18n.t('metrics.lexical_category')}</h5>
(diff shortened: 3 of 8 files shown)


Formasitchijoh:
This will be ready for merging when this [issue](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6877) is completed, so that the lexeme stats are moved to the appropriate section of the UI

ragesoss (changes requested):
> **Note on this feedback:** The following was produced with assistance from Claude Code. It is not meant to be trusted or acted on blindly — read it, verify the parts that matter, and use or discard each point based on your own judgment.

This PR wires up the UI and key-mapping for Lexeme stats cleanly, and the half that surfaces already-tracked categories (lemmas, forms, form claims, senses, glosses, sense claims) appears to work as soon as it merges. The 16 brand-new keys, however, look like they will be inert until a dependency bump lands.

**Blocking: `Gemfile.lock` does not appear to be updated, so the 16 new keys are dead as checked out.**

The new keys (`added_language`, `added_form_references`, `added_sense_qualifiers`, `added_lexical_category`, etc.) are emitted by the gem's PR #36 ("Include Lexeme support"), merged into `wikidata-diff-analyzer` `main` as `21a2914a32` on 2026-05-22. This PR does not touch `Gemfile`/`Gemfile.lock`; the lock still pins `683c88e` (gem PR #35, 2026-04-30 — pre-Lexeme).

A runtime check against the bundle on this branch suggests none of the new keys exist yet:

```
VERSION=2.2.0
language_analyzer.rb present? false
total.rb mentions added_language?          false
total.rb mentions added_form_references?   false
total.rb mentions added_sense_qualifiers?  false
total.rb mentions added_lexical_category?  false
```

If that holds, then because every new row uses `renderZero={true}`, the six new categories (language, lexical category, form references, form qualifiers, sense references, sense qualifiers) would render a permanent `0` and nothing would be summed into `CourseStat`. Worth bumping `wikidata-diff-analyzer` in `Gemfile.lock` to `21a2914a32` or later as part of this PR (and confirming the deploy runs `bundle install`), then re-checking that real Lexeme revisions populate the new rows.

Note the key spellings themselves look correct — they match gem PR #36's output exactly (`added_form_references` → `form_references_added`, and so on) — so once the gem is bumped the mapping should line up without further change.

**Secondary: the new spec cannot detect the above.**

`'counts new lexeme stat keys in built stats'` constructs `revision.summary` with the 16 keys directly and asserts they are summed. It passes regardless of whether the analyzer ever emits those keys (it is green now, against the pre-Lexeme gem). It exercises the mapping but not the data source. A spec that runs the analyzer over a known Lexeme revision (via VCR) would tie the assertion to real analyzer output and would have surfaced the missing bump.

**Minor:**
- `Metrics/ClassLength` is raised 130→145 to fit the larger constant. This moves away from the "ideally to 100" goal noted in the config; extracting `STATS_CLASSIFICATION` to its own constant file would avoid the bump, though that is a judgment call.
- RuboCop is clean; the three ESLint warnings on the changed file are all pre-existing hardcoded-string lines, not introduced here — the new rows correctly use `I18n.t`.

*Drafted in a Claude Code session (~20 min, ~3 interactions). Sage directed the review and approved this comment before posting, but the runtime gem-bump check, the gem PR #36 cross-reference, and the linter/spec runs were performed by Claude Code and were not all independently re-verified by Sage. The findings are provisional — confirm the `Gemfile.lock` gap before treating it as settled.*

ragesoss (approved):
> **Note on this feedback:** Produced with assistance from Claude Code. Not meant to be trusted or acted on blindly — read it, verify what matters, and use or discard each point on your own judgment.

Both findings from the prior review are resolved and verified locally.

**Blocking (gem bump) — resolved.** `Gemfile.lock` now pins `wikidata-diff-analyzer` at `21a2914a32`. A fresh `bundle install` against this branch installs that exact revision, and its source contains all the new keys (`added_language`, `added_form_references`, `removed_sense_qualifiers`, `added_lexical_category`, …) — the same runtime probe that came back `false` on the old lock now resolves them. The 16 new keys are live as checked out.

**Secondary (spec couldn't detect the gap) — resolved.** The new example exercises the analyzer over a real Lexeme revision (`2414438514`) rather than a hand-built `summary`, so the assertion is now tied to actual analyzer output. Verified the recorded interaction is a genuine `wikidata.org` Lexeme diff, and the example passes against the bumped gem.

Verified on a local checkout (master merged in): `update_wikidata_stats_timeslice_spec.rb` (7 examples) and `update_course_wiki_timeslices_spec.rb` (26 examples, against the regenerated `split_timeslices` fixtures) both green; `yarn build` clean; RuboCop clean; no new ESLint offenses. Every `statistics['…']` key the JSX reads maps to a `STATS_CLASSIFICATION` value, so no row renders a permanent zero from a mapping typo.

Non-blocking notes, neither needs to hold the merge:
- The `Gemfile.lock` change also carries incidental transitive patch bumps (`faraday`, `faraday-net_http`, `json`) not mentioned in the commit message — benign, just noting they ride along.
- `Metrics/ClassLength` is still raised as a global ceiling rather than scoped. A per-file `Exclude` or extracting `STATS_CLASSIFICATION` to its own constant file would keep the global limit tight, but that remains a judgment call.

Approving.

*Re-review drafted in a Claude Code session (~20 min, ~3 interactions). Sage directed the re-review and reviewed this comment before posting. The verification — `bundle install` + gem-revision/key probe, recording the new VCR cassette live and confirming it's a real Lexeme diff, running both spec files, the build, the linters, and the JSX↔label mapping cross-check — was performed by Claude Code; Sage spot-checked but did not independently re-run every step. Findings are provisional.*

PR #6908, opened 2026-06-15
Add admin-only Manage course flags view

gabina:
## What this PR does

Adds an admin-only "Manage course flags" page at `/course_flags` that lets
administrators view and toggle four course settings that previously required
a Rails console or direct database access:

- `use_acuwt` (flag) — opt a course into the ACUWT update path
- `very_long_update` (flag) — mark a course as too slow for normal update scheduling
- `debug_updates` (flag) — enable verbose update logging for a course
- `needs_update` (field) — immediately queue a full course update

A `CourseFlagsController` with `index` / `show` / `update` actions handles
the feature entirely on the server side (no frontend JS). Access is restricted
to admins via `require_admin_permissions`. A link to the new page is added to
the admin dashboard under "Other Useful links".

## AI usage

This PR was developed entirely with Claude Code (Sonnet 4.6) under human
direction. Claude Code drafted the controller, HAML views, route additions,
model predicate methods (`use_acuwt?`, `debug_updates?`), and feature specs.
The human (Gabina) provided all decisions about what to build, directed the
approach (dedicated controller modelled after `TimesliceDurationController`,
Ruby-only, no JS), and reviewed each step before committing.

One correction was needed: an initial draft added the action to
`CoursesController`; the human redirected to a standalone controller. The
spec run required one iteration — `be_axe_clean` assertions were catching
pre-existing nav/footer violations unrelated to the new views and were
dropped. All three commits were made within a single ~15-minute session.
Commit messages and this PR description were drafted by the `/commit` and
`/prepare-pr` Claude Code skills.

## Screenshots

**01 admin dashboard** — new "Manage course flags" link visible in Other Useful links

| Before | After |
|--------|-------|
| _(did not exist on master)_ |  01_admin_dashboard|

**02 course flags index** — search form to look up a course by ID

| Before | After |
|--------|-------|
| _(did not exist on master)_ | 02_course_flags_index  |

**03 course flags show** — current flag values with update form

| Before | After |
|--------|-------|
| _(did not exist on master)_ | 03_course_flags_show|

**04 course flags updated** — redirect notice after saving

| Before | After |
|--------|-------|
| _(did not exist on master)_ | 04_course_flags_updated |

- `use_acuwt` is a new flag with no existing consumer yet — this PR only                                                                                                 makes it settable. It's used on article-course-user-wiki-timeslices branch.
 `debug_updates` is already consumed by `UpdateDebugger`                                                                                                                    
(`lib/data_cycle/update_debugger.rb`) to gate Sentry capture of per-stage                                                                                                  
update timings; this PR makes it settable without a Rails console. 
- The page looks up courses by numeric ID rather than slug.

### DIFF
--- a/app/controllers/course_flags_controller.rb
+++ b/app/controllers/course_flags_controller.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+#= Admin-only controller for viewing and setting course flags and needs_update
+class CourseFlagsController < ApplicationController
+  respond_to :html
+  before_action :require_admin_permissions
+  before_action :set_course, only: %i[show update]
+
+  MANAGEABLE_FLAGS = %i[use_acuwt very_long_update debug_updates].freeze
+
+  def index; end
+
+  def show; end
+
+  def update
+    MANAGEABLE_FLAGS.each do |flag|
+      @course.flags[flag] = params[flag] == '1'
+    end
+    @course.needs_update = params[:needs_update] == '1'
+    @course.save
+    redirect_to course_flags_show_path(course_id: @course.id),
+                notice: "Flags updated for course #{@course.slug}"
+  end
+
+  private
+
+  def set_course
+    return unless params[:course_id].present?
+    @course = Course.find(params[:course_id])
+  rescue ActiveRecord::RecordNotFound
+    redirect_to course_flags_path, flash: { error: 'Course not found' }
+  end
+end
--- a/app/controllers/course_flags_controller.rb
+++ b/app/controllers/course_flags_controller.rb
@@ -10,9 +10,12 @@ class CourseFlagsController < ApplicationController
 
   def index; end
 
-  def show; end
+  def show
+    @manageable_flags = MANAGEABLE_FLAGS
+  end
 
   def update
+    return redirect_to(course_flags_path, flash: { error: 'Course not found' }) if @course.nil?
     MANAGEABLE_FLAGS.each do |flag|
       @course.flags[flag] = params[flag] == '1'
     end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -571,6 +571,14 @@ class Course < ApplicationRecord
     flags[:very_long_update].present?
   end
 
+  def use_acuwt?
+    flags[:use_acuwt].present?
+  end
+
+  def debug_updates?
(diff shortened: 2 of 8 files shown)


ragesoss:
Three changes were pushed to this branch (commit `04c8701`) during a review pass. They're offered as fixes to consider — read, verify, and keep or revert based on your own judgment; AI-driven changes shouldn't be trusted blindly.

**1. Flag list unified to a single source.** The form in `show.html.haml` iterated an inline `%i[use_acuwt very_long_update debug_updates]` literal while `update` iterated the `MANAGEABLE_FLAGS` constant — two independent lists. Adding a flag to the constant alone would leave it without a form control, so `update` would read a missing param (`params[flag] == '1'` → `false`) and silently force that flag off on every save. `show` now exposes `MANAGEABLE_FLAGS` as `@manageable_flags` and the form iterates that, so the two can't diverge.

**2. `update` guarded against a nil course.** `set_course` returns without redirecting when `course_id` is blank, leaving `@course` nil; `update` would then call `@course.flags` and raise `NoMethodError` (500). Unreachable from the UI (the form always sends the hidden `course_id`), but `update` now redirects to the index with "Course not found", mirroring the existing not-found path. Defensive only — no spec, as there's no UI path to it.

**3. Spec gap.** The update round-trip now also toggles and asserts `debug_updates`; previously 3 of the 4 controls were exercised.

The read-only display table was left listing the flags inline — that duplication is benign (a missing row is cosmetic, not data-corrupting) and keeping it preserves the `use_acuwt?` predicate reserved for the future branch. The `require_admin_permissions` gate was reviewed and left as-is. Two earlier notes — the `%hr` block in the index view and `js: true` on the spec — were checked against `timeslice_duration` and dropped; both are conventions copied from there.

Verification before push: RuboCop clean on the changed files, `course_flags_spec` 6/6 green, controller line coverage 100%.

*Drafted and pushed in a Claude Code session (~20 min, 5 interactions). Sage directed the review via the /review-pr skill, confirmed the `require_admin_permissions` gate was intentional, and approved the pushed fixes and this comment before posting. The fixes are AI-written; the session ran `course_flags_spec` (6/6) and RuboCop (clean) locally, and Sage reviewed the diff and those results but did not independently re-derive each finding by hand.*

PR #6913, opened 2026-06-18
Merge  article-course-user-wiki-timeslices branch

gabina:
## What this PR does

Introduces a new, finest-grained timeslice layer — **ArticleCourseUserWikiTimeslice (ACUWT)** — keyed by (course, wiki, article, user, time window), and re-architects the timeslice update flow around it for courses that opt in via the `use_acuwt` flag.

**The new layer**
- Adds the `article_course_user_wiki_timeslices` table and model, storing `revision_count`, `character_sum`, `references_count`, `new_article`, `first_revision`, wikidata `stats`, `needs_update`, and `tracked` per article/user/period.
- The three existing timeslice tables — ArticleCourseTimeslice, CourseUserWikiTimeslice, CourseWikiTimeslice — are now *re-derived from* ACUWT instead of recomputed by re-fetching revisions. New `update_from_acuwt` / `bulk_update_from_acuwt` methods aggregate the stored rows upward, coordinated by a new `needs_reaggregation` flag on `course_wiki_timeslices`.

**Cheaper updates and retries**
- `ReprocessArticleCourseUserWikiTimeslices` does *targeted* retries: when a single (article, user) ACUWT row failed scoring or wikidata-stats fetching, only that pair is re-fetched, then the affected CourseWikiTimeslice is flagged for reaggregation — instead of re-fetching every revision in the period.
- ACUWT rows are written with `upsert_all` (bulk) rather than per-row `find_or_create_by` + `save`.
- The shared services (`UpdateCourseWikiTimeslices`, `UpdateTimeslicesCourseUser`, `UpdateTimeslicesScopedArticle`, `UpdateTimeslicesUntrackedArticle`) gained `if use_acuwt? … else …` branches, so the legacy path is untouched for non-flagged courses.

**Controlled rollout**
- The per-course `use_acuwt` flag routes flagged courses to a new dedicated `acuwt_update` Sidekiq queue, served by a single-concurrency `sidekiq-acuwt` systemd process isolated from the normal update queues; `SystemMetrics` monitors it.

**Bundled fix (applies to all courses)**
- `Category#article_ids` now drops diacritic-variant articles that MySQL's accent-insensitive collation was wrongly matching (e.g. "Hanói" vs "Hanoi"), by re-filtering query results in Ruby with an exact string match.

## Why

Some very active ArticleScopedPrograms — notably the 2026 "Coordinate Me" campaign events — have caused severe processing backlogs. These courses track dynamic, PetScan-based Categories, and every time the set of articles in such a Category changes, the current system must recompute *all* of the course's timeslices. In practice this meant the timeslice system's whole point — efficient *incremental* updates — was not working for this class of course: routine churn in a category forced full rework over and over, and the affected courses fell badly behind.

Storing the finest-grained data once (ACUWT) lets the higher-level caches (ArticleCourse / CourseUserWiki / CourseWiki timeslices) be rebuilt by aggregating already-stored rows instead of re-fetching every revision from MediaWiki, and lets failed fetches be retried at the granularity of a single article/user. The goal is to handle these high-churn category-based courses with far less rework — so we can catch up on the backlog and then keep up with them incrementally. The dedicated `acuwt_update` queue keeps the new path isolated while it's validated on the opted-in courses.

(The bundled `Category#article_ids` diacritic fix is part of the same effort: it tightens which articles a PetScan/category-based course actually tracks.)

## Migrations

- Create `article_course_user_wiki_timeslices` (new empty table; 3 indexes including a unique natural-key index). Fast.
- Add `needs_reaggregation` (nullable boolean, default false) to `course_wiki_timeslices` — additive, INSTANT/metadata-only on MariaDB 10.3+, no backfill.

## Testing

Extensive specs added across the new model, services, and lib changes, covering both the ACUWT and legacy code paths for the refactored shared services.


ragesoss (approved):
This is AI-assisted review feedback from a Claude Code session. It's offered as observations to read and weigh, not conclusions to act on automatically — the analysis is provisional and may be wrong. Use or discard per your own judgment.

## Summary

A deep review of the ACUWT feature-branch merge. The change is well-structured and unusually well-isolated: fully opt-in per course via `use_acuwt`, routed to a dedicated single-concurrency `sidekiq-acuwt` process, with the legacy path preserved intact behind `if @course.use_acuwt? … else …` branches in every shared service.

## Verified locally

- All affected/PR specs pass (168 examples) plus the legacy-path service/lib specs (37 examples) — no regression in the default path from the shared-service refactor.
- RuboCop clean on all 16 changed Ruby files.
- `upsert_all` correctly serializes the Array `user_ids` and Hash `stats` text columns on Rails 8.1.3 (round-tripped via a one-off check) — the bulk-write path preserves the serialized values.
- Sidekiq isolation holds: every production systemd unit selects queues with explicit `--queue` flags, so adding `acuwt_update` to the global `:queues:` list affects only the test environment and the monitoring UI; no normal worker consumes acuwt jobs.
- Both migrations are additive (new empty table; nullable boolean default-false column) — safe inside a normal deploy.

## Observations (non-blocking)

1. The `@update_sercice` → `@update_service` typo fix changes behavior for *all* courses, not just ACUWT ones: on the previous code the legacy new-user revision fetch always received `update_service: nil`, so Replica API errors there were not recorded to the update logs. The fix looks correct and beneficial; flagging it only because it lives, un-gated, inside a feature-branch merge.

2. `mw_rev_count` derived from ACUWT undercounts deleted revisions (documented in-code as a known approximation, since ACUWT stores only live revisions). Worth confirming it can't cause an ACUWT course to incorrectly conclude there's nothing new to process. ACUWT-only.

3. `ArticleCourseUserWikiTimeslice#tracked` is set by the untrack/retrack paths but read only by those same paths (for idempotency); stat aggregation filters on `articles_courses.not_tracked`, not this column. Not a bug — noting it because a reader may expect the column to gate aggregation.

4. The new table's `schema.rb` entry is dumped with `utf8mb4_general_ci` while the rest of the schema is `utf8mb4_unicode_ci` (a dev-DB default leaking into the dump). No functional impact — ACUWT has no string keys — but it may cause schema churn on the next dump from a canonically-configured database.

## Coverage

Coverage is extensive across the new model, services, and lib. The one gap found was the new-user / removed-user ACUWT branches of `UpdateTimeslicesCourseUser` (and their `CourseRevisionUpdater` / `RevisionDataManager` helpers), which had no spec. A spec covering both branches was written during the review and added to the branch: the two examples assert `needs_reaggregation` / ACUWT-row outcomes the legacy branch cannot produce, so they fail if the `use_acuwt?` branch isn't taken. That AI-written spec was run locally and passes; it has not otherwise been independently audited.

---
*Reviewed in a Claude Code session (~1 hour, ~15 interactions). Sage directed a full deep review and verified specifics locally — ran the affected-area and legacy spec suites, checked `upsert_all` serialization with a one-off spec, confirmed queue isolation from the systemd units, and traced the new-user date-comparison against the existing `timeslice_manager` idiom — then wrote and committed the coverage spec. The full comment was reviewed before posting. Other observations were spot-checked but not each independently verified.*

(Review written by Claude Code.)

ragesoss:
Gabina's next steps list:

* Run the migrations (via capistrano deploy) and update webserver. That covers default queue as well.
* Pull the changes into the sidekiq server for long queue.
* Enable the use_acuwt flag for some course.
* Ensure it gets queued.
* Set sidekiq-long-4 to consume from the new acuwt queue and restart it.

PR #6921, opened 2026-06-25
Initial UX for "fact verification" exercise

ragesoss:
This adds the initial bunch of code and features for a new fact verification exercise for Wikipedia Student Program courses on Wiki Education Dashboard. We'll continue building it out for initial use in the Fall 2026 term, but will merge an initial version so new courses can be created with the wizard and be all set for the final form of the exercise and the system for preparing and displaying claims by the time the fall term actually starts.


ragesoss:
## Fact verification exercise — screenshot gallery

These were captured by driving the actual UI in headless-browser feature specs.
The article content in the viewer screenshots is test data standing in for a real
Wikipedia article (the harvested-claim and citation machinery runs for real over it).

### 1. The new wizard option

The "Getting started with editing" panel of the research-write wizard now offers
**Fact verification (recommended)**, selected by default alongside the existing
introductory assignments.

Wizard: Fact verification option

### 2. The exercise on the timeline

When selected, a **Fact verification** exercise block lands on the course timeline.
Because it's an in-app exercise (not a sandbox handoff), its launch button is a
React-Router link that keeps the student in the SPA.

Timeline: Fact verification exercise block

### 3. Choosing an article

Opening the exercise shows the article picker: a compact grid of candidate
(article, flagged-revision) tiles drawn from the pre-harvested claim pool and
prioritized to the course (see the matching write-up below).

Article picker

### 4. Seeing the claims in context

Clicking a tile opens the article at the flagged revision, in place, with its
harvested claims highlighted. The footer reports how many claims were found and
lets the student step between them.

Article viewer with highlighted claims

### 5. Inspecting a claim and its source

Clicking a highlighted claim opens an in-viewer panel with the claim sentence, its
cited source, and a link out to that source — and the button to select it.

Claim selection panel

### 6. The selected claim

After selecting, the student lands on their taken-claim view, with the claim, its
cited source, and the handoff to complete the verification in their Wikipedia
sandbox.

Taken claim

---

*Screenshots produced in a Claude Code session at Sage's direction, by driving the
real UI in feature specs (not mockups). The viewer's article text is test data, not
a live article. Sage reviewed the images before posting. (Comment written by Claude Code.)*

ragesoss:
## How articles are matched to a course for the fact-verification exercise

When a student opens the exercise inside their course, the Dashboard shows them a
short list of articles to choose from. Those articles aren't random — they come
from a shared, pre-built "claim pool," ordered to be as relevant as possible to
that particular course. End to end:

**1. Building the claim pool (done ahead of time, by staff).**
The pool is a library of individual factual claims pulled out of edits the
Dashboard had already flagged as likely AI-generated (its existing AI edit alerts).
For each flagged edit to a mainspace Wikipedia article, the system saves every
sentence that was *added with a citation*, along with where it came from: the
article, the exact revision, the course the edit was made in, and that course's
subject. Right now this harvesting is a manual step an admin runs — it isn't on an
automatic schedule yet.

**2. Picking which articles to show a given course.**
When a student starts the exercise, the system assembles up to 12 article "tiles"
from the pool, filling them in this priority order:

1. **Same-topic courses first.** Courses built with the wizard can carry one or
   more *topic tags* — the subject-specific guides the instructor selects (e.g.
   Ecology, Medicine, Biographies). The system finds other courses that share any
   of the same topic tags and surfaces the claims harvested from those courses
   first. So a student in an ecology-tagged course tends to see claims that came
   out of other ecology-tagged courses.
2. **Same free-text subject next.** If there's still room, it adds claims whose
   stored subject matches the course's own subject field.
3. **Anything else, to fill out the list.** If there still aren't enough, it tops
   up from the rest of the pool — so the picker is never empty when the pool has
   any claims.

Within each tier, articles with the most harvested claims appear first (ties break
toward the most recent flagged revision). Each tile is one (article, flagged
revision) pair; the claims it offers are exactly the ones harvested from that
revision.

**3. What the student does.**
They open a tile, which loads the article at the flagged revision with its
harvested claims highlighted, click a highlighted claim to see its cited source,
select it, and go verify it against that source in their Wikipedia sandbox.

### Intentional simplifications in this initial version

Noted for context — these are deliberate, not defects, and line up with the plan to
refine the student-facing selection before fall:

- **Claims aren't reserved.** Two students can be offered, or pick, the same claim;
  the only constraint is one active pick per student per course.
- **No "exclude my own work" filtering.** The list isn't yet filtered to skip
  claims a student added themselves or articles they're already working on.
- **Empty state is minimal.** When there are no candidate articles, the picker just
  says so; letting the student pick their own article in that case is marked as
  future work.
- **Harvesting is manual**, not yet scheduled.

*Description of current code behavior, written in a Claude Code session at Sage's
direction and reviewed by him before posting. (Comment written by Claude Code.)*

ragesoss:
## Review — Initial UX for "fact verification" exercise

_How to use this: AI-assisted review notes. Read them, sanity-check anything that
matters, and use or discard on your own judgment — they are observations to verify,
not conclusions to trust. The one finding verified directly this session is called
out as such; the rest came from automated review and should be read critically._

> **Update (edited):** the LLM subsystem this review originally flagged — the
> Anthropic adapter, the verdict/source-fetch verification path, the offline eval
> harness, and the `anthropic` gem — has since been cut from the branch, since
> nothing in the running app invoked it. The two findings about it (LLM-adapter
> hardening and the `SourceFetcher` SSRF shape) are therefore removed below.

Scope of this review is calibrated to the stated intent: lock in the schema,
assignment-wizard, and timeline integration now, with the student-facing exercise
UX still in flight.

### Migrations — safe to ship in a normal deploy

All four migrations are purely additive against brand-new, empty-in-production
tables (`verification_claims`, `verification_claim_assignments`): two `create_table`s
plus two `add_column`s (nullable `alert_id` + index, nullable `mw_rev_timestamp`).
No destructive ops, no inline backfill, no changes to existing columns. The
`NOT NULL`s are on columns of empty tables, and `add_index :verification_claims,
:subject` on a `string(255)` is well within InnoDB's prefix limit (other 255-char
columns are already indexed in the schema). `db/schema.rb` matches. The feature's
data only appears after the manual `rake claim_verification:harvest_pool`, so the
deploy itself ships empty tables — exactly the merge-now intent.

### Merge-now surface looks solid

- Routes order the deep `verify_claim/*` JSON paths ahead of the `courses#show`
  catch-all and the SPA fall-through is intentional.
- `TrainingModule#exercise_path` + `TrainingModuleDueDateManager#exercise_url` are
  clean additive accessors; both return `nil` when `exercise_path` is unset, so no
  behavior change for existing modules.
- Strong test coverage on this surface: both models, the harvest pipeline
  (`HarvestClaimPool` / `HarvestAiEditAlertClaims` / `HarvestRevisionClaims`),
  extraction, the request + feature flow, and the LLM client (WebMock-stubbed, so
  no real API calls in CI). React components are untested, consistent with the UX
  being in progress.

### Worth a look

- **Enrollment check on the exercise endpoints (verified directly).**
  `ClaimVerificationExercisesController` gates `state` / `annotated_article` /
  `take` on `require_signed_in` only; `course_from_slug` checks the course exists
  but not that `current_user` is enrolled. So any signed-in user can POST
  `/courses/<any-slug>/verify_claim/take` and create a `VerificationClaimAssignment`
  for themselves in a course they aren't in. Blast radius is low — assignments store
  no student work and aren't exclusive — but it's an unauthenticated-to-course
  write, and it's cheap to close before instructors create these pages this fall.
  (Confirmed by reading the controller + the request spec, which only guards on
  sign-in.)
- **`HarvestClaimPool` `limit:` is silently dropped** — `find_each` re-batches by
  PK and ignores the `@limit`/order; also `scope.count` + `find_each` runs the
  query twice. Cosmetic for the unlimited admin run.
- **ESLint:** the new claim-verification JSX (`claim_verification_admin.jsx`,
  `TakenClaim.jsx`) adds several `i18next/no-literal-string` warnings. Build still
  passes; consistent with UX-in-flight, but those strings are user-facing and will
  want localizing.
- **Dead/unwired, presumably intentional:** `HarvestCourseClaims` is referenced
  only by its own spec (the live pipeline uses the AiEditAlert harvest), and
  `VerificationClaim.student_added` is defined but unused. Worth a one-line comment
  or a note so it reads as deliberate.

### Bottom line

Schema, wizard, and timeline integration are in good shape to lock in now. The one
item worth an explicit decision before merge is the enrollment check on the
exercise endpoints — close it or consciously accept it. Everything else is either
cosmetic or correctly deferred with the rest of the student UX.

_Drafted in a Claude Code session at Sage's direction. Migration shapes, the
enrollment-check finding, and the wizard/timeline/claim-selection flows were
verified locally (ran the existing and new feature specs headless, read the
controller and migrations). The remaining notes (the dropped `limit:`, the ESLint
warnings, the unwired `HarvestCourseClaims`/`student_added`) came from automated
code-reading and were spot-checked, not exhaustively reproduced. (Review written by
Claude Code.)_

ragesoss:
## Re-review — Initial UX for "fact verification" exercise

_How to use this: AI-assisted review notes, produced by re-reviewing the branch
from scratch rather than trusting the earlier review. Read them, sanity-check
anything that matters, and use or ignore on your own judgment — they are
observations to verify, not conclusions to trust. Items reproduced locally this
session are called out as such; the rest came from automated code-reading and
were not all individually reproduced._

Baseline is clean: `yarn build`, RuboCop (`--force-exclusion`), and ESLint all
pass on the changed files, and the PR's own non-VCR specs run green (62 examples,
0 failures). The four items from the prior review — the enrollment gate on
`take`, the dropped `limit:`, the ESLint i18n warnings, and the dead
`HarvestCourseClaims`/`student_added` — are all resolved.

### Substantive

- **Sentence segmenter truncates claims at internal periods** —
  `lib/claim_verification/sentence_segmenter.rb:18`. `SENTENCE_RE` treats any
  `.`/`!`/`?` as a sentence end, so the citation binds to the trailing fragment
  and the stored/displayed claim loses everything before the last internal
  period. Reproduced by running the segmenter directly:
  - "The population grew from 2.5 billion in 1950 to 8 billion today.[1]" → claim
    becomes "5 billion in 1950 to 8 billion today."
  - "He lived in the U.S. for a decade.[1]" → claim becomes "for a decade."
  - "George W. Bush served two terms.[1]" → claim becomes "Bush served two terms."

  Clean sentences are fine; this hits any cited sentence with an internal period
  (abbreviation, initial, decimal, "e.g.", "U.S.") before the citation. `sentence`
  is the hero text in the taken-claim view, the selection panel, and the
  in-article highlight, so the fragment shows in all three (the full paragraph is
  still available via `context`). There is no `sentence_segmenter_spec` and the
  extractor spec only uses clean single-clause sentences, so this path is
  untested. Most impactful item given the segmenter feeds every harvested claim.

- **`AnnotateRevisionClaims#highlight` can crash on a ref id with an apostrophe** —
  `app/services/annotate_revision_claims.rb:52`:
  `doc.css("sup.reference a[href='##{ref_id}']")`. Confirmed locally that Nokogiri
  raises `Nokogiri::CSS::SyntaxError` when `ref_id` contains a single quote (e.g.
  `cite_note-Smith's_book-3`). Named refs like `<ref name="O'Brien 2020">` are
  common; whether the rendered `cite_note` id actually preserves the apostrophe
  depends on MediaWiki's HTML5 id-encoding, which was not verified against live
  output — worth a quick check. Either way it's a latent crash from interpolating
  an external-derived string into a quote-delimited selector (the
  `annotated_article` endpoint would 500 for that article). The sibling
  `ClaimCitationExtractor` is safe — it uses only literal-prefix selectors.

- **No unique index backing pool idempotency** — `verification_claims` has only a
  non-unique `[wiki_id, mw_rev_id]` index; dedup rests on `find_or_initialize_by`
  in `HarvestRevisionClaims#store`. The worker lock plus the incremental
  `NOT EXISTS` anti-join make real-world duplication unlikely, but a `rake` run
  (which bypasses the lock) concurrent with a worker, or a `full_rescan`, can
  double-insert. A `TEXT` sentence column can't be uniquely indexed without a
  prefix/hash, so this is "document the serial-only guarantee or add a hashed key,"
  not urgent.

### UX

- **Take failures surface in the page-top notification banner, behind the open
  viewer modal** (`useClaimHighlighting.js:163-172`) — both the 403 and the
  generic-error paths dispatch to `<Notifications />` in the course navbar, which
  the modal covers, so a failed take may go unseen. Surfacing the error inside the
  selection panel would be more reliable.
- **Annotated-article fetch has no stale-guard** (`useClaimHighlighting.js:55-73`)
  — toggling flagged↔current quickly can let an in-flight response paint
  highlights onto the wrong revision. Each tile is its own viewer instance, so the
  only race window is the in-viewer revision toggle. Standard `ignore`-flag
  cleanup fixes it.
- **"Choose a different claim" is a one-way door** — once on the picker, there's
  no affordance back to the current claim short of taking a new one or reloading.
- **Deep-link + already-taken-claim** — opening a `?showArticle=` deep link while
  holding a claim, then closing the viewer, leaves the student on the picker (the
  param is stripped via `history.replaceState` without re-rendering the parent)
  rather than their taken-claim view, until reload.
- **Claim-not-found `take` is a silent no-op** — a not-found claim returns 200 +
  `{assignment: null}` and bounces back to the picker with no message. Only
  triggers if a claim is deleted between render and take; not a crash.

### a11y

The exercise a11y is strong — claim markers are real `role="button" tabindex="0"`
with `aria-label` and Enter/Space + `preventDefault`, focus returns to the
originating claim on close, the legend is `aria-live`, and the highlight uses
non-color affordances. One gap: `ClaimSelectionPanel` is `role="dialog"` but isn't
a focus trap and has no Escape-to-close (`×` only).

### Nits

- `ArticleViewerShell.jsx:110-112` comment says the claim exercise uses
  `showPermalink=false`; it actually runs with the default `true` (which is what
  makes the deep link work).
- Double space in the new wizard blurb: "accurate&nbsp;&nbsp;for themselves".
- `absolutize_links` (`annotate_revision_claims.rb:80-83`) uses `a[href^="/"]`,
  which also matches protocol-relative `//host` URLs and would double-prefix them;
  a `:not([href^="//"])` guard is cheap insurance.

### To confirm (not a defect)

The default research-write timeline is restructured (`content.yml`): "Evaluate
Wikipedia" and "Improving Representation" move from week 2 to week 3, and a
default-on "Fact verification" block is added to week 2 — affecting every new
research-write course, not only opt-ins (the two moves are unconditional).
`course_creation_spec` is updated to match. Flagging only so it's a conscious
call.

_Drafted in a Claude Code session at Sage's direction (~30 min, ~4 interactions).
The segmenter truncation and the Nokogiri apostrophe crash were reproduced locally
this session; build, lint, and the PR's specs were run; the schema, routes, and
diff scope were checked directly. The remaining notes came from automated
code-reading and were not all individually reproduced. Sage reviewed this comment
before posting. (Comment written by Claude Code.)_

ragesoss:
## Follow-up — review items addressed

The items from the review notes above have been addressed in the commits since
that review (now at the PR head). Same caveat as before: these are AI-assisted
notes on AI-assisted changes — worth reading and verifying, not trusting blindly.

- **Apostrophe crash in `AnnotateRevisionClaims` (the "worth a look" #2).** Fixed
  in `85253e2`. `#highlight` no longer interpolates the ref id into a CSS
  selector (a `cite_note` id keeps an apostrophe from a named ref like
  `<ref name="O'Brien 2020">`, which broke the quoted selector with
  `Nokogiri::CSS::SyntaxError` and 500'd the `annotated_article` endpoint); it
  matches the href in Ruby instead. The reachability was confirmed against live
  parser output, and a regression spec was added (verified to fail on the
  pre-fix code).

- **Sentence-segmenter over-splitting.** The plain "split on `.!?`" truncated any
  cited sentence with an internal period — "He lived in the U.S. for a decade."
  was harvested as "for a decade.". `0b7478e` pinned that behavior in a spec;
  `09a00e7` then replaces the regex with `pragmatic_segmenter` (a corpus-tuned
  boundary detector) and turns those expectations green. A reference marker that
  the segmenter pushes to the start of the next sentence is folded back onto the
  sentence it trailed. Residual ambiguities remain and are documented in the
  spec, not solved: an abbreviation/acronym at a true sentence end before a
  capitalized sentence under-splits (merges, both citations still attach), and
  "Ph.D." normalizes to "Ph. D." (consistent across harvest and annotate, so
  matching still holds). The full claim-verification Ruby suite stays green.

- **Review nits.** `21ac6b4`: narrowed `absolutize_links` to skip
  protocol-relative `//host` URLs (would otherwise be double-prefixed); corrected
  a stale `ArticleViewerShell` comment; removed a double space in the wizard
  blurb.

- **Student-flow UX.** `1a3437c`: take failures now show inside the selection
  panel (the page-level notification banner sits behind the viewer modal, so they
  were easy to miss); Escape closes the panel; and the annotated-article fetch
  drops a stale response that arrives after the revision is toggled. The panel is
  left intentionally non-modal (no focus trap) so the article behind stays
  reachable for picking another claim.

Not covered here: the missing DB uniqueness index on `verification_claims` (the
`find_or_initialize_by` dedup is process-serial only) and the deep-link +
already-taken-claim edge case were left as-is for now.

_Implemented in the same Claude Code session as the review above, at Sage's
direction. The apostrophe crash and the segmenter truncation were each reproduced
and then re-verified by spec; the segmenter change keeps the full
claim-verification Ruby suite green; the JS changes pass ESLint and build, but the
end-to-end student flow was not run (its VCR cassettes aren't committed). Sage
reviewed this comment before posting. (Comment written by Claude Code.)_

ragesoss:
This feature is still in an incomplete state, but the assignment design wizard portion is ready to go, so I'm going to deploy this now and finish out the actual exercise user experience before it goes into action in the Fall term. We'll keep an eye out for summer courses in the meantime and manually remove the exercise from any that come in.

The UX for harvesting claims, displaying articles with claims for students to browse, viewing an individual article and the claims in it, and selecting a claim from there, are basically ready. We still need to figure out what exactly the exercise instructions will be, and potentially replace the sandbox launcher link with an in-Dashboard UI for submitting a structured response to the claim verification exercise (and how to surface that for instructors, if that is part of it as well).

PR #6922, opened 2026-06-26
Batch timeslice deletion and consolidate cleaner helpers

gabina:
## What this PR does

Makes timeslice deletion memory-safe and batched, and tidies up the two cleaner
classes that perform it. This is groundwork for upcoming work to bound the size
of the `article_course_user_wiki_timeslices` (ACUWT) table by purging timeslices
for old courses — but this PR contains **no behavior change** on its own.

Specifically:

- Adds `delete_in_batches`, which deletes straight from an
  `ActiveRecord::Relation` in primary-key batches (`in_batches(of: 5000).delete_all`).
  The previous helpers required callers to `pluck(:id)` the entire result set into
  Ruby before deleting, which does not scale for the courses with millions of
  timeslice rows that motivated this work. The new path never materializes the id
  list, and each batch is its own statement so InnoDB purge and replication can
  keep up.
- Adds `TimesliceCleaner#delete_all_timeslices_for_course`, a single entry point
  that deletes a course's rows across all four timeslice tables — including ACUWT,
  which the existing `course.destroy` cascade never touched (there is no
  association for it on `Course`, so those rows are currently orphaned on course
  deletion). Nothing calls this entry point yet; the purge job and the
  delete-course refactor that will use it are deliberately left for a follow-up PR.
- Routes every existing relation-based deletion in `TimesliceCleaner` and
  `ArticlesCoursesCleaner` through `delete_in_batches`, dropping the `pluck`/empty
  guards, and removes the now-dead `delete_*_timeslice_ids` helpers.
- Changes `get_article_course_timeslices_for_users` to return a single relation
  (OR-combining the per-user scopes) instead of a flattened Array, so the reset
  path no longer needs to pluck ids at all.
- Renames `ArticlesCoursesCleanerTimeslice` to `ArticlesCoursesCleaner` (the
  legacy class the suffix disambiguated from no longer exists) and removes the
  stale rename TODO.
- Extracts the shared `delete_in_batches` + batch-size constant into a
  `CleanerHelper` mixin, included by both cleaner classes, following the
  codebase's existing `*Helper` module convention.

## AI usage

This PR was developed with Claude Code (Opus 4.8) under interactive human
direction. The work began as a design discussion about how to bound ACUWT table
growth, and narrowed to this refactor as the first, self-contained step. Claude
Code drafted the code and the commit messages; the human set the direction, made
the design decisions (delete-don't-rebuild strategy, a time-based purge policy
for the future job, putting the deletion in `TimesliceCleaner`, following the
`*Helper` mixin convention), and corrected Claude on several points along the way
— including a mistaken claim that the timeslice tables hold the displayed stats
(they don't; the cache tables `courses` / `courses_users` / `articles_courses`
do), and an unnecessary relocation of a method that muddied a diff. Each commit
body carries a `## Process` section with the specifics.

Scale of effort: 6 commits in a single focused session, with iteration and review
between each step. RuboCop and the affected specs were run green after every change.

## Screenshots

No UI changes (backend refactor only).

- **No behavior change and nothing wired up yet.** This PR is intentionally just
  the deletion primitive and cleanup. The pieces that give it user-visible effect
  — a time-based purge job for archived `use_acuwt?` courses, removing
  `dependent: :destroy` from the three timeslice associations on `Course`, and
  splitting `DeleteCourseWorker` into delete-timeslices-then-`course.destroy` —
  are planned for a follow-up PR, along with specs for
  `delete_all_timeslices_for_course`.
- **Pre-existing orphan bug surfaced, not fixed here.** `course.destroy` does not
  delete ACUWT rows today (no association on `Course`). `delete_all_timeslices_for_course`
  is built to close that gap, but the delete-course path is not yet rewired to use
  it, so the orphaning still happens until the follow-up PR.
- `in_batches` cursors over the table's full primary-key range filtering by
  `course_id`, so batch *count* scales with the table's id span rather than the
  course's row count. This is fine for whole-course deletes (the intended use) and
  never loads records, but is worth noting.

### DIFF
--- a/app/services/update_timeslices_course_date.rb
+++ b/app/services/update_timeslices_course_date.rb
@@ -2,7 +2,7 @@
 
 require_dependency "#{Rails.root}/lib/timeslice_manager"
 require_dependency "#{Rails.root}/lib/timeslice_cleaner"
-require_dependency "#{Rails.root}/lib/articles_courses_cleaner_timeslice"
+require_dependency "#{Rails.root}/lib/articles_courses_cleaner"
 
 class UpdateTimeslicesCourseDate
   def initialize(course)
@@ -88,7 +88,7 @@ class UpdateTimeslicesCourseDate
     @timeslice_cleaner.delete_course_user_wiki_timeslices_prior_to_start_date
 
     # Delete articles courses
-    ArticlesCoursesCleanerTimeslice.clean_articles_courses_prior_to_course_start(@course)
+    ArticlesCoursesCleaner.clean_articles_courses_prior_to_course_start(@course)
   end
 
   def remove_timeslices_after_end_date
@@ -102,7 +102,7 @@ class UpdateTimeslicesCourseDate
     @timeslice_cleaner.delete_course_user_wiki_timeslices_after_end_date
 
     # Delete articles courses
-    ArticlesCoursesCleanerTimeslice.clean_articles_courses_after_course_end(@course)
+    ArticlesCoursesCleaner.clean_articles_courses_after_course_end(@course)
   end
 
   def add_wiki_timeslices_from_new_start_date(wiki)
--- a/app/services/update_timeslices_course_user.rb
+++ b/app/services/update_timeslices_course_user.rb
@@ -144,13 +144,12 @@ class UpdateTimeslicesCourseUser
     ArticlesCoursesCleanerTimeslice.clean_articles_courses_for_user_ids(@course, user_ids)
   end
 
-  # Returns ArticleCourseTimeslice records that have edits from users
+  # Returns an ArticleCourseTimeslice relation with edits from the given users
   def get_article_course_timeslices_for_users(user_ids)
-    timeslices = user_ids.map do |user_id|
-      ArticleCourseTimeslice.search_by_course_and_user(@course, user_id)
-    end
+    return ArticleCourseTimeslice.none if user_ids.empty?
 
     # These are the ArticleCourseTimeslice records that were updated by users
-    timeslices.flatten
+    user_ids.map { |user_id| ArticleCourseTimeslice.search_by_course_and_user(@course, user_id) }
+            .reduce(:or)
   end
 end
--- a/app/services/update_timeslices_course_user.rb
+++ b/app/services/update_timeslices_course_user.rb
@@ -2,7 +2,7 @@
 
 require_dependency "#{Rails.root}/lib/timeslice_cleaner"
 require_dependency "#{Rails.root}/lib/timeslice_manager"
-require_dependency "#{Rails.root}/lib/articles_courses_cleaner_timeslice"
+require_dependency "#{Rails.root}/lib/articles_courses_cleaner"
 require_dependency "#{Rails.root}/lib/revision_data_manager"
 require_dependency "#{Rails.root}/lib/course_revision_updater"
 
@@ -131,7 +131,7 @@ class UpdateTimeslicesCourseUser
       @article_course_timeslices_for_users
     )
     # Delete articles courses that were updated only for removed users
-    ArticlesCoursesCleanerTimeslice.clean_articles_courses_for_user_ids(@course, user_ids)
+    ArticlesCoursesCleaner.clean_articles_courses_for_user_ids(@course, user_ids)
   end
 
   def remove_courses_users_acuwt(user_ids)
@@ -141,7 +141,7 @@ class UpdateTimeslicesCourseUser
     # Delete ACUWT records for removed users
     @timeslice_cleaner.delete_acuwt_for_deleted_course_users(user_ids)
     # Delete articles courses that were updated only for removed users
-    ArticlesCoursesCleanerTimeslice.clean_articles_courses_for_user_ids(@course, user_ids)
+    ArticlesCoursesCleaner.clean_articles_courses_for_user_ids(@course, user_ids)
   end
 
   # Returns an ArticleCourseTimeslice relation with edits from the given users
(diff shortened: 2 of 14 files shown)


gabina:
@ragesoss this is ready for review. It's the first step to enable the automatic timeslices rows deletion strategy.

The idea is to have a Purge job that finds courses ended > 6 months ago and deletes their timeslices via `delete_all_timeslices_for_course`. I guess that one job could flag the courses as purgable, and another job could pick the oldest-end-date first and run the delete queries. Once a course is set as purgable, the course update schedule gets disabled too (to avoid a full reprocessing).

In addition, the plan involves splitting the existing course deletion process (`DeleteCourseWorker`) into two steps:
1. Delete the timeslices through `delete_all_timeslices_for_course`. 
2. Call `course.destroy`

This is because trying to delete such a huge number of timeslices rows through `course.destroy` may not be performant at all. I'll delete the existing `dependent: :destroy` clauses from the timeslices associations on Course (ACT, CUWT, CWT), so `course.destroy` doesn't include them. This means that if we run `course.destroy` in the console, we could leave orphan timeslices rows, but I think that's not a real problem since I believe we never run those things in production.

ragesoss (approved):
This review was assisted by Claude Code. AI-generated review feedback isn't meant to be trusted automatically or acted on blindly — it's offered as something to read, understand, and use or set aside based on your own judgment.

Reviewed as a no-behavior-change refactor. Verified locally on the branch merged with current master:

- RuboCop clean on all changed Ruby files.
- The cleaner specs plus the consuming service/lib specs all pass: `timeslice_cleaner`, `articles_courses_cleaner`, `duplicate_article_deleter`, the five `update_timeslices_*` services, `article_namespaces_manager`, `article_status_manager_timeslice` (76 examples, 0 failures).
- Repo-wide sweep confirms no remaining references to the old `ArticlesCoursesCleanerTimeslice` name, the old require path, or the four removed `delete_*_timeslice_ids` helpers — the rename and removal are complete.

A few correctness points that were traced through and appear sound:

- Every caller of `reset_timeslices_that_need_update_from_article_timeslices` now passes an `ActiveRecord::Relation` (required since the method reaches `in_batches` via `delete_in_batches`): the two `update_timeslices_untracked_article` callers, `mark_as_needs_update`, and the OR-combined relation from `get_article_course_timeslices_for_users`.
- `get_article_course_timeslices_for_users` building its result with `.reduce(:or)` is valid because `search_by_course_and_user` returns structurally identical relations. As a side benefit, the OR query dedups timeslices whose `user_ids` lists more than one removed user, where the previous `flatten` produced (harmless) duplicates.

### Coverage note

The only new public method, `TimesliceCleaner#delete_all_timeslices_for_course`, has no committed spec (the PR defers it to the follow-up). The following exploratory spec was written and run during review to exercise it — it confirms deletion across all four timeslice tables including ACUWT, course-scoping, and the empty-course no-op. It is AI-generated code; it was run locally (passes) and read, but is offered for adoption in the follow-up rather than as a required change:

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/timeslice_cleaner"

describe TimesliceCleaner do
  let(:course) { create(:course, start: '2024-01-01', end: '2024-04-20') }
  let(:other_course) do
    create(:course, slug: 'other/course', start: '2024-01-01', end: '2024-04-20')
  end
  let(:article) { create(:article, wiki_id: enwiki.id) }
  let(:cleaner) { described_class.new(course) }

  before do
    create(:article_course_user_wiki_timeslice, course:, article_id: article.id,
                                                user_id: 1, wiki: enwiki)
    create(:article_course_timeslice, course:, article: article)
    create(:course_user_wiki_timeslice, course:, user_id: 1, wiki: enwiki)
    create(:course_wiki_timeslice, course:, wiki: enwiki)

    create(:article_course_user_wiki_timeslice, course: other_course, article_id: article.id,
                                                user_id: 1, wiki: enwiki)
    create(:article_course_timeslice, course: other_course, article: article)
    create(:course_user_wiki_timeslice, course: other_course, user_id: 1, wiki: enwiki)
    create(:course_wiki_timeslice, course: other_course, wiki: enwiki)
  end

  it 'deletes all four timeslice types for the course and nothing for other courses' do
    cleaner.delete_all_timeslices_for_course

    expect(ArticleCourseUserWikiTimeslice.where(course:).count).to eq(0)
    expect(ArticleCourseTimeslice.where(course:).count).to eq(0)
    expect(CourseUserWikiTimeslice.where(course:).count).to eq(0)
    expect(CourseWikiTimeslice.where(course:).count).to eq(0)

    expect(ArticleCourseUserWikiTimeslice.where(course: other_course).count).to eq(1)
    expect(ArticleCourseTimeslice.where(course: other_course).count).to eq(1)
    expect(CourseUserWikiTimeslice.where(course: other_course).count).to eq(1)
    expect(CourseWikiTimeslice.where(course: other_course).count).to eq(1)
  end

  it 'is a no-op when the course has no timeslices' do
    expect { cleaner.delete_all_timeslices_for_course }.not_to raise_error
  end
end
```

### Minor

The "Do this to avoid running the query twice" comment at `update_timeslices_course_user.rb:127` is now slightly stale: `@article_course_timeslices_for_users` is a lazy relation re-evaluated a few times inside `reset_…` rather than a loaded set. Net cost is comparable and it's the intended batched-delete tradeoff, so no change needed — just flagging the comment.

*Reviewed in a Claude Code session (~15 min). Sage initiated the review, chose the approval path, and read the summary, but did not independently re-run the specs or re-trace the call paths — RuboCop, the spec runs, the reference sweep, the call-path tracing, and the exploratory spec were performed by Claude Code and reviewed at the summary level. The exploratory spec above is AI-generated and was run locally (passing).*

PR #6923, opened 2026-06-26
Delete course timeslices in batches before destroying the course

gabina:
## What this PR does

Splits complete course deletion into two explicit steps so a course's timeslice
tables are deleted in batches instead of through the ActiveRecord `destroy`
cascade, which does not scale for courses with millions of timeslice rows.

- Removes `dependent: :destroy` from the three timeslice associations on `Course`
  (`article_course_timeslices`, `course_user_wiki_timeslices`,
  `course_wiki_timeslices`). The cascade destroys rows one instantiated record at
  a time.
- `DeleteCourseWorker` now calls
  `TimesliceCleaner#delete_all_timeslices_for_course` (batched, all four timeslice
  tables) **before** `course.destroy`. Order matters — with the cascade gone,
  destroying first would orphan every timeslice row.
- This also fixes a pre-existing bug: `course.destroy` never deleted
  `ArticleCourseUserWikiTimeslice` (ACUWT) rows (there is no association for them
  on `Course`), so they were orphaned on every course deletion. The new path
  cleans them up.
- Adds a `DeleteCourseWorker` spec (there was none) asserting the course is
  destroyed, all four timeslice tables are cleared, and other courses' timeslices
  are left intact.

**This PR is stacked on #6922** (which adds `delete_in_batches` /
`delete_all_timeslices_for_course` and the cleaner refactors). Until #6922
merges, the diff here shows all of its commits too; the substantive changes in
*this* PR are the last two commits (the `Course` / `DeleteCourseWorker` change and
its spec). Once #6922 merges, this diff collapses to just those.

## AI usage

Developed with Claude Code (Opus 4.8) under interactive human direction, as part
of a larger effort to bound the growth of the ACUWT timeslice table. Claude Code
drafted the code and the commit messages and wrote the spec; the human set the
direction and made the decisions — including the delete-then-destroy ordering,
declining a `before_destroy` orphan guard for now, and structuring the work as a
stacked PR on top of #6922. Before changing the associations, Claude checked that
`DeleteCourseWorker` is the only app/lib caller of `course.destroy` and that no
spec relied on the cascade removing timeslices.

Scale of effort: the two commits unique to this PR were made in a single focused
session, with the spec verified green and RuboCop clean.

## Screenshots

No UI changes (backend only).

- **Depends on #6922** and should merge after it. The deletion primitive it calls
  (`delete_all_timeslices_for_course`) lives in that PR.
- **No `before_destroy` guard.** If `course.destroy` is ever called directly
  (rails console, future code) instead of through `DeleteCourseWorker`, timeslices
  will orphan silently — there is no DB-level foreign key to catch it. A cheap
  existence-check guard was considered and intentionally deferred.
- This is one step of a larger plan; the actual purge job for old courses (and
  blocking re-updates on purged courses) is planned as a separate follow-up PR.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -114,9 +114,14 @@ class Course < ApplicationRecord
 
   has_many :course_wiki_namespaces, class_name: 'CourseWikiNamespaces', through: :courses_wikis
 
-  has_many :article_course_timeslices, dependent: :destroy
-  has_many :course_user_wiki_timeslices, dependent: :destroy
-  has_many :course_wiki_timeslices, dependent: :destroy
+  # Timeslices are intentionally NOT dependent: :destroy. They can number in the
+  # millions for large courses, so deleting them through the destroy cascade (one
+  # instantiated record at a time) is prohibitively slow. DeleteCourseWorker
+  # deletes them in batches via TimesliceCleaner#delete_all_timeslices_for_course
+  # before destroying the course.
+  has_many :article_course_timeslices
+  has_many :course_user_wiki_timeslices
+  has_many :course_wiki_timeslices
 
   has_many :sandboxes, lambda {
     distinct.sandbox
--- a/app/services/update_timeslices_course_date.rb
+++ b/app/services/update_timeslices_course_date.rb
@@ -2,7 +2,7 @@
 
 require_dependency "#{Rails.root}/lib/timeslice_manager"
 require_dependency "#{Rails.root}/lib/timeslice_cleaner"
-require_dependency "#{Rails.root}/lib/articles_courses_cleaner_timeslice"
+require_dependency "#{Rails.root}/lib/articles_courses_cleaner"
 
 class UpdateTimeslicesCourseDate
   def initialize(course)
@@ -88,7 +88,7 @@ class UpdateTimeslicesCourseDate
     @timeslice_cleaner.delete_course_user_wiki_timeslices_prior_to_start_date
 
     # Delete articles courses
-    ArticlesCoursesCleanerTimeslice.clean_articles_courses_prior_to_course_start(@course)
+    ArticlesCoursesCleaner.clean_articles_courses_prior_to_course_start(@course)
   end
 
   def remove_timeslices_after_end_date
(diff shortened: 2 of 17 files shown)


ragesoss (approved):
This review was assisted by Claude Code. AI-generated review feedback isn't meant to be trusted automatically or acted on blindly — it's offered as something to read, understand, and use or set aside based on your own judgment.

Reviewed the delta unique to this PR (the `Course` association change, the `DeleteCourseWorker` change, and the new worker spec) against master with #6922 in place.

Verified locally:

- RuboCop clean on the three changed files.
- The new `delete_course_worker_spec` passes, along with `course_spec` and the two specs that call `course.destroy` directly (`survey_notification_spec`, `course_data_update_worker_spec`) — 103 examples green across the affected areas, no regressions.
- `DeleteCourseWorker` is the only caller of `course.destroy` in app/lib (the `delete_from_campaign` controller destroys `CampaignsCourses` join records, not courses; nothing in `lib/tasks` or docs destroys a course). So removing the cascade affects no other deletion path.
- There are no DB foreign keys on the timeslice tables and no `before_destroy`/`after_destroy` callbacks on `Course`, so removing `dependent: :destroy` interacts with nothing beyond the explicit deletion the worker now performs.
- Confirmed `Course` has no `article_course_user_wiki_timeslices` association, so ACUWT rows were indeed orphaned on the old cascade path; `delete_all_timeslices_for_course` now cleans them via its own course-scoped query.
- No spec relied on the cascade removing timeslices.

One point worth recording (not a blocker): timeslice deletion now runs as a series of batched transactions separate from `course.destroy`, so the operation is no longer atomic with the destroy the way the cascade was. If `course.destroy` raised after the timeslices were deleted, the course would persist with its timeslices already gone. This is inherent to the batched approach, and the related direct-`course.destroy` orphaning case is already noted in the PR's open questions with a deferred `before_destroy` guard.

*Reviewed in a Claude Code session (~10 min). Sage initiated the review, chose to approve, and read the summary, but did not independently re-run the specs or re-trace the call paths — the RuboCop run, spec runs, `course.destroy` caller search, foreign-key/callback checks, and association verification were performed by Claude Code and reviewed at the summary level.*

PR #6924, opened 2026-06-27
[WIP] Flag old courses as purgeable via a weekly job

gabina:
## What this PR does

Adds the first half of the timeslice-purge pipeline: a weekly job that flags old,
finished courses as **purgeable**, so a later PR can delete their timeslices and
keep the timeslice tables from growing without bound (ACUWT reached ~7–10 GB from
only a few very large courses). This PR only sets the flag — nothing is deleted yet.

A course is flagged `purgeable` when **all** of these hold:
1. it ended more than 6 months ago;
2. it was tracked in the timeslice system (has at least one course wiki timeslice)
   — this excludes legacy courses that never used timeslices and have nothing to purge;
3. it has no course wiki timeslice (CWT) still needing update or reaggregation, **and**
   no article-course-user-wiki timeslice (ACUWT) still flagged `needs_update`;
4. it is not possibly running an update — i.e. no unfinished update log started
   within the last 14 days (older unfinished logs are treated as stale leftovers
   from a deploy/restart, since real updates take hours, not weeks).

### Implementation

- **`MarkPurgeableCourses`** is driven from the `courses` table and uses only
  `course_id`-indexed `exists?` lookups, so it **never scans a whole timeslice
  table** — important because ACUWT is huge.
- It purges **all** old courses with timeslices, not only `use_acuwt` ones.
  `use_acuwt` is a migration-era marker that will eventually be universal; the
  real criterion is "tracked in the timeslice system."
- **`Course#update_possibly_running?`** (with `RUNNING_UPDATE_WINDOW = 14.days`)
  distinguishes a genuinely-running update from a stale unfinished log left by a
  deploy/restart, so the many courses carrying stale logs aren't blocked forever.
- Runs Saturday 1am PT, deliberately off the Sunday-night backup.

### ACUWT re-update alignment (last three commits)

The eligibility check is kept consistent with re-update eligibility so a course
is never purged while it could still be legitimately reprocessed:

- Added a composite index `(needs_update, course_id)` on
  `article_course_user_wiki_timeslices`. Leading with the boolean `needs_update`
  is intentional — every query constrains it with `= true` (equality), so the
  seek jumps straight to the rare `true` rows; pairing it with `course_id` makes
  the scheduler's global query index-only and keeps the plan robust. Estimated
  ~200–300 MB at ~8M rows.
- `MarkPurgeableCourses` now **excludes** courses that still have a failing ACUWT
  row (`needs_update: true`), served index-only by the new index.
- `Course.needs_partial_update` now selects a course when **any** of its
  timeslices still needs work — a CWT flagged `needs_update` **or**
  `needs_reaggregation`, or an ACUWT flagged `needs_update`. This makes those
  archived-course score retries actually reschedulable, and the purge check
  mirrors it exactly (including `needs_reaggregation`) so no course lands in
  permanent limbo — protected from purging yet never rescheduled.

Note this **reverses** an earlier decision on this branch. The initial version of
the service deliberately *did not* check ACUWT `needs_update`, because at the time
those retries could never run (`ready_for_update` keys off CWT `needs_update`,
which ACUWT never propagates to). The last three commits change that: now that
`needs_partial_update` considers ACUWT, the retries *can* run, so purge-eligibility
must exclude failing-ACUWT courses to stay consistent.

Specs cover every branch of the eligibility check (`spec/services/mark_purgeable_courses_spec.rb`,
`spec/workers/mark_purgeable_courses_worker_spec.rb`) and the rewritten
`.needs_partial_update` scope (`spec/models/course_spec.rb`). The worker was also
run manually against local data (87 courses flagged on the first run, 0 on the
second).

## AI usage

Developed with Claude Code (Opus 4.8) under detailed, iterative human direction.
Claude drafted the code and commit messages; the human drove the design and
corrected it repeatedly — widening scope from `use_acuwt`-only to all old courses,
adding the no-pending-work and not-updating constraints, replacing a naive "any
unfinished log" check with the 14-day staleness window, and probing the index
design at length (single-column vs composite, boolean leading column, real row
counts). The human also drove the two "shouldn't we also…" course corrections
that shaped the final scope: adding the missing `has_many` ACUWT association and
including `needs_reaggregation` in the pending-work check. The human ran the
migration and confirmed real-world behavior manually.

This was developed across 5 commits over roughly 5 days (2026-06-27 → 2026-07-02)
in several sessions — sustained iteration with heavy human review, not a single
sitting. The design changed direction mid-branch (the ACUWT exclusion the first
commit deliberately omitted was added back once the scheduler was taught to
consider ACUWT).

## Screenshots

No UI changes (backend only).

- **Accepted consequence:** because `ready_for_update` feeds `ScheduleCourseUpdates`,
  a permanently-failing ACUWT row (an error that never resolves) will now keep its
  course perpetually rescheduled and perpetually non-purgeable. This was accepted
  as an acceptable, *visible* condition — a stuck course signals a real problem to
  investigate rather than something silently swallowed. It was impossible before
  (the retry never ran) and should be rare.
- **Just the eligibility flag** — the job that actually deletes purgeable courses'
  timeslices, and the guard blocking re-updates on purged courses, are planned for
  a later PR.
- **Index DDL on a large table:** building the `(needs_update, course_id)` index is
  online DDL on the ~10 GB ACUWT table; run it off the backup window on production.
- **Unmeasured cost:** `ReprocessArticleCourseUserWikiTimeslices#periods_with_failing_acuwt`
  filters on `needs_update` and likely scans ~all of a mega-course's rows per update;
  its real cost still isn't measured (the MariaDB slow query log is disabled and the
  Sentry quota is exhausted). Worth a follow-up.


gabina:
Explaining the new query `Course.ready_for_update` runs on production

```
EXPLAIN SELECT courses.*
FROM courses
WHERE
(courses.id IN
	(SELECT course_wiki_timeslices.course_id
	FROM course_wiki_timeslices
	WHERE (course_wiki_timeslices.needs_update = TRUE
			OR course_wiki_timeslices.needs_reaggregation = TRUE))
OR courses.id IN
	(SELECT DISTINCT article_course_user_wiki_timeslices.course_id
	FROM article_course_user_wiki_timeslices
	WHERE article_course_user_wiki_timeslices.needs_update = TRUE))
```
	

|id|select_type|table|type|possible_keys|key|key_len|ref|rows|Extra|
|--|-----------|-----|----|-------------|---|-------|---|----|-----|
|1|PRIMARY|courses|ALL|||||18279|Using where|
|3|MATERIALIZED|article_course_user_wiki_timeslices|ALL|index_acuwt_unique,index_acuwt_on_course_id_and_wiki_id,index_acuwt_on_course_id_and_user_id||||9346326|Using where|
|2|DEPENDENT SUBQUERY|course_wiki_timeslices|index_subquery|course_wiki_timeslice_by_course_wiki_start_and_end|course_wiki_timeslice_by_course_wiki_start_and_end|4|func|68|Using where|

The new (needs_update, course_id) should fix the ACUWT query.

PR #6935, opened 2026-07-01
Add Fall 2026 research experiment opt-in framework

ragesoss:
## What this PR does

Adds a reusable **opt-in research-experiment framework**, with the Fall 2026
Princeton study (the "Ribeiro experiment") as its first instance. The flow has
two sides:

- **Instructors** opt a course in through a new panel in the research-write
  assignment wizard. The panel only appears for courses eligible for an active
  experiment (`Course#eligible_for_active_research_experiment?` →
  `OptInExperiment.for_course`); for the Fall 2026 study, eligibility is a
  Fall 2026 `ClassroomProgramCourse` on the Wiki Education deployment. The
  choice is persisted as a course tag.
- **Enrolled students** of a participating course are individually shown a
  consent modal on the course page and opt in or out. Opting in makes no edit on
  the student's behalf: they are shown the userscript import line plus a link to
  their own English Wikipedia `common.js`, and paste it in themselves. The
  Dashboard then confirms the script is really there by reading the page, and the
  install step reappears on each course-page visit until it is.

The two sides are gated independently, so **this ships with the student side
dark**: instructors can start opting courses in on deploy, while no student is
invited until the userscript is ready to hand out. See below.

All student- and instructor-facing copy is supplied by the operator (in the
experiment's Ruby definition, the wizard YAML, and a consent-form Markdown
file), and is intentionally kept out of the i18n pipeline since it is ephemeral
experiment text.

Key pieces: `lib/experiments/opt_in_experiment.rb` (base class + registry) and
`lib/experiments/fall_2026_research_experiment.rb` (the concrete instance);
`ExperimentCoursesUser` + migration (durable per-student opt-in/opt-out record);
`app/services/check_experiment_userscript.rb` (reads the student's `common.js`
and records the install); `app/controllers/experiments/opt_in_controller.rb`
(invitation lookup, opt-in/opt-out, and the outstanding install step); the wizard
panel in `config/wizard/researchwrite/wizard.yml`; the
`experiment_opt_in_invitation.jsx` modal with `_experiment_opt_in.styl`; and
`Features.fall_2026_research_student_optin?` gating the student side.

### Holding the student side back

`Features.fall_2026_research_student_optin?` reads
`ENV['fall_2026_research_student_optin']` and gates everything student-facing.
The setting is named for this experiment rather than generically, so a later
experiment gets its own.

The generic half lives on the base class: `student_invitations_open?` (true by
default, so an experiment is fully live unless it opts into staging) and
`open_to_student?`, which combines it with `course_participating?`.
`Fall2026ResearchExperiment` overrides the former to read the setting. Three
places consult it — `needs_response?`, the controller's `can_respond?`, and
`pending_participation` — so no invitation is offered, the opt-in/opt-out
endpoints return 422 when posted to directly, and a student who somehow already
has a record is not shown an install step either.

The instructor gate stays on `Course#eligible_for_active_research_experiment?`.
Both the wizard and the course JSON used to call that one method, so gating it
would have hidden the wizard panel too; the course JSON now carries a separate
`research_experiment_open_to_students`, which also means the React component
issues no invitation lookup at all while the student side is closed.

**The default is off, at the ENV level rather than in code.** `config/application.yml`
is gitignored and lives on the server, so deploying this branch introduces no
key, and a missing key is not `'true'` — the student side is off on arrival with
no config step required. Going live later is a one-line addition to the server's
`application.yml` plus a restart. `application.example.yml` documents it.

### Why students install the script themselves

The original design had the Dashboard write the import line to the student's
`User:<name>/common.js` using their own OAuth credentials, prompting for
re-authorization when the token lacked the `editmyuserjs` grant. **That grant was
not approved for the production Wiki Ed OAuth consumer**, so that path is gone —
`InstallExperimentUserscript` and its missing-grant handling were deleted.

Confirming the script by reading the page needs no grant at all, and it verifies
ground truth rather than trusting an edit response: a student who saves the line
without telling us stops being prompted automatically.

The student pastes the line in rather than getting a prefilled edit box, because
MediaWiki offers no way to prefill one here:

- `preload=<page>` is gated on `ContentHandler::supportsPreloadContent`, which
  defaults to **false** and is overridden to true only by
  `WikitextContentHandler` and `JsonContentHandler`. A userscript has to live in
  a `.js` page, so `preload` can never apply to it. (The Dashboard's existing
  `preload=Template:…` sandbox and bibliography links work precisely because
  those targets are wikitext.) This was confirmed by a manual test against
  en.wiki after an earlier round of anonymous probing proved inconclusive — an
  anonymous request cannot reach the create-form for another user's JS page at
  all, so it could not distinguish "preload ignored" from "no permission."
- No URL parameter injects literal text into an edit box either — `preloadtext`,
  `text`, `wpTextbox1` and `initialtext` are all unrecognized.

So the install step shows the exact line and a link to the student's own
`common.js` edit form. The userscript page is a user `.js` page, which MediaWiki
restricts to its owner and interface admins, so the code participants run cannot
be altered by anyone else.

A gadget-based alternative was considered and set aside. WMF's Product Safety and
Integrity team declined the `editmyuserjs` grant outright, and suggested shipping
the script as a gadget the Dashboard could enable through the lower-risk
`editmyoptions` grant. That would allow a one-click opt-in and would avoid a
public edit, but it depends on an enwiki interface administrator and, realistically,
a community discussion about running a research tool as site-wide infrastructure —
too much schedule risk for a Fall 2026 start, for only a modest gain over pasting
a single line.

## Screenshots

This UI is entirely new — there is no "before" state on master.

**Instructor wizard opt-in panel** — the first assignment-wizard step for an
eligible course:

Instructor wizard opt-in panel

**Student consent modal** — shown on the course page to enrolled students of a
participating course, with the full consent form in a scrollable area:

Student consent modal

**Install step** — after consenting. "Install script" opens the student's own
`common.js` edit form, and the line above it is what they paste in:

Install step

**Not yet installed** — after a re-check that did not find the line on the
student's `common.js`:

Install step, not found

## AI usage

This branch was written by **Claude Code** under Sage's direction, across two
sessions.

The first session (2026-06-29, one commit) produced the original framework:
parallel exploration of the wizard, OAuth-edit, and course-page surfaces, a short
clarification round, and an approved plan, after which Claude Code drafted the
model, migration, service, controller, wizard, and React layers and ran the
lint/test suites. Sage steered the design incrementally: dropping an out-of-scope
mailer, adding the durable student record, moving copy out of `en.yml` into Ruby,
switching a banner to a consent modal, and supplying every piece of user-facing
copy.

The second session (2026-08-04) rebased the branch onto master (108 commits; one
`db/schema.rb` conflict, resolved by keeping both the new table and master's
`facilitator_stats`) and then reworked the install mechanism after the OAuth grant
was denied. Claude Code established the `preload` constraints described above by
probing live en.wiki, deleted the OAuth install path, added
`CheckExperimentUserscript`, reworked the controller payload and the React modal,
and updated the specs. It marked the copy the change had falsified with
`[PLACEHOLDER]` markers rather than rewriting it — the consent form's *Procedures*
section had promised automatic installation — and Sage supplied the replacement
wording, the install-step copy, and the real IRB number.

Claude Code initially built a two-path install step, prefilling the edit box with
`preload` where it believed MediaWiki allowed it, and reported the JS-page case as
a small residual risk. Sage's manual test showed it did not work at all; the cause
is a content-model restriction that the earlier anonymous probing had been unable
to reach. That branch was removed and the design collapsed to the single paste
path.

Finally, Sage asked for the student side to be gated so the instructor side could
deploy ahead of it, and for the setting to be named for this experiment rather
than generically. Claude Code split the shared eligibility predicate in two so
the gate could not hide the wizard panel, and verified the default-off behaviour
empirically rather than by inspection, since "safe on deploy with no config step"
is the property that change rests on.

Claude Code also wrote the commit messages and this description.

The screenshots above were captured by a feature spec
(`spec/features/experiment_opt_in_screenshots_spec.rb`) written by Claude Code for
this PR; it is gated on `ENV['SCREENSHOT']` and does not run in the main suite.
This description's summary and analysis may contain errors.

- `User:Sage_(Wiki_Ed)/fall2026experiment.js` (a draft of the tool) is in place.
  `User:Sage_(Wiki_Ed)/fall2026experiment-install.js` was created as a preload
  source and is now unused, since `preload` turned out not to apply to JS pages;
  it can be deleted.
- **The draft tool script embeds `BACKEND_SECRET`.** Anything in that script is
  world-readable, since browsers have to fetch it from a public page, so a shared
  secret there authenticates nobody. The backend should treat every client as
  untrusted and validate/rate-limit server-side rather than gate on a
  client-held token. (`BACKEND_URL` still points at `localhost`, as expected for
  a draft.)
- The consent form's *Procedures* section changed to describe the self-install
  step. Since that is IRB-approved text, it likely needs an IRB amendment.
- Eligibility is time-based (`inferred_term == 'fall_2026'`), so the experiment
  becomes inactive on its own once courses roll past Fall 2026; there is no admin
  toggle to end it early.
- The `db/schema.rb` change was hand-edited rather than dumped, to avoid local
  MariaDB collation noise — worth a look to confirm it matches a real dump.


ragesoss:
Review of HEAD `8213d4ec7`, ahead of merging and deploying the instructor-opt-in
half while the student side stays behind its setting. AI-generated review
feedback like this is not meant to be trusted automatically or acted on blindly
— worth reading and understanding, then using or ignoring based on the
developer's own judgment.

## Blocker: the instructor opt-in panel never renders

`8213d4ec7` replaced the instructor-side serializer field in
`app/views/courses/course.json.jbuilder` instead of adding the new one alongside
it:

```diff
-  json.eligible_for_active_research_experiment @course.eligible_for_active_research_experiment?
+  json.research_experiment_open_to_students @course.research_experiment_open_to_students?
```

The wizard gate is evaluated client-side as `course[panel.only_if]`
(`app/assets/javascripts/reducers/wizard.js`), and
`config/wizard/researchwrite/wizard.yml` gates the new panel on
`only_if: 'eligible_for_active_research_experiment'`. With the field no longer
serialized that lookup is `undefined` for every course, so the panel is filtered
out — meaning the half of this PR intended to deploy first would collect no
opt-ins at all.

This was reproduced locally rather than only reasoned about, with a
contrasting-cases feature spec: as of `8213d4ec7`, an eligible Fall 2026
`ClassroomProgramCourse` advances from the meeting-days panel straight to
"Training"; with the one line restored, the eligible course sees the research
panel and a Spring 2026 course does not. Restoring the line is the entire fix.

CI does not catch it: no spec asserts the contents of `course.json`, and no spec
exercises `only_if` panel filtering. A regression spec covering the gate is being
committed along with the fix.

## Checked and found sound

- **Student side is genuinely dark.** `Features.fall_2026_research_student_optin?`
  is off for any value but `'true'`, which makes
  `research_experiment_open_to_students` false so the React component never
  fetches, and server-side `open_to_student?` also fails so the opt-in/opt-out
  endpoints return 422. Gated on both sides independently.
- **Migration.** New table, no backfill, two indexes created on zero rows — safe
  inside a normal deploy on both the Wiki Ed and P&E hosts.
- **Tag semantics on a re-run.** `WizardTimelineManager#add_tags` overwrites by
  panel key, so an instructor who re-runs the wizard and answers "No" flips
  `fall_2026_research_opted_in` to `fall_2026_research_opted_out` rather than
  ending up carrying both tags.
- Linters clean on all changed files; the PR's own Ruby specs (38 examples) and
  the JS component spec pass locally, as do `spec/lib/wizard_timeline_manager_spec.rb`,
  `spec/controllers/courses_controller_spec.rb` (73 examples) and
  `spec/features/course_creation_spec.rb`.

## Notes, not blockers

1. **The opt-in option is pre-selected.** The "Yes, my class will participate"
   option carries `selected: true`, so an instructor who clicks through the
   wizard without reading enrolls the course by default. Other panels in this
   wizard use the same pattern, but research participation is a different kind of
   choice — worth confirming this is intended.
2. **Consent copy is reachable before launch.** `GET
   /experiments/courses/:id/invitation` returns the `copy` payload — invitation
   text plus the full consent form — unconditionally: no login, no enrollment
   check, and no `student_invitations_open?` check. Impact is low since the text
   becomes public when the student side goes live, but gating `copy` on
   `open_to_student?` would keep it dark until then.
3. **Wizard-only reachability.** The panel can only be reached by running or
   re-running the researchwrite wizard, and re-running rebuilds the timeline
   (`WizardTimelineManager#save_timeline`). Fall 2026 courses that have already
   generated a timeline cannot realistically opt in through this path, so the
   earlier this deploys the more courses it can catch; already-set-up courses may
   need a separate route.
4. **`Course#inferred_term` has no spec** anywhere, despite being the only gate on
   course eligibility (five branches, including December mapping to spring of the
   following year).
5. **Stale comment.** `app/models/experiment_courses_user.rb` still describes a
   pending intervention as "awaiting OAuth re-authorization", left over from the
   OAuth-based design that was removed.

*Drafted in two Claude Code sessions (~35 min total, 3 interactions from the
reviewer; the first session was lost to a local environment failure and the
review was resumed from its transcript). Claude Code did the code reading, wrote
the exploratory spec, and ran the specs and linters locally. Sage directed the
review and approved posting this comment, but did not independently verify the
findings beyond reading them. The exploratory spec and this comment are
AI-generated.*

ragesoss:
Re-review of `ab6f18b63`, covering the two commits since the previous review of
`8213d4ec7`. As before: AI-generated review feedback like this is not meant to be
trusted automatically or acted on blindly — worth reading and understanding, then
using or ignoring based on the developer's own judgment.

## The two flagged items are fixed, and both fixes were checked by reverting them

**The wizard gate blocker is resolved** (`a2232b3bc`). Rather than only confirming
the line is back, the fix was checked against the regression spec that shipped
with it: deleting `json.eligible_for_active_research_experiment` again makes
`spec/features/research_experiment_wizard_optin_spec.rb` fail exactly the way the
original bug presented — the eligible Fall 2026 course advances from the
assignment panel straight to "Training" — and restoring it makes both examples
pass. So the new spec genuinely covers the gate rather than merely accompanying
the fix.

**The pre-selected opt-in is resolved** (`ab6f18b63`). Re-adding `selected: true`
makes the same spec fail on `have_no_css('.wizard__option.selected')`, and
`minimum: 1` is what disables Next in `panel.jsx` when nothing is selected, so an
instructor cannot click past the panel without answering.

Also addressed: `Course#inferred_term` now has eight examples pinning all five
branches. These were checked against `inferDefaultCampaign.js` — the Ruby month
boundaries match the JS exactly, including December mapping to the following
spring. And the stale "awaiting OAuth re-authorization" comment is gone.

## Still open from the previous review

**The consent copy is still served to logged-out strangers.** `GET
/experiments/courses/:id/invitation` returns the full `copy` payload with no
login, no enrollment check, and no `student_invitations_open?` check. This was
confirmed with a request spec rather than by reading: while
`fall_2026_research_student_optin` is off, a logged-out request for an eligible
Fall 2026 course returns 200 with the entire consent form — and does so even for
an eligible course whose instructor has not opted in, so the endpoint also
answers "is this course eligible?" for anyone who asks. Impact stays low, since
the text becomes public when the student side goes live, but gating `copy` on
`open_to_student?` would keep it dark until then.

Wizard-only reachability (note 3 last time) is unchanged; that one is operational
rather than a code change.

## New observations, not blockers

1. **Nothing exercises the student side end to end.** The modal is covered by the
   jest component spec against a mocked fetch, and the endpoints by the controller
   spec, but the only browser-level spec that drives the real component against
   the real controller is `experiment_opt_in_screenshots_spec.rb`, which is
   `ENV['SCREENSHOT']`-gated and so contributes nothing to CI. That is a
   reasonable gap while the student side ships dark — it is the coverage worth
   adding before `fall_2026_research_student_optin` is turned on.
2. **Re-running the wizard now requires re-answering, and a re-run can undo a
   prior "Yes".** `WizardController#wizard` renders the panel YAML as-is with no
   merge of the course's existing tags, so a re-run shows the panel blank rather
   than reflecting the earlier choice, and `add_tags` overwrites by key — so
   answering "No" on a re-run flips a previously opted-in course out. Requiring
   an explicit answer is still the better trade; this is just a note that the
   re-run path now silently discards an earlier opt-in.

## Checks run

RuboCop clean on all changed Ruby files; ESLint clean on the changed JS; `yarn
build` succeeds. Green locally with master merged in: the PR's Ruby specs
alongside `spec/models/course_spec.rb` (116 examples), the jest component spec (7),
the new `research_experiment_wizard_optin_spec.rb` (2),
`spec/lib/wizard_timeline_manager_spec.rb` (7), and
`spec/features/course_creation_spec.rb` (5). The screenshot spec correctly
contributes zero examples when `SCREENSHOT` is unset.

*Drafted in a Claude Code session (~25 min, 2 interactions from the reviewer).
Claude Code did the code reading, ran the specs and linters locally, verified
both fixes by reverting them and watching the regression spec fail, and confirmed
the unauthenticated `copy` exposure with a throwaway request spec. Sage asked for
the re-review and approved this comment before posting, but did not independently
verify the findings. This comment is AI-generated.*

PR #6940, opened 2026-07-03
[WIP] Add weekly worker to purge timeslices for purgeable courses

gabina:
## What this PR does

Adds the timeslice purge job from the ACUWT bloat-deletion plan: the piece
that actually deletes timeslices for courses flagged as purgeable.

`PurgeTimeslicesWorker` runs weekly (scheduled on Fridays via
`config/schedule.yml`), finds courses carrying the `purgeable` flag, and
deletes all four of their timeslice tables (ACUWT, ACT, CUWT, CWT) via the
existing `TimesliceCleaner#delete_all_timeslices_for_course`. After purging a
course it drops the `:purgeable` flag and sets a `:purged` flag, so the course
falls out of the query on subsequent runs instead of being re-scanned.

Design notes:
- **Friday schedule** is deliberate: a course stays flagged for a few days
  before its timeslices are actually removed, leaving a window to catch a
  mistake before the data is gone.
- **`COURSES_PER_RUN = 20`** caps how many courses are purged per run so a
  large backlog doesn't hit the DB with a flood of deletes at once; the
  remainder drains at 20/week.
- **Querying the flag**: `flags` is a serialized Hash stored as text, so it
  can't be filtered precisely in SQL. The query narrows with a raw
  `LIKE '%purgeable%'` and then confirms with the `purgeable?` predicate.
- **`daily_update` queue** was chosen for process isolation — it's serviced
  by a dedicated Sidekiq process, so a heavy batch of deletes can't starve the
  course-update queues. The queue is set only in the schedule entry (not in
  `sidekiq_options`), matching the other scheduled workers.
- `Course#purgeable?` is reused from the MarkPurgeableCourses PR (#6924),
  which sets the flag. That PR is not yet merged; this branch is otherwise
  independent of it and includes only the `purgeable?`/`purged?` predicates it
  needs. Until #6924 merges and starts setting the flag, this worker simply
  finds nothing to purge.

This is the "actual purge/delete job" listed as future work in the ACUWT
bloat-deletion plan; it builds on the already-merged
`TimesliceCleaner#delete_all_timeslices_for_course`.

## AI usage

The code, specs, and commit messages were written by Claude Code (Opus 4.8)
under my direction. I described the goal (a weekly Friday worker that purges
timeslices for purgeable courses) and pointed it at the `purgeable?` method to
reuse; Claude explored the codebase, wrote the worker, schedule entry, model
predicates, and spec, and ran the tests and RuboCop. I steered the design
through several corrections: renaming the marker flag to `purged`, keeping the
schedule rationale out of the worker, removing a redundant `queue:` from
`sidekiq_options`, confirming the worker belongs at the top level rather than
in the `daily_update/` sub-folder, and adding the per-run cap.

Note that this description — including the "What this PR does" summary — was
also drafted by Claude Code and may contain errors.

## Screenshots

No UI changes — this is a backend worker plus a scheduler entry.

- **Coordination with MarkPurgeableCourses (#6924).** That service should skip
  courses that are already `purged?` so it can't re-flag a purged course as
  `purgeable` and feed it back through this worker. Worth confirming when the
  two branches come together.
- **Blocking re-updates** of purged courses (so a purged course can't be
  re-inflated by a full update or zeroed by partial re-aggregation) is separate
  follow-up work from the ACUWT plan and is not in this PR. The new `purged?`
  predicate is the marker that future work would consume.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -588,6 +588,18 @@ class Course < ApplicationRecord
     flags[:use_acuwt].present?
   end
 
+  # True once MarkPurgeableCourses has flagged this course for timeslice purging
+  # (it ended long ago, was tracked in the timeslice system, and has no pending
+  # timeslice work).
+  def purgeable?
+    flags[:purgeable].present?
+  end
+
+  # True once PurgeTimeslicesWorker has deleted this course's timeslice records.
+  def purged?
+    flags[:purged].present?
+  end
+
   def debug_updates?
     flags[:debug_updates].present?
   end
--- a/app/workers/purge_timeslices_worker.rb
+++ b/app/workers/purge_timeslices_worker.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/timeslice_cleaner"
+
+# Deletes every timeslice record for courses that MarkPurgeableCourses has
+# flagged as purgeable. See Course#purgeable?.
+class PurgeTimeslicesWorker
+  include Sidekiq::Worker
+  sidekiq_options lock: :until_executed
+
+  # Cap how many courses are purged per run so a large backlog doesn't hit the
+  # DB with a flood of deletes at once. Remaining courses keep their purgeable
+  # flag and are picked up on the next run.
+  COURSES_PER_RUN = 20
+
+  def perform
+    purgeable_courses.each do |course|
+      TimesliceCleaner.new(course).delete_all_timeslices_for_course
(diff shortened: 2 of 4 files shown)


gabina:
Don't think it's related to these changes

```
Finished in 12 minutes 52 seconds (files took 4.58 seconds to load)
889 examples, 1 failure, 6 pending

Failed examples:

rspec ./spec/models/visiting_scholarship_spec.rb:155 # VisitingScholarship#scoped_article? considers assigned articles as scoped articles even though the title changed
```

Issue #6602, opened 2025-12-26
Clean up and test prototype accessibility improvements for ArticleViewer


ragesoss:
In fall 2025, a team from [Harvard Tech for Social Good](https://socialgood.hcs.harvard.edu/) (T4SG) worked on https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/2522

Their final PR  — https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6597 — includes a few enhancements for screen reader accessibility within ArticleViewer, including changes to how the highlighted text is annotated, and changes to how screen reader cursor location updates when the ArticleViewer scrolls to a particular user's next contribution.

The PR isn't ready to merge, but the semester-long project for the T4SG team has ended. I think the best strategy from here will be to open smaller clean PRs that extract the different usable aspects of the T4SG work into independent changesets (and add corresponding tests and demo videos).

The first feature is to change the Authorship Highlighting markup, so that each continuous section of text by a highlighted user is enclosed in a relevant HTML with relevant metadata. (This is in contrast to the current markup, where each highlighted work is its own span, annotated with the username of the editor... resulting in an unusable mess of confusing annotations via screen reader).

The next feature improves the accessibility of the ArticleViewere legend and the scroll behavior when clicking to advance to the next section of changes by a user.

katara-Jayprakash:
the one is also good! wana give it a try

ragesoss:
@katara-Jayprakash great! please do!

nexpectArpit:
@ragesoss is issue assignation is necessary to work and create a pull request?

Formasitchijoh:
@nexpectArpit  issues are not assigned here. Just pick up any issue you feel comfortable and indicate you are working on it to prevent conflicts. For issues someone indicated they are working on, but have  been inactive for a long time confirm with them if you can take up the issue.

katara-Jayprakash:
ya @Formasitchijoh thanx for reminding, and ya i am working on this currently

NIGHTFURY609:
Can i take up this issue?

ragesoss:
Sure.

Worksalohy:
Is anyone currently working on this issue? If not, I would like to give it a try.

NIGHTFURY609:
@Worksalohy hi, i am currently working on this issue.

nexpectArpit:
@ragesoss is this issue open for work?

ragesoss:
[...]

PR #6597, opened 2025-12-20
T4SG Final PR. Feel free to use or not use.

SamC1249:
Will send description and information to email.

PR #6945, opened 2026-07-06
ArticleViewer legend accessibility and scroll focus management

nexpectArpit:
- Accessible Buttons: Replaced the non-keyboard-focusable scroll arrow icons in article_viewer_legend.jsx with native <button> elements. These now have proper aria-label tags reflecting the editor's highlighting state.
- Focus Redirection: Updated ArticleScroll.js's scrollTo() method to return the DOM node of the scrolled-to target. The click/action handler in the legend now redirects keyboard and screen reader focus directly to the target paragraph (adding tabindex="-1" programmatically if not already present).
- Safe Fallbacks: Maintained clean mouse-user layouts and fallback links to talk pages (↗) when authorship highlighting is not available or disabled for a user.
- Crashes Prevented: Added a guard clause in ArticleViewer.jsx's fetchUserIds to return early when there are no editors in a course, preventing TypeError: No users provided! page crashes when previewing articles with no edits.
- Aesthetic Focus States: Added clean browser resets and focus outlines for the new legend buttons in the Stylus stylesheet.
- Unit Test Suite: Created a dedicated Jest test suite in test/components/article_viewer_legend.spec.js to ensure stability across legend button rendering, localized labels, focus management, and empty state fallbacks.
#### Verification
Verified in the browser:
- Tab navigation highlights the editor legend name buttons with focus rings.
- Pressing Enter / Space triggers smooth scrolling and transfers focus directly to the target paragraph.
- No TypeError crashes occur when previewing unedited/new articles in the course.
- Clean ESLint and unit test pass locally.
- Automated Tests: Ran Jest unit test suite (npx jest test/components/article_viewer_legend.spec.js) — 5/5 tests passed cleanly.

#### screen recording:

https://github.com/user-attachments/assets/711bb213-a34b-413d-87af-ffde000e33f3

addresses #6602

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/ArticleScroll.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/ArticleScroll.js
@@ -84,6 +84,7 @@ export class ArticleScroll {
         } else {
             this.scrollObject[name].index += 1;
         }
+        return paragraphs[index].paragraph;
     }
 
     findClosestEdit(name, paragraphs) {
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/ArticleViewerLegend.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/ArticleViewerLegend.jsx
@@ -23,29 +23,128 @@ const ArticleViewerLegend = ({ article, users, colors, status, allUsers, failure
       }
 
       setUserLinks(users.map((user, i) => {
-        let res;
         // The 'unhighlightedContributions' keeps track of the userids of users whose contributions
         // were not successfully highlighted in the article viewer.
         const UnhighlightedContributions = unhighlightedContributors?.find(x => x === user.userid);
         const userLink = UserUtils.userTalkUrl(user.name, article.language, article.project);
         const fullUserRecord = allUsers.find(_user => _user.username === user.name);
         const realName = fullUserRecord && fullUserRecord.real_name;
+        const colorClass = colors[i];
+
+        const handleScroll = () => {
+          if (!scrollBox) return;
+
+          const target = Scroller.scrollTo(user.name, scrollBox);
+
+          if (target && typeof target.focus === 'function') {
+            // Make sure it can receive programmatic focus
+            if (!target.hasAttribute('tabindex')) {
+              target.setAttribute('tabindex', '-1');
+            }
+            target.focus();
+          }
+        };
+
+        // Accessible label for the scroll button, tailored by state
+        let scrollAriaLabel;
         if (status === 'loading') {
-          res = <div key={`legend-${user.name}`} className={'article-viewer-legend'}><a href={userLink} title={realName} target="_blank">{user.name}</a></div>;
+          scrollAriaLabel = I18n.t('users.loading_authorship_for', { username: user.name });
         } else if (user.activeRevision === true) {
-          res = <div key={`legend-${user.name}`} className={`article-viewer-legend user-legend-name ${colors[i]}`}><a href={userLink} title={realName} target="_blank">{user.name}</a><button type="button" className="user-legend-hover-button" onClick={() => Scroller.scrollTo(user.name, scrollBox)}><img className="user-legend-hover" style={{ color: 'transparent' }} src="/assets/images/arrow.svg" alt="scroll to users revisions" width="30px" height="20px" /></button></div >;
+          scrollAriaLabel = I18n.t('users.scroll_to_users_edits', { username: user.name });
         } else if (UnhighlightedContributions) {
-          res = <div key={`legend-${user.name}`} className={'article-viewer-legend tooltip-trigger'}><p className={'tooltip large'} id={'popup-style'} >{I18n.t('users.contributions_not_highlighted', { username: user.name })}</p><a href={userLink} title={realName} target="_blank">{user.name}</a>{<span className="tooltip-indicator-article-viewer" />}</div>;
+          scrollAriaLabel = I18n.t('users.contributions_not_highlighted', { username: user.name });
         } else {
-          res = <div key={`legend-${user.name}`} className={'article-viewer-legend tooltip-trigger'}><p className={'tooltip large'} id={'popup-style'}>{I18n.t('users.no_highlighting', { editor: user.name })}</p><a href={userLink} title={realName} target="_blank">{user.name}</a>{<span className="tooltip-indicator-article-viewer" />}</div>;
+          scrollAriaLabel = I18n.t('users.no_highlighting', { editor: user.name });
         }
 
-        return res;
+        const isClickable = status !== 'loading' && user.activeRevision === true;
+
+        // Common wrapper class
+        const wrapperClassNames = [
+          'article-viewer-legend',
+          'user-legend-name',
+          (user.activeRevision === true ? colorClass : ''),
+          (UnhighlightedContributions || (!user.activeRevision && status !== 'loading') ? 'tooltip-trigger' : '')
+        ].filter(Boolean).join(' ');
+
+        // Base button element: focusable, screen-reader friendly
+        const scrollButton = isClickable && (
+          <button
+            type="button"
+            className="article-viewer-legend-button"
+            onClick={handleScroll}
+            aria-label={scrollAriaLabel}
+          >
+            <span aria-hidden="false">{user.name}</span>
+            <img className="user-legend-hover" style={{ color: 'transparent' }} src="/assets/images/arrow.svg" alt="" width="30px" height="20px" />
+          </button>
+        );
+
+        let tooltip = null;
+        if (status === 'loading') {
+          // No tooltip; message handled in usersStatus
+          tooltip = null;
+        } else if (user.activeRevision === true) {
+          // Active user — no tooltip needed, label explains behavior
+          tooltip = null;
+        } else if (UnhighlightedContributions) {
+          tooltip = (
+            <p className="tooltip large" id="popup-style">
+              {I18n.t('users.contributions_not_highlighted', { username: user.name })}
+            </p>
+          );
+        } else {
+          tooltip = (
+            <p className="tooltip large" id="popup-style">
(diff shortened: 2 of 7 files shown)


ragesoss (changes requested):
**How to read this feedback:** the notes below come from an AI-assisted review session. They aren't meant to be trusted or acted on automatically — read them, verify what's relevant, and use or discard each point based on your own judgment.

**Verified: the no-editors crash fix works.** The `useAuthorshipHighlighting` guard was reproduced end-to-end. A tracked `articles_courses` row with an empty `user_ids` makes `/articles/details` return `editors: []`, so the viewer's `users` prop becomes `[]` (truthy, so it passes the `isOpen && users` effect guard), and `AuthorshipURLBuilder` throws `TypeError: No users provided!`. A spec exercising this against **master** fails with an uncaught `No users provided!` that unmounts the React root (there's no error boundary); against this branch the viewer opens cleanly. Note the trigger is the article-*list* path — the article-finder path passes no `users` and was already safe.

**Worth addressing:**

- **Focus may fight the smooth scroll** (`ArticleViewerLegend.jsx`, `handleScroll`). After `Scroller.scrollTo(...)` — which animates `#article-scrollbox-id` with a 100 ms delay — the handler calls `target.focus()` synchronously. A default `.focus()` instantly scrolls the same container to bring the paragraph into view, producing a jump *before* the smooth animation runs. `target.focus({ preventScroll: true })` preserves the focus move (the accessibility goal) without the competing scroll. This was not verifiable in the test environment (authorship highlighting needs a live WhoColor call, which fails on CORS in CI — the same reason `article_viewer_spec.rb` is pending), so it warrants a manual check in the browser.

- **Hardcoded `aria-label="View User Talk Page"`** on the new talk link isn't localized, unlike the two keys added to `en.yml`. It's also identical on every row, so it doesn't identify whose talk page. Consider an I18n key that includes the username.

**Minor:**

- `user-legend-name` is now applied to every legend state, not just active editors, so loading/unhighlighted rows gain 10 px of left padding they didn't have. Cosmetic — worth a glance.
- For active editors the primary click target changed: the username now scrolls, and talk-page access moves to the small `↗`. Reasonable for the accessibility split, but a behavior change for mouse users.
- Focus outline uses a hardcoded `#2b79ff` rather than a Stylus variable; `scrollAriaLabel` is computed for states where it's never rendered (harmless).

**Tests.** #6602 asks for corresponding tests and demo videos, and this changeset adds neither. The legend rewrite and the scroll-focus behavior — the substantive, silent-failure-prone parts — are untested. The exploratory spec written during this review covers the crash-guard case and can be shared if useful as a starting point.

The dependency additions to the legend effect (`allUsers`, `article.language`, `article.project`) look correct — `allUsers` is the stable `state.users.users` reference, so there's no re-render loop, and late-arriving user records now resolve `realName`.

*Drafted in a Claude Code session (~35 min, ~4 interactions). Sage directed the review depth and phases and reviewed the full comment before posting. The crash-guard finding was verified by running an AI-written exploratory spec against both master (fails: `No users provided!`) and this branch (passes); that spec is AI-generated and was run but not line-by-line audited by the reviewer. The `preventScroll` and other findings were reasoned from the code and spot-checked, not independently reproduced.*

ragesoss:
All items from the earlier review have been addressed:

- `target.focus({ preventScroll: true })` resolves the focus-vs-smooth-scroll conflict.
- The talk-page link label is localized per-user via `users.view_user_talk_page`.
- `user-legend-name` and the color class are scoped to active editors again, restoring the original layout for loading/unhighlighted rows.
- The focus outline uses `$brand_primary`, and the unused `scrollAriaLabel` computation is gone.
- The new Jest suite covers the legend's aria-labels, class scoping, empty-users fallback, and the `preventScroll` focus call.

Verified during re-review: the new suite passes (5/5), the full JS suite passes (418 tests), ESLint is clean on the changed files, and a production build succeeds with only pre-existing warnings.

One minor gap remains: the `useAuthorshipHighlighting` empty-editors guard (the crash fix itself) has no committed test — the new suite exercises the legend with `users: []` but not the hook's early return. Fine to handle in a follow-up if at all.

*Drafted in a Claude Code session (~15 min, 2 interactions). Sage initiated the re-review and read the comment before posting. Verification (Jest suite runs, full JS suite, ESLint, production build, and code inspection of each fix against the prior review) was performed by Claude Code; Sage spot-checked the findings but did not independently reproduce them.*

ragesoss:
I don't like the ↗ links. I'm not sure what problem they are trying to solve, but they are too easy to click by accident next to the 'next segment' link that is already present.

What is the accessibility motive for changing the current behavior where the username itself is the link?

nexpectArpit:
> I don't like the ↗ links. I'm not sure what problem they are trying to solve, but they are too easy to click by accident next to the 'next segment' link that is already present.
> 
> What is the accessibility motive for changing the current behavior where the username itself is the link?

hey @ragesoss , thanks for the feedback! and apologies for late reply ,actually  the context on the design choices were:

The layout changes: specifically separating the scroll action from the talk-page link were adopted directly from the closed accessibility prototype in PR #6597 (created by the Harvard team).

In their original PR, the developers were following WCAG guidelines on semantic controls: they realized that having the username text act as both an external link and an inline scroll trigger violated standard screen-reader expectations. To solve this, they introduced the arrow button as a dedicated <button> for the scroll action, keeping it separate from the link. We adopted their design to preserve that accessibility baseline.

However, i agree with your concern about usability for mouse users regarding separate ↗ link which i added is a very small click target, and placing it right next to the scroll button makes it easy to misclick.

so the updated adjustment is to keep the layout clean and familiar for mouse users while still respecting the accessibility goals, i can update the code to: Revert the username text back to being a standard link (<a>) to visit the talk page (removing the ↗ icon). Keep the arrow next to it as the dedicated, focusable "button"  for scrolling.
This restores the original visual structure for mouse users while keeping the scroll control accessible. I will update the PR with this adjustment if this looks ok to you!

ragesoss:
Sure, go ahead with that adjustment. I think the behavior of that link is not a key accessibility priority; it's not something I've seen come up with real users, so I'd rather retain current behavior for that particular bit except for any enhancements that leave the mouse-based user experience unchanged. (We can always return to that issue separately, later on.)

nexpectArpit:
@ragesoss i updated the pr and the demo video. PTAL

ragesoss (changes requested):
**How to read this feedback:** the notes below come from an AI-assisted review session. They aren't meant to be trusted or acted on automatically — read them, verify what's relevant, and use or discard each point based on your own judgment.

All items from the prior review are addressed — `preventScroll: true`, the localized aria-label, scoped `user-legend-name`, `$brand_primary` focus outlines, simplified `scrollAriaLabel`, and the new test suites. The `↗` talk links were removed per earlier feedback, restoring the original behavior where the username itself is the talk-page link.

The new test suites (6 tests across two files) cover the key behaviors well: aria-labels, class scoping, empty-users safety, `preventScroll` focus integration, and the hook crash guard.

**Three pieces of dead code to clean up before merge:**

1. **`.user-legend-talk-link` CSS** (`_article_viewer.styl`, lines 304–311) — this class was added for the `↗` link element, which has since been removed. No element in the codebase references it.

2. **`loading_authorship_for` i18n key** (`en.yml`, line 1877) — this key was used in an earlier version of the multi-branch `scrollAriaLabel` computation, which was simplified to only produce a label for clickable buttons. No JS/JSX code references this key.

3. **`view_user_talk_page` test setup** (`article_viewer_legend.spec.js`, line 14) — this translation is injected in the test setup but is never asserted against, and the key does not exist in `en.yml`. Appears to be left over from an earlier iteration that included the `↗` talk link.

**Verified locally:** build succeeds, ESLint clean on all changed files, PR's own test suites pass (6/6), full JS suite passes (432/432, no regressions).

*Drafted in a Claude Code session (~20 min, ~8 interactions). Sage directed the review, read through the full walkthrough and coverage assessment, and reviewed this comment before posting. Local verification (build, ESLint, Jest suites) was performed by Claude Code; Sage did not independently reproduce those runs.*

nexpectArpit:
removed the dead code part, will start working soon on PR2 proposed in the plan!

ragesoss (changes requested):
**How to read this feedback:** This review comes from an AI-assisted session. It should not be trusted or acted on automatically; please verify the finding and use or disregard it based on your own judgment.

The latest cleanup resolves all three dead-code items from the prior review. However, one test-coverage issue remains:

- `test/components/use_authorship_highlighting.spec.js` passes `users: null`. That bypasses the effect which calls `fetchUserIds()`, so the empty-editor guard is never exercised. A temporary mutation replacing the guard with a deliberate exception still left this test passing. Changing the input to `users: []` reaches the intended crash path and passes with the current implementation; the test name should be updated accordingly.

Focused verification on the unmodified PR head: both Jest suites pass (6/6), and ESLint is clean on all changed JavaScript and test files.

*Drafted in a Claude Code session (~20 min, ~3 interactions). Sage directed the re-review, read the findings, and approved this review body before posting. Local verification—including the focused Jest suites, ESLint, code-path tracing, and a temporary mutation check—was performed by Claude Code; Sage did not independently reproduce those checks.*

ragesoss (approved):
**How to read this feedback:** This review comes from an AI-assisted session. It should not be trusted or acted on automatically; please verify and use or disregard based on your own judgment.

The remaining issue from the prior review is resolved. The new test passes `users: []`, which triggers the fetch effect and reaches the `fetchUserIds` guard clause. This was verified with a mutation check: replacing the guard's early return with a thrown exception makes the new test fail, confirming it exercises the intended code path. Keeping the original `users: null` case alongside it as a separate test works well.

Verified locally on the PR head: both Jest suites pass (7/7), ESLint is clean on the changed file, and the build succeeds.

With this, all items from the earlier reviews are addressed.

*Drafted in a Claude Code session (~10 min, 1 interaction). Sage initiated the re-review and approved this review body before posting. Local verification — the Jest suites, ESLint, the build, and the mutation check on the guard clause — was performed by Claude Code; Sage did not independently reproduce those checks.*

PR #6950, opened 2026-07-10
Move article tracked-status reset logic into ArticleNamespacesManager

gabina:
## What this PR does

Consolidates all the "reset articles whose tracked status changed" logic into `ArticleNamespacesManager`, which is now the single place that decides when a course's articles should be re-included in or excluded from course statistics. Previously this logic was split between `ArticleNamespacesManager` and the tail end of `ArticleStatusManagerTimeslice.update_article_status_for_course` (via `ArticlesCoursesCleaner.reset_articles_for_course`), and the untracked-namespaces reset ran **twice** per update for courses where both classes ran.

Concretely:

- `ArticleNamespacesManager` keeps its original namespace-move handling for every course (including article scoped programs, which rely on it as best-effort namespace-move detection since they never run the article status sync), and absorbs the deleted-articles and undeleted/re-tracked resets that previously ran from the status manager.
- The absorbed resets are gated behind a new `statuses_synced:` keyword so they run under exactly the same conditions as before (only when `UpdateCourseStats#should_update_article_status?` is true). The undeleted/re-tracked reset is additionally guarded against only-scoped-articles courses even if the flag is passed: non-scoped articles get article course timeslices but never an `articles_courses` record, so that reset would reprocess them on every update.
- `ArticleStatusManagerTimeslice` now purely syncs article records (title, namespace, deleted, mw_page_id) against Replica. The one reset it keeps is the duplicate-`mw_page_id` case in `handle_undeletion`, which is documented in the class comment.
- `ArticlesCoursesCleaner` keeps only the low-level `reset` primitive (shared with `reset_specific_articles` callers) and its unrelated cleanup methods.

No behavior change is intended other than removing the duplicated untracked-namespaces pass.

**Later commits on this branch** add info-level Sentry logging for every reset and status-change event (`Article retracked`, `Article untracked`, `Article namespace changed`, with course, reason, and article ids), to measure how frequently these resets happen in production — the triggering scenarios depend on on-wiki events landing in the middle of course updates, so they are nearly impossible to observe otherwise. They also rename `ArticleStatusManagerTimeslice` to `ArticleStatusManager`, now that it is the only class syncing article statuses.

## AI usage

The code in this PR was written by Claude Code (Fable 5) under close human direction. The session started as a code-analysis conversation about whether `ArticleNamespacesManager`'s "flag timeslices for reprocessing" comment was justified; during that analysis the duplicated reset behavior between the two classes was spotted, and Claude was directed to move all tracked-status reset logic into `ArticleNamespacesManager`. Claude identified that the `should_update_article_status?` gating is load-bearing (the re-tracked reset would loop forever on scoped courses), and the human supplied the key design context that the namespace-move detection must stay available for article scoped programs. Claude initially preserved old spec assertions by invoking `ArticleNamespacesManager` inside the status manager's spec file; the human flagged that specs should stay unit tests, and the specs were reworked accordingly. Claude also wrote the commit message.

Scale of effort: the original refactor was a single commit, developed and reviewed in one working session, with all spec runs and RuboCop passing on the first attempt. Five follow-up commits from a second session (also written by Claude Code under human direction) added the Sentry logging and the rename described above.

This "What this PR does" summary and the rest of this description were also drafted by Claude Code and may contain errors. This PR description was drafted using a Claude Code skill (`/prepare-pr`).

## Screenshots

No UI changes.

- The `statuses_synced:` flag preserves the existing gating exactly, but it means the deleted/re-tracked resets still never run for P&E courses whose updates take over 10 minutes (same as before). Whether those resets could safely run unconditionally (they are DB-only, no Replica calls) is left as possible follow-up.
- `moved_to_mainspace` is named after the common case, but it really detects moves into any tracked namespace; renaming it was left out of scope.


ragesoss (approved):

The refactor's behavior-preservation was checked by tracing the call paths against master, not only by reading the description:

- The removed second untracked-namespaces pass is redundant. `ArticlesCoursesCleaner#reset` only deletes `articles_courses` rows and never creates them, so no reset re-adds articles a later pass would need to catch. `reset_undeleted_or_retracked_articles` acts only on tracked-namespace articles with no `articles_courses` record (a no-op on that table), while `reset_articles_in_untracked_namespaces` acts only on untracked-namespace articles that have one — the two sets are disjoint by namespace, so their reordering does not change results. `moved_to_mainspace` articles sit in a tracked namespace and are never in scope for the untracked pass. Running the untracked pass once appears equivalent to the old twice.

- The `statuses_synced: false` path matches the old non-status-sync behavior, and the `true` path matches the old both-run behavior minus the duplicate.

- The gating is preserved: `statuses_synced` is fed from `should_update_article_status?`, the same predicate that gated the old status-manager reset.

- Caller check: `update_article_status_for_course` has a single production caller (`UpdateCourseStats`), and the removed public methods (`reset_articles_for_course`, class-level `reset_articles_in_untracked_namespaces`) have no remaining callers. `reset_specific_articles` is correctly retained for its other three callers.

Non-blocking observations:

- The `unless @course.only_scoped_articles_course?` guard (`article_namespaces_manager.rb:47`) is unreachable in the current production path — `should_update_article_status?` already returns `false` for scoped courses, so `statuses_synced` is always `false` there and the block is skipped. It reads as intentional defense-in-depth (consistent with the PR description's "even if the flag is passed"); noting it only so it is not later mistaken for load-bearing.

- `log_namespace_change` fires an info-level Sentry event per namespace change on every status sync, likely more frequently than the reset logs (which depend on mid-update on-wiki events). The description frames the instrumentation as intentional and temporary; flagging the relative volume only.

- Minor: `articles_in_tracked_namespaces` materializes full Article records where the old untracked path used `pluck(:id)`. Negligible.

RuboCop is clean on all changed files and the moved/added specs pass (38 examples). No migrations.

---

*Reviewed in a Claude Code session (~20 min, a handful of interactions) using the repo's `/review-pr` skill. Claude Code performed the call-path tracing and disjointness analysis, ran RuboCop and the affected specs (`article_namespaces_manager`, `article_status_manager`, `articles_courses_cleaner` — 38 examples, 0 failures) locally against the branch merged with master, and drafted this comment. Sage directed the review and read the findings before posting but did not independently re-derive the tracing. No integration-level spec was written for the `UpdateCourseStats#update_articles` wiring.*

PR #6956, opened 2026-07-10
Fix reference-counter API support for multilingual Wikisource and wikimedia wikis

gabina:
# Fix reference-counter API support for multilingual Wikisource and wikimedia wikis

## What this PR does

Fixes a steady stream of Sentry errors — `Non-200 response hitting references counter API: (404)` — coming from courses that track multilingual Wikisource (wikisource.org). That wiki is deliberately stored with `language = nil`, so `ReferenceCounterApi#batch_query_url` interpolated an empty final URL segment (`/api/v1/references/wikisource/`), which matches no route on the Toolforge app and 404s on every batch, through all 5 retries. The reference-counter API builds hostnames as `{language}.{project}.org`, and `www.wikisource.org` is the real multilingual Wikisource domain, so the fix is to substitute `www` for the nil language. Verified live with the exact rev_ids from the Sentry events (200 with per-rev counts) and with a page containing two literal `<ref>` tags (`num_ref: 2`).

While auditing related edge cases, we also found that the `wikimedia` project exclusion in `UNSUPPORTED_PROJECTS` is stale: the code comment claimed the API responds 400 "Language X is not a valid language" for wikimedia wikis, but live testing showed all three wikimedia wikis the Dashboard can actually have (`commons`, `incubator`, `meta` — anything else fails `Wiki` model validation) return 200 with correct counts, verified against pages containing literal `<ref>` tags. So this PR removes `wikimedia` from the exclusion list, and those wikis will start accumulating real reference counts. Wikidata remains excluded; the API rejects it server-side with "Project wikidata is not a valid project."
Note that we don't get "Language X is not a valid language" error anymore since we deleted the constant list of supported language from the reference-counter API source code. 

Relevant external context: [reference-counter on Toolforge](https://toolsadmin.wikimedia.org/tools/id/reference-counter).

## AI usage

This PR was developed with Claude Code (Claude Fable 5). Claude investigated the Sentry errors, traced the 404 to the nil-language URL segment, and initially proposed excluding multilingual Wikisource from the API — trusting the stale code comment. The human author pushed back and asked to verify empirically; Claude's live probes against the Toolforge API showed the `www` segment works, and a follow-up sweep requested by the author showed the wikimedia wikis work too, turning a proposed exclusion into a fix that recovers data. The author specified the shape of both changes (scope the `www` fallback inside `batch_query_url`; drop `wikimedia` from `UNSUPPORTED_PROJECTS`) and the two-commit split; Claude drafted the code, specs, and commit messages under that direction, and the author reviewed each edit.

Both commits were made within a single session on 2026-07-10, a few minutes apart; the investigation and live API verification took roughly an hour of back-and-forth. This summary and the rest of this description were also drafted by Claude Code and may contain errors.

## Screenshots

No UI changes.

- Removing `wikimedia` from the exclusion list means commons/incubator/meta revisions will start getting reference counts where they previously got none. Refs are meaningful on incubator (proto-Wikipedia articles) and present-but-rare on commons/meta; if that data is unwanted for some reason, the first commit can be dropped independently of the wikisource fix.
- The `www` fallback is scoped to wikisource specifically (not any nil-language wiki) so wikidata — also nil-language, but rejected by the API — can't slip through if its exclusion ever changes.

### DIFF
--- a/lib/reference_counter_api.rb
+++ b/lib/reference_counter_api.rb
@@ -153,7 +153,11 @@ class ReferenceCounterApi
   end
 
   def batch_query_url
-    "/api/v1/references/#{@project_code}/#{@language_code}"
+    # Multilingual Wikisource (wikisource.org) is stored with language = nil;
+    # the reference-counter API reaches it as www.wikisource.org.
+    language = @language_code
+    language = 'www' if @project_code == 'wikisource' && language.nil?
+    "/api/v1/references/#{@project_code}/#{language}"
   end
 
   # A batch lookup has no reason to take longer than this. Without timeouts,
--- a/lib/reference_counter_api.rb
+++ b/lib/reference_counter_api.rb
@@ -16,14 +16,10 @@ class ReferenceCounterApi
   RETRY_COUNT = 5
   MAX_NON_200_RESPONSE_LOGS = 5
 
-  # The reference-counter Toolforge API only supports language-edition wikis
-  # (en.wikipedia, fr.wiktionary, etc.). Wikidata is excluded because it has
-  # its own data model and reference-counting approach. The wikimedia
-  # pseudo-project (commons, meta, incubator, species, foundationwiki) is
-  # excluded because none of its members are language-edition wikis with
-  # article-style <ref> markup; the API responds with 400 "Language X is not
-  # a valid language" for any of them.
-  UNSUPPORTED_PROJECTS = %w[wikidata wikimedia].freeze
+  # Wikidata is excluded because it has its own data model and
+  # reference-counting approach; the API rejects it server-side with
+  # "Project wikidata is not a valid project."
+  UNSUPPORTED_PROJECTS = %w[wikidata].freeze
 
   def self.valid_wiki?(wiki)
     !UNSUPPORTED_PROJECTS.include?(wiki.project)
--- a/spec/lib/reference_counter_api_spec.rb
+++ b/spec/lib/reference_counter_api_spec.rb
@@ -19,10 +19,10 @@ describe ReferenceCounterApi do
     end.to raise_error(described_class::InvalidProjectError)
   end
 
-  it 'raises InvalidProjectError for wikimedia projects (commons, meta, etc.)' do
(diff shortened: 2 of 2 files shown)


ragesoss (approved):
**Note on this feedback:** The review below was produced with Claude Code. It is not meant to be trusted or acted on automatically — it is offered as input to read, understand, and use or discard based on your own judgment.

This change was reviewed locally against the current master. Both changes read as correct and well-scoped.

**Correctness.** The nil-language handling is complete: tracing the `Wiki` model, `ensure_valid_project` only permits `language = nil` for `wikidata` and `wikisource`. Wikidata remains excluded at construction; multilingual Wikisource is now handled by the `www` fallback; and every wikimedia wiki that can actually exist carries a real subdomain-as-language (`commons`/`incubator`/`meta`), so no nil-language wikimedia wiki can reach `batch_query_url` and reintroduce the empty-segment 404. The `www` substitution is correctly scoped to `wikisource` so wikidata can't slip through even if its exclusion ever changed, and language-edition wikisource (`es.wikisource`) keeps its real language segment.

**Verification performed.** The PR's own spec passes (11/11) and RuboCop is clean on both changed files. An exploratory spec was written during review to pin the URLs the newly-supported wikimedia wikis produce — `commons`/`incubator`/`meta` each yield a well-formed `/api/v1/references/wikimedia/{lang}`, and wikidata still raises `InvalidProjectError`. That spec (AI-generated, run but not adopted here) confirmed the wikimedia side of the change behaves as intended.

**Scope.** Additive data-recovery change with the existing retry/Sentry error handling intact; no migration, security, performance, or UI surface. Low risk.

Approving.

*Reviewed in a Claude Code session (~15 min, ~2 interactions) using the `/review-pr` skill. Sage directed the review and approved posting after reading the findings. The correctness tracing through the `Wiki` model, the local spec run, and the exploratory-spec verification were performed by Claude Code and spot-checked rather than independently re-derived by the reviewer.*

ragesoss:
LGTM

PR #6957, opened 2026-07-13
Stop storing zero-valued counters in timeslice wikidata stats

gabina:
## What this PR does

Shrinks the serialized `stats` field on timeslice records by storing only non-zero counters. The `article_course_user_wiki_timeslices` table is one of the largest in the database, and a big share of that size comes from its `stats` text column: for Wikidata revisions, `UpdateWikidataStatsTimeslice#build_stats_from_revisions` pre-filled all 68 `STATS_CLASSIFICATION` labels with 0 before counting, producing a ~1.5&nbsp;KB YAML blob per row in which most values were zero (~40 bytes once zeros are stripped). For a single large Wikidata course with ~900k ACUWT rows, that is over 1&nbsp;GB of mostly-zero YAML.

With this change, `build_stats_from_revisions` and `sum_up_stats` omit zero-valued counters — an absent key means zero. This applies consistently to both write paths: ACUWT rows and `course_wiki_timeslices` stats (whether built directly from revisions or aggregated from ACUWT rows). `CourseStat.stats_hash`, which is what the UI reads, still receives the complete zero-filled hash via a template merge in `update_wikidata_statistics`, so nothing changes for the frontend.

Old dense rows and new sparse rows coexist safely: `sum_up_stats` iterates each input hash's own keys, so legacy zeros are simply dropped the next time a timeslice is aggregated. No migration is required for correctness.

## AI usage

This PR was developed with Claude Code, which wrote the code, specs, and commit message under human direction. The problem diagnosis started from the human side: Gabina observed that ACUWT and ACT row counts are similar for the same course while ACUWT is far larger, hypothesized the dense `stats` text field was the cause, and proposed storing only non-zero values. Claude Code verified the hypothesis against the schema and code (ACT has no `stats` column; the dense blob measures 1,482 bytes vs 41 bytes sparse for a representative hash), confirmed every reader of the field tolerates sparse hashes, and then implemented the change as directed — including the explicit requirements that ACUWT and CWT stay consistent and that `course_stats` keep the complete hash. Local testing to validate the size measurements against real data was done by hand.

The work was completed in a single focused session (one commit), with all tests green on the first run. This summary and the rest of the description were also drafted by Claude Code (via the `/prepare-pr` skill) and may contain errors.

## Screenshots

No UI changes.

- This only affects newly written rows. Existing rows keep their dense blobs until their timeslice is next reprocessed; actually reclaiming disk space on the database server will eventually require a backfill of existing `stats` values plus `OPTIMIZE TABLE` (or equivalent), which is left as follow-up work.
- The `stats` serializer is still YAML. Switching to JSON would save roughly another 30% and deserialize faster, but it has to handle existing YAML rows, so it was deliberately kept out of scope.

### DIFF
--- a/app/services/update_wikidata_stats_timeslice.rb
+++ b/app/services/update_wikidata_stats_timeslice.rb
@@ -107,9 +107,11 @@ class UpdateWikidataStatsTimeslice
     revisions
   end
 
-  # Given an array of revisions, it builds the stats for those revisions
+  # Given an array of revisions, it builds the stats for those revisions.
+  # Zero-valued counters are omitted to keep serialized timeslice stats small;
+  # absent keys mean zero.
   def build_stats_from_revisions(revisions)
-    stats = STATS_CLASSIFICATION.values.to_h { |label| [label, 0] }
+    stats = Hash.new(0)
     revisions.each do |revision|
       revision.diff_stats&.each do |key, value|
         # Skip non-counter fields the analyzer may include (e.g. merge_target).
@@ -118,12 +120,13 @@ class UpdateWikidataStatsTimeslice
       end
     end
     stats['total revisions'] = revisions.count
-    stats
+    stats.reject { |_label, count| count.zero? }
   end
 
   # Given an array of indivual stats, it creates or updates the CourseStats row for it.
   def update_wikidata_statistics(individual_stats)
-    stats = sum_up_stats individual_stats
+    # CourseStat keeps the complete hash (zeros included), since it's what the UI reads.
+    stats = full_stats_template.merge(sum_up_stats(individual_stats))
     crs_stat = CourseStat.find_by(course_id: @course.id) || CourseStat.create(course_id: @course.id)
 
     # Update the stats_hash in the CourseStat model and save it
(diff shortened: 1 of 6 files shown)


ragesoss (approved):
Reviewed with Claude Code. AI-driven review notes below are provisional — worth reading and verifying, not to be trusted or acted on blindly.

The change is sound. Every consumer of the timeslice `stats` field was traced: `ACUWT.stats` and both `CWT.stats` write paths (`build_stats_from_revisions` and `sum_up_stats`) produce sparse hashes, while `CourseStat.stats_hash` — the value the UI, CSV builders, and Salesforce read — stays dense via the new `full_stats_template.merge(...)`. Because every `sum_up_stats` key is a subset of the template and `merge` preserves the template's key order, `CourseStat.stats_hash` comes out effectively identical to before, which is the invariant the frontend depends on. The no-migration claim holds up: legacy dense rows contribute their explicit zeros harmlessly and drop out on the next aggregation, and `CourseStat` is re-derived dense on every course update.

Verified locally: the PR's own and affected-area specs pass (`update_wikidata_stats_timeslice_spec`, `course_wiki_timeslice_spec`, `update_course_wiki_timeslices_spec` — 61 examples, 0 failures), and RuboCop is clean on the changed files. The regenerated `expected_timeslices.yml` fixtures are a genuine end-to-end check rather than orphaned data — `compare_course_wiki_timeslices` asserts them against real VCR-driven pipeline output with a full `match_array`, including the `stats` hash.

One minor, non-blocking observation: zero-revision Wikidata timeslices now serialize `stats == {}` (the `'total revisions' => 0` key is dropped, as the updated spec asserts). No current reader depends on that key's presence, so this is safe — noting it only in case future code assumes `stats` always carries `'total revisions'`.

*Reviewed in a short Claude Code session (~10 min, a couple of interactions). Sage directed the review; Claude Code ran the specs and RuboCop locally and traced the stats-field consumers. Sage reviewed this summary before approving but did not independently re-verify each finding.*

PR #6958, opened 2026-07-15
Reimplement ArticleNamespacesManager article resets for ACUWT courses

gabina:
## What this PR does

Makes `ArticleNamespacesManager` (the article reset logic introduced in #6950) work natively for ACUWT courses, using only the `article_course_user_wiki_timeslices` table plus the articles/articles_courses records — no full-period re-fetches and no reliance on `created_at` ordering that ACUWT reaggregation does not preserve. Each of the manager's four cases gets an ACUWT-appropriate mechanism, always gated on `course.use_acuwt?` with the legacy behavior preserved for non-ACUWT courses:

**1.** **Moved into a tracked namespace**: the legacy heuristic (articles_courses record created after some of its ACT rows) breaks under ACUWT — reaggregation rebuilds ACT rows, and an ACUWT-timestamp condition would either miss real moves or match forever and reset the article on every update. Instead, detection now happens at the moment it's observable: when `ArticlesCourses.update_from_course_revisions` creates a record for an article that already has ACUWT rows, those rows are marked `needs_update` (new `ArticlesCourses.create_records_and_mark_acuwt`, one extra UPDATE only when new records are inserted). `ReprocessArticleCourseUserWikiTimeslices` re-scores just those rows and reaggregates. Fires exactly once by construction.
**2.** **Moved to an untracked namespace** and **3.** **Deleted**: new `ArticlesCoursesCleaner#reset_excluded` — the article's stored ACUWT data is already correct, so exclusion only needs `reset_timeslices_for_reaggregation_from_acuwt` (DB-only) plus articles_courses deletion. ACUWT rows are kept so re-inclusion stays cheap.
**4.** **Undeleted / re-tracked**: new `ArticlesCoursesCleaner#reset_included` delegates to `create_records_and_mark_acuwt` — re-inclusion is exactly the "record created for an article with existing ACUWT history" case. Its retrieval switches to a new ACUWT-based `Course#articles_from_timeslices`.

Note that the old ACT-based retrieval query survives as `Course#articles_from_timeslices_legacy`. `TimesliceCleaner` uses it in three places, to decide which article-course-timeslice rows to delete during course/wiki/user cleanup, and stays on it permanently — that decision needs ACT semantics regardless of `use_acuwt?`.

Along the way, `UpdateTimeslicesScopedArticle`'s scope handling for `ArticleScopedProgram`/`VisitingScholarship` courses is unified onto the same mechanisms:
- **Scope entry**: its ~60-line bespoke fetch/score/upsert path (a near-duplicate of `ReprocessArticleCourseUserWikiTimeslices`) is deleted in favor of `create_records_and_mark_acuwt`, consumed in the same update cycle. Newly scoped ids are filtered to tracked namespaces first — assignments keep their `article_id` when an article moves namespace (see `AssignmentUpdater`), so a draftified assigned article stays in scope, and without the filter it would get its articles_courses record created here and deleted by `ArticleNamespacesManager` on every update (the old flow had a perpetual-rescore variant of that loop).
- **Scope exit**: now reuses `ArticlesCoursesCleaner#reset_excluded` instead of hand-rolling the same reaggregation-reset sequence — an article leaving scope is the same "exclude, keep ACUWT for cheap re-inclusion" situation as case 2/3 above.

Also fixed: for ACUWT courses, deleted articles' ACUWT rows were inflating `CourseWikiTimeslice` revision counts and wikidata stats forever, since exclusion resets (case 2/3 above) deliberately keep ACUWT rows and a full re-fetch only ever upserts rows, never deleting stale ones — unlike the legacy path, where counts are recomputed from freshly fetched revisions and deleted pages simply disappear from the replica. `update_revision_count_from_acuwt`, `update_mw_rev_count_from_acuwt`, and `update_stats_from_acuwt` now join `articles` and filter `deleted: false`, matching the filter the character/reference-count aggregation already applied.

## AI usage

All of the code, specs, and commit messages in this PR were written by Claude Code, working under close direction from the author. The design emerged from an extended back-and-forth: the author framed each problem (starting from "the moved_to_mainspace strategy won't survive reaggregation"), questioned or rejected several of Claude's intermediate proposals (an opt-out keyword argument for the scoped flow was replaced by full unification at the author's suggestion; the tracked-namespace filter was interrogated down to the assignments-keep-article_id-through-moves invariant before being kept; a proposed dedup of the new-articles scoped path onto `reset_included` was examined and rejected as not a real reuse opportunity), and decided the commit structure. Each commit message contains a `## Process` section describing that commit's specific back-and-forth.

The work spans 9 commits across two sessions: the bulk of the design and implementation (7 commits) in a single session on 2026-07-14, plus two smaller follow-up commits (the deleted-articles count/stats fix and the scope-exit dedup) added after review, days later. Enabling refactors (`reset` → `reset_legacy`, the `articles_from_timeslices` split) are committed separately from the behavior changes they support.

This PR description was also drafted by Claude Code (via the `/prepare-pr` skill) and may contain errors.

## Screenshots

No UI changes.

- For ACUWT courses, re-included (undeleted/re-tracked) articles are Sentry-logged twice, once by the manager (`undeleted_or_retracked`) and once by the primitive (`created_with_preexisting_acuwt_history`). Both are info-level frequency-learning telemetry; left as is.
- `ArticleStatusManager` still retrieves via `articles_from_timeslices_legacy`; switching it to the ACUWT-aware query is deferred to a separate PR. It's currently a dead code path for ACUWT either way: `UpdateCourseStats#should_update_article_status?` never runs the manager for `only_scoped_articles_course?` courses, and every ACUWT course today is scoped — so this only matters once a non-scoped course adopts ACUWT.
- The `create_records_and_mark_acuwt` Sentry log reuses the existing `'Article retracked'` message with a new reason (`created_with_preexisting_acuwt_history`), so existing Sentry queries keep aggregating these events.


gabina:
To handle in the following commits (already done):

Deleted articles still contribute to ACUWT-derived revision counts

For ACUWT courses, CourseWikiTimeslice#update_revision_count_from_acuwt and #update_mw_rev_count_from_acuwt sum ACUWT rows without joining against the articles table, so revisions to articles marked deleted: true keep counting toward revision_count and mw_rev_count after the article is excluded from the course (exclusion deliberately keeps ACUWT rows so undeleted articles can be re-included without a full re-fetch). The legacy path dropped these revisions naturally, since deleted pages disappear from the replica database and a full re-fetch never sees them. Character and reference sums are not affected — acuwt_mainspace_tracked_student_records already filters articles: { deleted: false }.

A possible fix is adding the same deleted: false join to those two sum queries. mw_rev_count needs extra care: it must stay comparable to the live fetched-revision count used by CourseRevisionUpdater#new_revisions? (see the existing approximation note on update_mw_rev_count_from_acuwt).

gabina:

ragesoss (approved):

Approving. The four case-by-case mechanisms hold up under tracing, and the invariants the redesign rests on were checked against the code rather than taken from the description.

**What was verified**

- **Single choke point for `articles_courses` creation.** Every insertion path in `app/` and `lib/` now funnels through `create_records_and_mark_acuwt` → `maybe_insert_new_records`; `CourseRevisionUpdater#create_articles_courses` is the only other entry. This is what makes "detect the move at creation time" sound as a replacement for a queryable heuristic — a bypass path would make a missed move permanently undetectable, and there isn't one.

- **When each mark is consumed.** `ArticleNamespacesManager` (`UpdateCourseStats#update_articles`) and `UpdateTimeslicesScopedArticle` (`PrepareTimeslices`, via `pre_update`) both run before the per-wiki loop, so their marks are consumed by `ReprocessArticleCourseUserWikiTimeslices` in the same cycle, as the description claims. The namespace-move marking in `update_from_course_revisions` happens during step-3 ingestion, which is after that wiki's reprocess pass, so that one lands on the following cycle. A one-cycle delay rather than a lost mark, and no worse than the legacy heuristic, which also only fired on the update after the record appeared.

- **No self-inflicted rework.** `create_articles_courses` runs before `bulk_upsert_from_revisions` within a timeslice, so a genuinely new article's just-written rows are not marked (`marked` is zero, no Sentry event). Separately, `bulk_upsert_from_revisions` lists `needs_update` in `update_only`, so an upsert can overwrite a pending mark — but `should_update_timeslice?` skips the upsert entirely when `only_new: true` and there is no new data, so a mark cannot be silently cleared by a no-op ingestion. Worth keeping in mind if that gate is ever relaxed.

- **The `deleted: false` change is complete rather than half-applied.** `CourseUserWikiTimeslice#update_cache_from_acuwt` already filtered deleted articles in both its character sums (`records_by_namespace`) and its revision count (`filtered_live_acuwt_records`), so `CourseWikiTimeslice` was the last remaining gap; the three methods this PR touches close it without leaving a sibling aggregate inconsistent.

- **`reset_excluded` does not leave stale rows behind.** `reset_timeslices_for_reaggregation_from_acuwt` deletes the ACT and CUWT rows for the affected pairs, so reaggregation rebuilds them and only the ACUWT rows are deliberately retained.

**Notes, none blocking**

- **The branch no longer merges cleanly.** Master's #6962 rewrote `UpdateTimeslicesScopedArticle#fetch_scores_and_update_acuwt`, the same ~60 lines this PR deletes. Taking the deletion appears to be the correct resolution: the surviving copy in `ReprocessArticleCourseUserWikiTimeslices` merges cleanly from master and already carries #6962's per-user/per-timeslice improvements, and `periods_for_articles` ends up with no remaining callers. That resolution is what the results below were produced against.

- **Case 4 fires exactly once because each update builds a fresh `Course`, not because of DB state.** `reset_undeleted_or_retracked_articles` computes `tracked - @course.articles.to_a`, which loads and caches the `articles` association *before* `reset_included` creates the new records. Running the manager twice against the same `Course` instance re-fires the case and re-marks the ACUWT rows on every pass; adding `course.reload` between passes makes it idempotent. Production looks safe, since `UpdateCourseStats` constructs the manager once per update against a freshly loaded course. Flagging it because every other case in this class fires once for a reason that survives object reuse (the `articles_courses` row is gone, or the article is filtered by `deleted`/namespace), and this one does not — so it would be easy to break later without any test noticing.

- **The `revision_count` fix changes numbers that are already on screen.** For ACUWT courses with deleted articles, `revision_count` and `mw_rev_count` will drop as timeslices reaggregate. That appears intended, but it may surface as an apparent regression report before anyone connects it to this PR. It also reads as a performance fix beyond the stats correctness: `new_revisions?` compares the fetched count against `mw_rev_count`, so an inflated `mw_rev_count` made that comparison mismatch every cycle and reprocess those timeslices indefinitely.

- **The largest remaining coverage gap is the wiring, not the units.** Each new mechanism has both-branch specs, but the marking → `ReprocessArticleCourseUserWikiTimeslices` → reaggregation handoff is not exercised end-to-end through `UpdateCourseStats`. That ordering is the load-bearing assumption of the whole design and was confirmed here by reading the pipeline, not by executing it. If it were wrong, the symptom would be silent — revisions staying unscored and stats drifting, with nothing raising — across all ACUWT courses.

- **Smaller items.** `reset_included(articles, wiki = nil)` ignores `wiki` in the ACUWT branch, since `create_records_and_mark_acuwt` marks course-wide; harmless while an article has exactly one wiki, but the signature implies a filter that is not applied. And `articles_from_timeslices` moves the `wiki_id` filter from `articles.wiki_id` to `article_course_user_wiki_timeslices.wiki_id`, which is equivalent only as long as those two never diverge.

The three items in the PR's own "open questions" section — the double Sentry log on re-inclusion, `ArticleStatusManager` still on legacy retrieval, and the reused `'Article retracked'` message — all read as reasonable calls as described.

**Exploratory spec**

The PR's specs cover one pass through each case, so nothing pinned the "fires exactly once by construction" claim. The spec below runs the manager twice across all four cases plus an exclude → undelete → re-include round trip, and is what surfaced the association-cache note above. It passes (6 examples) and is RuboCop clean. It was written by Claude Code and has not been line-by-line reviewed by a human; adopt or discard as useful.

<details>
<summary><code>spec/lib/article_namespaces_manager_acuwt_idempotence_spec.rb</code></summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/article_namespaces_manager"

describe ArticleNamespacesManager do
  describe 'ACUWT reset idempotence' do
    let(:start) { '2024-01-01'.to_datetime }
    let(:course) { create(:course, start:, end: '2024-04-20', home_wiki: enwiki) }
    let(:manager) { TimesliceManager.new(course) }
    let(:user) { create(:user) }

    # Case 3: tracked article that has just been detected as deleted on-wiki.
    let(:deleted_article) { create(:article, wiki: enwiki, namespace: 0, deleted: true) }
    # Case 2: article that has just moved out of a tracked namespace.
    let(:untracked_article) { create(:article, wiki: enwiki, namespace: 118) }
    # Case 4: tracked, live article with ACUWT history but no articles_courses record.
    let(:retracked_article) { create(:article, wiki: enwiki, namespace: 0) }

    def acuwt_for(article)
      ArticleCourseUserWikiTimeslice.where(course:, article:)
    end

    def cwt_at(date)
      course.course_wiki_timeslices.find_by(wiki: enwiki, start: date)
    end

    # Simulates the reprocess/reaggregation pass consuming the flags this manager set,
    # so a second manager run starts from a clean slate the way the next update would.
    # The reload matters: reset_undeleted_or_retracked_articles reads `@course.articles.to_a`,
    # which caches the association, and each real update builds a fresh Course object.
    def consume_flags
      # rubocop:disable Rails/SkipsModelValidations
      ArticleCourseUserWikiTimeslice.where(course:).update_all(needs_update: false)
      course.course_wiki_timeslices.update_all(needs_reaggregation: false, needs_update: false)
      # rubocop:enable Rails/SkipsModelValidations
      course.reload
    end

    before do
      stub_wiki_validation

      create(:articles_course, course:, article: deleted_article)
      create(:articles_course, course:, article: untracked_article)
      # retracked_article deliberately has no articles_courses record

      [deleted_article, untracked_article, retracked_article].each do |article|
        create(:article_course_user_wiki_timeslice, course:, article:, user:, wiki: enwiki,
               start: '2024-02-01', end: '2024-02-02', revision_count: 2)
      end

      manager.create_timeslices_for_new_course_wiki_records([enwiki])
    end

    it 'excludes a deleted article on the first pass only' do
      described_class.new(course, statuses_synced: true)

      expect(course.articles_courses.where(article: deleted_article)).to be_empty
      expect(acuwt_for(deleted_article).count).to eq(1)
      expect(cwt_at('2024-02-01').needs_reaggregation).to eq(true)

      consume_flags
      described_class.new(course, statuses_synced: true)

      # The articles_courses record must not come back, and nothing may be re-flagged:
      # the article is gone from course.articles, and the re-inclusion pass filters
      # deleted articles out.
      expect(course.articles_courses.where(article: deleted_article)).to be_empty
      expect(acuwt_for(deleted_article).pluck(:needs_update)).to eq([false])
      expect(cwt_at('2024-02-01').needs_reaggregation).to eq(false)
    end

    it 'excludes an untracked-namespace article on the first pass only' do
      described_class.new(course, statuses_synced: true)

      expect(course.articles_courses.where(article: untracked_article)).to be_empty
      expect(acuwt_for(untracked_article).count).to eq(1)

      consume_flags
      described_class.new(course, statuses_synced: true)

      expect(course.articles_courses.where(article: untracked_article)).to be_empty
      expect(acuwt_for(untracked_article).pluck(:needs_update)).to eq([false])
      expect(cwt_at('2024-02-01').needs_reaggregation).to eq(false)
    end

    it 'marks a retracked article for rescoring on the first pass only' do
      described_class.new(course, statuses_synced: true)

      expect(course.articles_courses.where(article: retracked_article)).not_to be_empty
      expect(acuwt_for(retracked_article).pluck(:needs_update)).to eq([true])

      consume_flags
      described_class.new(course, statuses_synced: true)

      # Now that the articles_courses record exists, the re-inclusion pass must skip it.
      expect(acuwt_for(retracked_article).pluck(:needs_update)).to eq([false])
    end

    it 'logs the retracked article to Sentry once across two passes' do
      allow(Sentry).to receive(:capture_message)

      described_class.new(course, statuses_synced: true)
      consume_flags
      described_class.new(course, statuses_synced: true)

      expect(Sentry).to have_received(:capture_message)
        .with('Article retracked', hash_including(
                extra: hash_including(reason: 'undeleted_or_retracked',
                                      article_ids: [retracked_article.id])
              )).once
    end

    it 'keeps ACUWT history through exclusion so a later undeletion re-includes cheaply' do
      described_class.new(course, statuses_synced: true)
      consume_flags

      # The article is restored on-wiki; ArticleStatusManager would clear the flag.
      deleted_article.update(deleted: false)
      described_class.new(course, statuses_synced: true)

      expect(course.articles_courses.where(article: deleted_article)).not_to be_empty
      # Re-inclusion reuses the retained ACUWT row rather than requiring a full re-fetch.
      expect(acuwt_for(deleted_article).pluck(:needs_update)).to eq([true])
      expect(acuwt_for(deleted_article).first.revision_count).to eq(2)
    end
  end

  describe 'ACUWT namespace-move detection idempotence' do
    let(:course) { create(:course, start: '2024-01-01', end: '2024-04-20', home_wiki: enwiki) }
    let(:user) { create(:user) }
    let(:article) { create(:article, wiki: enwiki, namespace: 0) }
    let(:revision) do
      build(:revision_on_memory, article_id: article.id, user_id: user.id, wiki_id: enwiki.id,
            mw_rev_id: 12_345, mw_page_id: article.mw_page_id,
            date: '2024-02-01'.to_datetime, scoped: true)
    end

    before do
      stub_wiki_validation
      create(:courses_user, course:, user:)
      # ACUWT row ingested while the article was still in an untracked namespace.
      create(:article_course_user_wiki_timeslice, course:, article:, user:, wiki: enwiki,
             start: '2024-02-01', end: '2024-02-02', revision_count: 1)
    end

    it 'marks preexisting ACUWT rows once, not on every ingestion' do
      ArticlesCourses.update_from_course_revisions(course, [revision])
      expect(ArticleCourseUserWikiTimeslice.where(course:, article:)
                                           .pluck(:needs_update)).to eq([true])

      # rubocop:disable Rails/SkipsModelValidations
      ArticleCourseUserWikiTimeslice.where(course:).update_all(needs_update: false)
      # rubocop:enable Rails/SkipsModelValidations

      # Same revision ingested again on a later update: the articles_courses record now
      # exists, so the article is no longer a "new" article and must not be re-marked.
      ArticlesCourses.update_from_course_revisions(course, [revision])

      expect(ArticleCourseUserWikiTimeslice.where(course:, article:)
                                           .pluck(:needs_update)).to eq([false])
    end
  end
end
```

</details>

**Results**

Run against the branch merged with master, using the conflict resolution described above:

- RuboCop clean on all 16 changed files.
- The PR's own specs and the specs for affected areas pass (188 examples, 0 failures).
- Full Ruby suite passes: 2946 examples, 0 failures, 14 pending — the pending ones are pre-existing unimplemented stubs and the network-dependent PetScan/PagePile specs.
- Exploratory idempotence spec passes (6 examples).
- No migrations, and no JS/CSS/asset files in the diff.

---

*Reviewed in a Claude Code session (~40 min, 3 interactions) using the repo's `/review-pr` skill. Claude Code read the full diff, traced the update-pipeline ordering and the `articles_courses` creation paths, resolved the master merge conflict locally, wrote and ran the exploratory idempotence spec, and ran RuboCop, the affected specs, and the full suite. Sage directed the review phases and chose to approve after reading the findings, but did not independently re-derive the call-path tracing or review the exploratory spec line by line. The pipeline-ordering conclusions come from code reading, not from an executed integration test — that gap is noted above as the main untested area.*

PR #6960, opened 2026-07-16
Gsoc26/system stats frontend

lakshita10341:
## What this PR does
This pull request introduces the new System & Facilitator Stats Dashboard for administrators. The dashboard simplifies how administrators monitor overall system performance and facilitator engagement by serving pre-computed weekly and daily database snapshots directly, eliminating heavy real-time table queries.

### Key Changes
1. **Frontend Dashboard Page (`system_stats_handler.jsx` & `_system_stats.styl`)**:
   - Added a new, responsive page showing key KPI counters (Total Edits, Article Views, Articles Created, Articles Improved, Characters Added, New Editors, Active Programs, Active Facilitators).
   - Added a "Monthly Activity Trends" chart visualizing system-wide activity metrics over the past 12 months.
   - Added a "Wiki Trends (Top 5)" multi-line chart highlighting activity across the top 5 home wikis.
   - Added a "Wiki Stats Breakdown" table. Columns are fully sortable, and client-side pagination (10 rows per page) is handled using the standard `react-paginate` component.
   - Added a "Facilitator Leaderboard" table displaying the top 100 facilitators. Columns are fully sortable, and client-side pagination (10 rows per page) is handled using the standard `react-paginate` component.

2. **Backend API endpoints (`SystemStatsController` & `routes.rb`)**:
   - Created three decoupled, high-performance API endpoints to serve data asynchronously:
     - `GET /system_stats.json`: Returns the overall KPI stats and monthly activity trends.
     - `GET /system_stats/wiki_trends.json`: Returns the historical wiki metrics and top 100 wiki stats.
     - `GET /system_stats/facilitators.json`: Serves the top 100 facilitator statistics.

3. **Specs (`system_stats_controller_spec.rb`)**:
   - Implemented comprehensive request specs verifying authentication/authorization (admin-only access) and JSON response payloads for all three endpoints.

## AI usage
I used AI:
-  to refactor the React frontend charts, implement robust Canvas-mounting hooks, style the layout with custom select dropdown arrows in Stylus, and optimize the backend queries.
- AI was also used to prepare unit test cases 

## Screenshots
image
620280700-be79b17a-5061-41ea-b452-66f76a32f127

### DIFF
--- a/app/assets/javascripts/components/system_stats/facilitator_leaderboard.jsx
+++ b/app/assets/javascripts/components/system_stats/facilitator_leaderboard.jsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState, useEffect, useMemo } from 'react';
 import ReactPaginate from 'react-paginate';
 import request from '../../utils/request';
 
@@ -48,15 +48,16 @@ const FacilitatorLeaderboard = () => {
     setCurrentPage(0);
   };
 
-  const sortedFacilitators = [...(facilitators || [])]
-    .sort((a, b) => {
-      let valA = a[sortField];
-      let valB = b[sortField];
+  const sortedFacilitators = useMemo(() => {
+    return [...(facilitators || [])].sort((a, b) => {
+      const valA = a[sortField] ?? 0;
+      const valB = b[sortField] ?? 0;
       if (typeof valA === 'string') {
         return sortOrder === 'asc' ? valA.localeCompare(valB) : valB.localeCompare(valA);
       }
       return sortOrder === 'asc' ? valA - valB : valB - valA;
     });
+  }, [facilitators, sortField, sortOrder]);
 
   const totalPages = Math.max(1, Math.ceil(sortedFacilitators.length / ROWS_PER_PAGE));
   const startIdx = currentPage * ROWS_PER_PAGE;
--- a/app/assets/javascripts/components/system_stats/facilitator_leaderboard.jsx
+++ b/app/assets/javascripts/components/system_stats/facilitator_leaderboard.jsx
@@ -1,0 +1,153 @@
+import React, { useState, useEffect } from 'react';
+import ReactPaginate from 'react-paginate';
+import request from '../../utils/request';
+
+const ROWS_PER_PAGE = 10;
+
+const SortIndicator = ({ field, sortField, sortOrder }) => {
+  if (sortField !== field) return null;
+  const indicatorClass = sortOrder === 'asc' ? 'sortable-indicator-asc' : 'sortable-indicator-desc';
+  return <span className={`${indicatorClass} ${sortOrder}`} />;
+};
+
+const FacilitatorLeaderboard = () => {
+  const [facilitators, setFacilitators] = useState(null);
+  const [loading, setLoading] = useState(true);
+  const [sortField, setSortField] = useState('edits');
+  const [sortOrder, setSortOrder] = useState('desc');
+  const [error, setError] = useState(null);
+  const [currentPage, setCurrentPage] = useState(0);
+
+  useEffect(() => {
+    setLoading(true);
+    request('/system_stats/facilitators.json')
+      .then(resp => {
+        if (!resp.ok) {
+          throw new Error('Failed to fetch facilitators');
+        }
+        return resp.json();
+      })
+      .then(data => {
+        setFacilitators(data.facilitators);
+        setLoading(false);
+      })
+      .catch(err => {
+        console.error(err);
+        setError(I18n.t('system_stats.errors.fetch_failed'));
+        setLoading(false);
+      });
+  }, []);
+
+  const handleSort = (field) => {
+    if (sortField === field) {
+      setSortOrder(sortOrder === 'asc' ? 'desc' : 'asc');
+    } else {
+      setSortField(field);
+      setSortOrder('desc');
+    }
+    setCurrentPage(0);
+  };
+
+  const sortedFacilitators = [...(facilitators || [])]
+    .sort((a, b) => {
+      let valA = a[sortField];
+      let valB = b[sortField];
+      if (typeof valA === 'string') {
+        return sortOrder === 'asc' ? valA.localeCompare(valB) : valB.localeCompare(valA);
+      }
+      return sortOrder === 'asc' ? valA - valB : valB - valA;
+    });
+
+  const totalPages = Math.max(1, Math.ceil(sortedFacilitators.length / ROWS_PER_PAGE));
+  const startIdx = currentPage * ROWS_PER_PAGE;
+  const paginatedFacilitators = sortedFacilitators.slice(startIdx, startIdx + ROWS_PER_PAGE);
+
+  return (
+    <div id="facilitators" className="container">
+      <div className="section-header system-stats__leaderboard-header">
+        <h2>{I18n.t('system_stats.facilitators.title')}</h2>
+      </div>
+      {error && (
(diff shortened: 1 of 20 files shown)


ragesoss (commented):
**How to read this feedback:** the notes below were produced with Claude Code during review. They are observations to weigh with your own judgment, not authoritative conclusions — use, verify, or discard as you see fit.

Reviewed locally: `yarn build` clean, RuboCop/ESLint clean on changed files, the PR's specs pass (9 examples), and an exploratory feature spec written during review passes (7 examples covering KPI formatting, chart mounting, sorting, pagination, empty states, and auth gating). The dashboard works well end-to-end. No migrations, no new JS dependencies.

**1. "Monthly Activity Trends" plots cumulative totals.** The snapshot values are running all-time totals (`Course.nonprivate.sum(:revision_count)` etc. in `SystemStatUpdateWorker`), and `trends_for` serves the raw month-end samples — so the chart shows an ever-rising cumulative curve rather than per-month activity, despite the title. The chart also mixes point-in-time gauges (`active_programs_count`, `active_facilitators_count`) with cumulative counters. Consider serving month-over-month deltas for the cumulative metrics, or relabeling the chart.

**2. `recent_monthly_snapshots` scales poorly as snapshots accumulate.** Snapshots are daily, so this loads ~365 rows per year — each including the serialized `wiki_stats` blob — and deserializes them all just to keep the last row per month; both JSON endpoints do this independently. Selecting only the needed month-end rows (or deferring `wiki_stats` where it isn't used) would keep these endpoints fast long-term. Also, `SystemStat.recent_months` already exists for the date window.

**3. `activeInYear: s.active_in_last_year ? 'Yes' : 'No'`** hardcodes English strings in the controller, and the frontend then sorts on them as strings. Serving the boolean and rendering the label client-side with `I18n.t` keeps it localizable.

**4. Wiki Ed will see a permanently empty dashboard.** `SystemStatUpdateWorker` no-ops when `Features.wiki_ed?`, but the admin-index link and route ship to both deployments. Consider gating the link on the P&E dashboard.

**Minor:** `system_stats.pagination.page_info` i18n key is unused; `.system-stats__search-input` and `.system-stats__leaderboard-actions` are dead CSS from the removed filter UI; the `data-current_user` attribute on `#react_root` copies a vestigial pattern (`revision_ai_scores_stats` uses a bare `#react_root`); `strftime('%b %Y')` month labels are English-only; a failed `wiki_trends` fetch only logs to the console while a failed KPI fetch shows the error banner.

The exploratory feature spec is available if useful — happy to share it for adoption, since the page's rendering, sorting, and pagination behavior currently have no committed coverage.

*Drafted in a Claude Code session (~45 min, ~3 interactions). Sage directed the review scope and read the full comment before posting. Verification was done by Claude Code locally: build, linters, the PR's specs, and an AI-generated 7-example exploratory feature spec that was run (all passing) but the findings above were not all independently re-verified by Sage.*

ragesoss:
I think we may as well enable this for Wiki Education Dashboard as well.

ragesoss (commented):
**How to read this feedback:** the notes below were produced with Claude Code during re-review. They are observations to weigh with your own judgment, not authoritative conclusions — use, verify, or discard as you see fit.

Re-reviewed commit 0d3fde6b6 locally: `yarn build` clean, RuboCop/ESLint clean on changed files, the PR's specs pass (22 examples), and 9 exploratory examples written during re-review pass (delta math, month/series alignment, localized leaderboard rendering, and boolean sorting in a real browser). The cumulative-trends, localization, month-label, error-banner, and cleanup items from the previous review are all addressed — the delta calculation checks out against multi-month fixtures.

**Remaining:**

1. `FacilitatorStatUpdateWorker` (line 13) still has `return if Features.wiki_ed?`. With the system-stat guard removed but this one left in place, the Facilitator Leaderboard will remain permanently empty on the Wiki Ed deployment. Removing it (and updating its spec the same way as `system_stat_update_worker_spec.rb`) would complete the Wiki Ed enablement.
2. `wiki_trends.json` still loads and deserializes every daily snapshot's `wiki_stats` blob (~395 rows for the 13-month window) to keep ~13 month-end rows. Not blocking for an admin-only page, but worth a follow-up as snapshots accumulate.

**New observations on the latest commit:**

1. The non-admin example in `spec/features/system_stats_spec.rb` cannot fail: `expect(page).not_to have_content('System & Facilitator Stats')` checks for text rendered by React, which doesn't execute under the default rack_test driver — so the assertion passes regardless of whether access is denied. Asserting on the response status/redirect (as the request specs already do) or running with `js: true` would make it meaningful. The admin example likewise only asserts `#react_root` exists, so none of the seeded stat records are exercised.
2. Until the first month boundary after deployment, the trends chart renders the all-time cumulative total as a single "monthly" data point — on Wiki Ed that would be tens of millions of edits. It self-corrects once a second month-group of snapshots exists; noted here so it isn't mistaken for a bug when first enabled.
3. `calculate_delta` clamps negative deltas to zero. Cumulative totals can legitimately decrease (courses deleted or made private, cache recalculation); such months will display 0. Likely the right display choice, but worth being aware it masks corrections.
4. Minor: `formatMonthLabel` is duplicated across both chart components, and the baseline-month-dropping logic (`length > 1 ? 1 : 0`) is repeated in three controller methods — the `months` array will silently misalign with the trend series if these ever diverge.

*Drafted in a Claude Code session (~25 min, 2 interactions). Verification was done by Claude Code locally: build, linters, the PR's 22 spec examples, and 9 AI-written exploratory examples (7 request-spec examples on the delta math, 2 browser examples on leaderboard localization/sorting and chart mounting). Sage read the full comment before approving posting; findings were not independently verified by hand.*

lakshita10341:
@ragesoss PTAL when you get time!

ragesoss:
**How to read this feedback:** the notes below were produced with Claude Code during re-review. They are observations to weigh with your own judgment, not authoritative conclusions.

Re-reviewed commit 2c729967c locally: `yarn build` clean, RuboCop/ESLint clean on changed files, the PR's 33 spec examples pass, and the pre-existing `system_stat_spec.rb` (untouched by the PR but covering the modified model) passes. All four actionable items from the previous review are resolved: the `FacilitatorStatUpdateWorker` Wiki Ed guard is removed with its spec updated, `recent_monthly_snapshots` no longer loads daily `wiki_stats` blobs, the non-admin feature-spec example now asserts a real 401, and both the `formatMonthLabel` and baseline-month-drop duplication are centralized.

The reworked `recent_monthly_snapshots` was verified with 6 AI-written exploratory examples (run, all passing): month-end row selection, ordering, window exclusion, blob deferral (accessing `wiki_stats` on the default path raises `MissingAttributeError`, confirming the optimization is real), and an end-to-end check that `/system_stats.json` delta math still lands on month-end values when fed daily snapshots. Since this method has no committed unit coverage, adopting something like these examples into `spec/models/system_stat_spec.rb` would be worthwhile — happy to share the file.

Two minor notes, neither blocking:
1. In both chart components, the new `import { formatMonthLabel }` statement sits mid-file where the deleted local function was. It works (imports hoist) and ESLint doesn't flag it, but placing it with the other imports at the top would match convention.
2. The admin-side feature spec example still only asserts `#react_root` exists, so the page's rendering/sorting/pagination behavior has no committed browser-level coverage. Noted since the first review; acceptable for an admin-only page.

*Drafted in a Claude Code session (~20 min, 2 interactions). Verification was done by Claude Code locally: build, linters, the PR's 33 spec examples, the pre-existing model spec, and 6 AI-written exploratory examples on the reworked snapshot query. Sage read the full comment before approving posting; findings were not independently verified by hand.*

PR #6962, opened 2026-07-16
Fetch new users' revisions per user and per timeslice in ACUWT path

gabina:
## What this PR does

Makes the ACUWT "new users" step of the course update survivable for huge courses, prompted by `Wikidata/Coordinate_Me_2026_DE_(2026)` (course 36020 on the P&E Dashboard), which had not completed an update since May 12: its 40 newly added users had 423k+ Wikidata revisions in the course window (three users with 100k+ each), and `UpdateTimeslicesCourseUser` fetched and processed all of them in a single un-checkpointed pass over the whole course period. The worker died before finishing, `courses_users.created_at` was never ratcheted, so the same users were treated as "new" again on every subsequent update — an infinite restart-from-zero loop.

Two changes:

1. **Per-user, per-timeslice processing in the ACUWT new-users path.** `add_user_ids_acuwt` now processes one user at a time and marks each as no-longer-new as soon as all their wikis complete, so an interrupted update never redoes finished users. Within a user, revisions are fetched one `CourseWikiTimeslice` at a time (bounds match the main update loop, including the `end - 1.second` adjustment for Replica's inclusive bounds, and iteration stops at the ingestion limit), so a single fetch/processing pass is never bigger than what the adaptive timeslice splitting already determined the course can handle. If a wiki fails for a user, the error is logged to Sentry (course/user/wiki) and that user's remaining wikis are skipped — they stay "new" and are fully retried next update. `CourseRevisionUpdater#fetch_revisions_for_new_users` early-returns on empty fetches so the many empty slices cost nothing beyond the Replica call.

2. **Skip parent revision lookups for wikis with no scoring API.** On Wikidata, `RevisionScoreImporter#get_revision_scores` made a real `parent_revision_ids` MediaWiki API call per 50-revision batch even though both reference-counter lookups return `{}` there — the parent ids only ever keyed into an always-empty parent-scores hash. A new `RevisionScoreApiHandler#scores_available?` predicate lets the importer return the revisions untouched before the batch loop, eliminating thousands of sequential no-op API calls per update on high-volume Wikidata courses.

## AI usage

This PR was developed with Claude Code, which wrote all of the code, specs, and commit messages, as well as this PR description, under direction from the human author. The session started from a production investigation: the human pasted syslog output from the stuck course update along with their hypothesis about the cause; Claude Code confirmed the hypothesis by reading the code and probing the course's public data (quantifying the 423k+ revision workload), and proposed fixes. The key design decisions were the human's: fixing incremental progress first, fetching per real `CourseWikiTimeslice` record rather than fixed windows (so split timeslices are respected), bailing out of a user's remaining wikis on first failure instead of continuing, and adding course/wiki ids to the Sentry log. Both commits were made within a single session on 2026-07-16, about half an hour apart. The analysis above was drafted by Claude Code and may contain errors; this description was prepared with the `/prepare-pr` Claude Code skill.

## Screenshots

No UI changes (backend-only).

- The per-timeslice fetch trades a few giant Replica requests for many small ones (one per user per timeslice, most returning empty). For Coordinate Me this is clearly the right trade — the giant requests are what killed the worker — but it does add request volume for courses with many timeslices and inactive new users.
- A related, larger optimization is deliberately out of scope: `WikidataDiffAnalyzer` still runs on all live revisions (not just scoped ones) to detect merges into in-scope items, which dominates the remaining cost for scoped Wikidata courses. An issue will be filed to handle that separately.
- The reaggregation pass (~25 min per split 1.5h timeslice observed on this course) is untouched and will likely become the visible bottleneck once this deploys.
- New users are processed strictly one at a time: each user's revisions are fetched per course wiki
  timeslice, and the user's courses_users.created_at is ratcheted as soon as all their wikis complete, so
  an interrupted update loses at most one user's partial work and never redoes finished users. This
  maximizes durability and failure isolation at the cost of request volume, since Replica accepts up to 10
  usernames per request (RevisionDataManager::MAX_USERNAMES) and per-user fetching forgoes that batching
  — for a course with many inactive new users and many timeslices, most of the extra requests are empty
  responses. If that overhead proves significant in practice, a possible follow-up is to process users in
  chunks of up to 10, fetching each timeslice for the whole chunk and ratcheting the chunk's created_at
  together: roughly 10× fewer Replica requests, in exchange for coarser checkpoints (a crash or a single
  user's failure would redo up to 10 users instead of 1). The expensive work — revision processing and, on
  Wikidata, diff analysis — scales with revision volume and is identical under both schemes, which is why
  the simpler per-user version was chosen first.

### DIFF
--- a/app/services/reprocess_article_course_user_wiki_timeslices.rb
+++ b/app/services/reprocess_article_course_user_wiki_timeslices.rb
@@ -59,9 +59,7 @@ class ReprocessArticleCourseUserWikiTimeslices
 
   def fetch_filtered_revisions(users, article_ids, ts_start, ts_end)
     revisions = revision_data_manager.fetch_revision_data_for_users_with_articles_only(
-      users,
-      ts_start.strftime('%Y%m%d%H%M%S'),
-      (ts_end - 1.second).strftime('%Y%m%d%H%M%S')
+      users, ts_start, ts_end
     )
     revisions.select { |r| article_ids.include?(r.article_id) }
   end
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -144,18 +144,8 @@ class UpdateCourseWikiTimeslices
   def fetch_only_revisions(wiki, timeslice_start, timeslice_end)
     # Fetches only revision for wiki
     @revisions = @revision_updater.fetch_revisions_for_course_wiki(
-      wiki,
-      real_start(timeslice_start).strftime('%Y%m%d%H%M%S'),
-      real_end(timeslice_end).strftime('%Y%m%d%H%M%S')
+      wiki, timeslice_start, timeslice_end
     )
-  end
-
-  def real_start(timeslice_start)
-    [timeslice_start, @course.start].max
-  end
-
-  def real_end(timeslice_end)
-    [timeslice_end - 1.second, @course.end].min
   end
 
   def should_update_timeslice?(wiki, only_new:)
--- a/app/services/update_timeslices_course_user.rb
+++ b/app/services/update_timeslices_course_user.rb
@@ -47,8 +47,12 @@ class UpdateTimeslicesCourseUser
       add_user_ids_acuwt(user_ids)
     else
       add_user_ids_legacy(user_ids)
+      update_created_at_for_users(user_ids)
     end
-    # Update created_at field for courses users to prevent re-marking them as newly added
+  end
+
+  # Update created_at field for courses users to prevent re-marking them as newly added
+  def update_created_at_for_users(user_ids)
     @course.courses_users.where(user_id: user_ids)
            .update_all(created_at: @course_update_start - 1.second) # rubocop:disable Rails/SkipsModelValidations
   end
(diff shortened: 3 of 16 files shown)


ragesoss (approved):
Automated review assistance follows. These are observations from a Claude Code session, provided to read and use or discard based on your own judgment — they are not authoritative and may be wrong.

Both changes read as behavior-preserving optimizations, and local verification supports that:

- The per-timeslice `real_start`/`real_end` in `UpdateTimeslicesCourseUser` match `UpdateCourseWikiTimeslices` line-for-line. Since course-wiki timeslices tile the window contiguously (`cwt[i].end == cwt[i+1].start`) and each fetch is `[start, end − 1s]`, the union of per-timeslice fetches covers the same period as the old whole-period fetch with no gaps or double-counting.
- `RevisionScoreApiHandler#scores_available?` resolves to exactly `ReferenceCounterApi.valid_wiki?`. On an unsupported wiki (wikidata) the old loop already produced empty `scores` and `parent_scores`, so `add_scores_to_revisions` left revisions untouched — the only real work was the wasted `parent_revision_ids` call. The early return yields identical output.

Verified locally: the PR's three spec files (28 examples) plus `course_revision_updater_spec` and `update_course_wiki_timeslices_spec` (38 examples) pass on the branch merged with master; RuboCop is clean on all changed files.

Two non-blocking notes:

1. **Multi-timeslice coverage.** The per-timeslice distribution — revisions in different timeslices landing in separate ACUWT records, each affected CWT flagged for reaggregation — is the core mechanism of this change, but the PR's `creates ACUWT rows` spec only exercises the first timeslice. An exploratory spec covering two non-adjacent timeslices was written and passes against this branch; it is AI-generated and was run but not otherwise hand-audited. Consider adopting it:

<details><summary>Exploratory multi-timeslice spec</summary>

```ruby
context 'when a new user has revisions across multiple timeslices (ACUWT path)' do
  # ...setup mirroring the existing "added (ACUWT path)" context...
  it 'creates one ACUWT per timeslice and flags each affected CWT for reaggregation' do
    rev_ts0 = build(:revision_on_memory, article_id: article1.id, user_id: user2.id,
                    wiki_id: enwiki.id, mw_rev_id: 12_345, mw_page_id: article1.mw_page_id,
                    date: start + 1.hour, scoped: true, new_article: false)
    rev_ts2 = build(:revision_on_memory, article_id: article2.id, user_id: user2.id,
                    wiki_id: enwiki.id, mw_rev_id: 12_346, mw_page_id: article2.mw_page_id,
                    date: start + 2.days + 1.hour, scoped: true, new_article: false)
    allow_any_instance_of(CourseRevisionUpdater)
      .to receive(:fetch_revisions_for_new_users) do |_u, _wiki, _users, ts_start, _ts_end|
      case ts_start
      when start.strftime('%Y%m%d%H%M%S') then [rev_ts0]
      when (start + 2.days).strftime('%Y%m%d%H%M%S') then [rev_ts2]
      else []
      end
    end

    described_class.new(course).run

    acuwt = ArticleCourseUserWikiTimeslice.where(course:, user: user2)
    expect(acuwt.count).to eq(2)
    expect(acuwt.pluck(:start)).to contain_exactly(start, start + 2.days)
    reaggregation = course.course_wiki_timeslices.where(needs_reaggregation: true)
    expect(reaggregation.pluck(:start)).to contain_exactly(start, start + 2.days)
  end
end
```

</details>

2. **Duplication.** `real_start`/`real_end` now exist identically in both `UpdateTimeslicesCourseUser` and `UpdateCourseWikiTimeslices`. Extracting them to a shared location would keep the Replica inclusive-bound adjustment in one place, though the methods are trivial.

The per-timeslice wikidata stats path (the primary target scenario) is exercised only per smaller window, not directly tested — worth an operational eye on the Coordinate Me course after deploy.

*Reviewed in a Claude Code session (~25 min, ~5 interactions). Sage directed the review, which ran the affected specs, RuboCop, and an AI-written exploratory spec locally against the branch merged with master, and confirmed the `real_start`/`real_end` and `scores_available?` equivalences by reading the code. Sage reviewed this comment before posting; the code-reading conclusions were spot-checked rather than each independently re-derived.*

ragesoss:
@gabina want to extract the `real_start` / `real_end`?

gabina:
It took me a while because the logic was duplicated in various places and it was a bit of a delicate change to delegate entirely to Claude.

ragesoss:
Re-review of the post-approval commit f494ce2. Automated review assistance follows — observations from a Claude Code session, to read and use or discard based on your own judgment.

The extraction looks correct and goes further than the original ask. Verified locally on the branch merged with master: 89 examples across the 9 affected spec files pass (including `reprocess_article_course_user_wiki_timeslices_spec` and `update_timeslices_scoped_article_spec`, which cover the two call sites that newly gained the course-bound clamp); RuboCop clean on all changed files; `yarn build` succeeds. The four call sites all pass raw `Time` bounds, the two `RevisionDataManager` entry points clamp and format them, no other production callers of the changed methods exist, and the legacy `fetch_revision_data_for_users` path is correctly untouched. Recomputing the clamped bounds in `fetch_revisions_for_course_wiki` preserves the `revisions[:end]` invariant needed for single-timeslice matching.

Two minor notes, neither blocking:

1. **Leftover commented-out code** in `spec/lib/course_revision_updater_spec.rb`: the old string-literal `let` definitions remain as comments (lines 10/12 and 207/209). These can be deleted.

2. **String-vs-Time comparison in an untouched mock.** At `spec/services/update_timeslices_course_user_spec.rb:242`, the `creates ACUWT rows` mock compares `ts_start == start.strftime('%Y%m%d%H%M%S')`, but `ts_start` is now an `ActiveSupport::TimeWithZone` rather than a formatted string. The spec still passes — confirmed by instrumenting the block — but only because ActiveSupport's comparison coercion parses the string back into a datetime. Comparing against `start` directly would be clearer. Note that a `case ts_start; when start.strftime(...)` form of the same pattern does *not* survive the convention change (`String === Time` never coerces) — the exploratory multi-timeslice spec from the original review needed exactly that adaptation before it passed again.

*Re-reviewed in a Claude Code session (~15 min, minimal interaction: Sage initiated the re-review and approved this comment; the session ran the affected specs, RuboCop, and the build locally against the branch merged with master, verified the caller graph of every changed method by grep, and empirically confirmed the coercion behavior in note 2 by instrumenting the mock. Sage read this comment before posting but did not independently re-derive the findings.)*

PR #6964, opened 2026-07-20
Resolve dead interwiki wikilinks in training-slide

### LINKED ISSUE #6963: Links not rendering
### What is happening?
The links on https://meta.wikimedia.org/wiki/Training_modules/dashboard/slides/12218-identifying-content-gaps-for-editathons render correctly on that page, but not on https://outreachdashboard.wmflabs.org/training/editathons/planning-well-before-the-event/identifying-content-gaps-for-editathons

Not sure if this is an error in the format entry, or somewhere in the rendering.

### System Information

 - OS: Linux
 - Browser: Firefox
 - Version: 152.0.2

Formasitchijoh:
## What this PR does

Fixes #6963. Training slide content is pulled from meta.wikimedia.org as MediaWiki wikitext and converted to Markdown via Pandoc. Plain external links convert fine, but MediaWiki interwiki links like `[[wikipedia:Wikipedia:Notability#General_notability_guideline|General Notabilty Guidelines]]` rely on MediaWiki's own interwiki table to resolve, which Pandoc doesn't have — so it passed the raw wiki target through as a dead href.

This adds `WikiLinkResolver` (`lib/utils/wiki_link_resolver.rb`), which rewrites resolvable `[[wikilink]]` targets into real URLs before Pandoc runs, wired into `Wikitext.mediawiki_to_markdown` (the only place this conversion happens, for training slides). Category directives, image embeds, same-page anchors, and unrecognized namespaces are deliberately left untouched. Covered by new/updated specs in `spec/lib/utils/wiki_link_resolver_spec.rb` and `spec/lib/wikitext_spec.rb`.

## AI usage

This PR was developed in a Claude Code session. I directed the investigation and design (asked for the root cause and approach before any code was written, required verification against the real reported slide's actual wikitext rather than a synthetic example, and raised edge cases like `[[:Category:Foo]]` and same-page anchors that shaped the final design). Claude Code wrote the fix, the tests, and this description.

## Screenshots
https://www.loom.com/share/4ccdaa7ed5ed45228d0dda2954f88a71

### DIFF
--- a/lib/utils/wiki_link_resolver.rb
+++ b/lib/utils/wiki_link_resolver.rb
@@ -1,0 +1,65 @@
+# frozen_string_literal: true
+
+# Pandoc's mediawiki-to-markdown conversion leaves internal/interwiki
+# [[wikilink]] targets as the raw wiki target (eg
+# "wikipedia:Wikipedia:Notability"), since it has no access to MediaWiki's
+# interwiki table and can't resolve them into real URLs on its own. This
+# rewrites the targets we can confidently resolve into MediaWiki's external
+# link syntax (`[url text]`), which Pandoc already converts correctly.
+#
+# Anything we're not confident about — category links, image embeds,
+# same-page anchors, or namespaces/prefixes we don't recognize — is left
+# untouched, so Pandoc handles it exactly as it does today.
+class WikiLinkResolver
+  # Interwiki project prefixes, mapped to the domain they resolve to by
+  # default (English-language, unless overridden by an explicit language
+  # sub-prefix, eg the 'eu' in "w:eu:Foo").
+  PROJECT_DOMAINS = {
+    'w' => 'wikipedia.org', 'wikipedia' => 'wikipedia.org',
+    'wikt' => 'wiktionary.org', 'wiktionary' => 'wiktionary.org',
+    'n' => 'wikinews.org', 'wikinews' => 'wikinews.org',
+    'b' => 'wikibooks.org', 'wikibooks' => 'wikibooks.org',
+    'q' => 'wikiquote.org', 'wikiquote' => 'wikiquote.org',
+    's' => 'wikisource.org', 'wikisource' => 'wikisource.org',
+    'v' => 'wikiversity.org', 'wikiversity' => 'wikiversity.org',
+    'voy' => 'wikivoyage.org', 'wikivoyage' => 'wikivoyage.org'
+  }.freeze
+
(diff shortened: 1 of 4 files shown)


ragesoss (approved):
**Note on this review:** Portions of this review were produced with AI assistance (Claude Code). The feedback below is meant to be read, understood, and used or ignored based on the developer's own judgment — not automatically trusted or blindly acted upon.

The fix was verified against the actual wikitext of the slide reported in #6963: an exploratory spec fetched the live page content, ran it through `WikiSlideParser` with and without the resolver, and confirmed the dead `class="wikilink"` hrefs become correct `https://en.wikipedia.org/wiki/...` links while plain external links and image templates are untouched. The conservative leave-untouched design means the worst case for unrecognized targets equals current behavior. PR specs, related existing specs (`wikitext_spec`, `wiki_slide_parser_spec`, `wiki_training_loader_spec`), RuboCop, and `yarn build` all pass locally.

Non-blocking observations, worth a look but fine to defer or skip:

1. `WIKILINK_RE` character classes (`[^|\]]+`, `[^\]]+`) match across newlines, so stray unmatched `[[ ... ]]` brackets spanning lines get rewritten into a malformed meta link with embedded newlines. MediaWiki never parses link targets across newlines, so excluding `\n` from both classes would match MediaWiki semantics and prevent the mangling.
2. The language sub-prefix check (`lang == lang.downcase && lang.length.between?(2, 3)`) accepts non-letter characters: `[[w:a/:evil|x]]` produces `https://a/.wikipedia.org/wiki/evil`. All real 2–3 character language codes are lowercase letters, so `lang =~ /\A[a-z]{2,3}\z/` is strictly tighter.
3. `escape` only converts spaces; a title containing `?` yields a URL where the `?` is read as a query-string separator. Rare in slide links — noting for completeness.
4. Possible follow-up: `c:` (Commons), `d:` (Wikidata), `m:` (Meta), `mw:`, and language-only prefixes like `[[:en:Foo]]` — all common on meta — remain unresolved, since they don't fit the `lang.domain` URL shape. Could be a follow-up issue if such links turn up in training content.

*Reviewed in a Claude Code session (~15 min, 2 reviewer interactions). Claude Code checked out the branch, ran the PR's specs plus related existing specs, RuboCop, and the build, and wrote/ran the exploratory spec that reproduced the bug and verified the fix against the real slide's wikitext. Sage read the session's findings and approved posting; the individual findings were not independently re-verified.*

PR #6968, opened 2026-07-23
Gsoc26/system wide csv export

lakshita10341:
## What this PR does
This pull request implements a system-wide CSV export feature with a robust asynchronous architecture. It allows administrators to export system statistics filtered by campaign, home wiki, date range, course type, and course status.
Specifically, it includes:
1. **Asynchronous Background Processing (Sidekiq)**:
   - Offloads CSV generation to `ReportCsvWorker` in the background to avoid web-request timeouts on large data sets.
   - Refactored `ReportsController#system_csv` to support background queuing of requests with JSON-serialized filters.
   - Uses `SystemCsvBuilder` to handle all filtering logic (`campaign_slug`, `start_date`, `end_date`, `wiki_domain`, `course_type`, and `status`).
2. **Frontend Polling & Download UI (`SystemCsvExportBar`)**:
   - Implemented a polling hook that periodically checks for file completion (every 6s, up to 20 attempts / 2 minutes).
   - Capping initial polling at 2 minutes is intentional to reduce unnecessary server load and avoid holding open browser polling loops during long-running exports.
   - If CSV generation takes longer than 2 minutes, the UI gracefully displays a `still_processing` notice informing the user that generation is continuing in the background, and they can click "Export CSV" again later to retrieve the cached file instantly.
   - Once the file is generated, it automatically triggers a direct browser download via a dynamic hidden anchor link with the `download` attribute, keeping administrators seamlessly on the dashboard page without reloads.
   
## AI usage
I used AI to:
- Design the asynchronous polling mechanism in the `SystemCsvExportBar` React component.
- Refactor `ReportsController#system_csv` to safely enqueue CSV generation.
- Build custom mock datasets in `db/seeds.rb` that skip external Wikipedia API calls.

## ScreenRecording
[Screencast from 2026-07-22 00-46-44.webm](https://github.com/user-attachments/assets/37d5409c-261d-4cc1-bafe-e7a0c46534f9)


ragesoss (changes requested):
This review was drafted with AI assistance (Claude Code). AI-generated review feedback is provisional — it is meant to be read, understood, and used or ignored based on the developer's own judgment, not automatically trusted or blindly acted upon.

The overall design — async generation, batched preloading, polling with file-existence check — is a good fit for this report. It mirrors the existing `CampaignCsvBuilder#courses_to_csv` pattern and is considerably safer than the existing synchronous `/all_courses_csv` endpoint. The build, RuboCop, ESLint, the PR's specs, and baseline specs for the affected areas were all run locally and pass.

The requested changes below are mostly about how this feature behaves on the Programs & Events Dashboard (outreachdashboard.wmflabs.org), which runs the same codebase against a much larger course table (~100K+ programs, 2,361 campaigns) with more constrained infrastructure.

### 1. Add `.distinct` to the wikis query in `SystemStatsController`

`index_json_data` now runs this on every page load:

```sql
SELECT wikis.language, wikis.project FROM wikis
INNER JOIN courses_wikis ON courses_wikis.wiki_id = wikis.id
INNER JOIN courses ON courses.id = courses_wikis.course_id
WHERE courses.private = FALSE
```

Without `DISTINCT`, this returns one row — and instantiates one `Wiki` object — per nonprivate course-wiki pair before deduplicating in Ruby. On the P&E Dashboard that is on the order of 100K+ rows per admin page view. Adding `.distinct` to the relation bounds the result to the number of distinct wikis (a few hundred).

Related, non-blocking: the dropdown is built from courses' *all* wikis (via `courses_wikis`), but `apply_wiki_filter` matches `home_wiki` only — so some dropdown options will always return zero courses. Building the list from `courses.home_wiki_id` instead would make the options match the filter's semantics.

### 2. Queue placement for the export job

`ReportCsvWorker` runs on the `default` Sidekiq queue. On the P&E Dashboard, that queue runs at low concurrency on the web server and handles interactive jobs (enrollment, mass enrollment, notifications, emails). An unfiltered system export there must aggregate `article_course_timeslices` for every nonprivate course — the only index leading on `course_id` is `[course_id, updated_at, article_id]`, so the job effectively reads every timeslice row for every course. That is plausibly an hours-long job occupying a default-queue thread, and two concurrent exports with different filters would occupy the queue entirely, delaying user-facing jobs.

Before this ships, there should be a decision about routing this report type (via `sidekiq_options queue:`, or a `set(queue:)` at enqueue time) to a queue that isn't shared with interactive work. Input from the maintainers on which queue is appropriate would be needed here.

### 3. Revision counts diverge from the single-course CSV for article-scoped courses

`SystemCsvBuilder#fetch_revision_counts` sums all tracked `ArticleCourseTimeslice` rows, but the canonical single-course export (`CourseCsvBuilder` with no preloaded data) uses `Course#scoped_article_timeslices`, which restricts `ArticleScopedProgram` / `VisitingScholarship` / `article_scoped`-flagged courses to their scoped articles. The divergence was reproduced with the spec below: a scoped program with a 10-revision assigned article and a 90-revision out-of-scope article reports 10 mainspace edits in the single-course CSV and 100 in the system CSV.

`CampaignCsvBuilder` has the same behavior, so this follows precedent — but article-scoped programs are among the most common course types on the P&E Dashboard, and out-of-scope timeslice rows do exist between updates for category-tracked courses. A decision is needed: either accept the campaign-CSV-consistent behavior (worth a code comment), or restrict the aggregate to scoped articles for those course types.

<details>
<summary>Reproduction spec (AI-generated during review; run locally, both examples pass)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/analytics/course_csv_builder"

describe 'SystemCsvBuilder vs CourseCsvBuilder for article-scoped courses' do

  let(:en_wiki) { Wiki.get_or_create(language: 'en', project: 'wikipedia') }

  let!(:course) do
    create(:course, slug: 'org/scoped_program_(term)',
                    title: 'Scoped Program',
                    type: 'ArticleScopedProgram',
                    home_wiki: en_wiki,
                    start: 2.months.ago,
                    end: 2.months.from_now)
  end

  let!(:in_scope_article) do
    create(:article, title: 'In_Scope', wiki: en_wiki,
                     namespace: Article::Namespaces::MAINSPACE)
  end
  let!(:out_of_scope_article) do
    create(:article, title: 'Out_Of_Scope', wiki: en_wiki,
                     namespace: Article::Namespaces::MAINSPACE)
  end

  before do
    # Only in_scope_article is assigned, so it alone is "scoped".
    create(:assignment, course:, article: in_scope_article,
                        article_title: in_scope_article.title, wiki: en_wiki)

    create(:article_course_timeslice, course:, article: in_scope_article,
                                      revision_count: 10, tracked: true,
                                      start: 1.month.ago, end: 1.month.ago + 1.day)
    create(:article_course_timeslice, course:, article: out_of_scope_article,
                                      revision_count: 90, tracked: true,
                                      start: 1.month.ago, end: 1.month.ago + 1.day)
  end

  mainspace_index = CourseCsvBuilder::CSV_HEADERS.index('mainspace_edits')

  it 'single-course CSV counts only scoped articles' do
    # Course.find returns the proper STI subclass (ArticleScopedProgram)
    row = CourseCsvBuilder.new(Course.find(course.id)).row
    expect(row[mainspace_index]).to eq(10)
  end

  it 'system CSV counts all tracked timeslices, including out-of-scope articles' do
    csv = SystemCsvBuilder.new(filters: { course_type: 'ArticleScopedProgram' }).generate_csv
    data_row = CSV.parse(csv)[1]
    expect(data_row[mainspace_index].to_i).to eq(100)
  end
end
```

</details>

### 4. Polling window vs. large-export reality

The frontend polls for 20 × 6s = 2 minutes, then shows "Could not generate CSV export. Please try again." while the background job keeps running. A full P&E export will essentially always exceed 2 minutes, so on that deployment the first attempt will always end in an error message even though the export succeeds later (re-clicking after completion downloads the file). The "gave up polling but the job is still running" case should be handled distinctly from an actual failure. Note that per project convention, any new user-facing message text needs to come from Wiki Education staff rather than being drafted in the PR — coordination in the PR thread is fine.

(Related observation, no change needed: repeated polling does not stack duplicate jobs — `sidekiq-unique-jobs` with `lock: :until_executed` suppresses re-enqueues with identical args. One caveat worth knowing: `lock_ttl` is nil, so if a worker process dies mid-job, that filter combination stays locked until the date-based filename changes the args the next day.)

### Minor, non-blocking

- `wiki_domain` is the only filter that isn't validated; it flows into the generated filename (slashes are stripped, so no traversal, but a >255-byte value would raise `Errno::ENAMETOOLONG` and return a 500). A length cap or membership check against known wiki domains would tighten this up.
- Namespace-edit columns render blank rather than 0 for courses with no matching timeslices (same as the campaign CSV — noting for awareness only).
- The comment in `system_csv_export_bar.jsx` says the polling window is 20 attempts; the PR description says 15.

*Drafted in a Claude Code session (~1 hour, ~4 interactions). Claude Code ran the build, linters, and specs locally, reproduced the scoped-article divergence with the spec above, and confirmed the generated SQL for the wikis query. Sage reviewed the findings summary and this comment before posting, but did not independently re-run the verifications. The P&E table-scale figures are estimates from public data, not production measurements.*

lakshita10341:
@ragesoss All the issues are addressed. PTAL!

ragesoss (commented):
This follow-up review was drafted with AI assistance (Claude Code). AI-generated review feedback is provisional — meant to be read, understood, and used or ignored based on the developer's own judgment, not automatically trusted or blindly acted upon.

The changes in `19af5734` address all four items from the prior review.

**Addressed**

- **Wikis query** — `SystemStatsController#index_json_data` now selects wikis by distinct `home_wiki_id` instead of instantiating one `Wiki` per nonprivate course-wiki pair, and building the list from `home_wiki_id` also aligns the dropdown options with `apply_wiki_filter`'s `home_wiki` matching.
- **Queue placement** — `ReportCsvWorker` now runs on `daily_update`, off the interactive `default` queue. This matches the precedent documented on `HarvestClaimPoolWorker` (a long, manually-triggered batch kept off the constant-update cycle).
- **Revision-count divergence for article-scoped courses** — documented at `fetch_revision_counts` as intentional, campaign-CSV-consistent behavior. A rerun of the earlier reproduction spec confirms the aggregate is unchanged (scoped single-course CSV = 10, system CSV = 100).
- **Polling timeout** — the exhausted-polling case now shows a distinct "still processing" notice instead of the failure message.
- **`wiki_domain` validation** — added (length ≤ 255 plus a character-set check), with a 422 spec; this closes the earlier `ENAMETOOLONG` → 500 path.

Build, RuboCop, ESLint, the PR's specs, and the baseline specs for the affected controllers were run locally against the branch merged with current master, and all pass.

**Non-blocking observations**

- `SystemStatsController#index_json_data` (`app/controllers/system_stats_controller.rb:34`): `Course.nonprivate.pluck(:home_wiki_id)` still loads every nonprivate course's `home_wiki_id` into memory before de-duplicating in Ruby — on the order of 100K+ integers per admin page load on the Programs & Events Dashboard. A subquery form such as `Wiki.where(id: Course.nonprivate.select(:home_wiki_id))` would push the de-duplication into the database. This is already a large improvement over the previous version; the remaining cost is integers rather than AR objects, so it is not blocking.
- The PR description still states the polling window is "up to 15 attempts"; the code and its comment use 20 (20 × 6s = 2 min).

*Re-review drafted in a Claude Code session (~20 min, ~3 interactions). Sage directed the re-review and reviewed this comment before posting. Claude Code ran the build, linters, PR and baseline specs, and reran the scoped-article reproduction spec locally against the branch merged with master; Sage did not independently re-run those verifications. The P&E table-scale figure is an estimate from public data, not a production measurement.*

ragesoss:
## Performance benchmark: CSV export load on the P&E Dashboard

A Claude Code session benchmarked `SystemCsvBuilder` to estimate worker time and
system load for this feature on Peony (P&E Dashboard) production.

**Bottom line:** a full unfiltered export takes roughly 5 minutes of worker time
and is ~99% database wait, with negligible CPU and memory cost. That runtime has
been judged acceptable for this feature, so no query optimization is needed to
merge. The one thing that does need changing is the **frontend polling window**,
which currently gives up before the export can finish.

### Method

The benchmark ran against a local Peony database snapshot, which has the real
`courses` table (18,329 nonprivate courses → 19 batches of 1,000) and the real
`articles` table (25.4M rows, 2.3 GB data + 3.2 GB index). The snapshot only had
`article_course_timeslices` rows for 20 courses, so timeslice volume was
synthesized into a throwaway clone table populated from the real 15.2M
`articles_courses` pairs, giving realistic article-ID and course distribution.
Measurements are end-to-end runs of `SystemCsvBuilder#generate_csv` with
per-query-class SQL timing, CPU time, and peak RSS instrumented.

### Headline results — full unfiltered export

| Timeslice rows | Wall time | Ruby CPU | Peak RSS | CSV output |
|---|---|---|---|---|
| 15.2M | 316 s | 2.6 s (1%) | +82 MB | 4.9 MB / 18,330 rows |
| 30.5M | 467 s | 2.9 s (1%) | +91 MB | 4.9 MB / 18,330 rows |

Two things stand out:

1. **The job is ~99% database wait.** Only ~3 s of Ruby CPU and <100 MB of RSS.
   Load on the Sidekiq host is negligible; effectively all cost lands on
   `peony-database`.
2. **`fetch_revision_counts` is ~99% of the runtime** (462 s of the 467 s at
   30.5M rows). Everything else is noise: `new_editors` ~3 s, tags 0.06 s,
   wikis 0.01 s, batch course fetch 0.13 s.

Per-batch cost is highly variable — at 30.5M rows the average batch took 24 s
but the worst took 77 s, since cost tracks timeslice rows rather than course
count.

### Needs attention: the polling window is shorter than the job

`system_csv_export_bar.jsx` polls 20 × 6 s = **120 s**, but an unfiltered export
needs roughly 300–470 s. An admin exporting without filters will always see
`still_processing` and has to click export roughly three times before the file
appears — the job does complete, but the UI reports failure while it is still
running. If ~5 minutes is the accepted worker time, the polling window needs to
cover it (or the UI needs to surface job status rather than poll to a fixed
deadline).

Each poll also re-enqueues the job. The `lock: :until_executed` option
deduplicates these, so behavior is correct, but it does depend on that lock
holding for the full run.

### Worth a second look: the queue change

This PR moves `ReportCsvWorker` from the `default` queue to `daily_update`. That
affects *every* CSV export, not just the new one — instructor course CSVs and
campaign CSVs now share a concurrency-2 queue with `DailyUpdateWorker`. A
five-minute system export consumes half that queue's capacity for its duration,
and conversely ordinary course CSV downloads now queue behind daily update work.
A dedicated `csv` queue would decouple the two directions.

Separately: because the generated filename is date-stamped and cached for the
whole day, a scheduled nightly job could pre-build the unfiltered export so the
common admin click is served from cache instantly.

### Where the time goes (for future reference)

`fetch_revision_counts` performs two reads per tracked timeslice row, both
against clustered indexes much larger than the buffer pool:

- reads a ~100-byte timeslice row to extract two integers (`article_id`,
  `revision_count`) — roughly 35% of query time
- probes a ~100-byte `articles` row to extract one byte of `namespace` —
  roughly 65% of query time (measured via `ANALYZE FORMAT=JSON`: 3,781 ms of
  5,800 ms on a sample batch)

If this ever does need to be faster, two covering indexes gave a 6× end-to-end
improvement with no application code change:

```sql
ALTER TABLE articles
  ADD KEY articles_id_namespace (id, namespace);
ALTER TABLE article_course_timeslices
  ADD KEY acts_covering (course_id, tracked, article_id, revision_count);
```

| Config (30.5M timeslices, unfiltered) | Wall time | Worst batch |
|---|---|---|
| Baseline | 467 s | 77 s |
| Both covering indexes | 78 s | 9 s |

Both plans flip to `Using index`. Not proposed for this PR — noting it in case
the export slows down as the timeslice table grows. The trade-off is that
`article_course_timeslices` is high-churn during course updates, so a
four-column index there adds write amplification; the `articles (id, namespace)`
index is lower risk and accounted for about half the gain on its own.

Two smaller options, also not needed now: 17% of nonprivate courses have
`revision_count = 0`, so filtering each batch's ID list on `revision_count > 0`
would skip that share of guaranteed-wasted probes. And `BATCH_SIZE = 1000`
happens to equal MariaDB's default `in_predicate_conversion_threshold`, which
converts the IN-list into a materialized derived table. At production data
density the optimizer still picks a good index-driven plan, so this is not a live
problem — but against a *sparse* timeslice table the plan flipped to a full table
scan per batch (999 IDs → 0.43 s index range; 1,000 IDs → 4.34 s derived + `ALL`;
82 s instead of ~4 s across a full export). `BATCH_SIZE = 500` would stay clear of
that threshold at no cost.

### One approach that measured worse

Hoisting the aggregation out of the per-batch loop into a single query for all
courses — the pattern `CampaignCsvBuilder` uses — is **slower**, not faster:
313 s versus 148 s on comparable batches. The GROUP BY working set grows to
14,864 groups against a 16 MB `tmp_table_size` and spills. The per-batch
structure in this PR is the right call and is worth keeping.

### Filters do not reliably reduce cost

Because cost tracks timeslice rows rather than course count, filtering does not
scale down the way course counts suggest. The `miscellanea` campaign is 39% of
courses but cost 270 s — 58% of the full export — with a *worse* per-batch
average (33.5 s vs 24.3 s), since the largest campaign holds the most active
courses. A narrow filter is genuinely cheap (a 53-course `status=active` export
took 4.8 s), but a broad one is not a mitigation, and the polling window issue
above applies to broad filters too.

### Caveats on these numbers

- Timeslice volume was synthesized, since the snapshot lacked it. Row width and
  article-ID distribution came from real data, but the 30.5M figure is
  **optimistic**: the second pass reused the same `(article_id, course_id)`
  pairs, giving better cache locality than real data would. The 15.2M run is the
  more trustworthy absolute datapoint.
- The local MariaDB buffer pool was the 128 MB default against a 5.5 GB
  `articles` table, so nearly every probe was a cache miss. If `peony-database`
  has a substantially larger pool, absolute times will be better than these;
  the *ratios* between configurations should hold up better than the absolutes.
- The `status=active` datapoint matched only 53 courses in this snapshot (its
  data ends mid-2026), so it is not representative of production active-course
  counts.
- Confirming production figures needs the real `article_course_timeslices` row
  count and `innodb_buffer_pool_size` from `peony-database`, which were not
  queried for this analysis.

---

These benchmarks were run and this comment was drafted in a Claude Code session.
Sage reviewed the findings and approved posting.

lakshita10341:
Hi @ragesoss , for the queue concern, I think making new queue for ReportCsvWorker is better because a nightly pre-build wouldn't fully solve this either, since any filtered export (e.g., a large campaign) would still take 3–5 minutes and wouldn't be pre-built. Let me know what looks good to you!

lakshita10341:
>Two smaller options, also not needed now: 17% of nonprivate courses have revision_count = 0, so filtering each batch's ID list on revision_count > 0 would skip that share of guaranteed-wasted probes. And BATCH_SIZE = 1000 happens to equal MariaDB's default in_predicate_conversion_threshold, which converts the IN-list into a materialized derived table. At production data density the optimizer still picks a good index-driven plan, so this is not a live problem — but against a sparse timeslice table the plan flipped to a full table scan per batch (999 IDs → 0.43 s index range; 1,000 IDs → 4.34 s derived + ALL; 82 s instead of ~4 s across a full export). BATCH_SIZE = 500 would stay clear of that threshold at no cost.

 I ran a benchmark on ~776,000 timeslices across 200 courses to compare the index options:
| Configuration | Query Time | Disk Speedup | Write Overhead / Build Time | 
|---|---|---|---|
| **Baseline** | 0.990 s | 1.0× | None |
| **`articles (id, namespace)`** | 0.970 s | **2.5–3× faster** | 0.07 s build; **Near-zero write cost** | 
| **`timeslices` covering index** | 0.809 s | 1.5–2× faster | 2.05 s build; **High write churn** | 
| **Both indexes** | 0.722 s | **6× faster** | High write amplification on 15M+ rows | 


The stats confirm that the **`articles (id, namespace)`** index is virtually free to maintain (0.07s build time on static data), whereas the **`timeslices` covering index** took 30× longer to build (2.05s) because updating `revision_count` on every 5-minute course update forces MySQL to re-index those rows constantly.

I will update the batch size and revision count filter for optimization. 

Since the `articles (id, namespace)` index is low risk, let me know if you'd like me to open a separate PR for it!

ragesoss:
sure, a separate PR for the `id-namespace` index sounds good.

ragesoss:
I don't think I want to add a separate queue. Putting it in the daily queue is fine. That worker is free most of the time anyway. The main limitation is that the CSV workers need to be on the same machine as the web server, unless/until we implement a solution for file access or shared storage so that users could download a file generated on a different instance than  the webserver.

lakshita10341:
>Needs attention: the polling window is shorter than the job
system_csv_export_bar.jsx polls 20 × 6 s = 120 s, but an unfiltered export
needs roughly 300–470 s. An admin exporting without filters will always see
still_processing and has to click export roughly three times before the file
appears — the job does complete, but the UI reports failure while it is still
running. If ~5 minutes is the accepted worker time, the polling window needs to
cover it (or the UI needs to surface job status rather than poll to a fixed
deadline).
Each poll also re-enqueues the job. The lock: :until_executed option
deduplicates these, so behavior is correct, but it does depend on that lock
holding for the full run.

Regarding the polling window,  ater 2 minutes, the UI shows: "Your CSV export is still generating in the background. Please check back shortly and click Export again."
Increasing the polling window would generate 50+ HTTP requests per export, unnecessarily increasing server load. Since the background job continues running and the completed file downloads instantly on the next click, the current timeout works well if that sounds good to you.

lakshita10341:
> I don't think I want to add a separate queue. Putting it in the daily queue is fine. That worker is free most of the time anyway. The main limitation is that the CSV workers need to be on the same machine as the web server, unless/until we implement a solution for file access or shared storage so that users could download a file generated on a different instance than the webserver.

Hi @ragesoss, the PR is ready for review! 
Regarding the multi-server setup, that makes total sense. For the future, if we ever switch to a multi-server setup, we can use shared cloud storage to check and serve generated CSVs from there.

ragesoss (changes requested):
This re-review was drafted with AI assistance (Claude Code). AI-generated review feedback is provisional — meant to be read, understood, and used or ignored based on the developer's own judgment, not automatically trusted or blindly acted upon. The reasoning below should be checked rather than assumed correct.

`26b49885a0` addresses the remaining items from the previous rounds: the wikis query now pushes de-duplication into the database via `Wiki.where(id: Course.nonprivate.select(:home_wiki_id))`, `BATCH_SIZE` moved to 500 to stay clear of MariaDB's `in_predicate_conversion_threshold`, and the spec `ACCEPT` headers were fixed. Build, RuboCop, ESLint, the PR's 35 specs, and baseline specs for the affected controllers were all run locally against the branch merged with current master, and all pass.

Changes are requested for one item: the `revision_count > 0` prefilter added in the same commit causes the export to silently report zero namespace edits for courses that have real data.

## Blocking: `active_course_ids` can zero out real revision data

`active_course_ids` gates the ACT aggregation on `courses.revision_count`:

```ruby
def active_course_ids(batch)
  batch.select { |c| c.revision_count.to_i.positive? }.map(&:id)
end
```

That counter is not derived from the table being aggregated. It is the sum over `course_wiki_timeslices`:

```ruby
# lib/course_cache_manager.rb:17
@course.revision_count = course_wiki_timeslices.sum(&:revision_count)
```

while `fetch_revision_counts` sums `article_course_timeslices`. When the two disagree, the consequence is silent rather than loud: `CourseCsvBuilder:106` is `return @revisions[[@course.id, namespace]] if @revisions`, so a skipped course yields `nil` and the cell renders blank. There is no fallback to the per-course query, because the hash is truthy even when the course is absent from it.

This was reproduced with the spec below — a course with 100 mainspace edits in `article_course_timeslices` and a stale `courses.revision_count` of 0:

```
when the cached courses.revision_count is in sync       → 100        (pass)
when the cached courses.revision_count is stale at zero  → 0          (expected 100)
matches what the canonical single-course CSV reports     → 0 vs 100   (expected equal)
```

Reverting only `lib/analytics/system_csv_builder.rb` to `19af573429` and rerunning the same spec gives 3/3 passing, which locates the change precisely rather than inferring it.

The reachable cases look ordinary rather than contrived: a course mid-first-update, where ACT rows are written before `CourseCacheManager` runs; an update that errors between those two writes; or CWT rows removed while ACT rows persist. That last case is worth particular attention given the active timeslice-deletion work on master (`Delete ACUWT rows in UpdateTimeslicesCourseDate`, `TimesliceCleaner#delete_timeslices_for_period`) — CWT/ACT consistency is an invariant currently in flux, and this prefilter turns any divergence into wrong CSV output.

Setting correctness aside, the trade appears unfavorable on its own terms. Courses with `revision_count = 0` hold few or no ACT rows by construction, so their probes are index seeks that find nothing — a small single-digit share of the query's I/O at most. Passing `batch_course_ids` instead of `active_course_ids(batch)` restores correctness at negligible cost.

Worth noting how this passed CI: `spec/lib/analytics/system_csv_builder_spec.rb` contains no reference to `revision_count`, `article_course_timeslice`, or `mainspace`. It covers filter construction thoroughly, but the method responsible for ~99% of the export's runtime has no assertions on its output values. A value-level spec would be worth adding regardless of how the prefilter is resolved.

<details>
<summary>Reproduction spec (AI-generated during review; run locally, results as shown above)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/analytics/course_csv_builder"

describe 'SystemCsvBuilder revision_count gate' do

  let(:en_wiki) { Wiki.get_or_create(language: 'en', project: 'wikipedia') }

  let!(:course) do
    create(:course, slug: 'org/stale_cache_(term)',
                    title: 'Stale Cache Course',
                    home_wiki: en_wiki,
                    start: 2.months.ago,
                    end: 2.months.from_now)
  end

  let!(:article) do
    create(:article, title: 'Mainspace_Article', wiki: en_wiki,
                     namespace: Article::Namespaces::MAINSPACE)
  end

  mainspace_index = CourseCsvBuilder::CSV_HEADERS.index('mainspace_edits')

  # Real mainspace edit data exists in article_course_timeslices...
  before do
    create(:article_course_timeslice, course:, article:,
                                      revision_count: 100, tracked: true,
                                      start: 1.month.ago, end: 1.month.ago + 1.day)
  end

  context 'when the cached courses.revision_count is in sync' do
    before { course.update_column(:revision_count, 100) }

    it 'reports the mainspace edits' do
      csv = SystemCsvBuilder.new(filters: {}).generate_csv
      expect(CSV.parse(csv)[1][mainspace_index].to_i).to eq(100)
    end
  end

  context 'when the cached courses.revision_count is stale at zero' do
    before { course.update_column(:revision_count, 0) }

    it 'still reports the mainspace edits from article_course_timeslices' do
      csv = SystemCsvBuilder.new(filters: {}).generate_csv
      expect(CSV.parse(csv)[1][mainspace_index].to_i).to eq(100)
    end

    it 'matches what the canonical single-course CSV reports' do
      single = CourseCsvBuilder.new(Course.find(course.id)).row
      csv = SystemCsvBuilder.new(filters: {}).generate_csv
      expect(CSV.parse(csv)[1][mainspace_index].to_i).to eq(single[mainspace_index])
    end
  end
end
```

</details>

## The accepted runtime estimate is roughly 4× low

The earlier benchmark on this PR concluded a full unfiltered export costs ~5 minutes of worker time, and that figure was judged acceptable. It rested on **synthesized** timeslice volume, and flagged its own gap: *"Confirming production figures needs the real `article_course_timeslices` row count and `innodb_buffer_pool_size` from `peony-database`, which were not queried for this analysis."*

Those figures have since been measured on `peony-database`, during the review of #6974. `article_course_timeslices` is **79,592,476 rows / 15.1 GB** and `innodb_buffer_pool_size` is **12 GB** — so the real table is 2.6× the largest synthetic stand-in. A single cold aggregation across all nonprivate courses (`CampaignCsvBuilder#revision_counts`, the same query shape) measured:

```
1451.21 s, physical reads 632575, groups 9018
```

632,575 × 16 KiB ≈ 9.65 GiB read from disk, at 6.8 MiB/s effective — entirely I/O-bound on random reads.

| Data point | ACT rows | Buffer pool | Result | Per row |
|---|---|---|---|---|
| Prior benchmark (synthetic) | 15.2M | 128 MB | 316 s | 20.8 µs |
| Prior benchmark (synthetic, optimistic locality) | 30.5M | 128 MB | 467 s | 15.3 µs |
| Production (measured, unbatched) | 79.6M | 12 GB | 1451 s | 18.2 µs |

The per-row cost holds at 15–21 µs across a **94× difference in buffer pool size**, which indicates the working set overwhelms the pool in every configuration and that row count is the term that matters. Projecting the batched export at real scale gives 79.6M × 15–18 µs ≈ 1220–1430 s, less roughly 15% now that `articles(id, namespace)` (#6974) has merged — so **on the order of 17–20 minutes** rather than 5.

These projections are linear extrapolations from measured page counts, not a direct timing of this PR's batched builder against the real table. The direction is well anchored; the specific minute figure is not.

Three earlier decisions rested on the 5-minute premise and may deserve reconsideration at 20:

- **Polling window.** 20 × 6 s = 120 s against a ~1,100 s job. An unfiltered export will always exhaust polling. The `still_processing` notice makes that graceful rather than a false failure, but extending the window to cover 20 minutes of 6-second polls is not a practical fix; surfacing real job status would be.
- **Queue occupancy.** `daily_update` runs at concurrency 2. A single export now holds half that capacity for ~20 minutes, and two concurrent exports with different filters would hold all of it. Ordinary course and campaign CSVs route through the same worker.
- **`lock: :until_executed` with `lock_ttl: nil`.** Exposure scales with runtime. A ~20-minute job interrupted by a deploy or restart leaves that filter combination locked until the date-stamped filename rolls over the following day.

The nightly pre-build idea raised earlier in this thread looks more attractive at this runtime than it did at 5 minutes, since it would serve the common unfiltered admin request from cache and sidestep the polling problem for the dominant case.

## Minor

The PR description still states the polling window is "up to 15 attempts"; the code and its comment use 20.

*Re-review drafted in a Claude Code session (~30 min, ~3 interactions), continuing a longer session in which Sage ran the `peony-database` measurements himself in a Rails console — the 79.6M row count, the 12 GB buffer pool figure, and the 1451-second aggregation above are his measurements, not estimates. Claude Code checked out the branch merged with master, ran the build, linters, the PR's specs and baseline controller specs, and wrote and ran the reproduction spec, including the single-file revert to `19af573429` that isolates the regression. Sage reviewed this comment before posting but did not independently re-run the spec or re-derive the per-row extrapolation, which is the part most worth checking — the regression itself is directly reproducible from the spec above.*

lakshita10341:
@ragesoss PTAL when you get time!

ragesoss (approved):
This re-review was drafted with AI assistance (Claude Code). AI-generated review feedback is provisional — meant to be read, understood, and used or ignored based on the developer's own judgment, not automatically trusted or blindly acted upon.

`dfa412dedf` resolves the blocking item from the previous round. Approving.

## The revision-count regression is fixed, and guarded

`build_batch_rows` passes `batch_course_ids` to `fetch_revision_counts` again and `active_course_ids` is gone. The round-3 reproduction spec was rewritten from scratch and now passes 3/3, where it previously passed 1/3: courses with a stale `courses.revision_count` of 0 again report their real `article_course_timeslices` totals, and the system CSV again agrees with what the canonical single-course CSV reports.

The new example at `spec/lib/analytics/system_csv_builder_spec.rb:266` was checked as a genuine regression guard rather than taken on faith — reintroducing `active_course_ids` locally makes it fail (`expected 100, got 0`), so the prefilter cannot come back silently. That also closes the value-level coverage gap noted last round; `fetch_revision_counts` now has an assertion on its output rather than only on filter construction.

## Runtime items accepted as designed

The previous round re-estimated a full unfiltered export at roughly 17–20 minutes rather than 5, and flagged three consequences. These are accepted rather than blocking:

- **Polling window** — 20 × 6 s = 120 s against a substantially longer job, so an unfiltered export will exhaust the window and land on the `still_processing` notice. The job completes regardless and the next click serves the cached file, which makes this graceful rather than a false failure.
- **Queue placement** — `daily_update` stays, per the maintainer decision earlier in this thread; that worker is idle most of the time, and CSV workers need to be on the web server host until shared file storage exists.
- **`lock: :until_executed` with `lock_ttl: nil`** — a job interrupted by a deploy or restart leaves that filter combination locked until the date-stamped filename rolls over the following day. Worth knowing about, not worth holding the PR for.

The `articles (id, namespace)` index from #6974 has since merged to master, which takes an estimated ~15% off that projection.

The new message strings in `config/locales/en.yml` are approved as written.

## Verification

Run locally against the branch merged with current master: `yarn build`, RuboCop on the eight changed Ruby files, ESLint on the two changed JSX files, the PR's 36 specs, and baseline specs for `ReportsController`, `SystemStatsController`, and `Course` (108 examples). All pass. The three build warnings are pre-existing `i18next/no-literal-string` warnings in unrelated components.

Five exploratory specs from earlier review rounds now fail against this branch, and all five fail because the bugs they pinned have been fixed — they assert the 500 responses that filter validation now correctly returns as 422, and a `new_editors` result from the old global `min(start)..max(end)` window that now correctly matches the per-course count.

## Non-blocking observations

- `app/controllers/system_stats_controller.rb:32`: `Campaign.select(:slug, :title).order(:title)` serializes every campaign into the page payload on each admin load — on the order of 2,300 rows on the Programs & Events Dashboard. Small next to what the wikis query cost before it was fixed, and not a regression, but the dropdown could be populated on demand if that payload ever becomes a concern.
- The polling component has no JavaScript test. `test/components/` contains no `system_csv_export_bar` coverage, so the attempt counting, the 422 branch, and the `ready`/`generating` branching are unexercised by CI.
- An exploratory spec written during the first review round asserts full row parity between `SystemCsvBuilder` and `CampaignCsvBuilder` over the same course set, and passes against the current branch. Since the two builders are expected to agree and the system builder deliberately follows the campaign builder's aggregation precedent, this is a compact guard against future drift and would be a reasonable addition to the PR's specs:

<details>
<summary>Row-parity spec (AI-generated during review; run locally, passes)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/analytics/campaign_csv_builder"

describe SystemCsvBuilder do

  let(:en_wiki) { Wiki.get_or_create(language: 'en', project: 'wikipedia') }

  describe 'row parity with CampaignCsvBuilder' do
    let(:campaign) { create(:campaign, slug: 'parity_campaign', title: 'Parity Campaign') }

    before do
      course_a = create(:course, slug: 'school/parity_a_(term)', title: 'Parity A',
                                 start: 2.months.ago, end: 2.months.from_now,
                                 type: 'ClassroomProgramCourse', home_wiki: en_wiki)
      course_b = create(:course, slug: 'school/parity_b_(term)', title: 'Parity B',
                                 start: 3.months.ago, end: 1.month.from_now,
                                 type: 'Editathon', home_wiki: en_wiki)
      campaign.courses << course_a
      campaign.courses << course_b
      create(:tag, course: course_a, tag: 'first_time_instructor')
    end

    it 'produces the same rows as the campaign courses CSV for the same course set' do
      system_csv = described_class.new(filters: { campaign_slug: 'parity_campaign' }).generate_csv
      campaign_csv = CampaignCsvBuilder.new(campaign).courses_to_csv
      expect(system_csv.lines.sort).to eq(campaign_csv.lines.sort)
    end
  end
end
```

</details>

One process note: the PR is still marked as a draft, so it will need to be marked ready for review before it can be merged.

*Re-review drafted in a Claude Code session (~15 min, 2 interactions). Claude Code checked out the branch merged with current master and ran the build, both linters, the PR's specs and the baseline controller and model specs; it also rewrote and ran the round-3 reproduction spec, and verified the PR's new spec is a real guard by reintroducing `active_course_ids` and confirming that example fails. Sage directed the re-review, read the findings summary, made the call to approve, and signed off on the new message strings, but did not independently re-run those verifications. The 17–20 minute runtime projection carried over from the previous round remains a linear extrapolation from measured page counts, not a direct timing of this builder against the production table.*

ragesoss:
@lakshita10341 let us know once you've had a chance to do the EXPLAIN queries and whether you want to make any more changes based on what you learn.

lakshita10341:
Hi @ragesoss @Abishekcs I ran `EXPLAIN ANALYZE` on MySQL across all 5 queries executed by `SystemCsvBuilder` (`BATCH_SIZE = 500`). Below are the exact estimated costs, execution times, and index plans:

| Query / Helper | Estimated Cost | Actual Execution Time | Access Path / Index Used | Resulting Behavior |
|---|---|---|---|---|
| **1. `fetch_revision_counts`** | `2.44` | **0.025 ms** (scan) / **0.049 ms** (aggregate) | `Index range scan` on `article_course_timeslice_by_updated_at` | Fast index range scan per 500-course batch; no full table scans |
| **2. `fetch_tags`** | `6.61` | **0.018 ms** | `Index range scan` on `index_tags_on_course_id_and_key` | Targeted index range lookup for instructor tags |
| **3. `fetch_new_editor_counts`** | `2.05` | **0.038 ms** (scan) / **0.045 ms** (aggregate) | `Covering index range scan` on `index_courses_users_on_course_id_and_user_id_and_role` | In-index covering scan for student registration date checks |
| **4. `fetch_wikis`** | `1.60` | **5.26 ms** | `Covering index scan` on `index_wikis_on_language_and_project` | In-memory DB-level deduplication via `Wiki.where(id: ...)` |
| **5. `course_scope`** | `14.70` | **1.41 ms** | Filtered scan on `courses.private = 0` | Scope filtering prior to batch processing |

---

 **No Full Table Scans:**
   - All 5 queries utilize **Index Range Scans** over `course_id IN (1, 2, ..., 500)` and **Covering Indexes**, confirming that MySQL jumps directly to batch rows rather than rescanning table data per batch.

---

###  Detailed Execution Trees

<details>
<summary><b>1. fetch_revision_counts (Timeslices & Articles Aggregation)</b></summary>

```text
-> Table scan on <temporary>  (actual time=0.0517..0.0517 rows=0 loops=1)
    -> Aggregate using temporary table  (actual time=0.0492..0.0492 rows=0 loops=1)
        -> Nested loop inner join  (cost=2.44 rows=0.3) (actual time=0.0299..0.0299 rows=0 loops=1)
            -> Filter: (article_course_timeslices.tracked = 1)  (cost=2.11 rows=0.3) (actual time=0.0256..0.0256 rows=0 loops=1)
                -> Index range scan on article_course_timeslices using article_course_timeslice_by_updated_at over (course_id = 1) OR (course_id = 2) OR ...
            -> Filter: (articles.namespace in (0,1,2))  (cost=1.33 rows=1)
                -> Single-row index lookup on articles using PRIMARY (id=article_course_timeslices.article_id)
```
</details>

<details>
<summary><b>2. fetch_tags</b></summary>

```text
-> Filter: (tags.tag in ('first_time_instructor','returning_instructor'))  (cost=6.61 rows=0.6) (actual time=0.0202..0.0202 rows=0 loops=1)
    -> Index range scan on tags using index_tags_on_course_id_and_key over (course_id = 1) OR (course_id = 2) OR ... (cost=6.61 rows=3) (actual time=0.0185..0.0185 rows=0 loops=1)
```
</details>

<details>
<summary><b>3. fetch_new_editor_counts</b></summary>

```text
-> Group aggregate: count(0)  (cost=2.05 rows=0.0333) (actual time=0.0459..0.0459 rows=0 loops=1)
    -> Nested loop inner join  (cost=2.05 rows=0.0333) (actual time=0.0432..0.0432 rows=0 loops=1)
        -> Nested loop inner join  (cost=1.72 rows=0.3) (actual time=0.0422..0.0422 rows=0 loops=1)
            -> Filter: ((courses_users.role = 0) and (courses_users.course_id in (1,2,3)))  (cost=1.61 rows=0.3)
                -> Covering index range scan on courses_users using index_courses_users_on_course_id_and_user_id_and_role over (course_id = 1) OR (course_id = 2) OR ...
            -> Single-row index lookup on courses using PRIMARY (id=courses_users.course_id)
        -> Filter: ((users.registered_at >= courses.start) and (users.registered_at <= courses.end))
            -> Single-row index lookup on users using PRIMARY (id=courses_users.user_id)
```
</details>

<details>
<summary><b>4. fetch_wikis</b></summary>

```text
-> Filter: (wikis.id in (1,2,3))  (cost=1.6 rows=3) (actual time=5.26..5.27 rows=3 loops=1)
    -> Covering index scan on wikis using index_wikis_on_language_and_project  (cost=1.6 rows=6) (actual time=5.26..5.26 rows=7 loops=1)
```
</details>

<details>
<summary><b>5. course_scope</b></summary>

```text
-> Filter: (courses.private = 0)  (cost=14.7 rows=5.7) (actual time=1.41..4.31 rows=57 loops=1)
    -> Table scan on courses  (cost=14.7 rows=57) (actual time=1.4..4.29 rows=57 loops=1)
```
</details>

---

All query execution plans are optimal and batching (`BATCH_SIZE = 500`) is performing cleanly. So the PR is ready to merge.

PR #6971, opened 2026-07-24
Reimplement ArticleStatusManager retrieval and reset on ACUWT

### LINKED ISSUE #6907: [ACUWT] Implement ArticleStatusManagerTimeslice
### Current Behavior:
`ArticleStatusManagerTimeslice` doesn't run for ASP courses, so it wasn't implemented in the `article-course-user-wiki-timeslices` branch for now.

### Desired Behavior:
Implement `ArticleStatusManagerTimeslice` for ACUWT flag so we can use the same updates for every course, no matter the kind of course.

gabina:
## What this PR does
Closes #6907

Adapts `ArticleStatusManager` to run on the ACUWT model (`ArticleCourseUserWikiTimeslice`) for ACUWT courses, so that its article retrieval and its reset logic are driven by ACUWT rather than by reading the ACT (`ArticleCourseTimeslice`) table. This is the last piece of the `ArticleStatusManager` ACUWT migration, following the `ArticleNamespacesManager` work.

Two touchpoints changed:

- **Article retrieval**: `ArticleStatusManager.update_article_status_for_course` now reads articles via `Course#articles_from_timeslices` (ACUWT-aware, with a legacy ACT fallback for non-ACUWT courses) instead of `articles_from_timeslices_legacy`. For ACUWT courses this surfaces the correct set of tracked *and* untracked articles from ACUWT rows.
- **Duplicate/undeletion reset** (`handle_undeletion` → `ArticlesCoursesCleaner.reset_specific_articles`): now branches on `use_acuwt?`. The ACUWT path derives the covering periods from the article's ACUWT rows and delegates to a new `TimesliceCleaner#reset_timeslices_for_update_from_acuwt` primitive that marks the covering CWTs `needs_update` (a real per-period re-fetch) and deletes the whole-period ACUWT and CUWT rows (plus the article's ACT rows) so the re-fetch regenerates them cleanly. Non-ACUWT courses keep falling back to `reset_legacy`.

A full per-period re-fetch is used for this reset (rather than a targeted ACUWT rescore or a reaggregation) because the duplicate/history-merge case is fundamentally about *re-attributing* revisions to the live article record — only an unfiltered re-fetch of the whole period can ingest a revision now resolving to a different `article_id`. This is deliberate; see "Open questions and concerns".

### A note on which tables this touches

This does not make `ArticleStatusManager` fully independent of ACT/CUWT — it's narrower than that:

- **ACUWT** is the new source of truth: retrieval reads it, and the reset derives its covering periods from it.
- **ACT** is no longer *read* for article retrieval (that was the switch), but the reset still *deletes* ACT rows.
- **CUWT** rows are *deleted* by the reset (whole-period).
- **CWT** is marked `needs_update` by the reset.

The ACT/CUWT deletions are cleanup of *derived* tables — they are rebuilt from ACUWT by the subsequent re-fetch. CUWT can't replace ACUWT anywhere here because it has no `article_id` dimension.

ACT is deleted only for the passed articles (it has no `wiki_id` to target a whole period); this can leave stale ACT rows for other zero-revision articles in the period, which is accepted and commented because the ACT table is slated for removal.

`DuplicateArticleDeleter` needs no change: it delegates to `reset_specific_articles` and so inherits the ACUWT-aware behavior.

Specs added for all three classes: the retrieval switch (`article_status_manager_spec`), the reset branch delegation (`articles_courses_cleaner_spec`), and the new primitive's whole-period vs article-scoped deletes (`timeslice_cleaner_spec`).

## AI usage

The code, specs, and commit message were written by Claude Code (`/commit`), under close human direction. The direction was hands-on and correction-driven rather than fire-and-forget — the reviewer:

- caught that the reset path (`handle_undeletion`), not just the retrieval line, needed changing, and chose a full per-period re-fetch for it;
- caught that the new reset primitive was wrongly modeled on the reaggregation primitive (reaggregation must keep ACUWT; a full reprocess must delete it) and that not deleting CUWT would leave stale rows — both fixed;
- decided to leave ACT deletion article-scoped (with a comment) rather than widen it, since ACT is being removed soon;
- confirmed, after analysis, that `DuplicateArticleDeleter` needs no change and that the targeted "mark ACUWT needs_update + reaggregate" alternative can't replace the re-fetch for this case;
- caught an overreaching "no longer depends on ACT/CUWT" claim in an earlier draft of this description and had it corrected to the more precise statement above.

Scale of effort: a single commit made in one focused session on 2026-07-23; the bulk of the human time went into the design discussion and corrections above, not volume of code (~234 lines including specs). This PR description was drafted using the `/prepare-pr` Claude Code skill, and — like the summary above — may contain errors; please review against the diff.

## Screenshots

No UI changes (backend only).

- **Currently a forward-looking path for ACUWT.** `UpdateCourseStats#should_update_article_status?` returns false for `only_scoped_articles_course?` before checking `use_acuwt?`, and every ACUWT course today is scoped, so `ArticleStatusManager` does not yet run on an ACUWT course. This matters once a non-scoped course adopts ACUWT.
- **Full re-fetch vs. targeted rescore.** We analyzed whether the `handle_undeletion` reset could avoid marking CWTs `needs_update` and instead mark specific ACUWT rows + reaggregate. It can't for this case: the re-attributed revisions resolve to the live article's `article_id`, which the targeted reprocess filters out (and its user set is scoped to the old article's editors), so the live copy would receive nothing. Kept the full re-fetch. A no-fetch "relabel `article_id` + reaggregate" optimization is possible but has merge/collision/partial-history-merge caveats — left for a future spike if this ever becomes a performance concern (it shouldn't; duplicates are rare).


ragesoss (commented):
Review from a Claude Code session, covering the full diff plus the surrounding
update path.

A note on how to use this: these are observations from an AI-assisted review,
not conclusions to act on automatically. Some may be wrong or may misread
intent. Worth reading and judging on the merits, then using or discarding.

Verified before the observations below: `bundle exec rubocop` is clean on all six
changed files; the PR's own specs pass (60 examples); and 201 existing examples
across ten related spec files (`duplicate_article_deleter`,
`update_timeslices_scoped_article`, `article_namespaces_manager`, `course`,
`update_course_wiki_timeslices`, `reprocess_article_course_user_wiki_timeslices`,
`timeslice_manager`, `course_wiki_timeslice`, the ACUWT model, and
`update_course_stats`) pass with no regressions.

Several claims in the description were checked against the code and hold up: the
`index_hash` column really is a generated column that is NULL when `deleted`,
under a unique index (`db/schema.rb:99-100`), so the duplicate-coexistence
comment is accurate; `reaggregate_timeslices_from_acuwt` guards with
`.where(needs_update: false)`, so reaggregation cannot clear a `needs_update`
flag this reset sets without a re-fetch having happened; `update_from_acuwt`
fully recomputes rather than incrementing; and `get_ingestion_start_time_for_wiki`
returns the latest period holding data, so a past period marked `needs_update` is
never skipped as "future".

## The ACUWT branch is a no-op when an article has no ACUWT rows

`articles_from_timeslices` replaces the ACT-based set rather than supplementing
it, and `reset_timeslices_for_update_from_acuwt` bails on
`return if acuwt_records.empty?`. An article whose only timeslice rows are ACT
rows is therefore invisible to the status sync; and if it reaches the reset by
another route, only the destructive half of the reset runs.

That state appears reachable. There is no ACT-to-ACUWT backfill anywhere in the
codebase — only a full rebuild (`PrepareTimeslices#recreate_timeslices`,
triggered by `needs_update`) populates ACUWT for historical periods. And
`CourseFlagsController` (`app/controllers/course_flags_controller.rb:19-23`)
exposes `use_acuwt` and `needs_update` as independent checkboxes. A course
flagged without also requesting a full update keeps ACT/CUWT/CWT with no ACUWT
rows for every pre-flag period, indefinitely: step 3 of the update only revisits
periods from `get_ingestion_start_time_for_wiki` forward.

Two symptoms follow.

**1. Through `DuplicateArticleDeleter` — reachable today.** It calls
`reset_specific_articles` for every course during revision fetching
(`lib/revision_data_manager.rb:131`), scoped ACUWT courses included. With no
ACUWT rows for the article, `reset_timeslices_for_update_from_acuwt` returns
early, so no period is marked `needs_update` and no ACT rows are cleaned — but
`delete_article_course(article_ids)` still runs unconditionally. The
`articles_courses` record is destroyed, the repair is skipped, and the stale ACT
rows remain. Minimal reproduction (first example passes, other two fail):

```ruby
describe ArticlesCoursesCleaner do
  let(:course) { create(:course, start: '2024-01-01', end: '2024-04-20') }
  let(:stale) { create(:article, wiki: enwiki, deleted: true) }

  before { stub_const('TimesliceManager::TIMESLICE_DURATION', 86400) }

  describe '.reset_specific_articles on an ACUWT course with no ACUWT rows' do
    before do
      stub_wiki_validation
      create(:articles_course, course:, article: stale)
      create(:article_course_timeslice, course:, article: stale,
             start: '2024-01-10', end: '2024-01-11')
      TimesliceManager.new(course).create_timeslices_for_new_course_wiki_records([enwiki])

      described_class.reset_specific_articles(course, Article.where(id: stale.id))
    end

    it 'deletes the articles_courses record' do
      expect(course.articles_courses.where(article: stale)).to be_empty
    end

    it 'marks the covering period for re-fetch' do
      timeslice = course.course_wiki_timeslices.find_by(wiki: enwiki, start: '2024-01-10')
      expect(timeslice.needs_update).to eq(true)
    end

    it 'deletes the stale ACT rows' do
      expect(course.article_course_timeslices.where(article: stale)).to be_empty
    end
  end
end
```

**2. Through `ArticleStatusManager` — latent, but the case this PR prepares for.**
With ACT-only rows, the duplicate pair is never examined at all: no deletion
detection, no title/namespace sync, no duplicate resolution. An equivalent
legacy-course scenario marks the period `needs_update` and deletes the stale ACT
rows, so this is a behavior change rather than a pre-existing gap being
inherited. It is latent today because `should_update_article_status?` returns
false for scoped courses — which is the same condition the description's "Open
questions" section already names.

Two possible directions: have `articles_from_timeslices` union ACUWT with ACT
(retrieval runs once per update per wiki, so the added cost is small), or fall
back to `reset_legacy` when the ACUWT relation comes back empty. `reset_excluded`
has the same shape and shares the exposure; that part, and
`articles_from_timeslices` itself, predate this PR — what is new here is routing
`mark_deleted!` and the undeletion repair through it.

## Smaller items

- **"On the next update" looks inaccurate.** Three of the new comments say the
  re-fetch happens on the next update. `UpdateCourseStats#initialize` runs
  `update_articles` (line 37) before `UpdateCourseWikiTimeslices#run` (line 38),
  and that service's first step is `fetch_data_and_reprocess_timeslices`, which
  consumes `needs_update`. The re-fetch therefore lands later in the same update.
  Worth correcting, since the same-cycle timing together with the
  `where(needs_update: false)` guard in `reaggregate_timeslices_from_acuwt` is
  what makes the design safe — that is the reasoning a later reader will need.

- **Four copies of one SQL builder.** `mark_timeslices_for_update`,
  `mark_timeslices_for_reaggregation`,
  `delete_course_user_wiki_timeslices_for_acuwt_pairs` and
  `delete_article_course_user_wiki_timeslices_for_pairs` each build the identical
  `"(wiki_id, 'Y-m-d H:M:S')"` list. `lib/timeslice_cleaner.rb` is already on the
  `Metrics/ClassLength` exclusion list in `.rubocop.yml` ("too big, and should be
  shrunk if feasible"); one private helper would remove three of the four.

- **Unreachable guard.** In `reset_timeslices_for_update_from_acuwt`,
  `return if wikis_and_starts.empty?` cannot fire after
  `return if acuwt_records.empty?` — a non-empty relation always plucks rows.
  Inherited from the two reaggregation methods, so it may be worth removing in
  all three.

- **Index coverage on the new whole-period ACUWT delete.**
  `delete_article_course_user_wiki_timeslices_for_pairs` filters on `course_id`
  plus a row-constructor `(wiki_id, start) IN (...)`. The only candidate index is
  `index_acuwt_on_course_id_and_wiki_id`, and MariaDB generally will not use an
  index for row-constructor IN, so this looks like a scan of the course's whole
  ACUWT partition, repeated per 5000-row `delete_in_batches` batch, on the
  largest of the four timeslice tables for exactly the high-churn courses ACUWT
  exists to serve. Likely acceptable given how rare duplicates are, but
  `where(course:, wiki_id: w, start: starts)` per wiki would be exact and
  index-friendly.

- **Naming.** The new instance `#reset_specific_articles` shadows the class method
  that calls it. Its siblings are `reset_legacy` / `reset_excluded` /
  `reset_included`, and the other entry points use distinct names
  (`clean_articles_courses_for_wiki_ids` calls
  `remove_articles_courses_for_wiki_ids`).

## Coverage note

The PR's specs cover each of the three classes in isolation and never the three
together, which is why the empty-ACUWT case did not surface. Worth noting too
that both new cleaner specs pass an `ActiveRecord::Relation`, while
`ArticleStatusManager#handle_undeletion` passes a bare `[article]` Array; that
path does work, but nothing in the PR exercises it.

One assumption behind symptom 2 being latent rather than live was not verified:
that every course currently flagged `use_acuwt` in production is scoped. That
needs a production check.

---

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage asked for the
review and approved posting after reading a summary, but did not independently
verify any finding. The verification described above was done by Claude Code:
branch checked out and merged with master, RuboCop run on the changed files, the
PR's specs and the ten related spec files run, and an exploratory spec written to
reproduce the empty-ACUWT case. The spec in this comment is AI-generated and was
run locally (1 pass, 2 failures as shown); Sage did not review or run it.*

gabina:
**The ACUWT branch is a no-op when an article has no ACUWT rows**

I think all the review in that section is based on the idea that a default course suddenly becomes ACUWT course without a right migration (without a complete reprocessing), and then it has no ACUWT records for historical data. I don't think we have to handle that case, we should think about a good migration process instead.

 **"On the next update" looks inaccurate**
 I only found two new comment saying that. Changed them in 5e8767eacd3289207f0c839d9239eaa707bc77cb
 
 **Four copies of one SQL builder.**
 Done on 51e5364799d8fab42a47482d1e19dde9cc92107f.
 
 **Unreachable guard.**
 Done on 5efcc1876

ragesoss (commented):
Follow-up from a Claude Code session, re-reviewing the three commits pushed on
2026-08-11 against the previous review.

Same caveat as before: these are observations from an AI-assisted review, not
conclusions to act on automatically. Some may be wrong or may misread intent.

Verified on the branch merged with current master: `bundle exec rubocop` is clean
on all six changed files; the PR's own specs pass (60 examples); and 102 examples
across seven related spec files (`timeslice_manager`, `duplicate_article_deleter`,
`article_namespaces_manager`, `update_course_wiki_timeslices`,
`reprocess_article_course_user_wiki_timeslices`, `course_wiki_timeslice`, the ACUWT
model) pass with no regressions.

## Addressed

- **"On the next update" comments** — reworded to "the next time
  `UpdateCourseWikiTimeslices` runs" in `5e8767ea`. Accurate now.
- **Four copies of the SQL builder** — `timeslices_for_pairs(model, id_column, pairs)`
  in `51e53647` converts all four call sites and removes about thirty lines from a
  class that is on the `Metrics/ClassLength` exclusion list. The extraction reads as
  behavior-preserving, and the spec runs above are consistent with that.
- **Unreachable guards** — removed from all three ACUWT methods in `5efcc187`.

## On the empty-ACUWT case

The position in the PR comment — that an ACUWT course with no ACUWT rows for
historical periods is a migration defect to be prevented rather than a state this
code should absorb — is a reasonable call, and the broader argument from the
previous review is not worth relitigating on that basis.

Two narrower points appear to survive it.

**1. The method is not atomic.** Re-running the reproduction from the previous
review against current HEAD still produces the same split outcome: the
`articles_courses` row is deleted, while the repair half returns early, so no
period is marked `needs_update` and the stale ACT rows remain. That asymmetry is
independent of how the empty state arises — under any migration policy, the
destructive half arguably should not run when the repair half does not. Hoisting
the emptiness check up into `reset_specific_articles`, or moving
`delete_article_course` inside the non-empty branch, would make the two halves
succeed or skip together at the cost of about one line.

**2. The migration discipline the argument relies on is not enforced in code.**
`CourseFlagsController` (`app/controllers/course_flags_controller.rb:19-23`) still
sets `use_acuwt` and `needs_update` from independent checkboxes, and there is no
ACT-to-ACUWT backfill anywhere in the codebase — only a full rebuild populates
ACUWT for historical periods. If the intended rule is that flipping `use_acuwt`
requires a full reprocess, the controller looks like the natural place to enforce
it. That seems better as a follow-up issue than as a change to this PR.

Reachability is narrower than the previous review stated. `DuplicateArticleDeleter`
(via `lib/revision_data_manager.rb:131`) remains the only route that reaches the
ACUWT branch today. The third caller, `app/services/update_timeslices_scoped_article.rb:105`,
sits inside the `use_acuwt? == false` branch, so it is not one.

## Resolved from the previous review's coverage note

The bare-Array path that `ArticleStatusManager#handle_undeletion` uses
(`[article]` rather than a relation) was flagged as working but unexercised. It
does work — confirmed by the spec below, which marks the covering CWT and deletes
the ACUWT rows as intended. The PR still has no spec for it, so this is offered
for adoption rather than as a finding:

```ruby
describe '.reset_specific_articles with a bare Array' do
  let(:course) { create(:course, start: '2024-01-01', end: '2024-04-20') }
  let(:stale) { create(:article, wiki: enwiki, deleted: true) }
  let(:user) { create(:user) }

  before do
    stub_wiki_validation
    create(:articles_course, course:, article: stale)
    create(:article_course_user_wiki_timeslice, course:, article: stale, user:,
           wiki: enwiki, start: '2024-01-10', end: '2024-01-11')
    create(:article_course_timeslice, course:, article: stale,
           start: '2024-01-10', end: '2024-01-11')
    TimesliceManager.new(course).create_timeslices_for_new_course_wiki_records([enwiki])
    described_class.reset_specific_articles(course, [stale])
  end

  it 'marks the covering period for re-fetch' do
    timeslice = course.course_wiki_timeslices.find_by(wiki: enwiki, start: '2024-01-10')
    expect(timeslice.needs_update).to eq(true)
  end

  it 'deletes the ACUWT rows' do
    expect(ArticleCourseUserWikiTimeslice.where(course:, article: stale)).to be_empty
  end
end
```

## Still open from the previous review

Neither was addressed in the response, and neither is blocking:

- **Index coverage on the whole-period ACUWT delete.** Confirmed against
  `db/schema.rb:80-82`: the candidate indexes are `index_acuwt_unique`
  (`course_id, article_id, user_id, wiki_id, start, end`) and
  `index_acuwt_on_course_id_and_wiki_id`. Neither can serve `course_id` plus a
  row-constructor `(wiki_id, start) IN (...)` beyond the `course_id` prefix, so
  this still looks like a scan of the course's ACUWT partition per batch. Likely
  acceptable given how rare duplicates are; `where(course:, wiki_id: w, start: starts)`
  per wiki would be index-friendly if it ever matters.
- **Naming.** The instance `#reset_specific_articles`
  (`lib/articles_courses_cleaner.rb:164`) still shadows the class method that calls
  it (line 30).

## New in the delta

Nothing. One item at the level of trivia: the `return if article_starts.empty?`
guard at `lib/timeslice_cleaner.rb:283` has the same shape as the three that were
removed and is equally unreachable from both of its call sites, which already
guard on `acuwt_records.empty?`. It sits in a private helper, where a defensive
guard is more defensible than in the public methods, so leaving it is a
defensible choice.

One item from the previous review remains unverified: whether every course
currently flagged `use_acuwt` in production is also scoped. That still needs a
production check, and it determines whether the `ArticleStatusManager` symptom is
latent or live.

---

*Drafted in a Claude Code session (~15 min, 2 interactions). Sage asked for the
re-review, read a summary of the findings, and directed the framing on the
empty-ACUWT item — specifically, to accept the migration argument and restate only
the atomicity and enforcement points rather than relitigate the original finding.
He did not independently verify any finding. The verification described above was
done by Claude Code: branch checked out and merged with master, RuboCop run,
the PR's specs and seven related spec files run, the previous review's
reproduction re-run against current HEAD, callers of `reset_specific_articles`
traced, and the schema indexes checked. The spec in this comment is AI-generated
and was run locally (both examples pass); Sage did not review or run it. The
production check above has not been performed.*

gabina:
> The method is not atomic. Re-running the reproduction from the previous
> review against current HEAD still produces the same split outcome: the
> articles_courses row is deleted, while the repair half returns early, so no
> period is marked needs_update and the stale ACT rows remain. That asymmetry is
> independent of how the empty state arises — under any migration policy, the
> destructive half arguably should not run when the repair half does not. Hoisting
> the emptiness check up into reset_specific_articles, or moving
> delete_article_course inside the non-empty branch, would make the two halves
> succeed or skip together at the cost of about one line.

The update system works with the assumption that we create ACUWT and ACT at the same time, and articles_courses records for a subset of those articles having revisions (that means, having ACUWT and ACT).
_The articles_courses row is deleted, while the repair half returns early, so no period is marked needs_update and the stale ACT rows remain._ Having articles courses records without them having timeslices is an unexpected scenario I think. It's not the way the update process is designed to work.


> The migration discipline the argument relies on is not enforced in code.
> CourseFlagsController (app/controllers/course_flags_controller.rb:19-23) still
> sets use_acuwt and needs_update from independent checkboxes, and there is no
> ACT-to-ACUWT backfill anywhere in the codebase — only a full rebuild populates
> ACUWT for historical periods. If the intended rule is that flipping use_acuwt
> requires a full reprocess, the controller looks like the natural place to enforce
> it. That seems better as a follow-up issue than as a change to this PR.

This would be handle in the future in a different PR.

ragesoss:
I'm happy with this.

PR #6972, opened 2026-07-24
Delete ACUWT rows in UpdateTimeslicesCourseDate

### LINKED ISSUE #6969: [ACUWT] Ensure ACUWT rows are deleted in UpdateTimeslicesCourseDate
### What is happening?
`UpdateTimeslicesCourseDate` class deletes timeslices for invalid dates but it could leave stale ACUWT rows.

This doesn't produce any incorrect statistics, but it leaves behind meaningless orphan rows.

gabina:
## What this PR does

Closes #6969. When a course's start or end date changes so that some
timeslices fall outside the new period, `UpdateTimeslicesCourseDate` removes the
now-invalid `CourseWikiTimeslice` and `CourseUserWikiTimeslice` rows but leaves
the corresponding `ArticleCourseUserWikiTimeslice` (ACUWT) rows behind. Those
orphans don't affect any statistics, but they accumulate as meaningless rows.

This PR adds ACUWT cleanup to both removal paths:

- `TimesliceCleaner` gains three methods mirroring the existing CUWT ones —
  `delete_article_course_user_wiki_timeslices_after_date`,
  `delete_article_course_user_wiki_timeslices_prior_to_start_date`, and
  `delete_article_course_user_wiki_timeslices_after_end_date`. ACUWT shares the
  same `wiki_id`/`start`/`end` columns as CUWT, so the queries are identical in
  shape.
- `UpdateTimeslicesCourseDate` calls the two new wrappers from
  `remove_timeslices_prior_to_start_date` and `remove_timeslices_after_end_date`.
  These paths already cover all three removal scenarios (start moved later, end
  moved earlier, and drastic non-overlapping change), so no ACUWT-specific branch
  was needed — the existing code reused cleanly.

The standalone `_after_date` method is only called internally by
`_after_end_date`; it's kept to stay consistent with the CWT/CUWT/ACT variants,
which follow the same pattern.

## AI usage

The code, specs, and commit message were drafted by Claude Code under the user's
direction. The user pointed it at the issue, proposed the two method names and
the observation that no ACUWT branch was needed, and made the call to keep the
`_after_date` helper for cross-table consistency after Claude questioned whether
it earned its place. The user ran RuboCop, the specs, and a local test before
committing.

This was a single focused session: one commit, all changes made within about an
hour on 2026-07-24.

This PR description was drafted using a Claude Code skill (`/prepare-pr`). The
"What this PR does" summary above was also drafted by Claude Code and may contain
errors.

## Screenshots

No UI changes — backend-only cleanup.

None.

### DIFF
--- a/app/services/update_timeslices_course_date.rb
+++ b/app/services/update_timeslices_course_date.rb
@@ -83,9 +83,10 @@ class UpdateTimeslicesCourseDate
     Rails.logger.info "UpdateTimeslicesCourseDate: Course: #{@course.slug}\
     Removing data prior to: #{@course.start}"
 
-    # Delete course and course user timeslices
+    # Delete course, course user and article course user timeslices
     @timeslice_cleaner.delete_course_wiki_timeslices_prior_to_start_date
     @timeslice_cleaner.delete_course_user_wiki_timeslices_prior_to_start_date
+    @timeslice_cleaner.delete_article_course_user_wiki_timeslices_prior_to_start_date
 
     # Delete articles courses
     ArticlesCoursesCleaner.clean_articles_courses_prior_to_course_start(@course)
@@ -97,9 +98,10 @@ class UpdateTimeslicesCourseDate
     Rails.logger.info "UpdateTimeslicesCourseDate: Course: #{@course.slug}\
     Removing data after to: #{@course.end}"
 
-    # Delete course and course user timeslices
+    # Delete course, course user and article course user timeslices
     @timeslice_cleaner.delete_course_wiki_timeslices_after_end_date
     @timeslice_cleaner.delete_course_user_wiki_timeslices_after_end_date
+    @timeslice_cleaner.delete_article_course_user_wiki_timeslices_after_end_date
 
     # Delete articles courses
     ArticlesCoursesCleaner.clean_articles_courses_after_course_end(@course)
--- a/lib/timeslice_cleaner.rb
+++ b/lib/timeslice_cleaner.rb
@@ -93,6 +93,14 @@ class TimesliceCleaner
     delete_in_batches(timeslices)
   end
 
+  # Deletes article course user wiki timeslices records with a start date later than
+  # the specific given date
+  def delete_article_course_user_wiki_timeslices_after_date(wikis, date)
+    timeslices = ArticleCourseUserWikiTimeslice.where(course: @course).where(wiki: wikis)
+                                               .where('start > ?', date)
+    delete_in_batches(timeslices)
+  end
+
   # Deletes article course timeslices records with a start date later than the
   # specific given date
   def delete_article_course_timeslices_after_date(wikis, date)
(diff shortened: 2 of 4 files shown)


ragesoss (approved):

Approving. The new methods are structural mirrors of the CUWT equivalents — same `start > ?` and `end <= ?` predicates, same `where(wiki: wikis)` on the `_after_date` variant, same absence of a wiki filter on `_prior_to_start_date`. That symmetry is what makes the fix safe: because ACUWT and CWT delete on identical predicates, no ACUWT row is removed whose timeslice period survives, and none survives whose period is removed.

Three things that could have gone wrong were checked specifically:

- **Ordering against `ArticlesCoursesCleaner`.** The new ACUWT deletions run before `clean_articles_courses_prior_to_course_start` / `_after_course_end` in both paths. If that cleaner decided which articles to drop by reading ACUWT, deleting first would silently change its behavior — it reads ACT (`article_ids_edited_beofre_start` / `article_ids_edited_after_end`), so there is no dependency.

- **The `_after_date` helper.** The description's justification holds up: all four `_after_date` methods are internal-only helpers, and `delete_article_course_timeslices_after_date` already has no caller anywhere in `app/` or `lib/`. The new one matches an existing pattern rather than introducing a novel unused method.

- **No `use_acuwt?` gate.** For non-ACUWT courses both queries match nothing, which is consistent with the other ungated ACUWT deletion in this class (`delete_acuwt_for_deleted_course_users`).

One note, not blocking: neither ACUWT index — `(course_id, article_id, user_id, wiki_id, start, end)` or `(course_id, wiki_id)` — places `start`/`end` where the date predicate becomes a range seek, so both new deletes scope by course (and wiki) through an index and then filter on the date column. That is the same shape as the existing CWT/CUWT deletions, but it lands on the timeslice table with the highest row count, so the sweep is larger by roughly the articles-per-user factor. Bounded by `delete_in_batches` at 5000 and only triggered by rare per-course date changes, so this is a characterization rather than a problem — worth knowing if ACUWT volume grows and date changes ever show up as slow.

An exploratory spec was written during review, asserting the cross-table invariant directly (surviving ACUWT rows equal the previously-existing rows whose period still exists as a `CourseWikiTimeslice`), which catches orphans and over-deletion in a single assertion rather than through per-call-site counts. It passes on this branch, and three of its four examples fail against master — so it reproduces #6969 rather than passing vacuously. It can be posted here if useful for the PR's own coverage; it was generated by Claude Code and has not been reviewed line by line by a human.

Verified locally against the branch merged with current master: RuboCop clean on all four changed files, the PR's own specs pass, and 310 examples across `spec/services` plus the timeslice cleaner/manager/articles-courses-cleaner specs pass with no failures. No migrations, and no asset files in the diff.

---

*Reviewed in a Claude Code session (~15 min, 2 interactions) using the repo's `/review-pr` skill. Claude Code traced the predicate symmetry against the CUWT methods, confirmed the `ArticlesCoursesCleaner` ordering and the `_after_date` caller claim by grep, checked the ACUWT indexes against the new queries, wrote the exploratory invariant spec and ran it against both the branch and master, and ran RuboCop and the spec sets named above. Sage directed the review and chose to approve after reading the findings, but did not independently re-derive the predicate comparison or review the exploratory spec line by line.*

PR #6973, opened 2026-07-28
Delete ACUWT rows in TimesliceCleaner#delete_timeslices_for_period

gabina:
## What this PR does
Add ACUWT row deletion to the `delete_timeslices_for_period` method. This method is called when a timeslice is split because it has more than 10K revisions. Before this PR, ACUWT rows created for the previous start and end dates were left orphaned when the timeslice was split. This did not affect any production courses, since those timeslices were split the first time they were processed.

## AI usage
Done with Claude.

### DIFF
--- a/lib/timeslice_cleaner.rb
+++ b/lib/timeslice_cleaner.rb
@@ -65,6 +65,7 @@ class TimesliceCleaner
     delete_course_wiki_timeslices_for_period(wikis, start_date, end_date)
     delete_course_user_wiki_timeslices_for_period(wikis, start_date, end_date)
     delete_article_course_timeslices_for_period(wikis, start_date, end_date)
+    delete_article_course_user_wiki_timeslices_for_period(wikis, start_date, end_date)
   end
 
   # Deletes course wiki timeslices records with a date prior to the current start date
@@ -277,6 +278,17 @@ class TimesliceCleaner
     timeslices = CourseUserWikiTimeslice.where(course: @course).where(wiki: wikis)
                                         .where('start >= ?', start_date)
                                         .where('end <= ?', end_date)
+    delete_in_batches(timeslices)
+  end
+
+  # Deletes article course user wiki timeslices records in the period [start_date, end_date].
+  # ACUWT rows are keyed by (course, wiki, article, user, start, end), so rows written for a
+  # period that no longer exists as a timeslice (for example, after that period was split)
+  # would otherwise linger and keep contributing to the aggregations derived from ACUWT.
+  def delete_article_course_user_wiki_timeslices_for_period(wikis, start_date, end_date)
+    timeslices = ArticleCourseUserWikiTimeslice.where(course: @course).where(wiki: wikis)
+                                               .where('start >= ?', start_date)
+                                               .where('end <= ?', end_date)
     delete_in_batches(timeslices)
   end
 
--- a/spec/lib/timeslice_cleaner_spec.rb
+++ b/spec/lib/timeslice_cleaner_spec.rb
@@ -114,6 +114,12 @@ describe TimesliceCleaner do
       end: end_date)
       create(:article_course_timeslice, course:, article: article3, start: end_date,
       end: end_date + 1.day)
+      create(:article_course_user_wiki_timeslice, course:, wiki: wikidata, article: article2,
+      user_id: 1, start: start_date, end: end_date)
+      create(:article_course_user_wiki_timeslice, course:, wiki: wikidata, article: article3,
+      user_id: 1, start: end_date, end: end_date + 1.day)
+      create(:article_course_user_wiki_timeslice, course:, wiki: enwiki, article: article1,
+      user_id: 1, start: start_date, end: end_date)
       course.reload
     end
 
@@ -128,6 +134,17 @@ describe TimesliceCleaner do
       expect(course.course_wiki_timeslices.size).to eq(221)
       expect(course.course_user_wiki_timeslices.size).to eq(2)
       expect(course.article_course_timeslices.size).to eq(2)
+    end
+
+    it 'deletes article course user wiki timeslices for the given period and wikis only' do
+      expect(ArticleCourseUserWikiTimeslice.where(course:).count).to eq(3)
+
+      timeslice_cleaner.delete_timeslices_for_period([wikidata], start_date, end_date)
+
+      # The wikidata record for a later period, and the enwiki record for the same period,
+      # both remain
+      expect(ArticleCourseUserWikiTimeslice.where(course:, wiki: wikidata).count).to eq(1)
+      expect(ArticleCourseUserWikiTimeslice.where(course:, wiki: enwiki).count).to eq(1)
     end
   end
 


ragesoss (approved):
The change looks correct and is symmetric with the three sibling `_for_period` methods. Notes below, none blocking.

A general note on how to use this feedback: it comes out of a Claude Code session and should not be automatically trusted or acted on. It's worth reading and understanding, then using or ignoring based on your own judgment.

**The impact is larger than the description suggests, and worth capturing.**

The description frames the old rows as orphaned. They also actively corrupt CWT stats. `CourseWikiTimeslice` selects ACUWT by `(course, wiki, start)` with no `end` predicate — `acuwt_live_article_records` (`app/models/course_wiki_timeslice.rb:162`) and `acuwt_mainspace_tracked_student_records` (`:172`). After `[start, end]` is split into `[start, mid]` and `[mid, end]`, the orphan's `start` is the first half's `start`, so the first half's CWT folds the whole pre-split period into its own totals.

Reproduced locally against a course with `use_acuwt`, one student and one mainspace article — an orphan row carrying `character_sum: 500, references_count: 9, revision_count: 11` alongside the first half's legitimate `100 / 1 / 2`:

| | no orphan | with orphan |
|---|---|---|
| `character_sum` | 100 | 600 |
| `references_count` | 1 | 10 |
| `revision_count` | 2 | 13 |

ACT (`article_course_timeslice.rb:117`) and CUWT (`course_user_wiki_timeslice.rb:63`) both match on `start` **and** `end`, so neither was ever affected. That asymmetry seems worth noting in the code comment, since it explains why CWT is the only casualty.

One downstream consequence worth checking: `mw_rev_count` is also derived from `acuwt_live_article_records`, and the comment at `course_wiki_timeslice.rb:144-147` notes it feeds `CourseRevisionUpdater#new_revisions?`. An inflated value there could plausibly mask genuinely new revisions. Not verified in this session.

**"Did not affect any production courses" may be narrower than "cannot happen."**

The mechanism checks out: `handle_timeslice` calls `maybe_split` before `process_leaf_timeslice` (`app/services/update_course_wiki_timeslices.rb:114-124`), so ACUWT rows only pre-exist when the period was already processed as a leaf. But that appears reachable — a timeslice processed below the threshold, later re-marked `needs_update`, accumulating past 10K on a subsequent run. If that's right, the distinction matters for whether existing data needs a cleanup pass, rather than only the forward fix.

**Stale comment in the caller.** `app/services/split_timeslice.rb:43` still reads "We also need to delete ACT and CUWT associated to that wiki and dates."

**Considered and appears fine:** deleting ACUWT before the halves are reprocessed can leave them empty for a run when `should_update_timeslice?` is false, but the new CWT rows are created with `needs_update: true` and re-derive on a later cycle — and this matches how ACT/CUWT/CWT already behaved before this PR. Also, the new deletion runs for courses without `use_acuwt?`; that looks harmless given `index_acuwt_on_course_id_and_wiki_id`, and desirable for any course switched off ACUWT.

**Exploratory spec.** Written during review to pin the over-count down as reproducible fact rather than an argument about code paths. All four examples pass on this branch; the last one fails on `master` (2 rows survive instead of 0), so it's a direct before/after check on the fix. Offered for adoption if useful — the first two examples are the ones that document the CWT/ACT asymmetry.

```ruby
# frozen_string_literal: true

require 'rails_helper'
require_dependency "#{Rails.root}/lib/timeslice_cleaner"

# Demonstrates the concrete harm of ACUWT rows orphaned by a timeslice split:
# CourseWikiTimeslice aggregates ACUWT by (course, wiki, start) WITHOUT matching
# `end`, so an orphan row for the pre-split period [start, end] is picked up by the
# post-split slice [start, midpoint], which shares the same `start`.
describe 'ACUWT rows orphaned by a timeslice split' do
  let(:start_date) { Time.zone.parse('2024-01-01') }
  let(:midpoint) { Time.zone.parse('2024-01-03') }
  let(:end_date) { Time.zone.parse('2024-01-05') }
  let(:course) do
    create(:course, start: start_date, end: end_date, flags: { use_acuwt: true })
  end
  let(:user) { create(:user) }
  let(:article) do
    create(:article, wiki: enwiki, namespace: Article::Namespaces::MAINSPACE, deleted: false)
  end

  before do
    # The post-split first half, and its legitimately re-derived ACUWT row.
    CourseWikiTimeslice.create!(course:, wiki: enwiki, start: start_date, end: midpoint)
    ArticleCourseUserWikiTimeslice.create!(course:, wiki: enwiki, article:, user_id: user.id,
                                          start: start_date, end: midpoint,
                                          character_sum: 100, references_count: 1,
                                          revision_count: 2)
  end

  # The orphan: written for the pre-split period, so it shares `start` with the
  # first half but has a different `end`.
  def create_orphan
    ArticleCourseUserWikiTimeslice.create!(course:, wiki: enwiki, article:, user_id: user.id,
                                          start: start_date, end: end_date,
                                          character_sum: 500, references_count: 9,
                                          revision_count: 11)
  end

  it 'aggregates the first half correctly when no orphan exists' do
    CourseWikiTimeslice.update_from_acuwt(course, enwiki, start_date, midpoint)

    cwt = CourseWikiTimeslice.find_by(course:, wiki: enwiki, start: start_date, end: midpoint)
    expect(cwt.character_sum).to eq(100)
    expect(cwt.references_count).to eq(1)
    expect(cwt.revision_count).to eq(2)
  end

  it 'over-counts the first half when an orphan shares its start date' do
    create_orphan

    CourseWikiTimeslice.update_from_acuwt(course, enwiki, start_date, midpoint)

    cwt = CourseWikiTimeslice.find_by(course:, wiki: enwiki, start: start_date, end: midpoint)
    # The orphan's stats are folded into the first half's totals.
    expect(cwt.character_sum).to eq(600)
    expect(cwt.references_count).to eq(10)
    expect(cwt.revision_count).to eq(13)
  end

  it 'does not leak the orphan into ACT, which matches on both start and end' do
    create_orphan

    ArticleCourseTimeslice.bulk_update_from_acuwt(course, enwiki, start_date, midpoint)

    act = ArticleCourseTimeslice.find_by(course:, article:, start: start_date, end: midpoint)
    expect(act.character_sum).to eq(100)
  end

  it 'removes the orphan via delete_timeslices_for_period, restoring correct aggregation' do
    create_orphan

    TimesliceCleaner.new(course).delete_timeslices_for_period([enwiki], start_date, end_date)

    # The orphan is gone; the first half's own ACUWT row was inside the deleted
    # period too, so it is re-derived when the half is reprocessed.
    expect(ArticleCourseUserWikiTimeslice.where(course:).count).to eq(0)
  end
end
```

Verified locally on this branch: RuboCop clean on both changed files, and 61 examples pass across `timeslice_cleaner_spec`, `split_timeslice_spec`, `update_course_wiki_timeslices_spec`, and `update_timeslices_course_user_spec`.

*Drafted in a Claude Code session (~10 min, 2 interactions). Sage requested the review and approved posting after reading a summary, but did not independently verify the findings or read the spec line by line. The session itself checked out the branch and ran locally: RuboCop on the changed files, the affected spec files listed above, and the exploratory spec on both this branch and `master` to confirm the before/after difference. The exploratory spec above is AI-generated code; it was run but has not been line-by-line reviewed by a human. The `mw_rev_count` / `new_revisions?` consequence and the reachability argument for previously-processed timeslices are reasoned from reading the code, not exercised by a test, and warrant independent confirmation.*

PR #6974, opened 2026-07-28
add composite index on articles(id, namespace)

lakshita10341:
## Summary
Adds a composite index on `articles(id, namespace)` to optimize bulk CSV export data fetching (`CampaignCsvBuilder#revision_counts` on `master` / `SystemCsvBuilder#fetch_revision_counts`).

## Rationale
During bulk CSV generation, queries join `article_course_timeslices` to `articles` to filter by `articles.namespace IN (0, 1, 2)`.

Without a covering index on `(id, namespace)`, MySQL reads full ~92-byte article rows off disk to access `namespace`. Because the covering index entry for `(id, namespace)` is significantly smaller (~14.1 bytes/row vs ~92 bytes/row for full rows), a covering index packs entries far more densely per page. This eliminates roughly **1.48 GB of disk reads** across large database scans by allowing MySQL to read `namespace` directly from the index.

---

## Production Measurements & Performance Impact

Measured on the P&E Dashboard production database (19.3M articles, 79.6M timeslices across a 9.65 GB I/O pass):

| Metric | Baseline (No Index) | With `articles(id, namespace)` | Impact / Migration |
|---|---|---|---|
| **Production I/O Scan** | 9.65 GB read | **~8.17 GB read** | **~15% reduction in disk I/O (~3.6 min saved)** |
| **Index Migration Time (19.3M rows)** | — | **~20 seconds** | Online / INPLACE (`add_index` safe for deploy) |
| **Disk Storage Added** | — | **~270 MB** | Low storage footprint |

---

##  Why Not Add the 4-Column Covering Index on `article_course_timeslices`?

While a covering index on `article_course_timeslices` (`course_id, tracked, article_id, revision_count`) could yield further I/O reductions on `ACT`:
**High Write Amplification:** `article_course_timeslices` is a high-churn, write-hot table updated continuously every 5 minutes during active course syncs. Because `revision_count` changes constantly, every update forces MySQL to re-index entries in the B-tree.

## Testing
- Ran migration `db/migrate/20260728203000_add_index_to_articles_on_id_and_namespace.rb` and verified `db/schema.rb`.

### DIFF
--- a/db/migrate/20260728203000_add_index_to_articles_on_id_and_namespace.rb
+++ b/db/migrate/20260728203000_add_index_to_articles_on_id_and_namespace.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddIndexToArticlesOnIdAndNamespace < ActiveRecord::Migration[7.0]
+  def change
+    add_index :articles, %i[id namespace], name: 'index_articles_on_id_and_namespace'
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_07_15_120000) do
+ActiveRecord::Schema[8.1].define(version: 2026_07_28_203000) do
   create_table "admin_course_notes", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.integer "courses_id"
     t.string "title"
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -99,6 +99,7 @@ ActiveRecord::Schema[8.1].define(version: 2026_07_15_120000) do
     t.virtual "index_hash", type: :string, as: "if(`deleted`,NULL,concat(`mw_page_id`,_utf8mb4'-',`wiki_id`))", stored: true
     t.index ["index_hash"], name: "index_articles_on_index_hash", unique: true
     t.index ["mw_page_id"], name: "index_articles_on_mw_page_id"
+    t.index ["id", "namespace"], name: "index_articles_on_id_and_namespace"
     t.index ["namespace", "wiki_id", "title"], name: "index_articles_on_namespace_and_wiki_id_and_title"
   end
 


ragesoss (commented):
Splitting the index change out into its own PR made this much easier to evaluate — the analysis below would not have been practical against the full #6968 diff.

The notes below come from a Claude Code review session. They are offered as observations to weigh rather than conclusions to act on directly; AI analysis is provisional and some of it may be wrong. Please use your own judgment on each point, and push back where the reasoning doesn't hold.

## The index does help — but by about 15%, not 2.5–3×

The query was measured on the Programs & Events Dashboard production database (MariaDB, 12 GB buffer pool, `articles` ≈ 19.3M rows / 4.2 GB, `article_course_timeslices` ≈ 79.6M rows / 15.1 GB). A single cold `CampaignCsvBuilder#revision_counts` call across all nonprivate courses:

```
run 1: 1451.21 s, physical reads 632575, groups 9018
```

632,575 × 16 KiB = **~9.65 GiB read from disk** in one pass. That closely matches the predicted clustered footprint of the two tables involved:

| Structure | Predicted |
|---|---|
| ACT clustered data (79.6M × ~109 B) | 8.08 GiB |
| `articles` clustered data | 1.74 GiB |
| **Total** | **9.82 GiB** |

Predicted 9.82 GiB vs. measured 9.65 GiB. So the query reads both tables' clustered data essentially in full, from disk, with effectively no cache benefit — 6.8 MiB/s effective throughput, i.e. entirely I/O-bound on random reads.

That lets the I/O be attributed by table: the ACT scan is ~84% of it, and the `articles` lookups this PR targets are **~18%**. Replacing that 1.74 GiB of clustered access with a ~270 MB covering index removes roughly 1.48 GiB, or ~15% of total I/O. For an I/O-bound query, that projects to about **1451 s → 1230 s** — a real gain of roughly 3.6 minutes, but a fifth of what the PR description claims.

Note that this 15% is a projection from measured page counts under a linear time∝I/O assumption, not a direct before/after measurement on production. It has not been verified by actually creating the index there.

## The mechanism in the description appears to be incorrect

The description states that MySQL "must fetch full article rows off disk just to read `namespace`." `EXPLAIN` on master shows `articles` reached via `type: eq_ref, key: PRIMARY`. In InnoDB the clustered index leaf page *is* the row, so `namespace` arrives on that same page read — there is no secondary-index bookmark lookup to eliminate. That cost model describes a non-primary access path, which isn't the plan here.

The mechanism that does appear to operate is density. An `(id, namespace)` entry measured 14.1 bytes/row when built locally, against ~92 bytes/row for the clustered row, so the covering index packs far more entries per page. Locally (`articles` at 28.8M rows, deliberately cache-starved) that produced:

```
before:  type: eq_ref   key: PRIMARY                             Extra: Using where
after:   type: ref      key: index_articles_on_id_and_namespace   Extra: Using where; Using index
```

with interleaved A/B runs (`IGNORE INDEX` to hold cache state comparable) of 4.0/4.5/5.5 s → 2.3/2.5/2.7 s, physical reads 33,732 → 23,459, and byte-identical results. That ~2× is *not* transferable to production — that bench had a 128 MB buffer pool and only 886K ACT rows, so it over-weighted the `articles` term relative to the real workload. The 15% figure above supersedes it.

This distinction matters practically, because the correct mechanism predicts when the gain appears: it scales with how much of `articles` is *not* resident. That is also why the benchmark in the PR description shows nothing. At 50,000 articles the table fits entirely in the buffer pool, leaving no physical reads to eliminate — hence 0.990 s → 0.970 s, a 2% delta within noise. The same table then labels that row "~2.5–3× faster on production disk I/O," a figure no data in the PR supports. Worth either re-running at a scale where `articles` exceeds `innodb_buffer_pool_size`, or dropping the table.

## The cited code is not on master

`SystemCsvBuilder#fetch_revision_counts` exists only in #6968, which hasn't merged. The equivalent query *is* on master as `CampaignCsvBuilder#revision_counts` (`lib/analytics/campaign_csv_builder.rb:54`) — same join, same `namespace IN (0,1,2)` filter — and is reachable synchronously through `AnalyticsController#all_courses_csv`. All measurements above used that method. Citing it instead would let this PR stand on its own rather than depending on #6968 landing first.

## Migration cost

Measured against a local copy of `articles` at 28.8M rows: `ADD INDEX` took **29.9 s** and added **407 MB** (14.1 bytes/row); `DROP INDEX` 0.1 s. Scaling to peony's 19.3M rows projects to **~20 s and ~270 MB**. InnoDB `ADD INDEX` is INPLACE/online on MariaDB, so concurrent DML is permitted throughout, and `add_index` inside `change` is automatically reversible. This looks safe to run inside a normal Capistrano deploy on either host.

Two costs the description doesn't mention, both probably acceptable but worth stating: ~270 MB of disk, and index maintenance on every `articles` INSERT and on every `namespace` UPDATE (page moves via `ArticleNamespacesManager`). `articles` is write-hot during course updates.

## A higher-value variant of the same idea

The measurement above shows 84% of the I/O is the ACT scan, which this index leaves untouched. The same covering-index trick applies there, and to a much larger term. The ACT side of this query needs only `course_id`, `tracked`, `article_id`, `revision_count` — an index on those four would carry ~13 bytes of payload plus the 8-byte PK, so roughly 2.2 GiB in place of scanning 8.08 GiB.

Projected on the same linear assumption: ~9.4 min instead of 24.2 min, versus ~20.6 min for the index in this PR. Both together would be ~6 min. This is unmeasured byte math rather than a benchmark, and ACT is a larger and more write-hot table, so it warrants more caution and its own testing — but it appears to be the higher-leverage change by roughly 4×, and it may be worth pursuing before or instead of this one.

## Relevant to #6968

`groups: 9018` corresponds to roughly 3,000 courses across three namespaces. `SystemCsvBuilder` batches at `BATCH_SIZE = 500` and calls `fetch_revision_counts` once per batch, so a system-wide export issues 6+ passes, each with `course_id IN (500 ids)` scattered across a 79.6M-row table. Given that one unbatched pass measured 24 minutes, it's worth measuring what the batched form costs on peony before that export ships — whether the optimizer turns each batch into range scans or a fresh full scan makes the difference between ~25 minutes and several hours, inside a Sidekiq job. No index on `articles` changes that shape.

*Drafted in a Claude Code session (~1 hour, ~10 interactions). Sage directed the review and ran all production measurements on the P&E Dashboard database himself in a Rails console, including the 24-minute query above; Claude Code produced and ran the local benchmark against a 28.8M-row copy of `articles` (index build timing, interleaved A/B query timings, `EXPLAIN` plans, buffer-pool read counts) and did the I/O attribution arithmetic. The local benchmark and the production page counts are measured; the 15% and the ACT-index projections are linear extrapolations from those counts and were not independently verified. Sage reviewed this comment before posting but did not re-derive the InnoDB clustered-index reasoning or the byte-width estimates, which are the parts most worth checking.*

PR #6975, opened 2026-07-28
Stop deleting derived timeslices when reaggregation

gabina:
## What this PR does
Reaggregation involves regenerating ACT, CUWT, and CWT rows from existing ACUWT rows. In most cases, we don't need to manually delete ACT and CUWT rows when marking a timeslice as `needs_reaggregation`, since they will be regenerated automatically during the reaggregation process. This PR avoids deleting derived timeslices when it isn't necessary.

Currently we mark timeslices as `needs_reaggregation` in the following places:

- New user added (`UpdateTimeslicesCourseUser`). There is no need to delete any timeslices; we only need to add the new ACUWT rows.
- User deleted (`UpdateTimeslicesCourseUser`). In this case, we need to delete the timeslices associated with the deleted user ID.
- Article manually excluded/reincluded (`UpdateTimeslicesUntrackedArticle`). We don't need to delete any timeslices in this case.
- ACUWT reprocessed (`ReprocessArticleCourseUserWikiTimeslices`). We don't need to delete timeslices in this case.
- Article excluded reset (`ArticlesCoursesCleaner`). This happens when an article is excluded because it changed namespace or because it fell out of scoped.  We don't need to delete the timeslices here, since the caches will be recalculated excluding this article.

## AI usage
Done with Claude Code.

### DIFF
--- a/app/services/reprocess_article_course_user_wiki_timeslices.rb
+++ b/app/services/reprocess_article_course_user_wiki_timeslices.rb
@@ -37,7 +37,7 @@ class ReprocessArticleCourseUserWikiTimeslices
       @course, @wiki, article_ids, ts_start
     )
     fetch_scores_and_update_acuwt(users, article_ids, ts_start, ts_end) if users.any?
-    reset_for_reaggregation(article_ids, ts_start)
+    mark_for_reaggregation(article_ids, ts_start)
   end
 
   def failing_article_ids_for(ts_start)
@@ -69,14 +69,15 @@ class ReprocessArticleCourseUserWikiTimeslices
     UpdateWikidataStatsTimeslice.new(@course).update_revisions_with_stats(live_revisions)
   end
 
-  # Marks the CWT for reaggregation (deleting the stale ACT/CUWT rows) and clears
-  # needs_update so the reaggregation pass picks it up. Reaggregation then rebuilds
-  # ACT, CUWT and CWT from ACUWT and re-derives needs_update from ACUWT state.
-  def reset_for_reaggregation(article_ids, ts_start)
+  # Marks the CWT for reaggregation and clears needs_update so the reaggregation pass picks it
+  # up. Reaggregation then rebuilds ACT, CUWT and CWT from ACUWT and re-derives needs_update
+  # from ACUWT state. The re-scored ACUWT rows are kept, so the rebuild covers every ACT and
+  # CUWT row of the period and there is nothing to delete first.
+  def mark_for_reaggregation(article_ids, ts_start)
     acuwt = ArticleCourseUserWikiTimeslice.where(
       course: @course, wiki: @wiki, start: ts_start, article_id: article_ids
     )
-    @timeslice_cleaner.reset_timeslices_for_reaggregation_from_acuwt(acuwt)
+    @timeslice_cleaner.mark_timeslices_for_reaggregation_from_acuwt(acuwt)
   end
 
   def revision_data_manager
--- a/app/services/reprocess_article_course_user_wiki_timeslices.rb
+++ b/app/services/reprocess_article_course_user_wiki_timeslices.rb
@@ -69,10 +69,8 @@ class ReprocessArticleCourseUserWikiTimeslices
     UpdateWikidataStatsTimeslice.new(@course).update_revisions_with_stats(live_revisions)
   end
 
-  # Marks the CWT for reaggregation and clears needs_update so the reaggregation pass picks it
-  # up. Reaggregation then rebuilds ACT, CUWT and CWT from ACUWT and re-derives needs_update
-  # from ACUWT state. The re-scored ACUWT rows are kept, so the rebuild covers every ACT and
-  # CUWT row of the period and there is nothing to delete first.
+  # Marks the CWT for the period as needing reaggregation, so that the re-scored ACUWT rows
+  # reach the derived caches (see TimesliceCleaner#mark_timeslices_for_reaggregation_from_acuwt).
   def mark_for_reaggregation(article_ids, ts_start)
     acuwt = ArticleCourseUserWikiTimeslice.where(
       course: @course, wiki: @wiki, start: ts_start, article_id: article_ids
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -215,19 +215,19 @@ class UpdateCourseWikiTimeslices
   end
 
   def update_article_course_user_wiki_timeslices_for_wiki(wiki, revisions)
-    timeslice = acuwt_timeslice_for(wiki, revisions)
+    timeslice = course_wiki_timeslice_for(wiki, revisions)
     ArticleCourseUserWikiTimeslice.bulk_upsert_from_revisions(
       @course, wiki, timeslice.start, timeslice.end, revisions[:revisions]
     )
   end
 
   def update_article_course_timeslices_from_acuwt_for_wiki(wiki, revisions)
-    timeslice = acuwt_timeslice_for(wiki, revisions)
+    timeslice = course_wiki_timeslice_for(wiki, revisions)
     ArticleCourseTimeslice.bulk_update_from_acuwt(@course, wiki, timeslice.start, timeslice.end)
   end
 
   def update_course_user_wiki_timeslices_from_acuwt_for_wiki(wiki, revisions)
-    timeslice = acuwt_timeslice_for(wiki, revisions)
+    timeslice = course_wiki_timeslice_for(wiki, revisions)
     revisions[:revisions].map(&:user_id).uniq.each do |user_id|
       CourseUserWikiTimeslice.update_from_acuwt(@course, user_id, wiki,
                                                timeslice.start, timeslice.end)
@@ -235,7 +235,7 @@ class UpdateCourseWikiTimeslices
   end
 
   def update_course_wiki_timeslices_from_acuwt_for_wiki(wiki, revisions)
-    timeslice = acuwt_timeslice_for(wiki, revisions)
+    timeslice = course_wiki_timeslice_for(wiki, revisions)
     CourseWikiTimeslice.update_from_acuwt(@course, wiki, timeslice.start, timeslice.end)
   end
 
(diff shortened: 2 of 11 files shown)


ragesoss (commented):
Review notes from a Claude Code session. As with any AI-assisted review, these are
observations to read and weigh, not conclusions to act on automatically — some of it may be
wrong, and the parts that are right are worth confirming independently before changing code.

## The ACT half of the argument checks out

`delete_article_course_timeslices_for_acuwt_pairs` deleted exactly the `(article_id, start)`
pairs drawn *from the ACUWT records passed in*, and `ArticleCourseTimeslice.bulk_update_from_acuwt`
upserts one row per article holding an ACUWT row in the window. The deleted set was therefore a
subset of the rebuilt set, and dropping the delete really does just turn an update into a
delete-plus-insert. Running mark→reaggregate and reset→reaggregate against the same fixture
produced identical ACT rows.

## The CUWT half is not symmetric

`delete_course_user_wiki_timeslices_for_acuwt_pairs` deleted **all** CUWT rows for the affected
`(wiki, start)` periods, not only the rows derivable from the ACUWT records passed in. The
rebuild is narrower: `reaggregate_timeslice_from_acuwt` iterates
`ArticleCourseUserWikiTimeslice.where(course:, wiki:, start:).pluck(:user_id).uniq`, so it only
rewrites CUWT rows for users that still hold an ACUWT row in that period.

A CUWT row whose user has no surviving ACUWT row in an affected period is therefore left in
place with its old values, where it previously would have been dropped. The stale value reaches
`CoursesUsers#character_sum_ms` and `#revision_count` through
`CoursesUsers#update_cache_from_timeslices`.

How reachable that shape is in practice is the open question, and the answer depends on ACUWT
invariants the author knows better than this session did. One path that appeared reachable:
`use_acuwt` is admin-toggleable via `CourseFlagsController::MANAGEABLE_FLAGS`, and flipping it
on without also setting `needs_update` skips `PrepareTimeslices#recreate_timeslices`, so
legacy-path CUWT rows survive into ACUWT operation. In-flow orphans — revisions with a nil
`article_id`, which `acuwt_records_from_revisions` rejects while
`update_course_user_wiki_timeslices_from_acuwt_for_wiki` still creates a CUWT row for the user —
come out zeroed by `update_from_acuwt`, so those look harmless.

Independent of reachability, the new doc comment on `mark_timeslices_for_reaggregation_from_acuwt`
states the guarantee without qualification:

> as long as the caller keeps its ACUWT rows, every row that could be stale is rewritten in place

That holds for ACT and not for CUWT. Whether or not any current caller can hit the gap, that
sentence is what a future caller will rely on when choosing between the two methods, so it may
be worth narrowing it to what the rebuild actually covers.

## Verification spec

The spec below was generated by Claude Code during the review and run locally — all five
examples pass on this branch. It runs the full mark→reaggregate cycle and contrasts it with
reset→reaggregate. `rebuilds the ACT rows identically whether marked or reset` is the example
confirming the ACT argument; `leaves the CUWT row of a user with no ACUWT row in the period
untouched` and `removes that same CUWT row when the reset variant is used` are the contrasting
pair for the gap above.

Offered as a reproduction, not as a suggested addition to the suite — the fixture constructs the
orphaned-CUWT shape directly rather than arriving at it through a real update, so it demonstrates
the mechanism without establishing that production can reach it.

<details>
<summary><code>spec/services/reaggregation_coverage_spec.rb</code></summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/data_cycle/update_debugger"
require "#{Rails.root}/lib/timeslice_cleaner"

describe 'reaggregation coverage after marking for reaggregation' do
  let(:start) { '2021-01-24'.to_datetime }
  let(:user1) { create(:user, username: 'Ragesoss') }
  let(:user2) { create(:user, username: 'Oleryhlolsson') }
  let(:article1) { create(:article, wiki: enwiki) }
  let(:course) do
    create(:course, start:, end: '2021-01-30', flags: { use_acuwt: true })
  end
  let(:manager) { TimesliceManager.new(course) }
  let(:reaggregator) do
    UpdateCourseWikiTimeslices.new(course, UpdateDebugger.new(course))
  end

  before do
    stub_wiki_validation
    JoinCourse.new(course:, user: user1, role: CoursesUsers::Roles::STUDENT_ROLE)
    JoinCourse.new(course:, user: user2, role: CoursesUsers::Roles::STUDENT_ROLE)
    manager.create_timeslices_for_new_course_wiki_records([enwiki])
    create(:articles_course, course:, article: article1)

    # Only user1 has an ACUWT row in this period.
    create(:article_course_user_wiki_timeslice,
           course:, article: article1, user: user1, wiki: enwiki,
           start:, end: start + 1.day, revision_count: 2, character_sum: 200)
    create(:article_course_timeslice, course:, article: article1, start:, end: start + 1.day)
    create(:course_user_wiki_timeslice,
           course:, user: user1, wiki: enwiki, start:, end: start + 1.day,
           character_sum_ms: 200, revision_count: 2)

    # user2 has a CUWT row in the same period but no ACUWT row backing it.
    create(:course_user_wiki_timeslice,
           course:, user: user2, wiki: enwiki, start:, end: start + 1.day,
           character_sum_ms: 9999, revision_count: 7)
  end

  # Runs one cleaner variant followed by the reaggregation pass, then yields so the caller
  # can snapshot the resulting rows. Rolls back so both variants start from the same state.
  def with_reaggregation(cleaner_method)
    snapshot = nil
    ActiveRecord::Base.transaction do
      TimesliceCleaner.new(course).public_send(
        cleaner_method, ArticleCourseUserWikiTimeslice.where(course:, wiki: enwiki, start:)
      )
      reaggregator.send(:reaggregate_timeslices_from_acuwt, enwiki)
      snapshot = yield
      raise ActiveRecord::Rollback
    end
    snapshot
  end

  it 'rewrites the CUWT row of a user that still has an ACUWT row in the period' do
    TimesliceCleaner.new(course).mark_timeslices_for_reaggregation_from_acuwt(
      ArticleCourseUserWikiTimeslice.where(course:, wiki: enwiki, start:)
    )
    reaggregator.send(:reaggregate_timeslices_from_acuwt, enwiki)

    row = CourseUserWikiTimeslice.find_by(course:, user: user1, wiki: enwiki, start:)
    expect(row.character_sum_ms).to eq(200)
  end

  it 'leaves the CUWT row of a user with no ACUWT row in the period untouched' do
    TimesliceCleaner.new(course).mark_timeslices_for_reaggregation_from_acuwt(
      ArticleCourseUserWikiTimeslice.where(course:, wiki: enwiki, start:)
    )
    reaggregator.send(:reaggregate_timeslices_from_acuwt, enwiki)

    row = CourseUserWikiTimeslice.find_by(course:, user: user2, wiki: enwiki, start:)
    expect(row).not_to be_nil
    expect(row.character_sum_ms).to eq(9999)
    expect(row.revision_count).to eq(7)
  end

  it 'removes that same CUWT row when the reset variant is used' do
    TimesliceCleaner.new(course).reset_timeslices_for_reaggregation_from_acuwt(
      ArticleCourseUserWikiTimeslice.where(course:, wiki: enwiki, start:)
    )
    reaggregator.send(:reaggregate_timeslices_from_acuwt, enwiki)

    expect(CourseUserWikiTimeslice.find_by(course:, user: user2, wiki: enwiki, start:)).to be_nil
  end

  it 'rebuilds the ACT rows identically whether marked or reset' do
    marked = with_reaggregation(:mark_timeslices_for_reaggregation_from_acuwt) do
      course.article_course_timeslices.order(:article_id, :start)
            .pluck(:article_id, :start, :revision_count, :character_sum)
    end
    reset = with_reaggregation(:reset_timeslices_for_reaggregation_from_acuwt) do
      course.article_course_timeslices.order(:article_id, :start)
            .pluck(:article_id, :start, :revision_count, :character_sum)
    end

    expect(marked).to eq(reset)
    expect(marked).to eq([[article1.id, start, 2, 200]])
  end

  it 'propagates the orphaned CUWT row into the CoursesUsers character sum cache' do
    TimesliceCleaner.new(course).mark_timeslices_for_reaggregation_from_acuwt(
      ArticleCourseUserWikiTimeslice.where(course:, wiki: enwiki, start:)
    )
    reaggregator.send(:reaggregate_timeslices_from_acuwt, enwiki)

    courses_user = CoursesUsers.find_by(course:, user: user2)
    courses_user.update_cache_from_timeslices
    expect(courses_user.character_sum_ms).to eq(9999)
  end
end
```

</details>

## Smaller notes

- `reprocess_article_course_user_wiki_timeslices.rb:72` — the reworded comment says the method
  "clears needs_update so the reaggregation pass picks it up", but
  `mark_timeslices_for_reaggregation` only sets `needs_reaggregation`, and nothing else in
  `mark_for_reaggregation` touches `needs_update`. This appears to be pre-existing wording
  carried into the rewrite, and looks harmless in practice (a CWT left at `needs_update: true`
  is skipped by `reaggregate_timeslices_from_acuwt` and handled by the normal fetch path
  instead), but the comment does not describe what the code does.
- The renamed `acuwt_timeslice_for` → `course_wiki_timeslice_for` reads as a straightforward
  improvement — the method returns a CourseWikiTimeslice.
- None of the five changed spec examples run the reaggregation pass, so the suite asserts the
  new intermediate state (rows survive the mark) without exercising the premise that the rebuild
  then covers them. An end-to-end example somewhere in the affected specs would pin that down.

## Checks run locally

- `bundle exec rubocop` on all 11 changed files — clean
- `spec/lib/article_namespaces_manager_spec.rb`, `spec/lib/articles_courses_cleaner_spec.rb`,
  `spec/lib/timeslice_cleaner_spec.rb`, `spec/services/reprocess_article_course_user_wiki_timeslices_spec.rb`,
  `spec/services/update_timeslices_scoped_article_spec.rb`,
  `spec/services/update_timeslices_untracked_article_spec.rb`,
  `spec/services/update_course_wiki_timeslices_spec.rb` — 83 examples, 0 failures
- `spec/services/update_course_stats_spec.rb`, `spec/services/update_timeslices_course_user_spec.rb`,
  and the four timeslice model specs — 84 examples, 0 failures

---

*Drafted in a Claude Code session (~25 min, 2 interactions). Sage asked for the review and read
the summary, but did not independently verify the findings or read the exploratory spec; the
spec was written and run by Claude Code. The reachability claim about the `use_acuwt` flag in
particular is a code-reading inference that has not been checked against production data.*

gabina:
It's an interesting review. The idea mentioned in the "CUWT half is not symmetric" is right, but it's no reachable. The path described is not reasonable (setting `use_acuwt` flag without setting `needs_update` is not something we're going to do). The key point is that the only place where we mark a timeslices for reaggregation deleting some ACUWT rows previously is when we delete a user, and in that case, we already call `reset_timeslices_for_reaggregation_from_acuwt` (manually deleting the CUWT rows).

I addressed the `mark_for_reaggregation` description comment mentioned in Smaller notes.

ragesoss:
okay. I'll go ahead and merge it, unless you'd like another round of review or you have any open concerns, @gabina.

gabina:
Ready for merging on my side.

Issue #6970, opened 2026-07-23
[ACUWT] Ensure timeslices are cleaned before reprocessing


gabina:
The current approach is to remove the timeslices before setting a row as `needs_update`. It could be better to just clean the timeslices before processing one timeslice with `needs_update` set to true. But we should check all the existing cases to be sure.

PR #6977, opened 2026-07-30
Clean stale timeslice rows before reprocessing a period

gabina:
## What this PR does

Addresses #6970. When a timeslice is reprocessed, the update re-fetches the whole period and rewrites it — but the writes only cover the `(article, user)` pairs the fetch returned. Rows for pairs whose revisions are no longer there survive and keep feeding the aggregated timeslices.

The cases that genuinely need this are the ones no marking site can predict, because nothing knows a revision is gone until a re-fetch reveals it:

- a revision deleted or suppressed on wiki while the page itself is still live (rows for deleted *pages* are kept on purpose, so they are unaffected);
- revisions re-attributed to a different `article_id` by a page move or history merge, where the old article is not flagged deleted.

It also happens to cover the course start/end date boundary period, whose re-fetch is truncated by `real_start`/`real_end`. That one is arguably better handled where the dates change, and is filed separately in issue #6978 (whether we're going to address it is still to be determined).
Removed course users are already handled at their own site by `delete_acuwt_for_deleted_course_users`.

How it works: `TimesliceCleaner#clean_timeslices_before_reprocessing` runs inside the `process_timeslices` transaction, only when `only_new` is false, and diffs the period's stored ACUWT rows against the fetched pairs — deleting only what the rewrite will not cover. A timeslice holds up to `SplitTimeslice::REVISION_THRESHOLD` (10k) revisions, so deleting the period wholesale would mean thousands of deletes and inserts on every reprocess; this way a reprocess that finds unchanged data deletes nothing, and surviving rows are updated in place, which preserves `tracked` (the one column `bulk_upsert_from_revisions` deliberately does not write). The ACT and CUWT deletes are keyed by the articles and users left with nothing for the period, since both unique indexes lead with `article_id` and `(course_id, user_id)` and would otherwise scan every row the course has. `ArticlesCoursesCleaner#remove_articles_courses_without_timeslices` then drops records for articles left without any timeslice in the course, after the rewrite rather than before. ACUWT courses only; the legacy path is untouched.

One related fix rides along: the CUWT rebuild now takes its user list from the period's stored ACUWT rows instead of the fetched revisions, so a user whose revisions are gone gets their row rebuilt rather than left stale.

## AI usage

Claude Code wrote the code, specs and both commit messages, under close and correction-driven direction. It began as analysis only, with the instruction to first verify the approach in #6970 was right. Several of its conclusions were wrong and were corrected: it claimed six `needs_update` sites were buggy (additive and retry-shaped marks cannot leave stale rows, so the real count was one); it proposed ACUWT-only scope, which was rejected, then a legacy branch was written and later removed once we decided that path is not worth hardening; the pair diff was dropped and then reinstated when the index layout and the 10k ceiling showed it was what keeps a reprocess cheap. The reviewer also rewrote the `process_timeslices` comments, renamed the write step to `write_timeslices`, and caught that the `articles_courses` cleanup must run after the rewrite.

Scale: two commits, but that understates it — one long interactive session dominated by design discussion rather than code volume (~350 lines including specs), after which three commits that belonged elsewhere were split into #6973 and #6975. The reviewer ran the suite; Claude ran RuboCop and the affected specs (69 examples) green.

The summary above was also drafted by Claude Code and may contain errors — please check it against the diff. This description was drafted using the `/prepare-pr` Claude Code skill.

## Screenshots

No UI changes (backend only).

- **The incremental path is deliberately not cleaned.** `only_new: true` runs many times a day per course, so cleaning there was ruled out on cost. If a revision is deleted on wiki and nothing marks the period `needs_update`, the orphan row survives — in practice `ArticleStatusManager` does mark, so it lands on the reprocess path anyway.

- **The `no_point_in_importing_revisions?` guard.** The clean is skipped when the fetch never asked the replica, which happens for an `ArticleScopedProgram` with no assignments and no categories. Such a course keeps its ACUWT rows on purpose while every fetch comes back empty, so without the guard a reprocess would read "not fetched" as "the period is empty" and wipe them permanently.

- **This does not remove the deletions elsewhere.** Each survives for a concrete reason: `reset_timeslices_for_update_from_acuwt` targets `deleted: true` articles, which this clean preserves by design; the reaggregation resets run on a path that never enters `process_timeslices`; `SplitTimeslice#delete_timeslices_for_period` uses pre-split period keys.

- **Possible index follow-up.** The period pluck filters `start`/`end` after an index range on `(course_id, wiki_id)` — no worse than what `act_records_from_acuwt` and `acuwt_live_article_records` already do per timeslice, but extending that index to include `start` would help all three.

### DIFF
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -202,7 +202,7 @@ class UpdateCourseWikiTimeslices
     return [] if @revision_updater.no_point_in_importing_revisions?
 
     revisions = @revisions[wiki]
-    timeslice = acuwt_timeslice_for(wiki, revisions)
+    timeslice = course_wiki_timeslice_for(wiki, revisions)
     @timeslice_cleaner.clean_timeslices_before_reprocessing(wiki, timeslice.start, timeslice.end,
                                                             revisions[:revisions])
   end
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -168,9 +168,9 @@ class UpdateCourseWikiTimeslices
   end
 
   # Processing timeslices involves:
-  # - deleting existing timeslices if reprocessing (only_new = false)
+  # - deleting existing timeslices if reprocessing (only_new = false) and use_acuwt? flag
   # - writing timeslices for the period on the db
-  # - deleting articles courses without timeslices if reprocessing
+  # - deleting articles courses without timeslices if reprocessing and use_acuwt? flag
   # - updating last mw rev datetime
   def process_timeslices(wiki, only_new:)
     @course.reload
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -2,6 +2,8 @@
 
 require_dependency "#{Rails.root}/lib/course_revision_updater"
 require_dependency "#{Rails.root}/lib/timeslice_manager"
+require_dependency "#{Rails.root}/lib/timeslice_cleaner"
+require_dependency "#{Rails.root}/lib/articles_courses_cleaner"
 require_dependency "#{Rails.root}/lib/data_cycle/update_debugger"
 require_dependency "#{Rails.root}/lib/revision_scanner"
 
@@ -16,6 +18,8 @@ class UpdateCourseWikiTimeslices
   def initialize(course, debugger, update_service: nil)
     @course = course
     @timeslice_manager = TimesliceManager.new(@course)
+    @timeslice_cleaner = TimesliceCleaner.new(@course)
+    @articles_courses_cleaner = ArticlesCoursesCleaner.new(@course)
     @splitter = SplitTimeslice.new(@course)
     @debugger = debugger
     @update_service = update_service
@@ -128,7 +132,7 @@ class UpdateCourseWikiTimeslices
     add_scores(only_new:)
     if should_update_timeslice?(wiki, only_new:)
       fetch_wikidata_stats(wiki) if wiki.project == 'wikidata'
-      process_timeslices(wiki)
+      process_timeslices(wiki, only_new:)
     end
     [start_date]
   end
@@ -163,15 +167,47 @@ class UpdateCourseWikiTimeslices
       @wikidata_stats_updater.update_revisions_with_stats(wikidata_revisions)
   end
 
-  def process_timeslices(wiki)
+  # Processing timeslices involves:
+  # - deleting existing timeslices if reprocessing (only_new = false)
+  # - writing timeslices for the period on the db
+  # - deleting articles courses without timeslices if reprocessing
+  # - updating last mw rev datetime
+  def process_timeslices(wiki, only_new:)
     @course.reload
     ActiveRecord::Base.transaction do
-      update_timeslices(wiki)
+      emptied_article_ids = only_new ? [] : clean_timeslices_before_reprocessing(wiki)
+      write_timeslices(wiki)
+      # The articles_courses cleanup has to run after rewriting the timeslices.
+      # Otherwise, it would drop the records of every article in the period.
+      @articles_courses_cleaner.remove_articles_courses_without_timeslices(emptied_article_ids)
       @timeslice_manager.update_last_mw_rev_datetime(@revisions)
     end
   end
 
-  def update_timeslices(wiki)
+  # A reprocess re-fetches the whole period and rewrites it, so rows for (article, user) pairs
+  # that no longer have revisions in the period have to go, or they would keep contributing to
+  # the aggregated timeslices. This is only done when reprocessing: the regular ingestion path
+  # runs many times a day for a single course, and there the fetch is not expected to return
+  # less data than the previous one.
+  #
+  # Only for ACUWT courses: the clean works by diffing the period's stored ACUWT rows against
+  # the fetched revisions. The legacy path is left as it was, since it is on its way out.
+  #
+  # Skipped when the fetch never asked the replica, which happens for an article scoped course
+  # with no assignments and no categories. Such a course keeps its ACUWT rows on purpose (see
+  # ArticlesCoursesCleaner#reset_excluded) while every fetch comes back empty, so cleaning
+  # would read 'not fetched' as 'the period has no revisions' and wipe them for good.
+  def clean_timeslices_before_reprocessing(wiki)
+    return [] unless @course.use_acuwt?
+    return [] if @revision_updater.no_point_in_importing_revisions?
+
+    revisions = @revisions[wiki]
+    timeslice = acuwt_timeslice_for(wiki, revisions)
+    @timeslice_cleaner.clean_timeslices_before_reprocessing(wiki, timeslice.start, timeslice.end,
+                                                            revisions[:revisions])
+  end
+
+  def write_timeslices(wiki)
     revisions = @revisions[wiki]
     if @course.use_acuwt?
       update_article_course_user_wiki_timeslices_for_wiki(wiki, revisions)
@@ -228,10 +264,19 @@ class UpdateCourseWikiTimeslices
 
   def update_course_user_wiki_timeslices_from_acuwt_for_wiki(wiki, revisions)
     timeslice = acuwt_timeslice_for(wiki, revisions)
-    revisions[:revisions].map(&:user_id).uniq.each do |user_id|
+    acuwt_user_ids(wiki, timeslice).each do |user_id|
       CourseUserWikiTimeslice.update_from_acuwt(@course, user_id, wiki,
                                                timeslice.start, timeslice.end)
     end
+  end
+
+  # Users come from the period's stored ACUWT rows rather than from the fetched revisions: a
+  # user whose revisions are no longer in the period still needs their CUWT row rebuilt, or it
+  # would keep stale values. This is what reaggregate_timeslice_from_acuwt already does.
+  def acuwt_user_ids(wiki, timeslice)
+    ArticleCourseUserWikiTimeslice.where(course: @course, wiki:, start: timeslice.start,
+                                         end: timeslice.end)
+                                  .distinct.pluck(:user_id)
   end
 
   def update_course_wiki_timeslices_from_acuwt_for_wiki(wiki, revisions)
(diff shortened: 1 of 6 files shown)


gabina:

ragesoss (changes requested):
This review was produced in a Claude Code session. It is not meant to be trusted or acted on automatically — read it, check the reasoning against the code, and use or discard each point on your own judgment. The one finding below was reproduced locally against both branches; the rest are observations that warrant verification.

## A failed replica fetch is indistinguishable from an empty period, and now deletes the data

`Replica#api_get` gives up after its three retries and returns `nil` (`lib/replica.rb:121-133`). It reports to Sentry through `log_error`, which ends in an explicit `return nil` and does not raise (`lib/errors/api_error_handling.rb:17`). `skip_replica_failure` also returns `nil` deliberately, for a wiki whose replica views are missing. `Replica#get_revisions` then turns that `nil` into an empty result set, silently:

```ruby
raw = get_revisions_raw(users, rev_start, rev_end)
@data = {}
return @data unless raw.is_a?(Enumerable)   # lib/replica.rb:32
```

So `revisions[:revisions]` reaches `process_timeslices` as `[]`. On the reprocess path `should_update_timeslice?` is true regardless of `new_data`, so `clean_timeslices_before_reprocessing` runs against an empty `fetched_pairs` and every stored row for the period is stale. Before this PR the same empty fetch was a no-op, because `bulk_upsert_from_revisions` returns early on `records.empty?`.

Partial failures look the same. `Utils.chunk_requests` (`lib/utils.rb:6-14`) batches users ten at a time and `concat`s each chunk's result, so one failed chunk drops just those users' revisions from the fetch — and only those users' rows get deleted.

Either way the loss is permanent within the normal cycle: the rewrite clears `needs_update` on the CWT, so nothing schedules another attempt.

A blanket "skip the clean when the fetch came back empty" guard would blunt part of what this PR is for, since a genuinely emptied period is one of the cases it targets. Distinguishing the two probably needs a fetch-succeeded signal propagated up from `Replica` / `RevisionDataManager`, which is a larger change than this PR.

### Reproduction

The spec below was run on `review/6977` merged with current `master` (**4 examples, 4 failures** — the ACUWT rows, the ACT row and the `articles_courses` record are all gone, and `needs_update` is back to `false`) and on `master` alone (**4 examples, 0 failures**). It stubs `Replica#get_revisions` to return `{}`, which is the exact value the silent-failure path produces.

<details>
<summary><code>spec/services/zz_review_6977_replica_failure_spec.rb</code></summary>

```ruby
# frozen_string_literal: true

# Exploratory spec written during review of PR #6977. Not intended to be merged as-is.
#
# Replica#get_revisions swallows a failed request: api_get returns nil after its three
# retries (and skip_replica_failure returns nil outright for a wiki whose replica views
# are missing), and get_revisions turns that nil into an empty result set. A reprocess
# therefore cannot tell "the period really has no revisions" from "the fetch failed".

require 'rails_helper'
require_dependency "#{Rails.root}/lib/timeslice_manager"
require_dependency "#{Rails.root}/app/services/update_course_wiki_timeslices"

describe 'reprocessing a period when the replica request fails' do
  let(:course) do
    create(:basic_course, start: '2018-11-24 00:00:00', end: '2018-11-30 23:55:00')
  end
  let(:user) { create(:user, username: 'Ragesoss') }
  let(:article) { create(:article, wiki_id: enwiki.id) }
  let(:timeslice) { course.course_wiki_timeslices.where(wiki: enwiki).first }
  let(:updater) { UpdateCourseWikiTimeslices.new(course, UpdateDebugger.new(course)) }

  before do
    stub_wiki_validation
    travel_to Date.new(2018, 12, 1)
    JoinCourse.new(course:, user:, role: 0)
    TimesliceManager.new(course).create_timeslices_for_new_course_wiki_records(course.wikis)

    # Stored data for the period, as a previous successful update would have left it.
    create(:articles_course, course:, article:)
    create(:article_course_user_wiki_timeslice, course:, wiki: enwiki, article:,
           user_id: user.id, start: timeslice.start, end: timeslice.end,
           revision_count: 3, character_sum: 500)
    create(:article_course_timeslice, course:, article:, start: timeslice.start,
           end: timeslice.end, revision_count: 3, character_sum: 500)
    create(:course_user_wiki_timeslice, course:, wiki: enwiki, user_id: user.id,
           start: timeslice.start, end: timeslice.end, revision_count: 3)
    timeslice.update(needs_update: true)

    # This is exactly what Replica#get_revisions returns when api_get gives up: an empty
    # @data hash, with no exception raised.
    allow_any_instance_of(Replica).to receive(:get_revisions).and_return({})
    allow_any_instance_of(Replica).to receive(:get_existing_revisions_by_id).and_return(nil)
    allow_any_instance_of(CourseRevisionUpdater).to receive(:fetch_scores_for_revisions)
    allow(RevisionScanner).to receive(:schedule_revision_checks)
  end

  after { travel_back }

  it 'keeps the stored ACUWT rows for the period' do
    updater.run(all_time: false)

    expect(ArticleCourseUserWikiTimeslice.where(course:, article:, user_id: user.id,
                                                start: timeslice.start).count).to eq(1)
  end

  it 'keeps the aggregated article course timeslice for the period' do
    updater.run(all_time: false)

    expect(ArticleCourseTimeslice.where(course:, article:, start: timeslice.start).count).to eq(1)
  end

  it 'keeps the articles_courses record for the article' do
    updater.run(all_time: false)

    expect(course.articles_courses.where(article:).count).to eq(1)
  end

  it 'does not leave the period permanently emptied' do
    updater.run(all_time: false)

    # needs_update is cleared by the rewrite, so nothing schedules another attempt.
    expect(timeslice.reload.needs_update).to eq(false)
    expect(ArticleCourseUserWikiTimeslice.where(course:, start: timeslice.start)).not_to be_empty
  end
end
```

</details>

## Secondary observations

**The deleted-article carve-out is only as good as `Article.deleted`.** `should_update_article_status?` (`app/services/update_course_stats.rb:145-157`) never runs `ArticleStatusManager` for `only_scoped_articles_course?` courses, and skips it on P&E for courses whose updates run longer than ten minutes — with a comment that says "This means we miss cases of article deletion." For those courses a page deleted on wiki is never flagged, so the carve-out does not cover it and the rows are deleted on the next reprocess. Whether that matters depends on whether the "keep the rows so undeletion is cheap" guarantee is meant to hold for those courses.

**The `no_point_in_importing_revisions?` comment is narrower than the guard.** It describes only the article-scoped-with-no-assignments-or-categories case, but the guard also returns true whenever `@course.students.empty?` (`lib/course_revision_updater.rb:17`). The behaviour is safe either way; the comment reads as exhaustive when it is not.

**Query count on the reprocess path.** It now runs three separate `(course_id, wiki_id)` index range scans over ACUWT per timeslice — the clean's `stored` pluck, `act_records_from_acuwt`, and the new `acuwt_user_ids` — where it previously ran one. The ingestion path gains one, since `acuwt_user_ids` replaced an in-memory `map(&:user_id).uniq`. The index follow-up already noted in the PR description looks worth doing before this runs on large P&E courses.

**Spec coverage of the empty-fetch path.** The new context in `spec/services/update_course_wiki_timeslices_spec.rb` sets up a fetch that returns `revisions: []` and asserts the cleaner is called with `[]`, stubbing `.and_return([])`. That pins down the call but not what the call does, so the case where the failure lands is not exercised.

Everything else read as intended: the index choices for the ACT and CUWT deletes, the ordering of the `articles_courses` cleanup after the rewrite, the set arithmetic in `emptied_article_and_user_ids`, and the `tracked` column on ACT (only written on the legacy path, so not at risk on the ACUWT path).

RuboCop is clean on all six changed files. The PR's own specs plus the existing specs for the affected areas pass: 69 examples, 0 failures.

---

*Drafted in a Claude Code session (~25 min, 2 interactions). Claude Code did the code reading, wrote the exploratory spec above, and ran it on both branches along with RuboCop and the affected specs; the pass/fail results reported here are from those runs. Sage read the session summary and approved posting, but did not independently reproduce the finding or line-by-line review the spec. The spec is AI-written and is offered as something to rerun, not as something verified by a human.*

gabina:
Interesting. It's true that a failed replica fetch is indistinguishable from an empty period. In fact, there are specs exercising that behavior in `spec/lib/replica_spec.rb`, and it seems to have been implemented that way on purpose a long time ago.

I agree that being unable to distinguish between failed requests and legitimate empty ones is not good for this PR, and we should be able to distinguish between those cases before implementing the row deletions.

I'm going to close this PR for now, since it isn't fixing anything specific to the new ACUWT branch. It also affects the current behavior. I'll open an issue to distinguish between failed revision requests and legitimate empty responses so this doesn't fall through the cracks, but not to handle it urgently.

On a side note, I'm trying to learn something from using Claude Code through this PR. It took me several hours to implement this (it also involved https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6973 and https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/6975). I spent a lot of time discussing it with Claude Code, and there was a lot of back and forth. After each small point we discussed, the AI would say, "Something that's worth noting..." and then write a huge paragraph about related side effects (and sometimes things that weren't even that related) that I had to think through, which ended up confusing me more and more. In the end, neither the AI nor I noticed this key point about emptying a timeslice just because the API failed, which is pretty sensitive.

The point is that I really think we've put a lot of careful thought into making the course update work reasonably well. Of course, there are bugs, performance improvements to implement, and things that can be made better in general. But I wouldn't want the use of artificial intelligence to undermine that: something which, despite its flaws, was carefully thought out. After all, with every PR that uses AI, the code generally becomes larger and less understandable to me (since I didn't write it line by line), which makes me feel more distant from truly understanding it. I think losing that understanding is undesirable, at least for now (regardless of me, of course).

ragesoss:
> But I wouldn't want the use of artificial intelligence to undermine that: something which, despite its flaws, was carefully thought out. After all, with every PR that uses AI, the code generally becomes larger and less understandable to me (since I didn't write it line by line), which makes me feel more distant from truly understanding it. I think losing that understanding is undesirable, at least for now (regardless of me, of course).

I agree very strongly with this. The codebase overall is already more complex that I can keep up with understanding, and for new feature areas that can be somewhat separated from existing core code paths, I'm okay with not understanding AI-written features like I would have if I had written it. But the update cycle has some core concepts that are relatively straightfoward and understandable, even though the implementation has a lot of complexity in the details.

It might be worth spending some time documenting and describing the core human-understandable aspects that we want to preserve. That could provide some guardrails against AI changes that take off in unexpected directions, and also be very useful for other developers who work on this part of the system in the future.

gabina:
> It might be worth spending some time documenting and describing the core human-understandable aspects that we want to preserve. That could provide some guardrails against AI changes that take off in unexpected directions, and also be very useful for other developers who work on this part of the system in the future.

I think that makes sense. I would like to do that.

PR #6427, opened 2025-07-25
Ensure no duplicate article titles for categories

gabina:
## What this PR does
Closes #6415 

Note that categories with duplicate articles will be deduplicated in the next refreshing (once this is deployed).
## Open questions and concerns
< anything you learned that you want to share, or questions you're wondering about related to this PR >

PR #6979, opened 2026-07-31
Memoize Course#scoped_article_titles and store it as a Set

gabina:
## What this PR does

`Course#scoped_article_titles` rebuilt its list on every call, and `Course#scoped_article?`
calls it once per article of every fetched timeslice. For an `ArticleScopedProgram` tracking
a large scope, that dominated the timeslice-processing loop: on a production-shaped course
with ~580k scoped titles, a single scope check cost ~102 ms, of which ~91 ms was rebuilding
the array (concat + `uniq`) and only ~4 ms was the lookup itself.

This PR memoizes the list per wiki and stores it as a `Set` instead of a uniq'd `Array`:

- **Memoize** — the list is now built once per `Course` object (so once per update) instead of
  once per article. This is where most of the win is.
- **`Set` instead of `uniq`** — a `Set` is inherently unique, so this replaces the dedup pass
  rather than adding one, and turns the lookup from a linear scan into a hash probe.

Measured on a dev database against two real production course shapes (a 229k-title scope and
a 579k-title scope), running the real `scoped_article?` code path:

| | 229k titles | 579k titles |
|---|---:|---:|
| `scoped_article?` per article, before | 28.3 ms | 102.5 ms |
| ├ of which: rebuilding the array | 25.8 ms | 91.5 ms |
| └ of which: `Array#include?` scan | 1.8 ms | 4.1 ms |
| `Set#include?` | 0.0004 ms | 0.007 ms |

For a timeslice touching ~6,300 distinct articles, that is ~645 s of scope checking before
this PR and ~0 after.

A note on why the `Set` conversion is correct now but was not in #6427: when `uniq` was added
there, a `Set` benchmarked *slower*, because the collection was rebuilt on every call and
building a `Set` costs ~0.57 s more than `uniq` at 579k titles. Without memoization a `Set`
pays that build cost on every single lookup. With memoization the build happens once and the
break-even is ~142 lookups, which a large scoped course crosses in the first few percent of
one timeslice.

## AI usage

Claude Code was used throughout, under the user's direction.

The user asked Claude Code to investigate why one article-scoped Wikidata course completed a
full update in under three days while a comparable one had been running for over a week.
Claude Code did the analysis (public dashboard JSON endpoints, the update-stage timings, and
a benchmark harness run against a dev database) and identified this method as the largest
per-timeslice cost. The user wrote the memoization commit; Claude Code wrote the `Set`
conversion, the explanatory code comment, and the four specs, and drafted this description.

The user directed the scope at each step, and pushed back on Claude Code's suggestions more
than once — Claude Code proposed adding a `clear_scoped_article_memos` invalidation hook, and
the user correctly rejected it as unreachable code, since the only caller of
`Category.refresh_categories_for` runs before anything reads the scope. The user also supplied
the #6427 benchmark that prompted the build-vs-lookup analysis above. The user ran RuboCop and
the specs; Claude Code did not.

This was developed across 2 commits within about a 45-minute session, on top of a longer
investigation session that preceded it. The "What this PR does" summary and the measurements
quoted above were drafted by Claude Code and may contain errors.

## Screenshots

No UI changes.

- The benchmark numbers come from a dev database. The scope-check figures are pure Ruby with
  no database access, so they should transfer to production closely; anything database-bound
  would not.
- This does not explain the whole gap between the two courses it came out of. Measured
  per-timeslice costs still account for well under half of the observed time for the slow
  course, so investigation continues separately.
- The memoized value lives for the lifetime of the `Course` object, which spans a whole
  update. That is intentional and matches the two lookups it is built from
  (`assigned_article_titles` and `category_article_titles`), which were already memoized the
  same way. `Category.refresh_categories_for` runs before anything reads the scope, so the
  memo cannot be built from stale categories.
- Adjacent, not addressed here: `Course#scoped_article_ids` is rebuilt on every call and is
  spliced into SQL as an `IN` list roughly once per user per timeslice by
  `CourseUserWikiTimeslice.update_from_acuwt` and `CourseWikiTimeslice#update_cache_from_acuwt`.
  On the same dev benchmark that costs 16.7 s plus 2.1 s per user per timeslice at 579k ids,
  and it scales worse than linearly with the size of the scope.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -382,10 +382,12 @@ class Course < ApplicationRecord
     article_course_timeslices.where(article_id: scoped_article_ids)
   end
 
+  # A Set rather than a uniq'd Array because scoped_article? looks a title up once per
+  # article of every fetched timeslice; the pricier build only pays off thanks to the memo.
   def scoped_article_titles(wiki)
     @scoped_article_titles ||= {}
     @scoped_article_titles[wiki] ||= (assigned_article_titles(wiki) + category_article_titles(wiki))
-                                     .uniq
+                                     .to_set
   end
 
   def assigned_article_titles(wiki)
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -383,7 +383,9 @@ class Course < ApplicationRecord
   end
 
   def scoped_article_titles(wiki)
-    (assigned_article_titles(wiki) + category_article_titles(wiki)).uniq
+    @scoped_article_titles ||= {}
+    @scoped_article_titles[wiki] ||= (assigned_article_titles(wiki) + category_article_titles(wiki))
+                                     .uniq
   end
 
   def assigned_article_titles(wiki)
--- a/spec/models/article_scoped_program_spec.rb
+++ b/spec/models/article_scoped_program_spec.rb
@@ -187,6 +187,10 @@ describe ArticleScopedProgram, type: :model do
       expect(course.scoped_article_titles(wiki)).to equal(course.scoped_article_titles(wiki))
     end
 
+    it 'returns a Set, so that looking a title up does not scan the whole scope' do
+      expect(course.scoped_article_titles(wiki)).to be_a(Set)
+    end
+
     it 'memoizes per wiki rather than for the whole course' do
       other_wiki = Wiki.get_or_create(language: 'es', project: 'wikipedia')
       course.scoped_article_titles(wiki)
--- a/spec/models/article_scoped_program_spec.rb
+++ b/spec/models/article_scoped_program_spec.rb
@@ -162,4 +162,42 @@ describe ArticleScopedProgram, type: :model do
       expect(course.scoped_article?(wiki, 'Unassigned_article', mw_page_id)).to eq(false)
     end
   end
+
+  # scoped_article? calls scoped_article_titles once per article of every fetched timeslice,
+  # so rebuilding the list on each call is the dominant cost for a course with a large
+  # scope. The memo below is what keeps that list from being rebuilt.
+  describe '#scoped_article_titles' do
+    before do
+      stub_wiki_validation
+      create(:assignment, course:, article:, article_title: article.title, wiki:)
+      create(:categories_courses, course:, category:)
+    end
+
+    let(:wiki) { Wiki.get_or_create(language: 'en', project: 'wikipedia') }
+    let(:course) { create(:article_scoped_program, start: '2018-01-01', end: '2018-12-31') }
+    let(:category) { create(:category, wiki:, article_titles: ['Category_article']) }
+    let(:article) { create(:article, title: 'Assigned_article') }
+
+    it 'includes both assigned and category article titles' do
+      expect(course.scoped_article_titles(wiki))
+        .to contain_exactly('Assigned_article', 'Category_article')
+    end
+
+    it 'returns the same object on repeated calls, without rebuilding the list' do
+      expect(course.scoped_article_titles(wiki)).to equal(course.scoped_article_titles(wiki))
+    end
+
+    it 'memoizes per wiki rather than for the whole course' do
+      other_wiki = Wiki.get_or_create(language: 'es', project: 'wikipedia')
+      course.scoped_article_titles(wiki)
+      expect(course.scoped_article_titles(other_wiki)).to be_empty
+    end
+
+    it 'picks up category changes for a freshly loaded course' do
+      course.scoped_article_titles(wiki)
+      category.update(article_titles: %w[Category_article New_category_article])
+      expect(Course.find(course.id).scoped_article_titles(wiki))
+        .to include('New_category_article')
+    end
+  end
 end


gabina:

ragesoss (approved):
Approving. A note on how this feedback is meant to be used: it was produced with heavy Claude Code involvement, and it should not be automatically trusted or acted on. It is worth reading and understanding, then using or ignoring based on your own judgment.

**What was checked locally**

The branch was merged with current master and the following were run: the PR's own specs plus `visiting_scholarship_spec`, `article_scoped_flag_spec`, `revision_data_manager_spec`, `revision_data_manager_merge_scoping_spec`, `update_wikidata_stats_timeslice_spec` (57 examples, 0 failures), and a wider regression set of `course_spec`, `course_user_wiki_timeslice_spec`, `course_wiki_timeslice_spec`, `update_course_wiki_timeslices_spec`, `category_spec` (160 examples, 0 failures, 2 pre-existing pendings). RuboCop is clean on both changed files; `yarn build` succeeds.

The performance claim reproduces. An exploratory spec was run on both this branch and master: 2,000 out-of-scope `scoped_article?` calls over a 30,000-title scope took **2,566 ms on master and 3.4 ms on this branch**, and an example asserting the category list is built once fails on master with 10 builds against 1 here. At the 579k scale in the description the ratio should be larger.

The `Array` → `Set` swap appears behavior-preserving. Because `Array#include?` compares with `==` while `Set#include?` uses `hash`/`eql?`, the cases where those diverge for strings were checked directly: UTF-8 versus ASCII-8BIT for an ASCII-only title, the same for a non-ASCII title, and a `String` subclass. All agree. Seven equivalence examples — including a batch comparison of `Set#include?` answers against the literal pre-PR `(assigned + category).uniq` answers — pass identically on both branches.

The memo also appears to add no new staleness window, for a reason worth stating explicitly since it makes the change safer than it looks: both inputs were already memoized per `Course` object with no invalidation anywhere, so `scoped_article_titles` was already effectively constant for the lifetime of an object. The ordering claim in the description also checks out — `UpdateCourseStats#initialize` calls `update_categories` → `Category.refresh_categories_for` before `UpdateCourseWikiTimeslices` and `update_wikidata_stats`, which are the only paths that reach `scoped_article?`.

**Non-blocking observations**

1. `scoped_article?` still linearly scans an Array on its other branch. `assigned_article_page_ids(wiki).include?(mw_page_id)` runs for every article that misses the title Set, which is the common case for a large scope, and the same one-line `.to_set` applies. Measured on synthetic integer arrays: negligible at 1k assignments, ~0.2 s per 6,300-article timeslice at 20k, ~1.9 s at 200k. Minor next to what this PR removes, and it only matters for assignment-heavy scopes — flagged because it is the same idiom left half-applied.

2. The commit message on d2c0b6d reads "Memoize scoped_article_ids", but the commit memoizes `scoped_article_titles`. `scoped_article_ids` is specifically the method the PR does not memoize, per its own Open Questions. Since merge commits are preserved in this repo, that message will point future `git log` and `git blame` readers at the wrong method.

3. `VisitingScholarship#scoped_article_titles` still returns an Array, so the new "returns a Set" spec now codifies a contract for the base method that a subclass does not follow. Nothing calls anything but `include?`, so this is harmless today. Worth noting as a trap for later, though: the obvious consistency fix of `assigned_article_titles(wiki).to_set` would build a fresh Set on every call and reintroduce exactly the cost this PR removes, so that override would need its own memo.

4. Retained memory rises, and given recent memory pressure on the Programs & Events infrastructure it is worth recording even though the trade clearly favors this change. At 579k titles the Set's internal hash measured 32 MB, and process RSS high-water rose about 94 MB while building it and did not return after the reference was dropped. Against that, the pre-PR code allocated roughly 8.8 MB of transient arrays per call, thousands of times per timeslice, so total GC pressure drops sharply. If peak allocation is ever a concern, `assigned_article_titles(wiki).to_set.merge(category_article_titles(wiki))` avoids the 579k-element intermediate array, though it does not change the retained figure.

5. Spec nit: the new `describe '#scoped_article_titles'` block repeats all four `let` definitions from the sibling `#scoped_article?` block verbatim, and its `stub_wiki_validation` is only needed by the one example that creates the `es` wiki.

One calibration note on the description's numbers: the absolute build costs did not reproduce here — `to_set` measured ~181 ms against `uniq` at ~50 ms for 579k titles, rather than the ~570 ms gap implied. The break-even argument is unaffected either way; these figures put it near 60 lookups instead of 142, which a large scoped course still crosses almost immediately.

**Suggested coverage**

The PR's specs cover the memo well, but nothing would catch a future regression back to per-call rebuilding at scale. The scale example below is the one from the exploratory spec used during this review; it fails on master and passes here, so it pins the actual property the PR exists for. It was written by Claude Code and is offered as a starting point rather than as reviewed code.

```ruby
describe 'at scale' do
  scope_size = 30_000
  let(:big_titles) { Array.new(scope_size) { |i| "Scoped_article_number_#{i}" } }
  let(:category) { create(:category, wiki:, article_titles: big_titles) }

  # Pre-memo this loop rebuilt and linearly scanned a 30k-title Array 2,000 times.
  # The bound is deliberately ~100x looser than the measured post-memo time.
  it 'checks 2,000 out-of-scope articles quickly' do
    course.scoped_article?(wiki, 'warm_the_memo', 1)
    elapsed = Benchmark.realtime do
      2_000.times { |i| course.scoped_article?(wiki, "Unscoped_#{i}", 500_000 + i) }
    end
    expect(elapsed).to be < 2.0
  end
end
```

A timing assertion is a judgment call for the committed suite; the alternative that avoids wall-clock entirely is asserting the build count, which also fails on master:

```ruby
it 'builds the category title list once across many scope checks' do
  allow(course).to receive(:category_article_titles).and_call_original
  10.times { |i| course.scoped_article?(wiki, "Article_#{i}", 100 + i) }
  expect(course).to have_received(:category_article_titles).once
end
```

*Drafted in a Claude Code session (~20 min, 3 interactions). Claude Code ran all of the spec runs and measurements described above, including the master-versus-branch contrast, the Set-versus-Array semantics checks, and the memory figures. Sage requested the review and read the resulting summary, but did not independently verify any of the findings, did not review the exploratory spec code, and approved posting on the strength of the summary alone.*

Issue #6910, opened 2026-06-16
Claim-verification exercise (v1): subject-matched claim pool + sandbox handoff


ragesoss:
# Claim-verification exercise (v1): subject-matched claim pool + sandbox handoff

## Summary

A new student activity that builds information-literacy skills around source
verification. The Dashboard assigns a student a real, cited factual claim drawn
from a Wikipedia article — matched to the student's course subject — and hands
them off to their own Wikipedia sandbox to check whether the cited source
actually supports the claim and record what they find.

The Dashboard's entire job in v1 is **selection and handoff**. The verification
work, and everything the student writes, happens on-wiki in their sandbox. The
Dashboard does not ingest, store, or grade the student's findings.

## Pedagogical model

1. The student is shown one assigned claim, its paragraph context, and the
   source cited for it (with a link out to the source).
2. They open their sandbox, pre-seeded with the claim, the source details, and a
   worksheet section to fill in.
3. They read the source, judge whether it supports the claim, and write up their
   assessment in the sandbox — graded/discussed off-Dashboard, as the instructor
   sees fit.

No LLM sits anywhere in this loop. The student is the judge.

## Scope boundary

**In scope for v1**

- A harvested **pool** of candidate claim/source pairs, each tagged with subject
  metadata.
- A background process that **populates and refreshes** the pool, biased toward
  the subject areas of currently active courses.
- **Matching + assignment**: choose a subject-relevant claim for each student and
  record which claim they were given (assignment only — not their answers).
- A **student-facing display** of the assigned claim and its source.
- A **"start in your sandbox"** handoff link that pre-seeds the sandbox.

**Explicitly out of scope for v1**

- Any in-Dashboard verification UI.
- Any storage of the student's judgment / findings / rationale.
- Any grading, scoring, or instructor-facing review of findings.
- Any LLM call in the student request path.
- Auto-detecting completion (cheap to add later — see Future work).

## Where the pieces come from

The `ClaimVerification` branch already contains the deterministic extraction
engine. v1 reuses it and adds the pool, matching, and handoff around it.

**Reused as-is (deterministic, no LLM):**

- `ClaimCitationExtractor` / `ExtractClaimsAndSources` (`:structural` mode) —
  one claim per cited sentence, paired to its citation.
- `GetRevisionHtmlWithCitations` + `WikiApi::DiffWikitextExtractor` — fetch
  article/diff HTML with citations intact.
[...]

PR #6981, opened 2026-08-05
Refine the fact-verification exercise: source-evaluation step, highlighting fix

ragesoss:
## What this PR does

Three related changes to the in-dashboard fact-verification exercise (#6910).

**A new source-evaluation step.** Before the student goes looking for the source, they now judge it from the citation alone: is this an appropriate kind of source for this topic and claim, and does it meet [Wikipedia's Reliable Sources policy](https://en.wikipedia.org/wiki/Wikipedia:Reliable_sources) — each a multiple choice plus a "why or why not". It becomes step 3; finding and verifying the source renumber to 4 and 5 on their own.

**The exercise's questions no longer live in the database schema.** Adding the step as a column per question would have meant a migration for every future refinement of the exercise's wording, options or ordering. Instead the answers moved into a serialized hash (the same pattern as `Alert#details`), and what the exercise asks is declared in `config/claim_verification_exercise.yml` — with model validation, permitted params, the React form and the instructor summary all derived from that one file. Adding or reordering a question is now that file plus locale copy: no migration, no model change, no JavaScript change. The migration carries existing answers across, and back out again on rollback.

**A highlighting bug fix.** In production the highlighted text in an article was often not the claim that came up when you clicked it. A claim's position was being located by walking backward from a citation marker counting characters, without ever checking that the text collected was the claim's sentence. That breaks when one source is cited on several sentences (every claim's data landed on the first of them), when a citation sits mid-sentence (the sentence continues past the marker, so backward-walking can never cover it), and silently in both cases because nothing verified the result. Claims are now located by finding their sentence text in the paragraph prose and wrapping exactly that range. On the article this was reported against, highlights went from 24 of 28 disagreeing with their claim to 0 of 28; across six (article, revision) pairs from a production dump — 391 highlights — 0 mismatches. No re-harvest is needed, since claims already store the sentence text the matching uses.

**Plus:** "Choose a different claim" had no `button` class (it rendered as bare text) and sat below the entire form, so it was easy to miss. It moves up beside the heading. That exposed a dead end behind it — the only ways out of the picker were taking a different claim or reloading the page — so the picker now offers "← Back to my claim" whenever the student already has one.

### Testing

87+ Ruby specs (model, request, lib, service) and 6 browser feature specs pass, along with the 413-test JS suite; RuboCop and ESLint are clean on every touched file. The three highlighting regression tests were checked against the old implementation to confirm they actually fail there. The migration's data round-trip was verified against real rows in a local production dump, and the diagnosis throughout was done against real production data rather than synthetic fixtures.

## AI usage

Claude Code wrote essentially all of the code in this PR, along with the commit messages and this description. It should be reviewed on that basis — particularly the new `ProseIndex` / `SentenceHighlighter` DOM-range logic, which is the least obvious code here.

What the human contributed: the decision to add the step and what it should ask; **all** user-facing copy (per the no-AI-text rule — Claude left `[PLACEHOLDER - …]` markers for every new string and they were filled in by hand, including "Back to my claim"); the architectural call that rejected Claude's first implementation for coupling the exercise's contents to the database structure, and the direction to use a serialized hash instead; the report of the highlighting bug with a specific production URL, plus the correct hypothesis that it involved sources used multiple times; and the choice of where the relocated button should go.

Claude's mistakes along the way are recorded candidly in the individual commit messages, and are worth skimming: two buggy diagnostic scripts (one of which briefly produced evidence *against* the approach that was ultimately correct), a crash on the first version of the range-wrapping code, a second bug caught by Claude's own new spec, and a regression test that initially passed against the old implementation for the wrong reason.

All three commits were made in a single session on 2026-08-05, spanning about two hours from first to last. That is fast for the amount of code here, which is a reason for reviewers to look closely rather than a sign of maturity.

This PR description was drafted using the `/prepare-pr` Claude Code skill, and — like the summary above — may contain errors.

## Screenshots

Captured on both branches with assets rebuilt on each side, so the "before" column is master's real behaviour.

**01 — a claim selected, in an article citing one source three times**

Before, the panel says the claim is *"Sea otters are a keystone species of the kelp forest ecosystem"* while the text actually highlighted is *"re hunted extensively fo…"* — a different sentence, and starting mid-word. After, the highlight matches the claim.

| Before | After |
|--------|-------|
| before | after |

**02 — the taken claim**

The new source-evaluation step, and "Choose a different claim" moved up beside the heading.

| Before | After |
|--------|-------|
| before | after |

**03 — the picker, after choosing to switch claims**

| Before | After |
|--------|-------|
| before | after |

**Deployment.** The migration drops six per-question columns in favour of a serialized `answers` hash. Existing responses are carried across (and back on rollback). The only responses in production so far are staff test submissions, but the copy-forward is there in case real student ones arrive before this deploys.

**Stale translations.** Step numbering is now composed in rather than written into the copy, so the English step names lost their "Step 3: " / "Step 4: " prefixes. That left the `nl` and `sk` translations carrying numbering of their own, which would have rendered as "Step 4: Stap 3: Zoek de bron", so those two strings were edited to drop it. Translatewiki will resync.

**Answer option values are identifiers.** They're stored in the hash, so rewording an option's label is free, but renaming a value after students have submitted leaves old answers keyed to the retired value. The summary falls back to displaying the raw value rather than hiding it, so nothing disappears, but old and new wouldn't aggregate. The scales are cheap to revise now, while the only responses are staff tests.

**Two pre-existing issues found but deliberately not fixed here:**

- A harvested sentence on *Toxic colonialism* is malformed — `'…"Waste colonialism." .'`, with a stray trailing space-period, apparently from diff-mode segmentation at harvest time — so it matches nothing in the full revision and is never highlighted. This was equally true before this PR.
- Pool claims outnumber selectable ones: 49 rows collapse to 28 distinct sentences on the article above, because `harvested_claims` keys by sentence and keeps the first. For a sentence carrying several citations, only one claim is ever offered. That looks intentional, but if each (sentence, source) pair should be its own verification task, that's a separate change.

**Also unchanged:** taking a different claim moves the single assignment while the earlier response stays with its own claim, so the instructor view lists that student as both answered and pending.

### DIFF
--- a/app/assets/javascripts/components/claim_verification_exercise/ArticlePicker.jsx
+++ b/app/assets/javascripts/components/claim_verification_exercise/ArticlePicker.jsx
@@ -20,12 +20,23 @@ import formatRevisionDate from '../../utils/format_revision_date';
 //
 // The shell's `?showArticle=` deep link is article-keyed; when one article has
 // several flagged-revision tiles we auto-open the first of them on a deep link.
-export const ArticlePicker = ({ articles, course, onTaken, showArticleId }) => {
+//
+// `onReturnToClaim` is given only when the student already has a claim, and is
+// the way back out of the picker without taking a different one — browsing the
+// candidates is not meant to cost them the claim they already have.
+export const ArticlePicker = ({ articles, course, onTaken, showArticleId, onReturnToClaim }) => {
   const showWiki = new Set(articles.map(toWikiDomain)).size > 1;
   const deepLinkedTile = articles.find(article => article.id === showArticleId);
 
   return (
     <div className="container narrow claim-verification-exercise claim-verification-exercise--articles">
+      {onReturnToClaim && (
+        <div className="cv-preview-return">
+          <button type="button" className="button" onClick={onReturnToClaim}>
+            ← {I18n.t('claim_verification.back_to_claim')}
+          </button>
+        </div>
+      )}
       <div className="claim-verification-exercise__intro">
         <p>{I18n.t('claim_verification.intro_p1')}</p>
         <p>{I18n.t('claim_verification.intro_p2')}</p>
@@ -88,6 +99,8 @@ ArticlePicker.propTypes = {
   onTaken: PropTypes.func.isRequired,
   // Article id from the ?showArticle= deep link, if any; auto-opens that tile.
   showArticleId: PropTypes.number,
+  // Present only when the student already has a claim to go back to.
+  onReturnToClaim: PropTypes.func,
 };
 
 export default ArticlePicker;
--- a/app/assets/javascripts/components/claim_verification_exercise/ClaimVerificationExercise.jsx
+++ b/app/assets/javascripts/components/claim_verification_exercise/ClaimVerificationExercise.jsx
@@ -37,7 +37,7 @@ import InstructorResponses from './InstructorResponses.jsx';
   "choosing" is kept as local state rather than a second, competing query param.
 */
 const ClaimVerificationExercise = ({ course }) => {
-  const [state, setState] = useState(null); // { assignment, response, articles }
+  const [state, setState] = useState(null); // { assignment, response, form, articles }
   const [choosing, setChoosing] = useState(false);
   const responsesPage = useLocation().pathname.endsWith('/responses');
   const dispatch = useDispatch();
@@ -46,7 +46,7 @@ const ClaimVerificationExercise = ({ course }) => {
     if (responsesPage) { return; }
     new ClaimVerificationAPI({ courseSlug: course.slug }).fetchState()
       .then(setState)
-      .catch(() => setState({ assignment: null, response: null, articles: [] }));
+      .catch(() => setState({ assignment: null, response: null, form: null, articles: [] }));
   }, [course.slug, responsesPage]);
 
   if (responsesPage) { return <InstructorResponses course={course} />; }
@@ -78,6 +78,7 @@ const ClaimVerificationExercise = ({ course }) => {
     return (
       <TakenClaim
         assignment={state.assignment}
+        form={state.form}
         response={state.response}
         courseSlug={course.slug}
         onChooseDifferent={() => setChoosing(true)}
--- a/app/assets/javascripts/components/claim_verification_exercise/ClaimVerificationExercise.jsx
+++ b/app/assets/javascripts/components/claim_verification_exercise/ClaimVerificationExercise.jsx
@@ -39,6 +39,10 @@ import InstructorResponses from './InstructorResponses.jsx';
 const ClaimVerificationExercise = ({ course }) => {
   const [state, setState] = useState(null); // { assignment, response, form, articles }
   const [choosing, setChoosing] = useState(false);
+  // Bumped whenever we change ?showArticle= ourselves. The id is read fresh from
+  // the URL on each render rather than held in state (the viewer shell owns that
+  // param), so clearing it only takes effect once something re-renders us.
+  const [, urlChanged] = useState(0);
   const responsesPage = useLocation().pathname.endsWith('/responses');
   const dispatch = useDispatch();
 
@@ -64,6 +68,16 @@ const ClaimVerificationExercise = ({ course }) => {
     window.history.replaceState(null, null, window.location.pathname);
     setState(prev => ({ ...prev, assignment, response: response || null }));
     setChoosing(false);
+  };
+
+  // Leave the picker for the claim the student already has. Browsing the other
+  // candidates shouldn't cost them their claim (or, once they've answered, send
+  // them back to a blank form), so this undoes "choose a different claim" and
+  // also drops any ?showArticle= they arrived on.
+  const returnToClaim = () => {
+    window.history.replaceState(null, null, window.location.pathname);
+    setChoosing(false);
+    urlChanged(tick => tick + 1);
   };
 
   const afterResponseSaved = (response) => {
@@ -93,6 +107,7 @@ const ClaimVerificationExercise = ({ course }) => {
       course={course}
       onTaken={afterTaken}
       showArticleId={showArticleId}
+      onReturnToClaim={state.assignment ? returnToClaim : null}
     />
   );
 };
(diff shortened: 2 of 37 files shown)


ragesoss (commented):
Reviewed with Claude Code. What follows is AI-generated analysis: worth reading and understanding, but not worth trusting automatically — take or leave each item on your own judgment. Some of it may simply be wrong.

No blocking objections; this reads as ready to merge. One item is worth fixing first — the locale regression in (1). Everything after it is optional.

**Verified locally against 51a4214** as it stands — the branch is 5 commits behind master and was *not* merged forward before testing, so nothing here speaks to interaction with those commits. RuboCop and ESLint clean on every changed file; `yarn build` compiles; 115 Ruby specs (lib, model, request, service) and the 6 browser feature specs pass; 413 JS tests pass. The migration was run down and back up again against a seeded row in a local test database. The highlighting was benchmarked and its edge cases reproduced in isolation. The production-dump result quoted in the description (391 highlights, 0 mismatches across six article/revision pairs) was **not** re-run, and is taken on the description's word.

---

### 1. `step_heading` is untranslated, so `nl` and `sk` now render half-English

Rendering the numbered headings in each locale that has this exercise's copy:

```
en: ["Step 3: Evaluate the source", "Step 4: Find the source", "Step 5: Verify the claim"]
nl: ["Step 3: Evaluate the source", "Step 4: Zoek de bron", "Step 5: Verifieer de bewering"]
sk: ["Step 3: Evaluate the source", "Step 4: Nájdite zdroj", "Step 5: Overte tvrdenie"]
```

On master, `nl` read "Stap 3: Zoek de bron" — fully Dutch. Composing the number in moved the word "Step" into `step_heading`, which exists only in `en.yml`, so both existing translations fall back to English for it. The new step falling back entirely is expected and fine; this is a different thing, and it's a regression rather than the stale-translation cleanup the description describes.

No new copy is needed to fix it — the translators' own words are in the lines this PR removes (`nl` had "Stap 3: ", `sk` had "Krok 3: "), so `step_heading: 'Stap %{number}: %{name}'` and `'Krok %{number}: %{name}'` reuse existing translator wording rather than inventing any.

### 2. A new `required:` question retroactively invalidates every stored response

A response carried forward by the migration, checked after running it:

```
valid?: false
errors: ["Answers source_appropriate is required", "Answers meets_rs_policy is required"]
save:   false
```

Harmless today: nothing re-saves a response in place, and a student editing theirs is held by `missingRequired` until they answer the new questions, so the flow degrades gracefully. But the point of the config file is to make adding a question routine, and each `required: true` addition silently turns every prior response unsaveable. Next time that may land on real student data rather than staff tests. Worth deciding now whether requiredness belongs as a submission-time gate rather than a record invariant.

### 3. Rollback is lossy, and the backfill rewrites `updated_at`

The round trip carries the six original columns across faithfully in both directions. But `source_appropriate`, `meets_rs_policy` and `source_appropriate_notes` have no columns to return to, so `down` discards them irrecoverably. "Existing responses are carried across (and back on rollback)" reads as fully reversible; it is reversible only for what predates this PR.

Separately, `updated_at` moved on each leg of the round trip (`22:07:55` → `22:08:22` → `22:08:57`), because the backfill uses `update!`. `update_columns` would preserve when the student actually submitted.

**Migration shape** otherwise looks safe inside a normal deploy: a tiny table, one nullable text column added and six dropped, no index or constraint work. The one thing worth stating rather than leaving implicit is that Capistrano runs migrations before the restart, so the old code is briefly live against a table missing the six columns it believes exist. On a feature with no traffic that is acceptable — but it is a real ordering hazard, not a non-issue.

### 4. Highlight over-wrap when a sentence boundary sits inside inline markup

`wrap_nodes`' comment says this "can take in a little adjacent text". It can take in a complete adjacent sentence:

```
input:     <p>He wrote <i>Otters Today. Volume Two</i> in 1990[1]. It sold well[2].</p>
claim:     "Volume Two in 1990."
highlight: "Otters Today. Volume Two in 1990."
```

Rare — it needs a sentence-final period inside an `<i>`/`<a>`/`<b>` — and still far better than what it replaces. Worth a spec pinning the behaviour, because of the follow-on: if the swallowed sentence is itself a pool claim, whichever is wrapped first blocks the other through `already_claimed?`, dropping it to marker-only tagging.

### 5. Highlighting cost, measured

`ProseIndex` is rebuilt per (claim × paragraph), and `prose_paragraphs` is recomputed per claim:

```
 40 paragraphs /  20,754 chars: 28 hits 0.14s / 28 misses 0.16s
120 paragraphs /  74,951 chars: 28 hits 0.29s / 28 misses 0.55s
200 paragraphs / 167,135 chars: 28 hits 0.61s / 28 misses 1.17s
```

"Misses" is the worst case: a claim whose sentence is nowhere in the prose (the *Toxic colonialism* case in the description) scans every paragraph. Against a MediaWiki API round trip this is noise, so not a blocker. Hoisting `prose_paragraphs(doc)` out of the per-claim loop is free; caching the indexes is not, since wrapping mutates the DOM.

### Smaller observations

- **No JS tests for `formDefinition.js`.** `visibleQuestions`, `missingRequired`, `answeredQuestions` and `displayValue` are pure functions with real branch logic, covered only through the browser feature spec. `test/` has no claim-verification coverage at all.
- **Nothing guards that a declared question has locale copy.** All ten resolve in en/nl/sk today (checked). But a config edit without the matching locale entry ships `translation missing: …` to students, and that is precisely the workflow this design invites. A spec walking `form.questions` and asserting no `translation missing` is a few lines.
- **No migration spec.** The round trip was run by hand and works; CI does not cover it.
- **`dangerouslySetInnerHTML` on translated copy.** `utils/markdown_it` forces `html: true` (merged after `opts`, so callers cannot opt out), meaning raw HTML in any translation renders. An established pattern in this codebase, but previously for operator and user content rather than locale strings arriving from Translatewiki.
- **`ProseIndex#reference_marker?` uses `class.include?('reference')`** — a substring test — while `SentenceSegmenter` matches the `.reference` class token and `already_claimed?` in this same PR uses `.split.include?`. A `<sup class="references-inline">` would be stripped from the index but kept in the segmenter's sentence, so that claim would match nothing and never highlight. Likely unreachable in real parser output; a consistency nit.
- **`← {I18n.t('claim_verification.back_to_claim')}`** hardcodes an LTR arrow outside the translation.

---

*Drafted in a Claude Code session (~25 minutes, 2 interactions). Claude Code ran the review workflow end to end and performed every verification named above; Sage read the resulting report and approved posting it, without independently re-checking any finding. The PR itself was written in the same tool, so this is a self-review rather than a second pair of eyes, and should be weighted accordingly. The two suggested locale lines reuse wording already present in the existing `nl` and `sk` translations rather than newly generated copy.*

PR #6982, opened 2026-08-06
Fix automatic backup when long updates running

gabina:
## What this PR does
This PR changes the approach we use to prevent the backup system from waiting indefinitely for a huge course update to finish.

Note that, while a backup row exists in the `waiting` or `running` state, all course update processes go to sleep until the backup finishes. At the same time, however, the backup doesn't start until all current course updates are sleeping. This means that, if the backup script runs while there is a long course update, all other course updates will sleep until the huge one finishes. That's undesirable.

The previous approach was to update the `current_backup` definition to ignore rows that had been updated more than 2 hours ago. This meant that a course update would sleep for at most 2 hours and, after that, regardless of whether the backup had run, processing would return to normal. That strategy had two main problems:
**(1)** the backup process running on the database server never finished because it waited forever. In fact, there were 4 backup processes running in the db server (all waiting for `/system/can_start_backup.json` endpoint returning OK). All of them finished successfully when I manually ran the new backup after failing the last row because the  `/system/can_start_backup.json` endpoint  finally returned OK.
**(2)** a backup row remained in the `waiting` state indefinitely, meaning that the next week's backup wouldn't run either unless we manually updated the row.

The new approach lives in the backup script rather than on the Ruby side. It waits until the system is ready to start a backup for up to 2 hours (sleeping up to 24 times for 300 seconds each). If 2 hours pass without the system becoming ready, the script marks the row as `failed` and exits. Since there are no longer any backup rows in the `waiting` or `running` state, `current_backup` returns nothing and all course updates wake up.

Note that removing `FRESH_WINDOW` would have a disadvantage: if the script terminated unexpectedly, for example because of an OOM condition or because the database server was restarted midway through the backup, a backup row would remain in the `waiting` or `running` state and the course update processes would continue sleeping. Because of that, we decided to keep `FRESH_WINDOW` and extend it to 5 hours (2 hours of waiting plus 3 hours to ensure the backup has enough time to run). A backup row that has remained in the `waiting` or `running` state for more than 5 hours is considered orphaned and ignored by the ruby side.

We also updated the `config/backup/waiting.sql` query so that, before creating a new `waiting` row when the backup script runs, it marks any `waiting` or `running` rows older than 6 days as `failed`. These are orphaned rows left over from the previous week's backup that never ran. Since we don't want those orphaned rows to prevent a new backup from starting, we clean them up before creating the new `waiting` row.

## AI usage
I used Claude Code to discuss some minor things, but I did the PR and the description myself (ChatGPT corrected some grammar and verb tenses).

### DIFF
--- a/app/models/system/backup.rb
+++ b/app/models/system/backup.rb
@@ -14,10 +14,10 @@
 #
 class Backup < ApplicationRecord
   IN_PROCESS = %w[waiting running].freeze
-  # A healthy backup run touches its row within minutes and waits at most 2 hours
-  # to start. Anything older is an orphaned row left behind by a crashed backup.sh
-  # (e.g., host restart), and must not indefinitely block CourseDataUpdateWorker via
-  # LogSidekiqStatus#pause_until_no_backup.
+  # A backup run waits at most 2 hours to start and then takes well under an hour.
+  # A row created longer ago than that is an orphan left behind by a crashed
+  # backup.sh (e.g., host restart), and must not indefinitely block
+  # CourseDataUpdateWorker via LogSidekiqStatus#pause_until_no_backup.
   FRESH_WINDOW = 5.hours
 
   def self.current_backup
@@ -25,7 +25,7 @@ class Backup < ApplicationRecord
     # than necessary since it doesn't detect the backup finished.
     ActiveRecord::Base.uncached do
       Backup.where(status: IN_PROCESS)
-            .where('updated_at >= ?', FRESH_WINDOW.ago)
+            .where('created_at >= ?', FRESH_WINDOW.ago)
             .order(id: :desc)
             .first
     end
--- a/app/models/system/backup.rb
+++ b/app/models/system/backup.rb
@@ -14,11 +14,11 @@
 #
 class Backup < ApplicationRecord
   IN_PROCESS = %w[waiting running].freeze
-  # A healthy backup run touches its row within minutes. Anything older is an
-  # orphaned row left behind by a crashed backup.sh (e.g., host restart), and
-  # must not indefinitely block CourseDataUpdateWorker via
+  # A healthy backup run touches its row within minutes and waits at most 2 hours
+  # to start. Anything older is an orphaned row left behind by a crashed backup.sh
+  # (e.g., host restart), and must not indefinitely block CourseDataUpdateWorker via
   # LogSidekiqStatus#pause_until_no_backup.
-  FRESH_WINDOW = 2.hours
+  FRESH_WINDOW = 5.hours
 
   def self.current_backup
     # Force uncaching because otherwise pause_until_no_backup may sleep more
--- a/config/backup/backup.sh
+++ b/config/backup/backup.sh
@@ -7,6 +7,11 @@ QUERY_ROUTE="ROUTE_TO_QUERIES"
 DASHBOARD_URL="URL_TO_DASHBOARD"
 JSON_ENDPOINT="$DASHBOARD_URL/system/can_start_backup.json"
 
+# Give up after MAX_WAIT_ATTEMPTS * WAIT_INTERVAL_IN_SECONDS (two hours) instead of
+# polling forever, so a course update that never goes to sleep can't hang the script.
+WAIT_INTERVAL_IN_SECONDS=300
+MAX_WAIT_ATTEMPTS=24
+
 log() {
   printf '%s : %s\n' "$(date '+%m%d%Y %T')" "$1" >> "$LOG_FILE"
 }
@@ -28,11 +33,19 @@ check_app_is_ready_for_backup() {
 }
 
 wait_until_ready_for_backup() {
+  local attempts=0
   local status=$(check_app_is_ready_for_backup)
 
   while [ $status != 200 ];
   do
-      sleep 300
+      if [ $attempts -ge $MAX_WAIT_ATTEMPTS ]; then
+        log "Waiting for more than two hours. Aborting"
+        mysql < "$QUERY_ROUTE/failed.sql" || true
+        exit 1
+      fi
+
+      attempts=$((attempts + 1))
+      sleep $WAIT_INTERVAL_IN_SECONDS
       status=$(check_app_is_ready_for_backup)
   done
 }
--- a/config/backup/backup.sh
+++ b/config/backup/backup.sh
@@ -7,10 +7,13 @@ QUERY_ROUTE="ROUTE_TO_QUERIES"
 DASHBOARD_URL="URL_TO_DASHBOARD"
 JSON_ENDPOINT="$DASHBOARD_URL/system/can_start_backup.json"
 
-# Give up after MAX_WAIT_ATTEMPTS * WAIT_INTERVAL_IN_SECONDS (two hours) instead of
-# polling forever, so a course update that never goes to sleep can't hang the script.
+# Give up once two hours have gone by, instead of polling forever, so a course update
+# that never goes to sleep can't hang the script. The settle sleep counts towards that
+# budget, so the script never spends more than MAX_WAIT_IN_SECONDS before the dump.
+MAX_WAIT_IN_SECONDS=7200
+SETTLE_SLEEP_IN_SECONDS=120
 WAIT_INTERVAL_IN_SECONDS=300
-MAX_WAIT_ATTEMPTS=24
+MAX_WAIT_ATTEMPTS=$(( (MAX_WAIT_IN_SECONDS - SETTLE_SLEEP_IN_SECONDS) / WAIT_INTERVAL_IN_SECONDS ))
 
 log() {
   printf '%s : %s\n' "$(date '+%m%d%Y %T')" "$1" >> "$LOG_FILE"
@@ -39,7 +42,7 @@ wait_until_ready_for_backup() {
   while [ $status != 200 ];
   do
       if [ $attempts -ge $MAX_WAIT_ATTEMPTS ]; then
-        log "Waiting for more than two hours. Aborting"
+        log "Waited two hours for the app to be ready. Aborting"
         mysql < "$QUERY_ROUTE/failed.sql" || true
         exit 1
       fi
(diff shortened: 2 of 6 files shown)


gabina:
@ragesoss this PR is ready for review.

ragesoss (commented):
Reviewed with Claude Code. As always with AI-assisted review: this is meant to be read and considered, not trusted automatically — use or ignore any of it based on your own judgment.

The approach makes sense, and the write-up of the production forensics made it easy to follow. Moving the give-up logic into the script addresses both failure modes directly. Widening `failed.sql` to cover `waiting` is worth calling out as an improvement in its own right: a crash between `waiting.sql` and `running.sql` previously left a permanent `waiting` orphan, and now it doesn't.

**Verified locally** (branch checked out, master merged): `backup_spec`, `backups_controller_spec` and `log_sidekiq_status_spec` pass (16 examples). RuboCop clean, `bash -n` clean. The new wait loop was exercised against a stubbed endpoint — it aborts after exactly `MAX_WAIT_ATTEMPTS` sleeps (24 × 300s = 2h), runs `failed.sql`, and exits 1, and the `ERR` trap does not re-fire on the explicit `exit 1`, so `failed.sql` isn't run twice. `waiting.sql` was run against a scratch MariaDB 11.8 with seeded rows: 7-day-old `waiting` and `running` orphans are failed and a new row is inserted, and an empty table inserts cleanly. The new `UPDATE` is covered by the existing `GRANT INSERT, SELECT, UPDATE ON dashboard.backups` for `dbbackup`, so no privilege change is needed on either server.

---

### 1. The two "orphan" thresholds don't line up

`FRESH_WINDOW` (5h) is the age at which the app stops seeing a row; `waiting.sql` (6 days) is the age at which the SQL clears it. A row whose age falls between the two is in the worst of both states — invisible to `current_backup`, so `can_start_backup` can never return 200, but still present, so `NOT EXISTS` blocks the insert of a new row.

Seeding a 30-hour-old `waiting` row and running the new `waiting.sql` against a local MariaDB:

```
| id | status  | age_hours |
|  1 | waiting |      30.0 |      <- not failed, and no new waiting row inserted
```

and the same row makes `/system/can_start_backup.json` return 503 indefinitely.

This does self-heal: the script polls for two hours, then the widened `failed.sql` clears the orphan on the way out. The cost is one skipped weekly backup. With a weekly cron the gap only opens on a manual re-run after a crash — which is more or less the situation that prompted this PR, so it may be worth closing. Matching the SQL cutoff to `FRESH_WINDOW` rather than 6 days would do it, on the reasoning that a row the app already ignores can't be a live backup. The 6-day value is the more conservative choice, so this is a question rather than a request.

Related, and lower stakes: `failed.sql` has no age bound, so if a manual run ever overlaps the cron, an abort in one script would fail the other's live row.

### 2. Worst-case course-update stall goes from 2h to 5h

Only in the crash-without-cleanup case — in the normal abort path the script fails the row at ~2h02m and the workers wake within a poll interval, which matches the old behaviour closely. The PR description names this trade-off, so this is just confirming it reads as intentional.

### 3. The freshness clock starts at row creation, not last touch

`running.sql` and `finished.sql` set `status`/`start`/`end` but never `updated_at`, so `current_backup`'s `updated_at >= FRESH_WINDOW.ago` filter measures from when `waiting.sql` inserted the row. That is what makes the "2h wait + 3h dump" arithmetic in the description correct, so the 5h value looks right — but it does mean the dump budget shrinks by however long the wait took, rather than resetting when the dump starts. Worth checking how long the production dump actually takes: if it can exceed 3h after a full 2h wait, workers would resume mid-dump.

It also makes the updated comment in `backup.rb` slightly misleading — "a healthy backup run touches its row within minutes" describes something the scripts don't do; the row's `updated_at` is written once and never again.

### 4. Minor

- `SERVER_CONFIG.md` still says the script "waits for 2 minutes and retries" between polls; the script sleeps 300s. Pre-existing, but it's in the paragraph this PR edits.
- The specs hardcode `6.hours.ago` against a 5-hour constant. `(described_class::FRESH_WINDOW + 1.hour).ago` would stop them needing an edit each time the window moves.

---

*Drafted in a Claude Code session (~20 min, 3 interactions). Sage directed the review and read the findings before posting, but did not independently re-run the verification described above — the spec runs, the shell-loop simulation and the local SQL scenarios were all executed by Claude Code, and the reasoning in each finding was reviewed rather than separately confirmed.*

gabina:
**1. The two "orphan" thresholds don't line up**

> FRESH_WINDOW (5h) is the age at which the app stops seeing a row; waiting.sql (6 days) is the age at which the SQL clears it.

That's true. But the only reason I updated the `waiting.sql` query to delete orphan rows that were created more than 6 days ago is to avoid the weekly automatic backup failing just because the backup from the previous week failed.

I want to keep the system simple. 

> A row whose age falls between the two is in the worst of both states — invisible to `current_backup`, so `can_start_backup` can never return 200, but still present, so NOT EXISTS blocks the insert of a new row.

That's true, but the backup process now has a `waiting` limit. It waits for up to two hours. A backup that remains in the `waiting` state for more than two hours is an unexpected situation (but possible if the backup process dies abruptly). In that case, however, I think it's OK to ignore the row. The only expected situation in which another backup process would start running in the same week while that waiting row still exists is if someone runs a manual backup. In that case, I think it's reasonable to require manually cleaning up the waiting backup row.

**2. Worst-case course-update stall goes from 2h to 5h**
That's true but it's intentional. Even stopping course updates for 5 hours we should be to catch up quickly.

**3. The freshness clock starts at row creation, not last touch**
I updated the `current_backup` definition to look at `created_at` field instead of `updated_at` in commit 763803d9bf0f4eda75f129398d77aa96f835d44f

I also updated SERVER_CONFIG.md since it had some minor errors and tested several scenarios locally.

ragesoss (approved):
Approving. The three follow-up commits resolve the substantive point from the last round, and the two findings that weren't taken were answered with reasoning that holds up under checking.

As before with AI-assisted review: this is meant to be read and considered, not trusted automatically — use or ignore any of it based on your own judgment.

**Addressed:**

- Switching `current_backup` to `created_at` (`763803d9`) fixes the misleading comment and, more usefully, puts the Ruby side and `waiting.sql`'s orphan `UPDATE` on the same clock — previously one measured `updated_at` and the other `created_at`. Confirmed that nothing in the codebase writes `updated_at` on a backup row (only the four SQL files touch them, and none set it), so the change is behaviour-neutral in production and clearer. The new `ignores a stale backup even if it was updated recently` example pins exactly the right thing.
- Deriving `MAX_WAIT_ATTEMPTS` from a stated budget (`2734a538`) reads better than a bare `24`, and the arithmetic checks out: `(7200−120)/300` = 23, truncating in the safe direction. Simulated against the real loop body with `sleep`/`curl`/`mysql` stubbed — 23 sleeps plus the 120s settle = 7020s (1h57m), inside the 7200s budget, and the `ERR` trap still doesn't re-fire on the explicit `exit 1`. That leaves 3h03m of dump budget inside the 5h `FRESH_WINDOW`.
- `SERVER_CONFIG.md` now says 5 minutes, and the `expected`→`unexpected` and `updated`→`created` fixes are correct.

**On the two declined findings** — both look right on closer inspection:

- The threshold gap fails *safe*, which wasn't spelled out last time. `can_start_backup` returns 503 when `current_backup` is nil, so a gap-aged row produces: no new row inserted, 503 indefinitely, a 2h poll, then `failed.sql` clears the orphan. The cost is a skipped backup, never an unpaused dump. Verified via a request spec exercising the composed path.
- The 6-day cutoff also has more margin than it first appears against a weekly cron: an orphan from week *N* is ~7 days old at week *N+1*'s run, since the wait shifts `created_at` by at most 1h57m. The cron path always self-heals; only a mid-week manual run lands in the gap, which matches the stated expectation of manual cleanup there.

**Possible follow-up, not for this PR:** a skipped backup is now entirely silent. `failed` becomes a reachable by-design state here, and it's also the only signal that a weekly backup didn't happen — but nothing surfaces it. No admin view (`routes.rb` exposes only `can_start_backup.json`), no mailer, no Sentry call, no worker checking status; `cap <stage> backup:download` lists dump files and requires going looking. Previously a stuck row announced itself through stalled course updates, which is how this bug surfaced at all. Removing that noise is the point of the PR, and it also removes the accidental alarm. Some cheap signal on a `failed` row — or on "no dump file newer than N days" — might be worth an issue.

**Minor:**

- The PR description still says "sleeping up to 24 times"; the derived constant makes it 23.
- The `Intended flow` section doesn't mention `waiting.sql`'s 6-day orphan cleanup, which is the least obvious part of the design and the thing a future maintainer would want when they hit the gap case.
- `running.sql` and `finished.sql` are unbounded in the same way `failed.sql` is, so a gap-aged orphan coexisting with a fresh row would be swept to `running` and then `finished` alongside it — recording success for a run that crashed. History-only, and only reachable via manual intervention.
- The specs still hardcode `6.hours.ago` against a 5-hour constant in four places; `(described_class::FRESH_WINDOW + 1.hour).ago` would stop them needing an edit whenever the window moves.

**Verified locally** (branch checked out, master merged): 17 existing examples pass, RuboCop and `bash -n` clean. Two `courses_controller_spec` failures (biographies email, CPM staff role) reproduce on master and are unrelated. An 8-example exploratory request spec covering the gap scenario end to end — including `LogSidekiqStatus` against the real `created_at` query, which the existing spec stubs out entirely — also passes; it was written by Claude Code and can be shared if useful for the `LogSidekiqStatus` coverage gap in particular.

---

*Drafted in a Claude Code session (~25 min, 2 interactions). Sage directed the re-review, read the findings, and made the call to approve and treat the alerting gap as a possible follow-up rather than a change request, but did not independently re-run the verification described above — the spec runs, the master comparison, the shell-loop simulation and the exploratory spec were all executed or written by Claude Code, and the cron-margin argument is arithmetic rather than an executed test. The exploratory spec was not reviewed line by line.*

gabina:
When do we plan to deploy this? Asking because I should update the backup scripts in the servers at the same time.

ragesoss:
I'll deploy tomorrow.

gabina:
Already updated the scripts and queries in both servers :heavy_check_mark:

Issue #6990, opened 2026-08-07
peony-web wedges from memory exhaustion every 1-2 days, requiring manual reboot


ragesoss:
## Summary

peony-web has been freezing every day or two, requiring an admin to notice and reboot it. The cause is not a recent code change: the host is configured to run more than its 7.8 GB of RAM can hold, and it has **no swap**. When it saturates it does not fail cleanly — it enters page-cache thrash and becomes unresponsive while its processes stay alive, so nothing crashes, nothing restarts, and no alert fires beyond the uptime monitor.

Only peony-web is affected. peony-sidekiq, peony-sidekiq-medium, peony-sidekiq-short, peony-database and p-and-e-dashboard-redis were all healthy throughout the most recent incident.

### Incident timeline, 2026-08-07 (all times UTC)

| Time | Event |
|---|---|
| 04:50 | Last course update processed; scheduling stops |
| 04:56 | Sentry uptime monitor (PEONY-225) records `timeout - Request timed out` — a timeout, not a connection refusal |
| 05:48 / 06:01 | peony-sidekiq-medium and peony-sidekiq-short finish in-flight work, then go idle with empty queues |
| 06:22 – 13:02 | peony-sidekiq continues `ConstantUpdateWorker` on its normal ~80 min cadence, entirely unaffected |
| 11:41 | journald stops being able to write on peony-web |
| 13:53 | peony-web still opening TCP connections to peony-database — kernel alive, userspace frozen |
| 14:21 | Admin reboots the box; sidekiq-cron then drains ~9.5 h of missed jobs back-to-back |
| 16:46 | OOM killer kills a Passenger worker: `ruby ... anon-rss:1638236kB` (1.6 GB) under `apache2.service` |

### Boot history — four wedges in four days

```
-4   Aug 03 00:25:32 → Aug 04 14:20:13
-3   Aug 04 14:34:41 → Aug 06 14:23:12
-2   Aug 06 14:34:01 → Aug 07 01:40:17
-1   Aug 07 01:53:40 → Aug 07 11:41:28
 0   Aug 07 14:21:21 → (current)
```

Each boundary is an admin noticing downtime and rebooting. The clustering around 14:20–14:34 UTC is ~07:20 Pacific — an artifact of when someone gets to their desk, not a scheduled job.

### Why it wedges instead of crashing

[...]

PR #6991, opened 2026-08-07
GSOC26: Optimize SystemCsvBuilder memory by streaming rows directly to CSV

lakshita10341:
## What this PR does
Optimizes memory consumption in `SystemCsvBuilder` by streaming batch rows directly into `CSV.generate` instead of accumulating them into an intermediate `csv_data` array first. This cuts peak memory usage during system-wide CSV generation roughly in half, preventing unnecessary memory bloat on memory-constrained servers. 

Addresses feedback discussed in issue #6990 regarding server stability and worker memory pressure.

## AI usage
Used AI to analyze the memory profile of `SystemCsvBuilder#generate_csv` and refactor the builder method to stream rows efficiently,.

### DIFF
--- a/lib/analytics/system_csv_builder.rb
+++ b/lib/analytics/system_csv_builder.rb
@@ -6,8 +6,9 @@ require_dependency "#{Rails.root}/lib/analytics/system_csv_filter_validator"
 
 # Generates system-wide CSV exports with dynamic filter support.
 # This is a standalone builder designed for admin-only, async exports
-# across all non-private programs. It preloads associated data in memory-safe
-# batches (1,000 courses at a time) to keep RAM consumption low.
+# across all non-private programs. It streams rows directly into the CSV
+# output and fetches associated data in batches (500 courses at a time)
+# to keep RAM consumption low.
 #
 # Filters supported:
 #   campaign_slug  — Courses belonging to a specific campaign
@@ -30,13 +31,12 @@ class SystemCsvBuilder
   end
 
   def generate_csv
-    csv_data = [CourseCsvBuilder::CSV_HEADERS]
-
-    course_scope.find_in_batches(batch_size: BATCH_SIZE) do |batch|
-      csv_data.concat(build_batch_rows(batch))
+    CSV.generate do |csv|
+      csv << CourseCsvBuilder::CSV_HEADERS
+      course_scope.find_in_batches(batch_size: BATCH_SIZE) do |batch|
+        build_batch_rows(batch).each { |row| csv << row }
+      end
     end
-
-    CSV.generate { |csv| csv_data.each { |line| csv << line } }
   end
 
   # Returns the filtered course scope. Public so it can be tested directly.


ragesoss (approved):
Reviewed with Claude Code. AI-generated review feedback of this kind is worth
reading and understanding, but not worth trusting automatically — use or ignore
it based on your own judgment of the code.

## Verification

The refactor is output-preserving. An exploratory spec (below) reconstructs the
pre-refactor accumulate-then-generate form against the same private helpers and
asserts byte equality with the new streamed output while forcing several
batches — the header is emitted exactly once, in the first row, with no rows
dropped at batch boundaries and filters still applied per batch.

RuboCop is clean on the changed file, the 19 existing `SystemCsvBuilder`
examples pass, and CI is green.

## The memory improvement is larger than the description claims

Peak RSS measured via `/proc/self/status` `VmHWM`, each variant in its own
process, using rows shaped like the real 24-column `CSV_HEADERS` (mixed
String/Integer/Time) batched at 500. Figures are net of a 12.3 MB bare
interpreter baseline:

| courses | output size | peak, before | peak, after | reduction |
|---:|---:|---:|---:|---:|
| 10,000 | 3.1 MB | 16.8 MB | 8.2 MB | 2.0× |
| 50,000 | 15.9 MB | 82.8 MB | 21.7 MB | 3.8× |
| 100,000 | 32.0 MB | 165.6 MB | 38.0 MB | 4.4× |

The old form's peak held steady at roughly **5.2× the size of the finished
export**, rather than the 2× estimated in #6990. The extra factor is object
overhead rather than text: an Array of 24-element Arrays holding `Time` and
`String` objects costs substantially more than the flat CSV it serializes to.
The streamed form lands at about 1.2× output at scale, because each row becomes
garbage as soon as it is written.

So "roughly in half" is the floor. At export sizes large enough to matter on a
7.8 GB host, the saving is closer to a factor of four.

This is a synthetic model of the row shapes rather than the real builder against
a populated database, so treat the absolute numbers as indicative. The
per-batch ActiveRecord objects are identical in both versions, so the delta
between them should hold.

## One wording concern

The revised class comment reads:

> It streams rows directly into the CSV output and fetches associated data in
> batches (500 courses at a time) to keep RAM consumption low.

"streams rows directly into the CSV output" can be read as streaming to a file.
It isn't — `CSV.generate` still materializes the complete CSV as a single
String, which `ReportCsvWorker#write_csv` then writes in one call. Peak memory
is still proportional to the size of the whole export, just with a much smaller
constant.

That distinction matters because an inaccurate memory claim in this same comment
is what the original review flagged. Stating explicitly that the full CSV is
still returned as a String would keep the file honest about the ceiling that
remains, and would leave the follow-up (writing straight to the destination
path) clearly signposted rather than looking already done.

The `1,000` → `500` correction is right.

## Suggested spec

Every pre-existing `#generate_csv` example creates about five courses against a
`BATCH_SIZE` of 500, so the multi-batch path — the one code path whose shape
this PR actually changes — has no coverage in the suite. The following spec was
generated by Claude Code during review and passes against this branch; adopting
it (or something like it) would pin the behavior this refactor depends on.

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe SystemCsvBuilder do

  let(:en_wiki) { Wiki.get_or_create(language: 'en', project: 'wikipedia') }

  # Enough courses to span several batches once BATCH_SIZE is stubbed down.
  let!(:courses) do
    (1..7).map do |i|
      create(:course, slug: "school/course_#{i}_(term)",
                      title: "Course #{i}",
                      start: 2.months.ago,
                      end: 2.months.from_now,
                      type: 'ClassroomProgramCourse',
                      home_wiki: en_wiki)
    end
  end

  describe '#generate_csv across multiple batches' do
    before { stub_const('SystemCsvBuilder::BATCH_SIZE', 2) }

    it 'emits the header exactly once' do
      rows = CSV.parse(described_class.new(filters: {}).generate_csv)
      header_rows = rows.select { |row| row == CourseCsvBuilder::CSV_HEADERS }
      expect(header_rows.count).to eq(1)
    end

    it 'places the header in the first row' do
      rows = CSV.parse(described_class.new(filters: {}).generate_csv)
      expect(rows.first).to eq(CourseCsvBuilder::CSV_HEADERS)
    end

    it 'emits one row per course with no rows dropped at batch boundaries' do
      rows = CSV.parse(described_class.new(filters: {}).generate_csv)
      slugs = rows.drop(1).map(&:first)
      expect(slugs).to match_array(courses.map(&:slug))
    end
  end

  describe '#generate_csv with a filter that spans batches' do
    before { stub_const('SystemCsvBuilder::BATCH_SIZE', 3) }

    it 'applies the filter across every batch' do
      create(:course, slug: 'school/editathon_(term)',
                      title: 'Editathon Course',
                      start: 2.months.ago,
                      end: 2.months.from_now,
                      type: 'Editathon',
                      home_wiki: en_wiki)

      rows = CSV.parse(described_class.new(filters: { course_type: 'Editathon' })
                                      .generate_csv)
      expect(rows.drop(1).map(&:first)).to eq(['school/editathon_(term)'])
    end
  end
end
```

Neither the comment wording nor the spec is a blocker. The change does what it
says, it is the right intermediate step, and the full file-streaming version
remains a sensible follow-up.

*Drafted in a Claude Code session (~20 min, 3 interactions). Sage directed the
review and read the full comment before posting, but did not independently
verify the findings. Claude Code ran the existing suite, RuboCop, and the
generated spec locally on the PR branch merged with master, and ran the memory
benchmark; the equivalence spec and the benchmark script are AI-written and were
not line-by-line reviewed by a human.*

PR #6992, opened 2026-08-07
Expand WikiWho authorship highlighting to all 70 supported languages

ragesoss:
## What this PR does

The WhoColor authorship-highlighting feature gated itself on a hardcoded list of 14 language codes, commented as accurate "as of 2023-05-15." The [WikiWho service](https://wikiwho-api.wmcloud.org/) has added many language editions since then, so the Dashboard was silently withholding authorship highlighting on wikis it now covers — among them ru, sv, uk, zh, fa, he, ko, vi, cs, fi, da, no, ro, and simple.

This replaces that list with the current set of 70 supported languages, read off the running service rather than the upstream source file (the old comment's GitHub line-number anchor had drifted). The list and the Wikipedia-only condition move into a small `wikiwhoLanguages.js` module so they're greppable and unit-testable, and the module's header comment records the command that regenerates the list from the live service plus the date it was last verified.

The gate itself is kept rather than dropped. Unsupported language codes return 404 from the WhoColor endpoint, so letting every request through would surface the authorship button and a failing legend on the ~270 Wikipedias that have no authorship data.

Relevant documentation: [WikiWho / WhoColor API](https://wikiwho-api.wmcloud.org/gesis_home), [source](https://github.com/wikimedia/wikiwho_api).

## AI usage

Claude Code wrote all of the code, the tests, and the commit message in this PR, working from short human prompts. The human contribution was direction and one substantive correction: Claude's first move was to read the supported-language list out of the upstream `settings_wmcloud.py` on GitHub, and the human interrupted and redirected it to the actual production server at `wikiwho-api.wmcloud.org`. That correction mattered — it produced a list read from what is actually deployed rather than from source that may not match. Claude then verified the codes against the live WhoColor endpoints directly (ru, sv, simple, vec returned 200; nn and xx returned 404), which confirmed both that the list is accurate and that unsupported languages hard-fail, which is what justifies keeping the gate at all.

This PR description, including the "What this PR does" summary above, was also drafted by Claude Code and may contain errors.

Scale of effort: one commit, produced in a single focused session of roughly a dozen turns. The human sent three short messages totalling about two sentences of text. There were no false starts in the implementation itself; the only correction was the redirect described above. Tests were green on the first run.

## Screenshots

No new UI. The change widens an existing gate, so the visible effect is that the "Current version with authorship highlighting" button and its contributor legend now appear on article viewers for 56 additional Wikipedia language editions instead of being hidden.

Before/after screenshots were not captured because reproducing the difference requires live WikiWho responses for a non-English wiki, which a local feature spec can't produce. To see it manually, open the article viewer for an article on any newly-covered wiki (for example a ru.wikipedia or sv.wikipedia article in a P&E Dashboard course) and check that the authorship-highlighting button is offered.

- The list is still a hardcoded snapshot and will go stale again as WikiWho adds languages. Fetching it from the service at runtime was considered and rejected as adding a network round-trip and a new failure mode to every article-viewer open. The refresh command in the module's header comment is the mitigation; a periodic check would be a reasonable follow-up if the drift proves annoying.
- `course_ores_plot.jsx` has a structurally similar `ORESSupportedWiki.languages` gate. That's the ORES/quality service with its own separate list, so it was deliberately left alone here — but it may be worth auditing for the same kind of staleness.
- `be-tarask` is intentionally absent: WikiWho covers `be` but not `be-tarask`, and they are separate wikis.

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/wikiwhoLanguages.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/wikiwhoLanguages.js
@@ -1,0 +1,34 @@
+/*
+  Wikipedia language editions covered by the WikiWho / WhoColor API at
+  https://wikiwho-api.wmcloud.org.
+
+  The service adds languages over time, so this list goes stale. To refresh it,
+  read the language codes out of the per-language endpoint URLs listed on the
+  service's index page:
+
+    curl -s https://wikiwho-api.wmcloud.org/ \
+      | grep -oE '/[a-z-]+/whocolor/v1\.0\.0-beta' \
+      | sed 's|/whocolor.*||; s|^/||' | sort -u
+
+  Codes not on the list return 404 from the WhoColor endpoint, which is why we
+  gate on it rather than letting every request through.
+
+  Last verified against the live service: 2026-08-07 (70 languages).
+*/
+export const WIKIWHO_LANGUAGES = [
+  'af', 'als', 'ar', 'az', 'be', 'bg', 'bn', 'bs', 'ce', 'cs',
+  'cy', 'da', 'de', 'dsb', 'el', 'en', 'eo', 'es', 'et', 'eu',
+  'fa', 'fi', 'fr', 'gl', 'he', 'hi', 'hr', 'hu', 'ia', 'id',
+  'it', 'ja', 'ka', 'kk', 'ko', 'ku', 'lt', 'lv', 'mk', 'ml',
+  'mr', 'ms', 'mt', 'ne', 'nl', 'no', 'pl', 'pt', 'ro', 'ru',
+  'sh', 'simple', 'sk', 'sl', 'sq', 'sr', 'sv', 'sw', 'ta', 'te',
+  'tg', 'th', 'tl', 'tr', 'uk', 'ur', 'uz', 'vec', 'vi', 'zh',
+];
+
+// WhoColor only covers Wikipedia. Other projects (Wiktionary, Wikisource, etc.)
+// have no authorship data even when their language code appears above.
+export const isWikiwhoSupported = article => (
+  article.project === 'wikipedia' && WIKIWHO_LANGUAGES.includes(article.language)
+);
+
+export default WIKIWHO_LANGUAGES;
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/wikiwhoLanguages.spec.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/wikiwhoLanguages.spec.js
@@ -1,0 +1,36 @@
+import { WIKIWHO_LANGUAGES, isWikiwhoSupported } from './wikiwhoLanguages';
+
+describe('WIKIWHO_LANGUAGES', () => {
+  it('contains no duplicate language codes', () => {
+    expect(new Set(WIKIWHO_LANGUAGES).size).toEqual(WIKIWHO_LANGUAGES.length);
+  });
+
+  it('includes the non-standard "simple" code used by Simple English Wikipedia', () => {
+    expect(WIKIWHO_LANGUAGES).toContain('simple');
+  });
+});
+
+describe('isWikiwhoSupported', () => {
+  it('is true for a supported language on Wikipedia', () => {
+    expect(isWikiwhoSupported({ language: 'en', project: 'wikipedia' })).toBe(true);
+  });
+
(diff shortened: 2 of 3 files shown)


ragesoss:
Review notes from a Claude Code session. These are provisional observations from AI analysis, not authoritative conclusions — they are offered to be read and weighed, not automatically trusted or acted on. Anything below may be wrong; the commands used are included so the claims can be rerun rather than taken on faith.

## Verified against the live service

The central factual claim of the PR — that the committed array matches what the service currently serves — was checked independently by running the refresh command from the module's own header comment and diffing the output against the committed list. **Zero differences, 70/70 codes.** The documented command works exactly as written and reproduces the list, which is worth noting on its own, since embedded refresh commands frequently stop working before the data they describe goes stale.

All 70 codes were also cross-checked against `Wiki::LANGUAGES` in `app/models/wiki.rb`. Every one is a valid Dashboard language value, so there are no inert entries that could never match a wiki.

Endpoint behavior was spot-checked beyond the codes named in the PR description, concentrating on the obscure newly-enabled ones:

| code | status | notes |
|---|---|---|
| `dsb`, `tg`, `vec` | 200 | newly enabled, serve data |
| `ce` | 200 | with its real Cyrillic title (`Берлин`) |
| `nn`, `xx` | 404 | unsupported route |

This supports the PR's reasoning for keeping the gate rather than dropping it.

## A refinement to the stated premise

The WhoColor error taxonomy appears to be three-way rather than two-way. The module comment states that codes off the list return 404, which holds, but there is a distinct middle case:

- **404** — language route unsupported
- **400** — language supported, article absent on that wiki: `{"error":"The article (Berlin) ... does not exist in chechen Wikipedia.","success":false}`

That 400 lands in what looks like a pre-existing rough edge rather than anything this PR introduces. `AuthorshipAPI.js:23` rejects with the raw `Response` object, so the `.catch` in `useAuthorshipHighlighting.js:139` reads `error.message` as `undefined`; both regex guards then test the string `"undefined"` and fail, and control falls through to `requestTitleVerification(true)` with `failureMessage` left unset. It is only reachable on title-mismatch cases, so widening the gate should not make it common — but it becomes reachable on 56 additional wikis. This seems like follow-up material rather than a merge blocker.

## Minor code note

`export default WIKIWHO_LANGUAGES;` has no consumers — only the named imports are used anywhere in the tree. Exporting the raw array as the module default creates an opening for a future caller to import the list and re-implement the gate while dropping the `project === 'wikipedia'` half of it. Removing the default export would leave `isWikiwhoSupported` as the module's only sanctioned entry point. (Separately, `isWhocolorLang` could be a value rather than a function, since `article` is stable within a render — but the function form is what keeps the diff to a single line, so this is purely optional.)

## Correction to the PR description

Open question #2 appears to rest on a misreading. `ORESSupportedWiki` is currently `{projects: [], languages: []}` — emptied deliberately in commit `882574e2` ("Disable ores plot feature for every course language/project") pending re-implementation per issue #6327. The nine-language list above it is commented out. The ORES gate is therefore switched off, not stale, and auditing it for list drift would misdirect the follow-up work; the actual blocker there is #6327. That bullet is worth editing before merge.

Also confirmed: there is no server-side counterpart list to keep in sync, unlike the ORES mapping, which carries an explicit "Update the server side mapping in sync" note.

## Checks run locally

- `yarn build` — exit 0
- ESLint on all three changed files — clean
- Jest across `ArticleViewer` — 6 suites, 39 tests, all passing, including the new `wikiwhoLanguages.spec.js`
- RuboCop and the Ruby suite — not applicable, no Ruby files touched
- GitHub CI was still pending at the time of writing

On coverage: `isWikiwhoSupported` is one line with two conditions, and the PR's spec exercises both in both directions, which is proportionate. No exploratory spec was written for the hook integration. The repo's React test infrastructure is enzyme on a React 17 adapter against a React 18 app, the hook is effect-driven, and the change substitutes a semantically identical predicate — so bug risk is low and a failure would be loud rather than silent. The live-service verification above is stronger evidence for this particular change than a local spec could provide.

---

*Drafted in a Claude Code session (~10 minutes, 2 interactions). This is a self-review, and the layering matters for calibration: the PR itself was written by Claude Code in an earlier session, and these notes were drafted by a fresh Claude Code session on the same account. Sage directed the review and approved posting after reading a summary, but did not independently verify any finding — the list diff, live endpoint queries, `Wiki::LANGUAGES` cross-check, build, lint, and spec runs were all performed by Claude Code. The verification claims above are reproducible rather than human-confirmed, which is why the commands are named.*

PR #6993, opened 2026-08-07
Exclude long-term Wikipedians and split retention CSV by prior participation

ragesoss:
## What this PR does

The "Retention predictors" CSV (admin-only, Scholars & Scientists courses) counted
every participant identically. That made two distinct populations invisible in the
aggregates: people who were already established Wikipedians before joining the
cohort, and people who had taken an earlier course and were therefore not having a
first-course experience at all. Both distort what the retention numbers can tell us.

Participants are now classified on two axes and the summary block is split
accordingly:

- **Long-term Wikipedians** — 1000 or more edits before the course start date — are
  still listed in the per-student detail block with all their metrics, but are
  excluded from every summary aggregate. A count row reports how many were excluded.
- **Returning participants** — anyone who took an earlier course as a student — get
  their own summary column, so their second-course numbers are not read as
  first-course outcomes.

Four columns are appended to the per-student detail block: `edits before course`,
`long-term Wikipedian (1000+ edits)`, `prior courses`, and `prior course slugs`. They
go at the end so existing column positions are unchanged for anyone with a
spreadsheet built on the old layout.

Both thresholds and definitions came from the report's operator: the 1000-edit bar,
prior edit count as the sole criterion (no account-age condition), and "any earlier
course where they were a student" as the prior-course rule.

Relevant API documentation: [`list=usercontribs`](https://www.mediawiki.org/wiki/API:Usercontribs).

## AI usage

Claude Code wrote all of the code in this PR, along with the commit message and this
description. The human contribution was direction and decisions: stating the two
problems to solve, choosing among four presented options for how to identify a
long-term Wikipedian and how to present the split, supplying the 1000-edit threshold
directly rather than picking from the offered values, and correcting a proposed
column label. Claude proposed the plan and asked those four questions before writing
anything; the human approved the plan and the implementation proceeded from there.

Note that the "What this PR does" summary above and the analysis throughout this
description were also drafted by Claude Code and may contain errors — review the diff
rather than trusting the summary.

Scale of effort: one commit, written in a single session of roughly fifteen turns
with three user messages totalling about a paragraph of text. This is a small,
focused change; there was no extended iteration or multi-day development. The commit
message's `## Process` section has a fuller accounting.

## Screenshots

No UI changes — the only visible surface is the generated CSV. Below is the same
illustrative fixture (four participants, one of them a long-term Wikipedian, one of
them a repeat participant) rendered by the old and new code. These are synthetic
fixture numbers, not real cohort data.

**Before:**

```
Summary
participants,4
total editing sessions during course,4
participants with no editing sessions during course,1
avg days to first independent edit,9.0
avg editing sessions in 30 days after course,0.8
participants who edited in 30 days after course,3
participants with 1+ edits in days 60-90,3
participants with 5+ edits in days 60-90 (survivors),1

Per-student detail
username,sessions during course,days to first independent edit,sessions in 30 days after course,edits in days 60-90
Newcomer,2,3,1,5
Quiet,0,30,0,0
Repeat,1,2,1,1
Veteran,1,1,1,2
```

**After:**

```
Summary,first course,returning participants
participants,2,1
long-term Wikipedians (excluded from all counts),1
total editing sessions during course,2,1
participants with no editing sessions during course,1,0
avg days to first independent edit,16.5,2.0
avg editing sessions in 30 days after course,0.5,1.0
participants who edited in 30 days after course,1,1
participants with 1+ edits in days 60-90,1,1
participants with 5+ edits in days 60-90 (survivors),1,0

Per-student detail
username,sessions during course,days to first independent edit,sessions in 30 days after course,edits in days 60-90,edits before course,long-term Wikipedian (1000+ edits),prior courses,prior course slugs
Newcomer,2,3,1,5,0,,0,
Quiet,0,30,0,0,0,,0,
Repeat,1,2,1,1,0,,1,Uni/Intro_(2024)
Veteran,1,1,1,2,1000+,yes,0,
```

Note how the "Veteran" row keeps its full metrics in the detail block while
contributing to no aggregate, and how "Repeat" moves out of the first-course column.

**The prior-edit count is scoped to the course's wikis, not global.** It sums
`usercontribs` across `course.wikis`. A participant arriving with 3,000 edits on
de.wikipedia and none on en.wikipedia would read as 0 prior edits in an en-only
cohort. `meta=globaluserinfo` would catch that in a single API call, but it reports
all-time edit counts including edits made *after* the course, which would break the
before-the-course framing and could exclude someone whose post-course editing is
exactly the retention success the report exists to measure. This PR chose accuracy on
the time window over breadth of wikis; if cross-wiki arrivals turn out to be common
in Scholars & Scientists cohorts, that tradeoff is worth revisiting.

**Concurrent enrollments do not count as prior courses.** The comparison is on course
start dates, so a participant enrolled in two overlapping courses is treated as
first-course in both. This seemed right — neither course precedes the other — but it
is a judgment call.

**The 1000-edit bar is measured only at the course start.** Someone who crossed 1000
edits during or after the course is not retroactively excluded, by design.

**Added API cost.** Each participant now costs up to two extra `usercontribs` pages
per course wiki (counting stops the moment it reaches 1000, so a 50,000-edit account
costs no more than a 1,000-edit one). For a 20-person, 2-wiki cohort that is at most
80 extra requests in an already-asynchronous Sidekiq job.

**Class extraction.** `RetentionPredictorsCsvBuilder` was close enough to RuboCop's
130-line class limit that adding a second summary column would have pushed it over,
so the summary block moved into its own `RetentionSummaryBlock` class. That is a
larger diff than the feature strictly required, but the aggregate helpers moved
wholesale rather than being rewritten.

### DIFF
--- a/lib/analytics/retention_participant_history.rb
+++ b/lib/analytics/retention_participant_history.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/wiki_api"
+
+# What a participant had already done before the course being reported on began:
+# how much they had edited, and whether they had already taken an earlier course.
+#
+# RetentionPredictorsCsvBuilder uses both to decide which participants belong in
+# which summary column. Long-term Wikipedians are listed in the report but left
+# out of every aggregate; returning participants are aggregated separately from
+# first-time ones, so a repeat participant's second-course numbers don't get read
+# as a first-course outcome.
+class RetentionParticipantHistory
+  # At or above this many edits before the course started, a participant is
+  # treated as an established Wikipedian rather than a retention subject.
+  LONG_TERM_EDIT_THRESHOLD = 1000
+
+  attr_reader :prior_edit_count, :prior_courses
+
+  def initialize(course, user, wikis)
+    @course = course
+    @user = user
+    @wikis = wikis
+    @prior_courses = find_prior_courses
+    @prior_edit_count = count_prior_edits
+  end
+
+  def long_term_wikipedian?
+    @prior_edit_count >= LONG_TERM_EDIT_THRESHOLD
+  end
+
+  def returning?
+    @prior_courses.any?
+  end
+
+  # Exact below the threshold; "1000+" once counting stopped at it.
+  def prior_edit_count_label
+    long_term_wikipedian? ? "#{LONG_TERM_EDIT_THRESHOLD}+" : @prior_edit_count.to_s
+  end
+
+  # Blank rather than an empty string, so a first-course participant's cell is
+  # genuinely empty in the CSV.
+  def prior_course_slugs
+    @prior_courses.map(&:slug).join('; ').presence
+  end
+
+  private
+
+  # Courses this user took as a student that started before this one did.
+  # Courses that merely overlap this one do not count as prior.
+  def find_prior_courses
+    Course.joins(:courses_users)
+          .where(courses_users: { user_id: @user.id,
+                                  role: CoursesUsers::Roles::STUDENT_ROLE })
+          .where('courses.start < ?', @course.start)
+          .order(:start)
+          .to_a
+  end
+
+  # Edits before the course start across every tracked wiki, summed. Counting
+  # stops at the threshold: an editor with 50,000 prior edits costs no more API
+  # calls than one with exactly 1,000.
+  def count_prior_edits
+    count = 0
+    @wikis.each do |wiki|
+      count += prior_edits_on(wiki, LONG_TERM_EDIT_THRESHOLD - count)
+      break if count >= LONG_TERM_EDIT_THRESHOLD
+    end
+    count
+  end
+
+  # Edits on one wiki before the course start, paginated until `limit` of them
+  # have been counted (or the user's contributions run out).
+  def prior_edits_on(wiki, limit)
+    api = WikiApi.new(wiki)
+    count = 0
+    continue = {}
+    loop do
+      response = api.query(prior_contribs_query.merge(continue))
+      break unless response
+      count += (response.data['usercontribs'] || []).count
+      break if count >= limit
+      continue = response['continue']
+      break unless continue
+    end
+    [count, limit].min
+  end
+
+  # usercontribs defaults to listing backwards from ucstart, so this walks from
+  # the moment before the course began into the past. The one-second offset keeps
+  # an edit made exactly at the course start out of the "before" count, since the
+  # report's during-course window already includes it.
+  def prior_contribs_query
+    {
+      list: 'usercontribs',
+      ucuser: @user.username,
+      ucstart: (@course.start - 1.second).strftime('%Y%m%d%H%M%S'),
+      ucprop: 'timestamp',
+      uclimit: 'max'
+    }
+  end
+end
--- a/lib/analytics/retention_predictors_csv_builder.rb
+++ b/lib/analytics/retention_predictors_csv_builder.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require_dependency "#{Rails.root}/lib/wiki_api"
+require_dependency "#{Rails.root}/lib/analytics/retention_participant_history"
+require_dependency "#{Rails.root}/lib/analytics/retention_summary_block"
 require 'csv'
 
 # Builds the "Retention predictors" CSV report for a course (currently the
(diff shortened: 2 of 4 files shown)


ragesoss:
A Claude Code review session went over this branch. The findings below are provisional
AI analysis, not conclusions to act on automatically — worth reading and understanding,
then using or ignoring per the developer's own judgment. Several were checked against a
locally-run spec; those are marked as verified, and the rest are code-reading only.

## Findings

**1. A failed prior-edit lookup silently demotes a long-term Wikipedian to a zero-edit newcomer.** *(verified locally)*

`retention_participant_history.rb:80` breaks out of the pagination loop on a falsy
response. `WikiApi#query` returns `nil` once its three retries are exhausted
(`wiki_api.rb:137`), so a transient API failure contributes 0 edits for that wiki with no
other signal.

The same nil-swallowing already exists in `edit_times`, but the consequence there is a
metric that reads low. Here it decides group membership. A participant with thousands of
prior edits whose lookup fails reads as `0` in the `edits before course` column —
indistinguishable from a genuine newcomer — is not flagged, and is folded into the
first-course aggregates. That is the contamination this PR exists to remove, and it leaves
no trace in the output.

Reproduced with a spec that stubs the prior-edit query to nil for one of two students with
identical timelines:

```ruby
allow(api).to receive(:query) do |params|
  if params.key?(:ucend)
    response_for(timeline)          # during-course query still succeeds
  elsif params[:ucuser] == 'user1'
    nil                             # WikiApi#query after retries are exhausted
  else
    prior_edits_response(0, params)
  end
end
```

With that stub, `long-term Wikipedians (excluded from all counts)` reports `0` and
`total editing sessions during course` doubles from 1 to 2.

Two directions worth weighing: raise rather than emit a report whose aggregates are
quietly wrong, or write a distinguishable marker into the `edits before course` cell and
drop that row from the aggregates the way an excluded participant is dropped.

**2. The excluded-count row's value lands under the "first course" header.** *(verified locally)*

`retention_summary_block.rb:44` emits `[EXCLUDED_LABEL, @excluded.size]` — a two-cell row
against a three-cell header. The count is a total across both groups, but it occupies the
column headed `first course`. Read column-wise, or summed by a spreadsheet formula over
that column, it reads as a first-course figure. The PR's own spec reads the value at index
1 as well, so the spec encodes the same ambiguity rather than catching it. Splitting the
count into two cells, one per group, would remove the ambiguity and would also answer a
question the current row cannot: whether the excluded participants were first-course or
returning.

**3. The class comment overstates the blank-versus-zero rule.**

`retention_summary_block.rb:17` states that a group with no members gets blanks rather than
a column of zeros. The `participants` row on line 43 bypasses the empty-group guard in
`value` that every metric row goes through, so a cohort with no returning participants
produces `participants,2,0` followed by `total editing sessions during course,2,`. The
behavior looks deliberate — a count of zero participants is a fact, whereas the metrics
over an empty group are undefined — but the comment does not describe it.

**4. Three untested paths.**

- A participant who is both long-term and returning. This is the only place the two
  classification axes interact. The behavior was checked locally and is correct — such a
  participant is excluded from the returning column too, not just the first-course one —
  but nothing in the suite pins it down.
- The API-failure path from finding 1.
- The multi-prior-course `'; '` join in `prior_course_slugs`; only the single-prior-course
  case is covered.

**5. `find_prior_courses` loads full Course records.**

`retention_participant_history.rb:51` uses only `.size` and `.slug` from the result.
`select(:id, :slug, :start)` would be leaner. N is small, so this is preference rather
than a problem.

## Checked and found sound

- `config.time_zone` is unset, so the app runs in UTC and the `strftime` calls in both
  query builders produce correct API timestamps — no offset bug.
- The `- 1.second` offset on `ucstart` pairs correctly with `edit_times`'s inclusive
  `ucend: @course.start`: an edit made exactly at the course start is counted once, in the
  during-course window.
- The cross-wiki budget arithmetic in `count_prior_edits` was traced at the boundaries
  (600+500, 999 spanning two pages, exactly 1000) and is correct in each.
- `find_prior_courses`'s `role: STUDENT_ROLE` matches `Course#students`
  (`courses_users.role = 0`), so both populations are defined consistently.
- `courses.start < ?` excludes the current course without a separate guard.
- Courses are not soft-deleted in this schema, so no `deleted: false` filter is missing.
- Both new classes are well under the 130-line limit.
- RuboCop is clean on all four files; the PR's own spec passes (18 examples) and the whole
  of `spec/lib/analytics` passes (75 examples), so no regression in the surrounding area.

The wiki-scoping tradeoff, the concurrent-enrollment rule, and the added API cost are all
already documented in the PR description, and that analysis holds up on inspection.

*Produced in a Claude Code review session (~15 min, 2 interactions). Sage invoked the
review and requested this comment; the code reading, the local spec runs, and the
exploratory spec that verified findings 1 and 2 were all done by Claude Code. Sage read
the summarized findings before approving this comment but did not independently re-derive
them, and did not review the AI-written exploratory spec, which is not committed to the
branch. Findings 3, 4, and 5 are code-reading only and were not separately verified.*

PR #6994, opened 2026-08-07
Scope the authorship fallback check to the displayed revision

ragesoss:
## What this PR does

When the article viewer's authorship highlighting can't color a particular editor's spans, it falls back to asking "does this editor own any text in this version?" so the legend can distinguish *"contributions present but not highlightable"* from *"none of this version is theirs"*. That check was answered by a second request to WikiWho's `rev_content` endpoint — and that request was never given a revision id. It therefore always described the **current** revision, even when the viewer had been toggled to the student's last revision, so the last-revision view could report the wrong answer.

Rather than thread the revision id through that request, this removes it. `AuthorshipAPI#fetchWhocolorHtml` already fetches a payload containing the same per-token authorship and was discarding everything except `extended_html`; it now returns `tokens` alongside the processed HTML. The fallback answer is scoped to the revision on screen by construction, with one fewer multi-megabyte request per viewer open and no way for the two to drift apart later. The methods that existed only to serve the old path (`fetchWikitextMetaData`, `wikiwhoColorRevisionURL`, and `generateWhocolorHtml`, which had already been dead since 506bb276e) are removed.

This came out of diagnosing a production report that authorship highlighting showed no contributions at all for [Stratovolcano](https://dashboard.wikiedu.org/courses/University_of_New_Mexico/Volcanology_Wikipedia_class_module_(Fall_2024)/articles/edited?showArticle=65780601). **That symptom is not fixed here and is not a Dashboard bug.** The deployed WhoColor at `wikiwho-api.wmcloud.org` misreads the namespace prefix in `[[File:...]]` as a URL scheme, which desynchronises its span annotator for the remainder of the page — an article whose lead opens with an image comes back with exactly one `editor-token` span in 140 KB of HTML. WikiWho's data itself is fine (the student owns 42% of current-revision tokens, 54% at their last revision). That is fixed upstream in [wikimedia/WhoColor#3](https://github.com/wikimedia/WhoColor/pull/3) (`9265d66`) but is not yet deployed; `wikiwho_api` also still pins its `lib/WhoColor` submodule at a pre-fix commit. The bug in this PR was a separate finding along the way.

No issue number is associated with this work.

## AI usage

Claude Code did essentially all of the work in this PR, under human direction.

The session began as a diagnosis request, not a request for a code change. Claude Code investigated by querying the production Dashboard's public JSON endpoints, the MediaWiki API, and the WikiWho whocolor/`rev_content` endpoints, then reproduced WhoColor's parser locally against the real wikitext and tokens across five historical revisions of that library, and rendered the annotated wikitext back through the MediaWiki parse API to confirm the fixed version produces 2,684 spans including the student's. It wrote the diagnosis, identified the revision-scoping bug reported here as a secondary finding, proposed two possible fixes with a recommendation, wrote all the code and specs once a direction was chosen, and wrote the commit message and this description.

The human contribution was direction and decisions: asking for the diagnosis, constraining how the WikiWho payloads were processed (memory), pointing out that local checkouts of the relevant upstream repos already existed, choosing between the two proposed approaches, and authorising the dead-code removal conditional on verifying it was genuinely unreachable.

One correction happened along the way: an early local parser run appeared to show current upstream master producing correct output, which contradicted a later run over pinned revisions. The discrepancy turned out to be a stale remote-tracking ref in a local checkout, resolved by diffing the two copies.

Scale of effort: one commit, written in a single session on 2026-08-07. The great majority of that session was investigation rather than implementation — the code change itself is small (63 insertions, 87 deletions across 5 files) and was written and verified quickly once the approach was settled. Human input across the whole session was roughly 90 words total, spread over six short messages, so reviewers should calibrate accordingly: this had light human review relative to the amount of analysis behind it.

The "What this PR does" summary and the analysis above were also drafted by Claude Code and may contain errors — in particular, the claims about the upstream WhoColor bug are worth independently confirming before acting on them.

## Screenshots

No UI changes. The legend renders the same components with the same strings before and after: `ArticleViewerLegend` matches only exact user ids out of `unhighlightedContributors`, so the status strings in that array are never displayed — they exist only to make the array non-empty. What changes is which revision the existing "contributions are present, but cannot be highlighted" tooltip is describing, plus the removal of a spurious failure state (see below).

Capturing a meaningful before/after would require driving WhoColor into a *partially* highlighted state for a specific revision, which the existing feature spec infrastructure doesn't do — `spec/features/article_viewer_spec.rb` depends on the live WikiWho service and both of its examples are already marked pending for an unrelated CI CORS issue.

- **The production symptom still needs an upstream deploy.** Nothing in this PR makes authorship highlighting work again on the affected articles. That needs `wikiwho-api.wmcloud.org` redeployed with WhoColor master, and `wikiwho_api`'s `lib/WhoColor` submodule pointer bumped past `9265d66`. Both are outside this repo.
- **Behaviour change worth a second look:** the fallback can no longer fail, because it no longer makes a request. Previously a failed metadata fetch called `setFailureMessage`, so the legend reported an authorship failure even when highlighting had actually succeeded, and left `unhighlightedContributors` empty — which stalled `legendStatus` on `loading`. Both of those states are now unreachable. I believe that's strictly an improvement, but it does mean the legend always resolves to `ready`.
- **Reliance on `tokens` being present** in the whocolor payload. It is present on every `success: true` response. If it were ever absent, the editor set is empty and every unhighlighted editor falls to the "no contributions" branch — the same outcome as today's failed metadata fetch, so it degrades no worse than the current code.
- **Dead-code removal** was verified against direct references, subclasses, dynamic dispatch, any other `rev_content` mention, and git history (`generateWhocolorHtml`'s last caller went in 506bb276e; the old class component's copy in 35931b584). Worth a reviewer's sanity check regardless, since deletions are harder to un-miss than additions.
- **Not covered by an automated test:** the specific end-to-end behaviour this fixes — that the fallback now answers for the toggled revision — is exercised only indirectly. The added specs cover the token pass-through at the API layer, not the hook's use of it.

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipAPI.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipAPI.js
@@ -23,15 +23,6 @@ export class AuthorshipAPI extends ArticleViewerAPI {
     });
   }
 
-  generateWhocolorHtml() {
-    const url = this.builder.wikiwhoColorRevisionURL();
-    return fetch(`${url}&origin=*`, {
-      headers: {
-        'Content-Type': 'application/javascript'
-      }
-    }).then(response => this.__handleFetchResponse(response));
-  }
-
   fetchWhocolorHtml(lastRevisionId) {
     let attempts = 0;
     const MAX_RETRY_ATTEMPTS = 5;
@@ -81,7 +72,12 @@ export class AuthorshipAPI extends ArticleViewerAPI {
         // The shell's __processHtml does the relative-link rewriting; if the
         // WhoColor payload lacked HTML, surface a failure marker (as before).
         const processed = this.__processHtml(response.extended_html);
-        return processed || { whocolorFailed: true };
+        if (!processed) return { whocolorFailed: true };
+        // The payload's per-token authorship rides along with the HTML. It is the
+        // same data the rev_content endpoint would return, already scoped to the
+        // requested revision, so callers can tell "edits present but not
+        // highlightable" from "no edits here" without a second request.
+        return { ...processed, tokens: response.tokens };
       });
   }
 
@@ -92,23 +88,6 @@ export class AuthorshipAPI extends ArticleViewerAPI {
         'Content-Type': 'application/javascript'
       }
     }).then(response => this.__handleFetchResponse(response));
-  }
-
-  fetchWikitextMetaData() {
-    const url = this.builder.wikiwhoColorRevisionURL();
-    return fetch(`${url}&origin=*`, {
-      headers: {
-        'Content-Type': 'application/javascript'
-      }
-    }).then(response => this.__handleFetchResponse(response))
-      .then((response) => {
-        if (response.error) throw new Error(this.__setException({ status: 404 }));
-        const revisionId = 'revisions'; // Get the first (and presumably only) revision ID
-        const revisionData = response[revisionId]?.[0];
-        if (!revisionData) throw new Error('Invalid response data');
-        const { tokens } = Object.values(revisionData)[0];
-        return { tokensForRevision: tokens };
-      });
   }
 }
 
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipAPI.spec.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipAPI.spec.js
@@ -26,6 +26,36 @@ describe('AuthorshipAPI', () => {
     expect(typeof api).toEqual('object');
   });
 
+  describe('.fetchWhocolorHtml()', () => {
+    it('should return the payload tokens alongside the processed html', async () => {
+      fetchMock.mockImplementationOnce(() => Promise.resolve({
+        ok: true,
+        json: () => Promise.resolve({
+          success: true,
+          extended_html: '<p><a href="/wiki/Bears">Bears</a></p>',
+          tokens: [[0, 'bears', 1, [], [], '48311939', 0]]
+        })
+      }));
+      const api = new AuthorshipAPI({ builder });
+      const actual = await api.fetchWhocolorHtml(1253430861);
+
+      // The tokens carry the per-token authorship for the requested revision, so
+      // the caller never has to ask a second endpoint which editors are present.
+      expect(actual.tokens).toEqual([[0, 'bears', 1, [], [], '48311939', 0]]);
+      expect(actual.html).toEqual('<p><a href="https://en.wikipedia.org/wiki/Bears">Bears</a></p>');
+    });
+
+    it('should return a failure marker when the payload has no html', async () => {
+      fetchMock.mockImplementationOnce(() => Promise.resolve({
+        ok: true,
+        json: () => Promise.resolve({ success: true, tokens: [] })
+      }));
+      const api = new AuthorshipAPI({ builder });
+
+      expect(await api.fetchWhocolorHtml(1253430861)).toEqual({ whocolorFailed: true });
+    });
+  });
+
   describe('.fetchUserIds()', () => {
     it('should make a request to the builder.wikiUserQueryURL', async () => {
       const api = new AuthorshipAPI({ builder });
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipURLBuilder.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipURLBuilder.js
@@ -3,7 +3,7 @@ import URLBuilder from '@components/common/ArticleViewer/utils/URLBuilder';
 const WIKIWHO_DOMAIN = 'https://wikiwho-api.wmcloud.org';
 
 // Builds the URLs the authorship/WhoColor feature needs: the WikiWho coloring
-// endpoints and the MediaWiki user-id query. Extends the shell's URLBuilder so it
+// endpoint and the MediaWiki user-id query. Extends the shell's URLBuilder so it
 // inherits wikiURL() (used by the user query); the shell never imports this.
 export class AuthorshipURLBuilder extends URLBuilder {
   wikiwhoColorURL(lastRevisionId) {
@@ -16,17 +16,6 @@ export class AuthorshipURLBuilder extends URLBuilder {
     // See https://github.com/wikimedia/wikiwho_api/pull/4/commits/8f421a20c62288a1a29bcef75fffa0a21d2c92a6
     const revisionId = lastRevisionId || 0;
     const url = `${WIKIWHO_DOMAIN}/${language}/whocolor/v1.0.0-beta/${encodeURIComponent(title)}/${revisionId}/`;
-    return url;
-  }
-
-  wikiwhoColorRevisionURL() {
-    const { language, title } = this.article;
-
-    if (!language) throw new TypeError('Article language is missing!');
-    if (!title) throw new TypeError('Article title is missing!');
-
-    const query = '?o_rev_id=true&editor=true&token_id=true&out=true&in=true';
-    const url = `${WIKIWHO_DOMAIN}/${language}/api/v1.0.0-beta/rev_content/${encodeURIComponent(title)}/${query}`;
     return url;
   }
 
--- a/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipURLBuilder.spec.js
+++ b/app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipURLBuilder.spec.js
@@ -28,24 +28,6 @@ describe('AuthorshipURLBuilder', () => {
     });
   });
 
-  describe('#wikiwhoColorRevisionURL', () => {
-    it('should create a wikiwhoColorRevisionURL when given a valid article', () => {
-      const helper = new AuthorshipURLBuilder({ article: defaults.article });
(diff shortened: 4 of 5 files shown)


ragesoss:
A Claude Code review pass over this branch. As with any AI-driven review, this is
provisional: it is worth reading and understanding, but not worth trusting
automatically or acting on without judgment. Some of it may be wrong.

## Verification performed

ESLint clean on all five changed files. Full Jest suite passes (65 suites, 422
tests). A grep across `app/ lib/ test/ spec/` for `generateWhocolorHtml`,
`fetchWikitextMetaData`, `wikiwhoColorRevisionURL`, and `tokensForRevision`
returns nothing, so the dead-code removal appears complete.

`TOKEN_EDITOR_INDEX = 5` was checked against three independent sources in the
upstream repos:

- `wikiwho_api/whocolor/handler.py:127-131` builds the tuple as
  `[conflict_score, str, o_rev_id, in, out, class_name, age]`
- `wikiwho_api/whocolor/swagger_data.py:28-40` documents the same order
- upstream WhoColor's own userscript uses the same index for the same purpose:
  `addClass("editor-token token-editor-" + token_data[5])`
  (`lib/WhoColor/userscript/whocolor.user.js:559`)

The revision does reach the tokens: `whocolor/views.py:73` passes `rev_id` into
`WhoColorHandler`, which calls `get_whocolor_data(self.rev_id)`
(`handler.py:98`). The bug described in the PR body looks real.

One improvement the description does not claim: index 5 is `class_name`, and
`parser.py:140` writes the span as `token-editor-{class_name}`. The fallback now
compares against the same identifier space the highlighting regex matches. The
old path compared against `rev_content`'s `editor` field, which is the raw editor
id — for anonymous editors that is `0|<IP>` where the span class is its md5
(`handler.py:109-112`). Not reachable via course users, but the new code is
structurally sounder and not merely revision-correct.

## Findings

**1. The choice of `tokens` over `present_editors` is load-bearing and
undocumented.** The same payload also carries `present_editors` —
`[editor_name, class_name, authorship_score]` per editor, a few dozen entries
rather than a multi-megabyte token array. It looks like exactly the answer this
code wants, and it is the obvious future optimization for anyone reading the
memory comment at `useAuthorshipHighlighting.js:125`. It would silently
reintroduce the bug. `present_editors` is accumulated by
`WikiMarkupParser.__set_token` (`parser.py:43-62`) only for tokens the parser
locates in the wikitext, and skipped entirely for tokens it cannot place — so it
is degraded by precisely the span-annotator desync this fallback exists to
survive, and on a Stratovolcano-class article it would be near-empty. `tokens`
comes straight off `whocolor_data` and is parser-independent. Worth a sentence in
the comment so this is not swapped later.

**2. The PR description overstates the old symptom.** It says the failed metadata
fetch meant "the legend reported an authorship failure even when highlighting had
actually succeeded." `ArticleViewerLegend` renders `failureMessage` only under
`status === 'failed'`, and the removed path set `failureMessage` without
`whoColorFailed`, so `legendStatus` computed to `'loading'` and the message was
never displayed. The second half of that sentence — that it stalled `legendStatus`
on `loading` — holds. This does not change the verdict on the removal.

**3. Pre-existing, but now pinned by a new spec:** `{ whocolorFailed: true }` has
no consumer anywhere in the codebase. When it fires, the hook calls
`setWhoColorHtml(undefined)`; `pending` is cleared only by `highlightAuthors` or
the `catch`, neither of which runs, so the viewer stays pending and the legend
stays on `loading` indefinitely. This behaviour is identical before and after this
PR, and the "(as before)" comment is accurate about intent. It is only worth noting
that `AuthorshipAPI.spec.js:55` now locks in a return shape that nothing reads.

Checked and cleared: the `setWhoColorEditors`-before-`setWhoColorHtml` ordering is
safe. The app mounts via `createRoot` (`components/app.jsx:15`), so React 18
auto-batches both updates into one render; it would also hold under legacy
semantics, for the reason the code comment gives. The intermediate array created
by `new Set(tokens.map(...))` is dwarfed by the token array itself.

## Coverage gap, and a spec that demonstrates the fix

The PR's specs cover the API pass-through. Nothing covers the hook, which is where
the fix lives. This repo has no `renderHook` available — `enzyme` is on a React 17
adapter and `react-test-renderer` is v19 against React 18 — but the hook can be
driven directly with `react-dom/client` under jsdom, and the returned `legend` is
an unrendered element whose props can be inspected without a redux Provider.

A throwaway spec was written during the review with three examples: token index 5
resolving to present-but-unhighlighted; an absent editor falling to "no
contributions"; and the toggled revision hitting `/whocolor/.../900/` exactly once
with no `rev_content` call. All three pass on this branch and all three fail when
the three source files are reverted to `HEAD~1`, so the behaviour change is
demonstrated rather than only argued.

<details>
<summary>Exploratory spec (AI-written, not committed)</summary>

```js
import React from 'react';
import { act } from 'react-dom/test-utils';
import { createRoot } from 'react-dom/client';

import useAuthorshipHighlighting from '@components/common/ArticleViewer/hooks/useAuthorshipHighlighting';

describe('useAuthorshipHighlighting', () => {
  const article = { title: 'Stratovolcano', language: 'en', project: 'wikipedia' };
  const STUDENT_USERID = 48311939;
  // [conflict_score, str, o_rev_id, in, out, editor/class_name, age]
  const tokenOwnedBy = (editor, str) => [0, str, 1253430861, [], [], editor, 0];

  // The WhoColor span-annotator bug: a payload whose HTML contains no span for the
  // student even though the student owns tokens in that revision. This is the state
  // the fallback exists to describe.
  const html = '<p><span class="editor-token token-editor-999" id="token-0">Lava</span></p>';

  let fetchCalls;
  let root;
  let container;

  const mockFetches = (tokens) => {
    fetchCalls = [];
    global.fetch = jest.fn((url) => {
      fetchCalls.push(url);
      if (url.includes('list=users')) {
        return Promise.resolve({
          ok: true,
          json: () => Promise.resolve({
            query: { users: [{ name: 'Student', userid: STUDENT_USERID }] }
          })
        });
      }
      if (url.includes('/whocolor/')) {
        return Promise.resolve({
          ok: true,
          json: () => Promise.resolve({ success: true, extended_html: html, tokens })
        });
      }
      return Promise.reject(new Error(`unexpected fetch: ${url}`));
    });
  };

  // Render the hook in a throwaway component. The legend it returns is an unrendered
  // React element, so its props can be inspected without a redux Provider.
  const renderHook = async (props) => {
    let latest;
    const Probe = () => {
      latest = useAuthorshipHighlighting(props);
      return null;
    };
    await act(async () => {
      root.render(<Probe />);
    });
    // Let the delayed whocolor request (setTimeout 0) and its promise chain settle.
    await act(async () => {
      jest.advanceTimersByTime(50);
      await Promise.resolve();
    });
    return () => latest;
  };

  const baseProps = {
    article,
    users: ['Student'],
    assignedUsers: [],
    showArticleFinder: false,
    isOpen: true,
    parsedSettle: { id: 1, ok: true },
    fetchArticleDetails: () => {},
    requestTitleVerification: () => {},
  };

  beforeEach(() => {
    global.IS_REACT_ACT_ENVIRONMENT = true;
    jest.useFakeTimers();
    container = document.createElement('div');
    document.body.appendChild(container);
    root = createRoot(container);
  });

  afterEach(() => {
    act(() => root.unmount());
    container.remove();
    jest.useRealTimers();
    delete global.fetch;
  });

  it('reads the editor id from token index 5 and reports the student as present-but-unhighlighted', async () => {
    mockFetches([tokenOwnedBy('999', 'Lava'), tokenOwnedBy(String(STUDENT_USERID), 'dome')]);

    const get = await renderHook({ ...baseProps, revisionId: 1253430861 });

    // The numeric user id in this array is what makes the legend show the
    // "contributions are present, but cannot be highlighted" tooltip.
    expect(get().legend.props.unhighlightedContributors).toContain(STUDENT_USERID);
  });

  it('reports no contributions when the student owns no tokens in the payload', async () => {
    mockFetches([tokenOwnedBy('999', 'Lava')]);

    const get = await renderHook({ ...baseProps, revisionId: 1253430861 });

    expect(get().legend.props.unhighlightedContributors).not.toContain(STUDENT_USERID);
    expect(get().legend.props.unhighlightedContributors).toContain(
      `No contributions found in this version for user ID ${STUDENT_USERID}`
    );
  });

  it('answers from the toggled revision, with no second wikiwho request', async () => {
    mockFetches([tokenOwnedBy(String(STUDENT_USERID), 'dome')]);

    await renderHook({ ...baseProps, revisionId: 900 });

    const wikiwhoCalls = fetchCalls.filter(url => url.includes('wikiwho-api'));
    expect(wikiwhoCalls).toHaveLength(1);
    expect(wikiwhoCalls[0]).toContain('/whocolor/v1.0.0-beta/Stratovolcano/900/');
    // Before this PR a second request went to rev_content with no revision id,
    // so the fallback described the current version instead of revision 900.
    expect(wikiwhoCalls.some(url => url.includes('rev_content'))).toBe(false);
  });
});
```

</details>

CI was still pending at the time of writing.

*Review performed in a Claude Code session (~10 minutes, 2 interactions). Sage
asked for the review and approved posting it, and did not independently verify any
of the findings — the code reading, the upstream-source checks, the spec runs, and
the `HEAD~1` comparison were all done by Claude Code and are reported here as run.
The exploratory spec above is AI-written; Claude Code ran it, Sage did not review
it. This is Sage's own PR, so this comment is a record of an AI review pass rather
than independent human review.*

PR #6995, opened 2026-08-08
Preserve YouTube iframe embeds in timeline block editor (Fixes #6966)

### LINKED ISSUE #6966: Embedded YouTube videos aren't easy to copy from one Timeline block to another
The new rich text editor for editing timeline blocks is buggy when it comes to handling embedded video.

The default timeline includes blocks that embed a few YouTube videos as iframes. When you click 'Edit' on such a block, the iframes are stripped out — and the HTML for them is gone even if you switch to HTML mode, so you can't easily copy and paste it.

It's possible to add and save the block content including the videos, if you paste it into HTML mode and then save without switching out of HTML mode, but any usage of the visual rich text editor will strip out the videos.

ArchieTyagi:
## What this fixes

Fixes #6966.

Embedded YouTube videos were being stripped when editing timeline blocks. Opening a block that contained an `<iframe>` in the editor removed the embed, and it couldn't be recovered from source view either, because the element had already been dropped from the editor's content on load.

## Cause

The timeline block editor was migrated from TinyMCE to a TipTap-based editor (`WysiwygEditor`). TipTap only keeps elements that its schema defines. The current schema is configured with StarterKit only, which has no node for `<iframe>`. So when a block's saved HTML is loaded into the editor, TipTap parses it against the schema, finds no matching node for the iframe, and discards it. Source view then reflects the already-stripped content, which is why the markup can't be recovered there.

## Change

Adds a custom TipTap node (`youtube_iframe.js`) that recognizes `<iframe>` elements and registers it in `WysiwygEditor`'s extensions list. This lets YouTube embeds survive the edit round-trip.

The node's `parseHTML` restricts matches to YouTube embed hosts (`youtube.com/embed` and `youtube-nocookie.com/embed`) and rejects any other iframe `src`. Since timeline content is user-generated, this avoids allowing arbitrary iframes, which would be an XSS/clickjacking surface. The node is marked `atom: true` so the embed is treated as a single, self-contained block.

Files changed:
- `app/assets/javascripts/components/common/youtube_iframe.js` (new)  the custom TipTap iframe node, restricted to YouTube hosts.
- `app/assets/javascripts/components/common/wysiwyg_editor.jsx`  imports the node and adds it to the editor's extensions.

## Testing

- Expected: a YouTube embed added to a timeline block persists through editing, saving, and reopening the block.
- Expected: a non-YouTube iframe is rejected (stripped), so only YouTube embeds are allowed.

I originally started this against the previous TinyMCE editor; master has since migrated to TipTap, so I reworked the fix for the new editor. I've verified the change builds, but my local dev environment is mid-migration to the newer Ruby/Node requirements on current master, so I haven't captured a local before/after recording yet  I'll add one once my environment is rebuilt. I'm relying on CI to validate the build in the meantime.

## Note for reviewers

I've scoped the fix to YouTube hosts at the editor level. If you'd prefer embeds handled differently  a broader allowlist , or leaving host validation to server-side sanitization  I'm happy to adjust.

### DIFF
--- a/app/assets/javascripts/components/common/wysiwyg_editor.jsx
+++ b/app/assets/javascripts/components/common/wysiwyg_editor.jsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useEditor, EditorContent } from '@tiptap/react';
 import StarterKit from '@tiptap/starter-kit';
+import YoutubeIframe from './youtube_iframe';
 
 // A lightweight TipTap-based rich text editor. It replaces the previous TinyMCE
 // editor used by TextAreaInput's `wysiwyg` mode. The external contract is the
@@ -45,6 +46,7 @@ const WysiwygEditor = ({ value, onChange, onFocus, onBlur, invalid }) => {
   const editor = useEditor({
     extensions: [
       StarterKit.configure({ link: { openOnClick: false } }),
+      YoutubeIframe,
     ],
     content: value || '',
     immediatelyRender: true,
--- a/app/assets/javascripts/components/common/youtube_iframe.js
+++ b/app/assets/javascripts/components/common/youtube_iframe.js
@@ -1,0 +1,38 @@
+import { Node, mergeAttributes } from '@tiptap/core';
+
+// Allows YouTube <iframe> embeds to survive the editor round-trip. TipTap's
+// schema drops any element it has no node for, which is why embeds were being
+// stripped from Timeline blocks. Restricted to YouTube hosts to avoid allowing
+// arbitrary iframes in user-generated content.
+const YOUTUBE_HOST = /^https:\/\/(www\.)?(youtube\.com|youtube-nocookie\.com)\/embed\//;
+
+export default Node.create({
+  name: 'youtubeIframe',
+  group: 'block',
+  atom: true,
+  selectable: true,
+
+  addAttributes() {
+    return {
+      src: { default: null },
+      width: { default: '560' },
+      height: { default: '315' },
+      frameborder: { default: '0' },
+      allowfullscreen: { default: 'true' },
+    };
+  },
+
+  parseHTML() {
+    return [{
+      tag: 'iframe',
+      getAttrs: (el) => {
+        const src = el.getAttribute('src') || '';
+        return YOUTUBE_HOST.test(src) ? {} : false;
+      },
+    }];
+  },
+
+  renderHTML({ HTMLAttributes }) {
+    return ['iframe', mergeAttributes(HTMLAttributes)];
+  },
+});
--- a/app/assets/javascripts/components/common/youtube_iframe.js
+++ b/app/assets/javascripts/components/common/youtube_iframe.js
@@ -2,8 +2,9 @@ import { Node, mergeAttributes } from '@tiptap/core';
 
 // Allows YouTube <iframe> embeds to survive the editor round-trip. TipTap's
 // schema drops any element it has no node for, which is why embeds were being
-// stripped from Timeline blocks. Restricted to YouTube hosts to avoid allowing
-// arbitrary iframes in user-generated content.
+// stripped from Timeline blocks. Matches are restricted to YouTube embed
+// hosts to keep embed content consistent. Note this is a content-fidelity
+// rule, not a security boundary: source mode passes raw HTML through unparsed.
 const YOUTUBE_HOST = /^https:\/\/(www\.)?(youtube\.com|youtube-nocookie\.com)\/embed\//;
 
 export default Node.create({
@@ -15,10 +16,14 @@ export default Node.create({
   addAttributes() {
     return {
       src: { default: null },
-      width: { default: '560' },
-      height: { default: '315' },
-      frameborder: { default: '0' },
-      allowfullscreen: { default: 'true' },
+      width: { default: null },
+      height: { default: null },
+      frameborder: { default: null },
+      allowfullscreen: { default: null },
+      allow: { default: null },
+      title: { default: null },
+      class: { default: null },
+      referrerpolicy: { default: null },
     };
   },
 


ArchieTyagi:
Hey @ragesoss, this should be ready for review when you have time. Please Review it Thank you!

ragesoss (commented):
A note on how to read this: the review below was produced with Claude Code, and it should be read rather than obeyed. Some of it is verified fact (marked as such), some is judgment that may be wrong. Use or discard each point based on your own reading of the code.

## What was verified locally

The diagnosis in the PR description holds up. The branch was checked out, `yarn build` run (exit 0), and the actual default-timeline markup from `config/wizard/researchwrite/content.yml:346` was round-tripped through two editors — one with StarterKit only (master's behavior) and one with the added node.

Master:

```
…<p></p><h4>Demo: publishing a new article</h4><p></p>
```

This branch:

```
…<p></p><h4>Demo: publishing a new article</h4><iframe src="https://www.youtube-nocookie.com/embed/gL3jWfMwawE" width="100%" height="400px" frameborder="0" allowfullscreen=""></iframe><p></p>
```

The delta is exactly the iframe. The dropped `<div style>` wrapper, the `target="_blank" rel="noopener…"` added to the link, and the stray empty `<p></p>` are all identical on master and are not caused by this change. That is a clean, well-scoped fix.

Also confirmed working: the block's original `width="100%"` / `height="400px"` survive rather than being overwritten by the node's defaults; a second edit round-trip is idempotent; the lookalike host `https://www.youtube.com.evil.example.com/embed/x` is correctly rejected; an embed written inline inside a paragraph survives (ProseMirror splits the paragraph); and multiple embeds in one block survive.

One hazard was specifically checked for and is not present: `htmlOf()` in `wysiwyg_editor.jsx:16` returns `''` when `editor.isEmpty`, and callers treat that as "no content". A block containing only a video reports `isEmpty === false`, so there is no wipe-on-save risk.

## Findings

**1. Protocol-relative and `http://` embed URLs are silently destroyed.** The regex is anchored on `^https:\/\/` and is case-sensitive, so all three of these lose the embed entirely when a block is opened in the visual editor:

```
//www.youtube.com/embed/abc
http://www.youtube.com/embed/abc
https://WWW.YOUTUBE.COM/embed/abc
```

This is the same data loss as #6966, narrowed to a smaller set of URLs, and it happens silently at parse time — the person editing the block gets no signal that a video was just deleted. Matching on host and path rather than the full URL string (and adding the `i` flag) would close this, and normalizing the stored `src` to `https` on parse would upgrade insecure embeds as a side benefit. Everything currently in the repo uses `https://www.youtube-nocookie.com`, so this is verified as a code behavior, not as a confirmed problem in production data — worth a look at what real timeline blocks actually contain before deciding how much it matters.

**2. The attribute round-trip is lossy.** `addAttributes` declares only `src`, `width`, `height`, `frameborder`, and `allowfullscreen`, so `class`, `title`, `allow`, and `referrerpolicy` are dropped when an existing embed passes through the editor. `title` is the one worth adding: an iframe with no accessible name is a WCAG 4.1.2 problem, and it would be silently stripped from any embed that currently has one. `class` is used on embeds in training content today (`7114-ai-detection-via-pangram.yml`, `7116-why-ai-is-bad-for-wikipedia-video.yml`) — a different surface than timeline blocks, but the same node would apply if this editor is reused there.

**3. Attributes are invented that were not in the source.** A bare `<iframe src="…">` comes back as `width="560" height="315" frameborder="0" allowfullscreen="true"`. Functionally harmless, but it means the editor rewrites markup the author did not write, and `560`/`315` is not this site's convention — the wizard content uses `100%` / `400px`. `default: null` for those attributes would be more faithful.

**4. The security framing in the PR description is stronger than what the code provides.** The description presents the host restriction as closing an XSS/clickjacking surface. In practice `onSourceChange` (`wysiwyg_editor.jsx:112-116`) passes the raw textarea value straight to `onChange` with no schema parsing, and no server-side sanitization of block content was found. So the allowlist is a content-fidelity rule, not a security boundary. Nothing here needs to change — but it is worth stating plainly in the description so that nobody later treats it as a boundary it is not.

**5. `@tiptap/core` is imported but is not a declared dependency.** `package.json` lists only `@tiptap/pm`, `@tiptap/react`, and `@tiptap/starter-kit`. The import resolves through hoisting and builds fine, but would break under a stricter installer (pnpm, Yarn PnP). Adding it as an explicit dependency is the safer form.

**6. No test coverage.** The editor has none today, so this is not a regression in standards. But this node is pure parse/serialize logic with a branch on URL shape, which makes it unusually cheap to test — and findings 1 through 3 above are all things a handful of assertions would have surfaced.

## Not verified

The in-editor interaction was not exercised in a real browser. A live `<iframe>` inside `contenteditable` generally captures pointer events, so selecting or deleting an embed may be awkward without a node view wrapping it in a non-editable container. Worth a manual check.

Findings 1 through 3 above were each confirmed by an assertion in a throwaway jest spec written during the review (18 examples, not included here). It can be shared if it would be useful as a starting point for test coverage — note that it pins *current* behavior, so the expectations covering findings 1 through 3 would need inverting if those are acted on.

---

*Drafted in a Claude Code session (~20 min, 4 interactions). Sage directed the session, reviewed the findings as they were presented, and approved posting; the individual findings were not independently re-derived. What was actually run locally: `yarn build` (exit 0), the full jest suite (66 suites / 444 examples, all passing), `spec/features/wysiwyg_input_spec.rb`, `spec/features/timeline_editing_spec.rb`, ESLint on the changed files, and the throwaway spec noted above — the round-trip outputs quoted at the top are real output from that run, not reasoning. Findings 4 and 5 are code reading, not execution. The in-editor interaction note is unverified.*

ragesoss:
I think points 2, 3, and 4 from this review merit changes (just to the PR description for 4). Point 1 seems okay to leave alone, as well as 5 and 6.

ArchieTyagi:
@ragesoss thanks for the review! i pushed fixes for 2, 3 and 4 : 
added title, class, allow and referrerpolicy so existing embeds don't lose them (title especially since it matters for accessibility)
changed the defaults to null so it stops adding width/height/frameborder that weren't there originally
updated the description to make clear the host check is just for keeping embeds consistent, not a security thing

left 1, 5 and 6 as you said. lmk if anything else needs fixing!
Happy to fix Anychange

ragesoss (approved):
Fixes for findings 2 and 3 were verified locally on the new commit (103f021f): `title`, `class`, `allow`, and `referrerpolicy` now survive the edit round-trip (checked against the real `class="youtube-short"` shape used in training content), and a bare `<iframe src="…">` comes back with no invented attributes. The wizard default-timeline embed still round-trips with its original `width="100%"`/`height="400px"`, non-YouTube and lookalike hosts are still rejected, and the round-trip is idempotent. `yarn build` (exit 0), the full jest suite (74 suites / 473 examples), and the wysiwyg and timeline-editing feature specs all pass on the branch merged with current master. No new issues were found in the delta.

*Drafted in a Claude Code session (~15 min, 2 interactions; Sage reviewed the findings and this comment before posting). What was actually run locally: `yarn build`, a 10-example verification spec written during review (AI-generated, exercised but not committed), the full jest suite, ESLint, and the two affected feature specs.*

PR #6996, opened 2026-08-10
feat: add daily stats system csv export

lakshita10341:
## What this PR does
This pull request implements an admin-only daily system statistics snapshot CSV export feature (SystemStat) with a configurable date range. It also updates the System Stats analytics dashboard UI with a clean, unified tabbed interface to easily switch between Course-Level Exports and Daily System Snapshots.

### Key Changes:
- **Backend & Analytics (`SystemStatsCsvBuilder`)**: Streams daily metric snapshots (`snapshot_date`, edits, views, articles, characters, active/archived programs, new editors, facilitators). Defaults to the past 30 days.
- **Async Worker (`ReportCsvWorker`)**: Adds background generation for `system_daily_stats_csv` via Sidekiq with automatic cleanup.
- **Controller & Endpoint (`ReportsController`)**: Adds `GET /system_daily_stats_csv` with admin authentication and date format/range validation (`422`).
- **Frontend Tabbed UI (`SystemCsvExportBar`)**: Adds tabs for "Course Level Exports" and "Daily System Snapshots", background polling, and direct file download.
- **Test Coverage**: Adds builder unit specs (`system_stats_csv_builder_spec.rb`) and controller request specs (`system_daily_stats_csv_spec.rb`).

## AI usage
AI was used for code writing, query and performance optimization, and database seeding for test scenarios.

[Screencast from 2026-08-11 02-55-45.webm](https://github.com/user-attachments/assets/d9d6fed3-ced2-4227-b4ad-513f2852d0b5)


Abishekcs:
LGTM!

ragesoss (changes requested):
Review notes from a Claude Code session. These are observations to read and weigh, not conclusions to act on automatically — AI analysis is provisional and some of it may be wrong. Use or ignore each item based on your own judgment of the code.

The feature works. The tabbed UI was exercised end to end in a browser spec (switch tabs, pick a range, export, poll, download) and the happy path passes; the 422 for a reversed range surfaces correctly in the UI. Two defects and one CI-blocking lint offense were reproduced locally, described below.

## CI

The current red build is unrelated: `spec/lib/revision_data_manager_spec.rb` hit a MediaWiki API 429. Worth noting that the workflow's "Ruby linting" step runs *after* rspec, so the flake meant RuboCop never ran on this branch — see the lint item below, which will fail a green rerun.

## Reproduced defects

### 1. CSV rows are emitted in id order, not `snapshot_date` order

`lib/analytics/system_stats_csv_builder.rb:28`

`SystemStat.for_date_range` ends in `.order(:snapshot_date)`, but `find_each` discards any scope ordering and forces primary-key order — Rails also logs a "Scoped order is ignored, it's forced to be batch order" warning on every run. Today the two orders coincide because `SystemStatUpdateWorker` writes one row per day in ascending order, so the export looks correct. Any row created out of sequence — a backfilled missing day, a re-created row — lands in the wrong place in the file.

Inserting three rows so that id order is the reverse of date order produces:

```
expected: ["2026-08-08", "2026-08-12", "2026-08-15"]
     got: ["2026-08-15", "2026-08-12", "2026-08-08"]
```

### 2. The generated filename is not sanitized, and a slash-delimited date breaks the job permanently

`app/controllers/reports_controller.rb:241`

`build_system_daily_stats_filename` interpolates the raw param strings. The sibling method for the existing endpoint, `build_system_csv_filename`, ends in `.tr('/', '-')` and carries a comment explaining why: `# Filename does not have to contain '/' char because it's interpreted as a route`.

`Date.parse` accepts slash-delimited dates, so `?start_date=2020/01/02` passes the new format validation and then the write fails:

```
Errno::ENOENT (No such file or directory @ rb_sysopen -
  public/system/analytics/system-daily-stats-from-2020/01/02-2026-08-17.csv)
```

With Sidekiq inline (test env) that surfaces as a 500. In production the controller returns 202, the job then fails on every Sidekiq retry, and the UI polls for the full two minutes before showing "still generating" — with no indication of what went wrong. The UI's `type="date"` inputs can't produce this, but a hand-edited URL can.

Relatedly, `Date.parse` is lenient enough to accept traversal-decorated values — `Date.parse('2020-01-02/../../../../tmp/pwn')` returns `2020-01-02`, and that whole string reaches the filename. Nothing was written outside `public/system/analytics` in testing, because the leading path segment (`system-daily-stats-from-2020-01-02`) isn't an existing directory, so this does not appear exploitable as written. But the only thing preventing it is a directory that happens not to exist.

Building the filename from the parsed `Date` objects rather than the raw params would close both cases at once (and keeping the `.tr('/', '-')` guard would be cheap defense in depth).

## RuboCop offense — will fail CI

```
lib/analytics/system_stats_csv_builder.rb:25:3: C: Metrics/MethodLength:
  Method has too many lines. [17/16]
```

The two parallel lists in `generate_csv` (`CSV_HEADERS` and the row array) are also a drift risk — they have to stay in the same order by hand. Deriving the row from the headers fixes the offense and the drift risk together, since every header is a column on `SystemStat`:

```ruby
snapshots.each do |snapshot|
  csv << CSV_HEADERS.map { |header| snapshot.public_send(header) }
end
```

## Suggestions

### Date validation already exists in `SystemCsvFilterValidator`

`lib/analytics/system_csv_filter_validator.rb:39` (`validate_dates`) already performs the same `Date.parse`-and-rescue check that this PR reimplements as `validate_daily_stats_date_formats` / `daily_stats_filter_errors` / `validate_daily_stats_date_range` in the controller. Reusing or extending that class would:

- drop the duplication (and the wording divergence: the new message reads `Invalid start date`, the existing one `Invalid start_date`);
- remove the need for the `# rubocop:disable Metrics/ClassLength` this PR wraps around the whole controller. `ReportsController` is within the 145-line limit on master and measures 193 with these additions, so the disable is masking real growth rather than an arbitrary threshold;
- let `system_csv` benefit from the new start-before-end check, which is a genuine improvement that endpoint currently lacks — it accepts a reversed range silently today.

### `wiki_stats` is loaded on every row and never used

`SystemStat#wiki_stats` is a serialized `text(65535)` column. The builder loads full rows, so every day in the range deserializes a per-wiki hash that no CSV column uses. `SystemStat.recent_monthly_snapshots` already avoids this deliberately (`select(column_names - ['wiki_stats'])`, with a comment about memory). Negligible for the 30-day default; less so for a multi-year admin export.

Selecting only the exported columns also happens to fix the ordering bug, since it requires dropping `find_each` (which needs the primary key in the select and overrides the order anyway):

```ruby
def snapshots
  SystemStat.for_date_range(@start_date, @end_date).select(*CSV_HEADERS)
end
```

This shape was checked locally: ordering preserved, `wiki_stats` not loaded.

### Tab state interactions

`app/assets/javascripts/components/system_stats/system_csv_export_bar.jsx:132,139`

Both tab handlers call `setNotice(null)` unconditionally, so switching tabs during an in-flight export wipes the "being generated in the background" message; it only reappears on the next 6-second poll tick. Confirmed in a browser spec.

Also confirmed: `exporting` is shared across both tabs, so an export started on the Daily tab leaves the Course tab's button disabled and relabelled "Generating Export...". Defensible as a "one export at a time" rule, but currently it reads as the wrong tab being busy. Worth a deliberate decision either way.

### Smaller notes

- The tab header uses bare `<button>`s with an `.active` class. Native buttons are keyboard-operable, so this is not an accessibility blocker, but there is no `role="tablist"` / `role="tab"` / `aria-selected` / `role="tabpanel"`. `app/assets/javascripts/components/common/OverviewStats/overview_stats_tab.jsx` already implements that pattern in this codebase.
- `#676eb4` is defined as `$brand_primary` in `app/assets/stylesheets/_variables.styl:46`. The `#7e22ce` → `#676eb4` recolor is a real improvement — it brings the export bar in line with the rest of the app — but it is not mentioned in the PR description, and the new tab styles introduce hardcoded greys (`#e2e8f0`, `#718096`, `#2d3748`) that aren't in the palette.
- `maxAttempts = 20` lost its `// 20 attempts x 6s = 120s (2 mins) polling window` comment in the refactor. That arithmetic isn't obvious from the constant alone.
- `startExport(buildUrlFn)` takes a thunk that is invoked immediately on the next line; passing the URL string would read the same and do the same thing.

## Coverage gaps

- No spec covers the tabbed UI. `spec/features/system_stats_spec.rb` only asserts that `#react_root` renders, so tab switching, the daily export flow, and the error path are untested.
- Neither reproduced defect above has spec coverage: row ordering and filename construction are both untested.
- `spec/lib/analytics/system_stats_csv_builder_spec.rb` — "returns empty result when start_date is after end_date" tests a state the endpoint can't reach, since the controller 422s that input. Not wrong, but it isn't exercising a reachable path.

## Question

The export includes `new_editors_count_with_preregistration` but not the plain `new_editors_count`, and omits `wiki_stats` entirely. Is that the intended column set?

---

Two exploratory specs were written during the review and are inlined below. They were AI-written and are not proposed as-is for merge, but the ordering and filename examples would be worth adopting in some form alongside the fixes.

One gotcha for rerunning the feature spec: `CsvCleanupWorker.perform_at` has to be stubbed, because Sidekiq runs inline in this suite and the real call deletes the CSV the moment it is written. Another: Chrome renders `type="date"` in the browser locale, so `fill_in` needs `MM/DD/YYYY` keystrokes — passing `%Y-%m-%d` silently yields values like `60810-02-02`, which `Date.parse` happily accepts, so the spec passes an absurd range to the server without failing.

<details>
<summary><b>Unit spec</b> — pins row ordering and filename construction (both examples currently fail)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/analytics/system_stats_csv_builder"

describe 'daily system stats CSV export (exploratory)' do
  describe SystemStatsCsvBuilder do
    # Rows inserted so that primary-key order is the reverse of snapshot_date order,
    # which is what happens whenever a missing day is backfilled after the fact.
    let!(:newest) { create(:system_stat, snapshot_date: 2.days.ago.to_date, total_edits: 300) }
    let!(:middle) { create(:system_stat, snapshot_date: 5.days.ago.to_date, total_edits: 200) }
    let!(:oldest) { create(:system_stat, snapshot_date: 9.days.ago.to_date, total_edits: 100) }

    it 'emits rows in snapshot_date order' do
      csv = described_class.new.generate_csv
      dates = CSV.parse(csv, headers: true).map { |row| row['snapshot_date'] }
      expect(dates).to eq([oldest, middle, newest].map { |s| s.snapshot_date.to_s })
    end
  end

  describe 'ReportsController#system_daily_stats_csv filename handling', type: :request do

    before { login_as admin }

    after do
      FileUtils.rm_rf('public/system/analytics')
    end

    it 'does not put a path separator in the generated filename' do
      # Date.parse accepts slash-delimited dates, so this passes format validation.
      get '/system_daily_stats_csv', params: { start_date: '2020/01/02' }
      expect(response).to have_http_status(:accepted)

      get '/system_daily_stats_csv', params: { start_date: '2020/01/02' }
      expect(response).to have_http_status(:ok)
      expect(response.parsed_body['url']).to eq(
        "/system/analytics/system-daily-stats-from-2020-01-02-#{Time.zone.today}.csv"
      )
    end

    it 'does not let a date param escape the analytics directory' do
      traversal = '2020-01-02/../../../../tmp/wikiedu-pr6996-pwn'
      get '/system_daily_stats_csv', params: { start_date: traversal }
      # Either rejected outright, or accepted with the separators stripped —
      # but never written outside public/system/analytics.
      expect(File).not_to exist('/tmp/wikiedu-pr6996-pwn-' \
                                "#{Time.zone.today}.csv")
    end
  end
end
```

The second filename example passes today, but only incidentally — the write fails with `ENOENT` rather than being rejected.

</details>

<details>
<summary><b>Feature spec</b> — covers the tab flows (4 of 5 pass; the notice-clearing example fails)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe 'System Stats CSV export tabs', type: :feature, js: true do
  let(:admin) { create(:admin, username: 'admin_user') }

  before do
    login_as(admin, scope: :user)
    # Sidekiq runs inline in specs, so the real perform_at would delete the CSV
    # the moment it is written.
    allow(CsvCleanupWorker).to receive(:perform_at)
    create(:system_stat, snapshot_date: Time.zone.today, total_edits: 50_000)
    create(:system_stat, snapshot_date: 3.days.ago.to_date, total_edits: 40_000)
  end

  after do
    FileUtils.rm_rf('public/system/analytics')
  end

  it 'shows the course export tab by default and switches to the daily snapshots tab' do
    visit '/system_stats'
    expect(page).to have_button('Course Level Exports')
    expect(page).to have_button('Daily System Snapshots')

    # Course tab is active by default: its filters are visible.
    expect(page).to have_css('#system-csv-campaign-select')
    expect(page).to have_no_css('#daily-csv-start-date')

    click_button 'Daily System Snapshots'
    expect(page).to have_css('#daily-csv-start-date')
    expect(page).to have_css('#daily-csv-end-date')
    expect(page).to have_no_css('#system-csv-campaign-select')
    expect(page).to have_button('Export Daily Stats CSV')
  end

  it 'generates a daily stats CSV for the chosen range' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'

    # Chrome renders type=date in the browser locale (en-US), so keystrokes
    # must be MM/DD/YYYY even though the submitted value is ISO.
    fill_in 'daily-csv-start-date', with: 7.days.ago.to_date.strftime('%m/%d/%Y')
    fill_in 'daily-csv-end-date', with: Time.zone.today.strftime('%m/%d/%Y')

    click_button 'Export Daily Stats CSV'
    # First poll returns 202; the polling interval is 6s, so allow for it.
    expect(page).to have_content('being generated in the background', wait: 10)
    expect(page).to have_no_button('Generating Export...', wait: 20)

    csv = Dir.glob('public/system/analytics/system-daily-stats-*.csv')
    expect(csv.size).to eq(1)
    expect(csv.first).to end_with(
      "system-daily-stats-from-#{7.days.ago.to_date}-to-#{Time.zone.today}-#{Time.zone.today}.csv"
    )
    body = File.read(csv.first)
    expect(body).to include('snapshot_date')
    expect(body).to include('50000')
    expect(body).to include('40000')
  end

  it 'surfaces the server-side validation error for a reversed date range' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'

    fill_in 'daily-csv-start-date', with: Time.zone.today.strftime('%m/%d/%Y')
    fill_in 'daily-csv-end-date', with: 7.days.ago.to_date.strftime('%m/%d/%Y')

    click_button 'Export Daily Stats CSV'
    expect(page).to have_content('start_date must be before end_date', wait: 10)
    expect(page).to have_button('Export Daily Stats CSV')
  end

  it 'keeps the in-flight export notice visible when switching tabs' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'
    click_button 'Export Daily Stats CSV'
    expect(page).to have_content('being generated in the background', wait: 10)

    click_button 'Course Level Exports'
    # The tab handler calls setNotice(null); the notice only comes back on the
    # next 6s poll tick, so check for it without Capybara's retry window.
    expect(page).to have_content('being generated in the background', wait: 0)
  end

  it 'shares the exporting state across both tabs' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'
    click_button 'Export Daily Stats CSV'
    expect(page).to have_content('being generated in the background', wait: 10)

    click_button 'Course Level Exports'
    # The course tab's own export button is disabled and relabelled by an
    # export the user started on the other tab.
    expect(page).to have_button('Generating Export...', disabled: true, wait: 0)
  end
end
```

</details>

---

*Produced in a Claude Code session (~2 hours, 2 interactions from Sage: one to start the review, one to request this comment). The session checked the branch out over current master, ran `yarn build`, the PR's own specs, and the existing specs for the affected areas (no regressions), ran RuboCop and ESLint on the changed files, and reproduced the two defects and the lint offense with new specs run locally. Sage read the findings summary and approved posting, but did not independently re-verify any individual finding. The exploratory specs mentioned above were written by Claude Code and run, but not line-by-line reviewed by Sage.*

lakshita10341:
`new_editors_count_with_preregistration` is used because this definition was finalized in the earlier discussions and is the standard metric displayed across the entire System Stats feature for consistency. `wiki_stats `was intentionally omitted as it is a serialized text/hash field containing nested per-wiki breakdowns, which doesn't fit cleanly into a flat daily time-series CSV format.

ragesoss (approved):
Re-review notes from a Claude Code session, covering the four commits since `bd7438ca5`. As before: these are observations to read and weigh, not conclusions to act on automatically — AI analysis is provisional and some of it may be wrong. Use or ignore each item based on your own judgment of the code.

Both defects from the previous review are fixed. Each fix was checked by reverting it and confirming the PR's own new spec catches the regression, so the guards are real rather than incidental:

- **Row ordering** — `find_each` → `each` plus `.select(*CSV_HEADERS)`. Reverting to `find_each` fails the new `emits rows in snapshot_date order, not id order` example (`["2026-08-08", "2026-08-16", "2026-08-13"]`). The `.select` also resolves the separate `wiki_stats` note: `snapshot.wiki_stats` now raises `MissingAttributeError`, so the serialized column is no longer deserialized per row.
- **Filename construction** — built from parsed `Date` objects with the `.tr('/', '-')` guard kept. `?start_date=2020/01/02` now yields `from-2020-01-02`, and the traversal-decorated value collapses to the same thing. Reverting the fix fails the new `normalizes slash-delimited dates in the filename` example with a 500.

Also resolved: the `Metrics/MethodLength` offense (CI is green now that the unrelated MediaWiki 429 flake isn't blocking the lint step); the tab-switch notice wipe; the shared `exporting` state mislabelling the other tab's button; the hardcoded hex values, now on `$brand_primary` / `$border_lt` / `$border_med` / `$text` / `$text_med` / `$text_dk` / `$contextual`; the `maxAttempts` arithmetic comment; and the `startExport` thunk.

Verified locally on the branch merged over current master: `yarn build` clean, RuboCop clean on all eight changed Ruby files, ESLint clean on the JSX, and 203 examples / 0 failures across `reports_controller`, `analytics_controller`, all of `spec/lib/analytics/`, `system_csv`, `system_daily_stats_csv`, the `system_stats` feature spec, and the `system_stat` model.

## Partially addressed

### `Metrics/ClassLength` disable

Moving validation into `SystemDailyStatsCsvFilterValidator` brought `ReportsController` from 193 lines to 173, but the file-wide `# rubocop:disable Metrics/ClassLength` is still needed and still present — the limit is 145, and `ReportsController` is under it on master. The disable is masking 28 lines of growth rather than an arbitrary threshold.

The extraction did land the wording fix (`Invalid start_date` now matches the existing validator), but two parts of the original note remain: `SystemDailyStatsCsvFilterValidator#validate_date_formats` still duplicates `SystemCsvFilterValidator#validate_dates`, and `system_csv` still accepts a reversed date range silently — it would have inherited the new start-before-end check for free if the range validation had been added to the existing class.

### ARIA tab wiring

The roles, `aria-selected`, `aria-controls`, and `aria-labelledby` are all in place. Two gaps remain, both measured in the browser:

```
tab="Course Level Exports"   selected=true   aria-controls="tabpanel-courses"  target_in_dom=true
tab="Daily System Snapshots" selected=false  aria-controls="tabpanel-daily"    target_in_dom=false
tab tabindex attrs: nil, nil
```

1. Because the inactive panel is conditionally rendered rather than hidden, the inactive tab's `aria-controls` points at an element that isn't in the DOM. Rendering both panels with `hidden` on the inactive one, or omitting `aria-controls` on the inactive tab, would both resolve it.
2. There is no focus management: both tabs stay in the tab order and there is no arrow-key handler, so `role="tab"` announces "tab, 1 of 2" and promises a keyboard pattern that isn't implemented.

Neither is a blocker — native `<button>`s remain Tab+Enter operable, and `app/assets/javascripts/components/common/OverviewStats/overview_stats_tab.jsx` is equally partial, so this matches existing precedent rather than falling below it. Worth naming because a Wiki Ed admin uses JAWS exclusively and reviews admin surfaces in practice.

## Notes on the new commits

- `72ec31a5a` adds `system_daily_stats_csv` to the `require_signed_in` list, but `require_admin_permissions` already calls `require_signed_in` internally and both 401 specs were already present at `bd7438ca5`. Harmless, but it does not close a hole that was open.
- `3bd96e3f9` merges two focused examples into one `handles full export lifecycle and returns ready status with URL`. `CLAUDE.md` asks for one focused `it` block per behavior; the consolidation is defensible here since the second example already re-asserted the first, but the name now describes two behaviors.
- Checked separately and clean: `params.permit(:start_date, :end_date)` drops non-scalar values, so `?start_date[a]=1` cannot reach `Date.parse` as a `Hash`.
- Still open from the previous review, and unchanged: `system_stats_csv_builder_spec.rb:54` (`returns empty result when start_date is after end_date`) exercises a state the endpoint 422s before the builder is reached.

## Remaining coverage gap

The tab UI is still untested. `spec/features/system_stats_spec.rb` is unchanged and only asserts that `#react_root` renders, so tab switching, the daily export flow, the 422 error path, the notice-preservation fix, and the label-scoping fix have no spec coverage. Blast radius is small — admin-only — but the two bugs just fixed lived in exactly that state machine and both failed silently.

The exploratory feature spec from the review is inlined below, updated for the new behavior. All six examples pass against the current branch; the two that assert the fixes (`keeps the in-flight export notice visible when switching tabs`, `scopes the Generating label to the tab that started the export`) failed before the fix commits. It is offered as a starting point if the tab flows are worth covering, not as a proposed diff.

Two gotchas for running it: `CsvCleanupWorker.perform_at` has to be stubbed, because Sidekiq runs inline in this suite and the real call deletes the CSV the moment it is written. And Chrome renders `type="date"` in the browser locale, so `fill_in` needs `MM/DD/YYYY` keystrokes — passing `%Y-%m-%d` silently yields values like `60810-02-02`, which `Date.parse` happily accepts, so the spec sends an absurd range to the server without failing.

<details>
<summary><b>Feature spec</b> — covers the tab flows and both state-machine fixes (6/6 pass)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe 'System Stats CSV export tabs', type: :feature, js: true do
  let(:admin) { create(:admin, username: 'admin_user') }

  before do
    login_as(admin, scope: :user)
    # Sidekiq runs inline in specs, so the real perform_at would delete the CSV
    # the moment it is written.
    allow(CsvCleanupWorker).to receive(:perform_at)
    create(:system_stat, snapshot_date: Time.zone.today, total_edits: 50_000)
    create(:system_stat, snapshot_date: 3.days.ago.to_date, total_edits: 40_000)
  end

  after do
    FileUtils.rm_rf('public/system/analytics')
  end

  it 'shows the course export tab by default and switches to the daily snapshots tab' do
    visit '/system_stats'
    expect(page).to have_button('Course Level Exports')
    expect(page).to have_button('Daily System Snapshots')

    expect(page).to have_css('#system-csv-campaign-select')
    expect(page).to have_no_css('#daily-csv-start-date')

    click_button 'Daily System Snapshots'
    expect(page).to have_css('#daily-csv-start-date')
    expect(page).to have_css('#daily-csv-end-date')
    expect(page).to have_no_css('#system-csv-campaign-select')
    expect(page).to have_button('Export Daily Stats CSV')
  end

  it 'exposes the tabs with tablist/tab/tabpanel roles' do
    visit '/system_stats'
    expect(page).to have_css('[role="tablist"]')
    expect(page).to have_css('[role="tab"][aria-selected="true"]', text: 'Course Level Exports')
    expect(page).to have_css('[role="tab"][aria-selected="false"]', text: 'Daily System Snapshots')
    expect(page).to have_css('[role="tabpanel"][aria-labelledby="tab-courses"]')

    click_button 'Daily System Snapshots'
    expect(page).to have_css('[role="tab"][aria-selected="true"]', text: 'Daily System Snapshots')
    expect(page).to have_css('[role="tabpanel"][aria-labelledby="tab-daily"]')
  end

  it 'generates a daily stats CSV for the chosen range' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'

    # Chrome renders type=date in the browser locale (en-US), so keystrokes
    # must be MM/DD/YYYY even though the submitted value is ISO.
    fill_in 'daily-csv-start-date', with: 7.days.ago.to_date.strftime('%m/%d/%Y')
    fill_in 'daily-csv-end-date', with: Time.zone.today.strftime('%m/%d/%Y')

    click_button 'Export Daily Stats CSV'
    expect(page).to have_content('being generated in the background', wait: 10)
    expect(page).to have_no_button('Generating Export...', wait: 20)

    csv = Dir.glob('public/system/analytics/system-daily-stats-*.csv')
    expect(csv.size).to eq(1)
    expect(csv.first).to end_with(
      "system-daily-stats-from-#{7.days.ago.to_date}-to-#{Time.zone.today}-#{Time.zone.today}.csv"
    )
    body = File.read(csv.first)
    expect(body).to include('snapshot_date')
    expect(body).to include('50000')
    expect(body).to include('40000')
  end

  it 'surfaces the server-side validation error for a reversed date range' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'

    fill_in 'daily-csv-start-date', with: Time.zone.today.strftime('%m/%d/%Y')
    fill_in 'daily-csv-end-date', with: 7.days.ago.to_date.strftime('%m/%d/%Y')

    click_button 'Export Daily Stats CSV'
    expect(page).to have_content('start_date must be before end_date', wait: 10)
    expect(page).to have_button('Export Daily Stats CSV')
  end

  it 'keeps the in-flight export notice visible when switching tabs' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'
    click_button 'Export Daily Stats CSV'
    expect(page).to have_content('being generated in the background', wait: 10)

    click_button 'Course Level Exports'
    # No Capybara retry window: the notice must survive the tab switch itself.
    expect(page).to have_content('being generated in the background', wait: 0)
  end

  it 'scopes the Generating label to the tab that started the export' do
    visit '/system_stats'
    click_button 'Daily System Snapshots'
    click_button 'Export Daily Stats CSV'
    expect(page).to have_content('being generated in the background', wait: 10)
    expect(page).to have_button('Generating Export...', disabled: true, wait: 0)

    click_button 'Course Level Exports'
    # The course tab keeps its own label; it is disabled but does not claim to
    # be generating an export the user started elsewhere.
    expect(page).to have_no_button('Generating Export...', wait: 0)
    expect(page).to have_button('Export CSV', disabled: true, wait: 0)
  end
end
```

</details>

---

*Produced in a Claude Code session (~25 min, 2 interactions from Sage: one to start the re-review, one to approve this comment). The session checked the branch out over current master and verified locally: `yarn build`, RuboCop and ESLint on the changed files, 203 examples across the affected areas, re-runs of both exploratory specs from the previous review, the two fix-revert checks described above, the `aria-controls` and `tabindex` measurements, and the `ReportsController` class length with the RuboCop disable removed. Sage read the findings summary and approved posting, but did not independently re-verify any individual finding. The inlined feature spec was written and run by Claude Code and is not proposed for the codebase; Sage did not review it line by line.*

PR #6999, opened 2026-08-17
Add admin CSV reports for instructor mentorship program matching

ragesoss:
## What this PR does

Adds two admin-only, campaign-scoped CSV downloads to the `/analytics` page to support the instructor mentorship program, which each term needs to match new instructors who asked for a mentor with experienced instructors who volunteered to be one:

- **Mentor Requests** (`/mentor_requests_csv?campaign[id]=N`) — instructors of campaign courses tagged `mentor_requested`, the tag the assignment wizard's "Mentorship program" panel applies when a first-time instructor asks to be matched with a mentor.
- **Mentor Volunteers** (`/mentor_volunteers_csv?campaign[id]=N`) — instructors who checked the "Become a mentor: …" option in an end-of-term instructor survey, scoped to responses attributable to a course in the selected campaign.

Both CSVs share one column shape: `real_name, username, email, institution, course_title, course_subject, course_term, course_slug`.

Implementation notes:

- Survey questions are cloned with new ids each term (sometimes `(Copy) `-prefixed), so the volunteers report matches the answer-option wording (`/become a mentor/i`, with a SQL `LIKE` prefilter) across all question ids rather than pinning an id. A line-level match on the `"\r\n"`-joined `answer_text` confirms the box was actually checked.
- The response→course link goes through the user's completed `SurveyNotification`s (the reliable path), falling back to `answer_groups.course_id` — a submit-time guess, guarded by the campaign filter — and skipping responses that can't be attributed at all.
- Follows the existing `tagged_courses_csv` pattern: builders in `lib/analytics/` (a shared `MentorshipCsvBuilder` base plus two small subclasses), inline `send_data` actions on `AnalyticsController` behind `require_admin_permissions`, and top-level routes (the `/analytics(/*any)` catch-all makes nested paths unreachable).

## AI usage

This feature was implemented by Claude Code (Fable 5) under direction from @ragesoss, who defined the feature (the staff request for a per-term mentee CSV, the two data sources, and the need for wording-robust survey matching), supplied the survey option wording, and made the scoping decisions (add email/username columns; scope volunteers by the course's campaign). Claude Code explored the codebase, proposed the plan, wrote all the code and specs, wrote the commit message, and drafted this PR description — including the summary above, which may contain errors.

Scale of effort: a single commit produced in one focused session on 2026-08-17, following a plan-mode design pass. The first spec run had two failures (test-setup issue: Rapidfire validates checkbox answers against the question's `answer_options`); everything passed after that fix, and the suite was re-verified after rebasing the work onto current master.

## Screenshots

**01 analytics tools page**

| Before | After |
|--------|-------|
| before | after |

**02 campaigns selected**

| Before | After |
|--------|-------|

- The button labels ("Mentor Requests" / "Mentor Volunteers") and the `campaign` field labels were chosen by Claude Code, matching the terse style of the existing forms on this page — rename if different wording is preferred.
- If a future survey rewords the option away from the phrase "become a mentor", the `MENTOR_OPTION` / `SQL_PREFILTER` constants in `MentorVolunteersCsvBuilder` are the single knob to update.
- The `answer_groups.course_id` fallback can name a sibling same-term course for multi-course instructors; the person and campaign membership are still correct, and the campaign filter discards wrong guesses outside the selected term.
- Submitting either form with the campaign prompt still selected raises (no campaign param) — the same behavior as the existing campaign-intersection form on this page.
- Possible follow-up, deliberately out of scope to keep both CSVs the same shape: an extra column with the exact option text the volunteer checked, useful when wording varies across terms.

### DIFF
--- a/app/controllers/analytics_controller.rb
+++ b/app/controllers/analytics_controller.rb
@@ -5,13 +5,16 @@ require_dependency "#{Rails.root}/lib/analytics/course_statistics"
 require_dependency "#{Rails.root}/lib/analytics/campaign_csv_builder"
 require_dependency "#{Rails.root}/lib/analytics/ungreeted_list"
 require_dependency "#{Rails.root}/lib/analytics/tagged_courses_csv_builder"
+require_dependency "#{Rails.root}/lib/analytics/mentor_requests_csv_builder"
+require_dependency "#{Rails.root}/lib/analytics/mentor_volunteers_csv_builder"
 
 #= Controller for analytics tools
 class AnalyticsController < ApplicationController
   layout 'admin'
   include CourseHelper
   before_action :require_signed_in, only: :ungreeted
-  before_action :require_admin_permissions, only: :tagged_courses_csv
+  before_action :require_admin_permissions,
+                only: %i[tagged_courses_csv mentor_requests_csv mentor_volunteers_csv]
 
   ########################
   # Routing entry points #
@@ -54,6 +57,16 @@ class AnalyticsController < ApplicationController
               filename: "all_courses-#{Time.zone.today}.csv"
   end
 
+  def mentor_requests_csv
+    send_data MentorRequestsCsvBuilder.new(selected_campaign).generate_csv,
+              filename: "mentor_requests-#{selected_campaign.slug}-#{Time.zone.today}.csv"
+  end
+
+  def mentor_volunteers_csv
+    send_data MentorVolunteersCsvBuilder.new(selected_campaign).generate_csv,
+              filename: "mentor_volunteers-#{selected_campaign.slug}-#{Time.zone.today}.csv"
+  end
+
   #################
   # WMF Analytics #
   #################
@@ -77,6 +90,10 @@ class AnalyticsController < ApplicationController
   end
 
   private
+
+  def selected_campaign
+    @selected_campaign ||= Campaign.find(params[:campaign][:id])
+  end
 
   # Mirrors #all_courses (JSON): admins see every course, including private;
   # non-admins see only nonprivate courses.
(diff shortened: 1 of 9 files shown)


ragesoss:
[Codex gpt-5.6-sol] Review findings:

- **[P1] Responses can be attributed to the wrong term/course.** `notification_courses` resolves every completed notification for every survey containing the question group. If a group belongs to Fall and Spring surveys, one matching Fall answer can also produce a Spring volunteer row—even if the Spring answer did not select mentoring. The answer must be correlated to one response/course context rather than expanded across all linked surveys. (`lib/analytics/mentor_volunteers_csv_builder.rb`, lines 55–57)

- **[P2] Volunteer rows discard course-specific real names.** `rows` calls `row(user, course)` without `courses_users.real_name`. Unlike the requests report, volunteers whose name exists only on their course enrollment get their username in `real_name`. (`lib/analytics/mentor_volunteers_csv_builder.rb`, line 16)

- **[P2] The report does not verify that volunteers are instructors.** `volunteer_pairs` accepts any user with matching answer text and an attributable campaign course. A student or other role taking an open/reused survey could therefore appear. The associated notification’s `courses_user.role`, or an equivalent enrollment lookup, should be checked. (`lib/analytics/mentor_volunteers_csv_builder.rb`, lines 20–22)

Validation: `git diff --check` and RuboCop passed, and GitHub CI is green. Local RSpec loading was blocked because MySQL was unavailable at `/run/mysqld/mysqld.sock`.

ragesoss:
None of those are worth fixing at the moment. We do one survey per term, so P1 is not expected to matter in practice. P2 is a low-stakes edge case. P3 is not an expected situation.

PR #7001, opened 2026-08-18
Add assume_ssl for Peony's proxy-terminated TLS

ragesoss:
## Summary

- Add `config.assume_ssl = true if ENV['ASSUME_SSL']` to production.rb and staging.rb, with an explanatory comment in production.rb
- Add `ASSUME_SSL: 'true'` to the P&E `application.yml` template

On 2026-08-17, the P&E Dashboard had a ~91-minute outage caused by a redirect loop: WMCloud's TLS-terminating proxy temporarily stopped sending `X-Forwarded-Proto: https`, so `config.force_ssl` saw every request as plain HTTP and redirected it — which the proxy forwarded right back, creating an infinite loop.

`config.assume_ssl` (Rails 7.1+) makes `request.ssl?` unconditionally true, eliminating the redirect-loop failure mode entirely. It's gated on `ENV['ASSUME_SSL']` so it only activates on Peony (behind the proxy); the Wiki Ed Dashboard terminates TLS directly and still needs `force_ssl` to redirect plain HTTP.

After merging and deploying, `ASSUME_SSL: 'true'` also needs to be added to the real `application.yml` on peony-web.

Relates to #6990.

## Test plan

- [ ] Deploy to Peony with `ASSUME_SSL: 'true'` in `application.yml`
- [ ] Verify `https://outreachdashboard.wmflabs.org/` loads normally
- [ ] Verify Wiki Ed production (no env var set) still redirects HTTP → HTTPS

### DIFF
--- a/config/application.programs-and-events.yml
+++ b/config/application.programs-and-events.yml
@@ -64,3 +64,8 @@
   update_debug_DailyUpdate: 'true'
 
   REDIS_URL: 'redis://p-and-e-dashboard-redis'
+
+# Peony sits behind WMCloud's TLS-terminating proxy. This tells Rails to
+# treat every request as SSL, avoiding redirect loops when the proxy
+# omits the X-Forwarded-Proto header.
+  ASSUME_SSL: 'true'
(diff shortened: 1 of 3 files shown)


ragesoss:
I've added the `ASSUME_SSL` to peony-web application.yml, so this should be all set upon next deploy.

Issue #700, opened 2016-02-28
selected state of wizard options is not accessible to screen readers


ragesoss:
It looks like the aria-selected property is the likely solution here: https://www.w3.org/TR/wai-aria/states_and_properties#aria-selected

PR #7002, opened 2026-08-18
Sentry unknown object response

Formasitchijoh:
## What this PR does

Fixes Sentry issue #700  which had logged 300+ events and was one of our largest active sources of Sentry noise. It also reduces volume on the related (`TypeError: Failed to fetch`, 10k+ events).

**Root cause:** across `api.js` and 17 other action/util files, our fetch error-handling threw the raw `Response` object instead of a real `Error` (`if (!response.ok) { ...; throw response; }`). A thrown `Response` has no message or stack trace, so it surfaces in Sentry as an opaque, undebuggable `[object Response]`. Worse, in `wikidata_actions.js`, `fetchWikidataLabels` had no `.catch()` at all on this rejection, so any failed Wikidata lookup (e.g. a 403 from rate limiting) became an unhandled promise rejection and crashed that request path outright.

**Fix:**
- Added `ApiError` (a real `Error` subclass carrying `status`/`statusText`/`url`/`responseText`) and an `ensureOk(response, prefix?)` helper to `utils/request.js`.
- Replaced every `if (!response.ok) { ...; throw response; }` call site with `ensureOk(response)`.
- Added the missing `.catch()` in `fetchWikidataLabels`, so a failed chunk now just skips those labels instead of crashing.
- Added a global `unhandledrejection` listener in `sentry.js` as a backstop — it wraps any non-`Error` rejection in a real `Error` before Sentry reports it, so this bug class can't recur silently elsewhere in the app.
- As a side effect of the `ensureOk` swap, found and fixed a pre-existing bug in `settings_actions.js`: `upgradeAdmin`/`downgradeAdmin`/`upgradeSpecialUser`/`downgradeSpecialUser` were missing a `return` on their follow-up promise.

Tests added: `request.spec.js`, `wikidata_actions.spec.js`, `sentry.spec.js`. Full suite (68 suites / 436 tests) green.

## AI usage

Used Claude  to investigate the Sentry issue (breadcrumbs, stack traces, live event sampling) and trace the root cause to specific files/lines in the real codebase; to design and implement the `ApiError`/`ensureOk` fix and apply it consistently across all 18 affected files; to write the new test coverage; and to independently review the resulting diff afterward for regressions (checked every remaining `response.ok` call site in the codebase, ran ESLint and the relevant Jest specs, and confirmed the fix actually addresses the confirmed root cause rather than just the symptom).

## Screenshots

Not applicable — this is an error-handling/logging fix with no UI change.

### DIFF
--- a/app/assets/javascripts/actions/campaign_actions.js
+++ b/app/assets/javascripts/actions/campaign_actions.js
@@ -1,16 +1,10 @@
 import { RECEIVE_COURSE_CAMPAIGNS, SORT_CAMPAIGNS_WITH_STATS, DELETE_CAMPAIGN, API_FAIL, RECEIVE_ALL_CAMPAIGNS, ADD_CAMPAIGN, RECEIVE_CAMPAIGNS_WITH_STATS, SET_FEATURED_CAMPAIGNS } from '../constants';
 import filterFeaturedCampaigns from '../utils/filter_featured_campaigns';
-import logErrorMessage from '../utils/log_error_message';
-import request from '../utils/request';
+import request, { ensureOk } from '../utils/request';
 
 const fetchCampaignsPromise = async (courseId) => {
   const response = await request(`/courses/${courseId}/campaigns.json`);
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
@@ -34,12 +28,7 @@ const removeCampaignsPromise = async (courseId, campaignId) => {
     method: 'DELETE',
     body: JSON.stringify({ campaign: { title: campaignId } })
   });
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
@@ -61,12 +50,7 @@ const addCampaignsPromise = async (courseId, campaignId) => {
     method: 'POST',
     body: JSON.stringify({ campaign: { title: campaignId } })
   });
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
@@ -85,12 +69,7 @@ export const addCampaign = (courseId, campaignId) => (dispatch) => {
 
 const fetchAllCampaignsPromise = async () => {
   const response = await request('/lookups/campaign.json');
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
@@ -110,12 +89,7 @@ export const fetchAllCampaigns = () => (dispatch) => {
 
 const fetchFeaturedCampaigns = async (dispatch) => {
   const response = await request('/campaigns/featured_campaigns');
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   const response_data = await response.json();
   const featured_campaigns = response_data.featured_campaigns;
   dispatch({ type: SET_FEATURED_CAMPAIGNS, data: response_data });
@@ -129,12 +103,7 @@ const fetchCampaignStatisticsPromise = async (userOnly, dispatch) => {
   // it is set to false if there are featured campaigns listed
   const newest = !(featured_campaigns.length > 0);
   const response = await request(`/campaigns/statistics.json?user_only=${userOnly}&newest=${newest}`);
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   const response_data = await response.json();
   const campaigns = filterFeaturedCampaigns(response_data, featured_campaigns);
   return { campaigns };
--- a/app/assets/javascripts/actions/category_actions.js
+++ b/app/assets/javascripts/actions/category_actions.js
@@ -1,16 +1,10 @@
 import { RECEIVE_CATEGORIES, ADD_CATEGORY, DELETE_CATEGORY, API_FAIL } from '../constants';
-import logErrorMessage from '../utils/log_error_message';
-import request from '../utils/request';
+import request, { ensureOk } from '../utils/request';
 import { stringify } from 'query-string';
 
 const fetchCategoriesPromise = async (courseSlug) => {
   const response = await request(`/courses/${courseSlug}/categories.json`);
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
@@ -32,12 +26,7 @@ const addCategoryPromise = async (payload) => {
     body: JSON.stringify(payload)
   });
 
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
@@ -63,12 +52,7 @@ const removeCategoryPromise = async (course_id, category_id) => {
     method: 'DELETE'
   });
 
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
--- a/app/assets/javascripts/actions/course_creation_actions.js
+++ b/app/assets/javascripts/actions/course_creation_actions.js
@@ -1,17 +1,11 @@
 import API from '../utils/api.js';
-import request from '../utils/request';
-import logErrorMessage from '../utils/log_error_message';
+import request, { ensureOk } from '../utils/request';
 
 import { RECEIVE_INITIAL_CAMPAIGN, CREATED_COURSE, RECEIVE_COURSE_CLONE, API_FAIL } from '../constants';
 
 const fetchCampaignPromise = async (slug) => {
   const response = await request(`/campaigns/${slug}.json`);
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
--- a/app/assets/javascripts/actions/new_account_actions.js
+++ b/app/assets/javascripts/actions/new_account_actions.js
@@ -1,17 +1,11 @@
 import * as types from '../constants';
-import logErrorMessage from '../utils/log_error_message';
 import API from '../utils/api.js';
-import request from '../utils/request';
+import request, { ensureOk } from '../utils/request';
 import { triggerNotificationsBellRefresh } from '../components/nav/notifications_bell';
 
 const _checkAvailability = async (newAccount) => {
   const response = await request(`https://meta.wikimedia.org/w/api.php?action=query&list=users&ususers=${newAccount.username}&usprop=cancreate&format=json&origin=*`);
-  if (!response.ok) {
-    logErrorMessage(response);
-    const data = await response.text();
-    response.responseText = data;
-    throw response;
-  }
+  await ensureOk(response);
   return response.json();
 };
 
--- a/app/assets/javascripts/actions/revisions_actions.js
+++ b/app/assets/javascripts/actions/revisions_actions.js
@@ -12,8 +12,7 @@ import {
   RECEIVE_REFERENCES_COURSE_SPECIFIC
 } from '../constants';
 import { fetchWikidataLabelsForRevisions } from './wikidata_actions';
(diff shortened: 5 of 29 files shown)


ragesoss (changes requested):
Review produced with Claude Code. AI-driven review feedback is not meant to be
automatically trusted or acted upon — it is worth reading and understanding, then
using or ignoring based on your own judgment. One finding below was verified by
running code locally; the rest come from reading the diff.

## The core fix looks right

The mechanical part of this PR holds up under checking:

- `throw response` is now completely absent from `app/assets/javascripts/` — a grep of
  the branch turns up zero remaining sites. Each of the 18 substitutions was read
  individually; all are faithful 1:1 replacements of the same 6-line boilerplate, and
  none silently gained or lost error handling in the swap.
- `ApiError` preserves exactly the fields the one real downstream consumer reads.
  `handleErrorNotification` (`app/assets/javascripts/reducers/notifications.js:17`)
  reads `responseText`, `responseJSON`, `errors`, `message`, and `statusText`; all
  still resolve as before.
- One benefit the description does not claim: because `ApiError` carries a real
  `.message`, the user-facing notification for a response with an empty `statusText`
  — which is every HTTP/2 response, so all of production — improves from a literal
  `{}` to `Request failed with status 500`.
- Full JS suite green (68 suites / 436 tests), ESLint clean on the changed files, and
  `yarn build` succeeds (3 warnings, all pre-existing `i18next/no-literal-string`).

The change requested below concerns the `sentry.js` addition, not the `ensureOk`
refactor.

## The unhandledrejection backstop appears to double Sentry volume rather than reduce it

`event.preventDefault()` cancels the browser's default action (console reporting of the
rejection). It does not stop other listeners from running. Sentry 7.120.4 installs its
own handler by assigning `window.onunhandledrejection`
(`node_modules/@sentry/utils/cjs/instrument/globalUnhandledRejection.js:25`), and that
handler never consults `event.defaultPrevented`.

The consequence is that both handlers fire, and a single non-`Error` rejection produces
two Sentry events instead of one:

```
"Unhandled rejection: Forbidden"                                     (mechanism: generic)
"Object captured as promise rejection with keys: status, statusText"  (mechanism: onunhandledrejection)
```

This was verified against the real SDK rather than inferred. The spec below was
generated by Claude Code during review, run locally, and passes on this branch:

```js
import '../app/assets/javascripts/sentry';

test('the listener adds a second Sentry event rather than replacing the opaque one', () => {
  const captured = [];

  window.Sentry.init({
    dsn: 'https://examplePublicKey@o0.ingest.sentry.io/0',
    autoSessionTracking: false,
    beforeSend: (event) => {
      captured.push(event.exception.values[0].value);
      return null; // don't actually send
    },
  });

  const event = new Event('unhandledrejection', { cancelable: true });
  event.reason = { status: 403, statusText: 'Forbidden' };

  // Sentry 7.x instruments window.onunhandledrejection (the property handler), not
  // addEventListener. A real browser fires both; jsdom's dispatchEvent only reaches
  // addEventListener listeners, so the property handler is invoked here too.
  window.dispatchEvent(event);
  window.onunhandledrejection(event);

  // preventDefault() suppressed the browser's console output...
  expect(event.defaultPrevented).toBe(true);
  // ...but Sentry's own handler ran anyway, so one rejection is now two events.
  expect(captured).toEqual([
    'Unhandled rejection: Forbidden',
    'Object captured as promise rejection with keys: status, statusText',
  ]);
});
```

`test/sentry.spec.js` does not surface this because it never calls `Sentry.init()`, so
Sentry's own handler is never installed in the test environment.

Since reducing Sentry noise is the purpose of this PR, a backstop that adds a second
event per rejection works against it. Possible directions, in rough order of
preference:

1. A `beforeSend` hook in `Sentry.init()` that normalizes non-`Error` exceptions on the
   event Sentry already builds. One hook, no parallel listener, no duplication.
2. Disable `onunhandledrejection` in the `GlobalHandlers` integration so the new
   listener is the only reporter.
3. Drop the listener. The `ensureOk` refactor already removes the source of these
   rejections, so the backstop may not be earning its complexity.

A secondary effect worth weighing regardless of direction: `event.preventDefault()`
also silences unhandled rejections in the browser console anywhere the sentry bundle
loads (`Features.sentry?`, so production and the P&E Dashboard). The `console.error`
fallback only runs when `Sentry` is undefined, which is the case in development where
the bundle is not loaded at all.

## Two remaining sites of the same bug class

The description states this bug class "can't recur silently elsewhere in the app." Two
sites appear to have been missed:

- `app/assets/javascripts/utils/onboarding_utils.js:12` and `:25` — `throw data;` where
  `data` is the response text **string**. A thrown string is also a non-`Error`
  rejection and produces the same opaque Sentry report. Same directory as `request.js`.
- `app/assets/javascripts/components/common/ArticleViewer/authorship/AuthorshipAPI.js:21`
  — `reject(response)` rejects with a raw `Response`. This one also has no `.catch()` on
  the `fetch`, so a network failure leaves the outer promise permanently pending rather
  than rejecting; that may be worth a look as a separate contributor to the
  `TypeError: Failed to fetch` volume the description mentions.

## Minor observations

None of these are blocking.

- `wikidata_actions.js` now logs each failure twice: `ensureOk` calls
  `logErrorMessage(response)`, then the new `.catch()` calls `logErrorMessage(error)`.
  Visible in the output of the new spec.
- `wikidata_actions.js:17` uses `await ensureOk(await request(...))`, while the other 17
  call sites use `await ensureOk(response)` on its own line. Style only.
- `article_finder_utils.js:15` — this `ensureOk` call is unreachable. `fetch-jsonp`
  resolves only with `{ ok: true, status: 200, json }` and its response object has no
  `.text()` method, so if the branch were reached `ensureOk` would throw a `TypeError`
  rather than an `ApiError`. Pre-existing, not introduced here.
- Sentry grouping: `ApiError`'s message is the bare `statusText`, so 500s from every
  endpoint will group into a single issue. `url` is on the error object but not in the
  message; folding the path into the message or setting a fingerprint would separate
  them if per-endpoint grouping is wanted.
- `ensureOk` introduces a `logErrorMessage` call where there previously was none in
  `api.js`'s `saveTimeline`/`saveCourse` and `mediawiki_revisions_utils.js`'s
  `fetchAll`. Harmless, though `fetchAll` runs in a continuation loop and could get
  chatty.

## Test coverage

| Area | Status |
|---|---|
| `ensureOk` / `ApiError` | Well covered — 5 cases including the `.text()`-rejects fallback |
| `fetchWikidataLabels`, both paths | Well covered |
| The 17 mechanical call sites | Uncovered, but low risk; every hunk was read |
| `settings_actions.js` `return` fix | **Uncovered.** Reverting only that file to master leaves `test/actions/settings_actions.spec.js` passing, so nothing currently discriminates. It is a real behavior change to four thunks — the returned promise now resolves after the list refresh rather than before. |
| `api.js` `saveTimeline`/`saveCourse` try/catch restructure | Uncovered; SentryLogger path with a silent failure mode |
| `sentry.js` listener | Covered, but the tests do not exercise the interaction with an initialized Sentry client |

*Drafted in a Claude Code session (~20 min, 3 interactions). Sage initiated the review
and chose to request changes on the Sentry finding, but did not independently verify any
of the findings beyond reading this summary. Verification performed by Claude Code within
the session: full Jest suite, ESLint on all changed files, and `yarn build` were run on
the branch merged with master; the double-reporting claim was verified by running the
spec above against the real `@sentry/browser` 7.120.4 SDK; the `settings_actions.js`
coverage claim was verified by reverting that file to master and rerunning its spec. The
inline spec is AI-generated code — it was run locally but has not been reviewed by Sage.*

ragesoss (approved):
Re-review produced with Claude Code. As before: AI-driven review feedback is not meant to
be automatically trusted or acted upon — it is worth reading and understanding, then using
or ignoring based on your own judgment.

## The blocking finding is fixed, and the fix was verified rather than assumed

`beforeSend` was the right call. The double-reporting is gone: a spec built the same way as
the one in the previous review — dispatching *both* delivery paths, `window.dispatchEvent`
and the `window.onunhandledrejection` property handler — now yields one event where it
previously yielded two, and the message is the readable one. Real `Error` rejections pass
through untouched, and an `ApiError` keeps its own message rather than being rewritten.

The SDK internals the hook depends on were checked rather than taken on trust:
`hint.originalException` is the rejection reason
(`@sentry/browser/cjs/integrations/globalhandlers.js:100-104`), and `hint.mechanism` is
applied to `event.exception.values[0]` in `@sentry/core/cjs/utils/prepareEvent.js:61`,
which runs before `processBeforeSend` in `baseclient.js:640`. So both the
`mechanism.type === 'onunhandledrejection'` guard and the `hint` read are reliable, not
incidental.

Dropping `preventDefault()` also resolves the secondary effect noted last time — unhandled
rejections are no longer silenced in the browser console wherever the sentry bundle loads.

The wiring looks complete. There is exactly one `Sentry.init` call site, and
`hot_javascript_tag` emits a plain synchronous `javascript_include_tag`
(`app/helpers/application_helper.rb:26-33`), so `window.normalizeUnhandledRejection` is
always defined before `init` runs.

Everything else raised last time also checks out: both missed rejection sites are fixed
(`AuthorshipAPI` no longer leaves its promise permanently pending on a network-level
failure, and both cases now have specs), the wikidata double-log and the nested-`await`
style are cleaned up, and the three coverage gaps are filled. The `settings_actions.js`
gap was verified closed the same way it was verified open: reverting only the four `return`
keywords, leaving `ensureOk` in place, now fails 4 of the 5 tests in that file. Worth
noting why that changed — on master the missing `return` was unobservable, and it is the
extra microtask that `await ensureOk(...)` introduces on the success path that makes it
observable. The `return` addition was load-bearing for the refactor, not incidental
cleanup.

## One item that may warrant a decision before merge

`AuthorshipAPI` switching from `reject(response)` to an `ApiError` changes a string users
see. `useAuthorshipHighlighting.js:134` does `setFailureMessage(error.message)`, and
`ArticleViewerLegend.jsx:136` renders it after `users.authorship_data_not_fetched`.
Previously `error.message` was `undefined`, so the legend read "Authorship data not
fetched: " with nothing after it. Now it reads "…: Not Found" — or, since `statusText` is
empty on every HTTP/2 response, "…: Request failed with status 404", which is the case that
will actually show in production.

That is arguably an improvement over a dangling colon, but `Request failed with status N`
is AI-authored copy now surfacing in the interface, which under this repo's conventions is
the operator's call rather than something to settle in review.

## Two observations, neither blocking

**The description's claim that this bug class "can't recur silently elsewhere in the app"
is still overstated.** A wider grep than the previous review's turns up roughly a dozen
more sites of the same shape — `Promise.reject(res)` or `throw res` with a raw `Response`:
`articles_actions.js:13`, `assignment_actions.js:17,100,122`, `alert_actions.js:57`,
`course_alert_actions.js:11`, `campaign_view_actions.js:14`,
`uploads_actions.js:55,86,133`, `tagged_courses_stats_action.js:11`, and
`AddToWatchlistButton.jsx:27`. Each was checked, and every one has a local `.catch()`, so
none produces an unhandled rejection and none contributes to the Sentry issue this PR
targets. Nothing here needs to change for this PR to land. It may be worth a follow-up on
its own merits, though: several of those catches return the raw `Response` as a resolved
value, and `articles_actions.js` then dispatches it as the `RECEIVE_ARTICLES` payload.

**`Sentry.captureException` in `wikidata_actions.js:51` is volume-neutral rather than a
reduction.** Before this PR those failures were already reaching Sentry as unhandled
rejections; they now arrive as explicit captures with a real message. Given the Wikidata
403s that have been running in production since around April, this path will keep producing
events at the same rate — better grouped and legible, but not fewer. That seems like the
right trade, and is only worth stating because volume reduction is the PR's stated goal.
It is also the first `Sentry.captureException` in an action file, so it establishes a
pattern.

## Coverage

The remaining untested surface is the `_head.html.haml` wiring itself: `Features.sentry?`
is false in the test environment, so no Ruby spec renders that block, and nothing would
catch it if `beforeSend` were wired to an undefined function. Load order was confirmed by
reading the helper rather than by exercising it. The 17 mechanical `ensureOk` call sites
also remain uncovered, as before — low risk, and each was read individually during the
first review.

Verification performed in this session: full JS suite (72 suites / 460 tests, all
passing; the count is higher than the 436 quoted last time because master has since
added specs of its own), ESLint clean on all 28 changed JS files, and `yarn build` succeeding with only
the three pre-existing `i18next/no-literal-string` warnings, all on the branch merged with
master.

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage asked for the re-review
and approved posting after reading the summary, but did not independently verify any of the
findings. Verification performed by Claude Code within the session: the full Jest suite,
ESLint, and `yarn build`; the single-event claim was checked with an AI-written spec run
against the real `@sentry/browser` 7.120.4 SDK and then discarded; the `settings_actions.js`
coverage claim was checked by reverting only the four `return` keywords and rerunning that
spec; the remaining rejection sites were each opened and checked for a local `.catch()`.
The verification spec was AI-generated and was not reviewed by Sage.*

ragesoss:
This looks good to me. Thanks @Formasitchijoh!

Formasitchijoh:
You are welcome @ragesoss. Please could you inform me when this changes are in production. There is an ongoing issue that is currently being logged on sentry but there are no detailed logged as to what is currently happening.I will be able to see the actual error once this fix is deployed.

ragesoss:
@Formasitchijoh in Peony production specifically, or both production instances?

Formasitchijoh:
@ragesoss  On Peony production

ragesoss:
cool. it'll probably deploy once https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/7006 is ready.

PR #7006, opened 2026-08-20
Add ReportCsvStore so report CSVs can be kept in object storage

### LINKED ISSUE #6606: Move CSV generation Sidekiq jobs off webserver
CSV generation can for report downloads can be resource-intensive (perhaps with a memory leak involved), so a large wave of CSV requests can bring down the Programs & Events Dashboard webserver.

We currently run those jobs on the web-server so that the files live on that server and can be served from there. To move them off the webserver, we'll need to implement a shared storage volume on Wikimedia Cloud, so that a sidekiq server can write the CSVs and the webserver can read them to serve the downloads.

It also needs to be implemented in a way that won't cause any trouble for Wiki Education Dashboard, where all the processes will continue to share a single server and a single storage volume.

gabina:
## What this PR does

Adds a `ReportCsvStore` abstraction so that report CSV exports can be kept either on the
local filesystem or in S3-compatible object storage, chosen per deployment by
configuration. This is the application-side prerequisite for
[#6606](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6606): today
`ReportCsvWorker` writes reports into `public/system/analytics` and Apache serves them from
there, which forces the CSV jobs to run on the web server, where they compete with the web
application for memory (a contributing factor in
[#6990](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/6990)).

`ReportCsvStore` exposes `exists?`, `write`, `url_for` and `delete`, and delegates to one of
two backends:

- `LocalReportCsvStore` — writes under `public/system/analytics` and returns a path served
  statically by the web server. This is the default and is what Wiki Education Dashboard,
  development and test all use.
- `S3ReportCsvStore` — puts the report in an object storage bucket and returns the object's
  public url, so users are redirected straight to it and the web server serves no bytes.

The backend is selected by whether `report_csv_bucket` is configured, so behaviour is
unchanged wherever it is unset. The five new settings are documented (commented out) in
`config/application.example.yml`. Programs & Events Dashboard will set them to a
[Wikimedia Cloud VPS object storage](https://wikitech.wikimedia.org/wiki/Help:Object_storage_user_guide)
container, which needs a bucket policy granting anonymous `s3:GetObject` for the redirect
to work.

The specs pin the suite to the local store (`spec/rails_helper.rb` deletes
`report_csv_bucket` from `ENV`, which figaro would otherwise populate from a developer's
`application.yml`), so the suite never depends on local configuration or reaches the
network. The object storage behaviour is covered by unit specs with a stubbed
`Aws::S3::Client`, plus one request spec that stubs the store to check the off-host
redirect.

## AI usage

Claude Code was used heavily throughout, in two distinct ways.

First, as a research and design partner: the user worked through the shared-storage options
for #6606 with it — NFS (`/data/project` vs `/data/scratch`), a Cinder volume attached to
two hosts, using Toolforge as a file host, pushing files to the web server over SSH, and
finally object storage — reading the relevant wikitech documentation and discarding each
option for concrete reasons. Some of Claude's early conclusions were wrong and were
corrected by the documentation or by testing, notably a claim that object storage was not
available for Cloud VPS, and a suggestion that Swift's `.rlistings` ACL could hide the
container listing (Ceph's RADOS gateway does not honour it, as a manual test showed).

Second, as an implementation assistant: Claude Code drafted `ReportCsvStore` and its two
backends, the three new spec files, the edits to `ReportsController`, `ReportCsvWorker` and
`CsvCleanupWorker`, and the `rails_helper` change. The user set the design — two backends
with class-method entry points, following the shape of `lib/originality_api.rb` — created
the object storage container and credentials by hand, verified the upload and anonymous
download path with `curl` and the AWS CLI, ran the specs, fixed the failures, reviewed
everything and wrote the commits.

The commits themselves were authored by the user, without AI-generated commit messages: six
commits within about three hours on 2026-08-20, following several days of intermittent
design discussion and manual infrastructure work before any code was written.

This description — including the summary above — was drafted by Claude Code using the
`/prepare-pr` skill, and may contain errors.

### System stats CSV downloaded from object storage
[Screencast from 2026-08-20 15-25-11.webm](https://github.com/user-attachments/assets/d359c49b-d4df-4ca8-b7fd-8a887462defa)

### Course overview CSV downloaded from object storage
[Screencast from 2026-08-20 15-30-42.webm](https://github.com/user-attachments/assets/0bd730e2-4476-4903-a5ea-975cb938e05d)

- **Production credentials are ready.** Already generated EC2 keys bound to new service
  developer account using the new peony email and asked for the `object_storage` role
  ([T435414](https://phabricator.wikimedia.org/T435414)).
- **This PR does not close #6606 on its own.** `ReportCsvWorker` still runs on the
  `daily_update` queue. Actually moving that queue off the web server is a deployment
  change, and `docs/WMCLOUD_SETUP/` still needs notes on the container, its bucket policy
  and the fact that listing suppression has to be done through S3 rather than Swift.
- **Cutover leaves files behind.** When P&E switches to object storage, the CSVs already in
  `shared/public/system/analytics` become invisible to the app (users simply regenerate
  them) and the directory needs sweeping manually once.
- **Worth one check after deploy:** objects uploaded with the AWS CLI came back carrying
  `content-encoding: aws-chunked`, which is an artifact of chunked request signing rather
  than a real encoding. The Ruby SDK's `put_object` with a string body is not expected to
  set it, but a browser download of a worker-written report is worth confirming.
- `ReportCsvStore` memoizes the selected backend, so changing `report_csv_bucket` requires
  a process restart. The spec that exercises backend selection resets that memo directly.


gabina:
Things to do **before** deployment:
Add production credentials as env variables to peony server.

Things to do right **after** deployment:
Manually clean reports from `public/system/analytics` in peony server.

ragesoss:
I think you have access to add the credentials?

gabina:
Yup, think I have access for doing both things. Just taking note so I wouldn't forget.

ragesoss (changes requested):
This review was produced in a Claude Code session. AI-generated review feedback should not
be trusted automatically or acted on blindly — it is worth reading and understanding, then
using or ignoring based on your own judgment. Some of what follows may be wrong.

The abstraction reads cleanly and the local path is genuinely unchanged, which makes this
safe to ship for Wiki Ed regardless of what happens with the object storage side. RuboCop is
clean on all 11 changed Ruby files, and the 69 examples covering the affected area (the three
new store specs plus `reports_controller_spec`, `system_csv_spec` and `system_daily_stats_csv_spec`)
pass against a merge with current master.

Changes are requested for one item — the first finding below — because it would break report
generation on Programs & Events in a way that would not show up until a user hit it.

## 1. `exists?` rescues only 404, and a 403 is likely

`S3ReportCsvStore#exists?` rescues `Aws::S3::Errors::NotFound`. S3-compatible gateways answer
**403, not 404**, for a HEAD on a key that does not exist when the caller lacks list permission
on the bucket — this is deliberate, so that object existence is not leaked. `Aws::S3::Errors::Forbidden`
is not a subclass of `NotFound`, so it escapes the rescue.

This was checked locally against a real `Aws::S3::Client` (send step stubbed to return a given
status with an empty body, which is the shape of a HEAD response):

| HEAD status | error class raised | caught by `rescue NotFound`? |
|---|---|---|
| 404 | `Aws::S3::Errors::NotFound` | yes → returns `false` |
| 403 | `Aws::S3::Errors::Forbidden` | no |
| 500 | `Aws::S3::Errors::Http500Error` | no |

The user-facing effect was also reproduced with a request spec: with `ReportCsvStore.exists?`
raising `Forbidden`, `GET /course_csv` returns **500 and never enqueues the job**, where the
local store renders "This file is being generated." Because `exists?` is false on every first
request for a report, this is not an edge case — if the production credentials do not have list
permission on the container, report generation is broken for every report.

Whether it fires depends on the container's permissions. If the EC2 keys own the container they
will have implicit full access and 404 is what comes back. Two ways to settle it:

- Confirm empirically: `head_object` against a key that has never existed, using the production
  credentials against the real container, and check that it 404s rather than 403s.
- Or make it moot in code: `rescue Aws::S3::Errors::NotFound, Aws::S3::Errors::Forbidden`. This
  costs the ability to distinguish "missing" from "misconfigured", so it is a trade-off rather
  than a strict improvement — but it fails toward regenerating the report rather than toward a
  500.

Either resolves this; the empirical check alone is enough if it comes back 404.

## 2. The request-path S3 client has no timeouts or retry ceiling

`Aws::S3::Client` defaults, confirmed against the version this PR locks (`aws-sdk-s3` 1.229.0 /
`aws-sdk-core` 3.254.1): `http_open_timeout: 15`, `http_read_timeout: 60`, `retry_limit: 3`,
legacy retry mode. `exists?` now runs inside the web request, so a slow or unreachable gateway
can hold a Puma thread for minutes, where `File.exist?` previously could not fail or block at
all. Given that the motivation here is reducing resource pressure on the web server (#6990),
tightening these on the client — something like `http_open_timeout: 5, http_read_timeout: 10,
retry_limit: 1` — seems worth doing at the same time.

## 3. The `aws-chunked` open question: not an issue on the worker path

Raised in the PR description as worth checking after deploy. It was checked here instead, by
capturing the HTTP request the SDK actually builds. `put_object` with a String body sends the
body **verbatim**, sets `content-type: text/csv`, sends the CRC32 as a plain
`x-amz-checksum-crc32` header, and sets **no** `content-encoding`. The `aws-chunked` value seen
from the AWS CLI is an artifact of chunked request signing over a file body and does not apply
here.

One related note: `request_checksum_calculation` defaults to `when_supported` in this SDK
version, so `x-amz-checksum-crc32` and `x-amz-sdk-checksum-algorithm: CRC32` are sent on every
upload. Some older Ceph RGW builds reject those headers. The CLI upload already exercised this
path against the real gateway and succeeded, so it is presumably fine — but if `put_object`
turns out to fail in production with a checksum or signature error, `request_checksum_calculation:
'when_required'` on the client is the fix.

## 4. Anonymous listing is the security-relevant thing to verify

The design requires anonymous `s3:GetObject`. This is not a regression: `public/system/analytics`
is already served by Apache without authentication, so report CSVs are already retrievable by
anyone who knows a filename. What changes is enumerability. Report filenames are guessable
(`<slug>-course_overview-<date>.csv`) and the contents include student usernames and course
data, so if anonymous listing is not denied, "guessable" becomes "enumerable across every report
in the retention window."

The PR description already notes that Ceph ignores Swift's `.rlistings` and that suppression has
to go through an S3 bucket policy, so this is likely handled — worth confirming that an anonymous
`GET /<bucket>?list-type=2` returns 403, and capturing that in the `docs/WMCLOUD_SETUP/` notes
that are already flagged as outstanding.

## 5. `Content-Disposition` is not set, and `download` does not work cross-origin

`triggerDownload` in `system_csv_export_bar.jsx:14` builds an `<a download>` to save the file.
Browsers honour the `download` attribute only for same-origin URLs, so against object storage it
degrades to a plain navigation. It still works because `text/csv` is not rendered inline — which
is what the screencast shows — but the behaviour depends on browser MIME handling rather than on
anything explicit. Passing `content_disposition: 'attachment'` to `put_object` would make it
deterministic, and would also cover the `redirect_to` endpoints.

## 6. Smaller notes

- **`Gemfile.lock` ordering.** The `aws-*` block landed between `ast` and `auth-sanitizer`;
  bundler sorts `auth-sanitizer` first. A plain `bundle install` on this branch rewrites it.
  Committing that keeps the reordering from riding along with an unrelated future bundle change.
- **`ReportCsvStore.local` / `.s3`** are public class methods used only by the private `store`;
  they could be `private_class_method` as well.
- **Delete semantics now differ by backend.** Local `File.delete` raises `Errno::ENOENT` when the
  file is already gone, so `CsvCleanupWorker` retries; S3 `delete_object` is idempotent. This is
  pre-existing behaviour for the local path, but the new abstraction is the natural place to make
  the two consistent.
- **`CsvCleanupWorker#perform` has no spec** — it is mocked everywhere it appears, so
  `ReportCsvStore.delete` is the one operation with no coverage on either backend.

## 7. Out of scope, but adjacent to #6606

`lib/analytics/histogram_plotter.rb` is a second local-disk CSV cache
(`public/assets/system/analytics`), still live: `article_list.jsx:318` renders `CourseOresPlot`
on every course's Articles tab, which reads it through `OresPlotController`, and
`UpdateCourseStats` (`app/services/update_course_stats.rb:92`) deletes it after every course
update to invalidate it.

On Programs & Events those two happen on different hosts — the read on `peony-web`, the delete
on the sidekiq hosts — so the invalidation is a silent no-op against the file the web host is
actually serving. The impact is bounded, so this is a note rather than a problem: `csv_filename`
embeds `Time.zone.today`, so the cache rotates daily on its own, and the explicit refresh button
works because `refresh_ores_data` deletes and rebuilds within the same web request.
`delete_csv` uses `FileUtils.rm_rf`, so the no-op does not raise inside the surrounding
transaction either.

Entirely independent of this PR and not something it should grow to cover — but it is the same
shared-storage problem in a second place, so worth knowing before #6606 is called done.

---

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage directed the review and chose
to request changes after reading the findings summary, but did not independently re-run the
checks. Verified locally by Claude Code, not by Sage: RuboCop on the 11 changed Ruby files; the
69 affected examples against a merge with master; the HEAD status/error-class mapping in finding
1, via a real `Aws::S3::Client`; the 500-instead-of-enqueue effect, via a throwaway request spec;
and the `put_object` request shape in finding 3, by capturing the built HTTP request. Finding 7's
reachability and no-op claims were traced through the routes, components and specs. Findings 4 and
5 are reasoning from the code and the linked documentation, and were not tested against the
production container.*

gabina:
@ragesoss I addressed comment **1. exists? rescues only 404, and a 403 is likely** in 9e6d3123d41592ab87eceb832d4d0a29b8961a61 and added a doc in 71de64861037024f2bb992268a7270fcab2c4653.

How would you like to proceed? One option could be to first set the credentials on the peony-web server and test the changes there, and then decide which server to move the queue to. Another option would be to set the credentials directly on the new server (for example, peony-sidekiq-medium) and test the changes directly on the new server.

ragesoss:
Either way is fine with me.

ragesoss (approved):
Follow-up on the requested changes. As before: AI-driven review feedback (this comment included) is worth reading and understanding, then using or ignoring based on your own judgment — some of it may be wrong.

The blocking finding is resolved. 9e6d312 rescues `Aws::S3::Errors::ServiceError`, reports to Sentry, and returns `false`. This was re-verified by re-running the original reproduction: with `head_object` answering 403 (`Aws::S3::Errors::Forbidden`) at the client boundary, `GET /course_csv` now enqueues the report and renders "This file is being generated" with a 200, and the error reaches Sentry — where it previously returned 500 without enqueuing. The rescue ordering is correct (`NotFound` is itself a `ServiceError` subclass, caught first and kept silent), and the fix lands better than either option the review offered, since it fails toward regenerating the report while keeping misconfiguration visible. One residual worth knowing about: `Seahorse::Client::NetworkingError` is not a `ServiceError`, so an unreachable or hanging gateway still surfaces as a 500 after the SDK's default timeouts (15s open / 60s read, 3 retries) — the same territory as the earlier client-timeout suggestion, which remains open. Relatedly, under a persistently erroring gateway, polling will not stack duplicate jobs (`lock: :until_executed` covers that); the visible symptom would be one Sentry event per poll.

The new `docs/WMCLOUD_SETUP/object_storage.md` captures the container policy requirement (anonymous GetObject, no ListBucket). Two things it does not yet record: that listing suppression has to be applied through the S3 API (bucket policy) rather than Swift ACLs, and whether an anonymous `GET /<bucket>?list-type=2` was empirically confirmed to return 403 against the real container. A line for each would close out that item, if they were done.

The remaining non-blocking notes from the prior review are unchanged and fine to defer or drop: request-path client timeouts, `content_disposition: 'attachment'` on `put_object`, Gemfile.lock ordering (a plain `bundle check` on this branch still rewrites the lock — re-confirmed here), `private_class_method` for `.local`/`.s3`, the delete-semantics divergence between backends, and spec coverage for `CsvCleanupWorker#perform`.

Checked for this re-review: RuboCop clean on the changed files, the 69 affected examples pass against a merge with current master, and `yarn build` succeeds.

*Drafted in a Claude Code session (~15 min, 2 interactions). Sage initiated the re-review and approved this comment before posting. The verifications above — the error-class hierarchy against the locked SDK, the 403 reproduction via a throwaway request spec written by Claude Code, the spec/linter/build runs, and the Gemfile.lock re-check — were performed by Claude Code and not independently re-run by Sage.*

Issue #7005, opened 2026-08-20
Updates re-scan every empty timeslice between a course's last student edit and today


ragesoss:
## Summary

For a course whose students have stopped editing but whose end date is still in the future, every course update re-fetches revision data for **every daily timeslice between the last tracked student edit and today** — one replica-revision-tools query per wiki per timeslice — even though all of those timeslices are empty and were already scanned by every previous update. The scan window never shrinks, and it grows by one timeslice per tracked wiki per day until the course end date. This is steady-state waste: the update completes "successfully" every cycle while doing thousands of no-op fetches.

## Observed case

P&E Dashboard course 10781, [`tlv_university/TLV_University,_Science_Oriented_Youth_-_Alpha_Program_-_Cycle_C_(2016-2017)`](https://outreachdashboard.wmflabs.org/courses/tlv_university/TLV_University,_Science_Oriented_Youth_-_Alpha_Program_-_Cycle_C_(2016-2017)):

- 50 students, 573 lifetime edits; tracked wikis: he.wikipedia + wikidata
- Course window 2016-11-17 → **2030-12-31**
- Last tracked student edit: he.wikipedia **2025-01-24**, wikidata **2017-08-14**
- Every update (per `flags.update_logs`, 2026-08-19/20): `processed` ≈ 3,865–3,867, duration 22.4–23.1 min, ~10 updates/day (`average_delay` ≈ 8,700 s), `reprocessed=0`, `error_count=0`

The arithmetic closes exactly: 573 daily slices (he.wikipedia, 2025-01-24 → today) + 3,294 (wikidata, 2017-08-14 → today) = **3,867**, matching the logged `processed` count, and the count grows by +2/day at the course-start-aligned 22:00 UTC slice boundary. Measured per-empty-slice cost: ~0.25–0.30 s per `revisions.php` query plus ~35% Rails-side overhead ≈ 0.35 s/slice → the observed ~22.7 min per update.

Net effect for this one dormant course: **~3.7 hours of sidekiq worker time and ~38,000 replica-revision-tools queries per day**, continuing (and slowly growing) until 2031.

## Mechanism

- `TimesliceManager#get_ingestion_start_time_for_wiki` returns the start of the timeslice with the max `last_mw_rev_datetime` — i.e., the slice containing the latest tracked student revision.
[...]

PR #7007, opened 2026-08-20
Skip empty timeslices via wide-window precheck during course updates

ragesoss:
## What this PR does

Addresses #7005.

For a dormant course with a far-future end date, every course update re-fetched every daily timeslice between the last tracked student edit and today — one replica-revision-tools query per wiki-timeslice — because empty timeslices never advance the ingestion watermark (`TimesliceManager#get_ingestion_start_time_for_wiki`). On P&E course 10781 that meant 3,867 no-op queries and ~22.7 minutes per update, roughly 10 times a day, growing by 2 timeslices per day until the course ends in 2030 (~3.7 hours of Sidekiq worker time and ~38,000 Toolforge queries daily for a course with no tracked activity since January 2025).

The fix: when the range of timeslices to process exceeds `GAP_PRECHECK_THRESHOLD` (7), a new `FindTimeslicesWithRevisions` service makes ~~**one** wide-window `revisions.php` query~~ **one wide-window `revisions.php` query per chunk of ten students (see Review updates below)** over the whole range and the update processes only the timeslices that actually contain revisions — plus, always, the first (watermark) timeslice, which is the only one in range that can already have recorded revisions and must be re-fetched to preserve deletion detection (the `mw_rev_count` comparison in `CourseRevisionUpdater#new_revisions?`) and partial re-ingestion. Skipping the rest is safe by construction: for an empty timeslice in `only_new` mode, `should_update_timeslice?` is false and nothing is persisted, so processing it was a pure no-op.

Design properties, per the priorities discussed for #7005 (simplicity first, then efficiency):

- **No schema change and no new persistent state.** The gap is still re-checked on every update — just with ~~one query~~ **a handful of queries** instead of thousands — so correctness is identical to the previous behavior.
- **Failure falls back to the old path.** The precheck uses `Replica#get_revisions_raw`, which distinguishes a failed query (`nil`) from a genuinely empty result (`[]`); on failure the full range is processed exactly as before. (`revisions.php` was also checked for server-side row limits — there are none, so a truncated wide result can't silently mark timeslices empty.) **Review update: that check didn't cover the endpoint's *memory* limit, which is the reason `RevisionDataManager` chunks its user list at ten. The precheck now chunks the same way, and treats a failure in any chunk as a failure of the whole precheck.**
- **Active courses are untouched.** Their scan ranges are short (the watermark is at their most recent edit), so they stay below the threshold and gain zero added queries.
- **The reprocess loop (`needs_update` timeslices) is deliberately untouched.** A flagged timeslice can carry stale data that reprocessing must clear, so "no revisions on wiki" does not imply no-op there. ~~Consequence: a flagged full update of a course with *no timeslices yet* (`recreate_timeslices` marks everything `needs_update`) does not benefit, while scheduled first updates and full updates of courses with existing timeslices do.~~ **Review update: this was wrong, and there is no uncovered case. `fetch_data_and_reprocess_timeslices` only reprocesses flagged timeslices whose `start` is at or before the ingestion start, while the precheck's range *begins* at the ingestion start — so the two sets overlap in at most the watermark timeslice, which the precheck keeps unconditionally anyway. A flagged full update of a brand-new course benefits like any other: `recreate_timeslices` flags everything, but the reprocess loop then clears the flag on every timeslice after the course start without reprocessing it, leaving the whole range to the precheck. Measured on a 30-timeslice course with everything freshly flagged: `processed=29` with the precheck disabled versus `processed=0` with it active.**

Expected impact on course 10781: each update drops from 3,867 fetches to ~~2 wide queries~~ **2 wide queries per chunk of ten students** + 2 watermark fetches — from ~22.7 minutes to a few seconds.

Validation beyond the specs: the service was smoke-tested against the live replica endpoint using a real course from a production data dump (110 timeslices): zero false negatives versus the recorded per-timeslice state, with the only extra selections being course-start timeslices containing system edits, which fetch and no-op harmlessly. The cost model and slice arithmetic were validated against course 10781's live update logs (predicted `processed=3867` exactly).

## Review updates

A Claude Code review session on this branch ([notes](https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/7007#issuecomment-5360386308)) found two ways the precheck added Replica load instead of removing it, plus the spec gap that hid both. Struck-through text above marks what changed. A follow-up commit on this branch addresses:

- **Courses where importing revisions is pointless paid for a query they previously didn't.** `CourseRevisionUpdater#no_point_in_importing_revisions?` already short-circuits every per-timeslice fetch for a course with no students, and for an `only_scoped_articles_course?` with no assignments and no categories — but the precheck had no equivalent guard, and `Course.ready_for_update` picks up every current course regardless of student count. The precheck now returns an empty set without querying in that case, which also skips the whole no-op range, so those courses come out ahead of pre-PR master rather than behind it.
- **The wide query bypassed the `MAX_USERNAMES` chunking** that exists to stay inside the Replica endpoint's memory limit, amplified by the widest window in the update. It now chunks at ten students, and collapses to the full-range fallback if any chunk fails. `Utils.chunk_requests` is deliberately not used: its `Array(info)` maps a failed chunk's `nil` to `[]`, which is indistinguishable from a genuinely empty result.
- **Both new spec files exercised the precheck on a course with zero students.** `JoinCourse` fails `student_joining_before_approval?` silently when the course has no campaign, and the `empty-gap precheck` block is a sibling of `describe 'UpdateCourseWikiTimeslices.run'` rather than nested inside it, so it never inherited that block's campaign setup. Every "fetches N timeslices" assertion was therefore measured on a course that in production issues no Replica query at all. Both setups now add a campaign, and that is what makes the two items above observable.
- Minor: `precheck_nonempty_timeslices` was issuing a `SELECT COUNT(*)` and then re-querying the same rows.

## AI usage

This PR was developed with Claude Code, which wrote all of the code, specs, and commit message under human direction. The investigation that produced #7005 (public-API recon against the live dashboard and replica endpoints, code reading, and the cost-model arithmetic), the implementation plan, and this implementation were all Claude Code sessions; Sage set the priorities (simple/understandable over maximally efficient), approved the plan's central design choice (a stateless wide-query precheck rather than a persisted scan watermark), and reviewed the work. The single implementation commit was made in one focused session on 2026-08-20, following a planning discussion in the same session. The commit's `## Process` section has a fuller account, including the one test misstep (a `stub_const` ordering issue in a new spec, fixed by removing the stub). This summary and the rest of this description were also drafted by Claude Code and may contain errors.

## Screenshots

No UI changes.

- `GAP_PRECHECK_THRESHOLD = 7` is a judgment call; anything from ~3 to ~30 is defensible. Existing specs' 7-slice courses sit exactly at the threshold, keeping them on the unmodified path.
- `processed` in update logs now counts only actually-fetched timeslices (it will drop from ~3,867 to ~2 for dormant courses), which makes the logs more meaningful but changes their scale for such courses.
- ~~The uncovered case — flagged full update of a course with no timeslices yet — could be closed by letting the reprocess loop skip provably-fresh empty timeslices, but that adds a multi-column "provably fresh" guard plus flag-clearing logic; it fires at most once per copied/new course and seemed not worth the added invariants. Can revisit if it turns out to matter.~~ **Withdrawn: there is no uncovered case — see the Review updates section.**

### DIFF
--- a/app/services/find_timeslices_with_revisions.rb
+++ b/app/services/find_timeslices_with_revisions.rb
@@ -1,17 +1,18 @@
 # frozen_string_literal: true
 
 require_dependency "#{Rails.root}/lib/replica"
+require_dependency "#{Rails.root}/lib/revision_data_manager"
 require_dependency "#{Rails.root}/lib/utils/replica_timeslice_bounds"
 
 #= Determines which of a course's wiki timeslices contain at least one tracked
-# revision, using a single wide-window Replica query instead of one query per
+# revision, using a wide-window Replica query instead of one query per
 # timeslice. UpdateCourseWikiTimeslices uses this to skip fetching timeslices
 # that are known to be empty — the dominant update cost for long-dormant
 # courses with far-future end dates, whose ingestion watermark never advances
 # past the last student edit (issue #7005).
 #
 # Exposes `slice_starts`: a Set of start datetimes for the timeslices that
-# contain revisions, or nil if the Replica query failed — callers should fall
+# contain revisions, or nil if a Replica query failed — callers should fall
 # back to a full per-timeslice scan in that case.
 class FindTimeslicesWithRevisions
   attr_reader :slice_starts
@@ -29,20 +30,40 @@ class FindTimeslicesWithRevisions
 
   def perform
     return if @timeslices.empty?
-    raw = fetch_wide_window_revisions
-    # A non-Enumerable result means the Replica query failed after retries;
-    # leave slice_starts nil so the caller knows the gap couldn't be checked.
-    return unless raw.is_a?(Enumerable)
-    rev_timestamps = raw.filter_map { |row| row['rev_timestamp'] }.sort
+    rows = fetch_wide_window_revisions
+    # nil means a Replica query failed after retries; leave slice_starts nil so
+    # the caller knows the gap couldn't be checked.
+    return if rows.nil?
+    rev_timestamps = rows.filter_map { |row| row['rev_timestamp'] }.sort
     @slice_starts = nonempty_slice_starts(rev_timestamps)
   end
 
+  # One query per chunk of users, each covering the whole timeslice range. The
+  # user list is chunked at the same limit RevisionDataManager uses, to stay
+  # within the Replica endpoint's memory limit; a wide window makes that limit
+  # more relevant here, not less. Utils.chunk_requests can't be used: it maps a
+  # failed chunk's nil to [], which is indistinguishable from a genuinely empty
+  # result and would mark timeslices empty in error. Returns nil if any chunk
+  # fails, so that the caller falls back to a full per-timeslice scan.
   def fetch_wide_window_revisions
-    Replica.new(@wiki, @update_service).get_revisions_raw(
-      @course.students,
-      ReplicaTimesliceBounds.real_start(@course, @timeslices.first.start),
-      ReplicaTimesliceBounds.real_end(@course, @timeslices.last.end)
-    )
+    @course.students.each_slice(RevisionDataManager::MAX_USERNAMES)
+           .each_with_object([]) do |users, rows|
+      raw = query_replica(users)
+      return nil unless raw.is_a?(Enumerable)
+      rows.concat raw.to_a
+    end
+  end
+
+  def query_replica(users)
+    Replica.new(@wiki, @update_service).get_revisions_raw(users, rev_start, rev_end)
+  end
+
+  def rev_start
+    @rev_start ||= ReplicaTimesliceBounds.real_start(@course, @timeslices.first.start)
+  end
+
+  def rev_end
+    @rev_end ||= ReplicaTimesliceBounds.real_end(@course, @timeslices.last.end)
   end
 
   # Replica rev_timestamps are '%Y%m%d%H%M%S' strings, the same format
--- a/app/services/find_timeslices_with_revisions.rb
+++ b/app/services/find_timeslices_with_revisions.rb
@@ -1,0 +1,64 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/replica"
+require_dependency "#{Rails.root}/lib/utils/replica_timeslice_bounds"
+
+#= Determines which of a course's wiki timeslices contain at least one tracked
+# revision, using a single wide-window Replica query instead of one query per
+# timeslice. UpdateCourseWikiTimeslices uses this to skip fetching timeslices
+# that are known to be empty — the dominant update cost for long-dormant
+# courses with far-future end dates, whose ingestion watermark never advances
+# past the last student edit (issue #7005).
+#
+# Exposes `slice_starts`: a Set of start datetimes for the timeslices that
+# contain revisions, or nil if the Replica query failed — callers should fall
+# back to a full per-timeslice scan in that case.
+class FindTimeslicesWithRevisions
+  attr_reader :slice_starts
+
+  # `timeslices` is a collection of CourseWikiTimeslice records for one wiki.
+  def initialize(course, wiki, timeslices, update_service: nil)
+    @course = course
+    @wiki = wiki
+    @timeslices = timeslices.sort_by(&:start)
+    @update_service = update_service
+    perform
+  end
+
+  private
+
+  def perform
+    return if @timeslices.empty?
+    raw = fetch_wide_window_revisions
+    # A non-Enumerable result means the Replica query failed after retries;
+    # leave slice_starts nil so the caller knows the gap couldn't be checked.
+    return unless raw.is_a?(Enumerable)
+    rev_timestamps = raw.filter_map { |row| row['rev_timestamp'] }.sort
+    @slice_starts = nonempty_slice_starts(rev_timestamps)
+  end
+
+  def fetch_wide_window_revisions
+    Replica.new(@wiki, @update_service).get_revisions_raw(
+      @course.students,
+      ReplicaTimesliceBounds.real_start(@course, @timeslices.first.start),
+      ReplicaTimesliceBounds.real_end(@course, @timeslices.last.end)
+    )
+  end
+
+  # Replica rev_timestamps are '%Y%m%d%H%M%S' strings, the same format
+  # ReplicaTimesliceBounds produces, so lexicographic comparison matches
+  # chronological order and no timezone parsing is needed.
+  def nonempty_slice_starts(rev_timestamps)
+    @timeslices.each_with_object(Set.new) do |timeslice, starts|
+      first_in_slice = rev_timestamps.bsearch do |ts|
+        ts >= ReplicaTimesliceBounds.real_start(@course, timeslice.start)
+      end
+      next if first_in_slice.nil?
+      starts << timeslice.start if within_slice?(first_in_slice, timeslice)
+    end
+  end
+
+  def within_slice?(rev_timestamp, timeslice)
+    rev_timestamp <= ReplicaTimesliceBounds.real_end(@course, timeslice.end)
+  end
+end
--- a/app/services/update_course_wiki_timeslices.rb
+++ b/app/services/update_course_wiki_timeslices.rb
@@ -13,10 +13,10 @@ require_dependency "#{Rails.root}/lib/revision_scanner"
 # and, if the timeslice exceeds the threshold, recursively splits it until all
 # timeslices are within limits.
 class UpdateCourseWikiTimeslices
-  # When the range of timeslices to process is longer than this, a single
-  # wide-window Replica query determines which of them actually contain
-  # revisions, so that empty ones can be skipped instead of fetched one by
-  # one. Short ranges (the common case for active courses) aren't worth the
+  # When the range of timeslices to process is longer than this, a wide-window
+  # Replica query (one per chunk of users) determines which of them actually
+  # contain revisions, so that empty ones can be skipped instead of fetched one
+  # by one. Short ranges (the common case for active courses) aren't worth the
   # extra query.
   GAP_PRECHECK_THRESHOLD = 7
 
@@ -90,20 +90,30 @@ class UpdateCourseWikiTimeslices
     @debugger.log_update_progress :"timeslices_processed_#{wiki.id}"
   end
 
-  # Skips timeslices that a single wide-window Replica query shows to contain
-  # no revisions; processing them would be a no-op. The first timeslice is
+  # Skips timeslices that a wide-window Replica query shows to contain no
+  # revisions; processing them would be a no-op. The first timeslice is
   # always kept: it's the only one in the range that can already have recorded
   # revisions (it contains the ingestion watermark), so it must be re-fetched
   # to detect on-wiki revision deletions and to re-ingest partial data. If the
   # wide query fails, fall back to processing the full range.
   def precheck_nonempty_timeslices(wiki, to_process)
-    return to_process if to_process.count <= GAP_PRECHECK_THRESHOLD
     slices = to_process.to_a
-    nonempty = FindTimeslicesWithRevisions.new(@course, wiki, slices,
-                                               update_service: @update_service).slice_starts
-    return to_process if nonempty.nil?
+    return slices if slices.size <= GAP_PRECHECK_THRESHOLD
+    nonempty = nonempty_slice_starts(wiki, slices)
+    return slices if nonempty.nil?
     watermark_start = slices.map(&:start).min
     slices.select { |t| t.start == watermark_start || nonempty.include?(t.start) }
+  end
+
+  # The starts of the timeslices that contain revisions, or nil if that couldn't
+  # be determined. When there's no point importing revisions for this course,
+  # every per-timeslice fetch would short-circuit before reaching Replica, so no
+  # timeslice can contain revisions and no query is needed — skipping the query
+  # keeps such courses from paying for a precheck they can't benefit from.
+  def nonempty_slice_starts(wiki, slices)
+    return Set.new if @revision_updater.no_point_in_importing_revisions?
+    FindTimeslicesWithRevisions.new(@course, wiki, slices,
+                                    update_service: @update_service).slice_starts
   end
 
   def fetch_data_and_reprocess_acuwt_timeslices(wiki)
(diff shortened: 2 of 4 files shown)


ragesoss:
@gabina you probably want to have a look at this. It's a shortcut for processing long-running courses where the users have stopped editing, which seems to be one major source of slow updates.

I noticed the issue when I was waiting for the last `long_update` course to finish processing so I could restart the worker as part of the deploy today, and was surprised that this course took so long even though it didn't have recent edits: https://outreachdashboard.wmflabs.org/courses/tlv_university/TLV_University,_Science_Oriented_Youth_-_Alpha_Program_-_Cycle_C_(2016-2017)

ragesoss:
## Review notes

These findings come from a Claude Code review session. They are not meant to be trusted automatically or acted on blindly — read them, check the reasoning, and use or discard each one on your own judgment. Four of the five were reproduced with executed specs (noted below); the remaining one comes from code reading and may be wrong.

Verified locally on the branch with master merged in: RuboCop clean on all four changed files; the PR's own specs 35/35; and 96/96 for existing specs in the affected areas (`prepare_timeslices`, `split_timeslice`, all four `update_timeslices_*`, `reprocess_article_course_user_wiki_timeslices`, `timeslice_manager`, `course_revision_updater`, `revision_data_manager`, `update_course_stats`, `constant_update`, `daily_update`). No regressions.

The two central safety arguments in the description hold up under reading:

- **The no-op claim.** `should_update_timeslice?(only_new: true)` → `new_data?` → `CourseRevisionUpdater#new_revisions?`, which for an empty timeslice with `mw_rev_count` 0 and `last_mw_rev_datetime` nil compares `0 != 0 || nil != nil` → false. Nothing is persisted, and `RevisionScanner.schedule_revision_checks` sits inside the guarded branch. Skipping is equivalent.
- **The watermark claim.** `TimesliceManager#update_timeslices` sets `last_mw_rev_datetime` from *all* fetched revisions including system ones, while `mw_rev_count` counts a strict subset — so no timeslice past the max-`last_mw_rev_datetime` slice can hold recorded revisions, and deletion detection can't be lost by skipping. The `real_end` 1-second subtraction also means no revision is assigned to two adjacent slices, and the `bsearch` find-minimum block is properly monotonic over the sorted string timestamps.

---

### 1. Courses with no students now pay a Replica query where they previously paid zero

`CourseRevisionUpdater#no_point_in_importing_revisions?` short-circuits *every* per-timeslice fetch when `course.students.empty?`, and also for an `only_scoped_articles_course?` with no assignments and no categories. `FindTimeslicesWithRevisions` has no equivalent guard, so it issues one wide `revisions.php` call per wiki per update for exactly those courses.

`Course.ready_for_update` includes every **current** course via `current_or_set_to_full_update`, regardless of student count, and any current course spanning more than `GAP_PRECHECK_THRESHOLD` timeslices qualifies for the precheck. On the P&E Dashboard that covers every empty event or course, multiplied by tracked wikis and by update frequency. With zero students `compile_usernames_query([])` serializes to `usernames[]=`, so the query carries no usernames at all.

Reproduced with a spec: with `precheck_nonempty_timeslices` stubbed to a pass-through, a no-student course's update makes **0** `get_revisions_raw` calls; unstubbed it makes **1**, with `{users: 0, start: "20181101000000", end: "20181130235500"}`.

This is also why the PR's own precheck specs pass — see finding 5. Both new spec files exercise a course with zero students, so what they currently assert *is* this regressive path.

Returning an empty `Set` (rather than `nil`) when there is no point importing revisions is strictly better than current master: no query, *and* the whole no-op range gets skipped.

### 2. The wide query bypasses the `MAX_USERNAMES` chunking

`RevisionDataManager#get_revisions` chunks the user list at `MAX_USERNAMES = 10`, with the comment "We limit the number of usernames per query in order to avoid hitting the memory limit of the Replica endpoint." `FindTimeslicesWithRevisions#fetch_wide_window_revisions` passes `@course.students` whole to `get_revisions_raw`, over the widest window in the update — so both dimensions the chunking exists to bound are amplified at once. Reproduced with a spec: 25 students go out in a single call.

The description says `revisions.php` was checked for server-side **row** limits. The concern documented in `RevisionDataManager` is a **memory** limit, which is a different failure and isn't ruled out by the row-limit check. Worth also checking GET URL length: `usernames[]=` repeated per student, against Apache's default 8190-byte `LimitRequestLine`.

Failure is safe (`nil` → full-range fallback) but not free: `api_get` retries three times with `sleep 2` and then reports to Sentry, so a course that trips a limit turns the intended savings into a recurring per-wiki, per-update error. Chunking at `MAX_USERNAMES` and unioning the timestamps keeps the win (10 queries instead of 3,867 for a 100-student course). Note that `Utils.chunk_requests` can't be used as-is: it maps a failed chunk's `nil` to `[]` via `Array(info)`, which is indistinguishable from a genuinely empty result and would mark timeslices empty in error. Any nil chunk needs to abort the whole precheck to `nil`.

### 3. The description's only stated limitation appears to be wrong — the fix is better than advertised

> a flagged full update of a course with *no timeslices yet* (`recreate_timeslices` marks everything `needs_update`) does not benefit

It does benefit. `fetch_data_and_reprocess_timeslices` **clears** `needs_update` without reprocessing for any timeslice whose `start > ingestion_start`, and for a brand-new course `ingestion_start == course.start` — so every slice but the first falls through to the process loop, where the precheck applies. Measured on a 30-timeslice course with everything freshly flagged: `processed=29, reprocessed=1` with the precheck stubbed out, versus `processed=0, reprocessed=1` with it active.

More generally there seems to be no uncovered case: reprocessing only touches slices with `start <= ingestion_start`, while `to_process` begins *at* `ingestion_start`, so the intersection is at most the watermark slice (plus any split children of it) — which is unconditionally kept anyway. If that holds, the third "Open questions" bullet can be dropped rather than deferred.

### 4. Two queries where one would do

`to_process.count` issues a `SELECT COUNT(*)` and then `to_process.to_a` re-queries the same rows. Loading once and checking `slices.size` avoids the extra round trip, and returning the loaded array in the below-threshold branch avoids a third.

### 5. Both new spec files exercise the precheck with zero students

`JoinCourse` fails `student_joining_before_approval?` for a course with no campaign and returns silently — `approved?` is `campaigns.any? && !withdrawn`. Measured: 0 students without a campaign, 1 with one.

- `find_timeslices_with_revisions_spec.rb` never adds a campaign, so `expect(replica).to receive(:get_revisions_raw).with(course.students, …)` asserts against an empty relation and is vacuous on the user dimension.
- The `describe 'empty-gap precheck'` block in `update_course_wiki_timeslices_spec.rb` is a **sibling** of `describe 'UpdateCourseWikiTimeslices.run'`, not nested inside it, so it doesn't inherit that block's `course.campaigns << Campaign.first`. Measured inside it: `students=0, campaigns=0, no_point_in_importing_revisions?=true`.

So every assertion in the PR about how many timeslices the precheck fetches was measured on a course that in production would issue no Replica query at all. Adding `course.campaigns << Campaign.first` to both setups is a one-line change, and it's what surfaces findings 1 and 2 — with it in place, the three existing `empty-gap precheck` examples fail against the current implementation.

---

Findings 1, 2 and 5 are the ones worth addressing before merge: 1 and 2 are load *regressions* in a change whose purpose is load reduction, and 5 is what hid them. All three are small fixes. 3 and 4 are cleanups.

Untested areas that remain, for the record: the `no_point_in_importing_revisions?` interaction (silent, and affects a large share of P&E courses), large unchunked user lists, and the multi-wiki case — the precheck runs per wiki, but only enwiki is exercised.

Fixes for 1, 2, 4 and 5, plus specs covering the first two, are prepared locally and will follow as a commit on this branch.

*Reviewed in a Claude Code session (~35 minutes, 2 interactions). Claude Code read the changed services and their collaborators, ran the linters and the existing and PR specs listed above, and wrote and ran an exploratory spec that reproduced findings 1, 2, 3 and 5. Finding 4 comes from code reading alone. Sage requested the review and read this comment before posting, but has not independently re-derived the findings; the analysis is Claude Code's and the reasoning should be checked rather than assumed.*

gabina:
I think this strategy makes sense. It's an optimization I had in mind a long time ago. It's the kind of change in the code that I'm never sure about making because it makes the algorithm more complex, and I don't think the time we're going to save is really meaningful, although maybe it is. My only real point here is that I would make `GAP_PRECHECK_THRESHOLD` greater. Although not usual (and undesirable), we may have 7 days of latency if something goes wrong with the queues, and I guess it would be nice to avoid going through the `FindTimeslicesWithRevisions` branch for every course in that case. Just to keep that case as the exception.

ragesoss:
That makes sense. I think a 30-day gap would be fine. I'll make that change.

The course that instigated this change was taking ~22 minutes every update cycle, which at current latency is about 4 hours of worker time per day.

ragesoss:
## Predicted production impact (pre-deploy baseline, 2026-08-20)

Documented before deploy so the actual change in update latency and system load can be measured against it.

**Cohort**: a prod console query identified 456 current courses whose latest update scanned more than the threshold's worth of timeslices. Public `course.json` data was available for 415 (the other 41 are private and excluded; their console-reported `processed` counts suggest they would add roughly 5–10% to the totals). Per-course steady state = median over the ~10 retained `update_logs` entries; frequency from `updates.average_delay`.

| Metric (415-course cohort) | Now | Predicted | Change |
|---|---|---|---|
| Timeslice scans/day | ~4.79M | ~80k | −98% |
| Replica-tools queries/day (incl. new precheck queries) | ~6.8M | ~210k | −97% |
| Update worker-time/day | ~156 h (estimate; see capacity note) | ~45–65 h | **−80 to −100 h/day** |
| Median inter-update interval | 2.4 h | (expected to fall) | — |

Savings are estimated per course as min(0.35 s × scans avoided, observed duration − a per-course overhead floor) — a flat count-based figure (~458 h/day) is not credible because measured per-slice costs range 0.01–0.35 s across the cohort (zero-student courses make no HTTP calls at all; those are covered by the `no_point_in_importing_revisions?` guard instead).

Capacity note: total course-update capacity is 6 workers × 24 h = **144 h/day** (the 2 ACUWT workers are idle and do not pull from other queues). The 156 h/day demand estimate slightly exceeds that ceiling — which both confirms the system runs at ~full saturation and indicates the `average_delay`-derived figures carry roughly 10% upward noise. The savings range above is reconciled to the capacity ceiling; the un-reconciled model gives −85 to −105 h/day.

Savings are concentrated: the top 15 courses account for roughly half. Expected before → after durations for the heaviest: Istituto ICAR 38 min → ≲1 min; Cochrane WiR 35 min → ≲1 min; UAE Wikimedians 33 min → ≲1 min; TLV University (course 10781, the original #7005 case) 23 min → seconds, with `processed` collapsing from 3,867 to ~4.

### Queue latency

Current Sidekiq snapshot (2026-08-20), with 6 course-update workers (3 long, 2 medium, 1 short) that pull from other queues when their own is empty; the 2 ACUWT workers are idle and do **not** cross-pull:

| Queue | Jobs waiting | Latency | Workers | Cohort demand now | Cohort demand predicted |
|---|---|---|---|---|---|
| long_update | 22 | 6,501 s (1.8 h) | 3 (72 h/day) | ~61 h/day | ~10 h/day |
| medium_update | 271 | 9,801 s (2.7 h) | 2 (48 h/day) | ~59 h/day | ~13 h/day |
| short_update | 472 | 1,099 s (18 min) | 1 (24 h/day) | ~37 h/day | ~29 h/day |

Method: queue latency ≈ time to drain the backlog ≈ Σ(queued job durations) ÷ workers. Cross-check: the snapshot's implied average job (latency × workers ÷ jobs) is 886 s / 72 s / 2.3 s for long/medium/short — matching the affected cohort's per-queue update durations, so these backlogs are dominated by precisely the courses this PR affects. Cohort demand per queue comes from assigning each course by the actual `CourseQueueSorting` rule (max recent update duration: ≤30 s short, ≤600 s medium, else long). The system currently runs at ~100% utilization: cohort demand alone is 84% of the long queue's capacity, 123% of medium's nominal capacity, and 152% of short's — the deficits covered only by the six update workers cross-pulling, with no spare capacity elsewhere (estimated demand ≈ the 144 h/day ceiling). That is why latencies sit at hours. When demand falls well below capacity, latency stops being backlog-drain time and collapses to "wait for the next free worker."

Predicted latencies:

- **long_update: 6,501 s → single-digit minutes** (bounded by occasional genuinely-long active-course jobs). 18 of the 23 cohort long-queue courses should re-sort into medium/short within about a day, as fast updates displace the slow entries in their update-log window.
- **medium_update: 9,801 s → low minutes** — this queue goes from over nominal capacity to ~25–30% of it; ~160 of its 197 cohort courses re-sort to short.
- **short_update: 1,099 s → several-fold lower but not collapsing** — its jobs are already ≤30 s so the precheck saves little directly, and it inherits ~180 re-sorted courses; the offset is freed long/medium workers idle-pulling short jobs.

Caveat on interpreting post-deploy numbers: as queues drain, courses get updated more often (`average_delay` shrinks toward the scheduling cadence), so some of the freed capacity converts into update freshness rather than idle workers. Worker-hours per update-cycle and per-course `update_logs` durations are the stable before/after metrics; queue latency and `average_delay` should both drop sharply but their exact values depend on that feedback.

**Post-deploy observables**: (1) `update_logs` durations and `processed` for the top-15 courses; (2) the three queue latencies vs the snapshot above; (3) `updates.average_delay` across all courses vs the 2.4 h cohort median; (4) replica-revision-tools request volume on Toolforge.

Assumes deploy at commit 5fa706cdd (threshold 30).

*Prediction assembled in a Claude Code session (~30 min, 5 interactions): Sage supplied the affected-course list, the queue snapshot, and the worker topology from prod, and reviewed the analysis and this comment before posting; the fetching, aggregation, and estimates are Claude Code's work from public course.json data, calibrated against the course-10781 measurements from the #7005 investigation.*

*Edited 2026-08-20: corrected the worker topology (ACUWT workers do not cross-pull, per Sage) — capacity is 144 h/day, not 192; savings restated as −80 to −100 h/day; the saturation diagnosis and latency predictions are unchanged (strengthened, if anything).*

ragesoss:
@gabina Claude estimates this will dramatically reduce latency, such that most courses will be updated every few minutes. Excited to find out if that's true.

gabina:
>  Claude estimates this will dramatically reduce latency, such that most courses will be updated every few minutes. Excited to find out if that's true.

Yes, it's pretty interesting. I think it sounds reasonable for historical courses with a lot of tracked wikis that maybe don't have data for the last years for some of them, and maybe that represent a lot of processing time in our system. Will see.

PR #7009, opened 2026-08-20
feat: implement precomputed retained editor metrics for courses_users…

### LINKED ISSUE #6984: Pre-compute `retained_new_editors` for System & Campaign CSV Exports
### Current Behavior:

`RetainedNewEditorsStats` calculates whether new editors remained active after a course by making **live Wikipedia API calls** (`list=usercontribs`) per course. It batches usernames in groups of 40 and queries the home wiki for post-course contributions (≥1 mainspace edit after `course.end + 7 days`).

This works for **single-course CSV exports** (`CourseCsvBuilder` with `@per_wiki = true`), but it **cannot scale** for system-wide or campaign-wide bulk exports because:

- Exporting 1,000+ courses would fire thousands of Wikipedia API requests, risking rate-limiting and timeouts.
- Each API call adds network latency, a system export could require ~5,000+ API calls.
- Sidekiq background workers would likely time out or block other jobs.

As a result, `SystemCsvBuilder` and `CampaignCsvBuilder` currently **cannot include `retained_new_editors`** in their CSV output.

### Desired Behavior:

Two new columns are added to the **`courses_users`** join table (rather than `users`, since retention is inherently course-specific):
- `retained_after_course` (boolean): whether the student continued editing after this specific course ended
- `retained_after_course_checked_at` (datetime):  when the check was performed, for auditability and debugging

A daily background worker (`RetainedEditorCheckWorker`) checks **new editors only** , students enrolled in a course (`role = 0`) who registered during the course window (`users.registered_at` between `courses.start` and `courses.end`). Students who already had accounts before the course are excluded from the check. The worker runs once a course has ended ≥60 days ago and saves `true`/`false` on the `courses_users` record permanently, each student-course record is checked once and never queried on Wikipedia again.

With this column on `courses_users`, all CSV builders can compute `retained_new_editors` count per course with a simple database query instead of live API calls:

```ruby
courses_users
  .joins(:user)
  .where(role: CoursesUsers::Roles::STUDENT_ROLE)
  .where('users.registered_at >= courses.start AND users.registered_at <= courses.end')
  .where(retained_after_course: true)
  .count
```

This enables `SystemCsvBuilder`  to include retained editor stats efficiently in bulk exports. `CourseCsvBuilder` and `CampaignCsvBuilder` can also be updated to read directly from the database with an API fallback for recent courses.

### Additional Context:

**Column definitions on `courses_users`:**
- `retained_after_course` (boolean, default: `nil`)
  - `nil` = not yet checked (course ended less than 60 days ago, or worker hasn't run yet)
  - `true` = student made ≥1 mainspace edit after `course.end + 7 days`
  - `false` = student did not make any mainspace edit after the threshold
- `retained_after_course_checked_at` (datetime, default: `nil`)
  - Records timestamp when the worker checked this record

**Worker behavior (`RetainedEditorCheckWorker`):**
- Runs daily as part of `DailyUpdate` (Sidekiq queue: `daily_update`).
- Finds **new editors only**: `courses_users` records with `role = 0` (student) where `retained_after_course IS NULL`, `users.registered_at` is between `courses.start` and `courses.end`, and `courses.end` was ≥60 days ago.
- Calls MediaWiki `usercontribs` API to check for mainspace edits after `course.end + 7 days` (namespace 0).
- Batches API calls in groups of 40 usernames (matching the `RetainedNewEditorsStats` pattern).
- Sets `retained_after_course = true/false` and `retained_after_course_checked_at = Time.zone.now` on `courses_users` permanently — never re-queries that record.
- If the MediaWiki API fails or returns `nil`, records remain `nil` for automatic retry on the next run.

### Implementation Steps:

1. **Migration**: Add `retained_after_course` (boolean) and `retained_after_course_checked_at` (datetime) columns (with index) to the `courses_users` table.
2. **Worker**: Create `RetainedEditorCheckWorker` background job.
3. **Daily Cycle**: Enqueue `RetainedEditorCheckWorker` in `DailyUpdate`.
4. **Rake Task**: Create `rake retained_editors:backfill` to process historical student records from courses that ended ≥60 days ago.
5. **CSV Builder**:
   - Update `SystemCsvBuilder` and  to batch-preload and include `retained_new_editors`.
6. **Specs**: Add unit and integration tests for worker, rake task, and CSV builders.

### Future Scope:
Once this core pre-computation infrastructure and backfill are merged and verified in production:
- **`CampaignCsvBuilder`**: Transition campaign CSV exports to use preloaded DB queries from `courses_users` instead of live queries.
- **`CourseCsvBuilder`**: Update single-course CSV exports to check the database first and only fall back to live API calls for recently ended courses (<60 days).

lakshita10341:
## What this PR does

This PR introduces pre-computed retained editor metrics to support scalable bulk reporting and system-wide stats exports without incurring live Wikipedia API calls.
Closes #6984 

### Background & Problem
`RetainedNewEditorsStats` previously calculated whether new editors remained active post-course by firing live `list=usercontribs` MediaWiki API calls per course. While this works for single-course CSV exports, it cannot scale for system-wide bulk CSV exports (`SystemCsvBuilder`) across thousands of courses due to API rate-limiting, latency, and Sidekiq worker timeouts.

### Solution & Key Changes
1. **Database Schema**:
   - Added `retained_after_course` (boolean, default: `nil`) and `retained_after_course_checked_at` (datetime) to `courses_users`.
   - Added `retained_new_editors_count` (integer, default: `0`) to `system_stats` and `facilitator_stats`.
2. **Background Worker (`RetainedEditorCheckWorker`)**:
   - Runs daily via `DailyUpdate` for courses ended ≥ 30 days ago.
   - Filters strictly for new-editor students (registered between `course.start` and `course.end`).
   - Queries `course.home_wiki` via `usercontribs` API in batches of 40 users for mainspace edits (namespace 0) after `course.end + 7 days`.
   - Utilizes pagination handling (`continue` parameter) with early exit when all batch users are confirmed active.
   - Saves results permanently to `courses_users` records.
3. **Stat Workers & System CSV Builder**:
   - Updated `SystemStatUpdateWorker` and `FacilitatorStatUpdateWorker` to calculate retained new editors using `NewEditorDateConditions::DURING_PROGRAM`.
   - Updated `SystemCsvBuilder` and `SystemStatsCsvBuilder` to pre-load `retained_new_editors` count per course via direct DB queries instead of live API calls.
4. **Historical Rake Task (`rake retained_editors:backfill`)**:
   - An idempotent, rate-limited Rake task to backfill historical courses ended ≥ 30 days ago with progress tracking.
5. **UI & Leaderboard Integration**:
   - Added `Retained Editors` KPI metric to System Stats overview.
   - Added sortable `Retained Editors` column to Facilitator Leaderboard.

 ```bash
 rake retained_editors:backfill
 ```
 **Why it is mandatory**: Running this task populates historical `retained_after_course` records for all existing courses ended ≥ 30 days ago. Executing the backfill prior to stat worker snapshot runs ensures that system stats, dashboard KPIs, facilitator leaderboards, and CSV exports reflect accurate historical retention data from day one without leaving unchecked courses as zero.

## AI usage

- Used AI coding assistant to assist with architectural design, drafting RSpec test suites for `RetainedEditorCheckWorker`, refactoring queries to use `NewEditorDateConditions::DURING_PROGRAM`, and ensuring schema/migration consistency.

- Currently, retention checks are conducted against `course.home_wiki` (matching existing `RetainedNewEditorsStats` behavior). Cross-wiki retention tracking across multi-wiki courses can be considered in future scope if needed.


ragesoss (changes requested):
Reviewed with the help of Claude Code. A note on how this feedback is meant to be used: it is not to be automatically trusted or blindly acted upon. Some of it may be wrong. Read it, understand the reasoning, and then use or ignore each point based on your own judgment.

This is a substantial and well-organized PR — the schema is hand-curated correctly (running `db:migrate` locally reorders 292 lines of `schema.rb`; this diff is surgical), the controller refactor appears behavior-preserving apart from JSON key ordering, and the worker spec covers more ground than most first-pass GSoC test suites.

**What was checked locally:** the branch was checked out, merged with current master (clean), `yarn build` run (clean), the 8 affected spec files run (72 examples, 0 failures), RuboCop and ESLint run on all changed files, and four findings below were reproduced in a throwaway spec included at the end. Two MediaWiki API assumptions were probed against live en.wikipedia.org.

**On CI:** the failing check is `spec/features/campaign_users_spec.rb` (axe `page-has-heading-one`). It passes locally on this branch and nothing here touches that page, so it reads as flaky/environmental rather than PR-introduced.

---

## Findings that look worth addressing before merge

### 1. RuboCop fails on this branch

```
app/workers/retained_editor_check_worker.rb:82:3: C: Metrics/MethodLength:
  Method has too many lines. [21/16]  def process_batch
```

Introduced by the third commit. CI did not report this: the `Ruby linting` step runs *after* rspec in the workflow, and the job exited at the rspec step, so the lint step never ran. The red X is currently hiding a second, separate failure.

### 2. One invalid username permanently poisons a whole 40-record batch

Probed against the live API — a single bad `ucuser` value fails the *entire* request rather than skipping that user:

```json
{"error":{"code":"baduser","info":"Invalid value \"Bad#Name\" for user parameter \"ucuser\"."}}
```

No `usercontribs` node comes back, so `query_usercontribs` returns nil, `fetch_active_usernames` returns nil, and `process_batch` returns 0 **without writing anything**. All 40 records stay `NULL` and are re-queried on the next run — and the next. `WikiApi#mediawiki` retries three times and calls `log_error` before returning nil, so each affected batch costs 3 API requests plus a Sentry event, every day, indefinitely. The 38 valid editors in that batch are never recorded.

Reproduced below (`batch poisoning by one bad username`): three `perform` calls, three queries, zero records written.

Valid-but-nonexistent usernames are fine — those return `"usercontribs":[]`.

Worth considering: catching the error and falling back to per-user queries for the failed batch, or validating usernames before batching. Note this is a *new* failure mode rather than a regression — `RetainedNewEditorsStats` had the same API exposure, but its consequence was one low CSV cell, not a permanent daily retry loop.

### 3. Pagination is unbounded, and the cost lands on the backfill

Probed against the live API: with a multi-valued `ucuser`, MediaWiki returns results **grouped by actor**, not interleaved by timestamp. The continue token is literally `"uccontinue":"actor|14213|20201229171937|997021250"`. So one prolific user in a batch gets paged through 500 edits at a time before the API reaches the next user.

Compounding this, the early exit only fires when *every* user in the batch has been seen:

```ruby
break if active_usernames.superset?(target_users)
```

Any batch containing a non-retained user — the common case — therefore pages to exhaustion.

For the daily path this is harmless: a 23-day window rarely exceeds one page. For the **backfill** it is not, because the window there is open-ended. A 2015 course whose new editor became a prolific Wikipedian costs one request per 500 of their lifetime mainspace edits, and the rake task's `sleep 0.2` is between *courses*, not between requests — so that burst is entirely unthrottled.

Reproduced below (`unbounded pagination cost`): 40 requests to determine two booleans.

Given the actor ordering, a cheap fix is to re-issue the query with only the not-yet-found usernames on each continuation, rather than following the continue token.

### 4. The daily worker has no cap on a single run

`eligible_course_ids` returns every non-private course ended ≥30 days ago with unchecked new editors — i.e. the entire history — and `perform` walks all of them in one invocation. Reproduced below (`no bound on a single run`).

The practical consequence is that the throttled backfill rake task never gets to be the thing that does the backfill: `DailyUpdate` fires the worker automatically the day this deploys, and it attempts the full history at once. It is also enqueued onto `daily_update` immediately *before* `SystemStatUpdateWorker`, at concurrency 2, so the daily stats snapshot can end up queued behind a very long retention scan.

A per-run limit (courses or API calls) would address both this and the earlier "designed to run across multiple invocations" guidance on #6984, which the rake task also does not currently implement.

---

## Design questions — these are calls for the maintainers, not defects in the implementation

### 5. The two KPIs displayed side by side use different new-editor definitions

`retained_new_editors_count` counts under `NewEditorDateConditions::DURING_PROGRAM`, but every "New Editors" figure shown next to it — KPI row, wiki breakdown, facilitator leaderboard — uses `WITH_PREREGISTRATION` (the 60-day pre-window).

Reproduced below (`denominator mismatch between the two KPIs`): a retained editor who registered 30 days before course start produces `new_editors_count_with_preregistration = 1` and `retained_new_editors_count = 0`.

The worker matching `RetainedNewEditorsStats`' `DURING_PROGRAM` definition is correct per #6984. The issue is the *pairing* in the UI: a reader treating the two numbers as a retention rate will get a wrong one.

### 6. The metric is frozen at check time, which makes it non-comparable across months

Backfilled historical courses get an open-ended observation window (end+7 → today, potentially years). Courses checked going forward get 23 days (end+7 → end+30) and are never rechecked — `retained_after_course_checked_at` is written but never read anywhere.

The consequence for the trend chart is that retention will appear to collapse for recent months as a methodology artifact rather than a real change. #6984 does specify "never re-queries that record", so this follows the spec; whether that consequence for the trend line was intended seems worth confirming before it ships to a public page.

### 7. Two different `retained_new_editors` columns now share a name

`CourseCsvBuilder` in per-wiki mode still computes this live with an open window; `SystemCsvBuilder` now reads the frozen DB flag. Same header, different meanings. Deferring `CourseCsvBuilder` was the agreed v1 scope, so this may only need documenting.

### 8. No index was added

Step 1 of #6984 called for one. `courses` has only a `slug` index, so the daily query full-scans `courses` and probes `courses_users` per course. That is likely tolerable once a day, but the decision is worth making deliberately now — adding an index to `courses_users` later is the migration that actually needs care.

On migration safety more generally: all three `add_column`s are nullable or literal-default, appended at end of table, with no backfill, no NOT NULL and no FK — the `ALGORITHM=INSTANT` shape for MariaDB 10.3+. `system_stats` and `facilitator_stats` are small. `courses_users` is the only large table and two instant ALTERs on it remain metadata-only. Assuming both production hosts are on MariaDB ≥10.3, this should be safe inside a normal deploy.

### 9. User-facing copy

`en.yml` adds "Retained Editors" in three places. Per this repo's conventions, user-facing text should come from the maintainers rather than an AI assistant — and the term's meaning here (≥1 mainspace edit on the home wiki, 7–30 days after course end, measured once and frozen) is not something a reader will infer from the label.

---

## Minor

- No spec for the backfill rake task, though step 6 of #6984 asked for one. It is the component that will touch production hardest.
- `spec/workers/retained_editor_check_worker_spec.rb` — the continuation example asserts that `inactive_user` ends up marked *retained*. That is correct for the stub, but the naming makes the example hard to read.
- `fetch_active_usernames` builds a `Set` and returns `.to_a`; the caller then does `include?` in a loop. Returning the `Set` would be simpler.
- Roughly 50 of the 86 changed lines in `system_stats_controller.rb` are renames, collapsed hash literals and removed blank lines — apparently to stay under the 130-line class limit. The reformatting appears behavior-preserving, but it makes the actual change hard to pick out. Extracting a class or a presenter would cost less diff noise than compressing the formatting.
- No screenshots for the new KPI tile, leaderboard column and chart series.

---

<details>
<summary>Exploratory spec used to reproduce findings 2, 3, 4 and 5 (click to expand)</summary>

This was generated by Claude Code during review and run locally against this branch — all four examples pass, which is what confirms the four behaviors. It has not been line-by-line reviewed by a human and is not proposed for merging as-is; it is included so the findings can be re-run rather than argued about. Save as `spec/workers/retained_editor_review_exploratory_spec.rb`.

```ruby
# frozen_string_literal: true

require 'rails_helper'
require_dependency "#{Rails.root}/app/workers/retained_editor_check_worker"
require_dependency "#{Rails.root}/app/workers/system_stat_update_worker"

describe 'PR #7009 exploratory findings' do

  def response_with(contribs, continue: nil)
    data = { 'usercontribs' => contribs }
    data['continue'] = continue if continue
    instance_double(MediawikiApi::Response, data:)
  end

  def new_editor_in(course, username:, registered_at: nil)
    user = create(:user, username:, registered_at: registered_at || (course.start + 1.day))
    create(:courses_user, course:, user:,
                          role: CoursesUsers::Roles::STUDENT_ROLE,
                          retained_after_course: nil)
  end

  # FINDING 2
  describe 'batch poisoning by one bad username' do
    let(:course) do
      create(:course, start: 90.days.ago, end: 70.days.ago, home_wiki: wiki, private: false)
    end

    before do
      new_editor_in(course, username: 'Good editor one')
      new_editor_in(course, username: 'Good editor two')
      new_editor_in(course, username: 'Bad#Name') # illegal char -> baduser
    end

    it 'leaves every record in the batch unchecked and re-queries them forever' do
      call_count = 0
      # WikiApi#query returns nil after retries when the API errors out.
      allow_any_instance_of(WikiApi).to receive(:query) do
        call_count += 1
        nil
      end

      3.times { RetainedEditorCheckWorker.new.perform }

      expect(CoursesUsers.where(retained_after_course: nil).count).to eq(3)
      expect(call_count).to eq(3)
    end
  end

  # FINDING 3
  describe 'unbounded pagination cost' do
    let(:course) do
      create(:course, start: 3.years.ago, end: 2.years.ago, home_wiki: wiki, private: false)
    end

    before do
      new_editor_in(course, username: 'Prolific editor')
      new_editor_in(course, username: 'Quiet editor') # zero post-course edits
    end

    it 'makes one request per 500 edits of the prolific user before giving up' do
      pages = 40 # 20,000 mainspace edits, plausible for a backfilled 2-year-old course
      call_count = 0

      allow_any_instance_of(WikiApi).to receive(:query) do
        call_count += 1
        last_page = call_count >= pages
        response_with(
          [{ 'user' => 'Prolific editor' }],
          continue: last_page ? nil : { 'uccontinue' => "actor|1|page#{call_count}" }
        )
      end

      RetainedEditorCheckWorker.new.perform

      expect(call_count).to eq(pages)
      expect(CoursesUsers.joins(:user)
                         .find_by(users: { username: 'Prolific editor' })
                         .retained_after_course).to be true
      expect(CoursesUsers.joins(:user)
                         .find_by(users: { username: 'Quiet editor' })
                         .retained_after_course).to be false
    end
  end

  # FINDING 4
  describe 'no bound on a single run' do
    before do
      5.times do |i|
        course = create(:course, slug: "School/Course_#{i}_(term)",
                                 start: (400 - i).days.ago, end: (380 - i).days.ago,
                                 home_wiki: wiki, private: false)
        new_editor_in(course, username: "Student #{i}")
      end
    end

    it 'processes every eligible historical course in one perform' do
      call_count = 0
      allow_any_instance_of(WikiApi).to receive(:query) do
        call_count += 1
        response_with([])
      end

      processed = RetainedEditorCheckWorker.new.perform

      expect(processed).to eq(5)
      expect(call_count).to eq(5) # one API round trip per course, unthrottled
      expect(CoursesUsers.where(retained_after_course: nil).count).to eq(0)
    end
  end

  # FINDING 5
  describe 'denominator mismatch between the two KPIs' do
    let(:course) do
      create(:course, start: 90.days.ago, end: 70.days.ago, home_wiki: wiki, private: false)
    end

    before do
      # Registered 30 days before the course started: inside the 60-day
      # preregistration window, outside DURING_PROGRAM.
      preregistered = create(:user, username: 'Preregistered editor',
                                    registered_at: 120.days.ago)
      create(:courses_user, course:, user: preregistered,
                            role: CoursesUsers::Roles::STUDENT_ROLE,
                            retained_after_course: true)
    end

    it 'counts a retained preregistered editor in the denominator but not the numerator' do
      SystemStatUpdateWorker.new.perform
      stat = SystemStat.find_by(snapshot_date: Time.zone.today)

      expect(stat.new_editors_count_with_preregistration).to eq(1)
      expect(stat.retained_new_editors_count).to eq(0)
    end
  end
end
```

</details>

---

*Drafted in a Claude Code session (~35 min, 3 interactions). Sage chose the PR and asked for a full review, but did not independently verify any of the findings and did not review the exploratory spec line by line. Everything described as "verified" or "reproduced" was verified by Claude Code, not by a human: the branch was checked out and merged with master, `yarn build`, RuboCop, ESLint and the 8 affected spec files were run locally, the four-example exploratory spec above was written and run, and the two MediaWiki `usercontribs` behaviors (the `baduser` whole-request failure and the per-actor continuation ordering) were probed against live en.wikipedia.org. The design questions in items 5–9 are judgment calls that have had no human review yet. Weight accordingly.*

ragesoss (changes requested):
Re-review of 5ff1682, again with the help of Claude Code. Same note as before on how to use this feedback: it is not to be blindly trusted — read it, check the reasoning, and use your own judgment.

**What was checked locally:** the branch was checked out and merged with current master (clean), `yarn build` run (clean), RuboCop run on all changed files (clean — finding 1 from the prior review is resolved), the PR's spec files plus the other affected specs run (71 examples, 0 failures), and the two new findings below reproduced in a fresh exploratory spec (included at the end).

**On CI:** the current red X is `spec/models/visiting_scholarship_spec.rb` plus `spec/features/oauth_spec.rb` (which drives a live Wikipedia OAuth flow). Neither is touched by this PR, and the first fails intermittently locally on a different example each time — both read as flaky/environmental, not PR-introduced.

**Prior findings:** 1 (RuboCop) and 4 (per-run cap, via `DEFAULT_PERFORM_LIMIT = 50` with the backfill task correctly bypassing it) are resolved. 2 (baduser) is addressed for the invalid-username case itself, and the new rake spec covers the earlier "no backfill spec" note at a smoke-test level. Findings 5–9 were maintainer calls and remain open. However, the fixes for 2 and 3 each introduce a new problem:

## New issue A — the narrowed re-query never executes, so every batch is decided by its first 500-result page

The new loop in `fetch_batch_active_usernames` intends to re-query with only the not-yet-found usernames instead of following continue tokens. But after any page that finds at least one new user, `next_pending_and_continue` returns `[new_pending, nil]`, and `batch_complete?` then evaluates

```ruby
pending == (target_users - active_usernames).to_a && continue_param.nil?
```

which is true by construction — `new_pending` was computed from that exact expression one line earlier, and `continue_param` was just set to nil. So every narrowing immediately breaks the loop, and the narrowed re-query never fires. Since `query_usercontribs` also forces `continue` to nil when a page is empty, there is no path to a second loop iteration at all: one API request per batch, always, and the continue-token plumbing is dead code.

The consequence: any user whose contributions lie beyond page 1 of the batch response is permanently written `retained_after_course: false`. The live API groups `usercontribs` results by actor, so one user with 500+ mainspace edits in the window fills page 1 and every other user in the batch is recorded as not retained. For the daily 23-day window that takes an unusually active editor; for the open-ended backfill window it will be routine. Combined with the never-recheck design, these false negatives freeze into the dataset permanently.

Reproduced below (`the narrowed re-query never executes`): two users, where the second user's edits would be returned by the narrowed re-query — exactly one API call happens and the retained user is written as not retained.

A fix needs the loop to actually issue a query with the narrowed list before concluding the batch — e.g. only treat the batch as complete after a query *with the current pending set* comes back with no continue token, rather than comparing `pending` against a set difference that was just used to compute it.

## New issue B — a transient API failure now writes permanent wrong data instead of leaving records for retry

The per-user fallback treats a nil query result the same as "no contributions":

```ruby
active_usernames.add(username) if result && result[:users].any?
```

`WikiApi#query` returns nil after three retries on *any* error — so an API outage during a daily run permanently marks up to 50 courses' worth of students as not retained. The previous revision left records nil and retried the next day; the old spec `'leaves records as nil for future retry'` asserted exactly that behavior, and this commit rewrote that spec rather than preserving the behavior alongside the fallback.

Reproduced below (`transient API failure writes permanent results`): all queries return nil, and every record is written false and permanently leaves the eligible pool.

The two nil cases are worth separating: on an individual-query error, skip writing that user's record (leave nil for retry); on a successful query with empty contribs, write false. That does mean a genuinely invalid username is retried once per day — one request per bad name, far cheaper than the original per-batch poisoning — unless usernames are validated before querying or the baduser error code is surfaced through `WikiApi`. Which of those is worth doing is a judgment call, but silently converting outages into permanent "not retained" data seems clearly worse than either.

## Minor

- `eligible_course_ids` applies `LIMIT` with no `ORDER BY`, so which 50 courses run on a given day is up to the database. Today this self-advances because every processed course gets written — but if issue B is fixed by leaving records nil on error, failing courses could recur in the same arbitrary window. An explicit order (e.g. `courses.end`) makes the cap predictable. The new spec `'only checks up to the limit of courses specified'` also implicitly depends on which course the database happens to return first.
- The continuation-pages spec was deleted rather than updated, so multi-page behavior — exactly where issue A lives — now has no coverage. When A is fixed, a spec exercising "a user found only via the narrowed re-query" would pin the behavior.

<details>
<summary>Exploratory spec reproducing issues A and B (click to expand)</summary>

Generated by Claude Code during re-review and run locally against this branch merged with master — both examples pass, which is what confirms the two behaviors. Not proposed for merging as-is. Save as `spec/workers/retained_editor_rereview_exploratory_spec.rb`.

```ruby
# frozen_string_literal: true

require 'rails_helper'
require_dependency "#{Rails.root}/app/workers/retained_editor_check_worker"

describe 'PR #7009 re-review exploratory findings' do
  let(:course) do
    create(:course, start: 90.days.ago, end: 70.days.ago, home_wiki: wiki, private: false)
  end

  def response_with(contribs, continue: nil)
    data = { 'usercontribs' => contribs }
    data['continue'] = continue if continue
    instance_double(MediawikiApi::Response, data:)
  end

  def new_editor_in(course, username:)
    user = create(:user, username:, registered_at: course.start + 1.day)
    create(:courses_user, course:, user:,
                          role: CoursesUsers::Roles::STUDENT_ROLE,
                          retained_after_course: nil)
  end

  # FINDING A: next_pending_and_continue narrows `pending` to the not-yet-found
  # users and sets continue to nil — but batch_complete? then evaluates
  # `pending == (target_users - active_usernames).to_a && continue_param.nil?`,
  # which is true by construction immediately after every narrowing. The loop
  # breaks before the narrowed re-query is ever issued.
  describe 'the narrowed re-query never executes' do
    before do
      new_editor_in(course, username: 'Prolific editor')
      new_editor_in(course, username: 'Second editor')
    end

    it 'marks a retained user not-retained after a single API request' do
      queries = []
      allow_any_instance_of(WikiApi).to receive(:query) do |_instance, params|
        queries << params[:ucuser]
        if params[:ucuser] == ['Second editor']
          # The re-query the narrowing logic intends to make would find these
          # edits — but it never happens.
          response_with([{ 'user' => 'Second editor' }])
        else
          # Page 1 of the batch query holds only the prolific editor's edits
          # (the live API groups usercontribs by actor), plus a continue token.
          response_with([{ 'user' => 'Prolific editor' }],
                        continue: { 'uccontinue' => 'actor|999|20260801000000|123' })
        end
      end

      RetainedEditorCheckWorker.new.perform

      expect(queries.size).to eq(1)
      expect(queries.first).to match_array(['Prolific editor', 'Second editor'])
      second = CoursesUsers.joins(:user).find_by(users: { username: 'Second editor' })
      expect(second.retained_after_course).to be(false) # permanent false negative
    end
  end

  # FINDING B: the baduser fallback treats a nil individual-query result
  # (API error after retries) the same as "no edits", so an API outage
  # permanently records not-retained for everyone in the batch. The previous
  # version of this worker left records nil for retry on the next run.
  describe 'transient API failure writes permanent results' do
    before do
      new_editor_in(course, username: 'Editor one')
      new_editor_in(course, username: 'Editor two')
    end

    it 'marks every user not-retained and never rechecks them' do
      allow_any_instance_of(WikiApi).to receive(:query).and_return(nil)

      RetainedEditorCheckWorker.new.perform

      expect(CoursesUsers.where(retained_after_course: false).count).to eq(2)
      expect(RetainedEditorCheckWorker.eligible_course_ids).to be_empty
    end
  end
end
```

</details>

---

*Drafted in a Claude Code session (~25 min, 2 interactions). Sage requested the re-review and approved posting after reading the draft, but did not independently verify the findings. Everything described as "verified" or "reproduced" was verified by Claude Code, not by a human: the branch was checked out and merged with master, `yarn build`, RuboCop and the affected spec files were run locally, and the two-example exploratory spec above (AI-generated, not line-by-line reviewed by a human) was written and run against this branch.*

lakshita10341:
@ragesoss I have made the changes. PTAL when you get time

ragesoss (commented):
Re-review of 7c1f5bf, again with the help of Claude Code. Same note as before on how to use this feedback: it is not to be blindly trusted — read it, check the reasoning, and use your own judgment.

**What was checked locally:** the branch was checked out and merged with current master (clean), `yarn build` run (clean), RuboCop and ESLint run on all changed files (clean), the PR's spec files plus the other affected specs run (78 examples, 0 failures), the exploratory spec from the previous review rerun (both examples now fail, which is what confirms the two fixes), and three new exploratory examples written and run (included at the end).

**Prior findings:** Issues A (the narrowed re-query never executing) and B (transient API failures writing permanent `false`) are both resolved, and the new specs pin the behavior. The `LIMIT` without `ORDER BY` note is addressed with `order(:end)`, and the continuation-page coverage is back. The design questions 5–9 from the first review remain open for the maintainers and are not blocking on the author's side.

## Observations on the new commit — none look blocking

### 1. A partial per-user failure is now a permanent "unknown"

When the batch query fails and the per-user fallback then fails for some users, those users get `retained_after_course_checked_at` set with a nil status and leave the eligible pool for good. The spec `sets checked_at for failed user but leaves status nil` pins this as intended. It is a reasonable way to stop retrying genuinely invalid usernames every day, but `WikiApi#query` returns nil for a `baduser` error and for a timeout alike, so a transient error that lands mid-fallback also freezes that user as unknown. Distinguishing the two would need the error code surfaced through `WikiApi`, or username validation before querying. Worth a deliberate decision rather than a fix necessarily.

### 2. No circuit breaker during an API outage

If the API is down, each batch costs one batch query plus one per-user query for each of up to 40 users, and `WikiApi` turns each of those into three HTTP attempts and a Sentry event before returning nil. The run then moves on to the next course and repeats. Reproduced below (`API outage across a whole run`): two courses, every query nil, six worker-level queries, nothing written. Scaled to the 50-course daily limit that is on the order of 6,000 failed requests and 2,000 Sentry events in one run. Since a batch that returns with nothing queryable means even single-user lookups failed, stopping the run at that point would be a cheap guard.

### 3. Oldest-first ordering and the backfill

`eligible_course_ids` now orders by `courses.end` ascending, so the daily job processes the oldest unchecked courses first. Reproduced below (`daily budget ordering`). Until `rake retained_editors:backfill` has been run to completion, the daily budget of 50 courses goes entirely to historical courses and no recently-ended course is checked — the new KPI would stay empty for recent months. This is fine as long as the backfill is run right after deploy, but it makes that step required rather than optional; the PR description could say so.

### Minor

- `order(:end)` has no tiebreaker. The spec `only checks up to the limit of courses specified` creates two courses with the same `end` value and still depends on which one the database returns first. `order(:end, :id)` in the worker, and distinct end dates in the spec, would make both deterministic.

<details>
<summary>Exploratory spec used for observations 2 and 3 (click to expand)</summary>

Generated by Claude Code during re-review and run locally against this branch merged with master — all three examples pass. Not proposed for merging as-is. Save as `spec/workers/retained_editor_rereview2_exploratory_spec.rb`.

```ruby
# frozen_string_literal: true

require 'rails_helper'
require_dependency "#{Rails.root}/app/workers/retained_editor_check_worker"

describe 'PR #7009 second re-review exploratory checks' do
  let(:course) do
    create(:course, start: 90.days.ago, end: 70.days.ago, home_wiki: wiki, private: false)
  end

  def response_with(contribs, continue: nil)
    data = { 'usercontribs' => contribs }
    data['continue'] = continue if continue
    instance_double(MediawikiApi::Response, data:)
  end

  def new_editor_in(course, username:)
    user = create(:user, username:, registered_at: course.start + 1.day)
    create(:courses_user, course:, user:,
                          role: CoursesUsers::Roles::STUDENT_ROLE,
                          retained_after_course: nil)
  end

  def retention_of(username)
    CoursesUsers.joins(:user).find_by(users: { username: }).retained_after_course
  end

  # Simulates the live API's per-actor grouping: a page holds only the first
  # queried user who has edits (a prolific editor fills all 500 rows), with a
  # continue token whenever another queried user also has edits.
  def actor_grouped_api(users_with_edits, queries)
    proc do |_instance, params|
      queried = params[:ucuser]
      queries << queried
      with_edits = queried.select { |u| users_with_edits.include?(u) }
      first = with_edits.first
      if first
        more = with_edits.size > 1
        response_with([{ 'user' => first }],
                      continue: more ? { 'uccontinue' => "actor|#{first}|x" } : nil)
      else
        response_with([])
      end
    end
  end

  describe 'narrowing chain with several prolific editors' do
    before do
      %w[Alpha Beta Gamma Delta].each { |u| new_editor_in(course, username: u) }
    end

    it 'issues one shrinking query per prolific editor and resolves everyone correctly' do
      queries = []
      allow_any_instance_of(WikiApi)
        .to receive(:query, &actor_grouped_api(%w[Alpha Beta Gamma], queries))

      RetainedEditorCheckWorker.new.perform

      expect(queries.map(&:size)).to eq([4, 3, 2, 1])
      expect(retention_of('Alpha')).to be(true)
      expect(retention_of('Beta')).to be(true)
      expect(retention_of('Gamma')).to be(true)
      expect(retention_of('Delta')).to be(false)
    end
  end

  describe 'API outage across a whole run' do
    before do
      new_editor_in(course, username: 'Editor one')
      new_editor_in(course, username: 'Editor two')
      course_2 = create(:course, slug: 'School/Course_2_(term)', start: 100.days.ago,
                                 end: 80.days.ago, home_wiki: wiki, private: false)
      new_editor_in(course_2, username: 'Editor three')
      new_editor_in(course_2, username: 'Editor four')
    end

    it 'leaves every record unchecked, but keeps querying course after course' do
      call_count = 0
      allow_any_instance_of(WikiApi).to receive(:query) do
        call_count += 1
        nil
      end

      RetainedEditorCheckWorker.new.perform

      # nothing written: all four records stay eligible for tomorrow
      expect(CoursesUsers.where.not(retained_after_course_checked_at: nil).count).to eq(0)
      # per batch: 1 batch query + 1 individual query per user; both courses processed.
      # Each WikiApi#query here is 3 real HTTP attempts + a Sentry event.
      expect(call_count).to eq(6)
    end
  end

  describe 'daily budget ordering' do
    before do
      [400, 300, 200].each do |days|
        c = create(:course, slug: "School/Course_#{days}_(term)", start: (days + 30).days.ago,
                            end: days.days.ago, home_wiki: wiki, private: false)
        new_editor_in(c, username: "Student #{days}")
      end
    end

    it 'spends the per-run limit on the oldest-ended courses first' do
      allow_any_instance_of(WikiApi).to receive(:query).and_return(response_with([]))

      RetainedEditorCheckWorker.new.perform(1)
      expect(retention_of('Student 400')).to be(false)
      expect(retention_of('Student 300')).to be_nil

      RetainedEditorCheckWorker.new.perform(1)
      expect(retention_of('Student 300')).to be(false)
      expect(retention_of('Student 200')).to be_nil
    end
  end
end
```

</details>

---

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage requested the re-review and approved posting after reading the draft, but did not independently verify the findings. Everything described as "verified" or "reproduced" was verified by Claude Code, not by a human: the branch was checked out and merged with master, `yarn build`, RuboCop, ESLint and the affected spec files were run locally, the previous review's exploratory spec was rerun, and the three-example exploratory spec above (AI-generated, not line-by-line reviewed by a human) was written and run against this branch.*

ragesoss:
I think the terminology and metrics definition issues (5 and 7 from the first AI review) should be addressed. We should make sure any field that is using the pre-registration version of the calculation is explicitly labeled, and the preregistration window size should be easy to see (probably by including the "60 day" limit in the label).

Is there a public spot where the current WMF definition / rationale of "pre-registration" and the window they use for it is documented? If so, we should point to that from the NewEditorDateConditions. I'm worried about this getting out of sync with future changes to WMF metrics, if it's something we only know from direct communication with WMF folks.

lakshita10341:
> I think the terminology and metrics definition issues (5 and 7 from the first AI review) should be addressed. We should make sure any field that is using the pre-registration version of the calculation is explicitly labeled, and the preregistration window size should be easy to see (probably by including the "60 day" limit in the label).
> 
> Is there a public spot where the current WMF definition / rationale of "pre-registration" and the window they use for it is documented? If so, we should point to that from the NewEditorDateConditions. I'm worried about this getting out of sync with future changes to WMF metrics, if it's something we only know from direct communication with WMF folks.

No, The 60-day pre-registration window came from direct communication with WMF folks.  I wasn't able to find any public documentation for it.  I've added the labels with the term. 
I've  updated the labels in en.yml so that it explicitly shows "New Editors (60-day pre-registration)" in the UI tables and KPIs.
For Retained Editors, I kept it using the standard program window (students registered during the course) so it stays consistent with how other CSV builders calculate it. Plus, precomputing retention on `courses_users` allows other builders to reuse this data directly rather than doing live API calls.

lakshita10341:
We'll need to run `rake retained_editors:backfill` once to backfill historical data.
Could you let me know roughly how many total courses are currently on production? Just want to estimate about how long the backfill script will take to finish.

lakshita10341:
>1. A partial per-user failure is now a permanent "unknown"
When the batch query fails and the per-user fallback then fails for some users, those users get retained_after_course_checked_at set with a nil status and leave the eligible pool for good. The spec sets checked_at for failed user but leaves status nil pins this as intended. It is a reasonable way to stop retrying genuinely invalid usernames every day, but WikiApi#query returns nil for a baduser error and for a timeout alike, so a transient error that lands mid-fallback also freezes that user as unknown. Distinguishing the two would need the error code surfaced through WikiApi, or username validation before querying. Worth a deliberate decision rather than a fix necessarily.

I've updated the fallback query logic in the worker to distinguish between permanent API errors and transient network timeouts.
Now:
- MediawikiApi::ApiError (confirmed bad/deleted usernames) -> stamped with checked_at (and retained_after_course: nil) so they leave the eligible pool and aren't retried every day.
- Transient network timeouts / 5xx errors -> left completely untouched (checked_at stays nil), so they automatically get retried on the next daily run.

ragesoss (changes requested):
Re-review of 2a98f78, again with the help of Claude Code. Same note as before on how to use this feedback: it is not to be blindly trusted — read it, check the reasoning, and use your own judgment.

**What was checked locally:** the branch was checked out and merged with current master (clean), `yarn build` run (clean), RuboCop run on all 14 changed Ruby files and ESLint on the 5 changed JSX files (both clean), the PR's specs plus the other affected specs run (77 examples, 0 failures), the `baduser` response shape re-probed against live en.wikipedia.org, and five new exploratory examples written and run (included at the end).

**Prior findings:** the circuit breaker (observation 2) is in, `order(:end, :id)` resolves the tiebreaker note, and the pre-registration labelling ask is done. Observation 1 — separating a confirmed bad username from a transient failure — is implemented correctly *in the per-user fallback*, but finding 1 below is that the fallback never gets reached in production.

## To answer the backfill sizing question

Peony has roughly **25,000 courses**. Two consequences worth designing around:

- At `sleep 0.2` between courses plus at least one API round trip each, `rake retained_editors:backfill` is on the order of **four or more hours** in a single foreground process — so it wants `nohup`/`screen`, and it needs to exit cleanly rather than by exception (finding 2).
- At the 50-course daily budget, the daily worker alone would need about **500 days** to cover 25,000 courses. So the backfill is genuinely required, not a nice-to-have, which is worth stating in the PR description as well as the code comment.

That number also makes design item 8 from the first review (no index) a decision worth making now rather than later: `eligible_course_ids` scans `courses_users` joined to `courses` and `users` on every daily run, and the unlimited call in the rake task materialises up to 25,000 ids.

## Finding 1 — the `baduser` handling added in 2a98f78 is unreachable, and a bad username kills the job

Re-probed against the live API: a `baduser` response is **HTTP 200** with a `mediawiki-api-error: baduser` header, so `MediawikiApi::Client#send_request` raises `ApiError` (the `status >= 400` branch is never taken).

The per-user fallback intercepts that correctly. The problem is that the **batch** query runs first and passes no interception block, so the `ApiError` falls through to `WikiApi#mediawiki`'s default handling: three retries, then

```ruby
extra = rate_limit_sentry_extra(e)   # -> retry_after_seconds(e)
raw = error.response.headers['Retry-After']
```

`ApiError#response` is a `MediawikiApi::Response`, which `def_delegators` only `:status` and `:success?`. So this raises

```
NoMethodError: undefined method 'headers' for an instance of MediawikiApi::Response
```

and `retry_after_seconds` rescues only `ArgumentError, TypeError`, so it escapes `WikiApi` entirely. It propagates through `query_usercontribs` → `fetch_batch_active_usernames` → `process_batch` → `check_course_new_editors` → `perform`, and the Sidekiq job dies. Nothing is written, the per-user fallback never runs, and the same bad username re-poisons the batch tomorrow — the original finding 2 from the first review, now as a job crash rather than a silent zero.

Reproduced below (`a real baduser ApiError reaching the batch query`), and the second example shows the intended behaviour does work once the error reaches the fallback.

**The root cause is not in this PR.** `retry_after_seconds` is maintainer code in `lib/wiki_api.rb`, added in e62075409 (April) and already deployed to both `production` and `wmflabs`; it is latent today only because existing callers rarely produce HTTP-200 API errors, and there is no matching Sentry issue in the last 90 days. A user-supplied multi-valued `ucuser` batch is exactly the call that does produce them. Two separable pieces of work:

- `lib/wiki_api.rb` should guard that header access — it affects every caller, and is arguably a separate PR.
- Independently, this worker can stop relying on `ApiError → nil` by passing the same interception block to the batch query and rescuing `MediawikiApi::ApiError` around it, falling back to per-user queries. That also removes three wasted HTTP attempts and a Sentry event per poisoned batch.

## Finding 2 — the backfill rake task raises on the new nil return

`check_course_new_editors` gained a `nil` return in eef1cfa as the circuit-breaker signal. `perform` handles it, but `lib/tasks/retained_editors.rake:29` still does:

```ruby
count = worker.check_course_new_editors(course)
total_records += count
```

so an API outage mid-backfill ends the run with `TypeError: nil can't be coerced into Integer` instead of the intended graceful stop. Reproduced below (`the backfill rake task and the new nil return`).

Progress is written per batch and eligibility self-advances, so a re-run resumes rather than redoing work — but over a four-hour run against 25,000 courses this is the difference between a clean stop with a progress line and a stack trace. `backfill` should treat `nil` the same way `perform` does: log and break.

The existing backfill spec stubs `check_course_new_editors` to return `1`, so it cannot catch this. A spec that lets the real method return `nil` would.

## Finding 3 — the ApiError/HttpError distinction has no real test coverage

`spec/workers/retained_editor_check_worker_spec.rb` stubs `WikiApi#query` itself and, inside the stub, calls the interception block and discards its return value before raising unconditionally:

```ruby
elsif block
  err = MediawikiApi::ApiError.new(double(data: { 'code' => 'baduser' }))
  block.call(err) # triggers error interception
  raise err
```

So the outcome is decided by the stub, not by the code under test. Verified by mutation: replacing the predicate in the worker with `{ |_e| false }` leaves all 14 examples in that file passing. The same is true of the sibling "transient network error" context — nothing there distinguishes the two paths either.

Stubbing `MediawikiApi::Client#query` instead runs the real `raise if caller_handles&.call(e)` path; the first three exploratory examples below do that, and they are what actually pin the distinction (and what surfaced finding 1).

## Finding 4 — the circuit breaker cannot tell one failed user from an outage

`fetch_active_usernames` returns `nil` whenever no user in a batch was queryable, which trips the breaker. On a **one-student course** that condition is a single failed user query. Combined with the new oldest-first ordering, one persistently unqueryable user in the oldest eligible course stops the whole daily run at that course, every day, and no later course is ever reached.

Reproduced below (`circuit breaker granularity on a one-student course`): the API is healthy for everyone except one student in the oldest course, and a perfectly healthy second course is never touched.

Tracking consecutive whole-course failures, or requiring some minimum number of failed queries before tripping, would keep the guard without letting one record wedge the queue.

## Still open from earlier reviews

- **Item 7 (first review)** — the two `retained_new_editors` columns still mean different things: `course_csv_builder.rb:125` computes it live via `RetainedNewEditorsStats` with an open-ended window, while `system_csv_builder.rb` reads the frozen flag. Same header, different metric. This was one of the two items flagged as wanting a fix.
- **The 60-day provenance** — the answer that it came from direct communication with WMF and has no public documentation is useful, and is exactly the thing that should live in `new_editor_date_conditions.rb` rather than only in a PR comment. That file is currently untouched by this PR and its comment still reads just `(WMF definition)`.
- Design items 6, 8 and 9 from the first review remain maintainer calls.

## Minor

- `process_batch` returns `batch.size` unconditionally, but in the per-user fallback a transiently-failed user is deliberately left unwritten. So "Processed N records" overstates what was actually recorded.
- The facilitator leaderboard's "New Editors" column header reuses `system_stats.kpis.new_editors`, which is now the much longer "New Editors (60-day pre-registration)". Worth eyeballing how that wraps in a table header — and the PR still has no screenshots of the new KPI tile, leaderboard column or wiki-breakdown column.

<details>
<summary>Exploratory spec used for findings 1, 2 and 4 (click to expand)</summary>

Generated by Claude Code during re-review and run locally against this branch merged with master — all five examples pass, which is what confirms the behaviours described above. RuboCop-clean, but not line-by-line reviewed by a human and not proposed for merging as-is. Save as `spec/workers/retained_editor_rereview3_exploratory_spec.rb`.

```ruby
# frozen_string_literal: true

require 'rails_helper'
require 'rake'
require_dependency "#{Rails.root}/app/workers/retained_editor_check_worker"

describe 'PR #7009 third re-review exploratory checks' do
  let(:course) do
    create(:course, start: 90.days.ago, end: 70.days.ago, home_wiki: wiki, private: false)
  end

  def new_editor_in(a_course, username:)
    user = create(:user, username:, registered_at: a_course.start + 1.day)
    create(:courses_user, course: a_course, user:,
                          role: CoursesUsers::Roles::STUDENT_ROLE,
                          retained_after_course: nil)
  end

  def record_for(username)
    CoursesUsers.joins(:user).find_by(users: { username: })
  end

  # Exactly what the gem builds for a baduser response: HTTP 200 plus a
  # `mediawiki-api-error` header, verified against live en.wikipedia.org.
  def real_baduser_error
    body = { 'error' => { 'code' => 'baduser',
                          'info' => 'Invalid value "Bad#Name" for user parameter "ucuser".' } }
    faraday = Faraday::Response.new(status: 200, body: body.to_json,
                                    response_headers: { 'mediawiki-api-error' => 'baduser' })
    MediawikiApi::ApiError.new(MediawikiApi::Response.new(faraday, ['error']))
  end

  # The PR's own spec stubs WikiApi#query and invokes the interception block
  # itself, so it never exercises WikiApi#mediawiki's real
  # `raise if caller_handles&.call(e)` path. These examples stub the underlying
  # MediawikiApi::Client instead, so the real WikiApi error handling runs.
  describe 'a real baduser ApiError reaching the batch query' do
    before do
      new_editor_in(course, username: 'Editor one')
      new_editor_in(course, username: 'Bad#Name')
    end

    it 'crashes the whole run instead of falling back to per-user queries' do
      allow_any_instance_of(MediawikiApi::Client).to receive(:query) { raise real_baduser_error }

      # WikiApi#retry_after_seconds calls error.response.headers, but
      # MediawikiApi::Response delegates only :status and :success?.
      expect { RetainedEditorCheckWorker.new.perform }
        .to raise_error(NoMethodError, /undefined method 'headers'/)

      expect(CoursesUsers.where.not(retained_after_course_checked_at: nil).count).to eq(0)
    end
  end

  describe 'a real baduser ApiError reaching only the per-user fallback' do
    before do
      new_editor_in(course, username: 'Editor one')
      new_editor_in(course, username: 'Bad#Name')
    end

    it 'stamps checked_at with nil status for the invalid user, on one attempt' do
      calls = []
      allow_any_instance_of(MediawikiApi::Client).to receive(:query) do |_i, params|
        calls << params[:ucuser]
        raise MediawikiApi::HttpError, 503 if params[:ucuser].size > 1 # batch fails transiently
        raise real_baduser_error if params[:ucuser] == ['Bad#Name']
        MediawikiApi::Response.new(
          Faraday::Response.new(status: 200,
                                body: { 'query' => { 'usercontribs' => [] } }.to_json),
          %w[query]
        )
      end

      RetainedEditorCheckWorker.new.perform

      # 3 batch attempts (HttpError is retried), then 1 attempt per user:
      # the ApiError one is intercepted on the first try, no retry, no Sentry.
      expect(calls.count { |u| u == ['Bad#Name'] }).to eq(1)
      expect(record_for('Bad#Name').retained_after_course).to be_nil
      expect(record_for('Bad#Name').retained_after_course_checked_at).not_to be_nil
      expect(record_for('Editor one').retained_after_course).to be(false)
      expect(RetainedEditorCheckWorker.eligible_course_ids).not_to include(course.id)
    end
  end

  describe 'a real transient HttpError' do
    before do
      new_editor_in(course, username: 'Editor one')
      new_editor_in(course, username: 'Editor two')
    end

    it 'leaves every record unchecked and trips the circuit breaker' do
      calls = 0
      allow_any_instance_of(MediawikiApi::Client).to receive(:query) do
        calls += 1
        raise MediawikiApi::HttpError, 503
      end

      RetainedEditorCheckWorker.new.perform

      # 3 (batch) + 3 + 3 (per user) = 9 HTTP attempts, then the run stops.
      expect(calls).to eq(9)
      expect(CoursesUsers.where.not(retained_after_course_checked_at: nil).count).to eq(0)
      expect(RetainedEditorCheckWorker.eligible_course_ids).to include(course.id)
    end
  end

  # check_course_new_editors gained a nil return in eef1cfa (the circuit-breaker
  # signal), but the backfill rake task still does `total_records += count`.
  describe 'the backfill rake task and the new nil return' do
    before do
      Rake.application.rake_require('retained_editors', [Rails.root.join('lib/tasks').to_s])
      Rake::Task.define_task(:environment)
      new_editor_in(course, username: 'Editor one')
    end

    after { Rake::Task['retained_editors:backfill'].reenable }

    it 'raises instead of stopping when the API is unreachable' do
      allow_any_instance_of(WikiApi).to receive(:query).and_return(nil)

      expect { Rake::Task['retained_editors:backfill'].invoke }
        .to raise_error(TypeError, /nil can't be coerced/)
    end
  end

  # The circuit breaker trips when fetch_active_usernames returns nil, which
  # happens whenever no user in a batch could be queried. On a one-student
  # course that is a single failed user query, not an outage.
  describe 'circuit breaker granularity on a one-student course' do
    let!(:older_course) do
      c = create(:course, slug: 'School/Older_(term)', start: 130.days.ago,
                          end: 110.days.ago, home_wiki: wiki, private: false)
      new_editor_in(c, username: 'Lone student')
      c
    end

    before { new_editor_in(course, username: 'Healthy student') }

    it 'stops the run and skips a healthy course when one user query fails' do
      allow_any_instance_of(WikiApi).to receive(:query) do |_i, params|
        if params[:ucuser].include?('Lone student')
          nil
        else
          instance_double(MediawikiApi::Response, data: { 'usercontribs' => [] })
        end
      end

      RetainedEditorCheckWorker.new.perform

      # older_course sorts first under order(:end, :id); its only student fails,
      # so the run breaks and the newer, healthy course is never touched.
      expect(record_for('Lone student').retained_after_course_checked_at).to be_nil
      expect(record_for('Healthy student').retained_after_course_checked_at).to be_nil
      expect(RetainedEditorCheckWorker.eligible_course_ids).to include(course.id)
    end
  end
end
```

</details>

---

*Drafted in a Claude Code session (~40 min, 3 interactions). Sage requested the re-review, supplied the production course count used in the backfill sizing above, decided this should be request-changes, and approved posting after reading the draft — but did not independently verify the findings. Everything described as "verified" or "reproduced" was verified by Claude Code, not by a human: the branch was checked out and merged with master, `yarn build`, RuboCop, ESLint and the affected spec files were run locally, the `baduser` HTTP status and header were probed against live en.wikipedia.org, the `MediawikiApi::Response` delegation and `NoMethodError` were reproduced with a realistic error object, finding 3 was checked by temporarily mutating the predicate in a local working copy, and the five-example exploratory spec above (AI-generated, not line-by-line reviewed by a human) was written and run. Sentry was searched for an existing occurrence of the `headers` NoMethodError and found none, which is weak evidence rather than proof that the path is currently cold.*

ragesoss:
Finding 1 from the review above is a pre-existing bug in `lib/wiki_api.rb`, not in this branch, so it is now tracked separately as #7033 — with the mechanism, a failing spec, and a verified one-line fix.

Two things that follow for this PR:

- The `MediawikiApi::ApiError` handling added in 2a98f78 is the right shape; it just cannot take effect on the batch query until #7033 is fixed. No need to work around it here.
- Passing the interception block to the batch query as well is still worth doing on its own merits: it avoids three wasted HTTP attempts and a Sentry event per batch that contains an invalid username, and it sidesteps #7033 entirely.

ragesoss (changes requested):
Re-review of ec8bea2c, again with the help of Claude Code. Same note as before on how to use this feedback: it is not to be blindly trusted — read it, check the reasoning, and use your own judgment.

**What was checked locally:** the branch was checked out and merged with current master (clean), `yarn build` run (clean), RuboCop run on the 20 changed Ruby files and ESLint on the 5 changed JSX files, the PR's specs plus the other affected specs run (81 examples, 0 failures), the ApiError coverage re-checked by mutation, and five new exploratory examples written and run (included at the end).

**On CI:** the two failures are `de_userfying_edit_alert_monitor_spec` (live log query returns nil) and `upload_importer_spec` (Commons thumbnail host is now `thumb.wikimedia.`). Neither is touched by this branch; both read as upstream drift.

**Prior findings, resolved:**

- The `baduser` path is fixed twice over — the batch query now passes the interception block and rescues `ApiError` into the per-user fallback, and master separately picked up the `retry_after_seconds` guard from #7043. Nice turnaround on #7033.
- The backfill rake task now breaks cleanly on the `nil` return, with a spec.
- The ApiError/HttpError coverage is now real. The specs stub `MediawikiApi::Client#query` instead of `WikiApi#query`, so the actual `raise if caller_handles&.call(e)` path runs. Re-running the mutation check from last time — replacing the predicate with `{ |_e| false }` — now fails an example, where previously all 14 passed.
- The 60-day provenance is recorded in `NewEditorDateConditions`.
- `process_batch` no longer overstates the records it wrote.

## Finding 1 — RuboCop fails on the branch, and CI is hiding it again

```
app/workers/retained_editor_check_worker.rb:107:101: C: Layout/LineLength: Line is too long. [110/100]
app/workers/retained_editor_check_worker.rb:144:101: C: Layout/LineLength: Line is too long. [101/100]
```

Both introduced by ec8bea2c. This is the same mechanism as finding 1 of the first review: the `Ruby linting` step runs after rspec in the workflow, rspec exited on the two unrelated failures above, so the lint step was skipped. The red X is again covering a second, separate failure.

Worth questioning ec8bea2c more broadly. Beyond causing the offenses, it semicolon-chains statements (`threshold = course.end + DAYS_AFTER_END.days; checked_count = 0`, and four assignments on line 107) and deletes the explanatory comments — including the `next_page_params` docstring and the DEPLOY NOTE's rationale about the 50-course daily budget. That last one is the context finding 2 depends on, so it is the comment least worth losing.

## Finding 2 — the new metrics read "not yet checked" as "not retained"

All four readers filter on `retained_after_course: true`:

- `SystemStatUpdateWorker#compute_retained_new_editors_count` → the Retained Editors KPI tile
- `SystemStatUpdateWorker#compute_retained_editors_by_wiki` → the wiki-breakdown column
- `FacilitatorStatUpdateWorker#compute_retained_editor_counts` → the leaderboard column
- `SystemCsvBuilder#fetch_retained_editor_counts` → the CSV column

Every production row starts at `nil`, and `nil` is indistinguishable from `false` in each of those queries. So from deploy until the backfill finishes, all four report 0 — as a plausible small number that grows daily, not as anything that looks broken. Nothing in the UI distinguishes "no editors were retained" from "these courses have not been checked yet."

The part that does not wash out once the backfill lands: `system_stats` rows are permanent dated snapshots (unique index on `snapshot_date`), and the page charts month-over-month deltas through `calculate_delta`. So the catch-up gets written into history as retention growth that never happened, and stays there.

Using last review's sizing — roughly 25,000 courses on Peony, four-plus hours for the backfill, and about 500 days for the 50-course daily budget to cover the same ground on its own — this is a long window, not a deploy-day blip.

Reproduced in the exploratory spec below. The first three examples show each reader returning 0 before the worker runs and the true count after, with the same courses, the same students and the same day; the fourth shows the month-over-month delta attributing the entire backfilled cohort to the month the backfill ran.

Options worth weighing, none of them obviously correct:

- Sequence the rollout: run the backfill before the stat workers start writing snapshots, and before the UI ships.
- Surface coverage — how many eligible courses have been checked — next to the metric, so a partial number is legible as partial.
- Have the readers exclude unchecked courses rather than counting them as zero, so the number is honest about its denominator.

Also worth noting: the PR's own specs seed `retained_after_course` directly as `true` or `false`, so nothing exercises the `nil` state that every production row actually starts in. That is why this did not show up in the suite.

## Finding 3 — the circuit breaker is better, but still wedgeable

`MAX_CONSECUTIVE_FAILURES = 3` plus `next` instead of `break` means a single unqueryable user no longer stops the whole daily run, which was the substance of last review's finding 4. What remains: with `order(:end, :id)`, three consecutive oldest courses each holding a permanently unqueryable user will stop the run at the same place every day, and no later course is ever reached. A failing course is also never marked, so each daily run re-pays those courses' full retry cost before moving on.

Lower severity than before, and possibly acceptable — but the failure mode is unchanged in kind, just three times harder to hit.

## Still open from earlier reviews

- **Item 7 (first review)** — the two `retained_new_editors` columns still mean different things. `course_csv_builder.rb:124` computes it live via `RetainedNewEditorsStats` with an open-ended window that keeps growing; `system_csv_builder.rb:192` reads the frozen point-in-time flag. Same header, two metrics, diverging further the older a course gets — and the frozen one reads 0 for anything unchecked, per finding 2.
- **Design item 8 (no index)** — `retained_after_course` and `retained_after_course_checked_at` are both unindexed, and all four new query paths filter on them. `eligible_course_ids` scans `courses_users` joined to `courses` and `users` on every daily run, and the unlimited call in the rake task materialises up to 25,000 ids. Still a maintainer call, but the numbers above make it a decision worth making now.
- Design items 6 and 9 remain maintainer calls.

## Migration safety

Both migrations are additive `add_column` — nullable with no default on `courses_users`, integer with a literal default on the two small snapshot tables. No backfill, no index, no NOT NULL. That is the safe shape. `courses_users` is large on Peony, so the one thing left to confirm is that both `production` and `peony-database` run MariaDB 10.3+, which makes these metadata-only under `ALGORITHM=INSTANT`. CI runs 10.5.29; the production versions have not been checked here. If they are 10.3+, this is safe inside a normal deploy.

## Minor

- The PR description still describes `rake retained_editors:backfill` as a historical backfill without saying it is required rather than optional, or how long it takes. Given finding 2, that belongs in the description.
- Still no screenshots of the new KPI tile, the leaderboard column or the wiki-breakdown column. The leaderboard's "New Editors" header now renders as "New Editors (60-day pre-registration)", which is worth eyeballing for wrapping in a table header.
- `check_course_new_editors` discards `checked_count` when a later batch in the same course returns `nil`, so records already written by earlier batches are not counted. The data is written and the next run picks up the rest, so this is accounting only.

<details>
<summary>Exploratory spec for finding 2 (click to expand)</summary>

Generated by Claude Code during this re-review and run locally against this branch merged with master — all five examples pass, which is what confirms the behaviour described above. RuboCop-clean, but not line-by-line reviewed by a human and not proposed for merging as-is. Save as `spec/workers/retained_editor_rereview4_exploratory_spec.rb`.

```ruby
# frozen_string_literal: true

require 'rails_helper'
require_dependency "#{Rails.root}/app/workers/retained_editor_check_worker"
require_dependency "#{Rails.root}/app/workers/system_stat_update_worker"
require_dependency "#{Rails.root}/app/workers/facilitator_stat_update_worker"

describe 'PR #7009 fourth re-review exploratory checks' do

  # A course that ended long enough ago to be eligible, with two students who
  # both registered during the program and both genuinely kept editing after it.
  let!(:course) do
    create(:course, slug: 'School/Retention_(term)', title: 'Retention Course',
                    start: 90.days.ago, end: 70.days.ago,
                    type: 'ClassroomProgramCourse', home_wiki: wiki, private: false)
  end

  let!(:students) do
    %w[Alice Bob].map do |name|
      user = create(:user, username: name, registered_at: 85.days.ago)
      user
    end
  end

  # Both students edited mainspace after course.end + 7 days, so the truth is
  # "2 retained". Stubbing at WikiApi#query is enough here: these examples are
  # about what the *readers* of the column report, not about API error handling.
  def api_says_both_retained
    allow_any_instance_of(WikiApi).to receive(:query).and_return(
      instance_double(MediawikiApi::Response,
                      data: { 'usercontribs' => students.map { |u| { 'user' => u.username } } })
    )
  end

  describe 'the three new readers treat "not yet checked" as "not retained"' do
    it 'reports 0 retained editors in the system stats KPI until the worker has run' do
      SystemStatUpdateWorker.new.perform
      expect(SystemStat.current.retained_new_editors_count).to eq(0)

      api_says_both_retained
      RetainedEditorCheckWorker.new.perform
      SystemStat.current.destroy
      SystemStatUpdateWorker.new.perform

      # Same courses, same students, same day: the only thing that changed is
      # that the backfill ran.
      expect(SystemStat.current.retained_new_editors_count).to eq(2)
    end

    it 'reports 0 in the system CSV retained_new_editors column until the worker has run' do
      column = SystemCsvBuilder::CSV_HEADERS.index('retained_new_editors')

      row = CSV.parse(SystemCsvBuilder.new(filters: {}).generate_csv)
               .find { |r| r[1] == 'Retention Course' }
      expect(row[column]).to eq('0')

      api_says_both_retained
      RetainedEditorCheckWorker.new.perform

      row = CSV.parse(SystemCsvBuilder.new(filters: {}).generate_csv)
               .find { |r| r[1] == 'Retention Course' }
      expect(row[column]).to eq('2')
    end

    it 'reports 0 in the facilitator leaderboard column until the worker has run' do
      instructor = create(:user, username: 'Facilitator', registered_at: 2.years.ago)
      create(:courses_user, course:, user: instructor,
                            role: CoursesUsers::Roles::INSTRUCTOR_ROLE)

      FacilitatorStatUpdateWorker.new.perform
      expect(FacilitatorStat.current.find_by(user: instructor).retained_new_editors_count).to eq(0)

      api_says_both_retained
      RetainedEditorCheckWorker.new.perform
      FacilitatorStat.current.destroy_all
      FacilitatorStatUpdateWorker.new.perform

      expect(FacilitatorStat.current.find_by(user: instructor).retained_new_editors_count).to eq(2)
    end
  end

  # system_stats rows are permanent, dated, one-per-day snapshots, and the
  # System Stats page charts month-over-month deltas of them. A snapshot taken
  # before the backfill completes is therefore not just temporarily low — it is
  # written into history, and the catch-up shows up as growth that never
  # happened.
  describe 'the backfill shows up as a fabricated growth spike in the trend' do
    it 'attributes the whole backfilled cohort to the month the backfill ran' do
      # Last month's snapshot, taken before the backfill.
      SystemStat.create!(snapshot_date: 1.month.ago.to_date, retained_new_editors_count: 0)

      # This month, after the backfill.
      api_says_both_retained
      RetainedEditorCheckWorker.new.perform
      SystemStatUpdateWorker.new.perform

      snapshots = SystemStat.recent_monthly_snapshots(13)
      deltas = snapshots.each_cons(2).map do |prev, curr|
        curr.retained_new_editors_count - prev.retained_new_editors_count
      end

      # No editor was actually retained this month; the entire jump is the
      # backfill filling in a column that read 0 before.
      expect(deltas.last).to eq(2)
    end
  end

  # The PR's own specs all seed retained_after_course directly (true/false), so
  # nothing exercises the nil state that every production row starts in.
  describe 'the nil state is what every production row starts in' do
    it 'is indistinguishable from false in all three count queries' do
      expect(CoursesUsers.where(retained_after_course: nil).count).to eq(2)

      counts = SystemCsvBuilder.new(filters: {})
                               .send(:fetch_retained_editor_counts, [course.id])
      expect(counts[course.id]).to be_nil

      CoursesUsers.update_all(retained_after_course: false)
      counts = SystemCsvBuilder.new(filters: {})
                               .send(:fetch_retained_editor_counts, [course.id])
      expect(counts[course.id]).to be_nil
    end
  end
end
```

</details>

---

*Drafted in a Claude Code session (~25 min, 3 interactions). Sage requested the re-review and decided it should be request-changes, and approved posting after reading the draft — but did not independently verify the findings. Everything described as "verified" or "reproduced" was verified by Claude Code, not by a human: the branch was checked out and merged with master, `yarn build`, RuboCop, ESLint and the affected spec files were run locally, the two RuboCop offenses were reproduced, the ApiError coverage was re-checked by temporarily mutating the interception predicate in a local working copy, and the five-example exploratory spec above (AI-generated, not line-by-line reviewed by a human) was written and run. Production MariaDB versions were not checked; the migration verdict is conditional on them.*

lakshita10341:
> Finding 2 — the new metrics read "not yet checked" as "not retained"

I've updated the PR description to make the backfill a mandatory deployment step. Running it immediately after deployment will prevent incorrect stats from day one.

> Finding 3 — the circuit breaker is better, but still wedgeable

This behavior is intentional based on previous review discussions. On network errors, the worker stops to avoid querying subsequent courses. On permanent API errors, checked_at is updated to prevent the course from being re-queried the following day.

lakshita10341:
>Item 7 (first review) — the two retained_new_editors columns still mean different things. course_csv_builder.rb:124 computes it live via RetainedNewEditorsStats with an open-ended window that keeps growing; system_csv_builder.rb:192 reads the frozen point-in-time flag. Same header, two metrics, diverging further the older a course gets — and the frozen one reads 0 for anything unchecked, per finding 2.

Updating `CourseCsvBuilder` to match `SystemCsvBuilder` is planned for the next step and will be handled in a separate PR after this one is merged.


> Design item 8 (no index) — retained_after_course and retained_after_course_checked_at are both unindexed, and all four new query paths filter on them. eligible_course_ids scans courses_users joined to courses and users on every daily run, and the unlimited call in the rake task materialises up to 25,000 ids. Still a maintainer call, but the numbers above make it a decision worth making now.

Regarding the decision not to add indexes to `retained_after_course` and `retained_after_course_checked_at`: retained_after_course is a boolean column with extremely low cardinality which will make the indexing uneffective. Hence, I have not added it.

ragesoss (approved):
Re-review of 90eb004, again with the help of Claude Code. Same note as before on how to use this feedback: it is not to be blindly trusted — read it, check the reasoning, and use your own judgment.

**What was checked locally:** the branch was checked out and merged with current master (clean), `yarn build` run (clean), RuboCop run on the 21 changed Ruby files and ESLint on the 5 changed JSX files (both clean), the PR's specs plus the other affected specs run (107 examples, 0 failures), and the five-example exploratory spec from the last review re-run (5 pass, meaning the behaviour it describes is unchanged, as expected since no code path changed).

**Since ec8bea2c:** two commits touching the branch's own code, both formatting-only. The semicolon chains are gone, the two RuboCop line-length offenses are fixed, and CI is green with the lint step actually running this time. The `loop`/`break` in the batch fetcher became `while pending`; that reads as equivalent and the pagination specs still pass. No behavioural change was found in the delta.

**Prior findings:**

- Finding 1: resolved. One correction to the last review: the explanatory comments were dropped in 3eca7da, not ec8bea2c (16 comment lines went to 6 there). Either way they are still absent. The two most worth restoring are the `next_page_params` return-value docstring and the DEPLOY NOTE's reason the backfill is mandatory — that the 50-course daily budget would otherwise be spent on historical courses for months.
- Finding 3: the reply holds up against the code. A `MediawikiApi::ApiError` on a user marks `checked_at` with a nil status, so a permanently unqueryable user does not wedge the run; only a nil return from `WikiApi#query` (network, exhausted retries) counts toward the three-failure stop, and users that hit that stay unchecked for the next run. Both paths are covered by the PR's own specs. Treating as resolved.
- Item 8 (index): the no-index call is defensible, though not quite for the stated reason. `EXPLAIN` against a local copy of Wiki Ed production data shows `eligible_course_ids` is driven by a scan of `courses` (~8.7K rows), then reaches `courses_users` through the existing `(course_id, user_id, role)` index; `retained_after_course_checked_at IS NULL` is a post-filter on rows already fetched by index. An index on either new column would not change that plan. The daily cost is a pass over the ~160K student rows of ended public courses, which is fine.
- Item 7 (`CourseCsvBuilder` divergence): deferred to a follow-up PR per the reply. Worth opening that issue now so it does not get lost, and worth a line in the PR description that the two `retained_new_editors` columns mean different things until then.

**Finding 2 — the remaining decision.** No code changed; the response is the deployment note making the backfill mandatory. That can work, because both snapshot writers upsert by date (`system_stats` on `snapshot_date`, `facilitator_stats` on `(snapshot_date, user_id)`), so a snapshot taken before the backfill finishes can be corrected by re-running the stat worker the same day. Sizing from Wiki Ed production data: 7,403 eligible courses and roughly 145K new-editor records. At 50 courses a day that is about 150 days; the backfill itself (batches of 40, roughly two API calls per batch, 0.2 s per course) should be on the order of one to three hours for Wiki Ed and longer for Peony. If the sequencing route is taken, the deploy checklist for each dashboard should be explicit in the PR description: deploy, run the backfill to completion (in `screen`/`tmux`, since it is a multi-hour foreground task; re-run if it stops on an API outage), then re-run `SystemStatUpdateWorker` (and `FacilitatorStatUpdateWorker` if a Sunday snapshot landed in between) so the day's snapshot reflects the backfilled data. The alternative — having the readers exclude unchecked courses, or surfacing checked-versus-eligible coverage next to the number — remains available if a process-only control turns out not to be enough.

**Migration safety:** Wiki Ed production was MariaDB 10.5.15 as of a June 2026 dump header, so both `add_column`s qualify for `ALGORITHM=INSTANT` there and are safe inside a normal deploy. Peony's version is still unconfirmed.

**Minor, unchanged:** no screenshots of the KPI tile, leaderboard column or wiki-breakdown column; `check_course_new_editors` still discards `checked_count` when a later batch returns nil (accounting only).

---

*Drafted in a Claude Code session (~30 min, 2 interactions). Sage requested the re-review and approved posting after reading the draft, but did not independently verify the findings. Everything described as checked was checked by Claude Code, not by a human: the branch was checked out and merged with master, `yarn build`, RuboCop, ESLint and the affected spec files were run locally, the prior exploratory spec was re-run, the circuit-breaker paths were traced in the worker source, and the row counts and `EXPLAIN` were taken from a local copy of Wiki Ed production data. Peony's MariaDB version was not checked.*

ragesoss:
Deploy checklist for the mandatory backfill, to go with the Deployment Note in the description (worth copying it in there). Same caveat as the reviews: drafted with Claude Code from the repository and the review findings, so check it against the actual environment before relying on it.

Both dashboards run `DailyUpdateWorker` at 11:30 UTC (sidekiq-cron, `daily_update` queue). It runs `RetainedEditorCheckWorker` with its 50-course budget and then `SystemStatUpdateWorker`, which upserts that UTC day's `system_stats` row. `FacilitatorStatUpdateWorker` runs Sundays at 02:00 UTC and upserts by `(snapshot_date, user_id)`. Because both upsert, a snapshot taken before the backfill finishes can be corrected by re-running the worker the same UTC day.

**Per dashboard** (Wiki Ed: cap stage `production`, branch `production`, `dashboard.wikiedu.org`. P&E: cap stage `peony`, branch `wmflabs`, `peony-web`):

1. **Confirm the MariaDB version** is 10.3 or later, so the two `add_column` migrations are metadata-only (`ALGORITHM=INSTANT`). Wiki Ed was 10.5.15 as of June 2026. Peony is unconfirmed — check `peony-database` with `SELECT VERSION();` before deploying.
2. **Pick the window.** Deploying shortly after that day's 11:30 UTC run has finished leaves roughly 24 hours before the next automatic snapshot, which is enough for the backfill on either dashboard. If the deploy has to happen before 11:30 UTC, expect that day's snapshot to read low and fix it in step 6.
3. **Deploy.** Capistrano runs the migrations. On P&E, also `git pull` + `bundle install` + restart on `peony-sidekiq`, `peony-sidekiq-medium` and `peony-sidekiq-3`, which Capistrano does not deploy, so every Sidekiq process runs the same code. The `daily_update` queue itself runs on `peony-web`, which Capistrano covers.
4. **Run the backfill** on the web server from the current release, inside `screen`/`tmux` since it is a multi-hour foreground task:
   ```
   RAILS_ENV=production bundle exec rake retained_editors:backfill
   ```
   It prints progress every 10 courses. Expected scale: Wiki Ed about 7,400 courses and 145K records, on the order of 1–3 hours; P&E roughly 25,000 courses, 4+ hours. If it stops with `MediaWiki API appears down, stopping backfill. Re-run to resume.`, re-run it; checked rows are skipped, so it resumes where it left off. Overlap with the 11:30 UTC daily run is harmless: both only touch rows with `retained_after_course_checked_at IS NULL`, and the writes are idempotent.
5. **Verify completion** in a production console:
   ```ruby
   RetainedEditorCheckWorker.eligible_course_ids.size                    # => 0
   CoursesUsers.where.not(retained_after_course_checked_at: nil).count   # ~145K on Wiki Ed
   ```
6. **Re-snapshot if needed.** If a daily snapshot was written that UTC day before the backfill finished, run `SystemStatUpdateWorker.new.perform` the same UTC day to overwrite it. If a Sunday 02:00 UTC facilitator snapshot landed between deploy and completion, also run `FacilitatorStatUpdateWorker.new.perform`. If the backfill completed before the next 11:30 UTC run, there is nothing to do; that run snapshots the backfilled data on its own.
7. **Sanity check** the System Stats page: the Retained Editors tile is non-zero and plausible against the numbers above, the leaderboard column populates, and the system CSV export's `retained_new_editors` column is non-zero for old courses.

Not covered by this checklist: `CourseCsvBuilder` still computes `retained_new_editors` live with an open-ended window (item 7, deferred to a follow-up PR). Courses ending from now on are picked up by the daily 50-course budget once they pass 30 days; if many courses end in the same week the queue clears over the following days, oldest first.

---

*Drafted in a Claude Code session at Sage's direction, as a follow-up to the approval above. Sage asked for the checklist to be posted and did not review this text before it went out. The schedule times, queue names, deploy stages, rake task behaviour and upsert semantics were read from the repository by Claude Code; the sizing comes from a local copy of Wiki Ed production data and the earlier Peony estimate. Nothing here was executed against production.*

ragesoss:
I've deployed this to both production servers. The backfill is complete for Wiki Education, and about halfway done (with perhaps 2 hours remaining) for Peony.

ragesoss:
Peony backfill also completed! Thanks @lakshita10341 !

lakshita10341:
Thanks @ragesoss! Glad to hear that.

PR #7011, opened 2026-08-25
Add new `report_csv` queue

gabina:
## What this PR does
This PR adds a new `report_csv` queue to put CSV generation and CSV cleanup. This is the last step for #6606.

As @Formasitchijoh [pointed out in Slack](https://wikieducation.slack.com/archives/C724RCXT7/p1787642285228759), `ReportCsvWorker` runs on the `daily_update` queue. Now that we want to move the CSV report generation to another server (I have `peony-sidekiq-medium` in mind) the easiest move is to create a new specific queue for that purpose. **Note this affects both deployments: wiki edu and peony.**

### Steps for peony

1. Copy production credentials as env variables into **1)** `peony-sidekiq-medium` (the server that process the new report_csv queue and creates and deletes the reports) and **2)** `peony-web` (the server that runs the controller code and it has to decide the report URL and check whether the report already exists). Already done :heavy_check_mark: :heavy_check_mark: 
2. Copy the new systemd unit file into `/etc/systemd/system` into `peony-sidekiq-medium` server and enable it (run `sudo systemctl daemon-reload` and `sudo systemctl enable --now sidekiq-report-csv`). I can do that right now, but I'm not sure if we would have to do it anyway manually right after deployment.

### Steps for wiki-edu

1. Copy the new system unit file into  `/etc/systemd/system` into single wiki edu server (using the right `WorkingDirectory` and `ExecStart` definitions) and enable it. I can do that.

## AI usage
Used Claude Code to discuss things.

### DIFF
--- a/app/services/system_metrics.rb
+++ b/app/services/system_metrics.rb
@@ -57,7 +57,8 @@ class SystemMetrics
     'long_update' => 1.day,
     'daily_update' => 1.day,
     'constant_update' => 15.minutes,
-    'acuwt_update' => 1.day
+    'acuwt_update' => 1.day,
+    'report_csv' => 15.minutes
   }.freeze
 
   def get_queue_status(queue_name, latency)
--- a/app/workers/csv_cleanup_worker.rb
+++ b/app/workers/csv_cleanup_worker.rb
@@ -3,6 +3,7 @@ require_dependency "#{Rails.root}/lib/analytics/report_csv_store"
 
 class CsvCleanupWorker
   include Sidekiq::Worker
+  sidekiq_options queue: 'report_csv'
 
   def perform(filename)
     ReportCsvStore.delete filename
--- a/app/workers/report_csv_worker.rb
+++ b/app/workers/report_csv_worker.rb
@@ -14,7 +14,7 @@ require_dependency "#{Rails.root}/lib/analytics/system_stats_csv_builder"
 
 class ReportCsvWorker
   include Sidekiq::Worker
-  sidekiq_options queue: 'daily_update', lock: :until_executed
+  sidekiq_options queue: 'report_csv', lock: :until_executed
 
   # Generate the csv for the given source (course or campaign)
   # if type is global, then can access source as nil
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -148,6 +148,7 @@ namespace :deploy do
       'sidekiq-long', # data updates for long-running courses, which may have long queue latency
       'sidekiq-daily', # once-daily long-running data update tasks
       'sidekiq-constant', # frequently-run tasks like adding courses to the update queues
+      'sidekiq-report-csv', # user-requested CSV reports
     ]
   end
   set :sidekiq_roles, -> { :app }
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1573,6 +1573,9 @@ en:
     purpose: Purpose
     queues_overview: Queues Overview
     queue: Queue
+    report_csv: Report CSV generation
+    report_csv_description: Specific queue for generating user-requested CSV reports.
+    report_csv_doc: Handles user-requested CSV reports.
     short_update: Short update
     short_update_description: Short updates process courses with fewer than 1,000 revisions.
     short_update_doc: Handles updates for small courses.
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -29,3 +29,4 @@ test:
  - constant_update
  - very_long_update
  - acuwt_update
+ - report_csv
--- a/server_config/systemd/sidekiq-report-csv.service
+++ b/server_config/systemd/sidekiq-report-csv.service
@@ -1,0 +1,38 @@
+# Based on https://github.com/mperham/sidekiq/blob/master/examples/systemd/sidekiq.service
+#
+# A process dedicated to the report_csv queue: user-requested CSV reports and the
+# cleanup jobs that delete them a week later. These are the only user-facing jobs
+# in the system, so they get their own queue instead of waiting behind course
+# updates. (The host running this needs the report_csv_* env vars.)
+
+[Unit]
+Description=sidekiq-report-csv
+After=syslog.target network.target
+
+[Service]
+Type=simple
+# On the Capistrano-deployed hosts this is /var/www/dashboard/current instead,
+# with User=ragesoss and Group=project-globaleducation.
+WorkingDirectory=/home/ragesoss/WikiEduDashboard
+ExecStart=/home/ragesoss/.rvm/bin/rvm default do bundle exec sidekiq -e production --queue report_csv --concurrency 1
+User=root
+Group=root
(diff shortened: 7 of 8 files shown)


ragesoss (approved):
This review was produced in a Claude Code session. AI-generated review feedback should not be trusted automatically or acted on blindly — it is worth reading and understanding, then using or ignoring based on your own judgment. Some of what follows may be wrong.

The change is complete and correct as far as the session could determine. Verified locally against a merge with current master: RuboCop clean on the changed Ruby files, `yarn build` succeeds, and 83 examples pass across the affected area (`system_metrics_spec`, `reports_controller_spec`, `system_csv_spec`, `system_daily_stats_csv_spec`, the three `ReportCsvStore` specs, and `sidekiq_cron_jobs_spec`), plus `spec/features/system_status_spec.rb`, which confirms `/status` renders the new row and stays axe-clean.

A few things checked that are worth recording, since none are obvious from the diff:

- The `sidekiq.yml` hunk header reads `@@ test:`, but the line is appended to the top-level `:queues:` list, which is what `SystemMetrics` reads. Parsed to confirm.
- Adding `sidekiq-report-csv` to `sidekiq_processes` in `deploy.rb` is safe on peony even though the process will live on `peony-sidekiq-medium`, because every systemctl call in the `sidekiq` namespace passes `raise_on_non_zero_exit: false` — the same reason `sidekiq-short`/`medium`/`long` already no-op on `peony-web`.
- The transition leaves no orphaned jobs. Queue names live in the Redis payload, so in-flight `ReportCsvWorker` jobs stay on `daily_update` and cleanup jobs scheduled up to a week out stay on `default`; both processes remain in the list.

## 1. Enabling the units before the deploy is safer than after

This answers the open question in step 2. Once this deploys, `csv_of` enqueues to `report_csv` whether or not anything consumes it, and the failure is silent — the user keeps seeing "This file is being generated. Please try again shortly." with nothing logged. Because the queue name is only a CLI argument, the unit can be installed and started on both hosts *before* the deploy, against the currently-deployed code, where it will sit harmlessly on an empty queue. The 15-minute threshold does mean `/status` would flag it eventually, which is a reasonable backstop.

## 2. `peony-sidekiq-medium` needs to be pulled to at least #7006 first

That host is updated by hand rather than by Capistrano. On pre-#7006 code every `report_csv` job would raise `NameError: uninitialized constant ReportCsvStore`, and without `aws-sdk-s3` (added in 70d878d52) the client can't be constructed either. From the user's side that looks identical to the failure above, so it is worth confirming the pull and `bundle install` happened before the unit is enabled.

## 3. Memory co-tenancy is worth watching after enabling

`peony-sidekiq-medium` already runs medium and short course updates, and CSV generation is the memory-heavy path that #6991 targeted. `--concurrency 1` looks like the right call for exactly that reason. A large system CSV running alongside a medium course update on one VM is the scenario that work was about, so RSS on that host is worth a look once the queue is live.

## 4. Nothing pins the `sidekiq.yml` ↔ `LATENCY_THRESHOLDS` invariant

This PR keeps them in sync. But `get_queue_status` does `latency < LATENCY_THRESHOLDS[queue_name]`, so the next queue added to `sidekiq.yml` without a matching threshold entry takes down `/status` with `ArgumentError: comparison of Integer with nil failed`. Confirmed by spec.

Separately, `spec_helper.rb` sets `Sidekiq.testing!(:inline)`, so no spec in the suite can observe which queue a job routes to — the existing `expect(CsvCleanupWorker).to receive(:perform_at)` assertions pass identically with or without this change.

Both gaps are covered by the following spec, which was written by Claude Code during the review and passes on this branch. Adopt it or ignore it as seems useful:

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/app/workers/report_csv_worker"
require "#{Rails.root}/app/workers/csv_cleanup_worker"

# The suite runs Sidekiq inline (see spec_helper.rb), so no other spec can
# observe which queue a job is routed to. These pin the routing directly.
describe 'report_csv queue routing' do
  it 'routes ReportCsvWorker to the report_csv queue' do
    expect(ReportCsvWorker.sidekiq_options['queue']).to eq('report_csv')
  end

  it 'routes CsvCleanupWorker to the report_csv queue' do
    expect(CsvCleanupWorker.sidekiq_options['queue']).to eq('report_csv')
  end

  it 'declares report_csv among the queues sidekiq is configured to process' do
    queues = YAML.safe_load_file('config/sidekiq.yml', permitted_classes: [Symbol],
                                                       aliases: true)[:queues]
    expect(queues).to include('report_csv')
  end

  it 'has a latency threshold for every queue the system status page reports on' do
    monitored = SystemMetrics.new.instance_variable_get(:@queues)
    expect(monitored - SystemMetrics::LATENCY_THRESHOLDS.keys).to be_empty
  end

  it 'cannot report status for a queue that has no latency threshold' do
    expect { SystemMetrics.new.get_queue_status('queue_with_no_threshold', 0) }
      .to raise_error(ArgumentError)
  end
end
```

## 5. Minor

- The "Open questions and concerns" section still contains the template placeholder.
- Pre-existing from #7006 rather than this PR: `LocalReportCsvStore#delete` uses a bare `File.delete`, which raises `Errno::ENOENT` if the file is already gone. `FileUtils.rm_f` would make cleanup idempotent. Only surfaces now because cleanup retries land on the user-facing queue.

Approving — findings 1 and 2 are deploy sequencing rather than code changes.

*Drafted in a Claude Code session (~20 min, 3 interactions). Sage asked for a full review and approved this comment before posting, but did not independently verify the findings. The session checked out the branch, merged current master, and verified locally: RuboCop on the changed Ruby files, `yarn build`, the 83 affected examples, the `/status` feature spec, and the parsed contents of `sidekiq.yml` and the new `en.yml` keys. Findings 1–3 concern server state that was not inspected — no SSH was performed — and are reasoning from the repo's deploy configuration, not observations of the hosts. The spec above was written by Claude Code and run on this branch; Sage did not review it line by line.*

gabina:
> 1. Enabling the units before the deploy is safer than after

I already copied the right unit files to the servers but didn't enable them. The review is not super clear for me because after point 1 says "it is worth confirming the pull and bundle install happened before the unit is enabled." Anyway, for wiki edu, I think it's going to be started automatically due to the capistrano deployment. I'm not sure if we have to run `bundle install` manually (I understand that's part of the deployment process). Anyway, let me know whenever you want to deploy this and I'll put an eye.

gabina:
Once this gets deployed I should update the **Sidekiq Servers** section in `docs/admin_guide.md`

ragesoss:
I'll deploy. `bundle install` is necessary for the Peony sidekiq servers that don't get capistrano deployment, but not needed manually for Wiki Education production.

gabina:
@Formasitchijoh the changes for moving CSV reports out of the web server (and the `daily_update` queue) is already deployed and working. Users should be able to download their reports pretty quickly now (since we have a specific queue and a specific worker for that).

PR #7016, opened 2026-08-27
Update admin docs with new sidekiq-report-csv

gabina:
## What this PR does
This PR updates some docs. Now that we moved CSV report generation and deletion out of the `daily_update` queue, and out of the `web-server` for peony, we have to update some docs.

`docs/WMCLOUD_SETUP/web_server.md` still says that we shouldn't run daily queue on the web server. Note that, while #6606 got closed, daily queue still runs on the web-server.

>  - Do **not** run `daily` on the web server. It carries one of the heaviest jobs in the system — `ImportRatingsWorker`, which sweeps all articles — and has been observed at over 1 GB resident while competing with the web application for the same RAM. See #6990.


ragesoss:
As long as we keep any heavy lifting out of the daily update, it's probably not critical to get that one off the web server, but I think at this point it doesn't matter which server we use. We could probably even move the default queue off of the web-server. But I don't think any further worker shuffling is urgent right now.

PR #7021, opened 2026-08-30
Fix ARIA compliance for dashboard info tooltip (#6095)

### LINKED ISSUE #6095: ARIA Compliance Issue: Information Tooltip Content Not Recognized by Screen Readers
With reference to #5298 

### What is happening?

When using the NVDA screen reader, the message inside the "i" (information) icon on the Dashboard is not announced. This creates an accessibility barrier for screen reader users, as they cannot access the additional information provided by the tooltip.

--------------------------------------------------------------------------------------------------
### To Reproduce

1. Open the Dashboard page with the "i" information icon.
2. Enable NVDA (or any screen reader) and navigate to the "i" icon.
3. Hover over the icon or navigate to it via keyboard.
4. Observe that the tooltip message is not announced by the screen reader

--------------------------------------------------------------------------------------------------

### Expected behavior
The tooltip message should be announced by the screen reader when the user hovers over or focuses on the "i" icon.

--------------------------------------------------------------------------------------------------

### Screenshots
Screenshot 2025-01-06 033552

image

--------------------------------------------------------------------------------------------------------------------

### Additional context

To install NVDA on Windows:

1. Download NVDA from [official website](https://www.nvaccess.org/download/)
2. Run the installer and follow the on-screen instructions to complete the installation.
3. Once installed, you can start NVDA by pressing Ctrl + Alt + N or by searching for "NVDA" in the Start Menu

MasterIfeanyi:
## What this PR does
Fixes #6095. The information "i" icon on the Dashboard page (rendered by `_open_course_creation.html.haml`) displayed a tooltip on hover, but the tooltip content was never announced by screen readers like NVDA.

Root cause: the tooltip trigger was a non-focusable `<div>`/`` pair with no ARIA relationship to the tooltip text, and the CSS only revealed the tooltip on `:hover`. Keyboard and screen reader users had no way to reach or hear it.

Changes:
- Made the info icon keyboard-focusable (`tabindex="0"`, `role="button"`)
- Added `aria-label` on the icon and `aria-describedby` linking it to the tooltip text
- Gave the tooltip container a unique `id` and `role="tooltip"`
- Added a `:focus-within` rule in `_tooltips.styl` alongside the existing `:hover` rule, so the tooltip also appears for keyboard users
- Changed the icon's `alt` text to empty, since the description is now conveyed via `aria-describedby` instead

## AI usage
I used Claude (Anthropic) throughout this fix. Specifically:
- Asked it to help me locate the relevant files in the codebase, since I wasn't familiar with the project structure
- Asked it to explain the WAI-ARIA pattern for accessible tooltips (focusable trigger + `aria-describedby` + `role="tooltip"`), since I hadn't written accessibility fixes like this before
- Asked it to walk me through Haml syntax line by line, since this was my first time writing Haml
- I wrote and tested the actual code changes myself, and reviewed each suggestion before applying it rather than pasting a generated diff wholesale

- I used `role="button"` on the `` to make it focusable and announce as interactive. If maintainers prefer, this could instead be refactored into a real `<button>` element wrapping the icon, which would be more semantically correct but touches more of the surrounding markup/CSS. Happy to go that route if preferred.

### DIFF
--- a/app/assets/stylesheets/modules/_tooltips.styl
+++ b/app/assets/stylesheets/modules/_tooltips.styl
@@ -5,6 +5,10 @@
     .tooltip
       opacity 1
       top 110%
+  &:focus-within
+  .tooltip
+    opacity 1
+    top 110%
   .tooltip-indicator
     &:after
       font-family 'icons'
--- a/app/assets/stylesheets/modules/_tooltips.styl
+++ b/app/assets/stylesheets/modules/_tooltips.styl
@@ -6,9 +6,9 @@
       opacity 1
       top 110%
   &:focus-within
-  .tooltip
-    opacity 1
-    top 110%
+    .tooltip
+      opacity 1
+      top 110%
   .tooltip-indicator
     &:after
       font-family 'icons'
--- a/app/views/dashboard/_open_course_creation.html.haml
+++ b/app/views/dashboard/_open_course_creation.html.haml
@@ -1,6 +1,6 @@
 %div.instructions.tooltip-trigger
-  %img{:src => "/assets/images/info.svg", :alt => "tooltip default logo"}
-  %div.tooltip.large
+  %img{:src => "/assets/images/info.svg", alt: "", tabindex: "0", role: "button", "aria-label": "More information", "aria-describedby": "course-creation-tooltip"}
+  %div.tooltip.large#course-creation-tooltip{role: "tooltip"}
     %p= t("dashboard.about_programs")
     %br
     %p= t("dashboard.about_creating_programs")
--- a/app/views/dashboard/_open_course_creation.html.haml
+++ b/app/views/dashboard/_open_course_creation.html.haml
@@ -1,5 +1,5 @@
 %div.instructions.tooltip-trigger
-  %img{:src => "/assets/images/info.svg", alt: "", tabindex: "0", role: "button", "aria-label": "More information", "aria-describedby": "course-creation-tooltip"}
+  %img{src: "/assets/images/info.svg", alt: "", tabindex: "0", "aria-label": t("dashboard.tooltip.aria_label"), "aria-describedby": "course-creation-tooltip"}
   %div.tooltip.large#course-creation-tooltip{role: "tooltip"}
     %p= t("dashboard.about_programs")
     %br
--- a/app/views/dashboard/_open_course_creation.html.haml
+++ b/app/views/dashboard/_open_course_creation.html.haml
@@ -1,5 +1,5 @@
 %div.instructions.tooltip-trigger
-  %img{src: "/assets/images/info.svg", alt: t("dashboard.tooltip.aria_label"), tabindex: "0", "aria-describedby": "course-creation-tooltip"}
+  %img{src: "/assets/images/info.svg", alt: t("dashboard.tooltip.alt"), tabindex: "0", "aria-describedby": "course-creation-tooltip"}
   %div.tooltip.large#course-creation-tooltip{role: "tooltip"}
     %p= t("dashboard.about_programs")
     %br
--- a/app/views/dashboard/_open_course_creation.html.haml
+++ b/app/views/dashboard/_open_course_creation.html.haml
@@ -1,5 +1,5 @@
 %div.instructions.tooltip-trigger
-  %img{src: "/assets/images/info.svg", alt: "", tabindex: "0", "aria-label": t("dashboard.tooltip.aria_label"), "aria-describedby": "course-creation-tooltip"}
+  %img{src: "/assets/images/info.svg", alt: t("dashboard.tooltip.aria_label"), tabindex: "0", "aria-describedby": "course-creation-tooltip"}
   %div.tooltip.large#course-creation-tooltip{role: "tooltip"}
     %p= t("dashboard.about_programs")
     %br
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -136,6 +136,8 @@ en:
       If you are running a single, independent event, you can create a
       Program now. If your event is part of something bigger, find or create the
       Campaign for it. Then you can create a new Program from the Campaign page.
+    tooltip:
+      aria_label: More information
   home:
     explore: Explore
     log_in: Log in with Wikipedia
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -137,7 +137,7 @@ en:
       Program now. If your event is part of something bigger, find or create the
       Campaign for it. Then you can create a new Program from the Campaign page.
     tooltip:
-      aria_label: More information
+      alt: More information
   home:
     explore: Explore
     log_in: Log in with Wikipedia


ragesoss (changes requested):
**Note on this feedback**: this review was produced with AI assistance (Claude Code). It is meant to be read, understood, and used or ignored per the developer's own judgment — not automatically trusted.

The ARIA changes in the Haml are well-aimed and likely fix the NVDA announcement on their own, since `aria-describedby` text is computed even for visually hidden elements. However, the CSS half of the change does not work as intended.

**1. The `:focus-within` rule compiles to the wrong selector (confirmed)**

In Stylus, two selectors on consecutive lines at the same indentation form a comma-separated group. As written:

```styl
  &:focus-within
  .tooltip
    opacity 1
    top 110%
```

compiles to:

```css
.tooltip-trigger:focus-within,
.tooltip-trigger .tooltip {
  opacity: 1;
  top: 110%;
}
```

rather than the intended `.tooltip-trigger:focus-within .tooltip`. As a result, no rule reveals the tooltip when the icon receives keyboard focus — the visual half of the fix is inert. (The stray `.tooltip-trigger .tooltip { opacity: 1 }` group member is overridden by the later base rule at equal specificity, so tooltips don't become permanently visible either.) Indenting `.tooltip` one level deeper fixes it:

```styl
  &:focus-within
    .tooltip
      opacity 1
      top 110%
```

This was verified with a browser-based feature spec (below): as written, focusing the icon leaves the tooltip at computed `opacity: 0`; with the corrected indentation, it transitions to 1. One testing note: the tooltip has a 0.2s opacity transition, so any check must wait before reading the computed style.

**2. `role="button"` announces an interactive control that does nothing**

Pressing Enter or Space on the icon has no effect, so announcing it as a button is misleading. For the ARIA tooltip pattern, the trigger only needs to be focusable and carry `aria-describedby`. Note, though, that removing `role="button"` while keeping `alt=""` would leave the image presentational-yet-focusable — also a problem. A cleaner shape: a meaningful `alt` (or `role="img"` with `aria-label`) plus `tabindex="0"` and `aria-describedby`, with no button role. (This overlaps with the open question already raised in the PR description; a real `<button>` wrapper would also resolve it.)

**3. `aria-label` bypasses i18n**

`"More information"` is hardcoded English while the rest of the partial uses `t(...)`. Screen-reader users on non-English locales would hear English. This should come from a locale key.

**Minor observations**
- Mixed Haml attribute syntax on one line (`:src => ...` next to `alt: ...`).
- Once corrected, the `:focus-within` rule applies to every `.tooltip-trigger` site-wide (~30 views/components), including triggers containing buttons/links, which will hold tooltips open while focused. This is probably the desired accessible behavior but is a broader change than the PR description implies.
- The full WAI-ARIA tooltip pattern also calls for Escape-to-dismiss, which a CSS-only approach can't provide; noting for completeness, not as a blocker.

**CI**: the failing build appears unrelated to this change (flaky specs: React duplicate-key warnings in `tickets_dashboard_spec`, live-Wikipedia network dependency in `oauth_spec`); `tickets_dashboard_spec` passes locally on this branch.

<details><summary>Feature spec used for verification (AI-generated during review; ran locally, fails as written / passes with the indentation fix)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe 'dashboard info tooltip accessibility', type: :feature, js: true do
  let(:user) { create(:user) }

  before do
    allow(Features).to receive(:open_course_creation?).and_return(true)
    login_as(user, scope: :user)
    visit root_path
    expect(page).to have_content 'My Dashboard'
  end

  it 'exposes the ARIA relationship between icon and tooltip' do
    icon = find('.instructions.tooltip-trigger img')
    expect(icon['tabindex']).to eq('0')
    expect(icon['aria-describedby']).to eq('course-creation-tooltip')
    expect(icon['aria-label']).to eq('More information')
    expect(page).to have_css('#course-creation-tooltip[role="tooltip"]', visible: :all)
  end

  it 'reveals the tooltip when the icon receives keyboard focus' do
    page.execute_script("document.querySelector('.instructions.tooltip-trigger img').focus()")
    focused = page.evaluate_script(
      "document.activeElement === document.querySelector('.instructions.tooltip-trigger img')"
    )
    expect(focused).to eq(true)
    sleep 0.5 # let the 0.2s opacity transition finish
    opacity = page.evaluate_script(
      "getComputedStyle(document.querySelector('.instructions.tooltip-trigger .tooltip')).opacity"
    )
    expect(opacity).to eq('1')
  end

  it 'does not shift the trigger element when the icon receives focus' do
    top_before = page.evaluate_script(
      "document.querySelector('.instructions.tooltip-trigger').getBoundingClientRect().top"
    )
    page.execute_script("document.querySelector('.instructions.tooltip-trigger img').focus()")
    top_after = page.evaluate_script(
      "document.querySelector('.instructions.tooltip-trigger').getBoundingClientRect().top"
    )
    expect(top_after).to eq(top_before)
  end
end
```
</details>

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage initiated the review and approved the comment before posting. The Stylus compilation output and the failing/passing spec runs (both the PR as written and the corrected indentation) were produced and verified locally by Claude Code; Sage reviewed the findings but did not independently re-run them.*

MasterIfeanyi:
@ragesoss  Please review


Fix ARIA tooltip focus styles and cleanup review feedback

- Fix Stylus indentation so &:focus-within nests .tooltip correctly, instead of compiling into a separate comma-separated selector
- Remove role="button" from the info icon since it isn't an interactive control; keep tabindex and aria-describedby
- Move hardcoded "More information" label into a locale key (dashboard.tooltip.aria_label) for i18n support
- Standardize Haml attribute syntax on the icon (use key: value style consistently instead of mixing hashrocket and colon syntax)

Addresses review feedback on #7021, part of #6095

ragesoss (changes requested):
**Note on this feedback**: this review was produced with AI assistance (Claude Code). It is meant to be read, understood, and used or ignored per the developer's own judgment — not automatically trusted.

Three of the four issues from the previous review are resolved, and the CSS half of the fix now genuinely works. One residual item from point 2 remains, and it is a one-line change.

## Resolved

**1. The `:focus-within` selector.** The indentation fix is correct. It now compiles to `.tooltip-trigger:focus-within .tooltip`, confirmed in the built `main`, `training`, `rtl-main`, and `rtl-training` bundles. Specificity is (0,3,0) against the base rule's (0,2,0), so it wins regardless of source order. In a browser, focusing the icon now takes the tooltip to computed `opacity: 1` — the same spec that failed against the previous commit passes against this one.

**3. i18n.** `t("dashboard.tooltip.aria_label")` resolves, and `config.i18n.fallbacks` is enabled, so untranslated locales fall back to English rather than rendering a "translation missing" string into the `aria-label`.

**4. Haml attribute syntax.** Consistent now.

## Remaining: `alt=""` alongside `aria-label` (point 2)

Removing `role="button"` was the right call. What remains is the second half of that point — the icon is now `alt=""` (implicitly presentational) while also being focusable and carrying `aria-label`:

```html
<img alt="" aria-describedby="course-creation-tooltip" aria-label="More information"
     src="/assets/images/info.svg" tabindex="0">
```

Two checks were run against this, and they are worth reporting separately because they do not agree:

**Chrome's accessibility tree handles it correctly.** Queried via CDP `Accessibility.getPartialAXTree`, the node comes back as:

```
role:        image
name:        "More information"   (sources: aria-label; alt="" marked superseded: true)
description: "This site tracks wiki activities through Programs — ..."
focusable:   true
ignored:     false
```

ARIA's presentational-roles conflict resolution applies as specified, so the practical impact on NVDA is probably nil. The prior review's concern about a "presentational-yet-focusable" image does not materialize in Chrome.

**axe-core still flags it**, as `presentation-role-conflict` (impact: minor), on that exact element. The rule exists because depending on conflict resolution is fragile across assistive technologies rather than because the computed result is wrong here.

Giving the image a meaningful `alt` instead resolves the conflict at the source:

```haml
%img{src: "/assets/images/info.svg", alt: t("dashboard.tooltip.aria_label"), tabindex: "0", "aria-describedby": "course-creation-tooltip"}
```

This was tested locally. The resulting accessibility tree is identical — same `role: image`, same name, same description, still focusable — and the axe scan of the trigger returns zero violations and zero incomplete results. No downside was found to making the swap.

## Minor notes, none blocking

- The `:focus-within` rule is now actually in effect, where before it compiled to a selector that never matched. It applies to all ~52 `.tooltip-trigger` sites, including triggers that wrap buttons and links (for example `uploads_handler.jsx` and `IconOpener.jsx`), which will now hold their tooltips open while a child has focus. This is very likely the desired accessible behavior; noting it because it is a site-wide behavior change that first takes effect with this commit.
- No `qqq.yml` entry accompanies the new key. Recent additions to `en.yml` in this repo skip qqq, so this is consistent with current practice — but "More information" carries no context for translators, and there is precedent here for documenting ambiguous messages (`3b4df62e8`).
- `dashboard.tooltip.aria_label` is a fairly generic key for a label specific to the course-creation info icon.

## Verification performed

`yarn build` completes cleanly. `spec/features/dashboard_spec.rb` and `spec/views/` pass (20 examples, 0 failures). No `.rb` or `.js` files are touched, so RuboCop and ESLint have no targets in this diff.

<details><summary>axe scan used to identify the finding (AI-generated during review; run locally — fails against this commit, passes with the <code>alt</code> change)</summary>

There is no need to add this to the PR; it is included so the finding can be reproduced independently.

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe 'dashboard info tooltip axe scan', type: :feature, js: true do
  let(:user) { create(:user) }
  let(:axe_source) { File.read(Rails.root.join('node_modules/axe-core/axe.min.js')) }

  before do
    allow(Features).to receive(:open_course_creation?).and_return(true)
    login_as(user, scope: :user)
    visit root_path
    expect(page).to have_content 'My Dashboard'
  end

  it 'reports no axe violations within the tooltip trigger' do
    page.execute_script(axe_source)
    result = page.evaluate_async_script(<<~JS)
      var done = arguments[arguments.length - 1];
      axe.run('.instructions.tooltip-trigger', {}, function (err, res) {
        done(err ? { error: String(err) } : {
          violations: res.violations.map(function (v) {
            return { id: v.id, impact: v.impact, nodes: v.nodes.map(function (n) { return n.html; }) };
          }),
          incomplete: res.incomplete.map(function (v) { return v.id; })
        });
      });
    JS
    expect(result['violations']).to eq([])
  end
end
```

Against this commit it fails with:

```
got: [{"id" => "presentation-role-conflict", "impact" => "minor",
       "nodes" => ["<img alt=\"\" aria-describedby=\"course-creation-tooltip\"
                    aria-label=\"More information\" src=\"/assets/images/info.svg\"
                    tabindex=\"0\">"]}]
```

With the `alt` change it returns `{"violations": [], "incomplete": []}`.

Note that the tooltip has a 0.2s opacity transition, so any spec reading computed opacity after focus needs to wait for it.
</details>

*Drafted in a Claude Code session (~25 min, 2 interactions). Sage initiated the re-review and decided that the remaining `alt` item was worth requesting. The Stylus compilation output, the Chrome accessibility-tree query, the axe scans against both the current markup and the proposed `alt` change, and the local spec runs were produced and verified by Claude Code; Sage reviewed the findings before posting but did not independently re-run them. The spec above is AI-generated and was run locally, but was not line-by-line reviewed by Sage.*

MasterIfeanyi:
@ragesoss  I have made the required changes, please review.

ragesoss (approved):
**Note on this feedback**: this review was produced with AI assistance (Claude Code). It is meant to be read, understood, and used or ignored per the developer's own judgment — not automatically trusted.

All items from the two previous reviews are resolved. The `alt` change is the last one, and it lands correctly.

## Resolved

**The `alt=""` / `aria-label` conflict (review #2).** Giving the image a meaningful, translated `alt` and dropping `aria-label` resolves the presentation-role conflict at the source. Renaming the locale key `aria_label` → `alt` alongside the change is a reasonable call — the key now matches what it populates. Two checks confirm the result:

The axe scan of the trigger, which previously returned `presentation-role-conflict` (impact: minor) on this element, now returns clean:

```
{"violations" => [], "incomplete" => []}
```

And Chrome's accessibility tree, queried via CDP `Accessibility.getPartialAXTree`, computes the name from `alt` with no superseded source:

```
role:        image
name:        "More information"    (source: alt; not superseded)
description: "This site tracks wiki activities through Programs — ..."
focusable:   true
describedby: course-creation-tooltip
ignored:     false
```

This matches what the previous review predicted the swap would produce: same role, same name, same description, still focusable — with the conflict gone.

**The `:focus-within` selector**, re-confirmed against this commit. It compiles to `.tooltip-trigger:focus-within .tooltip` in the built `main`, `training`, `rtl-main`, and `rtl-training` bundles, and in a browser, focusing the icon takes the tooltip to computed `opacity: 1`.

**i18n** and **Haml attribute syntax** remain resolved. No stale `dashboard.tooltip.aria_label` reference remains anywhere in the tree.

## CI has not run on the head commit

The workflow run for `6f6cc47f3f` came back `action_required` — it is waiting behind the first-time-contributor approval gate rather than having run and passed. The last green CI run on this branch was `85c8e407de`, which predates the `alt` change. Worth approving the run before merge, though the delta since then touches only a Haml attribute and a locale key.

## Test coverage

The PR adds no specs, so the checks described above and in the previous reviews live only in throwaway review specs. The change has no branching logic, so present bug risk is low — but the failure mode is silent, and this PR exists because a Stylus indentation slip silently disabled the CSS half of the fix. Nothing in the suite would catch a reintroduction, or catch `aria-describedby` being dropped in a future edit of the partial.

There is currently no spec covering `_open_course_creation` at all (it requires stubbing `Features.open_course_creation?`), so this would be new ground rather than an addition to existing coverage. Not treated as blocking, but a trimmed version of the spec below would be cheap insurance.

## Minor notes, none blocking

- No `qqq.yml` entry accompanies the new key. This is consistent with recent `en.yml` additions in this repo, but "More information" carries no context for translators.
- `dashboard.tooltip.alt` names the HTML attribute rather than the message's meaning, and is generic for a label specific to the course-creation info icon.
- The `:focus-within` rule applies to all ~52 `.tooltip-trigger` sites, including triggers wrapping buttons and links, which will hold their tooltips open while a child has focus. Very likely the desired accessible behavior; noted because it is a site-wide change that first takes effect here.
- The full WAI-ARIA tooltip pattern also calls for Escape-to-dismiss, which a CSS-only approach cannot provide. Noted for completeness.

## Out of scope, for later

Fifteen other instances of this same info-icon pattern still carry `alt="tooltip default logo"` with no `tabindex` or `aria-describedby` — four in `app/views/courses/_header_stats.html.haml` and eleven across JSX components (`student_stats.jsx`, `details.jsx`, `overview_stat.jsx`, `namespace_select.jsx`, and others). The new site-wide `:focus-within` rule does not help them, since those images are not focusable. Issue #6095 names the Dashboard icon specifically, so this is correctly outside this PR — but the markup here is the template those would follow.

## Verification performed

`yarn build` completes cleanly. `spec/features/dashboard_spec.rb` and `spec/views/` pass (20 examples, 0 failures). Four review specs pass against this commit. No `.rb` or `.js` files are touched, so RuboCop and ESLint have no targets in this diff.

<details><summary>Review specs used for verification (AI-generated during review; run locally, pass against this commit)</summary>

There is no need to add these to the PR as-is; they are included so the findings can be reproduced independently, and in case the coverage note above seems worth acting on.

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe 'dashboard info tooltip accessibility', type: :feature, js: true do
  let(:user) { create(:user) }
  let(:axe_source) { File.read(Rails.root.join('node_modules/axe-core/axe.min.js')) }

  before do
    allow(Features).to receive(:open_course_creation?).and_return(true)
    login_as(user, scope: :user)
    visit root_path
    expect(page).to have_content 'My Dashboard'
  end

  it 'exposes the ARIA relationship between icon and tooltip' do
    icon = find('.instructions.tooltip-trigger img')
    expect(icon['tabindex']).to eq('0')
    expect(icon['aria-describedby']).to eq('course-creation-tooltip')
    expect(icon['alt']).to eq('More information')
    expect(icon['aria-label']).to be_nil
    expect(page).to have_css('#course-creation-tooltip[role="tooltip"]', visible: :all)
  end

  it 'reveals the tooltip when the icon receives keyboard focus' do
    page.execute_script("document.querySelector('.instructions.tooltip-trigger img').focus()")
    focused = page.evaluate_script(
      "document.activeElement === document.querySelector('.instructions.tooltip-trigger img')"
    )
    expect(focused).to eq(true)
    sleep 0.5 # let the 0.2s opacity transition finish
    opacity = page.evaluate_script(
      "getComputedStyle(document.querySelector('.instructions.tooltip-trigger .tooltip')).opacity"
    )
    expect(opacity).to eq('1')
  end

  it 'does not shift the trigger element when the icon receives focus' do
    top_before = page.evaluate_script(
      "document.querySelector('.instructions.tooltip-trigger').getBoundingClientRect().top"
    )
    page.execute_script("document.querySelector('.instructions.tooltip-trigger img').focus()")
    top_after = page.evaluate_script(
      "document.querySelector('.instructions.tooltip-trigger').getBoundingClientRect().top"
    )
    expect(top_after).to eq(top_before)
  end

  it 'reports no axe violations within the tooltip trigger' do
    page.execute_script(axe_source)
    result = page.evaluate_async_script(<<~JS)
      var done = arguments[arguments.length - 1];
      axe.run('.instructions.tooltip-trigger', {}, function (err, res) {
        done(err ? { error: String(err) } : {
          violations: res.violations.map(function (v) {
            return { id: v.id, impact: v.impact, nodes: v.nodes.map(function (n) { return n.html; }) };
          }),
          incomplete: res.incomplete.map(function (v) { return v.id; })
        });
      });
    JS
    expect(result['violations']).to eq([])
    expect(result['incomplete']).to eq([])
  end
end
```

Note that the tooltip has a 0.2s opacity transition, so any spec reading computed opacity after focus needs to wait for it.

</details>

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage initiated the re-review and decided to approve. The compiled-CSS check, the Chrome accessibility-tree query, the axe scan, and the local spec runs were produced and verified by Claude Code; Sage read the findings summary before approving but did not independently re-run them. The specs above are AI-generated and were run locally, but were not line-by-line reviewed by Sage.*

PR #7022, opened 2026-08-30
Add screen reader edit boundary announcements and hide highlight token spans

nexpectArpit:
Eliminate confusing word-by-word screen reader announcements of editor highlight spans during authorship highlighting. Dynamically announce edit boundaries using visually hidden screen reader text while suppressing individual token span announcements.

#### Proposed Changes
1. HTML Processing (`ParsedArticle.jsx`):
   - Added `processAuthorshipHtml` to parse and process the article HTML string on the client side using `DOMParser`.
   - Filters `span.editor-token` elements with active `user-highlight-*` classes and editor name titles.
   - Groups contiguous editor spans by checking parent node matching, editor name equality, and whitespace-only text node transitions.
   - Injects boundary announcement spans:
     - `<span class="screen-reader">Edited by [Name]</span>` at the start of an edit block.
     - `<span class="screen-reader">End edit</span>` at the end of an edit block.
   - Appends `aria-hidden="true"` and `role="presentation"` to individual token spans.

2. Unit Tests (`parsed_article.spec.js`):
   - Added a new Jest test suite verifying contiguous span grouping, screen-reader boundary injection, and aria-hidden token masking.

#### Verification Results
All related component unit tests pass cleanly:

`npx jest test/components/parsed_article.spec.js test/components/article_viewer_legend.spec.js test/components/use_authorship_highlighting.spec.js `

### DIFF
--- a/app/assets/javascripts/components/common/ArticleViewer/components/ParsedArticle.jsx
+++ b/app/assets/javascripts/components/common/ArticleViewer/components/ParsedArticle.jsx
@@ -5,10 +5,93 @@ import PropTypes from 'prop-types';
 const httpLinkMatcher = /(<a href="http)/g;
 const blankTargetLink = '<a target="_blank" href="http';
 
+const processAuthorshipHtml = (html) => {
+  if (!html || typeof window === 'undefined' || !window.DOMParser) return html;
+
+  try {
+    const parser = new window.DOMParser();
+    const doc = parser.parseFromString(html, 'text/html');
+
+    const spans = Array.from(doc.querySelectorAll('span.editor-token'));
+    const isTargetSpan = (span) => {
+      const classList = span.className || '';
+      const hasHighlightClass = /user-highlight-\d+/.test(classList);
+      const hasTitle = span.hasAttribute('title') && span.getAttribute('title').trim() !== '';
+      return hasHighlightClass && hasTitle;
+    };
+
+    const targetSpans = spans.filter(isTargetSpan);
+    if (targetSpans.length === 0) return html;
+
+    const areContiguous = (spanA, spanB) => {
+      if (spanA.parentNode !== spanB.parentNode) return false;
+      if (spanA.getAttribute('title') !== spanB.getAttribute('title')) return false;
+
+      let next = spanA.nextSibling;
+      while (next && next !== spanB) {
+        if (next.nodeType === 3) { // Node.TEXT_NODE
+          if (next.nodeValue.trim() !== '') return false;
+        } else {
+          return false;
+        }
+        next = next.nextSibling;
+      }
+      return next === spanB;
+    };
+
+    const groups = [];
+    let currentGroup = [];
+
+    targetSpans.forEach((span) => {
+      if (currentGroup.length === 0) {
+        currentGroup.push(span);
+      } else {
+        const lastSpan = currentGroup[currentGroup.length - 1];
+        if (areContiguous(lastSpan, span)) {
+          currentGroup.push(span);
+        } else {
+          groups.push(currentGroup);
+          currentGroup = [span];
+        }
+      }
+    });
+    if (currentGroup.length > 0) {
+      groups.push(currentGroup);
+    }
+
+    groups.forEach((group) => {
+      const editorName = group[0].getAttribute('title');
+
+      const firstSpan = group[0];
+      const startSpan = doc.createElement('span');
+      startSpan.className = 'screen-reader';
+      startSpan.textContent = `Edited by ${editorName}`;
+      firstSpan.parentNode.insertBefore(startSpan, firstSpan);
+
+      const lastSpan = group[group.length - 1];
+      const endSpan = doc.createElement('span');
+      endSpan.className = 'screen-reader';
+      endSpan.textContent = 'End edit';
+      lastSpan.parentNode.insertBefore(endSpan, lastSpan.nextSibling);
+
+      group.forEach((span) => {
+        span.setAttribute('aria-hidden', 'true');
+        span.setAttribute('role', 'presentation');
+      });
+    });
+
+    return doc.body.innerHTML;
+  } catch (error) {
+    console.error('Failed to parse or process authorship HTML:', error);
+    return html;
+  }
+};
+
 export const ParsedArticle = ({ html, onInnerHTMLClick, onInnerHTMLKeyDown }) => {
   // This sets `target="_blank"` for all of the non-anchor links in the article HTML,
   // so that clicking one will open it in a new tab.
   const articleHTML = html?.replace(httpLinkMatcher, blankTargetLink);
+  const processedHTML = processAuthorshipHtml(articleHTML);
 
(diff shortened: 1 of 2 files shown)


nexpectArpit:
@ragesoss please review it

ragesoss (changes requested):
A note on how to read this: these findings come from an AI-assisted review session.
They are not meant to be trusted automatically or acted on blindly. Read them, check
the reasoning, and use or discard them on your own judgment.

## Main finding: the token spans carry the article text, so hiding them removes it

`aria-hidden="true"` removes an element **and its entire subtree** from the
accessibility tree. The `.editor-token` spans are not wrappers placed around already-
readable text. Under WhoColor each span *is* one word of the article. Marking them
hidden therefore deletes the highlighted words from what a screen reader reads,
leaving only the injected boundary announcements.

Given the markup `highlightAuthors` produces today, where brackets mark one student's
tokens:

```
The compound [is highly toxic] to fish
```

the accessible text changes from

```
The compound is highly toxic to fish
```

to

```
The compoundEdited by Student OneEnd edit to fish
```

The student's own words are the only part removed, and that is the text the feature
exists to surface.

Measured against a live WhoColor response (`/en/whocolor/v1.0.0-beta/Hedgehog/0/`,
422 KB, 2,440 token spans), treating the article's two most prolific editors as
highlighted students at 523 tokens between them:

| Measure | Value |
|---|---|
| Announcement groups | 203 |
| Injected `.screen-reader` spans | 406 |
| Groups containing exactly one token | 103 of 203 |
| Median group size | 1 token |
| Article characters removed from the a11y tree | 3,012 |
| Announcement characters added | 5,887 |

On that article a screen reader user would hear roughly twice as much "Edited by …
End edit" boilerplate as the article text it replaced, and half the announcement
pairs surround a single word that is itself hidden.

This is inherited rather than introduced here. The T4SG prototype in #6597 applied
`aria-hidden="true"` to *every* `.editor-token`, and the plan agreed in #6602
restated that step. This PR is narrower, hiding only highlighted tokens, but the
same consequence follows.

Worth noting that #6602 describes the goal differently: "each continuous section of
text by a highlighted user is enclosed in a relevant HTML [element] with relevant
metadata". A single wrapper element around each run, with the token spans left
visible to assistive technology, would match that description and would remove no
text. Whether the per-token `title` attribute is what actually produces the
word-by-word noise is worth confirming with a screen reader before picking a
direction.

## Other observations

1. **No memoization.** `processAuthorshipHtml` runs in the component body, so a full
   `DOMParser` parse and re-serialization of the article happens on every render of
   `ParsedArticle`, not only when `html` changes. `ArticleViewerShell` re-renders
   often; `setUnhighlightedContributors` appends one contributor per call, for
   instance. In jsdom the real 422 KB article takes about 230 ms per pass. A browser
   will be much faster, but the work is repeated and avoidable with `useMemo` keyed
   on `html`.

2. **Articles with no highlighting still pay for the parse.** The
   `targetSpans.length === 0` early return happens *after* the document has been
   parsed, so the claim-verification feature and the plain article view absorb the
   cost for no benefit. Benchmarked in jsdom against master's component at roughly
   830 KB, render time on an unhighlighted article went from 118 ms to 248 ms. A
   string check such as `html.includes('user-highlight-')` before parsing would skip
   it entirely.

3. **The announcement strings are hardcoded English.** `Edited by ${editorName}` and
   `End edit` bypass `I18n.t`, unlike
   `articles.show_current_version_with_authorship_highlighting` in the same feature,
   so they will not reach translatewiki. The wording itself should come from Wiki Ed
   rather than being settled in the PR.

4. **The injected spans have no surrounding whitespace**, so they run into adjacent
   text. The accessible string reads `compoundEdited by Student One` rather than
   `compound Edited by Student One`. Some screen readers will run those together.

5. **`.screen-reader` is positioned off-screen rather than clipped**, so the
   announcements are part of `textContent` and of any text selection. Copying a
   passage out of the article viewer now yields "Edited by …" and "End edit" mixed
   into the prose.

6. **`role="presentation"` is redundant** next to `aria-hidden="true"`, which already
   removes the element and its subtree.

## Checked and not a problem

- The `DOMParser` round-trip is faithful on real output. The only differences are
  HTML5 normalization: `<tbody>` insertion, void-element serialization, entity
  decoding. Elements the parser hoists into `<head>` would be dropped by
  `doc.body.innerHTML`, but real WhoColor HTML opens with
  `<div class="mw-parser-output">`, so nothing is lost.
- WhoColor does not place token spans inside `<a>` elements, 0 of 1,110 links in the
  sample, so links do not lose their accessible names.
- `.screen-reader` is already in the main CSS bundle, so no stylesheet change is
  needed.
- ESLint is clean, `yarn build` succeeds, and the full Jest suite passes at 81 suites
  and 508 tests with this branch merged onto current master.

## Test coverage

`parsed_article.spec.js` asserts that the boundary spans appear and that
`aria-hidden` is set, but nothing asserts what happens to the article text, which is
where the behavior actually changes. A test that computes the accessible text, by
walking the DOM and skipping `aria-hidden="true"` subtrees, would have caught this.

An exploratory spec doing exactly that is below. It was written by the AI agent
during this review and run against this branch, where all five examples pass and
document current behavior.

<details>
<summary>Exploratory spec (test/components/zz_review_7022_a11y.spec.js)</summary>

```js
// The accessibility-tree text: what a screen reader reads out. An element carrying
// aria-hidden="true" is removed along with its entire subtree.
const accessibleText = (root) => {
  const parts = [];
  const walk = (node) => {
    if (node.nodeType === 1 && node.getAttribute('aria-hidden') === 'true') return;
    if (node.nodeType === 3) { parts.push(node.nodeValue); return; }
    Array.from(node.childNodes).forEach(walk);
  };
  walk(root);
  return parts.join('').replace(/\s+/g, ' ').trim();
};

it('reads the full sentence before the change', () => {
  const bare = document.createElement('div');
  bare.innerHTML = html;
  expect(accessibleText(bare)).toBe('The compound is highly toxic to fish');
});

it('drops the student\'s words from the accessibility tree after the change', () => {
  const spoken = accessibleText(renderParsed(html));
  expect(spoken).toContain('Edited by Student One');
  expect(spoken).not.toContain('is highly toxic');
  expect(spoken).toBe('The compoundEdited by Student OneEnd edit to fish');
});
```

The full file also covers single words edited in place, where the output is
`ResearchersEdited by Student OneEnd edit foundEdited by Student OneEnd edit …`,
and two students editing adjacent words. It builds its fixtures by replaying the
two string transforms `useAuthorshipHighlighting.highlightAuthors` applies, so the
markup under test matches what `ParsedArticle` receives in production.

</details>

*Drafted in a Claude Code session (about 25 minutes, two interactions). The agent
checked out the branch onto current master, ran the Jest suite, benchmarked
`ParsedArticle` against master's version in-process, reproduced the accessibility-tree
behavior with a new spec, and measured the group and character counts against a live
WhoColor response for Hedgehog. Sage read the findings and the full comment before
approving this post, but did not independently verify them or run the code.*

Issue #1922, opened 2018-06-29
Add `Institution` model so that courses can be grouped by school/institution


ragesoss:
Having a new model for Institution would allow us to build features like an aggregrate page for all courses from the same university, and a map that shows the distribution of courses (based on the locations of their institutions).

The tricky part of this will be migrating the existing data; we will need a path to first create institution records based on all the different `school` values in existing `Course` records, and then have a way to merge duplicate values that actually refer to the same institution, without changing the `school` or `slug` fields on the `Course` records.

PR #7023, opened 2026-08-30
Add Institution model and link Course to it (#1922)

MasterIfeanyi:
## What this PR does

Addresses #1922. Adds a new `Institution` model so courses can eventually be grouped by institution, without changing any existing behavior.

- Create institutions table with unique name constraint
- Add institution_id reference on courses (optional, keeps school untouched)
- Add belongs_to/has_many association between Course and Institution
- Remove old alias_attribute :institution, :school (unused, conflicted with new association)
- Add rake task to backfill Institution records from existing Course#school values
- Add Institution.merge(survivor, duplicate) for manually merging duplicate institutions

## AI usage

I used Claude (Anthropic) throughout this work, as a step-by-step guide and pair-programming partner rather than to generate the solution outright. Specifically, I used it to:
- Investigate the codebase (using grep/sed) to find every place `school` and the `institution` alias were referenced, before writing any code
- Talk through the model design (what columns `Institution` needed, whether `name` should be unique)
- Talk through the tradeoff between automatic vs. manual merging of duplicate institutions, and decide on manual merging given the risk of automatic fuzzy-matching corrupting data
- Get plain-language explanations of Rails concepts (migrations, `belongs_to`/`has_many`, `find_or_create_by`, `update_all`) mapped against MERN/Mongoose, which I'm more familiar with
- Review and verify each step's output (migration results, console query results) before proceeding

I wrote and ran every command myself, and manually tested the migration and merge behavior in the Rails console before committing.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -104,6 +104,7 @@ class Course < ApplicationRecord
   # Metadata #
   ############
   belongs_to :home_wiki, class_name: 'Wiki'
+  belongs_to :institution, optional: true
 
   has_many :campaigns_courses, class_name: 'CampaignsCourses', dependent: :destroy
   has_many :campaigns, through: :campaigns_courses
@@ -246,7 +247,7 @@ class Course < ApplicationRecord
   ###########
   # Aliases #
   ###########
-  alias_attribute :institution, :school
+
 
   ##################
   # Course content #
--- a/app/models/institution.rb
+++ b/app/models/institution.rb
@@ -1,0 +1,10 @@
+class Institution < ApplicationRecord
+  has_many :courses
+
+  validates :name, presence: true, uniqueness: true
+
+  def self.merge(survivor, duplicate)
+    duplicate.courses.update_all(institution_id: survivor.id)
+    duplicate.destroy
+  end
+end
--- a/db/migrate/20260830200456_create_institutions.rb
+++ b/db/migrate/20260830200456_create_institutions.rb
@@ -1,0 +1,9 @@
+class CreateInstitutions < ActiveRecord::Migration[8.1]
+  def change
+    create_table :institutions do |t|
+      t.string :name, null: false
+      t.timestamps
+    end
+    add_index :institutions, :name, unique: true
+  end
+end
--- a/db/migrate/20260830205737_add_institution_to_courses.rb
+++ b/db/migrate/20260830205737_add_institution_to_courses.rb
@@ -1,0 +1,5 @@
+class AddInstitutionToCourses < ActiveRecord::Migration[8.1]
+  def change
+    add_reference :courses, :institution, foreign_key: true
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,31 +10,31 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[8.1].define(version: 2026_08_05_180000) do
+ActiveRecord::Schema[8.1].define(version: 2026_08_30_205737) do
   create_table "admin_course_notes", charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
     t.integer "courses_id"
-    t.string "title"
-    t.text "text"
-    t.string "edited_by"
     t.datetime "created_at", null: false
+    t.string "edited_by"
+    t.text "text"
+    t.string "title"
     t.datetime "updated_at", null: false
     t.index ["courses_id"], name: "index_admin_course_notes_on_courses_id"
   end
 
   create_table "alerts", id: :integer, charset: "utf8mb4", collation: "utf8mb4_unicode_ci", force: :cascade do |t|
-    t.integer "course_id"
-    t.integer "user_id"
     t.integer "article_id"
-    t.integer "revision_id"
-    t.string "type"
-    t.datetime "email_sent_at", precision: nil
+    t.integer "course_id"
     t.datetime "created_at", precision: nil, null: false
-    t.datetime "updated_at", precision: nil, null: false
-    t.text "message"
-    t.integer "target_user_id"
-    t.integer "subject_id"
-    t.boolean "resolved", default: false
     t.text "details"
+    t.datetime "email_sent_at", precision: nil
+    t.text "message"
+    t.boolean "resolved", default: false
+    t.integer "revision_id"
+    t.integer "subject_id"
+    t.integer "target_user_id"
+    t.string "type"
+    t.datetime "updated_at", precision: nil, null: false
+    t.integer "user_id"
     t.index ["article_id"], name: "index_alerts_on_article_id"
     t.index ["course_id"], name: "index_alerts_on_course_id"
     t.index ["revision_id"], name: "index_alerts_on_revision_id"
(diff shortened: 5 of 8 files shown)


ragesoss:
I'll have a look at this soon.

The biggest concern is backfilling the very messy data we currently have in the `school` field; the chief virtue of an Institution model would be to link (for example) courses at "UW Bothell" and "University of Washington, Bothell" to the same record, but there are enough records and enough ambiguity that we won't be able to programatically do this as part of the rollout, and will need to be robust to an interim period where there are still lots of messy duplicate records.

Sounds like this was on your radar. One thing I expect to need is a admin-facing UI for merging, rather than only supporting command line manual merging. Perhaps that can be a separate PR, as this one (based on the description) should just be laying the data model groundwork without changing any frontend behavior. But it will be good to explicitly talk through what kinds of UI changes will build on top of this, so that model has the right shape for the things we'll need from it.

MasterIfeanyi:
I agree I will love to be responsible in implementing the admin-facing UI for this fix, and also, what kind of UI changes would you like ?

ragesoss:
@MasterIfeanyi I'm not sure! Figuring out the UX side of it is the biggest challenge with this, I think.

ragesoss (changes requested):
This review was drafted with an AI coding agent. Treat the findings as leads to check rather than conclusions to act on — some may be wrong, and the judgment about what matters is yours.

The data-model groundwork here is the right shape for #1922, and splitting the admin merge UI into a follow-up is the right call. Most of what follows is about the interim period rather than the model itself.

## Blocking

**`/all_courses.json` loses the `institution` field.**

`app/views/analytics/all_courses.json.jbuilder:5` passes `:institution` to `json.call(course, ...)`. That previously resolved through `alias_attribute :institution, :school`. With the alias removed it resolves to the new association instead — `nil` for every course until the backfill runs, and an `Institution` object rather than a string afterwards.

`app/controllers/analytics_controller.rb` notes these endpoints were requested by WMF's Community Data and Evaluation team, so this is externally visible.

A spec written during review passes on master and fails on this branch:

```ruby
# spec/requests/all_courses_institution_review_spec.rb
describe 'GET /all_courses.json', type: :request do
  before do
    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(nil)
    create(:course, id: 1, school: 'University of Washington, Bothell',
                    start: 1.year.ago, end: 1.day.from_now)
  end

  it 'reports the institution name for the course' do
    get '/all_courses.json'
    expect(JSON.parse(response.body)['all_courses'].first['institution'])
      .to eq('University of Washington, Bothell')
  end
end
```

```
master:      1 example, 0 failures
this branch: 1 example, 1 failure   expected: "University of Washington, Bothell"  got: nil
```

The existing `#all_courses` spec only asserts the body parses as a Hash and contains a slug, which is why CI stayed green. Keeping the payload stable probably means changing the jbuilder to `json.institution course.school` and adding an assertion on the value.

**No test coverage, and the factory cannot build a record.**

`spec/models/institution_spec.rb` is the unmodified generator scaffold. `spec/factories/institutions.rb` has an empty body, so `create(:institution)` raises `Validation failed: Name can't be blank` — and the blank line inside it is the trailing whitespace currently failing CI's `git diff --check`. Adding `name { "University #{n}" }` via a sequence fixes both.

## Worth addressing before merge

**The backfill silently skips courses that fail validation.** `course.update(institution:)` returns `false` rather than raising, and `lib/tasks/institutions.rake` ignores the return value. A course with a nil `term` — the sort of legacy row production holds — is left unlinked with no output. Course also runs its whole `before_save` chain on each row; `ensure_required_params` can `throw :abort`, and `set_default_times` can mutate `start`/`end`, which then trips `set_needs_update_for_timeslice`.

Since this is a pure data link, skipping the model layer sidesteps both problems and is much cheaper — one statement per distinct school rather than one per course (8,660 courses span 1,746 distinct school values):

```ruby
Course.distinct.pluck(:school).compact_blank.each do |school|
  institution = Institution.find_or_create_by!(name: school)
  Course.where(school:).update_all(institution_id: institution.id)
end
```

**`Institution.merge` shadows ActiveRecord's own class-level `merge`.** `Institution.merge(Institution.where(name: 'x'))` now raises `ArgumentError: wrong number of arguments (given 1, expected 2)`. Nothing calls it today, so this isn't a live break, but the collision is worth avoiding. Moving it to `app/services/merge_institutions.rb` would match the repo's service conventions and give the admin UI somewhere to call in with real error handling. It also wants a transaction — if `duplicate.destroy` raises, the courses have already moved and stay moved.

**`has_many :courses` has no `dependent:` option.** With the foreign key defaulting to RESTRICT, `institution.destroy` raises `ActiveRecord::InvalidForeignKey` straight from MySQL rather than failing as a Rails validation. `dependent: :restrict_with_error` would make that a normal error path, which the merge UI will need.

**`db/schema.rb` carries ~590 lines of unrelated churn.** Every column in every table has been reordered alphabetically — an artifact of the local dump, not a real change. The substance is correct (checked against the merge base: nothing dropped, only the new table, `courses.institution_id`, its index and the FK), but it already causes trouble: merging current master conflicts, and one hunk is

```
<<<<<<< HEAD
    t.integer "user_id"
    t.integer "wiki_id"
=======
    t.integer "sample_id"
    t.index ["sample_id"], name: "index_revision_ai_scores_on_sample_id"
>>>>>>> master
```

where resolving toward HEAD silently drops master's new `revision_ai_scores.sample_id`. Taking master's `schema.rb` and hand-adding the four new lines brings the whole PR to 58 lines.

**The new table is the only non-utf8mb4 table in the schema.** It dumps as `charset: "utf8mb3", collation: "utf8mb3_general_ci"`, picked up from the local database default. CI runs `parallel:load_schema`, so CI and fresh dev setups would get utf8mb3 while production gets whatever its own default is. No practical breakage was found — cross-collation joins coerce fine on MariaDB, and none of the 1,746 current `school` values need 4-byte encoding — but pinning `charset: 'utf8mb4', collation: 'utf8mb4_unicode_ci'` in the migration would keep environments identical.

## Migration safety

Both migrations are additive and the `courses` table is small (~8,700 rows, 34 MB). The column add qualifies for `ALGORITHM=INSTANT`; the index is online; adding the foreign key forces a table copy, but at this size that is sub-second. Safe inside a normal deploy.

## Two questions about the shape, since the merge UI builds on this

**Merges aren't durable.** `merge` deletes the duplicate outright, so nothing records that "UW Bothell" resolves to "University of Washington, Bothell". If course creation ever links an Institution from the typed `school` string, the duplicate comes straight back and the curation work is lost. Somewhere to keep alternate names is what makes a merge decision stick — and it would also let the eventual UI suggest matches instead of relying on fuzzy string comparison at merge time.

**`school` and `institution_id` will drift.** `school` stays user-editable through `course_params`, and nothing keeps the two in sync, so after the backfill any course edit can leave a stale link. Deciding which field is authoritative during the interim — and whether a `before_save` hook should re-link on `school` change — probably wants settling before the backfill runs anywhere real.

For context on how messy the backfill target is: 8,660 courses across 1,746 distinct `school` values, 731 of which appear exactly once. No pure case or whitespace duplicates, but plenty of the harder kind:

```
University of California Berkeley   /  University of California, Berkeley
The University of British Columbia  /  University of British Columbia
Washington University in St Louis   /  Washington University in St. Louis  /  Washington University in Saint Louis
University of Hawaiʻi at Mānoa  /  University of Hawai'i at Mānoa  /  University of Hawaii at Mānoa
Washington College (3)              /  Washington Colege (1)
```

---

The exploratory specs written during this review (a request spec for the `/all_courses.json` payload, a model spec covering the validations, `merge` and `destroy` paths, and a rake-task spec) are available if they'd be a useful starting point for the coverage this PR is missing — just say so.

*Drafted in a Claude Code session (~45 min, 2 interactions). Sage directed the review but has not independently verified the findings. The `/all_courses.json` regression, the unusable factory, the `Institution.merge` argument error, the missing `dependent:` option, and both backfill findings were each reproduced locally by specs written during the review, run against both this branch and master; the specs for affected areas (103 examples) pass. Those specs are AI-generated and were run, not reviewed line by line. The full Ruby suite was not part of this verification.*

Issue #7017, opened 2026-08-28
Sentry noise: Wikidata label fetches on course pages return 403 for crawler traffic (~300 events/day)


ragesoss:
## Summary

The largest single consumer of our Sentry error quota is crawler traffic hitting course article pages and triggering Wikidata label fetches that fail with **HTTP 403 Forbidden**. Over the last 30 days this produced **10,676 events** ([PEONY-2NS], ~23% of all events in the peony project), at a steady ~300–350 events/day since late April.

Since the ~Aug 26 deploy of the `ensureOk`/`ApiError` work (fb4f95b11, cc4fd0933), the same failures surface with a real error message as **[PEONY-3BC] `ApiError: Forbidden`** (escalating, ~290/day) instead of the old stack-less `Error: [object Response]` ([PEONY-2NS] went silent at 2026-08-26 19:24 UTC; [PEONY-3BC] first seen 2026-08-26 20:03 UTC).

This issue is for understanding the failure completely before deciding whether/how to stop collecting it (per discussion, we don't want to jump straight to filtering).

[PEONY-2NS]: https://wiki-education.sentry.io/issues/PEONY-2NS
[PEONY-3BC]: https://wiki-education.sentry.io/issues/PEONY-3BC

## Cause chain (as understood so far)

1. A crawler running headless Chrome crawls `outreachdashboard.wmflabs.org` course pages — the top URLs are all `/courses/<school>/<title>/articles/assigned` and similar course tabs, spread thinly across hundreds of different courses (top URL has only 42 of 10,676 events).
2. The crawler executes the page's JavaScript. On course pages with Wikidata articles, `fetchWikidataLabels` (`app/assets/javascripts/actions/wikidata_actions.js`) fires chunked `wbgetentities` requests (30 IDs each) directly from the browser to `https://www.wikidata.org/w/api.php`.
3. Wikidata responds **403 Forbidden** to these requests. This is consistent with Wikimedia's edge blocking anonymous API traffic from datacenter/cloud IP ranges — the client geolocates to a Singapore datacenter.
4. Before the Aug 26 deploy, the raw `Response` object was thrown and escaped as an unhandled rejection, which Sentry recorded as the stack-less `Error: [object Response]`. After the deploy, `ensureOk` throws `ApiError: Forbidden` and the explicit `Sentry.captureException` in the wikidata label-fetch `catch` (added in 7859c1e0f) reports it.

## Evidence that this is one crawler, not real users

[...]

PR #7024, opened 2026-08-31
Attach Wikidata 403 response body to Sentry capture (PEONY-3BC)

Formasitchijoh:
## What this PR does

Attaches the actual Wikidata 403 response body to Sentry when `fetchWikidataLabels` fails, via `extra: { responseText, url }` on `Sentry.captureException`. Previously these fields were set on the thrown `ApiError` but silently dropped, since plain `Sentry.captureException(error)` doesn't serialize custom properties.

Diagnostic only — no behavior change. Addresses [#7017](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/7017): a crawler is triggering ~290 403s/day from Wikidata's API (PEONY-3BC, 0 real users impacted). The issue asks to confirm the failure mode before deciding whether to suppress it; this gives us the response body needed to make that call.

## AI usage

Prepared with Claude Code (Sonnet 5) via Claude Code CLI — investigation used Sentry MCP queries to confirm the crawler pattern, and code review to verify `ApiError` actually populates `responseText`/`url`. Single commit, one sitting. Commit message and this description were AI-drafted; review for accuracy.

## Screenshots

No UI changes.

- Confirming the crawler's IP/UA (nginx logs on peony-web) is still open and needs server access — not addressed here.
- No test coverage for `wikidata_actions.js`; verified manually.

### DIFF
--- a/app/assets/javascripts/actions/wikidata_actions.js
+++ b/app/assets/javascripts/actions/wikidata_actions.js
@@ -48,7 +48,17 @@ export const fetchWikidataLabels = (wikidataEntities, dispatch) => {
         // report here, a failure here would be invisible to Sentry (ensureOk
         // only console.logs) even though it's the dominant real-world source
         // of this bug (see Sentry issue PEONY-2NS).
-        if (typeof Sentry !== 'undefined') Sentry.captureException(error);
+        // extra.responseText/url surface the actual response body (e.g. Wikidata's
+        // 403 message) since Sentry.captureException(error) alone does not
+        // serialize custom properties on the thrown ApiError (see PEONY-3BC).
+        if (typeof Sentry !== 'undefined') {
+          Sentry.captureException(error, {
+            extra: {
+              responseText: error.responseText,
+              url: error.url
+            }
+          });
+        }
       });
   });
 };
--- a/test/actions/wikidata_actions.spec.js
+++ b/test/actions/wikidata_actions.spec.js
@@ -55,7 +55,8 @@ describe('fetchWikidataLabels', () => {
       ok: false,
       status: 403,
       statusText: 'Forbidden',
-      text: () => Promise.resolve(''),
+      url: 'https://www.wikidata.org/w/api.php?action=wbgetentities',
+      text: () => Promise.resolve('blocked'),
     });
     global.Sentry = { captureException: jest.fn() };
     const dispatch = jest.fn();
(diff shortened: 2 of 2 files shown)


ragesoss (approved):
Reviewed with Claude Code. Findings below are provisional AI-assisted analysis — they're offered to read and weigh, not to be trusted automatically or acted on without your own judgment. Push back on anything that looks wrong.

Approving. The change is diagnostic-only and does what the description says.

**Verification performed**

The Sentry API usage is correct for the pinned `@sentry/browser` 7.120.4. `parseEventHintOrCaptureContext` classifies any object containing an `extra` key as a `CaptureContext` rather than an `EventHint`, so the second argument is applied to the event as expected.

The premise also holds: `extraErrorDataIntegration` is not among v7's browser default integrations and is not added in `Sentry.init` (`app/views/shared/_head.html.haml`), so `ApiError`'s custom properties were in fact being dropped.

PEONY-3BC was checked against live Sentry. The events are `ApiError: Forbidden` with `handled: yes` / `mechanism: generic`, which means they originate from an explicit `captureException` — and `wikidata_actions.js` holds the only one in the JS bundle, so this patches the right call site. That the errors surface as `ApiError` rather than `TypeError: Failed to fetch` also indicates the 403 response carries CORS headers, so `response.text()` succeeds and `responseText` will actually be populated. The current events' context contains only browser/device/os/trace, confirming the fields were missing.

One piece of context for the follow-up decision in #7017: over the last 7 days, 1266 of 1288 events came from a single SG/Chrome/Mac source, but 22 came from CN, UG, US, and CG. Suppression scoped to the crawler would preserve that remaining signal; a blanket filter would not.

ESLint is clean on both changed files, and the PR's specs plus the full `test/actions` and `test/utils` suites (180 tests) pass locally.

**Non-blocking observations**

1. The catch also receives errors that are not `ApiError` — `json()` parse failures and network-layer rejections — and for those the new block reports `{ responseText: undefined, url: undefined }`. In the parse-failure case this discards the `requestUrl` that `utils/request.js` deliberately attaches. `url: error.url || error.requestUrl` would recover it.

2. `responseText` is forwarded unbounded. A WAF or proxy interstitial can run to several KB, and at roughly 215 events/day that adds up. Sentry caps the event so nothing breaks, but a `.slice(0, 500)` would make the payload predictable.

3. Worth a deliberate call rather than a change request: `extraErrorDataIntegration` is already exported by the installed `@sentry/browser`, so no new dependency is needed. Adding it to `Sentry.init` would surface `ApiError`'s `status`, `statusText`, `url`, and `responseText` everywhere — including unhandled rejections that never reach an explicit capture — instead of instrumenting one call site. Since `ensureOk` is shared app-wide, that may be the better long-term shape.

Observations 1 and 2 were demonstrated locally with a throwaway spec exercising the parse-failure, network-error, and large-body paths; that spec is not included here.

*Drafted in a Claude Code session (~10 min, 2 interactions). Sage directed the review and approved posting after a read-through, but did not independently verify the findings. Claude Code ran ESLint and the JS test suites locally, inspected the `@sentry/browser` 7.120.4 source to confirm the argument-parsing behavior, queried live Sentry for PEONY-3BC's event shape and geographic distribution, and wrote and ran a temporary spec to confirm observations 1 and 2. The approval reflects that verification, not a line-by-line human audit.*

Issue #7018, opened 2026-08-28
Sentry noise: 'TypeError: Failed to fetch' dominated by UA-rotating crawler traffic


ragesoss:
## Summary

The second-largest consumer of our Sentry error quota is **`TypeError: Failed to fetch`** from browser `fetch()` calls: **9,839 events** in the last 30 days under [PEONY-2P3] plus **829** under [PEONY-2NQ] — together ~22% of all events in the peony project. Ongoing since late April / early May at ~350/day combined.

The evidence below points to this being dominated by crawler traffic rather than real users, but per discussion we want to understand it fully before deciding whether to stop collecting it. Companion issue for the related 403 noise from the same code path: #7017.

[PEONY-2P3]: https://wiki-education.sentry.io/issues/PEONY-2P3
[PEONY-2NQ]: https://wiki-education.sentry.io/issues/PEONY-2NQ

## What the error is

`TypeError: Failed to fetch` is Chrome's message for a `fetch()` that failed at the **network level** — no HTTP response at all: connection refused/reset, DNS failure, CORS preflight failure, the page navigating away mid-request, offline clients, or content blockers cancelling the request.

- **PEONY-2P3** comes from the Wikidata label fetch: course pages with Wikidata articles fire chunked `wbgetentities` requests (30 IDs each) from the browser to `https://www.wikidata.org/w/api.php` via `fetchWikidataLabels` (`app/assets/javascripts/actions/wikidata_actions.js`, through `utils/request.js:45`).
- **PEONY-2NQ** is the same error from other `request()` call sites (e.g. the user-profile stats request in `components/user_profiles/user_profile.jsx:26`).

## Evidence that most of it is crawler traffic

- The 30-day browser histogram for PEONY-2P3 is **flat across old Chrome versions**: Chrome 131 (1,024), 116 (978), 133 (922), then ~490–508 each for Chrome 105, 108, 109, 110, 112, 120, 124… Real user traffic clusters on the latest two or three Chrome versions; a near-uniform spread across two years of Chrome versions is a user-agent-rotation signature.
- An explicit **`bingbot 2.0`** event appears in the last 24 hours.
- 0 identified users on both issues; events are spread thinly across hundreds of course URLs (top URL has 10 of 9,839 events).
[...]

PR #7025, opened 2026-09-01
Fix Sentry noise: 'TypeError: Failed to fetch'

Formasitchijoh:
## What this PR does

Addresses [#7018](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/7018) ("TypeError: Failed to fetch" dominating the Sentry error quota), with context from the companion issue [#7017](https://github.com/WikiEducationFoundation/WikiEduDashboard/issues/7017).

Investigation (via live Sentry data, not just the stack traces) found several `request()`/`fetch()` call sites with no `ensureOk` check and no `.catch()`, so failures surfaced as raw unhandled promise rejections — pure Sentry noise, no user-facing handling. Fixed each on its own merits:

- **`user_profile.jsx`** (PEONY-2NQ/2NT) and **`helpers.js#canUserCreateAccount`** (PEONY-2P4): now catch locally via `ensureOk`. Neither explicitly reports to Sentry — not needed once handled, and that's a deliberate per-site call rather than automatic. `helpers.js` fails open (`return true`), matching its existing de-facto behavior.
- **`media_wiki_classes.js`** (PEONY-2P5): same fix, plus a real correctness bug beyond the Sentry angle — one failing wiki no longer kills assessments for the whole batch via `Promise.all`'s fail-fast behavior.
- **`wikidata_actions.js`** (PEONY-2P3/PEONY-3BC) is **not** suppressed — it's the highest-volume issue currently hit and is under active investigation, so it's left reporting on purpose. Its catch and `sentry.js`'s `normalizeUnhandledRejection` (for anything else that goes truly unhandled app-wide) now attach `onLine`, `visibilityState`, `hasServiceWorkerController`, and the failed request URL — the exact diagnostics #7018 asks for to separate "user went offline / navigated away" from "request actually blocked or intercepted." `sentry.js` also aggregates these into one Sentry issue via a fixed fingerprint instead of letting them split by stack trace.

Two approaches were considered and rejected: centralizing the reporting inside `request.js`/`ensureOk` itself (would have double-reported errors already handled locally, or started reporting ~20+ previously-silent call sites and worsened the quota problem this PR exists to fix), and an `AbortError` guard added early on and then removed everywhere (none of the touched call sites use `AbortController`, so it was defensive code for a scenario that can't currently happen — real precedent for that guard exists only in `commons_image_picker.jsx`, which does use one).

## AI usage

This PR was prepared extensively with Claude Code (Claude Sonnet 5, via Claude Code CLI), across a long investigative session rather than a quick code-generation pass. Claude used the Sentry MCP integration to query live production issue data (event counts, browser/OS/geo breakdowns, breadcrumbs) directly rather than relying on pasted stack traces alone, cross-referenced that against the GitHub issue text, and used a background code-search agent to audit roughly 45 `request()`/`ensureOk` call sites across the codebase for regression risk before touching anything.

The human reviewer caught and corrected real mistakes along the way: rejected an initial proposal to centralize error reporting inside `request.js` (would have caused Sentry double-reporting or a quota-worsening spike in newly-reported call sites), had Claude remove a speculative `AbortError` guard added without checking whether any touched file actually used cancellation, and — after reading the issue text more closely than Claude had — caught that Claude's diagnostic enrichment was missing the literal "failed URL" the issue asked for, not just a derived cross-origin boolean. Claude also independently found and fixed one call site (`media_wiki_classes.js`) that the original file audit had missed due to a same-directory relative-import blind spot in the initial `grep`.

This branch has 4 commits, all landed within about a 10-minute window at the end of a much longer conversation — the commit timestamps alone understate the actual review effort, since most of the time went into investigation and back-and-forth correction before anything was committed.

This PR description was drafted using the Claude Code `/prepare-pr` skill.

## Screenshots

No UI changes — this is entirely error-handling and Sentry-diagnostic logic. Nothing to show before/after.

- `wikidata_actions.js`/PEONY-3BC is intentionally left reporting to Sentry rather than fixed — the actual decision (suppress it, block the crawler at the web-server level, or use a Sentry inbound filter) is still open pending a few days of the newly-enriched diagnostic data, per the "gather info before deciding" approach agreed on in #7017/#7018.


Formasitchijoh:
@ragesoss Once these two fix are deployed some of the errors will be handled and we can get more details into handling the prominent once as per the suggestions #7024

ragesoss (changes requested):
Reviewed with Claude Code. Findings below are provisional AI-assisted analysis — offered to read and weigh, not to be trusted automatically. Push back on anything that looks wrong; the volume figures in particular are worth re-checking in Sentry directly.

Two blocking items, plus three observations that don't block.

### 1. Rebase needed, and the obvious conflict resolution reverts #7024

#7024 merged after this branch was pushed, and it edits the same `Sentry.captureException` call. The conflict is in `wikidata_actions.js` and its spec.

The important part: this branch's side of the conflict does not include `responseText`, which is the entire purpose of #7024. Taking this side wholesale silently undoes it. The spec conflict compounds that — this branch loosens the assertion from an exact match to `expect.objectContaining({ onLine, visibilityState })`, which would let the regression pass CI.

The resolution wants the union of both extras: `responseText`, `url`, `onLine`, `visibilityState`, `hasServiceWorkerController`, with the spec asserting the full object. That was verified locally — the PR's own specs and the full 472-test JS suite pass with it.

### 2. `failedFetchUrlInfo` can attribute the wrong URL

`sentry.js` selects the most recent `category === 'fetch'` breadcrumb. Per Sentry's own instrumentation (`@sentry/browser/cjs/integrations/breadcrumbs.js:252-290`), a failed fetch breadcrumb is tagged `level: 'error'` with no `status_code`, while a successful one carries a `status_code` and no level. Breadcrumbs are appended in settle order, so any request that settles after the failed one wins — and `fetchWikidataLabels` itself chunks into multiple concurrent requests, so this is not an unusual interleaving.

The effect is that the URL and `crossOrigin` flag can describe a request that succeeded. Since the failed URL is the specific diagnostic #7018 asks for, that undercuts the goal.

Filtering on the error level fixes it:

```js
const lastFetch = breadcrumbs && [...breadcrumbs].reverse()
  .find(b => b.category === 'fetch' && b.level === 'error');
```

This spec fails against the current implementation and passes with that change:

```js
import '../app/assets/javascripts/sentry';

const buildEvent = (value = 'TypeError: Failed to fetch') => ({
  exception: {
    values: [
      { type: 'UnhandledRejection', value, mechanism: { type: 'onunhandledrejection', handled: false } },
    ],
  },
});

test('attributes the failed URL, not a later successful one', () => {
  const event = buildEvent();
  event.breadcrumbs = {
    values: [
      // The request that actually failed.
      { category: 'fetch', level: 'error', data: { url: 'https://www.wikidata.org/w/api.php' } },
      // A same-origin request that merely settled afterwards.
      { category: 'fetch', data: { url: `${window.location.origin}/articles.json`, status_code: 200 } },
    ],
  };
  const hint = { originalException: new TypeError('Failed to fetch') };

  const result = window.normalizeUnhandledRejection(event, hint);

  expect(result.extra.url).toBe('https://www.wikidata.org/w/api.php');
  expect(result.extra.crossOrigin).toBe(true);
});
```

### 3. `extra` is not indexed by Sentry; tags are

`event.fingerprint = ['unhandled-failed-to-fetch']` collapses every call site into a single issue, while the values that would distinguish them sit in `extra`, which Sentry does not make searchable or aggregatable. The offline-versus-blocked split #7018 asks for would then require opening events one at a time. The low-cardinality fields — `onLine`, `visibilityState`, `crossOrigin`, `hasServiceWorkerController` — would support that analysis directly as tags.

Separately, `extra.url` is easy to confuse with Sentry's built-in `url` tag, which holds the page URL. `requestUrl` or `failedUrl` would read less ambiguously.

### 4. Volume context, which may affect how this is prioritized

Production numbers over the last 24 hours: all `Failed to fetch` issues combined produced 28 events, twenty of which are the wikidata handled path this PR intentionally keeps. PEONY-3BC produced 166 over the same window.

The ~10.7k/30d figure behind #7018 appears to be largely a decayed historical tail rather than current load. Commit `7859c1e0f` (Aug 19) added the wikidata catch, and PEONY-2P3's unhandled count has gone 9631 over 30d, to 233 over 7d, to 0 over the last 24h. On those numbers this PR removes roughly 8 events per day.

None of that argues against the changes — the call-site fixes are correct regardless, and the `beforeSend` branch is reasonable as a forward-looking safety net. It does suggest the framing overstates present-day quota impact, and that PEONY-3BC is where the remaining pressure is.

### 5. Minor

`error.url = error.url || url` in `wikidata_actions.js` mutates the caught error. If anything ever rejects with a non-object, that assignment throws in strict mode and masks the original error. Low probability given the current call sites.

---

Confirmed working as described: `canUserCreateAccount`'s fail-open matches the prior de-facto behavior, since an unhandled rejection previously left `disabled` at its `false` initial state. `media_wiki_classes` returning nothing on failure is safe because `getAssessments` merges via `Object.assign`, which ignores `undefined`, and dropping the `Promise.all` fail-fast is a real fix. ESLint is clean across all nine files.

*Drafted in a Claude Code session (~15 min, 3 interactions). Sage directed the review and approved posting after a read-through, but did not independently verify the findings. Claude Code ran ESLint and the full 472-test JS suite, resolved the #7024 conflict locally and confirmed the union passes, read the `@sentry/browser` breadcrumbs source to establish the failed-breadcrumb shape, wrote and ran the spec above (it is AI-generated and was executed locally against both the current and corrected implementations), and queried live Sentry for the event volumes in section 4.*

ragesoss (changes requested):
Re-reviewed with Claude Code. As before, these findings are provisional AI-assisted analysis — worth reading and weighing, not trusting automatically. Push back on anything that looks wrong.

The delta since the last review is the master merge plus one line, so this is a short follow-up.

### Addressed

`responseText` is back in the `Sentry.captureException` extras in `wikidata_actions.js`, so #7024's runtime behavior is restored. The merge itself is clean and CI is green.

### 1. #7024's regression guard is still missing

The production code is fixed, but the spec that protected it is not. Master's version of the 403 test asserted `extra` exactly:

```js
{ extra: { responseText: 'blocked', url: 'https://www.wikidata.org/w/api.php?action=wbgetentities' } }
```

This branch still carries the loosened form introduced by the conflict resolution:

```js
{ extra: expect.objectContaining({ onLine: true, visibilityState: 'visible' }) }
```

So the one test where `responseText` carries a real value no longer checks it. Deleting `responseText: error.responseText` from `wikidata_actions.js` entirely leaves all four specs in that file green — that was run locally to confirm, not inferred.

The `responseText: undefined` added to the *network-failure* test does not cover this. `toHaveBeenCalledWith` uses `toEqual` semantics, under which a missing key and an explicitly-`undefined` key compare equal, so that line documents intent but cannot fail. The same experiment above demonstrates it: that test passed with the property removed from the source.

Restoring the exact assertion on the 403 test closes it:

```js
    expect(global.Sentry.captureException).toHaveBeenCalledWith(
      expect.objectContaining({ name: 'ApiError', status: 403, statusText: 'Forbidden' }),
      {
        extra: {
          url: 'https://www.wikidata.org/w/api.php?action=wbgetentities',
          responseText: 'blocked',
          onLine: true,
          visibilityState: 'visible',
          hasServiceWorkerController: false,
        },
      }
    );
```

That was verified in both directions locally: it passes against the current code and fails with `responseText` removed.

### 2. Items 2, 3, and 5 from the prior review appear untouched

`sentry.js` is byte-identical to the reviewed commit, and `error.url = error.url || url` is unchanged. Specifically still open:

- **Wrong URL attribution** — `failedFetchUrlInfo` still selects the most recent `category === 'fetch'` breadcrumb without filtering on `level === 'error'`, so a request that settles after the failed one wins. The spec from the prior review was re-run against the current branch and still fails the same way, resolving to `http://localhost/articles.json` rather than the wikidata URL. Since the failed URL is the specific diagnostic #7018 asks for, this is the item most worth landing before deploy.
- **`extra` is not indexed by Sentry** — the fingerprint collapses call sites into one issue while the distinguishing values sit in `extra`, which is not searchable or aggregatable. The offline-versus-blocked split would need events opened one at a time.
- **`error.url = error.url || url`** mutates the caught error; low probability given current call sites.

No signal either way on these in the PR, so it is unclear whether they are rejected, deferred, or simply not yet done. Any of those is a fine answer — a note saying which would help.

### Checks run

`yarn test` passes (475 tests, 76 suites). ESLint is clean across all nine changed files.

*Drafted in a Claude Code session (~10 min, 2 interactions). Sage directed the re-review and approved posting after reading the summary, but did not independently verify the findings. Claude Code ran the full JS suite and ESLint, diffed the branch against the previously-reviewed commit to establish which files changed, removed `responseText` from the source and reran the specs to confirm the guard no longer fires, verified the proposed assertion in both directions, and re-ran the prior review's breadcrumb spec against the current branch. The spec code above is AI-generated and was executed locally; it was not reviewed line-by-line by a human.*

Formasitchijoh:
@ragesoss  this is ready,  turns out I committed and forgot to push

ragesoss (approved):
Re-reviewed with Claude Code. As before, these findings are provisional AI-assisted analysis — worth reading and weighing, not trusting automatically.

### Addressed

All items from the prior two reviews have landed:

- The 403 spec asserts `extra` exactly again, so `responseText` is protected. Deleting it from the source fails the test — that was checked locally.
- `failedFetchUrlInfo` filters on `level === 'error'`, with the attribution spec included. Removing the filter fails that spec.
- Low-cardinality fields are tags in both `sentry.js` and `wikidata_actions.js`; the URL is renamed to `failedUrl` / `requestUrl` and stays in `extra`.
- The inner `error.url` mutation is guarded.

### One optional follow-up

The guard covers the inner assignment, but the outer `.catch` in `fetchWikidataLabels` still reads `error.url` and `error.responseText` off the same value. With `request` stubbed to reject `null`, the result locally was `TypeError: Cannot read properties of null (reading 'url')` as a fresh unhandled rejection, and `captureException` never fired. Nothing in `request.js` or `ensureOk` can reject with a non-Error today, so this is the same low-probability caveat as before. `error?.url` / `error?.responseText` would close it if wanted; not blocking.

### Checks run

`yarn build` succeeds. ESLint is clean across all nine changed files. `yarn test` passes (488 tests, 78 suites). An exploratory spec also confirmed the `beforeSend` branch fires end to end through a real `@sentry/browser` 7.120.4 client with its own `GlobalHandlers`, and that `undefined` `crossOrigin` / `failedUrl` are dropped at serialization rather than sent.

*Drafted in a Claude Code session (~10 min, 2 interactions). Sage directed the re-review and approved posting after reading the summary, but did not independently verify the findings. Claude Code diffed the branch against the previously-reviewed commit, ran the build, ESLint, and the full JS suite, mutation-tested both restored guards in both directions, and wrote and ran an AI-generated exploratory spec (not reviewed line-by-line by a human) covering the edge cases above.*

PR #7032, opened 2026-09-03
Scope the Article Finder search button styles so autocomplete suggestions render as a list

### LINKED ISSUE #7031: Drop-down list in "Article Finder" component is scrambled
### Issue

The dropdown list that appears when you start typing to find an article is really broken. I think it is part of the auto-complete feature.

### Additional Information
Image

### Environment
OS: Windows 11 pro, 
IDE: VS code.

### Component
Article Finder

samuelmbabhazi:
Fixes #7031

## What was wrong

`_article_finder.styl` styles the search bar's Search button with a bare `button` selector inside `.article-finder-search-bar` (position absolute, right 0, 8rem wide, purple background, white bold text). That rule predates 3d98791d, which turned the autocomplete suggestions into `<button class="autocomplete-item">` elements for accessibility. Since then every suggestion inherits the Search button styling: all of them are absolutely positioned at the same spot under the Search button, 8rem wide, purple with white text, which is the "scrambled" dropdown in the screenshot.

## The fix

- Scope the Search button rules to the direct child (`> button`, `> button:disabled`) so they no longer reach the suggestions.
- Let the suggestions render as list rows again (`display block`, `width 100%`, `text-align left`); they keep the existing padding, cursor and hover styles.

No JavaScript changes, so the Jest and ESLint suites are unaffected.

## How it was verified

Compiled the complete `main.styl` cascade (reset, utils, base, all modules) with the original and the patched partial. The diff between the two compiled stylesheets is exactly these lines. Rendering the component's markup with the full stylesheet in headless Chromium:

- before: the three suggestions are `position: absolute`, at the same coordinates, 120px wide, background `rgb(103, 110, 180)`, white text
- after: `position: static`, full width of the dropdown, stacked as rows, transparent background, regular text color, hover highlight intact

`.autocomplete-item` and `.autocomplete` are not used anywhere else in the codebase, and the only other button inside the search bar is the Search button itself.

## AI usage
I used Claude Code (Anthropic) as a coding assistant on this PR: to trace the regression to the commit that turned the autocomplete suggestions into buttons, to draft the stylesheet change, and to build the local verification harness (compiling the full Stylus cascade at the pre-regression commit, master, and this branch, then rendering the component in headless Chromium and comparing computed styles). I reviewed and tested the change myself; the before and after screenshots below come from that rendering. No user-facing text was written or changed.

## Screenshots
Before:

Article Finder autocomplete before the fix: suggestions stacked under the Search button

After:

Article Finder autocomplete after the fix: suggestions rendered as a list

### DIFF
--- a/app/assets/stylesheets/modules/_article_finder.styl
+++ b/app/assets/stylesheets/modules/_article_finder.styl
@@ -35,9 +35,9 @@
     border-radius 5px 5px 5px 5px
     padding 1rem 11rem 1rem 2rem
     transition border-radius linear 0.1s
-  button:disabled
+  > button:disabled
     opacity 50%
-  button
+  > button
     position absolute
     right 0
     height 100%
@@ -69,6 +69,10 @@
     box-shadow 0px 2px 4px rgba(0, 0, 0, 0.1)
     z-index 5
     .autocomplete-item
+      display block
+      width 100%
+      text-align start
+      line-height inherit
       padding 0.5rem 2rem
       cursor pointer
     .autocomplete-item:hover


MasterIfeanyi:
@ragesoss please review.

Formasitchijoh:
@samuelmbabhazi  for UI based fix, do provide a before and after screenshot or video as stated in the PR template

samuelmbabhazi:
> @samuelmbabhazi for UI based fix, do provide a before and after screenshot or video as stated in the PR template

**Before**
af-full-before-v2

**After**
af-full-after-v2

ragesoss:
Thanks @samuelmbabhazi ! Please also include the AI disclosure section in the PR description (see https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/master/PULL_REQUEST_TEMPLATE.md).

ragesoss (approved):
The fix is correct and minimal. Scoping the Search button rule to `> button` and giving `.autocomplete-item` block layout restores the dropdown without touching any other selector. The compiled CSS was checked in both the LTR and RTL bundles and only the expected rules changed.

Two notes, neither blocking:

- The regressing commit named in the description is not the one that introduced the bug. 3d98791d only added `type="button"` to the Search button. The suggestions were converted from `<div>` to `<button>` in 2bf734e9e (May 2026, jsx-a11y remediation), which stated that no CSS changes were needed because the class selectors match both elements. That claim missed this bare `button` rule.
- The PR has no regression test, and the existing `article_finder_spec.rb` never opens the dropdown. An exploratory spec written during this review opens the autocomplete, checks that suggestions are `position: static` and stacked on distinct rows, and clicks one through to results. It passes 5/5 on this branch and fails 3/5 against master (on master the click is intercepted by the overlapping sibling, so suggestions were unclickable, not just misrendered). It could be added as `spec/features/article_finder_autocomplete_spec.rb` if a regression guard is wanted; happy to share it.

A note on AI-driven review feedback: it is not meant to be trusted automatically or acted on blindly. Read it, understand it, and use or ignore it based on your own judgment.

*Drafted in a Claude Code session (~15 min, 2 interactions). Sage read the draft and approved posting. The exploratory spec is AI-generated and was run locally by the session on both this branch and master; Sage did not independently rerun it.*

PR #7034, opened 2026-09-03
Pangram 4 client, AI detector comparison system, and Originality.ai refresh

ragesoss:
## What this PR does

Pangram 3.3.2, the AI detector behind the Dashboard's AI edit alerts, shuts down on **September 30, 2026**, and Pangram 4 is only served through an asynchronous task endpoint. This PR is everything needed to compare detectors on real student edits *before* alerting is switched over, shipped as one rollout:

**1. Pangram 4 client** (`cd04b1c5c`)
- `PangramApi.v4` submits to the task endpoint with `model: pangram-4`, polls every 0.5 s (60 s ceiling), and returns the same parsed hash callers already expect. Non-success responses raise `PangramApi::RequestError` (carrying the status); a terminal task failure raises `TaskFailed`; running out of time raises `TaskTimeout`. `PangramApi.v3` remains.
- `PangramResponseParser` adds `humanized_window_count` and `max_humanizer_score` to alert details only when the response carries Pangram 4's per-window humanizer fields, so Pangram 3 alert details are unchanged.

**2. Detector comparison system and Originality.ai refresh** (`e37d5e30e`, `106544c1d`, `10a6f1f46`)
- **Registry and shared summary.** `AiDetector` (`lib/ai/ai_detector.rb`) maps each `RevisionAiScore#check_type` to an API client, a response parser and a vendor. Both parsers implement `DetectorSummary`, a vendor-neutral set of keys (max/mean window score, document score, label, fractions, humanizer counts, report URL), so exports and analysis never depend on the vendor. Vendor operating policy (admin partial, minimum input, credit estimation, balance lookup, error classes) lives in `AiDetector::Vendor`, so adding a detector is a client, a parser and one registry entry, and adding a vendor is one more struct. `CheckRevisionWithPangram::DETECTOR_KEY` becomes the single switch for the eventual cutover; production alerting **still uses Pangram 3** and the alert rule (max window score > 0.9) and emails are unchanged.
- **Originality.ai.** AI Allowance mode is added as two detectors, thresholds 15 (Originality's default) and 40 (their minimize-false-positives setting), alongside `turbo`, `academic` and `lite-102`. Live probes showed that an unrecognized `aiModelVersion` silently falls back to `lite-102` with no error, so the client asserts that the echoed `results.ai.aiModel` matches the request and raises `ModelMismatch` otherwise. It also refuses texts under Originality's 50-word minimum without sending a request, raises `RequestError` for HTTP and in-body errors, and can read the account's credit balance. `lite` (undocumented which version it maps to) is no longer selectable; its key is kept for historical rows.
- **Admin AI tools page** (`/ai_tools`) runs off the registry. Only the Pangram detectors are ticked by default, since Originality credits are limited. A detector error, including transport failures, renders as a notice for that detector instead of failing the whole request. The Originality partial shows the model actually used and per-block scores.
- **Samples.** A new `ai_detection_samples` table holds named sets of text units: the exact text sent to detectors (MEDIUMTEXT; cumulative course diffs exceed 64 KB), where it came from (wiki, revision, diff, term), `ground_truth` (`human`, `ai`, `ai_assisted`, or blank) separately from `provenance` (how we know: `pre_llm_term`, `staff_confirmed`, `experiment`, `synthetic`, `self_report`, or any description an ad-hoc source needs), free-text `notes`, and `factors`: named values such as `topic`, `author`, `model` or `prompt` that link units sharing a value, so pairs and groups need no extra tables. Four builders assemble samples: production-scored edits stratified by score band, alerts with student follow-up answers, the March 2026 cumulative-diff-by-term method, and generic rows naming units by URL or by text (also from a CSV). Student questionnaire answers are recorded as `self_report` provenance with no ground truth; they are never treated as truth. Text is fetched once at sampling time; units under 500 characters are skipped; units are never duplicated. Nothing is specific to any exemplar source: fetching and preprocessing from another wiki, a dump or a generation run happens outside the Dashboard and arrives as rows.
- **Scoring, export, import.** `ScoreAiDetectionSample` sends a sample to any set of detectors, storing results as `revision_ai_scores` rows with `check_origin = 'detector_comparison'` and a `sample_id`. It is resumable, retries failures in place, and its dry run reports pending words, estimated credits and the current balance for every vendor that meters, before anything is spent. `ExportAiDetectionComparison` writes a long-format CSV (one row per unit and detector, with a `factor_<name>` column per factor present). `ImportDetectorComparisonCsv` loads the March 2026 comparison CSV with the Pangram 3 / Originality Turbo / Academic scores it already holds, so only new detectors need to run on those 1000 units.
- **Analysis.** `docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/analyze.py` (pandas/matplotlib) produces positive rate and mean scores by term, ECDF score distributions split by the pre-ChatGPT baseline (human units with `pre_llm_term` provenance, where the expected hit rate is zero), threshold sweeps (false-positive rate on the baseline against positive rate elsewhere), pairwise scatters with a CSV of disagreements linking to both vendors' reports, a per-case report of challenge cases with each detector's verdict, a list of disputed alerts a detector still flags (candidates for a person to confirm), `--group-by` for rates per factor value (e.g. model, prompt, post-processing) and `--pair-by` for human-versus-AI units sharing a factor (e.g. topic) with a slope chart. The `README.md` in that directory documents the console workflow, the credit rules and guidance for synthetic exemplars.
- **`WikiUrlParser#revision_target`** centralizes the diff/oldid conventions. `diff=prev&oldid=X` now resolves to X against its parent (it previously produced `from_rev: 0`, which fetched the whole article); `diff=next` yields no target.

Existing stats pages filter on `check_origin`, so comparison rows do not appear in production score statistics.

References:
- [Pangram 4 migration guide](https://www.pangram.com/blog/pangram-4-migration-guide), [Pangram REST API reference](https://pangram.readthedocs.io/en/stable/api/rest.html)
- [Originality.ai: AI Allowance](https://originality.ai/blog/introducing-ai-allowance), [Originality.ai model accuracy post (July 2026)](https://originality.ai/blog/ai-accuracy). Their API documentation site is rendered client-side and could not be read by tooling; the parameter behavior described above comes from live probes on 2026-09-03.
- Plan documents: `.claude/pangram_4_migration_plan-2026-09-03.md`, `.claude/detector_comparison_system_plan-2026-09-03.md` (not committed).

## AI usage

Claude Code wrote all of the code, specs, documentation, the Python analysis script, the commit messages and this description, under direction from Sage Ross. Sage set the goals (a Pangram 4 client first, then a general system for comparing detectors so new models can be evaluated consistently, then a refresh of the Originality.ai integration), pointed to a March 2026 one-off comparison as the template, and made the design decisions: store the sent text in the database; ship everything on this branch as one rollout before the alert cutover; keep alert semantics and email copy unchanged at cutover; focus the Originality comparison on AI Allowance with conservative credit use (small samples, at most two Originality checks per unit, thresholds 15 and 40); never treat self-reported questionnaire answers as ground truth; allow a few synthetic known-AI exemplars varied the way students would vary them; build nothing around any particular exemplar source; and make linkage between exemplars expressible (which became factors).

Claude Code ran live probes against both vendors' APIs to establish facts the documentation did not provide: that Pangram 4 is async-only with compatible field names, and for Originality the silent model fallback, the snake_case allowance parameters, the 50-word minimum, that each allowance threshold re-scores the blocks (so thresholds cannot be re-derived offline), and that the four classic model identifiers returned identical scores on the probe texts. The Originality probes spent roughly 70 credits. Sage supplied the Originality accuracy post and the threshold decision mid-way.

A review of the first two commits written by GPT-5.6 and posted by Sage raised five findings (two data-integrity bugs, three design issues); all were judged valid and addressed in the fourth commit.

Scale of effort: four commits on 2026-09-03 across two working sessions, with Sage steering by short messages. Test trajectory: green on the first run for the registry work; three failures in the services from a spec stub keyed on an exact timestamp, passing after the stub was fixed; the later batches passed on their first runs. Final state across the touched suites: 152 examples, 0 failures; RuboCop clean.

The "What this PR does" summary and the rest of this description were also drafted by Claude Code and may contain errors. This PR description was drafted using a Claude Code skill (`/prepare-pr`).

## Screenshots

From a disposable screenshot spec run against stubbed API responses recorded from the live Pangram 3, Pangram 4 and Originality.ai endpoints. The page is admin-only and did not exist in this form on master.

**01 AI tools form.** Detectors come from the registry; only the Pangram boxes are ticked by default.

| Before | After |
|--------|-------|

**02 Pangram 3 and Pangram 4 results side by side**

| Before | After |
|--------|-------|

**03 Originality AI Allowance result** with the model actually used and per-block scores.

| Before | After |
|--------|-------|

- **Cutover is a separate PR.** After this deploys, the plan is to import the March 2026 sample, build a sample of recent production-scored edits, score both with Pangram 4 and the two Originality allowance levels in the production console (workflow in the detector comparison README), analyze, and only then change `CheckRevisionWithPangram::DETECTOR_KEY`. The Pangram 3 host goes away on September 30, so that has to land within the month.
- **Originality classic models scored identically.** `turbo`, `academic`, `lite` and `lite-102` returned byte-identical block scores on two very different probe texts (only `multilang` differed). Either they now route to one model or the echo is cosmetic. Worth a question to Originality support before spending credits on more than one classic model; the comparison plan uses only the allowance levels for new Originality spend.
- **Credit estimate is approximate.** The dry run assumes about one credit per 100 words per check and roughly double for an allowance scan, which matched the probes (90 words: 1 credit; 135 words with allowance: 3) but is not documented by Originality. Check the reported balance before and after the first real run.
- **Behavior change for `diff=prev` URLs on the admin page.** They now fetch the single edit rather than the full revision, which is what the URL denotes but differs from what the page did before.
- **`schema.rb` was hand-edited.** Rails 8.1 re-dumps every table with alphabetical columns, which turned the migration into a 600-line diff; the file was restored and only the new table, column, indexes and version were added. The next migration on master will produce the big re-dump for whoever runs it. The migration was amended in place while unshipped, so any checkout that ran the earlier version needs a rollback and re-migrate.
- **Originality `average_ai_likelihood` semantics are unchanged**: it is the vendor's whole-text confidence, not the mean of block scores. The summary carries both (`document_score` and `mean_window_score`) so analysis can pick, but the stored column means different things for Pangram and Originality rows, as before.
- **Window sizing.** Pangram 4 returned one 439-word window where Pangram 3 produced 405 + 34 on the same text. The alert emails describe "~400-word segments"; that copy is deliberately left alone for now and can be revisited with comparison data.
- The March 2026 CSV import re-fetches plaintext for its 1000 units from Wikipedia (roughly 20–30 minutes in a console) and depends on those revisions still being visible; deleted revisions are skipped and reported.


ragesoss:
[gpt-5.6-sol review]

I reviewed this both as a conventional code review and against the goal of making future detector comparisons easy and reproducible. The registry/sample/summary direction is solid, but I found several issues I think should be addressed before relying on this workflow:

1. **P1 — Analysis crashes when a detector has no report URLs.**

   [`pairwise()`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/e37d5e30ec2dcbdbe0f52378b1edb2ec59041d3d/docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/analyze.py#L214-L238) builds `reports` with `pivot_table`, which drops detector columns whose report URLs are entirely blank. The March importer does not populate report URLs, so a disagreement involving an imported detector reaches `reports.at[unit_id, a]` and raises `KeyError`. I reproduced this with an imported-style Pangram 3 row and a Pangram 4 row. Missing report links should remain nullable rather than aborting the analysis.

2. **P1 — `diff=prev&oldid=X` silently samples the whole article instead of the edit.**

   [`revision_target`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/e37d5e30ec2dcbdbe0f52378b1edb2ec59041d3d/lib/utils/wiki_url_parser.rb#L38-L49) converts `prev` to `0`, yielding `from_rev: 0`. `GetRevisionPlaintext` interprets that as the first revision and fetches the entire revision HTML. These are common Wikipedia diff URLs, so a URL-based dataset can silently contain fundamentally different text from what its URL denotes. For `diff=prev`, it should pass `rev_id: oldid`, `from_rev: nil`, and `diff_mode: true`, allowing the normal parent lookup.

3. **P2 — A third vendor cannot be added through only the documented registry steps.**

   [`ScoreAiDetectionSample`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/e37d5e30ec2dcbdbe0f52378b1edb2ec59041d3d/app/services/score_ai_detection_sample.rb#L39-L65) treats every non-Pangram detector as Originality: it imposes Originality's minimum, queries Originality's balance, and calls `client.expected_model` for credit estimation. A new detector implementing only the documented `#inference` interface will therefore crash during dry-run. The [admin results view](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/e37d5e30ec2dcbdbe0f52378b1edb2ec59041d3d/app/views/ai_tools/show.html.haml#L33-L39) likewise sends every non-Pangram result to the Originality partial. Detector-specific capabilities such as minimum input, cost estimation, error classes, and presentation should live in the registry/definition or optional adapters.

4. **P2 — Ordinary transport failures can turn the admin comparison into a 500.**

   [`detect_ai`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/e37d5e30ec2dcbdbe0f52378b1edb2ec59041d3d/app/controllers/ai_tools_controller.rb#L40-L43) catches vendor exceptions but not `Faraday::TimeoutError`, `Faraday::ConnectionFailed`, or malformed JSON. The thread pool swallows those exceptions, leaving neither a result nor an entry in `@errors`; the view then passes `nil` to a detector partial. Catching the same transport/parsing failures handled by the batch scorer, or rendering only when a result exists, would avoid this.

5. **P2 — `--rule vendor` is ignored for pairwise agreement and disagreements.**

   Although `load()` calculates `positive` from vendor labels, [`pairwise()`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/e37d5e30ec2dcbdbe0f52378b1edb2ec59041d3d/docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/analyze.py#L214-L237) always recalculates positives using `max_score > threshold`. Consequently, a report headed `rule: vendor` contains pairwise statistics generated by a different rule. Pivoting the already-calculated `positive` column would keep the outputs consistent.

Overall, storing sampled text and exposing a common summary are strong foundations. I would address the two silent data-integrity failures and move vendor-specific operational policy out of the generic scorer before relying on this for model evaluation.

Validation: both GitHub CI jobs passed, and RuboCop passed on the touched Ruby implementation files. Focused local RSpec execution was blocked by the local test database having the new migration pending.

ragesoss:
[gpt-5.6-sol re-review]

I re-reviewed the PR at `10a6f1f4`. The five findings from my first pass are substantively fixed. The ground-truth/provenance split and vendor-policy abstraction are strong improvements.

I found three new functional issues in the additional work:

1. **P2 — Self-report candidates are always omitted.**

   [`self_report_candidates()`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/analyze.py#L357-L366) searches for the literal substring `"self_reported_false_positive": true`, including a space after the colon. However, the [exporter serializes `unit.metadata` with `to_json`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/app/services/export_ai_detection_comparison.rb#L51-L58), which produces compact JSON such as `{"self_reported_false_positive":true}`. I reproduced this with the analysis function, and the candidate list was empty. Parsing the metadata as JSON—or exporting a dedicated boolean column—would avoid depending on serialization whitespace.

2. **P2 — A `diff=next` URL still causes the admin tool to score unrelated text.**

   [`WikiUrlParser#revision_target`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/lib/utils/wiki_url_parser.rb#L43-L63) correctly returns `nil` for `diff=next`, but [the controller treats every nil target as an article URL and loads the latest revision](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/app/controllers/ai_tools_controller.rb#L46-L57). Thus `?title=Foo&diff=next&oldid=X` silently scores the current latest full revision instead of the requested diff. The latest-revision fallback should apply only when the URL has no revision/diff selector; otherwise this case should be rejected with a notice or explicitly resolved.

3. **P2 — Challenge-case verdicts can be assigned to the wrong units.**

   In [`challenge_report()`](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/analyze.py#L337-L354), the detector scores are assigned by index, but the verdict list is built positionally from `zip(wide[detector], flagged_wide[detector], expected)`. The pivoted `wide` frame sorts its index, while `expected` retains CSV/input order. With reordered input, I reproduced a known-positive AI pangram being reported as `missed` and a human negative as `false positive`. Reindexing all three inputs to `report.index`, or constructing an indexed verdict Series, would keep labels attached to their units.

Two smaller follow-ups:

- The [README still says that self-report becomes ground truth](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/docs/analytics_scripts/ai_usage_research/detector_comparison/README.md#L60-L64), which contradicts the improved implementation where self-report is provenance and `ground_truth` remains unset.
- The chart labels hardcode “pre-LLM baseline” in both the [agreement chart](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/analyze.py#L176-L195) and [score-distribution chart](https://github.com/WikiEducationFoundation/WikiEduDashboard/blob/10a6f1f4672aca68c81498bacc9ab67a2412db75/docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/analyze.py#L278-L295), even when the user supplies a custom `--baseline-provenance` value.

Validation: RuboCop passed on the 18 touched Ruby implementation files. I also directly reproduced the self-report filtering and challenge-verdict alignment failures described above. The updated GitHub CI jobs were still pending when I reviewed this head.

PR #7035, opened 2026-09-03
Add self-service copying of Available Articles between courses

ragesoss:
## What this PR does

Adds a self-service way for instructors to reuse a previous course's Available Articles list. On the Articles tab → Available Articles, a new **Import from another course** button opens a dialog where the instructor picks one of their other courses from a dropdown (or pastes any course URL), optionally ticks **Include assigned articles?** to also bring over the articles students in that course were assigned, sees how many articles would be added (net of ones already present in this course), and confirms. Everything lands as Available Articles in the current course.

Until now the only options were the clone-time "Clone Available Articles list" checkbox (new courses only) or the console script in `docs/course_update_scripts/copy_assignments.rb`.

Implementation notes:

- `CopyAvailableArticles` service creates the records directly from the source rows (`article_title`/`article_id`/`wiki_id`, `flags: {available_article: true}`) — no `AssignmentManager`, so no per-title wiki API import. It dedupes by `(article_title, wiki_id)` against the target and within the source, and uses non-bang `create` so a legacy title that no longer validates (or a concurrent add) is counted as skipped rather than aborting the batch. `dry_run:` powers the preview.
- `CopyAvailableArticlesController` exposes `GET /copy_available_articles/preview.json` and `POST /copy_available_articles.json`. The target requires `can_edit?`; the source must be non-private or the user must be a member — the same visibility rule as `/courses/:slug/assignments.json`. Unknown and hidden sources both get the existing "No such course … exists." 404 so a private course's existence isn't revealed.
- `CourseUrlParser` (in `lib/utils`) resolves a bare slug, a `courses/…` path, or a full pasted course URL.
- New `Assignment.available` scope (`role: 0, user_id: nil`).
- The clone path (`CourseCloneManager#copy_assignments`) now delegates to the same service, so cloned Available Articles carry the `available_article` flag that the Remove button's unclaim-vs-destroy logic depends on. This also surfaced and fixes a pre-existing bug: the frontend always sends `copy_assignments` as the string `'true'`/`'false'`, and `'false'` was truthy, so every clone copied the articles regardless of the checkbox.
- Frontend: `copy_available_articles.jsx` mounted in `available_articles.jsx` behind the same `isInstructor || admin` gate as the add control. The preview is fetched straight into component state rather than through redux, because `fetchAssignments(sourceSlug)` would replace this course's list. Three new i18n strings; everything else reuses existing ones (`users.number_of_articles`, `error_no_course.explanation`, `error.invalid_assignment`, `application.confirm`/`cancel`).

Specs: service, parser, request spec (auth matrix, source visibility, checkbox, worker scheduling, preview), clone-manager `clone_assignments: true` (previously uncovered), clone-controller string param, a feature spec on the Articles tab, and Jest specs for the component, the gating, and the redux action.

## AI usage

This PR was built with Claude Code under Sage Ross's direction. Claude Code drafted all of the code, specs, the Stylus module, and the commit message, and ran the test suites. Sage asked for an overview of existing article-copying support, requested an implementation plan, decided the three product questions the plan surfaced (which articles to copy — unclaimed by default with an opt-in for student-assigned ones; which sources are allowed — own courses plus any course by URL; copy-all with a count preview rather than per-article selection), approved the plan, and then steered the result: cutting the user-facing message surface from 22 strings to 3 by reusing existing strings and dropping non-essential ones, ruling out Wikidata-specific wording, choosing the generic "Invalid assignment" error, and writing all three new strings.

Scale of effort: a single commit, developed in one Claude Code session of a few hours on 2026-09-03. Planning used plan mode with three exploration subagents and one design subagent; implementation went through one build pass and two simplification rounds driven by Sage's review. Tests were run about ten times; the Ruby specs needed one fix (a missing `require`) and the feature spec one corrected expectation.

The summary above and the rest of this description were also drafted by Claude Code and may contain errors. This PR description was drafted using a Claude Code skill (`/prepare-pr`).

## Screenshots

Before: the feature did not exist — the Available Articles header had only "How to find an article", "Add available articles", and "Find Articles".

After:

**1. Available Articles header with the new button**

Available Articles header showing the Import from another course button

**2. The import dialog**

Import dialog with course dropdown, course URL field, and Include assigned articles checkbox

**3. Choosing a source course from the instructor's other courses**

Dropdown open listing the instructor's other course

**4. Preview of what would be added**

Preview line reading Northwestern University - History of Science Fall 2025: 5 articles, with OK enabled

**5. After confirming**

Available Articles list populated with the five imported titles

- **Message surface is deliberately minimal.** The same-course error reuses "Invalid assignment", the not-found error reuses "No such course … exists.", and the confirm button is the generic "OK". There's no Wikidata-specific wording ("articles" will show on Wikidata courses too) — accepted as unlikely to matter for this feature.
- **Source visibility.** Anyone who can edit the target can import from any course they could already read `assignments.json` for, including by pasting a URL for a course they aren't enrolled in. That matches existing data exposure, but it does mean admins/staff can pull lists across instructors without the console.
- **Clone-path behaviour change.** Cloned Available Articles now get `flags[:available_article]`, and the `'false'` bug fix means clones will stop copying articles when the checkbox is unticked — which is what the UI always claimed. Bundled here because the fix fell out of delegating to the shared service; happy to split it out if preferred.
- **Synchronous copy.** Roughly three queries per row; hundreds of articles complete in a few seconds. No upper bound is enforced.
- The screenshot spec (`spec/features/copy_available_articles_screenshots_spec.rb`) is disposable and not committed.

### DIFF
--- a/app/assets/javascripts/actions/assignment_actions.js
+++ b/app/assets/javascripts/actions/assignment_actions.js
@@ -51,6 +51,15 @@ export const randomPeerAssignments = randomAssignments => (dispatch) => {
     .catch(response => dispatch({ type: types.API_FAIL, data: response }));
 };
 
+// Copies Available Articles from another course into this one. The response
+// carries the full refreshed assignments list, so it replaces the current one.
+export const copyAvailableArticles = opts => (dispatch) => {
+  dispatch({ type: types.LOADING_ASSIGNMENTS });
+  return API.copyAvailableArticles(opts)
+    .then(resp => dispatch({ type: types.RECEIVE_ASSIGNMENTS, data: resp }))
+    .catch(response => dispatch({ type: types.API_FAIL, data: response }));
+};
+
 export const deleteAssignment = assignment => (dispatch) => {
   return API.deleteAssignment(assignment)
     .then((resp) => {
--- a/app/assets/javascripts/components/articles/available_articles.jsx
+++ b/app/assets/javascripts/components/articles/available_articles.jsx
@@ -5,6 +5,7 @@ import { Link } from 'react-router-dom';
 
 import AssignCell from '@components/common/AssignCell/AssignCell.jsx';
 import ConnectedAvailableArticle from './available_article.jsx';
+import CopyAvailableArticles from './copy_available_articles.jsx';
 import AvailableArticlesList from '../articles/available_articles_list.jsx';
 import MyArticlesContainer from '../overview/my_articles/containers';
 import { ASSIGNED_ROLE } from '../../constants';
@@ -25,6 +26,7 @@ const AvailableArticles = (props) => {
   }, []);
 
   let assignCell;
+  let copyCell;
   let availableArticles;
   let elements = [];
   let findingArticlesTraining;
@@ -81,6 +83,12 @@ const AvailableArticles = (props) => {
         prefix={I18n.t('users.my_assigned')}
       />
     );
+    // Same gate as the add control inside AssignCell
+    if (current_user.isInstructor || current_user.admin) {
+      copyCell = (
+        <CopyAvailableArticles course={course} course_id={course_id} current_user={current_user} />
+      );
+    }
   }
 
   const showAvailableArticles = elements.length > 0 || current_user.isAdvancedRole;
@@ -92,6 +100,7 @@ const AvailableArticles = (props) => {
           <div className="section-header__actions">
             {findingArticlesTraining}
             {assignCell}
+            {copyCell}
             <Link to={`/courses/${course_id}/article_finder`}><button className="button border small ml2">{ArticleUtils.I18n('find', project)}</button></Link>
           </div>
         </div>
--- a/app/assets/javascripts/components/articles/copy_available_articles.jsx
+++ b/app/assets/javascripts/components/articles/copy_available_articles.jsx
@@ -1,0 +1,169 @@
+import React, { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import { useDispatch, useSelector } from 'react-redux';
+import Select from 'react-select';
+
+import Modal from '../common/modal.jsx';
+import selectStyles from '../../styles/single_select.js';
+import API from '../../utils/api.js';
+import { fetchCoursesForUser } from '../../actions/user_courses_actions';
+import { copyAvailableArticles } from '../../actions/assignment_actions';
+
+const PREVIEW_DEBOUNCE_MS = 400;
+
+// The backend reports why a source can't be used (unknown, private, or this
+// same course) in the JSON `message`; fall back to the generic error text.
+const errorMessageFrom = (error) => {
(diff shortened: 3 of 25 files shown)


ragesoss:
Review notes from a Claude Code session (attribution below). As with any AI-driven review, this is meant to be read and weighed, not automatically trusted or acted on.

**Regression, now fixed in 2b912eeaa.** `spec/features/course_cloning_spec.rb:113` failed on this branch and passed on master, and was the only failure in both CI runs. The spec's user-less fixture uses the bare `:assignment` factory, which sets no `role`, so the row had `role: nil`. The old `CourseCloneManager#copy_assignments` copied every `user_id: nil` row regardless of role; `CopyAvailableArticles` copies only `role: 0`, so the clone ended up with zero assignments. Against a prod dump, none of the 47,816 user-less assignments has a null role, so this was a fixture artifact rather than a production-visible change. The fixture now sets `role: 0`.

**Also fixed in 2b912eeaa:** the source-course react-select was rendering react-select's default English "Select..." placeholder; other selects in the Dashboard pass an I18n `placeholder`. Added `assignments.copy_available_select_placeholder` ("Select a course"). The PR description's count of three new i18n strings is now four.

Remaining minor, non-blocking observations, left as-is:
- The course URL input has only a placeholder, no label or `aria-label`; the react-select has `inputId` but no label either. Screen readers get the dialog title only.
- The preview counts a source row as "would be added" even when its title no longer passes validation, so `create` can return fewer than previewed. Only 4 such rows exist among ~48k prod available articles.
- The copy path skips `AssignmentManager`'s discouraged-article check. The clone path always did too, so not new.
- Pre-existing, out of scope: `API.cloneCourse` builds `?campaign_slug=X?copy_assignments=…` when a campaign is present, so `copy_assignments` never reaches the server from a campaign-scoped course creator. The `== 'true'` fix does not change that path. Follow-up issue material.

Verified locally: `yarn build`; RuboCop and ESLint on changed files; the PR's Ruby specs plus `assignments_controller_spec` and `assignment_spec` (115 examples); Jest for the three touched suites; the PR feature spec plus `assigned_articles_spec` and `multiwiki_assignment_spec`; `course_cloning_spec` and `course_copying_spec` on this branch and on master, and again after the fix. Two AI-written exploratory specs pass and were not committed: a feature spec covering the own-courses dropdown (now including the placeholder text), the include-assigned toggle, the private-course and same-course errors, a fully-duplicated source, and a student's view, using prod-shaped titles; and a service spec pinning the nil-role behavior above.

*Drafted in a Claude Code session (~1 hour, 5 interactions). Sage requested the review, read the findings, chose the fixture fix over loosening the service scope, picked the placeholder wording from three options, and approved this comment. Sage did not independently rerun the specs or verify the prod-dump query; all analysis, spec-writing, and test runs were done by Claude Code. The exploratory specs are AI-generated and were run but not reviewed by Sage.*

PR #7038, opened 2026-09-03
Speed up articles courses cache updates.

gabina:
## What this PR does

Speeds up the ArticlesCourses cache phase of a course update.
I did an experiment for course [Coordinate_Me_2026_NL ACUWT](https://outreachdashboard.wmflabs.org/courses/Test/TEST_Coordinate_Me_2026_NL_ACUWT/home)  (~6,000 articles) and deleted a user. For the update that catches the user deletion, the article course cache update phase was taking **6h 44m of a 7h 25m update — 91% of the total**, measured with the `debug_updates` flag.

| phase | duration |
|---|---|
| everything up to reaggregation | 2m 40s |
| timeslices reaggregated (125) | 37m 41s |
| **ArticlesCourses caches** | **6h 44m 47s** |
| CoursesUsers caches, course cache | 3s |

See raw debug updates:
```
{
0_start: 2026-09-04 22:10:10 UTC,
1_uploads_imported: 2026-09-04 22:12:07 UTC,
2_categories_updated: 2026-09-04 22:12:07 UTC,
3_article_namespaces_updated: 2026-09-04 22:12:08 UTC,
4_timeslices_adjusted: 2026-09-04 22:12:19 UTC,
5_timeslices_reprocessed_40: 2026-09-04 22:12:19 UTC,
6_acuwt_timeslices_reprocessed_40: 2026-09-04 22:12:21 UTC,
7_timeslices_processed_40: 2026-09-04 22:12:50 UTC,
8_timeslices_reaggregated_40: 2026-09-04 22:50:31 UTC,
9_average_pageviews_updated: 2026-09-04 22:50:31 UTC,
10_articles_courses_updated: 2026-09-05 05:35:18 UTC,
11_courses_users_updated: 2026-09-05 05:35:20 UTC,
12_course_cache_updated: 2026-09-05 05:35:21 UTC
}
```

This hits ACUWT courses specifically. `ArticleCourseTimeslice.bulk_update_from_acuwt`
bumps `updated_at` for every article in a reaggregated window, changed or not, so
`articles_courses_to_update` returns nearly every article in the course. The legacy path
writes ACT rows through `save`, whose partial writes only touch genuinely changed
articles. Courses that add and remove users daily, like the Coordinate Me series,
reaggregate constantly.

Two problems:

1. `ArticlesCourses.update_required_caches_from_timeslices` looked up records by
   `article_id` with no `course_id` filter, so it recomputed the caches of *every* course
   sharing those articles. On a P&E course whose articles are widely tracked this is a
   large multiplier — measured at 2× on a small course, and unbounded in general.
2. `ArticlesCourses#update_cache_from_timeslices` cost three queries per record: an
   `Article` load forced by the `article_course_timeslices` association lambda, a load of
   every timeslice as an ActiveRecord object, and a separate `MIN(first_revision)` round
   trip.

The aggregation now lives in `ArticlesCoursesCacheManager`, mirroring the existing
`CourseCacheManager`. It aggregates in SQL and writes each batch of 1000 records with one
`upsert_all`, following the pattern already used by
`ArticleCourseTimeslice.bulk_update_from_acuwt`. The unused instance method and the
`article_course_timeslices` association are removed.

Measured locally on course `Open Data Day Taiwan 2026`: **68 queries / ~813 ms → 5
queries / ~26 ms**. Cached values are unchanged; the one difference is that `upsert_all`
bumps `updated_at` unconditionally, where partial writes previously skipped no-op updates.

## AI usage

Claude Code did the analysis and wrote most of the code. It read the cache-update path,
identified the two problems above from a `pry` log the user captured on production data,
proposed the bulk-SQL design, and drafted `lib/articles_courses_cache_manager.rb`, the
model changes, and the specs.

The user directed the work throughout: chose which problems to pursue and in what order,
made the scoping fix in `c0af3a37d` herself, decided to defer the specs until the code had
been measured on production, created the manager file, and ran every test and lint pass.
Claude Code ran no specs in these sessions. Two of Claude Code's suggestions were dropped
after the user's production logs contradicted them — a `save if changed?` optimization
(Rails' partial writes already skip unchanged records) and an estimate attributing a
6½-hour update to a phase that the user's own comparison against the non-ACUWT branch
showed was elsewhere.

## Screenshots

No UI changes.

- `BATCH_SIZE = 1000` counts ArticlesCourses records, not timeslices. `user_ids` is a
  serialized array so it cannot be aggregated in SQL and is merged in Ruby, which means
  the pluck behind it returns one row per timeslice. On a course with heavy timeslice
  splitting, one batch could pull far more rows than intended; the constant may want
  lowering. Not yet measured on such a course.
- `upsert_all` skips validations and callbacks. `ArticlesCourses` has none on these
  columns, and `bulk_update_from_acuwt` already makes the same tradeoff on the same kind
  of data, but it is worth a reviewer's eye.
- `updated_at` is now written on every pass. Nothing in `app` or `lib` reads that column.
  Dropping it from `UPDATED_FIELDS` would restore the old "skip unchanged rows" behavior
  at the cost of the timestamp going stale.
- Not addressed here: `UpdateCourseStats#update_caches` wraps all three cache phases in a
  single `ActiveRecord::Base.transaction`, so a large course holds row locks and grows the
  undo log for the whole phase. Worth a follow-up.

### DIFF
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -87,7 +87,7 @@ class ArticlesCourses < ApplicationRecord
   end
 
   def self.update_required_caches_from_timeslices(course)
-    ArticlesCourses.where(article_id: articles_courses_to_update(course))
+    ArticlesCourses.where(course:, article_id: articles_courses_to_update(course))
                    .find_each(&:update_cache_from_timeslices)
   end
 
--- a/app/models/course_data/articles_courses.rb
+++ b/app/models/course_data/articles_courses.rb
@@ -21,6 +21,7 @@
 
 require_dependency "#{Rails.root}/lib/timeslice_manager"
 require_dependency "#{Rails.root}/lib/cumulative_diff_url_builder"
+require_dependency "#{Rails.root}/lib/articles_courses_cache_manager"
 
 #= ArticlesCourses is a join model between Article and Course.
 #= It represents a mainspace Wikipedia article that has been worked on by a
@@ -28,10 +29,6 @@ require_dependency "#{Rails.root}/lib/cumulative_diff_url_builder"
 class ArticlesCourses < ApplicationRecord
   belongs_to :article
   belongs_to :course
-
-  has_many :article_course_timeslices, lambda { |articles_courses|
-                                         where article: articles_courses.article
-                                       }, through: :course
 
   scope :live, -> { joins(:article).where(articles: { deleted: false }).distinct }
   scope :new_article, -> { where(new_article: true) }
@@ -48,15 +45,6 @@ class ArticlesCourses < ApplicationRecord
   ####################
   def cumulative_diff_url
     CumulativeDiffUrlBuilder.new(self).url
-  end
-
-  def update_cache_from_timeslices
-    self.character_sum = article_course_timeslices.sum(&:character_sum)
-    self.references_count = article_course_timeslices.sum(&:references_count)
-    self.user_ids = article_course_timeslices.sum([], &:user_ids).uniq
-    self.new_article = article_course_timeslices.any?(&:new_article)
-    self.first_revision = article_course_timeslices.minimum(:first_revision)
-    save
   end
 
   #################
@@ -87,12 +75,13 @@ class ArticlesCourses < ApplicationRecord
   end
 
   def self.update_required_caches_from_timeslices(course)
-    ArticlesCourses.where(course:, article_id: articles_courses_to_update(course))
-                   .find_each(&:update_cache_from_timeslices)
+    update_all_caches_from_timeslices(
+      where(course:, article_id: articles_courses_to_update(course))
+    )
   end
 
   def self.update_all_caches_from_timeslices(articles_courses)
-    articles_courses.find_each(&:update_cache_from_timeslices)
+    ArticlesCoursesCacheManager.new(articles_courses).update_caches_from_timeslices
   end
 
   # Creates missing ArticlesCourses records for the given course revisions whose
--- a/lib/articles_courses_cache_manager.rb
+++ b/lib/articles_courses_cache_manager.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+#= Service for updating the counts that are cached on ArticlesCourses objects
+class ArticlesCoursesCacheManager
+  BATCH_SIZE = 1000
+  UPDATED_FIELDS = %i[character_sum references_count user_ids new_article
+                      first_revision updated_at].freeze
+  # Values for an articles_courses record whose timeslices are all gone.
+  EMPTY_CACHE = { character_sum: 0, references_count: 0, user_ids: [],
+                  new_article: false, first_revision: nil }.freeze
+
+  def initialize(articles_courses)
+    @articles_courses = articles_courses
+  end
+
+  # Recalculates the cached fields from the article course timeslices, aggregating
+  # them in SQL and writing every record of a batch with a single upsert.
+  def update_caches_from_timeslices
+    @articles_courses.pluck(:id, :course_id, :article_id)
+                     .each_slice(BATCH_SIZE) { |batch| write_caches(batch) }
+  end
+
+  private
+
+  # Takes [id, course_id, article_id] triples and writes their recalculated caches.
+  def write_caches(records)
+    stats = timeslice_stats(records)
+    now = Time.zone.now
+    rows = records.map do |id, course_id, article_id|
+      { id:, course_id:, article_id:, updated_at: now,
+        **(stats[[course_id, article_id]] || EMPTY_CACHE) }
+    end
+    ArticlesCourses.upsert_all(rows, update_only: UPDATED_FIELDS)
+  end
+
+  # Returns { [course_id, article_id] => cache values }. Timeslices are queried per
+  # course so that the article_course_timeslices unique index applies.
+  def timeslice_stats(records)
+    stats = {}
+    records.group_by { |_id, course_id, _article_id| course_id }
+           .each do |course_id, triples|
+      stats.merge!(stats_for_course(course_id, triples.map(&:last)))
+    end
+    stats
+  end
+
+  def stats_for_course(course_id, article_ids)
+    timeslices = ArticleCourseTimeslice.where(course_id:, article_id: article_ids)
+    stats = aggregated_stats(timeslices)
+    user_ids_by_article(timeslices).each { |id, user_ids| stats[id][:user_ids] = user_ids }
+    stats.transform_keys { |article_id| [course_id, article_id] }
+  end
+
+  # SQL aggregates for every cached field but user_ids, keyed by article id.
+  def aggregated_stats(timeslices)
+    timeslices.group(:article_id)
+              .pluck(:article_id, 'SUM(character_sum)', 'SUM(references_count)',
+                     'MAX(new_article)', 'MIN(first_revision)')
+              .to_h do |article_id, chars, refs, new_article, first_revision|
+                [article_id,
+                 { character_sum: chars.to_i, references_count: refs.to_i,
+                   new_article: new_article.to_i.positive?, first_revision:,
+                   user_ids: [] }]
+              end
+  end
+
+  # user_ids is a serialized array, so it cannot be aggregated in SQL.
+  def user_ids_by_article(timeslices)
+    timeslices.pluck(:article_id, :user_ids)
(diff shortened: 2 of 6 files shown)


gabina:
Failures don't seem related to these changes now:

```
  1) DeUserfyingEditAlertMonitor.edits checks keys
     Failure/Error: expect(fedit.dig('logparams', 'target_title')).not_to be_nil

     NoMethodError:
       undefined method 'dig' for nil
     # ./spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb:49:in 'block (4 levels) in <top (required)>'
     # ./spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb:47:in 'block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:78:in 'block (2 levels) in <top (required)>'

  2) UploadImporter.import_urls_in_batches finds and saves Commons thumbnail urls
     Failure/Error: expect(peas_photo.thumburl[0...24]).to eq('https://upload.wikimedia/')

       expected: "https://upload.wikimedia/"
            got: "https://thumb.wikimedia./"

       (compared using ==)
     # ./spec/lib/importers/upload_importer_spec.rb:90:in 'block (4 levels) in <top (required)>'
     # ./spec/lib/importers/upload_importer_spec.rb:87:in 'block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:78:in 'block (2 levels) in <top (required)>'

Finished in 14 minutes 54 seconds (files took 4.2 seconds to load)
1284 examples, 2 failures, 6 pending
```

gabina:
Those specs are also failing locally for me, I'll fix them in a different PR.

ragesoss:
This seems like a safe change to me. It would be good to measure the effect of this change on a real ACUWT course like the one you experimented on. I'm guessing we shouldn't expect a direct extrapolation of the small local measurement you did (which took about 3% as long after the change, so the slow step would go from 6:44 to about 0:13), but if it's anything close, that'll make a big difference for our ACUWT capacity.

gabina:
Yes, this is what I know so far:

- Removing user Andre Engels from ACUWT version (course [37575](https://outreachdashboard.wmflabs.org/courses/Test/TEST_Coordinate_Me_2026_NL_ACUWT/home)) took ~7 hours, with the article course cache update step accounting for almost all of that time.
- Removing user Andre Engels from default version (course [37577](https://outreachdashboard.wmflabs.org/courses/Test/TEST_Coordinate_Me_2026_NL_DEFAULT_(August_2028)/home) took ~12 hours (but article course cache update step was fast). It took a lot of time because it reprocessed a lot of timeslices.

So, if we manage to reduce the aritcle course cache step time for ACUWT version by  at least one-third or something like that, I think we could start thinking about having 30 of those courses running at the same time without causing weeks of latency in the long queue (although it would still be a matter of days). And if the improvement is significantly better, we should be just fine.

I'll be testing that when we deploy this. Right now I'm re-adding the user to the ACUWT course, so a long update is expected for it.

I also would like to make an experiment with a huge category update (a lot of new articles), although I didn't decide yet how to reproduce that.

ragesoss (approved):
AI-assisted review feedback is provisional and should be independently evaluated rather than automatically trusted.\n\nThe implementation was reviewed for cache-equivalence, course scoping, SQL aggregation, batching, removed association callers, and test coverage. No blocking correctness issues were identified.\n\nVerified locally:\n\n- 75 affected Ruby specs passed\n- RuboCop passed on all six changed Ruby files\n- Started cleaning stale assets
[33mFinished cleaning stale assets after 0.017 seconds[0m
Started emitting i18n translations passed\n-  passed\n\nThe recorded CI failure is unrelated to this PR; both failing specs were fixed in merged PR #7046. The remaining operational risk is that timeslice-heavy courses may make a nominal 1,000-record batch fetch substantially more rows, as documented in the PR description.\n\nApproved, contingent on CI being rerun successfully.\n\n*Drafted in a Codex session (~35 minutes, ~3 user interactions). Sage directed the triage and full review and requested this approval, reviewed the summaries, and approved posting. Codex inspected the full PR diff, ran the affected specs, RuboCop, asset build, and whitespace checks locally; no exploratory AI-generated code was used.*

ragesoss (approved):
AI-assisted review feedback is provisional and should be independently evaluated rather than automatically trusted.\n\nThe implementation was reviewed for cache-equivalence, course scoping, SQL aggregation, batching, removed association callers, and test coverage. No blocking correctness issues were identified.\n\nVerified locally:\n\n- 75 affected Ruby specs passed\n- RuboCop passed on all six changed Ruby files\n- yarn build passed\n- git diff --check passed\n\nThe recorded CI failure is unrelated to this PR; both failing specs were fixed in merged PR #7046. The remaining operational risk is that timeslice-heavy courses may make a nominal 1,000-record batch fetch substantially more rows, as documented in the PR description.\n\nApproved, contingent on CI being rerun successfully.\n\n*Drafted in a Codex session (~35 minutes, ~3 user interactions). Sage directed the triage and full review and requested this approval, reviewed the summaries, and approved posting. Codex inspected the full PR diff, ran the affected specs, RuboCop, asset build, and whitespace checks locally; no exploratory AI-generated code was used.*

ragesoss (commented):
A note on how to read this: the findings below came out of an AI-assisted review session. They are meant to be read and weighed, not automatically trusted or acted on.

The change looks sound. Verified locally on the branch merged with current master: `yarn build`, RuboCop on the changed files, the PR's specs plus the existing specs for `ArticlesCourses`, `UpdateCourseStats`, `CourseCacheManager`, `ArticlesCoursesCleaner` and the course page feature spec (95 + 33 examples, all green), and an exploratory spec covering non-cache column preservation, query count, and the `updated_at` behavior.

Two observations, neither blocking:

**`updated_at` does have a reader.** `app/views/campaigns/articles.html.haml:78` uses `ac.updated_at` in the fragment-cache key for each campaign article row, so the unconditional bump invalidates every row's fragment on each course update rather than only changed rows. There appears to be a two-line fix: drop `updated_at` from `UPDATED_FIELDS` and from the row hash in `write_caches`. Rails' `upsert_all` then emits the touch itself as `updated_at = CASE WHEN (<every listed column> <=> VALUES(...)) THEN updated_at ELSE CURRENT_TIMESTAMP(6) END`, which restores the old "skip unchanged rows" timestamp semantics. Checked locally: the PR's specs still pass with that change, an unchanged row keeps its timestamp, and a changed row is bumped.

**Deleted-row resurrection window.** If an `articles_courses` row is deleted between the initial pluck and the batch upsert (for example a course deleted mid-update, since `Course` destroys `articles_courses` via `dependent: :destroy`), the upsert re-inserts it with its old id where the previous `save` path would have updated zero rows. The window is seconds and `bulk_update_from_acuwt` already accepts the same tradeoff, so this is noted for awareness only.

The CI failure is the two specs fixed in #7046; merging master should clear it.

*Drafted in a Claude Code session (~50 min, 2 interactions). Claude Code ran the build and specs listed above and wrote the exploratory spec; Sage read the full comment before posting but did not independently rerun the checks or the exploratory spec.*

gabina:
> Right now I'm re-adding the user to the ACUWT course, so a long update is expected for it.

Just in case, that update already finished so we're good to deploy whenever you want.

gabina:
Deleting user Andre Engels from ACUWT version. It seems much better than we expected
```
{
  0_start: 2026-09-09 20:05:10 UTC,
  1_uploads_imported: 2026-09-09 20:07:12 UTC,
  2_categories_updated: 2026-09-09 20:07:12 UTC,
  3_article_namespaces_updated: 2026-09-09 20:07:13 UTC,
  4_timeslices_adjusted: 2026-09-09 20:07:24 UTC,
  5_timeslices_reprocessed_40: 2026-09-09 20:07:24 UTC,
  6_acuwt_timeslices_reprocessed_40: 2026-09-09 20:07:26 UTC,
  7_timeslices_processed_40: 2026-09-09 20:07:30 UTC,
  8_timeslices_reaggregated_40: 2026-09-09 20:46:30 UTC,
  9_average_pageviews_updated: 2026-09-09 20:46:30 UTC,
  10_articles_courses_updated: 2026-09-09 20:46:43 UTC,
  11_courses_users_updated: 2026-09-09 20:46:45 UTC,
  12_course_cache_updated: 2026-09-09 20:46:46 UTC
}
```

gabina:
> which took about 3% as long after the change, so the slow step would go from 6:44 to about 0:13

exactly that BTW  **Edited**: even much less, because it took 13 seconds, not 13 minutes. Very surprising

PR #7043, opened 2026-09-04
Guard WikiApi#retry_after_seconds against responses without headers

### LINKED ISSUE #7033: WikiApi#retry_after_seconds raises NoMethodError on MediawikiApi::ApiError, so errors escape instead of returning nil
## Summary

`WikiApi#retry_after_seconds` raises `NoMethodError` when the error it is handed is a `MediawikiApi::ApiError`, so the exception escapes `WikiApi#mediawiki`'s rescue block instead of being logged and turned into a `nil` return. Any caller that relies on the documented "returns nil once retries are exhausted" contract will instead see an unhandled exception.

## Mechanism

`lib/wiki_api.rb:122` catches `StandardError`, retries three times, and then builds Sentry context:

```ruby
retry unless tries.zero?
extra = rate_limit_sentry_extra(e).merge(action:, query: args, api_url: @api_url)   # :134
```

which reaches `lib/wiki_api.rb:160`:

```ruby
def retry_after_seconds(error)
  return nil unless error.respond_to?(:response) && error.response
  raw = error.response.headers['Retry-After']    # :162
  Integer(raw) if raw.present?
rescue ArgumentError, TypeError
  nil
end
```

`MediawikiApi::ApiError#response` returns a `MediawikiApi::Response`, which `def_delegators` only `:status` and `:success?` — it has no `headers`. So line 162 raises:

```
NoMethodError: undefined method 'headers' for an instance of MediawikiApi::Response
```

and the `rescue ArgumentError, TypeError` on line 164 does not catch it. The `NoMethodError` propagates out of `WikiApi#query` / `WikiApi#action` to the caller.

`MediawikiApi::HttpError` is unaffected: its `@response` is a `Faraday::Response`, which does respond to `headers`. The bug is specific to `ApiError`.

## When this fires

`MediawikiApi::Client#send_request` raises `ApiError` when the response carries a `mediawiki-api-error` header **and** the HTTP status is below 400 — the `raise HttpError` check comes first. The MediaWiki action API returns plenty of errors that way. Verified against live en.wikipedia.org:

```
$ curl -sD - "https://en.wikipedia.org/w/api.php?action=query&list=usercontribs&ucuser=Bad%23Name&ucnamespace=0&format=json"
HTTP/2 200
mediawiki-api-error: baduser
```

So a malformed parameter value — an illegal username in `ucuser`, a bad title, too many values for a multi-value parameter — produces HTTP 200 plus the header, hence `ApiError`, hence the `NoMethodError`.

Callers that pass an interception block (`WikiApi#query`'s `&caller_handles`, added for exactly this purpose) are fine, because `raise if caller_handles&.call(e)` on line 129 fires before the Sentry path. Only the default no-block path is affected.

## Impact

Latent rather than active. A search of Sentry over the last 90 days found no occurrence of this `NoMethodError`, which is consistent with current callers mostly issuing well-formed queries over titles and revision IDs that the API accepts. But `retry_after_seconds` was added in e62075409 (2026-04-29) and is deployed on both `production` and `wmflabs`, so any new caller that puts user-supplied values into a multi-value API parameter can hit it.

That is how it surfaced: #7009 adds a worker that queries `list=usercontribs` with up to 40 dashboard usernames in `ucuser`, and relies on a `baduser` response coming back as `nil` so it can fall back to per-user queries. It instead crashes the Sidekiq job. See https://github.com/WikiEducationFoundation/WikiEduDashboard/pull/7009#pullrequestreview-5105017821 (finding 1) for the full trace and a reproduction; that PR should not have to work around this.

## Reproduction

```ruby
# frozen_string_literal: true

require 'rails_helper'
require_dependency "#{Rails.root}/lib/wiki_api"

describe WikiApi do

  # What MediawikiApi::Client#send_request builds for a baduser response:
  # HTTP 200 plus a mediawiki-api-error header.
  def baduser_error
    body = { 'error' => { 'code' => 'baduser', 'info' => 'Invalid value for user parameter.' } }
    faraday = Faraday::Response.new(status: 200, body: body.to_json,
                                    response_headers: { 'mediawiki-api-error' => 'baduser' })
    MediawikiApi::ApiError.new(MediawikiApi::Response.new(faraday, ['error']))
  end

  it 'returns nil instead of raising when the API reports an error' do
    allow_any_instance_of(MediawikiApi::Client).to receive(:query) { raise baduser_error }

    expect(described_class.new(wiki).query(list: 'usercontribs', ucuser: %w[A B])).to be_nil
  end

  it 'still parses Retry-After from a real HttpError' do
    faraday = Faraday::Response.new(status: 429, body: '{}',
                                    response_headers: { 'Retry-After' => '7' })
    api = described_class.new(wiki)
    expect(api.send(:retry_after_seconds, MediawikiApi::HttpError.new(faraday))).to eq(7)
  end
end
```

On current master the first example fails with `NoMethodError: undefined method 'headers' for an instance of MediawikiApi::Response`; the second passes.

## Suggested fix

Make the header lookup defensive, so only errors that actually carry an HTTP response are inspected. Something like:

```ruby
def retry_after_seconds(error)
  response = error.try(:response)
  return nil unless response.respond_to?(:headers)
  raw = response.headers['Retry-After']
  Integer(raw) if raw.present?
rescue ArgumentError, TypeError
  nil
end
```

The existing comment on the method already anticipates "the gem version in use doesn't expose the response on `HttpError`" — this is the same class of problem for a different error type.

That patch was applied locally and both examples above pass with it, so the `HttpError` path keeps working. `Retry-After` is only meaningful on a 429/503 anyway, which is always `HttpError`, so returning nil for `ApiError` loses nothing.

Both branches are worth keeping as specs, since the current suite covers neither.

(Issue description written by Claude Code.)

lakshita10341:
## What this PR does
Fixes #7033 

`WikiApi#retry_after_seconds` raises `NoMethodError` when handed a `MediawikiApi::ApiError`, causing the exception to escape `WikiApi#mediawiki`'s rescue block instead of logging the error and returning `nil`.

### Mechanism & Root Cause
1. `WikiApi#mediawiki` catches `StandardError`, retries up to 3 times, and then builds Sentry context using `rate_limit_sentry_extra(e)`.
2. `rate_limit_sentry_extra(e)` calls `retry_after_seconds(e)` to extract any HTTP `Retry-After` header.
3. `MediawikiApi::ApiError` (raised on HTTP 200 responses carrying MediaWiki API error payloads like `baduser` or `badtitle`) wraps a `MediawikiApi::Response` object. Unlike `MediawikiApi::HttpError` (which wraps `Faraday::Response`), `MediawikiApi::Response` delegates only `:status` and `:success?` — it does **not** implement or delegate `:headers`.
4. Calling `error.response.headers` on an `ApiError` throws `NoMethodError: undefined method 'headers' for #<MediawikiApi::Response:...>`, causing an unhandled exception for callers that expect `WikiApi#query` to return `nil` on failure.

### Changes Made
- Added a `respond_to?(:headers)` guard in `WikiApi#retry_after_seconds` (`lib/wiki_api.rb`).
- Added unit specs in `spec/lib/wiki_api_spec.rb` to verify `MediawikiApi::ApiError` returns `nil` without raising, and `MediawikiApi::HttpError` correctly parses `Retry-After`.

## AI usage
Used Google Antigravity AI agent to analyze the `MediawikiApi::Response` method delegations and write unit test case.

## Screenshots
N/A (backend error handling fix in `lib/wiki_api.rb`, no UI changes)

### DIFF
--- a/lib/wiki_api.rb
+++ b/lib/wiki_api.rb
@@ -154,12 +154,14 @@ class WikiApi
   end
 
   # Returns the integer seconds requested by the server's Retry-After header,
-  # or nil if the header is absent / unparseable / the gem version in use
-  # doesn't expose the response on HttpError. Wikimedia uses delay-seconds;
-  # HTTP-date form (RFC 7231) is not parsed.
+  # or nil if the header is absent / unparseable / the error's response object
+  # doesn't expose headers (e.g. MediawikiApi::ApiError wraps a Response that
+  # delegates only :status and :success?, not :headers).
+  # Wikimedia uses delay-seconds; HTTP-date form (RFC 7231) is not parsed.
   def retry_after_seconds(error)
-    return nil unless error.respond_to?(:response) && error.response
-    raw = error.response.headers['Retry-After']
+    response = error.try(:response)
+    return nil unless response.respond_to?(:headers)
+    raw = response.headers['Retry-After']
     Integer(raw) if raw.present?
   rescue ArgumentError, TypeError
     nil
--- a/spec/lib/wiki_api_spec.rb
+++ b/spec/lib/wiki_api_spec.rb
@@ -384,4 +384,45 @@ describe WikiApi do
       end
     end
   end
+
+  describe '#query with a MediawikiApi::ApiError (no interception block)' do
+    before { stub_wiki_validation }
+
+    # MediawikiApi::Client#send_request raises ApiError for HTTP 200 responses
+    # that carry a mediawiki-api-error header (e.g. baduser, badtitle).
+    # ApiError#response is a MediawikiApi::Response, which does NOT delegate
+    # :headers — so retry_after_seconds must not call .headers on it.
+    def baduser_error
+      body = { 'error' => { 'code' => 'baduser',
+                             'info' => 'Invalid value for user parameter.' } }
+      faraday = Faraday::Response.new(
+        status: 200, body: body.to_json,
+        response_headers: { 'mediawiki-api-error' => 'baduser' }
+      )
+      MediawikiApi::ApiError.new(MediawikiApi::Response.new(faraday, ['error']))
+    end
+
+    it 'returns nil instead of raising NoMethodError' do
+      allow_any_instance_of(MediawikiApi::Client)
+        .to receive(:query) { raise baduser_error }
+      wiki = Wiki.get_or_create(language: 'en', project: 'wikipedia')
+      expect(described_class.new(wiki).query(list: 'usercontribs',
+                                             ucuser: %w[A B])).to be_nil
+    end
+  end
+
+  describe '#retry_after_seconds' do
+    before { stub_wiki_validation }
+
+    it 'parses Retry-After from a real HttpError' do
+      faraday = Faraday::Response.new(
+        status: 429, body: '{}',
+        response_headers: { 'Retry-After' => '7' }
+      )
+      wiki = Wiki.get_or_create(language: 'en', project: 'wikipedia')
+      api = described_class.new(wiki)
+      expect(api.send(:retry_after_seconds,
+                      MediawikiApi::HttpError.new(faraday))).to eq(7)
+    end
+  end
 end


ragesoss (approved):
Reviewed with Claude Code. AI-generated review feedback is provisional: worth
reading and understanding, but not worth automatically trusting or acting on.
Use or ignore each point based on your own judgment.

## Verification

The change was checked out locally and checked against the pinned `mediawiki_api`
fork (`WikiEducationFoundation/mediawiki-ruby-api@3eebf48`, branch
`expose-response-on-http-error`):

- `MediawikiApi::Response` declares `def_delegators :@response, :status, :success?`
  and defines no `method_missing`, so `respond_to?(:headers)` is false and
  `.headers` raises. The mechanism in the PR description holds.
- `MediawikiApi::HttpError` in that fork exposes `attr_reader :response, :status`
  and accepts either a Faraday response or a bare status integer.
  `error.try(:response)` handles both forms; the integer form yields `nil`
  rather than raising.
- Running the two new examples against master's `lib/wiki_api.rb` fails the
  `ApiError` example with the expected `NoMethodError: undefined method 'headers'
  for an instance of MediawikiApi::Response`, via `retry_after_seconds` →
  `rate_limit_sentry_extra` → `mediawiki` → `query`, while the `HttpError`
  example passes. Both examples are therefore doing real work: one guards the
  fix, the other guards against the fix over-narrowing.
- `spec/lib/wiki_api_spec.rb` — 33 examples, 0 failures.
  `spec/services/update_wikidata_stats_timeslice_spec.rb` and `spec/lib/errors` —
  12 examples, 0 failures. RuboCop clean on both changed files.

One effect the description does not claim: the `respond_to?(:headers)` guard also
closes a second latent `NoMethodError`. `Faraday::Error#response` can be a plain
`Hash` (`{status:, headers:, body:}`), which likewise does not respond to
`headers`, so the old `error.response.headers` would have raised there too. That
path does not appear reachable through `WikiApi` today — the gem's Faraday stack
installs no `raise_error` middleware — so this reads as hardening rather than a
second live fix.

## Observations

**The untested half of the contract.** The `ApiError` example asserts only that
`query` returns `nil`. The behavior #7033 describes is "logging the error and
returning `nil`", and the logging half is unexercised: a regression that swallowed
the error entirely, or that broke the Sentry context build in some other way,
would still pass. Specs covering that gap were written during this review and
pass on the branch (reusing this PR's own `baduser_error` helper):

```ruby
it 'reports the error to Sentry without a retry_after extra' do
  allow(Sentry).to receive(:capture_exception)
  allow_any_instance_of(MediawikiApi::Client).to receive(:query) { raise baduser_error }
  described_class.new(wiki).query(list: 'usercontribs', ucuser: %w[A B])
  expect(Sentry).to have_received(:capture_exception).with(
    an_instance_of(MediawikiApi::ApiError),
    hash_excluding(extra: hash_including(:retry_after))
  )
end

it 'ticks the update service error stats' do
  update_service = double('UpdateService', sentry_tags: {}, update_error_stats: nil)
  allow_any_instance_of(MediawikiApi::Client).to receive(:query) { raise baduser_error }
  expect(update_service).to receive(:update_error_stats).once
  described_class.new(wiki, update_service).query(list: 'usercontribs', ucuser: %w[A B])
end

it 'does not sleep, since an ApiError is not a 429' do
  allow_any_instance_of(MediawikiApi::Client).to receive(:query) { raise baduser_error }
  expect_any_instance_of(described_class).not_to receive(:sleep)
  described_class.new(wiki).query(list: 'usercontribs', ucuser: %w[A B])
end
```

Also passing, and possibly worth folding in: `retry_after_seconds` returns `nil`
for `Faraday::TimeoutError`, for `MediawikiApi::ApiError.new` with no response,
and for `MediawikiApi::HttpError.new(429)` in the bare-integer form. That last
one is the back-compat constructor path in the fork, and nothing currently
covers it.

**The dropped comment clause.** The previous comment recorded that the header
lookup can return `nil` because "the gem version in use doesn't expose the
response on `HttpError`". That case is still live: the `Gemfile` pins the fork
specifically to get `HttpError#response`, with a note to switch back to plain
`mediawiki_api` once the upstream merge is released. The new wording covers it
under "e.g.", but the fork dependency is no longer recorded at the point where it
matters. Restoring a clause would keep that link visible to whoever eventually
unpins the fork.

Neither point blocks merging.

*Drafted in a Claude Code session (~15 min, 2 interactions). Sage asked for the
review and approved this comment after reading the session's summary; the
findings above were produced by Claude Code and were not independently reproduced
by Sage. The verifications named above — reading the fork's source, running the
PR's specs against master's `lib/wiki_api.rb`, and the spec and RuboCop runs —
were carried out in-session by Claude Code. The specs quoted above are
AI-generated; they were run on this branch and pass, but Sage has not reviewed
them line by line.*

PR #7046, opened 2026-09-08
Fix failing specs in CI

gabina:
## What this PR does
This PR fixes the two current failing specs in CI:

```
  1) DeUserfyingEditAlertMonitor.edits checks keys
     Failure/Error: expect(fedit.dig('logparams', 'target_title')).not_to be_nil

     NoMethodError:
       undefined method 'dig' for nil
     # ./spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb:49:in 'block (4 levels) in <top (required)>'
     # ./spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb:47:in 'block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:78:in 'block (2 levels) in <top (required)>'

  2) UploadImporter.import_urls_in_batches finds and saves Commons thumbnail urls
     Failure/Error: expect(peas_photo.thumburl[0...24]).to eq('https://upload.wikimedia/')

       expected: "https://upload.wikimedia/"
            got: "https://thumb.wikimedia./"

       (compared using ==)
     # ./spec/lib/importers/upload_importer_spec.rb:90:in 'block (4 levels) in <top (required)>'
     # ./spec/lib/importers/upload_importer_spec.rb:87:in 'block (3 levels) in <top (required)>'
     # ./spec/rails_helper.rb:78:in 'block (2 levels) in <top (required)>'
```

`DeUserfyingEditAlertMonitor` spec fails because the `recentchanges` response (de-userfying for the last 30 days) returns empty right now. The proposed solution is to avoid doing real requests (stop using VCR cassettes) and mock the response for both cases: no recent changes, existing recent changes.

`UploadImporter` spec fails because the `thumburl` field now (inside `imageinfo`) has a thumb (not upload) url domain. The proposed solution is just to update the expected URL in the specs. I don't think the domain would change soon, so this should be enough.

## AI usage
Discussed with Claude Code.

### DIFF
--- a/spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb
+++ b/spec/lib/alerts/de_userfying_edit_alert_monitor_spec.rb
@@ -43,15 +43,40 @@ describe DeUserfyingEditAlertMonitor do
   end
 
   describe '.edits' do
-    it 'checks keys' do
-      VCR.use_cassette 'recent_changes' do
-        fedit = mntor.edits.first
-        expect(fedit.dig('logparams', 'target_title')).not_to be_nil
-        expect(fedit.dig('user')).not_to be_nil
-        expect(fedit.dig('revid')).not_to be_nil
-        expect(fedit.dig('pageid')).not_to be_nil
-        expect(fedit.dig('logid')).not_to be_nil
-        expect(fedit.dig('timestamp')).not_to be_nil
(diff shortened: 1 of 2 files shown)


ragesoss:
LGTM

PR #7048, opened 2026-09-09
Added 'Get Help' button to training modules

GonzaFP:
## What this PR does
This PR fixes issue  #5486 i.e it adds a Get Help Button to training when a user is logged in. To accomplish this, the following was done:
1. Breadcrumbs rendering was moved from rails(slide_view) to react. i.e a BreadCrumbs component was created.
2. Training Hamburger component was created to house the SlideMenu component, the hamburger and the slide page label.
3. TrainingNav component was created to house Breadcrumbs, Training Hamburger and GetHelp Button. Training Nav was then included in Training Slide Handler Component.

In order for the  GetHelp Button to receive course data, the following were done for a user who is enrolled in atleast one course:
1. A get_recent_course function was created in users.rb to fetch a user's recent course.
2. This recent course and course_slug were accessed in training_controller. recent course was also used to get current user roles.
3. This data was embedded in html in the slide_view via #react_root.
4. Data was accessed in react and then used to update course state and current user state.
This way, the Get Help Button appears whether a user accessed a training module from course page or directly from training modules page.

For a logged in user who has not enrolled in any course and navigated from training, the GetHelp button appears with generic info since the recent course is set to null. If the unenrolled user navigates from course page, the GetHelp button appears with more information.

## Screenshots
Before:
image

After:
image

After (user who has not enrolled in a course and navigated from training)
image

After (user who has not enrolled in a course and navigated from courses)
image

### DIFF
--- a/app/assets/javascripts/training/components/bread_crumbs.jsx
+++ b/app/assets/javascripts/training/components/bread_crumbs.jsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { useParams } from "react-router-dom";
+
+const BreadCrumbs = () => {
+    const params = useParams();
+    const moduleName = document.getElementById('react_root').getAttribute('data-module-name');
+    console.log(params)
+    return (
+        <>
+            <ol className='breadcrumbs'>
+                <li><a href="/training">{I18n.t("training.training_library")}</a></li>
+                <li>
+                    <a href={`/training/${params.library_id}/${params.module_id}`}>
+                    {moduleName}
+                    </a>
+                    </li>
+            </ol>
+        </>
+    );
+}
+
+export default BreadCrumbs;
--- a/app/assets/javascripts/training/components/training_hamburger.jsx
+++ b/app/assets/javascripts/training/components/training_hamburger.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+import SlideMenu from './slide_menu.jsx';
+import { useDispatch, useSelector } from 'react-redux';
+import {toggleMenuOpen } from '../../actions/training_actions.js';
+import { useParams } from "react-router-dom";
+
+const TrainingHamburger = () => {
+    const dispatch = useDispatch();
+    const routeParams = useParams()
+    const training = useSelector(state => state.training);
+    const menuClass = training.menuIsOpen === false ? 'hidden' : 'shown';
+     const closeMenu_FC = (e) => {
+        if (training.menuIsOpen) {
+        e.stopPropagation();
+        dispatch(toggleMenuOpen({ currently: true }));
+        }
+    };
+
+    const toggleMenuOpen_FC = (e) => {
+        e.stopPropagation();
+        dispatch(toggleMenuOpen({ currently: training.menuIsOpen }));
+    };
+
+    return (
+        <div>
+        <div
+          role="button"
+          tabIndex={0}
+          className="pull-right training__slide__nav"
+          onClick={toggleMenuOpen_FC}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter' || e.key === ' ') {
+              e.preventDefault();
+              toggleMenuOpen_FC();
+            }
+          }}
+        >
+          <div className="pull-right hamburger">
+            <span className="hamburger__bar" />
+            <span className="hamburger__bar" />
+            <span className="hamburger__bar" />
+          </div>
+          <span className="pull-right training__slide__page-label">
+            {I18n.t('training.page_number', { number: training.currentSlide.index, total: training.slides.length })}
+          </span>
+        </div>
+        <SlideMenu
+          closeMenu={closeMenu_FC}
+          onClick={toggleMenuOpen_FC}
+          menuClass={menuClass}
+          currentSlide={training.currentSlide}
+          params={routeParams}
+          enabledSlides={training.enabledSlides}
+          slides={training.slides}
+        />
+        </div>
+    );
+}
+
+export default TrainingHamburger;
--- a/app/assets/javascripts/training/components/training_nav.jsx
+++ b/app/assets/javascripts/training/components/training_nav.jsx
@@ -1,0 +1,61 @@
+import React,{ useEffect } from 'react';
+import TrainingHamburger from "./training_hamburger.jsx";
+import BreadCrumbs from "./bread_crumbs.jsx";
+import GetHelpButton from '../../components/common/get_help_button.jsx';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { fetchCourse} from '../../actions/course_actions';
+import { fetchUsers } from '../../actions/user_actions.js';
+import {getCurrentUser} from '../../selectors';
+
+const TrainingNav = (props)=>{
+  useEffect(() => {
+    const courseSlug = getCourseSlug();
+    if (courseSlug){
+      props.fetchCourse(courseSlug);
+      props.fetchUsers(courseSlug);
+    }
+  }, []);
+  const getCourseSlug = () => {
+    const courseSlug = document.getElementById("react_root").getAttribute("data-course-slug");
+    if (courseSlug){
+       return courseSlug;
+    }
+  };
+  const usersignedin = document.getElementById("react_root").getAttribute("data-usersignedin")  === "true";
+ return (
+    <header className="training__slide-header">
+      <div className="training__nav-container">
+        <div className="training__breadcrumbs nav__item">
+          <BreadCrumbs/>
+        </div>
+        <div className="training__actions nav__item">
+          {usersignedin && (
+
+                      <div className="nav__button" id="get-help-button">
+
+                        <GetHelpButton course={props.course} currentUser={props.currentUser} key="get_help"/></div>)}
+          <div className="hamburger-wrapper ">
+            <TrainingHamburger/>
+          </div>
+        </div>
+      </div>
+  </header>
+)
+};
+
+TrainingNav.propTypes = {
+  course: PropTypes.object
+};
+
+const mapStateToProps = state => ({
+  course: state.course,
+  currentUser: getCurrentUser(state)
+});
+
+const mapDispatchToProps = {
+  fetchCourse,
+  fetchUsers,
+};
+
+export default connect(mapStateToProps,mapDispatchToProps)(TrainingNav);
--- a/app/assets/javascripts/training/components/training_slide_handler.jsx
+++ b/app/assets/javascripts/training/components/training_slide_handler.jsx
@@ -2,12 +2,12 @@ import React, { useEffect, useState } from 'react';
 import { useNavigate, useParams } from 'react-router';
 import { extend } from 'lodash-es';
 import { useDispatch, useSelector } from 'react-redux';
-import { fetchTrainingModule, setSlideCompleted, setCurrentSlide, toggleMenuOpen } from '../../actions/training_actions.js';
+import { fetchTrainingModule, setSlideCompleted, setCurrentSlide} from '../../actions/training_actions.js';
 import SlideLink from './slide_link.jsx';
-import SlideMenu from './slide_menu.jsx';
 import Quiz from './quiz.jsx';
 import Notifications from '../../components/common/notifications.jsx';
 import { FastTrainingAlert, fastTrainingAlertHandler } from './fast_training_alert';
+import TrainingNav from './training_nav.jsx'
 
 
 
@@ -64,11 +64,11 @@ const keys = { rightKey: 39, leftKey: 37 };
 
 const TrainingSlideHandler = () => {
   const training = useSelector(state => state.training);
+  //console.log(training)
   const routeParams = useParams();
   const navigate = useNavigate();
   const dispatch = useDispatch();
   const [baseTitle, setBaseTitle] = useState('');
-
   // useState for fastTrainingAlertHandler function
   const [isShown, setIsShown] = useState(false);
 
@@ -214,9 +214,6 @@ const TrainingSlideHandler = () => {
   }
 
   const { slideTitle, assessment, rawHtml } = getSlideInfo(training, I18n.locale);
-
-  const menuClass = training.menuIsOpen === false ? 'hidden' : 'shown';
-
   let quiz;
   if (training.currentSlide.assessment) {
     quiz = (
@@ -241,68 +238,25 @@ const TrainingSlideHandler = () => {
  if (training.currentSlide.wiki_page) {
    sourceLink = <span><a href={`https://meta.wikimedia.org/wiki/${training.currentSlide.wiki_page}`} target="_blank">{I18n.t('training.wiki_source')}</a></span>;
  }
-
-
-  const toggleMenuOpen_FC = (e) => {
-    e.stopPropagation();
-    dispatch(toggleMenuOpen({ currently: training.menuIsOpen }));
-  };
-
-  const closeMenu_FC = (e) => {
-    if (training.menuIsOpen) {
-      e.stopPropagation();
-      dispatch(toggleMenuOpen({ currently: true }));
-    }
-  };
-
   return (
     <div>
-      <Notifications />
-      <header>
-        <div
-          role="button"
-          tabIndex={0}
-          className="pull-right training__slide__nav"
-          onClick={toggleMenuOpen_FC}
-          onKeyDown={(e) => {
-            if (e.key === 'Enter' || e.key === ' ') {
-              e.preventDefault();
-              toggleMenuOpen_FC();
-            }
-          }}
-        >
-          <div className="pull-right hamburger">
-            <span className="hamburger__bar" />
-            <span className="hamburger__bar" />
-            <span className="hamburger__bar" />
-          </div>
-          <span className="pull-right training__slide__page-label">
-            {I18n.t('training.page_number', { number: training.currentSlide.index, total: training.slides.length })}
-          </span>
-        </div>
-        <SlideMenu
-          closeMenu={closeMenu_FC}
-          onClick={toggleMenuOpen_FC}
-          menuClass={menuClass}
-          currentSlide={training.currentSlide}
-          params={routeParams}
-          enabledSlides={training.enabledSlides}
-          slides={training.slides}
-        />
-      </header>
-      {loginWarning}
-      {pendingWarning}
-      <article className="training__slide">
-        {titlePrefix}
-        <h1>{slideTitle}</h1>
-        <div className="markdown training__slide__content" dangerouslySetInnerHTML={{ __html: rawHtml }} />
-        {quiz}
-        <footer className="training__slide__footer">
-          <span className="pull-left">{previousLink}</span>
-          {sourceLink}
-          <span className="pull-right">{nextLink}</span>
-        </footer>
-      </article>
+      <TrainingNav/>
+      <div className="container training">
+        <Notifications />
+        {loginWarning}
+        {pendingWarning}
+        <article className="training__slide">
+          {titlePrefix}
+          <h1>{slideTitle}</h1>
+          <div className="markdown training__slide__content" dangerouslySetInnerHTML={{ __html: rawHtml }} />
+          {quiz}
+          <footer className="training__slide__footer">
+            <span className="pull-left">{previousLink}</span>
+            {sourceLink}
+            <span className="pull-right">{nextLink}</span>
+          </footer>
+        </article>
+      </div>
     </div>
   );
 };
--- a/app/assets/stylesheets/modules/_popover.styl
+++ b/app/assets/stylesheets/modules/_popover.styl
@@ -6,6 +6,7 @@
 
   .pop
     background white
+    font-size 15px
     border 1px solid mischka
     border-radius 3px
     box-shadow 0 0 15px 0 rgba(0, 0, 0, .2)
@@ -99,4 +100,28 @@
 
   textarea
     height 180px
+    
++below(desktop)
+  .pop__container
+    display flex !important
+    justify-content center !important
+    width 100% !important
 
+    .pop
+      left 50% !important
+      right auto !important
+      transform translateX(-50%) scale(.7, .7) !important
+      -webkit-transform translateX(-50%) scale(.7, .7) !important
+      width 90vw !important
+      max-width 360px !important
+
+      &.open
+        transform translateX(-50%) scale(1, 1) !important
+        -webkit-transform translateX(-50%) scale(1, 1) !important
+
+      &:before,
+      &.open:before,
+      &.right:before
+        left 50% !important
+        right auto !important
+        margin-left -6px !important
--- a/app/assets/stylesheets/training.styl
+++ b/app/assets/stylesheets/training.styl
@@ -86,6 +86,8 @@ old-ie ?= false
 @import "modules/_tooltips"
 @import "modules/_popover"
 @import "modules/_nav_news"
+@import "modules/_controls"
+
 
 // Overwriting some table styles
 .training .training__slide
--- a/app/assets/stylesheets/training_modules/_nav.styl
+++ b/app/assets/stylesheets/training_modules/_nav.styl
@@ -19,11 +19,115 @@ body.training nav.top-nav
       content " > "
 
 .training__slide-header
-  padding 12px 0 16px 0
-  height auto
+  position relative
+  padding 16px 0
   background $training_primary
+  +below(desktop - 1)
+    margin-top 50px
+
+  .training__nav-container
+    display flex
+    align-items center
+    justify-content space-between
+    padding 0 40px
+    gap 30px
+
+    .training-btn
+      background-color #e5e7eb
+      color #1f2937
+
+    .training__breadcrumbs
+      flex 1
+      min-width 0
+
+      ol, ul, .breadcrumbs
+        margin 0
+        padding 0
+
+    .training__actions
+      display flex
+      align-items center
+      gap 230px
+
+  +below(desktop)
+    padding 8px 0 !important
+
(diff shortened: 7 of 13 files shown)


ragesoss (changes requested):
Thanks for taking this on — the extraction of `TrainingNav` / `TrainingHamburger` / `BreadCrumbs` is a sensible shape for the change, and the core feature does work: with `wiki_education` on, the Get Help button on a training slide correctly picks up the course and offers the staff contact options.

A note on how this review was produced: it was drafted with the help of an AI coding agent. It is not meant to be trusted automatically or acted on blindly. Read it, check the reasoning, and use or ignore each item on your own judgment — some of it may be wrong.

## CI

The `Check whitespace` step (`git diff --check`) is failing on five trailing-whitespace lines: `app/assets/stylesheets/modules/_popover.styl:103`, `app/controllers/training_controller.rb:92` and `:103`, and `app/views/training/slide_view.html.haml:3` and `:7`.

## Behavior changes that look like regressions

Each of the four below was reproduced locally against this branch and then re-checked against `master` to confirm it is introduced by this PR rather than pre-existing. The specs used are at the bottom of this comment.

**1. A non-existent module slug now returns 500 instead of 404.**

In `slide_view`, the assignment was moved above the nil guard:

```ruby
training_module = TrainingModule.find_by(slug: params[:module_id])
@training_module_name = training_module.name          # NoMethodError when nil
raise ActionController::RoutingError, 'not found' if training_module.nil?
```

`GET /training/students/lolnotarealmodule/<slide>` returns 404 on `master` and 500 here. Moving the assignment back below the guard fixes it; the name is only needed for the render path anyway.

**2. The Get Help button is not behind `Features.enableGetHelpButton`.**

`TrainingNav` gates the button on a new `data-usersignedin` attribute. The existing usage in `course_navbar.jsx:85` gates on `Features.enableGetHelpButton`, which `app/views/shared/_head.html.haml:41` defines as `Features.enable_get_help_button? && user_signed_in?` — and `enable_get_help_button?` is `ENV['wiki_education'] == 'true'`.

The new attribute reproduces the `user_signed_in?` half and drops the wiki_education half. With `enable_get_help_button?` stubbed false, the button still renders on the slide. The consequence is on the Programs & Events Dashboard: every signed-in user would get a button offering to contact "Wiki Education staff", with FAQ links hardcoded to `dashboard.wikiedu.org`, and a submission would be rejected by the guard at `alerts_controller.rb:95`.

`Features.enableGetHelpButton` is already available on this page — the training layout renders `shared/head` — so `usersignedin && Features.enableGetHelpButton` (or just the flag, since it already includes the signed-in check) should be enough, and the extra data attribute may not be needed at all.

**3. The module breadcrumb loses its translation.**

The Rails breadcrumb used `TrainingModule#translated_name` (`add_module_breadcrumb`). The React `BreadCrumbs` reads `data-module-name`, which the controller sets from `training_module.name`.

For a user with `locale: 'eu'` on `wikipedia-essentials`, `master` renders `Wikipediaren oinarria` and this branch renders `Wikipedia Essentials`. The first crumb still goes through `I18n.t`, so the two crumbs end up in different languages. `spec/features/training_translation_spec.rb` does not catch this because it only asserts on quiz text and the table of contents.

Setting `@training_module_name = training_module.translated_name` should restore it.

**4. `find_recent_course` can pick the wrong course, and can 404 the page.**

Two separate issues in the same method:

*Ordering.* `current_user.get_recent_course` is checked before `session[:training_return_to]`. For a user enrolled in course A and then course B, arriving at training from course A's page yields course B as the help context. The session value is the stronger signal about where the user actually came from, so it may be worth checking it first and falling back to the enrollment.

*Raising lookup.* Both branches call `find_course_by_slug`, which raises `ActionController::RoutingError` when the course is missing:

```ruby
def find_course_by_slug(slug)
  course = Course.find_by(slug:)
  raise ActionController::RoutingError.new('Not Found'), "Course #{slug} not found" if course.nil?
  return course
end
```

A `training_return_to` pointing at a course that has since been deleted therefore 404s the entire training slide — the slide becomes unreachable because of an incidental lookup for the help button's context. `Course.find_by(slug:)` with a nil check would degrade gracefully to the generic help content instead.

## Other observations

**`db/schema.rb` has 614 lines of unrelated churn.** The diff alphabetizes columns and adds `options: "ENGINE=InnoDB ROW_FORMAT=DYNAMIC"` on several tables — the signature of a local `db:schema:dump` against a differently-configured database. There is no migration in this PR, so `git checkout master -- db/schema.rb` should clear it.

**Global CSS changes from a training-scoped PR.** `app/assets/stylesheets/modules/_popover.styl` is picked up by `main.styl`'s `@import "modules/*"` glob, so both new rules apply app-wide, not just in training:

- `.pop { font-size: 15px }` restyles every popover in the app.
- The new `+below(desktop)` block re-centers and rescales every popover with `!important`.

Over 20 components use these classes (course-page Get Help, enroll button, category and tag editors, most of the settings dialogs). If the intent is training-only, scoping under a training selector or moving the rules into `training_modules/` would contain it.

Related: moving `margin-top 50px` out of the `+below($desktop)` block in `_training.styl` applies it at every width, and `.container.training` is also used by `app/views/training/index.html.haml` and `app/views/training/training_module.html.haml`, so the library and module index pages shift too.

**63 new `!important` declarations** across the stylesheet changes, plus some magic numbers (`gap 230px` in `.training__actions`). Much of this looks like it is fighting the `.pull-right` float utilities the markup still uses; switching the nav to flex layout without the floats would likely remove most of the need for the overrides.

**Debug output.** `console.log(params)` in `bread_crumbs.jsx:7`, and a commented-out `//console.log(training)` in `training_slide_handler.jsx`. Neither is caught by lint — `no-console` is not enabled in `eslint.config.mjs`.

**Dead code.** `slide_view` still calls `add_training_root_breadcrumb` and `add_module_breadcrumb(training_module)`, but the view no longer renders `render_breadcrumbs`, so those calls now do nothing on this page.

**Naming.** `User#get_recent_course` — the `get_` prefix is against Ruby convention (`Naming/AccessorMethodName`, though it is not enabled in this repo's RuboCop config). `recent_course` would fit the surrounding code better.

**Tests.** There are no spec changes. A new controller method, a new model method, and three new React components ship uncovered; all four issues above would have been caught by fairly small specs.

**Pre-existing, but in code this PR moves.** In `TrainingHamburger`, the `onKeyDown` handler calls `toggleMenuOpen_FC()` with no argument, so `e.stopPropagation()` throws when the menu is opened by keyboard. The same bug is on `master` at `training_slide_handler.jsx:270`, so it is not introduced here — but it is cheap to fix while the code is being relocated.

## Testing done

Ran on this branch merged with current `master`: `yarn build` (succeeds), RuboCop and ESLint on the changed files (clean as configured), `spec/controllers/training_controller_spec.rb`, `spec/models/user_spec.rb`, `spec/features/training_tool_spec.rb`, `spec/features/training_library_spec.rb`, `spec/features/training_translation_spec.rb`, and the ticket specs that exercise popover styles. All pass — no regressions in existing coverage. The gaps are the paths nothing currently covers.

<details>
<summary>Exploratory specs used to confirm the four issues (AI-written; run locally, not proposed for merge as-is)</summary>

These were written during review to turn the code-reading into something reproducible. They are AI-generated and were run on this branch and on `master`, but they have not been polished for merge — the naming and placement would need work, and the `find_course_by_slug` stub in the second example is crude. They may still be a useful starting point for real coverage.

`spec/controllers/exploratory_7048_slide_view_spec.rb` — three failures on this branch, all passing behavior on `master`:

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe TrainingController, type: :request do
  let(:user) { create(:user) }
  let(:library_id) { 'students' }
  let(:module_id) { TrainingModule.all.first.slug }
  let(:slide_id) { TrainingModule.all.first.slides.first.slug }

  before do
    TrainingModule.load_all
    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(user)
  end

  describe 'a module slug that does not exist' do
    it 'returns 404 rather than raising NoMethodError' do
      get "/training/#{library_id}/lolnotarealmodule/#{slide_id}"
      expect(response.status).to eq(404)
    end
  end

  describe 'course context for the Get Help button' do
    let(:course_a) { create(:course, slug: 'SchoolA/Course_A', title: 'Course A', school: 'SchoolA') }
    let(:course_b) { create(:course, slug: 'SchoolB/Course_B', title: 'Course B', school: 'SchoolB') }

    it 'uses the course the user navigated from, not their newest enrollment' do
      create(:courses_user, course: course_a, user:)
      create(:courses_user, course: course_b, user:)

      allow_any_instance_of(TrainingController).to receive(:session)
        .and_return({ training_return_to: '/courses/SchoolA/Course_A/students' })

      get "/training/#{library_id}/#{module_id}/#{slide_id}"
      expect(assigns(:course_slug)).to eq('SchoolA/Course_A')
    end

    it 'does not 404 when the session return_to points at a course that no longer exists' do
      allow_any_instance_of(TrainingController).to receive(:session)
        .and_return({ training_return_to: '/courses/Gone/Deleted_Course' })

      get "/training/#{library_id}/#{module_id}/#{slide_id}"
    end
  end
end
```

`spec/features/exploratory_7048_training_get_help_spec.rb` — the second example fails on this branch. Note the `have_css('.training__slide-header')` sync point in each example: without it, `have_no_css` resolves before React mounts and the example passes for the wrong reason.

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe 'Get Help button on training slides', type: :feature, js: true do
  let(:user) { create(:user) }
  let(:expert) { create(:user, username: 'Wiki Expert') }
  let(:course) { create(:course, slug: 'SchoolA/Course_A', school: 'SchoolA', title: 'Course_A') }
  let(:training_module) { TrainingModule.all.first }
  let(:slide) { training_module.slides.first }
  let(:slide_path) { "/training/students/#{training_module.slug}/#{slide.slug}" }

  before do
    TrainingModule.load_all
    create(:courses_user, course:, user: expert,
                          role: CoursesUsers::Roles::WIKI_ED_STAFF_ROLE)
    login_as user
  end

  describe 'feature gating' do
    it 'shows the button when enable_get_help_button? is true' do
      allow(Features).to receive(:enable_get_help_button?).and_return(true)
      visit slide_path
      expect(page).to have_css('.training__slide-header')
      expect(page).to have_css('#get-help-button')
    end

    it 'hides the button when enable_get_help_button? is false' do
      allow(Features).to receive(:enable_get_help_button?).and_return(false)
      visit slide_path
      expect(page).to have_css('.training__slide-header')
      expect(page).to have_no_css('#get-help-button')
    end
  end

  describe 'course context' do
    before { allow(Features).to receive(:enable_get_help_button?).and_return(true) }

    it 'offers the staff contact options for the course' do
      visit slide_path
      expect(page).to have_css('.training__slide-header')
      find('#get-help-button button').click
      expect(page).to have_css('.contact-wikipedia-help')
    end
  end
end
```

`spec/features/exploratory_7048_breadcrumb_translation_spec.rb` — fails on this branch, passes on `master`. Requires the `training/slide_translations` VCR cassette.

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/data_cycle/training_update"

describe 'Training slide breadcrumbs', type: :feature, js: true do
  let(:basque_user) { create(:user, id: 2, username: 'ibarra', locale: 'eu') }

  before do
    TrainingModule.destroy_all
    TrainingSlide.destroy_all
    TrainingLibrary.destroy_all
    allow(Features).to receive(:wiki_trainings?).and_return(true)
    VCR.use_cassette 'training/slide_translations' do
      TrainingUpdate.new(module_slug: 'all')
    end
    login_as(basque_user, scope: :user)
  end

  it 'shows the translated module name in the breadcrumb' do
    training_module = TrainingModule.find_by(slug: 'wikipedia-essentials')

    visit '/training/editing-wikipedia/wikipedia-essentials/five-pillars-quiz-1'
    expect(page).to have_css('.breadcrumbs')
    expect(find('.breadcrumbs')).to have_content(
      I18n.with_locale(:eu) { training_module.translated_name }
    )
  end
end
```

</details>

*Drafted in a Claude Code session (~40 min, 3 interactions). Sage set the scope and read the triage and full-review summaries, but did not independently verify the findings — the verification was done by the agent: each of the four regressions was reproduced with a spec on this branch and re-run against `master` to confirm it is PR-introduced, and `yarn build`, RuboCop, ESLint and the existing training specs were run locally. The exploratory specs quoted above are AI-written; they were run, but not reviewed line by line by a human.*

GonzaFP:
Hello @ragesoss 
I encountered an obstacle as I was writing an rspec test for the instance method that I added to the User model.
The method recent_course returns a user's recently enrolled course therefore to test it, I need at least two distinct courses. 
The problem is that spec/factories/courses.rb hardcodes every course's slug
So the second create(:course) in the same test fails:
 ActiveRecord::RecordNotUnique:
       Mysql2::Error: Duplicate entry 'WÏNTR/Ųnderwater_básket-weaving_(spring_2015)' for key 'index_courses_on_slug'
Could this be fixed so the slug is unique?
Thanks

ragesoss:
That one's easy to work around in the spec itself — no factory change needed.

The course factory hardcodes `slug` because most specs only need one course, but you can override any factory attribute at `create` time. Most specs that need two courses already do this (there are ~77 of them). For your test:

```ruby
let(:user) { create(:user) }
let(:older_course) { create(:course, slug: 'School/Older_(2024)', title: 'Older') }
let(:newer_course) { create(:course, slug: 'School/Newer_(2025)', title: 'Newer') }

it 'returns the most recently enrolled course' do
  create(:courses_user, course: older_course, user:)
  create(:courses_user, course: newer_course, user:)
  expect(user.get_recent_course).to eq(newer_course)
end
```

Passing `title` too keeps the two courses distinguishable in failure output, but only `slug` is required for uniqueness.

GonzaFP:
Thanks for the clarification.

GonzaFP:
Hello @ragesoss 
I have addressed the review comments in the latest commit.
Thanks!

ragesoss (changes requested):
Most of the previous round is addressed — the feature gating, the 404, the translated breadcrumb, the `find_recent_course` ordering and raising lookup, the debug output, the dead breadcrumb calls, the method naming, the keyboard handler, and nearly all of the CSS scoping. The `!important` count is down from 63 to 1, and the popover and `_training.styl` changes are now properly contained. Thanks for working through all of it.

Same note as last time on how this was produced: it was drafted with the help of an AI coding agent. It is not meant to be trusted automatically or acted on blindly. Read it, check the reasoning, and use or ignore each item on your own judgment — some of it may be wrong.

The blocking item is that the new tests don't pass.

## Five of the new tests fail

Run on this branch merged with current `master`:

**`spec/controllers/training_controller_spec.rb` — 2 failures**

```
NameError:
  undefined local variable or method 'course' for
  #<RSpec::ExampleGroups::TrainingController::FindRecentCourse::WhenThereIsNoSessionButTheUserHasARecentCourse>
# ./spec/controllers/training_controller_spec.rb:259
```

The context declares `old_course` and `new_course`, but the `before` block references `course`, which is only defined by the `let!` in the sibling context above it. Both new lets are lazy, so neither course is ever created:

```ruby
context 'when there is no session but the user has a recent course' do
  let(:old_course) { create(:course) }
  let(:new_course) { create(:course, slug: 'School/New_Course') }

  before do
    create(:courses_user, course_id: course.id, user_id: user.id)  # `course` is not in scope
  end
```

Worth fixing carefully rather than just renaming the variable: this is the one branch that `User#recent_course` exists to serve, and right now it has no working coverage. It also needs `old_course` to be created *before* `new_course` for the ordering assertion to mean anything — `let!` on both, or explicit creation in order.

**`spec/controllers/training_get_help_button_spec.rb` — 1 failure**

```
Capybara::ElementNotFound: Unable to find css ".training-btn"
# ./spec/controllers/training_get_help_button_spec.rb:23
```

`.training-btn` occurs in exactly two places in the repo: this line, and `app/assets/stylesheets/training_modules/_nav.styl:56`. Nothing renders that class. The button the spec wants is `#get-help-button button`.

**`test/components/bread_crumbs.spec.js` — both tests**

```
TypeError: act is not a function
```

Line 10 is `const { act } = require('react')`, which is the React 19 location. This repo is on React 18, and every other test under `test/` uses `require('react-dom/test-utils')`. Changing that one line makes both tests pass.

## A new regression: the breadcrumb is empty when signed out

Moving `@training_module_name` below the nil guard fixed the 404, but it also moved it inside `if current_user`. On `master` the module crumb came from `add_module_breadcrumb`, which ran for every visitor; now it comes from `data-module-name`, which is only set for signed-in users. A signed-out visitor on a training slide gets `Training Library >` with a dangling separator and an empty second crumb.

Reproduced on this branch and re-run against `master` to confirm it is introduced here:

```ruby
it 'names the training module in the second crumb' do
  visit "/training/students/#{module_id}/five-pillars"
  expect(page).to have_css('ol.breadcrumbs')
  expect(find('ol.breadcrumbs')).to have_content(
    TrainingModule.find_by(slug: module_id).translated_name
  )
end
```

```
expected to find text "Wikipedia policies" in "Training Library"
```

Setting `@training_module_name` between the nil guard and the `if current_user` block should fix it — the name doesn't depend on the user. (The existing `training_nav_spec.rb` example for breadcrumbs is signed out, but only asserts that `ol.breadcrumbs` exists, so it passes either way.)

## Still outstanding from the last round

**`db/schema.rb`.** Still ~1,100 diff lines vs `master`, and it now produces a merge conflict. Same schema version on both sides — the diff is alphabetized columns, `options: "ENGINE=InnoDB ROW_FORMAT=DYNAMIC"`, and changed text column sizes. `git checkout master -- db/schema.rb` clears it.

**Trailing whitespace.** Two of the five were fixed, and five new ones were added, for seven against `master`: `app/controllers/training_controller.rb:101`, `app/views/training/slide_view.html.haml:3` and `:7`, `spec/controllers/training_get_help_button_spec.rb:18` and `:20`, `spec/features/training_nav_spec.rb:8` and `:32`. Note the CI step is `git diff --check HEAD~1`, and `HEAD` is currently a merge commit, so it would pass by accident — the whitespace is still there.

## Other observations

**`js_error_expected: true` on three new examples.** That flag switches off severe-JS-error detection for the example. Running the same flow (visit the slide, open the Get Help popover) without it passes clean, so the flag looks like it can just be dropped.

**`'shows no contact staff options for no course'` can't fail.** The `before` block enrolls the user in a course, and the example never opens the popover, so `p.target-users` would be absent regardless.

**The `.pull-right` override in `_slides.styl`** is the one remaining `!important`:

```stylus
+below(desktop)
  .pull-right,
    .pull-right *
      float none !important
```

It's confined to the training bundle, but not to the nav. Below 920px it also unfloats the Next/Done slide button and its wrapper (`training_slide_handler.jsx:175`, `:255`), the slide-menu close button (`slide_menu.jsx:59`), the TOC slide count (`training_module_handler.jsx:51`) and the module progress badge (`show.html.haml:31`). Since the hamburger no longer uses `.pull-right`, it may not be needed at all now.

**Smaller things:**

- `training_controller.rb:88` ends with a stray `#` after `include?("/courses/")`.
- RuboCop reports one offense: `Layout/LineLength` at `spec/controllers/training_controller_spec.rb:240`.
- `describe TrainingController` was changed to the string `'TrainingController'` — unclear whether that was deliberate.
- Both new spec files are missing `# frozen_string_literal: true`.
- `spec/controllers/training_get_help_button_spec.rb` is a `type: :feature` spec living in `spec/controllers/`.
- `modules/_popover.styl` now has no functional change left — only removed blank lines, and it lost its trailing newline. Reverting the file would keep the diff honest.
- `.training-btn` in `_nav.styl:56` styles nothing.
- The mobile `.training__breadcrumbs { text-align: center }` rule is overridden by `.training__slide-header ol.breadcrumbs { text-align: left }`, so breadcrumbs stay left-aligned on narrow screens.
- `app/views/training/training_module.html.haml` has a whitespace-only change (`] }` → `]}`).

## Testing done

On this branch merged with current `master` (with `db/schema.rb` taken from `master` to resolve the conflict): `yarn build` succeeds; ESLint clean on the changed JS; RuboCop reports the one offense above; `spec/features/training_tool_spec.rb`, `training_library_spec.rb` and `training_translation_spec.rb` all pass (26 examples), so no regressions in existing coverage. The five failures above are all in code added by this PR. Desktop and mobile signed-in renders of the new nav both look correct.

CI has not run on the latest commit — the run is sitting in `action_required`, awaiting maintainer approval. Once it runs it should catch the five failures.

*Drafted in a Claude Code session (~35 min, 2 interactions). Sage asked for the review and approved posting after reading the summary, but did not independently verify the findings — the verification was done by the agent: the five test failures were reproduced locally, the signed-out breadcrumb regression was reproduced on this branch and re-run against `master` to confirm it is PR-introduced, the `react-dom/test-utils` one-line change was applied locally to confirm it fixes the JS tests, and the build, linters and existing training specs were run. The breadcrumb spec quoted above is AI-written and was run, but not reviewed line by line by a human.*

GonzaFP:
Hello @ragesoss 
I have addressed the review comments.
Thanks.

ragesoss:
Please mark it ready for review in the future, if it's ready for me to have another look.

ragesoss (changes requested):
The previous round is fully addressed: all five failing tests now pass (and the `recent_course` ordering test now genuinely exercises the ordering), the signed-out breadcrumb is fixed and covered, `db/schema.rb` matches `master`, the whitespace is gone, `js_error_expected` is dropped, the `.pull-right` override is removed, and the smaller items are all cleaned up. Desktop and mobile renders of the new nav look correct.

Same note as before: this was drafted with the help of an AI coding agent. It is not meant to be trusted automatically or acted on blindly — read it, check the reasoning, and use or ignore each item on your own judgment.

## A non-ASCII course slug in `return_to` makes the slide return 500

`explicit_return_to` stores `params[:return_to]`, which Rails has already percent-decoded. The dashboard's own course-page training links send one — `assignment_view_context.rb:112`, `student_status_context.rb:121` and `trainings_assignment_view_context.rb:151` build `?return_to=/courses/<slug>`. For a slug like `Школа/Курс_(2026)`, the session then holds the raw Unicode path, and `URI.parse` in `find_recent_course` raises `URI::InvalidURIError`. Every slide view after that returns 500 until the session changes.

`find_recent_course` runs for every signed-in user regardless of `enable_get_help_button?`, so this would reach the Programs & Events Dashboard, where non-ASCII slugs are common.

A related, quieter case: when the session value comes from a browser referer, it is percent-encoded, so `URI.parse` succeeds but the path segments stay encoded and `Course.find_by(slug:)` returns nil. The help button then loses its course context — and because the session branch doesn't fall back to `recent_course` when the lookup misses, it drops to the generic content even for an enrolled user.

`ClaimVerificationExercisesController#course_from_return_to` / `#slug_from_course_path` (`claim_verification_exercises_controller.rb:103`) already solves both: it `CGI.unescape`s the slug and rescues `URI::InvalidURIError`. Reusing that logic (perhaps extracted somewhere both controllers can share), and falling back to `recent_course` when the session lookup returns nil, would cover it.

Reproduced with this spec, which fails on this branch and — for the first example — passes on `master`:

```ruby
# frozen_string_literal: true

require 'rails_helper'

describe TrainingController, type: :request do
  let(:user) { create(:user) }
  let(:module_id) { TrainingModule.all.first.slug }
  let(:slide_path) { "/training/students/#{module_id}/five-pillars" }
  let(:slug) { 'Школа/Курс_(2026)' }
  let!(:course) { create(:course, slug:, title: 'Курс', school: 'Школа') }

  before do
    TrainingModule.load_all
    allow_any_instance_of(ApplicationController).to receive(:current_user).and_return(user)
    create(:courses_user, course:, user:)
  end

  it 'renders the slide after an explicit non-ASCII return_to' do
    get "/training/students/#{module_id}", params: { return_to: "/courses/#{slug}/home" }
    get slide_path
  end

  it 'finds the course from a percent-encoded referer' do
    encoded = "#{ERB::Util.url_encode('Школа')}/#{ERB::Util.url_encode('Курс_(2026)')}"
    get "/training/students/#{module_id}",
        headers: { 'HTTP_REFERER' => "http://www.example.com/courses/#{encoded}/home" }
    get slide_path
    expect(assigns(:course)).to eq(course)
  end
end
```

## Smaller things

- `slide_menu.jsx:27` has a stray `` `  1` `` after `=> {`. It's a no-op expression statement, so nothing breaks and lint doesn't flag it, but it's the only remaining change in that file.
- `slide_view.html.haml` is now a single 311-character line. Splitting the attribute hash across lines again (without the trailing spaces) would be easier to read.
- `spec/features/training_get_help_button_spec.rb` uses 4-space indentation and has no trailing newline.
- `training-btn` is now on the shared `GetHelpButton`, so it also renders on the course page (unstyled there, since the rule lives in the training bundle). Targeting `.training__nav-container #get-help-button button` in `_nav.styl` would keep the shared component free of training-specific classes. Optional.

## Testing done

On this branch merged with current `master`: `yarn build` succeeds; RuboCop and ESLint are clean on the changed files; `training_controller_spec.rb`, `user_spec.rb`, `training_get_help_button_spec.rb`, `training_nav_spec.rb`, `training_tool_spec.rb`, `training_library_spec.rb` and `training_translation_spec.rb` all pass (104 examples), as does `test/components/bread_crumbs.spec.js`. Desktop (1280px) and mobile (420px) renders of the nav and open popover were checked by screenshot.

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage asked for the review and approved posting after reading the draft, but did not independently verify the findings. The agent did the verification: it checked each prior item against the current code, reproduced the `return_to` 500 on this branch and re-ran it against `master` to confirm this PR introduces it, and ran the build, linters, PR specs and existing training specs. The spec quoted above is AI-written and was run, but no human reviewed it line by line.*

PR #7050, opened 2026-09-09
Fix 13 open Dependabot vulnerability alerts

ragesoss:
## What this PR does

Clears all 13 open Dependabot vulnerability alerts on this repository — one
RubyGems alert and twelve npm alerts — plus three additional findings that
`yarn npm audit` reported but Dependabot had not yet alerted on. After this
branch, `yarn npm audit --all --recursive` reports "No audit suggestions".

Most of the alerts were satisfied by version bumps or by `resolutions`
overrides, following the existing convention in the `resolutions` block of
`package.json` for pinning patched transitive dependencies. One alert needed
real code changes: reaching a patched `decode-uri-component` required moving
`query-string` from 7 to 9, which crosses a CommonJS-to-ESM boundary.

Alerts addressed, by package:

| Package | Change | Alerts |
|---|---|---|
| `mail` (rubygems) | 2.9.0 → 2.9.1 | 479 |
| `@tiptap/core` | 3.27.1 → 3.31.3, transitively — the diff bumps `@tiptap/pm`, `@tiptap/react` and `@tiptap/starter-kit` from `^3.27.1` to `^3.31.3` | 485, 475 |
| `qs` | 6.15.3 → 6.16.0 (new resolution) | 484, 478 |
| `browserslist` | 4.28.2 → 4.28.9 | 483, 482 |
| `fast-uri` | 3.1.5 → 3.1.7 (resolution bump) | 481, 480, 477, 476 |
| `postcss-selector-parser` | 7.1.1 → 7.1.6, 6.0.10 → 6.1.4 | 474 |
| `decode-uri-component` | 0.2.2 → 0.5.0, via `query-string` 9 | 473 |

Three non-Dependabot audit findings came along in the same pass, since all are
dev/build-time only and two already had `resolutions` entries being edited
anyway: `svgo` → `^2.8.4`, `js-yaml@^4.1.1` → `^4.3.2`, and a new `colord` →
`^2.9.4`.

### The query-string upgrade

`query-string@9.5.1` is the only release that depends on a patched
`decode-uri-component` (`^0.5.0`), so a `resolutions` override could not solve
alert 473 on its own — `decode-uri-component` 0.5.0 is ESM-only and
`query-string` 7 is CommonJS, so forcing the patched version under the old
parent would have broken the build.

query-string v9 is ESM-only and exposes everything on its **default** export,
with no named exports at all — in 9.5.1 the whole of `index.js` is
`import * as queryString from './base.js'; export default queryString;`, and
the named functions live in `base.js`, which the `exports` map does not expose
as a subpath. The 13 existing
`import { parse, stringify } from 'query-string'` call sites would therefore
have failed to build. Rather than rewrite every call site, this adds a small
shim at `app/assets/javascripts/utils/query_string.js` that re-exports the two
functions by name, so the 13 imports only needed their path repointed and no
call site changed.

`stringify`/`parse` output was checked against v7 for every shape this
codebase passes in — key sorting, `false`, `null`, `undefined`, and arrays —
and is byte-identical, so this is an import-mechanics change only.

`jest.config.js` gains `query-string`, `decode-uri-component`,
`split-on-first`, and `filter-obj` in `transformIgnorePatterns` so the new
ESM-only dependency chain is transformed to CJS for the test run.

Relevant docs: [query-string 9 release
notes](https://github.com/sindresorhus/query-string/releases/tag/v9.0.0).

### The prosemirror deduplication (second commit)

The `@tiptap/*` bump initially broke the wysiwyg editor. Raising `@tiptap/pm`
to 3.31.3 brought requirements of `prosemirror-model: ^1.25.11` and
`prosemirror-view: ^1.42.3`, while the lockfile's existing 1.25.9 / 1.41.9
entries still satisfied every other consumer's range — so yarn added a second
copy of each rather than lifting the shared one, leaving three copies of
`prosemirror-model` in `node_modules`. prosemirror requires a singleton: with
two copies loaded the editor throws `RangeError: Can not convert <> to a
Fragment (looks like multiple versions of prosemirror-model were loaded)`, and
paragraph breaks are dropped from ticket replies composed in the editor.

`spec/features/ticket_reply_formatting_spec.rb:16` caught it — the only failure
among CI's 4049 examples on the first commit.

The second commit fixes it with `yarn dedupe prosemirror-model
prosemirror-view`, collapsing each back to a single entry (1.25.11 / 1.42.3)
that satisfies all ranges. That is a lockfile-only change. Adding
`resolutions` entries was the other candidate fix and was passed over: this
repo's `resolutions` block is a security-pin list already flagged below as
overgrown, and prosemirror singleton-ness is not a vulnerability pin.
Recurrence on a future tiptap bump would not be silent — it throws, and the
ticket reply spec catches it.

The same commit reverts the `list.js` and `slick-carousel` checksum lines to
master's values. Those are git dependencies whose checksums are
environment-dependent; `git log -L` on that line shows it oscillating between
exactly two values across years of commits, so the branch should not carry a
flip. `yarn install --immutable` passes as CI runs it.

One duplicate was deliberately left in place: the bumps also produced two
`caniuse-lite` entries (1.0.30001791 + 1.0.30001810). It would dedupe cleanly,
but it is build-time browser data with no singleton requirement, and
collapsing it would nudge cssnano's minification inputs for no benefit.

## AI usage

Claude Code (Claude Opus 5) wrote all of the code in this PR, under short,
terse human direction — the operator's instruction was a single sentence
asking for a pass at the open Dependabot vulns on a fresh branch, followed by
a request to commit and open a PR. The human contribution here was direction
and the decision to do the work, not authorship; the alert triage, the choice
of `resolutions` versus direct bumps, the query-string shim approach, and the
verification steps were all Claude Code's. Claude Code also wrote the commit
message (marked as such in the commit, along with a `Co-Authored-By:`
trailer).

Claude Code got the query-string migration backwards on its first attempt. It
assumed v9 had dropped its default export and converted the one default import
to a named import, when the truth is the opposite — v9 has only a default
export. The test suite caught this: 7 failures across 3 suites, all tracing to
`stringify` being `undefined` and being swallowed by a `catch` in
`media_wiki_classes.js`. The shim module was the fix.

The second commit came out of a separate Claude Code session, prompted by a
one-sentence operator instruction to read the review on this PR and work in
fixes if the findings required them. It confirmed the reviewing session's
blocking finding independently rather than taking it on trust — reproducing the
spec failure against the first commit's lockfile, then re-running against the
deduped one — and made the dedupe-versus-`resolutions` call itself. The
operator's only other input was approval to commit, push, and edit this
description.

Scale of effort: two commits from two short, focused sessions — roughly half an
hour each — with terse single-sentence direction and no human iteration in
between. Reviewers should calibrate accordingly: the human contribution across
both sessions was direction and approval, not authorship or independent
verification.

The "What this PR does" summary and the analysis above were also drafted by
Claude Code and may contain errors. This PR description was drafted using
Claude Code and the `prepare-pr` skill.

## Screenshots

No UI changes. The `query-string` swap touches 13 UI and action modules, but
`stringify`/`parse` produce byte-identical output to the previous version, so
there is no visual difference to capture.

- **Correction to an earlier draft of this description.** It called the tiptap
  bump the least-covered change here and asked for a manual click-through of
  "the training module composer" before merging. That was wrong on both counts.
  No training module composer sits behind tiptap: `tiptap` is imported in
  exactly one file, `app/assets/javascripts/components/common/wysiwyg_editor.jsx`
  (plus its `youtube_iframe.js` extension), reached from exactly two
  `wysiwyg={true}` call sites — `components/tickets/new_reply_form.jsx:172` and
  `components/timeline/block.jsx:166`; `settings/views/site_notice_form.jsx`
  passes `wysiwyg={false}`. Both live call sites have feature-spec coverage, and
  that coverage is what caught the prosemirror duplication. The remaining tiptap
  risk is editor behaviour those specs do not assert, not an unspec'd surface.
- **The full Ruby suite has now run on this branch, via CI.** The first
  commit's run was 4049 examples with a single failure — the wysiwyg one fixed
  by the second commit. Local verification on the first commit covered: `yarn
  test` (78 suites, 495 tests, 0 failures), a production `yarn build` (0
  errors), `yarn lint` and `yarn lint-non-build` (0 errors; the 252
  `i18next/no-literal-string` warnings are pre-existing), and `bundle exec
  rspec spec/mailers` (35 examples, 0 failures) as a smoke test for the `mail`
  bump. The second commit adds: the wysiwyg and ticket feature spec set at 21
  examples / 0 failures, `yarn test` at 78 suites / 495 tests, four clean `yarn
  build` runs, `yarn npm audit --all --recursive` reporting no suggestions,
  ESLint on all changed JS, and `yarn install --immutable`.
- **Including the three non-Dependabot audit findings was a judgment call.**
  `svgo`, `js-yaml`, and `colord` were not in the alert list. They are all
  dev/build-time only and cheap to bump, and folding them in is what gets the
  audit to fully clean, but they can be split into a separate PR if reviewers
  would rather keep this one strictly scoped to the Dependabot list.
- `fast-uri` deliberately stayed on the 3.x line (`^3.1.6`, resolving to
  3.1.7) rather than jumping to the available 4.x major, since only the patch
  was needed.
- The `resolutions` block in `package.json` continues to grow. It is now well
  past 50 entries, many of them long-stale pins for vulnerabilities in
  dependencies that have since moved on. A separate cleanup pass to prune
  entries that are no longer doing anything would probably be worthwhile.


ragesoss (commented):
**Note on how to use this feedback:** this review was produced largely by an AI coding agent. It is not meant to be trusted automatically or acted on blindly — read it, evaluate the reasoning, and use or discard each point on your own judgment.

## Blocking: the tiptap bump duplicates prosemirror and breaks the WYSIWYG editor

CI is red on both build jobs, and the failure appears to be substantive rather than flake — both runs failed the same example with the same error:

```
spec/features/ticket_reply_formatting_spec.rb:16
  Uncaught RangeError: Can not convert <> to a Fragment
  (looks like multiple versions of prosemirror-model were loaded)
```

`@tiptap/pm@3.31.3` requires `prosemirror-model: ^1.25.11` and `prosemirror-view: ^1.42.3`. The lockfile's existing entries (1.25.9 / 1.41.9) still satisfy every other consumer's range, so yarn appears to resolve a second copy rather than lifting the shared one. `master` has no duplicated `prosemirror-*` packages; this branch has two.

The impact looks worse than console noise. The spec asserts the reply body should *not* be `"First paragraph.Second paragraph."`, and it is — paragraph breaks are dropped from ticket replies composed in the WYSIWYG editor.

Two fixes were tried locally, each taking the wysiwyg-related spec set from 21 examples / 1 failure to 21 / 0:

- `yarn dedupe 'prosemirror-*'` — lockfile-only, no `package.json` change
- adding `"prosemirror-model": "^1.25.11"` and `"prosemirror-view": "^1.42.3"` to `resolutions` — consistent with the convention this PR already uses elsewhere, and likely more durable against a later install re-splitting them

The PR's own open questions flag tiptap as the least-covered change, warranting a manual click-through. It was in fact covered — `ticket_reply_formatting_spec` caught it. The training module composer remains unspec'd, so that manual check still seems worth doing once the duplication is resolved.

## Claims that checked out

- **`stringify`/`parse` output identical between query-string 7.1.1 and 9.5.1** — checked by installing both versions side by side and diffing 23 cases: key sorting, `false` / `null` / `undefined`, arrays, repeated keys, unicode, `+`-as-space, and malformed percent-encoding. No differences. The shim also covers the whole surface in use; `parse` and `stringify` are the only imports anywhere in the tree.
- `yarn npm audit --all --recursive` reports no suggestions.
- `yarn test` (78 suites, 495 tests), `yarn build`, and ESLint on all 15 changed JS files are clean. `spec/mailers` passes 35/35 against `mail 2.9.1`. The `Gemfile.lock` diff is exactly the one-line `mail` bump.

## Two minor points

1. The description says `@tiptap/core` moves 3.27.1 → 3.31.3, but the diff bumps `@tiptap/pm`, `@tiptap/react`, and `@tiptap/starter-kit`; `@tiptap/core` moves transitively. Cosmetic, though the alert table credits 485/475 to a package the diff does not touch.
2. `list.js` and `slick-carousel` checksums differ from master's values. Since CI runs `yarn install --immutable`, this may be worth a glance — though git-dependency checksums are often environment-dependent, so this may be nothing.

*Drafted in a Claude Code session (~40 min, ~4 interactions). The human instruction was a single request to review the PR; the triage, root-cause analysis, local reproduction, and both fix verifications were the agent's. Sage read the summarized findings and approved posting, but did not independently reproduce the failure or verify the analysis. The version-differential test cited above is AI-written code that Sage did not review. Local verification ran in a working tree shared with another concurrent agent session, which adds some noise — the CI failure itself reproduced independently on GitHub, but point 2 above is the least reliable item here.*

PR #7051, opened 2026-09-09
Cut production AI detection over to Pangram 4

ragesoss:
## What this PR does

Cuts production AI-detection alerting over from Pangram 3 to Pangram 4, and commits the
analysis script that produced the evidence for the decision.

Pangram is shutting down the v3 API on **30 September 2026**. Phase 0 (#7034, merged and
deployed 4 September) built the v4 async client, response parser and detector registry
entry. Phase 1 scored six samples (2,637 units) under both models and compared them. This
PR is the cutover itself, plus the `pangram_cutover.py` analysis script that Phase 1 used.

The cutover is deliberately as close to no-change as possible: the alert rule
(`max_ai_likelihood > 0.9`), the AI-Assisted and humanizer handling, the alert detail
readers and every email template are untouched. The only intended difference is the score
shift between models.

**Commit 1 — analysis tooling.** `docs/analytics_scripts/ai_usage_research/detector_comparison/analysis/pangram_cutover.py`
is a companion to the existing `analyze.py`. Where `analyze.py` compares detectors on equal
terms, this one answers what changes for the alert pipeline: positive rate per production
score band reweighted to the production population, the old detector's reproducibility
against its own stored scores, threshold transitions per model version, v4-only signals
(humanized windows, score distribution per document label), window geometry, and baseline
false positives per model version. It imports `analyze.py` for palette, loading and table
rendering so both scripts' output stays one system. Also: a shared fix in `analyze.py`
(all-numeric tables printed integer counts as `58.000`), a README section, and a
`docs/analytics_scripts/**/*.csv` gitignore rule — the exports pair public Wikipedia diff
URLs with AI-likelihood scores, and this repo is public, so the data stays local while the
scripts are committed.

**Commit 2 — the cutover.**

- `CheckRevisionWithPangram::DETECTOR_KEY` moves to `PANGRAM_V4_KEY`. Phase 1 projects the
  alert rate at 9.5% under v4 against 10.8% actual production since May, with 0 false
  positives on 587 pre-ChatGPT baseline units.
- `already_checked?` now filters `check_origin` to `[nil, COURSE_UPDATE_ORIGIN]`. Without
  it, a detector-comparison row on a revision production had not yet checked would suppress
  the production check and its alert — which is why no revision from a running course could
  be sampled until now. `NULL` is included because the column was added in March 2026 and
  never backfilled: every earlier row carrying an `article_id` came from a production check,
  and excluding `NULL` would re-bill each pre-March revision that gets re-queued.
  `RevisionAiScoresStatsController#set_scores` already treats `NULL` the same way.
  `check_type` is deliberately *not* filtered, so revisions already scored under v3 stay
  checked rather than being re-billed under v4.
- Detector failures are now recorded rather than only raised. Each failure writes a
  `RevisionAiScore` row with nil `avg_ai_likelihood` — which `already_checked?` has always
  read as "check this revision again" — and only failures a retry could clear are re-raised.
  Terminal, so swallowed: a 4xx `RequestError` (malformed request, exhausted credits) and
  `TaskFailed`. Retryable, so re-raised: 5xx, `TaskTimeout`, Faraday and JSON errors.
- `AiDetectionWorker` gets `retry: 3` instead of Sidekiq's default 25. The async v4 endpoint
  adds failure modes and every attempt bills, so 25 attempts on a persistent failure was the
  expensive case.
- Both AI-edit-alert mailer previews move to the v4 `details` shape, taken from the
  3 September live probe rather than invented: one window, headline "AI Generated",
  prediction "We believe that this entire text is AI.", version 4.0, plus the
  `humanized_window_count` and `max_humanizer_score` keys the parser adds for v4. One window
  is the representative case — Phase 1 found v4 returns a single window for 79% of units.
  No email copy changed; the previews only feed it different data.
- `BuildAiDetectionSampleFromRecentScores` hardcoded the v3 key as its `check_type` default,
  so a post-cutover recent-scores sample would have silently sampled the retired detector.
  It now defaults to `CheckRevisionWithPangram::DETECTOR_KEY`.

Pangram 3 stays in the registry and `PangramApi.v3` stays callable, so last-chance v3 scoring
still works before the 30 September shutdown. Removing both is a follow-up after that date.

Relevant external documentation: [Pangram 4 migration guide](https://www.pangram.com/blog/pangram-4-migration-guide),
[Pangram REST API docs](https://pangram.readthedocs.io/en/stable/api/rest.html).

## AI usage

Both commits were written by Claude Code (Opus 5) under my direction, and it drafted both
commit messages and this PR description. Each commit carries a
`Co-Authored-By: Claude Opus 5` trailer and a `(Commit message written by Claude Code.)`
marker, and the commit bodies have `## Process` sections with more detail.

What the AI did: wrote `pangram_cutover.py` and tested it against a synthetic export before
the real one existed; implemented the cutover against a specification I had settled earlier;
wrote the new specs; and drafted the prose. What I contributed: the operator decisions that
constrain the cutover (keep the 0.9 rule, no AI-Assisted or humanizer alerting, leave email
copy alone — all decided 3 September, before any of this code), the Phase 1 data collection
runs in the production console, the design in the migration plan and runbook, the decision
to pull the cutover forward from the planned 15–19 September window, and review.

Two changes here came from the AI reading the code rather than from my specification, and
both deserve a reviewer's attention: the `[nil, COURSE_UPDATE_ORIGIN]` treatment of legacy
rows (my runbook said `COURSE_UPDATE_ORIGIN` alone, which would have re-billed pre-March
revisions), and the `BuildAiDetectionSampleFromRecentScores` default. The classification of
`TaskFailed` as terminal rather than retryable is a judgment call, not something the data
settles — see Open questions.

Scale of effort: two commits inside a single session on 9 September, about 17 minutes apart.
That is a small amount of human time on the *code*, and it should be read against a much
larger amount of prior human-directed work: the migration plan and runbook (3–4 September),
Phase 0 in #7034, and the Phase 1 scoring runs and analysis (5 September), which is where
the design decisions and the evidence actually come from.

The "What this PR does" summary and the analysis above were also drafted by Claude Code and
may contain errors. This PR description was drafted using Claude Code and the `prepare-pr`
skill.

## Screenshots

No UI changes. The alert emails render identically; only the sample data behind the mailer
previews changed, and the preview pages at `/mailer_previews` are the place to see that.

- **`TaskFailed` is treated as terminal.** A task the API itself finished in a failed stage
  is not retried, on the reasoning that resubmission is unlikely to succeed and every
  attempt bills. If `STAGE_FAILED` turns out to be transient in practice, this should move
  to the retryable list. We have no production v4 failure data yet either way.
- **`retry: 3`** is a guess at "bounded". The two LTI workers in this codebase use `retry: 5`.
- **Email copy is now known to be inaccurate.** Phase 1 found the median v3 window is 256
  words and v4 usually returns one window for the whole edit, so the emails' "Number of
  ~400-word segments with AI text" and "segments of approximately 400 words" no longer
  describe either model. This is deliberately out of scope — the copy is operator-written —
  but it should be revised soon.
- **Parked for after cutover, with Phase 1 data in hand:** under v4, `max window > 0.9` is a
  whole-text score for 79% of units and 27% of documents Pangram labels AI-generated fall
  below it, so whether to alert on the label, a lower max, or `windows_above_0_9` is a real
  rule decision. Humanizer flagging (13% of units, none in the human baseline) is also
  parked.
- **Pangram silently changed the v3 model in May 2026** (3.2 → 3.3.2 on 18 May, found during
  Phase 1). Expect the same from v4 point releases and watch the recorded `model_version`.
- **Follow-up after 30 September:** remove `PANGRAM_V3_KEY` from the registry and delete
  `PangramApi.v3`, keeping the constant for historical rows.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,5 @@ stats.json
 # Wikipedia diff URLs with AI-detection scores, and this repo is public; the
 # scripts and their READMEs are committed, the data they read and write is not.
 docs/analytics_scripts/**/*.csv
+# pangram_cutover.py imports analyze.py, so running it compiles a cache there.
+docs/analytics_scripts/**/__pycache__/
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,8 @@ stats.json
 !/fixtures/vcr_cassettes/cached/*
 /backups/
 .ltiaas
+
+# Research data exports under docs/analytics_scripts stay local. Rows pair public
+# Wikipedia diff URLs with AI-detection scores, and this repo is public; the
+# scripts and their READMEs are committed, the data they read and write is not.
+docs/analytics_scripts/**/*.csv
--- a/app/controllers/revision_ai_scores_stats_controller.rb
+++ b/app/controllers/revision_ai_scores_stats_controller.rb
@@ -29,12 +29,14 @@ class RevisionAiScoresStatsController < ApplicationController
     @scores_with_likelihood = @scores.where.not(avg_ai_likelihood: nil)
   end
 
+  # A failed check is stored as a row with nil likelihoods, so the distributions
+  # read the scored rows rather than every row.
   def set_avg_likelihoods
-    @avg_likelihoods = @scores.map { |s| { value: s.avg_ai_likelihood } }
+    @avg_likelihoods = @scores_with_likelihood.map { |s| { value: s.avg_ai_likelihood } }
   end
 
   def set_max_likelihoods
-    @max_likelihoods = @scores.map { |s| { value: s.max_ai_likelihood } }
+    @max_likelihoods = @scores_with_likelihood.map { |s| { value: s.max_ai_likelihood } }
   end
 
   # Sets an array of hashes with date, namespace, and count for historical scores.
@@ -86,6 +88,10 @@ class RevisionAiScoresStatsController < ApplicationController
   # Given arrays of possible namespaces and bins, and a partial array of hashes, returns a complete
   # array of hashes that includes all dates, namespaces and bins based on the partial one.
   def complete_hash(namespaces, bins, partial_stats)
+    # No rows means no date range to fill in. Reachable when every production row
+    # is a failed check, which carries no likelihood to bin.
+    return [] if partial_stats.empty?
+
     start_date = partial_stats.keys.map(&:first).min
     end_date   = partial_stats.keys.map(&:first).max
 
--- a/app/mailer_previews/ai_edit_alert_mailer_preview.rb
+++ b/app/mailer_previews/ai_edit_alert_mailer_preview.rb
@@ -64,17 +64,19 @@ class AiEditAlertMailerPreview < ActionMailer::Preview
 
   def example_alert_details(title)
     { article_title: title,
-      pangram_prediction: 'We are confident that this document is fully AI-generated',
-      headline_result: 'Fully AI Generated',
-      average_ai_likelihood: 0.974378,
-      max_ai_likelihood: 1.0,
+      pangram_prediction: 'We believe that this entire text is AI.',
+      headline_result: 'AI Generated',
+      average_ai_likelihood: 0.99999607,
+      max_ai_likelihood: 0.99999607,
       fraction_human_content: 0.0,
       fraction_ai_content: 1.0,
       fraction_mixed_content: 0.0,
-      window_likelihoods: [1.0, 0.9982278487261604, 0.9959831237792969],
-      predicted_ai_window_count: 3,
+      window_likelihoods: [0.99999607],
+      predicted_ai_window_count: 1,
       pangram_share_link: 'https://www.pangram.com/history/88ba317f-8572-4000-911f-2aa8fbea68fa',
-      pangram_version: '3.0',
+      pangram_version: '4.0',
+      humanized_window_count: 0,
+      max_humanizer_score: 0.00444,
       prior_alert_count_for_course: 0 }
   end
 
--- a/app/mailer_previews/scholars_ai_edit_alert_mailer_preview.rb
+++ b/app/mailer_previews/scholars_ai_edit_alert_mailer_preview.rb
@@ -31,17 +31,19 @@ class ScholarsAiEditAlertMailerPreview < ActionMailer::Preview
 
   def example_alert_details(title)
     { article_title: title,
-      pangram_prediction: 'We are confident that this document is fully AI-generated',
-      headline_result: 'Fully AI Generated',
-      average_ai_likelihood: 0.974378,
-      max_ai_likelihood: 1.0,
+      pangram_prediction: 'We believe that this entire text is AI.',
+      headline_result: 'AI Generated',
+      average_ai_likelihood: 0.99999607,
+      max_ai_likelihood: 0.99999607,
       fraction_human_content: 0.0,
       fraction_ai_content: 1.0,
       fraction_mixed_content: 0.0,
-      window_likelihoods: [1.0, 0.9982278487261604, 0.9959831237792969],
-      predicted_ai_window_count: 3,
+      window_likelihoods: [0.99999607],
+      predicted_ai_window_count: 1,
       pangram_share_link: 'https://www.pangram.com/history/88ba317f-8572-4000-911f-2aa8fbea68fa',
-      pangram_version: '3.0',
+      pangram_version: '4.0',
+      humanized_window_count: 0,
+      max_humanizer_score: 0.00444,
       prior_alert_count_for_course: 0 }
   end
 end
--- a/app/services/build_ai_detection_sample_from_recent_scores.rb
+++ b/app/services/build_ai_detection_sample_from_recent_scores.rb
@@ -8,8 +8,10 @@
 class BuildAiDetectionSampleFromRecentScores < BuildAiDetectionSample
   BANDS = { 'low' => 0.0...0.5, 'mid' => 0.5...0.9, 'high' => 0.9..1.0 }.freeze
 
+  # check_type defaults to whichever detector production alerting currently uses,
+  # so the default follows a cutover instead of silently sampling the old one.
   def initialize(sample_name:, per_band: 40, since: 30.days.ago,
-                 check_type: RevisionAiScore::PANGRAM_V3_KEY, verbose: false)
+                 check_type: CheckRevisionWithPangram::DETECTOR_KEY, verbose: false)
     super(sample_name:, verbose:)
     BANDS.each do |band, range|
       candidates(range, since, check_type).limit(per_band).each { |score| add_score(score, band) }
--- a/app/services/check_revision_with_pangram.rb
+++ b/app/services/check_revision_with_pangram.rb
@@ -26,7 +26,7 @@ class CheckRevisionWithPangram
     return unless fetch_pangram_inference
 
     parse_pangram_response
-    create_revision_ai_score
+    record_successful_check
 
     generate_alert if ai_likely?
   end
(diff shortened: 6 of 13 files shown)


ragesoss (commented):
A note on how this feedback is meant to be used: it comes out of an AI-assisted review session and should not be automatically trusted or acted on. It is worth reading and understanding, then using or ignoring based on the developer's own judgment. Anything asserted below may be wrong.

Two findings were reproduced locally with a throwaway spec; two smaller notes were not.

### `terminal_failure?` classifies 429 and 408 as terminal

`app/services/check_revision_with_pangram.rb:85` treats every 4xx `RequestError` as unretryable:

```ruby
when PangramApi::RequestError then error.status.to_i.between?(400, 499)
```

429 Too Many Requests is the canonical retryable 4xx, and 408 Request Timeout is another. The v4 async endpoint makes two or more HTTP calls per check (POST `/task`, then polling GETs) where v3 made one, so a course update that queues a burst of checks has more rate-limit surface than v3 did.

The consequence appears larger than a delayed alert. `RevisionScanner.schedule_revision_checks` is only reached from `UpdateCourseWikiTimeslices#update_timeslices`, over `revisions[:revisions]` for the timeslices being processed. A revision whose check was silently dropped is not re-offered on the next course update. The nil-likelihood row keeps it *eligible* for a re-check, but nothing schedules one — so in practice the check and its alert are lost until that timeslice is reset.

A narrower terminal set — 400, 401, 402, 403, 404 — would keep the out-of-credit case the change was built for while letting 429 and 408 consume a bounded retry.

### Swallowing 402 removes the only signal that alerting has stopped

Related to the above rather than separate. Exhausted credits are now recorded to a database row and nothing else; there is no `Sentry` capture anywhere in this path. Before this change, credit exhaustion would accumulate jobs in the Sidekiq dead set, which is visible on the Sidekiq dashboard. After it, production AI alerting can stop entirely with no operator-facing signal. If 402 stays terminal, something should probably notice a run of `RevisionAiScore` rows carrying `details['error']`.

### A transient failure followed by a success leaves two rows for one revision

`record_failed_check` (`:93`) writes its row before the re-raise at `:76`, so it also fires for retryable errors. `create_revision_ai_score` (`:133`) then calls `RevisionAiScore.create` on a fresh key rather than updating the row `record_failed_check` created.

Reproduced: raise `TaskTimeout`, then succeed on the retry, and `RevisionAiScore.where(revision_id: …).count` is 2 — one nil-likelihood failure row plus the real one. The PR's own example (`leaves the revision open to a later successful check`) asserts only `where.not(avg_ai_likelihood: nil).count == 1`, which passes in both worlds.

This reaches the stats page. `RevisionAiScoresStatsController#set_scores` scopes `@scores` to `check_origin: [nil, 'course_update']`, and `set_historical_scores_by_namespaces` counts all of them — so "scores by namespace" would begin counting failed attempts alongside real checks, and `set_avg_likelihoods` / `set_max_likelihoods` would gain nil entries. Having `create_revision_ai_score` use the same `find_or_initialize_by` key as `record_failed_check` would collapse the pair.

### Smaller notes

- `record_failed_check` uses `score.update` (`:98`) rather than `update!`. `ScoreAiDetectionSample#save_score` uses `update!` in the same find-or-initialize-then-write shape. A validation failure here returns false silently and records nothing, in the one case where recording is the entire point.
- `parse_population` in `pangram_cutover.py` merges partial `--population` input over `DEFAULT_POPULATION`, so `--population high=1305` alone silently combines a fresh count with the hardcoded 2026-09-04 preflight counts for `low` and `mid`. For a script whose headline output is a reweighted rate, either requiring all three bands or echoing the counts actually used would make that visible.
- `pangram_cutover.py` is the first script here to `import analyze`, so running it creates `docs/analytics_scripts/.../analysis/__pycache__/`, which is not ignored. This PR is already adding a gitignore rule for that tree.

### Checked and found clean

- Every `details[:...]` reader in `app/models/alert_types/ai_edit_alert.rb` maps to a key the v4 parser still populates; the email templates read only `predicted_ai_window_count`, so the two new humanizer keys are inert in the copy.
- Nil-likelihood rows are already handled by `AiDetectionSample#scored_by?`, `ScoreAiDetectionSample`, the stats controller's binning (which uses `@scores_with_likelihood`), and `ExportAiDetectionComparison`, which already reads `details['message']` — the exact key `record_failed_check` writes.
- The `docs/analytics_scripts/**/*.csv` rule matches at both depth 0 and depth 2 (`git check-ignore -v`), and no currently-tracked CSV is affected.
- RuboCop clean on all seven changed Ruby files. 87 examples across the PR's specs and the surrounding AI-detection specs pass with no failures.
- `yarn build` fails locally with 13 `import/named` errors on `query-string`, but fails identically on `master`. Pre-existing and unrelated to this PR.

*Drafted in a Claude Code session (~15 min, 2 interactions). Sage asked for a review of the PR and for this comment, and read the findings before posting, but did not independently verify them. The agent ran RuboCop and the surrounding spec files locally, reproduced both main findings with a throwaway AI-written spec (`TaskTimeout`-then-success leaving two rows; a 429 being swallowed without a raise), and compared the `yarn build` failure against master. That spec was written by the agent and run but not reviewed line-by-line by Sage, and is not included in the PR. The reasoning about `RevisionScanner` not re-offering dropped revisions was read out of the code, not demonstrated.*

ragesoss (commented):
Same note as before on how this is meant to be used: it comes out of an AI-assisted review session and should not be automatically trusted or acted on. Worth reading and understanding, then using or ignoring on your own judgment. Anything asserted below may be wrong.

`1bea425` addresses all six findings. Checked in their current state: the 408/429 denylist (the reasoning for preferring it to the proposed 400–404 allowlist holds — an allowlist would make 413/415/422 burn three attempts before being dropped anyway), the Sentry warning on terminal failures, the single `production_score` find-or-initialize collapsing the two-row case, `update!` on both paths, the `[default]` tagging in `parse_population`, and the `__pycache__` rule. The new `RevisionAiScore.count == 1` assertion is the one the previous spec was missing.

The `complete_hash` 500 found alongside them was verified as pre-existing: reverting only that file to `df4d6ed` and rerunning the new request spec fails 2 of 3 examples with `TypeError (can't iterate from NilClass)`. It also fires in the failed-checks-only state, not only on an empty table.

One new issue in the delta, and two notes.

### Retrying a failure that happened after submission re-bills the check

The 408/429 reclassification is right for a failure on the submitting POST, where nothing was billed and retrying is free. But `PangramApi#poll_task` runs `parse_response` on every polling GET (`lib/pangram_api.rb:99`), so a 429 *after* the task was accepted and billed escapes as the same retryable `RequestError`. Sidekiq re-runs `CheckRevisionWithPangram` from the top, and the service always starts at POST — the accepted task is abandoned and its bill paid for nothing.

Polling is the higher-volume request pattern in this flow: a GET every `POLL_INTERVAL` (0.5s) for up to 60s per job, against one POST. Across a course update's burst of concurrent workers, a 429 seems likelier on a poll than on a submission, which is the same rate-limit-surface argument the commit message makes for reclassifying 429 in the first place.

Reproduced with a throwaway spec — POST stubbed to succeed, polling GET to 429:

```ruby
before do
  stub_const('PangramApi::POLL_INTERVAL', 0)
  plaintext = instance_double(GetRevisionPlaintext, article_title: 'Five-a-side football',
                                                    plain_text: 'x ' * 400)
  allow(GetRevisionPlaintext).to receive(:new).and_return(plaintext)
end

it 'bills a POST for every attempt when the polling GET is rate limited' do
  submissions = stub_request(:post, PangramApi::V4_API_URL)
                .to_return(status: 200, body: { task_id: }.to_json)
  stub_request(:get, task_url).to_return(status: 429, body: 'Too Many Requests')

  # Sidekiq re-runs the whole service on retry; AiDetectionWorker allows 3 retries.
  4.times do
    expect { CheckRevisionWithPangram.new(attrs) }
      .to raise_error(PangramApi::RequestError, /HTTP 429/)
  end

  expect(submissions).to have_been_requested.times(4)
  expect(RevisionAiScore.count).to eq(1)
  expect(RevisionAiScore.last.avg_ai_likelihood).to be_nil
end

it 'billed only one POST for the same failure before the 429 reclassification' do
  allow_any_instance_of(CheckRevisionWithPangram)
    .to receive(:terminal_failure?).and_return(true)
  submissions = stub_request(:post, PangramApi::V4_API_URL)
                .to_return(status: 200, body: { task_id: }.to_json)
  stub_request(:get, task_url).to_return(status: 429, body: 'Too Many Requests')

  CheckRevisionWithPangram.new(attrs)

  expect(submissions).to have_been_requested.times(1)
end
```

Four submissions billed for one revision where the previous behaviour billed one. The same is true of a 5xx on a poll, but that was already retryable before this commit, so only the 4xx half is new here.

`poll_task` already owns a 60s deadline it could keep polling within, so absorbing a transient GET failure there — rather than letting it unwind to Sidekiq — would get the alert for one bill instead of up to four. It also gives the `retry: 3` open question a cost dimension that wasn't there before: at `retry: 5` the same rate-limited poll would be up to six submissions per revision.

### Smaller notes

- `set_historical_scores_by_namespaces` still reads `@scores` rather than `@scores_with_likelihood`, so a permanently failed check counts as a daily check. That reads as deliberate and is defensible for a chart labelled "daily checks", but it does mean the three charts on the page now have different denominators, and the namespace line will sit above the binned ones whenever failures accumulate. Nothing in the specs pins that difference down either way — the new request spec asserts `historical_scores_by_namespace` only for the empty case.
- Correcting the previous comment: `yarn build` succeeds on this branch (exit 0, three warnings). The 13 `import/named` errors on `query-string` reported last time were a stale local `node_modules` condition, not a property of the tree. The conclusion — unrelated to this PR — was right, but the failure itself was not real.

Verified in passing: the empty and failed-check-only states both render every section of the stats page in a real browser, so the `[]` return from `complete_hash` is safe through Vega. 99 examples green across the PR's specs and the surrounding AI-detection specs, three related feature spec files green, RuboCop clean on all nine changed Ruby files, and both Python scripts compile with the single `parse_population` caller updated.

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage asked for a re-review and chose how the new finding should be framed, then read this comment before posting, but did not independently verify any of it. The agent ran RuboCop, 99 rspec examples and three feature spec files locally; confirmed the `complete_hash` 500 is pre-existing by reverting that one file and rerunning the new request spec; reproduced the re-billing finding with the throwaway spec above; and ran `yarn build` to check the previous comment's claim about it. The spec pasted above was written by the agent and run but not reviewed line-by-line by Sage, and is not included in the PR. The claim about which HTTP verb is likelier to be rate limited is reasoning from `POLL_INTERVAL` and the code path, not measurement.*

PR #7056, opened 2026-09-10
Refuse to delete or rename event-linked courses; tolerate sync races

ragesoss:
## What this PR does

Two fixes for the Programs & Events Dashboard's integration with Wikimedia Event Registration, prompted by [T437639](https://phabricator.wikimedia.org/T437639), where participants registering for events linked to the Dashboard were blocked with "not connected to this event" and "unknown error" messages.

**Refuse to delete or rename an event-linked course.** The CampaignEvents extension looks the course up by slug on every registration and has no way to learn that the course went away. When an organizer deletes a linked course and recreates it under the same slug, the event ends up pointing at a course that was never linked, and every registration fails the extension's dry-run validation. That is exactly what happened to event 4296: the linked course was deleted on 2026-08-14 at 07:47 UTC and recreated three minutes later. `CoursesController#destroy` and `#update` now return 409 with a message when the course has an `event_sync` flag and the action would delete it or change its slug. The Delete button on the course page renders disabled with the same message as its tooltip. Updates that leave the slug alone still go through.

**Tolerate overlapping participant syncs.** The extension sends the full roster after every registration. Two overlapping syncs for the same course both pass the already-enrolled check and the model's uniqueness validation before either commits, and the loser 500s on the database's unique index (Sentry PEONY-2QH, four occurrences since 2026-08-25). `JoinCourse` now rescues `ActiveRecord::RecordNotUnique` and reports it as the existing `cannot_join_twice` failure, so the request completes and the remaining participants are still processed.

Relevant extension code: [WikiEduDashboard.php](https://github.com/wikimedia/mediawiki-extensions-CampaignEvents/blob/master/src/TrackingTool/Tool/WikiEduDashboard.php) in the CampaignEvents extension, which maps the Dashboard's `sync_not_enabled` and `course_not_found` codes to the messages participants see.

## AI usage

This PR was produced with Claude Code (Claude Fable 5.1) under human direction. The agent did the investigation behind it: it mapped the extension's error messages to Dashboard error codes by reading the CampaignEvents source, scanned recent Event Registration events through the CampaignEvents REST API and compared each linked course's stored event ID with the real one, and used Sentry's course-deletion log messages to reconstruct the delete-and-recreate sequence and to find the duplicate-key crash. It then wrote the code, the specs, and the commit message. The human decided what to fix, chose to include the rename guard, supplied both user-facing strings verbatim, and reviewed the diff.

One correction during the work: the first version of the race specs passed without exercising the rescue, because the model's uniqueness validation stopped the duplicate before the database did. The agent found this with a `rails runner` reproduction and adjusted the specs to stub the validation so that only the unique index catches the duplicate, matching production.

Scale of effort: a single commit, written in one session of roughly an hour of implementation after about an hour and a half of investigation. The "What this PR does" section and the rest of this description were also drafted by Claude Code and may contain errors.

## Screenshots

The Actions panel on an event-linked course, viewed by a facilitator. The tooltip on the disabled button is a native `title` attribute and does not appear in the screenshot.

| Before | After |
|---|---|
| before | after |

- The rename guard is a backstop. The details panel already locks title, institution, and term for linked courses, so it should only fire on a stale page or a direct API call.
- The race fix makes each insert safe but does not serialize syncs per course, so two interleaved syncs can still apply adds and removes in either order. Each sync carries the full roster, so the next one converges. A per-course lock felt like more machinery than the observed problem warrants.
- The Dashboard still has no way to tell the extension that a link is gone. A linked course that was deleted before this change leaves its event blocked with a "does not exist" error until the organizer unlinks it in Event Registration. I have raised on the Phabricator task whether the extension should treat that case as a desync rather than a hard failure.
- The screenshot spec (`spec/features/event_sync_delete_button_screenshots_spec.rb`) is disposable and not committed.

### DIFF
--- a/app/assets/javascripts/components/overview/available_actions.jsx
+++ b/app/assets/javascripts/components/overview/available_actions.jsx
@@ -113,9 +113,18 @@ const AvailableActions = ({ course, current_user, updateCourse, courseCreationNo
     }
     // If course is not published, show the 'delete' button to instructors and admins.
     if ((user.isAdvancedRole || user.admin) && (!course.published || !Features.wikiEd)) {
+      // Deleting a course that is linked to a Wikimedia Event Registration event
+      // would break registration for that event, so the server refuses it.
+      const linkedToEvent = Boolean(course.flags.event_sync);
+      let deleteTitle;
+      if (linkedToEvent) {
+        deleteTitle = I18n.t('courses.error.event_sync_delete');
+      } else if (Features.wikiEd) {
+        deleteTitle = I18n.t('courses.delete_course_instructions');
+      }
       controls.push((
-        <div title={Features.wikiEd ? I18n.t('courses.delete_course_instructions') : undefined} key="delete" className="available-action">
-          <button className="button danger" onClick={deleteCourseFunc}>
+        <div title={deleteTitle} key="delete" className="available-action">
+          <button className={linkedToEvent ? 'button danger disabled' : 'button danger'} onClick={deleteCourseFunc} disabled={linkedToEvent}>
             {CourseUtils.i18n('delete_course', course.string_prefix)}
           </button>
         </div>
--- a/app/controllers/courses/delete_from_campaign_controller.rb
+++ b/app/controllers/courses/delete_from_campaign_controller.rb
@@ -8,6 +8,9 @@ class Courses::DeleteFromCampaignController < CoursesController
     if @course.campaigns.size > 1
       remove_course_from_campaign_but_not_deleted
     else
+      # Removing the only campaign deletes the course, which is refused for
+      # event-linked courses. See CoursesController#reject_deletion_of_event_synced_course.
+      reject_deletion_of_event_synced_course { return }
       remove_and_delete_course_from_campaign
     end
   end
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -46,8 +46,9 @@ class CoursesController < ApplicationController
 
   def update
     validate
-    handle_course_announcement(@course.instructors.first)
     slug_from_params if should_set_slug?
+    reject_rename_of_event_synced_course { return }
+    handle_course_announcement(@course.instructors.first)
     @course.update update_params
     update_courses_wikis
     update_course_wiki_namespaces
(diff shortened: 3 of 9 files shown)


ragesoss (commented):
A note on how to use this: the findings below came from an AI-driven review session and are provisional. They are worth reading and checking, not acting on blindly.

**Finding, since fixed in bc21c3bd4: the deletion guard missed the endpoint the P&E Delete button actually uses.** `deleteCourseFunc` in `available_actions.jsx` dispatches `removeAndDeleteCourse` whenever `course.campaigns` is non-empty, and `CourseCreationManager#set_initial_campaign` falls back to the default campaign, so on P&E nearly every course takes that branch. The request lands in `Courses::DeleteFromCampaignController#remove_and_delete_course_from_campaign`, which called `DeleteCourseWorker.schedule_deletion` with no `controlled_by_event_center?` check. The disabled button covered the normal UI, but a stale tab or a direct request would still have deleted a linked course. An exploratory request spec against the first commit confirmed it: `schedule_deletion` expected 0 times, received 1 time.

The second commit applies the inherited `reject_deletion_of_event_synced_course` guard in that controller before the deletion branch, so a refused request leaves the course in its campaign, and removal from one of several campaigns is still allowed since that never deletes the course. It adds the first spec coverage for that controller (five examples). The exploratory spec was rerun against the fix and passed on all seven examples before being discarded.

**Verified as working on the first commit:** the rename guard with the payload the details panel sends, both when the name is unchanged and when title or term changes; the reorder of `handle_course_announcement`, which reads only the submitted flag and is Wiki Ed only; the `RecordNotUnique` rescue against the real unique index on course, user, and role; the 409 `message` body reaching the notifications reducer on both actions; and the React button state in a Jest render for linked, unlinked, and Wiki Ed cases, including a string-valued `event_sync`.

**Minor:** a disabled button is skipped by keyboard focus, so the reason in the wrapper `title` is unreachable for screen-reader users. Visible text or `aria-disabled` with `aria-describedby` would fix that. P&E only.

**PR description:** the "Refuse to delete or rename" paragraph names only `CoursesController#destroy` and `#update`; it should also mention the campaign-removal endpoint now that it is guarded.

CI on the second commit was still running when this was posted.

*Drafted in a Claude Code session (~25 min, 4 interactions). Sage directed the review, asked for the fix after reading the findings, and approved posting this updated comment. The agent wrote and ran the exploratory specs and wrote the fix commit and its spec; Sage read the agent's summary of the fix and its diff but did not independently rerun the specs.*

PR #7062, opened 2026-09-11
Put new courses into acuwt update

gabina:
## What this PR does

Starts the gradual migration of all courses onto the ACUWT update path by making
`use_acuwt` the default for newly created courses. Existing courses are untouched
and stay on the legacy path, so the migration proceeds as new courses appear
rather than through a bulk flip.

Three commits:

1. **Retire the `acuwt_update` queue.** `CourseQueueSorting#queue_for` checked
   `use_acuwt?` before every other rule, so defaulting the flag would have routed
   every new course — regardless of size — into a single dedicated queue. ACUWT
   courses now queue by the same timing-based rules as everything else. Removes
   the queue from `sidekiq.yml`, its latency threshold and locale strings from the
   system status page, its dedicated systemd unit, and the
   `CourseDataUpdateWorker` exception that kept `very_long_update` ACUWT courses
   running.
2. **Use ACUWT by default for new courses.** All three creation paths set the
   flag: the new-course form (via `update_use_acuwt` in the post-creation
   updates), `CourseCloneManager` (set on the clone regardless of the source
   course), and `CopyCourse` (forced on rather than inherited from the source
   dashboard's flags hash).
3. **Update flags in a single save query.** Unrelated cleanup made obvious by the
   commit above: the controller's flag setters each ran their own `@course.save`,
   so course creation did four saves and a course update did six. The setters now
   only mutate the in-memory flags hash, and `update_flags` and
   `handle_post_course_creation_updates` each persist once.

## AI usage

Claude Code (Opus 5) wrote the code and specs in all three commits, working from
the user's direction: the user set the migration strategy, decided to retire the
queue, chose to force the flag on in `CopyCourse`, decided against gating the
change per-deployment, moved the flag assignment out of `CourseCreationManager`
into `handle_post_course_creation_updates`, and identified the redundant per-flag
saves. Claude Code also drafted this description, using the `prepare-pr` skill, as
well as the "What this PR does" summary above — which may contain errors.

All three commits were made within a single session on 2026-09-11, roughly three
hours from first to last, including the user testing all three creation paths
locally.

## Screenshots

No UI changes.

- The deploy needs a manual step on peony: the running `sidekiq-acuwt` systemd
  unit must be stopped and disabled. Removing the unit file from the repo does not
  touch the live service.
- The existing ACUWT courses on the P&E Dashboard are test courses also flagged
  `very_long_update`. With the dedicated queue gone they fall into the
  `very_long_update` queue, which is intentionally never processed, so they stop
  updating. That is intended for now.
- What to do with the long-running courses still on the legacy path is
  deliberately out of scope; it depends on how the new ACUWT courses behave.
- `use_acuwt` is not in `Course::FEATURE_FLAG_KEYS`, so `CopyCourse` would have
  received it as a string key and `use_acuwt?` (which reads the symbol) would have
  missed it. `force_use_acuwt` deletes the string key before setting the symbol
  one, which sidesteps the issue for this path, but the key list may be worth
  revisiting separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gabina:
After deploying PR #7038, I run some updates on Coordinate Me courses - course [37575 acuwt](https://outreachdashboard.wmflabs.org/courses/Test/TEST_Coordinate_Me_2026_NL_ACUWT/home) and course [37577 default](https://outreachdashboard.wmflabs.org/courses/Test/TEST_Coordinate_Me_2026_NL_DEFAULT_(August_2028)/home).


Removing user Andre Engels from ACUWT version took ~45 minutes compared to ~12 hours in the default version.
Re-adding user Andre Engels from ACUWT version took ~55 minutes compared to ~12 hours in the default version.

Removing a small category from ACUWT version took ~3 minutes compared to ~7 minutes in the default version.
Re-adding a small category from ACUWT version took ~4 minutes compared to ~8 minutes in the default version.

ragesoss (approved):
A note on how to read this: the findings below came out of an AI-assisted review session. They are meant to be read and weighed, not automatically trusted or acted on.

Verified locally on the branch merged with current master: `yarn build`; RuboCop on the twelve changed Ruby files; the PR's own specs together with the existing specs for the courses controller, course cloning, copy course, queue sorting, schedule course updates, system metrics, the update worker, and `UpdateCourseStats` (157 examples); and eight feature specs covering course creation, cloning, copying, the course-flags admin page, and the system status page (26 examples). All green apart from one order-dependent failure in the controller spec's destroy example (a hardcoded user id colliding with a leftover row), which passes in isolation and is unrelated to this PR.

Two exploratory specs were also written and run:

- A full `UpdateCourseStats` run on a non-scoped `ClassroomProgramCourse` with `use_acuwt` set, against the same recorded data as the legacy `course_update` examples. This is the state the PR makes the default, and as far as could be determined it is the first time the ACUWT branches of `ArticleStatusManager` and `ArticleNamespacesManager` are exercised end to end on a non-scoped course: every production ACUWT course so far has been article-scoped, where `should_update_article_status?` returns false. `ArticleStatusManager` ran, the ACUWT and legacy paths produced identical course, article-course, and course-user caches, no timeslice was left `needs_update` or `needs_reaggregation`, and a second update was idempotent.
- UPDATE-statement counts for the create and update actions: the flags column is written exactly once per request, and the created course carries `use_acuwt` as a symbol key next to `timeslice_duration`.

The code changes look correct. The remaining points are operational.

## Deploy steps beyond the description

1. **Clear the orphaned `acuwt_update` queue on both deployments** after the deploy, for example `Sidekiq::Queue.new('acuwt_update').clear` from a console. Removing the queue from `sidekiq.yml` and the unit file does not touch jobs already in Redis. `CourseDataUpdateWorker` uses `lock: :until_and_while_executing` with a 30-day TTL, so a job left in that queue appears to hold the unique lock for its course id and would block the scheduler's re-enqueue of that course into any queue until the TTL expires or the queue is cleared (the orphan reaper only releases digests whose job is no longer present in any queue). `clear` also removes the queue from Sidekiq's queue set.
2. **Check the Wiki Ed deployment for existing `use_acuwt` courses.** The description covers the P&E test courses (ACUWT plus `very_long_update`, which now park in `very_long_update` as intended). Wiki Ed never ran a `sidekiq-acuwt` process, so any course flagged there has been idle in `acuwt_update`; after this change it routes by the normal timing rules unless it is also `very_long_update`. A console check:
   ```ruby
   Course.where("flags LIKE '%use_acuwt: true%'").map { |c| [c.id, c.slug, c.type, c.very_long_update?, c.only_scoped_articles_course?] }
   ```
3. Stop and disable `sidekiq-acuwt` on the peony host that runs it (already in the description).

## Rollout observation

With no per-deployment gate, every new course on both dashboards takes the ACUWT path from the next deploy, including the Wiki Ed fall term cohort. The exploratory spec above covers the correctness question that could be tested locally; production scale is the remaining test. The Sentry `Article retracked` and `Article namespace changed` info messages that the ACUWT branches emit should give an early signal on the first cohort.

One carry-over from the PR #6971 review becomes reachable now that `ArticleStatusManager#handle_undeletion` runs on ACUWT courses: `ArticlesCoursesCleaner#reset_specific_articles` deletes the `articles_courses` row even when the article has no ACUWT rows and the repair half returns early. For courses that start on ACUWT the rows exist from the first update, so this stays a corner case; noted for completeness, not as a blocker.

## Spec offered for adoption

The pipeline spec mirrors the three cache examples in the `when there are revisions` context of `spec/services/update_course_stats_spec.rb`, so the smallest form is a nested context with `let(:flags) { { use_acuwt: true } }` sharing those examples. The ACUWT examples (including the second-update run) recorded about 70 KB of new interactions on top of the `course_update` cassette, so the first run needs network access. The extra assertions used:

```ruby
context 'when use_acuwt is set' do
  let(:flags) { { use_acuwt: true } }

  # ...the same three cache examples as the legacy context...

  it 'runs article status sync for a non-scoped course' do
    expect(ArticleStatusManager).to receive(:update_article_status_for_course)
      .with(course).and_call_original
    VCR.use_cassette('course_update') { subject }
  end

  it 'leaves no timeslices pending after the update' do
    VCR.use_cassette('course_update') { subject }
    expect(course.course_wiki_timeslices.where(needs_update: true)).to be_empty
    expect(course.course_wiki_timeslices.where(needs_reaggregation: true)).to be_empty
    expect(ArticleCourseUserWikiTimeslice.where(course:, needs_update: true)).to be_empty
  end
end
```

Approved.

---

*Drafted in a Claude Code session (~40 min, 3 interactions). Sage asked for the review, read the triage, diff walkthrough and wrap-up summaries, and chose to approve, but did not independently rerun the checks. The verification described above was done by Claude Code: branch checked out and merged with master, build, RuboCop, the baseline and feature specs run, and the two exploratory specs written and run. The spec code in this comment is AI-generated and was run locally (all examples pass); Sage did not review or run it.*

PR #7065, opened 2026-09-11
Refresh the HECVAT, document the Cloud Firewall, and harden Oj and CI

ragesoss:
## What this PR does

Refreshes the published HECVAT (`docs/hecvat.md`, served at `/hecvat`) after the production server's in-place move from Debian 11 to Debian 12 on 2026-09-04, and closes several of its "No" answers with low-disruption changes made the same day:

- **Cloud Firewall.** `server_config/firewall.md` documents the Akamai (Linode) Cloud Firewall now attached to production, staging, and the rehearsal clone: why a network firewall rather than host nftables first, the facts the rules depend on (notably that the Wikimedia-allowlisted IPv6 is a SLAAC address that needs inbound ICMPv6), the rule set (inbound drop; allow TCP 22, TCP 80+443, ICMP from all IPv4 and IPv6; outbound accept), the four-phase rollout runbook with each phase's recorded result, and the change procedure. HECVAT FIDP-01, FIDP-02 and FIDP-05 change from No to Yes and point at it. The Phase 0 listener check also found that the bookworm upgrade had enabled `mariadb.socket`, leaving MariaDB on `*:3306` reachable from the internet (all accounts were `@localhost`, so no remote login was possible); that was fixed on the server the same day and the diagnosis and post-upgrade re-check are written into the runbook.
- **Oj strict mode.** `Oj.load` ran in Oj's default `:object` mode, which instantiates arbitrary Ruby classes from `"^o"` keys, and one call site parses training-content JSON from editable Meta-Wiki pages. `config/initializers/oj.rb` now sets `mode: :strict`; every call site only ever needed plain Hash/Array output.
- **Dependency audit in CI.** `bundler-audit` runs right after Ruby setup, before the test suites, and fails the build on any gem with a published advisory. Making it green required `crass` 1.0.6 → 1.0.7 (four CSS-parsing DoS advisories) and `rubyzip` 2.3.2 → 3.6.0 (CVE-2026-85396, path traversal on extraction; the Dashboard only writes zips).
- **HECVAT facts.** APPL-03 (OS and runtimes), DOCU-02 (owner and testing cadence), ITAC-06 (VPAT scope), APPL-06 and VULN-01 (the CI audit), DATA-01 (firewall), and a "Last reviewed" line, which is deliberately the document's only date.
- **Patch management.** `docs/patch_management.md` documents how each layer is patched and on what cadence; `unattended-upgrades` is now enabled on production, restricted to the Debian security suite with no automatic reboot. HECVAT PPPR-01 changes from No to Yes.
- **Recovery procedure.** `docs/deploy.md` said sslmate and a long-gone wmflabs backup host; it now describes Let's Encrypt via certbot, the same-Linode snapshot restore that preserves the allowlisted addresses, the firewall, and the real backup locations.

Related documentation: [Akamai Cloud Firewall](https://techdocs.akamai.com/cloud-computing/docs/cloud-firewall) and its [comparison with Linux firewalls](https://techdocs.akamai.com/cloud-computing/docs/comparing-cloud-firewalls-to-linux-firewall-software) (stateful, evaluated before host rules); [IPv6 on Linodes](https://techdocs.akamai.com/cloud-computing/docs/an-overview-of-ipv6-on-linode) (SLAAC requires ICMPv6); [Debian LTS schedule](https://wiki.debian.org/LTS); [Oj modes](https://github.com/ohler55/oj/blob/develop/pages/Modes.md).

## AI usage

This PR was produced with Claude Code (Claude Fable 5.1) under Sage Ross's direction. The agent read the existing HECVAT, the Debian upgrade notes and the security-status draft, researched the Akamai firewall and Oj documentation, drafted every file change and every commit message, and ran the specs and the external network probes. Sage decided what to do, ran every command on the production server and the clone himself and pasted the output back (the agent never had shell access to production), created and attached the firewall in Cloud Manager, and reviewed the HECVAT wording. The Brakeman scan that surfaced the Oj issue was run by the agent; the judgement that it was a real gap rather than a false positive came from verifying Oj's behaviour with the app loaded.

Scale of effort: eight commits, all on 2026-09-11. Seven came from a single session of several hours, roughly half of it the interactive firewall rollout; two of those are tidy-ups after Sage's review of the HECVAT wording (no dates or planned-change narration in the notes, apart from the header's review date and the OS support window), and the seventh adds the patch-management process after Sage chose it from a list of remaining easy wins. The eighth, from a short second session, addresses the Claude Code review posted on this PR: dead documentation references, the deploy-branch description, the audit step's position in CI, a pre-existing broken README link, and a spec pinning Oj strict mode. The "What this PR does" summary and the rest of this description were also drafted by Claude Code and may contain errors. This PR description was drafted using Claude Code and the `prepare-pr` skill.

## Screenshots

No UI changes. Verification evidence instead:

Oj mode with the app loaded, before and after the initializer change:

```
$ bundle exec rails runner 'p Oj.default_options[:mode]; p Oj.load(%q({"^o":"Object","x":1}))'
:object
#<Object:0x00007f519d1519e0 @x=1>        # before

:strict
{"^o" => "Object", "x" => 1}              # after
```

Dependency audit before and after the gem bumps:

```
$ bundle-audit check --update
Name: crass    Version: 1.0.6  ...  Solution: update to '>= 1.0.7'   (x4)
Name: rubyzip  Version: 2.3.2  Criticality: High  CVE-2026-85396   Solution: update to '>= 3.4.0'
Vulnerabilities found!

$ bundle exec bundle-audit check --update
No vulnerabilities found
```

Production behind the firewall, probed from outside on 2026-09-11 (unlisted ports time out instead of being refused, which is the Cloud Firewall dropping them):

```
prod :22 -> open
prod :80 -> open
prod :443 -> open
prod :3306 -> dropped after 6.0s
prod :8080 -> dropped after 6.0s
https -> 200 in 0.146662s
DB-backed page -> 200 in 0.099658s
http redirect -> 301 -> https://dashboard.wikiedu.org/
ping -> rtt min/avg/max/mdev = 20.827/20.861/20.895/0.034 ms
```

Spec runs: the twelve spec files covering every `Oj.load` call site (127 examples, green); `spec/lib spec/models spec/services spec/helpers spec/workers spec/mailers spec/presenters` (2644 examples, one failure in `upload_importer_spec` that fails identically on master because Commons now serves thumbnails from a different host); the zip export spec (8 examples); the HECVAT page spec (7 examples).

- The HECVAT "Last reviewed: 2026-09-11" line asserts Wiki Education's confirmation of the revised answers; Sage should re-read the changed notes (APPL-03, DOCU-02, FIDP-01/02/05, DATA-01, ITAC-06, APPL-06, VULN-01) before this merges, since `/hecvat` is public.
- `bundler-audit` will fail CI whenever a new advisory lands for a gem in the lockfile, including cases where the Dashboard's usage is unaffected (as rubyzip's was). The escape hatch is a `.bundler-audit.yml` ignore entry with a written justification; none is needed today.
- `rubyzip` 2 → 3 is a major bump. The only use is `Zip::OutputStream.write_buffer` in `ExportTrainingModuleDraft`, whose spec passes; `selenium-webdriver` allows `< 4.0`.
- Oj strict mode is a global default. No gem in the lockfile depends on `oj`, there is no `Oj.dump` in app or lib, and the Rails encoder installed by `optimize_rails` is unaffected, so the change only touches our own `Oj.load` calls.
- Not in this PR, flagged for a decision: whether the Fall 2026 study's wickie.ink enrollment call (not yet live) and the S3-compatible CSV store belong in the HECVAT's subprocessor list (DPAI-03); Brakeman's 40 warnings (13 High, none confirmed exploitable at a glance) need a reviewed baseline before it can join CI.
- `unattended-upgrades` will restart Apache or MariaDB for a few seconds when a security update to those packages lands (around 06:00 server time). Accepted by default; a package blacklist is the one-line alternative if that ever proves disruptive.
- Operational follow-ups recorded in the runbook: `certbot renew --dry-run` on prod, a check that the IPv6 `valid_lft` resets behind the firewall, the next Sunday dump, 2FA for the other Akamai account users, `PermitRootLogin prohibit-password`, a firewall for the other Akamai-hosted servers, and rebuilding staging off Debian 10.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,5 +181,10 @@ jobs:
       - name: Check whitespace
         run: git diff --check HEAD~1
 
+      # Fails the build when Gemfile.lock contains a gem with a published
+      # security advisory (ruby-advisory-db, refreshed on each run).
+      - name: Ruby dependency vulnerability audit
+        run: bundle exec bundle-audit check --update
+
       - name: Ruby linting
         run: bundle exec rubocop
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,13 @@ jobs:
            ruby-version: 3.4.8
            bundler-cache: true
 
+      # Fails the build when Gemfile.lock contains a gem with a published
+      # security advisory (ruby-advisory-db, refreshed on each run). Runs
+      # before the test suites so a new advisory surfaces in the first
+      # minutes of the build, and even when a later step fails.
+      - name: Ruby dependency vulnerability audit
+        run: bundle exec bundle-audit check --update
+
       - name: setup Node
         uses: actions/setup-node@v5
         with:
@@ -180,11 +187,6 @@ jobs:
 
       - name: Check whitespace
         run: git diff --check HEAD~1
-
-      # Fails the build when Gemfile.lock contains a gem with a published
-      # security advisory (ruby-advisory-db, refreshed on each run).
-      - name: Ruby dependency vulnerability audit
-        run: bundle exec bundle-audit check --update
 
       - name: Ruby linting
         run: bundle exec rubocop
--- a/Gemfile
+++ b/Gemfile
@@ -129,6 +129,7 @@ group :development do
 end
 
 group :development, :test do
+  gem 'bundler-audit', require: false
   gem 'pry-rails'
   gem 'byebug'
   gem 'rspec-rails', '6.1.0'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -138,6 +138,8 @@ GEM
     anonymous_loader (0.1.3)
       version_gem (~> 1.1, >= 1.1.14)
     ast (2.4.3)
+    auth-sanitizer (0.2.3)
+      version_gem (~> 1.1, >= 1.1.14)
     aws-eventstream (1.4.0)
     aws-partitions (1.1281.0)
     aws-sdk-core (3.254.1)
@@ -157,8 +159,6 @@ GEM
       aws-sigv4 (~> 1.5)
     aws-sigv4 (1.12.1)
       aws-eventstream (~> 1, >= 1.0.2)
-    auth-sanitizer (0.2.3)
-      version_gem (~> 1.1, >= 1.1.14)
     axe-core-api (4.11.3)
       dumb_delegator
       ostruct
@@ -187,6 +187,9 @@ GEM
       railties (>= 5.0)
     browser (5.3.1)
     builder (3.3.0)
+    bundler-audit (0.9.3)
+      bundler (>= 1.2.0)
+      thor (~> 1.0)
     byebug (11.1.3)
     capistrano (3.20.0)
       airbrussh (>= 1.0.0)
@@ -229,7 +232,7 @@ GEM
     connection_pool (3.0.2)
     crack (0.4.5)
       rexml
-    crass (1.0.6)
+    crass (1.0.7)
     cronex (0.15.0)
       tzinfo
       unicode (>= 0.4.4.5)
@@ -644,7 +647,7 @@ GEM
     ruby-graphviz (1.2.5)
       rexml
     ruby-progressbar (1.13.0)
-    rubyzip (2.3.2)
+    rubyzip (3.6.0)
     securerandom (0.4.1)
     selenium-webdriver (4.43.0)
       base64 (~> 0.2)
@@ -764,6 +767,7 @@ DEPENDENCIES
   bootsnap
   breadcrumbs_on_rails
   browser
+  bundler-audit
   byebug
   capistrano
   capistrano-bundler
--- a/config/initializers/oj.rb
+++ b/config/initializers/oj.rb
@@ -2,3 +2,10 @@
 
 # https://github.com/ohler55/oj/blob/develop/pages/Rails.md
 Oj.optimize_rails
+
+# Oj's default :object mode instantiates arbitrary Ruby classes from JSON keys
+# such as "^o", so Oj.load on any JSON we did not write ourselves (API responses,
+# training content read from wiki pages) would be an insecure-deserialization
+# hole. :strict mode returns only JSON-native types (Hash, Array, String, numbers,
+# booleans, nil). optimize_rails does not change this default.
+Oj.default_options = { mode: :strict }
--- a/docs/README.md
+++ b/docs/README.md
@@ -14,6 +14,7 @@
 - [Deployment](deploy.md)
 - [Admin Guide](docs/admin_guide.md)
 - [Upgrading dependencies](upgrade_dependencies.md)
+- [Patch management](patch_management.md)
 - [Tools & Integrations](tools.md)
 - [Model diagram](../erd.pdf)
 - [Miscellaneous Credits](miscellaneous_credits.md)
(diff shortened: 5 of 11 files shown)


ragesoss (commented):
**Review notes**

This review was produced in a Claude Code session and is offered as input to read, weigh, and use or discard on the developer's own judgment, not as findings to act on automatically.

**Verified locally** (branch checked out, master merged, `bundle install`):

- `bundle exec bundle-audit check --update` reports no vulnerabilities.
- RuboCop is clean on `config/initializers/oj.rb`; `yarn build` compiles.
- The spec files covering all twelve `Oj.load` call sites, the zip export, and the `/hecvat` page pass (122 examples, 0 failures, 2 pending).
- With the app loaded, `Oj.default_options[:mode]` is `:strict` and a `{"^o":"Object"}` payload returns a plain Hash. `JSON.parse` and `ActiveSupport::JSON.decode` were already pinned to compat mode by Oj's Rails mimic, so they are unchanged. Of the installed gems, only aws-sdk-core and i18n call Oj directly; aws-sdk-core passes an explicit `mode: :compat`, and the i18n call is in the KeyValue backend, which this app does not use.
- `rubyzip` 3.6.0 still provides `Zip::OutputStream.write_buffer` with the same shape used by `ExportTrainingModuleDraft`, and its spec passes.

**Findings**

1. `server_config/firewall.md` cites `.claude/debian_upgrade_plan-2026-09-01.md` as the source of the facts table. That file is not tracked in git, so the reference is dead for anyone reading the public doc.
2. `docs/patch_management.md` says the application deploys "from `master` via Capistrano". `config/deploy/production.rb` deploys the `production` branch, and `docs/deploy.md` describes merging master into production first.
3. The `bundler-audit` step runs after the full RSpec suite and before RuboCop. A new advisory therefore surfaces only at the end of a long CI run, and not at all if an earlier step fails. Placing it before the test steps would fail fast. Judgment call.
4. Two behavior changes accompany the strict-mode switch, both confirmed by probing. Object mode converted any string beginning with a colon into a Symbol (`":foo"` became `:foo`); strict mode keeps it a String, which is a fix and is relevant to training content read from wiki pages. Separately, `NaN` and `Infinity` now raise `Oj::ParseError` where object mode accepted them; `Utils.parse_json` already rescues that class, and the other call sites read MediaWiki or Toolforge JSON.
5. The firewall doc publishes the window during which MariaDB was reachable, the Cloud Firewall ID, both server IPs, and that SSH is open to the internet. These are discoverable independently, but the level of detail in a public record is worth a deliberate decision. The "After it is live: documentation updates" section also still reads as pending instructions, though this PR completes two of its three items.
6. Pre-existing and adjacent: the `docs/README.md` entry just above the new patch-management line links to `docs/admin_guide.md`, which resolves to a missing `docs/docs/admin_guide.md`.

**Not verified**: the server-side facts (unattended-upgrades configuration, the `mariadb.socket` fix, staging's OS and IP) cannot be checked from the repository and were taken as stated. CI was still running at the time of writing.

**Optional**: a spec asserting the default Oj mode is `:strict` and that a `^o` payload returns a Hash would pin the security property against a future refactor of the initializer.

*Drafted in a Claude Code session (~20 min, 2 interactions). Sage requested the review and reviewed this comment before posting. The local verification listed above (bundle-audit, RuboCop, yarn build, 122 spec examples, and the Oj mode probe) was run by Claude Code, not independently reproduced by Sage. No AI-generated code is included in this comment.*

PR #7066, opened 2026-09-11
Add Scholars & Scientists retention report card page

ragesoss:
## What this PR does

Adds an admin-only, Wiki Education Dashboard-only **report card** page for Scholars & Scientists courses, automating the staff spreadsheet that has been assembled by hand from the retention predictors CSV (`lib/analytics/retention_predictors_csv_builder.rb`). `/report_cards` lists campaigns that have computed data; `/report_cards/:campaign_slug` shows one row per `FellowsCohort` course in that campaign, in the spreadsheet's three column groups (General / During the course / After the course), with a totals row and a CSV download.

The retention metrics come from the live `usercontribs` API and are far too slow to compute on request, so this PR also adds a store for them: a `retention_stats` table with one row per course × student, filled by `UpdateCourseRetentionStats` from the per-student calculator extracted out of the CSV builder (`RetentionStudentStats`; the CSV's output is unchanged). A daily worker keeps the rows current for recently ended `FellowsCohort` courses only — each course is recomputed at most three times, at the checkpoints where a metric becomes final (end + 1, + 31 and + 91 days) — and `rake retention_stats:backfill[campaign_slug]` fills in historical cohorts one campaign at a time. Student-program courses are never touched, and the worker cannot reach courses older than four months, so it can never silently backfill history.

Semantics follow the CSV: long-term Wikipedians (500+ edits before the course) are counted under Participants and left out of every other figure. The spreadsheet's two "…of which is a long-term Wikipedian" sub-columns under the survival counts are therefore omitted (they would always be zero); words and uploads per participant divide by all participants, since those course totals include everyone's work.

**Review follow-up (second commit).** The review comment below found that a `usercontribs` request failing after `WikiApi`'s retries was read as "no edits", so a student whose edits could not be fetched would have been stored as a non-editor and never recomputed. A failed fetch (of any page, in either fetcher) now raises `RetentionFetchError`; the daily worker and the backfill task rescue it per course, so the course keeps its previous rows, stays due, and the other courses are not held up. The on-demand retention CSV now fails its job on a fetch error rather than producing a wrong download. Separately, a cell that will never fill in (nobody edited during the course, or every participant is a long-term Wikipedian) now shows a dash rather than "pending": each column carries the checkpoint its figure becomes final at, and the row knows which checkpoints its stored data has reached. Also from the review: a course with no students no longer stays due every day, the average-words figure rounds once, and the feature spec's totals check was tightened.

### Checked against the spreadsheet with real data

To illustrate the page with real numbers, the new backfill task was run locally against a copy of the production database (the 2026-06-21 dump) for `wikipedia_scholars__scientists_202425`: 20 courses, 299 students, about 20 minutes of serial `usercontribs` calls. The screenshots below are that data. Comparing the page's figures with the 2024–25 tab of the staff spreadsheet (which covers 13 of those 20 courses):

| Course | Spreadsheet: participants / editing sessions / edited in 30 days / survivors | Page | Why they differ |
|---|---|---|---|
| 250 by 2026-1 | 18 / 108 / 5 / 2 | 18 / 108 / 5 / 2 | Identical in every column. (The sheet's typed-in 31% is 29.4% by its own formula.) |
| 250 by 2026-2 | 28 / 228 / 4 / 1 | 28 / 228 / 4 / 1 | Identical in every column. |
| Anthology of Latino Poets | 17 / 75 / 2 / 1 | 17 / 75 / 2 / 1 | Identical in every column. |
| APS Wiki Scientists 8 | 17 / 66 / 1 / 0 | 17 / 66 / 1 / 0 | Identical in every column. |
| RAO Wiki Scientist Training-2 | 15 / 61 / 1 / 1 | 15 / 60 / 1 / 1 | One editing session fewer. |
| Histories of Social Movements and Political Change | 20 / 104 / 1 / 0 | 20 / 103 / 1 / 0 | One editing session fewer. |
| 250 by 2026-3 | 22 / 189 / 5 / 2 | 22 / 171 / 4 / 2 | One participant's edits no longer come back from the API at all; the survival counts are unchanged. |
| Latinx Past Wiki Scholars | 23 / 150 / 2 / 1 | 23 (1 long-term) / 108 / 1 / 0 | The one survivor is a long-term Wikipedian, excluded here. |
| Climate Finance Biographies | 26 (2 long-term) / 171 / 5 / 5 | 26 (3) / 78 / 3 / 4 | A third participant crosses the 500-edit threshold that the sheet still labels as 1000. |
| Global Approaches to Climate Finance 3 | 25 (3) / 141 / 7 / 0 | 25 (5) / 101 / 4 / 0 | Same threshold effect. |
| Global Approaches to Climate Finance 4 | 11 (3) / 315 / 8 / 4 | 11 (5) / 33 / 3 / 3 | Same. |
| Global Approaches to Climate Finance 5 | 11 (5) / 369 / 9 / 3 | 11 (5) / 33 / 4 / 3 | The sheet's during-course and 30-day figures include the five long-term Wikipedians; its survival figures don't. |
| The 2024 Election Editing Workshop | 14 / 169 / 8 / 5 | 14 (9) / 9 / 1 / 1 | Nine of the fourteen participants were long-term Wikipedians. |

Where no long-term Wikipedians are involved, the page reproduces the spreadsheet exactly. The session drift on three courses looks like a property of the source rather than the code: `usercontribs` does not return edits to pages that have since been deleted, so a participant whose sandbox or draft was deleted loses those sessions (inferred from the pattern, not verified per participant). Storing each course's figures once its windows close, as this PR does, also freezes them against that kind of drift. The other seven courses in the campaign (The 2024 Election-2 through -6, WITH Wiki Scientists 7 and 8) were never on the sheet and now appear alongside.

## AI usage

All of the code, specs and the commit message in this PR were written by Claude Code (Claude Fable 5.1) under Sage Ross's direction, in one session on 2026-09-11. Sage supplied the spreadsheet to automate and the goal; Claude Code did the reconnaissance (mapping every spreadsheet column to Dashboard data and checking the mapping against a prod-copy database), wrote a plan with the design decisions laid out as explicit choices, and Sage approved the recommended option on each. One question from Sage during planning caught a real flaw — the daily worker as first drafted would have backfilled every historical course on its first run — which was fixed in the plan before any code was written. Sage wrote the two pieces of user-facing copy (the legend and the empty state); the column headers are the spreadsheet's own text. The illustration below was produced by running the new backfill task locally against a copy of production data and screenshotting the resulting page.

The first commit was made at the end of that session (about an hour and a half of wall-clock time, planning included). The second commit came from a separate Claude Code session on the same day: Sage asked it to read the review findings (themselves posted from a Claude Code review session), decide which warranted changes, and make them. Claude Code triaged the eight items, implemented five and set three aside with reasons, in one autonomous pass; Sage's input on the result was to remove a `[PLACEHOLDER]` it had left in the legend for a sentence explaining the dash, judging the dash self-explanatory. This PR description, and the summary above, were also drafted by Claude Code using the `prepare-pr` skill, then revised by it for the second commit, and may contain errors.

## Screenshots

No "before": the page is new. These are the real 2024–25 figures described above, rendered in the test environment from an export of the computed rows (students replaced by placeholder accounts, which the page never shows; instructor names and course titles are the real, public ones). They predate the second commit; no 2024–25 course has a figure with nobody to count, so the dash marker does not appear in them.

**Report cards index**

| Before | After |
|--------|-------|

**Report card at the default page width (the table scrolls horizontally)**

| Before | After |
|--------|-------|

**The whole 2024–25 report card, all 20 courses and the totals row**

| Before | After |
|--------|-------|

- **Two judgment calls to confirm**: dropping the always-zero "…of which long-term Wikipedian" survival sub-columns, and dividing words/uploads by all participants rather than the counted ones. The review found both readings consistent with the exclusion rule; both are one-line changes if the other reading is preferred.
- **Set aside from the review, deliberately**: the "percentage of active participants who edited after" can exceed 100%, as the spreadsheet's own `M / (D − H)` does (a participant who edited only after the course counts in the numerator but not the denominator); noted in a code comment rather than changed. The predicted `RecordNotUnique` from running the backfill while the daily worker is on the same course cannot occur, since `insert_all` skips duplicates on MySQL. CSV cells are not formula-escaped, consistent with the other exports.
- **Instructor column**: the spreadsheet names one lead facilitator per course; the Dashboard only knows the instructor role, and Will Kent is a co-instructor on nearly every course. The page lists every instructor by real name.
- **Numbers will differ from the hand-built spreadsheet in places**: several of its "average" cells were typed over rather than computed, and the 2024–25 tab counted long-term Wikipedians in some denominators. The comparison above shows where the automated figures line up and where they don't.
- **After deploy**: run `bundle exec rake "retention_stats:backfill[wikipedia_scholars__scientists_202526]"` (and `_202425`) once on the server. Each is a few thousand serial API calls, about a minute per course.
- Follow-up candidate, not in this PR: serve the on-demand retention CSV from stored rows once a course has passed its last checkpoint, so that download becomes instant.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/assets/stylesheets/modules/_report_cards.styl
+++ b/app/assets/stylesheets/modules/_report_cards.styl
@@ -1,0 +1,44 @@
+// Admin-only Scholars & Scientists report cards (/report_cards).
+.report-card
+  &__legend
+    max-width 800px
+
+  &__scroll
+    overflow-x auto
+    margin-bottom 40px
+
+  &__table
+    min-width 2200px
+
+    thead th,
+    tbody td
+      padding 8px 10px
+      font-size 13px
+
+    thead th
+      vertical-align bottom
+      min-width 90px
+      max-width 150px
+      white-space normal
+
+    th.report-card__group
+      text-align center
+      font-size 15px
+      border-bottom 2px solid #ddd
+
+  &__title
+    min-width 180px
+
+  &__number
+    text-align right
+
+  td.report-card__number
+    white-space nowrap
+
+  &__pending
+    color #999
+    font-style italic
+
+  &__totals td
+    font-weight bold
+    border-top 2px solid #ddd
--- a/app/assets/stylesheets/modules/_report_cards.styl
+++ b/app/assets/stylesheets/modules/_report_cards.styl
@@ -39,6 +39,9 @@
     color #999
     font-style italic
 
+  &__not-applicable
+    color #999
+
   &__totals td
     font-weight bold
     border-top 2px solid #ddd
--- a/app/controllers/report_cards_controller.rb
+++ b/app/controllers/report_cards_controller.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require_dependency "#{Rails.root}/lib/analytics/retention_report_card"
+
+# Admin-only report cards for Scholars & Scientists (FellowsCohort) courses, one
+# per campaign, built from the RetentionStat rows UpdateRetentionStatsWorker and
+# the retention_stats:backfill task store. Wiki Education Dashboard only: the
+# program, and the daily job that computes the rows, exist only there.
+class ReportCardsController < ApplicationController
+  layout 'admin'
+  before_action :require_wiki_ed
+  before_action :require_admin_permissions
+
+  # Campaigns with at least one computed Scholars & Scientists course, newest
+  # first. Derived from stored data, so a campaign appears only once its courses
+  # have been computed.
+  def index
+    @campaigns = Campaign.joins(courses: :retention_stats)
+                         .where(courses: { type: 'FellowsCohort' })
+                         .select('campaigns.*, COUNT(DISTINCT courses.id) AS courses_count')
+                         .group('campaigns.id')
+                         .order(Arel.sql('MAX(courses.start) DESC'))
+  end
+
+  def show
+    campaign = Campaign.find_by(slug: params[:campaign_slug])
+    raise ActionController::RoutingError, 'Not Found' unless campaign
+
+    @report_card = RetentionReportCard.new(campaign)
+    respond_to do |format|
+      format.html
+      format.csv { send_data @report_card.to_csv, filename: csv_filename(campaign) }
+    end
+  end
+
+  private
+
+  def require_wiki_ed
+    raise ActionController::RoutingError, 'Not Found' unless Features.wiki_ed?
+  end
+
+  def csv_filename(campaign)
+    "report-card-#{campaign.slug}-#{Time.zone.today}.csv".tr('/', '-')
+  end
+end
--- a/app/helpers/report_cards_helper.rb
+++ b/app/helpers/report_cards_helper.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module ReportCardsHelper
+  # A numeric report card cell: formatted per the column, or the "pending"
+  # marker when the figure is not available yet.
+  def report_card_cell(value, format)
+    return tag.span(t('report_cards.pending'), class: 'report-card__pending') if value.nil?
+
+    case format
+    when :integer then number_with_delimiter(value)
+    when :decimal then number_with_precision(value, precision: 1)
+    when :percent then "#{number_with_precision(value, precision: 1)}%"
+    else value
+    end
+  end
+end
--- a/app/helpers/report_cards_helper.rb
+++ b/app/helpers/report_cards_helper.rb
@@ -1,16 +1,29 @@
 # frozen_string_literal: true
 
 module ReportCardsHelper
-  # A numeric report card cell: formatted per the column, or the "pending"
-  # marker when the figure is not available yet.
-  def report_card_cell(value, format)
-    return tag.span(t('report_cards.pending'), class: 'report-card__pending') if value.nil?
+  # A numeric report card cell: the row's figure for the column, formatted per
+  # the column; the "pending" marker when the figure has not been computed yet;
+  # the not-applicable marker when it has been reached but there was nobody to
+  # count, so it never will be.
+  def report_card_cell(row, column)
+    value = row.public_send(column.key)
+    return empty_report_card_cell(row, column) if value.nil?
 
-    case format
+    case column.format
     when :integer then number_with_delimiter(value)
     when :decimal then number_with_precision(value, precision: 1)
     when :percent then "#{number_with_precision(value, precision: 1)}%"
     else value
+    end
+  end
+
+  private
+
+  def empty_report_card_cell(row, column)
+    if row.reached?(column.stage)
+      tag.span(t('report_cards.not_applicable'), class: 'report-card__not-applicable')
+    else
+      tag.span(t('report_cards.pending'), class: 'report-card__pending')
     end
   end
 end
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -82,6 +82,7 @@ class Course < ApplicationRecord
            foreign_key: 'project_id',
            dependent: :destroy
   has_one :course_stat, class_name: 'CourseStat', dependent: :destroy
+  has_many :retention_stats, dependent: :destroy
 
   #########################
   # Activity by the users #
--- a/app/models/retention_stat.rb
+++ b/app/models/retention_stat.rb
@@ -32,17 +32,18 @@ class RetentionStat < ApplicationRecord
 
   validates :user_id, uniqueness: { scope: :course_id }
 
-  # Whether the course's rows lag behind what can be computed now: there are
-  # none yet, a metric window has closed since they were computed, or the
-  # roster has changed. Nothing is due before the first checkpoint, a day after
-  # the course ends.
+  # Whether the course's rows lag behind what can be computed now: a metric
+  # window has closed since they were computed, or they no longer match the
+  # roster (which is also what makes a course with students and no rows due, and
+  # a course with neither not due). Nothing is due before the first checkpoint,
+  # a day after the course ends.
   def self.update_due?(course, now: Time.zone.now)
     current_stage = RetentionStudentStats.stage(course, now)
     return false if current_stage.zero?
 
     rows = where(course_id: course.id)
-    return true if rows.none?
-    return true if RetentionStudentStats.stage(course, rows.minimum(:computed_at)) < current_stage
+    computed_at = rows.minimum(:computed_at)
+    return true if computed_at && RetentionStudentStats.stage(course, computed_at) < current_stage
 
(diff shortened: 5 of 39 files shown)


ragesoss:
Review notes from a Claude Code session. As with any AI-driven review, these are observations to read, check, and act on or set aside by the developer's own judgment, not conclusions to be trusted automatically.

**What was checked**: the branch merged with current master; `yarn build`; RuboCop on the changed Ruby files; the PR's specs together with the related existing specs (182 examples, 0 failures); the feature spec headless (2 examples, 0 failures); CI (green). The migration is an additive `create_table` with no backfill, so it looks safe to run inside a normal deploy on both stages. An exploratory spec (AI-written, not committed) was used to reproduce the behaviors below.

### Findings

1. **A failed `usercontribs` fetch is stored as a genuine non-editor and is never recomputed.** `WikiApi#query` rescues errors, logs to Sentry and returns `nil` once its three tries are exhausted. `RetentionStudentStats#edit_times` treats `nil` as the end of the data, so the student is stored with `sessions_during: 0`, `days_to_return: 30`, `sessions_after: 0`, `edits_60_90: 0`, and `RetentionParticipantHistory` likewise counts 0 prior edits. The row carries a current `computed_at`, so `RetentionStat.update_due?` returns `false` and nothing retries; at the final checkpoint the zeros are frozen. The comment in `UpdateCourseRetentionStats#store` that "an API failure leaves the previous rows in place" holds only for exceptions, which the wrapper never lets through. In the on-demand CSV this was a one-off wrong download; stored, it is permanent. One option: have `RetentionStudentStats` treat a `nil` response (including a failed continuation page) as a fetch failure and raise, at least when computing for storage, so the Sidekiq job fails, retries later, and the previous rows survive.

2. **"pending" is shown for figures that will never arrive.** `nil` currently means both "window not yet closed" and "undefined". Reproduced on two courses past their final checkpoint: where nobody edited during the course, the percentage cell reads "pending" forever; where every participant is a long-term Wikipedian, every retention cell does. The legend says pending cells fill in 31 and 91 days after a course ends, which is not true for these. Each column could carry the checkpoint it depends on, so the row can render a different marker once the course has passed it.

3. **A course with no students stays due on every daily run of its lifecycle.** With no rows, `update_due?` returns `true`; the service stores nothing; the next day repeats. No API calls are made, so the cost is only log noise for the ~4-month lifecycle per empty course.

4. **Informational: `pct_active_editors_after` can exceed 100%.** The numerator counts anyone who edited after the course; the denominator counts only those who edited during it. Two students (one with no during-course sessions who edited after, one who edited both) give 200.0. This mirrors the spreadsheet's `M / (D − H)`, so it is faithful rather than wrong, but worth knowing when reading the page.

Smaller notes:
- `RetentionReportCardRow#avg_words` rounds to one decimal and then to an integer, so 1000.45 becomes 1001 rather than 1000.
- In the feature spec, the totals-row check for `'18'` is also satisfied by the `18,462` words cell in the same row.
- Running the post-deploy backfill for `_202526` can overlap the daily worker on a course that ended within the last 121 days; the unique index would make one side fail with `RecordNotUnique`. Running the backfill outside the daily-update window avoids it.
- CSV cells are not formula-escaped. Admin-only and consistent with the other exports, so noted only.

On the two judgment calls in the description: both readings look consistent with the exclusion rule. The dropped sub-columns are structurally zero once long-term Wikipedians are excluded from the survival counts, and the words/uploads course totals do include everyone's work, so dividing by all participants matches what is being divided.

<details>
<summary>Exploratory spec excerpt for finding 1 (AI-written, not committed)</summary>

```ruby
let(:course) { create(:course, type: 'FellowsCohort', start: 130.days.ago, end: 100.days.ago) }
# ... one enrolled student, course.wikis = [wiki] ...
before do
  api = instance_double(WikiApi)
  allow(WikiApi).to receive(:new).with(wiki).and_return(api)
  allow(api).to receive(:query).and_return(nil) # what WikiApi#query returns after 3 failed tries
end

it 'does not store a row that looks like a genuine non-editor' do
  UpdateCourseRetentionStats.new(course, now:)
  row = course.retention_stats.find_by(user: student)
  expect(row).to be_nil.or(have_attributes(sessions_during: nil))
  # fails: sessions_during: 0, days_to_return: 30, sessions_after: 0, edits_60_90: 0
end

it 'leaves the course due for another attempt' do
  UpdateCourseRetentionStats.new(course, now:)
  expect(RetentionStat.update_due?(course, now:)).to be(true) # fails: false
end
```
</details>

*Drafted in a Claude Code session (one autonomous pass of roughly 20 minutes, 2 interactions). Sage asked for the review and for this comment, and read it before posting, but did not independently verify the findings. The spec runs, lint, build and CI results above were produced by the session; the exploratory spec that reproduces findings 1 through 4 was AI-written and was run by the session, not reviewed line by line by Sage.*

ragesoss:
Response to the review above, in commit 5ec0c45a5 (branch pushed).

**Changed**

1. **Failed fetch stored as a non-editor** — fixed. A nil response from `WikiApi#query`, for any page in either fetcher (edit timeline and pre-course edit count), now raises `RetentionFetchError`. `UpdateRetentionStatsWorker` rescues it per course (Sentry, with the course slug) and moves on, so the course keeps its previous rows and stays due for the next daily run; the backfill task prints a FAIL line and continues. The on-demand CSV now fails its Sidekiq job instead of silently reporting a non-editor. Both examples from the exploratory spec pass in the service spec: the previous rows survive and `update_due?` stays true.
2. **"pending" for figures that never arrive** — fixed as suggested. `RetentionReportCard::Column` carries the checkpoint stage its figure becomes final at, and `RetentionReportCardRow#reached?(stage)` says whether any of the row's courses has stored rows computed at or past it. A nil figure at a reached stage renders as a dash; otherwise "pending". Judged from the rows' `computed_at`, not the calendar, so a course past a checkpoint whose rows have not been recomputed yet still reads pending. The CSV keeps blanks for both.
3. **Empty course stays due** — fixed. `update_due?` no longer special-cases "no rows yet"; the roster comparison covers it, so a course with no students and no rows is not due.
4. **`avg_words` double rounding** and **the `'18'` feature-spec check** — fixed.

**Not changed**

- **`pct_active_editors_after` above 100%** — kept as is, since it mirrors the sheet's `M / (D − H)`; the code comment now says so.
- **Backfill overlapping the daily worker** — `RetentionStat.insert_all` skips duplicate rows on MySQL (`ON DUPLICATE KEY UPDATE`), so the predicted `RecordNotUnique` cannot occur; the worst case is two runs storing the same figures. Running the backfill outside the daily-update window is still the tidy choice.
- **CSV formula escaping** — consistent with the other exports; left alone.

*Drafted and posted by a Claude Code session. Sage asked for the triage and the changes and for the PR to be updated, and removed one `[PLACEHOLDER]` from the result; this comment was posted by the session without Sage reviewing it first.*

Issue #7026, opened 2026-09-01
Add LTI 1.1 support: launch-only "companion mode"


ragesoss:
## Why

A pilot instructor's institution supports only LTI 1.1 — they have not
completed a FERPA review process for LTI 1.3 applications. This is
unlikely to be the last institution in that position: LTI 1.1 remains
widely deployed even though 1EdTech deprecated it (certification ended
June 2021, support June 2022) and formally deprecated its OAuth 1.0a
security model.

## The shape of the problem

The Canvas integration has four pillars. Under LTI 1.1, Canvas offers
no equivalent for three of them:

| Pillar | LTI 1.1 status |
|---|---|
| Launch + opaque identity + enrollment | **Works.** OAuth 1.0a signed POST with an opaque `user_id`; Canvas's "Anonymous" privacy level withholds names/emails just as our 1.3 registration does |
| Roster sync (NRPS) | **No equivalent.** Canvas's LTI 1.1 membership service URL is documented "for future use only" |
| Bulk assignment import (Deep Linking via the Modules menu) | **No equivalent.** The `module_index_menu_modal` placement supports deep linking for 1.3 tools only; 1.1 Content-Item is one assignment at a time, with no points, no due date, no tag |
| Background grade push (AGS) | **Structurally different.** Basic Outcomes can only post a grade for a (student, assignment) pair *after that student has launched that assignment* — Canvas's docs call this "one of the major limitations of the LTI 1.1 Outcomes Service." It also cannot express our "Submitted, PendingManual, no score" policy for instructor-judged work |

So LTI 1.1 support is not a compatibility layer under the existing
feature set. It's a decision about what the 1.1 product *is*.

## Decision: launch-only companion mode

LTI 1.1 support is a **one-stop-shop iframe with no grade passback**:
the tool appears in course navigation, instructors link their Dashboard
course, students connect their Wikipedia account and are enrolled
automatically, and both see their Dashboard course data in-frame — with
links out to the full Dashboard in a new tab.

What that preserves is most of what the integration is for:

- Students reach the Dashboard from inside Canvas, with no enrollment
  passcode juggling — launch-time enrollment already exists
  (`LtiStudentEnrollment#ensure_launch_enrollment`) and already carries
  whole installs where the nav tab is the only placement.
- The identity model is unchanged: opaque LMS id + the student's own
  Wikipedia OAuth. The anonymized data posture holds — under 1.1 we
[...]

PR #7067, opened 2026-09-15
Add LTI 1.1 launch-only companion mode behind a feature flag

ragesoss:
## What this PR does

Adds **LTI 1.1 support as a launch-only "companion mode"**, per the plan of record in #7026. Addresses #7026.

A pilot institution can only install LTI 1.1 tools. Under 1.1, Canvas has no equivalent of three of the integration's four pillars: no roster service, no bulk Deep Linking from the Modules page, and a grade service that can only post a score for a student after that student has launched the assignment. So this is not a compatibility layer under the existing feature set. It is a deliberately reduced mode: **launch, identity linking, enrollment, and the in-frame views**, with no roster sync, no assignment import, and no grade passback. Instructors grade from Dashboard data, as every non-Canvas course already does.

How it works:

- **Token boundary.** LTIAAS delivers a 1.1 launch's token as `legacy-ltik`; the launch controller folds it into `params[:ltik]` at the boundary, so the idtoken API, the `LTIK-AUTH-V2` header, and every re-emitted link and form keep one name. From then on the idtoken's `ltiVersion` (LTIAAS labels 1.1 launches `"1.2.0"`) is the version signal, exposed as `LtiSession#legacy?` and recorded on a new `lti_course_bindings.lti_version` column (default `1.3.0`; `LtiCourseBinding#legacy?`, `lti_1_3` scope).
- **Gate.** `Features.lti_legacy_launches?` (`lti_legacy_launches_enabled`, alongside the Canvas flag). A legacy launch with it off is refused with a bare 403 and a Sentry report, mirroring the platform gate. Under 1.1 LTIAAS registers one global consumer key/secret for every LMS, so the Canvas allowlist is now the only platform filter and stays failing closed.
- **Platform identity.** A real 1.1 launch carries no `platform.id`, only the LMS's consumer-instance `guid` (Canvas's `tool_consumer_instance_guid`, per root account). Legacy bindings and contexts are keyed on that guid, which is also the per-institution scoping the shared key would otherwise lose. A launch with neither is refused at the gate.
- **Roles.** LTIAAS forwards 1.1 roles unnormalized (`Instructor`, `Learner`, `urn:lti:role:ims/lis/…`). Exact-match tables classify them alongside the 1.3 suffix table everywhere (`LtiSession`, `LtiContext`, `LtiMemberLinker`). The TA URN counts as staff because Canvas's 1.1 TaEnrollment sends only that role; institution- and system-level roles, Mentor/Observer, and ContentDeveloper stay unsupported.
- **Fail-closed in depth.** Both cron dispatchers select 1.3 bindings by version (not just by stored credentials), the three sync services return early, and `LtiServiceSession.new` raises for a legacy binding. A legacy launch persists no service credentials.
- **Experience.** Every 1.1 launch is the course-navigation launch (nothing to drill into). Instructors get `instructor_status_legacy`: the linked header, the course link, and the connected-accounts count, with no roster, sync, grade-sync or import rows. Students get the existing enrollment flow and overview unchanged. Deep-link placements render an in-frame refusal (reusing the existing one-word "Unavailable" header), `sync_grades` refuses, binding a course enqueues no syncs, and the first-link grade push is skipped. The course-page sidebar payload carries `legacy`, and its staff and student views drop the sync rows.
- **Tooling and docs.** `bin/canvas-lti11-tool` installs the 1.1 tool on the test Canvas through the external_tools API (the only way a hand-installed 1.1 tool gets a course-navigation placement without an XML config); `staging_specs/lti11_legacy_launch_spec.rb` drives the whole flow live for both personas and prints what LTIAAS sent. `docs/canvas_dev_setup.md` gains a developer section with the captured idtoken described field by field and what the first real launch settled; the admin guide gains a terse LTI 1.1 section (two tables and one line); `docs/canvas_integration_todos.md` tracks the follow-ups.

Verified live on staging against canvas.wikiedu.org through the testing LTIAAS tenant: instructor first launch through setup and the launch-only status view, student first launch through enrollment and the overview, binding keyed on the Canvas guid, both contexts classified from raw 1.1 roles.

Relevant external references: LTIAAS's LTI 1.1 quick-start (vendor-supplied PDF; legacy support is enabled manually per account), [Canvas LTI variable substitutions](https://developerdocs.instructure.com/services/canvas/external-tools/lti/file.tools_variable_substitutions) (the 1.1 membership URL is "for future use only"), [Canvas grade passback](https://developerdocs.instructure.com/services/canvas/external-tools/lti/file.assignment_tools) (the Basic Outcomes limitation), [1EdTech's LTI deprecation schedule](https://www.1edtech.org/lti-security-announcement-and-deprecation-schedule).

## AI usage

Claude Code (Claude Fable 5.1) wrote all of the code, specs, tooling, and docs in this PR, and every commit message, under Sage Ross's direction. Sage set the scope (the plan in #7026 and its LTIAAS addendum), reviewed each step, and made the copy decisions: the placeholder user-facing text was removed rather than written, with the deep-link refusal reusing an existing string and the legacy status view carrying no guidance text. Sage also handled the pushes, sent LTIAAS the account ID, retrieved the 1.1 key/secret from the LTIAAS portal, and installed the tool with the new script.

Claude Code ran the local specs, deployed to staging, ran the live staging smoke test, and diagnosed the one real bug it found (no `platform.id` on a legacy launch) by fetching the launch's idtoken from LTIAAS rather than by guessing; the fix is the third commit. The "What this PR does" summary and the rest of this description were also drafted by Claude Code and may contain errors.

Scale of effort: 4 commits within a single session of roughly an hour of wall-clock time between the first and last commit (2026-09-15, 09:56 to 10:55 PDT), preceded by the planning documented in #7026. Two live staging runs failed before the passing one (the platform-identity bug, then the spec reading the wrong log file).

## Screenshots

The first two are the real thing: the 1.1 tool installed on the test Canvas, launched through LTIAAS's legacy endpoint into staging, captured by the live smoke spec. The rest are from a local screenshot spec with a stubbed idtoken; the 1.3 counterparts are included so the reduction is visible.

| Before | After |
|---|---|
| _(did not exist on master)_ | Instructor's course-navigation tab in Canvas, LTI 1.1: the launch-only status view |
| _(did not exist on master)_ | Student's course-navigation tab in Canvas, LTI 1.1: the in-frame overview |
| Instructor status view under LTI 1.3, for contrast: roster, sync rows, grade sync, import step | Instructor status view under LTI 1.1: linked header, course link, connected count only |
| _(unchanged from 1.3)_ | Student overview under LTI 1.1, with timeline content |
| _(did not exist on master)_ | In-frame refusal when a 1.1 launch reaches a deep-linking placement |
| Course-page Canvas link panel under LTI 1.3, for contrast: last sync, roster count, connected count | Course-page Canvas link panel under LTI 1.1: course link and connected count only |

- **Canvas launch point for real institutions.** A manually installed 1.1 tool (key, secret, launch URL) gets no course-navigation tab; Canvas needs either an XML tool configuration or the external_tools API for the placement, and LTIAAS provides no XML. The test Canvas uses the API path via `bin/canvas-lti11-tool`. For an institution's own install we need to decide between hosting a small static cartridge XML on the Dashboard and documenting module items / external-tool assignments as the launch point. The admin guide's LTI 1.1 section deliberately stops short of install steps until that's decided.
- **One shared secret across every 1.1 institution.** A leak anywhere lets any holder forge launches for any 1.1 platform and user (under 1.3 a compromised registration only speaks for itself). The guid keying scopes the damage per institution's data, not the forgery itself. Rotation practice is an operator decision.
- **Still unexercised on a real launch:** a Canvas TA (1.1 sends only the TA URN) and a Canvas observer (`urn:lti:instrole:ims/lis/Observer` + `urn:lti:role:ims/lis/Mentor`, which must land in `unsupported_role`). Both are covered by unit specs against Canvas's published role table.
- **Grade passback is out of scope by design** (see #7026 for the full argument); the trigger to revisit is multiple committed institutions where 1.1 is the only path, passback is decisive, and their own 1.3 timeline is more than about a year out.
- **Staging note for reviewers:** the web app's Rails log on dashboard-testing lands in `/var/log/apache2/error.log` (Passenger captures stdout); `shared/log/staging.log` carries only Sidekiq and console output. Documented in the dev-setup doc.


ragesoss:
Closing: this was opened from the wrong head branch (`staging`, which is where the working copy happened to be checked out). The same commits are being re-opened from `lti-11-companion-mode`.

PR #7068, opened 2026-09-15
Add self-hosted LTI 1.1 support as a launch-only companion mode

ragesoss:
## What this PR does

Adds **LTI 1.1 support as a launch-only "companion mode"**, per the plan of record in #7026, and the Dashboard terminates those launches itself rather than routing them through LTIAAS. Addresses #7026.

Some institutions can only install LTI 1.1 tools. Under 1.1, Canvas has no equivalent of three of the integration's four pillars: no roster service, no bulk Deep Linking from the Modules page, and a grade service that can only post a score for a student after that student has launched the assignment. So this is not a compatibility layer under the existing feature set. It is a deliberately reduced mode: **launch, identity linking, enrollment, and the in-frame views**, with no roster sync, no assignment import, and no grade passback. Instructors grade from Dashboard data, as every non-Canvas course already does.

**LTIAAS is now 1.3-only.** The branch began by terminating 1.1 launches at LTIAAS, then replaced that with our own endpoint, because LTIAAS's legacy path issues one global consumer key and secret shared by every institution. Owning the credentials is what makes per-course keys, pinning, and revocation possible. The LTIAAS legacy terminator added earlier on this branch is removed here rather than kept alongside; there is one gate, `lti_legacy_launches_enabled`, and no coexistence.

How it works:

- **Credentials are per install, issued self-service.** `LtiConsumerKey` belongs to one Dashboard course and one issuing instructor, with the secret encrypted at rest through ActiveRecord encryption. An instructor generates a pair from an unlisted page at `/courses/<slug>/canvas` and pastes it into Canvas. Regenerating deactivates the previous key, so an install has one working secret at a time. A key that is issued but never launched expires after seven days, which closes the window on a secret that leaked from an inbox. A key is deleted with its course (a foreign key with cascade, as for bindings), so none can outlive the course it was issued for.
- **Two properties a shared secret cannot have.** The first launch pins the key to that Canvas's `tool_consumer_instance_guid`, after which it is refused from anywhere else. And a leak is bounded by one course rather than every 1.1 institution.
- **The page is unlisted, not unprotected.** Nothing in the interface links to it; Wiki Education emails the URL to the beta instructors who need 1.1, which keeps the 1.1 path from competing with 1.3 for everyone else's attention. Obscurity governs discovery only. Every request still requires the feature flags, a signed-in instructor of that course, and an approved course, and the page refuses when the course is already bound through a 1.3 install.
- **Verifying the launch.** `VerifyLtiLegacyLaunch` checks, in order: the message type, the key, that the key is active and unexpired, the OAuth 1.0a signature, the timestamp (±5 minutes), the nonce, that the launch names a Canvas instance at all, and finally the Canvas pin. A signed launch with no instance guid is refused rather than accepted unpinned: accepting it would activate the key without a pin and leave it usable from anywhere. The signature is verified with the `oauth` gem against a URL rebuilt from configuration, never from the request's own `Host`, and that same constant feeds the config XML so the two cannot drift. Replay protection is a unique index on (key, nonce) with an insert-and-rescue, not a read-then-write, so two simultaneous replays cannot both pass.
- **Every refusal renders identically**, whatever the reason, so the endpoint cannot be used to enumerate which consumer keys exist. The reason goes to the log. Routine refusals (a mistyped secret, a stale clock, an expired key) log at info; a key we never issued, a replay, or a wrong-Canvas launch goes to Sentry.
- **After verification, nothing downstream can tell the difference.** `NormalizeLtiLegacyLaunch` reshapes the raw form parameters into the idtoken shape LTIAAS would have produced, and the launch redirects into the ordinary flow at `/lti?ltik=...` with a token we mint. The key's course rides along as our own claim, so the instructor's first launch **auto-binds the course** and skips the setup picker. The claim is skipped if the course has been deleted since the token was minted. It does not auto-link identity, which would create a write-once lockout needing staff to undo.
- **The launch token references state, it does not carry it.** It began as a JWT holding the whole idtoken; that passed every unit test and failed against real Canvas with `CookieOverflow` at 5879 bytes, because the Wikipedia OAuth break-out stashes the token in the Rails session and the cookie jar caps at 4 KB. It is now a 32-byte random reference to an `lti_legacy_launches` row, which also made launches revocable. LTIAAS's own ltik is short for the same reason.
- **Platform identity.** A 1.1 launch carries no `platform.id`, only the consumer-instance guid (per Canvas root account). Legacy bindings and contexts key on that guid. A launch with no identity at all is refused at the launch endpoint, before the key is activated.
- **Roles.** 1.1 roles arrive unnormalized (`Instructor`, `Learner`, `urn:lti:role:ims/lis/…`). Exact-match tables classify them alongside the 1.3 suffix table everywhere. The TA URN counts as staff because Canvas's 1.1 TaEnrollment sends only that role; institution- and system-level roles, Mentor/Observer, and ContentDeveloper stay unsupported.
- **Fail-closed in depth.** Both cron dispatchers select 1.3 bindings by version rather than by stored credentials, the three sync services return early, and `LtiServiceSession.new` raises for a legacy binding. A legacy launch persists no service credentials.
- **Experience.** Every 1.1 launch is the course-navigation launch. Instructors get `instructor_status_legacy`: the linked header, the course link, and the connected-accounts count, with no roster, sync, grade-sync or import rows. Students get the existing enrollment flow and overview unchanged. Deep-link placements render an in-frame refusal, `sync_grades` refuses, binding enqueues no syncs, and the first-link grade push is skipped. The course-page Canvas panel's link into Canvas carries the same launch URL the config XML installs, so Canvas's `external_tools/retrieve` matches the 1.1 tool rather than the LTIAAS one.
- **Install path.** A hand-entered 1.1 tool gets no course-navigation placement, so the Dashboard serves a configuration XML at `/lti/legacy/config.xml`: the launch URL, the anonymous privacy level, and a default-enabled course-navigation placement, with no secrets in it and no `domain` (Canvas's uniqueness check compares domains, and omitting it keeps a 1.1 install from colliding with a 1.3 one). An instructor installs it "By URL" inside their own course under the default "LTI - add" permission, with no administrator. That path is what 1.1 uniquely enables, so it gets its own illustrated page at `/lti/guide/instructors`.
- **Styles.** The credentials page has its own module in `modules/`, rather than borrowing `.lti-status-list` and `.lti-setup__error` from `_lti_iframe.styl`. That partial is kept outside `modules/` so the main bundle's glob cannot sweep up selectors only the iframe layout uses, which meant the borrowed rules never reached this page. Its value rows are a grid whose label column sizes to the widest label, and long keys and secrets wrap instead of clipping.
- **Docs.** `docs/canvas_integration_guide.md` now positions LTI 1.3 as the institutional install and LTI 1.1 as a single-course install done by the instructor; the account-level 1.1 walkthrough is gone, because a course-scoped key cannot back one. `docs/canvas_dev_setup.md` gains a developer section, and `docs/hecvat.md` says LTIAAS is 1.3-only.

Validated live against canvas.wikiedu.org, end to end: an instructor installs per course with a Dashboard-issued key, Canvas posts a signed launch to our endpoint, the course auto-binds, the instructor connects their Wikipedia account, and a student enrolls. `staging_specs/lti11_launch_refusals_spec.rb` then signs its own launches and confirms what the endpoint rejects: a replayed signature, a wrong secret, a key we never issued, a 20-minute-old timestamp, a signature computed over another URL, a non-launch message type, a launch from a Canvas the key is not pinned to, and a signed launch that names no Canvas instance at all, after which it reads the key back to confirm nothing activated or pinned it. Both live specs were rerun on 2026-09-17 against staging at this branch's tip, and passed.

Relevant external references: [Canvas LTI variable substitutions](https://developerdocs.instructure.com/services/canvas/external-tools/lti/file.tools_variable_substitutions) (the 1.1 membership URL is "for future use only"), [Canvas grade passback](https://developerdocs.instructure.com/services/canvas/external-tools/lti/file.assignment_tools) (the Basic Outcomes limitation), [Canvas LTI XML configuration](https://developerdocs.instructure.com/services/canvas/external-tools/lti/file.tools_xml), [1EdTech's LTI deprecation schedule](https://www.1edtech.org/lti-security-announcement-and-deprecation-schedule).

## AI usage

Claude Code wrote all of the code, specs, tooling, and docs in this PR, and every commit message, under Sage Ross's direction, with one exception noted below. Two models did the work on the first day: the first nine commits, through the instructor install page, are Claude Fable 5.1; the nine that follow, from the self-hosted rework onward, are Claude Opus 5. The twentieth and twenty-first commits, on the second day, are Claude Fable 5.1 again, as are the twenty-second through twenty-fourth on the third day; the twenty-second's view changes are Sage's own writing. Each commit's `Co-Authored-By` trailer names the one that wrote it. Sage set the scope and made the design decisions that shaped it: per-install rather than per-institution keys, auto-binding on first launch without auto-linking identity, an unlisted URL rather than a discoverable course-page entry point during the beta, and LTIAAS becoming 1.3-only rather than serving both. He also handled the pushes and the staging configuration.

User-facing copy follows the repo's standing rule that AI does not write it. The earlier views reuse existing strings or carry none. The one AI-drafted exception is the Canvas guide, which says so on the page. The credentials page began as an AI first pass that Sage asked for, marked `[PLACEHOLDER` in both view files; on 2026-09-17 Sage rewrote every string on that page himself, and the twenty-second commit carries his copy, with Claude Code updating the spec assertions that quoted the old wording. No placeholder markers remain on the branch.

Claude Code ran the local specs, deployed to staging, ran the live staging specs, and diagnosed the bugs it found rather than guessing: the missing `platform.id` on a legacy launch, the session cookie overflow, and an ActiveRecord encryption configuration that was silently inert because it ran in an initializer instead of `on_load(:active_record)`. The "What this PR does" summary and the rest of this description were also drafted by Claude Code and may contain errors.

On 2026-09-16 Claude Code also reviewed the PR, using the repo's `review-pr` workflow, with Sage as the reviewer of his own PR. Reading the code raised three suspicions, and a temporary request spec turned them into reproducible failures: the course-page link for a 1.1 course pointed at LTIAAS's domain rather than our endpoint; a consumer key outlived its course, after which every launch 500ed inside Canvas; and a signed launch with no instance guid activated a key without pinning it. Sage directed the fixes for those three plus a sweep of comments and docs that still described the LTIAAS-terminated design, and at first declined a fourth suggestion (keeping unknown-key refusals out of Sentry). The fixes, eleven regression examples, and the staging database fixup (thirteen orphaned test keys removed and the new foreign key added, over SSH with Sage's one-off approval) are the twentieth commit and the session behind it.

A second Claude Code session the same day re-reviewed that fix commit with the same workflow, found the three defects resolved and no new ones, and then, at Sage's request, fixed three of the items the first review had observed and left in place: a course row lock so two simultaneous regenerations cannot leave two active keys, a credentials-page check for a 1.3 binding specifically rather than for any binding, and a once-per-address-per-hour throttle on Sentry reports of unknown-key refusals, which Sage chose over rate limiting or demoting the report. That is the twenty-first commit, with seven spec examples.

On 2026-09-17 a third session walked through the merge and deploy checklist with Sage, rendered the credentials page's five states in a visible browser so the placeholder copy could be reviewed in place, and, after Sage rewrote that copy in his editor, updated the specs, the todos doc, and this description. That is the twenty-second commit. Sage then chose one item from the outstanding list and set the rest aside: the twenty-third commit makes the generate action redirect after POST with the secret in the flash, so a browser refresh of the secret page no longer re-posts and silently replaces the key, with two new request examples. The twenty-fourth adds a live no-guid case to the refusals spec, which had unit coverage but no case against the deployed endpoint; after Sage redeployed staging, both live specs passed against the branch tip.

Scale of effort: 24 commits over three days. The first 19 took about five and a half hours of wall-clock time (2026-09-15, 09:56 to 15:29 PDT), preceded by the planning in #7026 and two planning documents written in the same session, with direction in roughly two dozen short messages; several live staging runs failed before passing ones, each on a real defect. The review-and-fix session on 2026-09-16 was a single sitting, with direction in four short messages; the re-review-and-fix session later that day was another, with six. The third day was Sage's own copy rewrite plus two small agent commits, directed in a handful of one-line messages.

## Screenshots

The first two are the real thing: the 1.1 tool installed on the test Canvas, launched into staging, captured by the live smoke spec. The rest are from a local screenshot harness, with the 1.3 counterparts alongside where one exists, so the PR shows what the 1.1 variant leaves out.

| Before | After |
|---|---|
| _(did not exist on master)_ | Instructor's course-navigation tab in Canvas, LTI 1.1: the launch-only status view |
| _(did not exist on master)_ | Student's course-navigation tab in Canvas, LTI 1.1: the in-frame overview |
| Instructor status view under LTI 1.3, for contrast: roster, sync rows, grade sync, import step | Instructor status view under LTI 1.1: linked header, course link, connected count only |
| _(unchanged from 1.3)_ | Student overview under LTI 1.1, with timeline content |
| _(did not exist on master)_ | In-frame refusal when a 1.1 launch reaches a deep-linking placement |
| Course-page Canvas link panel under LTI 1.3, for contrast: last sync, roster count, connected count | Course-page Canvas link panel under LTI 1.1: course link and connected count only |

The unlisted credentials page at `/courses/<slug>/canvas`, in each of its five states, with the operator's final copy. This is where the self-hosted path starts for an instructor, and all of it is new.

| Before | After |
|---|---|
| _(did not exist on master)_ | Credentials page before any key exists: the self-service path named, the 1.3 alternative noted, and the generate button |
| _(did not exist on master)_ | The one render that shows the shared secret, with the consumer key, secret and configuration URL and a copy button each |
| _(did not exist on master)_ | A key issued but not yet launched from Canvas, with the seven-day expiry noted and a regenerate option |
| _(did not exist on master)_ | A course connected through a 1.1 launch, showing when it connected and when it was last opened |
| _(did not exist on master)_ | The page refusing to issue a key for a course already bound through the 1.3 install |

The credential values in the second capture are placeholders substituted by the harness, not a real key and secret. They are the lengths a real pair has, 32 and 64 characters, so the capture shows the layout an instructor actually gets.

- **No rate limiting**, on either the launch endpoint or the key-issuing action. Deliberately left out rather than shipped quietly: a limit tight enough to matter can also refuse a computer lab launching at once, since the POST comes from each student's browser and a shared NAT address looks like one client. Rails 8.1 makes it a two-line controller change with no new gem, and the memcached store already supports the increment it needs. Sentry reports of unknown-key refusals are throttled per address, so a probe loop no longer floods Sentry, but nothing throttles the requests themselves.
- **A key that expires unlaunched still shows the waiting state.** After seven days the launch endpoint refuses the key, but the credentials page keeps saying it is waiting for the first launch and gives no sign the key is dead. A dedicated expired state needs operator copy that does not exist yet.
- **No staff view of issued keys.** Answering "which courses have a 1.1 key, and has it been used" means a console. Fine for a beta measured in a handful of instructors; not fine much beyond that.
- **Moving a linked course from 1.1 to 1.3 still needs staff.** A Dashboard course binds to one Canvas tool, so an instructor relaunching from a new 1.3 install is told the course is already linked. The guide says to contact Wiki Education first.
- **A Canvas TA and a Canvas observer have not been exercised live** under 1.1. The role tables classify them, and the specs cover them, but no real launch has come from either.
- **What links to the credentials page after the beta is deliberately unanswered.** Wiki Education will talk with each beta instructor about which path suited them, and the shape of a Canvas entry point on the course page follows from that rather than preceding it.
- **The keys migration was amended after staging had applied it**, to add the foreign key to courses. Staging was brought in line by hand; production will run the migration whole.


ragesoss:
## Review notes (AI-assisted, author's own review)

This PR was written by Claude Code under Sage's direction and then reviewed by Claude Code in a separate session using the repo's `review-pr` workflow, with Sage as the reviewer. As with all AI-driven review feedback in this repo: this is meant to be read, understood and weighed, not automatically trusted or acted on. Each finding below says how it was checked.

### Findings, and what was done about them

Three defects were confirmed by a temporary request spec written during the review (AI-generated, run locally, since deleted after its cases were turned into regression specs in the PR's own files). All three trace to the mid-branch pivot from terminating LTI 1.1 launches at LTIAAS to terminating them in the Dashboard. Fixed in be1c0a30a.

1. **The course-page Canvas link for a 1.1 course pointed at LTIAAS.** `LmsIntegrationStatusController#tool_launch_url` built the legacy URL on `LTIAAS_DOMAIN`, but the 1.1 tool is installed from the Dashboard's own config XML against the Dashboard's endpoint, and Canvas's `external_tools/retrieve` matches the installed tool by URL. The PR's own spec had pinned the wrong URL. Confirmed by comparing the `launch_url` served in the config XML with the URL embedded in the status JSON for a course bound through a real signed launch. Fixed by using `VerifyLtiLegacyLaunch.launch_url` for legacy bindings; the spec now expects the dashboard host, plus a contract test against the XML endpoint.

2. **A consumer key outlived its course, and every launch then 500ed.** `belongs_to` is not required-by-default in this app, so a launch with such a key verified, `claim_course` wrote the dead course id, and the bindings→courses foreign key rejected it at `/lti` with `InvalidForeignKey`. Confirmed by destroying the course and following the launch redirect. Fixed with a foreign key from `lti_consumer_keys` to `courses` with cascade (added to the create-table migration, since the PR is unmerged) and a `Course.exists?` guard in `claim_course`, because a launch token can name a course for 24 hours. Staging already had this in the wild: 13 keys whose courses the staging specs had deleted. Those were removed and the constraint added there by hand.

3. **A signed launch with no `tool_consumer_instance_guid` activated the key without pinning it.** After that, launches from any Canvas passed for good. Canvas always sends the guid, and only the secret holder could do this, so low severity, but the description promised the refusal and it happened downstream, after activation. Confirmed at the request level. Fixed by refusing such a launch in the verifier before the pin check and by having `usable_for?` refuse a blank guid; specs check the key stays unactivated.

Also changed: a sweep of comments and docs that still described the LTIAAS-terminated design (the flag's comment in `features.rb` and the example config, `lti_session.rb`'s "unverified" notes on roles and `productFamilyCode`, the launch-session, legacy-launches and deep-linking concerns, the version migration, the normalizer, and the todos doc).

Observed and left as is, at the author's decision:

- `unknown_key` refusals go to Sentry, and that is the one refusal reachable with no knowledge of any key, so an unauthenticated POST loop is a cheap way to flood Sentry. Combined with the already-noted absence of rate limiting.
- The credentials page's "already connected through 1.3" check is "a binding exists and no key does", so a course with an active 1.1 key that later gets a 1.3 binding shows the waiting state, and its next 1.1 launch lands in the setup picker with nothing linkable.
- Two fast POSTs to generate can leave two active keys; regeneration takes no lock. The unused one expires in seven days.
- The first migration is `Migration[7.0]`, the other three `[8.1]`.
- `courses/:slug/canvas` sits before the courses#show catch-all, so a course whose slug ends in `/canvas` would be shadowed. Existing `.json` routes have the same shape.

### Checks run

- `yarn build`, RuboCop and ESLint on the changed files: clean. The only RuboCop offenses were pre-existing ones in `config/routes.rb`, identical on master.
- The PR's specs plus every existing LTI spec, on a branch merged with master: 673 examples green before the fixes, 684 after (eleven new regression examples). The feature spec passed both times.
- Migration safety: additive only. The new column on `lti_course_bindings` is nullable-with-default on a table of a handful of rows, and the three new tables are new. Safe inside a normal deploy. The real deploy prerequisite is the three ActiveRecord encryption keys in production's application.yml before the flag goes on; without them, issuing a key 500s and launches log as `bad_signature` at info.
- Coverage: the verifier, token, normalizer, credentials page states, launch-controller legacy paths and sync guards are all exercised by the PR's specs. The three gaps found were exactly the three defects and are now covered. Still unexercised on a real 1.1 launch: a Canvas TA and a Canvas observer, as the description says.

*Drafted in a Claude Code session (~1 hour, 7 messages from the reviewer). Sage directed the review and chose which findings to fix, approved the staging fixup as a one-off, and read the findings summary and this comment before posting. The three confirmed defects were reproduced by an AI-written spec that Sage did not run himself; the remaining observations were not independently verified.*

ragesoss:
## Re-review notes (AI-assisted, author's own review)

Follow-up to the review notes above, covering the fix commit be1c0a30a. As before, this is AI-driven review feedback meant to be read and weighed, not acted on automatically.

All three defects from the first review are resolved in that commit, each with regression specs in the PR's own files: the course-page link for a 1.1 course now carries the Dashboard's own launch URL (with a contract test against the config XML); a consumer key is deleted with its course, and `claim_course` skips a course that no longer exists; and a signed launch with no instance guid is refused before it can activate the key. The comment and doc sweep removed the LTIAAS-terminated descriptions it set out to.

No new defects were found in the fix commit. Two things were checked because the fixes could have introduced them: the nonce table has no foreign key to consumer keys, so the new cascade cannot be blocked by leftover nonce rows (they are swept within ten minutes), and the foreign key joins matching integer columns.

Observed and left as is: four LTI spec files set `ENV['dashboard_url']` without restoring it. Running the status-controller spec ahead of the specs that read that variable for templates and sandbox names produced no failure, so this is cosmetic. The five items deferred in the first review remain deferred.

Checks: `yarn build`, RuboCop on the changed Ruby files (clean), all LTI specs plus the PR's modified specs on a branch merged with current master (694 examples, 0 failures), the feature spec via `bin/feature-spec`, and both CI runs green at the PR head.

*Drafted in a Claude Code session (~20 min, 1 message from the reviewer). Sage requested the re-review and read this comment before posting; he did not independently rerun the specs or re-read the fix commit in this session.*

ragesoss:
## Follow-up: three deferred items now fixed

Commit 1e96419a8 closes the first three items the review notes above listed as observed and left as is:

- **Two fast regenerations could leave two active keys.** `LtiConsumerKey.generate_for` now takes a row lock on the course inside its transaction, so the second of two simultaneous requests waits for the first and then replaces the key it made. A spec pins the `FOR UPDATE` read, since a real race cannot run inside a transactional example.
- **The credentials page's "already connected through 1.3" check** now asks for a 1.3 binding specifically, via the existing `lti_1_3` scope, rather than "any binding and no key". A course with a live 1.1 key that later gets a 1.3 install shows the standard-integration page and refuses to regenerate; a course's own legacy binding does not count as elsewhere.
- **Sentry flooding from unknown-key refusals.** Those refusals are still reported, but once per address per hour, counted with the cache store's increment. Every refusal still goes to the log, only that one reason is throttled, and a cache that returns nothing fails open and reports everything, as before.

Seven new spec examples cover the three changes. The dev-setup doc describes the throttle and the one-active-key guarantee. Still open, unchanged: the two cosmetic notes (migration version, route shadowing), rate limiting on the endpoints, and the other items in the PR description's open questions.

*Drafted in a Claude Code session, continuing the re-review session above. Sage chose the throttle over rate limiting or demoting the report, asked for all three fixes, and read this comment before posting. Claude Code wrote the code and specs; Sage did not rerun the specs himself.*

ragesoss:
## Deployment checklist: production rollout and instructor beta

AI-drafted from the code, the PR description, and the review sessions above. Check each item against the actual production configuration before acting on it; nothing here was executed against production.

### Before merging

- **Placeholder copy must not ship.** `app/views/course_canvas_credentials/show.html.haml` and `_regenerate.html.haml` carry `[PLACEHOLDER` markers over AI-drafted first-pass copy. Either approve the wording and remove the markers, or replace it. The instructor guide (`docs/canvas_instructor_install.md`) is AI-drafted too and says so on the page; confirm its wording and contact address are what testers should see.
- **Two credentials-page behaviors testers are likely to hit**, from the 2026-09-16 review pass: a browser refresh of the just-generated secret page re-POSTs and silently replaces the key the instructor has already pasted into Canvas, and a key that expires unused still shows the "waiting for your first launch" state. The first is worth fixing before inviting anyone, since a tester who refreshes breaks their own install and the page gives no hint why.
- The consolidation refactors from the same review are optional and can follow in a separate PR.

### Production configuration (`config/application.yml` on dashboard.wikiedu.org)

- Generate the three ActiveRecord encryption keys with `bin/rails db:encryption:init` (it only prints random values; run it anywhere) and add them as `active_record_encryption_primary_key`, `active_record_encryption_deterministic_key`, and `active_record_encryption_key_derivation_salt`. Production has no fallback: without them everything else works and the generate POST is a 500.
- Keep those three values wherever the rest of `application.yml` is backed up. Losing them makes every stored shared secret unreadable, which breaks every 1.1 install until each instructor regenerates.
- Set `lti_legacy_launches_enabled: 'true'`. `canvas_integration_enabled` is already on in production (the launch route redirects rather than 404s) and `dashboard_url` is already set; the new code reads both.
- Figaro loads `application.yml` at boot, so set these before the deploy's restart, or restart afterwards.

### Deploy

- Merge, then deploy the `production` branch with Capistrano as usual. The four migrations run inside the deploy: three new tables and one NOT NULL column with a literal default on `lti_course_bindings`, which is small. Instant on MariaDB 10.11; no out-of-band step.
- The `oauth` gem was already installed transitively; `bundle install` in the deploy picks up the explicit declaration.
- Programs & Events Dashboard: nothing to do. The migrations land at its next `wmflabs` deploy and create empty tables; the flag defaults to false there, and the Sidekiq hosts need nothing beyond their normal routine because the sync workers are gated on `canvas_integration_enabled`, which is off on Peony.

### Smoke test in production, before inviting anyone

1. `curl -sI https://dashboard.wikiedu.org/lti/legacy/config.xml` should return 200, and the body's `launch_url` should be `https://dashboard.wikiedu.org/lti/legacy/launch`. A 404 means one of the two flags is off.
2. `curl -s -o /dev/null -w '%{http_code}\n' -X POST https://dashboard.wikiedu.org/lti/legacy/launch -d x=1` should return 400 (the framable error page). This refusal is routine: logged at info as `reason=not_a_launch`, no Sentry event. A 500 means the endpoint itself is misconfigured.
3. On a staff-owned, approved production test course, sign in as its instructor, open `/courses/<slug>/canvas`, and click Generate. This is the one step that exercises the encryption keys; a 500 here means they are missing or malformed. In a console, `LtiConsumerKey.last.secret` should decrypt.
4. End to end against the test Canvas (canvas.wikiedu.org): install the tool in a test course by hand, following `/lti/guide/instructors` with the production key and the production config URL. Launch as the instructor (expect auto-bind with no picker, then the identity approval), then as a test student (expect enrollment). This is exactly what testers will do, so it doubles as a rehearsal of the guide. Do it before any other signed launch with that key: the first accepted launch pins the key to that Canvas.
5. Do **not** run `staging_specs/lti11_launch_refusals_spec.rb` against production. It provisions and deletes courses through `DashboardAdminClient` over SSH. Its refusal matrix was covered on staging.
6. Afterwards: Sentry should show nothing from steps 1 to 4. In a console, `LtiConsumerKey.last` should have `lms_instance_guid`, `activated_at`, and `last_launch_at` set; its course's binding should have `lti_version` "1.2.0" and `ltiaas_service_credentials` nil.
7. Find where the refusal lines land. They log at info as `[LTI 1.1] refused launch: reason=... key_id=...`. On staging they land in Apache's error.log under Passenger rather than the Rails log; confirm the equivalent on production before relying on them for tester support.

### Inviting instructors

**Who qualifies:** an instructor on an approved Dashboard course (in a campaign, not withdrawn), whose Canvas lets teachers add apps to their own course (the default "LTI - add" permission; the guide's troubleshooting section covers the case where it is off), and whose institution cannot install the 1.3 tool.

**What to send each tester:**

- Their credentials URL, `https://dashboard.wikiedu.org/courses/<slug>/canvas`. It is unlisted; nothing in the interface links to it.
- The install guide, `https://dashboard.wikiedu.org/lti/guide/instructors`.
- What the beta does and does not do: launch, Wikipedia account connection, automatic enrollment, and the in-Canvas overview; no gradebook, no roster, no assignment import. Grading stays on the Dashboard.
- That the shared secret is shown once. If lost, regenerate, and regenerating breaks the existing Canvas install until the new pair is pasted in.
- That unused credentials expire after seven days. Until the UX item above is fixed, the page does not say when that has happened.

**What to ask them to report:** whether the Apps tab and "+ App" were available; whether the By URL install worked on the first try; whether the first launch connected the course without a picker; whether a student could connect and enroll; any error page, with the time, so it can be matched to the log; and their Canvas hostname. If the course has a TA or an observer, ask for one launch from each: neither has been exercised on a real 1.1 launch yet. A TA should get the instructor view; an observer should get the "can't enroll you" page.

**Staff side, per tester:**

- After their first launch, check the key row is pinned (`lms_instance_guid`, `activated_at`) and the binding is right (`lti_version` "1.2.0", course set, course flagged). There is no staff view yet; `LtiConsumerKey.includes(:course, :user).order(:created_at)` in a console is the record.
- A mistyped secret appears in the log as `reason=bad_signature` with their key id. An unknown key means they pasted something that is not a key; Sentry reports it, once per address per hour.
- Common fixes: mistyped secret, edit the app under Canvas Settings, Apps, and re-enter both values; lost secret, regenerate and re-enter; tool installed but no navigation item, check Settings, Navigation; an institution that later installs the 1.3 tool, staff must remove or re-point the 1.1 binding before the instructor relinks.
- Revoking: `key.update!(active: false)` refuses further launches with that key; `LtiLegacyLaunch.delete_all` invalidates every outstanding launch token (they live 24 hours). Deleting a course cascades to its key.

### Rollback

Set `lti_legacy_launches_enabled: 'false'` and restart. The endpoint, the XML, and the credentials page 404; launches still holding a minted token get a 403 and one Sentry event each (`LegacyLaunchesDisabledError`), expected noise for up to 24 hours. Tables and keys stay, so turning the flag back on restores every install.

### Not in this rollout

No rate limiting on the launch or key-issuing endpoints (see "Open questions" in the description); no staff view of issued keys; moving a linked course from 1.1 to 1.3 needs staff; TA and observer roles untested on a live 1.1 launch.

*Drafted in a Claude Code session (~1 h, 3 messages from the reviewer) that also ran a complexity-and-UX review pass of this PR. Sage asked for the checklist and read it before posting. The production-facing claims (flag state, route behavior) were checked with unauthenticated HTTP requests against dashboard.wikiedu.org; nothing was executed against production.*

Issue #7073, opened 2026-09-16
Cloned courses never get advice emails: clone drops the research_write_assignment tag


ragesoss:
## Summary

Cloned courses do not receive the course advice drip emails, because cloning drops the `research_write_assignment` tag that gates the entire sequence.

`ScheduleCourseAdviceEmails#schedule_emails` (`app/services/schedule_course_advice_emails.rb:10-23`) returns early without it:

```ruby
def schedule_emails
  return unless @course.tag?('research_write_assignment')

  schedule_biographies_email
  schedule_generative_ai_email
  # ...eight more
end
```

It is called from `ListCourseManager#send_approval_notification_emails` (`lib/list_course_manager.rb:60`) when a course is approved, a path cloned courses go through normally. So the early return is silent: no error, no email, ten scheduled messages simply never queued.

This is a consequence of the same root cause as #7072 — `CourseCloneManager::TAG_KEYS_TO_CARRY_OVER` (`lib/course_clone_manager.rb:130-137`) is matched against `tag.key`, but lists tag **values**. `research_write_assignment` is stored under the `index` key (3957 rows), so the guard never passes and the tag is not carried over. Filing separately because the user-visible effect is different and arguably more consequential: #7072 is a reporting/cohort problem, this one means instructors do not get mail they were meant to get.

## Evidence

ClassroomProgramCourses started 2023 or later that have a timeline:

| origin | courses | missing `research_write_assignment` |
|---|---|---|
| cloned | 958 | 933 (97%) |
| not cloned | 1777 | 65 (3.7%) |

The 3.7% baseline is courses that genuinely were not research-write assignments. The 97% is the bug.

## Also affected in the same service

Two further gates depend on tags that cloning drops for the same reason, so even a fix to the outer gate alone would leave these broken:

- `schedule_biographies_email` (line 28) checks `tag?('biographies')`, stored under the `topics-biographies` key. `topics-*` keys are not in `TAG_KEYS_TO_CARRY_OVER` at all.
- `schedule_learning_to_edit_email` region (line 87) checks `tag?('instructor_learner')`, stored under the `instructor_learner` key, also absent from the list.

## Questions for whoever picks this up

[...]

Issue #7072, opened 2026-09-16
Course cloning silently drops the sandboxes tag (TAG_KEYS_TO_CARRY_OVER matches values, not keys)


ragesoss:
## Summary

`CourseCloneManager::TAG_KEYS_TO_CARRY_OVER` (`lib/course_clone_manager.rb:130-137`) is matched against `tag.key`, but four of its six entries are tag **values**, not keys. Those four never match, so cloning silently drops those tags — most visibly the `sandboxes` tag, which records whether students draft in sandboxes or edit live articles.

```ruby
TAG_KEYS_TO_CARRY_OVER = %w[
  tricky_topic_areas          # ✅ really is a key
  working_individually        # ❌ a VALUE of the working_in_groups key
  working_in_groups           # ✅ works — happens to be both a key and a value
  research_write_assignment   # ❌ a VALUE, stored under the `index` key
  yes_sandboxes               # ❌ a VALUE, stored under the `sandboxes` key
  no_sandboxes                # ❌ a VALUE, stored under the `sandboxes` key
].freeze

def tag_course
  # ...
  @course.tags.each do |tag|
    next unless TAG_KEYS_TO_CARRY_OVER.include?(tag.key)   # <-- compares against key
    tag_manager.add(tag: tag.tag, key: tag.key)
  end
end
```

`WizardTimelineManager#add_tags` keys each tag by the **wizard panel key**, so the sandboxes panel (`config/wizard/researchwrite/wizard.yml`, `key: sandboxes`) stores `key: 'sandboxes'` with `tag: 'yes_sandboxes'` or `'no_sandboxes'`. `yes_sandboxes` / `no_sandboxes` are therefore never keys, and the guard never passes.

## Evidence

How these tags are actually stored (production data):

| tag value | stored under key | rows |
|---|---|---|
| `research_write_assignment` | `index` | 3957 |
| `working_individually` | `working_in_groups` | 3551 |
| `working_in_groups` | `working_in_groups` | 1913 |
| `no_sandboxes` | `sandboxes` | 443 |
| `yes_sandboxes` | `sandboxes` | 389 |

`yes_sandboxes`, `no_sandboxes`, `working_individually` and `research_write_assignment` do not occur as keys at all.

Downstream effect, for ClassroomProgramCourses started 2025 or later that have a timeline:

| origin | courses | missing `sandboxes` tag |
|---|---|---|
| cloned | 333 | 320 (96%) |
[...]

PR #7074, opened 2026-09-16
Let admins insert standard wizard blocks and switch sandbox mode

ragesoss:
## What this PR does

The assignment wizard runs once, at course creation, and then discards the answers it was given. After that there is no way back into its content: if an instructor changes their mind, or a block gets deleted, or a block's canonical copy is revised upstream, an admin has to rebuild it by hand in the block editor.

This adds two admin-only tools to timeline edit mode:

1. **A standard-block picker.** A second button in each week header opens a catalogue of every block the research-write wizard can build, grouped by wizard week, annotated with the conditions each block depends on and whether those match this course. Picking one inserts it into that week pre-filled with its title, type, content, training modules and points, open in edit mode so the admin reviews it before saving.
2. **A sandbox mode switcher.** The sandboxes decision is one choice with three consequences — the `no_sandboxes` course flag, the `sandboxes` tag, and several timeline blocks that exist in two variants — which is why doing it by hand goes wrong. This does all three in one transaction and reports what it changed. `courses.no_sandboxes_info` currently warns admins to "change the tags and timeline content too"; this is the thing that does that.

Along the way it also fixes a one-character typo in the wizard content (`no_sandoxes`, missing a `b`) that had silently suppressed the "Drafting a new article in a sandbox" block for every course ever created. That went to master separately as ea8fa15d.

Two related bugs were found but deliberately **not** fixed here, and are filed instead:

- Addresses neither, but reports both: #7072 (course cloning drops the `sandboxes` tag) and #7073 (the same root cause drops `research_write_assignment`, which silently skips the entire course advice email sequence for cloned courses).

### Design notes worth a reviewer's attention

- **Block identity.** Blocks in `config/wizard/*/content.yml` had no stable identifier, only position. This adds an `id` slug to each. It is a config-side identifier only — `WizardTimelineManager#save_block` strips it alongside `if`/`unless`/`graded`, so nothing reaches the `blocks` table and there is no migration.
- **Not everything is knowable.** Only wizard options carrying a `tag:` leave a record. Options with `logic:` alone — graded training, the early editing exercises, article choosing, the discussions, the supplementary assignments — are gone for good. `WizardLogicState` is therefore three-valued and answers `:unknown` rather than guessing, which is why the picker shows "Unknown" against some rows.
- **Evidence vs. record.** Four sandbox-dependent blocks also branch on individual-vs-group work, which is often not recorded. `TimelineGroupMode` recovers it from the timeline instead, by matching the training modules on the blocks the wizard already built. That is evidence rather than a record, so it is kept in its own class, consulted only when nothing was recorded, and reports nil rather than guessing when blocks disagree or have been hand-edited. Measured against a production dump, this covers about 8% of 2025+ courses that would otherwise be reported unresolved (20% of cloned courses).
- **Which blocks swap is derived from `content.yml`**, not hardcoded, so editing the wizard keeps the switcher correct.
- **No new write path.** Inserted blocks persist through the existing timeline save; `TimelineController#timeline_params` already permits every field needed.

## AI usage

Nearly all of the code, specs and commit messages in this PR were written by **Claude Code** (Opus 5), under human direction, in a single working session. The human contribution was direction and decisions, not implementation: framing the feature and its key use case, choosing between three design options the agent presented (how to identify blocks, how much of the sandbox switch to automate, and whether the picker should filter or annotate), deciding that the group-mode inference was worth building and that it had to be a separate single-responsibility class, cutting four proposed interface messages as unnecessary, and writing all of the user-facing message text.

**All user-facing strings were written by the human operator, not the AI**, per `docs/ai_guidelines.md`. The agent shipped them as `[PLACEHOLDER - description]` markers and they were filled in afterwards. The one exception is `application.confirm_generic`, which reuses a sentence already present twice in `config/locales/en.yml` rather than introducing new phrasing.

Three defects in the agent's own work were caught by tests rather than by review, and are described in the commit messages: a stale-association bug where `course.weeks.pluck(:id)` returned `[]` while `course.weeks.count` returned 8 (this would have been an intermittent production no-op, not just a test artifact); a heading-level accessibility regression caught by an existing `be_axe_clean` assertion; and two colour-contrast failures below WCAG AA that an axe check had initially been racing past on an empty modal.

**Scale of effort:** two commits, thirteen minutes apart, in one sitting on 2026-09-16, plus the preceding typo fix. That is a small amount of human time for a change of this size, and reviewers should calibrate accordingly — the test suite is thorough (83 Ruby examples and 7 browser feature specs, all new) but the design has had only one pair of human eyes on it so far. The areas most worth a careful human read are `lib/wizard_logic_state.rb` (what can and cannot be recovered) and `app/services/switch_course_sandbox_mode.rb` (what it does when it cannot tell).

This PR description was drafted using Claude Code and the `prepare-pr` skill, and like the rest of the agent's analysis above it may contain errors.

## Screenshots

**01 — timeline in edit mode.** Before: no admin controls. After: "Add Standard Block" in each week header, and a "Sandbox Mode" panel in the sidebar.

| Before | After |
|--------|-------|
| before | after |

**02 — the standard block picker.** Grouped by wizard week. Each row shows the block type, whether it matches this course, whether it is already in the timeline, and the raw logic keys it is gated on.

| After |
|-------|
| after |

**03 — filtered.** 45 blocks is a lot to scroll, so the list filters by title.

| After |
|-------|
| after |

**04 — an inserted block, open for review.** Pre-filled title, block type, and training modules, in edit mode so nothing is saved until the admin agrees.

| After |
|-------|
| after |

**05 — the same block, saved.**

| After |
|-------|
| after |

**06 — after switching sandbox mode.** The sidebar now reads "Editing live articles", and reports which blocks were removed and added.

| After |
|-------|
| after |

- **`courses.no_sandboxes_info` is now misleading.** The Overview toggle still only sets the flag, so its "Be sure to change the tags and timeline content too" warning remains true for that control — but an admin who has the new switcher should be told to use it instead. That message is not changed in this PR because it needs human-written copy.
- **Two tools now set `no_sandboxes`.** The Overview toggle (flag only) and this switcher (flag, tag and blocks). That is deliberate for now, but the toggle is arguably the one that should be retired or redirected.
- **Removal matches on title.** Existing blocks carry no reference back to the catalogue entry that built them, so the switcher matches by title and reports anything it cannot find rather than guessing. A block an admin has retitled will be reported, not removed. Adding a provenance column would make this exact, at the cost of a migration; that seemed like the wrong trade for a repair tool.
- **The picker does not use the group-mode inference.** It only informs, so showing "Unknown" is accurate; only the switcher acts, so only the switcher falls back. Easy to extend if reviewers disagree.
- **Two minor visual nits I did not want to churn the screenshots over:** "Already in timeline" renders in the same warning colour as "Unknown", though they signal different things; and "Add Standard Block" is visually quieter than the adjacent "Add Block". Both are deliberate-ish (secondary admin action) but worth a designer's opinion.
- **Scope.** Only the research-write wizard is offered in the picker. `WizardBlockCatalog` is wizard-agnostic and the other three wizards' content files have had `id` slugs added, so translation and add-images are close to a one-line follow-up.


ragesoss (commented):
Review focused on screen reader accessibility (JAWS in particular), with a general pass over the rest. This is AI-driven review feedback: it is meant to be read, understood and then used or ignored on the developer's own judgment, not trusted or acted on automatically.

The picker and the switcher are structurally sound (the picker is a named `role="dialog"` with `aria-modal`, the switcher is a named `role="group"`, match state is conveyed in text rather than colour alone, and the existing `be_axe_clean` checks pass). The problems are all in what happens *dynamically*: focus, announcements, and button naming. None of them are caught by axe, which audits a static snapshot.

Each finding below was confirmed by an exploratory feature spec (AI-written, left untracked at `spec/features/pr7074_screen_reader_review_spec.rb`); 13 of its 15 examples fail on the PR branch, and each failure is one of the items below.

### Screen reader / JAWS

1. **Focus is lost at every step of the picker.** `document.activeElement` is `<body>` after opening the picker, after Cancel, and after Add. The trigger button unmounts when the modal replaces it, so there is nothing to keep focus on. Chromium only treats an `aria-modal` dialog as modal while focus is *inside* it, so JAWS is never told a dialog opened, the background page stays in the accessibility tree, and the virtual cursor is left near where the button used to be. After Add, the new block appears at the bottom of the week in edit mode, silently. Suggested: focus the filter input (or the `h2` with `tabIndex={-1}`) on open; keep a ref to the trigger and restore focus to it on Cancel; on Add, focus the inserted block's title input. `ClaimSelectionPanel.jsx` and `confirm.jsx` already do the focus-on-open half of this.

2. **The sandbox switch result is never announced.** After OK in the confirm dialog, focus drops to `<body>` (the confirm's button unmounts), the request runs, and the report renders into a `<div>` with no live region. A JAWS user hears nothing and has no way to know whether blocks were removed or added, or that something was left unresolved. This is WCAG 2.1 SC 4.1.3 (Status Messages), which the published VPAT claims AA for. Suggested: render the `.sandbox-mode__report` container unconditionally with `role="status"` (a live region has to exist *before* its content changes to be announced), and move focus back to the switch button once `switching` clears.

3. **45 buttons are all named "Add".** In the JAWS buttons list (Insert+Ctrl+B), forms list, or when tabbing, every one reads as "Add button" with no indication of which block. No new copy is needed: give each row title an `id` and use `aria-labelledby="<button-id> <title-id>"` so the name becomes "Add Start drafting your contributions"; `aria-describedby` pointing at the match/conditions line would let JAWS read "Matches" or "Unknown" after the name.

4. **Type and match status run together.** The meta line's spans have no whitespace between them, so the text content is `AssignmentMatches`, `In ClassUnknown`, `HandoutsUnknown`, `AssignmentDoes not match`. CSS margin gives sighted users a gap; the accessibility tree gets one word. A `{' '}` between the spans, or a visually hidden separator, fixes it.

5. **Escape does not close the picker**, and the shared `Modal` has no focus trap, so Tab walks out into the page behind the dialog. The trap and Escape are pre-existing `Modal` gaps shared with Confirm and others, so probably a follow-up on `Modal` itself; Escape handling for the picker could be added locally now (as `ClaimSelectionPanel.jsx` does).

6. **The filter is silent.** Nothing announces how many blocks remain as the user types, and when nothing matches the panel is simply empty, with no message. Needs operator copy: a `role="status"` count line and an empty-state line (`[PLACEHOLDER]` markers until then).

7. **A disabled Add gives no reason.** For the two handouts rows the button is just unavailable; nothing in the UI says the content is generated. `WizardBlockCatalog` says they are listed "rather than hidden with no explanation", but the explanation never reaches the user. A short line (operator copy) with `aria-describedby` on the button would do.

8. **The confirmation names nothing.** JAWS announces the dialog as "Are you sure you want to do this? dialog, OK button". For an action that deletes and creates timeline blocks, and for a user who has just lost the visual context of which button they pressed, the operation should be named. `initiateConfirm` already accepts an `explanation` prop; it just needs copy.

9. **Pre-existing, but worth knowing:** the week header renders Add Block / Add Standard Block / Delete Week *before* the week's `h2` in DOM order. A JAWS user who jumps to "Week 3" with H and reads downward meets Week 4's buttons at the end of Week 3's content. A cheap mitigation is an `id` on the week `h2` and `aria-describedby` on the three buttons, so each reads "Add Standard Block button, Week 4 (…)".

### Other

10. **A failed request leaves the UI stuck.** `wizardBlocks` does not handle `API_FAIL`. If the catalog fetch fails, `loading` stays `true`: the spinner never clears and the picker will never refetch (`!loaded && !loading`). If the switch fails, `switching` stays `true` and the button is disabled until reload.

11. **The catalog is fetched once per session but annotated per course state.** `match` and `in_timeline` go stale after a sandbox switch, after saving an inserted block, and when navigating to another course's timeline in the same SPA session (so "Matches" / "Already in timeline" can describe a different course). `lastSwitch` likewise persists across courses, so course B's sidebar can show course A's report. Refetching on every open is cheap, and `lastSwitch` could be keyed by slug or cleared on course change.

12. Minor: `@course.save` and `Block.create` inside the service transaction are non-bang, so a validation failure would be silent; `working_in_groups` also writes `retain_available_articles: true` (`FLAG_LOGIC`) but `WizardLogicState` does not use that as evidence, which may be deliberate; and the wizard offers 0/1/2 peer reviewers while `FLAG_LOGIC` and `content.yml` cover 1/2/3, so "Peer review three articles" is insertable from the picker but unreachable from the wizard (pre-existing drift, not introduced here).

### Verified locally

- `yarn build` clean; RuboCop and ESLint clean on all changed files.
- PR specs: 83 Ruby examples and 7 feature examples pass (plus the existing `wizard_timeline_manager_spec`).
- The one failed CI run is two `category_scopes_spec` failures, unrelated; the sibling run at the same SHA passed.
- Exploratory spec: 15 examples, 13 failures, each mapping to items 1 to 8 above.

*Drafted in a Claude Code session (~30 min, 2 interactions). Sage directed the review toward screen reader / JAWS accessibility and approved posting after reading the session's summary of the findings; the full comment text was drafted by the agent. The findings and the exploratory spec are AI-generated. Sage did not independently run the exploratory spec, and no JAWS session was run: the JAWS behaviour described is inferred from the DOM and focus state the spec measures.*

PR #7077, opened 2026-09-17
Fix minor asymmetries in course wiki timeslices

gabina:
## What this PR does

On the ACUWT path, `CourseWikiTimeslice` and `CourseUserWikiTimeslice` caches are
re-derived from ACUWT rows rather than from raw revisions. ACUWT rows are written for
every fetched revision regardless of namespace, and both aggregators excluded articles
only via `articles_courses.not_tracked` — which misses articles that have no
`articles_courses` record at all, as happens for any namespace the course doesn't
track. Mainspace edits in a course that doesn't track mainspace were therefore counted
in `character_sum` / `references_count`, inflating course words added and the editor
tab's chars added.

Both aggregators now filter positively on articles that have a tracked
`articles_courses` record, matching the legacy path's
`live_revisions_in_tracked_namespaces`. `character_sum_us` and `character_sum_draft`
stay unfiltered, also matching legacy.

Follow-up work from the ongoing ACT-to-ACUWT migration; this PR contains only the two
aggregation fixes.

## AI usage

Claude Code (Claude Opus 5) wrote both commits and their messages under the user's
direction. The user found the bug from a failing spec, reproduced it locally against a
paired ACUWT / default-branch course, and made the design calls — including deciding
not to loosen `character_sum_ms` to match the editor tab's user/draft columns, once it
turned out to feed course- and campaign-level aggregates. Claude Code traced the root
cause and drafted the fixes; the user confirmed the trace with its own debug run and
corrected several wrong intermediate claims along the way. Two commits in one session
on 2026-09-17, roughly an hour apart. This description was drafted with Claude Code and
the `prepare-pr` skill, and may contain errors.

## Screenshots

No UI changes. The effect is visible as corrected numbers in the editor tab's chars
added column and in course words added, for courses that don't track mainspace.

- `update_revision_count_from_acuwt` has the same negative-filter gap, but the legacy
  path may share it, so it was left out pending a comparison of `Course#revision_count`
  between the two paths.
- The two paths still differ in how `character_sum_us` / `character_sum_draft` are
  filtered. Not addressed here.


gabina:

ragesoss:
This review was produced with AI assistance. It is offered as something to read and weigh, not as a verdict to act on automatically — some of it may be wrong, and the parts that matter are worth confirming independently before changing anything.

## The fix reproduces and holds up

An exploratory spec written during review builds a draft-only course (tracked namespaces set via `course_wiki_namespaces`, so the mainspace article genuinely has no `articles_courses` record, matching what `ArticlesCourses.update_from_course_revisions` produces) alongside a control course that does track mainspace. Against `master` three examples fail — `character_sum` 4213, `references_count` 7, `character_sum_ms` 4213 — and all seven pass on this branch. The described leak is real and both hunks close it.

One detail that explains why this went unnoticed: a CWT/CUWT parity check (CWT `character_sum` vs the sum of student `character_sum_ms`) passes on `master` too. Both paths were inflated by the same amount, so the invariant held while the numbers were wrong.

## Both open questions in the description appear to resolve in the PR's favour

**`update_revision_count_from_acuwt`.** The deferral looks correct: the legacy path does share the gap. `revision.scoped` is set by `Course#scoped_article?` via `mark_scoped_articles` (`lib/revision_data_manager.rb:194`), which concerns only-scoped-article courses rather than namespaces, so the legacy `@revisions` carry every namespace and `CourseWikiTimeslice#update_revision_count` rejects only `not_tracked_article_ids`. Both paths over-count identically. The existing example "includes non-mainspace articles in revision_count" (expecting 9) already pins that as current behaviour on both sides.

**`character_sum_us` / `character_sum_draft`.** Confirmed to match legacy: `update_character_sum` passes `@liverevisions` (negative filter only) for US and DRAFT, and `tracked_namespace_revisions` (positive filter) only for MS. Leaving those two columns unfiltered mirrors that exactly. Separately, `article_course_user_wiki_timeslice.rb:83` applies the same non-negative-characters rule as the legacy `character_sum` helper, so the two paths' figures are genuinely comparable.

## Observation worth checking before merge: the CWT filter inverts a nearly-free predicate into an unbounded one

Rails generates `WHERE 1=1` for `where.not(article_id: [])` and `WHERE 1=0` for `where(article_id: [])`. Measured against a production-copy database (15.2M `articles_courses` rows):

| | old `not_tracked_article_ids` | new `tracked_article_ids` |
|---|---|---|
| Empty for | 14,656 of 15,399 courses (95%) — a free no-op | never, in practice |
| Largest per course | 4,667 | **1,053,802** (course 11168) |

For that course the pluck returns roughly 9.5 MB of raw ids, so the `IN (…)` literal is about 9.5 MB of SQL text. `acuwt_mainspace_tracked_student_records` memoises a relation rather than a result, so `.sum(:character_sum)` and `.sum(:references_count)` each execute it — two such statements. Local MariaDB 11.8 has `max_allowed_packet` at 16 MB, putting that course near 60% of the limit; the production value has not been checked here. Article-scoped courses append a second `IN` list on top.

The memoisation also does not span the aggregation loop: `CourseWikiTimeslice.update_from_acuwt` calls `find_by!(...)`, producing a fresh instance per timeslice, so this is one full pluck per timeslice — two, since `not_tracked_article_ids` is still plucked separately for `update_revision_count_from_acuwt`. `CourseUserWikiTimeslice.update_from_acuwt` reloads `course` per row through `find_or_create_by`, so `course.article_ids` re-queries per timeslice per user as well.

Scoping this honestly: no course in the database copy examined has `use_acuwt` set, so nothing is affected today. This reads as a scaling concern for the migration rather than a live problem, and `CopyCourse#force_use_acuwt` means the ACUWT population grows over time.

If it seems worth addressing, expressing the filter without materialising the id list would avoid the question entirely — `.where(article_id: course.articles_courses.tracked.select(:article_id))`, or an INNER JOIN on `articles_courses` (the unique index on `(course_id, article_id)` supports it). Whether the subquery form actually performs better on MariaDB is worth measuring rather than assuming. For `CourseUserWikiTimeslice`, bounding the lookup to the article ids present in the timeslice would be strictly cheaper than fetching `course.article_ids`.

## Minor: the two hunks use different definitions of "tracked"

`CourseWikiTimeslice` filters on `articles_courses.tracked`; `CourseUserWikiTimeslice` uses `course.article_ids`, which spans tracked and untracked rows alike. The latter is correct only because `excluded_article_ids` removed the untracked rows on the preceding line, and it does mirror the legacy path — but `mainspace_records_in_tracked_namespaces` reads as self-contained, and its dependence on that earlier line is not visible at the call site.

That difference also produces one real divergence: `where(tracked: true)` and `where(tracked: false)` both exclude NULL, so a NULL-`tracked` row would now be dropped by CWT while still counted by CUWT — the same parity break this PR fixes, in a different corner. In practice this appears unreachable: the database copy examined has 0 NULL rows out of 15.2M. (`ArticlesController#update_tracked_status` passes `params[:tracked]` straight into `update` without a guard, which is how one could arise, but that is a separate matter from this PR.)

## No regression guard ships with the fix

The PR contains no spec changes. All 148 existing examples across `course_wiki_timeslice`, `course_user_wiki_timeslice`, `article_course_timeslice`, `article_course_user_wiki_timeslice`, `update_course_wiki_timeslices`, `update_timeslices_untracked_article`, `update_timeslices_course_user`, `timeslice_cleaner` and `articles_courses_cleaner` pass on `master` as well as on this branch — the existing suite covers `tracked: false` but never the missing-`articles_courses`-record case. Reverting either hunk would currently be invisible to CI.

The exploratory spec below was written by Claude Code during this review and is offered for adoption if useful. It has not been reviewed line by line by a human; it was run against both branches, with the results reported above.

<details>
<summary>spec/models/acuwt_untracked_namespace_aggregation_spec.rb</summary>

```ruby
# frozen_string_literal: true

# Exploratory spec written during review of PR #7077.
#
# Reproduces the bug the PR fixes: on the ACUWT path, a course that does not track
# mainspace still counted its students' mainspace edits in character_sum /
# character_sum_ms / references_count, because both aggregators excluded articles
# only via a negative filter on `articles_courses.not_tracked`. Articles outside
# the course's tracked namespaces never get an articles_courses record at all, so
# the negative filter never saw them.
#
# Each example contrasts an untracked-namespace course against an otherwise
# identical course that does track mainspace.

require 'rails_helper'

describe 'ACUWT aggregation for untracked-namespace articles', type: :model do
  before { stub_const('TimesliceManager::TIMESLICE_DURATION', 86400) }

  let(:start) { 10.days.ago.beginning_of_day }
  let(:ts_end) { start + 1.day }
  let(:student) { create(:user, username: 'Student') }

  # Production-realistic magnitudes: a mainspace article edit big enough that its
  # leakage into the aggregates is unmistakable, and a smaller draft edit.
  let(:mainspace_chars) { 4213 }
  let(:mainspace_refs) { 7 }
  let(:draft_chars) { 812 }

  let(:mainspace_article) do
    create(:article, title: 'Sourdough starter', wiki:,
                     namespace: Article::Namespaces::MAINSPACE)
  end
  let(:draft_article) do
    create(:article, title: 'Sourdough hydration', wiki:,
                     namespace: Article::Namespaces::DRAFT)
  end

  # Builds a course on the ACUWT path with one student who edited both a mainspace
  # article and a draft article in the same timeslice. `tracked_namespaces` controls
  # which namespaces the course tracks, which in turn controls which articles get an
  # articles_courses record — the real mechanism behind the bug.
  def build_course(slug:, tracked_namespaces:)
    course = create(:course, slug:, start: 1.month.ago, end: 1.month.from_now,
                             home_wiki: wiki)
    courses_wiki = course.courses_wikis.find_by(wiki:)
    tracked_namespaces.each do |ns|
      create(:course_wiki_namespaces, courses_wikis: courses_wiki, namespace: ns)
    end
    create(:courses_user, course:, user: student, role: CoursesUsers::Roles::STUDENT_ROLE)
    TimesliceManager.new(course).create_timeslices_for_new_course_wiki_records([wiki])
    course
  end

  # Mirrors what ArticlesCourses.update_from_course_revisions does: an articles_courses
  # record is created only for articles in the course's tracked namespaces.
  def seed_acuwt(course)
    course.tracked_namespaces.each do |wiki_ns|
      article = [mainspace_article, draft_article].find { |a| a.namespace == wiki_ns[:namespace] }
      create(:articles_course, course:, article:) if article
    end
    create(:article_course_user_wiki_timeslice, course:, wiki:, article: mainspace_article,
           user_id: student.id, start:, end: ts_end, character_sum: mainspace_chars,
           references_count: mainspace_refs, revision_count: 3)
    create(:article_course_user_wiki_timeslice, course:, wiki:, article: draft_article,
           user_id: student.id, start:, end: ts_end, character_sum: draft_chars,
           references_count: 0, revision_count: 2)
  end

  def aggregate(course)
    CourseUserWikiTimeslice.update_from_acuwt(course, student.id, wiki, start, ts_end)
    CourseWikiTimeslice.update_from_acuwt(course, wiki, start, ts_end)
  end

  # Tracks DRAFT only, so the student's mainspace edit has no articles_courses record.
  let(:draft_only_course) do
    build_course(slug: 'School/Draft_only_(2026)',
                 tracked_namespaces: [Article::Namespaces::DRAFT])
  end

  # Control: tracks mainspace as well, so the mainspace edit does have a record.
  let(:mainspace_course) do
    build_course(slug: 'School/Tracks_mainspace_(2026)',
                 tracked_namespaces: [Article::Namespaces::MAINSPACE,
                                      Article::Namespaces::DRAFT])
  end

  describe 'CourseWikiTimeslice#update_cache_from_acuwt' do
    it 'excludes mainspace edits from character_sum when the course does not track mainspace' do
      seed_acuwt(draft_only_course)
      aggregate(draft_only_course)

      cwt = CourseWikiTimeslice.find_by(course: draft_only_course, wiki:, start:)
      expect(cwt.character_sum).to eq(0)
    end

    it 'excludes mainspace edits from references_count when the course does not track ' \
       'mainspace' do
      seed_acuwt(draft_only_course)
      aggregate(draft_only_course)

      cwt = CourseWikiTimeslice.find_by(course: draft_only_course, wiki:, start:)
      expect(cwt.references_count).to eq(0)
    end

    it 'still counts mainspace edits when the course does track mainspace' do
      seed_acuwt(mainspace_course)
      aggregate(mainspace_course)

      cwt = CourseWikiTimeslice.find_by(course: mainspace_course, wiki:, start:)
      expect(cwt.character_sum).to eq(mainspace_chars)
      expect(cwt.references_count).to eq(mainspace_refs)
    end
  end

  describe 'CourseUserWikiTimeslice#update_cache_from_acuwt' do
    it 'excludes mainspace edits from character_sum_ms when the course does not track ' \
       'mainspace' do
      seed_acuwt(draft_only_course)
      aggregate(draft_only_course)

      cuwt = CourseUserWikiTimeslice.find_by(course: draft_only_course, wiki:,
                                             user_id: student.id, start:)
      expect(cuwt.character_sum_ms).to eq(0)
      expect(cuwt.references_count).to eq(0)
    end

    it 'still counts mainspace edits when the course does track mainspace' do
      seed_acuwt(mainspace_course)
      aggregate(mainspace_course)

      cuwt = CourseUserWikiTimeslice.find_by(course: mainspace_course, wiki:,
                                             user_id: student.id, start:)
      expect(cuwt.character_sum_ms).to eq(mainspace_chars)
      expect(cuwt.references_count).to eq(mainspace_refs)
    end

    # Deliberate design decision recorded in the PR description: character_sum_us and
    # character_sum_draft stay unfiltered on the ACUWT path, matching the legacy path.
    it 'still counts draft edits that have no articles_courses record' do
      draft_only_course
      # Drop the draft article's articles_courses record so neither namespace is backed
      # by one, isolating the draft column from the tracked-article filter.
      seed_acuwt(draft_only_course)
      ArticlesCourses.where(course: draft_only_course).destroy_all
      aggregate(draft_only_course)

      cuwt = CourseUserWikiTimeslice.find_by(course: draft_only_course, wiki:,
                                             user_id: student.id, start:)
      expect(cuwt.character_sum_draft).to eq(draft_chars)
    end
  end

  describe 'CWT / CUWT parity' do
    it 'keeps CWT character_sum equal to the sum of student CUWT character_sum_ms' do
      seed_acuwt(draft_only_course)
      aggregate(draft_only_course)

      cwt = CourseWikiTimeslice.find_by(course: draft_only_course, wiki:, start:)
      cuwt_total = CourseUserWikiTimeslice.where(course: draft_only_course, wiki:, start:)
                                          .sum(:character_sum_ms)
      expect(cwt.character_sum).to eq(cuwt_total)
    end
  end
end
```

</details>

RuboCop is clean on both changed files and on the spec above.

*Drafted in a Claude Code session (~15 minutes, 3 interactions). Sage selected the PR, asked for a full review, and approved posting this as a comment; they did not independently verify the findings or read the exploratory spec. Verified locally by Claude Code: the inflated sums were reproduced on `master` and confirmed fixed on this branch using the spec above; 148 existing timeslice examples were run green on both branches; the tracked/untracked id-list sizes were measured against a production-copy database; and the SQL Rails emits for empty arrays was checked directly.*

gabina:

ragesoss (approved):
Re-review of the two new commits (`ce7b2ca00e`, `8f87fef2b1`), produced with AI assistance. As before, treat it as something to weigh rather than act on automatically.

**Earlier feedback is addressed.**

- **CWT query:** the `IN (SELECT …)` subquery doesn't build the id list. `EXPLAIN` against a production-copy database for course 11168 (1,053,802 `articles_courses` rows) shows MariaDB looking up each ACUWT row through the unique `(course_id, article_id)` index. It never builds a 1M-row intermediate result. The ACUWT table had no rows for that course locally, so this confirms how the query runs but not how fast it is on a populated table.
- **CUWT lookup:** now limited to the mainspace ids present in the timeslice, which fixes the per-user cost.
- **Regression specs:** each new spec guards its own fix. Reverting only the CWT change makes the CWT example fail (`character_sum` 4913 vs 700). Reverting only the CUWT change makes the CUWT example fail (`character_sum_ms` 4313 vs 100).

Still open, minor and optional: CWT filters on `articles_courses.tracked`, while CUWT accepts any `articles_courses` row and relies on the earlier `excluded_article_ids` step. The new method comment covers the reasoning. The two could only disagree for a NULL `tracked` value, and none exist in the data examined.

RuboCop is clean on the changed files, and 144 timeslice-related examples pass with current `master` merged in.

*Drafted in a Claude Code session (~5 min, 1 interaction). Sage asked for the re-review and approved posting; they did not independently verify the findings. Claude Code checked locally: ran the specs with each fix reverted in turn, ran the timeslice specs and RuboCop on the review branch, and ran `EXPLAIN` on the new CWT query against a production-copy database.*

PR #7079, opened 2026-09-18
Refactor ArticlesCoursesCacheManager

gabina:
## What this PR does

Part of the ongoing migration from `article_course_timeslices` (ACT) to
`article_course_user_wiki_timeslices` (ACUWT). `ArticlesCoursesCacheManager` is
now course-scoped — it takes a course alongside the `articles_courses` relation,
so it no longer groups records by `course_id` — and for courses with the
`use_acuwt` flag, both halves of the `ArticlesCourses` cache update now read
ACUWT instead of ACT: the stats themselves, and `articles_courses_to_update`,
which selects which records need refreshing. Those two have to move together —
computing from one table while selecting from the other is how stale caches
would slip through.

The values are the same either way: this is exactly what
`ArticleCourseTimeslice.act_stats_from_acuwt` already aggregates from ACUWT, with
the same sums and the same `revision_count > 0` rule for `user_ids`. The point of
deploying it is to measure the ACUWT path against the ACT path on the same
production course, since this cache phase dominates updates on large courses.

## AI usage

Claude Code (Opus 5) wrote the code, the specs, and the commit messages, under
the user's direction. The user chose the approach — course-scope the manager
first, then gate the table switch on `use_acuwt?`, split across separate commits —
reviewed each step, and ran the production queries that established the
`articles_courses_to_update` change was safe to include. The summary above was
also drafted by Claude Code and may contain errors.

Four commits within a single ~50-minute session, continuing from an earlier
session's investigation of the cache phase's cost.

## Screenshots

No UI changes.

The performance question is open by design — that is what this PR is for. The
first flagged production course is article-scoped and its cache phase touches
only ~8.5k of its 1.47M ACUWT rows, so it will confirm correctness but not cost;
a larger flagged course was created for the actual benchmark. No index changes
are proposed until those numbers exist. The `articles_courses_to_update` query
was timed on that course against both tables and is fast either way, even though
neither has an index on `updated_at`.

Other ACT-backed calculations are untouched and still correct, since
`bulk_update_from_acuwt` keeps ACT populated for flagged courses. Notably
`Course#tracked_article_course_timeslices` and `#scoped_article_timeslices`,
which feed the analytics and CSV builders, are handled on a separate branch.


ragesoss (changes requested):
Review notes from a Claude Code session. AI-driven review feedback isn't meant to be trusted automatically or acted on blindly — it's worth reading and understanding, then using or ignoring based on your own judgment. Anything below not marked as verified locally is provisional and may be wrong.

## 1. The selection switch drops the `touch_timeslices` signal

`ArticlesCoursesCleaner#touch_timeslices` (`lib/articles_courses_cleaner.rb:225`) exists solely to bump `ArticleCourseTimeslice.updated_at` so that the affected `ArticlesCourses` caches get recomputed on the next update. It runs when a course's start or end date moves and the now-out-of-range timeslices are deleted (`lib/articles_courses_cleaner.rb:86` and `:104`, reached from `UpdateTimeslicesCourseDate`).

There is no ACUWT equivalent. The ACUWT deletions go through `delete_in_batches`, which leaves the surviving rows' `updated_at` untouched. So once `articles_with_updated_timeslices` reads ACUWT for flagged courses, a date change no longer selects those articles, and their cached `character_sum` keeps counting the deleted period until something else happens to touch a timeslice for that article.

This was reproduced locally: the spec below passes on master and fails on this branch.

```
1) selects the article whose out-of-range timeslices were just deleted
   expected: [41]
        got: []

2) recomputes the cache so the deleted period no longer counts
   expected: 1000
        got: 1100
```

<details>
<summary>Spec used (AI-written, run on both branches, not reviewed line-by-line by a human)</summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require_dependency "#{Rails.root}/lib/timeslice_cleaner"
require_dependency "#{Rails.root}/lib/articles_courses_cleaner"

describe ArticlesCourses do
  let(:course) do
    create(:course, start: '2024-07-01'.to_datetime, end: '2024-08-01'.to_datetime)
  end
  let(:article) { create(:article) }
  let(:last_update) { 1.hour.ago }

  before do
    create(:articles_course, article:, course:, character_sum: 1100)
    # A timeslice before the new start date, and one after it.
    create(:article_course_user_wiki_timeslice, article:, course:, wiki:, user_id: 2,
           start: '2024-06-20', end: '2024-06-21', revision_count: 1, character_sum: 100)
    create(:article_course_user_wiki_timeslice, article:, course:, wiki:, user_id: 2,
           start: '2024-07-20', end: '2024-07-21', revision_count: 1, character_sum: 1000)
    create(:article_course_timeslice, article:, course:,
           start: '2024-06-20', end: '2024-06-21', character_sum: 100, user_ids: [2])
    create(:article_course_timeslice, article:, course:,
           start: '2024-07-20', end: '2024-07-21', character_sum: 1000, user_ids: [2])

    # Everything predates the last course update.
    ArticleCourseUserWikiTimeslice.update_all(updated_at: 2.hours.ago)
    ArticleCourseTimeslice.update_all(updated_at: 2.hours.ago)
    course.flags['update_logs'] = { 1 => { 'end_time' => last_update } }
    course.save

    # The course start date moves forward past the first timeslice, and the
    # date-change cleanup runs, exactly as UpdateTimeslicesCourseDate does it.
    course.update(start: '2024-07-05'.to_datetime)
    TimesliceCleaner.new(course).delete_article_course_user_wiki_timeslices_prior_to_start_date
    ArticlesCoursesCleaner.clean_articles_courses_prior_to_course_start(course)
  end

  it 'selects the article whose out-of-range timeslices were just deleted' do
    expect(described_class.articles_courses_to_update(course)).to eq([article.id])
  end

  it 'recomputes the cache so the deleted period no longer counts' do
    described_class.update_required_caches_from_timeslices(course)
    expect(described_class.find_by(course:).character_sum).to eq(1000)
  end
end
```

</details>

The stated rationale for moving the selection — that computing from one table while selecting from the other is how stale caches slip through — holds in general. But in this specific direction the ACT signal is a superset: `bulk_update_from_acuwt` rewrites `updated_at` on every reaggregation, so ACT is bumped everywhere ACUWT is, plus in the date-change case where only ACT is touched. An ACUWT `touch_all` alongside the existing ACT one in `touch_timeslices`, gated on `use_acuwt?`, would close the gap if the selection switch is kept.

## 2. The index premise in the description

The description states the query "is fast either way, even though neither has an index on `updated_at`." ACT has had one since the table was created: `article_course_timeslice_by_updated_at` on `(course_id, updated_at, article_id)` (`db/schema.rb:85`, from `db/migrate/20240522233334_create_article_course_timeslices.rb`). That index covers the whole query, so the ACT version is an index-only range scan.

ACUWT's three indexes (`db/schema.rb:105-107`) contain no `updated_at`, so the best available plan is a `course_id`-prefix scan of `index_acuwt_unique` followed by a row read per entry to test `updated_at`. On the article-scoped course described in the PR the absolute numbers may still be small, but the two timings were not measuring comparable plans, which is worth knowing before the benchmark course's numbers are interpreted.

## 3. The manager now assumes a course-scoped relation without enforcing it

`write_caches` writes `course_id: @course.id` for every row and looks stats up by `article_id` alone. If the relation ever contains another course's records, those rows get `EMPTY_CACHE` written — their caches zeroed — where the previous signature handled mixed-course relations correctly. The existing spec at `spec/lib/articles_courses_cache_manager_spec.rb:131` had to be rewritten for exactly this reason.

The only production caller is scoped, so this is latent rather than live. Scoping defensively in the initializer (`articles_courses.where(course:)`), or having the manager take just the course and derive the relation, would make the invariant structural rather than conventional.

## Checked and found consistent

- The ACUWT and ACT aggregations are equivalent: sums, `MAX(new_article)`, `MIN(first_revision)`, and the `revision_count > 0` rule for `user_ids` all match what `act_stats_from_acuwt` produces, and the `.distinct` pluck gives the same de-duplication the legacy `flat_map(&:last).uniq` did.
- `tracked` is ignored by both paths, consistently — `act_records_from_acuwt` doesn't filter on it either.
- `reset_excluded`, `reset_included`, and `reset_timeslices_for_update_from_acuwt` don't open a comparable staleness gap: each either deletes the `articles_courses` record or ends in a path that rewrites ACUWT and bumps its `updated_at`.
- RuboCop is clean on all changed files. `spec/lib/articles_courses_cache_manager_spec.rb`, `spec/models/articles_courses_spec.rb`, `spec/presenters/individual_statistics_timeslice_presenter_spec.rb`, `spec/lib/articles_courses_cleaner_spec.rb`, `spec/lib/timeslice_cleaner_spec.rb`, `spec/services/update_timeslices_course_date_spec.rb` and `spec/services/update_course_stats_spec.rb` all pass on the branch merged with master (87 examples).

*Drafted in a Claude Code session (~20 min, 2 interactions). The session read the diff and the surrounding ACUWT/ACT code paths, ran the affected specs and RuboCop locally on the branch, and reproduced finding 1 by running the AI-written spec above against both master and the branch. Sage approved posting after a read-through but did not independently re-verify the findings.*

PR #7080, opened 2026-09-21
Keep one course's failure from aborting the whole scheduler pass

gabina:
## What this PR does

`ScheduleCourseUpdates` did all of its per-course work inline in one loop, so a single
course raising ended the whole pass and every course after it in the iteration was never
enqueued. On `peony-web` this showed up as `ScheduleCourseUpdatesWorker` failing every five
minutes with `Locking a record with unpersisted changes is not supported. Changed
attributes: "flags". (RuntimeError)`, raised from `Course#add_flag` via
`course_queue_sorting.rb:76`, and as specific courses silently never reaching a sidekiq
queue.

This PR moves the loop body into `enqueue_course_update`, rescues `StandardError` per
course, and adds `report_enqueue_failure`, which logs the failure and reports it to Sentry
with the course id, slug and `course.changed`. One bad course can no longer starve the
rest of the pass, and the next occurrence will identify the offending row.

This is a containment fix, not a root-cause fix. Why a course arrives at the scheduler with
an unsaved change to its serialized `flags` column is still unknown — the extra Sentry
context added here is what should finally pin it down. `Course#add_flag` is deliberately
left raising rather than reloading: a silent `reload` would discard a caller's pending
changes, which is the silent-flag-loss failure mode that 37c8338c2 was written to prevent.

The existing `'reports logs to sentry even when it errors out'` example triggered the
top-level rescue through `CourseDataUpdateWorker.update_course`, which the new per-course
rescue now swallows. It raises from `Sidekiq::WorkSet.new` instead, which runs outside the
loop, and asserts the specific failure message rather than a bare call count.

## AI usage

Claude Code (Opus 5) was used for both the investigation and the implementation, under the
user's direction and with the user supplying every piece of production evidence.

The session began as a production diagnostic: courses were not being enqueued and the
initial suspicion was an orphaned `sidekiq-unique-jobs` lock in redis. Claude Code proposed
redis and console commands to confirm or rule that out; the user ran them and reported back
that no lock existed. It then proposed two further explanations — a course failing
validation so that `save` returned false inside `add_flag`, and a YAML round-trip mismatch
on the serialized `flags` column — and produced a SQL query and console checks to test them.
The user ran those too; the SQL returned nothing and the console check came back empty,
which ruled both out. Claude Code also disproved part of its own YAML theory with a local
round-trip test. The Sentry traceback the user then supplied identified the exact call site
and settled the diagnosis. Along the way Claude Code gave one wrong suggestion about which
log to grep, and a SQL query the user rejected as broken; both were corrected.

Claude Code wrote the implementation, the two new specs and the retargeted existing spec,
and drafted the commit message. The user made the design decision to contain the failure in
the scheduler rather than change `Course#add_flag`, after Claude Code laid out the tradeoff,
and ran all tests: one failure (an over-loose `have_received` count assertion) before a
clean pass.

The "What this PR does" summary above was also drafted by Claude Code and may contain
errors — in particular, the root cause of the dirty `flags` column remains unverified.

Scale of effort: one commit, written in a single session on 2026-09-21, following an
extended diagnostic conversation of roughly fifteen short user messages.

## Screenshots

No UI changes.

- **The root cause is still unknown.** Something leaves a `Course` with an unsaved change to
  its `flags` column at load time. Neither invalid records nor an unstable YAML round-trip
  reproduced it in checks against production, and by the time those checks ran the condition
  had cleared — any successful save of the course rewrites `flags` canonically and heals it.
  The Sentry `extra` added here should name the row the next time it happens.
- **`Course#add_flag` still violates its own stated contract.** 37c8338c2 intended that "a
  single invalid course cannot abort the scheduler loop", but `with_lock` raises on a dirty
  receiver, so it can. This PR works around that at the call site rather than changing
  `add_flag`, since the safe-looking fix (`reload if has_changes_to_save?`) would silently
  drop a caller's unsaved changes. Worth revisiting once the root cause is known.
- **Enqueue failures are now non-fatal too.** The per-course rescue also swallows failures
  from `CourseDataUpdateWorker.update_course`. If redis were unavailable, the pass would now
  report every course to Sentry and complete rather than failing loudly. That seemed an
  acceptable trade for not starving courses, but it is a deliberate behavior change worth a
  reviewer's opinion.


ragesoss:
A note on how to take this: it is AI-assisted review output, not a verdict. Worth reading and understanding, then using or ignoring on your own judgment.

## The dirty `flags` column has a root cause now

It is a *stale* serialization, not an unstable one — which is why the round-trip check in the diagnostic session came back clean. Psych used to emit a nil value as `key: `, with a trailing space, and now emits `key:`. Rails decides a serialized column changed in place by re-dumping the deserialized value and comparing bytes against what came out of the database (`ActiveRecord::Type::Serialized#changed_in_place?`), so any course holding a nil flag value whose row has not been saved since that Psych change reports itself dirty the moment it loads. `Course#add_flag` then raises before writing anything.

Measured on the two affected peony rows:

```
15518: stored 1709 B, fresh 1705 B, 4 nil-valued flags
18300: stored 2544 B, fresh 2542 B, 2 nil-valued flags

first divergence, both rows:
  stored: "academic_system: "
  fresh:  "academic_system:"
```

Exactly one byte per nil. Both rows report stable-after-rewrite, so re-storing the value repairs them permanently.

This also explains the console check that came back empty. `changed` and `will_save_change_to_flags?` report nothing for a serialized column unless the attribute has been read first — in-place comparison only engages once `has_been_read?` is true. A scan that never touches `course.flags` finds every row clean. Any future check needs the read.

And it means the condition is a backlog rather than a one-off: old courses carry pre-upgrade bytes until something saves them, and each surfaces as it is marked `needs_update` and becomes active again.

## The new specs do not exercise the path that fails

Both new examples raise from `CourseDataUpdateWorker.update_course`, which runs *after* `queue_for`. The production failure is raised *inside* `queue_for` → `update_longest_update_time` → `Course#add_flag`, before the worker is reached — the other half of `enqueue_course_update`. The containment holds either way, but the regression risk on that first half is currently uncovered.

The spec below builds a genuinely stale row by mutating the canonical bytes the way Psych did. It passes on this branch and fails on master with the production traceback (`course.rb:506` → `course_queue_sorting.rb:70` → `schedule_course_updates.rb:49`).

<details>
<summary><code>spec/lib/data_cycle/stale_flags_scheduler_spec.rb</code></summary>

```ruby
# frozen_string_literal: true

require 'rails_helper'
require "#{Rails.root}/lib/data_cycle/schedule_course_updates"

describe ScheduleCourseUpdates do
  let(:seven_second_log) do
    { 'update_logs' => { 1 => { 'start_time' => Time.zone.parse('2026-09-08 19:00:00'),
                                'end_time' => Time.zone.parse('2026-09-08 19:00:07') } } }
  end

  # Psych used to emit a nil value as "key: " and now emits "key:". Mutating the
  # canonical bytes that way is exactly what the stale production rows look like.
  def make_flags_stale(course)
    canonical = Course.connection.select_value("SELECT flags FROM courses WHERE id = #{course.id}")
    stale = canonical.sub("academic_system:\n", "academic_system: \n")
    raise 'canonical flags did not contain a nil academic_system' if stale == canonical
    Course.connection.exec_update(
      "UPDATE courses SET flags = #{Course.connection.quote(stale)} WHERE id = #{course.id}"
    )
  end

  let!(:stale_course) do
    course = create(:course, start: 1.day.ago, end: 2.months.from_now, slug: 'Stale/Course',
                             flags: { academic_system: nil }.merge(seven_second_log))
    make_flags_stale(course)
    course
  end

  let!(:healthy_course) do
    create(:course, start: 1.day.ago, end: 2.months.from_now, slug: 'Healthy/Course',
                    flags: seven_second_log)
  end

  before do
    allow(Sentry).to receive(:capture_exception)
    allow(Sentry).to receive(:capture_message)
    allow(CourseDataUpdateWorker).to receive(:update_course)
  end

  it 'raises from add_flag for the stale course, confirming the setup is faithful' do
    loaded = Course.find(stale_course.id)
    loaded.flags

    expect { loaded.add_flag(key: :longest_update, value: 7) }
      .to raise_error(RuntimeError, /Locking a record with unpersisted changes/)
  end

  it 'completes the pass and still enqueues the healthy course' do
    expect { described_class.new }.not_to raise_error

    expect(CourseDataUpdateWorker).to have_received(:update_course)
      .with(course_id: healthy_course.id, queue: anything)
  end

  it 'reports the stale course to sentry with flags named as the dirty attribute' do
    described_class.new

    expect(Sentry).to have_received(:capture_exception)
      .with(instance_of(RuntimeError),
            extra: hash_including(course_id: stale_course.id, slug: 'Stale/Course',
                                  changed: ['flags']))
  end

  # The containment saves the pass, but the stale course itself is still never
  # updated: queue_for raises before CourseDataUpdateWorker is reached. Only the
  # data cleanup restores it.
  it 'still does not enqueue the stale course itself' do
    described_class.new

    expect(CourseDataUpdateWorker).not_to have_received(:update_course)
      .with(course_id: stale_course.id, queue: anything)
  end
end
```

</details>

## Merging this does not restore the two known-bad courses

Because `queue_for` raises before `CourseDataUpdateWorker.update_course`, the offending course was still never enqueued. The containment protects the rest of the pass — on the 2026-09-20 data, roughly 920 of 940 courses — but 15518 and 18300 could not resume updating until their rows were rewritten. That is a data fix, not a code fix.

The script below has now been run against both production databases. Neither reported any row still stale after the rewrite, so the repair held in production as it did in local testing. It is worth re-running after any future Ruby or Psych upgrade, since that is the only thing that can reintroduce the drift.

<details>
<summary>Cleanup script, as run against production</summary>

```ruby
# Rewrites the `flags` column of courses whose stored YAML no longer round-trips.
# Run once per production server, and again after any Ruby/Psych upgrade.

# Reading `flags` is load-bearing: Rails only compares a serialized column against
# its stored bytes once the attribute has been read, so an unread `flags` looks
# clean whatever is in the column.
def stale_flags?(course)
  course.flags
  course.will_save_change_to_flags?
end

stale_ids = []
Course.select(:id, :flags).find_each(batch_size: 500) do |course|
  stale_ids << course.id if stale_flags?(course)
end
puts "Courses with stale flags: #{stale_ids.count}"

rewritten = []
still_stale = []

stale_ids.each do |id|
  # Re-read under a row lock rather than storing the copy the scan loaded: the
  # scheduler writes flags constantly, and writing back a copy read minutes earlier
  # would drop whatever landed in between (T437639). `Course.lock.find` builds a
  # fresh record, so unlike `with_lock` it cannot trip over the phantom change that
  # makes these rows stale in the first place.
  Course.transaction do
    locked = Course.lock.find(id)
    # update_column writes this column and nothing else: no validations, no
    # callbacks, no updated_at, no other attribute of the record.
    locked.update_column(:flags, locked.flags)
  end

  (stale_flags?(Course.select(:id, :flags).find(id)) ? still_stale : rewritten) << id
end

puts "Rewritten: #{rewritten.count}"
puts "Still stale, inspect by hand: #{still_stale.inspect}" if still_stale.any?
```

</details>

## On enqueue failures now being non-fatal

Answering the question raised in the description: the trade looks right, but there is a gap alongside it. `run_update` has its success-path `log_end_of_update` commented out, so a pass in which *every* course failed would now emit one `capture_exception` per course and no summary, and would report success. A count check at the end of the pass would close that cheaply.

Minor, not blocking: `course.changed` in `report_enqueue_failure` is only meaningful because `queue_for` happens to read `flags` before anything can raise. That holds today, but it makes the diagnostic quietly dependent on call order.

---

*Drafted in a Claude Code (Opus 5) session of roughly a dozen exchanges. Sage directed the investigation and supplied every piece of production evidence, including the console output behind the byte comparison above, and ran the cleanup script against both production databases — neither reported a remaining stale row. Verified locally by the session: the round-trip stability of each plausible `update_logs` value type, the cleanup script against a seeded stale row, and the exploratory spec passing on this branch and failing on master with the production traceback. The script and the spec are AI-written; Sage read and ran the script, but has not reviewed the spec line by line.*

Issue #7082, opened 2026-09-21
Exclude Wikidata edits from "Words added" and "Bytes added" stats


ragesoss:
## Problem

"Words added" is `WordCount.from_characters(character_sum)`: bytes divided by 5.175, an English Wikipedia prose estimate (`lib/word_count.rb`). The `character_sum` it divides is summed across **every** wiki a course tracks, including Wikidata. Wikidata revisions are JSON entity diffs, so a single claim can add hundreds of bytes and register as dozens of "words". Courses whose home wiki is English Wikipedia but also track Wikidata therefore show inflated prose numbers, and every campaign containing such a course inherits the inflation through `SUM(character_sum)`.

Wikidata work is already reported separately: `UpdateWikidataStatsTimeslice` writes claims/labels/descriptions/etc. into `CourseStat.stats_hash['www.wikidata.org']`, and the `WikidataOverviewStats` component renders that block on the course overview and the campaign page. Those stats are the right measure of Wikidata editing. The word and byte counts are meant to represent prose contributions to Wikipedia and other wikitext wikis, so Wikidata should be excluded from them entirely.

## Where the number comes from

Per-wiki rows are already separated. The leak happens at the two aggregation steps that collapse wikis into one cached column:

```
revisions (per wiki)
  └─ CourseUserWikiTimeslice.character_sum_ms/us/draft   (course × user × wiki × slice)
       ├─ CourseWikiTimeslice.character_sum               (course × wiki × slice)
       │    └─ [A] CourseCacheManager#update_cache_from_timeslices
       │           courses.character_sum = SUM over ALL wikis          ← leak
       └─ [B] CoursesUsers#update_values_from_timeslices
              courses_users.character_sum_* = SUM over ALL wikis       ← leak
```

Everything downstream reads those two cached columns; nothing else recomputes from timeslices.

[...]

PR #7083, opened 2026-09-21
Exclude Wikidata edits from words added and bytes added stats

ragesoss:
## What this PR does

Addresses #7082.

"Words Added" divides a course's cached character sum by 5.175, a bytes-per-word ratio derived from English Wikipedia prose. That character sum was summed across every wiki a course tracks, including Wikidata. Wikidata revisions are JSON entity diffs, so a single claim can add hundreds of bytes and register as dozens of words. Courses with an English Wikipedia home wiki that also tracked Wikidata reported inflated prose totals, and every campaign containing such a course inherited the inflation.

Wikidata contributions are already reported separately, through the wikidata stats block that `UpdateWikidataStatsTimeslice` writes into `CourseStat`. So this excludes Wikidata byte counts from the word and byte stats outright, rather than reweighting them. Three aggregation sites collapse per-wiki timeslices into a displayed number, and all three now skip excluded wikis: the cached `courses.character_sum`, the three per-student character sums on `courses_users`, and the per-namespace word count that feeds the course overview's detailed stats tabs.

Filtering at those choke points means the campaign totals, the programs table sort and word count filter, the productive and active course alerts, the term recap email gate, the Salesforce push, and the CSV exports all follow without being touched individually. Reference counts, revision counts and per-article character sums are deliberately unchanged: references were never affected (the reference counter API rejects Wikidata), an edit is an edit on any wiki, and a per-article row already names its wiki.

Scale of the problem, measured on this branch: the course in the `course_update` VCR cassette recorded 7991 characters, of which 124 came from English Wikipedia over 2 revisions and 7867 from Wikidata over 27. Its reported word count was 98% Wikidata.

Archived courses are left alone by design, so past campaigns keep their current totals. Courses still inside their update window correct themselves on the next `UpdateCourseStats` run.

## AI usage

Claude Code (Opus 5) wrote all of the code, specs and commit message in this PR, under my direction. I described the problem and the goal, chose the design tradeoffs, and supplied the user-facing tooltip wording myself; no interface text here was written by AI, per the project's guidelines.

How it went, specifically. I first asked for a plan rather than code, and the agent inventoried the data flow from timeslices through every consumer before proposing anything. That plan named four open questions, which I answered in one message: include the per-student sums, amend the tooltips with wording I supplied, leave archived courses alone rather than backfilling, and apply the same exclusion to "Bytes Added". The plan was then posted as issue #7082 and implemented on this branch.

The plan's inventory turned out to be incomplete. During implementation the agent grepped for remaining character-sum aggregations and found `UpdateWikiNamespaceStatsTimeslice#word_count`, a third site that would have kept showing an extrapolated Wikidata word count on a `www.wikidata.org-namespace-0` stats tab after the other two fixes landed. That is a good illustration of why the plan-first step and the follow-up check both mattered; issue #7082 does not mention that third site.

Two spec failures during the work were the agent's own (new examples created a Wikidata wiki without `stub_wiki_validation`, so they tried to reach the MediaWiki API). A third was real: the cassette expectations above. Rather than pasting in whatever the failing run reported, the agent wrote a throwaway spec to print the per-wiki breakdown, confirmed 124 was the English Wikipedia portion, and deleted it. Each new spec example was also run with the exclusion disabled to confirm it actually fails without the fix.

Verification: every spec in the suite that mentions `character_sum` or `word_count` passes, 502 examples, and RuboCop is clean on all changed files. The screenshots below were captured by a disposable spec that seeds one course tracking both English Wikipedia and Wikidata, run once on master and once on this branch against identical data.

Scale of effort: a single commit, written and tested in one sitting of roughly an hour, across four short instructions from me. Reviewers should calibrate accordingly and read the aggregation changes closely.

The "What this PR does" summary and the analysis above were also drafted by Claude Code using the `prepare-pr` skill, and may contain errors.

## Screenshots

Seeded with one course tracking both English Wikipedia and Wikidata: 29,000 characters of English Wikipedia prose over 24 edits, and 247,000 bytes of Wikidata entity diffs over 310 edits. The same fixture data was used for both columns.

**01 course overview words added**

| Before | After |
|--------|-------|
| before | after |

Words Added falls from 53.3K to 5.6K. Total Edits stays at 334, since revision counts still include Wikidata.

**02 course roster words added**

| Before | After |
|--------|-------|
| before | after |

The per-student mainspace figure falls from 53,526 to 5,604. The userspace and draft figures are unchanged.

**03 campaign overview words added**

| Before | After |
|--------|-------|
| before | after |

**04 campaign programs word count tooltip**

| Before | After |
|--------|-------|
| before | after |

The tooltip gains the sentence "(Wikidata edits are not included.)". Only `en.yml` is changed; other locales come from translatewiki.

- **Archived courses keep their old numbers.** This is a deliberate decision, not an oversight. Active courses self-correct on their next update, but courses past their update window keep inflated caches, so a past campaign containing one will still show an inflated total. On the Programs & Events Dashboard this affects a lot of old courses, since the historical `courses_wikis:migrate` task attached Wikidata to every course that existed at the time. If we ever want those corrected, archived courses that ran at least one timeslice update can be recomputed exactly and idempotently from their non-Wikidata timeslices with `update_columns`. It must not go through `update_cache_from_timeslices`, whose `view_sum` calculation uses `DATEDIFF(UTC_TIMESTAMP(), first_revision)` and would drift on ended courses.
- **Deploy note for the P&E Dashboard.** The Sidekiq hosts run `UpdateCourseStats` and are not deployed by Capistrano. Until they get their manual `git pull` and `bundle install`, active courses there will keep recomputing the old way.
- **Expect the numbers to drop visibly** on any course or campaign that tracks Wikidata. Worth knowing before deploy so the drop reads as the fix rather than as a regression.
- **`word_count_doc_wikidata` is left alone.** Both the course and generic variants of that string appear to be unreachable, because the Items list falls back to "Characters added" for Wikidata home wikis, which have no bytes-per-word figure. Not touched here; possible cleanup later.
- `WordCount.excluded_wiki_ids` runs a small query per call rather than memoizing, since wiki ids change between test examples and the query is tiny next to the work these methods already do. Worth a look if it ever shows up in a profile.


ragesoss:
Automated review notes follow. They are AI-generated and provisional: read, verify, and use or ignore them as judgment dictates.

**Wikidata-home courses lose their "Characters Added" figures.** When the course's home wiki is Wikidata, nearly all timeslices belong to Wikidata, so `character_sum_ms`, `_us`, `_draft` and `course.character_sum` all fall to 0. The roster for such a course has no `home_wiki_bytes_per_word`, so it shows "Characters Added (item | user | draft)" from those raw sums (`ContentAdded.jsx`, `student_list_keys.js`). That column would read `0 | 0 | 0`. The embed stats page, the admin course row, the programs "characters" column, the CSV exports and the user-profile characters figure are affected the same way. A contrasting-case spec (a Wikidata-home course with one 12,000-character Wikidata timeslice) gives 12000 on master and 0 on this branch for both the course and the student sums. The revision count is 40 on both. One possible fix is to skip the exclusion when the course's home wiki is itself an excluded project. Sage has judged this acceptable and not a merge blocker.

**Minor:** the `.compact` in `UpdateWikiNamespaceStatsTimeslice#update_stats` drops any nil stat, not only `word_count`. That's harmless today, but it's broader than the comment says.

**Existing bug, out of scope:** the namespace stats service writes `reference_count` and `view_count`, but `namespace_overview_stats.jsx` reads `references_count` and `views_count`. References and views therefore never render on the namespace tabs. Filed as #7086.

The PR's specs and the related existing specs pass locally (80 examples), and RuboCop is clean on the changed files.

*Drafted in a Claude Code session (~15 min, 2 interactions). Claude Code wrote and ran the contrasting-case spec locally. Sage read the findings and decided the Wikidata-home behavior is acceptable, but did not independently verify them.*

